PR #923, opened 2016-02-26
Revert "Use attribute translation orders"

jhawthorn:
Reverts solidusio/solidus#795

This caused master builds to fail https://circleci.com/gh/solidusio/solidus/2303

PR #795, opened 2016-02-03
Use attribute translation orders

Murph33:
This is to improve the Orders views ease of localization by using model attribute values more in translations as opposed to generic values.  

This is an ongoing effort as discussed in #735.

Issue #735, opened 2016-01-25
Use ActiveRecord translations for admin forms


cbrunsdon:
This is a meta-issue to track progress against the PR's required to move the admin translations to us the activerecord attributes.

The admin is far too unfriendly to i18n and we need to ensure it works properly for non english speakers, and the best step forward is to use the rails-provided translation mechanisms properly.

This is meant to carry through the exploratory work started in #549.

tvdeyen:
:+1: Let me know, if you need any help

Murph33:
Will do, thanks!

Murph33:
So I feel totally happy with the first commit I just have one thing I want to touch base on before I set up a PR.

https://github.com/magiclabs/solidus/commit/056dfdfb74ef137fc5d05454016df9176f3e8eaf#diff-3

Here you've changed away from including the span using a content tag as part of the label to just writing it out explicitly.  The only functional difference is obviously clicking the * now won't select the field.  Just wanted to address this other change before putting in a PR since it isn't really part of switching to better I18n use.  I don't' know if across the entire project this will increase or decrease consistency or if one of these is generally a better practice we want to use.

tvdeyen:
@Murph33 Yeah, I changed that primarily for performance reasons.There is no need to create a static HTML tags with costly rails view helpers. 

But, this is out of scope of this PR, so I guess we should skip this in order to not pollute the diff with non related stuff. What are @cbrunsdon and @jhawthorn thinking?

cbrunsdon:
Honestly I'm fine with it in this PR. 
Its the same output and more readable. Just do it in a separate commit and we're good IMO.

Murph33:
:+1: Will do.

jhawthorn:
@tvdeyen I have a different plan to remove the `*`. Those are a pain point for me because it messes up capybara's matching: you might have to `fill_in "Name *", with: "My Name"`. It also pollutes our markup with something which IMO is a style concern. I plan to open a PR where we change it to

``` html
<label class="required">Name</label>
```

``` css
label.required:after {
  content: " *";
}
```

tvdeyen:
@jhawthorn That's why the `<span>` is not part of the label in my change. I'm also fine with your solution. Easier to override for people that don't want the `*` in the forms.

Murph33:
Appears to be some orphan Spree.t(:back_to_countries_list) translation that doesn't even have an entry in the en.yml  
It only shows up on the admin page when you're adding or editing a country
[...]

PR #924, opened 2016-02-26
Use model attribute translations in order views.

Murph33:
This makes better use of I18n by using translations placed on specific model attributes.

This is part of an ongoing effort to improve I18n as discussed in #735.  

This was previously merged in #795 but removed temporarily in #923 because of breaking a test.  The test has been modified to reflect the changes this PR had introduced.


cbrunsdon:
Looks good to me, thanks @Murph33 :+1:

jhawthorn:
:+1:

PR #927, opened 2016-02-26
Alias cost to amount in ShippingRate

mamhoff:
We have a more or less pretty system of making money objects,
and Spree::ShippingRate wasn't using it. Also, I think nomenclature-wise
`amount` is a much better word than `cost` - the first times I came
across `cost` I always though it was somehow different from amount,
maybe negative or something like that.


Senjai:
Sure, seems reasonable :+1:

jhawthorn:
:+1:

cbrunsdon:
Good by me too, thanks :+1:

PR #929, opened 2016-02-26
Cleanup use of private

jhawthorn:
This fixes serveral incorrect uses of `private`, which rubocop was helpfully pointing out.


backend/app/controllers/spree/admin/variants_controller.rb
```diff
@@ -52,8 +52,6 @@ def load_option_types_values
         @option_values = @option_types.flat_map(&:option_values).uniq(&:presentation)
       end
 
-      private
```
Senjai: Why is this incorrect?

Senjai:
:+1:

cbrunsdon:
good by me thanks :+1:

gmacdougall:
:+1:

PR #933, opened 2016-02-29
Make cart taxes configurable via `Spree::Store` object

mamhoff:
In VAT countries, carts (orders without addresses) have to be shown with
adjustments for the country whose taxes the cart's prices supposedly include.
This might differ from `Spree::Store` to `Spree::Store`. We're introducting
the `cart_tax_country_iso` setting on Spree::Store for this purpose.

Previously the setting for what country any prices include
Spree::Zone.default_tax. That, however, would *also* implicitly tag all
prices in Spree as including the taxes from that zone. Introducing the cart
tax setting on Spree::Store relieves that boolean of some of its
responsibilities.

This PR also introduces the `Spree::TaxRate.for_address` scope. It's not fully being used here yet, but that's coming up later.

### DIFF
--- a/core/app/models/spree/store.rb
+++ b/core/app/models/spree/store.rb
@@ -23,6 +23,11 @@ module Spree
       where(default: true).first || new
     end
 
+    def default_cart_tax_location
+      @default_cart_tax_location ||=
+        Spree::Tax::TaxLocation.new(country: Spree::Country.find_by(iso: cart_tax_country_iso))
+    end
+
     private
 
     def ensure_default_exists_and_is_unique
--- a/core/app/models/spree/tax/tax_location.rb
+++ b/core/app/models/spree/tax/tax_location.rb
@@ -1,0 +1,33 @@
+module Spree
+  module Tax
+    # A class exclusively used as a drop-in replacement for a default tax address.
+    # It responds to `:country_id` and `:state_id`.
+    #
+    # @attr_reader [Integer] country_id the ID of a Spree::Country object
+    # @attr_reader [Integer] state_id the ID of a Spree::State object
+    class TaxLocation
+      attr_reader :country_id, :state_id
+
+      # Create a new TaxLocation object
+      #
+      # @see Spree::Zone.for_address
+      #
+      # @param [Spree::Country] country a Spree::Country object, default: nil
+      # @param [Spree::State] state a Spree::State object, default: nil
+      #
+      # @return [Spree::Tax::TaxLocation] a Spree::Tax::TaxLocation object
+      def initialize(country: nil, state: nil)
+        @country_id = country && country.id
+        @state_id = state && state.id
+      end
+
+      def ==(other)
+        state_id == other.state_id && country_id == other.country_id
+      end
+
+      def empty?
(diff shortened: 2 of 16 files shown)


mamhoff:
Added changelog entry, rebased on current master. Ready for review!

core/app/models/spree/store.rb
```diff
@@ -23,6 +23,12 @@ def self.default
       where(default: true).first || new
     end
 
+    def default_cart_tax_location
+      @default_cart_tax_location ||= begin
```
gmacdougall: Why begin?

core/app/models/spree/tax/tax_location.rb
```diff
@@ -0,0 +1,35 @@
...
+      #
+      # @see Spree::Zone.for_address
+      #
+      # @param [Integer] country_id the ID of a Spree::Country object, default: nil
+      # @param [Integer] state_id the ID of a Spree::State object, default: nil
+      # @param [Spree::Country] country a Spree::Country object, default: nil
+      # @param [Spree::State] state a Spree::State object, default: nil
+      #
+      # @return [Spree::Tax::TaxLocation] a Spree::Tax::TaxLocation object
+      def initialize(country: nil, country_id: nil, state: nil, state_id: nil)
```
gmacdougall: I would prefer that we deal strictly with objects rather than objects or id's. It makes things a bit simpler. It can likely be more efficient by just using the ID's, but I'd prefer the simpler code for now.

core/db/migrate/20160229133259_add_cart_tax_country_iso_to_spree_store.rb
```diff
@@ -0,0 +1,5 @@
+class AddCartTaxCountryIsoToSpreeStore < ActiveRecord::Migration
+  def change
+    add_column :spree_stores, :cart_tax_country_iso, :string, null: false, default: ""
```
gmacdougall: For adding this field, I would prefer allowing nulls, and setting the default to be null. Then we could validate that the country ISO is within the list of valid ISO's or nil.

core/spec/models/spree/tax_rate_spec.rb
```diff
@@ -1,6 +1,77 @@
...
+
+    subject(:rates_for_address) { Spree::TaxRate.for_address(address) }
+
+    context 'when address is in germany' do
+      let(:address) { create(:address, country_iso_code: "DE") }
+
+      it 'returns the eu and the german rate' do
+        expect(rates_for_address).to include(german_tax)
+        expect(rates_for_address).to include(eu_tax)
+        expect(rates_for_address.length).to eq(2)
```
gmacdougall: I'd rather do this with a single assertion:

```Ruby
is_expected.to contain_exactly(german_tax, eu_tax)
```

core/app/models/spree/store.rb
mamhoff: I thought because of the multiline expression. But - having written code that doesn't need it, I guess it can go here as well.

jhawthorn:
Looks reasonable to me :+1:

gmacdougall:
:+1:

PR #934, opened 2016-02-29
Add missing #orders relation on Spree::Store

mamhoff:
There's a `belongs_to` association on Spree::Order, but no corresponding `has_many` association on the Spree::Store.

The commit adds the missing `has_many` line.

### DIFF
--- a/core/app/models/spree/store.rb
+++ b/core/app/models/spree/store.rb
@@ -2,6 +2,7 @@ module Spree
   class Store < Spree::Base
     has_many :store_payment_methods, inverse_of: :store
     has_many :payment_methods, through: :store_payment_methods
+    has_many :orders, class_name: "Spree::Order"
 
     validates :code, presence: true, uniqueness: { allow_blank: true }
     validates :name, presence: true


tvdeyen:
:+1:

core/spec/models/spree/store_spec.rb
```diff
@@ -50,4 +50,14 @@
       expect(store.default).not_to be true
     end
   end
+
+  describe '#orders' do
```
jhawthorn: This should probably be `'.orders'` since it is on the class, not the instance.

mamhoff: No, it's on the instance :) see two lines below...

Senjai: I'm pro adding an association for this. But I don't think we need to test rails here. We're not really testing Solidus.

cbrunsdon:
:+1: with or without the spec

core/spec/models/spree/store_spec.rb
jordan-brough: +1 on @Senjai's comment

mamhoff:
Amended, no spec testing Rails anymore.

Senjai:
:shipit:

jhawthorn:
:+1:

PR #936, opened 2016-02-29
Set the order totals explicitly on promo_code_spec

cbrunsdon:
Previously we relied on magic values in the factories for the values,
which we tested on. This change makes it clearer what values we should
be getting before and after the promo is applied.


Senjai:
Seems reasonable :+1:

gmacdougall:
:+1:

jhawthorn:
:+1:

PR #937, opened 2016-02-29
Remove unclear test in unreturned_item_charger

cbrunsdon:
This test does not add to readability or understandability of the
behavior of the unreturned item exchanger. It does not help us ensure
its behaving as expected, and is higher value to remove than keep in.


Senjai:
:+1:

jhawthorn:
:+1: Was originally added believing that shipments were updated as part of the unreturned item change. It turns out they are not and it is explicitly disabled in `create_proposed_shipments`

gmacdougall:
:+1:

Issue #908, opened 2016-02-25
(Admin) Cannot add a payment for less than full order amount if order is not completed


dholdren:
### Reproduce:

(Using the generated sandbox app, tested on both v1.0 and v1.2 branches)
Take an order at the "Payment" stage with no payments.
As an admin, try to add a payment for less than the full amount. (e.g. a split payment)
### Expecting:

A new payment for the entered amount.
### Actual Result:

A new payment for the full amount of the order.

![](http://g.recordit.co/NXK14UffbJ.gif)
### Probable cause:

It looks like this line is the culprit:
https://github.com/solidusio/solidus/blob/v1.0/core/app/models/spree/order.rb#L674

cbrunsdon:
@dholdren and I chatted this through in slack and the decision was to not modify existing payments if no new store credit payments were created in that call.

Long term we need to refactor and clean it up, but that should be a good fix for now.

Thanks for hunting down the code for the specific issue @dholdren, super appreciated.

mamhoff:
@cbrunsdon Can we close this issue or do you want to keep it around until we have refactored and cleaned up?

dholdren:
Closing, this is no longer happening.

PR #938, opened 2016-03-01
Quick return out of add_store_credit_payment

cbrunsdon:
If we don't have a user, we don't need to be inside of
add_store_credit_payment.

If we have a user that doesn't have any credit, and our order doesn't
have any credit, we don't need to be inside of add_store_credit_payment.

To accomplish this we need to *change existing behavior* of
add_store_credit_payment to not add the error message given the previous
changes. I argue its previous implementation was overstepping its bounds
in assuming it should know when and how to add that error onto the
order.

Related to #908

### DIFF
--- a/backend/spec/features/admin/orders/new_order_spec.rb
+++ b/backend/spec/features/admin/orders/new_order_spec.rb
@@ -60,6 +60,35 @@ describe "New Order", type: :feature do
     end
   end
 
+  it 'can create split payments', js: true do
+    click_on 'Cart'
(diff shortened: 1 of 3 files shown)


jhawthorn:
I think we need a regression spec for the issues this is fixing.

cbrunsdon:
Added regression feature spec, felt that was the most appropriate way to do the regression.

dholdren:
👍 awesome. I'd love to see this merged so I can remove my local hack

jhawthorn:
Looks good to me :+1:

PR #939, opened 2016-03-01
Clean up price calculations

jhawthorn:
There are several conditions around not having a currency, which is not something which can happen in our system.

Removes a dead branch which calls `price_modifier_amount`, and cleans up some other areas.


cbrunsdon:
Looks fine to me, thanks @jhawthorn :+1:

gmacdougall:
:+1:

PR #940, opened 2016-03-01
Add currency preference to item total rule

jhawthorn:
Making a rule based on an item total makes no sense without a currency.
The default value of currency should hopefully keep this rule working
for anyone currently using it, as they are probable on a single currency
store.


cbrunsdon:
Do we need to take this in via the interface now too?

jhawthorn:
@cbrunsdon done

mamhoff:
:+1: Really nice.

PR #904, opened 2016-02-24
Add Spree::ShippingRateTax model

mamhoff:
Previously, shipping rate taxes were calculated on the fly every time
a shipping rate would be displayed. Now, shipping rate taxes are stored
on a dedicated table to look up.

There is a new model Spree::ShippingRateTax where the taxes are stored,
and a new Spree::Tax::ShippingRateTaxer that builds those taxes from within
Spree::Stock::Estimator.

The shipping rate taxer class can be exchanged for a custom estimator class
using the new Spree::Appconfiguration.shipping_rate_taxer_class preference.

PR #941, opened 2016-03-01
Calculate pre tax amounts on the fly

mamhoff:
This takes away some caching and slows down some of the tax calculations
in VAT countries, but it also speeds up tax calculation for Sales tax countries.

It also makes sure that for line items and shipments, the following always
holds true:

```
discounted_amount - included_tax_total = pre_tax_amount
```

The slowdown is significant, but will be remedied by two planned improvements
in the very near future:
* Looking up rates by address
* Separating handling of default taxes vs. order taxes

I had to adjust the spec setup for the default tax calculator spec and realised the order (!) calculator code has a bug: It takes the item's `total` rather than its `discounted_amount`, applying tax rates twice. Yay.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 ## Solidus 1.3.0 (unreleased)
 
+*   Removed `pre_tax_amount` column from line item and shipment tables
+
+    This column was previously used as a caching column in the process of
+    calculating VATs. Its value should have been (but wasn't) always the same as
+    `discounted_amount - included_tax_total`. It's been replaced with a method
+    that does just that. [#941](https://github.com/solidusio/solidus/pull/941)
+
+*   Renamed return item `pre_tax_amount` column to `amount`
+
+    The naming and functioning of this column was inconsistent with how
+    shipments and line items work: In those models, the base from which we
+    calculate everything is the `amount`. The ReturnItem now works just like
+    a line item.
+
+    Usability-wise, this change entails that for VAT countries, when creating
+    a refund for an order including VAT, you now have to enter the amount
+    you want to refund including VAT. This is what a backend user working
+    with prices including tax would expect.
+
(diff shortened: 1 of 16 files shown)


mamhoff:
Despite the horrid performance implications, I really want this in. It rids `Spree::TaxRate.rb` of any calculations except the bizarre `amount * -1` bit, and really gives a lot of power to the calculator, where it belongs. The performance can and will improve. It's also a prerequisite for #904, where I've baked these commits in in such a way that if this is merged, the other will be shorter. @jhawthorn @cbrunsdon @jordan-brough

tvdeyen:
Did you ran any benchmarks? Is this really so much slower?

mamhoff:
The taxation integration test suite, which is really exactly testing this lots and lots of times, goes from ~45s to ~53 seconds. Considering that much of that time is spent doing spec setup, yes, this must be a lot slower. 

The real reason why is that in the adjustments callback cycle, every adjustments amount get recalculated an insane number of times, which is why the caching (on the `pre_tax_amount` column) helped so much. 

This is a huge priority, but I think before this was the wrong spot in which to do the caching. If it's really necessary (I believe the more sanity we add to this code, the faster it will run).

jordan-brough:
:+1: I think the simplification here is worth the temporary slowdown, especially since @mamhoff has already done so much of the work to speed things up in upcoming commits.  I feel like this commit is helping detangle the code and helps me be able understand the existing code more clearly.

jordan-brough:
We'd probably want some sort of changelog note about removing the columns yes?

cbrunsdon:
Yep, I'm getting on the :+1: train too, excellent work as always @mamhoff. Temporary slowdown for vat is a small price to pay for moving this forward.

jhawthorn:
:+1:

gmacdougall:
:+1:

PR #942, opened 2016-03-01
Set billing address to default address

yeonhoyoon:
Set `bill_address` to default address, if user has one.

Currently when an order goes to `address` state, only `ship_address` gets filled in from a previously entered address. This is because `ship_address` is an alias for `default_address` in the `UserAddressBook` module.
https://github.com/solidusio/solidus/blob/master/core/app/models/concerns/spree/user_address_book.rb#L53
However, `bill_address` is not an alias, so it is not prefilled and user has to enter it every time. Rather than making it into an alias, just try default_address in `assign_default_address!` if user has one.

### DIFF
--- a/core/app/models/spree/order/checkout.rb
+++ b/core/app/models/spree/order/checkout.rb
@@ -295,11 +295,13 @@ module Spree
 
           def assign_default_addresses!
             if user
+              bill_address = (user.bill_address || user.default_address)
+              ship_address = (user.ship_address || user.default_address)
               # this is one of 2 places still using User#bill_address
-              self.bill_address ||= user.bill_address if user.bill_address.try!(:valid?)
+              self.bill_address ||= bill_address if bill_address.try!(:valid?)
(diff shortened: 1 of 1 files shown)


jgujgu:
Great catch...I was going to figure out a good way to do this myself, except less good. When I am beaten to the punch, it is usually because of another Asian guy in glasses.

yeonhoyoon:
:sunglasses:

mamhoff:
This is :+1: by me. However, would you care to merge the second (fixup) commit into the first one?

yeonhoyoon:
@mamhoff sure, squashed them.

mamhoff:
👍

tvdeyen:
Thanks! I would love to see tests for this

mamhoff:
@yeonhoyoon This is good to go, except it needs some specs.

tvdeyen:

PR #944, opened 2016-03-01
Split up two SQL statements in migration

jhawthorn:
This previously didn't work correctly under MySQL. In Rails 4.2.6.rc1 this raises an exception, in previous versions the second statement was ignored.


mamhoff:
Super, thank you. I didn't know two SQL statements in the same `execute` thingy don't work.

gmacdougall:
:+1:

PR #945, opened 2016-03-02
Get available calculators from configuration.

yeonhoyoon:
Currently, admin page pulls the list of available calculators from calculators that are already present in the shipping methods in the database. If you delete all the shipping methods in the database, you cannot select calculator when trying to add a new shipping method.

### DIFF
--- a/backend/app/controllers/spree/admin/shipping_methods_controller.rb
+++ b/backend/app/controllers/spree/admin/shipping_methods_controller.rb
@@ -39,7 +39,7 @@ module Spree
       def load_data
         @available_zones = Zone.order(:name)
         @tax_categories = Spree::TaxCategory.order(:name)
-        @calculators = ShippingMethod.calculators.sort_by(&:name)
+        @calculators = Rails.application.config.spree.calculators.shipping_methods
       end
     end
   end
--- a/backend/app/controllers/spree/admin/tax_rates_controller.rb
+++ b/backend/app/controllers/spree/admin/tax_rates_controller.rb
@@ -8,7 +8,7 @@ module Spree
       def load_data
         @available_zones = Zone.order(:name)
         @available_categories = TaxCategory.order(:name)
-        @calculators = TaxRate.calculators.sort_by(&:name)
+        @calculators = Rails.application.config.spree.calculators.tax_rates
       end
     end
   end


mamhoff:
I don't think the bug you're reporting is actually a bug, as the class method `.calculators` on both models runs exactly what you're proposing: 
```
# (with an entirely empty database) 
2.1.5 :003 > Spree::TaxRate.calculators
 => [Spree::Calculator::DefaultTax(id: integer, type: string, calculable_id: integer, calculable_type: string, created_at: datetime, updated_at: datetime, preferences: text)] 
2.1.5 :004 > Spree::TaxRate.all
  Spree::TaxRate Load (2.2ms)  SELECT "spree_tax_rates".* FROM "spree_tax_rates" WHERE "spree_tax_rates"."deleted_at" IS NULL
 => #<ActiveRecord::Relation []> 
2.1.5 :005 > Spree::TaxRate.calculators
 => [Spree::Calculator::DefaultTax(id: integer, type: string, calculable_id: integer, calculable_type: string, created_at: datetime, updated_at: datetime, preferences: text)] 
2.1.5 :006 > Spree::ShippingMethod.calculators
 => [Spree::Calculator::Shipping::FlatPercentItemTotal(id: integer, type: string, calculable_id: integer, calculable_type: string, created_at: datetime, updated_at: datetime, preferences: text), Spree::Calculator::Shipping::FlatRate(id: integer, type: string, calculable_id: integer, calculable_type: string, created_at: datetime, updated_at: datetime, preferences: text), Spree::Calculator::Shipping::FlexiRate(id: integer, type: string, calculable_id: integer, calculable_type: string, created_at: datetime, updated_at: datetime, preferences: text), Spree::Calculator::Shipping::PerItem(id: integer, type: string, calculable_id: integer, calculable_type: string, created_at: datetime, updated_at: datetime, preferences: text), Spree::Calculator::Shipping::PriceSack(id: integer, type: string, calculable_id: integer, calculable_type: string, created_at: datetime, updated_at: datetime, preferences: text)] 
2.1.5 :007 > Spree::ShippingMethod.all
  Spree::ShippingMethod Load (1.4ms)  SELECT "spree_shipping_methods".* FROM "spree_shipping_methods" WHERE "spree_shipping_methods"."deleted_at" IS NULL
 => #<ActiveRecord::Relation []>  
```

The proposed change also removes the sorting, which is a pity.

jhawthorn:
I agree with mamhoff. `.calculator` is defined [in CalculatedAdjustments](https://github.com/solidusio/solidus/blob/master/core/app/models/concerns/spree/calculated_adjustments.rb#L10-L12). Maybe there was some confusion because a server restart is needed to update the calculator configuration.

If we're mistaken, please reopen.

yeonhoyoon:
@mamhoff @jhawthorn 
You're right, I didn't look into the definition of `ShippingMethod.calculators` and thought it was just getting it from the database. Calculators were not loading in the new shipping method page, so I thought this might have been the cause.. maybe it was because I loaded the spree sample without a restart.

Thanks for taking the time to check the pull request & making a comment.

yeonhoyoon:
In case it might help anyone in the future, it was the problem with the IDE I was using(IntelliJ). I had both my app and solidus's code in the same project, and the ruby environment was loading duplicate classes into object space.. so `Spree::Calculator::Shipping::FlatRate < Spree::ShippingCalculator` was returning false. I removed the solidus code from the project and it worked.

PR #946, opened 2016-03-02
Fix incorrect test in Taxon Rule

mtomov:
* The correct behaviour is to test the parent taxon set on the rule
    for eligibility against an order with a child taxon


gmacdougall:
If this behaviour was incorrect, how was the test able to pass?

mtomov:
The test was passing, because the behaviour is correct, and the test is the same as the [main one at the top](https://github.com/solidusio/solidus/blob/master/core/spec/models/spree/promotion/rules/taxon_spec.rb#L20)

This is meant to test an additional behaviour in addition to the basic taxon matching.

gmacdougall:
@mtomov Thanks for the clarification. Makes sense.

jhawthorn:
Makes sense. This test now matches the stated context.

PR #947, opened 2016-03-02
Move method definitions out of ActiveSupport::Concern included blocks

jhawthorn:
`ActiveSupport::Concern`'s included block should only include logic which needs a `class_eval` on the including object. For any methods which don't need dynamic definition at include time we can just define them in the module for instance methods, or inside a `class_methods` block (or a module named `ClassMethods`) for class methods.

### DIFF
--- a/core/app/models/concerns/spree/adjustment_source.rb
+++ b/core/app/models/concerns/spree/adjustment_source.rb
@@ -1,23 +1,19 @@
 module Spree
   module AdjustmentSource
-    extend ActiveSupport::Concern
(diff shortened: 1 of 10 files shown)


cbrunsdon:
This all looks right to me, and I can't think of how this would affect anyone using these as they are now, but I might be missing something? Is this completely transparent or is there anything we need to make people aware of while upgrading?

jhawthorn:
It will be slightly different (methods defined on module instead of directly on class), but I don't think that necessitates a CHANGELOG entry.

PR #948, opened 2016-03-02
Show empty shipments in the admin

jhawthorn:
Previously, empty shipments would be hidden from admins. If there is a shipment on the order in the database, it should be shown to admins so that they are aware of the true status of their order.

### DIFF
--- a/backend/app/views/spree/admin/orders/_shipment.html.erb
+++ b/backend/app/views/spree/admin/orders/_shipment.html.erb
@@ -74,7 +74,7 @@
           <% end %>
 
           <td class="actions">
-            <% if( (can? :update, shipment) and !shipment.shipped?) %>
+            <% if can?(:update, shipment) && !shipment.shipped? %>
               <%= link_to '', '#', :class => 'edit-method fa fa-edit no-text with-tip', :data => {:action => 'edit'}, :title => Spree.t('actions.edit') %>
             <% end %>
           </td>
--- a/backend/app/views/spree/admin/orders/_shipment.html.erb
+++ b/backend/app/views/spree/admin/orders/_shipment.html.erb
@@ -1,122 +1,120 @@
 <% manifest_items = Spree::ShippingManifest.new(inventory_units: shipment.inventory_units.where(carton_id: nil)).items %>
 
-<% unless manifest_items.empty? %>
-  <div id="<%= "shipment_#{shipment.id}" %>" class="js-shipment-edit" data-hook="admin_shipment_form">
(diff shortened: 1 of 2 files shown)


cbrunsdon:
This came out of a discussion in slack where someone asked why we don't delete the shipments if we delete the last items inside of them (on a .complete) order.

With further discussion we found out the larger problem was we would not show admins empty shipments on an order, so by deleting the last order admin's did not know there were now "empty" shipments on that order.

I am strongly against automatically deleting shipments (as we lose things like tracking numbers that might be associated to them, other potential associations, etc) and would rather we present admins with the ability to delete them if that is the desired outcome.

jhawthorn:
Even if we were to delete empty shipments, that would be even more of a reason to delete this check :wink:

cbrunsdon:
Aye, that was my way of being :+1: on this, sorry. It might be worth sitting @Mandily down and having a talk from a UI/UX point of view though.

The admin screens, what they're showing (shipments or line items or who knows?) has always been pretty hard to understand and comprehend. She might already have plans for the future.

backend/app/views/spree/admin/orders/_shipment.html.erb
```diff
@@ -1,123 +1,121 @@
...
-            <% if shipment.tracking.present? %>
-              <strong><%= Spree.t(:tracking) %>:</strong> <%= shipment.tracking %>
-            <% else %>
-              <%= Spree.t(:no_tracking_present) %>
-            <% end %>
-          </td>
           <td class="actions">
-            <% if can? :update, shipment %>
-              <%= link_to '', '#', :class => 'edit-tracking fa fa-edit no-text with-tip', :data => {:action => 'edit'}, :title => Spree.t('actions.edit') %>
+            <% if( (can? :update, shipment) and !shipment.shipped?) %>
```
mamhoff: I think if we used the `&&` operator here, we could remove some of the parens, making that `if`statement look less like a method call: 
```ruby
if can? :update, shipment && !shipment.shipped?
```

mamhoff:
I think this makes sense, even as an incremental thing before @Mandily redesigns the page entirely. :+1:

backend/app/views/spree/admin/orders/_shipment.html.erb
jhawthorn: This line actually wasn't changed here, but looks like it because github's diff is confused by the massive whitespace change.

I've made this change in a second commit.

mamhoff:
:+1:

PR #949, opened 2016-03-02
Remove decorator from api

jhawthorn:
Thanks to @ayb for pointing this out.


cbrunsdon:
:+1:, thanks @ayb.

Gotta love class_eval'ing our own codebase.

PR #950, opened 2016-03-03
Reduce the surface area of Spree::Stock::Coordinator

jordan-brough:
Make 'shipments' the only public method of the Coordinator. It's the
only method called on Coordinator outside of the specs.

This was prompted by me looking at the Coordinator code while
reviewing some of the taxes work and thinking it might be nice to
refactor it a bit, which is easier with a reduced surface area.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Solidus 1.3.0 (unreleased)
 
+*   Removed Spree::Stock::Coordinator#packages from the public interface.
+
 *   Removed Spree::BaseHelper#gem_available? and Spree::BaseHelper#current_spree_page?
 
     Both these methods were untested and not appropriate code to be in core. If you need these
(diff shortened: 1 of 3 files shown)


mamhoff:
I like this. This would mean that the shipping rates estimation could be part of the `to_shipment` method on the package class, and I could get rid of the strange `attr_accessor` in `Spree::ShippingRate`.

cbrunsdon:
I also like this but reckon it will need a changelog if it goes in, as we are technically immediately removing a method from the public API.

I have such a poor understanding of what a "package" does I'd rather this be the only thing accessible by a co-ordinator, and agree it would give us more flexibility to (eventually) make some meaningful changes behind the scenes if we can get this in.

jhawthorn:
I wouldn't mind `#packages` being public, as I could see that being useful, but I don't feel that strongly.

:+1:

jordan-brough:
Great!  I added a changelog entry.

jordan-brough:
Merged in f8b889a

PR #951, opened 2016-03-03
Disable Rubocop SymbolProc

jordan-brough:
I personally usually agree with this Cop but it doesn't seem worth
hassling people over.
https://github.com/bbatsov/rubocop/blob/v0.37.2/lib/rubocop/cop/style/symbol_proc.rb

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -187,6 +187,9 @@ Style/StringLiterals:
   Enabled: false
   StyleGuide: http://relaxed.ruby.style/#stylestringliterals
 
+Style/SymbolProc:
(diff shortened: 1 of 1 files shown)


cbrunsdon:
I have no opinion one way or the other, whatever consensus we end up with is good by me.

jhawthorn:
I agree. This rule is 99% of the time correct, but often in rails it requires making the code awkward.

For example it recommends replacing:
```
respond_to do |format|
  format.html
end
```

For example it recommends replacing:
```
respond_to(&:html)
```

PR #870, opened 2016-02-17
Add Tabs Component

graygilmore:
First step in https://github.com/solidusio/solidus/issues/633

I would love some feedback on the JS. I liked bringing this in house. I know there's a library out there that probably does this plus a million things but that's a road I always try to avoid going down.

- [x] Re-add the `<nav>` element: https://github.com/solidusio/solidus/commit/fecc715a4b3f10e09c5a3d4a3036fa81d9212bbe#commitcomment-16168308
- [x] Don't trigger the window's resize: https://github.com/solidusio/solidus/commit/8c67bc3d7ce719ed31a0cc4ff644b577d6a988d7#commitcomment-16167954
- [x] Drop the need for the `.tab` class: https://github.com/solidusio/solidus/commit/fecc715a4b3f10e09c5a3d4a3036fa81d9212bbe#commitcomment-16168251
- [x] Initial page load on Safari is miscalculating widths and not creating the dropdown.
- [x] Shorten the wording on the order submenu: https://github.com/solidusio/solidus/pull/870#issuecomment-185484812

## Before
screen shot 2016-02-17 at 3 58 23 pm

## After
screen shot 2016-02-17 at 3 57 46 pm

Issue #634, opened 2015-12-30
Group settings pages and move them into the left nav and tabs


Mandily:
I’d like to take the long list of settings and break it up into groups to make it easier to scan.  
* reorder nav items so that it follows the natural progression of an order
* remove right side nav
* create parent group items settings pages
* put pages in tabs

Below is the proposed information architecture of the settings and further down a mockup of what that would look like. 

![settings-ia](https://cloud.githubusercontent.com/assets/15271353/12056024/75112298-aee7-11e5-9296-e72f4a5b23eb.png)

Ignoring obvious brand/colour discrepancies, this is what the settings area of the site would look like:
![settings](https://cloud.githubusercontent.com/assets/15271353/12056038/839d3392-aee7-11e5-96a7-a49f51f99fba.png)

I want to make sure this is expendable because I know a lot of stores have more than just out of the box functionality stored in here. Please provide feedback in the comments below.

dfranciosi:
Neat :gem: :+1:

tvdeyen:
I like the idea of grouping the settings in logical groups. But this comes with a downside I want to discuss here. 

We won't reduce clicks. Removing clicks was one of the mayor reasons to move the individual settings into the main navigation. 

Any ideas in how we address this? I absolutely see the benefits of grouping the settings and like to see this implemented, but I also like the settings to be easily reachable. 

Second: I don't think Taxonomies is Store related. It should be its own "Page" don't you think?

cbrunsdon:
@tvdeyen The settings are so infrequently used I don't think we should be prioritizing click reduction

hhff:
agree w/ @cbrunsdon - I think expandable sections will really help the admin feel "less overwhelming", which IMO is a bigger problem at the moment

tvdeyen:
Yeah, yep, you're all right. A cleaner settings section > less clicks. Granted.

What about putting Taxonomies on the Store settings page?

Mandily:
I'm a couple of comments behind, but the reason I want to move the settings into the main navigation is because that's technically where it belongs. It never fit in the second level at the top in the old model so it was put into the right nav. 

Once we started talking about it here, we realized it would be easier to scan if it was broken up into groups. More useable is the end goal. Less clicks is one of the ways to get there, but so is grouping. 

As for taxonomies on the store settings page, I think if we started combining pages, we should look at all of the settings, which I'm not sure if we want to do right now. What if we created a category called products? Would it be useful for other things people are customizing their stores with?

[...]

PR #952, opened 2016-03-03
Replace settings configuration menu with separate tabs partials

Murph33:
This was done to implement issue #634.  It uses the tab functionality introduced in #870 (not yet merged) by @graygilmore.   The only changes from the plans outlined in #634 (as given by @Mandily) were to change refunds and reimbursements to checkout and to add adjustment reasons to that section.

# Before

before tabs

# After

after tabs


Murph33:
I was a bit hasty opening this since #870 isn't merged yet and if that implementation changes this will need to be rebased possibly drastically.  Closing for now.

PR #953, opened 2016-03-03
Fix left nav submenus sharing an id

Murph33:
Currently each submenu partial has the same id of "sub_nav".  While they aren't used anywhere in the codebase currently it is incredibly bad practice to have duplicate ID attributes on a page.


jhawthorn:
:+1:

tvdeyen:
Do we need `id`s after all? No css/js refers to it and integration point for extensions should be the `[data-hook]` anyways. 

So, I vote for removing them.

graygilmore:
I agree with @tvdeyen.

Murph33:
Took them out.  No point in having code we aren't using and don't need.

cbrunsdon:
I'm good by this one now :+1:

jhawthorn:
:+1:

PR #955, opened 2016-03-03
Add bootstrap v4.0.0-alpha.2 into the admin

jhawthorn:
The goal of this is to get bootstrap included into the admin with minimal changes. After this is merged we can begin replacing components as we see fit.

Rather than including the bootstrap-sass gem, I have simply added the scss into `vendor/bundle`. This is to avoid tying the version of bootstrap we use in the admin with the version stores use in their frontend, and is how we should be handing all admin assets from now on.

After including bootstrap, the only major issue I noticed was broken pagination. Rather than fighting with the bootstrap classes, I chose to remove our pagination component in favour of the one included with bootstrap. So the styles of that have changed (and probably need their colours fixed).

Issues:
* [x] `<th>` is now left aligned instead of centered
* [x] Button alignment is slightly off
* [x] Match pagination colour with rest of theme
* [x] Links should not have an underline
* [x] Add js files
* [x] Size of header has changed
* [x] `<legend>`s at the top of fieldsets have lost their line separator
* [x] margins on existing grids are slightly off


jhawthorn:
> The small misalignment of things everywhere is a no-go to me. I'd prefer to see us importing things one component at a time. What was your thought on that?

We spoke IRL, most of our misalignments are going to be due to bootstrap's reboot and typography, and it isn't much use to include bootstrap without those. I'm going to try to fix up as many issues as possible before merging and have created a checklist of issues.

CHANGELOG.md
```diff
@@ -19,6 +19,14 @@
 
     More information is available in the [VersionCake README](https://github.com/bwillis/versioncake)
 
+*   Bootstrap 4.0.0-alpha.2 is included into the admin.
+
+*   Pagination now uses an admin-specific kaminari theme, which uses the
+    bootstrap4 styles. If you have a custom admin page with pagination you can
+    use this style with the following.
+
+        <%= paginate @collection, theme: "admin" %>
```
tvdeyen: This theme should be called `solidus`

jhawthorn: Done. I went with "solidus_admin"

jhawthorn:
I think this is good to go :ship:

I spent a while browsing through the admin and can no longer find anything noticeably broken. Spacing has of course changed slightly in a few places, but nothing I've found looks out of place.

jhawthorn:
I replaced the a-la-carte individual JS files with the single `bootstrap.js` file. It was pointed out to me that it would be nice for extension developers to have the entire suite of bootstrap js components available. This is also a lot simpler to include anyways.

tvdeyen:
Why not just deliver the files, but not require them. Then we could have a slim JS file for "default" Solidus and extension developers are still able to require components they need. 

For instance, the `Tab` component won't be used by default Solidus admin, so why require it and bloat the JS file?

Also, I thought we agreed on porting one component at a time, so we basically don't need any JS component right know?

Am I missing something? /c @graygilmore

jhawthorn:
I prefer including the entire bootstrap.js file. I suspect it will be a point of confusion for users if some bootstrap features don't work (the bootstrap docs aren't as clear as they could be about needing to include js for the `data-toggle` features).

IMO if we were going to sweat a couple extra KB, we wouldn't be using bootstrap in the first place.

tvdeyen:
As discussed in slack we go with the whole file

:+1: :shipit: already!

tvdeyen:
YOLO!

forkata:
Thanks @jhawthorn didn't get a chance to chime in before this got merged but :+1:  great work!

PR #958, opened 2016-03-04
Remove delegate_belongs_to

cbrunsdon:
This was added back in 2010 when the project pulled it its "plugins".
Who here has been rails-ing long enough to remember "plugins"?

Anyway. We don't need it.


jhawthorn:
I'd prefer leaving this in. It was deprecated fairly recently and was widely used before that.

tvdeyen:
@cbrunsdon you mean these svn externals, right?

:+1: for ditching it. It has been deprecated long enough

jhawthorn:
> It has been deprecated long enough

I disagree, this was deprecated in solidus 1.1. I would like to recommend to users upgrading to solidus to upgrade directly to the latest version, skipping intermediates, which I can't do if this is removed.

tvdeyen:
@jhawthorn This means we can't ever remove deprecated stuff, because upgraders are encouraged to always use the latest version, right? How should this ever work?

Upgraders need to use the version they are currently on and go on version for version. Remove departed stuff. Upgrade to next and so on. 

Or do you have any other idea?

jhawthorn:
I'd like to remove deprecations in 2.0

tvdeyen:
Ok. Sounds reasonable to me.

Issue #630, opened 2015-12-29
Move page title and action buttons into a sticky header


Mandily:
Currently the page title and action buttons are scrollable with the rest of the page content. I’d like to place these items in a sticky area of the page so that the user can always see what page they’re on and have easy access to functions like adding new items to lists. 

The page title would be left aligned and the action buttons would be right aligned to the screen. The content in the page would remain fixed and centered. The page title would truncate if it was exceptionally long and the screen was too small for it to fit on one line, although if the user hovered over it they could see the full page title in a browser styled tooltip.

See the mockup for details. (ignore the left nav visual design for now)

Please let me know your thoughts in the comments below.

![6-sticky-header](https://cloud.githubusercontent.com/assets/15271353/12044198/75ef11de-ae42-11e5-8ae2-70e9af93e44a.png)

BenMorganIO:
:+1:

tvdeyen:
Would love to see this happening. This is a great improvement. 

As next step I would love to see the content not being centered and narrowed any more. It does not fit into the overall look anymore and is a waste of screen space. 

@cbrunsdon How do you plan to release all of these backend improvements? As the individual PRs are great, releasing them individually would lead to a very inconsistent UI. So, I guess that if all UI PRs landed, we will release them all together, right?

Mandily:
I haven't posted all 4 of the following issues up yet, but we're starting to plan for the 1.3 release.
Assuming we don't run into any problems along the way, we'd like to get a sticky header, 100% width, tabs and moving the settings right side navigation into 1.3.

cbrunsdon:
@tvdeyen What is up there now (menu on the left, some colour changes) are what we're targeting for 1.2 (which we're hoping to get out the door sooner rather than later).

But aye, what @Mandily said. Kevin is AFK for the week but Hawth, Mandily and myself sat down yesterday and tried to see what would target 1.3 release. I think we came up with a good balance of consistency but still reasonable to do in a single release. 

We'll have to play it by ear as the work and feedback comes in, but will keep the releases in mind.

tvdeyen:
Ok, sounds like a good plan. Besides the colour (British Columbia ^_^, love that) changes in #603. This would lead to a strange mix of colours in the admin UI. I would move the colour changes all together in one release (1.3 then). Also #629 would be a better candidate for 1.3.

Moving around the navigation and flash notices is ok IMO, but having two colour themes is not.

How do you think about it @cbrunsdon @Mandily ?

[...]

PR #959, opened 2016-03-04
Improve admin header

Murph33:
Make the header sticky and adjust the display of the page title and actions in response to #630.  I'd like to draw attention to leaving the `<div class="toolbar" data-hook="toolbar"> ` in.  I wasn't certain it would be safe to take this out since someone might be hooking into the toolbar data-hook and it makes no difference in how the page renders. Also I took out the conditional for the page title to render the `<h1>` element.  I made this choice since the simple concise flex box setup I went with relied on having a title on the page (or at least an empty h1). I believe there are no pages in the admin that don't have content for the page_title.  Further, an empty h1 element shouldn't hurt anyone either on the off chance they removed the text for the page_title.

# Before

before-header

(I skipped including a view of the screen scrolled down with no header in view. It seemed pointless)

# After

after-header

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
@@ -107,13 +107,16 @@ body {
 }
 
 #content-header {
-  padding: 15px 0;
+  padding: 15px 14px;
   background-color: very-light($color-3, 4);
   border-bottom: 1px solid $color-border;
   position: fixed;
   left: $width-sidebar;
   width: calc(100% - #{$width-sidebar});
   z-index: 1;
+  @include display(flex);
+  @include align-items(center);
+  @include justify-content(space-between);
 
   .page-title {
     font-size: 20px;
--- a/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
@@ -65,6 +65,10 @@ body {
   @include clearfix;
 }
 
+.content-under-header {
+  padding-top: 89px;
+}
+
 // Header
 //---------------------------------------------------
 .admin-nav-header {
@@ -106,6 +110,10 @@ body {
   padding: 15px 0;
   background-color: very-light($color-3, 4);
   border-bottom: 1px solid $color-border;
+  position: fixed;
+  left: $width-sidebar;
+  width: calc(100% - #{$width-sidebar});
+  z-index: 1;
 
   .page-title {
     font-size: 20px;
--- a/backend/app/views/spree/admin/shared/_content_header.html.erb
+++ b/backend/app/views/spree/admin/shared/_content_header.html.erb
@@ -1,23 +1,13 @@
 <% if content_for?(:page_title) || content_for?(:page_actions) %>
   <div id="content-header" data-hook>
-    <div class="container">
-      <div class="sixteen columns">
-        <div class="block-table">
-          <% if content_for?(:page_title) %>
-            <div class="table-cell">
-              <h1 class="page-title <%= yield :page_title_classes %>"><%= yield :page_title %></h1>
-            </div>
-          <% end %>
+    <h1 class="page-title <%= yield :page_title_classes %>"><%= yield :page_title %></h1>
 
-          <% if content_for?(:page_actions) %>
-            <div class="page-actions table-cell toolbar" data-hook="toolbar">
-              <ul class="inline-menu">
-                <%= yield :page_actions %>
-              </ul>
-            </div>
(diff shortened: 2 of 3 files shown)


tvdeyen:
I like the way the admin is heading to. Nice work. Nothing to complain (can you believe it)

:+1:

backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -65,6 +65,10 @@ body {
   @include clearfix;
 }
 
+.content-under-header {
+  padding-top: 89px;
```
graygilmore: What does this value represent? If it's the header's height we should save that as a value: `$height-header` and use it here and on the header.

backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -103,9 +107,16 @@ body {
...
+  padding: 15px 14px;
   background-color: very-light($color-3, 4);
   border-bottom: 1px solid $color-border;
+  position: fixed;
+  left: $width-sidebar;
+  width: calc(100% - #{$width-sidebar});
+  z-index: 1;
+  @include display(flex);
+  @include align-items(center);
+  @include justify-content(space-between);
```
graygilmore: Let's just reorganize this a little bit:

```scss
@include display(flex);
@include align-items(center);
@include justify-content(space-between);
position: fixed;
left: $width-sidebar;
width: calc(100% - #{$width-sidebar});
padding: 15px 14px;
border-bottom: 1px solid $color-border;
background-color: very-light($color-3, 4);
z-index: 1;
```

backend/app/views/spree/admin/shared/_content_header.html.erb
```diff
@@ -1,23 +1,13 @@
...
   <div id="content-header" data-hook>
-    <div class="container">
-      <div class="sixteen columns">
-        <div class="block-table">
-          <% if content_for?(:page_title) %>
-            <div class="table-cell">
-              <h1 class="page-title <%= yield :page_title_classes %>"><%= yield :page_title %></h1>
-            </div>
-          <% end %>
+    <h1 class="page-title <%= yield :page_title_classes %>"><%= yield :page_title %></h1>
```
graygilmore: Do you know what `page_title_classes` is used for? Not sure if we want to keep this around. It doesn't make a lot of sense to me.

backend/app/views/spree/admin/shared/_content_header.html.erb
```diff
@@ -1,23 +1,13 @@
...
-          <% if content_for?(:page_actions) %>
-            <div class="page-actions table-cell toolbar" data-hook="toolbar">
-              <ul class="inline-menu">
-                <%= yield :page_actions %>
-              </ul>
-            </div>
-          <% end %>
-        </div>
+    <% if content_for?(:page_actions) %>
+      <div class="toolbar" data-hook="toolbar">
```
graygilmore: @jhawthorn is this a kind of class name that we need to keep around? I'd prefer to remove anything we're not actually using.

backend/app/views/spree/layouts/admin.html.erb
```diff
@@ -34,7 +34,7 @@
 
       <%= render :partial => 'spree/admin/shared/content_header' %>
 
-      <div class="container">
+      <div class="container<%= ' content-under-header' if (content_for?(:page_title) || content_for?(:page_actions)) %>">
```
graygilmore: As we discussed IRL, move this to `.main-content` and change the conditional class name to `.has-header`. This is a lot clearer on why the `padding-top` is being added.

backend/app/views/spree/admin/shared/_content_header.html.erb
graygilmore: Instead of having an empty `<h1>` we can tweak how we're using Flexbox.

```diff
#content-header {
  @include display(flex);
  @include align-items(center);
- @include justify-content(space-between);
}

+.page-actions {
+  @include flex-grow(1);
+  text-align: right;
+}
```

Murph33: This doesn't work quite like this since `.page-actions` is applied to the `ul` which is inside of that `div` with the `toolbar` class and datahook.  Works when applied to the toolbar div obviously but are we keeping that?  I would like to eliminate the empty h1 even though it shouldn't hurt, empty elements used to influence aren't great.

Murph33:
After talking about it we're not sure we want to push in a sticky header quite yet but are happy with the other changes so we will leave this for debate and ponderings in #630 and make a new PR for the skeleton work.

PR #960, opened 2016-03-04
Use conditional relationship rather than manual accessor.

eric1234:
This is basically just a change in implementation and no change in
functionality as it still does the same thing. `Spree::User#default_address`
still returns the related default address.

Why I prefer this implementation is it allows the address to be
preloaded preventing O(N) queries if we need to print some address
info out about each user (i.e. name, phone, etc). Printing the
name for each user seems like a common operation so it seems it
would be good to provide a way to do that for a list of users
efficiently.

The `default_address=` implementation is remaining and overriding the
implmenetation the relationship would provide. This will allow the
special handling of address assignment to remain.

### DIFF
--- a/core/app/models/concerns/spree/user_address_book.rb
+++ b/core/app/models/concerns/spree/user_address_book.rb
@@ -24,6 +24,10 @@ module Spree
 
       # bill_address is only minimally used now, but we can't get rid of it without a major version release
       belongs_to :bill_address, class_name: 'Spree::Address'
+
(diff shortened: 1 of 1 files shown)


jhawthorn:
Seems reasonable to me. Thank you. :+1:

cbrunsdon:
Yea, looks good to me too, thanks @eric1234

cbrunsdon:
@eric1234 just needs a rebase, when you get a sec

eric1234:
Looks like f4834987 caused the conflict. Resolved so once the tests finish we should be good to go.

PR #962, opened 2016-03-04
Hide niche stock restriction feature for cleaner backed UI.

eric1234:
The backend interface has in the user profile a "Stock Locations"
configuration. It's not clear from the UI what this feature is at all
(my first assumption was it was some indicator of where a customer should
get products shipped from by default).

Looked into it and it seems to be a order fullfillment feature to
ensure folks that are fullfilling the order can only do so out of the
stock locations they are assigned (so a worker in Seattle cannnot
fullfill with inventory from Atlanta).

This seems very much a niche feature as most stores only have one location
and even stores that do have multiple locations may not have problems
fulfilling from the wrong location. I would argue this entire feature should
be an extension.

But going down that path of moduralizing that feature is probably not worth
the effort at this point so just extended it to be less prominent. Added
a new preference that will indicate if this configuration should be shown.
Since I hate to have yet another configuration I made the default intelligent
so that it looks to see if one of the RestrictedStock permission sets are
being used to determine the default.

If you are using a custom permission set and therefore the intelligent
default fails you can still manually indicate you are using this feature but
hopefully for stores using it the feature will just work as before and for
stores not using it, they get a cleaner UI.

### DIFF
--- a/backend/app/views/spree/admin/users/_form.html.erb
+++ b/backend/app/views/spree/admin/users/_form.html.erb
@@ -31,7 +31,7 @@
       </div>
     <% end %>
 
-    <% if can?(:display, Spree::StockLocation) %>
+    <% if Spree::Config.can_restrict_stock_management && can?(:display, Spree::StockLocation) %>
       <div data-hook="admin_user_form_stock_locations" class="field">
         <%= label_tag nil, Spree.t(:stock_locations) %>
         <ul>
--- a/backend/app/views/spree/admin/users/_form.html.erb
+++ b/backend/app/views/spree/admin/users/_form.html.erb
@@ -31,7 +31,7 @@
       </div>
     <% end %>
 
-    <% if Spree::Config.can_restrict_stock_management && can?(:display, Spree::StockLocation) %>
+    <% if Spree::Config[:can_restrict_stock_management] && can?(:display, Spree::StockLocation) %>
       <div data-hook="admin_user_form_stock_locations" class="field">
         <%= label_tag nil, Spree.t(:stock_locations) %>
         <ul>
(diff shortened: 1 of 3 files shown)


cbrunsdon:
Rather than the code, I'd prefer we do this as a preference with a default value of off.

I certainly appreciate the attempt (and the spec), but default off with a note in changelog to re-enable seems reasonable to me.

eric1234:
It is a preference. The only bit of trickery here is that the default is halfway intelligent. Rather than break compatibility for folks that are using the permissions sets (I assume someone is since the feature was created) it detects the usage and sets the default according to that detection.

cbrunsdon:
@eric1234 aye, I appreciate the attempt at backwards compat. We ended up in this feature in core from before solidus dev was done publicly (it was included in a giant rebase to get a few branches together 18 months ago) and I know its not widely used enough to maintain backwards compat.

jhawthorn:
I'm wary of preferences being this smart. I also would prefer `can_restrict_stock_management` being false by default.

I expect this feature to be used by almsot nobody, in the future I'd like to consider moving it into a separate extension

jhawthorn:
:+1: Thank you

eric1234:
All cleaned up. I'll leave it up to you to decide how you want to version. Strictly speaking you are breaking compatibility so if you are a semantic versioning zealot you bump to version 2 but seeing as it impacts so few sites (problably one) the pragmatic approach would be just to note it in the release notes.

cbrunsdon:
I'm :+1: on this too.

Our long term goal is to be properly semver'd, but aye with the code base we have it'll take a while before that is achievable. For now I think the changelog entry is entirely reasonable.

Thanks a lot @eric1234

jhawthorn:
I aim to follow semver as closely as possible, but I feel this is fine to squeak by in a minor release as this only affects the admin interface and only for an infrequently used feature.

Thanks @eric1234

PR #640, opened 2015-12-31
Added prefix Spree:: to OrderMutex

sanchojaf:

PR #964, opened 2016-03-04
Allow the user sidebar partial to be used outside the Solidus engine

eric1234:
This should be a no-op for a stock solidus store. Where it comes in
handy is when you are trying to keep a separation of concerns between
custom code vs. Spree code while still maintaining a consistent UI for
the user.

My custom code operates in it's own context (the app or independent engines)
but I have data hanging off of the user model (i.e. info related to the user).
I don't want the user to have multiple admin screens so my custom code has
a light bit of integration to make it fit in with the Spree admin interface
(uses the same layout and uses shared partials such as this sidebar).

This setup generally works as many of the links already explicitly reference
the `spree` engine. For example the main menu does this at:

https://github.com/solidusio/solidus/blob/master/backend/app/helpers/spree/admin/navigation_helper.rb#L17

This commit updates this sidebar partial to follow the pattern set by the
main menu and explicitly indicate the engine. This allows it to work within the
spree engine as well as outside the spree engine.

### DIFF
--- a/backend/app/views/spree/admin/users/_sidebar.html.erb
+++ b/backend/app/views/spree/admin/users/_sidebar.html.erb
@@ -6,26 +6,26 @@
   <nav class="menu">
     <ul data-hook="admin_user_tab_options">
       <li<%== ' class="active"' if current == :account %>>
-        <%= link_to_with_icon 'user', Spree.t(:"admin.user.account"), edit_admin_user_path(@user) %>
+        <%= link_to_with_icon 'user', Spree.t(:"admin.user.account"), spree.edit_admin_user_path(@user) %>
       </li>
       <% if can?(:addresses, @user) %>
         <li<%== ' class="active"' if current == :address %>>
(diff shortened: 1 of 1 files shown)


cbrunsdon:
Personally I'm :+1: this, especially for a partial so likely to be overwritten for a store.

But I'm also not sure of our stance on these types of changes (see: #640, though this is definitely a different situation). Do we take them on an as-sent basis? Do we namespace all routes in our views in the admin going forward? Etc.

eric1234:
I def understand about not wanting to prefix everywhere. Too much boilerplate. But given the pluggable nature of users combined with this being a highly used extension point (most stores will have more extensive custom user data) I think like the main menu, the benefits outweighs the annoying boilerplate.

mamhoff:
I am definitely :+1: on prefixing path helpers, as it makes interaction with host apps and other engines a lot less error-prone.

tvdeyen:
:+1: we should always ever use the `spree` route proxy in all views and helpers, so they can be used outside of spree controllers.

jhawthorn:
:+1: thank you

PR #965, opened 2016-03-04
Connect shipping rates to shipments earlier in Stock::Coordinator

jordan-brough:
This will give the estimator the information it needs to calculate
taxes correctly in the pending tax work.

See comments in https://github.com/solidusio/solidus/pull/904#discussion-diff-54887938

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 ## Solidus 1.3.0 (unreleased)
 
+*   Changes to Spree::Stock::Estimator
+
+    * The package passed to Spree::Stock::Estimator#shipping_rates must have its
+      shipment assigned and that shipment must have its order assigned. This
+      is needed for some upcoming tax work in to calculate taxes correctly.
+    * Spree::Stock::Estimator.new no longer accepts an order argument. The order
+      will be fetched from the shipment.
+
+    https://github.com/solidusio/solidus/pull/965
+
 *   Made Spree::Order validate :store_id
 
     All orders created since Spree v2.4 should have a store assigned. We want to build more
--- a/api/spec/controllers/spree/api/checkouts_controller_spec.rb
+++ b/api/spec/controllers/spree/api/checkouts_controller_spec.rb
@@ -136,7 +136,7 @@ module Spree
 
       it "can update shipping method and transition from delivery to payment" do
         order.update_column(:state, "delivery")
-        shipment = create(:shipment, order: order)
+        shipment = create(:shipment, order: order, address: order.ship_address)
         shipment.refresh_rates
         shipping_rate = shipment.shipping_rates.where(selected: false).first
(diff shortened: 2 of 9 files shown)


mamhoff:
:heart: Can you look at the last failing test, and then I'll rebase #904.

jhawthorn:
> This will give the estimator the information it needs to calculate taxes correctly in the pending tax work.

If this is the case we need a spec for that

jordan-brough:
@mamhoff & @jhawthorn updated. Look good?  I don't love the spec I added but lmk what you think. Any ideas are definitely welcome.

mamhoff:
I tried rebasing this onto #904. My work then breaks a gazillion other things, and I think the ugly spec and this failure are related somehow: https://circleci.com/gh/solidusio/solidus/2473. I'm not so sure how to remedy this, but essentially our spec setup/our factories are set up in such a way that with this change AND my changes in, 700 specs break because apparently lots of shipping rates have no shipment when they're created. :crying_cat_face:

jordan-brough:
@mamhoff it's [this commit](https://github.com/magiclabs/solidus/commit/e0e9b58e43fc1281610b4da5fb7fccb7fb218637) where things fall apart right? I'm happy to try to help take a look when I have a few minutes.  I think it'd be very good to avoid the `attr_accessor :order` on the shipping rate if possible in PR 904 and I think it'd be good to understand what's breaking here.

Re the spec I added in 86e4dd8:  I think it's a not-great spec because really it's testing internal implementation rather than the public API of the class.  There's probably a better way to approach a spec but I'm not sure if there's a pragmatic one.  I'll think about that some more when I have a minute (open to ideas!).

jordan-brough:
FYI I chatted briefly with @mamhoff on this and I'm going to take some time tomorrow morning to code up a fix and improve the specs.

jordan-brough:
@mamhoff @jhawthorn I've added a couple of commits which broaden the scope of this PR a bit but I think it's worth it.  Thoughts?

These changes fix the errors that @mamhoff's [commit](https://github.com/magiclabs/solidus/commit/e0e9b58e43fc1281610b4da5fb7fccb7fb218637) was seeing.

There is some change to the public API so I've added a changelog for that.

@mamhoff you'll want to omit [this change](https://github.com/magiclabs/solidus/commit/e0e9b58e43fc1281610b4da5fb7fccb7fb218637#diff-503e05312113225cc0cdc3b8de8515c2) to `core/spec/models/spree/stock/estimator_spec.rb` if you rebase on top of this.

jhawthorn:
:+1:

mamhoff:
I said so before in real life, but this is a change that will make my life a lot easier. Thank you @jordan-brough!

gmacdougall:
Great work @jordan-brough 

:+1:

jordan-brough:
Thanks everyone!

PR #966, opened 2016-03-04
Move header title and actions out of 16 column skeleton

Murph33:
We are moving away from the skeleton and would like to move the page title and page actions out to the edges.  This is a stripped down version of #959 implementing @graygilmore's suggestions since we weren't sure about making the header sticky at crunch time (opinions at #630 please).  

# Before

before-header

# After

after-split

### DIFF
--- a/backend/app/views/spree/admin/shared/_content_header.html.erb
+++ b/backend/app/views/spree/admin/shared/_content_header.html.erb
@@ -1,23 +1,15 @@
 <% if content_for?(:page_title) || content_for?(:page_actions) %>
   <div id="content-header" data-hook>
-    <div class="container">
-      <div class="sixteen columns">
-        <div class="block-table">
-          <% if content_for?(:page_title) %>
-            <div class="table-cell">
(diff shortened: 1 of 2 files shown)


backend/app/views/spree/admin/shared/_content_header.html.erb
```diff
@@ -1,23 +1,13 @@
...
-              <ul class="inline-menu">
-                <%= yield :page_actions %>
-              </ul>
-            </div>
-          <% end %>
-        </div>
-      </div>
-    </div>
+    <% if content_for?(:page_actions) %>
+      <ul class="header-actions page-actions inline-menu">
```
graygilmore: I have a feeling like we might want to keep the `data-hook="toolbar"` here. @jhawthorn what are your thoughts on maintaining that kind of thing?

jhawthorn: I agree. We gain nothing by removing it here.

jhawthorn:
Merged in #1012

PR #967, opened 2016-03-06
Refactor Order factory

mamhoff:
This adds the possibility of running `build` on an order with a full
object graph that's mostly not persisted.

Mostly meaning that while associated models might still persist
associations, many many associations on the new order object will
not be persisted.

There is a minor fix in the reimbursement spec that tells actually
associates a refund to the reimbursement it checks. I do not fully grasp
why that ever worked.

Another minor fix is that the reimbursement total now does not expect
shipping to be refunded. I believe that is the standard behaviour, and
the reason it didn't work before is because of incomplete spec setup.


mamhoff:
This is a long project, I won't pursue it in the near future.

PR #968, opened 2016-03-07
WIP Add vat_country relation to prices

mamhoff:
I chose to use the ISO code as primary/foreign key as that is just so
much easier to understand than the ever-changing country IDs.

I believe with the 2-character limit on the column and Indexes that
should not be a major performance issue, and it makes our database dumps
and SQL queries a lot easier to read.

I don't want to touch the default price and would like that to always be
the net (export) price, as that makes prices easy to compare.

### DIFF
--- a/backend/config/routes.rb
+++ b/backend/config/routes.rb
@@ -50,6 +50,7 @@ Spree::Core::Engine.add_routes do
         end
       end
       resources :variants_including_master, only: [:update]
+      resources :prices, only: [:index, :edit, :destroy]
     end
     get '/products/:product_slug/stock', to: "stock_items#index", as: :product_stock
 
--- a/backend/spec/features/admin/products/prices_spec.rb
+++ b/backend/spec/features/admin/products/prices_spec.rb
@@ -1,0 +1,19 @@
+# encoding: utf-8
+require 'spec_helper'
+
+describe "Prices", type: :feature do
+  stub_authorization!
+
+  let(:product) { create(:product_with_option_types, price: "1.99", cost_price: "1.00", weight: "2.5", height: "3.0", width: "1.0", depth: "1.5") }
+  context "listing prices" do
+    let!(:variant) do
+      create(:variant, product: product, price: 19.99)
+    end
+
+    it "opens the prices tab in the product view" do
+      visit spree.edit_admin_product_path(id: product.slug)
+      click_link("Prices")
+      expect(page).to have_content("19.99")
+    end
+  end
+end
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -357,6 +357,9 @@ en:
       spree/payment_method:
         one: Payment Method
         other: Payment Methods
+      spree/price:
+        one: Price
+        other: Prices
       spree/product:
         one: Product
         other: Products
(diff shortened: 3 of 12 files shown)


core/db/migrate/20160303150239_add_vat_country_iso_to_price.rb
```diff
@@ -0,0 +1,7 @@
+class AddVatCountryIsoToPrice < ActiveRecord::Migration
+  def change
+    add_column :spree_prices, :vat_country_iso, :string, limit: 2, null: false, default: ''
+    add_index :spree_prices, :vat_country_iso
```
gmacdougall: We should probably have a combined unique index of some sort to ensure that there are not more than one price for the same set of things that should uniquely identify a price.

backend/config/routes.rb
```diff
@@ -50,6 +50,7 @@
         end
       end
       resources :variants_including_master, only: [:update]
+      resources :prices, only: [:index, :edit, :destroy]
```
gmacdougall: Do we have the forms for edit?

core/spec/models/spree/price_spec.rb
```diff
@@ -38,5 +50,37 @@
...
       it { is_expected.to be_valid }
     end
+
+    context "with default price set" do
+      subject { Spree::Price.new(variant: variant, is_default: true, vat_country: vat_country) }
+
+      context "when there is a vat country" do
+        let(:vat_country) { create(:country) }
+
+        it { is_expected.not_to be_valid }
```
gmacdougall: I would find this a bit easier to read if it were:

```
it { is_expected.to be_invalid }
```

core/config/locales/en.yml
```diff
@@ -1386,6 +1389,10 @@ en:
     presentation: Presentation
     previous: Previous
     price: Price
+    prices:
+      index:
+        title: "Prices for %{product_name}"
+        net: Nettopreis
```
gmacdougall: This is not English.

backend/spec/features/admin/products/prices_spec.rb
```diff
@@ -0,0 +1,19 @@
+# encoding: utf-8
+require 'spec_helper'
+
+describe "Prices", type: :feature do
+  stub_authorization!
+
+  let(:product) { create(:product_with_option_types, price: "1.99", cost_price: "1.00", weight: "2.5", height: "3.0", width: "1.0", depth: "1.5") }
```
gmacdougall: Could you please add some new lines in here to make it a bit easier to read, and shorten the line length?

Example:
```Ruby
let(:product) do
  create(
    :product_with_option_types,
    price: "1.99",
    cost_price: "1.00",
    weight: "2.5",
    height: "3.0",
    width: "1.0",
    depth: "1.5"
  )
end
```

I would also like to see some of those removed if possible, and only supply the data which is important to make this spec pass.

backend/config/routes.rb
mamhoff: Nope, I just needed the path helpers for now.

core/db/migrate/20160303150239_add_vat_country_iso_to_price.rb
mamhoff: I think there's a need to have more than one price for the same things present, probably something like time-based pricing (can we make this new price the default from Wednesday next week?) 

I'd rather propose something like superseding, editable `valid_from` dates. That way, you always take the newest past valid from date, and you get a whole lot of flexibility at the cost of only one datetime field. What do you think?

mamhoff:
I'm going to do this a little differently, in three steps: 

1. Add a user interface for prices in different currencies. 
2. Get #987 in
3. Then add the additional field.

Closing, for the time being.

PR #971, opened 2016-03-07
Remove "or" wording from admin backend

Murph33:
In many places the admin control gives us an option to `update` or `cancel`.  The or is unnecessary, ugly, and implied.  It is better without it.  I don't believe @Mandily has this in an issue but it is at her request so UX approved.

# before

before or

# after

after or

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
@@ -272,21 +272,6 @@ fieldset {
     > * + * {
       margin-left: 2em;
     }
-
-    span.or {
-      background-color: $color-1;
-      border-width: 5px;
-      margin-left: 5px;
-      margin-right: 11px;
-      position: relative;
-      z-index: 1;
-
-      -webkit-box-shadow: 0 0 0 5px $color-1;
-         -moz-box-shadow: 0 0 0 5px $color-1;
-          -ms-box-shadow: 0 0 0 5px $color-1;
-           -o-box-shadow: 0 0 0 5px $color-1;
-              box-shadow: 0 0 0 5px $color-1;
-    }
   }
 
   &.labels-inline {
--- a/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
@@ -268,6 +268,11 @@ fieldset {
       }
     }
 
+    // Always make sure the choices at the bottom of a fieldset are spaced out
+    > * + * {
+      margin-left: 2em;
+    }
+
     span.or {
       background-color: $color-1;
       border-width: 5px;
--- a/backend/app/views/spree/admin/adjustments/edit.html.erb
+++ b/backend/app/views/spree/admin/adjustments/edit.html.erb
@@ -15,7 +15,6 @@
 
     <div class="filter-actions actions" data-hook="buttons">
       <%= button Spree.t(:continue), 'arrow-right' %>
-      <span class="or"><%= Spree.t(:or) %></span>
       <%= link_to_with_icon 'remove', Spree.t('actions.cancel'), admin_order_adjustments_url(@order), :class => 'button' %>
     </div>
   </fieldset>
--- a/backend/app/views/spree/admin/adjustments/new.html.erb
+++ b/backend/app/views/spree/admin/adjustments/new.html.erb
@@ -16,7 +16,6 @@
 
     <div class="filter-actions actions" data-hook="buttons">
       <%= button Spree.t(:continue), 'arrow-right' %>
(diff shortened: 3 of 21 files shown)


backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -129,3 +129,11 @@ body {
   border-top: 1px solid $color-border;
   padding: 10px 0;
 }
+
+// Button Spacing
```
graygilmore: If we need comments like these we know we're putting this in the wrong place. I would prefer to see these styles in a file like `_forms.scss`.

What these styles do need, though, is a comment explaining their purpose.

Murph33: Good points, thanks once again Gray.  Slotting it into the `_forms.scss` also reminded me to remove the now unused `span.or` and also let me nest the styles into the same position in the stylesheet that `span.or` had been.

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -267,19 +267,14 @@ fieldset {
...
-      margin-right: 11px;
-      position: relative;
-      z-index: 1;
-
-      -webkit-box-shadow: 0 0 0 5px $color-1;
-         -moz-box-shadow: 0 0 0 5px $color-1;
-          -ms-box-shadow: 0 0 0 5px $color-1;
-           -o-box-shadow: 0 0 0 5px $color-1;
-              box-shadow: 0 0 0 5px $color-1;
+    // Previously there had been a span of class 'or' used for spacing.
```
graygilmore: So close! This makes a good commit message but it's entirely what's needed here. The goal here is to explain what this code is doing so that the next person looking at this doesn't have to scratch their head for 10min. Something like:

```scss
// Always make sure buttons are spaced out. Sometimes there are forms
// with single buttons inside so we can't always just do `.button + .button`.
```

jhawthorn:
:+1:

Mandily:
Spacing looks good to me. Thanks @Murph33 for getting this in.

tvdeyen:
:+1:

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -267,19 +267,11 @@ fieldset {
...
-
-      -webkit-box-shadow: 0 0 0 5px $color-1;
-         -moz-box-shadow: 0 0 0 5px $color-1;
-          -ms-box-shadow: 0 0 0 5px $color-1;
-           -o-box-shadow: 0 0 0 5px $color-1;
-              box-shadow: 0 0 0 5px $color-1;
+    // Always make sure the choices at the bottom of a fieldset are spaced out
+    form + form,
+    button + a,
+    form + a {
```
Sinetheta: It seems like you've got quite a few possible children here, but aren't addressing many of the permutations. Is that because these are the only ones that occur? Would `> * + *` be what you're looking for? http://alistapart.com/article/axiomatic-css-and-lobotomized-owls

Murph33: Those were the only locations that a `<span class="or">` showed up.  I can't off the top of my head confirm that the elements with `.filter-actions` only ever had two two elements in it so I'm not certain that a `* + *` selector would be safe.

Murph33: I went through and spot checked.  Those rules cover all existing cases and `* + *` would add no new ones.  The other occasions that did not include the `span class="or"` element are just for spots with individual buttons or links such as the search forms.

Sinetheta: That was my thought @Murph33. If the goal is to "inset space between any children" then I think `*` selectors might be what you're looking for (with child `>` selectors if necessary).

For example http://codepen.io/Sinetheta/pen/KzNpXN

mamhoff:
:+1: 

@Sinetheta @Murph33 frankly I don't quite understand the selector magic you're debating. Can you say whether this still needs work, or can it be merged?

Sinetheta:
:+1: from me, gj @Murph33

Murph33:
I am ambivalent on changing the selectors to the catchall or not.  It would really just come down to depending on the behaviour an admin would want if they decided to extend that part of the app.  I don't know how likely that is or not.  I was also just waiting to see if someone else had an opinion to share on it or if @Sinetheta wanted to lay down the :hammer:.

graygilmore:
I think it should probably be the `> * + *` but hopefully this area gets a reboot anyways.

mamhoff:
I think no-one's really said anything here, and apparently all possible options work just fine. I'd say we look at merging before next week's core meeting if no-one complains. OK?

jhawthorn:
Thank you

PR #972, opened 2016-03-07
Replace powerTip with bootstrap tooltips

jhawthorn:
Depends on #955 

This swaps out the existing jQuery PowerTip tooltip library with bootstrap's tooltips.

One advantage of this is that bootstrap's tooltips can listen on the body for hover events on elements, which avoid the need to re-initialize tooltips when content is dynamically added to the page.


tvdeyen:
#955 has landed. Now this can be rebased @jhawthorn

jhawthorn:
@tvdeyen done!

cbrunsdon:
Looks good to me, thanks @jhawthorn

tvdeyen:
If we change the selector to something more specific and closer to the event target element, we are good to go!

Nice work, John, as always 👌🏻

PR #279, opened 2015-08-12
Remove state and mandatory columns from adjustments

jhawthorn:
This comes from a conversation with @jordan-brough a while ago.

The state machine on adjustment serves no purpose, as it allows changing between both states without restriction. `open` and `close` are also states which need an explanation, vs `finalized`, which we've used elsewhere in the codebase with the same meaning. `open` also isn't ideal as it overrides `Kernel#open`, which causes state machine warnings unless configured to override that method.

This also removes the `mandatory` column, which has been unused since the Spree 2.2 adjustment changes. I'm not sure how this fits in with our versioning requirements, this column has been "gone" forever in that it hasn't done anything, but I could see an old store still setting it (or setting it to false) which would now error instead of just doing nothing.

PR #973, opened 2016-03-08
Update spree_adjustments.finalized to be non null

jordan-brough:
When we migrated from state=open/closed to finalized=true/false in
https://github.com/solidusio/solidus/pull/279 we added `finalized` as a nullable column.  This breaks the
scopes that we added in that commit -- when finalized is null the
adjustments are neither `finalized` nor `not_finalized`.

I noticed this while trying to add some code and specs using the
`not_finalized` scope.

We could try to tweak the scopes but I think it'd be better to just
avoid the ternary logic altogether by disallowing nulls.

In the old code we had `state_machine :state, initial: :open` which
prevented this problem even though the `state` column itself was
nullable.


jhawthorn:
Thanks jordan, that makes sense :+1:

gmacdougall:
Death to tri-state booleans!  :+1:

PR #974, opened 2016-03-08
Rails 5 preparations

jhawthorn:
These are some backports from the rails 5 branch. I believe they improve code quality so we might as well get them in and reduce the number of changes we'll eventually need.

* Use `AR::Relation#distinct` instead of `#uniq`, which is deprecated in rails 5
* Use `association.reload` instead of `association(true)`, which is deprecated in rails 5
* Remove once instance of `merge(nil)`
* Some fixes related to `ActionController::Parameters`, which are no longer a subclass of `Hash` in rails 5


gmacdougall:
Looks good. There's some scary stuff in there. I'm continually amazed that half of this stuff works at all.

:+1:

cbrunsdon:
All looks good to me, thanks hawth :+1:

PR #975, opened 2016-03-08
Replace settings configuration menu with separate tabs partials

Murph33:
This was done to implement issue #634.  It uses the tab functionality introduced in #870 by @graygilmore.   The only changes from the plans outlined in #634 (as given by @Mandily) were to change refunds and reimbursements to checkout and to add adjustment reasons to that section.

# Before

before tabs

# After

after tabs

## Change log

- Remove configuration menu (right nav) from settings pages
- Group settings pages into related partials as outlined in #634 
- Partials are rendered on pages owned by the partials as tabs as a top bar
- Admin-nav has a sub-menu for the settings now


gmacdougall:
Can we drop `configuration_menu.html.erb` now? Also I think that a Changelog entry for this would be appropriate.

Good work on this. I really like the usability improvements here.

Murph33:
Yes we can.  Should have removed it already come to think of it.  I will get some kind of a changelog thrown together as well.  Thanks.

backend/app/views/spree/admin/shared/_settings_sub_menu.html.erb
```diff
@@ -0,0 +1,27 @@
+<ul class="admin-subnav" data-hook="admin_settings_sub_tabs">
+
```
graygilmore: #nitpick can you remove this empty line?

graygilmore:
Good stuff. Happy to see the Tabs being used :smile:

cbrunsdon:
All looks reasonable to me, thanks @Murph33 :+1:

tvdeyen:
Could we rename the `checkout_tabs` partial into something less generic and more specific to the settings? Maybe `settings_checkout_tabs`?

Murph33:
I just wanted to draw attention to how I implemented https://github.com/Murph33/solidus/blob/change_settings_to_tabs/backend/app/views/spree/admin/shared/_settings_sub_menu.html.erb : My one concern is that currently it gives a link to the first thing in each of the conditional lists if any of the conditionals pass.  I don't know how likely it is for someone to be able to display `states` but not `zones` and I thought this would be fairly safe.  I was wondering if someone has a way that isn't too messy to ensure it links to something the user can display for certain without a messy conditional.

cbrunsdon:
@Murph33 I'm okay with that for now, we can address that as a separate issue if and when we need to.

I'm still :+1:  on this.

mamhoff:
:+1: I think this can go in. Virtually the entire core team have given their thumbs up, there's a changelog entry, and it looks great. What's stopping us?

PR #977, opened 2016-03-09
Remove normalize from backend.css

jhawthorn:
We now have normalize included as part of bootstrap


cbrunsdon:
Looks reasonable to me :+1:

gmacdougall:
:+1:

PR #980, opened 2016-03-11
Move the strange refunding logic from Spree::TaxRate into the default tax calculator

mamhoff:
When reviewing #904, we realized that we can not change the logic of
Spree::TaxRate#compute_amount if we want the migration included there to
always to the right thing. Because that method modifies the tax amount of things,
it should actually be in a calculator. I now moved that logic to the default tax
calculator, with the intention of deprecating use of that calculator entirely.

There is code for pre-Spree-2.2 order-based taxation in there, as well as code for
generating tax refunds, which in the future we won't need either.

### DIFF
--- a/core/spec/models/spree/calculator/default_tax_spec.rb
+++ b/core/spec/models/spree/calculator/default_tax_spec.rb
@@ -7,24 +7,32 @@ describe Spree::Calculator::DefaultTax, type: :model do
   let!(:rate) { create(:tax_rate, tax_category: tax_category, amount: 0.05, included_in_price: included_in_price, zone: zone) }
   let(:included_in_price) { false }
   let!(:calculator) { Spree::Calculator::DefaultTax.new(calculable: rate ) }
-  let!(:order) { create(:order, ship_address: address) }
-  let!(:line_item) { create(:line_item, price: 10, quantity: 3, tax_category: tax_category) }
-  let!(:shipment) { create(:shipment, cost: 15) }
 
   context "#compute" do
     context "when given an order" do
-      let!(:line_item_1) { line_item }
-      let!(:line_item_2) { create(:line_item, price: 10, quantity: 3, tax_category: tax_category) }
+      let(:order) do
+        create(
+          :order_with_line_items,
+          line_items_attributes: [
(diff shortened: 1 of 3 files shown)


gmacdougall:
I'd like to see the unit tests for default_tax.rb improved, covering the new functionality and providing better tests for the existing logic.

We also need to make the default_tax handle currencies that do not have two decimal places, but I'm fine with leaving that for another PR.

mamhoff:
Currencies with more than two decimal places are not supported at all in Solidus, I think that's an entirely different issue. And yeah, I'll provide unit tests for default tax.

gmacdougall:
I was thinking more currencies without decimal places like JPY or KRW which are generally supported, aside from some bad rounding logic here and there.

mamhoff:
Ha. Yeah, I hear you.

mamhoff:
Refactored the specs and added specs for some of the refunding logic @gmacdougall

mamhoff:
The failing spec on 6990e65 is due to a PhantomJS crash, I'm quite sure.

core/spec/models/spree/calculator/default_tax_spec.rb
```diff
@@ -70,51 +68,145 @@
...
+      expect(subject.method(:compute_shipping_rate)).to \
+        eq(subject.method(:compute_item))
+    end
+  end
+
+  context 'when given a line item' do
+    let(:line_item) { Spree::LineItem.new }
+
+    it 'calls #compute_line_item' do
+      expect(subject).to receive(:compute_line_item)
```
gmacdougall: I don't like stubbing methods on the file that we're writing the spec for. I don't think that this test provides value.

mamhoff:
Rebased and added specs for the actual behaviour of the calculator when called with line item, shipment or shipping rate. I think this is good to go now. @gmacdougall @jhawthorn

adammathys:
:+1: Looks good to me.

mamhoff:
@gmacdougall If you have a minute, this should be good to go.

mamhoff:
Please please review this, it's base work for #904, which is also ripe.

jhawthorn:
:+1: Looks good to me

gmacdougall:
:+1:

PR #981, opened 2016-03-11
Simplify Default Tax Calculator

mamhoff:
The logic for taxing line items, shipments and shipping rates is
already exactly the same. There's no need to maintain two methods
when one is enough.


gmacdougall:
:+1:

Good catch.

tvdeyen:
:+1:

jordan-brough:
:+1:

PR #982, opened 2016-03-11
Simplify payment row amount edit JS

jhawthorn:
Refactor to the inline edit feature on the payments page.

This started as a want to remove the JS being tied to the icon of the action buttons (`@$actions().find(".fa-#{@action}, .fa-cancel")`), but the existing JS was very opaque and using 145 lines for a simple inline edit seemed excessive.

This removes an automatic save whenever the amount input was unfocused, which I believe is bad UI. It's not clear that the user intends to save, they may intend to cancel, or even copy an amount from a previous payment. I ran this by @Mandily who agreed.

This replaces 145 lines of coffeescript soup with 34 well structured lines.
* All HTML is now declared in the view
* The edit button is visible on page load (previously it only showed up
  after an ajax request). This might solve some spurious CI failures on
  this page.
* No ajax requests on page load (previously one per payment).
* JS doesn't tie behaviour to the icon shown on the button.

Thanks to @Sinetheta who's recommendation of the `editing` class set most of this in motion.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/payments/edit.js.coffee
+++ b/backend/app/assets/javascripts/spree/backend/payments/edit.js.coffee
@@ -125,10 +125,6 @@ jQuery ($) ->
       $('<input />')
         .prop(id: 'amount', value: amount)
         .width(width)
-        .one
-          blur: =>
-            clicked = (@$buttons().filter -> $(@).data('clicked')).length
-            @save() unless clicked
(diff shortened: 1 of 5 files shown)


tvdeyen:
Wow. :+1:

cbrunsdon:
Aye, thanks @jhawthorn and co, looks good to me too.

PR #983, opened 2016-03-14
Better selected page detection for admin menu.

eric1234:
Previously if you used a custom URL you had to do something like this:

    <%= tab :leads, match_path: 'batch_communications/leads',
      url: main_app.admin_batch_communications_leads_path if can? :admin, Lead %>

Basically we are specifying the path twice. Once to indicate where we should link to and once to indicate if the we are at that path. Generally if we link to the path then we should also match that path. This patch allows the above to work without the `match_path` option.

If you specify a `match_path` that will take priority. This just provides a nice default to reduce the amount of boilerplate code.

### DIFF
--- a/backend/app/helpers/spree/admin/navigation_helper.rb
+++ b/backend/app/helpers/spree/admin/navigation_helper.rb
@@ -31,7 +31,8 @@ module Spree
         elsif options[:match_path]
           request.fullpath.starts_with?("#{admin_path}#{options[:match_path]}")
         else
-          args.include?(controller.controller_name.to_sym)
+          request.fullpath.starts_with?(destination_url) ||
+            args.include?(controller.controller_name.to_sym)
         end
         css_classes << 'selected' if selected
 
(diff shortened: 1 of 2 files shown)


backend/app/helpers/spree/admin/navigation_helper.rb
```diff
@@ -31,6 +31,7 @@ def tab(*args, &_block)
         elsif options[:match_path]
           request.fullpath.starts_with?("#{admin_path}#{options[:match_path]}")
         else
+          request.fullpath.starts_with?(destination_url) ||
           args.include?(controller.controller_name.to_sym)
```
jhawthorn: Could we have an indent here to show that this is part of the previous statement?

jhawthorn:
Fine by me.

eric1234:
Added the indent to conform to the project style. Once tests complete this should be good to merge.

mamhoff:
:+1: Looks good to me, too.

tvdeyen:
:+1: Nice

PR #984, opened 2016-03-14
Admin menu configuration partial to be usable outside of Solidus engine

eric1234:
A continuation of the work done on PR #964. Refer there for rational and discussion.
I also see application configuration as a common extension point and requiring everything
to be in the `spree` engine defeats the purpose of namespaces so making it usable
to other engines.

### DIFF
--- a/backend/app/views/spree/admin/shared/_areas_tabs.html.erb
+++ b/backend/app/views/spree/admin/shared/_areas_tabs.html.erb
@@ -2,16 +2,16 @@
   <nav>
     <ul class="tabs" data-hook="admin_settings_areas_tabs">
       <% if can? :display, Spree::Zone %>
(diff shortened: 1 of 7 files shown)


jhawthorn:
Looks good to me. Thanks again. :+1:

mamhoff:
Yep, I love namespacing path helpers. :+1:

mamhoff:
Would you be so kind to also rebase this on the current master?

eric1234:
I'll have to come back to it in a day or so as #975 restructured things so much (for the better) I'll have to re-implement.

eric1234:
Only took a month and a half :) but finally came back around to updating this PR so that the code works with the new admin config navigation.

jhawthorn:
Thanks again @eric1234

PR #985, opened 2016-03-14
Make use of Backbone in shipments edit page

jhawthorn:
The `shipments.js` file had become a sprawl of jQuery soup. This aims to introduce some structure by moving the events into a Backbone view. This also cleaned up the event handlers a fair bit since they now have access to the shipment's top level div as well as the order and shipment number.

This should not change any behaviour.

There's certainly more work to be done here, but I believe this is a good first step.


cbrunsdon:
The changes all look good to me so :+1:, with disclaimer I haven't pulled it down locally and ran it.

PR #987, opened 2016-03-16
WIP Superseding historical prices

mamhoff:
This is to replace the `is_default` boolean on prices. 

That boolean governed what price would be currently valid. I'd like to replace that with logic that can tell you what price a variant was at any given point. In order to do that, I introduce a `valid_from` field, which defaults to the time the price is created and indefinitely valid afterwards. There is no `valid_until` field because the product has `available_until` and we don't want to run into a product ever *not* having a price. 

@gmacdougall mentioned that this might have caching implications. @jhawthorn Please also have a look at this.

### DIFF
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -9,8 +9,6 @@ module Spree
         inverse_of: :variant,
         dependent: :destroy,
         autosave: true
-
-      after_save :save_default_price
     end
 
     def find_or_build_default_price
@@ -27,16 +25,6 @@ module Spree
 
     def has_default_price?
       !default_price.nil?
-    end
-
-    private
-
-    def default_price_changed?
-      default_price && (default_price.changed? || default_price.new_record?)
-    end
-
-    def save_default_price
-      default_price.save if default_price_changed?
     end
   end
 end
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -1,21 +1,44 @@
 module Spree
+  # Adds a `#default_price` `has_one` relation to the including model
+  #
+  # The default price is the price displayed in the admin as the main price for a product,
+  # for example on the Spree::Product#index page.
+  # It also defines getters for price-related properties on the variant that will return the
+  # properties of the default price.
+  # If you change the default price on a variant using `Spree::Variant#price=`, a new price object
+  # will be created and the old one soft-deleted. The deletion can not be avoided:
+  #
+  # ActiveRecord destroys stale `has_one` records, as one might expect from a 1:1 relation:
+  # https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/associations/has_one_association.rb#L35
+  # https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/associations/has_one_association.rb#L76-L91
+  # There is no way to prevent that, unfortunately, so we have to rely on `acts_as_paranoid`
+  # to keep historical records.
+  #
+  # We have to use an ActiveRecord association though, because the backend forms need it for Ransacks
+  # searching and sorting functionality.
+  #
   module DefaultPrice
     extend ActiveSupport::Concern
 
     included do
+      # Price to be used in the backend
+      # @return [Spree::Price] The backend price for the including object
       has_one :default_price,
-        -> { with_deleted.where(currency: Spree::Config[:currency]).valid_before_now },
+        -> { Spree::Price.default_prices },
         class_name: 'Spree::Price',
         inverse_of: name.demodulize.underscore.to_sym,
         dependent: :destroy,
         autosave: true
     end
+    # Delegate getters for price-related variant properties to the relation defined above.
     delegate :display_price, :display_amount, :price, :currency, to: :default_price, allow_nil: true
 
+    # Build a new price object and soft-delete the current one so we keep a history of prices.
     delegate :price=, to: :build_default_price
 
+    # @return [Boolean] Does the including object (ie: Variant) have a default price already?
     def has_default_price?
-      !default_price.nil?
+      default_price.present?
     end
   end
 end
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -4,7 +4,7 @@ module Spree
 
     included do
       has_one :default_price,
-        -> { where currency: Spree::Config[:currency], is_default: true },
+        -> { where(currency: Spree::Config[:currency]).valid_before_now },
         class_name: 'Spree::Price',
         inverse_of: :variant,
         dependent: :destroy,
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -4,24 +4,15 @@ module Spree
 
     included do
       has_one :default_price,
-        -> { where(currency: Spree::Config[:currency]).valid_before_now },
+        -> { with_deleted.where(currency: Spree::Config[:currency]).valid_before_now },
         class_name: 'Spree::Price',
         inverse_of: :variant,
         dependent: :destroy,
         autosave: true
     end
+    delegate :display_price, :display_amount, :price, :currency, to: :default_price, allow_nil: true
 
-    def find_or_build_default_price
-      default_price || build_default_price
-    end
-
-    delegate :display_price, :display_amount,
-      :price, :price=, :currency, :currency=,
-      to: :find_or_build_default_price
-
-    def default_price
-      Spree::Price.unscoped { super }
-    end
+    delegate :price=, to: :build_default_price
 
     def has_default_price?
       !default_price.nil?
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -6,7 +6,7 @@ module Spree
       has_one :default_price,
         -> { with_deleted.where(currency: Spree::Config[:currency]).valid_before_now },
         class_name: 'Spree::Price',
-        inverse_of: :variant,
+        inverse_of: name.demodulize.underscore.to_sym,
         dependent: :destroy,
         autosave: true
     end
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -5,7 +5,8 @@ module Spree
     MAXIMUM_AMOUNT = BigDecimal('99_999_999.99')
 
     belongs_to :variant, -> { with_deleted }, class_name: 'Spree::Variant', touch: true
-    before_save :valid_from_today!, if: -> { valid_from.blank? }
+
+    after_initialize :valid_from_today!, if: -> { valid_from.blank? }
 
     validate :check_price
     validates :amount, allow_nil: true, numericality: {
@@ -13,11 +14,11 @@ module Spree
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
 
+    scope :in_currency, -> (currency) { where(currency: currency) }
     scope :latest_valid_from_first, -> { order(valid_from: :desc) }
-    scope :valid_before, -> (date) { where("#{Spree::Price.table_name}.valid_from <= ?", date) }
+    scope :valid_before,
+          -> (date) { latest_valid_from_first.where("#{Spree::Price.table_name}.valid_from <= ?", date) }
     scope :valid_before_now, -> { valid_before(Time.current) }
-
-    after_save :set_default_price
 
     extend DisplayMoney
     money_methods :amount, :price
@@ -50,13 +51,6 @@ module Spree
 
     def check_price
       self.currency ||= Spree::Config[:currency]
-    end
-
-    def set_default_price
-      if is_default?
-        other_default_prices = variant.prices.where(currency: self.currency, is_default: true).where.not(id: id)
-        other_default_prices.update_all(is_default: false)
-      end
     end
   end
 end
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -20,6 +20,11 @@ module Spree
           -> (date) { latest_valid_from_first.where("#{Spree::Price.table_name}.valid_from <= ?", date) }
     scope :valid_before_now, -> { valid_before(Time.current) }
 
+    # Returns a cache key for all prices in a passed in relation
+    def self.cache_key
+      Digest::MD5.hexdigest(valid_before_now.pluck(:id, :updated_at).flatten.join("/"))
+    end
+
     extend DisplayMoney
     money_methods :amount, :price
 
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -1,5 +1,11 @@
 module Spree
   class Price < Spree::Base
+    # Because Rails destroys stale has_one records (see the Spree::DefaultPrice module for further
+    # information), we need `acts_as_paranoid` to keep historical records for us.
+    #
+    # The price relations on `Spree::Variant` are all scoped `with_deleted`, as we can not rely
+    # on the `deleted_at` accurately reflecting whether we have the right record. We rather use
+    # `valid_from` to figure out which price is the correct one to use.
     acts_as_paranoid
 
     MAXIMUM_AMOUNT = BigDecimal('99_999_999.99')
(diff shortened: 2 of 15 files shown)


gmacdougall:
An example of where we have caching that will be affected by this:

https://github.com/solidusio/solidus/blob/96f186c3bda4fa4636ea1dc352c03012ae293d16/frontend/app/views/spree/products/show.html.erb#L1

The product show page is currently cached on Locale, Currency, and Product and we rely upon Price touching Variant and Variant touching Product to bust the cache.

mamhoff:
That should work, though: Whenever I update, create or destroy a price belonging to a variant, that variant and in turn the associated product will be touched. 

Alternatively, we could create a more sophisticated cache key for products; something along the lines of
``` ruby
class Product
  def cache_key
    associated_object_timestamps = [updated_at] + variants.map(&:updated_at) + prices.map(&:updated_at) 
    hashed_timestamps = Digest::MD5.hexdigest associated_object_timestamps.map(&:to_s).join(",")
    super + hashed_timestamps
  end
end
```

Right now, that should bust the cache quite reliably whenever we update, add or remove any variants or prices. 

In the future, we'll also have to look at modifying the `*cache_key* methods in the core helpers to take into account the current order's tax location. But that is outside the scope of this PR.

mamhoff:
@gmacdougall and I talked IRL (well, on the interwebs) and he explained that time itself should expire the cache key. I've cooked something up by looking up `prices.currently_valid` (which uses `Time.current`) to figure out which prices are currently possibly valid, and that correctly expires the cache when a new price becomes relevant. I've also added a spec to check for changing of the current price (which does the touching chain `price -> variant -> product`, and also correctly busts the cache. Would this work?

mamhoff:
@gmacdougall @jhawthorn Please have a look at this, there is a caching strategy and a migration.

core/app/models/spree/product.rb
```diff
@@ -274,6 +274,10 @@ def find_variant_property_rule(option_value_ids)
       end
     end
 
+    def cache_key
+      super + "/prices/" + prices.currently_valid.pluck(:id).join("/")
```
gmacdougall: I think it would be best to digest the prices. I don't know if we'll encounter an issue with key length if we have thousands of prices, but it's better to err on the side of safety.

core/spec/models/spree/product_spec.rb
```diff
@@ -14,6 +14,38 @@ class Extension < Spree::Base
...
+            allow(Time).to receive(:current).and_return(1.day.from_now)
+            expect(product.cache_key).to eq(old_cache_key)
+          end
+
+          context 'when updating the price' do
+            it 'changes the cache key' do
+              old_cache_key = product.cache_key
+              allow(Time).to receive(:current).and_return(1.day.from_now)
+              current_price.update(amount: current_price.amount + 1)
+              expect(product.cache_key).to_not eq(old_cache_key)
```
gmacdougall: I think that this test might be best expressed using an assertion like:

```Ruby
expect { subject }.to chance { product.cache_key }
```

Since that's what's important here.

core/app/models/spree/product.rb
mamhoff: I thought about this but went for "the simplest thing that could possibly work". I do agree with digesting we'd be safe from key length issues. 

Something that's crossed my mind though: Should we maybe ask the currently used pricer make sure the correct cache key is used?

core/spec/models/spree/price_spec.rb
```diff
@@ -1,6 +1,63 @@
...
+        subject.save
+        expect(subject.valid_from).to eq(Date.new(1990, 01, 01))
+      end
+    end
+
+    context 'with no valid from date given' do
+      let(:valid_from) { nil }
+
+      before do
+        expect(Time).to receive(:current).at_least(:once).and_return(Date.new(2015, 02, 28))
```
gmacdougall: I would prefer to use the Timecop library for this, which is already a dependency in Solidus.

core/app/models/spree/product.rb
gmacdougall: Yes, I think that's makes sense. I'm not sure which PR you want to do that in.

mamhoff: Can absolutely be this one. I'm thinking after #994 is in, which gives us a first go at how a pricer would be structured, it could be quite neatly tucked into a new pricer that's less stupid than the `ConservativeLineItemPricer`.

mamhoff: Nope, coming to think of it: This PR should really just introduce this feature. `Variant#prices.currently_valid` will return all prices (all currencies and so on), and custom pricers will just add more attributes to select prices from - this caching strategy would still work in most cases. I'll work on your suggestions (use `timecop` and use the `change` matcher).

core/app/models/spree/price.rb
```diff
@@ -6,13 +6,25 @@ class Price < Spree::Base
...
+    scope :valid_before, -> (date) { where("#{self.table_name}.valid_from <= ?", date).order(valid_from: :desc) }
+    scope :currently_valid, -> { valid_before(Time.current) }
+
+    def self.default
+      currently_valid.first
+    end
+
+    # Returns a cache key for all prices in a passed in relation
+    def self.cache_key
+      Digest::MD5.hexdigest(pluck(:id, :updated_at).flatten.join("/"))
```
gmacdougall: We'll need to order the results from `pluck` somehow to ensure that they're returned in a consistent order.

core/spec/models/spree/product_spec.rb
```diff
@@ -14,6 +14,53 @@ class Extension < Spree::Base
...
+
+          context 'when updating the price' do
+            it 'changes the cache key' do
+              expect {
+                Timecop.travel(1.day.from_now)
+                current_price.update(amount: current_price.amount + 1)
+              }.to change { product.cache_key }
+            end
+
+            it 'does not change the length of the cache key' do
```
gmacdougall: I don't think this assertion is important.

mamhoff:
Wow, this is much more of a pain than I thought, mostly because saving a new item on a `has_one` relation *deletes* the old one, which is behaviour I'm not happy about. We need a relation though because otherwise Ransack won't work.

mamhoff:
@gmacdougall @jhawthorn This is done, as far as I'm concerned. Would you have a look? :)

core/db/migrate/20160318130914_migrate_default_to_current_prices.rb
```diff
@@ -0,0 +1,34 @@
+class MigrateDefaultToCurrentPrices < ActiveRecord::Migration
+  def up
+    say_with_time 'Setting default prices to be valid from now' do
```
adammathys: I don't think the `say_with_time` is necessary when all it's wrapping is an `execute`. But it doesn't really hurt to have it.

core/spec/support/concerns/default_price.rb
```diff
@@ -16,6 +16,31 @@
...
+        expect(default_price_association.options[:inverse_of]).to eq(:variant)
+      end
+    end
+
+    context 'when used outside of the Spree namespace' do
+      before do
+        module MyStore
+          class Bicycle < ActiveRecord::Base
+            include Spree::DefaultPrice
+            table_name = "spree_variants"
```
adammathys: I agree with @houndci-bot, I don't think `table_name` is doing anything here.

core/app/models/spree/price.rb
```diff
@@ -12,7 +13,12 @@ class Price < Spree::Base
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
 
-    after_save :set_default_price
+    scope :valid_before, -> (date) { where("#{self.table_name}.valid_from <= ?", date).order(valid_from: :desc) }
```
tvdeyen: I like to write this as `where("#{Spree::Price.table_name}.valid_from <= ?", date)` nowadays.

core/app/models/spree/price.rb
```diff
@@ -1,18 +1,35 @@
...
+    scope :valid_before_now, -> { valid_before(Time.current) }
+    scope :with_default_currency, -> { where(currency: Spree::Config.currency) }
+
+    # The scope to be passed into a `default_price` association for an object that needs a default price
+    # @return [ActiveRecord::Relation] A scope where the first object is the current default price
+    scope :default_price_scope, -> { with_deleted.with_default_currency.valid_before_now }
```
tvdeyen: I would like the scope to be called `Spree::Price.default_prices` or similar

core/app/models/concerns/spree/default_price.rb
```diff
@@ -1,42 +1,34 @@
...
+  # will be created and the old one soft-deleted.
   module DefaultPrice
     extend ActiveSupport::Concern
 
     included do
+      # Price to be used in the backend
+      # @return [Spree::Price] The backend price for the including object
       has_one :default_price,
-        -> { where currency: Spree::Config[:currency], is_default: true },
+        -> { default_price_scope },
```
tvdeyen: Can you write this more explicit as `Spree::Price.default_prices`?

core/app/models/spree/price.rb
```diff
@@ -6,13 +6,21 @@ class Price < Spree::Base
...
+    after_initialize :valid_from_today!, if: -> { valid_from.blank? }
+
     validate :check_price
     validates :amount, allow_nil: true, numericality: {
       greater_than_or_equal_to: 0,
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
 
-    after_save :set_default_price
+    scope :valid_before, -> (date) { where("#{self.table_name}.valid_from <= ?", date).order(valid_from: :desc) }
```
tvdeyen: If you want this scope to be called elsewhere, you need to include `with_deleted` scope

core/app/models/spree/price.rb
```diff
@@ -1,18 +1,35 @@
...
+
     validates :amount, allow_nil: true, numericality: {
       greater_than_or_equal_to: 0,
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
 
-    after_save :set_default_price
+    scope :valid_before, -> (date) { where("#{self.table_name}.valid_from <= ?", date).order(valid_from: :desc) }
+    scope :valid_before_now, -> { valid_before(Time.current) }
+    scope :with_default_currency, -> { where(currency: Spree::Config.currency) }
```
tvdeyen: also here you need to include `with_deleted` scope

core/app/models/spree/product.rb
```diff
@@ -274,6 +274,15 @@ def find_variant_property_rule(option_value_ids)
       end
     end
 
+    # Takes the product's cache key and adds a cache key for the current prices relation.
+    #
+    # This way, we expire the cache every time a new price becomes valid via its `valid_from`
+    # property.
+    #
+    def cache_key
+      super + "/prices/" + prices.valid_before_now.cache_key
```
tvdeyen: sweet solution :lollipop:

tvdeyen: As I read this now, I would like to have the scopes inside of `Spree::Price.cache_key`

tvdeyen: "ask, don't tell" ;)

core/app/models/spree/price.rb
```diff
@@ -1,18 +1,35 @@
...
+
+    # The scope to be passed into a `default_price` association for an object that needs a default price
+    # @return [ActiveRecord::Relation] A scope where the first object is the current default price
+    scope :default_price_scope, -> { with_deleted.with_default_currency.valid_before_now }
+
+    # Returns a fixed-length cache key for all prices in a passed in relation
+    #
+    def self.cache_key
+      Digest::MD5.hexdigest(pluck(:id, :updated_at).sort.flatten.join("/"))
```
tvdeyen: couldn't we spare the `.sort` as `valid_before` already has an order?

tvdeyen: This has possible implications I can't even imagine. But maybe I'm just anxious, since `cache_key` is called very often. @jhawthorn thoughts?

core/app/models/spree/variant.rb
```diff
@@ -39,6 +39,7 @@ class Variant < Spree::Base
     has_many :images, -> { order(:position) }, as: :viewable, dependent: :destroy, class_name: "Spree::Image"
 
     has_many :prices,
+      -> { with_deleted },
```
tvdeyen: we really need to explain, why we use `with_deleted` all the time on the prices.

core/app/models/spree/variant.rb
```diff
@@ -214,7 +215,10 @@ def option_value(opt_name)
     # @param currency [String] the desired currency
     # @return [Spree::Price] the price in the desired currency
     def price_in(currency)
-      prices.detect{ |price| price.currency == currency && price.is_default } || Spree::Price.new(variant_id: id, currency: currency)
+      prices.select { |price| price.valid_from <= Time.current }
```
tvdeyen: Isn't this just:

```ruby
prices.valid_before_now.where(Spree::Price.table_name => {currency: currency})
```

core/app/models/spree/variant.rb
```diff
@@ -214,7 +215,10 @@ def option_value(opt_name)
...
+        .sort_by(&:valid_from).reverse
+        .detect { |price| price.currency == currency } ||
+        Spree::Price.new(variant_id: id, currency: currency, valid_from: Time.current)
```
tvdeyen: And this:

```ruby
prices.build(currency: currency)
```

core/app/models/spree/variant.rb
```diff
@@ -328,6 +332,12 @@ def variant_properties
       end.flatten.compact
     end
 
+    # Dummy method for setting a variant's currency (which is impossible)
+    # @deprecated
+    def currency=(_currency)
+      ActiveSupport::Deprecation.warn "You can not meaningfully set a currency on a variant", caller
```
tvdeyen: If we deprecate stuff, we should at last implement the old, but wrong behavior.

```ruby
default_price.currency = currency
```

tvdeyen: Maybe also change the deprecation warning to something more explanatory.

core/app/models/spree/price.rb
```diff
@@ -1,18 +1,35 @@
...
+    # All price relations on `Spree::Variant` are scope `with_deleted`. The `valid_from` field
+    # tells us which price is currently valid.
     acts_as_paranoid
 
     MAXIMUM_AMOUNT = BigDecimal('99_999_999.99')
 
     belongs_to :variant, -> { with_deleted }, class_name: 'Spree::Variant', touch: true
 
-    validate :check_price
+    after_initialize :valid_from_today!, if: -> { valid_from.blank? }
```
tvdeyen: Isn't this better a column default in the database?

core/db/migrate/20160314162016_add_valid_from_to_spree_prices.rb
```diff
@@ -0,0 +1,5 @@
+class AddValidFromToSpreePrices < ActiveRecord::Migration
+  def change
+    add_column :spree_prices, :valid_from, :datetime
```
tvdeyen: Shouldn't we add `default: "NOW()"` here, so we can remove the `after_initialize` from price?

tvdeyen: Apparently this won't work in active record schema files.

core/app/models/spree/variant.rb
mamhoff: I elaborated more in the comments.

core/app/models/spree/price.rb
```diff
@@ -1,18 +1,41 @@
...
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
 
-    after_save :set_default_price
+    scope :in_currency, -> (currency) { where(currency: currency) }
+    scope :latest_valid_from_first, -> { order(valid_from: :desc) }
+    scope :valid_before,
+          -> (date) { latest_valid_from_first.where("#{Spree::Price.table_name}.valid_from <= ?", date) }
+    scope :valid_before_now, -> { valid_before(Time.current) }
+    scope :with_default_currency, -> { where(currency: Spree::Config.currency) }
```
tvdeyen: `in_currency(Spree::Config.currency)`

tvdeyen:
Perfect. Thanks for adding more docs and considering my feedback.

Someone please have a look into the SQL in the migrations as I'm not able to review it. @jhawthorn @gmacdougall 

:+1:

mamhoff:
This is now very out-of-date. Much of it has been taken in in other PRs. I'm closing this for the time being.

PR #990, opened 2016-03-17
Add LegacyLineItemPricer

mamhoff:
This moves setting the price of a line item to a special service object
called a `Pricer`. It supports price modifier options as well as not
changing the line item if no price modifiers are set and a price and
currency are already present.

Previously, the currency of a line item could be obtained from the
variant (which really can't know about the currency of a line item), so
that is newly streamlined into obtaining the currency from the if the
options don't say otherwise.

This is supposed to be a basis for discussion. 

I tried designing the interface of the object in such a way that existing 
`copy_price` overrides could easily be migrated into a `Pricer` object - 
hence the pattern of modifying the line item in-place.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -102,17 +102,8 @@ module Spree
     #
     # @param options [Hash] options for this line item
     def options=(options = {})
-      return unless options.present?
-
-      opts = options.dup # we will be deleting from the hash, so leave the caller's copy intact
-
-      currency = opts.delete(:currency) || order.currency
-
-      self.currency = currency
-      self.price    = variant.price_in(currency).amount +
-        variant.price_modifier_amount_in(currency, opts)
-
-      assign_attributes opts
+      Spree::Prices::LegacyLineItemPricer.set_price_for(self, options)
+      assign_attributes(options)
     end
 
     private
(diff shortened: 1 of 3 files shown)


cbrunsdon:
I'm never a big fan of calling anything "Legacy" something, is there a more descriptive name we can give this that better describes its behavior? (and maybe would be clear in its differences when we create other ones).

mamhoff:
`PriceModifierAwareOnlyChangesPriceWhenModifiersPresentPricer`? The current pricing behaviour has some quirks that prevent a descriptive, short name. I'm absolutely open for suggestions :D

adammathys:
Like @cbrunsdon, I'm not crazy about calling things `Legacy`. I don't have a better alternative though.

Big :+1: on the general direction of this change from me. Moving chunks of related logic out of the "core" models (`LineItem`, `Order`, `Shipment`, etc.) is something I could see being useful in a number of places. That being said, I'm not overly fond of objects that only modify fields on another object. Sort of goes against the "I" in SOLID, leading to code that ends up being harder to test and refactor later on. (In order to test a `Pricer`, we make assertions on a `Spree::LineItem`. This principle is violated all over the place in Solidus, which in my opinion is one of the many reasons it's so hard to write good tests for much of the project.)

Personally, I'd be happier with a toned back version of this change that introduced a `LegacyLineItemPrice` (:tm:) that was responsible for determining what amount and in which currency a `LineItem` should be priced. Leaving the logic for assigning the information to the columns on the `LineItem` itself.

jhawthorn:
In addition to others comments I'd love to see `LegacyLineItemPricer` use instance instead of class methods. In the current implementation the instance has no defined behaviour. All of the methods take in a line_item, storing that as an instance variable should help clean that up.

mamhoff:
Thanks everyone for the feedback. I think the difficulty in naming comes from the fact that this class does way too much, and I'll introduce smaller `Pricers` that are more easily names and (yep, @jhawthorn) take a line item as instance variable and go from there. 

This was a very valuable spike :)

mamhoff:
Closing in favour of https://github.com/solidusio/solidus/pull/994 and similar PRs

PR #992, opened 2016-03-18
Add timestamps to spree_prices table

mamhoff:
All other models have timestamps, and prices should have them, too!


Senjai:
:shipit:

jhawthorn:
:+1:

PR #993, opened 2016-03-18
Assign a line items currency as an attribute, not an option

mamhoff:
When extracting pricing to service objects I came across a number of
ways of setting a line item's currency. In `OrderContents`, the currency,
although an attribute of the line item, is passed to the line item
through an options Hash designed for an extension point introduced around
Spree 2.2/2.3 while it could just be assigned directly like any other
attribute.

This PR refactors that bit of weird code, hopefully relieving us of the
`options=` method on line item entirely in the future.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -104,15 +104,10 @@ module Spree
     def options=(options = {})
       return unless options.present?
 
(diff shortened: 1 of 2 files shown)


Senjai:
:+1:

mamhoff:
Also closing this in favour of #998. We don't even need this attribute, and it only creates confusion.

PR #994, opened 2016-03-18
Conservative line item pricer

mamhoff:
This is the pricer that does what `LineItem` does most of the time: Not change anything if there is a price, and if there isn't, take the one from the variant. 

It's modified a little bit so that it would always take the price in the currency from the current order/line item, but otherwise very much the same behaviour. 

Stay tuned for more `Pricers`!

### DIFF
--- a/core/app/models/spree/line_item/pricers/abstract.rb
+++ b/core/app/models/spree/line_item/pricers/abstract.rb
@@ -1,0 +1,26 @@
+module Spree
+  class LineItem
+    module Pricers
+      class Abstract
+        class VariantMissing < StandardError; end
+        class CurrencyMissing < StandardError; end
+
+        attr_reader :line_item
+
+        def initialize(line_item)
+          @line_item = line_item
+        end
+
+        def price
+          raise NotImplementedError, "Please implement '#price' in your pricer: #{self.class.name}"
+        end
+
+        private
+
+        def in_line_item_currency(amount)
+          Spree::Money.new(amount, currency: line_item.currency)
+        end
+      end
+    end
+  end
+end
--- a/core/app/models/spree/line_item/pricers/abstract.rb
+++ b/core/app/models/spree/line_item/pricers/abstract.rb
@@ -1,6 +1,16 @@
 module Spree
   class LineItem
+    # Pricers for line items expose `#price`, return a `Spree::Money` object,
+    # and are used to choose or calculate the correct price for a line item.
+    # Ideally, they are idempotent, that is: When run on the same line item, they will return
+    # the same price.
+    #
     module Pricers
+      # The `Abstract` pricer is here for other pricers to inherit from. It defines the common
+      # interface for pricers, as well as two practical errors, `VariantMissing` and
+      # `CurrencyMissing`.
+      #
+      # @attr [Spree::LineItem] The line item to calculate or choose a price for
       class Abstract
         class VariantMissing < StandardError; end
         class CurrencyMissing < StandardError; end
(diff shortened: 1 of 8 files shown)


core/app/models/spree/pricers/conservative_line_item_pricer.rb
```diff
@@ -0,0 +1,23 @@
...
+      class VariantMissing < StandardError; end
+      class CurrencyMissing < StandardError; end
+
+      attr_reader :line_item
+
+      def initialize(line_item)
+        @line_item = line_item
+      end
+
+      def price
```
gmacdougall: Could we get some YARD doc for these methods, and the class?

For a method called `price` I would expect it to return a `Spree::Price` but instead it returns a BigDecimal. I would like to avoid returning `BigDecimal` in any of these scenarios. I feel it makes more sense to return a Money object or Spree::Price.

core/app/models/spree/pricers/conservative_line_item_pricer.rb
```diff
@@ -0,0 +1,23 @@
+module Spree
+  module Pricers
+    class ConservativeLineItemPricer
+      class VariantMissing < StandardError; end
```
gmacdougall: I think these exceptions should be on a shared parent class.

core/app/models/spree/pricers/conservative_line_item_pricer.rb
```diff
@@ -0,0 +1,23 @@
+module Spree
+  module Pricers
+    class ConservativeLineItemPricer
```
gmacdougall: I don't think the Conservative portion of the name adds anything. This is just a standard line item pricer, so I'd call it `LineItemPricer` or `StandardLineItemPricer`.

mamhoff: I think it does. The pricer I want to see will change the price even if it's already present, and I'd like to make that the `standard` one. It's not there yet...

core/app/models/spree/pricers/conservative_line_item_pricer.rb
mamhoff: Yep, I agree.

core/app/models/spree/pricers/conservative_line_item_pricer.rb
mamhoff: YARD: Sure. 
Money object: Makes sense, requires a bit of tweaking here and there, and also the other `PriceModifierLineItemPricer` to be reworked a little.

core/app/models/spree/pricers/conservative_line_item_pricer.rb
gmacdougall: That makes sense. I think some docs would help there.

core/app/models/spree/pricers/conservative_line_item_pricer.rb
mamhoff: I don't want to use a `Spree::Price`, as that would imply having the currency stored redundantly AGAIN. :)

mamhoff:
@gmacdougall Implemented all suggestions, I think. This made adding a new `money_price=` setter on line item necessary that handles using a `Spree::Money` or just plain `Money` object as the price. 

It'll take some time for the tests to run, but I'm quite sure they're green. Ready to go. 

I've also changed the naming and namespacing of the pricers, linking them more closely to `Spree::LineItem`.

core/app/models/spree/line_item/pricers/abstract.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        attr_reader :line_item
+
+        def initialize(line_item)
+          @line_item = line_item
+        end
+
+        # Returns the calculated or chosen price
+        #
+        # @return [Spree::Money, Money] an money object representing line item's price
```
gmacdougall: I think this should just be `@return [Spree::Money]`

mamhoff: A pricer's return value will be sent `to_d` and `currency`, expecting `::Money`s signature. I thought I'd allow a plain money object to start phasing out `Spree::Money`.

adammathys:
:+1: I like the abstraction and general direction.

gmacdougall:
Looking good to me. Could you please rebase this?

mamhoff:
@gmacdougall rebased, ready to merge!

gmacdougall:
:+1:

PR #995, opened 2016-03-18
Get line item currency from order

mamhoff:
There seems to be a way of obtaining a line item's currency from
it's variant. That doesn't make much sense, as a variant might have
many prices in different currencies, but is not tied to a user or an
order which would know *which* of those currencies to use.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -136,7 +136,7 @@ module Spree
       # If the legacy method #copy_price has been overridden, handle that gracefully
       return handle_copy_price_override if respond_to?(:copy_price)
 
-      self.currency ||= variant.currency
+      self.currency ||= order.currency
       self.cost_price ||= variant.cost_price
       self.price ||= variant.price
     end


adammathys:
:+1:

jhawthorn:
:+1: Yikes. `variant.currency` is wrong and makes no sense

tvdeyen:
:+1:

mbj:
No specs?

mamhoff:
@mbj Good point, and thanks for sorta implicitly pointing me to your work on this. I'm closing this in favour of #998 - that one actually has specs and cleans the relationship with a line item and it's currency up in a more thorough way.

mbj:
@mamhoff This project currently does not automatically verify code changes are paired with covering specs. While people seem to review code changes, the reviews are naturally (as we are all humans) of varying quality. 

Having even an imperfect automated system in place that verifies the semantic change coverage will trigger the humans into performing a much higher quality review, as such systems never have a bad day.

When we have our tax hangout, I might be able to convince you to put such a system in-place.

PR #996, opened 2016-03-18
Remove hanging tooltips from removed DOM elements

jhawthorn:
This is a bit of a hack, but I don't know of a better alternative, and it's fairly self-contained.

When a DOM element is removed, any associated tooltips will remain on the page permanently, because the element will never receive the `mouseleave` event. Our previous tooltip solution, powerTip, [solved this using polling](https://github.com/stevenbenner/jquery-powertip/blob/dc5972bcbbd4229043bb1cf42ab7004c8846277e/src/tooltipcontroller.js#L359-L399). This does the same.

Hiding these tooltips when we remove elements isn't feasible because it's too common for us to remove elements from the DOM (both due to removal and re-rendering mustache templates).

It's possible this will be unnecessary in a future version of bootstrap, as they're planning to investigate using the new MutationObserver API to detect these changes to the DOM.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -15,14 +15,6 @@ Hopefully, this will evolve into a propper class.
 **/
 
 jQuery(function($) {
-  $('body').tooltip({selector: '.with-tip'})
-
-  $('body').on('inserted.bs.tooltip', function(e){
-    var $target = $(e.target);
-    var $tooltip = $("#" + $target.attr("aria-describedby"));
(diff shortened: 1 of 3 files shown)


stewart:
:+1: it's an effective hack that cleans up after itself well.

mamhoff:
:+1: - Again, my JS is limited, but this looks effective and not too intrusive as it only ever starts working when a tooltip element is instantiated. Great!

cbrunsdon:
:+1:. 

I can't find any specific discussion that seemed relevant on it actually getting implemented using the  ```MutationObserver``` other than a bullet point here: https://github.com/twbs/bootstrap/pull/17021

PR #997, opened 2016-03-19
Replace ul/li-s with a div, place weight last

docelic:
This commit replaces use of ul/li-s on the product form with plain
divs, achieving a much nicer visual layout for the product height/
width/depth/weight fields, which were previously clunky and misaligned.

The weight field has been moved from the 1st to the 4th position,
for improved intuitiveness in the new layout.

How it looked before:

before

How it looks now:

after

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -97,24 +97,24 @@
         <% end %>
       </div>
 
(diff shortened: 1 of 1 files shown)


jhawthorn:
Much better. Thank you. :+1:

gmacdougall:
:+1: Thanks for the contribution!

PR #635, opened 2015-12-30
Remove Order::CurrencyUpdater

jhawthorn:
I've split this off from #425, as I think the remainder of that PR (OrderUpdater change) scared people more than this portion. 

This was intended to be used to update line items in an order to a new currency. Nowhere do we allow reassigning a currency to an existing order, or to line items, so this should not be needed.

PR #998, opened 2016-03-20
Remove Line Item currency attribute

mamhoff:
Somehow, in the past, it was possible to have orders with line items
in multiple currencies. The code to remedy this (Order#homogenize_line_item_currencies)
disappeared in #635, and now we have a bunch of code in different places
that allows setting it, raises errors if it's not there, and so on.

This commit delegates the line items currency to the order.
Whenever people try *setting* a line item's currency, they get a
pretty deprecation warning telling them not to do so.

The error handling functionality that seems to be used in the order
importer is still present (ie: If we know an imported line item has
a currency different from the order, we'll error out).

This commit replaces all that behaviour with a simple delegate that
always obtains a line item's currency from it's order.

Adds a changelog entry.

I also refactored `OrderContents#add_to_line_item`, taking code from
MountainRoseHerbs/spree@461f975.

That method should now be much easier to understand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Solidus 1.3.0 (unreleased)
 
+*   Remove currency attribute on `Spree::LineItem`
+
+    A valid line item always has its order's currency. Therefore it is not necessary to save
+    that in the database.
+
 *   Taxes for carts now configurable via the `Spree::Store` object
 
     In VAT countries, carts (orders without addresses) have to be shown with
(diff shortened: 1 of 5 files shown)


gmacdougall:
I do think that this makes sense. However, I am concerned about potentially destroying data that people were relying upon if they had customized their store. I would appreciate some more opinions on this one.

mamhoff:
Closing in favour of #1007, which does not remove the column but makes handling line item currencies saner.

PR #999, opened 2016-03-21
Introduce PriceModifierLineItemPricer

mamhoff:
There is an extension point for modifying the price of a line item
that is used in the following extensions:

* https://github.com/godaddy/spree_product_sale
* https://github.com/v-may/spree_simple_sales

This extension point introduces some rather interesting code to
Spree::LineItem, which this commit extracts to a pricer object.

The extension point, as far as I can overview, needs the following to work:

- `*_price_modifier_amount_in` methods defined on the variant
- Fitting `*` methods on `Spree::LineItem`
- The Line Item to accept those methods as params

This commit extracts that code from Spree::LineItem, with the intention
of providing an example for how to write a custom pricer as well as with
the intention to remove this pricer and the associated extension point by
Solidus 2.0.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -103,16 +103,9 @@ module Spree
     # @param options [Hash] options for this line item
     def options=(options = {})
       return unless options.present?
-
-      opts = options.dup # we will be deleting from the hash, so leave the caller's copy intact
-
-      currency = opts.delete(:currency) || order.currency
-
-      self.currency = currency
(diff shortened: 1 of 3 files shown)


gmacdougall:
I like that we've extracted this functionality to a better place and put it under test. I would prefer to deprecate it with the intention of removing it. It can only be customized through monkey patching, and the pricer system as a whole will provide similar, but much less bad versions of that functionality.

mamhoff:
I'm closing this for the time being, as #994 has better naming and stuff. Let's do that first.

PR #1000, opened 2016-03-21
Stock management js cleanup

jhawthorn:
I started working on this because the JS was binding directly to buttons based on the font awesome icon class (yuck!). I then discovered that the view template was copy pasted between rails views and handlebars.

This simplifies the javascript in the stock management section. There should be no behaviour change.

* Use a backbone view for each row.
* Use backbone models to represent the StockItem
* StockItem "edit" rows are now rendered entirely in JS by handlebars rather than having duplicated views.
* We use a tbody to group rows all concerned with the same variant.
* Style even/odd rows in CSS instead of adding classes in JS
* Bind events to reasonable classes instead of icons liker `.fa-check`

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/sections/_stock_management.scss
+++ b/backend/app/assets/stylesheets/spree/backend/sections/_stock_management.scss
@@ -7,17 +7,17 @@
   }
 
   > tbody {
-    > tr {
-      > td {
-        background-color: $color-tbl-odd;
-      }
-      &.even td {
+    > tr.stock-item-edit-row {
+      td {
         background-color: $color-tbl-even;
       }
-      &:first-child {
-        td {
-          border-top: none;
-        }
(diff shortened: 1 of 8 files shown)


backend/app/assets/stylesheets/spree/backend/sections/_stock_management.scss
```diff
@@ -7,17 +7,17 @@
...
-      &.even td {
+    > tr.stock-item-edit-row {
+      td {
         background-color: $color-tbl-even;
       }
-      &:first-child {
-        td {
-          border-top: none;
-        }
+      &:nth-child(2n) td {
```
graygilmore: You can express this as `nth-child(even)` if you want it to read a little nicer.

cbrunsdon:
Looks good to me, thanks hawth. I guess these are all towards our eventual fixing of FA?

PR #1002, opened 2016-03-22
Add gem depencies on sass-rails and coffee-rails

jhawthorn:
These gems are required to run the frontend and backend projects, so we should declare them as necessary and require them instead of relying on them to be a part of the default rails application.

This should fix issues running specs for a few extensions where sass-rails was not being pulled in through other means.


cbrunsdon:
Looks good to me, thanks.

PR #1004, opened 2016-03-23
Encapsulate the logic for when an item is shippable.

eric1234:
Transitioning an item (InventoryUnit) to `shipped` is restricted [allowing an item to ask itself if it is shippable before it is marked as shipped](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/inventory_unit.rb#L58). This questioning is implemented by [verifying the item is on-hand](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/inventory_unit.rb#L118-L120).

But when the overall shipment is checking to see if the item is shippable and [it does it's own interrogation of the items in the shipment rather than asking the items itself](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/shipment.rb#L84).

Worse is that the implementation differs from the implementation in InventoryUnit. InventoryUnit questions if the item is "on-hand" while in Shipment it questions if the item is not backordered. Unsure if it is a real issue but it might mean we could get a case where the inventory unit is in neither state allowing the shipment to think it can ship while the inventory unit thinks it can't resulting in an error.

This commit encapulates the decision entirely in InventoryUnit allowing the item to determine if it is ready to ship. This logically should be the same as before only we now have better encapsulation.

The additional benefit (and primary motivation) is that the condition can customized now that it is properly encapsulated. I am working on an app with on-demand manufactoring with no real inventory. An item cannot be shipped until it is built. We can customize InventoryUnit as follows:

    Spree::InventoryUnit.class_eval do
      module CheckBuiltForShipping do
        def allow_ship?
          super && built?
        end
      end
      prepend CheckBuiltForShipping
    end

Prior to this commit, the system would allow the user to attempt to ship the item (since it is not "backordered") but the ship would fail since our customization made it not shippable. With this commit the shipment it now not shippable also meaning the user isn't even presented with the option.

In theory this commit should have required no test changes as it is just refactoring implementation details. Unfortunately due to the abuse of mocks there are some adjustments to the test to get things passing.

### DIFF
--- a/core/app/models/spree/inventory_unit.rb
+++ b/core/app/models/spree/inventory_unit.rb
@@ -113,11 +113,11 @@ module Spree
       return_items.not_expired.any?(&:exchange_requested?)
     end
 
-    private
-
     def allow_ship?
       on_hand?
     end
+
+    private
 
     def fulfill_order
       reload
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -81,7 +81,8 @@ module Spree
     end
 
     def can_transition_from_pending_to_ready?
-      order.can_ship? && !inventory_units.any?(&:backordered?) &&
+      order.can_ship? &&
+        inventory_units.all? { |iu| iu.allow_ship? || iu.canceled? } &&
         (order.paid? || !Spree::Config[:require_payment_to_ship])
     end
 
@@ -222,10 +223,9 @@ module Spree
     # ready      all other cases
     def determine_state(order)
       return 'canceled' if order.canceled?
+      return 'shipped' if shipped?
       return 'pending' unless order.can_ship?
-      return 'pending' if inventory_units.any?(&:backordered?)
-      return 'shipped' if state == 'shipped'
-      if order.paid? || !Spree::Config[:require_payment_to_ship]
+      if can_transition_from_pending_to_ready?
         'ready'
       else
         'pending'
--- a/core/spec/models/spree/shipment_spec.rb
+++ b/core/spec/models/spree/shipment_spec.rb
@@ -56,7 +56,7 @@ describe Spree::Shipment, type: :model do
     end
 
     it 'returns pending if backordered' do
-      allow(shipment).to receive_messages inventory_units: [mock_model(Spree::InventoryUnit, backordered?: true)]
+      allow(shipment).to receive_messages inventory_units: [mock_model(Spree::InventoryUnit, allow_ship?: false, canceled?: false)]
       expect(shipment.determine_state(order)).to eq 'pending'
     end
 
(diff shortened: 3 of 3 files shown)


eric1234:
Just added an additional commit of some additional areas where logic is duplicated and non consistent.

core/spec/models/spree/order_shipping_spec.rb
```diff
@@ -138,7 +138,7 @@ def emails
       before { Spree::OrderCancellations.new(order).short_ship([order.inventory_units.first]) }
 
       it "updates the order shipment state" do
-        expect { subject }.to change { order.reload.shipment_state }.from('ready').to('shipped')
+        expect { subject }.to change { order.reload.shipment_state }.from('pending').to('shipped')
```
jhawthorn: I think this might be an important behaviour change (and unfortunately this is the only spec that caught it).

A shipment which has some "canceled" inventory units should still be "ready" and shippable

eric1234: If this is the case then it was previously broken. Previously the shipment went to the `ready

eric1234: If that case it was previously broken. Previously the shipment went to the `ready` state only because it was checking if inventory units where "not backordered". But just because the shipment was once again ready the individual inventory units where still in the "canceled" state and there is not path from "canceled" to shipped because `allow_ship?` is only true if on-hold.

This is actually a good example of why having the logic in two places (and be different) is problematic. On `master` if you cancel an item then try to ship it you will likely get an error because you cannot `ship` the inventory unit anymore since it is canceled and not on-hold.

With my new code the logic has a single source of truth which means since you cannot ship the inventory unit the shipment is also moved back to pending. If you want to make the spec pass as originally designed then we need to update the `allow_ship?` method in InventoryUnit to be `on_hand? || canceled?`.

If that sounds good I'll make the change and revert this test to it's original check.

eric1234: I am wrong above. I did a bit of testing with this and it doesn't try to `ship` the canceled item as [there is a filter to remove all but the on-hand ones](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_shipping.rb#L18). I'm working right now on making another pass at the code to try to remove some of this duplicated logic while still satisfying the existing specs.

eric1234:
K, a26b808 now addresses the regression.

jhawthorn:
This is excellent. :+1:

mamhoff:
This is really good. Can you try to rebase the regression fix into the actual change commit, as well as fix the actual change commits up with the hound commit? That way, we have no state in the repository with a regression, which would be nice.

eric1234:
Commits have been squashed and rebased to master.

jhawthorn:
Thank you

PR #1007, opened 2016-03-24
Always set a line item's currency to order currency

mamhoff:
Before, a line item could under certain circumstances be assigned
the store's default currency, rather than its order's currency.

With this commit, we're making sure Solidus does not do that, as
well as deprecating setting the currency manually to something
other than the order's currency.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -104,15 +104,10 @@ module Spree
     def options=(options = {})
       return unless options.present?
 
-      opts = options.dup # we will be deleting from the hash, so leave the caller's copy intact
+      self.price = variant.price_in(currency).amount +
+                   variant.price_modifier_amount_in(currency, options)
 
-      currency = opts.delete(:currency) || order.currency
-
-      self.currency = currency
-      self.price    = variant.price_in(currency).amount +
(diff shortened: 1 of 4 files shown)


gmacdougall:
Looks good to me. Thanks for working on this. :+1:

adammathys:
:+1: Looks good to me.

core/app/models/spree/line_item.rb
```diff
@@ -175,7 +170,9 @@ def update_tax_charge
     end
 
     def ensure_proper_currency
-      unless currency == order.currency
+      if currency != order.currency
+        Spree::Deprecation.warn "You tried setting the currency for this line item manually." \
```
tvdeyen: Maybe not everyone intended to change the line item currency and something else change it. Without knowing that, the current warning could be misleading. 

Why not just: "The line items currency is different from it's order currency. This behavior is not supported anymore and will be deleted soon."?

tvdeyen:
:+1: otherwise

mamhoff:
Thanks @tvdeyen - I took your suggested wording for the deprecation message.

tvdeyen:
:+1:

gmacdougall:
:+1 again!

PR #1011, opened 2016-03-28
Make the creation of the sandbox consistent with the dummy app.

eric1234:
The dummy app [allows a DB_HOST variable](https://github.com/solidusio/solidus/blob/master/core/lib/generators/spree/dummy/templates/rails/database.yml#L30) to be set when creating test apps and the database username is `postgress`. The sandbox app on the other hand only allows the database type to be set. It is always assumed to be `localhost` and the user is assumed to be `sandbox`.

This commit makes the sandbox app consistent by allowing the DB_HOST variable to be set and it uses the same postgres user so that the a new special user does not need to be setup (i.e. we can use the same user as we are using for our tests).

### DIFF
--- a/lib/sandbox.sh
+++ b/lib/sandbox.sh
@@ -32,6 +32,14 @@ group :test, :development do
 end
 RUBY
 
+sed -i '/&default/a \
+  host: localhost' config/database.yml
+sed -i '/&default/a \
+  username: postgres' config/database.yml
+if [ -n "$DB_HOST" ]; then
+  sed -i "s/localhost/$DB_HOST/g" config/database.yml
+fi
+
 bundle install --gemfile Gemfile
 bundle exec rake db:drop db:create
 bundle exec rails g spree:install --auto-accept --user_class=Spree::User --enforce_available_locales=true


jhawthorn:
These values aren't going to work for databases other than postgres.

Is it possible that rails' `DATABASE_URL` would solve your needs here?

eric1234:
Ah yea, forgot to make the addition of `username: postgres` conditional on the DB being pg.

It's possible that `DATABASE_URL` might work. Is there a way I can provide that when creating a sandbox and that is used in the generated app?

jhawthorn:
`DATABASE_URL` won't replace the `database.yml` in the generated app, but will take precedence over the settings in `database.yml` when connecting to a database

eric1234:
I attempted to use `DATABASE_URL` but cannot seem to get it to work with me. First I tried:

`DATABASE_URL=postgres://postgres@127.0.0.1/sandbox_development rake sandbox`

This seemed to be trying to use the connect but since it didn't generate with the `pg` gem it get's an error.

```
Couldn't create database for {"adapter"=>"postgresql", "pool"=>5, "timeout"=>5000, "database"=>"sandbox_development", "username"=>"postgres", "host"=>"127.0.0.1"}
rake aborted!
Gem::LoadError: Specified 'postgresql' for database adapter, but the gem is not loaded. Add `gem 'pg'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).

Gem::LoadError: pg is not part of the bundle. Add it to Gemfile.
```

Next I attempted to also specify the database type in an attempt to get the gem loaded via `DB=postgresql DATABASE_URL=postgres://postgres@127.0.0.1/sandbox_development rake sandbox`. This resulted in the following error:

```
could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

So it seems to be correctly generated with the PG driver but now it's not using my connection string anymore and trying to connect via a domain socket. Any suggestions to getting it working with `DATABASE_URL`?

tvdeyen:

Issue #655, opened 2016-01-05
Solidus admin roadmap


Mandily:
A while ago I posted a git hub issue #487 to let the community know that Stembolt was ramping up efforts to improve the Solidus admin UI. Some effort has been put against planning how to roll these updates out, and we’d like to give the community insight into our roadmap at this time. We’ve already completed some issues for 1.2 and are starting to plan for 1.3 and 1.4. There’s a [post over on the solidus blog](http://solidus.io/blog/2016/01/05/solidus-improvements-roadmap.html) with more details, but a high level summary is included below. 

The high level plan is:
- Fix/adjust structural components first. This includes layout structure and any larger information architecture changes.
- Internal page navigation. Things like creating a sticky header and moving right side navigation into tabs.
- Branding and colour cleanup. While on the surface this seems purely visual, there is some function behind cleaning up interactive elements and decluttering the page to improve the user experience.
- Making the move to complete a task in one page. The current model has the user navigating away to complete edit or add functions. 
- Cleanup inconsistencies in the UI. Making the system look and behave the same across all pages will make it easier to extend in the future. It’ll also be more intuitive to users when they don’t need to remember multiple ways of interacting and label names.
- Improving the search layout. 
- New features.

**Completed in release 1.2**
#487 Improve layout with left navigation and full width
#505 Add flyout triggered by hover to left nav
#533 Solidus styleguide

**Planned for release 1.3**
#666 Allow left navigation to be collapsed
#630 Move page title and action buttons into a sticky header
#633 Add tabs to replace right side nav
#634 Group settings pages and move them into the left nav and tabs

**Planned for release 1.4**
#1076 Add navigable breadcrumbs in the header
#1078 Create 2 layout options for content area of admin UI

**Future release**
#520 Solidus admin branding - this will be worked on after the major structural changes have been completed.

Feel free to comment on the plan in the comments.

isaacfreeman:
Sounds great. I'm particularly keen on "complete a task in one page". Having product admin divided among multiple pages is a real source of confusion for clients.

sahal2080:
I strongly advise the Solidus core team to have a centralized and public place for the roadmap. This way new people can instantly see where this is heading and also feel that this project has higher chances of longevity. It's even better if the community was allowed to add and vote on features so that it can be used as one of the metrics for what to tackle next. 

Something like [Trello Voting Board](https://trello.com/b/HPJrNcQM/features-voting-board) or similar will do.

Mandily:
[...]

PR #1012, opened 2016-03-29
Prepare for new layouts

graygilmore:
This PR sets the stage for the layout work described for the 1.3 release here: https://github.com/solidusio/solidus/issues/655

In order to start work on that first we need to clean up the overall layout. At a high level, this PR reorganizes the admin layout and simplifies what `.main-content` actually contains.

This also pulls in the good work that @Murph33 did in https://github.com/solidusio/solidus/pull/966 which also helps to simplify things.

Other than the header changes from the PR mentioned above, there should be **no** visual change to the admin area.


backend/app/assets/stylesheets/spree/backend/shared/_helpers.scss
```diff
@@ -0,0 +1,35 @@
+.block-table {
```
jhawthorn: `helpers` is a bit of an overloaded term. How about `utilities` or `utils`?

If you believe helpers is the most descriptive, I'm also fine with that.

backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -1,16 +1,12 @@
...
-
 html {
   height: 100%;
+  box-sizing: border-box;
+}
+
+*,
+*:before,
+*:after {
+  box-sizing: inherit;
```
jhawthorn: All of this is provided by bootstrap's `_reboot.scss`. It can probably all be removed.

tvdeyen:
:+1: nice

PR #1015, opened 2016-03-30
Admin::VariantsController#update_positions issue

### LINKED ISSUE #1014: Admin::VariantsController#update_positions issue
1. Bug Scenario

  1) admin url - admin/products/xxx(slug)/variants
  2) try to change variant position in current page
  3) this event request to Admin::VariantsController#update_positions action
  4) throws error model_class.find(id).set_list_position(index)
    - undefined method `set_list_position' for #<Spree::Variant:0x007fc722f20580>

2. Reason
  - trying to call acts_as_list's method without including the required method in model

3. Solution
  - add the required method in model

devhyunjae:
fixes #1014

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -15,6 +15,7 @@ module Spree
   # inventory level determine "on_hand" level for the product.
   class Variant < Spree::Base
     acts_as_paranoid
+    acts_as_list scope: :product
 
     include Spree::DefaultPrice
 


mamhoff:
This makes sense, but would you be so kind to add a meaningful commit message (see http://chris.beams.io/posts/git-commit/ for what I mean by "meaningful").

Also, it'd be great to have a test that checks `Spree::Variant`s new behaviour.

cbrunsdon:
Ugh, so update_positions is inherited from the ResourceController. There is probably a *lot* of broken routes that we expose but don't actually support.

I'm :+1: on this, a test that it acts like a list would be nice, but getting it in to fix our existing broken route would take priority over a test for me.

Thanks @devhyunjae

jhawthorn:
Thank you.

PR #1017, opened 2016-03-30
Add missing timestamps for spree_* tables

mamhoff:
Many models do not have basic timestamps enabled. This commit adds timestamps
everywhere they might be useful. I passed `null: true`, as I expect these models to have historic data that we can't backfill.

### DIFF
--- a/core/db/migrate/20160330204846_add_missing_timestamp_columns.rb
+++ b/core/db/migrate/20160330204846_add_missing_timestamp_columns.rb
@@ -1,0 +1,20 @@
+class AddMissingTimestampColumns < ActiveRecord::Migration
+  def change
+    # Missing updated_at
+    add_column :friendly_id_slugs, :updated_at, :datetime, null: true
+
+    # Missing created_at
+    add_column :spree_countries, :created_at, :datetime, null: true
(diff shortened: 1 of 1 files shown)


tvdeyen:
I'm :+1: on this, **but** this can't happen accidentally. So, I ask myself, why where theses timestamps omitted in the first place? Performance? Thoughts?

gmacdougall:
I think it's reasonable so assume these were omitted by accident. I don't see any reason why they shouldn't be there.

mamhoff:
I think what happened depends on the model: 

- Some omitted it intentionally (`Spree::OrderMutex`)  - those are not included. 
- Some just missed adding them because they were created in ancient times (`Spree::Variant`) 
- Some used to be join tables and have since been upgraded to proper models (`Spree::ProductsTaxons`), and it was omitted by accident. 

For some of the join tables it might not be strictly necessary, but I'd like this schema to be unsurprising, and that's what this PR does. 
Re: The performance question: I think there's bigger performance holes in Solidus than adding timestamps here and there. :)

jhawthorn:
:+1:

(and I don't expect any measurable performance change)

gmacdougall:
:+1:

PR #1019, opened 2016-03-31
Re-organize variant edit form and flow

mamhoff:
In order to add somewhere to edit prices, I need the variant form
to be somewhat more structured than it has been in the past. Using
nested `fieldset` elements with `legend`s, this commit breaks the
edit variant form up in a "general", a "pricing" and a "shipping"
section, each with the relevant input fields.

In order to be consistent with the search form on the index page,
I chose a four-column layout.

All `data-hook`s are still there, although I moved some of them.

I also improved I18n, using the t('.something') helper which looks
translations up according to the name of the partial from which the
`t` is called. This pattern has been (sort of) used in the user and store
credit partials already, and does really good namespacing.

**New search box, and `deleted` checkbox next to it**
solidus administration variants 2016-04-05 18-08-25

**`New Variant` form, now full-width**
solidus administration variants 2016-04-05 18-08-54

**`Edit Variant` form, now full-width, without allowing the user to edit option values** 
solidus administration variants 2016-04-05 18-09-19

**When checking the `Show deleted variants` checkbox, deleted are shown in red**
solidus administration variants 2016-04-05 18-09-40

**Link to product details page if no option values are set for this product**
solidus administration variants 2016-04-05 18-10-49

### DIFF
--- a/backend/app/controllers/spree/admin/variants_controller.rb
+++ b/backend/app/controllers/spree/admin/variants_controller.rb
@@ -32,7 +32,7 @@ module Spree
 
       def collection
         if params[:deleted] == "on"
-          base_variant_scope ||= Variant.only_deleted.where(product_id: parent.id)
+          base_variant_scope ||= super.with_deleted
         else
           base_variant_scope ||= super
         end
--- a/backend/app/controllers/spree/admin/variants_controller.rb
+++ b/backend/app/controllers/spree/admin/variants_controller.rb
@@ -31,12 +31,10 @@ module Spree
       end
 
       def collection
-        @deleted = (params.key?(:deleted) && params[:deleted] == "on") ? "checked" : ""
-
-        if @deleted.blank?
-          base_variant_scope ||= super
-        else
+        if params[:deleted] == "on"
           base_variant_scope ||= Variant.only_deleted.where(product_id: parent.id)
+        else
+          base_variant_scope ||= super
         end
 
         search = Spree::Config.variant_search_class.new(params[:variant_search_term], scope: base_variant_scope)
--- a/backend/app/views/spree/admin/shared/_product_tabs.html.erb
+++ b/backend/app/views/spree/admin/shared/_product_tabs.html.erb
@@ -1,5 +1,5 @@
 <% content_for :page_title do %>
-  <%= Spree.t(:editing_product) %> &ldquo;<%= @product.name %>&rdquo;
+  <%= Spree::Product.model_name.human(count: :other) %> &#x2f; <%= @product.name %>
 <% end %>
 
 <% content_for :tabs do %>
--- a/backend/app/views/spree/layouts/admin.html.erb
+++ b/backend/app/views/spree/layouts/admin.html.erb
@@ -19,7 +19,7 @@
                 <%= yield :tabs %>
               </div>
             <% end %>
-            <div class="<%= if content_for?(:sidebar) then 'twelve' else 'sixteen' end %> columns">
+            <div class="js-content-below-tabs <%= if content_for?(:sidebar) then 'twelve' else 'sixteen' end %> columns">
               <%= render :partial => 'spree/admin/shared/table_filter' %>
               <%= yield %>
             </div>
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -725,7 +725,7 @@ en:
         new:
           new_variant: New variant
         edit:
-          edit_variant: Edit variant
+          edit_variant: "Edit variant - %{options_text}"
         form:
           dimensions: Dimensions
           pricing: Pricing
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -307,6 +307,7 @@ en:
         height: Height
         price: Price
         sku: SKU
+        tax_category: Variant tax category
         weight: Weight
         width: Width
       spree/zone:
@@ -728,6 +729,7 @@ en:
           edit_variant: "Edit variant - %{options_text}"
         form:
           dimensions: Dimensions
+          use_product_tax_category: Use Product Tax Category
           pricing: Pricing
           pricing_hint: These values are populated from the product details page and can be overridden below
     administration: Administration
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -721,6 +721,15 @@ en:
           amount_authorized_exceeds_total_credit: " exceeds the available credit"
           amount_used_not_zero: "is greater than zero. Can not delete store credit"
           update_reason_required: "A reason for the change must be selected"
+      variants:
+        new:
+          new_variant: New variant
+        edit:
+          edit_variant: Edit variant
+        form:
+          dimensions: Dimensions
+          pricing: Pricing
+          pricing_hint: These values are populated from the product details page and can be overridden below
     administration: Administration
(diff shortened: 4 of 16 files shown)


graygilmore:
Nice change. Probably a good candiate for tabs.

mamhoff:
I was actually surprised to see that the product form doesn't have them, yet! I do think you mean the sidebar, not the sections in this form, correct?

mamhoff:
I talked to @Mandily about this IRL, and I'll be updating this with lots of her input.

mamhoff:
Done updating. This has come out pretty nice! I have a few questions for reviewers, which I'll put as comments to individual code lines. Also: Check out the screencasts!

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -139,6 +139,10 @@ span.info {
 .field {
   padding: 10px 0;
 
+  &.centered {
+    margin-top: 13px;
+  }
```
mamhoff: I don't like this, but if I don't do something like that the checkbox that it's applied to ("Track inventory") is way too far up. I'm sure there's a better way to do this. Frontenders?

backend/app/views/spree/layouts/admin.html.erb
```diff
@@ -19,7 +19,7 @@
                 <%= yield :tabs %>
               </div>
             <% end %>
-            <div class="<%= if content_for?(:sidebar) then 'twelve' else 'sixteen' end %> columns">
+            <div id="content-without-nav" class="<%= if content_for?(:sidebar) then 'twelve' else 'sixteen' end %> columns">
```
mamhoff: This is an unfortunate name, but `content` has already been taken by the div above that contains both the `aside` nav and the actual content. If anyone comes up with a better name, I'd be happier.

core/config/locales/en.yml
```diff
@@ -715,6 +716,18 @@ en:
...
+          show_deleted: Show deleted variants
+        new:
+          new_variant: New variant
+        edit:
+          edit_variant: "Edit variant - %{options_text}"
+        form:
+          dimensions: Dimensions
+          no_tax_category_override: No override
+          pricing: Pricing
+          pricing_hint: These values are populated from the product details page and can be overridden below
```
mamhoff: This is what happens if you use the t('.something') convention. It's nice, and very well namespaced. We're losing a bit of generality here, but on the other hand it's very easy to find where to add a translation. Loved it.

backend/app/views/spree/layouts/admin.html.erb
graygilmore: Totally. I have completely reworked this for future layouts because I ran into the same problem: https://github.com/solidusio/solidus/compare/master...graygilmore:content_layout#diff-b1a47ee5eb885f91f6fe5ff9b2be5211R13

Since `#content-without-nav` is only being used for a JS hook maybe just something like `class="js-content-below-tabs`? Either way it's going to be ugly but I'd rather avoid something like `content-without-nav` as it implies that there is no nav on the page (if you were viewing it out of context, like in the `new.js.erb` file).

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
graygilmore: Try adding the `.checkbox` class to the div instead.

mamhoff: Perfect! Thanks loads!

mamhoff:
@graygilmore @Mandily Do have a look, I rebased with the current master - now we have: 

- Full-width table 
- Full-width forms

This is from older PRs:
- Working search in variants
- Tabs :confetti_ball: 
- Consistent view of deleted variants
- Deleted variants don't lose their values anymore

See updated screen shots in the description

core/config/locales/en.yml
tvdeyen: That's nice!

tvdeyen:
@mamhoff the screenshots above are also updated?

mamhoff:
@tvdeyen Yes they should be, does it look old to you?

tvdeyen:
No. Was just asking, because you made changes to the layout according to your last comment.  Thanks

backend/app/controllers/spree/admin/variants_controller.rb
```diff
@@ -31,12 +31,10 @@ def new_before
       end
 
       def collection
-        @deleted = (params.key?(:deleted) && params[:deleted] == "on") ? "checked" : ""
```
tvdeyen: Do people depend on that instance var?

mamhoff: I don't think so. It was only used for filtering, and that's been entirely revamped (we already "lost" `@option_types` and `@option_values`).

Mandily:
It seems to me that "EDIT VARIANT - SIZE: M, COLOR: BLUE" should actually be the title of this screen, instead of the title of the first section. Is there any chance we can add that to the page title? I'm not sure if we need the variant type in the title though, what do you guys think?
(Eventually I'd like to get some breadcrumb functionality in the title instead of having the 'back to products list' button. But that's probably a little bit of scope creep for this thread)

short title
long title

mamhoff:
Yes, @Mandily that's possible. Look!

solidus administration variants 2016-04-05 20-58-49

backend/app/views/spree/admin/shared/_product_tabs.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :page_title do %>
-  <%= Spree.t(:editing_product) %> &ldquo;<%= @product.name %>&rdquo;
+  <%= Spree::Product.model_name.human(count: :other) %> &#x2f; <%= @product.name %>
```
tvdeyen: :smirk:

mamhoff:
Yeah, let's do the breadcrumbing in a different PR. :)

Mandily:
I really like the page title update, that's harder to miss than where it was before. I have 3 more ideas for improving this:
1. I know I had section titles in my wires, but I think it's cleaner, takes up less space and still gets the job done without them. 
2. It would be more consistent with the rest of the app if we could label the 'override tax category" as 'variant tax category', and then have the default option be 'default'
3. Shorten the width of the tax category field

cleaned up

mamhoff:
1. Can we leave the "Edit Variant" in? I like the frame there, aesthetically :) Otherwise I'm fine with killing the sections headers and additional lines. 

2. I don't like calling anything "Default", because default is never actually defined and can mean different things in one or the other context. I could label the "No override option" with the tax category name from the product, but that's not really what we're storing in the database (we're storing nothing in this case). I'm curious for other opinions here. 

3. NP, will do.

mamhoff:
@Mandily here's the form without `legend`s - Still much better than what we had before! There's a larger gap towards the second row because the checkbox needs more space around itself than the other fields, but that's not something I want to tackle here. 

I went with "Use Product Tax Category" instead of "No override", which I think is understandable.

solidus administration variants 2016-04-07 00-49-32

Mandily:
This looks great! The "Use Product Tax Category" text is much clearer and it looks closer to fitting on one screen.  :+1:

jhawthorn:
:+1:

PR #1021, opened 2016-04-01
Remove unused Gateway.current method

jordan-brough:
The 'super' method referenced here doesn't exist.  It was removed in:
https://github.com/solidusio/solidus/commit/df32600


jhawthorn:
:+1: :100:

mamhoff:
:+1:

PR #1022, opened 2016-04-01
Make the reference to Spree::Store explicit in app config.

Senjai:
This appeared to be the source of a code reloading issue when this logic
was called. I've patched it by simply removing the deprecated logic, but
adding the spree namespace here seemed to improve the rails constant
autoloading issues around this particular part of the code.

The error I saw was: ArgumentError: A copy of Spree::AppConfiguration
was removed from the module tree but is still active. We narrowed this
down to the fact rails may have been looking for the constant on
AppConfiguration and using a more qualified name would make this more
robust.


jhawthorn:
This is a workaround more than anything, we should probably move `AppConfiguration` to `lib/`, as it should not be code-reloaded.

That said, this is a good workaround :+1:

mamhoff:
:+1:

PR #1023, opened 2016-04-01
Deprecate User#drop_payment_source & Gateway#disable_customer_profile

jordan-brough:
These are not used within Solidus itself.
Erasing data doesn't seem like the right way to 'disable' something.
`User#drop_payment_source` is not an accurate name for what that
method did.
It uses `update_column` for no reason that I could ascertain.
I did a quick survey of github and could find only one public repo
that is currently using this:
https://github.com/sprangular/sprangular/blob/v0.1.0/app/controllers/sprangular/credit_cards_controller.rb

I was looking at this while looking at what it takes to create a
non-credit-card payment source.

If people feel like it's important to keep this that's fine, I'm just looking
at things that could simplify having non-credit-card payment sources.

### DIFF
--- a/core/app/models/concerns/spree/user_payment_source.rb
+++ b/core/app/models/concerns/spree/user_payment_source.rb
@@ -15,7 +15,7 @@ module Spree
     end
 
     def drop_payment_source(source)
-      ActiveSupport::Deprecation.warn("User#drop_payment_source is deprecated")
+      ActiveSupport::Deprecation.warn("User#drop_payment_source is deprecated", caller)
       gateway = source.payment_method
       gateway.disable_customer_profile(source)
     end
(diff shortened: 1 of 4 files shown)


mamhoff:
Yikes. I do believe the idea is decent: Given that we want to keep a credit card record, but we want to make sure its not actually billed anymore, we destroy the token in our database. 

Really though: I think a payment source should know how to disable itself. I'd prefer a method signature like `payment_source.disable!`.

So, :+1: .

For more reading, I did multiple payment source support in Spree half a year ago. It's not super-pretty, but definitely something that works: https://github.com/spree/spree/pull/6831

jordan-brough:
Re: Your multiple payment source work -- that is extremely helpful, thank you!  I'm going to be looking through that in detail.

peterberkenbosch:
We need that here IMHO! (non creditcard stuff).

jhawthorn:
Works for me :+1:

PR #1025, opened 2016-04-04
Variant: Load default price also when including the relation

mamhoff:
Previously, the admin screen for deleted variants would always show
a price of zero, because the preloaded default_price relation would
not respect the `Spree::Price.unscoped { super }` override.

This commit introduces the `with_deleted` scope into the relation,
fixing the apparnent nilling of the price in the backend.

### DIFF
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -4,7 +4,7 @@ module Spree
 
     included do
       has_one :default_price,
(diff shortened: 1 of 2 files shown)


jhawthorn:
I have done this elsewhere and it has worked well (8090a4dce549fcb22821e6b63da48b2324d6609d). Not sure why I missed this one (oops).

:+1:

adammathys:
:+1:

PR #1026, opened 2016-04-04
Variant: Remove duplicated option value accessor spec

mamhoff:
These specs are still there, just a little above.


jhawthorn:
:+1:

adammathys:
:+1:

PR #1027, opened 2016-04-04
Do not delete option value configuration when destroying variant

mamhoff:
Before this change, a soft-deleted variant would have its
option value variants join table entries hard-deleted. The effect
of this behaviour was that in the backend, your deleted variants
would have no options and would be, therefore, indistinguishable.

The alternative would have been to make `Spree::OptionValuesVariants` also
act as paranoid. However, this leads to many complications. Instead, I just make sure
that all option values join table entries are actually destroyed when
the record is acutally removed from the database.

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -33,7 +33,7 @@ module Spree
     has_many :stock_locations, through: :stock_items
     has_many :stock_movements, through: :stock_items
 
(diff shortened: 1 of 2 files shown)


jhawthorn:
:+1: Makes sense (well, as much as anything when dealing with `acts_as_paranoid` models).

adammathys:
:+1: Definite improvement.

PR #1028, opened 2016-04-04
Use tabs on product edit page

mamhoff:
We have a great new tabs feature to replace the third-level sidebar
navigation. This commit uses the tabs feature on the product edit page.

For the time being, the only point where this messed up the existing layout
was `products#edit`. I solved it by making the first container full-width and
using the new sidebar space for pricing, variant and dimension options.

By the way: This has been amazing easy. Yay @graygilmore !

solidus administration products 2016-04-04 13-30-06

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -1,6 +1,6 @@
 <div data-hook="admin_product_form_fields">
 
-  <div class="left eight columns alpha" data-hook="admin_product_form_left">
+  <div class="left twelve columns alpha" data-hook="admin_product_form_left">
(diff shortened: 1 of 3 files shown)


tvdeyen:
:+1: Looking great! And the minimum amount of work needed is awesome. Big up @graygilmore

graygilmore:
A double mention! :smile:

Change looks good to me. So much better have those along the top.

jhawthorn:
:+1: great

jhawthorn:
Just a note that we might switch that 12:4 column ratio back to a 8:4 when we switch to bootstrap's 12 column grid. For now though, this is the right step forward.

PR #1029, opened 2016-04-04
Use Solidus variant searcher in backend variant index

### LINKED ISSUE #855: Broken product variants search when utilizing option types
When searching for a variant based on option types currently only the last option type sent is sent as a parameter.  After attempting the search the option types other than the last will be set to none.

broken_search

mamhoff:
Before this commit, the variant search was broken - it would
only send query data from the last select box present.

This commit replaces the Ransack search form with the variant search
we're already using in the stock locations tab.

New variant search form: 
solidus administration variants 2016-04-04 19-45-15

Old variant search form (the middle select won't have any effect): 
solidus administration variants 2016-04-04 19-46-35

Fixes #855.


jhawthorn:
:+1: Much simpler, and now actually works!

adammathys:
:+1:

PR #1030, opened 2016-04-04
Upgrade underscore 1.4.4 to lodash.core 4.8.2

jhawthorn:
We were running a fairly old version of underscore (January, 2013)

Lodash is a fork of backbone with better maintenance, 100% test coverage, and many new features.

Here I've included just the core build, which is most similar to what underscore offered and doesn't include the extra lodash features. This weighs in at ~4 kB gzipped vs ~22 kB for the full build.

Currently we barely use the lodash functionality, it might make sense to upgrade to the full build at a later date if this changes.


stewart:
:+1:

jhawthorn:
We're using more of underscore than I expected :disappointed:

PR #1032, opened 2016-04-04
Remove jquery.cookie

jhawthorn:
This was previously used in the old version of the taxon tree. We are no longer using it.


Senjai:
:shipit:

tvdeyen:
:+1: I :heart: removals!

PR #1033, opened 2016-04-04
Combine Order#token_based? and Gateway#supports?

jordan-brough:
This came out of a discussion with @mamhoff around non-credit-card
payment sources.

**This does remove a (private) method from Order and does slightly change
the behavior of the `supports?` method.**
It seems like it might be an OK change to make but extra eyes and thoughts
will be very appreciated.

`Order#token_based?` was a private method and was only used inside
`Order#handle_payment_preconditions` and only in conjunction with
the `supports?` method.  And the `supports?` method itself is only
used in that same spot.  So as best we could tell, `token_based?` was
being used to modify how `supports?` works.  Combining them seems more
straightforward.

The motivation here is to make it easier to support non-credit-card
payment sources by reducing complexity and reducing the number of methods
that a non-credit-card source is required to implement.

### DIFF
--- a/core/app/models/spree/gateway.rb
+++ b/core/app/models/spree/gateway.rb
@@ -39,8 +39,8 @@ module Spree
 
     def supports?(source)
       return true unless provider_class.respond_to? :supports?
-      return false unless source.brand
-      provider_class.supports?(source.brand)
+      return true if source.brand && provider_class.supports?(source.brand)
+      source.has_payment_profile?
     end
 
     def disable_customer_profile(source)
--- a/core/app/models/spree/payment/processing.rb
+++ b/core/app/models/spree/payment/processing.rb
@@ -134,7 +134,7 @@ module Spree
 
         if source
           if !processing?
-            if payment_method.supports?(source) || token_based?
+            if payment_method.supports?(source)
               yield
             else
               invalidate!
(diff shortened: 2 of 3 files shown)


mamhoff:
I think it's a great idea to remove `#token_based?` from payment, but I'm not so sure anymore about combining these two methods into one. 

The case in which `token_based?` would be run is a fairly remote one as we discussed IRL: 
1. Your credit card is not `supported` by your payment method
2. But it has a token
3. And the provider (which we're assuming here to be the same one as the provider of your payment method) knows how to bill using that token

`payment_method.supports?` should, IMHO, not provide the answer to Question 2.
Question 3 is always assumed to be `yes`. Ah well.  

What about this approach instead: 
In `processing.rb`: 
```
if payment_method.supports?(source) || source.has_payment_profile?
```

jordan-brough:
@mamhoff that would definitely be an improvement!  I'd love to go a bit further if I could.

I was hoping to move the whole "but it has a token" thing into PaymentMethod because then payment methods where this isn't an issue ("supporting" amex vs visa vs mastercard, etc) could could simply override `supports?` and not have to deal with this.  It feels a bit awkward that the Order has to know to call both `PaymentMethod#supports?` and then also call `Source#has_payment_profile?` because (if I'm reading things correctly) `supports?` doesn't quite do the full job, doesn't it?

mamhoff:
@jordan-brough But then, the method is still named awkwardly. 

What about deprecating `Spree::PaymentMethod#supports?`, and instead using a new method: 

```
def can_bill?(source)
   supports?(source) || source.has_payment_method?
end
```

mamhoff:
We talked IRL and I am :+1: on this.

jhawthorn:
:+1:

PR #1034, opened 2016-04-04
Remove unnecessary autosave callback

mamhoff:
The default price is autosaved via its `has_one` relation already,
so these 14 lines are not necessary.


jhawthorn:
:+1:

tvdeyen:
:thumpsup: :dart:

PR #1035, opened 2016-04-05
Price Modifier Pricer

mamhoff:
There is an extension point for modifying the price of a line item
that is used in the following extensions:

* https://github.com/godaddy/spree_product_sale
* https://github.com/v-may/spree_simple_sales

This extension point introduces some rather interesting code to
Spree::LineItem, which this commit extracts to a pricer object.

The extension point, as far as I can overview, needs the following to work:

- `*_price_modifier_amount_in` methods defined on the variant
- Fitting `*` methods on `Spree::LineItem`
- The Line Item to accept those methods as params

This PR extracts that code from Spree::LineItem, and deprecates this extension point.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -102,8 +102,10 @@ module Spree
       !sufficient_stock?
     end
 
-    # Sets the options on the line item according to the order's currency or
-    # one passed in.
+    # Sets options on the line item.
+    #
+    # The option can be arbitrary attributes on the LineItem. If no price is given in the options,
+    # this will call the legacy `PriceModifier` line item pricer.
     #
     # @param options [Hash] options for this line item
     def options=(options = {})
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -109,9 +109,10 @@ module Spree
     def options=(options = {})
       return unless options.present?
 
-      self.price = variant.price_in(currency).amount +
-                   variant.price_modifier_amount_in(currency, options)
-
+      # There's no need to call a pricer if we'll set the price directly.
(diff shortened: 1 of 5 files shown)


core/app/models/spree/line_item.rb
```diff
@@ -106,16 +106,19 @@ def insufficient_stock?
...
     #
     # @param options [Hash] options for this line item
     def options=(options = {})
       return unless options.present?
 
-      self.price = variant.price_in(currency).amount +
-                   variant.price_modifier_amount_in(currency, options)
-
+      # There's no need to call a pricer if we'll set the price directly.
+      unless options[:price].is_a?(Numeric)
```
jhawthorn: Is it common to have a price passed in here?

I'd prefer checking just `options.has_key?(:price)`, as both `line_item.price = '12.34'` and `line_item.price = nil` do work. Even if we don't expect options to be used that way, it's harder to reason about if it behaves differently based on the type of `:price` provided

mamhoff: There's specs for exactly that case which I don't want to see raising the deprecation warning for the `price_modifier` system. That was the motivation behind this check. 

Your suggestion to not check the type is good, will amend later today.

mamhoff: Hound complained that `has_key?` is deprecated, and I should use `key?` instead.

gmacdougall:
:+1: This isolates bad functionality to a place where it can be removed. This should be removed in Solidus 2.0.

jhawthorn:
> This isolates bad functionality to a place where it can be removed. This should be removed in Solidus 2.0.

Agree 100%. Isolation also means it could be trivially extracted into a gem if someone still needs it when we land Solidus 2.0.

PR #1036, opened 2016-04-05
Use Spree::DisplayMoney to convert Line Item amounts to Money

mamhoff:
We have a library to convert BigDecimals to Spree::Money objects for
display purposes. This commit uses `Spree::DisplayMoney` to generate those
methods instead of the line item object rolling its own thing.

This led to a few naming changes, for which aliases are provided.
The main method that's aliased in each case is now the one generated by `Spree::DisplayMoney`.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -55,12 +55,6 @@ module Spree
       amount + promo_total
     end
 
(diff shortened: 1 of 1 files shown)


gmacdougall:
:+1:

jhawthorn:
:+1:

PR #1037, opened 2016-04-05
Simplify Spree::Variant#price_in

mamhoff:
The commit changes behaviour. Before, `Spree::Variant#price_in` would
return a price of zero if no currency was given or no price was found in
the given currency. I think that is dangerous: You don't want your products
to randomly be sold for nothing.

This commit changes this behaviour to return the default currency if
no currency is given. It furthermore uses the database's finding capacities
rather than iterating over a potentially large array of prices, speeding things
up.

If no price for the given currency is found, returns nil.

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -214,7 +214,7 @@ module Spree
     # @param currency [String] the desired currency
     # @return [Spree::Price] the price in the desired currency
     def price_in(currency)
-      prices.detect{ |price| price.currency == currency && price.is_default } || Spree::Price.new(variant_id: id, currency: currency)
+      prices.find_by(currency: currency, is_default: true)
     end
 
     # Fetches the price amount in the specified currency.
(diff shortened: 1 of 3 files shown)


gmacdougall:
I might prefer returning nil when no price is available in the currency listed. I can also forsee issues when asking for the price in CAD, and am given a price in USD.

mamhoff:
It does return `nil` when there is no fitting price. It's only if you ask for a price in currency `nil` that you get the default currency.

gmacdougall:
That makes sense. I was confused.

jhawthorn:
Is there a good reason to return the price in the default currency if `nil` is used? The existing behaviour didn't allow that afaict.

I tend to view anywhere we use a "default" (price, locale, or store) as a place that is likely to accidentally introduce incorrect behaviour. My preference would be to require the caller to use `price_in(Spree::Config.currency)` if they indeed want the price in the default currency

mamhoff:
@jhawthorn Yeah, makes sense to me. Amending.

mamhoff:
@jhawthorn amended, green with no unrelated spec failures.

jhawthorn:
Great :+1:

PR #1040, opened 2016-04-05
Deleting an item from the shipments page should only delete it from that shipment

### LINKED ISSUE #1039: Deleting a line item deletes *all* matching line items from all shipments.
## Steps to reproduce:
1) Start with an order with multiple quantities of a line item
2) Split the single shipment into multiple shipments
3) click the trash icon on one shipment's line item

### Expected:
The unrelated shipments should be left alone.

### Actual (Error):
Both shipments become empty.

ogfij9uibn

jhawthorn:
Previously, removing an item from the admin shipments page would remove
it from all shipments.

This is a regression in solidus 1.2

This reverts c6fa6226d3028a2259461e66376c888588c02a5e

Fixes #1039

cc @dholdren @cbrunsdon

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/shipments.js
+++ b/backend/app/assets/javascripts/spree/backend/shipments.js
@@ -102,21 +102,6 @@ adjustShipmentItems = function(shipment_number, variant_id, quantity){
   }
 };
 
-deleteLineItem = function(line_item_id){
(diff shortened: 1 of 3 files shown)


dholdren:
thanks for this :+1:

cbrunsdon:
Yea, I'm good by this too, thanks @jhawthorn  :+1:, and for reporting @dholdren.

PR #1042, opened 2016-04-06
Do not set currency on variants

mamhoff:
Setting a currency on a variant doesn't make sense. This deprecates setting and getting the currency from a variant.


jhawthorn:
:+1:

gmacdougall:
:+1:

PR #1043, opened 2016-04-06
Get CustomerReturn currency from order

mamhoff:
Before this commit, any customer return, even for orders in a non-default currency,
would display their `amount`, `total` and `pre_tax_total` in the default currency.

This commit changes this so that customer returns always use their order's currency.


gmacdougall:
:+1: this seems significantly more correct.

tvdeyen:
:+1: Nicely said @gmacdougall

jhawthorn:
:+1: Anything using `Spree::Config[:currency]` is probably wrong.

PR #1045, opened 2016-04-06
Simplify Spree::Price#money

mamhoff:
This is a `money_method`. Even if `amount` is nil, it will return a
money object with zero cents, so the added `|| 0` doesn't make a huge amount of sense.


gmacdougall:
:+1:

jhawthorn:
:+1:

PR #627, opened 2015-12-29
Requirable factories #2 

mamhoff:
This PR builds on #623 (and includes @Sinetheta s work). I went through all factories, set up tests that don't require `testing_support/factories`, and made sure all factories `build` and `create`. As the factories are IMO code, and a pretty important part of the framework, I suggest to leave the tests in the repo. They can later be used to improve the factories even more. I do understand they seem pretty bogus at times, but having a systematic way of testing factories is a nice thing.

PR #1046, opened 2016-04-07
[WIP] Load factories instead of requiring them

Serabe:
As explained on spree/spree#6534 and rails/spring#412, there is a problem loading factories from an engine when using spring.

### DIFF
--- a/core/lib/spree/testing_support/factories.rb
+++ b/core/lib/spree/testing_support/factories.rb
@@ -1,5 +1,5 @@
 require 'factory_girl'
 
 Dir["#{File.dirname(__FILE__)}/factories/**"].each do |f|
-  require File.expand_path(f)
+  load File.expand_path(f)
 end


mamhoff:
Hm, we just put in a bit of work into making factories individually `require`able - allowing you to `require` just the factories you need for your tests instead of the > 100 factories Spree comes with. That functionality comes with `require` statements  

Maybe there's a better solution using a `config.to_prepare` block? Maybe even inside the `factory_girl_rails` railtie, which only loads `factory_girl`, and then does not run `FactoryGirl.find_definitons`.

mamhoff:
So I was wrong, it does run `FactoryGirl.find_definitions`, but does not add the Solidus factories to the load path. 

If you added a file `(spec/test)/factories/spree_factories.rb` directory with the following content:
```
require 'spree/testing_support/factories'
```
Does that solve your problem, too?

mamhoff:
@Serabe did any of my hints help solve your issue?

Serabe:
Sorry for the late response. I've been digging and making some tests (no pun intended) and this is something that really need to be solved in FactoryGirl / Spring side. Thank you for takin a look at this!

nikushi:
I am on the similar trouble with using factory_bot_rails and spring. By https://github.com/solidusio/solidus/pull/627, testing support factories provided by solidus become to load dependent factories by `require` not `load` internally, such as here https://github.com/solidusio/solidus/blob/master/core/lib/spree/testing_support/factories/order_factory.rb#L3

On the other hand, an app calls `FactoryBot.reload` on its initialization process https://github.com/thoughtbot/factory_bot_rails/blob/master/lib/factory_bot_rails/railtie.rb#L26. Can the app reload the testing support factories correctly? First `.reload` flushes all the factories loaded, then try to load factories again by `FactoryGirl.find_definitions` internally. But factories were already loaded with `require` so some factories would not be loaded again.

I finally apply the patch as follows, not good solution though:

```rb
config/initializers/factory_bot_rails.rb
module FactoryGirl
  def self.reload; end
end
```

In my understanding, calling `require` to load factories is incompatible with `FactoryBot.reload`.

PR #1048, opened 2016-04-07
Fix performance issue on coupon code application for large sets

dholdren:
When a Promotion has a large number of PromotionCodes (thousands),
applying a promotion is very slow / memory intensive.
Also, dumping a CSV has a similar problem.

### DIFF
--- a/backend/app/views/spree/admin/promotion_codes/index.csv.ruby
+++ b/backend/app/views/spree/admin/promotion_codes/index.csv.ruby
@@ -1,6 +1,6 @@
 CSV.generate do |csv|
   csv << ['Code']
-  @promotion.codes.order(:id).each do |code|
-    csv << [code.value]
+  @promotion.codes.order(:id).pluck(:value).each do |value|
+    csv << [value]
   end
 end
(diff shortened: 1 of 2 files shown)


tvdeyen:
:+1: obviously better and faster. Thanks!

backend/app/views/spree/admin/promotion_codes/index.csv.ruby
```diff
@@ -1,6 +1,6 @@
 CSV.generate do |csv|
   csv << ['Code']
-  @promotion.codes.order(:id).each do |code|
-    csv << [code.value]
+  @promotion.codes.order(:id).pluck(:value).find_each do |value|
```
mamhoff: Unfortunately, `find_each` doesn't work here, because `pluck` returns an array. Would `.each` not work, now that we con't create AR objects but simple strings?

dholdren: oh my bad, added that at the last minute. will change back to `.each`

jhawthorn:
:+1:

PR #1050, opened 2016-04-07
Remove bullets from text_list class

Murph33:
This class is only used in one location in the project and we want to remove the bullets from it. As requested by @Mandily 

## With Bullets
with_bullets

## No Bullets
nobullets


Murph33:
Other changes are coming as well.

tvdeyen:
:+1: can we also fix the indentation? Looks misplaced.

Murph33:
Finally got back to this.  Not too much more changed.  Original post updated to new image.

PR #1052, opened 2016-04-07
More shipments.js cleanup

jhawthorn:
This has one behaviour change, the rest is just cleanup.

It's invalid to split an item to the shipment it is already in. Previously, attempting to do so would do (essentially) `alert("undefined")`. Instead, this now only shows valid choices for splitting.

Cleanup involved moving more things into backbone views to stop needing to reach into the DOM in all event handlers. This should make it easier in the future to implement the proposed changes to the shipments page.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/shipments.js
+++ b/backend/app/assets/javascripts/spree/backend/shipments.js
@@ -102,63 +102,6 @@ adjustShipmentItems = function(shipment_number, variant_id, quantity){
   }
 };
 
-completeItemSplit = function(event) {
-  event.preventDefault();
-
-  if($('#item_stock_location').val() === ""){
-    alert('Please select the split destination.');
-    return false;
-  }
-
-  var link = $(this);
-  var stock_item_row = link.closest('tr');
-  var variant_id = stock_item_row.data('variant-id');
-  var quantity = stock_item_row.find('#item_quantity').val();
-
-  var stock_location_id = stock_item_row.find('#item_stock_location').val();
-  var original_shipment_number = link.closest('tbody').data('shipment-number');
-
-  var selected_shipment = stock_item_row.find($('#item_stock_location').select2('data').element);
-  var target_shipment_number = selected_shipment.data('shipment-number');
-  var new_shipment = selected_shipment.data('new-shipment');
-
-  if (stock_location_id != 'new_shipment') {
-    if (new_shipment != undefined) {
-      // TRANSFER TO A NEW LOCATION
-      Spree.ajax({
-        type: "POST",
-        url: Spree.routes.shipments_api + "/transfer_to_location",
-        data: {
-          original_shipment_number: original_shipment_number,
-          variant_id: variant_id,
-          quantity: quantity,
-          stock_location_id: stock_location_id
-        }
-      }).error(function(msg) {
-        alert(msg.responseJSON['message']);
-      }).done(function() {
-        window.Spree.advanceOrder();
-      });
-    } else {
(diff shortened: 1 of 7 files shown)


backend/app/assets/javascripts/spree/backend/shipments.js
```diff
@@ -193,7 +113,7 @@ addVariantFromStockLocation = function(stock_location_id, variant_id, quantity)
       url: Spree.routes.shipments_api,
       data: {
         shipment: {
-          order_id: order_number
+          order_id: window.order_number
```
mamhoff: Is this to make the globalness of `order_number` more explicit?

mamhoff:
My JS skills are limited, but from what I see this simplifies the existing codebase a lot. Very educational! :+1:

backend/app/assets/javascripts/spree/backend/shipments.js
tvdeyen: Yes and to ensure that always the global value is used and never a local one. Globals are bad, but it was global befor this change, so it's ok to make it more explicit here. We should change that on future versions to `Spree.current_order_number`

cbrunsdon:
Yea, same deal as the other big JS refactorings, went through commit-by-commit and it all looks good to me, thanks @jhawthorn :+1:

mtomov:
I would've only converted the code to CoffeeScript for better legibility. 

Otherwise I guess you're not heavily utilizing `Backbone.model`, but I'm not knowledgeable enough to point out exactly what should be there, and what in its View counterpart.

mamhoff:
@mtomov We've had the JS/CoffeeScript discussion in core and are in favour of incrementally moving our coffeescript to JS. There's been a long and insightful discussion in the Discourse (the forum software) community, and they had strong arguments in favour of this: https://meta.discourse.org/t/is-it-better-for-discourse-to-use-javascript-or-coffeescript/3153/17

I think the main points are that JS is incrementally getting better, and that only Rails devs know CoffeeScript. (I'm exaggerating, but you get my point).

mtomov:
Hah, ok. I think I support [this side](https://meta.discourse.org/t/is-it-better-for-discourse-to-use-javascript-or-coffeescript/3153/6) though. 

JS is indeed making progress with ECMA 2015, ES6; Typescript is also a great addition that enables large JS codebases to exist, but what you've decided is not bringing us to any of those.

If you want to follow the "future JS" direction, you'd need to put enablers in place - transpilers or requirement for modern browsers on admin. What you've chosen is putting off developers who don't want to be stuck with old JS just because... 

Have a look at Foundation 6.2 - they've converted all their js into ECMA2015 and backport with transpilers (which CoffeeScript is in fact).

mamhoff:
I think we're not discussing the actual PR here anymore 😉 . If you don't mind, let's move this to our Slack channel or an issue. I still think this PR is great work, and wouldn't want @jhawthorn to re-write it in CoffeeScript. John, do you mind rebasing this when you have a minute?

jhawthorn:
@mtomov 

With the solidus admin we're in a somewhat unique and difficult situation.

I don't think we can easily use a external transpiler (other than by going through sprockets). It's unreasonable to add extra work for users by making them setup a transpiler in order to use the admin. I also don't think we want to precompile and commit the result (this works well for small JS libs, but not so much for an application like solidus).

I really don't want to push coffeescript because generally JS developers don't care for coffeescript, and I would like JS devs to contribute to our JS.

So the current feelings of the core team is to accept contributions in either JS or coffeescript, but to avoid converting between the two.

mtomov:
:+1:  All right. Makes sense. Thanks!

PR #1058, opened 2016-04-11
Make order admin views HTML compilant

rodolfoasantos:
Just a little improvement to make the `<table>` tag more HTML compliant


tvdeyen:
Oh, that's correct. Thanks! :+1:

jhawthorn:
Thank you :+1:

PR #1059, opened 2016-04-12
Replace stock transfer confirmations with standard JS dialogs

jhawthorn:
Everywhere else in our admin we use the standard JS `confirm()` dialog. In the stock transfer section a custom warning banner was used which is awkward and not styled the same as the rest of our admin. The wording was awkward as well

This converts those confirmation dialogs to standard `confirm()` dialogs using rails ujs, removing a decent chunk of code. This also has the benefit of allowing keyboard usage (enter to confirm, escape to cancel) and working well on tablets. 

Before:

After:

cc @Mandily


forkata:
:+1: IMHO the after is more consistent with what we have on other actions throughout the backend. Also the the amount of code this removes is great!

mamhoff:
:+1:

Mandily:
This looks great. It's more clear and intuitive for users. :+1: for reducing cognitive load.

PR #1060, opened 2016-04-12
Port current_currency helper module from spree_multi_domain

mamhoff:
Previously, the `current_currency` helper would always return the
configured default currency for the Solidus installation. Solidus core
allows setting a default currency per `Spree::Store`, and this commit
adds a `current_currency` helper that respects that setting.

This does purposefully not include setting a currency via the session
(like the `spree_multi_currency` gem does. Solidus core does not allow
this, and it is much less complex to switch currencies with the configured
store.

This PR will also remove duplicate helper definitions from frontend and core.


core/lib/spree/core/controller_helpers/pricing.rb
```diff
@@ -0,0 +1,17 @@
...
+    module ControllerHelpers
+      module Pricing
+        extend ActiveSupport::Concern
+
+        included do
+          helper_method :current_currency
+        end
+
+        def current_currency
+          current_store.try(:default_currency).presence || Spree::Config[:currency]
```
gmacdougall: Do we expect current the current_store default currency to return a non-nil blank value (like `""`)? If not, we can simplify this to:

```Ruby
current_store.try!(:default_currency) || Spree::Config[:currency]
```

gmacdougall:
This is looking good so far.

core/lib/spree/core/controller_helpers/pricing.rb
mamhoff: I took some specs from `solidus_multi_domain`, and this case has been explicitly tested for: https://github.com/solidusio/solidus/pull/1060/files#diff-14ee7de4c10a42a47306354bec6b902eR20

mamhoff:
I added the `try!` you suggested; otherwise I think this can go in as-is (unless you insist on removing the `presence` call).

gmacdougall:
:+1:

jhawthorn:
:+1:

I think we'll want a CHANGELOG entry as this is a (beneficial) behaviour change that we now respect the `default_currency` option on a store

mamhoff:
@jhawthorn Added Changelog :)

PR #1061, opened 2016-04-12
Fix case where `track_inventory_levels` global flag not being honored

eric1234:
[At the instance level](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/variant.rb#L307-L309) the system looks both at the variant flag as well as the global flag, but when operating [at the collection level](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/variant.rb#L68-L74) only the variant flag was taken into account. This commit fixes this by making the `in_stock` scope a no-op if Solidus is configured to not track inventory levels globally.

I hit this issue when trying to create orders in the backend while configured to not track inventory. The product selector makes a call to the API and [indicates the `in_stock_only` flag](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/orders/edit.js#L4). Since none of the products had inventory (since inventory is not tracked) no products where returned.

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -66,6 +66,7 @@ module Spree
     # @param stock_locations [Array<Spree::StockLocation>] the stock locations to check
     # @return [ActiveRecord::Relation]
     def self.in_stock(stock_locations = nil)
+      return all unless Spree::Config.track_inventory_levels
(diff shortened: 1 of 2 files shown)


jhawthorn:
Looks good to me. Thank you :+1:

cbrunsdon:
Good by me too, thanks as always @eric1234

PR #1062, opened 2016-04-13
Simplify buggy tax rate migration

mamhoff:
This migration did not work, and requires a call to `TaxRate#compute_amount`.

We could have fixed the "the does not work", but the `compute_amount` call requires
half of Solidus' object graph (including the `Spree::Calculator` STI table as well
as `Spree::Order`, `Spree::Shipment`, `Spree::TaxRate` - this is still not exhaustive).

We felt adding all of those into the migration was less than ideal, especially if your
local shipping rate was taxed using a different system than the default tax calculator.

The downside of this is ONLY that in the backend for old shipping orders, you will not be
shown the shipping rate tax (included or additional).

The reduced-in-complexity migration now just adds the `tax_rate_id` column on `Spree::ShippingRate`.

### DIFF
--- a/core/db/migrate/20160225152313_remove_tax_rate_from_shipping_rate.rb
+++ b/core/db/migrate/20160225152313_remove_tax_rate_from_shipping_rate.rb
@@ -1,40 +1,9 @@
 class RemoveTaxRateFromShippingRate < ActiveRecord::Migration
-  class Spree::ShippingRate < Spree::Base; end
-  class Spree::TaxRate < Spree::Base
-    has_one :calculator, class_name: "Spree::Calculator", as: :calculable, inverse_of: :calculable, dependent: :destroy, autosave: true
-    def compute_amount(item)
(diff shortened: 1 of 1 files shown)


jhawthorn:
:+1:

tvdeyen:
I'm not happy with this just being removed and leaving the stores somehow "broken". Even within the change log entry are no hints on how to get the old state back, besides "updating the shipping rates". 

Why don't we take the code and move it into a rake task "solidus:recalculate_shipment_taxes"?

I totally agree that a migration isn't the right place, but leaving this "unfixed" isn't an option for me either. Or am I too strict here?

/c @cbrunsdon

mamhoff:
@tvdeyen Have a look at the upgrade strategy detailed in #1068.

tvdeyen:
:+1:

cbrunsdon:
Aye, good plan here, thanks folks :+1:

PR #1063, opened 2016-04-13
Include ControllerHelpers::Pricing from ControllerHelpers::Order

jhawthorn:
`ControllerHelper::Order` requires `current_currency` from the `ControllerHelper::Pricing` module in order to work.

cc @mamhoff

### DIFF
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -1,8 +1,11 @@
+require 'spree/core/controller_helpers/pricing'
+
 module Spree
   module Core
     module ControllerHelpers
       module Order
         extend ActiveSupport::Concern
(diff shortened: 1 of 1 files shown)


mamhoff:
Alright, no issue with this at all - I thought I'd `require`d the pricing module everywhere the order helpers module is included, but it does make sense to have the module require its dependency itself. Can you give a hint at where this failed?

jhawthorn:
It failed in an extension which only included `ControllerHelpers::Order`. https://github.com/solidusio-contrib/solidus_social/blob/master/app/controllers/spree/omniauth_callbacks_controller.rb

It could be an issue in custom controllers in apps as well.

mamhoff:
OK, no more questions asked :) :+1: !

PR #1065, opened 2016-04-13
Fix changelog entry after removing shipping rate tax migration

mamhoff:
We removed code from a migration in #1062 - this commit explains the
(minimal) consequences of that removal in the changelog.

cc @jhawthorn


jhawthorn:
Thanks

PR #1067, opened 2016-04-13
Rename "Store" to "General" and fix extension compatibility.

jhawthorn:
Previously, extensions often added their own configuration section by adding an item to the `configuration_menu` partial through deface. This was removed, breaking all those extensions.

This restores compatibility with those extensions by moving the "General" tabs to that exiting file. Any old overrides should now work without modifications.

This also renames "Store" to "General" (for which @Mandily is :+1:)

See, for example, repaired compatibility with solidus_social:


cbrunsdon:
Yea, I'm good by this, thanks. :+1:

tvdeyen:
Thanks!

PR #1068, opened 2016-04-14
Upgrade path for shipping rate tax data

mamhoff:
Previously, we had a migration that deleted the `tax_rate_id` column
from the `spree_shipping_rates` table. Within that migration, we could not
satisfactorily create shipping rate taxes because that required a call to
`Spree::TaxRate#compute_amount`, for which we can not guarantee the interface
for all future Solidus versions to come.

This compromise works as follows: We keep the column and provide a rake task to
migrate that data up. In a future release (maybe 2.0), we can then add a migration
to delete the column that errors out if there's still data on it (indicating the
migration hasn't been run).

### DIFF
--- a/core/lib/tasks/migrations/migrate_shipping_rate_taxes.rake
+++ b/core/lib/tasks/migrations/migrate_shipping_rate_taxes.rake
@@ -1,0 +1,17 @@
+namespace :solidus do
+  namespace :migrations do
+    namespace :migrate_shipping_rate_taxes do
+      task up: :environment do
+        puts "Adding persisted tax notes to historic shipping rates"
+        Spree::ShippingRate.where.not(tax_rate_id: nil).find_each do |shipping_rate|
+          tax_rate = Spree::TaxRate.unscoped.find(shipping_rate.tax_rate_id)
+          shipping_rate.taxes.find_or_create_by!(
+            tax_rate: tax_rate,
+            amount: tax_rate.compute_amount(shipping_rate)
+          )
+        end
+        Spree::ShippingRate.where.not(tax_rate_id: nil).update_all(tax_rate: nil)
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,11 +20,14 @@
     The shipping rate taxer class can be exchanged for a custom estimator class
     using the new Spree::Appconfiguration.shipping_rate_taxer_class preference.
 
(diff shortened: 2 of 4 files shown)


tvdeyen:
Perfect! Thank you very much @mamhoff

cbrunsdon:
I'm very much happy with this compromise :+1:

jhawthorn:
:+1:

core/lib/tasks/migrations/migrate_shipping_rate_taxes.rake
```diff
@@ -0,0 +1,17 @@
+namespace :solidus do
+  namespace :migrations do
+    namespace :migrate_shipping_rate_taxes do
+      task up: :environment do
+        puts "Adding persisted tax notes to historic shipping rates"
+        Spree::ShippingRate.where.not(tax_rate_id: nil).find_each do |shipping_rate|
+          tax_rate = Spree::TaxRate.unscoped.find_by(id: shipping_rate.tax_rate_id)
```
gmacdougall: Any reason just to not use `find`?

mamhoff: No particular reason. Should I amend?

core/lib/tasks/migrations/migrate_shipping_rate_taxes.rake
```diff
@@ -0,0 +1,17 @@
...
+          shipping_rate.taxes.find_or_create_by!(
+            tax_rate: tax_rate,
+            amount: tax_rate.compute_amount(shipping_rate)
+          )
+        end
+        Spree::ShippingRate.where.not(tax_rate_id: nil).update_all(tax_rate: nil)
```
gmacdougall: This seems odd to have inside the loop. Should this just be updating `shipping_rate` or should this be outside the loop?

mamhoff: I think it _is_ outside the loop :)

mamhoff:
Amended to use `find` instead of `find_by`.

core/lib/tasks/migrations/migrate_shipping_rate_taxes.rake
gmacdougall: Sorry. Must have been looking at my monitor cross eyed...

gmacdougall:
:+1:

PR #1069, opened 2016-04-14
[readme] Marked gem names as inline code and linked them to rubygems.

apepper:


README.md
```diff
@@ -15,19 +15,20 @@ is a fork of Spree.
...
-solidus gem you automatically require all of the necessary gem dependencies which are:
+[`solidus`](https://rubygems.org/gems/solidus) gem you automatically require all
+of the necessary gem dependencies which are:
 
-* solidus\_api (RESTful API)
-* solidus\_frontend (Cart and storefront)
-* solidus\_backend (Admin area)
-* solidus\_core (Essential models, mailers, and classes)
-* solidus\_sample (Sample data)
+* [`solidus_api`](https://rubygems.org/gems/solidus_api) (RESTful API)
```
tvdeyen: I would prefer GitHub Links, so the interested one could read the source code/README right along.

apepper: AFAIKS only `solidus_frontend` provides a `README.md`. All other gems do't provide a README file.

That's why I still think, that linking to rubygems is the better way.

jhawthorn: I'm not sure the rubygems page provides useful information. Maybe we should use github links with the intention of adding READMEs to the subprojects

apepper:
REVIEW (6f3189b)

* Pointed RubyGems links to github.
* Made localhost links clickable.
* Fixed a small syntax error.

mamhoff:
I'm alright with this, even though we don't have the README files for the individual Gems yet.

cbrunsdon:
I'm fine with this too, if it helps anyone read or navigate its good by me.

If you can squash the 4th typo fix into where it happened, and squash the first two together, I'll be :+1:

jhawthorn:
:+1: Thank you

PR #1071, opened 2016-04-15
Current price options and variant pricer

mamhoff:
First off: I'm deeply sorry for the humoungousness of this PR. 

This will allow us to use different things than just the currency for selecting prices. In order to do that, we add a `PricingOptions` module and a corresponding `Pricer` which will take the desired attributes from the pricing options module and yield a price. 

Since `current_currency` has also been used in availability and search scopes, those scopes also had to be touched (and tested). 

There were a couple of helpers doing really strange stuff in general. I did clean those up a little, because without that this feature wouldn't have worked either. 

At the end of this, I deprecate the  use of `current_currency` because there should be only one source of truth. 

I restructured the PR a bit to a) exclude `Spree::Money` being able to add and subtract and b) include getting rid of the line item pricers, which serve no perceptible purpose now that variant pricers exist.


core/app/models/spree/variant/pricer.rb
```diff
@@ -0,0 +1,19 @@
...
+      end
+
+      attr_reader :variant
+
+      def initialize(variant)
+        @variant = variant
+      end
+
+      def price_for(price_options)
+        variant.prices.currently_valid.find_by(price_options.desired_attributes).try(:money)
```
jhawthorn: `try` should be `try!`

core/app/models/spree/variant.rb
```diff
@@ -213,12 +230,39 @@ def option_value(opt_name)
...
+    def pricer
+      @pricer ||= Spree::Config.variant_pricer_class.new(self)
+    end
+
+    # Chooses an appropriate price for the given pricing options
+    #
+    # @see Spree::Variant::Pricer#price_for
+    # @param [Spree::Config.pricing_options_class] An instance of pricing options
+    # @return [Spree::Money] The chosen price as a Money object
+    delegate :price_for, to: :pricer
```
jhawthorn: Neat!

My only concern here is that `price_for` reads a lot like `price_in`, despite being very different (`price_in` returns a `Price`, `price_for` returns a `Money`). I'm not sure I have a better suggestion, though, as out `*amount*` methods usually just return a number.

core/app/models/spree/variant.rb
```diff
@@ -87,6 +87,23 @@ def self.active(currency = nil)
       joins(:prices).where(deleted_at: nil).where('spree_prices.currency' => currency || Spree::Config[:currency]).where('spree_prices.amount IS NOT NULL')
     end
 
+    # Returns default pricing options that can easily be used in scopes and instance methods
+    # This method is need in scopes and on instances.
+    #
+    # @return [Spree::Variant::Pricer::PricingOptions] the default pricing options
+    def self.default_pricing_options
+      @default_pricing_options ||= Spree::Config.pricing_options_class.new
```
jhawthorn: Would this make more sense on `Spree::Config`. Having class instance variables behaves strangely in development mode as classes are regularly reloaded.

jhawthorn:
This is scary because it's a big change, but it seems very good to me.

core/app/models/spree/variant.rb
mamhoff: I used `money_price` on the line item, would that be an alternative? `Spree::Variant#money_price_for_options`?

mamhoff:
Rebased on master so that the changeset becomes a little slimmer. A little.

core/app/models/spree/product.rb
```diff
@@ -162,10 +162,12 @@ def available?
...
     #
     # @param opt_type [String] the name of the option type to group by
+    # @param pricing_options [Spree::Config.pricing_options_class] the pricing options to search
+    #   for, default: the default pricing options
     # @return [Hash] option_type as keys, array of variants as values.
-    def categorise_variants_from_option(opt_type)
+    def categorise_variants_from_option(opt_type, pricing_options = Spree::Config.default_pricing_options)
       return {} unless option_types.include?(opt_type)
-      variants.active.group_by { |v| v.option_values.detect { |o| o.option_type == opt_type } }
+      variants.with_prices(pricing_options).group_by { |v| v.option_values.detect { |o| o.option_type == opt_type } }
```
mamhoff: Target for deprecation as it's called nowhere in the Solidus codebase.

core/app/models/spree/product/scopes.rb
```diff
@@ -177,8 +177,8 @@ def self.available(available_on = nil, _currency = nil)
     end
     search_scopes << :available
 
-    def self.active(currency = nil)
-      not_deleted.available(nil, currency)
+    def self.active(_currency = nil)
```
mamhoff: Deprecate this for a method that's not called `active`.

core/app/models/spree/variant.rb
```diff
@@ -214,12 +222,39 @@ def option_value(opt_name)
...
+
+    # Chooses an appropriate price for the given pricing options
+    #
+    # @see Spree::Variant::Pricer#price_for
+    # @param [Spree::Config.pricing_options_class] An instance of pricing options
+    # @return [Spree::Money] The chosen price as a Money object
+    delegate :price_for, to: :pricer
+
+    # Returns the difference in price from the master variant
+    def price_difference_from_master(pricing_options = Spreee::Config.default_pricing_options)
```
mamhoff: `Spreee` does not work.

core/app/models/spree/variant.rb
```diff
@@ -214,12 +222,39 @@ def option_value(opt_name)
...
+      master_price = product.master.price_for(pricing_options)
+      variant_price = price_for(pricing_options)
+      return unless master_price && variant_price
+      variant_price - master_price
+    end
+
+    def price_same_as_master?(pricing_options = Spree::Config.default_pricing_options)
+      price_difference_from_master(pricing_options) == Spree::Money.new(0, currency: pricing_options.desired_attributes[:currency])
```
mamhoff: price_difference_from_master(pricing_options).zero? if defined on `Spree::Money`

core/app/models/spree/variant/gift_wrap_pricer.rb
```diff
@@ -0,0 +1,16 @@
+module Spree
+  class Variant
+    class GiftWrapPricer < Spree::Variant::Pricer
```
mamhoff: Deprecate these classes

core/lib/spree/core/search/base.rb
```diff
@@ -4,20 +4,30 @@ module Search
...
 
+        def current_currency=(currency)
+          self.pricing_options = Spree::Config.pricing_options_class.new(
+            pricing_options.desired_attributes.merge(currency: currency)
+          )
+        end
+
+        def current_currency
+          pricing_options.currency
+        end
```
mamhoff: Deprecate these two methods

core/spec/helpers/products_helper_spec.rb
```diff
@@ -58,6 +48,11 @@ module Spree
         let(:product_price) { 100 }
         let(:currency) { 'JPY' }
 
+        before do
+          product; variant
```
mamhoff: No need to call product as variant will call product.

core/spec/helpers/products_helper_spec.rb
```diff
@@ -73,47 +68,46 @@ module Spree
     end
 
     context "#variant_price_full" do
+      let(:variant_2) { create(:variant, product: product, price: variant_2_price) }
```
mamhoff: Use `let!`

core/spec/helpers/products_helper_spec.rb
```diff
@@ -73,47 +68,46 @@ module Spree
...
+      let(:variant_2_price) { 20 }
+      let(:variant_price) { 15 }
+
       before do
         Spree::Config[:show_variant_full_price] = true
-        @variant1 = create(:variant, product: product)
-        @variant2 = create(:variant, product: product)
+        variant; variant_2
```
mamhoff: Only call `variant` as `variant2` is `let!`d.

core/spec/helpers/products_helper_spec.rb
```diff
@@ -73,47 +68,46 @@ module Spree
...
+        let(:variant_price) { 150 }
 
         before do
-          product.variants.active.each do |variant|
-            variant.prices.each do |price|
-              price.currency = currency
-              price.save!
-            end
-          end
+          product; variant
```
mamhoff: Leave out the call to `product`.

core/spec/models/spree/app_configuration_spec.rb
```diff
@@ -20,6 +20,14 @@
     expect(prefs.variant_search_class).to eq Spree::Core::Search::Variant
   end
 
+  it "uses variant pricer class by default" do
+    expect(prefs.variant_pricer_class).to eq Spree::Variant::Pricer
+  end
+
+  it "has a getter for the pricing options class provided by the variant pricer class" do
+    expect(prefs.pricing_options_class).to eq Spree::Variant::Pricer.pricing_options_class
+  end
```
mamhoff: Add a spec for `default_pricing_options`

core/spec/models/spree/line_item_spec.rb
```diff
@@ -217,12 +217,19 @@ def copy_price
...
+      end
+
+      it "updates the price based on the options provided" do
+        expect(Spree::Deprecation).to receive(:warn)
+        expect(line_item).to receive(:gift_wrap=).with(true)
+        expect(line_item).to receive(:gift_wrap).and_return(true)
+        expect(line_item.variant).to receive(:gift_wrap_price_modifier_amount_in).with("USD", true).and_return 1.99
+        line_item.options = { gift_wrap: true }
+        expect(line_item.price).to eq 21.98
+      end
```
mamhoff: This tests the pricers. Please delete.

core/spec/models/spree/line_item_spec.rb
```diff
@@ -236,4 +243,16 @@ def copy_price
       expect(line_item.currency).to eq(new_price.currency.iso_code)
     end
   end
+
+  describe "#pricing_options" do
+    let(:line_item) { stub_model(Spree::LineItem, currency: "RUB") }
```
mamhoff: Use `Spree::LineItem.new()` instead of `stub_model`.

core/spec/models/spree/variant/pricer_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  describe "#price_for(options)" do
+    let(:variant) { create(:variant) }
+
+    context "with the default currency" do
+      let(:pricing_options) { described_class.pricing_options_class.new(currency: "USD") }
+
+      it "returns the correct (default) price as a Spree::Money object" do
+        expect(subject.price_for(pricing_options)).to eq(variant.default_price.money)
```
mamhoff: Make the expectation clearly the thing we expect.

core/spec/models/spree/variant/pricing_options_spec.rb
```diff
@@ -0,0 +1,99 @@
+require 'spec_helper'
+
+describe Spree::Variant::PricingOptions do
+  subject { described_class.new }
+
+  context '.default_price_attributes' do
+    it 'can be passed into a WHERE clause on Spree::Prices' do
+      expect do
+        Spree::Price.where(described_class.default_price_attributes).to_a
```
mamhoff: Expect an empty array instead, and test whether it returns stuff if we setup stuff for it.

core/spec/models/spree/variant/pricing_options_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      end.not_to raise_error
+    end
+  end
+
+  context ".from_line_item" do
+    let(:line_item) { build_stubbed(:line_item) }
+    subject { described_class.from_line_item(line_item) }
+
+    it "returns the order's currency" do
+      expect(subject.desired_attributes[:currency]).to eq(line_item.order.currency)
```
mamhoff: Expect `"USD"` instead of `line_item.order.currency`.

core/spec/models/spree/variant/pricing_options_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    context "if line item has no order" do
+      before do
+        expect(line_item).to receive(:order).and_return(nil)
+      end
+
+      it "returns the line item's currency" do
+        expect(subject.desired_attributes[:currency]).to eq(line_item.currency)
```
mamhoff: Same

core/spec/models/spree/variant/pricing_options_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      end
+    end
+
+    context "if line item's order has no currency" do
+      before do
+        expect(line_item.order).to receive(:currency).and_return(nil)
+      end
+
+      it "returns the line item's currency" do
+        expect(subject.desired_attributes[:currency]).to eq(line_item.currency)
```
mamhoff: Same

core/spec/models/spree/variant/pricing_options_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      it "returns a Hash with the correct currency" do
+        expect(subject.desired_attributes[:currency]).to eq("EUR")
+      end
+    end
+  end
+
+  describe "#currency" do
+    context "when initialized with no currency" do
+      it "returns the default currency" do
+        expect(subject.currency).to eq(Spree::Config.currency)
```
mamhoff: Split this up into two explicit expecatations.

core/spec/models/spree/variant_spec.rb
```diff
@@ -182,6 +182,113 @@
...
+    it "returns an instance of a pricer" do
+      expect(variant.pricer).to be_a(Spree::Config.variant_pricer_class)
+    end
+
+    it "is instacached" do
+      expect(variant.pricer.object_id).to eq(variant.pricer.object_id)
+    end
+  end
+
+  context "#price_for(price_options)" do
```
mamhoff: Use https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/independent/delegate_method_matcher.rb#L49

mamhoff:
@jhawthorn This is ready for a second round of review now. Bad stuff extracted and slimmed to (I think) the minimum.

core/app/models/spree/variant.rb
```diff
@@ -214,12 +222,40 @@ def option_value(opt_name)
...
+      variant_price = price_for(pricing_options)
+      return unless master_price && variant_price
+      variant_price - master_price
+    end
+
+    def price_same_as_master?(pricing_options = Spree::Config.default_pricing_options)
+      diff = price_difference_from_master(pricing_options)
+      diff && diff.zero?
+    end
+
```
mtomov: If one was to anyhow customize `Spree::Config.default_pricing_options`, why would they want to call either of the 2 methods with a different argument?

I think the arguments are redundant and leak knowledge into the whatever will call them.

mtomov:
Thanks for your work on this. The new `solidusio-contrib/solidus_price_modifier` has really nice explanations of its functionality, but I'm a bit into the unknown as to how to use the newly provided functionality here. My two questions:

1) Can you give me a use case when we want to cache the variant's price on anything else, but the currency? Could it maybe be a special user role - _supplier_ So, being a _supplier_ would give prices, which are say 20% lower. Is that now possible?

2) In the `solidus_price_modifier` extension, you mentioned the example of a gift wrap, or any other option on line item affecting the price of the line item. Given that the old approach is mostly deprecated, is it possible to do it now with those new classes provided? If so, can you give an example.

I think some concrete examples would help other developers to also understand how to use them. Adding them to the code as comments above the classes would also be favourable in my opinion, but not sure if that is acceptable.

--

Thank you again. I've reviewed the code, and all looks good. My only comment would be to maybe keep those frontend view related methods for the price difference away from the variant model.

mamhoff:
Hi @mtomov, thanks for reviewing this - it's not an easy piece of code. 

This functionality is meant to be extended, and `solidus_price_modifier` is one (not great) example on how to use it. It's meant to replace functionality like the one provided by extensions like `spree_price_books` or `spree_price_lists` without adding another ActiveRecord table inbetween variants and prices. 

One use case that I have in mind will be pricing depending on the customer's `tax_address`, in order to get VAT pricing right and reduce complexity in the taxation system. The way that will be implemented is by adding a `vat_country_iso` field to prices and select prices based on that field (with a default fallback). 

Another example of where this would come in useful is exactly what you're describing: Different prices for e.g. wholesale buyers or registered users vs. non-registered users. All you need is a different `pricer` with the related `pricing_options` class, and probably another column on the `spree_prices` table. 

I agree that the `price_difference` stuff in `Spree::Variant` is not super-nice, and it's a good candidate for extraction, actually. For now I wanted to keep exisiting functionality though.  

Regarding comments: Yep, that's right, I'll add those. We're big fans of YARD docs.

core/app/models/spree/variant.rb
mamhoff: Once a user has, for example, a different tax address, these methods will need that user's or session's pricing options passed in here. I think that passing the argument is necessary (otherwise the user would get the price difference in the wrong currency, which would be wrong). I am, however, thinking it might be a good idea not to have a default argument and error out when these methods are called without arguments.

mtomov:
Thanks for explaining :+1:

jhawthorn:
Thank you @mamhoff this is excellent.

Also thank you @mtomov for the review. Lets try to get some good class-level documentation on these classes before the proper 1.3.0 release

alepore:
i'm late here but 💯 thanks for this, this is a big improvement and a very much needed one! 

i can probably build a clean and lightweight version of `spree_price_lists` on top of this...

PR #1072, opened 2016-04-15
Make tab dropdowns always visible in test env

jhawthorn:
Capybara+poltergeist can only click links which are visible and clickable. Our tab component collapses tabs into a dropdown when the screen is narrow. This commit forces that dropdown to be open in the
test environment and also changes it from an absolutely positioned element to a static elemnt so that it doesn't cover other on-screen items.

This fixes some failures in extensions, for example https://travis-ci.org/solidusio-contrib/solidus_related_products/jobs/119700687


tvdeyen:
:+1: looks good

mamhoff:
:+1:

mamhoff:
Thanks

PR #1075, opened 2016-04-18
Allow subtraction and addition for Spree::Money objects

mamhoff:
Spree::Money, being a thin wrapper around the Ruby `::Money` object, can
not add or subtract. This commit adds subtraction and addition capabilities.

### DIFF
--- a/core/lib/spree/money.rb
+++ b/core/lib/spree/money.rb
@@ -70,14 +70,17 @@ module Spree
     #
     # @see http://www.rubydoc.info/gems/money/Money/Arithmetic#%3D%3D-instance_method
     def ==(other)
+      raise TypeError, "Can't compare #{other.class} to Spree::Money" if !other.respond_to?(:money)
       @money == other.money
     end
 
     def -(other)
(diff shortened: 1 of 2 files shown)


gmacdougall:
I think this is very useful. My only question is if we should raise a specific error if `other` does not respond to `.money`. What do you think?

mamhoff:
This is a minimal implementation; if we were to do it really seriously, I would also like it if we could add `::Money` objects to `Spree::Money` objects. Ruby `::Money` has all that solved in a nice way, I kind of think it would be a duplication of effort to first implement a lot of `respond_to` and `is_a?` metamagic only to in a future life just let `Spree::Money` inherit from `::Money`... 

On the other hand, I have been confused in the past by `undefined method 'money' for nil:nilClass` :D - I'll add the error.

Senjai:
I feel like arithmetic on money is a good idea for ecommerce :trollface: :+1:

gmacdougall:
:+1: Thanks Martin!

jhawthorn:
:+1: Thanks

We might consider adding more of these from [Money::Arithmetic](http://www.rubydoc.info/gems/money/Money/Arithmetic). This is a great start.

PR #1081, opened 2016-04-19
Show an error when specifying an invalid DB value

jordan-brough:
Just to save myself from typing `postgresq` and ending up with `sqlite`.


jhawthorn:
:+1:

tvdeyen:
👍

Issue #894, opened 2016-02-23
Recalculate Tax amount after Zone change


dangerdogz:
Issue :

1) Client complete an order in TAX ZONE 1
2) Client made a mistake with his address and the shipping billing is really in TAX ZONE 2, so you edit his address in the admin panel
3) Tax amounts (on line items) are not updated (still TAX ZONE 1) unless you manually update each line items (by adding quantity for example)

I'm not sure what is the best way to fix it, it's possible to use update_totals or persist_total in the CustomerDetailsController but it seems out of place / heavy-handed

dangerdogz:
Gemfile :

https://gist.github.com/dangerdogz/996e7b887f2e35d40826

Gemfile.lock

https://gist.github.com/dangerdogz/06fa6c40c1e829a19205

AdnanTheExcellent:
When the address is updated, the taxrate should be adjusted if the @order.tax_zone changes.

pseudo code:

'''
old_tax_zone = @order.tax_zone

CODE_TO_UPDATE_SHIP_ADDRESS_HERE

unless old_tax_zone == @order.tax_zone
   Spree::TaxRate.adjust(@order.tax_zone, @order.line_items)
end
'''

dangerdogz:
Fixed by #1082

PR #1082, opened 2016-04-19
Recalculate the tax adjustments after customer address is updated

dangerdogz:
This is a fix for issue https://github.com/solidusio/solidus/issues/894

If you update an address in the order after the checkout (like by updating the Customer Details in the admin after an order is complete but before it's shipped), this can potentially put the Order in a new Tax zone and this is currently not supported.

Added by this pull request :

**Spree::Address**

A custom method to select only the fields related to Tax Zone matching (zipcode, country_id, state_id) to compare different addresses

**Spree::OrderContent**

Added a check in update_cart to recreate Tax adjustments in certain conditions

1) The order currently has no address (only an empty TaxLocation), we add one. Nothing to be done, it will be handled by the state transitions between :address and :delivery, and :delivery to :payment.

2) The order currently has an address, we update it but editing only trivial fields (firstname, lastname). We use address.tax_zone_attributes to compare the old and the new address and determine that nothing meaningful happened regarding tax zones, we do nothing.

3) The order currently has an address, we update the state / the country / the zipcode. We compute new Tax adjustments because we might be in a different zone.

4) old_tax_address will be  TaxLocation instead of an Address if there was no Address prior to the update, so we just return because it means that the order is either a VAT order and we don't need to recalculate Tax adjustments for now, or the order is a non-VAT order in the cart state, in which case we don't want to / we can't calculate Tax yet

**Potential code issues**

I wanted to use ActiveRecord::Dirty to determine if address was updated instead of using old_tax_zone / new_tax_zone but since Spree::Address factory if creating a new record each time we update a value, it's not possible to my knowledge. There might be a better way though

**To determine**

What if an order is paid? (like if you accept the pending payments, you double-check before shipping and notice that the address is wrong)

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -15,6 +15,7 @@ module Spree
     alias_attribute :last_name, :lastname
 
     DB_ONLY_ATTRS = %w(id updated_at created_at)
+    TAX_ZONE_ONLY_ATTRS = %w(state_id country_id zipcode)
 
     self.whitelisted_ransackable_attributes = %w[firstname lastname]
 
@@ -73,9 +74,14 @@ module Spree
       base.except!(*DB_ONLY_ATTRS)
     end
 
+
     # @return [Hash] hash of attributes contributing to value equality
     def value_attributes
       self.class.value_attributes(attributes)
+    end
+
+    def tax_zone_attributes
+      self.class.value_attributes(attributes.slice(*TAX_ZONE_ONLY_ATTRS))
     end
 
     # @return [String] the full name on this address
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -15,7 +15,7 @@ module Spree
     alias_attribute :last_name, :lastname
 
     DB_ONLY_ATTRS = %w(id updated_at created_at)
-    TAX_ZONE_ONLY_ATTRS = %w(state_id country_id zipcode)
+    TAXATION_ATTRS = %w(state_id country_id zipcode)
 
     self.whitelisted_ransackable_attributes = %w[firstname lastname]
 
@@ -74,14 +74,13 @@ module Spree
       base.except!(*DB_ONLY_ATTRS)
     end
 
-
     # @return [Hash] hash of attributes contributing to value equality
     def value_attributes
       self.class.value_attributes(attributes)
     end
 
-    def tax_zone_attributes
-      self.class.value_attributes(attributes.slice(*TAX_ZONE_ONLY_ATTRS))
+    def taxation_attributes
+      self.class.value_attributes(attributes.slice(*TAXATION_ATTRS))
     end
 
     # @return [String] the full name on this address
--- a/core/app/models/spree/order_contents.rb
+++ b/core/app/models/spree/order_contents.rb
@@ -22,6 +22,10 @@ module Spree
     end
 
     def update_cart(params)
+      # We need old_tax_address / new_tax_address because we can't rely on methods
+      # offered by ActiveRecord::Dirty to determine if tax_address was updated
+      # because if we update the address, a new record will be created
+      # by the Address.factory instead of the old record being updated
 
       old_tax_address = order.tax_address
 
@@ -29,7 +33,7 @@ module Spree
 
         new_tax_address = order.tax_address
 
-        unless new_tax_address.== old_tax_address
+        if should_recalculate_taxes?(old_tax_address, new_tax_address)
           order.create_tax_charge!
         end
 
@@ -66,6 +70,20 @@ module Spree
     end
 
     private
+
+    def should_recalculate_taxes?(old_address, new_address)
+      # Possible cases:
+      # [Old_address is a TaxLocation, New_address is a TaxLocation ]:
+      # NON-VAT order before Address state : we don't compute taxes.
+      # [Old_address is a TaxLocation, New_address is an Adress ]:
+      # Order who passed Address state : we compute taxes.
+      # [Old_address is an Adress, New_address is a TaxLocation ]:
+      # Should not happen, TaxLocation can be replaced by an Address but not the other way around.
+      # Order is a weird state and we do nothing because it should not happen.
+      return if old_address.is_a?(Spree::Tax::TaxLocation) || new_address.is_a?(Spree::Tax::TaxLocation)
+
+      old_address.try(:taxation_attributes) != new_address.try(:taxation_attributes)
+    end
 
     def after_add_or_remove(line_item, options = {})
       reload_totals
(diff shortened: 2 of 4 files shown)


core/spec/models/spree/order_contents_spec.rb
```diff
@@ -224,28 +224,34 @@
       }.to change { subject.order.total }
     end
 
-    context "submits an address in a different tax zone than the current address" do
-      let(:shipping_address) { create :address, state_code: "NY" }
+    context "submits an address in a different tax zone than the current address", :tag => :focus do
```
mamhoff: That `focus` tag shouldn't go into the codebase :)

core/spec/models/spree/order_contents_spec.rb
```diff
@@ -224,28 +224,34 @@
...
 
       it "updates tax adjustments" do
+        puts params.inspect
         expect(subject.order).to receive(:create_tax_charge!)
         subject.update_cart params
       end
     end
 
-    context "submits an address in the same tax zone than the current address" do
+    context "submits an address in the same tax zone than the current address"do
```
mamhoff: Can you add that space back in?

core/spec/models/spree/order_contents_spec.rb
```diff
@@ -224,28 +224,34 @@
...
 
       let(:default_address) { create :address, state_code: "NY", zipcode: "17402" }
       let(:updated_address) { create :address, state_code: "NY", zipcode: "17402"}
 
       before { order.update_attributes(ship_address: default_address, bill_address: default_address) }
 
       let(:params) do
-        { ship_address: updated_address}
+        { ship_address: updated_address, bill_address: updated_address}
```
mamhoff: And add a space here?

core/app/models/spree/address.rb
```diff
@@ -73,11 +74,16 @@ def self.value_attributes(base_attributes, merge_attributes = nil)
...
 
+
     # @return [Hash] hash of attributes contributing to value equality
     def value_attributes
       self.class.value_attributes(attributes)
     end
 
+    def tax_zone_attributes
+      self.class.value_attributes(attributes.slice(*TAX_ZONE_ONLY_ATTRS))
+    end
```
mamhoff: Does this method have any unit tests?

dangerdogz: I knew I forgot something, going to add some

mamhoff:
This is excellent work. I advised @dangerdogz IRL to do this at the model level, however I have an impression moving this to `Spree::Order` or `Spree::OrderUpdater` instead of `Spree::OrderContents` might be a worthwhile. 

There's one thing about the commits: While it's great to see your TDD, we want only commits that actually pass CI in the codebase. Can you group the commits that add functionality together with their new tests? Also, I agree with Hound most of the time.

Notwithstanding, good work!

core/app/models/spree/order_contents.rb
```diff
@@ -57,6 +70,16 @@ def approve(user: nil, name: nil)
 
     private
 
+    def should_recalculate_taxes?(old_address, new_address)
+      # Possible cases:
+      # [Old_address is a TaxLocation, New_address is a TaxLocation ]: non VAT order with no existing addresses who has not yet reached the address state, therefore we don't compute taxes
+      # [Old_address is a TaxLocation, New_address is an Adress ]: Order who reached at least the address state, therefore we can compute taxes
+      # [Old_address is an Adress, New_address is a TaxLocation ]: Should not happen, TaxLocation can be replaced by an Address but not the other way around. Order is a weird state and we do nothing because it should not happen
```
mamhoff: Can you re-format the comments so they fit in a GitHub code preview? :)

core/spec/models/spree/order_contents_spec.rb
```diff
@@ -224,6 +224,72 @@
       }.to change { subject.order.total }
     end
 
+    context "given an order with existing addresses" do
```
mamhoff: These specs are great, and even if we decide to move the logic somewhere else, they should absolutely stay.

core/app/models/spree/address.rb
```diff
@@ -15,6 +15,7 @@ class Address < Spree::Base
     alias_attribute :last_name, :lastname
 
     DB_ONLY_ATTRS = %w(id updated_at created_at)
+    TAX_ZONE_ONLY_ATTRS = %w(state_id country_id zipcode)
```
mamhoff: This is a naming thing: In my opinion, the `tax_zone` is a weird implementation-specific artifact. Can we name this constant `TAXATION_ATTRIBUTES` or something like that that does not imply `Spree::Zone` to be involved? Because it isn't (directly)

dangerdogz:
I fixed the various style issues, and addressed the feedback provided by @mamhoff 

Also added unit testing for the taxation_attributes method added

core/app/models/spree/order_contents.rb
```diff
@@ -57,6 +71,20 @@ def approve(user: nil, name: nil)
...
+      # [Old_address is a TaxLocation, New_address is a TaxLocation ]:
+      # NON-VAT order before Address state : we don't compute taxes.
+      # [Old_address is a TaxLocation, New_address is an Adress ]:
+      # Order who passed Address state : we compute taxes.
+      # [Old_address is an Adress, New_address is a TaxLocation ]:
+      # Should not happen, TaxLocation can be replaced by an Address but not the other way around.
+      # Order is a weird state and we do nothing because it should not happen.
+      return if old_address.is_a?(Spree::Tax::TaxLocation) || new_address.is_a?(Spree::Tax::TaxLocation)
+
+      old_address.try(:taxation_attributes) != new_address.try(:taxation_attributes)
```
jhawthorn: Sorry for the delay in reviewing. Could these be `.try!`?

Otherwise I think this is a great change and is a step in the right direction towards relying less on the state machine to get everything right.

dangerdogz: Comments look wrong, so I need to fix that but otherwise : 

I don't think TaxLocation needs to implement taxation_attributes because we're not supposed to compare TaxLocations for the following reasons : if old_address is a TaxLocation it means that that it's either a VAT order and it's handled elsewhere or it's a standard order who hasn't reached the Address state so the taxes will be handled by the checkout

The only case where this is suppose to apply is when you have two addresses, so this would means that you're either updating the customer details from the backend, or that a registered user is updating his default addresses so the taxes showing in the cart would be updated.

Does it make mores sense? 

Good point about using try! though, we can be extra sure that we're comparing Spree::Addresses because it will crash otherwise

dangerdogz:
Reworded the comments a bit, did a double check on the tests, should be ready for review again once build is finished

jhawthorn:
Looking good to me :+1:

mamhoff:
Yep, this is great! 💯

mamhoff:
Thanks again Fréderic!

PR #1083, opened 2016-04-19
Deprecate parsing in Spree::Money.new

jhawthorn:
Previously Spree::Money was always initialized with a string which would be parsed, like `"$1.00 USD"` or `"$2,10 EUR"`. This introduces a Spree::Money.parse which should be used instead for this type of input. Any currency passed in was appended to the `amount` string before parsing to end up with the correct currency.

This PR deprecates the complex parsing in `Spree::Money.new`, requiring either a Numeric or a string which is valid as input to `BigDecimal.new`. In solidus Money was already used in this way, but this deprecates the existing behaviour to be as safe as possible.

This now allows initializing `Spree::Money` with an existing `::Money` object. `Spree::Money.parse` was introduced to provide the existing behaviour, matching `Monetize.parse`.

This should be faster and less error-prone.

Inspired by @mamhoff's recent work improving Spree::Money

### DIFF
--- a/core/lib/spree/money.rb
+++ b/core/lib/spree/money.rb
@@ -12,6 +12,15 @@ module Spree
       def from_money(money)
         new(money)
       end
+
+      def parse(amount, currency = Spree::Config[:currency])
+        new(parse_to_money(amount, currency))
+      end
+
+      # @api private
+      def parse_to_money(amount, currency)
+        ::Monetize.parse(amount, currency)
+      end
     end
     self.default_formatting_rules = {
       # Ruby money currently has this as false, which is wrong for the vast
@@ -29,7 +38,16 @@ module Spree
       if amount.is_a?(::Money)
         @money = amount
       else
-        @money = Monetize.parse(amount, (options[:currency] || Spree::Config[:currency]))
+        currency = (options[:currency] || Spree::Config[:currency])
+        if amount.to_s =~ /\A-?\d+(\.\d+)?\z/
+          @money = Monetize.from_string(amount, currency)
+        else
+          @money = Spree::Money.parse_to_money(amount, currency)
+          Spree::Deprecation.warn <<-WARN.squish, caller
+            Spree::Money was initialized with #{amount.inspect}, which will not be supported in the future.
+            Instead use Spree::Money.new(#{@money.to_s.inspect}, options) or Spree::Money.parse(#{amount.inspect})
+          WARN
+        end
       end
       @options = Spree::Money.default_formatting_rules.merge(options)
     end
@@ -53,7 +71,7 @@ module Spree
     #   currency symbol
     # @return [String] the value of this money object formatted according to
     #   its options
-    def format(options={})
+    def format(options = {})
       @money.format(@options.merge(options))
     end
 
--- a/core/lib/spree/money.rb
+++ b/core/lib/spree/money.rb
@@ -6,6 +6,8 @@ module Spree
   # Spree::Money is a relatively thin wrapper around Monetize which handles
   # formatting via Spree::Config.
   class Money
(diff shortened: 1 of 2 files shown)


mamhoff:
I like this, of course! Now that `new` accepts a `Money` object, can you also get rid of `from_money`? I only introduced that because `new` couldn't do what it does two days ago. Now it can, so I think `from_money` can actually go. It has no specs, and is only used inside the class (in `+` and `-`).

core/lib/spree/money.rb
```diff
@@ -10,7 +10,16 @@ class <<self
...
       def from_money(money)
-        new(money.to_d, currency: money.currency.iso_code)
+        new(money)
+      end
+
+      def parse(amount, currency = Spree::Config[:currency])
+        new(parse_to_money(amount, currency))
+      end
+
+      # @api private
```
tvdeyen: Why not make it `private` then? ;)

core/lib/spree/money.rb
```diff
@@ -23,9 +32,33 @@ def from_money(money)
...
-    # @param options [Hash] the options for creating the money object
-    # @option options [String] currency the currency for the money object
+    # @param amount [Money, #to_s] the value of the money object
+    # @param options [Hash] the default options for formatting the money object See #format
+    def initialize(amount, options = {})
+      if amount.is_a?(::Money)
+        @money = amount
+      else
+        currency = (options[:currency] || Spree::Config[:currency])
+        if amount.to_s =~ /\A-?\d+(\.\d+)?\z/
```
tvdeyen: Please put this into a constant. We don't want "magic values" in our code base, right?

tvdeyen: I talked with @mamhoff and we think the Regex constant should be called `RUBY_NUMERIC_STRING`, so the reader knows what we want to check here.

core/lib/spree/money.rb
```diff
@@ -23,9 +32,33 @@ def from_money(money)
...
+          @money = Monetize.from_string(amount, currency)
+        else
+          @money = Spree::Money.parse_to_money(amount, currency)
+          Spree::Deprecation.warn <<-WARN.squish, caller
+            Spree::Money was initialized with #{amount.inspect}, which will not be supported in the future.
+            Instead use Spree::Money.new(#{@money.to_s.inspect}, options) or Spree::Money.parse(#{amount.inspect})
+          WARN
```
tvdeyen: After we remove this deprecation we should refactor this whole method ;)

core/lib/spree/money.rb
jhawthorn: It's called both from an instance of Money, and from the class so it can't be :(

We can make it private once the deprecation is removed.

core/lib/spree/money.rb
jhawthorn: Thank you. Excellent suggestion.

core/lib/spree/money.rb
```diff
@@ -6,11 +6,22 @@ module Spree
...
   # formatting via Spree::Config.
   class Money
+    RUBY_NUMERIC_STRING = /\A-?\d+(\.\d+)?\z/
+
     class <<self
       attr_accessor :default_formatting_rules
 
       def from_money(money)
-        new(money.to_d, currency: money.currency.iso_code)
+        new(money)
```
mamhoff: This is becoming the ultimate bikeshedding exercise, but: This method is now useless. Can you please please get rid of it?

core/lib/spree/money.rb
tvdeyen: Aight

tvdeyen:
Perfect. Thanks John. :+1:

Issue #1076, opened 2016-04-18
Add navigable breadcrumbs in the header


Mandily:
The title of many of the admin's pages is unintuitive and a duplicate of the parent pages. This issue exists to clean up page titles and introduce a navigable breadcrumb that better informs the user of where they are in the admin.

Breadcrumb labels should be a link when a specific parent page exists that can be linked to. For example the product listing page can be a link in the breadcrumbs, when the user is on a specific product. This means that we'll also be able to get rid of all "go back" buttons that currently exist.

A consistent page title will also be enforced. As "Listing Products" and "Products" are the same page, they will need to be made consistent with "Products" in all references.
<img width="355" alt="screen shot 2016-04-18 at 10 43 37 am" src="https://cloud.githubusercontent.com/assets/15271353/14613701/7d52546a-0552-11e6-8761-e14cbda2bd08.png">
<img width="301" alt="screen shot 2016-04-18 at 10 43 45 am" src="https://cloud.githubusercontent.com/assets/15271353/14613702/7d55b754-0552-11e6-8bf3-bd2b9308b74f.png">

Currently some pages use 'edit' in the title and others do not. To make it consistent and more concise across the app, we'll get rid of 'editing' in the breadcrumbs. When the user is on a page with fields instead of static text, there is enough affordance to tell them it is editable.

**Current**
<img width="972" alt="screen shot 2016-04-18 at 10 49 30 am" src="https://cloud.githubusercontent.com/assets/15271353/14613893/6641c4c6-0553-11e6-9a05-a5d8b4f418f1.png">
**Future**
<img width="964" alt="screen shot 2016-04-18 at 10 49 08 am" src="https://cloud.githubusercontent.com/assets/15271353/14613892/6641888a-0553-11e6-8f32-efc83075688c.png">

Please feel free to give feedback in the comments below.

tvdeyen:
100% acknowledged :+1:

Murph33:
Just like to add I'd started doing some work on this before getting side tracked with things.  I do intend to get back to finishing it when I have some available time.

PR #1085, opened 2016-04-20
Add breadcrumbs to page title in admin backend

Murph33:
Adding breadcrumbs will make for better UX.  As requested in #1076.  The button link to indexes have been taken out and replaced with navigatable breadcrumbs at the top of just about every page.

No new partials were added but that might be something we want to look at; pulling out the code in related pages when we can.  However in a lot of cases it would just be pulling 1-2 lines of code into a partial and then still requiring to add some page title code to that individual page.  

Just about every page was touched so the easiest way to get a feel for the changes might be to pull this down and just click around locally.

## Before Breadcrumbs

no_breadcrumbs_1
no_breadcrumbs_2
no_breadcrumbs_3
no_breadcrumbs_4

## After Breadcrumbs

afterbread1
afterbread2
afterbread3
afterbread4

### DIFF
--- a/backend/app/helpers/spree/admin/navigation_helper.rb
+++ b/backend/app/helpers/spree/admin/navigation_helper.rb
@@ -1,6 +1,27 @@
 module Spree
   module Admin
     module NavigationHelper
+      # Add items to current page breadcrumb heirarchy
+      def add_breadcrumb(*ancestors, &block)
+        breadcrumbs.concat(ancestors) if ancestors.present?
+        breadcrumbs.push(capture(&block)) if block_given?
+      end
+
+      def add_page_title_to_breadcrumbs
+        ActiveSupport::Deprecation.warn('content_for(:page_title) is deprecated, use add_breadcrumb')
+        add_breadcrumb(content_for(:page_title))
+      end
+
+      # Render Bootstrap style breadcrumbs
+      def render_breadcrumbs
+        add_page_title_to_breadcrumbs if content_for?(:page_title)
+        content_tag :ol, class: 'breadcrumb' do
+          safe_join breadcrumbs.collect { |level|
+            content_tag(:li, level, class: "separator #{level == breadcrumbs.last ? 'active' : ''}")
+          }
+        end
+      end
+
       # Make an admin tab that coveres one or more resources supplied by symbols
       # Option hash may follow. Valid options are
       #   * :label to override link text, otherwise based on the first resource name (translated)
@@ -135,6 +156,12 @@ module Spree
         content_tag(:li, options) do
           link_to(link_text, url)
         end
+      end
+
+      private
+
+      def breadcrumbs
+        @breadcrumbs ||= []
       end
     end
   end
--- a/backend/app/helpers/spree/admin/navigation_helper.rb
+++ b/backend/app/helpers/spree/admin/navigation_helper.rb
@@ -22,6 +22,18 @@ module Spree
         end
       end
 
+      def page_title
+        if content_for?(:title)
+          content_for(:title)
+        elsif content_for?(:page_title)
+          content_for(:page_title)
+        elsif breadcrumbs.any?
+          strip_tags(breadcrumbs.last)
+        else
+          Spree.t(controller.controller_name, default: controller.controller_name.titleize)
+        end
+      end
+
       # Make an admin tab that coveres one or more resources supplied by symbols
       # Option hash may follow. Valid options are
       #   * :label to override link text, otherwise based on the first resource name (translated)
--- a/backend/app/views/spree/admin/adjustment_reasons/edit.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/edit.html.erb
@@ -2,7 +2,7 @@
 
 <% add_breadcrumb Spree.t(:settings) %>
 <% add_breadcrumb Spree.t('admin.tab.checkout') %>
-<% add_breadcrumb link_to Spree::AdjustmentReason.model_name.human(count: :other), spree.admin_adjustment_reasons_path %>
+<% add_breadcrumb link_to plural_resource_name(Spree::AdjustmentReason), spree.admin_adjustment_reasons_path %>
 <% add_breadcrumb @object.name %>
 
 <%= render :partial => 'spree/shared/error_messages', :locals => { :target => @object } %>
--- a/backend/app/views/spree/admin/adjustment_reasons/edit.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/edit.html.erb
@@ -1,11 +1,13 @@
 <%= render 'spree/admin/shared/settings_checkout_tabs' %>
 
 <% content_for :page_title do %>
-  <%= Spree.t(:editing_adjustment_reason) %>
+  <%= Spree.t(:settings) %> &#x2f;
+  <%= Spree.t('admin.tab.checkout') %> &#x2f;
+  <%= link_to Spree::AdjustmentReason.model_name.human(count: :other), spree.admin_adjustment_reasons_path %> &#x2f;
+  <%= @object.name %>
 <% end %>
 
 <% content_for :page_actions do %>
-  <li><%= link_to_with_icon 'arrow-left', Spree.t(:back_to_adjustment_reason_list), collection_url, :class => 'button' %></li>
 <% end %>
 
 <%= render :partial => 'spree/shared/error_messages', :locals => { :target => @object } %>
--- a/backend/app/views/spree/admin/adjustment_reasons/edit.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/edit.html.erb
@@ -1,14 +1,9 @@
 <%= render 'spree/admin/shared/settings_checkout_tabs' %>
 
-<% content_for :page_title do %>
-  <%= Spree.t(:settings) %> &#x2f;
-  <%= Spree.t('admin.tab.checkout') %> &#x2f;
-  <%= link_to Spree::AdjustmentReason.model_name.human(count: :other), spree.admin_adjustment_reasons_path %> &#x2f;
-  <%= @object.name %>
-<% end %>
-
-<% content_for :page_actions do %>
-<% end %>
+<% add_breadcrumb Spree.t(:settings) %>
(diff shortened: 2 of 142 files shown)


backend/app/views/spree/admin/orders/customer_details/edit.html.erb
```diff
@@ -1,11 +1,10 @@
 <%= render :partial => 'spree/admin/shared/order_tabs', :locals => { :current => 'Customer Details' } %>
 
 <% content_for :page_title do %>
-  <i class="fa fa-arrow-right"></i> <%= Spree.t(:customer_details) %>
+  &#x2f; <%= Spree.t(:customer_details) %>
 <% end %>
 
 <% content_for :page_actions do %>
```
tvdeyen: Shouldn't we remove all empty `content_for :page_actions` blocks? Or do we want to keep them for potential deface overrides?

Murph33: Something along those lines was my thought when I was going through but in hindsight I suppose since they're just empty `content_for`s with no text in them they wouldn't even affect the rendering of that element on the page or not.  I'm not sure if that will cause the `page_actions` block to still render on the page but I imagine we would want to keep it for the potential deface overrides.

tvdeyen: Or just remove them. I don't think it's "good practice" to deface `content_for`s, right?

tvdeyen:
Wow, it just took me quite some time to figure out where the breadcrumbs for order views came from. Maybe this is the opportunity to move it out of `order_tabs` partial into the views, no? Or at least rename the partial into something more understandable? Don't have a good name in my mind right now, though.

tvdeyen:
Besides that 👍 this is a great improvement!

backend/app/views/spree/admin/orders/customer_details/edit.html.erb
Murph33: Oh, my thought was that without some text in that `content_for` that then there would no longer be a `page_actions` element on the page and possibly people would be defacing that but I haven't checked this hypothesis.

Mandily:
I pulled this down to look at it and noticed one small thing. In the reports list the name is 'sales total' (singular) in the breadcrumbs it's plural - I don't care which it is, but can you make it consistent please?

Otherwise it looks good!

backend/app/views/spree/admin/orders/customer_details/edit.html.erb
tvdeyen: Ah, you're absolutely right. We check with `content_for? :page_actions` and only render the div if something is present. We could remove the check or advice people to add `content_for :page_actions` to their overrides. I would vote for to go with latter. It was never guaranteed that this div exists anyway. Thoughts @jhawthorn ?

tvdeyen:
"She is called @Mandily sharp eyes"

Murph33:
With `order_tabs` I'd rather rename that to something more descriptive then pull the information out to every page.  Partials being responsible for some of the breadcrumbs does make sense in some of the cases.

backend/app/views/spree/admin/orders/customer_details/edit.html.erb
Murph33: Just checked this out.  Without adding a &nbsp; or something to the block the `<ul class='page-actions'>` element doesn't show up on the page.  I'm thinking we'd probably like to make sure that we don't lose those in case people are hooking into them.  I will make sure I change that.

Murph33:
I went with `Sales total` because the report list is calling an actual value on the `report` and changing that would seem out of scope but I think `Sales Totals` would make more sense in each place.

backend/app/views/spree/admin/adjustment_reasons/edit.html.erb
```diff
@@ -1,11 +1,13 @@
 <%= render 'spree/admin/shared/settings_checkout_tabs' %>
 
 <% content_for :page_title do %>
-  <%= Spree.t(:editing_adjustment_reason) %>
+  <%= Spree.t(:settings) %> &#x2f;
```
graygilmore: Instead of writing `&#x2f;` all of the time I think it would be better if we used CSS for this. Something like:

```scss
.page-title > *:before {
  content: ">";
}
```

graygilmore:
I don't think I like having all of these things shoved into the `h1`. They should be separate things. Maybe the active item could be the `h1`? Either way I think the implementation needs to change a bit.

backend/app/helpers/spree/admin/navigation_helper.rb
```diff
@@ -1,6 +1,39 @@
...
+      def add_page_title_to_breadcrumbs
+        ActiveSupport::Deprecation.warn('content_for(:page_title) is deprecated, use add_breadcrumb')
+        add_breadcrumb(content_for(:page_title))
+      end
+
+      # Render Bootstrap style breadcrumbs
+      def render_breadcrumbs
+        add_page_title_to_breadcrumbs if content_for?(:page_title)
+        content_tag :ol, class: 'breadcrumb' do
+          safe_join breadcrumbs.collect { |level|
```
graygilmore: `do`/`end` for multi-line blocks

Murph33:
So @Sinetheta hacked something much nicer together on the hackdays and kindly kept me involved by giving me a PR which apparently github doesn't notify you of.  I finally noticed today when I asked him how it was coming along so it is merged now.  I see Gray already commenting so I won't ping him.  Hoping you're still giving the thumbs up @tvdeyen!

Murph33:
I would just like to point out that the appearance of the breadcrumbs have changed since it's no longer in an H1 (which as Gray pointed out was a mistake in the first place) so if style opinionated people would chime in that would be appreciated.  All I did was bump the size some since they are in the old title's position and the small text size didn't look right to me.  I imagine this will get changed around some as we continue to make changes to the admin.

tvdeyen:
Is there a reason why we don't use `content_for` here? The API seams very unintuitive and tedious to me. 

```erb
<% content_for(:breadcrumbs) do %>
  <% link_to ... %>
  ...
<% end %>
```

Then when rendering the breadcrumbs we can take care of active state, etc.

Sinetheta:
@tvdeyen my thought was to get the breadcrumbs closer to meta-data. With a content_for we have no validation over what's being jammed up there. With an interface like `add_breadcrumb` I'm hoping that we can soon take this decision out of the views entirely and get the info from our navigation.

Currently we have a quasi configurable nav. Some of it is in [backend_configuration.rb](https://github.com/solidusio/solidus/blob/60ee1dffc7c3932cd7652cc2b4566f08a12f74b8/backend/app/models/spree/backend_configuration.rb) but mostly it's contained int he views. I've talked with @jhawthorn about how we might make the nav configurable and I think he liked this pattern https://github.com/codeplant/simple-navigation/wiki/Configuration

> Then when rendering the breadcrumbs we can take care of active state, etc.
How would we do this with a `content_for` approach?

tvdeyen:
I like the idea of having a configurable navigation. That would be a great integration point for extensions.

Murph33:
This has been updated to use the plural_resource_name helper and rebased on master.  Yes/No @tvdeyen, @Mandily, @Sinetheta , @mamhoff, @jhawthorn, @graygilmore, et al?

Mandily:
I know you removed the H1 from the breadcrumbs to be more semantic, but could you add a font-size of 16px (or 1rem, whatever unit we use) to the breadcrumbs to give it a bit more prominence on the page? That will make it smaller than what it was to account for more horizontal space being used, but larger than the 13px it and the rest of the body text is now.

Murph33:
Changed pictures in original post and bumped the font size.

backend/app/views/spree/admin/cancellations/index.html.erb
```diff
@@ -1,6 +1,7 @@
-<% content_for :page_title do %>
-  <%= link_to Spree.t(:order_number, number: @order.number), spree.edit_admin_order_path(@order) %> - <%= Spree.t(:cancel_inventory) %>
-<% end %>
+<% add_breadcrumb(plural_resource_name(Spree::Order), spree.admin_orders_path) %>
```
tvdeyen: I think we miss a `link_to` here?

tvdeyen:
Besides lots of (useless) additional new lines 👍

Thinking about having a `breadcrumbs` helper that takes a block or array of entries, so we can spare the repetitive `add_breadcrumb` calls. We just call the `add_breadcrumb` for each of the entries then. 

Would be much more easy to use, no?

Otherwise, great addition. Like that very much from the UI perspective.

Murph33:
I apologize while also blaming @Sinetheta for any unnecessary new lines that have appeared.  You were right @tvdeyen that did require a link_to in there, good catch.  I'm agreed with Kevin as well when you factor in the build up in partials we have that this while not necessarily being a perfect world implementation, is serviceable.   

Fixed the link_to and rebased on current master.

Murph33:
Rebased on current master.  Would really appreciate if we could get this merged or decide on some other change as I know @Mandily is eager to have it in so we can continue moving forward.

@tvdeyen, @jhawthorn, thanks.

backend/app/helpers/spree/admin/navigation_helper.rb
mamhoff: Both of these are valueable comments, I'd like to see them addressed (`map` > `collect`, `do / end` for multi-line blocks).

mamhoff:
I think this is good, but an intermediate step. Once we have a configurable navigation across levels, that should handle the breadcrumbs. It's better than what we had before though, so 👍 .

tvdeyen:
Yes, this is better as it's right now. 👍

backend/app/helpers/spree/admin/navigation_helper.rb
jhawthorn: `do`/`end` is actually a little more awkward here because `safe_join breadcrumbs.collect do |level|` will associate the block with `safe_join` rather than `collect`.

I agree with the hound, prefer `map` over `collect`.

jhawthorn:
There are several changes I think this needs but I will send them in a future PR.

Thank you @Murph33, @Sinetheta, and @Mandily

PR #1086, opened 2016-04-20
Interfaces for persist_user_credit_card and assign_default_credit_card

jordan-brough:
Extract the logic for persist_user_credit_card and
assign_default_credit_card into configurable classes.

Does this seem like something we'd like to do?  This is from a pairing session with
@mamhoff around non-credit card payment sources.  If it seems good then I'll
add some specs around it.

### DIFF
--- a/core/app/models/spree/wallet/add_after_order_complete.rb
+++ b/core/app/models/spree/wallet/add_after_order_complete.rb
@@ -1,0 +1,25 @@
+class Spree::Wallet::AddAfterOrderComplete
+  def initialize(order)
+    @order = order
+  end
+
+  # AddAfterOrderComplete is called after an order transitions to complete. It
+  # is responsible for saving payment sources in the user's "wallet" for future
+  # use.
+  def add_to_wallet
+    if !order.temporary_credit_card &&
+       order.user_id &&
+       order.valid_credit_cards.present?
+      # arbitrarily pick the first one for the default
+      default_cc = order.valid_credit_cards.first
+      # TODO: target for refactoring -- why is order checkout responsible for the user -> credit_card relationship?
+      default_cc.user_id = order.user_id
+      default_cc.default = true
+      default_cc.save
+    end
+  end
+
+  private
+
+  attr_reader :order
+end
--- a/core/app/models/spree/wallet/add_default_payment.rb
+++ b/core/app/models/spree/wallet/add_default_payment.rb
@@ -1,0 +1,24 @@
+class Spree::Wallet::AddDefaultPayment
+  def initialize(order)
+    @order = order
+  end
+
+  # Build a payment to be added to an order prior to moving into the "payment"
+  # state.
+  #
+  # @return [Payment] the unsaved payment to be added, or nil if none.
+  def build_payment
+    credit_card = order.user.try!(:default_credit_card)
+
+    if credit_card.try!(:valid?) && order.payments.from_credit_card.count == 0
+      Spree::Payment.new(
+        payment_method_id: credit_card.payment_method_id,
+        source: credit_card,
+      )
+    end
+  end
+
+  private
+
+  attr_reader :order
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 ## Solidus 1.4.0 (master, unreleased)
 
+*   Make some 'wallet' behavior configurable
+
+    NOTE: `Order#persist_user_credit_card` has been renamed to
+    `Order#add_payment_sources_to_wallet`. If you are overriding
+    `persist_user_credit_card` you need to update your code.
+
+    The following extension points have been added for customizing 'wallet'
+    behavior.
+
+    * Spree::Config.add_payment_sources_to_wallet_class
+    * Spree::Config.default_payment_builder_class
+
+    https://github.com/solidusio/solidus/pull/1086
+
 *   Backend: UI, Remove icons from buttons and tabs
 
 *   Backend: Deprecate args/options that add icons to buttons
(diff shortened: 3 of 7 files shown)


cbrunsdon:
I'm torn on having the "add after order complete" being a separate objects. I almost think it should be supported somehow by the wallet API itself to keep the api more cohese and clear for anyone using it.

I'd almost expect something more generic to be called like:

```ruby
wallet.events.order_complete(order)
```

which users could do as they wish with (including persist it, if desired).

The second one, the add_default_payment_source I'm weirded out about for another reason. It seems weird that a Wallet:: class would be responsible for creating a payment on an order after a transition. But thats maybe mostly related to how ugly it is now.

core/app/models/spree/wallet/add_default_payment.rb
```diff
@@ -0,0 +1,24 @@
+class Spree::Wallet::AddDefaultPayment
```
tvdeyen: We should consider to call this class `DefaultPaymentBuilder` as we actually don't add something, we build something. Otherwise great!

core/app/models/spree/wallet/add_after_order_complete.rb
```diff
@@ -0,0 +1,25 @@
+class Spree::Wallet::AddAfterOrderComplete
```
tvdeyen: I no this name too strictly named after where it's called from. I think it should be named more like what it does. My English vocabulary is limited, so I only came up with `Spree::Wallet::Storer`. But, besides that I really like the way we are going here. Configurable POROs that are responsible for the actual implementation. 👍

tvdeyen:
Good step in the right direction. I Like that.

core/app/models/spree/wallet/add_after_order_complete.rb
peterberkenbosch: Yes, I agree with @tvdeyen here. This will also open the possibility to add sources on other places. For example, adding bitcoin in "my account". Also.. naming things is hard :)

core/app/models/spree/wallet/add_default_payment.rb
peterberkenbosch: And with that name change, the method could be named just `build` I think.

peterberkenbosch:
I see a `Wallet` as an abstraction for the payment sources, and ideally very loosely coupled with any state event change on `Order`. I have some comments inline as well that reflect that opinion. 

Love to discuss this further and move forward :)

core/app/models/spree/wallet/add_default_payment.rb
jordan-brough: @tvdeyen @peterberkenbosch updated. Thanks!

core/app/models/spree/wallet/add_after_order_complete.rb
jordan-brough: @tvdeyen @peterberkenbosch hm, I am not sure about this one.  Right now the logic for this is very dependent on this happening during order completion.  `Wallet::AddAfterOrderComplete` seems like it might be the best we can get right now.  What if we leave it like this for the moment and take another stab at improving things as part of the Wallet work?

jordan-brough:
@cbrunsdon I hear you on the weirdness and current ugliness.  I'd lean toward getting this in since it solves the biggest problem of stores being able to customize this logic without monkey patching and then revisit during or after the main Wallet PR.  What do you think though?

> `wallet.events.order_complete(order)`

I like that.  How about I make that change in the main Wallet PR after we merge this? (Since the `wallet` namespace doesn't exist yet at this point).

core/app/models/spree/wallet/add_after_order_complete.rb
peterberkenbosch: 👍

peterberkenbosch:
Also, this PR is pointing to master right now. Should it be towards this branch? solidusio:jordan/non-credit-card-payment-sources-base

jordan-brough:
> Also, this PR is pointing to master right now. Should it be towards this branch?

@peterberkenbosch I did want to point it at master. I did this in a weird way...but this was one of three things in that other branch that I wanted to get into master before the main Wallet PR.

peterberkenbosch:
Got it @jordan-brough, I missed the add commit with the `Spree::Wallet` namespace here. Perfect!

peterberkenbosch:
wdyt @cbrunsdon @jhawthorn

core/app/models/spree/wallet/add_after_order_complete.rb
```diff
@@ -0,0 +1,25 @@
...
+  def initialize(order)
+    @order = order
+  end
+
+  # AddAfterOrderComplete is called after an order transitions to complete. It
+  # is responsible for saving payment sources in the user's "wallet" for future
+  # use.
+  def add_to_wallet
```
bbuchalter: Given this comment (https://github.com/solidusio/solidus/pull/1086/files#diff-a9fae1594c22d1ec005ac8501d80d7afR356) and the importance of documenting these interfaces, could we add some more YARD docs to this method?

bbuchalter:
I would find this PR helpful. We've got these customizations hacked into our app now and would love to have a nice interface for it!

cbrunsdon:
I've thought about this one for a few days and can't think of anything that would improve it, and I think its better than what we have now, so I'm at a :+1:  if @jhawthorn is.

peterberkenbosch:
Thanks @cbrunsdon! @jhawthorn wdyt? :)

jordan-brough:
@jhawthorn updated as per our chat
@bbuchalter I added some more comments/YARD docs also

jhawthorn:
:+1: Looks good

PR #1088, opened 2016-04-21
Fix migrating preferences from old spree versions

jhawthorn:
Previously this failed when migrating from an old version with:

    super: no superclass method 'preferences' for ...

This can be avoided by assigning to the attribute using `AR::Base#[]` instead of calling `super()`.


bbuchalter:
This there a chance this PR could also be backported to the 1.2.x series?

cbrunsdon:
I'm cool with this :+1:

PR #1089, opened 2016-04-21
Deprecate Spree::Product.active scope

mamhoff:
This scope is badly named (calling anything "active" in a rails
project makes searching hard, and it's hard to guess what `active`)
actually means). Besides, there is a `Spree::Variant#active` scope
that does a different thing. Furthermore, it takes two arguments, both of
which are ultimately unused.

Finally, the functionality offered by `.active` can just as well be
had by calling `.available`.


tvdeyen:
👍 and another obviously useless code gone!

gmacdougall:
:+1:

PR #1090, opened 2016-04-21
Improve coupon code JS applicator

mtomov:
* Separate Coupon code application from Saving & processing a payment 

    » allows users to first apply a code, and subsequently add payment details when they are certain of the final amount to be paid.

* Add new button Apply Code

* Separate payment js from coupon code js - allows for easier override
   as they are now in separate files.

* Specify your own css class as a main one, ex: `callout`. The JS applicator will automatically add `success` or `alert` classes  as appropriate.

* Allow customisation of the CSS classes applied on success and on error
   via data attributes `data-success-class` and `data-success-error`.

* Fix a refresh bug on coupon code application error - before, the page
   was refreshed even after displaying an AJAX brought error message.
   See issue & fix @Spree https://github.com/spree/spree/pull/7040

* Fixes a JS warning that a Synchronous AJAX request was executed, which could impact JS performance as it waits for response from the server before continuing the execution of the JS. (Was coming from the `async:false` setting)

* Switch to using the Spree API action for applying a code instead of the
   Frontend controller one at StoreController#apply_code.

   TODO: Delete the StoreController#apply_code similarly to Spree as
         it's not used anymore. See: https://github.com/spree/spree/pull/7284

--
coupon code applied successfully

coupon code doesnt exist

coupon code already applied

### DIFF
--- a/frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
+++ b/frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
@@ -1,0 +1,34 @@
+Spree.onCouponCodeApply = (e) ->
+  couponCodeField = $('#order_coupon_code')
+  couponCode = $.trim(couponCodeField.val())
+  return if couponCode == ''
+
+  couponStatus = $("#coupon_status")
+  successClass = couponStatus.data('success-class') || 'success'
+  errorClass = couponStatus.data('error-class') || 'error'
+  url = Spree.url(Spree.routes.apply_coupon_code(Spree.current_order_id),
+    {
+      order_token: Spree.current_order_token,
+      coupon_code: couponCode
+    }
+  )
+
+  couponStatus.removeClass([successClass,errorClass].join(" "));
+
+  req = Spree.ajax
+    method: "PUT",
+    url: url
+
+  req.done (data) ->
+    window.location.reload();
+    couponCodeField.val('')
+    couponStatus.addClass(successClass).html("Coupon code applied successfully.")
+
+  req.fail (xhr) ->
+    # handler = JSON.parse(xhr.responseText)
+    handler = xhr.responseJSON
+    couponStatus.addClass(errorClass).html(handler["error"])
+
+Spree.ready ($) ->
+  $('#coupon-code-apply-button').click (e) ->
+    Spree.onCouponCodeApply(e)
--- a/frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
+++ b/frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
@@ -4,8 +4,8 @@ Spree.onCouponCodeApply = (e) ->
   return if couponCode == ''
 
   couponStatus = $("#coupon_status")
-  successClass = couponStatus.data('success-class') || 'success'
-  errorClass = couponStatus.data('error-class') || 'error'
+  successClass = 'success'
+  errorClass = 'alert'
   url = Spree.url(Spree.routes.apply_coupon_code(Spree.current_order_id),
     {
       order_token: Spree.current_order_token,
--- a/frontend/app/views/spree/checkout/_payment.html.erb
+++ b/frontend/app/views/spree/checkout/_payment.html.erb
@@ -62,11 +62,9 @@
       <button type="button" class="button" id="coupon-code-apply-button">
         <%= Spree.t(:apply_code) %>
       </button>
+
     </p>
-    <div id='coupon_status'
-      data-success-class="success"
-      data-error-class="error">
-    </div>
+    <div id='coupon_status'></div>
 
   </div>
 </fieldset>
--- a/frontend/app/views/spree/checkout/_payment.html.erb
+++ b/frontend/app/views/spree/checkout/_payment.html.erb
@@ -59,7 +59,15 @@
     <p class='field' data-hook='coupon_code'>
       <%= form.label :coupon_code %><br />
       <%= form.text_field :coupon_code %>
+      <button type="button" class="button" id="coupon-code-apply-button">
+        <%= Spree.t(:apply_code) %>
+      </button>
     </p>
+    <div id='coupon_status'
+      data-success-class="success"
+      data-error-class="error">
+    </div>
+
   </div>
 </fieldset>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,11 @@
     * Partials are rendered on pages owned by the partials as tabs as a top bar
     * Admin-nav has a sub-menu for the settings now
 
+*   Coupon code application has been separated from the Continue button on the Payment checkout page
+
+    * JavaScript for it has been moved from address.js into its own `spree/frontend/checkout/coupon-code`
+    * Numerous small nuisances have been fixed [#1090](https://github.com/solidusio/solidus/pull/1090)
+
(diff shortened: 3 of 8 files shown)


frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
```diff
@@ -0,0 +1,39 @@
+Spree.onCouponCodeApply = (e) ->
+  coupon_code_field = $('#order_coupon_code')
+  coupon_code = $.trim(coupon_code_field.val())
+  return if coupon_code == ''
+
+  if $('#coupon_status').length == 0
```
jhawthorn: Is this necessary? The class is defined in the view, we should be able to rely on that.

frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
```diff
@@ -0,0 +1,39 @@
...
+    coupon_status = $("<div id='coupon_status'></div>")
+    coupon_code_field.parent().append(coupon_status)
+  else
+    coupon_status = $("#coupon_status")
+
+  successClass = coupon_status.data('success-class') || 'success'
+  errorClass = coupon_status.data('error-class') || 'error'
```
jhawthorn: I would prefer always using the same class name ex. `success` and `error`. Developers can customize that in css if need be.

frontend/app/views/spree/checkout/_payment.html.erb
```diff
@@ -59,7 +59,15 @@
     <p class='field' data-hook='coupon_code'>
       <%= form.label :coupon_code %><br />
       <%= form.text_field :coupon_code %>
+      <button type="button" class="button" id="coupon-code-apply-button">
+        <%= Spree.t(:apply_code) %>
+      </button>
     </p>
+    <div id='coupon_status'
```
jhawthorn: I believe this should be inside the above `<p>` tag to match the existing behaviour.

frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
mtomov: From my experience in integrating this code with Foundation, it was easier to specify the classes as data attributes than to add additional CSS to match this specific use case. I'd always avoid extra CSS if possible.

Unless others favour your opinion, I'd prefer to leave the data attributes in place.

frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
mtomov: I agree. Took it from the old script, sort of _backwards view compatible_, but I guess given that the view needs a button now, I can get rid of this.

mtomov:
Thanks for reviewing @jhawthorn !

I'd be happy to also hear any comments if anyone has.

cbrunsdon:
This looks pretty good to me, thanks @mtomov, :+1:. And I have nothing valuable to add to the open conversations.

mamhoff:
I'm just very happy about the `submit` Event listener being gone. 👍  -- also nothing valuabel to add to the existing conversations.

frontend/app/views/spree/checkout/_payment.html.erb
mtomov: Ah, I went about to change it, but it turns out that `<div>`s [can't be nested](http://stackoverflow.com/questions/8397852/why-p-tag-cant-contain-div-tag-inside-it) inside `<p>` tags.

Funny that I had tried to place it inside the `<p>` on my initial go, had discovered that it doesn't work, but had forgotten about it, and was trying it again ..  so bad ..

mtomov:
Hi @jhawthorn , I've made this one change in the JS (don't auto-create the div tag).

It's good that we don't attach to the `submit` event, as I have seen payment gateways ([Braintree](https://github.com/boomerdigital/spree_braintree_cse/blob/master/app/assets/javascripts/spree/braintree.js.coffee.erb#L19)) also attach to it, and it can get a bit messy when one needs to cancel the execution of the other..

Let me know if there's more I can adjust!

frontend/app/views/spree/checkout/_payment.html.erb
jhawthorn: That makes sense. Thanks for checking.

jhawthorn:
:+1:

cbrunsdon:
Still good by me, nice work :+1: 

We're almost definitely going to need a changelog entry for this change though.

frontend/app/views/spree/checkout/_payment.html.erb
```diff
@@ -59,7 +59,15 @@
...
+      data-success-class="success"
```
graygilmore: Since these values match the defaults here https://github.com/solidusio/solidus/pull/1090/files#diff-a065997c80be8446e236252c98a12800R7 we don't need to duplicate them here.

frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
```diff
@@ -0,0 +1,34 @@
+Spree.onCouponCodeApply = (e) ->
+  coupon_code_field = $('#order_coupon_code')
+  coupon_code = $.trim(coupon_code_field.val())
+  return if coupon_code == ''
+
+  coupon_status = $("#coupon_status")
```
graygilmore: I know that a bunch of this is being copied over but it would be a good opportunity to convert some of these variables into `camelCase`. The new stuff that you've added uses it 👍 so we might as well make everything in here a little nicer if we can.

frontend/app/views/spree/checkout/_payment.html.erb
mtomov: Yes, but they are present for documentation purposes here. Alternatively, I can remove the defaults in the JS, but I thought they don't add much noise, so why not have them.

graygilmore: They imply that they are overwriting default behaviour. If we want to document it both the changelog and the JS file make very good candidates for that.

mtomov: Yeah, but frontend developers (and even lazy people like me) don't tend to meddle with the middle of some JavaScript file to find out that they can actually customize the classes.

frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
mtomov: I can swap them. I can also give it a go with the changelog entry. It must be the one [at the root](https://github.com/solidusio/solidus/blob/master/CHANGELOG.md#solidus-130-unreleased) of the project, right? - just to add a bullet point at the bottom of 1.3?

frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js.coffee
graygilmore: I missed this because it was in an outdated diff. I'd lean towards agreeing with @jhawthorn here. Removing this would clean up both the JS and the markup and make the conversation below moot.

mtomov:
Changelog added + CSS classes streamlined. Let me know of anything else. Otherwise I can squash it. Thank you for the feedback!

jhawthorn:
:+1: Thank you

mtomov:
:+1:

PR #1092, opened 2016-04-21
Fix reference to Spree::Reimbursement::Credit class

adaddeo:


cbrunsdon:
Whoops. Thanks for catching that, @Murph33 when you get a sec can we get a spec that runs through this view please.

:+1:

PR #1094, opened 2016-04-22
Bugfix: Remove obsolete ::Spree::ProductScope class.

apepper:
Otherwise rails autoloading will create a module "::Spree::ProductScope", when autoloading core/models/spree/product_scope/scopes.rb

```
> bundle exec rails r "puts ::Spree::ProductScope.class"
> # should output "class"
> # previous to this commit outputed "module"
```

Without this commit the following error can occur:

> ProductScope is not a class
> 
> Extracted source (around line #2):
> 1 module Spree
> 2  class ProductScope < Spree::Base
>
> solidus_core (1.2.2) app/models/spree/product_scope/scopes.rb:2:in `<module:Spree>'

Quote from http://guides.rubyonrails.org/autoloading_and_reloading_constants.html :

> 6.3 Automatic Modules
> When a module acts as a namespace, Rails does not require the application to defines a file for it, a directory matching the namespace is enough.
>
> Suppose an application has a back office whose controllers are stored in app/controllers/admin. If the Admin module is not yet loaded when Admin::UsersController is hit, Rails needs first to autoload the constant Admin.
>
> If autoload_paths has a file called admin.rb Rails is going to load that one, but if there's no such file and a directory called admin is found, Rails creates an empty module and assigns it to the Admin constant on the fly.

__EDIT__

After feedback we agreed to simple remove the file. I also adjusted the title. Previous title: `Bugfix: Create "empty" ::Spree::ProductScope class.`


jhawthorn:
This `ProductScope` situation is a little confusing to me. This is defined in an ActiveRecord class, which is clearly incorrect as there's no tables for it. It also is in a file with the wrong name.

I think we should just remove the file, since it doesn't work and is unused.

cbrunsdon:
Yea I'm for cherry-picking that over. So long as we still get the same commit message.

Anyone that wants it back (which I can't imagine anyone would) can just add it back in themselves to their own project.

apepper:
REVIEW (4dd1e81)

* Removed `core/app/models/spree/product_scope.rb` by reverting my commit
* Removed `core/app/models/spree/product_scope/scopes.rb` by applying https://github.com/spree/spree/pull/5940 / https://github.com/spree/spree/commit/6880c4d80e5797d205a76293d8811861baedc7e5

mamhoff:
In the interest of a clean commit history, would you mind rebasing the two first commits out?

apepper:
REVIEW (82c1307)

* Rewrote branch and force pushed it, to only contain 82c1307.

cbrunsdon:
@apepper fantastic, thanks :+1:

apepper:
Would it be possible to add this to `1.3.0.beta2`?

PR #1096, opened 2016-04-22
Change around SKUS appearance on product page

Murph33:
We wanted to remove the bullets, italics, and the indent.  Did this through minor adjustments to a class we only use in one place and a tweak to the info class so it doesn't italicize in this location. As requested by @Mandily 

## Before
with_bullets

## After
final_changes

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
@@ -128,12 +128,14 @@ button, .button {
 }
 
 span.info {
-  font-style: italic;
   font-size: 85%;
   color: lighten($color-body-text, 15);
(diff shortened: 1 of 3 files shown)


mtomov:
Probably not part of your PR, but linking those to the edit screen on the variant would help a lot. Thanks!

Murph33:
That's definitely a good idea @mtomov and I think it's within scope for this.  After all we're just improving that little part of the page and them being links is much nicer.

mtomov:
It is! Thanks!

mamhoff:
:+1: This is good!

cbrunsdon:
I'm okay with this too, thanks for the change :+1: 

Though I imagine long term we'll want both those styles dead...

PR #1099, opened 2016-04-24
Use safe index changes

bbuchalter:
Some Spree stores migrating to Solidus may have backported these indexes
or used custom names to define these indexes. No need to stop the
migrations from running if they already exist or don't exist.

### DIFF
--- a/core/lib/spree/migration_helpers.rb
+++ b/core/lib/spree/migration_helpers.rb
@@ -1,0 +1,19 @@
+module Spree
+  module MigrationHelpers
+    def safe_remove_index(table, column)
+      remove_index(table, column) if index_exists?(table, column)
+    end
+
+    def safe_add_index(table, column, options = {})
+      if columns_exist?(table, column) && !index_exists?(table, column, options)
+        add_index(table, column, options)
+      end
+    end
+
+    private
+
+    def columns_exist?(table, columns)
+      Array.wrap(columns).all? { |column| column_exists?(table, column) }
+    end
+  end
+end
--- a/core/db/migrate/20140410141842_add_many_missing_indexes.rb
+++ b/core/db/migrate/20140410141842_add_many_missing_indexes.rb
@@ -1,17 +1,19 @@
 class AddManyMissingIndexes < ActiveRecord::Migration
+  include Spree::MigrationHelpers
+
   def change
-    add_index :spree_adjustments, [:adjustable_id, :adjustable_type]
-    add_index :spree_adjustments, :eligible
-    add_index :spree_adjustments, :order_id
-    add_index :spree_promotions, :code
-    add_index :spree_promotions, :expires_at
-    add_index :spree_states, :country_id
-    add_index :spree_stock_items, :deleted_at
-    add_index :spree_option_types, :position
-    add_index :spree_option_values, :position
-    add_index :spree_product_option_types, :option_type_id
-    add_index :spree_product_option_types, :product_id
-    add_index :spree_products_taxons, :position
-    add_index :spree_promotions, :starts_at
+    safe_add_index :spree_adjustments, [:adjustable_id, :adjustable_type]
+    safe_add_index :spree_adjustments, :eligible
+    safe_add_index :spree_adjustments, :order_id
+    safe_add_index :spree_promotions, :code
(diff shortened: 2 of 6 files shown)


bbuchalter:
@jhawthorn here is the second part of changes we made in https://github.com/jhawthorn/solidus/commits/fixup_old_preferences_migration3.

Just for reference, you had already opened one part of it here: https://github.com/solidusio/solidus/pull/1088

We need to still open a third, final PR for this part of the changes: https://github.com/jhawthorn/solidus/commit/c5398f75a2be26f7b298e4506dce9370a3da1cee#diff-fa901dae0e304366993ae2180c352c6fL28, but candidly, I'm not very clear on why that change is needed; I don't think I could write a useful commit message. Could you give it a try?

cbrunsdon:
This looks like a fine idea to me. Nothing worse than a migration that dies half way through for something easily avoidable.

Does this work with the arrays though? 

Since ```safe_add_index :spree_adjustments, [:adjustable_id, :adjustable_type]```
will call
```column_exists?(:spree_adjustments, [:adjustable_id, :adjustable_type])``` which I don't have any reasonable expectation works.

**Comment by bbuchalter** [comment]
@cbrunsdon I believe I've addressed your eagle-eyed concern. Thanks for the feedback.

File: core/lib/spree/migration_helpers.rb [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+      def add_multi_column_index
+        if all_columns_exist? && index_absent?
+          add_index(table, columns, options)
+        end
+      end
+
+      def all_columns_exist?
+        columns.all? { |column| column_exists?(column) }
```
bbuchalter: Note we discard the `options` parameter when checking for the existence of multiple columns. I believe this is a reasonable compromise. What do you think?

core/lib/spree/migration_helpers.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      def add_multi_column_index
+        if all_columns_exist? && index_absent?
+          migration.add_index(table, columns, options)
+        end
+      end
+
+      def all_columns_exist?
+        columns.all? { |column| migration.column_exists?(table, column) }
```
bbuchalter: Note that we drop the options here, although I suspect options would not be relevant in a multi-column index addition?

cbrunsdon: That commit gets pretty fancy, maybe something like: 

maybe something like: 

```ruby
    def safe_add_index(table, columns, options = {})
      if Array.wrap(columns).all? { |column| column_exists?(table, column) } 
        && !index_exists?(table, columns, options)
        add_index(table, columns, options)
      end
   end
```
on the original commit would be easier/cleaner?

or something with cleaner formatting.

bbuchalter: My initial reaction to writing a class for this was just like yours. It does have two benefits:
- The public methods read very nicely.
- Single column index additions get the options parameter passed along to `column_exists?`

If you feel this is too heavy weight for those benefits, I'll go back to the block-inside-conditional style like you've suggested.

cbrunsdon: you can still pass the options to column_exists in the block I pasted above.

Yea, little heavy for me. Happy to have anyone else weigh in, but aye, I'd say kill it for the sake of brevity.

adammathys: I agree with @cbrunsdon, this is a little too heavy. I would prefer the simpler `Array.wrap`, although I would probably just create a private method (instead of in-lining it) and call it a day:

```ruby
def safe_add_index(table, columns, options = {})
  if columns_exist?(table, columns) && !index_exists?(table, columns, options)
    add_index(table, columns, options)
  end
end

private

def columns_exist?(table, columns)
  Array.wrap(columns).all? { |column| column_exists?(table, column) }
end
```

I also don't think including the options in the `column_exits?` call is correct. The options passed in to that will check to see if there's a column with those restrictions. Whereas the options we're passing in are meant to apply to the index associated with that column, not the column itself.

bbuchalter:
One more time, with feeling!

cbrunsdon:
Looks good to me now, thanks :+1:

PR #1104, opened 2016-04-28
Don't stub nonexistant available_shipping_methods

cbrunsdon:
Eric Saupe brought this up in slack, this method does not exist and
should not be stubbed. Was discovered by using verifying doubles in his
local specs.


Senjai:
:+1:

jhawthorn:
remove stub :arrow_right: specs still pass :arrow_right: :+1:

Issue #1087, opened 2016-04-20
Submitting stock transfer form without a source location generates Ruby error


Mandily:
To reproduce:
1. Navigate to Stock > Transfers page, click "New Stock Transfer" button
2. On the New Stock Transfer page, click the continue button
3. View error.

<img width="1241" alt="screen shot 2016-04-20 at 2 20 54 pm" src="https://cloud.githubusercontent.com/assets/15271353/14690966/9e11c002-0703-11e6-9029-0828e773199d.png">

jhawthorn:
This should be fixed by #1105

PR #1105, opened 2016-04-28
Add regression spec + bugfix #1087 Submitting nil stock transfer form 

dangerdogz:
Added regression specs and small fix for #1087 

The fallback method was not updated after a refactoring of the load_stock_locations method


jhawthorn:
:+1: Thank you (especially for the specs)

cbrunsdon:
Yea, looks great, thanks @dangerdogz  :+1:

PR #1108, opened 2016-04-29
Fix spurious spec failure due to DB ordering

jhawthorn:
In #1052, I inadvertently introduced a spurious failure due to DB ordering. It is valid for the DB to return `order.shipments` in any order, so `order.shipments[0]` wasn't always the right shipment.

The error can be seen here: https://circleci.com/gh/solidusio/solidus/3275#artifacts


cbrunsdon:
I'm not sure I prefer this rather than ordering the shipments, but no reason we couldn't do both :+1:

jhawthorn:
I think this is the only correct approach. Even if shipments were ordered by time, this spec shouldn't be dependant on that fact. (also ordering by time have this same problem in MySQL due to timestamp resolution)

PR #1109, opened 2016-05-02
Remove redundant code in Api::ShipmentsController

adaddeo:
The removed code had already been refactored into its own method and is being called in a before_filter,  it was probably left in the create method by accident.


jhawthorn:
:+1: Looks reasonable. Thank you.

adaddeo:
Welcome.

cbrunsdon:
nice catch, thanks :+1:

PR #1117, opened 2016-05-06
Add Prices tab to Products overview

mamhoff:
We have great functionality for dealing with multiple prices for each variant. This is an
absolutely basic interface so that as a store admin, it's possible to edit and update prices.

I'm unsure about allowing deletion and have left it out for now: When you delete a price,
you might end up with no price that has `is_default` set, which will then make the variant
in question become "unavailable". As a new price gets the `is_default` boolean anyways, the old
one becomes so meaningless it's as if it was deleted.

There are plans for a much better interface, but I need something simple that works for 1.3
to get out the door.

### DIFF
--- a/backend/app/views/spree/admin/prices/_table.html.erb
+++ b/backend/app/views/spree/admin/prices/_table.html.erb
@@ -1,3 +1,5 @@
+<%= paginate prices, theme: "solidus_admin" %>
+
 <table class="index prices">
   <colgroup>
     <col style="width: 15%" />
@@ -47,3 +49,5 @@
     <% end %>
   </tbody>
 </table>
+
+<%= paginate prices, theme: "solidus_admin" %>
--- a/backend/app/views/spree/admin/prices/_table.html.erb
+++ b/backend/app/views/spree/admin/prices/_table.html.erb
@@ -23,11 +23,8 @@
   <% prices.each do |price| %>
     <tr id="<%= spree_dom_id price %>" data-hook="prices_row" class="<%= "deleted" if price.deleted? %> <%= cycle('odd', 'even')%>">
       <td class="align-center">
-        <% if price.variant.is_master? %>
-          <%= t(".master_variant") %>
-        <% else %>
-          <%= price.variant.options_text %> </td>
-        <% end %>
+        <%= price.variant.descriptive_name %>
+      </td>
       <td class="align-center">
         <% if price.is_default? %>
           <span class="state complete"> <%= Spree.t(:say_yes) %> </span>
--- a/backend/app/views/spree/admin/prices/_table.html.erb
+++ b/backend/app/views/spree/admin/prices/_table.html.erb
@@ -1,0 +1,49 @@
+<table class="index prices">
+  <colgroup>
+    <col style="width: 15%" />
+    <col style="width: 35%" />
+    <col style="width: 35%" />
+    <col style="width: 15%" />
+  </colgroup>
+
+  <thead data-hook="prices_header">
+    <tr>
+      <th><%= Spree::Variant.model_name.human %> </th>
+      <th><%= Spree::Price.human_attribute_name(:is_default) %></th>
+      <th><%= Spree::Price.human_attribute_name(:amount) %></th>
+      <th><%= Spree::Price.human_attribute_name(:currency) %></th>
+      <th class="actions"></th>
+    </tr>
+  </thead>
+
+  <tbody>
+  <% prices.each do |price| %>
+    <tr data-hook="prices_row" class="<%= "deleted" if price.deleted? %> <%= cycle('odd', 'even')%>">
+      <td class="align-center">
+        <% if price.variant.is_master? %>
+          <%= t(".master_variant") %>
+        <% else %>
+          <%= price.variant.options_text %> </td>
+        <% end %>
+      <td class="align-center">
+        <% if price.is_default? %>
+          <span class="state complete"> <%= Spree.t(:say_yes) %> </span>
+        <% else %>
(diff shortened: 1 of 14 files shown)


mamhoff:
Index view without deletion
solidus administration prices 2016-05-06 17-33-44

Edit view (you can only edit the price itself, for existing prices changing currency or variant is confusing and disabled)
solidus administration prices 2016-05-06 17-34-15

New view (all fields can be edited)
solidus administration prices 2016-05-06 17-34-33

backend/spec/features/admin/products/pricing_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(page).to have_content(product.name)
+      within(".tabs .active") do
+        expect(page).to have_content("Prices")
+      end
+
+      within('table.prices') do
+        expect(page).to have_content(master_price.money.to_s)
+        expect(page).to have_content(master_price.currency)
+        expect(page).to have_content(other_price.money.to_s)
+        expect(page).to have_content(other_price.currency)
```
gmacdougall: Can you please set expectations on the actual value (e.g. 'USD') instead?

backend/spec/controllers/spree/admin/prices_controller_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      let(:product) { create(:product) }
+
+      subject { spree_get :index, product_id: product.slug }
+
+      it { is_expected.to be_success }
+
+      it 'assigns a search variable' do
+        subject
+        expect(assigns(:search)).to be_present
+        expect(assigns(:prices)).to be_present
```
gmacdougall: I would like a better assertion here.

backend/spec/features/admin/products/pricing_spec.rb
mamhoff: Sure, will do.

mamhoff:
@gmacdougall incorporated your feedback, and added pagination.

backend/app/views/spree/admin/prices/_table.html.erb
```diff
@@ -0,0 +1,53 @@
...
+      </td>
+      <td class="align-center"><%= price.money.to_html %></td>
+      <td class="align-center"><%= price.currency %></td>
+      <td class="actions">
+        <% if can?(:update, price) %>
+          <%= link_to_edit(price, :no_text => true) unless price.deleted? %>
+        <% end %>
+        <% if can?(:destroy, price) %>
+          &nbsp;
+          <%#= link_to_delete(price, :no_text => true) unless price.deleted? %>
```
tvdeyen: Is this intentional to disable the delete link here?

backend/app/views/spree/admin/prices/_table.html.erb
```diff
@@ -0,0 +1,53 @@
...
+      <th><%= Spree::Price.human_attribute_name(:is_default) %></th>
+      <th><%= Spree::Price.human_attribute_name(:amount) %></th>
+      <th><%= Spree::Price.human_attribute_name(:currency) %></th>
+      <th class="actions"></th>
+    </tr>
+  </thead>
+
+  <tbody>
+  <% prices.each do |price| %>
+    <tr data-hook="prices_row" class="<%= "deleted" if price.deleted? %> <%= cycle('odd', 'even')%>">
```
tvdeyen: Would be great if we could follow rails convention here and put this into a `_price` partial and render the `@prices` collection.

core/config/locales/en.yml
```diff
@@ -134,6 +134,10 @@ en:
         name: Name
         preference_source: Preference Source
         type: Provider
+      spree/price:
+        currency: Currency
+        amount: Price
+        is_default: Currently Valid
```
tvdeyen: "Currently valid" seems like a strange translation for `is_default` Boolean. I can't imagine what this does. "Currently valid" sounds like something date related. Do we plan to support time based pricing in the future?

backend/app/views/spree/admin/prices/_table.html.erb
mamhoff: Yes it is - there's no route for deletion yet. I should add that I guess :sheep: .

core/config/locales/en.yml
mamhoff: I would love to see time-based pricing. That aside, `is_default` is the most poorly named boolean I've seen so far: It's `true` be default, and adding a new price for a given currency will set all other prices for that currency/variant to `is_default: false`. So "Currently valid" is actually a pretty accurate translation for what it does.

jhawthorn:
:+1: tried this out. More work is needed, but I think it's good enough to merge. We can then collaborate on making it better.

Future steps:
* Validate entered currency (it's possible currently to make a price in an invalid currency, ex. "XYZ")
* Grouping common prices (@Dkendal and @Mandily did some investigation of this)
* We have the master price both on the main product page as well as the pricing page. This will probably be cause of some confusion.
* Showing "Master Variant" in the pricing table feels like we're leaking an implementation detail

tvdeyen:
I also think, this is good enough. How to we ensure to not lose track on the follow up tasks? Maybe a "pricing todo" issue?

jhawthorn:
@tvdeyen done. We can use #1167 as a meta-issue to track these

tvdeyen:
Thanks John.

mamhoff:
#1169 should fix the validation issue. I've amended this PR to use a `select2` instead of a `text` input to make it harder to input invalid data. As for the last point: I've removed the `.master_price` translation from this PR and switched to instead using `Variant#descriptive_name` (which has the same leakage issue, but is already there as model-level logic).

jhawthorn:
Thanks as always Martin

Issue #1074, opened 2016-04-15
Multiple menus open in stock locations section


jhawthorn:
![](http://i.hawth.ca/s/TuhIJfG3.png)

graygilmore:
@jhawthorn would you prefer to see the `Stock Locations` tab removed from `Settings > Shipping`? That makes the most sense to me

graygilmore:
To clarify: and move it under `Stock` in the main navigation.

Mandily:
I agree that it should only be accessible from one place. But putting it in the stock menu item leads me down a rabbit hole of asking, when does something get put in settings, and when does it get put into a parent category of it's own? Playing devil's advocate here - some of those things look like they could go under 'orders'. 

Does anyone have any thoughts on what's a setting and what's a sub section of a related parent?

graygilmore:
I think this would be a good opportunity for us to put or "site architect" hats on and figure out the structure for the entirety of the admin area.

forkata:
Managing stock locations seems like a configuration change to me that wouldn't be done on a day-to-day basis and should probably be limited to "super" admins. The stock screens are something that is probably used on a daily basis by someone like an inventory manager. In my mind there are two ways to think about the organisation of things, one would be per area, i.e. stock, products etc., the other would be based on the frequency of use or permissions required to manage things, i.e. configuration vs. day-to-day operations.

PR #1119, opened 2016-05-09
Don't show multiple active navigation items

graygilmore:
Closes https://github.com/solidusio/solidus/issues/1074

For details check out the [commit message](https://github.com/solidusio/solidus/commit/57bf4ab5b8bd52b428db07ea7cf1df09403c9c97).

### DIFF
--- a/backend/app/views/spree/admin/shared/_settings_sub_menu.html.erb
+++ b/backend/app/views/spree/admin/shared/_settings_sub_menu.html.erb
@@ -21,6 +21,6 @@
   <% end %>
 
   <% if can?(:display, Spree::ShippingMethod) || can?(:display, Spree::ShippingCategory) || can?(:display, Spree::StockLocation) %>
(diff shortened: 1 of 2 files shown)


cbrunsdon:
Yea, this isn't amazing but I'm okay with it for now. We need to figure out how to do this better later.

graygilmore:
I agree. I don't think this is the nicest change but it fixes a pretty bad visual bug. I've create [an issue](https://github.com/solidusio/solidus/issues/1120) for discussing making this area better.

cbrunsdon:
(in case I wasn't clear before I was :+1: on this).

Kevin is also doing similar(ish) things right now with the breadcrumbs, might be a good idea to sit down tomorrow and talk some of this through IRL.

tvdeyen:
Solves the problem, even if it's not perfect. :thumbsup:

jhawthorn:
Thank you

PR #1124, opened 2016-05-09
Revert "Use eval_gemfile instead of plain eval"

adammathys:
This reverts commit 9cde4243667da97be5b524c4a7256d75ef270d40.

Unfortunately, this doesn't work for a number of people due to Bundler's API not being entirely stable. :frowning:


jhawthorn:
:+1:

This was broken by bundler/bundler#4244

jhawthorn:
Clarke gave this an IRL nod

PR #1125, opened 2016-05-09
Make the default pricer country-aware

mamhoff:
This is an addition to the standard pricer class to query for a specified country price first, and then query for a fallback price.

### DIFF
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -27,7 +27,7 @@ module Spree
     money_methods :amount, :price
     alias_method :money, :display_amount
 
-    self.whitelisted_ransackable_attributes = %w( amount variant_id currency )
+    self.whitelisted_ransackable_attributes = %w( amount variant_id currency country_iso )
 
     # An alias for #amount
     def price
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -50,6 +50,14 @@ module Spree
       country_iso.nil?
     end
 
+    def display_country
+      if country_iso
+        "#{country_iso} (#{country.name})"
+      else
+        I18n.t(:any_country, scope: [:spree, :admin, :prices])
+      end
+    end
+
     private
 
     def sum_of_vat_amounts
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -5,13 +5,16 @@ module Spree
     MAXIMUM_AMOUNT = BigDecimal('99_999_999.99')
 
     belongs_to :variant, -> { with_deleted }, class_name: 'Spree::Variant', touch: true
-    has_one :product, class_name: 'Spree::Product', through: :variant
+    belongs_to :country, class_name: "Spree::Country", foreign_key: "country_iso", primary_key: "iso"
+
+    delegate :product, to: :variant
 
     validate :check_price
     validates :amount, allow_nil: true, numericality: {
       greater_than_or_equal_to: 0,
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
+    validates :country, presence: true, if: -> { country_iso.present? }
 
     validates :currency, inclusion: { in: ::Money::Currency.all.map(&:iso_code), message: :invalid_code }
 
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -14,11 +14,12 @@ module Spree
       greater_than_or_equal_to: 0,
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
-    validates :country, presence: true, if: -> { country_iso.present? }
+    validates :country, presence: true, unless: -> { for_any_country? }
 
     validates :currency, inclusion: { in: ::Money::Currency.all.map(&:iso_code), message: :invalid_code }
 
     scope :currently_valid, -> { where(is_default: true) }
+    scope :for_any_country, -> { where(country: nil) }
     scope :with_default_attributes, -> { where(Spree::Config.default_pricing_options.desired_attributes) }
     after_save :set_default_price
 
@@ -42,6 +43,10 @@ module Spree
     end
 
     private
+
+    def for_any_country?
+      country_iso.nil?
+    end
 
     def check_price
       self.currency ||= Spree::Config[:currency]
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -18,7 +18,7 @@ module Spree
     validates :currency, inclusion: { in: ::Money::Currency.all.map(&:iso_code), message: :invalid_code }
     validates :country, presence: true, unless: -> { for_any_country? }
 
-    scope :currently_valid, -> { where(is_default: true) }
+    scope :currently_valid, -> { where(is_default: true).order("country_iso IS NULL") }
     scope :for_any_country, -> { where(country: nil) }
     scope :with_default_attributes, -> { where(Spree::Config.default_pricing_options.desired_attributes) }
     after_save :set_default_price
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -8,15 +8,15 @@ module Spree
     belongs_to :country, class_name: "Spree::Country", foreign_key: "country_iso", primary_key: "iso"
 
     delegate :product, to: :variant
+    delegate :tax_rates, to: :variant
 
     validate :check_price
     validates :amount, allow_nil: true, numericality: {
       greater_than_or_equal_to: 0,
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
-    validates :country, presence: true, unless: -> { for_any_country? }
-
     validates :currency, inclusion: { in: ::Money::Currency.all.map(&:iso_code), message: :invalid_code }
+    validates :country, presence: true, unless: -> { for_any_country? }
 
     scope :currently_valid, -> { where(is_default: true) }
     scope :for_any_country, -> { where(country: nil) }
@@ -43,7 +43,7 @@ module Spree
     end
 
     def net_amount
-      amount / (1 + sum_of_included_vats)
+      amount / (1 + sum_of_vat_amounts)
     end
 
     def for_any_country?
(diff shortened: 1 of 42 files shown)


core/app/models/spree/tax/tax_location.rb
```diff
@@ -25,6 +25,10 @@ def ==(other)
         state_id == other.state_id && country_id == other.country_id
       end
 
+      def country
+        Spree::Country.find_by(id: country_id)
```
bbuchalter: Should we consider memoizing this?

core/app/models/spree/variant/pricer.rb
```diff
@@ -12,7 +12,12 @@ def initialize(variant)
       end
 
       def price_for(price_options)
-        variant.prices.currently_valid.find_by(price_options.desired_attributes).try!(:money)
+        variant.prices
```
bbuchalter: Some docs for this class and this method would be very welcome.

core/app/models/spree/variant/pricing_options.rb
```diff
@@ -19,12 +22,38 @@ def initialize(desired_attributes = {})
         @desired_attributes = self.class.default_price_attributes.merge(desired_attributes)
       end
 
+      def arel_conditions
```
bbuchalter: I believe some docs for this new public method would be helpful.

core/app/models/spree/price.rb
```diff
@@ -5,15 +5,18 @@ class Price < Spree::Base
...
 
     validate :check_price
     validates :amount, allow_nil: true, numericality: {
       greater_than_or_equal_to: 0,
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
+    validates :country, presence: true, unless: -> { fallback? }
 
     scope :currently_valid, -> { where(is_default: true) }
-    scope :with_default_attributes, -> { where(Spree::Config.default_pricing_options.desired_attributes) }
```
bbuchalter: There is another method in this file which references `desired_attributes`: https://github.com/magiclabs/solidus/blob/b8d5bbaea0131abfbeb0c57d31dbe07b2a4a0cdf/core/app/models/spree/price.rb#L65

Does that need to be updated as well?

bbuchalter: There is also this method using `desired_attributes`: https://github.com/magiclabs/solidus/blob/b8d5bbaea0131abfbeb0c57d31dbe07b2a4a0cdf/core/app/models/concerns/spree/default_price.rb#L15

core/app/models/spree/tax/tax_location.rb
mamhoff: I wouldn't do it. The `Spree::Tax::TaxLocation` is usually a pretty throwaway object. The big reason why we need this method is that countries are referenced by ID rather that ISO code (in which case I could just use that and not use a DB lookup). 

Also: This query can return `nil`, in which case memoizing won't help.

core/app/models/spree/variant/pricer.rb
mamhoff: Absolutely, working on it.

core/app/models/spree/variant/pricing_options.rb
mamhoff: I'm considering naming it something else as it's anything that can be used as a `XXXX` to `variant.prices.where(XXXX)`. For the case we're looking at here, it could even be a hash.

core/app/models/spree/price.rb
mamhoff: No, both of these need the exact `desired_attributes` hash, not something that only `where` can use.

core/app/models/spree/price.rb
```diff
@@ -35,8 +38,24 @@ def price=(price)
...
+    def export?
+      country_iso.nil?
+    end
+    alias_method :fallback?, :export?
+
     private
 
+    def sum_of_included_vats
+      return 0 unless variant.tax_category
+      variant.tax_category.tax_rates.for_address(
```
mamhoff:
```
has_many :tax_rates
has_many :vat_rates, -> { where(included_in_price: true) } 
```

core/app/models/spree/price.rb
```diff
@@ -35,8 +38,24 @@ def price=(price)
...
+    end
+
+    def export?
+      country_iso.nil?
+    end
+    alias_method :fallback?, :export?
+
     private
 
+    def sum_of_included_vats
```
mamhoff: This is two times the same thing

core/app/models/spree/price.rb
```diff
@@ -35,8 +38,24 @@ def price=(price)
...
     end
 
+    def net_amount
+      amount / (1 + sum_of_included_vats)
+    end
+
+    def export?
+      country_iso.nil?
+    end
+    alias_method :fallback?, :export?
```
mamhoff: Remove the `fallback` naming

core/app/models/spree/price.rb
```diff
@@ -5,14 +5,17 @@ class Price < Spree::Base
...
 
     validate :check_price
     validates :amount, allow_nil: true, numericality: {
       greater_than_or_equal_to: 0,
       less_than_or_equal_to: MAXIMUM_AMOUNT
     }
+    validates :country, presence: true, unless: -> { fallback? }
 
     scope :currently_valid, -> { where(is_default: true) }
+    scope :fallback, -> { where(country: nil) }
```
mamhoff: `for_any_country`

core/app/models/spree/variant.rb
```diff
@@ -48,6 +48,7 @@ class Variant < Spree::Base
     before_validation :set_cost_currency
 
     validate :check_price
+    validate :build_vat_prices
```
mamhoff: Both of these should be `before_validate` as they actually set things.

core/app/models/spree/variant/price_generator.rb
```diff
@@ -0,0 +1,56 @@
...
+    # export (net) price for any country that doesn't have VAT.
+    # @example
+    #   The admin views German gross prices (Spree::Config.admin_vat_country_iso == "DE")
+    #   There is VATs configured for Germany (19%) and Finland (24%).
+    #   The price generator is run on a variant with a base (German) price of 10.00.
+    #   The Price Generator will generate:
+    #     - A price for Finland (country_id == "FI"): 10.42
+    #     - A price for export (country_id == nil): 8.40
+    #
+    class PriceGenerator
```
mamhoff: `VatPriceGenerator`

core/app/models/spree/variant/price_generator.rb
```diff
@@ -0,0 +1,56 @@
...
+      attr_reader :variant
+
+      def initialize(variant)
+        @variant = variant
+      end
+
+      def run
+        country_isos_requiring_price.each do |country_iso|
```
mamhoff: Add an early exit for sales tax stores.

core/app/models/spree/variant/pricer.rb
```diff
@@ -1,6 +1,15 @@
 module Spree
   class Variant
+    # This class is responsible for assigning a price to a variant.
```
mamhoff: ...for choosing the price of a variant when displaying or adding it to a cart.

core/app/models/spree/variant/pricer.rb
```diff
@@ -1,6 +1,15 @@
...
+    # A variant can have multiple or even dynamic prices. The `price_for`
+    # method determines which price applies under the given circumstances.
+    #
     class Pricer
```
mamhoff: Thomas says: Maybe another name that hints more at *choosing* or *retrieving* prices

mamhoff:
It's done - pending positive code reviews. 

This PR finally generates correct prices including VAT that can be changed by a backend user, and gets rid of the default zone. Much of the changed lines are specs for migrating an existing VAT store and class renames. Most notable is the class `Spree::Variant::PriceSelector`, which used to be called `Spree::Variant::Pricer` - that old name was confusing when trying to differentiate from `Spree::Variant::PriceGenerator` (which could have been named "Pricer", too). 

The interface for prices is still a bit rough around the edges, as #1167 testifies. For now, I believe it's absolutely usable, though.

cbrunsdon:
Quick warning from your friendly neighborhood core team! We're okay with this now in the "We can't see anything wrong or can suggest any better ways to do any of it", so unless we get any kind of feedback in the next 24 hours we're going to merge it!

jhawthorn:
:tada: :boom:

bbuchalter:
Cheers on this milestone PR!

core/app/models/spree/variant/pricing_options.rb
```diff
@@ -1,30 +1,95 @@
...
+      # new pricing options.
+      # @see Spree::Config.default_pricing_options
+      # @see #initialize
+      # @return [Hash] The attributes that admin prices usually have
+      #
       def self.default_price_attributes
-        { currency: Spree::Config.currency }
+        {
+          currency: Spree::Config.currency,
+          country_iso: Spree::Config.admin_vat_country_iso
```
flyfy1: why the `admin_vat_country_iso` is used here, rather than `default_country_iso`?

core/app/models/spree/variant/vat_price_generator.rb
```diff
@@ -0,0 +1,58 @@
...
+          foreign_price.amount = variant.default_price.net_amount * (1 + vat_for_country_iso(country_iso))
+        end
+      end
+
+      private
+
+      # nil is added to the array so we always have an export price.
+      def country_isos_requiring_price
+        return [nil] unless variant.tax_category
+        [nil] + variant_vat_rates.map(&:zone).flat_map(&:countries).flat_map(&:iso)
```
flyfy1: why do we need this `[nil]` (for all country)? What's an export price?

core/app/models/spree/variant/pricing_options.rb
mamhoff: Because when you create a new price for an item, you're most likely doing that in the backend as an admin, and you'll expect the new price to be with the VAT the other prices in the admin area. The default country iso is the one used in the frontend.

core/app/models/spree/variant/vat_price_generator.rb
mamhoff: An export price is a price without VAT, as in countries with VAT, if you export to a country where no VAT applies, you have to change the price to exclude VAT. We add `nil` because a price with `nil` is used for shipping to any country we have no information about. Otherwise we'd need to create a price for each country, which is redundant and tedious to work with.

PR #1126, opened 2016-05-10
Introduce Spree::Promotion.scheduled_after scope

stewart:
This commit introduces a new scope for the Spree::Promotion model,
`scheduled_after(time)`. This scope is useful for finding all promotions
scheduled to start after the provided time.


cbrunsdon:
@stewart is this going to be used somewhere in the codebase or is the point just add it for general support?

stewart:
General support, and possibly as part of some of the promotions admin changes, although I'd likely be using an instance method there, as all of the promotions are already loaded by that point. If this doesn't seem useful to the majority of people, I'm happy to close the PR.

PR #1128, opened 2016-05-10
Fix association to point to right objects

adammathys:
Having payment methods point to themselves doesn't really help any. Let's actually point the association at the object it should point to.


cbrunsdon:
:+1:

mamhoff:
👍

mamhoff:
Thank you!

PR #1134, opened 2016-05-10
Rename Spree::Promotion#expired? to Spree::Promotion#inactive?

stewart:
The new name better describes what the method is doing, and avoids confusion when calling it on promotions that have not run yet.


mamhoff:
👍 Thanks!

jhawthorn:
:+1:

PR #1135, opened 2016-05-10
Associate trackers to stores

adammathys:
This functionality was originally provided by [`solidus_multi_domain`](https://github.com/solidusio/solidus_multi_domain) but is just as valid as part of core. This is basically a continuation of the work from https://github.com/solidusio/solidus/commit/2fdfae5c5f7e1abba1f4b91d8dd5640d576a5b08.


mamhoff:
Is there tests for this functionality in `spree_multi_domain`? If so, it'd be good to move those over as well. If there aren't, I'd suggest actually writing them, because this is a bunch of functionality that deserves to be specced out.

PR #1136, opened 2016-05-10
Add one-click Heroku demo button

alexblackie:
Add a button for deploying the sandbox store to Heroku. This is made
possible by a CI job and a [separate repo][1].

[1]: https://github.com/solidusio/sandbox-demo

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,8 @@ custom frontend instead of using [`solidus_frontend`](https://github.com/solidus
 Getting started
 ---------------
 
+[![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy?template=https://github.com/solidusio/sandbox-demo)
+
 To add solidus, begin with a rails 4.2 application. Add the following to your
 Gemfile.
 


mamhoff:
👍

tvdeyen:
This is nice. Can we rename the button? "Deploy a demo" or simply "Demo" would be much clearer.

alexblackie:
I think we're stuck with Heroku's, unless we want to make our own image. A nice middle-ground could perhaps be some explanatory text around it.

> Want to try out Solidus? Deploy a demo store: [Button]

bbuchalter:
Where can the CI job be reviewed?

alexblackie:
@bbuchalter Currently it's living on the Stembolt Jenkins, which is not ideal (but was easiest and fastest). Will talk IRL with @jhawthorn on Monday to see if we can get it in Travis or something.

jhawthorn:
The [jenkins job](https://gist.github.com/jhawthorn/8d83d8581dc1199e7a044ad58e22f035) is too horrible to share in full (sorry alex), but the basics are that it clones down solidus master, runs a `rake sandbox`, and copies any changes over top of the [sandbox-demo](https://github.com/solidusio/sandbox-demo) repo.

This is our first shot at this. I think we can do better. What I really want is to have a buildpack which is responsible for running `rake sandbox` and saving the result as a slug, but this is a bit tricky to do (the heroku buildpacks really want an existing app). I do still really want to do this, but this hack was the easiest way to get started. I'd like to merge this as is because having a demo is better than none.

Something we can work on with this current demo hack that will carry forward is the [app.json](https://github.com/solidusio/sandbox-demo/blob/master/app.json) file. Which we should be able to use to configure the admin's email and password.

@alexblackie could we get some nicer text like @tvdeyen suggested.

jhawthorn:
Merged as #1173

Issue #1116, opened 2016-05-06
Remove order merge logic happening on every request


mtomov:
Hello,

[set_current_order](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/controller_helpers/order.rb#L60) is a method, which is called on [every request](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/controller_helpers/order.rb#L11) and results in 3 sql queries for logged in users:

``` sql
  Spree::Order Load (0.7ms)  SELECT  "spree_orders".* FROM "spree_orders" WHERE "spree_orders"."completed_at" IS NULL AND "spree_orders"."currency" = $1 AND "spree_orders"."guest_token" = $2 AND "spree_orders"."store_id" = $3 AND "spree_orders"."user_id" = $4 LIMIT 1  [["currency", "USD"], ["guest_token", "6YHZ-S_2WypaPxxky4fHVA"], ["store_id", 2], ["user_id", 1]]
  Spree::Order Load (0.4ms)  SELECT  "spree_orders".* FROM "spree_orders" WHERE "spree_orders"."user_id" = $1 AND "spree_orders"."frontend_viewable" = $2 AND "spree_orders"."store_id" = 2  ORDER BY "spree_orders"."created_at" DESC LIMIT 1  [["user_id", 1], ["frontend_viewable", "t"]]
  Spree::Order Load (0.3ms)  SELECT "spree_orders".* FROM "spree_orders" WHERE "spree_orders"."user_id" = $1 AND "spree_orders"."completed_at" IS NULL AND (id != 9)  [["user_id", 1]]


```

![spree orders - 2 queries](https://cloud.githubusercontent.com/assets/1651750/15068768/720a75ec-1371-11e6-9f22-6668cb8eef88.png)

The method performs an order merge operation if there's more than one incomplete order for the user. 

Unless I miss a use case, this _correct behaviour_ is have this operation be performed only after a user login - exactly in [UserSessionsController#create](https://github.com/solidusio/solidus_auth_devise/blob/master/lib/controllers/frontend/spree/user_sessions_controller.rb#L15). 

It is currently incorrectly executed there in a `before_action` instead of an `after_action` with the inclusion of the [`Order`](https://github.com/spree/spree_auth_devise/blob/master/lib/controllers/frontend/spree/user_sessions_controller.rb#L6) module.
## 

<del>
[...]

PR #1137, opened 2016-05-10
Remove extra set_current_order calls

Murph33:
Currently this method set_current_order is called on almost every request in the front end resulting in many unnecessary SQL requests.  We want to push this to where it should be which is just when we sign in to authentication systems.  This is currently done in solidus_auth_devise, 
https://github.com/solidusio/solidus_auth_devise/pull/51 ,  and is clarified in the patch notes for 1.4 so people using other gems can see this and address it.  This addresses #1116.

The one concern is that we need to make sure people use a recent enough version of Solidus Auth to match this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@
 
 ## Solidus 1.4.0 (unreleased)
 
+*   Order Merging on Sign In:
+
+    Previously a before filter added in `core/lib/spree/core/controller_helpers/order.rb`
+    would cause sql queries to be used on almost every request in the frontend.
+    If you do not use Solidus Auth you will need to hook into this helper and
+    call `set_current_order` where your user signs in. This merges incomplete
+    orders a user has going with their current cart.  If you do use
+    Solidus Auth you will need to make sure you use a current enough
+    version that includes this explicit call. If you do use Solidus Auth but do _not_
+    use Solidus Frontend you will also need to hook into this helper and call
+    `set_current_order` where your user signs in.
+
 *   Use in-memory objects in OrderUpdater and related areas.
 
     Solidus now uses in-memory data for updating orders in and around
--- a/api/app/controllers/spree/api/checkouts_controller.rb
+++ b/api/app/controllers/spree/api/checkouts_controller.rb
@@ -11,9 +11,6 @@ module Spree
       # TODO: Remove this after deprecated usage in #update is removed
       include Spree::Core::ControllerHelpers::PaymentParameters
 
-      # This before_action comes from Spree::Core::ControllerHelpers::Order
-      skip_before_action :set_current_order
-
       def next
         if @order.confirm?
           Spree::Deprecation.warn "Using Spree::Api::CheckoutsController#next to transition to complete is deprecated. Please use #complete instead of #next.", caller
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -8,8 +8,6 @@ module Spree
         include ControllerHelpers::Pricing
 
         included do
-          before_action :set_current_order
-
           helper_method :current_order
           helper_method :simple_current_order
         end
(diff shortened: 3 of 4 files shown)


mtomov:
Wow, thanks so much! I think it's really good what you've done both here and in the matching PR in `solidus_auth_devise`. 

Thanks a lot again!

mamhoff:
👍 Super!

CHANGELOG.md
```diff
@@ -17,6 +17,16 @@
...
 
+    #### Order Merging on Sign In:
+
+    Previously a before filter added in `core/lib/spree/core/controller_helpers/order.rb`
+    would cause sql queries to be used on almost every request in the frontend.
+    If you do not use Solidus Auth you will need to hook into this helper and
+    call `set_current_order` where your user signs in. This merges incomplete
+    orders a user has going with their current cart.  If you do use
+    Solidus Auth you will need to make sure you use a current enough
+    version that includes this explicit call.
```
bbuchalter: An additional note suggesting the following is likely very thoughtful:

> If you do not use Solidus Auth *or* you use Solidus Auth but do not use Solidus Frontend, you will need to restore order merging functionality yourself.

Murph33: Could someone correct me if I'm wrong but I was fairly certain that the order merging logic is all contained in Core and was just unnecessarily being called in the front end.  After this change the order merging should just happen on the sign in which should be handled by the authentication system a user chooses.  I will change it to clarify that calling `set_current_order` will cause order merging

CHANGELOG.md
```diff
@@ -17,6 +17,16 @@
     cart. You can set it through the new `Spree::Config.variant_pricer_class` setting. This
     class also knows which `PricingOptions` class it cooperates with.
 
+    #### Order Merging on Sign In:
```
jhawthorn: This will not be in 1.3, could you move this to the 1.4 section

CHANGELOG.md
bbuchalter: Hi @Murph33 I'm sorry I haven't come back to this thread, but I still think this is an open question. Perhaps I misunderstood the nature of this change. Is [this](https://github.com/solidusio/solidus_auth_devise/blob/master/config/initializers/warden.rb#L2) where order merging is done?

Murph33: No, that is where open orders, I believe, are associated with the signing in user in order to keep them from being lost.  Orders will now be merged [here](https://github.com/solidusio/solidus_auth_devise/pull/51/files) in solidus auth devise.  Something similar will need to be implemented in other authentication systems.

bbuchalter: Thus, we need to update this changelog entry. Right?

Murph33: I could change my entry to instruct them they could also just emulate the `set_current_order` logic themselves instead of hooking into the helper and using it if you would like, but I still don't see where Solidus Frontend comes into play.  The merging logic used to be called there but this is removing that.

bbuchalter:
👍

peterberkenbosch:
@Murph33 would you mind rebasing this one on master please? This should be okay to merge then @jhawthorn

Murph33:
Rebased on master, thanks @peterberkenbosch

peterberkenbosch:
Great! Thanks @Murph33! @jhawthorn ready to merge this from my point of view 👍

mtomov:
Hello, it will be great to have this merged in, ideally in 1.x, but I would understand if that is not ideal. Thank you!

mtomov:
Hello, do you think we can merge this one in now?

tvdeyen:
https://github.com/solidusio/solidus_auth_devise/pull/51 was merged, so this could be merged as well. Need an update to the CHANGELOG, though. We do not plan to back port this into 1.x

@Murph33 would you update the CHANGELOG or should we do this?

mtomov:
Ping : )

mtomov:
pong

swcraig:
I have updated the CHANGELOG and rebased onto master in #2185.

swcraig:
May we close this? #2185 was merged.

mamhoff:
Closing, thank you @swcraig

PR #1139, opened 2016-05-11
Improve reimbursement emails

kennyadsl:
This PR adds some improvements to the reimbursement email:

- it allows testing emails like other ones with the rails preview. To watch and work with this preview a real reimbursement (persisted in the database) is needed so it adds a reimbursement sample that is loaded with the `spree_sample:load` task and attached to one order previously created.
- it adds a basic html version of the reimbursement email.

I'm open to squash commits or discuss any change or improvement.

### DIFF
--- a/core/app/mailers/spree/reimbursement_mailer.rb
+++ b/core/app/mailers/spree/reimbursement_mailer.rb
@@ -1,11 +1,11 @@
 module Spree
   class ReimbursementMailer < BaseMailer
-      def reimbursement_email(reimbursement, resend = false)
-        @reimbursement = reimbursement.respond_to?(:id) ? reimbursement : Spree::Reimbursement.find(reimbursement)
-        store = @reimbursement.order.store
-        subject = (resend ? "[#{Spree.t(:resend).upcase}] " : '')
-        subject += "#{store.name} #{Spree.t('reimbursement_mailer.reimbursement_email.subject')} ##{@reimbursement.order.number}"
-        mail(to: @reimbursement.order.email, from: from_address(store), subject: subject)
-      end
+    def reimbursement_email(reimbursement, resend = false)
+      @reimbursement = reimbursement.respond_to?(:id) ? reimbursement : Spree::Reimbursement.find(reimbursement)
+      store = @reimbursement.order.store
+      subject = (resend ? "[#{Spree.t(:resend).upcase}] " : '')
+      subject += "#{store.name} #{Spree.t('reimbursement_mailer.reimbursement_email.subject')} ##{@reimbursement.order.number}"
+      mail(to: @reimbursement.order.email, from: from_address(store), subject: subject)
+    end
   end
 end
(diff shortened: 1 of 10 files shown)


kennyadsl:
What about to also adding a preference for Reimbursement Mailer class, similar to [Spree::CartonMailer](https://github.com/nebulab/solidus/blob/master/core/app/models/spree/app_configuration.rb#L331-L340) ?

core/app/mailers/spree/reimbursement_mailer.rb
```diff
@@ -1,11 +1,11 @@
...
-        subject += "#{store.name} #{Spree.t('reimbursement_mailer.reimbursement_email.subject')} ##{@reimbursement.order.number}"
-        mail(to: @reimbursement.order.email, from: from_address(store), subject: subject)
-      end
+    def reimbursement_email(reimbursement, resend = false)
+      @reimbursement = reimbursement.respond_to?(:id) ? reimbursement : Spree::Reimbursement.find(reimbursement)
+      store = @reimbursement.order.store
+      subject = (resend ? "[#{Spree.t(:resend).upcase}] " : '')
+      subject += "#{store.name} #{Spree.t('reimbursement_mailer.reimbursement_email.subject')} ##{@reimbursement.order.number}"
+      mail(to: @reimbursement.order.email, from: from_address(store), subject: subject)
+    end
```
mamhoff: Can you PR this in a separate PR that just fixes the indentation? I'm good with the change, just don't want to mix it up with some of the other stuff in this PR.

core/app/views/spree/reimbursement_mailer/reimbursement_email.text.erb
```diff
@@ -12,7 +12,7 @@
 <%= Spree.t('reimbursement_mailer.reimbursement_email.exchange_summary') %>
 ============================================================
 <% @reimbursement.return_items.exchange_requested.each do |return_item| %>
-<%= return_item.variant.sku %> <%= raw(return_item.variant.product.name) %> <%= "(#{raw(return_item.variant.options_text)})" if return_item.variant.options_text.present? -%> -> <%= return_item.exchange_variant.sku %> <%= raw(return_item.exchange_variant.product.name) if return_item.exchange_variant.options_text.present? %> <%= "(#{raw(return_item.exchange_variant.options_text)})" -%>
```
mamhoff: I don't understand why the changes you made are here. Does the view work without those changes?

mamhoff:
I would like this to be split up into one PR for fixing indentation, one for adding the HTML view and preview, and one for changing the Text version of the E-Mail.

core/app/views/spree/reimbursement_mailer/reimbursement_email.text.erb
kennyadsl: these are the changes made to this file:

1. removed `-` char at the end of some erb tags. They are useless and not used the in other tags;
2. when `return_item.variant.product.name` was used, it has been changed in `return_item.variant.name` since [that method is delegated from variant to its product](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/variant.rb#L25-L27);
3. in this portion: `<%= raw(return_item.exchange_variant.product.name) if return_item.exchange_variant.options_text.present? %>` the `if` condition is checking the presence of `option_text` but it's applied on the `name`. It had to be moved to the `option_text` tag. This is actually a bug since without this change the mail text will not have the name printed if the `option_text` is not present on that variant.

kennyadsl:
Closing in favor of:

- #1144 
- #1145 
- #1146 
- #1147

PR #1141, opened 2016-05-11
Add a spec for #transfer_to_location to verify the attributes of the new shipment

hectoregm:
When a shipment split is done the new shipment created by #transfer_to_location doesn't associate
the order shipping address to this new shipment so all new shipments have nil for address.

### DIFF
--- a/core/spec/models/spree/shipment_spec.rb
+++ b/core/spec/models/spree/shipment_spec.rb
@@ -24,6 +24,26 @@ describe Spree::Shipment, type: :model do
   let(:variant) { mock_model(Spree::Variant) }
   let(:line_item) { mock_model(Spree::LineItem, variant: variant) }
 
+  context '#transfer_to_location' do
+    it 'transfers unit to a new shipment with given location' do
+      order = create(:completed_order_with_totals, line_items_count: 2)
+      shipment = order.shipments.first
+      variant = order.inventory_units.map(&:variant).first
+
+      aggregate_failures("verifying new shipment attributes") do
+        expect do
+          shipment.transfer_to_location(variant, 1, stock_location)
+        end.to change { Spree::Shipment.count }.by(1)
+
+        new_shipment = order.shipments.last
+        expect(new_shipment.number).to_not eq(shipment.number)
+        expect(new_shipment.stock_location).to eq(stock_location)
+        expect(new_shipment.line_items.count).to eq(1)
+        expect(new_shipment.line_items.first.variant).to eq(variant)
+      end
+    end
+  end
+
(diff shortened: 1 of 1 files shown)


hectoregm:
The main problem is that splitting a completed order doesn't fill the address and the shipping method
for the new shipment, this later raises an error when doing a transition to shipped:

```
ActiveRecord::RecordInvalid: Validation of Address can't be blank, Shipping method can't be blank.
```


https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_shipping.rb#L51

when creating a Spree::Carton called on the after_transition hook for shipped.

mamhoff:
Thanks, this makes sense to me. @forkata Any idea why this was omitted in the first place? 

Btw really nice specs.

forkata:
@mamhoff It appears that the address association on shipment isn't used consistently and this was probably omitted by mistake. I looked through the history of this code and as far back as I could trace it, it seems this would have been an issue https://github.com/solidusio/solidus/commit/b252e9adcd02b5b0f0f89cdcfae583e72ae12fb7.

Related to this there is a proposed change to completely remove the address association on shipment here https://github.com/solidusio/solidus/pull/1138 and always use the `Order#ship_address` instead. If we decide to go down that direction, then this change may no longer be needed.

hectoregm:
@forkata yeah I read the proposal to delegate to `Order#ship_address` but I think we can use this spec
for `#trasnfer_for_location` right ?, I could remove address expectation and the injection of address.

forkata:
@hectoregm We can definitely use the spec, thank you for adding that. I would keep an eye on what the decision is regarding the `address` association on `Shipment` and make any changes based on that if needed.

core/spec/models/spree/shipment_spec.rb
```diff
@@ -24,6 +24,25 @@
   let(:variant) { mock_model(Spree::Variant) }
   let(:line_item) { mock_model(Spree::LineItem, variant: variant) }
 
+  context '#transfer_to_location' do
+    it 'transfers unit to a new shipment with given location' do
+      order = create(:completed_order_with_totals, line_items_count: 2)
+      shipment = order.shipments.first
+      variant = order.inventory_units.collect(&:variant).first
+
+      expect do
```
forkata: It would be nice if we can wrap all these expectations in an `aggregate_failures` block so we can see multiple failures in a single run.

http://www.rubydoc.info/github/rspec/rspec-expectations/RSpec%2FMatchers%3Aaggregate_failures

hectoregm:
@forkata made the changes we mentioned and added aggregate_failures block

jhawthorn:
:+1:

PR #1142, opened 2016-05-12
Use native Rails' `routes.draw` instead of custom `draw_routes`

dholdren:
* Solidus manages it's own route list, but Rails' RoutSet should be good enough.

### DIFF
--- a/api/config/routes.rb
+++ b/api/config/routes.rb
@@ -1,4 +1,4 @@
-Spree::Core::Engine.add_routes do
+Spree::Core::Engine.routes.draw do
   namespace :admin do
     resources :users do
       member do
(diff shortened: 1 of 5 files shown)


cbrunsdon:
This looks good to me, thanks, but I wasn't really clear on why we needed this in the first place. @jhawthorn?

For record I pulled it down into the sandbox, edited routes, and everything worked as expected.

jhawthorn:
The original reason this was added was related to rails/rails#12367, we'll have to see if that is still an issue in extensions.

jhawthorn:
@dholdren Thanks for your patience 

I tried this branch on a few extensions and couldn't spot any changes or issues :shipit: . Hopefully whatever this code was working around was fixed at some point since rails 4.0.

:+1: This is great. The backwards compatibility with deprecation warnings are especially appreciated.

PR #1144, opened 2016-05-14
Fix reimbursement mailer text

kennyadsl:
1. removed '-' char at the end of some erb tags. They are useless and
not used the in other tags;
2. when `return_item.variant.product.name` was used, it has been changed
in `return_item.variant.name` since that method is delegated from
variant to its product;
3. in this portion: `<%= raw(return_item.exchange_variant.product.name)
if return_item.exchange_variant.options_text.present? %>` the if
condition is checking the presence of `option_text` but it's applied on
the `name`. It had to be moved to the `option_text` tag. This is
actually a bug since without this change the mail text will not have the
name printed if the `option_text` is not present on that variant. 

The previous (wrong) result was:

`ROR-00012 Ruby on Rails Bag  -> ROR-00012  ()`

while the current (expected) result is:

`ROR-00012 Ruby on Rails Bag  -> ROR-00012 Ruby on Rails Bag`


mamhoff:
:+1: super!

jhawthorn:
:+1:

PR #1146, opened 2016-05-14
Add reimbursement email preview

kennyadsl:
It allows testing emails like other ones with the rails preview. To watch and work with this preview a real reimbursement (persisted in the database) is needed so it adds a reimbursement sample that is loaded with the spree_sample:load task and attached to one order previously created.


mamhoff:
:+1:

jhawthorn:
:+1: Thank you.

Having the sample as well is very helpful

PR #1151, opened 2016-05-16
Make product filters more configurable

mtylty:
Add `Spree::Config.search_filter_presenter_class` configuration variable
so that product filters (those that use the searcher_class to filter
product results) are more configurable.

This makes filtering more configurable because:
- `Spree::Config.search_filter_presenter_class` can be used to define which
  filters are presented on the view;
- The `partial_path` class method can be used to define which partial is
  rendered by the above views so that custom view logic can be implemented
  depending on the `searcher_class` and `search_filter_presenter_class`
  configurations.

I've made this PR as a proof-of-concept for what I think the product
filtering should look like in future releases.

What's your opinion on this?


core/app/models/spree/taxon.rb
```diff
@@ -35,6 +35,7 @@ class Taxon < Spree::Base
       content_type: { content_type: ["image/jpg", "image/jpeg", "image/png", "image/gif"] }
 
     # @note This method is meant to be overridden on a store by store basis.
+    # @deprecated - use Spree::Config.search_filter_presenter_class instead
     # @return [Array] filters that should be used for a taxon
     def applicable_filters
```
kennyadsl: Not sure what's the best way but probably it could be a good idea to change this method content to use the new class. If someone was using/overriding `taxon.applicable_filters` the interface would be the same and it should continue to work but under the hood it would start using the new `Spree::Core::Search::ProductFilter` which does exactly the same things.

core/lib/spree/core/search/product_filter.rb
```diff
@@ -0,0 +1,27 @@
+module Spree
+  module Core
+    module Search
+      class ProductFilter
```
kennyadsl: maybe `ProductFilters` if it returns a list of filters? or maybe a public method `all` or `available` which return the content of `@filters`?

core/lib/spree/core/search/product_filter.rb
```diff
@@ -0,0 +1,27 @@
...
+        attr_reader :filters
+
+        def initialize(taxon = nil)
+          @filters = []
+
+          if taxon.nil?
+            @filters << Spree::Core::ProductFilters.all_taxons
+          else
+            # @filters << Spree::Core::ProductFilters.taxons_below(taxon)
```
cbrunsdon: We can strip these comments now, I don't think they add any value

core/lib/spree/core/search/product_filter.rb
```diff
@@ -0,0 +1,27 @@
...
+            ## unless it's a root taxon? left open for demo purposes
+
+            @filters << Spree::Core::ProductFilters.price_filter
+            @filters << Spree::Core::ProductFilters.brand_filter
+          end
+        end
+
+        def self.partial_path
```
cbrunsdon: Rather than this being a static method on the class I'd really like it to be an instance, as it would allow the filter to potentially render different sidebars depending on its filter logic.

frontend/app/views/spree/products/index.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for :sidebar do %>
   <div data-hook="homepage_sidebar_navigation">
     <% if "spree/products" == params[:controller] && @taxon %>
-      <%= render :partial => 'spree/shared/filters' %>
+      <%= render partial: Spree::Config.search_filter_presenter_class.partial_path %>
```
cbrunsdon: if we instantiate the presenter class before the partial and pass it in as a parameter to the view we'd be able to make the partial_path an instance method and give stores tighter control of how things can be extended

core/lib/spree/core/search/product_filter.rb
cbrunsdon: I like either of those options too.

core/app/models/spree/taxon.rb
cbrunsdon: Honestly I'm fine with just a deprecation warning and kill it in the future.

core/lib/spree/core/search/product_filter.rb
mtylty: @cbrunsdon you're right, it could be an awesome feature.

Recently I needed something like this for a store that had two "filter areas", one at the top and one on the side (each with its own set of models to search on) and I think using an instance method would have saved me a lot of time.

frontend/app/views/spree/products/index.html.erb
```diff
@@ -1,7 +1,9 @@
 <% content_for :sidebar do %>
   <div data-hook="homepage_sidebar_navigation">
     <% if "spree/products" == params[:controller] && @taxon %>
-      <%= render :partial => 'spree/shared/filters' %>
+      <% Spree::Config.search_filter_presenter_class.new(taxon: @taxon).tap do |presenter| %>
```
mtylty: I'm starting to wander if it's worth moving this inside the controller and saved in an instance variable.

kennyadsl: Maybe it's easier to extend in the view?

mtylty: yes it is... It's just that I don't like using too many `.tap` and `.new` inside views 😛

kennyadsl: Yes, I agree and I don't know what's better. Does anyone else have an opinion on this? 
Please, feel free to point us to any issue on this topic or a place in the Solidus codebase where we adopted a similar approach?

cc/ @cbrunsdon @jhawthorn

tvdeyen:
Closing this PR as this is very outdated. Conflicts need to be fixed and if the changes still apply this needs to be opened again. Thanks for contributing.

PR #1154, opened 2016-05-16
Improve helpers used by StockTransfers section

jhawthorn:
Previously there were two helpers, `display_name` and `state`. By default, all helpers are available to all spree controllers (both frontend and backend). These names are obviously too generic so I have renamed the former `admin_stock_location_display_name` and removed the latter (which was much clearer inline in the one place it was used).

This also uses `admin_stock_location_display_name` in all of the stock transfer views to ensure that "source -> dest" is displayed consistently at the top of each page.


mamhoff:
👍

gmacdougall:
:+1:

PR #1158, opened 2016-05-17
Stop expecting the taxation system to adjust shipment amounts

mamhoff:
Once we simplify the taxation system to trust that prices are correct
from the get-go, we can not expect it to change shipment amounts.

This should also not be necessary as the shipping method system already
allows for stratified prices by zone.

Essentially, these expectations add an unnecessary implementation burden.

There's a couple of specs still trying to refund stuff - those will be gone once the default zone is gone.


gmacdougall:
:+1:

jhawthorn:
:+1:

PR #1160, opened 2016-05-18
fix loader vertical position (centering)

RitaD:
I think that loader position was a little imperfect, almost seems like a bug without a vertical position/margin. So I thought that vertical centering was a good solution..but if there are better ideas I’m willing to implement them

#### Before:

schermata 2016-05-18 alle 14 29 55

#### After:

schermata 2016-05-18 alle 14 29 33

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_progress.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_progress.scss
@@ -3,6 +3,7 @@
   position: fixed;
   top: 0;
   right: 0;
+  bottom: 0;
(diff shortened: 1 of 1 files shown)


mamhoff:
👍 "a little imperfect" :shipit:

Mandily:
👍

jhawthorn:
I think in the future we want to improve our loading indicators even more (progress shown on submit buttons, more granular indication of what sections of the page are waiting). For now, this is a huge improvement over the broken looking spinner we had at the top of the page.

Thank very much :+1:

Issue #763, opened 2016-01-29
Upgrade to select2 4.0


jhawthorn:
select2 4.0 is out (so is a new version of select2-rails)

jhawthorn:
Started looking into this, and although there are some significant improvements in select2 4.0, there are a number of incompatibilities (and maybe some style issues) that make the upgrade non-trivial.

I'll revisit this again later, maybe after the select2 4.0.2 release

pusewicz:
@jhawthorn 4.0.2 is already available!

jhawthorn:
I'm no longer sure it's a good idea for our project to upgrade to select2 4.0.3. I can't entirely rule it out, but it would be a significant investment (both from me and stores/extensions upgrading) for questionable, and maybe negative benefit.

Select2 4.0 has significant API changes. See for example what it takes to [change variant_autocomplete.js.coffee to use select2 4.0](https://github.com/jhawthorn/solidus/commit/ffc963917526e76349650411bd3127d5e3f3f308#diff-79737d43b7a39433915b5674813e0b0cR12).
* The signature of `processResults` has entirely changed
* The signature of `data` has entirely changed
* `initSelection` can be made to work with modifications, but actually generates warnings that it is [deprecated without replacement](http://i.hawth.ca/s/VFX5PBIV.png).
* `formatSelection` and `formatResult` have been replaced with `templateResult` and `templateSelection`
  * They take different arguments
  * Output is escaped by default (it can be configured not to be escaped, but `Select2.util.escapeMarkup`) is no longer provided.
  * formatResult is initially called with the data `{loading: true, text: "Searching...."}`. **What?!**
* Automatic expanding width to match the existing element no longer seems to work


Those are all frustrating and will be an issue for every single one of our existing select2 components. But there's a greater issue that I think makes this a non starter: 

* select2 can no longer be tied to an `<input `

There is supposed to be a [compatibility layer](956ac46dab3acb94af487bfa201595ec16c8bab3) for this but it [does not work](https://github.com/select2/select2/issues/3022#issuecomment-216344362) at all. The markup not expects to get a `<select>` element even when doing ajax (which I don't believe is even supported in 3.5). This is a fair bit more awkward because we have to [provide at least the `<option>` for initial selection](https://github.com/jhawthorn/solidus/commit/ffc963917526e76349650411bd3127d5e3f3f308#diff-a49b505b614b2659d9eae6c90d79f698R78).

[...]

PR #1163, opened 2016-05-19
Select results highlighted

RitaD:
I fixed some little visual defects on search results like:

* Color of result text highlighted
* Outside spacing
* Margin of inner elements

#### Before

schermata 2016-05-18 alle 16 30 13

schermata 2016-05-18 alle 16 25 27

#### After 

schermata 2016-05-19 alle 11 34 42

schermata 2016-05-19 alle 11 46 19


jhawthorn:
:+1: Thank you.

I looked into why the colours were off before and found:
 
Yuck! Ideally we wouldn't blindly be applying colours to all `dl` and header tags, but we currently do so this is a fine and necessary workaround.

Also want to note that we will need to make sure we don't lose this fix when implementing #763

mamhoff:
👍

mamhoff:
Thank you!

PR #1168, opened 2016-05-20
Extract methods inStock::Coordinator#stock_location_variant_ids

isaacfreeman:
Breaks up Spree::Stock::Coordinator#stock_location_variant_ids into smaller methods that are easier to override.

My goal here is to be able to use a decorator to modify the query in `location_variant_ids` (I want to retrieve only stock items that haven't been reserved by some other user) without having to bring all of `stock_location_variant_ids` into my decorator.

### DIFF
--- a/core/app/models/spree/stock/coordinator.rb
+++ b/core/app/models/spree/stock/coordinator.rb
@@ -79,32 +79,33 @@ module Spree
       # queries that are kept as performant as possible, and only loads the
       # minimum required ActiveRecord objects.
       def stock_location_variant_ids
-        # associate the variant ids we're interested in with stock location ids
-        location_variant_ids = StockItem.
-          where(variant_id: unallocated_variant_ids).
-          joins(:stock_location).
-          merge(StockLocation.active).
-          pluck(:stock_location_id, :variant_id)
-
-        # load activerecord objects for the stock location ids and turn them
-        # into a lookup hash like:
-        #   {
-        #     <stock location id> => <stock location>,
-        #     ...,
-        #   }
-        location_lookup = StockLocation.
-          where(id: location_variant_ids.map(&:first).uniq).
-          map { |l| [l.id, l] }.
-          to_h
-
-        # build the final lookup hash of
-        #   {<stock location> => <set of variant ids>, ...}
-        # using the previous results
         location_variant_ids.each_with_object({}) do |(location_id, variant_id), hash|
           location = location_lookup[location_id]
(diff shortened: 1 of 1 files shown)


mamhoff:
👍 for the method breakup. 

However, I would prefer this to become a customizable class, so that instead of writing a pseudo-decorator, you can do this: 

```
Spree::Config.stock.coordinator_class = MyShop::Stock::Coordinator

module MyShop
  module Stock 
    class Coordinator < Spree::Stock::Coordinator
      def stock_location_variant_ids
        ...
      end
    end
  end
end
```

Would you mind adding that extension point to `Spree::Stock::Configuration`?

jordan-brough:
+1 on @mamhoff's suggestion.  Adding `Spree::Config.stock.coordinator_class` ([like we have for Spree::Config.stock.estimator_class](https://github.com/solidusio/solidus/blob/v1.2.2/core/lib/spree/core/stock_configuration.rb)) seems like the direction we'd want to head in.

In addition to that, it would be really good for us to document which (if any) private methods here are "public api" that we're committing to keep as stable overridable methods.  E.g., we just recently [removed the previous private methods](https://github.com/isaacfreeman/solidus/commit/290c03b172a14a2fef1ea3b4ebb61f98fa8a6294#diff-6289a1a8d0052dc24e66d7b72ca93869L67) in this area.

On the method split-up:  I actually prefer the method as it originally was, though I wrote the latest version so I'm biased. ;)  I feel like the extracted method is very tied to the original method and separating them just makes it bit harder to read.  I won't fight too much about that if others feel strongly, but it sounds like the method break-up itself wasn't really the point anyway, especially if we haven't committed to keeping those methods around in the first place.

mamhoff:
I didn't see that these were private methods. 

@jordan-brough is right here, we as they're private, they are not part of the public interface and thus should not be an extension point at all. Sorry @isaacfreeman; I think the right way to go about this is subclassing and replacing the entire method as @jordan-brough suggests.

isaacfreeman:
That's cool. I also agree with @mamhoff's approach. My code started as minimal change for a single client, but his is better architecture for Solidus in general.

I'm not so sure about keeping a single large private method, inasmuch as when I subclass the coordinator I'll still need to duplicate a lot of code for the sake of inserting a single line in a query. But I appreciate the argument that private methods shouldn't be considered extension points, and I can see the value in explicitly declaring a `coordinator_class` instead of just providing hooks for monkey patching.

I'll close this and submit another PR along the suggested lines.

PR #1169, opened 2016-05-20
Validate currency on Spree::Price

mamhoff:
Currently, the currency of a Spree::Price instance can be set to anything, and the price object won't validate it. 

This commit changes this so that only the currencies the Ruby Money library knows are accepted.


tvdeyen:
👍

jhawthorn:
:+1:

PR #1173, opened 2016-05-20
One-click Heroku demo button

jhawthorn:
Building on top of @alexblackie's great work in #1136

This uses a modified heroku ruby buildpack to generate a sandbox application for deploy. https://github.com/solidusio/heroku-buildpack-solidus-demo/commit/1f7500a29f07189d1f413d8e6b948f1af4db169d. This is a bit of a hack but it doesn't require us to have additional infrastructure and we should be able to create demos of arbitrary github repos and branches!

Also improved from #1136 is that we now ask for an admin username and password when creating the app on heroku.


alexblackie:
LGTM! :+1: :heart:

tvdeyen:
This is just freakin' awesome!

PR #1174, opened 2016-05-20
Replace missing image images

jhawthorn:
I hadn't noticed previously that we had missed replacing the old spree "missing image" images.

Ideally stores provide images for all products, and likely override this image, but we should display something reasonable and *unbranded* if they have not.


adammathys:
:+1:

jasonfb:
are you sure you have your S3 and/or paperclip settings configured correctly?  This works fine for me on a Solidus 1.3 fresh install, you can see the images here on this sample app: http://solidus-1-3-helloworld.herokuapp.com

jasonfb:
👍

tvdeyen:
:+1:

mamhoff:
Thanks!

PR #1175, opened 2016-05-20
Improve mailer previews DB sampling

jhawthorn:
Previously these mailer previews just found the first object in the database and used that to render the preview.

Now they make an effort to find a suitable record in the DB (completed/cancelled order, etc.) and raise an understandable error if there is none to be found.

This also now finds the _last_ matching item in the database, which should allow a developer to, for example, cancel an order and then work on what the cancellation email would look like.

cc @Sinetheta @alexblackie

### DIFF
--- a/core/lib/spree/mailer_previews/reimbursement_preview.rb
+++ b/core/lib/spree/mailer_previews/reimbursement_preview.rb
@@ -2,7 +2,9 @@ module Spree
   class MailerPreviews
     class ReimbursementPreview < ActionMailer::Preview
       def reimbursement
-        ReimbursementMailer.reimbursement_email(Reimbursement.first)
+        reimbursement = Reimbursement.last
+        raise "Your database needs at least one Reimbursement to render this preview" unless reimbursement
+        ReimbursementMailer.reimbursement_email(reimbursement)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


core/lib/spree/mailer_previews/reimbursement_preview.rb
```diff
@@ -2,6 +2,8 @@ module Spree
   class MailerPreviews
     class ReimbursementPreview < ActionMailer::Preview
       def reimbursement
+        reimbursement = Reimbursement.last
+        raise "Your database needs at least one Reimbursement to render this preview" unless reimbursement
         ReimbursementMailer.reimbursement_email(Reimbursement.first)
```
kennyadsl: do you need reimbursement variable here?

mamhoff:
👍 After fixing the using the `reimbursement` variable in the `ReimbursementPreview`.

jordan-brough:
👍 (w/ the variable update)

core/lib/spree/mailer_previews/reimbursement_preview.rb
jhawthorn: Thank you. Nice catch.

PR #1178, opened 2016-05-23
adds empty div with data-hook for extension

jasonfb:
for https://github.com/solidusio-contrib/solidus_shipping_labeler

### DIFF
--- a/backend/app/views/spree/admin/return_authorizations/edit.html.erb
+++ b/backend/app/views/spree/admin/return_authorizations/edit.html.erb
@@ -17,15 +17,16 @@
 
 
 <%= render :partial => 'spree/shared/error_messages', :locals => { :target => @return_authorization } %>
-<%= form_for [:admin, @order, @return_authorization] do |f| %>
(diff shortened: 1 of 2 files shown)


jhawthorn:
Instead of an empty `<div>` at the end, it might be better to wrap the entire form in a `div`. This still allows for the same customisation using deface (and more) and makes more semantic sense.

I'm not sure the test for the selector is providing any value.

jasonfb:
@jhawthorn As you suggested, I wrapped the whole form in a div with a class ```.return-authorization-form-wrapper```

see associated commit in solidus_shipping_labeler https://github.com/solidusio-contrib/solidus_shipping_labeler/commit/b587cf3a5d7add0b06bbaa80d3d8d9ecca3e42c1

gmacdougall:
This looks reasonable to me.

👍

jhawthorn:
:+1: Looks good. Sorry for the delay

jhawthorn:
Merged in #1373. Thank you

PR #1179, opened 2016-05-23
Disable Rubocop trailing comma rules

jordan-brough:
Personally I like trailing commas, but more than that this just seems
like something that we don't need to make people worry about.


cbrunsdon:
I'm both pro-trailing-comma and pro-less-rules :+1:

mamhoff:
Thanks!

PR #1180, opened 2016-05-23
Configurable Spree::Config.stock.coordinator_class

isaacfreeman:
Provides the option of cleanly replacing Spree::Stock::Coordinator with a custom coordinator class.

```ruby
Spree::Config.stock.coordinator_class = MyShop::Stock::Coordinator

module MyShop
  module Stock 
    class Coordinator < Spree::Stock::Coordinator
      def stock_location_variant_ids
        ...
      end
    end
  end
end
```

Replaces #1168.

### DIFF
--- a/core/app/models/spree/exchange.rb
+++ b/core/app/models/spree/exchange.rb
@@ -20,7 +20,7 @@ module Spree
 
     def perform!
       begin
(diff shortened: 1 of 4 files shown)


mamhoff:
👍 Great!

jhawthorn:
:+1: Thank you.

In a future change we should better document the Coordinator, the interface it is expected to provide and what it should do.

isaacfreeman:
@jhawthorn: Sure. Can you point me to any existing documentation I can use as a model?

PR #1181, opened 2016-05-24
Set up all spree initializers before config/init

cbrunsdon:
Okay, bear with me here for a sec, I'm not 100% sure this is the right change.

While upgrading an app to 1.3 (from 1.2) I ran into a problem that spree these initializers were not run before my config/initializers/ were run. For instance, ```Rails.application.config.spree.promotions``` was nil, rather than being a ```Spree::Promo::Environment.new``` like it should have been.

I tracked the change that caused the issue to 4f6f209d8681e035ae647032cb2b8fb842602d33, which looks completely innocuous.

Poking through the rails docs on "initializers" (http://guides.rubyonrails.org/configuring.html#initializers), I'm fairly certain explicitly setting when we want these initializers to run is a good thing as we definitely want them ran before our app's config/initializers are ran.

I'm not too familiar with this area of rails though, so if anyone believes differently please speak up.

### DIFF
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -17,14 +17,14 @@ module Spree
         Spree::Config = app.config.spree.preferences # legacy access
       end
 
-      initializer "spree.default_permissions" do |_app|
+      initializer "spree.default_permissions", before: :load_config_initializers do |_app|
         Spree::RoleConfiguration.configure do |config|
           config.assign_permissions :default, [Spree::PermissionSets::DefaultCustomer]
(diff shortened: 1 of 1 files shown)


tvdeyen:
Sounds reasonable to me. Hopefully no one ever expects our initializers to be run after theirs. But let's try it. :thumbsup:

mamhoff:
Makes sense to me. 👍

jhawthorn:
I'd love to do a follow up on this, needing to add this to all initializers is a bit weird, but it does seem to solve the problem, so :+1:

PR #1182, opened 2016-05-24
Remove is_default Boolean from prices

mamhoff:
This boolean is either a very bare-bones version of `acts_as_paranoid` or
a very bare-bones version of `papertrail`. However: It's really badly named,
as `is_default` is not something I understand very well at all.

We can accomplish the same behavior by simply using an `order` clause in
the `currently_valid` scope on prices.

### DIFF
--- a/backend/app/views/spree/admin/prices/_table.html.erb
+++ b/backend/app/views/spree/admin/prices/_table.html.erb
@@ -4,7 +4,6 @@
   <thead data-hook="prices_header">
     <tr>
       <th><%= Spree::Variant.model_name.human %> </th>
-      <th><%= Spree::Price.human_attribute_name(:is_default) %></th>
       <th><%= Spree::Price.human_attribute_name(:amount) %></th>
       <th><%= Spree::Price.human_attribute_name(:currency) %></th>
       <th class="actions"></th>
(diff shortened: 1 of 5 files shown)


tvdeyen:
Although we just talked about this IRL, I don't understand the `order` clause thing. But anyway. This Boolean is not useful at all and we can't find any reference to prices in the current code base, that justify such thing. 

👍

mamhoff:
The order thing is there so that this test passes: https://github.com/magiclabs/solidus/blob/remove-is-default-boolean-from-prices/core/spec/models/spree/variant_spec.rb#L174-L180

Essentially, if you add a price to the prices array that has the same properties as the an existing price, the new price should be selected.

adammathys:
:+1:

jhawthorn:
This could use a CHANGELOG entry. Otherwise, :+1:

jhawthorn:
This could use a CHANGELOG entry. Otherwise, :+1:

peterberkenbosch:
@mamhoff any change you could rebase this PR and update the CHANGELOG? Thanks!

mamhoff:
Closing this in favor of #1469.

PR #1185, opened 2016-05-24
Allow setting nil in preferences

cbrunsdon:
Previously, a nil value on an integer preference would .to_i to 0.

Leaving it as nil rather than casting it to an int makes more sense
and would be more in line with what I would expect to happen.

This came out of an issue I had in an upgrade from 1.2 to 1.3.

I wanted to nil out default_country_id and switch to default_country_iso, but as I was unable to set default_country_id to nil I was unable to switch:
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/country.rb#L16


jhawthorn:
:+1:

Issue #1164, opened 2016-05-19
Pluralization issue with certain languages


dangerdogz:
Reposted from [Solidus I18](https://github.com/solidusio-contrib/solidus_i18n/issues/58)

References

[I18n Issue](https://github.com/svenfuchs/i18n/issues/239)
[ActiveAdmin Issue](https://github.com/activeadmin/activeadmin/pull/2791)
[Spree Issue](https://github.com/spree/spree/issues/5842)
[Spree Issue](https://github.com/spree/spree/pull/6059)

**Summary**

Basically, english use this function to generate plural names
[Rails i18n link](https://github.com/svenfuchs/rails-i18n/blob/master/lib/rails_i18n/common_pluralizations/one_other.rb)
/one_other.rb
```ruby
module RailsI18n
  module Pluralization
    module OneOther
      def self.rule
        lambda { |n| n == 1 ? :one : :other }
      end

      def self.with_locale(locale)
        { locale => {
            :'i18n' => {
              :plural => {
                :keys => [:one, :other],
                :rule => rule }}}}
      end
    end
  end
end
```

And some other languages like french use this one
[Rails i18n link](https://github.com/svenfuchs/rails-i18n/blob/master/lib/rails_i18n/common_pluralizations/one_upto_two_other.rb)
```ruby
module RailsI18n
  module Pluralization
    module OneUptoTwoOther
      def self.rule
        lambda { |n| n && n >= 0 && n < 2 ? :one : :other }
      end

      def self.with_locale(locale)
        { locale => {
            :'i18n' => {
              :plural => {
                :keys => [:one, :other],
                :rule => rule }}}}
      end
    end
  end
```

**Issue**

Throughout Solidus, the pluralization is triggered with this format :  ```<%= Spree::PromotionCategory.model_name.human(count: :many) %>```

It works with the english OneOther even if n is a symbol (:many) because :symbol == 1 will just return false (so you get the :other), but it crashes when you try to use another pluralization method because it's trying to determine if n ( a symbol, :many) is less or greater than an integer, thus you get :   ``` (comparison of Symbol with 0 failed)``` because you can't use this operator with a symbol

dangerdogz:
Reposting comment by @tvdeyen :

The only way to fix this is to pass in an actual number, right? Good catch and great investigation @dangerdogz

[...]

PR #1187, opened 2016-05-24
Improve I18n pluralization to avoid errors with some calculators

dangerdogz:
This is a fix for issue #1164 

1) First commit is to add a preference containing the "magic number"
2) Second commit is to add the helper to wrap the usage of ActiveRecord::Naming
3) Third commit is find/replace to use the helpers

It is likely that we'll need to add more keys to fully support every pluralization calculators (slavic languages use more keys like : [ few, many ] and we only use one/other) but this address the underlying issue of trying to compare if a Symbol is less or greater than an integer which was causing crashes with the OneUptoTwoOther calculator used by the french language.

I opted for the preference/helper approach so the number can be updated without patching or monkey patching core, and the usage can be updated without editing every file from the backend but I can revert to a simpler usage if needed

EDIT : I also confirm that it fixes the issues raised by https://github.com/luukveenis/solidus_i18n/pull/6

Suggestions welcome

### DIFF
--- a/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
@@ -1,7 +1,7 @@
 <%= render 'spree/admin/shared/settings_checkout_tabs' %>
 
 <% content_for :page_title do %>
-  <%= Spree::AdjustmentReason.model_name.human(count: :other) %>
+  <%= plural_resource_name(Spree::AdjustmentReason) %>
 <% end %>
 
 <% content_for :page_actions do %>
--- a/backend/app/views/spree/admin/customer_returns/edit.html.erb
+++ b/backend/app/views/spree/admin/customer_returns/edit.html.erb
@@ -54,7 +54,7 @@
 <% end %>
 
 <fieldset data-hook="reimbursements" class="no-border-bottom align-center">
-  <legend align="center"><%= Spree::Reimbursement.model_name.human(count: :other) %></legend>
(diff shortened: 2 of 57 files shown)


core/app/models/spree/app_configuration.rb
```diff
@@ -274,6 +274,14 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Boolean] Indicates if stock management can be restricted by location
     preference :can_restrict_stock_management, :boolean, default: false
 
+    # @!attribute [rw] i18n_generic_plural
+    #   This value is used as a count for the pluralization helpers related to I18n
+    #   ex: Spree::Order.model_name.human(count: Spree::I18N_GENERIC_PLURAL
+    #   Related to Solidus issue #1164, this is needed to avoid problems with
+    #   different pluralization calculators
+    #   @return [Integer] default: 2.1
+    preference :i18n_generic_plural, :integer, default: 2.1
```
jhawthorn: Is there a reason for this to be a preference? I think having it at `Spree::I18N_GENERIC_PLURAL` would probably be sufficient for everyone.

dangerdogz: Where should it live?

jhawthorn: Could be in here I suppose https://github.com/solidusio/solidus/blob/master/core/lib/spree/i18n.rb

dangerdogz: Noted, updating it

PR #1188, opened 2016-05-24
Simplify DB cleaning in backend specs

jhawthorn:
Previously we used wait_for_ajax after every request, which was error prone and gross.

As of Capybara 1.7, Capybara will wait for all requests to the server to finish. Because of this we can just make sure we run the DB cleaner after Capybara has been reset.


gmacdougall:
:+1:

adammathys:
:+1:

PR #1190, opened 2016-05-24
Fix payment#payment_method for deleted records

### LINKED ISSUE #1186: Soft-deleted payment methods cause errors in admin
After deleting a payment method in the admin, any references to it on past orders will lead to 500 errors. (e.g.: The show page and all events that can be triggered on `/admin/orders/:number/payments`)

jhawthorn:
Fixes #1186

### DIFF
--- a/core/app/models/spree/payment.rb
+++ b/core/app/models/spree/payment.rb
@@ -11,7 +11,7 @@ module Spree
 
     belongs_to :order, class_name: 'Spree::Order', touch: true, inverse_of: :payments
     belongs_to :source, polymorphic: true
-    belongs_to :payment_method, class_name: 'Spree::PaymentMethod', inverse_of: :payments
+    belongs_to :payment_method, -> { with_deleted }, class_name: 'Spree::PaymentMethod', inverse_of: :payments
 
     has_many :offsets, -> { offset_payment }, class_name: "Spree::Payment", foreign_key: :source_id
     has_many :log_entries, as: :source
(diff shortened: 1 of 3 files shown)


adammathys:
:+1:

mamhoff:
👍

core/app/models/spree/payment.rb
```diff
@@ -11,7 +11,7 @@ class Payment < Spree::Base
 
     belongs_to :order, class_name: 'Spree::Order', touch: true, inverse_of: :payments
     belongs_to :source, polymorphic: true
-    belongs_to :payment_method, class_name: 'Spree::PaymentMethod', inverse_of: :payments
+    belongs_to :payment_method, -> { with_deleted }, class_name: 'Spree::PaymentMethod', inverse_of: :payments
```
jasonfb: 👎

jasonfb:
this is an anti-pattern and should not be allowed to proliferate throughout the codebase, although I do realize it fixes a bug today.

mamhoff:
@jasonfb I agree with you. `acts_as_paranoid` is a bad replacement for properly time-windowed models. But as long as we don't have those, we need to work with what we have. I'm sorry.

jasonfb:
Yes fair enough. This does indeed fix a bug so... "bob's your uncle." (That's something Europeans say right?)

jhawthorn:
Wanted to note that I had to revert this because the new test caused a spurious failure :disappointed:

PR #1191, opened 2016-05-25
Improve I18n pluralization to avoid errors with some calculators

dangerdogz:
(Duplicate, closed the other one to avoid starting builds while I was working on it and I can't reopen it because I forced pushed the original branch, sorry about that)

This is a fix for issue #1164

1) First commit is to add a constant containing the "magic number" we want to use
2) Second commit is to add the helper to wrap the usage of ActiveRecord::Naming
3) Third commit is find/replace to use the helper

It is likely that we'll need to add more keys to fully support every pluralization calculators (slavic languages use more keys like : [ few, many ] and we only use one/other) but this address the underlying issue of trying to compare if a Symbol is less or greater than an integer which was causing crashes with the OneUptoTwoOther calculator used by the french language.

Replaced the original preference by a simple constant (thanks jhawthorn), but I kept the helper usage

Suggestions welcome

### DIFF
--- a/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
@@ -1,7 +1,7 @@
 <%= render 'spree/admin/shared/settings_checkout_tabs' %>
 
 <% content_for :page_title do %>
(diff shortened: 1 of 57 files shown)


tvdeyen:
👍

mamhoff:
👍

PR #1194, opened 2016-05-25
Rewrite payment method selection for admin new payment

jhawthorn:
Previously a credit card payment method would have to be named "Credit Card" in order to allow creating new payments through the admin. This is because it was using the payment method's name, running it through I18n, and using that as a selector to find a fieldset to show.

Now it just finds a div with the correct id.

Thanks @Murph33 who pointed this out a while ago


forkata:
:+1: Have definitely run into this before. Thanks for fixing this!

mamhoff:
👍 Beautiful!

gmacdougall:
:+1: Great work John!

PR #1200, opened 2016-05-27
Fix payment#payment_method for deleted records

jhawthorn:
Take two of #1190, which was reverted due to a broken test

### DIFF
--- a/backend/spec/features/admin/orders/payments_spec.rb
+++ b/backend/spec/features/admin/orders/payments_spec.rb
@@ -222,5 +222,29 @@ describe 'Payments', type: :feature do
         expect(page).to have_content("Payment Updated")
       end
     end
+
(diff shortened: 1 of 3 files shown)


bbuchalter:
I don't see any difference between this and #1190 - what changed?

jhawthorn:
Change was `payment.number` vs `payment.amount` :disappointed_relieved:

jordan-brough:
👍 

@jhawthorn just curious - you're saying the old spec was flaky?  If so, what made `number` flaky and `amount` reliable?

jhawthorn:
@jordan-brough `number` isn't displayed on the final page at all. So it was less "flaky" and more "entirely broken" I didn't notice because CircleCI decided not to test pull requests authored by me, so it actually wasn't being tested.

Hound made it look like it passed

PR #1202, opened 2016-05-30
WIP Update select2 to v 4.0.2

erikaxel:
The pull is a WIP.

I made the pull request to get feedback on the approach before refactoring all select2-boxes in the admin.

The following works so far:
* Select2 boxes without custom AJAX should work. I have not tested all.
* The select2 box to select variant filter on orders/index (admin front page) works

Questions:
**Styling**
1) I have opted for a minimal set of CSS-styling. I find the select2-default to be quite sensible and works IMHO out of the box. So instead of making design decisions in this pull I would prefer it to be done separately. 

**Setting selected values**
2) One issue that popped up during the upgrade was how to set initial default values on AJAX select2-boxes. An example would be the variant filter which can be applied on the admin orders page. There are two reasons why we can not use the old way: a) The new select2 library only accepts select boxes and not input boxes as before. b) In addition the old "initSelection" method has been removed. 
I have tried various ways, and landed on one where we set the option directly in .erb code, instead of using a AJAX-call. This eliminates the one call and uses the standard way of using HTML selects. Does this seem to be an ok approach?

**Testing**
3) I have changed the test for the orders page to work. I had to add some new select2 spec helpers make it work, since the old code for testing select2-boxes is heavily tied to select2 v1. The old spec helpers are in core, but I choose to only the new ones to the backend project since it is only used there for now. Should it  be in core together with the other select2 code?

Many of the other tests fail now, since I haven't gone though all select2 boxes, so this is expected.

The related issue is  #763


graygilmore:
> 1) I have opted for a minimal set of CSS-styling. I find the select2-default to be quite sensible and works IMHO out of the box. So instead of making design decisions in this pull I would prefer it to be done separately.

I believe that changing the styles from what it was before to this *is* a design decision. And since we're doing that we might as well make the styles match. Something as simple as just match to the inputs around it:

screen shot 2016-05-30 at 1 19 17 pm

erikaxel:
@graygilmore Very valid point. I have made some small changes to make it blend more in. 
Now it looks like this:
screenshot 2016-05-30 23 50 58

jhawthorn:
@erikaxel Thank you for starting on this.

The tests are going to need some work. Using `has_selector?` has some issues and shouldn't be used. It would also be nice if we could continue using the existing test methods, since those are likely to be used in extensions and test suites of solidus stores. I'd like to help with this, I'll try to send a PR to your branch soon.

jhawthorn:
@erikaxel Here's my first go at updating the existing methods, will probably need some more work to get all of the cases work, but this passes with the order listing specs. https://github.com/erikaxel/solidus/pull/1

It _should_ be possible to make this compatible with both select2 3.5 and select2 4, but I haven't attempted that yet.

jhawthorn:
Closing this due to inactivity. If you feel like picking this back up please open a new PR.

PR #1206, opened 2016-05-31
Update in memory line items in CreateItemAdjustments

jordan-brough:
Update in-memory line items in CreateItemAdjustments

Previously this:
```
Spree::PromotionHandler::Coupon.new(order).apply
```
would not update `adjustment_total` on any already-loaded
`order.line_items` because CreateItemAdjustments didn't operate on the
possibly-cached line items, nor did it reload/reset them.

### DIFF
--- a/core/app/models/spree/promotion/actions/create_item_adjustments.rb
+++ b/core/app/models/spree/promotion/actions/create_item_adjustments.rb
@@ -68,10 +68,12 @@ module Spree
         def line_items_to_adjust(promotion, order)
           excluded_ids = adjustments.
             where(adjustable_id: order.line_items.pluck(:id), adjustable_type: 'Spree::LineItem').
-            pluck(:adjustable_id)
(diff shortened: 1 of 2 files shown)


jhawthorn:
Removes stubs, reduces a query, fixes a bug

Awesome :+1:

cbrunsdon:
Yea, this is great. Actually fantastic. I can't imagine how many dev hours this is going to save people in the future.

:+1:

mtomov:
:+1:  

Don't know where the bad rep of stubs is coming from, as especially in Rspec - they are verified.

Change is great indeed! Thanks!

cbrunsdon:
@mtomov iirc receive_messages doesn't use a verifying double unless you configure it to, which will be default in future versions of rspec but not with what we're currently running in core.

mtomov:
:+1:

PR #1207, opened 2016-06-01
Cleaning up some inline js in admin settings

peterberkenbosch:
the default_currency and currency fields are not present anymore in the general settings.
this commit cleans up some remaining inline js code.

### DIFF
--- a/backend/app/views/spree/admin/general_settings/edit.html.erb
+++ b/backend/app/views/spree/admin/general_settings/edit.html.erb
@@ -85,11 +85,4 @@
     </fieldset>
 
   </div>
-
-
 <% end %>
-
-<script>
-  $('#store_default_currency').select2();
-  $('#currency').select2();
-</script>


flyfy1:
but why not allowing the admin to set the default_currency?

peterberkenbosch:
correct me if I'm wrong @jhawthorn but the idea is to have those setting in the spree.rb config setup, to keep them static and not hitting the database. This might change with the multi store ideas though.

gmacdougall:
There are two major ways to configure currency:

- Configure the global default currency through the application configuration

```
Spree.config do |config|
  config.currency = 'EUR'
end
```
- Configure the `default_currency` for a specific store on the `Spree::Store` model

Allowing admins to set the currency at runtime through the admin is something that broke a large number of assumptions about how the site operates (such as caching) in addition to the performance considerations.

flyfy1:
@gmacdougall but if `store.default_currency` is nil.. then "add to cart" would throw error.

(I think u can easily verify that with master branch of Solidus)

cbrunsdon:
@flyfy1 aye, running a store sans default currency doesn't make any sense, but we expect developers to set up the store in config rather than logging in as an admin.

The history of the "settings" page was aimed at supporting "hosted" spree, where admins could set up their stores using the settings pages, but it led to a bunch of very difficult situation:
* admins changing settings with unforseen consequences
* dev/test/prod/staging environments with different configs
* serious caching issues when switching live settings wouldn't invalidate caches

We have been phasing out of those pages in exchange for config blocks for many versions now, this commit just cleans up some JS that was accidentally left in.

:+1:

flyfy1:
Ahh Yeah thanks @cbrunsdon. I agree the pull request itself is nice, just trying to understand the code base also :)

jhawthorn:
:+1:

PR #1211, opened 2016-06-02
Restore price updating in LineItem#options=

jordan-brough:
LineItem#options= previously updated the price when the options
changed.  This was removed with the Pricer work but might cause
backward-compatiblity problems for stores that are using it.

This was a breaking change for our code while updating to Solidus v1.3.  We are using a gift_wrap attribute.

There doesn't seem to be any real point to having `LineItem#options=` other than updating the price. Perhaps we should also (or instead) consider deprecating or removing or renaming `LineItem#options=`?  As @mamhoff pointed out in chat, it's strange that if you update an attribute directly it won't affect the price, but if you update it through `options=` then the price does update.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -102,14 +102,20 @@ module Spree
       !sufficient_stock?
     end
 
-    # Sets options on the line item.
+    # Sets options on the line item and updates the price.
     #
     # The options can be arbitrary attributes on the LineItem.
     #
     # @param options [Hash] options for this line item
     def options=(options = {})
       return unless options.present?
+
       assign_attributes options
+
+      # There's no need to call a pricer if we'll set the price directly.
+      unless options.key?(:price) || options.key?('price')
+        self.money_price = variant.price_for(pricing_options)
+      end
     end
 
     def pricing_options
--- a/core/spec/models/spree/line_item_spec.rb
+++ b/core/spec/models/spree/line_item_spec.rb
@@ -214,6 +214,12 @@ describe Spree::LineItem, type: :model do
       line_item.options = { price: 123 }
       expect(line_item.price).to eq 123
     end
+
+    it "updates the price based on the options provided" do
+      expect(line_item).to receive(:gift_wrap=).with(true)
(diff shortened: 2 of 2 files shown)


mamhoff:
This actually looks alright to me. I guess the removal was triggered by associating this feature too closely with the `price_modifier_*`. 👍

jhawthorn:
:+1:

adaddeo:
`options=` has been useful and frustrating for us.

Currently there isn't one standard way to permit and assign additional attributes on line items.

[OrderContents](https://github.com/solidusio/solidus/blob/v1.3/core/app/models/spree/order_contents.rb#L89) provides a way to assign additional attributes, but [Core::Importer::Order] (https://github.com/solidusio/solidus/blob/v1.3/core/lib/spree/core/importer/order.rb#L103-L108) doesn't use it and instead calls `line_item.update_attributes` directly.

[Api::LineItemsController](https://github.com/solidusio/solidus/blob/v1.3/api/app/controllers/spree/api/line_items_controller.rb#L6) provides a way to permit additional attributes, but these don't help when using [Importer::Order](https://github.com/solidusio/solidus/blob/v1.3/api/app/controllers/spree/api/orders_controller.rb#L95-L102) to create an order with line items, and [OrderContents](https://github.com/solidusio/solidus/blob/v1.3/core/app/models/spree/order_contents.rb#L89) has it's own way of permitting additional attributes.


The way forward seems to be:
- Contain all logic for assigning additional attributes to line items in OrderContents
    - Remove `line_item.update_attributes` from [Importer::Order](https://github.com/solidusio/solidus/blob/v1.3/core/lib/spree/core/importer/order.rb#L103-L108)
- Leave all parameter permitting logic to the controllers
    - Remove ActionController::Parameters#permit from [OrderContents#add]((https://github.com/solidusio/solidus/blob/v1.3/core/app/models/spree/order_contents.rb#L89). I don't think there is a single other instance of parameters being permitted like this outside a controller in the codebase.
- Use the existing PermittedAttributes module to provide a DRY and extensible way of permitting additional line item attributes

I actually tried to take a shot at all this in #1112, but it could've been more direct.

PR #1213, opened 2016-06-02
Add an extra note to the changelog about the versioncake update

jordan-brough:
This would've been helpful for me while upgrading to v1.3.


jhawthorn:
:+1: Thank you

cbrunsdon:
Aye, good by me too :+1:. I solved this a different, likely worse way...

we should think about yanking versioncake from core. We don't use it and I doubt we ever will.

PR #1215, opened 2016-06-02
Fix failure in Splitter::ShippingCategory spec

jhawthorn:
After the change to use the `ClassConstantizer::Set` in the core engine initializers (#1203), it is possible for this spec to be run before `Spree::Stock::Splitter::ShippingCategory` is loaded. In that case, `ShippingCategory` as it is used here would resolve to `::ShippingCategory` (rails needs a const_missing to fire in order to provide it's autoloading feature).

This fixes the spurious error:

```
ArgumentError:
  When assigning attributes, you must pass a hash as an argument.
# ./spec/models/spree/stock/splitter/shipping_category_spec.rb:28:in `block (2 levels) in <module:Splitter>'
# ./spec/models/spree/stock/splitter/shipping_category_spec.rb:39:in `block (2 levels) in <module:Splitter>'
```

### DIFF
--- a/core/spec/models/spree/stock/splitter/shipping_category_spec.rb
+++ b/core/spec/models/spree/stock/splitter/shipping_category_spec.rb
@@ -2,46 +2,44 @@ require 'spec_helper'
 
 module Spree
   module Stock
-    module Splitter
-      describe ShippingCategory, type: :model do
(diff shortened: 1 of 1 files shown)


cbrunsdon:
Looks reasonable to me, thanks :+1:

PR #1217, opened 2016-06-03
Factory fixes for countries without subregions

jordan-brough:
I ran across these while upgrading to Solidus 1.3.

Worth merging to the 1.3 branch?


mamhoff:
👍 I wrote the initial state-aware factories, and I wasn't aware of states with no subregions. I'd be good with this in 1.3 as well.

jhawthorn:
:+1:

PR #1220, opened 2016-06-04
Re-write slow Order/Shipment migration with SQL

stewart:
This migration was originally written with SQL, but was updated to use
ActiveRecord logic due to an issue related to shipments with no cost:

https://github.com/spree/spree/commit/2b7638eeeb6c5c2f0dded0122ce9a82c1d0f6956

However, the performance of the ActiveRecord version of this query is
significantly impacted by large numbers of Orders and Shipments.

Fortunately for us, the previous SQL version only needs a `COALESCE()`
wrapper around the sub-select to address the same issue the AR version
was written to fix.


jhawthorn:
This is a very old migration (2.1 -> 2.2 I believe), but we might as well patch it now for anyone taking that upgrade path.

:+1: migrations should be written in SQL as much as possible (I argued for this when this migration was originally changed in spree). Both because it's faster and because it avoids using whatever behaviours are in the model (which change over time).

cbrunsdon:
Aye, I'm good with this too, making life easier for old stores to upgrade is nifty :+1:

PR #1224, opened 2016-06-07
Task and task spec fixes

jordan-brough:
See individual commits:

- Fix rake task specs - only run rake tasks once
- Fix bug in rake task "assure_store_on_orders"
- Change `abort` to `raise` in "assure_store_on_orders"

### DIFF
--- a/core/lib/tasks/migrations/assure_store_on_orders.rake
+++ b/core/lib/tasks/migrations/assure_store_on_orders.rake
@@ -16,9 +16,9 @@ namespace :solidus do
 
         spree_store_count = Spree::Store.count
         if spree_store_count == 0
-          abort "You do not have a store set up. Please create a store instance for your installation."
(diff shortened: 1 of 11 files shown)


jhawthorn:
This looks good to me :+1:

A few additional thoughts:
* We should probably rename `assure_store_on_orders` to `ensure_store_on_orders`
* Does ensuring there's a store_id belong in this rake task? It seems like it could be a migration.

jordan-brough:
I think a migration could make sense here.  Our store needs this rake task because in Solidus < 1.3 a store is only added if the Solidus-provided controllers are used to generate orders.  Our app doesn't use those controller so we have lots of orders without store ids.

jordan-brough:
Added a commit to rename assure -> ensure.

jordan-brough:
@mamhoff any thoughts on moving this to a migration?

jhawthorn:
Thanks Jordan :+1:

Issue #1150, opened 2016-05-15
Validation failed: Store can't be blank when adding to cart (on solidus hello-world install, v1.3)


jasonfb:
I get this when I click "Add to Cart" on a basic product. In my spree_stores table I have a single record for 'Sample Store' as it was populated by the generators (?)

```
ActiveRecord::RecordInvalid in Spree::OrdersController#populate
Validation failed: Store can't be blank
```

full stacktrace:
https://gist.github.com/anonymous/46b17070a302079102640f70fca219ae

Gemfile.lock
https://gist.github.com/jasonfb/024fc7c9435d30ea362979dad1f42884

jasonfb:
this was fixed by manually setting the 'default' flag on the 'Sample Store' record in spree_stores to true (1)

![2016-05-15_1946](https://cloud.githubusercontent.com/assets/59002/15277550/f7f02e34-1ad5-11e6-9530-0ea004ecf9bf.png)

changed to

![2016-05-15_1947](https://cloud.githubusercontent.com/assets/59002/15277552/fcdd4148-1ad5-11e6-8600-5260be6251cb.png)

jasonfb:
closing ; cross-reference https://stackoverflow.com/questions/37244912/on-solidus-1-3-fresh-install-validation-failed-store-cant-be-blank-when-addin/37244913#37244913

PR #1230, opened 2016-06-08
Migrations to fix up default store and store_id on orders

jhawthorn:
Spun off of #1224 and #1225

This PR accomplishes two things:

1)
A previous change to the `CreateStoreFromPreferences` migration caused the store to no longer be set as a default.

This updates that migration to set the store it is creating as the default. This also adds a later migration to set the first store to the default if there is not an existing default. Since the `CreateStoreFromPreferences` migration will be run very shortly after the creation of the Store table, this should be a safe assumption.

This should fix issues like #1150

2)
Adds a migration which is the equivalent of `rake solidus:migrations:ensure_store_on_orders:up`


mamhoff:
Thanks John! Does this not make the rake task obsolete and delete-worthy then? The migrations will be run before the rake tasks, so...

Otherwise, 👍

jhawthorn:
I will submit a second PR for removing the rake task

PR #1231, opened 2016-06-08
Use order factory in specs to ensure we have a valid model

jordan-brough:
These `creates` were actually failing because the order didn't have a
store, but we were using `create` instead of `create!` so we didn't
notice.

*Then* `order.contents.add` would get called, which persists using
`order.save!(validate: false)`, bypassing the store validation and
persisting the order in an invalid state.


mamhoff:
👍 Thanks Jordan!

jhawthorn:
:+1: In the future, if we do want to use `create` in a spec we should make sure to use `create!`

Issue #1189, opened 2016-05-24
Allow editing all Stores through the admin


jhawthorn:
Currently the "General" settings page is mostly (entirely?) for editing the default Store. It should be replaced with an interface for editing _any_ of the stores in the DB.

This would probably fix some awkwardness we see like #1176

jasonfb:
Are multiple store supported natively in core? I thought support for this was provided though the solidus_multi_domain gem. It would seem logical then that it is that extension's responsibility to provide this.

jhawthorn:
solidus_multi_domain includes some specifics about what multi-store behaviour means, but to some degree Stores have been natively supported since spree 2.3 (models in core, `current_store` method, association to orders) https://github.com/solidusio/solidus/blob/master/core/app/models/spree/store.rb

PR #1232, opened 2016-06-08
[RFC] Raise exception when there is no default store

jhawthorn:
We wanted to investigate what it would look like to raise an exception on `Store.current`/`Store.default` if there was no default store in the database.

As you can see there are a lot of spec changes needed to accomplish this, but not an obscene amount. We should think that it will be similar for users upgrading.

Some of the required changes were to avoid the "General Settings" page of the admin (which is actually a page to edit the default store). It would probably be helpful if we could make that page not error as part of #1189

### DIFF
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -903,7 +903,8 @@ RSpec.describe Spree::Order, type: :model do
   end
 
   context "#can_ship?" do
-    let(:order) { Spree::Order.create }
+    let(:store) { create(:store) }
+    let(:order) { Spree::Order.create(store: store) }
 
     it "should be true for order in the 'complete' state" do
       allow(order).to receive_messages(complete?: true)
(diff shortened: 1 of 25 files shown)


core/spec/models/spree/order_spec.rb
```diff
@@ -818,7 +813,8 @@ def merge!(other_order, user = nil)
 
   # Regression test for https://github.com/spree/spree/issues/4923
   context "locking" do
-    let(:order) { Spree::Order.create } # need a persisted in order to test locking
+    let(:store) { create(:store) } # need a persisted in order to test locking
+    let(:order) { Spree::Order.create(store: store) } # need a persisted in order to test locking
```
mamhoff: The comment is missing a word, maybe `store`?

mamhoff:
I'm a fan. 💯

mamhoff:
I'm still a fan of this. @jhawthorn whenever you have a minute would you mind rebasing this on the current master branch?

tvdeyen:
@jhawthorn any chance to look into this again?

PR #1235, opened 2016-06-09
Assert factories create valid objects

bbuchalter:
In previous versions of Spree factories could create invalid objects.
Let's simply assert that objects that get created are in fact valid.


jhawthorn:
Spec run failure looked unrelated retrying.

jhawthorn:
@bbuchalter could you rebase and reopen this targeting master instead of 1.2?

PR #1236, opened 2016-06-10
Assert factories create valid objects

bbuchalter:
In previous versions of Spree factories could create invalid objects.
Let's simply assert that objects that get created are in fact valid.


mamhoff:
👍  (please rebase on master though, this won't make it into 1.2)

jhawthorn:
:+1:

jhawthorn:
Thanks Brian

PR #1238, opened 2016-06-10
Remove comments and improve use of Store in emails

jhawthorn:
Remove comments from BaseMailer header and footer. These were unnecessary (they're comments), but were actually doing work (ERB was evaluated).

Changed OrderMailer to use the store tied to it's order was using the default store instead of the store tied to the order.

Changed TestMailer to use Store.default instead of Store.current with no argument.


mamhoff:
:+1: looks perfect to me.

PR #1239, opened 2016-06-10
Fix Store.current without args

jhawthorn:
When `Store.current` is called without arguments, it's supposed to return the default store.

Previously it was returning an arbitrary store since `Store.by_url(nil)` returns an SQL scope of `url LIKE '%%'`. It was most likely to return the first, which is why this wasn't noticed.

This also adds a deprecation warning for calling `Store.current` without an argument. Developers may believe it does the same thing as `current_store` (which it can't). I'd be interested in fully deprecating `Store.current` in the future in favour of `current_store` and `order.store`.

Waiting on:

* [x] #1238

### DIFF
--- a/core/app/models/spree/store.rb
+++ b/core/app/models/spree/store.rb
@@ -15,6 +15,7 @@ module Spree
     scope :by_url, lambda { |url| where("url like ?", "%#{url}%") }
 
     def self.current(store_key = nil)
+      Spree::Deprecation.warn "Spree::Store.current needs a code or URL as an argument. If you want the default store use Spree::Store.default", caller if !store_key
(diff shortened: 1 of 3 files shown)


mamhoff:
👍

jhawthorn:
I'm waiting on #1238 before merging this. Since that will fix the deprecation warnings this will introduce.

* [ ] #1238

tvdeyen:
Instead of the need of passing in arguments to get the current store, we should store it on the current thread, so we have one single point of truth for getting "the current store" (however this is been set). 

As a developer I would expect to get the "current store", not that I need to pass in an url or code, because that's what `find_by_url` or `find_by_code` are for. 

Why would I want to ask for "the current store", if I need to pass in the `url` or `code` anyway?

PR #1246, opened 2016-06-14
Upgrade rake task fixes

jordan-brough:
Fixes for #1245. See individual commits.


jordan-brough:
It'd be great to get more specs around these tasks and also to address the "TODO".  If anyone has the time or inclination to do so that would be great!  I think this at least gets us to a starting point though.

jhawthorn:
:+1:

jhawthorn:
I've sent you a PR to address the TODO https://github.com/jordan-brough/solidus/pull/3

mamhoff:
👍  Thank you!

PR #1249, opened 2016-06-14
Don't cache Spree::Config.default_pricing_options

jhawthorn:
This is built on top of #1246. The commit to be reviewed is https://github.com/solidusio/solidus/pull/1249/commits/62b247bc6a4e7f58f49e84cdd97a4e6576639846

I don't think there's a need to cache this as the only work it does is look up two Config settings.

Removing the caching allows removing a hack from the `PriceMigrator`.

### DIFF
--- a/core/app/models/spree/app_configuration.rb
+++ b/core/app/models/spree/app_configuration.rb
@@ -299,7 +299,7 @@ module Spree
     # @return [variant_price_selector_class] An instance of the pricing options class with default desired
     #   attributes
     def default_pricing_options
-      @default_pricing_options ||= pricing_options_class.new
+      pricing_options_class.new
     end
 
     attr_writer :variant_search_class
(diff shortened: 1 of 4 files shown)


tvdeyen:
👍

core/spec/lib/tasks/migrations/create_vat_prices_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+describe 'solidus:migrations:create_vat_prices' do
+  describe 'up' do
+    include_context(
+      'rake',
+      task_path: Spree::Core::Engine.root.join('lib/tasks/migrations/create_vat_prices.rake'),
+      task_name: 'solidus:migrations:create_vat_prices:up',
+    )
+
+    # TODO: Update to work if this is missing, and add more specs
```
mamhoff: If that zone is missing, the task should silently return.

mamhoff:
👍

PR #1250, opened 2016-06-15
Extract reports from controller

isaacfreeman:
When we have more than a handful of reports for a Solidus site, we typically end up with a long and messy decorator for the reports controller. This PR introduces a structure that allows us to extract the logic for each report into its own class, and declare which reports we want to be active in an initializer.

I appreciate this isn't ready to merge yet, but I'm opening the pull request for comment on whether this is a desirable approach. It's worked well for me with clients, but I don't know how universal my experience is. What do you think?

Notes:
- Specs currently aren't passing
- I've aimed to maintain backwards compatibility with existing decorators that use `Spree::ReportsController.add_available_report!`
- There's some metaprogramming to generate an action in the controller for each generated report. This could perhaps be made less magical by simply using a show method that receives the desired report as a param, but this would break backwards compatibility.
- I've added a second report showing variants that are out of stock. I'm not assuming that this specific report needs to also be merged into Solidus, and I'm open to removing it or moving it to a separate PR even if there's consensus on the general approach. I'll leave it in for now to illustrate how additional reports can be added.
- The seed data currently doesn't have any out-of-stock variants, so you'll see no rows in the table until you make one. In the console for the sandbox app, run `Spree::StockItem.first.set_count_on_hand(0)`.
- The out-of-stock variants report uses a `table_report` view that's intended to be generally useful for any report that generates a table.

### DIFF
--- a/backend/app/controllers/spree/admin/reports_controller.rb
+++ b/backend/app/controllers/spree/admin/reports_controller.rb
@@ -35,7 +35,7 @@ module Spree
 
       def show
         report_class = "Spree::Report::#{params[:id].camelize}".safe_constantize
-        if report_class.present?
+        if report_class
           report = report_class.new(params_for_report)
           respond_to do |format|
             format.html do
--- a/backend/app/controllers/spree/admin/reports_controller.rb
+++ b/backend/app/controllers/spree/admin/reports_controller.rb
@@ -34,7 +34,7 @@ module Spree
       end
 
       def show
-        report_class = "Spree::Reports::#{params[:id].camelize}".constantize
+        report_class = "Spree::Report::#{params[:id].camelize}".constantize
         report = report_class.new(params)
         respond_to do |format|
           format.html do
--- a/backend/app/controllers/spree/admin/reports_controller.rb
+++ b/backend/app/controllers/spree/admin/reports_controller.rb
@@ -12,62 +12,41 @@ module Spree
           if report_description_key.nil?
             report_description_key = "#{report_key}_description"
           end
-          @@available_reports[report_key] = { name: Spree.t(report_key), description: Spree.t(report_description_key) }
+          @@available_reports[report_key] = { name: Spree.t(report_key), description: report_description_key }
         end
       end
 
       def initialize
         super
-        ReportsController.add_available_report!(:sales_total)
+        Rails.application.config.spree.reports.each do |report|
+          Spree::Admin::ReportsController.add_available_report!(
+            report.name.demodulize.underscore.to_sym,
+            report.description
+          )
+        end
+      end
+
+      # Generate a method for each registered report
+      Rails.application.config.spree.reports.each do |report_class|
+        define_method(report_class.name.demodulize.underscore) do
+          report = report_class.new(params)
+          respond_to do |format|
+            format.html do
+              report.locals.each do |key, value|
+                instance_variable_set("@#{key}", value)
+              end
+              render report_class.template
+            end
+          end
+        end
       end
 
       def index
         @reports = ReportsController.available_reports
       end
 
-      def sales_total
-        params[:q] = {} unless params[:q]
-
-        if params[:q][:completed_at_gt].blank?
-          params[:q][:completed_at_gt] = Time.current.beginning_of_month
-        else
-          params[:q][:completed_at_gt] = begin
-                                           Time.zone.parse(params[:q][:completed_at_gt]).beginning_of_day
-                                         rescue
-                                           Time.current.beginning_of_month
-                                         end
-        end
-
-        if params[:q] && !params[:q][:completed_at_lt].blank?
-          params[:q][:completed_at_lt] = begin
-                                           Time.zone.parse(params[:q][:completed_at_lt]).end_of_day
-                                         rescue
-                                           ""
-                                         end
-        end
-
-        params[:q][:s] ||= "completed_at desc"
-
-        @search = Order.complete.ransack(params[:q])
-        @orders = @search.result
-
-        @totals = {}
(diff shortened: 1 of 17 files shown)


mtomov:
I like the approach! My suggestions would be:

- keep it to a single `show` action - it gets messy to have a controller action for every report. Metaprogramming helps indeed, but why not do it properly? I'm not sure that you will even be breaking compatibility, if you just add `show`, old reports with custom actions should theoretically work, no?

- have a base Reporter class, which would define the common interface `self.description`, `self.template`, `new`, `rows` .., and can contain some helpful methods, such as `def spree_routes; Spree::Core::Engine.routes.url_helpers; end;`

- rename `def locals` to something more revealing, like `data`, or something better

- table report is great!

backend/app/views/spree/admin/reports/table_report.html.erb
```diff
@@ -0,0 +1,58 @@
...
+      <%= submit_tag "Filter" %>
+    </div>
+  <% end %>
+<% end %>
+<div class="report orders-report">
+  <%= paginate @rows %>
+  <table style="table-layout: fixed">
+    <colgroup>
+      <% @headers.each do |header| %>
+        <col style="width: <%= header.values.first.length * 0.8 %>em">
```
mtomov: Wouldn't `table-layout: fixed` take care of the column widths?

backend/config/routes.rb
```diff
@@ -1,4 +1,4 @@
-Spree::Core::Engine.routes.draw do
+Spree::Core::Engine.add_routes do
```
mtomov: In the meantime, this [got deprecated](https://github.com/solidusio/solidus/blob/f831961f0c55dfcfeaf72bbb0804e7e325a6c671/core/lib/spree/core/routes.rb#L5)

backend/config/routes.rb
```diff
@@ -151,10 +151,12 @@
...
 
-    resources :reports, only: [:index] do
+    resources :reports, only: :index do
       collection do
-        get :sales_total
-        post :sales_total
+        Rails.application.config.spree.reports.each do |report|
+          get report.to_s.demodulize.underscore.to_sym
+          post report.to_s.demodulize.underscore.to_sym
+        end
```
mtomov: Mentioned it also in the other comment, but an alternative here is to keep the existing code as it is, just add the Deprecated warning to both `get sales_total` and `post sales_total`, and add a `show` method action, which would work without meta-programming.

isaacfreeman:
@mtomov: Thanks! Agree on all your points. In particular, you're right about the show method – I was too quick to assume I couldn't do it for backward compatibility reasons, but of course you're right that it's not a problem.

How would you feel about simply using `BaseReport` as a parent class name rather than `Reporter`? It's closer to the domain object as understood by users, and the :base: prefix is used elsewhere in Solidus for similar purposes.

Also, would it be more convenient to continue with this PR, or close it while I follow up and open a new one later?

mtomov:
Yes, either `Spree::BaseReporter` or maybe even `Spree::Reports::Base`. Maybe the singular is even better - `Spree::Report::Base` - your call.

By all means, do continue with this one. We can squash it all at the end.

mtomov:
Btw, if we are to consider an external dependency, this gem [Dossier](https://github.com/tma1/dossier) was recommended to me.

isaacfreeman:
@mtomov: Dossier looks good, but as you say it's an extra dependency, and I'd rather not add dependencies to the main Solidus distribution. However, if we go ahead with #1262 and move reports to their own `spree_reports` gem, I'd feel a lot more comfortable adding dependencies to that. We'd know then that people are opting in to using the reports, and we wouldn't be adding the dependency for people who'll never use it.

backend/app/controllers/spree/admin/reports_controller.rb
```diff
@@ -34,19 +34,34 @@ def index
...
 
       def show
-        report_class = "Spree::Report::#{params[:id].camelize}".constantize
-        report = report_class.new(params)
-        respond_to do |format|
-          format.html do
-            report.content.each do |key, value|
-              instance_variable_set("@#{key}", value)
+        report_class = "Spree::Report::#{params[:id].camelize}".safe_constantize
+        if report_class.present?
```
gmacdougall: This can just be `if report_class`

backend/spec/controllers/spree/admin/reports_controller_spec.rb
```diff
@@ -32,107 +40,28 @@
...
+      spree_get :show, id: :sales_total
+      expect(response).to be_ok
+    end
+    it 'passes params to the report' do
+      expect_any_instance_of(Spree::Report::SalesTotal).to receive(:initialize).with("foo" => "bar")
+      spree_get :show, { id: :sales_total, foo: :bar }
+    end
+    it "returns an error if there's no such report" do
+      spree_get :show, id: :some_report
+      expect(response).to redirect_to spree.admin_reports_path
```
gmacdougall: I would also like to see an expectation on the flash here.

backend/app/controllers/spree/admin/reports_controller.rb
isaacfreeman: I've always found `.present?` a bit more readable, but I'm happy to match code style elsewhere in Solidus.

backend/spec/controllers/spree/admin/reports_controller_spec.rb
isaacfreeman: Good idea, will do.

tvdeyen:
We want to extract reports into it's own gem like stated in #1262

PR #1251, opened 2016-06-15
Deprecate Variant.active, Variant#price_in, Variant#amount_in

mamhoff:
As noted in Slack by @jrochkind , all of these methods have an alternative now that can deal with more than just currency as a price determinator, and it's very hard to find out what that alternative is if you're not me. That's what these deprecations are doing.

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -94,9 +94,11 @@ module Spree
     # Returns variants that are not deleted and have a price in the given
     # currency.
     #
+    # @deprecated Please use the .with_prices scope instead
     # @param currency [String] the currency to filter by; defaults to Spree's default
     # @return [ActiveRecord::Relation]
     def self.active(currency = nil)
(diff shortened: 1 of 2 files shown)


jhawthorn:
:+1: Should we cherry-pick these back onto 1.3?

core/app/models/spree/variant.rb
```diff
@@ -94,9 +94,11 @@ def self.in_stock(stock_locations = nil)
     # Returns variants that are not deleted and have a price in the given
     # currency.
     #
+    # @deprecated Please use the .with_prices scope instead
     # @param currency [String] the currency to filter by; defaults to Spree's default
     # @return [ActiveRecord::Relation]
     def self.active(currency = nil)
+      Spree::Deprecation.warn("`Variant.with_prices(currency)` is deprecated. Please use `Variant.with_prices(pricing_options)` instead.", caller)
```
jasonfb: don't you mean to write ```Variant.active(currency)...``` instead

mamhoff: Absolutely, thanks!

mamhoff:
@jhawthorn I think so, they're a usability plus. :)

PR #1253, opened 2016-06-16
Remove shipping_method_id as permitted checkout attribute

jordan-brough:
I'm not sure what this could be used for. Orders don't have a shipping_method_id attribute do they?


bbuchalter:
Not in a long time @jordan-brough: https://github.com/solidusio/solidus/blob/60ee1dffc7c3932cd7652cc2b4566f08a12f74b8/core/db/migrate/20150626200816_remove_shipping_method_id_from_spree_orders.rb#L3

jhawthorn:
:+1:

PR #1256, opened 2016-06-16
Remove deprecated call to ActiveMerchant::Billing::Base.gateway_mode

mamhoff:
This has been deprecated in
https://github.com/activemerchant/active_merchant/commit/03883cb66e58002cf9e040ae0cb02a38f2413a1c

It does not raise an exception as they use their own deprecation class, but we should not call
deprecated methods in other gems, either.


jhawthorn:
:+1:

It looks like this has existed at least as far back as [1.48.0](https://github.com/activemerchant/active_merchant/blob/v1.48.0/lib/active_merchant/billing/base.rb#L26), our oldest supported version of AM.

PR #1257, opened 2016-06-16
Clean up records in promotion rules

cbrunsdon:
Product was the only one that had the (correct) dependent destroy on it.
When a rule is deleted we should no longer reference it in the join
table.


bbuchalter:
Is there any value in a migration to help clean up folks data?

mamhoff:
I'm halfway about a migration. The data doesn't really hurt I guess - once it bites somebody, I'd wait for that person to PR the migration.

jhawthorn:
A migration could be good, but I don't think that should block this PR.

:+1:

PR #1258, opened 2016-06-16
Always validate count_on_hand of StockItem

jhawthorn:
Previously, `count_on_hand` was only validated if it was changed and specifically if it was changed to a lower number than it was previously.

This isn't a good approach. A record should be valid based on it's current attributes. It shouldn't be considered valid just because it's "more" or "less" valid than before. The existing code would allow, for
example, creating a valid backorderable `StockItem` with a negative count_on_hand, and then setting it to be `backorderable=false`. This is an invalid state but would be allowed because count_on_hand wasn't changed.

This commit just changes the validation to just always run when `backorderable=false`.

### DIFF
--- a/core/app/models/spree/stock_item.rb
+++ b/core/app/models/spree/stock_item.rb
@@ -8,7 +8,7 @@ module Spree
 
     validates :stock_location, :variant, presence: true
     validates :variant_id, uniqueness: { scope: [:stock_location_id, :deleted_at] }, allow_blank: true, unless: :deleted_at
-    validates :count_on_hand, numericality: { greater_than_or_equal_to: 0 }, if: :verify_count_on_hand?
+    validates :count_on_hand, numericality: { greater_than_or_equal_to: 0 }, unless: :backorderable?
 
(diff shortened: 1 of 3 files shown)


core/app/models/spree/stock_item.rb
```diff
@@ -8,7 +8,7 @@ class StockItem < Spree::Base
 
     validates :stock_location, :variant, presence: true
     validates :variant_id, uniqueness: { scope: [:stock_location_id, :deleted_at] }, allow_blank: true, unless: :deleted_at
-    validates :count_on_hand, numericality: { greater_than_or_equal_to: 0 }, if: :verify_count_on_hand?
+    validates :count_on_hand, numericality: { greater_than_or_equal_to: 0 }, unless: :backorderable?
```
Senjai: The negative amount usually indicates that that many items are backorderd. If we adjusted below that, that would no longer mean the same thing. It makes sense to adjust positive amounts to fill backorders but I don't think it makes sense to adjust negatively.

Senjai:
:+1:

mamhoff:
👍

PR #1260, opened 2016-06-20
Modify factory completed_order_with_promotion

graygilmore:
The existing factory will not create a `Spree::OrderPromotion`. That was the original issue that lead into this refactor. I *could* have just added a `Spree::OrderPromotion.create()` at the bottom of the factory but I liked the fact that this change brings the factory more in line with a real world usage.

### DIFF
--- a/core/lib/spree/testing_support/factories/order_factory.rb
+++ b/core/lib/spree/testing_support/factories/order_factory.rb
@@ -108,7 +108,7 @@ FactoryGirl.define do
     end
   end
 
-  factory :completed_order_with_promotion, parent: :completed_order_with_totals, class: "Spree::Order" do
+  factory :completed_order_with_promotion, parent: :order_with_line_items, class: "Spree::Order" do
     transient do
       promotion nil
     end
@@ -118,6 +118,11 @@ FactoryGirl.define do
       promotion_code = promotion.codes.first || create(:promotion_code, promotion: promotion)
 
       promotion.activate(order: order, promotion_code: promotion_code)
+
+      # Complete the order after the promotion has been activated
+      order.refresh_shipment_rates
+      order.update_column(:completed_at, Time.current)
+      order.update_column(:state, "complete")
     end
   end
 end
--- a/core/lib/spree/testing_support/factories/order_factory.rb
+++ b/core/lib/spree/testing_support/factories/order_factory.rb
@@ -111,16 +111,13 @@ FactoryGirl.define do
   factory :completed_order_with_promotion, parent: :completed_order_with_totals, class: "Spree::Order" do
     transient do
       promotion nil
-      promotion_code nil
     end
 
     after(:create) do |order, evaluator|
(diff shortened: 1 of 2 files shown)


core/lib/spree/testing_support/factories/order_factory.rb
```diff
@@ -118,9 +118,12 @@
       promotion = evaluator.promotion || create(:promotion, code: "test")
       promotion_code = evaluator.promotion_code || promotion.codes.first
```
graygilmore: This is no longer needed since `.activate()` takes care of the promo code.

adammathys:
:+1: From me. Definite improvement over the original factory.

mamhoff:
👍  Great!

core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
```diff
@@ -45,4 +45,10 @@
 
     it_behaves_like 'a working factory'
   end
+
+  describe 'completed order with promotion' do
+    let(:factory) { :completed_order_with_promotion }
+
+    it_behaves_like 'a working factory'
+  end
```
bbuchalter: We seem to have a convention of not putting factories under the same level of testing as other parts of the system. Is that an intentional choice? If we were fixing any other regression outside of factories, wouldn't we be adding more robust tests?

mamhoff: No, we only recently put the factories under test, and the coverage is not perfect yet. They're very complex, and we'll happily accept contributions towards improving both the factories and their specs. Yes, the test suite itself kinda sorta tests the factories, but that's all very implicit and meh. 

This change is good as far as I'm concerned, but it does not test the improvements to the factory. @graygilmore If you want to add specs that test your improvements, I'd be happy - but I don't see that as a blocker for an obvious improvement.

graygilmore: I had definitely considered adding more robust tests but I wasn't sure if that's something we wanted to do for our factories. If anyone would like to see more work added to this PR just let me know.

mamhoff: I'll bring it up at the core meeting today, deferring until after that.

mamhoff: So we agreed we do want to have better specs for factories for sure. @graygilmore Just give us a heads-up if you're willing to spec out this factory a little more; @jhawthorn has given a fine example in #1264 on how to do this beautifully.

graygilmore: Great, I'll improve the specs using that example.

jhawthorn:
Thanks Gray. I'll add an extra spec in #1264 after rebasing

Issue #1167, opened 2016-05-19
Admin Pricing TODO


jhawthorn:
#1117 introduced a new interface for editing prices. Obviously editing prices is important, so we should try to make the interface as good as possible.

Some immediate suggestions:
- [x] Validate entered currency (it's possible currently to make a price in an invalid currency, ex. "XYZ") #1169
- [ ] Grouping common prices (@Dkendal and @Mandily did some investigation of this)
- [ ] We have the master price both on the main product page as well as the pricing page. This will probably be cause of some confusion.
- [ ] Showing "Master Variant" in the pricing table feels like we're leaking an implementation detail

mamhoff:
#1169 should fix the first issue. I've also amended the #1117  to use a `select2` instead of a text box for input.

Mandily:
I'd like to add two things to the to do list. The first one may solve @jhawthorn's last point above.

- [ ] Pull the master variant pricing and cost price off of the main product page, and out of the variant table to an area all it's own at the top of the page. 

This would also include a new feature I'd like to start implementing in the back end which is tooltips for more information. Great information for newbies, totally unneeded and thus out of the way for the experienced users.

<img width="464" alt="move master variant" src="https://cloud.githubusercontent.com/assets/15271353/15514717/4dcc1c9e-219f-11e6-885f-442adc3d1fad.png">

- [ ] Add an option types summary area at the top of the page so that the user can tell at a quick glance what's being used in the table below. I suspect the table could get messy at times, and a quick snapshot of what they're working with would be handy. This should fit to the right of the master variant pricing.

<img width="314" alt="option types summary" src="https://cloud.githubusercontent.com/assets/15271353/15514770/8b105c8c-219f-11e6-8d6d-ac97a6551dba.png">

jhawthorn:
We've made progress, as can be seen in the linked merged issues.

There's always going to be more we can do to improve the UI here, and this issue no longer has actionable steps forward.

Marking this done. Thanks everyone.

PR #1261, opened 2016-06-21
Add hint tooltips component to admin

jhawthorn:
@Mandily requested in #1167 that we add hint tooltips to provide more information in the admin.

This should be helpful to clarify the purpose of fields in the admin, which are not always obvious. For an example I've added a hint to `Product#promotionable` bca7849.

This adds both a helper `admin_hint(title, text)`, as well as an extra form helper method `<%= f.field_hint :field_name %>`. The form helper version looks up text in I18n under `en.spree.hints.#{model_scope}.#{method}`.


tvdeyen:
This is a great addition. 👍

mtomov:
Indeed :+1:

mamhoff:
👍

PR #1263, opened 2016-06-22
Remove broken `accessible_by` code for UserManagement role

jordan-brough:
The following code attempts to make the `update_email` permission work
with `accessible_by` but I don't think it actually works:

    # due to how cancancan filters by associations,
    # we have to define this twice, once for `accessible_by`
    can :update_email, Spree.user_class, spree_roles: { id: nil }
    # and once for `can?`
    can :update_email, Spree.user_class do |user|
      user.spree_roles.none?
    end

If the first `can` fails then `accessible_by` will move on to the
second one (the block form) and fail with this error:

    CanCan::Error:
    The accessible_by call cannot be used with a block 'can' definition.
    The SQL cannot be determined for :update_email Spree::LegacyUser

### DIFF
--- a/core/lib/spree/permission_sets/user_management.rb
+++ b/core/lib/spree/permission_sets/user_management.rb
@@ -4,10 +4,8 @@ module Spree
       def activate!
         can [:admin, :display, :create, :update, :save_in_address_book, :remove_from_address_book, :addresses, :orders, :items], Spree.user_class
 
-        # due to how cancancan filters by associations,
-        # we have to define this twice, once for `accessible_by`
-        can :update_email, Spree.user_class, spree_roles: { id: nil }
-        # and once for `can?`
+        # Note: This does not work with accessible_by.
+        # See https://github.com/solidusio/solidus/pull/1263
         can :update_email, Spree.user_class do |user|
           user.spree_roles.none?
         end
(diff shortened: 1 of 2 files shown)


tvdeyen:
You should be able to combine both into one `can` definition.

jordan-brough:
@tvdeyen how? If I remove the first `can` definition then `accessible_by` blows up.  If I remove the second `can` definition then the existing specs (before this PR) fail.

cc @Senjai since he maintains cancancan

jordan-brough:
Also, if I try to do this:
```ruby
can(:update_email, Spree.user_class, spree_roles: { id: nil }) do
  user.spree_roles.none?
end
```
then this error happens:
> CanCan::Error:
> You are not able to supply a block with a hash of conditions in update_email Spree::LegacyUser ability. Use either one.

tvdeyen:
Oh, sorry. It has to be a scope:

https://github.com/AlchemyCMS/alchemy_cms/blob/master/lib/alchemy/permissions.rb#L40-L42

Senjai:
@tvdeyen is correct.

@jordan-brough I made this work as a POC with this https://gist.github.com/Senjai/ec55a807e71c26224173539a0a0b814b

Feel free to use it if you want to. This stuff needs to be made better in cancancan and I'll throw it on the slate for 2.0.

jordan-brough:
@tvdeyen @Senjai thanks!  Great to know.

Perhaps in this case we should just yank the first `can` and not worry about `accessible_by` compatibility?  We obviously don't use `accessible_by` with this, since it was broken, and it feels weird to duplicate permission logic in two places.  We do have other `can`s that use blocks and don't provide `accessible_by` compatibility.  I'm OK with either way personally though.  Thoughts?

tvdeyen:
Yes. I'm ok with adding them, when we actually need them.

jordan-brough:
@tvdeyen updated. Look good?

PR #1264, opened 2016-06-22
Add specs on the order factories

jhawthorn:
Previously we just had smoke tests on factories to ensure that they didn't error. However out factories are rather complex, and often don't give quite realistic data. Since users of Solidus depend on these factories to work properly, we should be testing them thoroughly.

This commit adds some spec to the records created by the various order factories.

### DIFF
--- a/core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
+++ b/core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
@@ -8,42 +8,186 @@ RSpec.describe 'order factory' do
     let(:factory) { :order }
 
     it_behaves_like 'a working factory'
+
+    it "has the expected attributes" do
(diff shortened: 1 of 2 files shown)


core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
```diff
@@ -8,41 +8,185 @@
     let(:factory) { :order }
 
     it_behaves_like 'a working factory'
+
+    it "has the expected attributes" do
+      order = create(factory)
+      aggregate_failures do
```
tvdeyen: WAT? `aggregate_failures` is just awesome and I didn't know, yet? Thanks

mamhoff:
This is excellent. 👍

bbuchalter:
👍

PR #1269, opened 2016-06-23
Rabase `Rails5` onto `master`

Kingdutch:
I ran a rebase onto master. Sadly `DB=sqlite build bash.sh` is showing quite some rspec errors. I'm not sure if these were already there or whether I introduced them. 

I tried to resolve merge conflicts by looking at the issues that spawned them and the commits for both branches that conflicted and ending up with the intended code.

For version conflicts I used the highest version (i.e. Ruby 2.3.0 instead of 2.2.0 used by `rails5`)


jhawthorn:
@Kingdutch thank you. I've force pushed this to the rails 5 branch.

The spec failures already existed, and the branch is still in a pretty broken state. Notably there's some weirdness going on with state machine (when it gets redefined specifically) that I haven't been able to diagnose yet.

Issue #1266, opened 2016-06-23
Add hint tooltips to fields in the admin


jhawthorn:
In #1261 we introduced hint tooltips to help describe fields in the admin.

![](http://i.hawth.ca/s/UnN2Qttw.png)

Currently there's only one defined (Product promotionable), you can see aaef78d835444dd669c83ba0a49e88508892e150 as an example of how to add one. We don't want hints on every field, but it would be nice to have them on all fields whose purpose is non-obvious.

agustifernandez:
So a PR to add tooltips would make sense if someone was to give this a try? I would like to give back to core but need easy/simple entry-level stuff to be able to commit to it and success.

mamhoff:
@agustifernandez Yes, it'd be great if you could take a go at this. Good examples for not-so-easy-to understand form fields are: 

* `Variant#tax_category` (is an override, if not set, takes over the product)
* `Product#shipping_category`
* `Product#tax_category`
* `Price#country` (if empty, the price is valid for all countries)

jhawthorn:
We have added a few, I don't think this issue is helpful to leave open without specific recommendations

PR #1271, opened 2016-06-24
Add some admin hint tooltips

mtylty:
As requested in #1266, I've added the hint tooltips commented by @mamhoff.

### DIFF
--- a/backend/app/views/spree/admin/prices/_form.html.erb
+++ b/backend/app/views/spree/admin/prices/_form.html.erb
@@ -24,6 +24,7 @@
     <div class="three columns" data-hook="admin_product_price_form_country">
       <%= f.field_container :country do %>
         <%= f.label :country %>
+        <%= f.field_hint :country %>
         <%= f.collection_select :country_iso, available_countries, :iso, :name,
                                 {
                                   include_blank: t(:any_country, scope: [:spree, :admin, :prices]),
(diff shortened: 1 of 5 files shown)


Mandily:
Can we be consistent with the use of capitalization and use a capital letter after the colons in the 'default' messaging? I could be swayed to use all lowercase if someone thinks that's better. What I'm after here is consistency.

jhawthorn:
Thank you. Other than Amanda's suggestions this looks great :+1:

mtylty:
@Mandily I've asked myself the same question about capitalization.

I've ended up with a mixed solution because I've used capitalization whenever the default was the exact string (i.e. `Any Country` appears in the select dropdown) and lowercase to follow the example of the `checked` default provided in aaef78d835444dd669c83ba0a49e88508892e150.

I'm concerned with consistency as well. Should I change the lowercase defaults to uppercase? It makes more sense to me to display the exact string when the default is part of the available choices.

Mandily:
@mtylty Instead of being consistent with the inconsistency, how about we clean up the original strings if they need it too? Capitalization is related more to 1. whether the word is a proper noun and 2. the placement of the word in a sentence, title etc. I realize there's some less than desirable things in the admin right now, but I don't think we need to continue those bad habits in the meantime of cleaning them up.

gmacdougall:
:+1:

mamhoff:
👍

jhawthorn:
Thanks @mtylty

PR #1274, opened 2016-06-27
Include all adjustments in ItemAdjustments in adjustment_total

jordan-brough:
Why don't we include all adjustments in `ItemAdjustments`?

I'm opening this PR to see if anyone knows and/or what breaks if we do just include all adjustments.

### DIFF
--- a/core/app/models/spree/item_adjustments.rb
+++ b/core/app/models/spree/item_adjustments.rb
@@ -60,7 +60,10 @@ module Spree
 
       adjustments.select(&:cancellation?).map(&:update!)
 
-      @item.adjustment_total = @item.adjustments.eligible.sum(:amount)
+      @item.adjustment_total = @item.adjustments.
+        eligible.
+        reject(&:included?).
+        sum(&:amount)
 
       @item.update_columns(
         promo_total: @item.promo_total,
--- a/core/app/models/spree/item_adjustments.rb
+++ b/core/app/models/spree/item_adjustments.rb
@@ -60,7 +60,7 @@ module Spree
 
       adjustments.select(&:cancellation?).each(&:update!)
 
-      @item.adjustment_total = @item.adjustments.
+      @item.adjustment_total = adjustments.
         select(&:eligible?).
         reject(&:included?).
         sum(&:amount)
--- a/core/app/models/spree/item_adjustments.rb
+++ b/core/app/models/spree/item_adjustments.rb
@@ -61,7 +61,7 @@ module Spree
       adjustments.select(&:cancellation?).map(&:update!)
 
       @item.adjustment_total = @item.adjustments.
-        eligible.
+        select(&:eligible?).
         reject(&:included?).
         sum(&:amount)
 
--- a/core/app/models/spree/item_adjustments.rb
+++ b/core/app/models/spree/item_adjustments.rb
@@ -58,6 +58,8 @@ module Spree
       @item.included_tax_total = tax.select(&:included?).map(&:update!).compact.sum
       @item.additional_tax_total = tax.reject(&:included?).map(&:update!).compact.sum
 
+      adjustments.select(&:cancellation?).map(&:update!)
+
       @item.adjustment_total = @item.adjustments.eligible.sum(:amount)
 
       @item.update_columns(
--- a/core/app/models/spree/item_adjustments.rb
+++ b/core/app/models/spree/item_adjustments.rb
@@ -58,9 +58,7 @@ module Spree
       @item.included_tax_total = tax.select(&:included?).map(&:update!).compact.sum
       @item.additional_tax_total = tax.reject(&:included?).map(&:update!).compact.sum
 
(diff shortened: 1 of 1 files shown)


core/app/models/spree/item_adjustments.rb
```diff
@@ -58,9 +58,7 @@ def update
       @item.included_tax_total = tax.select(&:included?).map(&:update!).compact.sum
       @item.additional_tax_total = tax.reject(&:included?).map(&:update!).compact.sum
 
-      item_cancellation_total = adjustments.select(&:cancellation?).map(&:update!).compact.sum
```
bbuchalter: I would imagine tests are failing because of the lost call to `update!`. I do think there is room for improvement on how this is done though.

jordan-brough: @bbuchalter thanks!

core/app/models/spree/item_adjustments.rb
```diff
@@ -58,9 +58,12 @@ def update
...
+      adjustments.select(&:cancellation?).map(&:update!)
 
-      @item.adjustment_total = @item.promo_total + @item.additional_tax_total + item_cancellation_total
+      @item.adjustment_total = @item.adjustments.
+        eligible.
```
bbuchalter: Will the use of this AR scope trigger unnecessary SQL calls and Ruby object instantiation? We already have these in memory, yes?

jordan-brough:
TODO: Call `update!` on all all adjustments, not just promo/tax/cancellation adjustments.  I'm not sure how we best figure out what order to call `update!` on things though, when there are adjustments of unknown types.  Taxes probably need to be last.  That might be part of the reason why this wasn't done in an open-ended way to begin with?

core/app/models/spree/item_adjustments.rb
```diff
@@ -58,9 +58,12 @@ def update
...
+        reject(&:included?).
```
bbuchalter: What's the `included` column on adjustment do?

core/app/models/spree/item_adjustments.rb
jordan-brough: @bbuchalter updated. Thanks!

core/app/models/spree/item_adjustments.rb
jordan-brough: @bbuchalter it's really only for taxes.  It means that the adjustment is already built into the price of the item, so we don't count it toward the adjustment_total.

jordan-brough: e.g. VAT in Europe is an "included" tax whereas U.S. sales tax is an "additional" tax, in Solidus terms.

core/app/models/spree/item_adjustments.rb
```diff
@@ -58,9 +58,12 @@ def update
       @item.included_tax_total = tax.select(&:included?).map(&:update!).compact.sum
       @item.additional_tax_total = tax.reject(&:included?).map(&:update!).compact.sum
 
-      item_cancellation_total = adjustments.select(&:cancellation?).map(&:update!).compact.sum
+      adjustments.select(&:cancellation?).map(&:update!)
 
-      @item.adjustment_total = @item.promo_total + @item.additional_tax_total + item_cancellation_total
+      @item.adjustment_total = @item.adjustments.
```
jhawthorn: This should probably just be `adjustments`, which delegates and caches `item.adjustments`

core/app/models/spree/item_adjustments.rb
```diff
@@ -58,9 +58,12 @@ def update
       @item.included_tax_total = tax.select(&:included?).map(&:update!).compact.sum
       @item.additional_tax_total = tax.reject(&:included?).map(&:update!).compact.sum
 
-      item_cancellation_total = adjustments.select(&:cancellation?).map(&:update!).compact.sum
+      adjustments.select(&:cancellation?).map(&:update!)
```
jhawthorn: Since we aren't using the result this should be `each`

core/app/models/spree/item_adjustments.rb
jordan-brough: @jhawthorn thanks! Updated.

core/app/models/spree/item_adjustments.rb
jordan-brough: updated!

jordan-brough:
Note: I'm doing this mostly to see how much stuff breaks.  Specs are passing so that's good.  If anyone would like to pick this up and carry it forward that would be great. Or I'll try to think about it / look into it more when I can.

bbuchalter:
> Call update! on all all adjustments, not just promo/tax/cancellation adjustments. I'm not sure how we best figure out what order to call update! on things though

I know our application had to override parts of the old order updater to get adjustments to calculate in the sequence we need. However this gets, it should be easily customizable and composable. `ItemAdjustments` is an improvement, perhaps we can go even further.

PR #1275, opened 2016-06-27
Remove clear cache button

jhawthorn:
Our current "General Settings" page has a "Clear cache" button.

On stores I've worked on I wouldn't want this exposed to admins. We should (and afaik we do) expire caches every time the underlying DB object is changed.

This is motivated partly because I want to replace the `GeneralSettingsController` with a `StoreController`, and this is the only thing on that page that doesn't directly relate to a store.

Though this should be trivial to replace in one's own store, I'd love to hear if anyone considers this important in their admin or would be inconvenienced by it being removed.


bbuchalter:
We built our own clear cache button, but only developers used it. Eventually, we just built it into the deploy process. 👍 for removal.

jhawthorn:
Another point I should mention is that stores with heavy custom caching (which is who I assume this is designed for) are likely to need more than just `Rails.cache.clear`. The [existing specs hint at this](https://github.com/solidusio/solidus/blob/0a52dc579dd6ced31af57ea013b43b85085dbacd/backend/spec/controllers/spree/admin/general_settings_controller_spec.rb#L30). It would be simpler for these stores just to implement their own cache clear button than to try and hack into this existing one.

jasonfb:
👍 for removing button. Only developers should need to do platform operations like clearing the Rails cache, and we have access to the Rails console for this. Too dangerous to give non-developers things they can mess things up with (we have enough problems with them creating Promos that break our checkout!)

gmacdougall:
:+1:

jhawthorn:
Rebased and added note to changelog

Issue #1120, opened 2016-05-09
Admin Organization and Structure


graygilmore:
The navigation in the admin is accomplished through several partials that pass around regexes to determine what the "active" page is. This is both confusing to debug and difficult to modify and improve.

Our lack of structure in this area comes across to users of the system as things aren't necessarily in a logical place. For example, [this PR](https://github.com/solidusio/solidus/pull/1041) addressed an issue where navigation items were being duplicated under different hierarchies.

Running through a site architecture exercise should allow us to:
1. More easily add [breadcrumbs](https://github.com/solidusio/solidus/issues/1076)
2. Clean up menu partials ([like this one](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/shared/_settings_sub_menu.html.erb))
3. Provide a platform moving forward to add/remove new pages to the admin

Mandily:
I've started recruiting admin users for a cart sort that should help us organize the admin better. Will post back here when I have some results. We're looking for admin users, not developers that code on the platform. If that sounds like you or you know of anyone who wouldn't mind spending 10 - 15 minutes of time, please forward. 
The survey will be open until Friday, June 10th at midnight.

https://30336t5m.optimalworkshop.com/optimalsort/solidus-nav

tvdeyen:
@Mandily great idea. I forwarded the link to our "non-devs"

Mandily:
I looked at the results of the cardsort and saw that while there were some obvious moves and label changes, there was also some items that were split between being located in settings vs products. I ran another type of test, called a tree test where the navigation was set, and users were given tasks to follow to see if they went in the right direction. 

Again, not everyone thought the same way, but in creating a consistent convention of how we organize the navigation we should be able to account for the mental models of both groups. 

I've created a new navigation in a [google spreadsheet](https://docs.google.com/spreadsheets/d/15RBbFOfwqdtyRkJ3sPo5vxF-Uu8mK8suC53a5dN7KVo/edit#gid=0) with highlights for label changes. The new navigation is based around a structure where 'set it and forget' items (for example, things that end in 'types', 'categories', and 'methods') get created in 'settings' and then specifics for those get set in their respective section of the nav (for example, under orders and products). It's seems to be intuitive for about half of users, and the consistency should make it easier to learn by those it's not intuitive for. 

[...]

PR #1276, opened 2016-06-27
Change navigation and button labels

Mandily:
This is a precursor to #1120 

Some labels were inconsistent and some were confusing.
Based on a recent card sort and tree test on the navigation, I was able to figure out where some problem areas were. I've fixed the inconsistencies and made the confusing labels more clear.

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -415,8 +415,8 @@ en:
         one: Promotion Category
         other: Promotion Categories
       spree/property:
-        one: Property
-        other: Properties
+        one: Property Type
+        other: Property Types
       spree/prototype:
         one: Prototype
         other: Prototypes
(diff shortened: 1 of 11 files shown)


jhawthorn:
:+1: Looks good to me

core/config/locales/en.yml
```diff
@@ -476,8 +476,8 @@ en:
         one: Tax Rate
         other: Tax Rates
       spree/taxon:
-        one: Taxon
-        other: Taxons
+        one: Display Order
+        other: Display Order
```
gmacdougall: I don't believe that this change is safe to make.

Mandily: Safe from a technical or user perspective? I'm not sure of an alternative if it's code, but if it's the user you're worried about, we could use an interim 'Display Order (Taxon)' label that would be intuitive for familiar users and new users alike.

gmacdougall:
There are some failing tests now that the Taxon translation change has been reverted. Other than that, this is looking good.

PR #1279, opened 2016-06-28
Remove references to expand_variant

jhawthorn:
The JS this was designed for was removed in 2012 5f6bbdb12d50326af0df5e25ae4471ea687a7e1f.


mamhoff:
👍 Thanks John for digging these up :)

PR #1280, opened 2016-06-28
Rename before_filter/around_filter to before_action/around_action in controllers

adammiribyan:
Since before/around/after_filter callbacks are gonna be [deprecated](https://github.com/rails/rails/blob/v5.0.0.beta2/actionpack/lib/abstract_controller/callbacks.rb#L190-L193) soon, I thought this could be a nice improvement to keep things up to date.


jhawthorn:
:+1: Thank you. Looks like we had a mix of `*_filter` and `*_action` before.

mamhoff:
👍

PR #1281, opened 2016-06-28
Specify template for fresh_when

jhawthorn:
Noticed I was getting the error:

    Couldn't find template for digesting: store/cart_link

Indeed that template doesn't exist and we need to tell `fresh_when` which template we are actually using.


mamhoff:
👍

PR #1282, opened 2016-06-28
Allow editing multiple Stores

### LINKED ISSUE #1189: Allow editing all Stores through the admin

jhawthorn:
This replaces the existing "General settings" page with an interface allowing editing all orders.

Depends on #1275

Fixes #1189


cbrunsdon:
:+1:

PR #1284, opened 2016-06-29
Fix capybara-screenshot on CI

jhawthorn:
capybara-screenshot stopped working on CI. I think previously it was using `Capybara.save_and_open_page_path`, which has now been deprecated and replaced with `Capybara.save_path`


gmacdougall:
:+1:

PR #1289, opened 2016-06-30
Payment Factory should associate source with User

stewart:
In the process of creating a Payment via the factories, a new CreditCard
record is created, but not associated with any user.

If we create a Payment via one of the Order factories, this can be quite
confusing, as the created Order may have a payment, but the created
Credit Card associated with the payment is not associated with the User
that created the order.

To correct this, we can provide the order's associated user to the
credit card we create as the payment source.

### DIFF
--- a/core/lib/spree/testing_support/factories/payment_factory.rb
+++ b/core/lib/spree/testing_support/factories/payment_factory.rb
@@ -6,7 +6,7 @@ require 'spree/testing_support/factories/store_credit_factory'
 FactoryGirl.define do
   factory :payment, aliases: [:credit_card_payment], class: Spree::Payment do
     association(:payment_method, factory: :credit_card_payment_method)
-    association(:source, factory: :credit_card)
+    source { create(:credit_card, user: order.user) }
     order
     state 'checkout'
     response_code '12345'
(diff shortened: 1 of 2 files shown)


gmacdougall:
:+1: Thanks Andrew

core/lib/spree/testing_support/factories/payment_factory.rb
```diff
@@ -6,7 +6,7 @@
 FactoryGirl.define do
   factory :payment, aliases: [:credit_card_payment], class: Spree::Payment do
     association(:payment_method, factory: :credit_card_payment_method)
-    association(:source, factory: :credit_card)
+    source { create(:credit_card, user: order.user) }
```
mamhoff: I *think* this can be done with the `association` method, which will hit the database less when using `build_stubbed`: 

```
association :source, factory: [:credit_card, user: order.user]
```

stewart: I don't believe that will work, as `order` will be evaluated at the time of factory definition, as opposed to when a factory instance is created.

    [...]/payment_factory.rb:9:in `block (2 levels) in <top (required)>': undefined method `user' for #<FactoryGirl::Declaration::Implicit:0x007f0bfda556d0> (NoMethodError)

jhawthorn:
:+1:

PR #1292, opened 2016-07-01
Fix cancel order with store credits

Sinetheta:
When cancelling an order that had captured store credits the refunding of those credits errors and the order cancellation fails.

This is because methods which should have been invoked on the `PaymentMethod::StoreCredit` were instead being called on the `StoreCredit` directly. This led to bare responses being returned and caused the payment processing to choke [here](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment/processing.rb#L181 ) where it expects an `ActiveMerchant::Billing::Response`.

I have added a spec for `Order#cancel` and improved the specs for `PaymentMethod::StoreCredit#cancel` so that they confirm the desired outcome of refunded store credits.

### DIFF
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -40,7 +40,7 @@ describe Spree::Order, type: :model do
     end
   end
 
-  describe "#cancel!", pending: true do
+  describe "#cancel!" do
     context "with captured store credit" do
       let!(:store_credit_payment_method) { create(:store_credit_payment_method) }
       let(:order_total) { 500.00 }
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -40,6 +40,28 @@ describe Spree::Order, type: :model do
     end
   end
 
+  describe "#cancel!", pending: true do
+    context "with captured store credit" do
+      let!(:store_credit_payment_method) { create(:store_credit_payment_method) }
+      let(:order_total) { 500.00 }
+      let(:store_credit) { create(:store_credit, amount: order_total) }
+      let(:order) { create(:order_with_line_items, user: store_credit.user, line_items_price: order_total) }
+
+      before do
+        order.add_store_credit_payments
+        order.finalize!
+        order.capture_payments!
+      end
+
+      subject { order.cancel! }
+
+      it "cancels the order" do
+        expect{ subject }.to change{ order.can_cancel? }.from(true).to(false)
+        expect(order).to be_canceled
+      end
+    end
+  end
+
   context "#canceled_by" do
     let(:admin_user) { create :admin_user }
     let(:order) { create :order }
--- a/core/spec/models/spree/payment_method/store_credit_spec.rb
+++ b/core/spec/models/spree/payment_method/store_credit_spec.rb
@@ -251,8 +251,9 @@ describe Spree::PaymentMethod::StoreCredit do
       Spree::PaymentMethod::StoreCredit.new.cancel(auth_code)
     end
 
-    let(:store_credit) { create(:store_credit, amount_used: captured_amount) }
+    let(:store_credit) { create(:store_credit, amount: original_amount, amount_used: captured_amount) }
     let(:auth_code)    { "1-SC-20141111111111" }
+    let(:original_amount) { 100.0 }
     let(:captured_amount) { 10.0 }
 
     shared_examples "a spree payment method" do
@@ -269,9 +270,10 @@ describe Spree::PaymentMethod::StoreCredit do
                                         store_credit: store_credit)
       }
 
-      it "creates a store credit for the same amount that was captured" do
-        expect_any_instance_of(Spree::StoreCredit).to receive(:credit).with(captured_amount, auth_code, store_credit.currency)
-        subject
+      it "refunds the capture amount" do
+        expect { subject }.to change{ store_credit.reload.amount_remaining }.
+                              from(original_amount - captured_amount).
+                              to(original_amount)
       end
     end
 
@@ -284,9 +286,10 @@ describe Spree::PaymentMethod::StoreCredit do
                                           store_credit: store_credit)
         }
 
-        it "creates a store credit for the same amount that was captured" do
-          expect_any_instance_of(Spree::StoreCredit).to receive(:void).with(auth_code)
-          subject
+        it "refunds the capture amount" do
+          expect { subject }.to change{ store_credit.reload.amount_remaining }.
+                                from(original_amount - captured_amount).
+                                to(original_amount)
         end
       end
 
--- a/core/spec/models/spree/payment_method/store_credit_spec.rb
+++ b/core/spec/models/spree/payment_method/store_credit_spec.rb
@@ -255,6 +255,12 @@ describe Spree::PaymentMethod::StoreCredit do
     let(:auth_code)    { "1-SC-20141111111111" }
     let(:captured_amount) { 10.0 }
 
+    shared_examples "a spree payment method" do
+      it "returns an ActiveMerchant::Billing::Response" do
+        expect(subject).to be_instance_of(ActiveMerchant::Billing::Response)
+      end
+    end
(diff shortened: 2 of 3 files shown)


core/spec/models/spree/order_spec.rb
```diff
@@ -40,6 +40,28 @@ def compute(_computable)
     end
   end
 
+  describe "#cancel!", pending: true do
+    context "with captured store credit" do
+      let!(:store_credit_payment_method) { create(:store_credit_payment_method) }
+      let(:order_total) { 500.00 }
+      let(:store_credit) { create(:store_credit, amount: order_total) }
+      let(:order) { create(:order_with_totals, user: store_credit.user, line_items_price: order_total).tap(&:update!) }
```
bbuchalter: Why is `tap(&:update!)` required here?

core/spec/models/spree/payment_method/store_credit_spec.rb
```diff
@@ -269,9 +270,10 @@
                                         store_credit: store_credit)
       }
 
-      it "creates a store credit for the same amount that was captured" do
-        expect_any_instance_of(Spree::StoreCredit).to receive(:credit).with(captured_amount, auth_code, store_credit.currency)
-        subject
+      it "refunds the capture amount" do
+        expect { subject }.to change{ store_credit.reload.amount_remaining }.
```
bbuchalter: Although, there is no guidance in StemboltHQ/ruby-style-guide for these kinds of RSpec tests. At minimum, I would expect the spacing between `expect {` and `change{` to be consistent.

bbuchalter:
@Sinetheta this is fantastic work. Thank you so much for putting this together. I have questions about your commit messages:

> In the case of #void, the desired behaviour was actually taking place.
But stubbing the receiver doesn't exactly inspire much faith. We're
about to fix two bugs in #credit which were hidden by this type of
testing.

You suggest we're about to fix two bugs in `credit`, but I never see a regression test for these bugs. We get the spec on `order.cancel`, but I think specs that isolate the bug fixes in `credit` would be valuable too.

> This is a payment method, if it returns a boolean then
spree/payment/processing.rb will blow up. We are seeing this happen when
captured store credits are refunded and the handle_void_response method
tries to check the response.success? Although I have not added a test for the last code path, I think we can refactor it away when we're done with our changes.

Maybe I didn't follow the changes closely enough, but I didn't see clearly refactoring which made the need for testing this regression go away.

bbuchalter:
@jhawthorn I think this is a fairly serious issue that merits back-porting. Does Solidus have a written policy on version support or backporting?

core/spec/models/spree/order_spec.rb
Sinetheta: Just something I grabbed from another spec. I'll find out if this is required (and should be worked back into the factory) or not (and should be removed from other specs).

https://github.com/solidusio/solidus/blob/5f4b2a818829d9ef1e616df544be53bfd6abe540/core/spec/models/spree/order_spec.rb#L1121

Sinetheta:
Thanks @bbuchalter, regarding "the bugs in #credit" I think maybe I didn't phrase that very well. What I meant was that `StoreCredit#void` and `PaymentMethod::StoreCredit#void` have the same signature, but one returns a boolean and the other returns that boolean wrapped in an `ActiveMerchant::Billing::Response`. In the case of `#cancel` we needed to 

1. [Make gateway_options optional](https://github.com/solidusio/solidus/pull/1292/commits/d35fe39ec1704a4d1031ccac06d371b56bf9dd43)
2. [Send it an amount in cents](https://github.com/solidusio/solidus/pull/1292/commits/e221422b3d6f3fdbba6bf0c4828a8d0dcc97c291)

These aren't bugs, so much as differences that weren't apparent before.

My idea of a regression test was to verify that the correct amount of store credits are being returned to the user by both `#void` and  `#cancel`, but that doesn't really distinguish them. That's because I'm not super clear on the difference between how `#void` and `#cancel` are supposed to work for payment methods.

I think that they represent "undoing the authorization or capture", respectively. My understanding is that an `AUTHORIZE_ACTION` is a sort of "reserving of funds" which can be undone by cancelling, and that a `CAPTURE_ACTION` means the funds have actually been spent, so we need to refund the same amount. But I'm not super solid on it, so I just added a high level test.

I figured it was an improvement, because the previous spec simply verified that calls were made to the incorrect methods. But I would be happy to expand on them and would also like to clean up `#cancel` because I have a feeling that not all 4 code paths can be reached (both 1 and 4 mean we couldn't find a store credit event).

jasonfb:
@Sinetheta 
Keep in mind the refunds exist as their own first-class entity in Solidus I think, so that if "cancel" on an existing payment means refund the capture (which doesn't seem transactionally correct to me), then there should be a user interface that allows the operator to specify the type or reason for the refund. 

We use 'void' to mean "void out this payment record and no longer talk to the external gateway about it" and cancel to mean "cancel this payment and tell the external gateway about the cancellation"

In my view, cancelling should only happen before capture. After capture, you should go through the existing refund workflow instead and leave the original payment records as-is (in 'completed' state) since that is more appropriate from a book-keeping perspective.

bbuchalter:
> I figured it was an improvement, because the previous spec simply verified that calls were made to the incorrect methods. 

@Sinetheta Agreed. This is a big improvement.

> But I would be happy to expand on them and would also like to clean up #cancel because I have a feeling that not all 4 code paths can be reached (both 1 and 4 mean we couldn't find a 
store credit event).

I don't think this PR is the place to go deep diving into these interfaces. I think we've got a real improvement here as is.

core/spec/models/spree/order_spec.rb
jhawthorn: If it is required (I don't think it should be) it should be in the before block

Senjai:
I'm :+1: on this. Good work @Sinetheta

Sinetheta:
Thanks @jasonfb I agree that it's a little wacky that there's a button up top that just forces through a bunch of voids and refunds until everything is dead, but right now that's how it works. This is just a fix for orders with store credit.

core/spec/models/spree/order_spec.rb
```diff
@@ -40,6 +40,28 @@ def compute(_computable)
...
+        order.add_store_credit_payments
+        order.finalize!
+        order.capture_payments!
+      end
+
+      subject { order.cancel! }
+
+      it "cancels the order" do
+        expect{ subject }.to change{ order.can_cancel? }.from(true).to(false)
+        expect(order.canceled?).to be(true)
```
jhawthorn: This should be `expect(order).to be_canceled`

core/spec/models/spree/order_spec.rb
```diff
@@ -40,6 +40,28 @@ def compute(_computable)
     end
   end
 
+  describe "#cancel!", pending: true do
+    context "with captured store credit" do
+      let!(:store_credit_payment_method) { create(:store_credit_payment_method) }
+      let(:order_total) { 500.00 }
+      let(:store_credit) { create(:store_credit, amount: order_total) }
+      let(:order) { create(:order_with_totals, user: store_credit.user, line_items_price: order_total) }
```
jhawthorn: This factory is crap (sorry). The `order_with_totals` factory actually makes an order with no totals :disappointed_relieved: :rage4: 

As a result this spec wasn't actually failing after you removed the `.update!` :cry:.

Changing this to `create(:order_with_line_items` should fix the problem.

jhawthorn:
Other than two comments this is excellent. Thank you Kevin.

jhawthorn:
@bbuchalter Our [Versioning Guidelines](https://github.com/solidusio/solidus/wiki/Versioning-Guidelines) are in the wiki. Even though this changes some obviously broken behaviour it is too big of a change to go in a patch release. It will be a part of the upcoming 1.4.0 release.

jhawthorn:
:+1:

Senjai:
Still :+1: :tada:

gmacdougall:
👍 

Thanks Kevin!

Issue #1293, opened 2016-07-04
Can't view stock transfer if variant has been soft-deleted


isaacfreeman:
Visiting `/spree/admin/stock_transfers/transfer_number` results in a crash if one of the TransferItems refers to a variant that has since been deleted.

Obviously we don't want create new stock transfers for deleted variants, but I think we should be able to look at old stock transfers even if they refer to variants that are no longer available.

If there are no side-effects I'm unaware of, we should be able to fix this in https://github.com/solidusio/solidus/blob/master/core/app/models/spree/transfer_item.rb by changing

```
belongs_to :variant
```

to

```
belongs_to :variant, -> { with_deleted }
```

jhawthorn:
This was fixed by #1294

PR #1294, opened 2016-07-04
Allow TransferItem to retrieve a soft-deleted variant

isaacfreeman:
Addresses #1293

Specs passing, but it might worth someone with a deeper knowledge of the stock code checking for unintended side effects.


gmacdougall:
This looks quite reasonable. Could you please add a spec to show that this is functioning with deleted variants now? That way we don't accidentally break it in the future.

mamhoff:
👍

jasonfb:
conceptually it makes sense to me, and I agree screen should never crash on the Admin users, although I still do have a categorial objection to soft-deleting. (but, as has been explained to me before, until soft-deleting is removed this is the most effective strategy to delete these bugs)

My only suggestion here is a user experience one: Particularly, Merchandiser teams that operate Spree/Solidus by deleting old variants and replacing them with _identical SKUs_ (new records) -- which is possible because the SKU-uniquness is not enforced across the deleted/non-deleted records --- would have a very difficult them figuring out which of two identical SKUs on that screen was deleted and which was not deleted. 

I don't think that should block the bugfix, but it seems like a thorough implementation would change the UI to show the deleted ones with an indicator that they have been deleted (like strikethrough text or something). As well, it seems to me that one should be forcibly stopped from adding/removing stock for deleted variants (which I think actually works as-is because on that screen you input a SKU anyway). 

I guess all I'm getting at here is that by showing the deleted records there you create some confusion for the Merchandising team if there were two SKUs with the same value on the page and you couldn't tell which was which.

isaacfreeman:
@jasonfb: Makes sense to me. While this was intended to be a minor patch, and I don't have a philosophical objection to soft-deleting, I do think the admin UI should be very clear about the state of the system. It makes sense that people should know when they're looking at a variant that's since been deleted.

How about I do the following?
1. Show soft-deleted variants on a TransferItem in red text, as for deleted products in a search where Show Deleted has been checked.
2. Add an additional brief text to clarify the situation: either "This variant has since been deleted" or "This variant has since been replaced by another with the same SKU" depending on the situation.
3. Get @Mandily's input on the above as a short-term UX solution.
4. Ask @Mandily to consider the wider question of how to portray deleted entities across Solidus, and possibly cover this in the style guide.

Mandily:
We've been using red as a colour for the action that will delete a record in the admin. It's a helpful secondary indicator (the first being the delete **icon** or word **delete**) but I wouldn't want to establish it as the first and only indicator because that leaves the system not as usable for the colour-blind folk. 

I'd also like to establish a different visual treatment for **things that are already deleted** compared to **things that will perform a delete action**. I like the idea of a strikethrough, possibly even a lighter shade of the text. 

Some other ideas:
I'm not sure what the space considerations are, but this may be a good opportunity to use the "i" icon with tooltip that @jhawthorn used in #1266. Thinking a little ahead, if a user had three sku's, two of which had been deleted, it would be handy to have a 'deleted date' shown too. But that may be a little out of scope for what you're doing here. 

What do you think?

isaacfreeman:
@Mandily: I don't mind extending the scope – UX is always relevant. I would like though, to keep this issue scoped to just the TransferItem page, with the understanding that a more holistic review of how deleted items are rendered across all admin pages would be a separate issue.

So revised plan for deleted variants on this one page:
1. Lighter shade of text and strikethrough
2. Tooltip icon with first line of text saying "Variant deleted on [Date]"
3. Where relevant, an additional line of text saying "This variant has since been replaced by another with the same SKU"

Mandily:
That sounds great 👍

isaacfreeman:
It seems there's already a CSS class for deleted items in a table, so I'll stick with that.

Currently it looks like this:
screen shot 2016-07-08 at 17 13 06

And hovered:
screen shot 2016-07-08 at 17 13 12

Still some code to tidy up, but let me know if anyone wants changes.

Mandily:
I'm fine with tackling any styling issues in the brand overhaul. This looks good to me 👍

isaacfreeman:
OK, this is up to date with comments so far.

jhawthorn:
Thank you @isaacfreeman

Issue #1291, opened 2016-07-01
How to customize views HTML within solidus


acrolink:
I have noticed that the app/views & app/controllers folders are empty. My question, how to customize HTML of the rendered pages? Which files I need to target? Thanks.

jasonfb:
this information from the old Spree guides applies to your questions: http://guides.spreecommerce.org/developer/view.html#template-replacements

You might want to seek help on the #support channel of the Solidus Slack room.

acrolink:
Thanks @jasonfb .. I read the documentation you pointed to .. I think this task should be easier to do than the way it is now, nothing compares to directly editing views and controllers .. What I understood is that modifying the views/controllers files is possible but that would mean editing the global ones. Maybe I can avoid this by installing solidus/spree locally within my app.

jasonfb:
Spam detection software, running on the system "mail2.hsphere.cc", has
identified this incoming email as possible spam.  The original message
has been attached to this so you can view it (if it isn't spam) or label
similar future email.  If you have any questions, see
the administrator of that system for details.

Content preview:  Actually for views you copy the file from the gem views into
   your app (thus creating work for yourself when you upgrade and the original
   file changes --- this is why some people use deface, but deface has its own
   problems) [...] 

Content analysis details:   (5.1 points, 5.0 required)

 pts rule name              description
---- ---------------------- --------------------------------------------------
 2.5 HTML_IMAGE_ONLY_28     BODY: HTML: images with 2400-2800 bytes of words
 0.0 HTML_MESSAGE           BODY: HTML included in message
 2.5 MIME_QP_LONG_LINE      RAW: Quoted-printable line longer than 76 chars
 0.1 RDNS_NONE              Delivered to trusted network by a host with no rDNS

The original message was not completely plain text, and may be unsafe to
open with some email clients; in particular, it may contain a virus,
or confirm that your address can receive spam.  If you wish to view
it, it may be safer to save it to a file and open it with an editor.


X-Virus-Scan: Scanned by ClamAV 0.98.4 (no viruses);
  Sat, 02 Jul 2016 11:06:32 -0400
Received: from unknown (HELO ?100.121.54.16?) (jason.fb@datatravels.com@70.199.96.172)
  (de-)crypted with TLSv1: DHE-RSA-AES256-SHA [256/256] DN=unknown
  by mail2.hsphere.cc with ESMTPSAM; 2 Jul 2016 15:06:31 -0000
X-Originating-IP: 70.199.96.172
[...]

PR #1296, opened 2016-07-05
Move admin menu definition to configuration

gmacdougall:
This moves the admin menu to a configurable class. This allows people to
customize the menu by modifying it to their own will.

This can be accomplished by adding configuration to
Spree::Backend::Config.configure upon initialization of your app.

Example use:

```Ruby
  config.menu_items << Spree::BackendConfiguration::MenuItem.new(
    [:section],
    'icon-name',
    url: 'https://solidus.io/'
  )
end
```

### DIFF
--- a/backend/app/models/spree/backend_configuration.rb
+++ b/backend/app/models/spree/backend_configuration.rb
@@ -17,5 +17,105 @@ module Spree
     PROMOTION_TABS     ||= [:promotions, :promotion_categories]
     STOCK_TABS         ||= [:stock_items, :stock_transfers]
     USER_TABS          ||= [:users, :store_credits]
+
+    # An item which should be drawn in the admin menu
+    class MenuItem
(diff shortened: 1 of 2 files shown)


tvdeyen:
Just 👍

Excellent work.

mamhoff:
Wow! This is amazing stuff. 👍

jasonfb:
Love it. So much cleaner than deface. We should figure out how to update the "how to customize solidus" documentation out there to include references to this + other new patterns for customization, at least as "here's some new patterns we're trying out... try them yourself and tell us what you think" kind of thing. 

See recent GH conversation I was having on https://github.com/solidusio/solidus/issues/1291

jhawthorn:
:+1: This is great.

I'd love to eventually see the 3rd level (tab) menus built in a similar way, but that's definitely a task for a future PR.

PR #1297, opened 2016-07-05
Use spree RouteSet in navigation helper

gmacdougall:
This does not work correctly with admin pages outside of the Spree
namespace because the spree RouteSet is not explicitly invoked. The
spree RouteSet is used above in this class, but was not used correctly
here.


tvdeyen:
👍

jhawthorn:
:+1:

PR #1298, opened 2016-07-05
Properly namespace route in navigation helper

graygilmore:
Ran into this issue when working on a customization in the backend that didn't live under the `Spree` namespace.

### DIFF
--- a/backend/spec/helpers/admin/navigation_helper_spec.rb
+++ b/backend/spec/helpers/admin/navigation_helper_spec.rb
@@ -37,8 +37,7 @@ describe Spree::Admin::NavigationHelper, type: :helper do
 
       context "when match_path option is supplied" do
         before do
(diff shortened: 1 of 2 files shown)


backend/spec/helpers/admin/navigation_helper_spec.rb
```diff
@@ -37,8 +37,7 @@
 
       context "when match_path option is supplied" do
         before do
-          allow(helper).to receive(:admin_path).and_return("/somepath")
```
graygilmore: This didn't appear to provide any value to me and was causing issues in the test. If somebody can tell me why this is important to have I'll figure out a way to add it back in.

graygilmore:
And of course @gmacdougall beat me to it by 25min 😄 https://github.com/solidusio/solidus/pull/1297

PR #1301, opened 2016-07-06
Order details partial shows duplicate adjustments

SebAshton:
Usage of all_adjustments was showing duplicate promotions.


mamhoff:
Can you provide a spec for this?

mamhoff:
This is good work. It needs a spec. A view spec would be enough IMO. In order to test this behaviour, you'll need an order with both a line item level and an order level adjustment. @SebAshton Would you be willing to do this, or should we close this and make it an issue?

gmacdougall:
This was merged in #2257 after a rebase. Thanks for the work on this Seb!  Sorry for the very slow response on this.

PR #1306, opened 2016-07-06
Use with_model for DbMaximumLengthValidator

jhawthorn:
In rails 5 redefining the attribute doesn't change the columns_hash (which actually makes sense).

We shouldn't be redefining the slug on `Spree::Product` anyways, our tests should be run in isolation, and we want to test this against an actual table in the database.


tvdeyen:
👍

gmacdougall:
👍

PR #1308, opened 2016-07-06
Fix issues related to params being received as strings

jhawthorn:
In rails 5, controller tests treat `params` more correctly:

* values are converted to a string when posting form data.
* empty hash is not sent (can't be encoded)

Since these aren't realistic scenarios, we shouldn't be testing them.


tvdeyen:
Indeed 👍

PR #1309, opened 2016-07-06
Improved API pagination

jhawthorn:
## current_page

Previously, when displaying the current page in an API response, we used `params[:page]` when displaying the current page.

In many places we were using `params[:page] || 1`, which will display the page as a string (`"current_page":"2"` for `?page=2`). In other places we used `params[:page] ? params[:page].to_i : 1`, which usually displayed the page correctly as an integer, but would display `"current_page": 0` for non-integer input (`?page=foo`), even though the page was actually `1`.

Kaminari adds a `current_page` method to the scope, so we should just use that.

## per_page

We only displayed the `per_page` value in two places using `params[:per_page] || Kaminari.config.default_per_page`. This had the same problem of displaying a string instead of an integer. I've replaced this with `.limit_value`

### DIFF
--- a/api/app/views/spree/api/countries/index.v1.rabl
+++ b/api/app/views/spree/api/countries/index.v1.rabl
@@ -3,5 +3,5 @@ child(@countries => :countries) do
   attributes *country_attributes
 end
 node(:count) { @countries.count }
-node(:current_page) { params[:page] || 1 }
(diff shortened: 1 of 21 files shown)


tvdeyen:
Great. The specs are even better readable know. 👍

gmacdougall:
👍 

Much better. Thanks John!

PR #1313, opened 2016-07-08
Add by_store scope to #set_current_order

SebAshton:
On a multi domain site we've had carts in other stores being merged together for logged in users. Adding the by_store scope fixed this.

Happy to wrap in a regression test.


mamhoff:
I think this needs some test fixing, and maybe even a new spec. Otherwise, this is great!

dholdren:
a version of this PR with specs: https://github.com/solidusio/solidus/pull/1386

jhawthorn:
Fixed by #1386

Thank you both

PR #1314, opened 2016-07-08
Hide clear selection if no clear option.

eric1234:
See commit message on f5b07f0 for rational. After applying this commit selects work [as specified in the select2 documentation](http://select2.github.io/select2/#placeholders). See:

With and without clear

Note how the first one can be cleared but the second one cannot. The second one does not have a blank option while the first does. I also verified this didn't mess with multi-selects:

Multi-select


jhawthorn:
Thanks again. I pulled this down and it does indeed remove the inappropriate clear buttons. Much better.

This is another case (of many in our codebase) of `@extend .fa` wreaking havoc on the CSS.

:+1:

gmacdougall:
Thanks for the contribution Eric!

👍

Issue #1319, opened 2016-07-11
Add credential details to Sandbox section of README


bbuchalter:
The Sandbox instructions work, but provide no details on how to access the admin. These details are currently buried in https://github.com/stewart/solidus_auth_devise/blob/master/db/default/users.rb which would be extremely difficult for new developers to find.

jhawthorn:
Done in #1320

PR #1320, opened 2016-07-11
Add note of default admin username/password

stewart:
These defaults are hidden in a strange place in solidus-auth-devise, and
it's possible to miss the prompt during setup, or otherwise forget these
default credentials. So, let's make it a bit easier to get started, by
making it more clear how to login to a new store.

Somewhat, but not completely, corrects #1319.


isaacfreeman:
I frequently type in "spree123", then have to go and look up what it is now. This would be a help for people like me whose muscle memory is better than their regular memory.

jhawthorn:
:+!:

tvdeyen:
Thanks. Would love to see some kind of header above. I didn't spot this in the readme at first glance.

PR #1324, opened 2016-07-12
Simplify handling AC::Params in PaymentCreate

jhawthorn:
Previously, if PaymentCreate was passed attributes as `ActionController::Parameters` it would eventually pass them as `ActionController::Parameters` to `ActiveRecord`. This worked and was safe, but difficult to reason about.

This PR instead immediately converts the attributes to a hash of just the sanitized attributes. This is easier to reason about and verify its safety.

This change is needed for rails5 because the behaviour of `ActionController::Parameters#with_indifferent_access` has changed now that the class is no longer a subclass of `HashWithIndifferentAccess`.

### DIFF
--- a/core/app/models/spree/payment_create.rb
+++ b/core/app/models/spree/payment_create.rb
@@ -2,7 +2,7 @@ module Spree
   # Service object for creating new payments on an Order
   class PaymentCreate
     # @param order [Order] The order for the new payment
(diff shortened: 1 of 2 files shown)


mamhoff:
👍

stewart:
:+1:

PR #1327, opened 2016-07-13
Place Carton class in Spree module.

bbuchalter:
For consistent code style and class-lookup behavior, classes should be
defined inside a module explicitly.

### DIFF
--- a/core/app/models/spree/carton.rb
+++ b/core/app/models/spree/carton.rb
@@ -1,57 +1,59 @@
-class Spree::Carton < Spree::Base
-  belongs_to :address, class_name: 'Spree::Address'
-  belongs_to :stock_location, class_name: 'Spree::StockLocation', inverse_of: :cartons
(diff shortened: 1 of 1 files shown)


jordan-brough:
I prefer the non-nested style but I don't know if we need to mandate it one way or another. @bbuchalter do you know of anything that might break if we do this one way or the other?

bbuchalter:
@jordan-brough sadly, I don't have a way to demonstrate that the style of declaration will impact behavior. However, I do have a _memory_ that consistently is important for Ruby and Rails' code loading conventions. I'd understand if we want to close this.

jordan-brough:
@bbuchalter I think I have the same sort of haunting memory about Rails autoloading and class declarations. :)  I can't think of what exactly might break though either.  How about we close for now and see if we get someone reporting a failure?

bbuchalter:
Fine by me.

PR #1329, opened 2016-07-13
Refactor shipping_specs on admin product form

dfranciosi:
Works in the same way as in admin variants from
(https://github.com/solidusio/solidus/blob/60ee1dffc7c3932cd7652cc2b4566f08a12f74b8/backend/app/views/spree/admin/variants/_form.html.erb#L41)
Fixes a bug coming from Spree on weight on different locales
(https://github.com/spree/spree/issues/6480)

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -104,22 +104,12 @@
       </div>
 
       <div id="shipping_specs">
-        <div id="shipping_specs_height_field" data-hook="admin_product_form_height" class="field alpha two columns">
-          <%= f.label :height %>
-          <%= f.text_field :height, :size => 4 %>
-        </div>
-        <div id="shipping_specs_width_field" data-hook="admin_product_form_width" class="field omega two columns">
-          <%= f.label :width %>
-          <%= f.text_field :width, :size => 4 %>
-        </div>
-        <div id="shipping_specs_depth_field" data-hook="admin_product_form_depth" class="field alpha two columns">
-          <%= f.label :depth %>
-          <%= f.text_field :depth, :size => 4 %>
-        </div>
-        <div id="shipping_specs_weight_field" data-hook="admin_product_form_weight" class="field omega two columns">
-          <%= f.label :weight %>
-          <%= f.text_field :weight, :size => 4 %>
-        </div>
+        <% [:height, :width, :depth, :weight].each_with_index do |field, index| %>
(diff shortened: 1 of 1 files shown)


cbrunsdon:
:+1:, thanks.

For the record I spent an unfortunate amount of time trying to figure out what "specs" this would fix. :coffee:

mamhoff:
👍

backend/app/views/spree/admin/products/_form.html.erb
```diff
@@ -104,22 +104,12 @@
...
-        <div id="shipping_specs_depth_field" data-hook="admin_product_form_depth" class="field alpha two columns">
-          <%= f.label :depth %>
-          <%= f.text_field :depth, :size => 4 %>
-        </div>
-        <div id="shipping_specs_weight_field" data-hook="admin_product_form_weight" class="field omega two columns">
-          <%= f.label :weight %>
-          <%= f.text_field :weight, :size => 4 %>
-        </div>
+        <% [:height, :width, :depth, :weight].each_with_index do |field, index| %>
+          <div id="shipping_specs_<%= field %>_field" data-hook="admin_product_form_<%= field %>" class="field two columns <%= index.even? ? 'alpha' : 'omega' %>">
```
gmacdougall: I think we can lose the data-hook here. It was designed to be matched by deface overrides, but that won't be useful in this circumstance since deface operates on the template itself.

dfranciosi: @gmacdougall good catch, anyway we have tons of `data-hook` in this partial and around the codebase too, do you think it's purpose of this PR to remove this one in particular?

gmacdougall: As part of refactoring this, I feel this one should be removed as it can't be used for its intended purpose any longer.

dfranciosi: @gmacdougall Ok, perfectly fine with me, I'm going to add a commit.

gmacdougall:
Could you please remove my mention from the commit message. With the `@` it will message me any time someone cherry-picks that commit around to various branches.  Non `@` mention is fine though.  😄

dfranciosi:
@gmacdougall done 😉

cbrunsdon:
@dfranciosi would probably be good to strip @alepore from the commit message too, if you get a chance, thanks

dfranciosi:
@gmacdougall @cbrunsdon nobody is mentioned now 👍

jhawthorn:
:+1: Thank you

PR #1330, opened 2016-07-13
Use normal get/post/put/delete spec helpers instead of spree_*

jhawthorn:
The spree_get and similar helpers no longer do anything. The `ControllerRequests` module now [just sets the `routes`](https://github.com/solidusio/solidus/blob/master/core/lib/spree/testing_support/controller_requests.rb#L45-L47) and the `spree_*` helpers don't do anything other than call the standard helpers.

### DIFF
--- a/backend/spec/controllers/spree/admin/cancellations_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/cancellations_controller_spec.rb
@@ -4,7 +4,7 @@ describe Spree::Admin::CancellationsController do
   stub_authorization!
 
   describe "#index" do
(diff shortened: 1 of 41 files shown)


gmacdougall:
Yay! I always disliked those.

Could you also flag the existing `spree_*` helpers as deprecated?

tvdeyen:
:+1: and deprecation is great.

mamhoff:
👍

cbrunsdon:
Fantastic, big :+1:  from me too.

mtomov:
wow, brilliant, thanks a lot!

PR #1331, opened 2016-07-13
Add the ability to get simple results from taxon API

gmacdougall:
The taxon products API which is used to manage display order falls down
under its own weight in any sort of reasonable data set. In order to
simplify that, I have added the ability to customize the return data
from this controller.

In a simple case, the results will be the same as they are right now,
however you can pass `simple: 1` to the taxon/products API request and
it will return significantly less data.

Additionally, this updates the backend taxons.js.coffee to use this
simple view for all it needs. This reduces the amount of data that needs
to be queried and returned in order to make this fast. Previously I was
seeing 20+ second response times for a small number of products in a
complicated taxon structure. This reduces that down to the 100ms range.

### DIFF
--- a/api/app/controllers/spree/api/taxons_controller.rb
+++ b/api/app/controllers/spree/api/taxons_controller.rb
@@ -66,6 +66,17 @@ module Spree
         # Products#index does not do the sorting.
         taxon = Spree::Taxon.find(params[:id])
         @products = paginate(taxon.products.ransack(params[:q]).result)
+        @products = @products.includes(master: :default_price)
+
+        if params[:simple]
+          @exclude_data = {
+            variants: true,
+            option_types: true,
+            product_properties: true,
+            classifications: true
+          }
+          @product_attributes = %i(id name display_price)
+        end
         render "spree/api/products/index"
       end
 
--- a/api/app/views/spree/api/products/index.v1.rabl
+++ b/api/app/views/spree/api/products/index.v1.rabl
@@ -1,5 +1,5 @@
 object false
-node(:count) { @products.count }
+node(:count) { @products.size }
 node(:total_count) { @products.total_count }
 node(:current_page) { @products.current_page }
 node(:per_page) { @products.limit_value }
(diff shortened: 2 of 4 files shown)


gmacdougall:
Sample response after from the solidus sandbox:

```
> http "http://localhost:3000/api/taxons/products?id=7&simple=1" X-Spree-Token:cd0ebcf6ac7d236c7007f23ad05b50bba470c9812c57d8ab
HTTP/1.1 200 OK
Cache-Control: max-age=0, private, must-revalidate
Connection: Keep-Alive
Content-Length: 357
Content-Type: application/json; charset=utf-8
Date: Wed, 13 Jul 2016 19:25:29 GMT
Etag: W/"38d78a06563db8d140330f4323f18540"
Server: WEBrick/1.3.1 (Ruby/2.1.5/2014-11-13)
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Request-Id: c85d935d-39ad-4baa-a9c8-32575e8d1322
X-Runtime: 0.075606
X-Xss-Protection: 1; mode=block
```
```Json
{
    "count": 4,
    "current_page": 1,
    "pages": 1,
    "per_page": 500,
    "products": [
        {
            "display_price": "$19.99",
            "id": 7,
            "name": "Apache Baseball Jersey"
        },
        {
            "display_price": "$19.99",
            "id": 6,
            "name": "Ruby Baseball Jersey"
        },
        {
            "display_price": "$19.99",
            "id": 3,
            "name": "Ruby on Rails Baseball Jersey"
        },
        {
            "display_price": "$19.99",
            "id": 5,
            "name": "Ruby on Rails Ringer T-Shirt"
        }
    ],
    "total_count": 4
}
```

jhawthorn:
:+1:

I don't love the special case extra parameter of `?simple=1`, but I don't think we're ready for the generic solution of `?excludes=variants,option_types`. Our next steps for the API are uncertain, so I do agree with the minimal interface provided here.

jhawthorn:
@gmacdougall I think this is good to go. Could you give it a rebase and merge it?

gmacdougall:
Rebase incoming.

PR #1332, opened 2016-07-13
Control raising of deprecations through ENV var

jhawthorn:
In #1255 we started raising on any depredations in the test suite.

I think this has been great in core, but in extensions, which also use the test_app, I feel this has been a little too aggressive.

This allows extensions to opt-in or out of raising on deprecations.


tvdeyen:
Good idea. It was too aggressive. :+1:

mamhoff:
Will CircleCI still raise on deprecations? Because I'd love that, and that was the main intent of the original PR. Otherwise 👍

jhawthorn:
I've already set the environment variable on CircleCI, so it should still be raising. I think we'll want to turn this on for a few extensions as well (ex. solidus_auth_devise).

mamhoff:
Perfect. 👍

PR #1337, opened 2016-07-15
Fixes for Admin Payment with multiple Payment Methods

### LINKED ISSUE #1336: Multiple Payment Methods not working in Admin
With the changes in 1.3 the javascript on `admin_order_payments_path` is no longer firing. It looks like it is because `#new_payment` is gone.

https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/payments/new.html.erb#L13 renders 

```
<form class="edit_payment" id="edit_payment_3126575" action="/admin/orders/:order_id/payments" accept-charset="UTF-8" method="post">
  <input type="hidden" name="_method" value="patch">
...
</form>
```

and the javascript for that page is looking for `#new_payment` https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/payments/new.js#L3

This causes issues if there are more than one payment method set up where the form will submit all fields even for non-selected payment methods.

It's also rendering the `patch` hidden method and as far as I can tell there is no `PATCH` path for admin payments which causes `No route matches [PATCH] "/admin/orders/:order_id/payments"`

ericsaupe:
Fixes #1336 

Changing to `Spree::Payments.new` stops it from building an object without a source and causing errors in validation when a source is required and the user leaves the new page having built a payment without a source.

It also resolves the `PATCH` and javascript problems since we are not editing an already created object and instead are creating a brand new object.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/payments/new.js
+++ b/backend/app/assets/javascripts/spree/backend/payments/new.js
@@ -1,6 +1,6 @@
 //= require jquery.payment
 $(document).ready(function() {
-  if ($("#new_payment").is("*")) {
+  if ($(".edit_payment").is("*")) {
     $(".cardNumber").payment('formatCardNumber');
     $(".cardExpiry").payment('formatCardExpiry');
     $(".cardCode").payment('formatCardCVC');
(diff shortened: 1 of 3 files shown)


bbuchalter:
How do we feel about a feature test to protect this from future regression? Seems like a pretty important behavior, so it'd be worth having under test.

cbrunsdon:
What is it that lets it be valid if we .new it? Is the state 'invalid' so it skips the source check?

gmacdougall:
Based upon my findings in #1336 I am not confident this fixes an issue. I am going to close this PR for now pending new information.

PR #1339, opened 2016-07-18
Refund indexes

Senjai:
There were two indexes that were missing from the refunds table that slow down queries that use them. When the table was created, the migration created them as integer fields instead of using the `references` helper which adds a foreign key column and the index together.


jhawthorn:
:+1: Can confirm that there's only an index on `refund_reason_id` (which provides little value) and not `payment_id` or `reimbursement_id` (which both make sense).

gmacdougall:
:+1:

cbrunsdon:
:+1: , thanks @Senjai

PR #1340, opened 2016-07-18
Fix admin product sub-menu link view permissions for Taxon and Taxonomies

mtomov:
- as Default Customer has permission to view Taxonomy and Taxon links by default
    https://github.com/solidusio/solidus/blob/master/core/lib/spree/permission_sets/default_customer.rb#L24-L25

  - but we do not want them see the admin links if they don't have
    permission to administer them


cbrunsdon:
Makes sense to me, thanks @mtomov  :+1:

jhawthorn:
:+1: Thank you

PR #1341, opened 2016-07-19
Give populate_redirect its own controller action

jhawthorn:
For some reason setting the flash in the inline redirect doesn't work in rails 5.


cbrunsdon:
I'm fine with this, I wasn't too impressed with that in the routes.rb anyway
:+1:

gmacdougall:
👍

mamhoff:
👍

PR #1342, opened 2016-07-19
Rails 5

jhawthorn:
This still isn't quite ready to merge, and many parts I still hope to merge separately, but specs are passing and it should help visibility to have this PR.

### DIFF
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -73,8 +73,7 @@ module Spree
 
         error_notifier.call(exception, self) if error_notifier
 
-        render text: { exception: exception.message }.to_json,
-          status: 422
+        render json: { exception: exception.message }, status: 422
       end
 
       def gateway_error(exception)
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -148,7 +148,7 @@ module Spree
       def lock_order
         OrderMutex.with_lock!(@order) { yield }
       rescue Spree::OrderMutex::LockFailed => e
-        render text: e.message, status: 409
+        render plain: e.message, status: 409
       end
 
       def insufficient_stock_error(exception)
--- a/core/lib/spree/core/importer/product.rb
+++ b/core/lib/spree/core/importer/product.rb
@@ -7,8 +7,8 @@ module Spree
         def initialize(product, product_params, options = {})
           @product = product || Spree::Product.new(product_params)
 
-          @product_attrs = product_params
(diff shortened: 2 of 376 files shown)


jhawthorn:
This is ready to go :shipit:

gmacdougall:
Great Work John!  Big thanks to all the other contributors as well.

👍

core/lib/spree/core/importer/product.rb
```diff
@@ -7,8 +7,8 @@ class Product
         def initialize(product, product_params, options = {})
           @product = product || Spree::Product.new(product_params)
 
-          @product_attrs = product_params
-          @variants_attrs = options[:variants_attrs] || []
+          @product_attrs = product_params.to_h
+          @variants_attrs = (options[:variants_attrs] || []).map(&:to_h)
```
jordan-brough: What about Rails 5 made this necessary?  It feels like this should be done at the controller level to me.

jordan-brough: (Same question about the changes in `user_address_book.rb` and `importer/order.rb`.)

api/app/controllers/spree/api/base_controller.rb
```diff
@@ -73,8 +73,7 @@ def error_during_processing(exception)
 
         error_notifier.call(exception, self) if error_notifier
 
-        render plain: { exception: exception.message }.to_json,
-          status: 422
+        render json: { exception: exception.message }, status: 422
```
jordan-brough: Should we add this (and the others in "Use render :json insteda of render :text for json") to the non-Rails 5 branch also?

jordan-brough:
Looking great!  I left a couple questions, I'm not sure why github thinks they are outdated.

core/lib/spree/core/importer/product.rb
jhawthorn: Lower in this file there is

```
{ product: product }.merge(variant_attribute)
```

Where `variant_attribute` is a hash derived from `variants_attrs`. This causes a warning in rails 5 because it is converting an `ActionController::Parameters` to a `Hash` without sanitizing.

api/app/controllers/spree/api/base_controller.rb
jhawthorn: I think we should

core/lib/spree/core/importer/product.rb
jordan-brough: Hm, I still feel like the `.to_h` is more appropriate at the controller level, but I don't think it's important enough to hold this up.  I'm happy to see how this feels and readdress later if we feel like it.

jordan-brough:
👍

PR #1343, opened 2016-07-19
Avoid alias_method_chain in OptionValue promo rule

jhawthorn:
`alias_method_chain` is deprecated in rails5.


gmacdougall:
This does break API compatibility. Do we care?

jhawthorn:
> This does break API compatibility. Do we care?

I suppose it does remove the `preferred_eligible_values_without_numerification` method, but I don't expect anyone to be using that so IMO this is fine.

bbuchalter:
I know I'm just one voice, but my app isn't using this method.

mamhoff:
:+1: go for it

PR #1344, opened 2016-07-19
Remove calls to silence_stream

jhawthorn:
`Kernel#silence_stream` is deprecated in rails 4.2 (only in the docs, it doesn't issue a warning) and has been removed in rails 5.

`silence_stream` also was hiding any warnings (or errors) from migrations, generators, and seed in our test_app and sandbox.

Instead, this PR tries to make all of the setup steps as quiet as possible, without resorting to output redirection.

### DIFF
--- a/core/lib/generators/spree/install/install_generator.rb
+++ b/core/lib/generators/spree/install/install_generator.rb
@@ -108,24 +108,18 @@ Spree::Auth::Engine.load_seed if defined?(Spree::Auth)
 
     def install_migrations
       say_status :copying, "migrations"
-      silence_stream(STDOUT) do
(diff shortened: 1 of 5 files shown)


cbrunsdon:
These all look right to me, so :+1:.

Disclaimer I haven't ran it locally. Also I don't care about that hound warning.

gmacdougall:
:+1:

If we don't care about the hound warning we should disable it.

jhawthorn:
I like the `%x{` notation, but I actually agree with the hound warning. More people will be familiar with the backtick notation, so we should use that. Fixed.

PR #1349, opened 2016-07-24
Fix user picker search request

jeremyw:
The users API controller expects the search params to be contains in a
`q` node.


jeremyw:
Fixes #1350

jhawthorn:
:+1: I've added a spec in #1372 to test this

jhawthorn:
Thanks @jeremyw

Issue #1352, opened 2016-07-25
Edit promotions page - rules for users


simo163:
Hello,

I found a little issue when loading the edit promotion page, initializing the users into the rules if existing requires the API key access. I send a small PR for it. Kind regards!

jeremyw:
See #1350. The issue is not the API key but the params.

jhawthorn:
Duplicate of #1350

joelind:
Actually, there *is* an API key issue. When you go to edit an existing promotion with a user rule, the box that is supposed to list the users is empty. It appears that the UI is trying to fetch the users by ID, but it gets a 401 (since this API request has no token in it). See attached. I'm working on a PR for this.

<img width="1257" alt="screen shot 2016-08-12 at 8 44 16 am" src="https://cloud.githubusercontent.com/assets/39476/17622564/20695f1a-6069-11e6-87dc-e69b9486e584.png">

jasonfb:
@joelind -- are you certain you generated your own API key for the user you are logged in with? You do this by searching for yourself on the "Users" tab and then hit "Generate API Key" on your own page (This is associated with the Admin user you are logged in with)

Generally all Admins need API keys

joelind:
@jasonfb yes. The `initSelection` option that is passed to `select2` does a raw jQuery get, which doesn't seem to be sending the api_key with it. If I change the following, it works. Am I missing something?

If I edit `user_picker.js`, to change ```
 11     initSelection: function (element, callback) {
 12       $.get(Spree.routes.users_api, {
 13         ids: element.val()
 14       }, function (data) {
 15         callback(data.users);
 16       });
 17     },
```

to ```
 11     initSelection: function (element, callback) {
 12       $.ajax({
 13         url: Spree.routes.users_api,
 14         data: { ids: element.val() },
 15         headers: { "X-Spree-Token": Spree.api_key },
 16         success: function (data) {
 17           callback(data.users);
 18         }
 19       });
 20     },
```

simo163:
@joelind Thanks for explaining it better, this is exactly what I meant. I sent a PR for it in case #1353

PR #1353, opened 2016-07-25
fix user picker when initializing ... #1352

simo163:
fix user picker when initializing rules users select in the edit protion page

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/user_picker.js
+++ b/backend/app/assets/javascripts/spree/backend/user_picker.js
@@ -12,7 +12,8 @@ $.fn.userAutocomplete = function () {
       Spree.ajax({
         url: Spree.routes.users_api,
         data: {
-          ids: element.val()
+          ids: element.val(),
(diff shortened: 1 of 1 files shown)


gmacdougall:
`Spree.ajax` is supposed to pass the api_key in the headers of the request. If that's not working, we should fix things on that side rather than forcing this method of authentication.

Can you please provide a more detailed description of what problem you're seeing?

simo163:
It's strange, you're right Spree.ajax supposes to do it.

Doing this must reach the problem:
- create promotion
- add a rule of type User
- select some users and save
- reload the edit page
- can't load the users selected

I pass a [screeshot](http://img15.hostingpics.net/pics/585776promotionsuserrules.png)

jeremyw:
I think I fixed the root cause of this issue here: https://github.com/solidusio/solidus/pull/1349

gmacdougall:
I believe that does indeed fix the issue. Closing this PR.

Issue #1066, opened 2016-04-13
Can't cancel an order with a refund


travisp:
This is using solidus_gateway with Stripe, so I'm not sure if it's ultimately an issue with solidus_gateway or solidus, but:

If an order has a refund, and then an attempt to cancel the order is made, the cancellation fails with the error that a charge has already been refunded, because the payment is still marked as completed (even if it was refunded in full).

I'm not fully sure what the expected behavior here is, but after a refund has happened, it becomes impossible to cancel the order itself.

PR #1355, opened 2016-07-25
Allow cancelling orders that have been fully refunded

Sinetheta:
This fixes https://github.com/solidusio/solidus/issues/1066 but I could use some feedback.

After an order is cancelled, all completed payments are also cancelled [here](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L750). Each payment then tries to call cancel on it's payment method [here](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment/processing.rb#L82), the implementation of which is left to the provider [here](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment/processing.rb#L82).

I can't speak for the other gateways, but Braintree's cancel looks like this in solidus_gateway

``` rb
def cancel(response_code)
  provider
  transaction = ::Braintree::Transaction.find(response_code)
  # "A transaction can be refunded if its status is settled or settling.
  # If the transaction has not yet begun settlement, it should be voided instead of refunded.
  if transaction.status == Braintree::Transaction::Status::SubmittedForSettlement
    provider.void(response_code)
  else
    provider.refund(response_code)
  end
end
```

The reason why it's sufficient to simply check whether a payment is "settled" or not, is because the Braintree provider [docs say that](https://developers.braintreepayments.com/reference/request/transaction/refund/ruby#amount)

> If you do not specify an amount to refund, the entire transaction amount will be refunded.

and

> If you have already partially refunded a transaction, performing another refund without specifying the balance it will create a refund for the remaining non-refunded amount of the transaction.

but there's one more case, where a payment has already been fully refunded (through one or many refunds). In that case we get an error. This error is what causes the order cancellation to fail.

We could address this anywhere along the chain, but here are my thoughts.
1. Not in the Braintree ruby SDK, I can understand how a failure to create _any_ refund could be considered an error even if a partial refund is not.
2. Not in solidus_gateway, because we don't currently pass the `payment` when calling cancel on the gateway and I'm not sure we should.

Any ideas how I might add a regression test for this? The behaviour seems specific to one (albeit an important) gateway.

### DIFF
--- a/core/app/models/spree/payment.rb
+++ b/core/app/models/spree/payment.rb
@@ -146,6 +146,11 @@ module Spree
       credit_allowed > 0
     end
 
+    # @return [Boolean] true when this payment has been fully refunded
+    def fully_refunded?
+      refunds.map(&:amount).sum == amount
+    end
+
     # @return [Array<String>] the actions available on this payment
     def actions
       sa = source_actions
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -852,7 +852,7 @@ module Spree
 
     def after_cancel
       shipments.each(&:cancel!)
-      payments.completed.each(&:cancel!)
+      payments.completed.each { |payment| payment.cancel! unless payment.fully_refunded? }
       payments.store_credits.pending.each(&:void_transaction!)
 
       send_cancel_email
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -43,6 +43,8 @@ RSpec.describe Spree::Order, type: :model do
   end
 
   describe "#cancel!" do
+    subject { order.cancel! }
+
     context "with captured store credit" do
       let!(:store_credit_payment_method) { create(:store_credit_payment_method) }
       let(:order_total) { 500.00 }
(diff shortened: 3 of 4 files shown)


core/app/models/spree/payment.rb
```diff
@@ -145,6 +145,11 @@ def can_credit?
       credit_allowed > 0
     end
 
+    # @return [Boolean] true when this payment has been fully refunded
+    def fully_refunded?
+      refunds.sum(:amount) >= amount
```
peterberkenbosch: should `==` instead of `>=` be sufficient? If you refunded more then the amount paid there are some other issues going on IMHO.

peterberkenbosch:
This makes a lot of sense, curious why we do not have a `payment#refunded` state though.

gmacdougall:
I agree that this makes sense. I can't think of anywhere else I would prefer to deal with this.

👍

seantaylor:
Also ran into this issue on a store. 👍  for getting this in. Seems reasonable to me.

core/app/models/spree/payment.rb
Sinetheta: Fair enough. I don't like when we try to catch problems tangentially, but in it's only current use that case would raise an error, which would probably be useful. I'm cool with swicthing to `==`.

Senjai:
:+1: @jhawthorn any opinions?

jarednorman:
@Sinetheta Thoughts on testing this?

Sinetheta:
@jarednorman I'm open to thoughts on how we could test this. I've added the simple test, some assertions for the new payment method, but it doesn't address the real feature (cancelling orders with refunds) because we don't currently have anything in the way of refund specs.

core/app/models/spree/payment.rb
```diff
@@ -146,6 +146,11 @@ def can_credit?
       credit_allowed > 0
     end
 
+    # @return [Boolean] true when this payment has been fully refunded
+    def fully_refunded?
+      refunds.sum(:amount) == amount
```
mamhoff: I think we can save a DB lookup here by using the in-memory refunds: 

```
refunds.map(&:amount).sum == amount
```

PR #1356, opened 2016-07-26
Remove update_adjustments AR callbacks

jhawthorn:
The existing behaviour was that line_items and shipments had their adjustments recalculated after save. This would update the item's adjustments and the item's `*_total` columns. However this doesn't update the order's `*_total` columns until `order.update!` is run.

Since order.update! will update these values anyways, and since the order is in an inconsistent state until `order.update!` is run, we should remove these callbacks.

Related to #1252

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,12 @@
 
 *   Removed "Clear cache" button from the admin [#1275](https://github.com/solidusio/solidus/pull/1275)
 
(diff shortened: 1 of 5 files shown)


mamhoff:
This is excellent. 💯

jordan-brough:
👍 Thank you!

cbrunsdon:
I'm a serious :+1: on this as well

PR #1357, opened 2016-07-28
Cancelling Inventory updates tax_total & item total

brendandeere:
Currently, when inventory is cancelled on an order, the system creates a
negative line item adjustment for the the line item total including
taxes. This works for refunding the customer appropriately because the
order.total takes the adjustments into account. However the
order.item_total and order.tax_total still reflect the number of line
items the order was completed with.

I've provided two tests to the order cancellation spec which will pass
when the orders totals are correctly updated.

The question for the community is: Should cancelling and inventory unit
update the order's tax_total and/or item_total and how this should be
implemented with regards to the existing cancellation line item
adjustment.

### DIFF
--- a/core/spec/models/spree/order_cancellations_spec.rb
+++ b/core/spec/models/spree/order_cancellations_spec.rb
@@ -95,6 +95,27 @@ describe Spree::OrderCancellations do
       expect { subject }.to change { order.total }.by(-10.0)
     end
 
+    it 'adjusts the order tax_total', pending: 'community discussion' do
+      # set up required for tax adjustments on the line item
+      # Should probably me moved to apply to the whole spec if this is a
+      # change that is accepted
+      line_item = order.line_items.first
+      create :tax_rate, zone: order.tax_zone, tax_category: line_item.tax_category
+      line_item.send :update_tax_charge
+      order.update!
+
+      expect { subject }.to change { order.tax_total }.from(1).to(0)
+    end
+
+    it 'adjusts the order item total', pending: 'community discussion' do
+      # set up required for tax adjustments on the line item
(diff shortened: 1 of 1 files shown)


gmacdougall:
Hi Brendan,

I'm thinking through this and not coming up with any reasonable solution for this.

Lets take an example of an order for 3 Line Items costing $20 each, and a tax of $4 each.

This leads to the following situation:

Amount | Value
----------|----
Item Total | 60
Tax Total | 12
Total | 72

If one of the two line items is returned we get the following:

Amount | Value
----------|----
Item Total | 60
Tax Total | 12
Adjustment Total | -24
Total | 48

The total calculates correctly as you mentioned because of the adjustment total.

If we were to update the item total we would have the following:

Amount | Value
----------|----
Item Total (2 items only) | 40
Tax Total | 8
Adjustment Total | -24
Total | ???

In this case the total would be wrong because the sum of all the values is only 24.

As far as I can tell we have a few potential solutions:

1. Leave it as is and deal with the fallout
2. Stop representing cancelled line items as adjustments, and instead represent them with their own proper field (cancelled_item_total or something like that). Then we could explicitly ignore that field when calculating the order total, but still reflect the cancelled item correctly. We would need to add fields to both the `LineItem` and `Order` to deal with this correct.
3. We could update the line items on the order to reflect only non-cancelled items and trust `order.update!` to do the magic.

cbrunsdon:
Closing this as I don't think anyone has a solid plan on how to move this forward (but the PR is good information if anyone moves it forward).

I personally believe our best bet is to stop representing cancelled items as adjustments, but that is a pretty sizable change.

jhawthorn:
I've created #2260 to track various issues with UnitCancel

Issue #1161, opened 2016-05-18
Remove icons from buttons and tabs


Mandily:
This is a bullet from #520 Solidus Admin branding that I wanted to separate out so that any interested parties could take a crack at it. 

Icons are generally a universal language that don't require words to explain them. They're great for things like stop signs and bathrooms, and in our case edit functions and calendar pickers, etc. They're not great when they're not clear and need text beside them to explain them.

Currently, there are icons in the left side navigation, tabs and some of the buttons. The ones in the navigation are useful for the #666 collapsible nav feature. The ones that stand solo beside the tables are useful too. Many of the others in the admin do not add any value to the page, but instead clutter it. To reduce the clutter and save a little space, we can remove them.

This issue is left for anyone in the community who would like to try their hand at contributing back to core. Leave a message below if you'd like to claim it or have any questions.

tvdeyen:
👍

RitaD:
I’m currently working on this, expect a PR soon!

PR #1358, opened 2016-07-29
Remove all icons from tabs and buttons

kennyadsl:
This PR refers to #1161 issue. I'm not 100% sure we removed all of them but, hey, they are a lot and some of them are very hidden in the UI.

What we did:

- removed all icons from tabs and buttons
- fixed spacing in buttons without icons
- added deprecation warnings for `:icon` option in `button_link_for` and `icon_name` arg in `button` helpers methods since they are not used anymore. I'm not sure about this, but it could be good if we want to use another solution in future versions and we'll add some new style which could be incompatibile with them. Please, let me know your thoughts.

I'm available to squash commits as well...

I attach some screenshot of the result, just to be sure we did what needed:

schermata 2016-07-29 alle 16 19 31

schermata 2016-07-29 alle 16 20 00

schermata 2016-07-29 alle 16 52 03

cc @Mandily

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Solidus 1.4.0 (master, unreleased)
 
+*   Backend: UI, Remove icons from buttons and tabs
+
+*   Backend: Deprecate args/options that add icons to buttons
+
 *   Update Rules::Taxon/Product handling of invalid match policies
 
     Rules::Taxon and Rules::Product now require valid match_policy values.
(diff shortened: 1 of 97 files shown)


Mandily:
This looks so much cleaner!! Thanks for tackling this.

jhawthorn:
:+1: LGTM

tvdeyen:
:+1: great. Stashing some commits in smaller groups would be nice, but not mandatory.

kennyadsl:
@tvdeyen Yes, squashing was needed even in my opinion. Now we have 5 commits, much better. Let me know if you guys think something else is needed here.

mamhoff:
Great. 👍

PR #566, opened 2015-12-04
Raise InsufficientStock if and only if there is insufficient stock

jhawthorn:
Previously, `InsufficientStock` was raised if any StockLocations were fully out of inventory. This was incorrect because it was possible other stock locations could have fulfilled the inventory. This was also incorrect because the stock location could have some, but insufficient inventory, and not raise the exception (an incomplete package would be returned). Now the coordinator checks that the package is complete and raises `InsufficientStock` if it is incomplete for any reason.

--

Stock::Packer currently raises InsufficientStock under specific circumstances:
* There is a stock item for a variant in the stock location.
* There is exactly 0 inventory remaining for that variant in the stock location.
* The variant does not allow backordering.

This is odd, because it doesn't raise `InsufficientStock` when there is some but insufficient inventory (it adds the amount available to the package), or if the stock item doesn't exist at all in that stock location. These three cases should probably be handled in the same way.

The way the Packer is used is to build the fullest possible package for this stock location. As it's possible to fulfill an variant half from one location and half from another, this method should never raise `InsufficientStock` because it has insufficient information to do so.

PR #1359, opened 2016-07-29
Rescue Spree::Order::InsufficientStock in backend

DanielePalombo:
When an order contains a line_item which requires a quantity greater
than available in not
backorderable stock items, a Spree::Order::InsufficientStock exception
is raised, but there
are no rescue in controllers.

This change adds the rescue for Payments and CustomerDetails
controllers, displays an
`:insufficient_stock_for_order` error message and redirects to the
specific page.

Still needs to be ported in Spree::Api controllers

ref #566


jhawthorn:
:+1: (pending whitespace fixes)

Thanks for the specs

mamhoff:
👍

PR #872, opened 2016-02-18
Moving selected_shipping_rate method to has_one association

dfmedeiros:
Moving this method to a has_one association will help us to preload the association fixing a lot of N + 1 queries when dealing with shipments across the application.

PR #1360, opened 2016-07-29
Move :selected_shipping_rate to association

brchristian:
@jhawthorn I noticed that the `selected_shipping_method` has changed in Solidus relative to the current Spree master (my application is on Spree 2.4 currently), and so I just wanted to make sure that this proposed PR jives with your intentions in the previous refactor of this method.

This takes the existing `selected_shipping_rate` method and turns it into a `has_one` association. This enables eager loading, for instance `Spree::Shipment.includes(:selected_shipping_rate)`, which can speed up certain queries and avoid n-query issues in some cases.


mamhoff:
Hi @brchristian we've had exactly this PR in the past and decided to not accept it. For reasons, please look at the old PR - essentially we do not want the additional setter methods for this relation to be auto-created and believe the speedup can be achieved differently.

mamhoff:
https://github.com/solidusio/solidus/pull/872 was the old one, sorry, forgot to paste this before.

PR #1361, opened 2016-07-29
Don't throw error when an order doesn't have unprocessed payments

luukveenis:
This has been causing us grief with certain payment methods where the payment is already authorized when we get to the confirm step. In this case the payment may be pending already and it will raise an error.


peterberkenbosch:
💯

mamhoff:
👍

jhawthorn:
:+1:

PR #1364, opened 2016-08-02
[rails5] bump the rabl version to the version specified here

peterberkenbosch:
the API gemspec pointed to an older version of rabl, the fork that is used for rails5 is set to 0.13.0.. this PR updates the required version for rabl in the API to the version specified here https://github.com/jhawthorn/rabl/blob/rails5/lib/rabl/version.rb


mamhoff:
👍

gmacdougall:
👍 

Thanks Peter!

PR #1367, opened 2016-08-02
[rails5] render text is deprecated, using render plain instead.

peterberkenbosch:


jhawthorn:
Thanks peter. I've merged this into #1342

peterberkenbosch:
👍

PR #1368, opened 2016-08-02
Fix running ItemAdjustments multiple times against the same record

jhawthorn:
Rails 4.2 doesn't update [`ActiveModel::Dirty#changed_attributes`](http://api.rubyonrails.org/classes/ActiveModel/Dirty.html#method-i-changed_attributes) when using `update_columns`. This could cause incorrect behaviour when `ItemAdjustments#update` was called multiple times on the same in-memory object.

Example (buggy) scenario:
``` ruby
> item.changes #=> {}
> ItemAdjustments.new(item)
# adjustment is (somehow) updated to be 10.
# Change is persisted to database
> item.changes #=> {"adjustment_total"=>[BigDecimal(0.0), BigDecimal(10.0)], ...}
> item.changed? #=> true
# set as changed? even though we have persisted using update_columns
> ItemAdjustments.new(item)
# adjustment is updated back to 0.
# NOT PERSISTED TO DATABASE because changed? is false
> item.changes #=> {}
> item.changed? #=> false
```

This PR fixes this issue by manually removing the attributes we know to have been persisted and changed from the `attributes_changed_by_setter` hash.

On rails 5 `update_columns` seems to update the changed hash properly and this is unneeded.

### DIFF
--- a/core/app/models/spree/item_adjustments.rb
+++ b/core/app/models/spree/item_adjustments.rb
@@ -70,6 +70,13 @@ module Spree
         updated_at: Time.current
       ) if @item.changed?
 
+      # In rails 4.2 update_columns isn't reflected in the changed_attributes hash,
+      # which means that multiple updates on the same in-memory model will
+      # behave incorrectly.
+      # In rails 5.0 changed_attributes works with update_columns and this is
+      # unnecessary.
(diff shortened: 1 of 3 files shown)


gmacdougall:
👍 

Thanks for finding this John. It's a bit of a strange one.

PR #1370, opened 2016-08-03
Correct use of CONCAT() in very old migration

adammathys:
In MySQL, `CONCAT()` will return `NULL` if any of the passed in arguments are `NULL`. There are scenarios in which a user might have a first or last name but not both. In this case, this migration could be losing data.

Instead, we can use `CONCAT_WS()` which will simply skip `NULL` arguments. So we're less likely to lose data while maintaining the same result for all other cases.

[1] http://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_concat
[2] http://dev.mysql.com/doc/refman/5.7/en/string-functions.html#function_concat-ws


jhawthorn:
:+1: 

This is a very old migration (spree 1.1 -> 1.2), but there is no reason not to fix it. It may help others upgrading.

It looks like the postgres/sqlite migration is also broken in the same way, which another PR could fix.

gmacdougall:
👍

PR #1375, opened 2016-08-04
API - Countries: Do not query states for the index

ridem:
An index call would result in a ridiculous amount of queries. The states are not meant to be displayed for that view.

### DIFF
--- a/api/app/controllers/spree/api/countries_controller.rb
+++ b/api/app/controllers/spree/api/countries_controller.rb
@@ -5,7 +5,7 @@ module Spree
 
       def index
         @countries = Country.accessible_by(current_ability, :read).ransack(params[:q]).result.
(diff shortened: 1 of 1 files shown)


jhawthorn:
From what I can see this removes just one query.

```
  Spree::State Load (1.2ms)  SELECT "spree_states".* FROM "spree_states" WHERE "spree_states"."country_id" IN (3, 6, 62, 11, 1, 8, 5, 9, 4, 10, 7, 14, 13, 12, 16, 32, 23, 19, 18, 36, 20, 37, 25, 27, 33) ORDER BY "spree_states"."name" ASC
```

But it does seem like that query is unnecessary, and we should remove it.

:+1: Thank you

ridem:
👍 You're right, my bad, this was just one query!

PR #1376, opened 2016-08-04
Turbolinks 5.0 friendly frontend

mtomov:
This change allows all Spree frontend javascript to work seamlessly with Turbolinks 5.0, while not disturbing if Turbolinks is not present.

* I've tested with Turbolinks 5.0 the following sections with Spree JavaScript and they all work when fetched with Turbolinks (through get requests)
  - product page's thumbs
  - address validation on address stage checkout
  - credit card validation on payment stage checkout
  - cart delete button press
  - coupon code applicator

_I can re-arrange the commits and all if needed_

### DIFF
--- a/core/app/assets/javascripts/spree.js.coffee.erb
+++ b/core/app/assets/javascripts/spree.js.coffee.erb
@@ -1,7 +1,10 @@
 #= require jsuri
 class window.Spree
   @ready: (callback) ->
-    jQuery(document).ready(callback)
+    if typeof Turbolinks isnt 'undefined' and Turbolinks.supported
+      jQuery(document).on 'turbolinks:load', -> callback(jQuery)
+    else
+      jQuery(document).ready(callback)
 
   @mountedAt: ->
     "<%= Rails.application.routes.url_helpers.spree_path(trailing_slash: true) %>"
--- a/frontend/app/assets/javascripts/spree/frontend/checkout.js.coffee
+++ b/frontend/app/assets/javascripts/spree/frontend/checkout.js.coffee
@@ -7,6 +7,3 @@
 Spree.disableSaveOnClick = ->
   ($ 'form.edit_order').submit ->
     ($ this).find(':submit, :image').attr('disabled', true).removeClass('primary').addClass 'disabled'
-
-Spree.ready ($) ->
-  Spree.Checkout = {}
--- a/frontend/app/assets/javascripts/spree/frontend/checkout/address.js.coffee
+++ b/frontend/app/assets/javascripts/spree/frontend/checkout/address.js.coffee
@@ -1,4 +1,4 @@
-$ ->
+Spree.ready ($) ->
   if $('#checkout_form_address').is('*')
     # Hidden by default to support browsers with javascript disabled
     $('.js-address-fields').show()
--- a/frontend/app/assets/javascripts/spree/frontend/checkout/payment.js.coffee
+++ b/frontend/app/assets/javascripts/spree/frontend/checkout/payment.js.coffee
@@ -29,12 +29,11 @@ Spree.ready ($) ->
         ($ '#payment_method_' + @value).show() if @checked
       )
 
-      ($ document).on('click', '#cvv_link', (event) ->
+      ($ '#cvv_link').on 'click', (event) ->
         windowName = 'cvv_info'
         windowOptions = 'left=20,top=20,width=500,height=500,toolbar=0,resizable=0,scrollbars=1'
         window.open(($ this).attr('href'), windowName, windowOptions)
         event.preventDefault()
-      )
 
       # Activate already checked payment method if form is re-rendered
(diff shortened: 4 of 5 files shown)


tvdeyen:
@mtomov Turbolinks week for ya? 😆 

👍

jhawthorn:
I'm a little concerned that `Spree.ready` is now going to fire at least twice on a page (more if turbolinks is being used). Users may have used it in a way such that that will cause issues.

Sinetheta:
Does the Spree.Ready code really survive being reloaded like that?

It looks like the codebase has a bunch of doc readys and a few spree readys. If only the latter happens I would expect some information to be lost [checkout.js.coffee#L12](https://github.com/solidusio/solidus/blob/4eaef6ca53a11bfad7f442c3f6dca3c3e3b5b4b6/frontend/app/assets/javascripts/spree/frontend/checkout.js.coffee#L12)

or for delegated event handlers to pile up [payment.js.coffee#L32](https://github.com/solidusio/solidus/blob/4eaef6ca53a11bfad7f442c3f6dca3c3e3b5b4b6/frontend/app/assets/javascripts/spree/frontend/checkout/payment.js.coffee#L32)

mtomov:
@Sinetheta 

 - delegate event handlers are the best, since they never pile up, and don't have to be run after jQuery ready (Dom ready event) as they are run on the `document` node, which exists still before the html of the page has loaded. Turbolinks doesn't touch `document`, only replaces the `body` and merges the `head`

- `Spree.Checkout` - I wasn't aware of this one, but I couldn't find anything else in the codebase to use it. I think such an object definition is very hard to use even in a normal scenario as one cannot predict when it will be available (before or after another jQuery ready code). So, best way to go about this one is to actually delete it.

- If it really survives - it does indeed. As I mentioned above, I tested the product page with the thumbs, the checkout pages with the address validator and the credit card validator on the payment page. That's pretty much [all the frontend JS](https://github.com/solidusio/solidus/tree/master/frontend/app/assets/javascripts/spree/frontend). I think the only 2 I haven't looked at are:
  - coupon code applicator
  - cart.js

  I will see to test those two out as well.

@jhawthorn 

- I don't think that `Spree.ready` will fire more than once per page view with or without turbolinks, as `Spree.ready` fires when one of the 3 events happens - `jQuery.ready`, `page:load`, `turbolinks:load`. Neither of those events can happen on the same page load as any of the other two. 

  I can  though place some debugging code in the `Spree.ready`, and confirm that for certain.

mtomov:
Updated to only Support Turbolinks 5.0 as I didn't want to add the extra code to support 2.x. Tested, and all Dom ready events fire exactly once per page view, and all works. Let me know if you have any more concerns.

mtomov:
Also, just checked on `Spree.Checkout`, and it [has been used](https://github.com/solidusio/solidus/blob/ab5d75278c6f6b0aa1ea20769e6c2d1da0716ae7/frontend/app/assets/javascripts/spree/frontend/checkout/address.js.coffee) by `address.js` , but the following two commits:
- https://github.com/solidusio/solidus/commit/e49629ac7d7e3ff431e9ebffe08562598dfb83b3
- https://github.com/solidusio/solidus/commit/08d56b3f930ca4b9180db95e83f930ea7685d6ea

get rid of it all, and for good. I can add a commit to actually remove it from `checkout.js` as well.

Sinetheta:
Hey @mtomov, thanks for getting rid for `Spree.Checkout`. My concern with delegated events is that we have one bound *inside* of a `Spree.ready`, so each page load will add another.

- [x] Tell Spree.ready not to fire twice at page load.
- [x] Replace all document ready (and variants like `$ ->` with `Spree.ready` for code that interacts directly with the page.
- [] Move all event handlers delegated to document or body out of `Spree.ready`. Putting those in doc ready is probably fine (because it will never rerun). [payment.js.coffee#L32](https://github.com/mtomov/solidus/blob/b32dd92d388e5cb6808d405e8f0ef1e165ac5a0c/frontend/app/assets/javascripts/spree/frontend/checkout/payment.js.coffee#L32) 

I've never used turbolinks, but those are the only concerns I can think of. Luckily there isn't much JS in the frontend.

mtomov:
Right, I see what you mean, and you are right that there will be repetition of this code. Maybe in this case I can switch it to a normal event attachment. 

I will do that, and will remove `Spree.Checkout`. Thanks for the feedback! Much appreciated!

mtomov:
And updated! Thanks for the feedback.

daetherius:
Hi! I'm working on a project with Solidus v1.2 and we recently updated it to use TL5. I can confirm the address form validation is broken when coming from a turbolink. If this get merged, is there a way that this fix will be backported to the version I'm using? Or should I patch it myself?

Thanks in advance!

jhawthorn:
Thanks @mtomov

mtomov:
Not a problem, thank you as well :+1:

PR #1377, opened 2016-08-05
Add html templates

DanielePalombo:
- Add shipped_email.html.erb
  
  When the ShipmentMailer's template was moved under CartonMailer, only the text format was moved. 
  ref https://github.com/solidusio/solidus/commit/9ff3f310e0b3f274fce6f5a2b25fe7dd6859bab7 
  
  I change the test from:
  expect(shipment_email.body).not_to include(%{Out of Stock})
  in favour of
  expect(shipment_email).not_to have_body_text(%{Out of Stock})
  for use the default_part_body method.
- Add inventory_cancellation_email.html.erb based on https://github.com/solidusio/solidus/blob/c5b9e0242291a2249ea62b046cc973b530957d51/core/app/views/spree/order_mailer/inventory_cancellation_email.text.erb

### DIFF
--- a/core/app/views/spree/order_mailer/inventory_cancellation_email.html.erb
+++ b/core/app/views/spree/order_mailer/inventory_cancellation_email.html.erb
@@ -14,8 +14,8 @@
         <% @inventory_units.each do |item| %>
           <tr>
             <td><%= item.variant.sku %></td>
-            <td><%= raw(item.variant.product.name) %></td>
-            <td><%= raw(item.variant.options_text) -%></td>
+            <td><%= item.variant.product.name %></td>
+            <td><%= item.variant.options_text -%></td>
           </tr>
         <% end %>
       </table>
--- a/core/app/views/spree/order_mailer/inventory_cancellation_email.html.erb
+++ b/core/app/views/spree/order_mailer/inventory_cancellation_email.html.erb
@@ -1,0 +1,26 @@
+<table>
+  <tr>
+    <td>
+      <p class="lede">
+        <%= Spree.t('order_mailer.inventory_cancellation.dear_customer') %>
+      </p>
+      <p>
+        <%= Spree.t('order_mailer.inventory_cancellation.instructions') %>
+      </p>
+      <p>
+        <%= Spree.t('order_mailer.inventory_cancellation.order_summary_canceled') %>
+      </p>
+      <table>
+        <% @inventory_units.each do |item| %>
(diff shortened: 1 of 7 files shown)


mamhoff:
LGTM. 👍

core/app/views/spree/order_mailer/inventory_cancellation_email.html.erb
```diff
@@ -0,0 +1,26 @@
...
+      </p>
+      <p>
+        <%= Spree.t('order_mailer.inventory_cancellation.order_summary_canceled') %>
+      </p>
+      <table>
+        <% @inventory_units.each do |item| %>
+          <tr>
+            <td><%= item.variant.sku %></td>
+            <td><%= raw(item.variant.product.name) %></td>
+            <td><%= raw(item.variant.options_text) -%></td>
```
gmacdougall: I don't believe that these two should use `raw`. It is user input, so I don't think it should be trusted.

tvdeyen: If text formatting is the reason for using `raw`, we probably should use `sanitize` here.

DanielePalombo: I used `raw`  because i copied from https://github.com/solidusio/solidus/blob/c5b9e0242291a2249ea62b046cc973b530957d51/core/app/views/spree/order_mailer/confirm_email.html.erb#L18

Probably `sanitize` is better, i change also in other templates.

DanielePalombo:
Conflicts are fixed. CHANGELOG updated.

core/app/views/spree/reimbursement_mailer/reimbursement_email.html.erb
```diff
@@ -21,13 +21,13 @@
           <% @reimbursement.return_items.exchange_requested.each do |return_item| %>
             <tr>
               <td>
-                <%= return_item.variant.sku %> <%= raw(return_item.variant.name) %> <%= "(#{raw(return_item.variant.options_text)})" if return_item.variant.options_text.present? %>
+                <%= return_item.variant.sku %> <%=return_item.variant.name %> <%= "(#{return_item.variant.options_text})" if return_item.variant.options_text.present? %>
```
tvdeyen: missing space

DanielePalombo:
Updated. Thanks!

PR #1381, opened 2016-08-05
Document OrderContents#add

brendandeere:
Documents the params and in particular the options that are able to be
passed in.

I was rooting around here tracing what happens during this method (and the method it calls) and I thought it might be useful to document which options are used by any methods in this chain

### DIFF
--- a/core/app/models/spree/order_contents.rb
+++ b/core/app/models/spree/order_contents.rb
@@ -6,6 +6,19 @@ module Spree
       @order = order
     end
 
+    # Add a line items to the order if there is inventory to do so
+    # and populate Promotions
+    #
+    # @params [Spree::Variant] :variant The variant the line_item should
+    #   be associated with
+    # @params [Integer] :quantity The line_item quantity
+    # @param [Hash] :options Options for the adding proccess
+    #   Valid options:
+    #     shipment: [Spree::Shipment] LineItem target shipment
+    #     stock_location_quantities:
+    #       stock_location_id: The stock location to source from
+    #
+    # @return [Spree::LineItem]
     def add(variant, quantity = 1, options = {})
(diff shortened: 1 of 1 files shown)


core/app/models/spree/order_contents.rb
```diff
@@ -6,6 +6,23 @@ def initialize(order)
...
+    # @params [Integer] :quantity The line_item quantity
+    # @param [Hash] :options Options for the adding proccess
+    #   Valid options:
+    #     shipment: [Spree::Shipment] LineItem target shipment
+    #     stock_location_quantities:
+    #       stock_location_id: The stock location to source from
+    #       quanity: The quantity of stock to source from the location
+    #     id: Line item options id
+    #     variant_id: Line item options variant_id
+    #     quantity: line item options quantity
```
jhawthorn: Are `id`, `variant_id`, `quantity` really valid options? They probably should not be used.

brendandeere: [Strong params](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_contents.rb#L133) Here grab out these [keys](https://github.com/solidusio/solidus/blob/master/core/lib/spree/permitted_attributes.rb#L61)

They seem to get saved to the arbitrary `options` hash associated with line items [here](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/line_item.rb#L110)

Seems weird but I included them since it seems to be allowed ....

jhawthorn: We should probably not document them since they may have an effect but will likely produce strange behaviour

mamhoff:
👍

jhawthorn:
:+1: Thanks Brendan

brendandeere:
Thanks guys !

PR #1384, opened 2016-08-11
Replace spree_api_key with json web token format (JWT).

skukx:
See https://github.com/solidusio/solidus/issues/1383


skukx:
Failures on truncation for `spree_api_key` column. The column is no longer needed So maybe drop the column and add a boolean `allowed_api_access`

PR #1385, opened 2016-08-11
Add 'Create' and 'Destroy' API calls for Credit Card resource

photoshovel:
Enable use case of system external of Solidus handling the creation of payment instruments via 3rd party payment provider.

For example, mobile application's server API creates entries in Braintree for customers' credit cards.  Mobile application's server API then calls Solidus API to create credit card entities to use in Solidus commerce transactions.

Solidus test suite passes.  JMeter integration tests created around this case passed when used with Braintree gateway plugin and Braintree API sandbox.

### DIFF
--- a/api/app/controllers/spree/api/credit_cards_controller.rb
+++ b/api/app/controllers/spree/api/credit_cards_controller.rb
@@ -22,16 +22,16 @@ module Spree
       end
 
       def create
-        # @credit_card = Core::Importer::CreditCard.new(nil, @_params[:user_id], @_params[:credit_card]).create
-        @credit_card = Core::Importer::CreditCard.new(nil, @_params[:user_id], credit_card_create_params)
-        credit_card = @credit_card.create
+        authorize! :create, CreditCard
+        @credit_card = Core::Importer::CreditCard.import(user, credit_card_create_params)
+        respond_with(@credit_card, status: 201, default_template: :show)
+      end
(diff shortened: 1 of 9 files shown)


mamhoff:
I think it'd be great to have an API for creating payment sources like credit cards. 

However: 
- @jordan-brough is working on supporting payment sources other than credit cards, and it'd be great to support that work with this endpoint. 
- you're targeting the 1.3 branch, on which we won't accept new features
- the commits can IMO be squashed
- the commits are not connected to your github account, which is fine with us but might not be fine with you
- I know the API relies heavily on the order importer, but it really should use core instead. If we introduce new endpoints, it might be a good thing to just use the core models instead of the importer.

photoshovel:
Will try and connect with @jordan-brough to see how supporting other payment sources might look in the future.

I'll create a new pull request with squashed commits against Solidus master branch.  We needed the functionality now, so will just use our fork of 1.3.  Set email address in GitHub client to try and connect commits to my GitHub account.

Don't know enough about Solidus yet to understand order importer vs / core models.  I had tried to write in similar style of existing API calls.  Would someone be able to direct me to similar use of core models I could use as an example?

photoshovel:
Abandoning this pull request and replacing with [PR 1387](https://github.com/solidusio/solidus/pull/1387) against Solidus master.

PR #1387, opened 2016-08-12
Add 'Create' and 'Destroy' API calls for Credit Card resource

photoshovel:
Enable use case of system external of Solidus handling the creation of payment instruments via 3rd party payment provider.

For example, mobile application's server API creates entries in Braintree for customers' credit cards.  Mobile application's server API then calls Solidus API to create credit card entities to use in Solidus commerce transactions.

Solidus test suite passes.


mamhoff:
Would you mind rebasing the empty merge commit out of this PR? 

Also, I would love you to make Hound happy by removing the empty newlines. We use Hound in order to enforce a (pretty lenient) style guide that helps us reduce noise in diffs down the line, and ignore his advice only in rare, well-reasoned cases.

photoshovel:
@mamhoff  I will try and remove that merge commit.  Had a commit in there to clean-up those newlines but appears to have gotten left out/dropped when I rebased.  Will get that back in there too.

jhawthorn:
Thanks, but with the changes we've made to genericize our payment sources, our payment sources are not just credit cards.

I'm not eager to add another way to create credit cards in the system. The fewer endpoints we have for this the better.

photoshovel:
Bummer.  We've had this running on our own branch since last year and was hoping it would get folded-in.  For our Solidus deployment, we don't utilize the Admin/Customer web interfaces at all.  Run inside AWS Elastic Container Service.  Solidus is just another 3rd-party API to us so we can keep our 'books' in-house.  Would be great if the API was more richly-featured.

Created other new API calls I've been meaning to submit, mostly around promotions/coupon-codes to allow just-in-time creation of Customer-specific coupons, on demand, as well as for 'Brand Ambassador' promotion coupons.

mamhoff:
photoshovel: You can totally add this route to your app, and it'll probably be supported for a long time. The credit card model is not going away. If you were to rebuild it using the `wallet`, that would have a much better chance of being merged. 

We're understanding that the reality of payment sources is much richer than credit cards, and that's  the primary reason for this not being merged: We don't want to maintain an endpoint for one specific payment source, but for all the ones we support. 

Solidus is also meant to be a lean base for you to build what you need on top. Do open an issue next time describing what you intend to do and get early feedback before you put work into creating a PR. Your contributions are much appreciated!

PR #1388, opened 2016-08-13
Added preference for opting to use json web tokens

skukx:
TODO: Write migration to add column to user for `spree_api_access` then api authentication method to obtain json web token from server.

### DIFF
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -68,8 +68,9 @@ module Spree
       end
 
       def valid_token?
-        @token ||= JWT.decode(api_key,
-          signing_token, true, { algorithm: 'HS256' }).first rescue nil
+        @token ||= JWT.decode(api_key, signing_token, true, { algorithm: 'HS256' }).first
+      rescue
+        nil
       end
 
       def unauthorized
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -68,10 +68,8 @@ module Spree
       end
 
       def valid_token?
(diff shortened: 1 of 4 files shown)


skukx:
See: https://github.com/solidusio/solidus/issues/1383

cbrunsdon:
I don't think its reasonable to support hacking JWT in with an if/else like this. I think the approach we should take is make the authentication system pluggable and then offer JWT as a separate implementation (preferably in an extension?).

Doing it this way makes it tougher for people to add more complexity to authorization (like supporting two methods of authorization).

Also we'd need to add JWT as an explicit dependency of solidus_api.

jhawthorn:
As a toggle, I fear this would just split our focus and efforts in half, not something we want for a feature as important as authentication.

I'm interested in the idea of using JWT for _all_ of our authentication, but it would have to be a much more complete conversion than this. It would be nice if JWT was used to describe both users and order tokens. A complete JWT implementation would need to duplicate our "reset token on password change feature" and resetting tokens in general, which is a little awkward for JWT since it loses its statelessness (one of the main features). 

I also like Clarke's suggestion of moving authentication to a pluggable class and creating an extension. This is probably the most achievable way to introduce JWT in the short term.

skukx:
To work with "reset token on password change feature" it would be best to keep the token life short and implement a refresh token.  When validating refresh token, we could then check whether a password change has occurred. If so then we require sign in again.  Also while validating the refresh tokens we can check the db for any blacklisted tokens.  This limits the the db querying while staying secure as long as tokens life is short (~ 30 min).  I agree, much more work needs to be done to get this going.  Authentication would be handled by a Spree::Api endpoint using devise methods.

PR #1389, opened 2016-08-16
Use in-memory objects in OrderUpdater and remove Adjustment callbacks

jordan-brough:
My primary goal here is to make it easy to integrate 3rd party tax services
in a clean and reliable way by removing extraneous calls to
`ItemAdjustments#update`. (See https://github.com/solidusio/solidus/issues/1252.)

### Use in-memory objects in OrderUpdater:
Operate on `order.line_items` and `order.shipments` inside OrderUpdater
Instead of loading separate copies from the DB. This should give us:

- Better performance
- Return more accurate results. I.e. when calling `order.update!` the
`order.line_items` on that order object will get updated.

The second point is extra important as we work on removing extraneous calls
to `ItemAdjustments` (see below), which previously had the side effect of
updating some in-memory objects that won't otherwise get updated until
a `.reload` is invoked.

### Remove Adjustment callbacks:
Same idea as https://github.com/solidusio/solidus/pull/1356, and likewise related to https://github.com/solidusio/solidus/issues/1252.

Currently we do this:

- Add adjustment to item
  - Trigger `ItemAdjustments#update` via an after_create on Adjustment
- Call `order.update!` on items order
  - Trigger `ItemAdjustments#update` via OrderUpdater#update_totals

We should be able to eliminate the first call to
`ItemAdjustments#update` because we can rely on `order.update!` being
called.  If it weren't called then order-level amounts would be
incorrect.

### DIFF
--- a/core/app/models/spree/adjustment.rb
+++ b/core/app/models/spree/adjustment.rb
@@ -26,9 +26,6 @@ module Spree
     validates :amount, numericality: true
     validates :promotion_code, presence: true, if: :require_promotion_code?
 
-    after_create :update_adjustable_adjustment_total
-    after_destroy :update_adjustable_adjustment_total
-
     scope :not_finalized, -> { where(finalized: false) }
     scope :open, -> do
       Spree::Deprecation.warn "Adjustment.open is deprecated. Instead use Adjustment.not_finalized", caller
@@ -176,11 +173,6 @@ module Spree
     end
 
     private
-
-    def update_adjustable_adjustment_total
-      # Cause adjustable's total to be recalculated
-      ItemAdjustments.new(adjustable).update
-    end
 
     def require_promotion_code?
       promotion? && source.promotion.codes.any?
--- a/core/app/models/spree/order_updater.rb
+++ b/core/app/models/spree/order_updater.rb
@@ -31,7 +31,11 @@ module Spree
     end
 
     def recalculate_adjustments
-      all_adjustments.includes(:adjustable).map(&:adjustable).uniq.each { |adjustable| Spree::ItemAdjustments.new(adjustable).update }
+      adjustables = [*line_items, *shipments, order]
+
+      adjustables.each do |adjustable|
+        Spree::ItemAdjustments.new(adjustable).update
+      end
     end
 
     # Updates the following Order total values:
--- a/core/app/models/spree/order_updater.rb
+++ b/core/app/models/spree/order_updater.rb
@@ -67,7 +67,7 @@ module Spree
     end
 
     def update_shipment_total
-      order.shipment_total = shipments.sum(:cost)
+      order.shipment_total = shipments.to_a.sum(&:cost)
       update_order_total
     end
 
@@ -77,15 +77,14 @@ module Spree
 
     def update_adjustment_total
       recalculate_adjustments
-      order.adjustment_total = line_items.sum(:adjustment_total) +
-      shipments.sum(:adjustment_total) +
-      adjustments.eligible.sum(:amount)
-      order.included_tax_total = line_items.sum(:included_tax_total) + shipments.sum(:included_tax_total)
-      order.additional_tax_total = line_items.sum(:additional_tax_total) + shipments.sum(:additional_tax_total)
 
-      order.promo_total = line_items.sum(:promo_total) +
-      shipments.sum(:promo_total) +
-      adjustments.promotion.eligible.sum(:amount)
+      all_items = line_items + shipments
+
+      order.adjustment_total = all_items.sum(&:adjustment_total) + adjustments.eligible.sum(:amount)
+      order.included_tax_total = all_items.sum(&:included_tax_total)
+      order.additional_tax_total = all_items.sum(&:additional_tax_total)
+
+      order.promo_total = all_items.sum(&:promo_total) + adjustments.promotion.eligible.sum(:amount)
 
       update_order_total
     end
@@ -95,7 +94,7 @@ module Spree
     end
 
     def update_item_total
-      order.item_total = line_items.sum('price * quantity')
+      order.item_total = line_items.to_a.sum(&:amount)
       update_order_total
     end
 
--- a/core/spec/models/spree/order_cancellations_spec.rb
+++ b/core/spec/models/spree/order_cancellations_spec.rb
@@ -152,7 +152,13 @@ describe Spree::OrderCancellations do
       it "generates the correct total amount" do
         order.cancellations.short_ship([inventory_unit_1])
         order.cancellations.short_ship([inventory_unit_2])
-        expect(line_item.adjustments.map(&:amount)).to match_array([0.01, -0.83, -0.84])
+        expect(line_item.adjustments.map(&:amount)).to match_array(
+          [
+            0.01, # tax adjustment
+            -0.84, # short ship 1
+            -0.83, # short ship 2
+          ]
+        )
         expect(line_item.total).to eq 0
       end
     end
--- a/core/spec/models/spree/order_cancellations_spec.rb
+++ b/core/spec/models/spree/order_cancellations_spec.rb
@@ -131,13 +131,12 @@ describe Spree::OrderCancellations do
 
     context "when rounding is required" do
       let(:order) { create(:order_ready_to_ship, line_items_count: 1, line_items_price: 0.83) }
-      let(:line_item) { order.line_items.first }
+      let(:line_item) { order.line_items.to_a.first }
       let(:inventory_unit_1) { line_item.inventory_units[0] }
       let(:inventory_unit_2) { line_item.inventory_units[1] }
 
       before do
         order.contents.add(line_item.variant)
-        line_item.reload
 
         # make the total $1.67 so it divides unevenly
         Spree::Adjustment.tax.create!(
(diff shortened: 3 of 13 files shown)


core/spec/models/spree/order_cancellations_spec.rb
```diff
@@ -152,7 +152,9 @@
 
       it "generates the correct total amount" do
         order.cancellations.short_ship([inventory_unit_1])
+        inventory_unit_2.line_item.reload # UnitCancel#compute_amount needs updated amounts
         order.cancellations.short_ship([inventory_unit_2])
+        line_item.reload
```
jordan-brough: This and the change above it are unfortunate. :(  See the [commit message](https://github.com/solidusio/solidus/pull/1389/commits/164971be3e4c2ac9dc14606af532c6e0408a3345) for details.

cbrunsdon:
I'm totally aware this might break some code as stores might not be expecting us to operate on the in-memory objects, but I'm 100% for the change. I think the order updater should always be operating on the in-memory objects and the more of these changes we make the faster and cleaner our updating process will be.

core/spec/models/spree/order_cancellations_spec.rb
jordan-brough: Here's an idea for getting OrderUpdater to operate on in-memory objects, to try to avoid problems like the above.
```diff
diff --git a/core/app/models/spree/order_updater.rb b/core/app/models/spree/order_updater.rb
index caff2a6..b21a70c 100644
--- a/core/app/models/spree/order_updater.rb
+++ b/core/app/models/spree/order_updater.rb
@@ -31,7 +31,15 @@ module Spree
     end
 
     def recalculate_adjustments
-      all_adjustments.includes(:adjustable).map(&:adjustable).uniq.each { |adjustable| Spree::ItemAdjustments.new(adjustable).update }
+      adjustables = Set.new
+
+      # Prefer using in-memory objects when possible:
+      adjustables << order
+      adjustables.merge(order.line_items)
+      adjustables.merge(order.shipments)
+      # In case someone is attaching adjustments to something other than LineItems and Shipments:
+      adjustables.merge(all_adjustments.includes(:adjustable).map(&:adjustable))
+
+      adjustables.each do |adjustable|
+        Spree::ItemAdjustments.new(adjustable).update
+      end
     end
```
Two risks I can think of:
1) The in-memory objects may be stale.  That might be an acceptable risk in the interest of performance.
2) The above code will invoke `ItemAdjustments#update` on all shipments & line items, even if they don't currently have any adjustments on them. That will result in an extra `select * from spree_adjustments where adjustable_id = ?` for every line item/shipment that doesn't have at least one adjustment (extra compared to the existing code).

Any thoughts?

jhawthorn:
My main concern with this is that the `after_destroy` is currently necessary. I'm worried that when destroying the last adjustment on and item, an `order.update!` won't successfully update the item's total since it won't be included in `all_adjustments.map(&:adjustable)`. The change to `recalculate_adjustments` should fix that, but we should probably make a failing test to be sure.

core/spec/models/spree/order_cancellations_spec.rb
jhawthorn: I really want to do this. I don't think we should concern ourselves with someone attaching adjustments to something other than LineItems and Shipments, since the rest of the methods in `OrderUpdater` won't take that into account.

jordan-brough: @jhawthorn great point.  Here's a PR (against this PR) to do that: https://github.com/jordan-brough/solidus/pull/7  Shall I merge that into this PR?

jhawthorn: > Shall I merge that into this PR?

:+1:

core/spec/models/spree/order_updater_spec.rb
```diff
@@ -56,13 +56,17 @@ module Spree
       end
 
       it "update order adjustments" do
+        # TODO: Make this into a more valid test, so that stubbing
```
jhawthorn: I'm actually encouraged that this test is breaking. It's testing an obviously wrong thing (that the adjustment totals aren't updated despite being wrong).

core/spec/models/spree/order_cancellations_spec.rb
```diff
@@ -155,7 +155,7 @@
         inventory_unit_2.line_item.reload # UnitCancel#compute_amount needs updated amounts
         order.cancellations.short_ship([inventory_unit_2])
         line_item.reload
-        expect(line_item.adjustments.non_tax.sum(:amount)).to eq(-1.67)
+        expect(line_item.adjustments.map(&:amount)).to match_array([0.01, -0.83, -0.84])
```
jhawthorn: `[0.01, -0.83, -0.84].inject(:+) #=> -1.66`

and that 0.01 is the tax adjustment above :+1:

core/spec/models/spree/order_updater_spec.rb
jordan-brough: Great! I've gone ahead and made it simpler but more correct.

core/spec/models/spree/order_cancellations_spec.rb
jordan-brough: 👍

jordan-brough:
FYI I'm a bit worried that I've made the `spec/features/admin/orders/order_details_spec.rb` flaky somehow.  I haven't been able to repro these failures for that spec locally:
https://circleci.com/gh/solidusio/solidus/4156
https://circleci.com/gh/solidusio/solidus/4161
https://circleci.com/gh/solidusio/solidus/4151

core/spec/models/spree/order_cancellations_spec.rb
```diff
@@ -153,7 +152,7 @@
       it "generates the correct total amount" do
         order.cancellations.short_ship([inventory_unit_1])
         order.cancellations.short_ship([inventory_unit_2])
-        expect(line_item.adjustments.non_tax.sum(:amount)).to eq(-1.67)
+        expect(line_item.adjustments.map(&:amount)).to match_array([0.01, -0.83, -0.84])
```
mamhoff: This is a little strange, as the three values now add up to -1.66. Is there a tax adjustment hidden there?

mamhoff: Got it. Don't answer, Line 144-152 explain it all.

mamhoff:
This is good. 👍 !

core/spec/models/spree/order_cancellations_spec.rb
jordan-brough: @mamhoff I went ahead and updated a bit because this spec was hard for me to follow when I first read it and @jhawthorn also left a similar comment as yours.

core/app/models/spree/order_updater.rb
```diff
@@ -77,15 +77,14 @@ def update_order_total
...
+      all_items = line_items + shipments
+
+      order.adjustment_total = all_items.sum(&:adjustment_total) + adjustments.eligible.sum(:amount)
+      order.included_tax_total = all_items.sum(&:included_tax_total)
+      order.additional_tax_total = all_items.sum(&:additional_tax_total)
+
+      order.promo_total = all_items.sum(&:promo_total) + adjustments.promotion.eligible.sum(:amount)
 
       update_order_total
     end
```
mtomov: That's all pretty great! Thank you for it!

Just wanted to ask you on why you didn't go about to change `adjustments.eligible` to an in-memory call as well? I think the collection has already been loaded by now somewhere in the `ItemAdjustments`, and here you're calling a sum twice, so it should be a two SQL lookup save. A `#promotion?` method already exists on Adjustment, only an `#eligible?` one would need be added, which could be useful.

jordan-brough: @jhawthorn any thoughts?

jordan-brough:
Note: I found that the sporadic spec failures are specific to Postgres so I'm guessing it's something to do with DB result ordering but in any case at least I can repro the failures now.

core/app/models/spree/adjustment.rb
```diff
@@ -177,11 +174,6 @@ def update!(target = nil)
...
     private
 
-    def update_adjustable_adjustment_total
-      # Cause adjustable's total to be recalculated
-      ItemAdjustments.new(adjustable).update
-    end
-
     def require_promotion_code?
       promotion? && source.promotion.codes.any?
     end
```
mtomov: Just wanted to point out this bad boy here:

https://github.com/solidusio/solidus/blob/master/core/app/models/concerns/spree/adjustment_source.rb#L7

It sort of relied on the "automatic"  after destroy of adjustments. Should we maybe add an `order.update!' to all those incomplete orders, for which we have deleted an adjustment?

jordan-brough: I'm thinking perhaps we should leave that for a separate PR?  If it was relying on the after destroy then it was leaving the order in an inconsistent state anyway, because it didn't update the order-level totals.

core/app/models/spree/order_updater.rb
jhawthorn: :+1: I really want to do this. I tried it and it caused a few errors, so I'd like to do that as a separate follow-up to this PR

core/app/models/spree/adjustment.rb
mtomov: Sure; it was like that indeed :+1:

jordan-brough: Added a github issue for that: https://github.com/solidusio/solidus/issues/1396

jhawthorn:
:shipit:

bbuchalter:
Great work on moving this forward everyone. Thank you! ❤️

PR #1392, opened 2016-08-17
Fix FreeShipping promos that have codes

jordan-brough:
This commit:
https://github.com/solidusio/solidus/commit/4a62eca#diff-d7fd5424af325ba4827da883ac185ded
broke this fix:
https://github.com/solidusio/solidus/commit/7b50d73
because we stopped looking for promotions with codes.  So if you apply
a free shipping promo to your order via a code and then go backward in the order
state machine (and have your shipments regenerated) you won't get your
promotion re-applied when you move forward in the state machine again.

This also updates FreeShipping promotions to use the recently-added
apply_automatically promotion flag, rather than looking for the
absence of codes and paths.

### DIFF
--- a/core/app/models/spree/promotion_handler/free_shipping.rb
+++ b/core/app/models/spree/promotion_handler/free_shipping.rb
@@ -2,19 +2,21 @@ module Spree
   module PromotionHandler
     # Used for activating promotions with shipping rules
     class FreeShipping
-      attr_reader :order, :order_promo_ids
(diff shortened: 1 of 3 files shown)


jordan-brough:
@jhawthorn I verified just now that the new specs fail on the old code (in the PR I un-`pending`-fied one of the specs) so I think it might be ok?

jhawthorn:
Excellent. Thank you.

PR #1393, opened 2016-08-17
Fix query in Promotion::Actions::FreeShipping

jordan-brough:
The previous query can match the wrong source_type.

I'm thinking this is worth a spec but lmk if someone thinks otherwise.


gmacdougall:
:+1:

mamhoff:
👍

PR #1394, opened 2016-08-17
Remove require_email on email validation

graygilmore:
We only need `require_email` to be true to validate the presence of an email address. If, for whatever reason, you want to add an email address to an order *before* `require_email` is true then it should still go through the appropriate validations to make sure it's a valid email address.

Another option would be for each store to modify `require_email` but there are times where you may want to validate an email address without validating that `presence: true` so I believe this is the preferred method.


jhawthorn:
:+1:

cbrunsdon:
:+1:

PR #1395, opened 2016-08-17
Testing flaky specs

jordan-brough:
Ignore.  I'm just trying to debug some failing flaky specs that I can't repro locally.


jordan-brough:
The sporadic failures only happen on postgres. I can repro locally now.

PR #1400, opened 2016-08-20
Remove more usage of ItemAdjustments outside of OrderUpdater

jordan-brough:
Related to https://github.com/solidusio/solidus/issues/1252. Part of the same effort as https://github.com/solidusio/solidus/pull/1356 and https://github.com/solidusio/solidus/pull/1389.


jhawthorn:
:+1: Simpler and safer just to run the full order updater. I suspect performance is very comparable anyways.

gmacdougall:
:+1: 

Thanks Jordan!

PR #1401, opened 2016-08-24
Use in-memory objects for adjustments in ItemAdjustments

jordan-brough:
See individual commits.

This avoids extra queries and should improve performance.

It also should make refactoring easier without adding extra queries,
including some upcoming tax-refactoring I'm working on.

The two primary commits are these:

- "Automatically update `adjustments` relationships" - a workaround for backward-compatibility
- "Use in-memory objects for adjustments in ItemAdjustments" - to actually start using in-memory objects

I'm not super happy with the above workaround, but I couldn't think of a better option.  Also, it makes me nervous that we need a workaround like that in the first place -- it seems like this could be fragile.  See, for example, the "Don't use all_adjustments in OrderAdjuster" commit.  Using `all_adjustments` in the "wrong" way can result in outdated in-memory objects and thus incorrect calculations.

Relates to https://github.com/solidusio/solidus/issues/1252 and previous work done in https://github.com/solidusio/solidus/pull/1356, https://github.com/solidusio/solidus/pull/1389, and https://github.com/solidusio/solidus/pull/1400.

### DIFF
--- a/core/app/models/spree/tax/order_adjuster.rb
+++ b/core/app/models/spree/tax/order_adjuster.rb
@@ -16,10 +16,6 @@ module Spree
       def adjust!
         return unless order_tax_zone(order)
 
-        [order, *order.line_items, *order.shipments].each do |item|
-          item.adjustments.destroy(item.adjustments.select(&:tax?))
-        end
-
         (order.line_items + order.shipments).each do |item|
           ItemAdjuster.new(item, order_wide_options).adjust!
         end
@@ -32,7 +28,6 @@ module Spree
           rates_for_order_zone: rates_for_order_zone(order),
           rates_for_default_zone: rates_for_default_zone,
           order_tax_zone: order_tax_zone(order),
-          skip_destroy_adjustments: true
         }
       end
     end
--- a/core/app/models/spree/tax/order_adjuster.rb
+++ b/core/app/models/spree/tax/order_adjuster.rb
@@ -17,7 +17,7 @@ module Spree
         return unless order_tax_zone(order)
 
         [order, *order.line_items, *order.shipments].each do |item|
-          item.adjustments.tax.destroy_all
+          item.adjustments.destroy(item.adjustments.select(&:tax?))
         end
 
         (order.line_items + order.shipments).each do |item|
--- a/core/app/models/spree/tax/order_adjuster.rb
+++ b/core/app/models/spree/tax/order_adjuster.rb
@@ -16,7 +16,9 @@ module Spree
       def adjust!
         return unless order_tax_zone(order)
 
-        order.all_adjustments.tax.destroy_all
+        [order, *order.line_items, *order.shipments].each do |item|
+          item.adjustments.tax.destroy_all
+        end
 
         (order.line_items + order.shipments).each do |item|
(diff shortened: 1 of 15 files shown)


core/app/models/spree/tax/order_adjuster.rb
```diff
@@ -16,7 +16,9 @@ def initialize(order)
       def adjust!
         return unless order_tax_zone(order)
 
-        order.all_adjustments.tax.destroy_all
+        [order, *order.line_items, *order.shipments].each do |item|
+          item.adjustments.destroy(item.adjustments.select(&:tax?))
+        end
```
jhawthorn: Instead of this we should probably fix this line https://github.com/solidusio/solidus/blob/master/core/app/models/spree/tax/item_adjuster.rb#L31

and then remove the `skip_destroy_adjustments` option

jordan-brough: @jhawthorn thanks, updated!

mtomov:
Thank you for those :+1: 

Would they actually make it into the 1.x branch, or only onto the Rails5 one?

jhawthorn:
:+1:

core/app/models/spree/unit_cancel.rb
```diff
@@ -6,18 +6,18 @@ class Spree::UnitCancel < Spree::Base
   DEFAULT_REASON = 'Cancel'
 
   belongs_to :inventory_unit
-  has_one :adjustment, as: :source, dependent: :destroy
+  has_many :adjustments, as: :source, dependent: :destroy
```
bbuchalter: Perhaps this change would merit an entry in the CHANGELOG?

core/app/models/spree/unit_cancel.rb
```diff
@@ -6,18 +6,18 @@ class Spree::UnitCancel < Spree::Base
...
   # Creates necessary cancel adjustments for the line item.
   def adjust!
-    raise "Adjustment is already created" if adjustment
+    raise "Adjustment is already created" if adjustments.exists?
 
     amount = compute_amount(inventory_unit.line_item)
 
-    create_adjustment!(
-      adjustable: inventory_unit.line_item,
+    inventory_unit.line_item.adjustments.create!(
```
jhawthorn: Alternatively, I think this could be.

```
self.adjustment = inventory_unit.line_item.adjustments.create!(
```

This should keep both associations fresh without performing an extra query (no query run since source will already be set correctly). This avoids the need to change the `has_one` to a `has_many`

core/app/models/spree/unit_cancel.rb
jordan-brough: @bbuchalter good point.  I updated to @jhawthorn's solution though, so I've reverted back to `has_one`.

core/app/models/spree/unit_cancel.rb
jordan-brough: Updated. Thanks!

jordan-brough:
FYI I've added a Changelog entry for this and the related PRs.

cbrunsdon:
Looks great to me, really fantastic work, thanks :+1: and anyone that contributed eyes.

PR #1405, opened 2016-08-24
Fix mailer previews poor behaviour

adammathys:
ActionMailer::Preview.all will only look for new previews if the [descendents list is empty.](https://github.com/rails/rails/blob/4-2-stable/actionmailer/lib/action_mailer/preview.rb#L57) By manually requiring the previews, we populate it prematurely and cause any app using this extension to be unable to load their own previews in without additional manual effort.

Instead of causing extra grief for applications including Solidus, we can ensure we call ActionMailer::Preview.all before requiring our previews. This will ensure any previews that should be automatically loaded in get picked up and we can continue to manually include ours afterwards.

Unfortunately, this won't fix any issues where an application includes multiple engines attempting to add their own mailer previews.

### DIFF
--- a/core/lib/spree/core.rb
+++ b/core/lib/spree/core.rb
@@ -97,8 +97,4 @@ require 'spree/core/stock_configuration'
 require 'spree/permission_sets'
 require 'spree/deprecation'
 
-require 'spree/mailer_previews/order_preview'
-require 'spree/mailer_previews/carton_preview'
-require 'spree/mailer_previews/reimbursement_preview'
-
 require 'spree/core/price_migrator'
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -119,6 +119,19 @@ module Spree
       initializer "spree.core.checking_migrations", before: :load_config_initializers do |_app|
         Migrations.new(config, engine_name).check
       end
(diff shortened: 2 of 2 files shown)


jhawthorn:
Does it make sense to do this in all environments? When is `ActionMailer::Preview.all` normally called by rails?

adammathys:
It normally isn't called until [`Rails::MailersController#index`](https://github.com/rails/rails/blob/4-2-stable/railties/lib/rails/mailers_controller.rb#L10) is hit.

This isn't really different than how we're loading everything in now. [`ActionMailer::Preview.all`](https://github.com/rails/rails/blob/4-2-stable/actionmailer/lib/action_mailer/preview.rb#L56-L599) simply [requires the dependencies](https://github.com/rails/rails/blob/4-2-stable/actionmailer/lib/action_mailer/preview.rb#L97-L101) the same way we do right now. Personally, I don't think it makes sense to do this outside of the development environment, but decided to leave this as close to how it's currently working as possible while fixing the poor behavior.

jhawthorn:
That code in rails looks a little unfortunate, but this seems like a reasonable fix.

Let's do this only in dev.

:+1:

mamhoff:
👍

PR #407, opened 2015-10-01
FIx a stale data issue when using Spree::OrderCapturing.

Senjai:
As described in the comment, calling `payment.capture!` triggers an
after_save on Spree::Payment. In this after_save, things can happen, one
of those things is an `order.update!`.

That means if the `order.update!` happens within a payment after save, the order we have stored in `@order` is now out of date, and has to be reloaded.

If we don't reload it, the hooks in the order updater will be called
twice. If they make decisions around fresh state changes, idempotency
could be compromised. cc @athal7

PR #1406, opened 2016-08-25
Add `inverse_of` to order.payments

jordan-brough:
This doesn't seem to happen automatically and is more important now
that we're using cached association objects in OrderUpdater (see
#1389).  Solidus 1.4 broke some functionality in our app without it.  [This](https://github.com/solidusio/solidus/blob/v1.4.0.beta1/core/app/models/spree/payment.rb#L27-L28)
`after_save` in Payment triggers `order.update!` if the Order or the
Payment is complete.  That update needs to operate on the in-memory order
object or things can get out of sync.

It seems like this would be worth getting into Solidus 1.4.

In a Solidus sandbox without this patch:

    >> o = Spree::Order.create!
    >> p = o.payments.create!
    >> puts p.order.object_id, o.object_id
    70240633410460
    70240634674300

With this patch:

    >> o = Spree::Order.create!
    >> p = o.payments.create!
    >> puts p.order.object_id, o.object_id
    70145793491740
    70145793491740

There is some important history around this particular `inverse_of` --
see #407 and spree/spree_gateway#132.

Commit d7b8c73 removed part of the specs around this,
assuming that the `inverse_of` was automatically added, but it doesn't
actually seem to be (see above). The remaining tests around the issue
still pass.

I think I need to look further into the details of the prior issues to ensure
that we won't bring back bugs with this.  Any extra eyes or thoughts are
appreciated.  Any ideas on specs to add with this are also welcome.

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -47,7 +47,7 @@ module Spree
     belongs_to :store, class_name: 'Spree::Store'
     has_many :state_changes, as: :stateful
     has_many :line_items, -> { order(:created_at, :id) }, dependent: :destroy, inverse_of: :order
-    has_many :payments, dependent: :destroy
+    has_many :payments, dependent: :destroy, inverse_of: :order
     has_many :return_authorizations, dependent: :destroy, inverse_of: :order
     has_many :reimbursements, inverse_of: :order
     has_many :adjustments, -> { order(:created_at) }, as: :adjustable, inverse_of: :adjustable, dependent: :destroy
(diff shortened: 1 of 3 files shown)


jhawthorn:
:+1: it's time to do this.

We probably can also drop the `order.reload` added in #407, as I believe that was required due to rails 4.2's awkward behaviour for `changed?` when `update_columns` was being used.

cbrunsdon:
I'm good on this too, thanks jordan :+1:

jordan-brough:
I've gone ahead and removed the reload mentioned by @jhawthorn and added some more comments on the specs for that.  Assuming specs go green I'll merge this next week.

Shall we backport this to 1.4 as well?

bbuchalter:
For those who are investigating this issue be sure to also look at what impact this PR had by reviewing https://github.com/solidusio/solidus/pull/1416

PR #1407, opened 2016-08-25
API: Add Pagination Helpers + Extension Point

stewart:
This PR provides a configuration point for API response pagination at the controller, through a `default_per_page` private method.

This can be used to determine how many instances of a resource will be shown on a given "page" of API responses, and by default will return Kaminari's default count of 25.

Extracting the pagination controller method concern to a discrete method also provides a further extension point, for stores that might prefer to skip out on pagination entirely, or use their own solution besides Kaminari.

### DIFF
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -161,6 +161,16 @@ module Spree
           status: 422
         )
       end
+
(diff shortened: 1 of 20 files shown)


jhawthorn:
lgtm :+1:

mamhoff:
👍

PR #1408, opened 2016-08-25
Fetch all initial taxons for $.fn.taxonAutocomplete

stewart:
In cases where more than 25 taxons were already selected, this
autocomplete selector would only display the first 25, due to the
default pagination settings as defined by Kaminari.

By manually specifying a `per_page` parameter, we can ensure all
specified taxons are returned.


gmacdougall:
Thanks Andrew!

👍

jhawthorn:
:+1:

I could also see it making sense for the api not to use pagination when `ids=` is specified.

mamhoff:
👍

PR #1411, opened 2016-08-30
Admin logo easy config

gevann:
This PR adds a max-width to the admin-nav-header's logo, enabling admins to have their custom logos (set in the spree initializer) fit the nav-bar by default.

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
@@ -14,6 +14,9 @@ body {
 .admin-nav {
   @include position(absolute, 0 null 0 0);
   width: $width-sidebar;
+  img {
+    max-width: 100%;
+  }
 }
 
 .content-wrapper {


backend/spec/features/admin/configuration/logo_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    visit spree.admin_path
+    click_link "Settings"
+  end
+
+  context "visiting the main admin page" do
+    it "sets the ID on the header's logo" do
+      expect(page).to have_selector("img#admin-nav-logo", count: 1)
+    end
+  end
+end
```
mtomov: Having `js:true` over there makes this test very slow, compared to the benefit of testing that an ID is present, I'd think that's a negative gain of having the test. So, I'd rather choose not to have it.

mamhoff: I don't even see why `js: true` is there, and suspect the spec would pass without it. @gevann do you mind removing that bit?

gevann: Will do

jhawthorn:
I think this could be accomplished much simpler.

Instead of `max-width: $width-sidebar - (2 * $sidebar-margin);` we can just use `max-width: 100%`.

Rather than giving the image an `id` and styling that, it's probably reasonable to just add the rule on `.admin-nav img { max-width: 100%; }`

I agree that this change should need no specs.

gevann:
Simple approach used, and specs removed.

jhawthorn:
:+1:

backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -14,6 +14,9 @@ body {
 .admin-nav {
   @include position(absolute, 0 null 0 0);
   width: $width-sidebar;
+  img {
+    max-width: 100%;
+  }
 }
```
mtomov: I'd also use `> img` if possible to avoid the descendant selector for performance reasons.

https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Writing_efficient_CSS#Avoid_the_descendant_selector.21

jhawthorn: I don't think this is a concern. We use descendant selectors liberally in our styles. A `> img` here also won't achieve the desired effect here.

From the linked document

> Note: This document was originally written in 2000. Much has changed when it comes to writing CSS that is fast. This guide is not an accurate representation of the bottlenecks in the browser's rendering pipeline.

mtomov: Indeed rightly pointed. Thanks for the reply! It's also great that you have established a style.

PR #1413, opened 2016-09-01
Add missing translation

jordan-brough:
For Spree::Stock::InventoryValidator. I noticed this was missing while
doing an audit.

It does seem a bit strange because the "item" text won't be translated
for sites that serve multiple languages:

```ruby
line_item.errors[:inventory] << Spree.t(
  :inventory_not_available,
  item: line_item.variant.name
)
```
(https://github.com/solidusio/solidus/blob/v1.3.1/core/app/models/spree/stock/inventory_validator.rb#L6-L9)


gmacdougall:
:+1: Thanks Jordan!

mamhoff:
👍

jhawthorn:
:+1:

PR #1415, opened 2016-09-02
Remove callbacks from refund model

swively:
This PR removes two `after_create` callbacks from the refund model; refunds may now be created without automatically performing the transaction. 

The instance method `#perform_and_save!` was created to perform the transaction, and will raise an error if the transaction fails due to the validation of `transaction_id` on `update`
- `perform_and_save!` method is explicitly called from `reimbursement_helpers.rb`
- `create_log_entry` is called during the transaction

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Extract stock transfers to the [solidus_stock_transfers](https://github.com/solidusio-contrib/solidus_stock_transfers) gem. [#2379](https://github.com/solidusio/solidus/pull/2379) ([jhawthorn](https://github.com/jhawthorn))
 
+- Remove `#perform!` callback from refund model. Refunds can be created without performing the transaction implicitly; `#perform!` must be called for the transaction to be processed. [#1415](https://github.com/solidusio/solidus/pull/1415) ([swively](https:/github.com/swively))
+
 ## Solidus 2.4.0 (unreleased)
 
 ### Major changes
--- a/backend/app/controllers/spree/admin/refunds_controller.rb
+++ b/backend/app/controllers/spree/admin/refunds_controller.rb
@@ -8,6 +8,21 @@ module Spree
 
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
+      def create
+        @refund.attributes = refund_params
+        if @refund.perform!
+          flash[:success] = flash_message_for(@refund, :successfully_created)
+          respond_with(@refund) do |format|
+            format.html { redirect_to location_after_save }
+          end
+        else
+          flash.now[:error] = @refund.errors.full_messages.join(", ")
+          respond_with(@refund) do |format|
+            format.html { render action: 'new' }
+          end
+        end
+      end
+
       private
 
       def location_after_save
@@ -21,6 +36,10 @@ module Spree
 
       def refund_reasons
         @refund_reasons ||= Spree::RefundReason.active.all
+      end
+
+      def refund_params
+        params.require(:refund).permit!
       end
 
       def build_resource
--- a/backend/spec/controllers/spree/admin/refunds_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/refunds_controller_spec.rb
@@ -4,22 +4,46 @@ describe Spree::Admin::RefundsController do
   stub_authorization!
 
   describe "POST create" do
-    context "a Spree::Core::GatewayError is raised" do
-      let(:payment) { create(:payment) }
+    let(:refund_reason) { create(:refund_reason) }
+    let(:refund_amount) { 100.0 }
 
-      subject do
-        post :create,
-          params: {
-          refund: { amount: "50.0", refund_reason_id: "1" },
+    let(:payment) { create(:payment, amount: payment_amount) }
+    let(:payment_amount) { refund_amount * 2 }
+
+    subject do
+      post :create,
+        params: {
+          refund: {
+            amount: refund_amount,
+            refund_reason_id: refund_reason.id,
+            transaction_id: nil
+          },
           order_id: payment.order_id,
           payment_id: payment.id
         }
+    end
+
+    context "and no Spree::Core::GatewayError is raised" do
+      it "creates a refund record" do
+        expect{ subject }.to change(Spree::Refund, :count).by(1)
       end
 
-      before(:each) do
-        def controller.create
-          raise Spree::Core::GatewayError.new('An error has occurred')
-        end
+      it "calls #perform!" do
+        subject
+        # transaction_id comes from Spree::Gateway::Bogus.credit
+        expect(Spree::Refund.last.transaction_id).to eq("12345")
+      end
+    end
+
+    context "a Spree::Core::GatewayError is raised" do
+      before do
+        expect_any_instance_of(Spree::Refund).
+          to receive(:perform!).
+          and_raise(Spree::Core::GatewayError.new('An error has occurred'))
+      end
+
+      it "does not create a refund record" do
+        expect{ subject }.to_not change { Spree::Refund.count }
       end
 
       it "sets an error message with the correct text" do
--- a/core/app/models/spree/refund.rb
+++ b/core/app/models/spree/refund.rb
@@ -8,13 +8,9 @@ module Spree
 
     validates :payment, presence: true
     validates :reason, presence: true
-    validates :transaction_id, presence: true, on: :update # can't require this on create because the before_create needs to run first
     validates :amount, presence: true, numericality: { greater_than: 0 }
 
     validate :amount_is_less_than_or_equal_to_allowed_amount, on: :create
-
-    after_create :perform!
-    after_create :create_log_entry
 
     scope :non_reimbursement, -> { where(reimbursement_id: nil) }
 
@@ -35,21 +31,23 @@ module Spree
       payment.payment_method.name
     end
 
-    private
-
-    # attempts to perform the refund.
+    # Must be called for the refund transaction to be processed.
+    #
+    # Attempts to perform the refund,
     # raises an error if the refund fails.
     def perform!
       return true if transaction_id.present?
 
       credit_cents = money.cents
 
-      @response = process!(credit_cents)
+      response = process!(credit_cents)
+      log_entries.build(details: response.to_yaml)
 
-      self.transaction_id = @response.authorization
-      update_columns(transaction_id: transaction_id)
+      update!(transaction_id: response.authorization)
       update_order
     end
+
+    private
 
     # return an activemerchant response object if successful or else raise an error
     def process!(credit_cents)
@@ -69,10 +67,6 @@ module Spree
     rescue ActiveMerchant::ConnectionError => e
       logger.error(I18n.t('spree.gateway_error') + "  #{e.inspect}")
       raise Core::GatewayError.new(I18n.t('spree.unable_to_connect_to_gateway'))
-    end
-
-    def create_log_entry
-      log_entries.create!(details: @response.to_yaml)
     end
 
     def amount_is_less_than_or_equal_to_allowed_amount
--- a/core/app/models/spree/reimbursement_type/reimbursement_helpers.rb
+++ b/core/app/models/spree/reimbursement_type/reimbursement_helpers.rb
@@ -35,7 +35,7 @@ module Spree
         reason: Spree::RefundReason.return_processing_reason
       })
 
-      simulate ? refund.readonly! : refund.save!
+      simulate ? refund.readonly! : refund.perform!
       refund
     end
 
--- a/core/app/models/spree/payment/cancellation.rb
+++ b/core/app/models/spree/payment/cancellation.rb
@@ -29,7 +29,7 @@ module Spree
           if response = payment.payment_method.try_void(payment)
             payment.send(:handle_void_response, response)
           else
-            payment.refunds.create!(amount: payment.credit_allowed, reason: refund_reason)
+            payment.refunds.create!(amount: payment.credit_allowed, reason: refund_reason).perform!
           end
         else
           # For payment methods not yet implemeting `try_void`
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -69,7 +69,7 @@ RSpec.describe Spree::Order, type: :model do
       let(:payment) { create(:payment, order: order, amount: payment_amount, state: 'completed') }
 
       before do
-        create(:refund, payment: payment, amount: payment_amount)
+        create(:refund, payment: payment, amount: payment_amount).perform!
       end
 
       it "cancels the order" do
@@ -229,7 +229,7 @@ RSpec.describe Spree::Order, type: :model do
       order.cancellations.short_ship([order.inventory_units.first])
       expect(order.outstanding_balance).to be_negative
       expect(order.payment_state).to eq('credit_owed')
-      create(:refund, amount: order.outstanding_balance.abs, payment: payment, transaction_id: nil)
+      create(:refund, amount: order.outstanding_balance.abs, payment: payment, transaction_id: nil).perform!
       order.reload
       expect(order.outstanding_balance).to eq(0)
       expect(order.payment_state).to eq('paid')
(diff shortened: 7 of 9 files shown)


tvdeyen:
I'm fine with this, although I'm surprised that only in one place a refund is actually performed. Surely a reason why callbacks are evil here and this change is the right thing to do. 

We definitely need a Changelog entry. There will be people relying on the refund performing itself. 

:+1:

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Solidus 2.0.1 (unreleased)
```
jhawthorn: This should be for version 2.1.0

cbrunsdon:
@swively you'll also need to update the RefundController in the backend to perform! as that is what happens now.

core/app/models/spree/refund.rb
```diff
@@ -43,12 +38,15 @@ def perform!
       credit_cents = Spree::Money.new(amount.to_f, currency: payment.currency).money.cents
 
       @response = process!(credit_cents)
+      log_entries.create!(details: @response.to_yaml)
```
jhawthorn: This change is great, and allows another refactor here.

We can change this ugly instance variable to just a local variable in this `perform!` method (the only place it is used)

backend/app/controllers/spree/admin/refunds_controller.rb
```diff
@@ -8,6 +8,12 @@ class RefundsController < ResourceController
 
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
+      def create
+        super.tap do
```
jhawthorn: This is probably not quite right. The return value of a controller action has no meaning, so we don't need to return the value from our superclass here.

However, the `save` performed in the `ResourceController`, our superclass, may or may not succeed. We don't want to perform a refund which failed to save.

We probably need to duplicate the logic from the ResourceController here, but inserting the `@refund.perform!` in the "success" path.

backend/app/controllers/spree/admin/refunds_controller.rb
```diff
@@ -8,6 +8,26 @@ class RefundsController < ResourceController
...
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
+      def create
+        @refund.attributes = refund_params
+        if @refund.save
+          @refund.perform!
+          flash[:success] = flash_message_for(@refund, :successfully_created)
+          respond_with(@refund) do |format|
+            format.html { redirect_to location_after_save }
+            format.js   { render layout: false }
```
jhawthorn: We don't have a js view here, so we can remove this line. (it made sense in ResourceController because some controllers would have a JS view for create)

backend/app/controllers/spree/admin/refunds_controller.rb
```diff
@@ -8,6 +8,26 @@ class RefundsController < ResourceController
...
+          end
+        else
+          respond_with(@refund) do |format|
+            format.html do
+              flash.now[:error] = @refund.errors.full_messages.join(", ")
+              render action: 'new'
+            end
+            format.js { render layout: false }
```
jhawthorn: Same as above

backend/app/controllers/spree/admin/refunds_controller.rb
```diff
@@ -8,6 +8,22 @@ class RefundsController < ResourceController
 
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
+      def create
```
jordan-brough: Instead of defining our own `create` could we take advantage of `Spree::Backend::Callbacks` and do:
```ruby
create.after :perform_refund
...
def perform_refund
  @refund.perform!
end
```
?

backend/app/controllers/spree/admin/refunds_controller.rb
```diff
@@ -8,6 +8,22 @@ class RefundsController < ResourceController
...
+        @refund.attributes = refund_params
+        if @refund.save
+          @refund.perform!
+          flash[:success] = flash_message_for(@refund, :successfully_created)
+          respond_with(@refund) do |format|
+            format.html { redirect_to location_after_save }
+          end
+        else
+          flash[:error] = @refund.errors.full_messages.join(", ")
```
jordan-brough: This needs to be `flash.now[:error]` right? (Though if we use the `RefundController#create` this goes away anyway.

core/app/models/spree/refund.rb
```diff
@@ -33,22 +30,23 @@ def description
...
       return true if transaction_id.present?
 
       credit_cents = Spree::Money.new(amount.to_f, currency: payment.currency).money.cents
 
-      @response = process!(credit_cents)
+      response = process!(credit_cents)
+      log_entries.create!(details: response.to_yaml)
 
-      self.transaction_id = @response.authorization
+      self.transaction_id = response.authorization
```
jordan-brough: Now that this isn't invoked via a callback we can just do `update!(transaction_id: response.authorization)` right? (And remove the following `update_columns` line)

jordan-brough:
Something I just thought about -- after this change it's possible to have refunds in your database that haven't actually been refunded.  E.g.:

```ruby
refund = Spree::Refund.create!(...)
refund.perform! # <= raises an exceptoin
```

Previously the return of money happened in an after_create and if it failed (raised an exception) then the refund insert would get rolled back.  With this PR right now the `Refund.total_amount_reimbursed_for` method could easily become incorrect.

It seems like we need to either 1) Make it obvious that a Refund record should only exist if the return of money was successful, and make it easy to do that in a robust way. Or 2) Add something like a `state` column to `Refund` so that it can be marked as failed.

A solution for option 1 could be:

```ruby
def self.perform!(attributes)
  transaction do
    create!(attributes)
    perform!
  end
end
```

Then the "right way" would be fairly obvious and easy to do in a robust way.  And if someone *really* wants to create a `Refund` record without performing the refund (e.g. while importing data) they can still do so easily.

If we want to do option 2 then there is more code we need to change and probably a bigger release note we need to write.

backend/app/controllers/spree/admin/refunds_controller.rb
swively: @jhawthorn  and @cbrunsdon  expressed negativity towards `Spree::Backend::Callbacks`, so I've avoided using one in the controller in this case

core/app/models/spree/refund.rb
swively: Thank you for the feedback, Jordan! Forgive me, I can only reply to comments on the code.

You are correct that this change would potentially break `Refund.total_amount_reimbursed_for`,
so I have added a scope to make sure the method only performs on transacted refunds. 

The goal of this PR is to remove the implicit call of `perform!` so that refunds could be created
without automatically performing them, for the purposes of data importing

backend/app/controllers/spree/admin/refunds_controller.rb
jordan-brough: @jhawthorn @cbrunsdon ?  Are you guys wanting to deprecate `Spree::Backend::Callbacks`?  This seems like the kind of usage they were intended for.

jordan-brough:
> I have added a scope to make sure the method only performs on transacted refunds

This seems too indirect to me.  If we're going to introduce "state" to the Refund model I think we should make it more explicit than that (even if we don't use state_machine (preferrably not :)).

> The goal of this PR is to remove the implicit call of perform! so that refunds could be created
without automatically performing them, for the purposes of data importing

That makes complete sense.  But it seems like this:
```ruby
class Spree::Refund
  def self.perform!(attributes)
    transaction do
      create!(attributes)
      perform!
    end
  end
  ...
end
```
Would accomplish that in a much less invasive way (without changing the the meaning of what the presence of a `Refund` means)?

core/app/models/spree/refund.rb
swively: I believe you have to be a member of solidusio, regardless of whether you authored the PR

backend/app/controllers/spree/admin/refunds_controller.rb
cbrunsdon: I don't really like them, but it might be better than what we're doing here? @jhawthorn I'll leave you as tie breaking vote.

core/app/models/spree/refund.rb
swively: Sorry Jordan, I definitely agree with what you've suggested. 

Had to augment your method suggestion slightly, to be more verbose and to return the refund.

CHANGELOG.md
```diff
@@ -22,6 +22,10 @@
 
       https://github.com/solidusio/solidus/pull/1461
 
+    * Removed `#perform!` callback from refund model, it must be called manually
```
jordan-brough: I'd suggest wording this more strongly somehow.  Refunds really should always be performed via `create_and_perform!` unless you are taking care to delete the refund if it fails (via a transaction or otherwise).  And if you're importing data you won't be "performing" the refund.  Right?

core/app/models/spree/reimbursement_type/reimbursement_helpers.rb
```diff
@@ -35,7 +35,13 @@ def create_refund(reimbursement, payment, amount, simulate)
         reason: Spree::RefundReason.return_processing_reason
       })
 
-      simulate ? refund.readonly! : refund.save!
+      if simulate
+        refund.readonly!
+      else
+        refund.save!
+        refund.perform!
```
jordan-brough: We should use `create_and_perform!` here right?  Otherwise a failed `perform!` could leave an invalid refund in the DB.

core/app/models/spree/refund.rb
jordan-brough: That sounds good to me, thanks!

backend/app/controllers/spree/admin/refunds_controller.rb
jordan-brough: Hm, I think we actually can't use `create.after` anymore because we need to use `create_and_perform` instead of `save` + `perform!`.

backend/app/controllers/spree/admin/refunds_controller.rb
```diff
@@ -8,6 +8,22 @@ class RefundsController < ResourceController
 
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
+      def create
+        @refund.attributes = refund_params
+        if @refund.save
+          @refund.perform!
```
jordan-brough: We should use `create_and_perform` here also right?

swively: @jhawthorn suggested going a different route and making a `save_and_perform!` method that I'll get into next week

swively: @jordan-brough 
Changed to `#perform_and_save!` as per suggestions from @jhawthorn and @cbrunsdon 

Hopefully this appeases all parties!

core/app/models/spree/refund.rb
```diff
@@ -29,26 +26,32 @@ def total_amount_reimbursed_for(reimbursement)
       end
     end
 
+    def perform_and_save!
```
jordan-brough: Can we add some RDoc to this method, letting people know that they should always use this to save a refund unless they are importing data? (Until we do something like add state column and perhaps can get rid of this method entirely)

core/app/models/spree/refund.rb
```diff
@@ -29,26 +26,32 @@ def total_amount_reimbursed_for(reimbursement)
...
+      perform!
+      save!
+      self
+    end
+
     def description
       payment.payment_method.name
     end
 
-    private
```
jordan-brough: Can we restore the `private` here please? Since everyone should call this via `perform_and_save!`.

core/app/models/spree/refund.rb
```diff
@@ -29,26 +26,32 @@ def total_amount_reimbursed_for(reimbursement)
       end
     end
 
+    def perform_and_save!
+      perform!
+      save!
```
jordan-brough: `perform!` calls `update!`, so we don't need the `save!` here, right?

With the current state of this PR we could possibly even get rid of `perform_and_save!` and just have `perform!`.  But for now (until we add a `state` column) I personally think it's worth having `perform_and_save!`.

swively: At that point all `perform_and_save!` would do is call `perform!`, as it really doesn't need to return `self` either.

I think we should stick with the first iteration of having to call the public method `perform!` on a refund to perform it, and rely on the Changelog or the RDoc on the method to inform users of the need to explicitly perform their refunds.

backend/app/controllers/spree/admin/refunds_controller.rb
```diff
@@ -23,6 +38,10 @@ def refund_reasons
         @refund_reasons ||= RefundReason.active.all
       end
 
+      def refund_params
+        params.require(:refund).permit!
```
jordan-brough: Any reason not to use a whitelist here, rather than permitting everything?

core/app/models/spree/refund.rb
```diff
@@ -8,14 +8,11 @@ class Refund < Spree::Base
 
     validates :payment, presence: true
     validates :reason, presence: true
-    validates :transaction_id, presence: true, on: :update # can't require this on create because the before_create needs to run first
+    validates :transaction_id, presence: true, on: :update # can't require this on create, #perform! provides the transaction_id
```
jordan-brough: We should either remove the `on:` condition here or have a more robust way of determining whether a refund needs to have a transaction id.  With the updates from this PR we no longer need the `on:` for Solidus usage since the transaction id is fetched before the record is validated.

I'd be inclined to just remove the `on:` condition and require that things like data imports supply a manufactured id of some sort, if they don't have one already.

backend/spec/controllers/spree/admin/refunds_controller_spec.rb
```diff
@@ -4,18 +4,36 @@
...
+      expect{ subject }.to change(Spree::Refund, :count).by(1)
+    end
+
+    it "calls #perform!" do
+      subject
+      # transaction_id comes from Spree::Gateway::Bogus.credit
+      expect(Spree::Refund.last.transaction_id).to eq("12345")
+    end
+
+    context "a Spree::Core::GatewayError is raised" do
```
jordan-brough: Could we please also check here that no `Refund` object is created when a failure occurs?

backend/spec/controllers/spree/admin/refunds_controller_spec.rb
```diff
@@ -4,18 +4,36 @@
...
       before(:each) do
         def controller.create
           raise Spree::Core::GatewayError.new('An error has occurred')
```
jordan-brough: Could we change this to something like `expect_any_instance_of(Spree::Refund).to receive(:perform!).and_raise(...)`?  So that a test around not creating a Refund is a bit more likely to be meaningful?

backend/app/controllers/spree/admin/refunds_controller.rb
jhawthorn: This is done throughout the admin

core/app/models/spree/refund.rb
swively: I think opting to remove this validation might be the best route. It allows us to `create` refunds before calling `perform!` on them, rather than building them without saving to the database. It also allows data imports of refunds without `transaction_id`s, which is a potentital scenario. 

The error handling within the `#perform!` call should be enough to alert users that a refund transaction was unsuccessful, as a `transaction_id` is required for the method to succeed.

swively: Thank you for all your input on these changes Jordan, you've helped shed a lot of light on potential vulnerabilities with a change such as this. I know adding state to the refund model is something that not everyone will agree with.
Ultimately for the average user these changes won't have any negative implications, but for some users they will have more power over how their store handles refunds.

seand7565:
It's been over 2 years since this PR has been worked on - I think we should close it. If this is still a change we want to make, can we create an issue for it instead?

jarednorman:
I agree. I like the idea of this change, but @swively won't likely be completing this work.

aldesantis:
I will go ahead and close this one, but @swively feel free to reopen it!

Also, @seand7565 and @jarednorman thanks for the awesome triaging work you've been doing lately.

kennyadsl:
We should integrate this work with https://github.com/solidusio/solidus/pull/3181 before `v2.11` release.

kennyadsl:
Finally closed in favor of https://github.com/solidusio/solidus/pull/3641

PR #1416, opened 2016-09-02
Fix incorrect payment total calculation in process_payments_with

jhawthorn:
In 11de63f9d8956b3b0fad21833993b413413c22ea we added the inverse_of relation between order and payments, which caused some issues in `process_payments_with` from `Order::Payments`.

Payment has an `after_save` which triggers an `order.update!`, which updates the payment_total. process_payments_with was doing `self.payment_total += payment.amount` on each successful payment, making each payment amount counted twice.

This removes the addition in `process_payments_with` and relies on the callback performing the order update (which will update the payment total properly).

This commit also removes stubbing from a specs which now exposes this issue.

### DIFF
--- a/core/app/models/spree/order/payments.rb
+++ b/core/app/models/spree/order/payments.rb
@@ -43,10 +43,6 @@ module Spree
           break if payment_total >= total
 
           payment.public_send(method)
-
-          if payment.completed?
-            self.payment_total += payment.amount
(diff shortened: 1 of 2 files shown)


cbrunsdon:
:+1:

stewart:
:+1:

Issue #1366, opened 2016-08-02
Stock::Coordinator incorrectly splits packages


adaddeo:
On a 1.3 sandbox store:
1. In the admin, create a new order
2. Add 1 unit of any item
3. Update the quantity of item to 2
4. Enter customer details (anything will work) and update
5. Two shipments are created, each with 1 item

jhawthorn:
Confirmed. Looks like it's realated to an `OrderStockLocation` record being careted, which are used to configure packages specifically for one location.

DanielePalombo:
@jhawthorn i'm working on this issue, I'm almost done

DanielePalombo:
@adaddeo Have you try if my PR resolve your issue?

jhawthorn:
This should be fixed by #1709, which bypasses the "location configured package" feature, which is what caused the separate shipments.

PR #1417, opened 2016-09-04
Incorreclty splits packages

DanielePalombo:
rel #1366 

> On sandbox store:
> 1. In the admin, create a new order
> 2. Add 1 unit of any item
> 3. Update the quantity of item(bug 1) or re-add the same item(bug 2)
> 4. Enter customer details and click update button
> 5. Two shipments are created, the first with the first quantity added and the second with the remained quantity item

After investigation i see that the `order_stock_location` isn't update when `update_cart` is call (https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_contents.rb#L37-L67), and the `order_stock_location` creation is skipped if line_item isn't a new record(https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_contents.rb#L146).

solution to bug 1:
Update the quantity of `order_stock_location` when the `update_cart` method is called.
solution to bug 2:
Update the quantity of `order_stock_location` when an item is add to cart if already present an other `variant` and `stock_location` in the cart

### DIFF
--- a/api/app/controllers/spree/api/line_items_controller.rb
+++ b/api/app/controllers/spree/api/line_items_controller.rb
@@ -30,7 +30,14 @@ module Spree
 
       def update
         @line_item = find_line_item
-        if @order.contents.update_cart(line_items_attributes)
+        options = {
+          variant_stock_location_quantities: {
+            @line_item.variant_id => {
+              @line_item.variant.stock_locations.first.id => line_items_attributes[:line_items_attributes][:quantity]
+            }
+          }
+        }
+        if @order.contents.update_cart(line_items_attributes, options)
           @line_item.reload
           respond_with(@line_item, default_template: :show)
         else
--- a/core/app/models/spree/order_contents.rb
+++ b/core/app/models/spree/order_contents.rb
@@ -34,7 +34,7 @@ module Spree
       after_add_or_remove(line_item, options)
     end
 
-    def update_cart(params)
+    def update_cart(params, options = {})
       # We need old_tax_address / new_tax_address because we can't rely on methods
       # offered by ActiveRecord::Dirty to determine if tax_address was updated
       # because if we update the address, a new record will be created
@@ -43,6 +43,7 @@ module Spree
       old_tax_address = order.tax_address
 
       if order.update_attributes(params)
+        update_variant_stock_locations(order.line_items, options[:variant_stock_location_quantities])
 
         new_tax_address = order.tax_address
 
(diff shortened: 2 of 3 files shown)


api/app/controllers/spree/api/line_items_controller.rb
```diff
@@ -30,7 +30,14 @@ def create
 
       def update
         @line_item = find_line_item
-        if @order.contents.update_cart(line_items_attributes)
+        options = {
+          variant_stock_location_quantities: {
+            @line_item.variant_id => {
+              @line_item.variant.stock_locations.first.id => 3
```
cbrunsdon: is this right? should it be first.id: 3?

DanielePalombo: Thanks, i push the changes

jhawthorn:
This is tricky

Traditionally stock locations aren't associated until the packaging step. This allows for some "intelligent" packaging and allows stock to be drawn from whichever stock location has inventory available. This is the most common path that any frontend store will do, and is probably what out backend should do most of the time.

This changed with the introduction of the OrderStockLocation (which was not a particularly complete implementation IMNSHO). An OrderStockLocation ties some amount of a variant to a particular stock location. When packaging, it [builds these as separate shipments](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/stock/coordinator.rb#L41-L51).

We run into issues because under different circumstances our banckend behaves differently w/r/t adding an `OrderStockLocation`. The cart page usually does, and the shipments page doesn't.

I'm don't think adding an `OrderStockLocation` for the first stock location is the correct fix here, as it's possible that that location runs out of stock or could have other issues. I think we need to clarify to admin users when these records are being created, when they exist, and possibly stop creating them.

DanielePalombo:
Hi @jhawthorn  ,

i understand what do you mean, i agree with you probably it's better  clarify to the admin user the stock status and how the stock items in the order are splitted.

I propose to change the update quantity logic and give the possibility to change the quantity for each stock item. I drew a mockup to show you how i want implement this.

In order cart page with one item added in order, the user click on `edit` for change the quantity.
screenshot from 2016-10-07 14-37-36

The admin show the `order_stock_locations` details.
screenshot from 2016-10-07 16-04-06

The admin can change specific quantity for each `order_stock_locations` . When done click apply button to update the quantities.
screenshot from 2016-10-07 15-42-13

What do you think about it?

peterberkenbosch:
@DanielePalombo this needs a rebase with master.

DanielePalombo:
Before rebase or other changes, I'm waiting to know if my proposed solution can be alright.

peterberkenbosch:
Was about to mention that Daniele after I read the rest of this thread! 👍

jhawthorn:
@DanielePalombo sorry for the delay.

In addition to being assigned to a specific stock location, the default is that the variant can come from any stock location. IMO whatever UI we come up with should favour this default, since most of the that's probably what an admin would want.

DanielePalombo:
Sorry i don't understand what do you mean. 
Could you explaine more deeply please?

jhawthorn:
This should be fixed by #1696, which bypasses the "location configured package" feature, which is what caused the separate shipments

PR #1419, opened 2016-09-06
Don't nullify adjustment source association from completed orders

jordan-brough:
Stop setting spree_adjustments.source_id to null on completed orders
when deleting an adjustment source.

I don't think this shouldn't be necessary since in either case `adjustment.source`
will be nil. Also, the two sources using this within Solidus itself
(PromotionAction and TaxRate) both use `acts_as_paranoid` so deleting
the source_id deletes data that could still be useful because the
record is still actually in the database.

### DIFF
--- a/core/app/models/concerns/spree/adjustment_source.rb
+++ b/core/app/models/concerns/spree/adjustment_source.rb
@@ -1,19 +1,25 @@
 module Spree
   module AdjustmentSource
     def deals_with_adjustments_for_deleted_source
-      adjustment_scope = adjustments.joins(:order)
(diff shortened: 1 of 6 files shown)


jordan-brough:
@solidusio/admins any thoughts?

An alternative might be to avoid this entirely by raising an exception when trying to delete an adjustment source that is referenced by completed orders.  (A poor man's foreign key for a polymorphic relationship).  That would be more involved though.  This seems like a decent incremental improvement?

jhawthorn:
I think we need might want a `dependent: restrict` (or equivalent) to catch attempts to delete adjustment sources when it is not safe.

cbrunsdon:
Its been Agreed that since these are paranoid, we're better to keep the sources around.

gmacdougall:
Superseded by #2259

PR #1420, opened 2016-09-07
Add inverse_of to promotion actions and rules

jordan-brough:
This isn't happening automatically.

Without this patch:

    >> p = Spree::Promotion.create!(name: 'foo')
    >> p.promotion_rules << Spree::Promotion::Rules::FirstOrder.new
    >> p.promotion_actions << Spree::Promotion::Actions::FreeShipping.new

    >> p.object_id == p.promotion_rules.to_a.first.promotion.object_id
    => false

    >> p.object_id == p.promotion_actions.to_a.first.promotion.object_id
    => false

With this patch those queries return true.

It seems likely that there are a lot of other places where we should add this.  Or else figure out why it's not happening automatically.

### DIFF
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -7,10 +7,10 @@ module Spree
 
     belongs_to :promotion_category
 
-    has_many :promotion_rules, autosave: true, dependent: :destroy
+    has_many :promotion_rules, autosave: true, dependent: :destroy, inverse_of: :promotion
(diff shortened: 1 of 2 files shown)


jhawthorn:
:+1: 

I'd guess that rails doesn't understand our `Spree::` namespace well enough to add inverses automatically. Definitely needs further investigation.

jordan-brough:
FYI I added an issue to look into things further: https://github.com/solidusio/solidus/issues/1452

PR #1422, opened 2016-09-07
Add strikethrough to ineligible adjustments in admin

jhawthorn:
Previously there was nothing to show that adjustments were ineligible


tvdeyen:
We should consider to show a info or hint why this adjustment is striked through. It is now marked, which is better than before but why it's marked is still unclear to the user. 

I don't even know why an adjustment could exist but is not eligible. Aren't adjustments only get created if they are eligible? Is this really a case and when, why? Thanks is something we should explain. 

Maybe not in this PR, but something we should keep in mind.

PR #1424, opened 2016-09-08
942 billing to default address

peterberkenbosch:
update for #942 including specs changes.

The PR stalled, cherry-picked the commit and updated the spec suite to
reflect the changes.

Closes #942


mtomov:
The fix is really good, especially coupled with #1967 - just cherry-picked both, and about to deploy them to production. It is indeed great to have this usability improvement. 

Thanks for the fix @yeonhoyoon , and taking it up afterwards @peterberkenbosch !

mamhoff:
@tvdeyen I know you fixed this in a client project, do you agree this works well?

tvdeyen:
@peterberkenbosch there is a conflict, could you please resolve?

@mamhoff no, that wasn't me, but this change here is great

peterberkenbosch:
@tvdeyen Fixed the conflict, when CI is green this is mergeable.

peterberkenbosch:
And it's not :), failing on the introduced spec even. Will investigate and fix.

peterberkenbosch:
So, in #1954 we deprecated `Spree::Order#assign_default_addresses!` and introduced `Order.new.assign_default_user_addresses`, I have updated this PR accordingly.

peterberkenbosch:
@tvdeyen we are all green here!

PR #1427, opened 2016-09-08
Connect promotions to orders even if they aren't actionable yet

jordan-brough:
This changes behavior but seems worth it to me. I'm interested in
others' thoughts on it.

This patch makes it so that a customer can apply a coupon code to
their order before the promotion is actionable, as long as their order
meets the promotion rule requirements.

Example with a "Buy one get one 50% off" promotion:

Without this patch:
- Customer adds one item to their order
- Customer tries to apply the promotion but gets a confusing `:coupon_code_unknown_error` message (because none of the items in the cart are actionable yet)
- Customer adds a second item to their cart
- Customer successfully applies promotion and discount is given

With this patch:
- Customer adds an item to their order
- Customer successfully applies promotion  
  (note: they could be confused about not seeing a discount, if they don't understand the promo)
- Customer adds second item to cart and discount is automatically applied

This patch matches the behavior seen when a previously-actionable order
become non-actionable.  E.g. we don't disconnect a "buy one get one 50% off"
promotion from an order when a customer removes their second item from
their cart.  We leave it connected and if they add a second item the promotion
will automatically apply.

If a store really wants to prevent a coupon from being connected in a case
like this they could use `Rule`s to do so.

Thoughts?

### DIFF
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -112,13 +112,11 @@ module Spree
       # If an action has been taken, report back to whatever activated this promotion.
       action_taken = results.include?(true)
 
-      if action_taken
-        # connect to the order
-        order_promotions.find_or_create_by!(
-          order_id: order.id,
-          promotion_code_id: promotion_code.try!(:id)
-        )
-      end
+      # connect to the order
+      order_promotions.find_or_create_by!(
+        order_id: order.id,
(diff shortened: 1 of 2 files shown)


jordan-brough:
Closing. I opened this from the wrong repo.  Opened PR 1428 instead.

PR #1429, opened 2016-09-09
Add "only_allow_most_recent_promotion" configuration

jordan-brough:
Looking for input on this. If people think it's a reasonable then I'll add
specs for it.

UPDATE: I've opened #1433 as an alternative (more general/customizable
approach to solving this issue).

Add an `only_allow_most_recent_promotion` config that allows stores to
only allow a single promotion per order and for that promotion to be
the promotion most recently applied by the user.

This is a somewhat common way to work with promotions and helps reduce
unintended side effects from promotions interacting with each other.
E.g. accidentally giving more of a discount than a store intends to
when a clever customer applies multiple promotions.

This also requires adjusting PromotionHandler::Cart so that it won't
automatically apply "apply_automatically" promotions if this config is
enabled and the order in question already has a promotion applied to
it.

### DIFF
--- a/core/app/models/spree/app_configuration.rb
+++ b/core/app/models/spree/app_configuration.rb
@@ -274,6 +274,12 @@ module Spree
     #   @return [Boolean] Indicates if stock management can be restricted by location
     preference :can_restrict_stock_management, :boolean, default: false
 
+    # @!attribute [rw] only_allow_most_recent_promotion
+    #   @return [Boolean] When this is set to true then previous promotions will
+    #     be removed from an order when a new promotion is activate on the
+    #     order. (default: +false+)
+    preference :only_allow_most_recent_promotion, :boolean, default: false
+
     # searcher_class allows spree extension writers to provide their own Search class
     attr_writer :searcher_class
     def searcher_class
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -706,22 +706,6 @@ module Spree
       Spree::Money.new(total_available_store_credit - total_applicable_store_credit, { currency: currency })
     end
 
-    # Remove a promotion from this order along with any associated adjustments
-    # Note: OrderUpdater will need to be invoked after this is called
-    # @param promotion [Spree::Promotion] the promotion to remove
-    # @return [undefined]
-    def remove_promotion(promotion)
-      [self, *line_items, *shipments].each do |item|
-        item.adjustments.each do |adjustment|
-          if promotion.actions.include?(adjustment.source)
-            item.adjustments.destroy(adjustment)
-          end
-        end
-      end
-      # note: this destroys the join table entry, not the promotion itself
-      promotions.destroy(promotion)
-    end
-
     private
 
     def associate_store
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -706,6 +706,22 @@ module Spree
       Spree::Money.new(total_available_store_credit - total_applicable_store_credit, { currency: currency })
     end
 
+    # Remove a promotion from this order along with any associated adjustments
+    # Note: OrderUpdater will need to be invoked after this is called
+    # @param promotion [Spree::Promotion] the promotion to remove
+    # @return [undefined]
+    def remove_promotion(promotion)
+      [self, *line_items, *shipments].each do |item|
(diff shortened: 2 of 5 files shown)


jasonfb:
This feature is definitely desired, but I wonder if it is enough (sorry, not the answer you wanted). Specifically, we have two 'baseline' Promotions we would want applied *in addition* to a coupon-based Promo code entered by the user:

- Free shipping (applies over an order threshold)
- We have a special volume purchasing discount promotion that is applied to everyone on all orders

Ideally, if I were to design it for myself, I'd put a flag on the Promotions to indicate "User applicable" (this is sort of the opposite of the ```apply_automatically``` which is already there, so I guess you could invert that and consider any non-apply_automatically Promo to be "User applicable"-- fine, semantics).

Then, I'd make the setting be instead of ```only_allow_most_recent_promotion``` I'd change it to ```only_allow_most_recent_promotion_for_user_applied_promotions```

Then this new rule would apply only to user-applied promotions (AKA non-apply_automatically promotions), and still let the apply_automatically Promotions be 'stacked' on top of those User-entered Promo codes.

Just my 2¢, going down the road here is definitely an improvement in the right direction though.

mtomov:
Ha, I did a similar promotion destroy logic for a client just the other day. And the way it worked tied along the logic Jason was saying.

We considered the `user_applicable` promotions to be the ones with code. So, whenever a promo code is entered, all other code promotions are destroyed. We went even that far that even if you enter the very same promo code, it would still destroy & reapply, and give back a success message (again, client requested it that way).

So, I guess there's definitely a requirement for such a functionality.

jordan-brough:
@jasonfb @mtomov thanks for the feedback, very helpful!

A thought on `only_allow_most_recent_promotion_for_user_applied_promotions` -- that does seem reasonable, but it also seems reasonable that a store might want to have `apply_automatically` promotions be replaced by manually-entered promotions doesn't it?  i.e. have the `apply_automatically` promo be a sort of "default promo" that can be overridden by a manually entered-promo?  Maybe we should have the config setting be a three-valued config?  Or if there are other common scenarios that we should also consider then maybe we should make this even more generic somehow.

jasonfb:
Yes indeed and that is a use case for us too... in fact, I just hadn't presented it before. 

In our case, we have a "disable savings meter" (aka "volume discount" as mentioned above) which is specific to us, but an abstraction could include a setting on the *other* promotion to for ```evict_other_promotion_even_if_it_is_automatically_applied```. (this is a very wordy name but you get the idea)

it gets a little tricky with all these rules but I agree the use cases are there. as we have it what we're discussing here makes up a good part of our customization of Spree code (as we have special promotions)

jasonfb:
as an architectural standpoint it kind of seems like a can of worms here, because what you really want is a classification hierarchy where certain promotions kick off other promotions but then others can't, as set on a promotion-by-promotion basis. But if you really build that out it can be very complicated of course.

mtomov:
Another idea then:

As of current, Solidus applies _single_ promotion per Promotion Action Type (Order level, Line Item level, Shipping, etc). Then your current PR, as far as I understand, unites all of those, and only allows for a single promotion discount to be applied, regardless of the promotion action type. 

And what Jason is saying is that we still want to have those "groups" of promotions, out of which a single one would apply, but we don't want those groups to be Promotion Action Type based, right? We want to custom group the promotions per our own choosing, say 1) Free Shipping, 2) Volume Purchasing, 3) Other. So, if we have those 3 groups, a user can get a discount on all 3?

If I've got those right, then wouldn't that be the role of the _Promotion Category_ relationship? Apply the best promotion discount per  _Promotion Category_? So, if all promotions belong to a single category, user could only get a single discount (which is this PR?).

jordan-brough:
@jasonfb @mtomov I've opened PR 1433 as an alternative approach that might be more customizable without having to pull too much logic into Solidus itself.

jordan-brough:
@mtomov I think the promotion category idea is a good one. I know some stores use promotion categories for different purposes so I think I'd hesitate to bake that into Solidus, but in the other PR I've opened I believe you could do it that way using the provided extension points if a particular store wanted to.  Thoughts?

jordan-brough:
Closing this in favor of https://github.com/solidusio/solidus/pull/1433

PR #1433, opened 2016-09-12
Add extension points for handling multiple promotions on one order

jordan-brough:
This is an alternative to #1429.  Instead of supporting a single way to
handle multiple promotions, this attempts to provide extension points
where stores can implement their own requirements.

Looking for input on this. If people think it's a reasonable then I'll add
specs for it.

This PR adds these two extension points:
- Spree::Config.promotions_to_remove_from_order_class
- Spree::Config.automatic_promotion_decision_class

These are meant to allow customization around handling multiple
promotions on a single order.

One example scenario is only allowing a single promotion per order and
for that promotion to be the promotion most recently applied by the
user. This is a somewhat common way to work with promotions and helps
reduce unintended side effects from promotions interacting with each
other. E.g. accidentally giving more of a discount than a store
intends to when a clever customer applies multiple promotions.

With this PR, implementing the above should be possible via something
like this:

``` ruby
class CustomRemover < Spree::Promotion::PromotionsToRemoveFromOrder
  def promotions_to_remove
    last_promo = order.order_promotions.last.try!(:promotion)
    order.promotions - [last_promo]
  end
end

class CustomDecider < Spree::Promotion::AutomaticPromotionDecision
  def attempt_to_apply?
    order.promotions.none?
  end
end
```

There are many other ways that stores customize logic around multiple
promotions, so this is an attempt to make the behavior configurable,
rather than supporting only the use case mentioned above.

### DIFF
--- a/core/app/models/spree/app_configuration.rb
+++ b/core/app/models/spree/app_configuration.rb
@@ -371,6 +371,16 @@ module Spree
       @add_payment_sources_to_wallet_class ||= Spree::Wallet::AddPaymentSourcesToWallet
     end
 
+    attr_writer :promotions_to_remove_from_order_class
+    def promotions_to_remove_from_order_class
+      @promotions_to_remove_from_order_class ||= Spree::Promotion::PromotionsToRemoveFromOrder
+    end
+
+    attr_writer :automatic_promotion_decision_class
+    def automatic_promotion_decision_class
+      @automatic_promotion_decision_class ||= Spree::Promotion::AutomaticPromotionDecision
+    end
+
     def static_model_preferences
       @static_model_preferences ||= Spree::Preferences::StaticModelPreferences.new
     end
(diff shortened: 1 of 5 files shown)


mtomov:
I think it's really good!

I will try to update my comment later with an example of limiting the promotions to one per Promotion Category.

Wouldn't it be better to add the activated promotion as a second parameter to the `PromotionsToRemoveFromOrder#initialize`, similarly to the `AutomaticPromotionDecision` class:
```
  def initialize(order:, promotion:)
    @order = order
    @promotion = promotion
  end
```

as from your example it seems to be a bit of a struggle to take the `last_promo`?

gmacdougall:
Hi Jordan. We're going through stale PR's and this is something we're not feeling comfortable with. We're concerned with adding more customization to a confusing promotion system.

We're thinking we need something more in depth in order to fix some of these problems (similar to when we fixed up the tax system) rather than a simple PR like this.

Going to close for now, but we're open for further discussion on promotions.

PR #1434, opened 2016-09-12
Build payment factory from order.payments association

jhawthorn:
This keeps the order.payments association fresh.

Previously, calling `create(:payment)` would build a payment using `Spree::Payment.new(attributes)` before saving (the factory girl default). This commit changes the factory to use `order.payments.new`, so that the order's payments association will include the new payment.

Before

    order.payments #=> []
    create(:payment, order: order)
    order.payments #=> [] (cached)
    order.payments.reload #=> [#<Payment...

After

    order.payments #=> []
    create(:payment, order: order)
    order.payments #=> [#<Payment...

This PR includes a commit showing the cleanup this allows to the `order/payment_spec.rb`

This change allows #1426 to go green without any changes to specs

### DIFF
--- a/core/lib/spree/testing_support/factories/payment_factory.rb
+++ b/core/lib/spree/testing_support/factories/payment_factory.rb
@@ -27,6 +27,10 @@ FactoryGirl.define do
 
       refunds { build_list :refund, 1, amount: refund_amount }
     end
+
+    initialize_with do
+      order.payments.new(attributes)
+    end
   end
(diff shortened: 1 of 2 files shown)


gmacdougall:
Yay!  Good riddance to reloads!

Gregor

jordan-brough:
nice! 👍

cbrunsdon:
:+1:

PR #1435, opened 2016-09-12
Remove adjustment updates in Shipment#update_amounts

jhawthorn:
Previously calling `Shipment#update_amounts` (usually done by the `OrderUpdater`) would update the cost on the Shipment as well as call update! on all adjustments and update the Shipment's `adjustment_total`.

When used as part of an `OrderUpdate` this isn't necessary as the adjustments will be updated later in the process.

This relates to #1252 where @jordan-brough suggests we should no longer update adjustments outside of `OrderUpdater`.

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -282,11 +282,9 @@ module Spree
     def update_amounts
       if selected_shipping_rate
         self.cost = selected_shipping_rate.cost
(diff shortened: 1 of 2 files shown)


jordan-brough:
👍 nice!

gmacdougall:
👍 

Thanks John!

cbrunsdon:
:+1: :fireworks:

PR #1438, opened 2016-09-13
Remove trackers from Solidus

swively:
Previously, Solidus shipped with an admin panel that allowed creation of trackers. 
This functionality is now available in an extension, `solidus_trackers`

https://github.com/solidusio-contrib/solidus_trackers


cbrunsdon:
I am very pro this. I think most people end up implementing GA on their own and don't use these views, let alone the Tracker models. Maybe we want to have a different GA extension that implements a reasonable set of views? But definitely this is a good step forward.

:+1: 

Though we will probably want to move the external extension into solidus-contrib.

jhawthorn:
Trackers are a great fit for an extension. They integrate very loosely with the rest of solidus, and most users will be rolling their own analytics (especially as google is not the only name).

PR #1439, opened 2016-09-14
Prefer keeping order.order_promotions relationship up to date

jordan-brough:
Prefer keeping `Order#order_promotions` up to date instead of
`Promotion#order_promotions`.  The former is more likely to be
important.

It seems like keeping `Promotion#order_promotions` up to date might not be worth the hassle, but if we wanted to, some ideas are:

```ruby
order.order_promotions.find_or_create_by!(...)
order_promotions.reset # (this is different than order.promotions.reset)
```

or

```ruby
order_promotions << order.order_promotions.find_or_create_by!(...)
```
This generates an extra SQL query though.

or

```ruby
order_promotion = order.order_promotions.find_or_create_by!(...)
if order_promotions.loaded? && !order_promotions.include?(order_promotion)
  order_promotions.proxy_association.add_to_target(order_promotion)
end
```
That seems a bit excessive for this though.

### DIFF
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -114,10 +114,11 @@ module Spree
 
       if action_taken
         # connect to the order
-        order_promotions.find_or_create_by!(
-          order_id: order.id,
-          promotion_code_id: promotion_code.try!(:id)
+        order.order_promotions.find_or_create_by!(
+          promotion: self,
+          promotion_code: promotion_code,
         )
+        order.promotions.reset
       end
 
       action_taken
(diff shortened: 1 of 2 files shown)


jordan-brough:
@gmacdougall if we do want to make sure the relationships are correct, then since it's most likely that `promotion.order_promotions` isn't loaded and won't be used, perhaps the first option would be the best?

gmacdougall:
That's fine by me.

jordan-brough:
@gmacdougall I've updated with a couple more `reset`s.

PR #1440, opened 2016-09-14
Improve JS handling for OrdersController

mtomov:
* Using an instance variable allows JS populate operations to have
    access to the @order from within the rendered `.js.erb` files

  * Assign all line_item errors to the order's base

  * Allow to respond to JS requests for `#populate` and `#update` actions.
    It is left for the developer to actually implement the `js.erb` views

* Adjusts `#update` to perform the behave properly for JS actions as well

--

To consider:

  - to avoid the whole rescue operation, and manually assigning line item
    errors to the order's `:base`, we'd need to do to be doing `order.save`
    instead of `line_item.save!` in [`OrderContents`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_contents.rb#L151)

  -> That would have the effect to "automatically" assign the line item
     errors on the order in the form: `line_items.quantity` in Rails 4.x
     and "line_items[0].quantity" in Rails 5

  More info:
  - https://github.com/rails/rails/pull/8638
  - https://github.com/rails/rails/pull/19686

### DIFF
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -18,10 +18,11 @@ module Spree
 
     def update
       if @order.contents.update_cart(order_params)
+        @order.next if params.key?(:checkout) && @order.cart?
+
         respond_with(@order) do |format|
           format.html do
             if params.key?(:checkout)
-              @order.next if @order.cart?
               redirect_to checkout_state_path(@order.checkout_steps.first)
             else
               redirect_to cart_path
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -41,27 +41,30 @@ module Spree
 
     # Adds a new item to the order (creating a new order if none already exists)
     def populate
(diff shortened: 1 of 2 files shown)


frontend/app/controllers/spree/orders_controller.rb
```diff
@@ -5,6 +5,7 @@ class OrdersController < Spree::StoreController
     helper 'spree/products', 'spree/orders'
 
     respond_to :html
+    respond_to :js, only: [:populate, :update]
```
mtomov: I can actually get rid of this one if you prefer, as one can easily just add the line in a decorator.

jhawthorn: I would prefer that since the JS response will not be in core.

mtomov:
I've actually changed the code around, and also modified the `#update` method, so if you'd like to have a new look, that would be great.

jhawthorn:
Still looking good to me.

@mtomov could you please change the target of this PR to the `master` branch?

cbrunsdon:
Looks reasonable to me :+1:

jhawthorn:
Thanks again @mtomov

PR #1414, opened 2016-09-02
Wallet & Support for non-credit card payment sources - rebased on solidus:master

peterberkenbosch:
work done in #1091 was on different base repositories and I had some issues getting master rebased in there. This PR is mostly here to see if this is making the CI and Hound happy again.

PR #1442, opened 2016-09-14
Use payment_source_class to find reusable sources

stewart:
This method relied on the Gateway -> CreditCard association method `credit_cards`, which may not be present for custom gateways. By using the `payment_source_class` method, we can generalize this code for easier re-use.

### DIFF
--- a/core/app/models/spree/gateway.rb
+++ b/core/app/models/spree/gateway.rb
@@ -60,7 +60,10 @@ module Spree
       if order.completed?
         sources_by_order(order)
       elsif order.user_id
-        credit_cards.where(user_id: order.user_id).with_payment_profile
+        payment_source_class.where(
+          payment_method_id: id,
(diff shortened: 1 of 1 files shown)


jhawthorn:
The basic idea looks okay to me.

This would require that ALL payment sources support the `with_payment_profile` scope.

cc @jordan-brough who has done some work making generic wallets, which might be more of a full-featured solution than this.

jordan-brough:
@peterberkenbosch we don't address `Gateway#reusable_sources` in https://github.com/solidusio/solidus/pull/1414 and it seems like we should.  It seems like we would want to implement like this perhaps:
```ruby
def reusable_sources(order)
  if order.completed?
    sources_by_order(order)
  elsif order.user_id
    order.user.wallet.wallet_payment_sources
  else
    []
  end
```
Thoughts?

Issue #1196, opened 2016-05-27
Accept terms and conditions: How to do this well?


mamhoff:
Copying a Slack discussion here that arose in response to #1195 

> sebashton [1:22 PM] 
> Hi all, we’ve recently had to add terms and condition acceptance to a solidus checkout flow to comply with EU legislation. We thought this would be a nice bit of functionality to commit back to core. We'd appreciate feedback. https://github.com/solidusio/solidus/pull/1195
>  GitHub
> Add terms and conditions to comply with EU legislation by SebAshton · Pull Request #1195 · solidusio/solidus · GitHub
> This PR will add functionality for incorporating a terms and conditions checkbox onto the order confirmation page in the checkout flow. This is required for all stores with their main operating ad... 
> 
> clarke [2:30 PM] 
> sebashton: cool! thanks for contributing that back
> 
> [2:31] 
> sebashton: did you attempt any other ways to do this before you landed on the attr_accessor / before_transition?
> 
> sebashton [2:35 PM] 
> At first we had it as an attribute in the db itself, but we felt it was superfluous.
> 
> kennyadsl [2:35 PM] 
> @sebashton: we had the same requirement for one client and we used a validation on the order with a not persisted attribute
> 
> sebashton [2:36 PM] 
> did you use `acceptance_of`?
> 
> [2:36] 
> @kennyadsl ^
> 
> mamhoff [2:38 PM] 
> clarke, sebashton: We implemented it the exact same way before for a client. I'd be very happy for something like this to land in core, and I also don't think it has to be persisted.
> 
> kennyadsl [2:38 PM] 
> something like:

```
validates :field_name, acceptance: { accept: 'yes' }, if: ->(r) { # some condition }
```

(edited)

> [2:38] 
> I'd also like the step where the validation happen to be configurable
> 
> clarke [2:39 PM] 
> my biggest issue is "yet another before_transition" in the codebase (though I get it does almost nothing and is an optional configuration
> 
> kennyadsl [2:39 PM] 
> because for example we don't have the confirm step
> 
> [2:40] 
> and the client wants it on the address step
> 
> new messages
> sebashton [2:41 PM] 
> this is on the transition to complete, so it could be payment step
> 
> kennyadsl [2:41 PM] 
> you know, clients :troll:
> 
> sebashton [2:41 PM] 
> yea :slightly_smiling_face:
> 
> [2:42] 
> I guess its more of a UX question to where it appears
> 
> mamhoff [2:43 PM] 
[...]

PR #1448, opened 2016-09-16
Show term and condition to user when required acceptance

DanielePalombo:
When the preference `require_term_and_condition_acceptance` is true, the button
`place_order` is disable and over the button appear a textarea with
checkbox. The textarea contains the terms and conditions of checkout, and
is required to check the checkbox to remove disabled state from
`place_order` button, and procede to the next state.

rel #1196

### DIFF
--- a/frontend/app/assets/javascripts/spree/frontend/checkout.js.coffee
+++ b/frontend/app/assets/javascripts/spree/frontend/checkout.js.coffee
@@ -7,3 +7,11 @@
 Spree.disableSaveOnClick = ->
   ($ 'form.edit_order').submit ->
     ($ this).find(':submit, :image').attr('disabled', true).removeClass('primary').addClass 'disabled'
+
+Spree.ready ($) ->
+  termsCheckbox = ($ '#accept_terms_and_conditions')
+  termsCheckbox.change( () ->
+    submitBtn = $(this.closest('form')).find(':submit')
+    submitBtn.prop('disabled', !this.checked)
+    submitBtn.toggleClass('disabled', !this.checked)
+  )
--- a/frontend/app/models/spree/frontend_configuration.rb
+++ b/frontend/app/models/spree/frontend_configuration.rb
@@ -1,5 +1,8 @@
 module Spree
   class FrontendConfiguration < Preferences::Configuration
     preference :locale, :string, default: Rails.application.config.i18n.default_locale
+
+    # Add your terms and conditions in app/views/spree/checkout/_terms_and_conditions.en.html.erb
+    preference :require_terms_and_conditions_acceptance, :boolean, default: false
   end
 end
--- a/frontend/app/views/spree/checkout/_confirm.html.erb
+++ b/frontend/app/views/spree/checkout/_confirm.html.erb
@@ -7,6 +7,19 @@
 <br />
 
 <div class="form-buttons" data-hook="buttons">
-  <%= submit_tag Spree.t(:place_order), :class => 'continue button primary' %>
+  <% Spree::Frontend::Config[:require_terms_and_conditions_acceptance].tap do |requires_acceptance| %>
+    <% if requires_acceptance %>
+      <div class="terms_and_conditions" data-hook="terms_and_conditions">
+        <div class="policy"><%= render partial: "spree/checkout/terms_and_conditions" %></div>
+        <%= check_box_tag :accept_terms_and_conditions, 'accepted', false %>
+        <%= label_tag :accept_terms_and_conditions, Spree.t(:agree_to_terms_of_service) %>
+      </div>
+    <% end %>
+
+    <%= submit_tag Spree.t(:place_order),
+      disabled: requires_acceptance,
+      class: "continue button primary #{ 'disabled' if requires_acceptance }" %>
+  <% end %>
+
   <script>Spree.disableSaveOnClick();</script>
 </div>
--- a/frontend/app/views/spree/checkout/_terms_and_conditions.en.html.erb
+++ b/frontend/app/views/spree/checkout/_terms_and_conditions.en.html.erb
@@ -1,0 +1,1 @@
+Put your terms and conditions here
--- a/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
+++ b/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
@@ -935,6 +935,14 @@ p[data-hook="use_billing"] {
   }
 }
 
+.terms_and_conditions {
+  .policy {
+    height: 100px;
+    width: 100%;
+    overflow: scroll;
+  }
+}
+
 /*--------------------------------------*/
 /* Cart
 /*--------------------------------------*/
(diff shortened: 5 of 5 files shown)


frontend/app/views/spree/checkout/_confirm.html.erb
```diff
@@ -7,6 +7,5 @@
 <br />
 
 <div class="form-buttons" data-hook="buttons">
-  <%= submit_tag Spree.t(:place_order), :class => 'continue button primary' %>
-  <script>Spree.disableSaveOnClick();</script>
+  <%= render :partial => 'spree/shared/place_order', :locals => { :order => @order } %>
```
mtomov: If you only use the `place_order` partial only at a single place, then it's better to do without it, as it only introduces a level of indirection. Also the `spree/checkout/..` path would've been a better place then `spree/shared..`

frontend/app/assets/javascripts/spree/frontend/checkout.js.coffee
```diff
@@ -10,3 +10,8 @@ Spree.disableSaveOnClick = ->
 
 Spree.ready ($) ->
   Spree.Checkout = {}
+
+  terms_checkbox = ($ 'form.edit_order #accept_term_and_condition')
```
mtomov: Maybe just using the `id` will be enough - `accept_term_and_condition` sounds pretty unique, and people may decide to remove the `edit_order` class from the form.

frontend/app/assets/javascripts/spree/frontend/checkout.js.coffee
```diff
@@ -10,3 +10,8 @@ Spree.disableSaveOnClick = ->
...
+  terms_checkbox.change( () ->
+    ($ 'form.edit_order :submit').prop('disabled', !this.checked)
+  )
```
mtomov: Usually using jQuery's `closest does a good job at finding the nearest form, which would save you from specifying it with an ephemeral class.

frontend/app/views/spree/shared/_terms_and_conditions.en.html.erb
```diff
@@ -0,0 +1,64 @@
...
+
+Complaints
+We operate a complaints handling procedure which we will use to try to resolve disputes when they first arise, please let us know if you have any complaints or comments.
+
+Waiver
+If you breach these conditions and we take no action, we will still be entitled to use our rights and remedies in any other situation where you breach these conditions.
+
+Entire Agreement
+The above Terms of Service constitute the entire agreement of the parties and supersede any and all preceding and contemporaneous agreements between you and [Your Online Store URL]. Any waiver of any provision of the Terms of Service will be effective only if in writing and signed by a Director of [Your Online Store URL].
```
mtomov: Again, I'd prefer this to be in the `views/spree/checkout` folder instead of shared.

DanielePalombo:
@mtomov i hope now it's ok for you.

frontend/app/views/spree/checkout/_terms_and_conditions.en.html.erb
```diff
@@ -0,0 +1,64 @@
+____________________________________________________________
```
jhawthorn: I don't think `.html.erb` is appropriate here since html can't be rendered inside the textarea. Either this should be text, or the textarea should be changed to a different container.

frontend/app/views/spree/checkout/_confirm.html.erb
```diff
@@ -7,6 +7,17 @@
 <br />
 
 <div class="form-buttons" data-hook="buttons">
-  <%= submit_tag Spree.t(:place_order), :class => 'continue button primary' %>
+  <% if Spree::Frontend::Config[:require_term_and_condition_acceptance] %>
+    <div class="terms_and_conditions">
+      <textarea readonly style="width: 100%"><%= render partial: "spree/checkout/terms_and_conditions" %></textarea>
```
jhawthorn: I would prefer this had a class instead of using inline styles.

frontend/app/views/spree/checkout/_terms_and_conditions.en.html.erb
jhawthorn: Thought about this slightly more, and it's probably preferable to keep this HTML and to  change the `textbox` into a `div` with a fixed height and `overflow: scroll`

DanielePalombo: I think text is appropriate. In this way we prevent inserting the html code (link, images ecc...) and the documents can be printed in well format.

ref https://github.com/solidusio/solidus/issues/1196#issuecomment-222467628
```
The terms and conditions should be in a format that can be printed or saved – therefore avoid pop-up windows and ensure that they fit within the width of the page and are presented in a way that they will print properly.
```

jhawthorn: > In this way we prevent inserting the html code (link, images ecc...)

I don't see why this is a concern. Nobody is forcing a developer to add those to this file.

> The terms and conditions should be in a format that can be printed or saved

There is no reason a textarea is better for this. Using a div with `overflow: scroll` we even have the opportunity to expand it in a print-only stylesheet if that is a real concern. The "text only" feel of a textarea could be duplicated using `white-space: pre` for sites which really want that, and sites which already have an HTML version of their terms and conditions can use that.

jhawthorn:
Looks good to me.

Thank you @DanielePalombo

peterberkenbosch:
👍

frontend/app/assets/javascripts/spree/frontend/checkout.js.coffee
```diff
@@ -7,3 +7,11 @@
 Spree.disableSaveOnClick = ->
   ($ 'form.edit_order').submit ->
     ($ this).find(':submit, :image').attr('disabled', true).removeClass('primary').addClass 'disabled'
+
+Spree.ready ($) ->
+  Spree.Checkout = {}
```
jordan-brough: What does this do?

frontend/app/views/spree/checkout/_terms_and_conditions.en.html.erb
```diff
@@ -0,0 +1 @@
+Put here your terms and conditions
```
jordan-brough: Not a big deal at all, but "Put your terms and conditions here" would probably be the more standard english way to say this?

frontend/app/models/spree/frontend_configuration.rb
```diff
@@ -1,5 +1,7 @@
 module Spree
   class FrontendConfiguration < Preferences::Configuration
     preference :locale, :string, default: Rails.application.config.i18n.default_locale
+
+    preference :require_terms_and_conditions_acceptance, :boolean, default: false
```
jordan-brough: Could we add a comment here that they'll need to add the actual terms and conditions in `app/views/spree/checkout/_terms_and_conditions.en.html.erb` in their app?

jordan-brough:
Do we have a standard way of visually making a button look disabled in frontend?  I tried this out in a sandbox app and it was a little confusing that the button appeared enabled but was actually disabled.

jordan-brough:
Overall: This seems like a nice addition, thanks for doing it.

I left a few minor questions & comments but nothing that seems blocking to me if others are OK with how things are.

jhawthorn:
Thanks again

PR #1451, opened 2016-09-17
Add ability to remove promotions via Promotion#remove_from

jordan-brough:
Extracted from #1433 as per a chat with @jhawthorn.

This will allow promotions to be removed from orders and allows promotion actions to define how to reverse their side effects on an order.

For now this provides a default `remove_from` on the `PromotionAction` base class, with a deprecation warning that subclasses should define their own `remove_from` method.


core/app/models/spree/promotion.rb
```diff
@@ -220,6 +220,20 @@ def used_by?(user, excluded_orders = [])
       end
     end
 
+    # Removes a promotion and any adjustments or other side effects from an
+    # order.
+    # @param order [Spree::Order] the order to remove the promotion from.
+    # @return [undefined]
```
gmacdougall: If you don't care about the return value `@return [void]` should be used.

core/app/models/spree/promotion/actions/create_adjustment.rb
```diff
@@ -39,6 +39,16 @@ def compute_amount(calculable)
           [(calculable.item_total + calculable.ship_total), amount].min * -1
         end
 
+        # Removes any adjustments generated by this action from the order.
+        # @param order [Spree::Order] the order to remove the action from.
```
gmacdougall: Same here.

core/spec/models/spree/promotion/actions/create_item_adjustments_spec.rb
```diff
@@ -4,11 +4,11 @@ module Spree
   class Promotion
     module Actions
       describe CreateItemAdjustments, type: :model do
-        let(:order) { create(:order) }
+        let(:order) { create(:order_with_line_items, line_items_count: 1) }
         let(:promotion) { create(:promotion, :with_line_item_adjustment, adjustment_rate: adjustment_amount) }
         let(:adjustment_amount) { 10 }
         let(:action) { promotion.actions.first! }
-        let!(:line_item) { create(:line_item, order: order) }
+        let(:line_item) { order.line_items.to_a.first }
```
gmacdougall: The `to_a` shouldn't be needed

core/spec/models/spree/promotion/actions/create_adjustment_spec.rb
```diff
@@ -58,6 +58,27 @@
     end
   end
 
+  describe '#remove_from' do
+    let(:action) { promotion.actions.first! }
+    let(:promotion) { create(:promotion, :with_order_adjustment) }
+
+    # this adjustment should not get removed
```
gmacdougall: Could we name the `other_adjustment` something that implies this more directly?

core/app/models/spree/promotion.rb
jordan-brough: @gmacdougall thanks! Updated and opened a PR to remove more of these: https://github.com/solidusio/solidus/pull/1454

core/app/models/spree/promotion/actions/create_adjustment.rb
jordan-brough: Updated. Thanks.

core/spec/models/spree/promotion/actions/create_item_adjustments_spec.rb
jordan-brough: @gmacdougall this ensures that `line_item` refers to the same object that `OrderUpdater` and the promo code & etc will modify.  Our current implementation of the order factory happens to preload `order.line_items` properly, but that's not tested and I'm not sure it was intentional either.  Perhaps I could remove it here but make an additional PR that tests this explicitly in `order_factory_spec`?

core/spec/models/spree/promotion/actions/create_adjustment_spec.rb
jordan-brough: @gmacdougall sure, I've updated it to `unrelated_adjustment`.

core/spec/models/spree/promotion/actions/create_item_adjustments_spec.rb
jhawthorn: I'm also prefer keeping the `to_a.first`

gmacdougall: Sounds good

flyfy1:
Currently I can see the `remove_from` being defined.. however, where is this method called? I searched through the solidus core but I cannot find a single place where this `remove_from` method being called.

PR #1454, opened 2016-09-19
Change some yarddocs from @return [undefined] to @return [void]

jordan-brough:
See http://www.rubydoc.info/gems/yard/file/docs/Tags.md

@gmacdougall pointed this out in https://github.com/solidusio/solidus/pull/1451


cbrunsdon:
Good by me too thanks :+1:

PR #1460, opened 2016-09-20
Move deprecation forward to 3.0

jhawthorn:


mamhoff:
👍

jordan-brough:
There's a bunch of stuff that we've been telling people "will be removed from Solidus 2.0".  Should we actually remove those things in 2.0?  Or should we just merge this PR for now and be more vigilant in the future about removing deprecated code in major releases?

jordan-brough:
Note: We decided that for the 1.4/2.0 release at least we'll remove the deprecated methods in v2.1 instead of v2.0.

PR #1461, opened 2016-09-21
Remove deprecated code

jordan-brough:
In the deprecation warning we said we'd remove this in Solidus 2.0.

If instead we don't actually want to remove this yet, we should probably update the deprecation message.


jordan-brough:
@jhawthorn I've added a changelog entry and a "Removals" section. Look ok?

jordan-brough:
Went ahead and merged since I started using the "Removals" section in other PRs as well.  I'm happy to see that changed if we want different formatting or whatever though.

PR #1463, opened 2016-09-22
Calculate taxes in OrderContents instead of in a LineItem callback

jordan-brough:
And mark Tax::ItemAdjuster as api-private (to discourage stores &
extensions from calling it outside of the context of
Tax::OrderAdjuster).

This PR may result in a little more load because we're now calling
Tax::OrderAdjuster at some times when previously we would've just
called Tax::ItemAdjuster.

Also, we're now _always_ calling Tax::OrderAdjuster after
OrderContents#update_cart, instead of only when the location changes.

Benefits:
- It gets rid of an ActiveRecord callback
- It seems safer to call this in more cases
- It moves toward performing tax calculations at the order level
instead of at the line level.

See also https://github.com/solidusio/solidus/issues/1252

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## Solidus 2.1.0 (master, unreleased)
 
+*   Remove callback `Spree::LineItem.after_create :update_tax_charge`
+
+    Any code that creates `LineItem`s outside the context of OrderContents
+    should ensure that it calls `order.create_tax_charge!` after doing so.
+
+*   Mark `Spree::Tax::ItemAdjuster` as api-private
+
 *   Analytics trackers were removed from the admin panel; the extension
     `solidus_trackers` provides the same functionality
 
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -31,8 +31,6 @@ module Spree
     validates :price, numericality: true
     validate :ensure_proper_currency
 
-    after_create :update_tax_charge
-
     after_save :update_inventory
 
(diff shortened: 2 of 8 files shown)


mamhoff:
I'm 👍 on this.

bbuchalter:
First let me say, the change looks great from a technical level. Always happy to see more red than green. 👍 

Wanted to comment on this: 
> It moves toward performing tax calculations at the order level instead of at the line level.

I'll say one challenge with this is refunds, at least with TaxCloud, because you don't get any line item breakdown of the tax calculation. If a customer makes a partial refund of an order, without line item records of how much tax was collected, it's impossible to create tax-accurate refunds. This topic hasn't yet been broached in https://github.com/solidusio/solidus/issues/1252, so perhaps that's a better forum for it?

jordan-brough:
> without line item records of how much tax was collected

@bbuchalter to be clear -- the idea is still for taxes to be _calculated_ and _recorded_ in Solidus at the line level.  But we need the tax calculation to be _invoked_ at the order level (i.e. calculate taxes on all the order's items at the same time).  Does that address your concern at all?

bbuchalter:
Yes, absolutely. Thank you. Will be very interested to see how that is done in a performant way! Thanks for all your work and making the work so visible. ❤️

bbuchalter:
> I'll say one challenge with this is refunds, at least with TaxCloud, because you don't get any line item breakdown of the tax calculation.

This turns out not to be true. For TaxCloud the individual taxes are recorded in the `CartItem#amount` field. Thanks Jordan for following up on this with me.

PR #1465, opened 2016-09-22
Remove outdated comment in Tax::ItemAdjuster

jordan-brough:
This was changed in #1389 and OrderUpdater must now be run to update
item and order totals.


jordan-brough:
:)

PR #1466, opened 2016-09-22
Break ItemAdjustments apart and put the logic in OrderUpdater

jordan-brough:
This sets up tax updates in `OrderUpdater` so that all tax updates for an order
happen at the same time.

With `ItemAdjustments` we updated items in this order:

    - item 1
      - update promotions
      - update taxes
      - update cancellations
      - update totals
    - item 2
      - update promotions
      - update taxes
      - update cancellations
      - update totals

However, we need to invoke tax calculation on the order as a whole
(even though the results will be calculated & stored at the line/shipment level) in
order to provide a good extension point for external tax services. See #1252.
That means we need to do this:

    - update promotions
      - item 1
      - item 2
    - update taxes
      - item 1
      - item 2
    - update cancellations
      - item 1
      - item 2
    - update totals
      - item 1
      - item 2

I couldn't think of a reasonable way to refactor ItemAdjustments to be
able to do the above, without breaking backward compatibility in a
significant way. And if we're going to break backward compatibility
then moving this logic into OrderUpdater seemed like a better change
to me as a first step.

Once we have this in place we can start working on providing an
extension point in the "update taxes" step where external tax services
can be plugged in.


core/spec/models/spree/order_updater_spec.rb
```diff
@@ -67,6 +67,245 @@ module Spree
...
+                  order.reload
+                  expect(order.all_adjustments.count).to eq(2), 'Expected two adjustments'
+                  expect(order.all_adjustments.eligible.count).to eq(1), 'Expected one elegible adjustment'
+                  expect(order.all_adjustments.eligible.first.source.promotion).to eq(order_promo2), 'Expected promo2 to be used'
+                end
+              end
+            end
+
+            promo_sequences.each do |promo_sequence|
+              context 'with promo sequence #{promo_sequence}' do
```
jhawthorn: Single quotes prevents interpolation here

core/spec/models/spree/order_updater_spec.rb
```diff
@@ -67,6 +67,245 @@ module Spree
...
+
+                  order.reload
+                  expect(order.all_adjustments.count).to eq(2), 'Expected two adjustments'
+                  expect(order.all_adjustments.eligible.count).to eq(1), 'Expected one elegible adjustment'
+                  expect(order.all_adjustments.eligible.first.source.promotion).to eq(order_promo2), 'Expected promo2 to be used'
+                end
+              end
+            end
+
+            promo_sequences.each do |promo_sequence|
```
jhawthorn: It might make sense to group this loop with the previous identical one.

core/app/models/spree/order_updater.rb
```diff
@@ -164,5 +163,51 @@ def update_payment_state
...
+
+    def update_item_totals
+      [*line_items, *shipments].each do |item|
+        item_cancellation_total = item.adjustments.select(&:cancellation?).sum(&:amount)
+
+        item.adjustment_total = item.promo_total +
+                                item.additional_tax_total +
+                                item_cancellation_total
+
+        if item.changed?
```
jhawthorn: I think I disagree with our rubocop rules here.

core/app/models/spree/order_updater.rb
```diff
@@ -164,5 +163,51 @@ def update_payment_state
...
+
+    def update_cancellations
+      [*line_items, *shipments].each do |item|
+        item.adjustments.select(&:cancellation?).each(&:update!)
+      end
+    end
+
+    def update_item_totals
+      [*line_items, *shipments].each do |item|
+        item_cancellation_total = item.adjustments.select(&:cancellation?).sum(&:amount)
```
jhawthorn: It feels weird that `promo_total`, `included_tax_total`, and `additional_tax_total` are summed in their respective methods, but cancellations are not.

core/app/models/spree/order_updater.rb
jordan-brough: 👍 Updated in https://github.com/solidusio/solidus/pull/1467

core/spec/models/spree/order_updater_spec.rb
jordan-brough: Great idea. Done.

core/spec/models/spree/order_updater_spec.rb
jordan-brough: Thanks! Fixed by the update that groups this loop with the previous one.

core/app/models/spree/order_updater.rb
jordan-brough: Agreed. I've added a comment to that effect -- cancellation_total isn't persisted anywhere so we have to extract it here. (The others need to be done in their respective methods so that, e.g., tax calculations have the updated promo totals).

jordan-brough:
@jhawthorn thanks for the feedback!  I've added commits for all the comments.  I also rebased on latest master to get the updated rubocop rules.

core/app/models/spree/order_updater.rb
```diff
@@ -164,5 +174,71 @@ def update_payment_state
...
+        promotion_adjustments.each(&:update!)
+        Spree::Config.promotion_chooser_class.new(promotion_adjustments).update
+
+        item.promo_total = promotion_adjustments.select(&:eligible?).sum(&:amount)
+      end
+    end
+
+    # Update and select the best promotion adjustment for the order.
+    # We don't update the order.promo_total yet. Order totals are updated later
+    # in #update_adjustment_total since they include the totals from the ordre's
```
mtomov: typo in `ordre`

jordan-brough: Fixed. Thanks!

Issue #1453, opened 2016-09-17
"Country can't be blank" error inserting a new price


alejomo:
I'm trying to add a new price to a product in the Prices tab, I left the selected option "Any country" and is not working, it's showing the "Country can't be blank" error. I debugged the code and "country_iso" attribute is an empty string "", and the backend is asking to allow it empty if is nil. 

https://github.com/solidusio/solidus/blob/054e4bd75c65166d316f3f63be85efb48736bcdb/core/app/models/spree/price.rb#L49-L51

PR #1468, opened 2016-09-24
Convert blank country_iso on prices to nil

mamhoff:
When adding a new price for a product in the admin panel using the
prices tab and not selecting a country for this price ("Any Country"),
the browser will send an empty string as country iso. In the backend,
the intention of this user action is for the price to be saved with a
`country_iso` of `nil`, indicating that it is a fallback price valid
for any country.

This commit converts an empty string `country_iso` to `nil` before
validation, allowing admin users to actually create a new price.

Fixes https://github.com/solidusio/solidus/issues/1453

### DIFF
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -58,6 +58,10 @@ module Spree
       end
     end
 
+    def country_iso=(country_iso)
+      self[:country_iso] = country_iso.presence
+    end
+
     private
 
     def sum_of_vat_amounts
--- a/core/spec/models/spree/price_spec.rb
+++ b/core/spec/models/spree/price_spec.rb
@@ -55,6 +55,12 @@ describe Spree::Price, type: :model do
         it { is_expected.to be_valid }
       end
 
+      context 'when country iso is an empty string' do
+        let(:country_iso) { "" }
+
+        it { is_expected.to be_valid }
+      end
+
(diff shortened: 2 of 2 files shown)


jordan-brough:
Nice, thanks!  Possibly bikeshedding, but would it be more direct to do this in a setter instead of in a `before_validation`?

jordan-brough:
Or even at the controller level rather than the model level?  Just thinking out loud.

mamhoff:
I like the setter idea. I've thought about doing it at the controller level, where it admittedly belongs, but then I view that doing this at the model level is very much the Rails way and would confuse people less if they ever manipulate prices from another, custom controller. 

Will amend.

tvdeyen:
:+1:

jhawthorn:
I don't think this has the intended behaviour if the `country_iso` is being changed.

``` ruby
price = Spree::Price.new
price.country_iso = 'de'
price.country_iso #=> "de"
price.country_iso = nil
price.country_iso #=> "de"
```

I think the setter should be 

``` ruby
self[:country_iso] = country_iso.presence
```

mamhoff:
Thanks @jhawthorn - amended.

jhawthorn:
Great! Thank you

skukx:
Is it possible to merge this fix with past versions of solidus (1.4, 1.3, etc)? and release version bump to gems?

PR #1469, opened 2016-09-24
Remove is_default Boolean from prices (rebased)

mamhoff:
This boolean is either a very bare-bones version of `acts_as_paranoid` or
a very bare-bones version of `papertrail`. However: It's really badly named,
as `is_default` is not something I understand very well at all. It probably means
"the price to be used in the backend and mostly in the frontend too".

We can accomplish the same behavior by simply using an `order` clause in
the `currently_valid` scope on prices.

This PR is #1182 rebased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Solidus 2.1.0 (master, unreleased)
 
+*   Remove `is_default` boolean from `Spree::Price` model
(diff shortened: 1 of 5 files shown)


jordan-brough:
Seems reasonable but the specs are unhappy.

jhawthorn:
:+1:

PR #1474, opened 2016-09-27
Fix for repair adjustments code

jordan-brough:
Fix for Adjustment#repair_adjustments_associations_on_create

When the item gets deleted before after_commit runs.

I ran into this while working on some changes where some tax
adjustments would get created and destroyed in the same transaction.

Also add specs for this code.
See individual commits.

Is this worth adding to the 2.0 & 1.4 branches as well?

### DIFF
--- a/core/app/models/spree/adjustment.rb
+++ b/core/app/models/spree/adjustment.rb
@@ -184,7 +184,7 @@ module Spree
     end
 
     def repair_adjustments_associations_on_create
-      if adjustable.adjustments.loaded? && !adjustable.adjustments.include?(self)
+      if adjustable.adjustments.loaded? && !adjustable.adjustments.include?(self) && Spree::Adjustment.exists?(id)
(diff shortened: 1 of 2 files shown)


mamhoff:
Whether this should be on the 1.4 and 2.0 branches is a decision I'd like to leave up to @jhawthorn.

jordan-brough:
Thanks @cbrunsdon @mamhoff.

@jhawthorn mentioned offline that it might be nice to avoid the query using `destroyed?` so I've pushed a commit doing that.

@jhawthorn was also going to look at another way of attacking this problem that might allow using `after_save` instead of `after_commit`.  @jhawthorn do you want to do that in this PR or separately?

jordan-brough:
Note about 1.4:  My specs don't work for 1.4 unfortunately :( since after_commit doesn't work correctly in specs in Rails <= 4.  See [this readme](https://github.com/grosser/test_after_commit).

jordan-brough:
@jhawthorn merging this for now so that I can rebase https://github.com/solidusio/solidus/issues/1479 against master.  I assume you'll open another PR if you get the other method working.

PR #608, opened 2015-12-19
Update jquery payment js library

mtomov:
* Update old version from 3 Sep 2013 to the latest release.

  * includes lots of bug & browser compatibility fixes

  * Specify type="tel" for numerical input fields, which fixes
    non-working card expiry field on [Samsung android phones](http://github.com/stripe/jquery.payment/issues/112)

    Follows library [recommendations](https://github.com/stripe/jquery.payment#mobile-recommendations)

---

Second commit is a bit controversial, so you might consider excluding it.

    Autocomplete name on the card and card number, but forbid CVV autocomplete.

      * As per the autocomplete recommendations at:
        https://github.com/stripe/jquery.payment#autocomplete-recommendations

---

This _gateway is anyhow overridden by the gateways, so the autocomplete might not matter; 

On the other hand, the addition of the `type=tel` attribute is indeed important to fix annoying bugs on Samsung android phones, which occur without it (even with the latest version of the jquery.payment library).

PR #1477, opened 2016-09-28
Update CC brand detection regex

jhawthorn:
The existing hash was incomplete, missing some brands. It was also missing MasterCard's new (October 2016) BIN range (222100-272099).

This hash was originally added in spree/spree#3918 and has not been changed since. I don't know the original source, but I would rather trust the Hash taken from ActiveMerchant.

One notable difference is that this will treat maestro and MasterCard's as separate brands.

NB: For most stores, this code will not be run, and instead `jqyery.payment` on the frontend will be responsible for providing the credit card brand (already updated in #608 and understands MasterCard's new BIN. Thanks @mtomov).

### DIFF
--- a/core/app/models/spree/credit_card.rb
+++ b/core/app/models/spree/credit_card.rb
@@ -105,7 +105,10 @@ module Spree
     # @return [String] the credit card type if it can be determined from the
     #   number, otherwise the empty string
     def try_type_from_number
-      CARD_TYPES.find{ |type, pattern| return type.to_s if number =~ pattern }.to_s
+      CARD_TYPES.each do |type, pattern|
+        return type.to_s if number =~ pattern
+      end
+      ''
     end
 
     # @return [Boolean] true when a verification value is present
--- a/core/app/models/spree/credit_card.rb
+++ b/core/app/models/spree/credit_card.rb
@@ -29,14 +29,22 @@ module Spree
     # needed for some of the ActiveMerchant gateways (eg. SagePay)
     alias_attribute :brand, :cc_type
 
+    # Taken from ActiveMerchant
+    # https://github.com/activemerchant/active_merchant/blob/2f2acd4696e8de76057b5ed670b9aa022abc1187/lib/active_merchant/billing/credit_card_methods.rb#L5
     CARD_TYPES = {
-      visa: /^4[0-9]{12}(?:[0-9]{3})?$/,
-      master: /(^5[1-5][0-9]{14}$)|(^6759[0-9]{2}([0-9]{10})$)|(^6759[0-9]{2}([0-9]{12})$)|(^6759[0-9]{2}([0-9]{13})$)/,
-      diners_club: /^3(?:0[0-5]|[68][0-9])[0-9]{11}$/,
-      american_express: /^3[47][0-9]{13}$/,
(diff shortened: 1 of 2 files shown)


core/app/models/spree/credit_card.rb
```diff
@@ -97,8 +105,10 @@ def set_last_digits
     # @return [String] the credit card type if it can be determined from the
     #   number, otherwise the empty string
     def try_type_from_number
-      numbers = number.delete(' ') if number
-      CARD_TYPES.find{ |type, pattern| return type.to_s if numbers =~ pattern }.to_s
+      CARD_TYPES.each do |type, pattern|
```
tvdeyen: Wouldn't `detect` have the same effect? 

Also, I think you don't have to cast strings on `type`  Your hash keys are all strings already.

tvdeyen: Oh, nasty. You have the early return because you return empty string below if a number matches.

I would prefer `.select {...} || ''`, but it's on you to decide

core/app/models/spree/credit_card.rb
```diff
@@ -29,14 +29,22 @@ class CreditCard < Spree::Base
...
+    # https://github.com/activemerchant/active_merchant/blob/2f2acd4696e8de76057b5ed670b9aa022abc1187/lib/active_merchant/billing/credit_card_methods.rb#L5
     CARD_TYPES = {
-      visa: /^4[0-9]{12}(?:[0-9]{3})?$/,
-      master: /(^5[1-5][0-9]{14}$)|(^6759[0-9]{2}([0-9]{10})$)|(^6759[0-9]{2}([0-9]{12})$)|(^6759[0-9]{2}([0-9]{13})$)/,
-      diners_club: /^3(?:0[0-5]|[68][0-9])[0-9]{11}$/,
-      american_express: /^3[47][0-9]{13}$/,
-      discover: /^6(?:011|5[0-9]{2})[0-9]{12}$/,
-      jcb: /^(?:2131|1800|35\d{3})\d{11}$/
-    }
+      'visa'               => /^4\d{12}(\d{3})?(\d{3})?$/,
```
tvdeyen: Why not take the Regex from ActiveMerchant instead of copy it over. This way we have to always have to have an eye on this.

core/app/models/spree/credit_card.rb
tvdeyen: `.detect` obviously 🙄

jhawthorn: `.detect` would return us (for example) `['visa', /^4\d{12}(\d{3})?(\d{3})?$/]` so the code would need to be something like

```
card_type_pair = CARD_TYPES.detect { |type, pattern| numbers =~ pattern }
if card_type_pair
  card_type_pair[0]
else
  ''
end
```

I think the return is more clear

core/app/models/spree/credit_card.rb
cbrunsdon: I'm pro reducing and removing our reliance on active_merchant in core, so I'm :+1:  on copying it in (it also isn't something that changes frequently).

jordan-brough: > reducing and removing our reliance on active_merchant

Ditto

PR #1480, opened 2016-09-28
Remove unused code in Admin::PromotionsController

jordan-brough:
This was removed in c87726d1e93a07cdba1a868180b49d4ecd3779f7.


jordan-brough:
@cbrunsdon I took a harder look just now and it's because in that commit I referenced we overrode the `create` action in the controller so this doesn't get called anymore: https://github.com/solidusio/solidus/blob/v2.0.0/backend/app/controllers/spree/admin/resource_controller.rb#L53

PR #1483, opened 2016-09-29
OptionValue Promo Rule selected option value not appearing

dholdren:
Using the Spree::Promotion::Rules::OptionValue, after adding a
  product and option value, the selected option value
  was not appearing


dholdren:
yeah I wasn't sure which way to go. product_picker.js does it this way: https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/product_picker.js#L18

jhawthorn:
Thank you

PR #1485, opened 2016-09-29
Stop mutating input in LocalizedNumber.parse

jhawthorn:
Previously, `LocalizedNumber.parse` would perform it's `gsub!` replacements directly on the string passed in.


Senjai:
:+1:

Issue #1436, opened 2016-09-13
No flash[:error] when updating user in backend


ericsaupe:
In the admin/users_controller#update method there is no call to display what errors happened if the user doesn't save. This has caused complaints from our users who are confused why some actions aren't taking effect even though the page reloads.

https://github.com/solidusio/solidus/blob/master/backend/app/controllers/spree/admin/users_controller.rb#L36-L48

mamhoff:
I think this issue is fixed via #1487 .

PR #1487, opened 2016-09-30
Show flash error messages on admin users actions

kennyadsl:
taking the default resource controller behavior from
https://github.com/nebulab/solidus/blob/284cf74d00ea842553aeffc58eff50d3e12900dd/backend/app/controllers/spree/admin/resource_controller.rb#L44

ref #1436


jhawthorn:
Test failure is unrelated (sorry). Would you mind rebasing against master to fix it?

kennyadsl:
@jhawthorn done!

gmacdougall:
Thanks Alberto!

PR #1488, opened 2016-09-30
Change new order button text on users edit page

kennyadsl:
It wasn't immediately clear (to me) that this "Create Order" actually means "Create a new order for this user". Please feel free to close this PR if you guys think it's ok as is. 😄

### DIFF
--- a/backend/app/views/spree/admin/users/_user_page_actions.html.erb
+++ b/backend/app/views/spree/admin/users/_user_page_actions.html.erb
@@ -1,7 +1,7 @@
 <% content_for :page_actions do %>
   <% if can?([:admin, :create], Spree::Order) %>
     <li>
-      <%= button_link_to Spree.t(:create_new_order), spree.new_admin_order_path(user_id: @user.id) %>
+      <%= button_link_to t(".create_order"), spree.new_admin_order_path(user_id: @user.id) %>
     </li>
   <% end %>
 <% end %>
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -797,6 +797,9 @@ en:
           edit_price: Edit Price
         new:
           new_price: New Price
+      users:
+        user_page_actions:
+          create_order: Create order for this user
     administration: Administration
     agree_to_privacy_policy: Agree to Privacy Policy
     agree_to_terms_of_service: Agree to Terms of Service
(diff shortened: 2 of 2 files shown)


mamhoff:
If we're changing the key, I would rather change it to a properly namespaced key. There is the convention with `t(".create_use_order")` which will expand into a partial-specific path, also namespaced under `spree`: `spree.admin.users.user_page_actions.create_user_order`.

@jhawthorn What do you think? I would love to get rid of the `Spree.t` helper, as it really does not provide a lot of value over the standard Rails helper.

kennyadsl:
@mamhoff I can't find any reference about this convention in the codebase. Can you point me to something more specific please?

mamhoff:
It's not a solidus convention, it's a rails thing (not super-well-known, but 💯 ) From http://api.rubyonrails.org/classes/ActionView/Helpers/TranslationHelper.html#method-i-translate: 
> Second, if the key starts with a period translate will scope the key by the current partial. Calling translate(".foo") from the people/index.html.erb template is equivalent to calling translate("people.index.foo"). This makes it less repetitive to translate many keys within the same partial and provides a convention to scope keys consistently.

kennyadsl:
@mamhoff ah cool! I like this convention. IMHO scoping the key will also make it easier to understand the context to who will do the translation. I've made changes you proposed.

frankie-loves-jesus:
Can't you just provide context by making comments and properly sectioning your translation files? And like the guys mentioned, being more specific. Personally I gave up on scopes a long time ago. Didn't like the extra weight.

kennyadsl:
Thinking again, a downside of this approach could be that if that string is moved into another partial, we'll also need to change the translation keys, here and on solidus_i18n. Can this be acceptable?

mamhoff:
I think as long as the key is only used in that particular partial, we can just take that. If we need to reuse it somewhere else, we have two options: New key, or explicitly referring to the previous partial's name (`t(:new_user_order, 'spree.admin.user.user_page_actions.create_user_page_order'). Not a big deal. 

Regarding "Giving up on scopes": Do not do that. It's frequently misses subtleties in other languages, and we end up with terrible, terrible keys like "state" meaning BOTH "region" and "status", which e.g. in German are different things. Do not give up scopes.

jhawthorn:
Thanks everyone. I like us transitioning to the magic-scoped translation keys when we have a translation for a specific location.

PR #1493, opened 2016-10-03
Disable RuboCop Style/EmptyElse

jordan-brough:
I think an explicit nil sometimes makes code more readable. E.g.:

    if
      ...
    elsif
      ...
    elsif
      ...
    elsif
      ...
    else
      nil
    end

lets me know that the "else" case was specifically considered and not
just forgotten.  It's also a great spot to leave a comment sometimes.

In any case, it doesn't seem like something we need to mandate
globally.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,6 +32,9 @@ Style/MutableConstant:
 Style/ClassAndModuleChildren:
   Enabled: false
 
+Style/EmptyElse:
+  Enabled: false
+
 Style/GuardClause:
   Enabled: false
 


jordan-brough:
@jhawthorn are you thinking that empty else's are indicative of bugs?  I could see this being valid:
```ruby
if order.blah?
  ...do something...
elsif
  ...
elsif
  ...
elsif
  ...
else
  # don't do anything, order is foo
end
```
But if we think it's more likely to be a bug then I'm fine w/ `EnforcedStyle: empty`.  I just personally would love to keep Rubocop focused on things that are bugs or that cause obvious pain.

jhawthorn:
I suppose I'm fine with that too. I guess my ideal rule would be to allow either `nil` or a comment.

PR #1495, opened 2016-10-03
Remove is_default checkbox from prices edit

omnistegan:
As is_default in the Spree::Price model was removed in #1469 (#1182), the product prices page in admin backend should not have a checkbox for it as it prevents loading of the edit page with a no method error.

I've included a small test that ensures the edit page loads on being linked from the prices index page.

### DIFF
--- a/backend/app/views/spree/admin/prices/_form.html.erb
+++ b/backend/app/views/spree/admin/prices/_form.html.erb
@@ -41,12 +41,6 @@
       <% end %>
     </div>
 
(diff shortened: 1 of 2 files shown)


mamhoff:
Umm... Yes, that was my fault :). 👍

gmacdougall:
Thanks Graham. Should this be back-ported as a bugfix to any previous releases?

mamhoff:
The commit removing the `is_default` column has not yet been released, so no need for backporting.

gmacdougall:
Perfect   👍

PR #1496, opened 2016-10-03
Make PaymentsHelper available to Order views

eric1234:
The `confirm` screen shows the [payments info](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/orders/confirm.html.erb#L54). This [uses](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/payments/_list.html.erb#L22) the helper [`payment_method_name`](https://github.com/solidusio/solidus/blob/master/backend/app/helpers/spree/admin/payments_helper.rb#L4) without explicitly indicating that controller is using that helper.

In many Rails apps, this may be ok due to `config.action_controller.include_all_helpers` defaulting to `true`. But if the application has configured this to be `false` the confirm screen breaks. I am personally a fan of the old Rails behavior (i.e. don't put all helpers in a big pile but make separate helper modules actually mean something). Even though it is the old behavior it is still supported behavior by Rails so it seems an effort to work either way should be made.

### DIFF
--- a/backend/app/controllers/spree/admin/orders_controller.rb
+++ b/backend/app/controllers/spree/admin/orders_controller.rb
@@ -1,6 +1,8 @@
 module Spree
   module Admin
     class OrdersController < Spree::Admin::BaseController
+      helper 'spree/admin/payments'
+
       before_action :initialize_order_events
       before_action :load_order, only: [:edit, :update, :complete, :advance, :cancel, :resume, :approve, :resend, :unfinalize_adjustments, :finalize_adjustments, :cart, :confirm]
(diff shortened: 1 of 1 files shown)


jasonfb:
👍  for explicitness. I agree about not wanting to include all helpers in all controllers -- things get very messy on large apps when you do that.

eric1234:
Agreed. Unfortunately my upstream commits are limited to just stuff I hit while working on my primary goal. Systematically going through all screens to make sure everything is `include_all_helpers=false` happy is beyond my scope. But my config is set to false so if I hit any other screens as I'm hacking will send upstream.

PR #1499, opened 2016-10-03
Require ruby version 2.2.2 for all gems

### LINKED ISSUE #1494: Shouldn't v2.0 be on ruby version 2.2.2?
Since v2.0 is using Rails 5, shouldn't the ruby version be 2.2.2+

http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#ruby-2-2-2-required

jhawthorn:
Rails 5 requires ruby 2.2.2, so we can now do the same.

Fixes #1494

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -17,7 +17,8 @@ Gem::Specification.new do |s|
   s.files        = `git ls-files`.split("\n")
   s.require_path = 'lib'
 
-  s.required_ruby_version = '>= 2.1.0'
+  s.required_ruby_version     = '>= 2.2.2'
+  s.required_rubygems_version = '>= 1.8.23'
 
   s.add_dependency 'activemerchant', '~> 1.48'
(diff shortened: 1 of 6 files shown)


core/solidus_core.gemspec
```diff
@@ -17,7 +17,8 @@ Gem::Specification.new do |s|
   s.files        = `git ls-files`.split("\n")
   s.require_path = 'lib'
 
-  s.required_ruby_version = '>= 2.1.0'
+  s.required_ruby_version     = '>= 2.2.0'
+  s.required_rubygems_version = '>= 1.8.23'
```
tvdeyen: `>= 2.2.2`

jhawthorn: Thanks. Fixed

tvdeyen: Thank you

jhawthorn:
> Probably deserves a changelog entry?

I don't think so, we already require 2.2.2 by requiring rails 5.

PR #1500, opened 2016-10-03
Remove 'remove' class from cancel button on image edit page

omnistegan:
The 'Cancel' button in the edit images admin page inadvertently has a `remove` class attached to it.
This causes the unexpected behaviour of hiding the form-buttons instead of returning to the images index.

The `remove` class appears to only be used in promotions forms as is defined in [nested-attributes.js](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/nested-attribute.js).

### DIFF
--- a/backend/app/views/spree/admin/images/edit.html.erb
+++ b/backend/app/views/spree/admin/images/edit.html.erb
@@ -23,7 +23,7 @@
     <div class="clear"></div>
     <div class="form-buttons filter-actions actions" data-hook="buttons">
       <%= button Spree.t('actions.update') %>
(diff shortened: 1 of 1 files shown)


jhawthorn:
> The remove class appears to only be used in promotions forms as is defined in nested-attributes.js.

I investigated this and it actually isn't being used by promotion as far as I can tell. Though some promotion rules are using the `remove` class, the `nested-attributes.js` code runs only on page init and the promotion rules templates are only added after that.

I will open up a follow-up PR to remove that file.

Issue #1399, opened 2016-08-19
Why do we store currency on line items


jordan-brough:
Given:
- Orders have a currency
- We don't want to allow orders with line items of varying currencies

Could we remove currency from line items and instead delegate it to the order?

gmacdougall:
We can likely remove currency from LineItem. When I was originally adding support for currencies, I went in with a very basic assumption: everywhere that has an amount which is representing money gets a currency.

It was probably a mistake to add it to LineItem rather than rely on a join or delegation to make it happen.

mamhoff:
I have tried to remove it in the prices refactoring, but at the time we felt it to be unrelated, so I axed that change. I still would like to see that gone.

kennyadsl:
I'm working on this issue, for now I only made [this commit](https://github.com/nebulab/solidus/commit/48dc2f0b12a991b485a7a17033703a919271a3a2). I'm not sure how to proceed in terms of validation. If there's no `Spree::Price` with the currency of the current order, the line item will have nil price. Is this something that could happen or we always have a price in each currency?

mamhoff:
Hey @kennyadsl - I looked through the code and have issues with it :) - By default, a user should only be shown products and variants that have a price in the current currency (that's why the `Variant.with_prices` [scope] exists). (https://github.com/solidusio/solidus/blob/master/core/app/models/spree/variant.rb#L109-L112]). The default product searcher, which is used in the standard frontend products controller, will also exclude products without matching prices: https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/search/base.rb#L32-L34

I think the line item should be invalid and error out if no price with the right pricing options can be found.

kennyadsl:
Hey @mamhoff , thanks for reading my commit. 

I think that check is done [here](https://github.com/nebulab/solidus/commit/48dc2f0b12a991b485a7a17033703a919271a3a2#diff-d4464f8dddd86dd147739f1fcf6d589bR186) and tested [here](https://github.com/nebulab/solidus/commit/48dc2f0b12a991b485a7a17033703a919271a3a2#diff-e2c80773d9997dad992b4f6754a00847R146). Anyway there's still something to fix there, probably also find better names for errors and validation method. What do you think?

mamhoff:
Hey @kennyadsl I left some comments on the commit itself; I think we can simplify things a little bit more as some of the stuff in there has been deprecated for a while already.

kennyadsl:
[...]

PR #1507, opened 2016-10-07
Remove currency from line items

kennyadsl:
This PR removes currency from line items. That field is no more useful since all order contents will have the same currency (the order one). 

ref: #1399 

#### TODO:

- [x] Understand what happens when the order currency is changed. Can this happen? Shouldn't we recalculate prices for all line items with the new currency? Is this already handled by the order merger? [Handled with a separate Issue](https://github.com/solidusio/solidus/issues/1515) 
- [x] Remove currency from `spree_line_items` database table. I'm a bit concerned about this. What happens if there are users that rely on this field has custom logic related to line item currency (like statistics)? They could loose them. Can we proceed anyway or it's better to leave the field in the database and mark it for a future removal?
- [x] Update changelog

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -40,6 +40,7 @@ module Spree
     # @return [Spree::Product, nil] the product associated with this line
     #   item, if there is one
     delegate :product, to: :variant
+    delegate :currency, to: :order, allow_nil: true
 
     attr_accessor :target_shipment
 
@@ -133,10 +134,6 @@ module Spree
       Spree::Deprecation.warn 'Spree::LineItem#currency= is deprecated ' \
         'and will take no effect.',
         caller
-    end
-
-    def currency
-      order.currency if order
     end
 
     private
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -29,7 +29,6 @@ module Spree
       greater_than: -1
     }
     validates :price, numericality: true
-    validate :ensure_proper_currency
 
     after_save :update_inventory
 
@@ -88,12 +87,11 @@ module Spree
     alias display_total display_amount
     deprecate display_total: :display_amount, deprecator: Spree::Deprecation
 
-    # Sets price and currency from a `Spree::Money` object
+    # Sets price from a `Spree::Money` object
     #
-    # @param [Spree::Money] money - the money object to obtain price and currency from
+    # @param [Spree::Money] money - the money object to obtain price from
     def money_price=(money)
-      self.price = money.to_d
-      self.currency = money.currency.iso_code
+      self.price = money.to_d if money
     end
 
     # @return [Boolean] true when it is possible to supply the required
@@ -118,7 +116,10 @@ module Spree
 
       assign_attributes options
 
-      # There's no need to call a pricer if we'll set the price directly.
+      # When price is part of the options we are not going to fetch
+      # it from the variant. Please note that this always allows to set
+      # a price for this line item, even if there is no existing price
+      # for the associated line item in the order currency.
       unless options.key?(:price) || options.key?('price')
         self.money_price = variant.price_for(pricing_options)
       end
@@ -126,6 +127,16 @@ module Spree
 
     def pricing_options
       Spree::Config.pricing_options_class.from_line_item(self)
+    end
+
+    def currency=(_currency)
+      Spree::Deprecation.warn 'Spree::LineItem#currency= is deprecated ' \
+        'and will take no effect.',
+        caller
+    end
+
+    def currency
+      order.currency if order
     end
 
     private
@@ -149,7 +160,6 @@ module Spree
       # If the legacy method #copy_price has been overridden, handle that gracefully
       return handle_copy_price_override if respond_to?(:copy_price)
 
-      self.currency ||= order.currency
       self.cost_price ||= variant.cost_price
       self.money_price = variant.price_for(pricing_options) if price.nil?
       true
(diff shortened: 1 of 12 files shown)


core/app/models/spree/line_item.rb
```diff
@@ -118,14 +116,23 @@ def options=(options = {})
...
-      Spree::Config.pricing_options_class.from_line_item(self)
+    def currency=(_currency)
+      Spree::Deprecation.warn 'Spree::LineItem#currency= is deprecated ' \
+        'and will take no effect.',
+        caller
+    end
+
+    def currency
+      order.currency if order
     end
```
mamhoff: This can be written shorter as a delegate: 
``` ruby
delegate :currency, to: :order
```

core/app/models/spree/line_item.rb
```diff
@@ -88,12 +87,11 @@ def pre_tax_amount
     alias display_total display_amount
     deprecate display_total: :display_amount, deprecator: Spree::Deprecation
 
-    # Sets price and currency from a `Spree::Money` object
+    # Sets price from a `Spree::Money` object
     #
-    # @param [Spree::Money] money - the money object to obtain price and currency from
+    # @param [Spree::Money] money - the money object to obtain price from
     def money_price=(money)
```
mamhoff: I think this should have a Safeguard: It should raise a good error when trying to assign a price with a currency different to the order currency. Otherwise the information would just be lost. So something along these lines: 

``` ruby
class CurrencyMismatch < StandardError; end

def money_price=(money)
   return unless money
   raise CurrencyMismatch, "Line item price currency must match order currency!" if money.currency.iso_code != currency
   self.price = money.to_d
end

File: core/app/models/spree/line_item.rb [hunk 3]
```diff
@@ -118,14 +116,23 @@ def options=(options = {})
...
+      # it from the variant. Please note that this always allows to set
+      # a price for this line item, even if there is no existing price
+      # for the associated line item in the order currency.
       unless options.key?(:price) || options.key?('price')
-        self.money_price = variant.price_for(pricing_options)
+        self.money_price = price_from_options
       end
     end
 
-    def pricing_options
```
mamhoff: I like removing things from the public interface, but I'm not sure about moving this method to the private section.

core/app/models/spree/line_item.rb
```diff
@@ -149,9 +156,8 @@ def set_pricing_attributes
       # If the legacy method #copy_price has been overridden, handle that gracefully
       return handle_copy_price_override if respond_to?(:copy_price)
 
-      self.currency ||= order.currency
       self.cost_price ||= variant.cost_price
-      self.money_price = variant.price_for(pricing_options) if price.nil?
+      self.money_price = price_from_options if price.nil?
```
mamhoff: Is there a reason for extracting this method? I wouldn't do that. :)

core/app/models/spree/variant/pricing_options.rb
```diff
@@ -33,7 +33,7 @@ def self.default_price_attributes
       def self.from_line_item(line_item)
         tax_address = line_item.order.try!(:tax_address)
         new(
-          currency: line_item.order.try(:currency) || line_item.currency || Spree::Config.currency,
+          currency: line_item.order.try(:currency) || Spree::Config.currency,
```
mamhoff: With the delegate, this can actually be written shorter as `line_item.currency` :)

core/app/models/spree/line_item.rb
kennyadsl: I reverted this change for now. I'm probably missing something but at the moment I cannot see any scenario when something else then the line item itself could need to be aware of these specific line item pricing options. That's why I'd move it to the private section, but I'm ok leaving it in the public one as well.

core/app/models/spree/line_item.rb
kennyadsl: done, but [assigning nil to line item price when money is nil](https://github.com/solidusio/solidus/pull/1507/commits/73660fa3f4249b745bbd39fb7caa177cd4e18c12).

kennyadsl:
@mamhoff please check again when you have time, thanks a lot!

core/app/models/spree/line_item.rb
mamhoff: Yes, this is good. I have a feeling one could golf around in this method a bit more, but I'm fine leaving it as is because it does the right things.

core/app/models/spree/line_item.rb
```diff
@@ -118,7 +124,10 @@ def options=(options = {})
 
       assign_attributes options
 
-      # There's no need to call a pricer if we'll set the price directly.
+      # When price is part of the options we are not going to fetch
+      # it from the variant. Please note that this always allows to set
+      # a price for this line item, even if there is no existing price
+      # for the associated line item in the order currency.
```
mamhoff: Really good documentation. I wish I'd had that when first reading this method :)

core/app/models/spree/line_item.rb
```diff
@@ -88,12 +90,16 @@ def pre_tax_amount
...
     #
-    # @param [Spree::Money] money - the money object to obtain price and currency from
+    # @param [Spree::Money] money - the money object to obtain price from
     def money_price=(money)
+      unless money
+        self.price = nil
+        return
+      end
+      raise CurrencyMismatch, "Line item price currency must match order currency!" if money.currency.iso_code != currency
       self.price = money.to_d
```
jhawthorn: This is a little tough to parse with the early return inside the unless (which has me thinking about the rest of the method like the double negative unless...else).

I think it would read better as an if/elseif/else

``` ruby
  if !money
    self.price = nil
  elsif money.currency.iso_code != currency
    raise CurrencyMismatch, "Line item price currency must match order currency!"
  else
    self.price = money.to_d
  end
```

PR #1510, opened 2016-10-07
Admin pricing view

omnistegan:
This PR expands on the Admin Prices view to clarify master/variant pricing. See #1167 

Master and Variant related prices have been separated into their own tables utilising new scopes on the Price model. Additionally, Option Types and Option Values are shown in a table as a reference for the user. As recommended by @Mandily Tooltips are shown in the Master Variant and Options tables to clarify their content.

Tests regarding assigns in this context have been updated to reflect these changes.

Here is a screenshot for reference:
tooltip

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,9 @@
 
 *   The `lastname` field on `Address` is now optional. [#1369](https://github.com/solidusio/solidus/pull/1369)
 
+*   The admin prices listings page now shows master and variant prices
+    seperately. This changes `@prices` to `@master_prices` and `@variant_prices` in prices_controller
+
 *   Removals
 
     * Removed deprecated method `Spree::TaxRate.adjust` (not to be confused with
--- a/backend/app/controllers/spree/admin/prices_controller.rb
+++ b/backend/app/controllers/spree/admin/prices_controller.rb
@@ -7,8 +7,14 @@ module Spree
         params[:q] ||= {}
 
         @search = @product.prices.accessible_by(current_ability, :index).ransack(params[:q])
-        @prices = @search.result
+        @master_prices = @search.result
           .currently_valid
+          .for_master
+          .order(:variant_id, :country_iso, :currency)
+          .page(params[:page]).per(Spree::Config.admin_variants_per_page)
+        @variant_prices = @search.result
+          .currently_valid
+          .for_variant
           .order(:variant_id, :country_iso, :currency)
           .page(params[:page]).per(Spree::Config.admin_variants_per_page)
       end
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1177,6 +1177,8 @@ en:
     hints:
       spree/price:
         country: "This determines in what country the price is valid.<br/>Default: Any Country"
+        master_variant: "Changing master variant prices will not change variant prices below, but will be used to populate all new variants"
+        options: "These options are used to create variants in the variants table. They can be changed in the variants tab"
       spree/product:
         promotionable: "This determines whether or not promotions can apply to this product.<br/>Default: Checked"
         shipping_category: "This determines what kind of shipping this product requires.<br/> Default: Default"
@@ -1337,6 +1339,7 @@ en:
     manual_intervention_required: Manual intervention required
     manage_variants: Manage Variants
     master_price: Master Price
+    master_variant: Master Variant
     match_choices:
       all: All
       none: None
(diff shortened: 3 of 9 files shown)


backend/app/controllers/spree/admin/prices_controller.rb
```diff
@@ -7,8 +7,14 @@ def index
         params[:q] ||= {}
 
         @search = @product.prices.accessible_by(current_ability, :index).ransack(params[:q])
-        @prices = @search.result
```
mamhoff: Changing instance variable names in controllers can be hazardous if people overrode those views. This has not been around for long at all, and I do not suspect many people have used this instance variable name, so I won't ask for a deprecation warning, but maybe a short note in the ChangeLog.

mamhoff:
I love this. 👍

isaacfreeman:
👍 

I'd like to see master variants displayed to users throughout Solidus admin alongside regular variants. That is, instead of saying "this is a product with no variants", we should be saying "this is a product with one variant – click here if you want to make another". I appreciate the original intent was to avoid complexity for beginners, but I've worked on too many sites where clients and former devs had apparently never encountered the concept of a variant at all, and instead extended the product model in cumbersome ways. Variants aren't hard to understand, and the concept only has to be learnt once to forestall a lot of expensive re-writing later.

</rant>

So... this is a positive step.

backend/app/controllers/spree/admin/prices_controller.rb
omnistegan: Thanks, noted in the ChangeLog.

Mandily:
This looks great! My only comment/concern would be pulling those table titles out of the merged table cell and making them a proper heading so that it's a bit easier to scan the page. We could also then add one for "Variant pricing" which could be referred to in the tooltip messaging instead of "the table below".

core/config/locales/en.yml
```diff
@@ -1115,6 +1115,8 @@ en:
     hints:
       spree/price:
         country: "This determines in what country the price is valid.<br/>Default: Any Country"
+        master_variant: "Changing these prices will not change variant prices below, but will be used to populate all new variants"
```
Mandily: Change "these" to "master variant" to be more explicit

core/config/locales/en.yml
```diff
@@ -1115,6 +1115,8 @@ en:
...
+        options: "These are the options used in the table below. They can be changed in the variants tab"
```
Mandily: Be more explicit with the name of the table. "These options are used to create variants in the variant table. They can be changed in the _variants tab_"

omnistegan: Yes, but not if it is displayed as a Tooltip because the Tooltip is only displayed on mouseover. We could instead place this text so it is always visible and attach the link.

omnistegan:
Here is what it looks like with fieldset legend style headers.

fieldsets

core/config/locales/en.yml
Mandily: The new bootstrap layouts have a place for this kind of information where we could also put links. Let's leave it in the tooltip for now, and when we convert over to the new layouts we can augment the text with a link in the sidebar.

Mandily:
While not a fan of that fieldset style, I do like the hierarchy it gives to the page and it's a better middle ground for us to work from when we update styles. 👍

jhawthorn:
Though this works better for products with variants, it's now extremely confusing on products with only a master variant.



Maybe we need to special-case that to only show the master variant prices and lose the heading/hint.

CHANGELOG.md
```diff
@@ -36,6 +36,9 @@
 
 *   The `lastname` field on `Address` is now optional. [#1369](https://github.com/solidusio/solidus/pull/1369)
 
+*   Change backend `prices_controller` instance variable `@prices` to `@master_prices` and
```
jhawthorn: This should probably mention what user-facing change was made. Maybe:

> The backend prices listing page now shows master prices separate from variant prices. This changes instance variable `@prices` to `@master_prices` and `@variant_prices`.

jhawthorn:
I don't understand the purpose of the Options table here.

I think the option types relate only loosely to Prices (in that the options are what variants use, and variants have prices) and is going to be more confusing to users than helpful. New variants aren't created from this page.

I think that table would make sense on the Variants page, but not here.

mamhoff:
@jhawthorn Yep, I have to agree. The user needs to know which option types are selected for a particular variant she's pricing, but she doesn't create new variants here, which she would need that table for.

omnistegan:
@jhawthorn @mamhoff The options table was suggested to help the user quickly reference if they have prices set for all the relevant variants.
A select box of variants is available to the user when creating a new price, so options are not very useful to display on the prices index page.

omnistegan:
A couple of updated screenshots with options table removed, and only showing variants table if the product has variants.

Without variants:
no_variants

With variants:
with_variants

Issue #1472, opened 2016-09-27
Link to Store Credit on Order Payment Page


gmacdougall:
Currently, when a store credit is consumed in a payment, there is no link to that store credit record in either the order payments show page or the order payments index page.

This makes it difficult to determine which particular store credit was consumed, the history of that store credit, etc.

# Example of page where Link to Store Credit record should be

![image](https://cloud.githubusercontent.com/assets/764390/18875633/8f621f0c-8494-11e6-9930-817fc4cc37b0.png)

PR #1513, opened 2016-10-10
Ledger entries for store credits.

peterberkenbosch:
The `store_credit_event` is recording an audit trail to provide some insight in the lifeline of a `store_credit`. However, it would be nice to have an actual ledger for the store credits, that will focus only on the events that will have a financial consequence to the store owner.

This PR introduces the `store_credit_legder_entry` that will provide historically consistent reporting of store credit liabilities.

### DIFF
--- a/core/app/models/spree/store_credit.rb
+++ b/core/app/models/spree/store_credit.rb
@@ -17,6 +17,7 @@ class Spree::StoreCredit < Spree::Base
   belongs_to :category, class_name: "Spree::StoreCreditCategory"
   belongs_to :credit_type, class_name: 'Spree::StoreCreditType', foreign_key: 'type_id'
   has_many :store_credit_events
+  has_many :store_credit_ledger_entries
 
   validates_presence_of :user_id, :category_id, :type_id, :created_by_id, :currency
   validates_numericality_of :amount, { greater_than: 0 }
@@ -30,6 +31,7 @@ class Spree::StoreCredit < Spree::Base
   scope :order_by_priority, -> { includes(:credit_type).order('spree_store_credit_types.priority ASC') }
 
   after_save :store_event
+  after_create :add_init_ledger_entry
   before_validation :associate_credit_type
   before_validation :validate_category_unchanged, on: :update
   before_destroy :validate_no_amount_used
@@ -61,9 +63,14 @@ class Spree::StoreCredit < Spree::Base
         action_amount: amount,
         action_originator: options[:action_originator],
         action_authorization_code: authorization_code,
-
         amount_authorized: amount_authorized + amount
       })
+      Spree::StoreCreditLedgerEntry.debit(
+        self,
+        amount,
+        options[:action_originator],
+        Spree::StoreCreditLedgerEntry::PENDING
+      )
       authorization_code
     else
       errors.add(:base, Spree.t('store_credit.insufficient_authorized_amount'))
@@ -94,10 +101,26 @@ class Spree::StoreCredit < Spree::Base
           action_amount: amount,
           action_originator: options[:action_originator],
           action_authorization_code: authorization_code,
-
           amount_used: amount_used + amount,
           amount_authorized: amount_authorized - auth_event.amount
         })
+
+        Spree::StoreCreditLedgerEntry.debit(
+          self,
+          amount,
+          options[:action_originator],
+          Spree::StoreCreditLedgerEntry::LIABILITY
+        )
+
+        # we need to credit the amount on the non-liabitly ledger to
+        # proper balance the amounts.
+        Spree::StoreCreditLedgerEntry.credit(
+          self,
+          amount,
+          options[:action_originator],
+          Spree::StoreCreditLedgerEntry::PENDING
+        )
+
         authorization_code
       end
     else
@@ -113,9 +136,15 @@ class Spree::StoreCredit < Spree::Base
         action_amount: auth_event.amount,
         action_authorization_code: authorization_code,
         action_originator: options[:action_originator],
-
         amount_authorized: amount_authorized - auth_event.amount
       })
+
+      Spree::StoreCreditLedgerEntry.credit(
+        self,
+        auth_event.amount,
+        options[:action_originator],
+        Spree::StoreCreditLedgerEntry::PENDING
+      )
       true
     else
       errors.add(:base, Spree.t('store_credit.unable_to_void', auth_code: authorization_code))
@@ -184,7 +213,17 @@ class Spree::StoreCredit < Spree::Base
     self.action = ADJUSTMENT_ACTION
     self.update_reason = reason
     self.action_originator = user_performing_update
-    save
+    if save
+      if action_amount > 0
+        Spree::StoreCreditLedgerEntry.credit(self, action_amount, user_performing_update)
+        true
+      elsif action_amount < 0
+        Spree::StoreCreditLedgerEntry.debit(self, action_amount, user_performing_update)
+        true
+      end
+    else
+      false
+    end
   end
 
   def invalidate(reason, user_performing_invalidation)
@@ -193,11 +232,22 @@ class Spree::StoreCredit < Spree::Base
       self.update_reason = reason
       self.action_originator = user_performing_invalidation
       self.invalidated_at = Time.current
-      save
+      self.action_amount = liability_balance
+      if save
+        Spree::StoreCreditLedgerEntry.debit(self, liability_balance, user_performing_invalidation)
+      end
     else
       errors.add(:invalidated_at, Spree.t("store_credit.errors.cannot_invalidate_uncaptured_authorization"))
       return false
     end
+  end
+
+  def liability_balance
+    Spree::StoreCreditLedgerEntry.liability_balance(self)
+  end
+
+  def balance
+    Spree::StoreCreditLedgerEntry.balance(self)
   end
 
   class << self
@@ -211,15 +261,16 @@ class Spree::StoreCredit < Spree::Base
   def create_credit_record(amount, action_attributes = {})
     # Setting credit_to_new_allocation to true will create a new allocation anytime #credit is called
     # If it is not set, it will update the store credit's amount in place
-    credit = if Spree::Config[:credit_to_new_allocation]
-      Spree::StoreCredit.new(create_credit_record_params(amount))
+    if Spree::Config[:credit_to_new_allocation]
+      credit = Spree::StoreCredit.new(create_credit_record_params(amount))
+      credit.assign_attributes(action_attributes)
+      credit.save!
     else
       self.amount_used = amount_used - amount
-      self
+      assign_attributes(action_attributes)
+      save!
+      Spree::StoreCreditLedgerEntry.credit(self, amount, action_attributes[:action_originator])
     end
-
-    credit.assign_attributes(action_attributes)
-    credit.save!
   end
 
   def create_credit_record_params(amount)
@@ -254,6 +305,10 @@ class Spree::StoreCredit < Spree::Base
       originator: action_originator,
       update_reason: update_reason
     })
+  end
+
+  def add_init_ledger_entry
+    Spree::StoreCreditLedgerEntry.credit(self, amount, action_originator)
   end
 
   def amount_used_less_than_or_equal_to_amount
--- a/core/app/models/spree/store_credit_ledger_entry.rb
+++ b/core/app/models/spree/store_credit_ledger_entry.rb
@@ -1,0 +1,86 @@
+# Financial transaction entry for a specific `store_credit`
+# it will keep track of the liability and pending transactions.
+#
+# It will follow the debit and credit rules as described [here](https://en.wikipedia.org/wiki/Debits_and_credits)
+# so all debit calls will lower the liability and credit calls will add
+# an amount.
+#
+module Spree
+  class StoreCreditLedgerEntry < Spree::Base
+    PENDING = :pending
+    LIABILITY = :liability
+
+    belongs_to :store_credit
+    belongs_to :originator, polymorphic: true
+
+    scope :chronological, -> { order(:created_at) }
+    scope :reverse_chronological, -> { order(created_at: :desc) }
+    scope :liability, -> { where(liability: true) }
+    scope :pending, -> { where(liability: false) }
+
+    delegate :currency, to: :store_credit
+
+
+    # Will add the amount to the `store_credit`, `credit` the ledger.
+    # this amount will always be positive, when there is a need to remove
+    # an amount from the `store_credit` you will have to call the [#debit]
+    #
+    # @param store_credit [Spree::StoreCredit] the store credit for this credit call
+    # @param amount [BigDecimal] the amount for the credit, should always be a positive number.
+    # @param originator [Object] polymorphic origin that triggered this credit call
+    # @param liability_type [Symbol] is either an actual liability change, or a pending transaction
+    # @return [Spree::StoreCreditLedgerEntry] the created ledger entry, or an exception when not succesful
+    def self.credit(store_credit, amount, originator, liability_type=LIABILITY)
+      create!(
+        {
+          store_credit: store_credit,
+          amount: amount,
+          originator: originator,
+          liability: liability_type == LIABILITY
+        }
+      )
+    end
+
+    # Will remove the amount from the `store_credit`, `debit` the ledger.
+    # this amount will always be stored negative, when there is a need to add
+    # an amount to the `store_credit` you will have to call the [#credit]
+    #
+    # @param store_credit [Spree::StoreCredit] the store credit for this credit call
+    # @param amount [BigDecimal] the amount for the credit, should always be a positive number.
+    # @param originator [Object] polymorphic origin that triggered this credit call
+    # @param liability_type [Symbol] is either an actual liability change, or a pending transaction
+    # @return [Spree::StoreCreditLedgerEntry] the created ledger entry, or an exception when not succesful
+    def self.debit(store_credit, amount, originator, liability_type=LIABILITY)
+      # make sure debit amounts are stored as negative number.
+      amount = -amount if amount > 0
+      create!(
+        {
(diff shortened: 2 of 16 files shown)


core/app/models/spree/store_credit.rb
```diff
@@ -30,6 +31,7 @@ class Spree::StoreCredit < Spree::Base
   scope :order_by_priority, -> { includes(:credit_type).order('spree_store_credit_types.priority ASC') }
 
   after_save :store_event
+  after_save :create_ledger_entry
```
gmacdougall: I'm not super keen on building everything around callbacks. Could we try extracting the logic to some specific methods so I could do the following:

```Ruby
store_credit.credit(100, admin_user_who_granted_this)
store_credit.debit(50, payment)
```

Then those methods could create the ledger entry and modify the current balance.

core/db/migrate/20161006211441_create_spree_store_credit_ledger_entry.rb
```diff
@@ -0,0 +1,9 @@
+class CreateSpreeStoreCreditLedgerEntry < ActiveRecord::Migration[5.0]
+  def change
+    create_table :spree_store_credit_ledger_entries do |t|
```
gmacdougall: I would also like to see a source here in order to track where the action came from (for example, a payment or admin user).

core/app/models/spree/store_credit.rb
peterberkenbosch: I like that! More clear what the intention is. On it.

core/db/migrate/20161006211441_create_spree_store_credit_ledger_entry.rb
peterberkenbosch: yes.

core/spec/models/spree/store_credit_spec.rb
```diff
@@ -868,6 +957,25 @@
         expect { subject }.to change { store_credit.store_credit_events.where(action: Spree::StoreCredit::INVALIDATE_ACTION).count }.from(0).to(1)
       end
 
+      it "adds an entry to the ledger" do
+        expect { subject }.to change { Spree::StoreCreditLedgerEntry.count }.by(1)
```
gmacdougall: Rather than checking `Spree::StoreCreditLedgerEntry.count` I think it makes more sense to check: `store_credit.store_credit_ledger_entries.count`

core/spec/models/spree/store_credit_spec.rb
```diff
@@ -868,6 +957,25 @@
...
+      end
+
+      it "will made the ledger balance be 0.0" do
+        subject
+        expect(store_credit.ledger_balance).to eql 0.0
+      end
+
+      it "records the originator on the ledger entry" do
+        subject
+        expect(Spree::StoreCreditLedgerEntry.last.originator).to eq invalidation_user
```
gmacdougall: Same here:

```
expect(store_credit.store_credit_ledger_entries.last.originator)
```

core/app/models/spree/store_credit.rb
```diff
@@ -184,7 +189,17 @@ def update_amount(amount, reason, user_performing_update)
     self.action = ADJUSTMENT_ACTION
     self.update_reason = reason
     self.action_originator = user_performing_update
-    save
+    if save
```
gmacdougall: I still don't think this is quite what I'm looking for. I would prefer something like:

```Ruby
if action_amount > 0
  credit(action_amount, user_performing_update)
else
  debit(action_amount, user_performing_update)
else
```

and move to a system where that is what's happening. Then the credit and debit methods are the ones which alter the ledger.

peterberkenbosch: that is what happens here, did not use `credit` method though, since there is already a `credit` method here.

core/app/models/spree/store_credit.rb
```diff
@@ -193,13 +219,18 @@ def invalidate(reason, user_performing_invalidation)
       self.update_reason = reason
       self.action_originator = user_performing_invalidation
       self.invalidated_at = Time.current
-      save
+      self.action_amount = ledger_balance
+      debit(ledger_balance, { action_originator: user_performing_invalidation }) if save
```
jhawthorn: I don't doing work (like save) in a suffix `if` statement. I'd prefer a regular if statement to ensure it is noticed

```
if save
  debit(ledger_balance, { action_originator: user_performing_invalidation })
end
```

core/app/models/spree/store_credit.rb
```diff
@@ -123,25 +128,35 @@ def void(authorization_code, options = {})
     end
   end
 
-  def credit(amount, authorization_code, order_currency, options = {})
-    # Find the amount related to this authorization_code in order to add the store credit back
-    capture_event = store_credit_events.find_by(action: CAPTURE_ACTION, authorization_code: authorization_code)
+  def debit(amount, options = {})
+    # make sure debit amounts are stored as negative number
+    amount = -1 * amount unless amount < 0
```
jhawthorn: I found this just slightly tough to parse (have to flip the sign because of the unless). I'd prefer

``` ruby
amount = -amount if amount > 0
```

core/spec/models/spree/store_credit_ledger_entry_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+describe Spree::StoreCreditLedgerEntry do
+  describe "#generate_ledger_entries_for" do
+    context "store credit with just an allocation action" do
+      it "will create one ledger entry" do
```
jhawthorn: Missing spec

core/db/migrate/20161012151358_opening_ledger_entries_for_store_credits.rb
```diff
@@ -0,0 +1,7 @@
+class OpeningLedgerEntriesForStoreCredits < ActiveRecord::Migration[5.0]
+  def change
+    Spree::StoreCredit.find_each do |store_credit|
+      Spree::StoreCreditLedgerEntry.generate_opening_ledger_entry_for(store_credit)
```
jhawthorn: Is it possible to do this using raw SQL or using AR models defined directly in this migration?

Doing it this way may cause issues in the future if the behaviour of `StoreCredit` or `StoreCreditLedgerEntry.generate_opening_ledger_entry_for` changes.

core/app/models/spree/store_credit.rb
peterberkenbosch: yes, way better.. updated!

core/app/models/spree/store_credit.rb
peterberkenbosch: improved readability for sure! Updated.

core/db/migrate/20161012151358_opening_ledger_entries_for_store_credits.rb
peterberkenbosch: Yes, wanted to have something that was easily testable though. Looking into options here.

core/spec/models/spree/store_credit_ledger_entry_spec.rb
peterberkenbosch: yup. WIP atm since I'm looking at how to best make that migration happen.

core/db/migrate/20161012151358_opening_ledger_entries_for_store_credits.rb
peterberkenbosch: Updated the migration using a rake task and added the spec for that rake task.

core/db/migrate/20161012151358_opening_ledger_entries_for_store_credits.rb
```diff
@@ -0,0 +1,12 @@
+class OpeningLedgerEntriesForStoreCredits < ActiveRecord::Migration[5.0]
+  # Prevent everything from running in one giant transaction in postrgres.
+  disable_ddl_transaction!
```
jhawthorn: Is this necessary? I would rather have a transaction unless there is a reason not to.

core/lib/tasks/migrations/create_ledger_entries_for_store_credits.rake
```diff
@@ -0,0 +1,20 @@
+namespace :solidus do
+  namespace :migrations do
+    namespace :create_ledger_entries_for_store_credits do
+      task up: :environment do
+        print 'Creating opening ledger entries for store credits...'
```
jhawthorn: I would rather not print anything in our migrations (since this will show up in each `rake test_app`, `rake sandbox`, and `rspec` run). Though I believe `say` in an actual migration will be silent if `VERBOSE=false` or something.

core/lib/tasks/migrations/create_ledger_entries_for_store_credits.rake
```diff
@@ -0,0 +1,20 @@
...
+        Spree::StoreCredit.find_each do |store_credit|
+          current_balance_amount = store_credit.amount_remaining + store_credit.amount_authorized
+          store_credit.store_credit_ledger_entries.create!({
+            amount: current_balance_amount
+          })
```
jhawthorn: I would still prefer that this didn't use the standard AR models as that will cause problems if the implementation of `amount_remaining` ever changes. It seems really likely that that method will chnage, given that it might be better implemented in terms of summing these ledger entries.

Instead I'd like to see raw SQL which shouldn't cause any issues in the future.

``` ruby
execute <<-SQL
INSERT INTO spree_store_credit_ledger_entries
  (store_credit_id, amount, created_at, updated_at)
  SELECT id, amount - amount_used, '#{now}', '#{now}'
  FROM spree_store_credits
  WHERE invalidated_at IS NULL
SQL
execute <<-SQL
INSERT INTO spree_store_credit_ledger_entries
  (store_credit_id, amount, created_at, updated_at)
  SELECT id, 0, '#{now}', '#{now}'
  FROM spree_store_credits
  WHERE invalidated_at IS NOT NULL
SQL
```

gmacdougall:
Peter and I discussed this a little bit more, and there 's still some important information that we want to track regarding auth/void on the ledger. These are important events that affect the amount of available store credit and we want to have an appropriate history of them.

We also want to make sure that the public interface makes sense and you can't break assumptions that we have by using the public interface (methods like credit/debit).

core/lib/tasks/migrations/create_ledger_entries_for_store_credits.rake
peterberkenbosch: Thanks! I can confirm that `say` will indeed be silent with `VERBOSE=false`

core/lib/tasks/migrations/create_ledger_entries_for_store_credits.rake
peterberkenbosch: Yes, for sure! Updated that!

peterberkenbosch:
@gmacdougall @jhawthorn refactored a bunch! Ready for some more feedback. Thanks!

core/app/models/spree/store_credit_ledger_entry.rb
```diff
@@ -0,0 +1,49 @@
...
+          amount: amount,
+          originator: originator,
+          liability: liability_type == LIABILITY
+        }
+      )
+    end
+
+    def self.debit(store_credit, amount, originator, liability_type=LIABILITY)
+      # make sure debit amounts are stored as negative number
+      amount = -amount if amount > 0
```
bbuchalter: I think I might prefer to raise an exception here than this. I want consumers of this API to understand what's expected and not rely on magic.

peterberkenbosch: Yes, that and with documentation on the method. Magic vs Convenience, I will make sure to get the documentation on the method done first. :)

peterberkenbosch:
First option is solved with #1472 I guess. Will make this happen here and refer to the issue.

core/app/models/spree/store_credit_ledger_entry.rb
```diff
@@ -0,0 +1,74 @@
...
+
+
+    # Will add the amount to the `store_credit`, `credit` the ledger.
+    # this amount will always be positive, when there is a need to remove
+    # an amount from the `store_credit` you will have to call the [#debit]
+    #
+    # @param store_credit [Spree::StoreCredit] the store credit for this credit call
+    # @param amount [BigDecimal] the amount for the credit, should always be a positive number.
+    # @param originator [Object] polymorphic origin that triggered this credit call
+    # @param liability_type is either an actual liability change, or a pending transaction
```
gmacdougall:
```Ruby
# @param liability_type [String] ...
```

core/app/models/spree/store_credit_ledger_entry.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    # Will remove the amount from the `store_credit`, `debit` the ledger.
+    # this amount will always be stored negative, when there is a need to add
+    # an amount to the `store_credit` you will have to call the [#credit]
+    #
+    # @param store_credit [Spree::StoreCredit] the store credit for this credit call
+    # @param amount [BigDecimal] the amount for the credit, should always be a positive number.
+    # @param originator [Object] polymorphic origin that triggered this credit call
+    # @param liability_type is either an actual liability change, or a pending transaction
```
gmacdougall: Same as above

core/db/migrate/20161006211441_create_spree_store_credit_ledger_entry.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.decimal :amount, precision: 8, scale: 2, default: 0.0, null: false
+      t.references :store_credit, index: true
+      index_name = "idx_spree_sc_ledger_entries_on_orig_type_and_orig_id"
+      t.references :originator, polymorphic: true, index: { name: index_name }
+      t.boolean :liability, default: true
```
gmacdougall: `null: false` as well to prevent tri-state booleans.

core/app/models/spree/store_credit_ledger_entry.rb
```diff
@@ -0,0 +1,74 @@
...
+# it will keep track of the liability and pending transactions.
+#
+# It will follow the debit and credit rules as described [here](https://en.wikipedia.org/wiki/Debits_and_credits)
+# so all debit calls will lower the liability and credit calls will add
+# an amount.
+#
+module Spree
+  class StoreCreditLedgerEntry < Spree::Base
+    PENDING = 'pending'
+    LIABILITY = 'liability'
```
gmacdougall: We can probably use symbols here instead of strings. Not a big deal.

bbuchalter:
Given the massive number of commits, is a rebase in order?

bbuchalter:
A CHANGELOG entry is also in order.

peterberkenbosch:
yes @bbuchalter, rebase is definitely in order here, will be doing that when all lights are green :) Changelog is incoming as well.

peterberkenbosch:
This has updated and rebased. Ready for merge when approved. //cc @gmacdougall @jhawthorn

peterberkenbosch:
done

core/app/models/spree/store_credit.rb
```diff
@@ -116,14 +127,21 @@ def void(authorization_code, options = {})
...
+      )
       true
     else
       errors.add(:base, Spree.t('store_credit.unable_to_void', auth_code: authorization_code))
       false
     end
   end
 
-  def credit(amount, authorization_code, order_currency, options = {})
+  def credit(amount, authorization_code = nil, order_currency = "USD", options = {})
```
mamhoff: I'm uneasy about the default currency being `USD` here. If this isn't passed in in a context where the currency is anything else, this should actually throw an error IMO.

core/db/migrate/20161012151358_opening_ledger_entries_for_store_credits.rb
```diff
@@ -0,0 +1,9 @@
+class OpeningLedgerEntriesForStoreCredits < ActiveRecord::Migration[5.0]
+  def up
+    Rake::Task["solidus:migrations:create_ledger_entries_for_store_credits:up"].invoke
```
mamhoff: Why are these migrations extracted to `rake` tasks? I did this for data migrations which depended on model methods, but in this case the rake task really only has SQL and almost no ruby dependency. I believe these should be directly inside this migration.

peterberkenbosch: I wanted to have a way to properly spec it, could not find a better way to do this.

bbuchalter:
One of the design goals of this ledger was to ensure that information was only ever appended to the ledger so there would be a consistent historical record of store credit activity. While I believe we've followed that principle here, perhaps we should enforce it by marking all store credit ledger entries read-only?

core/spec/models/spree/store_credit_spec.rb
```diff
@@ -867,7 +867,7 @@
       it "creates a store credit event for the invalidation" do
         expect { subject }.to change { store_credit.store_credit_events.where(action: Spree::StoreCredit::INVALIDATE_ACTION).count }.from(0).to(1)
       end
-
+
```
bbuchalter: Extra whitespace. Surprised hound didn't see this.

peterberkenbosch:
@gmacdougall damn, you're right! It's missing the counter balance entries!

gmacdougall:
Can you please tidy up your commit messages?

http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

peterberkenbosch:
https://github.com/solidusio/solidus/pull/1513/commits/ec40ee8f1fdc2ec7f5152db9d6db1b586daebf7b

Have to rebase the last one in there still. Waiting for last round of feedback

gmacdougall:
Brian, Peter and I discussed this earlier and don't see a significant advantage in making the records read only. The ledger entries are only created, and not updated in the code here and readonly can be bypassed.

peterberkenbosch:
@gmacdougall will provide an improved version for this, so closing this one now.

PR #1514, opened 2016-10-11
Fix Shipment#update_attributes_and_order - run full order updater

jordan-brough:
Tax charges aren't being updated when the shipping method changes
because some parts of OrderUpdater aren't being run in
`Shipment#update_attributes_and_order`.

Also fix an issue in order_factory around shipment taxes.

See individual commits.

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -296,25 +296,9 @@ module Spree
       if update_attributes params
         if params.key? :selected_shipping_rate_id
           # Changing the selected Shipping Rate won't update the cost (for now)
-          # so we persist the Shipment#cost before calculating order shipment
-          # total and updating payment state (given a change in shipment cost
-          # might change the Order#payment_state)
+          # so we persist the Shipment#cost before running `order.update!`
           update_amounts
-
-          order.updater.update_shipment_total
-          order.updater.update_payment_state
-
-          # Update shipment state only after order total is updated because it
-          # (via Order#paid?) affects the shipment state (YAY)
-          update_columns(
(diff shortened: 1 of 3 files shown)


jhawthorn:
What was the issue in the `order_factory`? Can we test it in a spec?

jordan-brough:
@jhawthorn see the commit message on the [first commit](https://github.com/solidusio/solidus/pull/1514/commits/d278e970a37fc626c43b68f6326df3cd9526e9af):

>  Fix order_factory to create tax charge after shipment
> Otherwise the shipment doesn't get a tax adjustment set up.

Yah I can create a spec for that. Will do.

jordan-brough:
@jhawthorn I've added a spec for the factory

core/lib/spree/testing_support/factories/order_factory.rb
```diff
@@ -45,11 +45,11 @@
           create(:line_item, attributes)
         end
         order.line_items.reload
-        order.create_tax_charge!
 
         create(:shipment, order: order, cost: evaluator.shipment_cost, shipping_method: evaluator.shipping_method, stock_location: evaluator.stock_location)
         order.shipments.reload
 
+        order.create_tax_charge!
```
jhawthorn: I think this is no longer necessary thanks to your work in #1479

jhawthorn:
@jordan-brough I rebased and removed the first commit, which was unnecessary as of #1479 :tada:

mtomov:
I also think that this simplification is brilliant, and @gmacdougall said it all very well. Thanks a lot!

PR #1516, opened 2016-10-11
Do not require lastname in frontend address form

stewart:
Last names are now [optional][1] on Address records, and are no longer
generated by the default `:address` factory. As a result, this field
should no longer be considered required by solidus-frontend.

[1]: https://github.com/solidusio/solidus/pull/1369


jhawthorn:
Merging this earlier than I normally would because it fixes the `solidus_auth_devise` spec failures against Solidus master

jordan-brough:
Thanks @stewart!

PR #1521, opened 2016-10-13
Closed parenthesis on admin breadcrumb line

mdamian322:
This caused an error when creating a new reimbursement. The reimbursement edit page view is what failed.


tvdeyen:
Thanks!

bbuchalter:
When we catch errors like this, shouldn't we be added regression specs?

PR #1523, opened 2016-10-13
Simplify and improve stock locations form

jhawthorn:
This removes the cluttered layout of the existing form in favour of just having the fields from top to bottom. Fields are grouped into 3 logical sections.

This also adds informational hints to the settings checkboxes.

### DIFF
--- a/backend/app/views/spree/admin/stock_locations/_form.html.erb
+++ b/backend/app/views/spree/admin/stock_locations/_form.html.erb
@@ -1,105 +1,96 @@
-<div data-hook="admin_stock_locations_form_fields" class="row">
-  <div class="col-7" data-hook="stock_location_names">
-    <div data-hook="stock_location_name">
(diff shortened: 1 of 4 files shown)


tvdeyen:
Simpler it is. But I'm not sure if this is also optimal. But it's a good start to put this form either in an overlay/modal or use the space for some additional info. 

👍 For the hints

I would like to here what @Mandily has to say

Mandily:
I also like the idea of the tooltips on the checkboxes. I don't completely understand what all of these are for, but I wonder if there's a way to group them up into something more meaningful for users? Do any of you have a pulse on that?

The stock locations page in the new layout would look something like this
stock locations

I know @omnistegan has been working on implementing the new layouts over the last couple of weeks. I wonder if we shouldn't wait at this point for those new layouts instead of reconfiguring stuff in skeleton.

PR #1524, opened 2016-10-14
Add promotion code batch

vladstoick:
We (lostmy.name) have found that the current approach for creating promo codes doesn't work for creating thousands of promo codes as the request will time out by the time all codes are created. Therefore, we introduced promotion code batches to move the process of creating codes in a `ActiveJob`

This PR introduces a new model called `Spree::PromotionCodeBatch` which belongs to a `Spree::Promotion`. Each `PromotionBatch` has a `base_code` and a `number_of_codes` so it describes a batch of promo codes for a promotion.`PromotionBatch` also has an `email` field so we can send an email when all promo codes are created.

Generating the codes is done within a worker `Spree::PromotionCodeBatchJob` which calls `Spree::PromotionCode::BatchBuilder`. `PromotionCode::BatchBuilder` is very similar to `Spree::PromotionCode::CodeBuilder`) with a few optimisations (uses batches instead of creating all codes at once, uses more characters). `BatchBuilder` also updates `PromotionCodeBatch`'s `number_of_codes_processed` so we can display a status in the view.


bbuchalter:
Thanks for putting this forward. This is a regular pain point for our (tommyjohn.com) admins.

jordan-brough:
This will be a fantastic improvement for shops that make lots of codes. 👍 

Should we add a `spree_promotion_codes.promotion_code_batch_id` field?  E.g. then admins could download all the CSV for a particular batch.

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -0,0 +1,64 @@
...
+      codes_for_current_batch = Set.new
+      codes_to_generate = [self.class.batch_size, number_of_codes - i * batch_size].min
+
+      while codes_for_current_batch.size < codes_to_generate
+        new_codes = Array.new(codes_to_generate) { generate_random_code }.to_set
+        codes_for_current_batch += get_unique_codes(new_codes)
+      end
+
+      codes_for_current_batch.map do |code|
+        promotion.codes.build(value: code)
```
jordan-brough: Should we perhaps use `Spree::PromotionCode.create!` here instead of `promotion.codes.build ... promotion.save!`?  That would avoid loading a million objects into memory if you're generating a million codes.

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      promotion.save!
+
+      promotion_code_batch.update!(
```
jordan-brough: If we add a `spree_promotion_codes.promotion_code_batch_id` column then we could eliminate this yes?

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -0,0 +1,64 @@
...
+  def generate_random_code
+    suffix = Array.new(self.class.random_code_length) do
+      sample_characters.sample
+    end.join
+
+    "#{base_code}_#{suffix}"
+  end
+
+  def sample_characters
+    @sample_characters ||= ('a'..'z').to_a + (2..9).to_a.map(&:to_s)
```
jordan-brough: Should we make this a `class_attribute` as well, in case people want to customize what characters are used for codes?

backend/app/controllers/spree/admin/promotions_controller.rb
```diff
@@ -2,35 +2,24 @@ module Spree
...
       helper 'spree/promotion_rules'
 
       def create
-        @promotion_builder = Spree::PromotionBuilder.new(
-          permitted_promo_builder_params.merge(user: try_spree_current_user),
-          permitted_resource_params
-        )
-        @promotion = @promotion_builder.promotion
+        @promotion = Spree::Promotion.new(permitted_resource_params)
+        @promotion.codes.new(value: params[:single_code]) if params[:single_code].present?
```
jordan-brough: I like how this separates creating a single code (via this controller) from creating a batch of codes (via the batch controller).  There's probably even further improvement we can do along the lines of creating single codes, but it seems like that could be a separate PR.  Possibly part of #1509?

backend/app/views/spree/admin/promotion_code_batches/index.html.erb
```diff
@@ -0,0 +1,40 @@
...
+          <td><%= promotion_code_batch.base_code %></td>
+          <td><%= promotion_code_batch.number_of_codes %></td>
+          <td><%= promotion_code_batch.status %></td>
+          <td><%= promotion_code_batch.email %></td>
+        </tr>
+      <% end %>
+    </tbody>
+  </table>
+<% else %>
+  <div class="alpha twelve columns no-objects-found">
```
jhawthorn: We aren't using the skeleton classes anymore. This can juts be 

```
<div class="no-objects-found">
</div>

File: backend/app/views/spree/admin/promotion_code_batches/new.html.erb [hunk 6]
```diff
@@ -0,0 +1,25 @@
+<% admin_breadcrumb(link_to plural_resource_name(Spree::Promotion), spree.admin_promotions_path) %>
+<% admin_breadcrumb(link_to @promotion.name, spree.admin_promotion_path(@promotion.id)) %>
+<% admin_breadcrumb(plural_resource_name(Spree::PromotionCodeBatch)) %>
+
+<%= form_tag admin_promotion_promotion_code_batches_path do %>
+  <%= fields_for :promotion_code_batch do |f| %>
```
jhawthorn: This two lines can _probably_ be combined into just a `form_for`

backend/app/views/spree/admin/promotions/_form.html.erb
```diff
@@ -8,17 +8,10 @@
...
-          <% end %>
-
-          <%= bf.field_container :number_of_codes do %>
-            <%= bf.label :number_of_codes %>
-            <%= bf.text_field :number_of_codes, :readonly => !@promotion.new_record?, class: 'fullwidth' %>
-          <% end %>
-        <% end %>
+        <div class="field" id="promotion_single_code_field">
+          <%= label_tag :single_code %>
+          <%= text_field_tag :single_code, @promotion.codes.first.try(:value), :readonly => !@promotion.new_record?, :class => 'fullwidth' %>
```
jhawthorn: Please use `try!` and not `try`.

core/app/models/spree/promotion_code_batch.rb
```diff
@@ -0,0 +1,26 @@
...
+      if error.present?
+        error
+      elsif number_of_codes == number_of_codes_processed
+        "All codes created"
+      else
+        "Processing: #{number_of_codes_processed} / #{number_of_codes}"
+      end
+    end
+
+    def build
```
jhawthorn: I would prefer a name other than `build`, which will be confusing since ActiveRecord uses that name a lot.

Any of `build_codes`, `create_codes`, `process` would be fine.

vladstoick:
@jordan-brough I've added a `spree_promotion_codes.promotion_code_batch_id` field and also a download button so you can download codes for a single batch

core/app/models/spree/promotion_code_batch.rb
```diff
@@ -0,0 +1,31 @@
+module Spree
+  class PromotionCodeBatch < ActiveRecord::Base
+    belongs_to :promotion, class_name: "Spree::Promotion"
+    has_many :promotion_codes, class_name: "Spree::PromotionCode", dependent: :destroy
+
+    validates :number_of_codes, numericality: { greater_than: 0 }
+    validates_presence_of :base_code, :number_of_codes
+
+    def status
+      if error.present?
```
cbrunsdon: Can you return some kind of status here instead? something with a symbol or something. so we can use locales to display the actual text. When we hardcode english like this, people that speak weird languages become sad and complain to me in private messages.

backend/config/routes.rb
```diff
@@ -13,6 +13,9 @@
       resources :promotion_rules
       resources :promotion_actions
       resources :promotion_codes, only: [:index]
+      resources :promotion_code_batches, only: [:index, :new, :create] do
+        get '/download', to: "promotion_code_batches#download"
```
cbrunsdon: can we limit this to just respond to the :csv format?

core/app/mailers/spree/promotion_code_batch_mailer.rb
```diff
@@ -0,0 +1,22 @@
+module Spree
+  class PromotionCodeBatchMailer < ApplicationMailer
+    def promotion_code_batch_finished(promotion_code_batch)
+      mail(
+        to: promotion_code_batch.email,
+        body: "All #{promotion_code_batch.number_of_codes} codes are created now",
```
cbrunsdon: could you push these to locales too please

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -0,0 +1,63 @@
...
+        codes_for_current_batch += get_unique_codes(new_codes)
+      end
+
+      codes_for_current_batch.map do |value|
+        promotion.codes.build(value: value, promotion_code_batch: promotion_code_batch)
+      end
+
+      promotion.save!
+
+      promotion.reload
```
cbrunsdon: whats the reason this is getting reloaded here?

core/spec/models/spree/promotion_code/batch_builder_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+
+      it "updated the error on promotion batch" do
+        expect(promotion_batch.reload.error).to eq("#<RuntimeError: Error>")
+      end
+
+      it "sends an email email" do
+        expect(Spree::PromotionCodeBatchMailer)
+          .to have_received(:promotion_code_batch_errored)
+        expect(delivery).to have_received(:deliver_now)
```
cbrunsdon: can we cut down the stubbing in the mailers and instead look at the test deliveries to see if they contain the content we're expecting?

core/spec/models/spree/promotion_code/batch_builder_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      it "update the promotion codes count for the batch" do
+        expect(promotion_batch.promotion_codes.count).to eq(number_of_codes)
+      end
+
+      it "builds the correct number of codes" do
+        expect(subject.promotion.codes.size).to eq(number_of_codes)
+      end
+
+      it "builds codes with distinct values" do
+        expect(subject.promotion.codes.map(&:value).uniq.size).to eq(number_of_codes)
```
cbrunsdon: can you hardcode number_of_codes for all these tests? its more clear to asset something is equal to a specific thing rather than a variable.

core/spec/models/spree/promotion_code_batch_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      expect(Spree::PromotionCodeBatchJob)
+        .to have_received(:perform_later)
+        .with(subject)
+    end
+
+    context "with a batch that already started" do
+      before do
+        allow(subject).to receive_message_chain("promotion_codes.count")
+          .and_return(1)
```
cbrunsdon: can you kill the stubbing in this? I don't think its too costly to create a promo batch with a code to avoid the message chain

core/spec/models/spree/promotion_code_batch_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  describe "#process" do
+    it "should call the worker" do
+      allow(Spree::PromotionCodeBatchJob).to receive(:perform_later)
+
+      subject.process
+
+      expect(Spree::PromotionCodeBatchJob)
+        .to have_received(:perform_later)
+        .with(subject)
```
cbrunsdon: is this the same thing that was stubbed above? is there a way to kill those two things?

core/spec/models/spree/promotion_code/batch_builder_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      it "builds codes with the same base prefix" do
+        values = subject.promotion.codes.map(&:value)
+        expect(values.all? { |val| val.starts_with?("#{base_code}_") }).to be true
+      end
+
+
+      it "sends a success email" do
+        expect(Spree::PromotionCodeBatchMailer)
+          .to have_received(:promotion_code_batch_finished)
+        expect(delivery).to have_received(:deliver_now)
```
tvdeyen: Please expect `ActionMailer::Base.deliveries` to be present instead of stubbing implementation details. Thanks

core/spec/models/spree/promotion_code/batch_builder_spec.rb
tvdeyen: A great start in email testing are the Rails guides http://guides.rubyonrails.org/testing.html#testing-your-mailers

backend/app/views/spree/admin/promotion_code_batches/index.html.erb
```diff
@@ -0,0 +1,51 @@
...
+      <%= button_link_to Spree.t(:new_promotion_code_batch), new_object_url %>
+    <% end %>
+  </li>
+<% end %>
+
+<% if @promotion_code_batches.any? %>
+  <table>
+    <thead>
+      <tr>
+        <th>Base code</th>
```
tvdeyen: Please don't hardcore English in table headers, always use the model attributes translations.

backend/app/views/spree/admin/promotion_code_batches/index.html.erb
```diff
@@ -0,0 +1,51 @@
...
+        <th>Email</th>
+      </tr>
+    </thead>
+    <tbody>
+      <% @promotion_code_batches.each do |promotion_code_batch| %>
+        <tr>
+          <td><%= promotion_code_batch.base_code %></td>
+          <td><%= promotion_code_batch.number_of_codes %></td>
+          <td>
+            <%= promotion_code_batch.status %>.
```
tvdeyen: As Clarke already mentioned, please return a symbol as status and pass the status here to `Spree.t` with a meaningful scope (ie. `promotion_code_batches.status`, so this can be translated in foreign languages.

core/app/mailers/spree/promotion_code_batch_mailer.rb
vladstoick: https://github.com/solidusio/solidus/pull/1524#discussion_r83457606

vladstoick:
@cbrunsdon, tvdeyen: I think I went over everything you mentioned
- Use I18n.t everywhere
- Fix mail specs: I am now using `email = ActionMailer::Base.deliveries.last` to check that the email has the correct details.
- `download` route now has a default format of csv
- Fixed various where too many things were specs stubbed

Can you have a look over this again when you have time?

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -0,0 +1,63 @@
+class ::Spree::PromotionCode::BatchBuilder
+  attr_reader :promotion_code_batch
+  delegate :promotion, :number_of_codes, :base_code, to: :promotion_code_batch
+
+  class_attribute :random_code_length, :batch_size, :sample_characters
+  self.random_code_length = 8
```
jhawthorn: I would prefer we kept the old value here `6`, just to avoid surprise changes to users.

vladstoick: Done :)

backend/app/controllers/spree/admin/promotion_code_batches_controller.rb
```diff
@@ -0,0 +1,27 @@
+module Spree
+  module Admin
+    class PromotionCodeBatchesController < ResourceController
+      belongs_to 'spree/promotion'
+
+      create.after :build_promotion_code_batch
+
+      def download
+        require 'csv'
+        @promotion_code_batch = Spree::PromotionCodeBatch.find_by_id(params[:promotion_code_batch_id])
```
jordan-brough: Nitpick: Might as well use `find` here, rather than `find_by_id` so that it raises with a clear error message if the id is invalid?

core/app/models/spree/promotion_code/batch_builder.rb
jordan-brough: Might want to leave a code comment to that effect?

vladstoick:
@tvdeyen, @cbrunsdon did you have a chance to look over this again?

backend/app/controllers/spree/admin/promotion_code_batches_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        @promotion_code_batch = Spree::PromotionCodeBatch.find(
+          id: params[:promotion_code_batch_id]
+        )
+
+        respond_to do |format|
+          format.csv do
+            filename = "promotion-code-batch-list-#{@promotion_code_batch.id}.csv"
+            headers["Content-Type"] = "text/csv"
+            headers["Content-disposition"] = "attachment; filename=\"#{filename}\""
```
tvdeyen: Why not use [Rails' `send_file`](rails doc send_file) here? Then you can spare to set headers, Rails does this for you.

backend/app/views/spree/admin/promotion_code_batches/index.html.erb
```diff
@@ -0,0 +1,65 @@
...
+      <%= button_link_to Spree.t(:new_promotion_code_batch), new_object_url %>
+    <% end %>
+  </li>
+<% end %>
+
+<% if @promotion_code_batches.any? %>
+  <table>
+    <thead>
+      <tr>
+        <th><%= Spree.t("promotion_code_batches.base_code") %></th>
```
tvdeyen: Please don't invent your own translation keys, if Rails already has one for you.
Use `Spree::PromotionCodeBatch.human_attribute_name(:COLUMN_NAME)` instead.

Thanks

core/app/mailers/spree/promotion_code_batch_mailer.rb
```diff
@@ -0,0 +1,31 @@
...
+          promotion_code_batch_id: promotion_code_batch.id
+        ),
+      )
+    end
+
+    def promotion_code_batch_errored(promotion_code_batch)
+      mail(
+        to: promotion_code_batch.email,
+        body: Spree.t(
+        "promotion_code_batches.errored",
```
tvdeyen: wrong indention

backend/app/views/spree/admin/promotion_code_batches/download.csv.ruby
```diff
@@ -0,0 +1,6 @@
+CSV.generate do |csv|
+  csv << ['Code']
```
tvdeyen: Normally this header should be translatable, but this view could also be overwritten in the host app. So, I'm not super keen about it.

backend/app/views/spree/admin/promotions/_form.html.erb
```diff
@@ -8,17 +8,10 @@
...
-            <%= bf.label :base_code %>
-            <%= bf.text_field :base_code, :readonly => !@promotion.new_record?, :class => 'fullwidth' %>
-          <% end %>
-
-          <%= bf.field_container :number_of_codes do %>
-            <%= bf.label :number_of_codes %>
-            <%= bf.text_field :number_of_codes, :readonly => !@promotion.new_record?, class: 'fullwidth' %>
-          <% end %>
-        <% end %>
+        <div class="field" id="promotion_single_code_field">
```
tvdeyen: Since we don't use this ID anywhere we should rewrite this into a `data-hook` attribute, if I assume correctly that you added this for all the defacers out there.

core/app/mailers/spree/promotion_code_batch_mailer.rb
```diff
@@ -0,0 +1,31 @@
...
+  class PromotionCodeBatchMailer < ApplicationMailer
+    def promotion_code_batch_finished(promotion_code_batch)
+      mail(
+        to: promotion_code_batch.email,
+        body: Spree.t(
+          "promotion_code_batches.finished",
+          number_of_codes: promotion_code_batch.number_of_codes
+        ),
+        subject: Spree.t(
+          "promotion_code_batches.email_subject.batch_finished",
```
tvdeyen: Rails already has i18n scopes for mailer subjects. This mailer has the scope: 

```
en:
  spree:
    promotion_batch_mailer:
      promotion_code_batch_finished:
        subject: "Your batches are ready!"
```

core/app/mailers/spree/promotion_code_batch_mailer.rb
```diff
@@ -0,0 +1,31 @@
+module Spree
+  class PromotionCodeBatchMailer < ApplicationMailer
+    def promotion_code_batch_finished(promotion_code_batch)
+      mail(
+        to: promotion_code_batch.email,
+        body: Spree.t(
```
tvdeyen: I don't like to hard code the body in here. Please let's use mailer views as for all the other mailers.

core/app/mailers/spree/promotion_code_batch_mailer.rb
```diff
@@ -0,0 +1,31 @@
...
+          "promotion_code_batches.email_subject.batch_finished",
+          promotion_code_batch_id: promotion_code_batch.id
+        ),
+      )
+    end
+
+    def promotion_code_batch_errored(promotion_code_batch)
+      mail(
+        to: promotion_code_batch.email,
+        body: Spree.t(
```
tvdeyen: Same here: We shouldn't hard code the body in here.

core/app/mailers/spree/promotion_code_batch_mailer.rb
```diff
@@ -0,0 +1,31 @@
...
+        "promotion_code_batches.errored",
+          error: promotion_code_batch.error
+        ),
+        subject: Spree.t(
```
tvdeyen: Same as above: We don't need to introduce our own i18n scope.

tvdeyen: Ok, you are passing the id in here. That's why you need you own translation scope. 

Hmm, I doubt that any user has a real understanding where this id is coming from and what to do with it.

Maybe a link to the failed batch in the admin would be much more useful. And this correlates to my former comment: Let's use mailer views instead of a hard coded body.

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -0,0 +1,66 @@
...
+      promotion.reload
+    end
+  end
+
+  def generate_random_code
+    suffix = Array.new(self.class.random_code_length) do
+      sample_characters.sample
+    end.join
+
+    "#{base_code}_#{suffix}"
```
tvdeyen: This concatenation can not be configured anywhere, right? We should make this configurable, as I already tickets in my backlog ("remove the underscore form promotion codes") 😆

backend/app/views/spree/admin/promotion_code_batches/download.csv.ruby
cbrunsdon: I don't know if I agree with CSV headers being i18n'd, as they might need to be parsed by external systems... so I'm also a toss up. So I'm probably good with this as a default, and as Thomas says letting people override if they don't like.

core/app/models/spree/promotion_code/batch_builder.rb
cbrunsdon: I say ignore thomas and let him work on his own backlog, as this PR is separate from that feature.

core/app/models/spree/promotion_code_batch.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    def finished?
+      number_of_codes == promotion_codes.count
+    end
+
+    def process
+      if promotion_codes.count.zero?
+        PromotionCodeBatchJob.perform_later(self)
+      else
+        raise "Batch already started"
```
cbrunsdon: can you create a CantProcessStartedBatch exception and call that instead of the raise please? Makes finding different exceptions in external systems way easier.

tvdeyen: 👍 great catch @cbrunsdon

jhawthorn: This also won't prevent `process` from being called multiple times, since no codes will be created until the job starts running. Worth considering if we care about this race condition.

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def build_promotion_codes
+    generate_random_codes
+
+    Spree::PromotionCodeBatchMailer
+      .promotion_code_batch_finished(promotion_code_batch)
+      .deliver_now
+  rescue => e
+    promotion_code_batch.update!(error: e.inspect)
```
jordan-brough: @jhawthorn this will make the admin UI more helpful. If a batch fails you can go in and see the reason that it fails.  E.g. a customer service rep tries to create a promo, it fails for some reason, they tell a dev, and the dev can easily see what happened (and go look in their exception reporting tool or whatever for more details).

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -0,0 +1,66 @@
...
+    @promotion_code_batch = promotion_code_batch
+  end
+
+  def build_promotion_codes
+    generate_random_codes
+
+    Spree::PromotionCodeBatchMailer
+      .promotion_code_batch_finished(promotion_code_batch)
+      .deliver_now
+  rescue => e
```
jordan-brough: Moving the email to the Job makes sense to me but I think setting the error on the promo_code_batch should stay here.  We allow a row to exist in the `spree_promotion_code_batches` table when it may have failed part way through, so marking it with an error makes sense to me, and this seems like the place to do it.

core/app/models/spree/promotion_code_batch.rb
jordan-brough: @jhawthorn good point.  Since this model represents a "batch" which can't reasonably be completed within a transaction, perhaps we should have some sort of state column on it and not just infer the state by counting the codes on it.  Maybe `started_at`/`errored_at`/`finished_at` columns?  Or a `state` column?

core/db/migrate/20161017102621_create_spree_promotion_code_batch.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.timestamps
+    end
+
+    add_foreign_key(
+      :spree_promotion_code_batches,
+      :spree_promotions,
+      column: :promotion_id
+    )
+
+    add_column(
```
jordan-brough: We should add an index on this column

core/db/migrate/20161017102621_create_spree_promotion_code_batch.rb
```diff
@@ -0,0 +1,30 @@
+class CreateSpreePromotionCodeBatch < ActiveRecord::Migration[5.0]
+  def change
+    create_table :spree_promotion_code_batches do |t|
+      t.references :promotion, null: false
```
jordan-brough: We should add `index: true` on this

vladstoick:
I am adding a comment here as some changes are now outdated
@tvdeyen:
### Regarding: `send_file`:
I just used what was used in `promotion_codes_controller`. However, is it possible to use `send_file` for this? I was able to use `send_data` and `render_to_string`?
### Email body & URLs:
I have moved the body to individual erb files but I have noticed there's no default `action_mailer.default_url_options` so I can't generate URLs (I could add that in the spec but then by default the sandbox would have errors). I have added the promotion name in the meanwhile.

@cbrunsdon, @jordan-brough I think I covered all of your concerns. I have added a new column `state`

vladstoick:
@cbrunsdon, @tvdeyen, @jordan-brough: Is this still being consider or should I close this PR?

cbrunsdon:
Hey, sorry @vladstoick, I am super interested in getting this merged into core I just lost sight of the PR and you have my apologies, thanks for sticking through and poking us again.

PR #1528, opened 2016-10-17
Do not require 'wrapper_class' var in price checkbox partial

mamhoff:
In the bootstrap refactoring, the argument "wrapper_class" given to
rendering the `rebuild_vat_prices_checkbox` was lost, leading to
method not founds for stores with VAT. This commit changes the
reference from `wrapper_class` to `local_assigns[:wrapper_class]`
so that

  1. the partial does not error out when rendered without it and
  2. it is immediately apparent to the reader that `wrapper_class`
     is not a helper.

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -38,7 +38,7 @@
       </div>
 
       <% if show_rebuild_vat_checkbox? %>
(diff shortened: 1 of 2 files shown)


mamhoff:
I prefer the solution with `local_assigns[]` :)

flyfy1:
Great! I didn't know `local_assigns` before

PR #1529, opened 2016-10-18
Cancel authorized (pending) payments when cancelling an order

jordan-brough:
E.g., release funds on a credit card auth for cancelled orders whose
payments have been authorized but not yet captured.

This was already happening for store credit payments.

I tried to think of a reason not to void `pending` payments when cancelling an order and couldn't think of any reason.  Perhaps most people simply capture at order complete so this hasn't been an issue? And/or they don't worry about releasing authorizations on people's credit cards?

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -761,8 +761,7 @@ module Spree
 
     def after_cancel
       shipments.each(&:cancel!)
-      payments.completed.each(&:cancel!)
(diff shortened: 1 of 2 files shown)


gmacdougall:
I tried rebasing this to solve the conflicts but ran in to spec failures. We've changed the way that payment cancellation works recently, and I need to get a better understanding of this before I'm confident in what's going on.

gmacdougall:
Replaced by rebased version (#2265)

Issue #1475, opened 2016-09-27
Refunding Doesn't Update Credit Owed


gmacdougall:
When a refund is given, the "Credit Owed" in the admin and order payment state are not correctly updated.
# Steps to Reproduce

Start up a Solidus Sandbox
1. Purchase 3 items
2. Go to the admin page for that order
3. Capture the payment for the order
4. Cancel one of the three items ordered
5. Go to the payment page and see that credit is owed
6. Refund the amount owed to the customer
## Expected Results

The order should move into a paid state and balance due/credit owed should be $0
## Actual Results

The order's payment state is still credit owed, with -$20.99 showing as credit owed at the top of the page.
# Example

![image](https://cloud.githubusercontent.com/assets/764390/18876922/7c43f292-8499-11e6-8632-6b09dd53f7cc.png)

DanielePalombo:
My and @mtylty started to investigate this issue and we found these related issues: #1254, #1107, https://github.com/spree/spree/issues/6229

skukx:
The reason this is happening is because of this line:
https://github.com/solidusio/solidus/blob/v2.1/core/app/models/spree/order.rb#L355
```ruby
    def outstanding_balance
      # If reimbursement has happened add it back to total to prevent balance_due payment state
      # See: https://github.com/spree/spree/issues/6229
      adjusted_payment_total = payment_total + refund_total

      if state == 'canceled'
        -1 * adjusted_payment_total
      else
        total - adjusted_payment_total
      end
    end
```

After a `Spree::Refund` is created `perform!` is called
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/refund.rb#L16

In turn, this method calls `payment.order.updater.update` Which updates the `payment_total` column on the `Spree::Order`. This update already accounts for refunds when calculating `payment_total`.

See https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_updater.rb#L141

So TLDR; Get rid of `adjusted_payment_total` because `payment_total` is already adjusted.

jhawthorn:
I believe this is fixed by #2002

PR #1536, opened 2016-10-19
Fix how orders payment total is calculated

DanielePalombo:
In Spree::OrderUpdater the `update_payment_total` was incorrectly
removing  the refunds total from each order payment. This seems
incorrect because they are separate things. Refund total has to be
taken away into the outstanding_balance calculation instead.

rel #1475

### DIFF
--- a/core/app/models/spree/order_updater.rb
+++ b/core/app/models/spree/order_updater.rb
@@ -50,7 +50,7 @@ module Spree
 
     # Updates the following Order total values:
     #
-    # +payment_total+      The total value of all finalized Payments (NOTE: non-finalized Payments are excluded)
+    # +incoming_total+     The total value of all finalized Payments (NOTE: non-finalized Payments are excluded)
     # +item_total+         The total value of all LineItems
     # +adjustment_total+   The total value of all adjustments (promotions, credits, etc.)
     # +promo_total+        The total value of all promotion adjustments
@@ -73,7 +73,11 @@ module Spree
     end
 
     def update_payment_total
-      order.payment_total = payments.completed.includes(:refunds).map { |payment| payment.amount - payment.refunds.sum(:amount) }.sum
+      update_incoming_payment
+    end
+
+    def update_incoming_payment
+      order.incoming_payment = payments.completed.sum(&:amount)
     end
 
     def update_shipment_total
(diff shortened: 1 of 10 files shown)


core/app/models/spree/order_updater.rb
```diff
@@ -73,7 +73,7 @@ def update_shipments
     end
 
     def update_payment_total
-      order.payment_total = payments.completed.includes(:refunds).map { |payment| payment.amount - payment.refunds.sum(:amount) }.sum
+      order.payment_total = payments.completed.sum(&:amount)
```
jhawthorn: I'm concerned that we're changing the meaning of this column. Both because existing applications may expect the amount to have the refunds subtracted, and because existing records in the database will already exist with the existing values.

jhawthorn:
cc @jordan-brough who has written a good summary of the state of Refunds, payment_total and outstanding balanace in #1254

DanielePalombo:
After some insights i refactor the code to keep backward compatibility

bbuchalter:
@DanielePalombo this PR didn't work as expected for me. Steps I took:
1. Pulled your branch local
2. Wiped my sandbox
3. Created sandbox
4. Added an item to my cart
5. Completed my order
6. Went to admin and captured payment for order
7. Refunded order.

Expected to see balance due as $0.00 with a payment state of `paid`. Instead:
screen shot 2016-10-28 at 3 38 40 pm

Thank you very much for your effort on this. I'll be happy to continue to QA, but I believe we now need some feature tests to automate some of this QA if wish to avoid future regressions.

DanielePalombo:
@bbuchalter,
thanks for your attemps. 
 
> but I believe we now need some feature tests to automate some of this QA if wish to avoid future regressions.

I Agree with you!

I can confirm the result after the same steps, but maybe i found a workaround. After the refund i create an adjustment ,or delete the item from `cancel tab`, and the balance becomes 0 and the state `paid`.

screenshot from 2016-10-30 22-02-11
screenshot from 2016-10-30 22-02-04

I don't know if this can resolve your issues or if we can create an adjustment after the reimbursment... i would like to deepen the discussion.

bbuchalter:
I'm going to defer further conversation on this topic to the core team at the moment.

ericsaupe:
Any updates on this?

jhawthorn:
Thanks Daniele. I think this is fixed #2002. If not please file a new issue or PR

PR #1537, opened 2016-10-19
WIP remove Spree::Zone.default_tax and associated code paths

mamhoff:
Previously, Spree and Solidus had code paths for tax reimbursements:
If an orders tax address was outside the "default tax zone" and value added tax
applied from that default tax zone, Solidus would create negative tax
adjustments, i.e. Reimbursements. Since Solidus 1.3, the taxation system
has been overhauled and we do not need this hack anymore. Taking it out is
roughly this commit. It can (and should) be broken up into smaller commits,
and we need to come up with a good deprecation plan for many of these methods.


CHANGELOG.md
```diff
@@ -1,4 +1,12 @@
 ## Solidus 2.1.0 (master, unreleased)
+
+*   Remove the `default_tax` boolean from `Spree::Zone`, as well as
+    associated code paths.
+
+    In versions prior to Solidus 1.3, instead of changing prices for
+    different VAT territories, there was an assumption that backend
+    prices would include VAT, and the tax system was tasked with "refunding" that. This has since been changed so we could now delete a lot of code paths that performed these "refunds".
+
```
tvdeyen: I think it should be enough the link to the 1.3 blog post.

core/app/models/spree/order.rb
```diff
@@ -227,7 +227,7 @@ def backordered?
     # Returns the relevant zone (if any) to be used for taxation purposes.
     # Uses default tax zone unless there is a specific match
```
tvdeyen: Please remove this outdated comment

core/app/models/spree/zone.rb
```diff
@@ -42,13 +41,17 @@ class Zone < Spree::Base
     self.whitelisted_ransackable_attributes = ['description']
 
     def self.default_tax
-      where(default_tax: true).first
+      Spree::Deprecation.warn("There is no default tax zone anymore.", caller)
+      all.detect { |zone| zone.try(:default_tax) }
```
tvdeyen: If the column is gone, there is no need for this anymore, right? Maybe we should keep the column for at least one minor version instead?

core/db/migrate/20161019110308_remove_default_tax_from_zones.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveDefaultTaxFromZones < ActiveRecord::Migration[5.0]
+  class DefaultTaxZonePresent < StandardError; end
+  def change
+    if Spree::Zone.where(default_tax: true).any?
+      raise DefaultTaxZonePresent, "Please run the the solidus:migrations:create_vat_prices rake task."
+    end
+    remove_column :spree_zones, :default_tax, :boolean, default: false
```
tvdeyen: Maybe we keep the column for another round?

core/app/models/spree/zone.rb
```diff
@@ -42,13 +41,17 @@ class Zone < Spree::Base
...
     end
 
+
     # Returns the most specific matching zone for an address. Specific means:
     # A State zone wins over a country zone, and a zone with few members wins
     # over one with many members. If there is no match, returns nil.
     def self.match(address)
+      Spree::Deprecation.warn("Spree::Zone.match is deprecated. Please use Spree::Zone.for_adress instead.", caller)
```
peterberkenbosch: small typo. should be `Spree::Zone.for_address`

peterberkenbosch:
public API does not change right? This cleans up so nicely :) I really like it. That being said, I think these kind of changes should be introduced very defensively.

mamhoff:
Closing in favour of #1586 .

PR #1538, opened 2016-10-19
Use ActiveModel::Naming for default tax calculator

mamhoff:
The descriptions for calculators are not namespaced. This commit makes
`Spree::Calculator.description` are shortcut to `Spree::Calculator.model_name.human`.

This way, we can eliminate a few lines from the codebase, as well as
converging around Rails conventions for naming things.

This PR only touches the default tax calculator, and is supposed to be a blueprint / RFC type PR.

### DIFF
--- a/core/app/models/spree/calculator.rb
+++ b/core/app/models/spree/calculator.rb
@@ -17,9 +17,10 @@ module Spree
       end
     end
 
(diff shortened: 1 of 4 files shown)


jhawthorn:
:+1: I like it

PR #1540, opened 2016-10-19
Payment methods display on boolean

### LINKED ISSUE #1532: RFC: Split Spree::PaymentMethod.display_on to two boolean columns
RFC on changing the `display_on` column of Spree::PaymentMethod to two boolean columns `available_to_users` and `available_to_admin`.

gevann:
This PR changes the `display_on` column of Spree::PaymentMethods to two boolean columns, `available_to_users` and `available_to_admin`. This allows clearer selection than the string-based `display_on` column, enabling usage of scopes on the model, and improving readability and intention of the affected code. 
Closes #1532 

This PR will be reworked into two-value logic after the merge of PR #1541, which fixes invalid specs for payment methods.

### DIFF
--- a/backend/app/views/spree/admin/payment_methods/_form.html.erb
+++ b/backend/app/views/spree/admin/payment_methods/_form.html.erb
@@ -25,9 +25,13 @@
 
           <div class="info warning js-gateway-settings-warning"><%= Spree.t(:provider_settings_warning) %></div>
         </div>
-        <div data-hook="display" class="field">
-          <%= f.label :display_on %>
-          <%= select(:payment_method, :display_on, Spree::PaymentMethod::DISPLAY.collect { |display| [Spree.t(display), display == :both ? nil : display.to_s] }, {}, {:class => 'select2 fullwidth'}) %>
+        <div data-hook="available_to_user" class="field">
+          <%= label_tag nil, Spree::PaymentMethod.human_attribute_name(:available_to_users) %>
+          <%= f.check_box :available_to_users %>
+        </div>
+        <div data-hook="available_to_user" class="field">
+          <%= label_tag nil, Spree::PaymentMethod.human_attribute_name(:available_to_admin) %>
+          <%= f.check_box :available_to_admin %>
         </div>
         <div data-hook="auto_capture" class="field">
           <%= f.label :auto_capture %>
--- a/core/app/models/spree/payment_method.rb
+++ b/core/app/models/spree/payment_method.rb
@@ -14,6 +14,10 @@ module Spree
     has_many :stores, through: :store_payment_methods
 
     scope :ordered_by_position, -> { order(:position) }
+    scope :active, -> { where(active: true) }
+    scope :available_to_users, -> { where(available_to_users: true) }
+    scope :available_to_admin, -> { where(available_to_admin: true) }
+    scope :available_to_store, -> (store) { (store.present? && store.payment_methods.empty?) ? self : store.payment_methods }
 
     include Spree::Preferences::StaticallyConfigurable
 
@@ -32,11 +36,25 @@ module Spree
       raise ::NotImplementedError, "You must implement payment_source_class method for #{self.class}."
     end
 
-    def self.available(display_on = 'both', store: nil)
-      all.select do |p|
-        p.active &&
-          (p.display_on == display_on.to_s || p.display_on.blank?) &&
-          (store.nil? || store.payment_methods.empty? || store.payment_methods.include?(p))
+    def self.available(display_on=nil, store: nil)
+      Spree::Deprecation.warn "Spree::PaymentMethod.available is deprecated."\
+        "Please use .active, .available_to_users, and .available_to_admin scopes instead."\
+        "For payment methods associated with a specific store, use Spree::PaymentMethod.available_to_store(your_store)"\
+        " as the base applying any further filtering"
+
+      display_on = display_on.to_s
+
+      available_payment_methods =
+        case display_on
+        when 'front_end'
+          active.available_to_users
+        when 'back_end'
+          active.available_to_admin
+        else
+          active.available_to_users.available_to_admin
+        end
+      available_payment_methods.select do |p|
+        store.nil? || store.payment_methods.empty? || store.payment_methods.include?(p)
       end
     end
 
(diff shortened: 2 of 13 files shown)


backend/app/views/spree/admin/payment_methods/_form.html.erb
```diff
@@ -25,9 +25,31 @@
...
           <div class="info warning js-gateway-settings-warning"><%= Spree.t(:provider_settings_warning) %></div>
         </div>
-        <div data-hook="display" class="field">
-          <%= f.label :display_on %>
-          <%= select(:payment_method, :display_on, Spree::PaymentMethod::DISPLAY.collect { |display| [Spree.t(display), display == :both ? nil : display.to_s] }, {}, {:class => 'select2 fullwidth'}) %>
+        <div data-hook="available_to_user" class="field">
+          <%= label_tag nil, Spree::PaymentMethod.human_attribute_name(:available_to_users) %>
+          <ul>
+            <li>
+              <%= radio_button :payment_method, :available_to_users, true %>
```
jhawthorn: A checkbox might be simpler than a radio button

core/app/models/spree/payment_method.rb
```diff
@@ -33,10 +39,23 @@ def payment_source_class
     end
 
     def self.available(display_on = 'both', store: nil)
-      all.select do |p|
-        p.active &&
-          (p.display_on == display_on.to_s || p.display_on.blank?) &&
-          (store.nil? || store.payment_methods.empty? || store.payment_methods.include?(p))
+      Spree::Deprecation.warn "Spree::PaymentMethod.available is deprecated."\
+        "Please use .active, .available_to_users, and .available_to_admin scopes instead."\
+        "For payment methods associated with a specific store, use your_store.payment_methods."
```
jhawthorn: This isn't a precise replacement ,`your_store.payment_methods` won't return payments if none are associated (it should return all payments in this case). `Payment.available_to_store(store)` which I suggest above would solve this.

core/db/migrate/20161014221052_add_available_to_columns_and_remove_display_on_from_payment_methods.rb
```diff
@@ -0,0 +1,28 @@
+class AddAvailableToColumnsAndRemoveDisplayOnFromPaymentMethods < ActiveRecord::Migration[5.0]
+  def up
+    add_column(:spree_payment_methods, :available_to_users, :boolean, default: nil)
```
jhawthorn: What meaning does the default, `nil`, have?

I would prefer that the default to both be `true` (matching the old default behaviour) and that the column was `null: false` to avoid any tri-value logic.

core/db/migrate/20161014221052_add_available_to_columns_and_remove_display_on_from_payment_methods.rb
```diff
@@ -0,0 +1,28 @@
...
+    add_column(:spree_payment_methods, :available_to_admin, :boolean, default: nil)
+    execute("UPDATE spree_payment_methods "\
+             "SET available_to_users='true' "\
+             "WHERE (display_on='front_end' OR display_on='' OR display_on='both')")
+    execute("UPDATE spree_payment_methods "\
+             "SET available_to_admin='true' "\
+             "WHERE (display_on='back_end' OR display_on='' OR display_on='both')")
```
jhawthorn: :+1: Thank you for using plain SQL in the migration

core/app/models/spree/payment_method.rb
```diff
@@ -33,10 +39,23 @@ def payment_source_class
...
+      display_on = display_on.to_s
+
+      available_payment_methods =
+        case display_on
+        when 'front_end'
+          active.available_to_users
+        when 'back_end'
+          active.available_to_admin
+        else
+          active.available_to_users.available_to_admin + active.nil_available_to_users.nil_available_to_admin
```
jhawthorn: I don't follow this line. I don't think it replicates the existing functionality.

gevann: In the existing `core/spec/models/spree/payment_method_spec.rb` file's `describe .available` specs, there are four payment methods created, with `display_on` values of `'front_end'`, `'back_end'`, `'both'`, and `nil`.
The existing functionality filters with `(p.display_on == display_on.to_s || p.display_on.blank?)`. This behaviour of `.blank?` makes the live above (and the three value logic) necessary; the existing specs assert that there are two methods available for both. This should mean that there would be three each available for frontend and backend, as there are two for both and one each for those particular locations, however the specs show that this is not the case.

core/db/migrate/20161014221052_add_available_to_columns_and_remove_display_on_from_payment_methods.rb
gevann: I agree. Originally it was meant to be `default: true` and avoid tri-value logic, however the edge case of `nil` valued `display_on`s make tri-value logic necessary.

core/db/migrate/20161014221052_add_available_to_columns_and_remove_display_on_from_payment_methods.rb
```diff
@@ -0,0 +1,28 @@
+class AddAvailableToColumnsAndRemoveDisplayOnFromPaymentMethods < ActiveRecord::Migration[5.0]
+  def up
+    add_column(:spree_payment_methods, :available_to_users, :boolean, default: true)
+    add_column(:spree_payment_methods, :available_to_admin, :boolean, default: true)
+    execute("UPDATE spree_payment_methods "\
+             "SET available_to_users='false' "\
+             "WHERE NOT (display_on='front_end' OR display_on='' OR display_on IS NULL)")
+    execute("UPDATE spree_payment_methods "\
+             "SET available_to_admin='false' "\
+             "WHERE NOT (display_on='back_end' OR display_on='' OR display_on IS NULL)")
```
cbrunsdon: I'm not 100% sure thats safe, I'd prefer something like:
```ruby
"SET available_to_admin=#{ActiveRecord::Base.connection.quote(true)}"
```

peterberkenbosch:
@gevann this needs a rebase with master.

peterberkenbosch:
LGTM! 👍

PR #1546, opened 2016-10-20
Use actionmodel naming for calculators

### LINKED ISSUE #1544: Use ActiveModel::Naming for calculator descriptions
All of Spree's calculators and our internationalization would benefit from better use of `ActiveModel::Naming`, a module included in all Rails models. This is up for grabs, if you want a commit in Solidus, do this for other calculators in core! 

The base work for this is done in #1538 - all you'd have to do for each calculator is remove the `description` class method and add the correct translation key to the `spree.models` I18n namespace. 

Find the docs for `ActiveModel::Naming` [here](http://api.rubyonrails.org/classes/ActiveModel/Naming.html). 

These are the calculators still to be translated: 

- [ ] app/models/spree/calculator/flat_percent_item_total.rb
- [ ] app/models/spree/calculator/flat_rate.rb
- [ ] app/models/spree/calculator/flexi_rate.rb
- [ ] app/models/spree/calculator/free_shipping.rb
- [ ] app/models/spree/calculator/percent_on_line_item.rb
- [ ] app/models/spree/calculator/percent_per_item.rb
- [ ] app/models/spree/calculator/price_sack.rb
- [ ] app/models/spree/calculator/returns
- [ ] app/models/spree/calculator/returns/default_refund_amount.rb
- [ ] app/models/spree/calculator/shipping
- [ ] app/models/spree/calculator/shipping/flat_percent_item_total.rb
- [ ] app/models/spree/calculator/shipping/flat_rate.rb
- [ ] app/models/spree/calculator/shipping/flexi_rate.rb
- [ ] app/models/spree/calculator/shipping/per_item.rb
- [ ] app/models/spree/calculator/shipping/price_sack.rb
- [ ] app/models/spree/calculator/tiered_flat_rate.rb
- [ ] app/models/spree/calculator/tiered_percent.rb

gevann:
This PR is based on #1538 and closes #1544

### DIFF
--- a/core/spec/models/spree/calculator/default_tax_spec.rb
+++ b/core/spec/models/spree/calculator/default_tax_spec.rb
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'shared_examples/calculator_shared_examples'
 
 describe Spree::Calculator::DefaultTax, type: :model do
   let(:address) { create(:address) }
@@ -9,10 +10,7 @@ describe Spree::Calculator::DefaultTax, type: :model do
   let(:included_in_price) { false }
   subject(:calculator) { Spree::Calculator::DefaultTax.new(calculable: rate ) }
 
-  describe ".description" do
-    subject { described_class.description }
-    it { is_expected.to eq("Default Tax") }
-  end
+  it_behaves_like 'a calculator with a description'
 
   context "#compute" do
     context "when given an order" do
--- a/core/spec/models/spree/calculator/flat_percent_item_total_spec.rb
+++ b/core/spec/models/spree/calculator/flat_percent_item_total_spec.rb
@@ -4,6 +4,11 @@ describe Spree::Calculator::FlatPercentItemTotal, type: :model do
   let(:calculator) { Spree::Calculator::FlatPercentItemTotal.new }
   let(:line_item) { mock_model Spree::LineItem }
 
+  describe ".description" do
+    subject { described_class.description }
+    it { is_expected.to eq("Flat Percent") }
+  end
+
   before { allow(calculator).to receive_messages preferred_flat_percent: 10 }
 
   context "compute" do
--- a/core/spec/models/spree/calculator/flat_percent_item_total_spec.rb
+++ b/core/spec/models/spree/calculator/flat_percent_item_total_spec.rb
@@ -1,13 +1,11 @@
 require 'spec_helper'
+require 'shared_examples/calculator_shared_examples'
 
 describe Spree::Calculator::FlatPercentItemTotal, type: :model do
   let(:calculator) { Spree::Calculator::FlatPercentItemTotal.new }
   let(:line_item) { mock_model Spree::LineItem }
 
-  describe ".description" do
-    subject { described_class.description }
-    it { is_expected.to eq("Flat Percent") }
-  end
+  it_behaves_like 'a calculator with a description'
 
   before { allow(calculator).to receive_messages preferred_flat_percent: 10 }
 
--- a/core/app/models/spree/calculator/flat_percent_item_total.rb
+++ b/core/app/models/spree/calculator/flat_percent_item_total.rb
@@ -4,10 +4,6 @@ module Spree
   class Calculator::FlatPercentItemTotal < Calculator
     preference :flat_percent, :decimal, default: 0
 
-    def self.description
-      Spree.t(:flat_percent)
-    end
-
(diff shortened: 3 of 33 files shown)


core/spec/models/spree/calculator/flat_percent_item_total_spec.rb
```diff
@@ -4,6 +4,11 @@
   let(:calculator) { Spree::Calculator::FlatPercentItemTotal.new }
   let(:line_item) { mock_model Spree::LineItem }
 
+  describe ".description" do
+    subject{ described_class.description }
```
mamhoff: There's a space missing here before the opening `{`. Can you add that one space? I know it's lots of commits to go and edit, but interactive rebase makes that pretty straightforward. Read up [here](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase-i) about it, it's great. And yes, please force push afterwards, we don't mind for open PRs before they're merged.

tvdeyen:
Thanks for your contribution.

I don't see any value in testing these descriptions. Especially if we hard core translations inside tests. That only means broken builds if we update a translation. 

There is no business logic included at all that needed testing. 

So, from my point of view we should just remove them.

mamhoff:
OK, @tvdeyen this makes sense. My fault here for providing an example test... Maybe add a spec for the base calculator to call `model_name.human` instead? @gevann you can do that in an additional commit.

gevann:
@mamhoff no problem, I'll take care of that as you and @tvdeyen have suggested. Thanks for the feedback.

gevann:
@tvdeyen @mamhoff, spoke with @cbrunsdon, he says to keep the tests despite the hard-coding.

cbrunsdon:
Yea, I'm pro leaving the tests in. Since we're testing the output, which is the text string, not the implementation, which we know goes through standard rails i18n.

tvdeyen:
Testing a random string has no value for me. Why don't we test all other strings then, why exactly these strings? Are they special, and when yes, what makes them special?

The only value I see is that we ensure all calculators inherit from `Spree::Calculator`. Why not test this then? 

Sorry, still not convinced, but I don't resist on removing them. If anybody these value in them, please keep em. 

I only care about less burdens.

cbrunsdon:
I'd be okay with something like a shared example for calculators that just tested that there was a non-zero length string there

tvdeyen:
`it_behaves_like "having a description"` 👍

Like that

core/spec/models/spree/calculator/default_tax_spec.rb
```diff
@@ -9,6 +10,11 @@
   let(:included_in_price) { false }
   subject(:calculator) { Spree::Calculator::DefaultTax.new(calculable: rate ) }
 
+  describe ".description" do
```
tvdeyen: You can even put the whole `describe` block including the `subject` in the shared example so that you only have to leave `it_behaves_like` in the calculator specs.

core/lib/spree/testing_support/calculator_helpers.rb
```diff
@@ -0,0 +1,5 @@
+shared_examples_for 'having a description' do
+  it "has a description" do
+    expect(subject.size).to be > 0
```
tvdeyen: I think expecting a `String` would be better. Every enumerable has a `size` method.

tvdeyen:
Sorry for being so picky!

core/spec/models/spree/calculator/default_tax_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'spree/testing_support/calculator_helpers'
```
gevann: I think a new folder would be clearest -- I didn't want to add one myself as I'm new to Solidus, so I just chose a place where I found other shared examples being defined currently.

core/spec/models/spree/calculator/default_tax_spec.rb
mamhoff: @gevann I second this thought. In the individual calculator specs, we only want to see something like `it_behaves_like "a calculator with a description"`. The `describe` block can go into the shared example.

mtomov:
@tvdeyen I would've been on the same board as you, so not to worry : ), but the way the shared example is implemented is just great! Thanks!

tvdeyen:
@mamhoff you requested changes that where now made. 
Would you mind to review again? Thanks

PR #1547, opened 2016-10-20
validate empty option values to display new variant form

chukitow:
Admins users are able to add a new product variant only if the product has option types with option values but they are able to see the new product variant form if the access from the URL.


jhawthorn:
Other than the small nitpick this seems okay to me. We hide the link to "New Variant" when there are no option types, and the variants page provides a pretty good explanation.

chukitow:
I updated the method using the name that you suggested @jhawthorn :)

chukitow:
Hey @jhawthorn is there anything else that you want me to do for this PR? :)

peterberkenbosch:
👍

PR #1550, opened 2016-10-21
Fix css stock item informations inside td.

DanielePalombo:
The stock items informations are not well formatted. 

schermata 2016-10-20 alle 23 06 22

Fixed in scss.


peterberkenbosch:
👍

PR #1552, opened 2016-10-21
check empty option values just in one query

chukitow:
If a product has for example 10 option types assigned, the `empty_option_values?` method will execute the same amount of queries in order to know if there are option types without option values, this could be performed in a single query.

### DIFF
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -215,9 +215,7 @@ module Spree
 
     # @return [Boolean] true if there are no option values
     def empty_option_values?
-      options.empty? || options.any? do |opt|
-        opt.option_type.option_values.empty?
-      end
+      options.empty? || !option_types.left_joins(:option_values).where('spree_option_values.id IS NULL').empty?
     end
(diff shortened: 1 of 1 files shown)


core/app/models/spree/product.rb
```diff
@@ -215,9 +215,9 @@ def variant_option_values_by_option_type(variant_scope = nil)
 
     # @return [Boolean] true if there are no option values
     def empty_option_values?
-      options.empty? || options.any? do |opt|
-        opt.option_type.option_values.empty?
-      end
+      options.empty? || !option_types
+                        .joins("LEFT OUTER JOIN #{Spree::OptionValue.table_name} ON #{Spree::OptionValue.table_name}.option_type_id = #{Spree::OptionType.table_name}.id")
```
jhawthorn: This can probably be `.left_joins(:option_values)`

Rails 5! :tada:

chukitow:
I updated the PR using the new rails 5 methods instead of the raw query @jhawthorn

Issue #1518, opened 2016-10-13
Loading wrapper upon XHR request


flyfy1:
Currently, in `solidus_backend/app/assets/javascripts/spree/backend/progress.coffee`, it listens to ajax request globally. Whenever there's a AJAX request, the progress bar would show, something like this:

![image](https://cloud.githubusercontent.com/assets/1553760/19334954/3f1fe952-9132-11e6-9fce-8f8513d1e02e.png)

and this progress bar as a overlay would stop any activities on the backend interface.

When I was trying to provide a Direct File Upload from the Admin Backend, I build something like this:

![image](https://cloud.githubusercontent.com/assets/1553760/19335003/925231ca-9132-11e6-8e81-205abd89a590.png)

The idea as to let user upload images async.

However, the issue is that.. when a image is uploaded, the `Loading` would popup, and disable user from doing anything, until the upload is done (i.e., the AJAX request completes).

Any idea on how to solve this issue?

mtomov:
I would love to see the direct image upload land in core of solidus! It's such a pain to upload images!

@tvdeyen  I know you'd share my suffering, so just pinging you for solidarity : )

tvdeyen:
@mtomov you have my full solidarity. I would like to port Alchemys image upload over to Solidus. It's heavily based on https://github.com/blueimp/jQuery-File-Upload. Will see if so can free some time these days. The current image upload is just ridiculous, but as many stores don't even use the build in image solution and use services like Cloudinary, there was not much demand for a better solution yet. But anyway, this needs to be fixed. 

@flyfy1 I don't see an easy solution. As others will also be bothered by this we should think about replacing this with a custom trigger or event that all our own XHR requests call when we do expensive stuff. Any thoughts @jhawthorn?

aitbw:
I think this can be closed as per #1553, can you confirm this @kennyadsl ?

tvdeyen:
Yes. Thanks for that find.

PR #1553, opened 2016-10-22
Upload images in admin with drag and drop

mtomov:
I wanted to see how much effort it would be to do a modern drag and drop image upload (#1518), and that pretty much does it. Code is ~~not~~ ideal, but wanted to post it here if somebody wants to try it out, have any comments, etc ..

In summary:
- The JavaScript is just about ~~100 lines~~ (grew to ~ 200 with Backbone), even non-coffee-script (as I started with a non coffee example, but can convert it around). 
- Code I started from includes tests for feature detection (Drag and Drop, File Upload, Progress Bar). I decided to leave them for reference, but they can be removed, as I think that admin users should run the very latest of browsers, and if they don't - they can use the old functionality.
- All of the old way of uploading images is in place, so if someone is stuck with some unsupported functions, they can always fallback to that.
- There's a `javascript` response from the server, which appends a row in the images table .. not sure if it's an accepted practice
- I need to add a capybara test, but shouldn't be difficult, as I have also left in place a Browse button if somebody doesn't want to drag and drop
- Image is automatically assigned the master variant
  - so to make it a truly nice experience, the rows in the images table will have an select dropdown field with ability to change the variant, and click a Tick button on the right, similarly to the Stock Items page, and other in-place row edits.

And some pictures:

screenshot from 2016-10-27 23 30 25

screenshot from 2016-10-27 23 40 48

or in boxes:

screenshot from 2016-11-03 21 40 05

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/images/upload.js
+++ b/backend/app/assets/javascripts/spree/backend/images/upload.js
@@ -1,30 +1,76 @@
-// https://simplyian.com/2016/02/09/Using-Underscore-js-templates-within-ERB/
-_.templateSettings = {
-  interpolate: /\{\{\=(.+?)\}\}/g,
-  evaluate: /\{\{(.+?)\}\}/g
-};
-
 // Inspired by: http://html5demos.com/dnd-upload
 Spree.prepareImageUploader = function () {
   var uploadZone = document.getElementById('upload-zone');
   if(!uploadZone) return;
 
-  var tests = {
+  var UploadZone = Backbone.View.extend({
+    el: uploadZone,
+
+    events: {
+      "dragover" : "onDragOver",
+      "dragleave" : "onDragLeave",
+      "drop" : "onDrop",
+      'change input[type="file"]' : "onFileBrowserSelect"
+    },
+
+    progressZone: document.getElementById('progress-zone'),
+
+    // Hide or highlight supported browser features
+    initialize: function() {
+      "filereader formdata progress".split(' ').forEach(function (api) {
+        this.support[api].className = (this.tests[api] === false) ? 'red' : 'hidden'
+      }, this);
+    },
+
+    upload: function(file) {
+      if (!this.tests.formdata) return;
+
+      var progressModel = new ProgressModel({file: file});
+      progressModel.previewFile();
+      progressModel.uploadFile();
+
+      var progressView = new ProgressView({model: progressModel});
+      this.progressZone.appendChild(progressView.render().el);
+    },
+
+    onDragOver: function() {
+      this.el.className = 'hover';
+      return false;
+    },
+
+    onDragLeave: function() {
+      this.el.className = '';
+      return false;
+    },
+
+    onDrop: function(e) {
+      this.el.className = '';
+      e.preventDefault();
+
+      for (var i = 0; i < e.originalEvent.dataTransfer.files.length; i++) {
+        this.upload(e.originalEvent.dataTransfer.files[i]);
+      }
+    },
+
+    onFileBrowserSelect: function(e) {
+      for (var i = 0; i < e.target.files.length; i++) {
+        this.upload(e.target.files[i]);
+      }
+    },
+
+    tests: {
       filereader: typeof FileReader != 'undefined',
       dnd: 'draggable' in document.createElement('span'),
       formdata: !!window.FormData,
       progress: "upload" in new XMLHttpRequest
     },
-    support = {
+
+    support: {
       filereader: document.getElementById('filereader'),
       formdata:   document.getElementById('formdata'),
       progress:   document.getElementById('progress')
-    },
-    progressZone = document.getElementById('progress-zone'),
-    progressTmpl = document.getElementById('upload-progress-tmpl').innerHTML,
-    uploadForm   = document.getElementById('upload-form'),
-    variantId    = uploadForm.querySelector('input[name="image[viewable_id]"]').value;
-
+    }
+  });
 
   var ProgressModel = Backbone.Model.extend({
     initialize: function() {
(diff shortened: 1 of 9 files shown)


mtomov:
Hello,

This one is ready for official review. I've cleaned up the javascript using Backbone, and have placed the supporting bits (css, js, translations) to their relevant places.

This is first time I'm using Backbone and I'm quite pleasantly surprised how it put a nice structure overall. The events mapping is very clean, and the view and its model work very well together to provide for _live_ data binding.

You might've noticed that I'm using the _underscore's_ templating function `_template()` instead of the customary Handlebars, and the reason is that I don't like how all handlebar templates are placed in a god forsaken directory separated from their actual location of usage. In terms of performance - since the template is only parsed and evaluated once on the initial page load, and I'm using the simpler template syntax with Underscore, I don't think there is a negative burden on the page.

In terms of design - current one is a bit plain. An alternative is to make use of [Bootstrap's Cards](https://v4-alpha.getbootstrap.com/components/card/#groups). So, either multiple images per row, or again a single image per row, but [wrapped in a round border](https://v4-alpha.getbootstrap.com/components/card/#outline-variants).

Thank you!

tvdeyen:
@mtomov thanks! I will have a look later.

backend/app/assets/javascripts/spree/backend/images/upload.js
```diff
@@ -0,0 +1,204 @@
...
+
+      // send the image to the server
+      Spree.ajax({
+        url: window.location.pathname,
+        type: "POST",
+        dataType: 'script',
+        data: formData,
+        processData: false,  // tell jQuery not to process the data
+        contentType: false,  // tell jQuery not to set contentType
+        xhr: function () {
```
Sinetheta: Adding progress to ajax is cool, we may want to make this generic to all `Spree.ajax` in the future. Make it a callback and trigger an event.

backend/app/views/spree/admin/images/_image_row.html.erb
```diff
@@ -0,0 +1,22 @@
+<tr id="<%= spree_dom_id image %>" data-hook="images_row" class="<%= cycle('odd', 'even')%>">
```
Sinetheta: As we add more long lived pages `cycle('odd', 'even')` is really gonna show it's limitations, as new uploads don't get striped.

backend/app/assets/javascripts/spree/backend/images/upload.js
```diff
@@ -0,0 +1,204 @@
...
+      var progressModel = new ProgressModel({file: file});
+      progressModel.previewFile();
+      progressModel.uploadFile();
+
+      var progressView = new ProgressView({model: progressModel});
+      this.progressZone.appendChild(progressView.render().el);
+    },
+
+    onDragOver: function() {
+      this.el.className = 'hover';
```
mtomov: Right so. I've gone for that `dragClass: 'with-images'`

backend/app/assets/javascripts/spree/backend/images/upload.js
```diff
@@ -0,0 +1,204 @@
...
+      return false;
```
mtomov: Good observation. Corrected to that

backend/app/assets/javascripts/spree/backend/images/upload.js
```diff
@@ -0,0 +1,204 @@
...
+    },
+
+    onDragLeave: function() {
+      this.el.className = '';
```
mtomov: Indeed so. Used the native's `el.classList.remove/add`

backend/app/assets/javascripts/spree/backend/images/upload.js
```diff
@@ -0,0 +1,204 @@
...
+      return false;
```
mtomov: Thanks for looking in those details. Corrected now.

backend/app/assets/javascripts/spree/backend/images/upload.js
```diff
@@ -0,0 +1,204 @@
...
+    },
+
+    onDrop: function(e) {
+      this.el.className = '';
+      e.preventDefault();
+
+      for (var i = 0; i < e.originalEvent.dataTransfer.files.length; i++) {
```
mtomov: did it as per your example - all excellent working. Thanks!

backend/app/assets/javascripts/spree/backend/images/upload.js
```diff
@@ -0,0 +1,204 @@
...
+      }
+    },
+
+    onFileBrowserSelect: function(e) {
+      for (var i = 0; i < e.target.files.length; i++) {
+        this.upload(e.target.files[i]);
+      }
+    },
+
+    tests: {
```
mtomov: I'm not in favour of adding more libs. The code for those random pieces of text is ~20 links. As you see fit..

backend/app/assets/javascripts/spree/backend/images/upload.js
mtomov: I think it's too early at this point for anything such. As far as I understand, one only gets progress for large user submitted requests, i.e. files. For regular post requests when the server is just slow .. that's a different story that would like need server support.

mtomov:
Thanks @Sinetheta and John for the elaborate look. I've indeed corrected all points that you mentioned - switched to Handlebars, etc ..

Furthermore:

1. In 8887673 - I added a small piece out of stack overflow, which allows to resolve keys with dots inside. Ex `{{t 'admin.images.index.image_process_failed' }}` - that wasn't possible before

2. In 61a62ec - I've given it a try to use Bootstrap's cards with image inside. I think it looks better. From code perspective only thing I don't like is that I had to place `className: 'col-sm-6 col-md-4 margin-bottom-one',` inside the javascript due to Backbone wrapping the element in a div. 

  I've uploaded image at the top of how the page looks with 3 boxes per row. Naturally, I can get rid of the commit if it's too much. Thanks again!

Sanjana6221:
As I can see when I try to upload large image around 20 MB I see the progress bar show 100% but the loader still running, and when I check it with the server logs it looks like paperclip is still trying to create that file.

Could you please let me know is this working as expected as it implemented?

PR #1554, opened 2016-10-23
Feature/address geo points

skukx:
Adds latitude and longitude columns to Spree::Address table.
These columns are auto-populated by the geocoder gem.

See https://github.com/alexreisner/geocoder


tvdeyen:
Hi. Thanks for your interest in Solidus and your contribution. It's very much appreciated. 

This feature is better as an extension and is not something we want to have in core. It brings privacy issues, that not every store wants to deal with. 

So, as core feature I :-1: this. But as extension this feature would be great, as this is good work.

tvdeyen:
If you have interest in maintaining the extension I can create a repo under https://github.com/solidusio-contrib for you.

jhawthorn:
This is very cool, but I also don't think this belongs in Solidus core.

Adding extra API calls to the checkout process is not something that can be taken lightly.

This would make an excellent extension

skukx:
@tvdeyen That would be great to get a repo under solidusio-contrib.  My employer would find it very valuable to perform analytics using these geo-points.

tvdeyen:
@skukx done

https://github.com/solidusio-contrib/solidus_geocoding

PR #1556, opened 2016-10-24
Add tooltip to store cart tax iso country field

mamhoff:
This adds a tooltip using the new `field_hint` method.

However, `field_hint` does not currently work with `fields_for`, which is
why this PR removes the `fields_for` call and instead uses a full form object
for just the store.

Advantages: We now limit the `GeneralSettingsController` to only change things for the store.
Disadvantages: See advantages - if anyone patched their `GeneralSettingsController`,
this would be an issue.


jhawthorn:
I make further changes in #1282, which I will rebase to include this

Issue #1107, opened 2016-04-29
Incorrect payment_state when using refunds


dangerdogz:
### Issue ###

**outstanding_balance** from [order.rb](https://github.com/dangerdogz/solidus/blob/master/core/app/models/spree/order.rb)

```ruby
def outstanding_balance
   # If reimbursement has happened add it back to total to prevent balance_due payment state
   # See: https://github.com/spree/spree/issues/6229
   adjusted_payment_total = payment_total + refund_total

   if state == 'canceled'
     -1 * adjusted_payment_total
   else
     total - adjusted_payment_total
   end
end
```

It's used by the order_updater to determine the payment state of an order :

| outstanding_balance | state | 
| ------------- |:-------------:|
| outstanding_balance > 0 | balance_due |
| outstanding_balance < 0 | credit_owed |
| outstanding_balance = 0 | paid |

### History ###

Current behavior comes from [this spree issue](https://github.com/spree/spree/issues/6229)

Basically, if you refund something on a completed order, instead of being _paid_, it was balance_due. So this fix this, outstanding_balance takes into account the refunds

### Scenario ###

1) User place an order for a total of 25$

2) Admin captures the payment

3) For any reason (error in the order, need to use another payment method or credit card), you have the refund the payment

You know have an order with a "paid" payment_state, but without having any valid, completed payment at all

Outstanding balance will do the following calculation : 25 [total] - ( 0 [payment_total] + 25 [refund_total])

So at this point, the order will forever be in a "wrong state" because you can't add new payments (since outstanding balance is 0). You will be able to process and ship it, even though it's not paid for.
If you add an item or edit it in any way, you will be only be able to add and capture payments for the newly added items.

If we take the above order and add a 10$ mug, we will be able to add a 10$ payment, and if we capture it, we'll get:

35 [total] - ( 10 [payment_total] + 25 [refund_total]), so we're still short of 25$

The issue is compounded with certain gateways like Moneris where you can't void or cancel a payment, and thus you need to refund+add a new one if you need to change it.

### Problem ###

The behavior was added to Solidus with the following commit with the following commit [84eddb501822f2602ed7542def20278db0714223](https://github.com/solidusio/solidus/commit/84eddb501822f2602ed7542def20278db0714223) and is visible from 1.1 onwards

[...]

PR #1557, opened 2016-10-24
Correct behavior of Order#outstanding_balance

stewart:
This commit addresses an issue in `Spree::Order#outstanding_balance`, in
circumstances where an Order has had refunds applied.

The existing behaviour cancels out the amount of refunded payments,
resulting in a negative outstanding balance and an order state of
'credit_due', when in fact the order balance is zero. This is caused by
the calculation of `payment_total` already factoring in refunds.

Some more discussion about this behaviour can be found in the following
issues:

https://github.com/solidusio/solidus/issues/1107
https://github.com/solidusio/solidus/pull/1536

With these changes, `outstanding_balance` is now defined as
`total - payment_total`. This corrects the behaviour around refunds.

Tests have been updated accordingly, with some help from
DanielePalombo's commit:

https://github.com/solidusio/solidus/pull/1536/commits/c851404860a43f07a77d2c9a80dd6afcf2a54984

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -353,14 +353,10 @@ module Spree
     end
 
     def outstanding_balance
-      # If reimbursement has happened add it back to total to prevent balance_due payment state
-      # See: https://github.com/spree/spree/issues/6229
-      adjusted_payment_total = payment_total + refund_total
-
       if state == 'canceled'
-        -1 * adjusted_payment_total
+        -1 * payment_total
       else
-        total - adjusted_payment_total
+        total - payment_total
       end
     end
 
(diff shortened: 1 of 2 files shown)


jhawthorn:
This seems like a good change, but this reopens the issue which caused this in the first place, spree/spree#6229.

To summarize:
1. From a completed, paid, shipped order
2. Create an RMA
3. Create a Customer Return for that RMA.
4. Create a reimbursement

This leaves an order in the "balance due" state, which is wrong. We are not expecting more money from the customer since they have returned their item. Something needs to reflect this before this can be merged.

jordan-brough:
Please also see https://github.com/solidusio/solidus/issues/1254

PR #1559, opened 2016-10-25
Store credit originator link

### LINKED ISSUE #1471: Link Store Credit Usage to Payment
When a store credit is verified, authorized, or used it is not linked to an originator in the `Spree::StoreCreditEvent`. This makes it difficult for an admin to find which order/payment the store credit was consumed under.
# Current

image
# Desired

image

gmacdougall:
Add originator to gateway options

Adding the originator of the Payment in to the gateway options allows store credit events to bind their originator correctly providing better reporting.

Fixed #1471

### DIFF
--- a/core/app/models/spree/payment/processing.rb
+++ b/core/app/models/spree/payment/processing.rb
@@ -94,7 +94,11 @@ module Spree
           # payment gateways happy.
           #
           # For more information, please see Spree::Payment#set_unique_identifier
-          order_id: gateway_order_id
+          order_id: gateway_order_id,
+          # The originator is passed to options used by the payment method.
+          # One example of a place that it is used is in:
+          # app/models/spree/payment_method/store_credit.rb
+          originator: self
         }
 
         options[:shipping] = order.ship_total * 100
--- a/core/app/models/spree/payment/processing.rb
+++ b/core/app/models/spree/payment/processing.rb
@@ -85,15 +85,17 @@ module Spree
 
       def gateway_options
         order.reload
-        options = { email: order.email,
-                    customer: order.email,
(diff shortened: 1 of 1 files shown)


core/app/models/spree/payment/processing.rb
```diff
@@ -94,7 +94,8 @@ def gateway_options
           # payment gateways happy.
           #
           # For more information, please see Spree::Payment#set_unique_identifier
-          order_id: gateway_order_id
+          order_id: gateway_order_id,
+          originator: self
```
gmacdougall: I am going to look in to this more. There some confusing usage between `originator` and `action_originator` throughout the code.

gmacdougall: On further review the complex object is what's expected further down the chain:
https://github.com/solidusio/solidus/blob/v2.0.0/core/app/models/spree/store_credit.rb#L137

I am also not a fan of the generic term originator but that's what's expected here:
https://github.com/solidusio/solidus/blob/v2.0.0/core/app/models/spree/payment_method/store_credit.rb#L36

core/app/models/spree/payment/processing.rb
```diff
@@ -85,15 +85,18 @@ def cancel!
...
+          email: order.email,
+          customer: order.email,
+          customer_id: order.user_id,
+          ip: order.last_ip_address,
+          # Need to pass in a unique identifier here to make some
+          # payment gateways happy.
+          #
+          # For more information, please see Spree::Payment#set_unique_identifier
+          order_id: gateway_order_id,
+          originator: self
```
gmacdougall: Thanks for the reminder. Comment added.

PR #1560, opened 2016-10-26
Create a placeholder source form for store credit

### LINKED ISSUE #193: Missing partial for store credits in admin.
Should add something at `spree/admin/payments/source_views/_storecredit` otherwise store credit payments will result in server errors when viewing them in the admin.

bbuchalter:
# 193 reports that if you

enable the Store Credit payment method in the admin, you will be unable
to access the new payments page for an order.

This PR creates a form so a ActionView::MissingTemplate exception isn't
raised. Additionally, it explains that create a store credit payment via
the admin is not currently supported.

Fixes #193 

screen shot 2016-10-26 at 8 50 22 am

### DIFF
--- a/backend/app/views/spree/admin/payments/source_forms/_storecredit.html.erb
+++ b/backend/app/views/spree/admin/payments/source_forms/_storecredit.html.erb
@@ -1,0 +1,1 @@
+<%= t('.not_supported') %>
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -698,44 +698,23 @@ en:
     adjustment_total: Adjustment Total
     adjustments: Adjustments
     admin:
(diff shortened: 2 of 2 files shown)


bbuchalter:
> Not sure if it's worth having a translation for this message?

Easy enough to do, but I wouldn't know the proper namespace. Any suggestions?

jhawthorn:
> Easy enough to do, but I wouldn't know the proper namespace. Any suggestions?

We've started using the implicit leading `.` for strings we expect to only use once.

So `t('.not_supported')` would give us `spree.admin.payments.source_forms.storecredit.not_supported` (a bit long, but that's fine)

jhawthorn:
Thanks Brian

PR #1562, opened 2016-10-27
use ternary operator instead of if else block

chukitow:
I think that is cleaner use ternary operator instead of **if else  block**  for this simple condition.

### DIFF
--- a/frontend/app/views/spree/products/_cart_form.html.erb
+++ b/frontend/app/views/spree/products/_cart_form.html.erb
@@ -29,7 +29,7 @@
     <% end %>
 
     <% if @product.price_for(current_pricing_options) and !@product.price.nil? %>
-      <div data-hook="product_price" class="columns five <% if !@product.has_variants? %> alpha <% else %> omega <% end %>">
+      <div data-hook="product_price" class="columns five <%= !@product.has_variants? ? 'alpha' : 'omega' %>">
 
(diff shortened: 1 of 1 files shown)


frontend/app/views/spree/products/_cart_form.html.erb
```diff
@@ -29,7 +29,7 @@
     <% end %>
 
     <% if @product.price_for(current_pricing_options) and !@product.price.nil? %>
-      <div data-hook="product_price" class="columns five <% if !@product.has_variants? %> alpha <% else %> omega <% end %>">
+      <div data-hook="product_price" class="columns five <%= @product.has_variants? ? 'alpha' : 'omega' %>">
```
mamhoff: I think this is a fine change in principle, but the previous code had a `!` before the main condition, turning the two branches around. Maybe rather: 

```ruby
@product.has_variants? ? 'omega' : 'alpha'
```

chukitow: You are right, I thought this was going to be my first PR without comments hahaha, I'll change it ASAP

cbrunsdon: ha, nice catch Martin.

chukitow:
I already did the change that you requested @mamhoff 😬

PR #1570, opened 2016-11-01
Avoid re-creating tax adjustments

jhawthorn:
Previously, `ItemAdjuster#adjust!` would always destroy all tax adjustments on the item and recreate them. This was unnecessarily slow in the common case of needing no changes (especially following #1479).

This changes `ItemAdjuster#adjust!` to instead update the adjustment for a rate if it already exists, create adjustments if they are missing for a matching rate, and destroy adjustments for rates which are no longer matching.

With this change, the speed of an `order.update!` which requires no changes is comparable to before #1479 :tada: 

[Comparison of queries run](https://gist.github.com/jhawthorn/6807bcb46871c19b674850e2a25894dd)

### DIFF
--- a/core/app/models/spree/tax/item_adjuster.rb
+++ b/core/app/models/spree/tax/item_adjuster.rb
@@ -20,12 +20,31 @@ module Spree
         @rates_for_default_zone = options[:rates_for_default_zone]
       end
 
-      # Deletes all existing tax adjustments and creates new adjustments for all
-      # (geographically and category-wise) applicable tax rates.
+      # This updates the amounts for adjustments which already exist and
+      # creates and remove adjustments as needed to match the applicable
+      # (geographically and category-wise) tax rates.
       def adjust!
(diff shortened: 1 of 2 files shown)


jhawthorn:
One potential issue I've spotted is that this has changed the behaviour of an existing tax rate which had an amount and then later became`0` (but otherwise still matched). Previously, it would have been removed after becoming 0. This code, in it's current state, will leave the adjustment, but update the amount to be `0`. I'd like to change this so that it is consistent (it shouldn't have a "memory" of it's past tax rates).

Another (preexisting) issue is when `order_tax_zone(order)` has a zone but later becomes `nil`. `adjust!` will exit early, but it should probably remove all tax rates in that case.

peterberkenbosch:
I actually like the 0 amount better then no adjustment at all, there is a difference accounting-wise between 0 amount tax and no tax.

peterberkenbosch:
👍

PR #1571, opened 2016-11-02
Document promotion availability

dangerdogz:
This adds tooltips to the `starts_at` and `expires_at` promotion fields. It might not be obvious that if no values is set, the promotion will be immediately usable.

screen shot 2016-11-01 at 8 24 51 pm

screen shot 2016-11-01 at 8 26 20 pm

screen shot 2016-11-01 at 8 37 57 pm

### DIFF
--- a/backend/app/views/spree/admin/promotions/_form.html.erb
+++ b/backend/app/views/spree/admin/promotions/_form.html.erb
@@ -66,12 +66,12 @@
 
     <div id="starts_at_field" class="field">
       <%= f.label :starts_at %>
-      <%= f.text_field :starts_at, :value => datepicker_field_value(@promotion.starts_at), :class => 'datepicker datepicker-from fullwidth' %>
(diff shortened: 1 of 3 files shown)


backend/app/views/spree/admin/promotions/_form.html.erb
```diff
@@ -66,12 +66,14 @@
 
     <div id="starts_at_field" class="field">
       <%= f.label :starts_at %>
-      <%= f.text_field :starts_at, :value => datepicker_field_value(@promotion.starts_at), :class => 'datepicker datepicker-from fullwidth' %>
+      <%= f.field_hint :starts_at %>
+      <%= f.text_field :starts_at, :value => datepicker_field_value(@promotion.starts_at), :placeholder => "Immediately", :class => 'datepicker datepicker-from fullwidth' %>
```
tvdeyen: Please pass the placeholder value through `I18n.t` and preferable use a key with a leading period, so the key gets scoped to this particular view.

dangerdogz: Meant to do it and forgot, thanks

PR #1573, opened 2016-11-02
Allow title separator to be customized

rsiddle:
Companies may wish to use other types of separators for their `<title>` tags for branding purposes. This pull request keeps the default '-', however it allows others to be chosen through an initializer.

Examples
```
<title>Ruby on Rails Jr. Spaghetti - Site Name</title>
<title>Ruby on Rails Jr. Spaghetti | Site Name</title>
<title>Ruby on Rails Jr. Spaghetti at Site Name</title>
<title>Ruby on Rails Jr. Spaghetti @ Site Name</title>
```
Example overriding
```
# spree.rb
Spree.config do |config|
  config.title_site_name_separator = '|'
end
```

### DIFF
--- a/core/app/models/spree/app_configuration.rb
+++ b/core/app/models/spree/app_configuration.rb
@@ -62,6 +62,10 @@ module Spree
     #   @return [Boolean] When true, site name is always appended to titles on the frontend (default: +true+)
     preference :always_put_site_name_in_title, :boolean, default: true
 
+    # @!attribute [rw] title_site_name_separator
+    #   @return [String] When always_put_site_name_in_title is true, insert a separator character before the site name in the title (default: '-')
+    preference :title_site_name_separator, :string, default: '-'
+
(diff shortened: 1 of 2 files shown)


mamhoff:
I think this should be just customized by overriding the relevant view, and that method is a little superfluous anyways.

Don't get me wrong, I think this is well-executed and would be alright, but I'd rather limit the amount of configuration and helpers and instead just put things like that in templates.

tvdeyen:
I'm also pro templates, besides that this helper is optional and it's easy to write you're own. I don't see any need to introduce a new config option here. 

Thanks for your contribution, but we plan to reduce front specific code for future versions of solidus.

Sorry

jhawthorn:
Thanks for the pull request, but I'm afraid I agree with the others here.

This is probably best done by overriding `spree/shared/_head.html.erb` and replacing the `<title>` with `title_string` surrounded by the desired look.

PR #1574, opened 2016-11-03
Admin product edit view bootstrap grid upgrade

omnistegan:
This PR upgrades the product edit view to utilise Bootstrap grids based on designs by @Mandily. #1290 

Changes include:
- Implemented a few new styles to fulfil design requirements including form-sections, font-awesome input add-ons, and a left-aligned xl width layout.
- Rendered tabs full-width always. This will allow the tabs to remain stationary across views once they are migrated to the admin-layout styles introduced in #1114 
- Responsive Bootstrap Grid form views for the product edit view.

I've attached a few screenshots here.

Product without variants:
fullwidth
Product with variants:
with_variants

Below 992px
small_width

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/plugins/_select2.scss
+++ b/backend/app/assets/stylesheets/spree/backend/plugins/_select2.scss
@@ -213,3 +213,9 @@ label .select2-container {
     }
   }
 }
+
+// Set the width of placeholders to 100% to force respecting parent width
+// https://github.com/t0m/select2-bootstrap-css/issues/42
+input.select2-input[placeholder] {
+  width: 100% !important;
+}
--- a/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
@@ -27,6 +27,10 @@ textarea {
   line-height: 19px;
 }
 
+input.datepicker {
+  cursor: pointer;
+}
+
 .fullwidth {
   width: 100%;
 }
--- a/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
@@ -306,3 +306,27 @@ fieldset {
 .form-buttons {
   text-align: center;
 }
+
+.form-section {
+  margin: 10px 0px;
+  padding: 10px 0px;
+  background: very-light($color-3, 4);
+  border: 1px solid very-light($color-3, 32);
+}
+
+.text-field-addon {
+  position: absolute;
+  right: 25px;
+  bottom: 21px;
+  pointer-events: none;
+}
+
+.max-width-form {
+  @include margin(null null auto null);
+  max-width: map-get($container-max-widths, "xl");
+}
+
+.helpful-text-sidebar {
+  margin: 10px 0px;
+  padding: 10px 0px;
+}
--- a/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
@@ -308,6 +308,7 @@ fieldset {
 }
 
 .form-section {
+  @include make-row(0px);
   margin: 10px 0px;
   padding: 10px 0px;
   background: very-light($color-3, 4);
--- a/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
@@ -34,6 +34,11 @@ body {
     @include margin(null auto);
     max-width: map-get($container-max-widths, "xl");
   }
+
+  &.max-width {
+    @include margin(null null auto null);
+    max-width: map-get($container-max-widths, "xl");
+  }
 }
 
 .content {
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -5,16 +5,13 @@
     <div class="form-section">
 
       <% if !@product.has_variants? %>
-        <div class="col-xs-12">
-          <div class="row">
-            <div data-hook="admin_product_sku" class="col-xs-12 col-md-6">
-              <%= f.field_container :sku do %>
-                <%= f.label :sku, Spree::Variant.human_attribute_name(:sku) %>
-                <%= f.text_field :sku, :size => 16 %>
-              <% end %>
-            </div>
-          </div>
+        <div data-hook="admin_product_sku" class="col-xs-12 col-md-6">
+          <%= f.field_container :sku do %>
+            <%= f.label :sku, Spree::Variant.human_attribute_name(:sku) %>
+            <%= f.text_field :sku, :size => 16 %>
+          <% end %>
         </div>
+        <div class="col-md-6 hidden-dm-down"></div>
       <% end %>
 
       <div data-hook="admin_product_form_name" class="col-md-6 col-xs-12">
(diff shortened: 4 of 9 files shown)


tvdeyen:
This is awesome

backend/app/assets/stylesheets/spree/backend/plugins/_select2.scss
```diff
@@ -213,3 +213,9 @@ label .select2-container {
     }
   }
 }
+
+// Set the width of placeholders to 100% to force respecting parent width
+// https://github.com/t0m/select2-bootstrap-css/issues/42
+input[placeholder] {
```
omnistegan: Good suggestion, updated.

backend/app/views/spree/admin/products/_form.html.erb
```diff
@@ -1,194 +1,250 @@
...
-      </div>
-    </div>
-
-    <div class="right col-xs-3" data-hook="admin_product_form_right">
-      <div data-hook="admin_product_form_price">
-        <%= f.field_container :price do %>
-          <%= f.label :price, class: 'required' %>
-          <%= f.text_field :price, :value => number_to_currency(@product.price, :unit => ''), :required => true %>
-          <%= f.error_message_on :price %>
+<div class="container-fluid">
```
mtomov: Is this `container-fluid` really needed here? If it is needed for every view, is it not possible maybe move it to the admin layout, so that we don't have to nest all views within it, and add extra indentation.

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -306,3 +310,36 @@ fieldset {
...
+  @include margin (null auto);
+  max-width: map-get($container-max-widths, "xl");
+}
+
+.left-aligned-form {
+  @include margin (null null auto null);
+  max-width: map-get($container-max-widths, "xl");
+}
+
+.helpful-text-sidebar {
```
graygilmore: Is this for any sidebar? We might want to tweak our naming here.

backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -31,6 +31,11 @@ body {
     @include margin(null auto);
     max-width: map-get($container-max-widths, "xl");
   }
+
+  &.left-aligned {
```
graygilmore: I'm guessing the purpose of this is to add the `max-width` to it? I wonder if we can come up with a different name as everything* is left aligned by default.

*almost everything

backend/app/views/spree/admin/products/edit.html.erb
```diff
@@ -6,6 +8,7 @@
   <% end %>
 <% end %>
 
+
```
graygilmore: Empty line was added here.

backend/app/views/spree/admin/products/new.html.erb
```diff
@@ -6,60 +6,73 @@
...
-      <%= f.text_field :name, :class => 'fullwidth title', :required => true %>
-      <%= f.error_message_on :name %>
-    <% end %>
+    <div class="row">
+      <div class="col-lg-8 col-xs-12">
+        <div class="row">
+          <div class="col-xs-12">
+            <div class="row">
+              <% unless @product.has_variants? %>
+                <div data-hook="new_product_sku" class="col-md-6 col-xs-12">
```
graygilmore: This kind of nesting of rows and columns really concerns me. I'll load this up locally and see if there's anything we can do.

jhawthorn: I see above we have `col-xs-12`, which is the full width, and can probably be removed

omnistegan: The problem is with holding less than full-width columns on a row by themselves.
Here we want the SKU field to always be on it's own line visually, but if the SKU div and the Name div are in the same `row`, they will sit on the same line (at medium and above).

Some better spacing here could make the grid structure read clearer.

jhawthorn: We probably want to find a way not to need to do this. We should only use row where it is needed

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
jhawthorn: IMO the name is fine

backend/app/views/spree/admin/products/_form.html.erb
jhawthorn: There's no reason this page should have different outer padding than our other pages, so this should either be removed or applied to all pages.

backend/app/views/spree/admin/products/_form.html.erb
```diff
@@ -1,194 +1,250 @@
...
+  <div data-hook="admin_product_form_fields" class="row">
+    <div data-hook="admin_product_form_left" class="col-lg-8 col-xs-12">
+
+      <div class="row form-section">
```
jhawthorn: I notices that this has CSS applied strangely.

`.row` as defined by bootstrap has a margin left and right of `-0.9375rem`, which patches a padding of `0.9375rem` on the columns.

`.form-section` has `margin: 10px 0px`.

So adding these two together adds padding in a non-obvious way. We instead probably want to add some padding to `.form-section` and avoid using `row` here at all.

omnistegan:
Here are a few of the changes I was able to make thanks to all of your feedback. I hope I was able to address some of your concerns.

- Removed the `.container-fluid` from the edit form. Bootstrap encourages the use of a `container` or `container-fluid` as a parent of row classes. It may be possible to place one in the admin layout, but there are not any issues I could find placing rows inside a fieldset. That is, using row classes at the top level of the form partial and not having a container or container-fluid anywhere on the page.

- Used bootstrap mixin for `.form-section`. This way we can override bootstrap's gutter and set our own margins. Because the form-section itself is placed inside a bootstrap column, the row mixin needs to exist for nested columns to work.

- Position `.text-field-addon` absolutely.

- Reworked empty space to be filled with empty bootstrap columns rather than nesting rows.

e.g.
```html
<div class="row">
  <div class="col-md-6 col-xs-12">
    Some Content
  </div>
  <div class="col-md-6 hidden-sm-down"></div>
</div>
```
instead of
```html
<div class="row">
  <div class="col-xs-12">
    <div class="row">
      <div class="col-md-6 col-xs-12">
        Some Content
      </div>
    </div>
  </div>
</div>
```

backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
omnistegan: I've updated the `left-aligned` classes (this one, and [the form style](https://github.com/solidusio/solidus/pull/1574/files#diff-b7b0ef96f04e3ccff70950db8525b1deR329)) to be named `max-width`.

This seems to be in line with the function they serve and how they are being discussed in #1290 and other related comments.

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -306,3 +310,28 @@ fieldset {
...
+  background: very-light($color-3, 4);
+  border: 1px solid very-light($color-3, 32);
+}
+
+.text-field-addon {
+  position: absolute;
+  right: 25px;
+  bottom: 21px;
+  pointer-events: none;
+}
```
mtomov: Here, I think we should also rely [on the build in](https://v4-alpha.getbootstrap.com/components/input-group/#basic-example) `input-group-addon`, again for the reason of free documentation, but as we want to have those _add-ons_  styled a tad bit different, we can add a supplementary class, say 

```css
.input-group-addon.simple {
  background-color: transparent;
  border-left: 0;
}
```

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -306,3 +310,28 @@ fieldset {
...
+
+.max-width-form {
+  @include margin(null null auto null);
+  max-width: map-get($container-max-widths, "xl");
+}
```
mtomov: Maybe I miss something, but I think this is pretty much the role of the [`.container`](https://v4-alpha.getbootstrap.com/layout/overview/#containers) class. If I've guessed right, it will be great to get to use that. Otherwise, let me know what could be the problem with using it.

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -306,3 +310,28 @@ fieldset {
...
+
+.helpful-text-sidebar {
+  margin: 10px 0px;
+  padding: 10px 0px;
+}
```
mtomov: I think that this one can also be classified as a `form-section` or `card`, just without background and border. So, maybe something like:

```css
.card.light {
  border: 0;
}
```

If we minimize those "public api" classes, it will be easier for other developers to make use of them.

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -306,3 +310,28 @@ fieldset {
...
   text-align: center;
 }
+
+.form-section {
+  @include make-row(0px);
+  margin: 10px 0px;
+  padding: 10px 0px;
+  background: very-light($color-3, 4);
+  border: 1px solid very-light($color-3, 32);
+}
```
mtomov: The use of this class closely resembles the `.card` [component from Bootstrap](https://v4-alpha.getbootstrap.com/components/card/#header-and-footer). Do you think it will be an idea to either override the default `.card` class with our background and border, or maybe add a supplementary class to it, like:

```css
.card.form-section {
  background: very-light($color-3, 4);
  border: 1px solid very-light($color-3, 32);
}
```

backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -34,6 +34,11 @@ body {
     @include margin(null auto);
     max-width: map-get($container-max-widths, "xl");
   }
+
+  &.max-width {
+    @include margin(null null auto null);
+    max-width: map-get($container-max-widths, "xl");
+  }
```
mtomov: This seems to duplicate `max-width-form`, or the other way around, but either way I think we should change the terminology, and either use the build-in `.container` or an adjusted derivative of it if we can't - like `.container.forms`, so to be clearer to others what the purpose of this class is, ie. it's already documented in Bootstrap.

backend/app/views/spree/admin/products/_form.html.erb
```diff
@@ -1,89 +1,176 @@
...
-        <%= render "spree/admin/shared/rebuild_vat_prices_checkbox", form: f, model_name: "product" %>
-        <div class="clearfix"></div>
-      <% end %>
+      <div data-hook="admin_product_form_cost_currency" class="col-lg-3 col-xs-6">
+        <%= f.field_container :cost_currency do %>
+          <%= f.label :cost_currency %>
+          <%= f.text_field :cost_currency %>
+          <%= f.error_message_on :cost_currency %>
+        <% end %>
+      </div>
```
mtomov: With the new additions by currency widget by @jhawthorn  in #1793, I think that all prices can now occupy a single line.

jhawthorn:
This no longer applies and has become stale. Closing in favour of #1818

PR #1578, opened 2016-11-08
Add integration spec to cover Order#outstanding_balance

jhawthorn:
There are some issues with this as it is currently, as well as issues with the proposed solutions.

This PR aims to add feature specs so we can narrow down what the issues are with each approach.


tvdeyen:
This has been part of #2002 and can therefore be closed.

PR #1579, opened 2016-11-08
Provide a consistent sort order in admin for shipment contents

jordan-brough:
This adds a `sort_by` so that line items in each shipment are listed
in a consistent sort order in the admin.

The shipments themselves already have a consistent sort order (for
multiple shipments in a single order).

### DIFF
--- a/backend/app/views/spree/admin/orders/_shipment.html.erb
+++ b/backend/app/views/spree/admin/orders/_shipment.html.erb
@@ -1,7 +1,11 @@
 <% admin_breadcrumb(plural_resource_name(Spree::Shipment)) %>
 
 
-<% manifest_items = Spree::ShippingManifest.new(inventory_units: shipment.inventory_units.where(carton_id: nil)).items %>
+<%
+  manifest_items = Spree::ShippingManifest.new(
+    inventory_units: shipment.inventory_units.where(carton_id: nil),
+  ).items.sort_by { |item| item.line_item.created_at }
+%>
 
 <div id="<%= "shipment_#{shipment.id}" %>" class="js-shipment-edit" data-hook="admin_shipment_form">
   <fieldset class="no-border-bottom">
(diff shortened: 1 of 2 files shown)


backend/spec/features/admin/orders/order_details_spec.rb
```diff
@@ -438,6 +438,33 @@
...
+          @first_line_item  = order.contents.add(product2.master)
+          @first_line_item.update_columns(created_at: 1.day.ago)
+          @last_line_item  = order.contents.add(product3.master)
+          @last_line_item.update_columns(created_at: 1.day.from_now)
+        end
+
+        it 'orders the items in a shipment by created_at' do
+          visit spree.edit_admin_order_path(order)
+
+          stock_items = page.all(:css, '.stock-item')
```
jhawthorn: `all` isn't safe to use in feature specs, since it's possible for it to return `[]` if the desired items aren't on the page (ex. not yet fully loaded).

`page.all(:css, '.stock-item', count: 3)` should work

jordan-brough: Update. Thanks!

jordan-brough:
@jhawthorn updated as per feedback

backend/app/views/spree/admin/orders/_shipment.html.erb
```diff
@@ -1,7 +1,11 @@
 <% admin_breadcrumb(plural_resource_name(Spree::Shipment)) %>
 
 
-<% manifest_items = Spree::ShippingManifest.new(inventory_units: shipment.inventory_units.where(carton_id: nil)).items %>
+<%
+  manifest_items = Spree::ShippingManifest.new(
+    inventory_units: shipment.inventory_units.where(carton_id: nil),
+  ).items.sort_by { |item| item.line_item.created_at }
```
bbuchalter: I like how you chose to sort this by when the line item was created. Exactly how I would expect it to be sorted. 👍

PR #1580, opened 2016-11-08
Allow users to inline update the variant of an image in admin

mtomov:
This should allow users to update the variant of an image directly within the admin images table.

I have pretty much copy/pasted the code from the [`payments/edit`](https://github.com/solidusio/solidus/blob/94f630fd4ca31f52a8ed72956db0259668144796/backend/app/assets/javascripts/spree/backend/payments/edit.js.coffee) class, so thanks @jhawthorn for that! Indeed a great learner on syncing Backbone's models with the backend.

Not sure about the dotted underline below the small icons .. maybe I can remove that, but couldn't find an appropriate class for the purpose.

inline image variant update


tvdeyen:
Great feature. Didn't had the time to loon into the implementation details, but this looks good on a first glance. 

One thing coming to mind, though. From a UX point of view I'm asking myself why we don't always show the select box With the current one preselected. This spares some clicks and we could get rid of the extra icons. 

Thoughts @Mandily ?

jhawthorn:
I'm a little concerned that having two edit buttons will be confusing to users. @tvdeyen's suggestion would solve this.

Mandily:
I have a couple of thoughts on preselecting values in dropdowns. 
1. It's great if most of the time the user would select the same option anyway. This saves time, and cognitive load.
2. It's not so great when you want a user to actually think about what's in a box. You can risk them skimming over it and setting something up improperly. 

Without doing a full in depth exercise, I suspect that there's a good deal of dropdowns that would be suitable for a default value. 

With regards to the edit functionality of the tables - I think this is a great idea and I secretly thought it was going to take a lot longer before we'd start to see more of this added into the admin. So I'm excited you're doing it now @mtomov. 

Instead of using additional icons within the tables, it would be nice to use the same convention that the product stock screen uses - swapping out the edit and delete buttons for a checkmark and X. I think it streamlines both the interaction and the UI.
screen shot 2016-11-14 at 4 50 15 pm
 
In the future, we could take this one step further and make everything in the page editable without having to go into an edit mode, and automatically save changes on field blur. But in the meantime I'm more than happy to see edit functionality being merged into the view pages themselves.

mtomov:
Thanks @Mandily and the others for the feedback. It'll be great indeed if we have a _standard_ for those. 

I just didn't fully understand what the final look would be with the _checkmark_ and _X_buttons - do they go next to the current _Edit_ and _Delete_ buttons, is the select box always there available for change (no need to press Edit as Thomas suggested)?

The only drawback of having the selectbox immediately there available is that it might look a bit ugly, as the deep blue on those select boxes puts a great focus on the area, as if though something is off .. 

Naturally, it's actually easier indeed from a programming point to have the select box directly there available .. so it's really _up to you_ on how you're best seeing this.. as either option is sort of the same programming effort if you are concerned on that ..

Mandily:
The _checkmark_ and _X_ buttons replace the _edit_ and _delete_ buttons when the user is in **edit mode**. Clicking on _edit_ makes the whole row editable. Once the _checkmark_ or _X_ is clicked, the row goes back to view only, the edits are saved, and the _edit_ and _delete_ buttons appear again.

I don't think we should avoid introducing a new and improved interaction model because a visual style is too garish. At some point I'd either like to update the select 2 inputs to v4 or restyle them ourselves with the rebrand. @jhawthorn might be better able to speak to that though.

I realize this would also include a visual style change in the tables, but the best way I've seen this editable table thing done is that when a user hovers over a table row, it gets highlighted with the fields marked in white. 

I believe this is the best model, but before we go that way, I would want to understand if it's something that you think is doable across Solidus as a whole in a relatively short time period, because I would hate for a transition like this to take 4 years to complete.
screen shot 2016-11-15 at 3 17 54 pm

mtomov:
Amanda, the hover suggestion was brilliant, did it, wasn't difficult at all actually, hopefully can also be used by other similar in-table editing functionalities.  

With the hover, it's quite incredible how quickly one could change the variant on a whole bunch of images. 

- e86a884 was the actual commit with the change 



#### on hover
screenshot from 2016-12-01 23 07 37

#### on click
screenshot from 2016-12-01 23 07 47

.. finally the user needs to click on the check icon to confirm the change

mtomov:
I've actually gone about to re-write the javascript code for inline image update into a generic reusable `Editable Table` "plugin" of sorts, so that tables can be turned into inline editable only with html markup, similarly to the _sortable_ plugin and such.

It's all in the last commit, and as an example, to do it on a text field element in cell, one would need to create such markup:

```
  <td>
    <span class="editing-hide">
      <%= image.alt %>
    </span>

    <%= fields_for image do |f| %>
      <%= f.text_field :alt, class: 'editing-show' %>
    <% end %>
  </td>
```

Now, to eliminate a bit of duplication here - an improvement could be to check if the cell has an element with `editing-hide`, and if not - then create it using the value of the input (select) field.  That would reduce the html needed for such table to just an editable input box with a class of `editing-show`

Another good improvement is to have the `Cancel` button revert the input box value it its original one. Should be easily doable though this one.

In this particular case for the image - I don't think we need to have the `alt` box be inline-editable, but have changed it for demonstration purposes. Might have to add a test if we are to keep it.

screenshot from 2016-12-05 14 40 59

-- 

I think the option types/values admin section could be a good candidate for this kind of UI

Looking forward to any input if that is in the right direction. Thank you!

Mandily:
I downloaded your code and played around with the new functionality. I like it a lot! 

It would be nice if the styles between view and edit mode were the same, but I'm not sure if that's something we should tackle with this or leave it for the future rebrand. Especially as we can't do much with the dropdowns right now. What's your take on that? This might even be something I can do myself if you didn't want to get into the finicky css stuff.

I know we technically don't need a save button on this page because we're auto saving, but I think we need to keep the save button on every page for peace of mind for the user. In the future, I'd like to introduce states to buttons, so if the page was saved, the button would look disabled, whereas if it needed to be saved, it would be brighter to attract attention. We don't need to do that for this feature, but I'd like to keep the save button here for future use.

The third thing I wanted to bring attention to was that we don't have a stop gate for catching the user while in the middle of edit mode. If they start to edit a row and don't click on the check or x, and then navigate to another page, changes are lost. This is a behaviour that is consistent across Solidus, but I'd like to do something to remedy it in the near future: I see three options for this:
- Auto save rows when a user clicks outside of the row. This would negate the need for check and X buttons
- notify the user when they try to leave the page with a modal window.
- both of the above

I'm a little more on the side of the adding this kind of functionality here than the save button, but I could be persuaded to add it in later if you had a good reason to leave it. 


WRT your most recent post, I like the idea of keeping the alt text an editable field. I was actually trying to figure out how we could do edit the thumbnail image too, but I think that gets more complicated than we want to get at this point. The good thing about making every field in a table editable is that we can then get rid of the edit pages. (WOOHOO!) I think the option types page you mentioned and a good handful of others would benefit from users being allowed to edit right in the table and getting rid of the edit pages altogether. In some cases we might have to add a few more columns to the table, but it shouldn't be too big of a deal. 

Let me know if you want to hash out any more ideas or need help identifying those pages. I've got an ongoing list that I can add those "optional" features I mentioned above, so let me know you're thoughts on the stop-gate solutions so I can log them in there if you decide to do them later.

Thanks so much for taking this on!

tvdeyen:
> The good thing about making every field in a table editable is that we can then get rid of the edit pages. (WOOHOO!)

I don't think this a working for all tables. Often tables are just a list with little detail and the forms have way more options than ever will fit into a table row. Think of the product or order tables. For small tables I like the idea, though. Although I don't like auto save. What if the user change its mind or mistyped something? 

I like the current state with save and cancel buttons, although I think they should look different from the edit and delete buttons. It's very hard to tell for the user if the buttons currently visible are the edit buttons or save buttons. Maybe we should hide all buttons but the ones from the hovered row? Or we highlight the buttons if we are currently editing the row. I kinda like the first option. Would remove clutter from the views.

Mandily:
I agree, I don't think this would be appropriate for all tables, but any of the smaller ones where all data can be represented in a single row should eventually be moved over to something like this. To clarify -  it's EVERY FIELD in a small table, not EVERY TABLE.

Contrary to your auto-save point - if the user mistyped something, they can just go back in and edit it again. It's not like we're asking them to confirm with a dialog when they save it anyway, so that extra step isn't really reducing the risk of human error. 

I was also thinking on the styles that the check and X are presented in vs the edit buttons. If we remove the other row's buttons, we do remove the clutter, but we also remove the user's ability to edit more than one row at a time. At first I was thinking that would be a good idea, but the more I think on it, the more I think we need to give users the power and freedom to edit these things as quickly and easily as possible. I can't say for sure, but I imagine that when a user needs to edit something on one of these pages, it's probably more than just one row. This is also why I like the auto-save feature. It allows users to be more efficient in entering/changing their data.

In short - Solidus users are regular users of this software, I don't think we need to protect them from themselves as much as we would with a software that many people use only a few times.

mtomov:
Thanks for your comments!

Another round of updates in 0ac12bc focusing on simplification - have removed all the custom classes, reducing everything to a single `inline-editable` that need be applied to the table. Have also removed the double specification in the cell - we are left with a single input/select box in the cell. 

On suggestion from Thomas - have removed all buttons on initial view. On hover - the Edit and Delete buttons come up. On Edit - the Update and Cancel buttons come up.

I've dug out nice row context colours from spree's code depending on the button that you hover over. Have attached some screens as you might've only seen the Destroy (red) context colour and not the others. Hopefully that would help with the confusion between the buttons that you mentioned.

Finally, have added a real `cancel` functionality that I mentioned before would be nice to work.

hover and active

*Context colours:*

edit

cancel

save

Mandily:
By any chance can we get the cancel tooltip to be orange to match the rest? The red is associated with delete/destroy in my mind.

Rob117:
Is this not getting merged? If so, can we close it?
It looks useful, but leaving it open in the branches without comment for a month doesn't seem to be doing it justice.

mtomov:
I've rebased it onto latest master. Let me know what else I can do to demonstrate the achieved improvement.

mtomov:
Hello,

Although not a brand new PR, in the latest commit after speaking with @Mandily and @tvdeyen, I've:

1. Updated the styles to match the new Select2 styles
2. Added recognition of `Enter` and `Esc` keys when inline-editing a table row, as well as tests for both events
3. Have put back the Edit and Delete buttons to always be visible to keep the interface the same as existing tables
4. Re-organized the code to match the new Backbone Views organization


Thank you for the feedback!

screenshot from 2017-05-22 22 34 28

Mandily:
This looks great! I love how a small code change can make such a big difference to the usability of the admin.

@mtomov Eventually I'd like to get to the point where we remove the buttons on the side and only show them on hover. Do you have any ideas/timeline on when we should implement that part? I'm also curious to know what the plan is for applying this code to the rest of the tables in the Solidus admin.

mtomov:
Hi @Mandily - happy that you find it usable!

1. Remove the buttons and show them on hover - technically, doing that requires 1 line of css change. As I've noted above, I did have the right hand side buttons showing only on hover, and to me it looked good, but I've switched them back to always show as this was your preference. 

2. Ideas / timeline - that's all up to you

3. Plan for applying to other tables - I think someone must figure out what other tables could use a similar functionality, as they aren't that many suitable I think. Then, create an issue to describe the task, which someone might pick and make the code changes.

Thanks @mamhoff for reviewing!

mtomov:
Come on, one more approve : )

Let's see if the new members would be more easily wooed  @kennyadsl

Mandily:
I will make a new issue in the next couple of days with other tables and link back to this one.

mtomov:
Brilliant, thanks a lot @tvdeyen  - all great spots!

I think I've now addressed them all, and have squashed all my work into a single commit, which should describe it all.

tvdeyen:
🎊 Thanks!

I made a followup https://github.com/solidusio/solidus/issues/2042

mtomov:
Thanks all for helping out!

PR #1582, opened 2016-11-08
Styleguide form rules

Mandily:
This PR is to add the details from #1290 into the style guide. Layout details have been purposely left out as they can be added in with the PR's adding those layouts into core.

These are the details for how to implement each section, not examples of what they look like. If we'd like to include examples, a volunteer to assist in doing that would be helpful as I don't know code as well as I do design.

screen shot 2016-11-08 at 2 31 06 pm
screen shot 2016-11-08 at 2 32 04 pm
screen shot 2016-11-08 at 2 32 15 pm

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/sections/_style_guide.scss
+++ b/backend/app/assets/stylesheets/spree/backend/sections/_style_guide.scss
@@ -26,7 +26,8 @@
     @include flex(1);
   }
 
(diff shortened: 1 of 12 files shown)


isaacfreeman:
The value of a style guide isn't measured by how often it's referenced, but by how well it reduces friction everywhere else.

PR #1583, opened 2016-11-09
Simplify image upload UI

jhawthorn:
This replaces the new image upload area (previously a paragraph of text and a file upload field) with a single button inside the drop area with the text "Browse or drag files here to upload".

The original problem I set out to fix was that after uploading an image through the file input, the filename remained. I added some javascript to clear the field after uploading, which worked great, but now there was always the text "No file selected". After this I realized the entire area was a label and we can just hide the upload field.

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/sections/_products.scss
+++ b/backend/app/assets/stylesheets/spree/backend/sections/_products.scss
@@ -37,10 +37,15 @@ div[data-hook="admin_products_index_search_buttons"] {
 }
 
 #upload-zone {
-  border: 10px dashed $color-border;
+  border: 2px dashed $color-border;
   padding: 2rem 5rem;
   text-align: center;
-  margin-bottom: 1.5rem;
+  margin: 1rem auto;
+  width: 70%;
+
+  label.upload {
+    display: block;
+  }
 
   &.with-images {
     border-color: $color-success;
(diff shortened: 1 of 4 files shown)


Senjai:
:+1:

jhawthorn:
> Could we also get rid of that ugly 10px dashed border and use a solid light gray background instead? (No offense @mtomov)

I think (strangely) a dotted border is the universal symbol for "upload area". I do agree the wide dashed border doesn't fit particularly well here though.

How about a smaller dashed border? I also shrunk the area.

tvdeyen:
much better. Thanks @jhawthorn 

Lots of apps (Trello, Redmine, Alchemy) use the whole "content area" as an upload area. Much better to hit with your cursor then fiddling around with a small area, though. But this is fine for now and looks ok for me. 

/c @Mandily

Mandily:
I agree with @tvdeyen, It would be nice for the droppable area to be the full width of the content area with a lighter grey background. But I'm also fine with leaving out colour details until we do a full pass over of the UI with the rebrand. 

I'm not sure what your rationale is for combining the two messages into the button, but I think it's works better as two distinct instructions instead of one joined. The button is not being used to drag and drop, so the CTA text inside should not be referencing it. So the button is used to browse, and then the message that the user can drag and drop is located outside of it. 

Disclaimer - I did not mock this up, I just found online
drag-drop-file-upload

jhawthorn:
Split text from button and added a large icon

Mandily:
Perfect! And thank you for putting a lowercase 'u' on upload. 👍

tvdeyen:
Lovely @jhawthorn 😍

tvdeyen:
Thanks @Mandily 🌹

PR #1584, opened 2016-11-10
Remove temporary_credit_card attr from Order

peterberkenbosch:
Introduced in 2014 with [this PR](https://github.com/solidusio/solidus/commit/e891bf56dcbe188dd485e1a95da8445c23a9000e)

However, the attribute is never set on any other place except in the checkout spec and this can be removed IMHO.


peterberkenbosch:
@magnusvk can you remember the use case for having the `temporary_credit_card` attr? Working on #1414 and I do not see why we should still need it.

jhawthorn:
I don't like this attribute and it makes especially little sense on Order, but shouldn't we have some way of declining to save a credit card to a user's wallet?

magnusvk:
👍 to John's comment.

peterberkenbosch:
yes, agree that we should have an active configuration option there. Right now only creditcards that have a profile are stored in the wallet. Might be good to add that in #1414 I guess.

tvdeyen:
> shouldn't we have some way of declining to save a credit card to a user's wallet

👍 Absolutely 

> we should have an active configuration option there

This should be an choice the user makes during checkout, neither we nor the store owner should decide on storing credit card tokens on behalf of the user without asking.

We should revert what was introduced in e891bf56dcbe188dd485e1a95da8445c23a9000e IMO and add a way to let users decide if they want to store there credit card for later re-use.

Long story short, this PR is valid and ok IMO, but needs a rebase now that we introduced the wallet payment source.

@peterberkenbosch would you rebase this? If not we could take over.

PR #1586, opened 2016-11-11
Deprecate default zone

mamhoff:
Instead of fully deleting the terrible code paths that lead to "tax reimbursements", this PR deprecates the core methods for those code paths. 

This supersedes #1537, which we'll pull through in Solidus 3.

### DIFF
--- a/core/app/models/spree/zone.rb
+++ b/core/app/models/spree/zone.rb
@@ -55,6 +55,8 @@ module Spree
     # A State zone wins over a country zone, and a zone with few members wins
     # over one with many members. If there is no match, returns nil.
     def self.match(address)
+      Spree::Deprecation.warn("Spree::Zone.match is deprecated. Please use Spree::Zone.for_address instead.", caller)
+
       return unless address && (matches =
                                   with_member_ids(address.state_id, address.country_id).
                                   order(:zone_members_count, :created_at, :id).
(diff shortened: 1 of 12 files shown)


core/app/models/spree/zone.rb
```diff
@@ -41,14 +41,22 @@ class Zone < Spree::Base
 
     self.whitelisted_ransackable_attributes = ['description']
 
+    # Returns the zone marked as `default_tax`.
+    # @deprecated Please run the `solidus:migrations:create_vat_prices` rake task
     def self.default_tax
-      where(default_tax: true).first
+      default_tax_zone = where(default_tax: true).first
+      if default_tax_zone
+        Spree::Deprecation.warn("Spree::Zone.default_tax is deprecated. Please run the updater rake tasks.", caller)
```
Senjai: The exact name of the rake task may be helpful here

core/app/models/spree/zone.rb
```diff
@@ -41,14 +41,22 @@ class Zone < Spree::Base
...
+        default_tax_zone
+      end
     end
 
     # Returns the most specific matching zone for an address. Specific means:
     # A State zone wins over a country zone, and a zone with few members wins
     # over one with many members. If there is no match, returns nil.
     def self.match(address)
+      Spree::Deprecation.warn("Spree::Zone.match is deprecated. Please use Spree::Zone.for_adress instead.", caller)
```
Senjai: for_address is misspelled

mamhoff:
Addressed the nits! Thanks @senjaj

Issue #1564, opened 2016-10-27
RMA amount not correct when the quantity of a line item is greater than 3


mdamian322:
Steps to recreate:
1. Purchase 3 of a product
2. Create New RMA
3. The “Amount Before Sales Tax” field is autofilled with with an amount that is 19.99000000000000003398. I’m expecting it to be $19.99

<img width="1351" alt="amounts-before-sales-tax" src="https://cloud.githubusercontent.com/assets/13877138/19789535/9cd4a5b0-9c6d-11e6-83da-c493e9fc6b22.png">

swcraig:
This was fixed by #1590. I can't recreate on master.

aitbw:
As per #1590, this one's fixed @kennyadsl :tada:

kennyadsl:
Thanks @DanielePalombo for fixing this!

PR #1590, opened 2016-11-15
Fix RMA amount calculator

DanielePalombo:
RMA calculator use the percentage to calculate the weighted line item
amount, but when the percentage is a repeating decimal the calculator
makes an error. Replaced the percentage calculation with a division.

ref #1564

### DIFF
--- a/core/spec/models/spree/calculator/refunds/default_refund_amount_spec.rb
+++ b/core/spec/models/spree/calculator/refunds/default_refund_amount_spec.rb
@@ -2,7 +2,7 @@ require 'rails_helper'
 require 'shared_examples/calculator_shared_examples'
 
 RSpec.describe Spree::Calculator::Returns::DefaultRefundAmount, type: :model do
-  let(:line_item_quantity) { 2 }
+  let(:line_item_quantity) { 3 }
   let(:line_item_price) { 100.0 }
   let(:line_item) { create(:line_item, price: line_item_price, quantity: line_item_quantity) }
   let(:inventory_unit) { build(:inventory_unit, order: order, line_item: line_item) }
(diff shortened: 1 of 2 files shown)


core/spec/models/spree/calculator/refunds/default_refund_amount_spec.rb
```diff
@@ -36,7 +36,7 @@
         order.adjustments.first.update_attributes(amount: adjustment_amount)
       end
 
-      it { is_expected.to eq line_item_price - (adjustment_amount.abs / line_item_quantity) }
+      it { is_expected.to be_within(0.1).of(line_item_price - (adjustment_amount.abs / line_item_quantity)) }
```
jhawthorn: Thanks for the PR.

Being within 0.1 (10 cents) is probably not good enough here. We should test for an exact amount.

This test passes both before and after the change made in `default_refund_amount.rb`. Is it possible to get a spec which clearly shows how the existing code was incorrect.

DanielePalombo: The problem is that the existing code use a percentage to calculate the value of each line item. This works without problem when quantity is factor of 100 , but in other cases this returns a repeating decimal percentage and the result is a little bit different.

Ex:
existing code: 

```
def percentage_of_line_item(inventory_unit)
  1 / BigDecimal.new(inventory_unit) 
end
...

30 * percentage_of_line_item(3) # 30 * 0.333333333333333333 = 9.99999999999999999
```
what we want is 10 not 9.99999999999999999

with fix we have:

```
30 / 3 # 10 
```

Replaced the test from `/ line_item_quantity` to `* ( line_item_price / line_item_total )` because the code https://github.com/nebulab/solidus/pull/4/files#diff-f271b0c846fcbc382f43fdfa7f6a0c89R24 use a percentage instead a division

core/spec/models/spree/calculator/refunds/default_refund_amount_spec.rb
```diff
@@ -36,7 +37,7 @@
         order.adjustments.first.update_attributes(amount: adjustment_amount)
       end
 
-      it { is_expected.to eq line_item_price - (adjustment_amount.abs / line_item_quantity) }
+      it { is_expected.to eq line_item_price - (adjustment_amount.abs * ( line_item_price / line_item_total )) }
```
cbrunsdon: Can you replace this call with the actual number its supposed to be, please. Doing it on the computed amount from the setup makes it tougher to figure out what should be happening (though this calculation would be nice to keep in a comment or something).

You can take a look at the tax specs and see how much more readability it adds when you're really trying to figure out whats happening: https://github.com/solidusio/solidus/blob/master/core/spec/models/spree/calculator/default_tax_spec.rb

DanielePalombo:
👍

DanielePalombo:
Ok, I'll rebase it next weekend.

DanielePalombo:
I rebased it.

PR #1592, opened 2016-11-17
remove cruft

kwstannard:
This PR should cause no changes to behavior.

with_deleted scope on line 14 handles what the delegation was created
for. original commit with explanation: https://github.com/solidusio/solidus/commit/7c89051efbed7066a830ab193408c1ae99559c51

I can't get build.sh to run, but I did get passing tests in core.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -38,10 +38,6 @@ module Spree
     before_destroy :destroy_inventory_units
 
     delegate :name, :description, :sku, :should_track_inventory?, to: :variant
-    # @note This will return the product even if it has been deleted.
-    # @return [Spree::Product, nil] the product associated with this line
-    #   item, if there is one
-    delegate :product, to: :variant
     delegate :currency, to: :order, allow_nil: true
 
     attr_accessor :target_shipment


bbuchalter:
I also could not get `build.sh` to run...if we're talking about removing some cruft, I think that file should go and some clear README updates around running individual components test suites. I've just fought that battle pretty hard, so I'd be game if that'd be welcome.

jhawthorn:
@bbuchalter @kwstannard does #1595 fix the issues you were having with `build.sh`, if not could you please file a bug with whatever error you are seeing.

mamhoff:
I think that this delegation has a different purpose than line the association on line 14. 

The association defines `line_item.variant`, while the delegate, depending on the association, defines `line_item.product` - a method I find to be worthwhile. It might make sense to add `allow_nil: true` so that the delegation doesn't fail when the line item, for some reason, does not yet have a variant associated. 

Rather than see this delegate go, I would want to see a spec for it. I do believe people might rely on it.

jhawthorn:
@mamhoff There's also a [has_one :product, through: :variant](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/line_item.rb#L17), which is what makes this delegate unnecessary. It will have the same behaviour as though this delegate had `allow_nil: true`.

mamhoff:
@jhawthorn Thanks, good by me then :)

PR #1595, opened 2016-11-18
Sample test_app fixes

jhawthorn:
Instead of running seeds inside of `rake test_app` (which no other projects do), load seeds inside of the spec run.

This should fix some users were having running `build.sh` on the latest versions of bundler.

cc @bbuchalter

### DIFF
--- a/core/db/default/spree/store_credit.rb
+++ b/core/db/default/spree/store_credit.rb
@@ -4,7 +4,8 @@ Spree::PaymentMethod.create_with(
   name: "Store Credit",
   description: "Store credit",
   active: true,
-  display_on: 'none'
+  available_to_admin: false,
+  available_to_users: false
 ).find_or_create_by!(
   type: "Spree::PaymentMethod::StoreCredit"
 )
--- a/sample/Rakefile
+++ b/sample/Rakefile
@@ -7,7 +7,6 @@ desc "Generates a dummy app for testing"
 task :test_app do
   ENV['LIB_NAME'] = 'spree/sample'
   Rake::Task['common:test_app'].invoke
-  Rake::Task['common:seed'].invoke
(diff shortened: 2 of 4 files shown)


bbuchalter:
Looks like this surfaced a deprecation warning which caused the test suite to fail.

These changes look good. 👍 

However, this does not the underlying issue I was facing which we can discuss in depth in another forum. Ping me on Slack any time if you wish to discuss.

core/db/default/spree/store_credit.rb
```diff
@@ -4,7 +4,8 @@
   name: "Store Credit",
   description: "Store credit",
   active: true,
-  display_on: 'none'
+  available_to_admin: false,
+  available_to_users: false,
```
tvdeyen: Forgot to remove a trailing comma

core/lib/spree/testing_support/common_rake.rb
```diff
@@ -32,6 +32,6 @@
   task :seed do |_t, _args|
     puts "Seeding ..."
 
-    sh "bundle exec rake db:seed RAILS_ENV=test"
+    sh "rake db:seed RAILS_ENV=test"
```
tvdeyen: Why, is `bundle exec` causing any problems?

Removing this could be cause of conflicting gem errors, using the `bin/rake` (or even better `bundle exec`, because of a possible `spring` binstub) would solve this, no?

jhawthorn: I don't think we should use `bundle exec` here because we're already in ruby, and presumably already in a `bundle exec`. I'll remove this commit and submit it separately for discussion.

kwstannard:
This worked for me thanks :)

Issue #1588, opened 2016-11-11
Checkout as guest cannot be completed successfully due to validation error


andreas-venturini:
Reproduce:

1. Follow the steps in the [Getting started guide](https://github.com/solidusio/solidus/tree/v2.0#getting-started)
2. Point your webbrowser to http://127.0.0.1:3000/
3. Add any item to the cart
4. Click on checkout
5. In the "Checkout as guest" input enter "guest@example.com"
6. Click continue
7. On the next page (the address step of the checkout process) go back to the cart e.g. by clicking on the cart symbol
8. From the cart click on checkout again

**What you're expecting to happen:**

I should be redirected to the first step of the checkout process (address)

**What's actually happening**

The following validation error occurs and I can no longer complete the checkout process without deleting my browser session:

![screenshot from 2016-11-11 19-53-04](https://cloud.githubusercontent.com/assets/4801917/20226462/818ec2a4-a848-11e6-89ea-8698c58d45dd.png)

[Link to Gemfile](https://gist.github.com/andreas-venturini/fabd0e90b908e61bca32e41f848f1e63)
[Link to Gemfile.lock](https://gist.github.com/andreas-venturini/3dad5a151e7501fb12522429d4ea0a80)

jhawthorn:
Thanks for the detailed report. It really does help.

I can confirm that this happens in a sandbox, and seems to be at least somewhat related to auth_devise. I'm running a bisect to try and figure out what caused this.

jhawthorn:
This should be fixed in the latest version of solidus_auth_devise, 1.6.2.

Thanks again @andreas-venturini

PR #1596, opened 2016-11-18
Only run setup_for_current_state in edit, update

jhawthorn:
See also https://github.com/solidusio/solidus_auth_devise/pull/76 and https://github.com/solidusio/solidus/issues/1588

Because that before_action was running on all actions of the controller,
the extra actions added by solidus_auth_devise had this run before them,
which wasn't desirable.

This has been fixed via a skip_before_action in solidus_auth_devise, but
we might as well restrict it here as well, since it only makes sense for
these two specific actions.


bbuchalter:
This change seems reasonable. Thanks for the incremental improvement here.

PR #1598, opened 2016-11-21
Add a getting started guide for the uninitiated

mamhoff:
This is the first of a series of guides aimed at people familiar with Rails.
While this particular guide encompasses much that's in the Readme, it provides
more links and support for people not familiar with Solidus.

### DIFF
--- a/guides/getting_started.md
+++ b/guides/getting_started.md
@@ -1,0 +1,100 @@
+# Installing Solidus
+
+This article will explain how to setup a new Rails application with Solidus installed. We assume familiarity with Ruby and Rails, but will link to further resources to help you understand what we are doing.
+
+## Installing ImageMagick
+
+Solidus uses the ImageMagick library for manipulating images. Using this library allows for automatic resizing of product images and the creation of product image thumbnails. ImageMagick is not a Rubygem and it can be a bit tricky to install. There are, however, several excellent sources of information on the Web for how to install it. A basic Google search should help you if you get stuck.
+
+If you are using OSX, a recommended approach is to install ImageMagick using [Homebrew](http://brew.sh/). This can be done with the following command:
+
+```bash
+$ brew install imagemagick
+```
+
+If you are using Unix or Windows check out [Imagemagick.org](http://www.imagemagick.org/) for more detailed instructions on how to setup ImageMagick for your particular system.
+
+## Installing Ruby
+
+Installing Ruby varies widely from architecture to architecture. There is a good overview on how to
+install Ruby on the [Ruby homepage](https://www.ruby-lang.org/en/documentation/installation/).
+
+Please make sure your ruby version is at least 2.2.2:
+
+```
+$ ruby -v
+ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]
+```
+
+## Installing Rails
+
+Ruby comes with the package manager `rubygems`. However, many Ruby projects also use the package manager `bundler`. You use rubygems to install Rails as well as Bundler. If your Ruby is setup correctly, this should be as easy as running the following command:
+
+```
+$ gem install rails bundler
+```
+
+## Generating your host app
+
+It's important to understand that Solidus is a [Rails engine](http://guides.rubyonrails.org/engines.html), and that your host app simply requires that engine. The first step for installing Solidus is therefore generating a vanilla Rails application.
+
+Rails default `sqlite` database isn't well suited to real Solidus production usage. Solidus will runs best with the `postgresql` (or `MySQL`) database.
+
+The ecosystem in general uses the `rspec` testing framework, so we skip the standard `minitest` installation. To generate your host app with these options, run the following command:
+
+```
+rails new my-shop -d postgresql --skip-test
+```
+
+Switch into your newly generated app directory:
+
+```
+cd my-shop
+```
+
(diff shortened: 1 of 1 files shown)


guides/getting_started.md
```diff
@@ -0,0 +1,113 @@
...
+```
+bundle exec rails g spree:install
+bundle exec rails g solidus:auth:install
+bundle exec rake railties:install:migrations
+```
+
+These tasks will setup your Rails app as a basic store, with an authentication system and a storefront
+and checkout. It will also provide some sample products and configurations. You will be asked to set an admin email/password combination. The default values are `admin@example.com` and `test123`, respectively.
+
+Now, run the rails server (`rails s`) and, in your browser, navigate to `localhost:3000`. Voila: You'll find an app running an online shop.
```
bbuchalter: `To access the store admin, visit localhost:3000/admin.`

mamhoff: amended, thank you!

guides/getting_started.md
```diff
@@ -0,0 +1,113 @@
...
+Switch into your newly generated app directory:
+
+```
+cd my-shop
+```
+
+Now, run the tasks to create your development and test databases. If this does not work, please consult the postgres installation instructions for your architecture and the Rails Guide on [Configuring Rails applications](http://edgeguides.rubyonrails.org/configuring.html#rails-general-configuration).
+
+```
+bundle exec rake db:create db:migrate
```
tvdeyen: Running the migrations has no effect on a vanilla rails appandcan be omitted here

guides/getting_started.md
```diff
@@ -0,0 +1,113 @@
...
+end
+```
+
+Now, run `bundle install` to download `solidus` and `solidus_auth_devise` into your new app, then run
+the following generators to get started:
+
+```
+bundle exec rails g spree:install
+bundle exec rails g solidus:auth:install
+bundle exec rake railties:install:migrations
```
tvdeyen: Is this extra step really needed? I presume that the install generators already install these migrations? Or are there any additional migrations from other engines needed?

guides/getting_started.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+These tasks will setup your Rails app as a basic store, with an authentication system and a storefront
+and checkout. It will also provide some sample products and configurations. You will be asked to set an admin email/password combination. The default values are `admin@example.com` and `test123`, respectively.
+
+Now, run the rails server (`rails s`) and, in your browser, navigate to `localhost:3000`. Voila: You'll find an app running an online shop. To access the store admin, visit `localhost:3000/admin`.
```
tvdeyen: Voilá ;)

jhawthorn: "voilà"

guides/getting_started.md
mamhoff: Yes, this is true. The step is really only there for the `db:create` step as without it, the generators won't start (they include that step, too).

guides/getting_started.md
mamhoff: This is copied verbatim from the Readme. And you're right, that step can go, it doesn't do anything.

mamhoff:
...amended. :)

guides/getting_started.md
tvdeyen: Sure the `db:create` is mandatory. I meant the `db:migrate`, sorry or being unclear.

guides/getting_started.md
tvdeyen: Haha, my French teacher will kill me..

guides/getting_started.md
```diff
@@ -0,0 +1,112 @@
...
+This article will explain how to setup a new Rails application with Solidus installed and give you
+hints at how to start customizing your store. We assume familiarity with Ruby and Rails, but will
+link to further resources to help you understand what we are doing.
+
+## Installing Solidus
+
+This chapter will guide you through installing Solidus for the first time.
+
+
+### Installing Image Magick
```
tvdeyen: ImageMagick

guides/getting_started.md
```diff
@@ -0,0 +1,112 @@
...
+# Solidus and Solidus Authentication
+gem 'solidus', '~> 2.0'
+gem 'solidus_auth_devise'
+
+group :development, :test do
+  gem 'rspec-rails', '~> 3.5'
+end
+
+group :development do
+  gem 'spring'
```
tvdeyen: As `spring` is cause of lots of headaches, we should not promote its usage here, no?

Senjai: +1, spring is annoying

tvdeyen:
```sh
echo "export DISABLE_SPRING=1" >> .profile
```

guides/getting_started.md
```diff
@@ -0,0 +1,101 @@
...
+
+```
+$ gem install rails bundler
+```
+
+## Generating your host app
+
+It's important to understand that Solidus is a [Rails engine](http://guides.rubyonrails.org/engines.html), and that your host app simply requires that engine. The first step for installing Solidus is therefore generating a vanilla Rails application.
+
+Solidus will not work well with the `sqlite` database, and it runs best with the `postgresql` database.
```
jhawthorn: I worry that this will give the wrong impression, and sqlite might be fine for users who are just starting out and developing.

How about

> Rails default `sqlite` database isn't well suited to real Solidus production usage. Solidus will runs best with the `postgresql` (or `MySQL`) database.

PR #1599, opened 2016-11-21
adds spree token to users api request

bosskovic:
Without this, promotion rule for users is consistently failing in the admin when select2 tries to fetch the list of the users from the api; something similar could be needed at other places as well, such as around line #34

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/user_picker.js
+++ b/backend/app/assets/javascripts/spree/backend/user_picker.js
@@ -12,7 +12,8 @@ $.fn.userAutocomplete = function () {
       Spree.ajax({
         url: Spree.routes.users_api,
         data: {
-          ids: element.val()
+          ids: element.val(),
+          token: Spree.api_key
         },
         success: function(data) {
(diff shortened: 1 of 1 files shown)


jhawthorn:
Hi @bosskovic thanks for the PR.

As far as I know the user picker now works. There was a different issue fixed with it in #1349 and I added a spec in #1372 to ensure it is working. Do you have an example which fails without this PR?

Specifying a token here shouldn't be necessary because [`Spree.ajax` does that for us](https://github.com/solidusio/solidus/blob/cd5ca0167013d9bc03555ca170b02db19a5d6ffa/core/app/assets/javascripts/spree.js.coffee.erb#L47).

bosskovic:
@jhawthorn you are right, we are using an older version of solidus (1.3.1) and the user picker does seem to work in a fresh solidus app updated to the most recent version.
We rely on a number of forks of solidus gems and can not quickly update solidus in our app, but I expect that when we do, the problem will be gone there as well. I will close this request now.

PR #1603, opened 2016-11-22
Only credit back allowed amount

gmacdougall:
This fixes an issue with payment refunding. To reproduce:

1. Give a user $100 in store credit
2. Purchase an item with the user for less than $100.
3. Refund a portion of the payment through the admin
4. Attempt to cancel the order

This will result in an exception as it will attempt to give the user more
store credit than they could potentially have available.

This modifies the store credit payment cancel to give back the maximum allowed
rather than the full amount.

Additionally, this gets rid of some terrible logic which only works with currencies
that have two decimal places which multiplies and divides things by 100.

### DIFF
--- a/core/app/models/spree/payment.rb
+++ b/core/app/models/spree/payment.rb
@@ -141,6 +141,17 @@ module Spree
       amount - (offsets_total.abs + refunds.sum(:amount))
     end
 
+    # The total amount this payment can be credited as a Spree::Money object
+    #
+    # @return [Spree::Money] The amount of this payment minus offets and refunds
+    #   as a money object with the associated currency.
+    def credit_allowed_money
+      Spree::Money.new(
+        credit_allowed,
+        currency: currency
+      )
+    end
+
     # @return [Boolean] true when this payment can be credited
     def can_credit?
       credit_allowed > 0
--- a/core/app/models/spree/payment_method/store_credit.rb
+++ b/core/app/models/spree/payment_method/store_credit.rb
@@ -58,12 +58,16 @@ module Spree
         currency = gateway_options[:currency] || store_credit.currency
         originator = gateway_options[:originator]
 
-        store_credit.credit(
-          amount_in_cents / 100.0.to_d,
-          auth_code,
-          currency,
-          action_originator: originator
-        ) if amount_in_cents > 0
+        if amount_in_cents > 0
+          store_credit.credit(
+            ::Money.new(amount_in_cents, currency).to_d,
+            auth_code,
+            currency,
+            action_originator: originator
+          )
+        else
+          ActiveMerchant::Billing::Response.new(true, '')
+        end
       end
 
       handle_action(action, :credit, auth_code)
@@ -76,8 +80,12 @@ module Spree
       if store_credit_event.nil? || store_credit.nil?
         ActiveMerchant::Billing::Response.new(false, '', {}, {})
       elsif store_credit_event.capture_action?
-        amount_in_cents = (store_credit_event.amount * 100).round
-        credit(amount_in_cents, auth_code)
+        amount = if store_credit_event.originator_type == 'Spree::Payment'
+          store_credit_event.originator.credit_allowed_money.cents
+        else
+          (store_credit_event.amount * 100).round
+        end
+        credit(amount, auth_code, { originator: store_credit_event.originator })
       elsif store_credit_event.authorization_action?
         void(auth_code)
       else
--- a/core/app/models/spree/payment_method/store_credit.rb
+++ b/core/app/models/spree/payment_method/store_credit.rb
@@ -58,7 +58,12 @@ module Spree
         currency = gateway_options[:currency] || store_credit.currency
         originator = gateway_options[:originator]
 
-        store_credit.credit(amount_in_cents / 100.0.to_d, auth_code, currency, action_originator: originator)
+        store_credit.credit(
+          amount_in_cents / 100.0.to_d,
+          auth_code,
+          currency,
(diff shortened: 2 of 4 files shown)


core/spec/models/spree/payment_spec.rb
```diff
@@ -611,6 +611,53 @@
...
+        let(:currency) { 'JPY' }
+        it "it returns the total amount" do
+          expect(subject).to eq(100)
+        end
+      end
+    end
+
+    context "when there are offsets" do
+      before do
+        payment.offsets.create(
```
bbuchalter: Prefer `create!` to ensure errors are raised in test setup.

core/app/models/spree/payment_method/store_credit.rb
```diff
@@ -67,7 +67,7 @@ def credit(amount_in_cents, auth_code, gateway_options = {})
         originator = gateway_options[:originator]
 
         store_credit.credit(
-          amount_in_cents / 100.0.to_d,
+          ::Money.new(amount_in_cents, currency).to_d,
```
bbuchalter: Is the `.to_d` still required now that we're returning a `Fixnum` from `amount_in_cents`?

core/app/models/spree/payment_method/store_credit.rb
```diff
@@ -84,8 +84,10 @@ def cancel(auth_code)
       if store_credit_event.nil? || store_credit.nil?
         ActiveMerchant::Billing::Response.new(false, '', {}, {})
       elsif store_credit_event.capture_action?
-        amount_in_cents = (store_credit_event.amount * 100).round
-        credit(amount_in_cents, auth_code)
+        credit(
+          store_credit_event.originator.credit_allowed_in_cents,
```
bbuchalter: Can the `originator` here ever be something other than a payment? I assume so, and cause this will break.

core/spec/models/spree/payment_method/store_credit_spec.rb
```diff
@@ -267,15 +267,17 @@
         create(:store_credit_capture_event,
```
bbuchalter: The spacing here needs some help. Perhaps in a separate commit?

core/spec/models/spree/payment_method/store_credit_spec.rb
```diff
@@ -267,15 +267,17 @@
...
       }
+      let(:payment) { create(:payment, order: order, amount: 10) }
 
       it_behaves_like "a spree payment method"
 
       it "refunds the capture amount" do
         expect { subject }.to change{ store_credit.reload.amount_remaining }.
-                              from(original_amount - captured_amount).
-                              to(original_amount)
+          from(90).to(100)
```
bbuchalter: This is a great improvement, thank you!

core/app/models/spree/payment_method/store_credit.rb
```diff
@@ -66,7 +66,12 @@ def credit(amount_in_cents, auth_code, gateway_options = {})
         currency = gateway_options[:currency] || store_credit.currency
         originator = gateway_options[:originator]
 
-        store_credit.credit(amount_in_cents / 100.0.to_d, auth_code, currency, action_originator: originator)
+        store_credit.credit(
+          ::Money.new(amount_in_cents, currency).to_d,
+          auth_code,
+          currency,
+          action_originator: originator
+        ) if amount_in_cents > 0
```
bbuchalter: Can we add a regression test for this guard?

core/app/models/spree/payment_method/store_credit.rb
gmacdougall: This is creating a RubyMoney in order to reverse the amount_in_cents logic. Example:

```Ruby
[1] pry(main)> ::Money.new(1000, 'USD').to_d.to_s
=> "10.0"
[2] pry(main)> ::Money.new(1000, 'JPY').to_d.to_s
=> "1000.0"
```

core/spec/models/spree/payment_spec.rb
gmacdougall: Done.

core/app/models/spree/payment_method/store_credit.rb
bbuchalter: I just assumed since a polymorphic association is used, we should expect it possible that something other than a payment method be the originator. Maybe that should be changed to not be polymorphic?

core/app/models/spree/payment_method/store_credit.rb
gmacdougall: Added

gmacdougall:
Going to do some additional work on this. There's an issue with cancelling an order containing a fully refunded store credit payment.

gmacdougall:
I have corrected the error which occurred when cancelling an order which was previously fully refunded.

core/app/models/spree/payment.rb
```diff
@@ -144,6 +144,20 @@ def credit_allowed
...
     end
 
+    # The total amount this payment can be credited in cents.
+    #
+    # For currencies which do not support decimal places, it will return the
+    # full value.
+    #
+    # @return [Fixnum] The amount of this payment minus offets and refunds
+    #   in cents.
+    def credit_allowed_in_cents
```
cbrunsdon: Whats the rational here for returning the in_cents rather than the money object and letting the caller do with it as it wishes?

bbuchalter: @gmacdougall could you follow up here?

gmacdougall:
This was superseded by https://github.com/solidusio/solidus/pull/2111

PR #1605, opened 2016-11-22
Make frontend prices depend on `store.cart_tax_country_iso`

mamhoff:
Prior to this change, prices in the frontend portion of Solidus would
depend on the setting for selecting prices in the backend, which is not
store-specific, but global. However, we want prices to be different in the
frontend if we change store and there are different prices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Solidus 2.1.0 (master, unreleased)
 
+*   Make frontend prices depend on `store.cart_tax_country_iso`
+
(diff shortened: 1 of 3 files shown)


tvdeyen:
I think it's worth being mentioned in the changelog

mamhoff:
Added a short changelog entry (although I think it's really a bugfix).

tvdeyen:
Thanks. IMO bug fixes are also worth to be mentioned in a changelog. Someone might want to see if an update is worthwhile because the long awaited bug fix finally landed.

PR #1611, opened 2016-11-24
Replace ShippingMethod#display_on with ShippingMethod#available_to_users

jhawthorn:
Echoing the work done in #1540 to PaymentMethods

This replaces `ShippingMethod#display_on` with `ShippingMethod#available_to_users`.

Unlike with payment_methods, this doesn't add an `#available_to_admin` column, because the existing functionality did not provide a way to hide shipping methods from admins. Having `display_on="front_end"` was identical to having `display_on="both"` (or actually any string other than `"back_end"`).

### DIFF
--- a/backend/app/views/spree/admin/shipping_methods/_form.html.erb
+++ b/backend/app/views/spree/admin/shipping_methods/_form.html.erb
@@ -7,14 +7,6 @@
     <% end %>
   </div>
 
-  <div data-hook="admin_shipping_method_form_display_field" class="col-xs-5">
-    <%= f.field_container :display_on do %>
-      <%= f.label :display_on %><br />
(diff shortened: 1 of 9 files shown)


jhawthorn:
> There is this DeprecatedConstantProxy, although I doubt that anybody ever used this constant and it's safe to be used removed.

`DeprecatedConstantProxy` is really cool, but only works if there is a replacement. We can manually define `const_missing`, which we [have done elsewhere](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core.rb#L49).

What's best here is probably to use [`DeprecatedObjectProxy`](http://api.rubyonrails.org/classes/ActiveSupport/Deprecation/DeprecatedObjectProxy.html)

PR #1612, opened 2016-11-24
Use dismiss_confirm

gevann:
This PR uses capybara's `dismiss_confirm` in a feature spec, rather than `dismiss_prompt` as the modal is an alert not a prompt.


Senjai:
TIL :+1:

tvdeyen:
Thanks

PR #1613, opened 2016-11-24
Remove use of require_tree in admin javascripts

### LINKED ISSUE #1567: Overriding the Handlebars templates is impossible in production
I recently faced a bug on Solidus 1.3.1 related to the backend handlebars templates.
I needed for one of my clients to override the variants suggestions in the autocomplete forms.
I first implemented my override by writing a new template in:
/vendor/assets/javascript/spree/backend/templates/variants/autocomplete.hbs.erb

It worked fine in development but in production, the original template was still in place.
After some research I thought that this would be related to the assets pipeline compilation process and by reading the minified JS I discovered that both versions lived in the same file.

My version of the template is appended all the way at the end of the compiled JS file. Unfortunately, the code binding the template to the variant search form is way higher in the code and call the original template function.

Any ideas how this could be fixed ?
Thanks!

jhawthorn:
This came out of discussion with @tvdeyen 

When we use `require_tree` or a relative path to `require`, it makes it so that attempting to override these files in your own app won't work.

This PR replaces all `require_relative` calls with an individual hardcoded `require` per-file. Requires for templates were split into their own `index.js` file

Fixes #1567


backend/app/assets/javascripts/spree/backend.js
```diff
@@ -12,7 +12,62 @@
...
 //= require spree/backend/spree-select2
-//= require_tree ./backend/templates
-//= require_tree .
+//
+//= require spree/backend/templates
+//
+//= require spree/backend/address_states
+//= require spree/backend/adjustments
+//= require spree/backend/admin
+//= require spree/backend/backbone-overrides
```
tvdeyen: You already required this in L15

tvdeyen:
Using this fix in my demo app https://github.com/tvdeyen/solidus-asset-override-demo works 👍

tvdeyen:
This actually needs to be deployed somewhere in order to confirm its really working. My demo app is a proof though, but I want to be extra sure 😎

PR #1615, opened 2016-11-25
Cleanup - Specify Spree:: namespace in all Product calls

fylooi:
Some calls to Product are not namespaced with Spree::. This leads to a collision if the including application has an existing Product model defined.

### DIFF
--- a/core/app/models/spree/product/scopes.rb
+++ b/core/app/models/spree/product/scopes.rb
@@ -84,12 +84,12 @@ module Spree
     # note that it can test for properties with NULL values, but not for absent values
     add_search_scope :with_property_value do |property, value|
       joins(:properties)
-        .where("#{ProductProperty.table_name}.value = ?", value)
+        .where("#{Spree::ProductProperty.table_name}.value = ?", value)
         .where(property_conditions(property))
     end
 
     add_search_scope :with_option do |option|
-      option_types = OptionType.table_name
+      option_types = Spree::OptionType.table_name
       conditions = case option
                    when String     then { "#{option_types}.name" => option }
                    when OptionType then { "#{option_types}.id" => option.id }
@@ -100,10 +100,10 @@ module Spree
     end
 
     add_search_scope :with_option_value do |option, value|
-      option_values = OptionValue.table_name
+      option_values = Spree::OptionValue.table_name
       option_type_id = case option
-                       when String then OptionType.find_by(name: option) || option.to_i
-                       when OptionType then option.id
+                       when String then Spree::OptionType.find_by(name: option) || option.to_i
+                       when Spree::OptionType then option.id
         else option.to_i
       end
 
@@ -117,7 +117,7 @@ module Spree
     add_search_scope :with do |value|
       includes(variants_including_master: :option_values).
       includes(:product_properties).
-      where("#{OptionValue.table_name}.name = ? OR #{ProductProperty.table_name}.value = ?", value, value)
+      where("#{Spree::OptionValue.table_name}.name = ? OR #{Spree::ProductProperty.table_name}.value = ?", value, value)
     end
 
     # Finds all products that have a name containing the given words.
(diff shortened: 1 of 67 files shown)


flyfy1:
I don't quite understand.. if we have code like:

```ruby

class Product
  def m
    puts "Product::m"
  end
end


module Spree
  class Product
    def m
      puts "Spree::Product::m"
    end
  end

  class Caller
    def call_product
      Product.new.m
    end
  end
end

Spree::Caller.new.call_product

```

shouldn't the SpreeProduct to be called by the Caller by default?

fylooi:
@flyfy1 
Yes, that would be the case if the constants were all in one file.
In this case, it calls my preexisting Product model. Suspect Rails autoloading might have something to do with it. Specifying the namespace results in correct behaviour and aligns with the rest of the code.

cbrunsdon:
I'm with @flyfy1 on this one in that I'm not sure under what circumstances the wrong product will be called, could it be something like code you've copy-pasted into a decorator that is causing the issue? I tried a local repo and couldn't recreate the undesired behavior.

fylooi:
Hmm. Moved past that code and can't dig up the point of reproduction. I'll reopen this PR if I can get a handle on it.

flyfy1:
@fylooi When I was reading [Rails Docs upon module loading](http://guides.rubyonrails.org/autoloading_and_reloading_constants.html) today, at section 6.2, it mentioned sth below (cannot copy with style so I just include the screenshot):

image

I think it's something related to your issue (and which also means, seems to make ur pull request valid)

fylooi:
@flyfy1 
The problem is actually a combination of Rails autoloading and development code reloading. Have experienced a couple of other model name conflicts but still unable to reproduce this in a test app. Disabling Spring and setting `config.eager_load = true` in development appears to prevent these conflicts, with a slightly degraded dev experience.

Ensuring that all models in controllers are consistently invoked with the `Spree::` namespace should solve this. The current code is a mix of namespaced and non-namespaced model calls. 

If you guys are okay with this, I can do the changes and update this PR.

flyfy1:
@fylooi I haven't looked through the documentation yet.. nor do I sure if u should continue with the PR. probably need comments from the core team. Meanwhile, I think it's definitely helpful to create a sample App and surface the problem..

core/app/models/spree/product/scopes.rb
```diff
@@ -173,7 +173,7 @@ def self.property_conditions(property)
 
     # Can't use add_search_scope for this as it needs a default argument
     def self.available(available_on = nil, currency = nil)
-      Spree::Deprecation.warn("The second currency argument on Product.available has no effect, and is deprecated", caller) if currency
+      Spree::Deprecation.warn("The second currency argument on Spree::Product.available has no effect, and is deprecated", caller) if currency
       joins(master: :prices).where("#{Product.quoted_table_name}.available_on <= ?", available_on || Time.current)
```
tvdeyen: Here we have on `Product` missing its module namespace

fylooi:
Updated the PR to cover all model invocations in controllers and models.

mamhoff:
I'm also fine with this. Quoting the [Zen of Python](https://www.python.org/dev/peps/pep-0020/):

> Namespaces are one honking great idea -- let's do more of those!

;)

jhawthorn:
Thank you

fylooi:
Managed to reproduce one case:

1. Add a reference to `Country` in `before_action` in `ApplicationController`

fylooi:
Managed to reproduce one case. Documenting the steps for reference.

Assuming you have a `Country` model defined in your host application:
1. Add a reference to `Country` in an `around_action` in `ApplicationController`. `yield` after calling a method on `Country`
2. Visit `CustomerDetailsController#edit` (v1.4)

```
      def edit
          country_id = Country.default.id
          @order.build_bill_address(country_id: country_id) if @order.bill_address.nil?
          @order.build_ship_address(country_id: country_id) if @order.ship_address.nil?

          @order.bill_address.country_id = country_id if @order.bill_address.country.nil?
          @order.ship_address.country_id = country_id if @order.ship_address.country.nil?
        end
```

The `Country.default.id` call will call the host `Country` class, and will fail if the host class doesn't have `default` defined.

Issue #1608, opened 2016-11-23
Product#slug length validation prevents products with short names from being created


fschwahn:
I'm importing a product database, and I'm hitting a problem with the [slug length validation](https://github.com/solidusio/solidus/blob/78250a20ec57e86543b4d10302d14cc07850fd3b/core/app/models/spree/product.rb#L94). The product has the name `As`, which admittedly is not a great name, but unfortunately I have to work with the existing names.

The question is, why does this validation even exist? I see no reason to disallow short slugs. I traced this back to commit d4d955017102f31980cc3cf3c0ca7ad9f5b053bd, which is were the internal permalink solution was replaced with friendly_id. However, I also see no justification in the commit message why this validation is necessary. There is also no spec which would shed light on this.

I suggest that either
a) The length validation on slugs is removed altogether.
b) The length validation can be disabled by a configuration option.

I "solved" the issue by removing the validation like [described here](https://gist.github.com/brenes/4503386), but that's pretty horrible.

jhawthorn:
Good catch. Let's replace `length: { minimum: 3 }` with just `presence: true`, the same validation on `name`.

Would you be interested in making a PR @fschwahn? :)

tvdeyen:
Hmm, I guess the minimum of three is I18n related? AlchemyCMS has a minimum length of three validation because of possible locales in the route. 

But as Solidus has `/products/slug-of-product` as default product route I think it's ok to remove this, but I'm not 100% sure.

fschwahn:
good point @tvdeyen - I think this is not an issue because of the structure of the routes (as you mention), but I have only a very superficial knowledge of the solidus/solidus_i18n codebase.

tvdeyen:
The solidus_i18n codebase adds the non-default locale to the routes, but that's prepended to the product route, so

`/de/products/slug-of-product` would be the actual route. 

That said I think it's safe to remove the length validation of the product slug.

bbuchalter:
Can this issue be closed now that https://github.com/solidusio/solidus/pull/1616 is merged?

fschwahn:
Yes

PR #1616, opened 2016-11-25
Update product slug validation

fschwahn:
Replace the existing length validation with a presence validation, because the length validation could lead to errors with short product names (see #1608 for more information).

Should I add a model spec for creating products with short names?

One more thing I noticed is the uniqueness validation, I think it's unnecessary because that's essentially what friendly_id handles. At least the `allow_blank: true` is unnecessary because we validate the length / presence, so this can never happen.


jhawthorn:
Thanks @fschwahn

PR #1617, opened 2016-11-25
Modify getting started guide use Spree::User class

mamhoff:
Previously, the beginner following this guide would end up with
the `Spree::LegacyUser` instead of the `Spree::User` as her user
class configuration.


jhawthorn:
I skipped our normal "wait at least 24 hours to make sure everyone has a chance to review". I think our documentation should have a lower barrier to merge (and if there's an issue it's much easier to change back).

tvdeyen:
:+1: thought the same, but hestitated.

PR #1620, opened 2016-11-26
Checkout validation

oniram88:
This Concern give the ability to make validation of the
Order model with standard rails ActiveRecord::Validation
whithout stop saving that. 

I've see this solution with @mamhoff at the RubyDay on the workshop, I think this is a cleaner way to make validation before start the checkout process (better as override the checkout_allowed? method)


mamhoff:
I like the approach (like I said in the workshop). To more clearly state the purpose of this (which we talked about IRL): 

`Spree::Order` has a `checkout_allowed?` method which returns false if there's no line items present. In that case the state machine will refuse to advance to checkout, but only with a generic error message that doesn't really say much to the user. If one were to add more conditions (in the workshop, the example was that an order needs to be > 100 dollars total), the user would not get an appropriate error message. 

The solution proposed here adds a `checkout_errors` thingy filled with `ActiveModel::Error` objects that won't stop the order from being persisted but still add meaningful errors usable in the API or in the frontend. 

This PR I think needs a couple of things though: 

1. It can be simplified quite a bit I think. I wouldn't add a concern, but rather some PORO that does all of this. I would like the "no line items present" condition to also be implemented with this, and the `checkout_allowed?` method to query for checkout errors. 

The build is still failing, and we need tests for this behaviour. 

@oniram88 Would you be up for spending more time on this? Don't go right now though, as there will probably be a bit of discussion around architecture here. Nevertheless, thank you for coming up with this, I think it's a pretty neat feature.

oniram88:
That was more a proof of concept as a pr. What do you mean with PORO? It's better to make a class eval like in spree/order/*.rb??
Or it's better to expose a complete new class that encapsulate all the validation logic with delegation from-to order class?

I will make validation as soon as we have more stable solution.

mamhoff:
What I'm thinking of is refactoring `Spree::Order#checkout_allowed?` to something like the following:

```ruby
def checkout_allowed?
  Spree::Checkout.new(self).can_start?
end
```

So there would need to be a `Spree::Checkout` object. I would not make this persisted yet, just make it a PORO: 
```ruby
module Spree
  class Checkout
    def initialize(order)
      @order = order
    end
    def can_start?
      Spree::Config.checkout_blockers.any? { |blocker| blocker(@order).blocks_checkout? } 
    end
    def checkout_blocker_errors
      Spree::Config.checkout_blockers.map? { |blocker| blocker(@order).error }.compact 
    end
  end
end
```

So this in turn requires a setting for an array of Checkout blockers: 

```ruby
class Spree::AppConfiguration
  attr_writer :checkout_blockers
  def checkout_blockers
    @checkout_blockers ||= [Spree::Checkout::Blockers::LineItemsRequired]
  end
end
```

And, finally, a set of checkout blockers, structured like this: 

```ruby
class Spree::Checkout::Blockers::LineItemsRequired
  def initialize(order)
    @order = order
  end
  def blocks_checkout?
    @order.line_items.empty?
  end
  def error
    return unless blocks_checkout?
    ActiveModel::Error.new(@order, "Not enough items")
  end
end
```

All of these would have to tested and documented using [YARD docs](https://gist.github.com/chetan/1827484), implemented well, with proper spacing and stuff though. 

EDIT: I've run this by @clarke, who has a very different opinion on how to do this properly, with pretty good reasons (chief among these is that `Spree::Order#checkout_allowed?` is called from within the innards of the order state machine). 

I'll leave this solution here as a starting point for a discussion. Let's go slow on this.

mtomov:
Surprising as it might be the `#checkout_allowed?` method is not called from within the state machine, but only from the [`Frontend Checkout` controller](https://github.com/solidusio/solidus/blob/9250e11d033220c84e903ba8726136dd9486de23/frontend/app/controllers/spree/checkout_controller.rb#L109-L113)

In our case, we have been tying such validation checks directly to the state machine:

```ruby
    def ensure_logged_in_user
      unless user
        errors.add(:base, I18n.t(:logged_in_user_is_required, scope: "checkout")) && (return false)
      end
    end

    state_machine.before_transition to: [:address], do: :ensure_logged_in_user
```

which compared to the above approach has minor differences, but the proposed approach will certainly make it clearer for the inexperienced how to add validations.

oniram88:
as proposed by @mamhoff , I changed into a PORO version. It is not yet complete by tests, but first I wanted to be sure that this was the direction to take.
I picked up the pieces of @mamhoff  and composed by several small fixes.

I hope that at least the yard documentation is OK, it's my first pull-request for a project of this level, and I have yet to engage the correct methodology

mtomov:
Cool way of doing this with the OO Blockers! .. with the base class and each blocker in a separate other one ..

core/app/models/spree/app_configuration.rb
```diff
@@ -381,6 +381,16 @@ def tax_adjuster_class
       @tax_adjuster_class ||= Spree::Tax::OrderAdjuster
     end
 
+    # Allows providing your own classes for validate the checkout stard process
```
bbuchalter: _ stard_ looks like a typo?

also "for validate" might be better as "for validating"

oniram88: yes, it's a typo!! thanks

mamhoff:
I still think this would be a good path towards moving checkout validations into their own class. However, there has not been any activity on this PR since December 2016, so I'm closing it.

PR #1622, opened 2016-11-28
Optimize PromotionHandler::FreeShipping

jordan-brough:
This shouldn't have any behavior change.

Since https://github.com/solidusio/solidus/pull/1392 we've been doing this:

Loop through every active active free-shipping promotion and activate
it if:

a) it's already connected
or
b) it has apply_automatically=true.

With this update we instead do this:

1) Activate already-connected free-shipping promos
2) Activate any non-connected "apply automatically" free-shipping
promos.

This avoids ever loading free shipping promos that do not have apply_automatically=true, which a store might have a lot of.

### DIFF
--- a/core/app/models/spree/promotion_handler/free_shipping.rb
+++ b/core/app/models/spree/promotion_handler/free_shipping.rb
@@ -10,38 +10,49 @@ module Spree
       end
 
       def activate
-        promotions.each do |promotion|
-          if order_promotion = existing_order_promotion(promotion)
-            promotion.activate(
-              order: order,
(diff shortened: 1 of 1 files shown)


jhawthorn:
There's now a striking similarity between `PromotionHandler::FreeShipping` and `PromotionHandler::Cart`. I think we should look into removing `PromotionHandler::FreeShipping` in favour of just using `PromotionHandler::Cart`, or at least sharing some code between the two.

Definitely out of scope for this PR, though :+1:

PR #1623, opened 2016-11-28
Remove method missing helper

gevann:
This PR removes the method_missing override from core base helpers, which removes  the custom image helper functionality and accompanying specs.

### DIFF
--- a/core/app/helpers/spree/base_helper.rb
+++ b/core/app/helpers/spree/base_helper.rb
@@ -134,15 +134,6 @@ module Spree
        time.strftime("%l:%M %p")].join(" ")
     end
 
-    def method_missing(method_name, *args, &block)
-      if image_style = image_style_from_method_name(method_name)
-        define_image_method(image_style)
-        send(method_name, *args)
-      else
-        super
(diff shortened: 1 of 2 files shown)


core/app/helpers/spree/base_helper.rb
```diff
@@ -159,35 +150,10 @@ def plural_resource_name(resource_class)
...
 
-    # Returns style of image or nil
-    def image_style_from_method_name(method_name)
-      if method_name.to_s.match(/_image$/) && style = method_name.to_s.sub(/_image$/, '')
-        possible_styles = Spree::Image.attachment_definitions[:attachment][:styles]
-        style if style.in? possible_styles.with_indifferent_access
-      end
-    end
-
     def create_product_image_tag(image, product, options, style)
```
jhawthorn: We can remove this also

tvdeyen:
The day has finally come 👏🏻

mtomov:
Hah, indeed so Thomas .. those have been quite a trickery to grep. Thanks a lot!

PR #1625, opened 2016-11-29
Add missing index to spree_payments.number

jordan-brough:
We ensure uniqueness and we query by number inside of
[Payment#set_unique_identifier](https://github.com/solidusio/solidus/blob/v2.0.0/core/app/models/spree/payment.rb#L266-L271).  This should have an index.


jhawthorn:
Thanks Jordan

PR #1626, opened 2016-11-30
Create migration

gevann:
This PR splits the 'display_on' string column of shipping methods into two boolean columns in the same manner as #1540


jhawthorn:
I believe this is a duplicate of #1611

PR #1628, opened 2016-11-30
Add shipped state to shipped order factory

nhippenmeyer:
This adds the `shipped` shipment state to the shipped order factory (say 3 times fast)


jhawthorn:
I believe this was fixed a while back by #1772. Please re-file if it is not.

PR #1629, opened 2016-11-30
Sum order#item_count from in-memory LineItems

jhawthorn:
This should make the order update marginally faster.

### DIFF
--- a/core/app/models/spree/order_updater.rb
+++ b/core/app/models/spree/order_updater.rb
@@ -102,7 +102,7 @@ module Spree
     end
 
     def update_item_count
-      order.item_count = quantity
+      order.item_count = line_items.to_a.sum(&:quantity)
     end
 
     def update_item_total


cbrunsdon:
Personally I think we should always assume that if an order exists its line items are there as well. I consider them an integral part of the order.

jhawthorn:
@tvdeyen wrote:
> I always had the impression that summing in database is faster. Are the line items already in memory? And can we ensure that this is always the case?

It's much faster to sum in the database if the line_item association isn't already loaded. I changed this here because in the only place that `.quantity` is called, the OrderUpdater, we know that the line items are loaded (and are expected to be up to date).

I've changed the pull request to just inline the in-memory summation as part of the order updater, since that seems more correct to do w/r/t compatibility just in case someone somewhere is using `order#quantity`.

In a future PR, we should consider deprecating `order#quantity` in favour of `order#item_count`.

PR #1631, opened 2016-12-01
seed Spree::Store with valid cart_tax_country_iso value

alepore:
The current wrong value is something like `"#<Spree::Tax::TaxLocation:0x007fe12a2856a0>"` (because `Spree::Config.admin_vat_location` returns an object).
Is US a good default? :)

/cc @mamhoff 

p.s:
Will the default Store become something like "Solidus Demo Site" in the future? :)


mamhoff:
@alepore Thank you, this is good.

danmitchell-:
+1 for this. Caused my new store a big head ache :-/

PR #1635, opened 2016-12-02
Refactoring Of The Checkout Controller

dvassev:
This refactoring decouples the update method, removing state leakage, and simplification of the method overall.

### DIFF
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -25,11 +25,12 @@ module Spree
 
     # Updates the order and advances to the next state (when possible.)
     def update
-      if order_update
+      if update_order
 
         assign_temp_address
 
-        on_failure and return unless state_transition
+        redirect_on_failure and return unless transition_forward!
+
 
         if @order.completed?
           finalize_order
@@ -44,7 +45,7 @@ module Spree
 
     private
 
-    def order_update
+    def update_order
       OrderUpdateAttributes.new(@order, update_params, request_env: request.headers.env).apply
     end
 
@@ -52,12 +53,12 @@ module Spree
       @order.temporary_address = !params[:save_user_address]
     end
 
-    def on_failure
+    def redirect_on_failure
       flash[:error] = @order.errors.full_messages.join("\n")
       redirect_to(checkout_state_path(@order.state))
     end
 
-    def state_transition
+    def transition_forward!
       if @order.confirm?
         @order.complete
       else
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -29,7 +29,10 @@ module Spree
 
         assign_temp_address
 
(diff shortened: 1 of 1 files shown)


frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -25,33 +25,61 @@ class CheckoutController < Spree::StoreController
...
-        end
-        if !success
-          flash[:error] = @order.errors.full_messages.join("\n")
-          redirect_to(checkout_state_path(@order.state)) && return
-        end
+      if order_update
+
+        assign_temp_address
+
+        on_failure and return unless state_transition
```
mtomov: You might want to consider renaming `order_update`, `state_transition` and `on_failure` to their verb first equivalents to follow the schematics on the rest of the methods and to indicate that an action is happening. Say to `update_order`, `transition_forward!` and `redirect_on_failure`

dvassev:
@mtomov Thank you very much for this. I didn't notice that we you guys were using verb first and it makes a ton of sense.

frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -25,33 +25,62 @@ class CheckoutController < Spree::StoreController
...
+    def assign_temp_address
+      @order.temporary_address = !params[:save_user_address]
+    end
+
+    def redirect_on_failure
+      flash[:error] = @order.errors.full_messages.join("\n")
+      redirect_to(checkout_state_path(@order.state))
+    end
+
+    def transition_forward!
```
jhawthorn: I don't think this should be a "bang" methd. There's no non-bang version of this. We're calling `next` and `complete` on order, not `next!` and `complete!`.

I'd prefer it was just called `transition_forward` or similar.

frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -25,33 +25,62 @@ class CheckoutController < Spree::StoreController
...
-        end
-        if !success
-          flash[:error] = @order.errors.full_messages.join("\n")
-          redirect_to(checkout_state_path(@order.state)) && return
-        end
+      if update_order
+
+        assign_temp_address
+
+        redirect_on_failure and return unless transition_forward!
```
jhawthorn: I find this line tough to parse for a couple reasons.

I prefer using the conditional suffix only on functions that don't mutate state. It's kind of sneaky that `transition_forward!`, the main purpose of this whole action, is at the end of the line. It also makes the order these run in mildly confusing: `this_second and this_third unless this_first`.


I'd prefer that the `transition_forward!` was more obvious, and that the order would read top-to-bottom

``` ruby
unless transition_forward!
  redirect_on_failure
  return
end
```

dvassev: It is definitely more readable and more accessible even for people coming from other languages.

dvassev:
@jhawthorn Thank you for the suggestions.

Issue #1634, opened 2016-12-02
VAT prices are generated twice for each country for variants


fschwahn:
I have a tax rate which has `included_in_price` set to true. When I create a new product, prices get generated twice for all countries. I think the reason is that product calls `validate_master`, which runs validations, which triggers the `before_validation :build_vat_prices` callback. The validations are run again through rails associations, which trigger `before_validation :build_vat_prices` again. I just checked our production database, and this also happens for non-master variants, so there's probably something else going on).

That means `Spree::Variant::VatPriceGenerator` is run twice, and because nothing has been persisted to the database yet, `variant.prices.find_or_initialize_by` always builds a new record.

I'm not sure what's the right thing to do here:
- Make sure that the callback is only run once?
- Run the callback at a different time (eg. before_save)?
- Make search for existing prices in `Spree::Variant::VatPriceGenerator` work even if nothing has been saved yet (ie. put `variant.prices` in an array and use Array#select for finding existing prices)?
- ...?

fschwahn:
Some more information:

This only happens when building a new product model, the validations are then run twice because of association autosave. There is already a spec in place (https://github.com/solidusio/solidus/blob/b7ca586099d7dec60c37fb69801dc55c5b7e34bf/core/spec/models/spree/variant_spec.rb#L73) which checks that the correct number of prices are added which doesn't trigger this behaviour because it operates on already saved products / variants. I can submit a PR which modifies this spec to showcase this problem.

flyfy1:
Looking forward to the fix. This issue bugged me for quite a while also.

PR #1638, opened 2016-12-05
Fix for duplicate prices

fschwahn:
This is a failing spec, and a solution for `Variant#build_vat_prices` running several times for variants and subsequently creating duplicate prices (see #1634).

As mentioned in the issue I'm not sure the solution is right (eg. what happens if validations are run several times on purpose while changing tax categories?), but it was the easiest to implement.

### DIFF
--- a/core/app/models/spree/variant/vat_price_generator.rb
+++ b/core/app/models/spree/variant/vat_price_generator.rb
@@ -28,16 +28,19 @@ module Spree
           # Don't re-create the default price
           next if variant.default_price && variant.default_price.country_iso == country_iso
 
-          foreign_price = variant.prices.find_or_initialize_by(
-            country_iso: country_iso,
-            currency: variant.default_price.currency,
-          )
+          foreign_price = find_or_initialize_price_by(country_iso, variant.default_price.currency)
 
           foreign_price.amount = variant.default_price.net_amount * (1 + vat_for_country_iso(country_iso))
         end
       end
 
       private
+
+      def find_or_initialize_price_by(country_iso, currency)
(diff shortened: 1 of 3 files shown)


mamhoff:
I really appreciate your work on this! But the solution you're proposing doesn't look right. You want to go two routes, possibly: 

1. Make sure the callback is called only once somehow
2. Check in the VAT price generator whether the price we're looking for is already in the unsaved prices array on the variant in question (replace `find_or_create_by` with a `select` statement). 

Also, you changed the spec to actually create eight prices, when we would expect only four to be created.

fschwahn:
@mamhoff thanks for your feedback,

I think the second route you outlined is probably the solution which is more correct, I'll try to put something together.

As for the spec, this error only happens when the product itself has not been saved yet (ie. only when the variant is saved along with the product). The 8 prices come from 4 master prices, 4 new_variant prices which are created when the product & variant are created. I'll try to make the spec clearer.

mamhoff:
This looks like a cleaner approach to me, thank you very much for this contribution. One more thing: Can you squash the commits (I think one commit would be sufficient) and provide a [good commit message](http://chris.beams.io/posts/git-commit/) for it?

fschwahn:
Squashed & provided a commit message with some context.

mamhoff:
@fschwahn FYI we have a policy of not merging approved PRs for at least 24 hours. While this has a good chance of being merged, it might take a couple of days. This is to allow community feedback.

fschwahn:
Makes sense, thanks for the heads up.

PR #1640, opened 2016-12-05
Fix error accessing spree_roles in rails 5.0.1.rc1

jhawthorn:
Accessing this through association fails because it tries to use the `Spree::LegacyUser::SpreeRole` class, which does not exist.

    NameError:
      uninitialized constant Spree::LegacyUser::SpreeRole

I'll be taking a look to see if I can fix this in Rails proper (or at least file a bug), but I'd like to get this patched in Solidus in case this isn't fixed by the Rails 5.0.1 release.


cbrunsdon:
I have nothing against explicitly listing classes for associations :+1:

PR #1644, opened 2016-12-09
Avoid break of Validator if DB column is missing

niklas:
If the gem `solidus_globalize` is loaded, the `Spree::ProductProperty#value` column cannot be found anymore - either because it was removed (as instructed by `globalize` gem) or because it is ignored and skipped by `ActiveRecord::Base.columns_hash`.

More on the discussion: https://github.com/solidusio-contrib/solidus_globalize/pull/19

Example failure: https://travis-ci.org/solidusio-contrib/solidus_globalize/jobs/182244854

### DIFF
--- a/core/app/models/spree/validations/db_maximum_length_validator.rb
+++ b/core/app/models/spree/validations/db_maximum_length_validator.rb
@@ -10,7 +10,13 @@ module Spree
       end
 
       def validate(record)
-        limit = record.class.columns_hash[@field].limit
+        field = record.class.columns_hash[@field]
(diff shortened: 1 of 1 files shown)


gmacdougall:
I'm not sure this is the solution that we want. I think if we have a validation on something, we should assume it exists. If solidus_globalize doesn't want the column to exist, maybe it should remove the validation.

jhawthorn:
I agree with Gregor. As far as I can tell, this doesn't fix `solidus_globalize`, but moves the error elsewhere.

PR #1645, opened 2016-12-09
Remove conversions to floating point

jhawthorn:
Floating point should never be used for money, or really any number you're uncomfortable with being described as "about".

A quick example why:
```
> 0.1 + 0.1 + 0.1
=> 0.30000000000000004
```

We use `BigDecimal` extensively, and have done pretty good job at ensuring all new code doesn't do float math. However there are still a few pieces of older code which convert to or back from floats. It's time to remove these.

Ruby 2.4 has [changed the default round mode](https://bugs.ruby-lang.org/issues/12548) for floating point, so we want to make especially sure we aren't rounding any floats.

## Places `to_f` is found in models
 * [x] refund.rb
 * [ ] calculator/tiered_percent.rb
 * [x] calculator/flexi_rate.rb
 * [ ] calculator/tiered_flat_rate.rb
 * [ ] calculator/shipping/flexi_rate.rb
 * [x] calculator/shipping/flat_percent_item_total.rb
 * [x] order.rb
 * [x] promotion/actions/create_quantity_adjustments.rb
 * [x] promotion/actions/create_item_adjustments.rb
 * [x] promotion/actions/create_adjustment.rb
 * [x] payment/processing.rb

### DIFF
--- a/core/spec/models/spree/order/payment_spec.rb
+++ b/core/spec/models/spree/order/payment_spec.rb
@@ -218,15 +218,27 @@ module Spree
       end
     end
 
-    context "payment required?" do
+    context "payment_required?" do
+      before { order.total = total }
+
       context "total is zero" do
-        before { allow(order).to receive_messages(total: 0) }
-        it { expect(order.payment_required?).to be false }
+        let(:total) { 0 }
+        it { expect(order).not_to be_payment_required }
       end
 
-      context "total > zero" do
-        before { allow(order).to receive_messages(total: 1) }
-        it { expect(order.payment_required?).to be true }
+      context "total is once cent" do
+        let(:total) { 1 }
+        it { expect(order).to be_payment_required }
+      end
+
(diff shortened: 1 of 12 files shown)


core/spec/models/spree/order/payment_spec.rb
```diff
@@ -218,15 +218,27 @@ module Spree
...
+
+      context "total is once dollar" do
+        let(:total) { 1 }
+        it { expect(order).to be_payment_required }
       end
 
-      context "total > zero" do
-        before { allow(order).to receive_messages(total: 1) }
-        it { expect(order.payment_required?).to be true }
+      context "total is hige" do
```
tvdeyen: huge? ;)

jhawthorn:
I'm going to tackle the last 3 files in separate PRs, since they will probably involve some (minor) behaviour changes.

PR #1647, opened 2016-12-13
Update en.yml

Arpsara:
- Correct syntax in devise/oauth_callbacks/failure
- Reorder action verbs in spree / actions and remove duplicated "ship" verb


Arpsara:
Updated specs with success! :)

jhawthorn:
Thank you

PR #1649, opened 2016-12-14
Check "create" ability for the New Order button

timrossinfo:
The New Order button on the List Orders page should check for the "create" ability, not the "edit" ability.


jhawthorn:
Thanks :+1:

PR #1658, opened 2016-12-19
Correctly record last_ip_address on order

bbuchalter:
The assignment of last_ip_address was not persisted where it was set.
When this module is included in a custom storefront, the assignment of
last_ip_address is lost. There was also a case in simple_current_order
where the last_ip_address was never assigned.

### DIFF
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -37,7 +37,7 @@ module Spree
           @current_order = find_order_by_token_or_user(options, true)
 
           if options[:create_order_if_necessary] && (@current_order.nil? || @current_order.completed?)
-            @current_order = Spree::Order.new(current_order_params)
+            @current_order = Spree::Order.new(new_order_params)
             @current_order.user ||= try_spree_current_user
             # See issue https://github.com/spree/spree/issues/3346 for reasons why this line is here
             @current_order.created_by ||= try_spree_current_user
@@ -45,7 +45,7 @@ module Spree
           end
 
           if @current_order
-            @current_order.last_ip_address = ip_address
+            @current_order.update(last_ip_address: ip_address)
             return @current_order
           end
         end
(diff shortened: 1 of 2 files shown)


bbuchalter:
@cbrunsdon I added a spec that checks for persistence when `simple_current_order` is called a second time to show that it's not persisted. Did I misunderstand you or write the spec incorrectly?

cbrunsdon:
@bbuchalter doesn't subject memoize as well? I don't believe your spec addition actually runs `simple_current_order` twice.

cbrunsdon:
@bbuchalter hey, my apologies, the return in simple_current_order means if its memoized its only going to be executed once, so right you are.

cbrunsdon:
I have objections to the existence of `simple_current_order`, but are not appropriate for this PR, so thanks @bbuchalter and double thanks for helping me through that misunderstanding

bbuchalter:
@cbrunsdon would you like me to rebase the fixup or remove it?

core/lib/spree/core/controller_helpers/order.rb
```diff
@@ -21,10 +21,10 @@ def simple_current_order
           @simple_current_order = find_order_by_token_or_user
 
           if @simple_current_order
-            @simple_current_order.last_ip_address = ip_address
+            @simple_current_order.update(last_ip_address: ip_address)
```
jhawthorn: One though before merging: should this be `update!`? We don't check the response so it probably should.

bbuchalter: I agree this should be an `update!`. Also before merge, there is this question about my fixup commit in this PR: https://github.com/solidusio/solidus/pull/1658#issuecomment-269883417

jhawthorn: The test in the fixup looks good to me (it can't hurt to have). I'd like it rebased please :)

bbuchalter:
Done.

bbuchalter:
Test fail does not seem legit. I can't seem to trigger a rebuild. Thoughts?

gmacdougall:
I triggered a rebuild. Here's a link to the odd failure: https://circleci.com/gh/solidusio/solidus/4985

I was not able to reproduce it locally.

bbuchalter:
Poke.

bbuchalter:
I have rebased this PR on master. Given that `simple_current_order` is now deprecated, I've removed any of my changes related to that method.

tvdeyen:
Thanks!

PR #1659, opened 2016-12-20
Fix the 'Send Mailer' checkbox selection

aamyot:
The user selection for suppressing the shipping confirmation email is not respected when shipping a package from the admin.  It always sends it.

### DIFF
--- a/backend/spec/features/admin/orders/shipments_spec.rb
+++ b/backend/spec/features/admin/orders/shipments_spec.rb
@@ -37,6 +37,15 @@ describe "Shipments", type: :feature do
       expect(page).to have_content("shipped package")
       expect(order.reload.shipment_state).to eq("shipped")
     end
+
+    it "doesn't send a shipping confirmation email when ask to suppress the mailer" do
+      find("#send_mailer").set(false)
+      
+      expect {
+        find(".ship-shipment-button").click
+        wait_for_ajax
+      }.not_to change{ ActionMailer::Base.deliveries.count }
+    end
   end
 
   context "moving variants between shipments", js: true do
--- a/backend/app/assets/javascripts/spree/backend/shipments.js
+++ b/backend/app/assets/javascripts/spree/backend/shipments.js
@@ -47,7 +47,7 @@ var ShipShipmentView = Backbone.View.extend({
       type: "PUT",
       url: Spree.routes.shipments_api + "/" + this.shipment_number + "/ship",
       data: {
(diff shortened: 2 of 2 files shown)


aamyot:
also  https://github.com/solidusio/solidus/pull/1659

aamyot:
lol... of course!  this wasn't meant to be posted here.  Sorry about the confusion.
Thanks.

backend/spec/features/admin/orders/shipments_spec.rb
```diff
@@ -37,6 +37,15 @@
...
       expect(order.reload.shipment_state).to eq("shipped")
     end
+
+    it "doesn't send a shipping confirmation email when ask to suppress the mailer" do
+      find("#send_mailer").set(false)
+      
+      expect {
+        find(".ship-shipment-button").click
+        wait_for_ajax
+      }.not_to change{ ActionMailer::Base.deliveries.count }
```
gmacdougall: I'm not a big fan of the test.

It's testing with `.not_to` without a corresponding test to check when it does change. For all we know, it could not deliver anything, or this test isn't configured correctly and we wouldn't know it.

backend/spec/features/admin/orders/shipments_spec.rb
```diff
@@ -37,6 +37,15 @@
       expect(page).to have_content("shipped package")
       expect(order.reload.shipment_state).to eq("shipped")
     end
+
+    it "doesn't send a shipping confirmation email when ask to suppress the mailer" do
+      find("#send_mailer").set(false)
+      
+      expect {
+        find(".ship-shipment-button").click
+        wait_for_ajax
```
gmacdougall: I would like to see the `wait_for_ajax` eliminated. We should expect something on the page to happen.

backend/spec/features/admin/orders/shipments_spec.rb
jhawthorn: We can wrap the test above this one (`it "can ship a completed order"`) with the opposite email expectation. You'll probably want to rebase first, since that spec was changed in #1668

```
it "can ship a completed order" do
  expect {
    find(".ship-shipment-button").click

    expect(page).to have_content("shipped package")
    expect(order.reload.shipment_state).to eq("shipped")
  }.to change{ ActionMailer::Base.deliveries.count }.by(1)
end
```

jhawthorn:
Rebased and merged as #1716. Thanks @aamyot

PR #1660, opened 2016-12-20
Switch order search to using starts over cont

gmacdougall:
Due to the way that standard indexes work in modern databases. Use of
contains to find a string is very inefficient. Instead, if we find
things that start with the string, the index can find the records it is
looking for very efficiently.

This turns a table sequential scan into a simple index lookup
signifncantly speeding things up, or avoiding timeouts on stores with
millions of orders.

Example:

```
mysql> SELECT COUNT(`spree_orders`.`id`) FROM `spree_orders` WHERE (`spree_orders` .`email` LIKE '%foo@bar\\.com%' AND `spree_orders`.`completed_at` IS NULL);
+----------------------------+
| COUNT(`spree_orders`.`id`) |
+----------------------------+
|                          1 |
+----------------------------+
1 row in set (20.27 sec)

mysql> SELECT COUNT(`spree_orders`.`id`) FROM `spree_orders` WHERE (`spree_orders` .`email` LIKE 'foo@bar\\.com%' AND `spree_orders`.`completed_at` IS NULL);
+----------------------------+
| COUNT(`spree_orders`.`id`) |
+----------------------------+
|                          1 |
+----------------------------+
1 row in set (0.00 sec)
```

### DIFF
--- a/backend/app/views/spree/admin/orders/index.html.erb
+++ b/backend/app/views/spree/admin/orders/index.html.erb
@@ -38,12 +38,12 @@
 
           <div class="field">
             <%= label_tag nil, Spree.t(:promotion) %>
-            <%= f.text_field :order_promotions_promotion_code_value_cont, size: 25 %>
+            <%= f.text_field :order_promotions_promotion_code_value_start, size: 25 %>
           </div>
 
           <div class="field">
             <%= label_tag nil, Spree.t(:shipment_number) %>
-            <%= f.text_field :shipments_number_cont %>
+            <%= f.text_field :shipments_number_start %>
           </div>
 
         </div>
@@ -52,13 +52,13 @@
           <div class="row">
             <div class="col-12 col-xl-6">
               <div class="field">
-                <%= label_tag :q_number_cont, Spree.t(:order_number, number: '') %>
-                <%= f.text_field :number_cont %>
+                <%= label_tag :q_number_start, Spree.t(:order_number, number: '') %>
+                <%= f.text_field :number_start %>
               </div>
 
               <div class="field">
-                <%= label_tag :q_email_cont, Spree.t(:email) %>
-                <%= f.text_field :email_cont %>
+                <%= label_tag :q_email_start, Spree.t(:email) %>
+                <%= f.text_field :email_start %>
               </div>
             </div>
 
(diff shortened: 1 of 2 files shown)


gmacdougall:
Sure thing. I agree that having contains would be better than starts with on e-mail, but the reason that this came up is that we're getting timeouts on our site with 10+ million orders when attempting to search via e-mail.

gmacdougall:
Changelog added in an amendment.

stewart:
I believe the contains is also useful for promotion codes.

cbrunsdon:
I think this is probably worth reaching out to a few places CS departments and do a quick ask to see if they ever use email addresses with "contains" like that. I know people could override the view and switch the query param back, but its potentially a disruptive thing. It would be nice if we could support wildcards ourselves and let people decide when to use them.

jhawthorn:
@cbrunsdon @gmacdougall Can anything be done to move this forward? One of our promises with Solidus is to work well on large stores, which this change is necessary for.

Maybe we should start by changing the less controversial searches, like `number` and `shipment_number` to `_start`

gmacdougall:
I feel that all of the things listed here are large enough that non-indexed searches are inappropriate. It's not that hard to get millions of orders or promotion codes.

bbuchalter:
@gmacdougall can we get this one rebased on master and review this work again?

gmacdougall:
Change rebased and shall rise again!

backend/app/views/spree/admin/orders/index.html.erb
```diff
@@ -52,13 +52,21 @@
...
+<<<<<<< HEAD
                 <%= label_tag :q_number_cont, Spree.t(:order_number, number: '') %>
                 <%= f.text_field :number_cont %>
+||||||| parent of e13e004... Switch order search to using starts over cont
+                <%= label_tag :q_number_cont, Spree.t(:order_number, :number => '') %>
+                <%= f.text_field :number_cont %>
+=======
+                <%= label_tag :q_number_start, Spree.t(:order_number, :number => '') %>
+                <%= f.text_field :number_start %>
+>>>>>>> e13e004... Switch order search to using starts over cont
```
kennyadsl: @gmacdougall I think this has been introduced rebasing and fixing the CHANGELOG conflict.

gmacdougall: My bad. I'll fix that up.

gmacdougall:
This is now rebased (properly)!

mamhoff:
Resolved conflict with the current `master` branch so that the changelog entry is in the right place. I still think this is a worthwhile change.

tvdeyen:
@gmacdougall would you mind to rebase again so we can merge this?

gmacdougall:
Rebased the CHANGELOG out of this. Should be good to go.

mtomov:
Thank you!

PR #1661, opened 2016-12-21
Remove deprecated methods/behaviour for Solidus 2.1

jhawthorn:
Solidus 2.1 is the point which we are removing deprecated methods and behaviour.

All removed deprecations removed here had to exist at the release of Solidus `1.4`/`2.0`.

I didn't remove absolutely all deprecations that met that criteria here. After discussing with @mamhoff and others we decided to leave the deprecated default_tax functionality (and some related zone scopes) until Spree 3.0 to allow data migrations when upgrading from Solidus pre-`1.3` to post-`2.1`.


jhawthorn:
Merging this a few hours ahead of our normal 24-hour minimum. Sorry!

tvdeyen:
Absolutely ok. 👍

PR #1663, opened 2016-12-22
WIP Move instance methods from Spree::Order::Checkout to Spree::Order

mamhoff:
These methods are, although they are employed as state machine callbacks,
just simple instance methods on Spree::Order. This commit moves them there.

The advantage of this is reducing the complexity of Spree::Order::Checkout,
as well as the code being clearer at where these methods are being used.

This is work in progress, I'm cooperating with @tvdeyen here.


tvdeyen:
Merging in favor of #1664

PR #1664, opened 2016-12-23
Reduce metaprogramming in checkout.rb

jhawthorn:
This builds on top of @mamhoff's work in #1663, but is submitted as a separate PR to avoid complicating it.

The existing `Order::Checkout` module added all its methods to the order inside of a `class_eval` within the `included` method.

This PR adds an `Order::Checkout::ClassMethods` module to hold the methods defined on the class, and defined the instance methods directly on `Order::Checkout`.

To do this it was necessary to change the `class_attribute`s into plain module-level `attr_accessor`. Though there are workarounds to avoid needing this change, I don't think there's a reason to avoid doing this. `Order` is probably not being subclassed.


mamhoff:
Awesome, I'll close #1663 as this does all the things I planned to do. 👍

cbrunsdon:
Good by me too, nice step forward, thanks :+1:.

PR #1666, opened 2016-12-26
Only remove index if index exists

brchristian:
Our app didn’t have a pre-existing index for `spree_products.slug` (and I didn’t see an earlier migration for that), so it was throwing

```ruby
ArgumentError: Index name 'index_spree_products_on_slug' on table 'spree_products' does not exist
```

when we tried to run this migration. This change will make the migration work correctly whether or not the index already exists.

### DIFF
--- a/core/db/migrate/20141217215630_update_product_slug_index.rb
+++ b/core/db/migrate/20141217215630_update_product_slug_index.rb
@@ -2,7 +2,7 @@ class UpdateProductSlugIndex < ActiveRecord::Migration[4.2]
   include Spree::MigrationHelpers
 
   def change
-    safe_remove_index :spree_products, :slug
+    safe_remove_index :spree_products, :slug if index_exists?(:spree_products, :slug)
     safe_add_index :spree_products, :slug, unique: true
   end
 end


jhawthorn:
This should be what [`safe_remove_index`](https://github.com/solidusio/solidus/blob/master/core/lib/spree/migration_helpers.rb#L3-L5) already does

``` ruby
def safe_remove_index(table, column)
  remove_index(table, column) if index_exists?(table, column)
end
```

brchristian:
For some reason, the migration that installed for me was simply:

```ruby
# This migration comes from spree (originally 20141217215630)
class UpdateProductSlugIndex < ActiveRecord::Migration
  def change
    remove_index :spree_products, :slug if index_exists?(:spree_products, :slug)
    add_index :spree_products, :slug, unique: true
  end
end
```

We are using the rubygems `Solidus 1.2.2` release.

Is this a versioning thing? Perhaps the `1.2.2` release doesn’t have the `safe` methods? If that’s the case it might be worth a new minor version release - `1.2.3` or whatever.

brchristian:
@jhawthorn Looking back at the [1.2 branch](https://github.com/solidusio/solidus/commits/v1.2) it does seem like the `safe` methods were added after the `1.2.2` release, but there has not yet been a `1.2.3` release. Might be worthwhile if other people on the upgrade path may experience the same issue as we did.

gmacdougall:
Hi brchristian,

Thanks for your input on this subject. I think we're comfortable with leaving the existing migrations as they are. If we were to provide some level of backport for this, I think we'd want to backport `safe_remove_index`.

PR #1667, opened 2016-12-27
Deprecate Spree::Order#has_step?

mamhoff:
The Spree::Order::Checkout module defines a `has_checkout_step?` method
which does exactly the same, with a better name and a check for the
`step` argument being `nil`.

This came out of pairing with @tvdeyen.


kennyadsl:
👍

PR #1670, opened 2017-01-05
Add partial index on `spree_promotions` table

alexstoick:
Add a partial index on the `apply_automatically` to catch the very frequent used case of fetching the `sale_promotions`.

### DIFF
--- a/core/db/migrate/20170105164636_add_partial_index_apply_automatically_promotions.rb
+++ b/core/db/migrate/20170105164636_add_partial_index_apply_automatically_promotions.rb
@@ -1,0 +1,11 @@
+class AddPartialIndexApplyAutomaticallyPromotions < ActiveRecord::Migration[5.0]
+  def change
+    condition = "apply_automatically = #{connection.quoted_true}"
+    add_index(
+      :spree_promotions,
+      :apply_automatically,
+      where: condition,
+      name: "idx_apply_automatically_promotions",
+    )
+  end
+end


cbrunsdon:
Looks like a legit failure:

```

HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
: CREATE  INDEX  "idx_apply_automatically_promotions" ON "spree_promotions"  ("apply_automatically") WHERE apply_automatically = 1
```

jordan-brough:
Should we leave the existing `apply_automatically` index in Solidus? I imagine for most stores almost all promotions are `apply_automatically=false` so it might be better just to have this partial index?

core/db/migrate/20170105164636_add_partial_index_apply_automatically_promotions.rb
```diff
@@ -0,0 +1,15 @@
+class AddPartialIndexApplyAutomaticallyPromotions < ActiveRecord::Migration[5.0]
+  def change
+    condition = if connection.adapter_name =~ /sqlite/i
+                  "apply_automatically = 1"
```
jhawthorn: I believe SQLite actually uses `'t'` as its true value.

It's probably safer and simpler here to use

``` ruby
condition = "apply_automatically = #{connection.quoted_true}"
```

Otherwise :+1: 

It's worth noting that MySQL and its forks don't support partial indexes, but there should be no harm (and still considerable gain) in it adding this index without the condition.

alexstoick:
@jhawthorn Made the change requested 👍

gmacdougall:
Can you please explain why this is preferable and faster than the existing index? I don't see this being significantly faster.

alexstoick:
@gmacdougall An index is effective if the DB can partition itself into somewhat good sized chunks. Our use-case for this column would be that we have 1 or 2 values that are true and a bunch of them being false, which means that the DB would do a sequencial scan instead of using the index.

Here are the 2 queries with `EXPLAIN ANALYZE`:


WITH INDEX:
```
=> select id from spree_promotions where apply_automatically = true;
...
(18 rows)
Time: 1.345 ms
```

```
=> explain analyze select id from spree_promotions where apply_automatically = true;
QUERY PLAN                                                                       
-------------------------------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on spree_promotions  (cost=22.50..11609.82 rows=368878 width=4) (actual time=0.056..0.060 rows=18 loops=1)
   Filter: apply_automatically
   ->  Bitmap Index Scan on spree_promotions_apply_automatically_idx  (cost=0.00..4.05 rows=184439 width=0) (actual time=0.048..0.048 rows=18 loops=1)
         Index Cond: (apply_automatically = true)
 Total runtime: 0.079 ms
```

WITHOUT INDEX:
```
select id from spree_promotions where apply_automatically = true;
...
(18 rows)
Time: 101.613 ms
```

```
explain analyze select id from spree_promotions where apply_automatically = true;
QUERY PLAN                                                      
----------------------------------------------------------------------------------------------------------------------
 Seq Scan on spree_promotions  (cost=0.00..13247.27 rows=368878 width=4) (actual time=94.509..96.446 rows=18 loops=1)
   Filter: apply_automatically
   Rows Removed by Filter: 737738
 Total runtime: 96.464 ms
```


The difference from **1ms** to **100ms** is big!


Let me know if you have any other questions!

gmacdougall:
@alexstoick I don't believe that this is correct. Here's a basic example in my postgres.

```
gregor=# create table promotion_test (id SERIAL, apply_automatically BOOLEAN);
CREATE TABLE
gregor=# INSERT INTO promotion_test (apply_automatically) SELECT FALSE FROM generate_series(1,1000000);
INSERT 0 1000000
gregor=# UPDATE promotion_test SET apply_automatically = true where id IN (585215, 598755, 513940, 30532, 408250, 83469, 504287, 408492, 509115, 255546, 918519, 769406, 178516, 148653, 414456, 551209, 55718, 144677);
UPDATE 18
gregor=# create index on promotion_test(apply_automatically);
CREATE INDEX
gregor=# EXPLAIN ANALYSE select * from promotion_test where apply_automatically = TRUE;
                                                                       QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using promotion_test_apply_automatically_idx on promotion_test  (cost=0.42..9.60 rows=67 width=5) (actual time=0.035..0.044 rows=18 loops=1)
   Index Cond: (apply_automatically = true)
   Filter: apply_automatically
 Total runtime: 0.080 ms
(4 rows)
```

Can you try a `VACUUM ANALYZE spree_promotions` on your dataset?

jordan-brough:
@gmacdougall I'm interested to see the details and the `VACUUM ANALYZE` from @alexstoick, but whatever the case there isn't it much more efficient overall anyway to use the partial index (on postgres & mysql) given this particular data set?

gmacdougall:
@jordan-brough I don't believe so. If you look at my above example it find 18 rows out of 1 million in 0.080 ms

jordan-brough:
@gmacdougall I'm talking about in terms of indexing overhead and space consumed by the index.

jordan-brough:
Nevermind actually, for a boolean column the space consumed is going to be too small to worry about probably.

mamhoff:
@jhawthorn @jordan-brough I'll leave this PR to Team America.

gmacdougall:
I'm fairly confident my example shows that this is not needed. I'm open to being wrong though and if we can find a good counter-example, I'm open to revisiting this. Closing for now.

PR #1673, opened 2017-01-05
Fix spurious image upload test failure

jhawthorn:
It's possible for the image to be uploaded before these assertions were run. We can't reliably test the progress of the upload, but it should be sufficient to test the final result of the upload.

The failure can be seen in [this TravisCI build](https://travis-ci.org/jhawthorn/solidus/jobs/189278481)


tvdeyen:
Violated the 24h blocking rule, because related to tests.

PR #1675, opened 2017-01-06
Checked currency of order and promotion or shipping currency same

vishalzambre:
Order currency and shipping or promotion currency for flat rate must be checked.

### DIFF
--- a/core/app/models/spree/calculator/flat_rate.rb
+++ b/core/app/models/spree/calculator/flat_rate.rb
@@ -9,7 +9,7 @@ module Spree
       if object && preferred_currency.casecmp(object.currency).zero?
         preferred_amount
       else
-        0
+        BigDecimal.new(0)
       end
     end
   end
--- a/core/app/models/spree/calculator/flexi_rate.rb
+++ b/core/app/models/spree/calculator/flexi_rate.rb
@@ -12,13 +12,17 @@ module Spree
     end
 
     def compute(object)
-      items_count = object.quantity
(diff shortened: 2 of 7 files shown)


tvdeyen:
Thanks for contributing. Could you please provide some more background for this? That would help reviewing. Thanks

vishalzambre:
@tvdeyen  Hey,

In any calculators if it is flat rate discount or shipping it should check order currency because lets take an example :

Web portal is working in  multiple currency
If order currency is INR
and web portal default currency is USD and all shipping and promotions are created in USD.
Now in case of shipping if order in US requires 10$ of shipping and in INR (India) It may requires 100$ to ship. Same with promotion.

mamhoff:
Hey @vishalzambre, will you look into this again? I understand your concern: In a multi-currency store, calculators need to be aware of the currencies. However, while you did add some validations, I don't see how this plays out in the backend (how do you administer multi-currency calculators), and I'm also not sure `BigDecimal.new("0")` is always the answer. Also, the test suite is failing. 

Maybe have a look into the `available?` method on calculators, and make sure calculators only return true for objects they can calculate? 

Also, since this PR has been quiet for some time, we'd close this issue if it turns out you don't have the time/energy to do this.

vishalzambre:
@mamhoff Hey sorry for late reply yes, I'll look into this.

vishalzambre:
I worked around conversions for the above solution, I think not required to change in core.

PR #1677, opened 2017-01-09
Add an issue_template.md for github

jhawthorn:
This adds an `issue_template.md`.

I don't want this to be a hindrance to anyone filing an issue. Especially since overall the quality of issues filed on this project has already been :ok_hand:.

However, it would be nice to have with the issue the exact Solidus version used, as well as which extensions are in use, to save some guesswork. Other headings should be in any good bug report, and were borrowed from [rails' issue_template.md](https://github.com/rails/rails/blob/master/.github/issue_template.md)

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,14 @@
+### Steps to reproduce
+How can we make this happen
+
+### Expected behavior
+Tell us what should happen
+
+### Actual behavior
+Tell us what happens instead
+
+### System configuration
+**Solidus Version**:
+
+**Extensions in use**:
+


jhawthorn:
> I'd like a small comment indicating that it's okay to submit issues that are not a bug report, though. Things like: I'd like to have Solidus support a better returns system, and I'd want people's input as to what that means. That would be a valid issue IMO, but wouldn't fit this template.

I'd like that too, but couldn't find a way to put it in the template non-awkwardly. As far as I can tell, there's no way to add a comment (like a code comment). Maybe I can add more details to CONTRIBUTING.md and link to that.

PR #1681, opened 2017-01-11
Add a frontend views override generator

tvdeyen:
We promote overriding views instead of using Deface. This is great, but manually copying the views from GitHub or your local gem folder is tedious and error prone.

By providing a generator we help our users to get the right views into their app.

### DIFF
--- a/frontend/spec/generators/solidus/views/override_generator_spec.rb
+++ b/frontend/spec/generators/solidus/views/override_generator_spec.rb
@@ -1,0 +1,47 @@
+require 'spec_helper'
+require_relative '../../../../lib/generators/solidus/views/override_generator'
+
+RSpec.describe Solidus::Views::OverrideGenerator, type: :generator do
+  destination Rails.root.join('app', 'views', 'spree')
+
+  before(:all) do
+    prepare_destination
+  end
+
+  subject! do
+    run_generator arguments
+  end
+
+  let(:src) do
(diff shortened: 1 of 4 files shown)


peterberkenbosch:
Nice!!!

frontend/spec/generators/solidus/views/override_generator_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        Dir.glob(dest.join("**", "*")).each do |file|
+          next if File.directory?(file)
+          expect(file.to_s).to match("products")
+        end
+      end
+    end
+  end
+
+  after do
+    FileUtils.rm_rf destination_root
```
gmacdougall: What is `destination_root` and how comfortable are we `rm -rf`ing it?

mamhoff: It's `Rails.root.join('app', 'views', 'spree')` - in a spec context the `spree` directory of the dummy app, which should be empty always. I feel pretty comfortable `rm-rf`ing :)

tvdeyen: Yep, `destination_root` is Rails' generator method that returns the path to the file destination. In this case the dummy app's `app/views/spree` paths that is empty anyways.

But good catch @gmacdougall !

Issue #1650, opened 2016-12-14
Spree::Money should be comparable


eoinkelly:
Is there a reason why `Spree::Money` does not implement `<=>` and include `Comparable`? The [current implementation](https://github.com/solidusio/solidus/blob/master/core/lib/spree/money.rb)  manually implements `#==` but not `<=>`. I think it would be convenient to be able to sort `Spree::Money` instances directly rather than having to reach into their wrapped `Money` instances.

I'm happy to have a go at a PR for this. Is there any constraint/reason I'm not seeing that would make it a bad idea?

mamhoff:
We'd like `Spree::Money` to implement all the things that the base `Money` object supports, one by one. If this particular itches you, by all means submit a PR!

kennyadsl:
@eoinkelly How do you want to implement this? I think we only should allow comparing Money objects that has the same currency with `<=>`.

jhawthorn:
@kennyadsl I agree. That seems to be how `::Money` works already, so delegating to that will have the desired behaviour.

```
> ::Money.new(10) < ::Money.new(20)
=> true
> ::Money.new(10) > ::Money.new(20)
=> false
> ::Money.new(10, "USD") < ::Money.new(20, "CAD")
ArgumentError: comparison of Money with Money failed
```

@eoinkelly A PR would be most welcome

eoinkelly:
Not comparing across currencies seems sensible. I'll have a go at a PR and then we'll have something concrete to discuss.

PR #1682, opened 2017-01-12
Add Comparable to Spree::Money

graygilmore:
Fixes https://github.com/solidusio/solidus/issues/1650

### DIFF
--- a/core/lib/spree/money.rb
+++ b/core/lib/spree/money.rb
@@ -6,6 +6,8 @@ module Spree
   # Spree::Money is a relatively thin wrapper around Monetize which handles
   # formatting via Spree::Config.
   class Money
+    include Comparable
+    DifferentCurrencyError = Class.new(StandardError)
     RUBY_NUMERIC_STRING = /\A-?\d+(\.\d+)?\z/
 
     class <<self
@@ -90,6 +92,22 @@ module Spree
     # (see #to_s)
     def as_json(*)
       to_s
+    end
+
+    def <=>(other)
+      if !other.respond_to?(:money)
+        raise TypeError, "Can't compare #{other.class} to Spree::Money"
+      end
+      if self.currency != other.currency
+        # By default, ::Money will try to run a conversion on `other.money` and
+        # try a comparison on that. We do not want any currency conversion to
+        # take place so we'll catch this here and raise an error.
+        raise(
+          DifferentCurrencyError,
+          "Can't compare #{self.currency} with #{other.currency}"
+        )
+      end
+      @money <=> other.money
     end
 
     # Delegates comparison to the internal ruby money instance.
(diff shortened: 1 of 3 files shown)


core/spec/lib/spree/money_spec.rb
```diff
@@ -268,4 +268,32 @@
       end
     end
   end
+
+  describe "<=>" do
```
graygilmore: I thought about also adding tests for `<`, `>`, `.between?`, and `.sort` as those also get included with `Comparable` but I didn't want to get into the habit of testing another module. Any thoughts?

core/lib/spree/money.rb
```diff
@@ -92,6 +94,22 @@ def as_json(*)
...
+      if self.currency != other.currency
+        # By default, ::Money will try to run a conversion on `other.money` and
+        # try a comparison on that. We do not want any currency conversion to
+        # take place so we'll catch this here and raise an error.
+        raise(
+          DifferentCurrencyError,
+          "Can't compare #{self.currency} with #{other.currency}"
+        )
+      end
+      self.money <=> other.money
```
kennyadsl: I'm not sure which one is better here but in the rest of the file to read attributes we used `@money` instead of `self.money`. Is there any reason to change this here? (I think it also applies to `self.currency -> @currency`). If they are equivalent we should probably be more consistent.

core/spec/lib/spree/money_spec.rb
kennyadsl: What about testing that `Comparable` is included? This way, if we remove that module, we'll be sure that some test will fail.

tvdeyen: Yes to testing that `Comparable` is included 👍 

No need to test a well tested module from Rubys standard lib

core/lib/spree/money.rb
tvdeyen: `@money` would work as this is a `attr_reader`, won't work for `@currency`, though, as this is a delegate to `money`

👍 for being consistent on `@money` and `currency` usage. 

🤔 thinking to enable Rubocops "redundant self" checks...

core/spec/lib/spree/money_spec.rb
graygilmore: Makes sense to me, I'll include that.

core/lib/spree/money.rb
graygilmore: Will fix.

graygilmore:
@kennyadsl @tvdeyen fixed the `@money` and added a test that checks if `Comparable` is included. I went back and forth a few times on the value of adding that test.

graygilmore:
@jhawthorn changelog entry added.

PR #1683, opened 2017-01-13
Use the order bill address for credit card address

cbrunsdon:
Previously, a new address would be created and associated with the
credit card. I believe it makes more sense for the factories to use the
bill address associated with the order, if it exists.

In addition to being more sound logic, it will create less addresses
during testing.


tvdeyen:
Who finds the verb can keep it

PR #1685, opened 2017-01-16
Revert "Deprecate Spree::Address#empty?"

jhawthorn:
Reverts solidusio/solidus#1680

This caused test failures which were for some reason not caught in the PR.


jhawthorn:
I figured it out. PRs didn't have the CircleCI config which caused them to fail on depredations. Fixed now in CircleCI config :roll_eyes:

cbrunsdon:
Now we have to live with those red :x:'s _forever_.

PR #1689, opened 2017-01-23
Avoid methods other than update! on OrderUpdater

jhawthorn:


jhawthorn:
> The thing I'm curious about is why methods like update_shipment_state and update_payment_state become public.

They were previously public. They're still a little too intertwined to be made private immediately.

For example, in `Order#finalize!` both are called with intermediate work being done in-between.

gmacdougall:
Thanks @jhawthorn 

I confused myself with the diffs. It was all the other stuff that became private, not those becoming public.

mtomov:
Indeed great!

PR #1694, opened 2017-01-31
Add failing spec for location configured package

jhawthorn:
The coordinator doesn't properly check stock levels, since stock can be double-counted towards location configured packages and normal packages.


jhawthorn:
This has passed specs... but CircleCI :roll_eyes:

PR #1697, opened 2017-01-31
Added LineItem name to unavailable flash

ericsaupe:
We were seeing issues where users wouldn't know which of their
products in their cart were out of stock and causing the issue.
Adding the product name to the flash message helps them out.

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1282,7 +1282,7 @@ en:
     inventory: Inventory
     inventory_adjustment: Inventory Adjustment
     inventory_canceled: Inventory canceled
-    inventory_error_flash_for_insufficient_quantity: An item in your cart has become unavailable.
+    inventory_error_flash_for_insufficient_quantity: "%{item} has become unavailable."
     inventory_not_available: Inventory not available for %{item}.
     inventory_state: Inventory State
     inventory_states:
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1282,7 +1282,7 @@ en:
     inventory: Inventory
     inventory_adjustment: Inventory Adjustment
     inventory_canceled: Inventory canceled
-    inventory_error_flash_for_insufficient_quantity: "%{item} has become unavailable."
+    inventory_error_flash_for_insufficient_quantity: "#{names} has become unavailable."
     inventory_not_available: Inventory not available for %{item}.
     inventory_state: Inventory State
     inventory_states:
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1282,7 +1282,7 @@ en:
     inventory: Inventory
     inventory_adjustment: Inventory Adjustment
     inventory_canceled: Inventory canceled
-    inventory_error_flash_for_insufficient_quantity: "#{names} has become unavailable."
+    inventory_error_flash_for_insufficient_quantity: "'#{names}' has become unavailable."
     inventory_not_available: Inventory not available for %{item}.
     inventory_state: Inventory State
     inventory_states:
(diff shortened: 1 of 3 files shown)


frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -150,7 +150,8 @@ def ensure_order_not_completed
 
     def ensure_sufficient_stock_lines
       if @order.insufficient_stock_lines.present?
-        flash[:error] = Spree.t(:inventory_error_flash_for_insufficient_quantity)
+        out_of_stock_items = @order.insufficient_stock_lines.map { |l| "\"#{l.name}\"" }.join(', ')
```
tvdeyen: No need to convert name into a string, it is already one. 

Also I would prefer to join the names with Rails' `to_sentence` instead of `join` with comma.

tvdeyen:
The test failures seem related. Could you please have a look? Thanks

frontend/app/controllers/spree/checkout_controller.rb
ericsaupe: Will do. I'm actually wrapping each name in `"` to distinguish it from the rest of the message.

ericsaupe:
It turns out the specs were originally written to look for the product name in the translation, https://github.com/solidusio/solidus/blob/master/frontend/spec/controllers/spree/checkout_controller_spec.rb#L414. I have updated the code here to adhere to those old specs.

frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -150,7 +150,8 @@ def ensure_order_not_completed
 
     def ensure_sufficient_stock_lines
       if @order.insufficient_stock_lines.present?
-        flash[:error] = Spree.t(:inventory_error_flash_for_insufficient_quantity)
+        out_of_stock_items = @order.insufficient_stock_lines.map { |l| "'#{l.name}'" }.to_sentence
```
tvdeyen: Please put the apostrophe into the locale file. Then users can decide if they want to have it in their notice or not.

tvdeyen: This could be rewritten as

```
@order.insufficient_stock_lines.collect(&:name).to_sentence
```

core/config/locales/en.yml
```diff
@@ -1282,7 +1282,7 @@ en:
     inventory: Inventory
     inventory_adjustment: Inventory Adjustment
     inventory_canceled: Inventory canceled
-    inventory_error_flash_for_insufficient_quantity: An item in your cart has become unavailable.
+    inventory_error_flash_for_insufficient_quantity: "'#{names}' has become unavailable."
```
jhawthorn: Seems like quoting the outside of the sentence will read quite odd if multiple items are out of stock.

> 'blue shirt and red shirt' has become unavailable.

ericsaupe: I was thinking about that today as well. Initially, I had it wrapping each product name but @tvdeyen suggested moving the apostrophe to the translation file. I've also been slightly bothered by the `has` vs `have` not being changed properly. Thoughts?

tvdeyen: Valid points. I think we could just remove the apostrophe from the translation. Stores that want to keep it, can easily update their locale file. And "has become" could be "became" and everything should be fine, no?

WDYT @jhawthorn @ericsaupe

ericsaupe: `became` is great! I just couldn't think of a way to rephrase that. Thanks!

ericsaupe:
@tvdeyen anything else needed on this?

jhawthorn:
Thank you @ericsaupe

PR #1705, opened 2017-02-06
Fix check for `order.guest_token` presence

vfonic:
* In case the `order.guest_token` is an empty string, we don't want
  everyone to be able to view or update the order by passing empty
  `token=` parameter

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -2,11 +2,11 @@ require 'spree/core/validators/email'
 require 'spree/order/checkout'
 
 module Spree
-  # The customers cart until completed, then acts as permenent record of the transaction.
+  # The customers cart until completed, then acts as permanent record of the transaction.
   #
   # `Spree::Order` is the heart of the Solidus system, as it acts as the customer's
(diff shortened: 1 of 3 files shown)


vfonic:
Not sure if `.present?` should be added or if that should be removed altogether. Also not sure if this particular piece of code requires a test.

jhawthorn:
@vfonic is there any way (with the standard Solidus code) to end up with the empty string (`""`) as a `guest_token`? In my testing there wasn't.

Using `#present?` here seems fine to me, just to be sure against something going horribly wrong.

Alternatively (or in addition), we could add the following to the `Order` model

```
validates :guest_token, presence: { allow_nil: true }
```

vfonic:
As far as I could see, there's no way to end up with empty `guest_token`. However, saying that, and thinking about all the possible ways you can change the data in the database, I'd say in some strange edge case, it's possible.

I'll see to make the validation change on Order model.

vfonic:
@jhawthorn I'm not sure why would `validates :guest_token, presence: true` fail?

PR #1707, opened 2017-02-07
Spree::Wallet & Non credit card payment sources

chrisradford:
This PR takes the work from #1091 and #1414 by @jordan-brough and @peterberkenbosch, rebases it over the latest `master` and updates a single failing spec (squashed into another commit).

From the original PR:

> This is an attempt to update Solidus to fully support non-credit-card payment sources, and to make that easier to do it also introduces the concept of Spree::Wallet (see the code comments in wallet.rb for a summary of that class).

What else is needed for this to be merged? Assuming this is still desirable.


core/app/models/spree/payment_create.rb
```diff
@@ -44,20 +51,33 @@ def build_source
...
+          payment.source.user = order.user
+        end
       end
     end
 
+    def build_from_wallet_payment_source
+      wallet_payment_source_id = source_attributes.fetch(:wallet_payment_source_id)
+      raise(ActiveRecord::RecordNotFound) if order.user.nil?
+      wallet_payment_source = order.user.wallet.find(wallet_payment_source_id)
+      raise(ActiveRecord::RecordNotFound) if wallet_payment_source.nil?
```
tvdeyen: Could we add a `find!` method to the `Wallet` class instead? Then raise in that class rather than here.

core/app/models/spree/wallet.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  # Find a WalletPaymentSource in the wallet by id.
+  #
+  # @param wallet_payment_source_id [Integer] The id of the WalletPaymentSource.
+  # @return [WalletPaymentSource]
+  def find(wallet_payment_source_id)
+    user.wallet_payment_sources.find_by(id: wallet_payment_source_id)
+  end
+
```
tvdeyen: Could we add a `find!` method so we make usage of `find_by!` from active record?

core/app/models/spree/wallet.rb
```diff
@@ -0,0 +1,76 @@
...
+  # Find a WalletPaymentSource in the wallet by the id of its PaymentSource.
+  #
+  # @param payment_source_id [Integer] The id of the PaymentSource
+  # @return [WalletPaymentSource]
+  def find_by_payment_source(payment_source_id)
+    user.wallet_payment_sources.find_by(payment_source_id: payment_source_id)
```
tvdeyen: You don't use this method anywhere. Can we remove it? 

Also the current implementation is unsafe. The same `payment_source_id` could be used by another `payment_source_type`.

core/app/models/spree/wallet_payment_source.rb
```diff
@@ -0,0 +1,17 @@
+class Spree::WalletPaymentSource < ActiveRecord::Base
+  belongs_to :user, class_name: Spree.user_class, foreign_key: 'user_id'
+  belongs_to :payment_source, polymorphic: true
+
+  validates_presence_of :user
+  validates_presence_of :payment_source
```
tvdeyen: In Rails 5 `belongs_to` associations are required by default. We don't need these validations.

core/db/migrate/20160420044191_create_spree_wallet_payment_sources.rb
```diff
@@ -0,0 +1,18 @@
+class CreateSpreeWalletPaymentSources < ActiveRecord::Migration[4.2]
+  def change
+    create_table :spree_wallet_payment_sources do |t|
+      t.references :user, index: true, null: false
```
tvdeyen: Can we use `foreign_key: true` here instead and remove `index` and `null`?

core/lib/spree/core/controller_helpers/payment_parameters.rb
```diff
@@ -116,6 +116,60 @@ def move_existing_card_into_payments_attributes(params)
       params
     end
 
+    # This method handles the awkwardness of how the html forms are currently
+    # set up for frontend.
```
tvdeyen: Can we move this into the `frontend` gem?

core/spec/models/spree/credit_card_spec.rb
```diff
@@ -35,38 +37,7 @@ def self.payment_states
...
-    end
-
-    it "should be false when credit_allowed is zero" do
-      payment = mock_model(Spree::Payment, completed?: true, credit_allowed: 0, order: mock_model(Spree::Order, payment_state: 'credit_owed'))
-      expect(credit_card.can_credit?(payment)).to be false
-    end
-  end
-
-  describe "#valid?" do
+  context "#valid?" do
```
tvdeyen: Can we leave this a `describe` block?

core/spec/models/spree/credit_card_spec.rb
```diff
@@ -283,33 +254,7 @@ def self.payment_states
...
-      credit_card.name = "Ludwig van Beethoven"
-    end
-
-    it "extracts the last name" do
-      expect(credit_card.last_name).to eq "van Beethoven"
-    end
-  end
-
-  describe "#to_active_merchant" do
+  context "#to_active_merchant" do
```
tvdeyen: Can we leave this a `describe` block?

core/spec/models/spree/credit_card_spec.rb
```diff
@@ -329,44 +274,49 @@ def self.payment_states
...
-  it 'allows this card to save even if the previously default card has expired' do
-    user = FactoryGirl.create(:user)
-    first = FactoryGirl.create(:credit_card, user: user, default: true)
-    second = FactoryGirl.create(:credit_card, user: user, default: false)
-    first.update_columns(year: DateTime.current.year, month: 1.month.ago.month)
+      second.update_attributes!(default: true)
+    end
 
-    second.update_attributes!(default: true)
+    it 'allows this card to save even if the previously default card has expired' do
```
tvdeyen: This test is identical to the previous one.

core/spec/models/spree/payment_create_spec.rb
```diff
@@ -81,6 +81,10 @@ module Spree
         }
       end
 
+      around do |example|
+        ActiveSupport::Deprecation.silence { example.run }
```
tvdeyen: We don't want to silence `ActiveSupport::Deprecation` but `Spree::Deprecation` here

core/spec/models/spree/user_spec.rb
```diff
@@ -92,7 +92,9 @@
       end
 
       it "has payment sources" do
-        expect(user.payment_sources.first.gateway_customer_profile_id).not_to be_empty
+        ActiveSupport::Deprecation.silence do
```
tvdeyen: We don't want to silence `ActiveSupport::Deprecation` but `Spree::Deprecation` here

core/spec/support/concerns/payment_source.rb
```diff
@@ -0,0 +1,64 @@
...
+    it { is_expected.to respond_to(:imported) }
+    it { is_expected.to respond_to(:name) }
+  end
+
+  context "#associations" do
+    it { is_expected.to respond_to(:payments) }
+    it { is_expected.to respond_to(:user) }
+  end
+
+  context "#can_capture?" do
```
tvdeyen: Can we use a `describe` block for methods?

core/app/models/concerns/spree/user_methods.rb
```diff
@@ -23,6 +23,8 @@ module UserMethods
       has_many :store_credit_events, through: :store_credits
       money_methods :total_available_store_credit
 
+      has_many :wallet_payment_sources, foreign_key: 'user_id', class_name: 'Spree::WalletPaymentSource'
+
```
jordan-brough: Personally I'm not a huge fan of that module, and most of it is going away.  How about the reverse - bring the `has_many` from that module into here? Then the whole module goes away soon.

core/app/models/spree/payment_create.rb
```diff
@@ -44,20 +51,33 @@ def build_source
...
-        payment.source.user_id = order.user_id if order
+        if order && payment.source.respond_to?(:user=)
+          payment.source.user = order.user
+        end
       end
     end
 
+    def build_from_wallet_payment_source
+      wallet_payment_source_id = source_attributes.fetch(:wallet_payment_source_id)
+      raise(ActiveRecord::RecordNotFound) if order.user.nil?
```
jordan-brough: The method will only be used when there _is_ a user for the order.  The next line will fail with a `NoMethodError: undefined method `wallet' for nil:NilClass` if there isn't a user.

core/app/models/concerns/spree/user_methods.rb
tvdeyen: Absolutely!

core/app/models/spree/payment_create.rb
tvdeyen: Ok, but why raise then?

core/app/models/spree/wallet.rb
```diff
@@ -0,0 +1,76 @@
...
+    @user = user
+  end
+
+  # Returns an array of the WalletPaymentSources in this wallet.
+  #
+  # @return [Array<WalletPaymentSource>]
+  def wallet_payment_sources
+    user.wallet_payment_sources.to_a
+  end
+
```
chrisradford: Add:
```ruby
def reusable_payment_sources
  wallet_payment_sources.collect(&:payment_source).select(&:reusable?)
end
```

core/app/models/spree/gateway.rb
```diff
@@ -42,16 +42,18 @@ def supports?(source)
...
+    alias_method :sources_by_order, :reuable_sources_by_order
+    deprecate :sources_by_order, deprecator: Spree::Deprecation
 
     def reusable_sources(order)
       if order.completed?
-        sources_by_order(order)
+        reuable_sources_by_order(order)
       elsif order.user_id
-        credit_cards.where(user_id: order.user_id).with_payment_profile
+        order.user.wallet.wallet_payment_sources.collect(&:payment_source).select {|source| source.reusable?}
```
chrisradford: Can remove reusable? since all payment sources in the wallet are reusable

core/app/models/spree/payment_create.rb
chrisradford: Move this to start of method with:

`return if user.order.nil?`

core/app/models/spree/payment_create.rb
chrisradford: Change to a `return if wallet_payment_source.nil?`

core/app/models/spree/wallet.rb
```diff
@@ -0,0 +1,76 @@
...
+  #
+  # @param payment_source_id [Integer] The id of the PaymentSource
+  # @return [WalletPaymentSource]
+  def find_by_payment_source(payment_source_id)
+    user.wallet_payment_sources.find_by(payment_source_id: payment_source_id)
+  end
+
+  # Find the default WalletPaymentSource for this wallet, if any.
+  # @return [WalletPaymentSource]
+  def default
```
chrisradford:
```ruby
def default_wallet_payment_source
```

core/db/migrate/20160420044191_create_spree_wallet_payment_sources.rb
```diff
@@ -1,7 +1,7 @@
 class CreateSpreeWalletPaymentSources < ActiveRecord::Migration[4.2]
   def change
     create_table :spree_wallet_payment_sources do |t|
-      t.references :user, index: true, null: false
+      t.references :user, foreign_key: true
```
tvdeyen: You need to pass `{to_table: Spree.user_class.table_name}` to `foreign_key`

chrisradford: Ah, thanks!

Issue #1696, opened 2017-01-31
[RFC] Remove (or circumvent) location configured packages


jhawthorn:
Location configured packages are a feature added in Solidus 1.0 which allows specifying an quantity and variant which must be sent from a specific location before shipments are created. This feature is only on the "cart" page of the admin (and available through the API). This is **not** the behaviour used on the shipments page.

![](http://i.hawth.ca/s/b4aXYOn1.png)

The cart page forces the admin to specify which stock location they want to use inventory from. When building packages, the coordinator will build shipments specifically with inventory from these locations (in shipments separate from the automatically allocated inventory).

These are responsible for many bugs:

* Creates split shipments where it's not desired (since location configured packages are always separate) [#1366](https://github.com/solidusio/solidus/issues/1366)
* Double-counts stock if both location configured and standard packages are built [#1694](https://github.com/solidusio/solidus/issues/1694)
* If you are adding to an existing line item, it doesn't work [#1695](https://github.com/solidusio/solidus/issues/1695)

The UI is also extremely lacking ot misleading. These issues could also all be considered bugs

* There is no indication in the admin that inventory has been reserved for a specific location.
* There is no way not to configure the location when adding an item on the admin Cart page
* In contrast to the above, there is no way to specify quantities when updating a line_items's amount

## Action

We have three options for what action to take:
* Make location configured packages optional. Require an additional mouseclick to use them from. This is probably the least controversial change.
* Extract location configured packages to a separate gem, removing them from Solidus itself (my preference).
* Leave it as is. Fix what issues can be fixed, others, like [#1366](https://github.com/solidusio/solidus/issues/1366) are a design flaw and can't be easily fixed.

## Some relevant bits of code

* [line_items_on_order_edit.js](https://github.com/solidusio/solidus/blob/33797b434986cf8fdecf6369467d29d6bd31979b/backend/app/assets/javascripts/spree/backend/line_items_on_order_edit.js)
* [OrderContents](https://github.com/solidusio/solidus/blob/33797b434986cf8fdecf6369467d29d6bd31979b/core/app/models/spree/order_contents.rb#L101-L103)
* [Stock::Coordinator](https://github.com/solidusio/solidus/blob/33797b434986cf8fdecf6369467d29d6bd31979b/core/app/models/spree/stock/coordinator.rb#L41-L52)

tvdeyen:
In the last core team meeting we decided to go with option 2 (removing the feature).

PR #1709, opened 2017-02-08
Remove stock location configuration from admin "cart" page

jhawthorn:
For the reasons discussed in #1696.

This removes the stock location in the simplest way possible. A follow up PR will have UI improvements to the page which this allows.

Before:

After:

### DIFF
--- a/backend/spec/features/admin/orders/new_order_spec.rb
+++ b/backend/spec/features/admin/orders/new_order_spec.rb
@@ -26,7 +26,7 @@ describe "New Order", type: :feature do
     click_on 'Cart'
     select2_search product.name, from: Spree.t(:name_or_sku)
 
-    fill_in_quantity("table.stock-levels", "quantity_0", 2)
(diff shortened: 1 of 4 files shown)


backend/spec/features/admin/orders/new_order_spec.rb
```diff
@@ -130,7 +130,7 @@
       click_on 'Cart'
       select2_search product.name, from: Spree.t(:name_or_sku)
 
-      fill_in_quantity('table.stock-levels', 'quantity_0', 1)
```
tvdeyen: I think we can also deprecate and remove the `fill_in_quantity` Capybara extension as well.

jhawthorn: Done in #1710

tvdeyen: 👍

PR #1710, opened 2017-02-09
Admin cart page revamp

jhawthorn:
This builds on top of #1709.

## New Order

* We display the "new item" row immediately
* "Add item" is disabled (we are already adding a row)
* The "cancel" button is removed

## Validation failed

* After clicking save
* blank fields are highlighted in red

## After saving one item

* "Add item" button is enabled

## Adding second item

* "Add item" button is disabled
* Cancel button is available (clicking it removes the row and re-enables the "Add item button")


core/lib/spree/testing_support/capybara_ext.rb
```diff
@@ -20,6 +20,12 @@ def column_text(num)
   end
 
   def fill_in_quantity(table_column, selector, quantity)
+    Spree::Deprecation.warn <<-WARN.strip_heredoc
+      fill_in_quantity is deprecated. Instead use:
+        within(#{table_column.inspect}) do
+          fill_in #{selector.inspect}, with: #{quantity.inspect}
+        end
+    WARN
```
tvdeyen: Ah, you already did :)

Mandily:
👍 Looks good to me!

vfonic:
I've never seen the UI in which "Add Item" is disabled. From my experience, the button/link is enabled and, when the user clicks on it, it adds a new empty row (or if we don't want that, we can focus on the above empty row). If the user didn't want to do this, s/he can remove that row.
To me, this UI looks like:
I fill in the cart item details and then "Add Item" button is enabled, so I click on it to confirm adding **this first** item. Of course this is wrong, since "Add Item" doesn't confirm my action, but the button on the right does.

PS I've noticed that action buttons/icons in solidus admin are usually outside of the table and table row. I think if the action button is referring to the item in the table row, it should be inside the table itself:
Screenshot
Not sure why it was explicitly made look like it doesn't belong to the table.

PPS I'm trying to bring a new "outside" view. Maybe I'm wrong in some things so please either correct me or ignore me if you want to ship things faster. ;)

Mandily:
I also like the idea of putting the icons in the table to keep relative things grouped together, but it might be better to keep it separate for consistency right now, and then globally change the location of all icons in a separate PR. I realize in suggesting that though, someone would have to volunteer to do that chunk of work. This sounds like a thing that I might be capable of, but I'm not sure when I'll have the time. 

I can see how on first glance a user might think that the "add item" button at the bottom of the page would save the new row. I know they're figure it out and learn pretty quickly, but in the interest of a more intuitive UI, it would be nice if we could rename the label to "Add new item"

The note about the disabled button raises a bigger discussion about whether we want to put the user into a stateful interaction (which is how Solidus works now) to keep them focused on their task or make it stateless to give them more flexibility. Again, I'm inclined to say keep it consistent until we have a bigger bandwidth to think through and change things on a more global scale.

jhawthorn:
> Not sure why it was explicitly made look like it doesn't belong to the table.

This is the pattern used throughout the admin and is way outside the scope of this PR.

If you have thoughts on how the admin UI should be please file a new issue or come discuss it in our `#gui` slack channel

vfonic:
@jhawthorn 

I was talking generally:
> PS I've noticed that action buttons/icons **in solidus admin** are usually outside of the table...

I'll start the discussion in `#gui`

backend/app/assets/javascripts/spree/backend/templates/orders/line_item.hbs
```diff
@@ -0,0 +1,42 @@
...
+{{#if editing}}
+	<a class="save-line-item fa fa-ok no-text with-tip" href="#" title="{{ t "actions.save" }}"></a>
+	{{#unless noCancel}}
+	<a class="cancel-line-item fa fa-cancel no-text with-tip" href="#" title="{{ t "actions.cancel" }}"></a>
+	{{/unless}}
+{{else}}
+	<a class="edit-line-item fa fa-edit no-text with-tip" href="#" title="{{ t "actions.edit" }}"></a>
+	<a class="delete-line-item fa fa-trash no-text with-tip" href="#" title="{{ t "actions.delete" }}"></a>
+{{/if}}
+</td>
```
tvdeyen: The indenting of this file is a little off

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
```diff
@@ -1,18 +1,3 @@
...
-  .delete-line-item,
-  .line-item-qty-show {
-    display: none !important;
-  }
-}
-
-.line-item:not(.editing) {
-  .save-line-item,
-  .cancel-line-item,
-  .line-item-qty-edit {
```
tvdeyen: Can we remove these classes from the html as well?

backend/app/assets/javascripts/spree/backend/line_items.js.coffee
```diff
@@ -1,55 +1,98 @@
...
+      isNew: @model.isNew(),
+      noCancel: @noCancel
+    )
+    el = @$el.html(html)
+    @$("[name=variant_id]").variantAutocomplete({ in_stock_only: true })
+
+$ ->
+  if $("table.line-items").length
+    url = Spree.routes.orders_api + "/" + order_number
+    lineItemModel = Backbone.Model.extend
```
tvdeyen: Can we put this model into its own file and use the Spree.Model namespace discussed in #1715?

backend/app/assets/javascripts/spree/backend/line_items.js.coffee
```diff
@@ -1,55 +1,98 @@
...
-  Spree.ajax(
-    type: "DELETE"
-    url: url
-  ).done (msg) ->
-    $('#line-item-' + line_item_id).remove()
-    if $('.line-items tr.line-item').length == 0
-      $('.line-items').remove()
-    window.Spree.advanceOrder()
+
+Spree.CartLineItemView = Backbone.View.extend
```
tvdeyen: Maybe we can also introduce a Spree.Views namespace for views as we discussed for models in #1715?

backend/app/views/spree/admin/orders/_line_items.html.erb
```diff
@@ -1,50 +1,23 @@
...
-              <%= link_to '', '#', class: 'delete-line-item fa fa-trash no-text with-tip',  title: Spree.t('actions.delete') %>
-            <% end %>
-          </td>
-        <% end %>
-      <% end %>
-    </tbody>
-  </table>
-<% end %>
+  <tbody>
+  </tbody>
```
tvdeyen: It really helped me today that in the other backbone powered templates we have a comment noting that the content is handled via JS. Maybe we could also add a comment here.

backend/app/assets/javascripts/spree/backend/line_items.js.coffee
jhawthorn: Can't be done here, needs to be defined where we have access to `url`. #1715 is a follow up to this where this will be fixed.

backend/app/assets/javascripts/spree/backend/line_items.js.coffee
jhawthorn: I'd rather address this in the #1715, which is a follow up to this. This PR was intended only to be the UI change, with #1715 being a more complete switch to backbone.

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
jhawthorn: `.save-line-item` and `.cancel-line-item` are still in use and also used in specs (a more dubious use).

I'm not sure they aren't appropriate classes to have.

backend/app/views/spree/admin/orders/_line_items.html.erb
jhawthorn: :+1: Excellent idea

backend/app/assets/javascripts/spree/backend/templates/orders/line_item.hbs
jhawthorn: Rebased and fixed.

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
tvdeyen: `.line-item-qty-edit` and `.line-item-qty-show` are JS utility classes used in the old JS only AFAIK. Maybe there are others. Not mandatory, for sure. But I like to clean up as we have the chance.

backend/app/assets/javascripts/spree/backend/line_items.js.coffee
tvdeyen: Sure.

backend/app/assets/javascripts/spree/backend/line_items.js.coffee
tvdeyen: Okidoki

backend/app/views/spree/admin/orders/_line_items.html.erb
tvdeyen: Not mine ;) d2ec380de431ef498f28c4c2b80086c6fec81afc

jhawthorn:
Can't run test on CircleCI because of AWS issues, but this is passing locally (and was passing before adding CHANGELOG and a rebase)

:shipit: :fire:

PR #1715, opened 2017-02-16
Update admin cart page dynamically

jhawthorn:
Built on top of #1710 

This fully models the displayed information on the cart page in backbone, which allows it to update dynamically without relying on the previous `window.location.reload()`.

This allows for smoother interaction by the user, and avoids undesirable changes from the refresh like closing in-progress edits.


backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -124,6 +124,11 @@ button, .button {
     width: 100%;
     text-align: center;
   }
+
+  &[disabled], &.disabled {
```
vfonic: I think the more preferred way would be to use `:disabled` CSS3 selector. Although they are the same, some explanation why use one over the `[disabled]` can be found here:
http://stackoverflow.com/a/20146936/336806

It works in cross-browser:
http://caniuse.com/#feat=css-sel3

backend/app/views/spree/admin/orders/_adjustments.html.erb
```diff
@@ -1,21 +1,19 @@
...
-        <tr>
-          <th><%= Spree::Adjustment.human_attribute_name(:name)%></th>
-          <th><%= Spree::Adjustment.human_attribute_name(:amount)%></th>
+<fieldset class="no-border-bottom">
+  <legend><%= title %></legend>
+  <table>
+    <thead>
+      <tr>
+        <th><%= Spree::Adjustment.human_attribute_name(:name)%></th>
+        <th><%= Spree::Adjustment.human_attribute_name(:amount)%></th>
```
vfonic: These should be translated:
Screenshot:
Screenshot

vfonic: Nevermind, this is a `solidus_i18n` issue

backend/app/views/spree/admin/orders/_adjustments.html.erb
```diff
@@ -1,21 +1,19 @@
...
+      </tr>
+    </thead>
+    <tbody data-hook="order_details_adjustments"  class="with-border">
+      <% adjustments.eligible.group_by(&:label).each do |label, adjustments| %>
+        <tr class="total">
+          <td><strong><%= label %>:</strong></td>
+          <td class="total align-center"><span><%= Spree::Money.new(adjustments.sum(&:amount), currency: adjustments.first.order.try(:currency)) %></span></td>
```
vfonic: Not related to these changes, but you could add this fix in:
```rb
currency: adjustments.first.order.try(:currency)
```
=>
```rb
currency: order.currency
```

...since you're always passing `order` variable and it's always defined.

backend/spec/features/admin/orders/order_details_spec.rb
```diff
@@ -111,11 +108,14 @@
         expect(page).to have_content("Default")
       end
 
-      it "will show the variant sku" do
-        order = create(:completed_order_with_totals)
-        visit spree.edit_admin_order_path(order)
-        sku = order.line_items.first.variant.sku
-        expect(page).to have_content("SKU: #{sku}")
+      context "with a completed order" do
+        let!(:order) { create(:completed_order_with_totals) }
```
vfonic: I think it's better having this inlined inside of the it block:
```rb
order = create(:completed_order_with_totals)
```

It's always better to have the code contained in the single place, than trying to figure out where did this variable come from (especially if `let!` is used and not `let`).

core/config/locales/en.yml
```diff
@@ -711,6 +711,7 @@ en:
     add_action_of_type: Add action of type
     add_country: Add Country
     add_coupon_code: Add Coupon Code
+    add_line_item: Add Item
```
vfonic: Is there any process for reporting this new translation to `solidus_i18n`? Perhaps an issue could be created on the `solidus_i18n` repo when translation-changes/new-translation-is-added?

backend/app/assets/javascripts/spree/backend/format_money.js
```diff
@@ -0,0 +1,10 @@
+(function() {
+  var defaultLocale = document.documentElement.lang;
+
+  Spree.formatMoney = function(amount, currency, options) {
+    options = (options || {});
+    var locale = options.locale || defaultLocale;
+    var format = new Intl.NumberFormat(locale, {style: 'currency', currency: currency})
```
jhawthorn: Nope. Please feel free to file an issue.

backend/app/assets/javascripts/spree/backend/format_money.js
```diff
@@ -0,0 +1,10 @@
+(function() {
+  var defaultLocale = document.documentElement.lang;
```
jhawthorn: It will always be set. I don't think we need to be defensive here.

backend/app/assets/javascripts/spree/backend/format_money.js
```diff
@@ -0,0 +1,10 @@
...
+
+  Spree.formatMoney = function(amount, currency, options) {
```
jhawthorn: I can appreciate that, but I'm not 100% sure that this will be the permanent implementation of this method. We might have to end up including `accounting.js` or similar, which we would not want cluttering up the frontend. For that reason, I want to keep this backend-only.

backend/spec/features/admin/orders/order_details_spec.rb
jhawthorn: Using let is better for composability and is always preferred. This let actually overrides a `let!(:order)` higher up in the file, which the local definition would not do. This is a good thing.

core/config/locales/en.yml
jhawthorn: There is a rake task in `solidus_i18n` for this purpose. This question is way outside the scope of this PR.

backend/app/assets/javascripts/spree/backend/format_money.js
vfonic: Filed

backend/app/assets/javascripts/spree/backend/format_money.js
vfonic: I guess we can both agree that "temporary fix" tends to stay around much longer than what we planned it to stay. :)

backend/spec/features/admin/orders/order_details_spec.rb
vfonic: Do you mean it's always preferred in `solidus` project(s) or it's generally always preferred? Because I prefer not to use it and I can find some good arguments why it should be avoided:
https://robots.thoughtbot.com/lets-not
http://linduxed.com/blog/2014/08/24/writing-more-readable-rspec-tests/

Of course there are always arguments pro and contra. Just asking out of curiosity.

backend/app/assets/javascripts/spree/backend/format_money.js
jhawthorn: This is not a temporary fix. The frontend and backend have different javascript, different requirements, and different browser support.

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
tvdeyen: Good call 👍

backend/app/views/spree/admin/orders/_adjustments.html.erb
tvdeyen: Good call. We should address this in this PR

backend/app/assets/javascripts/spree/backend/orders/models.js
```diff
@@ -0,0 +1,48 @@
...
+Spree.Order.OrderModel.fetch = function(number, opts) {
+  var options = (opts || {});
+  var model = new Spree.Order.OrderModel({
+    number: number,
+    line_items: [],
+    shipments: []
+  });
+  model.fetch(options);
+  return model;
+}
```
tvdeyen: Maybe we should split this file into files for each model? So devs could override individual models (the order for instance)? Not sure though, if this could ever be the case.

backend/app/assets/javascripts/spree/backend/orders/line_items.js.coffee
```diff
@@ -0,0 +1,97 @@
...
+
+  events:
+    "click": "onClick"
+
+  onClick: ->
+    this.collection.push({})
+
+  render: ->
+    @$el.prop("disabled", !this.collection.length || this.collection.some( (item) -> item.isNew() ))
+
```
tvdeyen: I kinda confuses me, that some files are coffeescript and some are javascript. I would prefer to go with JS from now on. But we should be consistent.

tvdeyen: Also maybe splitting these views into separate overridable files would be preferable?

backend/app/assets/javascripts/spree/backend/orders/order_details.js
```diff
@@ -0,0 +1,56 @@
...
+    this.listenTo(this.model, "change", this.render);
+    this.render()
+  },
+
+  render: function() {
+    var lineItemCount = this.model.get("line_items").length
+    this.$el.toggleClass('hidden', !lineItemCount)
+    this.$('.order-total').text(this.model.get("display_total"))
+  }
+})
```
tvdeyen: Also consider to split this file up

backend/app/assets/javascripts/spree/backend/orders/order_summary.js
```diff
@@ -0,0 +1,41 @@
...
+    this.$('dd.order-included_tax_total').text(this.model.get("display_included_tax_total"))
+
+    this.$('.order-additional_tax_total').toggleClass("hidden", !Number(this.model.get("additional_tax_total")))
+    this.$('dd.order-additional_tax_total').text(this.model.get("display_additional_tax_total"))
+
+    this.$('.order-shipment_state').toggleClass("hidden", !this.model.get("completed_at"))
+    this.$('dd.order-shipment_state').html(this.renderState('shipment_state', this.model.get("shipment_state")))
+
+    this.$('.order-payment_state').toggleClass("hidden", !this.model.get("completed_at"))
+    this.$('dd.order-payment_state').html(this.renderState('payment_state', this.model.get("payment_state")))
```
tvdeyen: Not a backbone expert, but do we really need to use `this.$` over just `$`?

backend/app/assets/javascripts/spree/backend/templates/orders/line_item.hbs
```diff
@@ -0,0 +1,42 @@
...
+{{#if editing}}
+	<a class="save-line-item fa fa-ok no-text with-tip" href="#" title="{{ t "actions.save" }}"></a>
+	{{#unless noCancel}}
+	<a class="cancel-line-item fa fa-cancel no-text with-tip" href="#" title="{{ t "actions.cancel" }}"></a>
+	{{/unless}}
+{{else}}
+	<a class="edit-line-item fa fa-edit no-text with-tip" href="#" title="{{ t "actions.edit" }}"></a>
+	<a class="delete-line-item fa fa-trash no-text with-tip" href="#" title="{{ t "actions.delete" }}"></a>
+{{/if}}
+</td>
```
tvdeyen: I think the indentation in this file messed up somehow.

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
```diff
@@ -95,3 +80,14 @@ form.admin-ship-shipment {
...
   input[type='submit'] { margin-left: 5px; }
 }
+
+table.line-items {
+  input.error {
+    border-color: $color-error;
+  }
+
+  .select2-container.error {
+    border-radius: 4px;
```
tvdeyen: I believe we have a global `$border-radius` variable we can use here.

backend/app/views/spree/admin/orders/_form.html.erb
```diff
@@ -3,34 +3,10 @@
...
-        :order => order,
-        :title => Spree.t(:shipment_adjustments)
-   } %>
-  <%= render :partial => "spree/admin/orders/adjustments", :locals => {
-        :adjustments => @order.adjustments, 
-        :order => order,
-        :title => Spree.t(:order_adjustments)
-   } %>
+  <%= render :partial => "spree/admin/orders/carton", :collection => @order.cartons.order(:shipped_at), :locals => { :order => order } %>
+  <%= render :partial => "spree/admin/orders/shipment", :collection => @order.shipments.order(:created_at), :locals => { :order => order } %>
```
tvdeyen: We should only use Ruby 1.9 Hash syntax if we touch lines of code

backend/app/views/spree/admin/shared/_order_summary.html.erb
```diff
@@ -10,29 +10,28 @@
...
+    <dd class="order-payment_state <%= "hidden" unless @order.completed? %>" id='payment_status'><span class="state <%= @order.payment_state %>"><%= Spree.t(@order.payment_state, :scope => :payment_states, :default => [:missing, "none"]) %></span></dd>
+
     <% if @order.completed? %>
-      <dt><%= Spree::Shipment.model_name.human %>: </dt>
-      <dd id='shipment_status'><span class="state <%= @order.shipment_state %>"><%= Spree.t(@order.shipment_state, :scope => :shipment_states, :default => [:missing, "none"]) %></span></dd>
-      <dt><%= Spree::Payment.model_name.human %>: </dt>
-      <dd id='payment_status'><span class="state <%= @order.payment_state %>"><%= Spree.t(@order.payment_state, :scope => :payment_states, :default => [:missing, "none"]) %></span></dd>
       <dt data-hook='admin_order_tab_date_completed_title'><%= Spree::Order.human_attribute_name(:completed_at) %>:</dt>
       <dd id='date_complete'><%= pretty_time(@order.completed_at) %></dd>
     <% end %>
```
tvdeyen: This view could profit from some indenting and usage of Ruby 1.9 hash syntax.

backend/app/assets/javascripts/spree/backend/orders/line_items.js.coffee
jhawthorn: I also prefer JS, but this file was originally coffeescript and the changes were each incremental

backend/app/assets/javascripts/spree/backend/orders/models.js
jhawthorn: I agree. Going to split these into separate files and also move to a different namespace, `Spree.Models.Order`, which will also make more sense as we add more models (like `Spree.Models.Variant`)

backend/app/assets/javascripts/spree/backend/orders/order_summary.js
jhawthorn: Yes. `this.$` is scoped to the view's element. It's great!

backend/app/assets/javascripts/spree/backend/orders/line_items.js.coffee
tvdeyen: Ok. Fair enough.

backend/app/assets/javascripts/spree/backend/orders/models.js
tvdeyen: 👍🏻 for the namespace

backend/app/assets/javascripts/spree/backend/orders/order_summary.js
tvdeyen: TIL. Indeed very nice 👌🏻

jhawthorn:
@tvdeyen This is ready for review :)

Models and views are now namespaced and in the `models/` and `views/` directories. I believe I have addressed the other feedback.

Tests have passed, but CircleCI is being CircleCI and hasn't updated github.

tvdeyen:
@jhawthorn I see a fixup commit. Maybe you need to rebase once more?

tvdeyen:
🚀

PR #1717, opened 2017-02-17
Fix duplicated Shipments breadcrumb

jhawthorn:
:rofl: This is a silly one

We had the `admin_breadcrumb` declaration too deep within our partials, so it was called once per shipment, rather than a single time.

This moves the call so it is added only once.

### DIFF
--- a/backend/app/views/spree/admin/orders/_shipment.html.erb
+++ b/backend/app/views/spree/admin/orders/_shipment.html.erb
@@ -1,6 +1,3 @@
-<% admin_breadcrumb(plural_resource_name(Spree::Shipment)) %>
-
-
 <%
   manifest_items = Spree::ShippingManifest.new(
     inventory_units: shipment.inventory_units.where(carton_id: nil),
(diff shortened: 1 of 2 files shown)


backend/app/views/spree/admin/orders/_form.html.erb
```diff
@@ -1,3 +1,5 @@
+<% admin_breadcrumb(plural_resource_name(Spree::Shipment)) %>
```
vfonic: Wouldn't it be better to always have `admin_breadcrumb` at the top of every **non-partial** template?

Like this we ensure that the method is called only once. Otherwise, we can't reuse `_form` partial without getting the "Shipments" breadcrumb, no matter which page we're on.

tvdeyen:
Just for the record: We should start to think about how we organize the admin order templates and partials. Ideally we have a template for each order state (or better tab). Having an `edit` template for shipment state is somehow silly.

jhawthorn:
> Just for the record: We should start to think about how we organize the admin order templates and partials. Ideally we have a template for each order state (or better tab). Having an edit template for shipment state is somehow silly.

The names aren't ideal, but it's never made visible to the user. IMO it's not worth changing them due to the incompatibilities it will cause. Either way, ourside the scope of this PR.

vfonic:
> The names aren't ideal, but it's never made visible to the user. IMO it's not worth changing them due to the incompatibilities it will cause. Either way, outside the scope of this PR.

I think that developers should also be considered users. It will be easier for solidus developers to jump on certain code part and make changes, as well as to solidus developers-users, to figure out how to use the solidus code API (not JSON API).

Issue #666, opened 2016-01-07
Collapsible left nav


Mandily:
We can gain some screen real estate for heavy content pages by introducing a collapsible left navigation. Users that are working on small screens can manually reduce the width of the menu, while users on bigger screens can simply ignore it and not be affected. 

While the menu is collapsed, the icons will still be visible and link to their respective pages. All sub-navs will also be collapsed, but still accessible by hovering over the icon. 

Clicking on the collapsed menu will un-collapse the menu so that the user has automatic access to child menu items. 

Because some of the menus can bit longer and may even trigger a scrollbar on small screens, I've put the button at the top of the navigation instead of after the nav links.

![collapsible-nav](https://cloud.githubusercontent.com/assets/15271353/12185562/778dd79a-b550-11e5-95c4-6a11a7b8ce93.jpg)

Please let me know your thoughts in the comments below.

tvdeyen:
This would be a great addition. :clap: 

We need to assure, that the label of the menu item is visible while hovering over it. In current Spree 3 admin the label is not shown and therefore it's hard to guess what icon is the desired one.

<img width="1443" alt="monosnap 2016-01-08 00-25-23" src="https://cloud.githubusercontent.com/assets/42868/12185903/8a80a6ea-b59e-11e5-8a47-c3adabd61ec8.png">

Mandily:
That's a good point. I've run into that in other interfaces as well. Are you picturing a browser tooltip-like thing or something more like the UI of the sub nav flyout?

tvdeyen:
Like the fly out. We shouldn't introduce another visual for the main navigation. That could be confusing.

sukhchander:
@tvdeyen @Mandily i've implemented the collapsed nav as well as hover. who is working on the implementation of the refactored admin? i'd like to contribute or should i just create a fork?

expanded:
<img width="1230" alt="screenshot 2015-12-22 11 28 24" src="https://cloud.githubusercontent.com/assets/110444/12201818/5b6b5a70-b5f4-11e5-8f4b-f1f78ce22625.png">

collpased:
<img width="1228" alt="screenshot 2015-12-22 11 28 08" src="https://cloud.githubusercontent.com/assets/110444/12201827/6a073478-b5f4-11e5-9270-82d78876091c.png">

tvdeyen:
@sukhchander I guess @Sinetheta will do the implementation, but please feel free to send a PR. We are happy about every contribution.

[...]

PR #1718, opened 2017-02-17
[WIP] Collapsible menu 

davidedistefano:
This PR makes the Solidus admin more usable at small screen sizes.

At medium size devices, like small desktops or big tablets, the sidebar will collapse, leaving visible only the menu icons.

schermata 2017-02-06 alle 18 52 43

At mobile screen sizes instead, the menu will only be visible by clicking on the top right hamburger menu.

schermata 2017-02-06 alle 18 52 59

To make this PR works also on devices like smartphone or tablet the following line of code should be added inside the app `<head>` element:

`<meta content="text/html; charset=UTF-8" http-equiv="Content-Type" />
`

The admin layout needs some tweaks to work fine with new sidebar. I suggest to use a fluid layout instead of the old Skeleton grid.


backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -2,6 +2,49 @@ $padding-x-navbar: 26px;
...
+
+    .admin-content  {
+      transform: translateX($width-sidebar);
+      @include media($large-device) {
+        transform: none;
+      }
+    }
+  }
+}
+
```
tvdeyen: IMO a small transition would enhance UX:

```css
.admin-content, .admin-nav {
  transition: transform .2s ease-in;
}
```

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -69,40 +133,54 @@ nav.menu {
...
+        }
       }
-    }
 
-    &.selected li {
-      padding-left: 2em;
-    }
+      > ul {
+        @include media($desktop) {
+          @include flyout;
```
tvdeyen: IMO all breakpoints can profit from a flyout. Is there a particular reason you hide them from small viewports?

backend/app/assets/stylesheets/spree/backend/globals/_variables.scss
```diff
@@ -206,3 +207,14 @@ $z-index-navbar-flyout:          1000 !default;
 // Sidebar
 //--------------------------------------------------------------
 $border-sidebar:                 1px solid $color-sidebar-border !default;
+
+// Media queries
+//--------------------------------------------------------------
+$desktop-width: 959px;
```
tvdeyen: We should add `!default` here, so the variable can be overriden by shop devs.

backend/app/assets/stylesheets/spree/backend/globals/mixins/_media.scss
```diff
@@ -0,0 +1,28 @@
...
+///  .responsive-element {
+///    color: blue;
+///  }
+///  @media screen and (min-width: 480px) {
+///    .responsive-element {
+///      color: red;
+///    }
+///  }
+
+@mixin media($media-query) {
```
graygilmore: I don't think this media query is necessary. It actually requires more characters than not having it:

```scss
.element {
  background: blue;

  // With mixin
  @include media($desktop) {
    background: green;
  }

  // Without mixin
  @media #{$desktop} {
    background: green
  }
}
```

graygilmore:
I don't see a way to toggle the collapseness. This was a (kind of hidden) requirement from the issue: https://github.com/solidusio/solidus/issues/666#issuecomment-211585537

backend/app/assets/javascripts/spree/backend/navigation.coffee
```diff
@@ -4,7 +4,18 @@ navHeight = ->
...
 
-$ ->
-  $(".admin-nav-sticky, .admin-nav").stick_in_parent()
+checkSticky = ->
+  stickyElements = $('.admin-nav-sticky, .admin-nav')
+
+  if $(window).width() < 768
+    stickyElements.trigger 'sticky_kit:detach'
+  else
+    stickyElements.stick_in_parent(spacer: false)
```
davidedistefano: Thanks for your feedback! Can you please tell me what browser you are using and at which window width do you see this jumpy scrolling behaviour.

backend/app/assets/stylesheets/spree/backend/globals/mixins/_media.scss
davidedistefano: @graygilmore thanks for your feedback! We thought that having a mixin would be easier to add features to media query in future development, without updating code in the whole project. Also the mixin version seems to be a little more user friendly and more "readable". 

But your suggestion also makes sense and we're open to change code in this direction if you want.

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
davidedistefano: Actually we tried to add a CSS transition to the nav, but while using only CSS allows you to have a cleaner code, you don't have the same power as you have with Javascript :) 

We found a bug in particular that occurs while resizing the window, as explained by this GIF:

transition-menu-bug

So we thought that in order to avoid focusing too much on the transition effect, we'd implement a first version of the collapsible navigation without effect.

Any suggestion to avoid this bug?

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
Mandily: The flyout is triggered by hovering, so it doesn't actually work on a touch screen. it's more of a bonus for desktop users than a feature removed for mobile though. With a single tap they can still open the menu to access the lower levels.

tvdeyen: A small screen does not equal a touch screen and vice versa. And this PR actually added more code to make this work "only" for desktop. That's why I proposed to remove this refactoring and leave the flyout even for small screens.

backend/app/assets/stylesheets/spree/backend/globals/mixins/_media.scss
tvdeyen: Good catch @graygilmore

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
tvdeyen: I would not consider this a bug, but I also don't persist on this animation at all. It was just an idea for a little nicer UX.

backend/app/assets/javascripts/spree/backend/navigation.coffee
tvdeyen: Latest Chrome, small desktop viewport (~1024) lower the screen height so the sticky feature get's triggered.

backend/app/assets/stylesheets/spree/backend/globals/mixins/_media.scss
graygilmore: @davidedistefano I don't think the added complexity of the mixin helps us here.

backend/app/assets/javascripts/spree/backend/navigation.coffee
kennyadsl: it happens when page content is shorter than menu height, looking for a fix

kennyadsl: also, it doesn't seem to be related to having the spacer or not, it happens even with it

kennyadsl: @tvdeyen should be fixed with https://github.com/solidusio/solidus/pull/1718/commits/8e9a39f3ad78bda18324b46ed4e73462556116f6, what do you think?

backend/app/assets/stylesheets/spree/backend/globals/_variables.scss
kennyadsl: done

backend/app/assets/javascripts/spree/backend/navigation.coffee
tvdeyen: @kennyadsl I'm not sure about removing the sticky kit yet. @Sinetheta WDYT?

backend/app/assets/stylesheets/spree/backend/globals/_variables.scss
```diff
@@ -206,3 +207,14 @@ $z-index-navbar-flyout:          1000 !default;
...
+// Media queries
+//--------------------------------------------------------------
+$desktop-width: 959px !default;
+$tablet-width: 768px !default;
+$mobile-width: 480px !default;
+
+$large-device: "only screen and (min-width: #{$tablet-width})";
+$desktop: "only screen and (max-width: #{$desktop-width})";
+$tablet: "only screen and (max-width: #{$tablet-width})";
+$mobile: "only screen and (max-width: #{$mobile-width})";
```
mtomov: The problem with this stuff is that you then need documentation for it. That's why something standardized (Bootstrap) should provide for all that boilerplate. For example, it's very elegant in Foundation:

- http://foundation.zurb.com/sites/docs/media-queries.html#sass

```
  @include breakpoint(medium) {
    // All CSS in here goes inside the media query
  }
```

If Bootstrap doesn't provide for such a mixin, then we can probably steel the one from Foundation, although it's unlikely that there isn't one such there.

kennyadsl: @mtomov agree, but as long as we have skeleton layout I think we need to consider that exactly these breakpoints will trigger screen resizes. We just moved those into variables to be able to use them in all pages consistently.

mtomov: Yeah, but Bootstrap is already part of Solidus, so all mixins it has should be available.

- https://v4-alpha.getbootstrap.com/layout/overview/#responsive-breakpoints

Try the above ones, they should work:

```
@include media-breakpoint-up(xl) { ... }
```

kennyadsl: sorry, I don't get it. This way it will use the bootstrap sizes for `xl` etc screen sizes but I think we need to use the skeleton ones which are defined [here](https://github.com/solidusio/solidus/blob/master/frontend/app/assets/stylesheets/spree/frontend/_skeleton.scss) that could not match, right?

mtomov: The link are pointing to is for the frontend. 

And yes, hopefully they should match, otherwise we need them to match, otherwise we can't use Bootstrap ..

So, [this must be the bootstrap configuration](https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss#L107-L119) for the backend.

The [skeleton one for the backend](https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/shared/_skeleton.scss#L132) seems to have most of the values pretty fixed : ) .. so yeah, hopefully they match ..

kennyadsl: Sorry for the wrong link. Cool, thanks, I didn't noticed those variables in backend. I'll give these mixins a try soon! :)

mtomov: super : )

kennyadsl:
I've implemented a basic collapse menu system, basically it adds a label in the menu that change the checkbox that show the full menu, same as what the hamburger menu (only visible if <= tablet).

#### Collapsed

menu-collapsed


#### Open

menu-open


#### Tablet

menu-tablet


This PR still needs:

- [ ] understand if we can use bootstrap media queries, as @mtomov [pointed out in a comment](https://github.com/solidusio/solidus/pull/1718#discussion_r107908937).
- [ ] cleanup details, there's some alignment that is not working in the menu
- [ ] understand if we can remove stiky-kit jQuery plugin and move to css `position: sticky`
- [ ] make menu collapsed/not collapsed based on the devise width (now it always start collapsed)
- [ ] remember users choice about collapsed menus on each device (probably we can make a separate PR)
- [ ] add meta tag `<meta name="viewport" content="width=device-width, initial-scale=1">`
- [ ] show a label for elements that don't have a submenu
- [ ] remove hover state (color) on hamburger menu on touch devices
- [ ] try to add [a nice transition](https://github.com/solidusio/solidus/pull/1718#discussion_r102092456)
- [ ] make it work (if possible) with pages that don't have the`new-layout` body class

kennyadsl:
closing this one since lot of layout things has changed and after working on this I think we should use a different approach (more JS driven).

PR #1722, opened 2017-02-19
ruby 1.9 style hash colons

brchristian:


vfonic:
Finally someone with OCD like me! 🎉 

Although I like this change, I don't know whether it's considered to be a "religious war" that you're fighting, as I've realized that solidus code style guidelines are very loose. It looks like you can write your code any way you want as long as it looks readable (no fighting over hash styles, single/double quotes, etc.)
This change affects a lot of files and will require most of current PRs to be updated (or this PR to be updated again after those other PRs have been merged).

kennyadsl:
I'm ok updating the hash syntax to the new version but I think it should be done gradually to avoid losing a clear git history, which in this kind of projects is more important IMHO.

vfonic:
👍 I fully agree. I think the syntax should be updated as the changes are done to the file/to the code with old syntax.

brchristian:
If the general consensus is that a clean history is more important than consistent style, then that makes sense to me.

Would it be worth adding a small section to the contributor guide to the effect of "here’s our style guide, we want new commits to look this way, but as a policy do not accept strictly cosmetic PRs?"

kennyadsl:
@brchristian that section would be awesome anyway in my opinion and I think that the style guide for new commits should be done just following rubocop rules. 

Let's wait for other feedbacks on this PR since I'm not 100% sure everyone agree on this topic.

cbrunsdon:
We had a similar discussion much earlier (back when we were still a "new" fork) and decided not to do this as we thought it would be much harder to cherry-pick spree commits, and had a large number of other branches lying around.

Right now though I think we're in a pretty good state, and there hasn't been any commits we've cherry picked from spree in the last few years, so my personal preference is we go ahead and do this. Whether we do it slowly, or do it in a "big bang", the work has to be done and the "blame" for that line is going to be off, but right now we're only talking some ~650 lines which really isn't many.

brchristian:
(I realized that I had missed about 50 instances which were immediately preceded by an open brace or parenthesis, so I’ve added a second commit that takes care of those as well.)

kennyadsl:
@brchristian 👍  if we'll keep it maybe it's better to squash them in a single commit

brchristian:
@kennyadsl squashed! 👍

jhawthorn:
While I would like to have only migrated files as they were touched (and better preserving history), I feel like changes to hash syntax (and requests for those changes) clutter up code reviews. I think this will be a very good change for us.

Thank you @brchristian

PR #1725, opened 2017-02-20
Admin: Make actions table columns show inside of a table

vfonic:
As per discussion on slack and in here: https://github.com/solidusio/solidus/pull/1715

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/templates/variants/autocomplete_stock.hbs
+++ b/backend/app/assets/javascripts/spree/backend/templates/variants/autocomplete_stock.hbs
@@ -11,7 +11,7 @@
         <th>{{ t "location" }}</th>
         <th>{{ t "count_on_hand" }}</th>
         <th>{{ t "quantity" }}</th>
-        <th class="actions"></th>
+        <th class="actions">{{ t "actions" }}</th>
       </thead>
       <tbody>
         {{#each variant.stock_items}}
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -855,6 +855,8 @@ en:
         taxonomies: Taxonomies
         taxons: Taxons
         users: Users
+      table:
+        actions: Actions
       taxons:
         display_order: Display Order
       user:
--- a/backend/app/assets/stylesheets/spree/backend/components/_actions.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_actions.scss
@@ -24,8 +24,4 @@ table tbody tr {
       }
     }
   }
-
-  td.actions {
-    background-color: transparent !important;
-  }
 }
--- a/backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
@@ -33,8 +33,6 @@ table {
     }
 
     &.actions {
-      background-color: transparent;
-      border: none !important;
       text-align: center;
 
       span.text {
--- a/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
@@ -23,7 +23,7 @@
       <tr data-hook="adjustment_reasons_header">
         <th><%= Spree::AdjustmentReason.human_attribute_name(:name) %></th>
         <th><%= Spree::AdjustmentReason.human_attribute_name(:state) %></th>
-        <th class="actions"></th>
+        <th class="actions"><%= Spree.t('admin.table.actions') %></th>
       </tr>
     </thead>
     <tbody>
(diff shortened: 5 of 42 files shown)


backend/app/assets/javascripts/spree/backend/templates/variants/autocomplete_stock.hbs
```diff
@@ -11,7 +11,7 @@
         <th>{{ t "location" }}</th>
         <th>{{ t "count_on_hand" }}</th>
         <th>{{ t "quantity" }}</th>
-        <th class="actions"></th>
+        <th class="actions">{{ t "actions" }}</th>
```
vfonic: I'm not sure how does handlebars know to translate this correctly?

I haven't seen any translations for, for example: `{{ t "select_stock" }}` on line 2

core/config/locales/en.yml
```diff
@@ -855,6 +855,8 @@ en:
         taxonomies: Taxonomies
         taxons: Taxons
         users: Users
+      table:
```
vfonic: using singular here, like it's used for `tab` above, could change if needed

Mandily:
I'm not sure we need a label on the column "actions". But this works great otherwise.

isaacfreeman:
Having the "Actions" column label clarifies the purpose of the wordless action buttons. I think this is valuable, because the actions are typically pretty abstract and hard to capture in an icon. On many admin pages they're also used frequently, despite being small click targets. So anything that gives them some context could be a big help.

It might also be some small help to people using screen readers, but obviously there's much more to do there.

Mandily:
It sounds like you're trying to solve a less than intuitive button design here. 

If we're having issues with people understanding the icons in the button, I think we should redesign the button to make them more clearly understood on first glance. If users didn't know that these buttons cause "actions" to happen, we need to design a little more affordance into our button styles so they look interactive. 

Do you have any kind of data on this that would lead us to believe users are having difficulties with the buttons? I've wondered myself on and off for a while if simple text buttons would do better or worse, but without understanding the current situation, it's hard to improve.

isaacfreeman:
@Mandily: Good point that I'm conflating two different problems. If we were focusing on making the action buttons easy to understand, adding an "Actions" label at the top of the column isn't a solution I'd propose. And I don't have any particular data on users' experience with the action buttons, so it makes sense to treat that as a separate issue.

vfonic:
I just did a quick Google search and it seems like there's no consensus whether or not to use "Actions" label or leave it blank. I prefer having "Actions" heading as the icons might not be intuitive, as @isaacfreeman pointed out and it provides a little bit more of a structure, by answering a question: "What does this column represent/contain?"

Fun fact: On the other hand, code-wise, if I remove "Actions" title, this PR will be much smaller.
Here are the screenshots with and without the "Actions" title:
screenshot 2017-02-22 08 45 35
screenshot 2017-02-22 08 45 48

Please also note that, on tables that have sorting by column, actions title remains black/unsortable.

tvdeyen:
I opt in for let's add a Actions header.

backend/app/assets/javascripts/spree/backend/templates/variants/autocomplete_stock.hbs
jhawthorn: This should be `{{ t "admin.table.actions" }}`

jhawthorn:
:-1: From me, I'm afraid

I'd like to echo what @Mandily said. We need to identify exactly what problem we are trying to solve here and come up with the best solution for that problem.

Having the buttons inside the table looks worse to me. Maybe it's just familiarity with the existing look, but it seems really off.
 
I believe this is a worse UX language than what we had previously. Having the button visually outside the row makes it clear that the edit button refers to the entire row (the entire order in this screenshot).

tvdeyen:
Let me rephrase:

**If** we are putting the icons into the table row, then a 'Actions' table header would be preferable. 

But I leave the decision on are we putting the icons inside or outside the (visible) table row (because technically they are inside) to @Mandily

Mandily:
My future dream/end game is to have a thing that looks like this at some point. Moving the buttons inside the table is one step closer to highlighting a full row with them included. So my vote it inside the table. 

screen shot 2017-02-22 at 1 46 33 pm

vfonic:
@jhawthorn It was confusing for me, who is unfamiliar with the interface, so I asked and proposed the changes. I don't know of any UI where the action icons are outside of the table and table row they refer to. It's not clear, especially when there's only one row: there are some buttons on the right of the table, there are some buttons below the table...

I see others mostly agree with putting action icons inside of the table. Are you ok with that or you have strong opinion against?

What about "Actions" label? 👍 / 👎 ? I'd like to keep the label (obviously, since I made the change)

Mandily:
To go back to the point about solving specific problems - I don't think this solves a problem. I thought it might be a step in the direction of the screenshot I included above, but after thinking about it further, it's not, because that screenshot is about removing lines and cleaning up the overall appearance. Adding lines around the buttons now is not setting us up to do that in the future.

As all English and many other languages read from left to right first, I don't see a problem with understanding that the buttons on the right are indeed related to the row at the left.

I'm in favour of not merging this PR.

vfonic:
I think it does. I opened this PR exactly because to me, those buttons looked like they could belong to the table, but maybe also not. Also, it wasn't clear what those buttons represented. Of course, it's clear once you hover over them, but not before that.
I'm not talking about moving towards future version of design. I'm talking about improving current design. And that's what we should focus right now, while we don't have the new design.

Anyway, I guess this won't get merged. I've never seen UI look like this and I disagree with leaving these changes out.

PR #1727, opened 2017-02-20
Update code styles for /cart

vfonic:
* Use ruby 1.9 hash style
* Don't use @ instance variable in partial
* Replace `order_form.object` with `order`
* Replace
`render :partial => '...', :locals => { ... }`
with
`render '...', { ... }`

### DIFF
--- a/frontend/app/views/spree/orders/edit.html.erb
+++ b/frontend/app/views/spree/orders/edit.html.erb
@@ -16,7 +16,7 @@
         <div data-hook="inside_cart_form">
 
           <div data-hook="cart_items">
-            <%= render partial: 'form', locals: { order_form: order_form } %>
+            <%= render 'form', order_form: order_form %>
           </div>
 
           <div class="links columns sixteen alpha omega" data-hook="cart_buttons">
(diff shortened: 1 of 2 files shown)


frontend/app/views/spree/orders/edit.html.erb
```diff
@@ -16,7 +16,7 @@
         <div data-hook="inside_cart_form">
 
           <div data-hook="cart_items">
-            <%= render :partial => 'form', :locals => { :order_form => order_form } %>
+            <%= render 'form', order_form: order_form, order: @order %>
```
jhawthorn: Probably don't need to pass both `order_form` and `order` in, since `order` can be found via `order_form.object`.

If the goal is to make the code in the form partial nicer, it would be better to just store `order` as a local there.

vfonic: You mean to have this in _form:
```erb
<% order = order_form.object %>
```
?

vfonic: Updated

vfonic:
Resolved merge conflicts with `master`

jhawthorn:
Thank you

PR #1730, opened 2017-02-22
Improve JS i18n

jhawthorn:
A few improvements to our (very basic) JS i18n

In c871eea69c6769140255be3a5c0c1e183133db33 @mtomov made the Handlebars `t` helper understand the dot-separated paths. :+1:

This extracts that functionality to a `Spree.t`, helper, so that JS doesn't need to dig through the `Spree.translations` object manually.

``` js
Spree.t("thing.one.two")
```

This also adds a `scope` option similar to the one ruby's `18n.t` accepts. This is especially useful from handlebars templates, which can't themselves concatenate the string keys.

``` js
Spree.t(order.state, {scope: "order_state"})
```

``` handlebars
{{ t order.state scope="order_state" }}
```

Finally, this adds `Spree.human_attribute_name`, as well as ships the activerecord attribute translations to our JS.

``` js
Spree.human_attribute_name("spree/line_item", "description")
```

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/translation.js
+++ b/backend/app/assets/javascripts/spree/backend/translation.js
@@ -18,4 +18,8 @@
       return key;
     }
   }
+
+  Spree.human_attribute_name = function(model, attr) {
+    return Spree.t("activerecord.attributes." + model + '.' + attr);
+  }
 })();
--- a/backend/app/assets/javascripts/spree/backend/translation.js
+++ b/backend/app/assets/javascripts/spree/backend/translation.js
@@ -9,7 +9,11 @@
       }, obj || self);
   }
 
-  Spree.t = function(key) {
+  Spree.t = function(key, options) {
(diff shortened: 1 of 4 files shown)


backend/app/assets/javascripts/spree/backend/translation.js
```diff
@@ -0,0 +1,29 @@
...
+  Spree.t = function(key, options) {
+    options = (options || {});
+    if(options.scope) {
+      key = options.scope + "." + key;
+    }
+    var translation = resolveObject(key, Spree.translations);
+    if (translation) {
+      return translation;
+    } else {
+      console.error("No translation found for " + key + ".");
```
tvdeyen: I would consider to use a `console.warn` instead

jhawthorn: Good call. Done.

PR #1733, opened 2017-02-22
Add missing relation: `address.user_address` and `address.user`

vfonic:
I've got couple of questions that I need help answering:
1. Should I add any tests for this change?
2. I'm not sure whether this will work out-of-the-box when app uses custom User class

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -11,6 +11,9 @@ module Spree
     belongs_to :country, class_name: "Spree::Country"
     belongs_to :state, class_name: "Spree::State"
 
+    has_many :user_addresses, -> { active }
+    has_many :users, through: :user_addresses
+
     validates :firstname, :address1, :city, :country_id, presence: true
     validates :zipcode, presence: true, if: :require_zipcode?
     validates :phone, presence: true, if: :require_phone?


jhawthorn:
I don't think this can be done, since in the current implementation a single `Address` could be used by multiple Users.

vfonic:
Oh, I didn't know that. Well, in that case, I can replace `has_one` with `has_many`:
```rb
  has_many :user_addresses, -> { active }
  has_many :users, through: :user_addresses
```

This was my first implementation, but I thought it's impossible for an address to be shared by multiple users, so I changed it to `has_one`.

vfonic:
Hi @jhawthorn, what do you think of `Address.has_many :user_addresses`? I think this should work for every case.

vfonic:
Bump!

I think the changes I made after @jhawthorn's comment should work for every case and would not require users who need this functionality to have to do `class_eval` to add this functionality.

gmacdougall:
Hi Viktor,

Thanks for contributing this. Though your implementation is reasonable, we're concerned about exposing this complex relationship (an address having many users) to Solidus developers, as such we'd prefer to avoid doing this. It's a relationship we're not sure is going to exist in this manner for the long term, so we're going to leave this as is for now.

Thanks for the effort and discussion. Sorry for the slow response and please feel free to contribute again in the future!

PR #1736, opened 2017-02-22
Add a .eslintrc and lint our JS

jhawthorn:
This adds a fairly relaxed `eslintrc` for linting our JS.

This adds very few rules, as I'd like to avoid bikeshedding at this point in time, and instead relies on the [eslint:recommended](http://eslint.org/docs/rules/) rule which covers the most important rules. Notably, these rules will catch missed `var` declarations :tada:.

This works around two ugly globals we have `show_flash` and `update_state`, which should be removed at a later date, but are outside the scope of this PR.

I skipped a few files, which are addressed in my other open PRs
```
spree/backend/line_items_on_order_edit.js
spree/backend/payments/new.js
spree/backend/shipments.js
```

### DIFF
--- a/backend/.eslintrc
+++ b/backend/.eslintrc
@@ -1,0 +1,21 @@
+{
+  "rules": {
+    "no-unused-vars": ["error", { "vars": "all", "args": "none" }],
+    "linebreak-style": ["error", "unix"]
+  },
+  "env": {
+    "es6": false,
+    "browser": true,
+    "jquery": true
+  },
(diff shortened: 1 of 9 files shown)


tvdeyen:
👍 could you also enable javascript on `.hound.yml` then? 🍰

forkata:
:tada: This is great, surprised at how few changes were required!

stewart:
:+1:, thank you.

Issue #1237, opened 2016-06-10
ShippingMethod display: backend does not work properly


Murph33:
Setting a shipping method to display backend makes it not be displayed anywhere.  

https://github.com/solidusio/solidus/blob/master/core/app/models/spree/stock/estimator.rb#L14

https://github.com/solidusio/solidus/blob/master/core/app/models/spree/shipment.rb#L175-L198

https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/orders/_shipment.html.erb#L50-L52

The main (only?) way of generating shipping rates does not take a param to pass to the Estimator which defines new shipping rates so we always have `frontend_only = true`.  The shipping methods usable in the backend are found off of a shipments shipping rates.  In the front end shipping methods are commonly shown to customers through the shipments shipping rates so any order looked at in the backend would give the customer access to those shipping rates.

florianguenther:
What is the preferred solution for this?
Is someone working on the issue?

ericsaupe:
Is this resolved with #1984?

jhawthorn:
I believe so

PR #1739, opened 2017-02-23
Display Backend only rates in the backend

mamhoff:
Prior to the commit, shipping methods that had `display_on` set to
`back_end` would not be displayed in the backend. This commit changes
`Spree::Order#refresh_shipment_rates` and `Spree::Shipment#refresh_rates`
to accept an optional argument `frontend_only`. If that is set to `false`,
the estimator will also take into account backend only rates.

Also, the places where `@order.refresh_shipment_rates` are called
in the backend now pass `frontend_only: false` to those methods.

### DIFF
--- a/backend/spec/features/admin/orders/order_details_spec.rb
+++ b/backend/spec/features/admin/orders/order_details_spec.rb
@@ -118,6 +118,24 @@ describe "Order Details", type: :feature, js: true do
         end
       end
 
+      context "with a backend only shipping method" do
+        let!(:backend_only_shipping_method) { create(:shipping_method, name: "Backend Shipping", available_to_users: false) }
+
+        it "can change the shipping method for a backend only one" do
+          order = create(:completed_order_with_totals, frontend_viewable: false)
+          visit spree.edit_admin_order_path(order)
+          within("table.index tr.show-method") do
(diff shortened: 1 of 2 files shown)


jhawthorn:
The code looks good, I don't think this should be done in this way.

If a customer is in the process of checking out an order, and an admin views their order in the admin, the customer will be given the option of selecting the backend-only rates.

Maybe this could be done only for orders with `frontend_viewable = false` (only orders created in the admin)?

mamhoff:
What about the case where an admin edits a customer-created order to give them a shipping method the customer can not access? 

I get the point about not giving customers access to backend only rates; I'd amend the PR so that the frontend selects the rates by their frontend-viewability, too.

mamhoff:
If we do it via `frontend_viewable`, we can entirely pass on the second argument to `estimator.shipping_rates(package, frontend_only =  true)`.

mamhoff:
There is an [issue for this problem as well](https://github.com/solidusio/solidus/issues/1237)

puranjay6:
Can i now see the backend only shipping methods ?

mamhoff:
Closing this in favour of #1984.

PR #1743, opened 2017-02-24
Display Backend only rates in the backend

mamhoff:
Prior to the commit, shipping methods that had `display_on` set to
`back_end` or `available_to_users`set to `false` would not be displayed
in the backend. This commit changes `Spree::Order#refresh_shipment_rates`
and `Spree::Shipment#refresh_rates` to accept an optional argument
`frontend_only`. If that is set to `false`, the estimator will also
take into account backend only rates.

Also, the places where `@order.refresh_shipment_rates` are called
in the backend now pass `frontend_only: false` to those methods.

This is the same as https://github.com/solidusio/solidus/pull/1739, but for Solidus 1.4.

### DIFF
--- a/backend/app/controllers/spree/admin/orders/customer_details_controller.rb
+++ b/backend/app/controllers/spree/admin/orders/customer_details_controller.rb
@@ -29,7 +29,7 @@ module Spree
 
             unless @order.completed?
               @order.next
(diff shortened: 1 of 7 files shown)


mamhoff:
I'm waiting for the discussion on #1739.

PR #1744, opened 2017-02-24
Disable rubocop Metrics/BlockLength

jordan-brough:
This cop was added to rubocop a few months ago.  I agree shorter blocks are
usually better but this doesn't seem like it's something we need in rubocop and
it also has [some issues with specs](https://github.com/bbatsov/rubocop/issues/3664#issuecomment-255481296) apparently.

I noticed because it started spamming this PR: https://github.com/solidusio/solidus/pull/1707

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -226,6 +226,9 @@ Metrics/ClassLength:
 Metrics/ModuleLength:
   Enabled: false
 
+Metrics/BlockLength:
+  Enabled: false
+
 Metrics/CyclomaticComplexity:
   Enabled: false
 


peterberkenbosch:
There is a better option I think, we can disable this for the specs but still have it for the system. Ran into this some time ago as well: https://github.com/StemboltHQ/solidus_affirm/pull/6/commits/106096c6a57d575b9a4a41c02838444e18a6a439

jhawthorn:
I don't really care for any of the rules in `Metrics/*`. They're certainly good goals, but I think humans are the best judges of such things.

PR #1745, opened 2017-02-24
Implement Spree.formatMoney using accounting.js

jhawthorn:
This adds a `Spree.formatMoney(amount, currency)` helper for our JS components to use. The goal is for it to format money identically to how `::Money#format` does server-side.

We dump information about all supported currencies from `Money::Currency.all` to a JSON hash. This JSON stores the symbol, number of decimal places, and format (symbol first or last) for each currency.

This information is used to call [accounting.js](http://openexchangerates.github.io/accounting.js/)'s `formatMoney` with appropriate arguments for the currency.

``` js
Spree.formatMoney(10, 'USD')
"$10.00"
Spree.formatMoney(10, 'EUR')
"€10.00"
Spree.formatMoney(10, 'JPY')
"¥10"
Spree.formatMoney(10000, 'USD')
"$10,000.00"
```

### Why not `Intl.NumberFormat` ?
Originally I had wanted to implement this using  [the ECMAScript Internationalization API](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Intl), but that had a few issues:

* Formatting was inconsistent between ruby `::Money#format` and JS `Intl.NumberFormat` (`$10.00` vs `CA$ 10.00`). If we want to be able to use this to replace or enhance existing non-interactive components, they need to be identical.
* PhantomJS (even 2.1.1) doesn't support the ECMAScript Internationalization API at all.
* The full polyfill, [https://github.com/andyearnshaw/Intl.js/](Intl.js), is huge, about 600k. It's much smaller if only the current locale is included, but that presents its own challenges.


backend/app/assets/javascripts/spree/backend/format_money.js.erb
```diff
@@ -0,0 +1,27 @@
...
+    else
+      "%v %s"
+    end
+  [c.id.to_s.upcase, [
+    c.symbol || "¤",
+    c.exponent,
+    format
+  ]]
+}.to_h.to_json
+%>;
```
tvdeyen: We can't store this information like this as Sprockets won't pick up changes to `Money` settings while compiling new assets. Means, every time you change a `Money` setting you have to delete the assets on the server and force recompilation. 

To avoid this, we need to store this information in a html.erb file, like we already do for the JS translations data (`backend/app/views/spree/admin/shared/_translations.html.erb`)

backend/spec/features/admin/javascript_format_money_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  it 'should behave identically to Spree::Money#to_s' do
+    visit '/admin'
+
+    Money::Currency.all.map(&:id).map(&:to_s).map(&:upcase).uniq.each do |currency|
+      money = Spree::Money.new(1234, currency: currency)
+
+      js_result = page.evaluate_script("Spree.formatMoney(#{money.to_d}, '#{currency}')")
+
+      expect(js_result).to eq money.to_s
+    end
```
tvdeyen: I actually find this a very good way to test this. 👍

backend/app/assets/javascripts/spree/backend/format_money.js.erb
jhawthorn: A fair point, I hadn't considered them changing often enough to warrant this, but I suppose if they're changed at all the existing method would be a pain.

I've pushed up a change which combines this and the existing translations into a `_js_locale_data` partial

backend/app/views/spree/admin/shared/_head.html.erb
```diff
@@ -32,7 +32,7 @@
 
 <%= javascript_include_tag 'spree/backend/all' %>
 
-<%= render "spree/admin/shared/translations" %>
+<%= render "spree/admin/shared/js_locale_data" %>
```
tvdeyen: Good idea 👌🏻 Much more convenient. 

This deserves a CHANGELOG entry though, as people may have modified one of these files. Even deface overrides will not work anymore as the file name changed.

jhawthorn:
> Thanks. I think we should add a CHANGELOG entry as we moved around files.

Done

jhawthorn:
Can't run tests because of AWS issues affecting CircleCI, but this is passing locally (and was passing before adding CHANGELOG and a rebase)

:shipit: :fire:

PR #1747, opened 2017-02-27
Do not clear roles when no role params are present

tvdeyen:
There is a regression where updating a user in the admin without spree role ids in the params causes the current user roles being wiped out.

Also fixed a spec that only passed because of this bug being present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@
 *   New method Spree.formatMoney(amount, currency) is available to javascript in
     the admin.
 
+*   Fix an issue where updating a user in the admin without specifying roles in
+    the params would clear the user's roles.
+
 *   Deprecations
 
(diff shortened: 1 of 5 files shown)


tvdeyen:
@Senjai you introduced the fix in 72f8a0cc5f334d91e5424c2b2ef37b63bfec2063
Could you please review?

tvdeyen:
This also needs to be backported to v1.4 as this bug is also present in 1.4.0

tvdeyen:
Just saw that we need a fix for the missing "empty role ids param" in the users form and feature spec.

Let me add this as well

tvdeyen:
@Senjai done

Senjai:
@tvdeyen lgtm, thanks.

tvdeyen:
Oh. Thanks John. Let me rebase this.

tvdeyen:
@jhawthorn rebased

PR #1748, opened 2017-02-27
1.4 - Do not clear roles when no role params are present

tvdeyen:
Same as #1747 but for v1.4


mamhoff:
👍

tvdeyen:
We decided to not back port this into 1.4 because of our release policy.

PR #1749, opened 2017-02-27
update paperclip dependency

brchristian:
`paperclip` 4.x depends on `aws-sdk` 1.x, which is being deprecated.

This updates the dependency to `paperclip` 5.x, which depends on the more recent `aws-sdk` 2.x.

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'highline', '~> 1.7' # Necessary for the install generator
   s.add_dependency 'kaminari', '>= 0.17', '< 2.0'
   s.add_dependency 'monetize', '~> 1.1'
-  s.add_dependency 'paperclip', '~> 4.2'
+  s.add_dependency 'paperclip', ['>= 4.2', '< 6']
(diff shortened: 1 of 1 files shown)


jhawthorn:
It it possible to support both? I imagine some users will still have some random gems still requiring `aws-sdk 1.x`

brchristian:
@jhawthorn Great point. Would you suggest simply removing the version requirement altogether?

jhawthorn:
It's good for us to be pessimistic about supporting future major versions. They'll bump these versions only when they have a major change, so it's best to double check each time there is a new one that we can still use it.

What I'd like best here is:
```
s.add_dependency 'paperclip', ['=> 4.2', '< 6']
```

brchristian:
@jhawthorn That sounds great. Done & done.

jhawthorn:
Oops I had typod my suggestion (`=>` vs `>=`). Pushed up a fix (hope that's okay).

brchristian:
@jhawthorn No prob! But it looks like something weird got into the changelog by accident. Let’s just take that out before we merge.

jhawthorn:
> @jhawthorn No prob! But it looks like something weird got into the changelog by accident. Let’s just take that out before we merge.

Whoops

PR #1750, opened 2017-02-28
Add default lastname in address factory

vfonic:
* Allows easier creation of factories where `lastname` is used in specs

### DIFF
--- a/core/lib/spree/testing_support/factories/address_factory.rb
+++ b/core/lib/spree/testing_support/factories/address_factory.rb
@@ -11,7 +11,7 @@ FactoryGirl.define do
     end
 
     firstname 'John'
-    lastname nil
+    lastname 'Doe'
     company 'Company'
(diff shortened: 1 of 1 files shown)


vfonic:
I need `Address.lastname` to be filled in in my specs and I think the proper way would be to add it here in the solidus codebase. Is there any specific known reason why would the `lastname` be `nil` in the factory?

jhawthorn:
I think the intention is to have `lastname` `nil` it because it is optional. We generally want to have factories create the absolute minimal object (though we don't always succeed in doing this).

Since requiring a lastname is specific to your application, you might consider using [FactoryGirl.modify](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#modifying-factories)

vfonic:
That makes sense. I'll modify the factory in my code. 

Thanks for the explanation!

PR #1751, opened 2017-03-01
Fix issue where OrderInventory creates superfluous InventoryUnits

jhawthorn:
This fixes an issue where OrderInventory would miscount the number of existing inventory_units on a line item and create extra records.

InventoryUnits exist through the line_item.inventory_units, shipment.inventory_units, and order.inventory_units associations. Because of this, their associations can't be kept up to date in memory.

OrderInventory was counting inventory using inventory_units.size, which would return the in-memory count if the line_items.inventory_units association was loaded.

This issue was exposed by adding to the same line item multiple times without reloading the order. The line_item.inventory_units association would be become `loaded?`, because ActiveRecord's `#size` [marks a record as loaded](https://github.com/rails/rails/blob/5-0-stable/activerecord/lib/active_record/associations/has_many_association.rb#L82-L85) if the count returns 0 (TIL!). On the next pass it would see the wrong size because the association was out of date, and add two inventory units instead of one.

This avoids the issue by counting inventory using inventory_units.count, which doesn't consider the in-memory association.

### DIFF
--- a/core/app/models/spree/order_inventory.rb
+++ b/core/app/models/spree/order_inventory.rb
@@ -18,13 +18,13 @@ module Spree
     def verify(shipment = nil)
       if order.completed? || shipment.present?
 
-        if inventory_units.size < line_item.quantity
-          quantity = line_item.quantity - inventory_units.size
-
-          shipment = determine_target_shipment unless shipment
-          add_to_shipment(shipment, quantity)
-        elsif inventory_units.size > line_item.quantity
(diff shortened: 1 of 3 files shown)


jhawthorn:
Fixed a second issue while investigating this. Previously, when the inventory count reached `0`, it would call `shipment.destroy`, which would leave the shipment in the in-memory `order.shipments` association.

Issue #1752, opened 2017-03-02
Overeager parameter filtering


jhawthorn:
We set `app.config.filter_parameters` to filter out `number` and `verification_value`, because we absolutely don't want to log user's CreditCard numbers or CVVs.

However this also filters out ANY parameter with `number` in the name. For example `"original_shipment_number"=>"[FILTERED]"`

jhawthorn:
Fixed by #1755

PR #1755, opened 2017-03-03
Fix sensitive params filtering

kennyadsl:
- Use a regular expression to filter only parameters that start and end
  with sensitive keywords.
- Remove useless assignment of filter_parameters in backend since it's
  already defined in core engine.

ref #1752 

Not sure if it makes sense to add some specs here since we are just using the a rails configuration. I'm open to suggestions anyway.

### DIFF
--- a/backend/lib/spree/backend/engine.rb
+++ b/backend/lib/spree/backend/engine.rb
@@ -6,11 +6,6 @@ module Spree
       initializer "spree.backend.environment", before: :load_config_initializers do |_app|
         Spree::Backend::Config = Spree::BackendConfiguration.new
       end
-
-      # filter sensitive information during logging
-      initializer "spree.params.filter" do |app|
-        app.config.filter_parameters += [:password, :password_confirmation, :number]
-      end
     end
   end
 end
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -104,13 +104,14 @@ module Spree
         ]
       end
 
-      # filter sensitive information during logging
+      # Filter sensitive information during logging
       initializer "spree.params.filter", before: :load_config_initializers do |app|
         app.config.filter_parameters += [
-          :password,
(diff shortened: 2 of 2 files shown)


kennyadsl:
@jhawthorn I initially started writing something like this but wasn't sure about where is the best place to put it and if it's really useful:

```ruby
# core/spec/controllers/filtered_parameters_spec.rb

require 'spec_helper'

RSpec.describe ApplicationController, type: :controller do
  controller do
    def index
      head :ok
    end
  end

  describe "Passing sensitive parameters" do
    subject { request.filtered_parameters }

    it "filters correctly" do
      get :index, params: {
        password: 'secret password',
        password_confirmation: 'secret password confirmation',
        number: 'secret credit card number',
        verification_value: 'secret verification value',
        name: 'something not secret',
        number_eq: 'R123456789',
        payment_number_eq: '123',
        answer_number: '42'
      }

      expect(subject["password"]).to eq '[FILTERED]'
      expect(subject["password_confirmation"]).to eq '[FILTERED]'
      expect(subject["number"]).to eq '[FILTERED]'
      expect(subject["verification_value"]).to eq '[FILTERED]'
      expect(subject["verification_value"]).to eq '[FILTERED]'

      expect(subject["name"]).not_to eq '[FILTERED]'
      expect(subject["number_eq"]).not_to eq '[FILTERED]'
      expect(subject["number_eq"]).not_to eq '[FILTERED]'
      expect(subject["answer_number"]).not_to eq '[FILTERED]'
    end
  end
end
```

I used an anonymous controller that inherits from ApplicationController since this test is not related to a specific controller; the configuration will be valid for the whole application. 

My concern is that, if one day we'll change `filtered_parameters` configuration in frontend or backend engines this test will still be green. A possible solution could be duplicate this test in all engines that need parameters filtering. What do you think about it?

jhawthorn:
Thanks so much @kennyadsl

PR #1757, opened 2017-03-06
Simplified cart link url

bofrede:
There is no need for the link to include protocol and hostname.

Scenario: A setup with a reverse proxy server in front the the application server.
The Application server runs HTTP. The reverse proxy server runs HTTPS.
When the reverse proxy server serves it as https, it will not load the XHR from a http: location.


jhawthorn:
@bofrede you may want to make sure your reverse proxy sends `X-Forwarded-Proto`, which rails should understand.

PR #1758, opened 2017-03-07
Fit logo to available space by default

brchristian:
I think it would make sense to fit the backend logo to the available div by default, instead of requiring stores to supply appropriately-sized images.


jhawthorn:
Thank you

PR #1759, opened 2017-03-07
Use sprockets 3/4 style manifest.js

jhawthorn:
Sprockets 4 is coming along and is currently `4.0.0.beta4`. They have a nice [upgrade doc](https://github.com/rails/sprockets/blob/master/UPGRADING.md), and some really great features (source maps, built-in babel+es6).

One issue is that sprockets 4 doesn't seem to support "wildcards" (like `spree/backend/all*`) in the `config.assets.precompile` list. It wasn't obvious, but it's clear this removal was intentional. The behaviour was listed in sprockets 3.x as deprecated and was in [legacy.rb](https://github.com/rails/sprockets/blob/3.x/lib/sprockets/legacy.rb).

The sprockets upgrade notes push for using a `app/assets/config/manifest.js` instead of loading files into the `precompile array`, which seems reasonable. We can do the same, but we need a unique name so we don't collide with 

In lieu of wildcards, we can use `link_tree`  in most cases. This handles images cleanly.

The obvious ugliness is that we need to explicitly reference all our select2 locale files. I don't think there's a way around this while still using the `select2-rails` gem. We can't use `link_tree` because the relative path is unknown (in a gem). One solution would be to embed `select2-rails` in the `vendor/assets` directory (which we might want to do anyways).


jhawthorn:
I'll vendor select2 in a separate PR

PR #1760, opened 2017-03-08
Use an icon as missing image placeholder

jhawthorn:
This displays a nicer placeholder for representing a missing image on products or variants. It uses a font icon instead of linking to a PNG asset.

This has the advantage of being i18n friendly (the previous PNG was english), and is slightly more convenient to use technically (doesn't require sprockets resolving the image path). It also looks nicer.

Before:

After:

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/templates/orders/line_item.hbs
+++ b/backend/app/assets/javascripts/spree/backend/templates/orders/line_item.hbs
@@ -4,7 +4,11 @@
   </td>
 {{else}}
   <td class="line-item-image">
-    <img src="{{ image.mini_url }}">
+    {{#if image }}
+      <img src="{{ image.mini_url }}">
+    {{else}}
+      <span class="image-placeholder mini"></span>
+    {{/if}}
   </td>
   <td class="line-item-name">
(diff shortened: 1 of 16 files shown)


backend/app/assets/javascripts/spree/backend/templates/orders/line_item.hbs
```diff
@@ -4,7 +4,11 @@
   </td>
 {{else}}
   <td class="line-item-image">
-    <img src="{{ image.mini_url }}">
+    {{#if image }}
+      <img src="{{ image.mini_url }}">
+    {{else}}
+      <span class="image-placeholder mini"></span>
+    {{/if}}
```
tvdeyen: Could this be a handlebars partial?

jhawthorn: Not sure about that because we need to do the lookup of `mini_url`/`small_url`, but maybe a helper. I'll investigate.

jhawthorn: It can be done, but I think it warrants further discussion. I will open a follow-up PR.

PR #1761, opened 2017-03-08
Methods for Spree::Taxon for all products/variants from descendants

dgra:
This adds 2 primary methods to Spree::Taxon to allow users to retreive
all products and all variants that are under the taxon instance. Helpful for
a taxon page to show an overview of everything that is under that
taxon. Also, included is a shortcut method for
`taxon.descendants.pluck(:id)`.

### DIFF
--- a/core/app/models/spree/taxon.rb
+++ b/core/app/models/spree/taxon.rb
@@ -87,6 +87,19 @@ module Spree
       products.not_deleted.available
     end
 
+    # @return [ActiveRecord::Relation<Spree::Product>] all self and descendant products
+    def all_products
+      scope = Product.joins(:taxons)
+      scope.where(
+        spree_taxons: { id: self_and_descendants.select(:id) }
+      )
+    end
+
+    # @return [ActiveRecord::Relation<Spree::Variant>] all self and descendant variants, including master variants.
+    def all_variants
+      Variant.where(product_id: all_products.select(:id))
+    end
+
     # @return [String] this taxon's ancestors names followed by its own name,
     #   separated by arrows
     def pretty_name
--- a/core/spec/models/spree/taxon_spec.rb
+++ b/core/spec/models/spree/taxon_spec.rb
@@ -135,4 +135,39 @@ RSpec.describe Spree::Taxon, type: :model do
       taxonomy.root.children.unscoped.where(name: "Some name").first_or_create
     end
   end
+
+  context 'leaves of the taxon tree' do
+    let(:taxonomy) { create(:taxonomy, name: 't') }
+    let(:root) { taxonomy.root }
+    let(:taxon) { create(:taxon, name: 't1', taxonomy: taxonomy, parent: root) }
+    let(:child) { create(:taxon, name: 'child taxon', taxonomy: taxonomy, parent: taxon) }
+    let(:grandchild) { create(:taxon, name: 'grandchild taxon', taxonomy: taxonomy, parent: child) }
+    let(:product1) { create(:product) }
+    let(:product2) { create(:product) }
+    let(:product3) { create(:product) }
+    before do
+      product1.taxons << taxon
+      product2.taxons << child
+      product3.taxons << grandchild
+      taxon.reload
+
+      [product1, product2, product3].each { |p| 2.times.each { create(:variant, product: p) } }
+    end
+
+    describe '#all_products' do
+      it 'returns all descendant products' do
+        products = taxon.all_products
+        expect(products.count).to eq(3)
+        expect(products).to match_array([product1, product2, product3])
+      end
+    end
+
+    describe '#all_variants' do
+      it 'returns all descendant variants' do
+        variants = taxon.all_variants
+        expect(variants.count).to eq(9)
+        expect(variants).to match_array([product1, product2, product3].map{ |p| p.variants_including_master }.flatten)
+      end
+    end
+  end
 end


core/app/models/spree/taxon.rb
```diff
@@ -79,12 +79,32 @@ def to_param
...
       products.not_deleted.available
     end
 
+    # @return [ActiveRecord::Relation<Spree::Product>] all self and descendant products
+    def all_products
+      Product.joins(:taxons).where(spree_taxons: { id: descendant_ids.push(id) })
+    end
+
+    # @return [ActiveRecord::Relation<Spree::Variant>] all self and descendant variants, including master variants.
+    def all_variants
```
cbrunsdon: This can get cleaned up to 
```
Variant.where(product_id: all_products.pluck(:id))
```

core/app/models/spree/taxon.rb
```diff
@@ -79,12 +79,32 @@ def to_param
       permalink
     end
 
+    # @return [Array<Integer>] the ids of descendant taxons.
+    def descendant_ids
```
cbrunsdon: I don't think this is really worth being a public method, its more likely anyone that wants this will not know it exists and call `descendants.pluck(:id)`

core/app/models/spree/taxon.rb
dgra: @cbrunsdon I will change that, my original thinking for the other way was to save database calls but it only saves us one db call. Thank you for your feedback.

core/app/models/spree/taxon.rb
```diff
@@ -85,6 +85,16 @@ def active_products
       products.not_deleted.available
     end
 
+    # @return [ActiveRecord::Relation<Spree::Product>] all self and descendant products
+    def all_products
+      Product.joins(:taxons).where(spree_taxons: { id: descendant_ids.push(id) })
```
jhawthorn: It would be better to use `descendant_ids + [id]`.

Push would append to the array `descendant_ids` returns which we want to avoid (I don't know whether or not that would cause a real issue, but I can't know without reading its source, so we want to avoid it)

dgra: Good eye. Changing now. Thanks.

dgra:
@jhawthorn @cbrunsdon I just wanted to checkup on the state of this PR. Thank you guys.

dgra:
@jhawthorn @cbrunsdon Is there any update on this PR?

core/app/models/spree/taxon.rb
```diff
@@ -85,6 +85,16 @@ def active_products
       products.not_deleted.available
     end
 
+    # @return [ActiveRecord::Relation<Spree::Product>] all self and descendant products
+    def all_products
+      Product.joins(:taxons).where(spree_taxons: { id: descendant_ids + [id] })
```
gmacdougall: I have some query optimizations I think would be suitable here. Any time we use pluck, we take back the identifiers and put them in a ruby array which means a round trip to the database.

Currently `all_products` is required to do 2 queries, and `all_variants` is required to do 3, but we can reduce that to one each pretty easily.

Example of current behaviour:

```
[1] pry(main)> Spree::Taxon.first.all_variants.count
  Spree::Taxon Load (0.1ms)  SELECT  "spree_taxons".* FROM "spree_taxons" ORDER BY "spree_taxons"."id" ASC LIMIT ?  [["LIMIT", 1]]
   (0.2ms)  SELECT "spree_taxons"."id" FROM "spree_taxons" WHERE ("spree_taxons"."lft" >= 1) AND ("spree_taxons"."lft" < 12) AND ("spree_taxons"."id" != 1) ORDER BY "spree_taxons"."lft"
   (0.2ms)  SELECT "spree_products"."id" FROM "spree_products" INNER JOIN "spree_products_taxons" ON "spree_products_taxons"."product_id" = "spree_products"."id" INNER JOIN "spree_taxons" ON "spree_taxons"."id" = "spree_products_taxons"."taxon_id" WHERE "spree_products"."deleted_at" IS NULL AND "spree_taxons"."id" IN (3, 4, 5, 6, 7, 1)
   (0.1ms)  SELECT COUNT(*) FROM "spree_variants" WHERE "spree_variants"."deleted_at" IS NULL AND "spree_variants"."product_id" IN (1, 2, 8, 9, 4, 3, 5, 7, 6)
```

But if we try this:
```Ruby
    # @return [ActiveRecord::Relation<Spree::Product>] all self and descendant
    #   products
    def all_products
      scope = Product.joins(:taxons)
      scope.where(
        spree_taxons: { id: descendants.select(:id) }
      ).or(scope.where(spree_taxons: { id: id } ))
    end

    # @return [ActiveRecord::Relation<Spree::Variant>] all self and descendant
    #   variants, including master variants.
    def all_variants
      Variant.where(product_id: all_products.select(:id))
    end
```

Everything can be done in a single subquery:

```[3] pry(main)> Spree::Taxon.first.all_variants.count
  Spree::Taxon Load (0.2ms)  SELECT  "spree_taxons".* FROM "spree_taxons" ORDER BY "spree_taxons"."id" ASC LIMIT ?  [["LIMIT", 1]]
   (0.3ms)  SELECT COUNT(*) FROM "spree_variants" WHERE "spree_variants"."deleted_at" IS NULL AND "spree_variants"."product_id" IN (SELECT "spree_products"."id" FROM "spree_products" INNER JOIN "spree_products_taxons" ON "spree_products_taxons"."product_id" = "spree_products"."id" INNER JOIN "spree_taxons" ON "spree_taxons"."id" = "spree_products_taxons"."taxon_id" WHERE ("spree_products"."deleted_at" IS NULL AND "spree_taxons"."id" IN (SELECT "spree_taxons"."id" FROM "spree_taxons" WHERE ("spree_taxons"."lft" >= 1) AND ("spree_taxons"."lft" < 12) AND ("spree_taxons"."id" != 1) ORDER BY "spree_taxons"."lft") OR "spree_products"."deleted_at" IS NULL AND "spree_taxons"."id" = ?))  [["id", 1]]
=> 19
```

This means only one round trip to the DB, less work done in Ruby and hopefully happy developers!

dgra:
@gmacdougall No problem and thank you for looking at this. I knew it had a few db calls but I should have done a better job of reducing those. Thank you again for looking at this and reducing those calls.

core/app/models/spree/taxon.rb
```diff
@@ -124,5 +137,10 @@ def touch_ancestors_and_taxonomy
       # Have taxonomy touch happen in #touch_ancestors_and_taxonomy rather than association option in order for imports to override.
       taxonomy.try!(:touch)
     end
+
+    # @return [Array<Integer>] the ids of descendant taxons.
+    def descendant_ids
```
gmacdougall: This method is no longer used and can be removed.

dgra:
@gmacdougall You are right. I removed that method and I will rebase these commits into a single commit if/when this gets approved.

gmacdougall:
Looks reasonable to me. Thanks for addressing the feedback.

Rebase away!

dgra:
Thank you so much for providing feedback on this!

dgra:
@gmacdougall Just checking up on the status of this PR. Any update?

core/app/models/spree/taxon.rb
```diff
@@ -85,6 +85,19 @@ def active_products
       products.not_deleted.available
     end
 
+    # @return [ActiveRecord::Relation<Spree::Product>] all self and descendant products
+    def all_products
+      scope = Product.joins(:taxons)
+      scope.where(
+        spree_taxons: { id: descendants.select(:id) }
+      ).or(scope.where(spree_taxons: { id: id } ))
```
jhawthorn: This could probably be simpler as `where(spree_taxons: { id: self_and_descendants.ids } )`

jhawthorn:
I modified `all_products` to be a little cleaner using `self_and_descendants`.

Thanks for the change :+1:

gmacdougall:
Thanks for the work on this Dustin!  Sorry for the slow response.

PR #1762, opened 2017-03-09
Convert customer details page to backbone

jhawthorn:
This PR aims to replace the jQuery soup of [checkouts/edit.js](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/checkouts/edit.js) and [address_states.js](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/address_states.js) with backbone models and views.

There were a number of issues with the existing code:

* A number of race conditions were possible (though somewhat unlikely because the states API is very fast). For example switching countries rapidly could result in the wrong states being shown
* "Guest checkout" toggle didn't work correctly. If it was switched from "no" to "yes" and then back, it would still be submitted to the server as a guest checkout (no associated user)
* `checkouts.js` was callback hell. Worst case was setting the `state_id`, which had to be done [inside a select2 callback, and a callback from an ajax request inside of update_state](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/checkouts/edit.js#L64)

The conversion to backbone added a fair number of lines, which I feel I should justify. From about 135 lines between `edit.js` and `address_states.js` to about 250 lines between the four new backbone views.

This PR adds two way bindings between all the form inputs on this page. This allows our top level view to be oblivious to the details of the subview (in particular the awkwardness of state selection). Specifically I'd like to point to [the existing code when selecting a customer](https://github.com/solidusio/solidus/blob/80d63c2bcfa2f5904184e43ed1b5ba3fc9c74566/backend/app/assets/javascripts/spree/backend/checkouts/edit.js#L44-L68) vs the [new code](https://github.com/solidusio/solidus/blob/efc96579d5f326c8fd3acab259fe98b2f984e828/backend/app/assets/javascripts/spree/backend/views/order/customer_details.js#L48-L52)

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/state_select.js
+++ b/backend/app/assets/javascripts/spree/backend/views/state_select.js
@@ -30,25 +30,19 @@ Spree.Views.StateSelect = Backbone.View.extend({
 
   updateStates: function() {
     this.stopListening(this.states);
-    this.pending = true;
     var country_id = this.model.get("country_id");
     if (country_id) {
-      this.states = new Spree.Collections.States([], {country_id: country_id});
-      this.listenTo(this.states, "reset", this.onSync);
-      this.states.fetch({reset: true});
+      this.states = Spree.Views.StateSelect.stateCache(country_id);
+      this.listenTo(this.states, "sync", this.render);
+      this.render();
     }
-  },
-
-  onSync: function() {
-    this.pending = false;
-    this.render();
   },
 
   render: function() {
     this.$state_select.empty().select2("destroy").hide();
     this.$state_input.hide();
 
-    if (this.pending) {
+    if (!this.states.fetched) {
       this.$state_select.show().select2().select2("disable");
     } else if (this.states.length) {
       var $state_select = this.$state_select;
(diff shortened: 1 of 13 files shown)


jasonfb:
👍 for backbone; we use it widely (we use Mariontte/Backbone) and love it. in particular, it boots extremely fast and its source code is accessible (readable).

backend/app/assets/javascripts/spree/backend/views/state_select.js
```diff
@@ -0,0 +1,71 @@
...
+    this.$state_select.empty().select2("destroy").hide();
+    this.$state_input.hide();
+
+    if (!this.states.fetched) {
+      this.$state_select.show().select2().select2("disable");
+    } else if (this.states.length) {
+      var $state_select = this.$state_select;
+      this.states.each(function(state) {
+        $state_select.append(
+          $('<option>').prop('value', state.id).text(state.get("name"))
```
tvdeyen: Do we need a check for already existing states here? Or is the select always empty at this point?

jhawthorn: Yep, we empty the `<select>` on line 42.

tvdeyen: 👍

PR #1764, opened 2017-03-09
Add image partial to handlebars

jhawthorn:
This adds a Handlebars partial for rendering images or their placeholder. As suggested by @tvdeyen in #1760

To be able to index the correctly sized image url, I needed to use the handlebars `lookup` helper, which allows looking up by a dynamic attribute, as well as introduce a new `concat` helper to get the correctly named key (ex. `mini_url` from `"mini"`). A little hacky, but it works well and isn't noticeable to anyone calling the partial.

At the same time I've moved the code to dig the image data out of the variant JSON into handlebars, which is a bit simpler and nicer.


tvdeyen:
Who ever finds the verb can keep it 😅

PR #1765, opened 2017-03-13
Migrate all reusable credit card to wallet sources

jhawthorn:
Follow up to #1707

The behaviour prior to adding the Wallet was to allow reuse of any `CreditCard` which had a payment profile. However the migration was only migrating `CreditCard`s marked `default`.

This now migrates all credit cards with an associated `gateway_customer_profile_id`.


core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
```diff
@@ -1,15 +1,23 @@
...
 
     credit_cards.find_each do |credit_card|
-      Spree::WalletPaymentSource.create!(
+      Spree::WalletPaymentSource.find_or_create_by!(
         user_id: credit_card.user_id,
-        payment_source: credit_card,
-        default: credit_card.read_attribute(:default),
-      )
+        payment_source: credit_card
+      ) do |wallet_source|
```
tvdeyen: TIL 👍

core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
```diff
@@ -1,15 +1,23 @@
...
+    self.table_name = 'spree_credit_cards'
+  end
+  class WalletPaymentSource < ActiveRecord::Base
+    self.table_name = 'spree_wallet_payment_sources'
+  end
+
   def up
     credit_cards = Spree::CreditCard.
-      where(default: true).
+      where("gateway_customer_profile_id IS NOT NULL").
```
tvdeyen:
```ruby
where.not(gateway_customer_profile_id: nil)
```

core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
```diff
@@ -1,15 +1,23 @@
...
+class MigrateCreditCardsToWalletPaymentSources < ActiveRecord::Migration[4.2]
+  class CreditCard < ActiveRecord::Base
+    self.table_name = 'spree_credit_cards'
+  end
+  class WalletPaymentSource < ActiveRecord::Base
+    self.table_name = 'spree_wallet_payment_sources'
+  end
+
   def up
     credit_cards = Spree::CreditCard.
```
jordan-brough: Did you intend to use `CreditCard` here instead of `Spree::CreditCard`?

core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
```diff
@@ -1,15 +1,23 @@
...
-      where(default: true).
+      where("gateway_customer_profile_id IS NOT NULL").
       where.not(user_id: nil)
 
     credit_cards.find_each do |credit_card|
-      Spree::WalletPaymentSource.create!(
+      Spree::WalletPaymentSource.find_or_create_by!(
```
jordan-brough: `WalletPaymentSource` instead of `Spree::WalletPaymentSource`?

core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
```diff
@@ -1,15 +1,23 @@
...
         user_id: credit_card.user_id,
-        payment_source: credit_card,
-        default: credit_card.read_attribute(:default),
-      )
+        payment_source: credit_card
+      ) do |wallet_source|
+        wallet_source.default = credit_card[:default]
```
jordan-brough: why is this in a block instead of alongside the other attributes?

jordan-brough: if you use your in-migration classes you can just do `credit_card.default` right?

core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
```diff
@@ -19,7 +27,7 @@ def down
     add_column :spree_credit_cards, :default, :boolean, default: false, null: false
 
     wallet_payment_sources = Spree::WalletPaymentSource.
-      where(default: true, source_type: 'Spree::CreditCard').
+      where(default: true, payment_source_type: 'Spree::CreditCard').
       includes(:source)
```
jordan-brough: we'll need some tweaking for the `includes` to work with the in-migration classes right?

core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
```diff
@@ -1,15 +1,23 @@
...
-        payment_source: credit_card,
-        default: credit_card.read_attribute(:default),
-      )
+        payment_source: credit_card
+      ) do |wallet_source|
+        wallet_source.default = credit_card[:default]
+      end
     end
 
     remove_column :spree_credit_cards, :default, :boolean
```
jordan-brough: I wonder if we should hold off on actually dropping this column for another release.
Looking at it again right now I feel nervous about dropping a payment-related column like that.  Holding off would make the "down" migration simpler also.
Thoughts?

jhawthorn: :+1: As long as we provide deprecations.

core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
jhawthorn: Passing a block to `find_or_create_by` behaves the same as passing a block to `create` if the record doesn't already exist.

We want to do this here, because we absolutely only want to have one `WalletPaymentSource` per credit card, the block specifies the values if one doesn't already exist. This makes the up migration idempotent, which we want since the down migration doesn't remove `WalletPaymentSource`s.

jordan-brough: 👍 I stared at this at least 10 times and still managed to miss the fact that you'd changed it from `create` to `find_or_created_by` :)

core/db/migrate/20160420181916_migrate_credit_cards_to_wallet_payment_sources.rb
jordan-brough: The `Spree::CreditCard#default` method does have deprecations, but if someone was doing a `Spree::CreditCard.where(default: true)` I'm not sure how we could add a deprecation for that.  Perhaps we could rename the column to `deprecated_default` instead of dropping it?  Or just go ahead and drop it?

jhawthorn:
@jordan-brough I think I have addressed your feedback. It no longer removes the `spree_credit_cards.default` column, which eliminates the need for the down migrations's behaviour. It now only uses the `CreditCard` and `WalletPaymentSource` defined in the migration.

PR #1767, opened 2017-03-13
Remove StockMovements#new

jhawthorn:
I broke this page [over a year ago](https://github.com/solidusio/solidus/commit/0f5c0d0d30135c2ca40286592415054db4f2df56) (sorry). There is a javascript error when trying to use the stock item select. Nobody noticed so it should probably be safe to remove.

This page previously provided an interface to manually create a stock movement. I think this can more easily be done by editing stock from our stock management page.


jhawthorn:
I don't think this will be missed. It was introduced ages before we had our current stock editing interface, which can do the same task this does, but with much clearer UI.

If I'm mistaken, please reach out via GH issue or Slack.

PR #1770, opened 2017-03-14
Remove order association from inventory units

mamhoff:
Inventory units link line items to shipments. Both of these have an
order. Let's make this a has_many association, through :shipments, which
are required.

### DIFF
--- a/core/app/models/spree/customer_return.rb
+++ b/core/app/models/spree/customer_return.rb
@@ -61,7 +61,7 @@ module Spree
     end
 
     def return_items_belong_to_same_order
-      if return_items.select{ |return_item| return_item.inventory_unit.order_id != order_id }.any?
+      if return_items.select{ |return_item| return_item.inventory_unit.order != order }.any?
         errors.add(:base, Spree.t(:return_items_cannot_be_associated_with_multiple_orders))
       end
     end
--- a/core/app/models/spree/customer_return.rb
+++ b/core/app/models/spree/customer_return.rb
@@ -61,7 +61,7 @@ module Spree
     end
 
     def return_items_belong_to_same_order
-      if return_items.select{ |return_item| return_item.inventory_unit.order != order }.any?
+      if return_items.select{ |return_item| return_item.inventory_unit.order_id != order_id }.any?
         errors.add(:base, Spree.t(:return_items_cannot_be_associated_with_multiple_orders))
       end
     end
--- a/core/app/models/spree/order_cancellations.rb
+++ b/core/app/models/spree/order_cancellations.rb
@@ -25,7 +25,7 @@ class Spree::OrderCancellations
   #
   # @return [Array<UnitCancel>] the units that have been canceled due to short shipping
   def short_ship(inventory_units, whodunnit:nil)
-    if inventory_units.map(&:order).uniq != [@order]
+    if inventory_units.map(&:order_id).uniq != [@order.id]
       raise ArgumentError, "Not all inventory units belong to this order"
     end
 
--- a/core/app/models/spree/order_cancellations.rb
+++ b/core/app/models/spree/order_cancellations.rb
@@ -25,7 +25,7 @@ class Spree::OrderCancellations
   #
   # @return [Array<UnitCancel>] the units that have been canceled due to short shipping
   def short_ship(inventory_units, whodunnit:nil)
-    if inventory_units.map(&:order_id).uniq != [@order.id]
+    if inventory_units.map(&:order).uniq != [@order]
       raise ArgumentError, "Not all inventory units belong to this order"
     end
 
--- a/core/app/models/spree/reimbursement.rb
+++ b/core/app/models/spree/reimbursement.rb
@@ -154,7 +154,7 @@ module Spree
     end
 
     def validate_return_items_belong_to_same_order
-      if return_items.any? { |ri| ri.inventory_unit.order_id != order_id }
+      if return_items.any? { |ri| ri.inventory_unit.order != order }
         errors.add(:base, :return_items_order_id_does_not_match)
       end
     end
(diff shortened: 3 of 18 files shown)


cbrunsdon:
This relationship exists in the complicated it way is now because inventory units can be associated to multiple shipments through... returns? or exchanges?

You can see all the failures, we do in fact have behavior that relies on this (for better or worse), and these will not be easy changes to make.

mamhoff:
@cbrunsdon The spec suite doesn't catch any complex behaviour when removing the column. :)

core/app/models/spree/shipping_manifest.rb
```diff
@@ -7,7 +7,7 @@ def initialize(inventory_units:)
 
   def for_order(order)
     Spree::ShippingManifest.new(
-      inventory_units: @inventory_units.select { |iu| iu.order_id == order.id }
+      inventory_units: @inventory_units.select { |iu| iu.shipment.order_id == order.id }
```
tvdeyen: I would prefer to have 

```ruby
inventory_units: @inventory_units.select { |iu| iu.order == order }
```

here

core/db/migrate/20170319191942_remove_order_id_from_inventory_units.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def up
+    if InventoryUnit.
+        joins(:shipment).
+        where.not(
+          'spree_inventory_units.order_id = spree_shipments.order_id'
+        ).exists?
+      raise InconsistentInventoryUnitError, "You have inventory units with inconsistent order references. Please fix those before running this migration"
+    end
```
tvdeyen: 👍

core/app/models/spree/shipping_manifest.rb
jhawthorn: I'm slightly concerned that this change may introduce `N+1` queries (and tvdeyen's suggestion `2N+1`)

jhawthorn:
I think @cbrunsdon's concern is with code that has been extracted into the [solidus_expedited_exchanges](https://github.com/solidusio-contrib/solidus_expedited_exchanges) gem, which moves a shipment from one order to another in order to charge for it (yech!). I think it will be fine, as that code uses shipment creation time vs order creation time (yech again!) to detect an exchange. In any case, this shouldn't be a blocker.

Other than concern about additional queries 👍 from me.

mamhoff:
@jhawthorn I think you might be right here. What about having something like 
`delegate :order_id to: :shipment` in order to have concise usage of inventory units? @tvdeyen concern is a law-of-demeter concern where he's like "the manifest shouldn't know about the order's id", and that could "fix" both of your concerns (as well as possibly making the changeset a lot smaller. 

The PR at the moment always checks for the full `order`, which does introduce more objects being loaded.

core/app/models/spree/customer_return.rb
```diff
@@ -61,7 +61,7 @@ def generate_number
     end
 
     def return_items_belong_to_same_order
-      if return_items.select{ |return_item| return_item.inventory_unit.order_id != order_id }.any?
+      if return_items.select{ |return_item| return_item.inventory_unit.order != order }.any?
```
tvdeyen: Can we use?

```ruby
return_items.joins(inventory_unit: :shipment).where.not(Spree::Shipment.table_name => {order_id: order_id}).exists?
```

should be much faster and avoid `N+1`

core/app/models/spree/order_cancellations.rb
```diff
@@ -25,7 +25,7 @@ def initialize(order)
   #
   # @return [Array<UnitCancel>] the units that have been canceled due to short shipping
   def short_ship(inventory_units, whodunnit:nil)
-    if inventory_units.map(&:order_id).uniq != [@order.id]
+    if inventory_units.map(&:order).uniq != [@order]
       raise ArgumentError, "Not all inventory units belong to this order"
```
tvdeyen: Do we really need this check? Maybe tackle in a separate PR?

core/app/models/spree/reimbursement.rb
```diff
@@ -154,7 +154,7 @@ def generate_number
     end
 
     def validate_return_items_belong_to_same_order
-      if return_items.any? { |ri| ri.inventory_unit.order_id != order_id }
+      if return_items.any? { |ri| ri.inventory_unit.order != order }
```
tvdeyen: Can we use?

```ruby
return_items.joins(inventory_unit: :shipment).where.not(Spree::Shipment.table_name => {order_id: order_id}).exists?
```

should also be much faster and avoid `N+1`

core/app/models/spree/return_item.rb
```diff
@@ -258,7 +258,7 @@ def check_unexchange
     def belongs_to_same_customer_order
       return unless customer_return && inventory_unit
 
-      if customer_return.order_id != inventory_unit.order_id
+      if customer_return.order != inventory_unit.order
         errors.add(:base, Spree.t(:return_items_cannot_be_associated_with_multiple_orders))
```
tvdeyen: Could also be removed in a separate PR IMO

core/app/models/spree/shipping_manifest.rb
```diff
@@ -7,7 +7,7 @@ def initialize(inventory_units:)
 
   def for_order(order)
```
tvdeyen: This whole method needs to get nuked, because core doesn't create cartons from different orders. Separate PR of course

core/app/models/spree/customer_return.rb
mamhoff: This does not work, as the object being validated might have been changed in-memory and the DB query won't catch those changes.

core/app/models/spree/shipping_manifest.rb
mamhoff: This method exists because of shipping providers packing things for different order into same cartons.

mamhoff:
Rebased on current master, and fixed for Rails 5. From my point of view, this is still a very sensible thing to do. Oh Inventory Units.

mamhoff:
@tvdeyen would you mind re-reviewing?

PR #1773, opened 2017-03-14
Credit card fixes related to Wallet updates

jordan-brough:
The deprecated methods didn't handle all cases.

### DIFF
--- a/core/app/models/spree/credit_card.rb
+++ b/core/app/models/spree/credit_card.rb
@@ -46,12 +46,15 @@ module Spree
 
     def default
       Spree::Deprecation.warn("CreditCard.default is deprecated. Please use user.wallet.default_wallet_payment_source instead.", caller)
-      user.wallet.default_wallet_payment_source.payment_source == self
+      return false if user.nil?
+      user.wallet.default_wallet_payment_source.try!(:payment_source) == self
     end
 
     def default=(set_as_default)
       Spree::Deprecation.warn("CreditCard.default= is deprecated. Please use user.wallet.default_wallet_payment_source= instead.", caller)
-      if set_as_default # setting this card as default
+      if user.nil?
+        raise "Cannot set 'default' on a credit card without a user"
+      elsif set_as_default # setting this card as default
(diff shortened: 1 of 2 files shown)


core/app/models/spree/credit_card.rb
```diff
@@ -46,12 +46,15 @@ def self.default
...
       Spree::Deprecation.warn("CreditCard.default is deprecated. Please use user.wallet.default_wallet_payment_source instead.", caller)
-      user.wallet.default_wallet_payment_source.payment_source == self
+      return false if user.nil?
+      user.wallet.default_wallet_payment_source.try!(:payment_source) == self
     end
 
     def default=(set_as_default)
       Spree::Deprecation.warn("CreditCard.default= is deprecated. Please use user.wallet.default_wallet_payment_source= instead.", caller)
-      if set_as_default # setting this card as default
+      if user.nil?
```
cbrunsdon: Does this work reasonably with the current default behavior and guest checkouts?

https://github.com/solidusio/solidus/blob/master/core/app/models/spree/wallet/default_payment_builder.rb#L14

Its tough to trace all the way through easily (with all the aliases and whatnot) but I think we'll get the default set to nil if there is no user? Though none of our specs seem to be failing...

jordan-brough: @cbrunsdon that code you referenced (and the `assign_default_credit_card` method that calls it) is a no-op when `order.user` is nil, right?  Is that what you're asking?

jordan-brough: Also the [AddPaymentSourcesToWallet](https://github.com/solidusio/solidus/blob/ec58304541ad1655818042be0ea2ecb5863c7211/core/app/models/spree/wallet/add_payment_sources_to_wallet.rb#L14) code is a no-op if `order.user` is nil.

jordan-brough: It looks like our permitted attrs prevent setting credit card defaults via the api.
The admin doesn't allow setting the `default` attribute either.

I'm not seeing any other code that tries to set the default on a credit card, other than the above items.

PR #1776, opened 2017-03-16
Upgrade to Bootstrap 4.0.0.alpha6

jhawthorn:
Previously we were on Bootstrap 4.0.0.alpha2. Changes can be seen in their [Release notes](https://github.com/twbs/bootstrap/releases)

Notable to us:
* `col-xs-3` et al. are gone. Replaced with `col-3`.
* Flexbox is on by default :+1:  (we had already turned on the optional flexbox support)
* Classes for breadcrumbs changed (easy fix)
* Some variables and mixins changed

I can't be 100% sure this didn't mess up any styling, but in a quick browse through the admin I didn't spot anything off.


jhawthorn:
Hmm we seem to be running into https://github.com/twbs/bootstrap/issues/21607 so we may have to put off this upgrade

jhawthorn:
I tried setting `enable-transitions:        false`, and it didn't help

jhawthorn:
Closing for #1816

PR #1777, opened 2017-03-16
Raise font-size

tvdeyen:
All our forms are nested in field sets, which have a font-size of 90%. Also our form field font sizes are reduced by 10%. With our base font size of 13px this leads to a 10.5px font size. This is definitely too small for modern screen sizes.

### Before

solidus 13px

### After

orders 2017-03-19 21-49-48

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
@@ -14,7 +14,6 @@ nav.menu {
         text-align: left;
         border: 1px solid transparent;
         font-weight: $font-weight-bold;
-        font-size: 90%;
       }
 
       &.active a {
--- a/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
@@ -7,7 +7,7 @@ input[type="time"],
 input[type="url"],
 input[type="number"],
 input[type="tel"],
-textarea, fieldset {
+textarea {
   padding: 7px 10px;
   border: 1px solid $color-txt-brd;
   border-radius: $border-radius;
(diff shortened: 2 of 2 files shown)


jhawthorn:
If we're tweaking font sizes, I think we should remove some strange font size rules. Otherwise we risk calibrating the base font size wrong because the elements we look at are at `font-size: 90%`

For example:
* [fieldsets are set to 90%](https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss#L15)
* [labels are set to 90%](https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss#L75)

Mandily:
I would love to see a plan to deal with text that doesn't fit. I suspect if this happens at all it would be in the table headers. I'd be ok with truncating it like we do the admin email in the left hand column until we get a better layout sorted out.

tvdeyen:
> Fixing these should make the fonts even larger at the same time, which I think aligns with your goals 😄 .

👍

> truncating it like we do the admin email in the left hand column

I don't think it's a good idea to truncate table headers. IMO they transport important information. But I may be wrong. I would like to widen the default container width from 960px to ~1080px instead.

tvdeyen:
Most ideally we would use the newly introduced full width layout everywhere, but this would cause adjustments to the search forms most of the tables have. Maybe introduce a max width container for the search form?

Mandily:
The full width layout is for the listing pages with tables, but if there was a small table in a mostly form page I had hypothesized that text may not fit as shown in the example below. In the case that the text overflowed the table header that it was in, I thought it better to truncate it. Now thinking about it further, the best solution would be to simple wrap the text. :)

screen shot 2017-03-17 at 9 24 52 am

tvdeyen:
This could work. Alternatively we could remove columns that hold secondary information on smaller viewports. Hard to figure out which one is the secondary information though.

tvdeyen:
@Mandily First step taken https://github.com/solidusio/solidus/pull/1782

tvdeyen:
@Mandily @jhawthorn I removed the change of the body font size and fixed the sizing in `fieldset` and the navigation instead (see updated comment on top). Thanks for pointing me to this, John!

Mandily:
👍

jhawthorn:
This is looking much better.

@tvdeyen was there supposed to be a change to `$body-font-size` as well? Currently our legacy styles are using `13px` and bootstrap is specifying `15px`

tvdeyen:
@jhawthorn yes, but before we do that we want to do some preparations with #1780 and #1786 

Otherwise raising the font size just won't look good.

PR #1779, opened 2017-03-17
Add missing indexes for Spree::Taxon lft and rgt columns

vfonic:
* These are used by `awesome_nested_set` to perform ordering and
  therefore indexes should be added to them


vfonic:
Bump!

Info straight from `awesome_nested_set` Readme:
https://github.com/collectiveidea/awesome_nested_set/blob/fbc84118a0ba495b934ea5cadbdcc56847ec2aa0/README.md#indexes

(perhaps we should add even more indexes as suggested?)

PR #1780, opened 2017-03-17
Deprecate sass variables

tvdeyen:
In preparation for #1777 we deprecate our basic typography and colour variables and use the Bootstrap equivalent variable instead.

By dog feeding our current defaults the look is still the same as before. In future versions we will remove our variables, so extension developers and shops should start to use the Bootstrap variables instead. The depreciation warning helps by the upgrade.

I also removed some duplicated style an variable definitions that were defined by Bootstrap and us.


mtomov:
Tedious one indeed!, super :+1:

PR #1782, opened 2017-03-17
Convert admin orders table into full width layout

tvdeyen:
The admin orders table is too narrow on modern screen sizes. There is plenty of space we could use to display the full information on the orders table.

When adapting the full width layout we get too wide columns on the search form, though. This is fixed by utilizing different column sizes for the viewports. 

The result is pretty good without moving around any html elements (what potentially could break admin overrides of extensions or shops).

### Before

before-full-hd

### After

after-full-hd

after-large

after-ipad-landscape

after-ipad-portrait

after-mobile-portrait

### DIFF
--- a/backend/app/views/spree/admin/orders/index.html.erb
+++ b/backend/app/views/spree/admin/orders/index.html.erb
@@ -1,3 +1,5 @@
+<% admin_layout "full-width" %>
+
 <% admin_breadcrumb(plural_resource_name(Spree::Order)) %>
 
 
@@ -15,7 +17,7 @@
   <div data-hook="admin_orders_index_search">
     <%= search_form_for [:admin, @search] do |f| %>
       <div class="row">
-        <div class="field-block col-xs-3">
+        <div class="field-block col-xs-12 col-md-6 col-lg-4 col-xl-3">
           <div class="date-range-filter field">
             <%= label_tag :q_created_at_gt, Spree.t(:date_range) %>
             <div class="date-range-fields">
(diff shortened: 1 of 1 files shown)


tvdeyen:
@Mandily please have a look at this.

Mandily:
I'm definitely for the full width on the table. 

The third and fourth screenshots start to look a little odd to me with form fields moving around and empty space being on the left side of a row. Is there a way to clean that up at all without breaking extensions and shops as you mentioned above?

And just to double check - is this replicable with the other listing pages as well?

Mandily:
To play devil's advocate - did you consider putting a max width on the search area so preserve the order/groupings of form elements a little better? 

This was my idea in #1290 but I could see too how using less space up top could be more desirable. It also leads to the question of how well are those fields really organized at this point anyway? I wanted to move them around and even change the way a couple worked (as you can see in the checkboxes) but I was never super clear on how that would affect other stores. 

Any thoughts on this approach vs the full width one?
959f3d3c-4461-11e6-9e92-943a652aed60

tvdeyen:
> The third and fourth screenshots start to look a little odd to me with form fields moving around and empty space being on the left side of a row. Is there a way to clean that up at all without breaking extensions and shops as you mentioned above?

No. We would need to move fields into other containers and this would potentially break extensions or shops. 

> And just to double check - is this replicable with the other listing pages as well?

No. Every listing page need its own treatment. 

> did you consider putting a max width on the search area so preserve the order/groupings of form elements a little better?

Yes, I did. I decided against because we would limit the space for potential search field additions. Some shops may add additional fields to the search form. By limiting the width we would reduce the possibility to add additional rows to that form.

tvdeyen:
We definitely should reconsider the layout of that form. It is not ideal and has some elements in strange places. 

But: I wanted to make the changeset as small as possible. 

I think this is a great step forward with lots of room for improvement, sure. But better what we have right now.

Mandily:
I agree 👍

PR #1785, opened 2017-03-18
Do not run callbacks when deleting dependent inventory units

mamhoff:
An inventory unit should not exist without its line item and its shipment.
The protective callbacks on `Spree::InventoryUnit` just make sure that
one can not destroy an inventory unit for a shipment that cannot be destroyed.

Added a safeguard so that we're not deleting line items with shipped inventory units.


mamhoff:
For a line item with 5000 quantity, this is the difference in `destroy` performance: 

`master`: 
```
95.640000   0.900000  96.540000 ( 97.042166)
```

This PR: 
```
5.590000   0.140000   5.730000 (  5.788784)
```

mamhoff:
I've been thinking about this, and it's not the right way to do this.

PR #1786, opened 2017-03-19
New admin table layout

tvdeyen:
The content of all table headers, state columns and action cells have text wrapping disabled now. For cases where we want to disble text wrapping in data cells we provide the `.no-wrap` class.

This also makes (very sensitive) use of the Bootstrap default table style. The main advantage is that Bootstrap uses rem (a relative unit) for table cell padding. As we plan to raise the body font size this is a huge win for us.

### Before

table-before

### After

table-after-13px

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
@@ -1,10 +1,8 @@
 table {
-  width: 100%;
-  margin-bottom: 15px;
-  border-collapse: separate;
+  // Extend all tables with Bootstrap's default table style
+  @extend .table;
 
   th, td {
-    padding: 7px 5px;
     border-right: 1px solid $color-border;
     border-bottom: 1px solid $color-border;
     vertical-align: middle;
@@ -128,14 +126,16 @@ table {
 
   }
 
+  th, td.actions, td.no-wrap, .state {
+    white-space: nowrap;
+  }
+
   thead {
     th {
-      padding: 10px;
       border-top: 1px solid $color-border;
       border-bottom: none;
       background-color: $color-tbl-thead;
       text-align: center;
-      font-size: 90%;
       font-weight: $font-weight-bold;
     }
   }
--- a/backend/app/views/spree/admin/style_guide/topics/tables/_building_tables.html.erb
+++ b/backend/app/views/spree/admin/style_guide/topics/tables/_building_tables.html.erb
@@ -3,6 +3,7 @@
     <thead>
       <tr>
         <th colspan="2">Item Description</th>
+        <th>Long Item Text</th>
         <th>Price</th>
         <th>Quantity</th>
         <th>Total</th>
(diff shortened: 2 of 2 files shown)


tvdeyen:
@Mandily another preparation for #1777

backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
```diff
@@ -1,14 +1,20 @@
...
     vertical-align: middle;
     text-overflow: ellipsis;
+    overflow: hidden;
+
+    // Do not wrap text in all header and content cells.
+    // Overflowing content gets truncated and an ellipsis is shown.
+    // If you want to disable this, add the .wrap-text class to the column.
+    &:not(.wrap-text) {
+      white-space: nowrap;
+    }
```
jhawthorn: I'm not sure about this. I think our default should be to wrap, and to specify areas where wrapping is undesirable (like actions).

backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
```diff
@@ -1,14 +1,20 @@
 table {
-  width: 100%;
-  margin-bottom: 15px;
-  border-collapse: separate;
+  // Extend all tables with Bootstrap's default table style
+  @extend .table;
```
jhawthorn: This isn't worse than what was here before (bootstrap's `.table` sets pretty much the same CSS as we already had here). It would be ideal if we were to move away from styling all `table` elements like this and instead required a class. That way we avoid needing to un-style the styles when we want a table to be "normal".

backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
tvdeyen: I think the default should be no wrap. Most of the time we display where little information in cells, mostly values (order numbers, email addresses, etc. You don't want to wrap these.

The only occasion where wrapping was useful (not even needed) was the image caption. It would be a much larger change set to add a no-wrapp class to all cells where we want this to happen. 

But this is just my preference. I'm also fine with having the opposite being the default. I would then vote for also making the header cells non wrapping. Nobody wants a 

"Created
at"

cell, right?

backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
tvdeyen: What would a "normal" table look like? Browser default? I always thought this is our default table style. But I understand your concern. Im also fine with adding a table class to all tables. The change set will be much larger then, though.

tvdeyen:
@jhawthorn I updated the commit and took a more conservative approach. Only table headers, the `.actions` cell and states have text wrapping disabled. Added a `.no-wrap` class, so we can decide which cells we want to have text wrapping disabled as well.

backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
tvdeyen: Took the opposite approach like suggested.

backend/app/assets/stylesheets/spree/backend/shared/_tables.scss
jhawthorn: We can deal with this later. It's not a new issue and shouldn't block this PR.

Issue #1788, opened 2017-03-21
Support Rails 5.1


jhawthorn:
[Rails 5.1.0.rc1 is out](http://weblog.rubyonrails.org/2017/3/20/Rails-5-1-rc1/)!

I've done a quick sprint to add support for Rails 5.1, and it wasn't too much work [jhawthorn/solidus@rails_5_1](https://github.com/jhawthorn/solidus/tree/rails_5_1). All specs pass with only a handful of warnings.

I don't know yet how we should approach this version upgrade w/r/t Solidus versions: 

1. We could do what we did with Solidus 1.4/2.0 and release two versions where the rails version was the only difference
2. We could just require the upgrade when moving to the release with the rails version change
3. We could release a version of Solidus supporting both versions of rails.

Option 3 isn't out of the realm of possibility, and is the most developer friendly thing we can do, so we should try for it. If we go down this path we'll need to change our testing infrastructure to simultaneously test our codebase against both Rails 5.0 and 5.1.

The major change between the rails versions is how `changed?`, `$ATTR_was` et al. work in an `after_save` hook. I'll let the rails warnings explain the change:

> DEPRECATION WARNING: The behavior of `changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_changes?` instead.

This is a good change in my opinion. It's explained very well in [the commit introducing the change](https://github.com/rails/rails/commit/16ae3db5a5c6a08383b974ae6c96faac5b4a3c81).

Unfortunately it is a difference between rails 5.0 and 5.1, and so far is the only major one I see.

A minor change is that stubs on `has_many` associations no longer work, like `expect(order.line_items).to receive(:each)`. This is fine, we shouldn't do this anyways. We can fix this on our current Rails 5.0 codebase and be forwards compatible.

mtomov:
Btw, there's a pretty good reason for stores to upgrade to 5.1, as this performance issue has just been addressed there:

- https://github.com/rails/rails/issues/25970

and I've come to experience it myself - on a product page with 180 variants - it takes > 100ms just the call to `product.variants.includes(:option_values)`

So, thanks indeed for the effort!

PR #1790, opened 2017-03-21
Avoid some stubbing in payment spec

jhawthorn:
This stubbing no longer worked under rails 5.1 and was hiding behaviour of our system under test

Part of the work on #1788

### DIFF
--- a/core/spec/models/spree/payment_spec.rb
+++ b/core/spec/models/spree/payment_spec.rb
@@ -17,12 +17,12 @@ describe Spree::Payment, type: :model do
   let(:card) { create :credit_card }
 
   let(:payment) do
-    payment = Spree::Payment.new
-    payment.source = card
-    payment.order = order
-    payment.payment_method = gateway
-    payment.amount = 5
-    payment
+    Spree::Payment.create! do |payment|
+      payment.source = card
+      payment.order = order
+      payment.payment_method = gateway
+      payment.amount = 5
+    end
   end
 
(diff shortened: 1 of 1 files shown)


core/spec/models/spree/payment_spec.rb
```diff
@@ -17,12 +17,12 @@
   let(:card) { create :credit_card }
 
   let(:payment) do
-    payment = Spree::Payment.new
-    payment.source = card
-    payment.order = order
-    payment.payment_method = gateway
-    payment.amount = 5
-    payment
+    Spree::Payment.create! do |payment|
```
jordan-brough: Just curious - any reason not to use the payment factory here?

kennyadsl: I had the same doubt and I noticed that in this same file payments are also created using [Spree::PaymentCreate.new](https://github.com/jhawthorn/solidus/blob/bb561a2327fbc6fc93af563c6154bbba08ea087e/core/spec/models/spree/payment_spec.rb#L693-L696) Service Object. Is there any reason/general rule to prefer a method or another?

jhawthorn: I prefer using just `described_class.create` in the spec for the model itself, otherwise you are testing the factory's output at the same time.

We probably shouldn't be using `PaymentCreate` in this file (that should be in the `payment_create_spec.rb` file), I suspect that's here because PaymentCreate replaced an some old payment behaviour.

PR #1792, opened 2017-03-21
don't set webkit-tap-highlight-color

brchristian:
Similar to the thinking behind https://github.com/solidusio/solidus/pull/1738, this removes customization of the experimental and non-standard `webkit-tap-highlight-color`, which can be nontrivial to override, in favor of simply keeping with browser defaults.


jhawthorn:
Thank you

PR #1793, opened 2017-03-22
Number with currency widget

jhawthorn:
cc @Mandily @tvdeyen 

I would like to introduce a "Number with currency" widget to use on every form with a monetary amount.

Currently whenever we are looking to input a monetary value, we just have a plain text box for the amount, and a second field for the currency (if it can be changed by that form).

This widget combines both inputs. It comes in two flavours, either with a dropdown for the currency (if it is changeable here) or fixed (where we previously didn't show a currency). I think it is better to always show the currency, both to make it obvious what currency is being used and for consistency.

### Product edit
| After     | Before |
| ---      | ---       |
|  |  |

### Promotion Rules and calculators
**After**

**Before**

### Payment
**After**
_(layout and padding could use some fixing here before the final version)_

**Before**

## Try it

To try it out you can clone this down or try [a demo on codepen](http://codepen.io/jhawthorn/full/qrPrXV/)

## Implementation

This depends on having the font size match bootstrap's, which @tvdeyen has been working on. The widget uses bootstrap's styling of input addons, for the basic layout.

The selector (when enabled) is a standard HTML `<select>` (not select2) with custom styling to make it blend in better. I've done some cross browser testing, and what I have now seems to function well on chrome/FF/IE11/safari.

There's a small backbone.js view which watches for changes to the currency so it can update the symbol.


Senjai:
:heart_eyes: :+1:

Mandily:
I like this a lot. I'm assuming we'd use the same thing in the settings where you set the master currency for each store?

jhawthorn:
> I like this a lot. I'm assuming we'd use the same thing in the settings where you set the master currency for each store?

This probably only makes sense when entering an amount.

backend/app/assets/javascripts/spree/backend/views/number_with_currency.js
```diff
@@ -0,0 +1,20 @@
...
+      currency = this.$('[data-currency]').data("currency");
+    }
+    if (currency) {
+      var currencyInfo = Spree.currencyInfo[currency];
+      this.$('.currency-selector-symbol').text(currencyInfo[0])
+    } else {
+      this.$('.currency-selector-symbol').text("");
+    }
+  }
+});
```
Sinetheta: We can break this up a bit

```js
Spree.Views.NumberWithCurrency = Backbone.View.extend({
  events: {
    'change input,select': "render"
  },

  initialize: function() {
    this.defaultCurrency = this.$('[data-currency]').data("currency");
  },

  getCurrency: function() {
    return this.$('.currency-selector').val() || this.defaultCurrency;
  },

  updateSymbol: function(currency) {
    var symbol = '';
    if (currency) {
      symbol = Spree.currencyInfo[currency][0];
    }
    this.$('.currency-selector-symbol').text(symbol)
  },

  render: function() {
    var currency = this.getCurrency();
    this.updateSymbol(currency);
  }
});
```

Sinetheta:
Sorry about the nits, I didn't read the PR message. Looks great as a UI @jhawthorn 👍

backend/app/assets/stylesheets/spree/backend/components/_currency_selector.scss
```diff
@@ -0,0 +1,39 @@
...
+  height: 100%;
+  padding-left: .5rem;
+  border: 0;
+  background: transparent;
+
+  -webkit-appearance:none;
+  -moz-appearance:none;
+  appearance:none;
+
+  background: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='1024' height='640'><path d='M1017 68L541 626q-11 12-26 12t-26-12L13 68Q-3 49 6 24.5T39 0h952q24 0 33 24.5t-7 43.5z'></path></svg>") 90%/12px 6px no-repeat;
```
jhawthorn: Can't be done. This is the background of a `<select>`

tvdeyen:
I like this a lot. There is still room for improving, though. For me it's hard to distinguish between changeable and "not changeable" currency. The only differentiation is the little black arrow. Maybe we could remove the background from the "display only" or use a different bg colour? Not sure.

Thanks John! 🍺

backend/app/assets/stylesheets/spree/backend/components/_currency_selector.scss
tvdeyen: There is also `asset-data-url`

https://github.com/rails/sass-rails#asset-data-urlrelative-asset-path

then we can make this changeable **and** it's still a data uri.

backend/app/assets/stylesheets/spree/backend/components/_currency_selector.scss
```diff
@@ -0,0 +1,39 @@
...
+
+  font-family: inherit;
+  color: inherit;
+
+  &:-moz-focusring {
+    color: transparent;
+    text-shadow: 0 0 0 #000;
+  }
+}
```
tvdeyen: can we have a:

```scss
&:hover {
  cursor: pointer;
}
```

here please?

jhawthorn: :+1: 

I think I'm going to try `&:hover:not(:active)`. When testing with just `:hover` it felt a little weird. When opening the select, the cursor changed from `pointer` to `arrow` and then quickly back. Using `:hover:not(:active)`, it just switches to `arrow`, and stays that way while open. This is the same behaviour we get from the select2 controlled selects.

tvdeyen: 👌🏻

jhawthorn:
@tvdeyen to help differentiate between selectable and fixed currencies, I'm going to try make the background white when it is a select box. If we also go with the change I propose in [#1797 "Replace select2 styling"](https://github.com/solidusio/solidus/pull/1797) it should match other select boxes and be even more obvious.

backend/app/assets/stylesheets/spree/backend/components/_currency_selector.scss
mtomov: another option might be to use _::before_ / _::after_ selectors and `content: ` property with the encoded symbol (not entirely sure what the svg is) as a value and position it absolutely if the element has `position: relative`

jhawthorn:
This is no longer WIP and is good to go.

Here's how the selectable vs fixed currency looks now with a white background.

jhawthorn:
At @tvdeyen's suggestion I've swapped this to use bootstrap's custom select styling, which has the advantage of looking (more or less) identical across browsers and reduces the CSS we need. to achieve this.

backend/app/assets/stylesheets/spree/backend/components/_number_with_currency.scss
```diff
@@ -0,0 +1,34 @@
...
+      @extend .c-select;
+
+      display: inline-block;
+      width: 100%;
+      height: 31px;
+      border: 0;
+
+      &::-ms-expand {
+        // Required to see full text of selected value on IE11
+        display: none;
```
tvdeyen: Bootstrap [already takes care of hiding this](https://github.com/solidusio/solidus/blob/master/backend/vendor/assets/stylesheets/solidus_admin/bootstrap/_custom-forms.scss#L154).

backend/app/assets/stylesheets/spree/backend/components/_number_with_currency.scss
```diff
@@ -0,0 +1,34 @@
...
+  & &-select {
+    padding: 0;
+    background: #fff;
+
+    select {
+      @extend .c-select;
+
+      display: inline-block;
+      width: 100%;
+      height: 31px;
```
tvdeyen: Can we use Bootstraps `$input-height`here?

backend/app/assets/stylesheets/spree/backend/components/_number_with_currency.scss
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  &-addon {
+    width: 5.5rem;
+    text-align: left;
+  }
+
+  & &-select {
+    padding: 0;
+    background: #fff;
```
tvdeyen: Let's make use of Bootstraps `$input-bg` here.

backend/app/assets/stylesheets/spree/backend/components/_number_with_currency.scss
```diff
@@ -0,0 +1,34 @@
+.number-with-currency {
+  &-symbol {
+    min-width: 2.5em;
+    text-align: center;
```
tvdeyen: `text-align: center;` is already set by Bootstraps `.input-group-addon`

backend/app/assets/stylesheets/spree/backend/components/_number_with_currency.scss
jhawthorn: `opacity: 0` didn't do the trick on IE11. The "D" in "USD" was cut off, despite there being lots of room in the select box. See comment.

tvdeyen: Interesting. Ok. Thanks for clarification.

backend/app/assets/stylesheets/spree/backend/components/_number_with_currency.scss
jhawthorn: This leaves a `1px` gap on the top and bottom (height is `2.25rem == 29px`), but I don't think that's a huge issue, and is worth it to avoid this magic number and ensure the input will scale well if our font size changes.

jhawthorn:
@tvdeyen Thanks. I have addressed your feedback.

mtomov:
Just played with this for a bit, and it's super cool! The up and down keys from the keyboard work just as well, and the immediate feedback that you get with the currency sign change in front is awesome! Very nice piece of UI! Thanks a lot!

PR #1796, opened 2017-03-24
Remove any whitespace from a promo code.

eric1234:
There is no legit reason to have spaces in codes as the user would have to duplicate the spacing when they entered the code during checkout.

I have gotten real world cases where a store admin has inadvertently included a space before/after the code (due to a copy/paste error), as well as space within the code (due to font size making it hard to
see the space).

I considered putting a validation in place to correct the user rather than silently modifying their code but we are already silently modifying by downcasing the code. Even after being told of a spacing problems,
some users might have difficultly determing where the space is or understanding the error message. In the interest of just doing what they likely mean, I think silently modifying is the lesser of two evils.

### DIFF
--- a/core/app/models/spree/promotion_code.rb
+++ b/core/app/models/spree/promotion_code.rb
@@ -6,7 +6,7 @@ class Spree::PromotionCode < Spree::Base
   validates :value, presence: true, uniqueness: { allow_blank: true }
   validates :promotion, presence: true
 
-  before_save :downcase_value
+  before_save :normalize_code
 
   self.whitelisted_ransackable_attributes = ['value']
 
@@ -36,7 +36,7 @@ class Spree::PromotionCode < Spree::Base
 
   private
 
-  def downcase_value
-    self.value = value.downcase
+  def normalize_code
+    self.value = value.downcase.strip
   end
 end
--- a/core/spec/models/spree/promotion_code_spec.rb
+++ b/core/spec/models/spree/promotion_code_spec.rb
@@ -4,13 +4,25 @@ RSpec.describe Spree::PromotionCode do
   context 'callbacks' do
     subject { promotion_code.save }
 
-    describe '#downcase_value' do
-      let(:promotion) { create(:promotion, code: 'NewCoDe') }
+    describe '#normalize_code' do
+      let(:promotion) { create(:promotion, code: code) }
       let(:promotion_code) { promotion.codes.first }
 
-      it 'downcases the value before saving' do
-        subject
-        expect(promotion_code.value).to eq('newcode')
+      before { subject }
+
+      context 'with mixed case' do
+        let(:code) { 'NewCoDe' }
+
+        it 'downcases the value before saving' do
+          expect(promotion_code.value).to eq('newcode')
+        end
+      end
+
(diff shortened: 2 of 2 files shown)


core/app/models/spree/promotion_code.rb
```diff
@@ -39,4 +40,8 @@ def usage_limit
   def downcase_value
     self.value = value.downcase
   end
+
+  def remove_excess_whitespace
+    value&.gsub! /\s+/, ''
```
eric1234: I assume the "ambiguous first argument" is indicating I should add parens. So `gsub!(/\s+/, '')` (under the idea that somehow that is clearer).

Is the `unexpected token error` referring to the safe navigation operator? Is this a project decision (compat with older Ruby) or just hound not knowing the more modern syntax?

jordan-brough: It's compat with older Ruby. Solidus currently only requires Ruby 2.2.2 so we can't use the safe navigation operator.

jordan-brough: We already have a validation for code presence so I don't think we need to consider `nil` values.  Also I think we should use `self.value = value.strip` (not bang method and not worry about internal whitespace).  Open to others' thoughts on that.

eric1234: Up to you guys regarding the internal whitespace. I will say I have had a real world case where the admin put whitespace in the middle and couldn't see it because of the small font size. The question is how far do we go to make it fool proof since there is always a bigger fool?

We do know whitespace should never be in a code so doesn't seem to hurt to remove it. The person who pays my paycheck is the one who added the whitespace so if you prefer `strip` I'll just have to do a local override.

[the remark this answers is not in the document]
kennyadsl: I think `.strip` can be enough for fixing the main issue (whitespace before or after). Each store will be able to override the method easily if needed.

kennyadsl:
Talking with @mamhoff he convinced me that the current version is good enough to be merged. Having promo codes with spaces inside is not a documented feature so we should not care that much about it.

eric1234:
The last paragraphs in my original PR message explained why I opted for silently modifying over explicit validation. I saw it as the lesser of two evils but that is a bit of opinion. If you guys prefer validation then go ahead and close the PR.

eric1234:
I didn't expect so a small little change to make so much discussion. To get this thing closed and move onto more exciting things I have updated the PR to `strip` so that spaces within a code are supported.

I also went ahead and combined it with the existing downcasing method to reduce the clutter.

As if you should merge this or use @jhawthorn validation version (or both) I'll leave that up to you guys. I prefer the system just do what they mean rather than having the user try to remove characters they cannot see but understand if you prefer it the other way.

Regardless the PR is updated so it is ready if you do want to merge. If not then feel free to close.

mamhoff:
Hey @eric1234 thank you, very nice work (and yeah, this has generated a lot of discussion). 

Would you mind merging the three commits into one (and removing the `code nazi` language)?

Thank you again for your patience and hard work.

eric1234:
K, squashed and rebased to master. My frustration with overzealous code linters removed. :)

PR #861, opened 2016-02-12
Remove select2 styling

jhawthorn:
I believe the default styling of select2 is superior.

Before:

After:

If we ever upgrade to select2 4.0, the default style should integrate even better

PR #1797, opened 2017-03-24
Replace select2 styling

jhawthorn:
A year ago I submitted a similar PR (#861) which removed the select2 styling. At the time we tabled it because we intended to upgrade to select2 4.0, from which we preferred the styles. I no longer think we will be performing that upgrade (see #763), so giving this another shot.

The existing styles we had don't match the visual language we used for the rest of our inputs (flat, thin border, white background) and instead more closely matched buttons (solid blue background) which have a different purpose.

This replaces our existing custom select2 styles with much simpler styles intended to better match our other input elements. This also matches the look of [bootstrap-styled select elements](https://v4-alpha.getbootstrap.com/components/forms/#form-controls), which would allow us to use those in the future.

## Todo

* [x] Improve styling of multi-selects (ex. taxons and option types on a product)
* [x] Improve use of icons (search, clear, spinner). Old styles used font awesome, maybe bring those back

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
+++ b/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
@@ -336,8 +336,8 @@ $dropdown-border-width:          $border-width !default;
 $dropdown-divider-bg:            #e5e5e5 !default;
 
 $dropdown-link-color:            $gray-dark !default;
-$dropdown-link-hover-color:      darken($gray-dark, 5%) !default;
-$dropdown-link-hover-bg:         #f5f5f5 !default;
+$dropdown-link-hover-color:      #fff !default;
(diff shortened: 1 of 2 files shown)


jhawthorn:
Updated this to use the same dropdown arrow as bootstrap's `custom-select` styling.

mamhoff:
👏

jhawthorn:
This is out of WIP

For icons, I only brought back the "close" icon from font-awesome. The dropdown icon is provided by bootstrap, and I chose to forgo the search icon, which I don't think was useful.

Colours come from bootstrap variables, which will be slightly off until #1809 is merged.

jhawthorn:
@tvdeyen I agree with all your changes and cherry-picked your commit. I needed to make one change (001b702a3faa20f074823a19ebb836a331186209) to keep the close (clear) button for single-select in the right place.

Issue #1548, opened 2016-10-20
Use HTML5 Validations for Frontend Checkout


jhawthorn:
Extracted from #1542

@tvdeyen suggested

> As a bonus, as we're already here we could also add required: true to trigger browser based validation. 

@mtomov replies

> There's actually that beast here - jQuery.validate that takes care of javascript validations. Not sure if it will play well with the html validations as it's a rather old version.
> 
> As the address form is the only one to use this plugin, as far as I know, (payment form uses a different plugin), if you consider it appropriate, I can remove jQuery validate altogether and replace the presence validations with html5 ones. Whoever wants then can add a validation plugin of their own (such as the Foundation's one Abide, or I'm sure bootstrap must also have one)
> 
> I think even for experienced developers - having a reference to what exactly the correct auto-complete attributes are, without having to search for, is great to have.

@tvdeyen responds

> :+1: for removing jQuery.validate and replace it with HTML5 based validations. But this is just my 2 cents.

@cbrunsdon adds

> I'm pro swapping to html5 validations as well, though I know very little about them.

bbuchalter:
My team has learned the hard way that HTML5 form validations are not supported by Mobile Safari. If you wish to support that browser, you may not want to make that change.

tvdeyen:
@bbuchalter we still have the model validations. Front-end validations are just, and should always, be an UX addition for users. Also, the form we provide is only an example and most likely will never be used on production as is. Stores that care about front end validations on mobile Safari can easily drop in some JS lib. 

We can not provide front end templates that will fit all use cases. They are very easily replaceable by stores and are just an example/demo of a minimal working store and not a one-size-fit-em-all solution.

jhawthorn:
For reference
http://caniuse.com/#feat=form-validation
https://bugs.webkit.org/show_bug.cgi?id=28649

cbrunsdon:
As this is the default frontend, which is widely replaced anyway, I'd be pro putting in html5 validations with a warning in the changelog that someone might want to polyfill in something for iOS. 

It wouldn't stop individual stores from using whatever they wanted, and we'd be reducing our reliance on an external lib in a forward-looking way.

bofrede:
The iOS problem is finally about to go away.
https://webkit.org/blog/7093/release-notes-for-safari-technology-preview-19/

jhawthorn:
Completed by #2264

Issue #1794, opened 2017-03-23
Replace E-mail Validator


gmacdougall:
Our current e-mail validator is not good. It is a simplistic regular expression that incorrectly flags some e-mails as being valid though they are not.

The [validates_email_format_of](https://github.com/validates-email-format-of/validates_email_format_of) gem can be of assistance here.  It has a superior implementation of e-mail validation, and superior test cases.

We should consider using this solution rather than our current validator.

jasonfb:
https://davidcel.is/posts/stop-validating-email-addresses-with-regex/

jasonfb:
having said that, we really like this approach (which isn't really an email validator at all, more a suggester): https://github.com/mailcheck/mailcheck

You can check that out live by going to Kickstarter (log out if you are already logged in), then click 'Sign up'. Then enter your email but purposefully leave off the ```.com``` (or in your case ```.ca```), you'll see it suggest the right email to you

mtomov:
I think that so log as it does not flag correct emails as being wrong, it's fine as it is. I don't consider it to be a core responsibility of Solidus to handle such use cases with perfection. They can be implemented at the individual store level.

gmacdougall:
I tend to think that since we're [providing an e-mail validator](https://github.com/solidusio/solidus/blob/v2.1.0/core/lib/spree/core/validators/email.rb) we should make it decent.

We can also try to get out of that business altogether, but that's a bit more of an invasive change.

tvdeyen:
> We can also try to get out of that business altogether, but that's a bit more of an invasive change.

Why not let the user class provider handle this (in our case `solidus_auth_devise` could let Devise validate this email)?

That's definitely not something we should do in core.

dividedharmony:
This didn't have the Solidus Hackathon label, but I thought it was simple enough, so I put in PR #1965 if this issue is still relevant

PR #1799, opened 2017-03-26
Use HTML5 Validations for Frontend Checkout #1548

gitmihalis:
Hey there. This is my first contribution to open source! 

I removed the jQuery.validate plugin and added the HTML5 required attributes to the relevant input fields in checkout, as well as a couple of patterns to match basic phone number and email addresses. I don't think there are any new test to write seeing as none of the functionality has changed.

Looking forward to your feedback.

### DIFF
--- a/frontend/app/views/spree/checkout/payment/_gateway.html.erb
+++ b/frontend/app/views/spree/checkout/payment/_gateway.html.erb
@@ -3,7 +3,7 @@
 
 <p class="field">
   <%= label_tag "name_on_card_#{payment_method.id}", Spree.t(:name_on_card) %><span class="required">*</span><br />
-  <%= text_field_tag "#{param_prefix}[name]", "#{@order.billing_firstname} #{@order.billing_lastname}", { id: "name_on_card_#{payment_method.id}", utocomplete: "cc-name",  required: true } %>
+  <%= text_field_tag "#{param_prefix}[name]", "#{@order.billing_firstname} #{@order.billing_lastname}", { id: "name_on_card_#{payment_method.id}", autocomplete: "cc-name",  required: true } %>
 </p>
 
 <p class="field" data-hook="card_number">


frontend/app/views/spree/checkout/payment/_gateway.html.erb
```diff
@@ -3,12 +3,12 @@
 
 <p class="field">
   <%= label_tag "name_on_card_#{payment_method.id}", Spree.t(:name_on_card) %><span class="required">*</span><br />
-  <%= text_field_tag "#{param_prefix}[name]", "#{@order.billing_firstname} #{@order.billing_lastname}", { id: "name_on_card_#{payment_method.id}", autocomplete: "cc-name" } %>
+  <%= text_field_tag "#{param_prefix}[name]", "#{@order.billing_firstname} #{@order.billing_lastname}", { id: "name_on_card_#{payment_method.id}", utocomplete: "cc-name",  required: true } %>
```
mtomov: typo

mtomov:
I am all in favour of this change :+1: 

 - Foundation for example has a much lighter (read 100 times smaller) validation JS, which looks better and all, so I am having to always remove this jquery validation. It's just too much for 2 forms on the frontend.

frontend/app/views/spree/address/_form.html.erb
```diff
@@ -65,30 +67,31 @@
             class: !have_states ? 'required' : '',
             style: have_states ? 'display: none;' : '',
             disabled: have_states,
-            autocomplete: address_type + ' address-level1'
+            autocomplete: address_type + ' address-level1',
+            required: true
```
mtomov: The `required` here should maybe follow the same logic as the `required` class. `!have_states ? 'required' : ''`.

Do we maybe need to remove the `required` class from those inputs as well? And use maybe the `required` attribute as a selector in the css if there's need.

frontend/app/views/spree/address/_form.html.erb
```diff
@@ -65,30 +67,31 @@
...
   <p class="field" id=<%="#{address_id}zipcode" %>>
     <%= form.label :zipcode, Spree.t(:zip) %><% if address.require_zipcode? %><span class="required">*</span><% end %><br />
-    <%= form.text_field :zipcode, class: "#{'required' if address.require_zipcode?}", autocomplete: address_type + ' postal-code' %>
+    <%= form.text_field :zipcode, class: "#{'required' if address.require_zipcode?}", autocomplete: address_type + ' postal-code',  required: true %>
   </p>
 
   <p class="field" id=<%="#{address_id}phone" %>>
     <%= form.label :phone, Spree.t(:phone) %><% if address.require_phone? %><span class="required">*</span><% end %><br />
-    <%= form.phone_field :phone, class: "#{'required' if address.require_phone?}", autocomplete: address_type + ' home tel' %>
+    <%= form.phone_field :phone, class: "#{'required' if address.require_phone?}", autocomplete: address_type + ' home tel', required: true, pattern: '^([+]?\d{1,2}[-\s]?|)\d{3}[-\s]?\d{3}[-\s]?\d{4}$' %>
```
mtomov: I think that the RegEx is too complicated to sit in the views, and we'd need tests for it, as at the moment I have no clue what its expected functionality would be. 

In addition, any kind of validation in this field would have to consider any sort of telephone number that exists around the world, which I believe is a difficult job. Just recently I was to add an auto-format in this field for the US phone style, which would wrap the first 3 digits I believe it was in parentheses `(123) ...`

frontend/app/views/spree/checkout/edit.html.erb
```diff
@@ -12,7 +12,7 @@
         <% if @order.state == 'address' || !@order.email? %>
           <p class="field" style='clear: both'>
             <%= form.label :email %><br />
-            <%= form.text_field :email %>
+            <%= form.text_field :email, required: true, pattern: "^[a-zA-Z0-9\.]+@[a-zA-Z0-9]+(\-)?[a-zA-Z0-9]+(\.)?[a-zA-Z0-9]{2,6}?\.[a-zA-Z]{2,6}$", title: "username@example.com" %>
```
mtomov: Same thing as for the phone field is valid here - the logic is too complicated for the view, and it needs to be tested, so to understand what the expected validations are.

But before you go there, consider that there's already sort of an established RegEx in Solidus. See issue #1794  for more info.s

frontend/app/views/spree/address/_form.html.erb
gitmihalis: I don't think the `!have_states? 'required'` logic makes sense for anything else besides styles here. It's simple enough to remove the `required: true` attribute for a fringe case where `address-level1` would not be required, but the form is for Debit/Credit or Shipping which **do** require an address.

frontend/app/views/spree/checkout/edit.html.erb
gitmihalis: Yes, I could take it out.

kennyadsl:
This can be closed since we merged this via #2264, thanks @gitmihalis !

PR #1800, opened 2017-03-27
Do not consider pending inventory units cancelable

mamhoff:
Before an order is complete, all its inventory units are marked as "pending". Once
the order passes the `complete` step, the inventory units' `pending` attribute is
set to false.

Item cancelations were designed for completed orders only, as before that, one could
just change inventory on line items. Therefore, we should not consider pending inventory
units as cancelable.


mamhoff:
Once https://github.com/solidusio/solidus/pull/1787 is in, this will be green. I'll rebase then rather than create a merge conflict.

PR #1801, opened 2017-03-27
Don't display inactive payment methods on frontend or backend

luukveenis:
## The issue
After upgrading to Solidus 2.1, I noticed that my inactive payment methods were being displayed on the payment page in checkout and the `payments#new` page in the admin. The issue was introduced in #1540 because the `Spree::PaymentMethod.available` scope used to filter payment methods by `active`, but the new `available_to_users` and `available_to_admin` scopes only look at their respective attributes on the payment method. The `Spree::Order#available_payment_methods` method wasn't updated to include the `active` scope accordingly. Similarly, the `load_data` method in the admin payment methods controller is missing the `active` scope.

While writing the failing spec for this, I noticed that the `available_to_store` scope was not behaving as expected. It returned `self` if the store did not have any payment methods, which presumably was expected to return the current scope, but actually returns the `Spree::PaymentMethod` class constant. This wasn't breaking specs because we chained other scopes that would return payment methods as expected.

## The fix
I've added the `active` scope to `#available_payment_methods` and in `#load_data` to ensure we don't include inactive payment methods.

I also updated the `available_to_store` scope to raise an `ArgumentError` if the store provided is nil, and return all payment methods in the current scope if the store doesn't have any payment methods. I added specs that should catch this issue in the future.

### DIFF
--- a/backend/spec/controllers/spree/admin/payments_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/payments_controller_spec.rb
@@ -75,14 +75,28 @@ module Spree
       describe '#new' do
         # Regression test for https://github.com/spree/spree/issues/3233
         context "with a backend payment method" do
-          before do
-            @payment_method = create(:check_payment_method, available_to_admin: true)
+          context "and the payment method is active" do
+            before do
+              @payment_method = create(:check_payment_method, available_to_admin: true)
+            end
+
+            it "loads the payment method" do
+              get :new, params: { order_id: order.number }
+              expect(response.status).to eq(200)
+              expect(assigns[:payment_methods]).to include(@payment_method)
+            end
           end
 
-          it "loads backend payment methods" do
-            get :new, params: { order_id: order.number }
-            expect(response.status).to eq(200)
-            expect(assigns[:payment_methods]).to include(@payment_method)
+          context "and the payment method is inactive" do
+            before do
+              @payment_method = create(:check_payment_method, available_to_admin: true, active: false)
+            end
+
+            it "does not load the payment method", :pending do
+              get :new, params: { order_id: order.number }
+              expect(response.status).to eq(200)
+              expect(assigns[:payment_methods]).to be_empty
+            end
           end
         end
       end
--- a/backend/spec/controllers/spree/admin/payments_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/payments_controller_spec.rb
@@ -92,7 +92,7 @@ module Spree
               @payment_method = create(:check_payment_method, available_to_admin: true, active: false)
             end
 
-            it "does not load the payment method", :pending do
+            it "does not load the payment method" do
               get :new, params: { order_id: order.number }
               expect(response.status).to eq(200)
               expect(assigns[:payment_methods]).to be_empty
--- a/core/app/models/spree/payment_method.rb
+++ b/core/app/models/spree/payment_method.rb
@@ -17,7 +17,10 @@ module Spree
     scope :active, -> { where(active: true) }
     scope :available_to_users, -> { where(available_to_users: true) }
     scope :available_to_admin, -> { where(available_to_admin: true) }
-    scope :available_to_store, -> (store) { (store.present? && store.payment_methods.empty?) ? self : store.payment_methods }
+    scope :available_to_store, ->(store) do
+      raise ArgumentError, "You must provide a store" if store.nil?
+      store.payment_methods.empty? ? all : where(id: store.payment_method_ids)
+    end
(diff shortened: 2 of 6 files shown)


core/app/models/spree/payment_method.rb
```diff
@@ -17,7 +17,10 @@ class PaymentMethod < Spree::Base
     scope :active, -> { where(active: true) }
     scope :available_to_users, -> { where(available_to_users: true) }
     scope :available_to_admin, -> { where(available_to_admin: true) }
-    scope :available_to_store, -> (store) { (store.present? && store.payment_methods.empty?) ? self : store.payment_methods }
+    scope :available_to_store, ->(store) do
+      raise ArgumentError, "You must provide a store" if store.nil?
+      store.payment_methods.empty? ? all : where(id: store.payment_method_ids)
+    end
```
mtomov: The whole method is not very clearly written. It's sort of all in-reverse. I think it should be just:

```ruby
scope :available_to_store, -> (store) { where(store: store) }
```

That should also work if store is `nil`.

The `raise ArgumentError` is not much improvement over the Undefined method `payment_methods` on `Nil:Class` that one would usually get.

I think the whole method

luukveenis: Thanks for the feedback!

Unfortunately `where(store: store)` doesn't work, because the payment method <-> store association is a `has_many :through`.

The reason for the conditional is to keep the existing, or at least expected, behaviour where if the store doesn't have any payment methods, we return all of them instead of an empty collection.

I agree the ArgumentError is not a drastic improvement. I discussed this with @jhawthorn on Friday and would be okay just letting it error the usual way too.

adammathys: That won't quite work due to the nature of the relationship between stores and payment methods. It could be writen as:
```ruby
scope :available_to_store, -> (store) { joins(:stores).merge(Spree::Store.where(id: store.id)) }
```

However, this would mean a divergence from how we typically handle store associations. Normally, if a store doesn't have anything associated to it we assume everything belongs to that store. I believe this decision was made a while back to help handle single store apps by ensuring everything continued to work without having to create the missing associations.

While I'm definitely not opposed to changing that behavior, it would require a larger amount of work and a conscious decision to have all single store apps require some additional configuration behind the scenes.

jhawthorn:
This was a 2.1 regression, and I think we should backport it for a 2.1.1 release.

core/app/models/spree/payment_method.rb
mtomov: Absolutely you are both right, I did not see it through, and I am not that familiar with that part of the code. Thanks for the explanation!

core/app/models/spree/payment_method.rb
```diff
@@ -17,7 +17,10 @@ class PaymentMethod < Spree::Base
     scope :active, -> { where(active: true) }
     scope :available_to_users, -> { where(available_to_users: true) }
     scope :available_to_admin, -> { where(available_to_admin: true) }
-    scope :available_to_store, -> (store) { (store.present? && store.payment_methods.empty?) ? self : store.payment_methods }
+    scope :available_to_store, ->(store) do
+      raise ArgumentError, "You must provide a store" if store.nil?
```
jordan-brough: > I agree the ArgumentError is not a drastic improvement. I discussed this with @jhawthorn on Friday and would be okay just letting it error the usual way too.

^ I agree with removing the nil check.  The prior code didn't actually protect against nils and as @mtomov mentioned it doesn't really add much and we don't do it in most places

core/spec/models/spree/order_spec.rb
```diff
@@ -538,6 +538,16 @@ def merge!(other_order, user = nil)
       expect(order.available_payment_methods).to include(payment_method)
     end
 
+    it "does not include inactive payment methods" do
```
mamhoff: This spec surely wants to have a `pending` call to make sure this commit doesn't break the suite, no?

backend/spec/controllers/spree/admin/payments_controller_spec.rb
```diff
@@ -75,14 +75,28 @@ module Admin
...
         # Regression test for https://github.com/spree/spree/issues/3233
         context "with a backend payment method" do
-          before do
-            @payment_method = create(:check_payment_method, available_to_admin: true)
+          context "and the payment method is active" do
+            before do
+              @payment_method = create(:check_payment_method, available_to_admin: true)
+            end
+
+            it "loads the payment method" do
```
mamhoff: Can we have `pending` calls for both failing specs in this commit?

Issue #1803, opened 2017-03-28
Spree::Admin::VariantsController#index breaks for deleted products


bbuchalter:
### Steps to reproduce
```
chruby 2.4
git clone https://github.com/solidusio/solidus.git
cd solidus
bundle
bundle exec rake sandbox
cd sandbox
bin/rails s
```
* Visit http://localhost:3000/admin/products/ruby-on-rails-ringer-t-shirt/edit
* login as admin@example.com / `test123`
* Add the "Size" option type to the product and save
* Create a variant
* Return to the product admin index
* Search for "ringer"
* Delete the product
* Search for "ringer" with deleted and click on the product
* Go to the variants tab

### Expected behavior
I should see the variants for the product

### Actual behavior
```
NoMethodError (undefined method `variants' for nil:NilClass):

/Users/brian/workspace/solidus_master/backend/app/controllers/spree/admin/resource_controller.rb:189:in `collection'
/Users/brian/workspace/solidus_master/backend/app/controllers/spree/admin/variants_controller.rb:40:in `collection'
/Users/brian/workspace/solidus_master/backend/app/controllers/spree/admin/resource_controller.rb:144:in `load_resource'
```

### System configuration
**Solidus Version**: `master`

**Extensions in use**: none

PR #1804, opened 2017-03-28
Fix Spree::Admin::VariantsController#index for deleted products

bbuchalter:
Fixes https://github.com/solidusio/solidus/issues/1803

When a `Product` was deleted, `@parent` would not be set correctly when accessing `Spree::Admin::VariantsController#index`. Adding the `with_deleted` scope for the assignment to `Spree::Admin::ResourceController#parent` addresses this problem.

The heart of the change happens in https://github.com/solidusio/solidus/pull/1804/commits/dec876c158abd343fcd2c41b3153744a0f724e7d while the rest of the commits clean up the specs for `Spree::Admin::VariantsController`.

### DIFF
--- a/backend/app/controllers/spree/admin/variants_controller.rb
+++ b/backend/app/controllers/spree/admin/variants_controller.rb
@@ -55,6 +55,11 @@ module Spree
       def redirect_on_empty_option_values
         redirect_to admin_product_variants_url(params[:product_id]) if @product.empty_option_values?
       end
+
+      def parent
+        @parent ||= Spree::Product.with_deleted.find_by(slug: params[:product_id])
+        @product = @parent
+      end
     end
   end
 end
--- a/backend/spec/controllers/spree/admin/variants_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/variants_controller_spec.rb
@@ -27,25 +27,25 @@ module Spree
           end
         end
 
-        context "assigning @collection" do
+        context "the value of @collection" do
           let!(:variant) { create(:variant, product: product) }
           let!(:deleted_variant) { create(:variant, product: product) }
 
           context "with deleted variants" do
             before { deleted_variant.destroy }
 
-            context "deleted is not requested" do
-              it "does not assign deleted variants for a requested product" do
+            context "when deleted is not requested" do
+              it "excludes deleted variants" do
                 subject
                 expect(assigns(:collection)).to include variant
                 expect(assigns(:collection)).not_to include deleted_variant
               end
             end
 
-            context "deleted is requested" do
+            context "when deleted is requested" do
               let(:params) { { product_id: product.slug, deleted: "on" } }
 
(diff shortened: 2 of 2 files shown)


bbuchalter:
Obviously this isn't as simple as adding `with_deleted`, as this causes 111 specs to fail. I could add a check to see if `parent_data[:model_class].respond_to?(:with_deleted)`, but that's pretty ugly. However, I don't have any alternative suggestions. Going to let this sit for a day to see if I get any feedback on how we'd like to move forward with this, or come up with else.

jordan-brough:
Seems worth fixing in some way.  Some thoughts:

> I could add a check to see if parent_data[:model_class].respond_to?(:with_deleted), but that's pretty ugly

`parent_data[:model_class].paranoid?` is maybe a little better?

Any chance this might open up any vulnerabilities for custom models stores have that inherit from `ResourceController`?

We could also just override the required methods in `Spree::Admin::VariantsController` itself?

bbuchalter:
> We could also just override the required methods in Spree::Admin::VariantsController itself?

As I was thinking about it more, this was the direction I was taking. The question now is to what degree: do I just override `#parent` or do I stop trying to use `ResourceController` as the base of `VariantsController`? Is there an interest in moving away from `ResourceController`?

mamhoff:
I think the change with the smallest impact that fixes the bug would be overriding `#parent` on the `VariantsController`.

jordan-brough:
Another thought: Would it work to just do:
```ruby
belongs_to 'spree/product', find_by: :slug, model_class: Spree::Product.with_deleted
```
?

jordan-brough:
I'm not sure I love the thought of setting an option called `model_class` to a scope, but it seems like it could be OK.

bbuchalter:
@jordan-brough your solution is clever, but I think overriding `parent` is more clear.

bbuchalter:
Poke.

backend/app/controllers/spree/admin/variants_controller.rb
```diff
@@ -55,6 +55,13 @@ def variant_includes
       def redirect_on_empty_option_values
         redirect_to admin_product_variants_url(params[:product_id]) if @product.empty_option_values?
       end
+
+      def parent
+        if parent_data.present?
+          @parent ||= parent_data[:model_class].with_deleted.send("find_by_#{parent_data[:find_by]}", params["#{model_name}_id"])
+          instance_variable_set("@#{model_name}", @parent)
+        end
+      end
```
jhawthorn: We know what all these values are, so we can probably just

``` ruby
def parent
  @parent ||= Spree::Product.with_deleted.find_by(slug: params[:product_id])
  @product = @parent
end
```

bbuchalter: I agree this is an improvement. Will fix.

bbuchalter:
Feedback implemented. Ready for another review.

bbuchalter:
Poke.

PR #1806, opened 2017-03-28
Promotion sorting

eric1234:
bde9b56 is the meat of this request. 2dbba35 just removes an unnecessary constraint. See commit messages for full details.

### DIFF
--- a/backend/app/controllers/spree/admin/promotions_controller.rb
+++ b/backend/app/controllers/spree/admin/promotions_controller.rb
@@ -43,6 +43,7 @@ module Spree
         @search = @collection.ransack(params[:q])
         @collection = @search.result.
           includes(promotion_includes).
+          references(promotion_references).
           page(params[:page]).
           per(params[:per_page] || Spree::Config[:promotions_per_page])
 
@@ -55,6 +56,10 @@ module Spree
 
       def promotion_includes
         [:promotion_actions]
+      end
+
+      def promotion_references
+        []
       end
     end
   end
--- a/backend/app/controllers/spree/admin/promotions_controller.rb
+++ b/backend/app/controllers/spree/admin/promotions_controller.rb
@@ -41,7 +41,7 @@ module Spree
 
         @collection = super
         @search = @collection.ransack(params[:q])
-        @collection = @search.result(distinct: true).
+        @collection = @search.result.
           includes(promotion_includes).references(:codes).
           page(params[:page]).
           per(params[:per_page] || Spree::Config[:promotions_per_page])
--- a/backend/app/controllers/spree/admin/promotions_controller.rb
+++ b/backend/app/controllers/spree/admin/promotions_controller.rb
@@ -42,7 +42,7 @@ module Spree
         @collection = super
         @search = @collection.ransack(params[:q])
         @collection = @search.result(distinct: true).
-          includes(promotion_includes).
+          includes(promotion_includes).references(:codes).
           page(params[:page]).
           per(params[:per_page] || Spree::Config[:promotions_per_page])
 
@@ -54,7 +54,7 @@ module Spree
       end
 
       def promotion_includes
-        [:promotion_actions]
+        [:promotion_actions, :codes]
       end
     end
   end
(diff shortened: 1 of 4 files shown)


backend/app/controllers/spree/admin/promotions_controller.rb
```diff
@@ -41,8 +41,8 @@ def collection
 
         @collection = super
         @search = @collection.ransack(params[:q])
-        @collection = @search.result(distinct: true).
-          includes(promotion_includes).
+        @collection = @search.result.
+          includes(promotion_includes).references(:codes).
```
kennyadsl: I'd let this addition go into a new line, just to keep consistency with existing code

eric1234:
Yes, you are correct. I had thought about that but didn't think in practice folks would have that many codes on a single promotion. If that is a concern we could address in one of two ways:

* Just remove the code sorting similar to the usage sorting. I know my store won't be using codes at that scale so can just make it a local change. I think this is the best approach as sorting by code isn't probably a huge issues for most stores.
* You could define a new relationship that is scoped to just a single code and sort on that. This would scale to any size of codes but probably not worth the code complexity.

eric1234:
Yea, I'll just remove it. I was really thinking the sort would focus around promos with a single code and with multiple codes it would be more undefined. The store I am working with really just uses single codes so I'll just bring it local.

eric1234:
@jordan-brough 

158053c - This backs the code sorting back out.
4e03ad8 - This adds a hook to include references that I need to be able to do the code sorting locally. It's in the same design as the hook for includes. Defaults to a no-op.
ab90f50 - Not strictly related to sorting but related to the same page. A minor consistency fix I'm trying to sneak in. :)

backend/app/controllers/spree/admin/promotions_controller.rb
```diff
@@ -41,8 +41,9 @@ def collection
 
         @collection = super
         @search = @collection.ransack(params[:q])
-        @collection = @search.result(distinct: true).
+        @collection = @search.result.
           includes(promotion_includes).
+          references(promotion_references).
```
jordan-brough: Two thoughts:

1) Instead of `promotion_includes` and `promotion_references`, maybe we should just have `promotions_scope`
2) Maybe we should provide a config instead of requiring a monkey-patch

E.g.:
```ruby
merge(Spree::Backend::Config.promotions_index_scope).
```
?

backend/app/controllers/spree/admin/promotions_controller.rb
```diff
@@ -41,8 +41,9 @@ def collection
 
         @collection = super
         @search = @collection.ransack(params[:q])
-        @collection = @search.result(distinct: true).
+        @collection = @search.result.
```
eric1234: Crap, again my store not using multiple codes is biting me here as I'm not considering all the use cases. So basically:

* To allow searching by code we need the `distinct: true`.
* Having the `distinct: true` prevents us from sorting on custom criteria as anything in the `ORDER BY` must also be in the `SELECT`.

Option 1 - Something like `distinct: params[:q][:codes_value_cont].present?`. That way we are only doing `distinct` if searching on codes. Of course I would still need to resolve cases where the user is sorting on custom criteria but also searching on codes....
Option 2 - Drop this and the `promotion_references` completely as we get in a can of worms where one page has 10 extension points for every possible thing a developer might want to tweak. Are all those extension points more complex than just monkey-patching `collection`. Downside to this is I know solidus has a goal of discouraging monkey-patching.
Option 3 - Perhaps your `promotions_scope` hook could is the solution. Then I could also add a custom `select` that includes my custom sort (satisfying the SQL requirement that anything in `ORDER BY` also be in `SELECT` when using `DISTINCT`)

I think option 3 is the best. I can test that out to see if it works. Regarding monkey-patch vs config. If it was done as a config I wonder if we could do it at the Spree::ResourceController level. This way every resource controller has a hook to tweak it's scope to be customized. It could check for a config variable named after the controller name. If it doesn't exist then not tweaking needed.

mtomov: It seems that people have hit this `distinct` issue previously with product:

- https://github.com/solidusio/solidus/blob/master/core/app/models/spree/product/scopes.rb#L184-L207

kennyadsl:
Closing since there's no more activity here and probably this is not a feature needed by lot of solidus users. If we need it later we'll certainly keep this code as a starting point and as a reference about potential issues, especially with `distinct: true` in admin promotions controller.

PR #1809, opened 2017-03-29
Make our form/button styles match bootstrap's

jhawthorn:
This makes our forms (inputs and buttons) match those using bootstrap classes (`.btn` and `.form-control`). This should allow us to progressively change our current forms markup to bootstrap.

Similar to the changes in #1780, bootstrap variables are set to the legacy styles' values.

Buttons are styled as bootstrap buttons using `@extend`. `@extend` isn't an ideal tool to use, but I looked at the CSS products and it was okay here.

This necessitated one change: buttons hover color is now dark blue as it's [hardcoded in bootstrap](https://github.com/twbs/bootstrap/blob/v4-dev/scss/mixins/_buttons.scss#L7) to be 10% darker than the non-hovered BG. I hope this is okay. The new style is nicer. It currently disagrees with the select2 colours, but that will be fixed by #1797 

I tried using `@extend` for the form inputs, but the CSS produced was ridiculous. We also don't want some of the styles that would imply (like `width: 100%`). Instead I've just used the bootstrap variables here and pulled in their `placeholder` pseudo-element styling.

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
+++ b/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
@@ -289,8 +289,8 @@ $input-padding-y:                .375rem !default;
 $input-bg:                       #fff !default;
 $input-bg-disabled:              $gray-lighter !default;
 
-$input-color:                    $gray !default;
-$input-border-color:             #ccc !default;
+$input-color:                    $color-txt-text !default;
+$input-border-color:             $color-txt-brd !default;
 $input-btn-border-width:         $border-width !default; // For form controls and buttons
 $input-box-shadow:               inset 0 1px 1px rgba(0,0,0,.075) !default;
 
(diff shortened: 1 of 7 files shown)


backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
```diff
@@ -289,16 +289,16 @@ $input-padding-y:                .375rem !default;
 $input-bg:                       #fff !default;
 $input-bg-disabled:              $gray-lighter !default;
 
-$input-color:                    $gray !default;
-$input-border-color:             #ccc !default;
+$input-color:                    $color-txt-text !default;
```
jhawthorn: We've deprecated it, which means if a user sets `$color-txt-text`, they should see a warning, but it should continue working. This is the line that allows it to continue working

tvdeyen: Your are right. Ignore me

PR #1810, opened 2017-03-31
Fix styling of tiered promotions delete icon

jhawthorn:
Required for #1776

Previously this trash icon was being partially obscured by the row next to it, and only the left half of the icon could be clicked. It was also floating outside of the parent container. This commit fixes these issues
by adding a margin to the row, and placing the icon inside.

It also changes the cursor to a pointer on hover.

| Before | After |
|---|---|
|  |  |

Background red showing overlap. Trash can't be clicked inside the red.

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/sections/_promotions.scss
+++ b/backend/app/assets/stylesheets/spree/backend/sections/_promotions.scss
@@ -54,9 +54,14 @@
     position: relative;
     padding-bottom: 10px;
 
+    > .row {
+      margin-left: 10px;
+    }
+
     .remove {
       position: absolute;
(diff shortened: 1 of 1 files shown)


backend/app/assets/stylesheets/spree/backend/sections/_promotions.scss
```diff
@@ -54,9 +54,13 @@
     position: relative;
     padding-bottom: 10px;
 
+    >.row {
```
kennyadsl: I'd leave a space after `>`

jhawthorn:
Good call @kennyadsl. We might want to add a sass linter to catch these.

PR #1814, opened 2017-04-01
fix_display_orders for taxon

Perkir:
Althought i use  gem solidus_i18n, Display Order not work.
I think this commit fix it.


kennyadsl:
@Perkir Hello, what do you mean with "Display Order not work"? Can you please post steps to reproduce this issue or at least a screenshot. 
Also, in the code you are linking to the admin products page but with the "Orders" label, maybe you want to use `Spree::Product` instead?

Perkir:
Hello! 
I mean options in admin panel. 
Products -> Display Order

You're right i need to use Spree::Product.

Perkir:
nomethoderror at ru admin taxons 2017-04-02 22-06-20

This screen demonstration this bug

kennyadsl:
Can you please share which language (with `solidus_i18n`) and which solidus version are you using?

Perkir:
1.Language from  solidus_i18n its "ru"
2.Solidus version its 2.1.0

PR #1816, opened 2017-04-03
Upgrade to Bootstrap 4.0.0.alpha6

jhawthorn:
Previously we were on Bootstrap 4.0.0.alpha2. Changes can be seen in their [Release notes](https://github.com/twbs/bootstrap/releases)

Notable to us:
* `col-xs-3` et al. are gone. Replaced with `col-3`.
* Flexbox is on by default :+1:  (we had already turned on the optional flexbox support)
* Classes for breadcrumbs changed (easy fix)
* Some variables and mixins changed

I can't be 100% sure this didn't mess up any styling, but in a quick browse through the admin I didn't spot anything off.


backend/app/assets/stylesheets/spree/backend/shared/_typography.scss
```diff
@@ -1,5 +1,10 @@
 // Base
 //--------------------------------------------------------------
+
+html {
+  font-size: $font-size-root;
```
jhawthorn: I would like that, but I think we should do this in a separate PR. Removing this would move us from `13px` to `16px` (maybe), which is a really significant change, which I'm sure will cause some issues. I'd rather we look at that separately.

backend/vendor/assets/stylesheets/solidus_admin/bootstrap/_media.scss
```diff
@@ -1,90 +1,8 @@
...
-
-
-//
-// Media list variation
-//
-
-.media-list {
-  padding-left: 0;
-  list-style: none;
+.media-body {
```
jhawthorn: Right. I forgot this.

Bootstrap removed the `.media-left` class https://v4-alpha.getbootstrap.com/layout/media-object/ instead asking you to just specify `.mr-3`. :+1: Your padding solution is a good fix.

backend/app/assets/stylesheets/spree/backend/shared/_typography.scss
tvdeyen: As long as we set `$font-size-base` to `$body-font-size` everything should be fine.

jhawthorn: That changes the size of our `rem`, which I don't want to do. It's awkward that our base font size doesn't match a `rem`, making `rem` pretty useless if it doesn't match up with anything else on the page. If we're setting `$font-size-base` to a `px` value anyways we're not improving accessibility in any way.

I appreciate the argument that using the browser's default font size is a good thing. But if we aren't (and I think it's too big a change for this PR) the way bootstrap used to do it, and the way we're doing it here, by setting a size on the root element, is best.

jhawthorn:
@tvdeyen I've pulled in your changes other than changing the font size.

backend/app/assets/stylesheets/spree/backend/shared/_typography.scss
tvdeyen: Agreed!

PR #1818, opened 2017-04-03
Redesign product details page in admin

mtomov:
Hello, 

I've gone about to try out to implement #1290 instead of just criticizing on #1574. Initially, I went about to adjust what Graham had achieved thus far, but I ended up deleting and re-organizing most of the code, so have squashed, and only 2 commits are from the original.

The changes I've done reflect my following thoughts:

1. When introducing a new component, such as the admin section background, it is great to be able to associate it with an already existent component from Bootstrap, as this reduces the "learning curve" for other developers. So, instead of creating a new class `.form-section`, it's better to assign a default class from bootstrap and a custom one - `.card .card-spacial`. That way, a developer knows that what they are looking at is just a modification of an already known component.

    So, I've re-use the bootstrap class `.card-info`, as we have a light blue background color for the forms, though this might change, and we can go for a different name. I've gone to create a `card-outline-none`, to follow the logic for their other [`card-outline-*` classes](https://v4-alpha.getbootstrap.com/components/card/#outline-cards). `.input-group-addon` is actually unrelated, but fixes a horizontal alignment issue with those - can go in a separate PR.

2. I think that specifying a layout for every "new" form is not sustainable, and we should instead try to adapt the existing "old" layout to fit out case. 

    I've gone about to do just that, and there's actually not too much to do there. I've only deleted the `.container` specifications from Skeleton, so to rely on those provided by Bootstrap. This has the effect of increasing our container width from `960px` -> `1140px`, which has no negative consequences on any part of solidus backend, which uses the new bootstrap classes. I haven't checked any extension using the Skeleton classes, but I'd think that the `180px` difference would at large cause a very small input box to go a line up, which would be an insignificant "regression" in my opinion. Furthermore, we can prevent this "regression" by adjusting the skeleton widths (re-doing the column calculations with the new width).

    The existing "old" layout is actually meant to be our default layout, which I guess should be the  "max-width" version, and not the full-width one.

3. The specification says to align the "max-width" layout to the left due to tabs mis-behaviour, but I've discovered that this is questionable decision at this point:

    - left-aligned concentrates too much content on the left side where the Menu is located. The space between the Menu is not enough in my opinion.

    - on wide screen monitors, left-aligned, there's a huge space gap on the right, and the top right buttons are very disconnected. 

    - there's plenty of alternatives for the tabs - they can span full width and be centered; their border can span full width, but the tabs themselves can start from the content area.

    - I don't think that tabs will actually be present on any _full-width_ page, as tabs are generally Item edit specific, and full width pages are likely to be for an index page. See the Promotions#edit page - it's content is spread all over the place; in my view it's a mess, and needs to be put back to a _max-width_ one;
     In fact, I'd argue that there's not much advantage to having any full-width pages - looking at the orders index or the products index, which should be the heaviest tables of them all, they look better to me when _max-width_;

    Due to the possible controversy of the topic, I have left it to be implemented at any future PR, which would not have a negative impact on the currently implemented changes.

    It's actually very easy to play around to see how each options looks like. Just add `.container { margin-left: 0; }` in almost any css, and you'd see the left-aligned effect; personally, I really like the white space left and right of a _max-width_ container.

4. Unrelated directly, but I think that the [`admin_layout` feature](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/layouts/admin.html.erb#L7-L14) has already gone to be too complicated. It currently has a triple meaning of causing a "new-layout" class to come up, along with the specified class name of the "admin-layout", along with an `if` condition if the admin layout is specified or not, irrespective of its value. `new-layout` is a bit inappropriate, as when we do a v3 redesign, what would we call it then - `new-new-layout`?

    Unless I oversimplify, I think that this should be reduced to a boolean for `full-width`.

    In that respect, it's really great that you guys have not been rushing the UI changes, so that the "html api" can be polished up. Thanks for that!

5. I have moved the SKU to be below the Name and the Slug, contrary to the design, as it seems strange to me not to have the name on the upper row. Happy to change around.

6. I have followed the color implementations from the original PR #1574, and I don't have any color preferences, happy to even invert the colors completely as it is on the original design specs.

7. Have reduced the height around the logo a bit to make it match the breadcrumbs height.

8. Have added a carret down icon to all select2 input fields using a `before` css selector. I think the same approach should be used for the "calendar" icon.

Finally the screenhosts of a product without and with variants.

product details apache baseball jersey products 1

product details ruby on rails baseball jersey

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
+++ b/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
@@ -34,7 +34,7 @@ $gray-lightest:             #f7f7f9 !default;
 
 $brand-primary:             $color-3 !default;
 $brand-success:             $color-2 !default;
-$brand-info:                #5bc0de !default;
+$brand-info:                #eff5fc !default;
 $brand-warning:             $color-6 !default;
 $brand-danger:              $color-5 !default;
 
--- a/backend/app/assets/stylesheets/spree/backend/plugins/_select2.scss
+++ b/backend/app/assets/stylesheets/spree/backend/plugins/_select2.scss
@@ -90,3 +90,12 @@
     * { color: $dropdown-link-hover-color }
   }
 }
+
+.select2-choices::after {
+  content: "\f0d7"; // fa-caret-down
+  position: absolute;
+  font-family: "FontAwesome";
+  right: 10px;
+  bottom: 5px;
+  cursor: pointer;
+}
--- a/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
@@ -32,7 +32,7 @@ nav.menu {
 }
 
 .admin-nav-header {
-  padding: 15px;
+  padding: 8px 25px;
   height: 58px; // height of .page-title
   box-sizing: content-box;
   background-color: $color-1;
--- a/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
@@ -212,3 +212,15 @@ fieldset {
 .form-buttons {
   text-align: center;
 }
+
+.card.card-info {
+  border-color: darken($brand-info, 5%);
+}
+
+.input-group-addon {
+  line-height: initial;
+}
+
+.card-outline-none {
+  border: 0;
+}
--- a/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
@@ -29,11 +29,6 @@ body {
   body.new-layout & {
     @include padding(1rem $grid-gutter-width null);
   }
-
-  &.centered {
-    @include margin(null auto);
-    max-width: map-get($container-max-widths, "xl");
-  }
 }
 
 .content {
--- a/backend/app/assets/stylesheets/spree/backend/shared/_skeleton.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_skeleton.scss
@@ -21,7 +21,6 @@
 /* #Base 960 Grid
 ================================================== */
 
-    .container                                  { position: relative; width: 960px; margin: 0 auto; padding: 0; }
     .container .column,
     .container .columns                         { float: left; display: inline; margin-left: 10px; margin-right: 10px; }
     .row                                        { margin-bottom: 20px; }
@@ -77,7 +76,6 @@
     /* Note: Design for a width of 768px */
 
     @media only screen and (min-width: 768px + $width-sidebar) and (max-width: 959px + $width-sidebar) {
-        .container                                  { width: 768px; }
         .container .column,
         .container .columns                         { margin-left: 10px; margin-right: 10px;  }
         .column.alpha, .columns.alpha               { margin-left: 0; margin-right: 10px; }
@@ -130,7 +128,6 @@
     /* Note: Design for a width of 320px */
 
     @media only screen and (max-width: 767px + $width-sidebar) {
-        .container { width: 300px; }
         .container .columns,
         .container .column { margin: 0; }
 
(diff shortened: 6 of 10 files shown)


backend/app/assets/stylesheets/spree/backend/plugins/_select2.scss
```diff
@@ -90,3 +90,12 @@
     * { color: $dropdown-link-hover-color }
   }
 }
+
+.select2-choices::after {
```
jhawthorn: It also looks off on pages where the input expands to several lines.

backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
```diff
@@ -34,7 +34,7 @@ $gray-lightest:             #f7f7f9 !default;
 
 $brand-primary:             $color-3 !default;
 $brand-success:             $color-2 !default;
-$brand-info:                #5bc0de !default;
+$brand-info:                #eff5fc !default;
```
jhawthorn: I don't this is an appropriate variable to store the colour for the card backgrounds. This value should be usable for buttons, alerts, labels etc. Like in [this bootstrap example](https://v4-alpha.getbootstrap.com/components/buttons/#examples).

This won't work if we use it as a common background because it will totally blend in. It's also too light for use outside. 

 (`btn-info` on the right)

This should probably be in a dedicated variable.

tvdeyen:
I do appreciate all the hard work that has been put into this. Thanks for that. 🥂 

### But

I think we have to tackle some other UI problems first before we start to make changes to this form. As we can see on the screenshot very well, we need to get rid of this centred strip like thingy first. 

Especially on large screens this is not very efficient UI. Moving the form to the right of the screen could be helpful for this. Having a proper space for additional / secondary information is another one to be discussed before going any further with this IMO.

Besides that I don't think dark background and boxes are a great UI for this forms. I know [I was excited at first](https://github.com/solidusio/solidus/pull/1574#issuecomment-258250190), but as I revisit this today I doubt this whole coloured boxy layout is an actual improvement for us. I like the grouping of fields from an UX standpoint, but putting them in dark boxes is not something I would consider a modern and appealing UI. 

We could make the boxes white and make use of [Bootstrap card headers](https://v4-alpha.getbootstrap.com/components/card/#header-and-footer) to slightly improve this, but I'm still not 100% confident that this is the best we can do here:

product details - apache baseball jersey - products 2017-04-05 10-55-00

All of this is another discussion we should continue after we addressed the concerns from above at first.

Again, I really appreciate the work put into this, but for now this is a 👎 from me, sorry.

Please don't feel personally offended Martin, as this is not meant as an offence against you or others involved in this. This is UI, one of the hardest thing in our job.

tvdeyen:
> ... the admin_layout feature has already gone to be too complicated ... I think that this should be reduced to a boolean for full-width.

👍 

There is some stuff in this PR that I would like to see as separate PRs:

- [ ] Reduce the height around the logo a bit to make it match the breadcrumbs height
- [ ] Remove `.container` definitions from Skeleton and use those from Bootstrap

mtomov:
Oh, not to worry, _Request changes_ is my favourite button : ) In fact, the larger purpose of this PR was indeed to change the direction we were going with the css container structure and to overall express some general guidelines that I think we should follow : )

And, in terms of effort - I just know that reviewing all this, and trying it out, and writing out the problems is much more time consuming then the few lines of code I have changed, so thank you both!

@jhawthorn,

- `input-group-addon` - yeap, saw later that you mentioned that it was fixed by #1816

- the caret - I think I might be able to fix the multi-line bug by specifying it on the upper class, but I agree with you that more discussion is probably needed. 

- `brand-info` - very well spotted; it is indeed inappropriate to change the value of that variable. Better off as a new variable.


@tvdeyen,

> Besides that I don't think dark background and boxes are a great UI for this forms.

I share your opinion on that. I'm also in doubt about those default Bootstrap cards with headers. Better to leave it all as it is until we have a solid plan. I guess we might cover that at the conference. 


---

I've opened #1822 and #1823 as well suggested. Thanks indeed! 

I think we might as well close this and append the screenshots to #1290 as feedback.

kennyadsl:
Closing this one since most of its content has been moved/handled into other PRs (#1822 and #2127)

PR #1820, opened 2017-04-04
Add a configurable order number generator

tvdeyen:
The default order number generator was extracted into a class.
Provide your own class that takes an options hash and returns a
string when called for `#generate`.

    Spree::Config.order_number_generator = TrueNumberGenerator

    class TrueNumberGenerator
      def initialize(options = {})
      end

      def generate
        '42'
      end
    end

### DIFF
--- a/core/app/models/spree/app_configuration.rb
+++ b/core/app/models/spree/app_configuration.rb
@@ -397,6 +397,17 @@ module Spree
       @current_store_selector_class ||= Spree::CurrentStoreSelector
     end
 
+    # Allows providing your own class instance for generating order numbers.
+    #
+    # @!attribute [rw] order_number_generator
+    # @return [Class] a class instance with the same public interfaces as
+    #   Spree::Order::NumberGenerator
+    # @api experimental
+    attr_writer :order_number_generator
+    def order_number_generator
+      @order_number_generator ||= Spree::Order::NumberGenerator.new
+    end
+
     def static_model_preferences
       @static_model_preferences ||= Spree::Preferences::StaticModelPreferences.new
     end
--- a/core/app/models/spree/order/number_generator.rb
+++ b/core/app/models/spree/order/number_generator.rb
@@ -1,0 +1,43 @@
+module Spree
+  # Generates order numbers
+  #
+  # In order to change the way your order numbers get generated you can either
+  # set your own instance of this class in your stores configuration with different options:
+  #
+  #     Spree::Config.order_number_generator = Spree::Order::NumberGenerator.new(
+  #       prefix: 'B',
+  #       lenght: 8,
+  #       letters: false
+  #     )
+  #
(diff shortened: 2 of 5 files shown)


core/app/models/spree/order/number_generator.rb
```diff
@@ -0,0 +1,28 @@
...
+      possible = (0..9).to_a
+      possible += ('A'..'Z').to_a if letters
+
+      loop do
+        # Make a random number.
+        random = "#{prefix}#{(0...length).map { possible.sample }.join}"
+        # Use the random number if no other order exists with it.
+        if Spree::Order.exists?(number: random)
+          # If over half of all possible options are taken add another digit.
+          length += 1 if Spree::Order.count > (10**length / 2)
```
jhawthorn: This is a good catch by hound. I think there's an issue here between `length`  being a instance method and being assigned here as a variable.

jhawthorn:
I like this. It would be even better if we could share the code to generate Order numbers, Shipment numbers, and Payment numbers

tvdeyen:
> share the code to generate Order numbers, Shipment numbers, and Payment numbers

@jhawthorn I also thought about this, but I think we should not make the assumption that all numbers generate the same way. 

The only idea coming to my mind is the ask the `NumberGenerator` for `#order_number`, `#shipment_number`, `#payment_number` and so on. WDYT?

core/app/models/spree/order/number_generator.rb
tvdeyen: Oh, actually a good catch. So, the only solution I see is to use a `attr_accessor` for `length` then, right?

tvdeyen:
@jhawthorn addressed the issues. 

Still think this is valuable as order number generator on its own. Shipments and friends get their number through `permalinks.rb` and are not that visible to customers as the order number. 

Let's address this - totally valid concern - in a separate PR, ok?

core/app/models/spree/app_configuration.rb
```diff
@@ -365,6 +365,17 @@ def tax_adjuster_class
...
 
+    # Allows providing your own class for generating order numbers.
+    #
+    # @!attribute [rw] order_number_generator
+    # @return [Class] a class with the same public interfaces as
+    #   Spree::Order::NumberGenerator
+    # @api experimental
+    attr_writer :order_number_generator
+    def order_number_generator
+      @order_number_generator ||= Spree::Order::NumberGenerator
```
jhawthorn: One thought: maybe this should be an instance of `NumberGenerator` rather than the class. This would allow developers to specify a length, prefix, or letters without needing to write their own class

tvdeyen:
@jhawthorn good idea. I needed to remove the possibility for passing `options` into `Spree::Order#generate_order_number`, though. But I think this was barely used anyway.

PR #1822, opened 2017-04-05
Change admin logo height to match breadcrumbs height

mtomov:
~~The image itself seemed to contain white space underneath, so have modified the padding values accordingly to vertically center it.~~

_edit_: uses _line-height_ now, see conversation.

Removed a bit of css as well. 

- _Before_
  admin logo before

- _After_
  admin logo - after

Still looks a bit boxed, but oh well : )

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
@@ -32,22 +32,9 @@ nav.menu {
 }
 
 .admin-nav-header {
-  padding: 15px;
-  height: 58px; // height of .page-title
-  box-sizing: content-box;
-  background-color: $color-1;
+  padding: 14px 0 8px;
   border-bottom: 1px solid $color-border;
-
-  a {
-    display: inline-block;
-    line-height: 58px;
-  }
-
-  img {
-    vertical-align: middle;
-    max-width: 100%;
-    max-height: 100%;
-  }
+  text-align: center;
 }
 
 .admin-nav-menu {
--- a/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
@@ -32,9 +32,11 @@ nav.menu {
 }
 
 .admin-nav-header {
-  padding: 14px 0 8px;
   border-bottom: 1px solid $color-border;
   text-align: center;
+  // Using line height for proper vertical centering.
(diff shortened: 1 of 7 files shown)


backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -32,22 +32,9 @@ nav.menu {
 }
 
 .admin-nav-header {
-  padding: 15px;
-  height: 58px; // height of .page-title
-  box-sizing: content-box;
-  background-color: $color-1;
+  padding: 14px 0 8px;
```
tvdeyen: The solution I came up with

```scss
.admin-nav-header {
  padding: 0;
  border-bottom: 1px solid $color-border;
  text-align: center;
  // Using line height for proper vertical centering.
  // As line height does not take the border width into account we need to subtract it.
  line-height: $main-header-height - 1px;
}
```

tvdeyen: We should also use the (to be introduced) `$main-header-height` variable to set the height on the `.main-header`

```scss
$main-header-height:             75px !default;
```

```scss
.main-header {
  ...
  height: $main-header-height;
  ...
}
```

That fixes the height for empty main headers, in the login screen for instance

### Before

user sessions 2017-04-05 15-24-36

### After

user sessions 2017-04-05 15-25-09

backend/app/views/spree/admin/shared/_navigation_header.html.erb
```diff
@@ -1,5 +1,5 @@
 <header class="admin-nav-header">
-  <%= link_to spree.admin_path, class: 'brand-link' do %>
```
tvdeyen: Shops may have used this class for deface overrides. We should not remove it now.

mtomov:
Super, 

actually no logo resize was necessary after the line-height change .. which is good, as I didn't want to mess the frontend somehow.

I've allowed myself one more try with an automatic image resize in d2d0b89 . Reason is that I think that nowadays all images should be responsive, like in [foundation](https://github.com/zurb/foundation-sites/blob/develop/scss/_global.scss#L154). I'm not sure to what extend that would apply for such a project, but I've browsed through some product images, and everything looks normal, so it's pending your decision on this "global" one.

- `brand-link` - well pointed out

- line-height variable also fixed the login screen, well thought out :+1: 


There is nothing much different than before, but posting a screen nevertheless:

solidus logo after

PR #1823, opened 2017-04-05
Increase admin central area by 180px

mtomov:
by removing the container css specifications from Skeleton
  in favour of using Bootstrap's definition.

  This change will increase the width of the admin central area
  by 180px - from 960px to 1140px

  This will not have adversary effects to any part of the Solidus admin,
  as it has been fully converted to use the Bootstrap column classes,
  which are percentage based.

  Old Skeleton layouts might be negatively effected only in the following
  case:

   > 2 rows of form fields on large - the first row leaves 3 or more Skeleton
      columns empty (as 3 columns are 160px defined in [Skeleton](https://github.com/mtomov/solidus/blob/f8d32db1401dd8a9fc446a4adada70cc79d7445f/backend/app/assets/stylesheets/spree/backend/shared/_skeleton.scss#L36)), and 
      the second row starts with a field, which is 3
      or less columns wide. Then, the field from the second column would
      jump up on the previous line.

  As this is a very unlikely and minor regression, it is better to have it,
  then to re-calculate the Skeleton column widths, as doing so might cause
  in itself more trouble.

   Widths smaller than large are unaffected, as the Skeleton container widths 
match  with Bootstrap's

- _Before_

  increase container width - before

- _After_

  increase container width - after

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/shared/_skeleton.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_skeleton.scss
@@ -204,27 +204,6 @@
 
 /* #Clearing
 ================================================== */
-
-    /* Self Clearing Goodness */
-
-    /* Use clearfix class on parent to clear nested columns,
-    or wrap each row of columns in a <div class="row"> */
-    .clearfix:before,
-    .clearfix:after {
-      content: '\0020';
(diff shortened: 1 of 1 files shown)


mtomov:
Superseded by #2127

PR #1824, opened 2017-04-05
make the filter more modern

shcyiza:
I've change the input tag that was use for the check box into a check_box_tag method.
I've also added a sumbit onclick option in the method and erase the the sumbit button of the form to have the type of behavior consumer usually find in the filters of the e-commerce nowadays.

i hope u like it ;)

### DIFF
--- a/frontend/app/views/spree/shared/_filters.html.erb
+++ b/frontend/app/views/spree/shared/_filters.html.erb
@@ -11,17 +11,14 @@
           <% labels.each do |nm,val| %>
             <% label = "#{filter[:name]}_#{nm}".gsub(/\s+/,'_') %>
             <li class="nowrap">
-              <input type="checkbox"
-                     id="<%= label %>"
-                     name="search[<%= filter[:scope].to_s %>][]"
-                     value="<%= val %>"
-                     <%= params[:search] && params[:search][filter[:scope]] && params[:search][filter[:scope]].include?(val.to_s) ? "checked" : "" %> />
+            <%= check_box_tag "search[#{filter[:scope].to_s}][]", val,
+             params[:search] && params[:search][filter[:scope]] && params[:search][filter[:scope]].include?(val.to_s) ? true : false,
+             id: label , onclick: 'this.form.submit()' %>
               <label class="nowrap" for="<%= label %>"> <%= nm %> </label>
             </li>
           <% end %>
         </ul>
       </div>
     <% end %>
-    <%= submit_tag Spree.t(:search), name: nil %>
   <% end %>
 <% end %>


frontend/app/views/spree/shared/_filters.html.erb
```diff
@@ -11,17 +11,14 @@
...
+            <%= check_box_tag "search[#{filter[:scope].to_s}][]", val,
+             params[:search] && params[:search][filter[:scope]] && params[:search][filter[:scope]].include?(val.to_s) ? true : false,
+             id: label , onclick: 'this.form.submit()' %>
               <label class="nowrap" for="<%= label %>"> <%= nm %> </label>
             </li>
           <% end %>
         </ul>
       </div>
     <% end %>
-    <%= submit_tag Spree.t(:search), name: nil %>
```
cbrunsdon: I think this button was removed accidentally? Its needed to submit the form, and thats what made the tests fail. If you add it back in, I think the tests should start passing.

kennyadsl: I think submit is done via checkbox tag `onclick: 'this.form.submit()'` which I don't like a lot. It should be done in a javascript file but I'd avoid it completely because each store can implement that logic with a couple of lines of code if needed.

frontend/app/views/spree/shared/_filters.html.erb
```diff
@@ -11,17 +11,14 @@
           <% labels.each do |nm,val| %>
             <% label = "#{filter[:name]}_#{nm}".gsub(/\s+/,'_') %>
             <li class="nowrap">
-              <input type="checkbox"
-                     id="<%= label %>"
-                     name="search[<%= filter[:scope].to_s %>][]"
-                     value="<%= val %>"
-                     <%= params[:search] && params[:search][filter[:scope]] && params[:search][filter[:scope]].include?(val.to_s) ? "checked" : "" %> />
+            <%= check_box_tag "search[#{filter[:scope].to_s}][]", val,
```
kennyadsl: This line does not seem to be indented properly. It is inside the `<li>` element so it should have 2 more spaces, right?

kennyadsl: I think the best way to indent this checkbox params is using this [style guide](https://github.com/thoughtbot/guides/blob/master/style/erb/sample.html.erb):

```erb
<%= check_box_tag(
  param_1,
  param_2,
  etc
) %>
```

frontend/app/views/spree/shared/_filters.html.erb
shcyiza: kennyadsl is right my intend was to sumbit the form with the `onclick: 'this.form.submit()'` option... But it's no big deal... Eventhought the ecommerce design don't have no sumbit button for their filters... That's why i implemented it.

mamhoff:
I would rather keep this old-fashioned and working without JS. If stores want to do this, it'll take a few lines of code as @kennyadsl notes.

gmacdougall:
Given the current feedback, I don't think this is a change we're looking to make. There are failing tests and some additional JS that we're not looking to add.

Thanks for the contribution on this and please don't hesitate to send additional ideas our way.

PR #1826, opened 2017-04-05
Add teaspoon javascript test runner to backend

jhawthorn:
Our backend JS is growing, but at the same time is becoming more structured. It was suggested by @braidn in our slack that we add JavaScript unit tests, which we should.

This PR adds the [teaspoon](https://github.com/jejacks0n/teaspoon) Javascript test runner for Rails. Mocha and chai are used, which gives us a somewhat rspec-like syntax to write specs in. When invoked on the command line specs are run in a phantomjs browser, but can also be manually run in a browser.

```
describe('Foo', function() {
  it('is foo', function() {
    expect('foo').to.equal('foo');
  });
});
```

The purpose for these tests is to test JavaScript independently from our ruby/rails code. When integration testing we should still rely on capybara.

## Todo

* [x] Run teaspoon on CI, and as part of `build.sh`
* [x] Write at least one real spec
* [x] Check that teaspoon works under sprockets 4.0 (current gem release doesn't but git master does. Good enough for now.)
* [x] Make sure teaspoon raises syntax errors

### DIFF
--- a/backend/Gemfile
+++ b/backend/Gemfile
@@ -4,3 +4,6 @@ gem 'solidus_core', path: '../core'
 gem 'solidus_api', path: '../api'
 
 gemspec
+
+gem 'teaspoon'
+gem 'teaspoon-mocha'
--- a/backend/Rakefile
+++ b/backend/Rakefile
@@ -6,7 +6,36 @@ require 'spree/testing_support/common_rake'
 
 RSpec::Core::RakeTask.new
 
-task default: :spec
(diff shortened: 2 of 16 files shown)


jhawthorn:
This is ready for review. Teaspoon is run on CircleCI and should be able 

I've added some basic specs for some JS I'd written recently `Views.NumberWithCurrency`, `formatMoney`, and `Spree.t`

PR #1827, opened 2017-04-06
Fix datepicker style

kennyadsl:
Probably #1777 introduced some date-picker visual bugs.

This PR:

- adjusts width of the whole datepicker to adapt to its content, was: [`width: 17em`](https://github.com/solidusio/solidus/blame/master/backend/vendor/assets/stylesheets/jquery-ui.datepicker.css.erb#L299);
- gives more spaces to arrows that currently are not completely shown. Changes width from 10px to 13px.

**Before**

datepicker-before

**After**

datepicker-after


tvdeyen:
Thanks

PR #1833, opened 2017-04-07
Fix Solidus' behaviour when confirm state does not exist

mamhoff:
For stores that redefine their state machine in such a way that you can
transition to complete from other states than confirm (a requirement
that happens e.g. when not having a confirm step, and orders transition
directly from payment), we can not rely on the order's confirm state.
Instead, we can use state machines' `can_complete?` method, which
checks whether the `complete` transition can be executed from the
current state.

For Solidus' default state machine, this change does not change anything.

### DIFF
--- a/backend/app/controllers/spree/admin/orders_controller.rb
+++ b/backend/app/controllers/spree/admin/orders_controller.rb
@@ -98,7 +98,7 @@ module Spree
         else
           @order.contents.advance
 
-          if @order.confirm?
+          if @order.can_complete?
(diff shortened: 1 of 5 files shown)


jhawthorn:
I think @jordan-brough wanted to note that this isn't an endorsement for removing the `confirm` step. Though it can be done, it can cause issues unless care is taken.

Regardless `can_complete?` reads better than `confirm?`, and better expresses the intent of the calling code. :+1:

PR #1834, opened 2017-04-07
Only load active payment methods in the backend

luukveenis:
When the `display_on` column was removed and we switched to `available_to_users`/`available_to_admin`, we stopped using the `available` scope, which was restricted to active payment methods only.

The new scopes corresponding to the `available_to_*` columns are not restricted to active payment methods only, so we need to explicitly add that scope now.


luukveenis:
I added these commits to #1801 since they address the same issue.

PR #1835, opened 2017-04-07
Fix outstanding balance calculation with refunds

### LINKED ISSUE #1475: Refunding Doesn't Update Credit Owed

ericsaupe:
payment_total includes all refunds on the payments as well so there
is no reason to recalculate that in #outstanding_balance

Fixes #1475

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -365,14 +365,10 @@ module Spree
     deprecate create_tax_charge!: :update!, deprecator: Spree::Deprecation
 
     def outstanding_balance
-      # If reimbursement has happened add it back to total to prevent balance_due payment state
-      # See: https://github.com/spree/spree/issues/6229
-      adjusted_payment_total = payment_total + refund_total
-
       if state == 'canceled'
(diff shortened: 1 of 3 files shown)


skukx:
For more detailed explanation see: https://github.com/solidusio/solidus/issues/1475#issuecomment-292367955

ericsaupe:
@jhawthorn updated with your regression test from #1926

jhawthorn:
Thanks Eric and Taylor. Closing in favour of #2002, which includes the specs from this.

PR #1840, opened 2017-04-11
Fix background of nav header

jhawthorn:
In #1838 we lost the background colour of the nav header.

**Before**

**After**


jhawthorn:
Bypassing our normal 24-hour wait for reviews so that this can be included in 2.2.0.beta1.

PR #1848, opened 2017-04-19
Reorganize Image upload JS

jhawthorn:
This converts the image upload . The existing code was quite good so the changes required weren't too substantial.

The most important change was to decouple the Backbone views from the DOM (ourside of their `el`) and from each other, instead communicating through models and collections. This allows the Backbone Model and Views to be defined before the dom is loaded, which is considered the best practice.

This also includes some tweaks:
* Swap out `<progress>` element for bootstrap's progress, which looks better and is more consistent.
* Always display filename and upload size (was inconsistent previously)
* Remove browser feature tests since we only support modern browsers. Detecting these failures doesn't improve user experience.
* Use bootstrap utility class instead of custom `margin-bottom-one`

Look is changed very slightly (progress bar look and file size)

### DIFF
--- a/backend/app/assets/javascripts/spree/backend.js
+++ b/backend/app/assets/javascripts/spree/backend.js
@@ -36,7 +36,6 @@
 //= require spree/backend/gateway
 //= require spree/backend/handlebars_extensions
 //= require spree/backend/images/index
-//= require spree/backend/images/new
 //= require spree/backend/images/upload
 //= require spree/backend/navigation
 //= require spree/backend/number_field_updater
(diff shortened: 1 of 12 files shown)


mtomov:
Cool, thanks for the refactoring! I will be able to learn proper backbone organization. The size bit change is much cleaner indeed as well.

Btw, should I refactor #1580 to work with the new code, or the direction there is not appropriate?

jhawthorn:
@mtomov Thanks! This was an easy change because the existing code was very good.

Issue #1836, opened 2017-04-10
TaxRate <--> TaxCategory relation type


vladstoick:
# Background:
- A TaxRate belongs to a TaxCategory
- A Product/ Variant belongs to a Tax Category
# Problem:
When adding multiple categories you end up duplicating a lot of tax rates as VAT is applied differently across Europe based on the TaxCategory. For instance some countries have special tax reduction for certain tax categories, while other don't. A rate must exist for each category if you want to properly apply VAT:
```
def rates_for_item(item)
  applicable_rates(item.order).select { |rate| rate.tax_category_id == item.tax_category_id }
end
```
This means that even if a tax category doesn't get a special rate for that category you still end up creating a tax rate for that category and that zone. This can lead to the creation of a lot of TaxRate even when they are not necessary as the general VAT can apply to them. This can become problematic for instance when implementing EU VAT where each country (28 at the moment) has different rules (in our (LMN) case books have a VAT of 0% in some countries, while in others they might be fully taxed or other taxes might be implemented).

# Potential solution
Allow a TaxRate to have many categories if the rate is the same (the general VAT tax for instance) and make additional TaxRates required if the rate is going to be different.

mamhoff:
Go for it. I see a challenge in administering this nicely in the backend, but from a data architecture standpoint this makes a whole lot of sense. :+1:

PR #1851, opened 2017-04-21
TaxRate M-M TaxCategory

vladstoick:
Related to issue https://github.com/solidusio/solidus/issues/1836

- Transform TaxRate - TaxCategory in a many to many relation via TaxRateTaxCategory and update  all places that used `tax_rate.tax_category_id`.
- Update UI to use multiple from select2
image
- Fix all specs that use factory for tax_rate and update it to `tax_categories: []`

### DIFF
--- a/backend/app/views/spree/admin/tax_rates/index.html.erb
+++ b/backend/app/views/spree/admin/tax_rates/index.html.erb
@@ -29,7 +29,7 @@
       <tr data-hook="rate_header">
         <th><%= Spree::TaxRate.human_attribute_name(:zone) %></th>
         <th><%= Spree::TaxRate.human_attribute_name(:name) %></th>
-        <th><%= Spree::TaxRate.human_attribute_name(:tax_category_id) %></th>
+        <th><%= Spree::TaxRate.human_attribute_name(:tax_categories) %></th>
         <th><%= Spree::TaxRate.human_attribute_name(:amount) %></th>
         <th><%= Spree::TaxRate.human_attribute_name(:included_in_price) %></th>
         <th><%= Spree::TaxRate.human_attribute_name(:show_rate_in_label) %></th>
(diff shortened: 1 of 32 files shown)


backend/app/views/spree/admin/tax_rates/index.html.erb
```diff
@@ -42,7 +42,13 @@
       <tr id="<%= spree_dom_id tax_rate %>" data-hook="rate_row" class="<%= cycle('odd', 'even')%>">
         <td class="align-center"><%=tax_rate.zone.try(:name) || Spree.t(:not_available) %></td>
         <td class="align-center"><%=tax_rate.name %></td>
-        <td class="align-center"><%=tax_rate.tax_category.try(:name) || Spree.t(:not_available) %></td>
+        <td class="align-center">
+          <% if tax_rate.tax_categories.any? %>
+            <%= tax_rate.tax_categories.map(&:name).join(", ") %>
```
mamhoff: Here, you might want to use an internationalized separator, as not all locales will work with a comma separated list. Most will though, so I won't insist. :)

core/db/migrate/20170412103617_transform_tax_rate_category_relation.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def down
+    add_column :spree_tax_rates, :tax_category_id, :integer, index: true
+    add_foreign_key :spree_tax_rates, :spree_tax_categories, column: :tax_category_id
+
+    TaxRate.find_each do |tax_rate|
+      tax_category_ids = TaxRateTaxCategory.where(tax_rate_id: tax_rate.id).pluck(:tax_category_id)
+
+      if tax_category_ids.count == 1
```
jordan-brough: Nitpick: Don't need this `if` right? The logic in the `else` would handle it?

core/app/models/spree/tax_category.rb
```diff
@@ -4,7 +4,9 @@ class TaxCategory < Spree::Base
     validates :name, presence: true
     validates_uniqueness_of :name, unless: :deleted_at
 
-    has_many :tax_rates, dependent: :destroy, inverse_of: :tax_category
+    has_many :tax_rate_tax_categories, class_name: Spree::TaxRateTaxCategory, dependent: :destroy
```
jordan-brough: Since `inverse_of` is currently broken in Solidus it's probably worth adding it to these since we're starting fresh with them?

PR #1852, opened 2017-04-21
Add expiration to Spree::StoreCredit

luukveenis:
We currently have the notion of `expiring` and `non-expiring` store credits, but no means of telling whether a store credit _has_ expired. This PR adds an `expires_at` column to store credits and ensures only non-expired store credits are applied during checkout.

I tried to tie this in to the current system by adding a validation that requires the expiration date be set if the store credit has an expiring `Spree::StoreCreditCategory`. If the store credit is a non-expiring type, the validation is skipped and `nil` expiration values are treated as "never expires".

I added a datepicker to the store credit form to set the expiry:
store_credit_new

example of the validation error:
store_credit_error

I also displayed the expiration on the index/show pages:
store_credit_index_expiration

store_credit_edit_expiration

### DIFF
--- a/core/app/models/spree/store_credit.rb
+++ b/core/app/models/spree/store_credit.rb
@@ -23,6 +23,7 @@ class Spree::StoreCredit < Spree::PaymentSource
   validates_numericality_of :amount_used, { greater_than_or_equal_to: 0 }
   validate :amount_used_less_than_or_equal_to_amount
   validate :amount_authorized_less_than_or_equal_to_amount
+  validates :expires_at, presence: true, unless: ->(credit) { credit.category.non_expiring? }
 
   delegate :name, to: :category, prefix: true
   delegate :email, to: :created_by, prefix: true
@@ -231,6 +232,7 @@ class Spree::StoreCredit < Spree::PaymentSource
       created_by_id: created_by_id,
       currency: currency,
       type_id: type_id,
+      expires_at: expires_at,
       memo: credit_allocation_memo
     }
   end
--- a/core/app/models/spree/store_credit.rb
+++ b/core/app/models/spree/store_credit.rb
@@ -27,6 +27,7 @@ class Spree::StoreCredit < Spree::PaymentSource
   delegate :name, to: :category, prefix: true
   delegate :email, to: :created_by, prefix: true
 
+  scope :active, -> { where(expires_at: nil).or(where("expires_at > ?", DateTime.current)) }
   scope :order_by_priority, -> { includes(:credit_type).order('spree_store_credit_types.priority ASC') }
 
   after_save :store_event
--- a/core/app/models/spree/store_credit.rb
+++ b/core/app/models/spree/store_credit.rb
@@ -171,6 +171,10 @@ class Spree::StoreCredit < Spree::PaymentSource
     !amount_remaining.zero?
   end
 
+  def expired?
+    expires_at.present? && expires_at < Time.current
+  end
+
   def invalidateable?
     !invalidated? && amount_authorized.zero?
   end
--- a/core/app/models/spree/store_credit_category.rb
+++ b/core/app/models/spree/store_credit_category.rb
@@ -1,6 +1,12 @@
 class Spree::StoreCreditCategory < Spree::Base
-  class_attribute :non_expiring_credit_types
-  self.non_expiring_credit_types = [Spree.t("store_credit.non_expiring")]
+  class_attribute :non_expiring_credit_categories
+  self.non_expiring_credit_categories = [Spree.t("store_credit_category.default")]
+
+  class << self
+    alias_attribute :non_expiring_credit_types, :non_expiring_credit_categories
+    deprecate non_expiring_credit_types: :non_expiring_credit_categories, deprecator: Spree::Deprecation
+    deprecate :non_expiring_credit_types= => :non_expiring_credit_categories=, deprecator: Spree::Deprecation
+  end
 
   class_attribute :reimbursement_category_name
   self.reimbursement_category_name = Spree.t("store_credit_category.default")
@@ -11,6 +17,6 @@ class Spree::StoreCreditCategory < Spree::Base
   end
 
   def non_expiring?
-    self.class.non_expiring_credit_types.include? name
+    self.class.non_expiring_credit_categories.include?(name)
   end
 end
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -12,6 +12,7 @@ module Spree
     before_action :ensure_checkout_allowed
     before_action :ensure_sufficient_stock_lines
     before_action :ensure_valid_state
+    before_action :ensure_store_credits_eligible
 
     before_action :associate_user
     before_action :check_authorization
@@ -121,6 +122,18 @@ module Spree
       if params[:state] == "confirm" && @order.payment_required? && @order.payments.valid.empty?
         flash.keep
         redirect_to checkout_state_path("payment")
+      end
+    end
+
+    def ensure_store_credits_eligible
+      expired_payments = @order.payments.valid.store_credits.select { |p| p.source.expired? }
+
+      if expired_payments.any?
+        @order.payments.delete(*expired_payments)
+        @order.restart_checkout_flow
+        @order.update!
+        flash[:error] = Spree.t(:store_credits_expired_error)
+        redirect_to checkout_state_path(@order.state)
       end
     end
 
--- a/backend/app/controllers/spree/admin/store_credits_controller.rb
+++ b/backend/app/controllers/spree/admin/store_credits_controller.rb
@@ -68,7 +68,7 @@ module Spree
       private
 
       def permitted_resource_params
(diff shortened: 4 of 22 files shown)


luukveenis:
@adammathys thanks for the feedback :grin: 

1. The date picker fields in the admin submit the date as `yyyy/mm/dd` which gets parsed to a datetime as beginning of day. Since `Spree::Promotion` already has an `expires_at` field, which uses the default beginning of day behaviour, I thought it best to be consistent with that.
2. I added a similar check to promotions on the transition to `complete`. If the credit has expired, the store credit payments are removed and the checkout flow is restarted with an error message displayed to the user.
3. The current behaviour when reimbursing an order that was paid for with store credit is to "credit" the balance back to the existing `Spree::StoreCredit` record instead of creating a new one, so it would keep the same expiration date. I think this makes sense, but any feedback is appreciated.

core/app/models/spree/order/checkout.rb
```diff
@@ -112,6 +112,7 @@ def define_state_machine!
               before_transition to: :complete, do: :ensure_available_shipping_rates
             end
             before_transition to: :complete, do: :ensure_promotions_eligible
+            before_transition to: :complete, do: :ensure_store_credits_eligible
```
cbrunsdon: Anything that adds a transition I'm not a huge fan of, will talk offline though.

frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -124,6 +125,18 @@ def ensure_valid_state
       end
     end
 
+    def ensure_store_credits_eligible
+      expired_payments = @order.payments.valid.store_credits.select { |p| p.source.expired? }
```
jhawthorn: This amount of logic is awkward in a controller, especially since it's tied to the frontend. What about orders made through the admin, api, or a custom frontend?

core/app/models/spree/store_credit.rb
```diff
@@ -23,10 +23,12 @@ class Spree::StoreCredit < Spree::PaymentSource
   validates_numericality_of :amount_used, { greater_than_or_equal_to: 0 }
   validate :amount_used_less_than_or_equal_to_amount
   validate :amount_authorized_less_than_or_equal_to_amount
+  validates :expires_at, presence: true, unless: ->(credit) { credit.category.non_expiring? }
```
jhawthorn: Are we sure this should be a required field? Seems like it should only be allowed on a `non_expiring` category, but optional.

core/app/models/spree/store_credit_category.rb
```diff
@@ -1,6 +1,6 @@
 class Spree::StoreCreditCategory < Spree::Base
-  class_attribute :non_expiring_credit_types
-  self.non_expiring_credit_types = [Spree.t("store_credit.non_expiring")]
+  class_attribute :non_expiring_credit_categories
```
jhawthorn: Why has this changed?

If it must change, the old methods should be aliased with deprecation warnings.

core/app/models/spree/store_credit.rb
luukveenis: Sorry, I don't follow why it should only be allowed on a `non_expiring` category?

My thinking was that if we have an `expiring` category, one would expect those credits to have an expiration date (otherwise they wouldn't be expiring). Similarly if a credit is `non_expiring` it should not have an expiration date. If anything I would consider adding a validation that a `non_expiring` credit does not have an expiration date set.

core/app/models/spree/store_credit_category.rb
luukveenis: I changed the naming on this because it was confusing. We have both a `Spree::StoreCreditCategory` and a `Spree::StoreCreditType` model.

This attribute is called `non_expiring_credit_types`, but its value is set to the default `Spree::StoreCreditCategory` name, not a `Spree::StoreCreditType`. Also in the `non_expiring?` method, it checks if the current `CreditCategory`'s name is in the array, so calling it a credit type is misleading.

I found this confusing while working on it, but if you prefer to keep it as is please let me know. Otherwise I'll add deprecation warnings.

frontend/app/controllers/spree/checkout_controller.rb
luukveenis: I discussed this with @cbrunsdon offline because I initially had this in a `before_transition to: :complete`. He brought up that there would likely be scenarios where admins would want to push through an order with expired credit and having it in a transition hook would prevent them from being able to do that. In general, having it at the model level would make it hard to opt out of this behaviour if desired.

I could also add something to the API checkout controller to ensure similar behaviour, or is there a place you prefer this sort of logic to live?

core/app/models/spree/store_credit_category.rb
craigjbass: We were just looking at the Store Credits functionality this morning and agree with @luukveenis's assessment. We found the inconsistency in naming pretty confusing (note it's also inconsistent in the backend UI).

Also, we're actually lost as to what Category and Types are trying to achieve? I.e. what is the difference? Wondering if anyone here could shed some light onto this. From what we could see it didn't seem like there needed to be two models representing this concept.

jhawthorn: Name change is fine by me, but we need to old method to be deprecated before removal.

frontend/app/controllers/spree/checkout_controller.rb
jhawthorn: My concern is not only that we have the API and the admin, but users may have written their own frontend. Yet this behaviour is absolutely essential for store credits to work now.



> I discussed this with @cbrunsdon offline because I initially had this in a before_transition to: :complete. He brought up that there would likely be scenarios where admins would want to push through an order with expired credit and having it in a transition hook would prevent them from being able to do that. In general, having it at the model level would make it hard to opt out of this behaviour if desired.

I would expect the exact opposite of that behaviour to be desired.

core/app/models/spree/store_credit.rb
jhawthorn: That makes some sense, but are we sure that everything in an `expiring` category _must_ be expiring? Or does the category mean it _may_ be expiring

luukveenis: To be honest, I'm not sure what the intention was with the original implementation. Since there was never an expiry date set anywhere, all credits were essentially `non_expiring`, though some had the `expiring` type.

In my mind when I read that a store credit has an `expiring` type, I would expect it to have an expiry date. This does make existing `expiring` credits invalid, so I understand the argument for maintaining compatibility with current behaviour.

frontend/app/controllers/spree/checkout_controller.rb
luukveenis: > Yet this behaviour is absolutely essential for store credits to work now.

I don't entirely agree with this. For users with custom frontends, not including this change will simply maintain current behaviour. If they wish to use expiring credits, they will have to pull this into their frontend, similar to any of the other logic that happens in `before_action`s in the `CheckoutController`.

> I would expect the exact opposite of that behaviour to be desired. Admins shouldn't accidentally be able to use expired store credits.

In our store - and presumably others - there are cases where the order may be held for review before completion (ie: fraud review). For cases like this, we can't guarantee that the order state at the time of checkout will be the same as when it gets approved. It's possible that the credits were valid when the user checked out, but have expired by the time an admin has a chance to review the order.

Another very common customer service use case is users that make it to the confirm page and think they completed the order. When they realize they haven't gotten an order confirmation email, they call customer service to have them complete the order and we can run into the same issue.

There is some potential for admins to accidentally use expired credit, but I think that's better than putting this in a transition hook that prevents them from completing these orders altogether, when there are valid use cases for doing so.

core/app/models/spree/store_credit.rb
craigjbass: @luukveenis - yes. It also means the current implementation is far more complicated than it should be (to achieve the same thing). I'm personally new to Solidus eCom but seems like it is a really difficult affair to backtrack on design decisions, due to fear that it will break someone's proprietary setup?

tvdeyen:
@jhawthorn @luukveenis @cbrunsdon would you mind to revisit this and come to a decisions wether we can merge this or not? As I think store credits should be an extension actually I am not the right person to take any decision on this :P

luukveenis:
@kennyadsl thanks for the feedback.

I would be in favour of removing the `Spree::StoreCreditType` model/association and the logic around automatically assigning the type (expiring/non-expiring) based on the `Spree::StoreCreditCategory`.

https://github.com/solidusio/solidus/blob/master/core/app/models/spree/store_credit.rb#L272-L281

The credit type and category don't have a direct association and this is the main point of confusion I think. With the added `expires_at` field, we should just be able to infer whether a credit is expiring or non-expiring by whether it has an expiry date set.

If that code is removed, the `Spree::StoreCreditCategory` model only stores one attribute (`name`). In a new Solidus installation, we only create the `Default` store credit category. I'm not sure if this is used by anyone, but I also don't see any harm in keeping that.

seand7565:
Similar to #1602 the assignee doesn't work with Solidus anymore, and there hasn't been any activity in years - candidate for closing? @aldesantis

aldesantis:
I have closed this, although the PR did look cool and I'd be happy to see someone else take over this work.

PR #1855, opened 2017-04-21
Convert new payment JS to backbone

jhawthorn:
This replaces `payments/new.js` and `select_payments.js.coffee`, which were redundant and both acted on the same page and inputs, with two backbone views.

The first, `Spree.Views.Payment.New`, handles which payment method is selected and showing/hiding/disabling form inputs based on selection.

The second, `Spree.Views.Payment.EditCreditCard`, handles editing standard credit card fields using `jquery.payment` as well as selecting existing credit cards.


jhawthorn:
> The only concern I have is about the removed data-hook. I want to see them be gone as well, but stores "may" depend on it. Changelog?

Any store "may" depend on anything. One could target `[class=""]` in deface and it would make just as much sense as targettu `[data-hook]`. I try to leave in the "named" `data-hook`s, because that is/was a popular pattern, but I have no idea why empty `data-hook`s are a thing.

Regardless, I have added it back because there was no reason to remove it here (other than my desire to remove useless markup)

PR #1856, opened 2017-04-22
Remove js.erb response for prices#new and add breadcrumbs

jhawthorn:
The new link was previously using `remote: true` and a `js.erb` template which replaced the main body of the page with the new form. This PR changes it to link to the new url normally.

This fixes a minor bug where the NumberWithCurrency JS component wasn't being initialized on the dynamically added content. It also allows having the breadcrumbs and title change correctly.

If we wanted the benefits of the old behaviour (avoiding full page load), we should use turbolinks.

### DIFF
--- a/backend/app/views/spree/admin/prices/edit.html.erb
+++ b/backend/app/views/spree/admin/prices/edit.html.erb
@@ -1,4 +1,6 @@
 <%= render 'spree/admin/shared/product_tabs', current: 'Prices' %>
+<% admin_breadcrumb(link_to(plural_resource_name(Spree::Price), spree.admin_product_prices_url(@product))) %>
+<% admin_breadcrumb(Spree.t('actions.edit')) %>
 
 <%= form_for @price, url: spree.admin_product_price_path(@product, @price) do |f| %>
   <fieldset>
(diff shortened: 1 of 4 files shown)


tvdeyen:
> we should use turbolinks.

That's something I really want to look into next. Now that our JS is better organized this should be easy to accomplish.

mtomov:
> >    we should use turbolinks.

> That's something I really want to look into next. Now that our JS is better organized this should be easy to accomplish.

:+1: 

I have been thinking the same those past couple of days, and even looked briefly at [jquery.turbolinks](https://github.com/kossnocorp/jquery.turbolinks)

tvdeyen:
✅  https://github.com/solidusio/solidus/pull/1863

> I have been thinking the same those past couple of days, and even looked briefly at jquery.turbolinks

jquery.turbolinks does not support Turbolinks 5 and is not needed by us. There might be an extension or store that uses an old jquery plugin that makes internal use of `dom:ready`, but this is most likely very rare.

Extensions and stores DO have to use `Spree.ready` in stead of `jQuery.ready` though. But this is a simple search and replace update.

PR #1863, opened 2017-04-24
[RFC] Allow use of Turbolinks in admin

tvdeyen:
Using Turbolinks for navigating admin should give us a nice little speed gain.

For getting this to work it was needed to use `Spree.ready` instead of `jQuery.ready` to initialize the various JS modules we have.

Luckily `Spree.ready` already takes care of the `turbolinks:load` event, so this was easy for our own admin, for extensions we overwrote `jQuery.ready` to use `Spree.ready` instead (the same trick that `jquery-turbolinks` uses).

- [x] Disable Turbolinks per default and enable it only for new Stores
- [x] Override `jQuery.ready` and log deprecation notices to the JS console

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Solidus 2.5.0 (master, unreleased)
 
+- Add support for using Tubolinks in Solidus admin [#1863](https://github.com/solidusio/solidus/pull/1863) ([tvdeyen](https://github.com/tvdeyen))
+
 ## Solidus 2.4.0 (unreleased)
 
 ### Major changes
--- a/README.md
+++ b/README.md
@@ -139,6 +139,20 @@ include. This can be disabled by adding the following to
 config.assets.debug = false
 ```
 
+### Turbolinks
+
+To gain some extra speed you may enable Turbolinks inside of Solidus admin.
+
+Add `gem 'turbolinks', '~> 5.0.0'` into your `Gemfile` (if not already present) and append these lines to `vendor/assets/spree/backend/all.js`:
+
+```js
+//= require turbolinks
+//= require backend/app/assets/javascripts/spree/backend/turbolinks-integration.js
+```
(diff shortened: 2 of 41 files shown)


mtomov:
Super! Could you maybe push this code on a branch on the `solidusio` repo, as for some reason when I try to create a pull request towards your fork, I can't see it in the list ..

tvdeyen:
@mtomov sure https://github.com/solidusio/solidus/tree/turbolinks

mamhoff:
@mtomov is still working on this

tvdeyen:
This is now ready to review

tvdeyen:
@cbrunsdon rebased with latest `master` and ready to be reviewed.

backend/app/assets/javascripts/spree/backend/turbolinks-configuration.js.erb
```diff
@@ -0,0 +1,20 @@
+// comment line necessary for correct interpolation of use_turbolinks
+Turbolinks.supported = <%= Spree::Backend::Config.use_turbolinks %>;
```
jhawthorn: Instead would it be possible to do by having a user edit their `vendor/assets/javascripts/spree/backend/all.js`?

tvdeyen:
Like discussed in the last core team meeting we will ship everything that make Solidus admin compatible with Turbolinks, but **do not enable it by default**. Even not for newly created stores.

The chance to break something by an extension or usage of JS lib that is not compatible with Turbolinks exists and we do not want to maintain issues caused by usage of Turbolinks.

Use at your own risk. "It should work"™️

tvdeyen:
@cbrunsdon rewrote the flaky spec.

tvdeyen:
Rebased with latest `master`

mtomov:
That was a good decision. I like it better without the config option.

Thanks for all your work on that!

Issue #1726, opened 2017-02-20
Core: OrderContents#add quantity parameter not set to default when passed nil


vfonic:
### Steps to reproduce
I had code similar to `orders_controller#populate`:
```rb
# Adds a new item to the order (creating a new order if none already exists)
    def populate
      @order   = current_order(create_order_if_necessary: true)
      variant  = Spree::Variant.find(params[:variant_id])
      quantity = params[:quantity].to_i

      # 2,147,483,647 is crazy. See issue https://github.com/spree/spree/issues/2695.
      if !quantity.between?(1, 2_147_483_647)
        @order.errors.add(:base, Spree.t(:please_enter_reasonable_quantity))
      end

      begin
        @line_item = @order.contents.add(variant, quantity) # this line
      rescue ActiveRecord::RecordInvalid => e
        @order.errors.add(:base, e.record.errors.full_messages.join(", "))
      end

      respond_with(@order) do |format|
        format.html do
          if @order.errors.any?
            flash[:error] = @order.errors.full_messages.join(", ")
            redirect_back_or_default(spree.root_path)
            return
          else
            redirect_to cart_path
          end
        end
      end
    end
```

The difference is that:
1. I don't have reasonable quantity check (I think that's unnecessary code)
2. I have this line slightly different:
Solidus core code:
```rb
@line_item = @order.contents.add(variant, quantity)
```
Mine:
```rb
@line_item = @order.contents.add(variant, params[:quantity])
```

I looked at `OrderContents` and saw the default parameter value for quantity to be one and assumed that it would work with these params only:
```
Parameters: {"variant_id"=>"2"}
```

### Expected behavior
The quantity defaults to 1 when it's: nil or an empty string

### Actual behavior
The quantity defaults to 0.
The result of this is a new `line_item` added to order with a quantity of 0.

### System configuration
**Solidus Version**:
master: 4cfd1a2b64482329f2e36452b4b2470659b3c8a9

I tried to resolve this issue and ended up overcomplicating things. Should we just do something like this:
```rb
    def add(variant, quantity = 1, options = {})
      quantity = quantity.to_i
      quantity = 1 if quantity <= 0
      line_item = add_to_line_item(variant, , options)
      after_add_or_remove(line_item, options)
    end

    def remove(variant, quantity = 1, options = {})
      quantity = quantity.to_i
      quantity = 1 if quantity <= 0
      line_item = remove_from_line_item(variant, quantity.to_i, options)
[...]

PR #1866, opened 2017-04-26
Fix deleting a line item when asked to remove more than its quantity

omnistegan:
This PR addresses a small issue I found in `OrderContents#remove`.

In cases where `#remove` is called with a `quantity` greater than the `line_item.quantity`, the order would be left with a `line_item` with 0 quantity. It seems like the expected behaviour here would be to remove the line item if more than the current quantity is requested to be removed.

I addressed this simply by using `<=` instead of `==` for determining whether or not to destroy the `line_item`.

I'm also included a test that fails without the change.

### DIFF
--- a/core/app/models/spree/order_contents.rb
+++ b/core/app/models/spree/order_contents.rb
@@ -112,7 +112,7 @@ module Spree
       line_item.quantity -= quantity
       line_item.target_shipment = options[:shipment]
 
-      if line_item.quantity == 0
+      if line_item.quantity <= 0
         order.line_items.destroy(line_item)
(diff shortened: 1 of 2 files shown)


vfonic:
@omnistegan thanks for the PR! I've submitted an issue about similar bug when adding a line item with quantity being zero. If you find time to look into that one, would be great:
https://github.com/solidusio/solidus/issues/1726

I've also noticed that a lot of code in `add_to_line_item` and `remove_from_line_item` is the same. In fact, I'm sure these two methods could be merged into one.

PR #1867, opened 2017-04-26
[RFC] [WIP] [experiment] More configurable admin menu

jhawthorn:
Similar to what was done in #1296, this aims to make the admin menu.

As an RFC, I'd like to figure out how we'd like to ideally want to configure the menu. I haven't yet looked at backwards compatibility or implementation, so please hold off on reviewing that aspect of this.

The customization of existing menu, though greatly improved by #1296, leaves much to be desired. We can programmatically add top-level items, but not second-level items.2

How we determined what was active is a mess:
* For the top level, we store an array of controller names (ex. `BackendConfiguration::ORDER_TABS`), and consider it active if we are on any of those controllers
* On second level navigation, we (mostly) consider it active if the link 
* Alternatively a regex can be specified which is tested against the current URL

**This takes the following approach instead:**

"Second level" items, are now referred to as just "items". They have a label, url, and a "handle" (used to refer to them inside Solidus).

"Top level" items are considered sections, which have an icon and a label, they contain the other items. The section itself is a link to its first item, a convention we used previously, which this codifies. Because of this sections themselves don't have a URL defined.

I've tried to optimize for extensions and stores to add their own second-level items

``` ruby
menu.add_item(
  'settings/my_extension',
  url: :admin_my_model_path,
  condition: -> { can?(:admin, MyModel) }
)
```

Adding a new top level item is similar

``` ruby
menu.add_section(
  'my_extension'
  icon: 'rocket'
)
menu.add_item(
  'my_extension/my_extension',
  url: :admin_my_model_path,
  condition: -> { can?(:admin, MyModel) }
)
```

The next problem is how to detect which item is active. My current thought is to require declaring the active menu item in each view.

```
<%# admin/payment_methods/index.html.erb %>
<% admin_menu 'settings/payments' %>
```

One advantage of declaring both 1st-level sections and 2nd-level items in code (rather than as partials) is that we only have to detect the 2nd-level item being active and can infer that the 1st level is.

Which is a lot more reliable than the previous method, but a lot more verbose, needing a line in every top-level view. I'm not sure it's worth it vs specifying controllers.

Looking forward to hearing your thoughts.

**Still todo in the experiment phase:**
* [ ] 3rd level navigation (tabs)
* [ ] explore backwards compatibility


kennyadsl:
Great solution! We can even implement methods to remove items/sections that can be useful into stores/extension.

>  I'm not sure it's worth it vs specifying controllers.

What about implementing a "simple" finder by controller to find the active item and section?

```ruby
class Item
  # ...
  def controller
    @controller ||= Rails.application.routes.recognize_path(url).fetch(:controller)
  end
end
```

and somewhere in controllers land:

```ruby
def admin_menu
  @admin_menu_handle = Spree::Backend::Config.menu.items.find do |item| 
    item.controller == current_controller
  end

  # for this to work we'll probably need to pass section to items
  @admin_menu_section = @admin_menu_handle.section
end
```

Of course this is just a raw idea, I'm not sure this can work. Anyway this way we can maybe avoid adding both controllers list and view verbosity?

gmacdougall:
"This is also nothing" -- John Hawthorn

PR #1872, opened 2017-05-01
Fix BigDecimal errors under ruby 2.4

jhawthorn:
Under ruby 2.4, `BigDecimal.new` raises an `ArgumentError` when casting an invalid value (it used to return 0). Unfortunately, `#to_d` also (probably incorrectly) raises a `ArgumentError`, unlike `#to_i` and friends.

This works around this issue by capturing `ArgumentError` and returning 0 in that case.

With this change the test suite passes under ruby 2.4 (but still has a fair number of warnings from versioncake)


jhawthorn:
I thought about this a bit and couldn't find a better way to organize it. I don't want to introduce a new class (since this will be serialized). I want this to be as deletable as possible, since it's just a workaround for a ruby 2.4 bug (which I still hope will be fixed in a patch release)

PR #1876, opened 2017-05-03
Correctly set viewable_id in image upload (2.2.0 regression)

### LINKED ISSUE #1875: 2.3.0.alpha field  product images upload error
Hi, I have problems trying to upload images to products. After researching I realized that in a specific variant image uploading works but not in variant all. 
It's due to field of spree_assets table 'viewable_id' is 0 in variant all and not in specific variant.
I saw that file 'solidus/backend/app/assets/javascripts/spree/backend/images/upload.js' has been modified, maybe there is somwthing wrong here.

Could you check?

jhawthorn:
viewable_id was being set when the ImageUpload was added to the collection, but the upload was started before adding the item leaving `viewable_id` unset.

Fixes #1875

### DIFF
--- a/backend/spec/features/admin/products/edit/images_spec.rb
+++ b/backend/spec/features/admin/products/edit/images_spec.rb
@@ -64,6 +64,8 @@ describe "Product Images", type: :feature do
           expect(page).to have_xpath("//img[contains(@src,'ror_ringer')]")
         end
       end
+
+      expect(Spree::Image.last.viewable).to eq(product.master)
     end
   end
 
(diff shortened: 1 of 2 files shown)


jhawthorn:
We should probably also add a validation to the `image` model, but that's too big a change for a patch release.

backend/spec/features/admin/products/edit/images_spec.rb
```diff
@@ -64,6 +64,8 @@
           expect(page).to have_xpath("//img[contains(@src,'ror_ringer')]")
         end
       end
+
+      expect(Spree::Image.last.viewable).to eq(product.master)
```
Senjai: Err. Is there a way we can make this assertion without looking into the db here?

jhawthorn: Not easily. The regression here was that the value wasn't being sent to the server and not saved in the DB. I think this is the simplest test for that.

jhawthorn:
I was mistaken and this is not a 2.2.0 regression, and was only broken in master. Yay!

PR #1878, opened 2017-05-03
ActionController::Parameters fixes for Rails 5.1

jhawthorn:
From the [Rails 5.1 release notes](https://github.com/rails/rails/blob/5-1-stable/actionpack/CHANGELOG.md):

> *   Raise exception when calling `to_h` and `to_hash` in an unpermitted Parameters.
>
>     Before we returned either an empty hash or only the always permitted parameters
>    (`:controller` and `:action` by default).
>
>    The previous behavior was dangerous because in order to get the attributes users
>    usually fallback to use `to_unsafe_h that` could potentially introduce security issues.
>
>    *Rafael Mendonça França*

This makes our usage of ActionController::Parameters compatible with both Rails 5.0 and Rails 5.1


jhawthorn:
@tvdeyen not until #1895

PR #1882, opened 2017-05-04
Configure admin turbolinks

mtomov:
- Enabled for new applications

- Disabled for existing applications. Enable in an initializer with:
  ```ruby
    config.use_turbolinks = true
  end
  ```

- Handles gracefully existing javascript code, which still uses
    jQuery.ready and gives a warning of the location of those functions

Towards #1863 

_edit_: *I will see to fix the 18 test failures* .. *tests fixed!*

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/turbolinks-configuration.js.erb
+++ b/backend/app/assets/javascripts/spree/backend/turbolinks-configuration.js.erb
@@ -1,0 +1,20 @@
+// comment line necessary for correct interpolation of use_turbolinks
+Turbolinks.supported = <%= Spree::Backend::Config.use_turbolinks %>;
+
+Spree.jQueryReady = $.fn.ready;
+
+// override jQuery.ready to use Spree.ready even if it was not used explicitly
+$.fn.ready = function (callback) {
+  console.warn("jQuery.ready() is deprecated. Use Spree.ready() instead. Called from:", callback );
+  Spree.ready(callback);
+};
+
+Spree.ready = function(callback) {
+  if (Turbolinks.supported) {
+    jQuery(document).on('turbolinks:load', function() {
+      callback(jQuery);
+    });
+  } else {
+    Spree.jQueryReady(callback);
+  }
+};
--- a/backend/app/models/spree/backend_configuration.rb
+++ b/backend/app/models/spree/backend_configuration.rb
@@ -2,6 +2,9 @@ module Spree
   class BackendConfiguration < Preferences::Configuration
     preference :locale, :string, default: Rails.application.config.i18n.default_locale
 
+    # @!attribute [rw] use_turbolinks
+    preference :use_turbolinks, :boolean, default: false
+
     ORDER_TABS         ||= [:orders, :payments, :creditcard_payments,
                             :shipments, :credit_cards, :return_authorizations,
                             :customer_returns, :adjustments, :customer_details]
--- a/backend/app/assets/javascripts/spree/backend.js
+++ b/backend/app/assets/javascripts/spree/backend.js
@@ -18,6 +18,7 @@
 //= require spree/backend/namespaces
 //= require spree/backend/translation
 //= require spree/backend/backbone-overrides
+//= require spree/backend/turbolinks-configuration
 //= require spree/backend/format_money
 //
 //= require spree/backend/templates
--- a/backend/spec/models/spree/backend_configuration_spec.rb
+++ b/backend/spec/models/spree/backend_configuration_spec.rb
@@ -1,0 +1,9 @@
+require 'spec_helper'
+
+describe Spree::BackendConfiguration, type: :model do
+  let(:prefs) { Spree::Backend::Config }
+
+  describe '#use_turbolinks' do
+    specify { expect(prefs).to respond_to(:use_turbolinks) }
+  end
+end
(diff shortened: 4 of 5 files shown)


backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -159,7 +159,6 @@ Spree.ready(function(){
       return ui;
   };
 
-  $('table.sortable').ready(function(){
```
tvdeyen: This is a change in the actual implementation and not necessary for Turbolinks to work, right? I think this should not be in this PR.

backend/app/assets/javascripts/spree/backend/turbolinks-configuration.js.erb
```diff
@@ -0,0 +1,20 @@
+// comment line necessary for correct interpolation of use_turbolinks
+Turbolinks.supported = <%= Spree::Backend::Config.use_turbolinks %>;
+
+// override jQuery.ready to support Turbolinks
```
tvdeyen: This comment is misleading. You are doing this because we want to print out a deprecation warning and use `Spree.ready` even it it was not used explicitly. Turbolinks works with our this change.

backend/app/models/spree/backend_configuration.rb
```diff
@@ -2,6 +2,10 @@ module Spree
   class BackendConfiguration < Preferences::Configuration
     preference :locale, :string, default: Rails.application.config.i18n.default_locale
 
+    # @!attribute [rw] use_turbolinks
+    preference :use_turbolinks, :boolean, default: false
+
+
```
tvdeyen: Please satisfy Hound here.

backend/app/assets/javascripts/spree/backend/turbolinks-configuration.js.erb
```diff
@@ -0,0 +1,20 @@
+// comment line necessary for correct interpolation of use_turbolinks
+Turbolinks.supported = <%= Spree::Backend::Config.use_turbolinks %>;
```
tvdeyen: Neat little trick 👌 

But we remove the internal check of Turbolinks for supported browsers with this. Imagine someone enables Turbolinks in the backend and uses an unsupported browser, Turbolinks will not gracefully fall back anymore. 

I can live with that, because we encourage admin users to use latest browsers, but this can cause issues.

backend/spec/features/admin/orders/new_order_spec.rb
```diff
@@ -175,8 +170,6 @@
     end
 
     it "transitions to delivery not to complete" do
-      click_on 'Cart'
-
```
tvdeyen: Can we have these changes to the new order and customer details feature specs as operate PRs to master? I want to be sure they still pass on non-Turbolinks based backend.

backend/app/assets/javascripts/spree/backend/turbolinks-configuration.js.erb
mtomov: Thanks for the feedback! I will sort out the rest if we get the #1900 merged in. 

On this one, we can copy over the [few lines from turbolinks](https://github.com/turbolinks/turbolinks/blob/c73e134731ad12b2ee987080f4c905aaacdebba1/src/turbolinks/index.coffee#L9-L12), but I really hate to encourage people to use old browsers, and those unsupported browsers are already ice age by now, let alone by the time somebody actually starts using this feature here..

So, well pointed out; my preference is to keep this behaviour.

mtomov: Hi @tvdeyen , sorry to pester .. just whenever you get to it .. if you could rebase your branch on top of master now that that preparation PR has been merged .. as otherwise I think I won't be able to get the tests to green?

tvdeyen:
@mtomov could you please rebase on master, now that #1900 was merged?

backend/app/assets/javascripts/spree/backend/turbolinks-configuration.js.erb
tvdeyen: @mtomov sure. Sorry, I forgot the target of this PR, you are totally right!

tvdeyen:
@mtomov rebase `turbolinks` branch on `master`, now could you please rebase your branch on top of `turbolinks`?

backend/app/assets/javascripts/spree/backend/turbolinks-configuration.js.erb
mtomov: No problem, but I think that you might have updated `tvdeyen/turbolinks` instead of `solidusio/turbolinks` : )

tvdeyen: @mtomov updated `solidusio/turbolinks`

mtomov:
Super, thanks, updated from my side as well! 

Let me know! Thanks!

tvdeyen:
@mtomov just rebased again with master, could you please rebase again and while your at it fix the typo in the commit message ("Turobolinks" vs. "Turbolinks")? Thanks!

mtomov:
That's that done! All easy with small PRs : ) Thanks!

tvdeyen:
@mtomov Thanks, but I still see the typo in the commit message.

mtomov:
Correct, my bad, thanks for noticing! Should be good now.

PR #1883, opened 2017-05-04
Only touch product taxons when one exists

jhawthorn:
This avoids running weird queries like

    UPDATE "spree_taxons" SET "updated_at" = ... WHERE 1=0

### DIFF
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -339,10 +339,12 @@ module Spree
     # Iterate through this products taxons and taxonomies and touch their timestamps in a batch
     def touch_taxons
       taxons_to_touch = taxons.map(&:self_and_ancestors).flatten.uniq
-      Spree::Taxon.where(id: taxons_to_touch.map(&:id)).update_all(updated_at: Time.current)
+      unless taxons_to_touch.empty?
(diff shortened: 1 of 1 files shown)


jhawthorn:
Makes no performance difference, we always have an Array here. `any?` would be the best choice anyways.

tvdeyen:
On Arrays `any?` is `o(n)` while `empty?` is `o(1)`. From [the documentation](https://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-any-3F):

> If the block is not given, Ruby adds an implicit block of `{ |obj| obj }` that will cause `any?` to return `true` if at least one of the collection members is not `false` or `nil`.

This may not be a problem here, as we most likely will not have very large Arrays, but why not use the faster solution if we could?

tvdeyen:
I already approved this PR, so I don't insist on this change. Still thinking `present?` would be the "more correct question to ask", but it won't matter in this particular case. So, 👍

jhawthorn:
> I already approved this PR, so I don't insist on this change. Still thinking present? would be the "more correct question to ask", but it won't matter in this particular case. So,

:+1: I can get behind this. I don't think the performance would ever be an issue, but there is a mental overhead for the reader debating if it matters that any of the elements could be falsey.

Issue #1881, opened 2017-05-04
Solidus codebase has many dynamic finders


spaghetticode:
I am having issues integrating the `solidus_globalize` gem, issues that are related to the presence in solidus codebase of many dynamic finders such as `find_by_permalink!`, `find_by_name!` and so on, which as far as I know are/will be deprecated in rails.

I am only a beginner with solidus and its ecosystem, but I think that those finders should be replaced with on of these types of code:
```
where(field: "value").first!
find_by!(field: "value")
```
with the multiple benefits of: 
* fixing the issues with the `solidus_globalize` gems
* preparing `solidus` codebase for the moment when those dynamic finders will be completely removed

I think that myself (or somebody else at my company) may be able to handle this task and provide a PR in the next few days if there is general consensus that the issue is real, that this is something that should be done and that the proposed solution is good and sound.

tvdeyen:
👍 

First step taken by https://github.com/solidusio/solidus/pull/1880

AndreiMotinga:
I'd like to take a stab at it. Since I'm just starting out with open source,  any advice or guidance on how I should approach this would be appreciated.

jhawthorn:
> I am having issues integrating the solidus_globalize gem, issues that are related to the presence in solidus codebase of many dynamic finders such as find_by_permalink!, find_by_name! and so on, which as far as I know are/will be deprecated in rails.

These are *not* deprecated and there's no indication that they ever will be (quite the opposite). It is a misconception because some other dynamic finders were removed from rails (ages ago in rails 4.0).

From [activerecord-deprecated_finders README](https://github.com/rails/activerecord-deprecated_finders/blob/master/README.md):

> Note that find(primary_key), find_by..., and find_by...! are not deprecated.

So it is a globalize bug (globalize/globalize#423) and it should be fixed there. I'm not totally comfortable with us making changes simply because another library is broken.

AndreiMotinga:
> I'm not totally comfortable with us making sweeping changes simply because another library is broken.

I agree, it makes sense. However, in this particular case, the change required is very simple and won't change anything about the solidus, other than make it slightly more flexible.  

Plus, even when some other translation project will appear, will it face the same problem ?

spaghetticode:
@jhawthorn thanks for the insight. 

As you correctly wrote those dynamic finders are not actually deprecated *right now*, but according to https://github.com/rails/rails/blob/master/activerecord/lib/active_record/base.rb their usage is already sort of discouraged:

[...]

PR #1886, opened 2017-05-05
Replace ActiveRecord dynamic finders

spaghetticode:
See issue #1881


backend/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
```diff
@@ -8,7 +8,7 @@
 
     let(:order) { create(:order, number: "R123456789") }
 
-    before { allow(Spree::Order).to receive(:find_by_number!) { order } }
+    before { allow_any_instance_of(Spree::Order::ActiveRecord_Relation).to receive(:find_by!) { order } }
```
spaghetticode: I had to change the code as shown or the specs would fail. 

The reason behind this is unclear to me, as I didn't expect that changing only the message name (from `:find_by_number!` to `:find_by!`) would make the specs to fail, but that's what happens.

The relevant line for that `before` block is:
```
@order = Spree::Order.includes(:adjustments).find_by_number!(number: params[:order_id])
```

`Spree::Order.includes(:adjustments)` is not an instance of `Spree::Order`, but an instance of `Spree::Order::ActiveRecord_Relation`, so I am puzzled about how it worked before my changes.

tvdeyen: A `receive_message_chain(:adjustments, :find_by!)` should do the trick here.

backend/spec/controllers/spree/admin/orders_controller_spec.rb
```diff
@@ -30,7 +30,7 @@
     let(:adjustments) { double('adjustments') }
 
     before do
-      allow(Spree::Order).to receive_messages(find_by_number!: order)
+      allow_any_instance_of(Spree::Order::ActiveRecord_Relation).to receive_messages(find_by!: order)
       allow(order).to receive_messages(contents: Spree::OrderContents.new(order))
```
tvdeyen: We prefer using `receive_message_chain` here as we don't want to test internal active record implementation details. This specs will fail, if Rails renames it's relation class.

backend/app/controllers/spree/admin/option_types_controller.rb
```diff
@@ -21,7 +21,7 @@ def location_after_save
       end
 
       def load_product
-        @product = Spree::Product.find_by_param!(params[:product_id])
+        @product = Spree::Product.find_by!(param: params[:product_id])
```
tvdeyen: This method is defined by us in `core/lib/spree/core/permalinks.rb` and are included in all active record classes. This should not be replaced.

tvdeyen:
I'm shocked that the specs passes, although we changed `find_by_param` usages 🤔

backend/app/controllers/spree/admin/option_types_controller.rb
kennyadsl: I think this is not tested, this should probably be something like: 

```ruby
find_by: Spree::Product.permalink_field.to_sym => params[:product_id]
```

kennyadsl:
I'd remove those permalink finders methods from`core/lib/spree/core/permalinks.rb` if we are not using them or they can be replaced with `Model.permalink_field` using the default `find_by!` method.

backend/spec/controllers/spree/admin/orders_controller_spec.rb
spaghetticode: Thank you for the suggestion, `receive_message_chain` did the trick indeed

backend/app/controllers/spree/admin/option_types_controller.rb
spaghetticode: I think the best thing for me, with my limited knowledge of what is happening here, is to revert the code to what it was before.

kennyadsl: yes, it's probably better to do that in a separate PR anyway (if it is useful)

spaghetticode:
@mamhoff That's a good point, thank you. I changed the commit description and the PR title as well with a more descriptive text.

picazoH:
nice PR :clap: thank you @spaghetticode

mamhoff:
Please squash both commits into one (the first one would introduce a commit at which we'd have a failing test suite, which is no good when you try `git bisect` to find which commit introduced regressions). 

Also, your commit message's title is too long (50 characters is what you should aim for), and the body of the message reads like the commit title of another commit, not like an explanation of why you did that change. I encourage reading the aforementioned blog post again! 

Like I said before, I am absolutely fine with the content of the PR. But please be considerate to future readers of this code base.

spaghetticode:
@mamhoff Sorry I took the commit message matter too lightly. I hope it's now fine with you.

mamhoff:
@spaghetticode The commit message is much improved now, thank you for that! 

@tvdeyen if you have a minute could you re-review this? I think your concerns have been addressed, too.

PR #1887, opened 2017-05-05
add locales

AndreiMotinga:
These 2 are missing when used with solidus_i18n


jhawthorn:
Thanks, but this should be done in solidus_globalize, since those keys aren't referenced here.

PR #1888, opened 2017-05-05
Bug fix: Don't update wallet default payment method if the default is already …

ahoernecke:
…set to the same value

This fixes a bug where if the app tried to set the default payment method to the existing default, the wallet would end up with no default payment method

### DIFF
--- a/core/app/models/spree/wallet.rb
+++ b/core/app/models/spree/wallet.rb
@@ -61,6 +61,11 @@ class Spree::Wallet
       raise Unauthorized, "wallet_payment_source #{wallet_payment_source.id} does not belong to wallet of user #{user.id}"
     end
 
+    # Do not update the payment source if the passed source is already default
+    if default_wallet_payment_source.try(:id) == wallet_payment_source.id
+      return
+    end
+
     wallet_payment_source.transaction do
       # Unset old default
       default_wallet_payment_source.try!(:update!, default: false)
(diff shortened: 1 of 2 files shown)


core/app/models/spree/wallet.rb
```diff
@@ -61,6 +61,11 @@ def default_wallet_payment_source=(wallet_payment_source)
       raise Unauthorized, "wallet_payment_source #{wallet_payment_source.id} does not belong to wallet of user #{user.id}"
     end
 
+    # Do not update the payment source if the passed source is already default
+    if default_wallet_payment_source.try(:id) == wallet_payment_source.try(:id)
```
mamhoff: What happens if both of these return `nil`?

jhawthorn: It's probably better to also not use `.try(:id)`. `default_wallet_payment_source == wallet_payment_source`

ahoernecke: If both return nil, it should signify that the system is trying to clear the default payment source (wallet_payment_source=>nil) and that the default payment source is already not set (default_wallet_payment_source=>nil). Since we just return for this case, nothing would change, which I would think is what would be wanted.

ahoernecke: @jhawthorn, made the change to not compare against the ids.

PR #1890, opened 2017-05-05
Use ActiveJob :test queue adapter

jhawthorn:
Previously, we used the `:inline` queue adapter by default. We had one test where the queue adapter was switched to `:test`, and not switched back, which causes a test order dependency issue.

The :test queue adapter is great, faster (since it isn't being run by default), and gives us better control over our test suite. This commit uses the `:test` adapter for the entire test suite, clearing the performed
and enqueued jobs after each test.

### DIFF
--- a/api/spec/spec_helper.rb
+++ b/api/spec/spec_helper.rb
@@ -36,7 +36,7 @@ require 'spree/api/testing_support/caching'
 require 'spree/api/testing_support/helpers'
 require 'spree/api/testing_support/setup'
 
-ActiveJob::Base.queue_adapter = :inline
+ActiveJob::Base.queue_adapter = :test
 
 RSpec.configure do |config|
   config.backtrace_exclusion_patterns = [/gems\/activesupport/, /gems\/actionpack/, /gems\/rspec/]
@@ -64,6 +64,7 @@ RSpec.configure do |config|
     Spree::Api::Config[:requires_authentication] = true
   end
 
+  config.include ActiveJob::TestHelper
   config.include VersionCake::TestHelpers, type: :controller
   config.before(:each, type: :controller) do
     set_request_version('', 1)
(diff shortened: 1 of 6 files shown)


api/spec/spec_helper.rb
```diff
@@ -36,7 +36,7 @@
 require 'spree/api/testing_support/helpers'
 require 'spree/api/testing_support/setup'
 
-ActiveJob::Base.queue_adapter = :inline
+ActiveJob::Base.queue_adapter = :test
```
tvdeyen: Have you tried to just remove it?

I think the dummy app generated by Rails will most likely already have set this in the `environments/test.rb`.

jhawthorn:
@tvdeyen I agree. We should either be testing that the right job is being enqueued with the right params, or test the result of the jobs using `perform_enqueued_jobs`. I'll take a look next week.

jhawthorn:
@tvdeyen updated to be a less significant change. Now has the same assertions as before.

api/spec/spec_helper.rb
jhawthorn: The default is `:async`, our dummy app doesn't change the value.

api/spec/spec_helper.rb
```diff
@@ -64,6 +64,12 @@
     Spree::Api::Config[:requires_authentication] = true
   end
 
+  config.include ActiveJob::TestHelper
+  config.after(:each) do
+    ActiveJob::Base.queue_adapter.enqueued_jobs = []
+    ActiveJob::Base.queue_adapter.performed_jobs = []
```
jordan-brough: Just curious - So these don't happen automatically when you use the `:test` adapter?  And there aren't any nicer helper methods?

api/spec/spec_helper.rb
tvdeyen: Thanks for checking

api/spec/spec_helper.rb
jhawthorn: :+1: I think you're right this isn't necessary anymore. I only needed it before including ActiveJob::TestHelper, which adds a setup step which clears this.

PR #1892, opened 2017-05-08
New configurable tax calculator interface

adammathys:
One of the major barriers to implmenting a custom tax calculator is having to reimplement a large amount of logic around creating/updating tax adjustments.

In order to simplify this, the integration point for configurable tax calculation should change a little bit. The new interface this PR proposes takes in a `Spree::Order` in the initializer, then returns tax information in a set of POROs which can then be used by Solidus to create or update tax adjustments accordingly.

Here are the new POROs, in all of their simple glory:
```ruby
module Spree
  module Tax
    class TaxedOrder
      include ActiveModel::Model
      attr_accessor :id, :line_items, :shipments
    end

    class TaxedItem
      include ActiveModel::Model
      attr_accessor :id, :label, :tax_rate, :amount, :included_in_price
    end
  end
end
```

Eventually, it would be nice to also have the input for the tax calculator be a PORO as well. However, I felt that would be too large of a change for this initial PR. I have put some initial work into what those objects could look like and pushed them up to [a Gist](https://gist.github.com/adammathys/0e9a56e59bad4d9cb563e3f7cd7532dc).

TODO
----
- [x] More comprehensive unit tests of the new models.
- [x] Ensure existing adjuster overrides work
- [x] Shipping rate taxation
- [x] Good comments at the top of `TaxCalculator::Default`, to help future implementers


core/app/models/spree/tax/application.rb
```diff
@@ -0,0 +1,51 @@
+module Spree
+  module Tax
+    class Application
```
jhawthorn: I think we need a different class name here. `Application` has too strong another meaning

core/app/models/spree/tax/application.rb
```diff
@@ -0,0 +1,51 @@
...
+      def initialize(order, output)
+        @order = order
+        @output = output
+      end
+
+      def perform!
+        %i(line_items shipments).each do |items|
+          order.send(items).each do |item|
+            rates = output.send(items).select { |i| i.id == item.id }
```
jhawthorn: I think we could avoid the `send`s here (but the rest of the structure looks good).

core/app/models/spree/tax/output/item.rb
```diff
@@ -0,0 +1,10 @@
+module Spree
+  module Tax
+    module Output
```
jhawthorn: _Maybe_ we can come up with a more obvious name to what these classes represent than `Output`. Maybe `Spree::Tax::TaxedItem`? Naming is hard.

core/app/models/spree/order_updater.rb
```diff
@@ -226,7 +226,8 @@ def update_order_promotions
     end
 
     def update_taxes
-      Spree::Config.tax_adjuster_class.new(order).adjust!
+      output = Spree::Config.tax_calculator_class.new(order).calculate
+      Spree::Tax::Application.new(order, output).perform!
```
jhawthorn: If we move this change into the `Spree::Tax::OrderAdjuster`, we can keep overrides of that working (and delete the existing code in there)

core/app/models/spree/app_configuration.rb
```diff
@@ -360,8 +360,10 @@ def add_payment_sources_to_wallet_class
     # @return [Class] a class with the same public interfaces as
     #   Spree::Tax::OrderAdjuster
     # @api experimental
+    # @deprecated - use tax_calculator_class instead
     attr_writer :tax_adjuster_class
     def tax_adjuster_class
+      Spree::Deprecation.warn("Spree::AppConfiguration#tax_adjuster_class is DEPRECATED, please use #tax_calculator_class instead.", caller)
```
mamhoff: If we keep this interface and move the new stuff into the default order adjuster? As per @jhawthorns comment on a later commit?

core/app/models/spree/tax/output/item.rb
mamhoff: `TaxationData`?

core/app/models/spree/tax/output/item.rb
```diff
@@ -0,0 +1,10 @@
...
+      class Item
+        include ActiveModel::Model
+        attr_accessor :id, :tax_rate, :amount, :included_in_price
```
mamhoff: 1. I would like the `label` to be part of this. Avalara can add labels to whatever they return, and so should we.
2. I would like the matching not to be done by ID. When recording taxation API interactions with e.g. Avalara, it always hits us that matching is done by what is essentially a database artifact. Can we just try relying on sorting by ID and adding taxes - or no taxes - based on that? 
3. I would like our tax rate to be a hidden implementation detail. If using an external service, e.g. Avalara, we don't have tax rates to play with. 
```
attr_accessor :i_of_sorted_items, :label, :amount, :included_in_price
```

core/app/models/spree/tax/output/order.rb
```diff
@@ -0,0 +1,10 @@
+module Spree
+  module Tax
+    module Output
+      class Order
+        include ActiveModel::Model
+        attr_accessor :id, :line_items, :shipments
```
mamhoff: The current implementation allows for order-level tax adjustments. For all sane configurations I know we shouldn't allow that, just noting that this is conspicuously absent here.

core/app/models/spree/tax_calculator/default.rb
```diff
@@ -0,0 +1,46 @@
...
+      include Spree::Tax::TaxHelpers
+
+      def initialize(order)
+        @order = order
+      end
+
+      def calculate
+        Spree::Tax::Output::Order.new(
+          id: order.id,
+          line_items: line_item_rates,
```
mamhoff: This line reveals that `Spree::Tax::Output::Order#line_items` should be named `line_item_taxes`.

core/app/models/spree/tax_calculator/default.rb
```diff
@@ -0,0 +1,46 @@
...
+          shipments: shipment_rates
```
mamhoff: Similarly here, with `shipment_taxes`. 

By the way: We had a discussion earlier about the fact that for orders that are beyond the `delivery` step, it's actually preferable to have the shipment's manifest be taxed for contents rather than the line items. The shipments know where they will be shipped from, making correct taxation for multiple stock locations possible.

core/app/models/spree/tax_calculator/default.rb
```diff
@@ -0,0 +1,46 @@
...
+        end
+      end
+
+      def calculate_rates(item)
+        rates_for_item(item).map do |rate|
+          Spree::Tax::Output::Item.new(
+            id: item.id,
+            tax_rate: rate,
+            amount: rate.compute_amount(item),
+            included_in_price: rate.included_in_price
```
mamhoff: This should have the `label` thing as well. We want to abstract from the `Spree::TaxRate`, so `rate` should not be in here.

core/app/models/spree/tax/application.rb
mamhoff: `Adjuster`, `Performer`, `Applier`, `Applicator`, `OrderTaxer`. Just throwing in ideas here, I second Johns thought.

core/app/models/spree/tax/application.rb
```diff
@@ -0,0 +1,51 @@
...
+
+      attr_reader :order, :output
+
+      def update_adjustment(item, output_item)
+        tax_adjustment = item.adjustments.detect do |adjustment|
+          adjustment.source == output_item.tax_rate
+        end
+
+        tax_adjustment ||= item.adjustments.new(
+          source: output_item.tax_rate,
```
mamhoff: Ah! This is why `Spree::TaxRate` still plays a role. 

Adjustments need a `source` so they can be individually re-calculated. However, with any order-level, API-based taxation solution, this will never happen. 

We might need to think about something like a `Spree::Tax` model, similar to adjustments but NOT adjustments.

core/app/models/spree/tax/application.rb
adammathys: I agree. Just went with the the simplest placeholder name for the time being.

core/app/models/spree/tax/output/item.rb
adammathys: 1. Love the idea of moving labels to the output object. I'll put together that change in a few.
2. We need some way of tying the rate to the line item or shipment it should be applied to. The ID is our only real way of doing so. Sorting both lists won't always work. (For example, some items can be tax exempt and shouldn't have any rates associated to them.)
3. As you've commented later, this is due to adjustments requiring a source.

core/app/models/spree/tax/application.rb
adammathys: While I agree that using adjustments for taxes is not ideal, I believe that is much too large of a change to make in this PR.

core/app/models/spree/tax/output/item.rb
mamhoff: Tax exempt items can just have an adjustment_amount of `0`. We can calculate shipping rates without passing database IDs to external services, so why can we not do the same with taxation? :)

core/app/models/spree/tax/application.rb
mamhoff: I might make it a thing for the Hack days to replace tax adjustments with a new thing. Do you mind adding a comment that the `source` is just there to satisfy the requirement of a source on adjustments, and not actually (in terms of business logic) required?

core/app/models/spree/tax/output/order.rb
jhawthorn: We allow that only as legacy pre-spree-2.2 data. The current `Tax::OrderAdjuster` and `OrderUpdater` don't support it so neither should this.

core/app/models/spree/tax/output/item.rb
adammathys: Shipping rates are much easier to avoid passing database IDs. Since there's no need to tie the rate to any existing database record afterwards. If we were looking to make an interface for returning a list of tax rates for an item I would entirely agree with you.

However, we're looking to return a list of calculated taxes for items on an order. Meaning we need to tie those values together in some concrete way. Sorting the items and using their indices, giving them a new alphanumeric identifier, or using the database IDs are all functionally equivalent. Using the database IDs is just the simplest and most robust approach.

jhawthorn: I appreciate both sides here. We do want to associate tax amounts with specific items, and an ID is the most obvious and robust way to do that. That said, the more abstracted from the DB this is the better.

It's worth considering that passing using an ID here doesn't necessarily require that the ID is sent over the wire to an external service, the sort-by-id could be done in the extension.

core/app/models/spree/order_taxation.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    # Walk through the taxes for an item and update adjustments for it. Once
+    # all of the taxes have been added as adjustments, remove any old tax
+    # adjustments that weren't touched.
+    #
+    # @private
+    # @param [#adjustments] item a {Spree::LineItem} or {Spree::Shipment}
+    # @param [Array<Spree::Tax::TaxedItem>] rates a list of calculated taxes for an item
+    # @return [void]
+    def update_adjustments(item, rates)
```
mamhoff: Can we leave the rates out of the signature of this method, and rather attach them to the `TaxedItem` object? Because external services do not have actual rates.

adammathys: It currently is an array of `TaxedItem` objects. This looks to just be a on old name that got left behind in one of my rebases. I'll update it.

core/app/models/spree/app_configuration.rb
```diff
@@ -309,6 +309,16 @@ def shipping_rate_taxer_class
       @shipping_rate_taxer_class ||= Spree::Tax::ShippingRateTaxer
     end
 
+    # Allows providing your own class for calculating taxes on a shipping rate.
+    #
+    # @!attribute [rw] shipping_rate_tax_calculator_class
+    # @return [Class] a class with the same public interfaces as
+    #   Spree::TaxCalculator::ShippingRate
+    attr_writer :shipping_rate_tax_calculator_class
```
jordan-brough: Could we remove this until we've had a chance to look at options on the interface a bit?

adammathys: Do we want to remove both configuration points? Or just the shipping rate one?

core/app/models/spree/tax_calculator/shipping_rate.rb
```diff
@@ -0,0 +1,43 @@
+module Spree
```
kennyadsl: What about moving tax calculators folder into `models/spree/calculator`?

core/app/models/spree/app_configuration.rb
jordan-brough: Oops, yes, I meant both.

The PR looks really great! I'd love for us to merge it but it would be great to have a chance to collaborate (maybe even at SolidusConf?) on the interface before committing to it.

If that sounds OK could we perhaps also mark these two new classes as `@api private` for the moment, as we had done for the `Tax::ItemAdjuster` class?

jordan-brough:
Commenting here to connect this to issue #1892, which is advanced nicely by this PR

core/app/models/spree/app_configuration.rb
jhawthorn: I am becoming less convinced by the need for an "input"-type class, but we should absolutely discuss it at SolidusConf. I think marking it `@api private` for now is a good compromise.

WDYT @adammathys ?

adammathys: Definitely. I've marked both classes and their configuration points with `@api experimental`. (Which is what we used for the `tax_adjuster_class` configuration point.) I also added in a note to each class explaining that they're still in development and that the input specifically is likely to change.

core/app/models/spree/tax_calculator/shipping_rate.rb
adammathys: I didn't put the new tax calculators inside of `models/spree/calculator` because they aren't like the rest of the calculators in there. All of those calculators inherit from [`Spree::Calculator`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/calculator.rb) and as such have an already established interface.

The two new tax calculators share only the description of being a "calculator" with everything else in that namespace. Perhaps this is an indication that we should use a different name altogether for these new classes, because they are not meant to replace the existing calculators. (And the default implementations still use the [`Spree::Calculator::DefaultTax`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/calculator/default_tax.rb) at the lowest level to compute the tax amount.)

core/app/models/spree/order_taxation.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      tax_adjustment ||= item.adjustments.new(
+        source: tax_item.tax_rate,
+        order_id: item.order_id,
+        label: tax_item.label,
+        included: tax_item.included_in_price
+      )
+
+      unless tax_adjustment.finalized?
+        tax_adjustment.update(amount: tax_item.amount)
```
jordan-brough: If this fails where do we find out about it?

core/app/models/spree/order_taxation.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      tax_adjustment ||= item.adjustments.new(
+        source: tax_item.tax_rate,
+        order_id: item.order_id,
+        label: tax_item.label,
+        included: tax_item.included_in_price
+      )
+
+      unless tax_adjustment.finalized?
```
jordan-brough: This is likely outside the scope of this PR (since the same thing happens in the old code as of v2.1) but is this really what we want?  E.g. if an order is complete and I change the quantity (and thus price) of a line item, don't I need to update the VAT amount I have recorded for that item?

core/app/models/spree/tax/taxed_item.rb
```diff
@@ -0,0 +1,20 @@
...
+    # an item. (Either a {Spree::LineItem} or a {Spree::Shipment}.)
+    #
+    # @attr_reader [Integer] id the {Spree::LineItem} or {Spree::Shipment} ID
+    # @attr_reader [String] label information about the taxes
+    # @attr_reader [Spree::TaxRate] tax_rate will be used as the source for tax
+    #   adjustments
+    # @attr_reader [BigDecimal] amount the amount of tax applied to the item
+    # @attr_reader [Boolean] included_in_price whether the amount is included
+    #   in the items price, or additional tax.
+    class TaxedItem
```
jordan-brough: Would `ItemTax` be a clearer name for this? (Same question for `TaxedOrder`)

core/app/models/spree/tax/taxed_item.rb
```diff
@@ -0,0 +1,20 @@
+module Spree
+  module Tax
+    # Simple object used to hold tax data for an item.
+    #
+    # This generic object will hold the amount of tax that should be applied to
+    # an item. (Either a {Spree::LineItem} or a {Spree::Shipment}.)
+    #
+    # @attr_reader [Integer] id the {Spree::LineItem} or {Spree::Shipment} ID
```
jordan-brough: How about `item_id` instead of `id`? (Same question for the id in `TaxedOrder`)

core/app/models/spree/order_taxation.rb
jordan-brough: To me at least it doesn't seems clear that I need to go unlock tax adjustments before modifying completed orders.  (Either via the Admin or some other way)  Again, possibly outside the scope of this PR, but between v1.4 and v2.1 at least we did update taxes on completed orders, if I understand the code correctly.

adammathys: I believe this is out of scope of this PR. The goal isn't to change any of the current behavior, just to add in a new extension point. 

That being said, I agree that it doesn't make sense that taxes aren't recalculated on completed orders. I'll open a second PR to propose a reversion of that change.

core/app/models/spree/order_taxation.rb
adammathys: Ideally, this shouldn't ever fail. I'll update this to `update!` so that if it ever does it makes a noise.

core/app/models/spree/order_taxation.rb
jordan-brough: 👍 absolutely agreed. FYI I opened a PR just today actually: https://github.com/solidusio/solidus/pull/1917

jhawthorn:
:shipit:

PR #1895, opened 2017-05-08
Rails 5.1

jhawthorn:


jhawthorn:
We're now ready for Rails 5.1. All deps we need have releases compatible with rails 5.1 :tada:

I'll update this PR to also be compatible with rails 5.0

jarednorman:
This all looks quite sensible. :100:

ericsaupe:
Is it worth updating all form tags to use the new `form_with`? http://edgeguides.rubyonrails.org/5_1_release_notes.html#unification-of-form-for-and-form-tag-into-form-with. I know the Rails core team haven't said anything about deprecating current form tags but this is the unification of the two.

jarednorman:
I can't imagine it's worth the effort of doing that up front if the existing helpers aren't being deprecated.

ericsaupe:
@jarednorman that's what I thought as well. Just thought I'd bring it up!

peterberkenbosch:
👍

PR #1900, opened 2017-05-09
Prepare for turbolinks

mtomov:
- Corrects minor duplications in Capybara page visits
- Removes unnecessary calls to jQuery.ready
- Change to Spree.ready from jQuery.ready on a few places only

Towards #1863

based on suggestions from #1882 

_note: I have not fixed the indentation in `admin.js` not to introduce noise._

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -14,7 +14,7 @@ under the spree namespace that do stuff we find helpful.
 Hopefully, this will evolve into a propper class.
 **/
 
-jQuery(function($) {
+Spree.ready(function() {
   // Highlight hovered table column
   $('table').on("mouseenter", 'td.actions a, td.actions button', function(){
     var tr = $(this).closest('tr');
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -159,48 +159,44 @@ $(document).ready(function(){
       return ui;
   };
 
-  $('table.sortable').ready(function(){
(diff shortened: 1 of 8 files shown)


backend/spec/features/admin/promotion_adjustments_spec.rb
```diff
@@ -8,6 +8,7 @@
       visit spree.admin_path
       click_link "Promotions"
       click_link "New Promotion"
+      expect(page).to have_content("PromotionsNew Promotion")
```
jhawthorn: This will probably be cleaner if we assert on the title (which uses the same values now). I think drivers other than poltergeist sometimes insert a space between these elements :thinking: 

```
expect(page).to have_title("New Promotion - Promotions")
```

If this doesn't work, though, no big deal.

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -159,7 +159,6 @@ $(document).ready(function(){
       return ui;
   };
 
-  $('table.sortable').ready(function(){
     var td_count = $(this).find('tbody tr:first-child td').length
```
jhawthorn: The indentation here is now off

tvdeyen: > note: I have not fixed the indentation in admin.js not to introduce noise

but, yeah, we should fix that 👍

mtomov: no problem at all!, thanks both for reviewing!

The `have_title` suggestion was great, and I've gone about to actually change all those breadcrumb expectations to `have_title`, as when I was fixing those tests the other day, it took me a while to understand where "PromotionsPromotion" comes from.

All changed now.

Issue #1903, opened 2017-05-11
"Show Deleted" checkbox in /admin/products doesn't stay checked when appropriate


spaghetticode:
When filtering products in the admin page the checkbox `Show Deleted`, if checked, doesn't remain checked after form submission. 

This causes also a problem with pagination, you can't reach any other page expect the first with deleted products.

spaghetticode:
The reason is due to the following code in `backend/app/controllers/spree/admin/products_controller.rb:105`: 
```
@collection = @collection.with_deleted if params[:q].delete(:deleted_at_null) == '0'
```
I would change it as follow: 
```
@collection = @collection.with_deleted if params[:q][:deleted_at_null] == '0'
```
before opening a new PR with my changes I would like to ask you if there's a good reason for deleting `deleted_at_null` from `params[:q]` in that line. Thanks!

mtomov:
https://github.com/solidusio/solidus/commit/de2065b seems to be the commit to introduce the `delete()`. I guess if you are up for writing a test for this behaviour, there's nothing to stop you from making that change.

Actually, isn't it possible to directly use the `ransackable_scopes` to allow this `with_deleted` https://github.com/activerecord-hackery/ransack#using-scopesclass-methods

spaghetticode:
@mtomov I'm not sure I completely understood your last sentence. Did you mean that, instead of using `deleted_at_null` we may use a (new?) scope on the model, which we also use as a ransackable scope, and then we also change the param name on the `erb` view?

Anyway, I choose the simplest path for my PR: remove the `delete` method call. I also added a simple controller spec for it.

spaghetticode:
@mtomov with a little help from @kennyadsl I think I finally got your point. I am working on a new implementation that uses ransackable scopes, which has the extra benefit of tidying up a bit the code in the controller. I think that's an improvement over my first version, so I will update my PR ASAP. But if you don't prefer the new version then I can still go back to the first proposal ;)

mtomov:
Sure, take your time.

I'm just not sure if you will be able to capture the existing bug of the "not-checked" checkbox after searching with a `controller` test. Maybe a `feature` test might do one better.

spaghetticode:
Thanks for the suggestion, I've just added also an expectation in an existing feature spec as well.

mamhoff:
Fixed via #1904

PR #1904, opened 2017-05-12
Show deleted

spaghetticode:
Fix for issue #1903

### DIFF
--- a/backend/app/views/spree/admin/products/index.html.erb
+++ b/backend/app/views/spree/admin/products/index.html.erb
@@ -34,7 +34,7 @@
           <div class="col-2">
             <div class="field checkbox">
               <label>
-                <%= f.check_box :deleted_at_null, {checked: params[:q][:deleted_at_null] == '0'}, '0', '1' %>
+                <%= f.check_box :with_deleted, { checked: params[:q][:with_deleted] == 'true' }, 'true', 'false' %>
                 <%= Spree.t(:show_deleted) %>
               </label>
             </div>
--- a/backend/spec/controllers/spree/admin/products_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/products_controller_spec.rb
@@ -13,6 +13,29 @@ describe Spree::Admin::ProductsController, type: :controller do
       expect(assigns[:collection]).not_to be_empty
       expect(assigns[:collection]).to include(product)
     end
+
+    # Regression test for https://github.com/spree/spree/issues/1903
+    context 'when soft deleted products exist' do
+      let!(:soft_deleted_product) { create(:product, sku: "ABC123").destroy }
+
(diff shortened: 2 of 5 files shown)


backend/app/views/spree/admin/products/index.html.erb
```diff
@@ -34,7 +34,7 @@
           <div class="col-2">
             <div class="field checkbox">
               <label>
-                <%= f.check_box :deleted_at_null, {checked: params[:q][:deleted_at_null] == '0'}, '0', '1' %>
+                <%= f.check_box :with_deleted, {checked: params[:q][:with_deleted] == 'true'}, 'true', 'false' %>
```
kennyadsl: I'd take advantage of this change to add spaces inside `{}` to be consistent with the rest of the code .

backend/spec/controllers/spree/admin/products_controller_spec.rb
```diff
@@ -13,6 +13,28 @@
...
+
+        it 'filters out soft-deleted products by default' do
+          get :index, params: params
+          expect(assigns[:collection]).to_not include(soft_deleted_product)
+        end
+      end
+
+      context 'when params[:q][:with_deleted] is set to "true"' do
+        let(:params) { { q: { with_deleted: 'true' } } }
+        it 'includes soft-deleted products' do
```
kennyadsl: I'd leave an empty line after `let` here

core/app/models/spree/product.rb
```diff
@@ -103,6 +103,11 @@ def find_or_build_master
     self.whitelisted_ransackable_associations = %w[stores variants_including_master master variants]
     self.whitelisted_ransackable_attributes = %w[slug]
 
+    # This enables "Show Deleted" checkbox feature in admin/products page
```
kennyadsl: I'm not sure about this line of comment. If we'll add other scopes later this line had to be removed/changed in something more generic. I'm for removing it since we can understand why this code is here by using git history.

backend/spec/controllers/spree/admin/products_controller_spec.rb
kennyadsl: I'd leave an empty line after `let` here

core/app/models/spree/product.rb
kennyadsl: I'm not sure about this comment line. If we'll add other scopes later this line had to be removed/changed in something more generic. I'm for removing it since we can understand why this code is here by using git history.

backend/app/views/spree/admin/products/index.html.erb
kennyadsl: I'd take advantage of this change to add spaces inside `{}` to be consistent with the rest of the code .

mamhoff:
Thank you!

PR #1906, opened 2017-05-13
Allow refreshing rates for unshipped complete orders

mamhoff:
When an order is complete, but not shipped yet, admins need to be able
to refresh rates for a shipment.

This is especially the case when splitting shipments to a new location.
This will create a new shipment with no rates, and refreshing them is the
only way to get any.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Solidus 2.3.0 (master, unreleased)
 
+- Allow refreshing shipping rates for unshipped shipments on completed orders [\#1906](https://github.com/solidusio/solidus/pull/1906) ([mamhoff](https://github.com/mamhoff))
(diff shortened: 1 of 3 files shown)


mamhoff:
Added a changelog entry as well as a spec for the intended behaviour.

jhawthorn:
Just noting that previously `refresh_rates` was always called on completed orders 4ad32da96500ce56ea75aedeb4d3f868bf269eb5. That might make this a less-scary change, but it might also make this change not achieve what is desired (correct updating of rates when the amount changes).

mamhoff:
I part from the idea that it is possible to change the line items of an order when it is completed but not shipped yet. If that is possible, then the shipments must be able to change, too, as the line items will need to be shipped somehow.

mamhoff:
Rebased on current master.

mamhoff:
@jhawthorn This is primarily intended to make splitting shipments for completed orders work. Right now it doesn't at all because the new shipment won't have rates.

PR #1915, opened 2017-05-22
Deprecate simple_current_order

### LINKED ISSUE #1700: simple_current_order creates an Order without a Store
### Steps to reproduce
Try to access `available_payment_methods` from an order created by `simple_current_order` for a new user.

### Expected behavior
This should provide the available_payment_methods.

### Actual behavior
Can't find `payments`.

This is called to create the order:
```ruby
@simple_current_order = Spree::Order.new
```
(https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/controller_helpers/order.rb#L27)

If trying to access `available_payment_methods`, the call fails because of this:
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L448

What fixed it for me is to associate the order with the store:
```ruby
@simple_current_order = Spree::Order.new(store: current_store)
```

I can PR if confirmed

### System configuration
**Solidus Version**:
`master`, `v2.1`

ericsaupe:
Fixes the bug found in #1700 and deprecates `simple_current_order` because it's function is duplicated with `current_order`

### DIFF
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -24,7 +24,7 @@ module Spree
             @simple_current_order.last_ip_address = ip_address
             return @simple_current_order
           else
-            @simple_current_order = Spree::Order.new
+            @simple_current_order = Spree::Order.new(current_order_params)
           end
         end
 
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -14,7 +14,6 @@ module Spree
           helper_method :simple_current_order
         end
 
-        # Used in the link_to_cart helper.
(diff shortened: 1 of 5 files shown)


core/lib/spree/core/controller_helpers/order.rb
```diff
@@ -14,8 +14,9 @@ module Order
           helper_method :simple_current_order
         end
 
-        # Used in the link_to_cart helper.
         def simple_current_order
```
mamhoff: You can use 

```
deprecate simple_current_order: :current_order, deprecator: Spree::Deprecation
``` 
Which does the same thing more concisely.

ericsaupe: I like this a lot better, I'll make the change. Thanks!

Issue #1916, opened 2017-05-22
Tax adjustments aren't updated on completed orders


jordan-brough:
Existing tax adjustments aren't updated if an order is changed after it is complete.  This is because [we call `Adjustment#update!`](https://github.com/solidusio/solidus/blob/v2.2.1/core/app/models/spree/tax/item_adjuster.rb#L36) which quits if the adjustment is finalized (which happens automatically when an order completes).

Issue #1894, opened 2017-05-08
Taxation without adjustments


mamhoff:
Currently, we use the `Spree::Adjustment` model to save taxation data on line items, shipments and orders. 

Adjustments have issues: 

* They store labels in the database, making them hard to use across different locales
* They need a `source` and are recalculated every time the order changes. For taxation, that's not so desirable as we are moving towards doing taxation for the order all in one go. 
* They are polymorphic

Here's a proposal: Add a new thing called `Spree::Tax` that works just like an adjustment, except that it is somehow translatable through `I18n`, cannot be used for promotions, and is tied to a `Spree::Shipment` exclusively.

If line items need taxation, they can have a line item tax estimate, too - as a separate model.

PR #1917, opened 2017-05-22
Unfinalize tax adjustments before updating them

jordan-brough:
For #1916.

Also related to #1894.

Otherwise taxes don't get updated after an order completes (because all
adjustments are finalized when an order completes).

A different option:  Ignore `finalized` for all tax adjustments in `Adjustment#update!`?

### DIFF
--- a/core/app/models/spree/tax/item_adjuster.rb
+++ b/core/app/models/spree/tax/item_adjuster.rb
@@ -31,7 +31,7 @@ module Spree
           # Find an existing adjustment from the same source.
           # All tax adjustments already have source_type == 'Spree::TaxRate' so
           # we need only check source_id.
-          adjustment = tax_adjustments.detect{|a| a.source_id == rate.id }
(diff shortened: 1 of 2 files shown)


mamhoff:
@jordan-brough I think @gmacdougall s concern is valid. Seems like we need to add another line to that method. :)

jordan-brough:
@gmacdougall hm, it is not.  I wonder if my alternate suggestion would be better...?

> A different option: Ignore finalized for all tax adjustments in `Adjustment#update!`?

Do we really want to prevent tax adjustments from updating?

mamhoff:
We discussed IRL that both @jordan-brough and I agree that changing the semantics of `finalized` to never be applicable to tax rates on unshipped orders makes sense. If we have to unfinalize and finalize within the same method we're only adding complexity and database calls to the code at no advantage in code clarity.

jordan-brough:
@mamhoff @gmacdougall I've opened #1936 to replace this.

Issue #721, opened 2016-01-21
Options from products show not passed to orders_controller when product added to cart


cindyward1:
There is a bug in the populate method of frontend/app/controllers/spree/orders_controller.rb that prevents the options of the add to cart action of the product show view from being passed to the order.contents.add method.

The code in question was originally written in Spree when the OrderPopulator class was removed with Spree commit [c2870cf](https://github.com/spree/spree/commit/c2870cfa2fa22e91e8529edb0136c21cfa1b4e2a#diff-b186dacad723e791bc7a32692c875066). The code was then modified to remove the merge of the options argument with current_currency with commit [4e4d5b2](https://github.com/spree/spree/blob/4e4d5b2653cb739d348f43aa7c69f30fce695f0a/frontend/app/controllers/spree/orders_controller.rb).

The code from Spree commit c2870cf was modified incorrectly in Solidus with commit [40e4db2](https://github.com/solidusio/solidus/commit/40e6b89a0d78185d4f89afdca20561f5a3a97ba5#diff-b186dacad723e791bc7a32692c875066), which completely eliminated passing the options argument to the order.contents.add method. It should have been modified like it was in Spree commit 4e4d5b2 to read as shown below:

``` ruby
  # Adds a new item to the order (creating a new order if none already exists)
  def populate
      order    = current_order(create_order_if_necessary: true)
      variant  = Spree::Variant.find(params[:variant_id])
      quantity = params[:quantity].to_i
      options = params[:options] || {}   # This line needs to be added

      #2,147,483,647 is crazy. See issue https://github.com/spree/spree/issues/2695.
      if quantity.between?(1, 2_147_483_647)
        begin
            # options argument needs to be added to the order.contents.add method call
            order.contents.add(variant, quantity, options) 
```

We are trying to port spree_product_assembly to Solidus, and this bug is causing 4 test cases to fail with the error "Failure/Error: selected_variants[part.id.to_s], NoMethodError: undefined method `[]' for nil:NilClass". This same error also causes an abort when using the UI to add a bundle with a selection-deferred variant to the shopping cart.

cbrunsdon:
Hi @cindyward1, thanks for doing such a thorough backtrack of that codes history.

Do you have your WIP of the spree_product_assembly port handy? I'm unsure what is actually throwing the error here, as order.contents.add() sets the default for options to be {}, so it should never be nil and we should not need to set it inside of the populate method.

cindyward1:
[...]

PR #1918, opened 2017-05-22
Pass params[:options] through Spree::OrderContents#add

acreilly:
Now options can be passed to Spree::OrderContents#add which passes to Spree::OrderContents#add_to_line_item
#721

### DIFF
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -45,6 +45,7 @@ module Spree
       @order   = current_order(create_order_if_necessary: true)
       variant  = Spree::Variant.find(params[:variant_id])
       quantity = params[:quantity].present? ? params[:quantity].to_i : 1
+      options = params[:options] || {}
 
       # 2,147,483,647 is crazy. See issue https://github.com/spree/spree/issues/2695.
       if !quantity.between?(1, 2_147_483_647)
@@ -52,7 +53,7 @@ module Spree
       end
 
       begin
-        @line_item = @order.contents.add(variant, quantity)
+        @line_item = @order.contents.add(variant, quantity, options)
       rescue ActiveRecord::RecordInvalid => e
         @order.errors.add(:base, e.record.errors.full_messages.join(", "))
       end


jordan-brough:
Specs?

gmacdougall:
Can anyone think of reasons why we shouldn't accept user-defined input here? The line item options are validated with this: https://github.com/solidusio/solidus/blob/v2.2.1/core/app/models/spree/order_contents.rb#L99

but that's only one part of the thing that options are used for.

acreilly:
@gmacdougall Can't think of anything when it comes to solidus out of the box. I came across this issue when working on SolidusFlexiVariants where I had to override the method. https://github.com/boomerdigital/solidus_flexi_variants/blob/master/app/controllers/spree/orders_controller_decorator.rb#L16

@cindyward1 Also noticed it while working on a gem.

I guess I just find it odd that it is offered as an option, but never passed on.

mamhoff:
I think this is a beneficial change. One way to test it would be a controller test with stubbing `order.contents`, expecting that that double receives `add` with the options. Given that we allow this as an extension point but don't use it in core, there's no "real" way of testing it that does not involve a disproportiate amount of spec setup.

cindyward1:
I talked to Allison, and I'm working on some specs.

jhawthorn:
@gmacdougall wrote:
> The attributes should be validated at the lower level

I'm not sure, I think they should maybe be validated at this level via strong params, as they are [in the API](https://github.com/solidusio/solidus/blob/054e4bd75c65166d316f3f63be85efb48736bcdb/api/app/controllers/spree/api/line_items_controller.rb#L72)

cindyward1:
[The documentation for Spree::LineItem](http://docs.solidus.io/Spree/LineItem.html) says that `#options=`  "Sets options on the line item and updates the price. The options can be arbitrary attributes on the LineItem."

The ability to set arbitrary line item options does not seem to be tested in core/spec/models/spree/order_contents_spec.rb or core/spec/models/spree/line_item_spec_rb. line_item_spec.rb tests options that update price but not arbitrary options. order_contents_spec.rb appears to ignore options altogether.

@mamhoff may be right about the disproportionate amount of spec setup.

jhawthorn:
> The documentation for Spree::LineItem says that #options= "Sets options on the line item and updates the price. The options can be arbitrary attributes on the LineItem."

This is exactly the problem here. If we pass arbitrary user input to `#options=`, we allow users to set arbitrary options on line item.

I'm closing this, but feel free to open a new PR if it sanitizes the user's input.

PR #1920, opened 2017-05-22
Add data check to OrderController#populate for OrderContent#add quantity

dgra:
Should fix https://github.com/solidusio/solidus/issues/1726

### DIFF
--- a/frontend/spec/controllers/spree/orders_controller_spec.rb
+++ b/frontend/spec/controllers/spree/orders_controller_spec.rb
@@ -68,6 +68,34 @@ describe Spree::OrdersController, type: :controller do
             Spree.t(:please_enter_reasonable_quantity)
           )
         end
+
+        context "when quantity is empty string" do
+          it "should populate order with 1 of given variant" do
+            expect do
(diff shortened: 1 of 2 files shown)


frontend/spec/controllers/spree/orders_controller_spec.rb
```diff
@@ -68,6 +68,34 @@
             Spree.t(:please_enter_reasonable_quantity)
           )
         end
+
+        context "when quantity is empty string" do
+          it "should populate order with 1 of given variant" do
+            expect do
+              post :populate, params: { variant_id: variant.id, quantity: '' }
+            end.to change { user.orders.count }.by(1)
```
mamhoff: This tests the association of the order to the current user. If I understand correctly, the change this introduces does not have anything to do with the user, so we can harden the spec by only specifying that `post`ing changes the `Spree::Order`.count.

dgra: Good point. I will reduce the scope. Thank you!

vfonic:
Thank you!!! :) 

Really needed this and I overcomplicated it when I tried to fix it.

Issue #1919, opened 2017-05-22
Empty Cart does not remove coupon code from order


seand7565:
### Steps to reproduce
Create promotion of any kind
Add product to cart
Add promotion to cart
Empty cart
Add product to cart

### Expected behavior
There should be no promotions active

### Actual behavior
The previously entered promotion is still active

### System configuration
**Solidus Version**:Any

**Extensions in use**:None

seand7565:
This should be fixed by changing these lines in models/spree/order.rb from:

`    def empty!
      line_items.destroy_all
      adjustments.destroy_all
      shipments.destroy_all

      update!
    end`

to 
`    def empty!
      line_items.destroy_all
      adjustments.destroy_all
      shipments.destroy_all
      promotions.destroy_all

      update!
    end`

Will test and PR

gmacdougall:
This has been resolved by #1932

PR #1921, opened 2017-05-22
Destroy promotions on empty cart 

seand7565:
Addresses #1919

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -475,6 +475,7 @@ module Spree
       line_items.destroy_all
       adjustments.destroy_all
       shipments.destroy_all
+      promotions.destroy_all
 
       update!
     end


mamhoff:
I think this removes the promotion that created the promotion adjustment, not the adjustment. This would mean that if we have a promotion that applies to a thousand carts and a user clears their cart, that promotion would fail to apply for all remaining orders. I would also delete all promotion codes associated with it. 

I think it'd be good to look into what happens here exactly, because the existing code should destroy all adjustments for this order.

seand7565:
I do not believe it is the case that this code removes the promotion itself - we've been using this line of code on our local branch for the last year with no problems. I've tested this in console - after emptying the cart, the promotion still exists, however the order no longer has the promotion in Spree::Order.find(order_num).promotions

My current understanding of this issue is, while emptying the cart does clear the adjustment created by the promotion, the promotion is still associated to the order. The next time the order updates (when you add an item) the promotion re-applies the adjustment. Using promotions.destroy_all destroys all promotions in the order (but does not affect the promotion as a whole), which prevents the promotion from re-applying itself on cart update.

I'm working on tests now - will also look into this a bit more to make sure I have 100% understanding before pushing the tests.

seand7565:
Closing PR because I don't know how to Github. :no_good_man: Will submit a new PR in two seconds.

PR #1923, opened 2017-05-22
make it consistence with OPERATORS.

ranxiang:
In file: backend/app/views/spree/admin/promotions/rules/_item_total.html.erb line 5, it used ItemTotal::OPERATORS to render the form key of preferred_operator, and only can be 'gt' or 'gte'. but in the model item_total.rb, the default value of preferred_operator is '>', that's inconsistence and we'd better keep it same.

### DIFF
--- a/core/app/models/spree/promotion/rules/item_total.rb
+++ b/core/app/models/spree/promotion/rules/item_total.rb
@@ -6,7 +6,7 @@ module Spree
       class ItemTotal < PromotionRule
         preference :amount, :decimal, default: 100.00
         preference :currency, :string, default: ->{ Spree::Config[:currency] }
-        preference :operator, :string, default: '>'
+        preference :operator, :string, default: 'gt'
 
         OPERATORS = ['gt', 'gte']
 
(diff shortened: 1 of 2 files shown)


mamhoff:
This class is already well-tested, and will work with both default preferences. I think we don't need another test for this actually, but I whole-heartedly agree that adding a more helpful commit message would be great. Here's an article describing what great commit messages look like: https://chris.beams.io/posts/git-commit/

mamhoff:
A simple spec to test this would be

```ruby
RSpec.describe Spree::Promotion::Rules::ItemTotal do 
  let(:rule) { described_class.new }
  
  describe "#preferred_item_total" do 
    subject { rule.preferred_item_total }
    it { is_expected.to eq('gt') }
  end
end
```

We just discussed this IRL and both think this would make sense to add in.

ranxiang:
comment and spec added, please review.
This is the first time for me to open a pull request, let me know if anything still unclear.
Thanks.

mamhoff:
I'm fine adding this in. Please squash your commits into one, and try making the commit message for this commit adhere to [Chris Beams blog post](https://chris.beams.io/posts/git-commit/)

ranxiang:
Just saw the comment due to the time differences, thanks all, I'll improve the comment and squash the commits before create the pull request next time.

mamhoff:
You don't need to open a new PR, but can use `git push --force` once you have squashed your commits. Btw the build will also restart when you do this, so I'm not retriggering now.

mamhoff:
@ranxiang any news on this?

mamhoff:
Closing as stale.

PR #1924, opened 2017-05-22
Cast string values from params into boolean.

### LINKED ISSUE #1737: Stock Transfer receive dropdown autocomplete fails when all of stock in transit
I’m on the latest gem release of `2.0.x`. If I transfer all of the stock of a given product from one warehouse to another, I can’t "receive" the package correctly. The product is out of stock in all locations while it’s in transit, and when I go to log the received stock, the dropdown won’t show that product.

My workaround: go to Store Stock, manually change quantity at my receiving warehouse to 1, use the dropdown, change quantity manually back to 0, then receive the shipment.

gus4no:
This fixes #1737 

The issue was that the param was being set at `"false"` which is would add the `in_stock` scope to the variants because `"false" != false`.

So @jtapia and I ended up adding the query string param only if the condition was true.

### DIFF
--- a/api/spec/controllers/spree/api/variants_controller_spec.rb
+++ b/api/spec/controllers/spree/api/variants_controller_spec.rb
@@ -88,27 +88,56 @@ module Spree
       end
 
       context "stock filtering" do
-        subject { api_get :index, in_stock_only: true }
 
-        context "variant is out of stock" do
-          before do
-            variant.stock_items.update_all(count_on_hand: 0)
+        context "only variants in stock" do
+          subject { api_get :index, in_stock_only: "true" }
+
+          context "variant is out of stock" do
+            before do
+              variant.stock_items.update_all(count_on_hand: 0)
+            end
+
+            it "is not returned in the results" do
+              subject
+              expect(json_response["variants"].count).to eq 0
+            end
           end
 
-          it "is not returned in the results" do
-            subject
-            expect(json_response["variants"].count).to eq 0
+          context "variant is in stock" do
+            before do
+              variant.stock_items.update_all(count_on_hand: 10)
+            end
+
+            it "is returned in the results" do
+              subject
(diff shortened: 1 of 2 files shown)


gmacdougall:
Could you help my understanding how you reproduced the issue? Taking a look at how this is initialized, I can only see it getting initialized with boolean values (example: https://github.com/solidusio/solidus/blob/v2.2.1/backend/app/assets/javascripts/spree/backend/stock_transfers/receive.coffee#L3) and not seeing where the string "false" came from.

gus4no:
@gmacdougall About what we talked, what do you think of using this

`in_stock_only = ActiveRecord::Type::Boolean.new.cast(params[:in_stock_only])`

gmacdougall:
I think that makes sense.

gus4no:
@gmacdougall what do you think now?

gmacdougall:
I think this is great. I might consider adding another spec for the specific case that you were having trouble with. For example, when `in_stock_only` is `"false"`

gus4no:
@gmacdougall Tests added

gus4no:
@tvdeyen Squashed! :)

api/spec/controllers/spree/api/variants_controller_spec.rb
```diff
@@ -88,27 +88,56 @@ module Spree
...
+          subject { api_get :index, in_stock_only: "false" }
+
+          context "variant is out of stock" do
+            before do
+              variant.stock_items.update_all(count_on_hand: 0)
+            end
+
+            it "is not returned in the results" do
+              subject
+              expect(json_response["variants"].count).to eq 1
```
gmacdougall: This expectation doesn't seem to match the description here.

gus4no:
@gmacdougall I think it's ready now!

PR #1925, opened 2017-05-22
Allow to apply promotions if at least one line item is eligibile

### LINKED ISSUE #1437: Promotion eligibility? ambiguous behavior
On 1.3 (and on):
1. Create a coupon promotion that applies a discount to line items
2. Add a non-promotionable product to the cart
3. Add a promotionable product to the cart
4. Apply coupon

Expected behavior: the promotion applies to one line item but not the other
Got: the order is ineligible for this coupon

This is being missed due to a logical discrepancy in the [spec](https://github.com/solidusio/solidus/blob/v1.3/core/spec/models/spree/promotion_spec.rb#L511-L528):
there are two test cases in "eligible?" that aren't both possible:  
"and at least one item is non-promotionable" should be return false
"and at least one item is promotionable" should return true

What should the case where one is promotionable and is isn't? Both cases can't be right.

I'm guessing this arises from having both Order and LineItem level promotion actions, so I'm not sure how to resolve. Looking for guidance and I'll work on a PR.

softr8:
Fixes #1437

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - Adjustments without a source are now included in `line_item.adjustment_total`
 [\#1933](https://github.com/solidusio/solidus/pull/1933) ([alexstoick](https://github.com/alexstoick))
 
+- Add `Spree::Config.allow_promotions_with_non_promotionable_items` to control promotions eligibility when having a non promotional line item [\#1925](https://github.com/solidusio/solidus/pull/1925) ([softr8](https://github.com/softr8))
+
 ## Solidus 2.2.1 (2017-05-09)
 
 - Fix migrating CreditCards to WalletPaymentSource [\#1898](https://github.com/solidusio/solidus/pull/1898) ([jhawthorn](https://github.com/jhawthorn))
--- a/core/app/models/spree/app_configuration.rb
+++ b/core/app/models/spree/app_configuration.rb
@@ -46,6 +46,10 @@ module Spree
     #   @return [Boolean] When false, customers must create an account to complete an order (default: +true+)
     preference :allow_guest_checkout, :boolean, default: true
 
+    # @!attribute [rw] allow_promotions_with_non_promotionable_items
+    #   @return [Boolean] When true, customers can apply a promotion when the order has non promotionable items (default: +false+)
+    preference :allow_promotions_with_non_promotionable_items, :boolean, default: false
+
     # @!attribute [rw] allow_return_item_amount_editing
     #   @return [Boolean] Determines whether an admin is allowed to change a return item's pre-calculated amount (default: +false+)
     preference :allow_return_item_amount_editing, :boolean, default: false
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -237,8 +237,13 @@ module Spree
       when Spree::LineItem
         !promotable.product.promotionable?
       when Spree::Order
-        promotable.line_items.any? &&
-          promotable.line_items.joins(:product).where(spree_products: { promotionable: false }).any?
+        should_be_blacklisted = if Spree::Config.allow_promotions_with_non_promotionable_items
+                                  !promotable.line_items.joins(:product).where(spree_products: { promotionable: true }).exists?
+                                else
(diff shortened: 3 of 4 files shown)


mamhoff:
This makes sense to me: If we have any promotable line items, order level promotions should apply. The promotion rule can then decide what total to use, but generally the rules should come into effect if any line items are promotable. 

Could you be a little more elaborate in your commit message about what this change entails?

jhawthorn:
I don't think we should do this. I think that setting `promotionable` to `false` is intended to be a very hard "do not under any circumstances allow promotions on this item". Specifically, I think the use case is for cash-equivalent items (like gift cards), which we wouldn't want to allow promotions on just because there are other `promotionable` items in the cart.

gmacdougall:
What if we allow the behaviour to be configurable? Leave the default behaviour as is, but provide an option to allow people to implement the new behaviour based off of a Spree::Config attribute.

softr8:
Done! I was unsure how to name the config variable, suggestions welcomed!

softr8:
what do you think @gmacdougall ?

mamhoff:
@gmacdougall this is now configurable, and I still think this is a beneficial change. Would you mind re-evaluating?

jhawthorn:
Sorry, but I don't want this behaviour even if this is configurable.

Setting `promotionable=false` should be a hard guarantee that no promotions will ever apply to this item.

The intended use case of `promotionable=false` is to allow selling items like gift cards safely. If a customer buys a $100 gift card and a $1 pen, they shouldn't be able to receive a "$20 of orders over $100" discount. This is an extremely important rule to enforce, and though this would be a configuration option, I'm just not comfortable making it possible for that to be misconfigured.

Extension authors (ex. of a gift card extension) might want to have their extension create a `promotionable=false` item, and should have a guarantee that that works consistently in all stores.

To allow the behaviour this PR provides (promotions for some items in a cart), I believe it would be better to only allow it using item-level promotions.

fylooi:
@jhawthorn 
Just spent a few hours trying to figure out why my automatic free shipping promotion was failing sporadically. Finally traced it to one product with `promotionable=false`. Would love to have this as an option.

jacobherrington:
@nspinazz89 and I are revisiting this issue. I think this approach isn't terrible because putting this in a configuration option should give people control over how this works, especially if we default to enforcing the current behavior, right?

I'd love to get thoughts from reopening or borrowing from this PR? @solidusio/core-team

kennyadsl:
I see John's concerns here: setting that preference to true could break some behavior added into extensions or stores code.

On the other side, I searched in [`solidus-contrib` for promotionable](https://github.com/search?q=org%3Asolidusio-contrib+%22promotionable%22&type=Commits) but at a quick look I can't see any usage of that attribute in any extensions. 

I think that we should just better enforce what could happen if you enable the preference in the preference documentation, probably in the [YARD comment](https://github.com/solidusio/solidus/pull/1925/files#diff-a9fae1594c22d1ec005ac8501d80d7afR49), and we are good to go.

cedum:
Both (#3339, #1437) reported issues are related to line-items level promos: being unable to apply a line-item level promo seems indeed a bug.

But I don't think this should be applicable also for order-level promos: I expect an order promo to NOT be eligible if one of its items are not eligible. There're line-items promos if one needs a more granular control. Introducing this (order-level promos change) into the core, even behind a config flag, imho seems a "hack" for an edge case.

nspinazz89:
This is just an idea so feel free to shoot it down. What if we had order promotions operate on line items by default. Then say if a customer had a 20% off on your order with 3 items, only 2 of which are eligible for promotion, it would apply the 20% to both of those items and not the non-promotable third item and return a message explaining that only 2 items are eligible for promotion. 

In my opinion, this would be better than expecting customers to do 2 separate orders which would increase friction and shipping costs. 

In addition, having order promotions apply to line item would prevent a customer from getting a huge discount on cheap items if they have an order with a significantly more expensive non promotable item. Imagine if you had an expensive non-promotable item with 2 relatively cheap promotable items and tried to apply the 20% discount at the order level. It would be theoretically possible for the customer to get the cheap items essentially for free.

PR #1929, opened 2017-05-22
Fix schema price (offer) by adding content attribute to price span

rsiddle:
This change makes the schemas pass the [Google Schema Markup validator](https://search.google.com/structured-data/testing-tool/u/0/).

Products were failing on the currency code being inclusive of the price when the `content` attribute is not present. The `content` attribute should contain float values.

Schema: http://schema.org/Offer

### DIFF
--- a/frontend/app/views/spree/products/_cart_form.html.erb
+++ b/frontend/app/views/spree/products/_cart_form.html.erb
@@ -34,7 +34,7 @@
         <div id="product-price">
           <h6 class="product-section-title"><%= Spree.t(:price) %></h6>
           <div>
-            <span class="price selling" itemprop="price">
+            <span class="price selling" itemprop="price" content="<%= @product.price_for(current_pricing_options).to_d %>">
               <%= display_price(@product) %>
             </span>
             <span itemprop="priceCurrency" content="<%= current_pricing_options.currency %>"></span>
--- a/frontend/app/views/spree/shared/_products.html.erb
+++ b/frontend/app/views/spree/shared/_products.html.erb
@@ -32,7 +32,10 @@
           </div>
           <%= link_to truncate(product.name, length: 50), url, class: 'info', itemprop: "name", title: product.name %>
           <span itemprop="offers" itemscope itemtype="http://schema.org/Offer">
-            <span class="price selling" itemprop="price"><%= display_price(product) %></span>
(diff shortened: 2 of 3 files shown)


frontend/spec/features/products_spec.rb
```diff
@@ -78,6 +78,16 @@
     end
   end
 
+  describe 'schema.org markup' do
+    let(:product) { Spree::Product.available.first }
+
+    it 'displays price without currency sign' do
+      expect(page).to have_css("#product_#{product.id} span.price[content='#{product.price}']")
+      click_link product.name
+      expect(page).to have_css("#product-price span.price[content='#{product.price}']")
```
mtomov:
```ruby
expect(page).to have_css("[itemprop=price][content='19.99']")
```

`priceCurrency`


> has content attributes

frontend/app/views/spree/products/_cart_form.html.erb
```diff
@@ -34,7 +34,7 @@
         <div id="product-price">
           <h6 class="product-section-title"><%= Spree.t(:price) %></h6>
           <div>
-            <span class="price selling" itemprop="price">
+            <span class="price selling" itemprop="price" content="<%= @product.price %>">
```
mamhoff: `@product.price` will return something different than `display_price(@product)` in many circumstances (customer uses a different currency than the admins, for example). This would need to recreate the behavious of `display_price`.

mamhoff: You can use `@product.price_for(current_pricing_options` here.

frontend/app/views/spree/shared/_products.html.erb
```diff
@@ -32,7 +32,10 @@
           </div>
           <%= link_to truncate(product.name, length: 50), url, class: 'info', itemprop: "name", title: product.name %>
           <span itemprop="offers" itemscope itemtype="http://schema.org/Offer">
-            <span class="price selling" itemprop="price"><%= display_price(product) %></span>
+            <span class="price selling" itemprop="price" content="<%= product.price %>">
```
mamhoff: `product.price_for(current_pricing_options).to_d`

rsiddle:
@mamhoff Updated to include correct price.

Issue #1811, opened 2017-03-31
User address not added to order after signing in


aaronrussell:
### Reproduce

Given:
* I am a registered user with an address
* I am **not logged in**
* I have some items in my cart and am on the cart page

When:
* I click to proceed with checkout
* I am redirected to the login page
* I fill in my login details and click submit
* I am redirected to the checkout address step
* And my address is **not displayed** in the form

Expected:
* I expect my existing address to populate the order

*The above behaviour only occurs when following the steps above. If I am already logged in, then go to my cart and proceed to checkout, my existing address is correctly populated as expected.*

### System

* solidus - 2.1.0
* solidus_auth_devise - 1.6.2
* rails - 5.0.0

mtomov:
I could tell you why that happens, though I don't have a solution proposal:

[#assign_default_addresses! in Checkout](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order/checkout.rb#L93) is the one doing the magic, but only when you change the state of the order from `cart` to `address`. That actually happens when you click Checkout, irrespective if you are logged in or not.. 

So, the state `move` happens already before you are being redirected to log in, so after you log in, and get to the address page, there is no `state` change to trigger this method to run again.

```sql

Started PATCH "/cart" for 127.0.0.1 at 2017-03-31 12:00:42 +0100
Processing by Spree::OrdersController#update as HTML
  Parameters: {"utf8"=>"✓", "authenticity_token"=>"T4PkR/ARxR+9K4yvWQO/oOV1O1qxxWx7/UoriZ2fc6A0+V/Ey23cmyUxszFBE8j2cAq3sZHZ7pNtvmurZ98vBQ==", "order"=>{"line_items_attributes"=>{"0"=>{"quantity"=>"1", "id"=>"3"}}, "coupon_code"=>""}, "checkout"=>""}

  .............

  SQL (0.2ms)  UPDATE "spree_orders" SET "state" = ?, "updated_at" = ? WHERE "spree_orders"."id" = ?  [["state", "address"], ["updated_at", 2017-03-31 11:00:42 UTC], ["id", 3]]

[...]

PR #1930, opened 2017-05-22
Add specs for Spree::Order#assign_default_addresses!

kennyadsl:
Starting working on #1811 I found out that [assign_default_addresses!](https://github.com/solidusio/solidus/blob/2a5b2221a9720a53f6eb89b5056f9059b28c597b/core/app/models/spree/order.rb#L700) method wasn't tested at all. I know specs are a bit too coupled with implementation but I think this is better than what we have now (no tests). If someone has ideas to decouple it I'm open to suggestions.

I also think that this method name can be improved. It's not clear that it is assigning the deafult user address and it seems more that it is adding a generic default address, which is actually done [here](https://github.com/solidusio/solidus/blob/2a5b2221a9720a53f6eb89b5056f9059b28c597b/frontend/app/controllers/spree/checkout_controller.rb#L176-L178). My proposal is renaming it in `assign_default_user_addresses!`, I could do that in this same PR maybe. What do you think?

### DIFF
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -774,6 +774,88 @@ describe Spree::Order, type: :model do
     end
   end
 
+  context "#assign_default_addresses!" do
+    let(:order) { Spree::Order.new }
+
+    subject { order.assign_default_addresses! }
+
+    context "when no user is associated to the order" do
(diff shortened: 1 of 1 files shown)


mamhoff:
Please do the renaming in a separate PR. This is great as it adds specs though!

Issue #1812, opened 2017-03-31
TieredPercent and TieredFlatRate calculators should have a currency preference


jhawthorn:
Both the TieredPercent and TieredFlatRate calculators have different options which activate when the amount is above a certain monetary value. Unfortunately, they aren't tagged with a currency, so they aren't suitable for use on multi-currency stores.

graygilmore:
I'll have a look at this one.

PR #1931, opened 2017-05-22
Currency Preference on Tiered Calculators

graygilmore:
Closes https://github.com/solidusio/solidus/issues/1812

Used the implementation example from: https://github.com/solidusio/solidus/blob/master/core/app/models/spree/calculator/flat_rate.rb#L9-L13

Open to other methods of implementing this if we've found something better.

## Admin
screen shot 2017-05-22 at 3 18 26 pm

### DIFF
--- a/core/app/models/spree/calculator/tiered_percent.rb
+++ b/core/app/models/spree/calculator/tiered_percent.rb
@@ -4,6 +4,7 @@ module Spree
   class Calculator::TieredPercent < Calculator
     preference :base_percent, :decimal, default: 0
     preference :tiers, :hash, default: {}
+    preference :currency, :string, default: -> { Spree::Config[:currency] }
 
     before_validation do
       # Convert tier values to decimals. Strings don't do us much good.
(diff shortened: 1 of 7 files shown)


core/app/models/spree/calculator/tiered_percent.rb
```diff
@@ -22,8 +23,16 @@ class Calculator::TieredPercent < Calculator
...
     def compute(object)
       order = object.is_a?(Order) ? object : object.order
-      _base, percent = preferred_tiers.sort.reverse.detect{ |b, _| order.item_total >= b }
-      (object.amount * (percent || preferred_base_percent) / 100).round(2)
+
+      _base, percent = preferred_tiers.sort.reverse.detect do |b, _|
+        order.item_total >= b
+      end
+
+      if preferred_currency.casecmp(order.currency).zero?
```
mamhoff: Would `preferred_currency.upcase == order.currency.upcase` be clearer than this?

graygilmore: I noticed that @jhawthorn had switched to casecmp from the `upcase` comparison so I followed suit: https://github.com/solidusio/solidus/commit/4d2b5ba513ce4269eef638f706e4a293e47b8fc7

mamhoff: Yeah, super!

tvdeyen:
Thanks 🎉

PR #1932, opened 2017-05-22
Destroy promotions on empty cart

seand7565:
Addresses #1919

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -475,6 +475,7 @@ module Spree
       line_items.destroy_all
       adjustments.destroy_all
       shipments.destroy_all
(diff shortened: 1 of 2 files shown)


core/app/models/spree/order.rb
```diff
@@ -475,6 +475,7 @@ def empty!
       line_items.destroy_all
       adjustments.destroy_all
       shipments.destroy_all
+      promotions.destroy_all
```
mamhoff: Can you try `order_promotions.destroy_all` instead? Even if this line only destroys the join table, we want to be clear that we're targetting the join table, not the promotions table.

seand7565: Done, tested & works great.

Issue #1913, opened 2017-05-19
Remove order-level adjustments from core Solidus


jordan-brough:
Order-level adjustments are not first-class citizens in Solidus.  Can we get rid of them?

Some of the problems:

- Taxes don't work correctly when they are present
- They have unexpected side effects for promotions.  E.g. by default Solidus chooses one promotion to apply per line item, but if you apply an order promotion than a single item can be discounted twice.
- They make analytics and attribution more difficult.  E.g. if you want to know how much revenue a particular variant has brought in, you'll need to figure out how to divvy up order-level promotion amounts amongst line items.
- They make partial refunds more difficult.  If you want to refund a customer for one item, you again need to divvy up order-level promotion amounts to figure out the amount to refund.
- If you have some items that are not discountable (like gift cards) the logic for determining how to divvy up order-level promotions has an additional complication

On the other hand, it does make conceptual sense for a store to say "Here's a promotion for $5 off a purchase of $30 or more", which sounds like an order-level promotion/adjustment.  But a promotion like that could be stored in the DB as either:
- One -$5 adjustment on the order (order-level adjustment)
- Multiple adjustments totaling $5 (the $5 split up amongst the line items)
The former feels more intuitive to me but the latter makes a lot of other things easier.

jordan-brough:
Here's a few first ideas on TODO items for accomplishing the above:

- [ ] Reimplement `Spree::Promotion::Actions::CreateAdjustment` to generate line-level adjustments
- [ ] Allow having Admin adjustments at the line level (see #1274)
- [ ] Figure out what to do about existing order-level adjustments.  Write a tool for migrating them?  Continue to support but deprecate them?  Move support into an extension?

bbuchalter:
I agree this is a noble goal, but is likely a big breaking change for many, many stores. Would need to be part of a major version bump and a lot of care given to a sane upgrade path. Nonetheless, 👍 .

jasonfb:
I wonder if there's a way to bridge the old-and-new paradigms: 

An "order level adjustment" that *looks* to the customer & business as-if it is applied at the "order" level, but then there is a calculation of how much was discounted per line item.

[...]

PR #1933, opened 2017-05-22
Calculating the adjustment total including null source adjustments

alexstoick:
This is to help achieve #1913 - enabling admins to do refunds at line
item level instead of at order level. There is no UI for this yet, but
the functionality exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
   This fixes issue [\#1836](https://github.com/solidusio/solidus/issues/1836). By allowing a TaxRate to tax multiple categories, stores don't have to create multiple TaxRates with the same value if a zone doesn't have different tax rates for some tax categories.
 
+- Adjustments without a source are now included in `line_item.adjustment_total`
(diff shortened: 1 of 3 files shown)


mamhoff:
I'm not sure I understand why line item level adjustments have to be source-less.

jordan-brough:
@mamhoff the only other type of adjustment that solidus core generates (besides promos & taxes) are "admin" adjustments - adjustments an admin makes for an arbitrary reason.  Right now these can only be added at the order level, and that blocks us from getting rid of order-level adjustments (thus this is linked to #1913).

A nice improvement (but probably separate PR) would be to support calling `update!` on adjustments that have a source other than PromotionAction/TaxRate/nil.

jordan-brough:
And another nice improvement beyond that might be to record something as the `source` for admin-generated adjustments.

mamhoff:
Yeah, sounds good, actually. In the meantime, someone else got to editing the change log, do you mind rebasing?

alexstoick:
@mamhoff rebased!

Issue #991, opened 2016-03-17
No method credit for PaymentMethod Check


ajkamel:
In testing some of the order functionality I ran into this error.  It seems like I was not able to properly reimburse a customer that had paid by check.

mamhoff:
Yep, that might be the case. Feel like cherry-picking this for Solidus and PRing it? https://github.com/spree/spree/pull/6529/files It will need a test or two ...

PR #1934, opened 2017-05-22
Add credit method to PaymentMethod::Check

vladstoick:
Fixes https://github.com/solidusio/solidus/issues/991. Changes from https://github.com/spree/spree/pull/6529/files

### DIFF
--- a/core/spec/models/spree/payment_method/check_spec.rb
+++ b/core/spec/models/spree/payment_method/check_spec.rb
@@ -1,0 +1,78 @@
+require 'spec_helper'
+
+describe Spree::PaymentMethod::Check do
+  let(:order) { create(:order) }
+  subject { described_class.new }
+
+  context "#can_capture?" do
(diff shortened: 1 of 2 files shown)


core/spec/models/spree/payment_method/check_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  context "#void" do
+    it "succeds" do
+      expect(Spree::PaymentMethod::Check.new.void).to be_success
+    end
+  end
+
+  context "#credit" do
+    it "succeds" do
+      expect(Spree::PaymentMethod::Check.new.credit).to be_success
```
mtomov: `Spree::PaymentMethod::Check` - can be referenced using `described_class`, and the `new` form of it is usually referenced using `subject`, so at the top you can do:

```ruby
subject { described_class.new }
# ...
expect(subject.credit).to ....
```

PR #1935, opened 2017-05-22
Remove Dependency on ActiveMerchant

Dkendal:
This decouples Solidus from ActiveMerchant.

This does _NOT_ fix all existing gateways, nor does it fix their dependence on
ActiveMerchant.

The main focus of this pull request is to make core not rely on
`ActiveMerchant::Billing::Response`, and `ActiveMerchant::ConnectionError`.

A number of classes are copied from ActiveMerchant, and are available as an
analogous class under the Spree namespace:
- `ActiveMerchant::Billing::Response`  -> `Spree::BillingResponse`
- `ActiveMerchant::Billing::CVVResult`  -> `Spree::CVVResult`
- `ActiveMerchant::Billing::AVSResult`  -> `Spree::AVSResult`
- `ActiveMerchant::ConnectionError`  -> `Spree::BillingConnectionError`

Existing usages of the active merchant classes will continue to work, but
report deprecation notices.

Next steps would be to have existing Gateways return a `Spree::BillingResponse`
instead of an ActiveMerchant equivalent.

### DIFF
--- a/backend/spec/features/admin/orders/log_entries_spec.rb
+++ b/backend/spec/features/admin/orders/log_entries_spec.rb
@@ -7,7 +7,7 @@ describe "Log entries", type: :feature do
 
   context "with a successful log entry" do
     before do
-      response = ActiveMerchant::Billing::Response.new(
+      response = Spree::BillingResponse.new(
         true,
         "Transaction successful",
         transid: "ABCD1234"
@@ -31,7 +31,7 @@ describe "Log entries", type: :feature do
 
   context "with a failed log entry" do
     before do
-      response = ActiveMerchant::Billing::Response.new(
+      response = Spree::BillingResponse.new(
         false,
         "Transaction failed",
         transid: "ABCD1234"
--- a/core/app/models/spree/avs_result.rb
+++ b/core/app/models/spree/avs_result.rb
@@ -40,24 +40,22 @@ module Spree
 
     # Map vendor's AVS result code to a postal match code
     POSTAL_MATCH_CODE = {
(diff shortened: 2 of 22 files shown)


Dkendal:
@mamhoff rebased

Dkendal:
@jordan-brough While I agree that a simplified interface for gateways would be an excellent move forward, this is only intended to remove the dependency while still providing duck typing and backward compatibility with all existing gateways.

jordan-brough:
If we _can_ get to a simpler interface it feels like we should lead with that.

This feels very good as an exploratory step to see how many places ActiveMerchant is hooked in, but for this to provide any real benefit other libraries would need to start creating `Spree::BillingResponse` objects, and I think we'd want them to be able to create simpler objects.

jordan-brough:
@Dkendal we chatted about this in the core team meeting yesterday and we'd like to shoot for some simpler models rather than pulling in the ActiveMerchant ones verbatim.  There's a lot of code in there that we don't want to maintain or expose as an official `Spree` object.

That said, the work you did here is very helpful in identifying what would need to change to introduce a simplified billing response class.

From looking through your work here it appears that we could create a minimal `Spree::BillingResponse` object along with a translation layer that converts an `ActiveMerchant::Billing::Response` into a `Spree::BillingResponse`.  (Instead of having a `Spree::BillingResponse` with the exact same interface as the `ActiveMerchant::Billing::Response` class.)  If that's something you're interested in doing let me know, otherwise I might see if I have time to give that a try, based on your work here.

PR #1941, opened 2017-05-23
Update wallet payment source method names.

skukx:
Being inside the wallet, it should be expected that all methods adhere
to the wallet class. Therefore, having method names such as below should
be considered redundant.

```ruby
wallet.wallet_payment_sources
wallet.default_wallet_payment_source
```

This would read much better as
```ruby
wallet.payment_sources
wallet.default_payment_source
```

Being inside the wallet class should be enough to know that the payment
sources you are obtaining are the wallet payment sources.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,8 +129,8 @@
 
 - Deprecate `Order#has_step?` in favour of `has_checkout_step?` [#1667](https://github.com/solidusio/solidus/pull/1667) ([mamhoff](https://github.com/mamhoff))
 - Deprecate `Order#set_shipments_cost`, which is now done in `Order#update!` [\#1689](https://github.com/solidusio/solidus/pull/1689) ([jhawthorn](https://github.com/jhawthorn))
-- Deprecate `user.default_credit_card`, `user.payment_sources` for `user.wallet.default_wallet_payment_source` and `user.wallet.wallet_payment_sources`
-- Deprecate `CreditCard#default` in favour of `user.wallet.default_wallet_payment_source`
+- Deprecate `user.default_credit_card`, `user.payment_sources` for `user.wallet.default_payment_source` and `user.wallet.payment_sources`
+- Deprecate `CreditCard#default` in favour of `user.wallet.default_payment_source`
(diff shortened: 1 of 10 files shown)


jordan-brough:
@skukx we actually had those names originally and changed them to the longer ones.

There is both a `PaymentSource` and a `WalletPaymentSource` class, so `wallet.payment_sources` seemed a bit confusing since you're getting back `WalletPaymentSource`s, not `PaymentSource`s.

skukx:
@jordan-brough I have to disagree with the confusion.  Since we are within the context of the wallet, it should be expected that the payment sources being returned are of the `Spree::WalletPaymentSource` class

skukx:
Awwww I see what you're saying.  I was under the assumption that it was these methods were returning `Spree::PaymentSource` and that `Spree::WalletPaymentSource` was a join table to those classes.

jordan-brough:
@skukx yeah, I think ideally maybe we could've come up with a better name for `WalletPaymentSource` but that might be hard to change now.  It is a join table between the user and their payment sources, but the naming makes this a bit confusing perhaps.

skukx:
What is the reasoning for why the `Spree::WalletPaymentSource` (Join table) is returned rather than just the `Spree::PaymentSource`  In most cases won't you be calling `wallet.default_wallet_payment_source.payment_source`?

mamhoff:
@skukx Largely, we need to expose those `WalletPaymentSources` because frontend returns singular IDs, whereas payment sources are polymorphic and could be anything from a credit card to a roman debt token. 

Yes, in most cases we want the `payment_source`, but we opted for the not-so-pretty-looking but more correct `wallet.default_wallet_payment_source.payment_source`. That API allows us to do all the things we need while not exposing any methods called one thing and returning another.

skukx:
Ok, I finally understand what is happening. I mistook `Spree::PaymentSource` to be `Spree::Payment` and thought that `Spree::WalletPaymentSource` was just a join table to `Spree::Payment`. So calling `wallet.default_wallet_payment_source.payment_source` returned a `Spree::Payment` which had the polymorphic association.  However `Spree::WalletPaymentSource` is the actual polymorphic table (Not the join table).  I fully support this implementation understanding this now and closing the PR.

Issue #438, opened 2015-10-13
Remove admin ability to set a password, replace reset password email trigger.


athal7:
In the spirit of better security. relates to https://github.com/solidusio/solidus/pull/421

mamhoff:
@acreilly is working on this

seand7565:
An admin would still have the ability to change the email associated with the account and then trigger the email reset - essentially giving them the ability to change the password. Should the ability to change a users email also be disabled? I can't think of too many use cases in which you'd need to change a users email. @acreilly

ericsaupe:
The only time I could think that an admin would _need_ to be able to change the email address is if the user no longer has access to that email and would need their password changed. New account is the path there maybe?

acreilly:
@ericsaupe Lets say the data associated with that account is important to the user though, I think keeping the ability to change the email would be helpful. It isn't a common scenario, so if we really want to take it out, I guess a developer can go into the console and change it. What do you think?

ericsaupe:
Yeah, that's what I was thinking too. For our customer service reps I know they would be really angry if we simply told them to tell users to create a new account and lose all of their order history, saved cards, etc. Probably best to leave the ability to change emails in and let individual stores change that if they need to.

acreilly:
It doesn't appear that we use this view: https://github.com/solidusio/solidus_auth_devise/blob/master/lib/views/backend/spree/admin/user_passwords/edit.html.erb

Does having the reset_password_token increase security? Or do we just want to stick with adding the Send Reset Password Email button

ericsaupe:
It looks like it is required in the Devise method, https://github.com/plataformatec/devise/blob/master/app/controllers/devise/passwords_controller.rb#L32. I think it is simply for resetting the password through the controller but the issue here is to remove the change password form all together to simply send the reset email. So if it's not used it could probably be removed all together.

acreilly:
@ericsaupe I'm not sure if you are here today... but if you are would you mind helping me for a second? I am pretty easy to spot out in the room hah..

ericsaupe:
Of course! I'm next to that @skukx purple faced guy from that other PR. I'll come find you in a second.

mamhoff:
Why don't you let customer support change the email address and then send out a password reset token?

ericsaupe:
@mamhoff good idea. This task seems to be to add that ability to send the password reset using the admin.

tvdeyen:
[...]

PR #97, opened 2015-06-19
Be more strict about when inventory units and shipments are allowed to be destroyed.

athal7:
originally from https://github.com/bonobos/spree/pull/406

PR #1942, opened 2017-05-23
Remove password fields in Admin/Users view. Add a send reset password…

acreilly:
… email button.

#438 

This PR needs to be merged with [solidus_auth_devise PR #97](https://github.com/solidusio/solidus_auth_devise/pull/97)

### DIFF
--- a/backend/app/views/spree/admin/users/_form.html.erb
+++ b/backend/app/views/spree/admin/users/_form.html.erb
@@ -53,21 +53,4 @@
     <% end %>
 
   </div>
-
-  <div data-hook="admin_user_form_password_fields" class="col-6">
-    <%= f.field_container :password do %>
-      <%= f.label :password %>
-      <%= f.password_field :password, class: 'fullwidth' %>
-      <%= f.error_message_on :password %>
-    <% end %>
-
-
-    <% if f.object.respond_to?(:password_confirmation) %>
-      <%= f.field_container :password do %>
-        <%= f.label :password_confirmation %>
-        <%= f.password_field :password_confirmation, class: 'fullwidth' %>
-        <%= f.error_message_on :password_confirmation %>
-      <% end %>
-    <% end %>
-  </div>
 </div>
--- a/backend/app/views/spree/admin/users/edit.html.erb
+++ b/backend/app/views/spree/admin/users/edit.html.erb
@@ -23,7 +23,20 @@
         </div>
       <% end %>
     <% end %>
+
   </div>
+</fieldset>
+
+<fieldset id='forgot-password'>
+  <legend><%= Spree.t(:forgot_password) %></legend>
+
+  <%= form_for [@admin, Spree.user_class.new], as: :spree_user, url: spree.admin_reset_password_path do |f| %>
+      <%= f.hidden_field :email, value: @user.email %>
+
+    <div class="filter-actions actions">
+      <%= f.submit Spree.t(:reset_user_password), class: 'button primary' %>
+    </div>
+  <% end %>
 </fieldset>
 
 <% if can?(:update, @user) %>
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1794,6 +1794,7 @@ en:
     reports: Reports
     resend: Resend
     reset_password: Reset my password
+    reset_user_password: Reset user password
     response_code: Response Code
     restock_inventory: Restock Inventory
     resume: Resume


gmacdougall:
I'm concerned about the tight coupling between this change and the required change in solidus_auth_devise. We like to maintain compatibility between extension versions.

Should we consider a strategy where we detect what solidus_auth_devise supports and then provide the correct functionality. Perhaps with a deprecation warning as well.

There are also failing tests which need to be addressed.

acreilly:
I fixed the initial failing specs, but the specs will continue to fail due to the changes in solidus_auth_devise not being available on master.

I'm not exactly sure how to go about what you suggested. Since this change here would only be on master, it shouldn't affect other solidus versions. The change that was made on solidus_auth_devise won't break anything for older versions since that route is not used anywhere and is technically broken if anyone tried to use it.

BenMorganIO:
Personally, I think there should be both. If we go for the password reset, an admin won't be able to manually change another users password.

Often, when onboarding, an admin will add an employee via the admin, give them a role, and then set their password to something and ask the new employee to update their password for security. If we do both, we allow an admin to be able to create a new user and set their password and then send a reset password email.

Both cases are beneficial and I see a use for both of them. If you're an admin or the head honcho, you have the OSS and you're running it with your own database. Nothing is actually stopping an admin from going into the console and setting another users password. The least that we could do is make it easier for them via having a web interface for it. The only security issue here is if they have a poor password or left their session open.

Therefore, we could just ask the current user to provider _their_ password as confirmation and also allow them to send out reset password emails. We should probably also wrap this in a `can?`.

backend/app/views/spree/admin/users/_form.html.erb
```diff
@@ -53,21 +53,4 @@
...
-
-
-    <% if f.object.respond_to?(:password_confirmation) %>
-      <%= f.field_container :password do %>
-        <%= f.label :password_confirmation %>
-        <%= f.password_field :password_confirmation, class: 'fullwidth' %>
-        <%= f.error_message_on :password_confirmation %>
-      <% end %>
-    <% end %>
-  </div>
```
tvdeyen: I think we should keep these fields and only display them

```rb
if @user == try_spree_current_user
```

also, we should add a "current password" field. Then we should have addressed all security concerns.

backend/app/views/spree/admin/users/edit.html.erb
```diff
@@ -23,9 +23,22 @@
         </div>
       <% end %>
     <% end %>
+
   </div>
 </fieldset>
 
+<fieldset id='forgot-password'>
```
tvdeyen: We should add a `no-border-bottom` class. There are already too much borders on this page, also I don't not see any need for adding a DOM ID.

backend/app/views/spree/admin/users/edit.html.erb
```diff
@@ -23,9 +23,22 @@
...
+  <legend><%= Spree.t(:forgot_password) %></legend>
+
+  <%= form_for [@admin, Spree.user_class.new], as: :spree_user, url: spree.admin_reset_password_path do |f| %>
+      <%= f.hidden_field :email, value: @user.email %>
+
+    <div class="filter-actions actions">
```
tvdeyen: We should just add a `align-center` class and should not make use of the `filter-actions actions` classes, as these will add even more borders to this already overwhelming page.

backend/app/views/spree/admin/users/edit.html.erb
tvdeyen: We should hide this conditionally 

```rb
if @user != try_spree_current_user
```

That this is only visible for admins that don't manage their own password.

core/config/locales/en.yml
```diff
@@ -1731,6 +1731,7 @@ en:
     reports: Reports
     resend: Resend
     reset_password: Reset my password
+    reset_user_password: Reset user password
```
tvdeyen: By using the human attribute name, like suggested above, we can remove this.

tvdeyen:
@acreilly there is a rebase necessary. Would you mind to?

tvdeyen:
> Since this change here would only be on master, it shouldn't affect other solidus versions. The change that was made on solidus_auth_devise won't break anything for older versions since that route is not used anywhere and is technically broken if anyone tried to use it.

@gmacdougall @acreilly is right about this. Do you have anything particular in mind why this or the solidus auth PR should not be able to be merged?

seand7565:
Hi all! This change looks great but it looks like it's been needing work for a while. This looks like a good fix to our current oldest issue, #438 - are there any blockers in making the above changes (other than the version changes that have happened since this PR?)

backend/app/views/spree/admin/users/edit.html.erb
BenMorganIO: Why? That breaks consistency with the UI.

kennyadsl:
@acreilly hey there! I'm taking a look at this PR. There are some changes requested by @tvdeyen still to be done, especially the one about keeping fields if you are that user, since at the moment the UX generated with this PR is not ideal:

- as an Admin I need to change my password
- I visit my user page in admin
- I send the reset password link to myself
- I click on the link in the email
- I'm redirected to the homepage without any message and I can't change my password

This happens because to perform the reset it's required not to be authenticated:

```text
Redirected to http://localhost:3000/
Filter chain halted as :require_no_authentication rendered or redirected
Completed 302 Found in 6ms (ActiveRecord: 0.6ms)
```

Having the old password fields there for current users could be a solution for this problem.

Also, I think this needs a rebase. If you want or do not have time I can continue this PR with what's missing.

zabrador:
@kennyadsl this seems pretty stale, and I believe it was superseded by https://github.com/solidusio/solidus_auth_devise/pull/146. Can this now be closed?

kennyadsl:
Yes, you are right, thanks!

Issue #1131, opened 2016-05-10
Incompatible LineItem option permitting between Api::OrdersController and Api::LineItemsController


adaddeo:
Unable to create a new order with a line item with options. This is the issue/motivation behind #1112.

To reproduce:

Start by permitting a new option on to permitted on a line item:

```
PermittedAttributes.line_item_attributes << :some_option
```

Now each of the requests should create the line item with some_option set to 4:

To OrdersController#create

```
{ order: { line_items: [variant_id: 1, quantity:1, options: { some_option: 4 } ] } }
```

To LineItemsController#create

```
{ line_item: variant_id: 1, quantity:1, options: { some_option: 4 } }
```

However, the OrdersController fails to permit the option.

gus4no:
Working on it 👷

adaddeo:
This was fixed by @gus4no

PR #1943, opened 2017-05-23
Change how line item options are allowed in line items controller

gus4no:
#1131 

Don't really know if this is the right way.. but old logic and issue reported seemed confusing to me, apparently it was just a bad setup of permitted attributes.

I just had the line items controller changed to make more sense, and got rid of that class attribute.

Permitted controllers are working now, old test are still green and I added a new test to confirm that orders are being created properly with line items with options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Solidus 2.3.0 (master, unreleased)
 
+- Change how line item options are allowed in line items controller. [\#1943](https://github.com/solidusio/solidus/pull/1943)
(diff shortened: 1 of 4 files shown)


mamhoff:
This looks good to me, but needs a changelog entry for people who have overridden or used that class attribute. Also, would you mind squashing the Hound appeasement into the previous commit?

gus4no:
@mamhoff Changelog updated and commits squashed! :)

adaddeo:
Thanks, looking forward to trying this out!

PR #1944, opened 2017-05-23
Allow admin to edit variant option values

dividedharmony:
This PR reverts ce10040a which, on the edit admin product variants form, hid the ability to edit an existent variant's option value from the admin. However, we want maximize the flexibility for admins to edit variants. Thus, we should display this fieldset to the admin.

### DIFF
--- a/backend/app/views/spree/admin/variants/_form.html.erb
+++ b/backend/app/views/spree/admin/variants/_form.html.erb
@@ -18,20 +18,24 @@
     </div>
   </fieldset>
   <fieldset class="no-border-top no-border-bottom">
-    <% @product.option_types.each_with_index do |option_type, index| %>
-      <div class="field four columns <%= 'alpha' if index % 3 == 0 %><%= 'omega' if index % 3 == 2 %>" data-hook="presentation">
-        <%= label :new_variant, option_type.presentation %>
-        <%= f.collection_select 'option_value_ids',
-                                option_type.option_values,
-                                :id,
-                                :presentation,
-                                { include_blank: true },
-                                {
-                                  name: 'variant[option_value_ids][]',
-                                  class: "select2 fullwidth"
-                                } %>
-      </div>
-    <% end %>
+    <div class="row">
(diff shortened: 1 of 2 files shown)


jordan-brough:
@dividedharmony this doesn't seem to address some of the concerns that @mamhoff raised in [his commit](https://github.com/solidusio/solidus/pull/1019/commits/ce10040ae25f80f58f616c0bc1e402a0ea9cd694).

He mentioned that with the form, an admin:
> is able to change the variant's options to those of a variant that already
exists

Is that not still a danger?

dividedharmony:
@jordan-brough Good point. I can add a validation such that OptionValuesVariants are unique within the scope of any given product.

mamhoff:
@jordan-brough yes that could still happen, but I've since been convinced that admins should have that power. For example, stores that model packaging via variant options might want to change how a variant is packaged (in a bag or a carton) without having to re-create the variant and possibly losing the connection to the previous packaging for analytics.

backend/app/views/spree/admin/variants/_form.html.erb
```diff
@@ -17,28 +17,22 @@
...
-                                      {
-                                        name: 'variant[option_value_ids][]',
-                                        class: "select2 fullwidth"
-                                      } %>
-            </div>
-          </div>
-      <% end %>
+  <fieldset class="no-border-top no-border-bottom">
+    <% @product.option_types.each_with_index do |option_type, index| %>
+      <div class="field four columns <%= 'alpha' if index % 3 == 0 %><%= 'omega' if index % 3 == 2 %>" data-hook="presentation">
```
jhawthorn: This goes back to using the skeleton classes. Can this be updated to use the bootstrap classes (as the removed code did)

jhawthorn: This goes back to using the skeleton classes. Can this be updated to use the bootstrap classes (as the removed code did)

dividedharmony:
@jhawthorn Thanks for pointing that out! I've switched back over to the Bootstrap way of doing things.

PR #1948, opened 2017-05-23
Fix BigDecimal error in LocalizedNumber in ruby 2.4

omnistegan:
In ruby 2.4, calling `to_d` on an empty string raises an ArgumentError. Before it would return `0.0`. This commit rescues from the error and returns zero.

See also:
https://github.com/solidusio/solidus/pull/1872

### DIFF
--- a/core/lib/spree/localized_number.rb
+++ b/core/lib/spree/localized_number.rb
@@ -21,7 +21,8 @@ module Spree
       # then replace the locale-specific decimal separator with the standard separator if necessary
       number = number.gsub(separator, '.') unless separator == '.'
 
-      number.to_d
+      # Handle empty string for ruby 2.4 compatibility
+      BigDecimal.new(number.presence || 0)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


core/lib/spree/localized_number.rb
```diff
@@ -21,7 +21,14 @@ def self.parse(number)
       # then replace the locale-specific decimal separator with the standard separator if necessary
       number = number.gsub(separator, '.') unless separator == '.'
 
-      number.to_d
+      cast_to_d(number)
     end
+
+    def self.cast_to_d(value)
+      value.to_s.to_d
```
tvdeyen:
```ruby
BigDecimal.new(value.presence || 0)
```

should work without the need of `rescue`

omnistegan: I like this better. Also we can use it when we return the decimal instead of using a `cast_to_d` method. How would you feel about:

```ruby
    def self.parse
      ...

      # Handle empty string for ruby 2.4 compatibility
      BigDecimal.new(number.presence || 0)
    end
```

tvdeyen: Absolutely 👍

PR #1949, opened 2017-05-23
Distributed Amount Calculator

graygilmore:
This allows admins to create order level promotions that get applied through line item adjustments. For example "Save $50 off your order" and we spread the $50 through line item adjustments.

## Example

If you had a **$15** promotion that you wanted to spread over three line items with differing prices:

| Line Item | Line Item Price | Weighted Adjustment |
| -- | -- | -- |
| Socks | $5 | -$1.5 |
| Shoes | $30 | -$9 |
| Slippers | $15 | -$4.5 |

## Notes

This will help move us in the direction of removing order-level adjustments as discussed here: https://github.com/solidusio/solidus/issues/1913

This may even allow us to deprecate a flat rate order-level adjustment if we wanted to move in that direction.

## Admin Screenshot

screen shot 2017-05-23 at 2 14 37 pm

### DIFF
--- a/backend/app/views/spree/admin/promotions/actions/_create_item_adjustments.html.erb
+++ b/backend/app/views/spree/admin/promotions/actions/_create_item_adjustments.html.erb
@@ -1,2 +1,6 @@
-<%= render partial: 'spree/admin/promotions/actions/promotion_calculators_with_custom_fields',
-           locals: { calculators: Spree::Promotion::Actions::CreateItemAdjustments.calculators, promotion_action: promotion_action, param_prefix: param_prefix } %>
+<%= render(
+  "spree/admin/promotions/actions/promotion_calculators_with_custom_fields",
+  calculators: Spree::Promotion::Actions::CreateItemAdjustments.calculators,
+  promotion_action: promotion_action,
+  param_prefix: param_prefix
+) %>
--- a/backend/app/views/spree/admin/promotions/calculators/weighted_rate/_fields.html.erb
+++ b/backend/app/views/spree/admin/promotions/calculators/weighted_rate/_fields.html.erb
@@ -1,0 +1,54 @@
+<%= fields_for "#{prefix}[calculator_attributes]", calculator do |f| %>
+  <%= f.label :preferred_amount %>
+  <%= render "spree/admin/shared/number_with_currency", f: f, amount_attr: :preferred_amount, currency_attr: :preferred_currency %>
+<% end %>
+
+<div class="field">
+  <p>
+    <%= admin_hint(
+      calculator.model_name.human,
+      """
+        <p>
+          This amount will distributed to line items weighted relative to their
+          price. More expensive line items will receive a greater share of the
+          adjustment.
+        </p>
+
+        <p>
+          For example, with three line items and a preferred amount of $15 we
+          would end up with the following distribution:
+        </p>
+
+        <table>
+          <thead>
+            <tr>
+              <th></th>
+              <th>Price</th>
+              <th>Weighted Adj.</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr>
+              <td>Socks</td>
+              <td>$5</td>
+              <td>-$1.5</td>
+            </tr>
+            <tr>
+              <td>Shoes</td>
+              <td>$30</td>
+              <td>-$9</td>
+            </tr>
+            <tr>
+              <td>Slippers</td>
+              <td>$15</td>
+              <td>-$4.5</td>
+            </tr>
+          </tbody>
+        </table>
+      """
+    ) %>
+
+    This amount will be the <strong>total</strong> discount spread amongst all
+    of the line items.
+  </p>
+</div>
--- a/backend/app/views/spree/admin/promotions/calculators/weighted_rate/_fields.html.erb
+++ b/backend/app/views/spree/admin/promotions/calculators/weighted_rate/_fields.html.erb
@@ -9,9 +9,9 @@
       calculator.model_name.human,
       """
         <p>
-          This amount will distributed to line items weighted relative to their
-          price. More expensive line items will receive a greater share of the
-          adjustment.
+          This amount will be distributed to line items weighted relative to
+          their price. More expensive line items will receive a greater share
+          of the adjustment.
         </p>
 
         <p>
--- a/core/app/models/spree/calculator/distributed_amount.rb
+++ b/core/app/models/spree/calculator/distributed_amount.rb
@@ -1,0 +1,24 @@
+require_dependency 'spree/calculator'
+
+# This is a calculator for line item adjustment actions. It accepts a line item
+# and calculates its weighted adjustment amount based on the value of the
+# preferred amount and the price of the other line items. More expensive line
+# items will receive a greater share of the preferred amount.
+
+module Spree
+  class Calculator::DistributedAmount < Calculator
+    preference :amount, :decimal, default: 0
+    preference :currency, :string, default: -> { Spree::Config[:currency] }
+
+    def compute_line_item(line_item)
+      if line_item && preferred_currency.casecmp(line_item.currency).zero?
+        Spree::DistributedAmountsHandler.new(
+          line_item,
+          preferred_amount
+        ).amount
+      else
+        0
+      end
+    end
+  end
+end
--- a/core/app/models/spree/distributed_amounts_handler.rb
+++ b/core/app/models/spree/distributed_amounts_handler.rb
@@ -1,0 +1,43 @@
+module Spree
+  class DistributedAmountsHandler
+    attr_reader :line_item, :order, :total_amount
+
+    def initialize(line_item, total_amount)
+      @line_item = line_item
+      @order = line_item.order
+      @total_amount = total_amount
+    end
+
+    # @return [Float] the weighted adjustment for the initialized line item
+    def amount
+      distributed_amounts[@line_item.id].to_f
+    end
+
+    private
+
+    # @private
+    # @return [Hash<Integer, BigDecimal>] a hash of line item IDs and their
+    #   corresponding weighted adjustments
+    def distributed_amounts
+      remaining_amount = @total_amount
+
+      @order.line_items.each_with_index.map do |line_item, i|
+        if i == @order.line_items.length - 1
+          # If this is the last line item on the order we want to use the
+          # remaining preferred amount to ensure our total adjustment is what
+          # has been set as the preferred amount.
+          [line_item.id, remaining_amount]
+        else
+          # Calculate the weighted amount by getting this line item's share of
+          # the order's total and multiplying it with the preferred amount.
+          weighted_amount = ((line_item.amount / @order.item_total) * total_amount).round(2)
+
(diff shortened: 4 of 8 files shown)


core/app/models/spree/calculator/weighted_rate.rb
```diff
@@ -0,0 +1,19 @@
+require_dependency 'spree/calculator'
+
+module Spree
+  class Calculator::WeightedRate < Calculator
```
graygilmore: Not completed sold on the naming of this calculator.

core/app/models/spree/weighted_adjustments_handler.rb
```diff
@@ -0,0 +1,47 @@
+module Spree
+  class WeightedAdjustmentsHandler
```
graygilmore: Would happily rename this class to something else. Additionally, only the calculator is currently under test. I intend to switch to placing this under test and just making sure that the calculator calls this with the correct arguments.

backend/app/views/spree/admin/promotions/actions/_create_item_adjustments.html.erb
```diff
@@ -1,2 +1,5 @@
-<%= render partial: 'spree/admin/promotions/actions/promotion_calculators_with_custom_fields',
-           locals: { calculators: Spree::Promotion::Actions::CreateItemAdjustments.calculators, promotion_action: promotion_action, param_prefix: param_prefix } %>
\ No newline at end of file
+<%= render(
+  "spree/admin/promotions/actions/promotion_calculators_with_custom_fields",
+  calculators: Spree::Promotion::Actions::CreateItemAdjustments.calculators, promotion_action: promotion_action,
```
jordan-brough: did you intend to give a newline for the `promotion_action` option?

core/app/models/spree/weighted_adjustments_handler.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def initialize(line_item, preferred_amount)
+      @line_item = line_item
+      @order = line_item.order
+      @preferred_amount = preferred_amount
+    end
+
+    # Returns the weighted adjustment for the initialized line item
+    #
+    # @return [Integer]
```
graygilmore: Should probably be `[Float]`.

core/app/models/spree/calculator/weighted_rate.rb
```diff
@@ -0,0 +1,19 @@
...
+    preference :amount, :decimal, default: 0
+    preference :currency, :string, default: -> { Spree::Config[:currency] }
+
+    def compute_line_item(line_item)
+      if line_item && preferred_currency.casecmp(line_item.currency).zero?
```
jordan-brough: I'm not sure this is worth fixing, but we check the currency here at the line level but we don't check it again in WeightedAdjustmentsHandler.  That's fine in real life since we know that behind the scenes the line item delegates the currency to the order, but it seems wrong somehow to have to know that?

core/app/models/spree/calculator/weighted_rate.rb
jordan-brough: It would be great to have some comments explaining this calculator

core/app/models/spree/weighted_adjustments_handler.rb
```diff
@@ -0,0 +1,47 @@
+module Spree
+  class WeightedAdjustmentsHandler
+    attr_reader :line_item, :order, :preferred_amount
+
+    def initialize(line_item, preferred_amount)
```
jordan-brough: Maybe `total_amount` instead of `preferred_amount`?

core/app/models/spree/calculator/weighted_rate.rb
mamhoff: Maybe `DistributedAmounts`?

backend/app/views/spree/admin/promotions/calculators/weighted_rate/_fields.html.erb
```diff
@@ -0,0 +1,54 @@
+<%= fields_for "#{prefix}[calculator_attributes]", calculator do |f| %>
+  <%= f.label :preferred_amount %>
+  <%= render "spree/admin/shared/number_with_currency", f: f, amount_attr: :preferred_amount, currency_attr: :preferred_currency %>
+<% end %>
+
+<div class="field">
+  <p>
+    <%= admin_hint(
+      calculator.model_name.human,
+      """
```
mamhoff: Really like the use of admin hints here. :+1:

core/app/models/spree/weighted_adjustments_handler.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    # Returns a hash of line item IDs and their corresponding weighted
+    # adjustments
+    #
+    # @private
+    # @return [Hash<Integer, Integer>]
+    def weighted_amounts
+      remaining_amount = @preferred_amount
+
+      @order.line_items.each_with_index.map do |line_item, i|
```
jordan-brough: Not super important but could we add a `sort_by(&:id)` to this? Just to avoid the "rounding penny' from ending up on different items in different runs?

core/app/models/spree/weighted_adjustments_handler.rb
```diff
@@ -0,0 +1,47 @@
...
+        if i == @order.line_items.count - 1
```
jordan-brough: I think we want to use `line_items.length` here to force the association to load if it hasn't already, and avoid the db-query that `count` will always generate

core/app/models/spree/weighted_adjustments_handler.rb
```diff
@@ -0,0 +1,47 @@
...
+          # If this is the last line item on the order we want to use the
+          # remaining preferred amount to ensure our total adjustment is what
+          # has been set as the preferred amount.
+          [line_item.id, remaining_amount]
+        else
+          # Calculate the weighted amount by getting this line item's share of
+          # the order's total and multiplying it with the preferred amount.
+          weighted_amount = ((line_item.amount / @order.item_total) * preferred_amount).round(2)
```
jordan-brough: I think we have this problem all over the place, but we should round according to the number of fractional digits of the currency, not `2`.

I think it's fine if we leave it as `round(2)` in this PR (unless you want to have a go at fixing it) and I'll open a separate PR for the more general issue.

core/app/models/spree/calculator/weighted_rate.rb
jordan-brough: `WeightedAmounts`?

jordan-brough: Or maybe `DistributedAmount` (singular instead of plural, since the amount specified to the calculator is a single amount?)

graygilmore:
Great comments. Going to look at these now.

core/app/models/spree/weighted_adjustments_handler.rb
jordan-brough: agreed 👍

core/app/models/spree/weighted_adjustments_handler.rb
jordan-brough: Actually that would make the big-o worse for something that's already a bit worrisome.  Maybe that's not worth it.

core/app/models/spree/weighted_adjustments_handler.rb
jordan-brough: Something similar to whatever you choose for the calculator name seems fine imo.  Should this go in the `models/spree/promotion/` folder somewhere though?

jordan-brough:
You should add a screenshot of what the admin UI looks like for this calculator :) Best one I've seen I think.

core/app/models/spree/weighted_adjustments_handler.rb
graygilmore: I'm happy to leave this for now. Would prefer to avoid tacking too many things onto this PR.

core/app/models/spree/calculator/weighted_rate.rb
graygilmore: I liked having it in the calculator. I'm not sure it needs to exist in both places as it's the calculator that cares about this the most.

jordan-brough: Yah my concern was that we're only checking _one_ of the line item currencies. When you see `line_item.currency` that doesn't guarantee that all the line items are the same currency right?  (They _are_ but the interface doesn't make that clear.)

Personally i think it'd be more intuitive to see us check `line_item.order.currency` (or `line_item.order_currency`).

But like I said, probably not worth worrying about too much.

core/app/models/spree/weighted_adjustments_handler.rb
jordan-brough: 👍 FYI the github issue: https://github.com/solidusio/solidus/issues/1957

core/app/models/spree/calculator/weighted_rate.rb
graygilmore: Aahhh I gotcha. Right now `line_item.currency` actually delegates to the order already: https://github.com/solidusio/solidus/blob/master/core/app/models/spree/line_item.rb#L41

backend/app/views/spree/admin/promotions/calculators/weighted_rate/_fields.html.erb
```diff
@@ -0,0 +1,54 @@
...
+        <p>
+          This amount will distributed to line items weighted relative to their
```
isaacfreeman: Typo: "This amount will *be* distributed"

isaacfreeman: Possibly clearer first sentence:
"This amount will be distributed between line items proportionally to their price."

backend/app/views/spree/admin/promotions/calculators/weighted_rate/_fields.html.erb
```diff
@@ -0,0 +1,54 @@
...
+              <td>Slippers</td>
+              <td>$15</td>
+              <td>-$4.5</td>
+            </tr>
+          </tbody>
+        </table>
+      """
+    ) %>
+
+    This amount will be the <strong>total</strong> discount spread amongst all
```
isaacfreeman: I think this sentence is redundant, but if we want to emphasise that this amount is the total, it should appear before the example.

jordan-brough: @isaacfreeman the preceding text is the tooltip. This is the text that always shows. Please see the screenshot in the PR text.

backend/app/views/spree/admin/promotions/calculators/weighted_rate/_fields.html.erb
jordan-brough: Thanks @isaacfreeman. I went ahead and added a commit to fix the typo.

jordan-brough: I'm going to go ahead and merge this but would be happy to see a PR to improve it from either you or @graygilmore

PR #1951, opened 2017-05-23
Allow flexible separator for promotion code

ericgross:
Rather than always seperating the promotion base_code from the suffix
with an underscore (_), allow a seperator to be specified.

### DIFF
--- a/core/spec/models/spree/promotion_code/batch_builder_spec.rb
+++ b/core/spec/models/spree/promotion_code/batch_builder_spec.rb
@@ -12,7 +12,7 @@ describe Spree::PromotionCode::BatchBuilder do
     )
   end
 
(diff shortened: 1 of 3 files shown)


core/spec/models/spree/promotion_code/batch_builder_spec.rb
```diff
@@ -27,11 +27,12 @@
         expect(promotion_code_batch.reload.state).to eq("failed")
       end
     end
+
     context "with a successful build" do
       before do
         allow(Spree::PromotionCodeBatchMailer)
           .to receive(:promotion_code_batch_finished)
-          .and_call_original
+                .and_call_original
```
jordan-brough: Accidental change here?

Issue #610, opened 2015-12-20
Helper method conflict font-awesome-sass gem


harin:
Both solidus and font-awesome-sass uses the `icon` method. There should be a way to prevent helper method conflict. One way is to set `include_all_helpers` to false in the application config, but that would cause several features of solidus to not work. Or namespace the method like spree_icon etc.

cbrunsdon:
Probably best to address this as part of #348?

mamhoff:
This is still an issue. I would vote for deprecating usage of the `icon` helper and creating a new namespaced `solidus_icon` helper.

dividedharmony:
Working on it!

jhawthorn:
Fixed by #1952! :tada: 

We will remove the deprecated method completely for our next major release, Solidus 3.0.

Thanks!

PR #1952, opened 2017-05-23
Add Spree::Admin::NavigationHelper#solidus_icon

dividedharmony:
This PR deprecates Spree::Admin::NavigationHelper#icon as both solidus and font-awesome-sass uses the #icon method. By switching to to a #solidus_icon method, we avoid helper method conflict.

See issue https://github.com/solidusio/solidus/issues/610 for details.

### DIFF
--- a/backend/app/helpers/spree/admin/navigation_helper.rb
+++ b/backend/app/helpers/spree/admin/navigation_helper.rb
@@ -116,9 +116,12 @@ module Spree
         link_to(text, url, options)
       end
 
-      def icon(icon_name)
+      def solidus_icon(icon_name)
         icon_name ? content_tag(:i, '', class: icon_name) : ''
       end
+
+      alias_method :icon, :solidus_icon
+      deprecate icon: :solidus_icon, deprecator: Spree::Deprecation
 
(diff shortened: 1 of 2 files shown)


backend/app/helpers/spree/admin/navigation_helper.rb
```diff
@@ -116,10 +116,15 @@ def link_to_with_icon(icon_name, text, url, options = {})
...
 
-      def icon(icon_name)
+      def solidus_icon(icon_name)
         icon_name ? content_tag(:i, '', class: icon_name) : ''
       end
 
+      def icon(icon_name)
+        Spree::Deprecation.warn "Using `icon' is deprecated. Use `solidus_icon' instead", caller
+        solidus_icon(icon_name)
+      end
```
mamhoff: You can do this in two lines: 

```ruby
alias_method :solidus_icon, :icon
deprecate icon: :solidus_icon, deprecator: Spree::Deprecation
```

dividedharmony: Yes, that's much cleaner. I'll make that change now

Issue #1857, opened 2017-04-22
Time-aware tax rates


mamhoff:
Tax rates are currently valid "forever". If a tax rate changes in a jurisdiction a store sells to, the tax rate records have to be updated exactly when the new tax rates comes into effect. This is not very user-friendly. 

It would instead be nice if we could set a `valid_from` and `valid_until` date on tax rates, and have them apply when the current datetime is within these two dates. If both are `nil`, tax rates can still be indefinitely valid.

vladstoick:
I will start working on this

mtylty:
@vladstoick I'm working on this as well, would you mind splitting the work? We could work on `solidus_core` and `solidus_backend` separately.

vladstoick:
@mtylty I think this one might be easier if just one person works on it so I can let you work on it if you already started working

PR #1953, opened 2017-05-23
Add validity period for Spree::TaxRate

mtylty:
This code adds a valid_from and valid_until datetime field to `Spree::TaxRate` so that tax rates can be scoped in time.

Null values will be treated as valid so that a tax rate is valid by default and you can just ignore this feature if you don't need it.

The `Spree::Calculator::DefaultTax` will always return 0 in case the tax rate is not in its validity period.

I was unsure how to implement it so I started with a scope but ended up thinking it's not really needed. @mamhoff if you could point me in the right direction I'd refactor it.

Ref. #1857

### DIFF
--- a/backend/app/views/spree/admin/tax_rates/index.html.erb
+++ b/backend/app/views/spree/admin/tax_rates/index.html.erb
@@ -20,7 +20,8 @@
       <col style="width: 15%">
       <col style="width: 10%">
       <col style="width: 10%">
-      <col style="width: 10%">
+      <col style="width: 5%">
+      <col style="width: 5%">
       <col style="width: 10%">
       <col style="width: 15%">
       <col style="width: 15%">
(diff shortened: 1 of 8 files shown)


mtylty:
This is also missing validations for dates but I can imaging we want this to have validations so that there can only be one valid tax rate at a time.

mtylty:
I ended up duplicating code that's in `Spree::Promotion` to duplicate the same `starts_at`/`expires_at` logic. It made perfect sense to me to use that logic in `Spree::TaxRate` as well.

backend/app/views/spree/admin/tax_rates/index.html.erb
```diff
@@ -22,6 +22,7 @@
       <col style="width: 10%">
       <col style="width: 10%">
       <col style="width: 10%">
+      <col style="width: 10%">
```
mamhoff: We're now way over 100% with these `width` values.

mtylty: I'll see what I can do... I'm not that good with style and html 😅

PR #1954, opened 2017-05-23
Rename assign_default_addresses! to assign_default_user_addresses!

kennyadsl:
This one is just for better documenting code via a method name. I think it makes more clear that we are taking addresses from the user and not building a generic (empty) address like it happens [here](https://github.com/solidusio/solidus/blob/9c503a42c2adbf822ef2923bc58caf8344ffbcd4/frontend/app/controllers/spree/checkout_controller.rb#L173-L179).

This PR is related to #1811


mamhoff:
@kennyadsl Would you mind rebasing on the current `master` branch?

kennyadsl:
@mamhoff done!

kennyadsl:
rebased again due to another conflict with the CHANGELOG

mamhoff:
Thank you!

Issue #1321, opened 2016-07-12
Add tooltip usage to the Solidus style guide


Mandily:
I'd like to add some more context around when to use tooltips and which one to use. I'd like to add a two column layout to the style guide with the explanation on one side and the example and code on the other. Forgive the bad markdown table to demo.

Tooltips should only be used to add supplementary information to the object they’re being attached to. They should never hide content that a user needs to complete a task. This includes simple form field helper information like required input formats (ie DD-MM-YYYY)

| Type | Example |
| --- | --- |
| **Info tooltips** These are used to add supplementary information to form labels. The info icon is the coverable area for this tooltip and should be placed at the right side of the form label. | [Code and example for this tooltip goes here] |
| **Action icon tooltips** These are used to add textual context to the action icons used with tabular data like edit, delete, and clone. They are styled with a similar colour to the icon that they’re attached to. | [Code and example for this tooltip goes here] |

tvdeyen:
I like that. Excuse me for correcting the markdown table 😊

Mandily:
I missed the leading paragraph. I've just updated it. Thanks @tvdeyen for the headings. Much better.

gus4no:
On it

jhawthorn:
Fixed by #1955

PR #1955, opened 2017-05-23
Tooltip documentation

gus4no:
#1321 

Adding a few extra information for tooltips.
 I had to override some styling using `important!` I don't really like this but....

https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/shared/_tables.scss#L36

This line is removing borders, I tried removing `!important` but it would probably break something.

image

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/sections/_style_guide.scss
+++ b/backend/app/assets/stylesheets/spree/backend/sections/_style_guide.scss
@@ -106,6 +106,15 @@
       border: 1px solid darken($color-7, 10%);
     }
   }
+
+  table.with-actions-borders {
+    th, td {
+      &.actions {
+        border-right: 1px solid #cee1f4 !important;
+        border-bottom: 1px solid #cee1f4 !important;
+      }
+    }
+  }
 }
 
 .color-variables {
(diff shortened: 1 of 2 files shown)


backend/app/assets/stylesheets/spree/backend/sections/_style_guide.scss
```diff
@@ -106,6 +106,15 @@
       border: 1px solid darken($color-7, 10%);
     }
   }
+
+  table {
```
tvdeyen: The table needs a class otherwise all tables in the style guide will match this.

backend/app/assets/stylesheets/spree/backend/sections/_style_guide.scss
```diff
@@ -106,6 +106,15 @@
...
+    th, td {
+      &.actions {
+        border-right: 1px solid #cee1f4 !important;
+        border-bottom: 1px solid #cee1f4 !important;
```
tvdeyen: I'm ok with this as the problem is due to the original - wrong - implementation.

gus4no:
Ready @tvdeyen, sorry it took long for such a simple change, hope all is good now. :)

Issue #1502, opened 2016-10-05
Hidden seed data in migrations


tudorpavel:
I was using `rake db:reset` to reset my database when I realized some default Solidus seed data is missing. After searching through the code, I noticed that the Solidus migrations contain some hidden seed data such as [the default StockLocation](https://github.com/solidusio/solidus/blob/master/core/db/migrate/20130213191427_create_default_stock.rb#L17).

The problem is that when running `rake db:reset` it is doing `rake db:schema:load`, which only recreates database tables based on the `db/schema.rb` file and the hidden seed data from migrations is never created.

I think the expected behavior should be that migrations only create the schema, while data such as [the default StockLocation](https://github.com/solidusio/solidus/blob/master/core/db/migrate/20130213191427_create_default_stock.rb#L17) is extracted into the seeds.

BravoSimone:
Working on it! 👍

BravoSimone:
I've made a dump right after the migration, and after the reset.
I'm adding the "after migration dump" marked on what to me should be added and what not.
```
# NEED TO ADD
Spree::ShippingCategory.create!([
  {name: "Default"}
])

# NEED TO ADD
Spree::StockLocation.create!([
  {name: "default", default: false, address1: nil, address2: nil, city: nil, state_id: nil, state_name: nil, country_id: nil, zipcode: nil, phone: nil, active: true, backorderable_default: true, propagate_all_variants: true, admin_name: nil, position: 0, restock_inventory: true, fulfillable: true, code: nil, check_stock_on_transfer: true}
])

# THIS DIFFERS
Spree::Store.create!([
  {name: "Sample Store", url: "example.com", meta_description: nil, meta_keywords: nil, seo_title: nil, mail_from_address: "store@example.com", default_currency: nil, code: "spree", default: true, cart_tax_country_iso: nil}
])
# IN THIS WAY
# Spree::Store.create!([
#   {name: "Spree Demo Site", url: "demo.spreecommerce.com", meta_description: nil, meta_keywords: nil, seo_title: nil, mail_from_address: "spree@example.com", default_currency: nil, code: "spree", default: true, cart_tax_country_iso: "US"}
# ])

# GENERATED IN SEEDS TOO
Spree::ReturnReason.create!([
  {name: "Accidental order", active: true, mutable: true},
  {name: "Better price available", active: true, mutable: true},
  {name: "Damaged/Defective", active: true, mutable: true},
  {name: "Different from description", active: true, mutable: true},
[...]

PR #1962, opened 2017-05-23
Add hidden seeds from migrations

BravoSimone:
As requested in issue #1502 I'm adding seeds for data previously created only in migrations.
I've added only what I thought would be necessary, see the [comment](https://github.com/solidusio/solidus/issues/1502#issuecomment-303378552) for the complete list

### DIFF
--- a/core/db/default/spree/refund_reasons.rb
+++ b/core/db/default/spree/refund_reasons.rb
@@ -1,0 +1,1 @@
+Spree::RefundReason.find_or_create_by(name: "Return processing")
--- a/core/db/default/spree/shipping_categories.rb
+++ b/core/db/default/spree/shipping_categories.rb
@@ -1,0 +1,1 @@
+Spree::ShippingCategory.find_or_create_by(name: "Default")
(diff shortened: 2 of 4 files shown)


mamhoff:
When would these files be run?

BravoSimone:
`seed.rb` requires every file into `db/default/` recursively so each time `rake db:seed` is used it will also load this data

core/db/default/spree/refund_reason.rb
```diff
@@ -0,0 +1 @@
+Spree::RefundReason.find_or_create_by(name: "Return processing")
```
kennyadsl: I think this should be added [here](https://github.com/solidusio/solidus/blob/4354667e613ceeac29c4fc30c7fbc1e5abc56083/core/db/default/spree/return_reasons.rb) instead

mamhoff:
The `cart_tax_country_iso` determines for which country taxes are shown in the cart. As stores that sell to the US do not need to show taxes in the cart, I think it's fine to leave this as `nil`.

Issue #1201, opened 2016-05-27
Create multiple page states for stock transfers


Mandily:
I would like to introduce the concept of multiple page states in the admin. I plan to go through an exercise to evaluate what a page looks like when there's no data, a little data, and a lot of data. I also plan to look at whether it meets or doesn't meet any other criteria that is required to successfully interact with that page. This exercise will result in 1. page elements being shown/not shown and 2. new messages being created with the goal of making the UI more intuitive.

**The first page I'd like to visit is the stock transfer page. I'm creating this issue for the community, so if you're interested in working on it, please leave a comment below so we don't duplicate efforts.**

**State 1: Only one stock location set**
Display only a message that tells the user the purpose of the page, and a link to get to the page where they can set multiple stock locations.
![screen shot 2016-05-27 at 3 10 33 pm](https://cloud.githubusercontent.com/assets/15271353/15622408/38b385ec-241d-11e6-9543-93c857bd5cd9.png)

**State 2: Multiple locations exist, no stock transfers have been created**
Display a message to prompt user to create a stock transfer
![screen shot 2016-05-27 at 3 19 35 pm](https://cloud.githubusercontent.com/assets/15271353/15622534/7d94dde0-241e-11e6-9f6f-e964d7753f4a.png)

**State 3: Multiple locations exist, less than 10 stock transfers exist**
Display a table of stock transfers
![screen shot 2016-05-27 at 3 46 18 pm](https://cloud.githubusercontent.com/assets/15271353/15622865/34d39d4a-2422-11e6-91df-ea89e40cb7fa.png)

**State 4: Multiple locations exist, 10 or more stock transfers exist**
Display the search filter and a table of stock transfers.
![screen shot 2016-05-27 at 3 51 45 pm](https://cloud.githubusercontent.com/assets/15271353/15622938/15add574-2423-11e6-992c-afda4a341494.png)

Feel free to suggest other improvements to messaging in addition to the ones above.

mamhoff:
I <3 your high-level overview issues, @Mandily ! I feel the fourth page (with a lot of data) is missing pagination; otherwise I think this is great!

tvdeyen:
I like this. 

So this could be narrowed down to this: 

1. We have a page missing prerequisites. Displaying action to create it
2. A page with no data. Having a link to create a record 
3. A page with only few records. We omit the filters and pagination
4. A page with lots of records. We show the full feature set of filters and pagination. 

[...]

PR #1963, opened 2017-05-23
Create multiple page states for stock transfers

jtapia:
This add the changes mentioned on #1201

**State 1: Only one stock location set**
screen shot 2017-05-23 at 12 01 21 pm

**State 2: Multiple locations exist, no stock transfers have been created**
screen shot 2017-05-23 at 12 21 23 pm

**State 3: Multiple locations exist, less than 10 stock transfers exist**
screen shot 2017-05-23 at 3 31 40 pm

**State 4: Multiple locations exist, 10 or more stock transfers exist**
screen shot 2017-05-23 at 3 32 30 pm

**State 5: Keep the search bar when search results are less than 10**
screen shot 2017-05-23 at 3 32 45 pm

### DIFF
--- a/backend/app/views/spree/admin/stock_transfers/_stock_transfer_table.html.erb
+++ b/backend/app/views/spree/admin/stock_transfers/_stock_transfer_table.html.erb
@@ -1,0 +1,46 @@
+<% if @stock_transfers.any? %>
+  <%= render 'search' if @stock_transfers.size > 9 || params[:q].present? %>
+
+  <table class="index" id="listing_stock_transfers">
+    <thead>
+      <tr>
+        <th><%= sort_link @search, :number,          Spree.t(:number) %></th>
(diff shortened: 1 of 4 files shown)


backend/app/views/spree/admin/stock_transfers/_stock_transfer_table.html.erb
```diff
@@ -0,0 +1,46 @@
+<% if @stock_transfers.any? %>
+  <%= render 'search' if @stock_transfers.size > 9 || params[:q].present? %>
```
tvdeyen: I'm not sure about this one. I see the rain behind, but we don't have this kind of behavior in any other table we have. I think this should be discussed as an overall UI improvement on a separate issue/PR

jhawthorn:
Thank you!

Closing in favour of #2263

PR #1965, opened 2017-05-23
Use validates_email_format_of gem in EmailValidator

dividedharmony:
This PR replaces a simplistic regular expression that was flagging invalid emails as valid. Instead, we will now use the validates_email_format_of gem to provide more robust email validation. This gem has a superior implementation and is more thoroughly tested.

This PR resolves https://github.com/solidusio/solidus/issues/1794

### DIFF
--- a/core/lib/spree/core/validators/email.rb
+++ b/core/lib/spree/core/validators/email.rb
@@ -1,7 +1,14 @@
+require 'validates_email_format_of'
+
 class EmailValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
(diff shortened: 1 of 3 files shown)


mamhoff:
I don't think we should add another dependency to core for this task. Most authentication solutions come with email validation of sorts, and the needs people have for this vary wildly. 

I specifically didn't add the Hack Day label to this one because it's much harder than it initally seems...

mamhoff:
We discussed this in the core team, and we all think we should not try handling email validation in core. Closing this - thank you for the effort though!

PR #1966, opened 2017-05-24
Ported shipping guide documentation from Spree

Reddshift:
This could be polished a bit more - but Clarke just announced in SolidusConf that he would be hiring a technical writer to work on implementation guides and documentation.  so I thought I would create a pull request with what I have so far.  If this gets merged and helps some people in the meantime, great!  If we all feel like it would be better to wait for a more polished version - great too! 

I Started with a copy of the Spree doc was pulled (https://github.com/spree/spree-guides/blob/master/content/developer/core/shipments.md)  and made the following edits (among others)
* Confirmed as much as possible that Solidus still works as described in the spree documentation.  Fixed as needed.
* Refined descriptions and explanations in (hopefully) appropriate ways
* Only used ticks for things that could actually be in code or typed at a terminal or console.
* Reorganized examples to come after definitions
* Changed the “Note” callout pattern to a bold word with italic text instead of horizontal rules
* Replaced “Spree” with “Solidus” as appropriate

Questions for feedback:
* Stock splitting documentation - I just confirmed the paths are correct and files exist in solidus, I didn’t confirm that function in solidus has not changed
* Should we include the documentation (essentially a recommendation or promotion) of the solidus_active_shipping extension?  The extension itself acknowledges that it is old, complex, and needs some work. I didn’t have time to test the example code in solidus, but I did confirm that all objects/classes referred to exist, and the examples look sound.
* Wondering how to appropriately acknowledge that the document originated from the spree guide…

### DIFF
--- a/guides/shipping.md
+++ b/guides/shipping.md
@@ -1,0 +1,513 @@
+---
+title: "Shipments"
+section: "core"
+---
+
+## Overview
+
+This guide explains how Spree represents shipping options and how it calculates expected costs, and shows how you can configure the system with your own shipping methods. After reading it you should know:
+
+* how shipments and shipping are implemented in Spree
+* how to specify your shipping structure
+* how split shipments work
+* how to configure products for special shipping treatment
+* how to capture shipping instructions
+
+Spree uses a very flexible and effective system to calculate shipping, accommodating the full range of shipment pricing: from simple flat rate to complex product-type- and weight-dependent calculations.
+
+The Shipment model is used to track how items are delivered to the buyer.
+
+Shipments have the following attributes:
+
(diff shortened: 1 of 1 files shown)


mamhoff:
Clarkes announcement does not preclude the community from porting Spree documentation. The person that will be hired will have lots to do either way, so I suggest going on with community-delivered guides. 

@Reddshift this is an amazing effort, an I would very much like to see this merged as it still is a pretty good guide.

Reddshift:
Sounds good @mamhoff -- porting the Spree guides is fairly low effort compared to the extensive work that could be done to take documentation further. I agree that it is worth doing, even just to have some documentation until someone is hired.  I'll try to fit in porting some more...

PR #1967, opened 2017-05-24
Assign default user addresses in checkout controller

### LINKED ISSUE #1811: User address not added to order after signing in

kennyadsl:
It fixes #1811:

When **cart -> address** state change happens and user is not yet logged in, the `assign_default_addresses!` callback will not associate any address to the order regardless if the user will immediately later perform a login and had some address associated.
By calling `assign_default_addresses!` method in the `before_address` checkout controller's callback we'll be sure we always set addresses correctly before entering the address step.

The only cons I see is that we call this method much often (every time we hit the checkout address page) but it should be safe since, as @aaronrussell pointed out, it does nothing when the order has no user or when the order already has addresses associated.

I'm not sure if we want to remove the state machine callback now and move this functionality at a controller level, but we'll probably want to add `assign_default_addresses!` to backend and api controllers as well.

This PR also adds checkout feature specs that allows to reproduce this issue.

### DIFF
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -171,8 +171,9 @@ module Spree
     end
 
     def before_address
-      # if the user has a default address, a callback takes care of setting
-      # that; but if he doesn't, we need to build an empty one here
+      @order.assign_default_user_addresses!
+      # If the user has a default address, the previous method call takes care
+      # of setting that; but if he doesn't, we need to build an empty one here
       default = {country_id: Spree::Country.default.id}
       @order.build_bill_address(default) unless @order.bill_address
       @order.build_ship_address(default) if @order.checkout_steps.include?('delivery') && !@order.ship_address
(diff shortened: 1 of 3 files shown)


kennyadsl:
@jhawthorn @mamhoff this is not going to change any value on a GET request. It just [sets the address without saving](https://github.com/solidusio/solidus/blob/e6188ee31c25b65745735ffc7042a23860c99ae1/core/app/models/spree/order.rb#L701-L708).

Sinetheta:
Thanks @kennyadsl, I bumped into this problem just the other day. Can confirm this resolves the issue 😀

mtomov:
That's great! Thanks so much! It's those small usability things, which are most annoying. Thanks again!

PR #1974, opened 2017-05-29
Deprecate provider in favour of payment method type

tvdeyen:
Payment method types get referenced as providers in the admin payment methods configuration.

In order to remove confusion about what a payment method and a provider actually is we rename usages of provider to payment method type.

Later on a real payment provider class will get introduced, that hold all commons for multiple payment methods, like credentials and api communication through its gateway.

### DIFF
--- a/backend/app/controllers/spree/admin/payment_methods_controller.rb
+++ b/backend/app/controllers/spree/admin/payment_methods_controller.rb
@@ -2,8 +2,8 @@ module Spree
   module Admin
     class PaymentMethodsController < ResourceController
       skip_before_action :load_resource, only: :create
-      before_action :load_providers
-      before_action :validate_payment_method_provider, only: [:create, :update]
+      before_action :load_payment_method_types, except: [:index]
+      before_action :validate_payment_method_type, only: [:create, :update]
 
       respond_to :html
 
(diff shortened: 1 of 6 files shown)


tvdeyen:
Will rebase on top of #1975 if we decided to make this change.

backend/app/controllers/spree/admin/payment_methods_controller.rb
```diff
@@ -49,23 +49,38 @@ def collection
...
 
       def load_providers
-        @providers = Spree::PaymentMethod.providers.sort_by(&:name)
+        load_payment_methods
       end
+      deprecate load_providers: :load_payment_methods, deprecator: Spree::Deprecation
 
-      def validate_payment_method_provider
+      def load_payment_methods
+        @payment_methods = Rails.application.config.spree.payment_methods.sort_by(&:name)
```
kennyadsl: I think having a `@payment_methods` ivar can be confusing since, as a developer, I expect to find currently persisted `Spree::PaymentMethod` there instead now it contains possible types of `Spree::PaymentMethod`. What about some more semantic variable name like `@payment_method_types`?

tvdeyen: I see. `payment_method_types` is ok, or `configured_payment_methods` could also work, no? @mamhoff thoughts?

kennyadsl: or `available_payment_methods`?

mamhoff: I like `@payment_method_types`. Or even `@available_payment_method_types`. I agree with @kennyadsl s sentiment that `payment_methods` sounds like instances, but the collection is an array of class names.

tvdeyen: @kennyadsl @mamhoff updated. Please re-review

backend/app/controllers/spree/admin/payment_methods_controller.rb
```diff
@@ -49,16 +49,31 @@ def collection
...
+        deprecator: Spree::Deprecation
+
+      def validate_payment_method_type
         requested_type = params[:payment_method].delete(:type)
-        @payment_method_type = @providers.detect do |klass|
+        @payment_method_type = @payment_method_types.detect do |klass|
           klass.name == requested_type
         end
+        # TODO: Remove `@payment_method_type` instance var once `load_providers` gets removed.
+        @payment_method_type = @payment_method_type
```
mamhoff: If you move the comment up to new line 70, we shouldn't really need this line, should we?

tvdeyen: Oh, good catch. Due to the re-rename 😊

tvdeyen:
@mamhoff fixed

core/app/models/spree/payment_method.rb
```diff
@@ -25,6 +25,8 @@ class PaymentMethod < Spree::Base
     include Spree::Preferences::StaticallyConfigurable
 
     def self.providers
+      Spree::Deprecation.warn 'Spree::PaymentMethod.providers will be deleted without replacement. ' \
```
jhawthorn: It's not without replacement. The replacement is `Rails.application.config.spree.payment_methods`

tvdeyen: Fixed. Thanks

tvdeyen:
@jhawthorn rebased on latest master. Might have another look?

gmacdougall:
@tvdeyen Can we get a rebase on this?

tvdeyen:
@gmacdougall here you go

PR #1975, opened 2017-05-29
Use translated model names in admin payment methods form

tvdeyen:
This PR is one of many that follow to remove the confusion around payment methods, payment providers and gateways we have for a long time in our code base.

There is lots of confusing parts in administrating payment methods:

- The payment method type is called a provider (although it may be only one payment method implementation of many from the same provider) 
- The payment method classes listed as providers are represented by its class names instead of a meaningful human readable name
- Form params passed to the controller are confused by the STI nature of payment method classes instead of using one concise name
- The active flag is represented by two radio buttons instead of a checkbox

### Before

store credit - payment methods - before

### After

store credit - payment methods - after


tvdeyen:
Need to rebase #1974 if we decide to make this change

tvdeyen:
Need to rebase after #1977 was merged if we decide to make this change

tvdeyen:
Rebased, ready to review

tvdeyen:
Looks even better with #2040 

preference fields - payment method - after

tvdeyen:
Rebased again. Still ready for review

jhawthorn:
We're going to backport and include this in 2.3.0. Since 2.3.0 includes the renames of the bogus payment method classes, it makes sense to include this and have them human readable (and also would avoid changing how they're displayed here in two separate consecutive releases).

Issue #568, opened 2015-12-04
Lint Factories


cbrunsdon:
We should run all our factories through FactoryGirl.lint, as discussed on Slack I highly suspect we won't pass.

This should be a task accomplish-able by anyone with medium ruby/rails/factory girl experience, but could get very tricky depending on which factories don't build.

wgriffiths:
I will start looking into this

travisp:
Yes, I just tried importing spree factories into an app to use for testing, and the app lints all factories by default when all specs are run. It came up with these errors (after I modified the default user factory to have an 8 character password):

```
       * adjustment - Validation failed: Order can't be blank (ActiveRecord::RecordInvalid)
       * tax_adjustment - Validation failed: Order can't be blank (ActiveRecord::RecordInvalid)
       * customer_return_without_return_items - Validation failed: Return items can't be blank (ActiveRecord::RecordInvalid)
       * image - No such file or directory @ rb_sysopen - /Users/travis/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/solidus_core-1.1.0/spec/fixtures/thinking-cat.jpg (Errno::ENOENT)
       * refund - Validation failed: Amount is greater than the allowed amount. (ActiveRecord::RecordInvalid)
       * base_shipping_method - Validation failed: Calculator can't be blank (ActiveRecord::RecordInvalid)
       * stock_packer - undefined method `save!' for #<Spree::Stock::Packer:0x007feff84d6378> (NoMethodError)
       * stock_package - undefined method `save!' for #<Spree::Stock::Package:0x007fefdd080320> (NoMethodError)
       * stock_package_fulfilled - undefined method `save!' for #<Spree::Stock::Package:0x007feff84ae490> (NoMethodError)
       * stock_item - Validation failed: Variant has already been taken (ActiveRecord::RecordInvalid)
       * stock_movement - Validation failed: Variant has already been taken (ActiveRecord::RecordInvalid)
       * stock_transfer_with_items - Validation failed: Code has already been taken (ActiveRecord::RecordInvalid)
       * receivable_stock_transfer_with_items - Validation failed: Code has already been taken (ActiveRecord::RecordInvalid)
       * store_credit_event - PG::NotNullViolation: ERROR:  null value in column "action" violates not-null constraint
[...]

PR #1976, opened 2017-05-29
Fix the last factories lint offences.

AlessioRocco:
Related to this issue https://github.com/solidusio/solidus/issues/568.

I'm not sure we want the linter run at the beginning of each spec suite, for now I've put it here https://github.com/nebulab/solidus/blob/908cf3859140aae6077dab2af4623d1133ae5f93/core/lib/spree/testing_support/factories/factories_linter.rb.
I've also skip `stock_packer` factory because it's not an Active Record type, and `customer_return_without_return_items` because it's a wanted "wrong" factory, `customer_return` can't be without `return_items` due a validation.

### DIFF
--- a/core/lib/spree/testing_support/factories/factories_linter.rb
+++ b/core/lib/spree/testing_support/factories/factories_linter.rb
@@ -2,7 +2,7 @@ RSpec.configure do |config|
   config.before(:suite) do |example|
     DatabaseCleaner.start
     factories_to_lint = FactoryGirl.factories.reject do |factory|
-     [:stock_packer, :customer_return_without_return_items].include?(factory.name)
+      [:stock_packer, :customer_return_without_return_items].include?(factory.name)
     end
 
     FactoryGirl.lint factories_to_lint
--- a/core/lib/spree/testing_support/factories/stock_transfer_factory.rb
+++ b/core/lib/spree/testing_support/factories/stock_transfer_factory.rb
@@ -1,9 +1,12 @@
 FactoryGirl.define do
+  sequence(:source_code) { |n| "SRC#{n}" }
+  sequence(:destination_code) { |n| "DEST#{n}" }
+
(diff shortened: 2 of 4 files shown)


bbuchalter:
> I'm not sure we want the linter run at the beginning of each spec suite

We run Rubocop as part of the test suite, why wouldn't we lint the factories as well? I think the factories form the foundation upon which the test suite is built; if we don't maintain the factories, our test suite becomes difficult to maintain. I vote we run it as a breakable part of the build.

core/lib/spree/testing_support/factories/factories_linter.rb
```diff
@@ -0,0 +1,11 @@
+RSpec.configure do |config|
+  config.before(:suite) do |example|
```
jhawthorn: I'd prefer that this was a normal spec. As a `before(:suite)`, this makes it impossible to run a single other spec without running this first. It also makes this not report itself as a passing or failing spec.

BenMorganIO: @jhawthorn Yep, that can be a problem. We could provide a `config.files_to_run_one?` or something along those lines; can't really remember the methods name.

BenMorganIO:
I think it's completely fine to run the lint at the beginning of the suite. We could also have a rake task that simply lints all of the factories. I'm not sure if Factory Girl comes with this or not however, it wouldn't be too hard to build and add to CircleCI.

But, there really isn't much of a problem if we run it apart of the suite. I can only see it becoming a true problem if the suite takes a noticeably long time to run.

kennyadsl:
This is interesting but we still have no clear ideas about how to proceed here and what's the best way to lint factories inside the spec suite runs. Talking IRL with @AlessioRocco about moving this commit (138d6561ff16dfac42bffd1cdd49bf9b4942a9d7) into a new PR which fixes the `stock_transfer` factory if called multiple times.

PR #1977, opened 2017-05-29
Rename bogus gateways into payment methods

tvdeyen:
In the ongoing effort to remove confusion around payment methods, gateways and providers we rename:

- `Spree::Gateway::Bogus` into `Spree::PaymentMethod::BogusCreditCard` 
- `Spree::Gateway::BogusSimple` into `Spree::PaymentMethod::SimpleBogusCreditCard`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Solidus 2.3.0 (master, unreleased)
 
+- Renamed bogus payment methods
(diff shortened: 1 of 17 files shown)


tvdeyen:
Need to rebase #1975 if we decide to make this change

tvdeyen:
Closing as I plan to move all payment methods into the `PaymentMethods` module namespace instead, because of constant lookup issues in Rails in combination with STI.

PR #1978, opened 2017-05-29
Rename method_type into partial_name

tvdeyen:
`PaymentMethod#method_type` is a confusing name for a method returning the name of the partial to use.

As we cannot use `to_partial_path` (we use it in four different places), we use `partial_name` as more descriptive name.

This removes the key from the `/api/orders/:id` API (How ever this landed up there). As I can't imagine a good approach for deprecating API keys renaming is the only option I have.

### DIFF
--- a/api/app/views/spree/api/orders/show.v1.rabl
+++ b/api/app/views/spree/api/orders/show.v1.rabl
@@ -2,7 +2,7 @@ object @order
 extends "spree/api/orders/order"
 
 child :available_payment_methods => :payment_methods do
-  attributes :id, :name, :method_type
+  attributes :id, :name, :method_type, :partial_name
 end
 
 child :billing_address => :bill_address do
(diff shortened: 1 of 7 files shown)


api/app/views/spree/api/orders/show.v1.rabl
```diff
@@ -2,7 +2,7 @@ object @order
 extends "spree/api/orders/order"
 
 child :available_payment_methods => :payment_methods do
-  attributes :id, :name, :method_type
+  attributes :id, :name, :partial_name
```
jhawthorn: We probably want to keep the old field around (until the next major version)

tvdeyen: You mean additionally? Yes, this is a good idea

tvdeyen:
@jhawthorn rebased on master and re-added the `method_type` key to the api response.

gmacdougall:
@tvdeyen can we get a rebase on this?

tvdeyen:
@gmacdougall done

jhawthorn:
@tvdeyen Sorry, this was reverted in https://github.com/solidusio/solidus/pull/2057 as the deprecation warnings were causing CI failures

PR #1980, opened 2017-05-30
Improve order factories

bbuchalter:
This change mostly adds improved test coverage for the order factories, but does modify the behavior of the order factories as well:
* `order_with_totals` can now be built instead of just being created
* `completed_order_with_totals` was calling `refresh_shipment_rates` which was causing the `shipping_method` argument passed to the factory to be ignored. Removing that call resulted in no test failures and the requested `shipping_method` to be assigned correctly.

One question this PR raises for me is building the factories does not result in the same output as creating. If we don't care about being able to build, I would like to find some way to raise an exception when we call build on a factory known not to produce the expected outputs. I'm not sure how to do this.

### DIFF
--- a/core/lib/spree/testing_support/factories/order_factory.rb
+++ b/core/lib/spree/testing_support/factories/order_factory.rb
@@ -18,10 +18,14 @@ FactoryGirl.define do
       line_items_price BigDecimal.new(10)
     end
 
+    # TODO: Improve the name of order_with_totals factory.
     factory :order_with_totals do
-      after(:create) do |order, evaluator|
-        create(:line_item, order: order, price: evaluator.line_items_price)
-        order.line_items.reload # to ensure order.line_items is accessible after
+      after(:build) do |order, evaluator|
+        order.line_items << build(
+          :line_item,
+          order: order,
+          price: evaluator.line_items_price
+        )
       end
     end
 
@@ -37,7 +41,7 @@ FactoryGirl.define do
         stock_location { create(:stock_location) }
       end
 
-      after(:create) do |order, evaluator|
+      after(:build) do |order, evaluator|
         evaluator.stock_location # must evaluate before creating line items
 
         evaluator.line_items_attributes.each do |attributes|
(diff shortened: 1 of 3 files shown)


core/lib/spree/testing_support/factories/order_factory.rb
```diff
@@ -18,10 +18,14 @@
       line_items_price BigDecimal.new(10)
     end
 
+    # TODO: Improve the name of this factory.
```
tvdeyen: When using a tool to parse TODOs it would help to actually have the name of the factory in the message.

```ruby
# TODO: Improve the name of the order_with_totals factory
```

bbuchalter: Fixed in a fixup commit 745cdcf9

bbuchalter:
Thanks for the quick and positive feedback. There's still a lot more I'd like to do with these factories, but I think this is a good step forward.

bbuchalter:
I'm considering rebasing these changes into a single commit. I left them separate for the sake of review, but for the sake of history, I think they may belong all together. Thoughts?

tvdeyen:
👍 Agreed

bbuchalter:
Rebased into a single commit.

core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
```diff
@@ -1,6 +1,98 @@
...
+        total: 8.0,
+        state: expected_order_state
+      )
+    end
+  end
+
+  context 'when shipments should be taxed' do
+    let!(:ship_address) { create(:address) }
+    let!(:tax_zone) { create(:global_zone) } # will include the above address
+    let!(:tax_rate) { create(:tax_rate, amount: 0.10, zone: tax_zone, tax_categories: [tax_category]) }
```
bbuchalter: I'm trying to backport this PR to Solidus 1.3 to help with an upgrade I'm working on. I needed to change `tax_categories` here to `tax_category`, but it seems taxes aren't being added to shipments in my backport. Any thoughts?

PR #1981, opened 2017-05-30
Dynamically define methods for payment sources WIP

skukx:
(This is a work in progress)
In the wallet define methods for each descendant of the `Spree::PaymentSource` class.
This will allow users to do the following:

```ruby
user.wallet.credit_cards
user.wallet.gift_cards
```

The reason this is a work in progress is that this will not work
expectedly in Development. Because of lazy loading, the descendants may
not be loaded yet causing `Spree::PaymentSource.descendants` to return
an empty array.  Due to this, it may be best to not do this dynamically.

### DIFF
--- a/core/app/models/spree/wallet.rb
+++ b/core/app/models/spree/wallet.rb
@@ -12,6 +12,7 @@ class Spree::Wallet
 
   def initialize(user)
     @user = user
+    define_payment_source_methods
   end
 
   # Returns an array of the WalletPaymentSources in this wallet.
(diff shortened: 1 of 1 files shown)


core/app/models/spree/wallet.rb
```diff
@@ -73,4 +74,48 @@ def default_wallet_payment_source=(wallet_payment_source)
...
+  #
+  # @return [NilClass]
+  #
+  def define_payment_source_methods
+    # Loop through each available payment source class.
+    # @todo How do we compensate for lazy loading?
+    Spree::PaymentSource.descendants.each do |source|
+      # Underscore and pluralize
+      # Spree::CreditCard => 'credit_cards'
+      method_name = source.name.underscore.split('/').last.pluralize
```
tvdeyen:
```ruby
source.name.demodulize.tableize
```
should return the same result and is much shorter.

tvdeyen:
👏 I really like the idea

mamhoff:
I think I'm not a big fan of this. Payment Sources should quack like payment sources, and they have their respective partials for being rendered. If a specific store needs logic around types of payment sources in the wallet, they should implement it in their host apps. 

I would also not like a feature in core that breaks development mode in Rails. 

It is definitely an interesting idea, but drawbacks vs advantages I think yield a rather not...

skukx:
@mamhoff I agree, I don't think there is a good way to do this dynamically so I'm thinking of another way that this can be done similarly but extra methods will need to be defined by host apps

skukx:
Closing this in favor of https://github.com/solidusio/solidus/pull/1992

PR #1983, opened 2017-05-30
Make _report_order_criteria more shareable

skukx:
When creating a custom report on the reports controller, I noticed
that this shared form would always post the the :sales_total action of
the reports controller (`Spree::Admin::ReportsController`).

See
https://github.com/solidusio/solidus/blob/v2.1/backend/app/views/spree/admin/shared/_report_order_criteria.html.erb#L1

In order to allow future/custom reports be able to actually share this partial,
pass the key into the partial so that the partial will post to the right
action.

### DIFF
--- a/backend/app/views/spree/admin/reports/sales_total.html.erb
+++ b/backend/app/views/spree/admin/reports/sales_total.html.erb
@@ -10,7 +10,7 @@
 <% end %>
 
 <% content_for :table_filter do %>
-  <%= render partial: 'spree/admin/shared/report_order_criteria', locals: {} %>
+  <%= render partial: 'spree/admin/shared/report_order_criteria', locals: { action: :sales_total } %>
 <% end %>
 
 <table class="admin-report" data-hook="sales_total">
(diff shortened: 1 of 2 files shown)


backend/app/views/spree/admin/shared/_report_order_criteria.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= search_form_for @search, url: spree.sales_total_admin_reports_path  do |s| %>
+<%= search_form_for @search, url: spree.send("#{key}_admin_reports_path".to_sym)  do |s| %>
```
tvdeyen: I would like us to use:

```ruby
spree.url_for(controller: 'admin/reports', action: action)
```

instead of interpolating strings.

backend/app/views/spree/admin/reports/sales_total.html.erb
```diff
@@ -10,7 +10,7 @@
 <% end %>
 
 <% content_for :table_filter do %>
-  <%= render partial: 'spree/admin/shared/report_order_criteria', locals: {} %>
+  <%= render partial: 'spree/admin/shared/report_order_criteria', locals: { key: :sales_total } %>
```
skukx: @tvdeyen What if I use `report_key` to be consistent with the controller https://github.com/solidusio/solidus/blob/master/backend/app/controllers/spree/admin/reports_controller.rb#L11 ? otherwise I think action is a good name too

gmacdougall:
Thanks for the contribution!

PR #1984, opened 2017-05-31
Select all shipping rates in the Estimator

### LINKED ISSUE #1711: Not able to use backend shipping methods in admin
**Steps to Recreate:**
1. Select “Backend” from the Display dropdown on a Shipping Method in admin and save
2. Create an order in admin and get to the Delivery Step
3. Select a shipping method for the order. Notice that the shipping method you updated is not showing as an option, although “Backend” was selected

Expected Behavior: The shipping method should have been showing in the dropdown as an option since “Backend” was selected and admin users should be able use those shipping methods.

Actual Behavior: That shipping method is not showing as an shipping option

Solidus 2.0

ericsaupe:
This change selects all shipping rates using the Estimator and
dictates the selection of those rates to the various views. For
example, in the backend views we don't limit what rates are shown
whereas on the frontend we limit to only what users can see or
if a backend rate is already selected display that.

Possible fix for #1237 and another solution similar to #1739. Fixes #1711

### DIFF
--- a/core/app/models/spree/stock/estimator.rb
+++ b/core/app/models/spree/stock/estimator.rb
@@ -11,13 +11,16 @@ module Spree
       #   those marked frontend if truthy
       # @return [Array<Spree::ShippingRate>] the shipping rates sorted by
       #   descending cost, with the least costly marked "selected"
-      def shipping_rates(package, frontend_only = true)
+      def shipping_rates(package, frontend_only = nil)
         raise ShipmentRequired if package.shipment.nil?
         raise OrderRequired if package.shipment.order.nil?
+        if frontend_only.present?
+          Spree::Deprecation.warn "Using frontend_only arg is deprecated. All shipping rates will always be returned.", caller
+        end
 
         rates = calculate_shipping_rates(package)
-        rates.select! { |rate| rate.shipping_method.available_to_users? } if frontend_only
-        choose_default_shipping_rate(rates)
+        frontend_rates = rates.select { |rate| rate.shipping_method.available_to_users? }
+        choose_default_shipping_rate(frontend_rates)
         Spree::Config.shipping_rate_sorter_class.new(rates).sort
       end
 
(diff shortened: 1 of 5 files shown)


frontend/app/views/spree/checkout/_delivery.html.erb
```diff
@@ -38,6 +38,7 @@
           <h5 class="stock-shipping-method-title"><%= Spree.t(:shipping_method) %></h5>
           <ul class="field radios shipping-methods">
             <% ship_form.object.shipping_rates.each do |rate| %>
+              <% next unless ship_form.object.selected_shipping_rate == rate || rate.shipping_method.available_to_users? %>
```
kennyadsl: I think this logic could be too complex for a view. What about creating a method into `Spree::ShippingRate` similar to:

```ruby
def available_to_user?
  shipping_method.available_to_users? || shipment.selected_shipping_rate == self
end
```

and here just do:

```erb
<% ship_form.object.shipping_rates.select(&:available_to_user?).each do |rate| %>
```

Do you think something like this would work and reduce view complexity?

core/app/models/spree/stock/estimator.rb
```diff
@@ -11,13 +11,13 @@ class OrderRequired < StandardError; end
       #   those marked frontend if truthy
       # @return [Array<Spree::ShippingRate>] the shipping rates sorted by
       #   descending cost, with the least costly marked "selected"
-      def shipping_rates(package, frontend_only = true)
+      def shipping_rates(package, _frontend_only = true)
```
kennyadsl: do we need to deprecate this argument if it is not used anymore? Maybe also changing method documentation?

frontend/app/views/spree/checkout/_delivery.html.erb
ericsaupe: I like this idea. I actually was thinking about it after I submitted this PR and thought that would clean it up adding a method like that. I'll go ahead and do that.

core/app/models/spree/stock/estimator.rb
ericsaupe: I think so. What is the best way to deprecate an argument?

mamhoff: Just make it `nil` and add a `Spree::Deprecation.warn` if it is present.

isindexer:
Hi, will it fix this issue too : #1711 ?
Thx

ericsaupe:
@isindexer yes

isindexer:
thanks @ericsaupe

PR #1982, opened 2017-06-01
Clone action should be a POST

Draiken:
When cloning big products, specially ones with multiple variants and images, the clone action might take a lot of time to finish.

This is not a problem on itself, but it can generate 502 responses depending on the environment.

Once again, this is not a problem by itself, but in conjunction with the fact that this is done through a GET request, browsers may (and will) retry the request automatically, since GET requests are not supposed to change anything on the server side.

As a result, if a product takes too long to clone, a browser like Chrome might re-send the clone request multiple times creating several clones of the product. I've tested this locally on a particularly big product with several images (touching Google Cloud Storage, which is slow) and after a single click on "clone" I ended up with 6 clones.

Since logically a clone operation does create a new resource (similar to a create) if we use POST for this operation, this entire problem will be avoided in the future.

If people agree this is a problem I can submit a PR to fix this issue.

### DIFF
--- a/backend/app/helpers/spree/admin/navigation_helper.rb
+++ b/backend/app/helpers/spree/admin/navigation_helper.rb
@@ -79,6 +79,7 @@ module Spree
 
       def link_to_clone(resource, options = {})
         options[:data] = { action: 'clone' }
+        options[:method] = :post
         link_to_with_icon('copy', Spree.t(:clone), clone_object_url(resource), options)
       end
 
--- a/backend/config/routes.rb
+++ b/backend/config/routes.rb
@@ -45,7 +45,7 @@ Spree::Core::Engine.routes.draw do
         end
       end
       member do
-        get :clone
+        post :clone
       end
       resources :variants do
         collection do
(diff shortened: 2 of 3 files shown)


mamhoff:
I also agree this is a good fix to an existing problem. Would you mind adding a request spec for this change though?

jasonfb:
Safari is particularly egregious about pre-fetching GET requests, for example, when it renders the "frequently visited sites", it will prefetch all those URLs.. I've noticed non-idempodent GET requests that cause problems like this are especially exacerbated in Safari (which I think I may be one of the last users of on the planet)

Draiken:
@mamhoff of course... I'll add it as soon as I have time
When you say request spec, you mean a feature, or controller spec?

Draiken:
Added a controller spec, but since this behaviour is already tested on the model level by testing `#duplicate` I feel this spec doesn't add much value. The feature spec for it is more than enough IMO

Unless we are testing a route no longer exists, or the view to see the generated button, I'm not sure what you want me to test

mamhoff:
@Draiken The controller spec is perfect. It's more about documenting that we want this to be a `post` action, for the reasons you mention in the PR.

gmacdougall:
Thanks for the contribution!

PR #1986, opened 2017-06-01
Require credit card dependency in gateway class

tvdeyen:
Without requiring the spree/credit_card dependency in the Spree::Gateway class
Rails cannot lookup the constants correctly and will return Spree::CreditCard
when loading the not-even-existing Spree::Gateway::CreditCard constant.

This leads to strange errors with all payment methods inheriting from Spree::Gateway

Without this change:

    Spree::Gateway::CreditCard
    => Spree::CreditCard # Wrong Constant returned
    Spree::CreditCard
    => Spree::CreditCard # Constant now loaded
    Spree::Gateway::CreditCard
    => NameError: uninitialized constant Spree::Gateway::CreditCard

With this change:

    Spree::Gateway::CreditCard
    => NameError: uninitialized constant Spree::Gateway::CreditCard

### DIFF
--- a/core/app/models/spree/gateway.rb
+++ b/core/app/models/spree/gateway.rb
@@ -1,3 +1,5 @@
+require_dependency 'spree/credit_card'
+
 module Spree
   # A concrete implementation of `Spree::PaymentMethod` intended to provide a
   # base for extension. See https://github.com/solidusio/solidus_gateway/ for


tvdeyen:
🤔 

```ruby
>> Spree::Gateway::StoreCredit
=> Spree::StoreCredit
```

tvdeyen:
Closing. The Solution is to move all payment methods into the `PaymentMethods` (observe the plural name) module namespace, as Rails STI has mayor issues with class names having a module namespace that is also a class constant, like `PaymentMethod` is.

jordan-brough:
@tvdeyen STI auto-loading is a valid issue, I don't think your problem here is an STI issue, is it?

I think the same thing happens with any class that hasn't been loaded yet:
```ruby
>> defined?(Spree::Carton)
=> nil

>> Spree::Country::Carton
=> Spree::Carton(id: integer, number: string, external_number: string, stock_location_id: integer, address_id: integer, shipping_method_id: integer, tracking: string, shipped_at: datetime, created_at: datetime, updated_at: datetime, imported_from_shipment_id: integer)

>> Spree::Country::Carton
NameError: uninitialized constant Spree::Country::Carton
```

PR #1988, opened 2017-06-02
Do not show broken links in admin product view when product is deleted

### LINKED ISSUE #1987: Admin Product page for a deleted item shows errors when clicking around
### Steps to reproduce
1. delete an item from the products list
2. check 'Show Deleted' and hit 'Search'
3. click on deleted item
4. click 'Images', 'Variants', 'Prices' and 'Product Property' tabs --> throw errors

### Expected behavior
Those tabs should be hidden if they give errors

### Actual behavior
Tabs are all shown and clickable

### System configuration
**Solidus Version**:
solidusio/solidus/master at 4354667e61

[Gemfile.lock](https://gist.github.com/laurawadden/19efb31ecf42cdebdfa1447e663767df)

laurawadden:
Fixes #1987 - Admin product page for a deleted item shows errors when clicking around.

Hides broken tabs (Images, Variants, Prices, Product Stock) in Admin Product View if the product is deleted.

### DIFF
--- a/backend/app/views/spree/admin/shared/_product_tabs.html.erb
+++ b/backend/app/views/spree/admin/shared/_product_tabs.html.erb
@@ -9,19 +9,19 @@
       <% end if can?(:admin, Spree::Product) %>
       <%= content_tag :li, class: ('active' if current == 'Images') do %>
         <%= link_to Spree.t(:images), spree.admin_product_images_url(@product) %>
-      <% end if can?(:admin, Spree::Image) %>
+      <% end if can?(:admin, Spree::Image) && !@product.deleted? %>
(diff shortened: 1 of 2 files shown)


laurawadden:
Thanks for the comments/feedback.

I don't think it's worth adding the redirects in the controller unless you come across the problem (in real life) at some point in the future. This is really a small, dark, unvisited corner of the application.

@mamhoff As for looking more into the soft-deleted variants and prices (and why they aren't just there?), I think that's a good idea for the future.

mamhoff:
The variant view has been fixed in https://github.com/solidusio/solidus/pull/1804, so we can actually display that link. Would you mind removing the if condition there, and rebasing?

PR #1991, opened 2017-06-02
Add missing ) to admin Spree::Reimbursement edit view

acreilly:
Missing ) breaks the view for the admin reimbursement's edit page. This affects solidus versions v1.4 and v2.0


ericsaupe:
Just for reference, this PR is fixing older versions of Solidus, from 2.1 on the fix was in with this commit https://github.com/solidusio/solidus/commit/da223aa7466dc3f0cca16228a02e8271804e4bdc

jhawthorn:
For some reason there's an unrelated issue on CircleCI preventing this from passing :thinking: but the build is passing locally

jhawthorn:
I will try to cut a 1.4 patch release shortly

PR #1992, opened 2017-06-02
Define payment source helper methods on Wallet

skukx:
(This is a work in progress)
In the wallet define methods for each descendant of the `Spree::PaymentSource` class.
This will allow users to do the following:

```ruby
user.wallet.credit_cards
user.wallet.gift_cards
```

The reason this is a work in progress is that this will not work
expectedly in Development. Because of lazy loading, the descendants may
not be loaded yet causing `Spree::PaymentSource.descendants` to return
an empty array.  Due to this, it may be best to not do this dynamically.

### DIFF
--- a/core/spec/models/spree/wallet_spec.rb
+++ b/core/spec/models/spree/wallet_spec.rb
@@ -13,7 +13,7 @@ describe Spree::Wallet, type: :model do
     context 'when source is a Spree::PaymentSource' do
       it 'defines a helper method for the class' do
         described_class.payment_source_helper payment_source
-        expect(wallet.respond_to?(:fake_payment_sources)).to eql true
+        expect(wallet).to respond_to :fake_payment_sources
       end
     end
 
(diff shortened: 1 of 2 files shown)


skukx:
This will allow other apps by using a concern such as:
```ruby
module Spree::Wallet::GiftCards
  extend ActiveSupport::Concern
 
  included do
    payment_source_helper Spree::GiftCard
  end
end

Spree::Wallet.include Spree::Wallet::GiftCards

user.wallet.gift_cards #=> [Array<Spree::GiftCard>]
```

core/spec/models/spree/wallet_spec.rb
```diff
@@ -6,6 +6,24 @@
...
   subject(:wallet) { Spree::Wallet.new(user) }
 
+  describe '.payment_source_helper' do
+    let(:payment_source) { Object.const_set('FakePaymentSource', Class.new(Spree::PaymentSource)) }
+    let(:non_payment_source) { Object.const_set('FakeNonPaymentSource', Class.new) }
+
+    context 'when source is a Spree::PaymentSource' do
+      it 'defines a helper method for the class' do
+        described_class.payment_source_helper payment_source
+        expect(wallet.respond_to?(:fake_payment_sources)).to eql true
```
mtomov: I'd rather replace the `respond_to` expectation with one, which checks the actual contents of the returned collection. I think that `eq` is preferred to `eql`.

Great that you added a test for the ArgumentError.

kennyadsl:
@skukx this is a very good and well documented change. Anyway, talking with other core team members we decided to avoid introducing this code. The benefits it brings in are not enough to justify metaprogramming methods and the drawbacks that can be introduced from this. Thanks anyway!

Issue #1971, opened 2017-05-25
Document or deprecate use of HTTP_SPREE_STORE header


brchristian:
I have noticed three(!) SQL calls related to `Spree::Store` on every single pageview and wondered what was going on...it seems excessive.

<img width="1055" alt="screen shot 2017-05-25 at 3 37 48 pm" src="https://cloud.githubusercontent.com/assets/2460418/26473491/8851ac44-4160-11e7-9744-9910e45823b0.png">

The culprit is [this line](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/current_store.rb#L26) from `current_store.rb`:

```ruby
def store_key
  @request.headers['HTTP_SPREE_STORE'] || @request.env['SERVER_NAME']
end
```

It seems that (since [this commit](https://github.com/solidusio/solidus/commit/e3be3024e135f1f5bd04ffd51150c925d0458c91) from two years ago), the _default_ method of checking the current store is by looking at the `HTTP_SPREE_STORE` response header.

However, looking far and wide I could find a total of zero documentation both in the Solidus codebase, and on the entire internet (!), about how to set or use the `HTTP_SPREE_STORE` header.

My feeling here is that we need to either (a) deprecate, or (b) document the use of this.

brchristian:
e.g., perhaps this can be extracted into https://github.com/solidusio/solidus_multi_domain?

kennyadsl:
@brchristian Hi there, which solidus version are you using? I think this is fixed with [this commit](https://github.com/solidusio/solidus/commit/a682147000c7797a5ca35178e5816026aaca80cf). 

When `@request.headers['HTTP_SPREE_STORE']` or `@request.env['SERVER_NAME']` are not set only the default store query (last one) should be done.

brchristian:
@kennyadsl I’m using the latest gem release, which is `2.2.1`.

I note that out of the box, the default 2.2.1 store makes _six_ `Spree::Store` find calls per pageview! (Three for the main layout, and three for the cart partial.)

My hunch is that only a small fraction of the Solidus user base have more than one Store. My personal opinion is that six SQL calls per pageview is a lot of DB chatter to support what is (IMO) a somewhat niche use-case.

What do you think about trying to move this functionality out into the `solidus_multi_domain` gem and support a single store by default in the vanilla Solidus gem?

brchristian:
[...]

PR #1993, opened 2017-06-03
Select current store using a standard Config preference

kennyadsl:
This is a first step in the attempt of reducing the queries made at each request in a default solidus store that doesn't use multiple stores, as for #1971 .

The goal here is to:

1. make it more standard/easier to use another class to determine which store to use (this PR)
2. move out of core [the logic that checks requests info](https://github.com/solidusio/solidus/blob/befa551eb8bc783df94647a757fe1628e3238d27/core/lib/spree/core/current_store.rb#L25-L27) to determine if another store has to be used. This will be moved to [solidus_multi_domain](https://github.com/solidusio/solidus_multi_domain).
3. just use the `Spree::Store.default` on a solidus store without solidus_multi_domain avoiding multiple queries for each request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Solidus 2.3.0 (master, unreleased)
 
+- Deprecate `Spree::Core::CurrentStore` in favor of `Spree::CurrentStoreSelector`. [\#1993](https://github.com/solidusio/solidus/pull/1993)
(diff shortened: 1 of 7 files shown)


mtomov:
That's just great! I always end up overriding `Store.current` to avoid the pesky lookups on every request:

```ruby
  # class memoize it as we only have one store
  def self.current(domain = nil)
    @current_store ||= Spree::Store.default
  end
```

mamhoff:
@kennyadsl would you mind rebasing this on the current master? The changelog changed again. :)

kennyadsl:
@mamhoff sure, done!

Issue #1996, opened 2017-06-06
Backend, problem with multiple "select2"


oeN:
### Steps to reproduce
- Add a filter with a select box (with the select2 class) in the products index
- Click on the "New Product" button

### Expected behavior
- filters panel is hidden
- Shipping Category and Tax Category are styled correctly with **select2**

### Actual behavior
- filters panel is still visible
- Shipping Category and Tax Category are not styled

### System configuration
```
ruby 2.4

gem 'solidus', github: 'solidusio/solidus', branch: 'master'
```

**Extensions in use**:
```
gem 'solidus_auth_devise'
gem 'solidus_i18n', github: 'solidusio-contrib/solidus_i18n', branch: 'master'
gem 'globalize', github: 'globalize/globalize'
gem 'solidus_social', github: 'solidusio-contrib/solidus_social'
gem 'solidus_editor', github: 'solidusio-contrib/solidus_editor', branch: 'master'
gem 'solidus_paypal_braintree', github: 'solidusio/solidus_paypal_braintree'
gem 'solidus_globalize', github: 'solidusio-contrib/solidus_globalize'
```

oeN:
The problem is related to https://github.com/select2/select2/issues/2750

We need to change the way solidus init the `select2()`. 
To solve this specific problem is enough to override the `backend/app/views/spree/admin/products/new.js.erb` and change `$('.select2').select2();` to `$('select.select2').select2();`

I am changing all the points where solidus uses `$('.select2').select2();` to `$('select.select2').select2();` and then I'll open a PR

jhawthorn:
Fixed by #1997

PR #1997, opened 2017-06-06
Use heavier selector for select2

oeN:
See Issue https://github.com/solidusio/solidus/issues/1996

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/address_states.js
+++ b/backend/app/assets/javascripts/spree/backend/address_states.js
@@ -1,7 +1,7 @@
 window.update_state = function (region, done) {
   'use strict';
 
-  var country = $('span#' + region + 'country .select2').select2('val');
+  var country = $('span#' + region + 'country select.select2').select2('val');
   var state_select = $('span#' + region + 'state select.select2');
   var state_input = $('span#' + region + 'state input.state_name');
 
--- a/backend/app/assets/javascripts/spree/backend/images/index.js.coffee
+++ b/backend/app/assets/javascripts/spree/backend/images/index.js.coffee
@@ -10,4 +10,4 @@ $ ->
       url: @href
       success: (r) ->
         ($ '#images').html r
-        ($ '.select2').select2()
+        ($ 'select.select2').select2()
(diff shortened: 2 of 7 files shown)


gmacdougall:
Looks like a reasonable change. We have a test failure here that looks related.

I'd also like to see a better commit description.

Thanks for looking in to this!

oeN:
Thank you! I'm looking to the failure right now.

oeN:
@gmacdougall I've tried but I was unable to replicate the error, it seems that the failure spec fails randomly (at least in my environment). Is it possible? Do you know if someone has a problem like this?

With ruby 2.4.0 I was unable to even run the spec because of `poltergeist`.
With ruby 2.3.0 I'm able to run the spec but it fails randomly most of times because of `poltergeist` with errors like:
- `Request to 'http://127.0.0.1:62594/admin' failed to reach server, check DNS and/or server status`
- `Timed out with the following resources still waiting data:image/png;base64,iVBOR.....`

And sometimes the spec pass without problems. Do you have any advice to solve this and get the PR merged?

Thank you!!

graygilmore:
Should we be using an ID or `name`? That's what https://github.com/select2/select2/issues/2750#issuecomment-59634990 suggests.

Should allow us to clean up these selectors even more.

oeN:
Yes, I have opted for the `name` selector `select` instead of the ID to solve this problem quickly and with the minimum amount of code.

frankie-loves-jesus:
Wouldn't it be better if we switched to https://github.com/selectize/selectize.js altogether?

jhawthorn:
There are some good suggestions here, but I think they're all outside the scope of fixing this immediate issue. Many extensions use `<select class="select2">`, so we'll need that to continue working for the time-being.

I'd love to see us move away from that, to bootstrap's `custom-select` or similar, but this is a good fix to an immediate issue.

graygilmore:
> There are some good suggestions here, but I think they're all outside the scope of fixing this immediate issue. Many extensions use <select class="select2">, so we'll need that to continue working for the time-being.

👍 It wasn't fully clear to me why this was necessary so I pulled stuff down and tested things out. I think improving the commit message here will go a long way in clarifying things for future-us.

oeN:
@tvdeyen I've tried to add the information on the commit, let me know if is fine. I thought I had to stay in the character limit 😄

tvdeyen:
> I thought I had to stay in the character limit

This character only applies to the first sentence. The description body (separated by a new line) can and should be as long as you need to explain the rational behind our changes. 

A very good read on this topic is: https://chris.beams.io/posts/git-commit/

PR #1998, opened 2017-06-06
Add check for eligibility to free shipping promotion handler

nhippenmeyer:
### Steps to reproduce
1. Create a free shipping promotion and add a rule for item total to be greater than or equal to a certain amount.
2. Check out with an order that doesn't satisfy the minimum item total.

### Expected behavior
The free shipping promotion should not be applied to the order.

### Actual behavior
The free shipping promotion is applied to the order even when the rule isn't satisfied.

### System configuration
**Solidus Version**: 2.2.1

### DIFF
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -85,6 +85,7 @@ module Spree
     end
 
     def activate(order:, line_item: nil, user: nil, path: nil, promotion_code: nil)
+      return unless eligible?(order)
       return unless self.class.order_activatable?(order)
 
       payload = {
--- a/core/spec/models/spree/promotion_handler/free_shipping_spec.rb
+++ b/core/spec/models/spree/promotion_handler/free_shipping_spec.rb
@@ -3,8 +3,8 @@ require 'spec_helper'
 module Spree
   module PromotionHandler
     describe FreeShipping, type: :model do
-      let(:order) { create(:order) }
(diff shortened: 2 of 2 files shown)


nhippenmeyer:
This is causing the order_updater spec which distinguishes between adjustments and eligible adjustments to fail.  I'm unclear on why this distinction exists in the first place.

gmacdougall:
Looks like we have a legitimate test failure here. We'd also be looking for a better PR/commit description along with this one.

Thanks for taking a look at this!

nhippenmeyer:
@gmacdougall I updated the PR description with steps to reproduce this bug.  I'm not sure what to do about the failing test here, should the free shipping adjustment show up as an adjustment on the order if it's not being applied because the rules aren't satisfied?  My suspicion is that it should not in which case I can update the `order_updater_spec.rb`

danmitchell-:
This is high priority issue IMO. I hope it can be resolved soon 👍 

Just to reiterate, at present if you have a free shipping promo set up then all orders are getting Free Shipping regardless of their value

danmitchell-:
I now realise there is a more recent and up to date PR https://github.com/solidusio/solidus/pull/2187 so fingers crossed it goes into the next release or as a patch for 2.3

kennyadsl:
Closing this PR since there's a new up to date PR (#2187) which seems to be close to a final solution.

PR #2000, opened 2017-06-07
Rename bogus gateways

tvdeyen:
To remove confusion around payment methods, gateways and providers we rename the `Spree::Gateway::Bogus` and `Spree::Gateway::BogusSimple` payment methods into `Spree::PaymentMethod::BogusCreditCard` and `Spree::PaymentMethod::SimpleBogusCreditCard`.


gmacdougall:
Should we provide something like:

```Ruby
Spree::Gateway::Bogus = Spree::PaymentMethod::BogusCreditCard 
```

or even

```Ruby
class Spree::Gateway::Bogus < Spree::PaymentMethod::BogusCreditCard
  def initialize(*args)
    #deprecation notice
    super
  end
end
```

until we're ready for a major version release?

tvdeyen:
@gmacdougall I did this https://github.com/solidusio/solidus/pull/2000/commits/9443aec724f3a9ffd05a8ff1838db0c4a697fc97 ;)

gmacdougall:
That'll learn me for reading PR's too fast!

tvdeyen:
No worries, this is a large changeset

tvdeyen:
@mamhoff made a great comment IRL about making the migration class available from the outside (eg. `solidus_gateway`). Will do that now.

tvdeyen:
@mamhoff done!

core/lib/solidus/migrations/rename_gateways.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Solidus
+  module Migrations
+    class RenameGateways
+      DEFAULT_MAPPING = [
+        ['Spree::Gateway::Bogus', 'Spree::PaymentMethod::BogusCreditCard'],
```
kennyadsl: why not use an hash here?

core/spec/models/spree/payment_method/bogus_credit_card_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'spec_helper'
+
+module Spree
+  describe PaymentMethod::BogusCreditCard, type: :model do
+    let(:bogus) { create(:credit_card_payment_method) }
+    let!(:cc) { create(:credit_card, payment_method: bogus, gateway_customer_profile_id: "BGS-RERTERT") }
```
kennyadsl: why this file is needed?

tvdeyen: I just renamed the file.

core/lib/solidus/migrations/rename_gateways.rb
tvdeyen: A good catch, as the keys are unique this will work as well. Thanks!

core/spec/models/spree/payment_method/bogus_credit_card_spec.rb
tvdeyen: I say let's create a separate PR

core/lib/solidus/migrations/rename_gateways.rb
tvdeyen: Changed to Hash, thanks!

PR #2001, opened 2017-06-07
Rename Gateway into PaymentMethod::CreditCard

tvdeyen:
The `Spree::Gateway` is an implementation of a credit card payment method. The current name is confusing.

With moving this class into the `PaymentMethod` namespace and changing its name to `CreditCard` it is now more obvious what this class actually is.

Several hints inside the class prove that:

- It inherits from `Spree::PaymentMethod`
- The `source_class` is `Spree::CreditCard`
- In `supports?` it asks the gateway if it supports a certain credit card brand

Further actions are required. First on foremost `solidus_gateway` needs to be changed in a way that all credit card based payment methods inherit from `Spree::PaymentMethod::CreditCard` from now on and all non credit card payment methods inherit from `Spree::PaymentMethod` directly.

includes #2000


core/app/models/spree/payment_method.rb
```diff
@@ -1,5 +1,13 @@
...
-  # An abstract class which is implemented most commonly as a `Spree::Gateway`.
+  # A base class which is used for implementing payment methods.
+  #
+  # See https://github.com/solidusio/solidus_gateway/ for
+  # offically supported payment method implementations.
+  #
+  # Uses STI (single table inheritance) to store all implemented payment methods
+  # in one table (+spree_payment_methods+).
+  #
+  # This class is not ment to be instantiated. Please create instances of concrete payment methods.
```
mamhoff: This is excellent class documentation. Wow. 

There's a typo in here though: "ment" should read "meant".

core/app/models/spree/payment_method.rb
```diff
@@ -116,6 +131,22 @@ def self.active?
...
 
+    # Used as partial name for your payment method
+    #
+    # Currently your payment method needs to provide these partials:
+    #
+    #     1. app/views/spree/checkout/payment/_{method_type}.html.erb
+    #     The form your customer enters the payment information in during checkout
+    #
+    #     2. app/views/spree/checkout/existing_payment/_{method_type}.html.erb
+    #     The payment information of your customers resuable sources during checkout
```
mamhoff: Typo: "resuable" -> "reusable"

tvdeyen:
@mamhoff thanks! Fixed

Also now includes #2000

mtomov:
Indeed! I very much admire your guts to make such a major change in the name of making it all easier for us to work with the platform! I guess a big role in having the courage to do the refactoring play the tests, so thanks all as well for writing tests so good!

PR #1926, opened 2017-05-22
Add regression test for refund outstanding_balance

jhawthorn:

PR #2002, opened 2017-06-07
Update outstanding_balance to use reimbursements

jhawthorn:
This updates `outstanding_balance` to use the value of "reimbursements" when determining the payment total.

Previously the value of an order's refunds was used both for representing the money returned to the customer as well as the reduced "adjusted total" of the order (which we expected to be paid). This was
wrong in a number of cases.

This updates `outstanding_balance` to use the value of reimbursements for calculating the "adjusted total" of the order. This allows all cases to (at least) be represented properly. This better
matches the best practice double-entry bookkeeping style.

Reimbursements are created when going through the refund system so this should continue to work for cases where refunds where made through the returns system.

This PR includes [a test](63fe5fcca6385968a220f093141e499cdeb6823d) from @ericsaupe (showing breakage of the existing `outstanding_balance` as well as my #1578 integration tests (with all :pending specs now passing!) and #1926 regression test (already merged in master).

Thanks to everyone who had continued to report, prod and provide tests for this issue (too many to list but I have CC'd some of you fine folks). This has been a doozy, and I hope this is a viable solution.

CC @ericsaupe @skukx @DanielePalombo @jasonfb @bbuchalter @jordan-brough @mamhoff @stewart @luukveenis @Senjai 

If any of you can verify that this meets your needs, or observe it's effects on production-ish data that would be amazing.

Not addressed here is a data migration to update the `payment_state` of existing orders. This will require a detailed CHANGELOG entry.

### DIFF
--- a/core/spec/models/spree/order/outstanding_balance_integration_spec.rb
+++ b/core/spec/models/spree/order/outstanding_balance_integration_spec.rb
@@ -1,0 +1,116 @@
+require 'spec_helper'
+
+# This method in particular has been difficult to get right.
+# Many things will affect this amount
+#
+# See also:
+# https://github.com/solidusio/solidus/issues/1254
+# https://github.com/spree/spree/issues/6229
+# https://github.com/solidusio/solidus/issues/1107
+# https://github.com/solidusio/solidus/pull/1557
+# https://github.com/solidusio/solidus/pull/1536
+
+RSpec.describe "Outstanding balance integration tests" do
+  let!(:order) { create(:order_with_line_items, line_items_count: 2, line_items_price: 3, shipment_cost: 13) }
+  before { order.update_attributes!(state: 'complete', completed_at: Time.now) }
+
+  subject do
+    order.reload
+    order.update!
+    order.outstanding_balance
+  end
+
+  context 'when the order is unpaid' do
+    it { should eq order.total }
+    it { should eq 19 }
+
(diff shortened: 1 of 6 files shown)


core/spec/models/spree/order/outstanding_balance_integration_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        order.reload
+      end
+
+      it 'discounts the cancelled item amount' do
+        expect(order.refund_total).to eq(3)
+        expect(order.payment_total).to eq(16)
+        expect(order.outstanding_balance).to eq(0)
+
+        # Less sure about this one....
```
jordan-brough: Did you intend to leave this comment in? `19` seems OK to me (slightly confusing but does make sense from a certain POV)

core/spec/models/spree/order/outstanding_balance_integration_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        expect(order.total).to eq(19)
+      end
+
+      context 'and there is a full refund' do
+        it 'has the correct amount'
```
jordan-brough: Did you intend to implement this spec?

jordan-brough:
Link to https://github.com/solidusio/solidus/issues/1254 (since this fixes it)

PR #2003, opened 2017-06-08
Specify countries that don’t require states

isaacfreeman:
Some countries have subregions recorded in Carmen data, but people typically don’t use them in postal addresses. This adds an array of such countries to check before we assume that subregions are required.

For New Zealand, Carmen data includes provinces, and groups them by island, so Solidus checkout requires customers to choose "North Island", "South Island" or "Chatham Islands". This is not customary practice and is specifically recommended against in New Zealand Post's address standards (https://www.nzpost.co.nz/sites/default/files/uploads/shared/standards-guides/adv358-address-standards.pdf).

No doubt there are other countries where a similar situation applies, and which could be included in this pull request.

### DIFF
--- a/core/db/default/spree/countries.rb
+++ b/core/db/default/spree/countries.rb
@@ -1,5 +1,14 @@
 require 'carmen'
 
+# These countries have subregions, but typically don't use them in postal
+# addresses.
+COUNTRIES_THAT_DONT_REQUIRE_STATES = %w(NZ)
+
+def require_states?(country)
+  return false if COUNTRIES_THAT_DONT_REQUIRE_STATES.include? country.alpha_2_code
+  country.subregions?
+end
+
 ActiveRecord::Base.transaction do
   Carmen::Country.all.each do |country|
     Spree::Country.where(iso: country.alpha_2_code).first_or_create!(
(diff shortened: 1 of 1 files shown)


jhawthorn:
Thanks Isaac. I wonder if there's some repository we can find this information for all countries. We've been trying to avoid having any information about countries in Solidus itself because it tends to become out of date or incorrect.

isaacfreeman:
@jhawthorn: Yeah, I was in two minds about whether to submit this as an issue or a PR, but I figured we might as well start with some code.

I'll have a look for such a repository. My suspicion is that there's more likely to be something that records full address formats in all their glorious complexity than something that focuses on the specific question of whether to include states or not. But we'll see.

I'm happy to either leave this open for discussion, or close it and open an issue that links to it as one possible implementation we're not currently pursuing.

mamhoff:
The most complete repository for locale data is by far the [CLDR](https://cldr.unicode.org). There might very well be a flag hidden in there that we could use for this. It would require some digging though. The good news is we already have it as a dependency (and use it only for validating postal codes at the moment).

isaacfreeman:
Thanks @mamhoff. I'll see what I can find there, and close this.

PR #2004, opened 2017-06-08
Update RuboCop config

tvdeyen:
After we merged 2c775ab24e592b3949de417712017c1adb4de3f9 we decided to enable 
the cop for `Rails/DynamicFindBy`.

Along this I updated outdated configuration syntax and disabled/relaxed some
cops.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -106,6 +106,17 @@ Rails/DynamicFindBy:
     - find_by_param
     - find_by_param!
 
+# We use a lot of
+#
+#     expect {
+#       something
+#     }.to { happen }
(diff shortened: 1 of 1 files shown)


tvdeyen:
> What about also grouping config rules by their namespace?

Thought about that as well, but hesitated because the relaxed ruby styles config is on the bottom and our own additions at the top, but we could change that though.

kennyadsl:
Maybe we can leave relaxed ruby styles config at the bottom (I agree it's better to keep them separate to allow easy replacement if needed) and group only our rules? Does it make sense?

PR #2006, opened 2017-06-08
Fix spurious failures in order_factory_spec

jhawthorn:
It's valid for the DB to return the inventory units in any order, resulting in some spurious failures in this test.


bbuchalter:
🙏 thank you for cleaning this up

jhawthorn:
Merging without our normal 24 hour review window since this could avoid spec failures in other PRs

PR #2007, opened 2017-06-08
Specify Rails 5.0

brchristian:
Rails 5.1 is currently incompatible with any gem release of Solidus, so I think it makes sense to specify Rails 5.0.x in the guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Getting started
 
 Begin by making sure you have [Imagemagick](http://imagemagick.org/script/download.php) installed, which is required for Paperclip. (You can install it using [Homebrew](https://brew.sh) if you're on a Mac.)
 
-To add solidus, begin with a Rails 5 application. Add the following to your
+To add solidus, begin with a Rails 5.0 application. Add the following to your
 Gemfile.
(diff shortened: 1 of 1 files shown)


jhawthorn:
Master now runs Rails 5.1, but we might need a better description of which versions support which versions of rails in the README

brchristian:
@jhawthorn Oh I actually didn’t know that about the `master` branch. But that said, the README does specify using the rubygems version of Solidus, so that’s why I was suggesting we say Rails 5.0, just so people don’t get tripped up. The alternative, which would also work, is to suggest they use the `master` github version of Solidus (rather than the rubygems release).

jhawthorn:
It would be nice to have a section describing what Solidus versions support which rails versions (4.2 for 1.x, 5.0 for 2.0-2.2, 5.1 for 2.3+), but since master is Rails 5.1, this is soon to be invalid

asecondwill:
I'm not quote sure which to use still though.  If I want to get a stable (2.4.2) version running smoothly with quite a few extensions, which version of rails should i go back to? 5.0?

PR #2010, opened 2017-06-08
Avoid initializing git repo in dummy app

jhawthorn:
Rails 5.1 now automatically makes a git repo for you. Neat! but not what we want when running `rake test_app`.

This also skips a number of other unnecessary steps in the dummy app generation.

### DIFF
--- a/core/lib/generators/spree/dummy/dummy_generator.rb
+++ b/core/lib/generators/spree/dummy/dummy_generator.rb
@@ -30,6 +30,15 @@ module Spree
       opts[:database] = 'sqlite3' if opts[:database].blank?
       opts[:force] = true
       opts[:skip_bundle] = true
+      opts[:skip_gemfile] = true
+      opts[:skip_git] = true
+      opts[:skip_keeps] = true
+      opts[:skip_listen] = true
+      opts[:skip_puma] = true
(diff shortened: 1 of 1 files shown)


core/lib/generators/spree/dummy/dummy_generator.rb
```diff
@@ -30,6 +30,13 @@ def generate_test_dummy
       opts[:database] = 'sqlite3' if opts[:database].blank?
       opts[:force] = true
       opts[:skip_bundle] = true
+      opts[:skip_gemfile] = true
+      opts[:skip_test] = true
+      opts[:skip_git] = true
+      opts[:skip_keeps] = true
+      opts[:skip_spring] = true
+      opts[:skip_listen] = true
+      opts[:skip_puma] = true
```
kennyadsl: looking at [here](https://github.com/rails/rails/blob/5-1-stable/railties/lib/rails/generators/app_base.rb) maybe we can also add `skip_yarn` and `no_rc `?

jhawthorn: :+1: done

mtomov:
Thanks for this one as well!

PR #2012, opened 2017-06-09
Admin::ReportsController reusable search params

oeN:
I'm adding a new report to my project, and I want to reuse the same filter as the `sales_total`.
I think that would be great if we can have the search params logic in an another method like `search_params`

### DIFF
--- a/backend/app/controllers/spree/admin/reports_controller.rb
+++ b/backend/app/controllers/spree/admin/reports_controller.rb
@@ -26,27 +26,7 @@ module Spree
       end
 
       def sales_total
-        params[:q] = {} unless params[:q]
-
-        if params[:q][:completed_at_gt].blank?
-          params[:q][:completed_at_gt] = Time.current.beginning_of_month
-        else
-          params[:q][:completed_at_gt] = begin
-                                           Time.zone.parse(params[:q][:completed_at_gt]).beginning_of_day
-                                         rescue
-                                           Time.current.beginning_of_month
-                                         end
-        end
-
-        if params[:q] && !params[:q][:completed_at_lt].blank?
-          params[:q][:completed_at_lt] = begin
-                                           Time.zone.parse(params[:q][:completed_at_lt]).end_of_day
-                                         rescue
-                                           ""
-                                         end
-        end
-
-        params[:q][:s] ||= "completed_at desc"
+        params[:q] = search_params
 
         @search = Order.complete.ransack(params[:q])
         @orders = @search.result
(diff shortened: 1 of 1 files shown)


backend/app/controllers/spree/admin/reports_controller.rb
```diff
@@ -68,6 +48,29 @@ def sales_total
...
+          q[:completed_at_gt] = adjust_start_date q[:completed_at_gt]
+          q[:completed_at_lt] = adjust_end_date(q[:completed_at_lt]) if q[:completed_at_lt].present?
+          q[:s] ||= 'completed_at desc'
+        end
+      end
+
+      def adjust_start_date(string_date = nil)
+        return Time.current.beginning_of_month unless string_date
+        Time.zone.parse(string_date).beginning_of_day
+      rescue
```
jhawthorn: Can this rescue something more specific? It's easy for bugs to sneak in with the catchall rescue.

backend/app/controllers/spree/admin/reports_controller.rb
```diff
@@ -68,6 +48,29 @@ def sales_total
...
+        Time.current.beginning_of_month
+      end
+
+      def adjust_end_date(string_date)
+        Time.zone.parse(string_date).end_of_day
+      rescue
```
jhawthorn: ditto here.

backend/app/controllers/spree/admin/reports_controller.rb
```diff
@@ -68,6 +48,29 @@ def sales_total
...
+      def search_params
+        params.fetch(:q, {}).tap do |q|
+          q[:completed_at_gt] = adjust_start_date q[:completed_at_gt]
+          q[:completed_at_lt] = adjust_end_date(q[:completed_at_lt]) if q[:completed_at_lt].present?
+          q[:s] ||= 'completed_at desc'
+        end
+      end
+
+      def adjust_start_date(string_date = nil)
+        return Time.current.beginning_of_month unless string_date
```
jhawthorn: The previous code checked the equivalent of `string_date.blank?`, should we still be doing that (I suspect yes since we may get an empty string from the user)?

kennyadsl:
@oeN also, I think that if you rebase against master the test suite failure will disappear, thanks!

oeN:
@kennyadsl Thanks for the comments! About the "unparsable" format, I think that we can avoid this check, for now, for two reasons:

- the inputs are managed by JS code
- adding this check feels like a kind of defensive programming

PR #2017, opened 2017-06-13
Emit Spree.url deprecation warning in all envs

jhawthorn:
This has been deprecated forever, but I'm still not confident enough to remove it. It's possible that it hasn't been seen by users because `Spree.env` has been unset (as it is in our default frontend).

This commit changes the warning to be issued always (so long as console and console.warn are defined).

This also removes a call to `Spree.url` from solidus_frontend.

### DIFF
--- a/core/app/assets/javascripts/spree.js.erb
+++ b/core/app/assets/javascripts/spree.js.erb
@@ -24,7 +24,7 @@ Spree.pathFor = function(path) {
 
 Spree.url = function(uri, query) {
   if (console && console.warn) {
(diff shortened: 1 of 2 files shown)


jhawthorn:
> We should maybe also change the warning message. Often Spree.url is used only to get a URL, like in the example above. So, adding "use Spree.routes to get URLs" could help developers understand what is wrong.

I don't think `Spree.routes` provides the functionality `Spree.url` has either. I think I will just drop the suggestion from the warning.

PR #2019, opened 2017-06-14
Remove ! from assign_default_user_addresses!

jhawthorn:
There was some confusion when reviewing #1967 about whether `assign_default_user_addresses!` would persist changes or not. It doesn't, but the `!` suggests that it might.

There's no version of this method without an exclamation mark. Additionally, this doesn't have any of the normal tells of a ! method, it doesn't make changes to the database, shouldn't raise exceptions, and
doesn't do anything "dangerous".

This commit renames the method to `assign_default_user_addresses`, and keeps `assign_default_user_addresses!` as a deprecated alias. It also adds some YARD docs to the method.

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -698,6 +698,9 @@ module Spree
       self.ship_address = Spree::Address.immutable_merge(ship_address, attributes)
     end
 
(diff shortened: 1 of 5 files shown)


mtomov:
Thanks a lot!

PR #2020, opened 2017-06-14
Validate presence of product on a Variant

### LINKED ISSUE #2011: Variant validation allows no product associated
### Steps to reproduce
Just follow [this](https://asciinema.org/a/ck2bktzpcdpapoij6h0ztasr1) asciinema

### Expected behavior
Variants must **always** have an associated product to be valid

### Actual behavior
If I set `variant.product` as `nil` doing `variant.save` the validation doesn't fail.
So variants can be product orphan.

### System configuration
Solidus master branch

jhawthorn:
It might also make sense to add a DB constraint.

Fixes #2011

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -20,7 +20,7 @@ module Spree
     attr_writer :rebuild_vat_prices
     include Spree::DefaultPrice
 
-    belongs_to :product, -> { with_deleted }, touch: true, class_name: 'Spree::Product', inverse_of: :variants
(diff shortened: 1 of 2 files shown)


core/app/models/spree/variant.rb
```diff
@@ -61,6 +61,7 @@ class Variant < Spree::Base
 
     validate :check_price
 
+    validates :product, presence: true
```
tvdeyen: Can't we instead use `optional: false` on the `belongs_to :product` above?

PR #2022, opened 2017-06-15
[WIP] Always use the default store

kennyadsl:
This commit avoids doing 2 extra queries for every request (#1971)
even when solidus store is not using solidus_multi_domain extension.

This logic has now been moved to the solidus_multi_domain extension
since it is needed only when multiple stores are expected
(solidusio/solidus_multi_domain#67).

*NOTE:* there's something I'm not too happy about this PR. I'm moving out to solidus_multi_domain extension a spec about filtering api orders by store. But the [code that filters api orders at the moment is still in core](https://github.com/solidusio/solidus/blob/befa551eb8bc783df94647a757fe1628e3238d27/api/app/controllers/spree/api/orders_controller.rb#L75). 
Since the new version of the "store selector class" always returns the default store, I'm not sure it makes sense to test that logic in core, even if the code is here. What about moving the controller code into the extension as it is done for [shipments](https://github.com/nebulab/solidus_multi_domain/blob/178a4b5a63fc139614bfffd1589f0057d17b085b/app/controllers/spree/api/shipments_controller_decorator.rb#L4)?

### DIFF
--- a/api/spec/controllers/spree/api/orders_controller_spec.rb
+++ b/api/spec/controllers/spree/api/orders_controller_spec.rb
@@ -128,27 +128,32 @@ module Spree
       end
     end
 
-    it "cannot view all orders" do
-      api_get :index
-      assert_unauthorized!
-    end
+    context 'displaying orders' do
+      let(:current_api_user) { order.user }
+      let(:order) { create(:order, line_items: [line_item], store: Spree::Store.default) }
 
-    context "the current api user does not exist" do
-      let(:current_api_user) { nil }
+      it 'is not authorized to view an order of another user' do
+        another_user_order = create(:order, store: Spree::Store.default)
+        api_get :show, id: another_user_order.to_param
 
-      it "returns a 401" do
-        api_get :mine
         expect(response.status).to eq(401)
       end
-    end
 
-    context "the current api user is authenticated" do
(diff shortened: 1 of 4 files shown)


mtomov:
Thanks!

I haven't used more than 1 store, but if it is a requirement to use the `solidus_multi_domain` extension to work with more than one store, then I guess the logic should belong in the extension. Then again, I'm not too familiar if core is meant to support multiple stores the extension.

jhawthorn:
I think we've been moving more towards having the multi_domain behaviour included in core, just because of how frequent a need it is (and how invasive the changes are otherwise). But I appreciate how wasteful those three queries are on single-store apps.

Maybe we want to include several store selectors side-by-side `Spree::StoreSelector::Default`/`Spree::StoreSelector::ByDomain` for users to select. WDYT?

tvdeyen:
I like the idea of provided selectable Store selector classes. 

Why do we even still have this extension? Shouldn't we deprecate it?

kennyadsl:
Since the direction is to gradually deprecate `solidus_multi_domain`, I'm going to change this PR trying to pull stuff from that extension into core instead of the opposite. I'll update this soon!

kennyadsl:
Closing this one in favor of #2022 which is a better solution.

PR #2025, opened 2017-06-15
Fix admin SQL issues with DISTINCT products

### LINKED ISSUE #2016: Order clause expression not in select list on MySQL 5.7+
### Steps to reproduce
Using MySQL 5.7+ with default configuration
Try to order the products by price on the admin

### Expected behavior
It should order the products by price

### Actual behavior
500 error.
Example of the error:
```
Mysql2::Error: Expression #1 of ORDER BY clause is not in SELECT list, references column 'vff_production.spree_prices.amount' which is not in SELECT list; this is incompatible with DISTINCT: 

SELECT DISTINCT `spree_products`.* 
FROM `spree_products` 
LEFT OUTER JOIN `spree_variants` 
ON `spree_variants`.`product_id` = `spree_products`.`id` 
AND `spree_variants`.`is_master` = 1 
AND `spree_variants`.`deleted_at` IS NULL 
LEFT OUTER JOIN `spree_prices` 
ON `spree_prices`.`variant_id` = `spree_variants`.`id` 
AND `spree_prices`.`currency` = 'BRL' 
AND `spree_prices`.`country_iso` IS NULL 
AND `spree_prices`.`deleted_at` IS NULL 
WHERE `spree_products`.`deleted_at` IS NULL AND (1=1) 
ORDER BY `spree_prices`.`amount` ASC 
LIMIT 10 OFFSET 0
```

### System configuration
**Solidus Version**:
2.2.0
**Extensions in use**:
solidus_auth_devise
solidus_prototypes

Related functionality on MySQL 5.7+: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_only_full_group_by
Note that this is the default SQL mode now. This is also what other DBs already do, so I think it makes sense to fix this.

jhawthorn:
Fixes #2016

MySQL 5.7+ previously errored when trying to sort products by price.

> ORDER BY clause is not in SELECT list,
> references column 'spree_prices.amount' which is not in SELECT list;
> this is incompatible with DISTINCT

Additionally, under PostgreSQL, searching for an SKU which matches multiple variants of the same product and then sorting by price, multiple of the same Product would be returned in the search.

This commit removes the Product.distinct_by_product_ids hack replaces the join search on products->variants with a WHERE EXISTS subquery.

### DIFF
--- a/backend/app/controllers/spree/admin/products_controller.rb
+++ b/backend/app/controllers/spree/admin/products_controller.rb
@@ -102,7 +102,6 @@ module Spree
         # @search needs to be defined as this is passed to search_form_for
         @search = super.ransack(params[:q])
         @collection = @search.result.
-              distinct_by_product_ids(params[:q][:s]).
               includes(product_includes).
               page(params[:page]).
               per(Spree::Config[:admin_products_per_page])
--- a/backend/app/views/spree/admin/products/index.html.erb
+++ b/backend/app/views/spree/admin/products/index.html.erb
@@ -26,8 +26,8 @@
 
           <div class="col-4">
             <div class="field">
-              <%= f.label :variants_including_master_sku_cont, Spree::Variant.human_attribute_name(:sku) %>
-              <%= f.text_field :variants_including_master_sku_cont, size: 15 %>
+              <%= f.label :with_variant_sku_cont, Spree::Variant.human_attribute_name(:sku) %>
+              <%= f.text_field :with_variant_sku_cont, size: 15 %>
             </div>
           </div>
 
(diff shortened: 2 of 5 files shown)


mtomov:
Woow, that's so elegant! and builds on top of the refactor in #1904 by @spaghetticode setting an excellent example for any such scopes. 

I've actually had to deal with this one just recently to handle searching through thousands of products, and I didn't know about the exists way .. will definitely use it next time around!


Maybe it's a good time to ping on #1660. There was an excellent discussion there, and in my view we, the benefits of lookups and search working on large stores far outweight the usability negatives of not being able to do partial matching. And by large stores I don't even mean ours (4.5mil variants, 3mil) - I'm sure that even with a few thousand products, each with 8-10 variants .. the lookups become hardly usable. 

So if you do make a decision for better support of large stores, then we can add PRs similar to #1660 as well as conditionally add Postgres (or MySQL) specific indices such as [Operators](https://www.postgresql.org/docs/current/static/indexes-opclass.html), which would reduce those lookup times to milliseconds.

jhawthorn:
@mamhoff I have re-added the scope and deprecated it instead.

PR #2034, opened 2017-06-21
Replace select2 with <select class="custom-select">

jhawthorn:
This replaces the select2 component with a plain `<select>` (with bootstrap's custom-select styling) throughout the admin.


mtomov:
Hm.. Ain't it a more valid approach to style all `select` with your custom styles, but maybe add a `no-custom-style` or so class for those occasions when you want to opt out of it. That should make it easier for developers as they won't have to learn about this new class unless they have a very very special case to "opt out". Would certainly help extensions as well.

jhawthorn:
`custom-select` has styling from bootstrap. It is somewhat heavy handed, so we probably don't want to `@extend` it onto every select element.

PR #2035, opened 2017-06-21
Fix typo on the new promotions form

swcraig:
The hint for the expiration date field had an extra "s" at the end of
the sentence. The hint used to read, "If no value is specified, the
promotion will never expires."

This commit removes the extra "s" in "expires".


tvdeyen:
Thanks.

mtomov:
Thanks!

PR #2036, opened 2017-06-22
[WIP] Add foreign keys

jhawthorn:


mamhoff:
I've opened an issue linking to this PR.

jhawthorn:
This was not merged because in tests on older long lived stores there was a lot of broken data.

I still think this should be merged and it should be the responsibility of users to fix up their data (or skip the foreign keys migration).

I will not be able to get around to updating this.

Issue #2032, opened 2017-06-20
Implement a payment capture events GUI


tvdeyen:
Following these discussions

- [Spree PR](https://github.com/spree/spree/issues/2962#issuecomment-43803337)
- [Spree commit introducing capture events](https://github.com/spree/spree/commit/486f0dc078bb917c4e17d2a028749af92b2d9832#commitcomment-5325335)
- [Slack discussion](https://solidusio.slack.com/archives/C0HGMA1QF/p1497429695729061)

we are missing an interface for capturing partial payments (`Spree::PaymentCaptureEvent`)

So, @moritzps, @pascalj and @rbjoern84 from @bitspire started a discussion on Slack about implementing one.

This is the original screenshot of the UI they proposing 

![solidus_multiple_captures_v2](https://user-images.githubusercontent.com/42868/27320719-babe95ac-5597-11e7-8907-629d8d5bfe2d.png)

### TODO

- [ ] Display the form above/below the payment capture events table
- [ ] Display the form under the payment details if no capture events are present yet
- [ ] Prefill remaining amount in the field (payment amount minus sum of all captured events)
- [ ] Do not display form if no money can be captured anymore (the full amount of the payment is reached)

tvdeyen:
One thing I discussed with @pascalj IRL was to add a from for capturing partial payments to the payments detail page. Introducing a modal (we did not finally decided to have in our UI now) for this rarely used feature seems overkill for us. 

#### Payment Detail page with capture events

![payments - r987654321 - orders](https://user-images.githubusercontent.com/42868/27321309-0b2c953c-559a-11e7-8cc6-20170a29b422.jpg)

We need to keep in mind that payments could be uncaptured and therefor do not have an `Spree::PaymentCaptureEvent` yet. But we still have the payments detail page, so we could still display the form and pre-fill the amount into the field

#### Payment Detail page without capture events

![credit card pending - payments - r123456789 - orders 2017-06-20 09-23-57](https://user-images.githubusercontent.com/42868/27321427-78cdadba-559a-11e7-8bf2-6086343d67d1.jpg)

The form **should not be displayed** if no many can be captured anymore (aka. the full amount of the payment is reached)

pascalj:
[...]

PR #2039, opened 2017-06-23
[WIP] partial captures

pascalj:
Please see #2032 for the details.

This is currently WIP because I'd like to get some feedback regarding the overall behaviour.

### DIFF
--- a/backend/app/controllers/spree/admin/payments_controller.rb
+++ b/backend/app/controllers/spree/admin/payments_controller.rb
@@ -53,9 +53,7 @@ module Spree
       def fire
         return unless (event = params[:e]) && @payment.payment_source
 
-        # Because we have a transition method also called void, we do this to avoid conflicts.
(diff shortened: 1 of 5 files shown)


cbrunsdon:
Hi @pascalj, sorry about the slow reply, I've been internally going back and forth on this and its generated a few pretty hefty conversations.

I definitely think the change is a good direction (multiple captures when payment gateways support it), but I'm worried that this is going to be a big change that everyone will get, even when multiple capture is not supported by payment gateways.

I do however think this is a very reasonable, and not too intrusive for an extension, would you consider supporting it in that way? If it lives there for a while and the kinks get worked out, I would be happy to re-examine it for core in a future release.

pascalj:
Thanks @cbrunsdon! I responded in #2032 to keep the discussion there. Closing this MR in favor of an extension.

PR #2040, opened 2017-06-23
[RFC] Introduce preference fields partials

tvdeyen:
Preference fields are used in several places all over the Solidus admin. Some of them are:

- Shipping and tax rate calculator fields
- Payment method preferences
- Promotion action calculator fields

Instead of using a string concatinator helper - that just concatinate the fields and labels with `<br>`
- we introduce partials that can easily be overriden by stores and extensions.

This deprecates the follwoing preference fields helpers:

- `preference_field_tag`
- `preference_field_for`
- `preference_field_options`
- `preference_fields`

### Before

preference fields - promotion action - before
preference fields - payment method - before
preference fields - shipping rate - before

### After

preference fields - promotion action - after
preference fields - payment method - after
preference fields - shipping calculator - after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Solidus 2.4.0 (master, unreleased)
 
+- Deprecates several preference fields helpers in favor of preference field partials. [\#2040](https://github.com/solidusio/solidus/pull/2040) ([tvdeyen](https://github.com/tvdeyen))
+  Please render `spree/admin/shared/preference_fields/\#{preference_type}' instead
+
 - Remove `set_current_order` calls in `Spree::Core::ControllerHelpers::Order`
   [\#2185](https://github.com/solidusio/solidus/pull/2185) ([Murph33](https://github.com/murph33))
 
--- a/backend/app/views/spree/admin/shared/preference_fields/_boolean.html.erb
+++ b/backend/app/views/spree/admin/shared/preference_fields/_boolean.html.erb
@@ -1,0 +1,13 @@
+<% label = local_assigns[:label].presence %>
+<% html_options = local_assigns[:html_options] || {} %>
+
+<div class="field">
+  <% if local_assigns[:form] %>
+    <%= form.check_box attribute, html_options %>
+    <%= form.label attribute, label %>
+  <% else %>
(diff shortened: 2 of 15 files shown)


backend/app/views/spree/admin/shared/preference_fields/_boolean.html.erb
```diff
@@ -0,0 +1,13 @@
+<% label = local_assigns[:label].presence %>
+<% html_options = local_assigns[:html_options] || {} %>
+
+<div class="field">
+  <% if local_assigns[:form] %>
+    <%= form.check_box attribute, html_options %>
+    <%= form.label attribute, label %>
+  <% else %>
+    <%= hidden_field_tag name, 0 %>
```
jhawthorn: I wonder if we can simplify these cases using the new [`form_with`](https://github.com/rails/rails/issues/25197) :thinking:

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Solidus 2.3.0 (master, unreleased)
 
+- Deprecates several preference fields helpers in favor of preference field partials. [\#2040](https://github.com/solidusio/solidus/pull/2040) ([tvdeyen](https://github.com/tvdeyen))
```
jhawthorn: Can we bump this forward to 2.4?

backend/app/views/spree/admin/shared/preference_fields/_boolean.html.erb
tvdeyen: Good question. Would be nice though. Will give it a try

tvdeyen: Turns out this would be tricky, because we would need to rewrite all forms rendering these fields. In terms of the calculator fields this means to rewrite the calculator forms. I think we should tackle this in a separate PR. WDYT?

CHANGELOG.md
tvdeyen: Done

backend/app/views/spree/admin/shared/preference_fields/_boolean.html.erb
jhawthorn: Yeah, let's tackle this later.

PR #2043, opened 2017-06-27
Make Address.find_all_by_name_or_abbr case-insensitive

jordan-brough:
This is helpful when importing data from third parties. E.g. PayPal.

Note: I believe the spec failures are due to #2045

### DIFF
--- a/core/app/models/spree/state.rb
+++ b/core/app/models/spree/state.rb
@@ -6,7 +6,11 @@ module Spree
     validates :country, :name, presence: true
 
     def self.find_all_by_name_or_abbr(name_or_abbr)
-      where('name = ? OR abbr = ?', name_or_abbr, name_or_abbr)
+      where(
+        arel_table[:name].matches(name_or_abbr).or(
+          arel_table[:abbr].matches(name_or_abbr)
+        )
+      )
     end
 
     # table of { country.id => [ state.id , state.name ] }, arrays sorted by name
--- a/core/app/models/spree/state.rb
+++ b/core/app/models/spree/state.rb
@@ -5,12 +5,16 @@ module Spree
 
     validates :country, :name, presence: true
 
-    def self.find_all_by_name_or_abbr(name_or_abbr)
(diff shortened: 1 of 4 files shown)


core/app/models/spree/state.rb
```diff
@@ -6,7 +6,7 @@ class State < Spree::Base
     validates :country, :name, presence: true
 
     def self.find_all_by_name_or_abbr(name_or_abbr)
-      where('name = ? OR abbr = ?', name_or_abbr, name_or_abbr)
+      where('lower(name) = lower(:name_or_abbr) OR lower(abbr) = lower(:name_or_abbr)', name_or_abbr: name_or_abbr)
```
gmacdougall: Rather than using `lower` using arel matches would be preferred. This will use `ILIKE` in postgres, and `LIKE` in MySQL.

These types of comparisons can make better use of indexes than relying upon the SQL function return value.

Example:

```Ruby
where(
  arel_table[:name].matches(name_or_abbr).or(
    arel_table[:abbr].matches(name_or_abbr)
  )
)
```

core/spec/models/spree/state_spec.rb
```diff
@@ -1,10 +1,37 @@
...
+      it { is_expected.to be_empty }
+    end
+
+    context 'by name' do
+      let(:search_term) { 'California' }
+      it { is_expected.to include(state) }
+    end
+
+    context 'by abbr' do
+      let(:search_term) { 'California' }
```
gmacdougall: This doesn't match the context.

core/app/models/spree/state.rb
```diff
@@ -6,7 +6,7 @@ class State < Spree::Base
     validates :country, :name, presence: true
 
     def self.find_all_by_name_or_abbr(name_or_abbr)
```
jordan-brough: Completely agree.  I'll switch to `with_name_or_abbr` with a deprecation for the old method.

jordan-brough:
@gmacdougall @jhawthorn updated.  If all looks good I'll squash these down.

PR #2044, opened 2017-06-27
Include database requirement in README

bbuchalter:
Following these steps, it'd be easy to miss that you need a valid database for `bundle exec rails g spree:install` to work properly. You'd get this error otherwise: `bin/rails: FATAL:  database "solidus_demo_development" does not exist (ActiveRecord::NoDatabaseError)`


mamhoff:
We just ran into this today. The culprit is `solidus_auth_devise`, which uses its preferences singleton to figure out whether it should load the frontend routes or not. 

Adding an initializer `config/initializers/solidus_auth_devise_fix.rb` fixes it: 
```
Spree::Auth::Config.use_static_configuration!
```

PR #2046, opened 2017-06-27
Use `@collection` instead of `@collection.present?` in some admin controllers

jordan-brough:
I'm pretty sure `defined?` was the intended behavior.

Using `present?` can generate an extra query for the presence check, and will
also prevent the instance-variable caching from working when the query returns
no results.


jordan-brough:
> `return @collection if @collection` would also work

Good point, I've updated to that.

PR #2047, opened 2017-06-27
Use `.all` instead of `.where(nil)`

jordan-brough:
This is more readable IMO


jhawthorn:
I'd guess this was done for a much older version of rails (4.0 would have issued a warning). Due time we updated it :+1:

Issue #1901, opened 2017-05-09
Support testing with other Capybara drivers


jhawthorn:
Currently our test suite only passes when using phantomjs/poltergeist.

When I originally converted us over (spree/spree#4383) it supported both poltergeist/phantomjs and selenium/firefox, but we've since accidentally added/changed some specs which now break outside of poltergeist.

Since [phantomjs's future is uncertain](https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE) and has the [worst es6 support](https://kangax.github.io/compat-table/es6/#phantom) of all browsers we support (including IE11!), we may want to migrate away from it at some point (but hard to say for sure).

Supporting selenium could (maybe) also let us test across our supported browsers.

* [x] #2049
* [x] #2034

jhawthorn:
We are now using selenium + chrome headless for the backend :tada: :taco:

PR #2049, opened 2017-06-27
Spec fixes for selenium

jhawthorn:
This (as well as #2034) gets us closer to fixing #1901 and supporting chrome in it's new `--headless` mode.


gmacdougall:
I like the direction this is heading, but there's a couple of failures to take care of right now.

jhawthorn:
@gmacdougall Fixed by #2045 and another small change. Mind taking another look?

PR #2052, opened 2017-06-30
Convert API controller specs to request specs

jhawthorn:
Controller specs have been soft-deprecated in favour of using request specs. Request specs are more suitable for what we are testing here in most cases (we want to render views, and care about routing).


mtomov:
Pretty impressive! I never thought that anyone would go about to switch all those tests to the recommended type. Thanks a lot!

Issue #2050, opened 2017-06-28
DB Seed Scripts Exceeding max_questions Limit (MySQL)


pmn4:
This is absolutely a favor, as this is in response to something well outside the code base, but when seeding the database with countries and states, individual insert statements are used (I presume for re-runability), which results in > 3600 queries, and Heroku's `max_questions` limit is exceeded.

How do you feel about, in a first step, identifying all of the countries/states already in the database, and in a second step, combining the inserts into a single statement?

for reference, here is the [Spree Core version of the seed script](https://github.com/spree/spree/blob/master/core/db/default/spree/states.rb#L24-L27).

I will take a stab at a PR if you are comfortable with the change.
thanks!

### Steps to reproduce
Deploy the app to Heroku using MySQL and run `heroku run rake db:seed`

### Expected behavior
Databases seed scripts should complete

### Actual behavior
Heroku's MySQL `max_questions` limit is hit, stopping the scripts.

### System configuration
**Solidus Version**:
`2.3.0.alpha` (`9e00618`)

**Extensions in use**:

* solidus_auth_devise
* solidus_multi_domain

mamhoff:
That'd be a great PR!

jhawthorn:
Merged as #2097. Thanks!

PR #2053, opened 2017-07-03
[#2050] Refactor Countries Seed Script to use Fewer SQL Statements

pmn4:
RE: Issue #2050 - `countries.rb`

This PR eliminates between 246 and 990* _(explained below)_ SQL statements in the database seed process.
The resulting code queries the database for all existing entries and then crafts an insert statement for the ones that remain. The `INSERT` statement allows us to skip ActiveRecord hooks/validations, and since we are reasonably sure that there will not be a conflict, I think this is ok. (LMK if this is a bad assumption and I will add the extra checks)

I plan to refactor the `states.rb` and `zones.rb` assuming this passes your code review process.

* - 246 queries: 1 for each country to determine whether it is already in the database
if all countries already exist, that is the end of the script.
if any country does not exist, two additional uniqueness queries are performed, on the new countries `name` and `iso_name`, before the insert statement is performed.

### DIFF
--- a/core/db/default/spree/countries.rb
+++ b/core/db/default/spree/countries.rb
@@ -1,13 +1,43 @@
 require 'carmen'
 
-ActiveRecord::Base.transaction do
-  Carmen::Country.all.each do |country|
-    Spree::Country.where(iso: country.alpha_2_code).first_or_create!(
-      name: country.name,
-      iso3: country.alpha_3_code,
-      iso_name: country.name.upcase,
-      numcode: country.numeric_code,
-      states_required: country.subregions?
-    )
-  end
+# Insert Countries into the spree_countries table, checking to ensure that no
+# duplicates are created, using as few SQL statements as possible (2)
+
+connection = ApplicationRecord.connection
+
+country_mapper = ->(country) do
+  name            = connection.quote country.name
+  iso3            = connection.quote country.alpha_3_code
+  iso             = connection.quote country.alpha_2_code
+  iso_name        = connection.quote country.name.upcase
+  numcode         = connection.quote country.numeric_code
+  states_required = connection.quote country.subregions?
+
+  [name, iso3, iso, iso_name, numcode, states_required].join(", ")
+end
+
+country_values = -> do
+  carmen_countries = Carmen::Country.all
+
+  # find entires already in the database (so that we may ignore them)
+  existing_country_isos =
+    Spree::Country.where(iso: carmen_countries.map(&:alpha_2_code)).pluck(:iso)
+
+  # create VALUES statements for each country _not_ already in the database
+  carmen_countries
+    .reject { |c| existing_country_isos.include?(c.alpha_2_code) }
+    .map(&country_mapper)
+    .join("), (")
+end
+
+country_columns = %w(name iso3 iso iso_name numcode states_required).join(', ')
+country_vals = country_values.call
+
+if country_vals.present?
+  # execute raw SQL (insted of ActiveRecord.create) to use a single
+  # INSERT statement, and to avoid any validations or callbacks
+  connection.execute <<-SQL
+    INSERT INTO spree_countries (#{country_columns})
+    VALUES (#{country_vals});
+  SQL
 end


core/db/default/spree/countries.rb
```diff
@@ -1,13 +1,53 @@
...
+
+  [name, iso3, iso, iso_name, numcode, states_required].join(", ")
+end
+
+country_values = -> do
+  carmen_countries = Carmen::Country.all
+
+  # find entires already in the database (so that we may ignore them)
+  exiting_country_isos =
+    Spree::Country.where(iso: carmen_countries.map(&:alpha_2_code)).map(&:iso)
```
tvdeyen: I’m sure we can speed this up even more by using `pluck(:iso)` instead of `map`

core/db/default/spree/countries.rb
```diff
@@ -1,13 +1,53 @@
...
+  states_required = connection.quote country.subregions?
+
+  [name, iso3, iso, iso_name, numcode, states_required].join(", ")
+end
+
+country_values = -> do
+  carmen_countries = Carmen::Country.all
+
+  # find entires already in the database (so that we may ignore them)
+  exiting_country_isos =
```
tvdeyen: Typo in variable name `existing` vs. `exiting`

core/db/default/spree/countries.rb
```diff
@@ -1,13 +1,53 @@
...
+    Spree::Country.where(iso: carmen_countries.map(&:alpha_2_code)).map(&:iso)
+
+  # create VALUES statements for each country _not_ already in the database
+  carmen_countries
+  .reject { |c| exiting_country_isos.include?(c.alpha_2_code) }
```
tvdeyen: Please indent these lines

core/db/default/spree/countries.rb
```diff
@@ -1,13 +1,53 @@
...
+  columns = %w(name iso3 iso iso_name numcode states_required)
+
+  if connection.adapter_name =~ /MySQL/i
+    columns.join(", ")
+  else
+    "\"#{columns.join('", "')}\""
   end
 end
+
+country_cols = country_columns.call
```
tvdeyen: Is there any reason why we put this code in a callable instead of just evaluate the code in place?

core/db/default/spree/countries.rb
pmn4: funny, `rubocop` autocorrected those lines! happy to switch them back

core/db/default/spree/countries.rb
pmn4: none. just thought it would read nicely next to the following line.
will call in place

core/db/default/spree/countries.rb
pmn4: TIL!
will do.

core/db/default/spree/countries.rb
```diff
@@ -1,13 +1,52 @@
...
+
+country_columns = -> do
+  columns = %w(name iso3 iso iso_name numcode states_required)
+
+  if connection.adapter_name =~ /MySQL/i
+    columns.join(", ")
+  else
+    "\"#{columns.join('", "')}\""
   end
 end
```
kennyadsl: I tried the script locally and `columns.join(", ")` seems to work even with `postgresql` and `sqlite3`. Why the `else` version is needed?

pmn4: good question. tbh, I copied that piece from [Spree core version](https://github.com/spree/spree/blob/master/core/db/default/spree/countries.rb#L22) and never tested it

kennyadsl: Just found some (maybe) useful comments [here](https://github.com/spree/spree/pull/5730#issuecomment-66370008).

pmn4:
I hear you on the readability.
I wrote an initial version which used `Spree::Country.create(array_of_attribute_hashes)`, but found that ActiveRecord creates individual INSERT statements.

jhawthorn:
Merged as #2053. Thank you :tada:

PR #2060, opened 2017-07-04
Use CircleCI 2.0

jhawthorn:
This switches our CircleCI config to CircleCI 2.0 (which is in beta). This allows us to use docker containers and gives us better control over caching.

The big challenge in making this was getting the caching to work since we don't have a Gemfile.lock checked into the repo. CircleCI has odd, but useful, behaviour when the cache key isn't an exact hit: it will
find the most recent cache key which our specified key is a prefix of.

Because of this we can save to the cache key

    solidus-gems-v1-{{ .Branch }}-{{ checksum "Gemfile.lock" }}

but restore from the key

    solidus-gems-v1-{{ .Branch }}

and we will always get the most recent cache for each branch, and won't waste time saving cache for each run, only when our bundle install finds new gems.

This uses CircleCI's provided 2.4.1-node-browsers docker container. We could use our own custom docker container, but this should be a good choice for us for now and avoids us needing to fetch phantomjs or Node each run.


bbuchalter:
Any data on test runtime changes using the in-memory postgres?

jhawthorn:
> Any data on test runtime changes using the in-memory postgres?

@bbuchalter It is a bit faster, and more similar to the speeds we had on CircleCI 1 (making me think CIrcleCI probably did this previously). Are you able to view build timing in our CircleCI? https://circleci.com/gh/jhawthorn/solidus/1134#build-timing https://circleci.com/gh/jhawthorn/solidus/1120#build-timing

bbuchalter:
Can't see build timing:
screen shot 2017-07-06 at 1 18 04 pm

jhawthorn:
Rats. Here's timing without details.

*Before*


*After*


Mostly notably this shaves about a minute off of the "backend" builds. My guess is that this is because `TRUNCATE` is faster.

PR #2062, opened 2017-07-05
Display a pointer cursor hovering add variant buttons

kennyadsl:
The button tag does not display a pointer cursor. Using an anchor tag also uniforms the markup with other [similar buttons](https://github.com/solidusio/solidus/blob/ba852180327f1e43eb9ad9d8f9395211af0c9d7b/backend/app/assets/javascripts/spree/backend/templates/variants/split.hbs#L23-L24) present in the line items management area.

*Before*

schermata 2017-07-05 alle 15 50 58

*After*

schermata 2017-07-05 alle 15 50 22

### DIFF
--- a/backend/app/views/spree/admin/orders/_shipment.html.erb
+++ b/backend/app/views/spree/admin/orders/_shipment.html.erb
@@ -62,9 +62,9 @@
           </td>
           <td class="actions">
             <% if can? :update, shipment %>
-              <%= link_to '', '#', class: 'save-method fa fa-check no-text with-tip',
+              <%= button_tag '', class: 'save-method fa fa-check no-text with-ti, type: :button',
                 data: {'shipment-number' => shipment.number, action: 'save'}, title: Spree.t('actions.save') %>
-              <%= link_to '', '#', class: 'cancel-method fa fa-cancel no-text with-tip',
+              <%= button_tag '', class: 'cancel-method fa fa-cancel no-text with-ti, type: :button',
                 data: {action: 'cancel'}, title: Spree.t('actions.cancel') %>
             <% end %>
           </td>
@@ -85,7 +85,7 @@
 
           <td class="actions">
             <% if can?(:update, shipment) && !shipment.shipped? %>
-              <%= link_to '', '#', class: 'edit-method fa fa-edit no-text with-tip', data: {action: 'edit'}, title: Spree.t('actions.edit') %>
+              <%= button_tag '', class: 'edit-method fa fa-edit no-text with-tip', data: {action: 'edit'}, title: Spree.t('actions.edit'), type: :button %>
             <% end %>
           </td>
         </tr>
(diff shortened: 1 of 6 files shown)


mtomov:
Well, I'd say that the problem is in the CSS rather than the markup. Button can even be thought to be better semantically, because the action that you are about to perform is not a `GET` one.

Also, I'd doubt that this is the only place, which uses buttons, so applying such a change site-wide using css should a 1 liner.

kennyadsl:
@mtomov thanks for your feedback! I initially thought the same thing and I was surprised that buttons do not have `cursor: pointer` class by default. Maybe there are buttons where this is not needed? Anyway I moved my steps in this direction (using an anchor) mainly to uniform the markup between all the other similar buttons, like [here](https://github.com/solidusio/solidus/blob/ba852180327f1e43eb9ad9d8f9395211af0c9d7b/backend/app/assets/javascripts/spree/backend/templates/variants/split.hbs#L23-L24). I agree with you about the better semantic using a button, even if that's not a real form since the action is performed only via js and I don't think that page will work without it.

jhawthorn:
I don't think these elements should have the "pointer" cursor. I think the buttons which do have the pointer cursor are a mistake due to them being `<a>` tags (also a mistake IMO, but that's another discussion)

["Buttons shouldn’t have a hand cursor" by Adam Silver](https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b) argues for this and has some reference to authorities saying the same: [Microsoft](https://msdn.microsoft.com/en-us/library/windows/desktop/dn742466%28v=vs.85%29.aspx), [Apple](https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/Pointers.html), [w3c](https://www.w3.org/TR/CSS21/ui.html#propdef-cursor).

cc @Mandily who may have some thoughts on the matter

Mandily:
I'm of the thought that anything interactive on the page should have the affordance of a pointer cursor, or a text cursor if it's an input. I understand where the author of that linked post is coming from, that in theory interactive things should look like they can be interacted with, and therefore don't need the cursor to tell users. But in practice, we have flat designs with less affordance, and the pointer cursor only helps to reinforce.

kennyadsl:
I read that article before doing this PR but I don't fully agree with the author. Like a lot of comments do, I also think that it's more important to focus on user expectations. 

I think that Microsoft and Apple resources linked refer to Desktop application, which maybe is a bit different from a web application. w3c [takes a strong position](http://lists.w3.org/Archives/Public/public-css-testsuite/2010Jul/0024.html) in favor of not using pointer cursors for elements that are not links.

Anyway I'm starting to feel that this discussion is too much general and would require a separate issue for discussion. If we decide to go with removing pointer from buttons (and other elements that are not links) we probably have to change a lot of stuff in both backend and frontend markup/style.

kennyadsl:
Updated the PR so that we now have `<button>` elements instead of `<a>` for shipment/shipment manifest views and js templates, which is semantically better for everyone. The pointer cursor is added via css to all buttons inside a `table td.actions`.

backend/app/views/spree/admin/orders/_shipment.html.erb
```diff
@@ -62,9 +62,9 @@
           </td>
           <td class="actions">
             <% if can? :update, shipment %>
-              <%= link_to '', '#', class: 'save-method fa fa-check no-text with-tip',
+              <%= button_tag '', class: 'save-method fa fa-check no-text with-ti, type: :buttonp',
```
tvdeyen: Typo at the end of the line

backend/app/views/spree/admin/orders/_shipment.html.erb
```diff
@@ -62,9 +62,9 @@
...
                 data: {'shipment-number' => shipment.number, action: 'save'}, title: Spree.t('actions.save') %>
-              <%= link_to '', '#', class: 'cancel-method fa fa-cancel no-text with-tip',
+              <%= button_tag '', class: 'cancel-method fa fa-cancel no-text with-ti, type: :buttonp',
```
tvdeyen: Ditto

backend/app/views/spree/admin/orders/_shipment.html.erb
kennyadsl: thank you!

backend/app/views/spree/admin/orders/_shipment.html.erb
kennyadsl: both fixed now!

graygilmore:
Big fan of changing empty anchors to buttons 👍

mtomov:
thanks!

PR #2064, opened 2017-07-06
Fix Paperclip::Errors::NotIdentifiedByImageMagickError on invalid images

karlentwistle:
In the paperclip README it says NOTE: Post processing will not even start if
the attachment is not valid according to the validations. Your callbacks and
processors will only be called with valid attachments.

However find_dimensions is still being called and unless a valid attachment is
attached to Image the error Paperclip::Errors::NotIdentifiedByImageMagickError
is raised.

This pull requests adds a guard clause to after_post_process so it only calls
find_dimensions if the image is valid. There wasn't any unit tests for image
prior to this PR so it also introduces a very simple happy path test for save
and the inverse to prevent future regression of the aforementioned issue.

### DIFF
--- a/core/spec/models/spree/image_spec.rb
+++ b/core/spec/models/spree/image_spec.rb
@@ -1,0 +1,23 @@
+require 'spec_helper'
+
+describe Spree::Image, type: :model do
+  context '#save' do
+    context 'invalid attachment' do
+      let(:invalid_image) { File.open(__FILE__) }
+      subject { described_class.new(attachment: invalid_image) }
+
+      it 'returns false' do
+        expect(subject.save).to be false
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


core/spec/models/spree/image_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+describe Spree::Image, type: :model do
+  context '#save' do
+    context 'invalid attachment' do
+      let(:invalid_image) { File.open(__FILE__) }
```
jhawthorn: I love this

mtomov: Same here! 

And congrats on the first tests on the `Image` model : )

PR #2065, opened 2017-07-06
WIP: Order not triggering callbacks on shipments

skukx:
When `complete!` is called on order, it then updates all the shipments
to `ready`. The problem is that this update skips state_machine
transition callbacks. Write spec to show this behavior.

### DIFF
--- a/core/spec/models/spree/order/finalizing_spec.rb
+++ b/core/spec/models/spree/order/finalizing_spec.rb
@@ -32,7 +32,10 @@ describe Spree::Order, type: :model do
     end
 
     it "should change the shipment state to ready if order is paid" do
-      Spree::Shipment.create(order: order)
+      Spree::Shipment.state_machine.after_transition to: :ready, do: :callback
+      shipment = Spree::Shipment.create(order: order)
+      expect(shipment).to receive(:callback)
+
       order.shipments.reload
 
       allow(order).to receive_messages(paid?: true, complete?: true)


kennyadsl:
Also, from Rails 4+ that [`update!` method on `Spree::Shipment`](https://github.com/solidusio/solidus/blob/6d5d40d9827cc449dfc9a7682cfce827a1f33f5c/core/app/models/spree/shipment.rb#L309-L322)  is overriding the [ActiveRecord update! method](https://apidock.com/rails/ActiveRecord/Persistence/update%21) and I don't think it has been done intentionally. 

P.S. We probably have the same problem on `Spree::Order`

cbrunsdon:
Hi @skukx, a few of us are in the middle of a conversation about this IRL and we're trying to get a better picture about how people are using the state machine on shipments (because as you can see it has a lot of issues).

Can you share what it is you're looking for on ready being called? Is there a "better" time you actually want it called, but are hooking into `ready` because it should be easy, or is `shipment.ready` for sure the exact time you want something to be executed?

skukx:
@cbrunsdon right now we are working on integrating with our WMS.  The idea was to send our wms what items to ship as the `Solidus::Shipment` moves to ready.  This would ensure that payment is full, and that there is enough stock.  Normally the WMS could probably account for the stock portion  but the WMS we have a contract with doesn't have the most friendly API.

jhawthorn:
We understand the problem here and desire to have a better way to hook into state changes. We'd like to find a better way for stores to hook into events like orders completing, payments being received, and shipments being shipped.

Unfortunately the Shipment state machine isn't really a state machine. It's mostly a calculated state field. I'm not even sure the "ready" state should exist.

Closing because this isn't mergeable or immediately actionable.

PR #2070, opened 2017-07-07
[RFC] Add FulfillmentChanger

mamhoff:
(built on top of #1906, without which this does not work either)

This class takes two shipments and moves a certain amount of stuff
between them. This used to be called "splitting" a shipment, but really we
are changing fulfillment for those variants, hence the name.

This will replace the `transfer_to_*` methods on `Spree::Shipment`. 

It is designed to be quite a bit faster, since it does not move inventory units one by one, but in bulk.

There's a slight change in behaviour: When splitting a shipment such that the new shipment would have more inventory than the original one, that's not possible anymore, because I can't imagine a situation where that would be beneficial. 

I had to adjust one spec for the aforementioned change in behaviour. I also had to amend one spec so that the order starts out with only on_hand inventory units.


core/app/models/spree/fulfilment_changer.rb
```diff
@@ -0,0 +1,92 @@
...
+    validate :enough_stock_at_desired_location, if: :handle_stock_location_counts?
+
+    def run!
+      return false if invalid?
+      desired_shipment.save! if desired_shipment.new_record?
+
+      current_on_hand_quantity = [current_shipment.inventory_units.on_hand.size, quantity].min
+      new_on_hand_quantity = [desired_shipment.stock_location.count_on_hand(variant), quantity].min
+
+      if handle_stock_location_counts?
```
tvdeyen: Maybe it makes sense to put all the stock and shipment changing code into a transaction to avoid potential stock count mismatches?

core/app/models/spree/fulfilment_changer.rb
```diff
@@ -0,0 +1,92 @@
+module Spree
+  class FulfilmentChanger
```
tvdeyen: I would like to see some documentation on this class.

backend/app/assets/javascripts/spree/backend/shipments.js
```diff
@@ -189,8 +189,12 @@ var ShipmentSplitItemView = Backbone.View.extend({
     }
     jqXHR.error(function(msg) {
       alert(Spree.t("split_failed"));
-    }).done(function() {
-      window.Spree.advanceOrder();
+    }).done(function(response) {
+      if (response.success) {
+        window.Spree.advanceOrder();
+      } else {
+        alert(response.message);
```
mamhoff: Using `show_flash` resulted in quite a lot of testing trouble for some reason: Locked Sqlite databases, hung phantomjs, weird things. Can we move that to a separate PR?

tvdeyen: Sure

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Solidus 2.3.0 (master, unreleased)
 
+- Change HTTP Status code for Api::ShipmentsController#transfer_to_* to be always 202 Accepted rather than 201 Created or 500.
+  Speed up changing fulfilment for parts of a shipment [\#2070](https://github.com/solidusio/solidus/pull/2070) ([mamhoff](https://github.com/mamhoff))
```
jhawthorn: These will be in 2.4.0

mamhoff: Changed

mamhoff:
Thank you John for updating the changelog. I've had too much on my plate.

PR #2072, opened 2017-07-07
[RFC] Rename order.update! to order.recalculate

jhawthorn:
I expect this to be contentious, and it definitely introduces some work for users upgrading, which I'm always cautious of.

`ActiveRecord::Base` has an `update!(attributes)` which as of Rails 4.0 is the recommended way to update the attributes on a model. [The PR introducing it](https://github.com/rails/rails/pull/8705) describes the older `update_attributes!` as "soft-deprecated".

Because we've overridden `update!`, we can't use that method to update the the attributes on order (as `update!` does for every other model), and have to use `update_attributes!`.

This isn't great because Rails devs (as well as documentation) are going to expect `update!` to work and might not even be aware of `update_attributes!`. This is doubly confusing because we have access to ActiveRecord's `order.update`, and one really expects that to do approximately the same thing as the `!` version.

This PR adds `Order#recalculate`, which does exactly what `Order#update!` did, invoking the `OrderUpdater`.

`Order#update!` will now, if given attributes, will behave like AR's `update!`. If given no attributes it will emit a deprecation warning and call `recalculate`.

If we decide to do this, we should probably do the same for `Adjustment#update!` and `Shipment#update!`

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -254,8 +254,17 @@ module Spree
       @updater ||= Spree::OrderUpdater.new(self)
     end
 
-    def update!
+    def recalculate
       updater.update
+    end
+
+    def update!(*args)
+      if args.empty?
+        Spree::Deprecation.warn "Calling order.update! with no arguments as a way to invoke the OrderUpdater is deprecated, since it conflicts with AR::Base#update! Please use order.recalculate instead"
+        recalculate
+      else
+        super
+      end
     end
 
     def assign_billing_to_shipping_address
@@ -471,7 +480,7 @@ module Spree
       shipments.destroy_all
       order_promotions.destroy_all
 
-      update!
+      recalculate
     end
 
     alias_method :has_step?, :has_checkout_step?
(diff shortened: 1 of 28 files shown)


core/app/models/spree/order.rb
```diff
@@ -254,10 +254,20 @@ def updater
       @updater ||= Spree::OrderUpdater.new(self)
     end
 
-    def update!
+    def recalculate
       updater.update
     end
 
+    alias_method :update!, :recalculate
```
tvdeyen: Why this alias?

jhawthorn: Mistake. Fixed

gmacdougall:
I am also on board with this.

subhashsaran:
`recalculate ` Sounds better than spree `update_with_updater!`
https://github.com/spree/spree/blob/v3.1.5/core/app/models/spree/order.rb#L235-L243
👍

jhawthorn:
Thanks for the feedback everyone. This was a lot less contentious than I expected.

Merging now. As the 2.3 branch has been forked now, this will be in our future 2.4.0 release.

PR #2073, opened 2017-07-10
Don't @extend compound selectors in sass

jhawthorn:
The ability to extend compound selectors has been deprecated and will be removed from sass.

This fixes the following deprecation warnings

```
DEPRECATION WARNING on line 49 of 
Extending a compound selector, [class^="icon-"]:before, is deprecated and will not be
supported in a future release.
See https://github.com/sass/sass/issues/1599 for details.
```

See https://github.com/sass/sass/issues/1599


mamhoff:
`.flash.error` apparently also counts as a compound selector. 
[This line](https://github.com/solidusio/solidus/blob/master/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss#L404) still triggers that warning in the sandbox. 

Otherwise, things look still good. :)

jhawthorn:
Good catch @mamhoff. Fixed

PR #2075, opened 2017-07-11
Invalidate existing non store credit payments during checkout

tvdeyen:
When creating a new payment for an order having existing payments
we want all previous payments in checkout state to be invalidated; except
if they are store credit payments.
                        
Former implementation only invalidated payments with same payment method.
That led to multiple valid payments per order for stores having more then
one payment method available for customers to choose from.


tvdeyen:
Waits for #2076

PR #2076, opened 2017-07-11
Deprecate Spree::OrderCapturing class

tvdeyen:
This task was added to automate capturing pending payments on a regular basis via a rake task. This kind of custom business logic should not be maintained by us. Half of the class features is not even used by the provided rake task and it makes further improvements around payments harder to test.


tvdeyen:
Prerequisite for #2075

tvdeyen:
@jordan-brough the tests in #2075 fail because the specs for `Spree::OrderCapturing` fail.

PR #2077, opened 2017-07-11
Revert "Select all shipping rates in the Estimator"

jhawthorn:
Reverts solidusio/solidus#1984

Unfortunately, this would allow (mildly) malicious customers to select backend-only rates for their others. 

Steps to reproduce:
* On a store with a (supposed to be secret) free backend-only shipping rate
* Perform a checkout as an unprivileged user up to the delivery step
* Use inspector (or similar) to change the radio button's value to the value of a backend only shipping rate (this can be guessed easily by numbers near the id of the customer-available rates)
* Submit the page, receive the free backend-only shipping rate

### DIFF
--- a/core/app/models/spree/shipping_rate.rb
+++ b/core/app/models/spree/shipping_rate.rb
@@ -35,10 +35,6 @@ module Spree
     end
     alias_method :display_cost, :display_price
 
-    def available_to_user?
-      shipping_method.available_to_users? || shipment.selected_shipping_rate == self
-    end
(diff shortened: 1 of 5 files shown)


jhawthorn:
Agree with martin, this was good and desirable other than the issue of users being able to bypass restrictions. We absolutely want to support admin-only shipping rates.

I've tagged the issue for milestone 2.4.0, let's see if we can find a solution next release.

PR #2080, opened 2017-07-12
Deprecate tasks in core/lib/tasks

cbrunsdon:
None of these tasks should be used by people.

This deprecates those tasks, and removes a dependency that exists just to support these tasks (highline).

### DIFF
--- a/core/db/seeds.rb
+++ b/core/db/seeds.rb
@@ -1,5 +1,18 @@
 # Loads seed data out of default dir
 default_path = File.join(File.dirname(__FILE__), 'default')
 
-Rake::Task['db:load_dir'].reenable
-Rake::Task['db:load_dir'].invoke(default_path)
+%w(
+  stores
+  store_credit
+  countries
+  return_reasons
+  states
+  stock_locations
+  zones
+  refund_reasons
+  roles
+  shipping_categories
+).each do |seed|
+  puts "Loading seed file: #{seed}"
+  require_relative "default/spree/#{seed}"
+end
--- a/core/lib/tasks/core.rake
+++ b/core/lib/tasks/core.rake
@@ -10,6 +10,7 @@ namespace :db do
 use rake db:load_file[/absolute/path/to/sample/filename.rb]'
 
   task :load_file, [:file, :dir] => :environment do |_t, args|
+    Spree::Deprecation.warn("load_file has been deprecated. Please load your own file.")
     file = Pathname.new(args.file)
 
     puts "loading ruby #{file}"
@@ -18,6 +19,7 @@ use rake db:load_file[/absolute/path/to/sample/filename.rb]'
 
   desc "Loads fixtures from the the dir you specify using rake db:load_dir[loadfrom]"
   task :load_dir, [:dir] => :environment do |_t, args|
+    Spree::Deprecation.warn("rake spree:load_dir has been deprecated and will be removed with Solidus 3.0. Please load your files directly.")
     dir = args.dir
     dir = File.join(Rails.root, "db", dir) if Pathname.new(dir).relative?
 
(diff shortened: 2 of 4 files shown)


core/db/seeds.rb
```diff
@@ -1,5 +1,8 @@
 # Loads seed data out of default dir
 default_path = File.join(File.dirname(__FILE__), 'default')
 
-Rake::Task['db:load_dir'].reenable
-Rake::Task['db:load_dir'].invoke(default_path)
+['stores', 'store_credit', 'countries', 'return_reasons', 'states',
+ 'stock_locations', 'zones', 'refund_reasons', 'roles', 'shipping_categories'].each do |seed|
```
kennyadsl: I like formatting this kind of array with each element in a separate line, mainly to keep git commits history clean if we change something, but it's just my personal preference, I'm ok with this as well.

tvdeyen: Yes. I also would like to have this separated by line 👍 

And maybe use `%w()` syntax, but I don't care about this very much.

core/lib/tasks/core.rake
```diff
@@ -1,10 +1,16 @@
...
+  print prompt
+  ["y", "yes"].include? STDIN.gets.strip.downcase
+end
+
 namespace :db do
   desc 'Loads a specified fixture file:
 use rake db:load_file[/absolute/path/to/sample/filename.rb]'
 
   task :load_file, [:file, :dir] => :environment do |_t, args|
+    ActiveSupport::Deprecation.warn("load_file has been deprecated. Please load your own file.")
```
tvdeyen:
```ruby
Spree::Deprecation.warn
```

core/lib/tasks/core.rake
```diff
@@ -13,6 +19,7 @@ use rake db:load_file[/absolute/path/to/sample/filename.rb]'
 
   desc "Loads fixtures from the the dir you specify using rake db:load_dir[loadfrom]"
   task :load_dir, [:dir] => :environment do |_t, args|
+    ActiveSupport::Deprecation.warn("load_dir has been deprecated. Please load your own files.")
```
tvdeyen:
```ruby
Spree::Deprecation.warn
```

core/lib/tasks/core.rake
tvdeyen: I think we can be a little more verbose/descriptive on this deprecation notice

> `rake spree:load_dir` has been deprecated and will be removed with Solidus 3.0. Please `load` your seed files directly from your app's `db/seeds.rb` file.

core/lib/tasks/core.rake
```diff
@@ -32,9 +39,9 @@ use rake db:load_file[/absolute/path/to/sample/filename.rb]'
 
   desc "Migrate schema to version 0 and back up again. WARNING: Destroys all data in tables!!"
   task remigrate: :environment do
-    require 'highline/import'
+    ActiveSupport::Deprecation.warn("remigrate has been deprecated. Please demigrate and then migrate again, I guess?")
```
tvdeyen: I think `rake db:reset` is what we want to recommend, no?

Also `Spree::Deprecation.warn`

cbrunsdon:
Cleaned up the list of seeds and removed any trace of humour from the deprecations

core/lib/tasks/core.rake
```diff
@@ -13,6 +19,7 @@ use rake db:load_file[/absolute/path/to/sample/filename.rb]'
 
   desc "Loads fixtures from the the dir you specify using rake db:load_dir[loadfrom]"
   task :load_dir, [:dir] => :environment do |_t, args|
+    Spree::Deprecation.warn("rake spree:load_dir has been deprecated and will be removed with Solidus 3.0. Please load your files directly file.")
```
mamhoff: `Please load your files directly file.` I think there's a file too much.

PR #2082, opened 2017-07-12
Mark ffaker as a common Spree dependency

dangerdogz:
FFaker is used by the test suite and should not be a production
dependency.


kennyadsl:
@dangerdogz I think ffaker is still needed as dependency since it is used to build [sample data](https://github.com/solidusio/solidus/blob/f9f610bdd13dc20f7350c8f29cc80d738ce608a9/sample/db/samples/addresses.rb#L6) which could even be run in production.

dangerdogz:
Will investigate more

PR #1956, opened 2017-05-23
Remove custom responders

omnistegan:
This PR removes custom responders, in favor of placing them in an extension.

I'm currently hosting the extension on my github [here](https://github.com/omnistegan/solidus_responders).

This resolves the first part of https://github.com/solidusio/solidus/issues/1869

PR #2083, opened 2017-07-12
Remove spree responders

cbrunsdon:
Pull responders out of the project.

I don't know what they do and I'm not smart enough to figure it out, so
I'm almost positive they aren't used.


cbrunsdon:
Closing this for right now. Learnt what these did and realize I have to kill them harder.

kennyadsl:
@cbrunsdon also, there's another PR ( #1956 ) for this same task.

cbrunsdon:
@kennyadsl oh nice :D. should have looked first. So @omnistegan how is #1956 looking? (also I'm looking at you right now, when you're curious about why I'm staring at you later).

PR #2084, opened 2017-07-12
Fix shipping method factory for stores with alternate currency

Sinetheta:
Without this change, a store with an alternate default currency will not find shipping rates for it's orders and shipments for those orders won't be shipable.

### DIFF
--- a/core/lib/spree/testing_support/factories/shipping_method_factory.rb
+++ b/core/lib/spree/testing_support/factories/shipping_method_factory.rb
@@ -23,7 +23,7 @@ FactoryGirl.define do
 
     transient do
       cost 10.0
-      currency 'USD'
+      currency { Spree::Config[:currency] }
     end
 
     before(:create) do |shipping_method, _evaluator|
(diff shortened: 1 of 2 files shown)


core/lib/spree/testing_support/factories/shipping_method_factory.rb
```diff
@@ -19,11 +19,11 @@
     carrier 'UPS'
     service_level '1DAYGROUND'
 
-    calculator { |s| s.association(:shipping_calculator, strategy: :build, preferred_amount: s.cost, preferred_currency: s.currency) }
+    calculator { |s| puts s.currency; s.association(:shipping_calculator, strategy: :build, preferred_amount: s.cost, preferred_currency: s.currency) }
```
tvdeyen: Development artifact?

Sinetheta: 😳  this is why I never `git add .`

bbuchalter:
💚 specs for factories!

PR #2089, opened 2017-07-13
Make rubocop inherit_from-friendly

jordan-brough:
I inherit from Solidus rubocop rules in my Solidus application.
This change makes this rule work in my app as well as in Solidus.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -116,6 +116,7 @@ Rails/DynamicFindBy:
 Lint/AmbiguousBlockAssociation:
   Exclude:
     - '*/spec/**/*'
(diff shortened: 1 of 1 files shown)


mamhoff:
I think you would need to add `spec/**/*` rather than changing it. Because Solidus is split up in submodules, the actual tests are in `core/spec`, `backend/spec` and so on, and `*` apparently doesn't match nothing. :)

jordan-brough:
@mamhoff that's what I did isn't it? The original is still there. :)

(Btw Another option could be to change it from `*/spec/**/*` to `**/spec/**/*`, but that felt like it might be too broad)

PR #2090, opened 2017-07-14
Remove responders dependency from core

cbrunsdon:
Moves the responders dependency to the API, the place where we actually use it.

I'll rebase this once #1956 is merged, I just want to run tests.


cbrunsdon:
Okay with #1956 gone this is good to go whenever someone has a sec

PR #2091, opened 2017-07-14
Add missing default_currency field on admin/stores

oeN:
The Spree::Store have a default_currency field that is used in
`core/lib/spree/core/controller_helpers/pricing.rb` but never valorized
because the field was missing from the store form.

### DIFF
--- a/backend/spec/features/admin/stores_spec.rb
+++ b/backend/spec/features/admin/stores_spec.rb
@@ -1,0 +1,35 @@
+require 'spec_helper'
+
+describe 'Stores', type: :feature do
+  stub_authorization!
+
+  context 'when adding a store' do
+    before { visit spree.new_admin_store_path }
+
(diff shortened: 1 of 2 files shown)


backend/spec/features/admin/stores_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    fill_in 'store_name', with: 'Solidus Store'
+    fill_in 'store_code', with: 'solidus'
+    fill_in 'store_url', with: 'example.solidus.io'
+    fill_in 'store_mail_from_address', with: 'from@solidus.io'
+    select 'EUR', from: 'store_default_currency'
+    click_button "Create"
+
+    @store = Spree::Store.last
+
+    expect(@store.default_currency).to eq 'EUR'
```
kennyadsl: @oeN maybe we should also test that the currency wasn't `EUR` before selecting/saving it?

PR #2095, opened 2017-07-18
Invalidate old payments done with different payment method

kennyadsl:
Previously if a user makes a payment with payment method A, go to confirm, go back to payment and makes another payment with payment method B, both payments remains in 'checkout' state.

The correct behavior should be invalidating old payments whatever payment method they were created with.

I have some doubt about the in memory `order` object, which need a `.reload` to have both payments available. In fact when the `after_create :invalidate_old_payments` is called, it only contains the old payments without the newly added one.

### DIFF
--- a/core/app/models/spree/payment.rb
+++ b/core/app/models/spree/payment.rb
@@ -230,9 +230,8 @@ module Spree
 
     def invalidate_old_payments
       if !store_credit? && !['invalid', 'failed'].include?(state)
-        order.payments.select do |payment|
+        order.reload.payments.select do |payment|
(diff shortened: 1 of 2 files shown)


tvdeyen:
Potential duplicate of #2075 ?

kennyadsl:
@tvdeyen yes! Didn't noticed that PR. I noticed that when `invalidate_old_payments`, `order.payments` didn't contain the newly created payment. It could not be an issue since we want to invalidate only old payments but it can cause issues in other places maybe, what do you think?

Going to close this one anyway!

PR #2096, opened 2017-07-18
Revert rename of spree.rb initializer

### LINKED ISSUE #2033: Initializer rename breaks extensions
I just want to point out that the rename of the `spree.rb` initializer to `solidus.rb` in #1864 breaks some extension generators.
For example i have this problem with [solidus_affirm](https://github.com/StemboltHQ/solidus_affirm/blob/affirm-payment-method/lib/generators/solidus_affirm/install/install_generator.rb#L17) and [solidus_scaffold](https://github.com/freego/solidus_scaffold/blob/master/lib/generators/solidus_scaffold/scaffold/scaffold_generator.rb#L46).

Can we maybe add some helper to `solidus_support` to help backward compatibility?

### Steps to reproduce
- Install latest solidus
- Run any extension generator that writes on `spree.rb`

### Expected behavior
The generator should run

### Actual behavior
`Errno::ENOENT: No such file or directory @ rb_sysopen - (...)/solidus_affirm/spec/dummy/config/initializers/spree.rb`

### System configuration
**Solidus Version**: 2.3.0.alpha
**Extensions in use**: solidus_affirm

jhawthorn:
This partially reverts 49ab6ef873d6b671e8b3a388efd227fa64d1fc8f

We wanted to rename the `spree.rb` initializer to `solidus.rb`, but too many extension installations still expect it to be at `spree.rb` (and will probably want to for a while since that will be the name for upgrading
stores).

Fixes #2033


tvdeyen:
Short term merge as this is actually a bug fix.

PR #2097, opened 2017-07-18
Reduce number of SQL statements in countries seeds

jhawthorn:


jhawthorn:
This is a rebase of #2053, which is already approved.

PR #2098, opened 2017-07-18
Fix class/module nesting

cbrunsdon:
We had some cases in solidus where we were unclearly nesting classes and modules inside other classes and modules.

This breaks out the unclear definitions into what they should be.


cbrunsdon:
Ahh, I guess require_dependency is what the autoloader uses to do these things. But yea, I'm happy explicitly spelling it out.

bbuchalter:
> a class in a module in a class in a module
😆

PR #2092, opened 2017-07-14
[WIP] tables

Mandily:
I've started some work to clean up the tables in the Solidus admin. I think a lot of the existing table code is redundant now that we've got Bootstrap in the project, so I removed what I could to clean it up, and then added a couple of styles back in to "clean up" what was left. I've just tackled the index pages (except for the stock page) at the moment, and I'd like some feedback before I go through the rest of the views. 

Orders
screen shot 2017-07-13 at 5 01 01 pm

Products
screen shot 2017-07-13 at 5 01 19 pm

Reports
screen shot 2017-07-13 at 5 00 41 pm

Promotions
screen shot 2017-07-13 at 5 05 50 pm

Users
screen shot 2017-07-13 at 5 22 22 pm

@Sinetheta  @jhawthorn @tvdeyen @graygilmore can I get your thoughts on this?

PR #2100, opened 2017-07-19
[RFC] New action icons style

tvdeyen:
No more background and rounded corner. Using the icon only.
More cleaner and reduced interface.

new action buttons

In preparation for #2092 

## Even more beautiful with the new table style

orders with new action buttons

products with new action buttons

/c @Mandily


kennyadsl:
I like this change, the interface looks more clean and professional in my opinion.

PR #2101, opened 2017-07-19
Preview Images in a Modal

graygilmore:
Additionally this PR adds a shared modal partial for the admin and instructions on its usage to the style guide.

modal

### DIFF
--- a/backend/app/views/spree/admin/images/_image_row.html.erb
+++ b/backend/app/views/spree/admin/images/_image_row.html.erb
@@ -7,9 +7,19 @@
   </td>
 
   <td>
-    <%= link_to image.attachment.url(:product) do %>
+    <span class="modal-image-trigger" data-toggle="modal" data-target="#modal-image-<%= image.id %>">
       <%= render 'spree/admin/shared/image', image: image, size: :mini %>
-    <% end %>
+    </span>
+    <%= render(
+      "spree/admin/shared/modal",
+      target: "modal-image-#{image.id}",
+      title: image.alt,
+      content: %{
+        <div class='align-center'>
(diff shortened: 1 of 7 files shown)


backend/app/views/spree/admin/images/_image_row.html.erb
```diff
@@ -7,9 +7,15 @@
...
-    <%= link_to image.attachment.url(:product) do %>
+    <span class="modal-image-trigger" data-toggle="modal" data-target="#modal-image-<%= image.id %>">
       <%= render 'spree/admin/shared/image', image: image, size: :mini %>
-    <% end %>
+    </span>
+    <%= render(
+      "spree/admin/shared/modal",
+      target: "modal-image-#{image.id}",
+      title: image.alt,
+      content: image_tag(image.attachment.url(:large))
```
tvdeyen: Just my personal taste, but wouldn’t it be nice to wrap the image in `div` that centers the image?

graygilmore:
Updated to include centering of the images. This will likely not happen on normal stores as the images we're selecting to go in this modal are larger than the modal itself but I'm happy to provide a fallback for smaller images.

graygilmore:
@tvdeyen a separate PR makes sense to me. Unfortunately overriding the value here is really hard. I tried adding:

```scss
[id^="modal-image-"] img {
  // Images in tables automatically have a border around them but we don't want
  // that in our modal.
  border: none;
}
```

But of course it's not strong enough to override this incredible selector:

```
table tbody tr.even td img {
  border: 1px solid #e7f0fa;
}
```

tvdeyen:
@graygilmore https://github.com/solidusio/solidus/pull/2143

PR #2106, opened 2017-07-21
fix name for gateways from solidus_gateway

ccarruitero:

PR #2107, opened 2017-07-21
Override model_name.human for PaymentMethod

jhawthorn:
Recently we started using `model_name.human` instead of the raw class name when displaying the names of payment methods.

Unfortunately, if there isn't an i18n key for that specific payment method, model_name.human will fall back to "Payment Method", which is not at all helpful in distinguishing it.

This commit changes `PaymentMethod.model_name` to return a custom `ActiveModel::Name` subclass, which will use exact matches for translations but doesn't fall back to "Payment Method", and instead will
use the "humanized" class name if there is no key.

Fixes #2106

### DIFF
--- a/core/app/models/spree/payment_method.rb
+++ b/core/app/models/spree/payment_method.rb
@@ -37,6 +37,21 @@ module Spree
 
     include Spree::Preferences::StaticallyConfigurable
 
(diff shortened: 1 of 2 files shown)


tvdeyen:
Ah, you already added the milestone. Thanks

Issue #2042, opened 2017-06-26
Add more inline editable tables


tvdeyen:
As a follow up to the `inline-editable-table` introduced in #1580 we should have a look into all tables in the admin and see where this new UI component could fit. Also we need to update the style guide

- [ ] Update style guide and show the new UI component
- [ ] Compile a list with tables in the admin that could make use of this component

Mandily:
@mtomov worked on a PR that added inline editing capabilities to the image variants screen to make it quicker and easier for admins to update fields. I think this could be extended to other tables in the admin so that users could both edit and create simple table rows inline instead of needing to go to other pages. I've broken up the opportunities into:
1. Simply making existing inline editing capabilities available upon hovering of rows. Sometimes we'd be able to remove the edit button from the side if we were also removing a separate edit page, but in the case of pages with a large number of inputs, we'd keep the edit buttons to be able to link the user to that page.
2. Making interactions consistent and moving _save_ or _update_ functions tied to a table row into icons located at the right side of the table, and removing them from underneath the table.
3. Moving _edit_ functionality into a table from a separate page (and thus removing the need for the separate page and for an edit button at the right side of the table)
4. Moving _new_ functionality into a table from a separate page (and thus removing the need for a separate page and for a new button in the header or elsewhere on the page.)

I’ve made a list of tables and what would need to be done to move edit and new functionality into the table [here](https://docs.google.com/spreadsheets/d/13N_inWDx-JCAu6orKDNZbKquL_xMoA6Ky9YH48a9H8Q/edit?usp=sharing)

Let me know if there's any questions/concerns about this.

Issue #1162, opened 2016-05-18
Investigate button styles


Mandily:
I'd like to make buttons in the admin more intuitive by giving them multiple states. Buttons that submit forms and do in page functions should also have a progress spinner inside as well. The purpose of this issue is to discuss what styles need to be created to do this. This will inform a future issue about creating different visual styles (including a spinner button) for the button states.

Once we're done exploring this issue, I will update the mockups in #520 Admin branding to reflect our decisions here.
### Button states (not all buttons have all states)

Each state is distinguished visually different from the others
- Static no action required (example: download list, filter results, save (when nothing has changed in the page) - things that don't **need** to be done on a page)
- Static action required (example: save your edits, cancel edits - when something needs to be done to commit user input)
- Static no action required: hover
- Static action required: hover
- Active 
- Active with spinner
- Disabled

States could look something like this, with button type styling being applied as colour at a later step:
<img width="360" alt="screen shot 2016-05-18 at 3 57 36 pm" src="https://cloud.githubusercontent.com/assets/15271353/15373022/63aa1c42-1d11-11e6-99ea-0fb8bdc8bf3b.png">
### Button types

Should different types look different from each other?
- Filter/search
- Create a new ______
- Edit existing ______
- Save/update
- Cancel
- Clear cache
- Download list
- Edit, clone, and delete **icons** beside tables
### Should different types look different from each other?

There's two options for this question
1. All buttons regardless of their type look the same.
2. Buttons look different based on their type. (ie negative vs positive actions, or any other way we want to break them down)

The thread in #520 Admin branding was discussing using different colours for positive and negative actions, but we were inconclusive on what defined a negative action and we never reached a resolution.

At this point, after considering all of the various states a button could have, I'm leaning towards a cleaner look with buttons using one colour (blue) throughout, with no emphasis on positive, negative or important to notice buttons. We'll use the Spree Blue now, and when we do our brand overhaul we'll change it to the new Blue Steel palette.

<img width="796" alt="screen shot 2016-05-18 at 3 41 18 pm" src="https://cloud.githubusercontent.com/assets/15271353/15372579/1888530c-1d0f-11e6-8973-9a62ab4412df.png">

I'd like to take it to the community - Do you prefer different styles for different types of buttons or consistency throughout? If you prefer multiple styles, how do you define which buttons get which styles?

@davekiss @tvdeyen @Sinetheta @mamhoff @graygilmore

graygilmore:
[...]

PR #2109, opened 2017-07-24
Inline Table Editing

graygilmore:
This PR does a few different things:

1. Refactors the inline editable table row JS to handle a few different scenarios of editing (like removing the save/cancel buttons if no fields have changed).
2. Only apply our custom form styling to elements that aren't already using Bootstrap classes (e.g. `form-control`)
  - A note about this change: this is likely something that we'll want to do throughout the admin so we do not have selector specificity collisions. https://github.com/graygilmore/solidus/commit/8777e5f78ec6c92899a89190d5b03e62ea1f0026
3. Add an API endpoint for updating tax categories and implement the inline editable table to the index page as another example implementation (the first being the product images table)

This PR addresses but does not complete https://github.com/solidusio/solidus/issues/2042. There are quite a few more tables to go which I will be opening up later PRs for.

## TODO

- [x] fix spec failures
- [ ] remove skip auth
- [ ] fix checkbox reset
- [ ] ~~investigate moving this functionality to Backbone~~

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/tables/editable_table_row.js.coffee
+++ b/backend/app/assets/javascripts/spree/backend/views/tables/editable_table_row.js.coffee
@@ -26,8 +26,11 @@ Spree.Views.Tables.EditableTableRow = Backbone.View.extend
       method: 'put'
       success: (response) =>
         @$el.removeClass("editing")
+        $(".has-danger", @$el).removeClass("has-danger")
         @storeValues()
       error: (response) =>
+        for field, errors of response.responseJSON.errors
+          $("input[name*='[#{field}]']", @$el).parent().addClass("has-danger")
         show_flash 'error', response.responseJSON.error
 
   ENTER_KEY: 13
--- a/backend/app/assets/javascripts/spree/backend/views/tables/editable_table_row.js.coffee
+++ b/backend/app/assets/javascripts/spree/backend/views/tables/editable_table_row.js.coffee
@@ -1,6 +1,7 @@
 Spree.Views.Tables.EditableTableRow = Backbone.View.extend
   events:
     "focus :input": "onEdit"
+    "blur :input": "onBlur"
     "click [data-action=save]": "onSave"
     "click [data-action=cancel]": "onCancel"
     'keyup input': 'onKeypress'
@@ -10,6 +11,10 @@ Spree.Views.Tables.EditableTableRow = Backbone.View.extend
 
   onEdit: ->
     @$el.addClass('editing')
+
+  onBlur: (e) ->
+    if $(":input", @$el).serialize() == @originalFormData
+      @onCancel(e)
 
   onCancel: (e) ->
     e.preventDefault()
@@ -43,6 +48,8 @@ Spree.Views.Tables.EditableTableRow = Backbone.View.extend
       when @ESC_KEY then @onCancel(e)
 
   storeValues: ->
+    @originalFormData = $(":input", @$el).serialize()
+
     $(":input", @$el).each ->
       $input = $(this)
       $input.data 'original-value', $input.val()
--- a/backend/app/assets/javascripts/spree/backend/views/tables/editable_table_row.js.coffee
+++ b/backend/app/assets/javascripts/spree/backend/views/tables/editable_table_row.js.coffee
@@ -1,7 +1,6 @@
 Spree.Views.Tables.EditableTableRow = Backbone.View.extend
   events:
-    "select2-open": "onEdit"
-    "focus input": "onEdit"
+    "focus :input": "onEdit"
     "click [data-action=save]": "onSave"
     "click [data-action=cancel]": "onCancel"
     'keyup input': 'onKeypress'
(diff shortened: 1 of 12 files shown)


kennyadsl:
Fantastic work @graygilmore! 

I just noticed a little issue when changing a checkbox input and trying to restore original form data:

ezgif com-video-to-gif


I suppose it is because `$input.val(originalValue)` could not change checkbox value.

Waiting your feedback on this issue before approving. 👍

mtomov:
That's cool! Thanks! Didn't know about the `:input` selector.

`skip_before_action :authenticate_user` seem a bit weird there at the top of the controller

graygilmore:
> I just noticed a little issue when changing a checkbox input and trying to restore original form data

I knew there was still something to do! The checkboxes were finicky, I'll have another look at this.

> That's cool! Thanks! Didn't know about the :input selector.

Neither did I! It's a jQuery specific thing.

> `skip_before_action :authenticate_user` seem a bit weird there at the top of the controller

You're right, I do not believe that should be there. Will fix.

backend/app/assets/javascripts/spree/backend/views/tables/editable_table_row.js.coffee
```diff
@@ -43,3 +46,16 @@ Spree.Views.Tables.EditableTableRow = Backbone.View.extend
...
+
+    $(":input", @$el).each ->
+      $input = $(this)
+      $input.data 'original-value', $input.val()
+
+  restoreOriginalValues: ->
+    $(":input", @$el).each ->
+      $input = $(this)
+      originalValue = $input.data('original-value')
+      $input.val(originalValue).change()
```
tvdeyen: Instead of rolling our own set/reset mechanism we should consider to use a [Backbone model](http://backbonejs.org/#Model) like [in the stock management views](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/stock_management/index_update_forms.coffee)

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -8,27 +8,29 @@ input[type="url"],
...
-  border-radius: $input-border-radius;
-  background: $input-bg;
-  font-size: $font-size-base;
-  line-height: $input-line-height;
-
-  &:focus {
-    border-color: $input-border-focus;
-    outline: 0;
-  }
+  &:not(.form-control) {
```
tvdeyen: 👍 like that!

backend/app/views/spree/admin/tax_categories/index.html.erb
```diff
@@ -34,18 +33,47 @@
...
-         @edit_url = edit_admin_tax_category_path(tax_category)
-         @delete_url = admin_tax_category_path(tax_category)
-    %>
+    <% @tax_categories.each do |tax_category| %>
       <tr id="<%= spree_dom_id tax_category %>" data-hook="tax_row" class="<%= cycle('odd', 'even')%>">
-        <td class="align-center"><%= tax_category.name %></td>
-        <td class="align-center"><%= tax_category.tax_code %></td>
-        <td class="align-center"><%= tax_category.description %></td>
-        <td class="align-center"><%= tax_category.is_default? ? Spree.t(:say_yes) : Spree.t(:say_no) %></td>
+        <td class="align-center">
```
tvdeyen: A Backbone template could be useful here. We use them in other inline tables as well. Combined with a model this should reduce code. See the stock management view for instance.

graygilmore: I think it could reduce some markup but it would drastically increase the amount of JS, no? If that's the direction we'd like to go that's fine but I think the net footprint will be much larger (since each page will require its own JS).

graygilmore: Having looked at this a little bit more I see some opportunity here to perhaps generalize the EditStockItemView and allow other pages to inherit from it. I will investigate this today.

mtomov: I'm strongly in favour if sticking to server side logic wherever possible. 

I think the only thing here to improve is to eliminate a bit of duplication with the `fields_for` by placing `<%= fields_for tax_category do |f| %>` above the `<tr>`

tvdeyen: Ignore me, I thought this whole line gets replaced with fields. The fields are already on screen. 👌 No need for Backbone templates then.

graygilmore:
@Mandily is this the direction you want to see us heading? I know you've referenced other implementations where the admin is able to edit multiple items in the table and then save them all at once. Before we continue down this path I think it's important to make sure we're spending time on something that we actually want!

Mandily:
Allowing admins to edit multiple things at once and worry about saving when they're done - would be a huge win for efficient design. In an ideal world, we have a save button in an easy-to-find spot on every page with states (#1162) that tell a user whether the page needs to be saved or not, instead of multiple save buttons down the page. We could even add a dialog/modal to warn the user if they tried to navigate away from the page without saving it first. Or maybe we even auto save. 

But to get back to the original question - yes, edit multiple items and save once would be huge.

tvdeyen:
> edit multiple items in the table and then save them all at once

This is where Backbone models in conjunction with Backbone collections really shine.

graygilmore:
Closing this to remove some clutter. Given the conversation above I think this area of the system needs to move in a slightly different direction.

PR #2111, opened 2017-07-25
[RFC] Cancel payments refunds correctly

### LINKED ISSUE #2104: [RFC] Canceling payments. The more correct way.
## What happens?

When canceling orders Solidus [calls `cancel!` on each completed payment](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L851-L858).

[`Spree::Payment#cancel!` calls `payment_method.cancel`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment/processing.rb#L81-L84) and handles the response as a void response.

## The problem

That assumption is unfortunate. Lots of payment providers only allow voiding payments if they are not settled yet. Braintree is such a provider. After a transaction has settled you cannot void anymore but have to refund instead.

The `solidus_braintree` and `solidus_paypal_braintree` integrations handle this by checking the status of the transaction first and refund if the transaction cannot be voided anymore. Although this seems correct gateway wise, it is not for Solidus.

For a refunded payment we need to have a `Spree::Refund` object. Handling every response as a void and not as refund - if it should - is wrong and leads to incorrect payment state.

## Solution

The payment method itself should not know how to create refunds for payments. That's why I propose to add a new method to `Spree::PaymentMethod` called `try_void`.

`Spree::Payment#cancel!` will be changed to tell the payment method to try a void. If that succeeds we handle the response as a void - like we do currently -, but if that fails we create a refund with the full amount and "Order canceled" reason. `Spree::Refund` already talks with the gateway and credits accordingly. So, nothing more to do here.

With this change payment methods that do not have voidable states can return a valid void response like before, but payment methods that have voidable states can decide to return `false`.

tvdeyen:
## The problem

When canceling payments Solidus handles all response as a void response.

That assumption is unfortunate. Lots of payment providers only allow voiding payments if they are not settled yet. Braintree is such a provider. After a transaction has settled you cannot void anymore but have to refund instead.

The `solidus_braintree` and `solidus_paypal_braintree` integrations handle this by checking the status of the transaction first and refund if the transaction cannot be voided anymore. Although this seems correct gateway wise, it is not for Solidus.

For a refunded payment we need to have a `Spree::Refund` object. Handling every response as a void and not as refund - if it should - is wrong and leads to incorrect payment state.

## Solution

A new method `Spree::PaymentMethod#try_void`

`Spree::Payment#cancel!` has changed to tell the payment method to try a void. If that succeeds we handle the response as a void - like we do currently -, but if that fails we create a refund with the full amount and a "Order canceled" reason. `Spree::Refund` already talks with the gateway and credits accordingly.

Fixes #2104

### DIFF
--- a/core/app/models/spree/payment/cancellation.rb
+++ b/core/app/models/spree/payment/cancellation.rb
@@ -1,0 +1,60 @@
+module Spree
+  class Payment
+    # Payment cancellation handler
+    #
+    # Cancels a payment by trying to void first and if that fails
+    # creating a refund about the full amount instead.
+    #
+    class Cancellation
+      DEFAULT_REASON = 'Order canceled'.freeze
+
+      attr_reader :reason
+
+      # @param reason [String] (DEFAULT_REASON) -
+      #   The reason used to create the Spree::RefundReason
+      def initialize(reason: DEFAULT_REASON)
+        @reason = reason
+      end
+
+      # Cancels a payment
+      #
+      # Tries to void the payment by asking the payment method to try a void,
+      # if that fails create a refund about the allowed credit amount instead.
+      #
+      # @param payment [Spree::Payment] - the payment that should be canceled
+      #
+      def cancel!(payment)
+        # For payment methods already implemeting `try_void`
+        if try_void_available?(payment.payment_method)
+          if response = payment.payment_method.try_void(payment)
+            payment.send(:handle_void_response, response)
+          else
+            payment.refunds.create!(amount: payment.credit_allowed, reason: refund_reason)
+          end
+        else
+          # For payment methods not yet implemeting `try_void`
+          deprecated_behavior(payment)
+        end
+      end
+
+      private
+
+      def refund_reason
+        Spree::RefundReason.where(name: reason).first_or_create
+      end
+
+      def try_void_available?(payment_method)
+        payment_method.respond_to?(:try_void) &&
+          payment_method.method(:try_void).owner != Spree::PaymentMethod
+      end
+
+      def deprecated_behavior(payment)
+        Spree::Deprecation.warn "#{payment.payment_method.class.name}#cancel is deprecated and will be removed. " \
+          'Please implement a `try_void` method instead that returns a response object if void succeeds ' \
+          'or `false|nil` if not. Solidus will refund the full amount of the payment then.'
+        response = payment.payment_method.cancel(payment.response_code)
+        payment.send(:handle_void_response, response)
+      end
+    end
+  end
+end
--- a/core/app/models/spree/payment/cancellation.rb
+++ b/core/app/models/spree/payment/cancellation.rb
@@ -23,7 +23,7 @@ module Spree
       #
       # @param payment [Spree::Payment] - the payment that should be canceled
       #
-      def cancel!(payment)
+      def cancel(payment)
         # For payment methods already implemeting `try_void`
         if try_void_available?(payment.payment_method)
           if response = payment.payment_method.try_void(payment)
@@ -51,7 +51,7 @@ module Spree
       def deprecated_behavior(payment)
         Spree::Deprecation.warn "#{payment.payment_method.class.name}#cancel is deprecated and will be removed. " \
           'Please implement a `try_void` method instead that returns a response object if void succeeds ' \
-          'or `false|nil` if not. Solidus will refund the full amount of the payment then.'
+          'or `false|nil` if not. Solidus will refund the payment then.'
         response = payment.payment_method.cancel(payment.response_code)
         payment.send(:handle_void_response, response)
       end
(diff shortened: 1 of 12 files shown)


core/app/models/spree/app_configuration.rb
```diff
@@ -352,6 +352,17 @@ def default_payment_builder_class
       @default_payment_builder_class ||= Spree::Wallet::DefaultPaymentBuilder
     end
 
+    # Allows providing your own class for canceling payments.
+    #
+    # @!attribute [rw] payment_canceller
+    # @return [Class] a class instance that responds to `cancel!(payment)`
+    attr_writer :payment_canceller
+    def payment_canceller
+      @payment_canceller ||= Spree::Payment::Cancellation.new(
```
kennyadsl: I just have some doubt about the naming:

I expected this class to be named `Spree::Payment::Canceler` because it is the entity that performs the cancelation. `Spree::Payment::Cancellation` also makes sense but it's more an action, maybe I'd expect the cancelation to happen right when it is initialized, instead this will not performs anything until the `cancel!` method is called. 

As discussed in private with @tvdeyen it is just a personal preference, both makes sense, I just wanted to express mine 😸 .

core/app/models/spree/payment/cancellation.rb
```diff
@@ -0,0 +1,60 @@
...
+    # creating a refund about the full amount instead.
+    #
+    class Cancellation
+      DEFAULT_REASON = 'Order canceled'.freeze
+
+      attr_reader :reason
+
+      # @param reason [String] (DEFAULT_REASON) -
+      #   The reason used to create the Spree::RefundReason
+      def initialize(reason: DEFAULT_REASON)
```
kennyadsl: Maybe this should not be an optional argument since we are passing it on [initialization](https://github.com/solidusio/solidus/pull/2111/files#diff-a9fae1594c22d1ec005ac8501d80d7afR362).

core/app/models/spree/payment/cancellation.rb
```diff
@@ -0,0 +1,60 @@
...
+      #
+      # @param payment [Spree::Payment] - the payment that should be canceled
+      #
+      def cancel!(payment)
+        # For payment methods already implemeting `try_void`
+        if try_void_available?(payment.payment_method)
+          if response = payment.payment_method.try_void(payment)
+            payment.send(:handle_void_response, response)
+          else
+            payment.refunds.create!(amount: payment.amount, reason: refund_reason)
```
jordan-brough: Might need to use the `credit_allowed` method here instead of `amount`?

tvdeyen: Thanks, Jordan. Fixed

core/app/models/spree/payment/cancellation.rb
tvdeyen: I want to be able to initialize without the reason given

tvdeyen:
It is worth mentioning that a payment method that does not implement `try_void` yet will fall back to the current behavior. So, this is safe for existing stores.

tvdeyen:
@jordan-brough I fixed the refund amount. It now uses `credit_allowed`

core/app/models/spree/payment/cancellation.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+
+      # Cancels a payment
+      #
+      # Tries to void the payment by asking the payment method to try a void,
+      # if that fails create a refund about the allowed credit amount instead.
+      #
+      # @param payment [Spree::Payment] - the payment that should be canceled
+      #
+      def cancel!(payment)
```
jhawthorn: This should probably be just `cancel` (without the `!`)

mamhoff:
I am very much for this, but would like a dedicated changelog entry detailing that this is happening and what it means.

core/app/models/spree/payment_method/bogus_credit_card.rb
```diff
@@ -55,7 +55,7 @@ def void(_response_code, _credit_card, _options = {})
       ActiveMerchant::Billing::Response.new(true, 'Bogus Gateway: Forced success', {}, test: true, authorization: '12345')
     end
 
-    def cancel(_response_code)
+    def try_void(_payment)
```
jhawthorn: A YARDoc comment would be helpful here, the return values are a little confusing:
* `ActiveMerchant::Billing::Response` with `true` if the void succeeded
* `ActiveMerchant::Billing::Response` with `false` if the void failed
* Just `false` if it can't be voided at this time.

tvdeyen:
> I am very much for this, but would like a dedicated changelog entry detailing that this is happening and what it means.

It means nothing to existing stores as all existing payment methods do not implement `try_void` yet. Besides a deprecation notice stores will not see any difference to current behaviour. 

But I will add a note about the deprecation

core/app/models/spree/payment/cancellation.rb
tvdeyen: Fixed

core/app/models/spree/payment_method/bogus_credit_card.rb
tvdeyen: added

tvdeyen:
@mamhoff @jhawthorn addressed your feedback

PR #2112, opened 2017-07-25
Default to static preferences

jhawthorn:
We've included static preferences for a while, and they've been default for new stores, but required a call to `use_static_preferences!`

This commit makes the static preferences store the default with no additional configuration. It also adds a way for stores to opt-in to the legacy DB-driven behaviour.

If accepted this should have a prominent changelog entry.


mamhoff:
Will this also affect extension's `Configuration` objects?

jhawthorn:
> Will this also affect extension's Configuration objects?

It will

Issue #2114, opened 2017-07-25
Remove the ability to edit countries and states


graygilmore:
Editing these resources can negatively impact a store. Removing the editing functionality from the admin will reduce the risk of something being accidentally modified.

If a store absolutely requires a change to these they could still do so through a migration or the console but the majority of stores should not be managing these resources directly.

## The Future

Perhaps this could also be a step towards using something like Carmen or removing these items from the DB in another way.

graygilmore:
**Note:** currently working on this.

PR #2117, opened 2017-07-25
Preload assets before backend specs

jhawthorn:
We've been seeing some builds fail with "Request failed to reach server" on the first request.

This happens because we have quite a few assets and it can take some time to compile them. Since this was done on the first request, this sometimes caused poltergeist to error.

This commit gets sprockets-rails to preload the assets before the suite is run (if there is a feature spec included in the run).

This has the added advantage of not including the time to precompile specs as part of the first spec's run time. Running `rspec --profile=10` should now be more useful in showing actually slow specs.

Alternative to #2116

PR #2118, opened 2017-07-25
Remove admin's ability to edit countries and states

graygilmore:
From the commit:

```
Allow admins to edit this part of the system could have dramatic
consequences on a store. Developers will still be able to edit these
resources through a migration or directly in the console for extreme
edge cases.

This may also be a step in the direction of using an externally sourced
service such as Carmen.
```

If anyone can think of any other parts of the system that can also be removed with this change I will happily add to this PR. I think I've found them all but there could be some sneaky ones (like the JS `. observe_field` that @jhawthorn pointed out).

Closes https://github.com/solidusio/solidus/issues/2114

## TODO

- [x] Add changelog
- [x] Fix tests


cbrunsdon:
Spec failure is unrelated, should be fixed by #2117

kennyadsl:
I don't know, I'm not against this but I have some concerns about how much difficult/long it would be to extend solidus in case a store needs to add some editable field into countries/states. For example options specific for country that would need a custom field editable by admin into the country edit form. I know we have other resources that do not have an admin UI but I have the feeling that countries/states are more used.

I understand the concern about making some of those fields not editable though. What about just disable critical fields in the form?

Of course if our plan is to migrate to Carmen in the future this makes perfect sense.

graygilmore:
> A good change. I don’t see a reason to keep - and even add code for- the possibility to list countries and states. What is the benefit of this?

You're right – if we're removing the editing of this we might as well nuke the whole thing. I'd be interested in adding that change if that's something we can agree on.

> For example options specific for country that would need a custom field editable by admin into the country edit form.

I think I would rather see these kinds of modifications in a separate model: `YourApp::CountryModification` – what do you think?

kennyadsl:
@graygilmore we discussed about this one in the core team meeting. We'd go with removing this, including listing, from core for several reasons:

- Solidus will slowly move towards Carmen. It also allow better countries/states I18n;
- The only country's attribute that could be safely changed is `states_required`. It is usually a "dev thing" and it can easily be done via console;

We plan to move this backend functionalities to a separate extension so it can be added back to stores that need those easily.

Thanks!

kennyadsl:
@graygilmore I've created an extension ([solidusio-contrib/solidus_countries_backend](https://github.com/solidusio-contrib/solidus_countries_backend)) that should have all the countries/states UI code. It currently points to this PR [repo/branch](https://github.com/solidusio-contrib/solidus_countries_backend/blob/master/Gemfile#L3) to run specs against Solidus without this code.

graygilmore:
@kennyadsl excellent. I will make the changes to this PR to remove these entirely from the backend.

graygilmore:
This has been updated with a full removal.

kennyadsl:
@graygilmore there's a couple of feature specs failures, due to Locations -> Zones I18n key rename.

graygilmore:
Updated the `zones_spec.rb` and added a changelog entry.

kennyadsl:
The failure is not related, rebasing against master should fix it, Thanks!

kennyadsl:
Also, I'd mention the replacement extension ([`solidusio-contrib/solidus_countries_backend`](https://github.com/solidusio-contrib/solidus_countries_backend)) in both commit message and PR description so that it's super easy to find it if needed, wdyt?

graygilmore:
@kennyadsl added a note about the extension to the changelog entry. Tests now passing after rebasing with master 🎉

PR #2119, opened 2017-07-26
Deprecate pagination in searcher

cbrunsdon:
This deprecates doing pagination in the base searcher and instead pushes the responsibility onto the caller (in our cases the controllers).

We don't gain anything from doing the pagination inside the searcher, and it doesn't have anything to do with the core goal of it (searching).

Also throws the explicit dependencies on kaminari on frontend and backend (as they call kaminari functions).


jhawthorn:
Should solidus_api also receive an explicit dependency as it calls kaminari functions?

cbrunsdon:
Good call hawth, added.

ahwong5:
@cbrunsdon You have omitted to update the home controller as well, therefore the frontend homepage has lost the ability to paginate.

And also when performing Search, it doesn't paginate as well. I'm not sure which file need to update.

Thanks.

PR #2123, opened 2017-07-28
Pill Component

graygilmore:
This PR introduces a new "pill" component. This component is used to display the state of something through six distinct colors:

- Complete
- Active
- Inactive
- Pending
- Warning
- Error

This should make scanning tables significantly easier and provide a slightly more modern feel to the admin (looks great with the new tables in https://github.com/solidusio/solidus/pull/2092).

## TODO

- [x] fix test errors

## Style Guide

screen shot 2017-07-28 at 11 32 11 am

## Orders (before)

screen shot 2017-07-28 at 11 27 07 am

## Orders (after)

screen shot 2017-07-28 at 11 26 11 am

## Promotions (before)

screen shot 2017-07-28 at 11 27 25 am

## Promotions (after)

screen shot 2017-07-28 at 11 27 48 am


Mandily:
I like the content of what you're trying to say in the writeup in the style guide, but could we write it in a clearer way? Maybe something like this: 

"Pill styling adds a second, easily scannable way to recognize status. They're most commonly used in, but not restricted to tables."

tvdeyen:
I like that! 😍 

Maybe it's just me, but shouldn't `complete` be green? I assume shop owners love to see lots of green in their orders table.

backend/app/assets/stylesheets/spree/backend/components/_pills.scss
```diff
@@ -0,0 +1,43 @@
+.pill {
+  display: inline-block;
+  padding: 2px 10px;
+  border-radius: 10px;
+  font-size: 11px;
```
tvdeyen: The `font-size` should have a relative unit.
Also the paddings.

backend/app/views/spree/admin/promotions/index.html.erb
```diff
@@ -85,11 +77,26 @@
...
+          </td>
+          <td class="align-center">
+            <%= promotion.usage_count %>
+          </td>
+          <td class="align-center">
+            <%= promotion.starts_at.to_date.to_s(:short_date) if promotion.starts_at %>
+          </td>
+          <td class="align-center">
+            <%= promotion.expires_at.to_date.to_s(:short_date) if promotion.expires_at %>
+          </td>
```
tvdeyen: You are changing the table far more than just adding the pill component. Could you please do this in a separate commit, or not even PR?

core/config/locales/en.yml
```diff
@@ -170,13 +170,14 @@ en:
...
+        expires_at: End
         name: Name
         path: Path
         per_code_usage_limit: Per Code Usage Limit
         promotion_uses: Promotion Uses
-        starts_at: Start date
-        usage_limit: Overall Usage Limit
+        starts_at: Start
+        status: Status
+        usage_limit: Usage Limit
```
tvdeyen: These changes do not belong in this PR IMO. At least it should be a separate commit.

backend/app/helpers/spree/admin/reimbursements_helper.rb
```diff
@@ -1,14 +0,0 @@
...
-        case reimbursement.reimbursement_status
-        when 'reimbursed' then 'success'
-        when 'pending' then 'notice'
-        when 'errored' then 'error'
-        else raise "unknown reimbursement status: #{reimbursement.reimbursement_status}"
-        end
-      end
-    end
-  end
-end
```
tvdeyen: Could we move this removal in a separate commit after this commit?

backend/app/helpers/spree/admin/stock_transfers_helper.rb
```diff
@@ -17,10 +17,6 @@ def stock_transfer_edit_or_ship_path(stock_transfer)
         end
       end
 
-      def stock_transfer_status(stock_transfer)
-        stock_transfer.closed? ? Spree.t(:closed) : Spree.t(:open)
-      end
```
tvdeyen: this as well, please

backend/app/views/spree/admin/cancellations/index.html.erb
```diff
@@ -34,7 +34,15 @@
...
           </td>
           <td class="inventory-unit-quantity align-center">1</td>
-          <td class="inventory-unit-state align-center"><%= inventory_unit.state %></td>
+          <td class="inventory-unit-state align-center">
+            <span class="pill pill-<%= inventory_unit.state %>">
+              <%= Spree.t(
+                inventory_unit.state,
+                scope: :inventory_states,
+                default: inventory_unit.state.titleize
+              ).titleize %>
```
tvdeyen: I like that this is now translatable, but please do not `titleize` this. This is what should be done in the translation itself, upper and lower case rules are very specific to languages.

Also, this should be a separate commit.

backend/app/views/spree/admin/customer_returns/_reimbursements_table.html.erb
```diff
@@ -20,8 +20,8 @@
         </td>
         <td><%= reimbursement.display_total %></td>
         <td>
-          <span class="state <%= reimbursement_status_color(reimbursement) %>">
-            <%= reimbursement.reimbursement_status %>
+          <span class="pill pill-<%= reimbursement.reimbursement_status %>">
+            <%= reimbursement.reimbursement_status.titleize %>
```
tvdeyen: Please do not transform strings like this, pass it through `I18n` or just leave it like it is.

backend/app/views/spree/admin/customer_returns/_return_item_decision.html.erb
```diff
@@ -39,7 +39,9 @@
           <%= return_item.acceptance_status_errors %>
         </td>
         <td class-"align-center">
-          <%= return_item.reception_status.humanize %>
+          <span class="pill pill-<%= return_item.reception_status %>">
+            <%= return_item.reception_status.humanize %>
```
tvdeyen: We should only use `humanize` if we pass it through `I18n` before and humanize the fallback translation.

```ruby
Spree.t(return_item.reception_status, 
  scope: 'return_item_reception_statuses', 
  default: return_item.reception_status.humanize)
```

backend/app/views/spree/admin/customer_returns/_return_item_selection.html.erb
```diff
@@ -38,7 +38,9 @@
           <%= return_item.display_amount %>
         </td>
         <td class="align-center">
-          <%= return_item.inventory_unit.state %>
+          <span class="pill pill-<%= return_item.inventory_unit.state %>">
+            <%= return_item.inventory_unit.state.titleize %>
```
tvdeyen: Same comment as above. I suggest to make a separate commit (not even a PR) first that changes the presentation via `I18n`.

backend/app/views/spree/admin/orders/confirm/_payments.html.erb
```diff
@@ -30,8 +30,12 @@
...
           </td>
           <td class="align-center">
-            <span class="state <%= payment.state %>">
-              <%= Spree.t(payment.state, scope: :payment_states, default: payment.state.capitalize) %>
+            <span class="pill pill-<%= payment.state %>">
+              <%= Spree.t(
+                payment.state,
+                scope: :payment_states,
+                default: payment.state
+              ).titleize %>
```
tvdeyen: Ditto

backend/app/views/spree/admin/orders/index.html.erb
```diff
@@ -150,11 +150,27 @@
       <tr data-hook="admin_orders_index_rows" class="state-<%= order.state.downcase %> <%= cycle('odd', 'even') %>">
         <td class="align-center"><%= l (@show_only_completed ? order.completed_at : order.created_at).to_date %></td>
         <td class="align-center"><%= link_to order.number, edit_admin_order_path(order) %></td>
-        <td class="align-center"><span class="state <%= order.state.downcase %>"><%= Spree.t("order_state.#{order.state.downcase}") %></span></td>
-        <td class="align-center"><span class="state <%= order.payment_state %>"><%= link_to Spree.t("payment_states.#{order.payment_state}"), admin_order_payments_path(order) if order.payment_state %></span></td>
-          <% if Spree::Order.checkout_step_names.include?(:delivery) %>
-            <td class="align-center"><span class="state <%= order.shipment_state %>"><%= Spree.t("shipment_states.#{order.shipment_state}") if order.shipment_state %></span></td>
+        <td class="align-center">
+          <span class="pill pill-<%= order.state.downcase %>">
+            <%= Spree.t("order_state.#{order.state.downcase}").titleize %>
```
tvdeyen: Same comment because of titleization applies here

backend/app/views/spree/admin/users/items.html.erb
```diff
@@ -52,15 +50,6 @@
...
               <td class="align-center item-total"><%= item.money.to_html %></td>
-              <td class="align-center order-state">
-                <div class="state <%= order.state.downcase %>"><%= Spree.t("order_state.#{order.state.downcase}") %></div>
-                <% if order.payment_state %>
-                  <div class="state <%= order.payment_state %>"><%= link_to Spree.t("payment_states.#{order.payment_state}"), admin_order_payments_path(order) %></div>
-                <% end %>
-                <% if Spree::Order.checkout_step_names.include?(:delivery) && order.shipment_state %>
-                  <div class="state <%= order.shipment_state %>"><%= Spree.t("shipment_states.#{order.shipment_state}") %></div>
-                <% end %>
-              </td>
```
tvdeyen: Should be a separate commit as well

tvdeyen:
The build errors are actually related to the reimbursements helper removal.

Mandily:
re: "Shouldn't complete be green?" 

Green is usually associated with an action - like "go" or "active" in this case. In the case of grey being used for complete, it means that a thing is done and you don't have to worry about doing anything else for it. The colours stand out more than the greys because they're the things still in flux that an admin might want to keep an eye on. The neutral things are the ones they don't have to worry about.

The caveat here being that most if not all of my colour theory is based on North American norms. Maybe these colours mean different things to you.

Make sense?

graygilmore:
> Maybe it's just me, but shouldn't complete be green? I assume shop owners love to see lots of green in their orders table.

No. Green in this case specifies something that is `active`. It gives the feeling of movement and that something is currently happening. A `complete` state is something that needs no further action and is finished, and the grey-blue color better helps signal that.

backend/app/assets/stylesheets/spree/backend/components/_pills.scss
graygilmore: Any particular reason?

backend/app/views/spree/admin/cancellations/index.html.erb
graygilmore: 👍 will make changes to the translations and then `downcase` or somehow slugify for the class names.

backend/app/assets/stylesheets/spree/backend/components/_pills.scss
tvdeyen: paddings not passings. (Sorry iOS)

backend/app/views/spree/admin/cancellations/index.html.erb
tvdeyen: > we're not i18n-ing those!

😆

tvdeyen:
> Re: Re: Green

This makes total sense. Thanks for the explanation.

backend/app/assets/stylesheets/spree/backend/components/_pills.scss
graygilmore: I'm more curious about who the perspective is here. This is a very specific component that I have sized specifically. If users viewing the admin bump up their font size this will scale along with it.

If we decide to change the root font size of the admin we'll need to go through components one by one and figure out what needs to change. I'd normally be happy to use `rem` but our admin root font size is `13px` which I can see changing in the future but I wouldn't want this component to change.

tvdeyen: I like us to move forward to relative units all over the place. That’s the way all frontend frameworks went for a reason. 

I see the problem with the small base font size we have. So, I won’t insist to make this change right now. 

But, could you please make this font size a variable then?

graygilmore:
Wooo tests are green. @tvdeyen I think I've addressed all of your other concerns. I went a little overly verbose with the number of commits but that should make things a little cleaner to understand in the future.

backend/app/views/spree/admin/shared/_order_summary.html.erb
```diff
@@ -1,7 +1,15 @@
 <header id="order_tab_summary" data-hook>
   <dl class="additional-info">
     <dt class="order-state" id="order_status" data-hook><%= Spree.t(:status) %>:</dt>
-    <dd class="order-state"><span class="state <%= @order.state %>"><%= Spree.t(@order.state, scope: :order_state) %></span></dd>
+    <dd class="order-state">
+      <span class="pill pill-<%= @order.state %>">
+        <%= Spree.t(
+          @order.state,
+          scope: :order_state,
+          default: @order.state.humanize
```
jhawthorn: I'd rather not use `humanize` as a default. We want to see where we are missing translations (and avoid the cost of calling the inflector)

core/config/locales/en.yml
```diff
@@ -1645,6 +1645,7 @@ en:
       pending: Pending
       processing: Processing
       void: Void
+      invalid: Invalid
```
graygilmore: Nice! Thanks for that.

Eth3rnit3:
capture d ecran 2017-10-18 a 09 49 22
I have this error message after the update to 2.4 do you have any idea where it might come from?
server-side : 
`ActionView::Template::Error (".state.sold" failed to @extend ".pill-sold".
The selector ".pill-sold" was not found.
Use "@extend .pill-sold !optional" if the extend should be able to fail.
):
    4: <meta content="width=device-width, initial-scale=1.0, maximum-scale=1" name="viewport">
    5: <%== meta_data_tags %>
    6: <%= canonical_tag(current_store.url) %>
    7: <%= favicon_link_tag 'favicon.ico' %>
    8: <%= stylesheet_link_tag 'spree/frontend/all', media: 'screen' %>
    9: <%= csrf_meta_tags %>
   10: <%= javascript_include_tag 'spree/frontend/all' %>`

mamhoff:
@Eth3rnit3 Have you overridden any style sheets in your host app? Compare them with the ones in Solidus 2.4. You're probably not requiring the pills SCSS file.

Eth3rnit3:
yes only the file variables_override. scss I'm going to compare it right now and I'm telling you

Eth3rnit3:
no unfortunately the file is empty in solidus

Eth3rnit3:
you're right, if I add the line: 

> @import 'spree/backend/components/pills'
to my file, it works; -) Thank you

Eth3rnit3:
not really sorry, actually. 
only a css problem that made a mistake earlier

Eth3rnit3:
actually the problem was that I overloaded the $state class in my variables_override file. thank you, maybe we'll see you soon.

PR #2126, opened 2017-07-31
Allow dev mode code reloading of configured classes

jhawthorn:
Previously if a change was made to one of the classes configured in `Spree::Config`, the changes wouldn't apply until the development server was restarted. This is because we stored a reference to the exact class, and it would not use the new version of the class loaded by code reloading.

This PR changes the attributes to store the class `name`, and to `#constantize` it when accessed. This is similar to the change made in #1203 to the arrays of classes for calculators, payment methods, etc.

### DIFF
--- a/core/app/models/spree/app_configuration.rb
+++ b/core/app/models/spree/app_configuration.rb
@@ -256,20 +256,14 @@ module Spree
     preference :can_restrict_stock_management, :boolean, default: false
 
     # searcher_class allows spree extension writers to provide their own Search class
-    attr_writer :searcher_class
-    def searcher_class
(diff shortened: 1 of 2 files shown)


kennyadsl:
I don't know this topic very well but I'm curious why we didn't used the Rails `class_attribute` here. It should also have a [`default`](https://github.com/rails/rails/pull/29270) option now.

jhawthorn:
> I don't know this topic very well but I'm curious why we didn't used the Rails class_attribute here. It should also have a default option now.

The purpose of this is to call `constantize` whenever accessed

PR #2127, opened 2017-07-31
Remove Legacy Admin Layout

graygilmore:
This removes the legacy admin layout and gives us half-decent responsiveness due to the nature of using Bootstrap classes. Previously, using `.container` our content would sit at a maximum width of `960px` and shrink down to awkward sizes depending on the browser size.

This PR also finally removes the Skeleton Grid CSS from the admin. If stores are still relying on it in order to make overrides to their stores they can pull that CSS directly into their app.

**Oooo, fancy!**

responsive

## Before at awkward breakpoint

screen shot 2017-07-31 at 3 37 10 pm

## After

screen shot 2017-07-31 at 3 37 24 pm


graygilmore:
Looks like the order page is borked. Will fix.

graygilmore:
Fixed the order page. The bug was from a few releases ago but it went unnoticed because none of the pages using the new layout also had a sidebar.

mtomov:
That's great! Thank you! 

Happy to see `@admin_layout` go.

Issue #2099, opened 2017-07-19
Canceled orders in `complete` scope


brchristian:
We have been experimenting with allowing users to cancel completed orders if they haven’t been fulfilled yet, and one of the things that has stood out to me as unintuitive is that the `Order.complete` scope includes cancelled orders.

This makes sense in the context of the admin `Orders#index` page, where it is useful for backend administrators to see canceled as well as completed orders, but it causes problems elsewhere.

For instance, the Sales Total report will still include revenue from canceled orders. This seems incorrect.

My proposal here would be to do something like the following:

1. change the `Order.complete` scope to include the additional restriction of `.where.not(status: 'canceled')` by default
2. create an `Order.canceled` scope to have a canonical way to select canceled orders (presumably it would just check for `status: 'canceled'` but maybe we also want to check for `.not(completed_at: nil)` or something)
3. create an `Order.complete.with_canceled` or `Order.complete_with_canceled` to include both fully completed orders as well as canceled ones. I see this mostly being useful on the backend in views like `Orders#index`.

If the above (or some variation on the above) seems like a useful approach to the Solidus team / community, let me know and I am more than happy to prepare a PR.

bbuchalter:
Hey @brchristian - you raise an interesting point. Thanks for the issue submission. It's great to see you still hacking on this stuff; I remember when we did a bit of work together for `spree_tax_cloud` a long time ago.

Anyway, from my perspective, this issue is mostly one of semantics. From a certain perspective, for an order to be canceled, it must have be _complete_. If you only want to find orders that are completed and not canceled, there is a `by_state` scope you can use to accomplish what you've suggested in your second proposal.

As for the issue with the Sales Report, I think that's a very reasonable fix that would deserve it's own issue and a PR would likely be very welcome.

Just my two cents.

brchristian:
@bbuchalter Agreed that it’s a semantics issue!

As you say, yes, the alternative and more minimal fix is just a PR to restrict the default Sales Total scoping in the Reports controller.

I can see the argument in either direction and am happy to go wherever majority opinion lies. Curious to kind of take the temperature of the community here and see what people think makes most sense.

brchristian:
Closing here and linking to the recent discussion at https://github.com/solidusio/solidus/pull/2131

PR #2131, opened 2017-08-02
Exclude canceled orders from sales report

brchristian:
As I mentioned in issue #2099, by default the Sales Total report does not exclude canceled orders from the sales totals. This is the "minimum viable PR" to fix that issue.

I’m open to further discussion from the Solidus community around the following:

1. Should we make proper `canceled` and `not_canceled` scopes for the `Order` model, to go along with frequently used scopes like `complete`?

2. Should the `complete` scope by _default_ exclude canceled orders unless explicitly called for `complete_including_canceled`, the way that `Product#variants` by default excludes the master variant unless and we specify `variants_including_master`, or how various scopes exclude deleted objects unless we use `with_deleted` scopes?

If "yes" to either (1) or (2), chime in on this thread and I’m happy to whip up another easy PR.

### DIFF
--- a/backend/app/controllers/spree/admin/reports_controller.rb
+++ b/backend/app/controllers/spree/admin/reports_controller.rb
@@ -28,7 +28,7 @@ module Spree
       def sales_total
         params[:q] = search_params
 
-        @search = Order.complete.ransack(params[:q])
+        @search = Order.complete.not_canceled.ransack(params[:q])
         @orders = @search.result
 
         @totals = {}
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -175,6 +175,14 @@ module Spree
       where(completed_at: nil)
     end
 
+    def self.canceled
+      where(state: 'canceled')
+    end
+
+    def self.not_canceled
+      where.not(state: 'canceled')
+    end
+
     # Use this method in other gems that wish to register their own custom logic
     # that should be called after Order#update
     def self.register_update_hook(hook)
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -61,6 +61,14 @@ RSpec.describe Spree::Order, type: :model do
         expect{ subject }.to change{ order.can_cancel? }.from(true).to(false)
         expect(order).to be_canceled
       end
+
+      it "places the order into the canceled scope" do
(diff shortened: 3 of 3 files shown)


mamhoff:
> 1. Should we make proper canceled and not_canceled scopes for the Order model, to go along with frequently used scopes like complete?

Yes, that would be nice. You can add those to this PR or open a new one.

> 2. Should the complete scope by default exclude canceled orders unless explicitly called for complete_including_canceled, the way that Product#variants by default excludes the master variant unless and we specify variants_including_master, or how various scopes exclude deleted objects unless we use with_deleted scopes?

No, `complete` just means that the order has, at some point in the past, been completed. In the core team, we were discussing we might be missing a name for a scope that is `completed.not_canceled`, but we did not come up with one.

brchristian:
@mamhoff That all sounds good to me. I’ve closed the open issue at #2099 and am including here two additional PRs: one to add the `canceled` and `not_canceled` scopes, and another to refactor the first commit in this branch to use that new syntax.

brchristian:
@mamhoff I think I’ve addressed your notes here, so let me know if there’s anything else I can add to this!

mamhoff:
@brchristian any news on this? Will you be able to add a request spec?

brchristian:
@mamhoff Yes, happy to add a spec! Will try to get to that before the end of this week.

brchristian:
@mamhoff Actually, just added the specs now. That should take care of it. Let me know if you want anything further and I’m happy to add it; otherwise I think we are probably good to merge.

mamhoff:
I actually do. I think the model-level changes (including the spec, which is perfect) should go into one commit, and then we should have another commit that changes the controller to use the new scope. 

Sorry to be so picky, but we'll all be thankful when we go back to the history of this. :)

brchristian:
@mamhoff We good to go with this one?

brchristian:
Hi @mamhoff just wanted to ping and see if we are good to go, would it be possible to merge this? We have been adding this in our app through a decorator and it’d be nice to remove that and simply be able to use the latest `master`.

I’ve just rebased again if that’s helpful!

mamhoff:
@brchristian thanks for following up! Yes, in my opinion this is good to go, but needs another pair of eyes from the core team.

brchristian:
Hi @tvdeyen are we good to merge this? Lmk if there is anything more we need and I am happy to hop on it. Cheers!

PR #2135, opened 2017-08-08
Enable custom shipping promotions via config.spree.promotions.shipping_actions

jordan-brough:
This updates the shipping promotion handler so that it will look for promotion
action types via `Rails.application.config.spree.promotions.shipping_actions`,
instead of only looking for actions of type
`Spree::Promotion::Actions::FreeShipping`.

This should allow applications to create their own shipping promotion actions.

### DIFF
--- a/core/app/models/spree/promotion_handler/shipping.rb
+++ b/core/app/models/spree/promotion_handler/shipping.rb
@@ -1,7 +1,7 @@
 module Spree
   module PromotionHandler
     # Used for activating promotions with shipping rules
-    class FreeShipping
+    class Shipping
       attr_reader :order
       attr_accessor :error, :success
 
@@ -29,7 +29,7 @@ module Spree
       end
 
       def automatic_promotions
-        @automatic_promotions ||= active_free_shipping_promotions.
+        @automatic_promotions ||= active_shipping_promotions.
           where(apply_automatically: true).
           to_a.
           uniq
@@ -39,20 +39,16 @@ module Spree
         @connected_promotions ||= order.order_promotions.
           joins(:promotion).
           includes(:promotion).
-          merge(active_free_shipping_promotions).
+          merge(active_shipping_promotions).
           to_a.
           uniq
       end
 
-      def active_free_shipping_promotions
+      def active_shipping_promotions
         Spree::Promotion.all.
           active.
           joins(:promotion_actions).
-          merge(
-            Spree::PromotionAction.of_type(
-              Spree::Promotion::Actions::FreeShipping
-            )
-          )
+          merge(Spree::PromotionAction.shipping)
       end
     end
   end
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -105,6 +105,12 @@ module Spree
         ]
       end
 
+      initializer 'spree.promo.register.promotions.shipping_actions', before: :load_config_initializers do |app|
+        app.config.spree.promotions.shipping_actions = %w[
+          Spree::Promotion::Actions::FreeShipping
+        ]
+      end
+
       # Filter sensitive information during logging
       initializer "spree.params.filter", before: :load_config_initializers do |app|
         app.config.filter_parameters += [
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -536,10 +536,11 @@ module Spree
       end
     end
 
-    def apply_free_shipping_promotions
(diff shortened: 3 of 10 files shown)


core/lib/spree/core/engine.rb
```diff
@@ -105,6 +105,12 @@ class Engine < ::Rails::Engine
         ]
       end
 
+      initializer 'spree.promo.register.promotions.shipping_actions', before: :load_config_initializers do |app|
```
kennyadsl: do we need to keep [`Spree::Promotion::Actions::FreeShipping` into `spree.promo.register.promotions.actions`](https://github.com/jordan-brough/solidus/blob/4db04cac8946c0ca2c041b9302e2f478624d1d93/core/lib/spree/core/engine.rb#L104) as well?

jordan-brough: At the moment we do, unfortunately.  I'm not sure if it's straightforward to combine them either, since you might remove a shipping promotion from `promotions.actions` so that they don't show up in the admin UI, but you'd still want existing orders to apply existing free shipping promotions?

Very open to ideas on this.

core/app/models/spree/promotion_handler/shipping.rb
```diff
@@ -1,7 +1,7 @@
 module Spree
   module PromotionHandler
     # Used for activating promotions with shipping rules
-    class FreeShipping
+    class Shipping
```
jhawthorn: We probably want to keep `Spree::PromotionHandler::FreeShipping` as a deprecated constant.

This is probably a good case for [ActiveSupport::Deprecation::DeprecatedConstantProxy](http://api.rubyonrails.org/classes/ActiveSupport/Deprecation/DeprecatedConstantProxy.html). I believe we re-add a `promotion_handler/free_shipping.rb` with

```
module Spree
  module PromotionHandler
    ActiveSupport::Deprecation::DeprecatedConstantProxy.new('FreeShipping', 'Shipping', Spree::Deprecation)
  end
end
```

jordan-brough: @jhawthorn nice, I hadn't seen that before.  Added.

core/lib/spree/core/engine.rb
kennyadsl: Thanks for the clarification. I have no good ideas other than the one you proposed. Anyway at the moment I can't figure out a scenario when removing something from that `actions` list would break existing orders and promotions. I mean, isn't the same if we remove another action from that `actions` list, like `Spree::Promotion::Actions::CreateAdjustment`?

jordan-brough: @kennyadsl here's the case that I'm thinking of:

- Developer creates a `OurGreatShippingAction` promo action and adds it to the list
- Admin creates a promo using that action
- The store decides not to do any _more_ of those promo actions and a developer removes the `OurGreatShippingAction` from the list so it doesn't show up in the admin for _new_ promos anymore.
- However, if this *also* removes the action from `Spree::PromotionHandler::Shipping#active_shipping_promotions` then that means that the existing promo will cease to work as well.

So we need 2 lists -- one list that indicates what actions should be available in the admin, and one list that indicates what promotions to consider in `Spree::PromotionHandler::Shipping#active_shipping_promotions`.

The difference from other promo actions is that other promo actions don't need a special invocation to work correctly -- i.e. other actions don't need a `before_transition from: :delivery, do: :apply_shipping_promotions` in order to work correctly.

jordan-brough: A couple other possibilities:

A) Just decide we don't care about the scenario above

B) Figure out how to make shipping promotions work correctly without the `before_transition from: :delivery, do: :apply_shipping_promotions` transition (thus eliminating the need to identify them in `Spree::PromotionHandler::Shipping`).

C) Think of another way to identify Shipping promo actions in `Spree::PromotionHandler::Shipping`.  Perhaps all Shipping promo actions must inherit from a base class?  Or have a `is_shipping` field on the action?

It might be worth just going w/ what's in this PR and iterating on it?

kennyadsl: Got it, thanks for taking the time to explain this better. IMO it's reasonable to care about the above scenario. As you said, there could be promotions that would stop working if a developer removes that class from the list before all promos with that actions are expired or disabled.

I'm fine merging and iterating on this, maybe with time we'll be able to simplify this logic a lot since it seems like we just want to automatically try to apply some promotions after a specific checkout step (probably I'm too optimistic and it's much more complex than this 😄 ).

PR #2136, opened 2017-08-08
Fix error when passing coupon_code to api/checkouts#update

jhawthorn:
Previously, when given a `coupon_code`, this action would always error. `handler` wasn't the `PromotionHandler`, but the return value of `apply` (a hash). This has been the case since Spree 2.2.

The error wasn't noticed in specs because of the [rescue_from in BaseController](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/base_controller.rb#L26), which I am working on removing.

This also makes the return status of a _real_ coupon application error (which previously was inaccessible) 422.


jhawthorn:
Merged as part of #2139

PR #2137, opened 2017-08-08
Quick Switch

graygilmore:
This component allows power user admins to navigate directly to a resource if they already know the primary key for it (an order's number, for example). This is heavily inspired both by how developers navigate to particular files or the Spotlight in macOS (though both of those use fuzzy searching which makes them much more powerful).

Example:

quick-switch

## TODO

- [x] Allow this component to be extended – this might not be necessary as a first step but I imagine stores would want a way to be able to hook into this.
- [x] Add a visual indication that a resource has been found and that the admin is being redirected.
- [x] Explore new way of making this extendable and reducing controller complexity (see WIP commit)

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/quick_switch.js
+++ b/backend/app/assets/javascripts/spree/backend/views/quick_switch.js
@@ -1,0 +1,71 @@
+Spree.Views.QuickSwitch = Backbone.View.extend({
+  events: {
+    "shown.bs.modal": "onShow",
+    "hidden.bs.modal": "onHide",
+    "submit form": "onSubmit"
+  },
+
+  initialize: function() {
+    this.$el = $(this.el);
+    this.$dialog = $("[data-js='quick-switch-dialog']", this.$el);
+    this.$form = $("form", this.$el);
+    this.$input = $("[data-js='quick-switch-input']", this.$el);
+    this.$tooltip = $("[data-js='quick-switch-tooltip']", this.$el);
+		this.render();
+  },
+
+  triggerShortcut: function() {
+    this.onShortcut();
+  },
+
+  onShortcut: function() {
+    this.$el.modal("toggle");
+  },
+
+  onShow: function() {
+    this.$input.focus();
+  },
+
+  onHide: function() {
+    this.$input.val("");
+    $(".alert", this.$dialog).remove();
+  },
+
+  onSubmit: function(e) {
+    e.preventDefault();
+    this.toggleSearchClass();
+    this.sendQuery();
+  },
+
+  toggleSearchClass: function() {
+    this.$form.toggleClass("searching");
+  },
+
+  sendQuery: function() {
+    var view = this;
+    Spree.ajax({
+      dataType: "json",
+      url: this.$form.attr("action"),
+      type: "GET",
+      data: this.$form.serializeArray(),
+      global: false, // disable the backend's default loading messaging
+      success: function(response) {
+        if(response.redirect_url) {
+          window.location = response.redirect_url;
+        }
+      },
+      error: function(response) {
+        var message = JSON.parse(response.responseText).message;
+        view.toggleSearchClass();
+        if($(".alert", view.$el).length) {
+          $(".alert", view.$el).text(message);
+        } else {
+          view.$dialog.append("<div class='alert alert-warning'>" + message + "</div>");
+        }
+      }
+    });
+  },
+
+  render: function() {
+  }
+});
--- a/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
+++ b/backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
@@ -67,6 +67,7 @@ $border-radius:          3px !default;
 $input-color:                    $color-3 !default;
 $input-border-color:             $color-border !default;
 $input-focus-border-color:       $color-2 !default;
+$input-placeholder-color:        #aaa !default;
 
 $custom-select-focus-border-color: $input-focus-border-color !default;
 
--- a/backend/app/assets/stylesheets/spree/backend/components/_quick-switch.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_quick-switch.scss
@@ -1,0 +1,56 @@
+.quick-switch-form {
+  position: relative;
+  display: flex;
+}
+
+.quick-switch-query-input {
+  flex: 1;
+}
+
+.quick-switch-button,
+.quick-switch-spinner {
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+}
+
+.quick-switch-spinner {
+  position: absolute;
+  top: 50%;
+  right: 8px;
+  width: 20px;
+  height: 20px;
+  margin-top: -10px;
+  border-width: 2px;
+  border-left-color: $color-3;
+  background: $color-1;
+
+  .quick-switch-form:not(.searching) & {
+    display: none;
+  }
+}
+
+.quick-switch-help {
+  font-size: 0.8rem;
+}
+
+.quick-switch-help button {
+  padding-left: 0;
+  box-shadow: none;
+  font-size: inherit;
+  cursor: pointer;
+
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
+.quick-switch-help ul {
+  margin: 0.5rem 0 0;
+  padding-left: 2rem;
+}
+
+.quick-switch-dialog .alert {
+  margin-top: 1rem;
+}
--- a/backend/app/controllers/spree/admin/quick_switch_controller.rb
+++ b/backend/app/controllers/spree/admin/quick_switch_controller.rb
@@ -1,0 +1,48 @@
+module Spree
+  module Admin
+    class QuickSwitchController < Spree::Admin::BaseController
+      layout false
+
+      def find_object
+        quick_switch_item = Spree::Backend::Config.quick_switch_items.detect do |item|
+          item.search_triggers.include? searched_key.to_sym
+        end
+
+        if quick_switch_item
+          if object = quick_switch_item.finder.call(searched_value)
+            render(
+              json: {
+                redirect_url: quick_switch_item.url.call(object)
+              },
+              status: :ok
+            )
+          else
+            render(
+              json: {
+                message: quick_switch_item.not_found_text(searched_value)
+              },
+              status: :not_found
+            )
+          end
+        else
+          render(
+            json: {
+              message: I18n.t("invalid_query", scope: "spree.quick_switch")
+            },
+            status: :bad_request
+          )
+        end
+      end
+
+      private
+
+      def searched_key
+        params[:quick_switch_query].split(" ")[0]
+      end
+
+      def searched_value
+        params[:quick_switch_query].split(" ")[1]
+      end
+    end
+  end
+end
--- a/backend/app/models/spree/backend_configuration.rb
+++ b/backend/app/models/spree/backend_configuration.rb
@@ -84,5 +84,85 @@ module Spree
         )
       ]
     end
+
+    # New quick switch items can be added to the admin:
+    #
+    # Spree::Backend::Config.configure do |config|
+    #   config.quick_switch_items << config.class::QuickSwitchItem.new(
+    #     search_triggers: [:o, :order],
+    #     finder: ->(searched_value) do
+    #       Spree::Order.find_by(number: searched_value)
+    #     end,
+    #     url: ->(order) do
+    #       Spree::Core::Engine.routes.url_helpers.edit_admin_order_path(order)
+    #     end,
+    #     help_text_key: :order_help,
+    #     not_found_text_key: :order_not_found
+    #   )
+    # end
+    #
+    # @!attribute quick_switch_items
+    #   @return [Array<Spree::QuickSwitchItem>]
+    attr_writer :quick_switch_items
+
+    # Return the quick switch items that administrators can search by
+    #
+    # @api public
+    # @return [Array<Spree::QuickSwitchItem>]
+    def quick_switch_items
+      @quick_switch_items ||= [
+        Spree::QuickSwitchItem.new(
+          search_triggers: [:o, :order],
+          finder: ->(searched_value) do
+            Spree::Order.find_by(number: searched_value)
+          end,
+          url: ->(order) do
+            Spree::Core::Engine.routes.url_helpers.edit_admin_order_path(order)
+          end,
+          help_text_key: :order_help,
+          not_found_text_key: :order_not_found
+        ),
+        Spree::QuickSwitchItem.new(
+          search_triggers: [:s, :shipment],
+          finder: ->(searched_value) do
+            Spree::Shipment.find_by(number: searched_value)
+          end,
+          url: ->(shipment) do
+            Spree::Core::Engine.routes.url_helpers.
+              edit_admin_order_path(shipment.order)
+          end,
+          help_text_key: :shipment_help,
+          not_found_text_key: :shipment_not_found
+        ),
+        Spree::QuickSwitchItem.new(
+          search_triggers: [:email, :u, :user],
+          finder: ->(searched_value) do
+            Spree.user_class.find_by(email: searched_value)
+          end,
+          url: ->(user) do
+            Spree::Core::Engine.routes.url_helpers.edit_admin_user_path(user)
+          end,
+          help_text_key: :user_help,
+          not_found_text_key: :user_not_found
+        ),
+        Spree::QuickSwitchItem.new(
+          search_triggers: [:p, :product, :sku, :v, :variant],
+          finder: ->(searched_value) do
+            Spree::Variant.find_by(sku: searched_value)
+          end,
+          url: ->(variant) do
+            if variant.is_master?
+              Spree::Core::Engine.routes.url_helpers.
+                edit_admin_product_path(variant.product)
+            else
+              Spree::Core::Engine.routes.url_helpers.
+                edit_admin_product_variant_path(variant.product, variant)
+            end
+          end,
+          help_text_key: :variant_help,
+          not_found_text_key: :variant_not_found
+        )
+      ]
+    end
   end
 end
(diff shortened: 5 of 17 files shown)


backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
```diff
@@ -68,6 +68,7 @@ $border-radius:          3px !default;
 $input-color:                    $color-3 !default;
 $input-border-color:             $color-border !default;
 $input-border-focus:             $color-2 !default;
+$input-color-placeholder:        lighten($gray-light, 20) !default;
```
graygilmore: Cannot use `$gray-light` here since it's no longer defined until later on... this is tricky.

graygilmore:
Needs https://github.com/solidusio/solidus/pull/2158 in order to effectively use Bootstrap's `btn-link` styling.

backend/app/controllers/spree/admin/quick_switch_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      def searched_key
+        params[:quick_switch_query].split(" ")[0]
+      end
+
+      def searched_value
+        params[:quick_switch_query].split(" ")[1]
+      end
+
+      def find_and_redirect_to_order
```
graygilmore: Now that we have a `QuickSwitchItem` class I wonder if that's where these methods should be defined.

graygilmore:
Here's a really simple example of how somebody would extend this in their own app:

```rb
  config.quick_switch_items << Spree::QuickSwitchItem.new(
    [:promo],
    :find_and_redirect_to_promotion,
    "promo PROMOTION_CODE"
  )
end
```

```rb
# app/controllers/spree/admin/quick_switch_controller_decorator.rb
Spree::Admin::QuickSwitchController.class_eval do
  private

  def find_and_redirect_to_promotion
    if promotion = Spree::Promotion.with_coupon_code(searched_value)
      redirect_to_url spree.edit_admin_promotion_path(promotion)
    else
      not_found("Could not find promotion with code #{searched_value}")
    end
  end
end
```
Does this seem suitable? This is my first time getting into some arguably meta programming with `method(quick_switch_item.method).call`.

backend/app/models/spree/backend_configuration.rb
```diff
@@ -62,61 +29,109 @@ def initialize(
...
+    # Return the quick switch items that administrators can search by
+    #
+    # @api public
+    # @return [Array<Spree::QuickSwitchItem>]
+    def quick_switch_items
+      @quick_switch_items ||= [
+        Spree::QuickSwitchItem.new(
+          [:o, :order],
+          :find_and_redirect_to_order,
+          Spree.t("quick_switch.help_actions.order")
```
graygilmore: Looks like these translations are not quite available when this code gets run. Does anyone know a way to get around this?

tvdeyen: You could only pass the translation key here and translate in the output. We also could think about some sensible defaults. Ie. An order quick switch item has an order translation key (assuming a proper scope, like above).

This also would be great for the methods that get called. I don’t like the “you need to class eval” kind of extension point. A callable object (lambda) or assumptions we make would be nice. Passing the class name of the model that gets called and an array of attributes that gets searched for instance.

backend/app/controllers/spree/admin/quick_switch_controller.rb
tvdeyen: I would rather let us make some assumptions and a way to let users define the class and attributes. Something like

```ruby
QuickSwitchItem.new(
  model_class_name: 'Spree::Image',
  attributes: [:alt_text, :title]
)
```

On the quick switch controller we would then just search the value on that model class' attributes

tvdeyen:
I would like us to not promote `class_eval` as a way to extend this functionality, but provide an API instead. Example above. The quick switch item object could take a model class name and an array of attributes. This should be enough to implement the search on our side instead.

backend/app/models/spree/quick_switch_item.rb
```diff
@@ -0,0 +1,22 @@
...
+    # @param help_text [String] A helpful guide for admins to understand how to
+    #   use this QuickSwitchItem
+    #
+    # @example
+    #   Spree::QuickSwitchItem.new(
+    #     [:o, :order],
+    #     :find_and_redirect_to_order,
+    #     "o ORDER_NUMBER"
+    #   )
+    def initialize(search_terms, method, help_text)
```
tvdeyen: Please use named arguments, so the api is more obvious for users.

backend/app/models/spree/quick_switch_item.rb
```diff
@@ -0,0 +1,22 @@
...
+      @search_terms = search_terms
```
tvdeyen: isn't `search_trigger` a more understandable name instead of `search_term`?

backend/app/models/spree/quick_switch_item.rb
```diff
@@ -0,0 +1,22 @@
...
+      @method = method
```
tvdeyen: instead of `method` I would like to use a Ransack `search_query` (i.e.: `promo_code_cont`)

backend/app/models/spree/quick_switch_item.rb
```diff
@@ -0,0 +1,22 @@
...
+      @help_text = help_text
+    end
```
tvdeyen: Add a method for the help text here.

```rb
def help_text
  Spree.t(@help_key, scope: 'quick_switch.help_texts')
end
```

tvdeyen:
Or we could just use Ransack.

To take your example from above, I think of something like:

```rb
  config.quick_switch_items << Spree::QuickSwitchItem.new(
    search_triggers: [:p, :promo],
    model_class_name: 'Spree::Promotion'
    search_query: :coupon_code_eq, # a ransack search query 
    help_text:  "promo PROMOTION_CODE"
  )
end
```

```rb
# app/controllers/spree/admin/quick_switch_controller.rb
Spree::Admin::QuickSwitchController.class_eval do
  def find_object
    quick_switch_item = Spree::Backend::Config.quick_switch_items.detect do |item|
      item.search_triggers.include? searched_key.to_sym
    end
    if quick_switch_item && search_record
      redirect_to_url spree.edit_admin_promotion_path(search_record)
    else
      not_found("Could not find promotion with code #{searched_value}")
    end
  end

  private

  def search_record
      model_class = quick_switch_item.model_class_name.constantize
      model_class.ransack(quick_switch_item.search_query => searched_value).result.first
  end
end
```

Alternatively we could use `find_by` as we mostly want to find exactly one record, but Ransack would be more flexible for users to extend this behavior.

WDYT?

backend/app/controllers/spree/admin/quick_switch_controller.rb
```diff
@@ -0,0 +1,94 @@
...
+        quick_switch_item = Spree::Backend::Config.quick_switch_items.detect do |item|
+          item.search_terms.include? searched_key.to_sym
+        end
+
+        if quick_switch_item
+          method(quick_switch_item.method).call
+        else
+          message = Spree.t("quick_switch.invalid_query")
+          respond_to do |format|
+            format.html { redirect_back(fallback_location: spree.admin_path, flash: { error: message }) }
```
tvdeyen: Do we really need the html response? We are only consuming the json response

backend/config/routes.rb
```diff
@@ -2,6 +2,7 @@
   namespace :admin do
     get '/search/users', to: "search#users", as: :search_users
     get '/search/products', to: "search#products", as: :search_products
+    post "/quick-switch", to: "quick_switch#find_object", as: :quick_switch
```
tvdeyen: I think `get` would be more appropriate then `post`.

backend/app/models/spree/quick_switch_item.rb
graygilmore: Yea, @jhawthorn pointed out that I had set this up a little wrong and should instead of storing the key. 👍

graygilmore:
> Or we could just use Ransack.

I really like this simplification. This was exactly the kind of feedback I was hoping for!

backend/config/routes.rb
graygilmore: The only issue that I have is that it returns a redirect. It seemed odd to me to have a `GET` change your location but I'll gladly change it if it makes more sense.

backend/app/controllers/spree/admin/quick_switch_controller.rb
graygilmore: I did this so people could still hit this without JS (also what the controller's spec is doing). I can remove it and instead switch to a feature spec only.

backend/config/routes.rb
tvdeyen: Doesn't `get` requests in regular search forms change location all the time? 

But yeah, semantics `¯\_(ツ)_/¯`

As this is an Ajax request I actually don't care this much.

backend/app/controllers/spree/admin/quick_switch_controller.rb
tvdeyen: I get the test argument. But it is very unlikely that someone will use this html endpoint ever. Switching to a request spec should be enough and assuming that the response gives you the `redirect_url`. I don't think we need a fully blown js driven feature spec here.

Mandily:
I really like the spinner that replaces the search icon and the animations for extending the "how do I use this?" and the errors. One small UX request - Can we put a hover state and cursor: pointer on that link? I looked around a bit and couldn't find links anywhere else. This might be a new style worth putting in the style guide.

backend/app/assets/javascripts/spree/backend/views/quick_switch.js
```diff
@@ -0,0 +1,71 @@
...
+    this.$input = $("[data-js='quick-switch-input']", this.$el);
+    this.$tooltip = $("[data-js='quick-switch-tooltip']", this.$el);
+		this.render();
+  },
+
+  triggerShortcut: function() {
+    this.onShortcut();
+  },
+
+  onShortcut: function() {
```
tvdeyen: I didn’t find any usage of this function besides from `triggerShortcut`. Could we combine these?

backend/app/models/spree/backend_configuration.rb
```diff
@@ -62,61 +29,135 @@ def initialize(
...
           url: :admin_users_path
         )
       ]
     end
+
+    # New quick switch items can be added to the admin:
+    #
+    # Spree::Backend::Config.configure do |config|
+    #   config.quick_switch_items << config.class::QuickSwitchItem.new(
+    #     [:o, :order],
```
tvdeyen: We should update the rdoc

backend/app/models/spree/backend_configuration.rb
```diff
@@ -62,61 +29,135 @@ def initialize(
...
+
+    # Return the quick switch items that administrators can search by
+    #
+    # @api public
+    # @return [Array<Spree::QuickSwitchItem>]
+    def quick_switch_items
+      @quick_switch_items ||= [
+        Spree::QuickSwitchItem.new(
+          search_triggers: [:o, :order],
+          finder: -> (searched_value) do
```
tvdeyen: I think hound is right

backend/app/models/spree/backend_configuration.rb
```diff
@@ -62,61 +29,135 @@ def initialize(
...
+        Spree::QuickSwitchItem.new(
+          search_triggers: [:p, :product, :sku, :v, :variant],
+          finder: -> (searched_value) do
+            return Spree::Variant.find_by(sku: searched_value)
+          end,
+          url: -> (variant) do
+            if variant.is_master?
+              return Spree::Core::Engine.routes.url_helpers.edit_admin_product_path(variant.product)
+            else
+              return Spree::Core::Engine.routes.url_helpers.edit_admin_product_variant_path(
```
tvdeyen: Do we actually need these `return`s in all the lambdas?

backend/app/models/spree/quick_switch_item.rb
```diff
@@ -0,0 +1,47 @@
...
+    #     search_triggers: [:o, :order],
+    #     finder: -> (searched_value) do
+    #       return Spree::Order.find_by(number: searched_value)
+    #     end,
+    #     url: -> (order) do
+    #       return Spree::Core::Engine.routes.url_helpers.edit_admin_order_path(order)
+    #     end,
+    #     help_text_key: :order_help,
+    #     not_found_text_key: :order_not_found
+    #   )
```
tvdeyen: I love this API

backend/config/routes.rb
kennyadsl: I also think `GET` is better here. This returns a redirect but it's still data retrieving (more like a question then the request to change something).

backend/app/assets/stylesheets/spree/backend/components/_quick-switch.scss
```diff
@@ -0,0 +1,51 @@
...
+  .quick-switch-form:not(.searching) & {
+    display: none;
+  }
+}
+
+.quick-switch-help {
+  font-size: 0.8rem;
+}
+
+.quick-switch-help button {
```
kennyadsl: why not nest into `.quick-switch-help` here?

backend/app/assets/javascripts/spree/backend/views/quick_switch.js
```diff
@@ -0,0 +1,71 @@
...
+    this.$el.modal("toggle");
+  },
+
+  onShow: function() {
+    this.$input.focus();
```
Sinetheta: How about saving the user's current target if they're already using a form element and returning it to them afterwards?

backend/app/assets/javascripts/spree/backend/views/quick_switch.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+
+  onHide: function() {
+    this.$input.val("");
+    $(".alert", this.$dialog).remove();
+  },
+
+  onSubmit: function(e) {
+    e.preventDefault();
+    this.toggleSearchClass();
```
Sinetheta: Is there a circumstance where you would want `onSubmit` to _remove_ the search class? You might want to explicitly enable and disable instead of relying on implicit toggle.

backend/app/assets/javascripts/spree/backend/views/quick_switch.js
Mandily: 👍

Mandily:
@Sinetheta If the quick switch field was in the header like Shopify, I suspect many would expect it to behave like the omnisearch in Shopify, which is not what this is. IMO it's better to leave it a little hidden, yet easily accessible by quick key like the Mac OS.

jhawthorn:
I wrote a commit to fix the spec failure here https://github.com/jhawthorn/solidus/commit/54c309beccc209a84bca65784c190f4b7f5bce20

backend/app/assets/javascripts/spree/backend/views/quick_switch.js
graygilmore: If they are already in a form element we won't kick them out.

backend/app/assets/javascripts/spree/backend/views/quick_switch.js
graygilmore: In its current state there is no case where it needs this but that can certainly come if this ever gets more complicated.

backend/app/assets/javascripts/spree/backend/views/quick_switch.js
graygilmore: I've found this to be a common pattern in JS-land. In this case something else triggers a shortcut and then our object knows what to do `onShortcut`. I'd prefer to keep it to keep the pattern alive but could be persuaded.

graygilmore:
Actionable comments have been addressed! Thanks for everyone's patience on this one.

backend/app/assets/javascripts/spree/backend/views/quick_switch.js
tvdeyen: If you would use a custom event name and listener, agreed. 

Anyways. It is great as it is right now. I will not block this great feature.

backend/app/models/spree/quick_switch_item.rb
```diff
@@ -0,0 +1,47 @@
...
+      @help_text_key = help_text_key
+      @not_found_text_key = not_found_text_key
+    end
+
+    def help_text
+      Spree.t(@help_text_key, scope: "quick_switch")
+    end
+
+    def not_found_text(value)
+      Spree.t(@not_found_text_key, scope: "quick_switch", value: value)
```
tvdeyen: We deprecated `Spree.t` in favor of `I18n.t` with a `scope: 'spree'`. Would you please make this change before we merge this?

backend/app/models/spree/quick_switch_item.rb
```diff
@@ -0,0 +1,47 @@
...
+    )
+      @search_triggers = search_triggers
+      @finder = finder
+      @url = url
+      @help_text_key = help_text_key
+      @not_found_text_key = not_found_text_key
+    end
+
+    def help_text
+      Spree.t(@help_text_key, scope: "quick_switch")
```
tvdeyen: Same here. Use `I18n.t` over `Spree.t`

graygilmore:
> One last change request.

Changed `Spree.t` to `I18n.t` 👍

tvdeyen:
@kennyadsl @jhawthorn could you please re-review and give approval? I think all concerns are addressed and this feature is ready to 🚢

kennyadsl:
I'd still add some capybara specs since there's a some JS behaviors which is untested at the moment.

tvdeyen:
I vote against Capybara specs. Request specs for the controller should be enough. The feature is not business critical and feature specs for this particular feature will be hard to write and - even worse - brittle and hard to maintain.

tvdeyen:
@graygilmore would you mind rebasing again? @jhawthorn @kennyadsl are you fine with the changes addressed?

kennyadsl:
Sure 👍 I suggested to have some capybara specs just to have a bell ringing if feature stops working, but if it's just me I'm totally fine merging this PR. We can even add these specs later if something will break in the future.

tvdeyen:
@graygilmore do you mind to write a simple feature smoke test? Thanks

kennyadsl:
Closing in favor of #2809

PR #2138, opened 2017-08-08
Better restrict deleting users with orders

jhawthorn:
Previously, there was a before_destroy on `LegacyUser` which would prevent deleting users which had an attached order by raising `DestroyWithOrdersError`.

`LegacyUser` shouldn't be used in real stores, so this is pointless.

This replaces that check with `dependent: :restrict_with_exception` inside of `UserMethods,` which will restrict deletions of any user model (like `Spree::User` from `solidus_auth_devise`). This also updates the API (the only place which handled this) to understand `DeleteRestrictionError` instead, and to return a better error response.


jhawthorn:
Merged as part of #2139

PR #2139, opened 2017-08-09
Remove rescue_from StandardError in Api::BaseController

jhawthorn:
Previously `Api::BaseController` would swallow all exceptions (or all `StandardError` as of Solidus 1.0) and silently render an exception using `error_during_processing`.

This has long hid errors in specs (see the last few commits for examples) as well as in production usage.

This removes the `rescue_from StandardError`, the `error_during_processing` method, and the configurable `error_notifier`. Instead, errors bubble up and are handled by Rails middleware, as they would be normally.

This allows exceptions to be seen in API specs, stack traces to be seen in development, and errors to be reported to error collection services in production.

This will change the responses for unhandled exceptions:

Previously, unhandled exceptions would return `422 unprocessable entity` (which is just wrong) with JSON and an `exception` key.

Now unhandled exceptions will return `500 server error` (which is correct) and JSON with an `errors` key (if not requesting from a web browser in development, which would instead display a nice error page)

I feel this is an unavoidable API change to the responses, and a good one.

This builds on #2138 and #2136, and includes a few other commits to make the removal possible.

### DIFF
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -24,6 +24,7 @@ module Spree
       before_action :load_user_roles
 
       rescue_from StandardError, with: :error_during_processing
+      rescue_from ActionController::ParameterMissing, with: :parameter_missing_error
       rescue_from ActiveRecord::RecordNotFound, with: :not_found
       rescue_from CanCan::AccessDenied, with: :unauthorized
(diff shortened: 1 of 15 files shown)


mtomov:
Oh, thank you so much. When I used to work with the API, I always had to start with commenting out this rescue. And when I forget .. thanks a lot!

PR #2140, opened 2017-08-09
Remove the ffaker sample data dependency

cbrunsdon:
Continues the idea of #2082, this removes the dependency of ffaker but
also yanks it out of the sample requirement. I generated a small data
set with 15.times.map { FFaker::Name.first_name } etc.

It should be noted that there isn't any reasonable way to export development dependencies to people, and anyone trying to include our factories will get ffaker missing errors. I personally think thats pretty reasonable in exchange for removing ffaker as a useless runtime dependency in prod.

### DIFF
--- a/common_spree_dependencies.rb
+++ b/common_spree_dependencies.rb
@@ -35,6 +35,7 @@ group :test do
   gem 'with_model'
   gem 'rspec_junit_formatter'
   gem 'rails-controller-testing'
+  gem 'ffaker'
 end
 
 group :test, :development do
--- a/sample/db/samples/addresses.rb
+++ b/sample/db/samples/addresses.rb
@@ -1,26 +1,23 @@
 united_states = Spree::Country.find_by!(iso: "US")
 new_york = Spree::State.find_by!(name: "New York")
 
-# Billing address
-Spree::Address.create!(
-  firstname: FFaker::Name.first_name,
-  lastname: FFaker::Name.last_name,
-  address1: FFaker::Address.street_address,
-  address2: FFaker::Address.secondary_address,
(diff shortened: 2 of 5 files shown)


common_spree_dependencies.rb
```diff
@@ -35,6 +35,7 @@
   gem 'with_model'
   gem 'rspec_junit_formatter'
   gem 'rails-controller-testing'
+  gem 'ffaker'
```
tvdeyen: Could we add `require: false` here? We don't need to require `ffaker` every time we boot the dummy apps.

sample/db/samples/addresses.rb
```diff
@@ -1,26 +1,23 @@
...
-  firstname: FFaker::Name.first_name,
-  lastname: FFaker::Name.last_name,
-  address1: FFaker::Address.street_address,
-  address2: FFaker::Address.secondary_address,
-  city: FFaker::Address.city,
-  state: new_york,
-  zipcode: 16_804,
-  country: united_states,
-  phone: FFaker::PhoneNumber.phone_number)
+first_names = ["Sterling", "Jennette", "Salome", "Lyla", "Lola", "Cheree", "Hettie", "Barbie", "Amelia", "Marceline", "Keeley", "Mi", "Karon", "Jessika", "Emmy"]
```
tvdeyen: Personally I prefer the `%w(Sterling Jennette)` syntax for arrays of strings. Also having a new line for each word is far better readable, but this should not block this PR.

tvdeyen:
We probably want to add a changelog entry

tvdeyen:
Closed in favor of #2163

PR #2143, opened 2017-08-09
Remove the border from images in tables

tvdeyen:
As prerequisite to #2092 and #2101 we remove the border from images in tables.


jhawthorn:
:boat:

PR #2145, opened 2017-08-10
Fix customer_details Backbone always setting Guest checkout True

ericsaupe:
When the page is rendered the value of #user_id is set and the Backbone
render was resetting that. This changes the render to fallback to that
value if the model doesn't have the user_id already present.

**Steps to Recreate:**
Create order in Admin
Get to address step
Use Customer Search to add a User to the order
Notice Guest Checkout is now No
Complete Address Step
Go back to Address Step
Guest Checkout is now Yes even though a user is attached to order

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/order/customer_details.js
+++ b/backend/app/assets/javascripts/spree/backend/views/order/customer_details.js
@@ -53,7 +53,7 @@ Spree.Views.Order.CustomerDetails = Backbone.View.extend({
   },
 
   render: function() {
-    var user_id = this.model.get("user_id")
+    var user_id = this.model.get("user_id") || $("#user_id").val()
     this.$("#user_id").val(user_id);
     this.$('#guest_checkout_true')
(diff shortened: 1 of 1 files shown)


jhawthorn:
Is it possible to get a regression spec for this?

ericsaupe:
The model does have the `user_id` set. When going back that the Backbone JS it isn't initializing with that data.

tvdeyen:
> Backbone JS isn't initializing with that data

Can we fix this then instead?

ericsaupe:
That's what this does. It either uses the data that has been set in Backbone through another method or gets the value from the rendered page while initializing.

cbrunsdon:
We can make @swcraig write a regression feature spec for this.

swcraig:
I am working on the regression spec for this.

cbrunsdon:
Closing in favor of  #2176 which includes the regression

ericsaupe:
Thanks!

PR #2146, opened 2017-08-10
Convert admin RABL views to Jbuilder

jhawthorn:


backend/solidus_backend.gemspec
```diff
@@ -31,6 +31,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'jquery-ui-rails', '~> 5.0.0'
   s.add_dependency 'font-awesome-rails', '~> 4.0'
   s.add_dependency 'kaminari', '>= 0.17', '< 2.0'
+  s.add_dependency 'jbuilder'
```
tvdeyen: We should set version numbers

jhawthorn: :+1: done

cbrunsdon:
When you see anything that looks like this: `my_instance.()` it calls the `call` method on that object, if its an object, or will invoke a proc if its a proc.

jhawthorn:
> When you see anything that looks like this: `my_instance.()` it calls the `call` method on that object, if its an object, or will invoke a proc if its a proc.

Yeah, JBuilder uses this (maybe too) cleverly as a shorthand for multiple attributes. I replaced a few spots where it didn't make sense to use this over explicitly declaring the attributes on separate lines.

PR #2147, opened 2017-08-11
Rewrite API RABL views as Jbuilder

jhawthorn:
This rewrites all of our (previously RABL) API views in Jbuilder

## TODO

* [x] Re-add caching.


cbrunsdon:
This all looks good to me, are the changes here 100% compatible (as far as you can tell) or is there something we should maybe be aware of that behaves differently?

jhawthorn:
> This all looks good to me, are the changes here 100% compatible (as far as you can tell) or is there something we should maybe be aware of that behaves differently?

Changes _should_ be 100% compatible.

One thing to watch out for is we no longer have RABL's implicit checking for records being `nil`, so we have to check explicitly for associations being nil. I've tried to do this but it's possible I've missed some that aren't exposed by our tests.

jhawthorn:
I did a hacky test with our spec suite to ensure that the responses were compatible. For each get/put/post/delete in our test suite I recorded the response (normalized by sorting the JSON keys, starting from the same seed) and compared the output JSON.

I found one discrepancy (associations being `null` vs just skipping the key) which is fixed by the latest commit (81a5638).

Afaik, this is good to go :shipit:

api/app/views/spree/api/addresses/_address.json.jbuilder
```diff
@@ -0,0 +1,17 @@
+json.cache! address do
```
kennyadsl: I know we just want to reflect views as they are, I just want to take note that this is the only place where `cache!` is used without `I18n.locale` key. Probably stuff for another PR.

cbrunsdon:
We'll need a serious change log entry for this. Everyone that has extended or changed rabl templates will have to adapt.

PR #2056, opened 2017-07-04
Fixed insufficient inventory error message.

husam212:

PR #2152, opened 2017-08-14
Fix typo in core en.yml

isaacfreeman:


isaacfreeman:
One spec not passing. Not sure why, but I'll investigate.

swcraig:
I didn't investigate the one failing spec, but I believe this typo was fixed in 74d7b940 from PR #2056.

https://github.com/solidusio/solidus/pull/2056/files#diff-3329c5ee0df3de06561549745b116615R1367

PR #2163, opened 2017-08-17
Remove the ffaker sample data dependency (revised)

swcraig:
Addresses comments brought up in @cbrunsdon's [original PR](https://github.com/solidusio/solidus/pull/2140).

Original PR description
--------------
Continues the idea of #2082, this removes the dependency of ffaker but
also yanks it out of the sample requirement. I generated a small data
set with 15.times.map { FFaker::Name.first_name } etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Solidus 2.4.0 (master, unreleased)
 
+- Remove `ffaker` as a runtime dependency in production [\#2140](https://github.com/solidusio/solidus/pull/2140) ([cbrunsdon](https://github.com/cbrunsdon))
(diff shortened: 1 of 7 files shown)


swcraig:
https://github.com/solidusio/solidus/pull/2140#discussion_r132139967

@tvdeyen I didn't change the name arrays to use `%w` syntax because the other arrays in this file use multi-word values and I felt that consistency in the file would be favoured. Instead I spread the arrays over multiple lines when they grew past 80 characters. Let me know if you want me to revert the arrays to sit on single lines again (or whatever solution you find reasonable).

mtomov:
Thanks!

PR #2164, opened 2017-08-17
Tweak font styles on admin shipments page

jhawthorn:
This PR:
* Removes **bold** from shipping method name (this is a value, not a heading)
* Removes **bold** from adjustment name values (this is a value, not a heading)
* *Italicizes* "No tracking details provided." (to make it clearer this is a placeholder)

**Before**

**After**


Mandily:
Looks good to me 👍

tvdeyen:
WIP, but this is what I have right now

shipments - r987654321 - orders 2017-08-18 00-56-08

tvdeyen:
> but i guess we can do that in a different PR.

Upcoming

PR #2165, opened 2017-08-17
Show table borders on action columns

jhawthorn:
Previously, the action icons were shown clearly outside the table row they belonged to. I suspect this is just because it looked bad.

With the new table design, I believe it looks better to group the actions under the same row separator as the rest of the row, and it makes it more obvious what it is operating on.

This also right-aligns the action cells.

**Before**

*Note extra whitespace at right*

**After**

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_actions.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_actions.scss
@@ -26,8 +26,6 @@ table tbody tr {
   }
 
   td.actions {
-    background-color: transparent !important;
-
     button {
       cursor: pointer;
     }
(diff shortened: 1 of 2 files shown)


tvdeyen:
Some people may find it weird that the actions are right aligned, especially on tables with unequal actions item counts

payments - r987654321 - orders 2017-08-17 22-53-52

shipments - r987654321 - orders 2017-08-17 22-54-13

I like it 👍

Mandily:
When thinking about the alignment of the action buttons - the thing that I would most want to align to is the same action above or below the current row. So a particular action could always be found by scanning straight down. It's one more little piece (location) that helps a user identify what the action is. Kind of like how a colour blind person knows that the top light on a traffic light is stop even though they don't see red. 

For example: all of the X's would line up straight down the side. 
29433353-2071b138-839f-11e7-9ccd-3e6d90f5720c

Thoughts?

jhawthorn:
Aligning the identical actions looks good (for the cases it makes sense), but I suspect we'd have to do it with placeholders. This PR won't help or hurt our ability to do that in the future.

PR #2120, opened 2017-07-26
Checkout email input field should use email_field

notapatch:
- on frontend checkout/edit
 - mobile users will have a virtual keyboard that now includes
   an @ symbol.
 - fallback for none HTML 5 browsers is to text type

PR #2166, opened 2017-08-17
Checkout email input field should use email_field (revised)

swcraig:
Rebasing and pushing @notapatch's [original PR](https://github.com/solidusio/solidus/pull/2120). CircleCI was [giving a phantom failure](https://github.com/solidusio/solidus/pull/2120#issuecomment-318725277), but that should be resolved now.

Also very subtly changes the commit message (get rid of a period and change "none" to "non").

Original PR description
---------------

- on frontend checkout/edit
 - mobile users will have a virtual keyboard that now includes
   an @ symbol
 - fallback for non HTML 5 browsers is to text type


mamhoff:
Closing as #2120 has been merged.

PR #2167, opened 2017-08-17
Fix action button hover style

tvdeyen:
The action button elements must not have a backgorund color on mosue hover. Also fixes the icon hover color for shippment table action buttons.

### Before

monosnap 2017-08-17 23-07-57

### After

monosnap 2017-08-17 23-06-58

- [x] Fix the tooltips color for split shipment action


tvdeyen:
Tooltip color also fixed now

monosnap 2017-08-17 23-42-43

PR #2169, opened 2017-08-18
Add headers to shipment method and tracking number

tvdeyen:
Without headers it is hard to guess what "UPS same day" on a order shiments
table actually means.

Even more important, when in edit mode a form label is displayed in top of
the form field what leads to jumpy and confusing UI.

By adding headers to all view states this table row can have, we get a more
consistent UI.

### Before

shipping-edit-before

### After

shipping-edit-after


tvdeyen:
Please excuse the poor GIF quality on the second image. The background is actually not gray ;)

PR #2171, opened 2017-08-18
Make checkout billing address inputs full width

notapatch:
- shipping address already has input elements full width
 - billing address dropdown boxes are already full width
 - change makes billing address consistent with shipping address


notapatch:
Without the change what you see is the input boxes taking up part of the width whereas other elements use the full-width - you have to uncheck the "Use Billing Address" checkbox to see the difference. This change brings it in line with the Billing Address.

PR #2173, opened 2017-08-21
Make `Spree::Config` a singleton class

swcraig:
At the moment `Spree::Config` is a constant that is being assigned as an instance of `Spree::Core::Environment.new.preferences`. Using a singleton object here is a more reasonable approach. 

`Spree::Config.instance` is the actual `Spree::AppConfiguration`. Also note that `Spree::Config#method_missing` forwards calls to `Spree::Config.instance` so `Spree::Config` can continue to be used the same way as before.

### DIFF
--- a/core/lib/spree/config.rb
+++ b/core/lib/spree/config.rb
@@ -1,0 +1,23 @@
+module Spree
+  class Config
+    def self.instance
+      @@instance || Spree::AppConfiguration.new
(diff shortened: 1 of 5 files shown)


cbrunsdon:
Killing this one as #2178 accomplishes the same goals and is a better move forward, thanks though @swcraig

PR #2174, opened 2017-08-22
Change product's price color away from link color

notapatch:
- price on products page has a link color but it is not a link
  - links should be distinct to improve usability
  - chosen color variable product_body_text_color

### DIFF
--- a/frontend/app/assets/stylesheets/spree/frontend/_variables.scss
+++ b/frontend/app/assets/stylesheets/spree/frontend/_variables.scss
@@ -12,6 +12,7 @@ $link_text_color:           #00ADEE !default;
 $product_background_color:  #FFFFFF !default;
 $product_title_text_color:  #404042 !default;
 $product_body_text_color:   #404042 !default;
+$product_price_text_color:  #252525 !default;
 $product_link_text_color:   #BBBBBB !default;
 
 /*--------------------------------------*/
--- a/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
+++ b/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
@@ -452,7 +452,7 @@ nav#taxonomies {
 
 %price_text {
   font-weight: bold;
-  color: $link_text_color;
+  color: $product_price_text_color;
 }
 
 span.price {
@@ -520,7 +520,7 @@ ul#products {
     }
 
     .price {
-      color: $link_text_color;
+      color: $product_price_text_color;
       font-size: $product_list_price_font_size;
       padding-top: 5px;
       display: block;


notapatch:
I am open to suggestions on what color variable you would like the price text to be but since it isn't a link ... this is more an "anything but link color" pull request.

cbrunsdon:
I'm not picky on colors at all (or the default frontend, for that matter), but @Mandily might have an opinion.

Mandily:
I agree with @notapatch - it is confusing to have the prices looking like links. Swapping the variable for the product_body_text_color looks fine to me. Screenshot for anyone wondering what this looks like: 👍 

screen shot 2017-08-22 at 1 26 29 pm

notapatch:
I didn't explain it very well - I did not mention this was for the products#show page (I will do print screens to help reviewers) and not the index page as in the above image. 

However, as @Mandily points out .. the products/index page is also using the link colors for non-links. If people are agreeable, I can extend the pull request to include the price on the product's index page - I will also check if price appears consistently throughout the front end application.

Mandily:
Sounds good to me @notapatch Can you post screenshots once you've made changes to the other areas please and thanks!

notapatch:
# What do shopping sites do?

I took the first 20 sites from [Alexa shopping category](1) \[1\] and 3/4 of the websites had the non-sale price in black. Four sites used a brand or accent color on #index and the #show page and one site had different price colors on index and show. 

In other words, if I show the price in black and leave a variable to change the price color then we can mock the requirements for 95% of the websites in the top 20. 

# what did I change?

I chose an application unique dark gray color so that anyone picking the color can grep and find it quickly - #252525.

What pages were changed?

products#index, products#show, orders#edit, checkout order summary panel (affecting Address, Delivery, Payment and Confirm).

# Products#index  _Before_
solidus-products-index-before

# Products#index _After_
solidus-products-index-after

# Orders#edit _Before_

solidus-orders-edit-before

# Orders#edit _After_

solidus-orders-edit-after

# Checkout order summary panel _Before_

solidus-checkout-address-before

# Checkout order summary panel _After_

- it would probably look nicer if the "Order Summary" panel had a more consistent look with the Billing and Shipping panels

solidus-checkout-address-after

# Checkout - confirm _Before_

solidus-checkout-confirm-before

# Checkout - confirm _After_
* "Sub total" and "order total" should be de-emphasized as the shopper is more interested in the amounts than the titles  - this will give a "cleaner" look to the page.

solidus-checkout-confirm-after


# Conclusion

For a few lines of code, this is a big change to the appearance - as you can imagine I willing to take my time on this - we might only want the change affecting products#index and products#show? However, the demo app should eventually have links distinctive from any other element on the page throughout the application because some people will judge the quality of the code on first appearances and distinctive links is a basic usability requirement.

\[1\] - I did not count Amazon.co.uk as I assumed it was the same as Amazon.com, Netflix because they required a login and I assume they aren't going to be a standard shopping site.

[1]: https://www.alexa.com/topsites/category/Shopping

Mandily:
Looks good to me! 👍

PR #2176, opened 2017-08-22
Check if user_id is already set (with regression spec)

swcraig:
From #2145. This is @ericsaupe's commit modified to include the regression spec.

Original PR
----------------
When the page is rendered the value of #user_id is set and the Backbone
render was resetting that. This changes the render to fallback to that
value if the model doesn't have the user_id already present.


ericsaupe:
Thanks @swcraig for the regression spec!

PR #2177, opened 2017-08-22
Capitalize event buttons in `OrdersHelper`

swcraig:
lowercase-order-buttons

These seem to be the only buttons that are not capitalized in the
backend. They look especially weird since they can sit beside the
"Resend" button which is capitalized.

**Edit**: This only changes the translation values in `en.yml`. They don't appear to be used anywhere else. The one problem with this approach is that the [alert text that is generated](https://github.com/solidusio/solidus/blob/master/backend/app/helpers/spree/admin/orders_helper.rb#L11) will have the order event capitalized (e.g "Are you sure you want to Cancel this order?"). I think having an "errant" capitalization in the alert text is more desirable than lowercase button text.

### DIFF
--- a/backend/spec/features/admin/orders/cancelling_and_resuming_spec.rb
+++ b/backend/spec/features/admin/orders/cancelling_and_resuming_spec.rb
@@ -21,7 +21,7 @@ describe "Cancelling + Resuming", type: :feature do
 
   it "can cancel an order" do
     visit spree.edit_admin_order_path(order.number)
-    click_button 'cancel'
+    click_button 'Cancel'
     within(".additional-info") do
       expect(find('dt#order_status + dd')).to have_content("Canceled")
     end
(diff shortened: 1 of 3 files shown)


backend/app/helpers/spree/admin/orders_helper.rb
```diff
@@ -6,7 +6,8 @@ def event_links
         links = []
         @order_events.sort.each do |event|
           next unless @order.send("can_#{event}?")
-          links << button_link_to(Spree.t(event), [event, :admin, @order],
+          links << button_link_to(Spree.t(event).capitalize,
```
jhawthorn: Probably don't want to do it like this. The choice to capitalize or not is language specific, so it should exist somewhere in translations.

swcraig: Thanks John, I didn't even think about that. I'll get something else for this.

PR #2183, opened 2017-08-23
Fix StoreCredit with multiple currencies

jordan-brough:
Please see [individual commits](https://github.com/solidusio/solidus/pull/2183/commits).

High level overview:

- Only apply StoreCredit to an Order if the StoreCredit currency matches the Order currency
- Group StoreCredits by currency in the admin

It took some effort to _not_ attempt a refactor of a bunch of the scary or weird-looking StoreCredit code I noticed while working on this.


jordan-brough:
FYI: I force-pushed after Thomas's review to fix some specs and add some specs.  No code changes.

PR #2184, opened 2017-08-23
Fix backend data-action across multiple files

kennyadsl:
In backend there are multiple links with inconsistent `data-action` attributes. 

For example `data-action="red"` or `data-action="green"`, which are both not semantic and do not change tooltip colors when needed (tooltip colors are defined with [actions names](https://github.com/solidusio/solidus/blob/96b30d6f4e644383cfa13fb2dd208bcd3589cb81/backend/app/assets/stylesheets/spree/backend/plugins/_bootstrap_tooltip.scss#L34-L42), not color).

With this PR we also start using these semantic `data-action` into js to target buttons instead of rely on class names that can change in the future, like icons ones (`fa-something`).

The only real effect of this PR on the UI is that buttons and tooltips use the right color now:

##### Before

stock_cancel_before

##### After

stock_cancel_after

##### Before

stock_transfer_cancel_before
stock_transfer_save_before

##### After

stock_transfer_cancel_after
stock_transfer_save_after

##### Before

store_credit_cancel_before

##### After

store_credit_cancel_after

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/number_field_updater.coffee
+++ b/backend/app/assets/javascripts/spree/backend/number_field_updater.coffee
@@ -24,7 +24,7 @@ class NumberFieldUpdater
     $("#number-update-#{id} span").css('display', cssDisplay)
 
   toggleForm = (id, show) ->
-    toggleButtonVisibility('void', id, show)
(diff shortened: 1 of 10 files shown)


mtomov:
Cool that!

PR #2185, opened 2017-08-23
Remove extra set_current_order calls (updated CHANGELOG and rebase)

swcraig:
This rebases and updates the CHANGELOG entry for @Murph33's #1137.

In the CHANGELOG entry I've included what the "current enough version" of solidus_auth_devise is (the `set_current_order` call was [added in v1.5.0](https://github.com/solidusio/solidus_auth_devise/blob/master/CHANGELOG.md#solidus-auth-devise-v150-2016-07-18). I've also included [the issue it addressed](https://github.com/solidusio/solidus/issues/1116).

Original PR
---------------
Currently this is calling on almost every request in the front end.  We
want to push this to where it should be which is just when we sign in
authentication systems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 ## Solidus 2.4.0 (master, unreleased)
 
+- Remove `set_current_order` calls in `Spree::Core::ControllerHelpers::Order`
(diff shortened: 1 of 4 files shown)


mtomov:
Thanks a lot!

Issue #2155, opened 2017-08-15
Payment step, coupon code ambiguous behaviour


shiroginne:
Hello,

I've faced ambiguous behaviour of coupon codes on payment step:
when user has saved payment methods (credit card, for example) and he tries to enter wrong coupon code (page reloads), section with saved payment methods is gone.
If user clicks on "Save and Continue" with empty coupon code, he got credit card validation error but saved payment methods are on place.

### Steps to reproduce
1. buy something with credit card (save payment method)
2. try to buy something else
3. enter wrong coupon code
4. press 'Save and Continue'

<img width="979" alt="screen shot 2017-08-15 at 14 49 11" src="https://user-images.githubusercontent.com/63340/29316801-9370b88c-81c9-11e7-81e7-6ddf45b17b52.png">
<img width="964" alt="screen shot 2017-08-15 at 14 49 19" src="https://user-images.githubusercontent.com/63340/29316803-95e2bcc8-81c9-11e7-8072-b0c209c11e24.png">
<img width="1011" alt="screen shot 2017-08-15 at 14 49 31" src="https://user-images.githubusercontent.com/63340/29316807-975b1866-81c9-11e7-8861-35eac08fadd7.png">

### Expected behavior
Saved payment methods present on page with coupon code error.

### Actual behavior
Saved payment methods present absent on payment page when wrong coupon code entered.

### System configuration
**Solidus Version**:
- 2.0.2
- 2.2.1

**Extensions in use**:
```ruby
gem 'solidus'
gem 'solidus_auth_devise'
```

I'm ready to help/make a fix for this, just not sure where to start to find out why it behaves like this.

cbrunsdon:
@benjaminwil can you give a shot at re-creating this please

kennyadsl:
[...]

PR #2186, opened 2017-08-25
[WIP]: Fix setting checkout steps context applying coupon codes

kennyadsl:
Ref: #2155  

In case of errors applying coupon codes, [the edit view is rendered and controller callbacks execution ends](https://github.com/solidusio/solidus/blob/548b54f4744696ef3c4dcc774639a7f8911583dd/frontend/app/controllers/spree/store_controller.rb#L30). 

But checkout steps specific ivars are set with `setup_for_current_state` callbacks which happens after coupon code application, for example [wallet payments options are set before payment](https://github.com/solidusio/solidus/blob/4ade94e2c9f45bcccd93794336823f9f05bb8ab8/frontend/app/controllers/spree/checkout_controller.rb#L199). 

This PR forces calling `setup_for_current_state` if we are stopping callbacks execution after a coupon application error.  

Since apply_coupon_code is also called from `OrdersController` a `respond_to?` is required to be sure we only call it when needed.

### DIFF
--- a/frontend/app/controllers/spree/store_controller.rb
+++ b/frontend/app/controllers/spree/store_controller.rb
@@ -19,6 +19,7 @@ module Spree
     # which needs it)
     def apply_coupon_code
       if params[:order] && params[:order][:coupon_code]
+        Spree::Deprecation.warn("Passing params[:order][:coupon_code] is deprecated and will have no effect in a future version.", caller)
         @order.coupon_code = params[:order][:coupon_code]
 
         handler = PromotionHandler::Coupon.new(@order).apply
--- a/frontend/app/controllers/spree/store_controller.rb
+++ b/frontend/app/controllers/spree/store_controller.rb
@@ -24,6 +24,11 @@ module Spree
         handler = PromotionHandler::Coupon.new(@order).apply
 
         if handler.error.present?
+          # If there are errors applying the coupon code we need to render the
+          # edit view and stop the request execution. Checkout steps need a
+          # setup before rendering the edit view or they will miss some ivars.
+          send(:setup_for_current_state) if respond_to?(:setup_for_current_state, true)
+
           flash.now[:error] = handler.error
           respond_with(@order) { |format| format.html { render :edit } } && return
         elsif handler.success
--- a/frontend/spec/controllers/spree/checkout_controller_spec.rb
+++ b/frontend/spec/controllers/spree/checkout_controller_spec.rb
@@ -429,4 +429,33 @@ describe Spree::CheckoutController, type: :controller do
       post :update, params: { state: "payment" }
     }.to change { order.line_items.to_a.size }.from(1).to(0)
   end
+
+  context 'trying to apply a coupon code' do
+    let(:order) { create(:order_with_line_items, state: 'payment', guest_token: 'a token') }
+    before { cookies.signed[:guest_token] = order.guest_token }
+
+    context "when coupon code is applied" do
+      let(:coupon_code) { "coupon_code" }
+      before { create(:promotion, :with_order_adjustment, code: coupon_code) }
+
+      it "continues checkout flow normally" do
+        put :update, params: { state: order.state, order: { coupon_code: coupon_code } }
+
+        expect(response).to redirect_to(spree.checkout_state_path('confirm'))
+        expect(flash.now[:success]).to eq(Spree.t(:coupon_code_applied))
+      end
+    end
+
+    context 'when coupon code is not applied' do
+      let(:coupon_code) { 'wrong_coupon_code' }
+
+      it "renders edit view and setups current checkout step correctly" do
+        put :update, params: { state: order.state, order: { coupon_code: coupon_code } }
+
+        expect(response).to render_template :edit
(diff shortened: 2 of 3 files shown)


frontend/app/controllers/spree/store_controller.rb
```diff
@@ -26,6 +26,11 @@ def apply_coupon_code
         handler = PromotionHandler::Coupon.new(@order).apply
 
         if handler.error.present?
+          # If there are errors applying the coupon code we need to render the
+          # edit view and stop the request execution. Checkout steps need a
+          # setup before rendering the edit view or they will miss some ivars.
+          send(:setup_for_current_state) if respond_to?(:setup_for_current_state, true)
```
kennyadsl: maybe an alternative approach could be redirecting to the right checkout step?

gmacdougall: This is pointing to a larger problem in this bit of code. I think it might be best to apply some refactoring to here in order to avoid the `respond_to?` and `send`.

Some suggestions of things I would try:

1. Extract `apply_coupon_code` to a common base class (which extends `StoreController`) and modify the checkout controller and orders controller to inherit from that.
2. Extract the error handling to a method, allowing it to be overridden to handle the error condition and call `super` to finish the job.

frontend/spec/controllers/spree/checkout_controller_spec.rb
```diff
@@ -429,4 +429,46 @@ def post_address
...
+    let(:order) do
+      create(:order_with_line_items).tap do |order|
+        until order.state == 'payment'
+          order.next!
+        end
+      end
+    end
+
+    before do
+      allow(controller).to receive_messages current_order: order
```
gmacdougall: I would like it if we could avoid stubbing the controller under test.

frontend/spec/controllers/spree/checkout_controller_spec.rb
```diff
@@ -429,4 +429,46 @@ def post_address
       post :update, params: { state: "payment" }
     }.to change { order.line_items.to_a.size }.from(1).to(0)
   end
+
+  context 'trying to apply a coupon code' do
+    let(:order) do
+      create(:order_with_line_items).tap do |order|
+        until order.state == 'payment'
```
gmacdougall: Should this be a factory trait of some sort?

frontend/spec/controllers/spree/checkout_controller_spec.rb
```diff
@@ -429,4 +429,46 @@ def post_address
...
+
+    context 'when coupon code is not applied' do
+      let(:coupon_code) { 'wrong_coupon_code' }
+
+      it "renders edit view and setups current checkout step correctly" do
+        put :update, params: { state: order.state, order: { coupon_code: coupon_code } }
+
+        expect(response).to render_template :edit
+        expect(flash.now[:error]).to eq(Spree.t(:coupon_code_not_found))
+        expect(assigns(:wallet_payment_sources)).not_to be_nil
```
gmacdougall: What should it be?

frontend/spec/features/coupon_code_spec.rb
```diff
@@ -32,50 +32,106 @@ def create_basic_coupon_promotion(code)
...
+          )
+
+          order.reload
+          order.user = user
+          order.recalculate
+          order
+        end
+
+        before do
+          allow_any_instance_of(Spree::CheckoutController).to receive_messages(current_order: order)
```
gmacdougall: I don't like the use of stubs here. The feature specs are intended to interact with the site as a user would. Here we create an order for them and stub things rather than browsing through the site to add stuff to a cart.

frontend/spec/features/coupon_code_spec.rb
```diff
@@ -32,50 +32,106 @@ def create_basic_coupon_promotion(code)
...
+            click_button "Save and Continue"
+            # To payment screen
+            click_button "Save and Continue"
+          end
+
+          it "shows wallet payments on coupon code errors" do
+            fill_in "order_coupon_code", with: "coupon_codes_rule_man"
+            click_button "Save and Continue"
+
+            expect(page).to have_content(Spree.t(:coupon_code_not_found))
```
gmacdougall: I think it's helpful to use the actual string here. I've seen more than a few translation bugs get caught when people were specifying the key wrong in both places.

frontend/spec/controllers/spree/checkout_controller_spec.rb
kennyadsl: yes, I can just do `create(:order_with_line_items, state: 'payment')`

frontend/spec/features/coupon_code_spec.rb
kennyadsl: yes, you are right. I changed the code to avoid stubbing the current order

kennyadsl:
Just an update on this PR: I cleaned up everything following @gmacdougall suggestions. Anyway the core of the fix is still something that needs to be addressed. Talking with the rest of the core team we decided to:

1. deprecate this `apply_coupon_code` callback to apply coupon code  
2. use a separate controller which will handle, with a specific request the coupon code application

kennyadsl:
closing in favor of #2327

Issue #1999, opened 2017-06-06
Free shipping promotion applied when rules are not satisfied


nhippenmeyer:
### Steps to reproduce
1. Create a free shipping promotion and add a rule for item total to be greater than or equal to a certain amount.
2. Check out with an order that doesn't satisfy the minimum item total.

### Expected behavior
The free shipping promotion should not be applied to the order.

### Actual behavior
The free shipping promotion is applied to the order even when the rule isn't satisfied.

### System configuration
**Solidus Version**: 2.2.1

**Pull request with fix**: https://github.com/solidusio/solidus/pull/1998

fylooi:
Just upgraded to 2.2.1, seeing this as well.

loicginoux:
Do you still have the issue with 2.4 or master branch ?

jacobherrington:
This seems to be fixed in 2.7.

aitbw:
This seems fixed via #2187

PR #2187, opened 2017-08-28
Add eligibility checking to automatic free shipping promotions

fylooi:
This PR is an alternate fix for issue #1999 since the original one seems to have stalled. 

`Spree::PromotionHandler::FreeShipping` currently does not check whether the promotion is eligible before activating it. This is an issue for automatic promotions as any associated promotion rules (eg. minimum order total) are ignored. 

No problem for non-automatic promotions as they are validated upon activation by other handlers.

### DIFF
--- a/core/spec/models/spree/promotion_handler/shipping_spec.rb
+++ b/core/spec/models/spree/promotion_handler/shipping_spec.rb
@@ -13,8 +13,20 @@ module Spree
       context 'with apply_automatically' do
         let!(:promotion) { create(:promotion, apply_automatically: true, promotion_actions: [action]) }
 
-        it "creates the adjustment" do
-          expect { subject.activate }.to change { shipment.adjustments.count }.by(1)
+        context 'for eligible promotion' do
+          it "creates the adjustment" do
+            expect { subject.activate }.to change { shipment.adjustments.count }.by(1)
+          end
+        end
+
+        context 'for ineligible promotion' do
+          let!(:promotion) do
+            create(:promotion, :with_item_total_rule, item_total_threshold_amount: 1_000, apply_automatically: true, promotion_actions: [action])
+          end
+
+          it "does not create the adjustment" do
+            expect { subject.activate }.to change { shipment.adjustments.count }.by(0)
+          end
         end
       end
 
--- a/core/spec/models/spree/promotion_handler/shipping_spec.rb
+++ b/core/spec/models/spree/promotion_handler/shipping_spec.rb
@@ -43,6 +43,22 @@ module Spree
               subject.activate
             }.to change { shipment.adjustments.count }
           end
+
+          context 'when currently ineligible' do
+            let(:promotion) do
+              create(:promotion, :with_item_total_rule, item_total_threshold_amount: 1_000, code: 'freeshipping', promotion_actions: [action])
+            end
+
+            before do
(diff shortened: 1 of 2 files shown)


core/app/models/spree/promotion_handler/free_shipping.rb
```diff
@@ -18,7 +18,7 @@ def activate
         end
 
         not_connected_automatic_promotions.each do |promotion|
-          promotion.activate(order: order)
+          promotion.activate(order: order) if promotion.eligible?(order)
```
jordan-brough: Would you mind making this a normal `if` instead of a trailing `if`?  The condition is quite important, as you've pointed out, so it'd be nice to make sure it stands out

core/spec/models/spree/promotion_handler/free_shipping_spec.rb
```diff
@@ -13,8 +13,20 @@ module PromotionHandler
       context 'with apply_automatically' do
         let!(:promotion) { create(:promotion, apply_automatically: true, promotion_actions: [action]) }
 
-        it "creates the adjustment" do
-          expect { subject.activate }.to change { shipment.adjustments.count }.by(1)
+        context 'for eligible promotion' do
+          before { allow_any_instance_of(Spree::Promotion).to receive(:eligible?).and_return(true) }
```
jordan-brough: You don't need this because the default promotion has no rules and is eligible, right?

core/spec/models/spree/promotion_handler/free_shipping_spec.rb
```diff
@@ -13,8 +13,20 @@ module PromotionHandler
...
+
+          it "creates the adjustment" do
+            expect { subject.activate }.to change { shipment.adjustments.count }.by(1)
+          end
+        end
+
+        context 'for ineligible promotion' do
+          before { allow_any_instance_of(Spree::Promotion).to receive(:eligible?).and_return(false) }
```
jordan-brough: Rather than stubbing would you mind instead creating a promotion that will in fact not be eligible?  e.g. 
```ruby
let(:promotion) { create(:promotion, :with_item_total_rule, item_total_threshold_amount: 1_000, ... }
```

fylooi:
@jordan-brough 

Thanks for the review. Have applied the requested changes, merged your PR and rebased master.

core/spec/models/spree/promotion_handler/shipping_spec.rb
```diff
@@ -13,8 +13,20 @@ module PromotionHandler
...
-          expect { subject.activate }.to change { shipment.adjustments.count }.by(1)
+        context 'for eligible promotion' do
+          it "creates the adjustment" do
+            expect { subject.activate }.to change { shipment.adjustments.count }.by(1)
+          end
+        end
+
+        context 'for ineligible promotion' do
+          let(:promotion) do
+            create(:promotion, :with_item_total_rule, item_total_threshold_amount: 1_000, code: 'freeshipping', promotion_actions: [action])
```
jordan-brough: You want to set `apply_automatically: true` in this context rather than `code: 'freeshipping'`, right?

fylooi: My bad, copied that as I wasn't familiar with the factories. Also needs to be a `let!`.

mamhoff:
We hope this fixes the issue, we're reasonably sure it got introduced in #1392 and this seems to fix it.

PR #2188, opened 2017-08-29
Update Solidus to work correctly with non-promotion line-level adjustments

jordan-brough:
In #1933 we starting allowing non-promotion adjustments on line items
and shipments.  However, we didn't update some other important code (including
the DefaultTax calculator) to consider these adjustments.

This PR fixes the above issue in its first commit and renames several existing methods in its second commit.

Please see individual commits.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -59,30 +59,37 @@ module Spree
     end
     deprecate discounted_amount: :total_before_tax, deprecator: Spree::Deprecation
 
+    # @return [BigDecimal] the amount of this line item, taking into
+    #   consideration all its adjustments.
+    def total
+      amount + adjustment_total
+    end
+    alias final_amount total
+    deprecate final_amount: :total, deprecator: Spree::Deprecation
+
     # @return [BigDecimal] the amount of this item, taking into consideration
     #   all non-tax adjustments.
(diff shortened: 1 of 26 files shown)


core/app/models/spree/line_item.rb
```diff
@@ -57,6 +57,13 @@ def amount
     def discounted_amount
       amount + promo_total
     end
+    deprecate discounted_amount: :final_amount_without_additional_tax, deprecator: Spree::Deprecation
+
+    # @return [BigDecimal] the amount of this item, taking into consideration
+    #   all non-tax adjustments.
+    def final_amount_without_additional_tax
+      amount + adjustments.reject(&:tax?).sum(&:amount)
```
jhawthorn: This probably needs to check `eligible?` (and a spec catching that would be great)

jordan-brough: 👍 thanks. Added a spec as well.

jordan-brough:
`final_amount_without_tax` might be a better name than `final_amount_without_additional_tax`?  Having both `included` and `additional` types of tax make things more confusing. I'm not sure which name is better. Or if there's something else even better.  The fact that we also have `LineItem#pre_tax_amount` makes things even more confusing, since that method is dealing with `included` tax.

PR #2190, opened 2017-08-30
Remove autoload on product_filters

cbrunsdon:
I'm almost positive this does nothing.

EDIT: it did something. but not that much.


cbrunsdon:
I thought I could remove the autoload, but I also needed to add the require.

PR #2193, opened 2017-08-30
Remove task to copy shipped shipments to cartons

cbrunsdon:
This was supposed to be ran back in the 1.2 days, and we don't need to
keep it around anymore. It'll still be in the repo history if anyone
needs it for whatever reason.


samkpower:
@askl56 Agreed, just wanted to chime in and share this was a breaking change on my end as well. I believe general best practices are to keep the migration history intact, no? I too love to delete code, but if it were me, I think I would've kept the migration as is and left the rake task blank.

kennyadsl:
Fair point, we'll take that into account for the next releases. FYI, I think we already started deprecating rake tasks before removal (that only happens in major releases). Thanks for letting us know about the trouble!

PR #2194, opened 2017-08-30
Remove PriceMigrator

cbrunsdon:
This was intended to be run in the 1.3 build, and should be safe to
remove now as we head into 2.4.

The task will still be there for anyone that wants it, but its more
likely that if they're moving to 2.4 this isn't going to do them any
good.


mamhoff:
Can I PR more removals to this PR? If we get rid of this migration, I would also like to remove Spree::Zone.default_tax and all the code paths that do the weirdo "tax refund" thing.

mamhoff:
I'll PR that other removal separately.

PR #2198, opened 2017-09-01
Disable Rubocop Style/TernaryParentheses

jordan-brough:
This is sometimes more readable and doesn't seem worth complaining about.

e.g.:
```
core/lib/spree/core/search/base.rb:113:34: C: Omit parentheses for ternary conditions.
@properties[:page] = (params[:page].to_i <= 0) ? 1 : params[:page].to_i
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -236,6 +236,9 @@ Style/StringLiterals:
 Style/SymbolProc:
   Enabled: false
 
+Style/TernaryParentheses:
+  Enabled: false
+
 Style/WhileUntilModifier:
   Enabled: false
   StyleGuide: http://relaxed.ruby.style/#stylewhileuntilmodifier


swcraig:
If we are interested, there is a style `require_parentheses_when_complex` that enforces parentheses around the conditional if it includes more than a single entity (like the expression in your PR description).

https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/ternary_parentheses.rb#L41-L51

```
Style/TernaryParentheses:
   EnforcedStyle: require_parentheses_when_complex
```

jordan-brough:
Thanks @swcraig.  My personal opinion is that I'd rather turn off purely-stylistic rules like this completely.

jhawthorn:
Spec failure here can't be related

Issue #2122, opened 2017-07-27
Prioritizer pulls from third stock location when there is enough in first + second


bradwbradw:
### Steps to reproduce
- set up 3 stock locations, override `sort_packages` so that you have control over which one comes last.  For brevity, they can be named "A", "B", and "C", and prioritized in that order.
- set up the stock counts of a variant as such: A: 2, B: 4, C: 5
- set up an order, and specify that a line item has 5 of the variant

### Expected behavior
Tell us what should happen
- there should be two shipments: 2 coming from A, and 3 coming from B
- the prioritizer should be returning 2 inventory units in A, and 3 in B

### Actual behavior
- there are three shipments: 2 coming from A, 2 coming from B, and 1 coming from C
- the prioritizer is returning 2 inventory units in A, 2 in B, and 1 in C

### System configuration
**Solidus Version**:
`2.0.0`
**Extensions in use**:
none, just common_spree_dependencies

I reproduced this by adding to the prioritizer_spec.rb in the solidus gem directory and running rspec in there (`.rvm/gems/ruby-2.2.5/bundler/gems/solidus-blabla`)

here is a test that is expected to fail that can be added to https://github.com/solidusio/solidus/blob/master/core/spec/models/spree/stock/prioritizer_spec.rb#L86

```

      it '1st has some, 2nd has remaining, 3rd is empty' do
        5.times { build_inventory_unit }

        package1 = pack do |package|
          2.times { |i| package.add inventory_units[i] }
        end
        package2 = pack do |package|
          4.times { |i| package.add inventory_units[i] }
        end
        package3 = pack do |package|
          5.times { |i| package.add inventory_units[i] }
        end

        packages = [package1, package2, package3]
        prioritizer = Prioritizer.new(inventory_units, packages)
        packages = prioritizer.prioritized_packages

        expect(packages.count).to eq 2
        expect(packages[0].quantity).to eq 2
        expect(packages[1].quantity).to eq 3
      end
```

jhawthorn:
Thanks for the issue. I'd guess that the root cause is the same as #583

jhawthorn:
I believe this will be fixed by #2199. I've added the (greatly appreciated) spec here as ec7b241c1c0a7f38adf0aa5d993c0eb2137b204b.

PR #2199, opened 2017-09-01
Replace Stock::Coordinator with Stock::SimpleCoordinator

### LINKED ISSUE #583: Packaging fails if multiple stock locations are required to fulfill quantity
Discovered this while writing tests for #566.

Because of the way the Stock Coordinator, Packer and Prioritizer work, it won't properly split inventory across stock locations if multiple locations are required to fulfill the quantity of a variant (ie. no stock location has the full amount requested, but combined they do). I believe this was a regression in spree 2.4
## Example scenario

Consider packaging an order with 2 of the same variant. In this example we have two stock locations, each with 1 `count_on_hand` of the variant.
- In the `Stock::Coordinator` we have two unallocated_inventory_units, we'll call them `unit1` and `unit2`.
- Next a `Sock::Packer` is made for each stock location to generate the candidate packages. They both create the identical package, containing only `unit1`. Neither contains `unit2` because each only has enough stock for one.
- The `Stock::Prioritizer` removes the unit from the second candidate package and then removes the empty package entirely.
- The `Stock::Coordinator` finally returns a single package with the single inventory unit.

jhawthorn:
This PR introduces `Spree::Stock::SimpleCoordinator`, which is intended to fix the issues with and replace:
* `Spree::Stock::Coordinator`
* `Spree::Stock::Adjuster`
* `Spree::Stock::Packer`
* `Spree::Stock::Prioritizer`

This solves several issues and bugs with the existing Coordinator, where stock coordination would fail despite actually having enough stock. It should also be faster than the existing Coordinator by making use of the new Stock::Availability class.

This fixes #583 and #2122

It is also intended to contain all the business logic for how stock is packaged in the `SimpleCoordinator` class itself (vs split across Packer/Adjuster/Prioritizer). This should make it a better starting point for stores which want to implement their own coordinator class.

## `Spree::StockQuantities`
This introduces `StockQuantities`, a value class representing counts for a number of variants.

This class provides set-like operations to make to make it easier to deal with and reason about stock.

## `Stock::Availability`

This commit adds Spree::Stock::Availability, a class to manage checking stock availability efficiently for a set of Variants and StockLocations.

This is similar to the existing Spree::Stock::Quantifier, but more efficient by checking multiple variants at once. This needs only two queries to check stock for any number of variants and stock_locations
(though passing fewer will have faster queries).

This would be a good candidate to be a pluggable class. It could in theory be overridden to provide oft-requested features like stock reservations and querying other sources (like 3PL or ERP). I haven't done this at this time because the interface isn't the best suited to be generic and instead is just the most convenient for the one place its used now. To do this the class would also have to include
unstocking and restocking.


core/app/models/spree/stock_quantities.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    # Adds two StockQuantities together
+    # @return [Spree::StockQuantities]
+    def +(other)
+      combine_with(other) do |_variant, a, b|
+        (a || 0) + (b || 0)
+      end
+    end
+
+    # Subtracks another StockQuantities from this one
```
mamhoff: Subtracts (typo)

jhawthorn:
@swcraig is working on the `solidus_legacy_stock_system` gem. I think this can be merged before that's 100% finished.

PR #2208, opened 2017-09-08
Run validations in order updater

mamhoff:
We don't validate when we save orders in the order updater. This is problematic because we save associated records, even if they're invalid - leading to validation errors down the line, i.e. when advancing to checkout or trying to modify an order made invalid by its associated records. 

This shows what is necessary in order to run validations in the order updater, which I think would be a good thing.

### DIFF
--- a/core/spec/models/spree/order_contents_spec.rb
+++ b/core/spec/models/spree/order_contents_spec.rb
@@ -307,7 +307,13 @@ describe Spree::OrderContents, type: :model do
   end
 
   context "completed order" do
-    let(:order) { Spree::Order.create! state: 'complete', completed_at: Time.current }
+    let(:order) do
+      Spree::Order.create!(
+        state: 'complete',
+        completed_at: Time.current,
+        email: "test@example.com"
+      )
+    end
 
     before { order.shipments.create! stock_location_id: variant.stock_location_ids.first }
 
--- a/core/app/models/spree/order_merger.rb
+++ b/core/app/models/spree/order_merger.rb
@@ -54,11 +54,13 @@ module Spree
       end
 
       set_user(user)
-      persist_merge
+      if order.valid?
+        persist_merge
 
-      # So that the destroy doesn't take out line items which may have been re-assigned
-      other_order.line_items.reload
-      other_order.destroy
+        # So that the destroy doesn't take out line items which may have been re-assigned
+        other_order.line_items.reload
+        other_order.destroy
+      end
     end
 
     private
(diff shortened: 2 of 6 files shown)


tvdeyen:
Could you please add a changelog entry?

mamhoff:
I will once other people have weighed in and also think this is the right thing to do :)

jhawthorn:
I'm in favour of this behaviour, but a little concerned about performance. What queries are going to be added by validations?

mamhoff:
The updater loads shipments and line items anyways, those will be loaded. Only new or changed records will be validated. I'm guessing we'd see additional queries for variants on changed line items. New addresses will load states and countries. 

I think the benefits of running validations justifies a couple of additional queries. 

Shipments have no validations to speak of.

core/spec/models/spree/order_contents_spec.rb
```diff
@@ -307,7 +307,13 @@
   end
 
   context "completed order" do
-    let(:order) { Spree::Order.create! state: 'complete', completed_at: Time.current }
+    let(:order) do
+      Spree::Order.create!(
```
bbuchalter: Using a factory would have avoided needing to fix this. Maybe we can start using a factory now?

mamhoff: Hey! Good to hear from you :) 

This spec tests something hard to test with a factory if I remember correctly. Let's not do this in this PR.

mamhoff:
This still needs a :+1: from @kennyadsl or someone from the core team. IIRC the decision was to merge it and eat the performance penalty in the interest of debuggability.

cbrunsdon:
We added the 3.0 milestone as we're going to hold off merging this behaviour until a larger version bump. We're worried about how the validations will impact stores for speed and unexpected behaviour.

kennyadsl:
@mamhoff Hey Martin :wave:, we decided to merge this in v2.11 with a preference like `Spree::Config[:run_order_validations_on_order_updater]` and a deprecation warning to start asking people to change the behavior in their apps before releasing v3.0. Is it something you want/can take care of? Otherwise, someone in the core team will handle rebasing and adding this deprecation. Let us know if you have time, thanks!

mamhoff:
@kennyadsl Thanks for flagging me - wow, this is finally getting attention :) I'm really sorry, but at the moment I'm more busy than I'd like to be. If someone else can handle rebasing this, I'd be quite happy actually. :)

kennyadsl:
Closed with #3645

PR #2209, opened 2017-09-08
add Product Orders admin tab

brchristian:
At our company we’ve found that a frequent use case is wanting to know which orders contained a given product. Solidus already has a helpful [`Product#orders`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/product.rb#L56) association, so this PR adds backend views via a tab for seeing all of the completed orders of a given product.

If this functionality seems like it would be helpful to other stores in similar situations, I’m happy to contribute it here as a PR to the backend.

### DIFF
--- a/backend/app/controllers/spree/admin/products/orders_controller.rb
+++ b/backend/app/controllers/spree/admin/products/orders_controller.rb
@@ -1,0 +1,25 @@
+module Spree
+  module Admin
+    module Products
+      class OrdersController < Spree::Admin::BaseController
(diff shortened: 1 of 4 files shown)


jhawthorn:
The order index allows searching by variant? Maybe a shortcut to that would be more appropriate than it's own page.

mamhoff:
Yeah, I agree with @jhawthorn that we don't need an extra page for what is essentially a search.

PR #2211, opened 2017-09-09
Use headless chrome for backend feature tests

jhawthorn:
This required removing two ([broken anyways](https://github.com/solidusio/solidus/pull/2210)) image specs, which we'll probably want to fix or change the actual behaviour of instead of just removing.

This switches our backend feature specs away from the now somewhat unmaintained phantomjs to headless chrome. This might require some extra setup from developers, but should be a more reliable and future proof testing solution (and we might even be able to test against other selenium browsers like gecko and IE).


backend/spec/features/admin/promotion_adjustments_spec.rb
```diff
@@ -188,14 +188,14 @@
     end
 
     it "should allow an admin to create a promo with generated codes" do
-      fill_in "Name", with: "Promotion"
+      fill_in "Name", with: "SAVE SAVE SAVE"
```
tvdeyen: 😆

backend/spec/features/admin/products/edit/images_spec.rb
```diff
@@ -52,85 +52,6 @@
...
-
-      # And then go back to Size S variant, but using Enter key
-      targetted_select2 "Size: S", from: "#s2id_image_viewable_id"
-      find("#s2id_image_viewable_id").send_keys(:return)
-
-      expect(page).to have_content "Size: S"
-      expect(page).to have_field "image[alt]", with: "ruby on rails t-shirt"
-    end
-  end
-
```
tvdeyen: Could we `xit` or even pend these specs and add a note why these fail or don’t test the right thing? I still think this is a great feature, but need a rework someday.

gmacdougall: This spec didn't work, and as such I think it's best we just remove it for now. There are quite a few areas where we could improve the specs in the backend, and I don't think keeping the code around is going to make this any easier.

tvdeyen:
🤗

PR #2212, opened 2017-09-09
use class instead data-hook for gateway partial

ccarruitero:
Given that other fields in the partial have a data-hook and class to allow easy customization, I think will be good that `card name` field have the same

### DIFF
--- a/frontend/app/views/spree/checkout/payment/_gateway.html.erb
+++ b/frontend/app/views/spree/checkout/payment/_gateway.html.erb
@@ -1,12 +1,12 @@
 <%= image_tag 'credit_cards/credit_card.gif', id: 'credit-card-image' %>
 <% param_prefix = "payment_source[#{payment_method.id}]" %>
 
-<p class="field">
+<p class="field card_name">
   <%= label_tag "name_on_card_#{payment_method.id}", Spree.t(:name_on_card) %><span class="required">*</span><br />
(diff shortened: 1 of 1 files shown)


frontend/app/views/spree/checkout/payment/_gateway.html.erb
```diff
@@ -1,9 +1,9 @@
 <%= image_tag 'credit_cards/credit_card.gif', id: 'credit-card-image' %>
 <% param_prefix = "payment_source[#{payment_method.id}]" %>
 
-<p class="field">
+<p class="field" data-hook="card_name">
```
kennyadsl: I think the general feeling is to remove `data-hook` attributes. To follow this path you could use a class instead, like:

```
<p class="field payment-form-card-name">
```

and add a similar class to the other elements for consistency (number and expiration).

You should be able to hook this element with a class as well, right?

ccarruitero: Yeah, there shouldn't  be a problem.

I'm going to update the elements to use class.

kennyadsl:
@ccarruitero Can you please rebase this?

kennyadsl:
@ccarruitero ping 🙂

aitbw:
Closing now that #3182 was merged, thanks @ccarruitero!

PR #2216, opened 2017-09-11
Deprecate .calculators

cbrunsdon:
There are almost no reasonable reasons why these should be called
through the class models the way they are, near as I can tell they're
only ever used for display/form purposes in the backend.


dhonig:
This makes sense.  Going through the Config opposed to the class  further separates the usage from the implementation. +1

PR #2225, opened 2017-09-18
Move editing of shipment method and tracking into JS views and templates

jhawthorn:
This is the first step towards future work on two separate goals:
* Allowing editing shipments fully through ajax without the page being refreshed
* Allowing backend-only shipping methods to be used by admins #1711

For both of these, we need the editing of shipping methods to be renderable by JS (it was previously an ERB view). This also does the same to he tracking method, which made sense to do at the same time since they used to share code.

This creates a Shipment backbone model and collection, and fetches it from the order API endpoint when loading the shipments page.

Two new handlebars helpers were needed to allow this: `human_model_name` which works like `model_name.human` in ruby (but only supports the singular translation for now), and `format_money` which gives access to the existing `Spree.formatMoney` function.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/templates/orders/shipment_tracking.hbs
+++ b/backend/app/assets/javascripts/spree/backend/templates/orders/shipment_tracking.hbs
@@ -1,0 +1,22 @@
+<th>
+  {{ human_attribute_name "spree/shipment" "tracking" }}
+</th>
+<td colspan="4">
+  {{#if editing}}
+    <input type="text" name="tracking" class="fullwidth" value="{{tracking}}">
+  {{else}}
+    {{#if tracking}}
+      {{ tracking }}
+    {{else}}
+      {{ t "no_tracking_present" }}
+    {{/if}}
+  {{/if}}
+</td>
+<td class="actions">
+  {{#if editing}}
+    <button class="js-save fa fa-check no-text with-tip" data-action="save" title="{{ t "actions.save" }}"></button>
+    <button class="js-cancel fa fa-cancel no-text with-tip" data-action="cancel" title="{{ t "actions.cancel" }}"></button>
+  {{else}}
+    <button class="js-edit fa fa-edit no-text with-tip" data-action="edit" title=""{{ t "actions.edit" }}"></button>
+  {{/if}}
+</td>
--- a/backend/app/assets/javascripts/spree/backend/templates/orders/shipment_tracking.hbs
+++ b/backend/app/assets/javascripts/spree/backend/templates/orders/shipment_tracking.hbs
@@ -3,7 +3,9 @@
 </th>
 <td colspan="4">
   {{#if editing}}
-    <input type="text" name="tracking" class="fullwidth" value="{{tracking}}">
+    <form>
+      <input type="text" name="tracking" class="fullwidth" value="{{tracking}}">
+    </form>
   {{else}}
     {{#if tracking}}
       {{ tracking }}
--- a/backend/app/assets/javascripts/spree/backend/views/order/shipment_tracking.js
+++ b/backend/app/assets/javascripts/spree/backend/views/order/shipment_tracking.js
@@ -1,0 +1,43 @@
+Spree.Views.Order.ShipmentTracking = Backbone.View.extend({
+  tagName: 'tr',
+  className: 'edit-tracking',
+
+  events: {
+    "click .js-edit":   "onEdit",
+    "click .js-save":   "onSave",
+    "click .js-cancel": "onCancel",
+  },
+
+  initialize: function(options) {
+    this.render();
+  },
+
+  onEdit: function(event) {
+    this.editing = true;
+    this.render();
+  },
+
+  onSave: function(event) {
+    this.editing = false;
+    this.model.save({
+      tracking: this.$('input[type="text"]').val()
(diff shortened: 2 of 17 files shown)


backend/app/assets/javascripts/spree/backend/views/order/shipment_tracking.js
```diff
@@ -0,0 +1,43 @@
+Spree.Views.Order.ShipmentTracking = Backbone.View.extend({
+  tagName: 'tr',
+  className: 'edit-tracking',
+
+  events: {
+    "click .js-edit":   "onEdit",
+    "click .js-save":   "onSave",
```
tvdeyen: It's very inconvenient one can't submit all our Backbone views with Enter key. We should change that.

If we would listen to `submit form` events instead and wrap the fields into a `form` (and use that magic submit button trick shown above) we can have both behaviours (enter key and click on the button) with the same event handler.

I will submit a PR with the changes to your PR.

backend/app/assets/javascripts/spree/backend/templates/orders/shipment_tracking.hbs
```diff
@@ -0,0 +1,22 @@
+<th>
+  {{ human_attribute_name "spree/shipment" "tracking" }}
+</th>
+<td colspan="4">
+  {{#if editing}}
+    <input type="text" name="tracking" class="fullwidth" value="{{tracking}}">
```
tvdeyen: If we would wrap this in a `form` we could use the `submit` event instead.

```html
<form id="tracking_form">
```

Notice that we need to have that `id` to connect the "save" button with this form as it needs to be outside of the form. But there is [this barely known attribute of submit buttons](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form) called `form`. 

Very handy for our use case!

backend/app/assets/javascripts/spree/backend/templates/orders/shipment_tracking.hbs
```diff
@@ -0,0 +1,22 @@
...
+    {{#if tracking}}
+      {{ tracking }}
+    {{else}}
+      {{ t "no_tracking_present" }}
+    {{/if}}
+  {{/if}}
+</td>
+<td class="actions">
+  {{#if editing}}
+    <button class="js-save fa fa-check no-text with-tip" data-action="save" title="{{ t "actions.save" }}"></button>
```
tvdeyen: In order to use the same `submit` event we need to add 

```html
<button type="submit" form="tracking_form">
```

the `form` attribute is important here.

backend/app/assets/javascripts/spree/backend/views/order/shipping_method.js
```diff
@@ -0,0 +1,52 @@
+Spree.Views.Order.ShippingMethod = Backbone.View.extend({
+  tagName: 'tr',
+  className: 'edit-shipping-method',
+
+  events: {
+    "click .js-edit":   "onEdit",
+    "click .js-save":   "onSave",
+    "click .js-cancel": "onCancel",
+    "change select": "onChange",
```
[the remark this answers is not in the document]
tvdeyen: Just saw, that the `onChange` function is not defined in this view.

backend/app/assets/javascripts/spree/backend/views/order/shipment_tracking.js
Mandily: 👍

tvdeyen:
Just as a heads up. Besides the PR to this PR here I am working on updating all Backbone based views to also support submitting via enter key (where it’s appropriate)

backend/app/assets/javascripts/spree/backend/views/order/shipment_tracking.js
jhawthorn: We can't listen to `submit form` because we can't wrap only one table row in a `<form>`.

backend/app/assets/javascripts/spree/backend/views/order/shipping_method.js
```diff
@@ -0,0 +1,52 @@
...
+
+  onSave: function(event) {
+    this.editing = false;
+    this.model.save({
+      selected_shipping_rate_id: this.$('select').val()
+    }, {
+      patch: true,
+      success: function() {
+        // FIXME: should update page without reloading
+        window.location.reload();
```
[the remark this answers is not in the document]
jhawthorn: Changing the shipping method may change the amount of the order, so the sidebar needs updating. This will eventually be done without a reload, but is outside the scope of this PR (which is the first step towards removing it).

This doesn't introduce the reload, it was moved from [here](https://github.com/jhawthorn/solidus/blob/12565691ca470afd93cc577e3908f1551a8248f1/backend/app/assets/javascripts/spree/backend/shipments.js#L310)

jhawthorn:
I believe I've addressed the suggestions here.

The last 4 commits here are poorly named and will be squashed into earlier commits, but I have kept them separate to ease reviews.

backend/app/assets/javascripts/spree/backend/views/order/shipping_method.js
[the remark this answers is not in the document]
tvdeyen: Thanks for clarification

PR #2227, opened 2017-09-19
remove unused update_params_payment_source method

ccarruitero:
This method [was moved from `Spree::Order::Checkout`](https://github.com/solidusio/solidus/commit/7ec8e0b82d6d9d6c127a9267bd757d949bb04a52), and seems to be unused now that we use [`set_payment_parameters_amount`](https://github.com/solidusio/solidus/blob/7289efd2a2cd112665c03fa5ff360a6d75adbc35/core/lib/spree/core/controller_helpers/payment_parameters.rb#L210) method for the same.

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -787,6 +787,7 @@ module Spree
         @updating_params[:order][:payments_attributes].first[:amount] = total
       end
     end
+    deprecate update_params_payment_source: :set_payment_parameters_amount, deprecator: Spree::Deprecation
 
     def associate_store
       self.store ||= Spree::Store.default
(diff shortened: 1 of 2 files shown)


mamhoff:
I'm all for this. However, we don't just delete methods like that. Please deprecate it, and we can remove it in a future version. People might have used this method in their host apps, and just removing it without a warning would be a terrible dev experience.

core/app/models/spree/order.rb
```diff
@@ -782,11 +782,12 @@ def process_payments_before_complete
     #   }
     #
     def update_params_payment_source
-      if @updating_params[:order] && (@updating_params[:order][:payments_attributes] || @updating_params[:order][:existing_card])
+      if @updating_params&[:order] && (@updating_params&[:order]&[:payments_attributes] || @updating_params&[:order]&[:existing_card])
```
mtomov: Better figure out how to set some dummy data to this instance variable `@updating_params` in the test rather than changing the method. Given that it hasn't been tested so far, any change can break it really.

PR #2229, opened 2017-09-19
Do a solidus 1.4 roll-up migration

cbrunsdon:
The logical break between solidus 1.4 and 2.0 means we can take some
liberties with some clean up in a pretty safe way.

This rolls up the migrations (something we haven't done for the last 5
years), cleaning up our migration directory and importing fewer
migrations into new stores.

It also has the added benefit of not needing to run through the entire
migration history for a test_app, or having to maintain any of the old
migrations.


core/db/migrate/20160101010000_solidus_one_four.rb
```diff
@@ -0,0 +1,1153 @@
...
+  def up
+    # This migration is just a compressed version of all the previous
+    # migrations for spree_core. Do not run it if one of the core tables
+    # already exists. Assume the best.
+    return if table_exists?(:spree_addresses)
+
+    # this table should not technically exist in the database (as its provided by auth_devise),
+    # but spree_api depends on it existing. This defininition comes from solidus_auth_devise's first migration,
+    # and creates a table equivolent to it
+    create_table "spree_users", force: true do |t|
```
tvdeyen: Can't we fix `solidus_api` instead? It should not rely on this table.

cbrunsdon: Discussed IRL, not reasonable to do with this PR (which doesn't change behavior). We'll need to change behavior to fix it elsewhere.

tvdeyen: 👌🏻

PR #2230, opened 2017-09-20
Switch factories to strings instead of constants

cbrunsdon:
When we define the factories we don't actually care if the constants are
resolved to real objects or not

### DIFF
--- a/core/spec/lib/spree/core/validators/email_spec.rb
+++ b/core/spec/lib/spree/core/validators/email_spec.rb
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'spree/core/validators/email'
 
 describe EmailValidator do
   class Tester
--- a/core/lib/spree/testing_support/factories/address_factory.rb
+++ b/core/lib/spree/testing_support/factories/address_factory.rb
@@ -3,7 +3,7 @@ require 'spree/testing_support/factories/country_factory'
 require 'twitter_cldr'
 
 FactoryGirl.define do
-  factory :address, class: Spree::Address do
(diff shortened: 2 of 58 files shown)


core/spec/lib/spree/core/validators/email_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'spree/core/validators/email'
```
kennyadsl: is this supposed to stay here?

mtomov:
Say again, what is the advantage of using strings? 

I was thinking that using real objects sort of adds a validation check that there really exists this object, which I thought might be good.

tvdeyen:
One of the advantages is that the constant doesn't have to be loaded when loading the factories, only when you actually use it.

I get the validation argument and first thought the same, but as our own test suite uses all of these factories we have confidence that they are valid ;)

mtomov:
Makes sense, thanks!

core/spec/lib/spree/core/validators/email_spec.rb
cbrunsdon: I split it into another commit. Long story short, without the order const loaded, EmailValidator isn't properly loaded and now that the order const isn't loaded when its factory is, this isn't required anymore.

kennyadsl:
For the validation argument, we also have [specs for factories](https://github.com/solidusio/solidus/tree/3f485cf0cddf37085183a5d55976712d011b2e30/core/spec/lib/spree/core/testing_support/factories).

PR #2232, opened 2017-09-20
Use number field for decimal preference field

swcraig:
There are two benefits to this. The UX benefit of the
increment/decrement widget on the input field itself, and now a user can
no longer enter text into this field (which is rejected upon submit and
the preference value can be reset to 0).

Adding `step: :any` allows both integers and floats to be entered, with the widget stepping up and down by 1.

I know that the preference_field_tag is deprecated, but this threw me off when I was testing something related.


kennyadsl:
@tvdeyen good catch! I also found another PR (#1821) that adds other thoughts to this topic. I'm going to remove my approval since I feel like I need to revisit this changes better.

jhawthorn:
My feeling from working on #1821 is that number inputs are terrible for decimal inputs. Having the input step by 1 isn't desirable for, say, entering a tax rate. The bad browser behaviour w/r/t locale only makes this all worse.

I'd rather continue using `input type=text`, even if it's semantically less correct.

mamhoff:
Closing due to localization issues with browser vendors.

PR #2233, opened 2017-09-20
Remove Postal Code Format Validation (and Twitter CLDR dependency)

mamhoff:
The removes the validation of postal code format. 

This validation is done through a pretty expensive dependency, `twitter_cldr`, and actually provides little tangible benefit: We don't know whether the postal code enters actually exists if it passes, we only check the format through a regular expression. 

Stores that rely on this functionality are encouraged to validate the zipcode through something like https://github.com/dgilperez/validates_zipcode, which is a much smaller dependency. We could add that to core, and it would keep the functionality and reduce the size of our bundle, but really I question the benefit of this validation entirely.

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -1,5 +1,3 @@
-require 'twitter_cldr'
-
 module Spree
   # `Spree::Address` provides the foundational ActiveRecord model for recording and
   # validating address information for `Spree::Order`, `Spree::Shipment`,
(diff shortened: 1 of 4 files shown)


mamhoff:
The presence validation is still there, two lines above the deleted line.

BenMorganIO:
I like this. For business owners, address verification is done when generating the shipping label and checking fraud.

For me, I can verify the address of orders through [EasyPosts address Verification API](https://www.easypost.com/address-verification-api) and [Stripes Radar](https://stripe.com/radar). This covers malicious users. For non-malicious users, or rather, the ones making honest mistakes, we can detect them through services such as Google Maps.

PR #2236, opened 2017-09-22
[RFC] Pull rails dependency out of core into solidus_rails

cbrunsdon:
Solidus core doesn't need to be a mountable rails engine.

Core only provides its rails engine for use in solidus_frontend, solidus_backend, and solidus_api. By creating a new gem (solidus_rails), and pushing the rails dependencies into it, we can lesson the surface area of core.

This would allow us more room to try other technologies, build smaller modules to enhance core, and allow us to develop significantly faster (without constantly needing a full rails env to do anything). This removes the requirement to run test_app or use the spec/dummy to run tests.

It will also give us a new layer to keep compatibility in with existing _frontend, and _backend while continuing to move core forward (think: pulling out pre-complete-states) 

You can see from the db_helper.rb, and the inclusion of standalone_migrations how simple it is to continue to maintain activerecord migrations in core.

This transition will take a while as we move all dependencies to solidus_rails, but should be pretty seamless for anyone currently relying on frontend or backend.

Big TODOs:

- [ ] Pull mailers out of core
- [ ] Move the rest of app/models to /lib (and fix requires)
- [ ] Move Spree::Core::Engine's work into solidus_rails (or continue to deprecate)
- [ ] Move the few assets in core to solidus_rails (spree.js.erb)

Happy for feedback or to discuss this change here, on slack, or over a hangout if anyone wanted to talk face to face.


solidus_rails/solidus_rails.gemspec
```diff
@@ -0,0 +1,26 @@
+# encoding: UTF-8
+require_relative '../core/lib/spree/core/version.rb'
+
+Gem::Specification.new do |s|
+  s.platform    = Gem::Platform::RUBY
+  s.name        = 'solidus_rails'
+  s.version     = Spree.solidus_version
+  s.summary     = 'Provides the rails glue for the legacy solidus frontend/backend/api'
```
mamhoff: Do we want to call these `legacy` just yet? :)

core/lib/spree.rb
```diff
@@ -1,4 +1,28 @@
...
+  # Used to configure Spree.
+  #
+  # Example:
+  #
+  #   Spree.config do |config|
+  #     config.track_inventory_levels = false
+  #   end
+  #
+  # This method is defined within the core gem on purpose.
+  # Some people may only wish to use the Core part of Spree.
```
mamhoff: s/Spree/Solidus ?

solidus_rails/solidus_rails.gemspec
```diff
@@ -0,0 +1,26 @@
...
+
+  s.files        = `git ls-files`.split("\n")
+  s.require_path = 'lib'
+
+  s.required_ruby_version     = '>= 2.2.2'
+  s.required_rubygems_version = '>= 1.8.23'
+
+  s.add_dependency 'solidus_core', s.version
+
+  s.add_dependency 'rails', '~> 5.1.0'
```
BenMorganIO: This can be split out to just actionmailer and activerecord.

cbrunsdon: I will once the deps are actually removed, aye. There are still lots of rails calls inside the codebase (and we still have the engine in there).

jhawthorn:
My initial feeling towards this is :-1:

I don't see any benefit to making these changes. Furthermore I expect this will be confusing to users as it is already confusing to me.

Even after all these changes, `solidus_core` still depends on `rails`. Making the split pretty ineffective. `solidus_core` is always going to depend on `activerecord`, so it seems like a better first step would be reducing the requirements of `solidus_core`. Since `solidus_core` will never be without `activerecord` it might as well continue being an engine. If this is just for spec speed purposes we can just avoid requiring the engine if that's the goal here; there's no need for the complexity of an extra gem.

As for the `controller_helpers`, it would be simpler for everyone to keep them in `solidus_core`, but simply not require them by default. I think that would be a good change.

cbrunsdon:
Yea, it'll be a fair number of more changes before I can actually remove the rails dep from solidus_core, but without this work that'll never be possible of course.

core/script/rails
```diff
@@ -1,8 +0,0 @@
-#!/usr/bin/env ruby
-# This command will automatically be run when you run "rails" with Rails 3 gems installed from the root of your application.
-
-ENGINE_ROOT = File.expand_path('../..', __FILE__)
-ENGINE_PATH = File.expand_path('../../lib/spree/core/engine', __FILE__)
-
-require 'rails/all'
-require 'rails/engine/commands'
```
BenMorganIO: We should probably cherry-pick this out into another PR.

core/app/models/spree/calculator/shipping/flexi_rate.rb
```diff
@@ -1,30 +1,32 @@
...
-          if (max == 0 && i == 0) || (max > 0) && (i % max == 0)
-            sum += preferred_first_item.to_f
-          else
-            sum += preferred_additional_item.to_f
+        def compute_from_quantity(quantity)
+          sum = 0
+          max = preferred_max_items.to_i
+          quantity.times do |i|
+            # check max value to avoid divide by 0 errors
+            if (max == 0 && i == 0) || (max > 0) && (i % max == 0)
```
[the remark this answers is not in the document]
tvdeyen: Addressed this rubocop annoyance with https://github.com/solidusio/solidus/pull/2314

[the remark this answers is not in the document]
cbrunsdon: I'll end up pulling this commit out, as I sent it separably as #2312

cbrunsdon:
Here is a quick diagram to help illustrate what this PR is working to achieve:
image

The goal is for solidus_core to provide ecommerce functionality independently from what web framework we're using to deliver the content.

Anecdotally, this removes 20 more gems from the requirements of solidus_core.

More importantly, it lets me use and work with the solidus models without also requiring all of rails, a huge benefit to me when I'm developing stores and libraries locally.

core/script/rails
cbrunsdon: Just pulled this out entirely, can be dealt with later

core/lib/spree.rb
cbrunsdon: Was just a move, but updated.

solidus_rails/solidus_rails.gemspec
cbrunsdon: Updated

BenMorganIO:
@cbrunsdon oh man, looking at that, I feel like Solidus Core should be ActiveSolidus and Solidus Rails should be ActionSolidus XD

cbrunsdon:
Here are some timings for running the state_spec.rb 5 times in three different environments:

1) Solidus Core
2) My Branch, using my rails_helper
3) My branch, using my db_helper

In host applications we can expect the same spec would take significantly longer (9+ seconds each) as we'd be booting up entire rails applications. 
```
for i in `seq  5`; do bundle exec rspec spec/models/spree/state_spec.rb; done;
```

Solidus Core:
`./state_spec.sh  11.54s user 1.66s system 99% cpu 13.306 total`

My Branch, rails_helper.rb:
`./state_spec.sh  7.23s user 0.84s system 98% cpu 8.172 total`

My branch, db_helper.rb:
`./state_spec.sh  4.74s user 0.53s system 98% cpu 5.348 total`

core/lib/spree.rb
```diff
@@ -0,0 +1,34 @@
+require 'active_support/core_ext/module/attribute_accessors'
```
jhawthorn: We can't name a file spree.rb, because of [the other file of the same name](https://github.com/solidusio/solidus/blob/master/lib/spree.rb)

cbrunsdon: Hmmm, I like the idea of a `spree.rb` defining the `Spree` module. (independent of the rest of these changes). Can we look at killing the other one? 

I am not a big fan of the "solidus" gem itself doing much.

jhawthorn: That file doesn't do much, but we should definitely deprecate it before removing it, various extensions and likely stores, which have been upgraded from spree, are expecting that require to require all of Solidus.

cbrunsdon: I am not super worried about deleting the original without deprecating as I don't think it would create any runtime errors in stores, and could be caught and fixed very easily by anyone reading the changelog.

But now that i sit down and think about it, I'm not 100% sure how/when solidus is usually required into a project.

Anecdotally I grep'd my code folder and we don't have any projects, either inherited or greenfield that do a `require 'spree'`.

cbrunsdon: Yea, I just checked and it looks like we usually get solidus required as part of:
`Bundler.require(*Rails.groups)` in config/application.rb's.

I doubt anyone is going to get caught by this, even if they are an old spree upgrade.

kennyadsl:
Closing this one since there seems to be no one anymore interested and it's a big change that requires some ownership at least.

PR #2239, opened 2017-09-27
[core] remove rails and use only actionmailer and activerecord

BenMorganIO:
Before:

```shell
$ bundle
Bundle complete! 30 Gemfile dependencies, 126 gems now installed.
```

After:

```shell
$ bundle
Bundle complete! 30 Gemfile dependencies, 123 gems now installed.
```

---

What I particularly like about this change is that when we move to 5.2, we won't get active storage out of anywhere. We'll get to add it in when we're ready.

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -19,6 +19,10 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = '>= 2.2.2'
   s.required_rubygems_version = '>= 1.8.23'
 
+  %w[activerecord actionmailer].each do |rails_dep|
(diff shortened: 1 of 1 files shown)


tvdeyen:
Already dealt with in #2236

But thanks. Pulling out Rails out of core seems like a good idea.

BenMorganIO:
Not really, but I'll leave a comment on that this change is totally possible within that branch.

BenMorganIO:
@tvdeyen @jhawthorn After the closing of https://github.com/solidusio/solidus/pull/2236#issuecomment-334582147 I think this can be reopened and rebased.

These changes will help remove actioncable and the rails meta gem.

PR #2243, opened 2017-09-29
Improve API key on user edit page

jhawthorn:
This improves a few aspects of how API keys are handled and displayed on the user edit page.

* I18n translations for this page moved out of the API gem, and keys named relative to the view.
* Confirmation dialogs added for clearing or regenerating the key.
* API key is displayed for admins viewing their own user page
* Fixed changed bad heading translation ("Stores -> User")
* Embolden "**Key:**" and italicize "_(hidden)_"

### DIFF
--- a/backend/app/views/spree/admin/users/edit.html.erb
+++ b/backend/app/views/spree/admin/users/edit.html.erb
@@ -35,13 +35,8 @@
         <div id="current-api-key"><%= t('.key') %>: (<%= Spree.t('.hidden') %>)</div>
       </div>
       <div class="filter-actions actions">
-        <%= form_tag spree.clear_api_key_admin_user_path(@user), method: :put do %>
-          <%= button t('.clear_key') %>
-        <% end %>
-
-        <%= form_tag spree.generate_api_key_admin_user_path(@user), method: :put do %>
-          <%= button t('.regenerate_key') %>
-        <% end %>
+        <%= button_link_to t('.clear_key'), spree.clear_api_key_admin_user_path(@user), method: :put, data: { confirm: t('.confirm_clear_key') }, class: 'btn btn-primary' %>
+        <%= button_link_to t('.regenerate_key'), spree.generate_api_key_admin_user_path(@user), method: :put, data: { confirm: t('.confirm_regenerate_key') }, class: 'btn btn-primary' %>
       </div>
 
     <% else %>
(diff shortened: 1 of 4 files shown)


backend/app/views/spree/admin/users/edit.html.erb
```diff
@@ -31,8 +31,14 @@
     <legend><%= t('.api_access') %></legend>
 
     <% if @user.spree_api_key.present? %>
-      <div class="field">
-        <div id="current-api-key"><%= t('.key') %>: (<%= Spree.t('.hidden') %>)</div>
+      <div id="current-api-key">
+        <strong><%= t('.key') %>: </strong>
+        <% if @user == try_spree_current_user %>
```
tvdeyen: Maybe for easier customisation we should add a

```rb
can? :reveal_api_key, @user
```

permission?

tvdeyen:
Somehow my comment is on an outdated commit.  Copy that here:

> Maybe for easier customisation we should add a
> ```rb
> can? :reveal_api_key, @user
> ```
> permission?

jhawthorn:
@tvdeyen 

> Maybe for easier customisation we should add a
> 
> can? :reveal_api_key, @user
> permission?

Because our default "admin" role is assigned `SuperUser` permissions they will be able to view all api_keys, a bigger change than I intended. Are we okay with that?

tvdeyen:
> Because our default "admin" role is assigned SuperUser permissions they will be able to view all api_keys, a bigger change than I intended. Are we okay with that?

Revealing should only be possible for the owner. But I’m fine with Super Admins being able to reset the key. WDYT?

jhawthorn:
> Revealing should only be possible for the owner. But I’m fine with Super Admins being able to reset the key. WDYT?

With the `SuperUser` permission, that isn't possible. We give the "admin" role the `:manage` permission, which allows all actions. https://github.com/solidusio/solidus/blob/master/core/lib/spree/permission_sets/super_user.rb#L5

tvdeyen:
> With the SuperUser permission, that isn't possible. We give the "admin" role the `:manage` permission, which allows all actions.

Right, the super user thing... Something we should think about changing in the future.

Ok, for now we should do what you are proposing in this PR and compare the ownership right inside the view for the revealing. For clearing and resetting I think we could use a permission.

PR #2244, opened 2017-09-29
Add form submit events to several backbone views

tvdeyen:
Lots of our backbone based table rows do have input fields. Users are used to be able to submit changes by hitting the enter key inside a form field.
This is default browser behaviour and should be supported by all our backbone based table row views.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/shipments.js
+++ b/backend/app/assets/javascripts/spree/backend/shipments.js
@@ -3,7 +3,8 @@
 var ShipmentAddVariantView = Backbone.View.extend({
   events: {
     "change #add_variant_id": "onSelect",
-    "click .add_variant": "onAdd"
+    "click .add_variant": "onAdd",
+    "submit form": "onAdd"
   },
   onSelect: function(e) {
     var variant_id = this.$("#add_variant_id").val();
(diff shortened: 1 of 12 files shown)


mtomov:
Thanks!

backend/app/assets/javascripts/spree/backend/calculator.js
```diff
@@ -14,6 +17,14 @@ Spree.CalculatorEditView = Backbone.View.extend({
...
       $(this).toggle(selected);
     });
+  },
+
+  // Somehow backbone seems to prevent events from bubbling up the DOM
+  submit: function(e) {
+    var code = e.keyCode || e.which;
+    if (code === 13) {
+      this.$el.closest('form').submit();
+    }
```
jhawthorn: I'd rather find a different solution to this. I believe I've found the issue and we should address it in a separate PR. It's not a backbone issue.

The is that the "Add" buttons from the Tiered calculators, which aren't submit buttons, but their disabled state causes chrome to not submit the form on "Enter". This can probably be better solved by making these buttons an `<a class="btn"` instead.

tvdeyen: Great. I’m more than happy to remove this

tvdeyen: Converted the `button`s into `a`nchors. The actual problem was that hitting the enter key added new tiers and that prevents the form from being submitted.

tvdeyen:
@jhawthorn review again, please

tvdeyen:
Rebased with latest `master` and fixed conflicts

PR #2247, opened 2017-10-04
Change how shipping methods are selected from the admin

jhawthorn:
Not quite ready for review

This will allow admins to select backend-only shipping methods.


jhawthorn:
Will re-create

PR #2248, opened 2017-10-04
Fix class definition of variant

cbrunsdon:
We shouldn't have multiple inconsistent definitions of the same class.


tvdeyen:
Short term merge, as there is no breaking behaviour. Quite contrary

PR #1402, opened 2016-08-24
adding the promotion categories to the subnav

mgharbik:

PR #2250, opened 2017-10-04
Fix Admin Homepage Spec Link Expectations

mamhoff:
Prior to this commit, we were using a terse syntax for
testing the admin navigation that also did not work.

This changes the syntax to be more readable and actually
testing the links in question.

This is extracted from #1402 by @mgharbik.


mamhoff:
This is taken from #1402, which by itself was stale but contained good stuff.

mtomov:
super!

PR #2125, opened 2017-07-28
Remove order association from inventory units

jhawthorn:

PR #2256, opened 2017-10-04
Remove order association from inventory units

mamhoff:
This is the third attempt to remove the `order_id` column from the `spree_inventory_units` table. It is redundant, as Inventory Units are associated with Shipments, which also have an order ID. 

This will slow down some operations, as we're removing what is essentially a cache column from inventory units. It will also speed up some operation, as we are getting rid of a column with an indexes. 

Most importantly, this simplifies our data model.

Compared to the previous attempts, this is simplified since work on the stock system has removed the necessity for packages and packers to be initialized with orders.

Previous related PRs: #1770 , #2124, #2125.

### DIFF
--- a/core/lib/spree/testing_support/factories/inventory_unit_factory.rb
+++ b/core/lib/spree/testing_support/factories/inventory_unit_factory.rb
@@ -5,8 +5,18 @@ require 'spree/testing_support/factories/shipment_factory'
 
 FactoryBot.define do
   factory :inventory_unit, class: 'Spree::InventoryUnit' do
+    transient do
+      order nil
+    end
+
     variant
-    line_item { build(:line_item, variant: variant) }
+    line_item do
+      if order
+        build(:line_item, variant: variant, order: order)
+      else
+        build(:line_item, variant: variant)
+      end
+    end
     state 'on_hand'
     shipment { build(:shipment, state: 'pending', order: line_item.order) }
     # return_authorization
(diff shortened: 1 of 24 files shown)


core/lib/spree/testing_support/factories/inventory_unit_factory.rb
```diff
@@ -5,8 +5,18 @@
 
 FactoryGirl.define do
   factory :inventory_unit, class: 'Spree::InventoryUnit' do
+    transient do
+      order nil
```
tvdeyen: I would consider this wrong as this implies (from reading the specs) that we could set an order on a inventory unit, what we recently prohibited.

The shipment already has the order. Could we change the specs that we set the correct order on the shipments instead?

core/spec/models/spree/stock/package_spec.rb
```diff
@@ -165,7 +165,7 @@ def build_inventory_unit
 
           it "returns an order" do
             expect(subject.order).to be_a_kind_of Spree::Order
-            expect(subject.order).to eq unit.order
+            expect(subject.order).to eq order
```
tvdeyen: could we amend or at least correct the commit message?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Solidus 2.5.0 (master, unreleased)
 
+- Replace order foreign key from `Spree::InventoryUnit` with a delegate to the inventory unit's `Spree::Shipment`. `Spree::Stock::Package` and `Spree::Stock::Packer` now have to be initialized with a mandatory order argument. [\#2125](https://github.com/solidusio/solidus/pull/2125) ([jhawthorn](https://github.com/jhawthorn), [mamhoff](https://github.com/mamhoff))
```
tvdeyen: `Spree::Stock::Package` and `Spree::Stock::Packer` has not been touched in this PR

Issue #1861, opened 2017-04-23
Prevent user from deleting the default master variant price


robinboening:
### Steps to reproduce
1. Create a product
2. Go to `/admin/products/xyz/prices`
3. Delete the one and only price

### Expected behavior
The one and only price should not be deletable.

### Actual behavior
I can just delete the one and only price. When I go back to "product details" I can see its still using that (paranoidly deleted) price as its default price, but I am not able to edit the record in the prices#edit view anymore.

### System configuration
**Solidus Version**: 2.3.0.alpha

**Extensions in use**: None

laurawadden:
working on it

laurawadden:
@mamhoff mentioned that you actually *should* be able to delete the master price.

However, it is a bug that when you go back to "product details", the master price is still shown (even though it's already deleted). Expected behaviour should be that the field should be blank.

jasonfb:
my colleague @reidcooper and I literarily just noticed this yesterday — spree_prices seems to have acts_as_paranoid on it (using a soft delete with a ```deleted_at``` field), although for some reason Spree's code somehow works around the gem's mechanism. If you make a change to a price after deleting it, it still updates the "deleted" (soft) record in the database.

PR #2262, opened 2017-10-04
Add required to product master price field

gregdaynes:
This adds `required: true` to the master price field on the products
form. The label for the field indicates that it is required, but the
field can be cleared, and the form submitted - resulting in the price
being defaulted to `0.00`

This is part of issue #1861. Though in the issue report it talks of the
product price showing a deleted value when removed (which has been
resolved). This further makes it a conscious decision to set the price
to `0`.

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -31,8 +31,10 @@
     <div class="right col-3" data-hook="admin_product_form_right">
       <div data-hook="admin_product_form_price">
         <%= f.field_container :price do %>
-          <%= f.label :price, class: 'required' %>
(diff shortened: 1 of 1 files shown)


backend/app/views/spree/admin/products/_form.html.erb
```diff
@@ -32,7 +32,7 @@
       <div data-hook="admin_product_form_price">
         <%= f.field_container :price do %>
           <%= f.label :price, class: 'required' %>
-          <%= render "spree/admin/shared/number_with_currency", f: f, amount_attr: :price,
+          <%= render "spree/admin/shared/number_with_currency", f: f, amount_attr: :price, required: true,
```
jhawthorn: We have a `Spree::Config.require_master_price` (which I don't know how widely is used but exists).

Can we change this line to `required: Spree::Config.require_master_price`?

PR #2263, opened 2017-10-04
Create multiple page states for stock transfers

jhawthorn:
Rebase and modifications to #1963


jhawthorn:
Modifications made from #1963:

* Stock transfer table should be shown if there are `>= 2` stock locations. Not just `> 0`
* Filters/search should always be shown, regardless of number of stock transfers

PR #2265, opened 2017-10-04
Cancel authorized (pending) payments when cancelling an order

gmacdougall:
E.g., release funds on a credit card auth for cancelled orders whose
payments have been authorized but not yet captured.

This was already happening for store credit payments.

I tried to think of a reason not to void `pending` payments when cancelling an order and couldn't think of any reason.  Perhaps most people simply capture at order complete so this hasn't been an issue? And/or they don't worry about releasing authorizations on people's credit cards?

Closes #1529

This is a rebased version of #1529

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -852,8 +852,9 @@ module Spree
 
     def after_cancel
       shipments.each(&:cancel!)
-      payments.completed.each { |payment| payment.cancel! unless payment.fully_refunded? }
-      payments.store_credits.pending.each(&:void_transaction!)
+      payments.select do |payment|
+        (p.pending? || p.completed?) && !payment.fully_refunded?
+      end.each(&:cancel!)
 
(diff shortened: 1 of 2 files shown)


kennyadsl:
@gmacdougall we'd like to have this one in v2.11 under a preference that users should opt-in for, with a deprecation warning. Moving to the 2.11 Milestone.

seandawes:
This is a good one.

One thing to note is @peterberkenbosch looked at this issue in regards to orders having more than 1 pending (authorized) payment on an order dealing with Affirm. We experience this all the time when a user changes method, cart total etc we have orders with numerous pending approvals. 

So potential this could be stacked where we address multiple authorizations first and the auto cancel just needs to cancel the last one in the flow?

Adding @modded-nate from our team here

kennyadsl:
Closing, see https://github.com/solidusio/solidus/pull/3662

PR #2268, opened 2017-10-05
Sticky admin header

### LINKED ISSUE #630: Move page title and action buttons into a sticky header

tvdeyen:
Closes #630

localhost-3000-admin-stock_items

localhost-3000-admin-products 1

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_pagination.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_pagination.scss
@@ -1,0 +1,5 @@
+.pagination .page-item.active .page-link {
+  // overwrite a weird `z-index: 2` from Bootstrap,
+  // so it does not peak out of the sticky header
+  z-index: auto;
+}
--- a/backend/app/assets/stylesheets/spree/backend/globals/mixins/_line_through.scss
+++ b/backend/app/assets/stylesheets/spree/backend/globals/mixins/_line_through.scss
@@ -10,13 +10,11 @@
     width: 100%;
     height: 1px;
     background: $color;
-    z-index: 0;
   }
 
   > * { // usually a <span> but we'll use an * just in case
     @include padding(null 1rem);
     position: relative;
     background: $background;
-    z-index: 1;
   }
 }
--- a/backend/app/assets/stylesheets/spree/backend/shared/_header.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_header.scss
@@ -1,10 +1,14 @@
 .main-header {
   display: flex;
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: $main-header-height;
   align-items: center;
-  padding: 15px $grid-gutter-width;
+  padding: 15px $grid-gutter-width 15px $width-sidebar + $grid-gutter-width;
   background-color: very-light($color-3, 4);
   border-bottom: 1px solid $color-border;
-  height: $main-header-height;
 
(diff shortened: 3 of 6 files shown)


tvdeyen:
@Mandily @graygilmore FYI

kennyadsl:
@tvdeyen Great! We don't need to take into account smaller screen versions for now right?

kennyadsl:
I actually just found an issue with elements inside the `.content-wrapper` that has a z-index applied, like datepicker:

schermata 2017-10-05 alle 16 16 18

If I got it correctly, even if we respect the stacking order and give more priority to an element pushing it at the end of the list there could be other child elements inside previous elements with z-index (new stacking context) that would take precedence in display order.

I made a little codepen to simplify the situation: https://codepen.io/kennyadsl/pen/vepxwO

jhawthorn:
Code seems fine, but I'm not a fan of having a sticky header. I don't think users benefit from having the information in the header visible at all times. That information is less important than that on the rest of the page, and it eats quite a bit of screen real estate.

I also see some z-index issues.

Mandily:
For a little more context - the idea here was that the action buttons would always be in view and accessible by putting them in the header. Create actions like "new product" would live there as well as save and cancel buttons once a user was in those new or edit views.

As for small/mobile screens, we could do the fancy thing of scrolling down the header goes away but scrolling up brings it back into view.

mtomov:
I'm also of the opinion that screen estate for content is more important than saving a _scroll up_ to do a button press. The buttons, which sit at the top right, in my view, are actually not as often used, as the actions related to the general peruse of the data.

tvdeyen:
> I don't think users benefit from having the information in the header visible at all times. That information is less important than that on the rest of the page, and it eats quite a bit of screen real estate.

The header mostly contains breadcrumbs and is therefore a navigation bar, not an information bar. I do believe that users will benefit from having the header navigation sticking at the top, like we have the main navigation on the left side sticking. 

The actual change necessary is not that intrusive. If you are concerned about the change DOM position I could leave it as is and use a very high `z-index` value instead. but using the DOM positioning is strongly recommended, though.

> Create actions like "new product" would live there 

👍 on having navigable buttons ("New product" is actually a navigation action) inside the header is good UX IMO and that's why I consider a fixed header a good thing.

> as well as save and cancel buttons once a user was in those new or edit views.

I would strongly argue against putting "save" and "cancel" buttons into the header, but this is another discussion actually. 

> I also see some z-index issues.

Besides the aforementioned date picker I did not found any more z-index issues. Could you please mention them, so I can investigate?

jhawthorn:
> Besides the aforementioned date picker I did not found any more z-index issues. Could you please mention them, so I can investigate?

Price inputs seem to float above the header.

tvdeyen:
Thanks. Just saw, that this happens even in non sticky header branches:

z-index price fields

Will take a look

tvdeyen:
We decided to not have a sticky header

PR #2270, opened 2017-10-05
Remove remnants of location configured packages

jhawthorn:
As per #1696, we decided to remove the "location configured packages" feature without replacement.

In Solidus 2.2 (#1709) we removed the ability to configure them from the admin
In Solidus 2.4 (#2199) we removed the code which build location configured packages based on the OrderStockLocation records

This PR finishes the removal by removing the ability to create the records through the API (which once created would have no effect anyways) and removes the OrderStockLocation model.


jhawthorn:
> Do we also need to remove the database table?

I've removed the migration creating the table. (I don't think we should make a migration to remove it if it already exists)

PR #2271, opened 2017-10-05
Use HTML5 to prevent spaces in promotion codes

jhawthorn:
Alternative to #1796

This prevents promotions from being created with an accidental whitespace.

This is a UX concern rather than a business logic/technical/security constraint, to it makes sense to keep the solution as close to that level as possible (in the HTML).

This could be made even nicer by listening to keydown events on this input and preventing spaces from being entered, but I think this is a minor and infrequent enough issue for users that that isn't worth the effort.


tvdeyen:
I'm 👎 for the same argument you already mentioned in #1796 

> An admin might really want to enter "SAVE MORE" as their coupon code.

jhawthorn:
ok

PR #2272, opened 2017-10-05
Make solidus_core depend on actionmailer and activerecord instead of rails

BenMorganIO:
We originally closed #2239 because we thought we were going to get it in from #2236 but never did. So... Here's a new PR!

---

Before:

```shell
$ bundle
Bundle complete! 30 Gemfile dependencies, 126 gems now installed.
```

After:

```shell
$ bundle
Bundle complete! 30 Gemfile dependencies, 123 gems now installed.
```

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -19,6 +19,13 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = '>= 2.2.2'
   s.required_rubygems_version = '>= 1.8.23'
 
+  %w[
+    actionmailer actionpack actionview activejob activemodel activerecord
+    activesupport railties
+  ].each do |rails_dep|
+    s.add_dependency rails_dep, '~> 5.1.0'
+  end
+
   s.add_dependency 'activemerchant', '~> 1.48'
   s.add_dependency 'acts_as_list', '~> 0.3'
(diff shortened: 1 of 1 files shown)


cbrunsdon:
At the very least I think we'll need to add activesupport, activejob, and railties, which are the three I know we depend on inside of core.

cbrunsdon:
@BenMorganIO aye, `Rails` is from railties, but we don't have that dependency, we get it through other gems that do explicitely have it listed, which I don't think is good to rely on.

jhawthorn:
> At the very least I think we'll need to add activesupport, activejob, and railties, which are the three I know we depend on inside of core.

We do get these from other gems, but it would make sense to add them to our deps

core/solidus_core.gemspec
```diff
@@ -19,7 +19,10 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = '>= 2.2.2'
   s.required_rubygems_version = '>= 1.8.23'
 
-  %w[activerecord actionmailer].each do |rails_dep|
+  %w[
+    actionmailer actionpack actionview activejob activemodel activerecord
+    activesupport
```
cbrunsdon: railties too please

PR #2276, opened 2017-10-09
Mutating shipments must not mutate line items

Gamabunta94:
This pull request wants to solve this issue: solidusio#2235

If there are two or more shipments for one order and the user delete one of these, the line items of order are modified.
To overcome this problem I removed destroy action from shipment so to force the user to use split action to add, modify or delete shipments.


Gamabunta94:
Please don't consider this PR. Sorry was my fault.

PR #564, opened 2015-12-03
[WIP] Remove after_rollback from LogEntry

jhawthorn:
WIP because I'm not yet 100% certain that removing that has no effect, need to test a bit more.

`after_rollback` has all kinds of problems. When it's run as part of our specs, the filter happens after our spec run (each spec is run inside a transaction), meaning that extra useless blank log entries end up in our database. Gross!

In core's test suite:

```
config.after(:suite){ p Spree::LogEntry.count } # => 149
```

I don't think we ever want this, if we were rolling back a log entry, we would also be rolling back a state change on a payment, which is _bad_.

This is similar to bc19fcb607377fcb02ecf0289ba6430b9d2cffd3

PR #2277, opened 2017-10-09
Issue with Transactional Rollbacks during Test Environments with Spree Log Entries

reidcooper:
I am proposing this fix to the `Spree::LogEntry` model. 

When running my tests with DatabaseCleaner (using `:transaction`), after it asserts my first example using `rspec`, DatabaseCleaner kicks in to transactionally rollback all my test data. Unfortunately, due to this feature, I run into issues where the `after_rollback` is called on the `Spree::LogEntry` model which in turn tries to append some information that does not exist anymore since DatabaseCleaner had its way and cleared out my test data. This causes an issue where the record of a variant is not found anymore. 

For those who use DatabaseCleaner, to ignore trying to create new Spree::LogEntries during test environments. I have attached a gist of a log with a backtrace of my `byebug` statement. Notice at the top of the gist, `BEFORE @lpes_calculator; AFTER @lpes_calculator` is wrapped around my `expect` statement.

I am open to potentially better solutions.

https://gist.github.com/reidcooper/2e09d5dc21f5783b26d0ae7a633a22f5

### DIFF
--- a/core/app/models/spree/log_entry.rb
+++ b/core/app/models/spree/log_entry.rb
@@ -4,7 +4,7 @@ module Spree
 
     # Fix for https://github.com/spree/spree/issues/1767
     # If a payment fails, we want to make sure we keep the record of it failing
-    after_rollback :save_anyway, if: Proc.new { !Rails.env.test? }
+    after_rollback :save_anyway, if: proc { !Rails.env.test? }
 
     def save_anyway
       log = Spree::LogEntry.new
--- a/core/app/models/spree/log_entry.rb
+++ b/core/app/models/spree/log_entry.rb
@@ -4,7 +4,7 @@ module Spree
 
     # Fix for https://github.com/spree/spree/issues/1767
     # If a payment fails, we want to make sure we keep the record of it failing
-    after_rollback :save_anyway
+    after_rollback :save_anyway, if: Proc.new { !Rails.env.test? }
 
     def save_anyway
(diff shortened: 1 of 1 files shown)


mamhoff:
There used to be https://github.com/solidusio/solidus/pull/564, which is a more comprehensive solution to this thing. @jhawthorn What do you think, should we just revive that?

cbrunsdon:
Have we ever talked about having a second storage mechanism for stuff like this, separate from our AR?

We're fighting against our technology with things like the log_entries, and it seems to me it would make more sense to shove it in some kind of document or object storage, its not like we ever actually do anything with it in terms of querying.

jasonfb:
An abstracted storage mechanism — in which stores (no pun intended) could choose where to put their spree_log_entries — makes total sense to us. But it seems to me that for core code to dictate a document or object storage, that would expand the (already tight & small) stack requirements for setting up a new Solidus installation. 

We recognize that this is a small piece of paper mache that covers up a bandage that itself was originally covering up what is (arguably) a code smell. It seem to us the root of the problem is that exceptions (in this case, the rollback) are relied on for too much flow control, which is essentially the root of the code smell. We fully understand this is not a fix to the root of the problem. 

Also note it can be decorated as the decorated ```after_rollback``` seems to over-ride the core code definition. (this worked fine in our app)

So do with that information what you will.

PR #2278, opened 2017-10-10
Fix preference_field_tag when given no type

jhawthorn:
Previous to being deprecated, this [defaulted to](https://github.com/solidusio/solidus/blob/v2.3/backend/app/helpers/spree/admin/base_helper.rb#L67-L68) `:text` if no type was given. This PR restores that behaviour.


jhawthorn:
Found [in `solidus_editor`](https://github.com/solidusio-contrib/solidus_editor/blob/0c1a4e1d6188ed9f0d7a238eea644c92dc2dea4b/app/views/spree/admin/editor_settings/edit.html.erb#L26)

PR #2280, opened 2017-10-11
Remove after_rollback from LogEntry

jhawthorn:
Follow up to #2277 #564 

`after_rollback` has all kinds of problems. When it's run as part of our specs, the filter happens after our spec run (each spec is run inside a transaction), meaning that extra useless blank log entries end up in our database. Gross!

In core's test suite:

```
config.after(:suite){ p Spree::LogEntry.count } # => 149
```

I don't think we ever want this, if we were rolling back a log entry, we would also be rolling back a state change on a payment, which is _bad_.

This is similar to bc19fcb607377fcb02ecf0289ba6430b9d2cffd3

### DIFF
--- a/backend/spec/features/admin/orders/log_entries_spec.rb
+++ b/backend/spec/features/admin/orders/log_entries_spec.rb
@@ -14,7 +14,6 @@ describe "Log entries", type: :feature do
       )
 
       payment.log_entries.create(
-        source: payment.source,
         details: response.to_yaml
       )
     end
--- a/core/app/models/spree/log_entry.rb
+++ b/core/app/models/spree/log_entry.rb
@@ -2,17 +2,6 @@ module Spree
   class LogEntry < Spree::Base
     belongs_to :source, polymorphic: true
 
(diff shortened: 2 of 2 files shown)


jasonfb:
we're not quite sure why referencing a source would cause a rollback. All we can think is for the case where ```payment.source``` has errors on it, it won't have saved correctly.

this fix seems like it will remove the association between spree_log_entries and payments, which will surely make debugging harder, no?

I'd say a more defensive fix would be
```
source_id: (!payment.source.new_record?) ? payment.source.id : nil ,
```

or
```
source_id: (!payment.source.errors?) ? payment.source.id : nil ,
```

or even
```
source_id: (! payment.id.nil?) ? payment.source.id : nil ,
```

This would eliminate the awkward after_rollback construction but also keep the association (when there really is a payment record to associate)

jhawthorn:
> this fix seems like it will remove the association between spree_log_entries and payments, which will surely make debugging harder, no?

This is not being changed.

jasonfb:
Oh ok I misunderstood in that case.

PR #2281, opened 2017-10-12
Replace generated dummy test_app with single file rails application

jhawthorn:
This PR removes the need to run `rake test_app` before running specs. Migrations are automatically run before specs if the database is new or has missing migrations.

```
$ rspec
Re-running migrations
rake db:reset VERBOSE=false
Dropped database 'db/solidus_backend_test.sqlite3'
Created database 'db/solidus_backend_test.sqlite3'

Randomized with seed 14974
..................
```

This is done by adding a checked-in single-file(ish) Rails application requirable at `spree/testing_support/dummy_app`. This replaces the need to generate the `spec/dummy` application, and allows us to make changes to the dummy app's configuration and have them apply without re-running `rake test_app`.

The `rake test_app` task has been replaced by a new `rake db:reset` task, which calls `db:drop`, `db:create`, and `db:migrate` as one would expect. Most users will never have to run this, as the spec_helper for each project will detect the need to re-run migrations and automatically run `rake db:reset`. `test_app` has been kept as an alias for `db:reset`, which will be helpful to keep in the short-term to allow git bisects. Top-level rake tasks have been similarly adjusted.

```
jhawthorn@muta:~/src/solidus/core
$ rake -T
rake console     # Open a sandboxed console in the test environment
rake db:create   # Creates the test database
rake db:drop     # Drops the test database
rake db:migrate  # Migrates the test database
rake db:reset    # Recreates the test database and re-runs all migrations
rake spec        # Run RSpec code examples
```

A new `rake console` command opens up a sandboxed (in a transaction) pry console in the `test` environment. I'm hoping this meets most of the needs of developers who would previously `cd` into `spec/dummy` and fire up `rails c`.

``` ruby
jhawthorn@muta:~/src/solidus/core
$ rake console
Loading test environment in sandbox (Rails 5.1.4)
Any modifications you make will be rolled back on exit
[1] pry(main)> require "spree/testing_support/factories"
[2] pry(main)> order = FactoryGirl.create(:order)
[3] pry(main)> variant = FactoryGirl.create(:variant)
[4] pry(main)> order.contents.add(variant)
=> #<Spree::LineItem:0x0055fb94895220
 id: 1,
 variant_id: 2,
 order_id: 2,
 quantity: 1,
 price: BigDecimal(19.99),
 created_at: Fri, 13 Oct 2017 22:32:55 UTC +00:00,
 updated_at: Fri, 13 Oct 2017 22:32:55 UTC +00:00,
 cost_price: BigDecimal(17.0),
 tax_category_id: 1,
 adjustment_total: BigDecimal(0.0),
 additional_tax_total: BigDecimal(0.0),
 promo_total: BigDecimal(0.0),
 included_tax_total: BigDecimal(0.0)>
[4] pry(main)> exit
   (0.4ms)  rollback transaction
```

Todo:
* [x] Convert `api` and `frontend`
* [x] Investigate entirely eliminating `rake test_app`  (auto run migrations)
* [x] Rebase and make sensible commits
* [x] Prefix databases with project name
* I haven't investigated using this is extensions, but I would like to tackle separately


tvdeyen:
I have some difficulties to build this locally.

### First try in root of Solidus

Pulled fresh branch, ran bundle update and

```
tvd and ruby-2.4.2 in solidus/ on jhawthorn-dummy_app
› be rake spec

-----------
Solidus api
-----------
rake spec
/Users/tvd/.rbenv/versions/2.4.2/bin/ruby -I/Users/tvd/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rspec-core-3.6.0/lib:/Users/tvd/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rspec-support-3.6.0/lib /Users/tvd/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/rspec-core-3.6.0/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
Re-running migrations
rake db:reset VERBOSE=false
Dropped database 'db/solidus_api_test.sqlite3'
Created database 'db/solidus_api_test.sqlite3'

[...]

Failures:

  1) Spree::Api::ReturnAuthorizationsController as the order owner cannot delete a return authorization
     Failure/Error: delete spree.api_order_return_authorization_path(order, 0)

     NoMethodError:
       undefined method `spree_path' for #<Module:0x00007f9ba9236288>
     # ./spec/requests/spree/api/return_authorizations_controller_spec.rb:57:in `block (3 levels) in <module:Spree>'

  [...]
```

#### Fails with hundreds of errors. All the same. Seems to be routes related.

### Second try in `./api`

```
tvd and ruby-2.4.2 in api/ on jhawthorn-dummy_app
› be rake test_app
[Spree WARNING] Missing migrations.
[Spree WARNING] add_lft_and_rgt_indexes_to_taxons from spree is missing.
[Spree WARNING] solidus_one_four from spree is missing.
[Spree WARNING] Run `bundle exec rake railties:install:migrations` to get them.

Database 'db/solidus_api_test.sqlite3' does not exist
Created database 'db/solidus_api_test.sqlite3'
```

#### Prints warnings, but it builds successfully.

```
tvd and ruby-2.4.2 in api/ on jhawthorn-dummy_app
› beer

[Spree WARNING] Missing migrations.
[Spree WARNING] add_lft_and_rgt_indexes_to_taxons from spree is missing.
[Spree WARNING] solidus_one_four from spree is missing.
[Spree WARNING] Run `bundle exec rake railties:install:migrations` to get them.

[...]

Finished in 1 minute 56.92 seconds (files took 4.86 seconds to load)
579 examples, 0 failures, 3 pending

Randomized with seed 1494
```

### The same with `./core`

## Conclusion:

1. Builds successfully for individual tests suites, but fails from root
2. The warning about missing migrations seems to be wrong and we should find a way to avoid it

frontend/spec/controllers/spree/home_controller_spec.rb
```diff
@@ -16,11 +16,12 @@
...
     context "different layout specified in config" do
-      before { Spree::Config.layout = 'layouts/application' }
+      before { Spree::Config.layout = 'layouts/my_super_awesome_layout' }
 
       it "renders specified layout" do
-        get :index
-        expect(response).to render_template(layout: 'layouts/application')
+        expect {
+          get :index
+        }.to raise_error(ActionView::MissingTemplate, %r{Missing template layouts/my_super_awesome_layout})
```
tvdeyen: I think we should not change the spec like this. Is there a way to tell the dummy app the view paths like you did with the assets?

frontend/app/controllers/spree/orders_controller.rb
```diff
@@ -1,7 +1,6 @@
 module Spree
   class OrdersController < Spree::StoreController
     before_action :check_authorization
-    rescue_from ActiveRecord::RecordNotFound, with: :render_404
```
tvdeyen: I think it is very common in stores to overwrite `render_404` and do stuff instead of rendering the 404 page. Building a dynamic 404 page for instance. Is there a way to still provide this functionality for stores? Could you tell the dummy app where to find the views like you did with the assets paths for instance?

frontend/app/controllers/spree/products_controller.rb
```diff
@@ -3,7 +3,6 @@ class ProductsController < Spree::StoreController
     before_action :load_product, only: :show
     before_action :load_taxon, only: :index
 
-    rescue_from ActiveRecord::RecordNotFound, with: :render_404
```
tvdeyen: ditto

api/db/migrate/20100107141738_add_api_key_to_spree_users.rb
```diff
@@ -1,7 +0,0 @@
-class AddApiKeyToSpreeUsers < ActiveRecord::Migration[4.2]
```
cbrunsdon: I think we should deal with these migrations in a separate PR. The weirdness of them came up when I did the roll-up but I didn't want to change any behaviour.

frontend/app/controllers/spree/orders_controller.rb
jhawthorn: I think this should be removed anyways because it's a bad way to override the 404 behaviour.

Stores could regain this functionality by either:
* Using `config.exceptions_app`, the proper way to have a dynamic 404 page
* Setting up the `rescue_from` from `ApplicationController`
* Class-evaling (or `send`ing) the rescue_from (if overrides are being done anyways)

But I can submit this as a separate PR if it will get this merged

jhawthorn: Right. I can't remove this because the render_404 behaviour is very unusual.

``` ruby
render status: :not_found, file: "#{::Rails.root}/public/404", formats: [:html], layout: nil
```

This doesn't use views, it renders the `public/404.html` file relative to the Rails.root directory. There's really no way to get around this, it's bad code.

frontend/spec/controllers/spree/home_controller_spec.rb
jhawthorn: I do, but can send a separate PR

api/db/migrate/20100107141738_add_api_key_to_spree_users.rb
jhawthorn: Removed this change.

Replaced the db:migrate task with a custom migration task which runs migrations one engine at a time (which simulates running `railties:install:migrations` before `db:migrate`).

jhawthorn:
I've addressed most of the feedback.

> Builds successfully for individual tests suites, but fails from root

This should be fixed. We were incorrectly running specs using the top-level Gemfile instead of the individual project Gemfiles. It now runs `bundle exec` from a clean environment under each project.

> The warning about missing migrations seems to be wrong and we should find a way to avoid it

This is because Spree::Migrations is detecting the old dummy_app migrations (it assumes the migrations paths instead of checking the app's config.paths as it should). `rm -Rf spec/dummy` or `rake clean` at the top-level will fix it. We could detect if we're running inside `DummyApp` from `Spree::Migrations`, but I'm not sure that's worth doing.


I was able to re-add the `application/layout`, but can't find a reasonable alternative to `render_404`.

frontend/app/controllers/spree/orders_controller.rb
tvdeyen: TIL `config.exceptions_app` 🎉 

You are right, the code is bad and should be removed. Although it's blocking your PR I think we need a separate PR due to visibility reasons. I think we can forego with deprecating this, though.

PR #2283, opened 2017-10-13
Translate reception states and fix translation key

rbngzlv:
Screen capture:

captura de pantalla 2017-10-13 a las 8 11 27

And the translation key for `Canceled` state is with double `l`: https://github.com/solidusio/solidus/blob/master/core/app/models/spree/return_item.rb#L83

### DIFF
--- a/core/app/models/spree/return_item.rb
+++ b/core/app/models/spree/return_item.rb
@@ -212,7 +212,7 @@ module Spree
       event_paths.delete(:expired)
       event_paths.delete(:unexchange)
 
-      status_paths.map{ |s| s.to_s.humanize }.zip(event_paths)
+      status_paths.map{ |s| I18n.t("spree.reception_states.#{s}", default: s.to_s.humanize) }.zip(event_paths)
     end
 
     def part_of_exchange?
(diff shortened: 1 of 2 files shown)


rbngzlv:
Done!

swcraig:
Just as a note, we use the (generally) American spelling of `canceled` throughout most of our English translations. Should an issue be made to transition into using `cancelled` with two `l`s? It's sort of a nitpick, but we may as well make all of the uses uniform.

[Both spellings seem to be acceptable and correct.](https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#Doubled_in_British_English)

tvdeyen:
The problem is that the state is called `cancelled` in the source code and therefore the translation key also has to be `cancelled`. But we could revert that translation change for sure.

rbngzlv:
Reverted the translation change.

rbngzlv:
Rebased on master and change `Spree.t(:key)` to `I18n.t('spree.key')`

rbngzlv:
@tvdeyen rebased on master. More than three months have passed, is something blocking this small PR from being merged?

tvdeyen:
@rbngzlv we need a second approval. From @kennyadsl maybe? 

Sorry, that this took so long and thanks for reminding.

rbngzlv:
@kennyadsl and @tvdeyen I'm doing a friendly reminder about this PR.

kennyadsl:
This just needed someone to press the button, sorry again for taking so long to merge!

PR #2284, opened 2017-10-13
Fix return url after fire in return_authorizations controller

rbngzlv:


kennyadsl:
@rbngzlv Can you please rebase? No more `Spree.t(:key)` in favor of `I18n.t('spree.key')`

rbngzlv:
@kennyadsl Done!

Issue #2285, opened 2017-10-13
Fulfilment changer uncorrectly restocking and unstocking items


DanielePalombo:
We found some problem when moving backordered inventory units across shipments. Specifically the stock items of source and destination aren't restocking unstocking the correct quantity.

### Steps to reproduce
- Create 2 stock location
- Choose one product and set stock items to quantity 10 and both to backorderable.
- Complete an order that requires 15 inventory units.
- Now you have `10 X on hand` and other `5 X backordered` inventory units, something like:
<img width="850" alt="screen shot 2017-10-06 at 12 00 52" src="https://user-images.githubusercontent.com/387690/31286790-92f87eac-aabf-11e7-8d6f-18a13c78d2c9.png">

- Into product stocks section the related stock item is correctly calculated.
<img width="1099" alt="screen shot 2017-10-06 at 12 01 05" src="https://user-images.githubusercontent.com/387690/31286953-035b00d4-aac0-11e7-9ac8-11c01dae946b.png">

- Return to order page and move all inventory units to a new shipment in the other stock location.
<img width="840" alt="screen shot 2017-10-06 at 12 01 45" src="https://user-images.githubusercontent.com/387690/31287003-245af578-aac0-11e7-882b-90a5d8f16b8b.png">

### Expected behavior
The count on hand of source stock item should be restocked to the original value (10), the count on hand of stock item where you move the inventory units to should be negative (-5).

### Actual behavior
The count on hand on default stock item isn't restored, and the second stock location is 0 but should be -5
<img width="1109" alt="screen shot 2017-10-06 at 12 02 20" src="https://user-images.githubusercontent.com/387690/31287042-4bde5f04-aac0-11e7-825d-458024ef989d.png">

### System configuration
2.4, master

mamhoff:
In my understanding, a stock item can not have a negative count on hand. Backordered items also should not affect the count on hand. What's the scenario you want the negative count on hand for?

DanielePalombo:
Thanks for the fast answer.

If I have backorderable stock item with count on hand 0, I can receive new order for this product, then after the order the related count on hand should be negative. This allow me to understand how many items are out of stock. 

Even in the previous screenshots the count on hand is negative after I made an order with more quantity than availability.

PR #2286, opened 2017-10-13
Fix restocking and unstocking items in FulfilmentChanger

DanielePalombo:
This PR fixes issues described in #2285.

Current code is not taking into account some important scenarios, like:
- Desired stock item as negative count on hand amount
- Restock of inventory in backordered state


mamhoff:
I don't think stock item's should have a negative count on hand. In my understanding, backordered stock items don't affect the count on hand.

mamhoff:
I was wrong, just re-checked the way `Spree::OrderInventory` handles these cases. In fact, negative stock items are desired in order to account for the case that inventory is added and used up by backordered stock items.

PR #2287, opened 2017-10-14
Update translation for Adjustment Label

gregdaynes:
Updating the translation to reflect what the attribute is called to correct the error message when providing attributes to create. `Spree::Adjustment.create!` requires a description with message
'Description can't be blank', providing it using `description: 'string'` outputs "unknown attribute 'description' for Spree::Adjustment."


gregdaynes:
Tests have been fixed to account for new translations - Sorry, I should have commented earlier.

Issue #2251, opened 2017-10-04
Make cart total/subtotal background color configurable


mamhoff:
As identified by @akihitoalextsuboi in #1708, we have a hardcoded color variable for the background of the cart subtotal and total. We should change this to be a configurable, well-named variable. 

A good variable name would be `$cart-total-background-color`.

kennyadsl:
It should also probably take into account `$product_price_text_color` now.

gregdaynes:
I'll look into this later today.

PR #1708, opened 2017-02-08
set default color to cart subtotal

akihitoalextsuboi:
Class .cart-subtotal & .cart-total color should be set as default color, not as direct color.

PR #2288, opened 2017-10-14
Convert Cart total/subtotal CSS colors to vars

gregdaynes:
This is for issue #2251 

Following up on the feedback for #1708, I've taken Martin's suggestion for a variable name, but opted for underscores to match the rest of the variable names. 

I've also opted to not use `$product_price_text_color` for the text - It would make it default to black, which is not inline with the current style, as well is not a fitting name for the context. Instead I've created a new var `$cart_total_text_color ` and assigned it white.

### DIFF
--- a/frontend/app/assets/stylesheets/spree/frontend/_variables.scss
+++ b/frontend/app/assets/stylesheets/spree/frontend/_variables.scss
@@ -15,6 +15,9 @@ $product_body_text_color:   #404042 !default;
 $product_price_text_color:  #252525 !default;
 $product_link_text_color:   #BBBBBB !default;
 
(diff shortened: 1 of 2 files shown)


kennyadsl:
@tvdeyen I generally agree with not using extra spaces for indentation (I'm also in favor of commas in last element of arrays and hashes to preserve git history). In this case I'd go with being consistent with the rest of the file, since I can't see an easy solution for the current state of the file: I don't think we'll change that file so often to bring it into an ideal indentation state and probably it will be completely changed in the future instead. 

Of course this is a non blocking comment and I'm perfectly fine with the current indentation. 👍

PR #2289, opened 2017-10-16
update no_objects_found partial to allow not show new resource link

### LINKED ISSUE #2150: Getting method "new_admin_stock_location_stock_movement_url" not found while trying to access stock movement for a newly created stock location
### Steps to reproduce
In the admin, access shipping then stock location
create a new stock location
then trying to access stock movement for the newly created stock location

### Expected behavior
Access the stock movement configuration

### Actual behavior
System fails with a NoMethodError
`undefined method `new_admin_stock_location_stock_movement_url' for #<ActionDispatch::Routing::RoutesProxy:0x007fe34123ac90>
`
### System configuration
**Solidus Version**: 2.4.0.alpha

**Extensions in use**: solidus_auth_devise

ccarruitero:
Fix #2150 

Like described in the issue mentioned above there is an error when we try to generate a `new_object_url` for stock movements, this because stock movements only allow [index route](https://github.com/solidusio/solidus/blob/09b0d748b9e3f94f05fb8768a25673f9941d10d6/backend/config/routes.rb#L170).

### DIFF
--- a/backend/app/views/spree/admin/shared/_no_objects_found.html.erb
+++ b/backend/app/views/spree/admin/shared/_no_objects_found.html.erb
@@ -1,4 +1,4 @@
 <%= t('spree.no_resource', resource: plural_resource_name(resource)) %>
-<% if can? :create, resource %>
+<% if can? :create, resource and local_assigns[:new_resource_url] %>
   <%= link_to t('spree.create_one'), new_resource_url %>
 <% end %>
--- a/backend/app/views/spree/admin/stock_movements/index.html.erb
+++ b/backend/app/views/spree/admin/stock_movements/index.html.erb
@@ -40,8 +40,7 @@
 <% else %>
   <div class="col-9 no-objects-found">
     <%= render 'spree/admin/shared/no_objects_found',
-                 resource: Spree::StockMovement,
-                 new_resource_url: new_object_url %>
+                 resource: Spree::StockMovement %>
   </div>
 <% end %>
 


backend/app/views/spree/admin/shared/_no_objects_found.html.erb
```diff
@@ -1,4 +1,4 @@
 <%= Spree.t(:no_resource, resource: plural_resource_name(resource)) %>
-<% if can? :create, resource %>
+<% if can? :create, resource and !new_resource_url.nil? %>
```
tvdeyen: We should use `local_assigns` here to make it optional.

```rb
if can?(:create, resource) && local_assigns[:new_resource_url]
```

backend/app/views/spree/admin/stock_movements/index.html.erb
```diff
@@ -41,7 +41,7 @@
   <div class="col-9 no-objects-found">
     <%= render 'spree/admin/shared/no_objects_found',
                  resource: Spree::StockMovement,
-                 new_resource_url: new_object_url %>
+                 new_resource_url: nil %>
```
tvdeyen: We can remove this local entirely if we use `local_assigns` in the partial above.

backend/app/views/spree/admin/shared/_no_objects_found.html.erb
ccarruitero: Yeah, is more clean like you propose :+1: 

I'll update the pulll request accordingly

tvdeyen:
Sorry that this took so long. Please feel free to ping us, if it takes too long for an approved PR to be merged.

PR #2291, opened 2017-10-16
Allow access to Spree::Core::Environment through Spree::Config

cbrunsdon:
This PR makes the Spree::Core::Environment (previously only accessible through Rails.application.config.spree) available via the Spree::Config.

This makes configuring Solidus easier in a few ways:
- We no longer need to understand or rely on rails initializers (which are very confusing, even for long-time rails users)
- Providing access via Spree::Config.environment is clearer than telling people to set some things on config, and others on rails environment (see doc changes to simple_coordinator)
- The environment works more like the other preferences, and can be reset through reset_spree_preferences (no need to keep the "old" configs around)

I've left the initializers in in case applications or gems are still using them, and they should be able to continue using them as before, but I think we should think about deprecating them in the future.

### DIFF
--- a/core/app/models/concerns/spree/calculated_adjustments.rb
+++ b/core/app/models/concerns/spree/calculated_adjustments.rb
@@ -22,7 +22,7 @@ module Spree
       end
 
       def spree_calculators
-        Rails.application.config.spree.calculators
+        Spree::Config.environment.calculators
       end
     end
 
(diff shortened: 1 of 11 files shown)


adaddeo:
> (see doc changes to simple_coordinator)

Could you link this? Definitely been using Rails/Solidus for a while still get confused by the initialization process.

PR #2292, opened 2017-10-16
Remove rspec-collection-matchers.

cbrunsdon:
The only functionality we were using from it was error_on, which can be
just as easily replaced with slightly different syntax.

### DIFF
--- a/core/spec/models/spree/price_spec.rb
+++ b/core/spec/models/spree/price_spec.rb
@@ -20,8 +20,9 @@ RSpec.describe Spree::Price, type: :model do
     context 'when the amount is less than 0' do
       let(:amount) { -1 }
 
-      it 'has 1 error_on' do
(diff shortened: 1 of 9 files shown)


cbrunsdon:
@tvdeyen `error_on` re-checks `.valid?`, which is first checked in the before block on the parent context, but not again after the object is modified in the nested before. pushed up a fix.

core/spec/models/spree/price_spec.rb
```diff
@@ -20,8 +20,9 @@
     context 'when the amount is less than 0' do
       let(:amount) { -1 }
 
-      it 'has 1 error_on' do
-        expect(subject.error_on(:amount).size).to eq(1)
+      it 'has 1 error on amount' do
+        subject.valid?
```
tvdeyen: Maybe `expect(subject).to_not be_valid`?

PR #2294, opened 2017-10-16
remove unused methods in stock_transfers_controller

ccarruitero:
`source_location` and `destination_location` methods in `stocks_transfers_controller` seems to be unused.

Judging for the parameters that use, I think was used for the old stock transfers views but not anymore.


jhawthorn:
We can probably just remove these IMO, since they're private methods on a backend controller it's unlikely that anyone would be relying on them.

ccarruitero:
I just update the pull request removing the methods

jhawthorn:
Thank you!

PR #2302, opened 2017-10-18
Remove assert usage in checkout_spec

cbrunsdon:
This isn't an rspec feature, it was provided through rspec/rails weird
bridge layer to test/unit, and its easy to re-write to rspec.


cbrunsdon:
The circle error was my fault, typo'd something, fixed it and its fine now.

PR #2304, opened 2017-10-18
Generate js_locale_data using JSON.dump

jhawthorn:
We generate some fairly large JSON here, and Rails's `to_json` is a fair bit slower than plain `JSON.generate` due to smart handling of ActiveRecord and other objects (which we are guaranteed not to have in translation data).

```
               I18n.t()  177.064k (± 4.5%) i/s
       I18n.t().to_json  155.028  (± 1.3%) i/s
JSON.generate(I18n.t())    1.045k (± 1.1%) i/s
```

So using `JSON.generate` is ~5x faster faster than using `to_json`.

It would be even faster to use [Oj](https://github.com/ohler55/oj) if available, but it's unfortunately not as easy as just using multi_json, which will call rails' to_json if Oj isn't available :disappointed:.

```
            I18n.t()               188.690k (± 1.3%) i/s
    I18n.t().to_json               165.058  (± 0.6%) i/s
JSON.generate(I18n.t())              1.149k (± 0.7%) i/s
   Oj.dump(I18n.t())                 4.417k (± 9.4%) i/s
MultiJson.dump(I18n.t()) w/ JSON   150.526  (± 1.3%) i/s
MultiJson.dump(I18n.t()) w/ Oj       4.253k (± 6.4%) i/s
```

### DIFF
--- a/backend/app/views/spree/admin/shared/_js_locale_data.html.erb
+++ b/backend/app/views/spree/admin/shared/_js_locale_data.html.erb
@@ -3,40 +3,42 @@
     var Spree = {}
   }
   Spree.translations = <%==
-    I18n.t("spree").merge({
-      date_picker:    Spree.t(:js_format,
-                               scope: 'date_picker',
-                               default: 'yy/mm/dd'),
-      abbr_day_names:         I18n.t(:abbr_day_names, scope: :date),
-      destroy:                Spree.t(:destroy, scope: :actions),
-      edit:                   Spree.t(:edit,    scope: :actions),
-      save:                   Spree.t(:save,    scope: :actions),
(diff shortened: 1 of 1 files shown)


jhawthorn:
> In the PR title and commit message you talk about JSON.generate while in the code you use JSON.dump (which uses JSON.generate internally). This can be confusing. Maybe update the commit message?

Updated.

> There is also JSON.fast_generate. Did you tried that as well?

I just tried `fast_generate` and saw no difference in performance

```
JSON.generate(I18n.t())
                          1.047k (± 1.0%) i/s -      5.304k in   5.065674s
JSON.fast_generate(I18n.t())
                          1.047k (± 1.1%) i/s -      5.304k in   5.064887s
```

PR #2309, opened 2017-10-19
Replace Spree.t with plain I18n.t

jhawthorn:
The `Spree.t` method provides debatable benefit over the standard `I18n.t` method and `t` helper.

* `Spree.t(:foo)` is one character shorter than `t('spree.foo')`
* `Spree.t('foo.bar')` is the same length as `t('spree.foo.bar')`
* `Spree.t` doesn't support the `t(".relative")` shorthand (so we currently use both `t` and `Spree.t`)
* `Spree.t` generates `<span class="translation_missing"` on errors (like the `t` helper), which is surprising because it looks more like `I18n.t` (which doesn't emit html)
* Use of `Spree.t` prevents us from using tools like [i18n-tasks](https://github.com/glebm/i18n-tasks)

This PR replaces `Spree.t` with `I18n.t` throughout the project.


cbrunsdon:
:+1:

PR #2311, opened 2017-10-21
Require only part of activemerchant

mamhoff:
ActiveMerchant is 2.5 megs of glue code that we don't need. This PR only requires those files of ActiveMerchant that we do need: Errors, Billing Response and Credit Card. 

It also bumps the required ActiveMerchant Version to 1.66, as that is the first version that works with Rails 5.1.

### DIFF
--- a/core/lib/spree/core.rb
+++ b/core/lib/spree/core.rb
@@ -1,5 +1,4 @@
 require 'rails/all'
-require 'active_merchant'
 require 'acts_as_list'
 require 'awesome_nested_set'
 require 'cancan'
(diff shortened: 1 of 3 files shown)


tvdeyen:
`factory_girl` was finally renamed to `factory_bot`. Pulling it out into a separate PR makes sense

tvdeyen:
@mamhoff please rebase now that the factory bot rename is merged

sample/spec/spec_helper.rb
```diff
@@ -12,6 +12,7 @@
 require 'rspec/rails'
 require 'ffaker'
 require 'spree_sample'
+require 'factory_girl'
```
jhawthorn: This will generate a warning and I don't think it is necessary after a rebase (the Gemfile will cause factory_bot to be required).

It turns out we also don't need factories at all for this test suite, so I have removed them in #2330

jhawthorn:
rebased and removed (now unnecessary) last commit

PR #2312, opened 2017-10-23
Fix module/class nesting in calculator/*

cbrunsdon:
Spree::Calculator is a class (that inherits from Spree::Base), but in
these definitions we don't clarify it as such.

### DIFF
--- a/core/app/models/spree/calculator/free_shipping.rb
+++ b/core/app/models/spree/calculator/free_shipping.rb
@@ -1,3 +1,5 @@
+require_dependency 'spree/calculator'
+
 module Spree
   # @deprecated This calculator will be removed in future versions of Spree.
   #   The only case where it was used was for Free Shipping Promotions.  There is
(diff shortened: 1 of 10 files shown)


core/app/models/spree/calculator/default_tax.rb
```diff
@@ -1,58 +1,60 @@
 require_dependency 'spree/calculator'
```
jhawthorn: Shouldn't this require make the change unnecessary?

cbrunsdon: Its not require_dependency'd everywhere its defined/used. 

I'd say we fix the definitions to be consistent and not have to worry about it.

cbrunsdon:
Okay, I don't personally agree with the require_dependency over keeping the class definitions the same, but I'll switch it. Are we okay with me leaving them broken out so its clear Calculator is a class and not a module?

PR #2316, opened 2017-10-23
Indent nested taxon menues and highlight the selected taxons.

bofrede:
To see this in action you need:
* Create multiple nested taxons in the backend.
* `  config.max_level_in_taxons_menu = 3` in `config/initializers/spree.rb`

### DIFF
--- a/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
+++ b/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
@@ -359,9 +359,14 @@ nav#taxonomies {
   }
 
   .taxons-list {
+    ul {
+      margin-left: 1em;
+    }
     li {
-      a {
-        font-size: $main_navigation_font_size
+      font-size: $main_navigation_font_size;
+      font-weight: normal;
+      &.current {
+        font-weight: bold;
       }
     }
   }


kennyadsl:
Hello! It would be great if you could explain better which problem this PR will solve, and maybe post some screenshot of the result so it is immediately clear to everyone. 

Thanks for your contribution!

bofrede:
## Without this fix, the taxon menu looks like this
* Cars
* Ferrari
* Lotus
* Mercedes
* Boats
* Regal
* Wellcraft

## With this fix, the taxon menu looks like this
* **Cars**
  * Ferrari
  * Lotus
  * **Mercedes**
* Boats
  * Regal
  * Wellcraft

frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
```diff
@@ -359,11 +359,21 @@ nav#taxonomies {
...
+    }
+    ul li {
+      font-weight: normal;
       a {
         font-size: $main_navigation_font_size
       }
     }
+    .current {
+      font-weight: bold;
+    }
```
kennyadsl: I think this code can be simplified in this way:

```scss
  ul.taxons-list {
    ul {
      margin-left: 1em;
    }

    li {
      font-size: $main_navigation_font_size;
      font-weight: normal;

      &.current {
        font-weight: bold;
      }
    }
  }
```

bofrede: I agree it can be simplified a little.
However `&.current` is a too specific selector.

kennyadsl: I think it works with `&.current` since it is applies to `ul.taxons-list li`, which is exactly elements that need this style.

tvdeyen: @kennyadsl I think @bofrede meant that the selector is too strong in terms of "Do not over specify css selectors" rule of thumb. It would be hard to override this if we over specify selectors. 90% of `!important` usage is because CSS authors uses too specific css selectors.

kennyadsl: @tvdeyen thanks for your feedback. Yes, it's probably better to have it less specific.

bofrede:
Now simplified and squashed as per @kennyadsl's request.

kennyadsl:
Thanks!

tvdeyen:
Sorry that this took so long to be merged. Please feel free to ping us, if an approved PR takes too long to be merged.

PR #2320, opened 2017-10-24
Use transactional fixtures in frontend and backend

jhawthorn:
Previously, we needed to run tests outside of a transaction because our test and server thread required separate database connections.

However as of Rails 5.1, due to the work implementing system tests, this is no longer necessary.

This should be significantly faster for postgresql and mysql, which otherwise had to truncate all tables after every feature spec.

This keeps `DatabaseCleaner` around to truncate the database before running specs. We might consider removing that as well, but is outside the scope of this PR.

See also
[rails/rails#28083](https://github.com/rails/rails/pull/28083)
[rails/rails#26703](https://github.com/rails/rails/pull/26703)

### DIFF
--- a/backend/spec/features/admin/promotion_adjustments_spec.rb
+++ b/backend/spec/features/admin/promotion_adjustments_spec.rb
@@ -192,8 +192,11 @@ describe "Promotion Adjustments", type: :feature, js: true do
       choose "Multiple promotion codes"
       fill_in "Base code", with: "testing"
       fill_in "Number of codes", with: "10"
-      click_button "Create"
-      expect(page).to have_title("SAVE SAVE SAVE - Promotions")
+
+      perform_enqueued_jobs {
+        click_button "Create"
+        expect(page).to have_title("SAVE SAVE SAVE - Promotions")
+      }
 
       promotion = Spree::Promotion.find_by(name: "SAVE SAVE SAVE")
       expect(promotion.path).to be_nil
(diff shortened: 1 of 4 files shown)


jhawthorn:
Unscientific test:

**Before:** [Finished in 9 minutes 46 seconds](https://circleci.com/gh/solidusio/solidus/6485#build-timing/containers/1)
**After:** [Finished in 7 minutes 42 seconds](https://circleci.com/gh/jhawthorn/solidus/1348#build-timing/containers/1)

jhawthorn:
I've been trying to diagnose the CI error. It's repeatable with the same seed, but weird. Some concurrency error when loading classes. I have to assume it's caused by this change.

jhawthorn:
Aha! We were having a deadlock related to us using the `:async` ActiveJob runner (which we really shouldn't do in tests). This should be fixed

backend/spec/features/admin/promotion_adjustments_spec.rb
```diff
@@ -192,8 +192,11 @@
       choose "Multiple promotion codes"
       fill_in "Base code", with: "testing"
       fill_in "Number of codes", with: "10"
-      click_button "Create"
-      expect(page).to have_title("SAVE SAVE SAVE - Promotions")
+
+      perform_enqueued_jobs {
```
jhawthorn: I think this will also fix a spurious error we've had in this spec. Bonus!

jhawthorn:
I started running into a different spurious error in `promotion_adjustments_spec` locally. I'm going to spend more time investigating that issue before merging.

jhawthorn:
> I started running into a different spurious error in promotion_adjustments_spec locally. I'm going to spend more time investigating that issue before merging.

Fixed by improving the spec. I'll merge once CI passes.

PR #625, opened 2015-12-26
Add concept of Image gallery

cbrunsdon:
This PR comes out of the conversation of #495 

We need a better way for stores to customize how different items relate to images, I believe we're missing the concept of a "gallery" that gives stores that control.

A single ActiveRecord association is too simple for stores to model the relationship between a variant (or product) and its images, this PR creates the concept of a "Gallery" for stores to implement and model that relationship as it pertains to their store.

To do this I needed to:
- [x] Shift existing calls onto this new api
- [x] mark old API as deprecated
- [x] create ProductAssociation for Spree::Product's
- [x] make sure preload_params are used everywhere to cut down on queries

I'm open to any and all feedback on how this is being executed, and especially around me documentation which I'm specifically trying to improve upon.

PR #2323, opened 2017-10-24
Add `#gallery` to `Spree::Variant` and `Spree::Product`

swcraig:
This is a lightweight version of #625.

Related to #2324, I think this is a good starting point to begin opening up an interface for different image upload implementations.

### DIFF
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -280,6 +280,19 @@ module Spree
       end
     end
 
+    # An enumerable of media that goes along with a product
+    # Stores may not necessarily want to attach images to products (variants)
+    # This provides a more robust interface
+    #
+    # @return [Enumerable] of media for a product (and its variants)
+    def gallery
+      if Product.method_defined?(:variant_images)
+        association(:variant_images).reader
+      else
+        []
+      end
+    end
+
     private
 
     def any_variants_not_track_inventory?
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -350,6 +350,15 @@ module Spree
       end.flatten.compact
     end
 
+    # An enumerable of media that goes along with a variant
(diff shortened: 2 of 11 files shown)


frontend/app/views/spree/products/_thumbnails.html.erb
```diff
@@ -1,19 +1,10 @@
 <%# no need for thumbnails unless there is more than one image %>
```
jhawthorn: We can't change the behaviour of this file. We can't have it so users upgrade and the logic for displaying images on the frontend changes (I think it's likely in a ton of stores that this wouldn't be caught before production).

core/app/models/spree/variant.rb
```diff
@@ -350,6 +350,15 @@ def variant_properties
       end.flatten.compact
     end
 
+    # An enumerable of media that goes along with a variant
+    # Stores may not necessarily want to attach images to variants
+    # This provides a more robust interface
+    #
+    # @return [Enumerable] of media for a variant
+    def gallery
+      Variant.reflect_on_association(:images) ? association(:images).reader : []
```
jhawthorn: We define this association, so we shouldn't test for it. If this is to prepare for a future paperclip removal PR, it should be done later.

core/app/models/spree/product.rb
```diff
@@ -280,6 +280,19 @@ def find_variant_property_rule(option_value_ids)
       end
     end
 
+    # An enumerable of media that goes along with a product
+    # Stores may not necessarily want to attach images to products (variants)
+    # This provides a more robust interface
+    #
+    # @return [Enumerable] of media for a product (and its variants)
+    def gallery
+      if Product.method_defined?(:variant_images)
```
jhawthorn: We define this association, so we shouldn't test for it. If this is to prepare for a future paperclip removal PR, it should be done later.

swcraig:
Closed in favour of #2337

PR #2327, opened 2017-10-25
Move checkout coupon code section into summary

kennyadsl:
This change is needed if we want to stop passing from checkout controller to apply coupon codes.

At the moment we have several issues with this behavior:

- apply_coupon_code is defined into frontend store_controller. This potentially allows all controllers' actions to apply coupon code, just passing `[:order][:coupon_code]` along with other params. We don't want this anymore since those actions are designed to work for orders and checkout controllers only. See [`render :edit`](https://github.com/solidusio/solidus/blob/88f531a2f0568967c28de2ccf591ba5b066fbc2f/frontend/app/controllers/spree/store_controller.rb#L17-L33) for example, it could not work in all scenarios. To fix this issue I duplicated this method in both orders and checkout controllers, I think in this case duplication is not bad, since real shared logic is handled by `PromotionHandler::Coupon.new(@order).apply` and we'll need a different behavior in those controllers;
- #2155 : Checkout controller needs extra ivars setup before rendering, after coupon code is applied with errors. This is now fixed since we can add this setup into the checkout-specific code that applies coupon code.
- Coupon Code application UX is "strange". We have the coupon code field and submit button into the payment form. If we add an invalid code and submit the payment form with payment data, we'll lost all payment data we just added. If we add a valid coupon code and submit with ajaxy Apply Button behavior page will reload and we'll lost all payment data we just added. I think we need to make clear that those forms are separated so user would not fill  both of them uselessy. To improve UX I moved the coupon code form into the summary box. Also, I made the coupon code a real form, even if it is always submitted with JS, both for semantic and to have a real working  fallback if for some reason JS is not working.

In my plans next steps will be:

- Deprecate accepting `[:order][:coupon_code]` param in checkout controller. If this PR is merged we'll just have this behavior called as a fallback if for some reason JS is not working.
- Investigate if we can/makes sense to move the coupon code into a separate form also for the cart. I feel like it's not that terrible to have it in a single form there in terms on UX.
- If we want to deprecate, we'll need another fallback endpoint, so I was thinking about creating a separate controller that will only handle coupon code application, maybe also for the api.

### DIFF
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -173,12 +173,13 @@ module Spree
         handler = PromotionHandler::Coupon.new(@order).apply
 
         if handler.error.present?
-          setup_for_current_state
           flash.now[:error] = handler.error
-          respond_with(@order) { |format| format.html { render :edit } } && return
         elsif handler.success
           flash[:success] = handler.success
         end
+
+        setup_for_current_state
+        respond_with(@order) { |format| format.html { render :edit } } && return
       end
     end
 
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -173,6 +173,7 @@ module Spree
         handler = PromotionHandler::Coupon.new(@order).apply
 
         if handler.error.present?
+          setup_for_current_state
           flash.now[:error] = handler.error
           respond_with(@order) { |format| format.html { render :edit } } && return
         elsif handler.success
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -166,6 +166,21 @@ module Spree
       spree.order_path(@order)
     end
 
+    def apply_coupon_code
+      if params[:order] && params[:order][:coupon_code].present?
+        @order.coupon_code = params[:order][:coupon_code]
+
+        handler = PromotionHandler::Coupon.new(@order).apply
+
+        if handler.error.present?
+          flash.now[:error] = handler.error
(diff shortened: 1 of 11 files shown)


mtomov:
That's all great! And all in for all the 3 _next steps_ 

A current problem I'm always ending up solving with the coupon code not being a separate form is to detect when the enter key has been pressed, and submit just it .. as otherwise an Enter key would submit the payment form.

```javascript
  $('#coupon_code').keypress (e) ->
    if(e.which == 13)
      Spree.onCouponCodeApply()
      e.preventDefault()
```

Separate controller and action is definitely the way to go!

Thanks again

frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -165,6 +165,23 @@ def completion_route
       spree.order_path(@order)
     end
 
+    def apply_coupon_code
+      if params[:order] && params[:order][:coupon_code].present?
+        @order.coupon_code = params[:order][:coupon_code]
```
tvdeyen: We should use `params.require(:order).permit(:coupon_code)` instead. This is more secure and will most likely fail  in Rails >= 5.1 otherwise.

frontend/app/controllers/spree/orders_controller.rb
```diff
@@ -118,5 +118,20 @@ def assign_order
         redirect_to(root_path) && return
       end
     end
+
+    def apply_coupon_code
+      if params[:order] && params[:order][:coupon_code].present?
+        @order.coupon_code = params[:order][:coupon_code]
```
tvdeyen: Ditto

kennyadsl:
@tvdeyen sure, I'll fix and rebase after we merge #2690 so we can test it on the CI with both versions.

kennyadsl:
@tvdeyen changes requested done!

PR #2328, opened 2017-10-25
Move Paperclip into an extension

swcraig:
Removes all Paperclip specific code (including all JavaScript related to uploading images in the admin) from Solidus. All of this functionality has been moved into a branch in (the currently temporary repo) [solidus_paperclip](https://github.com/swcraig/solidus_paperclip/commits/pull-paperclip-into-extension).

Views in the admin/frontend render empty space for images. This way, future image uploading extensions only need to attach images to variants and they will show up in the expected places. Image upload implementations shouldn't need to know about specific views in Solidus.

This PR also adds `#gallery` methods to `Spree::Variant` and `Spree::Product`. These methods return an empty `Enumerable`, which future image upload extensions would be expected to override. I can open a separate PR for this change, but it is cleaner to have `#gallery` before this is merged. Most notably for JBuilder endpoints in the API and also the Handlebars templates in the admin. By using the null object pattern here I can leave those files without modification for now.

`Spree::Taxon#icon` has been removed, and a decorator has been added on the extension side. If this needs a better interface, I will work on addressing that.

This is a relatively large PR, and would be difficult to review easily. If there is an obvious way to split this up logically, please let me know.

I am more than happy to discuss any aspect of this change, and will do everything I can to move this forward in the best possible way.

**Before this is merged** solidus_paperclip needs to exist on solidusio-contrib and that location needs to be referenced in `lib/sandbox.sh`. At the moment it's a hardcoded relative path on my machine.


swcraig:
I think the best approach will be to merge #2337 before pulling out Paperclip.

dankmitchell:
Can this move forward now?

ericsaupe:
@swcraig there are a number of conflicting files but this is a great direction!

ericsaupe:
Closing in favor of #2974 now that @elia has picked it up and @swcraig has mentioned he doesn't have bandwidth right now.

Thanks for the work on getting the ball rolling @swcraig!

Issue #2141, opened 2017-08-09
Solidus 2.3 doesn't finalize shipments after an order is completed


CallMeSH:
### Steps to reproduce
- Create a new order
- Complete it
- In the backend, add a product to the cart
- Go to the `Cancel Items` tab, you can't find your newly added product

### Expected behavior
When you add a product to the cart after the checkout, we already know that the product is available in stock. This is why its inventory units should have their `pending` value set to false.

### Actual behavior
All products created after the order's completion get their `pending` value to true as if the order was not completed yet.

Calling `finalize!` on the shipment solves the issue but it requires to open a rails console. Either this is a bug and fixing it will solve the issue, or there should be a button to `finalize` a shipment in the backend.

### System configuration
**Solidus Version**: 2.3.0
**Extensions in use**: Various home-made extensions but nothing covering this scope of the app

Thanks!

PR #2331, opened 2017-10-27
Finalize shipment after inventory units are added to completed order

DanielePalombo:
This PR is going to fix the issue https://github.com/solidusio/solidus/issues/2141 .

Create a new class Spree::Stock::InventoryUnitsFinalizer
to finalize the inventory units provided.
It unstocks the desired quantity and marks as not pending the inventory units processed.
To fix the issue we run it, both when a shipment is finalized and when a variant is added to a shipment in a completed order.

### DIFF
--- a/core/spec/models/spree/inventory_unit_spec.rb
+++ b/core/spec/models/spree/inventory_unit_spec.rb
@@ -142,6 +142,7 @@ RSpec.describe Spree::InventoryUnit, type: :model do
     }
 
     it "should create a stock movement" do
+      expect(Spree::Deprecation).to receive(:warn)
       Spree::InventoryUnit.finalize_units!(inventory_units)
       expect(inventory_units.any?(&:pending)).to be false
     end
--- a/core/spec/models/spree/order_inventory_spec.rb
+++ b/core/spec/models/spree/order_inventory_spec.rb
@@ -35,12 +35,19 @@ RSpec.describe Spree::OrderInventory, type: :model do
     end
 
     context "order is not completed" do
-      before { order.update_columns completed_at: nil }
+      let(:inventory_unit_finalizer) { double(:inventory_unit_finalizer, run!: [true]) }
 
-      it "doesn't unstock items" do
-        expect {
-          subject.verify(shipment)
-        }.not_to change { stock_item.reload.count_on_hand }
+      before do
+        allow(Spree::Stock::InventoryUnitsFinalizer)
+          .to receive(:new).and_return(inventory_unit_finalizer)
+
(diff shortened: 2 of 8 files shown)


core/spec/models/spree/inventory_unit_spec.rb
```diff
@@ -142,6 +142,7 @@
     }
 
     it "should create a stock movement" do
+      expect(Spree::Deprecation).to receive(:warn)
       Spree::InventoryUnit.finalize_units!(inventory_units)
```
kennyadsl: why not using the new call here and remove the Deprecation `:warn` stub?

core/spec/models/spree/order_inventory_spec.rb
```diff
@@ -35,9 +35,18 @@
...
+
+      before do
+        allow(Spree::Stock::InventoryUnitsFinalizer)
+          .to receive(:new).and_return(inventory_unit_finalizer)
+
+        order.update_columns completed_at: nil
+      end
 
       it "doesn't unstock items" do
+        expect(inventory_unit_finalizer).to_not receive(:run!)
```
kennyadsl: I have the feeling we can avoid this expectation: if `run!` is called the next expectation will fail, right? Is there any other reason you wanted to set this expectation?

core/spec/models/spree/inventory_unit_spec.rb
DanielePalombo: This is a test for `Spree::InventoryUnit#finalize_units!`.

core/spec/models/spree/stock/inventory_units_finalizer_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      subject { described_class.new([inventory_unit]).run! }
+
+      it "updates the associated inventory units" do
+        inventory_unit.update_columns(updated_at: 1.hour.ago)
+        expect { subject }.to change { inventory_unit.reload.updated_at }
+      end
+
+      it "updates the inventory units to not be pending" do
+        expect { subject }.to change { inventory_unit.reload.updated_at }
```
gmacdougall: This test doesn't match its description.

DanielePalombo: Fixed!

kennyadsl:
@gmacdougall can you please review again now?

kennyadsl:
@DanielePalombo since it's so old, can you please rebase against latest master? Thanks!

DanielePalombo:
@kennyadsl I have rebased it!

PR #2332, opened 2017-10-27
Explicitly define backend/promotions routes

vladstoick:
Currently the `show` route is also include but it's never defined which leads to a
`AbstractController::ActionNotFound` being thrown in case someone reaches
that page

### DIFF
--- a/backend/app/controllers/spree/admin/promotions_controller.rb
+++ b/backend/app/controllers/spree/admin/promotions_controller.rb
@@ -5,6 +5,10 @@ module Spree
 
       helper 'spree/promotion_rules'
 
+      def show
+        redirect_to action: :edit
+      end
+
       def create
         @promotion = Spree::Promotion.new(permitted_resource_params)
         @promotion.codes.new(value: params[:single_code]) if params[:single_code].present?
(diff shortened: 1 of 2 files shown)


backend/config/routes.rb
```diff
@@ -9,7 +9,7 @@
       end
     end
 
-    resources :promotions do
+    resources :promotions, only: [:new, :edit, :update, :create, :destroy,] do
```
kennyadsl: I'd write `except: :show` instead?

vladstoick: Changed it 👍

kennyadsl:
Thinking twice, it's probably better to redirect to the `:edit` route, as we did in [backend product controller](https://github.com/solidusio/solidus/blob/3fe329d0b3862ccc9f0145e494bfd4e2b2af7a08/backend/app/controllers/spree/admin/products_controller.rb#L10-L12).

mtomov:
> Thinking twice, it's probably better to redirect to the :edit route, as we did in backend products controller.

this has been the best redirect ever, as I can relatively easy add `admin` to the frontend products page url to get to the admin one. Not as easy the opposite way around.

In the promotions, I don't think it matters much.

kennyadsl:
The thing I was thinking at is that by only removing the `:show` route we still have a 404 error (instead of the current `AbstractController::ActionNotFound`), which does not solve the case someone reaches that page. With the redirect in controller instead we wont have any error.

kennyadsl:
@vladstoick thanks!

PR #2337, opened 2017-10-30
Add `#gallery` to `Variant` and `Product`

swcraig:
[Paperclip is deprecated](https://robots.thoughtbot.com/closing-the-trombone) and we need to remove it from stock Solidus.

In order to remove Paperclip properly we need to add extension points so that we can plug a different image uploading library (like Active Storage). Adding galleries to `Product` and `Variant` will give us more control over what we send to the view layer, and protect our core models from image implementation details.

To plug this interface a developer needs to create custom galleries for both `Variant` and `Product` and make sure that the `#images` method of those return a collection of images that respond to the new "presentational" image interface the frontend expects. This PR includes changes to `Spree::Image` which transitions it into a presentation role for Paperclip. Specifically, the presentational image interface that the view layer expects is:

```
id: int
url(:size): string
filename: string
viewable_id: int # the ID of the variant this is "attached" to
```

Before Paperclip can be extracted into its own extension properly we need this PR to be merged.

Much of the work around the the `Gallery` classes are heavily influenced by/verbatim the work Clarke did in #625.

### DIFF
--- a/backend/app/views/spree/admin/orders/_shipment_manifest.html.erb
+++ b/backend/app/views/spree/admin/orders/_shipment_manifest.html.erb
@@ -4,7 +4,9 @@
       data-variant-id="<%= item.variant.id %>"
       >
     <td class="item-image align-center">
-      <%= render 'spree/admin/shared/image', image: item.variant.display_image, size: :mini %>
+      <%= render 'spree/admin/shared/image',
+        image: (item.variant.gallery.images.first || item.variant.product.gallery.images.first),
+        size: :mini %>
     </td>
     <td class="item-name">
       <%= link_to item.variant.product.name, edit_admin_product_path(item.variant.product) %><br><%= "(" + variant_options(item.variant) + ")" unless item.variant.option_values.empty? %>
--- a/backend/app/views/spree/admin/shared/_image.html.erb
+++ b/backend/app/views/spree/admin/shared/_image.html.erb
@@ -1,7 +1,7 @@
 <% size ||= :mini %>
 
-<% if image && image.attachment? %>
-  <%= image_tag image.attachment(size) %>
+<% if image_url = image.try(:url, size) %>
+  <%= image_tag image_url %>
 <% else %>
   <span class="image-placeholder <%= size %>"></span>
 <% end %>
--- a/backend/app/views/spree/admin/stock_items/_stock_management.html.erb
+++ b/backend/app/views/spree/admin/stock_items/_stock_management.html.erb
@@ -28,7 +28,9 @@
         <td class="no-padding" rowspan="<%= row_count %>">
           <div class='variant-container'>
             <div class='variant-image'>
-              <%= render 'spree/admin/shared/image', image: variant.display_image(fallback: false), size: :small %>
+              <%= render 'spree/admin/shared/image',
+                image: variant.gallery.images.first,
+                size: :small %>
             </div>
             <div class='variant-details'>
               <table class='stock-variant-field-table'>
--- a/core/app/models/spree/gallery/product_gallery.rb
+++ b/core/app/models/spree/gallery/product_gallery.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Spree
+  module Gallery
+    class ProductGallery
+      def initialize(product)
+        @product = product
+      end
+
+      # A list of all images associated with this gallery
+      #
+      # @return [Enumerable<Spree::Image>] all images in the gallery
+      def images
+        @images ||= @product.variant_images
+      end
+    end
+  end
+end
--- a/core/app/models/spree/gallery/variant_gallery.rb
+++ b/core/app/models/spree/gallery/variant_gallery.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Spree
+  module Gallery
+    class VariantGallery
+      def initialize(variant)
+        @variant = variant
+      end
+
+      # A list of all images associated with this gallery
+      #
+      # @return [Enumerable<Spree::Image>] all images in the gallery
+      def images
+        @images ||= @variant.images
+      end
+    end
+  end
+end
--- a/core/app/models/spree/image.rb
+++ b/core/app/models/spree/image.rb
@@ -19,8 +19,10 @@ module Spree
     # we need to look at the write-queue for images which have not been saved yet
     after_post_process :find_dimensions, if: :valid?
 
-    # used by admin products autocomplete
     def mini_url
+      Spree::Deprecation.warn(
+        'Spree::Image#mini_url is DEPRECATED. Use Spree::Image#url(:mini) instead.'
+      )
       attachment.url(:mini, false)
     end
 
--- a/core/app/models/spree/image.rb
+++ b/core/app/models/spree/image.rb
@@ -24,6 +24,14 @@ module Spree
       attachment.url(:mini, false)
     end
 
+    def url(size)
+      attachment.url(size)
+    end
+
+    def filename
+      attachment_file_name
+    end
+
     def find_dimensions
       temporary = attachment.queued_for_write[:original]
       filename = temporary.path unless temporary.nil?
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -381,6 +381,7 @@ module Spree
     #   not from this variant
     # @return [Spree::Image] the image to display
     def display_image(fallback: true)
+      Spree::Deprecation.warn('Spree::Variant#display_image is DEPRECATED. Choose an image from Spree::Variant#gallery instead')
       images.first || (fallback && product.variant_images.first) || Spree::Image.new
     end
 
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -394,6 +394,14 @@ module Spree
       end.flatten.compact
     end
 
+    # The gallery for the variant, which represents all the images
+    # associated with it
+    #
+    # @return [Spree::Gallery] the media for a variant
+    def gallery
+      @gallery ||= Spree::Config.variant_gallery_class.new(self)
+    end
+
     private
 
     def rebuild_vat_prices?
--- a/frontend/app/views/spree/products/_thumbnails.html.erb
+++ b/frontend/app/views/spree/products/_thumbnails.html.erb
@@ -1,17 +1,19 @@
 <%# no need for thumbnails unless there is more than one image %>
-<% if (@product.images + @product.variant_images).uniq.size > 1 %>
+<% if @product.gallery.images.size > 1 %>
   <ul id="product-thumbnails" class="thumbnails inline" data-hook>
-    <% @product.images.each do |i| %>
-      <li class='tmb-all tmb-<%= i.viewable.id %>'>
-        <%= link_to(image_tag(i.attachment.url(:mini)), i.attachment.url(:product)) %>
+
+    <% @product.gallery.images.each do |image| %>
+      <% next if image.viewable_id != @product.master.id %>
+      <li class='tmb-all tmb-<%= image.viewable_id %>'>
+        <%= link_to(image_tag(image.url(:mini)), image.url(:product)) %>
       </li>
     <% end %>
 
     <% if @product.has_variants? %>
-      <% @product.variant_images.each do |i| %>
-        <% next if @product.images.include?(i) %>
-        <li class='vtmb tmb-<%= i.viewable.id %>'>
-          <%= link_to(image_tag(i.attachment.url(:mini)), i.attachment.url(:product)) %>
+      <% @product.gallery.images.each do |image| %>
+        <% next if image.viewable_id == @product.master.id %>
+        <li class='vtmb tmb-<%= image.viewable_id %>'>
+          <%= link_to(image_tag(image.url(:mini)), image.url(:product)) %>
         </li>
       <% end %>
     <% end %>
--- a/frontend/app/views/spree/shared/_image.html.erb
+++ b/frontend/app/views/spree/shared/_image.html.erb
@@ -1,11 +1,11 @@
 <% size ||= :mini %>
 <% itemprop ||= nil %>
 
-<% if image && image.attachment? %>
-  <% if itemprop %>
-    <%= image_tag image.attachment(size), itemprop: itemprop %>
+<% if image_url = image.try(:url, size)  %>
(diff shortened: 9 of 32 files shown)


core/app/models/spree/variant.rb
```diff
@@ -350,6 +350,16 @@ def variant_properties
       end.flatten.compact
     end
 
+    # An Enumerable of media that goes along with a variant
+    # Stores may not necessarily want to attach images to variants
+    # This provides a more robust interface
+    # NOTE: In the future this will return an empty array
```
jhawthorn: Why would this be useful? What is a developer supposed to do with this information.

I assume this is for extension points, since we want to avoid `class_eval`ing, maybe we should look at introducing a `Spree::Config.variant_gallery_class` and a `Spree::Config.product_gallery_class`.

tvdeyen: I like that idea of configurable image attachment classes. If we want extensions to be able to hook into we want to provide a real integration point. 

Also I don’t think `gallery` is the correct term. `gallery_images` would follow the “plural is a collection” convention of Rails better. We could even override the `images` getter. Btw. We need to deprecate the `images =` setter as well or delegate this to the images provider class.

core/app/models/spree/gallery/variant_gallery.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      # (see Spree::Gallery::Base#primary_image)
+      def primary_image
+        @variant.images.first || product_images.first
+      end
+
+      private
+
+      def product_images
+        @variant.product.try(:variant_images) || []
```
jhawthorn: I don't think this needs to be a `try`. Is `@variant` ever `nil`?

core/app/models/spree/gallery/base.rb
```diff
@@ -0,0 +1,23 @@
+module Spree
```
jhawthorn: I'm not sure there's a need for this base. It should be sufficient to duck-type have the two gallery classes "quack" similarly :duck: :duck: :duck: :duck: :duck:

swcraig: :duck: 

I guess the reason I would want the base class is so that it's obvious the interface that should be implemented for galleries in general. But I understand your point and can make the changes.

swcraig:
This is ready for another review. There have been a few changes since the last reviews. I have added `Variant#gallery.best_image`. This method will fallback to using the first associated product image if a variant itself has no images.

core/app/models/spree/gallery/variant_gallery.rb
```diff
@@ -0,0 +1,38 @@
...
+      #
+      # @return [Spree::Image] best image for the variant gallery
+      def best_image
+        @variant.images.first || product_images.first || Spree::Image.new
+      end
+
+      private
+
+      def product_images
+        @variant.product.variant_images
```
tvdeyen: I don’t see a need for this private method as long as we don’t access it from two different methods at least.

core/app/models/spree/gallery/variant_gallery.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      # The "best" {Spree::Image} associated with the gallery
+      # This will return the first associated product image if there are no
+      # variant images
+      #
+      # @return [Spree::Image] best image for the variant gallery
+      def best_image
+        @variant.images.first || product_images.first || Spree::Image.new
+      end
```
tvdeyen: Why not remove these image collector methods from here and move into the views. This might be more duplication, but is also more flexible. Although I think the naming is difficult and misleading. Let’s not make these decisions here, but in the views. They are easier to override in a per use case basis.

swcraig:
I think this is good to go. I've also added `ImagesHelper#image_or_default` for the admin/frontend views.

swcraig:
Thank you for the feedback.
 
The galleries have an `#image_urls` method in addition to an the `#images` method. The JBuilder API endpoints. the product thumbnails partial on the frontend, and `images/_image_row`  in the backend expect `Spree::Image`s (that respond to Paperclip API calls). I can't see an easy way to rewrite these with only URLs and keep the same behaviour.

The actual display logic no longer needs to know Paperclip's API (there is no more accessing of the `#attachment` of an image). Another sort of unfortunate result is that the views need to specify the image size twice (in the case that there is no image and we need to fallback to the CSS of `class= placeholder <size>`. [Like here for example](https://github.com/swcraig/solidus/blob/20d4b5cad2f5a7eedc9285c69c9639263f2b6226/backend/app/views/spree/admin/cancellations/index.html.erb#L28-L30).

I think this is closer to getting rid of the Paperclip specific logic out of the views, and I'm interested in hearing what we think of where this is going.

backend/app/views/spree/admin/shared/_image.html.erb
```diff
@@ -1,7 +1,7 @@
 <% size ||= :mini %>
 
-<% if image && image.attachment? %>
-  <%= image_tag image.attachment(size) %>
+<% if image %>
+  <%= image_tag image %>
```
mamhoff: This is also missing the `:mini` option.

core/app/models/spree/gallery/product_gallery.rb
```diff
@@ -0,0 +1,16 @@
...
+    class ProductGallery
+      def initialize(product)
+        @product = product
+      end
+
+      # A list of all images associated with this gallery
+      #
+      # @return [Enumerable<Spree::Image>] all images in the gallery
+      def images
+        @images ||= (@product.images + @product.variant_images).uniq
```
tvdeyen: There is no need to merge these two collections, because `@product.variant_images` already [includes images of the master variant](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/product.rb#L75) - what are [the images returned by `@product.images`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/product.rb#L72)

backend/app/views/spree/admin/stock_items/_stock_management.html.erb
```diff
@@ -28,7 +28,9 @@
         <td class="no-padding" rowspan="<%= row_count %>">
           <div class='variant-container'>
             <div class='variant-image'>
-              <%= render 'spree/admin/shared/image', image: variant.display_image(fallback: false), size: :small %>
+              <%= render 'spree/admin/shared/image',
+                image: variant.gallery.images.first.try(:url, :small),
```
tvdeyen: The thing you're passing into the partial as `image` is actually an `url`. Shouldn't we rename the variable as well?

backend/app/views/spree/admin/orders/_shipment_manifest.html.erb
```diff
@@ -4,7 +4,9 @@
       data-variant-id="<%= item.variant.id %>"
       >
     <td class="item-image align-center">
-      <%= render 'spree/admin/shared/image', image: item.variant.display_image, size: :mini %>
+      <%= render 'spree/admin/shared/image',
+        image: (item.variant.gallery.images.first || item.variant.product.gallery.images.first).try(:url, :mini),
```
tvdeyen: It makes sense to have a `first_image_url` method in the gallery classes.

```rb
item.variant.gallery.first_image_url(size: :mini) || item.variant.product.gallery.first_image_url(size: :mini)
```

core/app/models/spree/gallery/product_gallery.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Gallery
+    class ProductGallery
+      def initialize(product)
+        @product = product
+      end
+
+      # A list of all images associated with this gallery
+      #
+      # @return [Enumerable<Spree::Image>] all images in the gallery
+      def images
```
tvdeyen: I think we would gain much more isolation from  this class if it would have an `image_urls` method that takes a `size:` parameter.

core/app/models/spree/gallery/variant_gallery.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Gallery
+    class VariantGallery
+      def initialize(variant)
+        @variant = variant
+      end
+
+      # A list of all images associated with this gallery
+      #
+      # @return [Enumerable<Spree::Image>] all images in the gallery
+      def images
```
tvdeyen: Here as well, a `image_urls` method that takes a `size:` parameter would help us decoupling from paperclip

core/app/models/spree/image.rb
```diff
@@ -19,9 +19,18 @@ class Image < Asset
...
+      Spree::Deprecation.warn('Spree::Image#mini_url is DEPRECATED.')
       attachment.url(:mini, false)
     end
 
+    def filename
+      attachment_file_name
+    end
+
+    def url(size)
+      attachment.url(size)
```
tvdeyen: We are not able to decouple from paperclip if we have this method inside of here. It would be better to put this url retrieving logic inside the gallery classes. Extensions like `solidus_paperclip` would then implement such gallery classes.

frontend/app/views/spree/products/_thumbnails.html.erb
```diff
@@ -1,17 +1,19 @@
 <%# no need for thumbnails unless there is more than one image %>
-<% if (@product.images + @product.variant_images).uniq.size > 1 %>
+<% if @product.gallery.images.size > 1 %>
   <ul id="product-thumbnails" class="thumbnails inline" data-hook>
-    <% @product.images.each do |i| %>
-      <li class='tmb-all tmb-<%= i.viewable.id %>'>
-        <%= link_to(image_tag(i.attachment.url(:mini)), i.attachment.url(:product)) %>
+
+    <% @product.gallery.images.each do |image| %>
+      <% next if image.viewable_id != @product.master.id %>
```
swcraig: I don't see how I am going to be able to write this `_thumbnails` partial using only image URLs. At some point an image is going to need the context of which variant it is for. What do you think, @tvdeyen?

If we have situations like this (and the `admin/_image_row` partial), I think it would make sense for the galleries to return an "Image" class. If a developer wants to drop in a different implementation they can provide their own image class that responds to `#url`, `#id`, `#viewable_id`, and `#alt`. I was thinking something like this:

```
class MyImageClass
  def initialize(image)
    @image = image
  end

  def url(size)
    @image.the_way_this_image_library_gets_urls(size)
  end

  # etc...
end

class MyVariantGallery
  def images
    however_the_images_are_stored.map { |image| MyImage.new(image) }
  end
end
```

backend/app/views/spree/admin/shared/_image.html.erb
swcraig: This partial is given an image url (the caller has already decided the size and retrieved the relevant url). I've updated the naming to be clearer.

Does that make sense?

frontend/app/views/spree/shared/_image.html.erb
```diff
@@ -1,11 +1,11 @@
 <% size ||= :mini %>
 <% itemprop ||= nil %>
 
-<% if image && image.attachment? %>
+<% if image_url %>
   <% if itemprop %>
-    <%= image_tag image.attachment(size), itemprop: itemprop %>
+    <%= image_tag image_url, itemprop: itemprop %>
```
swcraig: Something to note here. We don't use an image's `alt` attribute in either frontend or backend. Would being able to access `image.alt` here be an argument for using a custom image class [like I suggest in this PR comment](https://github.com/solidusio/solidus/pull/2337/files#r154390915)? That image class could have a strict interface `#url(size)`, `#id` (for that single-use modal on `backend/image/_image_row`), `#viewable_id`, and `#alt`.

Is there a reason why we haven't historically been rendering an image's `alt` attribute? @mamhoff @tvdeyen

frontend/app/views/spree/products/_thumbnails.html.erb
tvdeyen: I like the idea of a separate `Image` class 👌

So we would have a repository (`Gallery`) and a presenter (maybe `Picture`?). This sounds like a good plan to me.

frontend/app/views/spree/shared/_image.html.erb
tvdeyen: > Would being able to access image.alt here be an argument for using a custom image class

Yes! A separate image class (`Picture` maybe?) makes total sense to abstract the presenter API 👍

tvdeyen: > Is there a reason why we haven't historically been rendering an image's alt attribute?

Not sure. We should render it (at least in the frontend).

swcraig:
I have this to the point where we don't reference Paperclip specific calls anywhere in the views.  We do continue to have Paperclip specific logic in the [solidus_api controller for images](https://github.com/swcraig/solidus/blob/147c57aac515f072fd09734dd4954e8688fdb518/api/app/controllers/spree/api/images_controller.rb), and in the [actual form for uploading images](https://github.com/swcraig/solidus/blob/147c57aac515f072fd09734dd4954e8688fdb518/backend/app/views/spree/admin/images/_form.html.erb). We can probably deal with these when someone goes to pull Paperclip into an extension.

frontend/app/views/spree/products/_thumbnails.html.erb
swcraig: After thinking about this, I think that we can transition `Spree::Image` to function as this presenter. I don't think there is much value in adding another class (`Spree::Picture`), when we can use `Spree::Image` to do the same thing. I've updated the PR to reflect this.

swcraig:
This is ready for another review. I plan on [transitioning `Spree::Image` to function as a "presenter" class for images on the frontend](https://github.com/solidusio/solidus/pull/2337#discussion_r156772037). It's sort of weird right now because it represents a Paperclip image, but instead of getting rid of it and using a new model (`Spree::Picture`) if/when someone rips out Paperclip we could just keep it and make it define the interface that views understand.

swcraig:
Could we possibly take another look at this? With [Rails 5.2/ActiveStorage close to release](http://weblog.rubyonrails.org/2018/1/30/Rails-5-2-RC1-Active-Storage-Redis-Cache-Store-HTTP2-Early-Hints-Credentials/) I think there should be some consideration to get Paperclip out.

swcraig:
[With Paperclip officially being deprecated](https://robots.thoughtbot.com/closing-the-trombone) can we take a serious look at merging this? When this is merged it will make extracting Paperclip into an extension much easier. I also think this is a better general approach to dealing with images in core/the view layer.

@tvdeyen feedback would be appreciated.

tvdeyen:
🎉 🎁

PR #2343, opened 2017-11-01
Revert "Use transactional fixtures in frontend and backend"

jhawthorn:
Reverts solidusio/solidus#2320

We've had a few [CI failures due to deadlocks](https://circleci.com/gh/solidusio/solidus/6555), which is probably related :disappointed_relieved: 

I'd like to revert this while investigating further.


tvdeyen:
:+1:

PR #2350, opened 2017-11-03
Use paranoia_ prefixed methods

jhawthorn:


backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -85,7 +85,15 @@ def update_positions
...
-    if @object.destroy
+
+    destroy_result =
+      if @object.respond_to?(:paranoia_destroy)
+        @object.paranoia_destroy
+      else
+        @object.destroy
+      end
+
+    if destroy_result
```
tvdeyen: 👍 good catch. Lots of extensions might inherit from this.

jhawthorn:
> I like the explicitness of this, although I think the name of your methods should be paranoid_*. Makes it more readable. Will comment on the paranoia PR as well.

I agree it sounds better and I definitely thought about it when changing adding these aliases, but the gem's name is paranoia and all the other methods in the gem are prefixed `paranoia_` (`paranoia_destroyed?`, `paranoia_scope`, `paranoia_column`) so I think I will keep with that naming.

PR #2352, opened 2017-11-03
Use transactional fixtures in frontend and backend (take 2)

jhawthorn:
A second attempt at #2320. Now with fewer deadlocks!

I believe the issue last time was that frontend also needed to switch to `queue_adapter = :test`. Now all of our projects run with this setting (I probably should have done this to begin with).


jhawthorn:
Frontend passed 10 times locally :crossed_fingers: :shipit:

tvdeyen:
👏

PR #2359, opened 2017-11-07
Declare test dependencies explicitly in gemspecs

jhawthorn:
Build on #2358, which is split out for visibility.

This moves our spec dependencies out of `common_spree_dependencies.rb` and into the individual gemspecs for each project. This adds explicitness (good) at the expense of repetition (bad) and makes the
spec suites faster (good).

### DIFF
--- a/core/config/initializers/assets.rb
+++ b/core/config/initializers/assets.rb
@@ -1,1 +1,3 @@
-Rails.application.config.assets.precompile << 'solidus_core_manifest.js'
+if Rails.application.config.respond_to?(:assets)
+  Rails.application.config.assets.precompile << 'solidus_core_manifest.js'
+end
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -73,8 +73,10 @@ module DummyApp
 
     config.action_controller.include_all_helpers = false
 
-    config.assets.paths << File.expand_path('../dummy_app/assets/javascripts', __FILE__)
-    config.assets.paths << File.expand_path('../dummy_app/assets/stylesheets', __FILE__)
+    if config.respond_to?(:assets)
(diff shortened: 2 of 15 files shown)


ccarruitero:
I'm agree to move testing dependencies from `common_spree_dependencies`, but in order to reduce the repetition I think that we could introduce new files for wrap common testing dependencies.

I was thinking in one file like `common_testing_dependencies` where add testing gems that are repeated in all solidus gems, excluding sample. I think will be mainly the gems added in core without `timecop` and `with_model`. 

An another  like `common_integration_testing_dependencies` where add common integration gems. Here I think will be only `capybara` and `capybara-screenshot` for now.

core/config/initializers/assets.rb
```diff
@@ -1 +1,3 @@
-Rails.application.config.assets.precompile << 'solidus_core_manifest.js'
+if Rails.application.config.respond_to?(:assets)
```
ccarruitero: I think we should remove this changes since are in https://github.com/solidusio/solidus/pull/2358

core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -73,8 +73,10 @@ class Application < ::Rails::Application
 
     config.action_controller.include_all_helpers = false
 
-    config.assets.paths << File.expand_path('../dummy_app/assets/javascripts', __FILE__)
-    config.assets.paths << File.expand_path('../dummy_app/assets/stylesheets', __FILE__)
+    if config.respond_to?(:assets)
```
ccarruitero: same as comment above

core/config/initializers/assets.rb
jhawthorn: This is built on top of #2358, which is necessary for this to pass.

jhawthorn:
Closing in favour of #2407

Issue #2360, opened 2017-11-07
Error using deface with the admin layout in solidus 2.4.0.rc1


afdev82:
I already opened [this issue](https://github.com/solidusio-contrib/solidus_editor/issues/12), I'm preparing a PR for that.
I have this issue with the v2.4.0.rc1, but not with v.2.3

afdev82:
We can close this now.

PR #2361, opened 2017-11-07
use single quote to fix deface override, see #2360

afdev82:
This change seems to fix the #2360 (and so the https://github.com/solidusio-contrib/solidus_editor/issues/12)

### DIFF
--- a/backend/app/views/spree/layouts/admin.html.erb
+++ b/backend/app/views/spree/layouts/admin.html.erb
@@ -10,7 +10,7 @@
     <%= render "spree/admin/shared/flash" %>
     <%= render "spree/admin/shared/spinner" %>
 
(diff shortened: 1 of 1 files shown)


jhawthorn:
I'd prefer this was fixed in deface. It seems like these regexes https://github.com/DefaceCommunity/deface/blob/master/lib/deface/parser.rb#L14-L16
are incorrect

They might need to be
```
/([\w-]+)(\s?=\s?)(")(([^"]*<%.*?%>)+[^"]*)/m,
/([\w-]+)(\s?=\s?)(')(([^']*<%.*?%>)+[^']*)'/m,
/([\w-]+)(\s?=\s?)()(<%.*?%>)(?:\s|>|\z)/m
```

mamhoff:
Oh I had no idea how terrible Deface is. Still, pro this for the reasons @tvdeyen points out.

PR #2362, opened 2017-11-07
Remove unused spec/support/test_gateway

jhawthorn:


jhawthorn:
> Might break an extension, but thats their problem.

Extensions use things from `lib/spree/testing_support`. There's no easy way for outside apps or extensions to require this so I think we are safe.

PR #2366, opened 2017-11-09
Fix rake teaspoon:server in backend

jhawthorn:
`rake teaspoon:server` starts up a server so we can run our JS tests against any web browser.

I broke this when introducing the new dummy_app (sorry).

In fixing it I've also removed the `listen` gem, which is no longer required.

### DIFF
--- a/backend/Rakefile
+++ b/backend/Rakefile
@@ -29,6 +29,10 @@ end
 
 namespace :teaspoon do
   task :server  do
+    require 'teaspoon'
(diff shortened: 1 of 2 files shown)


jhawthorn:
> Should we actually run these tests on CircleCI?

We do run `rake teaspoon` on CI, which runs the tests inside phantomjs (I wonder if we can change this to headless chrome). `rake teaspoon:server` just starts a server and requires the user to point the browser at it.

PR #2372, opened 2017-11-10
Allow code reloading of role configurations

jhawthorn:
This uses the `ClassConstantizer::Set` when storing the mapping of role names to permission sets. In development, this allows those classes to be reloaded if changes are made to a role defined inside the
application.


kennyadsl:
Closing this one since all code here has already been merged in master with #2374

PR #2375, opened 2017-11-15
Locations documentation

benjaminwil:
I've created three short articles about locations in Solidus:

- An overview about zones, countries, and states.
- An article specifically about countries and states. This article may be built upon in the future.
- An article specifically about zones, which includes an example of how a store might set up zones.

My goals are to:

- Give more context to Solidus's basic locations concepts. Developers who are new to Solidus should be able to understand locations from a business perspective and then a development perspective.
- Use article titles, headings, and sub-headings to make it easier for developers to discover the specific information about shipments that they are looking for. They should be able to skim and scan each article quickly.
- Provide some scaffolding for when there's additional documentation about locations.
- Provide more a realistic example of how zones might be set up in a store.
       
### DIFF
--- a/guides/locations/countries-and-states.md
+++ b/guides/locations/countries-and-states.md
@@ -2,7 +2,7 @@
 
 Solidus uses the list of countries and states provided by [Carmen][carmen-repo],
 which uses the countries and states available in the [Debian `iso-codes`
-package][debian-iso-codes]. Solidus creates new `Spree::Country` and
+package][debian-iso-codes]. Solidus creates new `Spree::Country` and 
 `Spree::State` objects for each country and state.
 
 Carmen is generally up-to-date and should provide Solidus with any country or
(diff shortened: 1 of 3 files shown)


guides/locations/countries-and-states.md
```diff
@@ -0,0 +1,35 @@
...
+Rails console:
+
+```
+Spree::Country.create!(iso_name:"NEW COUNTRY", name:"New Country", states_required: true)
+```
+
+## States
+
+A state is any sub-region of a country, whether that is a province, district, or
+territory. If the state you require is not recognized, you can add it directly
```
mamhoff: For some countries, dividing them up into sub-territories makes no sense. For those cases, we can determine that the `states_required` field on the country is set to false. 

If we do that, then the subregions for that country are not chosen for addresses. 

This makes sense for many European countries and any small country.

guides/locations/countries-and-states.md
```diff
@@ -0,0 +1,35 @@
...
+Carmen is generally up-to-date and should provide Solidus with any country or
+state you would ever need.
+
+[carmen-repo]: https://github.com/carmen-ruby/carmen
+[debian-iso-codes]: https://packages.debian.org/sid/all/iso-codes
+
+## Countries
+
+If a country or state you require is not recognized, you can add it directly
+from the Solidus admin (**Settings -> Locations -> Countries**) or from your
```
benjaminwil: You actually can't add states or countries from **Settings -> Locations -> ...** anymore. This needs to be updated.

guides/locations/countries-and-states.md
benjaminwil: I added a note about this. :+1:

PR #2378, opened 2017-11-15
Convert frontend's CoffeeScript to javascript

jhawthorn:
For the same reasoning as #2376

The frontend was previously almost entirely coffeescript, but there wasn't really all that much of it.

Some automated conversion + formatting followed by some manual tweaking gets us a frontend that doesn't depend on coffee-rails.


tvdeyen:
We should not move the `var` declarations from top of the function body’s. JavaScript var hoisting is doing this implicitly and by not having them on top can lead to hard to track down bugs.

PR #2379, opened 2017-11-15
Extract stock transfers to solidus_stock_transfers extension

jhawthorn:
We would like to extract stock transfers to a gem ~(which I haven't done yet, therefore WIP)~ [and here it is!](https://github.com/solidusio-contrib/solidus_stock_transfers)

This removes the stock transfers interface from the admin, extracting it to the new [solidus_stock_transfers](https://github.com/solidusio-contrib/solidus_stock_transfers) gem. This went really well! Nothing really referenced stock transfers, so it's quite a clean split.

Also extracted with this is the `NumberField` component, a JavaScript helper which was only ever used by stock transfers.


mtomov:
That is a good decision! Thanks!

Issue #1837, opened 2017-04-10
Shipping calculators still include cancelled items in shipping cost


ericsaupe:
### Steps to reproduce

1. Complete order with many items and flexirate shipping calculator
2. Cancel items in shipment
3. Adjustment created does not take into account shipping cost
4. Cost on shipping was not changed even though less items will be shipped

### Expected behavior
After an item is cancelled the calculator should only calculate cost based on what is actually going to ship.

### Actual behavior
Shipping calculators are calculating cost on all items even if they are cancelled:

- https://github.com/solidusio/solidus/blob/master/core/app/models/spree/calculator/shipping/flexi_rate.rb#L12
- https://github.com/solidusio/solidus/blob/master/core/app/models/spree/calculator/shipping/flexi_rate.rb#L12
- https://github.com/solidusio/solidus/blob/054e4bd75c65166d316f3f63be85efb48736bcdb/core/app/models/spree/calculator/shipping/per_item.rb#L14
- etc.

### System configuration
**Solidus Version**: 2.1 (seems to be in all)

**Extensions in use**: core, frontend, backend

jacobherrington:
This is still present in Solidus 2.7.

kennyadsl:
@ericsaupe should be closed with #3059, can you please check?

ericsaupe:
It still seems to be a little bit off. Maybe it's just the taxes? Anyway so the setup here is

Shipping Rate Calculator - Flexible Rate by item with first item $1.5 and addition items $1

- Added three different products to the cart
- Completed checkout
- Canceled two of the three items
- The adjustments created did include the additional $1 for shipping but the shipping cost, which is taxed, remained the same. Also the tax adjustments were readjusted and now show a negative adjustment of a few cents. The tax adjustment for the remaining item seems to have disappeared as well.

![screen shot 2019-02-05 at 2 56 26 pm](https://user-images.githubusercontent.com/5138892/52309985-a42f4b80-2956-11e9-86f9-17ef4b01ef7f.png)


This was done against the current `2.9.0.alpha`

Correct me if I'm misunderstanding it all here but I think it should adjust the cost of shipping by the canceled item change, adjust the line item by the cost of the item, and maybe additional adjustments to remove taxes? Just thinking about it from an accounting standpoint where adjustments were created to charge money from a customer and then negative adjustments were created to restore money to the customer against what they paid for.

mdesantis:
[...]

PR #2380, opened 2017-11-16
Cancelled Items totalling issue

gaurav945:
Steps to reproduce the issue : 
1. Place an order, say, containing 1 line item & quantity 4, & the price of item being 100 $, hence the total cost becomes 400 $.

2. Now, go to the Cancel Items tab in the corresponding orders edit page & select all the 4 inventory units & cancel them, together.

3. The final order price should be 
400 - (400/4) - (300/3) - (200/2) - (100/1), which is 0 $, 
but the total comes out to be 
400 - (400/4) - (400/3) - (400/2) - (400/1), which is not 0 $.

4. ALSO, if we cancel all the items, one by one, it works fine.

5. On top of it, there's this open issue as well : https://github.com/solidusio/solidus/issues/1837#issue-220715796

### DIFF
--- a/core/app/models/spree/unit_cancel.rb
+++ b/core/app/models/spree/unit_cancel.rb
@@ -29,6 +29,8 @@ class Spree::UnitCancel < Spree::Base
   # This method is used by Adjustment#update to recalculate the cost.
   def compute_amount(line_item)
     raise "Adjustable does not match line item" unless line_item == inventory_unit.line_item
+    line_item.order.update!
+    line_item = Spree::LineItem.find(line_item.id)
     -(line_item.total.to_d / line_item.inventory_units.not_canceled.reject(&:original_return_item ).size)
   end
 end


cbrunsdon:
Ha, if your description is right this is for sure a big problem. If you're able to write a regression test that would be very appreciated, otherwise someone like @derrickp will take a look.

core/app/models/spree/unit_cancel.rb
```diff
@@ -29,6 +29,8 @@ def adjust!
   # This method is used by Adjustment#update to recalculate the cost.
   def compute_amount(line_item)
     raise "Adjustable does not match line item" unless line_item == inventory_unit.line_item
+    line_item.order.update!
+    line_item = Spree::LineItem.find(line_item.id)
```
mamhoff: I hope this line is not necessary. This amounts to a line item reload, which would result in an out-of-sync object graph - something we've been working pretty hard to avoid.

Besides, I second @cbrunsdon in that this but sounds pretty important and deserves a spec. 

Good find!

gaurav945: Thanks, I just came across it by chance !!
Also, I am not as good a Rails developer as you guys, so, I only understood like 10-20 % of what you guys are talking about, but I am glad that someone noticed it & is working on a fix for it. :)

Also, if you don't mind, @mamhoff could you please explain to me what is an out-of-sync object graph ? I heard this term for the first time.

mamhoff: In almost every request, we load an order with line items, shipments, inventory units, and so on. When reloading anything other than the order, we have two copies of the same database entry in memory - that means the object graph, the network on in-memory-objects is not in sync.

gaurav945: Hey @mamhoff thanks for the explanation. I had no clue of those things & I have been using Solidus for the past 8-9 months. :-P
I guess I still have way lot more to learn. Thanks :))

kennyadsl:
Closed with #3059

PR #2381, opened 2017-11-16
Fix issue loading esmx on checkout frontend

jtapia:
### What it does?
- Fix issue loading es-MX locale on checkout frontend

### Info
```
ActionView::Template::Error (The asset "jquery.validate/localization/messages_esmx.js" is not present in the asset pipeline.):
    33:
    34: <% if I18n.locale != :en %>
    35:   <%= javascript_include_tag 'jquery.validate/localization/messages_' + I18n.locale.to_s.downcase.gsub('-', '') %>
    36: <% end %>
```


jhawthorn:
We're looking to remove jquery.validate in #2264, so this may be unnecessary

jhawthorn:
Thanks for the PR.

I've merged #2264, so I think this is now unnecessary. `es-MX` or any other locale should now work just fine on the frontend by relying on browser validation :tada:.

PR #2382, opened 2017-11-17
Fixes for custom user generator

tvdeyen:
While building an install generator for [alchemy-solidus](https://github.com/AlchemyCMS/alchemy-solidus)
I came across some quirks in the current custom user generator.

First the migration generator does not use the active record migration generator but its super class.
This made is necessary to write our own `next_migration_number` generator. This implementation had some
flaws that lead to duplicated migrations problems. We now use the active record migration generator and
do not need to ship our own custom migration number generator anymore.

Also the generator used to inject the spree initializer and set the `Spree.user_class` to the generated 
custom user. Somehow this got lost. Replaced this by an gsub that ensures that the `Spree.user_class` 
is never set twice.

This removes an unused module (`Rails::Generators::ResourceHelpers`) that was used to deal with controller names
what we do not do anymore in this generator.

On the way I made some code cleanups.

Unfortunately it is not super easy to write generator specs as we do not have a dummy test app anymore. 
We do not have tests for any generator actually. We should, but this is not purpose if this PR.

I hope [the test suite of alchemy-solidus](https://travis-ci.org/AlchemyCMS/alchemy-solidus) is proof enough 
that this generator works. In [its own generator we use the solidus custom user generator](https://github.com/tvdeyen/alchemy-solidus/blob/solidus-2/lib/generators/alchemy/solidus/install/install_generator.rb#L43) to setup the dummy app and test the integration
of Alchemy and Solidus. Would this generator fail, the builds of alchemy-solidus would fail.

### DIFF
--- a/core/lib/generators/spree/custom_user/custom_user_generator.rb
+++ b/core/lib/generators/spree/custom_user/custom_user_generator.rb
@@ -12,10 +12,10 @@ module Spree
     end
 
     def check_for_constant
-        klass
+      klass
     rescue NameError
-        @shell.say "Couldn't find #{class_name}. Are you sure that this class exists within your application and is loaded?", :red
-        exit(1)
+      @shell.say "Couldn't find #{class_name}. Are you sure that this class exists within your application and is loaded?", :red
+      exit(1)
     end
 
     def generate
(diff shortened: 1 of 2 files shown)


tvdeyen:
> This also fixes the duplicate migration issue reported here: AlchemyCMS/alchemy-solidus#14 

Indeed! Nice find @mamhoff

PR #2383, opened 2017-11-17
Remove stringex as a dependency of core

swcraig:
We already have ActiveSupport as part of Solidus, and stringex doesn't
do much else that ActiveSupport doesn't already do. We lose support for
translating non-ascii characters, but `ActiveSupport#parameterize` does
a pretty good job of replacing special characters for ones that can be
used in a valid URL.

This PR also introduces a pluggable `taxon_url_parametizer_class` that defaults to [`ActiveSupport::Inflector`](http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html).

### DIFF
--- a/core/app/models/spree/taxon.rb
+++ b/core/app/models/spree/taxon.rb
@@ -59,7 +59,7 @@ module Spree
     # name and its parents permalink (if present.)
     def set_permalink
       permalink_tail = permalink.split('/').last if permalink.present?
-      permalink_tail ||= name.to_url
+      permalink_tail ||= Spree::Config.taxon_url_parametizer_class.parameterize(name)
       self.permalink_part = permalink_tail
     end
 
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -363,6 +363,13 @@ module Spree
     #   Spree::CurrentStoreSelector
     class_name_attribute :current_store_selector_class, default: 'Spree::StoreSelector::ByServerName'
 
+    # Allows providing your own class for creating urls on taxons
+    #
+    # @!attribute [rw] taxon_url_parametizer_class
+    # @return [Class] a class that provides a `#parameterize` method that
+    # returns a String
+    class_name_attribute :taxon_url_parametizer_class, default: 'ActiveSupport::Inflector'
+
     # Allows providing your own class instance for generating order numbers.
     #
     # @!attribute [rw] order_number_generator
--- a/core/lib/spree/core/permalinks.rb
+++ b/core/lib/spree/core/permalinks.rb
@@ -1,5 +1,3 @@
-require 'stringex'
-
 module Spree
   module Core
     module Permalinks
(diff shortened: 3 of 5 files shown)


core/app/models/spree/taxon.rb
```diff
@@ -59,8 +59,8 @@ def seo_title
     # name and its parents permalink (if present.)
     def set_permalink
       permalink_tail = permalink.split('/').last if permalink.present?
-      permalink_tail ||= name.to_url
-      self.permalink_part = permalink_tail
+      permalink_tail ||= Spree::Config.string_inflection_class.parameterize(name)
+      self.permalink_part = Spree::Config.string_inflection_class.dasherize(permalink_tail)
```
jhawthorn: It's not clear to me why we are using two different methods here. Previously we only used one.

core/lib/spree/app_configuration.rb
```diff
@@ -363,6 +363,14 @@ def payment_canceller
     #   Spree::CurrentStoreSelector
     class_name_attribute :current_store_selector_class, default: 'Spree::StoreSelector::ByServerName'
 
+    # Allows providing your own class for inflecting strings.
+    # For example, to create the permalinks for taxons.
+    #
+    # @!attribute [rw] string_inflection_class
+    # @return [Class] a class that provides `#parameterize`, and `#dasherize`
+    # methods that both return Strings
+    class_name_attribute :string_inflection_class, default: 'ActiveSupport::Inflector'
```
jhawthorn: This needs a name more clearly scoped to what it does. We (through rials) use `ActiveSupport::Inflector` heavily for other things and this name shouldn't imply that we are overriding that. It should be clear that this is for URL parameterization only.

The question here is do we intend to use this for. If we intend to only use this class from Taxon, it should be called `taxon_parametizer_class` (or maybe `taxon_url_parametizer_class`). If we intend to move everything else over, we could call it `parametizer_class` (or maybe `url_parametizer_class`).

It's worth noting that friendly_id has some of its own logic and configuration for parameterization (though I believe it relies on `Inflector.parameterize`). I'm not sure we would want to lose that logic, but it would be nice if we generated URLs consistently.

core/app/models/spree/taxon.rb
swcraig: stringex would "dasherize" everything as part of its `to_url` method. `ActiveSupport::Inflector` doesn't do that, so some [specs related to this](https://github.com/solidusio/solidus/blob/master/core/spec/models/spree/taxon_spec.rb#L71) would fail (`t2_child` is expected to be convereted to `t2-child`, which `ActiveSupport#parameterize` alone does not do).

cbrunsdon: I'd say delete the specs... but thats just me.

jhawthorn: We can either create a new class which calls `parameterize` then `dasherize`, or we can just use `parameterize`.

Since we're already changing behaviour here I think it is fine to just use `parameterize` and update the specs to pass.

PR #2385, opened 2017-11-20
Allow ShipmentsController#remove on ready shipment

alepore:
On the admin forms of ready shipments we have a delete button for each item.
These buttons are not working because the `ShipmentsController#remove` action is only enabled for `pending` shipments. This behaviour comes from https://github.com/solidusio/solidus/pull/97.
I don't know the reasons for that but i see that ready shipments are supposed to be editable (https://github.com/solidusio/solidus/pull/1784).

This PR enables ready shipments item deletion and thus fixes the admin delete buttons, following what was done on #1784.


kennyadsl:
@alepore maybe related to #2331?

PR #2353, opened 2017-11-03
Shipments documentation

benjaminwil:
I am taking the shipping guide documentation merged in #1966 and I am breaking it into several articles about shipments. My goals are to:

- Give more context to Solidus's basic shipping concepts. Developers who are new to Solidus should be able to understand shipments flow from a business perspective and then a development perspective.
- Use article titles, headings, and sub-headings to make it easier for developers to discover the specific information about shipments that they are looking for. They should be able to skim and scan each article quickly.
- Provide more realistic examples of how shipments might be set up in a store. Use zones, shipping methods, and shipping categories in the same way a successful ecommerce business might use them.    

This PR splits the `guides/shipping.md` article into multiple `guides/shipments/*.md` articles. Much of the content has been rewritten and all of it has been reformatted.

The most remarkable changes are the the "Shipment setup examples"  and "Split shipments" articles. I wanted to create more realistic shipment setups, and as of #2199 (Solidus v2.4.0) split shipments use the `Spree::Stock::SimpleCoordinator` instead of using coordinators, adjusters, packers, etc.
       
_This is part a larger project to improve Solidus's documentation. [See this gist with the high-level table of contents](https://gist.github.com/benjaminwil/5e2d2da3d9def6b949b833b1aa2549d3). Where and how this documentation will exist is still up for discussion._

PR #2388, opened 2017-11-21
add payments guide

ccarruitero:
I just ported the guide for payments from spree.

Let me know what you think.

### DIFF
--- a/guides/payments.md
+++ b/guides/payments.md
@@ -189,25 +189,6 @@ debugging why a payment has failed.
 You can get a list of these log entries by calling the `log_entries` on any 
 `Spree::Payment` object. To get the `Active::Merchant::Billing::Response` out of these `Spree::LogEntry` objects, call the `details` method.
 
-## Supported Gateways
-
-Access to a number of payment gateways is handled with the usage of the [solidus_gateway](https://github.com/solidusio/solidus_gateway) extension. This extension currently supports the following gateways:
-
-* Authorize.Net
-* Balanced
-* Beanstram
-* Braintree
-* eWAY
-* LinkPoint
-* Moneris
-* PayPal
-* Sage Pay
-* Samurai
-* Skrill
-* Stripe
-* USA ePay
-* WorldPay
-
 ## Adding your custom gateway
 
 In order to make your custom gateway available on backend list of payment
--- a/guides/payments.md
+++ b/guides/payments.md
@@ -11,7 +11,7 @@ Payment methods typically represent a payment processor. Payment methods will
 process credit card payments, and may also include non credit card payment
 methods such as Check or StoreCredit, which are provided in Solidus by default.
 
-The `Payment` model in Solidus tracks payments against [Orders](orders).
+The `Payment` model in Solidus tracks payments against Orders.
 Payments relate to a `source` which indicates how the payment was made, and a
 `PaymentMethod`, indicating the processor used for this payment.
 
--- a/guides/payments.md
+++ b/guides/payments.md
@@ -1,0 +1,220 @@
+# Payments
+
+## Overview
+
+Solidus has a highly flexible payments model which allows multiple payment
+methods to be available during checkout. The logic for processing payments is
+decoupled from orders, making it easy to define custom payment methods with
+their own processing logic.
+
+Payment methods typically represent a payment processor. Payment methods will
+process credit card payments, and may also include non credit card payment
+methods such as Check or StoreCredit, which are provided in Solidus by default.
+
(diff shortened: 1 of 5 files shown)


guides/payments.md
```diff
@@ -0,0 +1,220 @@
...
+from a payment gateway, it will serialize the object as YAML and store it in the
+database as a log entry for a payment. These responses can be useful for
+debugging why a payment has failed.
+
+You can get a list of these log entries by calling the `log_entries` on any 
+`Spree::Payment` object. To get the `Active::Merchant::Billing::Response` out of these `Spree::LogEntry` objects, call the `details` method.
+
+## Supported Gateways
+
+Access to a number of payment gateways is handled with the usage of the [solidus_gateway](https://github.com/solidusio/solidus_gateway) extension. This extension currently supports the following gateways:
```
ccarruitero: I just used `Stripe` from `solidus_gateway`. So, not sure if this list of supported gateways apply to `solidus_gateway` right now?

guides/payments.md
```diff
@@ -0,0 +1,220 @@
...
+* Moneris
+* PayPal
+* Sage Pay
+* Samurai
+* Skrill
+* Stripe
+* USA ePay
+* WorldPay
+
+## Adding your custom gateway
```
ccarruitero: I think this reference for how to add a custom gateway is too vague. I think we should extend this information in another guide maybe.

guides/payments.md
```diff
@@ -0,0 +1,220 @@
...
+Responses from payment gateways within Solidus are typically
+`ActiveMerchant::Billing::Response` objects. When Solidus handles a response
+from a payment gateway, it will serialize the object as YAML and store it in the
+database as a log entry for a payment. These responses can be useful for
+debugging why a payment has failed.
+
+You can get a list of these log entries by calling the `log_entries` on any 
+`Spree::Payment` object. To get the `Active::Merchant::Billing::Response` out of these `Spree::LogEntry` objects, call the `details` method.
+
+## Supported Gateways
```
ccarruitero: I'm not sure if we should maintain a list of supported gateways here. I think that reference to another resource where we can find gateway extensions could be better.

What do you think?

I was thinking in [extensions.solidus.io](http://extensions.solidus.io). I like how show the builds for each version, so you can know if a project is maintained, but don't have a way to filter the extensions. Another option could be [solition](http://soliton.nebulab.it) that find extensions from github (finding by repository description and name, I think so). Another option?

guides/payments.md
benjaminwil: This is actually on my to-do list!

guides/payments.md
```diff
@@ -0,0 +1,220 @@
...
+Solidus has a highly flexible payments model which allows multiple payment
+methods to be available during checkout. The logic for processing payments is
+decoupled from orders, making it easy to define custom payment methods with
+their own processing logic.
+
+Payment methods typically represent a payment processor. Payment methods will
+process credit card payments, and may also include non credit card payment
+methods such as Check or StoreCredit, which are provided in Solidus by default.
+
+The `Payment` model in Solidus tracks payments against [Orders](orders).
```
benjaminwil: Right now the `[Orders](orders)` link here doesn't go anywhere, and we don't have an equivalent page worth linking to yet.

benjaminwil:
Thanks for this!

In case you haven't seen #2353 and #2375, I'm working through all of the existing Solidus documentation and adding more. In the near future I would like to break up this payments documentation into multiple articles.

I think this would give us more space to provide examples and tutorials, like you pointed out in your comment on the "Adding your custom gateway" sub-article.

cbrunsdon:
Defering to @benjaminwil on my :+1:, but thanks for sending the PR! Updating docs is a terrible, thankless task but we really appreciate it.

guides/payments.md
kennyadsl: I agree we'll probably don't need to keep this list, even because the direction on gateways is trying to build a gem for each geteway for several reason. They don't share any logic in `solidus_gateway` gem and 99% of times stores just use one gateway from that list.

guides/payments.md
kennyadsl: Agree this is useless here as it currently is

guides/payments.md
ccarruitero: I'll remove that link

ccarruitero:
Thanks all for your comments.

I'll update this accordingly

benjaminwil:
Thanks for the updates. I would :+1: this.

There's still more work to be done here (for example, around getting started with custom payment gateways), but the essentials are covered clearly.

Thanks for your work on this @ccarruitero.

PR #2394, opened 2017-11-22
[RFC] Do not render complex preference types as form fields

tvdeyen:
When rendering preference form fields we currently just render every
preference type defined on the preferable class. 

Some classes - like [solidus_paypal_braintree's Gateway](https://github.com/solidusio/solidus_paypal_braintree/blob/e7d8be0443465f686201f2b0b0455d7ff5e72f23/app/models/solidus_paypal_braintree/gateway.rb#L34-L35) - use complex types (`Array` and `Hash`) to store developer facing preferences.

Currently this [leads to bugs](https://github.com/solidusio/solidus_paypal_braintree/issues/132) because these preference field partials are not available.

As arrays and hashes won't represent well in a form field, we should not render them IMO. 
The settings stored in these preferences are considered developer facing only. 
Admins should not change these kind of values.

### DIFF
--- a/backend/app/views/spree/admin/payment_methods/_form.html.erb
+++ b/backend/app/views/spree/admin/payment_methods/_form.html.erb
@@ -19,9 +19,10 @@
             <% if @object.preference_source.present? %>
               <%= t('spree.preference_source_using', name: @object.preference_source) %>
             <% elsif !@object.new_record? %>
-              <% @object.preferences.keys.each do |key| %>
-                <%= render "spree/admin/shared/preference_fields/#{@object.preference_type(key)}",
-                   form: f, attribute: "preferred_#{key}", label: t(key, scope: 'spree') %>
+              <% @object.admin_form_preference_names.each do |name| %>
+                <%= render "spree/admin/shared/preference_fields/#{@object.preference_type(name)}",
+                  form: f, attribute: "preferred_#{name}",
+                  label: t(name, scope: 'spree', default: name.to_s.humanize) %>
               <% end %>
             <% end %>
           </div>
--- a/backend/app/views/spree/admin/promotions/calculators/_default_fields.html.erb
+++ b/backend/app/views/spree/admin/promotions/calculators/_default_fields.html.erb
@@ -1,5 +1,6 @@
-<% calculator.preferences.keys.map do |key| %>
-  <%= render "spree/admin/shared/preference_fields/#{calculator.preference_type(key)}",
-     name: "#{prefix}[calculator_attributes][preferred_#{key}]",
-     value: calculator.get_preference(key), label: t(key.to_s, scope: 'spree') %>
+<% calculator.admin_form_preference_names.map do |name| %>
+  <%= render "spree/admin/shared/preference_fields/#{calculator.preference_type(name)}",
+    name: "#{prefix}[calculator_attributes][preferred_#{name}]",
+    value: calculator.get_preference(name),
+    label: t(name.to_s, scope: 'spree', default: name.to_s.humanize) %>
 <% end %>
--- a/backend/app/views/spree/admin/shared/_calculator_fields.html.erb
+++ b/backend/app/views/spree/admin/shared/_calculator_fields.html.erb
@@ -11,9 +11,10 @@
       <% calculator = f.object.calculator.class == calculator_class ? f.object.calculator : calculator_class.new %>
       <div class="js-calculator-preferences" data-calculator-type="<%= calculator_class %>">
         <%= f.fields_for :calculator, calculator do |calculator_form| %>
-          <% calculator.preferences.keys.each do |key| %>
-            <%= render "spree/admin/shared/preference_fields/#{calculator.preference_type(key)}",
-               form: calculator_form, attribute: "preferred_#{key}", label: t(key, scope: 'spree') %>
+          <% calculator.admin_form_preference_names.each do |name| %>
+            <%= render "spree/admin/shared/preference_fields/#{calculator.preference_type(name)}",
+              form: calculator_form, attribute: "preferred_#{name}",
+              label: t(name, scope: 'spree', default: name.to_s.humanize) %>
           <% end %>
         <% end %>
       </div>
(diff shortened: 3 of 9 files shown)


backend/spec/features/admin/configuration/shipping_methods_spec.rb
```diff
@@ -35,6 +35,46 @@
...
+        fill_in "shipping_method_name", with: "bullock cart"
+
+        within("#shipping_method_categories_field") do
+          check first("input[type='checkbox']")["name"]
+        end
+
+        click_on "Create"
+        select 'Complex Shipments', from: 'Base Calculator'
+        click_on "Update"
+        save_and_open_page
```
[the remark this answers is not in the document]
tvdeyen: UPS 😊

kennyadsl:
Great, I was thinking at what should I do if I need to show just one preference hash or array field, for example because I know it's safe in that point and I want admin to be able to change only that. Is this a scenario we want to support?

kennyadsl:
> Great, I was thinking at what should I do if I need to show just one preference hash or array field, for example because I know it's safe in that point and I want admin to be able to change only that. Is this a scenario we want to support?

Ok, ignore it. I figured out I can just render another input in the view for that field outside the preferences partials code, so there's no need to support that in core.

tvdeyen:
Updated this PR so that it is now possible to overwrite the class method `allowed_admin_form_preference_field_types` on the extending class in order to change the allowed fields. One have to provide their own preference field partials if they do so.

core/spec/models/spree/preferences/preferable_spec.rb
```diff
@@ -105,6 +105,46 @@ class B < A
       })
     end
 
+    describe '#admin_form_preference_field_types' do
+      subject { C.new.admin_form_preference_field_types }
+
+      before do
+        class C
```
mamhoff: `C` is very short, and maybe someone else creates an example class `C`, with different behaviour. In order to avoid conflicts and possibly test order issues because of the conflict, you might want to consider naming the class `ClassWithPreferences` or the like.

mamhoff: Uh-oh, there's also `A`, `B`, and `D`. Not concern of this PR, but 🗡

backend/app/views/spree/admin/payment_methods/_form.html.erb
```diff
@@ -19,9 +19,9 @@
             <% if @object.preference_source.present? %>
               <%= t('spree.preference_source_using', name: @object.preference_source) %>
             <% elsif !@object.new_record? %>
-              <% @object.preferences.keys.each do |key| %>
-                <%= render "spree/admin/shared/preference_fields/#{@object.preference_type(key)}",
-                   form: f, attribute: "preferred_#{key}", label: t(key, scope: 'spree') %>
+              <% @object.admin_form_preference_field_types.each do |type| %>
```
mamhoff: It seems to me this is a wrong change.  Given a class with preferable `:color, type: :string`, previously this would render the form with attribute `preferred_color`, whereas now it seems like it renders `preferred_string`, no?

core/lib/spree/preferences/preferable.rb
```diff
@@ -101,6 +113,30 @@ def default_preferences
...
+      #
+      # Overwrite +allowed_admin_form_preference_field_types+ in your class that
+      # includes +Spree::Preferable+ if you want to provide more fields.
+      # If you do so, you also need to provide a preference field partial
+      # that lives in:
+      #
+      # +app/views/spree/admin/shared/preference_fields/+
+      #
+      # @return [Array]
+      def admin_form_preference_field_types
```
mamhoff: Previously, I misread code in a future commit because this method includes `types`, prompting me to thing the result of this method is an array of `type` strings. Let's consider renaming it to `admin_editable_preferences`, so it communicates more clearly it returns an array of preferences?

backend/app/views/spree/admin/payment_methods/_form.html.erb
mamhoff: No, we're just not rendering a `type` here.

tvdeyen: Ah, yes. You are right. Dang. Thanks for catching this!

tvdeyen:
@mamhoff please review again

backend/app/views/spree/admin/promotions/calculators/_default_fields.html.erb
```diff
@@ -1,5 +1,5 @@
-<% calculator.preferences.keys.map do |key| %>
-  <%= render "spree/admin/shared/preference_fields/#{calculator.preference_type(key)}",
-     name: "#{prefix}[calculator_attributes][preferred_#{key}]",
-     value: calculator.get_preference(key), label: t(key.to_s, scope: 'spree') %>
+<% calculator.admin_form_preference_names.map do |type| %>
```
mamhoff: `type` should be `name` here.

backend/app/views/spree/admin/payment_methods/_form.html.erb
```diff
@@ -19,9 +19,9 @@
             <% if @object.preference_source.present? %>
               <%= t('spree.preference_source_using', name: @object.preference_source) %>
             <% elsif !@object.new_record? %>
-              <% @object.preferences.keys.each do |key| %>
+              <% @object.admin_form_preference_names.each do |key| %>
```
mamhoff: `key` should be `name` here.

backend/app/views/spree/admin/shared/_calculator_fields.html.erb
```diff
@@ -11,9 +11,9 @@
       <% calculator = f.object.calculator.class == calculator_class ? f.object.calculator : calculator_class.new %>
       <div class="js-calculator-preferences" data-calculator-type="<%= calculator_class %>">
         <%= f.fields_for :calculator, calculator do |calculator_form| %>
-          <% calculator.preferences.keys.each do |key| %>
-            <%= render "spree/admin/shared/preference_fields/#{calculator.preference_type(key)}",
-               form: calculator_form, attribute: "preferred_#{key}", label: t(key, scope: 'spree') %>
+          <% calculator.admin_form_preference_names.each do |type| %>
```
mamhoff: `type` should be `name` here.

tvdeyen:
@mamhoff addressed

tvdeyen:
@kennyadsl could you please re-review and give approval? Thanks

PR #2396, opened 2017-11-23
Convert product, variant, stock item, prices to discard

jhawthorn:
This deprecates the `#paranoia_destroy` (and therefore also `#destroy`) method on these models in favour of using #discard.

`after_discard` callbacks have been added to these models to duplicate the exact current behaviour of calling `#destroy` (`before_destroy` callbacks and destroying of `dependent: :destroy` associations). In the future we probably want to change this behaviour (why do we delete images when soft-deleting variants?), but that isn't being done here.

### DIFF
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -1,3 +1,5 @@
+require 'discard'
+
 module Spree
   # Products represent an entity for sale in a store. Products can have
   # variations, called variants. Product properties include description,
@@ -12,6 +14,18 @@ module Spree
     friendly_id :slug_candidates, use: :history
 
     acts_as_paranoid
+    include Spree::ParanoiaDeprecations
+
+    include Discard::Model
+    self.discard_column = :deleted_at
+
+    after_discard do
+      variants_including_master.discard_all
+      self.product_option_types = []
+      self.product_properties = []
+      self.classifications = []
+      self.product_promotion_rules = []
+    end
 
     has_many :product_option_types, dependent: :destroy, inverse_of: :product
     has_many :option_types, through: :product_option_types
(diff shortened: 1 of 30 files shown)


core/app/models/spree/product.rb
```diff
@@ -13,6 +15,25 @@ class Product < Spree::Base
...
+      self.product_option_types = []
+      self.product_properties = []
+      self.classifications = []
+      self.product_promotion_rules = []
+    end
+
+    def paranoia_destroy
+      Spree::Deprecation.warn <<-WARN.strip_heredoc
+        Calling #destroy (or #paranoia_destroy) on a Product currently performs a soft-destroy using the paranoia gem.
+        In Solidus 3.0, paranoia will be removed, and this will perform a HARD destoy instead. To continue soft-deleting, use #discard instead.
```
swcraig: "...perform a HARD _destoy_".  All three of the deprecation warnings have this.

jhawthorn: This also prompted me to move these warnings to a common `ParanoiaDeprecations` module.

core/spec/models/spree/variant_spec.rb
```diff
@@ -654,19 +654,43 @@
...
+      variant.images = [create(:image)]
 
-    before { variant.paranoia_destroy }
+      expect(variant.stock_items).not_to be_empty
+      expect(variant.prices).not_to be_empty
+      expect(variant.currently_valid_prices).not_to be_empty
 
-    it "should keep its price if deleted" do
-      expect(variant.display_price).to eq(previous_variant_price)
+      variant.destroy
```
kennyadsl: shouldn't be `variant.discard`?

PR #2397, opened 2017-11-24
[WIP] fix Importer::Order to allow create a shipments

ccarruitero:
I found some problems when trying to create and order with shipment through api.

The errors comes from `Importer::Order`, that the api use for [create orders](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/orders_controller.rb#L30)

The first error comes from the [shipment creation](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/importer/order.rb#L51-L84), that was using parameters that are [not permitted](https://github.com/solidusio/solidus/blob/master/core/lib/spree/permitted_attributes.rb#L82)

And other problem was that when create line items, how this call [order.contents.add](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/importer/order.rb#L102), after item is [added](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_contents.rb#L74) the  order [finish destroying all shipments](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L556)

This update `create_shipments_from_params` to use permitted parameters and move the method to be called just after line items to avoid be deleted.

### DIFF
--- a/api/spec/requests/spree/api/orders_controller_spec.rb
+++ b/api/spec/requests/spree/api/orders_controller_spec.rb
@@ -732,6 +732,20 @@ module Spree
           expect(response.status).to eq 201
           expect(json_response["user_id"]).to eq(user.id)
         end
+
+        it "with a item and shipment" do
+          stock_location = create(:stock_location)
+          post spree.api_orders_path, params: {
+            order: {
+              line_items: {
+                "0" => { variant_id: variant.to_param, quantity: 5 }
+              },
+              shipments_attributes: [{ stock_location_id: stock_location.id }]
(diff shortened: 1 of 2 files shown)


kennyadsl:
@ccarruitero thanks for this PR but it's stale from a lot of time and we are going to close it. If you want to keep working on this we can reopen it, just let us know!

ccarruitero:
Oopps... I missed this. It's from some time ago.

I'll take a look into the changes and current master, and maybe open a new PR if there is something to fix yet

PR #2399, opened 2017-11-24
Harden frontend checkout spec

tvdeyen:
Capybara's `current_path` is not an auto-waiting matcher, that's why the
spec fails from time to time. By using a waiting matcher `has_content`
we make use of the Capybara auto-waiting and retry feature. This should
stabalize the spec.

### DIFF
--- a/frontend/spec/features/checkout_spec.rb
+++ b/frontend/spec/features/checkout_spec.rb
@@ -341,7 +341,7 @@ describe "Checkout", type: :feature, inaccessible: true do
       }.not_to change { Spree::CreditCard.count }
 
       click_on "Place Order"
-      expect(page).to have_current_path(spree.order_path(Spree::Order.last))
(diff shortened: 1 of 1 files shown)


jhawthorn:
As far as I can tell `current_path` [does wait for the path to chage](https://github.com/teamcapybara/capybara/blob/7e2e6e433ba83617bf98cd4c11d8b54581eaac94/lib/capybara/session/matchers.rb#L73).

Do you have an example build on CI where this has failed?

tvdeyen:
@jhawthorn yes: https://circleci.com/gh/tvdeyen/solidus/160

jhawthorn:
Hmm. I can get that failure locally. It looks like the payment isn't succeeding.



When I take the screenshot before submitting the button it looks like the credit card hasn't been entered correctly.



So I think this is the same issue as #2386, but with the CC field, not the expiry.

tvdeyen:
Close in favour of #2404

PR #2401, opened 2017-11-24
Recalculate order inside UnitCancel adjust!

derrickp:
This follows along with #2380 and provides a spec for the issue and an alternate fix. I have changed the `order.update!` call to be `order.recalculate` and moved it into the `adjust!` method. This keeps the original compute function clear of side effects.

### DIFF
--- a/core/app/models/spree/unit_cancel.rb
+++ b/core/app/models/spree/unit_cancel.rb
@@ -24,6 +24,9 @@ class Spree::UnitCancel < Spree::Base
       eligible: true,
       finalized: true
     )
+
+    inventory_unit.line_item.order.recalculate
+    adjustment
   end
 
   # This method is used by Adjustment#update to recalculate the cost.
--- a/core/spec/models/spree/unit_cancel_spec.rb
+++ b/core/spec/models/spree/unit_cancel_spec.rb
@@ -55,5 +55,27 @@ RSpec.describe Spree::UnitCancel do
         expect { subject }.to raise_error RuntimeError, "Adjustable does not match line item"
       end
     end
+
+    context "multiple inventory units" do
+      let(:line_item) { create(:line_item, quantity: 4, price: 100.0) }
+
+      let(:inventory_units) do
+        [
+          create(:inventory_unit, line_item: line_item),
+          create(:inventory_unit, line_item: line_item),
+          create(:inventory_unit, line_item: line_item),
+          create(:inventory_unit, line_item: line_item),
+        ]
+      end
+
+      it "properly creates adjustments for line_item" do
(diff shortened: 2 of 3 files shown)


core/app/models/spree/unit_cancel.rb
```diff
@@ -24,6 +24,9 @@ def adjust!
       eligible: true,
       finalized: true
     )
+
+    inventory_unit.line_item.order.recalculate
```
kennyadsl: what about manually updating the adjustment_total instead?

```ruby
inventory_unit.line_item.adjustment_total += amount
```

This would ensure it's updated for next iterations without hard refreshing all the order content

core/spec/models/spree/unit_cancel_spec.rb
```diff
@@ -55,5 +55,27 @@
...
+        ]
+      end
+
+      it "properly creates adjustments for line_item" do
+        inventory_units.each do |inventory_unit|
+          Spree::UnitCancel.create!(inventory_unit: inventory_unit, reason: Spree::UnitCancel::SHORT_SHIP).adjust!
+          inventory_unit.cancel!
+          line_item.reload
+        end
+        expect(line_item.total.to_d).to eq(0)
```
kennyadsl: I think it's giving a false positive here.

You should test removing reload and with:

```
expect(inventory_units.first.line_item.total.to_d).to eq(0)
```

core/app/models/spree/unit_cancel.rb
derrickp: I tried that, but the line items that are on the inventory units were all different objects that pointed to the same line item record. So each time the code came through the adjustment_total was still at 0.

kennyadsl: mhh, I tried this code as well and it seems to work with your spec (removing the `line_item.reload`)

derrickp: It worked for me in the spec as well, but once I tried it in the sandbox app it failed. I will try again however and see if it was just a fluke failure.

derrickp:
I've added a new spec to `order_cancellations` and I've refactored the spec in `unit_cancel` so that it doesn't do a reload anymore.

aldesantis:
@kennyadsl @derrickp FYI, there's also a problem with taxes when cancelling items. The taxes should be subtracted from `line_item.total` because they're already recalculated on top of the new net total:
```ruby
-((line_item.total - line_item.additional_tax_total).to_d / line_item.inventory_units.not_canceled.reject(&:original_return_item).size)
```

kennyadsl:
Closed with #3059

PR #2407, opened 2017-11-28
Use a single top-level Gemfile for test dependencies

jhawthorn:
Similar to the goals (and some shared changes) from #2359, but with a different approach.

Currently we have one Gemfile per solidus sub-project. This was done to keep our projects isolated ex. so solidus_backend isn't loaded when we're only testing solidus_frontend.

However, we can achieve the same separation by simply not `require`ing the other gems. We can either set `require: false` in the Gemfile, or stop calling `Bundle.require`. I did both. The new dummy_app no longer has any knowledge of rubygems or bundler, and all test gems are now `require: false`.

Advantages to this:
* Startup (especially core specs) should be much faster since it doesn't load unnecessary gems
* Simpler for developers (only need to bundle install in one folder)
* Faster to bundle install once vs 5 times
* CircleCI (or any CI) gem caching is much simpler
* Users of `BUNDLE_PATH: "vendor/bundle"` :wave: don't have to fight bundler installing gems 5 times.

Disadvantages:
* We might miss a dependency, if we `require` it without adding it to the `gemspec`.

This is the same approach to multi-sub-project Gemfile management taken by [rails](https://github.com/rails/rails/blob/master/Gemfile). So I have some confidence it is okay.


Gemfile
```diff
@@ -1,3 +1,52 @@
-eval(File.read(File.dirname(__FILE__) + '/common_spree_dependencies.rb'))
+source 'https://rubygems.org'
 
-gemspec
+gemspec require: false
+
+platforms :ruby do
+  gem 'mysql2'
```
tvdeyen: It seems unnecessary for me to require all database connectors at once. 
Shouldn't we only load those we actually use via `ENV['DB']`?

jhawthorn: I've added `require: false` here too

PR #2408, opened 2017-11-28
Deprecate weird taxon product filters

cbrunsdon:
This doesn't look like these were ever used, and I can't imagine anyone
finding them randomly and using them.

Better to pull these out of core and let people add it themselves if they
want it.

Also, they're untested.

### DIFF
--- a/core/lib/spree/core/product_filters.rb
+++ b/core/lib/spree/core/product_filters.rb
@@ -164,6 +164,7 @@ module Spree
       # This scope selects products in any of the active taxons or their children.
       #
       def self.taxons_below(taxon)
+        Spree::Deprecation.warn "taxons_below is deprecated in solidus_core. Please add it to your own application to continue using it."
         return Spree::Core::ProductFilters.all_taxons if taxon.nil?
         {
           name:   'Taxons under ' + taxon.name,
--- a/core/lib/spree/core/product_filters.rb
+++ b/core/lib/spree/core/product_filters.rb
@@ -181,6 +181,7 @@ module Spree
       #
       # idea: expand the format to allow nesting of labels?
       def self.all_taxons
+        Spree::Deprecation.warn "all_taxons is deprecated in solidus_core. Please add it to your own application to continue using it."
         taxons = Spree::Taxonomy.all.map { |t| [t.root] + t.root.descendants }.flatten
         {
           name:   'All taxons',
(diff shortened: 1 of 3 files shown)


cbrunsdon:
@kennyadsl @mamhoff I threw another commit on here before I saw your approval. I can either split this up into another PR or you can weigh in on it being okay.

core/lib/spree/core/product_filters.rb
```diff
@@ -164,11 +164,11 @@ def self.selective_brand_filter(taxon = nil)
       # This scope selects products in any of the active taxons or their children.
       #
       def self.taxons_below(taxon)
+        Spree::Deprecation.warn "taxons_below is deprecated in solidus_core. Please add it to your own application to continue using it."
         return Spree::Core::ProductFilters.all_taxons if taxon.nil?
         {
           name:   'Taxons under ' + taxon.name,
-          scope:  :taxons_id_in_tree_any,
-          labels: taxon.children.sort_by(&:position).map { |t| [t.name, t.id] },
+          scope:  :taxons_id_in_tree_any, labels: taxon.children.sort_by(&:position).map { |t| [t.name, t.id] },
```
kennyadsl: I'd keep it on 2 lines

cbrunsdon: nice catch, sorry. fixed

frontend/app/views/spree/shared/_filters.html.erb
```diff
@@ -1,4 +1,4 @@
-<% filters = @taxon ? @taxon.applicable_filters : [Spree::Core::ProductFilters.all_taxons] %>
+<% filters = @taxon.applicable_filters %>
```
kennyadsl: Maybe this partial is or was rendered in some view that doesn't have `@taxon`?

cbrunsdon: not anywhere in solidus_frontend, the only time its rendered I can find is here:

https://github.com/solidusio/solidus/blob/master/frontend/app/views/spree/products/index.html.erb#L1-L9

And in that case only if `@taxon` exists.

I personally have a vendetta against these methods because I only barely understand what is happening and don't think they're used or valuable.

mamhoff:
I keep my approval.

frontend/app/views/spree/shared/_filters.html.erb
kennyadsl: Oh, yes, seems like it's safe to remove them. I agree they are pretty useless, thanks!

PR #2409, opened 2017-11-28
Move Spree::Taxon#applicable_filters

cbrunsdon:
How we handle product filters in core is pretty janky, as they're only used by our frontend but interwoven through core.

As is clear from [the documentation of product_filters](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/product_filters.rb):

> THIS FILE SHOULD BE OVER-RIDDEN IN YOUR SITE EXTENSION!
>   the exact code probably won't be useful, though you're welcome to modify and reuse
>   the current contents are mainly for testing and documentation

By moving the `Spree::Taxon#applicable_filters` call (which was untested) into the frontend, we can eventually remove the applicable_filters call in `Spree::Taxon` (along with its `@note This method is meant to be overridden on a store by store basis.`), which is just as easy to customize in a helper.

With #applicable_filters removed, we can then remove the product_filters requirement of Spree::Taxon.

### DIFF
--- a/frontend/app/helpers/spree/taxon_filters_helper.rb
+++ b/frontend/app/helpers/spree/taxon_filters_helper.rb
@@ -1,0 +1,11 @@
+require 'spree/core/product_filters'
+
+module Spree
+  module TaxonFiltersHelper
+    def applicable_filters_for(_taxon)
+      [:brand_filter, :price_filter].map do |filter_name|
+        Spree::Core::ProductFilters.send(filter_name) if Spree::Core::ProductFilters.respond_to?(filter_name)
+      end.compact
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


kennyadsl:
There's a conflict to be fixed, anyway I'm good with this change, thanks!

frontend/app/helpers/spree/taxon_filters_helper.rb
```diff
@@ -0,0 +1,11 @@
+require 'spree/core/product_filters'
+
+module Spree
+  module TaxonFiltersHelper
+    def applicable_filters_for(_taxon)
+      [:brand_filter, :price_filter].map do |filter_name|
+        Spree::Core::ProductFilters.send(filter_name) if Spree::Core::ProductFilters.respond_to?(filter_name)
```
jhawthorn: I know this was copy-pasted from the existing implementation, but it's weird since both methods are already defined on `ProductFilters`. Expecting users to `remove_method` as the means of customization seems quite odd.

This of course is probably desirable for compatibility and shouldn't block this PR.

PR #2255, opened 2017-10-04
Some admin UI enhancements

tvdeyen:
This PR includes a number of small fixes I collected while working on several other PRs.

This is best reviewed commit by commit. 

If it is preferred to open separate PRs for some of this changes I am happily willing to do so.

### Screens

Select2 Dropdown opened above instead of below

#### Before

select2 above - before

#### After

select2 above - after

Variant auto complete

#### Before

variant autocomplete before

#### After

variant autocomplete padding

Credit card based payment methods form

#### Before

new payment - payments - r929121340 - orders 2017-09-21 11-40-03

#### After

new payment - payments - r929121340 - orders 2017-09-21 11-20-52

PR #2411, opened 2017-11-29
Admin order payments ui cleanup

tvdeyen:
Extracted from #2255 

Made some changes to the admin order payments UI:

- Fixes the gateway payment method form
- Make the number with currency component suitable for smaller widths
- Re-arrange the order payments table

### Screenshots

localhost_3000_admin_orders_r123456789_payments_new laptop with hidpi screen 1

localhost_3000_admin_orders_r123456789_payments_new laptop with hidpi screen

payments - before

payments - after


backend/app/views/spree/admin/payments/_list.html.erb
```diff
@@ -1,31 +1,47 @@
...
             <%= t(payment.state, scope: 'spree.payment_states') %>
           </span>
         </td>
+        <td class="amount align-right">
+          <div class="editing-show">
+            <div class="input-group">
+              <div class="input-group-addon number-with-currency-symbol">
+                <%= ::Money::Currency.find(@order.currency).symbol %>
+              </div>
+              <%= number_field_tag :amount, payment.amount, class: 'js-edit-amount align-right form-control', step: :any %>
```
kennyadsl: We decided to [use "text" input for currency fields](https://github.com/solidusio/solidus/pull/2232), right?

tvdeyen: @kennyadsl Fixed it

backend/app/assets/javascripts/spree/backend/views/payment/payment_row.js.coffee
```diff
@@ -0,0 +1,24 @@
+Spree.Views.Payment.PaymentRow = Backbone.View.extend
```
ccarruitero: I think the file should be in javascript instead coffee.

From [backend README](https://github.com/solidusio/solidus/tree/master/backend):

>Though we have existing CoffeeScript files, any new files should be in javascript (ES5).

kennyadsl: Ah, you are right, good catch!

backend/app/assets/javascripts/spree/backend/views/payment/payment_row.js.coffee
```diff
@@ -0,0 +1,24 @@
...
+  events:
+    "click .js-edit": "onEdit"
+    "click .js-save": "onSave"
+    "click .js-cancel": "onCancel"
+
+  onEdit: (e) ->
+    e.preventDefault()
+    @$el.addClass("editing")
```
ccarruitero: Surely is outside the scope of this changes.

But we have similar backbone views that use a conditional value `editing` with a handlebars template.  (ex: [orders/shipping_methods](https://github.com/solidusio/solidus/blob/926ea6880b232b353d8cea1479a2f1bcc41ac279/backend/app/assets/javascripts/spree/backend/views/order/shipping_method.js#L18)) I think that approach is better than show/hide with css.

Maybe we can create a common view for use in this cases.

backend/app/assets/javascripts/spree/backend/views/payment/payment_row.js.coffee
tvdeyen: Will do, thanks!

backend/app/assets/javascripts/spree/backend/views/payment/payment_row.js.coffee
tvdeyen: Absolutely. Will see if I can sneak this into this PR

tvdeyen:
@kennyadsl made the requested changes

PR #2417, opened 2017-11-30
Fix address validation switching to a country without states

kennyadsl:
Saving an address with a country without states, after saving an address that instead has states will raise a validation error, since we are not resetting the state. We always had this issue, it's just raising after #2371, which describes the same issue but for backend.

This PR fixes this problem by adding hidden fields that reset `state_id` and/or `state_name` if they are disabled (via js).

#### Why WIP?

It needs some specs as well, but I'd like to understand if it's better to perform this kind of state resets at model level instead. Any comment is appreciated!

### DIFF
--- a/frontend/app/assets/javascripts/spree/frontend/checkout/address.js
+++ b/frontend/app/assets/javascripts/spree/frontend/checkout/address.js
@@ -43,6 +43,7 @@ Spree.ready(function($) {
       var statePara = $("#" + region + "state");
       var stateSelect = statePara.find("select");
       var stateInput = statePara.find("input");
+      var stateSpanRequired = statePara.find('[id$="state-required"]');
       if (states.length > 0) {
         var selected = parseInt(stateSelect.val());
         stateSelect.html("");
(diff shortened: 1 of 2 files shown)


mamhoff:
I think that it makes sense to do this at the model level: If the country is changed to one that does not require states, set the `state_id` to `nil`.

kennyadsl:
@tvdeyen rebased and ready for a second review!

frontend/app/assets/javascripts/spree/frontend/checkout/address.js
```diff
@@ -43,6 +43,7 @@ Spree.ready(function($) {
       var statePara = $("#" + region + "state");
       var stateSelect = statePara.find("select");
       var stateInput = statePara.find("input");
+      var stateSpanRequired = statePara.find('[id$="state-required"]');
```
tvdeyen: Where is this var used?

kennyadsl: mhhh we've probably had some change in that file in the meantime, or I missed something while rebasing. I honestly can't recall why that var was required, I'll investigate soon, thanks :)

kennyadsl:
I think I'll go with fixing this at a model level if everyone agree it's preferable, thanks for the feedback!

kennyadsl:
Closing in favor of #3129

PR #2419, opened 2017-12-01
Use I18n date format for `pretty_time` helper

tvdeyen:
Instead of concatenating strings we should use a `i18n` date format instead.

Also introduces support for custom Solidus' time formats. Currently we are using the Rails `:long` date format that Rails apps and extensions might override. Instead we use our own `solidus` namespace to avoid conflicts.

```yml
en:
  time:
    formats:
      solidus:
        long: ...
```


tvdeyen:

PR #2420, opened 2017-12-01
Improve admin return authorization controller

kennyadsl:
With #2284 we noticed we had no specs for the `fire` action in admin ReturnAuthorization controller. 

This PR adds specs for that method and handle when trying to fire a method that does not exist on the ReturnAuthorization instance.

### DIFF
--- a/backend/app/controllers/spree/admin/return_authorizations_controller.rb
+++ b/backend/app/controllers/spree/admin/return_authorizations_controller.rb
@@ -12,8 +12,8 @@ module Spree
       def fire
         action_from_params = "#{params[:e]}!"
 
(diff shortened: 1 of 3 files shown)


tvdeyen:

kennyadsl:
After talking with @gmacdougall we agreed that it's better to whitelist the accepted methods here. As a next step it will probably make sense to try to refactor the logic that required this `fire` action to exist.

kennyadsl:
The only possible event is `cancel!` so it does not make a lot of sense having this metaprogrammed stuff in there just for that. 

As next iteration we could remove this `fire` action and create another controller that is responsible just for canceling return authorization instead.

PR #2421, opened 2017-12-01
Remove zombie tooltips with MutationObserver

kennyadsl:
Previously, we was removing tooltips of elements that are removed from DOM (they remain on the page permanently) by [polling the element to see if it's still hovered](https://github.com/solidusio/solidus/pull/996).

Instead of polling the tooltip to see if it is still hovered, this commit introduce a mutation observer.


tvdeyen:

PR #2426, opened 2017-12-01
Store credit admin UI fixes

tvdeyen:
Builds on top of #2419 

### Before

store credit - admin example com - users 2017-12-01 23-44-25

edit - store credit - store credits - admin example com - users 2017-12-01 23-45-29

### After

store credit - admin example com - users 2017-12-01 23-49-31

edit - store credit - store credits - admin example com - users 2017-12-01 23-45-08

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -868,15 +868,17 @@ en:
           address: Address
       store_credits:
         add: "Add store credit"
-        amount_authorized: "Amount Authorized"
-        amount_credited: "Amount Credited"
-        amount_used: "Amount Used"
+        amount_authorized: Authorized
+        amount_credited: Credited
+        amount_used: Used
+        created_at: Issued On
         back_to_edit: "Back to edit"
         back_to_user_list: "Back to user list"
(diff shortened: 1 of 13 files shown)


core/config/locales/en.yml
```diff
@@ -305,10 +305,10 @@ en:
         cart_tax_country_iso: Tax Country for Empty Carts
       spree/store_credit:
         amount: Amount
-        amount_authorized: Amount Authorized
-        amount_credited: Amount Credited
-        amount_used: Amount Used
-        category_id: Credit Type
+        amount_authorized: Authorized
+        amount_credited: Credited
+        amount_used: Used
```
kennyadsl: I'm not sure these 3 keys are ok, since it could not be clear that they refer to the amount. 

Eg. 

```
Used cannot be greater than the credited amount
```

I don't get what we gain removing the term `Amount`

tvdeyen: You are right. I will change that

tvdeyen:
@kennyadsl made the requested changes. Found old unused translation keys and used them for the table headers.

PR #2428, opened 2017-12-04
Remove delimiter on number with currency field

lgiacalone3:
Noticed a weird bug when editing a promotion - after updating the promotion, the item total dropped from 1000 down to 1. Looks like the delimiter is causing the amount to get truncated when values are 1000 and over.

Probably due to this:
```
irb(main):001:0> '1,000'.to_i
=> 1
```

This removes the delimiter from the `number_with_currency` partial, which is used on the promotion edit page. If there are any other fields that should get a similar treatment, feel free to let me know - happy to make the change.

### DIFF
--- a/backend/app/views/spree/admin/shared/_number_with_currency.html.erb
+++ b/backend/app/views/spree/admin/shared/_number_with_currency.html.erb
@@ -5,7 +5,7 @@
 
 <div class="input-group number-with-currency <%= "number-with-currency-with-select" unless currency %> js-number-with-currency">
   <div class="input-group-addon number-with-currency-symbol"></div>
-  <%= f.text_field amount_attr, value: number_to_currency(f.object.public_send(amount_attr), unit: ''), class: 'form-control number-with-currency-amount', required: required %>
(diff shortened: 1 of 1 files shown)


backend/app/views/spree/admin/shared/_number_with_currency.html.erb
```diff
@@ -5,7 +5,7 @@
 
 <div class="input-group number-with-currency <%= "number-with-currency-with-select" unless currency %> js-number-with-currency">
   <div class="input-group-addon number-with-currency-symbol"></div>
-  <%= f.text_field amount_attr, value: number_to_currency(f.object.public_send(amount_attr), unit: ''), class: 'form-control number-with-currency-amount', required: required %>
+  <%= f.text_field amount_attr, value: number_to_currency(f.object.public_send(amount_attr), unit: '', delimiter: ''), class: 'form-control number-with-currency-amount', required: required %>
```
luukveenis: This can probably be a `number_field` as well. It's only used in admin forms for numeric values.

lgiacalone3:
@luukveenis fixed!

lgiacalone3:
@jhawthorn @luukveenis Just removed that last commit.

jhawthorn:
Thank you

Issue #2430, opened 2017-12-05
[RFC] Events in Solidus


swcraig:
@derrickp and I have been working on making user notifications more customizable.

Historically, Solidus has handled “transactional” communication to users with email, sent using [ActionMailer](https://github.com/rails/rails/tree/master/actionmailer). “Transactional” emails notify users that:
- Order has been confirmed
- Order has been cancelled
- Shipment has shipped

Additionally, in some cases we use email to notify admins that their operations have finished processing, as is the case in reimbursements and order inventory issues. Popular extensions have also used our ActionMailer settings, with [solidus_auth_devise](https://github.com/solidusio/solidus_auth_devise) using ActionMailer to send notifications for password reset or recovery.

How stores handle notifications to users is an integral part of the experience to shopping online. As Solidus stores evolved, external notification platforms (SMS, email-templating-as-a-service, A/B testing of email), individual stores have had to implement these alternate notification channels using very intrusive overrides to solidus core. While ActionMailer is the only notification system offered in core, it is not widely used and is often turned off and discarded.

We believe that Solidus core needs to support a clear, extendable, configurable system to give stores tighter control over how and when notifications are sent to users. Our goal would be to allow stores to implement whatever notification systems they desire without the need to override core. 

As almost all large Solidus stores have had to completely re-write or overwrite core to build complex user notifications, it is a long-standing problem in the Spree/Solidus ecosystem.

There have already been attempts to make behaviour customizable by adding “hooks” around specific actions. For example:
- [in reimbursements](https://github.com/solidusio/solidus/blob/354eb21a5469cc5487440e978ac780a06f5676cc/core/app/models/spree/reimbursement.rb#L108)
- [in orders](https://github.com/solidusio/solidus/blob/354eb21a5469cc5487440e978ac780a06f5676cc/core/app/models/spree/order.rb#L431)

These have been largely unsuccessful as they are obscure, undocumented, and insufficient for most stores needs.

In addition to at least [one large-scale attempt](https://github.com/solidusio/solidus/pull/1218) to tackle the problem, we’ve made two large and distinct attempts to achieve the goal of a customizable system:
  - [Command based approach](https://github.com/swcraig/solidus/pull/11)
  - [Observer based approached](https://github.com/swcraig/solidus/pull/9)

We don’t believe any of our previous approaches, or any attempt made or pitched to Solidus Core, have gone far enough to solve the problems mentioned above. We need to take a more “radical” approach to solve the problem.

[...]

PR #2431, opened 2017-12-07
Add event bus

swcraig:
Related to #2430, this is an "event bus" implementation that @derrickp and I have been working on.

`Spree::MailProcessor` subscribes to `Spree::EventBus` which publishes events.

`Spree::MailProcessor` can be used as a template for how to create new processors (and how to subscribe to events). The processor implements a `#register!` method defining its subscriptions. This class will presumably be ripped out into its own extension after this PR is merged.

This PR does **not** handle hierarchical events or asynchronous/distributed processing.


core/app/models/spree/order.rb
```diff
@@ -432,7 +432,10 @@ def finalize!
 
       touch :completed_at
 
-      deliver_order_confirmation_email unless confirmation_delivered?
+      Spree.event_bus.publish(
+        :order_confirmed,
+        Spree::Events::OrderConfirmedEvent.new(order: self)
```
kennyadsl: what about using configurable classes instead? `Spree::Config.order_confirmed_event_class.new(...)`

core/lib/spree/core/engine.rb
```diff
@@ -39,6 +39,10 @@ class Engine < ::Rails::Engine
         Migrations.new(config, engine_name).check
       end
 
+      initializer 'spree.core.register_mail_processor' do |_app|
```
kennyadsl: what about a more generic `initializer 'spree.core.register_event_processors'`?

core/lib/spree/events/processors/mail_processor.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Events
+    module Processors
+      class MailProcessor
+        class << self
+          def register!
+            Spree::Config.event_bus.subscribe(:order_confirmed,           ->(event) { send_confirm_email(event) })
+            Spree::Config.event_bus.subscribe(:order_cancelled,           ->(event) { send_cancel_email(event) })
+            Spree::Config.event_bus.subscribe(:carton_shipped,            ->(event) { send_carton_shipped_emails(event) })
+            Spree::Config.event_bus.subscribe(:reimbursement_processed,   ->(event) { send_reimbursement_email(event) })
+            Spree::Config.event_bus.subscribe(:order_inventory_cancelled, ->(event) { send_inventory_cancellation_email(event) })
```
kennyadsl: is it possible (easy) for stores or extensions to add other subscriptions to this list? One way could be creating a custom `MailProcessor` class that inherits from this one, but we should probably use something like

```ruby
Spree::Config.mail_processor_class.register!
```

instead of

```ruby
Spree::Events::Processors::MailProcessor.register!
```

into `core/engine.rb`

core/lib/spree/events/processors/mail_processor.rb
```diff
@@ -0,0 +1,42 @@
+module Spree
+  module Events
+    module Processors
+      class MailProcessor
+        class << self
+          def register!
+            Spree::Config.event_bus.subscribe(:order_confirmed,           ->(event) { send_confirm_email(event) })
```
kennyadsl: not sure about this, but maybe `method(:send_confirm_email)` is the same here.

core/lib/spree/events/processors/mail_processor.rb
cbrunsdon: I think the idea is to eventually pull this mail processor out entirely, and move it to an solidus_action_mailer extension, eventually, so a Config for mail_processor_class doesn't need to be supported by core at all.

core/app/models/spree/order.rb
cbrunsdon: I think we want to ensure we *have* a strong API, so anyone that wishes to count on these events through their app or extensions can expect events to look a certain way.
If someone wanted to write their own event class (that, for instance, supported more things), they could catch this one and re-broadcast a different one, or override the code that triggers the event.

core/lib/spree/events/processors/mail_processor.rb
derrickp: The only time you'd need to modify this list specifically was if you wanted to change the `MailProcessor` list, which as @cbrunsdon mentioned eventually might be pulled out entirely. 

However, if all you wanted to do was subscribe to the same event, any other object can call `Spree::Config.event_bus.subscribe(:order_cancelled, OTHER_PROC)` and it will not change the existing subscriptions. You can have any number of subscribers to an event, so a store can fairly easily call that function to add themselves to the list if they had some functionality they wanted to made sure happened in response to the event. The only concern then is that they call `subscribe` early enough in the startup of the app so as to be able to get the events when they happen.

aaronlifton2:
I really like this idea - the only problems I see are not being able to easily modify or delete the default spree subscriptions.

fylooi:
This is a very extensible approach, but I feel that care needs to be taken to maintain accessibility for Solidus beginners and/or stores which don't need to do much customization. 

My experience replacing the Order mailers with a Mailchimp service call as a first time Solidus user: 
- Figuring out where the mailer gets triggered was a challenge. 
- Had to follow the order flow through `solidus_frontend` into `solidus_core` through `Order` and `Checkout. 
- Monkey patched `Spree::OrderMailer.confirm_email` using an `OrderMailerDecorator`. 

With this approach, I would need to understand the event bus design pattern & implementation in addition to Solidus's core modeling and checkout structure. Some documentation in the code would go a long way.

swcraig:
Thank you for the feedback @fylooi and @aaronlifton2.

Would the sort of documentation I've added [`Spree::EventBus`](https://github.com/swcraig/solidus/blob/56e9d1fc3f4f464e9f437b4fcbb391321cf9b9f4/core/lib/spree/event_bus.rb#L2-L17) and [`Spree::MailProcessor`](https://github.com/swcraig/solidus/blob/56e9d1fc3f4f464e9f437b4fcbb391321cf9b9f4/core/lib/spree/events/processors/mail_processor.rb#L6-L17) be of value?

core/app/models/spree/order_shipping.rb
```diff
@@ -71,17 +71,12 @@ def ship(inventory_units:, stock_location:, address:, shipping_method:,
       shipment.update_columns(state: 'shipped', shipped_at: Time.current)
     end
 
-    send_shipment_emails(carton) if stock_location.fulfillable? && !suppress_mailer # e.g. digital gift cards that aren't actually shipped
+    Spree.event_bus.publish(
+      :carton_shipped,
+      Spree::Events::CartonShippedEvent.new(carton: carton, suppress_mailer: suppress_mailer)
```
tvdeyen: I don't like our event depend on `suppress_mailer` here. Can't we just pass the `order` instead?

core/lib/spree/events/processors/mail_processor.rb
```diff
@@ -0,0 +1,54 @@
...
+    module Processors
+      class MailProcessor
+        class << self
+          # This subscribes the MailProcessor to the relevant events in Solidus.
+          # It is called in an initializer BEFORE events start happening.
+          #
+          # To override/modify which notifications are sent, add a decorator for
+          # this file to define your own `#register!` method. For example:
+          #
+          # Spree::Events::Processors::MailProcessor.class_eval do
```
tvdeyen: If we plan to move this class into an action mailer extension anyway and people are able to write their own mail processors we do not need to imply `class_eval` is a good idea to use the event bus. This violates the whole purpose of an event bus, no?

core/lib/spree/events/processors/mail_processor.rb
tvdeyen: I am in favour of removing the class out of core and move this into an action mailer extension instead. Then provide an easy (as in not `class_eval`) configuration for people wanting to enable/disable certain event handlers. But this can be discussed in that new extension instead of core.

core/lib/spree/events/processors/mail_processor.rb
```diff
@@ -0,0 +1,54 @@
...
+          def send_inventory_cancellation_email(event)
+            Spree::OrderMailer.inventory_cancellation_email(event.order, event.inventory_units).deliver_later
+          end
+
+          def send_reimbursement_email(event)
+            Spree::ReimbursementMailer.reimbursement_email(event.reimbursement.id).deliver_later
+          end
+
+          def send_carton_shipped_emails(event)
+            return if event.suppress_mailer
```
tvdeyen: If we pass the `order` instead of `surpress_mailer`

```rb
return if event.order.suppress_mailer
```

core/lib/spree/app_configuration.rb
```diff
@@ -389,6 +389,10 @@ def stock
       @stock_configuration ||= Spree::Core::StockConfiguration.new
     end
 
+    def event_bus
+      @event_bus_class ||= Spree::EventBus.instance
+    end
```
tvdeyen: I don't think this should live in `Spree::Config`. The class is not configurable, but the name implies it is.

Defining this in `Spree.event_bus` should be sufficient.

core/lib/spree/core.rb
```diff
@@ -42,6 +42,11 @@ def self.config(&_block)
     yield(Spree::Config)
   end
 
+  # Convenience method to access the configured event bus
+  def self.event_bus
+    Spree::Config.event_bus
```
tvdeyen: Why use the `Spree::Config` namespace? Could we use

```rb
def self.event_bus
  @event_bus_class ||= Spree::EventBus.instance
end
```

instead?

core/app/models/spree/order_shipping.rb
mamhoff: Or actually pass the `carton` - it holds a reference to the order(s).

tvdeyen: It is already passing the `carton` and as the carton only nows it's `orders` we do not know the `order` object that sends this event, though.

swcraig: I agree this is not ideal. I've removed `suppress_mailer` from the event object, and have made a (possibly presumptuous) decision within the `MailProcessor`:
```
return if event.carton.shipments.any?(&:suppress_mailer)
```
The `suppress_mailer` flag is defined on `Shipment`. Having the carton and order in the event does not give the necessary context of which _shipment_ just triggered this `CartonShippedEvent`.

Thoughts?

tvdeyen: Do not send the event if suppress_mailer is true? This is it’s whole purpose right? “Do not send” as in “do not notify the user”, whatever the method of notification is.

swcraig: That works for customer-related notifications, but there could possibly be other processors that don't care about the `suppress_mailer` flag. For example, a processor that sends event information to a service like statsd.

core/app/models/spree/reimbursement.rb
```diff
@@ -51,10 +51,12 @@ class IncompleteReimbursementError < StandardError; end
     # These are called if the call to "reimburse!" succeeds.
     class_attribute :reimbursement_success_hooks
     self.reimbursement_success_hooks = []
+    deprecate :reimbursement_success_hooks, 'Add a processor for the :reimbursement_processed event instead.', deprecator: Spree::Deprecation
```
jhawthorn: Deprecating implies that this will be removed in a future version, but as far as I can tell the changes later in this file have made this entirely non-functional.

core/app/models/spree/order_shipping.rb
tvdeyen: Yeah. You are right we should not suppress events ever. But then we need a way for stores to be able to alter already registered event handlers.

core/lib/spree/event_bus.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def publish(name, event)
+      @subscriptions[name].each do |subscription_proc|
+        subscription_proc.call(event)
+      end
+    end
+
+    def subscribe(name, proc_to_call)
+      subscription = @subscriptions[name]
+      already_subbed = subscription.any? { |sub_proc| sub_proc == proc_to_call }
```
jhawthorn: Could `include?` work here?

jhawthorn:
I like a lot of where this is going.

My big concern is still: at what time and in what context do the event callbacks run?

The current implementation has a lot of behaviour implied:
* Callbacks might run inside a transaction from the "publishing" code
  * An exception in the callback will roll back the transaction :bomb:
  * The transaction might rolled back after the callback is run (either by a later callback or by containing code) :bomb:
* An exception in a callback prevents all further callbacks
* Callbacks are (maybe?) run within a request
  * `I18n.locale` will be set correctly for the user :+1:
  * An exception in the callback will cause the request to return an error respones (500, 404, etc. Depending on exception)
* The callback is passed a live AR object
  * there may be unsaved changes from the "publishing" code (we could be careful to avoid this)
  * unsaved attribute changes from the callback may be saved by the containing code :bomb:
    * or a subsequent callback :bomb: 
  * There may be stale associations (we've worked hard to fix associations becoming stale, but it can probably still happen, especially in areas involving inventory_units)

Since we are currently only using this for emails, these aren't issues, thanks to ActiveJob, but as soon as we intend these extensions to be used by users or extensions, some behaviours here are going to cause problems and others are going to be relied upon.

I don't know the best solution, but I think we should consider running all callbacks through ActiveJob, which solves most of these problems.

fastjames:
Every time I read this PR I get a little bit giddy, as it would make the job of extension authors (like me) much easier.

@jhawthorn I understand what you are saying about a callback not getting a consistent environment.
I would be fine with events always running through ActiveJob as I think that enforces the right kind of decoupling. If someone then wanted a synchronous event, it feels like "publish event and wait for some sort of resolution" could be built as a feature atop the first part.

I would love to talk more about this feature and what, if anything, we can do to help it see the light of day.

swcraig:
Thank you for the encouraging words @fastjames.

I've gone ahead and stopped sending live AR objects to the processors. I have also moved the publishing of events to always be _after_ our transactions have completed. Thank you @jhawthorn for the feedback.

I think that it should be up to specific processors to make the decision whether to run their code in-process or through ActiveJob. This PR is not intended to implement an asynchronous message queue. The event bus (we are very open to changing this nomenclature) provides a better mechanism for stores to add their own custom callbacks to run given the firing of an event. In my opinion, if one of my callbacks (subscribers) throws an exception, I _want_ my system to blow up. I don't want my application to continue execution in a possibly compromised state. If I want to write a subscriber that runs asynchronously, or does aggressive exception handling, I am still free to do that.

This implementation does not stop someone from writing asynchronous event subscribers, and I think defaulting to ActiveJob makes it more difficult to write in-process handlers, rather than vice-versa.

If this still isn't what we are going for, or we have misunderstood the concerns about this, we are very open to continue to make changes and get this into a merge-able state.

askl56:
Is there any update on this being merged/where it fits into the roadmap?

swcraig:
This is good for a merge when we are ready.

fastjames:
Still burning a :candle: for this one. To blend with v2.7 it needs to be modified slightly, pointing to `Spree::Config.order_mailer_class` and `Spree::Config.reimbursement_mailer_class` rather than calling the classes directly.

swcraig:
Thanks for pointing out the stale code @fastjames. I think the community is still unsure whether they want to see this change in Solidus. Once we get the green light I will rebase and update the branch with the changes.

dhonig:
I'd like to see this merged.

I'd also like to set it up somewhere people can kick the tires and observe
it in action.
I also think that the Hierarchy around MailProcessor could be cleaned up a
bit to make it a more clear design.
For example adding a "BaseProcessor" would be a simple improvement.  Maybe
this is there and I've missed it?

core/app/models/spree/order.rb
```diff
@@ -884,12 +881,11 @@ def after_cancel
       payments.completed.each { |payment| payment.cancel! unless payment.fully_refunded? }
       payments.store_credits.pending.each(&:void_transaction!)
 
-      send_cancel_email
```
fastjames: I ran into a situation recently where the removal of this method caused problems with an extension that tried to override it. Would it be possible to leave the `send_foo_email` methods in place and have them call the publish method instead? That would allow extension authors to update their tricks before things stop working.

kennyadsl:
Closing in favor of #3081, which has been now merged, thanks a lot for this work anyway!

PR #2433, opened 2017-12-08
Getting started documentation

benjaminwil:
I rewrote the getting started documentation to focus on first-time installations. Other "getting started"-related information can reside in this subdirectory alongside this article.

The article is opinionated in that it focuses on macOS users. The rationale for this is that a) Linux users are more likely to know how to install the required system dependencies and don't need that hand-holding. We also don't shell commands for every package manager out there. b) Windows users have to jump through additional hoops to get Solidus up and running. (We don't officially support or recommend using Windows right now.)

I split out information about common installation options and developing Solidus into separate articles so that they can blossom into their own full-fledged guides. Much of the information in them right now is borrowed from the README.md.


guides/getting-started/develop-solidus.md
```diff
@@ -0,0 +1,127 @@
...
+## Testing
+
+Solidus uses [RSpec](http://rspec.info/) for testing.
+
+If you intend to submit your work to Solidus as a pull request, it must pass all
+of the Solidus test suites before it is merged. We use CircleCI to run tests on
+all incoming pull requests.
+
+To run the test suites for `solidus_frontend` and `solidus_backend`, you need to
+install [PhantomJS](http://phantomjs.org/) on your system first.
```
tvdeyen: We don't use phantoms anymore.

You need to install [chromedriver](https://sites.google.com/a/chromium.org/chromedriver/home) on your system first.

guides/getting-started/develop-solidus.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+This runs using PostgresQL by default, but it can be overridden by specifying
+`DB=sqlite` or `DB=mysql` in the environment. For example:
+
+```shell
+DB=mysql bash build.sh
+```
+
+Note that this will fail if you have not installed PhantomJS on your system.
```
tvdeyen: chromedriver

guides/getting-started/develop-solidus.md
```diff
@@ -0,0 +1,127 @@
...
+```shell
+DB=postgresql bundle exec rspec
+```
+
+#### Fail fast
+
+You can enable `FAIL_FAST` to stop any test run after the first failure:
+
+```shell
+FAIL_FAST=true bundle exec rspec spec/models/state_spec.rb
```
tvdeyen: We should not promote a custom ENV var, but the rspec `--fail-fast` feature instead. But this is common rspec knowledge, I don't think we need to mention it in our guides at all.

guides/getting-started/installation-options.md
```diff
@@ -0,0 +1,94 @@
+# Installation options
+
+## Bleeding edge Solidus
```
tvdeyen: This headline belongs to the next paragraph and this headline should be something like 

## Stable Solidus

guides/getting-started/develop-solidus.md
```diff
@@ -0,0 +1,127 @@
...
+
+[contributing]: https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md
+[solidus-auth-devise]: https://github.com/solidusio/solidus_auth_devise
+
+## Testing
+
+Solidus uses [RSpec](http://rspec.info/) for testing.
+
+If you intend to submit your work to Solidus as a pull request, it must pass all
+of the Solidus test suites before it is merged. We use CircleCI to run tests on
```
tvdeyen: and you need to provide tests for your bug fix or new feature as well.

guides/getting-started/develop-solidus.md
benjaminwil: Thank you. I will also update this in the README.

guides/getting-started/develop-solidus.md
benjaminwil: I actually took this from the [README](https://github.com/solidusio/solidus#running-an-individual-test-suite):

> You can also enable fail fast in order to stop tests at the first failure
> `FAIL_FAST=true bundle exec rspec spec/models/state_spec.rb`

Would you :+1: my removing this from the README as well?

I would agree that we should remove this if it's common rspec knowledge. Instead I'll just explicitly state to reference the rspec documentation for testing needs.

benjaminwil:
@tvdeyen Thanks for the review. All of the issues are fixed.

PR #2435, opened 2017-12-10
Use backbone for admin product sorting, fix admin stock item UI

aaronlifton:
- Moved taxon js into a folder
- converted product sorting coffeescript to js and made it use a backbone view
- fixed stock item table UI

**Before**
image

**After**
image

**Before**
image

**After**
image


tvdeyen:
### Firefox and multiple stock locations

stock 2 locations firefox

Looks the same on chrome. Using `rowspan` like this is not how we can solve this issue. The table needs to have a nested table for the stock locations. I already invested lots of time into trying to solve this issue and this is the only solution I came up with.

### Firefox with one stock location 

stock firefox

aaronlifton2:
Ahh ok, thanks. Will do.

aaronlifton2:
I thought about this, perhaps using css columns would work better than nested tables here? Unless you already got it working.

aldesantis:
@aaronlifton I noticed you opened #2438 which doesn't contain the stock location UI change. Do you want to close this PR or keep it to continue work on stock locations once #2438 is merged?

PR #2436, opened 2017-12-11
Adding pagination to home_controller.

aosorio1:
This change is needed in Solidus 2.4.0 to add pagination to home_controller (products_controller and taxons_controller already have it).


jhawthorn:
Thanks for the report and PR. I think the change we made to the searcher was not the right thing to do and I have proposed reverting it in #2455

jhawthorn:
Closing in favour of #2455. Thanks

PR #2445, opened 2017-12-12
Remove FAIL_FAST environment variable

jhawthorn:
`fail_fast` can be set just by specifying `rspec --fail-fast`. Previously, our use of the `FAIL_FAST` was overriding this feature, though.

I've removed this from the README because it
a) isn't important - nobody needs this to get started developing
b) is described in rspec's docs along with other options


jhawthorn:
Let's merge #2442 first, and then I will rebase this

PR #2449, opened 2017-12-13
Avoid changing method visibility when deprecating a method

jordan-brough:
`ActiveSupport::Deprecation::MethodWrapper#deprecate_methods` changes method
visibility to `public`.  This seems unintentional and I will submit a PR to
Rails to see if that can be updated.

For the meanwhile, this update changes method visibility back to the original
visibility. It's a bit more convoluted than might be expected because
ActiveSupport uses `Module.prepend` to override the original method, so we can't
just write `private :some_method` because that only affects the original method,
whose visibility has not been changed.

If people think it's worth creating a helper for this let me know and I will do so.

### DIFF
--- a/backend/app/controllers/spree/admin/payment_methods_controller.rb
+++ b/backend/app/controllers/spree/admin/payment_methods_controller.rb
@@ -49,10 +49,9 @@ module Spree
       end
 
       def load_providers
+        Spree::Deprecation.warn('load_providers is deprecated. Please use load_payment_method_types instead.', caller)
         load_payment_method_types
       end
-      deprecate load_providers: :load_payment_method_types, deprecator: Spree::Deprecation
-      instance_method(:load_providers).owner.send(:private, :load_providers)
 
       def load_payment_method_types
         @payment_method_types = Rails.application.config.spree.payment_methods.sort_by(&:name)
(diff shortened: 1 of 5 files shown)


jordan-brough:
I've submitted a Rails PR here: https://github.com/rails/rails/pull/31433 to fix this in ActiveSupport.

gmacdougall:
Thanks for finding this Jordan.

I prefer to leave this as is and update to a new rails version when this is fixed upstream.

jordan-brough:
@gmacdougall I think it's worth going ahead and fixing this.  My fix was only merged into Rails master and Rails v5.2 is already in beta so I imagine it'll be a good while before that trickles back to Solidus.  It's good to have correct method visibility for the apps using Solidus imo, even for deprecated methods.

jhawthorn:
Nice catch.

> My fix was only merged into Rails master and Rails v5.2 is already in beta so I imagine it'll be a good while before that trickles back to Solidus

I _think_ rails master is still tracking 5.2, so it shouldn't be too long.

Alternatively, we could use `Spree::Deprecation.warn` explicitly within the private method, which avoids the slight strangeness of `instance_method(...).owner.send(:private, ...)`. I am fine either way.

jordan-brough:
> we could use `Spree::Deprecation.warn` explicitly within the private method

Great point, I'll update with that 👍

jordan-brough:
@jhawthorn @gmacdougall I removed a broken deprecation no `Spree::PaymentMethod#provider_class` and updated the remaining ones to use `Spree::Deprecation.warn`.  Look good?

PR #2453, opened 2017-12-14
Split builds more intelligently in build-ci.rb

jhawthorn:
When building on CircleCI, we split the subproject builds between several containers in order to finish faster.

Previously we did this by just splitting the projects round-robin. This actually worked out well due to our number of workers (3) and the alphabetization of our projects.

* Worker 1: solidus_api, solidus_frontend
* Worker 2: solidus_backend (rspec), solidus_backend (JS), solidus_sample
* Worker 3: solidus_core

core and backend's specs take about the same amount of time to run, so this worked fairly okay because backend's JS tests and solidus_sample are both pretty quick. If we were to add/remove/rename a subproject, rename an existing one, or consider backend's JS tests separately from rspec (the change which motivated this), we would be less lucky and our CI would take longer.

This change splits backend JS from backend rspec, and gives each project a weight in build-ci.rb.

With this change, our CI runs:

* Worker 1: solidus_core
* Worker 2: solidus_backend (rspec)
* Worker 3: solidus_frontend, solidus_api, solidus_sample, solidus_backend (JS)

It's not a huge improvement, but it will probably make CI builds 20-30 seconds faster now by giving backend a worker all of its own, and it should stay fast regardless of what changes we make to our project structure in the future.

### DIFF
--- a/build-ci.rb
+++ b/build-ci.rb
@@ -4,7 +4,7 @@
 require 'pathname'
 
 class Project
-  attr_reader :name
+  attr_reader :name, :title, :weight
 
   NODE_TOTAL = Integer(ENV.fetch('CIRCLE_NODE_TOTAL', 1))
   NODE_INDEX = Integer(ENV.fetch('CIRCLE_NODE_INDEX', 0))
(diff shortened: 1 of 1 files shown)


build-ci.rb
```diff
@@ -17,11 +17,21 @@ class Project
...
   end
 
-  ALL = %w[api backend core frontend sample].map(&method(:new)).freeze
+  ALL = [
+    new('api', weight: 50),
+    new('backend', weight: 215),
+    new('backend', test_type: :teaspoon, title: "backend JS", weight: 15),
+    new('core', weight: 220),
+    new('frontend', weight: 95),
+    new('sample', weight: 22)
```
kennyadsl: curiosity: how did you find/choose these weights?

jhawthorn: Just basically `time rspec` on my local machine. Not very accurate (my computer is a lot faster than circle), but it just needs to be an estimate

PR #2454, opened 2017-12-14
Use scope.cache_key in cache_key_for_products

jhawthorn:
This was written before rails supported calling `cache_key` on a scope. This saves us one query, since rails selects both the max and count in a single query.

### DIFF
--- a/core/app/helpers/spree/products_helper.rb
+++ b/core/app/helpers/spree/products_helper.rb
@@ -66,9 +66,7 @@ module Spree
 
     # @return [String] a cache invalidation key for products
     def cache_key_for_products
-      count = @products.count
-      max_updated_at = (@products.maximum(:updated_at) || Date.today).to_s(:number)
-      "#{I18n.locale}/#{current_pricing_options.cache_key}/spree/products/all-#{params[:page]}-#{max_updated_at}-#{count}"
+      "#{I18n.locale}/#{current_pricing_options.cache_key}/spree/products/all-#{params[:page]}-#{@products.cache_key}"
     end
   end
 end
--- a/core/lib/spree/testing_support/caching.rb
+++ b/core/lib/spree/testing_support/caching.rb
@@ -7,16 +7,6 @@ module Spree
         @cache_write_events
       end
 
-      def assert_written_to_cache(key)
-        unless @cache_write_events.detect { |event| event[:key].starts_with?(key) }
-          fail %{Expected to find #{key} in the cache, but didn't.
-
-  Cache writes:
-  #{@cache_write_events.join("\n")}
-          }
-        end
-      end
-
       def clear_cache_events
         @cache_write_events = []
       end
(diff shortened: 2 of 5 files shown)


gmacdougall:
I like the motivation behind these changes but there's a couple failing tests which need to be addressed.

kennyadsl:
I've found something about those spec failures:

Using `ActiveRecord::Relation#cache_key` produce cache keys for our products search similar to

```
query-9a8681e154ac694950cb63b934444259-2-20180328191149899341/695d3742aae7eec904440fe5dfbf1300
```


- `9a8681e154ac694950cb63b934444259` is the digest (MD5) of the query so it changes every time the query changes
- `2` is the size of the collection
- `20180328191149899341/695d3742aae7eec904440fe5dfbf1300` is timestamp of the most recently updated record in the collection.

Specs fail because **the first part is different in each query to retrieve products**. This is the content of `cache_writes` when specs fail:

```text
{:controller=>"home", :action=>"index", :key=>"views/en/USD/spree/products/all--spree/products/query-9a8681e154ac694950cb63b934444259-2-20180328191149899341/695d3742aae7eec904440fe5dfbf1300"}, 
{:controller=>"home", :action=>"index", :key=>"views/en/USD/spree/products/all--spree/products/query-c8672fcf8f2cb4d058f7f3f08a9dace2-2-20180328191149899341/695d3742aae7eec904440fe5dfbf1300"}
```

I looked in depth at both queries and it looks like they are actually different. The difference is into this piece:

```text
("spree_products".available_on <= '2018-03-30 08:43:59.215588') 
```

which is generated by the [available scope](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/product/scopes.rb#L182) that we use in our [searcher base scope](https://github.com/solidusio/solidus/blob/a779d170d3c4224fb6aa3f6551bf061374d9a6f0/core/lib/spree/core/search/base.rb#L56). That query uses `Time.current` so ActiveRecord is right thinking that he needs to generate the query again. 

Actually it's something we are not taking into account when generating the cache key for products on Solidus. If a product "expires" it will be visibile in the products list until the cache key expires.

I'll continue to think at a solution for this. Any suggestion is welcome!

kennyadsl:
Overriding `Spree::Product.collection_cache_key` fix specs. But we loose the [query signature](https://github.com/rails/rails/blob/c680080967c8941e8a7a6eb44ee3d9e697dd3c91/activerecord/lib/active_record/collection_cache_key.rb#L6) functionality provided by ActiveRecord. Anyway for that kind of query I think we cannot use the query digest since it contains the `Time.now` part and it's different every time.

Something like this lets the specs pass:

```ruby
# core/app/models/spree/product.rb

def self.collection_cache_key(collection = all, timestamp_column = :updated_at)
  count = collection.count
  max_updated_at = (collection.maximum(:updated_at) || Date.today).to_s(:number)
  "#{collection.model_name.cache_key}/#{count}-#{max_updated_at}"
end
```

jhawthorn:
The best way to handle this would probably be to decide a maximum resolution of `available_on` (say, 5 minutes) and round times and cache based on that.

I will not get around to this.

PR #2457, opened 2017-12-15
Support rspec-rails 3.7

jhawthorn:
We previously had errors on rspec-rails 3.7 due to not having puma, but that is no longer an issue.


Gemfile
```diff
@@ -17,7 +17,7 @@ end
 gem 'database_cleaner', '~> 1.3', require: false
 gem 'factory_bot_rails', '~> 4.8', require: false
 gem 'rspec-activemodel-mocks', '~>1.0.2', require: false
-gem 'rspec-rails', '~> 3.6.0', require: false
+gem 'rspec-rails', '~> 3.7.1', require: false
```
tvdeyen: Shouldn’t we able to allow `~> 3.7`?

PR #2458, opened 2017-12-15
Remove broken `Shipment.has_many :cartons` relationship

jordan-brough:
This has been broken since Solidus 2.3 (Rails 5.1).

Solidus 2.2 gave this warning:

> DEPRECATION WARNING: uniq is deprecated and will be removed from Rails 5.1 (use distinct instead) (called from block in <class:Shipment>

No one is using it apparently :) and I think it's probably best not to declare
this as a `has_many` anyway, for the sake of simplifying the in-memory object
situation for orders/shipments/inventory units/cartons.

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -10,7 +10,6 @@ module Spree
     has_many :shipping_rates, -> { order(:cost) }, dependent: :destroy, inverse_of: :shipment
     has_many :shipping_methods, through: :shipping_rates
     has_many :state_changes, as: :stateful
-    has_many :cartons, -> { uniq }, through: :inventory_units
 
     before_validation :set_cost_zero_when_nil
 


loicginoux:
just to let you know, we use this relation in our code, but overwrote this association like this:
```
Spree::Shipment.has_many :cartons, -> { distinct }, through: :inventory_units
```

as suggested in rails release note 5.0 http://edgeguides.rubyonrails.org/5_0_release_notes.html#active-record-deprecations

jhawthorn:
@loicginoux that looks good to me, would you mind sending a PR. What do you think @jordan-brough

jordan-brough:
Well the `distinct` change is what's recommended in the deprecation warning ("(use distinct instead)") but I'm just a little sad that we get separate `Carton` instances for all of the following:

```ruby
shipment.order.cartons
shipment.inventory_units.map(&:carton)
shipment.cartons
```

I was thinking it would be nice to eliminate this path for now since it's already broken.
I don't really have a great idea for a better replacement at the moment though so I can change the approach if we'd prefer.

jhawthorn:
Unfortunately most things with inventory_units end up with us having different copies of the same record in memory.

If we want to remove this, we should instead move the deprecation onto `cartons`, and then remove it in the next major version. The previous deprecation only warns users that we (in solidus_core) are doing something deprecated, not that they are.

jordan-brough:
Updated to fix the `has_many` rather than deleting it.

PR #2461, opened 2017-12-18
New preference to control available currencies

softr8:
This change allows stores to control what currencies will be available, it could improve usability a bit in some dropdowns by limiting the number of options.

### DIFF
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -257,6 +257,15 @@ module Spree
     #   @return [Boolean] Indicates if stock management can be restricted by location
     preference :can_restrict_stock_management, :boolean, default: false
 
+    # Allows restricting what currencies will be available.
+    #
+    # @!attribute [r] available_currencies
+    #   @returns [Array] An array of available currencies from Money::Currency.all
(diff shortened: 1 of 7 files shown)


core/lib/spree/app_configuration.rb
```diff
@@ -257,6 +257,12 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Boolean] Indicates if stock management can be restricted by location
     preference :can_restrict_stock_management, :boolean, default: false
 
+    # Allows restricting what currencies will be available.
+    #
+    # @!attribute [r] available_currencies
+    #   @returns [Array] An array of available currencies from Money::Currency.all
+    preference :available_currencies, :array, default: ::Money::Currency.all
```
jhawthorn: Only concern is that this should _maybe_ be a proc. We load preferences quite early and customizations might be made to `::Money`. (this is really quite an edge case, so I don't know if it's necessary, but can't hurt)

jhawthorn:
This seems to be failing because `money` isn't loaded yet when `app_configuration` is

jhawthorn:
Thank you!

PR #2467, opened 2017-12-21
Initial promotions documentation

benjaminwil:
I ported the Spree promotions guide, rewrote, and expanded it to be a number of articles.

I have not yet mentioned all of the aspects of the promotions system. For example, I didn't even acknowledge that `Spree::PromotionCode` exists. But what's in this PR is a good starting point. Future promotions documentation PRs will be smaller.

There is some repetition between the `overview.md` and other articles, but I think introducing major concepts in an overview article is helpful.

I would appreciate any feedback you have.


guides/promotions/promotion-rules.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Available promotion rule types
+
+The following classes are [subclasses of the `Spree::Promotion::Rules`
+model][promotion-rules]:
+
+- `FirstOrder`: Eligible for a customer's first order only.
+- `FirstRepeatPurchaseSince`: Eligible for a customer's first repeat purchase
+  since a specified date.
+- `NthOrder`: Eligible for a customer's *n*th order only.
```
benjaminwil: The italics at `*n*th order` here messes up the syntax highlighting for the rest of the article, but GitHub renders the Markdown fine, and my text editor's syntax highlighting is fine. Weird?

guides/promotions/overview.md
```diff
@@ -0,0 +1,195 @@
...
+  for the promotion to be eligible. When set to `any`, just one of the
+  [promotion rules](#promotion-rules) must be met.
+- `path`: If the promotion is activated when the customer visits a URL, this
+  value is the path for the URL.
+- `per_code_usage_limit`: Specifies how many codes can be used before the
+  promotion becomes inactive.
+- `apply_automatically`: If `true`, the promotion is activated and applied
+  automatically once all of the [eligibility checks](#eligibility) have passed.
+
+Note that you can access promotion information using the `promotion` method its
```
swcraig: Was the word "on" missed after the word "method"?

guides/promotions/overview.md
```diff
@@ -0,0 +1,195 @@
...
+1. The administrator creates a `Spree::Promotion` from the Solidus backend.
+  - They create a name, description, and optional category for the promotion.
+  - They choose not to set a usage limit for the promotion.
+  - They choose not to set a start or end date for the promotion.
+  - They choose the "Apply to all orders" activation method. Alternatively, they
+    could have chosen to apply promotions via a promotion code or a URL.
+2. The administrator creates `Spree::PromotionRule`s for the promotion.
+   - In this case, they use the rule type "Item Total"
+     (`Spree::Promotion::Rules::ItemTotal`) and set the rule so that the
+     promotion must be greater than $100 USD.
```
swcraig: Do you mean the order must be greater than $100 USD?

guides/promotions/overview.md
```diff
@@ -0,0 +1,195 @@
...
+    `Spree::Promotion::Actions::Shipping` model. In this case, the only
+    available action is "Makes all shipments for the order free".
+  - Because the promotion action requires a shipment, the
+    `Spree::PromotionHandler::Shipping` will be used when it is time to activate
+    the promotion.
+
+Different types of promotions would change the customer's experience of
+promotion activation. For example, the customer might be required to enter a
+promotion code to activate some promotions, while a free shipping promotion
+would be applied automatically.
```
adammathys: > while a free shipping promotion would be applied automatically.

That isn't always the case. You could have free shipping that only applies after a customer enters a code or visits a specific page.

guides/promotions/promotion-actions.md
```diff
@@ -0,0 +1,71 @@
...
+model][promotion-actions]:
+
+- `CreateAdjustment`: Creates a single adjustment for the current
+  `Spree::Order`.
+- `CreateItemAdjustments`: Creates an adjustment for each `Spree::LineItem` in
+  the current order.
+- `CreateQuantityAdjustments`: Creates per-quantity adjustments. For example,
+  you could create an action that gives customers a discount on the first three
+  t-shirts that they order.
+- `FreeShipping`: Creates an adjustment that subtracts all shipping charges.
```
adammathys: I would use "negates" instead of "subtracts", but that might just be personal preference.

guides/promotions/promotion-actions.md
```diff
@@ -0,0 +1,71 @@
...
+  Similarly, we should link to the adjustments documentation once it's merged.
+-->
+
+## Available promotion action types
+
+The following classes are [subclasses of the `Spree::Promotion::Actions`
+model][promotion-actions]:
+
+- `CreateAdjustment`: Creates a single adjustment for the current
+  `Spree::Order`.
```
adammathys: Creates a single adjustment associated to the order. I might also add something about not recommending its use. Order-level adjustments don't work particularly well with taxes or refunds.

guides/promotions/promotion-actions.md
```diff
@@ -0,0 +1,71 @@
...
+- `CreateItemAdjustments`: Creates an adjustment for each `Spree::LineItem` in
+  the current order.
```
adammathys: Will create adjustments on each **applicable** line item. (e.g.: If there's a product rule, it will only apply to products matching the rule.)

guides/promotions/promotion-actions.md
```diff
@@ -0,0 +1,71 @@
...
+- `CreateQuantityAdjustments`: Creates per-quantity adjustments. For example,
+  you could create an action that gives customers a discount on the first three
+  t-shirts that they order.
```
adammathys: Creates adjustments based on groups of items, not just the first set of matching items. So the action would apply based on the number of groups of t-shirts.

e.g.: 0-2 shirts - doesn't apply, 3-5 shirts - applies once, 6-8 shirts - applies twice, etc.

guides/promotions/promotion-actions.md
```diff
@@ -0,0 +1,71 @@
...
+Note that whenever an order, line item, or shipment with a promotion adjustment
+on it is updated, the [eligibility][eligibility] of the promotion is re-checked
+and the promotion actions are re-applied.
+
+[eligibility]: overview.md#eligibility
+
+## Register a custom promotion action
+
+You can create a new promotion action for Solidus by creating a new class that
+inherits from `Spree::PromotionAction`:
```
adammathys: I might also mention that it must implement the `perform(options = {})` method that should return a boolean declaring whether the action was applied successfully. 

Also, that it's recommended that it define a `remove_from(order)` as well.

guides/promotions/promotion-rules.md
```diff
@@ -0,0 +1,111 @@
...
+- `ItemTotal`: Eligible if the order total is less than or greater than a
```
adammathys: I would try to be more specific that it's the order's item total. (i.e.: Before taxes/shipping/other discounts/etc.)

guides/promotions/promotion-rules.md
```diff
@@ -0,0 +1,111 @@
...
+  Once that documentation is merged, we can add those link.
+-->
+
+[promotion-rules]: https://github.com/solidusio/solidus/tree/master/core/app/models/spree/promotion/rules
+
+## Eligibility
+
+Note that whenever an order, line item, or shipment with a promotion adjustment
+on it is updated, the [eligibility][eligibility] of the promotion is re-checked
+and the promotion actions are re-applied.
```
adammathys: > the promotion actions are re-applied.

Might be a bit misleading. The action isn't really re-applied. The amount will be re-calculated and the eligibility of the adjustment will be changed based on the eligibility and if the amount becomes $0. But I'm pretty sure all of the logic in the action's `perform(options = {})` method won't be re-run.

benjaminwil:
@adammathys Thanks for your review. I've addressed your comments. If you could re-review whenever you've got a chance that'd be great. 👍

PR #2472, opened 2017-12-22
Make the install generator idempotent

tvdeyen:
Currently if you run the install generator a second time the

    mount Spree::Core::Engine

route gets injected a second time.

Thor's insert_into_file should be an reversible operation but
I guess because we inject multiple lines into the routes file
it doesn't work as expected.

### DIFF
--- a/core/lib/generators/spree/install/install_generator.rb
+++ b/core/lib/generators/spree/install/install_generator.rb
@@ -4,6 +4,8 @@ require 'bundler/cli'
 
 module Spree
   class InstallGenerator < Rails::Generators::Base
+    CORE_MOUNT_ROUTE = "mount Spree::Core::Engine"
+
     class_option :migrate, type: :boolean, default: true, banner: 'Run Solidus migrations'
     class_option :seed, type: :boolean, default: true, banner: 'load seed data (migrations must be run)'
     class_option :sample, type: :boolean, default: true, banner: 'load sample data (migrations must be run)'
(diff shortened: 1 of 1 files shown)


core/lib/generators/spree/install/install_generator.rb
```diff
@@ -4,6 +4,8 @@
 
 module Spree
   class InstallGenerator < Rails::Generators::Base
+    SPREE_MOUNT_ROUTE = "mount Spree::Core::Engine"
```
jhawthorn: This is already namespaced within `Spree::`, so I'm not sure that's descriptive. It could be `MOUNT_ROUTE` or `ENGINE_MOUNT_ROUTE` or `CORE_MOUNT_ROUTE`

core/lib/generators/spree/install/install_generator.rb
```diff
@@ -149,23 +151,26 @@ def load_sample_data
     end
 
     def install_routes
-      insert_into_file File.join('config', 'routes.rb'), after: "Rails.application.routes.draw do\n" do
-        <<-ROUTES
+      routes_file_path = File.join('config', 'routes.rb')
+      unless File.read(routes_file_path).match Regexp.new(SPREE_MOUNT_ROUTE)
```
jhawthorn: Treating `SPREE_MOUNT_ROUTE` as a regex is going to work fine because it doesn't have any regex characters (`*`, `(`, `.`, etc), but it feels like the wrong thing to do.

This should probably be

    File.read(routes_file_path).include?(SPREE_MOUNT_ROUTE)

tvdeyen: Check

core/lib/generators/spree/install/install_generator.rb
tvdeyen: changed

tvdeyen:
@jhawthorn thanks. Addressed your concerns.

Issue #2469, opened 2017-12-21
Replace jQuery UI components with something more modern


jhawthorn:
We use only a few jQuery UI components in the admin. jQuery UI is a bit dated and is no longer being regularly updated (but we are on an older version even so). The few places we do use it do not fit into our UI well, even though some have received some customization.

There likely exist better, more modular, purpose built libraries that can replace the widgets we are using.

Our current use of jQuery UI probably weighs about 100k of JS (minified), 20k of CSS (minified), and a half-dozen small images. This isn't the main motivation to remove it, but we can likely do better.

* [ ] Sortable #2523
* [x] Datepicker #2477, #2506, #2510
* [ ] Autocomplete #2509

## Sortable

Probably the hardest to replace, this is used in two different ways:
* Allowing admins to sort items in tables
  * [generically on all tables with class "sortable"]( 
https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/admin.js#L148-L194)
  * [sorting products within taxons](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/taxons.js.coffee)
* [Arranging taxons into a tree](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/taxonomy.js.coffee) (which is a bit more demanding and complicated)

Potential alternatives:
* https://github.com/RubaXa/Sortable
* https://johnny.github.io/jquery-sortable/
* https://github.com/lukasoppermann/html5sortable
* (many more)

## Datepicker

We use the datepicker in a few places, both to represent a date range (a start and end date), and to represent a single date. Ideally we would avoid including Moment.js as part of this (which is a great library, but huge).

Potential alternatives:
* [`<input type="date">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) (browser support is too patchy to use this without polyfills or something for IE11)
* ✔️ https://github.com/chmln/flatpickr (suggested by mdesantis)
* https://github.com/dbushell/Pikaday (looks quite good)
* http://bootstrap-datepicker.readthedocs.io/en/latest/
* https://fengyuanchen.github.io/datepicker/

## autocomplete

We use autocomplete only in one place: filling in product properties. It uses ajax to fetch the possible property names.

Potential alternatives:
* Nothing? we might not need this functionality
* ✔️ HTML5 datalist (browser support is [patchy](https://caniuse.com/datalist) but maybe we are happy with progressive enhancement)
[...]

PR #2477, opened 2018-01-02
Switch from jQuery UI datepicker to flatpickr

mdesantis:
I reached a good point about Flatpickr integration, which was discussed in [this issue](https://github.com/solidusio/solidus/issues/2469). It works and it seems good to me. Could you give it a look? 

Also, I have some considerations:

1. class names: I managed to leave HTML untouched, so I left `datepicker` class name. It seems to me more consistent as it defines the functionality, but if you prefer to use the component name it could be replaced with `flatpickr`.
1. default options: the previous datepicker instance had some options, mainly for l10n, but flatpickr has different options so I preferred to not to mess with them. 
1. theme: flatpickr has some themes, I left the default theme which seems to me the best match with solidus admin; anyway a custom theme could be developed
1. input: a good idea could be switching from `<input type="text">` to `<input type="date">` for dates, to `<input type="datetime-local">` for datetime and to `<input type="time">` for time. They don't hurt browsers if they are not supported (they fallback to `"text"`) and allow polyfill using `$('input[type=...]'")`, using flatpickr `enableTime` option
1. range behaviour: at the moment, dates range support consists in setting min and max dates on value change. I found a better UX in flatpickr range plugin, which reorders the dates on value change. Unfortunately flatpickr range plugin is in beta, but I managed to emulate the same behaviour with two flatpickr instances. What do you think about it?
1. vendor vs yarn: I noticed solidus still uses vendor folder approach for thrid party libraries, so I put flatpickr files in vendor, which is pretty troublesome, since you have to: find and download the dist files (which more and more project do not include in GitHub repos), divide them for type, and move them to the right folders (javascript, stylesheets, images). I think yarn would make third party libraries usage far easier.


backend/app/assets/javascripts/spree/backend.js
```diff
@@ -12,6 +12,8 @@
 //= require solidus_admin/backbone-nested-models
 //= require solidus_admin/popover
 //= require solidus_admin/bootstrap
+//= require solidus_admin/flatpickr/flatpickr
+//= require solidus_admin/flatpickr/plugins/rangePlugin
```
kennyadsl: is loading this plugin needed?

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -53,24 +53,48 @@ $.fn.radioControlsVisibilityOfElement = function(dependentElementSelector){
...
 
-  // Correctly display range dates
-  $('.date-range-filter .datepicker-from').datepicker('option', 'onSelect', function(selectedDate) {
-    $(".date-range-filter .datepicker-to" ).datepicker( "option", "minDate", selectedDate );
-  });
-  $('.date-range-filter .datepicker-to').datepicker('option', 'onSelect', function(selectedDate) {
-    $(".date-range-filter .datepicker-from" ).datepicker( "option", "maxDate", selectedDate );
-  });
+  // Handle range dates
+  if ($('.date-range-filter .datepicker-from, .date-range-filter .datepicker-to').length) {
```
kennyadsl: What about trying to create a plugin for this behavior?

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -53,24 +53,48 @@ $.fn.radioControlsVisibilityOfElement = function(dependentElementSelector){
...
-  $('.datepicker').datepicker({
-    dateFormat: Spree.t('date_picker.js_format', { default: 'yy/mm/dd' }),
-    dayNames: Spree.t('abbr_day_names'),
-    dayNamesMin: Spree.t('abbr_day_names'),
-    firstDay: parseInt(Spree.t('date_picker.first_day', { default: '0' }), 10),
-    monthNames: Spree.t('month_names'),
-    prevText: Spree.t('previous'),
-    nextText: Spree.t('next'),
-    showOn: "focus"
+  $('.datepicker').flatpickr({
```
kennyadsl: since we are making changes here, what about moving all the datepicker code to a separate file?

backend/app/assets/javascripts/spree/backend.js
mdesantis: No, it is a typo! Nice catch!

backend/app/assets/javascripts/spree/backend/admin.js
mdesantis: I actually already implemented the former behavior, and I saved that code in the case it was decided to preserve that behavior. If requested I can switch to that.

kennyadsl: I meant moving this behavior you implemented into a new `flatpickr` plugin that we'll include instead of putting this code here directly but it's just an idea for better isolate this functionality that we could choose to replace later. Anyway maybe it's better to wait for other core team members opinion on flatpickr before putting too much effort here. Thanks!

mdesantis:
>  maybe we don't need files for themes and plugins we are not going to use. WDYT?

I think it's bad either way. The right way to import an external client side library in Rails nowadays is using yarn, but I didn't detect any use of it in solidus. Copy pasting external libraries is bad, it introduces all the issues that we, as Rails developers, had on server side before bundler was introduced, that is third party libraries not synced with upstream, hardly upgradable (as I pointed out in 6.) and thus rarely updated. Not to mention that sometimes developers feel the right to edit copy pasted third party libraries, making updatings a nightmare.

In order to keep as most as possible third party libraries synced with upstream I usually import all the libraries code, even the unused code, but it's definitely possible to cherry pick the used files. I think the choice depends by the current solidus way to manage third party libraries.

backend/app/assets/javascripts/spree/backend/admin.js
mdesantis: Yes, we could talk together about that in the issue discussion maybe? I don't know if all the points I made above in the pull request comment will be considered :flushed:

mdesantis:
@kennyadsl I moved the datepicker code to a single file, is it ok? :-)

jhawthorn:
This looks great to me :+1:

I hope you don't mind I've taken it to #2477 with a few small changes:
* Rebased (Fixing the build)
* Added localization
* Fixed CRLF line endings
* Removed unused flatpickr files

mdesantis:
Yes, it's fine for me. If you want help just ask ;)

---

Maurizio De Santis

Il 17 gen 2018 12:02 AM, "John Hawthorn" <notifications@github.com> ha
scritto:

jhawthorn:
Thank you! Merged as #2477

PR #2479, opened 2018-01-02
[RFC] Use microsecond timestamps in MySQL

jhawthorn:
MySQL in Rails by default only tracks those datetimes with 1s of precision. This can cause issues since we cache based on that `updated_at` timestamp.

Now that that's somewhat widely supported (MySQL 5.6+/MariaDB 10+) I wonder if we should be specifying that precision in core.

If merged, this migration would be **VERY**  slow.

### DIFF
--- a/core/db/migrate/20180102185108_add_timestamp_precision_to_mysql.rb
+++ b/core/db/migrate/20180102185108_add_timestamp_precision_to_mysql.rb
@@ -1,0 +1,90 @@
+class AddTimestampPrecisionToMysql < ActiveRecord::Migration[5.1]
+  TABLES = %W[
+    spree_addresses
+    spree_adjustment_reasons
+    spree_adjustments
+    spree_assets
+    spree_calculators
+    spree_cartons
+    spree_countries
+    spree_credit_cards
+    spree_customer_returns
+    spree_inventory_units
+    spree_line_item_actions
+    spree_line_items
+    spree_log_entries
+    spree_option_types
+    spree_option_values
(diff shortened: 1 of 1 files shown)


core/db/migrate/20180102185108_add_timestamp_precision_to_mysql.rb
```diff
@@ -0,0 +1,90 @@
...
+    spree_variants
+    spree_zone_members
+    spree_zones
+  ]
+
+  def up
+    return unless Spree::Order.connection.adapter_name == "Mysql2"
+    TABLES.each do |table|
+      change_column(table, :updated_at, :datetime, precision: 6)
+      change_column(table, :created_at, :datetime, precision: 6)
```
tvdeyen: MySQL allows multiple columns in a single `ALTER TABLE` statement. Maybe this will speed up the migration a bit?

bbuchalter:
While I think this migration is necessary to have effective testing, it will require most stores to take some downtime. If the community wants to proceed, perhaps it could be isolated into its own dedicated point release to ensure proper attention is paid.

tvdeyen:
Since migrations are most of time part of deployment strategies I think it is probably not a good idea to add this as migration. Why not make it a rake task so then stores can decide when to deploy this asynchronous year to their automated deployment?

jhawthorn:
We've merged #2598

It would be nice to provide this migration to older stores (maybe as a gist or installed by rake task), but it no longer makes sense in this form.

PR #2480, opened 2018-01-03
adding meta_title to api json product response

loicginoux:
the api response for a product gives the `meta_description` and `meta_keywords` but not the `meta_title`. I don't see why this attribute should not be in the api response.


jhawthorn:
Thank you

PR #2482, opened 2018-01-03
Tweaks to override-solidus-assets.md

jhawthorn:
Fixes a JS typo and warns about view-like overriding of files.

At some point in the future it's possible we will use an asset bundler other than sprockets (like webpack or rollup). If we do so we won't be able to support sprocket's view-like file lookup and overriding (and it is an unusual thing to support for JS in the first place).

### DIFF
--- a/guides/assets/override-solidus-assets.md
+++ b/guides/assets/override-solidus-assets.md
@@ -93,6 +93,9 @@ For example, to replace the `solidus_frontend`'s
 replacement to `your_app/app/assets/stylesheets/spree/frontend/_variables.scss`
 with your own definitions inside.
 
+This is more brittle than overriding single definitions, as described above,
+and isn't guaranteed to work in future Solidus versions.
+
 Note that this method *completely* replaces any functionality provided by the
 stylesheet or JavaScript file.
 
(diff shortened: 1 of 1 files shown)


guides/assets/override-solidus-assets.md
```diff
@@ -24,7 +24,8 @@ might not be effective for the next.
 
 If you want to tweak a CSS definition or change the way a particular JavaScript
 methods works, you can override it by redefining the definition in a file in
-your application's `app/assets` tree.
+your application's `app/assets` tree, although this isn't guaranteed to work in
+future versions of Solidus.
```
tvdeyen: As far as I read this paragraph it handles overriding a single CSS rule or JS function in a separate file, not overwriting the defining file itself. This is handled below. And overwriting things like this is always supported, no? Or am I misreading something?

jhawthorn: You're right I was confused about what this was describing. We should probably change this section (in a future PR) because "redefining" isn't an accurate description of what happens when adding additional CSS (and is likely misleading for JS too).

benjaminwil: :-1: My bad.

PR #2488, opened 2018-01-05
Split existing configuration guide

benjaminwil:
I've cleaned up and rewritten some of the existing configuration guide. More noticeably, I've split the existing guide up into a number of new articles that now reside in different subdirectories.

- The `getting-started` directory now has a stub about why you might want to create a private Solidus fork.
- A new `extensions` directory houses an article about installing extensions and (for now) an article about decorators (though I think this needs another home).
- A new `views` directory has an article about overriding views.
- A new `preferences` directory has articles about site-wide preferences configuration, class extension points, and creating model-specific preferences.

None of this is entirely new content, but used to all live together in one article. Splitting things up into more focused directories will hopefully help developers find the information they need faster as the Solidus documentation continues to grow.


guides/extensions/installing-extensions.md
```diff
@@ -0,0 +1,44 @@
+# Installing extensions
+
+Solidus extensions are gems that change or extend Solidus's functionality. Many
+Solidus extensions can be installed in a few simple steps: 
+
+1. Add the extension gem to your project's `Gemfile`.
+2. Run `bundle install to install the gem and its dependencies.
```
swcraig: I think you are missing an ending ` for `bundle install`.

guides/preferences/app-configuration.md
```diff
@@ -0,0 +1,70 @@
+# App configuration 
+
+Solidus includes many preferences with default settings that are appropriate for
+typical stores.  For a list of Solidus's preferences their default values, see
```
adammathys: > Solidus's preferences **and** their default values . . .

benjaminwil:
Thanks for the review @adammathys. I've committed changes for all your comments.

If you could approve that would be great. :+1:

PR #2489, opened 2018-01-05
Convert store credits to discard

jhawthorn:


core/app/models/spree/store_credit.rb
```diff
@@ -33,6 +39,7 @@ class Spree::StoreCredit < Spree::PaymentSource
   before_validation :associate_credit_type
   before_validation :validate_category_unchanged, on: :update
   before_destroy :validate_no_amount_used
```
jhawthorn: This existing line I'm pretty sure actually didn't work. Though errors are added to the model, `deleted_at` is still set and persisted.

PR #2490, opened 2018-01-08
creating an order should activate promotions

loicginoux:
- I have a promotion that can apply on a product
- via API, I create an order, adding the product with promotion

Expected: I should have the promotion applied to the order.

Current: no promotion applied


jhawthorn:
I don't know why this originally failed CI. Passing and merged as #2576. Thanks!

PR #2493, opened 2018-01-09
Replace new product inline form with normal form

jhawthorn:
Previously, creating a new product involved filling out a small inline form which held with only the required attributes of a product.

This requires admin users to understand and learn to use two separate forms with the same purpose (filling in product data). It's also awkward if the user wants to fill in some of the information not on this page as they have to create the product and then fill in the rest of the information they cared about.

This commit removes the inline new product form in favour of linking to a new product page with the same form as the edit page.

**Before:**

**After:**

**Todo:**
* [x] ~Specifying slug doesn't work on a new product~ It does work, but is cleared if validations fail due to a bug in friendly_id, which I will file there
* [x] Slug should not be required on new records (it is generated automatically)
* [x] Missing breadcrumbs
* [x] ~Should we add the tabs? Disabled?~ Can't render. We can revisit if this is important
* [x] Fix specs?

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -14,7 +14,7 @@
       <div data-hook="admin_product_form_slug">
         <%= f.field_container :slug do %>
           <%= f.label :slug, class: ('required' if !f.object.new_record?) %>
-          <%= f.text_field :slug, class: 'fullwidth title', required: !f.object.new_record? %>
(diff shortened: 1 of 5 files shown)


jhawthorn:
A compromise required to make this work: the slug field needs to be disabled when creating a new product in order to keep the behaviour we had previously (slug generated on create, slug required when updating).

I think this is a reasonable solution and we can address it later if there is a preferred UX design.

gmacdougall:
Great work John!

PR #2495, opened 2018-01-10
 Avoid JSON serializing Float::INFINITY 

### LINKED ISSUE #2486: With oj gem - ActionView::Template::Error Infinity not allowed in JSON
### Steps to reproduce

Load an order from the Solidus Checkouts API with a Line Item Variant with `track_inventory: false`, using the following responder:

`respond_with(@order, default_template: 'spree/api/orders/show', status: 200)`

### Actual behavior

The API will throw: `ActionView::Template::Error Infinity not allowed in JSON.`

Source of the issue:
https://github.com/solidusio/solidus/blob/96b30d6f4e644383cfa13fb2dd208bcd3589cb81/api/app/views/spree/api/variants/_small.json.jbuilder#L8

As it corresponds to:
https://github.com/solidusio/solidus/blob/fb70a41064aa34d9e350e096998ce638fd83bec7/core/app/models/spree/stock/quantifier.rb#L26

### Expected behavior

I would expect that in the case `track_inventory` is false, the `_small.json.jbuilder` template would not include the `total_on_hand` attribute.

### System configuration
**Solidus Version**: 2.4.2

jhawthorn:
This works fine under the default JSON library (it ends up being NULL), but it errors under OJ.

Fixes #2486

### DIFF
--- a/api/app/views/spree/api/variants/_big.json.jbuilder
+++ b/api/app/views/spree/api/variants/_big.json.jbuilder
@@ -1,7 +1,6 @@
 json.cache! [I18n.locale, Spree::StockLocation.accessible_by(current_ability), variant] do
   json.(variant, *variant_attributes)
   json.partial!("spree/api/variants/small", variant: variant)
-  json.total_on_hand(variant.total_on_hand)
   json.variant_properties(variant.variant_properties) do |variant_property|
     json.(variant_property, *variant_property_attributes)
   end
--- a/api/app/views/spree/api/variants/_small.json.jbuilder
+++ b/api/app/views/spree/api/variants/_small.json.jbuilder
@@ -5,7 +5,12 @@ json.cache! [I18n.locale, variant] do
   json.track_inventory(variant.should_track_inventory?)
   json.in_stock(variant.in_stock?)
   json.is_backorderable(variant.is_backorderable?)
-  json.total_on_hand(variant.total_on_hand)
+
+  # We can't represent Float::INFINITY in JSON
+  # Under JSON this woulb be NULL
+  # Under oj this would error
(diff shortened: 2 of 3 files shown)


hhff:
LGTM but i'd guess that `in_stock?` and potentially `is_backorderable?` also doesn't need to be serialized when we're not tracking inventory.  Just a thought!

jhawthorn:
> LGTM but i'd guess that in_stock? and potentially is_backorderable? also doesn't need to be serialized when we're not tracking inventory. Just a thought!

I've spent some time thinking about this. You're right that it's confusing to have values that should basically be ignored when track_inventory is off. I think in stock should probably keep `in_stock` even if we were to hide backorderable, as we do in most places consider track_inventory=false to imply in_stock.

Related to this is #2522, where we are considering adding another attribute/scope to represent (in_stock || backorderable || !track_inventory).

For now, at least, I think we should leave all of these fields in for compatibility.

joshhepworth:
I just ran into this issue as well, and actually had to address this in the Products API as well due to this line: https://github.com/solidusio/solidus/blob/master/api/app/helpers/spree/api/api_helpers.rb#L54

The because the `Spree::Product#any_variants_not_track_inventory?` method is private, I ended up with this in `_product.json.jbuilder`:

```
  json.(product, *@product_attributes - [:total_on_hand])

  json.total_on_hand(product.total_on_hand.to_f.infinite? == 1 ? nil : product.total_on_hand)
```

I'm not sure if this is acceptable, or if you'd rather have a solution that cleaned this up a little more extensively. I'm also not sure if the `product_attributes` helper ends up touching other endpoints as well.

PR #2496, opened 2018-01-10
Fix migration to be reversible

brchristian:
The migration as written is not reversible unless we specify the data type. This fixes the issue and allows for an easily reversible migration.


gmacdougall:
Thanks for the contribution!

PR #2498, opened 2018-01-10
Initial orders documentation

benjaminwil:
I ported the orders documentation over from the Spree guides. Then, I rewrote it with Solidus in mind. I also split the article into four smaller articles.

The main overview article summarizes how some other models (like `Spree::Adjustment` and `Spree::Shipment`) supplement the order, but in some cases I was not able to link to additional documentation because it doesn't exist yet (or just hasn't been merged). I'll come back and do that when the time is right.


benjaminwil:
@avremel Thanks for catching that error. Fixed!

guides/orders/display-total-methods.md
```diff
@@ -0,0 +1,41 @@
...
+@order.display_total.to_html 
+# => "$10.99"
+```
+
+Because `Spree::Money` objects are based on the [Ruby Money
+library][ruby-money], you can further change what information is displayed using
+its [`format`][ruby-money-format] method:
+
+```ruby
+@order.display_total.format(:with_currency => true)
```
kennyadsl: what about using the new ruby syntax here? `.format(with_currency: true)`

guides/orders/overview.md
```diff
@@ -0,0 +1,162 @@
...
+Whenever information about the order is updated, the `Spree::OrderUpdater`
+should update the order. If your store creates or changes the functionality
+around the order, you may need to [manually call the order
+updater][update-orders].
+
+Orders have the following attributes:
+
+- `number`: The unique identifier for this order. It begins with the letter `R`
+  and ends in a nine-digit number (for example, `R123456789`). This number is
+  shown to the users, and can be used to find the order by calling
```
kennyadsl: maybe here we should link the [`Order::NumberGeneratore`](https://github.com/solidusio/solidus/blob/c74c7d0351cf678406dee218a19070756bcbfbbf/core/app/models/spree/order/number_generator.rb#L19) which allows to change how this number is generated. WDYT?

guides/orders/overview.md
```diff
@@ -0,0 +1,162 @@
...
+  machine][order-state-machine] article for more information.
+- `adjustment_total`: The sum of all adjustments on this order.
+- `user_id`: The ID for the order's corresponding user. Stored only if the order
+  is placed by a signed-in user.
+- `completed_at`: The timestamp that logs when the order is completed.
+- `bill_address_id` and `ship_address_id`: The IDs for the related
+  `Spree::Address` objects with billing and shipping address information.
+- `payment_total`: The sum of all the *finalized* payments on the order.
+- `shipment_state`: The current [shipment state][shipment-states] of the order.
+- `payment_state`: The current payment state of the order.
```
kennyadsl: this state needs a link to its own documentation like `shipment state`, right? I assume it's not here since the payment docs is not merged yet, what about a todo? is it possible within the list?

benjaminwil: Yes, good call. Thank you!

guides/orders/overview.md
benjaminwil: Done! Thanks for the suggestion.

guides/orders/display-total-methods.md
benjaminwil: :+1:

benjaminwil:
@kennyadsl Thanks for the review!

PR #2501, opened 2018-01-11
Fix calculator class check bug

pelargir:
This PR fixes a bug found when attempting to change a shipping method's calculator to a different calculator that is a subclass of its current calculator. The existing implementation of this code prevents switching between calculators that subclass other calculators.

Using `#is_a?` to check the calculator class doesn't play well with calculators that are subclasses of each other. For example, given calculator A and calculator B, where B is a subclass of A: if a shipping method is configured to use calculator B, it becomes impossible to switch the shipping method to calculator A since `A.is_a?(B)` returns true. Whereas `A.instance_of?(B)` returns false.

If the intention is to prevent the same calculator from being set twice, `#instance_of?` is the correct choice since it disregards subclasses and only returns true if both are the exact same class.

### DIFF
--- a/core/app/models/concerns/spree/calculated_adjustments.rb
+++ b/core/app/models/concerns/spree/calculated_adjustments.rb
@@ -32,7 +32,7 @@ module Spree
 
     def calculator_type=(calculator_type)
       klass = calculator_type.constantize if calculator_type
-      self.calculator = klass.new if klass && !calculator.is_a?(klass)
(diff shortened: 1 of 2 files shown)


mamhoff:
Hey @matt, thank you! Would you mind adding a spec for the intended behaviour? I know there is one :)

pelargir:
@mamhoff done!

PR #2504, opened 2018-01-15
Improve pricing options flexibility

softr8:
The last piece, I think, to make pricing options totally flexible is to allow how it's built from a view context, allowing developers to decide how to find the prices based, example: users role

It also adjusts the API to use `object.price_for(current_pricing_options)`

I can add tests if you think this is a good idea.

### DIFF
--- a/api/app/helpers/spree/api/api_helpers.rb
+++ b/api/app/helpers/spree/api/api_helpers.rb
@@ -51,7 +51,7 @@ module Spree
       end
 
       @@product_attributes = [
-        :id, :name, :description, :price, :display_price, :available_on,
+        :id, :name, :description, :available_on,
         :slug, :meta_description, :meta_keywords, :shipping_category_id,
         :taxon_ids, :total_on_hand, :meta_title
       ]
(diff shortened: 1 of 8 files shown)


softr8:
Done @mamhoff, let me know what you think now... Thanks!

mamhoff:
Looks good, but can you look into the failing spec?

mamhoff:
Hm. The spec failure fix makes me wary. Why do the first two commits change the return value of `variant.default_price`? If we break that method, many admin views will cease to function.

softr8:
Actually, the failing spec was a test that sometimes fails, I saw it red in other PR's. My guess is that a simple rebuild would've made it green.

I didn't change `default_price` method at all, the reason I changed the form was because we have an external job that imports prices in bulk from a Google SpreadSheet, and if a variant didn't have any price yet, the variant edition was breaking, that's why I used find or build.

softr8:
Any comment @mamhoff ? I can rollback the view change if you feel it is not that safe.

kennyadsl:
@jtapia Hey I think this is a great change, can you please rebase against master?

spaghetticode:
@softr8 thanks for this PR 👏 
Can you please rebase this with current master?

softr8:
Rebased!

kennyadsl:
Thanks @softr8 !!

PR #2507, opened 2018-01-17
Remove eager touching from OptionValue

brendandeere:
**Before:**
Any time an option value is saved with changes (or touched) all
associated variants are also touched.

In the core test suite this is called 803 times creating roughly 2 db
requests per call.

**After**:
Touching variants after an option value is updated now occurs in the
admin controllers when option types or option values are modified there.

**Some stats:** 
**Runtimes**
before -> bundle exec rake spec:core  270.94s user 4.17s system 99% cpu 4:37.70 total
after -> bundle exec rake spec:core  266.91s user 4.36s system 99% cpu 4:33.93 total

**Number of times run/module:**
api:  158
backend: 71
core: 803

touch_all_variants

### DIFF
--- a/backend/app/controllers/spree/admin/option_types_controller.rb
+++ b/backend/app/controllers/spree/admin/option_types_controller.rb
@@ -3,6 +3,8 @@ module Spree
     class OptionTypesController < ResourceController
       before_action :setup_new_option_value, only: :edit
 
+      update.after :touch_variants
+
       def update_values_positions
         params[:positions].each do |id, index|
           Spree::OptionValue.where(id: id).update_all(position: index)
@@ -34,6 +36,11 @@ module Spree
         else
           Spree::OptionType.all
         end
+      end
+
+      def touch_variants
+        option_values = @object.option_values
+        ActiveRecord::Base.transaction { option_values.each(&:touch_all_variants) }
       end
     end
   end
--- a/core/app/models/spree/option_value.rb
+++ b/core/app/models/spree/option_value.rb
@@ -9,9 +9,6 @@ module Spree
     validates :name, presence: true, uniqueness: { scope: :option_type_id, allow_blank: true }
     validates :presentation, presence: true
 
-    after_save :touch, if: :saved_changes?
-    after_touch :touch_all_variants
-
     delegate :name, :presentation, to: :option_type, prefix: :option_type
 
     self.whitelisted_ransackable_attributes = ['presentation']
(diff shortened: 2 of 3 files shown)


gmacdougall:
I'm not a huge fan of moving from model based touching to controller based touching. I would be much happier if we could find a solution where the model still has that responsibility.

With Solidus being an extensible system, I can foresee people having extensions to their application which create or modify option values that this would break for them.

mamhoff:
I agree with @gmacdougall here. Shaving four seconds off the test suite for potentially introducing caching errors does not seem worth it to me. Also, I have seen code bases where option values are changed from Rake tasks...

brendandeere:
How would you feel about leaving `touch_all_variants` in the model and simply removing the callbacks instead?

That would keep the api consistent and it would be easy for extensions to call the touching method as needed.

mtomov:
Sort of related: with lots of products and variants, updating option values and types can bog down the database, so on one such app I've done:

```ruby
# option_value.rb

  # Re-implement to handle lots of variants and products
  def touch_all_variants
    Spree::OptionValue::TouchAllVariants.perform_later(id)
  end

  class Spree::OptionValue::TouchAllVariants < ActiveJob::Base
    def perform(option_value_id)
      current_time = Time.current

      Spree::Variant.joins(:option_values_variants).
        where(spree_option_values_variants: { option_value_id: option_value_id } ).
        select(:id, :product_id).find_in_batches(batch_size: 5000) do |batch|
          variant_ids = batch.map(&:id)
          product_ids = batch.map(&:product_id).uniq

          Spree::Variant.where(id: variant_ids).update_all(updated_at: current_time)
          Spree::Product.where(id: product_ids).touch_all
      end
    end
  end
```

and

```ruby
# option_type.rb

  # Re-implement to handle lots of products
  def touch_all_products
    return unless changed?

    Spree::Product.joins(:product_option_types).
      where(spree_product_option_types: { option_type_id: id } ).
      select(:id).find_in_batches(batch_size: 500) do |batch|
        product_ids = batch.map(&:id)
        Spree::Product.where(id: product_ids).touch_all
    end
  end
```

But overall, I don't think that this is also too good.

I think that some sort of non-imperative system (EventBus) is probably better, as it must be the responsibility of the relevant model (Product, Variant, Taxon) to *hook itself* and *listen* for updates (on Option Type, Option Variant, etc.. ) and decide on it's own how to go about updating itself  - be it in batches, in background jobs, which columns if you will, and even throw further update events..

gmacdougall:
Going to close this as I don't foresee us coming up with a reasonable solution for dealing with the drawbacks of this. It's something people may want to consider in their own stores, but as it it, is too potentially dangerous to merge.

PR #2509, opened 2018-01-18
Use datalist instead of jquery-ui/autocomplete

jhawthorn:
Replaces `jquery-ui/autocomplete`, which is only used in the product properties page, with the html5 `<datalist>` element.

[Browser support for datalist](https://caniuse.com/#feat=datalist) is patchy (notably no safari), but it is entirely fine for this page not to have autocompletion.

We had apparently already decided that the number of properties is few enough that they can be just dumped on the page as json. I'm keeping this assumption and just dumping them on the page as HTML.

**Before**

**After** (display will be browser and OS dependent)

See also #2469

### DIFF
--- a/backend/app/assets/javascripts/spree/backend.js
+++ b/backend/app/assets/javascripts/spree/backend.js
@@ -4,7 +4,6 @@
 //= require jquery_ujs
 //= require jquery-ui/datepicker
 //= require jquery-ui/sortable
-//= require jquery-ui/autocomplete
 //= require jquery.sticky-kit.min
 //= require solidus_admin/select2
 //= require solidus_admin/underscore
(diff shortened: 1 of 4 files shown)


mtomov:
Good choice!

tvdeyen:
Just for reference. Select2 also has autocomplete functionality.

jhawthorn:
> Select2 also has autocomplete functionality.

I don't think select2 can do what we need here. We want to have autocomplete (selection from a list), but we also want to be able to type a value not in our autocomplete list. select2 allows selecting a single value from a list (it's default behaviour), and it allows taking arbitrary typed input when it's taking multiple values as token input, but I don't think it can accept a single arbitrary value with suggestions (I could be wrong though).

tvdeyen:
It’s a little arkward but it’s doable https://github.com/AlchemyCMS/alchemy_cms/blob/master/app/assets/javascripts/alchemy/alchemy.autocomplete.js.coffee

I am not saying we should use select2, but as we already have the dependency it’s worth looking into it.

PR #2511, opened 2018-01-18
Initial calculators documentation

benjaminwil:
I've added an overview of `Spree::Calculator` (and its sub-classes). It goes over calculator attributes, preferences, and what a "calculable" is. 

Information about creating custom calculators will live in the promotions, shipments, and taxation documentation respectively. So far I've only fleshed out the "Custom shipping calculators" article. The "Custom tax calculators" remains as a stub for now. An article about custom promotion actions has yet to be merged. 

In the future, there will also be articles summarizing the functionality/features of all of the built-in calculators.

### DIFF
--- a/guides/calculators/overview.md
+++ b/guides/calculators/overview.md
@@ -13,8 +13,9 @@ charges a flat rate of $10 USD, you would be creating an instance of the
 `Spree::Calculator::Shipping::FlatRate` calculator:
 
 - The new calculator instance would have `calculable_type` of
-  `Spree::ShippingMethod` (because it is associated with a shipping method, not a
-  tax rate or promotion action).
+  `Spree::ShippingMethod`. This is because it has a polymorphic relationship
+  with a shipping method, rather than a tax rate or promotion action. See
+  [calculables](#calculables) for more information about this relationship.
 - The new calculator instance's `preferences` attribute would have a value of
   `{:preference=>10, :currency=>"USD"}`.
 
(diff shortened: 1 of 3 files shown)


guides/calculators/overview.md
```diff
@@ -0,0 +1,114 @@
...
+
+Whenever you create a new shipping method, tax rate, or promotion action, you
+also create a new instance of a `Spree::Calculator`.
+
+For example, if you create a new shipping method called "USPS Ground" that
+charges a flat rate of $10 USD, you would be creating an instance of the
+`Spree::Calculator::Shipping::FlatRate` calculator:
+
+- The new calculator instance would have `calculable_type` of
+  `Spree::ShippingMethod` (because it is associated with a shipping method, not a
```
gmacdougall: It would be best here that `calculable` is a polymorphic relationship and possibly link to the rails documentation on that.

guides/calculators/overview.md
```diff
@@ -0,0 +1,114 @@
...
+So, the rate for each shipment is calculated by the associated
+`Spree::Calculator::Shipping::FlatRate` object.
+
+Similarly, each tax rate in your store is calculable. So, instance of the
+`Spree::Calculator::DefaultTax` calculator is created and calculates the amount
+of tax that should be applied to line items, shipments, or orders.
+
+## Preferences
+
+Each `Spree::Calculator` has a `preferences` attribute. Its value is a hash that
```
gmacdougall: Static model preferences should be introduced here.

guides/shipments/custom-shipping-calculators.md
```diff
@@ -50,15 +50,115 @@ Typically, a calculator uses the following order information:
...
+You may want to make the calculator availability change depending on some aspect
+of the current order. To do this, you can override the `available?` method in
+your custom calculator:
+
+```ruby
+module MyStore
+  class Calculator::Shipping::CustomShippingCalculator < Spree::ShippingCalculator
+    ...
+
+    def available?(object)
```
gmacdougall: I would recommend users against defining something as `object`. What is the object being passed in here? Lets explain that.

guides/calculators/overview.md
benjaminwil: Good call. Some initial docs I wrote introducing static model preferences was just merged, so I will link there.

guides/shipments/custom-shipping-calculators.md
benjaminwil: I'm trying to check that this is possible now. Would it make sense to change `object` to `order` for this example? Or is the whole example kind of unrealistic, maybe?

gmacdougall: Yup. That makes sense. The example is fine as is as a trivial example, but will be helpful for people wondering what `object` is.

Issue #2512, opened 2018-01-18
Orders api view not respecting other payment sources


skukx:
### Steps to reproduce
- Create an order with a payment from a non-traditional payment source
- Go back to cart page
- Open debugger in browser
- Refresh page

You should see an api request which returns a 500 status code in the response. This also causes the page to not load entirely (Add Item button greyed out).

The error is as follows:
```
ActionView::Template::Error (undefined method `month' for #<Spree::GiftCard:0x00007fcdb9bee298>):
    25:   json.payment_method { json.(payment.payment_method, :id, :name) }
    26:   json.source do
    27:     if payment.source
    28:       json.(payment.source, *payment_source_attributes)
    29: 
    30:       if @current_user_roles.include?("admin")
    31:         json.(payment.source, :gateway_customer_profile_id, :gateway_payment_profile_id)
```
The reason this happens is due to the view trying to render attributes for payment sources which not all payment sources might have.

See: https://github.com/solidusio/solidus/blob/master/api/app/views/spree/api/orders/_big.json.jbuilder#L28-L32L144

See: https://github.com/solidusio/solidus/blob/master/api/app/helpers/spree/api/api_helpers.rb#L142-L144

### System configuration
**Solidus Version**:
2.4

### Possible Resolution
I think what should be done is much like what is done on the frontend.

See: https://github.com/solidusio/solidus/tree/master/frontend/app/views/spree/checkout/payment
See: https://github.com/solidusio/solidus/blob/master/frontend/app/views/spree/checkout/_payment.html.erb#L53

Where each payment source is given a partial which is rendered based on the payment source type.

PR #2513, opened 2018-01-18
Fix view for orders api (Fixes Issue #2512)

skukx:
Refactor _big.json.jbuilder so that it dynamically loads the payment
source view based on that source's payment method.

Add payment views for each payment method which will be used to render a
payment source as json.

Fixes https://github.com/solidusio/solidus/issues/2512.

### DIFF
--- a/api/app/views/spree/api/orders/_big.json.jbuilder
+++ b/api/app/views/spree/api/orders/_big.json.jbuilder
@@ -27,11 +27,10 @@ json.payments(order.payments) do |payment|
   json.payment_method { json.(payment.payment_method, :id, :name) }
   json.source do
     if payment.source
-      json.(payment.source, *payment_source_attributes)
-
-      if @current_user_roles.include?("admin")
-        json.(payment.source, :gateway_customer_profile_id, :gateway_payment_profile_id)
-      end
+      json.partial!(
+        "spree/api/payments/source_views/#{payment.payment_method.partial_name}",
+        payment_source: payment.source
+      )
     else
       json.nil!
     end
--- a/api/app/views/spree/api/orders/_big.json.jbuilder
+++ b/api/app/views/spree/api/orders/_big.json.jbuilder
@@ -26,10 +26,17 @@ json.payments(order.payments) do |payment|
   json.(payment, *payment_attributes)
   json.payment_method { json.(payment.payment_method, :id, :name) }
   json.source do
-    if payment.source
+    ##
+    # payment.source could be a Spree::Payment. If it is then we need to call
+    # source twice.
+    # @see https://github.com/solidusio/solidus/blob/v2.4/backend/app/views/spree/admin/payments/show.html.erb#L16
+    #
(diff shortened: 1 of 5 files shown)


skukx:
Yeah I can add a test for this. As far as the naming convention, I followed convention on how Spree admin chooses to display payment sources.

See: https://github.com/solidusio/solidus/tree/master/backend/app/views/spree/admin/payments/source_views

If we want to move away from that convention I will change the views mimic that payment source rather than the method

mamhoff:
Sorry, I thought we changed the partial name with deprecating the `Spree::Gateway` class, but we didn't: 
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment_method/credit_card.rb#L15

So, keep the partial name, but do add a spec. :heart:

skukx:
Let me know if the specs added are sufficient

api/spec/requests/spree/api/orders_controller_spec.rb
```diff
@@ -317,6 +317,21 @@ module Spree
           expect(credit_cards[0]['id']).to eq payment.source.id
           expect(credit_cards[0]['address']['id']).to eq credit_card.address_id
         end
+
+        it 'renders the payment source view for gateway' do
+          subject
+          expect(response).to render_template partial: "spree/api/payments/source_views/_#{payment.payment_method.partial_name}"
```
mamhoff: These are sufficient, but should be more explicit. Either `render_template('spree/api/payments/source_views/_gateway.html.erb`), or test directly for the right thing being rendered as JSON.

skukx:
Made changes as requested

api/app/views/spree/api/orders/_big.json.jbuilder
```diff
@@ -25,11 +25,10 @@ json.payments(order.payments) do |payment|
   json.payment_method { json.(payment.payment_method, :id, :name) }
   json.source do
     if payment.source
```
skukx: It is possible that `payment.source` is a `Spree::Payment`.  
See: https://github.com/solidusio/solidus/blob/v2.4/backend/app/views/spree/admin/payments/show.html.erb#L16

We will need to account for that situation here as well:
```ruby
json.source do
    ##
    # payment.source could be a Spree::Payment. If it is then we need to call
    # source twice.
    # @see https://github.com/solidusio/solidus/blob/v2.4/backend/app/views/spree/admin/payments/show.html.erb#L16
    #
    payment_source = payment.source.is_a?(Spree::Payment) ? payment.source.source : payment.source
    
    if payment_source
      json.partial!(
        "spree/api/payments/source_views/#{payment.payment_method.partial_name}",
        payment_source: payment_source
      )
    else
      json.nil!
    end
  end
```

When I get a chance I'll add this to the PR

skukx:
What is the status of this?

gmacdougall:
There's a red CircleCI build on this right now. Do you know the source of the errors?

skukx:
I just pushed something new so it may have to do with that. I'll take a look

skukx:
Look like I just needed rebase my branch on master

PR #2525, opened 2018-01-23
Add documentation that summarizes Solidus's built-in shipping calculators

benjaminwil:
This pull request extends my previous pull request for calculators documentation (#2511). I recommend that that pull request be reviewed and merged first.

### DIFF
--- a/guides/calculators/shipping-calculators.md
+++ b/guides/calculators/shipping-calculators.md
@@ -1,0 +1,101 @@
+# Shipping calculators
+
+Solidus comes with a set of default calculators that account for typical
+shipping scenarios:
+
+- [Flat percent](#flat-percent)
+- [Flat rate](#flat-rate) (per order)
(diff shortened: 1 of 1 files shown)


guides/calculators/shipping-calculators.md
```diff
@@ -0,0 +1,101 @@
+# Shipping calculators
+
+Solidus comes with a set of default calculators that account for typical
+shipping scenarios:
+
+- [Flat percent](#flat-percent)
+- [Flat rate](#flat-rate) (per order)
+- [Flexi rate](#flexi-rate)
+- [Flat rate per package item](#flat-rate-per-package-item)
```
kennyadsl: what about moving `Flat rate per package item` immediately under `Flat rate`? Maybe being closer could help readers following their meanings?

benjaminwil: Good call. Fixed!

PR #2526, opened 2018-01-23
Add a stub article that summarizes Solidus's built-in tax calculator

benjaminwil:
This pull request extends my previous pull request for calculators documentation (#2511). It depends on #2511 being merged first.


guides/calculators/tax-calculator.md
```diff
@@ -0,0 +1,24 @@
...
+
+Solidus comes with a tax calculator that is used to calculate both sales tax
+(United States-style taxes) and value-added tax (VAT):
+[`Spree::Calculator::DefaultTax`][default-tax-calculator]. Typically, this
+calculator should be the only tax calculator required by your store.
+
+Taxes can apply to line items, shipments, or an entire order.
+
+The tax calculator uses its calculable – a `Spree::TaxRate` – to calculate tax
+totals. Tax rates are represented as decimal between `0.01` and `1.00`.
```
gmacdougall: I don't believe this is true. A tax rate has a precision of 8, scale 5 at the DB level and no such restrictions in rails validations that I know of.

guides/calculators/tax-calculator.md
```diff
@@ -0,0 +1,24 @@
...
+
+Solidus comes with a tax calculator that is used to calculate both sales tax
+(United States-style taxes) and value-added tax (VAT):
+[`Spree::Calculator::DefaultTax`][default-tax-calculator]. Typically, this
+calculator should be the only tax calculator required by your store.
+
+Taxes can apply to line items, shipments, or an entire order.
+
+The tax calculator uses its calculable – a `Spree::TaxRate` – to calculate tax
+totals. Tax rates are represented as decimal between `0` and `1`.
```
gmacdougall: The amounts that can be represented here are actually between -999.99999 and 999.99999 to potentially accounting for tax rates over 100% and (for some reason) negative tax rates.

I think it would be best to use a simple example here that a sales tax of 5% should be represented as `0.05` and not `5` as 5 is 500%.

PR #2527, opened 2018-01-24
Remove Jquery UI

jhawthorn:
Fully removes `jquery_ui_rails` and image assets stored in `backend/vendor/assets`.

Depends on:

* [x] #2523 
* [x] #2510
* [x] #2509


jhawthorn:
Now ready to go :shipit:

PR #2528, opened 2018-01-24
Improve HTTP usage in update_positions

jhawthorn:
This fixes some mildly bad HTTP behaviour that didn't really affect anything, but we might as well do it more properly.

* We were responding with `Content-Type: text/javascript` with a body of `"Ok"`. `"Ok"` is not valid JSON :no_entry_sign:
* `OptionTypesController#update_values_positions` had an unused `format.html` response, which didn't work.
* We were doing ajax with `dataType: 'script'` (ie. get some javascript and eval it), which isn't actually what we wanted.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/components/sortable_table.js
+++ b/backend/app/assets/javascripts/spree/backend/components/sortable_table.js
@@ -19,7 +19,7 @@ Spree.ready(function() {
         });
         Spree.ajax({
           type: 'POST',
-          dataType: 'script',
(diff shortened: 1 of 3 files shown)


jhawthorn:
I've spotted an `InvalidAuthenticityToken` issue with this on some pages.

jhawthorn:
> I've spotted an InvalidAuthenticityToken issue with this on some pages.

User error (mine)! Everything's good I had customized something else in my sandbox which was breaking this.

PR #2533, opened 2018-01-26
Change instances of "udpate" to "update"

dsojevic:
Fixed an instance of a permissions check on shipping categories checking for "udpate" instead of "update" and fixed the same spelling mistake in an API spec title/description.


jhawthorn:
tahnk you.

PR #2537, opened 2018-01-27
Use carmen to translate available_countries helper

jhawthorn:
WIP: will revisit and test properly next week

In https://github.com/solidusio-contrib/solidus_i18n/pull/106 we are looking to remove a feature from `solidus_i18n` which provides translations for each country based on the translations in the [`i18n_data`](https://github.com/grosser/i18n_data) gem.

We don't want to support this odd magic feature as part of i18n, and we'd like to remove the dependency on i18n_data, but at the same time we still want countries to be localized properly.

Ideally this would also be faster than the existing implementation as `available_countries` is a little bit slow.

This PR replaces the the existing behaviour of using `I18n.t("spree.country_names.#{iso_code}")` with using Carmen's data to perform translations.

This is the first time we are actually requiring and using carmen outside of tests and seeds, so we should consider the effect this has on startup time and memory usage.

## TODO

* [x] Continue to support overrides from `I18n` `spree.country_names.*`
* [x] Change `Carmen.i18n_backend` with each request
* [x] Benchmark
* [x] Consider startup time and memory usage

### DIFF
--- a/core/app/helpers/spree/base_helper.rb
+++ b/core/app/helpers/spree/base_helper.rb
@@ -119,6 +119,8 @@ module Spree
         [country.code, country.name]
       end.to_h
 
+      country_names.update I18n.t('spree.country_names', default: {}).stringify_keys
+
       countries.collect do |country|
         country.name = country_names.fetch(country.iso, country.name)
         country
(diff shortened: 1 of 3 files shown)


jhawthorn:
*After*
```
 available_countries    223.391  (± 4.5%) i/s -      1.116k in   5.006031s
```

*Before*
```
 available_countries     80.574  (± 6.2%) i/s -    406.000  in   5.057310s
```

kennyadsl:
I'm preparing a PR on solidus_i18n to remove that feature.

PR #2538, opened 2018-01-30
Add documentation overview of inventory

benjaminwil:
Adds an article that functions as an overview of Solidus's inventory system. It has links to other articles that will be committed in other pull requests. For now, these links are commented out.

### DIFF
--- a/guides/inventory/overview.md
+++ b/guides/inventory/overview.md
@@ -1,0 +1,131 @@
+# Overview
+
+Solidus can manage your store's inventory. This allows you to keep track of your
+current stock, [backorder][backorder] stock, as well as sold and shipped
+inventory. Because inventory touches orders, shipments, and the contents of your
+physical warehouse, there are many essential moving parts:
+
+- `Spree::InventoryUnit`: Represents a stock item that has been sold. For more
+  information, see [Inventory units](#inventory-units).
+- `Spree::StockItem`: Counts the inventory for a specific `Spree::Variant` at a
+  specific `Spree::StockLocation`. For more information, see [Stock
+  items](#stock-items). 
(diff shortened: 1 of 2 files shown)


guides/inventory/overview.md
```diff
@@ -0,0 +1,131 @@
...
+#### Stock transfers
+
+If you manage multiple stock locations and inventory frequently moves between
+them, you may benefit from the
+[`solidus_stock_transfers`][solidus-stock-transfers] extension. This extension
+adds a user interface for managing transfers in the `solidus_backend`.
+
+[solidus-stock-transfers]: https://github.com/solidusio-contrib/solidus_stock_transfers
+
+### Return authorizations
```
kennyadsl: It's not immediately clear (at least to me) how this section is related with stock management. I'm not saying it's not related but I think we should specify how here.

We should probably say that the return authorization is something that allows to create `ReturnItems`. Each `ReturnItem` in associated with an inventory unit that can be restocked when order items are received back. You can follow the code [here](https://github.com/solidusio/solidus/blob/9cac52900fadaf79b3157ecd3f0dc2016ae56528/core/app/models/spree/return_item.rb#L81) to see what happens. 

Do you think it could make sense to add these information?

benjaminwil: Thank you. This prompted me to restructure the documentation around returns. All of the RMA-related documentation I wrote now lives in it's own `returns/` section, and I've added a stub specifically about `Spree::ReturnItem`s to the inventory overview.

There is definitely more to say about RMAs, `ReturnItem`s and other aspects of the returns system. That will have to be added in future PRs.

kennyadsl: Cool, makes sense!

PR #2539, opened 2018-01-30
Add documentation for stock items and stock movements

benjaminwil:
This pull request adds articles about `Spree::StockItem`s and `Spree::StockMovement`s. (Stock locations documentation is summarized in #2538.)

### DIFF
--- a/guides/inventory/stock-items.md
+++ b/guides/inventory/stock-items.md
@@ -1,0 +1,48 @@
+# Stock items
+
+On-hand inventory is tracked using the `Spree::StockItem` model. Each
+`Spree::Variant` in a store has a corresponding `Spree::StockItem` object with a
+`count_on_hand` value that represents the number of items you have in stock.
+
+`Spree::StockItem` objects have the following attributes:
+
+- `stock_location_id`: The ID for the `Spree::StockLocation` where the stock
+  item is located.
+- `variant_id`: The ID for the `Spree::Variant` that this stock item represents.
+- `count_on_hand`: The number of units currently in inventory. See [Count on
+  hand](#count-on-hand) for more information.
+- `backorderable`: Sets whether the stock item should be
+  [backorderable](#backorderable-stock-items).
+- `deleted_at`: A timestamp that logs when this stock item was deleted from
+  inventory. Otherwise, the value is `nil`.
+
+## Count on hand
+
+When administrators create a new product using the `solidus_backend`, they can
+set an initial inventory "Count On Hand" value for the new product.
+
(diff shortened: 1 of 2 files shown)


guides/inventory/stock-items.md
```diff
@@ -0,0 +1,48 @@
+# Stock items
+
+On-hand inventory is tracked using the `Spree::StockItem` model. Each
+`Spree::Variant` in a store has a corresponding `Spree::StockItem` object with a
+`count_on_hand` value that represents the number of items you have in stock.
```
kennyadsl: > Each `Spree::Variant` in a store has a corresponding `Spree::StockItem` object

it has a corresponding `Spree::StockItem` object for each stock location present in the store. So the same variant can have more than one `Spree::StockItem` associated objects.

guides/inventory/stock-items.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Changing the count on hand value
+
+While a `Spree::StockMovement` object logs the increase or decrease of the
+`count_on_hand` value, administrators can also edit the count on hand from the
+`solidus_backend`.
+
+Whenever an administrator updates the count on hand, they are discarding the old
+value completely. So, if a stock item is
+[backorderable](#backorderable-stock-items) and its `count_on_hand` is `-5`,
```
kennyadsl: for the sake of the example, is it relevant to say that the item should be backorderable? Maybe it's more clear if we remove this concept from here and use a positive number for the initial `count_on_hand`. WDYT?

guides/inventory/stock-items.md
```diff
@@ -0,0 +1,48 @@
...
+
+See the [Stock movements][stock-movements] article for more information.
+
+[stock-movements]: stock-movements.md
+
+## Backorderable stock items
+
+If a `Spree::StockItem` is `backorderable`, then customers can continue to order
+it after the product is sold out. When a sold out product continues to sell, the
+`count_on_hand` would become a negative integer.
```
kennyadsl: Maybe here we should say that orders that contains backorderable stock item/inventory units (not sure which concept it's better to mention here in order to not increase complexity) can not be shipped until that stock item has been restocked in the stock location. Not sure if we have other guide for this topic, maybe we can just link there?

benjaminwil: There aren't any other guides on this topic yet, so I'm going to add a bit more information here. Thanks for pointing this out!

benjaminwil:
@kennyadsl Thanks for the review. I've addressed all your comments. :+1:

PR #2540, opened 2018-01-30
Add initial RMA and return authorization documentation

benjaminwil:
This pull request adds stub articles for `Spree::InventoryUnit`s and `Spree::ReturnAuthorization`s. The return authorizations article introduces the concept of RMAs but not the larger returns/reimbursements system in Solidus, which will come later as its own section.

### DIFF
--- a/guides/inventory/inventory-units.md
+++ b/guides/inventory/inventory-units.md
@@ -1,0 +1,38 @@
+# Inventory units
+
+<!-- TODO:
+  This article is a stub.
+-->
+
+A `Spree::InventoryUnit` object is created every time that an item is sold.
+It tracks a sold item as it changes location: from being a sold item waiting in the
+warehouse to be shipped, to being a shipped item (or potentially a returned item).
+
+Note that the `Spree::InventoryUnit` tracks an item as an object associated with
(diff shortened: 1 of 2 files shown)


guides/inventory/inventory-units.md
```diff
@@ -0,0 +1,40 @@
...
+allows you to more closely track the status of an order and the line items and
+shipments associated with it.
+
+A `Spree::InventoryUnit` object has the following attributes:
+
+- `state`: The current state of the inventory unit. The state value can be
+  `on_hand`, `backordered`, `shipped`, or `returned`.
+- `variant_id`: The ID for the `Spree::Variant` corresponding with the inventory
+  unit that has been sold.
+- `order_id`: The ID for the `Spree::Order` associated with the current
```
kennyadsl: I'm not sure if it's worth specifying that this attribute is no more [on the Inventory Unit but it's calculated from the associated shipment of the Inventory Unit](#2377).

benjaminwil: Thank you for pointing this out! For now, I will simply remove this item from the list.

PR #2544, opened 2018-01-31
Use relative path on New Order button to support multiple stores

ericsaupe:
With the way the default #current_store method and
current_store_selector_class work the absolute path generated by
_url can cause the new order created to fallback to a different
domain than the one the user is currently on.

### DIFF
--- a/backend/app/views/spree/admin/orders/index.html.erb
+++ b/backend/app/views/spree/admin/orders/index.html.erb
@@ -5,7 +5,7 @@
 
 <% content_for :page_actions do %>
   <li>
-    <%= button_link_to t('spree.new_order'), new_admin_order_url, id: 'admin_new_order' %>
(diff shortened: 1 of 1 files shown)


ericsaupe:
Now that I look a little deeper it seems that there are many buttons under `content_for :page_actions` that are `_url` while many of the other links on these backend pages are `_path`. Should they all be changed to `_path` to better support multiple stores?

jhawthorn:
What's the issue with using `_url`? Shouldn't it use the domain of the current request?

ericsaupe:
Looking into it further it looks like it's because we do something like:
```
config.action_controller.default_url_options = {
    host: 'myfirststoredomain.com',
    protocol: 'https'
  }
```
in our environment configs. This causes all `_url` paths to have `myfirststoredomain.com` instead of the current domain. I don't think we actually need that and removing that `host` param does resolve it. I'll close this for the time being.

Thanks

PR #2546, opened 2018-01-31
Convert trivial coffee to plain JS

jhawthorn:
This converts many of the remaining CoffeeScript files to plain JS.

This focuses on converting our remaining CoffeeScript files which can be straightforwardly converted to plain JS. These aren't huge files, are reasonably well organized already, and don't heavily use any of CoffeeScript's more powerful features (like classes).

This converts ~200 lines of CoffeeScript into ~280 lines of JS.

After this we are left with 250 lines of coffee in only a few sections, which I will tackle in separate PRs:
* components/tabs (which is a large class)
* promotions (which could be better organization)
* stock_management (which should have views moved into `views/` directory)
* store_credits (which could be better organized)


kennyadsl:
Have you used any tool to perform this conversion?

jhawthorn:
> Have you used any tool to perform this conversion?

I used coffeescript (1.x version, since 2.0+ emits es6) and manually edited the output.

backend/app/assets/javascripts/spree/backend/gateway.js
```diff
@@ -0,0 +1,22 @@
+Spree.ready(function() {
+  var $gateway_type = $('select.js-gateway-type');
+  var $preference_source = $('select.js-preference-source');
+  var original_gtwy_type = $gateway_type.val();
```
kennyadsl: what about taking advantage of this PR to change this variable name to `original_gateway_type`?

PR #2550, opened 2018-02-02
Add store select to payment method admin

adammathys:
Payment methods belong to stores, but there was previously no way of managing that association in the admin. This adds a select field allowing us to easily control that association.

store-payment-admin


backend/app/assets/javascripts/spree/backend/store_autocomplete.js
```diff
@@ -0,0 +1,45 @@
+$.fn.storeAutocomplete = function() {
+  "use strict";
+
+  this.select2({
+    minimumInputLength: 1,
+    multiple: true,
+    initSelection: function(element, callback) {
+      console.log(element.val());
```
jhawthorn: !

PR #2552, opened 2018-02-02
Add store promotion rule

adammathys:
Introduces a new promotion rule (originally part of [`solidus_multi_domain`](https://github.com/solidusio/solidus_multi_domain/blob/master/app/models/spree/promotion/rules/store.rb)) that allows for scoping promotions to specific stores.

store-promo-rule

### DIFF
--- a/core/app/models/spree/promotion/rules/store.rb
+++ b/core/app/models/spree/promotion/rules/store.rb
@@ -1,0 +1,20 @@
+module Spree
+  class Promotion
+    module Rules
+      class Store < PromotionRule
+        has_many :promotion_rule_stores, class_name: "Spree::PromotionRuleStore",
+                                         foreign_key: :promotion_rule_id,
+                                         dependent: :destroy
+        has_many :stores, through: :promotion_rule_stores, class_name: "Spree::Store"
+
+        def applicable?(promotable)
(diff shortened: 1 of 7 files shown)


core/app/models/spree/promotion/rules/store.rb
```diff
@@ -0,0 +1,20 @@
...
+                                         foreign_key: :promotion_rule_id,
+                                         dependent: :destroy
+        has_many :stores, through: :promotion_rule_stores, class_name: "Spree::Store"
+
+        def applicable?(promotable)
+          promotable.is_a?(Spree::Order)
+        end
+
+        def eligible?(order, _options = {})
+          stores.none? || stores.include?(order.store)
```
jhawthorn: it seems really weird to check `stores.none?` here

adammathys: It is a little odd. I'm not sure why it was originally written this way in [`solidus_multi_domain`](https://github.com/solidusio/solidus_multi_domain/blob/master/app/models/spree/promotion/rules/store.rb#L11-L13), but I figured it would be best to keep it the same.

tvdeyen:
Thanks

PR #2555, opened 2018-02-02
Splitted shipment should update order totals/payment status

### LINKED ISSUE #2554: Order totals and payment status are not recalculated after splitting shipment
### Steps to reproduce
1. Place an order with (at least) 2 line items in the cart
2. Go to the admin panel and select the placed order
(Optional) Capture the pending payment
3. Split the created shipment into 2 separated shipment with the usual flow
4. Verify how order totals (total, ship total, tax)
If you captured the payment, check how payment status didn't change from `paid` to `balance due`

### Expected behavior
Ship total should change (double)
Tax total should change (double, based on the shipping method's tax rate)
Order total should reflect the increase of ship total and tax total
Order's payment status should change from `paid` to `balance due` if the payment was captured

### Actual behavior
New shipment is created properly but order is not updated, totals nor payment status

Here's a gif with the problem in the sandbox:

issue

### System configuration
**Solidus Version**: Latest from `master` branch

**Extensions in use**: None

VzqzAc:
Fixes #2554, detailed issue info is in there

Here's a gif with the solution in place:

working

I also added a small test to make sure after calling `run!` in `Spree::FulfilmentChanger` the order is properly recalculated

### DIFF
--- a/core/app/models/spree/fulfilment_changer.rb
+++ b/core/app/models/spree/fulfilment_changer.rb
@@ -96,6 +96,10 @@ module Spree
       # are up-to-date, too.
       desired_shipment.refresh_rates
 
+      # In order to reflect the changes in the order totals
+      desired_shipment.order.reload
+      desired_shipment.order.recalculate
+
       true
     end
 
--- a/core/spec/models/spree/fulfilment_changer_spec.rb
+++ b/core/spec/models/spree/fulfilment_changer_spec.rb
@@ -57,6 +57,15 @@ RSpec.describe Spree::FulfilmentChanger do
       subject
     end
 
+    it 'updates order totals' do
+      original_total = order.total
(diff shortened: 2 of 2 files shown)


VzqzAc:
I'm having troubles with `fulfilment_changer_spec` because of `RuntimeError: Can't modify frozen hash`, debugging it's happening with the shipment that is supposed to be destroyed in both `when the current shipment is emptied out by the transfer` context, I'm working on how to get rid of it

VzqzAc:
Thanks for your response @jhawthorn, I'd also prefer not to use it but yeah, it's necessary, in case `current_shipment` is deleted it is frozen at the time `order.updater.update` is called and causes runtime errors while trying to update items adjustments in `OrderUpdater` model

PR #2556, opened 2018-02-05
Remove coffeescript

jhawthorn:
(Depends on unmerged #2553)

This removes the coffee-script dependency from solidus_backend. It also updates the README and removes `coffeelint.json`.

Extensions which use coffeescript should now declare their coffee-rails dependency explicitly (or switch to plain JS as well).

Initially after doing this I was getting:
```
LoadError:
  cannot load such file -- coffee_script
```
but clearing `spec/dummy/tmp/cache` fixed this.

Stores which use coffeescript should have no issues so long as they declare `gem "coffee-rails"` in their Gemfile (which they likely do as that is the rails default).

## TODO
* [ ] For some reason teaspoon is failing :bomb:


jhawthorn:
I don't think teaspoon can be made to work with this. Oh well.

PR #2557, opened 2018-02-06
Add association between stores and shipping

adammathys:
Like payment methods, shipping methods are typically store specific.

store-shipping-admin

### DIFF
--- a/core/app/models/spree/shipping_method.rb
+++ b/core/app/models/spree/shipping_method.rb
@@ -37,6 +37,11 @@ module Spree
 
     validate :at_least_one_shipping_category
 
+    scope :available_to_store, ->(store) do
+      raise ArgumentError, "You must provide a store" if store.nil?
+      store.shipping_methods.empty? ? all : where(id: store.shipping_method_ids)
+    end
+
     # @param shipping_category_ids [Array<Integer>] ids of desired shipping categories
     # @return [ActiveRecord::Relation] shipping methods which are associated
     #   with all of the provided shipping categories
--- a/core/app/models/spree/shipping_method.rb
+++ b/core/app/models/spree/shipping_method.rb
@@ -30,6 +30,9 @@ module Spree
     has_many :shipping_method_stock_locations, dependent: :destroy, class_name: "Spree::ShippingMethodStockLocation"
     has_many :stock_locations, through: :shipping_method_stock_locations
 
+    has_many :store_shipping_methods, inverse_of: :shipping_method
+    has_many :stores, through: :store_shipping_methods
+
     validates :name, presence: true
 
     validate :at_least_one_shipping_category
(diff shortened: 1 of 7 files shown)


core/app/models/spree/shipping_method.rb
```diff
@@ -30,10 +30,18 @@ class ShippingMethod < Spree::Base
...
+    has_many :store_shipping_methods, inverse_of: :shipping_method
+    has_many :stores, through: :store_shipping_methods
+
     validates :name, presence: true
 
     validate :at_least_one_shipping_category
 
+    scope :available_to_store, ->(store) do
+      raise ArgumentError, "You must provide a store" if store.nil?
+      store.shipping_methods.empty? ? all : where(id: store.shipping_method_ids)
```
jhawthorn: This is the opposite of the direction we usually have the "none means all" rule.

Normally we would have "shipping method with no store is available to all stores" not "a store with no shipping methods can use all shpiping methods"

jhawthorn: Apparently payment methods do the same

https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment_method.rb#L39-L42

I'm not really a fan of that either

adammathys: This is the direction we use for most of these types of associations. (As you found with payment methods.)

I agree that it's confusing and I'm also not a fan of the "none means all" pattern we've adopted, but it's probably best to at least keep them consistent.

core/db/migrate/20180202222641_create_store_shipping_methods.rb
```diff
@@ -0,0 +1,10 @@
+class CreateStoreShippingMethods < ActiveRecord::Migration[5.1]
+  def change
+    create_table :spree_store_shipping_methods do |t|
+      t.references :store, foreign_key: { to_table: "spree_stores" }
```
jhawthorn: I don't think this syntax is valid. It fails under MySQL with

```
Mysql2::Error: Can't create table `solidus_core_test`.`spree_store_shipping_methods` (errno: 150 "Foreign key constraint is incorrectly formed")
```

jhawthorn: Reading more, `to_table` does seem like a valid argument (I had thought it was not), but for some reason this isn't working.

jhawthorn: Fails on
```
   (60.4ms)  CREATE TABLE `spree_store_shipping_methods` (`id` bigint NOT NULL AUTO_INCREMENT PRIMARY KEY, `store_id` bigint, `shipping_method_id` bigint, `created_at` datetime NOT NULL, `updated_at` datetime NOT NULL,  INDEX `index_spree_store_shipping_methods_on_store_id`  (`store_id`),  INDEX `index_spree_store_shipping_methods_on_shipping_method_id`  (`shipping_method_id`), CONSTRAINT `fk_rails_1ec4e4430e`
FOREIGN KEY (`store_id`)
  REFERENCES `spree_stores` (`id`)
, CONSTRAINT `fk_rails_24cc4a3cd1`
FOREIGN KEY (`shipping_method_id`)
  REFERENCES `spree_shipping_methods` (`id`)
) ENGINE=InnoDB
```

jhawthorn: The issue is that this is creating an FK from a bigint to a normal size int.

Removed the FK entirely in in #2596

PR #2559, opened 2018-02-06
Add locale chooser to admin

jhawthorn:
I want to add a locale selector to the admin.

It's currently too hard to change the locale of the admin. My hope is that by making it easier we will have more users switch to a non-english locale and then through that hopefully get more help keeping them up to date.

I want to implement this in Solidus backend rather than solidus_i18n. This lets us avoid the need for `deface`, and makes it more clear that i18n is a concern of Solidus itself.

It would be nice if we could recommended that all new users include `solidus_i18n`, and listed the gem inside our README's "Getting started" section. To do this we would probably want it to only provide YAML files and there's been some work towards that.

## Todo

* [x] Have the selector actually set the locale
* [x] Currently this uses I18n.available_locales, we should probably restrict it to locales which have at least _some_ translations under the `spree` namespace.
* [x] Hide selector if there is only one available locale.

### DIFF
--- a/backend/app/views/spree/admin/shared/_locale_selection.html.erb
+++ b/backend/app/views/spree/admin/shared/_locale_selection.html.erb
@@ -1,0 +1,12 @@
+<div class="admin-locale-selection">
+  <select class="custom-select fullwidth">
+    <%=
+      options_for_select(
+        I18n.available_locales.map do |locale|
+          [I18n.t('spree.i18n.this_file_language', locale: locale, default: locale.to_s, fallback: false), locale]
+        end.sort,
+        selected: I18n.locale
+      )
+    %>
+  </select>
+</div>
--- a/backend/app/views/spree/admin/shared/_locale_selection.html.erb
+++ b/backend/app/views/spree/admin/shared/_locale_selection.html.erb
@@ -1,6 +1,6 @@
-<div class="admin-locale-selection">
-  <% available_locales = Spree.i18n_available_locales %>
-  <% if available_locales.size > 1 %>
+<% available_locales = Spree.i18n_available_locales %>
+<% if available_locales.size > 1 %>
+  <div class="admin-locale-selection">
     <select class="js-locale-selection custom-select fullwidth">
       <%=
         options_for_select(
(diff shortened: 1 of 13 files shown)


backend/app/views/spree/admin/shared/_locale_selection.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="admin-locale-selection">
```
kennyadsl: Maybe we should put this div into the `if available_locales.size > 1`? I think there's no reason to keep the html if empty.

core/spec/lib/i18n_spec.rb
```diff
@@ -5,9 +5,15 @@
 
 RSpec.describe "i18n" do
   before do
+    # This reload avoids an issue with I18n.available_locales being cached
```
kennyadsl: Do you have ideas about why this is needed only in this specs and not in other files that uses `I18n`?

backend/app/views/spree/admin/shared/_locale_selection.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <% available_locales = Spree.i18n_available_locales %>
+  <% if available_locales.size > 1 %>
+    <select class="js-locale-selection custom-select fullwidth">
+      <%=
+        options_for_select(
+          available_locales.map do |locale|
+            [I18n.t('spree.i18n.this_file_language', locale: locale, default: locale.to_s, fallback: false), locale]
```
kennyadsl: Have you considered using some more explicit string to mark locales available, like `active: true`? 
I know it's not a proper translation but how can a user easily understand that this key is what will make the translation file available?

tvdeyen: I think this is good for now and works with existing locales w/o any further changes.

backend/app/views/spree/admin/shared/_locale_selection.html.erb
jhawthorn: done.

core/spec/lib/i18n_spec.rb
jhawthorn: I guess the other files don't need a new locale to be autodetected. We should keep an eye out for new failures related to this, though

Issue #2460, opened 2017-12-16
Rails 5.2


jhawthorn:
Rails 5.2 is in beta, and we should get ready for it

Gems needing fixes:
* [ ] [polyamorous](https://github.com/activerecord-hackery/polyamorous) is broken
* [ ] [ransack](https://github.com/activerecord-hackery/ransack) is broken
* [x] acts_as_list has an open PR to fix deprecations swanandp/acts_as_list#294
* [x] [friendly_id](https://github.com/norman/friendly_id) has an open PR to fix deprecations norman/friendly_id#849
* [x] [awesome_nested_set](https://github.com/collectiveidea/awesome_nested_set) has an open PR to fix deprecations collectiveidea/awesome_nested_set#383
* [ ] [cancancan](https://github.com/cancancommunity/cancancan) is broken (as of rails 5.2.0.rc1)

Gems needing only gemspec dependency bumps:
* [ ] font-awesome-rails
* [ ] paranoia
* [ ] canonical-rails

Solidus:
* [x] #2465
* [x] #2563

halo:
I believe polyamorous has been fixed in https://github.com/activerecord-hackery/polyamorous/pull/28 and ransack has been fixed in https://github.com/activerecord-hackery/ransack/pull/868

damienlethiec:
Hi @jhawthorn 

Sorry to come back to you! Rails 5.2.0.rc2 is out (http://weblog.rubyonrails.org/2018/3/20/Rails-5-2-RC2/). Any news about the status of the gems Solidus depends on? Do you know when will we be able to use Solidus with Rails 5.2?

Thanks a ton :)

damienlethiec:
Hello @jhawthorn :)

Any news?

Thanks a lot!! :)

davideghz:
up

benbonnet:
just tried on 5.2, source in the gemfile from master; getting the following :

```
solidus was resolved to 2.6.0.alpha, which depends on
      solidus_core (= 2.6.0.alpha) was resolved to 2.6.0.alpha, which depends on
        actionpack (~> 5.1.0)
```

PR #2563, opened 2018-02-08
Preparations for Rails 5.2 (part 2)

jhawthorn:
There were some changes in 5.2.0.rc1, since the betas, which we needed to adjust for.

### DIFF
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -55,6 +55,11 @@ module DummyApp
     config.active_support.deprecation                 = :stderr
     config.secret_key_base                            = 'SECRET_TOKEN'
 
+    if config.active_record.sqlite3
+      # Rails >= 5.2
+      config.active_record.sqlite3.represent_boolean_as_integer = true
+    end
+
     # Avoid issues if an old spec/dummy still exists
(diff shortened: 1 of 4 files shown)


damienlethiec:
Hello @jhawthorn . Thanks for the great work. I am beginning a new solidus app with Rails 5.2 soon. Should I use the Github version of the gem for the moment? When with the 5.2 compatible version of Solidus released? Thank you :)

jhawthorn:
@damienlethiec Unfortunately there isn't yet a Rails 5.2 compatible version of Solidus yet. Biggest blocker right now is `cancancan`, which does not work on Rails 5.2 (see #2460).

damienlethiec:
Thanks a ton for the answer!! Any idea of when a compatible version could be ready (even if I understand you don't have any control over CanCanCan)

jhawthorn:
> Thanks a ton for the answer!! Any idea of when a compatible version could be ready (even if I understand you don't have any control over CanCanCan)

With this change it's just the gems we depend on (barring any changes between rc and final release). cancancan doesn't want to merge any changes before final release, so some time after that.

Solidus 2.5 will likely ship without Rails 5.2 support, but I would like to cut a Solidus 2.6 within a few weeks of the final release.

damienlethiec:
Great, thank you!

PR #2566, opened 2018-02-13
Revert "Add specs and convert Admin::Variant#destroy to discard"

jhawthorn:
Reverts solidusio/solidus#2520

Master started failing... did this break it?


jhawthorn:
Nope! Wasn't this.

PR #2567, opened 2018-02-13
Fix build maybe?

jhawthorn:


jhawthorn:
Fixed by #2568 instead.

PR #2572, opened 2018-02-14
Specify inheritance for Spree::Promotion

SamuelMartini:
it makes clear that Spree::Promotion inherit from Spree::Base and then
ActiveRecord.

With a promotion decorator present in the app, there are cases where calling a class method from an another gem on `Spree::Promotion` result in a `NoMethodError`. This is because the receiver has no superclass(just Object) and not Spree::Base as expected. It probably depend on which definition is loaded first.

The app has the following definitions:
```ruby
module Spree
  Promotion.foo
end
```

and a custom action defined like the default on solidus_core
```ruby
module Spree
  class Promotion
    module Actions
      class Custom
```
In some cases this results in a `NoMethodError` when `Promotion` receives `foo` and the same when it receives `count` or any other active record methods.
Adding `< Spree::Base` in the custom actions and rules in the app was enough to solve the issue, but I think that Spree::Promotion should inherit from Spree::Base in solidus_core as well.

### DIFF
--- a/core/app/models/spree/promotion/actions/create_adjustment.rb
+++ b/core/app/models/spree/promotion/actions/create_adjustment.rb
@@ -1,5 +1,5 @@
 module Spree
-  class Promotion
+  class Promotion < Spree::Base
     module Actions
       class CreateAdjustment < PromotionAction
         include Spree::CalculatedAdjustments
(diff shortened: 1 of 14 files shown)


jhawthorn:
Thank you

PR #2574, opened 2018-02-14
Add mutable false to default refund reason record

### LINKED ISSUE #2573: Refund reason seed creates an incorrect default record
The current seed which creates the default refund reason does not set `mutable: false`:

https://github.com/solidusio/solidus/blob/master/core/db/default/spree/refund_reasons.rb#L1

So [`Spree::RefundReason.return_processing_reason`](https://github.com/solidusio/solidus/blob/v2.4.0/core/app/models/spree/refund_reason.rb#L10) raises and `ActiveRecord::RecordNotFound` exception attempting to find the default refund reason filtering for `mutable: false`. This exception causes bad behaviours on reimbursements management, e.g. it makes it raise an error trying to visit an Original Payment reimbursement edit page, since `Spree::Reimbursement#simulate` method fails. 

### Steps to reproduce

```
git clone git@github.com:solidusio/solidus.git
cd solidus
bundle install
bundle exec rake sandbox
cd sandbox
bin/rails runner 'p Spree::RefundReason.return_processing_reason'
```

### Expected behavior

It should return the default refund reason created in seeds

### Actual behavior

It displays an `ActiveRecord::RecordNotFound` error

### System configuration
**Solidus Version**: master, 2.4.0

mdesantis:
Fixes #2573

### DIFF
--- a/core/db/default/spree/refund_reasons.rb
+++ b/core/db/default/spree/refund_reasons.rb
@@ -1,1 +1,1 @@
-Spree::RefundReason.find_or_create_by(name: "Return processing")
+Spree::RefundReason.find_or_create_by!(name: "Return processing", mutable: false)


kennyadsl:
In another PR we could remove this `mutable` thing, if it's just used to do not allow admin to change its name (since it has to be the same of the [constant defined here](https://github.com/solidusio/solidus/blob/9cac52900fadaf79b3157ecd3f0dc2016ae56528/core/app/models/spree/refund_reason.rb#L5)) in favor of a `default` boolean on the refund reason table. WDYT?

mdesantis:
If the `mutable`thing is used just for in order to mark the default record yes, it shouldn't be much more complex than rename `mutable` to `default` and swap the values; if otherwise the `mutable` thing is there to represent some separated logic we could add a `default` column in order to manage this record and leave the `mutable` thing alone.

jhawthorn:
Thank you!

PR #2579, opened 2018-02-15
Add options to PromotionCode::BatchBuilder and spec for unique promotion code contention

jhawthorn:
The issue fixed by #2578 wasn't caught by our specs. It was hard to write a spec to test that issue because the `PromotionCode::BatchBuilder` was configured by `class_attribute`s.

This PR deprecates the existing class-attributes in favour of having a defaults option hash.

It uses this to add an additional spec which runs the BatchBuilder with settings that make it extremely likely that there will a conflict: from a space of 100 possible codes (`00` to `99`) we generate 50.

An online birthday problem calculator tells me this is 99.99997% likely that the generator will encounter at least one conflict. That seems plenty. I haven't verified this, but intuitively it seems very likely.

This would have caught the issue prior to #2578 
```
Failures:

  1) Spree::PromotionCode::BatchBuilder#build_promotion_codes with likely code contention creates the correct number of codes
     Failure/Error: expect(promotion.codes.size).to eq(number_of_codes)

       expected: 50
            got: 61

       (compared using ==)
     # ./spec/models/spree/promotion_code/batch_builder_spec.rb:71:in `block (4 levels) in <top (required)>'
```

### DIFF
--- a/core/app/models/spree/promotion_code/batch_builder.rb
+++ b/core/app/models/spree/promotion_code/batch_builder.rb
@@ -1,15 +1,32 @@
 class ::Spree::PromotionCode::BatchBuilder
-  attr_reader :promotion_code_batch
+  attr_reader :promotion_code_batch, :options
   delegate :promotion, :number_of_codes, :base_code, to: :promotion_code_batch
 
-  class_attribute :random_code_length, :batch_size, :sample_characters, :join_characters
-  self.random_code_length = 6
-  self.batch_size = 1_000
-  self.sample_characters = ('a'..'z').to_a + (2..9).to_a.map(&:to_s)
-  self.join_characters = "_"
+  DEFAULT_OPTIONS = {
+    random_code_length: 6,
+    batch_size: 1000,
+    sample_characters: ('a'..'z').to_a + (2..9).to_a.map(&:to_s),
+    join_characters: "_"
+  }
 
-  def initialize(promotion_code_batch)
+  [:random_code_length, :batch_size, :sample_characters, :join_characters].each do |attr|
+    define_singleton_method(attr) do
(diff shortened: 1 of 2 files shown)


jhawthorn:
Failure should be fixed by #2580

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -2,14 +2,30 @@ class ::Spree::PromotionCode::BatchBuilder
...
+  [:random_code_length, :batch_size, :sample_characters, :join_characters].each do |attr|
+    define_singleton_method(attr) do
+      Spree::Deprecation.warn "#{name}.#{attr} is deprecated. Use #{name}::DEFAULT_OPTIONS[:#{attr}] instead"
+      DEFAULT_OPTIONS[attr]
+    end
+
+    define_singleton_method(:"#{attr}=") do |val|
+      Spree::Deprecation.warn "#{name}.#{attr}= is deprecated. Use #{name}::DEFAULT_OPTIONS[:#{attr}]= instead"
+      DEFAULT_OPTIONS[attr] = val
+    end
```
tvdeyen: 👍nice

core/app/models/spree/promotion_code/batch_builder.rb
```diff
@@ -2,14 +2,30 @@ class ::Spree::PromotionCode::BatchBuilder
...
+
+    delegate attr, to: self
+  end
+
+  def initialize(promotion_code_batch, options = {})
     @promotion_code_batch = promotion_code_batch
+    @options = DEFAULT_OPTIONS.merge(options)
```
tvdeyen: Not necessary, but could we use a attr_reader for @options? Consistency

kennyadsl: Agree with @tvdeyen , `attr_reader` for `@options` will make this class more consistent.

PR #2581, opened 2018-02-17
Add custom authentication (User model) setup article

benjaminwil:
I wrote an article outlining the requirements for setting up a custom `User` model or some non-`solidus_auth_devise` authentication solution. I followed my own instructions while setting up the [Passwordless](https://github.com/mikker/passwordless) gem with decent results. 👍 

### DIFF
--- a/guides/users/custom-authentication.md
+++ b/guides/users/custom-authentication.md
@@ -1,0 +1,172 @@
+# Custom authentication
+
+*You can use the official [`solidus_auth_devise`][solidus-auth-devise] gem
+to provide a `Spree::User` model and basic authentication for Solidus. See its
+documentation for additional setup instructions.*
+
+Stores require a `User` model in order to take advantage of all of Solidus's
+features. This model can have any name, and Solidus can integrate with your
+existing Rails application's existing `User` model.
+
+By default, Solidus provides a [`Spree::LegacyUser` model][legacy-user] that
+offers the bare minimum functionality of a user. The model is intended to be
+modified by extensions.
+
+The rest of this article outlines the steps required should you decide to create
+a `User` model from scratch, use an authentication solution like
+[Devise][devise], or integrate your application's existing `User` model.
+
+Note that while your user model can have whatever name you like, this article
+uses the model `MyStore::User` for its examples. 
+
+[legacy-user]: https://github.com/solidusio/solidus/blob/master/core/app/models/spree/legacy_user.rb
+[solidus-auth-devise]: https://github.com/solidusio/solidus_auth_devise
+
+## Set the `Spree.user_class`
+
+No matter what gem or extension you use for your store's `User` model, your
+application's `Spree.user_class` value needs to be set. By default, Solidus sets
+the `Spree.user_class` to `Spree::LegacyUser`. 
+
+You can configure a custom `Spree.user_class` in your application's
+`/config/initializers/spree.rb` file:
+
+```ruby
+# /config/initializers/spree.rb
+Spree.user_class = "MyStore::User"
+```
+
+If you use the `solidus_auth_devise` gem, your `Spree.user_class` is set to
+`Spree::User` when you run the gem's `solidus:auth:install` generator.
+
+## Prepare your user model for Solidus
+
+Once you have set the `Spree.user_class`, you can start integrating with the
+features that are associated with the `user_class`.
+
+### Custom user generator
+
+After you have created your custom `User` model and its corresponding database
+table, you can use the `spree:custom_user` generator to generate Solidus's
+required `User` table columns and some authentication helpers:
+
+```shell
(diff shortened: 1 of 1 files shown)


guides/users/custom-authentication.md
```diff
@@ -0,0 +1,172 @@
...
+features. This model can have any name, and Solidus can integrate with your
+existing Rails application's existing `User` model.
+
+By default, Solidus provides a [`Spree::LegacyUser` model][legacy-user] that
+offers the bare minimum functionality of a user. The model is intended to be
+modified by extensions.
+
+The rest of this article outlines the steps required should you decide to create
+a `User` model from scratch, use an authentication solution like
+[Devise][devise], or integrate your application's existing `User` model.
```
tvdeyen: Missing link

guides/users/custom-authentication.md
```diff
@@ -0,0 +1,172 @@
...
+
+```shell
+bundle exec rails db:migrate
+```
+
+If you use the `spree:custom_user` generator:
+
+- The `Spree.user_class` is updated to your specified class.
+- Authentication helpers are set up for the `solidus_frontend` and
+  `solidus_backend` views and are sent to the `Rails::ApplicationController`,
```
tvdeyen: The `ApplicationController` is not nested inside the `Rails` module

guides/users/custom-authentication.md
```diff
@@ -0,0 +1,172 @@
...
+to provide a `Spree::User` model and basic authentication for Solidus. See its
+documentation for additional setup instructions.*
+
+Stores require a `User` model in order to take advantage of all of Solidus's
+features. This model can have any name, and Solidus can integrate with your
+existing Rails application's existing `User` model.
+
+By default, Solidus provides a [`Spree::LegacyUser` model][legacy-user] that
+offers the bare minimum functionality of a user. The model is intended to be
+modified by extensions.
```
tvdeyen: This user is only meant as test user and should not be modified by extensions. Also it is not suitable for production environments.

guides/users/custom-authentication.md
tvdeyen: Also this article does not handle the usage of Devise. Do you plan to extend this article? If not this paragraph should be removed as this confuses

benjaminwil: I just ran through the requirements in this article with a `User` model that I set up with Devise. If someone was already using Devise for their user model they could use this article to integrate it with Solidus.

Why do you feel that mentioning Devise here as an example confuses?

guides/users/custom-authentication.md
benjaminwil: Maybe [this](https://github.com/solidusio/solidus/blob/9f5c53aee433fa0e72ae97301a161192e147132e/core/app/models/spree/legacy_user.rb#L4-L5) is just a very old comment then!

```
  # Default implementation of User.
  #
  # @note This class is intended to be modified by extensions (ex.
  #   spree_auth_devise)
```
When I edit this article in my next commits, I'll be sure to update this comment as well.

tvdeyen: Thanks!

guides/users/custom-authentication.md
tvdeyen: Ah, ok, I misread the comma and the "or". Makes sense. Sorry for the confusion.

guides/users/custom-authentication.md
```diff
@@ -0,0 +1,174 @@
...
+  `Rails::ApplicationController` and the `Spree::Api::BaseController`, making it
+  available throughout your application.
+
+### Minimum requirements
+
+Solidus requires that your `User` model's database table includes at least the
+following columns:
+
+- `spree_api_key`: A string with a user's API key. This should be limited to 48
+  characters (`:limit => 48`).
```
kennyadsl: is the text between parenthesis really needed?

guides/users/custom-authentication.md
```diff
@@ -0,0 +1,174 @@
...
+It also requires that you have a [`spree_current_user`](#spree-current-user)
+helper method.
+
+#### User passwords 
+
+Note that if you use the stock `solidus_frontend` or `solidus_backend` gems,
+your user should also have a `password` column. You can set up a password column
+however you see fit.
+
+### `spree_current_user` <span id="spree-current-user"></span>
```
kennyadsl: should we say that this code (`spree_current_user` and authentication helpers)  goes into a controller? Or is it too obvious?

guides/users/custom-authentication.md
benjaminwil: :joy: Probably not. Thanks for catching that.

guides/users/custom-authentication.md
benjaminwil: No, I think that I should mention that. Thank you!

PR #2582, opened 2018-02-18
Distribute over eligible line items

Sinetheta:
There's a surprising interaction between the distributed amount calculator and the product promotion rule. The entire discount amount is distributed among _all_ of an order's line_items, but only adjustments on line_items that match the promotion rule are considered "actionable" and contribute towards the order promo total.

https://github.com/solidusio/solidus/blob/master/core/app/models/spree/promotion/actions/create_item_adjustments.rb#L32

This means that as you add products to your cart, your discount can actually go _down_, which I don't imagine is a behaviour anyone would want.

screen shot 2018-02-18 at 10 11 59 am

Here the total discount has dropped from the preferred amount of $10 to ($22.99 / $38.98) = $5.90 the portion of the discount which was distributed to the item on sale.

I think there's larger discussion to be had about line item adjustments, but for now I'd just like to fix this one calculator. If only "actionable" line items can receive an adjustment then we need to distribute the desired amount among only those line items.

screen shot 2018-02-18 at 10 41 00 am

### DIFF
--- a/core/app/models/spree/distributed_amounts_handler.rb
+++ b/core/app/models/spree/distributed_amounts_handler.rb
@@ -1,17 +1,16 @@
 module Spree
   class DistributedAmountsHandler
-    attr_reader :line_item, :order, :promotion, :total_amount
+    attr_reader :line_items, :total_amount
 
-    def initialize(line_item, promotion, total_amount)
-      @line_item = line_item
-      @order = line_item.order
-      @promotion = promotion
+    def initialize(line_items, total_amount)
+      @line_items = line_items
       @total_amount = total_amount
     end
 
-    # @return [BigDecimal] the weighted adjustment for the initialized line item
-    def amount
-      distributed_amounts[@line_item.id].to_d
+    # @param [LineItem] one of the line_items distributed over
+    # @return [BigDecimal] the weighted adjustment for this line_item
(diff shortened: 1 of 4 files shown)


adammathys:
I don't believe this is the best way to fix this issue. Now that the handler is reliant on knowing about the promotion, there's little purpose for it being its own class. It is directly tied to the calculator and to only ever being used for promotions. 

A better approach would be passing the handler the list of items and an amount to distribute. Then updating the the interface to query for the amount that should be applied to a specific item.

tvdeyen:
@adammathys I agree, but was hesitant to refuse the PR because of that. But on second thought we should think about a solution to only action on eligible items.

core/app/models/spree/distributed_amounts_handler.rb
```diff
@@ -1,17 +1,16 @@
 module Spree
   class DistributedAmountsHandler
-    attr_reader :line_item, :order, :promotion, :total_amount
+    attr_reader :distributed_amounts, :line_items, :total_amount
```
kennyadsl: I think we don't need this `distributed_amounts`  reader method here

Sinetheta: 👍 thanks

PR #2583, opened 2018-02-19
Use solidus_auth_devise GitHub master for sandbox

tvdeyen:
We need at least the 2.1 gem release of solidus auth devise, because we dropped the deface dependency in Solidus 2.5.0.
    
Without this might break the sandbox (the build in demo app), if people have an older version of solidus auth devise installed.

### DIFF
--- a/lib/sandbox.sh
+++ b/lib/sandbox.sh
@@ -36,7 +36,7 @@ fi
 
 cd ./sandbox
 echo "gem 'solidus', :path => '..'" >> Gemfile
-echo "gem 'solidus_auth_devise'" >> Gemfile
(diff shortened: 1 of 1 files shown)


jhawthorn:
> The current stable gem release of solidus auth devise fails with
> current solidus core master, because we dropped the deface dependency.

This shouldn't be the case. Version `2.1.0` is the same as current master

tvdeyen:
Somehow this is now working for me. 🤔 

I guess because I had `solidus_auth_devise` 2.0 installed it fulfilled bundlers needs, but failed because we do not have `deface` installed anymore. Now that I have `solidus_auth_devise` 2.1 installed it works like expected. 

We should consider to add `>= 2.1` to the sandbox's `Gemfile`

wdyt @jhawthorn ?

PR #2586, opened 2018-02-19
Add frozen_string_literal: true to all files

jhawthorn:
Also remove any encoding: UTF-8 comments which should no longer have any effect.


jhawthorn:
> Nice. Out of curiosity did you test how much memory actually gets saved on a regular store?

I haven't, and I'm not sure what exactly I would measure (maybe run stack_prof or memory_prof on a full checkout?). I'm not expecting a major improvement (but would love to be surprised by one).

PR #2588, opened 2018-02-20
Allowing admin to show all countries regardless checkout zone

### LINKED ISSUE #2562: Admin should not restrict countries to checkout_zone
From @softr8 in `#solidus` slack

> I need to create some stock locations with a different country than the one available for the storefront(Spree::Config.checkout_zone), is this some special case for me or this is something that would be more general?

This seems like a bug. In both frontend and backend we use [available_countries](https://github.com/solidusio/solidus/blob/2fcabc9518175cddebd6076fea3a7f914e702022/core/app/helpers/spree/base_helper.rb#L107-L120) to build a list of countries.

There's a `Spree::Config.checkout_zone` setting, which is intended to limit the countries which can be chosen on the frontend. Because this helper is shared and also used on the admin, it also limits the countries which can be seen and selected there.

We should change this so that the admin can always select all countries.

softr8:
Fixes #2562

### DIFF
--- a/core/app/helpers/spree/base_helper.rb
+++ b/core/app/helpers/spree/base_helper.rb
@@ -106,8 +106,8 @@ module Spree
       end
     end
 
-    def available_countries
-      checkout_zone = Zone.find_by(name: Spree::Config[:checkout_zone])
+    def available_countries(restrict_to_zone: Spree::Config[:checkout_zone])
+      checkout_zone = Zone.find_by(name: restrict_to_zone)
 
       if checkout_zone && checkout_zone.kind == 'country'
         countries = checkout_zone.country_list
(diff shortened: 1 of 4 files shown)


core/app/helpers/spree/base_helper.rb
```diff
@@ -106,8 +106,8 @@ def taxons_tree(root_taxon, current_taxon, max_level = 1)
       end
     end
 
-    def available_countries
-      checkout_zone = Zone.find_by(name: Spree::Config[:checkout_zone])
+    def available_countries(zone_name = Spree::Config[:checkout_zone])
```
tvdeyen: I would like to have a more meaningful parameter here. Just passing `nil` to restrict the countries to a specific zone seams unintuitive to me.

Maybe something like?

```rb
def available_countries(restrict_to_zone: Spree::Config[:checkout_zone])
```

softr8:
Done! let me know if it still makes sense.

Thanks!

jhawthorn:
Thank you!

PR #2589, opened 2018-02-20
Fix admin payment actions style table issue

jtapia:
### What it does?
- Fix for admin tables on resolutions < 1200px

### Preview
http://recordit.co/hR8Qjd0k61

### Reference
- https://v4-alpha.getbootstrap.com/content/tables/

### DIFF
--- a/backend/app/views/spree/admin/payments/_list.html.erb
+++ b/backend/app/views/spree/admin/payments/_list.html.erb
@@ -1,4 +1,4 @@
-<table class="index" id='payments' data-order-id='<%= @order.number %>'>
+<table class="index table-responsive" id='payments' data-order-id='<%= @order.number %>'>
   <colgroup>
     <col style="width: 15%"/>
     <col style="width: 15%"/>


jtapia:
@tvdeyen I just changed the approach, what do you think?

backend/app/views/spree/admin/payments/_list.html.erb
```diff
@@ -1,4 +1,4 @@
-<table class="index" id='payments' data-order-id='<%= @order.number %>'>
+<table class="index table table-sm table-responsive" id='payments' data-order-id='<%= @order.number %>'>
```
tvdeyen: Could we not change the table style here? Won’t adding just `table-responsive` be enough to fix the issue?

jtapia: I just tested it and yes, we can use only `table-responsive`, I sent the changes https://github.com/solidusio/solidus/pull/2589/files

PR #2591, opened 2018-02-21
Documentation touchup

jormon:
Links were confusingly switched: `backend` link was going to `frontend` and vice-versa.


benjaminwil:
Thank you. :clap:

PR #2595, opened 2018-02-22
Revert "Add association between stores and shipping"

jhawthorn:
Reverts solidusio/solidus#2557


jhawthorn:
Revert avoided by #2596

PR #2597, opened 2018-02-23
Build mysql on circle ci

tvdeyen:
Although most of our users use postgresql nowadays, we still support
mysql as well. We want to build for mysql as well, so we can see potential
issues before they happen.


tvdeyen:
Closing in favour of #2599

PR #2607, opened 2018-02-28
Fix spec failure due to DB order

jhawthorn:
The shipments association here was being loaded in the recalculate. The association itself isn't ordered, so the shipments could be in any order. Because of this, shipments.last wasn't necessarily the most recent shipment, and this spec could fail.

This probably started failing after #2555 was merged

### DIFF
--- a/core/spec/models/spree/shipment_spec.rb
+++ b/core/spec/models/spree/shipment_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe Spree::Shipment, type: :model do
           end
         end.to change { Spree::Shipment.count }.by(1)
 
-        new_shipment = order.shipments.last
+        new_shipment = order.shipments.order(:created_at).last
         expect(new_shipment.number).to_not eq(shipment.number)
         expect(new_shipment.stock_location).to eq(stock_location)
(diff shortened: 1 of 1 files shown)


core/spec/models/spree/shipment_spec.rb
```diff
@@ -39,7 +39,7 @@
           end
         end.to change { Spree::Shipment.count }.by(1)
 
-        new_shipment = order.shipments.last
+        new_shipment = (order.shipments.to_a - [shipment])[0]
```
tvdeyen: Why not `order.shipments.order(:created_at).last` or `order.shipments.sort_by(&:created_at).last` if we don't want to reload the shipments again? Would be more readable.

jhawthorn: I did this out of habit due to MySQL second timestamp issues, but this isn't necessary as of #2598 :tada:. I will use your suggestion.

tvdeyen: Ah. Thanks for clarifying. I always forget the darn mysql default precision.

tvdeyen:
Merging ahead of our usual time period, as this fixes spec issues.

PR #2608, opened 2018-03-01
Test eslint on hound attempt 2

jhawthorn:
Hound still doesn't always run on our repo, so I'm not sure the odds of this working.


jhawthorn:
Hound still doesn't work :disappointed:

PR #2609, opened 2018-03-01
adding link to product in backend order summary

loicginoux:
That won't be the PR of the month.... I am just adding an handy link back to the product when you are in the backend order summary.

screenshot 2018-03-01 17 07 09

### DIFF
--- a/backend/app/views/spree/admin/orders/_carton_manifest.html.erb
+++ b/backend/app/views/spree/admin/orders/_carton_manifest.html.erb
@@ -4,7 +4,7 @@
       <%= render 'spree/admin/shared/image', image: item.variant.display_image, size: :mini %>
     </td>
     <td class="item-name">
-      <%= item.variant.product.name %><br><%= "(" + variant_options(item.variant) + ")" unless item.variant.option_values.empty? %>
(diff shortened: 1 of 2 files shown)


backend/app/views/spree/admin/orders/_carton_manifest.html.erb
```diff
@@ -4,7 +4,7 @@
       <%= render 'spree/admin/shared/image', image: item.variant.display_image, size: :mini %>
     </td>
     <td class="item-name">
-      <%= item.variant.product.name %><br><%= "(" + variant_options(item.variant) + ")" unless item.variant.option_values.empty? %>
+      <%= link_to item.variant.product.name, edit_admin_product_path(item.variant.product), target: "_blank" %><br><%= "(" + variant_options(item.variant) + ")" unless item.variant.option_values.empty? %>
```
jhawthorn: Looks great, but I'd rather not use `target: "_blank"` here. Links should open in the same tab.

loicginoux: Ok, I think when someone works on an order he will want to keep it open while looking at a product details quickly but doesn't matter. I changed it.

PR #2610, opened 2018-03-04
More flexible frontend JS includes

brchristian:
Instead of requiring individual checkout JS files by name, we can require the `/checkout` directory instead. This should have the effect of making it easier for extensions or individual stores to add JS files without needing to override `frontend.js` or `frontend/checkout.js` to make sure that those files are included.

### DIFF
--- a/frontend/app/assets/javascripts/spree/frontend/checkout.js
+++ b/frontend/app/assets/javascripts/spree/frontend/checkout.js
@@ -1,8 +1,6 @@
 //= require jquery.payment
 //= require_self
-//= require spree/frontend/checkout/address
-//= require spree/frontend/checkout/payment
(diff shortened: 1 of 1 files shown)


brchristian:
@tvdeyen Oh, you are completely correct! My apologies and thanks for the heads up. I have updated the PR accordingly.

tvdeyen:
@bbuchalter maybe I was not clear enough. But your PR is not doing what you would expect it does. We need to keep it as it is right now. I am sorry, but this is not "fixable" by us. This needs to be fixed in Sprockets.

brchristian:
@tvdeyen Sorry for any confusion. Closing this!

tvdeyen:
@brchristian no need to be sorry. The goal is very noble and it would be great to have a solution for this.

PR #2621, opened 2018-03-09
The shipment of a partial expedition should return to :ready

sebfie:
Hello,

On my solidus website, we need to ship order’s shipments partially, after sending the first part of a shipment, the shipment stays on state `:pending` instead of `:ready`. We think it should be in state `ready`  because the button “ship” does not appear and we cannot send the rest  of the items

This commits fixed my issue.

The problem seems to be in `can_transition_from_pending_to_ready?` because to transition you check that all  inventory_units are allowed to be ship even for the products already sent so you cannot transition to ready

I think we should only check for inventory units on state `:on_hand` to correct the case of sending partial shipments

Maybe am I missing something?

Thx for you comments

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -82,7 +82,7 @@ module Spree
 
     def can_transition_from_pending_to_ready?
       order.can_ship? &&
-        inventory_units.all? { |iu| iu.allow_ship? || iu.canceled? } &&
+        inventory_units.on_hand.all? { |iu| iu.allow_ship? || iu.canceled? } &&
         (order.paid? || !Spree::Config[:require_payment_to_ship])
     end
 
--- a/core/spec/models/spree/order_shipping_spec.rb
+++ b/core/spec/models/spree/order_shipping_spec.rb
@@ -208,6 +208,10 @@ RSpec.describe Spree::OrderShipping do
         expect { subject }.to change { unshipped_inventory.map(&:reload).map(&:state) }.from(['on_hand']).to(['shipped'])
       end
 
+      it "shipment should be ready" do
+        expect(shipment.reload.state).to eq("ready")
+      end
+
       it "creates a carton with the shipment's inventory units" do
         expect { subject }.to change { order.cartons.count }.by(1)
         expect(subject.inventory_units).to match_array(unshipped_inventory)
--- a/core/spec/models/spree/shipment_spec.rb
+++ b/core/spec/models/spree/shipment_spec.rb
@@ -81,6 +81,7 @@ RSpec.describe Spree::Shipment, type: :model do
 
     it 'returns pending if backordered' do
       allow(shipment).to receive_messages inventory_units: [mock_model(Spree::InventoryUnit, allow_ship?: false, canceled?: false)]
+      allow(shipment.inventory_units).to receive_messages(on_hand: shipment.inventory_units)
       expect(shipment.determine_state(order)).to eq 'pending'
     end
 
@@ -301,6 +302,7 @@ RSpec.describe Spree::Shipment, type: :model do
         shipment.update_attributes!(state: 'ready')
 
         allow(shipment).to receive_messages(inventory_units: [mock_model(Spree::InventoryUnit, allow_ship?: false, canceled?: false)])
+        allow(shipment.inventory_units).to receive_messages(on_hand: shipment.inventory_units)
         expect(shipment).to receive(:update_columns).with(state: 'pending', updated_at: kind_of(Time))
         shipment.update_state
       end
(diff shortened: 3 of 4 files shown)


kennyadsl:
Have you considered splitting the shipment to reflect the partial shipments made?

sebfie:
No, I just want to send severals cartons for my shipment.

kennyadsl:
Sorry, I can't get if shipping only part of the shipment without splitting is possible using solidus default functionalities or it is something you implemented in your store?

sebfie:
In fact, In my code i just use: 

`
Spree::OrderShipping.new(order).ship(
          inventory_units: inventory_units,
          stock_location: stocklocation,
          address: order.shipping_address,
          shipping_method: order.shipments.first.shipping_method,
          shipped_at: Time.now,
          external_number: "test_number",
          tracking_number: "12345",
          suppress_mailer: true
        )
`

You can specify inventory_units you want to ship, so I suppose you can send a shipment on multiple cartons? Is not why carton exists?

isindexer:
Hi
I am working with sebfie on the business side. 
The problem when using the split functionnality is that the total amount of the order is recalculated. If a promotion has ended during the picking packing shipping process, it leads to a gap between the original amount, the payment and the recalculated amount (payment_state = balance_due or credit_owned). 
Creating as many cartons as needed to fully shipped the order prevents us from this problem. But we need the shipment to be  ready for that. 
Hope it helps to understand the aim of this PR

core/app/models/spree/shipment.rb
```diff
@@ -82,7 +82,7 @@ def can_transition_from_pending_to_shipped?
 
     def can_transition_from_pending_to_ready?
       order.can_ship? &&
-        inventory_units.all? { |iu| iu.allow_ship? || iu.canceled? } &&
+        inventory_units.on_hand.all? { |iu| iu.allow_ship? || iu.canceled? } &&
```
kennyadsl: I think that this won't work if the current shipment only has inventory units in a different state, eg `backordered`. In this case `all?` returns true and it will incorrectly move the shipment to ready

kennyadsl:
Sure, now I got it, thanks for the detailed background of this PR. 
You are right, this is the scenario carton was made for, thanks for pointing out the issue.

A question: when you hit the `ship` button on an already partially shipped shipment, will solidus only ship the rest of the shipping correctly understanding what has already been shipped, or do you need extra code to make it work?

isindexer:
Correct me if i am wrong Sebfie but we don’t need extra-code to ship the rest of the shipping. 
To be honnest, we never use the ´ship´ button. We only use a worker

core/app/models/spree/shipment.rb
loicginoux: Thats what I though too but we actually used on_hand inventory units only because we saw this method from inventory_unit.rb:

```
def allow_ship?
     on_hand?
   end
```

kennyadsl:
Ok, shipping the rest of the order programmatically makes sense, but it should work without any code even via Admin UI since [here](https://github.com/solidusio/solidus/blob/daea52ff1ce4f860bd061b81f62804df2c229234/core/app/models/spree/order_shipping.rb#L20) it ships only `shippable` inventory units, which should filter out what you previously shipped.

core/app/models/spree/shipment.rb
sebfie: I saw and fixed some rspec does not break by this code

sebfie:
Yes, but to be able to send the shippable, the shipment should have the state "ready"

core/app/models/spree/shipment.rb
kennyadsl: I think that with `InventoryUnit.pre_shipment` we'll also cover `backorders` inventory units

sebfie: So we should change the `def allow_ship?` method

kennyadsl: No, I meant changing `inventory_units.on_hand.all? {}` into `inventory_units.pre_shipment.all? {}` but I'm still not sure we'll cover all scenarios since we'll filter out all `canceled` inventory units (making the `|| iu.canceled?` check in the block useless). I need to understand better what that code does, and which kind of inventory units we expect to have when the shipment is ready.

sebfie: So I wait your feedback before to add some tests?

kennyadsl: yes, please let me (and/or other users as well) think a bit about this, thanks again!

kennyadsl:
I've made #2622 with what I've found, I think it's better to discuss there before changing other code to be sure we are going in the right direction and we are not missing anything. Thanks again!

PR #2624, opened 2018-03-09
Rewrite payment processing documentation; add Spree::Order#payment_states documentation

benjaminwil:
I have rewrote the `payments/payment_processing.md` article and removed/moved some of the content from it.

- The new article goes more into detail about the `authorize!` and `purchase!` methods used for payment processing.
- It moves information about payment states (on `Spree::Order`s) to a new article in the future `orders/` section of the documentation.
- It removes information about logging. Later, I will reintroduce documetation about logging in a new section.
- I made more linear descriptions of what is involved in Solidus's processing process.

This is a pretty dense article. Payment processing is confusing. In the future, I'd like to take this article in a more tutorialized direction set up a payment method that uses a real payment service provider like Stripe.


guides/orders/payment-states.html.markdown
```diff
@@ -0,0 +1,37 @@
+# Payment states
+
+<!-- TODO:
+  This article is a stub. It may need to be merged with another orders article
+  in the future.
+-->
+
+You can check the state of the `Spree::Payment` associated with any order:
```
gmacdougall: There are two ways that the state of payments are represented. There is the state of the payment status for the order, which is what is described below and the state of an individual payment.

As an example:

An order for $100 has a single payment of $50 that is complete. The status of the one payment is complete, but the order is in a payment state of balance_due.

If the order has multiple payments:

- $80 - failed
- $80 - completed
- $80 - completed

The order will be in a payment_state of credit owed.

Refunds play in to this as well as those can bring the payment state back in to a state of pair if the customer overpaid in the above case. This is complicated enough that it needs an explanation of what the difference between order `payment_state` and payment's `state` is.

benjaminwil: I am going to include an example like this, thank you.

guides/orders/payment-states.md
```diff
@@ -0,0 +1,94 @@
...
+A `Spree::Order` can have one of the following `payment_state`s:
+
+The following is a list of the available payment states:
+
+- `balance_due`: This state indicates that a payment on the order is required.
+- `failed`: This state indicates that at least one payment for the order failed.
+- `credit_owed`: This state indicates that the order has been paid for in excess
+  of its total.
+- `paid`: This state indicates that the order has been paid for in full. All of
+  the `Spree::Payment`s on the order have been satisfied.
```
gmacdougall: An order can also be in the state `void`

guides/orders/payment-states.md
```diff
@@ -0,0 +1,94 @@
...
+For more information about `Spree::Payment`s and the Solidus payments system,
+see the [Payments][payments] documentation.
+
+[payments]: ../payments/overview.md
+[payment-sources]: ../payments/payment-sources.md
+
+## Completed payments
+
+The `Spree::Order`'s `payment_state` can only be `paid` if each `Spree::Payment`
+has a state of `completed`.
```
gmacdougall: This is not true.

guides/payments/payment-processing.md
```diff
@@ -0,0 +1,167 @@
...
+the way the `Spree::PaymentMethod` is configured.
+
+In summary, the `process!` method processes the payment in one of the following
+ways:
+
+- If the `Spree::PaymentMethod`'s `auto_capture` attribute is set to `true`,
+  `purchase!` is called, meaning the payment is authorized and captured. This
+occurs even if the payment's state is already `completed`.
+- If the `Spree::PaymentMethod`'s `auto_capture` attribute is set to `false`,
+  then the payment is authorized but not captured. This occurs even if if the
```
DanielePalombo: `if` repeted twice

guides/payments/payment-processing.md
```diff
@@ -0,0 +1,167 @@
...
+## Interfacing with a payment service provider
+
+When you set up your own payment methods, you need it to integrate with a payment
+service provider. While you can create an integration in any way that you want,
+you may want to note the following conventions set up by Solidus's
+`Spree::PaymentMethod` and `Spree::Payment::Processing` classes:
+
+- `Spree::PaymentMethod` has a similar interface to API provided by
+  [active_merchant][active-merchant].
+- The `Spree::PaymentMethod#gateway` method uses the
```
DanielePalombo: Maybe I'm wrong, but seems like `gateway` method returns an instance of this payment method. Gateway_options is the variable provided to the gateway class.

benjaminwil: Thank you for making a note of this. While the `options` value is created by the `Spree::Payment::Processing#gateway_options` the way that I wrote this didn't really make sense.

I'm going to simplify it shortly and leave explanations to the "Gateway options" section below.

PR #2626, opened 2018-03-13
Make frontend's LocaleController compatible with solidus_i18n

jhawthorn:
Written on top of #2559

This adjusts the behaviour of frontend's `LocaleController#set` so that it can be used by `solidus_i18n` without needing to make any changes to the controller.

Specifically, this plays nicely with `filter :locale` in routes from the `routing-filter` gem, which is included by solidus_i18n.

### DIFF
--- a/frontend/app/controllers/spree/locale_controller.rb
+++ b/frontend/app/controllers/spree/locale_controller.rb
@@ -13,7 +13,8 @@ module Spree
       else
         flash[:error] = t('spree.locale_not_changed')
       end
-      redirect_back_or_default(spree.root_path)
+
+      redirect_to spree.root_path
     end
   end
 end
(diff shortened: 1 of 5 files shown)


jhawthorn:
> Only works currently with solidusio-contrib/solidus_i18n#114

It won't cause any problems without that merged (it just won't do anything). This change will allow us to remove the decorator entirely from solidus_i18n :tada:

tvdeyen:
> This change will allow us to remove the decorator entirely from solidus_i18n 

solidus_18n also takes the preferred_available_locales is the current store into account. What this PR is not. Do you plan to support this feature as well in core?

jhawthorn:
> solidus_18n also takes the preferred_available_locales is the current store into account. What this PR is not. Do you plan to support this feature as well in core?

#2627 :wink: 

My implementation is a bit different and has different attribute names.

Issue #1654, opened 2016-12-16
Duplicate Variant shown with Variant Having multiple product


flyfy1:
## Setup:

- define TaxCategory to have a TaxRate with tax included in price
- create product with multiple Variant

## How to see the issue:

In the SolidusFrontend, view the product

## ScreenShot:

![image](https://cloud.githubusercontent.com/assets/1553760/21250974/82b8847e-c384-11e6-8bb0-1ab3dfbac690.png)

flyfy1:
@mamhoff What's your idea about this issue?

cbrunsdon:
@flyfy1 can you paste a snippet or something of some factory code that would exactly expose this problem on the sandbox? Would help a lot for debugging purposes.

isindexer:
Hi @flyfy1 
Did you solve this issue ?

seand7565:
@flyfy1 @mamhoff Having trouble replicating this issue. I've done the following: 

Changed the tax rate to "include in price"

![image](https://cloud.githubusercontent.com/assets/5720486/26301141/2ca7bd3c-3ed8-11e7-9819-4cebacf9e36d.png)

Created a product with multiple variants and assigned it the tax category

![image](https://cloud.githubusercontent.com/assets/5720486/26301178/45c0b256-3ed8-11e7-9b09-0503834933f9.png)

However I see the correct amount of variants on the product page:

![image](https://cloud.githubusercontent.com/assets/5720486/26301232/73923e8e-3ed8-11e7-86b4-f732aa6ada77.png)


This issue was from 2016 and there hasn't been any update on it since - is this still a problem?

sechix:
Hello, I have the same problem, how did you fix it?

seand7565:
@sechix Can you take us through the steps to replicate this problem on a clean server? I was unable to replicate this issue following the steps detailed in the initial post. Also: What version of Solidus are you on?

mntmn:
I'm having the same problem with 2.5.0beta2. Also the variant prices don't include the taxes but add them on top, even if specified otherwise.

[http://dump.mntmn.com/solidus-variants-bug.png](http://dump.mntmn.com/solidus-variants-bug.png)

mntmn:
The reason is that `variants_and_option_values_for` creates a query that contains something like this:

`... AND ("spree_prices"."country_iso" = 'DE' OR "spree_prices"."country_iso" IS NULL) `

[...]

PR #2630, opened 2018-03-15
Fix duplicate variants on product page

mamhoff:
This fixes issue #1654. The default pricing options class has a
`search_arguments` which adds `nil` as a matching possibility for the
price's country field (in order to accomodate for the fallback price).

In Spree::Variant.with_prices, Spree::Price.where(search_arguments) is
merged into a Variant query. This results in duplicate variants being
returned.

From the standard product page, `Spree::Variant.with_prices` is called
by `Spree::Product#variants_and_option_values_for(current_pricing_options)`.

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -138,11 +138,23 @@ module Spree
     end
 
     # Returns variants that have a price for the given pricing options
+    # If you have modified the pricing options class, you might want to modify this scope too.
     #
     # @param pricing_options A Pricing Options object as defined on the price selector class
     # @return [ActiveRecord::Relation]
     def self.with_prices(pricing_options = Spree::Config.default_pricing_options)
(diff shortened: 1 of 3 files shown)


jhawthorn:
I made a similar fix to Product#with_master_price by [`using an EXISTS subquery`](https://github.com/solidusio/solidus/blob/d803ee9e9b89d96d985fde4a4d2c897a9f54a21c/core/app/models/spree/product/scopes.rb#L176-L178) (see #2021). That might work better here than using a DISTINCT

mamhoff:
Thank you John, good tip! Weirdly though, in this particular context ActiveRecord seems to have difficulty generating the correct SQL for the pricing options' search arguments, so I had to write some SQL by hand to make it work. This means that if a store overrides their pricing options class they might have to override the `with_prices` scope, too. If anyone can come up with a more generic solution, I'm all for it.

jhawthorn:
I think as long as it works under both MySQL and Postgres (and it seems to), the raw SQL is fine.

PR #2632, opened 2018-03-15
Add yard rake task

jhawthorn:
This adds a `rake yard` task to core, and sets the default options we should be using for documentation


tvdeyen:
We probably want to `.gitignore` 

```
.yardoc
```

as well

PR #2635, opened 2018-03-15
Move EmailValidator under Spree namespace

tvdeyen:
Change all usages of

    validates :email, email: true

to

    validates :email, 'spree/email' => true

### DIFF
--- a/core/lib/spree/core/validators/email.rb
+++ b/core/lib/spree/core/validators/email.rb
@@ -1,9 +1,30 @@
 # frozen_string_literal: true
 
-class EmailValidator < ActiveModel::EachValidator
-  def validate_each(record, attribute, value)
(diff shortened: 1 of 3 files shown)


core/lib/spree/core/validators/email.rb
```diff
@@ -1,9 +1,30 @@
...
+
+    def validate_each(record, attribute, value)
+      unless value =~ EMAIL_REGEXP
+        record.errors.add(attribute, :invalid, { value: value }.merge!(options))
+      end
     end
   end
 end
+
+# @private
```
kennyadsl: why `@private`?

kennyadsl: Got it after reading https://github.com/solidusio/solidus/pull/2629

tvdeyen:
@jhawthorn 😘

PR #2646, opened 2018-03-22
Disable backend footer profile edit link if role cannot edit users

gianlucarizzo:
A user with a role that can access the backend main page but has no permissions to edit users will get a misleading "Authorization Failure" error.
With this fix the user will be cleanly routed to the main page without displaying any errors.

### DIFF
--- a/backend/app/views/spree/admin/shared/_navigation_footer.html.erb
+++ b/backend/app/views/spree/admin/shared/_navigation_footer.html.erb
@@ -6,9 +6,16 @@
 <% if try_spree_current_user %>
   <ul id="login-nav" class="admin-login-nav">
     <li data-hook="user-account-link">
-      <%= link_to spree.polymorphic_path([:edit, :admin, try_spree_current_user]) do %>
-        <i class='fa fa-user'></i>
-        <%= try_spree_current_user.email %>
+      <% if can?(:admin, try_spree_current_user) %>
+        <%= link_to spree.edit_admin_user_path(try_spree_current_user) do %>
+          <i class='fa fa-user'></i>
+          <%= try_spree_current_user.email %>
+        <% end %>
+      <% else %>
+        <a>
+          <i class='fa fa-user'></i>
+          <%= try_spree_current_user.email %>
+        </a>
       <% end %>
     </li>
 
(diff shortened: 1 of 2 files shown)


backend/spec/views/spree/admin/shared/navigation_footer_spec.rb
```diff
@@ -6,16 +6,39 @@
   let(:user) { FactoryBot.build_stubbed(:admin_user) }
   before do
     allow(view).to receive(:try_spree_current_user).and_return(user)
+    @ability = Object.new.extend(CanCan::Ability)
```
kennyadsl: I think this can be a let outside the before block

backend/spec/views/spree/admin/shared/navigation_footer_spec.rb
```diff
@@ -6,16 +6,39 @@
...
+    allow(@controller).to receive(:current_ability).and_return(@ability)
```
kennyadsl: I think `@controller` [can be called as `controller` without the `@` here](https://relishapp.com/rspec/rspec-rails/v/3-7/docs/view-specs/view-spec).

backend/spec/views/spree/admin/shared/navigation_footer_spec.rb
```diff
@@ -6,16 +6,39 @@
...
     render
     expect(rendered).to have_selector("#login-nav")
   end
 
-  it "has a user-account-link" do
-    render
-    expect(rendered).to have_link(user.email, href: Spree::Core::Engine.routes.url_helpers.edit_admin_user_path(user))
+  context "authorized user" do
+    it "has a user-account-link that links to edit_admin_user_path" do
+      @ability.can :admin, user
```
kennyadsl: you can move this ability setting in a before that lives into the context `authorized user` since it will be applied to all its scenarios.

backend/app/views/spree/admin/shared/_navigation_footer.html.erb
```diff
@@ -6,9 +6,16 @@
...
-      <%= link_to spree.polymorphic_path([:edit, :admin, try_spree_current_user]) do %>
-        <i class='fa fa-user'></i>
-        <%= try_spree_current_user.email %>
+      <% if can?(:admin, try_spree_current_user) %>
+        <%= link_to spree.edit_admin_user_path(try_spree_current_user) do %>
+          <i class='fa fa-user'></i>
+          <%= try_spree_current_user.email %>
+        <% end %>
+      <% else %>
+        <%= link_to spree.admin_path do %>
```
jhawthorn: I'm not sure it makes sense to link to the root. Maybe we should drop the link entirely (just plain text) if we don't have permissions?

gianlucarizzo: fixed!

gianlucarizzo:
I think it's all okay now

PR #2648, opened 2018-03-26
taxon.rb icon attachment size

nxtbz:
# icon attachment

Updated the has_attachemend_file icon sizes for Taxonomies.

```
     has_attached_file :icon,
- styles: { mini: '32x32>', normal: '128x128>' },
+ styles: { mini: '32x32>', normal: '128x128>', medium: '400x400>', large: '600x600>' },

```

### DIFF
--- a/core/app/models/spree/taxon.rb
+++ b/core/app/models/spree/taxon.rb
@@ -26,7 +26,7 @@ module Spree
     after_touch :touch_ancestors_and_taxonomy
 
     has_attached_file :icon,
-      styles: { mini: '32x32>', normal: '128x128>' },
+      styles: { mini: '32x32>', normal: '128x128>', medium: '400x400>', large: '600x600>' },
       default_style: :mini,
       url: '/spree/taxons/:id/:style/:basename.:extension',
       path: ':rails_root/public/spree/taxons/:id/:style/:basename.:extension',


nxtbz:
Mysql failure?

kennyadsl:
If you rebase against master now the Mysql specs failure should go away. It's not clear why it's needed to update those taxons icon sizes honestly. Can you please explain better?

core/app/models/spree/taxon.rb
```diff
@@ -26,7 +26,7 @@ class Taxon < Spree::Base
     after_touch :touch_ancestors_and_taxonomy
 
     has_attached_file :icon,
-      styles: { mini: '32x32>', normal: '128x128>' },
+      styles: { mini: '32x32>', normal: '128x128>', medium: '400x400>', large: '600x600>' },
```
tvdeyen: Not all stores need all these sizes. This is a rarely used feature, but if a store is using this we force them to pre-process four sizes now, what is not what we want to force stores into. 

Could you make this change in your particular store instead? Here is an article on how to achieve this https://github.com/solidusio/solidus/blob/master/guides/products-and-variants/product-images.md#paperclip-settings (Just replace `Spree::Image` with `Spree::Taxon`)

PR #2649, opened 2018-03-26
Clean CheckoutsController routes

gevann:
The Api::Checkouts::Controller does not have any actions to match the
order_concern routes (empty, cancel, apply_coupon_code).


jhawthorn:
Maybe those missing routes could be moved out of order_concerns and onto `resources :orders`

gevann:
Requested changes have been made, thanks!

PR #2656, opened 2018-03-27
Remove state machine gem from Spree::Shipment

jgayfer:
The [state_machines gem](https://github.com/state-machines/state_machines) has been used throughout Solidus since the earliest days of Spree. We use it to track the "state" our shipments/payments/orders are in, as well as plug custom functionality into the transitions between those states (`before_transition`/`after_transition`). 

However, our use of intermingled state machines between Payments, Shipments and Orders causes problems for both new and established developers. The state machine transitions are difficult to understand and debug, the order the transitions are executed in can be tough to control, and properly controlling database transactions during a transition is very poorly understood. 

We believe the downsides of implementing our state machines using the state_machines gem to be a net negative for the project.

This PR changes the `Spree::Shipment` class, explicitly defining the important methods that would have been generated from the state_machines metaprogramming. The advantages of doing this are improved code clarity, debug-ability, as well as making it more easy to completely override methods regarding state.

The first two commits of this pull request remove the `state_machine` from [`Spree::Shipment`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/shipment.rb) without changing the external API of the model. The last two commits go a bit further and deprecate the [transition methods](https://github.com/solidusio/solidus/blob/fd474db749d4386f52a23efbb59d30336351028d/core/app/models/spree/shipment.rb#L79-L91) that were used by the state machine in an attempt to improve code readability. If we don't want to go ahead and deprecate the transition methods, the first two commits can just be taken on their own, keeping the external API of [`Spree::Shipment`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/shipment.rb) intact.

While we could re-implement the `before_transition`/`after_transition` calls on the `Spree::Shipment` model, we believe the following examples would be a better way to support customising functionality before or after a transition (with `ship` used as an example).

```ruby
module AfterShipAction
  # Note that returning true/false is done to preserve the original return values of ship
  # This pattern works for all of ship, resume, pend, ready, and cancel
  def ship
    if super
      additional_after_ship_procedure
    end
  end

  def additional_after_ship_procedure
    # Do something interesting!
    true
  end
end

module BeforeShipCondition
  def ship
    if pre_ship_condition
      super
    end
  end

  def pre_ship_condition
    # Some interesting condition as to when we can ship
    true
  end
end

Spree::Shipment.prepend AfterShipAction
Spree::Shipment.prepend BeforeShipCondition
```

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -4,6 +4,14 @@ module Spree
   # An order's planned shipments including tracking and cost.
   #
   class Shipment < Spree::Base
+    class InvalidStateChange < StandardError; end
+
+    READY = 'ready'
+    PENDING = 'pending'
+    SHIPPED = 'shipped'
+    CANCELED = 'canceled'
+    DEFAULT_STATES = [READY, PENDING, SHIPPED, CANCELED]
+
     belongs_to :order, class_name: 'Spree::Order', touch: true, inverse_of: :shipments
     belongs_to :stock_location, class_name: 'Spree::StockLocation'
 
@@ -13,6 +21,8 @@ module Spree
     has_many :shipping_methods, through: :shipping_rates
     has_many :state_changes, as: :stateful
     has_many :cartons, -> { distinct }, through: :inventory_units
+
+    validate :is_valid_state?
 
     before_validation :set_cost_zero_when_nil
 
@@ -26,49 +36,111 @@ module Spree
 
     make_permalink field: :number, length: 11, prefix: 'H'
 
-    scope :pending, -> { with_state('pending') }
-    scope :ready,   -> { with_state('ready') }
-    scope :shipped, -> { with_state('shipped') }
+    scope :pending, -> { with_state(PENDING) }
+    scope :ready,   -> { with_state(READY) }
+    scope :shipped, -> { with_state(SHIPPED) }
     scope :trackable, -> { where("tracking IS NOT NULL AND tracking != ''") }
     scope :with_state, ->(*s) { where(state: s) }
     # sort by most recent shipped_at, falling back to created_at. add "id desc" to make specs that involve this scope more deterministic.
     scope :reverse_chronological, -> { order('coalesce(spree_shipments.shipped_at, spree_shipments.created_at) desc', id: :desc) }
     scope :by_store, ->(store) { joins(:order).merge(Spree::Order.by_store(store)) }
 
-    # shipment state machine (see http://github.com/pluginaweek/state_machine/tree/master for details)
-    state_machine initial: :pending, use_transactions: false do
-      event :ready do
-        transition from: :pending, to: :shipped, if: :can_transition_from_pending_to_shipped?
-        transition from: :pending, to: :ready, if: :can_transition_from_pending_to_ready?
-      end
+    def cancel!
+      cancel || raise(InvalidStateChange)
+    end
 
-      event :pend do
-        transition from: :ready, to: :pending
-      end
+    def cancel
+      return false unless can_cancel?
+      change_state!(CANCELED)
+      after_cancel
+      true
+    end
 
-      event :ship do
-        transition from: [:ready, :canceled], to: :shipped
-      end
-      after_transition to: :shipped, do: :after_ship
+    def can_cancel?
+      ready_or_pending?
+    end
 
-      event :cancel do
-        transition to: :canceled, from: [:pending, :ready]
-      end
-      after_transition to: :canceled, do: :after_cancel
+    def canceled?
+      state == CANCELED
+    end
 
-      event :resume do
-        transition from: :canceled, to: :ready, if: :can_transition_from_canceled_to_ready?
-        transition from: :canceled, to: :pending
-      end
-      after_transition from: :canceled, to: [:pending, :ready, :shipped], do: :after_resume
+    def resume!
+      resume || raise(InvalidStateChange)
+    end
 
-      after_transition do |shipment, transition|
-        shipment.state_changes.create!(
-          previous_state: transition.from,
-          next_state:     transition.to,
-          name:           'shipment'
-        )
+    def resume
+      return false unless can_resume?
+      can_transition_from_canceled_to_ready? ? change_state!(READY) : change_state!(PENDING)
+      after_resume
+      true
+    end
+
+    def can_resume?
+      canceled?
+    end
+
+    def ship!
+      ship || raise(InvalidStateChange)
+    end
+
+    def ship
+      return false unless can_ship?
+      previous_state = state
+      change_state!(SHIPPED)
+      after_ship
+      after_resume if previous_state == CANCELED
+      true
+    end
+
+    def can_ship?
+      ready? || canceled?
+    end
+
+    def shipped?
+      state == SHIPPED
+    end
+
+    def ready!
+      ready || raise(InvalidStateChange)
+    end
+
+    def ready
+      return false unless state == PENDING
+      if can_transition_from_pending_to_shipped?
+        change_state!(SHIPPED)
+        after_ship
+      elsif can_transition_from_pending_to_ready?
+        change_state!(READY)
+      else
+        return false
       end
+      true
+    end
+
+    def ready?
+      state == READY
+    end
+
+    def can_ready?
+      pending? && (can_transition_from_pending_to_shipped? || can_transition_from_pending_to_ready?)
+    end
+
+    def pend!
+      pend || raise(InvalidStateChange)
+    end
+
+    def pend
+      return false unless can_pend?
+      change_state!(PENDING)
+      true
+    end
+
+    def pending?
+      state == PENDING
+    end
+
+    def can_pend?
+      ready?
     end
 
     self.whitelisted_ransackable_associations = ['order']
@@ -264,13 +336,13 @@ module Spree
     # shipped    if already shipped (ie. does not change the state)
     # ready      all other cases
     def determine_state(order)
-      return 'canceled' if order.canceled?
-      return 'shipped' if shipped?
-      return 'pending' unless order.can_ship?
+      return CANCELED if order.canceled?
+      return SHIPPED if shipped?
+      return PENDING unless order.can_ship?
       if can_transition_from_pending_to_ready?
-        'ready'
+        READY
       else
-        'pending'
+        PENDING
       end
     end
 
@@ -355,7 +427,7 @@ module Spree
           state: new_state,
           updated_at: Time.current
         )
-        after_ship if new_state == 'shipped'
+        after_ship if new_state == SHIPPED
       end
     end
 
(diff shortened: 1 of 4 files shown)


core/app/models/spree/shipment.rb
```diff
@@ -156,16 +156,25 @@ def inventory_can_ship?
...
       !requires_shipment?
     end
 
     def can_transition_from_pending_to_ready?
+      Spree::Deprecation.warn \
+        'Spree::Shipment#can_transition_from_pending_to_ready? is deprecated. ' \
+        'Use Spree::Shipment#inventory_can_ship? instead.'
       order.can_ship? &&
         inventory_units.all? { |iu| iu.shipped? || iu.allow_ship? || iu.canceled? } &&
         (order.paid? || !Spree::Config[:require_payment_to_ship])
```
tvdeyen: could we call `inventory_can_ship?` in here?

core/app/models/spree/shipment.rb
```diff
@@ -14,6 +21,10 @@ class Shipment < Spree::Base
     has_many :state_changes, as: :stateful
     has_many :cartons, -> { uniq }, through: :inventory_units
 
+    validates :state, presence: true, inclusion: { in: [READY, PENDING, SHIPPED, CANCELED] }
```
tvdeyen: `presence: true` is superfluous if we also check the inclusion in something that's not having `nil`.

Maybe we could also add default value for the `state` column?

core/app/models/spree/shipment.rb
```diff
@@ -14,6 +21,10 @@ class Shipment < Spree::Base
...
+
+    after_initialize :set_default_state
```
tvdeyen: We could have a default value for the `state` column instead.

core/app/models/spree/shipment.rb
```diff
@@ -430,5 +518,28 @@ def ensure_can_destroy
...
+      )
+    end
+
+    def change_state!(new_state)
+      previous_state = state
+      if new_state != previous_state
+        update_attributes(
+          state: new_state,
+          updated_at: Time.current
+        )
```
tvdeyen: ActiveRecord already sets `updated_at` for us.

Also, the bang at the end of the method name indicates that this method raises if the record couldn't be saved. We should use `update!` here.

```rb
update!(state: new_state)
```

tvdeyen: We should consider to move the methods `change_state!` and `store_state_change` into a module we share with all models that have `state` changes.

core/app/models/spree/shipment.rb
```diff
@@ -26,67 +37,144 @@ class Shipment < Spree::Base
...
+        'Use Spree::Shipment#inventory_can_ship? instead.'
       order.can_ship? &&
         inventory_units.all? { |iu| iu.shipped? || iu.allow_ship? || iu.canceled? } &&
         (order.paid? || !Spree::Config[:require_payment_to_ship])
     end
 
     def can_transition_from_canceled_to_ready?
+      Spree::Deprecation.warn \
+        'Spree::Shipment#can_transition_from_pending_to_ready? is deprecated. ' \
+        'Use Spree::Shipment#inventory_can_ship? instead.'
```
tvdeyen: dito

core/app/models/spree/shipment.rb
```diff
@@ -26,67 +37,144 @@ class Shipment < Spree::Base
...
+    def inventory_can_ship?
+      order.can_ship? &&
+        inventory_units.all? { |iu| iu.shipped? || iu.allow_ship? || iu.canceled? } &&
+        (order.paid? || !Spree::Config[:require_payment_to_ship])
+    end
+
     def can_transition_from_pending_to_shipped?
+      Spree::Deprecation.warn \
+        'Spree::Shipment#can_transition_from_pending_to_shipped is deprecated. ' \
+        'It is equivalent to the negation of Spree::Shipment#requires_shipment?'
```
tvdeyen: dito

core/app/models/spree/shipment.rb
```diff
@@ -26,67 +37,144 @@ class Shipment < Spree::Base
...
+    def ready?
+      state == READY
+    end
+
+    def can_ready?
+      pending? && (!requires_shipment? || inventory_can_ship?)
+    end
+
+    def pend!
+      pend || (raise InvalidStateChange)
```
tvdeyen: weird usage of parenthesis here.

Either

```rb
raise InvalidStateChange
```

or

```rb
raise(InvalidStateChange)
```

core/app/models/spree/shipment.rb
```diff
@@ -26,67 +37,144 @@ class Shipment < Spree::Base
...
+    def can_ship?
+      ready? || canceled?
+    end
+
+    def shipped?
+      state == SHIPPED
+    end
+
+    def ready!
+      ready || (raise InvalidStateChange)
```
tvdeyen: Ditto, weird usage of parenthesis

core/app/models/spree/shipment.rb
```diff
@@ -26,67 +37,144 @@ class Shipment < Spree::Base
...
+      after_resume
+      true
+    end
+
+    def can_resume?
+      canceled?
+    end
+
+    def ship!
+      ship || (raise InvalidStateChange)
```
tvdeyen: Ditto, parenthesis

core/app/models/spree/shipment.rb
```diff
@@ -430,5 +518,28 @@ def ensure_can_destroy
...
+      if new_state != previous_state
+        update_attributes(
+          state: new_state,
+          updated_at: Time.current
+        )
+        store_state_change(previous_state, new_state)
+      end
+    end
+
+    def set_default_state
```
tvdeyen: A database default would be preferable

jgayfer:
@tvdeyen Thanks for all the great feedback! I've gone ahead and made _most_ of the requested changes. 

I haven't pulled out the `change_state!` method into a module yet, as I wasn't able to decide on the best way to do so. While there is shared functionality in the `change_state!` method, the four models in question use it a bit differently. For example, not all of the models need to create a `StateChange`, creating a `StateChange` requires a value for it's `name`, and `Reimbursement` uses the `reimbursement_status` attribute instead of `state`. It's entirely possible to pull it out into a module, but because of the reasons just mentioned, my initial sketch had more parameters than I would have liked.

With all that being said, I'd love to hear any insight you may have on a better way to pull out the shared functionality!

Lastly, in regards to things like `CANCELABLE_STATES` on inventory units (and other constants) and potentially turning them into methods (or adding them to the config), I haven't made any changes yet. I wasn't sure if this was something we were wanting to look into now, or if it should be a future change instead.

Thanks again for taking time to review my work

core/db/migrate/20180404185904_add_default_state_to_shipment.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDefaultStateToShipment < ActiveRecord::Migration[5.1]
+  def change
+    change_column_default(:spree_shipments, :state, Spree::Shipment::PENDING)
```
tvdeyen: We usually don't want to rely on code inside the migrations. Although very unlikely that the value for `Spree::Shipment::PENDING` will change ever, it still is more reliable to use the value here instead.

core/app/models/spree/shipment.rb
```diff
@@ -430,5 +508,20 @@ def ensure_can_destroy
...
+      state_changes.create!(
+        previous_state: previous_state,
+        next_state:     new_state,
+        name:           'shipment'
+      )
+    end
+
+    def change_state!(new_state)
+      previous_state = state
+      return if new_state == previous_state
```
tvdeyen: We may can make use of `ActiveModel::Dirty` here?

core/app/models/spree/shipment.rb
```diff
@@ -26,69 +35,138 @@ class Shipment < Spree::Base
...
+    def ship!
+      ship || raise(InvalidStateChange)
+    end
+
+    def ship
+      return false unless can_ship?
+      previous_state = state
+      change_state!(SHIPPED)
+      after_ship
+      after_resume if previous_state == CANCELED
```
tvdeyen: Can we try to make use of `ActiveModel::Dirty` here?

ccarruitero:
I'm not agree with remove the `state_machines` gem.

I feel that reimplementing `state_machines` functionality on each model is something repetitive.
Aside, we are assuming that everyone want to use the default states, I think if someone require states completely different for their use case, would be necessary more work to do than is currently.

Like you mention maybe `state_machines` is difficult to understand, but instead remove I think we should improve documentation in that field.

Finally if you want to remove `state_machines`, I think a better way could be find and alternative or do a module replacement for that, and allow configure in some way what to use for the states.

jgayfer:
@tvdeyen I made the change regarding the database migration (which definitely makes sense). I will make this change to my other PRs early next week.

Regarding `ActiveModel::Dirty`, I just want to confirm that the goal is to use it to check what the previous state was without having to assign an extra variable? (For example, when I use `previous_state` to make note of what the original state was).

tvdeyen:
> Regarding ActiveModel::Dirty, I just want to confirm that the goal is to use it to check what the previous state was without having to assign an extra variable? (For example, when I use previous_state to make note of what the original state was).

@jgayfer yes. Like I said not mandatory, but still worth trying, as we are duplicating AR behavior here.

core/app/models/spree/shipment.rb
```diff
@@ -14,6 +21,8 @@ class Shipment < Spree::Base
     has_many :state_changes, as: :stateful
     has_many :cartons, -> { distinct }, through: :inventory_units
 
+    validates :state, inclusion: { in: [READY, PENDING, SHIPPED, CANCELED] }
```
tvdeyen: We should extract this into a method, as stores might want to extend/change states.

jgayfer:
@tvdeyen Changes have been made.

I wasn't entirely sure exactly how someone might want to go about changing what states are valid, so I tried to keep the options open.

There is a new constant `VALID_STATES` which can be overridden should someone desire to change what states are allowed. I also added a validation function `validate_state` which checks to see if the current state is included in `VALID_STATES`. The inclusion of this function allows for a potentially more detailed custom state validation process as changing the `VALID_STATES` constant may not cover every use case.

core/app/models/spree/shipment.rb
```diff
@@ -14,6 +22,8 @@ class Shipment < Spree::Base
     has_many :state_changes, as: :stateful
     has_many :cartons, -> { distinct }, through: :inventory_units
 
+    validate :validate_state
```
tvdeyen: Can we call this method `valid_state` or `is_valid_state`? Reads better.

core/app/models/spree/shipment.rb
```diff
@@ -4,6 +4,14 @@ module Spree
   # An order's planned shipments including tracking and cost.
   #
   class Shipment < Spree::Base
+    class InvalidStateChange < StandardError; end
+
+    READY = 'ready'
+    PENDING = 'pending'
+    SHIPPED = 'shipped'
+    CANCELED = 'canceled'
+    VALID_STATES = [READY, PENDING, SHIPPED, CANCELED]
```
tvdeyen: Can we rename this constant to `DEFAULT_STATES`? We don't want to give the impression that overwriting a constant is a good extension point (because it's not)

jgayfer:

kennyadsl:
reopened in #3038

PR #2658, opened 2018-03-28
Update Orders state machine doc removing confirmation_required text

kennyadsl:
This feature has been removed here:

https://github.com/solidusio/solidus/commit/8c8c84c975636d72e77501f244e69630c58e3e4c

cc @benjaminwil

### DIFF
--- a/guides/orders/order-state-machine.md
+++ b/guides/orders/order-state-machine.md
@@ -16,11 +16,8 @@ There are six order states by default:
 If you go through the checkout provided in the `solidus_frontend` gem, you can
 see that there is a clearly defined step for each of these states during checkout.
 
-In some circumstances, the `payment` and `confirm` states are optional:
-
-- The `payment` state is only triggered if `payment_required?` returns `true`.
-- The `confirm` state is only triggered if `confirmation_required?` returns
-  `true`.
+The `payment` state is optional and it's only triggered if `payment_required?` 
+returns `true`.
 
 The `complete` state is triggered in one of two ways:
 


guides/orders/order-state-machine.md
```diff
@@ -16,12 +16,8 @@ There are six order states by default:
...
 see that there is a clearly defined step for each of these states during checkout.
 
-In some circumstances, the `payment` and `confirm` states are optional:
-
-- The `payment` state is only triggered if `payment_required?` returns `true`.
-- The `confirm` state is only triggered if `confirmation_required?` returns
-  `true`.
-
+The `payment` state is optional and it's only triggered if `payment_required?` 
+returns `true`.
```
benjaminwil: We should add a newline after this paragraph to separate it from the paragraph below.

benjaminwil:
@kennyadsl Thanks for the mention :+1:

guides/orders/order-state-machine.md
kennyadsl: Ok! I thought that since it's a single line it would be better to compact those paragraphs but here is your land, I will update shortly! 😄

kennyadsl: done!

kennyadsl:
@benjaminwil do you think it's ok now?

benjaminwil:
@kennyadsl Yes, thanks. 😊

guides/orders/order-state-machine.md
```diff
@@ -16,11 +16,8 @@ There are six order states by default:
 If you go through the checkout provided in the `solidus_frontend` gem, you can
 see that there is a clearly defined step for each of these states during checkout.
 
-In some circumstances, the `payment` and `confirm` states are optional:
-
-- The `payment` state is only triggered if `payment_required?` returns `true`.
-- The `confirm` state is only triggered if `confirmation_required?` returns
-  `true`.
+The `payment` state is optional and it's only triggered if `payment_required?` 
+returns `true`.
```
benjaminwil: I agree, we could have alternatively merged the paragraphs. But they are about different states.

PR #2659, opened 2018-03-28
Add a shipment_id index to shipping_rates table

tvdeyen:
This missing index speeds up the `Spree::Shipment#selected_shipping_rate` query
for large applications from 4 seconds to 4 milliseconds. This is for a table with
1,297,260 shipping rates.

Beware that this migration could take a while (~2 seconds for above mentioned table).


jhawthorn:
I thought that the [index on `["shipment_id", "shipping_method_id"]`](https://github.com/solidusio/solidus/blob/master/core/db/migrate/20160101010000_solidus_one_four.rb#L786) would cover this (but I could be wrong).

tvdeyen:
@jhawthorn somehow 🤔 I am missing that index on my database. Git log showed me that someone actually removed that index because of problems with the uniqueness constraints. Can confirm that the same index even w/o the uniqueness constraint is mitigating the performance problems in the dame way the single column index does.

PR #2663, opened 2018-03-29
Remove promotion_code from adjustment_attributes

gevann:
Adjustments do not have an attribute called 'promotion_code'.

### DIFF
--- a/api/app/helpers/spree/api/api_helpers.rb
+++ b/api/app/helpers/spree/api/api_helpers.rb
@@ -133,7 +133,7 @@ module Spree
 
       @@adjustment_attributes = [
         :id, :source_type, :source_id, :adjustable_type, :adjustable_id,
-        :amount, :label, :promotion_code,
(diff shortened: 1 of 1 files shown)


api/app/helpers/spree/api/api_helpers.rb
```diff
@@ -133,7 +133,7 @@ def required_fields_for(model)
 
       @@adjustment_attributes = [
         :id, :source_type, :source_id, :adjustable_type, :adjustable_id,
-        :amount, :label, :promotion_code,
```
kennyadsl: There's a [promotion_code_id on `Spree::Adjustment`](https://github.com/solidusio/solidus/blob/a779d170d3c4224fb6aa3f6551bf061374d9a6f0/core/app/models/spree/adjustment.rb#L22)s. Maybe it's better to update that attribute instead of removing it?

gevann:
Requested changes have been made, thanks :smile:

PR #2664, opened 2018-03-29
Remove state machine gem from Spree::Payment

jgayfer:
This PR removes the state machine gem from `Spree::Payment` while keeping the external API intact. The logic that was previously hidden within the state machine is now contained within the `Spree::Payment` model.

### DIFF
--- a/core/app/models/spree/payment.rb
+++ b/core/app/models/spree/payment.rb
@@ -8,8 +8,19 @@ module Spree
   class Payment < Spree::Base
     include Spree::Payment::Processing
 
+    class InvalidStateChange < StandardError; end
+
     alias_attribute :identifier, :number
     deprecate :identifier, :identifier=, deprecator: Spree::Deprecation
+
+    PENDING = 'pending'
+    PROCESSING = 'processing'
+    CHECKOUT = 'checkout'
+    COMPLETED = 'completed'
+    INVALID = 'invalid'
+    VOID = 'void'
+    FAILED = 'failed'
+    DEFAULT_STATES = [PENDING, PROCESSING, CHECKOUT, COMPLETED, INVALID, VOID, FAILED]
 
     IDENTIFIER_CHARS    = (('A'..'Z').to_a + ('0'..'9').to_a - %w(0 1 I O)).freeze
     NON_RISKY_AVS_CODES = ['B', 'D', 'H', 'J', 'M', 'Q', 'T', 'V', 'X', 'Y'].freeze
@@ -43,6 +54,7 @@ module Spree
     validates :amount, numericality: true
     validates :source, presence: true, if: :source_required?
     validates :payment_method, presence: true
+    validate :is_valid_state?
 
     default_scope -> { order(:created_at) }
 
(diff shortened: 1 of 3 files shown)


core/app/models/spree/payment.rb
```diff
@@ -8,9 +8,19 @@ module Spree
...
     alias_attribute :identifier, :number
     deprecate :identifier, :identifier=, deprecator: Spree::Deprecation
 
+    PENDING = 'pending'
+    PROCESSING = 'processing'
+    CHECKOUT = 'checkout'
+    COMPLETED = 'completed'
+    INVALID = 'invalid'
+    VOID = 'void'
+    FAILED = 'failed'
```
BenMorganIO: We should freeze these strings since they shouldn't be mutatable.

jgayfer: Will do!

core/app/models/spree/payment.rb
```diff
@@ -43,6 +53,7 @@ class Payment < Spree::Base
     validates :amount, numericality: true
     validates :source, presence: true, if: :source_required?
     validates :payment_method, presence: true
+    validates :state, inclusion: { in: [PENDING, PROCESSING, CHECKOUT, COMPLETED, INVALID, VOID, FAILED] }
```
BenMorganIO: I feel like a method should be used here. So that way people can easily extend payment states if they need to.

core/app/models/spree/payment.rb
jgayfer: Oh, just realised the `frozen_string_literal` at the top of the file takes care of this already.

BenMorganIO: Ah, right. We should delete all of the other freeze calls then :p

jgayfer:
@tvdeyen Requested changes have been made. See my comments on #2656

core/app/models/spree/payment.rb
```diff
@@ -280,5 +364,20 @@ def create_eligible_credit_event
...
+      state_changes.create!(
+        previous_state: previous_state,
+        next_state:     new_state,
+        name:           'payment'
+      )
+    end
+
+    def change_state!(new_state)
+      previous_state = state
+      return if new_state == previous_state
```
tvdeyen: Can we try to make use of `ActiveModel::Dirty` here?

core/db/migrate/20180404215857_add_default_state_to_payment.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDefaultStateToPayment < ActiveRecord::Migration[5.1]
+  def change
+    change_column_default(:spree_payments, :state, Spree::Payment::CHECKOUT)
```
tvdeyen: We should use the value of the constant here so we do not rely on the code may change in the future.

jgayfer:
@tvdeyen I've changed the database migration to use the actual value. Same question applies as in #2656  about `ActiveModel::Dirty` though.

core/app/models/spree/payment.rb
tvdeyen: Very good point Ben. @jgayfer would you mind to address this?

jgayfer:
@tvdeyen Made some changes. See my comments on #2656.

jgayfer:

BenMorganIO:
Are we able to place state machine specific methods in their own class? I feel like that would make managing the model so much easier.

tvdeyen:
> Are we able to place state machine specific methods in their own class? I feel like that would make managing the model so much easier.

@BenMorganIO a `module` would be fine 👍 But we agreed to add this after we merge these PRs

kennyadsl:
reopened in #3039

PR #2666, opened 2018-03-29
Remove state machine gem from Spree::Reimbursement

jgayfer:
This PR removes the state machine gem from `Spree::Reimbursement` while keeping the external API intact. The logic that was previously hidden within the state machine is now contained within the `Spree::Reimbursement` model.

### DIFF
--- a/core/app/models/spree/reimbursement.rb
+++ b/core/app/models/spree/reimbursement.rb
@@ -3,6 +3,12 @@
 module Spree
   class Reimbursement < Spree::Base
     class IncompleteReimbursementError < StandardError; end
+    class InvalidStateChange < StandardError; end
+
+    PENDING = 'pending'
+    ERRORED = 'errored'
+    REIMBURSED = 'reimbursed'
+    DEFAULT_REIMBURSEMENT_STATUSES = [PENDING, ERRORED, REIMBURSED]
 
     belongs_to :order, inverse_of: :reimbursements
     belongs_to :customer_return, inverse_of: :reimbursements, touch: true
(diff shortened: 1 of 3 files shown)


jgayfer:
@tvdeyen Changes have been made. See my comments on #2656. Thanks for the input!

core/db/migrate/20180404235352_add_default_status_to_reimbursements.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDefaultStatusToReimbursements < ActiveRecord::Migration[5.1]
+  def change
+    change_column_default(:spree_reimbursements, :reimbursement_status, Spree::Reimbursement::PENDING)
```
tvdeyen: We should use the value of the constant here so we do not rely on the code may change in the future.

jgayfer:
@tvdeyen I've changed the database migration to use the actual value. Thanks for reviewing my work!

tvdeyen:
@jgayfer could you please revisit the failing specs?

jgayfer:
@tvdeyen Specs are green now :heavy_check_mark:

core/app/models/spree/reimbursement.rb
```diff
@@ -14,12 +19,13 @@ class IncompleteReimbursementError < StandardError; end
 
     validates :order, presence: true
     validate :validate_return_items_belong_to_same_order
+    validates :reimbursement_status, inclusion: { in: [PENDING, ERRORED, REIMBURSED] }
```
tvdeyen: Stores might want to extend/change the states. This should be a method instead.

jgayfer:
@tvdeyen Made some changes. Please see my comment on #2656.

jgayfer:

kennyadsl:
reopened in #3040

PR #2668, opened 2018-04-03
Remove state machine gem from Spree::InventoryUnit

jgayfer:
This PR removes the state machine gem from `Spree::InventoryUnit` while keeping the external API intact. The logic that was previously hidden within the state machine is now contained within the `Spree::InventoryUnit` model.

### DIFF
--- a/core/app/models/spree/inventory_unit.rb
+++ b/core/app/models/spree/inventory_unit.rb
@@ -4,9 +4,18 @@ module Spree
   # Tracks the state of line items' fulfillment.
   #
   class InventoryUnit < Spree::Base
-    PRE_SHIPMENT_STATES = %w(backordered on_hand)
-    POST_SHIPMENT_STATES = %w(returned)
-    CANCELABLE_STATES = ['on_hand', 'backordered', 'shipped']
+    class InvalidStateChange < StandardError; end
+
+    ON_HAND = 'on_hand'
+    BACKORDERED = 'backordered'
+    RETURNED = 'returned'
+    SHIPPED = 'shipped'
+    CANCELED = 'canceled'
+    DEFAULT_STATES = [ON_HAND, BACKORDERED, RETURNED, SHIPPED, CANCELED]
+
+    PRE_SHIPMENT_STATES = [BACKORDERED, ON_HAND]
+    POST_SHIPMENT_STATES = [RETURNED]
+    CANCELABLE_STATES = [ON_HAND, BACKORDERED, SHIPPED]
 
     belongs_to :variant, -> { with_deleted }, class_name: "Spree::Variant", inverse_of: :inventory_units
     belongs_to :shipment, class_name: "Spree::Shipment", touch: true, inverse_of: :inventory_units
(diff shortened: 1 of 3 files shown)


core/app/models/spree/inventory_unit.rb
```diff
@@ -59,24 +69,81 @@ def order=(_)
...
+    end
+
+    def cancel
+      return false unless can_cancel?
+      change_state!(CANCELED)
+      true
+    end
+
+    def can_cancel?
+      CANCELABLE_STATES.include?(state)
```
tvdeyen: Later we should change `CANCELABLE_STATES` to be a method as stores might want to change that.

Ideally this would even be configurable, so stores won't have to `class_eval` or `Module#prepend` this class

core/app/models/spree/inventory_unit.rb
```diff
@@ -4,9 +4,17 @@ module Spree
...
+
+    ON_HAND = 'on_hand'
+    BACKORDERED = 'backordered'
+    RETURNED = 'returned'
+    SHIPPED = 'shipped'
+    CANCELED = 'canceled'
+
+    PRE_SHIPMENT_STATES = [BACKORDERED, ON_HAND]
+    POST_SHIPMENT_STATES = [RETURNED]
+    CANCELABLE_STATES = [ON_HAND, BACKORDERED, SHIPPED]
```
tvdeyen: I know this is just keeping things like before, but we should consider to make this configurable in the future. Either via a method or configuration.

jgayfer:
@tvdeyen Made some of the requested changes! See my comments on #2656.

core/app/models/spree/inventory_unit.rb
```diff
@@ -155,5 +221,11 @@ def ensure_can_destroy
         throw :abort
       end
     end
+
+    def change_state!(new_state)
+      previous_state = state
```
tvdeyen: Can we try to make use of `ActiveModel::Dirty` here?

core/db/migrate/20180405181459_add_default_state_to_inventory_unit.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDefaultStateToInventoryUnit < ActiveRecord::Migration[5.1]
+  def change
+    change_column_default(:spree_inventory_units, :state, Spree::InventoryUnit::ON_HAND)
```
tvdeyen: We should use the value of the constant here so we do not rely on the code may change in the future.

jgayfer:
@tvdeyen I've changed the database migration to use the actual value. Same question applies as in #2656  about `ActiveModel::Dirty` though.

core/app/models/spree/inventory_unit.rb
```diff
@@ -26,18 +34,19 @@ def order=(_)
     end
 
     validates_presence_of :shipment, :line_item, :variant
+    validates :state, inclusion: { in: [ON_HAND, BACKORDERED, RETURNED, SHIPPED, CANCELED] }
```
tvdeyen: The states may change for stores. We should allow them to overwrite a method instead.

jgayfer:
@tvdeyen Made the requested changes. Check out my comment on #2656.

jgayfer:

kennyadsl:
reopened in #3041

PR #2669, opened 2018-04-03
Load email validator from core instead of order

jhawthorn:
It may be desirable to use the email validator from elsewhere before order.rb has been loaded.

Additionally, this fixes an issue where EmailValidator would be unloaded in development mode. Spree::Order is autoloaded, and because Spree::EmailValidator is inside the same namespace, this confuses rails into thinking Spree::EmailValidator was defined in this file and can be reloaded. I believe this to be a bug in rails.

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'spree/core/validators/email'
 require 'spree/order/checkout'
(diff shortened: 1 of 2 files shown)


jhawthorn:
For anyone curious I do think this is a rails bug and spent some time yesterday evening making a test case for it https://github.com/jhawthorn/rails/commit/cf9c10e1b3e78c188677080fcf684f52ded55206

PR #2670, opened 2018-04-03
Move Spree::Taxon#applicable_filters (rebase)

jhawthorn:
Rebased since it conflicted after #2408 was merged

Fixes #2409


jhawthorn:
> Thanks. Could we satisfy the Hound, though?

done

PR #2673, opened 2018-04-04
Remove has checkout step

### LINKED ISSUE #2502: Behaviour should not depend on has_checkout_step?
We have a few places where the names of checkout steps influence behaviour. We take things like `has_checkout_step?("delivery")` to mean "this order doesn't need shipments". This is a bad way to do this. A store might want to create shipments, but not have a delivery step.

However this behaviour has been allowed for a long time and is being depended upon. See #626 and others.

```
$ ag 'has_checkout_step\?\'
frontend/app/views/spree/shared/_order_details.html.erb
3:  <% if order.has_checkout_step?("address") %>
10:    <% if order.has_checkout_step?("delivery") %>
31:  <% if order.has_checkout_step?("payment") %>

core/app/models/spree/line_item.rb
195:      if (saved_changes? || target_shipment.present?) && order.has_checkout_step?("delivery")

core/app/models/spree/order.rb
845:      if has_checkout_step?("delivery")
```

gevann:
This PR aims to resolve #2502 by enabling branching on a configuration option rather than on `order.has_checkout_step?`, thereby allowing shipments to be either built or not independent of the names of checkout steps.

### DIFF
--- a/core/app/models/spree/line_item.rb
+++ b/core/app/models/spree/line_item.rb
@@ -194,7 +194,7 @@ module Spree
     end
 
     def update_inventory
-      if (saved_changes? || target_shipment.present?) && order.has_checkout_step?("delivery")
+      if saved_changes? || target_shipment.present?
         Spree::OrderInventory.new(order, self).verify(target_shipment)
       end
     end
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -844,7 +844,7 @@ module Spree
     end
 
     def ensure_inventory_units
-      if has_checkout_step?("delivery")
+      if Spree::Config.build_shipment_predicate_class.call(self)
         inventory_validator = Spree::Stock::InventoryValidator.new
 
         errors = line_items.map { |line_item| inventory_validator.validate(line_item) }.compact
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -17,6 +17,7 @@
 # a.get :color
 # a.preferred_color
 #
+require 'spree/core/shipments/predicates/build'
 require "spree/core/search/base"
 require "spree/core/search/variant"
 require 'spree/preferences/configuration'
@@ -267,6 +268,16 @@ module Spree
     def available_currencies
       @available_currencies ||= ::Money::Currency.all
     end
+
+    # Allows implementing custom shipment building criterion based on an order
+    # instance. Warning: change this setting is not recommended.
+    # @!attribute [rw] build_shipment_predicate_class
+    #   @note Setting this to false is not recommended. This setting is provided
+    #     to help users who have made heavy modifications to disable shipments
+    #     in their store.
+    #   @return [Class] The predicate class that responds to #call(order) to
+    #     determine if shipments should be built.
+    class_name_attribute :build_shipment_predicate_class, default: 'Spree::Core::Shipments::Predicates::Build'
 
     # searcher_class allows spree extension writers to provide their own Search class
     class_name_attribute :searcher_class, default: 'Spree::Core::Search::Base'
--- a/core/lib/spree/core/shipments/predicates/build.rb
+++ b/core/lib/spree/core/shipments/predicates/build.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Spree
+  module Core
+    module Shipments
+      module Predicates
+        class Build
+          def self.call(_order)
+            true
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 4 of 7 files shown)


core/app/models/spree/line_item.rb
```diff
@@ -194,7 +194,7 @@ def handle_copy_price_override
     end
 
     def update_inventory
-      if (saved_changes? || target_shipment.present?) && order.has_checkout_step?("delivery")
+      if (saved_changes? || target_shipment.present?)
```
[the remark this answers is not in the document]
gevann: Resolved.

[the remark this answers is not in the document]
gevann: Resolved.

core/lib/spree/app_configuration.rb
```diff
@@ -88,6 +88,10 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Boolean]
     preference :binary_inventory_cache, :boolean, default: false
 
+    # @!attribute [rw] build_shipments
+    #   @return [Boolean] should shipments be built (default: +true+)
+    preference :build_shipments, :boolean, default: true
```
jhawthorn: This seems a little weird as a preference because it isn't a behaviour we properly want to support. Maybe we could add some wording here to dissuade users from setting it to false.

Maybe:

> This setting is provided to help users who have made heavy modifications to disable shipments in their store. It should not be changed.

core/lib/spree/core/shipments/predicates/build.rb
```diff
@@ -0,0 +1,13 @@
+module Spree
```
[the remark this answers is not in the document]
gevann: Resolved.

gevann:
I've addressed @kennyadsl concern, enabling stores that wish to do this to easily configure whether or not they build shipments based upon an order without having to decorate any spree classes.

frontend/app/views/spree/shared/_order_details.html.erb
```diff
@@ -7,7 +7,7 @@
       <%= render partial: 'spree/shared/address', locals: { address: order.bill_address } %>
     </div>
 
-    <% if order.has_checkout_step?("delivery") %>
+    <% if Spree::Config.build_shipment_predicate_class.call(self) %>
```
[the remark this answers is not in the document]
gevann: Changes made, thanks!

frontend/spec/features/checkout_spec.rb
```diff
@@ -498,9 +498,18 @@
...
+  context "order has only payment step and does not build shipments" do
+    class DisallowShipments
+      def self.call(_order)
+        false
+      end
+    end
+
     before do
       create(:credit_card_payment_method)
+      @original_predicate_class = Spree::Config.build_shipment_predicate_class
```
kennyadsl: why not using `let!` here?

gevann: I thought it would be simpler to leave it in the same `before` and `after` blocks where the rest of the setup and teardown were taking place. I can move it out to a `let!` if that's preferred.

kennyadsl: I tend to avoid instance variables in specs as much as possible and I'd like to even discourage future developers to use them (watching them here could make them feel like it's ok to use). I understand here it makes sense though. Maybe using `around` instead of `before/after` would fix the issue and will avoid using any instance variable?

gevann:
Requested changes have been made, thanks!

gevann:
@kennyadsl I believe your changes have been addressed.

core/lib/spree/app_configuration.rb
```diff
@@ -268,6 +269,16 @@ def available_currencies
       @available_currencies ||= ::Money::Currency.all
     end
 
+    # Allows implementing custom shipment building criterion based on an order
+    # instance. Warning: change this setting is not recommended.
+    # @!attribute [rw] build_shipment_predicate_class
+    #   @note Setting this to false is not recommended. This setting is provided
```
kennyadsl: Maybe `Setting this to false` is misleading? 
Is this statement referring to the result of the `call` method on the class specified in this preference? If yes maybe this is not the right place where to put this information or it needs to be adjusted?

core/app/models/spree/order.rb
```diff
@@ -844,7 +844,7 @@ def require_email
     end
 
     def ensure_inventory_units
-      if has_checkout_step?("delivery")
+      if Spree::Config.build_shipment_predicate_class.call(self)
```
kennyadsl: #### This comment is not blocking.

Maybe it's just me but I find this preference naming a bit complex to understand. For me it's not immediately clear what it does and what's the purpose of this preference. Maybe it would be better if it was wrapped into a `Spree::Order#needs_shipments_building?` or something similar? What do you think?

tvdeyen: I agree with this. And as changing configuration options in later releases is hard, we should agree on a better name before merging this.

jacobherrington:
Is work still being done on this @gevann?

gevann:
No. Work on this was stopped when it was originally approved.

Issue #2530, opened 2018-01-25
Improve the way stock item "Count On Hand" values are updated in the backend


benjaminwil:
In this issue, I want to argue that the functionality for changing the "Count On Hand" value for stock items (in the backend) could be more clear and simpler to use than it is now.

I'm going to set up a scenario that demonstrates why I think that the current implementation is error prone and forces administrators to do math that they don't actually need to do.

![product-stock-ui](https://user-images.githubusercontent.com/883581/35364655-3dec0930-0125-11e8-981d-b463b628cd7b.png)

#### Part 1

This part is just setting up a mock scenario for your store administrator in Part 2:

1. Create a Solidus sandbox and start it up.
2. In the backend, check the product stock for the "Ruby on Rails Mug" product. The "Count On Hand" value should be 10 by default.
3. Check that the Ruby on Rails Mug is backorderable. By default, it is.
4. In the frontend, add 15 Ruby on Rails Mug units to your cart and complete the checkout process.
5. Now, check the "Count On Hand" value for the Ruby on Rails Mug product. It should be -5.

#### Part 2

In this part, watch how one of your store administrators uses the backend UI and deals with physical products:

1. One of your store administrators received 100 new Ruby on Rails Mug units in the mail.
2. They log into the the backend, go to the **Stock -> Product Stock** page and change the Ruby on Rails Mug count on hand to 100.
3. They ship the 5 backordered mugs.
4. The Ruby on Rails Mug "Count On Hand" value still reads 100, even though you only have 95 mugs in available to sell.
5. Another store administrator logs into the backend and sees that there are 100 Ruby on Rails Mugs in stock. Misinformation continues to spread.

### Expected behavior

The backend UI is technically working as expected. It is the store administrator who is acting unexpectedly.

### Actual behavior

While you might say that the store administrator "should know better," than to enter 100 in this scenario, I don't know that Solidus is helping an administrator to _not_ make this mistake.

We could create a more complicated scenario where it is even fuzzier. Say you have 1387 mugs on hand, then you receive 150 mugs in the mail but 23 of them are broken. :-1: The store administrator should update the "Count On Hand" to be 1514--but they make a mistake and enter 1511 instead.

The "Count On Hand" value counts the items that are available to sell. The administrator *should* be able to figure out that if the on hand value is -5, and they have just received 100 mugs, that they need to update the count on hand to **95**, not 100. (`100 + -5 = 95`.)

[...]

PR #2682, opened 2018-04-06
Negative coh red

gevann:
This PR aims to improve the user experience of the admin, specifically with the goal of alleviating the #2530. Since we do not know if generally admins prefer to get _setting_ the count on hand (as they do now), or would rather mostly be _adjusting_ it instead, this PR aims instead to provide the admin better visual feedback such that they are more aware when the number they are dealing with is negative or not.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -7,6 +7,7 @@
 //= require spree/backend/product_picker
 //= require spree/backend/option_value_picker
 //= require spree/backend/taxons
+//= require spree/backend/highlight_negative_numbers
 
 /**
 This is a collection of javascript functions and whatnot
--- a/backend/app/assets/stylesheets/spree/backend/shared/_typography.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_typography.scss
@@ -98,6 +98,9 @@ dl {
 .red    { color: $color-5 }
 .yellow { color: $color-6 }
 
+.negative { color: $red }
+input[type="number"].negative { color: $red } // needed to override blue style that is more specific than '.negative'
+
 .no-objects-found {
   text-align: center;
   font-size: 120%;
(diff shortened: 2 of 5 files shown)


backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -136,3 +136,18 @@ Spree.ready(function(){
...
+    var isInputNumber = el instanceof HTMLInputElement && el.type == 'number';
+    if(isInputNumber){
+      if(el.value < 0){
+        el.classList.add("negative");
+      } else {
+        el.classList.remove("negative");
+      }
+    }
+  });
+});
```
tvdeyen: Could we move this in its own file? Then this is more easily adjustable by stores.

backend/app/assets/stylesheets/spree/backend/shared/_typography.scss
```diff
@@ -98,6 +98,8 @@ dl {
 .red    { color: $color-5 }
 .yellow { color: $color-6 }
 
+.negative { color: $color-5 !important }
```
tvdeyen: Could we use a meaningful color variable (we have `$red`) here?

tvdeyen: Do we really need the `!important` here? Especially with this generic class this could cause headaches. Would a stronger selector like

```css
input[type="number"].negative
```

work as well?

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -136,3 +136,18 @@ Spree.ready(function(){
     });
   }
 });
+
+Spree.ready(function() {
+  // Highlight negative numbers in red.
+  document.querySelector('body').addEventListener('input', function(e){
+    var el = e.target;
+    var isInputNumber = el instanceof HTMLInputElement && el.type == 'number';
+    if(isInputNumber){
```
tvdeyen: Could we please the human eye by adding some spacing here?

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -136,3 +136,18 @@ Spree.ready(function(){
...
+      if(el.value < 0){
```
tvdeyen: Ditto?

tvdeyen:
Note to self. Enable javascript syntax linting on HoundCI

gevann:
Requested changes made, thanks!

PR #2684, opened 2018-04-06
UI fixes for admin locale select

tvdeyen:
Now that we have a admin local select and store owners can configure available locales for their stores we should improve the UI/UX a little bit.

1. ~~Admins should only be able to choose from configured available locales instead of all locales in the system~~  

2. The store administration form uses two columns to display all the fields.

3. The locale select uses select2, because it is a multi select and this is where select2 still shines. 

4. When the local select is displayed alone, as on the login screen, it has correct margins.

### Screenshot

sample store - stores 2018-04-07 01-04-12

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
@@ -162,7 +162,6 @@ nav.menu {
 
 .admin-locale-selection {
   margin: 1em;
-  margin-bottom: 0;
 }
 
 .admin-nav.fits .admin-nav-footer {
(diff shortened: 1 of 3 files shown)


backend/app/views/spree/admin/shared/_locale_selection.html.erb
```diff
@@ -1,7 +1,7 @@
-<% available_locales = Spree.i18n_available_locales %>
+<% available_locales = current_store.available_locales %>
```
jhawthorn: I think the `current_store.available_locales` should apply to the frontend only

tvdeyen: Ideally we would have two configurations, but this seems overkill to me. We could keep it as is, if we encourage the users of `solidus_i18n` to copy over only the files they need in their app (what I recommend anyway).

tvdeyen:
We should be more descriptive then

monosnap 2018-04-07 01-13-06

backend/app/views/spree/admin/shared/_locale_selection.html.erb
tvdeyen: Reverted

PR #2686, opened 2018-04-07
Remove default address dependency

### LINKED ISSUE #2645: When a reimbursement is created, a store credit is not created only a store reimbursement credit.
### Steps to reproduce
Ship order 
Create returns authorisation
Create customer return with reimbursement type of `Spree::StoreCredit`
Create reimbursement

### Expected behavior
Spree::StoreCredit created

### Actual behavior
Reimbursement credit is created with a `creditable_type` of `Spree::StoreCredit` but a `creditable_id` of `nil`. 

### System configuration
**Solidus Version**: 2.4.2

After falling down the rabbit hole debugging this I think I have worked out what is going on. 

    module Spree::ReimbursementType::ReimbursementHelpers
      #...
      def create_creditable(reimbursement, unpaid_amount)
        Spree::Reimbursement::Credit.default_creditable_class.new(
          user: reimbursement.order.user,
          amount: unpaid_amount,
          category: Spree::StoreCreditCategory.reimbursement_category(reimbursement),
          created_by: Spree::StoreCredit.default_created_by,
          memo: "Refund for uncreditable payments on order #{reimbursement.order.number}",
          currency: reimbursement.order.currency
        )
      end
    end

The first issue is that the `create_creditable` method requires a `created_by` attribute to be set, and `Spree::StoreCredit.default_created_by`, which turns out to be... 

    class Spree::StoreCredit < Spree::PaymentSource
      #...
      DEFAULT_CREATED_BY_EMAIL = "spree@example.com"
      #....
      class << self
        def default_created_by
          Spree.user_class.find_by(email: DEFAULT_CREATED_BY_EMAIL)
        end
      end
    end

So unless you still have a user with `spree@example.com` as an email address in your database the store credits will never save.

gevann:
This PR is _blocked on #2687_.

This PR aims to resolve #2645 by eliminating a hard-coded dependency on a user in the database with the email address `'spree@example.com'`. In order to achieve this, the user that creates the reimbursement has to be passed down the call chain from the backend. Since the call chain for reimbursements is particularly long, this PR has many changes in it, however they are all aimed at ensuring that when a `Spree::StoreCredit` is eventually created by a reimbursement, that it has as the creator without having to rely on specific email addresses existing.

### DIFF
--- a/backend/app/controllers/spree/admin/reimbursements_controller.rb
+++ b/backend/app/controllers/spree/admin/reimbursements_controller.rb
@@ -13,7 +13,7 @@ module Spree
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
       def perform
-        @reimbursement.perform!
+        @reimbursement.perform!(try_spree_current_user)
         redirect_to location_after_save
       end
 
@@ -57,7 +57,7 @@ module Spree
       end
 
       def load_simulated_refunds
-        @reimbursement_objects = @reimbursement.simulate
+        @reimbursement_objects = @reimbursement.simulate(try_spree_current_user)
       end
 
       def spree_core_gateway_error(error)
--- a/core/app/models/spree/order_cancellations.rb
+++ b/core/app/models/spree/order_cancellations.rb
@@ -83,13 +83,17 @@ class Spree::OrderCancellations
   # @api public
   # @param [Array<InventoryUnit>] inventory_units the inventory units to be reimbursed
   # @return [Reimbursement] the reimbursement for inventory being canceled
-  def reimburse_units(inventory_units)
+  def reimburse_units(inventory_units, creator = nil)
+    unless creator
+      creator = Spree.user_class.where(email: 'spree@example.com').first
+      Spree::Deprecation.warn("Calling #reimburse_units on #{self} without creator is deprecated")
+    end
     reimbursement = nil
 
     Spree::OrderMutex.with_lock!(@order) do
       return_items = inventory_units.map(&:current_or_new_return_item)
       reimbursement = Spree::Reimbursement.new(order: @order, return_items: return_items)
-      reimbursement.return_all
+      reimbursement.return_all(creator)
     end
 
     reimbursement
--- a/core/app/models/spree/reimbursement.rb
+++ b/core/app/models/spree/reimbursement.rb
@@ -98,12 +98,16 @@ module Spree
       total - paid_amount
     end
 
-    def perform!
+    def perform!(creator = nil)
+      unless creator
+        creator = Spree.user_class.where(email: 'spree@example.com').first
+        Spree::Deprecation.warn("Calling #perform on #{self} without creator is deprecated")
+      end
       reimbursement_tax_calculator.call(self)
       reload
       update!(total: calculated_total)
 
-      reimbursement_performer.perform(self)
+      reimbursement_performer.perform(self, creator)
 
       if unpaid_amount_within_tolerance?
         reimbursed!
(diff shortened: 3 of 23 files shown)


gevann:
This test passes for me locally, and seems unlikely to have been caused by these changes. If anybody has any guidance on where to get started on resolving this one I would much appreciate it! 😄

jhawthorn:
> This PR is blocked on #2685.

Do you mean #2687?

gevann:
Yes I did.

gmacdougall:
My main concern about this is that we're changing the public API for some of these key methods.

Could we user `creator = nil` as a default action, with deprecation warnings that maintain the current behaviour until the next major version release?

gevann:
I see your point. I've removed the default user constant in this PR - what do you think about rather than using `creator=nil` that  we use `creator=<user>` where `<user>` is the `'spree@example.com'` user it used to be? That way nobody who wants the current behaviour will have to change anything but we can still remove that constant and stop depending on it?

gmacdougall:
Yeah. That's what I was thinking.

gevann:
I'll make that change, thanks!

gevann:
I have run this locally and it passes for me with the same seed, with `DB=postgres`. Does anybody have any ideas of what I can poke in to here to replicate? Does this fail or pass for any of you?

gevann:
Fixed. Thanks for the feedback!

gevann:
@kennyadsl you are correct, thanks!

core/app/models/spree/order_cancellations.rb
```diff
@@ -83,13 +83,17 @@ def cancel_unit(inventory_unit, reason: Spree::UnitCancel::DEFAULT_REASON, whodu
   # @api public
   # @param [Array<InventoryUnit>] inventory_units the inventory units to be reimbursed
   # @return [Reimbursement] the reimbursement for inventory being canceled
-  def reimburse_units(inventory_units)
+  def reimburse_units(inventory_units, creator = nil)
```
tvdeyen: Could we use a named argument here?

```rb
def reimburse_units(inventory_units, creator: nil)
```

core/app/models/spree/reimbursement.rb
```diff
@@ -98,12 +98,16 @@ def unpaid_amount
       total - paid_amount
     end
 
-    def perform!
+    def perform!(creator = nil)
```
tvdeyen: A named parameter would help the self-documentation of this method.

backend/app/controllers/spree/admin/reimbursements_controller.rb
```diff
@@ -13,7 +13,7 @@ class ReimbursementsController < ResourceController
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
       def perform
-        @reimbursement.perform!
+        @reimbursement.perform!(try_spree_current_user)
```
tvdeyen: A named argument would explain the purpose of this argument:

```rb
@reimbursement.perform!(creator: try_spree_current_user)
```

backend/app/controllers/spree/admin/reimbursements_controller.rb
```diff
@@ -57,7 +57,7 @@ def load_stock_locations
       end
 
       def load_simulated_refunds
-        @reimbursement_objects = @reimbursement.simulate
+        @reimbursement_objects = @reimbursement.simulate(try_spree_current_user)
```
tvdeyen: Here as well

```rb
@reimbursement.simulate(creator: try_spree_current_user)
```

core/app/models/spree/reimbursement.rb
```diff
@@ -116,12 +120,16 @@ def perform!
       end
     end
 
-    def simulate
+    def simulate(creator = nil)
```
tvdeyen: Ditto: Named argument

jacobherrington:
@gevann is this something you're still working on?

kennyadsl:
@jacobherrington I've opened #2802 to complete this one, I think we can close this.

PR #2687, opened 2018-04-09
Remove order-dependency from locale specs

gevann:
Previously, `bundle exec rspec --seed 13070 --format documentation` would fail in the front-end app due to test ordering. This PR removes this order dependency.


gevann:
Cannot reproduce this test failure when running `DB=postgresql bundle exec rspec --seed 23685`; does anybody have any suggestions?

PR #2690, opened 2018-04-10
Rails 5.2

jhawthorn:
This PR isn't totally ready to merge (needs some unreleased changes in other gems), but Rails 5.2 is now out.

We need:
* [x] New gem of paranoia
* [x] New gem of canonical-rails
* [x] New gem of friendly_id

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  postgres:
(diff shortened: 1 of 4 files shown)


jhawthorn:
> The first commit seems to be empty. Is it still useful?

Nope! I merged it a while ago in a separate PR

PR #2691, opened 2018-04-10
Remove assert_written_to_cache

jhawthorn:
This is error prone and breaks in Rails 5.2, since the key being logged is an array.

The key we see from AS::Notifications is now an array and has additional hashes on the template names (I assume russian doll caching related). It's too hard to predict where these values come from, so we can't continue testing this.

We should have most of the benefit and testing of this by just counting the number of cache writes (which we were doing already).

### DIFF
--- a/core/lib/spree/testing_support/caching.rb
+++ b/core/lib/spree/testing_support/caching.rb
@@ -7,16 +7,6 @@ module Spree
         @cache_write_events
       end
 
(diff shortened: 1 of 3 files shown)


jhawthorn:
I'd like to avoid removing even testing related methods in a SemVer minor release, but I think it's unlikely this is widely used, and it would break for everyone in Rails 5.2 regardless.

Issue #2692, opened 2018-04-12
Unauthorized redirections fail on Solidus engines mounted with namespace


spaghetticode:
When Solidus routes are mounted with a custom namespace, for example:

```mount Spree::Core::Engine, at: '/shop'```

then the app frontend redirects to `/unauthotized` instead of `/shop/unauthorized`. 
Tested with Solidus master branch.

kennyadsl:
Closing this since this should be changed in the auth system as done in [solidus_auth_devise](https://github.com/solidusio/solidus_auth_devise/blob/01459b4339b0130241886bbcc3b35af1853f1e74/lib/spree/auth/engine.rb#L58-L67), as suggested in the [related PR](https://github.com/solidusio/solidus/pull/2693).

PR #2693, opened 2018-04-12
Enable route namespacing on unauthorized redirects

spaghetticode:
Hardcoded URLs don't work as expected when Solidus routes are mounted within a
namespace: the namespace part is simply lost. In order to avoid this problem
and include the namespace, when present, a routing helper must be used.

This fixes issue #2692


spaghetticode:
I am now thinking I should close this PR. I believe that the simplest way to solve my issue is to customize the `BaseController.unauthorized_redirect` lambda where necessary, just like the gem `solidus_auth_devise` does in https://github.com/solidusio/solidus_auth_devise/blob/master/lib/spree/auth/engine.rb

PR #2695, opened 2018-04-13
Add frontend customer orders page

### LINKED ISSUE #4010: Legacy backend users index breaks with a custom `Spree.user_class`
Using a Spree.user_class other than 'User' (e.g. Account) causes users/index view to break because the method link_to_edit uses a polymorphic method of determining the URL helper and it results in 'undefined method edit_admin_account_url' error. It appears like Solidus needs this  edit_admin_user_url.  

**Solidus Version**:
2.11.7

**To Reproduce**
Setup Spree to use a user_class called 'Account' and visit the admin users index page. It will error.

spaghetticode:
The code is basically a cut & paste of the relevant parts from the gem
solidus_auth_devise.

The rationale for this change is that the orders history page better belongs
inside the solidus main project, because the feature does not have much to do
with the actual authentication process, and because it allows projects that
have an existing authentication system (ie. don't use solidus_auth_devise) to
have the orders page feature out of the box.

Closes #4010

### DIFF
--- a/frontend/app/controllers/spree/users_controller.rb
+++ b/frontend/app/controllers/spree/users_controller.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+class Spree::UsersController < Spree::StoreController
+  skip_before_action :set_current_order, only: :show, raise: false
+  prepend_before_action :load_object, only: :show
+
+  include Spree::Core::ControllerHelpers
+
+  def show
+    @orders = @user.orders.complete.order('completed_at desc')
+  end
+
+  private
+
+  def load_object
+    @user ||= Spree.user_class.find_by(id: try_spree_current_user&.id)
+    authorize! params[:action].to_sym, @user
+  end
+
+  def accurate_title
+    I18n.t('spree.my_account')
+  end
+end
--- a/frontend/app/views/spree/users/show.html.erb
+++ b/frontend/app/views/spree/users/show.html.erb
@@ -1,0 +1,43 @@
+<h1><%= accurate_title %></h1>
+
+<div data-hook="account_summary" class="account-summary">
+  <dl id="user-info">
+    <dt><%= t('spree.email') %></dt>
+    <dd><%= @user.email %> (<%= link_to t('spree.edit'), edit_account_path %>)</dd>
+  </dl>
+</div>
+
+<div data-hook="account_my_orders" class="account-my-orders">
+
+  <h3><%= t('spree.my_orders') %></h3>
+  <% if @orders.present? %>
+    <table class="order-summary">
+      <thead>
+      <tr>
+        <th class="order-number"><%= Spree::Order.human_attribute_name(:number) %></th>
+        <th class="order-date"><%= Spree::Order.human_attribute_name(:date) %></th>
+        <th class="order-status"><%= Spree::Order.human_attribute_name(:status) %></th>
+        <th class="order-payment-state"><%= Spree::Order.human_attribute_name(:payment_state) %></th>
+        <th class="order-shipment-state"><%= Spree::Order.human_attribute_name(:shipment_state) %></th>
+        <th class="order-total"><%= Spree::Order.human_attribute_name(:total) %></th>
+      </tr>
+      </thead>
+      <tbody>
+      <% @orders.each do |order| %>
+        <tr class="<%= cycle('even', 'odd') %>">
+          <td class="order-number"><%= link_to order.number, order_url(order) %></td>
+          <td class="order-date"><%= l order.completed_at.to_date %></td>
+          <td class="order-status"><%= t("spree.order_state.#{order.state}") %></td>
+          <td class="order-payment-state"><%= t("spree.payment_states.#{order.payment_state}") if order.payment_state %></td>
+          <td class="order-shipment-state"><%= t("spree.shipment_states.#{order.shipment_state}") if order.shipment_state %></td>
+          <td class="order-total"><%= order.display_total %></td>
+        </tr>
+      <% end %>
+      </tbody>
+    </table>
+  <% else %>
+    <p><%= t('spree.you_have_no_orders_yet') %></p>
+  <% end %>
+  <br>
+
+</div>
--- a/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
+++ b/frontend/app/assets/stylesheets/spree/frontend/screen.css.scss
@@ -1117,6 +1117,12 @@ table.order-summary {
             color: $link_text_color;
           }
         }
+
+        &.order-status,
+        &.order-payment-state,
+        &.order-shipment-state {
+          text-transform: capitalize;
+        }
       }
     }
   }
--- a/frontend/config/routes.rb
+++ b/frontend/config/routes.rb
@@ -31,4 +31,6 @@ Spree::Core::Engine.routes.draw do
   get '/unauthorized', to: 'home#unauthorized', as: :unauthorized
   get '/content/cvv', to: 'content#cvv', as: :cvv
   get '/cart_link', to: 'store#cart_link', as: :cart_link
+
+  resource :account, controller: 'users', only: [:show, :edit]
 end
--- a/frontend/spec/controllers/spree/users_controller_spec.rb
+++ b/frontend/spec/controllers/spree/users_controller_spec.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Spree::UsersController, type: :controller do
+  let(:user) { Spree.user_class.new }
+
+  describe '#show' do
+    context 'when not authenticated' do
+      before { allow(Spree.user_class).to receive(:find_by) { nil } }
+
+      it 'redirects to unauthorized path' do
+        get :show
+        expect(response).to redirect_to '/unauthorized'
+      end
+    end
+
+    context 'when authenticated' do
+      before { allow(Spree.user_class).to receive(:find_by) { user } }
+
+      it 'redirects to signup path if user is not found' do
+        get :show
+        expect(response).to be_successful
+      end
+    end
+  end
+end
--- a/frontend/spec/features/user_orders_spec.rb
+++ b/frontend/spec/features/user_orders_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe "User orders page", type: :feature do
+  before do
+    allow_any_instance_of(Spree::UsersController).to receive_messages(try_spree_current_user: user)
+  end
+
+  context 'when the user is not authenticated' do
+    let(:user) { nil }
+
+    it 'fails authentication' do
+      visit spree.account_path
+      expect(page).to have_content 'Authorization Failure'
+    end
+  end
+
+  context 'when the user is authenticated' do
+    let(:user) { create :user }
+
+    let!(:cart_order) { create :order, user: user }
+    let!(:complete_order) { create :completed_order_with_totals, user: user }
+
+    it 'lists user complete orders' do
+      visit spree.account_path
+      expect(page).to have_content 'My Account'
+      expect(page).to have_content user.email
+      expect(page).to have_content complete_order.number
+      expect(page).not_to have_content cart_order.number
+    end
+  end
+end


kennyadsl:
It makes sense but we should find a way to handle duplication of routes/controllers/views for users that will update `solidus` but not `solidus_auth_devise`, where we should also remove those files if they are already here. We should conditionally avoid including those files in some way? maybe Solidus version number?

kennyadsl:
Also, we are thinking to move all the controller/views/helpers from solidus_auth_devise to core, keeping `solidus_auth_devise` as solely an authentication system, so it's better to wait until we finalize this decision. Any thought is welcome.

frontend/app/controllers/spree/users_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  private
+
+  def load_object
+    @user ||= spree_current_user
+    authorize! params[:action].to_sym, @user
+  end
+
+  def accurate_title
+    Spree.t(:my_account)
```
tvdeyen: We deprecated `Spree.t` in favor of `I18n.t(:my_account, scope: :spree)`. 
And I don't see why we define this in the controller. Just use `t(:my_account, scope: :spree)` directly in the view

frontend/app/views/spree/shared/_login_bar_items.html.erb
```diff
@@ -1 +1,4 @@
 <%# solidus_auth_devise or a custom auth system can replace this partial %>
+<% if try_spree_current_user %>
+  <li><%= link_to Spree.t(:my_account), spree.account_path %></li>
```
tvdeyen: We deprecated `Spree.t` in favor of `I18n.t(:my_account, scope: :spree)`.

frontend/app/views/spree/users/show.html.erb
```diff
@@ -0,0 +1,48 @@
+<h1><%= accurate_title %></h1>
+
+<div data-hook="account_summary" class="account-summary">
+  <dl id="user-info">
+    <dt><%= Spree.t(:email) %></dt>
```
tvdeyen: We deprecated `Spree.t` in favor of `I18n.t(:email, scope: :spree)`.

frontend/app/views/spree/users/show.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    <dd>
+      <%= @user.email %>
+      <% if spree.respond_to? :edit_account_path %>
+        (<%= link_to Spree.t(:edit), spree.edit_account_path %>)
```
tvdeyen: We deprecated `Spree.t` in favor of `I18n.t(:edit, scope: :spree)`.

frontend/app/views/spree/users/show.html.erb
```diff
@@ -0,0 +1,48 @@
...
+</div>
+
+<div data-hook="account_my_orders" class="account-my-orders">
+
+  <h3><%= Spree.t(:my_orders) %></h3>
+  <% if @orders.present? %>
+    <table class="order-summary">
+      <thead>
+      <tr>
+        <th class="order-number"><%= I18n.t(:number, scope: 'activerecord.attributes.spree/order') %></th>
```
tvdeyen: Please use

```rb
Spree::Order.human_attribute_name(:number)
```

frontend/app/views/spree/users/show.html.erb
```diff
@@ -0,0 +1,48 @@
...
+        <th class="order-date"><%= Spree.t(:date) %></th>
+        <th class="order-status"><%= Spree.t(:status) %></th>
+        <th class="order-payment-state"><%= Spree.t(:payment_state) %></th>
+        <th class="order-shipment-state"><%= Spree.t(:shipment_state) %></th>
+        <th class="order-total"><%= Spree.t(:total) %></th>
```
tvdeyen: Please use

```rb
Spree::Order.human_attribute_name
```

for all the attributes here

frontend/app/views/spree/users/show.html.erb
```diff
@@ -0,0 +1,48 @@
...
+      </tr>
+      </thead>
+      <tbody>
+      <% @orders.each do |order| %>
+        <tr class="<%= cycle('even', 'odd') %>">
+          <td class="order-number"><%= link_to order.number, order_url(order) %></td>
+          <td class="order-date"><%= l order.completed_at.to_date %></td>
+          <td class="order-status"><%= Spree.t("order_state.#{order.state}").titleize %></td>
```
tvdeyen: Please do not use `titleize` here. As this is cosmetic change that should be done solely in CSS

tvdeyen: Also do not use `Spree.t`

kennyadsl:
By merging this PR stores will end up having the same route/action/view defined in both `solidus_frontend` and `solidus_auth_devise`. Do you see a possible solution to this?

spaghetticode:
@kennyadsl regarding your question about routes, what you wrote is certainly true, but I think I am missing something as I don't see why this can be a problem. Can you explain the reason?

kennyadsl:
Sure, I'm thinking that if a user upgrades to latest Solidus (that will include code from `solidus_auth_devise`) he will have the same code defined in both `solidus` and `solidus_auth_devise`. If we want to change something in core I'm afraid some changes (like changes on views) could not be doable since they will be overridden from the same files in the extension since they should be loaded later, right?

spaghetticode:
I think I understand your concerns. Looking at a concrete example, `spree_auth_devise` file `lib/frontend/spree/users/show.html.erb` overwrites the new file `frontend/app/views/spree/users/show.html.erb`. 

One way to look at this, is to consider the new files in Solidus codebase as sensible defaults that are naturally overwritten by the current version of `solidus_auth_devise`.  After all, their code is the same right now, so no harm done. 

Issues arise when one of the two files changes. 

Solidus changes will be masked by `spree_auth_devise` file as long as that file exist also in the gem. I think this first scenario is not a big issue, again as I'd rather consider Solidus file as (hidden) default, at least initially. `spree_auth_devise` is still in charge.

IMHO a more significant issue is when the file inside `spree_auth_devise` gets updated, as the updates should probably also be manually backported to Solidus' file.

I think one way to minimize code backporting may be to address all current issues in `spree_auth_devise`, this should give us some stability, and basically freeze that gem's development. 
As soon as a new version of Solidus with the new file is released, a new version of  `spree_auth_devise` stripped of the very same file can be released. The new gem version would require the new Solidus release, of course. 

What do you think? Does this strategy seem too clumsy to you?

kennyadsl:
@spaghetticode your proposed solution makes sense but no one can ensure us that user will update `solidus_auth_devise` after the `Solidus` update. 

We should think at least to some warning or suggestion into the changelog entry that asks users to update `solidus_auth_devise` along with `solidus`. If we go in this direction, I'd prefer to ask/force Solidus users to update `solidus_auth_devise` once, by moving all the code unrelated to authentication from `solidus_auth_devise` to `solidus` in a single version release.

spaghetticode:
@kennyadsl 👍 on deprecations/warnings when an old version of `solidus_auth_devise` is found.

jarednorman:
This looks like a good change that we can roll into v3.0, but it requires a little bit of coordination.

`solidus_auth_devise` is currently pinned to Solidus versions less than v3.0. In order to avoid conflicts where someone is running a version of `solidus_auth_devise` that defines this controller on a version of Solidus that also does, we'll have to merge this change into v3.0 and cut a new version of `solidus_auth_devise` at the same time that removes the controller and bumps the version requirement to Solidus `>= 3.0`.

spaghetticode:
I've been experimenting a bit with this PR, and I think it is safe to merge anytime, as the code in `solidus_auth_devise` will always take precedence over the one defined here. I tested this in both Rails 5.2 and 6.0 (Zeitwerk) and basically what happens is that extension directories come first in `autoload_paths` (Rails 5.2) and `root_dirs` (Zeitwerk) compared to the ones defined in Solidus. For example, this is `autoload_paths` from a sandbox application:

```
/Users/andrea/solidus/sandbox_5.2/app/assets
/Users/andrea/solidus/sandbox_5.2/app/channels
/Users/andrea/solidus/sandbox_5.2/app/controllers
/Users/andrea/solidus/sandbox_5.2/app/helpers
/Users/andrea/solidus/sandbox_5.2/app/jobs
/Users/andrea/solidus/sandbox_5.2/app/mailers
/Users/andrea/solidus/sandbox_5.2/app/models
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/solidus_auth_devise-2.5.2/lib/decorators/backend/controllers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/solidus_auth_devise-2.5.2/lib/decorators/frontend/controllers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/solidus_auth_devise-2.5.2/app/mailers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/solidus_auth_devise-2.5.2/app/models
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/solidus_auth_devise-2.5.2/lib/controllers/backend
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/solidus_auth_devise-2.5.2/lib/controllers/frontend
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/devise-4.7.3/app/controllers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/devise-4.7.3/app/helpers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/devise-4.7.3/app/mailers
/Users/andrea/solidus/frontend/app/assets
/Users/andrea/solidus/frontend/app/controllers
/Users/andrea/solidus/frontend/app/helpers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/canonical-rails-0.2.11/app/helpers
/Users/andrea/solidus/backend/app/assets
/Users/andrea/solidus/backend/app/controllers
/Users/andrea/solidus/backend/app/helpers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/font-awesome-rails-4.7.0.5/app/assets
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/font-awesome-rails-4.7.0.5/app/helpers
/Users/andrea/solidus/api/app/controllers
/Users/andrea/solidus/api/app/helpers
/Users/andrea/solidus/core/app/assets
/Users/andrea/solidus/core/app/controllers
/Users/andrea/solidus/core/app/helpers
/Users/andrea/solidus/core/app/jobs
/Users/andrea/solidus/core/app/mailers
/Users/andrea/solidus/core/app/models
/Users/andrea/solidus/core/app/models/concerns
/Users/andrea/solidus/core/app/subscribers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/activestorage-5.2.4.4/app/assets
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/activestorage-5.2.4.4/app/controllers
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/activestorage-5.2.4.4/app/controllers/concerns
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/activestorage-5.2.4.4/app/javascript
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/activestorage-5.2.4.4/app/jobs
/Users/andrea/.rvm/gems/ruby-2.6.3/gems/activestorage-5.2.4.4/app/models
/Users/andrea/solidus/sandbox5/test/mailers/previews
```
So, for example, `Spree::UsersController` will be found first in `solidus_auth_devise` and the one in `solidus_frontend` will never be loaded. 

The code in this PR becomes alive only when its counterpart is removed from `solidus_auth_devise`.

kennyadsl:
@spaghetticode What about the routes? Don't we have a conflict of both core and solidus_auth_devise defining the same one?

spaghetticode:
@kennyadsl I didn't find any issue with that when I tested the feature. The last routing file loaded will simply override the previous route, if they're the same.

tvdeyen (approved):
I think this is fine to be merged as @spaghetticode already stated.
I also agree with @jarednorman that we should cut a appropriate release of solidus_auth_devise.

Thanks for this. This is very helpful for people just getting started 🙏🏻

waiting-for-dev:
Closing as we've deprecated solidus_frontend. We could move it to solidus_starter_frontend, though.

PR #2696, opened 2018-04-13
cleanup order validate_payments_attributes

ccarruitero:
I think it's not necessary make a query over `available_payment_methods`. Until I can see `available_payment_methods` already return a list of payment methods that are available for the order.


tvdeyen:
Thanks

PR #2704, opened 2018-04-17
Improve the admin UX for a product's "Available On" field

benjaminwil:
In the `solidus_backend` interface, a product's "Available On" field should be used set to a date.

If `available_on` is not set, then the product is not displayed on the storefront.

Because it's not technically a required field, it might not be clear to an admin why their new product does not appear on the storefront after it's been created. (The reason is because `available_on` has no value.)

This PR addresses the "Available On" UX in the following ways:

- It improves the admin UX by making it clear that the "Available On" field is a date picker.
- the "Available On" field now has placeholder text that clearly states "Hey, this is a *date picker* that sets the *availability* date for the product."
- Lastly, there was a SASS variable called `$input-placeholder-color` that wasn't being set anywhere, so I set it. This is the color that is now used for any `<input>` tags placeholder text across the admin now.

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -64,6 +64,7 @@
       <div data-hook="admin_product_form_available_on">
         <%= f.field_container :available_on do %>
           <%= f.label :available_on %>
+          <%= f.field_hint :available_on %>
 
           <%= render "spree/admin/shared/datepicker", f: f, date_attr: :available_on %>
 
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -64,8 +64,10 @@
       <div data-hook="admin_product_form_available_on">
         <%= f.field_container :available_on do %>
           <%= f.label :available_on %>
+
+          <%= render "spree/admin/shared/datepicker", f: f, date_attr: :available_on %>
+
           <%= f.error_message_on :available_on %>
-          <%= f.text_field :available_on, value: datepicker_field_value(@product.available_on), class: 'datepicker' %>
         <% end %>
       </div>
 
--- a/backend/app/views/spree/admin/shared/_datepicker.html.erb
+++ b/backend/app/views/spree/admin/shared/_datepicker.html.erb
@@ -1,0 +1,8 @@
+<div class="input-group">
+  <div class="input-group-prepend">
+    <span class="input-group-text">
+      <%= solidus_icon('fa fa-calendar') %>
+    </span>
+  </div>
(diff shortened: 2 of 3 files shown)


backend/app/views/spree/admin/shared/_datepicker.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="input-group">
+  <div class="input-group-prepend">
+    <span class="input-group-text">
+      <%= icon('fa fa-calendar') %>
```
tvdeyen: `icon` is deprecated (deprecations let tests fail), please use `solidus_icon`.

backend/app/views/spree/admin/shared/_datepicker.html.erb
```diff
@@ -0,0 +1,8 @@
...
+    </span>
+  </div>
+  <%= f.text_field date_attr, value: datepicker_field_value(f.object.public_send(date_attr)), class: 'form-control datepicker', placeholder: placeholder_attr %>
```
tvdeyen: Could we add an option to set the `required` attribute?

backend/app/views/spree/admin/products/_form.html.erb
```diff
@@ -64,8 +64,10 @@
       <div data-hook="admin_product_form_available_on">
         <%= f.field_container :available_on do %>
           <%= f.label :available_on %>
+
+          <%= render "spree/admin/shared/datepicker", f: f, date_attr: :available_on, placeholder_attr: t('spree.available_on_placeholder') %>
```
tvdeyen: Would be great to have an option to set the `required` attribute.

backend/app/views/spree/admin/shared/_datepicker.html.erb
benjaminwil: Would making "Available On" input required complicate the product creation process in other ways?

I was hesitant to suggest this for this PR as I think it is another can of worms.

benjaminwil:
@tvdeyen Thank you for the comments.

I wonder if a `required` attribute would complicate product creation and product imports? While I agree that it would improve the admin UX it might be a bigger issue than this PR should address.

benjaminwil:
@kennyadsl Thank you for pointing this out.

I am not sure what the best way to deal with placeholder text should be. It may be confusing to use a standard date like `01-01-1985`. The placeholder text would also have to account for different date formats if the user is using a format like `MM.DD.YY` instead. For now, I have just removed the placeholder all together.

I have instead added a more descriptive tooltip.

benjaminwil:
@kennyadsl Thanks for your patience – I should dropped those changes before.

I've now dropped those older commits entirely.

tvdeyen:
Thanks, Ben

Issue #2653, opened 2018-03-27
Spree::Shipment#cartons is an array


aldesantis:
This is happening because the `cartons` association in `Spree::Shipment` uses `uniq` instead of `distinct`, and `uniq` returns an array as of Rails 5.1.

The following monkey-patch fixes the problem:

```ruby
Spree::Shipment.class_eval do
  has_many :cartons, -> { distinct }, through: :inventory_units
end
```

### Steps to reproduce

* Call `Spree::Shipment#cartons` in any Solidus app

### Expected behavior

An AR collection proxy should be returned.

### Actual behavior

```
NoMethodError: undefined method `extensions' for []:Array
Did you mean?  extend
from /Users/aldesantis/.rvm/gems/ruby-2.5.0@myproject/gems/activerecord-5.1.5/lib/active_record/associations/association.rb:140:in `extensions'
```

### System configuration

**Solidus Version**: 2.3

**Extensions in use**: 

* solidus_affirm
* solidus_auth_devise
* solidus_comments
* solidus_gateway
* solidus_paypal_braintree
* solidus_product_assembly
* solidus_redirector
* solidus_related_products
* solidus_reviews
* solidus_sale_prices
* solidus_sendwithus
* solidus_sitemap
* solidus_social
* solidus_static_content
* spree_paypal_express
* solidus_trackers

Leoxxid:
I can do this? I want start in the open source world.

jhawthorn:
Fixed by #2653

PR #2710, opened 2018-04-20
Replace `uniq` with `distinct` for cartons association

DanielePalombo:
This PR is going to fix the issue #2653 

Spree::Shipment#cartons should return a carton collection, to do this `distinct` is provided instead of `uniq` as scope parameter.


jhawthorn:
Thank you!

PR #2713, opened 2018-04-24
Remove hardcoded admin store attributes

jtapia:


core/config/locales/en.yml
```diff
@@ -595,6 +595,7 @@ en:
       spree/store:
         one: Store
         other: Stores
+        default: Default
```
tvdeyen: This needs to be in the `activerecord.attributes` section

jtapia:
@tvdeyen right, my bad, fixed

core/config/locales/en.yml
jtapia: fixed

PR #2718, opened 2018-04-26
Add a documentation contributors guide

benjaminwil:
I tried to keep this short and sweet. This is a list of guidelines for anyone wanting to contribute documentation.

I don't expect all of these to get followed 100% of the time, but I think it is a useful resource nonetheless.

### DIFF
--- a/guides/contributing.md
+++ b/guides/contributing.md
@@ -1,6 +1,6 @@
 # Contributing
 
-We hope that you will consider contributing to the Solidus documentation. 
+We hope that you will consider contributing to the Solidus documentation.
 
 We want to provide the Solidus community with consistent, easy-to-read, and
 easy-to-maintain articles. If you decide to submit a pull request, please try to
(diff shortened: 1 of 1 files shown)


guides/contributing.md
```diff
@@ -0,0 +1,60 @@
...
+- Prefer to write in the present tense. [The present tense is simpler and easier
+  to read][present-tense].
+
+[present-tense]: https://www.plainlanguage.gov/guidelines/conversational/use-the-present-tense/
+
+### Code
+
+- Do not abbreviate terminal commands. (`bundle exec rails server`, not `bundle
+  exec rails s`.)
+- If your code is longer than one line or 20 characters, use a code block rather
```
tvdeyen: Code should be in blocks if it’s longer than one word. Otherwise it may break the line and this isn’t really readable. 

I would even consider to put code always in a block, so it stands out. Code inside paragraphs seldomly make sense. Only if you want to reference a `variable` or `constant` it makes sense. Also syntax highlighting only works for code blocks.

benjaminwil: Good call, thank you. Fixed.

PR #2719, opened 2018-04-30
Move all developer guides to guides/developers/

benjaminwil:
Now that we are expanding the Solidus guides to include guides that are not aimed at a only at a developer audience, we should make a clear distinction of audience when we create new articles. 

Developer documentation should go in a `developers/` subdirectory. This PR simply moves everything that's currently in the `guides/` directory to the `guides/developers/` directory.


tvdeyen (commented):
If we begin adding more guide categories we should think about moving them into their own repository. Especially if we start to add assets, screenshots, maybe videos.

tvdeyen:
@benjaminwil this needs a rebase.

tvdeyen (approved):
Looks good 👍

benjaminwil:
This pull request is no longer necessary because #2740 was merged.

PR #2728, opened 2018-05-03
Fix fieldset legend position in Firefox

tvdeyen:
Fieldset legends are a weird html tag as they have magic layout attrobutes that are different in every browser. This unifies them with less css as possible.

### Before

screenshot-2018-5-3 default - stock locations - shipping - settings

### After

screenshot-2018-5-3 default - stock locations - shipping - settings 1


tvdeyen:
There is unfortunately much more involved than I thought. Closing for now

PR #2734, opened 2018-05-14
Require sass >= 3.5.2

gmacdougall:
Bootstrap 4.x requires at least sass 3.5.2.

For more information see:
https://github.com/twbs/bootstrap-rubygem/pull/122
https://github.com/sass/sass/issues/2383


jhawthorn:
> In order to support SassC (a much faster, C based Sass implementation) we should actually remove the sass and sass-rails dependencies from our gemspecs and tell our users to add the Sass implementation of their liking.

I think this still works with both included. `sassc-rails` even [depends on sass](https://github.com/sass/sassc-rails/blob/master/sassc-rails.gemspec#L26-L28)

PR #2737, opened 2018-05-16
Remove promotions admin UI that misleadingly doesn't do anything on the solidus_frontend

benjaminwil:
While the UI I am removing are candidates for useful features, these features are not actually implemented or working on the `solidus_frontend`. I also do not think that the UI really "helps" developers who may want to implement these features or features like them.

I think that we should remove these UI elements for now because they mislead store administrators. When we re-add them to the admin UI, we should make sure that developers have a clear path forward for implementing the features.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Solidus 2.7.0 (master, unreleased)
 
+### Admin
+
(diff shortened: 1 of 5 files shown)


tvdeyen (approved):
I absolutely agree. Thanks! A changelog entry would be nice, though.

benjaminwil:
I am under the impression that the changelog was just generated from commit messages, except for the major changes block for each version.

benjaminwil:
Looks like I also am making some tests fail. I will fix this shortly.

gmacdougall (approved):

PR #2739, opened 2018-05-18
Add docs site shell

jgayfer:
This PR contains the shell of a static site built to house the Solidus developers documentation.

A working version of the site can be viewed at: http://jgayfer.s3-website.us-east-1.amazonaws.com/

This site is modelled after the new [Solidus marketing site](https://solidus.io) to ensure a consistent user experience across both sites.

This site could also be used for future Solidus docs sites (API and Admin) to further the goal of a unified Solidus experience on the web.

### Rationale

Keeping the source for the docs site within the Solidus repository has several advantages:

- It allows the docs to be versioned alongside the code. When major changes are introduced in Solidus, it is clear where and how the docs should also be changed.
- Having the docs stay within the Solidus repository ensures that they will be evaluated (by the community and the core developer team) at the same high standard that Solidus code contributions are.
- There's no risk in adding the docs site to the main Solidus repository. The docs are self-contained and don't affect the code base. And, for the forseeable future, they don't significantly increase the size of `git clone`s of Solidus.

We can decide to move the docs to another repository in the future. But first, we would need to create a plan to make sure the docs aren't treated as a second-class citizen.

### Next steps

Note that this PR is a shell for the docs site:

- In a [later PR](#2740), we would need to move the docs to `docs_site/source/developers` directory. 
- We would need to change each article extension from `.md` to `.html.md`, as this is the format Middleman requires. 
- We would also need to go through each article and change each relative link from `.md` to `.html`.
- Regarding deployment, this site is setup so that it can easily be deployed to an AWS S3 bucket using the `middleman-s3_sync` gem.


docs_site/source/topic.html.erb
```diff
@@ -0,0 +1,111 @@
+---
+title: Home
```
kennyadsl: is this file needed?

jgayfer: @kennyadsl No, it's not. It was just a template/example of the styles we are using. I'll remove it!

PR #2745, opened 2018-05-25
Add initial user management documentation for end users

benjaminwil:
This adds initial documentation about Solidus's user management for end users. The audience for this documentation is store administrators or anyone who has to touch the `solidus_backend` as a business person.


guides/users/users/overview.md
```diff
@@ -0,0 +1,78 @@
+# Overview
+
+Solidus includes a straightforward user management system. Users are your stores
```
tvdeyen: I think we can remove "straightforward" here or replace it with "sophisticated". But leaving it out is most likely the best option.

guides/users/users/overview.md
```diff
@@ -0,0 +1,78 @@
...
+customers, as well as your store administrators who need access to the Solidus
+admin interface.
+
+You can search and sort users using a few types of information:
+
+- **Email**: Users access your store and make orders using an identifying email
```
tvdeyen: Users sign into your store is more accurate as they can access the store without leaving an email

guides/users/users/overview.md
```diff
@@ -0,0 +1,78 @@
...
+- **Member Since**: The date that the user's account was created.
+
+## User information 
+
+Solidus stores additional information about users. This information relates the
+users to other aspects of your store like payments and orders.
+
+The following information is tracked against your users:
+
+- **Addresses**: Customers have at least one address on file. This address can
```
tvdeyen: Maybe explain that every time a user enters a new address we associate the address with the users account.

guides/users/users/overview.md
```diff
@@ -0,0 +1,78 @@
...
+
+You may want to create additional roles like `customer_support` that allow users
+to only access customer orders, shipments, and returns.
+
+By default, users with no roles can only access pages on the storefront. If
+users without a role try to access the Solidus admin, they are greeted with an
+access denied page.
+
+Talk to your developers about adding additional roles and the requirements that
+you have for those roles.
```
tvdeyen: 👌 maybe explain why it is needed to involve a developer.

tvdeyen (commented):
A good first version. I left some comments

guides/users/users/overview.md
benjaminwil: Thanks for drawing attention to this, I will tweak this sentence.

But I think that this is expected behavior and doesn't need to be explained.

If a customer enters a new address in their second order, then the new address "updates" the address on file that is exposed in the backend UI.

And at the moment, the backend UI doesn't expose the list of old or alternative addresses.

tvdeyen (approved):

kennyadsl (approved):
Thanks!

PR #2748, opened 2018-05-25
Allow `register_ability` to accept strings

gmacdougall:
This deals with Rails autoloading better and allows classes to be specified
as strings instead of class names.


gmacdougall:
This obviously does not work. Will take another pass at it.

PR #2751, opened 2018-05-29
Dynamically render ReportsController translations

stewart:
This PR corrects an issue that occurs when attempting to extend `Spree::Admin::ReportsController` with additional reports. The `ReportsController.add_available_report!` class method currently translates its arguments with `I18n.t` when called, rather than when the `/admin/reports` page is actually rendered.

If additional reports are being decorated onto this controller via `class_eval` or prepended modules, there is a possibility that the `i18n` library has not yet been initialized – this results in static "translation missing" errors in the admin. In particular, this is the case with `config.to_prepare`, which is commonly used to load these kinds of Solidus modifications.

To correct this issue, we can switch to dynamically rendering the translations in the view, rather than generating them statically in the controller at startup.

### DIFF
--- a/backend/spec/controllers/spree/admin/reports_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/reports_controller_spec.rb
@@ -9,24 +9,16 @@ describe Spree::Admin::ReportsController, type: :controller do
     it 'should contain sales_total' do
       expect(Spree::Admin::ReportsController.available_reports.keys.include?(:sales_total)).to be true
     end
-
-    it 'should have the proper sales total report description' do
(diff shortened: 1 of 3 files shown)


backend/spec/controllers/spree/admin/reports_controller_spec.rb
```diff
@@ -9,24 +9,16 @@
     it 'should contain sales_total' do
       expect(Spree::Admin::ReportsController.available_reports.keys.include?(:sales_total)).to be true
     end
-
-    it 'should have the proper sales total report description' do
-      expect(Spree::Admin::ReportsController.available_reports[:sales_total][:description]).to eql('Sales Total For All Orders')
-    end
```
stewart: It may be worth replacing this test with a view spec, but I'm uncertain how much value testing translations provides in this context.

PR #2754, opened 2018-05-30
Add minor change for API taxons controller

jtapia:


api/app/controllers/spree/api/taxons_controller.rb
```diff
@@ -4,21 +4,10 @@ module Spree
...
-        if taxonomy
-          @taxons = taxonomy.root.children
-        elsif params[:ids]
-          @taxons = Spree::Taxon.accessible_by(current_ability, :read).where(id: params[:ids].split(','))
-        else
-          @taxons = Spree::Taxon.accessible_by(current_ability, :read).order(:taxonomy_id, :lft).ransack(params[:q]).result
-        end
-
-        @taxons = paginate(@taxons)
+        @taxons = paginate(taxons)
```
kennyadsl: It is a bit twisted for me that `@taxons` will contains the paginated version of `taxon` which sets an ivar named `@taxons` which contains the taxon scope. 

What about naming the memoizing method `taxons_scope` or `all_taxons` or something similar?

jtapia:
I think this would make the code more readable

jarednorman:
I'm generally not in favour of extracting methods like this without a good reason. The controller action was simple enough to begin with (not even ten actual lines of code!) and now to see how it works I have to look a hundred lines down in the file.

If someone needs to override the functionality in the extracted method or something, then the argument could be made, but without that I don't think we should merge this.

kennyadsl:
Closing this one, there not seems to be interest, thanks for the contribution anyway.

PR #2756, opened 2018-05-30
Add promotion rules documentation for end users

benjaminwil:
This adds documentation about Solidus's promotion rules for end users. The audience for this documentation is store administrators or anyone who has to touch the `solidus_backend` as a business person.


guides/source/users/promotions/promotion-rules.html.md
```diff
@@ -1 +1,62 @@
 # Promotion rules
+
+Promotion rules determine whether a promotion can be applied to an order.
+Promotions are only *eligible* if it meets all of the rules that have been
```
gmacdougall: Below, it is stated that this can be configured to match all or match any.

gmacdougall (changes requested):

benjaminwil:
A promotion rules article has already been merged.

PR #2765, opened 2018-06-04
Remove x-overflow rule from admin nav container

jacquesporveau:
This CSS rule is not needed.

There is JS to handle overflowing tabs inside the Solidus Project that
currently resides in:

https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/components/tabs.js

Furthermore this CSS can cause issues for views. When the `.content-main`
div does not have enough content to be tall enough for the dropdown tab to
fully display a scroll bar is introduced on the right hand side of the page
and looks silly.

## Before
bad

## After
good

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
+++ b/backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
@@ -38,7 +38,6 @@ body {
 
 .content-main {
   @include make-col(12);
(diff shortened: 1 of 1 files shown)


jacquesporveau:
Eh I had an old version of Solidus and the PR will look different than this. Closing for now and will open another PR when I have the right solution implemented.

Issue #2519, opened 2018-01-20
Country States Not Populating - Sandbox


grichardomi:
### Steps to reproduce
New sandbox installation

### Expected behavior
States to change based on "Country" selection

### Actual behavior
States are not changing for selected countries

### System configuration
**Solidus Version**: Latest

**Extensions in use**:
None

kennyadsl:
Do you mean in frontend checkout, address step?

grichardomi:
Alberto,

Thanks for responding. I did not check the front-end. It is the back-end,
that I'm most concerned about.

Guy



<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail&utm_term=icon>
Virus-free.
www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail&utm_term=link>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

jhawthorn:
I can't reproduce this, it is working here. Can you provide more details?

![](http://i.hawth.ca/s/pD7HepUB.png)

grichardomi:
Thanks John for the feedback.

Admin -> user -> address was not working in my solidus app. So I cloned and
ran sandbox version over two weeks ago. States are not working for me.

There could something abnormal, not necessarily the sandbox but in my
server environment. I'm using ubuntu instance. Is this possible?

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail&utm_term=icon>
Virus-free.
www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail&utm_term=link>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

jacobherrington:
I believe I have reproduced what @grichardomi is describing.

![screenshot from 2018-04-13 22-35-23](https://user-images.githubusercontent.com/11466782/38763900-01728b4e-3f6b-11e8-8322-4251f6eae801.png)

I am on Fedora 26 and followed the readme to build a sandbox application with Solidus. The screenshot was taken at `http://localhost:3000/admin/users/1/addresses`.

@jhawthorn I am new to Solidus, but happy to try and figure this one out.

jacobherrington:
[...]

PR #2766, opened 2018-06-06
Fix bug with user address forms

jacobeubanks:
refs #2519 

Bug: states would not reload with change in country on user address forms

Stock Location JS solved this for stock location forms so it was renamed and used in user address forms.

Notes: Considering this change a refactor so no new tests were added. Let me know if tests should be written.


jacobherrington:
Hey @tvdeyen, thanks for the quick review. 

The issue occurs in the backend when updating a user's address information. When the country is changed, the states collection does not update. See #2519 for a screenshot of the issue. 

You can see the behavior in the sandbox app or any vanilla solidus app at `http://localhost:3000/admin/users/1/addresses`.

Existing code in [`stock_locations.js`](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/stock_location.js) (now `locations.js`) handled the problem by targeting a CSS class. We added that class to the user's address form, as opposed to rewriting the same code.

We changed the name of the JS file and the CSS class name to reflect an increase in usage. It is now being used in [`admin/users/_addresses_form.html.erb`](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/users/_addresses_form.html.erb) as well as [`admin/stock_locations/_form.html.erb`](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/stock_locations/_form.html.erb). 

Let me know if there are any other changes needed.

jacobeubanks:
@tvdeyen Agreed! Change made.

Thanks for the speedy responses!

backend/app/assets/javascripts/spree/backend/addresses.js
```diff
@@ -3,7 +3,7 @@
 Spree.ready(function() {
   "use strict";
 
-  _.each(document.querySelectorAll('.js-stock-locations-form'), function(el) {
+  _.each(document.querySelectorAll('.js-locations-form'), function(el) {
```
tvdeyen: Could we also rename the class to `js-addresses-form`?

jacobeubanks: @tvdeyen Good idea! Change made. Thanks!

jacobherrington:
@tvdeyen @kennyadsl just curious, is there anything else we need to do on this PR?

kennyadsl:
No I think we just need to merge, doing it now!

jacobeubanks:
@kennyadsl Thank you!

Issue #2758, opened 2018-06-01
Deprecate _skeleton.scss in favor Bootstrap in the frontend


jacobherrington:
The erb templates in frontend are currently reliant on Skeleton V1.2 ([stored locally]( https://github.com/solidusio/solidus/blob/master/frontend/app/assets/stylesheets/spree/frontend/_skeleton.scss)), which uses fixed width columns.

I'm working on a project that will deprecate Skeleton in favor of Bootstrap for a few these templates. I'm happy to change all of the templates and PR back to frontend if there aren't any objections.

jacobherrington:
Looking at #580, I'm not sure if this would be a welcome change. Open to feedback.

At the least, I would prefer to replace [_skeleton.css](https://github.com/solidusio/solidus/blob/master/frontend/app/assets/stylesheets/spree/frontend/_skeleton.scss) with a grid system that isn't fixed width.

tvdeyen:
@jacobherrington I say give it a shoot.

jacobherrington:
Based on my own _anecdotal_ experience, I think Solidus frontend is very important for newcomers to Solidus and potential adopters. I also think the current frontend is a hurdle for newcomers because it is a bit dated. 

I would love to build a new frontend with a powerful framework like React or Vue, but In the interest of keeping our footprint small...

**Short-term**
We will replace [_skeleton.css](https://github.com/solidusio/solidus/blob/master/frontend/app/assets/stylesheets/spree/frontend/_skeleton.scss) with an extremely light, more-modern CSS grid system. I feel like that is a good incremental quality of life change that will help new Solidus developers in building their own shop frontend.

**Long-term**
In light of our previous conversation @tvdeyen, I will look in to building a modern starter frontend for Soldius that will function as a drop-in replacement for the current frontend.

tvdeyen:
Probably also an option is to use no framework at all. Flex box and CSS grids have broad support and are a good starter for a modern store frontend.

kennyadsl:
Hey there, sorry for being late in answering and thanks for your thoughts and work on this topic. 

[...]

PR #2767, opened 2018-06-08
Replace skeleton grid with bootstrap grid

jacobherrington:
Refs #2758 

**The issue**
Solidus frontend currently relies on [_skeleton.scss](https://github.com/solidusio/solidus/blob/master/frontend/app/assets/stylesheets/spree/frontend/_skeleton.scss), which is a grid system using fixed widths. It is  outdated (last updated in 2011). It might confuse or deter newcomers because it is not mobile friendly and difficult to iterate on.

**What this PR changes**
We added _only_ the grid system from Bootstrap 4.1.1 [(bootstrap-grid.css)](https://github.com/jacobherrington/solidus/blob/chore/replace-skeleton-css/frontend/app/assets/stylesheets/spree/frontend/bootstrap-grid.css), which should feel familiar to most newcomers and makes it very easy to drop in the full Bootstrap framework if someone wants to extend the existing frontend. On the other hand, it is almost as easy to extract [bootstrap-grid.css](https://github.com/jacobherrington/solidus/blob/chore/replace-skeleton-css/frontend/app/assets/stylesheets/spree/frontend/bootstrap-grid.css) if someone wants to use another solution for styling. This makes the frontend easier to iterate on.

**Caveats**
There are cases in which we could have made slight design changes, but we left those choices out of this PR and may revisit them in the future. Instead, we chose to make as few design changes as possible. The goal of this PR was to move from a fixed-width CSS grid to a flex-based CSS grid, not redesign the look and feel of the frontend.


gmacdougall:
Thanks for your work on this. We've been talking about making major changes to the front end for some time, but looking at this change, I think it can provide a decent quality of development life for people who are using the front end as is right now.

I think the big thing we're concerned about is breaking people's existing customization which they have made using skeleton CSS and markup.

For people just starting their stores now, I do think this is a better options for them.

jacobeubanks:
Thanks for your response!

I agree with @gmacdougall and @jacobherrington. Though I think this is an improvement, I'm not sure if this small change is worth breaking frontends.

tvdeyen:
We should be able to keep the skeleton grid until the next version if Solidus and remove it later. Stores using it (very unlikely they actually do) have the chance to migrate to a grid system of their liking (even keep using skeleton by copy it into their app). 

We should even be able to print out a deprecation warning with Sass if people are still using our skeleton file.

ericsaupe:
I like the idea of deprecating it with an eye to a future release. Solidus 3.0 would have the new styles for example since a full point release usually means things are going to break. Out of the box Solidus should look good and look good on mobile.

jacobherrington:
Going to close this until I can revisit removing Skeleton in a better way.

PR #2768, opened 2018-06-08
Add support for multiple tables of contents in the Solidus Guides

benjaminwil:
Right now, the Solidus Guides include one guide: the Solidus Developers Guide.

In the future, we intend to publish other guides with separate tables of contents. For example, a guide for end users (which is partially written and available at `/guides/source/users/`) should be displayed separately from the Developers Guide, as the content is written for a non-developer audience.

These commits prepare the Solidus Guides to handle those additional TOCs.


benjaminwil:
I’m going to change a few things here and then reopen this.

Issue #2741, opened 2018-05-22
Negative order totals


kushniryb:
### Steps to reproduce
Create 2 separate promotions with the following configurations:
1. Whole-order adjustment with the Flat Rate of 500 
2. Per-line-item adjustment with the Flat Percent of 100 

Buy some items so that order total is less than 500, go to the cart and apply them one after another.

### Expected behavior
• Order total is 0

### Actual behavior
• Order total is negative

### System configuration
**Solidus Version**: `master`

**Extensions in use**: Default `sandbox`

jacobherrington:
I am able to replicate this:
![screenshot from 2018-06-11 11-49-41](https://user-images.githubusercontent.com/11466782/41245417-9954b5b0-6d6d-11e8-8e00-48384500d26d.png)

kennyadsl:
We discussed this issue in the core team meeting, and we think that this scenario is basically not realistic. We think that no business could ever create this kind of cumulative promotions, isn't it? 

Despite we know that technically this could happen, in this case it's more like an issue in how you set promotions up for your store and we think that for now the best trade-off between patching complex code, working around the issue as for (#2769) and do something here is to just raise an exception when calculating total, if it's less than 0. Stores that want to ignore this scenario can just ignore it while other store can rescue the exception and print an error message to the user.

@jacobherrington again, thanks for your contribution, we really appreciate the effort, if you'd like to take this task it would be great!

jacobherrington:
I still plan on working on this when I have time, but I'm also happy to pair-with/hand-off-to someone looking to get involved in this project.

peterberkenbosch:
Work is being done here right now. https://github.com/solidusio/solidus/pull/2855

Issue #2593, opened 2018-02-21
RFC: Higher level interface for applying promotions


jhawthorn:
Our current promotion interface often doesn't allow the desired behaviour.

We want all promotions to calculate a promotion amount at a line-item level, but often stores want to emulate a per-order discount. This results in calculators like [DistributedAmount](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/calculator/distributed_amount.rb) which attempt to do this, but are awkward and involve error prone recalculations involving the whole order per-line item.

I believe we should take the lessons learned addressing similar issues in taxation #1892, and try to apply them to promotions.

PR #2769, opened 2018-06-11
fix negative total displaying in cart by displaying zero if total is …

jacobherrington:
…less than zero

Refs #2741 

If an admin set the promotions in a specific way, it was possible to make the cart display a negative value before checkout. The bug is purely visual as Solidus sets the checkout value to $0.00. This PR simply makes the frontend reflect that behavior in the cart.

Feedback is welcome.

### DIFF
--- a/core/app/helpers/spree/base_helper.rb
+++ b/core/app/helpers/spree/base_helper.rb
@@ -12,7 +12,7 @@ module Spree
         text = "#{text}: (#{t('spree.empty')})"
         css_class = 'empty'
       else
-        text = "#{text}: (#{current_order.item_count})  <span class='amount'>#{protect_from_negative(current_order)}</span>"
+        text = "#{text}: (#{current_order.item_count})  <span class='amount'>#{display_order_total(current_order)}</span>"
         css_class = 'full'
       end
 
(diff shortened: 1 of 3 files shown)


jacobherrington:
Don't merge this, just realized there is still some work to be done for this bug.

jacobherrington:
Renamed the method - feel free to give some feedback! Thanks!

kennyadsl:
I think the best way to solve this issue is to do not allow promotion to apply (or apply it partially) when the order total reach a negative amount, but at the moment I don't know if this is easy to achieve with current promotion implementation. I'm not sure about what to do here. I'll bring this topic into the next core team meeting and will let you know!

jacobherrington:
@kennyadsl sounds good. 

However, it is worth mentioning, I think this is 99% presentation as the checkout still reflects a total of $0.00.

kennyadsl:
Yes, checkout could work, but the issue is at data level if I got it correctly. We should not allow that since an order with negative amount makes no sense.

jacobherrington:
@kennyadsl I agree, with the caveat that I don't think we should make it difficult to set a product to zero as some store owners might intentionally give a product away.

Is #2593 relevant to this issue? 

Could you kill two birds with one stone by handling this issue at the promotions instead of the orders level? Just a thought.

Issue #2008, opened 2017-06-08
Missing user email not evaluated until confirm


brchristian:
When I create a brand new Solidus 2.2 application and try to place an order from the backend, I notice that if I fill out the customer details with no email address, there is no validation error. Then later, when I go to Confirm >> Refresh Totals, I get an error message saying `Customer email can’t be blank`.

<img width="864" alt="screen shot 2017-06-08 at 4 20 07 pm" src="https://user-images.githubusercontent.com/2460418/26949084-99619f7e-4c66-11e7-980a-33a08599cd96.png">

It seems to me that we should run this validation on the _customer_ page, along with the other customer details validations (e.g., that state and country are present, etc.).

jacobherrington:
This was solved by #2771.

brchristian:
Closing issue. Thanks, @jacobherrington!

PR #2771, opened 2018-06-12
make customer email field required when an admin is making a new order

jacobherrington:
This adds simple HTML validation for the email field. 

I personally think that is enough to solve the problem.

refs #2008

### DIFF
--- a/backend/app/views/spree/admin/orders/customer_details/_form.html.erb
+++ b/backend/app/views/spree/admin/orders/customer_details/_form.html.erb
@@ -7,7 +7,7 @@
       <div class="col-9">
         <div class="field">
           <%= f.label :email %>
-          <%= f.email_field :email, class: 'fullwidth' %>
+          <%= f.email_field :email, required: true, class: 'fullwidth' %>
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


BenMorganIO:
I think you should also add a validation in the model as well. If the order is a guest order, then it should require the customer email.

jacobherrington:
Hey @BenMorganIO! Thanks for weighing in.

Inside the Order model line 127 is `validates :email, presence: true, if: :require_email` which makes sure that you can't have a guest checkout without an email. 

The issue is that the frontend doesn't stop you from progressing though several long screens before the validation fails, causing the user to lose significant time if they have to remake the order. This change will just keep the user from progressing to the next screen without inputting an email - ultimately saving time for an admin user.

BenMorganIO:
Ah, I see. :+1: good sir.

jacobherrington:
@tvdeyen anything else I need to do on this PR. Seems pretty trivial to me, but happy to do some fixups if needed.

tvdeyen:
@jacobherrington no, all fine. We need a second review of a core team, then we are ready to go.

Issue #546, opened 2015-11-26
non-admin api users can update order after it is complete


carchrae:
we've noticed that a non-admin user can update an order through the api, even if it is in the complete state.  eg, using the order_token and making line_item adjustments.

i can't see why this should be allowed.  it means a user can change line items, etc.  (due to a bug in our UI code, this happened)

i locked down the order by adding the following can can rule:

`app/models/spree/ability_decorator.rb`

``` ruby
# See http://github.com/ryanb/cancan for more details on cancan.
require 'cancan'
class AbilityDecorator
  include CanCan::Ability

  def initialize(user)
    user ||= Spree.user_class.new

    if user.respond_to?(:has_spree_role?) && user.has_spree_role?('admin')
      # if you are an admin, we don't apply the rule in the else block
    else
      # not allowed to update completed orders - wtf, why isn't this a default!
      cannot [:update], Spree::Order do |order|
        order.completed?
      end
    end

  end
end

Spree::Ability.register_ability(AbilityDecorator)
```

now, clearly the UI code should note that the user's order is complete and should not allow them to update the order (yes, we had that bug, yes, we fixed the UI), but client side code should not check rules like this, so we added it to the server as well.

i'll file a pull request to update the can can rules, but i have to ask, was this a design choice to allow a user to change a completed order?  or is this simply a case that has been handled by controller logic before?

seand7565:
Looks like this was fixed with #2787 - can anyone verify?

kennyadsl:
@seand7565 yep, thanks for pointing out!

PR #2772, opened 2018-06-14
Fix the ability to update cart content on completed order

sebfie:
Hello,

We just got an issue with our solidus api : 
- We have a react frontend and we use solidus as an api
- We can update order content event if it's paid or completed

Step to reproduce : 
- Create an order, pay it
- Create a curl request on api to update the order content
- It responses with a 200

At the moment, I just created a rspec test to show the issue, in fact I need some advise to know how to fix it : 
- Should I use cancancan?
- Should I check order status on a before filter on lineItemsController?

To make a fast fix on my website, I did : 

```
module Spree
  module Api
    module LineItemsControllerDecorator
      def check_order_state
        unauthorized if ["complete", "canceled", "confirm"].include?(@order.state) # Maybe check if it's a final state?
      end
    end
  end
end

Spree::Api::LineItemsController.prepend Spree::Api::LineItemsControllerDecorator
Spree::Api::LineItemsController.before_action :check_order_state, only: [:create, :update]
```

Thank you !

### DIFF
--- a/api/spec/requests/spree/api/line_items_controller_spec.rb
+++ b/api/spec/requests/spree/api/line_items_controller_spec.rb
@@ -154,6 +154,17 @@ module Spree
             post spree.api_order_line_items_path(order), params: { line_item: { variant_id: product.master.to_param, quantity: 1 } }
             expect(order.reload.shipments).not_to be_empty
           end
+
+          it "cannot update a line item on order" do
+            line_item = order.line_items.first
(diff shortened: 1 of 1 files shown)


kennyadsl:
Hey, this is a duplicate of #546. I also think this should not be the default behavior in core. I'm preparing a fix but I have a lot of specs failing so I need to understand better if there is some side effect of this or some specs are just not well written.

sebfie:
Ok, thank you

kennyadsl:
Fixed in #2787, thanks anyway!

PR #2775, opened 2018-06-15
Add initial settings documentation for end users

benjaminwil:
This adds documentation about the Solidus settings interface for end users. The audience for this documentation is store administrators or anyone who has to touch the `solidus_backend` as a business person.

It also adds to the English locales for the the interface, which is a minor UX improvement. The `code` field was very ambiguous.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Support this project by becoming a Key Stakeholder. Your logo will show up here 
 
 <a href="https://nebulab.it/"><img src="https://images.opencollective.com/proxy/images?src=https%3A%2F%2Fopencollective-production.s3-us-west-1.amazonaws.com%2F3cc3c170-20cc-11e9-9582-214168b65c9a.png&height=100"></a>
 <a href="https://www.enginecommerce.com/"><img src="https://images.opencollective.com/proxy/images?src=https%3A%2F%2Flogo.clearbit.com%2Fenginecommerce.com&height=100"></a>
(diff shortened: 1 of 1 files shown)


spaghetticode:
@benjaminwil can you rebase this with master?

Also, I think that while sorting translation keys is a good idea, it also makes harder to see actual changes with this PR.

core/config/locales/en.yml
```diff
@@ -1298,8 +1300,9 @@ en:
         fulfillable: "When unchecked, this indicates that items in this location don't require actual fulfilment. Stock will not be checked when shipping and emails will not be sent.<br/> Default: Checked"
         check_stock_on_transfer: "When checked, inventory levels will be checked when performing stock transfers.<br/> Default: Checked"
       spree/store:
-        cart_tax_country_iso: "This determines which country is used for taxes on carts (orders which don't yet have an address).<br/> Default: None."
         available_locales: "This determines which locales are available for your customers to choose from in the storefront."
+        cart_tax_country_iso: "This determines which country is used for taxes on carts (orders which don't yet have an address).<br/> Default: None."
+        code: "An identifier for your store. Developers may need this value if you operate multiple storefronts.
```
aitbw: A quote is missing at the end of this line.

benjaminwil:
As you can see, when i rebased this with `master`, it seems that all of this has already been merged via another PR. 💯

PR #2779, opened 2018-06-20
Refund shipping costs

### LINKED ISSUE #2628: Refunds on shipping cost
Hi,
We struggle to understand the best way to  refund shipping costs.

At the moment the best way we found is to do it directly from the payment tab, by entering manually the amount of the shipping cost and adding a reason like “refund shipping cost”.  I would like to see  if there is a smoother way for admins to do this.
Maybe having the line shipping cost in the “customer returns” tab like it is done for products. 
screen shot on 2018-03-14 at 15_29_14

Do you think this way would be possible ? and/or a good thing to add to solidus ? I could probably work on a PR for that.
As noted by @gmacdougall, it doesn't really fit in to the customer return model, since they can't return the shipment...

I'd like to have opinions on that before starting anything.
Thanks.

fkoessler:
This PR allows to refund shipping costs and fixes #2628.

A new Settlement model has been added with an association to the Reimbursement and Shipment models.
Settlements and ReturnItems share similarities, they are both "reimbursement items" so they share a common interface and must both implement the #total method.

To create a settlement, you must first create a customer refund and a reimbursement.
From the reimbursement page, you can then add a settlement for a shipment and update the reimbursement.
When adding a settlement, eligibility validators are run against the settlement to decide its state: accepted, rejected or manual_intervention_required. This behaviour is similar to return items.

screenshot-localhost 3000-2018-06-20-14-06-15

The settlement is taken into account when creating refunds and/or store credits.

The interface also provides a button to reject settlements in case we don't want them any more.

Let me know if you have any question / feedback.
I hope this will be useful.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/returns/return_item_selection.js
+++ b/backend/app/assets/javascripts/spree/backend/returns/return_item_selection.js
@@ -16,7 +16,8 @@ Spree.ready(function() {
   }
 
   var formFields = $("[data-hook='admin_customer_return_form_fields'], \
-                     [data-hook='admin_return_authorization_form_fields']");
+                     [data-hook='admin_return_authorization_form_fields'], \
+                     [data-hook='admin_reimbursement_settlements_form_fields'] ");
 
   if(formFields.length > 0) {
     updateSuggestedAmount();
--- a/backend/app/controllers/spree/admin/reimbursements_controller.rb
+++ b/backend/app/controllers/spree/admin/reimbursements_controller.rb
@@ -40,7 +40,7 @@ module Spree
       end
 
       # We don't currently have a real Reimbursement "new" page. And the only
-      # built-in way to create reimburesments via Solidus admin is from the
+      # built-in way to create reimbursements via Solidus admin is from the
       # customer returns admin page via a button that supplies the
       # "build_from_customer_return" parameter. The "edit" page is not
       # suitable for use here for that reason as well.
--- a/backend/app/controllers/spree/admin/reimbursements_controller.rb
+++ b/backend/app/controllers/spree/admin/reimbursements_controller.rb
@@ -10,6 +10,7 @@ module Spree
       before_action :load_stock_locations, only: :edit
       before_action :load_simulated_refunds, only: :edit
       before_action :load_settlements, only: :edit
+      after_action :attempt_accept_new_settlements, only: :update
 
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
@@ -69,6 +70,12 @@ module Spree
         end
 
         @existing_settlements = @reimbursement.settlements.for_shipment.not_pending
+      end
+
+      def attempt_accept_new_settlements
+        @reimbursement.settlements.pending.each do |settlement|
+          settlement.attempt_accept!
+        end
       end
 
       def load_simulated_refunds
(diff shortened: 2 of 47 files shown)


tvdeyen:
> So, can we merge it? We need it in our project :)

@sebfie PRs need two approvals from core team members. This PR is a larger feature that will need manual review besides code review. Since all core team members doing this on a voluntary basis and mostly in free time, please be patient. 

Meanwhile you can run a fork of solidus that includes this PR in your app. 

Thanks for your understanding

sebfie:
Alright, I will wait 👍

dhonig:
@tvdeyen @kennyadsl   Just bringing this one up for attention since it came up in support this morning.  My own thoughts are that this is really good work but it breaks new ground in a direction that we need to be careful of.   What I mean is that the same problem can be solved purely through adjustments.   When a new model is introduced we have to weigh the benefit of the new model vs increase in complexity it brings.  In my own experience with large stores, there are anomalies that happen with adjustments and incomplete orders that can be very hard to track down. This design could help in those cases, but the last thing I want to see are more models chucked into core without very careful scrutiny and a vision for how they the a coherent future direction.   That vision could be simply expressed as favor adjustments for these concerns and we'll use implicit models for these other concerns.

peterberkenbosch:
@dhonig my experience is that adjustments can clutter the reporting and exact intent of said adjustment when it's used for refunds. Agree 100% on the remark about new terrain. This needs some proper thought and discussion. I do like the first look of it though!

skukx:
This is definitely something `Deseret Book` would much appreciate. From a quick glance everything looks good to me. I'll take a closer look later though.  My only request would be to add more documentation above each method and class.

fkoessler:
Thanks a lot for taking the time to review this PR.

My experience is that adjustments are not adapted for refunds because:
- a negative adjustment will recompute the order's total amount and change its status to 'credit owed', which does not reflect the reality
- after the adjustment is created, we still have to create a reimbursement of the same amount as the adjustment
- it makes it difficult to analyse what has happened, we need to look at adjustment labels to identify refunds. This is much easier done if refunds are stored in a settlement table

Shipping cost refunds is very common in Europe in our line of business.

Settlements could also allow for more than just refunding shipping costs, it's actually a model that can be used to store any kind of refund / compensation.
For example, it could be used to store a commercial gesture. In this case, the settlement would not be associated with a shipment. To allow for such feature, the model can be kept as is, it's just a matter of adding the UI to allow for this action.
I would not say this is a minor feature.

That being said, I understand the reticence to introduce another model with its database table that will need to be maintained.
Let us know your decision about what should be done: document / improve & merge or create a solidus extension.

And thanks again for the great work on solidus :)

ericsaupe:
The way we get around this at Deseret Book is we issue the Refund for however much we are giving back to the customer, might just be shipping. That results in a `balance_due` in which case we create an Adjustment with the negative of the amount we refunded to bring the order back to balance.

For example:
- User calls in wanting shipping refunded
- Admin clicks refund on payment, enters amount of `$3.50`
- Order now shows balance due of `$3.50`
- Create new Adjustment for `-$3.50` with a reason of Shipping (I think we added that reason, can't remember if it is default in Solidus)
- Order now shows everything is balanced

It sort of follows double entry accounting in that we have both the debit of money in the Refund and the credit in the Adjustment.

tvdeyen:
Reviewing this again we decided in the core meeting today that we do not want have this in core. We think this is lot of code for a very specific need one could address in an extension or their custom stores. Thanks for the contribution.

PR #2781, opened 2018-06-21
Update promotion-actions.html.md

bazfer:
Action must be namespaced.

### DIFF
--- a/guides/source/developers/promotions/promotion-actions.html.md
+++ b/guides/source/developers/promotions/promotion-actions.html.md
@@ -82,7 +82,7 @@ You must then register the custom action in an initializer in your
 `config/initializers/` directory:
 
 ```ruby
-Rails.application.config.spree.promotions.actions << MyPromotionAction
+Rails.application.config.spree.promotions.actions << Spree::Promotion::Actions::MyPromotionAction
 ```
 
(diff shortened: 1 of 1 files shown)


guides/source/developers/promotions/promotion-actions.html.md
```diff
@@ -82,7 +82,7 @@ You must then register the custom action in an initializer in your
 `config/initializers/` directory:
 
 ```ruby
-Rails.application.config.spree.promotions.actions << MyPromotionAction
+Rails.application.config.spree.promotions.actions << Spree::Promotion::Actions::MyPromotionAction
```
gmacdougall: I don't believe this is a pattern we wish to encourage. If this has resolution issues as it is I think that:

```Ruby
Rails.application.config.spree.promotions.actions << ::MyPromotionAction
```

or 


```Ruby
Rails.application.config.spree.promotions.actions << MyNamespace::MyPromotionAction
```

would be preferrable.

jacobherrington:
Hey @bazfer, are you planning on making this change?

guides/source/developers/promotions/promotion-actions.html.md
bazfer: What I suggested is what worked for me, but if either one of these two alternatives resolve as well and the pattern is preferred, sure, why not.

PR #2782, opened 2018-06-21
Allow to easily extend the TaxonsController

coorasse:


frontend/app/controllers/spree/taxons_controller.rb
```diff
@@ -4,19 +4,22 @@ module Spree
   class TaxonsController < Spree::StoreController
     helper 'spree/products', 'spree/taxon_filters'
 
+    before_action :load_taxon, only: [:show]
+
     respond_to :html
 
     def show
-      @taxon = Spree::Taxon.find_by!(permalink: params[:id])
-      return unless @taxon
```
kennyadsl: This line was useless, right?

coorasse: yes. exactly. I also double checked to be sure but the `find_by!` raises an exception if the element is not found, returning 404.

coorasse:
Good catch. Yes, that would happen. 
Do you want me to add a note to the Changelog? Another option would be to call `@taxon = load_taxon` directly inside the action...even though I would prefer the current version.

kennyadsl:
@coorasse let's go with just adding a line to the Changelog, thanks!

coorasse:
Good. I updated the CHANGELOG and squashed. 👍

PR #2783, opened 2018-06-23
Stock location sorters

aldesantis:
This introduces the concept of a stock location sorter, which takes a collection of stock locations and returns them in the right order for inventory allocation.

It also updates `SimpleCoordinator#allocate_inventory` to respect the order of stock locations returned by the sorter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Solidus 2.8.0 (master, unreleased)
 
+### Core
+
+- Implement stock location sorters [#2783](https://github.com/solidusio/solidus/pull/2783) ([aldesantis](https://github.com/aldesantis))
+
 ## Solidus 2.7.0 (2018-09-14)
 
 ### Major Changes
--- a/core/app/models/spree/stock/simple_coordinator.rb
+++ b/core/app/models/spree/stock/simple_coordinator.rb
@@ -23,7 +23,7 @@ module Spree
         @order = order
         @inventory_units = inventory_units || InventoryUnitBuilder.new(order).units
         @splitters = Spree::Config.environment.stock_splitters
-        @stock_locations = Spree::StockLocation.active
+        @stock_locations = Spree::Config.stock.location_sorter_class.new(Spree::StockLocation.active).sort
 
         @inventory_units_by_variant = @inventory_units.group_by(&:variant)
         @desired = Spree::StockQuantities.new(@inventory_units_by_variant.transform_values(&:count))
@@ -85,7 +85,9 @@ module Spree
       end
 
       def allocate_inventory(availability_by_location)
-        availability_by_location.transform_values do |available|
+        sorted_availability = sort_availability(availability_by_location)
+
+        sorted_availability.transform_values do |available|
           # Find the desired inventory which is available at this location
           packaged = available & @desired
 
@@ -94,6 +96,16 @@ module Spree
 
           packaged
         end
+      end
+
+      def sort_availability(availability)
+        sorted_availability = availability.sort_by do |stock_location_id, _|
+          @stock_locations.find_index do |stock_location|
+            stock_location.id == stock_location_id
+          end
+        end
+
+        Hash[sorted_availability]
       end
 
       def get_units(quantities)
--- a/core/lib/spree/core/stock_configuration.rb
+++ b/core/lib/spree/core/stock_configuration.rb
@@ -5,6 +5,7 @@ module Spree
     class StockConfiguration
       attr_writer :coordinator_class
       attr_writer :estimator_class
+      attr_writer :location_sorter_class
 
       def coordinator_class
         @coordinator_class ||= '::Spree::Stock::SimpleCoordinator'
@@ -14,6 +15,11 @@ module Spree
       def estimator_class
         @estimator_class ||= '::Spree::Stock::Estimator'
         @estimator_class.constantize
+      end
+
+      def location_sorter_class
+        @location_sorter_class ||= '::Spree::Stock::LocationSorter::Unsorted'
+        @location_sorter_class.constantize
       end
     end
   end
--- a/core/spec/lib/spree/core/stock_configuration_spec.rb
+++ b/core/spec/lib/spree/core/stock_configuration_spec.rb
@@ -20,6 +20,7 @@ RSpec.describe Spree::Core::StockConfiguration do
       end
     end
   end
+
   describe '#estimator_class' do
     let(:stock_configuration) { described_class.new }
     subject { stock_configuration.estimator_class }
@@ -34,6 +35,24 @@ RSpec.describe Spree::Core::StockConfiguration do
 
       it "returns the constant" do
         is_expected.to be MyEstimator
+      end
+    end
+  end
+
+  describe '#location_sorter_class' do
+    let(:stock_configuration) { described_class.new }
+    subject { stock_configuration.location_sorter_class }
+
+    it "returns Spree::Stock::LocationSorter::Unsorted" do
+      is_expected.to be ::Spree::Stock::LocationSorter::Unsorted
+    end
+
+    context "with another constant name assiged" do
+      MySorter = Class.new
+      before { stock_configuration.location_sorter_class = MySorter.to_s }
+
+      it "returns the constant" do
+        is_expected.to be MySorter
       end
     end
   end
--- a/guides/source/developers/shipments/stock-location-sorters.html.md
+++ b/guides/source/developers/shipments/stock-location-sorters.html.md
@@ -1,0 +1,52 @@
+# Stock location sorters
+
+This article explains the purpose, interface and correct usage of custom stock location sorters.
+
+Your app's stock location sorter defines the order in which stock locations are used to allocate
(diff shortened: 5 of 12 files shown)


core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -96,6 +102,16 @@ def allocate_inventory(availability_by_location)
         end
       end
 
+      def sort_availability(availability)
```
BenMorganIO: Maybe add some docs on how sorting works.

aldesantis:
Improved this with a better, configurable design.

core/app/models/spree/stock/simple_coordinator.rb
aldesantis: @BenMorganIO you mean on this specific method or in the guides?

core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -23,7 +23,7 @@ def initialize(order, inventory_units = nil)
         @order = order
         @inventory_units = inventory_units || InventoryUnitBuilder.new(order).units
         @splitters = Spree::Config.environment.stock_splitters
-        @stock_locations = Spree::StockLocation.active
+        @stock_locations = Spree::Config.stock_location_sorter_class.new(Spree::StockLocation.active).sort
```
kennyadsl: I'd use another variable or method `@sorted_stock_lcoations` for that. It should also help backward compatibility if someone extended this class and was using the `@stock_locations` ivar. what do you think?

core/app/models/spree/stock/simple_coordinator.rb
kennyadsl: I think that if we document this change well in the PR and in the guides it would be enough

core/app/models/spree/stock/simple_coordinator.rb
aldesantis: @kennyadsl the problem with that is if someone overrides @stock_locations and we sort in a different variable (let's say @sorted_stock_locations), they also have to override @sorted_stock_locations, otherwise it will contain the result of the sorting of the original @stock_locations variable.

kennyadsl: Yes, you are right. One option could be creating another `@active_stock_locations` ivar, use that in the whole class instead of `@stock_locations` which could be deprecated with [DeprecatedInstanceVariableProxy](https://apidock.com/rails/ActiveSupport/Deprecation/DeprecatedInstanceVariableProxy). I feel like it's too complex though, maybe a changelog line is enough here. I'd say, let's see what others think before.

kennyadsl: Talked with @gmacdougall about this. Since the default sorted is returning the same thing as before maybe no one will have unexpected behavior with this, so I'm ok leaving it as is.

aldesantis:
@kennyadsl this was documented in the guides. Please let me know how they look.

guides/source/developers/shipments/stock-location-sorters.html.md
```diff
@@ -0,0 +1,54 @@
...
+
+For example, you can register it in your `/config/application.rb` initializer:
+
+```ruby
+# /config/application.rb
+module MyStore
+  class Application < Rails::Application
+    # ...
+
+    initializer 'spree.register.calculators' do |app|
```
kennyadsl: why `calculators`?

aldesantis: @kennyadsl fixed.

guides/source/developers/shipments/stock-location-sorters.html.md
```diff
@@ -0,0 +1,54 @@
...
+class Spree::Stock::Sorter::Priority < Spree::Stock::Sorter::Base
+  def sort
+    stock_locations.order(priority: :asc)
+  end
+end
+```
+
+### Switching the sorter
+
+Once you have created the logic for the new shipping calculator, you need to
```
kennyadsl: maybe this text as well?

kennyadsl:
@aldesantis can you please rebase against master here?

gmacdougall:
There's just a failing spec that needs to be addressed.

core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -23,7 +23,7 @@ def initialize(order, inventory_units = nil)
         @order = order
         @inventory_units = inventory_units || InventoryUnitBuilder.new(order).units
         @splitters = Spree::Config.environment.stock_splitters
-        @stock_locations = Spree::StockLocation.active
+        @stock_locations = Spree::Config.stock.sorter_class.new(Spree::StockLocation.active).sort
```
tvdeyen: We should name this 

```rb
Spree::Config.stock.location_sorter_class
```

core/app/models/spree/stock/sorter/base.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Spree
+  module Stock
+    module Sorter
```
tvdeyen: The `Sorter` module should be renamed

```rb
Spree::Stock::LocationSorter::Base
```

core/app/models/spree/stock/sorter/default_first.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Spree
+  module Stock
+    module Sorter
```
tvdeyen: ditto

core/lib/spree/core/stock_configuration.rb
```diff
@@ -5,6 +5,7 @@ module Core
     class StockConfiguration
       attr_writer :coordinator_class
       attr_writer :estimator_class
+      attr_writer :sorter_class
```
tvdeyen:
```rb
location_sorter_class
```

core/lib/spree/core/stock_configuration.rb
```diff
@@ -15,6 +16,11 @@ def estimator_class
         @estimator_class ||= '::Spree::Stock::Estimator'
         @estimator_class.constantize
       end
+
+      def sorter_class
```
tvdeyen: ditto

core/spec/lib/spree/core/stock_configuration_spec.rb
```diff
@@ -37,4 +37,21 @@
       end
     end
   end
+  describe '#sorter_class' do
```
tvdeyen: Missing newline

core/app/models/spree/stock/simple_coordinator.rb
BenMorganIO: Yeah, method docs plus guide docs would be great. I recently needed something like this.

core/app/models/spree/stock/sorter/base.rb
BenMorganIO: It would also be really good to add some top-level docs as well.

core/spec/models/spree/stock/sorter/default_first_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+module Spree
+  module Stock
+    module Sorter
+      RSpec.describe DefaultFirst, type: :model do
```
BenMorganIO: This looks weird. Why not just do:

```ruby
describe Spree::Stock::Sorter::DefaultFirst do
end
```

BenMorganIO: Why the need for `type: :model`?

core/lib/spree/core/stock_configuration.rb
BenMorganIO: Can also do:

```ruby
attr_writer :coordinator_class, :estimator_class, :sorter_class
```

core/spec/models/spree/stock/sorter/default_first_spec.rb
aldesantis: @BenMorganIO I agree that it's weird, but this is how specs are written for the other models too (including `type: :model`), so I chose to stick with the convention here.

aldesantis:
@BenMorganIO @tvdeyen @kennyadsl I addressed the requested changes (sorry for the delay!) in separate commits so you can look at them. Let me know if it all looks good and I will squash the commits together.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Solidus 2.7.0 (2018-09-14)
 
+### Core
```
kennyadsl: Wait, this should stay under 2.8.0 instead!

aldesantis: Whoops, fixing!

aldesantis: @kennyadsl this was fixed, can you check?

aldesantis:
@tvdeyen @BenMorganIO I think this might be ready for merging?

kennyadsl:
@aldesantis specs failures seem related, can you please take a look?

aldesantis:
@kennyadsl sorry about that, it should be okay now.

aldesantis:
🎉

PR #547, opened 2015-11-26
(api) don't allow customers to update completed orders

carchrae:
don't allow customers to update a completed order (eg, line_item adjustments, etc)

see https://github.com/solidusio/solidus/issues/546

PR #2787, opened 2018-06-29
Fix permissions for users to change their own orders 

kennyadsl:
ref #546 and #2772 

With this PR the default permissions set does not allow to change an order after its completion by its owner (other users cannot change it anyway 🔐 )

This is what the majority of users expect. 

I also changed some specs which were just testing the wrong thing, more details in the commit messages.

cc @sebfie

### DIFF
--- a/core/lib/spree/permission_sets/default_customer.rb
+++ b/core/lib/spree/permission_sets/default_customer.rb
@@ -11,6 +11,9 @@ module Spree
         can [:read, :update], Order do |order, token|
           order.user == user || (order.guest_token.present? && token == order.guest_token)
         end
+        cannot :update, Order do |order|
+          order.completed?
+        end
         can :create, ReturnAuthorization do |return_authorization|
           return_authorization.order.user == user
         end
--- a/api/spec/requests/spree/api/checkouts_controller_spec.rb
+++ b/api/spec/requests/spree/api/checkouts_controller_spec.rb
@@ -329,8 +329,7 @@ module Spree
       it "returns the order if the order is already complete" do
         order.update_columns(completed_at: Time.current, state: 'complete')
         put spree.api_checkout_path(order.to_param), params: { order_token: order.guest_token }
-        expect(json_response['number']).to eq(order.number)
-        expect(response.status).to eq(200)
+        assert_unauthorized!
       end
 
       # Regression test for https://github.com/spree/spree/issues/3784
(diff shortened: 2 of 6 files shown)


core/lib/spree/permission_sets/default_customer.rb
```diff
@@ -11,6 +11,9 @@ def activate!
         can [:read, :update], Order do |order, token|
           order.user == user || (order.guest_token.present? && token == order.guest_token)
         end
+        cannot :update, Order do |order|
+          order.completed?
+        end
```
tvdeyen: Why not have two seperate rules? One read and one update rule. You mentioned that in your commit message. What was the reason you did it this way?

kennyadsl: I started with that in mind and this has been my [first attempt commit](https://github.com/nebulab/solidus/commit/5bba2fbe572433c04ddf615fed4a32a39cf4ef16). Than I saw #547 and investigate a bit more in the cancancan doc and I found [this page in their wiki](https://github.com/CanCanCommunity/cancancan/wiki/Ability-Precedence) which says that rules are logically or'ed and I think it's a good way to avoid duplicate logic or move it into a method which I don't know where to place.

What do you think? I'd go more with this and change the commit message to better reflect its content (thanks for noticing that)

sebfie: For me, it seems the right way to do.

core/lib/spree/permission_sets/default_customer.rb
```diff
@@ -11,6 +11,9 @@ def activate!
         can [:read, :update], Order do |order, token|
           order.user == user || (order.guest_token.present? && token == order.guest_token)
         end
+        cannot :update, Order do |order|
+          order.completed?
```
BenMorganIO: Could this be done as `cannot :update, Order, &:completed?`

core/lib/spree/permission_sets/default_customer.rb
tvdeyen: It is absolutely fine 👌

core/lib/spree/permission_sets/default_customer.rb
kennyadsl: Sure, but I prefer this way for clarity and cleaner git history if the block content changes in the future.

kennyadsl:
[Specs ran successfully](https://circleci.com/gh/nebulab/solidus/tree/fix-completed-order-permissions), not sure what's happening, I'll try to re-trigger this build.

Issue #2791, opened 2018-07-04
only one taxon_id and option_type_id possible when creating product


hefan:
When you create a product and try to add more than one taxon_id or more than one option_type_id only the first one is saved.

During Update it works as it should.

### Steps to reproduce
Create a new product and add more than one option_type_id or taxon_id to it. Click Create.

### Expected behavior
The new Product should be created with more than one option type or taxon assigned

### Actual behavior
Only the first option type or taxon is assigned. you need to assign the rest by updating the product again.

### System configuration
**Solidus Version**:
v2.6.0

**Extensions in use**:
* solidus (2.6.0)
* solidus_api (2.6.0)
* solidus_auth_devise (2.1.0)
* solidus_backend (2.6.0)
* solidus_core (2.6.0)
* solidus_frontend (2.6.0)
* solidus_i18n (2.0.0)
* solidus_menus (1.0)
* solidus_sample (2.6.0)
* solidus_static_content (1.0.0 eae8e82)
* solidus_support (0.2.1)

### Possible Solution

Just Split the taxon_ids and option_type_ids params similar like in the update function:
https://github.com/solidusio/solidus/blob/master/backend/app/controllers/spree/admin/products_controller.rb

Something like:

```
before_action :split_params, only: [:create]

  def split_params
    if params[:product][:taxon_ids].present?
      params[:product][:taxon_ids] = params[:product][:taxon_ids].split(',')
    end
    if params[:product][:option_type_ids].present?
      params[:product][:option_type_ids] = params[:product][:option_type_ids].split(',')
    end
  end
```

jacobherrington:
I am able to replicate this issue - planning on taking a look at it later today.

jacobherrington:
Fix for this in #2840 pending review.

jacobherrington:
@kennyadsl can this issue be closed now that you have merged #2840?

kennyadsl:
Sure, thanks for pointing out!

PR #2793, opened 2018-07-06
Allow multiple taxons and/or option possible when creating product

jacobherrington:
add private methods for splitting up taxons and option_type_ids on the products controller, addressing a bug that only allowed for one of each to be applied on product creation

refs #2791

### DIFF
--- a/backend/app/controllers/spree/admin/payment_methods_controller.rb
+++ b/backend/app/controllers/spree/admin/payment_methods_controller.rb
@@ -67,10 +67,11 @@ module Spree
       end
 
       def validate_payment_method_type
-        requested_type = params[:payment_method].delete(:type)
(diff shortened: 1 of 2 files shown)


backend/app/controllers/spree/admin/products_controller.rb
```diff
@@ -6,6 +6,7 @@ class ProductsController < ResourceController
       helper 'spree/products'
 
       before_action :load_data, except: [:index]
+      before_action :split_taxons, :split_option_types, only: [:create, :update]
```
kennyadsl: I'd prefer splitting these in two lines, so git history is easier to read when these change. But it's just a non-blocking preference 🙂

jacobherrington:
This was closed unintentionally. I'll make a new PR for this issue shortly.

Issue #2743, opened 2018-05-23
Require Ruby 2.3+


gmacdougall:
Ruby 2.2 has passed its end of life. We should migrate our hound and other required ruby versions to Ruby 2.3.

PR #2794, opened 2018-07-06
Upgrade ruby to 2.3

jtapia:
Issue: https://github.com/solidusio/solidus/issues/2743

Ruby 2.2 has passed its end of life. This changed the required ruby versions to Ruby 2.3

### DIFF
--- a/guides/source/developers/getting-started/first-time-installation.html.md
+++ b/guides/source/developers/getting-started/first-time-installation.html.md
@@ -1,7 +1,7 @@
 # First-time installation
 
 This article will help you install and run Solidus on your local machine for the
-first time. This guide is aimed specifically at developers running macOS. 
+first time. This guide is aimed specifically at developers running macOS.
 
 If you run Linux or Windows, or you don't use [Homebrew][brew] on your Mac, you
 can still follow this guide. However, you may want to consult other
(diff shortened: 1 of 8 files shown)


guides/source/developers/getting-started/first-time-installation.html.md
```diff
@@ -63,14 +63,14 @@ Homebrew][ruby-homebrew] if you need to upgrade from your system's Ruby.
...
 gem install rails
 ```
 
 See more detailed installation information below.
 
-### Upgrade Ruby on macOS 
+### Upgrade Ruby on macOS
 
 If you run macOS Sierra or an older OS, you system's version of Ruby will need
 to be upgraded from 2.0.x to 2.2.2 or newer. You can check what version of Ruby
```
kennyadsl: maybe we should also change this line?

jtapia: done, thanks

BenMorganIO:
I'm :-1: on this as well. Because if a store isn't able to reach 2.3, we could leave them out as Solidus goes forward.

ericsaupe:
I agree with @tvdeyen we should support whatever the Rails versions are supporting.

BenMorganIO:
@ericsaupe That would go against Solidus principles, which is to make upgrades easier for users. If we remove a version of ruby, we need a practical reason for it. Mainly, I think, it's when that version of ruby stops working with the minimal version of Rails we're permitting. So, if Rails 5.2 supports 2.2, which I think it does, we should keep it.

PR #2795, opened 2018-07-06
Allow customizing the reimbursement mailer class

jtapia:
Adds the ability to replace default reimbursement mailer class with a custom one, so there's no need to override default partials

### DIFF
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -316,6 +316,14 @@ module Spree
     #   signature as Spree::PromotionCodeBatchMailer.promotion_code_batch_finished.
     class_name_attribute :promotion_code_batch_mailer_class, default: 'Spree::PromotionCodeBatchMailer'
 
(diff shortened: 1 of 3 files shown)


core/lib/spree/app_configuration.rb
```diff
@@ -307,6 +307,14 @@ def default_pricing_options
     # @api experimental
     class_name_attribute :shipping_rate_tax_calculator_class, default: 'Spree::TaxCalculator::ShippingRate'
 
+    # Allows providing your own Mailer for order mailer.
```
kennyadsl: this comment needs to be changed

jtapia: thanks, done

kennyadsl:
Specs are green here as well:

https://circleci.com/gh/solidusio/solidus/8573
https://circleci.com/gh/solidusio/solidus/8572
https://circleci.com/gh/solidusio/solidus/8571
https://circleci.com/gh/solidusio/solidus/8570

PR #2796, opened 2018-07-06
Allow customizing the promotion code batch mailer class

jtapia:
Adds the ability to replace default promotion code batch mailer class with a custom one, so there's no need to override default partials

### DIFF
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -307,6 +307,15 @@ module Spree
     # @api experimental
     class_name_attribute :shipping_rate_tax_calculator_class, default: 'Spree::TaxCalculator::ShippingRate'
 
+    # Allows providing your own Mailer for promotion code batch mailer.
+    #
+    # @!attribute [rw] promotion_code_batch_mailer_class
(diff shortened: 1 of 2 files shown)


core/lib/spree/app_configuration.rb
```diff
@@ -307,6 +307,15 @@ def default_pricing_options
     # @api experimental
     class_name_attribute :shipping_rate_tax_calculator_class, default: 'Spree::TaxCalculator::ShippingRate'
 
+    # Allows providing your own Mailer for order mailer.
```
kennyadsl: This comment needs to be updated

core/lib/spree/app_configuration.rb
```diff
@@ -307,6 +307,15 @@ def default_pricing_options
...
+    #
+    # @!attribute [rw] order_mailer_class
```
kennyadsl: here as well

core/lib/spree/app_configuration.rb
jtapia: thanks, done

core/lib/spree/app_configuration.rb
jtapia: thanks, done

PR #2800, opened 2018-07-18
Skip forgery protection in api controllers

### LINKED ISSUE #2799: API endpoints produces InvalidAuthenticityToken with POST
### Steps to reproduce
``curl -X POST -k -H 'X-Spree-Token: 1111111111111111111111111111111' -i 'https://example.com/api/orders'``

I get ``"ActionController::InvalidAuthenticityToken in Spree::Api::OrdersController#create"
in actionpack (5.2.0) lib/action_controller/metal/request_forgery_protection.rb:211:in `handle_unverified_request'``

### Expected behavior
The API endpoints should not require an authenticity_token.

In Spree, the CSRF protection is disabled for JSON/XML requests:
``protect_from_forgery unless: -> { request.format.json? || request.format.xml? }`` (https://github.com/spree/spree/blob/master/api/app/controllers/spree/api/base_controller.rb)

this is missing in https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/base_controller.rb

Maybe I am missing something, but I did not found a way to get the API to work correctly.

### System configuration
**Solidus Version**: 2.6.0

**Extensions in use**: none

kennyadsl:
Fixes #2799

Rails is now enabling forgery protection by default so we need to explicitly disable it for api requests, as described here:

http://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html

This PR also enables forgery protection by default in the specs dummy app so that we can really test that the api code is working in a real Rails 5.2+ environment.

### DIFF
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -13,8 +13,15 @@ Rails.env = 'test'
 require 'solidus_core'
 
 # @private
+def forgery_protected_by_default?
+  Gem::Version.new(Rails.version) >= Gem::Version.new('5.2')
+end
+
+# @private
 class ApplicationController < ActionController::Base
-  protect_from_forgery with: :exception
+  if !forgery_protected_by_default?
+    protect_from_forgery with: :exception
+  end
 end
 
 # @private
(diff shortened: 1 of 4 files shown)


core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -12,9 +12,16 @@
 
 require 'solidus_core'
 
+# @private
+def forgery_protected_by_default?
+  Gem::Version.new(Rails.version) >= Gem::Version.new('5.2')
```
BenMorganIO: Is this still needed with the latest master?

BenMorganIO: Yep, still needed: https://github.com/solidusio/solidus/blob/master/core/solidus_core.gemspec#L27

PR #2802, opened 2018-07-18
Remove default address dependency part 2

### LINKED ISSUE #2930: Reimbourses with store credits is not assigning credits
### Steps to reproduce

1. Create a shipped order associated to a registered user
2. Create a RMA with Reimbursement Type: Store Credits
3. Create a Return
4. Reimbourse the Order
5. Go to the admin user page under `admin/users/:id/store_credits`

### Expected behavior

I should find the credits assigned to that user.

### Actual behavior

No store credits are assigned to that user.

### System configuration

**Solidus Version**: master (via sandbox)
**Extensions in use**: none

kennyadsl:
Continuation of #2686
Closes #2930 

This PR continues the work of @gevann removing hard-coded dependency of having a specific user in the database to perform a set of operations related to reimbursements. 

@tvdeyen this basically seconds your comments on that PR, let me know if it's enough!

I've also deprecated `whodunnit` argument from some methods in favor of a more consistent `created_by`.

### DIFF
--- a/core/app/models/spree/reimbursement.rb
+++ b/core/app/models/spree/reimbursement.rb
@@ -98,12 +98,16 @@ module Spree
       total - paid_amount
     end
 
-    def perform!
+    def perform!(creator: nil)
+      unless creator
+        creator = Spree.user_class.find_by(email: 'spree@example.com')
+        Spree::Deprecation.warn("Calling #perform on #{self} without creator is deprecated")
+      end
       reimbursement_tax_calculator.call(self)
       reload
       update!(total: calculated_total)
 
-      reimbursement_performer.perform(self)
+      reimbursement_performer.perform(self, creator: creator)
 
       if unpaid_amount_within_tolerance?
         reimbursed!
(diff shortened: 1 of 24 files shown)


core/app/models/spree/reimbursement.rb
```diff
@@ -98,12 +98,16 @@ def unpaid_amount
       total - paid_amount
     end
 
-    def perform!
+    def perform!(created_by: nil)
+      unless created_by
+        created_by = Spree.user_class.where(email: 'spree@example.com').first
```
BenMorganIO: Use `find_by` instead of `where.first`.

kennyadsl: Done and squashed changes into 3f33fa199e889251c517a60c4ac98be958248b4c

kennyadsl:
@ericsaupe @gmacdougall I've addressed the concern we talked about in the core team meeting. This should be fine now!

PR #2803, opened 2018-07-19
Add article that documents a Spree::Order journey

benjaminwil:
The initial version of this article tracks a `Spree::Order`, from before it is created (by a customer adding an item to the cart), to the point when an order is shipped and received by the customer.

### DIFF
--- a/guides/source/developers/getting-started/customer-flow.html.md
+++ b/guides/source/developers/getting-started/customer-flow.html.md
@@ -1,0 +1,249 @@
+# Customer flow
+
+This article outlines a typical customer flow in Solidus on a technical level.
+The [`Spree::Order`][spree-order] model ties much of this flow together.
+However, every order placed touches many Solidus models.
+
+Note that this article does not detail _every_ single interaction that a Rails
+server logs. Nor does it detail every single attribute on every single relevant
+model. For example, it does not mention the `special_instructions` that a
+customer might add to an order, or any of the order cancellation process.
+
+## Standard flow
+
+Customer arrives at the store's URL.
+
+- Solidus determines whether the customer is _new_ or _returning_.
+  - If returning, Solidus determines whether the customer is a guest.
+    - If customer was logged in during a previous session, Solidus determines
+      whether customer should be logged in again.
+  - If new, `guest_token` cookie might be generated for the customer.
+- Solidus routes the user to the correct view (including the correct
+  permissions).
+- Store loads entrance page for user.
+
+Customer navigates to a product page.
+
+Customer chooses a variant of the product.
+
+Customer chooses a quantity of the product.
+
+Customer adds the product to their cart.
+
+- A [`Spree::Order`][spree-order] object is created and given an `id`.
+  - The order has a `state` of `cart`.
+- A `Spree::LineItem` object is created for each item in the cart.
+  - The line item records which [`Spree::Variant`][spree-variant] of a product
+    the customer added to the cart.
+  - The line item records the quantity of the `Spree::Variant` added to
+    the cart.
+  - The price of the variant is added to the line item.
+  - The line item is associated with the customer's order using its
+    `order_id` value.
+  - The taxes and price adjustments start at `0.0`: they cannot be
+    calculated until after the Solidus knows the customer's address.
+  - The `tax_category_id` for each item is set according to the [tax
+    category][tax-categories] of each item in the cart.
+
+[spree-order]: overview.html
+[spree-variant]: ../products-and-variants/variants.html
+[tax-categories]: ../taxation/overview.html#tax-categories
+
+If the customer is not logged in, the user is asked to log into their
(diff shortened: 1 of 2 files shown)


guides/source/developers/getting-started/customer-flow.html.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Standard flow
+
+Customer arrives at the store's URL.
+
+- Solidus determines whether the customer is _new_ or _returning_.
+  - If returning, Solidus determines whether the customer is a guest.
+  - If customer was logged in during a previous session, Solidus determines
+    whether customer should be logged in again.
+    - If new, `guest_token` cookie might be generated for the customer.
```
kennyadsl: too much indent?

guides/source/developers/getting-started/customer-flow.html.md
```diff
@@ -0,0 +1,247 @@
...
+[spree-order]: overview.html
+[spree-variant]: ../products-and-variants/variants.html
+[tax-categories]: ../taxation/overview.html#tax-categories
+
+If the customer is not logged in, the user is asked to log into their
+account or continue the checkout process as a guest.
+
+- Customer logs in or continues checkout as a guest.
+  - The `Spree::Order` is updated.
+    - If the customer logs into an existing account, their `Spree::User`
```
kennyadsl: maybe we should mention that at this point the order could be merged with a previous user's order?

guides/source/developers/getting-started/customer-flow.html.md
```diff
@@ -0,0 +1,247 @@
...
+[spree-payment-method]: ../payments/payment-methods.html 
+[spree-payment-source]: ../payments/payment-sources.html
+
+Customer enters their preferred payment details and submits them.
+
+Customer reviews the confirmation screen and places order.
+
+- The `Spree::Payment`'s `state` changes from `checkout` to `pending`.
+
+Administrator approves the shipment.
```
kennyadsl: I'm not sure that this is needed actually. I think that shipments become ready when the payment is paid (no balance due), no approval needed.

guides/source/developers/getting-started/customer-flow.html.md
```diff
@@ -0,0 +1,247 @@
...
+
+- The `Spree::Order`'s `approver_id` and `approved_at` values are no
+  longer `nil`.
+- The [`Spree::StockItem`'s][spree-stock-item] `count_on_hand` value decrease by
```
kennyadsl: doesn't this happen when the shipment is shipped?

kennyadsl (commented):
Left some questions, thanks!

guides/source/developers/getting-started/customer-flow.html.md
benjaminwil: Thank you. This was intentional, but I think that it makes more sense to indent these lists this way:

```
 - If returning, Solidus determines whether the customer is a guest.
    - If customer was logged in during a previous session, Solidus determines
      whether customer should be logged in again.
  - If new, `guest_token` cookie might be generated for the customer.
```

Because the "If customer was logged in..." item relates specifically to the returning customers item above.

guides/source/developers/getting-started/customer-flow.html.md
benjaminwil: Thank you, I had overlooked that. I will be adding a new list item to account for this functionality.

guides/source/developers/getting-started/customer-flow.html.md
benjaminwil: I will run through an order and double-check this now. Thank you.

My understanding was that a sold stock item is no longer counted as on hand, because it has been sold and will be shipped.

benjaminwil:
@kennyadsl Thanks for your comments. I just updated the commit to fix the issues you commented on.

guides/source/developers/getting-started/customer-flow.html.md
```diff
@@ -0,0 +1,249 @@
...
+
+Customer reviews the confirmation screen and places order.
+
+- The `Spree::Payment`'s `state` changes from `checkout` to `pending`.
+
+Administrator can opt to manually approve the shipment. (Optional.)
+
+- The `Spree::Order`'s `approver_id` and `approved_at` values are no
+  longer `nil`.
+- The [`Spree::StockItem`'s][spree-stock-item] `count_on_hand` value decrease by
```
kennyadsl: Sorry, I still think this should not stay here, doesn't this happen when the order is completed without the need for the admin to do any approval?

benjaminwil: Oops, thank you! Fixing.

gmacdougall (approved):
Thanks Ben!

tvdeyen (approved):
This is very useful. Thanks for writing that

PR #2804, opened 2018-07-19
Change i18n selection for select2

afdev82:
I have a problem with the locale "en-US".
The select2 doesn't find the translations.
This PR it's a way to fix the problem.

### DIFF
--- a/backend/app/views/spree/admin/shared/_js_locale_data.html.erb
+++ b/backend/app/views/spree/admin/shared/_js_locale_data.html.erb
@@ -33,7 +33,7 @@
 <script data-hook='admin-custom-translations'>
 </script>
 
-<% if I18n.locale != :en %>
+<% if I18n.locale.to_s.split('-').first != 'en' %>
   <%= javascript_include_tag "solidus_admin/select2_locales/select2_locale_#{I18n.locale}",
     data: {turbolinks_track: 'reload'} %>
 <% end %>


kennyadsl:
is `en-US` the only locale that you are using?

afdev82:
No, I'm using also 'de', 'en', 'fr'

kennyadsl:
Which version of solidus and solidus_i18n are you using?

kennyadsl:
Anyway, what about creating a new `select2_locale_en-US.js` file [here]( https://github.com/solidusio/solidus/tree/master/backend/vendor/assets/javascripts/solidus_admin/select2_locales) instead?

afdev82:
Solidus is the v2.4.2, but the same code is also in the master.
Solidus_18n is the v1.2.0.

afdev82:
Because I don't need different settings for my locale, in general the locale of the store is set to "en-US" because I need different currency formats, but for select2 it should fallback to en.
IMHO it's not handled correctly in the code, the fix is not completely general, because if I wanted to have the js file for "en-US" it doesn't work, but as far as I saw the files are only simply translations in different languages and I think it's fine in this way.

> Anyway, what about creating a new select2_locale_en-US.js file here instead?

Anyway, I'm using custom forks of the two projects with small changes. This PR was only to contribute.

kennyadsl:
> Anyway, I'm using custom forks of the two projects with small changes. This PR was only to contribute.

Thanks for contributing but I think that, as you said, this change is going to make life harder if someone just want have different content in their `en-US` file. I think we can try another way that is good for everyone. If you want/have time to find alternatives please go ahead, otherwise I'll try to take a deeper look at this soon. Thanks again.

afdev82:
See #2805

PR #2805, opened 2018-07-19
add en-US locale to select2

afdev82:
This adds the en-US file to avoid errors with en-US locale.

It could be used to overwrite some translations if needed.


kennyadsl:
Merging since failures on Netlify are not related and are solved on master

PR #2808, opened 2018-07-20
Add Json Web Token functionality to Spree

skukx:
Enable Spree api to use json web tokens as an authentication method.
The api should still support `spree_api_tokens` by falling back if no
json web token is provided.

The algorithm, secret, ayload options, and expiration should all be
configurable through `Spree::Config`.

See: https://jwt.io/introduction/
See: https://github.com/jwt/ruby-jwt


api/app/controllers/spree/api/base_controller.rb
```diff
@@ -103,6 +107,14 @@ def api_key
       end
       helper_method :api_key
 
+      def json_web_token
+        @json_web_token ||= JWT.decode(api_key, Spree::Config.jwt_secret,
+                              true, algorithm: Spree::Config.jwt_algorithm).first
+      rescue JWT::DecodeError => e
+        Rails.logger.warn e
```
skukx: May be too high of a log level

core/lib/spree/app_configuration.rb
```diff
@@ -137,6 +137,26 @@ class AppConfiguration < Preferences::Configuration
...
+    # @!attribute [rw] jwt_expiration
+    #   @return [String] How long until the token expires in seconds.
+    #   (default: +3600+)
+    preference :jwt_expiration, :integer, default: 3600
+
+    # @see https://github.com/jwt/ruby-jwt#algorithms-and-usage
+    # @!attribute [rw] jwt_options
+    #   @return [String] The options to pass into `Spree::User#as_json` when
+    #   when creating the jwt payload. (default: `{ only: %i[email first_name id last_name] }`)
+    preference :jwt_options, :hash, default: { only: %i[email first_name id last_name] }
```
skukx: Not a fan of this

core/app/controllers/spree/base_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'cancan'
+require 'jwt'
```
kennyadsl: why this is needed here?

core/lib/spree/core/controller_helpers/auth.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'cancan'
+require 'jwt'
```
kennyadsl: is it needed here?

core/lib/spree/core/controller_helpers/auth.rb
```diff
@@ -28,6 +29,17 @@ module Auth
           rescue_from CanCan::AccessDenied do
             instance_exec(&unauthorized_redirect)
           end
+
+          # Triggers after devise has logged a user in.
+          def after_sign_in_path_for(resource)
```
kennyadsl: does this happen only when a user logs in via web app?

api/app/controllers/spree/api/base_controller.rb
```diff
@@ -42,7 +42,11 @@ def permitted_line_item_attributes
       end
 
       def load_user
-        @current_api_user ||= Spree.user_class.find_by(spree_api_key: api_key.to_s)
+        @current_api_user ||= if json_web_token.present?
```
kennyadsl: We could even use another separate header maybe? Is there some benefit in using `X-Spree-Token` for JWT as well?

core/app/models/concerns/spree/user_api_authentication.rb
```diff
@@ -2,6 +2,21 @@
...
+      # @see https://github.com/jwt/ruby-jwt#support-for-reserved-claim-names
+      extras = {}
+      extras['exp'] = Time.current.to_i + expires_in if expires_in.present?
+      extras['iat'] = Time.current
+
+      payload = as_json(options)
+      payload.merge(extras)
+
+      JWT.encode payload, Spree::Config.jwt_secret, Spree::Config.jwt_algorithm
+    end
```
kennyadsl: who saves this token on the User?

api/app/controllers/spree/api/base_controller.rb
skukx: I figured it would create backward compatibility as well as forward compatibility by allowing both options

core/app/models/concerns/spree/user_api_authentication.rb
skukx: The token never gets saved to the database and has no need to.  It is encrypted with a secret, (By default Rails.application.secret_key_base). If the token fails to decrypt it is unauthentic or has been tampered with.

core/lib/spree/core/controller_helpers/auth.rb
skukx: Correct
see: https://www.rubydoc.info/github/plataformatec/devise/Devise/Controllers/Helpers:after_sign_in_path_for

When a user has logged in through the Devise controller this method is called.

core/lib/spree/core/controller_helpers/auth.rb
skukx: Probably not, I wondered that myself

core/app/controllers/spree/base_controller.rb
skukx: Probably not, I wondered that myself

skukx:
@kennyadsl You're right I haven't written specs around this yet. If this is something the community/core team likes then I'll continue work on it.

kennyadsl:
Today I did some work on JWT and I have a concern regarding refresh policy/endpoint. Do you have any thought on that? Do you think we should include some work for that in this PR?

skukx:
@kennyadsl That part gets a little complicated and right now it would piggy back off of Devise Sessions controller.

Think of whatever server is performing our authentication as our AuthServer.

When a user token expires, front-end app redirects to the AuthServer. The AuthServer would either recognize the user or it wouldn't.

If the AuthServer recognizes the user (still logged in), then it generates a new token and sends it back. See: https://github.com/solidusio/solidus/blob/b8258bffd9c42ff934c184cf598eb47512ccb26f/core/lib/spree/core/controller_helpers/auth.rb#L34-L42

However if the user is no longer authenticated with the AuthServer, then the AuthServer renders the login page for the user to authenticate.

core/app/controllers/spree/base_controller.rb
BenMorganIO: Do `p require 'jwt'`. If you get `false`, you'll know soon enough.

ericsaupe:
I wonder if this would be better started as an extension to establish the additional features of JWTs like refresh policy, timeout config, etc.

Thoughts @skukx ?

skukx:
Yes, when I get time, I'm probably going to do this. It can be hard to make a big change like this without already seeing it in action. I'll probably release a gem that does this then if it is proven useful we can move it in.

skukx:
This has been moved to an extension. See: https://github.com/skukx/solidus_jwt

PR #2809, opened 2018-07-20
Quick switch part 2

kennyadsl:
This PR is a continuation of #2137 to add missing feature specs.

This also removes an useless variable from JS.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/components/shortcuts.js
+++ b/backend/app/assets/javascripts/spree/backend/components/shortcuts.js
@@ -1,0 +1,19 @@
+$(function() {
+  var quickSwitch = new Spree.Views.QuickSwitch({
+    el: $("[data-js='quick-switch']")
+  });
+
+  var onKeypress = function(e) {
+    if(
+      e.target.tagName === "INPUT" ||
+      e.target.tagName === "SELECT" ||
+      e.target.tagName === "TEXTAREA"
+    ) {
+      return
+    } else if(e.key === "@") {
+      quickSwitch.triggerShortcut()
+    }
+  };
+
+  $(document.body).on("keypress", onKeypress);
+});
--- a/backend/app/controllers/spree/admin/quick_switch_controller.rb
+++ b/backend/app/controllers/spree/admin/quick_switch_controller.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module Spree
+  module Admin
+    class QuickSwitchController < Spree::Admin::BaseController
+      layout false
+
+      def find_object
+        quick_switch_item = Spree::Backend::Config.quick_switch_items.detect do |item|
+          item.search_triggers.include? searched_key.to_sym
+        end
+
+        if quick_switch_item
+          if object = quick_switch_item.finder.call(searched_value)
+            render(
+              json: {
+                redirect_url: quick_switch_item.url.call(object)
+              },
+              status: :ok
+            )
+          else
+            render(
+              json: {
+                message: quick_switch_item.not_found_text(searched_value)
+              },
+              status: :not_found
+            )
+          end
+        else
+          render(
+            json: {
+              message: I18n.t("invalid_query", scope: "spree.quick_switch")
+            },
+            status: :bad_request
+          )
+        end
+      end
+
+      private
+
+      def searched_key
+        params[:quick_switch_query].split(" ")[0]
+      end
+
+      def searched_value
+        params[:quick_switch_query].split(" ")[1]
+      end
+    end
+  end
+end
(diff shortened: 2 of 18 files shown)


kennyadsl:
I've deleted houndci-bot comments since it was not detecting frozen_string_literal comment (that I've added) correctly.

backend/spec/features/admin/quick_switch_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+  context 'as admin user' do
+    stub_authorization!
+
+    context "visiting any page" do
+      let(:initial_page) { spree.admin_orders_path }
+      before(:each) { visit initial_page }
+
+      it 'can open the quick switch' do
+        open_quick_switch
+        expect(page).to have_css("#quick-switch", visible: true)
```
kennyadsl: @tvdeyen after your review I just changed this line from:

```ruby
expect(page).to have_content("How do I use this") # TODO: check visible class
```

since I did left that `TODO` but totally forgot to fix it before pushing.

tvdeyen: 👍

backend/app/assets/javascripts/spree/backend/components/shortcuts.js
```diff
@@ -0,0 +1,19 @@
...
+
+  var onKeypress = function(e) {
+    if(
+      e.target.tagName === "INPUT" ||
+      e.target.tagName === "SELECT" ||
+      e.target.tagName === "TEXTAREA"
+    ) {
+      return
+    } else if(e.key === "@") {
+      quickSwitch.triggerShortcut()
```
BenMorganIO: Missing semicolon.

backend/app/assets/javascripts/spree/backend/components/shortcuts.js
```diff
@@ -0,0 +1,19 @@
...
+    el: $("[data-js='quick-switch']")
+  });
+
+  var onKeypress = function(e) {
+    if(
+      e.target.tagName === "INPUT" ||
+      e.target.tagName === "SELECT" ||
+      e.target.tagName === "TEXTAREA"
+    ) {
+      return
```
BenMorganIO: Missing semicolon.

backend/app/controllers/spree/admin/quick_switch_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+          end
+        else
+          render(
+            json: {
+              message: I18n.t("invalid_query", scope: "spree.quick_switch")
+            },
+            status: :bad_request
+          )
+        end
+      end
```
BenMorganIO: This is a really big method. Maybe move the renders to their own private methods? Like `render_redirect_to`, `render_not_found`, and `render_invalid_query`. You may have better naming than I.

backend/app/models/spree/backend_configuration.rb
```diff
@@ -64,60 +31,140 @@ def initialize(
...
+          end,
+          help_text_key: :user_help,
+          not_found_text_key: :user_not_found
+        ),
+        Spree::QuickSwitchItem.new(
+          search_triggers: [:p, :product, :sku, :v, :variant],
+          finder: ->(searched_value) do
+            Spree::Variant.find_by(sku: searched_value)
+          end,
+          url: ->(variant) do
```
BenMorganIO: Since this is multi-line, the lambda should be `lambda do |variant|`

backend/app/models/spree/menu_item.rb
```diff
@@ -0,0 +1,36 @@
...
+    # @param url [String] A url where this link should send the user to
+    def initialize(
+      sections,
+      icon,
+      condition: nil,
+      label: nil,
+      partial: nil,
+      url: nil
+    )
+
```
BenMorganIO: Extra new line.

backend/config/routes.rb
```diff
@@ -4,6 +4,7 @@
   namespace :admin do
     get '/search/users', to: "search#users", as: :search_users
     get '/search/products', to: "search#products", as: :search_products
+    get "/quick-switch", to: "quick_switch#find_object", as: :quick_switch
```
BenMorganIO: :-1: Would prefer something more restful such as:

```ruby
resource :quick_switch, only: [:create]
```

jarednorman:
My two cents: This is a cool feature, but seems like something that might be better off in an extension rather than having to maintain this whole extensible quick change interface that the average store owner won't even know about in core.

kennyadsl:
💯 

I agree. If someone wants to keep this code and create an extension, we can add it to `solidusio-contrib` org!

PR #2810, opened 2018-07-24
Add an allocator class to extend the Solidus initial allocation logic

vassalloandrea:
With this PR we add the possibility to change the stock allocation logic using a custom class without overriding the allocate_inventory method in SimpleCoordinator.

### Default class 
Without changing the Solidus default stock allocation, I wrote a `Spree::Stock::Allocator::OnHandFirst` allocator with the same logic of the `allocate_inventory` [method](https://github.com/solidusio/solidus/blob/fa98d6523571142fe4c23120cf281f0f6d1b0aa7/core/app/models/spree/stock/simple_coordinator.rb#L87).

### How to override Solidus default stock allocation
You can tell the `SimpleCoordinator` to use a custom allocator overriding the `Spree` configuration: `Spree::Config.stock.allocator_class`

### How to write a custom allocator
To write a custom stock allocator you can extend `Spree::Stock::Allocator::Base` if you also want to change the allocation priority ( on hand first or backorderable first ). Otherwise, you can extend `Spree::Stock::Allocator::OnHandFirst` to only change the allocation logic.

### DIFF
--- a/core/app/models/spree/stock/simple_coordinator.rb
+++ b/core/app/models/spree/stock/simple_coordinator.rb
@@ -11,6 +11,9 @@ module Spree
     #   * Repeat but for backordered inventory
     #   * Combine allocated and on hand inventory into a single shipment per-location
     #
+    # Allocation logic can be changed using a custom class (as
+    # configured in Spree::Config::stock_allocator_class )
+    #
     # After allocation, splitters are run on each Package (as configured in
     # Spree::Config.environment.stock_splitters)
     #
@@ -31,6 +34,8 @@ module Spree
           variants: @desired.variants,
           stock_locations: @stock_locations
         )
+
+        @allocator = Spree::Config.stock.allocator_class.new(@availability)
       end
 
       def shipments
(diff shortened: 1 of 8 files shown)


core/lib/spree/app_configuration.rb
```diff
@@ -381,6 +381,13 @@ def payment_canceller
     # returns a String
     class_name_attribute :taxon_url_parametizer_class, default: 'ActiveSupport::Inflector'
 
+    # Allows providing your own class to allocate the inventory into packages
+    #
+    # @!attribute [rw] taxon_url_parametizer_class
+    # @return [Class] a class that provides a `#allocate_on_hand` and `#allocate_backorderable` method that
+    # returns a Spree::StockQuantities
+    class_name_attribute :stock_allocator_class, default: 'Spree::Stock::Allocator::OnHandFirst'
+
```
tvdeyen: This should go into the stock configuration

core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -31,6 +34,8 @@ def initialize(order, inventory_units = nil)
           variants: @desired.variants,
           stock_locations: @stock_locations
         )
+
+        @allocator = Spree::Config.stock_allocator_class.new(@availability)
```
tvdeyen:
```rb
Spree::Config.stock.allocator_class
```

Issue #1262, opened 2016-06-22
Move reports to their own gem


isaacfreeman:
Working on PR #1250 prompted a discussion with @dhonig, wherein we agreed that having reports in Solidus is a two-edged sword. It's pretty easy to write a report that chews up all available resources and degrades performance on the same server that's serving the front end to customers. On the other hand,  the convenience of having reports right there in Solidus Admin is valuable for many merchants.

It would be good to encourage developers to make a conscious decision about how reports should be provided for each client. Moving them out of the main Solidus distribution into a separate gem would provide that decision point.

jhawthorn:
This makes sense to me. The builtin reports section has been [stagnant for a long time](https://github.com/solidusio/solidus/commits/master/backend/app/controllers/spree/admin/reports_controller.rb) (at least before #1250, which looks great). Moving them to a gem I hope would allow for making faster and more opinionated changes. I'm sure there are more "common" reports we could share in an extension which are desirable across many stores.

jacobherrington:
#2814 will close this issue.

PR #2814, opened 2018-07-28
Move reports from backend into an extension

jtapia:
#### Issue
Based on https://github.com/solidusio/solidus/issues/1262

#### What it does?
- Move Reports admin section into an extension - https://github.com/jtapia/solidus_reports


kennyadsl:
Thanks, I think we should also move the extension into solidusio-contrib organization before merging. What do you think?

jtapia:
@kennyadsl sure, what is the process to move the exension into the `solidus-contrib` organization?

kennyadsl:
I think you just need to give me admin access to that repo and I can transfer it. Let's talk in Slack for details if you want!

jtapia:
@kennyadsl I just sent to you the invite, thanks

kennyadsl:
@jtapia I think I have to be admin to make the transfer

jtapia:
@kennyadsl right, my bad, I just did it

Issue #2801, opened 2018-07-18
Better Error Reporting


skukx:
I'm not sure how the community feels about this but I would be willing to work on this myself if there is enough interest.  Currently many of the Solidus errors are logged like so:

```ruby
Rails.logger.error(<String>)

# Or

Rails.logger.error(e.message)
```

This is fine in a lot of cases, however in certain cases this is not optimal. The reason I say it isn't optimal is because stack trace is eliminated making it harder to debug.  When using reporting services such as `Rollbar` it is nice to have those stack traces.

### Non-Optimal Examples
![image](https://user-images.githubusercontent.com/1091314/42895395-a7f4cec0-8a6e-11e8-9196-edd7d11c3bbf.png)

![image](https://user-images.githubusercontent.com/1091314/42895434-cb9dd092-8a6e-11e8-8505-36e9be9f162a.png)

![image](https://user-images.githubusercontent.com/1091314/42895464-dd7df1ac-8a6e-11e8-9067-bc1d587fed52.png)

### Better Example

![image](https://user-images.githubusercontent.com/1091314/42895503-fb6d19c2-8a6e-11e8-87cb-88a980e1d7bf.png)

### Summary

When ever we rescue an error we should pass the error object itself to the `Logger#error` method. Here are some examples where this practice could be implemented.

https://github.com/solidusio/solidus/blob/485781561beaf5c504c24aa069da62069ea7afe9/frontend/app/controllers/spree/store_controller.rb#L25-L28

https://github.com/solidusio/solidus/blob/859143f3f061de79cc1b385234599422b8ae8e21/backend/app/controllers/spree/admin/payments_controller.rb#L49-L52

https://github.com/solidusio/solidus/blob/859143f3f061de79cc1b385234599422b8ae8e21/backend/app/controllers/spree/admin/payments_controller.rb#L49-L52

Again this is nothing major but more of a quality of life change.

dhonig:
@skukx  Really glad you brought this up.   Really good suggestions on the changes.  It would be great to modernize how errors are reported.  I'm not sure anyone has given this some good thinking in ages.

tvdeyen:
@skukx thanks. Absolutely. Error handling needs some overhaul.

But instead of change the way we log, we should think about adding a global error handler, because like you said we want to be able to report to error tracking services or log, or do both. Having an configurable error handler (that defaults to `Rails.logger.error`) would be a huge benefit.

mtomov:
[...]

PR #2816, opened 2018-08-01
Add error handler discussed in #2801

skukx:
Implements #2801 

Once merge you can log errors with:
```ruby
Spree::Core::ErrorReporter.report(error)
Spree::Core::ErrorReporter.report(error, :debug)
Spree::Core::ErrorReporter.report(error, :info)
Spree::Core::ErrorReporter.report(error, :warn)
Spree::Core::ErrorReporter.report(error, :error)
Spree::Core::ErrorReporter.report(error, :fatal)
Spree::Core::ErrorReporter.report(error, :unknown)
```

If you want to add your own reporter then:
```ruby
class SolidusRollbarReporter < Spree::Core::ErrorReporter::Base
  def self.report(error, severity, metadata)
    Rollbar.send(severity, error, metadata)
  end
end

# Register our new reporter
Spree::Core::ErrorReporter.add_reporter(SolidusRollbarReporter)
```


skukx:
The current structure for errors currently looks looks something like:
```ruby
module Spree
  class Exchange < Spree::Base
    class UnableToCreateShipments < StandardError; end
  end
end
```

Is that a practice we want to follow as far as declaring/defining the errors within the models? Or would we rather do something like:
```
module Spree
  class Error < StandardError; end
  class PaymentSourceError < Error; end
  class CreditCardError < Error; end
end
```
Essentially where the errors are defined in one place rather than disbursed throughout the models.

core/lib/spree/app_configuration.rb
```diff
@@ -390,13 +404,20 @@ def payment_canceller
...
     class_name_attribute :taxon_url_parametizer_class, default: 'ActiveSupport::Inflector'
 
+    ##
+    # Logger for solidus
+    # @return [Logger] The logger which by default outputs to STDOUT.
+    #
+    def logger
+      @logger ||= Logger.new(STDOUT)
+    end
+
```
mtomov: I'm not sure if it's worth keeping a separate logger for Solidus only, and that be different than the `Rails.logger`

core/lib/spree/core/error_handler/base.rb
```diff
@@ -0,0 +1,32 @@
...
+      class Base
+        class << self
+          ##
+          # Generic error handling
+          # @param error [StandardError] The error you want to handle.
+          # @param serverity [Symbol, String] The severity (i.e. debug, info, warn, error, fatal)
+          #
+          def handle(_error, _severity = :error)
+            raise NoMethodError
+          end
```
mtomov: I'd think that we might want to think about the exact sequence and type of parameters, taking into consideration reporting to external services, such as Honeybadger:

https://docs.honeybadger.io/ruby/getting-started/reporting-errors.html#passing-additional-options-to-honeybadger-notify

for example, by leaving out 

```ruby
message = error.respond_to?(:message) ? error.message : error
```

One can't do 

```ruby
ErrorHandler.report("ma mia")
```

which of course might be desirable, not sure..

Also, `severity` sounds like just one possible option, which an error can have. For example Honeybadger supports all of those:

image

And lastly, `context` in the context of Honeybadger is an object with all sorts of additional information, can be for example 

```ruby
{ user_id: 99 }
```

Which usually comes useful to be able to pass as well, either as part of the above `options`, or maybe as an explicit parameter being so useful, as in the example that I suggested:


```ruby
class ErrorHandler
  class << self
    def report(error, context = {}, options = {})
      Honeybadger.notify(error, options.merge(context: context))

      severity = options[:severity] || :error
      message = error.respond_to?(:message) ? error.message : error
      Rails.logger.send(severity, message)
    end
  end
end
```

---

I'm just mentioning those as considerations, feel free to do without them if not deeming necessary at this point.

core/lib/spree/app_configuration.rb
skukx: I think this should be a decision for the developers. Some developers prefer to keep their logs aggregated. However I think it would be ok to default it to `Rails.logger` instead of creating a new logger by default

core/lib/spree/core/error_handler/base.rb
skukx: We need to be careful not to build this with one error reporter in mind.  The implementation does allow overriding if needed. Also:

```ruby
ErrorHandler.report("ma mia")
# and
error = StandardError.new("ma mia")
ErrorHandler.report(error)
```
Will have the same result when it comes to the `Logger` class

There is no need to explicitly perform
```ruby
message = error.respond_to?(:message) ? error.message : error
```

skukx: btw, thank you for your input. This is great stuff

skukx:
From #2816 
Here is a list of current Spree errors being raise for reference:
### List of current spree custom errors
```ruby
Spree::Core::GatewayError
Spree::Exchange::UnableToCreateShipments
Spree::LineItem::CurrencyMismatch
Spree::OrderCapturingFailures
Spree::OrderMutex::LockFailed
Spree::Order::CannotRebuildShipments
Spree::Order::InsufficientStock
Spree::PromotionCodeBatch::CantProcessStartedBatch
Spree::Reimbursement::IncompleteReimbursementError
Spree::StockLocation::InvalidMovementError
Spree::Stock::ShipmentRequired
Spree::Stock::OrderRequired
Spree::Wallet::Unauthorized
```

### List of generic `raise` errors
These are errors that don't raise any specific exception. For example:
```ruby
raise 'Some error'
```

These errors may be fine as is or we may want to create a custom error for
some so they are not classified `RuntimeError` directly. For example:
```ruby
raise "Cannot set 'default' on a credit card without a user"
# Versus
raise Spree::CreditCardError, "Cannot set 'default' on a credit card without a user"
```

```ruby
raise "Cannot set 'default' on a credit card without a user"
raise "The order association has been removed from InventoryUnit. The order is now determined from the shipment."
raise 'perform should be implemented in a sub-class of PromotionAction'
raise "Attempted to call code on a Spree::Promotion. Promotions are now tied to multiple code records"
raise "Attempted to call code= on a Spree::Promotion. Promotions are now tied to multiple code records"
raise "unexpected match policy: #{preferred_match_policy.inspect}"
raise "Implement me"
raise "Adjustment is already created"
raise "Adjustable does not match line item"
raise 'No master variant found to infer price'
raise "Unexpected originator type #{originator.class}"
raise "Your database needs at least one completed order to render this preview"
```

dustineichler:
@skukx LGTM. Probably a good idea to update Solidus guides with implementation and use details.

dhonig:
Looks good to me so far. Nice work!

skukx:
I think what I would do is consider this Merge Request complete as is. Then open up another merge request which will re-evaulate the error classes and raised runtime exceptions.

skukx:
@tvdeyen As the logger goes. I think it is an important feature to allow developers to decide where logs from `solidus` are sent. For the many that won't use this feature, this won't affect them. However for the few that may want to aggregate their logs, this is a simple and clean solution.
```ruby
config.logger = Logger.new('solidus.log')
```

As for the error reporter, I totally understand where you're coming from. I actually had coded this in the way you are speaking of. However I felt the solution fell short as far as extensibility. So extensibility was the reason I refactored it to be this way and allows extensions to be able to plugin. This also helps keep the code clean by keeping the different reporters separated.  If the community isn't in agreement on this I can switch it back. However I did feel the original solution was more complicated in the long run.

ericsaupe:
I love the `config.logger` option. 

The possibilities of the extensibility of the error reporting are really big. I can see the concern @tvdeyen has though. It does add complexity for something that may not be used by everyone. At least in my experience, we log everything to Rollbar and Filebeat and to a log file. This would make it easy to tie into all three of those services with little effort and as things change with new reporters/services it would be simple to reconfigure things. So for me I see the benefit but the question is whether or not it's worth it to the project at large. Like you said, it's there and if it's unused by most it shouldn't make a difference but gives another point of extensibility to developers.

skukx:
@tvdeyen As an owner and maintainer of the project you've got the final say on this. What direction would you like this to go? I'll make changes as needed

tvdeyen:
> I live the `config.logger` option.

I am fine with having logger separate from a error handler 👌

> At least in my experience, we log everything to Rollbar and Filebeat and to a log file. This would make it easy to tie into all three of those services with little effort and as things change with new reporters/services it would be simple to reconfigure things

I don't see why this is not handable by one error handler in your project. 

> As an owner and maintainer of the project you've got the final say on this. 

I don’t have. We need to have two approvals from core team members and if we agree on willing to maintain the multiple error handlers overhead, I am fine with this approach as well. 

Personally I would still say one error handler per store is enough, but YMMV and I am happily accepting examples and like to hear from other core team members what they think.

skukx:
>> At least in my experience, we log everything to Rollbar and Filebeat and to a log file. This would make it easy to tie into all three of those services with little effort and as things change with new reporters/services it would be simple to reconfigure things

> I don't see why this is not handable by one error handler in your project.

The problem is that from our own project we can't send errors which occur within solidus.
For example:

https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/api/app/controllers/spree/api/checkouts_controller.rb#L26-L29

So within this rescue say we want to log the error to some 3rd party reporter. In order to do so, we'd need to override that method.

The `ErrorReporter` class allows us to tie into that. Whether that is by inheritance or by using the `Composite` pattern.

skukx:
Example Usage:
Lets say we want to create a gem `solidus_rollbar` which reports solidus errors to rollbar.
```ruby
class SolidusRollbarReporter < Spree::Core::ErrorReporter::Base
  class << self
    def report(error, severity, metadata)
      Rollbar.send(rollbar_severity(severity), error, metadata)
    end

    private

    def rollbar_severity(severity)
      level = Logger.const_get(severity.upcase)
      [:debug, :info, :warning, :error, :critical, :error][level] || :error
    end
  end
end

# Register our new reporter
Spree::Core::ErrorReporter.add_reporter(SolidusRollbarReporter)
```

Now all stores need to do is add `solidus_rollbar` to their gemfile and logs are automatically sent.

As for filebeat, we do a very very custom implementation where we log with a json formatted string.
```ruby
class MyFilebeatReporter < Spree::Core::ErrorReporter::Base
  def self.report(error, severity, metadata)
    if error.respond_to?(:to_builder)
      Filebeat.logger.send(severity, error.target!)
    end
  end
end

Spree::Core::ErrorReporter.add_reporter(MyFilebeatReporter)
```

ericsaupe:
I wonder if simply fixing the way Solidus can sometimes swallow errors will remove the need to have custom error tracking configurations. If Solidus is handling errors like a standard Rails app these other services should be able to run without additional configuration and handle the errors appropriately.

skukx:
@ericsaupe That is not necessarily true. Solidus currently does pass a string rather than the exception itself into the logger. However even if that were changed it doesn't mean those errors will be reported to other services unless those services extend from the Rails logger. Not all services do that, and those that support it are not always configured to do so.

tvdeyen:
@skux most error trackers I know hook into the exception class and not into the Rails logger. 

I think we need to discuss the two things separately. Logging and exception tracking are two different things. 

@ericsaupe this is an interesting thought that we should discuss further.

skukx:
@tvdeyen I completely agree. I am more referencing rescued exceptions which won't bubble to those services. Some of those should still indeed be sent in my opinion

ericsaupe:
I guess the question then is, should we be rescuing? Should those methods/core be written in a way that allow those errors to bubble up?

skukx:
@ericsaupe I think rescuing is fine, in certain cases we may want to log the exception rescued though. An example would be, 
```ruby
def foo
  # Implementation
rescue GatewayError => error
  # Ok
  Rails.logger.error "Payment failed"

  # Better
  Rails.logger.error error
end
```

And if it is an exception we want to track in a 3rd party then we add that implementation in the rescue as well. This is because the error won't bubble up, nor do we want it to. This is of course is a minority of cases however.  Most rescues we don't care about tracking that exception rescued because technically we've already handled it gracefully.

tvdeyen:
@skukx you are right. Let's get back to topic.

I think having the ability to register several reporters is a actually good thing, the footprint is low but the possibilities are many. Thanks for the examples. 👌 

But I am still not convinced about the error reporter base class and repository. I think we can implement this with less code.

### 1. The Solidus default reporter

```rb
class Spree::ErrorReporter
  def report(error, severity, _metadata)
    Spree::Config.logger.send(severity, error)
  end
end

# Register the default reporter
Spree::AppConfiguration
  ...
  attr_writer :error_reporters
  def error_reporters
    @error_reporters ||= [Spree::ErrorReporter.new]
  end
end
```

### 2. A store having their own reporter

```rb
class MyStore::PutsReporter < Spree::ErrorReporter
  attr_reader :log_file

  def initialize(log_file)
    @log_file = log_file
  end

  def report(error, severity, metadata)
    log_file.puts error
  end
end

# Register the store reporter
Spree::Config.configure do |config|
  config.error_reporters << MyStorePutsReporter.new(File.open('log/errors.log'))
end
```

### 3. For an extension

```rb
class SolidusRollbarReporter < Spree::ErrorReporter
  def report(error, severity, metadata)
    Rollbar.send(rollbar_severity(severity), error, metadata)
  end

  private

  def rollbar_severity(severity)
    level = Logger.const_get(severity.upcase)
    [:debug, :info, :warning, :error, :critical, :error][level] || :error
  end
end

# Register the extension reporter
Spree::Config.configure do |config|
  config.error_reporters << SolidusRollbarReporter.new
end
```

### Several things to notice here. 

1. We should hold instances of reporters not classes.
2. We should not use the `Core` module namespace. This namespace is reserved for internal gem organisation not part of the public api.

We should split up this PR into two. One for the logger introduction, one for the error reporting. This helps keep focused on the topic :)

Again, super appreciate all the work going into this and I like to hear from other community members what they think about this.

🚀

skukx:
@tvdeyen Thanks for input. I like your implementation. There is one thing missing which is the execution of each reporter:
```ruby
def report(error, severity = :error, _metata = {})
  error_reporters.each do |reporter|
    reporter.report(error, severity, metadata)
  end
end
```

We could maybe do something like:
```ruby
module Spree::ErrorReporter
  def self.report(error, severity = :error, _metata = {})
    Spree::Config.error_reporters.each do |reporter|
      reporter.report(error, severity, metadata)
    end
  end
end

class Spree::ErrorReporter::Logger
  def report(error, severity, _metadata)
    Spree::Config.logger.send(severity, error)
  end
end

# Register the default reporter
Spree::AppConfiguration
  ...
  attr_writer :error_reporters
  def error_reporters
    @error_reporters ||= [Spree::ErrorReporter::Logger.new]
  end
end
```

However with this implementation we might as well create a base reporter to create a contract that all reporters must uphold (implemented `#report`).

core/lib/spree/core/error_reporter.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      ##
+      # Add a reporter
+      # @param reporter_class [ErrorReporter::Base]
+      # @return [Array<ErrorReporter::Base>]
+      #
+      def add_reporter(reporter_class)
+        unless reporter_class < Spree::Core::ErrorReporter::Base
+          raise ArgumentError, 'Reporter is not type Spree::Core::ErrorReporter::Base'
+        end
```
BenMorganIO: Why require it to be of `ErrorReport::Base`? I would recommend just checking if it `responds_to? :report`. That would be much more in line with Ruby philosophy.

core/lib/spree/core/error_reporter/base.rb
```diff
@@ -0,0 +1,34 @@
...
+      class Base
+        class << self
+          ##
+          # Generic error handling
+          # @param error [StandardError] The error you want to handle.
+          # @param serverity [Symbol, String] The severity (i.e. debug, info, warn, error, fatal)
+          # @param metadata [Hash] Any metadata we want to pass in
+          #
+          def report(_error, _severity = :error, _metadata = {})
+            raise NoMethodError
```
BenMorganIO: Might be better to have it `raise NotImplementedError`

BenMorganIO: As a developer, if I see an error that says `NotImplementedError`, that really tells me something :p

But if you see `NoMethodError` I would check to see if I spelled the method right. `NoMethodError` covers a wider array of reasons why it gets raised.

If I see someone do: `rescue NoMethodError` I would wonder why. Are they getting `nil` and `rescue`'ing if it ain't there? So many questions with `NoMethodError`.

The purpose of `NotImplementedError` was to communicate what needs to be implemented. You've spelled your method right and if you rescue it, it's obvious to give a default response when a method isn't implemented or to raise a new default error that is better formatted to the appropriate context.

BenMorganIO: Ah, just read the link. This puts a spin on things: http://ruby-doc.org/core-2.5.1/NotImplementedError.html

> Raised when a feature is not implemented on the current platform. For example, methods depending on the fsync or fork system calls may raise this exception if the underlying operating system or Ruby runtime does not support them.
>
> Note that if fork raises a NotImplementedError, then respond_to?(:fork) returns false.

kennyadsl:
@skukx Hey there, we discussed this old PR in the core team meeting today and we think there's still a lot of value here. @aldesantis did have an interesting idea that we wanted to share with you: maybe using events here can be beneficial to avoid building an error reporting system. We could just fire an event for each exception raised in core and ship a super basic event listener that just logs them as it's currently doing. Other apps can build their own event listener based on the reporting system they are using. What do you think?

skukx:
Yeah, I think using the event pipeline would be a perfect use of this

aldesantis:
I also think the key here is to come up with a naming scheme that allows people to easily subscribe to all interesting events via a regex (es. `info.*`, `error.*` etc., maybe with an additional namespace like `info.order.*`).

skukx:
I curious what you intend to send through the event bus? Will these be error objects? Uncaught errors only? Message strings? Caught errors etc?

It may be more robust to decorate the logger much like `ActiveSupport::TaggedLogging` does.

```
logger = Logger.new(STDOUT)
logger = ActiveSupport::TaggedLogging(logger)
```

But

```ruby
logger = Logger.new(STDOUT)
logger = Spree::Logging.new(logger)

# Assuming extensions have been written for this purpose

logger.add_plugin(rollbar)
logger.add_plugin(datadog)

# OR

Spree::Logger.configure do |config|
  plugin :rollbar
  plugin :datadog
end
```

Only the internals change

aldesantis:
@skukx my idea here was to pass caught exceptions to the event bus. Whenever Solidus swallows an exception that may be interesting for the user to track, we emit an event so that they can decide what they want to do with that exception (log it, report it etc.)

Decorating the logger provides a more standardized API I guess, but at the expense of writing more code. At the end of the day, what we're doing is simply notifying the application that something of interest happened, which seems like a good job for the event bus.

PR #2818, opened 2018-08-02
Fix of typo of E-commerce

johnlow95:


jacobherrington:
I think this should stay all lowercase. That's purely opinion though.

kennyadsl:
Closing, @benjaminwil has the final word here 🙂 . Thanks anyway!

PR #2823, opened 2018-08-07
Backend: Extra cancancan validations for some URL resource links

memotoro:
Hello

In the backend system, some resource link are accesible without any cancancan validations. These extra simple conditionals will help to avoid a leak on those resources for users without the right roles and permissions sets. These changes simply avoid rendering the "create new resource" links if the user has no permissions to do so, given a cleaner backend user experience.

Regards

### DIFF
--- a/backend/app/controllers/spree/admin/zones_controller.rb
+++ b/backend/app/controllers/spree/admin/zones_controller.rb
@@ -20,7 +20,7 @@ module Spree
 
       def load_data
         @countries = Spree::Country.order(:name)
-        @states = Spree::State.order(:name)
+        @states = Spree::State.includes(:country).order(:name)
         @zones = Spree::Zone.order(:name)
       end
     end


backend/app/views/spree/admin/images/index.html.erb
```diff
@@ -13,6 +13,7 @@
   <%= render 'new', product: @product, image: Spree::Image.new(viewable: @product) %>
 </div>
 
+<% if can?(:create, Spree::Image) %>
```
kennyadsl: The inner content needs an indentation here

backend/app/views/spree/admin/orders/index.html.erb
```diff
@@ -7,7 +7,7 @@
   <li>
     <%= link_to t('spree.new_order'), new_admin_order_url, id: 'admin_new_order', class: 'btn btn-primary' %>
   </li>
-<% end if can? :create, Spree::Order %>
+<% end if can? :manage, Spree::Order %>
```
kennyadsl: why ``manage` if the link is for creating orders only?

backend/app/views/spree/admin/payments/index.html.erb
```diff
@@ -3,7 +3,9 @@
 <% content_for :page_actions do %>
   <% if @order.outstanding_balance? %>
     <li id="new_payment_section">
-      <%= link_to t('spree.new_payment'), new_admin_order_payment_url(@order), class: 'btn btn-primary' %>
+      <% if can? :manage, Spree::Payment %>
```
kennyadsl: why `manage` and not `create`?

backend/app/views/spree/admin/return_authorizations/index.html.erb
```diff
@@ -3,7 +3,7 @@
 <% content_for :page_actions do %>
   <% if @order.shipments.any? &:shipped? %>
     <li>
-      <% if can? :create, Spree::ReturnAuthorization %>
+      <% if can? :manage, Spree::ReturnAuthorization %>
```
kennyadsl: same question as above

memotoro:
Hello @kennyadsl 

Thanks for you feedback. Sorry, I need to check the specs in order to pass the tests and I might require more work to make the changes work properly. Could you drop completely this pull request and I'll create a new one with the right code later once all the issues are fixed?

Regards

backend/app/views/spree/admin/payments/index.html.erb
memotoro: @kennyadsl You are right here. Create should be enough. I'll fix it.

backend/app/views/spree/admin/orders/index.html.erb
memotoro: Hello @kennyadsl . 

There are two classes to manage the permissions sets Spree::PermissionSets::OrderManagement and Spree::PermissionSets::OrderDisplay. The OrderManagement establishes the ':manage' permissions. The display establishes the ':display' and ':admin' permissions. If I want to restrict a user with only the permission OrderDisplay, I have to check for ':manage' permission as for some reasons, the ':create' permission on OrderDisplay is still allowing the button to be rendered, allowing the user to click the New Order button, and creating an empty order that it cannot be modified, not great user experience I guess. I'm not sure why the ':create' is still available on OrderDisplay. Do you know how to fix this issue? Maybe I misunderstood the OrderDisplay role. Let me know your thoughts about it.

backend/app/views/spree/admin/images/index.html.erb
memotoro: @kennyadsl . You're right. I'll fix it.

backend/app/views/spree/admin/return_authorizations/index.html.erb
memotoro: @kennyadsl Same comment with the OrderDisplay issue mentioned above.

backend/app/views/spree/admin/orders/index.html.erb
memotoro: Hello @kennyadsl 

Do you have any thoughts about my previous comments on PermissionSets? 

Regards

kennyadsl: By default only [this permissions set](https://github.com/solidusio/solidus/blob/master/core/lib/spree/permission_sets/default_customer.rb) is activated. I think that at line 10 it sets the create ability for Order. Are you setting the other permissions for a specific user in some way via code?

tvdeyen:
I think `:manage` is fine for orders and returns. It only makes sense to be able to create an order if you are also able to update it.

Please fix the specs. Thanks for the contribution

memotoro:
@kennyadsl @tvdeyen . Changes as requested. Let me know your thoughts. Regards

ericsaupe:
Looks good. Would you mind squashing your commits?

memotoro:
Hello @ericsaupe @tvdeyen @kennyadsl 

Sorry for the late replay. I had issues merging last time that I've squashed my commits to make it a single commit. Have a look and it is passing the CI tests.

Regards

backend/app/views/spree/admin/orders/index.html.erb
aitbw: From what I can understand from @memotoro's comment and the proposed patch is that, `OrderDisplay` works **only** to create orders without any further actions, whereas `OrderManagement` allows full control over them. Since we're only talking about `backend`, this feels reasonable.

As @tvdeyen said on his comment below, if you can create an order, you probably should be able to update it as well.

memotoro:
Hello @kennyadsl @aitbw . Rebase the commit against the current master. Specs were added before to test the feature. Regards

kennyadsl:
Thanks @memotoro!

> Specs were added before to test the feature.

What do you mean?

memotoro:
Hello @kennyadsl. What I meant is that I already changed a spec to test this feature, and I did it before the rebase that I just did. This commit is only a rebase against the current master, so it will pass the test cases now. Sorry for the misunderstanding. The spec I modified is `backend/spec/features/admin/orders/listing_spec.rb`. Hopefully it gets merge into the next release. Regards

spaghetticode:
@memotoro nice work! 🎉 

I was wondering if there's some plan of adding the missing integration specs. I also see there are some unanswered questions left by @aitbw, I think we're pretty close to having this finally approved, it just needs a bit more TLC 🔧⚒... thank you!

kennyadsl:
Hey @memotoro! Is there any chance for you to complete the PR. I think the last missing piece is [this Angel concern](https://github.com/solidusio/solidus/pull/2823#discussion_r267738858), and are good to go. Thanks!

spaghetticode:
@memotoro I'll be pushing this PR forward, thanks for the contribution!

Issue #2825, opened 2018-08-08
Rails 5.2.1 seems to be breaking Sandbox migrations


jacobherrington:
### Steps to reproduce
From Solidus `bundle exec rake sandbox`

### Expected behavior
A new instance of Sandbox

### Actual behavior
The [remove_order_id_from_inventory_units](https://github.com/solidusio/solidus/blob/425240b84f670558aab61df2655be9a75e00dda0/core/db/migrate/20170319191942_remove_order_id_from_inventory_units.rb#L22) migration fails.

<img width="1008" alt="screen shot 2018-08-08 at 11 22 30 am" src="https://user-images.githubusercontent.com/11466782/43850372-611c71ac-9afd-11e8-8d6f-18b20ef10b6c.png">

Thanks to @VzqzAc for pointing out that the issue originates from https://github.com/activerecord-hackery/polyamorous

It seems we need to set Rails to 5.2.0 to get the migrations working.

### System configuration
**Solidus Version**: 2.7.0.alpha

jacobherrington:
Related: https://github.com/rails/rails/issues/33557

VzqzAc:
Issued on polyamorous/Related?: https://github.com/activerecord-hackery/polyamorous/issues/45

Also seems like any migration with `joins` will fail, not only `remove_order_id_from_inventory_units`

pedrofurtado:
@VzqzAc @jacobherrington The ActiveRecord 5.2.1 changed the number of parameters used by some classes in Private API. Because of this, some gems from activerecord-hackery (ransack, polyamorous, etc.) is broken.

kennyadsl:
Ransack released a new gem version today ([1.8.9](https://rubygems.org/gems/ransack/versions/1.8.9)). Unfortunately they decided (still not sure if that was intentional) to lock Rails compatibility up to 5.1.1. I opened a [new issue on Ransack](https://github.com/activerecord-hackery/ransack/issues/943) to understand better how to go head.

kennyadsl:
To be more clear, this is making impossible for us to support Rails 5.2.1 since ransack 1.8.8 is generating the error above and 1.8.8 does not support Rails 5.2 at all.

kennyadsl:
Working on a PR (#2826).

docelic:
@eduardopatrick to just use Rails 5.2.0 and have it all working with Solidus, there's just a small change you need to do:

[...]

PR #2826, opened 2018-08-09
Compatibility with Rails 5.2.1 & Ransack

kennyadsl:
ref #2825 

- Update ransack dependency to make it work with Rails 5.2.1
- Fix some specs
- Lock monetize dependency since the new version does not work with our DistributedAmount calculator.

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -35,7 +35,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'discard', '~> 1.0'
   s.add_dependency 'friendly_id', '~> 5.0'
   s.add_dependency 'kaminari-activerecord', '~> 1.1'
-  s.add_dependency 'monetize', '~> 1.1'
+  s.add_dependency 'monetize', '~> 1.8.0'
   s.add_dependency 'paperclip', ['>= 4.2', '< 6']
   s.add_dependency 'paranoia', '~> 2.4'
   s.add_dependency 'ransack', '~> 2.0'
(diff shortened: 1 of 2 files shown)


kennyadsl:
[This is the issue](https://github.com/activerecord-hackery/ransack/issues/947) on ransack that is making our spec fail, let's see what happen. 

In the meantime ransack 2.0 has been released so I'm going to bump our dependency up in the core gemspec. This way Rails < 5.2 will use ransack 1.8 and Rails >= 5.2 will use ransack 2.0.

huoxito:
@kennyadsl see https://github.com/nebulab/solidus/pull/15 that might help

that last spec failure might be related to a monetize version change. I noticed that latest [master build](https://circleci.com/gh/solidusio/solidus/8621) was still _Using activerecord 5.1.6_ I tried to run the test on that version but it still failed. When I diff my local bundle with the one in circleci a monetize gem bump made me think it could be related.

I don't know yet whether it's a problem with the test setup or something wrong with monetize. It seems worth checking though, calculations shouldn't break like that.

core/solidus_core.gemspec
```diff
@@ -35,9 +35,9 @@ Gem::Specification.new do |s|
   s.add_dependency 'discard', '~> 1.0'
   s.add_dependency 'friendly_id', '~> 5.0'
   s.add_dependency 'kaminari-activerecord', '~> 1.1'
-  s.add_dependency 'monetize', '~> 1.1'
+  s.add_dependency 'monetize', '~> 1.8.0'
```
skukx: Minor suggestion would be to add a comment as to why this is locked with a link to the issue that needs to be resolved until it can be relaxed again.

kennyadsl: Thanks for the suggestion, I've documented that in the commit message. I prefer not to add comments in the code and if we forgot and keep monetize locked to 1.8 is not that critical. If someone will need to update monetize to 1.9 a `git blame` will document the why of this. Does it make sense?

PR #2829, opened 2018-08-15
Tracking Number link to Tracking URL page

JuanCrg90:
This change allows showing the tracking URL in the admin panel instead of only the tracking number.

### DIFF
--- a/backend/app/views/spree/admin/orders/_carton.html.erb
+++ b/backend/app/views/spree/admin/orders/_carton.html.erb
@@ -50,7 +50,12 @@
       <tr class="show-tracking total">
         <td colspan="5">
           <% if carton.tracking.present? %>
-            <strong><%= Spree::Carton.human_attribute_name(:tracking) %>:</strong> <%= carton.tracking %>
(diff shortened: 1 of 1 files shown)


backend/app/views/spree/admin/orders/_carton.html.erb
```diff
@@ -50,7 +50,12 @@
       <tr class="show-tracking total">
         <td colspan="5">
           <% if carton.tracking.present? %>
-            <strong><%= Spree::Carton.human_attribute_name(:tracking) %>:</strong> <%= carton.tracking %>
+            <strong><%= Spree::Carton.human_attribute_name(:tracking) %>:</strong>
+            <% if carton.tracking_url.nil? %>
```
jarednorman: I'd probably just write this as
```ruby
<% if carton.tracking_url %>
  <%= link_to carton.tracking, carton.tracking_url, target: "_blank" %>
<% else %>
  <%= carton.tracking %>
<% end %>
```
but either way this seems fine.

JuanCrg90: Updated, Thank you 😄

JuanCrg90:
😄 Thanks

PR #2830, opened 2018-08-17
Make Spree::Image more extensible

fastjames:
We have some special cases where we need to modify how `Spree::Image`
objects behave, so we need to override a few of the parameters in the image class.


fastjames:
The extension point for styles is unneeded and breaks some parts of the admin. Back to the drawing board.

PR #2831, opened 2018-08-20
Move factory_bot static attrs to dynamic

fastjames:
Factory_bot is now flagging static factory attributes as deprecated,
pending removal in version 5.0. This is adding a lot of noise to test output, which makes it harder to see more important failures.


tvdeyen:
@fastjames a rebase with latest master should fix the build errors

PR #2835, opened 2018-08-28
@wip

fastjames:
Change FactoryGirl to FactoryBot

We need to backport this change to maintain the branch, since `factory_girl` is no longer available as a gem.

I don't know how much interest there is in backporting changes like this, but I figured it would be useful since this branch is still under maintenance until November or so.


kennyadsl:
Personally, I'm against changing this in old versions branch/es. I know they should be maintained still for some time but the maintenance is more for security updates, or at least bugfix. 

Let's wait for some other opinions on this, thanks anyway.

fastjames:
I would prefer to leave it as-is as well, so maybe I need to lock it to a pre-name change version? That would be a less invasive change. My goal here is to help extension authors who run CI against older versions. I had trouble getting versions with `factory_girl` to pass so I might have been doing something wrong.

kennyadsl:
What about locking `factory_bot` gem to 4.10.0 in solidus 2.3 instead? It should still have the factory_girl file in `/lib` which [has been removed with this commit](https://github.com/thoughtbot/factory_bot/commit/40eed88bb9222ee09001fbc7d301527836e0d085). What do you think?

BenMorganIO:
Are we able to kill using `FactoryBot` explicitly in specs? Using the monkey patched version of `create` and `build` are far more succinct, use less code, and uses `FactoryBot` as it was originally meant to be used.

BenMorganIO:
@kennyadsl the community is providing a valid change to help with backward compatibility. Solidus team members are committed to security patches for older versions. If a community member provides a fix, especially one that may be holding them back, I don't see why the Solidus team cannot merge it if the work has already been paid for.

tvdeyen:
We have a policy of very stable branches. Even small changes may have impact on stores. Since this change does not affect the stability I am fine with it, but I am not sure why this change is needed. Do specs fail without that change? The factory_girl is still available. It hasn’t been removed from rubygems AFAIK. Maybe we need to pin the specific version.

kennyadsl:
@tvdeyen specs fail because [they removed the `FactoryGirl`](https://github.com/thoughtbot/factory_bot/commit/40eed88bb9222ee09001fbc7d301527836e0d085) namespace at all in the latest released versions.

@BenMorganIO I wasn't aware that this was breaking extension specs on older Solidus versions when I said I was against this PR. Now that this is clear I'm open to merge this change, even if we should make the same thing for the other older versions maybe?

tvdeyen:
> @tvdeyen specs fail because they removed the FactoryGirl namespace at all in the latest released versions.

@kennyadsl but not if we pin the version of `factory_girl` to `4.8.1`, right? That's the last version that does not have all the deprecation notices in it. This should work well and is the least intrusive change for this branch IMO.

@fastjames would you please revert all the `FactoryGirl` -> `FactoryBot` changes and just lock the `factory_girl` version? We should be fine then.

Thanks for the contribution

kennyadsl:
@tvdeyen I think we could even switch to factory_bot and just use `4.10.0` version, which is the last that contains the old `FactoryGirl` namespace.

tvdeyen:
@kennyadsl This only produces a lot of deprecation noise, no? This might confuse people and make them think they should upgrade. Are there any notable fixes in 4.10 that we need?

I am not opposed to it, just think we should reduce deprecation notices to a minimum.

kennyadsl:
@tvdeyen Yes, I was just afraid that `factory_girl` gem one day could be deleted but honestly I don't even know if it's doable and for sure when this happens those Solidus versions will have passed their EOL. 

👍 for your proposal

fastjames:
Glad I could generate some discussion! To answer an earlier question from @tvdeyen I came to this because I had an extension whose specs were failing, and the errors seemed to point to the uphill `factory_girl` dependency as the culprit (because it was not pinned). With time comes understanding, and I would be happy to pin to either of the mentioned versions. If 4.10.0 creates deprecation messages and we don't want to fix them, then I would prefer to pin to 4.8.1. I am pushing a new commit with that in it now.

BenMorganIO:
@fastjames are you trying to use factory bot in your current project?

BenMorganIO:
If so, you may be able to cheat with this initializer:

```ruby
# config/initializers/factory_bot.rb

FactoryBot = FactoryGirl
```

And then as you bump Solidus or try and add factory bot with factory girl (if that's possible :/) it should make the journey somewhat easier on you.

fastjames:
I can reproduce the poltergeist timeout error, but not the other one; it seems to be related to the time being taken to compile assets.

BenMorganIO:
@fastjames yeah, I've ran into that issue before.

tvdeyen:
@kennyadsl what happens if you lock the factory girl version in Solidus Auth Devise? Do they still fail?

tvdeyen:
> I'm afraid my suggestion was incorrect and locking factory_girl dependency in common_spree_dependencies.rb is not a solution since they are dependency needed for the solidus gem development and they are not considered as extensions dependencies.

I think this change is still correct. If someone wants to build Solidus 2.3 we want the tests to pass. 

For extensions though we need to make sure that the extension is using factory girl 4.8.1 if they run specs for Solidus 2.3.

tvdeyen:
> Yes but currently almost all extensions are built to work with all solidus versions, adding some extra code or branch in all extension is worst than change Factory Girl to Factory Bot in 2.2, 2.3, 2.4 branches here IMHO.

Extensions already have code that distinguish Solidus versions.

See `solidus_auth_devise` as an example

1. https://github.com/solidusio/solidus_auth_devise/blob/master/Gemfile
2. https://github.com/solidusio/solidus_auth_devise/search?q=solidus_gem_version&unscoped_q=solidus_gem_version

The change extensions would need to make is not that hard actually.

In `solidus_auth_devise` it would be to add one line in the `Gemfile`

```rb
# Gemfile
...
group :test do
  ...
  if branch < "v2.5"
    gem 'factory_girl', '4.8.2' # has both `factory_girl` and `factory_bot` in it.
  else
    gem 'factory_bot', '> 4.8.2'
  end
  ...
end
```

and remove the `factory_bot` dependency from the gemspec.

We can also update the `solidus_cmd` `Gemfile` template, as Solidus 2.4 (the last version that still uses `factory_girl`) reaches EOL in May 2019 and we need to support it until then.

tvdeyen:
> In terms of the danger of doing that, I think it's just matter of specs so there could not be issues. If stores that still uses Factory Girl (and are importing solidus factories into their specs) updates to latest Solidus patch version (like 2.3.1) they will be forced to update to Factory Bot, but is that a real issue?

It think we should avoid breaking the builds of people upgrading a minor (or patch version) of Solidus, if we can.

tvdeyen:
@kennyadsl I will prepare a PR for `solidus_auth_devise`

kennyadsl:
@tvdeyen ok thanks, let's see how it comes 🙂

tvdeyen:
https://github.com/solidusio/solidus_auth_devise/pull/130

tvdeyen:
## Actually the fix is even easier

Just use `factory_bot`, '4.10.0' for all Solidus versions. What FactoryBot feature of < 4.10 do we desperately need that justifies all the hassle?

jacobherrington:
It should be mentioned in [writing extensions](https://github.com/solidusio/solidus/blob/master/guides/source/developers/extensions/writing-extensions.html.md) as well.

@tvdeyen what about someone who works with a store that is on an older Solidus? They might want to build an extension and would care to support the old versions.

tvdeyen:
> It should be mentioned in writing extensions as well.

Yes, do it. But keep in mind that Solidus 2.4 will reach EOL in 6 months anyway 🤷‍♂️ 

> @tvdeyen what about someone who works with a store that is on an older Solidus? They might want to build an extension and would care to support the old versions.

Easy. Add `factory_bot` 4.10.0 to your `Gemfile`/`gemspec`. Done

kennyadsl:
Cough, cough https://github.com/solidusio/solidus/pull/2835#issuecomment-418301511 🤣 I'm happy if we all agree that this is a good solution. 

https://github.com/solidusio/solidus_auth_devise/pull/130 is not needed anymore, right?

tvdeyen:
> solidusio/solidus_auth_devise#130 is not needed anymore, right?

It is. Without that change it would load newer versions of `factory_bot` even for old versions of Solidus.

kennyadsl:
Oh right, I keep forgetting that the Solidus dev dependencies are not reflected into extensions. 👍 

RE "the plan":

Right now:

- update all extensions with something similar to https://github.com/solidusio/solidus_auth_devise/pull/130/commits/d902b8ea46883f51f31df6fd59cda7b805ab0849
- merge this PR
- update template on  and documentation to explain why this is needed (I'm working on this)

Once the EOL of 2.4 is reached (2019-05-07):

- just revert https://github.com/solidusio/solidus_auth_devise/pull/130/commits/d902b8ea46883f51f31df6fd59cda7b805ab0849 on all extensions

tvdeyen:
This PR is not needed anymore, right?

PR #2836, opened 2018-09-01
Money#allocate calculates weights already

huoxito:
Related to https://github.com/solidusio/solidus/pull/2826/commits/3f027e24805f146507d032404a8c727d815d778a. Assuming build is still green for both monetize 1.8 and 1.9 we can loose that dependency again.

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -35,7 +35,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'discard', '~> 1.0'
   s.add_dependency 'friendly_id', '~> 5.0'
   s.add_dependency 'kaminari-activerecord', '~> 1.1'
-  s.add_dependency 'monetize', '~> 1.8.0'
+  s.add_dependency 'monetize', '~> 1.9'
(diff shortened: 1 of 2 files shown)


huoxito:
also related https://github.com/RubyMoney/monetize/issues/118#issuecomment-416904973

core/solidus_core.gemspec
```diff
@@ -35,7 +35,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'discard', '~> 1.0'
   s.add_dependency 'friendly_id', '~> 5.0'
   s.add_dependency 'kaminari-activerecord', '~> 1.1'
-  s.add_dependency 'monetize', '~> 1.8.0'
+  s.add_dependency 'monetize', '~> 1.9'
```
tvdeyen: 1.8 does not work?

tvdeyen: Yes. Please. We want to be as conservative as possible with dependencies

huoxito:
I've set monetize back to `~> 1.8` thanks

PR #2838, opened 2018-09-04
Update guides ffi gem

kennyadsl:
This gem version had a security issue:

https://nvd.nist.gov/vuln/detail/CVE-2018-1000201


kennyadsl:
Merging with 1 review. This is on guides and only on development, it's not critical

PR #2840, opened 2018-09-07
allow multiple taxons on product creation

jacobherrington:
add private methods for splitting up taxons and option_type_ids on the products controller, addressing a bug that only allowed for one of each to be applied on product creation

refs #2791

### DIFF
--- a/backend/spec/controllers/spree/admin/products_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/products_controller_spec.rb
@@ -52,18 +52,21 @@ describe Spree::Admin::ProductsController, type: :controller do
 
   # regression test for https://github.com/spree/spree/issues/2791
   context "adding taxons to a product" do
-    let!(:product) { create(:product) }
-    let!(:first_taxon) { create(:taxon) }
-    let!(:second_taxon) { create(:taxon) }
+    let(:product) { create(:product) }
+    let(:first_taxon) { create(:taxon) }
 
     it "adds a single taxon to a product" do
       put :update, params: { id: product.to_param, product: { taxon_ids: first_taxon.id.to_s } }
       expect(flash[:success]).to eq("Product #{product.name.inspect} has been successfully updated!")
     end
 
-    it "adds multiple taxons to a product" do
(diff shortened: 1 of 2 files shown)


backend/spec/controllers/spree/admin/products_controller_spec.rb
```diff
@@ -50,6 +50,23 @@
     end
   end
 
+  # regression test for https://github.com/spree/spree/issues/2791
+  context "adding taxons to a product" do
+    let!(:product) { create(:product) }
+    let!(:first_taxon) { create(:taxon) }
+    let!(:second_taxon) { create(:taxon) }
```
BenMorganIO: This taxon is needlessly created for the first spec. `let!` can also be problematic in creating records when you don't need them and can slow test suites down. Using `let` will solve this.

jacobherrington: Thanks for the feedback @BenMorganIO, changes made. Let me know if there is anything else you can think of!

backend/spec/controllers/spree/admin/products_controller_spec.rb
```diff
@@ -50,6 +50,26 @@
...
+
+    context "when there are mulitple taxons" do
+      let(:second_taxon) { create(:taxon) }
+
+      it "adds multiple taxons to a product" do
+        put :update, params: { id: product.to_param, product: { taxon_ids: "#{first_taxon.id}, #{second_taxon.id}" } }
+        expect(flash[:success]).to eq("Product #{product.name.inspect} has been successfully updated!")
+      end
+    end
+  end
```
tvdeyen: You are testing product update, while the PR says product creation. 

Could you add a spec for creation as well?

BenMorganIO: Yep, especially since the before action also applies to create. The code change did expand the behavior to create.

jacobherrington: Thanks for the feedback, just pushed some more tests!

PR #2843, opened 2018-09-12
Fix Solidus v2.6 and prepare a 2.6.1 release

kennyadsl:
It basically applies the fixes we've done on master (#2826 and #2836), which should be good for Solidus 2.6 as well.

Thanks @huoxito

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -35,7 +35,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'discard', '~> 1.0'
   s.add_dependency 'friendly_id', '~> 5.0'
   s.add_dependency 'kaminari-activerecord', '~> 1.1'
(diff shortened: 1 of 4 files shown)


core/solidus_core.gemspec
```diff
@@ -35,9 +35,9 @@ Gem::Specification.new do |s|
   s.add_dependency 'discard', '~> 1.0'
   s.add_dependency 'friendly_id', '~> 5.0'
   s.add_dependency 'kaminari-activerecord', '~> 1.1'
-  s.add_dependency 'monetize', '~> 1.1'
+  s.add_dependency 'monetize', '~> 1.8.0'
```
huoxito: think this could be `~> 1.8` since it works with both 1.8 and 1.9 versions

kennyadsl: you are right, fixed!

PR #2847, opened 2018-09-17
standardize JS semi-colon style in backend

jacobherrington:
This is a pretty low priority PR. I noticed that semi-colon usage was pretty erratic in the JS. It seemed that generally it was preferable to end statements with a semi-colon, so I went ahead and tried to standardize that across all of the JS.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -19,8 +19,8 @@ Spree.ready(function() {
   // Highlight hovered table column
   $('table').on("mouseenter", 'td.actions a, td.actions button', function(){
     var tr = $(this).closest('tr');
-    var klass = 'highlight action-' + $(this).data('action')
-    tr.addClass(klass)
+    var klass = 'highlight action-' + $(this).data('action');
+    tr.addClass(klass);
 
     var observer = new MutationObserver(function(mutations) {
       tr.removeClass(klass);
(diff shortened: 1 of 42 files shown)


tvdeyen:
Semicolons are totally optional in Javascript, that’s why I think we should not change a bunch of source files. I does not provide any benefit but adding noise to git blame. And that alone is reason enough to reject this change. 

But, I think we should think about adding JS linting to our workflow. But this is a whole different story. 

For the scope if this PR I think we should not do it.

jacobherrington:
@tvdeyen That's fair, totally agree with turning this down in favor of adopting some kind of linter. I feel like there are a lot of opportunities to increase our code quality, particularly in JS.

jacobherrington:
Opened #2849 in response to this feedback. Would love to hear some opinions there so that we can do something about JS code quality. Thanks! 👍

ericsaupe:
@tvdeyen how is this different than the `rubocop -a` git changes? This is simply a manual lint change. I can see not doing it in favor of a proper solution for #2849 which would be just like the `rubocop -a` changes but I think code clean up, even in small commits, is valuable.

tvdeyen:
@ericsaupe the difference is that we proposing a change here that would preferable be disabled in a linter (because semicolons are optional even in JS) and we did not in the past and should not continue to care about them. Being too strict with linters can harm contributions. We want only the bare minimum of style linting as possible. The same reasons why we have lots of Rubcops disabled. 

IMO We should:

1. Set up JS linting
2. Disable all checks
3. Enable one check
4. Fix the issues
5. Repeat from 3

That's how we did it with Rubocop and this is how we should do this with JS linting.

ericsaupe:
Should we close this PR in favor of the solution to #2849?

jacobherrington:
Sounds good to me @ericsaupe

PR #2853, opened 2018-09-21
 Fix a bug with the settings subnav 

jacobherrington:
There is currently a UI bug causing the settings subnav to display incorrectly on some routes.

`/admin/return_reasons` and `/admin/adjustment_reasons` fail to display the subnav at all:
image

While `/admin/reimbursement_types` doesn't highlight the active tab:
image

This PR fixes both issues. 

I also want to verify that removing `:return_authorization_reasons` from line 18 of `backend_configuration.rb` is okay. If that is used anywhere, we can put it back.

### DIFF
--- a/backend/app/models/spree/backend_configuration.rb
+++ b/backend/app/models/spree/backend_configuration.rb
@@ -16,8 +16,7 @@ module Spree
                             :payment_methods, :shipping_methods,
                             :shipping_categories, :stock_locations,
                             :refund_reasons, :reimbursement_types,
-                            :return_authorization_reasons, :return_reasons,
-                            :adjustment_reasons]
+                            :return_reasons, :adjustment_reasons]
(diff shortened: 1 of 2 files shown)


backend/app/models/spree/backend_configuration.rb
```diff
@@ -15,7 +15,8 @@ class BackendConfiguration < Preferences::Configuration
                             :tax_rates, :zones,
                             :payment_methods, :shipping_methods,
                             :shipping_categories, :stock_locations,
-                            :refund_reasons, :reimbursement_types, :return_authorization_reasons]
+                            :refund_reasons, :reimbursement_types,
+                            :return_reasons, :adjustment_reasons]
```
BenMorganIO: uh oh, I think you meant to do `:return_authorization_reasons`.

PR #2855, opened 2018-09-21
Stop negative totals during checkout

### LINKED ISSUE #2741: Negative order totals

jacobherrington:
Currently it is possible to check out with a negative total. 

image

This PR stops the transition to the payment step if the cart total is below zero.

Would love some feedback on this as I'm sure there are other solutions.

Closes #2741

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -862,6 +862,12 @@ module Spree
       end
     end
 
+    def ensure_positive_total
+      if outstanding_balance < 0
+        errors.add(:base, I18n.t('spree.cart_balance_negative')) && (return false)
+      end
+    end
+
     def ensure_available_shipping_rates
       if shipments.empty? || shipments.any? { |shipment| shipment.shipping_rates.blank? }
         # After this point, order redirects back to 'address' state and asks user to pick a proper address
--- a/core/app/models/spree/order/checkout.rb
+++ b/core/app/models/spree/order/checkout.rb
@@ -83,6 +83,8 @@ module Spree
               after_transition to: :complete, do: :add_payment_sources_to_wallet
               before_transition to: :payment, do: :add_default_payment_from_wallet
 
+              before_transition to: :payment, do: :ensure_positive_total
+
               before_transition to: :confirm, do: :add_store_credit_payments
 
               # see also process_payments_before_complete below which needs to
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1085,7 +1085,8 @@ en:
       one: Subtotal (1 item)
       other: Subtotal (%{count} items)
     carton_external_number: External Number
-    carton_orders: Other orders with Carton
+    carton_orders: 'Other orders with Carton'
+    cart_balance_negative: "Cart balance cannot be negative"
     categories: Categories
     category: Category
     character_limit: Limit of 255 characters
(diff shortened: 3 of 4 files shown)


peterberkenbosch:
what are your thoughts about zero-ing out the order instead of rendering an error message?

jacobherrington:
@peterberkenbosch I think this was closer to @kennyadsl's solution. This allows shops that might want unique behavior to rescue the error. I have no strong feelings about it though!

peterberkenbosch:
After some more thought, I do think it’s actually a better way to raise an exception for something that should not happen in the first place. This leaves any "magic" zero value orders out too.

core/app/models/spree/order.rb
```diff
@@ -870,6 +870,12 @@ def ensure_line_items_present
       end
     end
 
+    def ensure_positive_total
+      if outstanding_balance < 0
+        errors.add(:base, I18n.t('spree.cart_balance_negative')) && (return false)
```
BenMorganIO: I don't think there's really any need to do a return of `false` here. The return value of this method will never be used and it's more for possibly adding errors to the record.

jacobherrington: Will happily remove it! Thanks for the feedback, appreciate it!

BenMorganIO: Weird. I would suspect AASM to just check the errors. It's scary then that AASM lets me move forward in steps _with errors_ if the method is~, as Java would put it,~ `void`.

peterberkenbosch: the last method is returning !false hence the state machine continues. It's not raising an error mind you. It's just adding a model validation. if we were to `raise` an `Exception` here we would have stopped the state machine.

jacobherrington: @peterberkenbosch do you think raising an exception is a better alternative?

peterberkenbosch: I'm not sure, I think added the model validation presents it cleanly to the end user. When raising an exception, when not caught somewhere, we end up with 500 screens. 

On the other hand, when there is some sort of error tracking in place, we do send a quick and pressing message that something is not going as expected when we do raise an exception.

jacobherrington: Because it is possible for a user to hit this (given the right circumstances), I think we should leave it with AR. Thanks for clearing up our confusion!

BenMorganIO: If raising an exception, would the state machine catch it and politely return `false` when doing `order.next` vs `order.next!`?

peterberkenbosch: `errors.add` is not returning `nil` it returns an Array with error messages

spaghetticode:
@jacobherrington 👏 thank you for this improvement! 

I understand now `order.next!` raises an error, but it's not clear to me how the store frontend/API are affected by this change. Maybe we could add a feature spec that verifies what happens in case of negative amount? 

Also, currently the PR needs some rebasing with master 🚧

jacobherrington:
I don't feel like this is a good solution, so I'm closing this PR

PR #2860, opened 2018-09-24
Fix a bug causing non-eligible adjustments to appear in the the admin…

### LINKED ISSUE #2484: Non Eligible Order Adjustments Showing in Admin Order Cart Page
### Steps to reproduce
 - Go To Admin
 - Create a promotion with name `50% Off` and exactly mimics the below settings
screen shot 2018-01-03 at 1 28 02 pm

- Go to Orders
- Create a new Order
- Add an item that the promotion won't apply to

### Expected behavior
Under Order Adjustments, The `50% Off` promotion should not appear since it isn't eligible for the order.

### Actual behavior
The `50% Off` promotion appears under Order Adjustments even though eligible is false.
screen shot 2018-01-03 at 1 33 16 pm

### System configuration
Solidus 2.3

jacobherrington:
… cart

Fixes #2484.

ES6 would make this significantly cleaner, but it seems we favor ES5 for now (happy to make that change though!)

I don't have a ton of experience with Backbone.js, so if there is a more performant built-in solution for this I'd love to hear about it.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/order/details_adjustments.js
+++ b/backend/app/assets/javascripts/spree/backend/views/order/details_adjustments.js
@@ -8,7 +8,14 @@ Spree.Views.Order.DetailsAdjustments = Backbone.View.extend({
     var totals = {};
     var collection = this.collection ? this.collection.chain() : _.chain([this.model]);
     collection
-      .map(function(item){ return item.get("adjustments") || [] })
+      .map(function(item) {
+        return (item.get("adjustments") || [])
+        .filter(function(adjustment) {
+          if (adjustment.eligible === true) {
+            return adjustment;
+          }
+        });
+      })
       .flatten(true)
       .each(function(adjustment){
         var label = adjustment.label;


jacobherrington:
@kennyadsl essentially the frontend is displaying all adjustments including those which are ineligible on the order. They are not actually being applied to the order, just displayed on the page this JS effects. 

Filtering them by eligibility causes this function to return an empty array which is what should happen when there are not adjustments affecting the order.

kennyadsl:
Can you also please take a look at the specs for this? I'm seeing that we should already [be covered](https://github.com/solidusio/solidus/blob/4c75c8bc4b2321aca22c1b478648580c30d1112d/backend/spec/features/admin/orders/adjustments_spec.rb#L43-L45) but it looks like that test is a false positive.

jacobherrington:
@kennyadsl that test is actually running on another page. That is running `admin/orders/:id/adjustment` (and should pass), but this change affects `admin/orders/:id/cart`. I will look into writing a regression spec for this change.

ericsaupe:
@jacobherrington any update on this?

jacobherrington:
@ericsaupe I haven't had a chance to revisit it unfortunately!

Issue #2028, opened 2017-06-16
Table rows should be consistent height


Mandily:
On the store stock page, there's a table with multiple height table cells. 
All table cells should be the same height across a row.
<img width="1014" alt="screen shot 2017-06-16 at 10 54 30 am" src="https://user-images.githubusercontent.com/15271353/27238592-63f44112-5282-11e7-97ce-d5d256d2d592.png">

tvdeyen:
Yes. There is some very weird table markup going on. Have a branch laying around that fixes some of these issues. We publish it soon.

samdking:
Hi, I have this same problem. Could you refer me to the branch you mentioned @tvdeyen? Is there any plan to merge this into a release?

tvdeyen:
I did not have published a branch yet because @graygilmore volunteered to continue working on this particular area. This whole page needs a serious and well thought overhaul. My solution was more like a hotfix. Mainly I calculated the row height by the count of stock locations. This is a weak solution and we decided to not merge this.

graygilmore:
Yes, the markup on this page is nothing short of a nightmare. I haven't dug into this area much further, to be honest. I think it needs a completely fresh solution.

aitbw:
Fixed via #2862, @kennyadsl

tvdeyen:
Yes

PR #2862, opened 2018-09-24
[RFC] New stock management

### LINKED ISSUE #2857: Misaligned table on the Stock page
This is a small UI bug on the Stock page. Stock Location, Back orderable, Count on Hand are all aligned incorrectly, this is because these td elements are wrapped in tr instead of being inside the same tr as the other elements on the row. 

image

tvdeyen:
The current stock management view had some flaws:

1. You have to input absolute values. Something that is not reflecting how people actually set stock in their stores
2. The layout was broken, because the HTML we was using is invalid
3. The "add stock item" validation was broken, because the code was not updated after we are not using select2 anymore.

This view could definitely be improved even more, but for now I am proposing to make these ~small~ changes.

### Update stock

solidus-stock-items-modify

### Propagate stock to location not fulfilling the variant yet

solidus-add-stock-item

Closes #2857 #2028 #2530

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/templates/stock_items/stock_location_stock_item.hbs
+++ b/backend/app/assets/javascripts/spree/backend/templates/stock_items/stock_location_stock_item.hbs
@@ -1,5 +1,5 @@
-<td>{{stockLocationName}}</td>
-<td>
+<td class="stock-location-name">{{stockLocationName}}</td>
+<td class="align-center">
   <input id="backorderable-{{id}}"
          name="backorderable"
          type="checkbox"
@@ -8,7 +8,7 @@
          {{#if backorderable}} checked="checked" {{/if}}
    >
 </td>
-<td>
+<td class="align-center">
   {{#if editing}}
     <form>
       <input
--- a/backend/app/assets/javascripts/spree/backend/templates/stock_items/stock_location_stock_item.hbs
+++ b/backend/app/assets/javascripts/spree/backend/templates/stock_items/stock_location_stock_item.hbs
@@ -4,33 +4,32 @@
          name="backorderable"
          type="checkbox"
          value="backorderable"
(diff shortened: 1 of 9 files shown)


kennyadsl:
Thanks @tvdeyen . I agree we need a UX/UI person opinion but in the meantime I personally think it's better than what we have now.

tvdeyen:
Thanks @kennyadsl  I think so as well.

ccarruitero:
@tvdeyen the changes looks good. Definitely better than the current view.

But, I'm not sure we should allow the user just update the stock. I think will be better keep a record about why the stock change, maybe generating a stock movement,  that way is more easy track an error when occurred.

tvdeyen:
@ccarruitero thanks for the feedback.

> But, I'm not sure we should allow the user just update the stock. I think will be better keep a record about why the stock change, maybe generating a stock movement, that way is more easy track an error when occurred.

This PR does not make any changes to our current stock API at all. It behaves exactly like before. It's only UI.

The only difference is that instead of passing the `?force=true` parameter ([to set an absolute stock value](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/stock_items_controller.rb#L40)), it uses the default behaviour of our API that [increases and decreases current stock by utilizing stock movements](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/stock_items_controller.rb#L78-L84).

tvdeyen:
I updated the PR in that you are not able to create negative stock via UI. This [is prevented in the API already](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/stock_items_controller.rb#L80), but I call this better UX if check that already in the UI.

ericsaupe:
Agreed, I like to have the UI handle those edge cases before a server call is made when possible. Good change!

PR #2863, opened 2018-09-24
Chore/make admin more responsive

jacobherrington:
This PR adds Bootstrap breakpoints to most of the forms in the admin area. 

A good next step would be making the sidebar animated on mobile resolutions, but that's a more drastic change.

This probably needs to be squashed if everyone is okay with the change.

### DIFF
--- a/backend/app/views/spree/admin/adjustment_reasons/shared/_form.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/shared/_form.html.erb
@@ -1,5 +1,5 @@
 <div data-hook="admin_adjustment_reason_form_fields" class="row">
-  <div class="col-3">
+  <div class="col-12 col-md-6 col-lg-4">
     <%= f.field_container :name do %>
       <%= f.label :name, class: 'required' %><br />
       <%= f.text_field :name, class: 'fullwidth' %>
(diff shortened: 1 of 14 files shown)


jacobherrington:
@tvdeyen add a viewport tag. Should be useful if we want to make admin mobile friendly.

kennyadsl:
@jacobherrington please take a look at [this issue](https://github.com/solidusio/solidus/issues/666) and [this closed PR](https://github.com/solidusio/solidus/pull/1718) also. We started some work for the collapsible menu but, the way it was required to be, it turned out to be a big change.

At the moment I think we need something more simple, like removing the ability to expand/reduce the menu in mobile/desktop.

jacobherrington:
@kennyadsl Ah, well this PR isn't doing that quite yet. Instead this PR just makes some of the forms collapse on smaller window sizes. 

I will read through those threads if I get to a sidenav PR! Thanks for the info!

tvdeyen:
@jacobeubanks sorry, my bad. Now with the meta tag the admin does not scale any more on small devices. This is an issue for all tablet users, as long as we don't collapse the sidebar as well. I think we should fix that sidebar problem before we take care of the rest.

jacobherrington:
Sounds good @tvdeyen. I will look into that!

jacobherrington:
Closing this until we have a resolved #2961, then I will revisit a more complete version of this.

PR #2869, opened 2018-09-26
Fix Money :html deprecation notice

VitaliyAdamkov:
```
[DEPRECATION] `html` is deprecated - use `html_wrap` instead. Please note that `html_wrap` will wrap all parts of currency and if you use `with_currency` option, currency element class changes from `currency` to `money-currency`.
```

### DIFF
--- a/core/lib/spree/money.rb
+++ b/core/lib/spree/money.rb
@@ -81,9 +81,9 @@ module Spree
     # @param options [Hash] additional formatting options
     # @return [String] the value of this money object formatted according to
     #   its options and any additional options, by default as html.
-    def to_html(options = { html: true })
+    def to_html(options = { html_wrap: true })
       output = format(options)
-      if options[:html]
+      if options[:html_wrap]
         # 1) prevent blank, breaking spaces
         # 2) prevent escaping of HTML character entities
         output = output.sub(" ", "&nbsp;").html_safe


core/lib/spree/money.rb
```diff
@@ -81,12 +81,12 @@ def format(options = {})
     # @param options [Hash] additional formatting options
     # @return [String] the value of this money object formatted according to
     #   its options and any additional options, by default as html.
-    def to_html(options = { html: true })
```
kennyadsl: can you please also:

- update documentation on line 80, that mention the `html: true` param
- deprecate the `html` option into solidus, so that if stores devs are still using it, they will be notified that they need to change their code to pass `html_wrap` instead

core/lib/spree/money.rb
```diff
@@ -81,12 +81,12 @@ def format(options = {})
...
+    def to_html(options = { html_wrap: true })
       output = format(options)
-      if options[:html]
+      if options[:html_wrap]
         # 1) prevent blank, breaking spaces
         # 2) prevent escaping of HTML character entities
-        output = output.sub(" ", "&nbsp;").html_safe
+        output = output.html_safe
```
kennyadsl: Can you please document this change? why is this needed?

kennyadsl:
What do you mean? I'm ok with the change, I just think it needs some more details to be better valuable by who is in charge to review/merge this and by devs that will need to understand why this code changed in their stores.

If there's something I can do to help you with what's requested feel free to ask here on in our Slack channel!

VitaliyAdamkov:
Ruby 2.4.2, Rails 5.1.6, Solidus 2.4.2
At production puma's error log mostly consists from deprecation notice lines: 67%.
I wanted to stop that & thought that just replacing :html with :html_wrap would be enough.
But in that case behavior of display_price changes to inappropriate:

- it renders view with another structure, what breaks css styles
- in some cases it renders plain html as text
- & sometimes it renders broken tags, that are separated with "&nbsp;"'s instead of regular spaces

jacobherrington:
I was working on this last night actually. I agree that getting rid of the huge number of deprecation warnings would be great. 

screen shot 2018-09-25 at 11 03 23 pm

The problem is that back in June [RubyMoney/money deprecated `html` in favor of `html_wrap`](https://github.com/RubyMoney/money/commit/1a752273bb905c1093853de27a1de2376d005820#diff-ed0226fe7de46664f666cd7a6e7e7b46), which doesn't exactly do the same thing.

From [RubyMoney/money docs](https://www.rubydoc.info/gems/money/Money/Formatter):

> Money.ca_dollar(570).format(html: true, with_currency: true)
#=> "$5.70 <span class=\"currency\">CAD</span>"

VS

> Money.ca_dollar(570).format(html_wrap: true, with_currency: true)
#=> "<span class=\"money-currency-symbol\">$</span><span class=\"money-whole\">5</span><span class=\"money-decimal-mark\">.</span><span class=\"money-decimal\">70</span> <span class=\"money-currency\">CAD</span>"


Unfortunately because `html_wrap` and `html` actually do different things, this change breaks a handful of important tests. Also if you look at the frontend after making this change you will see this:

screen shot 2018-09-26 at 6 49 55 am

So it's possible that getting rid of these deprecation warnings would take more than just using `html_wrap` where we were using `html`.

JDutil:
@kennyadsl @VitaliyAdamkov I took over where this issue left off and opened https://github.com/solidusio/solidus/pull/2912/files if you guys want to review my fixes there.

PR #2876, opened 2018-09-27
Remove .travis.yml

jacobherrington:
We decided to not require Ruby 2.3, so we should
make sure we are aware of any issues on Ruby 2.2
and Ruby 2.5.1 once broke Solidus, so it seems smart
to test against it as well.

Creating this PR to test CI and facilitate discussion. 

Related: 
https://github.com/solidusio/solidus/pull/2794
https://github.com/solidusio/solidus/issues/2825


jacobherrington:
image 🤦‍♂️

PR #2877, opened 2018-09-27
Round calculator values based on order currency

### LINKED ISSUE #1957: Round by currency fractional digits rather than by 2 digits
In places like [this](https://github.com/solidusio/solidus/blob/v2.2.1/core/app/models/spree/calculator/flat_percent_item_total.rb#L8) we have code like this:

```ruby
computed_amount = (object.amount * preferred_flat_percent / 100).round(2)
```

The `round(2)` should really round by the number of fractional units for the order's currency.  E.g. `JPY` has no fractional units and we should round to whole numbers.

ericsaupe:
Fixes #1957 by getting the exponent of the desired currency from `Money::Currency`

### DIFF
--- a/core/app/models/spree/calculator/flat_percent_item_total.rb
+++ b/core/app/models/spree/calculator/flat_percent_item_total.rb
@@ -7,7 +7,10 @@ module Spree
     preference :flat_percent, :decimal, default: 0
 
     def compute(object)
-      computed_amount = (object.amount * preferred_flat_percent / 100).round(2)
+      order = object.is_a?(Order) ? object : object.order
+      preferred_currency = order.currency
+      currency_exponent = ::Money::Currency.find(preferred_currency).exponent
+      computed_amount = (object.amount * preferred_flat_percent / 100).round(currency_exponent)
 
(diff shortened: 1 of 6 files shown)


core/app/models/spree/calculator/flat_percent_item_total.rb
```diff
@@ -5,9 +5,11 @@
 module Spree
   class Calculator::FlatPercentItemTotal < Calculator
     preference :flat_percent, :decimal, default: 0
+    preference :currency, :string, default: ->{ Spree::Config[:currency] }
 
     def compute(object)
-      computed_amount = (object.amount * preferred_flat_percent / 100).round(2)
+      currency_exponent = ::Money::Currency.find(preferred_currency).exponent
```
kennyadsl: Shouldn't be the currency taken from the order where the calculator lives instead?

ericsaupe: Yeah, I like that better. I'll make the changes.

ericsaupe:
@kennyadsl this is ready to be looked at again. Thanks!

kennyadsl:
Thanks @ericsaupe !

PR #2879, opened 2018-10-01
Documentation tweaks

jacobherrington:
This PR contains mostly capitalization and small formatting changes
most of this is subjective and can easily be changed if there are strong
opinions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Try out Solidus with one-click on Heroku:
 
 [![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy?template=https://github.com/solidusio/solidus)
 
-## Getting started
+## Getting Started
 
 Begin by making sure you have
 [Imagemagick](http://imagemagick.org/script/download.php) installed, which is
@@ -89,7 +89,7 @@ As part of running the above installation steps, you will be asked to set an adm
 
 The best way to ask questions is via the [#support channel on the Solidus Slack](https://solidusio.slack.com/messages/support/details/).
 
-Installation options
+Installation Options
 --------------------
 
 Instead of a stable build, if you want to use the bleeding edge version of
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Run the `bundle` command to install.
 After installing gems, you'll have to run the generators to create necessary
 configuration files and migrations.
 
-```shell
+```bash
 bundle exec rails g spree:install
 bundle exec rails g solidus:auth:install
 bundle exec rake railties:install:migrations
@@ -68,13 +68,13 @@ bundle exec rake railties:install:migrations
 
 Run migrations to create the new models in the database.
 
-```shell
+```bash
 bundle exec rake db:migrate
 ```
 
 Finally start the rails server
 
-```shell
+```bash
 bundle exec rails s
 ```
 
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ Try out Solidus with one-click on Heroku:
 
 [![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy?template=https://github.com/solidusio/solidus)
 
-## Getting started
+## Getting Started
```
benjaminwil: The contributing article says that the Solidus guides use sentence-style capitalization (`Getting started`) for article titles and subheadings, not title-style (`Getting Started`). See https://guides.solidus.io/contributing.html

If we change this, we should change all of the subheadings throughout the guides, and update the contributors guide as well.

jacobherrington: We can make the inverse change if you feel that is the right way to go -- I just tried to make the fewest changes possible based on the README. I'm all for being consistent when it comes to formatting 👍

benjaminwil: I see what you mean now. Carry on as you see fit. 💯

jacobherrington: I'd rather be consistent with the docs. Made that change. Thanks!

.github/issue_template.md
```diff
@@ -1,14 +1,15 @@
 ### Steps to reproduce
-How can we make this happen
+<!-- How can we make this happen -->
```
kennyadsl: Not sure where/what changes having this text as comment, can you please explain better?

jacobherrington: Sorry, here is the explanation in the commit message:

>Add comments to the issue template
>
>Personally I always end up removing these instructions so it made
sense to me to go ahead and make them comments. This commit also
adds instructions to get the current Solidus version from Rails
console.

This just makes them not show up in the issue if someone doesn't remove them. It's purely my preference and I'm happy to remove that change.

kennyadsl: I did read the commit description but still can't understand the final effect of this. Will that text be invisible in GH? If yes, what about the `Steps to reproduce` title that will still be visible without any content inside?

jacobherrington: ### Steps to reproduce

Example

### Expected behavior

Example

### Actual behavior

Example

### System configuration
**Solidus Version**:

2.7.0
**Extensions in use**:
N/A

# Source for this comment:
```
### Steps to reproduce
<!-- How can we make this happen -->
Example

### Expected behavior
<!-- Tell us what should happen -->
Example

### Actual behavior
<!-- Tell us what happens instead -->
Example

### System configuration
**Solidus Version**:
<!-- Unsure? Check with `Spree.solidus_version` -->
2.7.0
**Extensions in use**:
N/A
```

jacobherrington: This is how it should work.

jacobherrington: @kennyadsl Like I said, I'm am happy to get rid of this change if you think it doesn't work.

kennyadsl: Gotcha, thanks for the explanation. I have no strong opinion around that but I'm seeing that it's also used [here](https://github.com/stevemao/github-issue-templates) so 👍 . What about opening a new PR for this? I think it's not really related to this PR and could make sense to have a separate discussion around that.

jacobherrington: @kennyadsl done!

Issue #2882, opened 2018-10-02
End of Life announced for sass dependency.


JDutil:
Ruby Sass is deprecated and will be unmaintained as of 26 March 2019. It would be good to consider changing the sass dependency to their suggested `sassc-ruby` or more likely `sassc-rails` library to ensure Solidus dependencies are maintained in the future.

### Steps to reproduce
1. Uninstall sass gem: `gem uninstall sass`
2. Install sass gem: `gem install sass`

### Expected behavior
No EOL messages for dependencies.

### Actual behavior
```
Post-install message from sass:

Ruby Sass is deprecated and will be unmaintained as of 26 March 2019.

* If you use Sass as a command-line tool, we recommend using Dart Sass, the new
  primary implementation: https://sass-lang.com/install

* If you use Sass as a plug-in for a Ruby web framework, we recommend using the
  sassc gem: https://github.com/sass/sassc-ruby#readme

* For more details, please refer to the Sass blog:
  http://sass.logdown.com/posts/7081811
```

### System configuration
**Solidus Version**: Master

**Extensions in use**: None

PR #2883, opened 2018-10-02
Move to sassc-rails

jacobherrington:
Ruby Sass is deprecated and will be unmaintained. This commit
moves Solidus to the recommended sassc-rails gem.

Refs #2882

### DIFF
--- a/backend/solidus_backend.gemspec
+++ b/backend/solidus_backend.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'jbuilder', '~> 2.6'
   s.add_dependency 'jquery-rails'
   s.add_dependency 'kaminari', '~> 1.1'
-  s.add_dependency 'sass-rails'
+  s.add_dependency 'sassc-rails'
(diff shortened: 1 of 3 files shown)


jacobherrington:
This is a WIP, but wanted to make sure this was the direction we should go.

backend/solidus_backend.gemspec
```diff
@@ -29,8 +29,8 @@ Gem::Specification.new do |s|
   s.add_dependency 'jbuilder', '~> 2.6'
   s.add_dependency 'jquery-rails'
   s.add_dependency 'kaminari', '~> 1.1'
-  s.add_dependency 'sass-rails'
-  s.add_dependency 'sass', '>= 3.5.2'
+  s.add_dependency 'sassc-rails'
+  s.add_dependency 'sassc'
```
tvdeyen: We should be fine with only depending on `sassc-rails`.

jacobherrington: Okay, thanks I will make that change and follow up by making sure the documentation is up to date with this change.

PR #2887, opened 2018-10-03
Use logo.svg in the README

jacobherrington:
I noticed that we don't use the SVG in the readme.

Here is the difference:

screen shot 2018-10-03 at 5 37 39 am


jacobherrington:
We can copy the SVG over to the root of this repo if that's better?

README.md
```diff
@@ -1,5 +1,5 @@
 
-![](https://raw.githubusercontent.com/solidusio/solidus/master/solidus.png)
+<img src="./logo.svg" width=350>
```
tvdeyen: Why don't we use markdown anymore?

ericsaupe: Regular HTML works in markdown but I'd prefer the markdown version too.

ericsaupe: AFAIK there isn't a way to manipulate size using Markdown.

ericsaupe: Yeah, I usually prefer Markdown over Markdown + HTML but I'm not totally against having HTML in the README. Is it worth it/possible to resave the SVG to a smaller size or will it simply enlarge itself by default of how the Markdown is rendered? Just thinking of other options if we really were set on not doing the HTML

jacobherrington: It's pretty huge, but if that's okay markdown works

Issue #2864, opened 2018-09-24
Bootstrap tooltips are broken on Firefox


jacobherrington:
Fairly sure this is an issue with Bootstrap, but there seem to be some fixes that we could implement for Solidus. I'm also not sure what browsers we "support." 

https://www.useloom.com/share/318e1a8ff47e4c0f909b6749b13e767a

PR #2889, opened 2018-10-03
Must hide tooltips before re-rendering or they'll persist after eleme…

JDutil:
…nts removed from DOM.

refs #2864 

The issue with tooltips persisting on line item form is caused by the original tooltip element being removed from the DOM when the templates re-rendered after clicking an action.  The recommended workaround for this is to first hide your tooltip before removing the elements from the DOM: https://github.com/twbs/bootstrap/issues/2298

I haven't browsed around looking for this issue in other spots, but it wouldn't surprise me if there are other instances of this to fix. Would like to get feedback on this solution before checking for other places to fix.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -87,6 +87,7 @@ Spree.ready(function(){
         },
         dataType: 'script',
         success: function(response) {
+          el.tooltip('dispose')
           el.parents("tr").fadeOut('hide', function() {
             $(this).remove();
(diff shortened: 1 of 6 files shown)


kennyadsl:
Could this be related to #2421 ?

JDutil:
@kennyadsl yea it looks like it.  I'll probably close this in favor of fixing in that file so solution works everywhere, but my initial attempts don't seem to work.

JDutil:
Closed in favor of https://github.com/solidusio/solidus/pull/2890

PR #2891, opened 2018-10-04
Fix RubyMoney/Money i18n deprecation warning

jacobherrington:
This PR fixes the warning message coming from RubyMoney/money discussed in https://github.com/RubyMoney/money/issues/816

```
[DEPRECATION] `use_i18n` is deprecated - use `Money.locale_backend = :i18n` instead
```

It might be worth waiting to see if this is fixed in the money gem before we merge this PR.

### DIFF
--- a/core/config/initializers/money.rb
+++ b/core/config/initializers/money.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+
+Money.locale_backend = :i18n


jacobherrington:
😭 
screen shot 2018-10-04 at 4 30 11 pm

kennyadsl:
As far as I get from [this comment](https://github.com/RubyMoney/money/issues/816#issuecomment-427148418) in the issue on `RubyMoney/money`, you have to set that configuration up in each application. I'm not sure that initializer works but setting that [here](https://github.com/solidusio/solidus/blob/fc8c0723eff7438546db4b26a3f4bc6950dbc737/core/lib/spree/testing_support/dummy_app.rb#L91) it removes the deprecations in specs. I think we need to put that setting in a place with is also configurable by users with the other `Spree::Money` configurations.

jacobherrington:
That might be a better solution, although this does fix it when I run tests which was what I was trying to fix. I'll try to make it a configuration option instead though.

tvdeyen:
> I think we need to put that setting in a place with is also configurable by users with the other Spree::Money configurations.

Yes, I agree. There is also `money-rails` that [has an install generator](https://github.com/RubyMoney/money-rails/blob/master/lib/generators/money_rails/initializer_generator.rb), that creates the above mentioned initializer in the host app. Although I don't like to add another dependency, I think we should at least consider it or implement our own generator during Solidus install.

jacobherrington:
Work is being done on #2912

PR #2897, opened 2018-10-05
Improve trustworthiness of a content items price.

jacquesporveau:
A line items price is more trustworthy than a variants price in regards
to what a customer is paying for something.

In the case that there are applications that charge different prices for variants based on criteria then using a line items price is a more accurate representation and a better source of truth as to what someone will pay for an item.

For shipping calculators that depend on how much something is costs this is an improvement.

### DIFF
--- a/core/app/models/spree/stock/content_item.rb
+++ b/core/app/models/spree/stock/content_item.rb
@@ -31,7 +31,7 @@ module Spree
       end
 
       def price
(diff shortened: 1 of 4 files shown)


tvdeyen:
Please rebase to get rid of the Netlify errors.

jacquesporveau:
Thanks guys!

Issue #2868, opened 2018-09-26
Products at mulitple Stock Locations appear as unique variants


jacobherrington:
### Steps to reproduce
**1.** Create a store with two (or three) Stock Locations.
<img width="1163" alt="screen shot 2018-09-25 at 7 23 06 pm 1" src="https://user-images.githubusercontent.com/11466782/46050364-60e7e500-c0f9-11e8-9894-a30fc4374dd1.png">
**2.** Choose a product with no variants (I used the RoR Mug). 
**3.** Add stock to the product in each location.
<img width="879" alt="screen shot 2018-09-25 at 7 30 16 pm" src="https://user-images.githubusercontent.com/11466782/46050387-74934b80-c0f9-11e8-99fc-71b68e859ba7.png">
**4.** Create a new order from `admin/orders` and try to add the item. You will see the two (or three) variants with the same SKU.
<img width="1139" alt="screen shot 2018-09-25 at 7 23 58 pm" src="https://user-images.githubusercontent.com/11466782/46050399-7d841d00-c0f9-11e8-8968-719dc615f206.png">

### Expected behavior
One variant should exist in the dropdown, regardless of the number of Stock Locations.

Please correct me if this is not ideal.

### Actual behavior
There are as many variants listed in the drop down as there are Stock Locations with stock.

### System configuration
**Solidus Version**: 2.7.0

**Extensions in use**: Sandbox

mayanktap:
Hi @jacobherrington , I am able to reproduce this issue in my local machine and tried to fix. I guess I would be able to fix this. 
I would like to work on this. Can I ?

jacobherrington:
Please do @mayanktap! Let me know if you need help and I'll try to give you some guidance. :)

mayanktap:
@jacobherrington , According to my understanding and investigation, I made some changes and tested it thoroughly. I will raise the PR so that you can review and take decision accordingly. 
I would just need some guidance in the process of making PR.
 
I want to verify one point regarding the process of raising PR: 
1.) I am raising PR as base branch as **master**. That means the new branch that I will create to push the change will be from **master** branch. Is it correct?

mayanktap:
Is there any naming convention regarding the name of the branch? I didn't find anything related to this in CONTRIBUTION documentation.

jacobherrington:
Wow that's great 🎉 

Yes, we usually branch off of master and PR back to master and branch names aren't a big deal to Solidus.

Thanks a ton for your help! 👍

[...]

PR #2898, opened 2018-10-06
Issue 2868: Products at mulitple Stock Locations appear as unique variants.

mayanktap:
Please review the change and suggest if this requires some more work.

Fix for Issue: https://github.com/solidusio/solidus/issues/2868


ericsaupe:
I think this addresses it nicely. 

A couple of notes, I would redo your commit message to better describe what is actually being changed and, if possible, write a regression test that fails for current versions of Solidus but is resolved with your fix.

Thanks!

jacobherrington:
Great first PR! 🎉 

Regarding Eric's suggestion on the commit messages, we have some great resources for good commit messages in [contributing.md](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines), if you're interested. 

Thanks for fixing this bug!

mayanktap:
@ericsaupe , @jacobherrington Thanks for your review comments. 

Actually, I am familiar in using GIT from terminal instead from UI and because of that I was not able to figure out where GIT commit message is to be changed. I will make the necessary changes in the commit message. 

Regarding the failing test, I check the test case and tried to run the test with and without my code which leads to the failing of same test. So as per my understanding, the failing of test case is not because of the this change. 
@ericsaupe , Can you please specify what kind of regression test are you talking about?

ericsaupe:
@mayanktap sure! Basically, a test that would do something like this:

```
it "returns unique products" do
  Spree::Variant.suppliable.count == variants.count
end
```

That's a very simple test that is obviously mostly pseudocode but the idea is that you're going to call the `suppliable` method and see how many variants come back and compare it with the distinct number of variants we expect. Without your fix this should fail since the `suppliable` method will return duplicated variants but after your fix the test should pass.

Does that make sense?

mayanktap:
@ericsaupe , @jacobherrington

I have improved my commit history with proper commit message. Can you please review this PR? I am not sure how to add reviewers in this PR.

spaghetticode:
@mayanktap after the `push --force` I see there are some spurious changes in the commits for `CHANGELOG.md` file. Can you rework your commits and remove these changes?

mayanktap:
Hi @spaghetticode ,
I guess you are talking about [this](https://github.com/solidusio/solidus/commit/014feb3ddce2fc8cbf8a3777a8bb3196d4ab4b14) change 
I removed those CHANGELOG.md changes because that change was not available in solidus/master and I don't want to merge any changes to solidus/master which are not in the scope of this PR. 

To avoid more confusion, I am closing this PR and creating the new one. The new PR is: https://github.com/solidusio/solidus/pull/3063

PR #2900, opened 2018-10-07
symbol can't be argument for :count in east slavic locales

bitberries:
Hi guys!

Symbol can't be argument for :count in east slavic locales, east slavic pluralization rule requires number because it uses modulo division, [https://github.com/svenfuchs/rails-i18n/blob/master/lib/rails_i18n/common_pluralizations/east_slavic.rb](https://github.com/svenfuchs/rails-i18n/blob/master/lib/rails_i18n/common_pluralizations/east_slavic.rb)

### DIFF
--- a/backend/spec/features/admin/users_spec.rb
+++ b/backend/spec/features/admin/users_spec.rb
@@ -37,7 +37,7 @@ describe 'Users', type: :feature do
     end
 
     it 'can go back to the users list' do
-      expect(page).to have_link Spree::LegacyUser.model_name.human(count: :other), href: spree.admin_users_path
+      expect(page).to have_link Spree::LegacyUser.model_name.human(count: Spree::I18N_GENERIC_PLURAL), href: spree.admin_users_path
     end
 
     it 'can navigate to the account page' do
--- a/backend/app/views/spree/admin/promotion_categories/index.html.erb
+++ b/backend/app/views/spree/admin/promotion_categories/index.html.erb
@@ -1,5 +1,5 @@
 <% admin_breadcrumb(link_to plural_resource_name(Spree::Promotion), spree.admin_promotions_path) %>
-<% admin_breadcrumb(Spree::PromotionCategory.model_name.human(count: :many)) %>
+<% admin_breadcrumb(plural_resource_name(Spree::PromotionCategory)) %>
 
 
 <% content_for :page_actions do %>


jacobherrington:
@bitberries thanks for making this PR! 🎉 

We have some great resources for writing good commit messages in our [contributing.md](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines) file, if you're interested!

bitberries:
Changed commit message, is it okay now?

kennyadsl:
@bitberries I'm not sure to get the issue with current code and the link provided does not help understanding (for me). Can you please point to some other resource or explain the issue better?

bitberries:
Exception screenshot
P.S. Spree also has this bug last 2 years

bitberries:
Also I never saw that count parameter was not numerical. Everywhere it is number. It is assumed by pluralization rules that it will be a number, check it https://github.com/svenfuchs/rails-i18n/tree/master/lib/rails_i18n/common_pluralizations - everywhere comparisons only with numbers.

This code is just randomly working correctly for the English language.

kennyadsl:
> This code is just randomly working correctly for the English language.

That's what I was trying to understand, thanks. It looks like it's not supported by `rails-i18n` so we should avoid using it here as for your PR.

I've also found the initial work for this in Solidus: #1191

Are there other occurrences of this wrong pluralization in Solidus? Is it worth to update all of them in this PR maybe, what do you think?

kennyadsl:
@bitberries I've just found this one: https://github.com/solidusio/solidus/blob/fac5b359bb0b27de638d2ae4abb82abc584ad177/backend/spec/features/admin/users_spec.rb#L40

This should be unsupported as well. I know that this will not break any thing since specs does not run with east slavic locale but we should not encourage bad practices in the codebase. Every line of code could be taken as example and reused again, maybe somewhere where it will be dangerous.

bitberries:
Yep, you are right. I fixed it.

backend/spec/features/admin/users_spec.rb
```diff
@@ -37,7 +37,7 @@
     end
 
     it 'can go back to the users list' do
-      expect(page).to have_link Spree::LegacyUser.model_name.human(count: :other), href: spree.admin_users_path
+      expect(page).to have_link Spree::LegacyUser.model_name.human(count: Spree::I18N_GENERIC_PLURAL), href: spree.admin_users_path
```
kennyadsl: why can't we use `plural_resource_name ` here?

bitberries:
Now using plural_resource_name in spec.
Squashed into one commit.

backend/spec/features/admin/users_spec.rb
kennyadsl: chatted with @bitberries on Slack, I think that's the best way as well.

PR #2902, opened 2018-10-07
fix translations keys

bitberries:
Hi guys,

In admin, at order page in cart tab summary renders without translations for shipment and payment states.
I fixed it

.


bitberries:
Thought that branch name will be enough.
Changed the commit message :)

kennyadsl:
Now there are 3 commits, the goal is to have just the correct one, I think you just need to squash them, thanks!

bitberries:
Squashed

PR #2905, opened 2018-10-08
add bcc fields to stores

afdev82:
This PR adds the possibility to specify BCC email addresses for the confirmation order emails in the Store admin tab.


kennyadsl:
Why bcc should only apply to the order confirm email?

afdev82:
Because I needed only for that... But once we have this field we could use it also for other methods that sends emails.
In my case the need was to have only the confirmation order emails in BCC to other email addresses.

kennyadsl:
I think that we need to provide a way to let devs choose which emails they want to add bcc into if we want to introduce this feature.
Other users could need to add it to shipment confirmation emails also for example and if they need to override code to make that happen I think this is not worth the change since they could just change code to add the bcc to the order confirm email as well.

IMO if we don't go in this direction it's better to keep this customization out of core, maybe an extension or directly into apps.

afdev82:
What you say it makes sense.
I just thought that the functionality to receive an email when an order is performed instead of having to log in in the website and check there, is pretty common and nothing stop to add the same functionality for other emails too.
I wanted only to reduce the need to override views / methods in my app just to add the bcc field to the emails, because it's always a problem with upgrades.
If it's worth to have in the core or in an extension, I don't know.

ericsaupe:
I agree with @kennyadsl, I think there is value in having BCC fields for stores but I'd like to see this solution expanded to include the possibility of having BCCs on all emails sent. It might be a good idea to have this start as an extension and then reopen with a full solution into core once it's been fully implemented.

afdev82:
Ok :+1:

PR #2907, opened 2018-10-10
DRY Spree::OrderContents

spaghetticode:
Some duplicated code has been removed from `Spree::OrderContents` and replaced
by a direct call to `Spree::Order#recalculate`

### DIFF
--- a/core/app/models/spree/order_contents.rb
+++ b/core/app/models/spree/order_contents.rb
@@ -79,12 +79,8 @@ module Spree
       line_item
     end
 
-    def order_updater
-      @updater ||= Spree::OrderUpdater.new(order)
-    end
-
     def reload_totals
-      order_updater.update
+      @order.recalculate
     end
 
     def add_to_line_item(variant, quantity, options = {})


core/app/models/spree/order_contents.rb
```diff
@@ -79,12 +79,8 @@ def after_add_or_remove(line_item, options = {})
       line_item
     end
 
-    def order_updater
-      @updater ||= Spree::OrderUpdater.new(order)
-    end
-
```
tvdeyen: I think we should deprecate removals first. There might be people that decorated this class like this:

```rb
module OrderUpdaterDecorator
  def order_updater
    super
    do_something
  end
end

Spree::OrderContents.prepend OrderUpdatesDecorator
```

That would now fail.

kennyadsl: I thought the same but it's a private method, do we want to use deprecations for those as well?

elia: It's probably worth deprecating it in v2.7.x and keep the removal from master, there's no risk of leaving the deprecation there forever

kennyadsl: @elia We don't have a real policy around that but we usually keep deprecated method around for at least a couple of `MINOR` releases, often much more than 2 👻

tvdeyen: We agreed on not deprecating a private method

tvdeyen:
We agreed on not to deprecate a private method

PR #2908, opened 2018-10-10
Spree::FulfilmentChanger stock allocation fix

spaghetticode:
When splitting a shipment item to the same stock location, if the stock location has currently zero availability for that product then the state of the target shipment inventory unit should be determined by the original shipment inventory unit state:

* when  originally backordered, as there is still no available stock, then it  should remain backordered

* when originally on hand, then it should remain on hand, as there is no actual  stock movement in this case (the fact that the stock location now has no  availability for that product is not actually relevant)

A clarifying example for the second scenario:

* stock location L has 1 unit for product A
* customer purchases 1 unit of product A
* the order is created with product A on hand
* product A is automatically included in shipment S
* stock location L is now empty (0 units) for product A
* if the admin splits product A from shipment S to a new shipment from the  very same stock location L, product A should remain on hand also in the new shipment

### DIFF
--- a/core/app/models/spree/fulfilment_changer.rb
+++ b/core/app/models/spree/fulfilment_changer.rb
@@ -46,7 +46,7 @@ module Spree
       desired_shipment.save! if desired_shipment.new_record?
 
       # Retrieve how many on hand items we can take from desired stock location
-      available_quantity = [desired_shipment.stock_location.count_on_hand(variant), 0].max
+      available_quantity = [desired_shipment.stock_location.count_on_hand(variant), default_on_hand_quantity].max
 
       new_on_hand_quantity = [available_quantity, quantity].min
       unstock_quantity = desired_shipment.stock_location.backorderable?(variant) ? quantity : new_on_hand_quantity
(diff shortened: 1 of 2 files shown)


kennyadsl:
@spaghetticode can you please rebase against master? Failures should go away

core/app/models/spree/fulfilment_changer.rb
```diff
@@ -114,6 +114,14 @@ def handle_stock_counts?
       current_shipment.order.completed? && current_stock_location != desired_stock_location
     end
 
+    def default_on_hand_quantity
+      if current_stock_location != desired_stock_location
+        0
+      else
+        current_shipment.inventory_units.where(variant: variant).on_hand.limit(quantity).count
```
kennyadsl: Why the `limit(quantity)` is needed here? Shouldn't `current_shipment.inventory_units.where(variant: variant).on_hand` already return only the inventory units present in the shipment for this variant, which should not be more than `quantity`?

spaghetticode: @kennyadsl I think your note makes sense, that code is not necessary. I'm changing the code soon.

kennyadsl:
Thanks!

Issue #2910, opened 2018-10-11
Flaky checkout feature spec


kennyadsl:
We have some random failures on the CI, for example [this one on master](https://circleci.com/gh/solidusio/solidus/9321#tests/containers/2).

This always happens on the same spec: [spec/features/checkout_spec.rb:364](https://github.com/solidusio/solidus/blob/fac5b359bb0b27de638d2ae4abb82abc584ad177/frontend/spec/features/checkout_spec.rb#L349-L365) and looking at the [artifacts on the build](https://circleci.com/gh/solidusio/solidus/9321#artifacts/containers/2) and there is a clear evidence that it is an issue with credit card.

![Error on Ci](https://9321-30985840-gh.circle-artifacts.com/2/test-artifacts/screenshot_2018-10-11-18-05-03.429.png)

This error happens when the credit card number is not valid, that's [the only check done](https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/core/app/models/spree/payment_method/bogus_credit_card.rb#L28) on the Bogus payment method used in specs.

I think this is related to these issues: 

- https://github.com/stripe/jquery.payment/issues/149
- https://bugs.chromium.org/p/chromedriver/issues/detail?id=1771&q=sendkeys&sort=-id&colspec=ID%20Status%20Pri%20Owner%20Summary

There are some suggested solutions around, like [this one](https://github.com/eliotsykes/rails-testing-toolbox/blob/master/stripe_helper.rb) but I'd like to avoid changing the normal Capybara behavior as much as possible.

PR #2911, opened 2018-10-11
Fill credit card input with force in checkout spec

kennyadsl:
This fixes the flaky spec described into #2910.

Capybara `fill_in` sometime does not fill the input with all the string provided so this fix will fill it with JS bypassing the default behavior.


kennyadsl:
I'm re-pushing this branch several times to see if this really fixes it

kennyadsl:
This seems to work, I don't love it but can't find any other solution at the moment. WDYT?

tvdeyen:
Merged because all PRs are currently suffering of breaking builds.

PR #2912, opened 2018-10-11
Fix money deprecation warning.

### LINKED ISSUE #2870: Deprecation warnings from RubyMoney/money
### Steps to reproduce
`bundle exec rspec` in `solidus/backend`.

Spree::Money.to_html logs a bunch of deprecation warrnings.

https://github.com/solidusio/solidus/blob/8caf58aabc3d5390051d1f866a537df34f7f6181/core/lib/spree/money.rb#L84

### Expected behavior
Test should run with a minimal amount of noise.

### Actual behavior
An absurd number of deprecation warnings log during the tests:
screen shot 2018-09-25 at 11 03 23 pm

### System configuration
**Solidus Version**: 2.7.0

---

#2869 attempted to fix this, I will repost my comment from that PR here:

I was working on this last night actually. I agree that getting rid of the huge number of deprecation warnings would be great. 

screen shot 2018-09-25 at 11 03 23 pm

The problem is that back in June [RubyMoney/money deprecated `html` in favor of `html_wrap`](https://github.com/RubyMoney/money/commit/1a752273bb905c1093853de27a1de2376d005820#diff-ed0226fe7de46664f666cd7a6e7e7b46), which doesn't exactly do the same thing.

From [RubyMoney/money docs](https://www.rubydoc.info/gems/money/Money/Formatter):

> Money.ca_dollar(570).format(html: true, with_currency: true)
#=> "$5.70 <span class=\"currency\">CAD</span>"

VS

> Money.ca_dollar(570).format(html_wrap: true, with_currency: true)
#=> "<span class=\"money-currency-symbol\">$</span><span class=\"money-whole\">5</span><span class=\"money-decimal-mark\">.</span><span class=\"money-decimal\">70</span> <span class=\"money-currency\">CAD</span>"

Unfortunately because `html_wrap` and `html` actually do different things, this change breaks a handful of important tests. Also if you look at the frontend after making this change you will see this:

screen shot 2018-09-26 at 6 49 55 am

So it's possible that getting rid of these deprecation warnings would take more than just using `html_wrap` where we were using `html`.

JDutil:
fixes #2870

### DIFF
--- a/core/config/initializers/money.rb
+++ b/core/config/initializers/money.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+require 'money'
+
+Money.locale_backend = :i18n
--- a/core/lib/spree/money.rb
+++ b/core/lib/spree/money.rb
@@ -1,9 +1,5 @@
 # frozen_string_literal: true
 
-require 'money'
-require 'monetize'
-require 'active_support/core_ext/string/output_safety'
-
 module Spree
   # Spree::Money is a relatively thin wrapper around Monetize which handles
   # formatting via Spree::Config.
@@ -77,16 +73,22 @@ module Spree
       @money.format(@options.merge(options))
     end
 
-    # @note If you pass in options, ensure you pass in the html: true as well.
+    # @note If you pass in options, ensure you pass in the { html_wrap: true } as well.
     # @param options [Hash] additional formatting options
     # @return [String] the value of this money object formatted according to
-    #   its options and any additional options, by default as html.
-    def to_html(options = { html: true })
+    #   its options and any additional options, by default with html_wrap.
+    def to_html(options = { html_wrap: true })
       output = format(options)
-      if options[:html]
-        # 1) prevent blank, breaking spaces
-        # 2) prevent escaping of HTML character entities
-        output = output.sub(" ", "&nbsp;").html_safe
+      # Maintain compatibility by checking html option renamed to html_wrap.
+      if options[:html] || options[:html] == false
+        Spree::Deprecation.warn <<-WARN.squish, caller
+          Spree::Money#to_html called with Spree::Money#to_html(html: #{options[:html]}),
+          which will not be supported in the future.
+          Instead use :html_wrap e.g. Spree::Money#to_html(html_wrap: #{options[:html]})
+        WARN
+      end
+      if options[:html_wrap] || options[:html]
+        output = output.html_safe
       end
       output
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - Implement stock location sorters [#2783](https://github.com/solidusio/solidus/pull/2783) ([aldesantis](https://github.com/aldesantis))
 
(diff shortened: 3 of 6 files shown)


jacobherrington:
Does this cause the UI to break horrendously (see #2870) or is it working?

JDutil:
@jacobherrington I don’t see the html safe issues from the other PR, but there are style breaks that will need to be fixed. Doesn’t look major at all just a line break was added between the currency symbol and value, but I still need to fix specs too.

JDutil:
@jacobherrington I believe specs should be passing now if you want to take another look. I didn't end up having to make any styling changes I just had to remove the gsub changing whitespace into html entity for whitespace as that was causing invalid html to render making the line break. The spec changes are mainly just taking into account the spans that are now rendered, and were not rendered before.  In theory this should be fine for most stores, but there is always the possibility some custom styles my conflict...

core/lib/spree/money.rb
```diff
@@ -4,6 +4,8 @@
 require 'monetize'
 require 'active_support/core_ext/string/output_safety'
 
+Money.locale_backend = :i18n
```
tvdeyen: I think we should not add global configuration in a class like this.

Could we move this into it's own `config/initializers/money.rb` of the host app instead? And tell users to do so in their current apps and add it into the install generator? I like to be explicit with these things, especially if it's not our gem we are configuring here.

JDutil: I agree this should be done in a more configurable manner, but I don’t think we should require everyone to update their apps with an initializer by default. I’m thinking just allow apps to override the setting if they really want to. WDYT?

tvdeyen: I’m not a fan of adding a setting in our app that’s a proxy to a foreign gem. 

What about putting it inside a `before_initialize` block? Are engine `config/initialize` files read before host apps files? That we be preferable. Someway where the host app is able to set it in a default (read: mentioned in money gems readme) kind of way and we still act as fallback/default?

core/config/initializers/money.rb
```diff
@@ -0,0 +1,3 @@
+require 'money'
```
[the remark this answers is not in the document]
JDutil: This was fixed, but comment didn't get hidden.

JDutil:
@tvdeyen I updated this PR so that Money is now initialized in `core/config/initializers/money.rb`, but that meant I needed to change the spec to use `rails_helper` instead of `spec_helper`.  Now the main apps using solidus can simply override the Money defaults by adding their own initializer if they want to.

JDutil:
@ericsaupe squashed 👍

kennyadsl:
This could be a breaking change for some store right? Maybe we should add a specific line into the CHANGELOG. @JDutil what do you think?

JDutil:
@kennyadsl added to the changelog.

kennyadsl:
Made a test locally and I can't see any UI change from what we previously have, thanks @JDutil 👍 🎉

JDutil:
@kennyadsl yea there shouldn't be visual changes unless you've customized how your price is displayed for something like having a big number and small decimals or something along those lines or are applying css to a span tag instead of a class. 

I don't think anyone should really be effected unless they went out of their way to really change the pricing display, but going forward these new spans rendered will allow people to more easily do any custom styles for big currency symbol or number and small decimal etc...

Issue #2652, opened 2018-03-27
Card used for payment being made the default


ghost:
### Steps to reproduce

1. Make sure that user has atleast two credit cards, say Visa and MasterCard associated with his account and one of the two say Visa is the default credit card
2. Perform a checkout but during checkout use the non default card, in our case MasterCard for payment.

### Expected behavior

Visa should stay the default card.

### Actual behavior

MasterCard becomes the default card.

### System configuration

Solidus Version: 1.4.2 

### Extensions in use

1. solidus_auth_devise
2. solidus_braintree

When `order.next` changes the order state from `payment` to `complete`, the card used for payment is automatically made the default. Why is the card used for payment forcibly made the default card ?. And, how can this be prevented ?

Thanks

PR #2913, opened 2018-10-12
Add make_default method on AddPaymentSourcesToWallet class

vassalloandrea:
Ref: #2652

### After some consideration with @tvdeyen 

With this PR we want to separate the `add_to_wallet` behavior from the `make_default` one to help if someone wants to change only the `make_default` logic without overriding the `add_to_wallet`.

## How it works

Now the `add_to_wallet` method calls the `make_default` method after adding on the user wallet the payment sources.
To change the `make_default` behavior now we can create own class and override the `make_default` method.

E.g.: https://github.com/solidusio/solidus/pull/2913#issuecomment-429913171

### DIFF
--- a/core/app/models/spree/wallet/add_payment_sources_to_wallet.rb
+++ b/core/app/models/spree/wallet/add_payment_sources_to_wallet.rb
@@ -24,14 +24,19 @@ class Spree::Wallet::AddPaymentSourcesToWallet
       # add valid sources to wallet and optionally set a default
       if sources.any?
         # arbitrarily sort by id for picking a default
-        wallet_payment_sources = sources.sort_by(&:id).map do |source|
+        sources.sort_by(&:id).each do |source|
           order.user.wallet.add(source)
         end
 
-        order.user.wallet.default_wallet_payment_source =
-          wallet_payment_sources.last
+        make_default
       end
     end
+  end
+
+  protected
+
+  def make_default
+    order.user.wallet.default_wallet_payment_source = order.user.wallet_payment_sources.last
   end
 
   private
--- a/core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
+++ b/core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Spree::Wallet::AddPaymentSourcesToWallet, type: :model do
+  let(:order) { create(:order_ready_to_complete) }
+
+  describe '#add_to_wallet' do
+    subject { described_class.new(order) }
+
+    it 'saves the payment source' do
+      expect { subject.add_to_wallet }.to change {
+        order.user.wallet.wallet_payment_sources.count
+      }.by(1)
+    end
+  end
+end


kennyadsl:
@tvdeyen I think you are right but we are already storing the credit card in the database, this just changes which one is considered the default one. 

I think it does not change which cards are being shown to the user on the next purchase but the last one used will be preselected I think, while at the moment the preselected one was the first added. Right @vassalloandrea ?

tvdeyen:
> I think you are right but we are already storing the credit card in the database

We store the payment source used on the payment, yes, but also we store it as default payment source in the users wallet. And this should not be the case IMO unless the user wants to. They should opt-in for this.

Also the additional configuration option in this PR seems to be very store specific and should not be in core IMO. What could be in core is a user account page that let the use decide which payment source that they decided to be stored is their default one. Having the store deciding this is wrong IMO.

kennyadsl:
> Also the additional configuration option in this PR seems to be very store specific and should not be in core IMO. What could be in core is a user account page that let the use decide which payment source that they decided to be stored is their default one. Having the store deciding this is wrong IMO.

👍 

What about a class that each store can customize? Like:

```ruby
Spree::Config::Wallet.default_payment_method_setter_class
```

which by default acts exactly as the one we have in core right now?

tvdeyen:
> What about a class that each store can customize?

Why not replace `add_payment_sources_to_wallet_class` with your custom code in your stores config? 

Maybe it makes sense to extract the "make payment method the default one" into a method, that a store then can override?

```rb
Spree.config do |config|
  ...
  config.add_payment_sources_to_wallet_class = 'My::AddPaymentSourcesToWallet'
  ...
end
```

```rb
class My::AddPaymentSourcesToWallet < Spree::Wallet::AddPaymentSourcesToWallet
  def make_default
    order.user.wallet.default_wallet_payment_source = wallet_payment_sources.first
  end
end
```

core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    it 'saves the payment source and calls make_default' do
+      expect(subject).to receive(:make_default)
+      subject.add_to_wallet
+      expect(order.user.wallet.wallet_payment_sources.count).to eq(1)
+    end
+  end
+
+  describe '#make_default' do
+    subject(:call) { -> { described_class.new(order).make_default } }
```
tvdeyen: Why this extra lambda?

core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:order) { create(:order_ready_to_complete) }
+
+  describe '#add_to_wallet' do
+    subject { described_class.new(order) }
+
+    it 'saves the payment source and calls make_default' do
+      expect(subject).to receive(:make_default)
+      subject.add_to_wallet
+      expect(order.user.wallet.wallet_payment_sources.count).to eq(1)
+    end
```
tvdeyen: We do not need to explicitly test the implementation of `make_default` here, as you already have an expectation about the default payment source in the next test.

Also, I prefer to test changes like this:

```rb
expect { subject.add_to_wallet }.to change {
  order.user.wallet.wallet_payment_sources.count
}.by(1)
```

core/app/models/spree/wallet/add_payment_sources_to_wallet.rb
```diff
@@ -24,16 +24,19 @@ def add_to_wallet
...
         end
 
-        order.user.wallet.default_wallet_payment_source =
-          wallet_payment_sources.last
+        make_default
       end
     end
   end
 
+  def make_default
```
tvdeyen: I suggest to make this a `protected` method.

vassalloandrea:
Hey @tvdeyen, I made the changes 👍

core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Spree::Wallet::AddPaymentSourcesToWallet, type: :model do
+  let(:order) { create(:order_ready_to_complete) }
+
+  describe '#add_to_wallet' do
+    subject { described_class.new(order) }
+
+    it 'saves the payment source and calls make_default' do
```
aldesantis: This doesn't seem to actually test that the `make_default` method is called. I'm not sure whether we should do it at all tbh: on one end, it's an extension point and it's probably worth testing; on the other, it's a private interface and it's generally considered bad practice to test them. What do you think?

aldesantis:
@vassalloandrea there's something off with the specs. Could you take a look?

aldesantis:
@solidusio/core-team since some time has passed since we originally discussed this, can someone take another look and make sure we're okay with this one?

aldesantis:
Thanks @vassalloandrea!

PR #2914, opened 2018-10-12
Add BackorderedFirst allocator 

vassalloandrea:
Add a new allocator to build the packages giving the priority on the backorderable stock locations.

## How it works

Using `BackorderableFirst` allocator if the stock location *marked as `default`* can't stock all the required inventory units, all they are shipped from the backorderable stock location. 

*Documentation link*: [documentation](https://github.com/solidusio/solidus/guides/source/developers/shipments/stock-allocator.html.md)

### DIFF
--- a/core/app/models/spree/stock/allocator/backordered_first.rb
+++ b/core/app/models/spree/stock/allocator/backordered_first.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Spree
+  module Stock
+    module Allocator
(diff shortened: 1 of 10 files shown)


kennyadsl:
It's worth to say that this PR depends on #2810

vassalloandrea:
After thinking about this PR I'm going to close it whereas `BackorderedFirst ` allocator can be used in a specific situation and shouldn't be added on the Solidus core.

PR #2915, opened 2018-10-12
Add deprecation warning for Spree.t

### LINKED ISSUE #2773: Deprecate Spree.t
So I'm currently working on a Solidus project and I saw that @jhawthorn has changed `Spree.t` => `t('spree.')`. I'm thinking we should deprecate `Spree.t` as I've been going into the project I've been on and changing `t('spree.')` to `Spree.t`; I'm regretting that now. I think a warning should begin to be emitted to programmers using `Spree.t` and remove it in Solidus 3 or 4.

Issue #2309

What are some thoughts on this?

JDutil:
fixes #2773

### DIFF
--- a/core/spec/lib/i18n_spec.rb
+++ b/core/spec/lib/i18n_spec.rb
@@ -30,23 +30,28 @@ RSpec.describe "i18n" do
   end
 
   it "translates within the spree scope" do
+    expect(Spree::Deprecation).to receive(:warn).twice
     expect(Spree.t(:foo)).to eql("bar")
     expect(Spree.translate(:foo)).to eql("bar")
   end
 
   it "prepends a string scope" do
+    expect(Spree::Deprecation).to receive(:warn)
     expect(Spree.t(:foo, scope: "bar")).to eql("bar within bar scope")
   end
 
   it "prepends to an array scope" do
+    expect(Spree::Deprecation).to receive(:warn)
     expect(Spree.t(:foo, scope: ["bar"])).to eql("bar within bar scope")
   end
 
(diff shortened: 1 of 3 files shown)


kennyadsl:
@JDutil specs failure are actually related, can you please take a look? 🙏

core/spec/lib/i18n_spec.rb
```diff
@@ -29,25 +29,35 @@
     I18n.reload!
   end
 
-  it "translates within the spree scope" do
-    expect(Spree.t(:foo)).to eql("bar")
-    expect(Spree.translate(:foo)).to eql("bar")
-  end
+  describe '#translate' do
+    after do
+      Spree::Deprecation.behavior = :raise
```
kennyadsl: Another approach could be stubbing the deprecation warning 

```
expect(Spree::Deprecation).to receive(:warn)
```

tvdeyen: Leaving it like that means that from the moment we ran this spec we always raise for deprecations even if we change this in the spec helper in the future. Either wrap the spec in a

```rb
around do |example|
  Spree::Deprecation.silence { example.run }
end
```

or do what @kennyadsl proposes.

Third possibility is to remove specs for deprecated methods.

JDutil:
I prefer @kennyadsl solution will update :+1:

kennyadsl:
@tvdeyen in this case I'd leave the specs since we are not really removing this functionality but just printing a deprecation warning. I think it's better to be sure stores and extensions that still relies on that method can work.

PR #2920, opened 2018-10-17
Add response with error message on failure destroy action

bitberryru:
Hello again!

Now I noticed bug in admin when validations prevent destroy. 
Your resource controller doesn't handle js format when destroy failures but your js code expects error message 

Test and fix in commit! :)

### DIFF
--- a/backend/spec/controllers/spree/admin/resource_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/resource_controller_spec.rb
@@ -32,6 +32,13 @@ describe Spree::Admin::WidgetsController, type: :controller do
     model do
       acts_as_list
       validates :name, presence: true
+      before_destroy :check_destroy_constraints
+
+      def check_destroy_constraints
+        return unless name == 'undestroyable'
+        errors.add :base, "You can't destroy undestroyable things!"
+        throw(:abort)
+      end
     end
   end
 
@@ -145,6 +152,21 @@ describe Spree::Admin::WidgetsController, type: :controller do
 
     it 'destroys the resource' do
       expect { subject }.to change { Widget.count }.from(1).to(0)
+    end
+
+    context 'failure' do
+      let(:widget) { Widget.create!(name: 'undestroyable') }
+      let(:params) { { id: widget.id } }
+
+      context 'js format' do
+        subject { delete :destroy, params: params, format: 'js' }
+
+        it 'responds with error message' do
+          subject
+          expect(response).to be_unprocessable
+          expect(response.body).to eq assigns(:widget).errors.full_messages.join(', ')
+        end
+      end
     end
   end
 
--- a/backend/app/controllers/spree/admin/resource_controller.rb
+++ b/backend/app/controllers/spree/admin/resource_controller.rb
@@ -108,6 +108,7 @@ class Spree::Admin::ResourceController < Spree::Admin::BaseController
       invoke_callbacks(:destroy, :fails)
       respond_with(@object) do |format|
         format.html { redirect_to location_after_destroy }
+        format.js { render status: :unprocessable_entity, plain: @object.errors.full_messages.join(', ') }
       end
     end
   end


backend/spec/controllers/spree/admin/resource_controller_spec.rb
```diff
@@ -146,6 +153,21 @@ def model_class
...
+      let(:widget) { Widget.create!(name: 'undestroyable') }
+      let(:params) { { id: widget.id } }
+
+      context 'js format' do
+        subject { delete :destroy, params: params, format: 'js' }
+
+        it 'responses with error message' do
+          subject
+          expect(response).to be_unprocessable
+          expect(response.body).to eq assigns(:widget).errors.full_messages.join(', ')
```
kennyadsl: Since you have the error hardcoded at [line 39](https://github.com/solidusio/solidus/pull/2920/files#diff-d39c34ac52242598fd564473683289eaR39) maybe it's better to check the expected result so that we are super sure we are testing the right thing, for example if both `assigns(:widget).errors.full_messages` and `response.body` are empty we could have this test passing even if it's not what we expect.

My suggestion is:

```ruby
expect(response.body).to eq 'You cant destroy undestroyable things!'
```

bitberries: The point is that I want to test that the answer contains exactly validation errors same as in other resource controller's actions
For example:
https://github.com/solidusio/solidus/blob/518765fd145b7072d80c141503cea2cce484793f/backend/spec/controllers/spree/admin/resource_controller_spec.rb#L97

backend/spec/controllers/spree/admin/resource_controller_spec.rb
```diff
@@ -32,6 +32,13 @@ def model_class
     model do
       acts_as_list
       validates :name, presence: true
+      before_destroy :check_destroy_constraints
+
+      def check_destroy_constraints
+        return unless name == 'undestroyable'
+        errors.add :base, 'You cant destroy undestroyable things!'
```
jacobherrington: I'm not crazy about `You cant destroy undestroyable things!` I think there is a better way to phrase this.

We definitely need to add an apostrophe: `You can't destroy undestroyable things!`

Also, we probably want to add this to `en.yml` instead of just writing the string.

bitberries: This validation is only for testing as you can see. Adding it to en.yml is weird. But if you do not like the phrase itself, then you can offer absolutely any one to your taste.

jacobherrington: @bitberries That's fair, let's not worry about putting it in yaml then! I know it's picky, but I'd like to at least see the apostrophe :)

bitberries: @jacobherrington done

kennyadsl:
@bitberryru failures should go away if you rebase against master, thanks!

bitberryru:
> @bitberryru failures should go away if you rebase against master, thanks!

Thank you, done.

backend/spec/controllers/spree/admin/resource_controller_spec.rb
```diff
@@ -146,6 +153,21 @@ def model_class
...
     end
+
+    context 'failure' do
+      let(:widget) { Widget.create!(name: 'undestroyable') }
+      let(:params) { { id: widget.id } }
+
+      context 'js format' do
+        subject { delete :destroy, params: params, format: 'js' }
+
+        it 'responses with error message' do
```
aitbw: Can you change `responses` for `responds`? Thanks!

kennyadsl: @aitbw good catch!

bitberryru: Done

PR #2922, opened 2018-10-19
Remove dangerous query method warning

potomak:
Warning message:

    DEPRECATION WARNING: Dangerous query method (method whose arguments
    are used as raw SQL) called with non-attribute argument(s):
    "coalesce(spree_shipments.shipped_at, spree_shipments.created_at)
    desc". Non-attribute arguments will be disallowed in Rails 6.0.
    This method should not be called with user-provided values, such as
    request parameters or model attributes. Known-safe values can be
    passed by wrapping them in Arel.sql(). (called from block in
    <class:Shipment> at solidus/core/app/models/spree/shipment.rb:37)

SQL generated by the scope before this update:

    > Spree::Shipment.reverse_chronological.to_sql
     => "SELECT \"spree_shipments\".* FROM \"spree_shipments\" ORDER BY
      coalesce(spree_shipments.shipped_at, spree_shipments.created_at)
      desc, \"spree_shipments\".\"id\" DESC"

SQL generated by the scope after this update:

    > Spree::Shipment.reverse_chronological.to_sql
     => "SELECT \"spree_shipments\".* FROM \"spree_shipments\" ORDER BY
      COALESCE(\"spree_shipments\".\"shipped_at\",
      \"spree_shipments\".\"created_at\") DESC,
      \"spree_shipments\".\"id\" DESC"

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -33,7 +33,7 @@ module Spree
     scope :trackable, -> { where("tracking IS NOT NULL AND tracking != ''") }
     scope :with_state, ->(*s) { where(state: s) }
     # sort by most recent shipped_at, falling back to created_at. add "id desc" to make specs that involve this scope more deterministic.
-    scope :reverse_chronological, -> { order('coalesce(spree_shipments.shipped_at, spree_shipments.created_at) desc', id: :desc) }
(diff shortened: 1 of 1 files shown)


potomak:
The same _issue_ has been addressed by https://github.com/solidusio/solidus/pull/2921. The approach is slightly different, but both PRs fix this deprecation warning.

kennyadsl:
I'm sure this is a great way to fix the issue but I'm not an Arel expert so I'm more for the easier to understand solution provided in #2921 . But let's wait to see if we have other opinions here. 

In the meantime thanks for the contribution!

PR #2923, opened 2018-10-19
Guides improvements

kennyadsl:
This PR switches Guides package manager to yarn, improves the README and updates `randomatic`, a vulnerable package (that is not dangerous since we only use that in development).


kennyadsl:
I'm actually seeing a UI change in the header in desktop:

Before:

schermata 2018-10-19 alle 23 38 25

After:

schermata 2018-10-19 alle 23 38 33

I'll try to take a look soon but I have no idea what could be, but I ❤️Netlify previews even more. 🙂

kennyadsl:
Ok I think it's related to this issue on Twitter Bootstrap: https://github.com/twbs/bootstrap/pull/26820

PR #2925, opened 2018-10-24
Add translation for details attribute of LogEntry model

bitberries:
Hi guys,

Noticed place without i18n and invalid font-awesome icons. Fixed.


bitberries:
Also, In solidus_gateway I noticed these partials https://github.com/solidusio/solidus_gateway/tree/master/lib/views/backend/spree/admin/log_entries

But I did not find support of these partials in solidus. Seems that long time ago there was custom log entry rendering, in history I found this commit https://github.com/spree/spree/issues/5475

What you think about returning back this feature?

bitberries:
@jacobherrington Okay, I will open an issue for discuss :)

kennyadsl:
🎉

PR #2927, opened 2018-10-29
Remove duplicated CHANGELOG

yono:
https://github.com/solidusio/solidus/pull/2737 appears twice in CHANGELOG.

So I removed one line of the two.


yono:
@kennyadsl 

Thank you for your approval!

> How did you noticed that? 🙂

When I upgrade solidus version in our product, I read the version's whole CHANGELOG.
Now I trying to upgrade solidus version to 2.7, so I read the PR's CHANGELOG twice.

PR #2928, opened 2018-10-29
Remove user prereq from First Order promorule

fastjames:
When a customer applies a promotion with a "first order" rule, the rule
checks for a User or email associated with the order. This is not
strictly necessary, since the order can be checked after the coupon is
applied and if the user is found to have an existing order the coupon
gets removed then.

### DIFF
--- a/frontend/spec/features/first_order_promotion_spec.rb
+++ b/frontend/spec/features/first_order_promotion_spec.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.feature "First Order promotion" do
+  given!(:promotion) do
+    FactoryBot.create(
+      :promotion_with_first_order_rule,
+      :with_order_adjustment,
+      code: "FIRSTONEFREE",
+      per_code_usage_limit: 10
+    )
+  end
+
+  background do
+    create(:store)
+    product = FactoryBot.create(:product)
+    visit spree.root_path
+    click_link product.name
+    click_button "Add To Cart"
+  end
+
+  scenario "Adding first order promotion to cart and checking out as guest" do
+    fill_in "Coupon code", with: "FIRSTONEFREE"
+    click_button "Apply Code"
+    expect(page).to have_content("The coupon code was successfully applied to your order")
+
+    within("#cart_adjustments") do
+      expect(page).to have_content("-$10.00")
+    end
+  end
+
+  scenario "Trying to reuse first order promotion" do
+    previous_user = FactoryBot.create(
+      :user,
+      email: "sam@tom.com"
+    )
+    _previous_order = create(:completed_order_with_totals, user: previous_user)
+    fill_in "Coupon code", with: "FIRSTONEFREE"
+    click_button "Apply Code"
+    expect(page).to have_content("The coupon code was successfully applied to your order")
+    click_on "Checkout"
+    fill_in "Customer E-Mail", with: "sam@tom.com"
+    fill_in_address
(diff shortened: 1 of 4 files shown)


fastjames:
I can see a possible case for reintroducing the guard with a switch related to the config setting that requires login.

frontend/spec/features/first_order_promotion_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    create(:store)
+    product = FactoryBot.create(:product)
+    visit spree.root_path
+    click_link product.name
+    click_button "Add To Cart"
+  end
+
+  scenario "Adding first order promotion to cart and checking out as guest" do
+    fill_in "Coupon code", with: "FIRSTONEFREE"
+    click_button "Update"
```
aitbw: Now that #2958 is merged, you can use `click_button "Apply Code"` to take advantage of the new endpoints and avoid deprecation warnings :tada:

frontend/spec/features/first_order_promotion_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  scenario "Trying to reuse first order promotion" do
+    previous_user = FactoryBot.create(
+      :user,
+      email: "sam@tom.com"
+    )
+    _previous_order = create(:completed_order_with_totals, user: previous_user)
+    fill_in "Coupon code", with: "FIRSTONEFREE"
+    click_button "Update"
```
aitbw: Same as the previous comment

fastjames: Updated

frontend/spec/features/first_order_promotion_spec.rb
fastjames: Updated

fastjames:
@aitbw thanks for the feedback, I have made the requested changes and rebased against today's master.

frontend/spec/features/first_order_promotion_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+end
+
+def fill_in_address
+  address = "order_bill_address_attributes"
+  fill_in "#{address}_firstname", with: "Ryan"
+  fill_in "#{address}_lastname", with: "Bigg"
+  fill_in "#{address}_address1", with: "143 Swan Street"
+  fill_in "#{address}_city", with: "Richmond"
+  select "United States of America", from: "#{address}_country_id"
+  # select "Alabama", from: "#{address}_state_id"
```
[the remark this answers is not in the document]
fastjames: Agreed. Maybe at one point there was not a state selector on the test checkout, but I guess that's no longer the case.

frontend/spec/features/first_order_promotion_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    expect(page).to have_content("The coupon code was successfully applied to your order")
+    click_on "Checkout"
+    fill_in "Customer E-Mail", with: "sam@tom.com"
+    fill_in_address
+    click_on "Save and Continue"
+    expect(page).to_not have_content("#summary-order-charges")
+  end
+end
+
+def fill_in_address
```
spaghetticode: I think that, by defining this method here, you're polluting the `main` namespace. I think that this method can be moved inside the `Rspec.feature` block or, since this method is used only once here, it could be removed and the code inlined directly where it's used.

fastjames: Good point, I don't want to be a polluter. I have moved it into the feature block.

PR #2929, opened 2018-10-30
remove unused routes and actions from return authorizations

ccarruitero:
`add` and `receive` actions for this controller was removed in https://github.com/solidusio/solidus/commit/a8b51afb8e33a98b116eb88940ce0bde0aeec4be#diff-34985b492fae3d51c9ad1b6e583f06db and https://github.com/solidusio/solidus/commit/8ae1e15f7ca423518200488e97ca0bb928e66092#diff-34985b492fae3d51c9ad1b6e583f06db, so is not necessary keep in routes or controller


kennyadsl:
Thanks @ccarruitero ! 🎉

Issue #2926, opened 2018-10-26
Refund created for uncompleted payments


bitberry-dev:
### Steps to reproduce

1. Create order with check payment method.
2. Create refund in admin -> order -> payments.
![2018-10-26 17 55 34](https://user-images.githubusercontent.com/3278464/47574686-cfae9d00-d948-11e8-8f12-5e3a015424a9.png)

### Expected behavior

I expect error when creating refund from uncompleted payment.
Also do not show refund icon for uncompleted payments and do not run refunds#new action.

### Actual behavior

Refund created for payment in "invalid" state.
![2018-10-26 17 55 58](https://user-images.githubusercontent.com/3278464/47574691-d5a47e00-d948-11e8-8935-02cb0df0750c.png)

### System configuration
**Solidus Version**:

"2.8.0.alpha"
master, latest commit 622aa25e6c351c4de16af87e9cba70de596ad959

Is this the right behavior? If not I can fix it.
Also refund model doesn't validate amount on updates.

kennyadsl:
I don't think it is expected, it does not make sense to refund a payment that never happened. Feel free to submit a patch, thanks!

jarednorman:
This was half-fixed by #6094.

PR #2932, opened 2018-11-05
Fix issue refunding uncompleted payments

jtapia:
Issue: https://github.com/solidusio/solidus/issues/2926

#### Includes
- Remove `Refund` button for uncompleted payments
- Remove `Credit` button for payment on `void` and `invalid` state

### DIFF
--- a/backend/app/views/spree/admin/payments/_list.html.erb
+++ b/backend/app/views/spree/admin/payments/_list.html.erb
@@ -56,12 +56,15 @@
               <%= link_to_with_icon 'edit', t('spree.actions.edit'), nil, no_text: true, class: "js-edit", data: {action: 'edit'} %>
             <% end %>
             <% allowed_actions = payment.actions.select { |a| can?(a.to_sym, payment) } %>
+            <%= allowed_actions %>
             <% allowed_actions.each do |action| %>
               <% if action == 'credit' %>
+                <% next if payment.invalid? || payment.failed? %>
                 <%= link_to_with_icon 'reply', t('spree.refund'), new_admin_order_payment_refund_path(@order, payment), no_text: true %>
-              <% elsif action == 'capture' && !@order.completed? %>
-                <%# no capture prior to completion. payments get captured when the order completes. %>
               <% else %>
+                <% next if action == 'capture' && !@order.completed? %>
+                <% next if action == 'void' && (payment.invalid? || payment.failed?) %>
+
                 <%= link_to_with_icon action, t(action, scope: 'spree'), fire_admin_order_payment_path(@order, payment, e: action), method: :put, no_text: true, data: {action: action} %>
               <% end %>
             <% end %>


backend/app/views/spree/admin/payments/_list.html.erb
```diff
@@ -58,10 +58,12 @@
             <% allowed_actions = payment.actions.select { |a| can?(a.to_sym, payment) } %>
             <% allowed_actions.each do |action| %>
               <% if action == 'credit' %>
+                <% next unless payment.completed? %>
```
ccarruitero: I think will be better add the extra checking with the conditional

backend/app/views/spree/admin/payments/_list.html.erb
```diff
@@ -57,12 +57,14 @@
             <% end %>
             <% allowed_actions = payment.actions.select { |a| can?(a.to_sym, payment) } %>
             <% allowed_actions.each do |action| %>
-              <% if action == 'credit' %>
+              <% if action == 'credit' && !payment.completed? %>
```
jtapia: @ccarruitero what do you think?

jtapia:
@kennyadsl sure, checking it

backend/app/views/spree/admin/payments/_list.html.erb
spaghetticode: I think that the condition should be a bit different Now it's verified when the action is `credit` and the payment is not completed. I think it should be:

```<% if action == 'credit' && payment.completed? %>```

as we want to show the button only when the payment is `complete`

File: backend/app/views/spree/admin/payments/_list.html.erb [hunk 3]
```diff
@@ -57,12 +57,14 @@
...
                 <%= link_to_with_icon 'reply', t('spree.refund'), new_admin_order_payment_refund_path(@order, payment), no_text: true %>
               <% elsif action == 'capture' && !@order.completed? %>
                 <%# no capture prior to completion. payments get captured when the order completes. %>
               <% else %>
-                <%= link_to_with_icon action, t(action, scope: 'spree'), fire_admin_order_payment_path(@order, payment, e: action), method: :put, no_text: true, data: {action: action} %>
+                <% if payment.void? || payment.invalid? %>
```
spaghetticode: I think this `if` condition is responsible for most (if not all) the failing specs, I think it should be different.

What's happening now is that, when all previous conditions are not met, then the action link is printed only if the payment is `void` or `invalid`.

I guess that among these actions that are filtered out by the conditional there's something that should be printed anyway. These actions come from this line:

```
<% allowed_actions = payment.actions.select { |a| can?(a.to_sym, payment) } %>
```

jtapia: @spaghetticode I just changed the approach, please let me know what do you think

spaghetticode: @jtapia sorry for the late feedback. I think it now may work properly. 

But I noticed there's no spec for this PR, and I think at some point in the past they were present, as some comments suggest. Not sure what happened to them, maybe they got lost during some rebase?

backend/app/views/spree/admin/payments/_list.html.erb
```diff
@@ -56,12 +56,16 @@
               <%= link_to_with_icon 'edit', t('spree.actions.edit'), nil, no_text: true, class: "js-edit", data: {action: 'edit'} %>
             <% end %>
             <% allowed_actions = payment.actions.select { |a| can?(a.to_sym, payment) } %>
+            <%= allowed_actions %>
             <% allowed_actions.each do |action| %>
-              <% if action == 'credit' %>
+              <% case action %>
```
spaghetticode: I prefer `case` to `if` clauses when there are multiple `when`, but maybe in this context (only one `when`) `if` would be a nicer choice? 

My preference would go to `if` because while they both work fine here, using `case` produces 2 lines of ERB (maybe it's just me, but the less ERB the better 😄) when one suffices with `if`.

kennyadsl:
Closing this one since specs are failing and there has not been activity for a while. Thanks for the contribution anyway.

fthobe:
@kennyadsl do you have any idea if this has been fixed in the meantime?

jtapia:
@kennyadsl I ran the tests locally and didn't found any failures. Is ok if I add test cases for the changes and reopen it?

fthobe:
@jtapia i think everybody would really appreciate it!

PR #2933, opened 2018-11-06
improve decorators / contributing docs

elia:
A couple of fixes to docs. I'd also like to address this comment:

```html
<!-- TODO:
   An article about decorators doesn't really belong in the `/extensions`
   directory. In the future, there will be a better home for it.
 -->
```

but I need to know if there's any plan to do that or what would be the desired new home for the decorators docs.

My proposal would be something like *Local Changes* or *Patching*, which would work for both the general application using Solidus directly and the extension developer adding functionality.

Also if there's any collective wisdom to share, would be great to add some guidance on how to patch templates (erb/jbuilder/ecc.) 😄.

### DIFF
--- a/guides/source/developers/extensions/decorators.html.md
+++ b/guides/source/developers/extensions/decorators.html.md
@@ -13,15 +13,13 @@ For example, if you want to add a method to the `Spree::Order` model, you could
 create `/app/models/mystore/order_decorator.rb` with the following contents:
 
 ```ruby
-module MyStore
-  module OrderDecorator
-    def total
-      super + BigDecimal(10.0)
-    end
+module MyStore::OrderDecorator
+  def total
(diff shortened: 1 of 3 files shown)


guides/source/developers/extensions/decorators.html.md
```diff
@@ -13,15 +13,13 @@ For example, if you want to add a method to the `Spree::Order` model, you could
...
-      super + BigDecimal(10.0)
-    end
+module MyStore::OrderDecorator
+  def total
+    super + BigDecimal(10.0)
   end
-end
 
-Spree::Order.prepend MyStore::OrderDecorator
+  Spree::Order.prepend self
```
kennyadsl: I alway use this syntax, which avoid useless duplication so 👍

PR #2939, opened 2018-11-11
Add ordering to menu items

jacobherrington:
This change allows menu items to be ordered. If the order value is nil
the menu item will come at the end of the list.

After:
b63a27f7-a7f5-449f-b233-3c4d964d3bd9

### DIFF
--- a/backend/app/models/spree/backend_configuration.rb
+++ b/backend/app/models/spree/backend_configuration.rb
@@ -22,7 +22,9 @@ module Spree
 
     # An item which should be drawn in the admin menu
     class MenuItem
-      attr_reader :icon, :label, :partial, :condition, :sections, :url, :position
+      attr_reader :icon, :label, :partial, :condition, :sections, :url
+
+      attr_accessor :position
 
       # @param sections [Array<Symbol>] The sections which are contained within
       #   this admin menu section.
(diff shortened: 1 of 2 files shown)


kennyadsl:
Hey I think it's a good idea but how can the menu be reordered by a user? For example what should I do in my app to put a new entry after the `Products` item? What about an overridable method that returns the order of the items instead?

jacobherrington:
@tvdeyen Changes made, does that work for you?

@kennyadsl I added some comments to explain how the position attribute should be used. In most cases, the value doesn't need to be present and the behavior will be exactly the same as it currently is in Solidus. However, if someone wants to alter the ordering they simply add or change the position attribute to match the order they would like the menu to display. Does that make sense, or do I need further changes?

kennyadsl:
@jacobherrington I got how you can add new items but I still don't get how you can change existing (in solidus backend) menu items positions into your own app, maybe it's simpler than what I think but can you please provide a simple example?

jacobherrington:
@kennyadsl for example `Spree::Backend::Config.menu_items.find { |i| i.label == :promotions }.position = 0` would make promotions first. Do you think there is a better way to achieve this?

jacobherrington:
@kennyadsl @tvdeyen please let me know if you have any more thoughts on this one!

PR #2940, opened 2018-11-11
Use filter over search for filter options

jacobherrington:
Search seems like wrong word for this section

screenshot_2018-11-10 orders

### DIFF
--- a/backend/app/views/spree/admin/orders/index.html.erb
+++ b/backend/app/views/spree/admin/orders/index.html.erb
@@ -10,7 +10,7 @@
 <% end if can? :create, Spree::Order %>
 
 <% content_for :table_filter_title do %>
-  <%= t('spree.search') %>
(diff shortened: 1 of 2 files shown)


kennyadsl:
Thanks! In the screenshot provided I can see `Filter Result` as button's text but from the code it looks like you used another i18n key (`spree.filter`). Is that just an issue with the screenshot provided?

jacobherrington:
@kennyadsl I'm not sure I understand your question. I didn't change the button text, rather the section heading text from `Search` to `Filter`. Because this is not a conventional search, instead it's a set of filters to narrow results.

@tvdeyen I think `Search` really might be incorrect here because it's not a search. It's a series of filters. Your call though 👍

kennyadsl:
@jacobherrington sorry, I thought it was the button. 😄

PR #2943, opened 2018-11-11
Standardize capitalization

jacobherrington:
I went through and tried to clean up some of the more inconsistent sets of capitalization. This may not be perfect, but it's much more consistent and adds some polish to the UX.


jacobherrington:
Just realized this will probably break some UI tests. 😆 I'll try to make those changes where necessary.

kennyadsl:
Maybe you could squash all commits into a single one with a better description?

jacobherrington:
@kennyadsl done!

PR #2944, opened 2018-11-12
Rename solidus admin

trottomv:
I doubt whether to keep the old "backend" directory for projects in production


kennyadsl:
Hey thanks @trottomv!

A couple of things:

- [Solidus Support `backend_available?` method](https://github.com/solidusio/solidus_support/blob/master/lib/solidus_support.rb#L48-L50): I think we need to deprecate this method since it should now be `admin_available?`  This is used into extensions [like here](https://github.com/solidusio/solidus_auth_devise/blob/master/lib/spree/auth/engine.rb#L69)
- Lot of extensions are injecting assets expecting the backend directory exists, can we fix this somehow in your opinion? Different examples of its usage [here](https://github.com/solidusio/solidus_auth_devise/blob/master/lib/spree/auth/engine.rb#L29-L30) and [here](https://github.com/solidusio/solidus_multi_domain/blob/master/lib/generators/solidus_multi_domain/install/install_generator.rb#L7)

tvdeyen:
> Lot of extensions are injecting assets expecting the backend directory exists, can we fix this somehow in your opinion? Different examples of its usage here and here

Good point Alberto!

But it's not only appending to existing files. It's also about adding new files into folders for a) overwriting or b) introducing new templates and partials.

This change is huge in terms of its impact and should not be done before 3.0. 

We should deprecate it first and then make the final move.

tvdeyen:
The move from `factory_girl` to `factory_bot` could be an example on how to make this transition.

kennyadsl:
Closing this one, thanks anyway but we are not able to change the name of this component at this time.

Issue #2952, opened 2018-11-14
StockLocation is missing `acts_as_list`


rymai:
### Steps to reproduce

Create two stock locations and try to reorder them.

### Expected behavior

The reordering update the stock locations' `position` attribute.

### Actual behavior

It doesn't work, and the following error is raised:

```
Started POST "/admin/stock_locations/update_positions" for ::1 at 2018-11-14 23:27:28 +0100
Processing by Spree::Admin::StockLocationsController#update_positions as JSON
  Parameters: {"positions"=>{"1"=>"1", "2"=>"2"}}
  Spree::User Load (0.9ms)  SELECT  "spree_users".* FROM "spree_users" WHERE "spree_users"."deleted_at" IS NULL AND "spree_users"."id" = $1 ORDER BY "spree_users"."id" ASC LIMIT $2  [["id", 96], ["LIMIT", 1]]
  Spree::Role Load (0.6ms)  SELECT "spree_roles".* FROM "spree_roles" INNER JOIN "spree_roles_users" ON "spree_roles"."id" = "spree_roles_users"."role_id" WHERE "spree_roles_users"."user_id" = $1  [["user_id", 96]]
   (0.3ms)  BEGIN
  Spree::StockLocation Load (0.4ms)  SELECT  "spree_stock_locations".* FROM "spree_stock_locations" WHERE "spree_stock_locations"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
   (0.2ms)  ROLLBACK
Completed 500 Internal Server Error in 37ms (ActiveRecord: 2.5ms)

NoMethodError (undefined method `set_list_position' for #<Spree::StockLocation:0x00007fba41b60050>):
  
activemodel (5.2.1) lib/active_model/attribute_methods.rb:430:in `method_missing'
solidus_backend (2.7.0) app/controllers/spree/admin/resource_controller.rb:79:in `block (2 levels) in update_positions'
actionpack (5.2.1) lib/action_controller/metal/strong_parameters.rb:338:in `block in each_pair'
actionpack (5.2.1) lib/action_controller/metal/strong_parameters.rb:337:in `each_pair'
actionpack (5.2.1) lib/action_controller/metal/strong_parameters.rb:337:in `each_pair'
solidus_backend (2.7.0) app/controllers/spree/admin/resource_controller.rb:78:in `block in update_positions'
[...]

PR #2953, opened 2018-11-14
Add acts_as_list to Spree::StockLocation

rymai:
Fixes https://github.com/solidusio/solidus/issues/2952.

/cc @octave

### DIFF
--- a/core/app/models/spree/stock_location.rb
+++ b/core/app/models/spree/stock_location.rb
@@ -7,6 +7,8 @@ module Spree
   class StockLocation < Spree::Base
     class InvalidMovementError < StandardError; end
 
+    acts_as_list
+
     has_many :shipments
     has_many :stock_items, dependent: :delete_all, inverse_of: :stock_location
     has_many :cartons, inverse_of: :stock_location


rymai:
Actually I realized that it doesn't look so... I bumped into this issue when trying to reorder the stock locations in the Admin (the UI still allows to do that, so maybe we should just remove that...).

That also triggers a more interesting discussion: my original thought was that the stock locations ordering would be used in the simple shipment coordinator, but it's not. At least I think we should change `@stock_locations = Spree::StockLocation.active` to `@stock_locations = Spree::StockLocation.order_default.active` there... but that's another discussion. :D

kennyadsl:
I think you want to look at this recently merged PR #2783 that would allow to use the `position` if needed creating and using a new StockLocation sorter class. I think we need to merge this PR anyway since there's a bug in the UI. Thanks again!

rymai:
> I think you want to look at this recently merged PR #2783 that would allow to use the `position` if needed creating and using a new StockLocation sorter class.

@kennyadsl Thanks for the pointer! That will fulfill my need perfectly! Looking forward for the next version! :)

rymai:
@tvdeyen Done, thanks! 🎉

tvdeyen:
Thank you for your contribution

PR #2955, opened 2018-11-15
Add script to skip CircleCI on guides

jacobherrington:
This is based on the discussion here: https://github.com/Moya/Moya/issues/1286

I'm not super familiar with Circle, but if this doesn't work for some obvious reason feel free to close.


aitbw:
:thinking: but you can skip CI builds by adding `[ci skip]` to the commit message, [here's more information about it](https://circleci.com/docs/2.0/skip-build/)

jacobherrington:
@aitbw Yes, that is correct, but this is automated. Not all who wish to contribute to the docs will know to use `[ci skip]` or `[skip ci]` in the commit message.

jacobherrington:
Not to mention that I'd just like to hear opinions on this idea. There may be a very good reason to run CI on doc changes. 🤷‍♂️

.circleci/config.yml
```diff
@@ -16,6 +16,7 @@ jobs:
           POSTGRES_USER: root
     parallelism: &parallelism 3
     steps: &steps
+      - run: skip-ci.sh && circleci step halt || true # skip Circle for guides
```
kennyadsl: Is it possible to have the script inline without the need to add another script file? I personally prefer a less readable line than a new file in the project that is used in a single point.

jacobherrington:
Unfortunately this isn't a CircleCI feature, though I wish it was... It requires some creative thinking to solve. I think I could probably figure out a way to do this inside of the CircleCI config, but I'm not sure I see the benefit of doing it that way.

I will see if I can find a way to do that, but it might be simpler to stick with this implementation. 

Would it help if we created a `scripts/` directory to put things like this? `build.sh` could live there as well.

.circleci/config.yml
elia: A good place for the script could be `bin/circleci-halt-for-doc-only-changes` or `.circleci/halt-for-doc-only-changes` and move everything inside it. That way what it does should be self-evident and well contained.

jacobherrington: I like that suggestion @elia do you agree with this @kennyadsl?

kennyadsl: Yeah, `.circleci/bin/halt-for-doc-only-changes.sh` would be my preference

jacobherrington:
@kennyadsl Done!

kennyadsl:
@jacobherrington sorry, I meant squash commits 😬

jacobherrington:
Done @kennyadsl 👍

PR #2956, opened 2018-11-15
Added a link to the frontend product from the backend product edit page

seand7565:
It's handy to be able to quickly get to the frontend product page from the backend product edit page.

### DIFF
--- a/backend/app/views/spree/admin/products/edit.html.erb
+++ b/backend/app/views/spree/admin/products/edit.html.erb
@@ -4,6 +4,11 @@
       <%= link_to t('spree.new_product'), new_object_url, id: 'admin_new_product', class: 'btn btn-primary' %>
     </li>
   <% end %>
+  <% if defined?(Spree::Frontend::Engine) %>
+    <li id="view_product_link">
+      <%= link_to t('spree.view_product'), spree.product_path(@product), id: 'admin_view_product', class: 'btn btn-primary ml-2' %>
+    </li>
+  <% end %>
 <% end %>
 
 <%= render partial: 'spree/admin/shared/product_tabs', locals: { current: 'Product Details' } %>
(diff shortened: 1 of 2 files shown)


jacobherrington:
Could we add some padding between the buttons?
image

Adding `<%= link_to t('spree.new_product'), new_object_url, id: 'admin_new_product', class: 'btn btn-primary mr-2' %>` makes it a bit better, don't you think?
image

jacobherrington:
There is also a danger that we are assuming the individual is using both `solidus_frontend` and solidus_backend`. I think a fair number of people choose to forgo `solidus_frontend`.

seand7565:
@jacobherrington Very good point about people not using frontend. Any suggestions on how to handle that? The [admin navigation footer](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/shared/_navigation_footer.html.erb) wraps the link to the frontend in `if spree.respond_to? :root_path` - a similar approach should work here, right?

jacobherrington:
@seand7565 That is something I had thought of as well, the question is: Do we want to perpetuate that pattern? I think that's a call for the core team.

tvdeyen:
I like the idea, but we need to address stores not having the frontend installed. `solidus_support` uses  `defined?(Spree::Frontend::Engine)`

kennyadsl:
I like this but maybe it's time to port [these methods](https://github.com/solidusio/solidus_support/blob/master/lib/solidus_support.rb#L44-L54) into `solidus_core`?

kennyadsl:
> Or add `solidus_support` as runtime dependency?

I don't think this could work as is since `solidus_support` already depends on `solidus_core`.

jacobherrington:
> We could assume that extensions or shops that install solidus_support also will have solidus_core installed, no?

@tvdeyen Yes. `solidus_support` depends on `solidus_core` so a site using `solidus_support` without `solidus_core` wouldn't work anyway.

PR #2957, opened 2018-11-16
Add a button to hide nav

jacobherrington:
Video: https://www.useloom.com/share/2a72f50345c74fb095c58b939683f165

This is a really, super basic way to hide the nav. Happy to take input from anyone with some UI knowledge on animations, etc.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/components/admin_nav.js
+++ b/backend/app/assets/javascripts/spree/backend/components/admin_nav.js
@@ -12,12 +12,14 @@ Spree.ready(function() {
     adminNav.style.display = "none";
   }
 
-  adminNavToggle.addEventListener("click", function() {
-    if (document.body.classList.contains("admin-nav-extended")) {
-      showNav();
-    } else {
-      hideNav();
-    }
-    document.body.classList.toggle("admin-nav-extended");
-  });
+  if (adminNavToggle) {
+    adminNavToggle.addEventListener("click", function() {
+      if (document.body.classList.contains("admin-nav-extended")) {
+        showNav();
+      } else {
+        hideNav();
+      }
+      document.body.classList.toggle("admin-nav-extended");
+    });
+  }
 });
--- a/backend/app/assets/javascripts/spree/backend/components/admin_nav.js
+++ b/backend/app/assets/javascripts/spree/backend/components/admin_nav.js
@@ -1,0 +1,23 @@
+Spree.ready(function() {
+  var adminNavToggle = document.querySelector("#admin-nav-toggle");
+  var adminNav =  document.querySelector(".admin-nav");
+
+  function showNav() {
+    adminNavToggle.innerHTML = '<i class=\"fa fa-chevron-left fa-2x\">';
+    adminNav.style.display = "block";
+  }
+
+  function hideNav() {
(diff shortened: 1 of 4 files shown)


backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -17,6 +24,8 @@ body {
   @media print { display: none }
 }
 
+.admin-nav-extended { padding-left: 0; }
+
```
tvdeyen: Here you could add all the CSS magic for showing and hiding the nav bar. For instance:

```scss
.admin-nav {
  transform: translate3d(0, 0, 0);
  transition: transform 350ms ease-in-out;
}

#admin-nav-toggle . fa-chevron-right {
  transform: rotate(0);
  transition: transform 350ms linear;
}

.admin-nav-extended { 
  padding-left: 0; 

  .admin-nav { 
    transform: translate3d(-100%, 0, 0);
  }

  #admin-nav-toggle .fa-chevron-right {
    transform: rotate(180deg);
  }
}
```

backend/app/views/spree/admin/shared/_navigation.html.erb
```diff
@@ -1,3 +1,4 @@
+<span id="admin-nav-toggle"><i class="fa fa-chevron-left fa-2x"></i></span>
```
tvdeyen: Instead of changing the innerHTML of the toggle icon in JS we could also use some CSS magic to rotate it if we toggle the body class.

backend/app/assets/javascripts/spree/backend/components/admin_nav.js
```diff
@@ -0,0 +1,25 @@
+Spree.ready(function() {
+  var adminNavToggle = document.querySelector("#admin-nav-toggle");
+  var adminNav =  document.querySelector(".admin-nav");
+
+  function showNav() {
+    adminNavToggle.innerHTML = '<i class=\"fa fa-chevron-left fa-2x\">';
```
tvdeyen: I don't think we need this change of innerHTML if we use CSS to transform the icon

jacobherrington:
@tvdeyen thoughts on this? https://www.useloom.com/share/9270e32141c14ef1baf7e34d6473401c

tvdeyen:
Much better, thanks. Could you try to make the toggle as high as the header and the icon half the size, so it fits the rest of the icons?

jacobherrington:
@tvdeyen What do you mean by as high as the header? Do you want it to be level with another element or taller? I can make the change, just want to make sure I understand.

tvdeyen:
As high as the logo and the breadcrumb.

jacobherrington:
@kennyadsl does this solve that issue for you? I'm not sure it's the best way, but I'm not sure how I should accomplish this in vanilla JS.

jacobherrington:
@tvdeyen I actually agree with everything you have listed here. I am glad to hear there is interest in implementing this in a more correct way. 

As long as we are comfortable with a more significant UI change, I agree that we should move towards something like you have shared. I will approach this with your thoughts in mind. I am going to open an issue to get some further discussion from other members of the community.

PR #2959, opened 2018-11-20
Add no other promotions rule

jacobherrington:
This is a promotion rule that seems like it should ship with Solidus. The primary reason I feel this way is because it requires modifying the Spree::PromotionHandler::Coupon class.

Another solution might be extracting `promotion_stacking_disabled?` into some kind of flag on the promotion model instead of a method that is less than efficient.

### DIFF
--- a/backend/app/views/spree/admin/promotions/rules/_first_order.html.erb
+++ b/backend/app/views/spree/admin/promotions/rules/_first_order.html.erb
@@ -1,0 +1,1 @@
+<%# Don't delete this file! It is required for a promotion rule %>
--- a/backend/app/views/spree/admin/promotions/rules/_no_other_promotion.html.erb
+++ b/backend/app/views/spree/admin/promotions/rules/_no_other_promotion.html.erb
@@ -1,0 +1,1 @@
+<%# Don't delete this file! It is required for a promotion rule %>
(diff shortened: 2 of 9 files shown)


jacobherrington:
I'd love to hear any constructive feedback on this idea.

ericsaupe:
I like this idea but I'm afraid of how it would work. The way it's currently written the NoOtherPromotions promo wouldn't apply if other promotions were already there. What if that was a better promotion than what is currently applied? AFAIK the default frontend doesn't allow users to manage promos applied to an order which would entirely lock the user out of using this promo. Should the application of this promo remove other promos if better?

Not really sure what the best path is just thinking out loud.

jacobherrington:
@ericsaupe I don't disagree with you. However, I do think there should be an option for stores who don't want coupon stacking of any kind and this is a rather generic implementation of that which can be extended for more specific use cases (e.g. picking the best coupon).

I would also really like if there was a global config option that disabled coupon stacking on ALL coupons, but that is a little less granular than this method.

Just to clarify, the way it works, a promotion with this rule could not be applied if another promotion was already applied AND no other other promotion could be applied after it was applied. So it is only possible to use such a promotion as the ONLY promotion.

jacobherrington:
I think it's worth revisiting this later, but I'll leave the PR open to gather thoughts.

jacobherrington:
@ericsaupe we put this in production and we are having some trouble with it. I'd advise not merging this until I can revisit it and do something more robust.

PR #2963, opened 2018-11-21
Add i18n-tasks and normalize translations

afdev82:
This PR adds the `i18n-tasks` gem to check the translations and reorders them like how is already done in the `solidus_i18n` project.

In this way it's easier to contribute, because the translations are in the same position.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ end
 
 gem 'database_cleaner', '~> 1.3', require: false
 gem 'factory_bot_rails', '~> 4.8', require: false
+gem 'i18n-tasks', '~> 0.9', require: false
 gem 'rspec-activemodel-mocks', '~>1.0.2', require: false
 gem 'rspec-rails', '~> 3.7', require: false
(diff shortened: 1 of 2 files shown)


afdev82:
I just run `bundle exec i18n-tasks normalize` inside the `core` directory

kennyadsl:
@afdev82 Hey can you please rebase on master since in the meantime some en key changed? I think it could be easier to revert changes and re-run the script at this point since there are a lot of conflicts. Thanks!

afdev82:
@kennyadsl yes, I will do it soon ;)

afdev82:
maybe we could add `bundle exec i18n-tasks health` to the ci, we do the same for our projects and it's useful

tvdeyen:
> maybe you could add bundle exec i18n-tasks health to the ci, we do the same for our projects and it's useful

This is probably a good idea, but I don't know how this will impact the current state of translations. We might want to check the health first, fix all issues and than enable it.

kennyadsl:
Thanks!

PR #2965, opened 2018-11-22
[WIP] Add option for user to store creditcard

peterberkenbosch:
following up on #1584 (finally) this PR will update the temp cc and introduce an option for users to save the creditcard profile in the wallet. 

- [ ] think about subscriptions and how we can provide a hook for non stored payment options

### DIFF
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -30,6 +30,7 @@ module Spree
       if update_order
 
         assign_temp_address
+        assign_temp_payment_source
 
         unless transition_forward
           redirect_on_failure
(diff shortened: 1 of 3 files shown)


spaghetticode:
@peterberkenbosch thank you for this PR 👏 

Can you confirm this PR is ready for review? I see in the description there's an unchecked checkbox, and also this [comment](https://github.com/solidusio/solidus/issues/3003#issuecomment-447906821) suggests me that it's not... am I right?

jarednorman:
@peterberkenbosch What's the status of this PR? Any plans to continue with this work?

peterberkenbosch:
Have not looked into this again. Not sure if this is still valid with the current versions. Will look and update.

waiting-for-dev:
Closing as, if we still need it, it should go now to solidus_starter_frontend. Please, feel free to move it there, and thanks again for your collaboration.

PR #2967, opened 2018-11-22
[WIP] Add payment redirect support for checkout

peterberkenbosch:
`CheckoutController` never had the concept of redirecting the user to the "offsite" payment method site. This situation was handled in various bad ways in the past, mostly with the
substitution of the checkout submit button with a link tag, using some JS.

This PR adds the possibility to redirect if the active payment wants to redirect to its site, before proceeding to the next checkout step.

Based on the work Alessandro Lepore did for solidus_paybright

- [ ] add guides samples
- [ ] add configuration for base callback webhooks

### DIFF
--- a/core/app/models/spree/payment.rb
+++ b/core/app/models/spree/payment.rb
@@ -159,6 +159,11 @@ module Spree
       payment_method.try!(:store_credit?)
     end
 
+    # @return [String] the (computed) redirection url for offsite payment. returning nil for onsite payment methods
+    def redirect_url
+      payment_method&.redirect_url(self)
+    end
+
     private
 
     def source_actions
--- a/core/app/models/spree/payment_method.rb
+++ b/core/app/models/spree/payment_method.rb
@@ -265,6 +265,11 @@ module Spree
       is_a? Spree::PaymentMethod::StoreCredit
     end
 
+    # provides hook for redirecting to offsite payment method.
+    def redirect_url(_payment)
+      nil
+    end
+
     protected
 
     # Represents the gateway class of this payment method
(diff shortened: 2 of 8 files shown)


kennyadsl:
Hey thanks @peterberkenbosch !

In my experience these kind of gateways also have a callback URL (or more, like one for success and one for failure) where they will land after the payment offsite is done.

With this callback the payment is registered in solidus and the user is redirected to the next step (or still on payment if there's an error).

Do you think this logic should be implemented separately into each gateway that should add its own routes/controllers or is this something that we can generalize and add support for that into core?

peterberkenbosch:
Thanks @kennyadsl, 

My approach was to have the most minimal support needed to have offsite payment providers. We could for sure explore a callback controller base class that will lighten the load for those implementations for sure. Will post some more code here soonish with that to discuss further.

core/app/models/spree/payment.rb
```diff
@@ -196,6 +196,11 @@ def store_credit?
       payment_method.try!(:store_credit?)
     end
 
+    # @return [String] the (computed) redirection url for offsite payment. returning nil for onsite payment methods
+    def redirect_url
+      payment_method.redirect_url(self)
```
spaghetticode: Just a minor annoyance, but I think that `Spree::Payment.new.redirect_url` would raise an error, as `payment_method` is `nil`

peterberkenbosch: that is correct, the redirect_url needs a payment method, we could use a safe operator here to just return nil instead of throwing an error though.

jarednorman:
@peterberkenbosch Is this being actively worked on? We're trying to pin down what belongs in our upcoming milestones.

peterberkenbosch:
@jarednorman yes, I started rebasing master on this one this morning actually.

elia:
@peterberkenbosch I think this should probably be reissued on https://github.com/solidusio/solidus_starter_frontend, feel free to just close it or turn it into an issue on SSF, leaving it open so you handle it the way you prefer 👍

kennyadsl:
@peterberkenbosch closing this. Feel free to reopen if you want to continue this nice work.

Issue #2950, opened 2018-11-14
Admin Navigation Tab Double Highlighting


jacobherrington:
### Steps to reproduce

In the admin area, navigate to a product and then click "Product Stock."

### Expected behavior

On the nav, the Stock tab should be highlighted.

### Actual behavior

Both the Product and Stock navigation tabs are highlighted.
![screenshot_2018-11-13 store stock - ruby on rails baseball jersey - products](https://user-images.githubusercontent.com/11466782/48492149-c0f03300-e7ee-11e8-9c44-d40a7bca3dd8.png)

### System configuration
**Solidus Version**:

2.8.0.alpha

**Extensions in use**:
N/A

aitbw:
I'll have a look at this one :)

PR #2968, opened 2018-11-22
Fix double highlight on admin menu when Product Stock tab is active

aitbw:
Refs: #2950

Here's a screenshot to demonstrate how the menu looks after the proposed patch

screenshot_2018-11-22 store stock - apache baseball jersey - products

Open to suggestions! :raised_hands:

### DIFF
--- a/backend/app/helpers/spree/admin/navigation_helper.rb
+++ b/backend/app/helpers/spree/admin/navigation_helper.rb
@@ -77,6 +77,11 @@ module Spree
         if options[:css_class]
           css_classes << options[:css_class]
         end
+
+        if params[:product_slug].present? && link.include?('Stock')
+          css_classes.delete('selected')
+        end
+
         content_tag('li', link + (yield if block_given?), class: css_classes.join(' ') )
       end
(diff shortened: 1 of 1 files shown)


jacobherrington:
@tvdeyen I agree, if it seems like there are more examples of this problem, but if it is only on the product stock page this solves that issue.

jacobherrington:
@aitbw let's try to approach the problem from another perspective. Instead of removing the class from the incorrect element, we should make sure it never gets applied. Does that make sense?

tvdeyen:
This is my alternative approach https://github.com/solidusio/solidus/pull/2969 to this issue. 

@aitbw thanks for working on this and sorry for the rejection, but we would like to fix the issue at the root cause. Having to deal with issue like this in a helper is not ideal. That's why I made this alternative approach. Thanks for the contribution. That's very much appreciated.

aitbw:
Don't worry @jacobherrington @tvdeyen, I completely understand and I'm more fond of a solution like #2969 :relaxed:

PR #2969, opened 2018-11-22
Fix "Stock" admin nav double highlight

tvdeyen:
Without setting `match_path` the `tab` helper uses the current controller name as criteria to activate the tab in the menu. Since the stock items controller can be accessed for a single product or all products we need to make sure it only matches the `/admin/stock_items` path for the "Stock" menu tab.

store stock - apache baseball jersey - products 2018-11-22 22-05-07

### DIFF
--- a/backend/app/models/spree/backend_configuration.rb
+++ b/backend/app/models/spree/backend_configuration.rb
@@ -115,6 +115,7 @@ module Spree
           condition: -> { can?(:admin, Spree::StockItem) },
           label: :stock,
           url: :admin_stock_items_path,
+          match_path: '/stock_items',
           position: 3
         ),
         MenuItem.new(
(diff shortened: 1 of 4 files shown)


backend/app/models/spree/backend_configuration.rb
```diff
@@ -46,7 +46,8 @@ def initialize(
         label: nil,
         partial: nil,
         url: nil,
-        position: nil
+        position: nil,
+        match_path: nil
```
jacobherrington: Would you mind expanding the above comment to explain what `match_path` is used for?

andyyang: Could I help to add the comment?

tvdeyen:
@jacobherrington done

backend/app/models/spree/backend_configuration.rb
jacobherrington: @andyyang We would love to have you help out! I think Thomas has this covered. Would you take a look at https://github.com/solidusio/solidus/issues for another issue? Thanks!

PR #2971, opened 2018-11-23
Move script files to bin + documentation

elia:
- ➡️ Using `bin/` for internal utils is the default for both Rails and Bundler.
The script has now no extension and is marked as executable improving its
_usability_ (the user doesn't need to know how to deal with each extension) and
_maintainability_ (it's easy to switch to a ruby script just changing the
shebang line).
- 📝  The doc is updated and env variables are passed with `env` making life easier
for some non-bash shells such as fish-shell.
- 💁 Added some hints to the README on how to fix common setup issues.


peterberkenbosch:
nice.

README.md
```diff
@@ -211,16 +211,24 @@ You can see the build statuses at
...
+env DB=mysql bin/build
 ```
 
-The `build.sh` script runs using PostgreSQL by default, but it can be overridden by setting the DB environment variable to `DB=sqlite` or `DB=mysql`. For example:
+If the command fails with MySQL related errors you can try creating a user with this command:
 
 ```bash
-DB=mysql bash build.sh
+# Creates a user with the same name as the current user and no restrictions.
+mysql -uroot -e"CREATE USER '$USER'@'localhost'; GRANT ALL PRIVILEGES ON * . * TO '$USER'@'localhost';"
```
jacobherrington: I believe you are missing a space after `mysql -uroot -e`

elia: It's fine, for single letter options you're allowed to keep the argument attached to the option. But I'll change them to their expanded version because it's more understandable for the casual reader.

tvdeyen:
@jacobherrington @elia the commits are great. There is no need to squash them.

kennyadsl:
Thanks @elia !

PR #2972, opened 2018-11-23
Remove devise translations

afdev82:
This PR remove devise translations, since they are also removed in the solidus_i18n project (see https://github.com/solidusio/solidus_i18n/pull/107) and they are provided by the devise_i18n gem (https://github.com/tigrish/devise-i18n).

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -12,13 +12,11 @@ en:
         spree/fulfilment_changer:
           attributes:
             current_shipment:
-              can_not_have_backordered_inventory_units: has backordered inventory
-                units
+              can_not_have_backordered_inventory_units: has backordered inventory units
               has_already_been_shipped: has already been shipped
             desired_shipment:
               can_not_transfer_within_same_shipment: can not be same as current shipment
-              not_enough_stock_at_desired_location: not enough stock in desired stock
-                location
+              not_enough_stock_at_desired_location: not enough stock in desired stock location
   activerecord:
     attributes:
       spree/address:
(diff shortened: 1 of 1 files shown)


kennyadsl:
@afdev82 sorry, can you explain better why is the reformatting commit needed?

afdev82:
@kennyadsl usually the translations are alphabetically ordered and so on "normalized". In this way in both projects (solidus_i18n and solidus) they are in the same place and it's easier to translate them. I would like to add the check for the health of the translations in the ci, in this way every time someone adds a translation it should be in the right place.
At the moment there will be a lot of work to get it working, I tried to check and I have to fine tuning the configuration of the i18n-tasks gem.

kennyadsl:
> usually the translations are alphabetically ordered and so on "normalized".

Got that but shouldn't this be handled by `i18n-tasks` gem? 
What I don't get is why the PR #2963 moved things around and we are reverting some of its changes here?

afdev82:
Hm... I just run bundle exec i18n-tasks normalize inside the core directory, and then commit the result

kennyadsl:
Yes but look [at this for example](https://github.com/solidusio/solidus/pull/2963/files#diff-3329c5ee0df3de06561549745b116615R20). It does the opposite of the normalization happened in this PR, maybe some options changed on you env in the meantime?

afdev82:
I see... I have to check, I don't know why now.

kennyadsl:
@afdev82 👋do you have any update here?

afdev82:
No sorry, I didn't have time to continue on that :(

kennyadsl:
Reopened as #3132

PR #2981, opened 2018-11-30
Update CHANGELOG.md

brchristian:


kennyadsl:
I manually re-ran the failed job since it looks like a random failure:

```
Failed examples:

rspec ./spec/features/admin/orders/order_details_spec.rb:110 # Order Details as Admin cart edit page with a completed order can change the shipping method

Failure/Error: raise Capybara::ExpectationNotMet.new('Timed out waiting for Selenium session reset') if (Capybara::Helpers.monotonic_time - start_time) >= 10
```

We should open an issue if this happens frequently.

PR #2982, opened 2018-12-01
Lint ERB files via HoundCI

kennyadsl:
Now that [Hound supports ERB](https://github.com/houndci/hound/pull/1641) linting I think it's a good idea to start using that linter in Solidus. 

Linting is done with this tool: https://github.com/Shopify/erb-lint

~Since we are already linting Ruby code with Rubocop I configured hound to use the same rules that we have for the rest of the application.~ See https://github.com/solidusio/solidus/pull/2982#issuecomment-447416364

What do you think?

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,5 +1,7 @@
 ruby:
   config_file: .rubocop.yml
+erb_lint:
+  enabled: true
 scss:
   enabled: false
 javascript:


kennyadsl:
I'm not sure if we also want to lint all the current files. It's a lot of work and we could ask some help to the `erblint --autocorrect` command, even if it would need an accurate check of the result. 

But honestly I'm more worried about breaking the commits history making git blame on specific lines more complex, do you think it's worth it?

.hound.yml
```diff
@@ -1,5 +1,8 @@
 ruby:
   config_file: .rubocop.yml
+erb:
```
gylaz: We've switched to naming these by linter name (rather than language), so this should either be `erb_lint` or its alias `erblint`.

jacobherrington:
I personally think it's worth some short-term pain to increase the friendliness and quality of our ERB templates. Of course, it seems that you're the one doing most of the work, so it's up to you. 😉 

I personally don't mind adding to the Git history, but other people might feel differently.

.hound.yml
kennyadsl: @gylaz Thank! So also the other entries are not valid?

gylaz: For legacy reasons, old keys are still supported. It's just the new keys that are added (and all the new documentation) are now by linter name (given that for certain languages we now support multiple linters).

kennyadsl: Done, thanks!

aldesantis:
@kennyadsl I'm all for opportunistic refactoring, so I wouldn't fix current files all in one go as it's a lot of work. Instead, let's fix issues as they arise, what do you think?

kennyadsl:
Yes, probably that's the best option, considering also that our admin UI could need a look and feed refresh sooner or later and we'd touch those files anyway.

aitbw:
My two cents: I think the best approach when fixing linting issues is to tackle one folder per PR. For example: under `backend/app/views/spree/admin` we have `products` and `orders` —suppose all templates on these folders have issues. I would fix `products` issues on one PR and `orders` in another one. It's far easier to review changes with this approach, and it's mostly conflict-free.

kennyadsl:
Ok, considering the feedback received (if linting on existing files needs to happen we need one or more separate PRs) I think this PR is ready to accept reviews :)

kennyadsl:
I've removed the link to the `.rubocop.yml` config from the `.erblint.yml` file since it's still not supported by HoundCI. I've also removed the `.erblint.yml` file at all since that was the only configuration present. This is now ready to be merged in my opinion.

jacobherrington:
I'm still completely in support of this getting merged.

PR #2985, opened 2018-12-04
Add margin to billing fields

jacobherrington:
Really small, annoying style thing I noticed.

Before: 
screen shot 2018-12-03 at 9 13 32 pm

After:
screen shot 2018-12-03 at 9 13 50 pm

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
+++ b/backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
@@ -53,7 +53,7 @@
   margin-bottom: 15px;
 }
 
-#billing {
+label[for="order_bill_address_attributes_firstname"] {
   margin-top: 45px;
 }
 
--- a/backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
+++ b/backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
@@ -49,6 +49,10 @@
   margin-bottom: 18px;
 }
 
+[data-hook="complete-order-button"] {
+  margin-bottom: 15px;
+}
+
 #billing {
   margin-top: 45px;
 }
--- a/backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
+++ b/backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
@@ -49,6 +49,10 @@
   margin-bottom: 18px;
 }
 
+#billing {
+  margin-top: 45px;
+}
+
 #risk_analysis legend {
   background-color: #c00;
   color: #FFF;


kennyadsl:
Thanks @jacobherrington! Why the margin change to the button?

jacobherrington:
@kennyadsl I noticed it later and didn't want to make a separate PR.

The button is sitting at the very bottom of the page, it just adds a tiny amount of whitespace so that it's not against the bottom edge of the window.

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
```diff
@@ -49,6 +49,14 @@
   margin-bottom: 18px;
 }
 
+[data-hook="complete-order-button"] {
+  margin-bottom: 15px;
+}
+
+#billing {
+  margin-top: 45px;
```
tvdeyen: Every div with an id of `billing` now gets a top margin? Could we use a more specific selector instead? Or add a utility class that handles these cases? There are probably more places in admin where a form is built like this form?

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
```diff
@@ -49,6 +49,14 @@
   margin-bottom: 18px;
 }
 
+[data-hook="complete-order-button"] {
+  margin-bottom: 15px;
```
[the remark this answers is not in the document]
jacobherrington: I don't see this on the other sections. It might be present and I just haven't noticed it, but I don't think we should add bottom padding to the whole admin. My sandbox is currently broken because of #2989, but I will look for more instances of this.

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
jacobherrington: You are correct. There are also other uses of this partial where this rule needs to be applying. I'll make this selector more specific and it probably needs to live in a different SCSS file as well. Not sure what I was thinking when I added this 😬 Thanks!

jacobherrington:
@tvdeyen I agree with your comments, this was sloppy CSS, but I do think we should have a presentable UI. As long as we have best practices CSS is fairly easy to overwrite.

jacobherrington:
@tvdeyen I updated the selector for adding margin to the billing field. The UI change is the same, but this is much more specific to this single element. 

I looked for other places this CSS was needed and didn't find any, so I left the rules where they were. Please let me know if I need to do anything else for your approval.

kennyadsl:
@tvdeyen can you please review again?

PR #2988, opened 2018-12-05
Normalize API I18n keys

aitbw:
Based on the work established by #2963, this PR normalizes the I18n keys for the API.

That said, some issues are pending when running `bundle exec i18n-tasks health`, as you can see below:

image

We have two (2) options to fix this:

* Assuming the `scope` usage is the reason for these issues, remove this option and replace it with `I18n.t('spree.api.my_key')`

* Include a `i18n-tasks.yml` config file under `config` and consider all `spree.api` I18n keys used

We need to address this last issue to include an I18n health check to the CI (first mentioned in [here](https://github.com/solidusio/solidus/pull/2963#issuecomment-440978247)) in order to stay consistent within this regard.

### DIFF
--- a/api/app/controllers/spree/api/shipments_controller.rb
+++ b/api/app/controllers/spree/api/shipments_controller.rb
@@ -113,7 +113,7 @@ module Spree
         )
 
         if fulfilment_changer.run!
-          render json: { success: true, message: t('spree.shipment_transfer_success') }, status: :accepted
+          render json: { success: true, message: t('spree.api.shipment.transfer_success') }, status: :accepted
         else
           render json: { success: false, message: fulfilment_changer.errors.full_messages.to_sentence }, status: :accepted
         end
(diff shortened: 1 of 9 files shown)


kennyadsl:
@aitbw 

> Include a i18n-tasks.yml config file under config and consider all spree.api I18n keys used

What do you mean? Which configuration should this file contain in order to make the CI step work?

aitbw:
@kennyadsl, I meant maybe we could do something [like this](https://github.com/consul/consul/blob/master/config/i18n-tasks.yml#L122), but I'm more fond of the first option.

kennyadsl:
if it's possible, I'm ok removing scope.

jacobherrington:
@aitbw I'm not certain we need it, but can you rebase just to be safe?

aitbw:
Done @jacobherrington!

Issue #2989, opened 2018-12-05
Solidus breaking with Rails 5.2.2


aitbw:
### Steps to reproduce

* Run `bundle exec rake sandbox` on your Solidus source folder

### Expected behavior

* Sandbox app should be created

### Actual behavior

* Rake task to create sandbox app crashes with the following error:

```
rake aborted!
NameError: undefined method `apply_free_shipping_promotions' for class `#<Class>'
```

A "workaround" is available by commenting [this line](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L555) but doing so raises two (2) more errors:

```
rake aborted!
NameError: undefined method `Spree::Store.by_url is DEPRECATED' for class `#<Class>'
```

and

```
rake aborted!
ActiveRecord::RecordNotFound: Couldn't find Spree::Zone
```

### System configuration

* Latest master revision (7a33685)

**Extensions in use**:

* `solidus_auth_devise`

jacobherrington:
Most likely related to #2978

aitbw:
@jacobherrington Maybe but I doubt it; the issue presented in #2978 failed from time to time (I'm assuming that's because of some random value being taken into consideration when seeding the database) while this one always fails.

kennyadsl:
Could it be something related to the last Rails release happened yesterday?

kennyadsl:
I don’t think it’s related since Ransack released a version that should fix that issue a couple of hours ago.

jacobherrington:
@aitbw I see this now, I thought this was intermittent. This failure is new.

kennyadsl:
I think it's related to some [ActiveSupport deprecation change](https://github.com/solidusio/solidus/blob/dde082e96b56a96ea0b535637685f37f0a96e8bf/core/app/models/spree/order.rb#L555) in latest Rails. I'd lock Rails support to 5.2.1 for now until we know what's going on.

tvdeyen:
It seems like `ActiveSupport::Deprecation` now checks if the deprecated method actually exist, what it does not in this case. 

Adding a simple `alias_method :apply_free_shipping_promotions, :apply_shipping_promotions` should do the trick.

kennyadsl:
👍 where did you find that? I couldn't find any info about that in the ActiveSupport changelog

kennyadsl:
Also, I don't think it's the only place where we are using `deprecate` like that.

tvdeyen:
[...]

PR #2991, opened 2018-12-06
Lock Rails version up to 5.2.1

kennyadsl:
There are incompatibilities between core code and Rails 5.2.2 that we need to investigate better.

Locking Rails to 5.2.1 should allow us to have all the ecosystem working in the meantime, for example this should immediately fix:

- extensions tests against master
- sandbox
- Heroku demo

ref #2989


tvdeyen:
I do not think this is necessary. Ransack 2.1.1 is out and the missing method alias is our own fault. Will provide a PR

tvdeyen:
Emergency merge

PR #2992, opened 2018-12-06
Fix deprecations to make Solidus work with Rails 5.2.2

kennyadsl:
This PR fixes a couple of wrong calls of `deprecate` method. These errors are raising errors from Rails `5.2.2` since [this PR](https://github.com/rails/rails/pull/33325) has been merged.

This PR is also reverting the `5.2.1` lock of Rails version.

Unfortunately, if this works, I think we should also apply these fixes to Solidus `v2.6` and `v2.7` otherwise Rails `5.2.2` cannot be used by users that use those versions via RubyGems.

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -552,6 +552,7 @@ module Spree
       Spree::PromotionHandler::Shipping.new(self).activate
       recalculate
     end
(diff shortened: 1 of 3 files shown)


kennyadsl:
Not sure what Hound wants here 🐶

PR #2993, opened 2018-12-06
[v2.7] Fix deprecations to make Solidus work with Rails 5.2.2

kennyadsl:
See #2992 for more details.


kennyadsl:
@ericsaupe I'm re-running the failed job. I think it's a flaky spec but let's see.

PR #3000, opened 2018-12-13
Fix spree_promotions.code not being ignored properly

### LINKED ISSUE #2979: Calling Spree::Order#promotions calls code which raises an error
### Steps to reproduce
Calling `promotions` on an order calls `code` on the object, because the `spree_promotions` table has a `code` column. This raises an error. 

### Expected behaviour
The `columns` method should exclude `code` from being called

### Actual behaviour
Raises error: `RuntimeError (Attempted to call code on a Spree::Promotion. Promotions are now tied to multiple code records)`

### System configuration
**Solidus Version**: 2.6.0

The `columns` method should be a class method. I can create a quick PR for this.

aldesantis:
Fixes #2979.

### DIFF
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -50,7 +50,7 @@ module Spree
     self.whitelisted_ransackable_attributes = ['path', 'promotion_category_id']
 
     # temporary code. remove after the column is dropped from the db.
-    def columns
+    def self.columns
       super.reject { |column| column.name == "code" }
     end
 


core/app/models/spree/promotion.rb
```diff
@@ -50,7 +50,7 @@ class Promotion < Spree::Base
     self.whitelisted_ransackable_attributes = ['path', 'promotion_category_id']
 
     # temporary code. remove after the column is dropped from the db.
-    def columns
+    def self.columns
```
kennyadsl: This should fix the issue but what about using [ingored_columns](https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-ignored_columns) instead?

kennyadsl:
@aldesantis I've opened #3028 which removes the `code` column at all. Maybe we can move forward with that PR removing this method at all. WDYT?

PR #3002, opened 2018-12-13
Include records on API Order / Product queries

fastjames:
When querying for Orders or Products via the API, the significant number
of related records needed to render the JSON for each Order / Product
can lead to a lot of extra database traffic. By using `includes` we can
reduce the query traffic (and thus response time) significantly.

### DIFF
--- a/api/app/controllers/spree/api/orders_controller.rb
+++ b/api/app/controllers/spree/api/orders_controller.rb
@@ -51,7 +51,18 @@ module Spree
 
       def index
         authorize! :index, Order
-        @orders = paginate(Spree::Order.ransack(params[:q]).result)
+        orders_includes = [
+          :user,
+          :payments,
+          :adjustments,
+          :line_items
+        ]
+        @orders = paginate(
+          Spree::Order
+            .ransack(params[:q])
+            .result
+            .includes(orders_includes)
+        )
         respond_with(@orders)
       end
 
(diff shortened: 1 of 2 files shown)


api/app/controllers/spree/api/products_controller.rb
```diff
@@ -8,7 +8,14 @@ def index
...
         else
-          @products = product_scope.ransack(params[:q]).result
+          query = product_scope.ransack(params[:q])
+          product_includes = [
+            :variants,
+            :option_types,
+            :product_properties,
+            { classifications: :taxon }
+          ]
+          @products = query.result.includes(product_includes)
```
kennyadsl: Same here:

```ruby
# Let's evaluate moving this into a controller's private method
products_includes = [
  :variants,
  :option_types,
  :product_properties,
  { classifications: :taxon }
]

@products = product_scope
  .includes(products_includes)
  .ransack(params[:q])
  .result
```

WDYT?

fastjames: This seems reasonable, done.

api/app/controllers/spree/api/orders_controller.rb
```diff
@@ -51,7 +51,15 @@ def empty
...
-        @orders = paginate(Spree::Order.ransack(params[:q]).result)
+        query = Spree::Order.ransack(params[:q])
+        order_includes = [
+          :user,
+          :payments,
+          :adjustments,
+          :line_items
+        ]
+        orders_base = query.result.includes(order_includes)
+        @orders = paginate(orders_base)
```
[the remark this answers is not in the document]
fastjames: Agreed on maintaining some level of backcompat. I have made this change.

PR #3006, opened 2018-12-18
Allow :state to be a permitted checkout attribute

ericsaupe:
Using the API can be problematic when updating the order once it has
made it to the confirm step. This is due to calling next when
updating the order when it is in the confirm step. Passing in a new
state fixes this and flows through the state machine starting from
the passed in state.

### DIFF
--- a/api/spec/requests/spree/api/checkouts_controller_spec.rb
+++ b/api/spec/requests/spree/api/checkouts_controller_spec.rb
@@ -412,6 +412,17 @@ module Spree
           expect(response.status).to eq(422)
           expect(json_response['errors']['state'][0]).to eq('is invalid')
         end
+
+        it "cannot move to complete" do
+          expect(order.state).to eq('cart')
+          put spree.api_checkout_path(order),
+            params: { order_token: order.guest_token, order: {
+              state: 'complete'
+            } }
+
+          expect(response.status).to eq(422)
+          expect(json_response['errors']['state'][0]).to eq('cannot transition via "next"')
+        end
       end
     end
 
--- a/api/spec/requests/spree/api/checkouts_controller_spec.rb
+++ b/api/spec/requests/spree/api/checkouts_controller_spec.rb
@@ -4,17 +4,16 @@ require 'spec_helper'
 
 module Spree
   describe Api::CheckoutsController, type: :request do
+    let(:state) { create(:state) }
+    let(:country) { state.country }
+    let(:country_zone) { create(:zone, name: 'CountryZone') }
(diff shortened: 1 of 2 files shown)


api/spec/requests/spree/api/checkouts_controller_spec.rb
```diff
@@ -7,13 +7,13 @@ module Spree
     before(:each) do
       stub_authentication!
       Spree::Config[:track_inventory_levels] = false
-      country_zone = create(:zone, name: 'CountryZone')
+      @country_zone = create(:zone, name: 'CountryZone')
```
kennyadsl: I know there's no easy way to get these zones set up correctly with factories and we should give a simpler interface around this but in the meantime I'd rather start moving these `ivar` into `let!` instead of adding new ones. Do you think it's doable here?

ericsaupe: Yeah, let me give it shot.

kennyadsl:
Thinking to some security implications of merging this PR:

- will the state machine validations be skipped setting the state without calling `next`?
- can a user set the state of the order to `complete` without actually have all the steps done?

ericsaupe:
@kennyadsl good questions

1. The order still follows through with calling `next` which, surprisingly will get the test order from `cart` to `confirm` without issue (something we may need to look at elsewhere). For example, calling `next` on the test order from its initial state it passes through each step without error so you can call `next` three times and end up at confirm`
2. I've added a test to verify that you can't just pass `complete` and get the order to complete.

kennyadsl:
Re-running specs

spaghetticode:
I'm no expert here, so please forgive me if this is nonsense. 

I share @kennyadsl's concerns in regards to not using the state machine. Having a state machine and being (very easily) able to not follow it kind of invalidates it, IMHO. 

Also, in general I try to follow the principle "there should be only one way to do it" when it comes to complicate things. This is because having multiple paths to achieve one goal can easily lead to confusion and inconsistencies. 

I wonder if the scope of this PR can be limited to previous states. To me it seems kinda less dangerous to revert the state of the order only to previous ones.

ericsaupe:
@spaghetticode definitely a valid concern. Let me see if I can write some more tests to really make sure that isn't a concern and if it is shore up the code to make sure it doesn't happen.

kennyadsl:
We talked about this in the weekly Core Team meeting. It's better to try other approaches like adding a new endpoint that has the `state` parameter permitted and let you move to another state using the state machine. Something like `/recede/address`? 
It will be an extra API call but we thought it's legit in the API world.

kennyadsl:
Closing this one as discussed in the core team meetings

PR #3007, opened 2018-12-18
Verify ownership of payment_source when creating WalletPaymentSource

ericsaupe:
This merge request takes care of these things:

1. It's currently possible for payment sources created by one user to be added to the wallet of another user. This potentially allows one user to use payments stored for other users.
2. Cleans up data by restricting payment sources being added to a user's wallet if that payment source has already been added to the wallet.
3. Matches namespacing and inheritance for the Spree::WalletPaymentSource class to be the same as the other models in `spree_core` inheriting from `Spree::Base`

### DIFF
--- a/core/app/models/spree/wallet_payment_source.rb
+++ b/core/app/models/spree/wallet_payment_source.rb
@@ -13,12 +13,22 @@ module Spree
     }
 
     validate :check_for_payment_source_class
+    validate :validate_payment_source_ownership
 
     private
 
     def check_for_payment_source_class
       if !payment_source.is_a?(Spree::PaymentSource)
         errors.add(:payment_source, :has_to_be_payment_source_class)
+      end
+    end
+
+    def validate_payment_source_ownership
+      return unless payment_source.present?
+
+      if payment_source.respond_to?(:user_id) &&
+         payment_source.user_id != user_id
+        errors.add(:payment_source, :not_owned_by_user)
       end
     end
   end
--- a/core/app/models/spree/wallet_payment_source.rb
+++ b/core/app/models/spree/wallet_payment_source.rb
@@ -1,19 +1,21 @@
 # frozen_string_literal: true
 
-class Spree::WalletPaymentSource < ActiveRecord::Base
-  belongs_to :user, class_name: Spree::UserClassHandle.new, foreign_key: 'user_id', inverse_of: :wallet_payment_sources
-  belongs_to :payment_source, polymorphic: true, inverse_of: :wallet_payment_sources
+module Spree
+  class WalletPaymentSource < Spree::Base
+    belongs_to :user, class_name: Spree::UserClassHandle.new, foreign_key: 'user_id', inverse_of: :wallet_payment_sources
+    belongs_to :payment_source, polymorphic: true, inverse_of: :wallet_payment_sources
 
-  validates_presence_of :user
-  validates_presence_of :payment_source
+    validates_presence_of :user
+    validates_presence_of :payment_source
 
-  validate :check_for_payment_source_class
+    validate :check_for_payment_source_class
 
-  private
+    private
 
-  def check_for_payment_source_class
-    if !payment_source.is_a?(Spree::PaymentSource)
-      errors.add(:payment_source, :has_to_be_payment_source_class)
+    def check_for_payment_source_class
+      if !payment_source.is_a?(Spree::PaymentSource)
+        errors.add(:payment_source, :has_to_be_payment_source_class)
+      end
     end
   end
 end
--- a/core/app/models/spree/wallet_payment_source.rb
+++ b/core/app/models/spree/wallet_payment_source.rb
@@ -7,6 +7,10 @@ module Spree
 
     validates_presence_of :user
     validates_presence_of :payment_source
+    validates :user_id, uniqueness: {
+      scope: [:payment_source_type, :payment_source_id],
+      message: :payment_source_already_exists
+    }
 
     validate :check_for_payment_source_class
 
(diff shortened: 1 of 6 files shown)


core/app/models/spree/wallet_payment_source.rb
```diff
@@ -1,19 +1,31 @@
...
-  belongs_to :user, class_name: Spree::UserClassHandle.new, foreign_key: 'user_id', inverse_of: :wallet_payment_sources
-  belongs_to :payment_source, polymorphic: true, inverse_of: :wallet_payment_sources
+module Spree
+  class WalletPaymentSource < Spree::Base
+    belongs_to :user, class_name: Spree::UserClassHandle.new, foreign_key: 'user_id', inverse_of: :wallet_payment_sources
+    belongs_to :payment_source, polymorphic: true, inverse_of: :wallet_payment_sources
 
-  validates_presence_of :user
-  validates_presence_of :payment_source
+    validates_presence_of :user
```
skukx: I would prefer `validates :user, presence: true`. Not a big deal though

core/app/models/spree/wallet_payment_source.rb
```diff
@@ -1,19 +1,31 @@
...
+    validates_presence_of :payment_source
+    validates :user_id, uniqueness: { scope: [:payment_source_type, :payment_source_id] }
```
skukx: It may be worth adding a custom message here. I found the default message on this is going to be `User id has already been taken`. That message doesn't really denote that the violation is due to the tuple key

core/app/models/spree/wallet_payment_source.rb
```diff
@@ -1,19 +1,31 @@
...
+    validate :check_for_payment_source_class
```
skukx: I'm guessing the reasoning for this is to mimic what happens in the `Spree::Wallet#add` method? I'm a fan prepending validation methods with `validate` so you know the purpose of that method is to validate and if called will add to the object errors. `validate_payment_source_class`

core/app/models/spree/wallet_payment_source.rb
```diff
@@ -1,19 +1,31 @@
 # frozen_string_literal: true
 
-class Spree::WalletPaymentSource < ActiveRecord::Base
-  belongs_to :user, class_name: Spree::UserClassHandle.new, foreign_key: 'user_id', inverse_of: :wallet_payment_sources
-  belongs_to :payment_source, polymorphic: true, inverse_of: :wallet_payment_sources
+module Spree
```
kennyadsl: can you please change the namespace/inheritance stuff into one or more other commits, so they are independently  documented? Also, why are these change needed?

core/app/models/spree/wallet_payment_source.rb
kennyadsl: Also, can we please document why this change is needed in this PR?

skukx:
To clarify some of these changes (each change should probably be documented in it's own commit):
```ruby
validates :user_id, uniqueness: { scope: [:payment_source_type, :payment_source_id] }
```

This change is due to a unique tuple key on the database. Adding the validation mirrors what is enforced at the database layer and allows activerecord to handle it.  What was a `PG::UniqueViolation` error becomes an `ActiveRecord::RecordInvalid` error.

----------------------

For
```ruby
validate :check_for_payment_source_ownership
```

It is possible that a user's wallet can have a payment source added to it which does not belong to the user.  This means a user could make transactions on a payment source that does not belong to them.

core/app/models/spree/wallet_payment_source.rb
ericsaupe: It's to follow the format used in the other model files.

core/app/models/spree/wallet_payment_source.rb
kennyadsl: @skukx agree but these validations were already there, they have just been indented due to the module change. I think that if we want to change them we need to open another PR and deprecate existing methods.

core/app/models/spree/wallet_payment_source.rb
```diff
@@ -1,19 +1,35 @@
...
+    belongs_to :user, class_name: Spree::UserClassHandle.new, foreign_key: 'user_id', inverse_of: :wallet_payment_sources
+    belongs_to :payment_source, polymorphic: true, inverse_of: :wallet_payment_sources
 
-  validates_presence_of :user
-  validates_presence_of :payment_source
+    validates_presence_of :user
+    validates_presence_of :payment_source
+    validates :user_id, uniqueness: {
+      scope: [:payment_source_type, :payment_source_id],
+      message: :payment_source_already_exists
```
kennyadsl: This could potentially create issues with payment sources that have already been added to a user and are no more valid - for example, if you already have duplicates and try to update one of them, right?

I'm not sure what is the best way to handle this, maybe use this validation only [on create](https://guides.rubyonrails.org/active_record_validations.html#on) so it does not apply to already added payment sources?

Also, how did you have this issue? Can you share the reason/scenario behind this problem (which I see exists)?

kennyadsl: just read https://github.com/solidusio/solidus/pull/3007#issuecomment-448727125 - nevermind

core/config/locales/en.yml
```diff
@@ -488,6 +488,9 @@ en:
           attributes:
             payment_source:
               has_to_be_payment_source_class: has to be a Spree::PaymentSource
+              not_owned_by_user: does not belong to user
+            user_id:
+              payment_source_already_exists: already has the Spree::PaymentSource in their wallet
```
tvdeyen: This is not a very human friendly error message. Could we use

"already has this payment source in their wallet"?

aitbw: I'm assuming @ericsaupe took inspiration from the message above. I think we could fix both in this PR.

PR #3015, opened 2018-12-31
Enable partial doubles verification for RSpec (part 2)

cedum:
This is a follow up of #3005 and enables the partial double verification for RSpec also in Solidus API, Frontend and Backend.


frontend/spec/support/example_groups/partial_double_verification_example_group.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module PartialDoubleVerificationExampleGroup
```
kennyadsl: Can't we share the support file introduced in core via #3005 ? Maybe moving the module in `core/lib` into the `testing_support` folder?

cedum:
Pushed some changes that DRYes the partial double verification support file as per @kennyadsl suggestion.
I refactored also the module into a plain `RSpec.configure` block, mostly because I think there're no benefits to isolate it into a module (+ it was requiring ActiveSupport).
It made sense to me to extract also the `c.verify_partial_doubles = true` setting; in cases where a different setting is required in a single component, it can be easily overridden.
Wdyt?

PR #3017, opened 2019-01-02
Revert PR #2672 [v2.6]

aitbw:
This patch aims to revert changes introduced in PR #2672 to offer fully-functioning stores for those using the [legacy stock system](https://github.com/solidusio-contrib/solidus_legacy_stock_system) extension

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -73,6 +73,11 @@ module Spree
     has_many :variants, through: :line_items
     has_many :products, through: :variants
 
+    if Spree.solidus_gem_version > Gem::Version.new('2.5.0')
+      has_many :order_stock_locations, class_name: 'Spree::OrderStockLocation'
+      has_many :stock_locations, through: :order_stock_locations
+    end
+
     # Shipping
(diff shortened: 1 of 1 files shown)


core/app/models/spree/order.rb
```diff
@@ -73,6 +73,11 @@ class CannotRebuildShipments < StandardError; end
     has_many :variants, through: :line_items
     has_many :products, through: :variants
 
+    if Spree.solidus_gem_version > Gem::Version.new('2.5.0')
+      has_many :order_stock_locations, class_name: 'Spree::OrderStockLocation'
+      has_many :stock_locations, through: :order_stock_locations
+    end
```
kennyadsl: What about adding this code to the [legacy stock system](https://github.com/solidusio-contrib/solidus_legacy_stock_system) extension instead?

jacobherrington: ☝️ This seems like a lower touch fix.

aitbw: @kennyadsl @jacobherrington I've been trying to implement this extension-level, but so far it is not working :-( I'll let you know if I come up with a more clean solution

PR #3019, opened 2019-01-03
Setup Netlify site for yard docs

tvdeyen:
Preparations for deploying the yard docs on Netlify.


kennyadsl:
@tvdeyen is the Netlify failure normal here?

tvdeyen:
@kennyadsl No. I need to make some more adjustments. Stay tuned

tvdeyen:
@kennyadsl ready ✅

tvdeyen:
Ideally we would only install gems that are necessary to build the docs (namely `yard`), but this would make it necessary to alter the install command on Netlify. Anyhow the gems are cached and this is working for now. We can make this change later if we want to safe some resources 🌳

tvdeyen:
@kennyadsl thanks. 

I was able to disable all gem groups we don't need with setting the ENV var

```
BUNDLE_WITHOUT=frontend backend utils ci
```

in the Netlify admin.

This should safe some resources 🐳

tvdeyen:
With that gem grouping change we should be able to only install the gems needed to build the current sub project. Currently we are [always installing all gems for each subproject](https://github.com/solidusio/solidus/blob/master/.circleci/config.yml#L28). I will prepare another PR to make the changes.

PR #3023, opened 2019-01-04
Rescue from `Spree::Order::InsufficientStock` on `frontend` checkout flow

spaghetticode:
Ref #3023 

When two users try to purchase the last item remaining from a non-backordeable
stock location at the same time then the last one will experience an unhandled
error `Spree::Order::InsufficientStock`.

This happens only if there is a second backorderable stock location for the
product.

When there is no backorderable stock location the controller `before_action`
`ensure_sufficient_stock_lines` is enough to catch the issue in advance.

The error is generated by this line in Spree::Order model:

`before_transition to: :complete, do: :validate_line_item_availability`

Generally, `ensure_sufficient_stock_lines` prevent customers to complete the
checkout process when there is not enough stock availability, but the case
above is not caught here.

So, by using `rescue_from` the customer is now redirected to the checkout
`address` page and shown an error message that suggests to repeat the
checkout process.

The order's shipments will be rebuilt using the backorderable stock location
in the delivery step, allowing them to (hopefully!) complete the purchase.

`rescue_from` is already used on the `api` and `backend` section in order to
manage `InsufficientStock` errors, so this was a natural choice also on the
`frontend`.

### DIFF
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -24,6 +24,7 @@ module Spree
     helper 'spree/orders'
 
     rescue_from Spree::Core::GatewayError, with: :rescue_from_spree_gateway_error
+    rescue_from Spree::Order::InsufficientStock, with: :insufficient_stock_error
 
     # Updates the order and advances to the next state (when possible.)
     def update
@@ -232,6 +233,23 @@ module Spree
 
     def check_authorization
       authorize!(:edit, current_order, cookies.signed[:guest_token])
+    end
+
+    def insufficient_stock_error
+      packages = @order.shipments.map(&:to_package)
+      if packages.empty?
+        flash[:error] = I18n.t('spree.insufficient_stock_for_order')
+        redirect_to cart_path
+      else
+        availability_validator = Spree::Stock::AvailabilityValidator.new
+        unavailable_items = @order.line_items.reject { |line_item| availability_validator.validate(line_item) }
+        if unavailable_items.any?
+          item_names = unavailable_items.map(&:name).to_sentence
+          flash[:error] = t('spree.inventory_error_flash_for_insufficient_shipment_quantity', unavailable_items: item_names)
+          @order.restart_checkout_flow
+          redirect_to spree.checkout_state_path(@order.state)
+        end
+      end
     end
   end
 end
--- a/frontend/spec/features/checkout_confirm_insufficient_stock_spec.rb
+++ b/frontend/spec/features/checkout_confirm_insufficient_stock_spec.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe "Checkout confirm page submission", type: :feature do
+  include_context 'checkout setup'
+
+  context "when the product from the order is not backorderable but has enough stock quantity" do
+    let(:user) { create(:user) }
+
+    let(:order) { Spree::TestingSupport::OrderWalkthrough.up_to(:payment) }
+    let(:order_product) { order.products.first }
+    let(:order_stock_item) { order_product.stock_items.first }
+
+    before do
+      order_stock_item.update! backorderable: false
+      order_stock_item.set_count_on_hand(1)
+      allow_any_instance_of(Spree::CheckoutController).to receive_messages(current_order: order)
+      allow_any_instance_of(Spree::CheckoutController).to receive_messages(try_spree_current_user: user)
+      allow_any_instance_of(Spree::OrdersController).to receive_messages(try_spree_current_user: user)
+    end
+
+    context 'when there are not other backorderable stock locations' do
+      context 'when the customer is on the confirm page and the availabilty drops to zero' do
+        before do
+          visit spree.checkout_state_path(:confirm)
+          order_stock_item.set_count_on_hand(0)
+        end
+
+        it 'redirects to cart page and shows an unavailable product message' do
+          click_button "Place Order"
+          expect(page).to have_content "#{order_product.name} became unavailable"
+          expect(page).to have_current_path spree.cart_path
+        end
+      end
+    end
+
+    context 'when there is another backorderable stock location' do
+      before do
+        create :stock_location, backorderable_default: true, default: false
+      end
+
+      context 'when the customer is on the confirm page and the availabilty drops to zero' do
+        before do
+          visit spree.checkout_state_path(:confirm)
+          order_stock_item.set_count_on_hand(0)
+        end
+
+        it "redirects to the address checkout page and shows an availability changed message" do
+          click_button "Place Order"
+          error_message = "Quantity selected of #{order_product.name} is not available. Still, items may be available from another stock location, please try again."
+          expect(page).to have_content error_message
+          expect(page).to have_current_path spree.checkout_state_path(:address)
+        end
+
+        it "can still complete the order using the backorderable stock location by restarting the checkout" do
+          click_button "Place Order"
+          expect(page).to have_current_path spree.checkout_state_path(:address)
+          click_button "Save and Continue"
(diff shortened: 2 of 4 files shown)


frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -161,6 +162,19 @@ def ensure_sufficient_stock_lines
...
 
+    def ensure_sufficient_shipment_stock_quantities
+      if @order.passed_checkout_step?('delivery')
+        availability_validator = Spree::Stock::AvailabilityValidator.new
+        unless @order.line_items.all? { |item| availability_validator.validate(item) }
+          unavailable_items = @order.line_items.map { |item| item.errors.full_messages.to_sentence }.to_sentence
+          flash[:error] = t('spree.inventory_error_flash_for_insufficient_shipment_quantity', unavailable_items: unavailable_items)
+          @order.restart_checkout_flow
+          @order.next!
+          redirect_to spree.checkout_state_path(:delivery)
```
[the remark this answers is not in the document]
spaghetticode: Yes, the goal is to revert the order status to `delivery`, and there is no apparent need to restart from the `cart` state, but my guess is that using `#restart_checkout_flow` and add one `#next!` call may be better, at least in the long run, as `#ensure_updated_shipments` might develop in a more complex method that adds more (required) behavior to the restarting process. I'd rather follow the path than create a new one with code such as `@order.update state: :delivery` or someting similar.

On the other side what makes me frown a bit is the `@order.next!` line, as it may have one meaning today (advance the order state to `delivery`) but that can change in the future. 

Maybe the best thing would be to `restart_checkout_flow` and remove the `next!`.

I'm not sure about what's the best solution, so I'm committed to following the advice of people with a better understanding of the checkout process and its intricacies.

frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -161,6 +162,19 @@ def ensure_sufficient_stock_lines
       end
     end
 
+    def ensure_sufficient_shipment_stock_quantities
+      if @order.passed_checkout_step?('delivery')
```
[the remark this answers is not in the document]
spaghetticode: @kennyadsl It's not necessarily assuming that the following step is the last... but it must be executed after the order passed the `delivery` step, for some reasons.  

I'll share some context regarding why I added this `if` clause, I think it may help us in finding a better solution for the issue you exposed. 

The reason is that, without that `if`, this [spec](https://github.com/solidusio/solidus/blob/master/frontend/spec/features/checkout_unshippable_spec.rb) fails. The spec seems to be still relevant to me, so adding the `if` allowed me to add the functionality while still keeping the existing behavior.  

Today I tried to dig deeper in the issue regarding the spec mentioned above, and this is what I found out. After moving my `before_action` to the end of the chain things improved a little, but not enough. 

The reason for having that specific `if` is that the `before_action` `CheckoutController::before_payment` may manipulate the order, specifically these lines:

```ruby
        @differentiator = Spree::Stock::Differentiator.new(@order, packages)
        @differentiator.missing.each do |variant, quantity|
          @order.contents.remove(variant, quantity)
        end
```

Besides, the line `@order.contents.remove(variant, quantity)` has the side effect of changing the order state from `payment` to `address` 🤔. Have not checked why, but I guess some `ActiveRecord` callback is responsible for that. 

Missing contents must be removed before running the new `before_action`, or behavior will change, and as said before, the removal happens only when the customer tries to see the `checkout/payment` page, which happens after `delivery` is completed. I think the state ordering `delivery -> payment` is quite robust, as generally order total should be calculated only after shipping charges and taxes are known, both depending on addresses. But YMMV, so robust may not be enough.

Another option would be to move the new code in the `before_payment` method, so it's executed when it starts to be relevant, but that would limit its execution only to that precise step, eventually limiting its usefulness, that's why I think this is not a viable solution. 

The current solution has the strong disadvantage that `@order.passed_checkout_step?('delivery')` does not make clear why the step must be passed, and the strong connection with the existing `before_payment` method. 

A different approach may be to leverage the existing code in `before_payment` and change the `if` to this, for example:

```ruby
      packages = @order.shipments.map(&:to_package)
      @differentiator = Spree::Stock::Differentiator.new(@order, packages)
      if @differentiator.missing.empty?
      # ...
```

This way, there's no more explicit dependence on order steps and their names. What do you think?

frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -233,5 +234,21 @@ def rescue_from_spree_gateway_error(exception)
     def check_authorization
       authorize!(:edit, current_order, cookies.signed[:guest_token])
     end
+
+    def insufficient_stock_error
+      packages = @order.shipments.map(&:to_package)
+      if packages.empty?
+        flash[:error] = I18n.t('spree.insufficient_stock_for_order')
+        redirect_to cart_admin_order_url(@order)
```
kennyadsl: This path should probably not contain `_admin_`. Also maybe we should use `_path` instead of `_url`, right?

frontend/spec/features/checkout_confirm_insufficient_stock_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      context 'when the customer is on the confirm page and the availabilty drops to zero' do
+        before do
+          visit spree.checkout_state_path(:confirm)
+          order_stock_item.set_count_on_hand(0)
+        end
+
+        it 'redirects to cart page and shows an unavailable product message' do
+          click_button "Place Order"
+          expect(page).to have_content "#{order_product.name} became unavailable"
+          expect(page).to have_current_path spree.cart_path
```
kennyadsl: not sure how this could pass if the path in the controller was wrong, can you please double check?

frontend/app/controllers/spree/checkout_controller.rb
spaghetticode: good catch! 👍

frontend/spec/features/checkout_confirm_insufficient_stock_spec.rb
spaghetticode: Specs pass because they don't exercise the path that was wrong. I'll fix the path and add another test for that branch of the `if`.

spaghetticode:
@kennyadsl I updated the code according to your comments, adding a couple of commits to facilitate your understanding of the changes. 

When it's ok for you I'd like to rebase/squash and update the PR description/commit message in order to reflect the latest changes introduced. Just let me know when it's fine for you (or please add more comments if something is still OK 😸 )

kennyadsl:
@spaghetticode go ahead with the commits squash!

frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -233,5 +234,21 @@ def rescue_from_spree_gateway_error(exception)
...
+
+    def insufficient_stock_error
+      packages = @order.shipments.map(&:to_package)
+      if packages.empty?
+        flash[:error] = I18n.t('spree.insufficient_stock_for_order')
+        redirect_to cart_path
+      else
+        availability_validator = Spree::Stock::AvailabilityValidator.new
+        unless @order.line_items.all? { |item| availability_validator.validate(item) }
+          unavailable_items = @order.line_items.map(&:name).to_sentence
```
rubenochiavone: `@order.line_items.all? { |item| availability_validator.validate(item) }` means that at least one line item is not valid, right? So, I'm not sure that `unavailable_items = @order.line_items.map(&:name).to_sentence` is right because it lists all line items as unavailable. Perhaps

```ruby
unavailable_items = @order.line_items.select { |line_item| !availability_validator.validate(line_item) }.map(&:name).to_sentence
```

would do (haven't tested). WDYT?

spaghetticode: You're right, something went wrong with my cut & paste refactoring. It should be fine now, thank you for spotting the bug 🐛

PR #3028, opened 2019-01-06
Remove code column from spree_promotions table

kennyadsl:
# Description

This PR removes `code` column form `Spree::Promotion` since promotions are now tied to multiple codes via `Spree::PromotionCode`. 

Ref. #2979 

# TODO:

- [x] [Remove columns method on Spree::Promotion](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/promotion.rb#L52-L55). This should be done in sync with #3000

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,36 @@ PR description for more info.
 
 - Replace jquery_ujs with rails-ujs in frontend and backend [#3027](https://github.com/solidusio/solidus/pull/3027) ([kennyadsl](https://github.com/kennyadsl))
 
+**Removed code from Spree::Promotion**
+
+Previously Solidus used `code` column on `spree_promotions` to add a code
+to promotions that could be used as coupon code by users. This is no more a
+thing since we support multiple coupon codes associated to a single promotion.
+
+This change is important because it's quite common for old stores to have some
+promotion with `code` field still present in the database, even if it's not used.
+When performing the migration present in this PR it will raise an exception if
+there are records in the `spree_promotions` table with that field present.
+It's up to each store to understand how to handle this scenario before running
+this migration. We also provide other two ways to handle this, and users can
+just change the migration after it has been copied into their store.
+It's just matter of changing the content of the
+`RemoveCodeFromSpreePromotions.promotions_with_code_handler` method and make it
+return one of the following:
+
+- `Solidus::Migrations::PromotionWithCodeHandlers::MoveToSpreePromotionCode`:
+  it will convert Spree::Promotion#code to a `Spree::PromotionCode` before
+  removing the `code` column.
+- `Solidus::Migrations::PromotionWithCodeHandlers::DoNothing`: it will print
+  a message to track what we are deleting.
+
+Alternatively users can create their own class to handle data and return that
+class. The new class could inherit from `PromotionsWithCodeHandler` and
+should respond to `call`.
+
+- Remove `code` column from `spree_promotions` table.
+[#3028](https://github.com/solidusio/solidus/pull/3028) ([kennyadsl](https://github.com/kennyadsl))
+
 ### Core
 
 - Fix Spree::Variant inconsistency due to lack of product association [#3043](https://github.com/solidusio/solidus/pull/3043) ([rubenochiavone](https://github.com/rubenochiavone))
--- a/core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb
+++ b/core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class RemoveCodeFromSpreePromotions < ActiveRecord::Migration[5.1]
+  def up
+    remove_index :spree_promotions, name: :index_spree_promotions_on_code
+    remove_column :spree_promotions, :code
+  end
+
+  def down
+    add_column :spree_promotions, :code, :string
+    add_index :spree_promotions, :code, name: :index_spree_promotions_on_code
+  end
+end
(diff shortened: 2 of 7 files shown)


kennyadsl:
I have to add a guard to let people know that they have to remove data from the column before running the migration. Also, I need to add a changelog entry

CHANGELOG.md
```diff
@@ -1,5 +1,36 @@
...
+## Major Changes
+
+### Removed code from Spree::Promotion
+
+Previously Solidus used `code` column on `spree_promotions` to add a code
+to promotions that could be used as coupon code by users. This is no more a
+thing since we support multiple coupon codes associated to a single promotion.
+
+This change is important because it's quite common for old stores to have some
+promtion with `code` field still present in the database, even if it's not used.
```
mdesantis: promtion -> promotion

CHANGELOG.md
```diff
@@ -1,5 +1,36 @@
...
+When performing the migration present in this PR it will raise an exception if
+there are records in that database with that field filled.
+It's up to each store to understand how to handle this scenario before running
+this migration. We also provide other 2 ways to handle this, and each user can
```
mdesantis: 2 -> two (you are indicating a quantity)

core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb
```diff
@@ -0,0 +1,89 @@
...
+  class MoveToSpreePromotionCode < PromotionsWithCodeHandler
+    def call
+      # This is another possible approach, it will convert Spree::Promotion#code
+      # to a Spree::PromotionCode before removing the `code` field.
+      #
+      promotions.each do |promotion|
+        migration_context.say "Creating Spree::PromotionCode with value '#{promotion.code}' "\
+          "for Spree::Promotion with id '#{promotion.id}'"
+
+        promotion.codes.create(value: promotion.code)
```
mdesantis: What about `create!`?

core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb
```diff
@@ -0,0 +1,89 @@
...
+        "\n"
+    end
+  end
+
+  class MoveToSpreePromotionCode < PromotionsWithCodeHandler
+    def call
+      # This is another possible approach, it will convert Spree::Promotion#code
+      # to a Spree::PromotionCode before removing the `code` field.
+      #
+      promotions.each do |promotion|
```
spaghetticode: I suppose the number of promotions might be quite high in some databases. In that case `find_each` should  work better than `each`

core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb
kennyadsl: Thanks for pointing this out @mdesantis !

There's a uniqueness validation on `value` so if that value is already used (let's say the code has been already moved manually into the `Spree::PromotionCode`) it will raise an exception.

What about using `Spree::PromotionCode.find_or_create_by!(value: value, promotion_id: promotion.id)` instead? 
I think this way it will only raise if the value is not unique and that value has been used on a `Spree::PromotionCode` associated with another `Spree::Promotion`. It will do nothing if the current `Spree::Promotion` already has a `Spree::PromotionCode` with the same value.

mdesantis: I totally agree with that :+1:

core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def up
+    promotions_with_code = Spree::Promotion.where.not(code: nil)
+
+    if promotions_with_code.any?
+      # You have some promotions with "code" field present! This is not good
+      # since we are going to remove that column.
+      #
+      self.class.promotions_with_code_handler.new(self, promotions_with_code).call
```
elia: I think it would be better to use a simple callable instead of a class that responds to `#call` and pass options directly to `call`, also a constant would be more straightforward than a method:

```suggestion

      deprecated_code_handler = StopWithError
      # deprecated_code_handler = DoNothing
      # deprecated_code_handler = MoveToSpreePromotionCode
      deprecated_code_handler.call(self, promotions_with_code)
```

and then below:

```rb
# Please note that this will block the current migration and rollback all
# the previous ones run with the same "rails db:migrate" command.
StopWithError = ->(migration, promotions) {
  raise StandardError, "You are trying to drop 'code' column from "\
    "spree_promotions table but you have at least one record with that "\
    "column filled. Please take care of that or you could lose data. See:" \
    "\n" \
    "https://github.com/solidusio/solidus/pull/3028"\
    "\n"
}
```

This is much simpler and has the advantage of having the configuration near the error location, making it easier to find

kennyadsl: This is very similar to what I did first but to test all the possible proposed approaches I ended up creating a method. I know it's not ideal but I can't think to other solutions for this. Feel free to suggest how can I keep this code tested reducing its complexity.

tvdeyen: I like Elias idea of using simple callables, but Albertos approach is working well and is tested (not that easy with lambdas), so I think it's fine. The outcome is the same. 

What I do care about is the very thoughtful work that is done here and want to give an extra appreciation! Very nicely done, Alberto! 👏

core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb
```diff
@@ -0,0 +1,99 @@
...
+      end
+    end
+  end
+
+  class DoNothing < PromotionsWithCodeHandler
+    def call
+      # This approach will delete all codes without taking any action. At
+      # least we could print a message to track what we are deleting.
+      #
+      promotions.each do |promotion|
```
spaghetticode: There's also this instance of `each` that can be replaced with `find_each`. Sorry for not finding it earlier.

kennyadsl:
There's some issue with the database in test environment when it tries to roll back to the previous state after the test ends, I need to investigate more on why it's happening.

PR #2996, opened 2018-12-07
Deprecate 'X-SPREE-TOKEN' header

Nittarab:
I changed the use of custom header `X-SPREE-TOKEN` with `Authorization: Bearer my_secret_token`
Suggested in #2934

Issue #2934, opened 2018-11-06
Deprecate 'X-SPREE-TOKEN' header


skukx:
Rather than using a custom header we should move this to be under the `Authorization` header.
Example:
```
Authorization: Bearer my_secret_token
```

We then could update the `load_user` method to look like
```ruby
def load_user
  @current_api_user ||= authenticate_with_http_token do |token, options|
    Spree.user_class.find_by(spree_api_key: token)
  end
end
```
See: https://github.com/solidusio/solidus/blob/8fd6b1f9175b936b5db3d1724c87bd1fe426425a/api/app/controllers/spree/api/base_controller.rb#L45-L47

kennyadsl:
Closed with #3029, thanks @Nittarab and @twist900!

PR #3029, opened 2019-01-09
Deprecate 'X-SPREE-TOKEN' header 2

twist900:
# Description

This PR is a continuation of #2996. It keeps the X-SPREE-TOKEN alongside the newly introduced bearer authorization token for backward compatibility. A deprecation message is displayed when the X-SPREE-TOKEN header is detected.

Also, it fixes some linting issue.

ref #2934 

# Checklist:

### DIFF
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -100,9 +100,26 @@ module Spree
       end
 
       def api_key
-        request.headers["X-Spree-Token"] || params[:token]
+        bearer_token || spree_token || params[:token]
       end
       helper_method :api_key
+
+      def bearer_token
+        pattern = /^Bearer /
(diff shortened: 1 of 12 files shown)


api/app/controllers/spree/api/base_controller.rb
```diff
@@ -100,10 +100,27 @@ def invalid_resource!(resource)
...
 
       def api_key
-        request.headers["X-Spree-Token"] || params[:token]
+        bearer_token || spree_token || params[:token]
       end
       helper_method :api_key
 
+      def bearer_token
+        pattern = /^Bearer /
+        header  = request.headers["Authorization"]
```
kennyadsl: I'd remove one space after `header`, it's aligned to `pattern` on the line above but if that line changes in the future we'll be forced to change this one as well and this is not ideal for git history. What do you think?

PR #3031, opened 2019-01-09
fixed nested attributes delete issue

forever-sumit:
# Description
I found an issue , when we try to add the a nested record like in product property page if we add a  new property, and fill any data in the newly added fields. after that I'll click on delete icon, without making field empty. 

In this case I see that newly added field not shown in page, but if I click on save or edit button, I see that the records are created in the database.

Here the issue is , when we click on the delete button, Javascript code try to find that dom element and try to change val, but the dom element not found. so the destroy value not set and it submit as a nil value.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -101,7 +101,7 @@ Spree.ready(function(){
 
   $('body').on('click', 'a.spree_remove_fields', function() {
     var el = $(this);
-    el.prev("input[type=hidden]").val("1");
+    el.next("input[type=hidden]").val("1");
     el.closest(".fields").hide();
     if (el.prop("href").substr(-1) == '#') {
       el.parents("tr").fadeOut('hide');


kennyadsl:
Thanks for the contribution @forever-sumit, would you mind also adding a spec that fails before this fix?

forever-sumit:
@kennyadsl , before this fix no any specs are failing, I will try to add spec that cover my fix.

jacobherrington:
Hey @forever-sumit, any update on this? Would love to get this merged!

forever-sumit:
Hey @jacobherrington, Due to other work load, I'm not able to write the spec yet.

if you guys feel my changes is okay to merge then please merge it. I'll write the spec later when I get a chance.

kennyadsl:
I'd say to wait for specs anyway. I'm not confident merging this without specs since I'm not 100% confident on the context where this issue happens. If someone else (maybe from the core team) has time to go deeper and understand better, that person could also probably add specs to finish this PR.

kennyadsl:
@hefan Hey there, do you think this will be fixed by your recent work?

hefan:
This should not happen with #3592
At least if 
/admin/products/:product/product_properties
or
/admin/option_types/:option_type/edit
is meant.

The non persistent field is not created when its removed before updating.
`$('body').on('click', 'a.spree_remove_fields', function() {`
does no longer mess then with ids, it just removes the line.

kennyadsl:
closed with https://github.com/solidusio/solidus/pull/3592

PR #3034, opened 2019-01-11
PoC: Decorators / Presenters [WIP]

mdesantis:
# PoC: Decorators / Presenters [WIP]

## Introduction

Hello, I'm trying to find a better way to handle view logic on Solidus.

### Brief recap about view logic management patterns

Helpers have always been the  Achilles' heel of the full object oriented of Rails. Even [Steve Klabnik complained about them](http://blog.steveklabnik.com/posts/2011-09-09-better-ruby-presenters). But, what are the alternatives?

#### Presenter pattern

The Presenter pattern couples view logic with the controller/action. So, for example, if you have a PagesController with `home` and `about` actions, you could have:

```ruby
class PagesPresenter < ApplicationPresenter
  def nav_items
    [ { name: 'Home', path: home_path } ,
      { name: 'About', path: about_path } ]
  end
end

class Pages::HomePresenter < PagesPresenter
  def greeting
    "Hi, #{current_user.nickname}!" if current_user
  end
end
```

#### Presentation Model (aka Decorator) pattern

The Presentation Model pattern couples view logic with the model instance. So, for example, if you have a User model, you could have:

```ruby
class UserDecorator
  def full_name
    "#{first_name} #{last_name}"
  end
end

UserDecorator.decorate(User.new)
```

Essentially, a presenter is a class which takes as input the object to be presented (let's call it _subject_) and returns an instance of it after having applied some view logic related augmentations (that are often called _decorations_, as presenters are often known as _decorators_, but let's avoid that name, since on Solidus _decorators_ already indicate overrides of the platform business logic).

The main advantage of it is that you have a place where to put view logic related to models, while, without presenters, you are forced whether or to dirty the model class, or to use helpers, thus precluding OOP (see e.g. https://github.com/solidusio/solidus/blob/aabb0967c0796b7c301c963f88dbf8947236773c/core/app/helpers/spree/base_helper.rb#L113-L126).

## My effort

First, I looked into the current state of things regarding existing Ruby presenters implementations. This is the summary:

#### API samples

```ruby
Article.find(params[:id]).decorate
```

#### pros

- full featured

- has been around for a while

- pretty popular (~ 4.5k GH stars)

#### cons

- a bit bloated

- maintenance is "meh"

- `Decorator` suffix can't be customized (and we already have decorators concept which would lead to confusion)

#### API samples

```ruby
# inside views
@author.full_name
# Explicit call
```

#### pros

- from Kaminari's author

- simple and clean

- `Decorator` suffix can be customized

#### cons

- not very popular (~ 200 GH stars)

- has auto-decoration. To me it means that we could have many issues like "why is country.name in the controller different from Spree::Country.find(1).name in the console?!"

#### pros

- Uses `Presenter` instead of `Decorator`

#### cons

- still in early stage

- few GH stars (5)

## https://gitlab.com/gitlab-org/gitlab-ce/blob/master/app/presenters/README.md

```ruby
class Spree::CountriesPresenter
  presents :countries
end
# Spree::Presenter.present(countries, attributes: { current_user: user }, as: 'Spree::CountryList')
Spree::CountryListPresenter.present(countries, attributes: { current_user: user })
```

## Performance

Unfortunately, abstraction doesn't come for free: it comes with runtime overhead.

### DIFF
--- a/core/app/helpers/spree/base_helper.rb
+++ b/core/app/helpers/spree/base_helper.rb
@@ -119,6 +119,10 @@ module Spree
         countries = Country.all
       end
 
+      Spree::CountryListPresenter.new(countries)
+    end
+
+    def countries_for_select(countries)
       country_names = Carmen::Country.all.map do |country|
         [country.code, country.name]
       end.to_h
--- a/core/app/presenters/spree/country_active_decorator_presenter.rb
+++ b/core/app/presenters/spree/country_active_decorator_presenter.rb
@@ -1,0 +1,19 @@
+module Spree
+  module CountryActiveDecoratorPresenter
+    def name
+      "CountryActiveDecoratorPresenter: #{translated_name || carmen_name || super}"
+    end
+
+    def carmen_country
+      Carmen::Country.coded iso
+    end
+
+    def carmen_name
+      carmen_country.try :name
+    end
+
+    def translated_name
+      I18n.t("spree.country_names.#{iso}", default: nil)
+    end
+  end
+end
--- a/core/app/presenters/spree/country_list_presenter.rb
+++ b/core/app/presenters/spree/country_list_presenter.rb
@@ -1,0 +1,15 @@
+module Spree
+  class CountryListPresenter < Spree::Presenter::Delegated
+    def initialize(subject)
+      subject = prepare_countries(subject)
+
+      super
+    end
+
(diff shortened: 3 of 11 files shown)


mdesantis:
It's too early to open this PR here, I'm going to close it and open again later

PR #3035, opened 2019-01-12
Fix Spree::Order::InsufficientStock error raised when confirming an order

### LINKED ISSUE #3020: Spree::Order::InsufficientStock error raised when confirming an order
### Steps to reproduce

See [this gist](https://gist.github.com/spaghetticode/35dd0d91625a65d75aee32e383af795c) for integration tests that expose the issue. Specifically the last spec is failing. 

When two users try to purchase the last item remaining from a non-backordeable stock location at the same time then the last one will experience an unhandled error `Spree::Order::InsufficientStock`. This happens only if there is a second backorderable stock location for the product.

The error is generated by this line in `Spree::Order` model:

```ruby
before_transition to: :complete, do: :validate_line_item_availability
```

So, controller validations (provided here via `before_action :ensure_sufficient_stock_lines`) are following a different path, calling different methods, and this causes that error to be raised later in the order finalization process. Unifying the process seems to me the most reasonable way to fix the issue.

As a side note, but I think this is quite relevant, if the customer restarts the checkout process at least from the `delivery` step then the error does not happen anymore, as the order shipments are recreated from scratch this time using the backorderable stock location.

### Expected behavior

IMHO the customer should be redirected to the checkout `delivery` page, showing them an error message that suggests to restart the checkout process due to some changes in product availabilities. 

### Actual behavior

When the customer presses the order confirmation button then checkout process is halted by `Spree::Order::InsufficientStock`. The customer experiences a `500 server error` page.

### System configuration
**Solidus Version**:

`master` and previous versions, at least down to `2.6.x`

**Extensions in use**:

none

rubenochiavone:
# Description

When two users try to purchase the last item remaining from a non-backordeable stock location at the same time then the last one will experience an unhandled error `Spree::Order::InsufficientStock`. This happens only if there is a second backorderable stock location for the product.

## Fix

Improved `Spree::Order.insufficient_stock_lines` method to also map `@order.line_items` that are not validated using `Spree::Stock::AvailabilityValidator`.

## Tests

Added tests from @spaghetticode's [gist](https://gist.githubusercontent.com/spaghetticode/35dd0d91625a65d75aee32e383af795c/raw/184f921564770c1d2c8c1006d9e1714171b3f6cd/checkout_insufficient_stock_spec.rb). Had to improve on it since now validation prevents user to advance to next checkout page.

Closes #3020.

# Checklist:

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -465,7 +465,8 @@ module Spree
     end
 
     def insufficient_stock_lines
-      line_items.select(&:insufficient_stock?)
+      availability_validator = Spree::Stock::AvailabilityValidator.new
+      line_items.select { |line_item| line_item.insufficient_stock? || !availability_validator.validate(line_item) }
     end
 
     ##
--- a/frontend/spec/features/checkout_insufficient_stock_spec.rb
+++ b/frontend/spec/features/checkout_insufficient_stock_spec.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe "Checkout confirm page", type: :feature do
+  include_context 'checkout setup'
+
+  context 'when there is not enough stock at the default stock location' do
+    context "when the product is not backorderable" do
+      let(:user) { create(:user) }
+
+      let(:order) { Spree::TestingSupport::OrderWalkthrough.up_to(:payment) }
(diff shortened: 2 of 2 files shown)


rubenochiavone:
Didn't saw https://github.com/solidusio/solidus/pull/3023 :smile: since it was not linked with issue

PR #3037, opened 2019-01-16
Add countries to state selection for zones

jacobherrington:
# Description

This PR adds countries to the state selection when editing a zone. 

screen shot 2019-01-15 at 8 10 31 pm

This helps you keep from accidentally trying to ship things to Maryland, Liberia when you'd like to ship things to Maryland, USA.  😬 

# Checklist:

~- [ ] Documentation/Readme have been updated accordingly~
~- [ ] Changes are covered by tests (if possible)~

### DIFF
--- a/core/app/models/spree/state.rb
+++ b/core/app/models/spree/state.rb
@@ -36,5 +36,9 @@ module Spree
     def to_s
       name
     end
+
+    def state_with_country
+      "#{name} (#{country})"
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


core/app/models/spree/state.rb
```diff
@@ -36,5 +36,9 @@ def <=>(other)
     def to_s
       name
     end
+
+    def state_with_country
+      "#{name}, (#{country})"
```
aitbw: I think we could remove the comma from here, WDYT?

jacobherrington: I'm fine with whatever, this is just a minor annoyance we noticed a couple of days ago. Whatever makes this easy to understand. 😄

jacobherrington:
@kennyadsl done!

PR #3038, opened 2019-01-16
Remove state machine gem from Spree::Shipment (updated PR #2656)

cedum:
This resolves the existing conflict and rebases from the latest `master` the PR #2656 by @jgayfer:

> The [state_machines](https://github.com/state-machines/state_machines) gem has been used throughout Solidus since the earliest days of Spree. We use it to track the "state" our shipments/payments/orders are in, as well as plug custom functionality into the transitions between those states (`before_transition`/`after_transition`).
>
> However, our use of intermingled state machines between Payments, Shipments and Orders causes problems for both new and established developers. The state machine transitions are difficult to understand and debug, the order the transitions are executed in can be tough to control, and properly controlling database transactions during a transition is very poorly understood.
>
> We believe the downsides of implementing our state machines using the state_machines gem to be a net negative for the project.
>
> This PR changes the `Spree::Shipment` class, explicitly defining the important methods that would have been generated from the state_machines metaprogramming. The advantages of doing this are improved code clarity, debug-ability, as well as making it more easy to completely override methods regarding state.
>
> The first two commits of this pull request remove the `state_machine` from [`Spree::Shipment`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/shipment.rb) without changing the external API of the model. The last two commits go a bit further and deprecate the [transition methods](https://github.com/solidusio/solidus/blob/fd474db749d4386f52a23efbb59d30336351028d/core/app/models/spree/shipment.rb#L79-L91) that were used by the state machine in an attempt to improve code readability. If we don't want to go ahead and deprecate the transition methods, the first two commits can just be taken on their own, keeping the external API of [`Spree::Shipment`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/shipment.rb) intact.
>
>While we could re-implement the `before_transition`/`after_transition` calls on the `Spree::Shipment` model, we believe the following examples would be a better way to support customising functionality before or after a transition (with `ship` used as an example).
>
> ```ruby
> module AfterShipAction
>   # Note that returning true/false is done to preserve the original return values of ship
>   # This pattern works for all of ship, resume, pend, ready, and cancel
>   def ship
>     if super
>       additional_after_ship_procedure
>     end
>   end
> 
>   def additional_after_ship_procedure
>     # Do something interesting!
>     true
>   end
> end
> 
> module BeforeShipCondition
>   def ship
>     if pre_ship_condition
>       super
>     end
>   end
> 
>   def pre_ship_condition
>     # Some interesting condition as to when we can ship
>     true
>   end
> end
> 
> Spree::Shipment.prepend AfterShipAction
> Spree::Shipment.prepend BeforeShipCondition
> ```

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -161,14 +161,17 @@ module Spree
     def can_transition_from_pending_to_shipped?
       !requires_shipment?
     end
+    deprecate can_transition_from_pending_to_shipped?: '!requires_shipment?', deprecator: Spree::Deprecation
 
     def can_transition_from_pending_to_ready?
       inventory_can_ship?
     end
+    deprecate can_transition_from_pending_to_ready?: :inventory_can_ship?, deprecator: Spree::Deprecation
 
     def can_transition_from_canceled_to_ready?
       can_transition_from_pending_to_ready?
     end
+    deprecate can_transition_from_canceled_to_ready?: :inventory_can_ship?, deprecator: Spree::Deprecation
 
     extend DisplayMoney
     money_methods(
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -4,6 +4,14 @@ module Spree
   # An order's planned shipments including tracking and cost.
   #
   class Shipment < Spree::Base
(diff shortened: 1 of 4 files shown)


kennyadsl:

PR #3039, opened 2019-01-16
Remove state machine gem from Spree::Payment (updated PR #2664)

cedum:
This rebases from the latest master the PR #2664 by @jgayfer:

> This PR removes the state machine gem from `Spree::Payment` while keeping the external API intact. The logic that was previously hidden within the state machine is now contained within the `Spree::Payment` model.
> 


kennyadsl:

PR #3040, opened 2019-01-16
Remove state machine gem from Spree::Reimbursement (updated PR #2666)

cedum:
This rebases from the latest master the PR #2666 by @jgayfer:

> This PR removes the state machine gem from `Spree::Reimbursement` while keeping the external API intact. The logic that was previously hidden within the state machine is now contained within the `Spree::Reimbursement` model.
> 


kennyadsl:

PR #3041, opened 2019-01-16
Remove state machine gem from Spree::InventoryUnit (updated PR #2668)

cedum:
This rebases from the latest master the PR #2668 by @jgayfer:

> This PR removes the state machine gem from `Spree::InventoryUnit` while keeping the external API intact. The logic that was previously hidden within the state machine is now contained within the `Spree::InventoryUnit` model.
> 


kennyadsl:

PR #3042, opened 2019-01-17
Updating readme to include OpenCollective information

seand7565:
Adds OpenCollective badges to the readme - as well as a Key Stakeholder avatars.

Related to PR #2931

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,24 @@
 - [solidus-security](https://groups.google.com/forum/#!forum/solidus-security) mailing list
 
 ## Table of Contents
-1. [Summary](#summary)
-2. [Demo](#demo)
-3. [Getting Started](#getting-started)
-4. [Installation Options](#installation-options)
-5. [Performance](#performance)
-6. [Developing Solidus](#developing-solidus)
-7. [Contributing](#contributing)
+1. [Key Stakeholders](#key-stakeholders)
+2. [Summary](#summary)
+3. [Demo](#demo)
+4. [Getting Started](#getting-started)
+5. [Installation Options](#installation-options)
+6. [Performance](#performance)
+7. [Developing Solidus](#developing-solidus)
+8. [Contributing](#contributing)
+
+## Key Stakeholders
+
+Thank you to all our donors! 🙏 [Become a donor](https://opencollective.com/solidus#backer)
+
+Support this project by becoming a Key Stakeholder. Your logo will show up here with a link to your website. [Become a Key Stakeholder](https://opencollective.com/solidus#key-stakeholder)
+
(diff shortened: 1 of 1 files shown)


jacobherrington:
Re-running specs, touching a markdown file shouldn't case a feature spec to fail.

README.md
```diff
@@ -275,3 +278,12 @@ If you want to write an extension for Solidus, you can use the
...
 [CONTRIBUTING.md](CONTRIBUTING.md) before contributing.
+
+
+## Sponsors
+
+Thank you to all our backers! 🙏 [Become a backer](https://opencollective.com/solidus#backer)
+
+Support this project by becoming a Key Stakeholder. Your logo will show up here with a link to your website. [Become a Key Stakeholder](https://opencollective.com/solidus#key-stakeholder)
+
+<img src="https://opencollective.com/solidus/tiers/key-stakeholder.svg">
```
kennyadsl: This image links to itself for some reason and this is not ideal since we are saying on the previous line:

> Your logo will show up here with a link to your website.

README.md
```diff
@@ -275,3 +278,12 @@ If you want to write an extension for Solidus, you can use the
 
 Solidus is an open source project and we encourage contributions. Please read
 [CONTRIBUTING.md](CONTRIBUTING.md) before contributing.
+
+
+## Sponsors
+
```
kennyadsl: We are using `Sponsor` here and saying thanks to `bakers` into the text. This is kinda confusing and I think we should explain better the difference between the two.

README.md
jacobherrington: Ah, so we may have to add those images manually. @seand7565 do you think you can do that for us? I can keep up with it as more people join that tier. 😄

README.md
jacobherrington: This is true because in OC those are names for the tiers we are using. Backer is $10-50 and Sponsor is $100-500.

Maybe `## Sponsors` can change to `## Key Stakeholders` and perhaps we can say: 
``` 
Thank you to all our donors! 🙏 [Become a donor](https://opencollective.com/solidus)
```

I'd also like to see this higher in the readme to demonstrate the support + provide more value to those willing to donate such a significant amount of money.

README.md
seand7565: Yeah, absolutely. How should I deal with the image storage? Directly on the repo, like the Solidus logo? https://github.com/solidusio/solidus/blob/master/logo.svg

Or do we have another place to store these images?

jacobherrington: Can we grab them from OpenCollective profiles? For example `https://images.opencollective.com/proxy/images?src=https%3A%2F%2Flogo.clearbit.com%2Fenginecommerce.com`

seand7565: You got it! https://github.com/seand7565/solidus/tree/readme_oc#key-stakeholders

README.md
seand7565: Agreed and changed. How's that look? https://github.com/seand7565/solidus/tree/readme_oc#key-stakeholders

jacobherrington:
I'm happy with this as is, what do you think @kennyadsl?

PR #3043, opened 2019-01-18
Fix Spree::Variant inconsistency due to lack of product association

### LINKED ISSUE #3036: Spree::Variant could not be consistent
**Describe the bug**

Right now we can have a `Spree::Variant` instance that is not consistent since the price cannot be created without a product associated with the variant.

When [set_price]
(https://github.com/solidusio/solidus/blob/8c8986a092c9eb989811856f2eed4e2a1b212e90/core/app/models/spree/variant.rb#L75) is called before validation it raises an error if the product is not set.

**Solidus Version**:

master

**To Reproduce**

1. Open a rails console on a sandbox generated on `master`
2. Run: 
```ruby
v = Spree::Variant.new
=> #<Spree::Variant id: nil, sku: "", weight: 0.0, height: nil, width: nil, depth: nil, deleted_at: nil, is_master: false, product_id: nil, cost_price: nil, position: nil, cost_currency: nil, track_inventory: true, tax_category_id: nil, updated_at: nil, created_at: nil>
irb(main):090:0> v.valid?
        1: from (irb):90
RuntimeError (No master variant found to infer price)
```

**Expected behavior**

It should not raise an exception and an error should be added to the object instead.

rubenochiavone:
# Description

`Spree::Variant` new instance throw error during validation because it lacks a product association. The lack of product causes

- Error to be raised because price is inferred using `product.master.price`
- `Module::DelegationError` to be raised because to generate VAT prices, `variant.tax_category` is required and this method is delegated to `product.tax_category`

To fix it, removed raising an error if price can't be inferred using `product.master.price`. Reporting missing product is enough for an invalid variant instance. Also add product verification before generating VAT prices for variant.

Closes #3036

# Checklist:

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -72,9 +72,10 @@ module Spree
       autosave: true
 
     before_validation :set_cost_currency
-    before_validation :set_price
-    before_validation :build_vat_prices, if: -> { rebuild_vat_prices? || new_record? }
+    before_validation :set_price, if: -> { product && product.master }
+    before_validation :build_vat_prices, if: -> { rebuild_vat_prices? || new_record? && product }
 
+    validates :product, presence: true
     validate :check_price
 
     validates :cost_price, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
@@ -418,10 +419,7 @@ module Spree
 
     # Ensures a new variant takes the product master price when price is not supplied
     def set_price
-      if price.nil? && Spree::Config[:require_master_price] && !is_master?
-        raise 'No master variant found to infer price' unless product && product.master
-        self.price = product.master.price
-      end
+      self.price = product.master.price if price.nil? && Spree::Config[:require_master_price] && !is_master?
     end
 
     def check_price
--- a/core/spec/models/spree/variant_spec.rb
+++ b/core/spec/models/spree/variant_spec.rb
@@ -3,6 +3,8 @@
 require 'rails_helper'
 
 RSpec.describe Spree::Variant, type: :model do
+  it { is_expected.to be_invalid }
+
   let!(:variant) { create(:variant) }
 
   it_behaves_like 'default_price'
@@ -21,6 +23,8 @@ RSpec.describe Spree::Variant, type: :model do
     it "should require a product" do
       expect(variant).to be_valid
       variant.product = nil
+      expect(variant).to be_invalid
+      variant.price = nil
       expect(variant).to be_invalid
     end
   end


core/app/models/spree/variant.rb
```diff
@@ -418,10 +418,7 @@ def set_master_out_of_stock
 
     # Ensures a new variant takes the product master price when price is not supplied
     def set_price
-      if price.nil? && Spree::Config[:require_master_price] && !is_master?
-        raise 'No master variant found to infer price' unless product && product.master
-        self.price = product.master.price
-      end
+      self.price = product.master.price if price.nil? && Spree::Config[:require_master_price] && !is_master? && product && product.master
```
jacobherrington: I'm not sure about a guard clause being this long, but it's certainly not blocking.

core/app/models/spree/variant.rb
```diff
@@ -73,7 +73,7 @@ class Variant < Spree::Base
 
     before_validation :set_cost_currency
     before_validation :set_price
```
kennyadsl: what about moving `if: -> { product && product.master }` here?

core/app/models/spree/variant.rb
```diff
@@ -73,7 +73,7 @@ class Variant < Spree::Base
...
-    before_validation :build_vat_prices, if: -> { rebuild_vat_prices? || new_record? }
+    before_validation :build_vat_prices, if: -> { rebuild_vat_prices? || new_record? && product }
 
     validate :check_price
```
kennyadsl: What about adding a `presence` validation on the `product` here? There are too many places where Solidus gives for granted that a product should be associated with a variant and I think it could make sense to don't allow saving a variant without a product associated. Also, this way users will have the right validation message attached to the invalid object.

core/app/models/spree/variant.rb
rubenochiavone: Agreed, it got quite long. Ruby's `statement if condition` is to good to don't use it. :smile:

@kennyadsl contribution might reduce it though - https://github.com/solidusio/solidus/pull/3043#discussion_r249722989.

core/app/models/spree/variant.rb
rubenochiavone: Sounds good

core/app/models/spree/variant.rb
rubenochiavone: :+1:

core/app/models/spree/variant.rb
jacobherrington: Yes, I agree with his feedback.

core/app/models/spree/variant.rb
rubenochiavone: Hm, after reviewing the code I'm having second thoughts. How can we validate product presence in variant if product requires a master variant to be present?

The funny thing is that product automatically builds master variant, but I was unable to find `build_master` method definition. Maybe I'm missing something :frowning_face: 

https://github.com/solidusio/solidus/blob/adc75347b93fa9a51195051530b892f3d3379486/core/app/models/spree/product.rb#L74-L76

Any idea?

kennyadsl: `build_master` is a [method coming from has_one association](https://guides.rubyonrails.org/association_basics.html#methods-added-by-has-one). That variant will be associated with the product automatically so I think it will be valid, even with the proposed product presence validation.

rubenochiavone: I see. :+1:

rubenochiavone: Done

core/app/models/spree/variant.rb
rubenochiavone: Done

core/spec/models/spree/variant_spec.rb
```diff
@@ -3,6 +3,13 @@
 require 'rails_helper'
 
 RSpec.describe Spree::Variant, type: :model do
+  context "new instance" do
```
spaghetticode: @rubenochiavone I think this example readability may benefit from using Rspec [implicit subject](https://relishapp.com/rspec/rspec-core/v/3-5/docs/subject/implicitly-defined-subject). In Rspec the default `subject` is an instance of the class being specified (similar to `described_class.new(args)` but without any arguments... see [here](https://relishapp.com/rspec/rspec-core/docs/metadata/described-class)):

```it { expect(subject).to be_invalid }```

or, even more [concisely](https://relishapp.com/rspec/rspec-core/docs/subject/one-liner-syntax):

```it { is_expected.to be_invalid }```

Comment (rubenochiavone) [reply core/spec/models/spree/variant_spec.rb]: Whoa, that's great and works perfectly! Do you mind if I add a reference to implicit subject?

```ruby
it { is_expected.to be_invalid } # implicitly defined subject, see https://relishapp.com/rspec/rspec-core/v/3-5/docs/subject/implicitly-defined-subject
```

kennyadsl: I'm 👍 on using this syntax but 👎 on adding a code comment. I think you could add a reference in the commit description instead (please keep it all in a single commit 🙂).

rubenochiavone: Ok, cool. :+1:

rubenochiavone: Done

kennyadsl:
@rubenochiavone thanks! 🎉

PR #3045, opened 2019-01-20
Extend Decorator documentation

jacobherrington:
This change adds an example of using ActiveRecord inside a model
decorator. Largely adapted from a gist:
https://gist.github.com/kushniryb/101c161eab8d2d2e2e35e4a17b781d1b

### DIFF
--- a/guides/source/developers/extensions/decorators.html.md
+++ b/guides/source/developers/extensions/decorators.html.md
@@ -34,7 +34,7 @@ total by $10 (or whatever your currency is).
 
 ## Using ActiveRecord methods in decorators
 
-In order to access ActiveRecord methods, you'll need to define a special
+In order to access some class-level methods, you'll need to define a special
 method.
 
 ```ruby
@@ -46,8 +46,6 @@ module MyStore::ProductDecorator
     base.has_many :comments, dependent: :destroy
     base.scope    :sellable, -> { base.where(...).order(...) }
     base.delegate :something, to: :something
-
-    base.singleton_class.prepend ClassMethods
   end
 
   ...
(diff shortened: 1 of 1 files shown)


guides/source/developers/extensions/decorators.html.md
```diff
@@ -32,6 +32,34 @@ total by $10 (or whatever your currency is).
 
 [monkey-patch]: https://en.wikipedia.org/wiki/Monkey_patch
 
+## Using ActiveRecord methods in decorators
```
kennyadsl: I'd not say `ActiveRecord methods` here, they are more generic class-level methods. For example, the `acts_as_something` methods coming from other libraries are not `ActiveRecord`.

guides/source/developers/extensions/decorators.html.md
```diff
@@ -32,6 +32,34 @@ total by $10 (or whatever your currency is).
...
+module MyStore::ProductDecorator
+
+  # This is the place to define custom associations, delegations, scopes and
+  # other ActiveRecord stuff
+  def self.prepended(base)
+    base.has_many :comments, dependent: :destroy
+    base.scope    :sellable, -> { base.where(...).order(...) }
+    base.delegate :something, to: :something
+
+    base.singleton_class.prepend ClassMethods
```
kennyadsl: I think this line is not really needed, we should have the same result without it.

kennyadsl: To expand a little bit, in the example gist provided they also have:

```ruby
    # Class methods
    module ClassMethods
      def ransackable_associations
        %w(...)
      end
    end
```

which contains methods injected at a class level but we haven't added that module into the example so that prepend is not needed. I'm fine both with and without adding that extra scenario.

jacobherrington: I'd rather have less code in the example where possible

guides/source/developers/extensions/decorators.html.md
jacobherrington: Changed to class-level, does that make sense?

PR #3047, opened 2019-01-23
Add Api::CouponCodesController#destroy endpoint

### LINKED ISSUE #1641: Remove Coupon Code?
Currently, we can apply coupon code via endpoint `/api/orders/:id/apply_coupon_code`.

However, how to remove a coupon code?

aitbw:
# Description

This PR aims to close #1641 by adding a new API endpoint to the `CouponCodesController` to remove coupon codes from orders.

Said endpoint wasn't added to `CouponCodesController` under `frontend/`, but could be added if necessary.

Thanks to @flyfy1, @mtomov, and @loicginoux for their contributions on how to implement this, much appreciated! :raised_hands: :bowing_man:

# Checklist:

- [x] Documentation/Readme have been updated accordingly


core/app/models/spree/promotion_handler/coupon.rb
```diff
@@ -8,7 +8,7 @@ class Coupon
 
       def initialize(order)
         @order = order
-        @coupon_code = order.coupon_code && order.coupon_code.downcase
+        @coupon_code = order.coupon_code(&:downcase)
```
kennyadsl: I'm not sure about this. It's both not really related to this change and I think it works only since `Spree::Order#coupon_code` is a method but it won't work if you apply this syntax to a variable, for example:

```console
> coupon_code = 'TEst'
> coupon_code(&:downcase)
NoMethodError: undefined method `coupon_code' for main:Object


> def coupon_code; "TEst"; end
> coupon_code(&:downcase)
"TEst"
```

Also, I think it's not working as expected (the string is not downcased).

core/app/models/spree/promotion_handler/coupon.rb
```diff
@@ -25,6 +25,20 @@ def apply
         self
       end
 
+      def remove
+        if !coupon_code.present? || !promotion.present?
+          set_error_code :coupon_code_not_found
+        elsif !promotion_exists_on_order?(order, promotion)
+          set_error_code :coupon_code_already_removed
```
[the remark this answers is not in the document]
jacobherrington: Yes, I'd prefer `coupon_code_not_present`.

core/app/models/spree/promotion_handler/coupon.rb
```diff
@@ -25,6 +25,20 @@ def apply
         self
       end
 
+      def remove
+        if !coupon_code.present? || !promotion.present?
```
spaghetticode: As `present?` is `!blank?` (see [here](https://apidock.com/rails/Object/present%3F)) then  this can become `if coupon_code.blank? || promotion.blank?`

PR #3059, opened 2019-01-28
Unit cancel fix tests

mdesantis:
# Description

This follows along with #2401, and in addition to its `UnitCancel#adjust!` fix, it adds tests checking that, when `OrderCancellations#short_ship` is called with an `inventory_units` argument with multiple inventory units that do not map to the same objects of its `@order.inventory_units` it calculates the correct adjustment amount, where before #2401 they would fail.

Ref #2401

# Checklist:


kennyadsl:
Re-running CI, there's a flaky spec. 😠

PR #3061, opened 2019-01-29
Tentative fix for flaky specs in `backend/spec/features/admin/users_spec.rb`

spaghetticode:
**Description**

These specs are failing often on circleCI builds.
    
The added expectation ensures that the click toggled the table header HTML.
This has the double benefit to ensure HTML is consistent and leaves more time
to successfully verify the flaky expectation:

`expect(text_match_1).to appear_before text_match_2`

**Checklist:**

### DIFF
--- a/backend/spec/features/admin/users_spec.rb
+++ b/backend/spec/features/admin/users_spec.rb
@@ -57,19 +57,20 @@ describe 'Users', type: :feature do
   shared_examples_for 'a sortable attribute' do
     before { click_link sort_link }
 
-    it "can sort asc" do
+    it "can sort asc", :js do
       within_table(table_id) do
+        expect(page).to have_selector '.sort_link.asc'
         expect(page).to have_text text_match_1
         expect(page).to have_text text_match_2
         expect(text_match_1).to appear_before text_match_2
       end
     end
 
-    it "can sort desc" do
+    it "can sort desc", :js do
       within_table(table_id) do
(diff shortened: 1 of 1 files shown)


backend/spec/features/admin/users_spec.rb
```diff
@@ -57,19 +57,20 @@
   shared_examples_for 'a sortable attribute' do
     before { click_link sort_link }
 
-    it "can sort asc" do
+    it "can sort asc", :js do
```
kennyadsl: what about this `, :js`? Why is this needed?

spaghetticode: Not sure why it happens, but when using  `:rack_test` driver `expect(page).to have_selector '.sort_link.desc'` fails. It always finds `sort_link.asc` only.

These classes don't seem to be changed by javascript, as clicking the link generates a page reload, so I can't explain why:

```ruby
Capybara.current_driver = :selenium_chrome
=> :selenium_chrome
visit '/admin/users?q[s]=created_at+desc'
=> nil
expect(page).to have_selector '.sort_link.desc'
=> true
Capybara.current_driver = :rack_test
=> :rack_test
visit '/admin/users?q[s]=created_at+desc'
=> nil
expect(page).to have_selector '.sort_link.desc'
RSpec::Expectations::ExpectationNotMetError: expected to find visible css ".sort_link.desc" within #<Capybara::Node::Element tag="table" path="/html/body/div[5]/div/div/table"> but there were no matches
```

PR #3062, opened 2019-01-30
Testing tools improvements

kennyadsl:
**Description**

This PR: 

- bumps Capybara to 3.x (we still are on 2.x), here's the [upgrading guide](https://github.com/teamcapybara/capybara/blob/master/UPGRADING.md)
- removes Poltergeist from Gemfile, which is not used anymore
- switches Capybara server to Puma, to reflect the new standard in both Capybara and Rails. I don't think we'll have some real advantage using Puma over WEBRick, but this is the new standard and probably what all developers expect (I don't have strong opinions here). 

# Checklist:

### DIFF
--- a/backend/spec/features/admin/orders/order_details_spec.rb
+++ b/backend/spec/features/admin/orders/order_details_spec.rb
@@ -191,9 +191,14 @@ describe "Order Details", type: :feature, js: true do
       end
 
       context 'splitting to location' do
-        # can not properly implement until poltergeist supports checking alert text
-        # see https://github.com/teampoltergeist/poltergeist/pull/516
-        it 'should warn you if you have not selected a location or shipment'
+        it 'should warn you if you have not selected a location or shipment' do
+          visit spree.edit_admin_order_path(order)
+
+          within('tr', text: line_item.sku) { click_icon 'arrows-h' }
+          accept_alert 'Please select the split destination.' do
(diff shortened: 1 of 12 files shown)


backend/spec/features/admin/orders/order_details_spec.rb
```diff
@@ -191,9 +191,16 @@
...
-        it 'should warn you if you have not selected a location or shipment'
+        it 'should warn you if you have not selected a location or shipment' do
+          visit spree.edit_admin_order_path(order)
+
+          within('tr', text: line_item.sku) { click_icon 'arrows-h' }
+          click_icon :ok
+          alert = page.driver.browser.switch_to.alert
+
+          expect(alert.text).to eq 'Please select the split destination.'
+          alert.accept
```
aitbw: I think you can remove lines 199, 201, 202, and replace it with `accept_confirm { Please select the split destination }`

kennyadsl: Thanks for pointing this out @aitbw! 

Following [docs](https://github.com/teamcapybara/capybara/blob/0fe411a6dadf5028c12560ce77b43358ad1a019f/lib/capybara/session.rb#L610-L630) I've done this way at the end:

```ruby
accept_alert 'Please select the split destination.' do
  click_icon :ok
end
```

PR #3064, opened 2019-01-30
Add channel attribute to import order creation

jtapia:
#### Description

Based on documentation https://github.com/solidusio/solidus/blob/master/guides/source/developers/orders/overview.html.md the `channel` attribute is an option value for order creation, but in the importer https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/importer/order.rb never is referenced, this allows importer to use `channel` on order creation.


spaghetticode:
@jtapia thank you for this PR.

The `channel` attribute is not set explicitly, but at line [32](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/importer/order.rb?#L32) `order.update_attributes!(params)` adds it to the order.

PR #3066, opened 2019-01-30
Accept source as permitted attribute importing orders

jtapia:
#### Description
Sending the `source` data inside `payments` payload to `spree/api/orders#create` is returning an error:
`ActiveRecord::RecordInvalid: Validation failed: Source can't be blank`

### DIFF
--- a/api/app/controllers/spree/api/orders_controller.rb
+++ b/api/app/controllers/spree/api/orders_controller.rb
@@ -9,6 +9,9 @@ module Spree
       class_attribute :admin_order_attributes
       self.admin_order_attributes = [:import, :number, :completed_at, :locked_at, :channel, :user_id, :created_at]
 
+      class_attribute :admin_payment_attributes
+      self.admin_payment_attributes = [:payment_method, :amount, :state, source: {}]
+
       skip_before_action :authenticate_user, only: :apply_coupon_code
 
       before_action :find_order, except: [:create, :mine, :current, :index]
@@ -151,6 +154,14 @@ module Spree
       def permitted_shipment_attributes
         if can?(:admin, Spree::Shipment)
           super + admin_shipment_attributes
+        else
+          super
+        end
+      end
+
+      def permitted_payment_attributes
+        if can?(:admin, Spree::Payment)
+          super + admin_payment_attributes
         else
           super
         end
--- a/api/spec/requests/spree/api/orders_controller_spec.rb
+++ b/api/spec/requests/spree/api/orders_controller_spec.rb
@@ -815,6 +815,43 @@ module Spree
           expect(response.status).to eq 201
           expect(json_response["user_id"]).to eq(user.id)
         end
+
+        context "with payment" do
+          let(:params) do
+            {
+              payments: [{
+                amount: '10.0',
+                payment_method: create(:payment_method).name,
+                source: {
+                  month: "01",
+                  year: Date.today.year.to_s.last(2),
+                  cc_type: "123",
+                  last_digits: "1111",
+                  name: "Credit Card"
+                }
+              }]
+            }
+          end
+
+          context "with source" do
+            it "creates a payment" do
+              post spree.api_orders_path, params: { order: params }
+              payment = json_response['payments'].first
+
+              expect(response.status).to eq 201
+              expect(payment['amount']).to eql "10.0"
+              expect(payment['source']['last_digits']).to eql "1111"
+            end
+
+            context "when payment_method is missing" do
+              it "returns an error" do
+                params[:payments][0].delete(:payment_method)
+                post spree.api_orders_path, params: { order: params }
+                expect(response.status).to eq 404
+              end
+            end
+          end
+        end
       end
 
       context "updating" do


jtapia:
@spaghetticode I just added a test, please let me know what do you think

spaghetticode:
@jtapia I think the tests don't verify that the payment source is correctly saved. They only test that the controller action returns a 201 or a 404.

Also, I ran the tests commeting the code you added in Spree::Api::OrdersController and they still pass, so I'm a bit puzzled 🤔... did not investigate more though.

api/spec/requests/spree/api/orders_controller_spec.rb
```diff
@@ -815,6 +815,38 @@ module Spree
...
+                  month: "01",
+                  year: Date.today.year.to_s.last(2),
+                  cc_type: "123",
+                  last_digits: "1111",
+                  name: "Credit Card"
+                }
+              }]
+            }
+          end
+          let!(:payment_method) { create(:payment_method) }
```
spaghetticode: I think that, since you're using `payment_method` only once, you can inline this `let` directly where it's used and save one line without compromising readability: 

```ruby
amount: '10.0',
payment_method: create(:payment_method).name,
source: {
```

api/spec/requests/spree/api/orders_controller_spec.rb
```diff
@@ -815,6 +815,38 @@ module Spree
           expect(response.status).to eq 201
           expect(json_response["user_id"]).to eq(user.id)
         end
+
+        context 'with payment', js: true do
```
spaghetticode: I think `js: true` is not needed for these specs

api/spec/requests/spree/api/orders_controller_spec.rb
```diff
@@ -815,6 +815,38 @@ module Spree
...
+          end
+          let!(:payment_method) { create(:payment_method) }
+
+          context "with source" do
+            it "creates a payment" do
+              post spree.api_orders_path, params: { order: params }
+              expect(response.status).to eq 201
+            end
+
+            it "returns an error" do
```
spaghetticode: I think that this spec would benefit from a context that explains what's happening here, for example
`context "when payment_method is missing" do`

api/spec/requests/spree/api/orders_controller_spec.rb
```diff
@@ -815,6 +815,38 @@ module Spree
...
+                }
+              }]
+            }
+          end
+          let!(:payment_method) { create(:payment_method) }
+
+          context "with source" do
+            it "creates a payment" do
+              post spree.api_orders_path, params: { order: params }
+              expect(response.status).to eq 201
```
spaghetticode: I think it's worth expanding a bit the expectations for this spec. The title says that `it creates a payment` but actually here we just check that the response is successful. I think adding something like this may be an improvement: 
```
payment = json_response['payments'].first
expect(payment['amount']).to eql "10.0"
expect(payment['source']['last_digits']).to eql "1111" 
```

api/spec/requests/spree/api/orders_controller_spec.rb
jtapia: done here https://github.com/solidusio/solidus/pull/3066/files#diff-72468e52a59aedfe8a3443597f1a6f3aR836

api/spec/requests/spree/api/orders_controller_spec.rb
jtapia: done here https://github.com/solidusio/solidus/pull/3066/files#diff-72468e52a59aedfe8a3443597f1a6f3aR836

api/spec/requests/spree/api/orders_controller_spec.rb
jtapia: done here https://github.com/solidusio/solidus/pull/3066/files#diff-72468e52a59aedfe8a3443597f1a6f3aR836

api/spec/requests/spree/api/orders_controller_spec.rb
jtapia: done here https://github.com/solidusio/solidus/pull/3066/files#diff-72468e52a59aedfe8a3443597f1a6f3aR836

jtapia:
@spaghetticode please let me know what do you think, thanks

jtapia:
@spaghetticode should we remove the `Needs Tests` tag?

spaghetticode:
@jtapia let's wait for somebody from the core team to chime in for that.

api/app/controllers/spree/api/orders_controller.rb
```diff
@@ -9,6 +9,9 @@ class OrdersController < Spree::Api::BaseController
       class_attribute :admin_order_attributes
       self.admin_order_attributes = [:import, :number, :completed_at, :locked_at, :channel, :user_id, :created_at]
 
+      class_attribute :admin_payment_attributes
+      self.admin_payment_attributes = [:payment_method, :amount, :state, source: [:name, :month, :year, :cc_type, :last_digits]]
```
tvdeyen: Shouldn’t we allow all attributes a payment source can have? There is more than credit cards. Since this is admin only I am fine with 

```suggestion
      self.admin_payment_attributes = [:payment_method, :amount, :state, source: []]
```

jtapia: @tvdeyen done here https://github.com/solidusio/solidus/pull/3066/files#diff-173268f0a85c59ef13e0753726e16ba8R13, thanks

Issue #2924, opened 2018-10-24
Exception when trying to edit a variant without a default price


oneiros:
### Steps to reproduce

Configure a default currency in `config/initializers/spree.rb`:

```ruby
  config.currency = "EUR"
```

Log into the admin backend, select a product and on the tab "Variants" add a new variant.

Go to the tab "Prices" and under "Variant Pricing" edit the price of the newly created variant.

Set the Currency to "USD" and save the price.

Go back to the "Variants" tab and edit the variant.

### Expected behavior

At the very least, no exception should be raised

### Actual behavior

The following exception is raised:

```
ActionView::Template::Error (undefined method `currency' for nil:NilClass):
    65:       <div class="col-3">
    66:         <div class="field" data-hook="price">
    67:           <%= f.label :price %>
    68:           <%= render "spree/admin/shared/number_with_currency", f: f, amount_attr: :price, currency: @variant.default_price.currency %>
    69:         </div>
    70:       </div>
    71: 
  
solidus_backend (2.7.0) app/views/spree/admin/variants/_form.html.erb:68:in `_779db0d19337b31fccae3d69b3329154'
```

The reason is quite obvious: The partial expects a `default_price` to always be present. The `default_price` is being determined in ` solidus/core/app/models/concerns/spree/default_price.rb` using, amongst others, the scope `with_default_attributes`. This in turn uses the configured currency. So if no price exists for the configured currency, no `default_price` can be found.

I guess a possible solution would be to fix the view partial to not raise an exception (and I would be happy to prepare a PR for that), but I wonder if it should even be possible for a variant to not have a `default_price`. If no, the actual fix would be quite different.

### System configuration

**Solidus Version**: 2.7.0

**Extensions in use**:

* 'solidus_product_assembly'
* 'solidus_print_invoice'
* 'solidus_reviews'

jacobherrington:
I think not raising an exception is a good idea. I'm not too familiar with doing international stores, so anyone can correct me, but I think people should be able to edit prices for different currencies in this way.

jtapia:
I created this PR https://github.com/solidusio/solidus/pull/3067 that fixed that issue, let me know your thoughts

kennyadsl:
Closed with #2504, thanks for reporting the issue.

PR #3067, opened 2019-01-31
Fix issue when trying to edit a variant without a default price

jtapia:
#### Description
Issue reported on https://github.com/solidusio/solidus/issues/2924, that cause that the variant editing page gets broken if you choose another currency that the default

#### Issue
- https://github.com/solidusio/solidus/issues/2924

### DIFF
--- a/backend/app/views/spree/admin/variants/_form.html.erb
+++ b/backend/app/views/spree/admin/variants/_form.html.erb
@@ -65,7 +65,7 @@
       <div class="col-3">
         <div class="field" data-hook="price">
           <%= f.label :price %>
-          <%= render "spree/admin/shared/number_with_currency", f: f, amount_attr: :price, currency: @variant.default_price.currency %>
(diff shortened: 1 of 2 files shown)


jtapia:
@spaghetticode please try this case https://github.com/solidusio/solidus/pull/3067/files#diff-910a995f7d92d191856d89f34b920949R82

kennyadsl:
I think we should solve this issue by not letting users add a price with another currency if a price with the default currency has not yet been added to that variant. Maybe we can disable the input field and also a tooltip that explains this. WDYT?

kennyadsl:
Also, I think it's already fixed in https://github.com/solidusio/solidus/pull/2504, which I'd really love to merge. If someone wants to rebase that PR (or open it again with the same commits rebased on master) it would be great.

jtapia:
@kennyadsl I asked @softr8 to rebase the PR #2504, I'm going to close this PR

PR #3070, opened 2019-02-01
Bump the minimal Ruby version to 2.3

elia:
# Description

Support for 2.2 has been dropped on March, 2018:

https://www.ruby-lang.org/en/news/2018/03/28/ruby-2-2-10-released/

# Checklist:

- [x] Documentation/Readme have been updated accordingly

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,7 @@ AllCops:
     - '*/spec/dummy/**/*'
     - 'sandbox/**/*'
     - '**/templates/**/*'
-  TargetRubyVersion: 2.2
+  TargetRubyVersion: 2.3
 
 # Sometimes I believe this reads better
 # This also causes spacing issues on multi-line fixes
--- a/solidus.gemspec
+++ b/solidus.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.require_path = 'lib'
   s.requirements << 'none'
 
-  s.required_ruby_version     = '>= 2.2.2'
+  s.required_ruby_version     = '>= 2.3'
   s.required_rubygems_version = '>= 1.8.23'
 
   s.author       = 'Solidus Team'


kennyadsl:
We had some discussion on the same topic here https://github.com/solidusio/solidus/pull/2794. There are valid points to avoid doing this, unless until we support Rails 5.1.

coorasse:
I am also against dropping support for ruby 2.2 only because it reached EOL. Unless you face a specific issue and maintaining it would costs too much, I see no reason to drop the support.

kennyadsl:
An argument in favor of this is that we use some gems that are starting dropping support of Ruby < 2.3, for example [FactoryBot](https://github.com/thoughtbot/factory_bot/commit/27edb29a966365ca476d486323c1d7b60815b773). Using their latest version will force users to update to a more recent Ruby version anyway.

elia:
I'm ok with both directions, supporting older version is very kind to people stuck there (I've been that guy once or twice). 

On the other hand, upgrading from 2.2 to 2.3 is quite smooth (it's no 1.8 to 1.9!), and allows code to use [a couple new tricks](http://damien.la/2015/12/02/ruby2.3.0), like `&.` (this is what got me started in the first place) and `#dig`.

jacobherrington:
After revisiting the conversation I think we should support Ruby 2.2.2+ as that is what Rails 5 supports.

elia:
> After revisiting the conversation I think we should support Ruby 2.2.2+ as that is what [Rails 5 supports](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#ruby-versions).

👍 Closing then, can be easily reopened if the decision is revised.

One last thing I'd like to mention is that Rails 6 will require 2.5+, going from 2.2 to 2.3 seems to me a reasonable requirement in exchange for a bunch of language features. It  doesn't sound likely for a project to be able to upgrade to the latest solidus and being "stuck" on 2.2. 

My last 2¢ anyway 😄.

PR #3073, opened 2019-02-04
Relax rails dependency

opsb:
The rails version was pinned to 5.1.* which is overly restrictive as all versions 5.1 <= v < 6.0 will be backwards compatible.


kennyadsl:
@opsb I think you are looking at some old code. That dependency lock is not present in master, can you please take a look?

opsb:
apologies, PR is for a fork, will close this one

PR #3081, opened 2019-02-06
ActiveSupport notifications for Events Handling 

spaghetticode:
This is the code resulted from some experiments I did with `ActiveSupport::Notifications` after [this issue](https://github.com/solidusio/solidus/issues/2430). 

I think using ActiveSupport Notifications may simplify the handling of events, as most of the code is already implemented, if the Solidus team decides eventually to go that way.

A bonus of this implementation is that native `ActiveSupport` code is wrapped inside the `Spree::Event` module, which should be easy to customize in order to use another event library, if that may be the case... something like a custom event bus or an external gem like `rails_event_store`. 

Differently, a configuration settings may be introduced in order to allow to use an entirely different class instead of `Spree::Event`. 

I think the code is pretty simple and straightforward to understand, it can be surely improved but my primary goal was to provide a POC to be discussed.

I provided a couple of usage examples related to sending emails, as this is a hot topic for this feature.

The first one relates to the order finalization process, and now courtesy of `Spree::Event::MailerProcessor` the email is sent via event subscription. Removing the subscription is easy and can be done in an initializer, for example:

```Spree::Event.unsubscribe Spree::Event::MailerProcessor.order_finalize_subscription```

All subscriptions from `MailerProcessor` can be removed with this line: 

```Spree::Event::MailerProcessor.unregister!```

Some specs show the usage with code.

Adding more events should be quite straightforward by following the examples in the code.

### CODE REVIEW

File: core/lib/spree/event.rb [hunk 1]
```diff
@@ -0,0 +1,20 @@
+module Spree
```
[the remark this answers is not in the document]
spaghetticode: 🐕 the file is updated, but the hound did not acknowledge the change... resolving myself

core/lib/spree/event/mailer_processor.rb
```diff
@@ -0,0 +1,45 @@
+module Spree
```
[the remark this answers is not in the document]
spaghetticode: 🐕 the file is updated, but the hound did not acknowledge the change... resolving myself

mdesantis:
@spaghetticode , what about fully embracing the JavaScript metaphore and replace `publish`, `subscribe` and `unsubscribe` respectively with `trigger`, `on` and `off`?

spaghetticode:
@mdesantis I'm not sure about that. I feel that this is Ruby so there's no particular benefit in adhering to an interface/naming that it's not strictly ours. Also, `on off trigger` are mainly jQuery AFAIK .

Another reason why I'm not much into this change is that while I like easy to type (short) and remember methods, I learned in the past that they can be a PITA to search in a codebase, especially when they're as generic as `on` and `off` can be.

spaghetticode:
I've been working a bit more on this implementation. The main improvements are:

* Spree::Event now delegates the bulk of its methods to the `adapter`, which defaults to `Spree::Event::Adapter::ActiveSupportNotifications`. This will make super-easy switching to a different implementation by simply changing the adapter: 
`Spree::Event.adapter = Spree::EventBus`

* `Spree::Event.listeners` returns a hash with all Solidus related subscriptions, grouped by event name. The implementation is again delegated to the adapter, as `ActiveSupport::Notifications` already has a nice way for managing this

* `Spree::Event.unsubscribe` now accepts a subscription object (it will unsubscribe only that subscription) or an event name (it will unsubscribe all subscriptions to that event)

* some specs for `Spree::Event` module in order to improve coverage

core/lib/spree/core/engine.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'spree/config'
+require 'spree/event/mailer_processor'
```
kennyadsl: if we plan to have multiple event processors maybe we can change the folders/modules structure to accomodate this? Something like: `spree/event/processors/mailer`. WDYT?

core/lib/spree/event.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+require_relative 'event/adapters/active_support_notifications'
+require_relative 'event/configuration'
+
+module Spree
+  module Event
+    POSTFIX = '.spree'
```
kennyadsl: I always see `SUFFIX` for this kind of things. Anyway, what about making this configurable?

core/lib/spree/core/engine.rb
```diff
@@ -44,6 +45,11 @@ class Engine < ::Rails::Engine
         Migrations.new(config, engine_name).check
       end
 
+      # This code can be injected by an external gem/part of Solidus:
```
kennyadsl: I don't get the comment, can you please expand a little bit?

core/lib/spree/event.rb
spaghetticode: 👍 I think making this configurable is a nice addition

Regarding the naming, I think [both are fine](https://english.stackexchange.com/questions/81263/postfix-or-suffix) but yes, probably `suffix` is more common

core/lib/spree/core/engine.rb
spaghetticode: 👍makes sense

core/lib/spree/core/engine.rb
spaghetticode: I meant that this code (the initializer callback) could be moved to another gem, for example to a solidus extension/part that is responsible of sending emails. 

More in general adding subscriptions using an initializer seems to me reasonably simple and convenient, it may be the standard way to do this... what do you think?

kennyadsl: Ok, I agree that it can be moved but I'd remove the comment then.

core/lib/spree/event/adapters/active_support_notifications.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Spree
+  module Event
+    module Adapters
+      module ActiveSupportNotifications
```
skukx: This is more of a question and less of a request. Would it be worth implementing more of an interface or strategy pattern? So instead of defining a module, have a base abstract class which all adapters should inherit from an therefore expected to implement `instrument`, `subscribe`, etc.

I know I'm splitting hairs here but I love to learn

core/app/models/spree/order.rb
```diff
@@ -418,23 +418,23 @@ def valid_credit_cards
...
-      updater.update_payment_state
-      shipments.each do |shipment|
-        shipment.update_state
-        shipment.finalize!
-      end
-
-      updater.update_shipment_state
-      save!
-      updater.run_hooks
+      Spree::Event.instrument 'order_finalize', order: self do
```
skukx: I think we should have a constant for each event topic.
```ruby
Spree::EventTopics::ORDER_FINALIZE
#or 
Spree::EventTopics::Order::FINALIZE
```

skukx:
I'm really excited for this!

core/lib/spree/event.rb
```diff
@@ -0,0 +1,110 @@
...
+    # @see Spree::AppConfiguration
+    def adapter
+      Spree::Config.events.adapter
+    end
+
+    # The suffix used for namespacing Solidus events, defaults to
+    # `.spree`
+    #
+    # @see Spree::Event::Configuration#suffix
+    def suffix
```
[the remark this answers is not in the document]
skukx: I just read that `Rails` uses the suffix convention which I find interesting.
See: https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events

> You should follow Rails conventions when defining your own events. The format is: event.library. If your application is sending Tweets, you should create an event named tweet.twitter.

I'm not a fan of it, but probably should stick to rails convention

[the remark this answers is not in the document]
spaghetticode: Yes, that's the reason why I chose the suffix. Sticking with conventions is usually the best choice 👍

core/lib/spree/event/adapters/active_support_notifications.rb
spaghetticode: I guess it mostly depends on specific situations. The main advantage of using an adapter is that you can easily change the adapter at runtime and still call the same interface and get the work done. 

If you are thinking about any specific implementation that may work better in this situation then please share it, all improvements are welcome :)

core/app/models/spree/order.rb
spaghetticode: Yes, maybe. The benefit I can see for using constants is that we can kind of document all the events at the top of each class.  It would be easier to spot all of them. Do you have other benefits in mind? 

There's no hard constraint that would force devs to use constants, so the rule can be easily bypassed. A rule easily bypassable is usually not worth much.

AFAIK Rails codebase does not use any constant with ActiveSupport notifications, so maybe we can use the same direct approach without loosing much? Still pondering our options...

skukx: I think you're right, we don't lose much by not having constants. However thinking from and extension stand point it would be easier to manage if constants were available.

Lets say I have many subscribers in my codebase defined with the static string `'order_finalize'` and one day down the road that topic name changes. When I run bundle update, all of sudden my code breaks.  If I were to define my subscribers with the constant, running bundle update wouldn't break a thing. In general I always think it's best to avoid magic values when possible

core/lib/spree/event/adapters/active_support_notifications.rb
skukx: I think this is great, I say splitting hairs because they all accomplish the same exact thing. Some are just more strictly enforced than others

core/app/models/spree/order.rb
spaghetticode: 👍 I agree constants are stronger than simple strings, but constant renaming and behavior changes may anyway break things in the future. I'm still thinking about other ways to make this more robust. Let's wait for some other feedback on this, thank you!

kennyadsl: > Lets say I have many subscribers in my codebase defined with the static string `'order_finalize'` and one day down the road that topic name changes. When I run bundle update, all of sudden my code breaks.

That's true, even if no one ensures us that someone just skipped the usage of the constant and used its content (the string) instead to subscribe to events. In that case, those stores would break in the same way. I imagine a way to change an event name could be:

```ruby
Spree::Event.instrument 'new_event_name', order: self do
  # Do stuff

  # Instrument the old event as well, so old subscribers will be happy
  Spree::Event.instrument 'old_event_name', order: self
end
```

We should also find a way in the Spree::Event#subscribe to deprecate events so subscribers to old events would be notified that it's time to change the name of the event.

My point is that if we decide to use constants but we still need to implement something similar cause we can't be sure devs used strings or constants while subscribing to event, I can't see a real advantage introducing them.

skukx: I agree, I've come to a similar conclusion. It would make more sense for the extension that may subscribe many times to define the constant themselves at their discretion.

skukx:
@tvdeyen I can't answer all but maybe a few of the issues you raised.

> Why do we have the concept of adapters
This is what I actually like most about this implementation. Using the adapter pattern allows developers to not be locked into one event store.  For instance if I wanted to use `Redis` as in event store, I could write an adapter and plug it in to use that instead.  

What I don't know, is that if I already have redis setup for my ActiveSupport caching, is it essentially accomplishing the same thing?

> Spree::Event.instrument does not seem like a good name for dispatching events. fire, commit, dispatch seem better candidates IMO

I believe the reason `instrument` is used is because it mimics the rails api for issuing notifications:
https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events

core/spec/lib/spree/event_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+          let!(:subscription) { Spree::Event.subscribe(subscription_name) { item.do_something } }
+
+          before do
+            Spree::Event.subscribe(subscription_name) { item.do_something_else }
+          end
+
+          it 'can unsubscribe from single event by object' do
+            subject.unsubscribe subscription
+            Spree::Event.instrument subscription_name
+            expect(item).not_to have_received :do_something
```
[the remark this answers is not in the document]
spaghetticode: Sure, I hope this clarifies the spec. 

This line adds a subscription to the event. When the event is fired `item.do_something` will be called:
```ruby
let!(:subscription) { Spree::Event.subscribe(subscription_name) { item.do_something } }
```
in the `before` block a new subscription is added, so there's going to be a total of two subscriptions for the event:
```ruby
Spree::Event.subscribe(subscription_name) { item.do_something_else }
```
But before firing the event, we unsubscribe the first subscription:
```ruby
subject.unsubscribe subscription
```
Later the rspec example verifies that the first block is not called (subscription removed), but it verifies that the second one is. This also serves the purpose of checking that when you remove a subscription all other subscriptions are still in place.

core/spec/lib/spree/event_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+    describe '#listeners' do
+      context 'when there is no subscription' do
+        it { expect(subject.listeners).to be_empty }
+
+        context 'after adding a subscription' do
+          before do
+            Spree::Event.subscribe(subscription_name) { item.do_something }
+          end
+
+          it 'includes the new subscription with custom postfix' do
```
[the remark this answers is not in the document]
spaghetticode: They mean the same thing 😄 but surely I agree it's better to be consistent with naming also here 👍

core/spec/models/spree/reimbursement_spec.rb
```diff
@@ -154,7 +154,8 @@
     context 'when reimbursement cannot be fully performed' do
       let!(:non_return_refund) { create(:refund, amount: 1, payment: payment) }
 
-      it 'raises IncompleteReimbursement error' do
+      it 'does not send a reimbusement email and raises IncompleteReimbursement error' do
```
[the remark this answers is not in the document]
spaghetticode: 👍

spaghetticode:
@skukx thanks for your answer to @tvdeyen questions, what you wrote is much similar to what floats in mind 👍

Still I'd like to expand a bit on the questions: 

> Why wrap the code in a block for instance? Active support uses this to instrument the code, but we don’t have to
I'm not 100% sure what you mean here... if are you referring for example to the block enclosed in this:
```ruby
Spree::Event.instrument 'order_finalize', order: self do
  all_adjustments.each(&:finalize!)
  # ...
end
```
then yes, there's no _real_ need to wrap the code inside a block. We could move the event instrumentation code at the bottom of the method, but I personally prefer the other way for some reasons:
* the block clarifies what actions are part of the order finalization process (you may argue that in this specific example the method is already named  `finalize!`, so there's no need to add another wrapper, but this is just a basic example... if I had to stick to a convention I'd wrap rather than not, at the cost of being sometimes redundant)
* you get the ability to do what Rails does, for example calculating the time required to perform the action (the code inside the block)

Anyway, I think both ways can be used depending on the situation

> We need to have more events than just one when something updates.

The examples I provided are just examples that show how to change some previous code to something that is still very similar but can emit events we can leverage on. The order finalization event was a low-hanging fruit, as the method was already there, and it made sense being able to subscribe to an event about _order finalization_. 

Of course I agree this is just a step in the process, so there's much more that can be done, but I personally don't feel comfortable in modifying the existing processes to that extent, at least not now. 
Maybe someone more experienced than me with Solidus and its processes can come up with better examples that go in that direction. 

> Spree::Event.instrument does not seem like a good name for dispatching events. fire, commit, dispatch seem better candidates IMO

I also rather prefer a different naming, but I'd stick with ActiveSupport names for these reasons:
* we're all Rails programmers, so it makes totally sense to me using the same interface (many of us are possibly already accustomed to it)
* ATM the only provided adapter is for ActiveSupport::Notification, so to me it makes sense sticking with its interface
* I don't know how successful the ActiveSupport::Notification adapter is going to be, but if it is going to be the default one, I think choosing the same interface makes sense

mdesantis:
Regarding `instrument` with or without block: I think that if the reported timings are not used the block is redundant and even misleading: looking at it it tells me that there is a reasoning behind it and that the timings are used somewhere.

> the block clarifies what actions are part of the order finalization process (you may argue that in this specific example the method is already named finalize!, so there's no need to add another wrapper, but this is just a basic example...

Exactly, there's the method for that

> if I had to stick to a convention I'd wrap rather than not, at the cost of being sometimes redundant

if I had to stick to a convention I'd not wrap rather than yes, because we are using `ActiveSupport::Notification` as event dispatcher, we aren't using it as timings reporter, so the blocks are redundant

> you get the ability to do what Rails does, for example calculating the time required to perform the action (the code inside the block)

But since does timings are not used it just adds unneeded complexity; also, even if you need those timings, the code wrapped by the block should _accidentally_ correspond with the code that you need to measure, so it's pretty unreliable to me

tvdeyen:
@spaghetticode I agree with @mdesantis here.

Only because the `ActiveSupport::Notifications` is our current implementation we should not use its interface to design ours and not tightly couple us to it. Rather abstract the interface into something that can easily be used in our code case and is adapter agnostic. All the integration work should be done in the adapter instead.

I am in favor of 

- Renaming `Spree::Event.instrument` into `Spree::Event.fire`
- Making it simple call that does not wrap code in a block

So the only change we need to make is add a single line into the order `finalize!` method:

```rb
Spree::Event.fire('order_finalize', order: self)
```

spaghetticode:
@tvdeyen @mdesantis thank you for your feedback.

As no other feedback came supporting my opinions I decided it was best to follow your suggestions. So I removed the blocks in the examples, and renamed `Spree::Event::instrument` as `Spree::Event::fire`.

core/app/models/spree/reimbursement.rb
```diff
@@ -111,11 +111,14 @@ def perform!(created_by: nil)
...
         reimbursed!
-        reimbursement_success_hooks.each { |h| h.call self }
-        send_reimbursement_email
+        reimbursement_success_hooks.each { |h| h.call self } # TODO: these should become event subscriptions
       else
         errored!
-        reimbursement_failure_hooks.each { |h| h.call self }
+        reimbursement_failure_hooks.each { |h| h.call self } # TODO: these should become event subscriptions
+      end
+      Spree::Event.fire 'reimbursement_perform', reimbursement: self
```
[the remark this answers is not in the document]
spaghetticode: @kennyadsl I think an event may suffice, but it depends on what we want to prioritize.

For example, in this case, the mailer processor acts only when the reimbursement processing was successful by using some simple logic: `if reimbursement.reimbursed?`

The rule of thumb I am suggesting to use here is to add events only when strictly necessary, otherwise there may be an event proliferation to the extreme that, for each `if` or `case` statement in a code block, a different event is fired. 
We're talking about two events instead of one here, so there's no real risk of overdoing it, but I think that having a conversation about _when_ adding new events right now may be beneficial. 

Of course limiting events has the downside that the subscriber has to implement some more logic, just like in the example above.

As both approaches have pros and cons, I think there's no clear winner, at least in my mind. What do you think?

core/lib/spree/core/engine.rb
spaghetticode: 👍

core/lib/spree/event/processors/mailer_processor.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Spree
+  module Event
+    module Processors
+      module MailerProcessor
+        SUBSCRIPTIONS = [:order_finalize_subscription, :reimbursement_perform_subscription]
```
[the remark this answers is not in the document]
spaghetticode: 👍

core/app/models/spree/reimbursement.rb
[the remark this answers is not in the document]
kennyadsl: I'm proposing two events mainly because I saw that in this specific case there are two hooks, one for success and one for failure. If we are suggesting users to replace hooks with events subscription we should at least provide the same hook points that we used to provide before. I mean that right now it's not used into Solidus but it could be something that stores or extensions are using and this is hard to know. 

Generally speaking, I think it's very hard to provide the right balance between too few and too many events.

core/lib/spree/event/processors/mailer_processor.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module Spree
+  module Event
+    module Processors
+      module MailerProcessor
```
skukx: Would it be worth providing a registration interface? Something like:
```ruby
module Spree
  module EventRegistration
    def subscriptions
      @subscriptions ||= []
    end

    def add_subscription(method_name = nil)
      subscriptions << method_name.to_sym
      send(:mattr_accessor, "#{method_name}_handle")
      subscriptions
    end

    def register!
      subscriptions.each do |subscription|
        instance_variable_set("@#{subscription}_handle", send(:subscription))
      end
    end

    def unregister!
      subscriptions.each do |subscription|
        Spree::Event.unsubscribe(send("#{subscription}_handle"))
      end
    end
  end
end
```

Then in any processor

```ruby
module Spree
  module Event
    module Processors
      module MailerProcessor
        extend self
        extend Spree::EventRegistration
 
        add_subscription :order_finalize
        add_subscription :reimbursement_perform

        def order_finalize
          # Do work
        end

        def reimbursement_perform
          # Do work
        end
      end
    end
  end
end
```

It's a bit much to look at, but may allow for more extensibility as well as shareability.

skukx: I know this probably isn't the route, but when obtaining a handle to subscribers it may be worth implementing a consistent methodology so that store may unsubscribe from out of the box subs.

kennyadsl: I think this is a good concern, we should provide one suggested way to do things that we know will be needed in stores and extensions.

spaghetticode: @skukx 👍 thanks for the valuable input, I'll be working on this!

core/app/models/spree/reimbursement.rb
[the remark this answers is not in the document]
spaghetticode: @kennyadsl 👍 
I'll update the code to include two events, I agree in this specific situation it makes more sense having two separate events.

spaghetticode:
I followed @skukx's idea on refactoring the `MailerProcessor` module and evetually extract an [abstraction](https://github.com/solidusio/solidus/pull/3081/files#diff-d6a7a37486ba2958ea8f0d091d2df618) that helps to create new subscriber modules. What came out is quite similar to what he was proposing 👍 

Rspec examples and documentation in the module file should help in understanding how this works.

It's still a two steps process: one first declares a method as `event_action` and, if the name of the method is not the same of the event, then the extra parameter `event_name: 'name_of_event'` must be specified. 

After the declarative steps, one must explicitly subscribe the events somewhere in the code (for example in an initializer). Here's some pseudocode example:   

```ruby
module EmailSender
  include Spree::Event::Subscriber

  event_action :order_finalized
  event_action :confirm_reimbursement, event_name: :reimbursement_reimbursed

  def order_finalized(event)
    Mailer.send_email(event.payload[:order])
  end

  def confirm_reimbursement(event)
    Mailer.send_email(event.payload[:reimbursement])
  end
end

EmailSender.subscribe!
```

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@
 
 ### Backend
 
+**Added Spree::Event **
+
+Solidus now includes an event library that allows to use different adapters.
+The default adapter is based on `ActiveSupport::Notifications` library.
+Events should allow developers to customize and extend Solidus behavior
+more easily by simply subscribing certain events. Sending emails may be a
+simple use case for this new feature.
```
kennyadsl: I think this should not be bold, maybe it's the space after `Added Spree::Event `?

spaghetticode: 👍 thanks for noticing it

kennyadsl:
@spaghetticode can you please rebase, it should remove the spec failure, I hope.

I think it's time to start considering to merge this one. Any comment here from the community at this point would be great.

core/spec/lib/spree/event_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  let(:notifier) { ActiveSupport::Notifications.notifier }
+
+  subject { described_class }
+
+  it 'has default adapter' do
+    expect(subject.adapter).to eql Spree::Event::Adapters::ActiveSupportNotifications
+  end
+
+  before do
+    # ActiveSupport::Notifications expose no interface to clean all
```
aitbw: I'd word this differently, like `ActiveSupport::Notifications does not provide an interface...`, WDYT?

guides/source/developers/events/overview.html.md
```diff
@@ -0,0 +1,55 @@
...
+
+The adapter can be changed using this code, for example in a initializer:
+
+```ruby
+Spree::Config.events.adapter = "Spree::EventBus.new"
+```
+
+## Subscribing to events
+
+`Spree::Event.subscribe` allows to subscribe to a certain event. The event name is mandatory, the optional block will be executed everytime  the event is
```
aitbw: Two things in here:

* There's an trailing whitespace after `everytime`
* `fired:` is on a new line, I think it should be on the same line as the paragraph above

guides/source/developers/events/overview.html.md
```diff
@@ -0,0 +1,55 @@
...
+
+`Spree::Event.fire` allows to fire an event. The event name is mandatory, then both a hash of options (it will be available as the event payload) and an optional code block can be passed: 
+
+```ruby
+Spree::Event.fire 'order_finalized', order: @order do
+  @order.finalize!
+end
+```
+
+This is an alternative way to basically have the same functionality but  without the block:
```
aitbw: Trailing whitespace after `but`

guides/source/developers/events/overview.html.md
spaghetticode: 👍

core/spec/lib/spree/event_spec.rb
spaghetticode: 👍

guides/source/developers/events/overview.html.md
spaghetticode: 👍

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@
 
 ### Backend
 
+**Added Spree::Event**
+
+Solidus now includes an event library that allows to use different adapters.
+The default adapter is based on `ActiveSupport::Notifications` library.
+Events should allow developers to customize and extend Solidus behavior
+more easily by simply subscribing certain events. Sending emails may be a
```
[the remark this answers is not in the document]
spaghetticode: 👍

guides/source/developers/events/overview.html.md
```diff
@@ -0,0 +1,55 @@
+# Events
+
+The `Spree::Event` module allows to fire and subscribe events inside Soldius codebase and extensions.
```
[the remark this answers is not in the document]
spaghetticode: 👍

PR #3088, opened 2019-02-07
Lock sqlite3 version to 1.3

### LINKED ISSUE #3087: sqlite3 dependency version conflicts with activerecord sqlite adapter
On February 04, 2019 sqlite3 1.4.0 has been released. This version breaks ActiveRecord 5.2.2 SQLite adapter, which requires sqlite3 ~> 1.3.6: https://github.com/rails/rails/blob/v5.2.2/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L12

**Solidus Version**: `2.9.0.alpha@a736ec34e`

**To Reproduce**

1. Checkout Solidus repo
2. Run `bundle exec rake sandbox`

**Current behavior**

It fails showing the following error:

```
rake aborted!
LoadError: Error loading the 'sqlite3' Active Record adapter. Missing a gem it depends on? can't activate sqlite3 (~> 1.3.6), already activated sqlite3-1.4.0. Make sure all dependencies are added to Gemfile.
/home/maurizio/Sviluppo/Nebulab/solidus/sandbox/config/environment.rb:5:in `<main>'
/home/maurizio/.asdf/installs/ruby/2.6.0/bin/bundle:23:in `load'
/home/maurizio/.asdf/installs/ruby/2.6.0/bin/bundle:23:in `<main>'

Caused by:
Gem::LoadError: can't activate sqlite3 (~> 1.3.6), already activated sqlite3-1.4.0. Make sure all dependencies are added to Gemfile.
/home/maurizio/Sviluppo/Nebulab/solidus/sandbox/config/environment.rb:5:in `<main>'
/home/maurizio/.asdf/installs/ruby/2.6.0/bin/bundle:23:in `load'
/home/maurizio/.asdf/installs/ruby/2.6.0/bin/bundle:23:in `<main>'
Tasks: TOP => db:drop => db:load_config => environment
(See full trace by running task with --trace)
```

**Expected behavior**

It should successfully create the sandbox

mdesantis:
Fixes #3087. It includes the following changes:

**Lock sqlite3 version to 1.3**

ActiveRecord 5.2.2 has sqlite3 dependency locked to `~> 1.3.6`, we need to reflect that in order to avoid conflicts due to sqlite3 1.4 version releasement.

**Fix sqlite3 version conflict on sandbox generation**

`rails new` generates a Gemfile with `gem 'sqlite3'` dependency declaration, but since sqlite3 1.4 has been released it conflicts with activerecord sqlite3 adapter dependency declaration, which is `sqlite3 ~> 1.3.6`.

**Checklist:**

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ group :backend, :frontend, :core, :api do
     when /postgres/
       gem 'pg', '~> 1.0', require: false
     else
-      gem 'sqlite3', require: false
+      gem 'sqlite3', '~> 1.3.6', require: false
       gem 'fast_sqlite', require: false
     end
   end
--- a/lib/sandbox.sh
+++ b/lib/sandbox.sh
@@ -49,6 +49,10 @@ group :test, :development do
 end
 RUBY
 
+# Ensure sqlite3 version to match ActiveRecord SQLite adapter requirement
+# (see https://github.com/solidusio/solidus/issues/3087 for details)
+sed -i "/gem 'sqlite3'/c\gem 'sqlite3', '~> 1.3.6'" Gemfile
+
 bundle install --gemfile Gemfile
 bundle exec rake db:drop db:create
 bundle exec rails g spree:install --auto-accept --user_class=Spree::User --enforce_available_locales=true


peterberkenbosch:
sweet! I was just working on this. Trying to figure out how to "hack" the generated `Gemfile`. This will work!

kennyadsl:
This has been fixed in [Rails](https://github.com/rails/rails/issues/35161) (I think just in master for now). 
What do you think about [this solution](https://github.com/thoughtbot/factory_bot_rails/commit/e6509920a69ed3b8922763251e26800ef32cb91e) made on factory_bot_rails using `gsub_file` instead of `sed`?

```ruby
gsub_file "Gemfile", /^gem 'sqlite3'$/, 'gem "sqlite3", "~> 1.3.6"'
```

I think it's a bit more "Rails friendly".

mdesantis:
@kennyadsl I guess they have a template for the new Rails app, but we don't have it, and I wouldn't add it just for this edit. Anyway, since in `lib/sandbox.sh` there is also

 ```sh
cat <<RUBY >> Gemfile

gem 'solidus', path: '..'
gem 'solidus_auth_devise', '>= 2.1.0'
gem 'rails-i18n'
gem 'solidus_i18n'

group :test, :development do
  platforms :mri do
    gem 'pry-byebug'
  end
end
RUBY
```

I could convert both of them to the "Rails template way", WDYT?

kennyadsl:
If possible, that would not be bad in my opinion. But maybe we can do that into another PR?

mdesantis:
@kennyadsl :+1: I'll do that in another PR

mdesantis:
@jacobherrington here we are! can you please check that what I wrote makes sense? I feel it could be expressed better

peterberkenbosch:
> This has been fixed in [Rails](https://github.com/rails/rails/issues/35161) (I think just in master for now).
> What do you think about [this solution](https://github.com/thoughtbot/factory_bot_rails/commit/e6509920a69ed3b8922763251e26800ef32cb91e) made on factory_bot_rails using `gsub_file` instead of `sed`?
> 
> ```ruby
> gsub_file "Gemfile", /^gem 'sqlite3'$/, 'gem "sqlite3", "~> 1.3.6"'
> ```
> 
> I think it's a bit more "Rails friendly".

This is not working in `sandbox.sh` since it's a shell script, not a ruby file.

kennyadsl:
@peterberkenbosch yeah, we talked about that. The only way to use `gsub_file` is moving that part into a Gemfile template and use that template with the `-m` option on the `rails new` command. Definitely not worth it. 🙂

PR #3089, opened 2019-02-07
Fix flaky specs in `backend/spec/features/admin/users_spec.rb`

spaghetticode:
*(This is a reopening for PR #3061 closed by mistake)*

These specs are failing often on circleCI builds.

The added expectation ensures that the click toggled the table header HTML.
This has the double benefit to ensure HTML is consistent and leaves more time
to successfully verify the flaky expectation:

`expect(text_match_1).to appear_before text_match_2`

Also, the `within_table` in the second example must be splitted in order to get
green specs. This is seems to be caused by rack_test not being able to detect 
updates inside the selected HTML block after a page reload.
    
Re-selecting the HTML block with a second `within_table` does the job.

**Checklist:**

### DIFF
--- a/backend/spec/features/admin/users_spec.rb
+++ b/backend/spec/features/admin/users_spec.rb
@@ -59,6 +59,7 @@ describe 'Users', type: :feature do
 
     it "can sort asc" do
       within_table(table_id) do
+        expect(page).to have_selector '.sort_link.asc'
         expect(page).to have_text text_match_1
         expect(page).to have_text text_match_2
         expect(text_match_1).to appear_before text_match_2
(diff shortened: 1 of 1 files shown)


kennyadsl:
@spaghetticode maybe we should change the first commit message? Or are we still uncertain it will fix?

spaghetticode:
@kennyadsl I'm still kind of uncertain. 

I mean, there was no real reason in the first place for these specs to be broken, in fact most of the time they were green anyway... apart from the real issue I fixed with the second commit (which anyway was not relevant for the  original `appear_before` failing check, as `appear_before` uses the unscoped `page.body`) I could not find something deterministic in the failures.

I think only the future can make us certain about this.

kennyadsl:
@solidusio/core-team can we have another review here, please?

Issue #3086, opened 2019-02-07
special characters should not be allowed in taxon permalinks


loicginoux:
**Solidus Version**:
2.7.1

**To Reproduce**
- via backend: create a taxon
- edit the "new-node"
- add some special character in permalink field, like `spécïal & character`
- save
**Current behavior**
- it saves it as `spécïal & character`
- when using this permalink in front end urls, it causes some errors because url does not handle correctly this special characters.

**Expected behavior**
- it should save it as `special-character`

**Additional context**
this is in `Taxon#set_permalink`, `permalink_tail` should be parameterize.
we should do 
```
permalink_tail.parameterize
```
 or do 
```
Spree::Config.taxon_url_parametizer_class.parameterize(permalink_tail)
```

when updating taxon, we never pass by this line 
```
Spree::Config.taxon_url_parametizer_class.parameterize(name)
```
because permalink is never nil.

If someone can confirm the bug, I can submit a PR for it

PR #3090, opened 2019-02-08
Parameterize taxon's permalink also on update

loicginoux:
**Description**
fix bug #3086 
the fix allows to use the ```Spree::Config.taxon_url_parametizer_class``` on taxon permalink update, and not only on the name. We insure permalink is always parameterized on save.

### DIFF
--- a/core/app/models/spree/taxon.rb
+++ b/core/app/models/spree/taxon.rb
@@ -59,9 +59,8 @@ module Spree
     # Sets this taxons permalink to a valid url encoded string based on its
     # name and its parents permalink (if present.)
     def set_permalink
-      permalink_tail = permalink.split('/').last if permalink.present?
-      permalink_tail ||= Spree::Config.taxon_url_parametizer_class.parameterize(name)
-      self.permalink_part = permalink_tail
+      permalink_tail = permalink.present? ? permalink.split('/').last : name
+      self.permalink_part = Spree::Config.taxon_url_parametizer_class.parameterize(permalink_tail)
     end
 
     # Update the permalink for this taxon and all children (if necessary)
--- a/core/spec/models/spree/taxon_spec.rb
+++ b/core/spec/models/spree/taxon_spec.rb
@@ -26,6 +26,14 @@ RSpec.describe Spree::Taxon, type: :model do
       expect(taxon.permalink).to eql "ruby-on-rails"
     end
 
+    context "updating a taxon permalink" do
+      it 'parameterizes permalink correctly' do
+        taxon.save!
+        taxon.update_attributes(permalink: 'spécial&charactèrs')
+        expect(taxon.permalink).to eql "special-characters"
+      end
+    end
+
     context "with parent taxon" do
       let(:parent) { FactoryBot.build(:taxon, permalink: "brands") }
       before       { allow(taxon).to receive_messages parent: parent }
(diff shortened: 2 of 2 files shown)


core/app/models/spree/taxon.rb
```diff
@@ -59,9 +59,8 @@ def seo_title
     # Sets this taxons permalink to a valid url encoded string based on its
     # name and its parents permalink (if present.)
     def set_permalink
-      permalink_tail = permalink.split('/').last if permalink.present?
-      permalink_tail ||= Spree::Config.taxon_url_parametizer_class.parameterize(name)
-      self.permalink_part = permalink_tail
+      permalink_tail = permalink.present? ? permalink.split('/').last : name
```
kennyadsl: Do you think we can use [`permalink_part_changed?`](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html) here to change the permalink_part only when it is changed?

core/spec/models/spree/taxon_spec.rb
```diff
@@ -26,6 +26,14 @@
       expect(taxon.permalink).to eql "ruby-on-rails"
     end
 
+    context "updating a taxon permalink" do
+      it 'parameterize permalink correctly' do
+        taxon.save!
+        taxon.update_attributes(permalink: 'spécial&charactèrs')
+        expect(taxon.permalink).to eql "special-characters"
+      end
+    end
```
kennyadsl: This change also has effects on permalinks of children. Can we test that it works with children as well?

loicginoux: ok, I'll add this test.

core/app/models/spree/taxon.rb
loicginoux: I'll change that, it's better yes.

loicginoux: Actually, we cannot really use that condition because this method is called also for all children via:
```    
after_update :update_child_permalinks, if: :saved_change_to_permalink?
...

def update_permalinks
  set_permalink

  # This will trigger update_child_permalinks if permalink has changed
  save!
end
````
it breaks current test, I will leave it this way.

File: core/spec/models/spree/taxon_spec.rb (hunk 2 above)
Comment (loicginoux) [reply core/spec/models/spree/taxon_spec.rb]: Test added

File: core/spec/models/spree/taxon_spec.rb [hunk 3]
```diff
@@ -26,6 +26,14 @@
       expect(taxon.permalink).to eql "ruby-on-rails"
     end
 
+    context "updating a taxon permalink" do
+      it 'parameterize permalink correctly' do
```
kennyadsl: it's `parameterizes` here, right?

core/spec/models/spree/taxon_spec.rb
kennyadsl: I think the test you added is testing that the permalink is correctly saved when it has a parent taxon. I mean to test when the taxon with special chars has children instead (like `spécial&charactèrs/child` for example) so we can see the parameterization reflected on those taxons as well. Anyway the test you added can stay, better to have that as well. 🙂

core/app/models/spree/taxon.rb
kennyadsl: Makes sense, can't see any clean solution here. I agree it's better to keep it as is.

core/spec/models/spree/taxon_spec.rb
loicginoux: yes updated

core/spec/models/spree/taxon_spec.rb
loicginoux: I added 2 more tests

loicginoux:
ok, all in one commit now. thx for the review !

loicginoux:
Ok I changed commit description and removed one `correct grammar on test description`

kennyadsl:
Thanks @loicginoux!

PR #3093, opened 2019-02-11
Add support for multiple tables of contents in the Solidus Guides

kennyadsl:
This PR is a reopening of the great @benjaminwil work on #2770:

> Right now, the Solidus Guides include one guide: the Solidus Developers Guide.
> 
> In the future, we intend to publish other guides with separate tables of contents. For example, a guide for end users (which is partially written and available at /guides/source/users/) should be displayed separately from the Developers Guide, as the content is written for a non-developer audience.
>
> These commits prepare the Solidus Guides to handle those additional TOCs.

This PR adds some commit to actually start displaying Users Guides in our guides website, making its link accessible via a new navigation in the header.

**Mobile**

schermata 2019-02-11 alle 14 16 07

**Desktop**
schermata 2019-02-11 alle 14 16 26

**NOTE**
I know it does not look perfect on a UI/UX point of view. We'll work on that later if others think this is an acceptable solution for now.

**Checklist:**


kennyadsl:
@benjaminwil if you want to take a look it would be awesome, thanks!

kennyadsl:
Thanks for the review @tvdeyen, but I can't find links without `cursor: pointer` in both navigation and sidebar.

tvdeyen:
@kennyadsl Chrome I assume? In Firefox the links have no pointer. I guess they removed the default pointer for links and leave it to the site designers to use pointer cursors. Anyway, not important for this PR something we should think about later.

kennyadsl:
@tvdeyen yes, Chrome. I agree we can tackle this into a separate PR.

PR #3094, opened 2019-02-11
Fixing inventory unit finalizer

seand7565:
**Description**
In an order with more than one inventory_unit, the total of inventory_units in the shipment would be subtracted from the stock of each item. For example, if a customer ordered 5 of "Product A" and 3 of "Product B", both products would have their stock decreased by 8, because the code is summing the inventory_units in general, not the units specific to the line_item.


seand7565:
Test added, all passing!

DanielePalombo:
Thank @seand7565 to have fixed this bug. LGTM !

PR #3096, opened 2019-02-13
Evented interactors

spaghetticode:
Basic interactors can be written using the `interactor` gem. 

Also, augmented interactors that automatically trigger events are added, see `Spree::EventedInteractor`

Events proposed in [this PR](https://github.com/solidusio/solidus/pull/3081) are integrated with the EventedIteractor for simpler use.


spaghetticode:
Sorry for the noise, opened by mistake 😞

PR #3099, opened 2019-02-14
No "Add product" at orders shipments page

### LINKED ISSUE #3085: Backend: Add product via shipment tab buggy
**Solidus Version**:
2.7.1

**To Reproduce**
- via backend, "new order"
- skip "cart" tab
- add user to order
- go to shipment tab and add product via "add product" block
- change shipping method of the shipment newly created
- create a check payment
- go back to shipments tab

**Current behavior**
The shipment loose its custom shipping method and instead the default shipping method is applied. 

**Expected behavior**
the shipment should not change. When doing the same step bug adding product via "cart" tab, the shipment is not changed when creating a check payment.

**Additional context**
What's happening is that during this `create_proposed_shipments`
```
before_transition to: :delivery, do: :create_proposed_shipments
```

it destroys and rebuilds shipments for the order. In the case "add product", the state is still in "cart" and when adding payment, we rebuild shipments after having updated its shipping method and thus we loose changes. When doing via "cart" tab, we trigger a state change, we pass by delivery step before being able to do any changes to shipments, so shipments are recreated before any custom update to shipments.

I see 2 solutions:
- we update state of the order when adding product via "add product", it would make the state being "delivery" before applying any changes to shipments.
- I would advise to remove completely this block "add products", and we would be able to add products only via the cart tab. We loose the feature to being able to select the stock location but I don't see the point of it if you already have a stock location coordinator... It also means a UI more simple and less code to maintain :)

What do you think ? I don't mind fixing this in a PR. Let me know the solution you'd prefer to see.

tvdeyen:
**Description**

Removes the "Add Product" section from Admin "Orders / Shipment" Tab. The current feature is misleading and buggy. We must not allow to add products into an order with shipments. Using the "Cart" tab instead.

Closes #3085 

**Checklist:**
- [x] I have added a CHANGELOG entry for this change


CHANGELOG.md
```diff
@@ -2,10 +2,14 @@
 
 ### Major Changes
 
+- Removes the "Add product" section from admin orders shipments tab
+
 ### Core
 
 ### Backend
 
+- Removes the "Add product" section from admin orders shipments tab
```
kennyadsl: What do you think about keeping the format of the rest of the CHANGELOG? I mean, adding a reference to this same PR since the number could be useful to easily find the context of this change.

tvdeyen: Actually I thought about not adding this at all, as we generate the changelog from PRs before we make a release. But the PR description asked me to add changelog entries. We should remove the changelog note from the PR description IMO. We can't expect from contributors to maintain a changelog in a PR (The don't know the PR # and the release this is being merged in anyway)

kennyadsl: You are right, going to remove that line now

spaghetticode:
@tvdeyen I think this PR is a significant improvement for Solidus backend, I would love to see this merged soon.

I see at the moment the PR needs rebasing and maybe some little TLC for CI? If you need some help for pushing this forward I'll be glad to help.

tvdeyen:
Unfortunately I won’t be able to finish this PR soon. 
Please take over. Thanks.

tvdeyen:
@spaghetticode will tkae this over and open a new PR

Issue #3103, opened 2019-02-15
[DEMO] non-heroku demo?


rubyFeedback:
Is there some way to get a non-heroku demo?

Reason I ask is that heroku requires sign-in when I clicked the Demo link on the main readme here ("You'll need a Heroku account to deploy the app") and I really don't want to sign-in just to 
try out solidus.

Anyway, if this is too much work or unwanted, please feel free to close the issue request - ultimately I could just go ahead and sign up at heroku, but I'll rather skip that altogether (I am a lazy person, I know).

kennyadsl:
If you can clone the repo locally, you could just run

```
bundle install
bundle exec rake sandbox
cd sandbox
bundle exec rails s
```

to have a working Solidus demo store. Alternatively, there's no way at the moment.

JuanCrg90:
@kennyadsl, with the docker demo addition, is this issue still valid?

kennyadsl:
@JuanCrg90 yes, you are right. Thanks!

PR #3106, opened 2019-02-16
Enable Docker for demoing purposes

kinduff:
**Description**
Enables Solidus to be ran using Docker for demoing purposes. This provides an easy way for newcomers to test the storefront out without going through the install process and avoiding a Heroku deployment.

Ref #3103

**Checklist:**

**How to test it out:**
As the README states, just run the following command to download the image and run it at [http://localhost:3000](http://localhost:3000).

```
docker run --rm -it -p 3000:3000 kinduff/solidus-demo:latest
```

The admin can be found at [http://localhost:3000/admin/](http://localhost:3000/admin/), credentials are `admin@example.com` and `test123`.

**Notes & Questions:**
This PR is not ready to be merged since I pushed the Docker image to my own Dockerhub account.

Wanted to submit and ask before adding more work into this if it would be worth to create some sort of Rake task to generate an image for each of the existing versions or just rely on master.

Also to ask if anyone has insights about how to automate this with Circle CI and have an up to date image everytime there's a change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,18 @@ Try out Solidus with one-click on Heroku:
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/solidusio/solidus)
 
+Alternatively, you can use Docker to run a demo on your local machine. Run the
+following command to download the image and run it at
+[http://localhost:3000](http://localhost:3000).
+
+```
+docker run --rm -it -p 3000:3000 solidusio/solidus-demo:latest
+```
+
+The admin interface can be accessed at
+[http://localhost:3000/admin/](http://localhost:3000/admin/), the default
+credentials are `admin@example.com` and `test123`.
+
 ## Getting started
 
 Begin by making sure you have
--- a/lib/demo/Dockerfile
+++ b/lib/demo/Dockerfile
@@ -1,0 +1,21 @@
+#
+# This image is intended to be used to test and demo Solidus
+# it is not intended for production purposes 
+#
+FROM ruby:2.5.1
+
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+
+RUN apt-get update -qq && apt-get install -y build-essential libpq-dev nodejs
+
+RUN mkdir /solidus
+
+WORKDIR /solidus
+
+ADD . /solidus
+
+RUN bundle install
+
+RUN bundle exec rake sandbox
+
+CMD ["sh", "./lib/demo/docker-entrypoint.sh"]
--- a/lib/demo/README.md
+++ b/lib/demo/README.md
@@ -1,0 +1,27 @@
+# Solidus Docker Demo
+The Dockerfile in this folder is intended for demoing purposes, meaning that a
+user can run an installation of Solidus using a Docker container.
+
+## How to build the image
+Make sure Docker is installed in your local and run the following command:
+
+```shell
+docker build -t solidusio/solidus-demo:latest -f lib/demo/Dockerfile .
+```
+
+## How to run the image
+You can either run the image you built locally or run the official image pushed
+in Dockerhub.
+
+```shell
+docker run --rm -it -p 3000:3000 solidusio/solidus-demo:latest
+```
+
+## How to push the image
+If you want to push the image you can use the following command, just note that
+this specific command will push the image to the official Solidus Dockerhub
+account.
+
+```shell
+docker push solidusio/solidus-demo:latest
+```
--- a/lib/demo/docker-entrypoint.sh
+++ b/lib/demo/docker-entrypoint.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+export RAILS_ENV=development
+
+cd sandbox && \
+bundle exec rails s -p 3000 -b '0.0.0.0'


kennyadsl:
@kinduff thanks for the contribution. My first thought has been that I don't want to maintain this Docker Demo feature, mostly because I don't have a lot of experience with Docker and there's no way to know if it's still working for all solidus features (eg. file upload or other things which could depend on the system). On the other end, the same could happen (and happens often) with Heroku. For sure we'll talk about this on the next Core Team meeting.

dhonig:
+1 for docker.
It opens up lots of cloud platforms.  Maintenance shouldn’t be too hard. Ill be glad to help

kennyadsl:
> Also to ask if anyone has insights about how to automate this with Circle CI and have an up to date image everytime there's a change.

@kinduff maybe a GitHub Action?

aitbw:
Hi @kinduff, thanks a lot for PR. I have a few improvements / questions on this:

* I think the `Dockerfile` is still missing some essential packages, such as `postgresql-client` and `imagemagick` —I think we should add them in order to provide a seamless Docker experience

* I'm no Docker expert, but I think we should be using `docker compose` to handle database responsabilities. What do you think?

* Does this work without any extra configuration on Solidus?

* The last time I worked with Docker, when we ran the commands to build the container, Docker did not respect the files' ownership —this means we had to use `sudo` for every command every time we were not using Docker, which was very annoying. Is this the case here? If so, can we think of a solution?

* Does the full spec suite runs green without any extra configuration? I remember the `--no-sandbox` flag had to be included to the Headless Chrome setup under Docker containers.

Let me know what you think! :hugs:

kinduff:
Thanks for your input @aitbw, I'm quoting and responding.

> I think the `Dockerfile` is still missing some essential packages, such as `postgresql-client` and `imagemagick` —I think we should add them in order to provide a seamless Docker experience

`imagemagick` is already installed. I'm not sure about `postgresql-client` since by default sandbox environment is using `sqlite`, so for a demo purposes it works well.

> I'm no Docker expert, but I think we should be using `docker compose` to handle database responsabilities. What do you think?

I'm all in with docker-compose files, but since this is a demo just to run and see Solidus, I think we should treat it as is, since it's using `sqlite`. If we want to provide the developers a Docker guidance, as an official way to setup the project - for dev, staging and production, I can help with that. 

> Does this work without any extra configuration on Solidus?

Works with Solidus out of the box, since it's just a demo meant to be ran in local and not a development platform.

> The last time I worked with Docker, when we ran the commands to build the container, Docker did not respect the files' ownership —this means we had to use `sudo` for every command every time we were not using Docker, which was very annoying. Is this the case here? If so, can we think of a solution?

It's not the case since it's using a `sandbox` user to lock it out, since the owner of the app is this user, you don't need sudo. If you need to `/bin/bash` into containers to modify things, then the setup for Docker and the application is not set correctly.

> Does the full spec suite runs green without any extra configuration? I remember the `--no-sandbox` flag had to be included to the Headless Chrome setup under Docker containers.
I'm working on adjustments to enable the tests suits on this image related to `chromedriver`, it's necessary but it's shouldn't be a blocker since it doesn't break the backend/frontend demo as this image is intended to provide.

aitbw:
Hey @kinduff, thanks for the answers! Just one question:

> `imagemagick` is already installed. I'm not sure about `postgresql-client` since by default sandbox environment is using `sqlite`, so for a demo purposes it works well.

Are you sure about this? Because on the last project I worked on, we had to install `imagemagick` when building the container.

kinduff:
> Are you sure about this? Because on the last project I worked on, we had to install `imagemagick` when building the container.

Yes @aitbw, give it a try:

```
docker run --rm -it kinduff/solidus-demo:latest convert --version
```

Should output:

```
Version: ImageMagick 6.9.7-4 Q16 x86_64 20170114 http://www.imagemagick.org
Copyright: © 1999-2017 ImageMagick Studio LLC
License: http://www.imagemagick.org/script/license.php
Features: Cipher DPC Modules OpenMP 
Delegates (built-in): bzlib djvu fftw fontconfig freetype jbig jng jp2 jpeg lcms lqr ltdl lzma openexr pangocairo png tiff wmf x xml zlib
```

I also updated the README to point to the correct image, sorry about that. You should be able to run it too as a demo server.

kennyadsl:
In the last Core Team meeting, we had a chat about this and we are ok with this PR if the image is hosted on a Solidus Docker Hub account. There's an existing `solidusio` org on Docker Hub, trying to understand who created that.

kinduff:
@kennyadsl That's awesome. Let me know how I can help or if I can be added as a collaborator :+1:

dhonig:
@kennyadsl would be great to get @kinduff on the core team to improve
communication and increase our contributions!

README.md
```diff
@@ -66,6 +66,14 @@ Try out Solidus with one-click on Heroku:
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/solidusio/solidus)
 
+Alternatively, you can use Docker to run a demo in your local machine. Run the following command to download the image and run it at [http://localhost:3000](http://localhost:3000).
+
+```
+docker run --rm -it -p 3000:3000 kinduff/solidus-demo:latest
+```
+
+The admin can be found at [http://localhost:3000/admin/](http://localhost:3000/admin/), credentials are `admin@example.com` and `test123`.
```
jacobherrington: Perhaps this could say ```The admin interface can be accessed at [http://localhost:3000/admin/](http://localhost:3000/admin/), the default credentials are `admin@example.com` and `test123`.```, what do you think?

kinduff:
@jacobherrington I :heart: your suggestions. I was thinking to use a namespace like `Dockerfile.demo` but your `lib` suggestion sounds way better. I'll address these changes in a bit, and also the changes on the README file.

kinduff:
@jacobherrington Just pushed the latest updates by your request. How do you think we could push this demo image to the official Solidus Dockerhub?

docker-entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
kennyadsl: what about also moving this file into the Demo folder?

kinduff:
Looks like everything is set up, we're just missing to have the docker image in the official Dockerhub account.

kennyadsl:
@kinduff great, thanks. I've opened a support ticket on DockerHub to understand how we could get the `solidusio` or `solidus` organization ownership, which already exist but no one seems to have created. 

In the meantime, maybe you could rebase the PR since I think one single, well-documented commit in enough for this change, thanks!

kennyadsl:
We did it: https://cloud.docker.com/u/solidusio/repository/docker/solidusio/solidus-demo 🎉

tvdeyen:
🎉 

The correct link https://hub.docker.com/r/solidusio/solidus-demo

dhonig:
Nice work all!

If there is interest in a development setup I would be happy to donate that.

PR #3115, opened 2019-02-20
Set mininum line item quantity in admin cart

mamhoff:
**Description**
It does not make much sense to add zero of anything to an order. This
commit changes the minimum quantity for a line item in the admin cart to
one.

**Checklist:**

I don't think a test is necessary for this change.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/templates/orders/line_item.hbs
+++ b/backend/app/assets/javascripts/spree/backend/templates/orders/line_item.hbs
@@ -20,7 +20,7 @@
 {{#if editing}}
   <td class="line-item-qty-edit">
     <form>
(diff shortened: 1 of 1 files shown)


kennyadsl:
I'm giving a live try to this PR. I think it's ok as is since it does not allow you to set 0 as value by using broswer's number input controls. Anyway, a user is still able to fill that input with `0` and submit. I think this happens since the submit of that form is never really called. Maybe it's just matter of triggering the validation programmatically in [the backbone view](https://github.com/solidusio/solidus/blob/e55f3279c2046dc82f42b8cb7703006fa42987a2/backend/app/assets/javascripts/spree/backend/views/cart/line_item_row.js#L43)?

PR #3116, opened 2019-02-20
Add a stock locations filter configurable class

kennyadsl:
**Description**

This class is helpful to filter stock locations that need to be taken into account by the `SimpleCoordinator` while generating shipments.

**Checklist:**


core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -26,7 +26,10 @@ def initialize(order, inventory_units = nil)
         @order = order
         @inventory_units = inventory_units || InventoryUnitBuilder.new(order).units
         @splitters = Spree::Config.environment.stock_splitters
-        @stock_locations = Spree::Config.stock.location_sorter_class.new(Spree::StockLocation.active).sort
+
+        filtered_stock_locations = Spree::Config.stock.location_filter_class.new(Spree::StockLocation.all, @order).filter
+        sorterd_stock_locations = Spree::Config.stock.location_sorter_class.new(filtered_stock_locations).sort
```
aitbw: `sorterd` -> `sorted`

core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -26,7 +26,10 @@ def initialize(order, inventory_units = nil)
...
+        @stock_locations = sorterd_stock_locations
```
aitbw: `sorterd` -> `sorted`

guides/source/developers/shipments/stock-locations-filter.html.md
```diff
@@ -0,0 +1,54 @@
...
+stock locations.
+
+[split-shipments]: split-shipments.html
+
+## Pre-configured filters
+
+Solidus ships with just one filter:
+
+- [Active](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/stock/location_filter/active.rb),
+    which returns only stock locations active.
```
aitbw: Should be worded as `active stock locations.`

guides/source/developers/shipments/stock-locations-filter.html.md
```diff
@@ -0,0 +1,54 @@
...
+# /config/application.rb
+module MyStore
+  class Application < Rails::Application
+    # ...
+
+    initializer 'spree.register.stock_location_filters' do |app|
+      app.config.spree.stock.location_filter_class = 'Spree::Stock::LocationFilter::SameOrderCountry'
+    end
+  end
+end
```
tvdeyen: Shouldn't we advice to use the `config/initializers/spree` file instead?

kennyadsl: This is what is being suggested in other parts of the guides, eg: https://guides.solidus.io/developers/shipments/custom-shipping-calculators.html

We use that initializer in our apps so I'm totally for that. For consistency, I'd like to change this everywhere in guides, what do you think?

tvdeyen: We should make this consistence over the guides

kennyadsl: Done, thanks for the suggestion!

PR #3118, opened 2019-02-22
When a controller action fails to be authorized, redirect back or default to /unauthorized

genarorg:
**Description**
This change allows for a user to be redirected back whenever a request is not authorized. The current behavior redirects the user to `/unauthorized` all the time, but this is not ideal when browsing the admin, since you need to manually go back to where you came from.

**Changes:**
* `Spree::Core::ControllerHelpers::Auth`: In `self.unauthorized_redirect`, use `redirect_back` with a fallback location.

**Checklist:**

### DIFF
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -220,6 +220,12 @@ module Spree
     #   @return [Boolean] (default: +true+)
     preference :raise_with_invalid_currency, :boolean, default: true
 
+    # @!attribute [rw] redirect_back_on_unauthorized
+    #   Whether to try to redirect users back when they try to access
+    #   unauthorized routes, before redirect them to /unauthorized.
+    #   @return [Boolean] (default: +false+)
+    preference :redirect_back_on_unauthorized, :boolean, default: false
+
     # @!attribute [rw] require_master_price
     #   @return [Boolean] Require a price on the master variant of a product (default: +true+)
     preference :require_master_price, :boolean, default: true
--- a/core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
+++ b/core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
@@ -34,6 +34,11 @@ Spree.config do |config|
   # See https://github.com/solidusio/solidus/pull/3456 for more info.
   config.raise_with_invalid_currency = false
 
+  # Set this configuration to false to always redirect the user to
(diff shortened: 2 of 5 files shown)


kennyadsl:
Thanks! Are you using `solidus_auth_devise` as well? Because I think that Proc [is overridden in that extension](https://github.com/solidusio/solidus_auth_devise/blob/91687769600cc9292e198684be8d15b6796f07e2/lib/spree/auth/engine.rb#L58-L67) as well. Maybe something we should take into account?

genarorg:
ah! thanks for pointing that out. I can make the change in `solidus_auth_devise`, so that the behavior would be consistent to that of core. How does that sound?

kennyadsl:
Sound good to me, please note that for the Admin controller there's another point where this is set, other than what I've previously linked: https://github.com/solidusio/solidus_auth_devise/blob/335991f79557bc36ea08943b9e1fbb151154d020/lib/spree/auth/engine.rb#L37

kennyadsl:
If we create a preference for letting people opt-in for this new behavior, we can deprecate the old one and use this preference into https://github.com/solidusio/solidus_auth_devise/pull/192 as well.

core/lib/spree/app_configuration.rb
```diff
@@ -213,6 +213,12 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Boolean] (default: +true+)
     preference :raise_with_invalid_currency, :boolean, default: true
 
+    # @!attribute [rw] redirect_back_on_unauthorized
+    #   Whether to try to redirect users back when they try to access
+    #   unauthorized routes, before redirect them to /unauthorized.
+    #   @return [Boolean] (default: +false+)
+    preference :redirect_back_on_unauthorized, :boolena, default: false
```
spaghetticode, line 220: This should be `:boolean`

kennyadsl: Wow, and it's working?

spaghetticode: 🤷‍♂️

spaghetticode: It works because that information is used for [typecasting](https://github.com/solidusio/solidus/blob/16ed773a6554782e33476e2ffc41b70320ddb46e/core/lib/spree/preferences/preferable.rb#L144). In that case, no typecasting would occur as the type would not be recognized.

aldesantis:
@kennyadsl this is good to merge I think?

PR #3124, opened 2019-02-27
Remove unnecessary decimal conversion

brchristian:
**Description**
This removes an unnecessary conversion of the count of completed orders (which is always necessarily an integer!) into a decimal number.

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/user_reporting.rb
+++ b/core/app/models/concerns/spree/user_reporting.rb
@@ -10,7 +10,7 @@ module Spree
     end
 
     def order_count
-      BigDecimal(spree_orders.complete.count)
+      spree_orders.complete.count
     end
 
     def average_order_value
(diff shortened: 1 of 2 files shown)


brchristian:
@kennyadsl That’s a good idea! I’m surprised that current didn’t fail, actually.

What if we change from something like this:

```ruby
expect(subject.order_count).to eq BigDecimal(order_count)
```

to something like this:

```ruby
expect(subject.order_count).to eq subject.orders.complete.count
```

If that seems reasonable to you, I can add a commit to this PR.

kennyadsl:
@brchristian I'm more for adding the exacxt number that we expect so just leaving:

```ruby
expect(subject.order_count).to eq order_count
```

brchristian:
@kennyadsl sounds good, I have added the spec to this PR!

Issue #3098, opened 2019-02-13
Cannot change shipping address to country without states


brchristian:
Users whose shipping address defaults to a country with states (e.g., USA) cannot create an order that ships to a country without states (e.g., Anguilla). They will be stopped from getting past the Address phase of the checkout flow with an error saying `Shipping address state does not match the country`

**Solidus Version**:

2.8.1

**To Reproduce**

1. On a clean install of the latest version of Solidus.
2. Log in with default credentials (admin@example.com)
3. Check out with an order with billing address & shipping address in a country that has states (e.g., USA)
4. Create second order, which will default billing and shipping addresses to the last-used address
5. Change shipping address to a country without states (e.g., Anguilla)
6. Attempt to proceed
7. Receive error: `1 error prohibited this record from being saved: There were problems with the following fields: Shipping address state does not match the country`

**Current behavior**

User cannot check out.

**Expected behavior**

User can check out.

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Safari
 - Version: 12.0.3

VzqzAc:
By a bit of investigation, I see that the problem is that https://github.com/solidusio/solidus/blob/master/frontend/app/controllers/spree/checkout_controller.rb#L54 which invokes https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_update_attributes.rb#L18-L25 does not perform a ship address update, I thought on a way to solve it, by following kind of what is being made with the payments in the same class, something like
```ruby
    def apply
      order.validate_payments_attributes(@payments_attributes)

      assign_order_attributes
      assign_payments_attributes
      assign_ship_address_attributes unless order.use_billing.in?([true, 'true', '1']) # Or moving the `Spree::Order#use_billing?` method to be public and call it

      order.save
    end 
    # ...
    def assign_ship_address_attributes
        # update order.ship_address attributes
    end 
```

But I'm not sure if this is a path we'd like to follow, that was just my first thought to fix this

kennyadsl:
Thanks @brchristian for reporting and @VzqzAc for these initial thoughts. I think this is also related to https://github.com/solidusio/solidus/pull/2417, which I started fixing on the view layer (by resetting the state to `nil`in the form). We have fixed this already the same way in backend but I think we should do something on the model layer. I know @cedum started working on this already. 

[...]

PR #3129, opened 2019-03-01
Refactor address state validation

cedum:
**Description**

Deprecates and refactors state validation by extracting out the normalization and validation logic into a separate service-like object, which might be also user-provided, for example:
```
Spree::Address.state_validator_class = UserProvided::Address::StateValidator
```

The following private methods have been deprecated from `Spree::Address` model:
- `state_validate`
- `validate_state_matches_country`

This PR introduces the following configuration flag `Spree::Config.use_legacy_address_state_validator` that enables to switch from the old state validation behavior to the new in a smooth way.
For newly generated applications, this flag is set to `false`, which means it will use the new state validator class.
For existing applications upgrading to Solidus v2.11 it is set to `true`, hence keeping the same behavior as in the previous Solidus versions.

**Checklist:**
- [ ] Write a recap in this PR description with the normalization and validation rules

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -218,6 +218,12 @@ module Spree
     end
 
     def state_validate
+      Spree::Deprecation.warn \
+        "#{self.class}#state_validate private method has been deprecated" \
+        " and will be removed in Solidus v3." \
+        " Check https://github.com/solidusio/solidus/pull/3129 for more details.",
+        caller
+
       # Skip state validation without country (also required)
       # or when disabled by preference
       return if country.blank? || !Spree::Config[:address_requires_state]
@@ -253,6 +259,12 @@ module Spree
     end
 
     def validate_state_matches_country
+      Spree::Deprecation.warn \
+        "#{self.class}#validate_state_matches_country private method has been deprecated" \
+        " and will be removed in Solidus v3." \
+        " Check https://github.com/solidusio/solidus/pull/3129 for more details.",
+        caller
+
       return unless country
 
       self.state = nil if country.states.empty?
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -8,6 +8,9 @@ module Spree
   class Address < Spree::Base
     extend ActiveModel::ForbiddenAttributesProtection
 
+    mattr_accessor :state_validator_class
+    self.state_validator_class = Spree::Address::StateValidator
+
     belongs_to :country, class_name: "Spree::Country", optional: true
     belongs_to :state, class_name: "Spree::State", optional: true
 
@@ -16,8 +19,20 @@ module Spree
     validates :phone, presence: true, if: :require_phone?
 
     validate :validate_name
-    validate :state_validate
-    validate :validate_state_matches_country
+
+    validate do
+      if Spree::Config.use_legacy_address_state_validator
+        begin
+          @silence_state_deprecations = true
+          state_validate
+          validate_state_matches_country
+        ensure
+          @silence_state_deprecations = false
+        end
+      else
+        self.class.state_validator_class.new(self).perform
+      end
+    end
 
     alias_attribute :first_name, :firstname
     alias_attribute :last_name, :lastname
@@ -218,11 +233,13 @@ module Spree
     end
 
     def state_validate
-      Spree::Deprecation.warn \
-        "#{self.class}#state_validate private method has been deprecated" \
-        " and will be removed in Solidus v3." \
-        " Check https://github.com/solidusio/solidus/pull/3129 for more details.",
-        caller
+      unless @silence_state_deprecations
+        Spree::Deprecation.warn \
+          "#{self.class}#state_validate private method has been deprecated" \
+          " and will be removed in Solidus v3." \
+          " Check https://github.com/solidusio/solidus/pull/3129 for more details.",
+          caller
+      end
 
       # Skip state validation without country (also required)
       # or when disabled by preference
@@ -259,11 +276,13 @@ module Spree
     end
 
     def validate_state_matches_country
-      Spree::Deprecation.warn \
-        "#{self.class}#validate_state_matches_country private method has been deprecated" \
-        " and will be removed in Solidus v3." \
-        " Check https://github.com/solidusio/solidus/pull/3129 for more details.",
-        caller
+      unless @silence_state_deprecations
+        Spree::Deprecation.warn \
+          "#{self.class}#validate_state_matches_country private method has been deprecated" \
+          " and will be removed in Solidus v3." \
+          " Check https://github.com/solidusio/solidus/pull/3129 for more details.",
+          caller
+      end
 
       return unless country
 
--- a/core/app/models/spree/address/state_validator.rb
+++ b/core/app/models/spree/address/state_validator.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+module Spree
+  class Address::StateValidator
+    attr_reader :address
+    delegate :state, :state_name, :country, to: :address
+
+    def initialize(address)
+      @address = address
+    end
+
+    def perform
+      return unless state_required?
+
+      if country.present?
+        normalize_state if state.present?
+        normalize_state_name if state_name.present?
+      end
+
+      validate_not_blank
+      validate_matches_country
+    end
+
+    private
+
+    def normalize_state
+      # discard the 'state' attribute when having a country with no states
+      address.state = nil if country.states.blank?
+    end
+
+    def normalize_state_name
+      # discard the 'state_name' when having a valid 'state' and country combo
+      if state.present? && state.country == country
+        address.state_name = nil
+        return
+      end
+
+      # set the state from the state name if the country contains one with that name
+      states_from_name = country.states.with_name_or_abbr(state_name)
+      if states_from_name.size == 1
+        address.state = states_from_name.first
+        address.state_name = nil
+      end
+    end
+
+    def validate_not_blank
+      if state.blank? && state_name.blank?
+        address.errors.add(:state, :blank)
+      end
+    end
+
+    def validate_matches_country
+      if state.present? && state.country != country
+        address.errors.add(:state, :does_not_match_country)
+      end
+    end
+
+    # Don't require a state if disabled at config level or
+    # the associated country doesn't require states
+    def state_required?
+      Spree::Config.address_requires_state && country_requires_states?
+    end
+
+    def country_requires_states?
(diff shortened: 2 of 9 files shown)


cedum:
Moved the PR back to WIP 🚧. It requires some more work. I'll try to make it less break. Besides this, by digging deeper into it I see I misinterpreted some state behaviours and also introduced a few bugs 😅

core/app/models/spree/address.rb
```diff
@@ -8,15 +8,22 @@ module Spree
...
     belongs_to :state, class_name: "Spree::State"
 
     validates :firstname, :address1, :city, :country_id, presence: true
     validates :zipcode, presence: true, if: :require_zipcode?
     validates :phone, presence: true, if: :require_phone?
 
-    validate :state_validate
-    validate :validate_state_matches_country
+    validate do
+      self.class.state_normalizer_class.new(self).perform!
```
kennyadsl: Normalization is also happening into the validation, right? Why it is needed to perform it here as well?

core/app/models/spree/address/state_validator.rb
```diff
@@ -0,0 +1,50 @@
...
+  class Address::StateValidator
+    attr_reader :address
+    delegate :state, :state_name, :country, to: :address
+
+    def initialize(address)
+      @address = address
+
+      # normalize state and state_name attributes before
+      # performing the validation
+      Spree::Address.state_normalizer_class.new(address).perform!
```
kennyadsl: What about passing this class to the `initialize` of this class?

```ruby
def initialize(address, address_normalizer_class = Spree::Address.state_normalizer_class)

# and in the caller
self.class.state_validator_class.new(self, self.class.state_normalizer_class).perform!
```

This way it will still be possible to change the class as an address level preference but we can also have more flexibility if needed for some customization or testing needs.

core/app/models/spree/address/state_normalizer.rb
```diff
@@ -0,0 +1,45 @@
...
+module Spree
+  class Address::StateNormalizer
+    attr_reader :address
+    delegate :state, :state_name, :country, to: :address
+
+    def initialize(address)
+      @address = address
+    end
+
+    def perform!
```
kennyadsl: Personal preference, but I'd use `def call` here. Did you have a specific reason to use `perform!` instead?

core/app/models/spree/address/state_validator.rb
cedum: This wasn't intended to be here. I have moved it out into `validate` block in `Spree::Address`. Forgot to remove it from here :|

core/app/models/spree/address/state_normalizer.rb
cedum: I have followed the convention of other service-like objects in Solidus that are using `perform` as their main action method. I have no preferences though.
Probably it's worth defining a project-wise convention for this kind of objects

core/app/models/spree/address.rb
cedum: The original idea was to move it here from the validator service object so they can be completely decoupled. Just forgot to remove it from there (see also the comment bellow related to this one)

cedum:
Ready for a second review 🙏 

@kennyadsl I've researched a few methods to deprecate them by showing a warning message only if they have been monkey patched, but the final solution was quite hacky and not working in all cases since there're multiple ways to patch them.
Being both of them private methods, you think it's worth the effort to deprecate them before removing? I've invested some time into it but I wasn't able to find a +/- clean way to deprecate them.

cedum:
Rebased against master and resolved conflicts

aldesantis:
@cedum can you provide some info about the current state of the deprecation warning in this PR? It looks like everything should work properly if the private methods are being overridden, am I mistaken?

core/app/models/spree/address.rb
```diff
@@ -8,15 +8,31 @@ module Spree
...
 
-    validate :state_validate
-    validate :validate_state_matches_country
+    validate do
+      self.class.state_normalizer_class.new(self).perform
+      self.class.state_validator_class.new(self).perform
+
+      %w[state_validate validate_state_matches_country].each do |method|
+        if self.class.method_defined?(method)
+          Rails.logger.warn(
```
aldesantis, line 29: What if we called the custom methods instead of the new behavior in case the methods have been added by the user in a decorator? We may even do that in the `state_normalizer_class` and `state_validator_class` themselves.

kennyadsl:
I recall we decided to merge this in 3.0 but I think it wasn't a good decision since, if possible, we should make this optional and deprecate the old behavior in 2.11. This way people have time to fix this in their application before the 3.0 upgrade.

core/app/models/spree/address.rb
```diff
@@ -218,6 +228,12 @@ def validate_name
     end
 
     def state_validate
+      Spree::Deprecation.warn \
```
kennyadsl, line 231: Until the preference is not changed users will see 3 deprecation warnings. It would be awesome to find a way to only print one deprecation warning in that case, but keep the individual ones if these methods are called from somewhere else. I'm not sure it's a good idea but what about passing an arg to these deprecated methods to conditionally silence the deprecation warning?

This way we can do something like:

```
      if Spree::Config.legacy_address_state_validator
        state_validate(silence_deprecation: true)
        validate_state_matches_country(silence_deprecation: true)
      else
```

Users still see the main deprecation warning if using default code and the specific ones if they used `state_validate` or `validate_state_matches_country` in other ways. I'm not particularly concerned about changing the arity of those methods since they'll go away soon. 

What's your thoughts?

core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -119,6 +119,7 @@ class Application < ::Rails::Application
   config.raise_with_invalid_currency = false
   config.redirect_back_on_unauthorized = true
   config.run_order_validations_on_order_updater = true
+  config.legacy_address_state_validator = false
```
kennyadsl, line 122: Can we alphabetize this new preference?

core/app/models/spree/address.rb
cedum: 👍 agree!
By introducing a new parameter it will break any existing prepended patches of these methods. I'm not aware of a way to introduce a new parameter w/o being breaking. I'm looking if there's another way

kennyadsl: Maybe just setting an accessor on the class?

cedum: Another way might be to wrap the two calls in a `silence` block, this avoids polluting the Address class API with new attributes/methods. The downside is it will silence any eventual deprecations generated by these two calls.
What do you think?

```ruby
validate do
  if Spree::Config.legacy_address_state_validator
    Spree::Deprecation.silence do
      state_validate
      validate_state_matches_country
    end
  else
    self.class.state_validator_class.new(self).perform
  end
end
```

kennyadsl: It's more elegant for sure but I think (might be worth a double-check) we are not using `Spree::Deprecation.silence` anywhere in the codebase (outside specs). I'm not completely sure why to be honest and if this could cause issues that we can't anticipate at the moment. Anyway, for now this seems like the best solution let's go head with this and let's see what the other core team member think about it.

cedum: just discovered that `silence` seems to not be thread-safe 🤦 Here's a simple test case that sometimes leaves silenced class attribute set to `true`: https://gist.github.com/cedum/8315ebdd59781ee7fcc7f452a405542f
I'm refactoring this to a simple instance variable switch 😄, something like this:
```ruby
validate do
  if Spree::Config.legacy_address_state_validator
    begin
      @silence_state_deprecations = true
      state_validate
      validate_state_matches_country
    ensure
      @silence_state_deprecations = false
    end
  else
    self.class.state_validator_class.new(self).perform
  end
end
```

kennyadsl (approved):
👍

kennyadsl:
@cedum we discussed this in the core meeting and we agreed that it would be much better if we have the fix for #2417 and #3098 in a different commit, just for documentation in case someone cannot update and wants to just understand what changed to backport that in their own application. Do you think it's doable?

cedum:
> @cedum we discussed this in the core meeting and we agreed that it would be much better if we have the fix for #2417 and #3098 in a different commit, just for documentation in case someone cannot update and wants to just understand what changed to backport that in their own application. Do you think it's doable?

Yep, makes sense! I'm on it

core/app/models/spree/address.rb
```diff
@@ -253,9 +276,17 @@ def state_validate
...
-      return unless country
+      unless @silence_state_deprecations
+        Spree::Deprecation.warn \
+          "#{self.class}#validate_state_matches_country private method has been deprecated" \
+          " and will be removed in Solidus v3." \
+          " Check https://github.com/solidusio/solidus/pull/3129 for more details.",
+          caller
+      end
 
+      return unless country
```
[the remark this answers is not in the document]
cedum: ok boss

cedum:
I've extracted the fix for #3098 into a separate PR (merged into master), this way it should be easier to backport the fix on stores unable to upgrade. 
I've just rebased also this PR and I think it's ready for another review 🙏

kennyadsl (approved):
👍 Thanks @cedum !

core/lib/spree/core/engine.rb
```diff
@@ -74,6 +74,15 @@ class Engine < ::Rails::Engine
             caller
           )
         end
+
+        if Spree::Config.use_legacy_address_state_validator != false
+          Spree::Deprecation.warn(
+            'Spree::Config.use_legacy_address_state_validator set to true is ' \
+            'deprecated and will not be possibile in Solidus 3.0. Please switch this ' \
+            'value to false and check that everything works as expected.',
```
aldesantis, line 82: Can we improve this message to explain that validation will be done through an external class moving forward?

aldesantis (changes requested):
@cedum just left a comment on a deprecation message, but other than that it's good to go!

core/lib/spree/core/engine.rb
cedum: @aldesantis what do you think about this message?

> DEPRECATION WARNING: Spree::Config.use_legacy_address_state_validator set to true has been deprecated and will be removed in Solidus 3.0. The Spree::Address state validation has been extracted into a configurable external class. Switch Spree::Config.use_legacy_address_state_validator to true to start using the external validator class.

aldesantis: @cedum sounds great!

spaghetticode (approved):
@cedum thank you 👍

aldesantis (approved):

PR #3130, opened 2019-03-04
Allow orders with different shipping categories

### LINKED ISSUE #2998: Cannot order items with different shipping categories
### Steps to reproduce
In the sandbox of Solidus 2.7.1, create two shipping categories and assign them to two products.
Add the two products to the cart and go until the address  step, compile and click "save and continue".

### Expected behavior
It should go to the Delivery step

### Actual behavior
It throws the error "We are unable to calculate shipping rates for the selected items."

### System configuration
**Solidus Version**: 2.7.1 (sandbox)
It worked with v2.4, actually I'm updating from 2.4 to 2.7.

I had a look and I found that the second `shipment` doesn't have any `shipping_rate` (`shipments.second.shipping_rates` is empty).
If I add the single items to the cart, they work.

### LINKED ISSUE #3315: Unable to calculate shipping rates for the selected items
Hello. I found and issue with calculating shipping methods in v2.8.4 and v2.9.0. It seems that the shipping methods ids associated to a particular products are not loaded correctly.

**Solidus Version**:
v2.8.4 
v2.9.0
v2.9.1

**To Reproduce**

Admin Section
1. Bootstrap a solidus store with sample data.
2. Make sure all the shipping methods belong to the default store (e.g Sample Store)
3. Create a new shipping category (e.g Big Products)
4. Change or Create one shipping method associated to that new Category. (e.g UPS Ground (USD) to use only Big Products category) for the default store (e.g Sample Store)
5. Change any product (e.g Ruby on Rails Mug) to use the new shipping category (e.g Big Products)

Customer
1. Put in your basket two products, one with Default shipping category (e.g Ruby on Rails Stein) and the one with the new shipping category (e.g Ruby on Rails Mug)
2. Go through the checkout and fill the form with a valid delivery address and click 'Save and Continue' to go to Delivery Options
3. You get an error saying: 'We are unable to calculate shipping rates for the selected items.'

**Expected**

The expected result should be that the customer has two different shipping options, one for each product as each product has a different shipping category.

**Information**

Digging the code and the queries, it seems that the `estimator` line 58 https://github.com/solidusio/solidus/blob/v2.8/core/app/models/spree/stock/estimator.rb#L58 call the scope `available_to_store` with options `package.shipment.order.store`. For some reasons, the scope is not returning all the shipping associated to the store, it returns only a partial of that. As a result, not all the shipping rates are calculated.

If the shipping methods are not associated to a Store, it works, as the scope for the query in model `shipping_method` line 44 https://github.com/solidusio/solidus/blob/v2.8/core/app/models/spree/shipping_method.rb#L44 include all the shipping if no association with store is provided. Has anyone else seen these?

I found a way to patch, it seems that if I called `@order.store.shipping_method_ids` somewhere before the delivery state is change (e.g `before_delivery` method in `checkout_controller`https://github.com/solidusio/solidus/blob/v2.8/frontend/app/controllers/spree/checkout_controller.rb#L202), the query picks up all the shipping method associated to the store, not an elegant solution. Is the query being cached? Why is not picking up all the shipping methods? If I'm missing something in the config, please let me know. Any suggestions are welcome. Check the queries in the log to see the values for the shipping and you will see some of them are missing.

If anyone in the solidus team needs more details or the steps are not clear, please drop me a message here and I'll try to provide details. At the moment I have the patch that I mention but I wonder if anyone else has found it as I think it is pretty standard to have different shipping categories for different types of products. It should work OK unless I'm doing something wrong.

Regards

aitbw:
**Description**

This PR allows users to submit orders with items that belong to different shipping categories.

This issue is present from Solidus v2.6 through v2.8, as well as in `master`

Once merged:
- Closes #2998 
- Closes #3315

See also:
  * solidusio/solidus_multi_domain#99
  * solidusio/solidus_multi_domain#105

**Checklist:**

### DIFF
--- a/core/app/models/spree/shipping_method.rb
+++ b/core/app/models/spree/shipping_method.rb
@@ -41,7 +41,7 @@ module Spree
 
     scope :available_to_store, ->(store) do
       raise ArgumentError, "You must provide a store" if store.nil?
-      store.shipping_methods.empty? ? all : where(id: store.shipping_method_ids)
+      store.shipping_methods.empty? ? all : where(id: store.shipping_methods.ids)
     end
 
     # @param shipping_category_ids [Array<Integer>] ids of desired shipping categories
--- a/core/spec/models/spree/shipping_method_spec.rb
+++ b/core/spec/models/spree/shipping_method_spec.rb
@@ -257,4 +257,36 @@ RSpec.describe Spree::ShippingMethod, type: :model do
       it { should == 'back_end' }
     end
   end
+
+  describe '.available_to_store' do
+    let(:store) { create(:store) }
+    let(:first_shipping_method) { create(:shipping_method, stores: [store]) }
+    let(:second_shipping_method) { create(:shipping_method, stores: [store]) }
+
+    subject { [first_shipping_method, second_shipping_method] }
+
+    it 'raises an exception if no store is passed as argument' do
+      expect {
+        described_class.available_to_store(nil)
+      }.to raise_exception(ArgumentError, 'You must provide a store')
+    end
+
+    context 'when the store has no shipping methods associated' do
+      before { store.shipping_methods = [] }
+
+      it 'returns all shipping methods' do
+        expect(store.shipping_methods).to eq([])
+        expect(described_class.available_to_store(store)).to eq(subject)
+      end
+    end
+
+    context 'when the store has shipping methods associated' do
+      before { create(:shipping_method) }
+
+      it 'returns the associated records' do
+        expect(store.shipping_methods).to eq(subject)
+        expect(described_class.available_to_store(store)).to eq(subject)
+      end
+    end
+  end
 end
--- a/frontend/spec/features/checkout_spec.rb
+++ b/frontend/spec/features/checkout_spec.rb
@@ -672,6 +672,57 @@ describe "Checkout", type: :feature, inaccessible: true do
     end
   end
 
+  # Regression test for: https://github.com/solidusio/solidus/issues/2998
+  context 'when two shipping categories are available' do
+    let!(:first_category) { create(:shipping_category) }
+    let!(:second_category) { create(:shipping_category) }
+
+    let!(:first_shipping_method) do
+      create(:shipping_method,
+             shipping_categories: [first_category],
+             stores: [store])
+    end
+
+    let!(:second_shipping_method) do
+      create(:shipping_method,
+             shipping_categories: [second_category],
+             stores: [store])
+    end
+
+    context 'assigned to two different products' do
+      let!(:first_product) do
+        create(:product,
+               name: 'First product',
+               shipping_category: first_category)
+      end
+
+      let!(:second_product) do
+        create(:product,
+               name: 'Second product',
+               shipping_category: second_category)
+      end
+
+      before do
+        stock_location.stock_items.update_all(count_on_hand: 10)
+      end
+
+      it 'transitions successfully to the delivery step', js: true do
+        visit spree.product_path(first_product)
+        click_button 'add-to-cart-button'
+        visit spree.product_path(second_product)
+        click_button 'add-to-cart-button'
+
+        click_button 'Checkout'
+
+        fill_in_address
+        fill_in 'order_email', with: 'test@example.com'
+        click_button 'Save and Continue'
+
+        expect(Spree::Order.last.state).to eq('delivery')
+      end
+    end
+  end
+
   def fill_in_credit_card(number: "4111 1111 1111 1111")
     fill_in "Name on card", with: 'Mary Doe'
     fill_in_with_force "Card Number", with: number


core/app/models/spree/shipping_method.rb
```diff
@@ -41,7 +41,7 @@ class ShippingMethod < Spree::Base
 
     scope :available_to_store, ->(store) do
       raise ArgumentError, "You must provide a store" if store.nil?
-      store.shipping_methods.empty? ? all : where(id: store.shipping_method_ids)
+      store.shipping_methods.empty? ? all : where(id: store.shipping_methods.ids)
```
kennyadsl: What's the difference here? Shouldn't they return the same thing?

aitbw: They should; I checked the generated SQL on the sandbox and the output was the same for both calls, but somehow, the result is not the same. (at least, not in this scenario.)

I will investigate further to understand what's going on here.

aitbw: @kennyadsl I added this:

```ruby
puts store.shipping_method_ids.inspect
```

to the scope and found that `store.shipping_method_ids` is returning **just one** shipping method ID, whereas `store.shipping_methods.ids` is returning **all** the IDs for the shipping methods associated to the store.

Can't understand why this is happening, as the generated SQL is exactly the same for both, and some unit tests I created for this work as expected with either. Maybe there's a bug present in Rails (as mentioned first by @afdev82 in [here](https://github.com/solidusio/solidus/issues/2998#issuecomment-446933585)) or something else we're missing out.

kennyadsl: Trying this out and it returns **just one** shipping method ID also for:

```ruby
ActiveRecord::Base.uncached { store.shipping_method_ids } # => [3]
```

in the test you wrote.

Also, the query seems a bit too much for what it should do:

```sql
(0.6ms)
SELECT DISTINCT "spree_shipping_methods"."id"
FROM "spree_shipping_methods"
LEFT OUTER JOIN "spree_shipping_method_stock_locations" ON "spree_shipping_methods"."id" = "spree_shipping_method_stock_locations"."shipping_method_id"
INNER JOIN "spree_store_shipping_methods" ON "spree_shipping_methods"."id" = "spree_store_shipping_methods"."shipping_method_id"
WHERE "spree_shipping_methods"."deleted_at" IS NULL
  AND "spree_shipping_methods"."id" IN
    (SELECT "spree_shipping_methods"."id"
     FROM "spree_shipping_methods"
     INNER JOIN "spree_shipping_method_categories" ON "spree_shipping_method_categories"."shipping_method_id" = "spree_shipping_methods"."id"
     WHERE "spree_shipping_methods"."deleted_at" IS NULL
       AND "spree_shipping_method_categories"."shipping_category_id" = ?
     GROUP BY spree_shipping_methods.id
     HAVING COUNT(DISTINCT "spree_shipping_method_categories"."id") = 1)
  AND ("spree_shipping_methods"."available_to_all" = 1
       OR "spree_shipping_method_stock_locations"."stock_location_id" = 1)
  AND "spree_store_shipping_methods"."store_id" = ? [["shipping_category_id", 2],
      ["store_id", 1]]
```

There are some suspect join with categories, which I'm not sure why are there and if they are needed here. Can you please investigate a little bit more?

aitbw: Yeah sure, let me have a further look at it

spaghetticode:
@aitbw 👋hi Angel, did you have any chance to continue working on this?

aitbw:
@spaghetticode, hello Andrea! I've been trying to find out why my proposed change works, but with no avail so far :sweat:

cedum:
This seems to be fixed in Rails 6 with this commit:
https://github.com/rails/rails/commit/17f2f3054c218485fff598966c39cd3d8a6f6056#diff-c47e1c26ae8a3d486119e0cc91f40a30
By applying this patch to Rails 5.2 `store.shipping_method_ids` starts to behave as expected.

In a scenario with two products with different shipping categories, we get two calls to this scope for each product: the first one seems correct, the second one instead returns the cached result from the previous one.
I'm quite ignorant about ActiveRecord internals, also the scope chaining here is extreme to be able to understand the behavior easily, however, it should be safe accepting the solution in this PR.

kennyadsl:
@cedum nice finding! 

@aitbw 👋👋👋can you please give us (@cedum or me) access to your solidus fork so we can rebase this PR against master? Or if you prefer and/or have time, please do that by yourself.

aitbw:
@kennyadsl, hi Alberto! :wave: Done! :heavy_check_mark: 

Removed the first commit based on @cedum's comment, but I can add it again if necessary.

cedum:
hey @aitbw ! Thank you 🙏 
It would be nice to have also the second commit since it fixes this scope for Rails 5.2 (the commit from the rails repo I linked has been merged only in rails 6 unfortunately)

aitbw:
Ooooh ok ok, got it! @cedum, sorry about that :sweat_smile:

ericsaupe:
Sounds like this fixes #3315 as well if you wanted to put that in the description

Issue #3138, opened 2019-03-13
RFC: Deleting users with associated orders


aitbw:
After talking with @elia and @jacobherrington about how we should tackle #3126, I think it's better for everybody to have a say on this particular issue. (I think #2221 is also related to this?)

Initially, I plan to submit a PR to hide the `delete` button for those users that have **at least** one order associated and then work from there on whatever the community consensus is.

The proposed ideas to completely fix the aforementioned issue are the following:

* Create a ghost / fake user, like Github does with @ghost, to inherit the orders of the deleted users
* Keep the users, but remove all the sensitive data (i.e.: email address, billing/shipping addresses) and replace their names with some like `Deleted user <some_id>`

Take into consideration that this is something that stores administrators **must be** aware of, it should be GDPR compliant and last but not least, orders **can't** be deleted whatsoever.

Let me know what you think! :raised_hands:

stem:
@epicery, we keep the user object and scramble it's data.
Doing so enable us to keep stats about our average lifetime value, average order interval, average order count, etc. Migrating all orders to a "ghost" user will certainly break that.

jacobherrington:
I agree with deleting as much PII as possible and renaming the user to "Deleted user <some_id>. I'm not sure on how far we can go with scrambling data before we run into other problems.

kennyadsl:
I agree that we should explore the approach proposed by @stem if that's compliant with GDPR. Are there some rules to respect for not being able to reverse scrambled data with real users data?

seand7565:
We could use SecureRandom to generate a new email - eg.`"#{SecureRandom.hex}@deleted.com"` - to remove unscrambling as a possibility.

brchristian:
Hi all, have we reached consensus here? We are getting requests from users to delete their accounts but I can’t tell if there is a Solidus best practice for that. Where has this conversation arrived now that we’re 2 years out? I’m happy to submit PRs if needed.

kennyadsl:
Hey @brchristian, I think we should go with extracting the current logic that deletes users in a configurable class that contains the logic for that action. In the beginning, it should just be identical to what we have now. After that, we can provide another opt-in class that does something more advanced like scrambling data. I suspect this is not enough because each store can have users' references in other tools (like marketing or OMS integrations) and they will need to add custom logic to remove users from those platforms as well. What do you think?

brchristian:
[...]

Issue #3126, opened 2019-02-27
Deleting a user with orders throws Missing Template exception


brchristian:
When attempting to delete a user from the backend who has orders, the deletion will fail and a `Missing Template` error screen will be shown. (Screenshot below.)

**Solidus Version**:
2.8.1

**To Reproduce**
Create new 2.8.1 app.
Complete an order with a user.
Attempt to delete user from backend.

**Current behavior**
Missing Template exception will be seen.

**Expected behavior**
I would expect either to see an explanation to the effect of ``Cannot delete user with orders``, or a dialogue ``This user has orders. Continue?``, or a successful deletion.

**Screenshots**
<img width="1253" alt="screen shot 2019-02-27 at 12 56 19 pm" src="https://user-images.githubusercontent.com/2460418/53522611-9e2d1600-3a8f-11e9-873f-45166f08c0f8.png">

**Desktop (please complete the following information):**
 - OS: MacOS Mojave
 - Browser Safari
 - Version 12.0.3

jacobherrington:
This still exists in master. I'll try to make a fix today.

jacobherrington:
Just as an update, the change in #3139 does not close this issue. It is only a temporary fix.

tvdeyen:
Fixed with https://github.com/solidusio/solidus/pull/4415

PR #3139, opened 2019-03-13
Don't allow deletion for users with orders

aitbw:
**Description**

This PR hides the `delete` button for users that have **at least** one order associated to their profiles, which aims to serve as a starting point to solve #3126.

Ref: #3138

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/users/index.html.erb
+++ b/backend/app/views/spree/admin/users/index.html.erb
@@ -89,7 +89,7 @@
           <% if can?(:edit, user) %>
             <%= link_to_edit user, no_text: true %>
           <% end %>
-          <% if can?(:destroy, user) %>
+          <% if can?(:destroy, user) && user.orders.count.zero? %>
             <%= link_to_delete user, no_text: true %>
           <% end %>
         </td>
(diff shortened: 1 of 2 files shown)


kennyadsl:
@aitbw thanks!

backend/app/views/spree/admin/users/index.html.erb
```diff
@@ -89,7 +89,7 @@
           <% if can?(:edit, user) %>
             <%= link_to_edit user, no_text: true %>
           <% end %>
-          <% if can?(:destroy, user) %>
+          <% if can?(:destroy, user) && user.orders.count.zero? %>
```
JuanCrg90: what is the advantage of doing `user.orders.count.zero? ` instead of `user.orders.empty?`

aitbw: None, but since both snippets generate a SQL query, they're pretty much exchangeable. Your suggestion is less verbose and feels more natural, though.

aitbw:
Hi @kainio, sorry for the late reply —yes, you're correct; this PR aims to serve as a workaround.

Since the bug reported in #3126 only affects the backend side of Solidus, I think it's safe to assume admins won't be trying to hack their own stores :smile_cat:

I personally am working on a proper fix for the aforementioned issue, but we still haven't reached consensus on #3138 —feel free to join the discussion :slightly_smiling_face:

PR #3135, opened 2019-03-07
Align some deprecation messages in Order model

elia:
**Description**

After update! => recalculate rename.

ref:
- https://github.com/solidusio/solidus/pull/1689
- https://github.com/solidusio/solidus/pull/2072

~~❓should `create_tax_charge!` implementation be replaced with a call to `recalculate` too?~~

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
- [x] I have updated Guides and README accordingly to this change (if needed)
- [x] I have added tests to cover this change (if needed)

PR #3140, opened 2019-03-13
Fix order deprecation messages

twist900:
**Description**

Fix deprecation warnings as both `#set_shipments_cost` and `#create_tax_charge`
are now handled by a single call to `#recalculate` and not by `#update!`.

**Checklist:**


aitbw:
I think this is a duplicate of #3135?

twist900:
@aitbw you are right, closing

PR #3141, opened 2019-03-13
Tentative fix for flaky specs

kennyadsl:
**Description**

This PR is an attempt to fix flaky specs:

- [Admin Product, chaning taxons feature spec](https://circleci.com/gh/solidusio/solidus/11375#tests/containers/1)
- [Location Filter Unit spec](https://circleci.com/gh/nebulab/solidus/1701)
- [Frontend checkout sources specs](https://circleci.com/gh/nebulab/solidus/1726#tests/containers/2)

For more info it's better to take a look at individual commits.

**Checklist:**

### DIFF
--- a/core/spec/models/spree/stock/location_filter/active_spec.rb
+++ b/core/spec/models/spree/stock/location_filter/active_spec.rb
@@ -8,16 +8,13 @@ module Spree
       RSpec.describe Active, type: :model do
         subject { described_class.new(stock_locations, order) }
 
-        let(:stock_locations) { instance_double('Spree::StockLocation::ActiveRecord_Relation') }
+        let!(:active_stock_location) { create(:stock_location) }
+        let!(:inactive_stock_location) { create(:stock_location, active: false) }
+        let(:stock_locations) { Spree::StockLocation.all }
         let(:order) { instance_double('Spree::Order') }
 
         it 'returns only active stock locations' do
-          active_stock_locations = instance_double('Spree::StockLocation::ActiveRecord_Relation')
-
-          without_partial_double_verification do
-            expect(stock_locations).to receive(:active) { active_stock_locations }
-          end
-          expect(subject.filter).to eq(active_stock_locations)
+          expect(subject.filter).to eq([active_stock_location])
         end
       end
     end
--- a/core/spec/models/spree/stock/location_filter/active_spec.rb
+++ b/core/spec/models/spree/stock/location_filter/active_spec.rb
@@ -13,8 +13,10 @@ module Spree
 
         it 'returns only active stock locations' do
           active_stock_locations = instance_double('Spree::StockLocation::ActiveRecord_Relation')
-          expect(stock_locations).to receive(:active) { active_stock_locations }
 
+          without_partial_double_verification do
+            expect(stock_locations).to receive(:active) { active_stock_locations }
+          end
           expect(subject.filter).to eq(active_stock_locations)
         end
       end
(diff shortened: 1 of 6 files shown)


kennyadsl:
This is passing, I rerun [this workflow](https://circleci.com/gh/nebulab/workflows/solidus/tree/kennyadsl%2Ffix-flacky-specs) several times and it is always green (first failure is due to a force push, not a real failure). I think we can try to merge this.

kennyadsl:
It failed again on the CI :(

https://circleci.com/gh/nebulab/solidus/1625#tests/containers/1

kennyadsl:
By looking at the [last failure on CircleCI](https://circleci.com/gh/nebulab/solidus/1625#tests/containers/1) it's evident what happens:

The page fails in this state (at least when the screenshot is taken):

 img

and the failure says:

```
Failure/Error: expect(page).to have_css(".select2-search-choice", text: taxon.name)
  expected to find visible css ".select2-search-choice" with text "Clothing" but there were no matches. Also found "Ruby on Rails", which matched the selector but not all filters. 
```

This is the failing code:

https://github.com/solidusio/solidus/blob/7ca285d9475192cd7277cbd51d660fe2eaadeb10/core/lib/spree/testing_support/capybara_ext.rb#L93-L98

`page.find` should wait for 10 ([`Capybara.default_max_wait_time`](https://github.com/solidusio/solidus/blob/7ca285d9475192cd7277cbd51d660fe2eaadeb10/.circleci/config.yml#L8)) seconds, so I don't think it's an issue related to the page waiting for div to be populated correctly. Also, if `Clothing` were not found and clicked in the results div, Capybara would raise an exception. So for some reason, that result is found, clicked but it's not doing anything. I'll put some more debugging code, so it will be more clear what's happening.

core/spec/models/spree/stock/location_filter/active_spec.rb
```diff
@@ -13,8 +13,10 @@ module LocationFilter
 
         it 'returns only active stock locations' do
           active_stock_locations = instance_double('Spree::StockLocation::ActiveRecord_Relation')
-          expect(stock_locations).to receive(:active) { active_stock_locations }
 
+          without_partial_double_verification do
+            expect(stock_locations).to receive(:active) { active_stock_locations }
+          end
```
tvdeyen: Hmm, I am not a huge fan of these kind of specs. Could we rewrite the spec into something that actually test the result instead of stubbing the implementation? 

Create two stock locations, one active, one not. Expect the active one is the result.

jacobherrington:
+1 for getting rid of Select2. Something I'd love to tackle if I had the time.

tvdeyen:
> +1 for getting rid of select2. Something I'd love to tackle if I had the time.

@jacobherrington Awesome! keep in mind, though, that we already only using select2 for multiple selects and autocomplete. Normal selects are already browser selects. John looked into an alternative, but didn’t finished his work.

kennyadsl:
@tvdeyen I've done the suggested changes to specs. I also changed Stock Location Sorter specs that I wrongly used as a reference when writing these ones.

kennyadsl:
Aaaaand... there's another flaky: https://circleci.com/gh/nebulab/solidus/1726#tests/containers/2

tvdeyen:
These feature specs are all written very poorly. One does not expect database changes in a feature spec. One expects changes in the request outcome (html). This particular one should be easily fixed by waiting for a string present on the screen before making that assertion about a change in the database (that happen in another thread and therefore can’t be predictable)

Issue #3075, opened 2019-02-04
Avoid runnning validation during Order#record_ip_address


elia:
`Order#record_ip_address` shouldn't risk blocking the request in case a validation check is not met. Instead of `update_attributes!` I'd suggest using `update_column`.

**Solidus Version**:
Found in 2.4, latest code seems to behaving the same way.

**To Reproduce**
Hard to reproduce, but got notifications for exceptions due to the validation on email.

**Current behavior**
Raises an exception if the order is invalid.

**Expected behavior**
No unrelated exceptions raised when recording the ip.

**Additional context**
related: https://github.com/solidusio/solidus/pull/2068

kennyadsl:
That PR's commit message is [Avoid running validations in current_order](/solidusio/solidus/pull/2068/commits/a5828b8e4343175b5a10281edeacdc1d94a0a2c6) but I think it refers to running validation only when we need to update the order - if IP is not changed it won't.

At the moment I can't see any cons skipping validation there. So 👍 for me!

PR #3145, opened 2019-03-15
Don't run validations in Order#record_ip_address

cedum:
When updating the user's latest IP address for an Order, don't run order validations and callbacks.
The IP address tracking should not interfere with/block the request in case validation check is not met.

Related issue: #3075 

**Checklist:**


kennyadsl:
Thanks @cedum !

PR #3146, opened 2019-03-15
Ensure return from CSS function

fastjames:
**Description**
When compiling SCSS, the newly included `sassc` compiler seems to be more vigilant about logic checks. In particular, it returns an error if a CSS function ends without a `@return` statement. The `very-light` function has the possibility of doing that, which causes such an error when we test our solidus app against v2.8.2.

I am not sure if it's possible to write tests for css functions, and I do not see any existing tests that appear to target them so I assume that's not a requirement right now.

**Checklist:**


kennyadsl:
@fastjames thanks! Can you please post the error you are receiving as well?

fastjames:
Sorry for the delay, I have been unable to get back to this today to reproduce and record the error. Will do so as soon as I can.

PR #3148, opened 2019-03-18
Remove cancancan customizations

coorasse:
**Description**
While testing solidus against cancancan 3.0 I noticed that the usage was "not correct (*Euphemismus*)".
This PR aims to remove the customizations introduced to cancancan, which are not necessary.
The current customizations are not compatible with version 3.0.0 of cancancan since Solidus makes use of `merge` of Abilities. The `3.0.0` has a breaking change which changes how `aliased_actions` are merged.

I removed all custom aliases and replaced `display` with `read`. Also, I corrected the rule which would have caused issues during the migration.

This is probably a breaking change, because it will cause problems for the plugins.
Plugins should now use standard `cancancan` actions and replace `display` with `read`.

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -133,13 +133,13 @@ module Spree
 
       def product_scope
         if can?(:admin, Spree::Product)
-          scope = Spree::Product.with_deleted.accessible_by(current_ability, :read).includes(*product_includes)
+          scope = Spree::Product.with_deleted.accessible_by(current_ability).includes(*product_includes)
 
           unless params[:show_deleted]
             scope = scope.not_deleted
           end
         else
-          scope = Spree::Product.accessible_by(current_ability, :read).available.includes(*product_includes)
+          scope = Spree::Product.accessible_by(current_ability).available.includes(*product_includes)
         end
 
         scope
@@ -159,7 +159,7 @@ module Spree
 
       def authorize_for_order
         @order = Spree::Order.find_by(number: order_id)
-        authorize! :read, @order, order_token
+        authorize! :show, @order, order_token
       end
 
       def lock_order
--- a/api/app/controllers/spree/api/orders_controller.rb
+++ b/api/app/controllers/spree/api/orders_controller.rb
@@ -50,7 +50,7 @@ module Spree
       end
 
       def index
-        authorize! :index, Order
+        authorize! :admin, Order
         orders_includes = [
           :user,
           :payments,
--- a/core/app/models/spree/ability.rb
+++ b/core/app/models/spree/ability.rb
@@ -29,24 +29,11 @@ module Spree
     def initialize(current_user)
       @user = current_user || Spree.user_class.new
 
-      alias_actions
       activate_permission_sets
       register_extension_abilities
     end
 
     private
-
-    def alias_actions
-      clear_aliased_actions
-
-      # override cancan default aliasing (we don't want to differentiate between read and index)
-      alias_action :delete, to: :destroy
-      alias_action :edit, to: :update
-      alias_action :new, to: :create
-      alias_action :new_action, to: :create
-      alias_action :show, to: :read
-      alias_action :index, :read, to: :display
-    end
 
     # Before, this was the only way to extend this ability. Permission sets have been added since.
     # It is recommended to use them instead for extension purposes if possible.
--- a/core/lib/spree/permission_sets/default_customer.rb
+++ b/core/lib/spree/permission_sets/default_customer.rb
@@ -4,11 +4,11 @@ module Spree
   module PermissionSets
     class DefaultCustomer < PermissionSets::Base
       def activate!
-        can :display, Country
-        can :display, OptionType
-        can :display, OptionValue
+        can :read, Country
+        can :read, OptionType
+        can :read, OptionValue
         can :create, Order
-        can [:read, :update], Order do |order, token|
+        can [:show, :edit, :update], Order do |order, token|
           order.user == user || (order.guest_token.present? && token == order.guest_token)
         end
         cannot :update, Order do |order|
@@ -17,20 +17,20 @@ module Spree
         can :create, ReturnAuthorization do |return_authorization|
           return_authorization.order.user == user
         end
-        can [:display, :update], CreditCard, user_id: user.id
-        can :display, Product
-        can :display, ProductProperty
-        can :display, Property
+        can [:read, :update], CreditCard, user_id: user.id
+        can :read, Product
+        can :read, ProductProperty
+        can :read, Property
         can :create, Spree.user_class
         can [:read, :update, :update_email], Spree.user_class, id: user.id
-        can :display, State
-        can :display, StockItem, stock_location: { active: true }
-        can :display, StockLocation, active: true
-        can :display, Taxon
-        can :display, Taxonomy
+        can :read, State
+        can :read, StockItem, stock_location: { active: true }
+        can :read, StockLocation, active: true
+        can :read, Taxon
+        can :read, Taxonomy
         can [:save_in_address_book, :remove_from_address_book], Spree.user_class, id: user.id
-        can [:display, :view_out_of_stock], Variant
-        can :display, Zone
+        can [:read, :view_out_of_stock], Variant
+        can :read, Zone
       end
     end
   end
--- a/core/spec/models/spree/ability_spec.rb
+++ b/core/spec/models/spree/ability_spec.rb
@@ -159,25 +159,27 @@ RSpec.describe Spree::Ability, type: :model do
 
       context 'requested by same user' do
         before(:each) { resource.user = user }
-        it_should_behave_like 'access granted'
+        it { expect(ability).to be_able_to(:show, resource) }
         it_should_behave_like 'no index allowed'
       end
 
       context 'requested by other user' do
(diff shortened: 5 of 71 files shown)


api/app/controllers/spree/api/base_controller.rb
```diff
@@ -133,13 +133,13 @@ def find_product(id)
 
       def product_scope
         if can?(:admin, Spree::Product)
-          scope = Spree::Product.with_deleted.accessible_by(current_ability, :read).includes(*product_includes)
+          scope = Spree::Product.with_deleted.accessible_by(current_ability).includes(*product_includes)
```
coorasse: this is not necessary anymore. The default is `:index` which is by default aliased by `:read`.

api/app/controllers/spree/api/orders_controller.rb
```diff
@@ -49,8 +49,9 @@ def empty
         render plain: nil, status: 204
       end
 
+      # TODO: this action is never accessible. shall we simply remove it?
```
coorasse: this method was never accessible. why don't we remove it?

core/app/models/spree/ability.rb
```diff
@@ -29,25 +29,12 @@ def self.remove_ability(ability)
...
       activate_permission_sets
       register_extension_abilities
     end
 
     private
 
-    def alias_actions
-      clear_aliased_actions
-
-      # override cancan default aliasing (we don't want to differentiate between read and index)
```
coorasse: this is supported already. you can use `read`.

core/app/models/spree/ability.rb
```diff
@@ -29,25 +29,12 @@ def self.remove_ability(ability)
...
-      alias_action :delete, to: :destroy
```
coorasse: `delete` was never used

core/app/models/spree/ability.rb
```diff
@@ -29,25 +29,12 @@ def self.remove_ability(ability)
...
-      alias_action :edit, to: :update
-      alias_action :new, to: :create
-      alias_action :new_action, to: :create
```
coorasse: `new_action` was never used

core/lib/spree/permission_sets/default_customer.rb
```diff
@@ -4,11 +4,11 @@ module Spree
...
       def activate!
-        can :display, Country
-        can :display, OptionType
-        can :display, OptionValue
+        can :read, Country
+        can :read, OptionType
+        can :read, OptionValue
         can :create, Order
-        can [:read, :update], Order do |order, token|
+        can [:show, :edit, :update], Order do |order, token|
```
coorasse: block conditions should be defined only for actions on object instances, not classes, therefore this rule applies only on `:show, :edit, :update`. This was actually working by mistake since version 2.0 of cacancan

core/spec/models/spree/ability_spec.rb
```diff
@@ -254,7 +256,7 @@ def initialize(_user)
       context 'requested by same user' do
         let(:resource) { user }
         it_should_behave_like 'access granted'
-        it_should_behave_like 'no index allowed'
+        it_should_behave_like 'index allowed'
```
coorasse: why should it not be allowed?

kennyadsl:
@coorasse wow, thank you, I love this change! Do you think there's a way to deprecate the old usage any way before removing it?

coorasse:
I cannot really think of a deprecation warning: we could override `can?` and `accessible_by` methods of `cancancan` and raise a warning when `can?` is called with `delete`, `new_action`, `read` or `display` and when `accessible_by` is called with `:read` or `:display`.

The calls to `accessible_by(ability, :read)` should be replaced by `accessible_by(ability)` but this would not work before this update! ⚠️ 

Also `can? :read, resource` should be changed to the specific action `can? :show` or `can? :edit`, etc, but again: it would not work before this change.

The calls to `can? :delete, resource` could already be replaced by `can? :destroy, resource`.

Also the calls to `can? :new_action, resource` could already be replaced by `can? :create, resource`.

The main problem was this alias:
```
alias_action :show, to: :read
alias_action :index, :read, to: :display
```
which caused `read` to be the same of `show` and `display` to be both `read` and `index`...
I am not even sure what this means 😄. If you notice, in fact, the changes I had to do in the code itself were not actually many. It is more a renaming issue. I can make a new comment and show in detail how/why the problem is raised only in version `3.0.0`.

I let you guys decide about how to move here: I have not enough knowledge about the project.
I pointed the finger 😄 and I can provide all help you need but I think this might break quite some stuff.

coorasse:
In cancancan 3.0.0 when two abilities are merged, also the aliased_actions are merged.
Solidus defines the following aliases in base ability:
```
{
:destroy=>[:delete], 
:update=>[:edit], 
:create=>[:new, :new_action], 
:read=>[:show], 
:display=>[:index, :read]
}
```

base ability is the merged with other abilities where the default aliases are defined. The following:

```
{
  :read => [:index, :show],
  :create => [:new],
  :update => [:edit]
}
```

with the following aliases as a result:

```
{
:destroy=>[:delete], 
:update=>[:edit], 
:create=>[:new], 
:read=>[:index, :show], 
:display=>[:index, :read]
}
```

The call that Solidus performs on `Spree::Order.accessible_by(ability, :read)` in version 3.0.0 fails because the following rule:
```
can [:read, :update], Order do |order, token|
  order.user == user || (order.guest_token.present? && token == order.guest_token)
end
```
 is now identified as possible interesting rule and cancancan raises an issue because this is a block rule (`do...end`) defined on a collection action `read`

api/app/controllers/spree/api/base_controller.rb
jkojro: Hi @coorasse! But if `:read` alias was `:show` shouldn't it be changed like this `scope = Spree::Product.with_deleted.accessible_by(current_ability, :show).includes(*product_includes)` to keep same logic? Isn't your change changing scope from show to index now?

coorasse: it changes from `:show` to `:index` but all the rules are defined for `:read` instead of `:display` which was including also `:read`. it's quite complicated, I know, I don't know why all these aliases were generated before.  `accessible_by(ability, :show)` would not make sense because you want a list of objects. `:show` is for single instances.

jkojro: Yeah, that make sense. So for now, when is `alias_action :show, to: :read` this scope with argument `(current_ability, :read)` had no authorization at all I gues?

api/app/controllers/spree/api/orders_controller.rb
jkojro: @coorasse Could you explain why is it unaccessible?

coorasse: it was defined with `:index` and since `:read` was an alias of `:show`, no rule was defined for `:index`. in fact, I replaced it with a random action called `:impossible_action` and no test failed. There is only one test and it checks that this endpoint should not be accessible 🤷‍♂️ . That's why I asked myself: "why don't we remove it?"

api/app/controllers/spree/api/base_controller.rb
coorasse: `:read` was aliased by `:display` and rules were (almost) always defined on `:display`. You can see that most of the changes are about that: https://github.com/solidusio/solidus/pull/3148/files#diff-82b30f937458e7c1052c326354dc18cbL8

api/app/controllers/spree/api/orders_controller.rb
jkojro: @coorasse As I've tested, `:manage` gives acces to all rules, also theese not defined, so a user with `can :manage, Order` will go through `authorize! :impossible_action, Order` too. So it could be replaced with `authorize! :admin, Order` I think and it would keep eligible logic.

api/app/controllers/spree/api/orders_controller.rb
```diff
@@ -49,8 +49,9 @@ def empty
...
       def index
-        authorize! :index, Order
```
jkojro: `authorize! :admin, Order`

api/app/controllers/spree/api/orders_controller.rb
coorasse: Oh right! `manage` authorizes you! I forgot that 🤦‍♂️ . To do it properly here we should leave `authorize! :index, Order` but then [this test](https://github.com/solidusio/solidus/blob/29fe2cc432ac336ee4dd407f677512c28b35c001/api/spec/requests/spree/api/orders_controller_spec.rb#L201) will fail. So I'd say we could use `authorize! :admin, Order` and keep the current behaviour.

api/app/controllers/spree/api/base_controller.rb
tvdeyen: I agree that `:index` is the right thing here, as `:show` would need to check every single instance of a Product if its able to be shown to the user, what is definitely not what we want here.

api/app/controllers/spree/api/orders_controller.rb
```diff
@@ -50,7 +50,7 @@ def empty
       end
 
       def index
-        authorize! :index, Order
+        authorize! :admin, Order
```
tvdeyen: Why this change?

kennyadsl:
Closing in favor of #3701.

PR #3150, opened 2019-03-20
When editing prices keep the currency locked.

peterberkenbosch:
**Description**

When editing a price that has a different currency then the default currency set, the edit form defaults the currency to the default currency. 

This PR fixes that by passing in the price currency to the currency form partial. It also improves the feature spec for editing prices by actually verifying that the edit worked.

before: https://cl.ly/420fb991710b
after: https://cl.ly/2b141684337f

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/prices/_form.html.erb
+++ b/backend/app/views/spree/admin/prices/_form.html.erb
@@ -26,7 +26,7 @@
       <div data-hook="admin_product_price_form_amount" class="col-4">
         <%= f.field_container :price do %>
           <%= f.label :price %>
-          <%= render "spree/admin/shared/number_with_currency", f: f, amount_attr: :price, currency_attr: :currency %>
+          <%= render "spree/admin/shared/number_with_currency", f: f, amount_attr: :price, currency_attr: :currency, currency: @price.currency %>
         <% end %>
       </div>
   </div>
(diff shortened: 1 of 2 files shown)


peterberkenbosch:
seems there are some unrelated failing specs on the rails51, I do see them in master now and then too actually. @kennyadsl could you restart the circleci workflow?

kennyadsl:
@peterberkenbosch done!

kennyadsl:
@peterberkenbosch Thanks, I think this works, I just have a doubt about if the currency choice should be disabled in the form. Can't/Shouldn't the price currency be edited by an admin user?

peterberkenbosch:
I kept the existing expected(?) functionality in tact in this PR. Noticed the same though. Wanted to keep the PR to the bugfix, but I can update it to have the currency change as well. WDYT @kennyadsl

kennyadsl:
I'm fine with the fix, just wondering. Maybe some git blame could help us remember why currency is not changeable.

peterberkenbosch:
The currency component was introduced like that. It's used on multiple places actually.

kennyadsl:
@peterberkenbosch can you please rebase now, I hope flaky specs are fixed with #3141 .

peterberkenbosch:
@kennyadsl indeed! Specs are passing! Thanks

peterberkenbosch:
Thanks @jacobherrington, will make another PR for that ;)

Issue #3149, opened 2019-03-18
Breadcrumbs displays invalid emails modifying users 


kennyadsl:
When changing a user's email in the backend (`/admin/users/:id`), if there is a validation error on the email, breadcrumbs of the page are updated with the wrong, invalid value.

**Solidus Version**:

master

**To Reproduce**

In sandbox:

1. Go to Admin -> User -> Any user -> Account tab
2. Change user's email with the email taken by another user and save
3. Click Update button

**Current behavior**

Breadcrumbs are updated with the wrong email, even if there's an error on the page

**Expected behavior**

The old, persisted and valid email is displayed in breadcrumbs

**Screenshots**

<img width="1098" alt="breadcrumbs-bug" src="https://user-images.githubusercontent.com/167946/54568033-1ffcc900-49d6-11e9-9ef1-ecf72681bf33.png">

👆 I'm actually editing `admin@example.com`

**Additional context**

I suppose this happens because breacrumbs is using the ivar (maybe @user, or @object from [`Spree::Admin::ResourceController`](https://github.com/solidusio/solidus/blob/14a9eb15eb22c92ef5e00e4daa74f96e58ced7b9/backend/app/controllers/spree/admin/resource_controller.rb)) that contains the instance of the User with errors (and previously submitted params) and not the persisted one, which is valid.

I'm afraid this could happen in other pages as well, where the value that the user is trying to change to an invalid one is used to create breadcrumbs of that page.

jtapia:
@kennyadsl I'm not sure if this is the best approach to tackle this issue, but here is a possible solution https://github.com/solidusio/solidus/pull/3152

chrean:
A PR was accepted to fix an edge case, and we are closing this because we are in the process of building a new version of the Solidus Admin, which will make this issue soon irrelevant.

PR #3152, opened 2019-03-21
Fix issue with user breadcrumbs

jtapia:
### Issue
- https://github.com/solidusio/solidus/issues/3149

### Preview
Screen Shot 2019-03-21 at 4 35 12 PM 1

### DIFF
--- a/backend/app/views/spree/admin/users/edit.html.erb
+++ b/backend/app/views/spree/admin/users/edit.html.erb
@@ -1,6 +1,5 @@
 <% admin_breadcrumb(link_to plural_resource_name(Spree::LegacyUser), spree.admin_users_path) %>
-<% admin_breadcrumb(@user.email) %>
-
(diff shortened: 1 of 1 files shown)


kennyadsl:
This could be a solution (I'm wondering if using only `@user.email_was` without any if fixes the issue as well since I think it always returns the right email) anyway I think this issue could be present in several other admin pages and we should come up with a more robust long-term solution.

I'm taking a look at https://github.com/weppos/breadcrumbs_on_rails which is old and unmaintend but it's very few lines of code that we could add to solidus easily maybe. It provides an interface in the controller to set breadcrumbs, this way we could maybe define breadcrumbs before calling the `update_attributes` that makes the instance dirty? How does it play with ResourceController? Maybe worth exploring a little bit more.

aldesantis:
@jtapia would like to get this in. Could you rebase?

jtapia:
@aldesantis done

aldesantis:
@jtapia thanks!

PR #3157, opened 2019-04-02
Improve promotion statuses

JuanCrg90:
**Description**

The current promotions index page only shows two statuses, active and inactive, this doesn't give enough information when the store has many promotions. This adds the unstarted and expired promotion statuses looking for an improvement in the user experience in this admin section.

Before
image

After
image

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/promotions/index.html.erb
+++ b/backend/app/views/spree/admin/promotions/index.html.erb
@@ -82,7 +82,7 @@
           </td>
           <td>
             <span class="pill pill-<%= promotion.active? ? 'active' : 'inactive' %>">
-              <%= t(promotion.active?? :active : :inactive, scope: 'spree') %>
+              <%= t(admin_promotion_status(promotion), scope: 'spree.admin.promotion_status') %>
             </span>
           </td>
           <td>
--- a/backend/spec/features/admin/promotions/promotion_spec.rb
+++ b/backend/spec/features/admin/promotions/promotion_spec.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+feature 'Promotions' do
+  stub_authorization!
+
+  context 'index' do
+    context 'when no promotions' do
+      scenario 'shows no promotions found message' do
+        visit spree.admin_promotions_path
+        expect(page).to have_content('No Promotions found.')
+      end
+    end
+
+    context 'when promotion is active' do
+      given!(:promotion) { create :promotion }
+
+      scenario 'promotion status is active' do
+        visit spree.admin_promotions_path
+
+        within_row(1) do
+          expect(column_text(3)).to eq("Active")
+        end
+      end
+    end
+
+    context 'when promotion is in the future' do
+      given!(:promotion) { create :promotion, starts_at: 1.day.after }
+
+      scenario 'promotion status is not started' do
+        visit spree.admin_promotions_path
+
+        within_row(1) do
+          expect(column_text(3)).to eq("Not started")
+        end
+      end
+    end
+
+    context 'when promotion is in the past' do
+      given!(:promotion) { create :promotion, expires_at: 1.day.ago }
+
+      scenario 'promotion status is expired' do
+        visit spree.admin_promotions_path
+
+        within_row(1) do
+          expect(column_text(3)).to eq("Expired")
+        end
+      end
+    end
+  end
+end
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -64,9 +64,24 @@ module Spree
       super
     end
 
+    def not_started?
+      !started?
+    end
+
+    def started?
+      starts_at.nil? || starts_at < Time.current
+    end
+
+    def expired?
+      expires_at.present? && expires_at < Time.current
+    end
+
+    def not_expired?
+      !expired?
+    end
+
     def active?
-      (starts_at.nil? || starts_at < Time.current) &&
-        (expires_at.nil? || expires_at > Time.current)
+      started? && not_expired?
     end
 
     def inactive?
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -850,6 +850,11 @@ en:
           expires_at_placeholder: Never
           general: General
           starts_at_placeholder: Immediately
+      promotion_status:
+        active: Active
+        expired: Expired
+        inactive: Inactive
+        not_started: Not started
       stock_locations:
         form:
           address: Address
--- a/core/spec/models/spree/promotion_spec.rb
+++ b/core/spec/models/spree/promotion_spec.rb
@@ -384,6 +384,86 @@ RSpec.describe Spree::Promotion, type: :model do
     end
   end
 
+  describe '#not_started?' do
+    let(:promotion) { Spree::Promotion.new(starts_at: starts_at) }
+    subject { promotion.not_started? }
+
+    context 'no starts_at date' do
+      let(:starts_at) { nil }
+      it { is_expected.to be_falsey }
+    end
+
+    context 'when starts_at date is in the past' do
+      let(:starts_at) { Time.current - 1.day }
+      it { is_expected.to be_falsey }
+    end
+
+    context 'when starts_at date is not already reached' do
+      let(:starts_at) { Time.current + 1.day }
+      it { is_expected.to be_truthy }
+    end
+  end
+
+  describe '#started?' do
+    let(:promotion) { Spree::Promotion.new(starts_at: starts_at) }
+    subject { promotion.started? }
+
+    context 'when no starts_at date' do
+      let(:starts_at) { nil }
+      it { is_expected.to be_truthy }
+    end
+
+    context 'when starts_at date is in the past' do
+      let(:starts_at) { Time.current - 1.day }
+      it { is_expected.to be_truthy }
+    end
+
+    context 'when starts_at date is not already reached' do
+      let(:starts_at) { Time.current + 1.day }
+      it { is_expected.to be_falsey }
+    end
+  end
+
+  describe '#expired?' do
+    let(:promotion) { Spree::Promotion.new(expires_at: expires_at) }
+    subject { promotion.expired? }
+
+    context 'when no expires_at date' do
+      let(:expires_at) { nil }
+      it { is_expected.to be_falsey }
+    end
+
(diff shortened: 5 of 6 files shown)


core/app/models/spree/promotion.rb
```diff
@@ -64,9 +64,24 @@ def as_json(options = {})
       super
     end
 
+    def unstarted?
```
spaghetticode: I may be splitting hairs here, but I think it may be worth being more consistent with naming in order to honor the principle of least surprise :neckbeard: 

We have `not_expired` but also`unstarted`: I think we may prefer to change one of them in order to have the same prefix. So it would be either `unexpired` or `not_started`. What do you think?

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,73 @@
     end
   end
 
+  context '#unstarted?' do
```
spaghetticode: _(Still splitting hairs)_ I think this should be more appropriately a `describe` instead of a `context`. I understand this file uses both methods indifferently, still `describe` sounds more appropriate to me when referring to a method.

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,73 @@
...
+    it 'should not be unstarted if no starts_at date' do
```
spaghetticode: (more hair splitting 💇‍♂️) in my experience using `should` is a bit out of fashion with rspec. [This link](http://www.betterspecs.org/#should) may be one of the reasons, together with the new `expect` syntax that replaced `should`. 
Again, I understand this file is already full of old code that uses `should` and if we want to be consistent we should keep using it...

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,73 @@
...
+      promotion.starts_at = nil
+      expect(promotion.unstarted?).to eq(false)
```
spaghetticode: I find `.to be true` and `.to be false` more easy on the eyes

kennyadsl: there are `to be_truthy` and `to be_falsey` as well.

core/app/models/spree/promotion.rb
JuanCrg90: I will go with `not_started`. what do you think should I update too on the front end?

image

core/spec/models/spree/promotion_spec.rb
JuanCrg90: I will update it, I was trying to keep the file convention. Thanks for the feedback 😄

core/app/models/spree/promotion.rb
spaghetticode: 👍 yes, I think that it's best to be consistent also there with the chosen naming.

JuanCrg90:
@spaghetticode , @kennyadsl I have updated the tests and added integration tests, what do you think 😄 ?

spaghetticode:
@JuanCrg90 looks good to me 👍 

There are a few tests failing, rebasing with master should make everything green... thank you!

backend/spec/features/admin/promotions/promotion_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      scenario 'promotion status is not started' do
+        visit spree.admin_promotions_path
+
+        within_row(1) do
+          expect(column_text(3)).to eq("Not started")
+        end
+      end
+    end
+
+    context 'when promotion was in the past' do
```
aitbw: I think it should be `in the past` in here, WDYT?

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
     end
   end
 
+  describe '#not_started?' do
+    let(:promotion) { Spree::Promotion.new(starts_at: starts_at) }
+    subject { promotion.not_started? }
+
+    context 'no starts_at date' do
+      let(:starts_at) { nil }
+      it{ is_expected.to be_falsey }
```
aitbw: Could you add the missing space between `it` and the opening curly brace? :smile_cat:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+
+    context 'when starts_at date is in the past' do
+      let(:starts_at) { Time.current - 1.day }
+      it{ is_expected.to be_falsey }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+
+    context 'when starts_at date is not already reached' do
+      let(:starts_at) { Time.current + 1.day }
+      it{ is_expected.to be_truthy }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+  end
+
+  describe '#started?' do
+    let(:promotion) { Spree::Promotion.new(starts_at: starts_at) }
+    subject { promotion.started? }
+
+    context 'when no starts_at date' do
+      let(:starts_at) { nil }
+      it{ is_expected.to be_truthy }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+
+    context 'when starts_at date is in the past' do
+      let(:starts_at) { Time.current - 1.day }
+      it{ is_expected.to be_truthy }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+
+    context 'when starts_at date is not already reached' do
+      let(:starts_at) { Time.current + 1.day }
+      it{ is_expected.to be_falsey }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+  end
+
+  describe '#expired?' do
+    let(:promotion) { Spree::Promotion.new(expires_at: expires_at) }
+    subject { promotion.expired? }
+
+    context 'when no expires_at date' do
+      let(:expires_at) { nil }
+      it{ is_expected.to be_falsey }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+
+    context 'when expires_at date is not already reached' do
+      let(:expires_at) { Time.current + 1.days }
+      it{ is_expected.to be_falsey }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+
+    context 'when expires_at date is in the past' do
+      let(:expires_at) { Time.current - 1.days }
+      it{ is_expected.to be_truthy }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+  end
+
+  describe '#not_expired?' do
+    let(:promotion) { Spree::Promotion.new(expires_at: expires_at) }
+    subject { promotion.not_expired? }
+
+    context 'when no expired_at date' do
+      let(:expires_at) { nil }
+      it{ is_expected.to be_truthy }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+
+    context 'when expires_at date is not already reached' do
+      let(:expires_at) { Time.current + 1.days }
+      it{ is_expected.to be_truthy }
```
aitbw: Same as before :ok_woman:

core/spec/models/spree/promotion_spec.rb
```diff
@@ -384,6 +384,86 @@
...
+    end
+
+    context 'when expires_at date is in the past' do
+      let(:expires_at) { Time.current - 1.days }
+      it{ is_expected.to be_falsey }
```
aitbw: Same as before :ok_woman:

backend/app/views/spree/admin/promotions/index.html.erb
```diff
@@ -82,7 +82,15 @@
           </td>
           <td>
             <span class="pill pill-<%= promotion.active? ? 'active' : 'inactive' %>">
-              <%= t(promotion.active?? :active : :inactive, scope: 'spree') %>
+              <% if promotion.active? %>
+                <%= t(:active, scope: 'spree') %>
```
kennyadsl: what about adding a scope to this translation? Like `spree.promotion_status.active`?

JuanCrg90:
Updated @kennyadsl 😄 what do you think?

core/config/locales/en.yml
```diff
@@ -847,6 +847,11 @@ en:
           expires_at_placeholder: Never
           general: General
           starts_at_placeholder: Immediately
+      promotion_status:
+          active: Active
```
kennyadsl: Too much indented here?

JuanCrg90: updated, thanks for the observation 😄

core/app/models/spree/promotion.rb
```diff
@@ -64,9 +64,24 @@ def as_json(options = {})
       super
     end
 
+    def not_started?
+      starts_at.present? && starts_at > Time.current
+    end
+
+    def started?
+      !not_started?
```
ericsaupe: I think I would switch where the logic resided so `started?` contained the time check and `not_started?` could be `!started?` which reads a bit better. As it's written here it's sort of a double-negative for `started?` being "not not_started?"

JuanCrg90: I have applied the change @ericsaupe thanks for the feedback. 

What do you think 😄 ?

PR #3161, opened 2019-04-03
Enable extension developers to customize the namespace

mdesantis:
**Description**

If you are developing an extension and you want to use `Rake::Task['extension:test_app']` you are forced to have a file named exactly as the capitalized version of the extension namespace name, so if you name your namespace extension e.g. SolidusGraphQL you are forced to create a file named `solidus_graph_q_l`, otherwise the following error is raised:

```
> bundle exec rake test_app --trace
** Invoke default (first_time)
** Invoke first_run (first_time)
** Execute first_run
** Invoke test_app (first_time)
** Execute test_app
** Invoke extension:test_app (first_time)
** Execute extension:test_app
** Invoke common:test_app (first_time)
** Execute common:test_app
rake aborted!
LoadError: cannot load such file -- solidus_graph_q_l
```

 This change allows the developer to customize the extension namespace setting a `LIB_NAMESPACE` environment variable.

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/common_rake.rb
+++ b/core/lib/spree/testing_support/common_rake.rb
@@ -24,7 +24,7 @@ namespace :common do
     begin
       require "generators/#{ENV['LIB_NAME']}/install/install_generator"
       puts 'Running extension installation generator...'
-      "#{ENV['LIB_NAME'].camelize}::Generators::InstallGenerator".constantize.start(["--auto-run-migrations"])
+      "#{ENV['LIB_NAMESPACE'] || ENV['LIB_NAME'].camelize}::Generators::InstallGenerator".constantize.start(["--auto-run-migrations"])
     rescue LoadError
       # No extension generator to run
     end


core/lib/spree/testing_support/common_rake.rb
```diff
@@ -24,7 +24,7 @@
     begin
       require "generators/#{ENV['LIB_NAME']}/install/install_generator"
       puts 'Running extension installation generator...'
-      "#{ENV['LIB_NAME'].camelize}::Generators::InstallGenerator".constantize.start(["--auto-run-migrations"])
+      "#{ENV['NAMESPACE'] || ENV['LIB_NAME'].camelize}::Generators::InstallGenerator".constantize.start(["--auto-run-migrations"])
```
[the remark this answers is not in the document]
mdesantis: Good point, I agree

core/lib/spree/testing_support/common_rake.rb
```diff
@@ -24,7 +24,7 @@
     begin
       require "generators/#{ENV['LIB_NAME']}/install/install_generator"
       puts 'Running extension installation generator...'
-      "#{ENV['LIB_NAME'].camelize}::Generators::InstallGenerator".constantize.start(["--auto-run-migrations"])
+      "#{ENV['LIB_NAMESPACE'] || ENV['LIB_NAME'].camelize}::Generators::InstallGenerator".constantize.start(["--auto-run-migrations"])
```
ericsaupe: Does `LIB_NAMESPACE` need to be `camelize` as well? Should this be `#{(ENV['LIB_NAMESPACE'] || ENV['LIB_NAME']).camelize}`

mdesantis: @ericsaupe I don't think so, because `LIB_NAMESPACE` is meant to be the Solidus extension gem namespace, it isn't used as a file path, so it doesn't make sense to pass it in a file path form; I meant to use `LIB_NAMESPACE` in order to skip transformations, so f.e. in this way:

```
LIB_NAME=solidus_graphql_api LIB_NAMESPACE=SolidusGraphQLAPI bundle exec rake test_app
```

Do you think we could refactor in a better way the flow? I was thinking about that but unfortunately it didn't come me to mind anything better

ericsaupe: Makes sense to me! Thanks for the explanation.

kennyadsl:
@mdesantis can you please rebase against master now?

mdesantis:
@kennyadsl FYI I rebased with master a couple of hours ago and surprisingly tests are still failing; I'm going to try again

kennyadsl:
I've just merged https://github.com/solidusio/solidus/pull/3202 that should fix the build

Issue #2241, opened 2017-09-28
bug: user with display stock permission can edit product stock


loicginoux:
### Steps to reproduce
- set user permission to display stock and variants, permission sets: `StockDisplay` and `ProductDisplay` ), s without `manage` access.
- go to backend product stock page -> /admin/products/your-product/stock

### Expected behavior
- icon "edit" should not be visible
- ajax edit action should not be possible

### Actual behavior
- icon is visible 
- ajax edit action is not possible, "ressource not found" (good!), so at least the backend does not allow the modification

### System configuration
**Solidus Version**: 2.2.1

So basically, that's a matter of hiding the edit icon when user has no stock management permission.
files to look at: 
- `solidus_backend-2.2.1/app/views/spree/admin/stock_items/_stock_management.html.erb`
- `solidus_backend-2.2.1/app/assets/javascripts/spree/backend/stock_management/index_update_forms.coffee`

If this is a valid bug, I could submit a PR.

BravoSimone:
Hi, I really want to take care of this issue but I can't understand the first step to reproduce the bug, can you provide more detailed instructions?

loicginoux:
Hi, Sorry that may be my bad english :S 
You need to set up a role that has only "display stock permission" and "display product". He should not be able to manage neither stock nor product.
Let me know if you have any more questions

seand7565:
I believe this issue has been solved by #3163

loicginoux:
Thanks for the PR ! 👏

PR #3163, opened 2019-04-05
Disable adjust stock field when user does not have the correct permission

seand7565:
**Description**
Currently, if you only have display permission for StockItems, you can still see and interact with the stock edit field on the Stock Item page. When you attempt to submit, the API returns 401 unauthorized - however this might be confusing to admin users who don't know about their permission level. 

This PR disables the stock edit field (and "Back Orderable" tick boxes) if you do not have the correct permissions to edit stock items, with an explanation tooltip as to why you can't edit the stock items.

This should solve issue #2241 

**Checklist:**


seand7565:
Seems like two unrelated payment specs failed. I'm guessing as part of #3160 ?

kennyadsl:
@seand7565 Hey there! Can you please rebase against master so we'll see the CI green, thanks!

PR #3165, opened 2019-04-08
Ensure cartons find soft deleted shipping methods

pelargir:
**Description**

Ensure cartons can find soft deleted shipping methods they belong to. If they can't, the `#tracking_url` method blows up when it tries to call `#build_tracking_url` on a nil shipping method.

**Checklist:**

### DIFF
--- a/core/app/models/spree/carton.rb
+++ b/core/app/models/spree/carton.rb
@@ -3,7 +3,7 @@
 class Spree::Carton < Spree::Base
   belongs_to :address, class_name: 'Spree::Address', optional: true
   belongs_to :stock_location, class_name: 'Spree::StockLocation', inverse_of: :cartons, optional: true
-  belongs_to :shipping_method, class_name: 'Spree::ShippingMethod', inverse_of: :cartons, optional: true
+  belongs_to :shipping_method, -> { with_discarded }, class_name: 'Spree::ShippingMethod', inverse_of: :cartons, optional: true
 
   has_many :inventory_units, class_name: "Spree::InventoryUnit", inverse_of: :carton, dependent: :nullify
   has_many :orders, -> { distinct }, through: :inventory_units
--- a/core/spec/models/spree/carton_spec.rb
+++ b/core/spec/models/spree/carton_spec.rb
@@ -5,6 +5,14 @@ require 'rails_helper'
 RSpec.describe Spree::Carton do
   let(:carton) { create(:carton) }
 
+  describe 'shipping method' do
+    it 'returns soft deleted shipping method' do
+      carton = create(:carton)
+      carton.shipping_method.discard
+      expect(carton.reload.shipping_method).to be_present
+    end
+  end
+
   describe "#create" do
     subject { carton }
 


kennyadsl:
I'm not sure we should do this. Every carton will now also return discarded records as well by default but I still can't figure out how to make this kind of things work with discard.

jacobherrington:
@kennyadsl that was my concern, but I'm not sure if that is bad or acceptable.

jarednorman:
Currently shipping methods have both paranoia and discard on them (both using the `deleted_at` column.)

I think a better solution might to be remove paranoia from the model, making the extra scope unnecessary. I'm not 100% on what else this might affect, though.

jacobherrington:
Oops @tvdeyen I dismissed your review. 🤦‍♂️

pelargir:
Rebased from master to resolve conflict.

ericsaupe:
Ideally it would only show the soft deleted `shipping_method` if it's already attached but not list it if choosing another, right?

kennyadsl:
Yes, but my concern is if we are using (in core, extensions or users' stores) some code to check if there's a shipping method attached to the carton. 

```ruby
if carton.shipping_method
  # do something dangerous
end
```

The current behavior is returning `false` if the shipping method has been soft-deleted, if we merge this PR it would return `true`. I'm not sure if this is actually critical and we should considered this to be a breaking change. We have a [single occurrence](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/orders/_carton.html.erb#L33) of this in core though. 🤔

In terms of functionality, do we really want to still build tracking urls for deleted shipping methods? Maybe not generating the tracking at all is the best thing as default behavior here, so probably a check of the shipping method presence in the `tracking_url` method is enough to fix the reported bug?

pelargir:
@kennyadsl I see your point. If a carton's shipping method was soft deleted, wouldn't you want to still display it as part of the historical record for that carton? If it's important to show when a carton's shipping method has been deleted, the shipping method itself could be checked for its deletion status and an appropriate message displayed. I'm having trouble imagining a scenario where a dangerous operation would be performed based on the presence or absence of a carton's shipping method.

core/app/models/spree/carton.rb
```diff
@@ -3,7 +3,7 @@
 class Spree::Carton < Spree::Base
   belongs_to :address, class_name: 'Spree::Address', optional: true
   belongs_to :stock_location, class_name: 'Spree::StockLocation', inverse_of: :cartons, optional: true
-  belongs_to :shipping_method, class_name: 'Spree::ShippingMethod', inverse_of: :cartons, optional: true
+  belongs_to :shipping_method, -> { with_deleted }, class_name: 'Spree::ShippingMethod', inverse_of: :cartons, optional: true
```
kennyadsl: I think we should use `with_discarded` instead

kennyadsl:
@pelargir I'm sure it's the right thing to do and I'm also having trouble imagining a scenario when this could cause issues, maybe just some recurring task that makes sure that all cartons have a shipping method associated for consistency but... :man_shrugging: 

Let me talk with the rest of the core team during the next meeting about this and I'll let you know.  I left a comment in the meantime, we should use discard scopes rather than paranoia ones.

core/app/models/spree/carton.rb
pelargir: Done. I force-pushed this change.

pelargir:
@kennyadsl I've been attempting to fix the CircleCI build but I keep running into the error outlined in #3374.

pelargir:
@kennyadsl the build is passing now. What else can I do to help get this approved?

spaghetticode:
Also, not strictly related to the issue with `#tracking_url` but I think it helps in getting more context, we have a validation on `Spree::Carton` that requires the presence of a `shipping_method`, so all cartons with a soft-deleted shipping method are invalid.

kennyadsl:
@solidusio/core-team can we have a second review here?

kennyadsl:
Thanks, @pelargir!

Issue #2961, opened 2018-11-20
The navigation menu should be a collapsible


jacobherrington:
This is a continuation of #666 based on feedback from an MVP attempt at this functionality in #2957 and the earlier work done in #1718.

One of the more pressing issues with Solidus' admin UI is a lack of responsiveness. A key step in achieving a more responsive, modern feeling UI is recovering some lost screen real estate by reworking the navigation menu.

Ideally, the navigation menu could be collapsed much like the navigation menu in GitLab and Spree:

<img width="207" alt="screen shot 2018-11-20 at 4 27 39 pm" src="https://user-images.githubusercontent.com/11466782/48806843-49843d00-ece1-11e8-9ecc-6a04a33ad018.png">

## Some of the requirements laid down in previous issues and discussions:

- [ ] Default open on large screens.
- [ ] Default close on small screens.
- [ ] Remember user preferences on the computer/device they're using.
- [ ] Stays open/close until user changes the preference.
- [ ]  Label on collapsed nav items to appear on hover. Tooltip should look consistent with admin, but different from menu styles.

## An example mockup:

![image](https://user-images.githubusercontent.com/11466782/48807548-79344480-ece3-11e8-9e13-520f3780440a.png)

jacobherrington:
I am currently working on this, but would love to take suggestions here!

tvdeyen:
I have a branch with a working prototype as well. Just tell me if I should send it to you.

tvdeyen:
@jacobherrington https://github.com/tvdeyen/solidus/tree/collapsible-nav

seand7565:
@jacobherrington If you're not still working on this I'd love to take a crack at finishing it - this is a feature I'd love to see in the next release.

jacobherrington:
@seand7565 Please do, I haven't had much free time in the last few months. 👍

tvdeyen:
@seand7565 this would be great 👍

seand7565:
Great! I've started working on it here -> https://github.com/seand7565/solidus/tree/nav-collapse

jacobherrington:
@seand7565 Please let me know if I can help, I don't have much time, but happy to give feedback. Thanks Sean! 🥇

jacobherrington:
https://github.com/solidusio/solidus/pull/3322 Work is being done in this PR!

PR #3166, opened 2019-04-09
Admin-side navbar collapse

seand7565:
**Description**
Makes the admin-side navbar collapsible by clicking a collapse icon at the top of the page.

navbar-collapse

Additional functionality:

- The navbar defaults to closed on iPad sized screens or smaller.
- When the navbar is collapsed or opened, a cookie is stored that remembers your preference - on the next page load, the navbar will be in the same state as it was on the previous page.
- When the navbar is closed, a tooltip is set for the navbar icons which - on hover- displays the name of the button you have selected. If you prefer a closed navbar, but don't know what the icons mean, you can still find your way around.

image

@jacobherrington and @tvdeyen did most of the heavy lifting here, I just cleaned it up and added a few things.

ref #2961 - should close this issue if everything looks good!

**Checklist:**

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/components/admin_nav.js
+++ b/backend/app/assets/javascripts/spree/backend/components/admin_nav.js
@@ -1,9 +1,9 @@
 Spree.ready(function() {
-  toggleTooltips()
-  if(window.screen.width <= 1024 && !document.cookie.includes("admin_nav_hidden")){
-    //Set default nav to collapse on small screens - but don't override user preference
+  toggleTooltips();
+  if (window.screen.width <= 1024 && !document.cookie.includes("admin_nav_hidden")) {
+    // Set default nav to collapse on small screens - but don't override user preference
     document.body.classList.add("admin-nav-hidden");
-    document.cookie = "admin_nav_hidden=true; expires=Fri, 31 Dec 9999 23:59:59 GMT"
+    document.cookie = "admin_nav_hidden=true; expires=Fri, 31 Dec 9999 23:59:59 GMT";
   }
 
   var adminNavToggle = document.querySelector("#admin-nav-toggle");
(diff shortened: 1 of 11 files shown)


seand7565:
Got some tests to fix here - capybara is trying to click on nested navbar links, but those are hidden when the navbar is collapsed. The navbar defaults to closed on small screens, which is the problem for these tests. 

The two solutions I have are to increase capybaras window size, or just click the navbar-toggle button to open the navbar before running the tests. I'll pop something in soon to get the tests to pass.

tvdeyen:
Thanks, this looks great! Regarding the specs: My vote is to raise the capybara window size.

aitbw:
My vote too goes to raise Capybara window size. This looks very good, @seand7565! Thanks a lot :raised_hands:

One small thing: is it possible you could rebase this PR? That'd would help us avoid the merge commit in here

seand7565:
(Mostly) passing now! @aitbw The screen size actually wasn't the issue here like I thought, I used the wrong operator in the javascript file determining the default screen size (greater than rather than less than). D'oh!

Not sure why the mysql rails 5.1 tests failed - and on the flash too, which wasn't touched in this commit. I'll see if I can find out why it failed!

Thanks all for the feedback! ❤️ I rebased and squashed some of the smaller changes - not sure how far you folks like PRs squashed. Should this all be one commit?

seand7565:
The test that was previously failing here appears to be passing now after a rebase 🤷‍♂

spaghetticode:
@seand7565 Thanks for the PR 👏 

I see it now needs some rebasing, some asset files are out-of-sync, can you please update them? Thank you!

seand7565:
This doesn't play well with the new admin styling - will have to make some changes to get it back in line. Give me a few days and I'll have it done!

seand7565:
Should be good to go now!

image

image

backend/app/assets/javascripts/spree/backend/components/admin_nav.js
```diff
@@ -0,0 +1,27 @@
+Spree.ready(function() {
+  toggleTooltips();
+  if (window.screen.width <= 1024 && !document.cookie.includes("admin_nav_hidden")) {
+    // Set default nav to collapse on small screens - but don't override user preference
+    document.body.classList.add("admin-nav-hidden");
+    document.cookie = "admin_nav_hidden=true; expires=Fri, 31 Dec 9999 23:59:59 GMT";
```
elia: Any reason not to use LocalStorage instead? 
Browser support is pretty wide https://caniuse.com/#feat=namevalue-storage
I would also suggest namespacing the key with `solidus.` or `solidus/`, e.g. `window.localStorage.get('solidus.minimize-admin-nav')`

jacobherrington: I believe I was initially using LocalStorage, but there was some feedback on that decision from either @tvdeyen or @seand7565. I do not remember it 🤦‍♂

kennyadsl:
I'm ok with this but I think it will need some extra design iterations after the merge. If you agree I'd merge it after v2.9 release which already contains a lot of backend changes.

mfrecchiami:
Hi @seand7565 and everyone!

I tried to improve a little bit more this new feature, moving the collapse icon inside the `admin-nav-header` and leaving the Solidus mark always visible.

minimize-menu

minimize-menu-closed

Then I tried to move to the below the collapse button:

minimize-menu-bottom

minimize-menu-bottom-closed

**Which one do you think that could be the best solution?**

**Note:** when you log in with a user whit role of Admin, there is a select for change the language, at the bottom of the sidebar; as you can see by the screenshots above, I choose to hide it when the navbar is collapsed

seand7565:
That looks fantastic - thanks @mfrecchiami ! I like the button being inside the collapsible panel, so it doesn't affect the layout of the rest of the page.

mfrecchiami:
Hi @seand7565 !

I did my job starting from your repo 💪 
I've made my branch, did the rebase with `master` and added my commits
I believed to find your repo in the _compare_ step, but it wasn't 😞 (still don't know why)
So I made another PR related to yours: https://github.com/solidusio/solidus/pull/3322

Do you want to do it in a different way? Let me know!

seand7565:
@mfrecchiami No, that's perfect - thanks! I'll close this PR in favor of yours.

PR #3167, opened 2019-04-09
Add basic tooling for JS linting

### LINKED ISSUE #2849: Including JS linting in the Solidus workflow
After making #2847 and hearing @tvdeyen's feedback, I feel like we need to adopt some JS standards for Solidus. 

Solidus isn't a JS project; it doesn't seem like this is something we should obsess over. However the project relies on having a functioning admin UI, so it's pragmatic to make sure we have high code quality when it comes to JS.

I'm not sure what the best solution for Solidus is (and don't want to start an argument), but I think it's worth discussing with other contributors. Thoughts?

aitbw:
**Description**

See #2849 for further discussion prior to this PR

This PR aims to add basic tooling for JS linting based on @airbnb's [JavaScript Style Guide](https://github.com/airbnb/javascript) by doing the following:

* Adds a `package.json` file with the bare minimum to work as expected —including a Yarn task to automatically check if the rules for all JS files on Solidus (minus those under `guides`, `sandbox`, `node_modules` and `vendor`) are being followed. The dependencies for this to work are based on [this NPM package](https://www.npmjs.com/package/eslint-config-airbnb-base), which doesn't include linting specific to JS frameworks, such as React

* Ignores unnecessary newly-introduced files and folders (I'm looking at you, `node_modules` :eyes:)

* Adds ESLint config file with all the warnings/errors raised when running `yarn lint` disabled

* Adds ESLint inline config comments on several files under `backend`, as `yarn lint` was complaining even with the config file

* Enable JS linting for Hound based on the newly-introduced `.eslintrc.json` file

* Whitelist Solidus-specific global variables

This should allow for other future PRs to re-enable rules and fix its related issues.

Let me know what you think! :raised_hands:

**Checklist:**

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,27 @@
+{
+  "extends": "airbnb-base",
+  "rules": {
+    "camelcase": 0,
+    "comma-dangle": 0,
+    "dot-notation": 0,
+    "func-names": 0,
+    "no-console": 0,
+    "no-extra-semi": 0,
+    "no-global-assign": 0,
+    "no-redeclare": 0,
+    "no-shadow": 0,
+    "no-undef": 0,
+    "no-unused-vars": 0,
+    "no-use-before-define": 0,
+    "no-var": 0,
+    "object-shorthand": 0,
+    "operator-linebreak": 0,
+    "prefer-arrow-callback": 0,
+    "prefer-destructuring": 0,
+    "prefer-template": 0,
+    "quotes": 0,
+    "radix": 0,
+    "space-before-function-paren": 0,
+    "vars-on-top": 0
+  }
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -7,10 +7,12 @@
     "func-names": 0,
     "no-console": 0,
     "no-extra-semi": 0,
-    "no-global-assign": 0,
+    "no-global-assign": ["error", {
+        "exceptions": ["Tabs"]
+    }],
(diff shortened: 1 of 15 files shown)


kennyadsl:
But I think it should not close #2849, since it also expects other PR to re-introduce disabled rules.

aitbw:
Correct, @kennyadsl —let me edit the description to avoid automatic closure of that issue.

.eslintrc.json
```diff
@@ -0,0 +1,27 @@
+{
+  "extends": "airbnb-base",
+  "rules": {
+    "camelcase": 0,
+    "comma-dangle": 0,
+    "dot-notation": 0,
+    "func-names": 0,
+    "no-console": 0,
+    "no-extra-semi": 0,
+    "no-global-assign": 0,
```
tvdeyen: Instead of disabling the rule we could also whitelist all our global constants. So that we don’t allow new ones.

.eslintrc.json
```diff
@@ -0,0 +1,27 @@
...
+    "no-redeclare": 0,
+    "no-shadow": 0,
+    "no-undef": 0,
```
tvdeyen: Also with this rule. As we have at least the global `Spree` object we could whitelist it here and keep the rule enabled.

aitbw:
Added a new commit to comply with your suggestions, @tvdeyen :raised_hands: let me know what you think!

PR #3168, opened 2019-04-10
Use taxon children when searching classification

fkoessler:
**Description**

In current implementation, `eligible?` checks the configured taxons and all descendants and `actionable?` only seems to check against the taxons themselves. This leads to buggy behavior. See: https://github.com/solidusio/solidus/pull/1409#issuecomment-243260852

This commit modifies `actionable?` and have it check against configured taxons and all descendants in order to have a coherent behavior.

### DIFF
--- a/core/spec/models/spree/promotion/rules/taxon_spec.rb
+++ b/core/spec/models/spree/promotion/rules/taxon_spec.rb
@@ -151,14 +151,14 @@ RSpec.describe Spree::Promotion::Rules::Taxon, type: :model do
     let(:order) { create :order_with_line_items }
     let(:taxon) { create :taxon, name: 'first' }
 
-    before do
-      rule.preferred_match_policy = 'invalid'
-      rule.save!(validate: false)
-      line_item.product.taxons << taxon
-      rule.taxons << taxon
-    end
-
     context 'with an invalid match policy' do
+      before do
+        rule.preferred_match_policy = 'invalid'
+        rule.save!(validate: false)
+        line_item.product.taxons << taxon
+        rule.taxons << taxon
+      end
+
       it 'logs a warning and uses "any" policy' do
         expect(Spree::Deprecation).to(
           receive(:warn).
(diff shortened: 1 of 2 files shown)


fkoessler:
Hello @aitbw , thanks for reviewing. I added some specs that cover the change. Let me know if I can improve the PR.

core/spec/models/spree/promotion/rules/taxon_spec.rb
```diff
@@ -170,5 +170,42 @@
...
     end
+
+    context 'when a product has the taxon of a taxon rule' do
+      before do
+        product.taxons << taxon
+        rule.taxons << taxon
+        rule.save!
+      end
+
+      it 'is actionnable' do
```
aitbw: Typo, should be `actionable`

core/spec/models/spree/promotion/rules/taxon_spec.rb
```diff
@@ -170,5 +170,42 @@
...
+
+    context 'when a product has a taxon child of a taxon rule' do
+      before do
+        taxon.children << taxon2
+        product.taxons << taxon2
+        rule.taxons << taxon
+        rule.save!
+      end
+
+      it 'is actionnable' do
```
aitbw: Typo, should be `actionable`

core/spec/models/spree/promotion/rules/taxon_spec.rb
```diff
@@ -170,5 +170,42 @@
...
+    end
+
+    context 'when a product does not have taxon or child taxon of a taxon rule' do
+      before do
+        product.taxons << taxon2
+        rule.taxons << taxon
+        rule.save!
+      end
+
+      it 'is not actionnable' do
```
aitbw: Typo, should be `actionable`

core/spec/models/spree/promotion/rules/taxon_spec.rb
```diff
@@ -170,5 +170,42 @@
         ).to be_truthy
       end
     end
+
+    context 'when a product has the taxon of a taxon rule' do
```
aitbw: This should be `has a taxon`, right?

fkoessler: that's right, fixed.

fkoessler:
There will be a performance overhead since `actionable?` now calls to `#rule_taxon_ids_with_children`. I'm not sure if it is significant or not.

`actionable?`  will be called on each line_item of the order, see `Spree::Promotion` `#line_item_actionable?`
The performance overhead will probably depend on the depth of the taxonomy, the number of line_items in the order. Should this be benchmarked?
Performance improvements from https://github.com/solidusio/solidus/pull/2258 are kept.

In any case, the taxon promotion rule won't work as expected until `eligible?` and `actionable?` are made coherent.

kennyadsl:
I'm fine with a performance regression if we are making things work as expected. The overhead will probably be higher depending on how many nested taxons you have and if the promotion is auto-applied (without the promo code), but in that case, I'd left to each store to fix eventual performance issues on their own. Thanks for the accurate explanation.

PR #3170, opened 2019-04-10
Remove unused variable assignment from Admin::OrdersController#index action

aitbw:
**Description**

This variable assignment was introduced back in commit 1a38331; since it is now unused and does not affect the code in any way, it can be safely removed

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/orders_controller.rb
+++ b/backend/app/controllers/spree/admin/orders_controller.rb
@@ -14,7 +14,6 @@ module Spree
       respond_to :html
 
       def index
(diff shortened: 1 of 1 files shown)


kennyadsl:
I think params here are restored so they can be presented to the user in the view again. If they are left empty user would lose the date that he/she entered before submitting, right?

aitbw:
@kennyadsl oh snap, you're absolutely right! Sorry about that :bowing_woman:

There's still this, that I think we can safely remove: `query_present = params[:q]`

PR #3172, opened 2019-04-11
Improving the way positioning update happens in classifications

softr8:
**Description**
It used to rely on item position coming from a html index, but
if a product was deleted, its classification was destroyed as well
leaving a gap, so classification positions are no longer consecutive
breaking the dependency of the index in the html, this uses internal
index instead of the one coming from the frontend

The issue:
consider the following classifications
[product_id, taxon_id, position]
1, 1, 1
3, 1, 3
4, 1, 4
5, 1, 5
6, 1, 6

And we wanted to move product_id 6 to before product_id 5, the javascript code sends position 4, so the code updated the product 6 before the product 4 instead of the product 5, this change fixes this issue.

The reason of the position gaps is because product has dependent: :delete_all in classifications but it does not rebuild the position index to consecutive values.

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/classifications_controller.rb
+++ b/api/app/controllers/spree/api/classifications_controller.rb
@@ -6,14 +6,15 @@ module Spree
       def update
         authorize! :update, Product
         authorize! :update, Taxon
-        classification = Spree::Classification.find_by(
-          product_id: params[:product_id],
-          taxon_id: params[:taxon_id]
-        )
-        # Because position we get back is 0-indexed.
-        # acts_as_list is 1-indexed.
-        classification.insert_at(params[:position].to_i + 1)
-        head :ok
+
+        taxon = Spree::Taxon.find(params[:taxon_id])
+        classification = taxon.classifications.find_by(product_id: params[:product_id])
+
+        if classification.insert_at(params[:position].to_i)
+          head :ok
+        else
+          render json: { errors: classification.errors.full_messages }, status: :unprocessable_entity
+        end
       end
     end
   end
--- a/api/spec/requests/spree/api/classifications_controller_spec.rb
+++ b/api/spec/requests/spree/api/classifications_controller_spec.rb
@@ -20,7 +20,7 @@ module Spree
 
     context "as a user" do
       it "cannot change the order of a product" do
-        put spree.api_classifications_path, params: { taxon_id: taxon, product_id: taxon.products.first, position: 1 }
+        put spree.api_classifications_path, params: { taxon_id: taxon.id, product_id: taxon.products.first.id, position: 1 }
         expect(response.status).to eq(401)
       end
     end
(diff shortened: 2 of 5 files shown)


api/spec/requests/spree/api/classifications_controller_spec.rb
```diff
@@ -30,14 +30,24 @@ module Spree
...
 
+      it "can change the order of a product regardless gaps in positions due discarted products" do
+        taxon.classifications.reload
+        taxon.classifications.second.update_attribute(:position, 3)
+        taxon.classifications.third.update_attribute(:position, 5)
+        classification = taxon.classifications.find_by(product_id: last_product.id)
+        put spree.api_classifications_path, params: { taxon_id: taxon.id, product_id: last_product.id, position: 1 }
+        expect(response.status).to eq(200)
+        expect(classification.reload.position).to eq(3)
+      end
```
kennyadsl: I think something should be fixed here. I change this code (which I think it's legit) to make a test and it raises an exception:

```
        taxon.classifications.second.update_attribute(:position, 3)
        taxon.classifications.third.update_attribute(:position, 4)
        classification = taxon.classifications.find_by(product_id: last_product.id)
        put spree.api_classifications_path, params: { taxon_id: taxon.id, product_id: last_product.id, position: 3 }
        expect(response.status).to eq(200)
```

Can you please give it a try as well? 

I think that with your solution there's the possiblity that the array could be smaller than the position passed via params.

softr8: Uhm, the js plugin to reorder the products works from 0 to size - 1, the product used to test has 3 taxons, so, a valid position should be from 0 to 2, isn't it?

api/spec/requests/spree/api/classifications_controller_spec.rb
```diff
@@ -30,14 +30,24 @@ module Spree
...
-      it "can change the order a product" do
+      it "can change the order of a product" do
         classification = taxon.classifications.find_by(product_id: last_product.id)
         expect(classification.position).to eq(3)
         put spree.api_classifications_path, params: { taxon_id: taxon.id, product_id: last_product.id, position: 0 }
         expect(response.status).to eq(200)
         expect(classification.reload.position).to eq(1)
       end
 
+      it "can change the order of a product regardless gaps in positions due discarted products" do
```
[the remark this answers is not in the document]
softr8: thanks! fixing

api/app/controllers/spree/api/classifications_controller.rb
```diff
@@ -10,9 +10,11 @@ def update
           product_id: params[:product_id],
           taxon_id: params[:taxon_id]
         )
-        # Because position we get back is 0-indexed.
-        # acts_as_list is 1-indexed.
-        classification.insert_at(params[:position].to_i + 1)
+        # We can not rely on html element position index, there could be gaps between
+        # positions due discarted products, their classification is destroyed
```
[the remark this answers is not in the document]
softr8: Thanks! fixing

api/spec/requests/spree/api/classifications_controller_spec.rb
kennyadsl: Ah ok, so it's not possible for the plugin to make a request with position > "number of non-discarded elements", which would raise the exception I had, because `params[:position].to_i` is not an index of that array. 

Makes sense, thanks for the clarification!

kennyadsl: Maybe we should still make an extra check to avoid this because the request could come from any other place, even not the JS plugin?

softr8: yes, I agree on that, what do you suggest to return? like a unprocessable_entity with the error in json ?

kennyadsl: yes, I think that would make sense, thanks!

softr8: Done! @kennyadsl

api/spec/requests/spree/api/classifications_controller_spec.rb
```diff
@@ -30,21 +30,43 @@ module Spree
...
         expect(classification.position).to eq(3)
         put spree.api_classifications_path, params: { taxon_id: taxon.id, product_id: last_product.id, position: 0 }
         expect(response.status).to eq(200)
         expect(classification.reload.position).to eq(1)
       end
 
+      it "can change the order of a product regardless gaps in positions due discarded products" do
+        taxon.classifications.reload
+        taxon.classifications.second.update_attribute(:position, 3)
+        taxon.classifications.third.update_attribute(:position, 5)
```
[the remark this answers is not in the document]
tvdeyen: We need to disable this cop as this is ridiculous. We update columns all over the place and do this for good reasons (AR callbacks are evil is one of them)

kennyadsl:
We discussed in real life yesterday about trying to move this logic into a Classification method or a different object since that action starting to become too complex.

softr8:
Closing this PR in favor of #3439

PR #3173, opened 2019-04-11
Update class methods to be instance methods.

jacquesporveau:
**Description**
There are no class methods defined on `Spree::Order` with the names the docs suggest. These are supposed to be instance methods.

### DIFF
--- a/guides/source/developers/adjustments/overview.html.md
+++ b/guides/source/developers/adjustments/overview.html.md
@@ -137,20 +137,20 @@ If you want to retrieve the line item adjustments, you can use the
 `line_item_adjustments` method:
 
 ```ruby
-Spree::Order.line_item_adjustments
+Spree::Order.find(1).line_item_adjustments
 ```
 
 Or, if you want to retrieve the shipment adjustments, you can use the
 `shipment_adjustments` method:
 
 ```ruby
-Spree::Order.shipment_adjustments
+Spree::Order.find(1).shipment_adjustments
 ```
 
 Finally, if you want to retrieve all of the adjustments on the order, you can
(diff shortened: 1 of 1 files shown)


guides/source/developers/adjustments/overview.html.md
```diff
@@ -137,20 +137,20 @@ If you want to retrieve the line item adjustments, you can use the
 `line_item_adjustments` method:
 
 ```ruby
-Spree::Order.line_item_adjustments
+Spree::Order.find(1).line_item_adjustments
```
kennyadsl: What about using `order.line_item_adjustments. I think it should be pretty clear that it refers to an instance of the `Spree::Order` class. WDYT?

jacquesporveau: Going by the how the rest of the file is formatted I think we should take this route. 

The adjustment scopes section is all formatted like this.

```ruby
- `Spree::Order.find(1).adjustments.eligible`: Returns all of the eligible
  adjustments on the order with ID `1`.
- `Spree::LineItem.find(1).adjustments.eligible`: Returns all of the eligible
  adjustments on the line item with ID `1`.
- `Spree::Shipment.find(1).adjustments.eligible`: Returns all of the eligible
  adjustments on the shipment with ID `1`.
```

as well as part of the other adjustments docs.

```ruby
Spree::Order.find(1).adjustments
```

That makes me imagine that this is probably a pattern found throughout these docs and for the sake of consistency we should follow it.

kennyadsl: Yes, I think we should be consistent as well, thanks for pointing that out!

PR #3174, opened 2019-04-12
Upgrade to CanCanCan 3.0

coorasse:
This PR aims to upgrade to cancancan 3.0 without the assle of rewriting all the permissions logic. see https://github.com/solidusio/solidus/pull/3148

It basically overrides the changes to the `merge` method of Ability.rb which were breaking solidus permissions mechanism. 
It should be seen as a temporary patch, so that the "proper" way of managing permissions can be released in the future in a major 3.0 release.
This allows to migrate to Rails 6.0.

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'activemerchant', '~> 1.66'
   s.add_dependency 'acts_as_list', '~> 0.3'
   s.add_dependency 'awesome_nested_set', '~> 3.0', '>= 3.0.1'
-  s.add_dependency 'cancancan', '~> 2.2'
+  s.add_dependency 'cancancan', '~> 3.0'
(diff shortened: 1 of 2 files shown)


core/solidus_core.gemspec
```diff
@@ -30,7 +30,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'activemerchant', '~> 1.66'
   s.add_dependency 'acts_as_list', '~> 0.3'
   s.add_dependency 'awesome_nested_set', '~> 3.0', '>= 3.0.1'
-  s.add_dependency 'cancancan', '~> 2.2'
+  s.add_dependency 'cancancan', '~> 3.0'
```
tvdeyen: It seams we are still able to support cancancan 2.x as well? We should not force everybody else to upgrade to 3.0 as well. Could we set a lower bound of `>= 2.2` and upper bound of `< 4.0`?

aldesantis:
Closing because it doesn't look like we need this anymore. @coorasse feel free to correct me if I'm wrong!

PR #3180, opened 2019-04-15
Remove name column from default ransackable attributes

### LINKED ISSUE #3179: Several models have invalid "name" included in ransackable_attributes
I stumbled upon a Ransack misconfiguration: https://github.com/solidusio/solidus/blob/4d53bdc921a2738f301eb398ab6a4963cb1cdf66/core/app/models/concerns/spree/ransackable_attributes.rb#L10
`name` is included by default in all the models, even the ones which don't have a `name` column, such as `Spree::Variant`, causing Ransack searches like this:

```ruby
Spree::Variant.ransack('name_cont' => 'asd', 's' => 'id desc').result.to_sql
```

fail with the following error:

```ruby
> bundle exec rails c
Loading development environment (Rails 5.2.3)
[1] pry(main)> Spree::Variant.ransack('name_cont' => 'asd', 's' => 'id desc').result.to_sql
NoMethodError: undefined method `type' for nil:NilClass
from gems/ransack-2.1.1/lib/ransack/adapters/active_record/context.rb:28:in `type_for'
```

Instead of discarding the `name` matcher.

**Solidus Version**: `2.9.0.alpha@4d53bdc92`

**To Reproduce**

```
cd sandbox
bundle exec rails r "puts Spree::Variant.ransack('name_cont' => 'asd', 's' => 'id desc').result.to_sql"
```

**Current behavior**

```
undefined method `type' for nil:NilClass
```

**Expected behavior**

```
SELECT "spree_variants".* FROM "spree_variants" WHERE "spree_variants"."deleted_at" IS NULL ORDER BY "spree_variants"."id" DESC
```

mdesantis:
Fixes #3179. `name` column included by default within default ransackable attributes causes ransack searches with search matchers based on `name` column to crash when appied on models which don’t have a `name` column. This PR removes the `name` column from default ransackable attributes and adds it explicitly on models which support search by name.

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/ransackable_attributes.rb
+++ b/core/app/models/concerns/spree/ransackable_attributes.rb
@@ -7,7 +7,7 @@ module Spree::RansackableAttributes
     class_attribute :whitelisted_ransackable_attributes
 
     class_attribute :default_ransackable_attributes
-    self.default_ransackable_attributes = %w[id name]
+    self.default_ransackable_attributes = %w[id]
   end
 
   class_methods do
--- a/core/app/models/spree/country.rb
+++ b/core/app/models/spree/country.rb
@@ -8,6 +8,8 @@ module Spree
 
     validates :name, :iso_name, presence: true
 
+    self.whitelisted_ransackable_attributes = %w[name]
+
     def self.default
       if Spree::Config.default_country_id
         Spree::Deprecation.warn("Setting your default country via its ID is deprecated. Please set your default country via the `default_country_iso` setting.", caller)
--- a/core/app/models/spree/option_type.rb
+++ b/core/app/models/spree/option_type.rb
@@ -25,6 +25,8 @@ module Spree
     after_touch :touch_all_products
     after_save :touch_all_products
 
+    self.whitelisted_ransackable_attributes = %w[name]
+
     def touch_all_products
       products.find_each(&:touch)
     end
--- a/core/app/models/spree/option_value.rb
+++ b/core/app/models/spree/option_value.rb
@@ -16,7 +16,7 @@ module Spree
 
     delegate :name, :presentation, to: :option_type, prefix: :option_type
 
-    self.whitelisted_ransackable_attributes = ['presentation']
+    self.whitelisted_ransackable_attributes = %w[name presentation]
 
     # Updates the updated_at column on all the variants associated with this
     # option value.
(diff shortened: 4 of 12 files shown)


mdesantis:
@kennyadsl do you think this is an important change? because previous implementation could expect `default_ransack_attributes` to include `name`? I was thinking about that

kennyadsl:
`Important` label is needed to mark PRs that need special attention in the Changelog. I think this could potentially break user application if they were relying on having that `name` attribute in the ranskack attributes, so they have to be aware that, in case they really need it, they have to re-add it manually, right?

mdesantis:
@kennyadsl , in this PR I fixed the tested models, but if we want to be :100:% sure that the functionality remains the same for the remaining models we could add `name` to their `whitelisted_ransackable_attributes` too. Here is the models list which have a `name` column but doesn't have `name` within `whitelisted_ransackable_attributes`:

```ruby
Spree::AdjustmentReason
Spree::BillingIntegration
Spree::CreditCard
Spree::Gateway
Spree::Gateway::Bogus
Spree::Gateway::BogusSimple
Spree::PaymentMethod
Spree::PaymentMethod::BogusCreditCard
Spree::PaymentMethod::Check
Spree::PaymentMethod::CreditCard
Spree::PaymentMethod::SimpleBogusCreditCard
Spree::PaymentMethod::StoreCredit
Spree::PromotionCategory
Spree::RefundReason
Spree::ReimbursementType
Spree::ReimbursementType::Credit
Spree::ReimbursementType::Exchange
Spree::ReimbursementType::OriginalPayment
Spree::ReimbursementType::StoreCredit
Spree::ReturnReason
Spree::Role
Spree::ShippingCategory
Spree::ShippingMethod
Spree::StateChange
Spree::Store
Spree::StoreCreditCategory
Spree::StoreCreditReason
Spree::StoreCreditType
Spree::TaxCategory
Spree::TaxRate
```

I extracted this list with the following script:

```ruby
ActiveRecord::Base.descendants.select do |model|
  begin
    model.respond_to?(:whitelisted_ransackable_attributes) && 
    model.columns.any? { |column| column.name == 'name' } && (
      !model.whitelisted_ransackable_attributes ||
      !model.whitelisted_ransackable_attributes.include?('name')
    )
  rescue ActiveRecord::StatementInvalid; false
  end
end.map(&:to_s).sort
```

PR #3181, opened 2019-04-16
Introduce deprecation warning when using `Spree::Refund#perform!` callback

aitbw:
**Description**

This PR introduces a deprecation warning when relying on `Spree::Refund#perform!` implicitly (i.e.: calling it as a callback rather than calling it explicitly); aims to serve as a starting point for #1415

**Checklist:**

### DIFF
--- a/backend/spec/features/admin/orders/new_refund_spec.rb
+++ b/backend/spec/features/admin/orders/new_refund_spec.rb
@@ -12,12 +12,16 @@ RSpec.describe 'New Refund creation', :js do
 
   it 'creates a new refund' do
     visit spree.new_admin_order_payment_refund_path(order, payment)
+
     expect(page).not_to have_selector 'td', text: amount
+    expect(Spree::Deprecation).to receive(:warn)
+
     within '.new_refund' do
       fill_in 'refund_amount', with: amount
       select reason.name, from: 'Reason'
       click_button 'Refund'
     end
+
     expect(page).to have_content 'Refund has been successfully created!'
     expect(page).to have_selector 'td', text: amount
   end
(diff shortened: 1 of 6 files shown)


kennyadsl:
Thanks for this change! I think we should submit this change with #1415 commits as well since we would otherwise ship a deprecation warning without the code that does the "new thing". WDYT?

aitbw:
@kennyadsl, sounds good to me. Do you want me to rebase that PR and then add these commits or cherry pick them into this branch?

kennyadsl:
I think we can use this PR cherry-picking and close that one, thanks!

aitbw:
@kennyadsl, I noticed something while re-reading your first comment: if we include the changes proposed in #1415 alongside this PR, we'd render this warning useless as the callback would be removed **prior** to Solidus v3.0 (which I assume is not the intended) and we would not be giving the developers any previous notice that the `#perform!` callback is going to be removed.

What do you think?

kennyadsl:
I think that what we have to do is:

- call `perform` explicitly after save (no more callbacks), always using the `perform_after_creation=false` attribute
- provide a way to tell users that are saving the refund in their store to:
  - add the `perform_after_creation=false`  attribute before save (or they will see the deprecation)
  - call `.perfrom` manually

In the next version released, we can safely deprecate `perform_after_creation` since we can assume all stores are calling `.perform` manually, and they no longer need to set `perform_after_creation`.

Does it make sense?

aitbw:
@kennyadsl, yes, that makes sense, but my concern is focused on the stores that rely on that callback to be performed automatically —would that change break their stores if we don't give them a previous warning?

kennyadsl:
If they don't set `perform_after_creation=false` the default is true and they should have the `perform` called, isn't it?

aitbw:
No, the `perform_after_creation` only serves to show the deprecation warning; the `#perform!` callback is completely removed from #1415

kennyadsl:
Closed in favor of https://github.com/solidusio/solidus/pull/3641

PR #3182, opened 2019-04-16
Use classes alongside data-hook attributes for gateway partial

aitbw:
**Description**

Given that other fields in the aforementioned partial have `data-hook` attributes and classes to allow easy customization, I think it could be good that card-related fields have the same structure

Rebased version of #2212, all the credit goes to @ccarruitero 

**Checklist:**

### DIFF
--- a/frontend/app/views/spree/checkout/payment/_gateway.html.erb
+++ b/frontend/app/views/spree/checkout/payment/_gateway.html.erb
@@ -1,14 +1,15 @@
 <%= image_tag 'credit_cards/credit_card.gif', id: 'credit-card-image' %>
 <% param_prefix = "payment_source[#{payment_method.id}]" %>
 
-<div class="field field-required">
+<div class="field field-required card_name" data-hook="card_name">
   <%= label_tag "name_on_card_#{payment_method.id}", t('spree.name_on_card') %>
-  <%= text_field_tag "#{param_prefix}[name]", "#{@order.billing_firstname} #{@order.billing_lastname}", { id: "name_on_card_#{payment_method.id}", autocomplete: "cc-name" } %>
+  <%= text_field_tag "#{param_prefix}[name]", "#{@order.billing_firstname} #{@order.billing_lastname}", { id: "name_on_card_#{payment_method.id}", autocomplete: "cc-name", class: 'cardName' } %>
 </div>
 
-<div class="field field-required" data-hook="card_number">
+<div class="field field-required card_number" data-hook="card_number">
   <%= label_tag "card_number", t('spree.card_number') %>
   <%= text_field_tag "#{param_prefix}[number]", '', {id: 'card_number', class: 'required cardNumber', size: 19, maxlength: 19, autocomplete: "cc-number", type: "tel" } %>
+
   <span id="card_type" style="display:none;">
     ( <span id="looks_like" ><%= t('spree.card_type_is') %> <span id="type"></span></span>
       <span id="unrecognized"><%= t('spree.unrecognized_card_type') %></span>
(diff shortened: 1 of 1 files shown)


frontend/app/views/spree/checkout/payment/_gateway.html.erb
```diff
@@ -1,27 +1,28 @@
...
 <% param_prefix = "payment_source[#{payment_method.id}]" %>
 
-<div class="field field-required">
+<div class="field field-required card_name">
   <%= label_tag "name_on_card_#{payment_method.id}", t('spree.name_on_card') %>
-  <%= text_field_tag "#{param_prefix}[name]", "#{@order.billing_firstname} #{@order.billing_lastname}", { id: "name_on_card_#{payment_method.id}", autocomplete: "cc-name" } %>
+  <%= text_field_tag "#{param_prefix}[name]", "#{@order.billing_firstname} #{@order.billing_lastname}", { id: "name_on_card_#{payment_method.id}", autocomplete: "cc-name", class: 'cardName' } %>
 </div>
 
-<div class="field field-required" data-hook="card_number">
```
kennyadsl: I think we should keep both when the data-hook is already there otherwise we'll break stores that rely on those data-hook

aitbw: Fair enough. Do you think we still should add those classes **and** keep the `data-hook`s?

aitbw:
Done @kennyadsl :heavy_check_mark: I don't know if it's necessary to squash the commits and/or change the PR's title, let me know what you think

kennyadsl:
I'd also go for squashing commits, thanks @aitbw!

aitbw:
@kennyadsl @jacobherrington squashed!

frontend/app/views/spree/checkout/payment/_gateway.html.erb
```diff
@@ -1,27 +1,28 @@
 <%= image_tag 'credit_cards/credit_card.gif', id: 'credit-card-image' %>
 <% param_prefix = "payment_source[#{payment_method.id}]" %>
 
-<div class="field field-required">
+<div class="field field-required card_name" data-hook="card_name">
```
tvdeyen: I would prefer to not add additional `data-hook` attributes. They are a relict of the past.

kennyadsl: We can't remove the other `data-hook` from the same context (the other credit card fields). I think that keeping both here could be a good compromise to be minimum consistent. Not having the `data-hook` only in the card name will not enforce the deprecation in my opinion.

tvdeyen: I am not saying that an existing hook should be removed. As far as I can tell from the git diff a new `data-hook` is introduced.

kennyadsl: Yes you are right but my concern is that in the same file/form we would have:

```erb
<div class="field field-required card_name">
<div class="field field-required card_number" data-hook="card_number">
<div class="field field-required card_expiration" data-hook="card_expiration">
<div class="field field-required card_code" data-hook="card_code">
```

which is inconsistent. Anyway, I'm not too attached to this, I can live either way so if you think we should not add other `data-hook`s I'm ok asking @aitbw to remove it.

aitbw: I think @kennyadsl's reasoning makes sense for the sake of consistency and backwards compatibility (in case existing hooks are being used by stores)

tvdeyen: > sake of consistency and backwards compatibility 

Adding a new data-hook is not backwards compatibility, but anyway, this is not super important for me.

PR #3185, opened 2019-04-16
Add promotion rules article for Solidus admins

benjaminwil:
This is a small contribution that adds information around promotion rule types and eligibility. This is for Solidus admins.

**Checklist:**

### DIFF
--- a/guides/data/nav/users.yml
+++ b/guides/data/nav/users.yml
@@ -21,12 +21,12 @@
     dropdown:
       - title: "Overview"
         href: "/users/promotions/overview.html"
+      - title: "Promotion Rules"
+        href: "/users/promotions/promotion-rules.html"
       - title: "Promotion Actions"
         href: "/users/promotions/promotion-actions.html"
       - title: "Promotion Calculators"
         href: "/users/promotions/promotion-calculators.html"
-      - title: "Promotion Rules"
-        href: "/users/promotions/promotion-rules.html"
 
   - title: "Settings"
     dropdown:
--- a/guides/source/users/promotions/promotion-rules.html.md
+++ b/guides/source/users/promotions/promotion-rules.html.md
@@ -1,1 +1,51 @@
 # Promotion rules
+
+Promotion rules determine whether a promotion is eligible to be applied.
+Promotions let you control which customers can get promotional discounts and
+which ones cannot. For example, you may want to offer a promotion that is only
+valid for new customers.
+
+Many conventional types of promotion rules are available in Solidus by default.
+If you require a more specialized rule, speak with your developers about
+creating one.
+
+## Available promotion rule types
+
+- **First Order**: Eligible for a customer's first order only.
+- **First Repeat Purchase Since**: Eligible for a customer's first repeat
+  purchase since a specified date.
+- **Nth Order**: Eligible for a customer's *n*th order only.
+- **Item Total**: Eligible if the order total (before any adjustments) is less
+  than or greater than a specified amount.
+- **One Use Per User**: Eligible for use one time per user.
+- **Product(s)**: Eligible for specified products only.
+- **Option Value(s)**: Eligible for specified variants (product option values)
+  only.
+- **Taxon**: Eligible for products with specified taxons.
+- **User**: Eligible for specified users.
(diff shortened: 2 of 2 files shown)


guides/data/nav/users.yml
```diff
@@ -21,12 +21,12 @@
...
       - title: "Overview"
         href: "/users/promotions/overview.html"
+      - title: "Promotion Rules"
+        href: "/users/promotions/promotion-rules.html"
       - title: "Promotion Actions"
         href: "/users/promotions/promotion-actions.html"
       - title: "Promotion Calculators"
         href: "/users/promotions/promotion-calculators.html"
-      - title: "Promotion Rules"
-        href: "/users/promotions/promotion-rules.html"
```
aitbw: Is this change really necessary for this PR? I'm assuming most users would expect to have an alphabetically-sorted menu, what do you think?

guides/source/users/promotions/promotion-rules.html.md
```diff
@@ -1 +1,50 @@
...
+
+## Available promotion rule types
+
+- **First Order**: Eligible for a customer's first order only.
+- **First Repeat Purchase Since**: Eligible for a customer's first repeat
+  purchase since a specified date.
+- **Nth Order**: Eligible for a customer's *n*th order only.
+- **Item Total**: Eligible if the order total (before any adjustments) is less
+  than or greater than a specified amount.
+- **One Use Per User**: Eligible for use one time for each user.
```
aitbw: I think `Eligible for use one time per user` reads better, what do you think?

kennyadsl (approved):
Thanks Ben!

guides/source/users/promotions/promotion-rules.html.md
```diff
@@ -1 +1,50 @@
...
+
+## Rule matching
+
+There are two types of rule matching in Solidus' promotion system: **Match all
+of these rules** or **Match any of these rules**.
+
+- **Match all of these rules**: This setting ensure that every rule must be
+  matched before the promotion is applied. This is great if you only want to
+  apply promotions for a very specific kind of order, like for first-time
+  customers who order more than $100 USD of the T-Shirt or Tote Bag products.
```
aitbw: I think `(...) whose order is valued at $100 USD or more` reads better and it's more concise, what do you think?

guides/source/users/promotions/promotion-rules.html.md
```diff
@@ -1 +1,50 @@
...
+- **Match any of these rules**: This setting allows you to create more flexible
+  promotions. For example, if you wanted to give a discount to customers who
+  buy more than $100 USD of products or are a first-time customer.
```
aitbw: I think it's better to use other examples in here, to avoid any possible confusions; what do you think?

aitbw (commented):
Hey @benjaminwil, thanks for showing the docs some love! :heart:

I've left you some comments I'd like to discuss first before giving you my approval, let me know what you think!

jacobherrington (approved):
Thanks Ben! I think @aitbw has great feedback, but I'm happy with this either way. 🥇

guides/data/nav/users.yml
benjaminwil: I would expect administrators to read about "Promotion Rules" before they read about "Promotion Actions." Perhaps my commit message should have stated that more explicitly.

While the guides may not be read linearly, we can prioritize articles by putting them up higher in the menus. We do this with "Overview" articles.

This is a low-risk change so I will include it. We can revert later if we formalize menu orders.

guides/source/users/promotions/promotion-rules.html.md
benjaminwil: Great edit!

guides/source/users/promotions/promotion-rules.html.md
benjaminwil: I see what you mean!

benjaminwil:
Responded and committed fixes for @aitbw's feedback. :+1:

guides/data/nav/users.yml
kennyadsl: I think it makes sense to read rules before actions since this is also the order we are presenting information in the UI.

PR #3187, opened 2019-04-17
Privatize PaymentMethod::Check#simulated_successful_billing_response

jacquesporveau:
**Description**
This method should only be called from within itself. The response is
specific to this class and being able to obtain this response from
outside of it doesn't make sense to me.

**Other notes**
I did an `ag` for occurrences of this method in the project and it looks like it is only ever called from within the class so this is a safe change to make.

Screen Shot 2019-04-17 at 1 33 45 PM

**Checklist:**

### DIFF
--- a/core/app/models/spree/payment_method/check.rb
+++ b/core/app/models/spree/payment_method/check.rb
@@ -33,6 +33,8 @@ module Spree
       false
     end
 
+    private
+
     def simulated_successful_billing_response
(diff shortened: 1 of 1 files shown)


kennyadsl:
@jacquesporveau Hey there, I re-ran specs and they are still failing. Can you please try rebasing against `master` and running failing specs locally to see if they are real failures?

PR #3189, opened 2019-04-17
Fix broken links in guides.

jacquesporveau:
**Description**
These links both 404'd. They were missing parts of an extension.

**Checklist:**

### DIFF
--- a/guides/source/developers/payments/payment-sources.html.md
+++ b/guides/source/developers/payments/payment-sources.html.md
@@ -12,7 +12,7 @@ Solidus includes some payment sources such as `Spree::CreditCard` and
 `Spree::StoreCredit`. However, your [payment method][payment-methods] could
 define any custom payment source in its `payment_source_class` method.
 
-[payment-methods]: payment-methods.html
+[payment-methods]: payment-methods.html.md
 
 ## Credit cards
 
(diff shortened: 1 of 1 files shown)


jacquesporveau:
Ah I see. It looks like these links probably works within the context of the guides website but not within the github repo.

https://github.com/solidusio/solidus/blob/master/guides/source/developers/payments/payment-sources.html.md

The links on that page aren't working for me. Maybe this isn't something that we super care about since most folk are probably using that site anyways.

benjaminwil:
Just double-checked this locally. It breaks the guides website.

This creates HTML links with `<a href="payment-methods.html.md"></a>` which do not exist on the site.

We'd have to either write a Middleman Markdown extension to deal with this, or move away from Middleman.

As an example, VuePress doesn't have this issue: Try clicking the `configureWebpack` link in both the [site view](https://vuepress.vuejs.org/guide/assets.html) and the [repo view](https://github.com/vuejs/vuepress/blob/master/packages/docs/docs/guide/assets.md).

jacquesporveau:
I think it's better for the online guides to work rather than the github repo.

kennyadsl:
I tried in the preview area generated for this PR by Netlify and I think it was working but maybe I clicked on the wrong links?

PR #3192, opened 2019-04-24
Update admin color palette and font

mfrecchiami:
With the aim of defining a solid brand image for Solidus, we updated the color palette and the backend font with Inter, a 100% Open Source font family; then we'll replace the product images with the Solidus design, then all the icons (now there's _font-awesome_ ) will be replaced with a custom icon font.

After this first phase, we'll deal with more structural improvements related to the elements such as forms, main navigation etc.

Our goal must be to make the Solidus product attractive also to the decision makers and to do so it is necessary to invest in the visual as well as marketing part.
Furthermore, a cured (as well as performed) product also gives the idea of solidity.

Some screenshots here:

screencapture-orders-2019-04-29

screencapture-products-2019-04-29

screencapture-products-edit-2019-04-29

screencapture-products-prices-2019-04-29


dhonig:
Great work!  Again I'd like to suggest that we drop the spree_frontend altogether and start working on starter kits in react/vue.js etc.

benjaminwil:
> then all the icons (now there's font-awesome ) will be replaced with a custom icon font

I agree that this will more attractive to decision makers, and I am happy so much care is going into the backend.

I hope we will still keep font-awesome or some other stock icon library easily available for developers. It is great to have ready-made icons available for extensions and custom backend views that Solidus does not provide out of the box.

backend/app/assets/stylesheets/spree/backend/globals/_fonts.scss
```diff
@@ -0,0 +1,15 @@
...
+       url("Inter-Regular.woff") format("woff");
+}
+
+@font-face {
+  font-family: 'Inter';
+  font-style:  normal;
+  font-weight: 600;
+  src: url("Inter-SemiBold.woff2") format("woff2"),
+       url("Inter-SemiBold.woff") format("woff");
+}
```
elia: Final newline missing

backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
```diff
@@ -75,9 +79,13 @@ $custom-select-focus-border-color: $input-focus-border-color !default;
...
 
 // Breadcrumbs
 
 $breadcrumb-bg:                 transparent !default;
-$breadcrumb-active-color:       $color-4 !default;
+$breadcrumb-active-color:       $color-primary !default;
+
+// Popover
+
+$popover-header-bg:             #e3ebfd !default;
```
elia: Final newline missing

backend/app/assets/stylesheets/spree/backend/components/_messages.scss
```diff
@@ -1,14 +1,16 @@
 .errorExplanation {
-  padding: 5px;
-  border: 1px solid very-light(theme-color("danger"), 12);
-  background-color: very-light(theme-color("danger"), 6);
-  border-radius: 3px;
-  color: very-light(theme-color("danger"), 30);
+  padding: 10px 15px;
+  //border: 1px solid very-light(theme-color("danger"), 12);
+  border-top: 3px solid theme-color("danger");
+  //background-color: very-light(theme-color("danger"), 6);
```
elia: These commented CSS rules can be removed safely ✂️

backend/app/assets/stylesheets/spree/backend/globals/_variables_override.scss
```diff
@@ -4,4 +4,4 @@
   To set your own colors, sizes or fonts just override this 
   file in your application and set variables according to
   globals/_variables.scss file.
-  --------------------------------------------------------- */
+  --------------------------------------------------------- */
```
elia: Final newline missing

mfrecchiami:
Hi @benjaminwil thank you for your feedback!
Actually we thought a lot about the font icon issue: maybe the best solution is to create custom icons only for the main navigation and replace Font Awesome with another open source font that has the same look and feel of the custom navigation icons.
We're looking for the best solution, both for the brand and for the devs ;)

backend/app/assets/stylesheets/spree/backend/globals/_fonts.scss
mfrecchiami: Thanks @elia !
Actually few weeks ago I played with Atom settings, and from that moment something went weird. I'm thinking about reset settings to my last backup.

mfrecchiami:
@jacobherrington and @ericsaupe thank you! 🙏

backend/app/assets/stylesheets/spree/backend/globals/_variables.scss
```diff
@@ -10,59 +10,68 @@ $font-size-pill: 11px;
 //--------------------------------------------------------------
 
 // Basic color palette for admin
-$color-1:                        #FFFFFF !default;    // White
-$color-2:                        #9FC820 !default;    // Green
-$color-3:                        #5498DA !default;    // Light Blue
-$color-4:                        #6788A2 !default;    // Dark Blue
-$color-5:                        #C60F13 !default;    // Red
-$color-6:                        #FF9300 !default;    // Yellow
-$color-7:                        #F4F4F4 !default;    // Light Gray
```
kennyadsl: I've just one concern about removing variables. If extensions or store were using them they'll end up having an error while compiling. What about moving them into a `_deprecated-variables.scss` file instead? 

I'm not sure if there's a way to actually deprecate them with scss. I've made some test with this code:

```scss
@if variable-exists(color-1) {
  @warn '$color-1 will be removed, please use $color-primary;';
  $color-1: $color-primary;
}
```

and it works, but it only prints a deprecation message if you re-define `$color-1`, not if you just use it. It also has to be declared **after** the `$color-1` definition, which is basically useless for both extensions and stores since they will eventually declare/override variables after solidus files are loaded.

@tvdeyen do you have any idea around this?

tvdeyen: Great point @kennyadsl 

We already have a way to deprecate variables https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/globals/_deprecation.scss

You use it like this https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/backend/app/assets/stylesheets/spree/backend/globals/_variables.scss#L42

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -101,13 +104,15 @@ nav.menu {
...
     color: $color-navbar;
-    font-weight: normal;
+    font-weight: $font-weight-bold;
   }
 
   .icon_link {
     text-indent: 2em;
 
     &:before {
+      color: $color-8;
```
tvdeyen: Could we not use color variables inside of our sass files, but use dedicated variables, please? So in this case add a new variable (`$navigation-icon-color`) and give it a default of `$color-8`?

I know that the whole admin sass files are build like this, but this is not something we want to continue.

backend/app/assets/stylesheets/spree/backend/shared/_layout.scss
```diff
@@ -18,6 +22,7 @@ body {
 }
 
 .content-wrapper {
+  background-color: $color-1;
```
tvdeyen: Could we add a `$content-wrapper-bg-color` variable instead of using a color variable here directly?

backend/app/assets/stylesheets/spree/backend/components/_breadcrumb.scss
```diff
@@ -1,4 +1,16 @@
 .breadcrumb {
   font-size: 16px;
   margin-bottom: 0;
+  
+  a {
+    color: $color-8;
```
tvdeyen: Please add a variable for this use case and give it a default of `$color-8` instead of using the color variable directly.

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -191,6 +191,7 @@ nav.menu {
 
   a {
     display: inline-block;
+    font-weight: 600;
```
tvdeyen: Please use a Bootstrap variable for the font weight

backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
```diff
@@ -7,24 +7,27 @@
 //
 // Grayscale and brand colors for use across Bootstrap.
 
-$blue:   $color-3 !default;
-$red:    $color-5 !default;
-$orange: $color-6 !default;
-$green:  $color-2 !default;
+$primary:          $color-primary !default;
+$secondary:        #C7CDD5 !default;
```
tvdeyen: Could we make this a variable?

backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
```diff
@@ -64,9 +67,10 @@ $border-radius:          3px !default;
 
 // Forms
 
-$input-color:                    $color-3 !default;
-$input-border-color:             $color-border !default;
-$input-focus-border-color:       $color-2 !default;
+$input-color:                    $color-dark !default;
+$input-placeholder-color:        $color-dark-light !default;
+$input-border-color:             #D7DAE0 !default;
```
tvdeyen: Could we make this a variable?

backend/app/assets/stylesheets/spree/backend/_bootstrap_custom.scss
```diff
@@ -75,9 +79,14 @@ $custom-select-focus-border-color: $input-focus-border-color !default;
...
 
 // Breadcrumbs
 
 $breadcrumb-bg:                 transparent !default;
-$breadcrumb-active-color:       $color-4 !default;
+$breadcrumb-active-color:       $color-primary !default;
+
+// Popover
+
+$popover-header-bg:             #e3ebfd !default;
```
tvdeyen: Could we make this a variable?

backend/app/assets/stylesheets/spree/backend/components/_buttons.scss
```diff
@@ -0,0 +1,14 @@
+.btn-default {
+  background: linear-gradient(#FFF, #ECEDF1);
```
tvdeyen: Could we make use of variables in this file?

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
```diff
@@ -17,7 +17,9 @@
   .order-total {
     font-size: 35px;
     font-weight: $font-weight-bold;
-    color: theme-color("success");
+    background: #e3ebfd;
```
tvdeyen: Could we make this a variable?

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
```diff
@@ -28,7 +30,9 @@
   .customer {
     font-size: 20px;
     font-weight: $font-weight-bold;
-    color: theme-color("success");
+    background: #e3ebfd;
```
tvdeyen: Could we make this a variable?

backend/app/assets/stylesheets/spree/backend/sections/_orders.scss
```diff
@@ -37,10 +41,17 @@
...
+      padding: .1em .3em;
+      border-radius: 3px;
     }
     .stock-location-name {
-      color: theme-color("success");
+      color: #5B6778;
+      background: #e3ebfd;
+      padding: .1em .3em;
+      border-radius: 3px;
+      word-break: break-word;
```
tvdeyen: Could we make use of variables in this file?

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -41,6 +41,8 @@ textarea {
 }
 
 label {
+  font-size: 12px;
```
tvdeyen: We should make this a relative font size

backend/app/assets/stylesheets/spree/backend/shared/_forms.scss
```diff
@@ -64,7 +66,18 @@ button,
...
+    border-color: #C7CDD5;
+    
+    &:hover {
+      background: linear-gradient(#FFF, darken(#ECEDF1, 3%));
+    }
+    
+    &.disabled,
+    &:disabled {
+      color: $color-disabled;
+    }
```
tvdeyen: Could we make use of variables in this file?

backend/app/assets/stylesheets/spree/backend/shared/_typography.scss
```diff
@@ -17,6 +17,11 @@ hr {
   border-left: none;
 }
 
+strong,
+b {
+  font-weight: 600;
```
tvdeyen: There should be a Bootstrap variable for this. Or we should add one for this weight.

backend/app/assets/stylesheets/spree/backend/globals/_fonts.scss
```diff
@@ -0,0 +1,16 @@
...
+  font-style:  normal;
+  font-weight: 400;
+  src: url("Inter-Regular.woff2") format("woff2"),
+       url("Inter-Regular.woff") format("woff");
+}
+
+@font-face {
+  font-family: 'Inter';
+  font-style:  normal;
+  font-weight: 600;
```
tvdeyen: We should use a font weight variable here that is then used wherever we use this font face and weight

backend/app/views/spree/admin/shared/_head.html.erb
```diff
@@ -5,9 +5,6 @@
 
 <title><%= admin_page_title %></title>
 
-<!-- Get "Open Sans" font from Google -->
-<link href='//fonts.googleapis.com/css?family=Open+Sans:400italic,600italic,400,600&subset=latin,cyrillic,greek,vietnamese' rel='stylesheet' type='text/css'>
-
```
tvdeyen: 👍 💯 finally!

tvdeyen:
We probably want to update the date picker styles as well? 

Monosnap 2019-05-09 17-34-46

tvdeyen:
There are still some places that use the old button classes and therefore get the secondary button style. Product -> Images for instance.

Screen Shot 2019-05-09 at 17 37 00

BTW: I really like the secondary button style, as it looks like a button (a mayor downside of the flat UI trend). But it looks a bit misplaced next to the primary buttons that are still flat. Any plans to make them non-flat buttons as well? I would appreciate that.

tvdeyen:
> I hope we will still keep font-awesome or some other stock icon library easily available for developers. It is great to have ready-made icons available for extensions and custom backend views that Solidus does not provide out of the box.

I agree with Benjamin here. Lots of stores and extensions use the provided icons. We can't just replace them without breaking them.

> Actually we thought a lot about the font icon issue: maybe the best solution is to create custom icons only for the main navigation and replace Font Awesome with another open source font that has the same look and feel of the custom navigation icons.
We're looking for the best solution, both for the brand and for the devs ;)

I understand the thought from a brand standpoint, but we are an open source framework that is used because of its extensibility and ease of customisation. That includes icons IMO. Custom drawn icons is a maintenance burden for us. I advice us to use open source icons whenever possible.

FontAwesome 5 is a great resource for [free icons](https://fontawesome.com/icons?d=gallery&m=free) that has [lots of modern integration options](https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use).

Please keep that in mind.

Thanks for all the great work on the new admin. I really appreciate the work that has been done here.

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
mfrecchiami: Hi @tvdeyen many thanks for your feedbacks and advices! I replaced all missed HEX colors in the scss files,and the font-weight style, also.

backend/app/assets/stylesheets/spree/backend/shared/_fonts.scss
```diff
@@ -0,0 +1,15 @@
...
+  src: font-url("../../fonts/inter/Inter-Regular.woff2") format("woff2"),
+       font-url("../../fonts/inter/Inter-Regular.woff") format("woff");
+}
+
+@font-face {
+  font-family: 'Inter';
+  font-style:  normal;
+  font-weight: $font-weight-bold;
+  src: font-url("../../fonts/inter/Inter-SemiBold.woff2") format("woff2"),
+       font-url("../../fonts/inter/Inter-SemiBold.woff") format("woff");
```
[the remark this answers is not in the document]
mfrecchiami: Hi @tvdeyen thank you for your precious suggestions!
I've moved the `fonts` path to `vendor/assets/` path. 
I cannot make it before, because I didn't know I should have added `config.assets.precompile << /\.(?:svg|eot|woff|ttf)\z/` to let Rails to compile font formats 😅

backend/app/assets/stylesheets/spree/backend/globals/_variables.scss
mfrecchiami: Hi guys! I've just pushed a commit to resolve this issue; as you can seen the old numbered color variables has been re-added in a their own scss partial, and I used the existing sass mixin, suggest by @tvdeyen (I just duplicate it and renamed the old one because it refers to Bootstrap variables, and its warning message, also). 
Thank you @kennyadsl for your feedback!

The message that I wrote refers to a future 3.0 version of Solidus, so we can keep the deprecated variables on 2.8 and 2.9vs. What do you think about that?

tvdeyen: Awesome 🍰

backend/app/assets/stylesheets/spree/backend/shared/_fonts.scss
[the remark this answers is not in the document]
tvdeyen: 👍

jacquesporveau:
Great job!

mtomov:
Indeed, thank you for that!

PR #3193, opened 2019-04-25
More error codes to apply_coupon_code api response

fkoessler:
Makes it possible for decoupled frontend to identify error and display custom error messages.

**Description**

With the current implementation, if a promotion is not eligible, the api endpoint returns the translated eligibility error message. There is no error code sent with the response.

This PR adds error codes to some of the api error responses. If no code is present in the error, the generic `coupon_code_not_eligible` code is returned.

It makes it possible for a decoupled frontend to identify the error when applying a coupon code and display a custom error message.

The only way I found to pass the error code from a promotion rule to the coupon promotion handler is by adding the code as a detail to the error, see: https://guides.rubyonrails.org/active_record_validations.html#validations-overview-errors-details
It is then possible to retrieve the code in the coupon promotion handler with: `promotion.eligibility_errors.details[:base].first[:code]`

Let me know if there's a better way to do that or if I can improve the PR in any other way.

**Checklist:**

### DIFF
--- a/core/app/models/spree/promotion/rules/first_order.rb
+++ b/core/app/models/spree/promotion/rules/first_order.rb
@@ -16,7 +16,7 @@ module Spree
 
           if user || email
             if !completed_orders.blank? && completed_orders.first != order
-              eligibility_errors.add(:base, eligibility_error_message(:not_first_order))
+              eligibility_errors.add(:base, eligibility_error_message(:not_first_order), error_code: :not_first_order)
             end
           end
 
--- a/core/app/models/spree/promotion/rules/item_total.rb
+++ b/core/app/models/spree/promotion/rules/item_total.rb
@@ -20,7 +20,7 @@ module Spree
           return false unless order.currency == preferred_currency
           item_total = order.item_total
           unless item_total.send(preferred_operator == 'gte' ? :>= : :>, BigDecimal(preferred_amount.to_s))
-            eligibility_errors.add(:base, ineligible_message)
+            eligibility_errors.add(:base, ineligible_message, error_code: ineligible_error_code)
           end
 
           eligibility_errors.empty?
@@ -37,6 +37,14 @@ module Spree
             eligibility_error_message(:item_total_less_than, amount: formatted_amount)
           else
             eligibility_error_message(:item_total_less_than_or_equal, amount: formatted_amount)
+          end
+        end
+
+        def ineligible_error_code
+          if preferred_operator == 'gte'
+            :item_total_less_than
+          else
+            :item_total_less_than_or_equal
           end
         end
       end
--- a/core/app/models/spree/promotion/rules/one_use_per_user.rb
+++ b/core/app/models/spree/promotion/rules/one_use_per_user.rb
@@ -11,10 +11,10 @@ module Spree
         def eligible?(order, _options = {})
           if order.user.present?
             if promotion.used_by?(order.user, [order])
-              eligibility_errors.add(:base, eligibility_error_message(:limit_once_per_user))
+              eligibility_errors.add(:base, eligibility_error_message(:limit_once_per_user), error_code: :limit_once_per_user)
             end
           else
-            eligibility_errors.add(:base, eligibility_error_message(:no_user_specified))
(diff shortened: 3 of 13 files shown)


fkoessler:
Hello @kennyadsl , thanks for the review.

How about renaming the `code` parameter to `status_code`?
That way it's obvious it's not the promotion code, and it's coherent with the existing implementation of `set_error_code` :

```
      def set_error_code(status_code, options = {})
        @status_code = status_code
        @error = options[:error] || I18n.t(status_code, scope: 'spree')
      end
```

I'll update the PR if we agree on the name.

kennyadsl:
I think it could work but what about `error_code`? `status_code` makes me think to an HTTP status code. 🙂

ericsaupe:
I second the change to `error_code` but overall great improvement.

fkoessler:
PR updated:
  - renamed `code` to `error_code`
  - change in `PromotionHandler::Coupon` `#set_error_code` sets `@error` with translated promotion rule validation error message instead of error code (`@status_code` contains the error code returned by the api)

kennyadsl:
@fkoessler one is a real failure, the other will go away rebasing against `master`.

fkoessler:
@kennyadsl 

I wanted to get rid of the second parameter in function call `set_error_code(promotion.eligibility_errors.details[:base].first[:error_code], promotion.eligibility_errors.details[:base].first) if error_code_present?(promotion)` in `Spree::PromotionHandler::Coupon` `#handle_present_promotion` by retrieving the translated error message from the error code.

However it did not work because parameters for the translation are missing in `Spree::PromotionHandler::Coupon`.

So I reverted the changes.
Let me know if you think of a way to improve the PR.

kennyadsl:
What about only changing the scope of the translation if we have an eligibility error?

```ruby
      def set_error_code(status_code, scope: 'spree')
        @status_code = status_code
        @error = I18n.t(status_code, scope: scope)
      end
```

Calling it with `scope: [:spree, :eligibility_errors, :messages]` should fix the issue, right?

fkoessler:
@kennyadsl that was my previous commit with the failing spec.

It won't fix the issue, the variable part of the translation will be missing.

For example, `formatted_amount` in `Spree::Promotion::Rule::ItemTotal` `#ineligible_message`

```
        def ineligible_message
          if preferred_operator == 'gte'
            eligibility_error_message(:item_total_less_than, amount: formatted_amount)
          else
            eligibility_error_message(:item_total_less_than_or_equal, amount: formatted_amount)
          end
        end
```
Translation parameters could be passed as error detail, the same `error_code` is passed.

I'm not sure it's a good idea though, I don't think it's the way ActiveModel Errors are supposed to be used.

kennyadsl:
Got it, you are right. 

Have you considered creating another method to set the errors codes and messages for these eligibility errors? Something like:

```ruby
      def set_promotion_eligibility_error_code(promotion)
        return unless eligibility_error_code_present?(promotion)

        eligibility_error = promotion.eligibility_errors.details[:base].first

        @status_code = eligibility_error[:error_code] || :generic_code
        @error = eligibility_error[:error] || "generic_message"
      end
```

I think it could be better in order to encapsulate all the eligibility errors logic into a different place. WDYT?

Also, another last-minute doubt: can we have 2 or more different eligibility errors for the same promotion?

fkoessler:
@kennyadsl I updated the PR, it's cleaner with all the eligibility errors logic in its dedicated method.

I think it's starting to look decent. Let me know what you think and if it can be further improved.

To answer your question, yes there can be many eligibility errors for the same promotion. Each existing promotion rule can potentially generate an eligibility error. Presently only the first error is returned by the API endpoint.

kennyadsl:
Cool, supporting multiple errors could be another improvement done into another PR. Can you please take a look at failing specs, afterward we are good to go IMO!

fkoessler:
I rebased against master, the failing specs is now fixed.

kennyadsl:
@fkoessler I've opened https://github.com/solidusio/solidus/issues/3227 to follow up with the last point discussed, thanks again for the contribution here!

PR #3195, opened 2019-05-02
Prevent promotion rule to show user multiple times

### LINKED ISSUE #3183: User picker proposes the same users at promotion edit page
When we have a user, who has few addresses then he will be proposed with user_picker.js multiple times.

**Solidus Version**:
2.4.2 (checked at https://solidus-master-demo.herokuapp.com/ also)
[Gemfiles](https://gist.github.com/VitaliyAdamkov/738512cfe887ead87d852be34c5ec99a)

**To Reproduce**
1. Make an order for default user with Address1.
2. Make an order for default user with Address2.
3. Create a promotion.
4. Go to the promotion edit page.
5. Add the user promotion rule.
6. Type any letter from default user's email.

**Current behavior**
API call runs and select shows default user at least two times.

**Expected behavior**
Select is expected to show unique users.

**Screenshots**
изображение

 - OS: Ubuntu 18.04
 - Browser Mozilla Firefox
 - Version 66.0.2

vzqzac:
**Description**

The ransack query used to fetch users for the promotion rule dropdown checked for attributes that may return repeated results. This adds the `distinct` flag in the result's call to prevent repeated records to be shown.
Fixes #3183 

**This affects the API's `ResourceController`, so the scope of the update may be wide, although the change will likely fix un-encountered issues (IMO we shouldn't be returning repeated records independently of the API call, but open to discuss)**

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/resource_controller.rb
+++ b/api/app/controllers/spree/api/resource_controller.rb
@@ -9,7 +9,7 @@ class Spree::Api::ResourceController < Spree::Api::BaseController
       ids = params[:ids].split(",").flatten
       collection_scope = collection_scope.where(id: ids)
     else
-      collection_scope = collection_scope.ransack(params[:q]).result
+      collection_scope = collection_scope.ransack(params[:q]).result(distinct: true)
     end
 
     @collection = paginate(collection_scope)
--- a/api/spec/controllers/spree/api/resource_controller_spec.rb
+++ b/api/spec/controllers/spree/api/resource_controller_spec.rb
@@ -34,6 +34,18 @@ module Spree
       model do
         acts_as_list
         validates :name, presence: true
+        has_many :components
+      end
+    end
+
+    with_model 'Component', scope: :all do
+      table do |t|
+        t.string :name
+        t.belongs_to :widget
+      end
+
+      model do
+        belongs_to :widget
       end
     end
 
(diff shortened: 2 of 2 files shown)


vzqzac:
I'm not totally sure if the approach I followed for the spec is the best one, I'd happily change it if a better approach can be accomplished

kennyadsl:
We discussed this in the core meeting and we think that this is a bit too extreme as solution. 😬 

If we are having multiple results there is the possibility that there are wrong table joins and it would be great to understand where is the problem and fix just that. What about going a little bit deeper and try to understand where the duplicate come from?

vzqzac:
@kennyadsl it is indeed! The actual source of the _problem_ relies in the user picker javascript: https://github.com/solidusio/solidus/blob/a3cc823e34f627681bd4507cf54b43dccd5569a3/backend/app/assets/javascripts/spree/backend/user_picker.js#L28-L33

Whit that Ransack query, the resulting pg query we obtain when giving an input `"a"` is something like:
```
[2] pry(#<Spree::Api::UsersController>)> collection_scope.ransack(params[:q]).result
  Spree::User Load (0.8ms)  SELECT "spree_users".* FROM "spree_users" LEFT OUTER JOIN
 "spree_user_addresses" ON "spree_user_addresses"."user_id" = "spree_users"."id" AND
 "spree_user_addresses"."archived" = $1 LEFT OUTER JOIN "spree_addresses" ON
 "spree_addresses"."id" = "spree_user_addresses"."address_id" WHERE "spree_users"."deleted_at" IS
 NULL AND (("spree_users"."email" ILIKE 'a%' OR "spree_addresses"."firstname" ILIKE 'a%') OR
 "spree_addresses"."lastname" ILIKE 'a%')  [["archived", false]]
```

The easiest workaround I found was to just set the `distinct true` parameter, but I totally understand how extreme this is, given the controller where this happens. This could be done in a view too, but unfortunately, the call is done via AJAX, we can update that to sanitize though, but doesn't seem like ta proper solution in my opinion.  

Do you have any suggestion? I'm happy to discuss this and update the code with whatever we come up to 😃

kennyadsl:
@vzqzac thanks for the detailed report! 

I was investigating a little bit further and I found a few things:

- From what I can read in the [Ransack REAMDE](https://github.com/activerecord-hackery/ransack#problem-with-distinct-selects), this is a known issue, and to fix it `distinct: true` alone could not work with all the scenarios. It looks like sometime you also have to add some explicit `includes`/`joins` to the ransack call. If this is true we may break some extensions or stores that are using Ransack relying on ResourceController, right?
-  [This piece of code in the backend users controller](https://github.com/solidusio/solidus/blob/a3cc823e34f627681bd4507cf54b43dccd5569a3/backend/app/controllers/spree/admin/users_controller.rb#L106). It looks like we are doing something very similar there with a custom query. I'm not sure it's better and honestly, I don't even know if we are using it but it may be worth investigating why that piece of code exists and if we need something similar in the users API controller as well.

Let me know your thoughts. 🙂

vzqzac:
I'm closing this since I haven't worked on it from a long time now, I'll try to open a new pull request based on @kennyadsl suggestions

Issue #3198, opened 2019-05-03
Order state does not change after a customer return is received


spaghetticode:
When an order has more than one line item, creating a return for all the items at once and setting the reception status to `Received` fails to change the order state to `Returned`, like it does with orders with one single line item. See attached screenshot.
 
**Solidus Version**:

master

**Screenshots**

<img width="979" alt="Schermata 2019-05-03 alle 17 22 38" src="https://user-images.githubusercontent.com/141220/57147972-ff8f9f00-6dc8-11e9-988f-90d2249c4406.png">

jacobherrington:
@spaghetticode Can this issue be closed after #3199?

spaghetticode:
@jacobherrington sure!

PR #3199, opened 2019-05-03
Allow order with multiple line items to be marked as "Returned"

spaghetticode:
**Description**

This PR fixes issue #3198 

When a customer return is created for an order with multiple line items and the
return state of all the items is `Received` then the order state should change
to `Returned` just like it happens with orders with only one line item.

**Checklist:**

### DIFF
--- a/core/app/models/spree/return_item.rb
+++ b/core/app/models/spree/return_item.rb
@@ -130,6 +130,8 @@ module Spree
       after_transition any => any, do: :persist_acceptance_status_errors
     end
 
+    attr_accessor :skip_customer_return_processing
+
     # @param inventory_unit [Spree::InventoryUnit] the inventory for which we
     #   want a return item
     # @return [Spree::ReturnItem] a valid return item for the given inventory
@@ -233,10 +235,12 @@ module Spree
 
     def process_inventory_unit!
       inventory_unit.return!
-
       if customer_return
         customer_return.stock_location.restock(inventory_unit.variant, 1, customer_return) if should_restock?
-        customer_return.process_return!
+        unless skip_customer_return_processing
+          Deprecation.warn 'From Solidus v2.9 onwards, #process_inventory_unit! will not call customer_return#process_return!'
+          customer_return.process_return!
+        end
       end
     end
 
--- a/backend/app/controllers/spree/admin/customer_returns_controller.rb
+++ b/backend/app/controllers/spree/admin/customer_returns_controller.rb
@@ -10,6 +10,7 @@ module Spree
       before_action :load_form_data, only: [:new, :edit]
       before_action :build_return_items_from_params, only: [:create]
       create.fails  :load_form_data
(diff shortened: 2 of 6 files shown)


core/app/models/spree/order.rb
```diff
@@ -286,7 +286,7 @@ def allow_cancel?
     end
 
     def all_inventory_units_returned?
-      inventory_units.all?(&:returned?)
+      inventory_units.returned.count == inventory_units.count
```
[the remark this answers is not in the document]
spaghetticode: This code is executed for each return item, for a better understanding of the process you can start from `ReturnItem#process_inventory_unit!`. The first inventory unit is marked as `returned`, and eventually `order.inventory.units` is called. 

The query is executed only for the first return item, and as we know there's no issue as long as the customer return is for one single item. But when there's more than one, from the second time on, `order.inventory.units` refers to stale data, and must be updated. Using `count` ensures a new (lightweight) query is performed. 

Basically, it happens something similar to this:
```ruby
o = create :shipped_order, line_items_count: 2
o.inventory_units.all? &:shipped?
=> true

u = Spree::InventoryUnit.find(o.inventory_units.first.id)
u.return!
u.reload.returned?

o.inventory_units.all? &:shipped?
=> true
```

core/app/models/spree/return_item.rb
```diff
@@ -233,10 +235,12 @@ def currency
 
     def process_inventory_unit!
       inventory_unit.return!
-
       if customer_return
         customer_return.stock_location.restock(inventory_unit.variant, 1, customer_return) if should_restock?
-        customer_return.process_return!
+        unless skip_customer_return_processing
```
kennyadsl: I think we should also test that a deprecation warning is raised in some circumstances.

backend/spec/controllers/spree/admin/reimbursements_controller_spec.rb
```diff
@@ -38,7 +38,10 @@
     let(:order) { customer_return.order }
     let(:return_item) { customer_return.return_items.first }
     let(:payment) { order.payments.first }
-    before { return_item.receive! }
+    before do
+      return_item.skip_customer_return_processing = true
```
kennyadsl: Do you think it could be a good idea to add this as default in the return item factory?

core/app/models/spree/return_item.rb
spaghetticode: Makes sense, I added this spec: https://github.com/solidusio/solidus/pull/3199/files#diff-0f55d3185362ecf4697cad5429703bd0

backend/spec/controllers/spree/admin/reimbursements_controller_spec.rb
spaghetticode: I totally agree 👍 
This also reduced the changes I had to make to spec files

PR #3202, opened 2019-05-07
Fixes for discard 1.1.0

kennyadsl:
**Description**

This PR fixes a bug with the last version of discard. I think the bug was more related to how we were using discard than to the discard update itself.

**TLDR;**

`discarded?` was returning true, even when the record updated to `deleted_at` wasn't yet persisted. This behavior seems to be changed (probably correctly?) so this validation

```
validate :validate_no_amount_used, if: :discarded?
```

was never performed on the first `discard` call, which is exactly what we need. 

This PR changes that way to validate the record into a `before_discard` callback (similar to what we have with the `after_destroy`), [throwing `abort` to interrupt the callback chain](https://api.rubyonrails.org/classes/ActiveModel/Callbacks.html).

Asking for a @jarednorman review as well, since he's the new discard maintainer. 🎉 

**Checklist:**

### DIFF
--- a/core/app/models/spree/store_credit.rb
+++ b/core/app/models/spree/store_credit.rb
@@ -39,7 +39,7 @@ class Spree::StoreCredit < Spree::PaymentSource
   before_validation :associate_credit_type
   before_validation :validate_category_unchanged, on: :update
   before_destroy :validate_no_amount_used
-  validate :validate_no_amount_used, if: :discarded?
+  before_discard :validate_no_amount_used
 
   attr_accessor :action, :action_amount, :action_originator, :action_authorization_code, :store_credit_reason
 
--- a/core/app/models/spree/store_credit.rb
+++ b/core/app/models/spree/store_credit.rb
@@ -268,6 +268,7 @@ class Spree::StoreCredit < Spree::PaymentSource
   def validate_no_amount_used
     if amount_used > 0
       errors.add(:amount_used, 'is greater than zero. Can not delete store credit')
+      throw :abort
     end
   end
 


kennyadsl:
@jarednorman do you think we need to update the discard dependency along with this change? I'm not sure the old discard version is compatible with this new code. Let me know!

jarednorman:
In reviewing the `discard` gem, I'm not sure what caused this change. The only change I see to how `discarded?` works since v1.0.0 is [this one](https://github.com/jhawthorn/discard/commit/dfdd984b37bf3d646b276d68e47f7f998c6aaad9#diff-0a62a055b1aeb5c4b72b9318cebd32f2R66), which doesn't seem like it should change the behaviour.

kennyadsl:
I've just run specs with discard `1.0.0` and they pass so I think it's backward compatible. 🎉 

<details>
<summary>Investigation that leads me to open jhawthorn/discard/issues/51</summary>

<br>
I've also tried to run `master` specs with discard `1.0.0` and they pass.
Next attempt done is using previous specs and discard `1.1.0`, which had the failure that originated this PR. I changed in discard locally:

```ruby
run_callbacks(:discard) do
  update_attribute(self.class.discard_column, Time.current)
end
```

to:

```ruby
with_transaction_returning_status do
  run_callbacks(:discard) do
    self[self.class.discard_column] = Time.current
    save
  end
end
```

and specs pass, so I assumed the problem is here. I then changed that block into:

```ruby
run_callbacks(:discard) do
  self[self.class.discard_column] = Time.current
  save
end
```

removing `with_transaction_returning_status` and they still pass, so I assume the problem is with the `with_transaction_returning_status` change.

From [with_transaction_returning_status](https://apidock.com/rails/ActiveRecord/Transactions/with_transaction_returning_status) API reference, it:

> Executes method within a transaction and captures its return value as a status flag. If the status is true the transaction is committed, otherwise a ROLLBACK is issued. In any case the status flag is returned.

I think that removing this method somehow changes the state the record is when it's time to execute validations, but I'm not sure if it's an issue worth spending too much time on so I've opened [this PR on discard](https://github.com/jhawthorn/discard/issues/51).
</details>

I'd say to wait to see if we receive some feedback on that issue otherwise, we can safely merge this.

PR #3203, opened 2019-05-09
Fix partner image

jarednorman:
Not sure what happened, but the Super Good logo disappeared from Open Collective. It has been reuploaded, and this just updates the URL in the README. You can see [here](https://github.com/jarednorman/solidus/tree/super-good) that it is actually fixed.

**Checklist:**


kennyadsl:
I think we should start hosting those image in some repository here on GH instead of relying on OC, which apparently changes paths very often.

PR #3206, opened 2019-05-10
Use a rails application template for Heroku + example-app

elia:
**Description**

The current heroku deploy button is partially broken and uses a custom buildpack. 
This PR introduces a reusable [rails application template](https://guides.rubyonrails.org/rails_application_templates.html) that can be used to generate and update a demo-app (presumably published at https://github.com/solidusio/solidus-example-app).

The app will have its own app.json and will be a good base for deploys to heroku, but also
for forks. Heroku deploys should be already on the right path for production use (cache, images, etc.)

The template can also be used as an executable to build the official demo app:

```bash
~/Code/solidus $ cd ..
# the default APP_NAME is solidus-example-app
~/Code $ solidus/bin/rails-application-template [APP_NAME]
# …generates the app…
```

**Demo app**

The resulting code is available here: https://github.com/nebulab/solidus-example-app
I just had to change the heroku button url to point at the _nebulab_ org instead of _solidusio_, other than that it's straight out of the template.

See it in action: https://elia-solidus-demo.herokuapp.com

**Screenshots**

image

image

image

**Checklist:**


bin/rails-application-template
```diff
@@ -0,0 +1,139 @@
...
+
+# This file can both be used as rails application template and as executable
+if $0 == __FILE__
+  require 'shellwords'
+  app_name = ARGV.shift || default_app_name
+  options = ARGV
+  options.unshift "--database=postgresql"
+  options.unshift "--force" if app_name == default_app_name
+  command = "rails", "new", app_name, "-m", __FILE__, *options
+  warn "This file is intended to be used as Rails Application Template"
```
jacobherrington: "This file is intended to be used as *a* Rails Application Template"

bin/rails-application-template
```diff
@@ -0,0 +1,139 @@
...
+  rails_command "db:create"
+  rails_command "db:migrate"
+
+  git :init
+  git add: "."
+  git commit: %{ -m 'Generate app with solidus-example-app template' }
+
+  if ENV['UPDATE_EXAMPLE_APP_REPO']
+    git remote: %{ add origin git@github.com:solidusio/solidus-example-app.git }
+    git push: %{ --force --set-upstream origin master }
```
jacobherrington: I wasn't sure about using `--force` here without asking the user for confirmation, but I talked myself into being okay with it.

elia:
> Looks good to me. This does make some assumptions as to services like Cloudinary and Memcachier but since this is meant as an example app I think those are fine choices that are easily changed by the user.

Yes, given it's going to be deployed on heroku I looked for services that had a free tier and were at the same time viable for production use (using a cache and a CDN are both very good practices for any production store).

bin/rails-application-template
elia: The idea is to have something that can be automated and requires as few interaction as possible. That's only for the chore of updating the example app after a release of course, it doesn't happen when used as an app template.

jacobherrington: Yes, and worst case scenario we just regenerate the example app. The repo we are pushing to will not contain anything we would be afraid to lose.

kennyadsl:
@elia I moved the repo from `nebulab` org to `solidusio`: https://github.com/solidusio/solidus-example-app

Are we good to go?

elia:
> Are we good to go?

I'll remove the last commit and we're good! 🚀

elia:
@kennyadsl done, and also regenerated the app to with the solidusio heroku button link 👍

PR #3210, opened 2019-05-14
Improve jbuilder serialization for Oj gem

kennyadsl:
**Description**

This PR makes sure we are always nullifying total_on_hand before passing trying to serialize it, if it is a `Float::Infinity`, which was causing [issues with `oj`](https://github.com/solidusio/solidus/pull/2495#issuecomment-365352792).

**Checklist:**

### DIFF
--- a/api/app/helpers/spree/api/api_helpers.rb
+++ b/api/app/helpers/spree/api/api_helpers.rb
@@ -183,6 +183,10 @@ module Spree
           @@variant_attributes
         end
       end
+
+      def total_on_hand_for(object)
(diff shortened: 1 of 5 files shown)


pelargir:
v2.9.1 fixes this problem. No need to backport anything.

https://github.com/rails/jbuilder/commit/821f514741a3e9102082a6e98eb59a08671f75d0

kennyadsl:
@pelargir thanks! I think we can still keep these changes, even it's no more a bug.

kennyadsl:
@ericsaupe @jacobherrington This is no more strictly required but it's a nice to have if we want to support serializing with `Oj`

PR #3212, opened 2019-05-22
Improve JS linting

aitbw:
**Description**

Following #3167, this PR aims to further improve JS linting by enabling non-offending rules, taking advantage of ESLint environments and whitelisting backend-specific global variables.

**Checklist:**

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,9 @@
 {
   "extends": "airbnb-base",
+  "env": {
+    "browser": true,
+    "jquery": true
+  },
   "rules": {
     "camelcase": 0,
     "comma-dangle": 0,
@@ -29,11 +33,9 @@
     "vars-on-top": 0
   },
   "globals": {
-    "$": "readonly",
     "accounting": "readonly",
     "addVariantFromStockLocation": "writable",
     "adjustShipmentItems": "writable",
-    "document": "readonly",
     "error": "writable",
     "flatpickr": "readonly",
     "Handlebars": "readonly",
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -1,19 +1,28 @@
...
     "camelcase": 0,
     "comma-dangle": 0,
     "dot-notation": 0,
     "func-names": 0,
-    "no-console": 0,
-    "no-extra-semi": 0,
+    "no-console": ["warn", {
+        "allow": ["warn"]
+    }],
+    "no-extra-semi": "warn",
```
mdesantis: Why a warning instead of an error for this rule? Do we expect to use extra semicolons while developing?

aitbw: Given the discussion in #2847, I think the general consensus is to keep semicolons and only remove the ones that are completely unnecessary.

As for the warning instead of an error, I think that was an arbitrary decision :sweat_smile: I feel like the warnings should be used for stuff like this, whilst errors should be used when variable shadowing, unused vars or no `var` usage issues are present, but I'm completely willing to discuss what's the better choice for this. Let me know what you think! :raised_hands:

mdesantis: @aitbw but in #2847 there isn't a discussion about extra semicolons :confused: for what I read `no-extra-semi` rule is for code like this: `;;` while the semicolons toggle rule is named `semi`, isn't it?

jacobherrington: https://eslint.org/docs/rules/no-extra-semi 

I think I'm reading the same way as @mdesantis. This is a non-blocker for me, but I would go ahead and make it an error personally.

aitbw: You're right, @mdesantis. Fixing it as we speak! :wink:

kennyadsl:
Thanks @aitbw, you are great!

PR #3215, opened 2019-05-29
Add preferred_reimbursement_type_id as permitted attributes

ibudiallo:
**Description**

When using the api, we can create return authorizations but cannot set the reimbursement type. The reason being the field `preferred_reimbursement_type_id` was not included in the list of **permitted attributes.**

**Checklist:**

### DIFF
--- a/api/spec/requests/spree/api/return_authorizations_controller_spec.rb
+++ b/api/spec/requests/spree/api/return_authorizations_controller_spec.rb
@@ -17,13 +17,21 @@ module Spree
       it "can create a new return authorization" do
         stock_location = FactoryBot.create(:stock_location)
         reason = FactoryBot.create(:return_reason)
+        reimbursement = FactoryBot.create(:reimbursement_type)
+        unit = FactoryBot.create(:inventory_unit)
         rma_params = { stock_location_id: stock_location.id,
                        return_reason_id: reason.id,
+                       return_items_attributes: [{
+                         inventory_unit_id: unit.id,
+                         preferred_reimbursement_type_id: reimbursement.id,
+                       }],
                        memo: "Defective" }
         post spree.api_order_return_authorizations_path(order), params: { order_id: order.number, return_authorization: rma_params }
         expect(response.status).to eq(201)
         expect(json_response).to have_attributes(attributes)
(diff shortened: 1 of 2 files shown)


ibudiallo:
@ericsaupe I've added the field in the spec. Thanks

api/spec/requests/spree/api/return_authorizations_controller_spec.rb
```diff
@@ -17,8 +17,14 @@ module Spree
...
         stock_location = FactoryBot.create(:stock_location)
         reason = FactoryBot.create(:return_reason)
+        reimbursement = FactoryBot.create(:reimbursement_type)
+        unit = FactoryBot.create(:inventory_unit)
         rma_params = { stock_location_id: stock_location.id,
                        return_reason_id: reason.id,
+                       return_items_attributes: [{
+                         inventory_unit_id: unit.id,
+                         preferred_reimbursement_type_id: reimbursement.id,
+                       }],
```
spaghetticode: @ibudiallo I can see you changed the setup part of the example, adding also the `preferred_reimbursement_type_id` parameter, but I think the feature is not actually exercised in any of the subsequent expectations. 

I think adding something like: 
```ruby
        return_authorization = Spree::ReturnAuthorization.last
        expect(return_authorization.return_items.first.preferred_reimbursement_type).to eql reimbursement
```
would ensure that the new code is actually working, what do you think?

ibudiallo: Completely make sense. I've added the test now.

kennyadsl:
Thank you @ibudiallo, just one last thing: can you please squash commits into a single one?

ibudiallo:
Thanks guys. I squashed my commits.

kennyadsl:
Hey @ibudiallo I'd love to merge this PR but it should have a single commit. Maybe you rebased against your local copy of solidus, which is a bit behind the upstream one?

I suggest you to add a new upstream remote and redo the rebase:

```bash
git remote add upstream git@github.com:solidusio/solidus.git
git fetch -a upstream
git rebase upstream/master
```

This would clean-up the branch, I'm available here or on our Slack Workspace for further help if needed. Thanks!

PR #3216, opened 2019-05-30
Fix non thread safe gateway initialization

bitberryru:
Hi!

Noticed non thread safe code here https://github.com/solidusio/solidus/blob/e7260a27a7c292908a835f374d5ba73fe7284cd0/core/app/models/spree/payment_method.rb#L134
It can cause problems with payments. And this kind of setup should be used only in initializers.

Instead of this I instantiate gateway object with test option based on :server and :test_mode.

btw, :test_mode option has no effect in active merchant gateways.

### DIFF
--- a/core/app/models/spree/payment_method.rb
+++ b/core/app/models/spree/payment_method.rb
@@ -130,9 +130,13 @@ module Spree
     def gateway
       gateway_options = options
       gateway_options.delete :login if gateway_options.key?(:login) && gateway_options[:login].nil?
-      if gateway_options[:server]
-        ActiveMerchant::Billing::Base.mode = gateway_options[:server].to_sym
-      end
+
+      # All environments except production considered to be test
+      test_server = gateway_options[:server] != 'production'
+      test_mode = gateway_options[:test_mode]
+
+      gateway_options[:test] = (test_server || test_mode)
+
       @gateway ||= gateway_class.new(gateway_options)
     end
     alias_method :provider, :gateway
--- a/core/app/models/spree/billing_integration.rb
+++ b/core/app/models/spree/billing_integration.rb
@@ -9,8 +9,13 @@ module Spree
 
     def gateway
       integration_options = options
-      ActiveMerchant::Billing::Base.integration_mode = integration_options[:server].to_sym
-      integration_options[:test] = true if integration_options[:test_mode]
+
+      # All environments except production considered to be test
+      test_server = integration_options[:server] != 'production'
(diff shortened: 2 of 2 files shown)


core/app/models/spree/payment_method.rb
```diff
@@ -130,9 +130,13 @@ def find_with_destroyed(*args)
...
-        ActiveMerchant::Billing::Base.mode = gateway_options[:server].to_sym
-      end
+
+      # All environments except production considered to be test
+      test_server = gateway_options[:server] != 'production'
+      test_mode = gateway_options[:test_mode]
+
+      gateway_options[:test] = (test_server || test_mode)
+
       @gateway ||= gateway_class.new(gateway_options)
```
mdesantis: Current implementation is not thread safe, but `||=` operator isn't thread safe neither: https://tenderlovemaking.com/2011/10/20/connection-management-in-activerecord.html

We should use a mutex in order to make it thread safe: https://lucaguidi.com/2014/03/27/thread-safety-with-ruby/

bitberryru: Wow, the question here is not in `||=` operator used to set instance variable :)) but that you can't change the class variable (global) and expect that the gateway instance will use exactly this value. In next call this class variable will be overwritten and all gateway instances will use it, not one from their preference.

mdesantis: Oh ok, I got it now :) my apologies!

spaghetticode:
@bitberryru I think the changes make sense, thank you! 👏 

For reference and understandability, as the changes deal with the `activemerchant` gem codebase, I'm cut & pasting the relevant code from the currently bundled version:

```ruby
# from lib/active_merchant/billing/gateway.rb

# Initialize a new gateway.
#
# See the documentation for the gateway you will be using to make sure there are no other
# required options.
def initialize(options = {})
  @options = options
end

# Are we running in test mode?
def test?
  (@options.has_key?(:test) ? @options[:test] : Base.test?)
end
```

bitberryru:
@kennyadsl Yes, it should not break any store, but when updating it is still worth checking the settings of payment methods :) Only payment methods where server preference set to production and test_mode turned off will work in "production" mode.

> I also noticed we are doing something similar here but I'm not sure we are actually using that class anywhere, maybe we can deprecate it... 🤔

Wow, I didn't see it before and never use 😃 This code must also be fixed, if you are not going to delete this class, I can do it.

bitberryru:
`Spree::BillingIntegration` also fixed.

PR #3217, opened 2019-05-30
Change payment method partial name convention

bitberryru:
Hi!

I propose to use underscore in the name of the payment method partial. It is more convenient and clearer. What do you think, guys?


mdesantis:
I agree :)

spaghetticode:
👍 This is definitely an improvement, but I do have some concerns: besides some failing specs that should be addressed,  this may be a breaking change for some stores and/or extensions, as some partials will possibly need to be renamed?

bitberryru:
Once everyone agrees I'll fix the tests.

> this may be a breaking change for some stores and/or extensions, as some partials will possibly need to be renamed

Yes

tvdeyen:
I won’t consider this breaking, but a fix instead. The only difference is that a payment method name "Credit Card" would currently be `credit card` and now `credit_card`. The former will most likely lead to bugs. Since this is a rather new feature and even aliased to the former `method_type` method I think this is save.

bitberryru:
> I won’t consider this breaking, but a fix instead. The only difference is that a payment method name "Credit Card" would currently be `credit card` and now `credit_card`. The former will most likely lead to bugs. Since this is a rather new feature and even aliased to the former `method_type` method I think this is save.

No, it is 'gateway' already :)
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment_method/credit_card.rb#L17

tvdeyen:
`gateway` is the fallback for all payment methods that do neither implement `method_type` nor `partial_name`

https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment_method.rb#L204

Anyway. This fix is good and appreciated. Thanks for the work.

PR #3220, opened 2019-06-07
Allow only stubbed changes to `Spree::Config` in specs

### LINKED ISSUE #3219: Improve specs reliability when changing `Spree::Config.track_inventory_level`
I recently imported some specs from Solidus in my codebase in order to broaden test coverage locally, before applying some patches to the original code. 

I only later realized that some of these specs changed the value of a relevant global setting without putting it back to the original value, thus creating random failures in other subsequent specs that relied on the original value of that parameter.

While this does not seem to affect Solidus test suite, I think the issue is still relevant as I believe it's generally good practice, after changing a global parameter, to put it back to the original value.
One example is [this spec](https://github.com/solidusio/solidus/blob/master/core/spec/models/spree/variant_spec.rb#L791):

```ruby
    context "inventory levels globally not tracked" do
      before { Spree::Config.track_inventory_levels = false }

      it "includes all variants" do
        expect( subject ).to include(in_stock_variant, backordered_variant, out_of_stock_variant)
      end
    end
```

and, more generally, all the times `Spree::Config.track_inventory_levels` is changed in a `before` block but not returned to the original value.

spaghetticode:
By freezing `Spree::Config.preference_store` and allowing changes to `Spree::Config` settings only via RSpec stubs we ensure no global settings state change will leak outside the lifecycle of each spec example.

Also, stubbing changes to `Spree::Config` removes the necessity to reset these settings before each spec is run, which in the past encouraged the _not-so-good_ practice of changing the global state and not returning it to the original values.

When writing tests, the preferred way to change `Spree::Config` settings is by using the helper method `stub_spree_preferences`:

```ruby
stub_spree_preferences(currency: 'EUR', track_inventory_levels: true)
```

The helper method `with_unfrozen_spree_preference_store ` can be used when for exceptional reasons one needs to use the unfrozen version of the `preference_store`. This helper method has the advantage of automatically returning the configuration state to the original value:

```ruby
around do |example|
  with_unfrozen_spree_preference_store do
    example.run
  end
end
```

resolve #3219 

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/preferences.rb
+++ b/core/lib/spree/testing_support/preferences.rb
@@ -49,7 +49,9 @@ module Spree
       #   Solidus #3219 for more details and motivations.
       def stub_spree_preferences(preferences)
         preferences.each do |name, value|
-          allow(Spree::Config).to receive(:[]).and_call_original
+          if Spree::Config.method(:[]).owner >= Spree::Config.class
+            allow(Spree::Config).to receive(:[]).and_call_original
+          end
           allow(Spree::Config).to receive(:[]).with(name) { value }
           allow(Spree::Config).to receive(name) { value }
         end
--- a/core/lib/spree/testing_support/preferences.rb
+++ b/core/lib/spree/testing_support/preferences.rb
@@ -56,6 +56,27 @@ module Spree
           allow(Spree::Config).to receive(name) { value }
         end
       end
+
+      # This method allows to temporarily switch to an unfrozen Spree::Config preference
+      # store with all proper preferences values set.
+      #
+      # It should be used sparingly, only when `stub_spree_preferences` would not work.
+      #
+      # @example Temporarily switch to an unfrozen store and change some preferences:
+      #   with_unfrozen_spree_preference_store do
+      #     Spree::Config.currency = 'EUR'
+      #     Spree::Config.track_inventory_levels = false
+      #
+      #     expect(Spree::Config.currency).to eql 'EUR'
+      #   end
+      # @see Spree::TestingSupport::Preferences#stub_spree_preferences
+      def with_unfrozen_spree_preference_store
+        frozen_store = Spree::Config.preference_store
+        Spree::Config.preference_store = Spree::Config[:unfrozen_preference_store].dup
+        yield
+      ensure
+        Spree::Config.preference_store = frozen_store
+      end
     end
   end
 end
--- a/core/lib/spree/testing_support/preferences.rb
+++ b/core/lib/spree/testing_support/preferences.rb
@@ -29,6 +29,31 @@ module Spree
         find("#preferences_#{preference}")['checked'].should be false
         Spree::Config[preference].should be false
       end
+
+      # This is the preferred way for changing temporarily Spree preferences during
+      # tests via stubs, without changing the actual values stored in Spree::Config.
+      #
+      # By using stubs no global preference change will leak outside the lifecycle
+      # of each spec example, avoiding possible unpredictable side effects.
+      #
+      # This method  may be used for stubbing one or more different preferences
+      # at the same time.
+      #
+      # @param [Hash] preferences names and values to be stubbed
+      #
+      # @example Stubs `currency` and `track_inventory_levels` preferences
+      #   stub_spree_preferences(currency: 'EUR', track_inventory_levels: false)
+      #   expect(Spree::Config.currency).to eql 'EUR'
+      #
+      # @see https://github.com/solidusio/solidus/issues/3219
+      #   Solidus #3219 for more details and motivations.
+      def stub_spree_preferences(preferences)
+        preferences.each do |name, value|
+          allow(Spree::Config).to receive(:[]).and_call_original
+          allow(Spree::Config).to receive(:[]).with(name) { value }
+          allow(Spree::Config).to receive(name) { value }
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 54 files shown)


core/spec/models/spree/promotion_handler/shipping_spec.rb
```diff
@@ -73,6 +73,8 @@ module PromotionHandler
       end
 
       context 'with a custom shipping action' do
+        let!(:default_actions) { Spree::Config.environment.promotions.shipping_actions }
```
kennyadsl: Is it not possible to stub this one as well?

frontend/spec/features/currency_spec.rb
```diff
@@ -14,7 +14,9 @@
     click_link "RoR Mug"
     click_button "Add To Cart"
     # Now that we have an order...
-    Spree::Config[:currency] = "AUD"
-    visit spree.root_path
+    with_unfrozen_spree_preference_store do
```
kennyadsl: I think we should document in the commit message why it's not possible to stub this preference (mainly because is not entirely clear to me yet :P ).

core/lib/spree/testing_support/preferences.rb
```diff
@@ -29,6 +29,33 @@ def assert_preference_unset(preference)
...
+          allow(Spree::Config).to receive(name) { value }
+        end
+      end
+
+      def with_unfrozen_spree_preference_store
+        frozen_store = Spree::Config.preference_store
+        Spree::Config.preference_store = Spree::Config[:unfrozen_preference_store]
+        yield
+      ensure
+        reset_spree_preferences
```
kennyadsl: I'm not sure to get why this `reset_spree_preferences` is needed. If we are replacing the preference store with the frozen one generated at the test suite boot we should have it in the same state without the need of a reset, right?

core/spec/models/spree/promotion_handler/shipping_spec.rb
spaghetticode: 👍 yes it is, thanks for noticing it!

core/lib/spree/testing_support/preferences.rb
spaghetticode: 👍 Especially after my latest update, it's not needed anymore. 

One thing worth noticing that surprised me is that `reset_spree_preferences` returns configuration values to the very default ones for the preference. 

For example, `Spree::Config.mails_from` is returned to [this default value](https://github.com/solidusio/solidus/blob/3af846ba42f517c218d5ec010f42471acee2fc6b/core/lib/spree/app_configuration.rb#L242) instead of the one set in [dummy_app.rb]( https://github.com/solidusio/solidus/blob/3af846ba42f517c218d5ec010f42471acee2fc6b/core/lib/spree/testing_support/dummy_app.rb#L114). 

I think this is another good reason for avoiding to use it, or deprecate it as you already suggested.

frontend/spec/features/currency_spec.rb
jacobherrington: I am also curious about this one.

core/lib/spree/testing_support/preferences.rb
```diff
@@ -29,6 +29,33 @@ def assert_preference_unset(preference)
         find("#preferences_#{preference}")['checked'].should be false
         Spree::Config[preference].should be false
       end
+
+      def stub_spree_preferences(preferences)
```
jacobherrington: I would love to see a comment explaining why we are doing this, including perhaps a link to #3219.

frontend/spec/features/currency_spec.rb
spaghetticode: Being an old regression spec I wanted to keep the code as similar as the original as possible in order to avoid making the spec irrelevant by overlooking some details, so I did not stubbed it in the beginning, but after running this spec against the original code for spree/spree#2340 I think this can be safely stubbed as well. See also the fix for issue 2340: https://github.com/spree/spree/commit/58585fc55cad6027a4fa01a747e14999d5dc49ae

As far as I understand this spec basically ensured that this old line would find no order: 

```ruby
current_order = Spree::Order.find_by_id_and_currency(session[:order_id], current_currency, :include => :adjustments)
```

by changing the backend currency, thus making the line that followed (before the fix) raise an exception:

```ruby
current_order.last_ip_address = ip_address
```

I think this spec is not much relevant anymore, as that part of the code is long gone, so I guess it may be safely removed. What do you think? Maybe in another PR to keep things clean and focused here.

core/lib/spree/testing_support/preferences.rb
spaghetticode: @jacobherrington I added some Rdoc/Yard documentation to the new helper methods in `preferences.rb`

spaghetticode:
@kennyadsl I agree with you on deprecating `reset_spree_preferences`, so I've just added a commit for that.

core/spec/lib/spree/core/testing_support/preferences_spec.rb
```diff
@@ -4,7 +4,23 @@
 
 RSpec.describe Spree::TestingSupport::Preferences do
   describe 'resetting the app configuration' do
+    around do |example|
+      with_unfrozen_spree_preference_store do
+        behavior = Spree::Deprecation.behavior
```
kennyadsl: I think you should use  `Spree::Deprecation.silence {}` instead here, right?

spaghetticode: @kennyadsl 👍 that's much better, thank you!

PR #3226, opened 2019-06-11
Avoid removing product_properties and properties when discard a product

jtapia:
#### Description
Removing(soft-delete) `products` is deleting the `product_properties`, but when you want to bring back the product, the `product_properties` are gone, this change add the soft-delete to `product_property` model

### DIFF
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -24,7 +24,7 @@ module Spree
     after_discard do
       variants_including_master.discard_all
       self.product_option_types = []
-      self.product_properties = []
+      product_properties.discard_all
       self.classifications = []
       self.product_promotion_rules = []
     end
--- a/core/app/models/spree/product_property.rb
+++ b/core/app/models/spree/product_property.rb
@@ -1,12 +1,17 @@
 # frozen_string_literal: true
 
+require 'discard'
+
 module Spree
   class ProductProperty < Spree::Base
     include Spree::OrderedPropertyValueList
 
     acts_as_list scope: :product
 
-    belongs_to :product, touch: true, class_name: 'Spree::Product', inverse_of: :product_properties
+    include Discard::Model
+    self.discard_column = :deleted_at
+
+    belongs_to :product, -> { with_deleted }, touch: true, class_name: 'Spree::Product', inverse_of: :product_properties
     belongs_to :property, class_name: 'Spree::Property', inverse_of: :product_properties
 
     self.whitelisted_ransackable_attributes = ['value']
--- a/core/spec/models/spree/product_spec.rb
+++ b/core/spec/models/spree/product_spec.rb
@@ -477,6 +477,17 @@ RSpec.describe Spree::Product, type: :model do
         expect(product.possible_promotions).to include(promotion)
       end
     end
+
+    context "#discard" do
+      let(:product_property) { create(:product_property) }
+      let(:product) { product_property.product }
+
+      it "does not delete product properties objects" do
+        product.discard
+        expect(product.product_properties).to be_empty
+        expect(product_property.reload.deleted_at).not_to be_nil
+      end
+    end
   end
 
   context "#images" do
--- a/api/app/controllers/spree/api/product_properties_controller.rb
+++ b/api/app/controllers/spree/api/product_properties_controller.rb
@@ -48,7 +48,7 @@ module Spree
       def destroy
         if @product_property
           authorize! :destroy, @product_property
-          @product_property.destroy
+          @product_property.discard
           respond_with(@product_property, status: 204)
         else
           invalid_resource!(@product_property)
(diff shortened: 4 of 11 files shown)


core/app/models/spree/product.rb
```diff
@@ -24,7 +24,7 @@ class Product < Spree::Base
     after_discard do
       variants_including_master.discard_all
       self.product_option_types = []
-      self.product_properties = []
+      self.product_properties.discard_all
```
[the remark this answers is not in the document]
kennyadsl: Probably Hound is right here, I think we can avoid `self` now.

core/app/models/spree/product_property.rb
```diff
@@ -1,12 +1,20 @@
...
+
 module Spree
   class ProductProperty < Spree::Base
     include Spree::OrderedPropertyValueList
 
     acts_as_list scope: :product
 
-    belongs_to :product, touch: true, class_name: 'Spree::Product', inverse_of: :product_properties
+    acts_as_paranoid
+    include Spree::ParanoiaDeprecations
```
kennyadsl: If it wasn't under Paranoia before we can avoid calling `acts_as_paranoid` and its deprecations, or is it adding some other value that I can't see at the moment?

core/lib/spree/testing_support/factories/product_property_factory.rb
```diff
@@ -7,5 +7,6 @@
   factory :product_property, class: 'Spree::ProductProperty' do
     product
     property
+    deleted_at { nil }
```
kennyadsl: I don't think this is needed, should be nil by default, right?

core/spec/models/spree/product_spec.rb
```diff
@@ -477,6 +477,18 @@ class Extension < Spree::Base
         expect(product.possible_promotions).to include(promotion)
       end
     end
+
+    context "#discard" do
+      let(:product_property) { create(:product_property) }
+      let(:product) { product_property.product }
+
+      it "should not delete product properties objects" do
```
kennyadsl: Can you please avoid "should" here? `it "does not delete...` form is preferred.

core/db/migrate/20160101010000_solidus_one_four.rb
```diff
@@ -426,10 +426,12 @@ def up
       t.integer "property_id"
       t.datetime "created_at", precision: 6
       t.datetime "updated_at", precision: 6
+      t.datetime "deleted_at", precision: 6
```
kennyadsl: I think we can't change this migration but we need a new one, otherwise who updates Solidus won't take this database change.

core/app/models/spree/product.rb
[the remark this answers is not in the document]
jtapia: done https://github.com/solidusio/solidus/pull/3226/files#diff-327ceda9367add32373743b9f3692285R27

core/app/models/spree/product_property.rb
jtapia: done https://github.com/solidusio/solidus/pull/3226/files#diff-df2aa0068c69d3f1a851f4d2e7ed3e20R11

core/spec/models/spree/product_spec.rb
jtapia: done https://github.com/solidusio/solidus/pull/3226/files#diff-72c8de7678be26819204ff95fe964da9R485

core/lib/spree/testing_support/factories/product_property_factory.rb
jtapia: done

core/db/migrate/20160101010000_solidus_one_four.rb
jtapia: I added it here because of this error https://circleci.com/gh/solidusio/solidus/12317, do you know if I need to do something else to detect the new migration?

jtapia:
@kennyadsl I think there's no way to identify a previous deleted product property, but we still have the hard-delete(destroy) option and now the `discard` one, what do you think?

kennyadsl:
I'm fine with soft-deleting product properties, I was just thinking that the requirement you described for implementing this feature

> Removing(soft-delete) products is deleting the product_properties, but when you want to bring back the product, the product_properties are gone

is not totally satisfied since to bring product properties back you when you restore a soft-deleted product, you'll need to manually select which ones you want to restore. We have the same for `Spree::Variant` and other models, I think soft-deletion is not done to bring things back but to avoid losing references to records in the database, since they have important information for data reporting.

Another thought: what should happen if you delete a `Spree::Property` instead?

waiting-for-dev:
Thanks, @jtapia. As @kennyadsl said, it's kind of a catch-22 situation, as we can't easily discern which records we want to bring back. On top of that, it probably goes beyond the business need that soft deletion addresses. Closing it, but let me thank you again for your work on it.

PR #3230, opened 2019-06-17
Purify guides search terms before using them

kennyadsl:
**Description**

We had an XSS vulnerability that allows injecting code in the page. Purifying the string that is passed to the search engine fixes the issue.

**Checklist:**
- ~I have updated Guides and README accordingly to this change (if needed)~
- ~I have added tests to cover this change (if needed)~


kennyadsl:
Merging without respecting the 24-hour waiting policy for security reasons.

PR #3232, opened 2019-06-17
Support class reloading for Event Subscribers

elia:
**Description**

Before this change, editing an instance of Spree::Event::Subscriber would
result in not being reloaded or being subscribed twice.

With this change each includer of Event::Subscriber is registered by its name
and later unsubscribed just before class unload (using old module instance),
then resubscribed after the new code is loaded.

_Credit: This work has been done together with the estimable @AlessioRocco_

**Checklist:**


core/lib/generators/spree/install/install_generator.rb
```diff
@@ -79,35 +79,55 @@ def create_overrides_directory
...
+    initializer 'solidus.decorators' do |app|
+      config.to_prepare do
+        Dir.glob(Rails.root.join('app/**/*_decorator*.rb)) do |path|
+          require_dependency(path)
+        end
+      end
+    end
+
+    # Load application's view overrides
+    initializer 'solidus.decorators' do |app|
```
kennyadsl: `decorators` -> `overrides`

core/lib/spree/app_configuration.rb
```diff
@@ -220,6 +220,18 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Boolean] Don't escape HTML of product descriptions. (default: +false+)
     preference :show_raw_product_description, :boolean, default: false
 
+    # @!attribute [r] subscriber_names
```
kennyadsl: I'm not sure this is the right place for these methods. They look like global containers, while here we should only list configurable preferences.

core/lib/spree/event/subscriber.rb
```diff
@@ -26,6 +26,8 @@ def self.included(base)
 
         base.mattr_accessor :event_actions
         base.event_actions = {}
+
+        Spree::Config.subscriber_names << base.name
```
kennyadsl: What is name here?

elia:
@tvdeyen thanks for the early heads up!

> From what I can tell in the moment this looks like an implemenation issue we have with the current subscribers implementation, right? If so, could we try to fix those instead?

That's correct, namely the problem is with AS::Notifications, although, that's likely something that will be needed for any other subscriber implementation that organizes subscribers in different classes/modules across different files (which seems a pretty standard way of doing it). And allowing an implementation to provide a list of constant-names to be re-subscribed at each reload seems a general-enough api.

To better exemplify the problem I'll write some psedo-code that will hopefully make it clear for anyone not super-familiar with rails reloading, the following is based on the current implementation / examples.

_The app starts and subscribes all subscribers_
```rb
# config/initializers/spree_event_subscribers.rb
OrderSubscriber.subscribe!
PaymentSubscriber.subscribe!
```

elia:
@tvdeyen thanks for the early heads up!

---

> I really don't like to add this amount of code into every store that gets created out there. Could you elaborate why this change is necessary and maybe discuss another way of fixing those issues?

I'll defer to the section below for explaining how this is necessary for the current subscriber implementation (`AS::Notifications`) and most probably for alternative ones. I also agree that keeping that amount of code there is not ideal. What about leaving just the glob and subscribe with an internal API? E.g.:

```rb
Dir.glob(Rails.root.join('app/subscribers/*_subcriber.rb)) do |path|
  Spree::Event.register_subscriber(path)
end
```

Alternatively we can also define the glob in a configuration, e.g. `Spree::Config.events.files = Dir.glob(Rails.root.join('app/subscribers/*_subcriber.rb))`, and do everything else internally.

---

> From what I can tell in the moment this looks like an implemenation issue we have with the current subscribers implementation, right? If so, could we try to fix those instead?

That's correct, namely the problem is with AS::Notifications, although, that's likely something that will be needed for any other subscriber implementation that organizes subscribers in different classes/modules across different files (which seems a pretty standard way of doing it). And allowing an implementation to provide a list of constant-names to be re-subscribed at each reload seems a general-enough api.

To better exemplify the problem I'll write some psedo-code that will hopefully make it clear for anyone not super-familiar with rails reloading, the following is based on the current implementation / examples.

I apologize if this is known stuff, but I thought it might be a useful refresher for some 🙂

### Scenario 1: loading subscribers with an initializer

_The app starts and subscribes all subscribers_
```rb
# config/initializers/spree_event_subscribers.rb
OrderSubscriber.subscribe! # Rails will load app/subscribers/order_subscriber.rb
PaymentSubscriber.subscribe! # Rails will load app/subscribers/payment_subscriber.rb
```

_Rails will load the missing constants in this way_

```rb
unless const_defined? :OrderSubscriber
  load 'app/subscribers/order_subscriber.rb'
  mark_as_autoloaded :OrderSubscriber, 'app/subscribers/order_subscriber.rb'
  p OrderSubscriber # => #<Module:123>
end
```

_AS::Notification will store the contents of `OrderSubscriber` inside a global list of subscribers_

```rb
AS::Notification.subscribers << OrderSubscriber # that's an instance of Module: #<Module:123>
```

_The developer changes the contents of app/subscribers/order_subscriber.rb_

```rb
if file_changed? 'app/subscribers/order_subscriber.rb'
  remove_const :OrderSubscriber
end

# at this point AS::Notification still holds a reference to the old version of OrderSubscriber  
p AS::Notification.subscribers # => [#<Module:123>, …]
```

_The new version of OrderSubscriber is never loaded since the constant is not referenced anywhere except in the initializer_

### Scenario 2: loading subscribers with `to_prepare`

_Goes like scenario 1, but `order_subscriber.rb` is loaded within the `to_prepare` block which is called at initialization and after each application reload_

_When the app reloads the constant is removed as above, and redefined with the new implementation of OrderSubscriber, let's call it #<Module:456>, which is a completely different instance of Module_

_The problem remains with the global `AS::Notifications` global list that still holds the old reference, plus the new reference: `[#<Module:123>, …, #<Module:456>, …]`_

_The result in this case is that we have duplicate subscribers with different implementations_

core/lib/spree/app_configuration.rb
elia: I'll move them to `Spree::Config.events`

core/lib/spree/event/subscriber.rb
elia: It's `Module#name`, e.g. `Foo::Bar.name # => "Foo::Bar"`

core/lib/generators/spree/install/install_generator.rb
```diff
@@ -79,35 +79,40 @@ def create_overrides_directory
     end
 
     def configure_application
-      application <<-APP
-
-    config.to_prepare do
-      # Load application's model / class decorators
-      Dir.glob(File.join(File.dirname(__FILE__), "../app/**/*_decorator*.rb")) do |c|
-        require_dependency(c)
+      application <<-RUBY
```
kennyadsl: I'm ok with keeping these changes (both commits for this file) but maybe now it's better to move them into a separate PR. They don't seem to be related to this one anymore, right?

elia: Done ✅

core/lib/spree/core/engine.rb
```diff
@@ -45,8 +45,16 @@ class Engine < ::Rails::Engine
         Migrations.new(config, engine_name).check
       end
 
-      initializer 'spree.core.subscribe_event_mailer_processor' do
-        Spree::Event::Processors::MailerProcessor.subscribe!
+      # Setup Event Subscribers
+      config.paths.add "app/subscribers", eager_load: true, glob: "**/*_subscriber.rb"
```
kennyadsl: @elia does this apply only to `app/subscribers` of Solidus core or it will add to the path also subscribers defined into extensions and host applications?

elia: @kennyadsl good call, after reading a bunch of rails code and some trial and error I verified I misinterpreted how paths worked. I'll push an update that fixes the behavior for both engine and app code.

kennyadsl:
We are discussing this IRL and we decided to try to use an approach similar to calculators so we'll probably end up having a list of subscribers classes into a preference like

```ruby
spree.config.event.subscribers = [
  Spree::Core::Event::MailSubscriber,
  SpreeExtensions::AnotherSubscriber,
  MyApp::YetAnotherSubscriber
]
```

This looks like a good solution for a couple of reasons:

1. it is something easy to manage to add change or remove classes in the list
2. it is a well-known pattern for Solidus users since we are using it in other places. If we find a more Rails-way solution for this kind of object we can easily replace everything and deprecate setting the configuration.

elia:
@kennyadsl @tvdeyen ready for a new (final?) pass 🔍🙂

PR #160, opened 2015-07-09
Add bourbon and replace spin.js with simple CSS spinner

jhawthorn:

PR #3236, opened 2019-06-21
Add support for Rails 6

aldesantis:
**Description**

Adds Rails 6 support. 🍾 

**Progress:**

Before we can merge this, the following needs to be addressed:

- [x] solidus_auth_devise needs to use `secret_key_base` instead of `secret_token` (see [#160](https://github.com/solidusio/solidus_auth_devise/pull/160)).
- [x] Ransack will only support Rails 6 after a stable version is released (see [their readme](https://github.com/activerecord-hackery/ransack#rails-6-support)). In the meantime, I've forked and fixed Ransack so we can work on Solidus.
- [x] [Awesome Nested Set](https://github.com/collectiveidea/awesome_nested_set) needs a new version released for Rails 6 support.
- [ ] RSpec needs to release a stable 4.x release. The beta works but we'd rather not have it in the Gemfile.

**Checklist:**

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,8 +46,8 @@ group :frontend do
 end
 
 group :backend do
-  gem 'teaspoon', require: false
-  gem 'teaspoon-mocha', require: false
+  gem 'teaspoon', github: 'jejacks0n/teaspoon', require: false
+  gem 'teaspoon-mocha', github: 'jejacks0n/teaspoon', require: false
 end
 
 group :utils do
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ group :backend, :frontend, :core, :api do
   gem 'database_cleaner', '~> 1.3', require: false
   gem 'factory_bot_rails', '~> 4.8', require: false
   gem 'rspec-activemodel-mocks', '~> 1.1', require: false
-  gem 'rspec-rails', '~> 3.7', require: false
+  gem 'rspec-rails', '~> 4.0.0.beta2', require: false
   gem 'simplecov', require: false
   gem 'with_model', require: false
   gem 'rails-controller-testing', require: false
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ source 'https://rubygems.org'
 group :backend, :frontend, :core, :api do
   gemspec require: false
 
-  rails_version = ENV['RAILS_VERSION'] || '~> 5.2.0'
+  rails_version = ENV['RAILS_VERSION'] || '~> 6.0.0'
   gem 'rails', rails_version, require: false
 
   platforms :ruby do
--- a/core/app/models/spree/product/scopes.rb
+++ b/core/app/models/spree/product/scopes.rb
@@ -2,159 +2,162 @@
 
 module Spree
   class Product < Spree::Base
-    cattr_accessor :search_scopes do
-      []
-    end
+    module Scopes
+      def self.prepended(base)
+        base.class_eval do
+          cattr_accessor :search_scopes do
+            []
+          end
 
-    def self.add_search_scope(name, &block)
-      singleton_class.send(:define_method, name.to_sym, &block)
-      search_scopes << name.to_sym
-    end
+          def self.add_search_scope(name, &block)
+            singleton_class.send(:define_method, name.to_sym, &block)
+            search_scopes << name.to_sym
+          end
 
-    def self.property_conditions(property)
-      properties = Property.table_name
-      case property
-      when String   then { "#{properties}.name" => property }
-      when Property then { "#{properties}.id" => property.id }
-      else { "#{properties}.id" => property.to_i }
-      end
-    end
+          def self.property_conditions(property)
+            properties = Property.table_name
+            case property
+            when String   then { "#{properties}.name" => property }
+            when Property then { "#{properties}.id" => property.id }
+            else { "#{properties}.id" => property.to_i }
+            end
+          end
 
-    scope :ascend_by_updated_at, -> { order(updated_at: :asc) }
-    scope :descend_by_updated_at, -> { order(updated_at: :desc) }
-    scope :ascend_by_name, -> { order(name: :asc) }
-    scope :descend_by_name, -> { order(name: :desc) }
+          scope :ascend_by_updated_at, -> { order(updated_at: :asc) }
+          scope :descend_by_updated_at, -> { order(updated_at: :desc) }
+          scope :ascend_by_name, -> { order(name: :asc) }
+          scope :descend_by_name, -> { order(name: :desc) }
 
-    add_search_scope :ascend_by_master_price do
-      joins(master: :default_price).order(Spree::Price.arel_table[:amount].asc)
-    end
+          add_search_scope :ascend_by_master_price do
+            joins(master: :default_price).order(Spree::Price.arel_table[:amount].asc)
+          end
 
-    add_search_scope :descend_by_master_price do
-      joins(master: :default_price).order(Spree::Price.arel_table[:amount].desc)
-    end
+          add_search_scope :descend_by_master_price do
+            joins(master: :default_price).order(Spree::Price.arel_table[:amount].desc)
+          end
 
-    add_search_scope :price_between do |low, high|
-      joins(master: :default_price).where(Price.table_name => { amount: low..high })
-    end
+          add_search_scope :price_between do |low, high|
+            joins(master: :default_price).where(Price.table_name => { amount: low..high })
+          end
 
-    add_search_scope :master_price_lte do |price|
-      joins(master: :default_price).where("#{price_table_name}.amount <= ?", price)
-    end
+          add_search_scope :master_price_lte do |price|
+            joins(master: :default_price).where("#{price_table_name}.amount <= ?", price)
+          end
 
-    add_search_scope :master_price_gte do |price|
-      joins(master: :default_price).where("#{price_table_name}.amount >= ?", price)
-    end
+          add_search_scope :master_price_gte do |price|
+            joins(master: :default_price).where("#{price_table_name}.amount >= ?", price)
+          end
 
-    # This scope selects products in taxon AND all its descendants
-    # If you need products only within one taxon use
-    #
-    #   Spree::Product.joins(:taxons).where(Taxon.table_name => { id: taxon.id })
-    #
-    # If you're using count on the result of this scope, you must use the
-    # `:distinct` option as well:
-    #
-    #   Spree::Product.in_taxon(taxon).count(distinct: true)
-    #
-    # This is so that the count query is distinct'd:
-    #
-    #   SELECT COUNT(DISTINCT "spree_products"."id") ...
-    #
-    #   vs.
-    #
-    #   SELECT COUNT(*) ...
-    add_search_scope :in_taxon do |taxon|
-      includes(:classifications)
-        .where('spree_products_taxons.taxon_id' => taxon.self_and_descendants.pluck(:id))
-        .select(Spree::Classification.arel_table[:position])
-        .order(Spree::Classification.arel_table[:position].asc)
-    end
+          # This scope selects products in taxon AND all its descendants
+          # If you need products only within one taxon use
+          #
+          #   Spree::Product.joins(:taxons).where(Taxon.table_name => { id: taxon.id })
+          #
+          # If you're using count on the result of this scope, you must use the
+          # `:distinct` option as well:
+          #
+          #   Spree::Product.in_taxon(taxon).count(distinct: true)
+          #
+          # This is so that the count query is distinct'd:
+          #
+          #   SELECT COUNT(DISTINCT "spree_products"."id") ...
+          #
+          #   vs.
+          #
+          #   SELECT COUNT(*) ...
+          add_search_scope :in_taxon do |taxon|
+            includes(:classifications)
+              .where('spree_products_taxons.taxon_id' => taxon.self_and_descendants.pluck(:id))
+              .select(Spree::Classification.arel_table[:position])
+              .order(Spree::Classification.arel_table[:position].asc)
+          end
 
-    # This scope selects products in all taxons AND all its descendants
-    # If you need products only within one taxon use
-    #
-    #   Spree::Product.taxons_id_eq([x,y])
-    add_search_scope :in_taxons do |*taxons|
-      taxons = get_taxons(taxons)
-      taxons.first ? prepare_taxon_conditions(taxons) : where(nil)
-    end
+          # This scope selects products in all taxons AND all its descendants
+          # If you need products only within one taxon use
+          #
+          #   Spree::Product.taxons_id_eq([x,y])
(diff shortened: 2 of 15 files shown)


aldesantis:
On @spaghetticode's advice, I have removed the commit that converted `update_attributes` to `update`, so that the PR is less likely to cause conflicts as new work is merged into `master`. Also took the chance to rebase against the latest `master`.

kennyadsl:
@aldesantis makes sense, are you going to open a new PR with the `update_attribute` change?

aldesantis:
@kennyadsl yes, probably after we merge this one.

kennyadsl:
@aldesantis isn't better to start using `update` before merging the Rails 6 support instead? This way we won't have the deprecation warning at all, right?

aldesantis:
@kennyadsl definitely, what I meant was that instead of opening a PR now and then having to constantly keep it up to date with `master`, I will just wait for this PR to be merged, pull the latest `master` and then make the change.

skukx:
@aldesantis Can you expand on rspec 3.8 and rails 6 support? I've not seen any issues on any of my rails 6 projects. I'm not saying there aren't issues, but I am curious what those could be.

aldesantis:
Hi @skukx! 👋 The issue I have encountered specifically is this: https://github.com/rspec/rspec-rails/issues/2086, which was resolved in `4-0-dev`. Let me know if I missed something though!

aldesantis:
We're almost there guys, we're just waiting for rspec-rails 4 at this point!

tvdeyen:
> We're almost there guys, we're just waiting for RSpec 4 at this point!

@aldesantis you can use `gem 'rspec-rails', '>= 4.0.0.beta2'`

Gemfile
```diff
@@ -28,7 +28,11 @@ group :backend, :frontend, :core, :api do
   gem 'database_cleaner', '~> 1.3', require: false
   gem 'factory_bot_rails', '~> 4.8', require: false
   gem 'rspec-activemodel-mocks', '~> 1.1', require: false
-  gem 'rspec-rails', '~> 3.7', require: false
+  gem 'rspec-rails', github: 'rspec/rspec-rails', branch: '4-0-dev', require: false
+  gem 'rspec-core', github: 'rspec/rspec-core', require: false
+  gem 'rspec-mocks', github: 'rspec/rspec-mocks', require: false
+  gem 'rspec-support', github: 'rspec/rspec-support', require: false
+  gem 'rspec-expectations', github: 'rspec/rspec-expectations', require: false
```
tvdeyen: You can just use `gem 'rspec-rails', '>= 4.0.0.beta2'` instead.

aldesantis: @tvdeyen the betas are not good unfortunately, they don't provide full support for Rails 6. We need to wait for 4.0.0 stable. 😞

core/solidus_core.gemspec
```diff
@@ -29,8 +29,8 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'activemerchant', '~> 1.66'
   s.add_dependency 'acts_as_list', '~> 0.3'
-  s.add_dependency 'awesome_nested_set', '~> 3.0', '>= 3.0.1'
-  s.add_dependency 'cancancan', '~> 2.2'
+  s.add_dependency 'awesome_nested_set', '~> 3.2', '>= 3.0.1'
+  s.add_dependency 'cancancan', '~> 3.0'
```
tvdeyen: There is no need to use 3 and above. It also works with `['>= 2.1', '< 4.0']`
We want to keep our dependencies always as alow as possible to avoid conflicts with other gems.

Gemfile
tvdeyen: What exactly is the issue? I have good results with the betas in Alchemy.

core/solidus_core.gemspec
aldesantis: @tvdeyen I tried with >= 2.1, < 4.0, but I get tons of these:
```
 111) Taxonomies show should display existing taxonomies
       Failure/Error: @search = ~Spree~::~Order~.accessible_by(current_ability, :index).ransack(params[:q])
       
       CanCan::Error:
         The accessible_by call cannot be used with a block ‘can’ definition.The SQL cannot be determined for :index Spree::Order
```

Gemfile
aldesantis: @tvdeyen I think you're right, not sure what kind of problem I experienced initially but it doesn't seem to happen anymore. It was probably not related to RSpec.

core/solidus_core.gemspec
tvdeyen: Interesting. This does not seem to be Rails 6 related to me. :thinking:

kennyadsl: I think we need https://github.com/solidusio/solidus/pull/3174 as well? Maybe loading that code conditionally if Rails >= 6

aldesantis: With CanCanCan 3, everything seems to work without #3174. Not sure if #3174 would allow us to continue using CanCanCan 2?

tvdeyen: I commented on #3174 that we should try to allow 2 as well. It should work as it only removes lots of quirks.

aldesantis: Even with the patch from #3174, I keep seeing lots of test failures when using CanCanCan 2. I think an upgrade to 3 is required, but again, let me know if you find a way to make the suite pass.

core/spec/lib/search/base_spec.rb
```diff
@@ -60,7 +60,11 @@
                search: { "price_range_any" => ["Under $10.00", "$10.00 - $15.00"] } }
     searcher = Spree::Core::Search::Base.new(params)
     expect(searcher.send(:get_base_scope).to_sql).to match /<= 10/
-    expect(searcher.send(:get_base_scope).to_sql).to match /between 10 and 15/i
+    if Rails.gem_version.segments.first == 6
+      expect(searcher.send(:get_base_scope).to_sql).to match /between 10\.0 and 15\.0/i
+    else
+      expect(searcher.send(:get_base_scope).to_sql).to match /between 10 and 15/i
+    end
```
kennyadsl: can you please explain why this is needed in the commit message?

core/spec/lib/spree/event_spec.rb
```diff
@@ -18,14 +18,26 @@
...
     end
 
     after do
-      notifier.instance_variable_set '@subscribers', @old_subscribers
+      if Rails.gem_version.segments.first == 6
+        notifier.instance_variable_set '@string_subscribers', @old_string_subscribers
+        notifier.instance_variable_set '@other_subscribers', @old_other_subscribers
+      else
+        notifier.instance_variable_set '@subscribers', @old_subscribers
+      end
```
kennyadsl: here as well, I'm not getting why this is needed. A couple of line in the commit message would be helpful

core/app/models/spree/product/scopes.rb
```diff
@@ -168,93 +171,97 @@ def self.property_conditions(property)
...
+                    Spree::Taxon.where("#{taxons}.permalink LIKE ? OR #{taxons}.permalink = ?", "%/#{t}/", "#{t}/").first
+                end
+              }.compact.flatten.uniq
+            end
           end
-        }.compact.flatten.uniq
+        end
       end
+
+      ::Spree::Product.prepend self
```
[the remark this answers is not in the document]
aldesantis: @kennyadsl that's correct!

aldesantis:
@kennyadsl I have rebased with more granular and detailed commits. Let me know what you think!

Gemfile
```diff
@@ -28,7 +28,7 @@ group :backend, :frontend, :core, :api do
   gem 'database_cleaner', '~> 1.3', require: false
   gem 'factory_bot_rails', '~> 4.8', require: false
   gem 'rspec-activemodel-mocks', '~> 1.1', require: false
-  gem 'rspec-rails', '~> 3.7', require: false
+  gem 'rspec-rails', '~> 4.0.0.beta2', require: false
```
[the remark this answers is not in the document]
aldesantis: @ericsaupe yes it is, see https://github.com/solidusio/solidus/pull/3236#issuecomment-520341615.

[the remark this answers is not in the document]
aldesantis: @ericsaupe yeah, I think so. Since this will be released with 2.10 (I guess) there's no rush anyway, I imagine a stable release is behind the corner for RSpec.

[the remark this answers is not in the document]
tvdeyen: I don’t see why we need to wait until a stable rspec 4 has been released. We don’t ship this to any of our users but solely use this Gemfile for testing. So, IMO its totally fine.

core/solidus_core.gemspec
```diff
@@ -24,13 +24,13 @@ Gem::Specification.new do |s|
     actionmailer actionpack actionview activejob activemodel activerecord
     activesupport railties
   ].each do |rails_dep|
-    s.add_dependency rails_dep, ['>= 5.1', '< 5.3.x']
+    s.add_dependency rails_dep, ['>= 5.1', '< 7.0.x']
```
tvdeyen: Knowing from history of a lots of Rails Upgrades we should lower this to 6.1 ;)

core/spec/lib/spree/event_spec.rb
```diff
@@ -18,14 +18,26 @@
     before do
       # ActiveSupport::Notifications does not provide an interface to clean all
       # subscribers at once, so some low level brittle code is required
-      @old_subscribers = notifier.instance_variable_get('@subscribers').dup
+      if Rails.gem_version.segments.first == 6
```
tvdeyen: Sometimes you use `Rails.gem_version >= Gem::Version.new('6.0.0')` sometimes this, I would prefer to use either the one or the other but not both.

core/solidus_core.gemspec
tvdeyen: I would like to see some of these errors. I ask myself what a Rails update should change in terms of these errors from cancancan. The issue is that we try to use the same ability rule for scopes as well for single records. Then cancancan is not able to generate the correct sql for the scope. Very interesting. 

I really would like to not raise a commonly used dependency like cancancan if we find a way to not need it. I know this might be disappointing and we all want to land Rails 6 support, but we also have responsibility for not breaking extensions and stores that might use dependencies that are not supporting cancancan 3 yet.

core/lib/spree/permission_sets/default_customer.rb
```diff
@@ -8,7 +8,7 @@ def activate!
         can :display, OptionType
         can :display, OptionValue
         can :create, Order
-        can [:read, :update], Order do |order, token|
+        can [:read, :update], Order, Order.where(user: user) do |order, token|
```
tvdeyen: I think we need more of these. The third parameter is for scopes, the block for single records. Then we should be able to still support cancancan 2

Gemfile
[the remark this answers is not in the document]
aldesantis: @tvdeyen I don't think there's a specific reason, it's just that there's no rush to merge this IMO and using the stable now will ensure we don't run into any unexpected issues down the road when the stable is released. I'm fine with both courses of action though, we can either wait for stable or merge with the beta.

core/spec/lib/spree/event_spec.rb
aldesantis: @tvdeyen you're absolutely right, I fixed these to always use `Gem::Version.new('6.0.0')` which as far as I can see is the standard way of doing this stuff.

core/solidus_core.gemspec
aldesantis: @tvdeyen you mean because 6.1 might introduce breaking changes? If that's the case, I see your point but I think it would be painful for users who want to upgrade to 6.1 to have to wait for a new Solidus release. On the other hand, locking to 6.0 in case a breaking change is introduced is quite easy.

I don't have a very strong opinion on this though, so I'm happy to lock to < 6.1 if you think a higher upper limit is likely to cause problems.

tvdeyen: I my experience Rails x.0 upgrades are always the easiest, x.1 updates not necessarily. But we can leave it as is and adjust if necessary.

tvdeyen: You must not forget that Rails does not follow semver.

aldesantis:
@tvdeyen you were right on CanCanCan, I went with `['>= 2.1', '< 4.0']` and it all seems to work. Sorry about the long discussion there, I incorrectly assumed you wanted to stay on `~> 2.2` which is I didn't see it working. Thanks!

core/lib/spree/permission_sets/default_customer.rb
aldesantis: @tvdeyen I'm actually surprised it worked before, since we didn't have a scope. Anyway, this seems to be enough to make it work.

tvdeyen:
> you were right on CanCanCan, I went with ['>= 2.1', '< 4.0'] and it all seems to work. Sorry about the long discussion there, I incorrectly assumed you wanted to stay on ~> 2.2 which is why I didn't see it working. Thanks!

@aldesantis No worries. You did an amazing job!

> By the way, should we use >= 2.2 instead of >= 2.1, since the previous constraint was ~> 2.2?

Yes, sorry, I just copyied it over from another project ;)

Gemfile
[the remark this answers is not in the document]
tvdeyen: I see lets leave it as is and merge this PR so people can start testing master with Rails 6

core/lib/spree/permission_sets/default_customer.rb
tvdeyen: :man_shrugging:

aldesantis:
Bumped CanCanCan's lower bound to 2.2!

core/lib/spree/permission_sets/default_customer.rb
kennyadsl: @coorasse if you have some time, can you take a quick look here, please? Just to be sure we are not missing something important. Thanks!

aldesantis:
I have also made Rails 6 the default in CircleCI and in the Gemfile, as per @kennyadsl's request.

ericsaupe:
Are we waiting for a full release of `rspec-rails`?

core/solidus_core.gemspec
```diff
@@ -29,8 +29,8 @@ Gem::Specification.new do |s|
 
   s.add_dependency 'activemerchant', '~> 1.66'
   s.add_dependency 'acts_as_list', '~> 0.3'
-  s.add_dependency 'cancancan', '~> 2.2'
   s.add_dependency 'awesome_nested_set', '~> 3.2'
+  s.add_dependency 'cancancan', ['>= 2.2', '< 4.0']
```
tvdeyen: Just saw that we should update the commit message

aldesantis: @tvdeyen you're right, it wasn't very clear. What about now?

tvdeyen: :+1: thanks

tvdeyen:
:tada: :champagne: :100:

PR #2974, opened 2018-11-23
ActiveStorage support

elia:
📝 _This is an attempt to resurrect the great work started by @swcraig ([here](https://github.com/solidusio/solidus/pull/2328) and [here](https://github.com/swcraig/solidus_paperclip)) and building on the gallery API introduced in #2337_ 

### The old approach

After attempting to stay within the bounds of the old attempt I realized there was a problem with the fact that the gallery implementation wasn't really completed (several places still access images directly instead of going through the new API). 

### The new approach

The new approach is not to create an extension but rather extract all the Paperclip stuff to a concern and provide an alternative implementation based on ActiveStorage. That turned out to be way more clean and required way less code.

### Migrating the assets and Paperclip/AS bridge

The nice thing is that it's also possible to offer a better migration path thanks to ActiveStorage mirroring. The idea is to create an ActiveStorage service based on the existing Paperclip storage that will serve the images from their current locations (e.g. S3 with the Paperclip defined folder structure) and have at the same time an AS mirror that points to the AS-S3 service. At this point new uploads would go to both service and there's plenty of time for a progressive migration of the assets base until the old service can be dismissed.

_This will be part of a different PR, so this one won't be blocked and new installation can start with AS as soon as possible_

## 🛣 Roadmap

- [x] move image/paperclip stuff to a concern
- [x] implement an alternative image/activestorage concern
- [x] allow to switch image concern via an app-configuration
- [x] move taxon/paperclip stuff to a concern
- [x] implement an alternative taxon/activestorage concern
- [x] allow to switch taxon concern via an app-configuration
- [x] evaluate extracting common code between the concerns / refactoring

## Old roadmap (discarded)
- ~setup the extension inside the repo making it a dependency of the main gem~
- ~move paperclip-related code piece by piece to the extension~
- ~add an alternative ActiveStorage-based gallery implementation~
- ~(maybe) add a specialized ActiveStorage service that can work with paperclip assets, to be as [mirror service](https://edgeguides.rubyonrails.org/active_storage_overview.html#mirror-service)~
- ~extract the extension to a different repo~

## Reference links
- https://edgeguides.rubyonrails.org/active_storage_overview.html
- https://github.com/solidusio/solidus/pull/2337
- https://github.com/spree/spree/pull/8709

PR #3237, opened 2019-06-21
Attachment adapters

elia:
**Description**

_This PR has been extracted from https://github.com/solidusio/solidus/pull/2974 as [suggested](https://github.com/solidusio/solidus/pull/2974#issuecomment-503125134) by @kennyadsl_

Paperclip has been officially deprecated and is no more maintained, but the supposed replacement with ActiveStorage is not possible to due to lack of proper support for public long-term urls. 

This PR allows Image and Taxon to switch their attachment implementation, extracting Paperclip to a concern, and opening the door for alternative implementations (e.g. Carrierwave, Shrine, AttachmentFu, [et cetera](https://www.ruby-toolbox.com/categories/rails_file_uploads)).

 
**Checklist:**

### DIFF
--- a/core/app/models/spree/image.rb
+++ b/core/app/models/spree/image.rb
@@ -2,55 +2,13 @@
 
 module Spree
   class Image < Asset
-    validate :no_attachment_errors
-
-    has_attached_file :attachment,
-                      styles: { mini: '48x48>', small: '100x100>', product: '240x240>', large: '600x600>' },
-                      default_style: :product,
-                      default_url: 'noimage/:style.png',
-                      url: '/spree/products/:id/:style/:basename.:extension',
-                      path: ':rails_root/public/spree/products/:id/:style/:basename.:extension',
-                      convert_options: { all: '-strip -auto-orient -colorspace sRGB' }
-    validates_attachment :attachment,
-      presence: true,
-      content_type: { content_type: %w(image/jpeg image/jpg image/png image/gif) }
-
-    # save the w,h of the original image (from which others can be calculated)
-    # we need to look at the write-queue for images which have not been saved yet
-    after_post_process :find_dimensions, if: :valid?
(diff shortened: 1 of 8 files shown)


core/app/models/spree/image.rb
```diff
@@ -2,55 +2,13 @@
...
-
-    # if there are errors from the plugin, then add a more meaningful message
-    def no_attachment_errors
-      unless attachment.errors.empty?
-        # uncomment this to get rid of the less-than-useful interim messages
-        # errors.clear
-        errors.add :attachment, "Paperclip returned errors for file '#{attachment_file_name}' - check ImageMagick installation or image source file."
-        false
-      end
+      url(:mini, false)
```
spaghetticode: although tests are green, I think there's an issue here. `Spree::Image#url` only takes one argument: 

```ruby
i = Spree::Image.new

show-method i.url
def url(size)
  attachment.url(size)
end

show-method mini_url
  Spree::Deprecation.warn(
    'Spree::Image#mini_url is DEPRECATED. Use Spree::Image#url(:mini) instead.'
  )
  url(:mini, false)

i.mini_url
DEPRECATION WARNING: Spree::Image#mini_url is DEPRECATED. Use Spree::Image#url(:mini) instead. (called from irb_binding at (pry):1)
ArgumentError: wrong number of arguments (given 2, expected 1)

**Comment by kennyadsl** [comment]
I've tested this PR in a clean application that also has a custom definition of `Spree::Image` style:

```ruby

Spree::Image.attachment_definitions[:attachment][:styles] = {
  mini: '48x48>',
  small: '100x100>',
  product: '240x240>',
  large: '600x600>',
  mobile: '400x500>',
}
```

Everything works as expected but we need to fix what @spaghetticode spotted, thanks @elia !

PR #3239, opened 2019-06-26
Refactor Promotions Environment Configuration

kennyadsl:
**Description**

~🚨This PR lives on top of #3238 and should be merged after that one. 🚨~ -> Merged and rebased.

At the moment, there are some small issues with how `spree.config.promotions` is defined. 

First of all, it's an "`Environment`" object, like `Spree::Core::Environment::Calculators` (both include the `EnvironmentExtension` concern), but they are in two different places on the filesystem, without any valid reason (only looking at git history).  

Another issue is that [we are currently redefining it every time we create the main Environment object](https://github.com/solidusio/solidus/blob/b8dbd1fd7ea216d0a4cf197ad7d8d15712a926dd/core/lib/spree/app_configuration.rb#L482), but that assignment is useless since [it's already initialized with the same thing](https://github.com/solidusio/solidus/blob/b8dbd1fd7ea216d0a4cf197ad7d8d15712a926dd/core/lib/spree/core/environment.rb#L18).

This PR fixes these issues and deprecates the old class used to represent `env.promotions` (`Spree::Promo::Environment`).

**Checklist:**


kennyadsl:
@eric1234 sure, I'll add this to the CHANGELOG but please keep in mind that this should be transparent for users and if they customized the deprecated class, they will receive a deprecation warning with customizations preserved. I've just added some specs around this here to be sure it works as expected: https://github.com/solidusio/solidus/pull/3239/files#diff-31874dd9132638da9df25acb9f1e3d1f

PR #3247, opened 2019-07-03
Add zone link

jacobherrington:
I found this comment that was showing up, so I went ahead and added the missing link
image

### DIFF
--- a/guides/source/developers/taxation/overview.html.md
+++ b/guides/source/developers/taxation/overview.html.md
@@ -49,13 +49,15 @@ In Solidus, a tax rate consists of at least four values:
 
 - The descriptive name for the tax rate. For example, "Minnesota Sales Tax" for
   a Minnesota state tax rate.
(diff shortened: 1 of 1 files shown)


guides/source/developers/taxation/overview.html.md
```diff
@@ -49,7 +49,7 @@ In Solidus, a tax rate consists of at least four values:
 
 - The descriptive name for the tax rate. For example, "Minnesota Sales Tax" for
   a Minnesota state tax rate.
-- The zone that the tax rate should apply to. <!-- TODO: Link to zones article.-->
+- The [zone](../locations/zones.html) that the tax rate should apply to.
```
kennyadsl: Can we use the link format at the end of the paragraph as the rest of guides pages (I think). Should be something like:

```md
...
The [zone][zone] that the tax rate should apply to.
...

[zone]: ../locations/zones.html
```

jacobherrington: Sure. 👍

PR #3248, opened 2019-07-03
Add missing CSS classes

jacobherrington:
This is a small UI mistake. We are missing a CSS class here which makes this button look different than the other Product buttons.

Before:
image

After:
image

Example of another Product button:
image


jacobherrington:
I found these CSS classes were missing from the "New Variant" button too.
After:
image

PR #3253, opened 2019-07-04
Upgrade to CircleCI 2.1

aldesantis:
**Description**

Upgrades to CircleCI 2.1, which allows us to use Orbs. This is especially important for https://github.com/solidusio/solidus/pull/3252, since we want to use [this Orb](https://github.com/nebulab/circleci-orbs-stoplight) for keeping the OAS documentation in sync with Stoplight.

**Checklist:**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   postgres: &postgres
     working_directory: &workdir ~/solidus
@@ -72,7 +72,6 @@ jobs:
       RAILS_VERSION: '~> 5.1.0'
 
 workflows:
-  version: 2
   build:
     jobs:
       - postgres
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
       - run: |
           bundle install --path=vendor/bundle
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - run: |
           bundle install --path=vendor/bundle
 
-      - cache-save:
+      - save_cache:
```
kennyadsl: I'm sure there's a valid reason but can you explain this change better in the commit message, please?

aldesantis: @kennyadsl `cache-save` is not a valid step name in CircleCI, I'm actually not sure how this has worked so far. The right name is `save_cache`.

kennyadsl: 😱

I'm also noticing that we are trying to restore the key `solidus-gems-v1`, which is never saved...

aldesantis: @kennyadsl that's actually fine, because it will look for any keys that start with `solidus-gems-v1`.

kennyadsl: Found some reference about the [`cache-save`](https://thoughtbot.com/blog/circleci-2-rails) step, maybe it's the CircleCI legacy way of calling that type of jobs.

Issue #3233, opened 2019-06-19
RFC: View hooks beyond Deface


tvdeyen:
## The problem

Extensions and stores want to make additions (or even changes) to the admin views. As of yet, the defacto standard for this is [Deface](https://github.com/spree/deface). Deface brings a lot of other problems - first and foremost it hides view changes into places where they are hard to track down or introduces "where does this checkbox coming from?" problems. Also it is barely known outside the Spree/Solidus ecosystem. Additionally we need to maintain HTML hooks (`data-hook`) that exist solely for Deface to work properly. That makes admin views refactoring very hard.

## Proposal

I am proposing an idea that originally came up from @mamhoff while working with him on a client store that makes lots of changes to the Solidus admin. We removed Deface from that store a while ago and introduced another way to make admin view changes maintainable:

**Rails partials**

By splitting up the admin views into smaller chunks we make it easier for us to keep our admin maintainable during Solidus updates.

### This is how it works

The Admin Orders table for example.

We split the `admin/orders/index.html.erb` template into several partials:

#### 1. The filter fields

```erb
...
<% content_for :table_filter do %>
  <div data-hook="admin_orders_index_search">
    <%= search_form_for [:admin, @search] do |f| %>
      <%= render 'filter_fields', f: f %>
      ...
    <% end %>
  </div>
<% end %>
...
```

#### 2. The table

```erb
...
<% if @orders.any? %>
  <%= render 'table', orders: @orders, ransack_search: @search, show_only_completed: @show_only_completed %>
<% else %>
  ...
<% end %>
...
```

#### 3. The table row

```erb
<table class="index" id="listing_orders" data-hook>
  <colgroup>
    ...
  </colgroup>
  <thead>
    ...
  </thead>
  <tbody>
    <%= render collection: orders, partial: 'spree/admin/orders/order' %>
  </tbody>
</table>
```

## Summary

For index views I propose implementing a pattern like:

```
admin/resource/index.html.erb
  - admin/resource/_filter_fields.html.erb
  - admin/resource/_table.html.erb
    - admin/resource/_resource.html.erb
```

For edit views I propose implementing a pattern like:

```
admin/resource/edit.html.erb
  - admin/resource/_form.html.erb
  - admin/resource/_form_additions.html.erb
  - admin/resource/_side_bar_additions.html.erb
```

[...]

PR #3257, opened 2019-07-09
RFC: Admin configurable tabs

kennyadsl:
**Description**

Following the discussion in #3233, I'm working on this spike to allow easy customization of admin tabs for stores and extensions, without using Deface.

This proposed structure is just a working attempt to programmatically control the tabs in the product page. If it works for everyone, we can extend it to all the other pages that use tabs.

Suppose you want to add a new tab for a product reviews section, that you
implemented custom in your store.

You can create a class with your logic, similar to this one:

```ruby
# lib/spree/backend/tabs/product/reviews.rb

require 'spree/backend/tabs/product/base'

module Spree
  module Backend
    module Tabs
      class Product
        class Reviews < Base
          def name
            'Reviews'
          end

          def presentation
            view_context.t('your_app.reviews')
          end

          def url
            view_context.spree.admin_product_reviews_url(product)
          end

          def visible?
            view_context.can?(:admin, Spree::Review)
          end
        end
      end
    end
  end
end
```

Next step is adding this class to the list of tabs to load:

```ruby
# config/initializer/spree.rb

  config.product_tabs.items << 'Spree::Backend::Tabs::Product::Reviews'
end
```

🗣Any feedback in this early stage is highly appreciated.

**TODO**

- [ ] Wait for feedback
- [ ] Write tests
- [ ] Add remaining tabs
- [ ] Write documentation for other tabs or write a generic documentation for tabs
- [ ] Understand if we can deprecate the usage of `data-hook="admin_product_tabs"` via Deface
- [ ] Consider backporting to supported versions, so we can remove deface overrides in extensions for tabs

**Checklist:**


backend/app/models/spree/backend_configuration.rb
```diff
@@ -1,9 +1,24 @@
 # frozen_string_literal: true
 
+require 'spree/backend/environment/product_tabs'
+
 module Spree
   class BackendConfiguration < Preferences::Configuration
     preference :locale, :string, default: Rails.application.config.i18n.default_locale
 
+    def product_tabs
+      @product_tabs ||= Backend::Environment::ProductTabs.new.tap do |product_tabs|
```
kennyadsl: I can probably do something like this instead: https://github.com/solidusio/solidus/pull/3232/files#diff-c379205c70b6f279fcb4141310bbef4dR9

cc @elia

backend/app/views/spree/admin/shared/_product_tabs.html.erb
```diff
@@ -4,24 +4,11 @@
...
-      <%= content_tag :li, class: ('active' if current == 'Prices') do %>
-        <%= link_to plural_resource_name(Spree::Price), spree.admin_product_prices_url(@product) %>
-      <% end if can?(:admin, Spree::Price) && !@product.deleted? %>
-      <%= content_tag :li, class: ('active' if current == 'Product Properties') do %>
-        <%= link_to plural_resource_name(Spree::ProductProperty), spree.admin_product_product_properties_url(@product) %>
-      <% end if can?(:admin, Spree::ProductProperty) && !@product.deleted? %>
-      <%= content_tag :li, class: ('active' if current == 'Stock Management') do %>
-        <%= link_to t('spree.stock_management'), spree.admin_product_stock_url(@product) %>
-      <% end if can?(:admin, Spree::StockItem) && !@product.deleted? %>
+      <% product_tabs(current: current).each do |tab| %>
```
tvdeyen: Could we pass the `product: @product` in here so we do not need to use and assign instance variables on the view context?

backend/lib/spree/backend/tabs/product/base.rb
```diff
@@ -0,0 +1,50 @@
...
+          def css_classes
+            return 'active' if active?
+
+            ''
+          end
+
+          private
+
+          def product
+            view_context.assigns["product"]
```
tvdeyen: I would like to not do this. We should pass the product into here instead.

backend/lib/spree/backend/tabs/product/images.rb
```diff
@@ -0,0 +1,29 @@
...
+  module Backend
+    module Tabs
+      class Product
+        class Images < Base
+          def name
+            'Images'
+          end
+
+          def presentation
+            view_context.t('spree.images')
```
tvdeyen: There is no need to use the `view_context`. `I18n.t('spree.images')` should work as well.

backend/lib/spree/backend/tabs/product/images.rb
```diff
@@ -0,0 +1,29 @@
...
+          end
+
+          def url
+            view_context.spree.admin_product_images_url(product)
```
tvdeyen: Could we pass the `spree` url proxy into the base class instead of using the `view_context`? Or try to make use of `Spree::Core::Engine.routes.url_helpers`

backend/lib/spree/backend/tabs/product/images.rb
```diff
@@ -0,0 +1,29 @@
...
+          end
+
+          def visible?
+            view_context.can?(:admin, Spree::Image) && !product.deleted?
```
tvdeyen: We should pass the `current_ability` into the base class instead of using `view_context`

backend/lib/spree/backend/tabs/product/prices.rb
```diff
@@ -0,0 +1,29 @@
...
+          def name
+            'Prices'
+          end
+
+          def presentation
+            view_context.plural_resource_name(Spree::Price)
+          end
+
+          def url
+            view_context.spree.admin_product_prices_url(product)
```
tvdeyen: Ditto we should use `Spree::Core::Engine.routes.url_helpers` instead.

backend/lib/spree/backend/tabs/product/prices.rb
```diff
@@ -0,0 +1,29 @@
...
+          end
+
+          def visible?
+            view_context.can?(:admin, Spree::Price) && !product.deleted?
```
tvdeyen: Ditto we should pass the `current_ability` instead.

backend/lib/spree/backend/tabs/product/base.rb
```diff
@@ -0,0 +1,50 @@
...
+          attr_reader :view_context, :current
+
+          def initialize(view_context:, current: nil)
+            @view_context = view_context
+            @current = current
+          end
+
+          def name
+            raise NotImplementedError
+          end
```
tvdeyen: We do not use the name anywhere. Do we need it then?

kennyadsl: We use it into the `active?` method actually.

kennyadsl:
@tvdeyen thanks, all your feedback makes sense. I'll try to rewrite this following your suggestion and other internal notes that I received.

backend/lib/spree/backend/tabs/product/prices.rb
```diff
@@ -0,0 +1,29 @@
...
+  module Backend
+    module Tabs
+      class Product
+        class Prices < Base
+          def name
+            'Prices'
+          end
+
+          def presentation
+            view_context.plural_resource_name(Spree::Price)
```
kennyadsl: @tvdeyen unfortunately, we don't have this `plural_resource_name` method into `I18n`, it's something we've built as a helper [here](https://github.com/solidusio/solidus/blob/1ec3fac0b65dbbf97be3dbf119b0f0133017484a/core/app/helpers/spree/base_helper.rb#L150) and I can't think of a clean way to use it here without using the `view_context`, do you have any suggestion?

tvdeyen: Yes, you should be able to do

```rb
Spree::Price.model_name.human(count: Spree::I18N_GENERIC_PLURAL)
```

no?

kennyadsl: Sure, but there's this not ideal duplication of logic which I'd like to avoid.

tvdeyen: Not all duplication is bad. I consider this a good thing actually. We do not want to rely on that the `view_context` has magically access to the `plural_resource_name` helper. I kind of embrace explicitness over Rails magic lately ;)

backend/lib/spree/backend/tabs/product/base.rb
tvdeyen: Oh, my bad :)

backend/lib/spree/backend/tabs/product/images.rb
tvdeyen: Maybe add a `url_helpers` method on the base class

```rb
def url_helpers
  Spree::Core::Engine.routes.url_helpers
end
```

and make use of it here?

```rb
def url
  url_helpers.admin_product_images_path(product)
end
```

kennyadsl:
I've made a couple of other attempts:

[**v2**](https://github.com/solidusio/solidus/commit/6f84ef2de47fbd883af510cb873e6bd58ecc417c)

This follows the @tvdeyen suggestions of not using a view_context. I've created another object called `TabsContext` that stores all the information shared between tabs in the same tabs set. I think that the advantage of having a specific object to store context data is that we don't have to explicitly pass all data to all tabs every time. 

I still think this has a big issue, the main object of the tabs context, in this case the `Spree::Product` instance, could be more than one object for other tabs sets, for example in the configuration pages that have tabs but are not related to .a specific object.

[**v3**](https://github.com/solidusio/solidus/commit/6bf4f689c27f229c15d337ebbdff68398f70fc48)

This is similar to the approach used for `MenuItems` and it's probably the approach that I like the most at this point. We inject instances of `Spree::Backend::Tabs` into the configuration array and use them in the view. I think this work, and it's more flexible but I'm not sure about the extensive usage of `instance_eval` in the views. 

Any thoughts?

Thought

tvdeyen:
@kennyadsl I also prefer the v3. Much easier and `instance_exec` is a good trade-off here. I left some comments on the commit itself. Somehow GitHub is not displaying them here. Hope they reached you?

waiting-for-dev:
Hey, @kennyadsl, it'd be nice to have this one. Pinging you in case you find the time to complete it 🙂

chrean:
Closing because we are in the process of completely redoing the admin, starting from a UX design.

PR #3260, opened 2019-07-10
Make all belongs_to associations optional

tvdeyen:
Since Rails 5.1 `belongs_to` associations are required by default. We used to disable this option globally with a ActiveRecord class setting. With this we make the Rails 5.1 default setting obsolote and turn this off app wide.

```rb
Rails.application.config.active_record.belongs_to_required_by_default # => true
Spree::Base.belongs_to_required_by_default # => false
ActiveRecord::Base.belongs_to_required_by_default # => false
```

The fun fact is, the tests all pass with this removed. But this has the potential to break stores if their data is not consistent.

We have several options to solve this

1. We tell our users to clean up the database before running the next Solidus update where this included.
2. We include this in a mayor version update
3. We explicitely make all our `belongs_to` associations `optional`

I am in favor of the latter, so I made all `belongs_to` associations optional for now.

**Description**

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/user_address_book.rb
+++ b/core/app/models/concerns/spree/user_address_book.rb
@@ -25,7 +25,7 @@ module Spree
       has_many :addresses, through: :user_addresses
 
       # bill_address is only minimally used now, but we can't get rid of it without a major version release
-      belongs_to :bill_address, class_name: 'Spree::Address'
+      belongs_to :bill_address, class_name: 'Spree::Address', optional: true
 
       has_one :default_user_address, ->{ default }, class_name: 'Spree::UserAddress', foreign_key: 'user_id'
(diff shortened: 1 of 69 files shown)


kennyadsl:
Thanks @tvdeyen, I agree with this approach and I think this way it can be added in a minor version update with a good changelog entry but, should we also update all the extensions?

tvdeyen:
Ideally we also would introduce foreign keys for all our associations as John has done in #2036

Also something we would do in a mayor version, though.

tvdeyen:
@kennyadsl do you think we can merge this?

tvdeyen:
CircleCI wont catch up my updates. Do not know why. Will re-open this PR

tvdeyen:
Created #3309

PR #3269, opened 2019-07-12
Add a color to menu selected items

mfrecchiami:
There's an issue related to my previous PR (the update of colors palette): when there is a first level menu item, the user can't understand in which section he actually is, if he look only at the sidebar navigation (e.g.: **Orders** or **Users**) because any color or background, changes on the first level menu items (see picture below):

issue

This is a mistake done thought about the future possibility of a navigation without popover but with an active dropdown (like most of modern navigation sidebar).
So, I add a background color on the first level menu item that has the `.selected` class.
This is the  result:

selected_sample_02

selected_sample_01


mfrecchiami:
@kennyadsl thanks for your reporting!

kennyadsl:
@mfrecchiami can you please rebase against the updated upstream master to fix the specs?

PR #3270, opened 2019-07-12
Upload new sample images

mfrecchiami:
The sample images of the online store demo has been updated with new items, branded by Solidus and by a redesigned version of the Ruby logo.

Also image sizes has been changed, to support the retina screens (but maybe an helper that load ` @2x` pictures on retina screens only, could be useful).
The smaller size called `mini` is still the same: the issue is that on the product page (and somewhere in the backend, also) there is any _width_ or _max-width_ defined, for the thumbnail list item.

new_products_home

I noticed that where are more than two colors variant, there is no way to show all thumbnails.
Actually the thumbs are (in order):

- the home page product image

- the thumbs related to the variant selected by the default _Variants_ input radio

3_colors_2_thumbs

IMO on the product page, it could be better show first the item with the same variation, e.g. color, that is shown on the home page.
For now it is possible only to put same color variation as the first option in _Variations_ list, but then all the thumbs shows the same color, so it seems like duplicated items (until the user select another option).


sample/db/samples/product_properties.rb
```diff
@@ -2,61 +2,122 @@
...
+    # {
+    #   "Manufacturer" => "Jerseys",
+    #   "Brand" => "Conditioned",
+    #   "Model" => "WM6001",
+    #   "Shirt Type" => "Skinny",
+    #   "Sleeve Type" => "Short",
+    #   "Made from" => "90% Cotton, 10% Nylon",
+    #   "Fit" => "Slim",
+    #   "Gender" => "Women's"
+    # }
```
tvdeyen: This can probably be removed?

mfrecchiami: @tvdeyen thanks! I just removed it 😉

mfrecchiami:
@kennyadsl thanks 👍 

The goal, is to have a product page like this: 
[Considering a product with multiple variation colors and which Black is the color showed on home page]

sample

kennyadsl:
We'll need an extensive CHANGELOG entry on this one since users could be relying on those images size and updating Solidus could be dangerous for their layout.

jacobeubanks:
Love the refresh! I think it's important for potential users to see a more modern store demo and this is a big step toward that.

kennyadsl:
@mfrecchiami can you please rebase this one as well when you have time? (no hurry!)

kennyadsl:
@mfrecchiami `sample/db/samples/products.rb` is changed on `master`, can you please rebase when you have time? 🙏

mfrecchiami:
@kennyadsl now the branch is updated!

kennyadsl:
> It might be worth noting the changes to the attachment sizes

@ericsaupe yes, a CHANGELOG entry for that should be enough, what do you think? If someone is relying on those sizes we can provide instructions to easily change them back.

Issue #3271, opened 2019-07-15
CircleCI MySQL builds are broken


kennyadsl:
We have all the CircleCI MySQL builds broken. 

Here's an example: https://circleci.com/gh/solidusio/solidus/12551

I think this could be due to [recent updates of the CircleCI docker images](https://hub.docker.com/r/circleci/mysql/tags), here's the error:

```
LoadError:
  Error loading the 'mysql2' Active Record adapter. Missing a gem it depends on? libmariadbclient.so.18: cannot open shared object file: No such file or directory - /home/circleci/solidus/vendor/bundle/ruby/2.5.0/gems/mysql2-0.5.2/lib/mysql2/mysql2.so
```

kennyadsl:
Before each specs run our CircleCI config retrieves the cached gems saved into a folder shared between all builds. Gems cached include native extensions that are needed for some gems to work on the OS used (eg. nokogiri, mysql). At the moment I think that the recent mysql docker image update changed some path or library used by our cached native extension.

The only thing that we can do is clear the cache (by bumping the key used) and regenerate native extension for that specific mysql gem version using the new docker image released.

I'm doing some experiments [here](https://circleci.com/gh/nebulab/solidus/tree/kennyadsl%2Ftest-circle-no-cache) and if that works I'll submit a PR soon.

kennyadsl:
Closed with #3272

PR #3272, opened 2019-07-15
Bump CircleCi gems cache key

kennyadsl:
**Description**

Ref #3271 

The docker image that we are using for MySQL has been recently updated: https://hub.docker.com/r/circleci/mysql

This update changed some internal path or library that were cached into our gem cache in CircleCI for `mysql2 0.5.2`. This happens because native extensions are cached along with the gems and are not rebuilt at every job run.

To clear CircleCI cache we can only change the cache key to rebuild it from scratch, as described in their docs:

https://circleci.com/docs/2.0/caching/#clearing-cache

**Checklist:**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,14 +44,14 @@ commands:
 
       - restore_cache:
           keys:
-            - solidus-gems-v1-{{ .Branch }}
-            - solidus-gems-v1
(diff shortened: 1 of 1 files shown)


kennyadsl:
I'm also making an attempt to enter the CircleCI container via `ssh` and run `bundle pristine` which should reinstall all gems and reinstall native extensions.

PR #3274, opened 2019-07-17
Update Tab colors with its own variables

mfrecchiami:
Avoid using navbar variables and general variables color, about Tabs.
Dedicated variables has been added.

This is necessary to fix the color of the active tab item, and to avoid such mistakes on the future.

white_color

The right view:

corrected_view


kennyadsl:
The build is actually 🍏 https://circleci.com/workflow-run/a6e7a7cc-2124-40aa-8f8c-8e4037660a97

PR #3292, opened 2019-08-03
Rename one letter variables

jacobherrington:
**Description**

This is a small tweak, personally, I find single-character variable names like these to be less than ideal. IMO no reason to not change these. 🤠 

**Checklist:**

### DIFF
--- a/backend/app/helpers/spree/admin/navigation_helper.rb
+++ b/backend/app/helpers/spree/admin/navigation_helper.rb
@@ -33,7 +33,7 @@ module Spree
         elsif content_for?(:page_title)
           content_for(:page_title)
         elsif admin_breadcrumbs.any?
-          admin_breadcrumbs.map{ |x| strip_tags(x) }.reverse.join(' - ')
+          admin_breadcrumbs.map { |breadcrumb| strip_tags(breadcrumb) }.reverse.join(' - ')
         else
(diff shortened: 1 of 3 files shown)


jacobherrington:
@spaghetticode There are a bunch of these throughout the codebase. I might go in and do a more sweeping change in the future, but it would also make a good first-time contribution for anyone new to open source 😉

spaghetticode:
@jacobherrington I think that's a great idea! I will open an issue and mark it as beginner-friendly

backend/app/helpers/spree/admin/navigation_helper.rb
```diff
@@ -33,7 +33,7 @@ def admin_page_title
         elsif content_for?(:page_title)
           content_for(:page_title)
         elsif admin_breadcrumbs.any?
-          admin_breadcrumbs.map{ |x| strip_tags(x) }.reverse.join(' - ')
+          admin_breadcrumbs.map{ |breadcrumb| strip_tags(breadcrumb) }.reverse.join(' - ')
```
kennyadsl: What about taking advantage of the change on this line and also add a space after `map` for consistency?

jacobherrington: Sure, good call. 👍

PR #3293, opened 2019-08-03
Change nil check to use safe navigation operator

jacobherrington:
**Description**

~This change is fairly subjective, but it does remove a duplicate method
call and in my opinion clarifies the intention of the code~

This replaces a boolean && with the safe navigation operator.

**Checklist:**

### DIFF
--- a/api/app/helpers/spree/api/api_helpers.rb
+++ b/api/app/helpers/spree/api/api_helpers.rb
@@ -177,7 +177,7 @@ module Spree
       ]
 
       def variant_attributes
-        if @current_user_roles && @current_user_roles.include?("admin")
+        if @current_user_roles&.include?("admin")
           @@variant_attributes + [:cost_price]
         else
           @@variant_attributes


api/app/helpers/spree/api/api_helpers.rb
```diff
@@ -177,15 +177,16 @@ def required_fields_for(model)
...
           @@variant_attributes + [:cost_price]
         else
           @@variant_attributes
         end
       end
 
       def total_on_hand_for(object)
-        object.total_on_hand.finite? ? object.total_on_hand : nil
+        total_on_hand = object.total_on_hand
+        total_on_hand if total_on_hand.finite?
```
kennyadsl: I'm not sure about my feelings on this one. This removes clarity in reading the code since the reader has to think that if the last if statement on a method doesn't result in any code being run, it returns nil, which is not explicitly stated. Is this change a Rubocop suggestion?

ericsaupe: I agree, this doesn't read as cleanly imo

jacobherrington: This was a code smell from reek, but if this is not preferable we don't need to make the change.

I have changed the PR just to swap that nil check with the safe navigation operator.

Issue #3205, opened 2019-05-10
Promotion eligibility? usage count


cedum:
When performing the eligibility check for a promotion, the usage limit check should not count the promotable that has been already applied for the promotion.
I suspect the same must be done also when re-computing the amounts in promo actions, otherwise, it sets them to zero.

Here's a test case (I did my best to explain it without letting my brain go into a loop 💥):
- Given I have an active promotion with 1 usage limit
- Given I have a completed order with the promotion above applied
- When I edit and save the promotion' adjustment and the promo has _0 usages left_
  - The adjustment is saved as _ineligible_ and amount set to `0`
  - The promotion changes to 1 usage left
- When I edit and save again the promotion' adjustment
  - The adjustment is saved as eligible and amount to the promo discount
  - The promotion changes back to "0 usages left"

Rinse and repeat to switch the adjustment to ineligible and back

Reproduced against solidus versions: `2.9.0.alpha` and `2.7.1`

loicginoux:
Just checking but was your order completed ? it should only count promotion usage for completed order normally.

cedum:
@loicginoux yes, tested it on a completed order (I've updated the issue description to reflect this)

kennyadsl:
I think you are right, this happens because the eligibility check is done when there are no eligible "usage" left. I think we should add that "usage" to the count somehow to make it work.

PR #3278, opened 2019-07-19
Attempt to fix flaky specs

kennyadsl:
**Description**
This is an attempt to fix a couple of flaky specs that we have:

- https://circleci.com/gh/nebulab/solidus/2687#tests/containers/1 🔴 
- https://circleci.com/gh/nebulab/solidus/2686#tests/containers/1 🔴 

Using a more standard way of checking HTML elements state for Capybara could fix the issue since, using the current code, it could not be waiting for the element to change state/appear on the page.

Labeled as WIP. I'd like to re-run specs a lot of times before merge. Progress can be monitored here:

https://circleci.com/gh/nebulab/workflows/solidus/tree/kennyadsl%2Ffix-flaky-attempt

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
~- [ ] I have updated Guides and README accordingly to this change (if needed)~
~- [ ] I have added tests to cover this change (if needed)~

PR #3297, opened 2019-08-08
Update usage count in Promotion eligibility check

filippoliverani:
**Description**

This PR addresses issue #3205.

Eligibility check on an Adjustment bound to a Promotion with a usage limit should not include in usage count the current Adjustment itself.

With this PR the eligibility check does not consider the Order of the Adjustment it is applied to.
I updated the check adding an optional array of Orders to exclude from the computation (in the same vein as Promotion `#used_by?`) to avoid any breaking change to Promotion and PromotionCode puplic APIs.

**Checklist:**

### DIFF
--- a/core/app/models/spree/adjustment.rb
+++ b/core/app/models/spree/adjustment.rb
@@ -56,6 +56,14 @@ module Spree
     extend DisplayMoney
     money_methods :amount
 
+    def self.adjusted_orders(excluded_orders: [])
+      eligible.
+      joins(:order).
+      merge(Spree::Order.complete).
+      where.not(spree_orders: { id: excluded_orders }).
+      distinct
+    end
+
     def finalize!
       update!(finalized: true)
     end
--- a/core/app/models/spree/adjustment.rb
+++ b/core/app/models/spree/adjustment.rb
@@ -56,6 +56,10 @@ module Spree
     extend DisplayMoney
     money_methods :amount
 
+    # Returns Adjustments of completed Orders.
+    #
+    # @param excluded_orders [Array<Spree::Order>] Orders to exclude from query
+    # @return [ActiveRecord::Relation] Scoped Adjustments
     def self.in_completed_orders(excluded_orders: [])
       joins(:order).
       merge(Spree::Order.complete).
--- a/core/app/models/spree/adjustment.rb
+++ b/core/app/models/spree/adjustment.rb
@@ -56,8 +56,7 @@ module Spree
     extend DisplayMoney
     money_methods :amount
 
-    def self.adjusted_orders(excluded_orders: [])
-      eligible.
+    def self.in_completed_orders(excluded_orders: [])
       joins(:order).
       merge(Spree::Order.complete).
       where.not(spree_orders: { id: excluded_orders }).
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -177,13 +177,9 @@ module Spree
     #
     # @return [Integer] usage count
     def usage_count(excluded_orders = [])
-      Spree::Adjustment.eligible.
-        promotion.
+      Spree::Adjustment.promotion.
+        adjusted_orders(excluded_orders: excluded_orders).
         where(source_id: actions).
-        joins(:order).
-        merge(Spree::Order.complete).
-        where.not(spree_orders: { id: excluded_orders }).
-        distinct.
         count(:order_id)
     end
 
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -126,9 +126,11 @@ module Spree
     # called anytime order.recalculate happens
     def eligible?(promotable, promotion_code: nil)
       return false if inactive?
-      return false if usage_limit_exceeded?(excluded_orders: [promotable])
-      return false if promotion_code&.usage_limit_exceeded?(excluded_orders: [promotable])
       return false if blacklisted?(promotable)
+
+      excluded_orders = eligibility_excluded_orders(promotable)
+      return false if usage_limit_exceeded?(excluded_orders: excluded_orders)
+      return false if promotion_code&.usage_limit_exceeded?(excluded_orders: excluded_orders)
 
       !!eligible_rules(promotable, {})
     end
@@ -260,6 +262,16 @@ module Spree
 
       errors.add(:apply_automatically, :disallowed_with_code) if codes.any?
       errors.add(:apply_automatically, :disallowed_with_path) if path.present?
+    end
+
+    def eligibility_excluded_orders(promotable)
+      if promotable.is_a?(Spree::Order)
+        [promotable]
+      elsif promotable.respond_to?(:order)
+        [promotable.order]
+      else
+        []
+      end
     end
   end
 end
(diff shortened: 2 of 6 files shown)


kennyadsl:
@filippoliverani Hey there! Can you please rebase against master now that https://github.com/solidusio/solidus/pull/3278 has been merged? It should fix the flaky spec, thanks!

core/app/models/spree/adjustment.rb
```diff
@@ -56,6 +56,14 @@ class Adjustment < Spree::Base
     extend DisplayMoney
     money_methods :amount
 
+    def self.adjusted_orders(excluded_orders: [])
```
kennyadsl: I have a couple of questions about this class method/scope:

1. Maybe a different name could help understand what this method does: what about a simpler `in_completed_orders()`? I think that maybe with `adjusted_orders()` could not be immediately clear that it's just searching into completed orders.
2. I'm not sure that `eligible` should be part of this method. Chaining it from outside adds a bit of duplication but makes the code more flexible IMO, what do you think?

filippoliverani:
@kennyadsl Thank you for your suggestions, I'll rebase and resolve the problems 👍 .

core/app/models/spree/adjustment.rb
filippoliverani: I agree on both points.
At first I had used a much longer name but it was quite hard to read, `in_completed_orders` strikes a good balance.

core/app/models/spree/promotion.rb
```diff
@@ -166,22 +168,20 @@ def products
     # Whether the promotion has exceeded it's usage restrictions.
     #
     # @return true or false
-    def usage_limit_exceeded?
+    def usage_limit_exceeded?(excluded_orders = [])
```
kennyadsl: We need to also update the YARD documentation here.

kennyadsl: Also, can we use a keyword argument here so it's clear, even when this method is called, that the array contains orders to exclude? I feel like this 

```
usage_limit_exceeded?([promotable])
```

could be better for the reader in this form:

```
usage_limit_exceeded?(excluded_orders: [promotable])
```

core/app/models/spree/promotion.rb
```diff
@@ -166,22 +168,20 @@ def products
...
     #
     # @return [Integer] usage count
-    def usage_count
-      Spree::Adjustment.eligible.
-        promotion.
-        where(source_id: actions.map(&:id)).
-        joins(:order).
-        merge(Spree::Order.complete).
-        distinct.
+    def usage_count(excluded_orders = [])
```
kennyadsl: Same for YARD documentation and keyword argument

filippoliverani:
Thank you @kennyadsl, I'm still very new to Solidus and your help is much appreciated!

core/app/models/spree/promotion.rb
filippoliverani: > We need to also update the YARD documentation here.

Missed that 😱 



> can we use a keyword argument here so it's clear, even when this method is called, that the array contains orders to exclude?

Absolutely agree, I used a positional argument just to be consistent with `#used_by?` but here I'd prefer a keyword one too.

core/app/models/spree/promotion.rb
```diff
@@ -126,9 +126,10 @@ def activate(order:, line_item: nil, user: nil, path: nil, promotion_code: nil)
     # called anytime order.recalculate happens
     def eligible?(promotable, promotion_code: nil)
       return false if inactive?
-      return false if usage_limit_exceeded?
-      return false if promotion_code && promotion_code.usage_limit_exceeded?
+      return false if usage_limit_exceeded?(excluded_orders: [promotable])
```
cedum: I think we must handle here the different cases a `promotable` object can be:
- `Spree::Order`
- `Spree::LineItem`
- `Spree::Shipment`

You can simulate these cases by creating different promos with different [_Action_' adjustment types.](https://cedum.s3-eu-west-2.amazonaws.com/yfzqwsbz4).

core/app/models/spree/adjustment.rb
```diff
@@ -56,6 +56,17 @@ class Adjustment < Spree::Base
...
 
+    # Returns Adjustments of completed Orders.
+    #
+    # @param excluded_orders [Array<Spree::Order>] Orders to exclude from query
+    # @return [ActiveRecord::Relation] Scoped Adjustments
+    def self.in_completed_orders(excluded_orders: [])
+      joins(:order).
+      merge(Spree::Order.complete).
+      where.not(spree_orders: { id: excluded_orders }).
+      distinct
```
cedum: Not strictly related to this PR, since the `distinct` comes from the previous implementation. Just a heads-up.

Last time I checked it, the generated query was very slow because of `DISTINCT(*)` on a relatively big table (~400k rows). It was something like ~1ms query response with `group by` vs ~500ms using `distinct(*)` on my local machine.
It's worth double checking it and, if confirmed, opening a separate issue.

filippoliverani:
@cedum Thank you for the feedback, in the meantime I'll check the existing specs too :+1:!

core/app/models/spree/promotion.rb
filippoliverani: I had this doubt but I thought that I could ignore LineItems and Shipments when computing eligibility check on them because usage count was not directly affected.

Do you think that updating`#eligible?`to handle non Order promotables excluding Orders to which LineItems and Shipments belong to would be enough?

core/app/models/spree/adjustment.rb
filippoliverani: This could be the right time to speed up the query, I'll profile both versions with some test data, thanks.

core/app/models/spree/promotion.rb
cedum: > Do you think that updating `#eligible?` to handle non Order promotables excluding Orders to which LineItems and Shipments belong to would be enough?

👍 yes. AFAIK, a promotion's usage count is always the number of distinct orders that promo is applied. I cannot think of cases where it's counted by a different object.

core/app/models/spree/adjustment.rb
filippoliverani: I tested it on about 500k rows of random genereated rows but I couldn't find any big difference. If you have a more realistic dump without sensitive data could you please share it with me?

filippoliverani:
@cedum now `#eligible?` should handle all promotable types. I updated and refactored a bit its specs to increase coverage.
I saw some other duplicated and maybe obsolete specs in `promotion_spec.rb` but maybe it would be better to clean them up in an other PR,  WDYT?

core/app/models/spree/adjustment.rb
cedum: 🤔 thanks for looking into this. I'll try to create a reproducible case with dummy data then eventually will open a separate issue with steps to reproduce it.

cedum:
> @cedum now `#eligible?` should handle all promotable types. I updated and refactored a bit its specs to increase coverage.

💯LGTM

> I saw some other duplicated and maybe obsolete specs in `promotion_spec.rb` but maybe it would be better to clean them up in an other PR, WDYT?

agree 🤝

PR #3300, opened 2019-08-09
Add docs for partials that need to be provided

skukx:
**Description**

Due to !2513 we need to provide a partial for the api to render a
payment method.

**Checklist:**

### DIFF
--- a/core/app/models/spree/payment_method.rb
+++ b/core/app/models/spree/payment_method.rb
@@ -199,6 +199,9 @@ module Spree
     #     4. app/views/spree/admin/payments/source_views/_{partial_name}.html.erb
     #     The view that represents your payment method on orders in the backend
     #
+    #     5. app/views/spree/api/payments/source_views/_{partial_name}.json.jbuilder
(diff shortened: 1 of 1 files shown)


core/app/models/spree/payment_method.rb
```diff
@@ -199,6 +199,9 @@ def display_on
     #     4. app/views/spree/admin/payments/source_views/_{partial_name}.html.erb
     #     The view that represents your payment method on orders in the backend
     #
+    #     5. app/views/spree/api/payments/source_views/_{partial_name}.json.jbuilder
+    #     The view that represents your payment method on orders in the backend
```
kennyadsl: `in the backend` ➡️ `in the api`?

skukx: Good Catch!

kennyadsl:
@skukx not sure what's happened here, can you please restart the build in your CircleCI account?

skukx:
I wonder if the force push broke the integration with circle

PR #3305, opened 2019-08-19
Raise exception if dividing by 0

ericsaupe:
While this should never happen it is possible to call this method
and subsequently try to save a NaN value if the line_item has
already been fully cancelled.

**Description**

This came up in our store today. I was unable to recreate it but somehow one of our customer service representatives short shipped an item and the adjustment, and subsequent adjustment total on the order, went to `NaN`. Looking into the calculation it looks like it may be possible for the short ship compute amount to return that result. This PR raises an exception if, somehow, that method is called and tries to divide by 0.

**Checklist:**


core/app/models/spree/unit_cancel.rb
```diff
@@ -45,6 +45,9 @@ def weighted_line_item_amount(line_item)
   end
 
   def quantity_of_line_item(line_item)
-    BigDecimal(line_item.inventory_units.not_canceled.reject(&:original_return_item).size)
+    not_canceled_inventory_units_count = line_item.inventory_units.not_canceled.reject(&:original_return_item).size
+    raise "Line Item does not have any inventory units available to cancel" if not_canceled_inventory_units_count.zero?
```
jacobherrington: Could we put this in i18n?

kennyadsl: I think we should raise this error in the `weighted_line_item_amount ` method since this one is only a getter for the size and maybe, semantically, is allowed to return 0. `weighted_line_item_amount`, instead, cannot perform the operation it has been written for so I think that's the right place where to put this raise. What do you think?

ericsaupe: I like it. Makes sense to raise when the calculation is about to fail.

kennyadsl:
@ericsaupe just thinking: wouldn't this lead to another exception [ZeroDivisionError](https://ruby-doc.com/core-2.4.0/ZeroDivisionError.html) without the need to explicitly raise it?

ericsaupe:
@kennyadsl because `quantity_of_line_item` is a `BigDecimal` it actually returns `BigDecimal('-Infinity')` when `quantity_of_line_item` is `BigDecimal(0)`

PR #3308, opened 2019-08-21
Making taxon form to render attachment definitions dynamically

softr8:
**Description**
Solidus now allows to inject easily more attachment definitions via
configurable modules (#3237), if a user adds a new attachment, it would be needed
to override the partial _form and add this new attachment, this change
takes all definitions and render the inputs dynamically.

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/taxons/_form.html.erb
+++ b/backend/app/views/spree/admin/taxons/_form.html.erb
@@ -20,11 +20,7 @@
       </div>
     <% end %>
 
-    <%= f.field_container :icon do %>
-      <%= f.label :icon %><br />
-      <%= f.file_field :icon %>
-      <%= image_tag f.object.icon(:mini) if f.object.icon_present? %>
-    <% end %>
+    <%= render "spree/admin/taxons/attachment_forms/#{f.object.attachment_partial_name}", f: f %>
   </div>
 
   <div class="col-5">
--- a/core/app/models/spree/taxon/paperclip_attachment.rb
+++ b/core/app/models/spree/taxon/paperclip_attachment.rb
@@ -18,4 +18,8 @@ module Spree::Taxon::PaperclipAttachment
   def icon_present?
     icon.present?
   end
+
+  def attachment_partial_name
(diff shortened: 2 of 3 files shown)


spaghetticode:
@softr8 can you please handle the small nitpick from Hound? 

Apart from that, I think this is a great improvement, thank you for noticing that there was room for it 👍 ❤️

backend/app/views/spree/admin/taxons/_form.html.erb
```diff
@@ -20,10 +20,12 @@
       </div>
     <% end %>
 
-    <%= f.field_container :icon do %>
-      <%= f.label :icon %><br />
-      <%= f.file_field :icon %>
-      <%= image_tag f.object.icon(:mini) if f.object.icon_present? %>
+    <% f.object.class.attachment_definitions.each do |attachment, definition| %>
```
kennyadsl: Doesn't this work for Paperclip only? I don't think ActiveStorage has a `attachment_definitions` class method. Maybe we should also make this partial loaded conditionally if we are using the paperclip adapter, similar to what we do for payment method partials. What do you think?

core/app/models/spree/taxon/paperclip_attachment.rb
```diff
@@ -18,4 +18,8 @@ module Spree::Taxon::PaperclipAttachment
   def icon_present?
     icon.present?
   end
+
+  def attachment_partial_name
+    'paperclip_attachment'
```
kennyadsl: Maybe it's better to have only `paperclip` here since it could be used also in other context. 

We can use 

```ruby
<%= render "spree/admin/taxons/attachment_forms/#{f.object.attachment_partial_name}", f: f %>
```

So we'll be able to have a folder for the forms and one for the show views (if needed in the future). This is exactly the same pattern we are using for payments, see [here](https://github.com/solidusio/solidus/tree/df79573cc09335dd8d2b193d8f54cd0e03a2cc05/backend/app/views/spree/admin/payments). WDYT?

PR #3316, opened 2019-08-28
Added empty cart button in admin cart

ericsaupe:
**Description**

The frontend has an empty cart button and the API supports an
empty cart call. This adds that functionality to the backend
interface.

Screen Shot 2019-08-28 at 12 06 52 PM

**Checklist:**

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/cart/empty_cart_button.js
+++ b/backend/app/assets/javascripts/spree/backend/views/cart/empty_cart_button.js
@@ -1,0 +1,29 @@
+Spree.Views.Cart.EmptyCartButton = Backbone.View.extend({
+  initialize: function() {
+    this.listenTo(this.collection, 'update', this.render);
+    this.render();
+  },
+
+  events: {
+    "click": "onClick"
+  },
+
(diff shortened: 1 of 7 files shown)


backend/app/assets/javascripts/spree/backend/views/cart/empty_cart_button.js
```diff
@@ -0,0 +1,29 @@
...
+  },
+
+  onClick: function(e) {
+    e.preventDefault()
+    if (!confirm(Spree.translations.are_you_sure_delete)) {
+      return;
+    }
+
+    this.model.empty({
+      success: () => {
```
pelargir: Because of this one line, Uglifier has to be put in harmony mode or Solidus 2.10 won't run. This should be changed back to ES5 syntax for better compatibility. There is no need to run Uglifier in harmony mode just to parse this one line.

tvdeyen: Good catch, Matt! We would need to either process this file with Babel or update the syntax. I vote for the latter. 

@pelargir want to send a PR?

Issue #3311, opened 2019-08-23
Background color of backend tabs is too dark


delphaber:
**Solidus Version**:
`master` ( == `4828eea14e41e0062c0946c89ed11631a96fbf26`)

**To Reproduce**
1. Go to any page with a tab, like [http://localhost:3000/admin/users/1/edit](http://localhost:3000/admin/users/1/edit)

**Current behavior**
In solidus backend I noticed that non active tabs have a dark background color, that makes the tab title unreadable.

**Expected behavior**
A lighter background color, like this:

<img width="1008" alt="Schermata 2019-08-23 alle 17 01 52" src="https://user-images.githubusercontent.com/139284/63602455-c83b3d00-c5c7-11e9-8011-71d6e5bd608c.png">

**Screenshots**
<img width="1003" alt="Schermata 2019-08-23 alle 16 53 23" src="https://user-images.githubusercontent.com/139284/63601899-9f667800-c5c6-11e9-9dab-b82fe95fb541.png">

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Firefox dev edition
 - Version: 69.0b15

**Additional context**
Tabs background color is defined [here](https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/globals/_variables.scss#L58):

`$color-tab-bg:                  very-light($color-dark-light, 4) !default;` 

I suppose the [`very-light` function](https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/globals/_functions.scss#L2) is not working as expected 🤔 , but I don't know why.

I fixed the problem on my project by creating a file in `app/assets/stylesheets/spree/backend/globals/_variables_override.scss` with:

```scss
$color-dark-light: #808C9E;
$color-tab-bg: lighten($color-dark-light, 40%);
```

( I don't know why, but placing the file in `vendor/assets/stylesheets/spree/backend/globals/_variables_override.scss` did not work. )

Sass versions:

```
    sassc (2.1.0)
      ffi (~> 1.9)
    sassc-rails (2.1.2)
      railties (>= 4.0.0)
      sassc (>= 2.0)
      sprockets (> 3.0)
      sprockets-rails
      tilt
```

kennyadsl:
@delphaber thanks for reporting, we'll take a look as soon as possible!

kennyadsl:
Closed with https://github.com/solidusio/solidus/pull/3320

PR #3320, opened 2019-08-30
Fix tab background color, too dark #3311

Ajmal:
**Description**

In solidus backend I noticed that non active tabs have a dark background color, that makes the tab title unreadable.

Ref #3311

**Checklist:**


tvdeyen:
There is no need to wait 24h for a CSS :)

Issue #2721, opened 2018-05-01
Not update available on date


sechix:
### Steps to reproduce
In products show a different date in input datapicker field that in database
![image](https://user-images.githubusercontent.com/24247266/39479469-9c87fcb0-4d65-11e8-8194-458574a6f440.png)

### System configuration
**Solidus Master last version**

jtapia:
Hi, are you using translations?

jacobherrington:
Hey @sechix I can't replicate this issue, it seems to be working for me. 

If you hit update at the bottom of the product page, does your database match?

Could you post some reproduction steps?

Thanks!

sechix:
![image](https://user-images.githubusercontent.com/24247266/41254387-996aef64-6dc3-11e8-8cdd-854979d8da11.png)


Update product:
![image](https://user-images.githubusercontent.com/24247266/41254415-a96e8bc8-6dc3-11e8-9d9a-109379116e45.png)

After update (even refreshing page):

![image](https://user-images.githubusercontent.com/24247266/41254452-cb241940-6dc3-11e8-9a95-cc00fd196eba.png)

jtapia:
with `es-MX` you can see this:
![screen shot 2018-06-13 at 11 40 43 am](https://user-images.githubusercontent.com/957520/41370131-69e800d4-6f0c-11e8-9cca-d9668b4eb0c8.png)

and later I got a `missing translation` error, I created this PR https://github.com/solidusio/solidus_i18n/pull/129 to fix that, with `es, es-EC and es-CL` is working fine

Draiken:
I have the same issue. Seems to be related to localization since I use pt-BR and have the exact same issue. My guess is that the datepicker format has a specific format and i18n changes the default date formats, causing the disconnect between the actual value from the input and the datepicker's display

Draiken:
To expand a bit: on the context of solidus_i18n you see a field called `spree.date_picker.format` and it makes sense to localize that to the local format. However by doing that you break the datepickers on the admin (haven't tested others, but I imagine everywhere that this helper is used will do that).

I propose we hard-code the datepicker format to something that always works, since this is not a presentational value but instead a value that is used by a third party library.

PR #3321, opened 2019-08-30
Fix issue with available_on datepicker

coorasse:
This PR fixes https://github.com/solidusio/solidus/issues/2721.
I'd like to fix it already in version 2.7 and release a 2.7.4 if possible.
The datepicker should take the `js_format` from i18n file (not in use anywhere else).
I'll open also a PR on solidus_i18n to set it already correctly for the other locales.

I am not sure if this is a breaking change. If someone has overridden `js_format` with something weird it will probably stop working.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/datepicker.js
+++ b/backend/app/assets/javascripts/spree/backend/datepicker.js
@@ -11,7 +11,8 @@ Spree.ready(function(){
   });
 
   $('.datepicker').flatpickr({
-    allowInput: true
+    allowInput: true,
+    dateFormat: Spree.t('date_picker.js_format')
   });
 
   // Handle range dates
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1467,7 +1467,7 @@ en:
     date_picker:
       first_day: 0
       format: "%Y/%m/%d"
-      js_format: yy/mm/dd
+      js_format: Y/m/d
     date_range: Date Range
     default: Default
     default_refund_amount: Default Refund Amount


kennyadsl:
Thanks, @coorasse. I'm rerunning specs, failures do not seem related but I did never see those before. Let's see what happens.

aldesantis:
@coorasse we're going through old PRs and this seems like a fix we want to have. Unfortunately, 2.7 is not maintained anymore, so I don't know if it makes sense to add it there. If you still have the time, would you be willing to rebase against master or 2.8?

kennyadsl:
For reference, if this is still valid, we should make this PR against master (and eventually backport).

kennyadsl:
@coorasse I rebased this against master and resolved a conflict. If specs pass, I'll go head and add the backport labels to have this on recent support versions of Solidus and will ask for a review by other core team members. Thanks again!

kennyadsl:
@waiting-for-dev do you have any idea about why ensure and triage actions have not been triggered here? 🤔

waiting-for-dev:
> @waiting-for-dev do you have any idea about why ensure and triage actions have not been triggered here? 🤔

@kennyadsl, I think the [synchronize activity type](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) doesn't include changing the base branch. However, making a change to the commits on the head branch does. I forced another push and the workflow has run now.

kennyadsl (approved):

waiting-for-dev (approved):

elia (approved):

PR #3322, opened 2019-08-30
Style collapsing sidebar

mfrecchiami:
Ref. to @seand7565 PR: https://github.com/solidusio/solidus/pull/3166
Closes https://github.com/solidusio/solidus/issues/2961

In the first prototype about this new feature, I want to let the Solidus mark only when the sidebar was collapsed; but then I remember that the logo on header is customizable and many times there are no mark (or no pretty ways) to show in a small place.
So I opted to leave the logo container always visible, also when the user choose to collapse the sidebar: so, no matter if you've uploaded a custom logo or not, and furthermore you know always which admin area you are logged in.

Also the languages select shows in a smart way when the sidebar is collapsed.

uncollapsed

collapsed

collapsing-nav720

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
@@ -2,6 +2,91 @@ $padding-x-navbar: 26px;
 $padding-y-navbar: 13px;
 $padding-y-navbar-submenu: 9px;
 
+#admin-nav-toggle {
+  border-bottom: $border-sidebar;
+  padding: ($padding-y-navbar/ 2) $padding-x-navbar;
+  width: 100%;
+  text-align: left;
+  color: $color-navbar;
+  font-weight: $font-weight-bold;
+  position: relative;
+  overflow: hidden;
+  display: block;
+  height: 39px;
+  z-index: 1;
+
+  span {
+    font-size: $font-size-sm;
+    font-weight: $font-weight-bold;
+  }
+
+  &.btn {
+    background-color: transparent;
+  }
+
+  &.fa {
+    text-indent: 2em;
+
+    &:before {
+      color: $color-icon-navbar;
+      font-weight: $font-weight-normal;
+      position: absolute;
+      left: 1em;
+      top: 0;
+      line-height: 39px;
+      transform: translateX(-50%);
+      transition: transform .5s ease-in-out;
+    }
+
+    &:focus {
+      outline: 0;
+      box-shadow: none;
+    }
+  }
+}
+
+.admin-nav-hidden {
+  padding-left: $width-sidebar-collapsed;
+
+  .admin-nav,
+  .admin-nav-footer {
+    width: $width-sidebar-collapsed;
+  }
+
+  .text {
+    transform: translateX(-50px);
+    display: none;
+  }
+
+  .admin-login-nav a {
+    text-overflow: clip;
+  }
+
+  .brand-link {
+    overflow: hidden;
+
+    img {
+      max-width: 125px;
+    }
+  }
+
+  .admin-nav-menu {
+    li {
+      &.selected:not(:hover) > ul  {
+        visibility: hidden;
+      }
+
+      &.selected:not(:hover) .admin-subnav {
+        display: none;
+      }
+    }
+  }
+
+  .admin-nav-menu .selected li {
+    padding-left: 0;
+  }
+}
+
 nav.menu {
   ul {
     list-style: none;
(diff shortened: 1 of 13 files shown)


mfrecchiami:
Hi @tvdeyen  thanks for your feedback!

- "choose language" official icon: we evaluated it but icons in the sidebar are really small and this official icon is pretty detailed and we thought that was better choose something simpler (like the globe). Maybe we could consider to resize menu icons, but I'm not so sure that the language icon will display better than now. (In any case I'll try to do it and will send some screenshots with these changes 😉 )

- The icons in the lower left area do not have the same color as the main menu items: I didn't change it and I think that it was done at the beginning to differentiate the two types of menus, and also because the "user" icon is the same icon used for _UserS_ in the main menu, I guess.

We should update the _UserS_ icon with this one: https://fontawesome.com/v4.7.0/icon/users
but, again, we should to consider to resize the menu icons;
we could also think about another background color for the bottom left menu.

In any case I would split the restyle of bottom left menu, in another PR.

TODO:

- [ ] Hide the Language name if sidebar is collapsed [test it on main browsers]
- [ ] The chevron is the other way round if I collapsed the sidebar and reload the page

🙏 thank you

mfrecchiami:
Hi e @tvdeyen !
I fixed the two issues on TODO list 💪 
Now, I'm changing the language icon (globe to [official icon](https://fontawesome.com/v4.7.0/icon/language)). I attach two screenshots, one with the globe icon and one with the language icon; I think that the globe one is more readable than the official one. What do you think?

globe-icon

official-language-icon

tvdeyen:
You are right. The globe looks better than the official icon. I thought Don’t Awesome also has a version of the official icon, but they don’t. That’s good for now. Thanks for making these two screens. That really helped to make the decision. 

I also agree we should move the styling of the bottom menu into a separate PR. And I agree in that we should use larger icons in the main menu as well. But this is also something for later. 

Again. Great work on this PR.

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -2,6 +2,92 @@ $padding-x-navbar: 26px;
...
+    width: $width-sidebar-collapsed;
+  }
+
+  .text {
+    transform: translateX(-50px);
+    opacity: 0;
+  }
+
+  .admin-login-nav a {
+    text-overflow:clip;
```
spaghetticode: nitpick: missing space after `:`

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -177,8 +239,59 @@ nav.menu {
...
+    background: $color-white;
+    box-shadow: 1px 0 4px 0 rgba(0, 0, 0, 0.1);
+    border-radius: 0 4px 4px 4px;
+    border: 1px solid $color-border;
+    @include caret($direction: left, $color-caret: $color-white);
+
+  &:before {
+    z-index: 1;
+    top: 1.5em;
+    pointer-events:none;
```
spaghetticode: missing space after `:`

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -177,8 +239,59 @@ nav.menu {
...
+    display: block;
+    position: absolute;
+    border-width: 1em 1em;
+    border-style: solid;
+    border-color: transparent;
+    border-right-color: $color-border;
+    left: -2.1em;
+    top: 1.5em;
+    margin-top: -1em;
+    pointer-events:none;
```
spaghetticode: missing space after :

backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
kennyadsl: HoundCI should catch those 🤔

mfrecchiami: @kennyadsl If I remember well, HoundCI notify you when the white space is missed, at the end of the file

kennyadsl:
We found some visual bugs with this PR that we are trying to solve into https://github.com/nebulab/solidus/pull/28. 

Once ready, for this PR, I suggest creating a single commit co-authored with everyone involved. We have a lot of commits here but they are kinda reverting each other work several times so I don't think it's actually helping browsing git history. Let me know your thoughts on this. 🙂

kennyadsl:
Done, I removed transitions for now since they conflict with the jQuery plugin that handles the sticky sidebar elements when a long page is totally scrolled down. I had a real-life conversation with @davidedistefano about this and we both think we can add back transition with other PRs later. 

This is ready to go in my opinion, I'd just wait @seand7565 @mfrecchiami @jacobherrington and @tvdeyen opinion on squashing commits into a [single co-authored one](https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors).

kennyadsl:
Thanks to anyone involved in shipping this feature. We made it! 🎉

PR #3326, opened 2019-09-04
Avoid installing webpacker in sandbox

kennyadsl:
**Description**

With rails/rails#34275 `rails new` option named `--skip_yarn` is changed into `--skip_javascript`. This PR reflects this change and keeps allowing us to still use sprockets for JS in our sandbox.

**TODO**

- [x] ~Using this option with Rails 6, sandbox does not contain any `app/assets/javascripts` folder, so I think Solidus manifests is not loaded correctly. This needs to be investigated.~ My fault, that folder is not needed in the sandbox since the sandbox application will use the manifests defined in the solidus gems. 

**Checklist:**

### DIFF
--- a/lib/sandbox.sh
+++ b/lib/sandbox.sh
@@ -27,7 +27,7 @@ bundle exec rails new sandbox --database="$RAILSDB" \
   --skip-rc \
   --skip-spring \
   --skip-test \
-  --skip-yarn
+  --skip-javascript
 
 if [ ! -d "sandbox" ]; then
   echo 'sandbox rails application failed'


lib/sandbox.sh
```diff
@@ -27,7 +27,7 @@ bundle exec rails new sandbox --database="$RAILSDB" \
   --skip-rc \
   --skip-spring \
   --skip-test \
-  --skip-yarn
+  --skip_javascript
```
tvdeyen: Shouldnt this be?

```suggestion
  --skip-javascript
```

kennyadsl: Yes but I swear that I tried and it worked the same! 😅 Thanks!

PR #3327, opened 2019-09-04
Fix dummy app generator to skip Bootsnap and Webpack in Rails 6

aldesantis:
**Description**

Fixes the dummy app generator, which now fails because Rails 6 uses Bootsnap and Webpacker by default, which are not included in our Gemfile (and not needed for the dummy app).

**Checklist:**

### DIFF
--- a/core/lib/generators/spree/dummy/dummy_generator.rb
+++ b/core/lib/generators/spree/dummy/dummy_generator.rb
@@ -42,6 +42,8 @@ module Spree
       opts[:skip_spring] = true
       opts[:skip_test] = true
       opts[:skip_yarn] = true
+      opts[:skip_bootsnap] = true
+      opts[:skip_webpack_install] = true
 
       puts "Generating dummy Rails application..."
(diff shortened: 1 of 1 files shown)


malasaad82:
Hi,
Do we need to remove the bootsnap from /sandbox/config/boot.rb 
require 'bootsnap/setup' # Speed up boot time by caching expensive operations.

And remove it as welll from the gemfile ?
/sandbox/Gemfile
# Reduces boot times through caching; required in config/boot.rb
gem 'bootsnap', '>= 1.4.2', require: false

I tried to install extensions and I does not work,  I got error:
after I run this:
bundle exec bin/rails g solidus_comments:install

```
	3: from bin/rails:3:in `<main>'
	2: from bin/rails:3:in `require_relative'
	1: from /Development/App/solidus/sandbox/config/boot.rb:4:in `<top (required)>'
/Development/App/solidus/sandbox/config/boot.rb:4:in `require': cannot load such file -- bootsnap/setup (LoadError)
```

PR #3328, opened 2019-09-04
Remove Rails 5.1 support

kennyadsl:
**Description**
Our code is still compatible with Rails 5.1 but that version is in its `EOL` phase, so it's not receiving security patches anymore. 

This PR removes support for Rails 5.1, and it's a follow-up of #3333. We should merge this PR after we release the next version, probably `2.10`. 

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/dummy_app/rake_tasks.rb
+++ b/core/lib/spree/testing_support/dummy_app/rake_tasks.rb
@@ -43,10 +43,8 @@ namespace :db do
           [path],
           ActiveRecord::SchemaMigration
         ).migrate
-      elsif Rails.gem_version >= Gem::Version.new('5.2.0')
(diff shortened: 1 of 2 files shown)


kennyadsl:
@tvdeyen I think that document is outdated, according to https://rubyonrails.org/security/

kennyadsl:
I think this is the last one: https://github.com/rails/rails/blob/6d68bb5f695414b1801c1c3952ef955a5b0b6c6a/guides/source/maintenance_policy.md

kennyadsl:
@tvdeyen what about printing a deprecation warning if Rails 5.1 is used and just remove it from the CI?

tvdeyen:
This sounds like a good idea.

kennyadsl:
I've created #3333 and updated this PR, so it's ready to merge after the next release.

PR #3331, opened 2019-09-11
Use default sass function to lighten colors

mfrecchiami:
Noticed that the custom sass `@function very-light` doesn't work
properly and this fact is generating display issues, both in tables
with actions, and in components such as Taxonomies.

### The issue

Schermata 2019-08-23 alle 11 22 12 (1)

Screenshot from 2019-09-10 16_01_37

### Resolved issue

Schermata 2019-09-11 alle 14 47 35

Schermata 2019-09-11 alle 14 47 15

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_messages.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_messages.scss
@@ -40,15 +40,15 @@
   color: $body-color;
 
   &.notice {
-    background-color: rgba(very-light($color-notice, 15), .95);
+    background-color: rgba(lighten($color-notice, 15), .95);
     border-top-color: $color-notice;
   }
   &.success {
-    background-color: rgba(very-light($color-success, 30), .95);
+    background-color: rgba(lighten($color-success, 30), .95);
     border-top-color: $color-success;
   }
   &.error {
-    background-color: rgba(very-light($color-error, 30), .95);
+    background-color: rgba(lighten($color-error, 30), .95);
     border-top-color: $color-error;
   }
 }
--- a/backend/app/assets/stylesheets/spree/backend/globals/_variables.scss
+++ b/backend/app/assets/stylesheets/spree/backend/globals/_variables.scss
@@ -89,8 +89,8 @@ $color-style-guide-table-border:    #CEE1F4;
 
 // Table colors
 $color-tbl-odd:                  $color-white !default;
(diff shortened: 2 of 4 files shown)


mfrecchiami:
Hi @tvdeyen !
The `very-light` function is a custom function:
https://github.com/solidusio/solidus/blob/master/backend/app/assets/stylesheets/spree/backend/globals/_functions.scss

so actually we could remove it 👍

tvdeyen:
Dang, then we should `@deprecate` it first.

tvdeyen:
We probably have to release a fixed supported versions with a locked `sassc` version. We had some issues with lighten colors in Solidus 2.6 and above with sassc 2.2.1. Locking it to 2.0.1 helped.

kennyadsl:
@tvdeyen this fix has been released in Solidus 2.6, if I recall correctly. What is the problem exactly?

tvdeyen:
@kennyadsl due to changes in Sassc in the `lighten` function if you have a custom theme that sets new values for `$color-x` variables all the colors are too dark. Thats why in this change https://github.com/solidusio/solidus/pull/3331/files#diff-c48fd398d057bc5e3c809c376edd3ad9 all the color values have been taken by 5. We could override every scss file that has `very-light` in it and replace it with `lighten` and take the vallue by 5, but this is meh. We locked the sassc version in our code base to fix this, but we probably should lock the version in `solidus_backend` as well to help other stores as well. WDYT?

kennyadsl:
Got it but I'm not sure since this lock would only be needed if someone has custom css that overrides the default Solidus admin theme but this will not allow using most recent Sassc versions to who is not using a custom theme, right? 

We support changing admin colors but if the "unsupported" custom code is inside user applications, that's probably the best place where the lock is needed.

kennyadsl:
I mean, we are also printing a deprecation warning for that function, is it working?

tvdeyen:
> We support changing admin colors but if the "unsupported" custom code is inside user applications, that's probably the best place where the lock is needed.

@kennyadsl it is not unsupported code in our custom code. It is incompatible code inside of `solidus_backend`.

tvdeyen:
Solidus backend makes use of the `very-light` function that is not working anymore in sassc 2.2. And if you replace it with `lighten` you have to take the value by `5` like @mfrecchiami did in this PR. But this  fix has not been backported.

kennyadsl:
Ah, I thought it was backported. I'd go with backporting this fix, then.

PR #3334, opened 2019-09-16
Replace update_attributes with update

aldesantis:
**Description**

`update_attributes` was deprecated in Rails 6 and will be removed in Rails 6.1. This PR replaces all instances of `update_attributes` with `update`.

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/shipments_controller.rb
+++ b/api/app/controllers/spree/api/shipments_controller.rb
@@ -146,7 +146,7 @@ module Spree
       end
 
       def update_shipment
-        @shipment.update_attributes(shipment_params)
+        @shipment.update(shipment_params)
         @shipment.reload
       end
 
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -326,7 +326,7 @@ module Spree
 
     # Update Shipment and make sure Order states follow the shipment changes
     def update_attributes_and_order(params = {})
-      if update_attributes params
+      if update(params)
         if params.key? :selected_shipping_rate_id
           # Changing the selected Shipping Rate won't update the cost (for now)
           # so we persist the Shipment#cost before running `order.recalculate`
--- a/api/app/controllers/spree/api/addresses_controller.rb
+++ b/api/app/controllers/spree/api/addresses_controller.rb
@@ -15,7 +15,7 @@ module Spree
         authorize! :update, @order, order_token
         find_address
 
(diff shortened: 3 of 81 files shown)


api/app/controllers/spree/api/shipments_controller.rb
```diff
@@ -50,7 +50,7 @@ def create
       end
 
       def update
-        @shipment.update_attributes_and_order(shipment_params)
```
tvdeyen: Is this correct?

api/app/controllers/spree/api/checkouts_controller.rb
```diff
@@ -112,7 +112,7 @@ def state_callback(before_or_after = :before)
         send(method_name) if respond_to?(method_name, true)
       end
 
-      def after_update_attributes
```
tvdeyen: Do we want to rename this method as well?

core/app/models/spree/shipment.rb
```diff
@@ -325,8 +325,8 @@ def update_amounts
     end
 
     # Update Shipment and make sure Order states follow the shipment changes
-    def update_attributes_and_order(params = {})
-      if update_attributes params
+    def update_and_order(params = {})
```
tvdeyen: We probably don’t want to rename this method?

core/lib/spree/permitted_attributes.rb
```diff
@@ -9,7 +9,7 @@ module PermittedAttributes
       :address_attributes,
       :address_book_attributes,
       :checkout_attributes,
-      :credit_card_update_attributes,
+      :credit_card_update,
```
tvdeyen: I think this is not meant to be renamed

aldesantis:
@tvdeyen fixed the ones that were not supposed to be there. :)

kennyadsl:
@aldesantis failures seem relevant to the change, can you please take a look when you do have time?

aldesantis:
@kennyadsl we should be good now! I apologize this took three rounds to get right. 🙏

core/app/models/spree/shipment.rb
```diff
@@ -326,7 +326,7 @@ def update_amounts
 
     # Update Shipment and make sure Order states follow the shipment changes
     def update_attributes_and_order(params = {})
-      if update_attributes params
+      if update params
```
kennyadsl: what about taking advantage of the change to add parenthesis here?

aldesantis: @kennyadsl done. :)

PR #3338, opened 2019-09-19
Allow splitting shipments when not tracking inventory

### LINKED ISSUE #2817: Items on Back Order when Track Inventory is false
### Steps to reproduce
* Set config.track_inventory_levels = false
* Create one or two shipment locations
* Submit and order for two products or a quantity of two for a single product
* In the order section, press the Split button
* Choose a different shipment location or even the same shipment location

### Expected behavior
* Item should be shippable

### Actual behavior
* Item says it is back ordered and you can't get out of that state

### System configuration
**Solidus Version**:
2.6.0

**Extensions in use**:
```
gem 'solidus' 
gem 'solidus_auth_devise'
gem 'deface'
gem 'solidus_wishlist', github: 'boomerdigital/solidus_wishlist', branch: 'master'
gem 'solidus_reviews', github: 'solidusio-contrib/solidus_reviews'
gem "searchkick", github: "ankane/searchkick"
gem 'solidus_cmd'
```

https://gist.github.com/brettallred/fe054c6dd4df820b0a510cbb4872f610

nspinazz89:
**Description**

Closes #2817.

When `Spree::Config.track_inventory_levels` is `false` we are not supposed to change inventory levels when shipping orders.

At the moment though, if we try to split a shipment with the following scenario:

- `track_inventory_levels` is false
- there is no inventory in the second stock location for the related stock item

we end up creating a second shipment with items backorderable, making the second shipment impossible to ship.

This is because the FulfilmentChanger class at the moment does not take into account this scenario. But it was currently doing a lot of things that are not required when inventory levels are not tracked.

This fix proposes two different paths, driven by a new argument required in the initialization of the class, that performs a
complex stock movement when we are tracking inventory and a more simple otherwise.

Not passing this new argument to the class' `new` method is deprecated, assuming the old behavior for backward compatibility.

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/shipments_controller.rb
+++ b/api/app/controllers/spree/api/shipments_controller.rb
@@ -127,7 +127,8 @@ module Spree
           current_shipment: @original_shipment,
           desired_shipment: @desired_shipment,
           variant: @variant,
-          quantity: @quantity
+          quantity: @quantity,
+          track_inventory: Spree::Config.track_inventory_levels
         )
 
         if fulfilment_changer.run!
--- a/core/app/models/spree/fulfilment_changer.rb
+++ b/core/app/models/spree/fulfilment_changer.rb
@@ -17,18 +17,27 @@ module Spree
   # @attr [Integer] quantity How many units we want to move
   #
   class FulfilmentChanger
+    TRACK_INVENTORY_NOT_PROVIDED = Object.new.freeze
     include ActiveModel::Validations
 
     attr_accessor :current_shipment, :desired_shipment
-    attr_reader :variant, :quantity, :current_stock_location, :desired_stock_location
+    attr_reader :variant, :quantity, :current_stock_location, :desired_stock_location, :track_inventory
 
-    def initialize(current_shipment:, desired_shipment:, variant:, quantity:)
+    def initialize(current_shipment:, desired_shipment:, variant:, quantity:, track_inventory: TRACK_INVENTORY_NOT_PROVIDED)
       @current_shipment = current_shipment
       @desired_shipment = desired_shipment
       @current_stock_location = current_shipment.stock_location
       @desired_stock_location = desired_shipment.stock_location
       @variant = variant
       @quantity = quantity
+      @track_inventory = if track_inventory == TRACK_INVENTORY_NOT_PROVIDED
+        Spree::Deprecation.warn(
+          "Not passing `track_inventory` to `Spree::FulfilmentChanger` is deprecated." \
+        )
+        true
+      else
+        track_inventory
+      end
     end
 
     validates :quantity, numericality: { greater_than: 0 }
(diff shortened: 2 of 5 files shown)


octave:
👍

kennyadsl:
Checking if this is still a bug. Locally the provided specs are not failing to remove the change to the fulfillment changer class.  Just pushed the spec only to see if we are still effected. If everything is green, we can close this without merging.

kennyadsl:
I can reproduce the failing spec. Will push an update soon.

kennyadsl:
I pushed an updated version with more unit specs and the implementation of [this suggestion](https://github.com/solidusio/solidus/pull/3338#pullrequestreview-291083699).

rainerdema (approved):
Fantastic work, @kennyadsl @nspinazz89! Everything seems to be working perfectly. Thank you!

api/app/controllers/spree/api/shipments_controller.rb
```diff
@@ -127,7 +127,8 @@ def transfer_to_shipment
           current_shipment: @original_shipment,
           desired_shipment: @desired_shipment,
           variant: @variant,
-          quantity: @quantity
+          quantity: @quantity,
+          track_inventory: Spree::Config.track_inventory_levels
```
spaghetticode, line 131: I'm thinking this new param is not necessary since `Spree::Config.track_inventory_levels` could be accessed directly within `FulfilmentChanger`, but given we need to deprecate the previous behavior I can't think of any other way than this 🤔 . I'm wondering if we should try to remove it later, when we finally remove the default behavior.

core/app/models/spree/fulfilment_changer.rb
```diff
@@ -20,15 +20,23 @@ class FulfilmentChanger
     include ActiveModel::Validations
 
     attr_accessor :current_shipment, :desired_shipment
-    attr_reader :variant, :quantity, :current_stock_location, :desired_stock_location
+    attr_reader :variant, :quantity, :current_stock_location, :desired_stock_location, :track_inventory
 
-    def initialize(current_shipment:, desired_shipment:, variant:, quantity:)
+    def initialize(current_shipment:, desired_shipment:, variant:, quantity:, track_inventory: nil)
```
spaghetticode, line 25: I'm thinking `nil` might be somehow a valid value for `Spree::Config.track_inventory_levels`, as it still evaluates to `false`. I'm thinking the default value for `track_inventory` may be a constant with an arbitrary value and use that instead of `nil`, so we're 100% sure we didn't receive that param from the caller. Something like what's being done [in this PR](https://github.com/solidusio/solidus/pull/4909/files) with `PAYMENT_NOT_PROVIDED`. WDYT?

api/app/controllers/spree/api/shipments_controller.rb
kennyadsl: My idea was that at the end it's not that bad because this `track_inventory` value determines heavily the logic used in the class, and it's better to be as explicit as possible about it when it's initialized. After the deprecation period, we could evaluate removing it (with another deprecation), though.

core/app/models/spree/fulfilment_changer.rb
kennyadsl: I thought people always need to pass something that is `true` or `false` explicitly here. Do you mean that someone might pass `nil` initializing this class, but they actually mean to pass false? I'm not sure when that could happen, but open to changing this if you have a concrete example.

spaghetticode: @kennyadsl looks great, thanks!

spaghetticode (approved):
Thanks @kennyadsl , LGTM! 
I see there are some CI runs failing but I think they're not related to these changes.

PR #3342, opened 2019-09-20
[backend] Disable submit buttons after first click

spaghetticode:
**Description**
It may happen that admin users click multiple times (purposely or not) on submit buttons in the backend area. 

This can be a problem as at times it may silently generate unwanted duplicate records, or show an error page due to duplicate records. This is much more likely to happen on real shops when unanticipated server load makes the application unresponsive (at first click nothing seems to happen, so then why not click again?) and where some controller `create` actions may be patched and be slower.

By changing `button_tag` to `f.submit` or `submit_tag`  the button gets disabled after the first click, thanks to the automatically added attribute `data-disable-with="Create"`

**Checklist:**

### DIFF
--- a/backend/spec/features/admin/configuration/payment_methods_spec.rb
+++ b/backend/spec/features/admin/configuration/payment_methods_spec.rb
@@ -32,8 +32,8 @@ describe "Payment Methods", type: :feature do
     end
   end
 
-  context "admin creating a new payment method" do
-    it "should be able to create a new payment method" do
+  context "admin creating a new payment method", :js do
+    it "creates a new payment method and disables the form" do
       click_link "Payments"
       expect(page).to have_link 'Payment Methods'
       click_link "admin_new_payment_methods_link"
@@ -43,6 +43,15 @@ describe "Payment Methods", type: :feature do
       select Spree::PaymentMethod::Check.model_name.human, from: "Type"
       click_button "Create"
       expect(page).to have_content("successfully created!")
+
+      visit spree.new_admin_payment_method_path
+      page.execute_script "$('form').submit(function(e) { e.preventDefault()})"
+      fill_in "payment_method_name", with: "check90"
+      fill_in "payment_method_description", with: "check90 desc"
+      select Spree::PaymentMethod::Check.model_name.human, from: "Type"
+      click_button "Create"
+
+      expect(page).to have_button("Create", disabled: true)
     end
   end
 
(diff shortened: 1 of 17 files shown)


backend/spec/features/admin/configuration/payment_methods_spec.rb
```diff
@@ -44,6 +44,17 @@
...
+
+    it 'disables the button at submit', :js do
+      visit spree.new_admin_payment_method_path
+      page.execute_script "$('form').submit(function(e) { e.preventDefault()})"
+      fill_in "payment_method_name", with: "check90"
+      fill_in "payment_method_description", with: "check90 desc"
+      select Spree::PaymentMethod::Check.model_name.human, from: "Type"
+      click_button "Create"
+
+      expect(find('input[type="submit"]')).to be_disabled
```
tvdeyen: Could we move this into the other spec instead? We should avoid to have multiple `it` blocks in feature specs. This unnecessarily slows down our test suite

backend/spec/features/admin/configuration/shipping_methods_spec.rb
```diff
@@ -38,6 +38,21 @@
...
+      page.execute_script "$('form').submit(function(e) { e.preventDefault()})"
+      fill_in "shipping_method_name", with: "bullock cart"
+
+      within("#shipping_method_categories_field") do
+        check first("input[type='checkbox']")["name"]
+      end
+
+      click_on "Create"
+
+      expect(find('input[type="submit"]')).to be_disabled
```
tvdeyen: Here as well.

backend/spec/features/admin/promotions/promotion_code_batches_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      expect(page).to have_content "Promotion Code Batch has been successfully created!"
+    end
+
+    it "disables the button at submit", :js do
+      page.execute_script "$('form').submit(function(e) { e.preventDefault()})"
+
+      create_code_batch
+
+      expect(find('input[type="submit"]')).to be_disabled
```
tvdeyen: We don’t need multiple it blocks to test this feature.

seandawes:
This is also an issue in areas where double click occurs on refunds for example. It will double refund. We solved this on our store and you ( @spaghetticode ) actually fixed the issue we spec'd out for you . Dec 2018 Modded Euros issue 870 in our codebase for reference

backend/spec/features/admin/configuration/shipping_methods_spec.rb
spaghetticode: @tvdeyen I updated the files according to your suggestion. I think they don't read as good as before, so we're loosing something here. But I agree that saving some seconds is a good thing, so why not 👍

Issue #3345, opened 2019-09-22
Locale spelled wrong for Logged In Successfully


nolantait:
The locale for t('spree.logged_in_succesfully') is spelled wrong. Just had a time trying to figure out why it wasn't working. Successfully should have two s.

**Solidus Version**:
2.9.1

**To Reproduce**
1. Start an application and try to use the In8n translation for spree.logged_in_successfully in console.

askl56:
@jacobherrington Can we close this topoic?

PR #3346, opened 2019-09-23
fix spelling of locale logged_in_successfully

nspinazz89:
**Description**
Fix for #3345 

**Checklist:**

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1518,7 +1518,7 @@ en:
     lock: Lock
     log_entries: Log Entries
     logged_in_as: Logged in as
-    logged_in_succesfully: Logged in successfully
+    logged_in_successfully: Logged in successfully
     logged_out: You have been logged out.
     login: Login
     login_as_existing: Login as Existing Customer
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1518,6 +1518,7 @@ en:
     lock: Lock
     log_entries: Log Entries
     logged_in_as: Logged in as
+    logged_in_succesfully: Logged in successfully
     logged_in_successfully: Logged in successfully
     logged_out: You have been logged out.
     login: Login


jacobherrington:
So this is a bit more complicated than I'd like it to be.

I think we are okay to merge this, but we will need to make a similar change in [solidus_auth_devise](https://github.com/solidusio/solidus_auth_devise/search?q=logged_in_succesfully&unscoped_q=logged_in_succesfully).

It seems like this is something people have used for some time ([based on a GitHub search](https://github.com/search?q=logged_in_succesfully&type=Code)). Hopefully the changelog will be enough to help people find this...

jacobherrington:
Here are all of the places where we have used the misspelling: https://github.com/search?q=org%3Asolidusio+logged_in_succesfully&type=Code

kennyadsl:
Here's another extension to be fixed: https://github.com/solidusio/solidus_i18n/search?q=logged_in_succesfully&unscoped_q=logged_in_succesfully

kennyadsl:
Just commented on https://github.com/solidusio/solidus_auth_devise/pull/173, and now I have other concerns (sorry): if a user updates `solidus` without updating `solidus_auth_devise` they'll end up having no translation for the old wrong string which is still used into `solidus_auth_devise`. The worst thing is that in this scenario the i18n fallback will auto-generate a wrong `"Logged in succesfully"`string. 

That said, I'm tempted to ask you to leave both strings, but the [solution proposed here](https://github.com/solidusio/solidus_auth_devise/pull/173#issuecomment-534534274) won't have any effect at that point. 

Maybe the safest thing for users is to keep both things (double keys here (and `solidus_i18n`) and the fallback default in https://github.com/solidusio/solidus_auth_devise/pull/173). This way, users that update solidus will have an extra string in the i18n and in the future (maybe solidus 3.0), we will be able to safely remove the duplicated (wrong) string. What do you think?

spaghetticode:
I agree with @kennyadsl, in order to play nice with every scenario we need to keep also the old key, and only later with a major release remove the misspelled one.

kennyadsl:
@nspinazz89 can you please rebase this one?

spaghetticode:
@nspinazz89 In order to improve git history readability, can you squash the commits into a single one? 🙏

Issue #3339, opened 2019-09-19
Non-promotable items blacklist an entire order


nspinazz89:
**Solidus Version**:
Master

**To Reproduce**
1. Mark a product as non-promotable

2. Create a per line item adjustment promotion

3. Create an order with the non-promotable item and promotable items

4. Attempt to apply the coupon in the cart 

**Current behavior**
The coupon fails to apply because the order is not eligible for promotion. 

**Expected behavior**
This is the correct behavior for a whole order adjustment, but the incorrect behavior for a per line item adjustment. In a per line item adjustment one would expect, that the promotion is applied to each eligible item and the non-eligible item is skipped. 

**Additional context**
Interested in alternative solutions, otherwise I will tackle this bug.

jacobherrington:
This is a bug we found, hopefully we'll have a solution soon, but I'd love to hear of anything people have come up with to solve this elsewhere. 🤠

jacobherrington:
We are evaluating the approach in #1925. I think this is what I would do in a vacuum, so I want to re-evaluate it and maybe even re-open that PR.

cedum:
Check @loicginoux [comment](https://github.com/solidusio/solidus/issues/1437#issuecomment-447410940). I think this might be a better approach to solve this issue instead of the one proposed in the PR#1925.

kennyadsl:
@jacobherrington @nspinazz89 I'm working with @cedum to another approach to face this problem in #3348, let us know what do you think! This should solve the issue taking into account concerns with applying discounts to non-promotionable items.

fthobe:
@jarednorman I think with the fresh promotion system the time for this bug to be solved has come. Is this something on the roadmap before 4.5?

PR #3348, opened 2019-09-25
Allow order level promotions with non promotable items

kennyadsl:
🚧🚧🚧   

**Description**
This PR allows applying order-level promotions to orders with some non-promotionable line items. 
Those line items will not be taken into account in promotion computations since store owners could have their good reasons to exclude them from being promotionable.

ref: #1925, #1437 and #3339 

**WIP** This is just a spike, it still needs:

- validation, does it work or am I missing something?
- refactor: I think this code can be improved, I will do that if this works for everyone
- tests
- a configuration that allows not changing existing stores behavior, see #1925

Thanks, @cedum for the help in the in-depth analysis. 

**Checklist:**


core/app/models/spree/promotion.rb
```diff
@@ -240,8 +240,8 @@ def blacklisted?(promotable)
       case promotable
       when Spree::LineItem
         !promotable.product.promotionable?
-      when Spree::Order
-        promotable.line_items.joins(:product).where(spree_products: { promotionable: false }).exists?
+      # when Spree::Order
+      #   promotable.line_items.joins(:product).where(spree_products: { promotionable: false }).exists?
```
adaddeo: Another option could be to perform a check like so:
```
        promotable.line_items.any? &&
          !promotable.line_items.joins(:product).where(spree_products: { promotionable: true }).any?
```

kennyadsl: Thanks for your suggestion. Do you mean that we could perform this check instead of removing this code?

kennyadsl:
Closing this. It tries to fix a bug but I don't know if it's still the best way to handle the problem now. This spike will still be attached to the main issue, which will be kept open.

PR #3352, opened 2019-09-26
Handle deprecation for Rails 6 in DummyApp

cedum:
`.represent_boolean_as_integer=` is always true. Setting this is
deprecated and will be removed in Rails 6.1

Not sure if this is the right way to handle these types of deprecations.

Ref: https://github.com/rails/rails/commit/f59b08119bc0c01a00561d38279b124abc82561b

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -13,6 +13,7 @@ Rails.env = 'test'
 require 'solidus_core'
 
 RAILS_52_OR_ABOVE = Gem::Version.new(Rails.version) >= Gem::Version.new('5.2')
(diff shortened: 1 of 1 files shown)


core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -64,7 +65,11 @@ class Application < ::Rails::Application
 
     if RAILS_52_OR_ABOVE
       config.action_controller.default_protect_from_forgery = true
-      config.active_record.sqlite3.represent_boolean_as_integer = true
+      unless RAILS_6_OR_ABOVE
+        # `.represent_boolean_as_integer=` is always true.
+        # Setting this is deprecated and will be removed in Rails 6.1
```
kennyadsl: I think moving this comment into the commit message description is enough documentation.

PR #3356, opened 2019-09-27
Extract the state machines into replaceable modules

cedum:
This PR extracts the existing state machines into replaceable modules.

A while ago I took ownership of some PRs that completely removes the state machines from a few models (#3038, #3039, #3040, #3041).
While I fully agree with the rationale behind those PRs, I think that rewriting manually each dynamically generated `state_machines`'s method might not be ideal.

AFAIK, the state machines in these models don't need to be customized for the majority of the shops (please correct me). Also, having different state-related logic implementations for different models could be confusing.

By discussing this a bit with @kennyadsl, we thought that it might be worth introducing first an intermediary step that lets developers tinker more easily with the existing state machines instead of removing them completely.
This will enable the (few?) shops which require a custom state-related logic to provide their own while keeping a standard implementation for the rest. Hopefully, it could also lead to more ideas 💡addressing the concerns in the original PR (#3038).

**Checklist:**


BenMorganIO:
This is a really good idea :+1:

cedum:
> I'd love to also see some Guides pages update with some simple examples, but it's not blocking. Just let me know if you plan to do so and I'll add a WIP tag to the PR

@kennyadsl 👍I'll work on it in this PR

cedum:
I've pushed a draft for the documentation ([here's the preview](https://deploy-preview-3356--solidus-guides.netlify.com/developers/customizations/state-machines.html)).

Instead of describing it for each model, I preferred to add a generic guide under _Customizations_.
It would be awesome to have some feedback 🙏

kennyadsl:
I think it's a perfect starting point, thanks!

Issue #3065, opened 2019-01-30
RFC: Deprecate Spree namespace


dustineichler:
I'd like to see Solidus deprecate the `Spree` namespace. It's time we move away from being a "fork" of Spree. Obviously this would be a big lift, but necessary while comparisons between Solidus and Spree are no longer 1:1 and haven't been for a while.

Questions, comments and concerns below.

jarednorman:
Yeah, I'm personally pretty indifferent on this one (though it would be nice,) but if we do it, we definitely have to go with whatever approach causes the least problems.

dustineichler:
@jarednorman agreed. i've looked into this. i see two issues: updating namespace and db migration e.g. - `solidus_`.

dhonig:
Obvious issue with the plugins and echosystem as well.

patleb:
@jacobherrington, project namespace is an implementation detail, I don't see how this could affect the stakeholders matter (the project is already marketed as Solidus).

I don't mean to be harsh, but I really don't see any advantages to change namespacing just for the sake of it: I would say that it could potentially do more harm than good and it doesn't clarify anything (quite the contrary, you know from where the code comes from --> Spree).

As someone who works with both projects (Spree and Solidus), I can tell that there is more similar code than completely different one and both are using almost identical data modeling and ideas. I understand that the project evolved in parallel to Spree at some point, but it hasn't been a rewrite, it's an evolution and wouldn't make things easier to read, write or understand to change the namespace.

An example of fork of a known project which kept the namespace of the original project is [OpenProject](https://github.com/opf/openproject) which is a fork of [Redmine](https://github.com/redmine/redmine) and both are doing pretty well as is.

dustineichler:
@patleb "implementation details" and stakeholder issues aren't mutually exclusive. No one is disputing this is a cosmetic up. No one is disputing potential impact. No one is saying we're doing it for the sake of it. We are looking for better anecdotes in this discussion than, "could potentially do more harm than good" and "i don't see any advantages".

Can you speak to some of the work you're doing and how this change might affect the projects you're working on.

patleb:
> "implementation details" and stakeholder issues aren't mutually exclusive

In what way if I might ask? I'm talking about this specific issue, not about all those in existence. In this case, I'm pretty sure this is irrelevant: when was the last time a stakeholder spoke up about how you name your classes or how some files are named?

> No one is disputing this is a cosmetic up

[...]

PR #3364, opened 2019-10-04
Deprecate Spree namespace

aldesantis:
**Description**

Ref: #3065 

Renames the `Spree` namespace to `Solidus` and creates proxies to allow usage of the old namespace in legacy code, table names and view paths.

All of the proxies/monkey-patches are self-contained, so that we don't have to remove them from a million files when we complete the migration. We may even allow users to enable/disable them via the configuration, in case they want to try to migrate completely right away (e.g. `config.solidus_namespace_proxies = true`, or at a more granular level by enabling/disabling individual modules).

At this point, this is very much experimental, but the approach seems promising and CI is passing. Interested to know what people think about the architecture and whether we're interested in doing this at all.

Before merging this, we should also test it in real applications, which will be the ultimate test of its reliability.

**Checklist:**


BenMorganIO:
Let's get this in sooner before people notice the large change.

aldesantis:
We decided not to move forward with this eventually, rationale in https://github.com/solidusio/solidus/issues/3065#issuecomment-543135460.

PR #3365, opened 2019-10-06
A promotion without actions should log errors

### LINKED ISSUE #2777: A promotion without actions should log errors
### Steps to reproduce
1. create a sandbox app
2. start the server and navigate to localhost:3000/admin/promotions
3. create any promotion with a coupon code, but don't add any actions
4. checkout using the promotion code

### Expected behavior
The server should log a warning that the coupon does nothing. I feel this way because the promotion is "active." If the admin doesn't want the promotion to attempt to apply, it should be set to inactive. As an admin, I would expect that any "active" promotion will at least attempt to apply itself. 

### Actual behavior
The coupon gives an error: "The coupon code you entered doesn't exist. Please try again." 

That error message is totally fine for the frontend as someone shopping on a Solidus store shouldn't see that they tried to use an empty promotion, but the server should log something more useful to the admin. e.g., "Someone tried to use an empty coupon code..." 

This way the admin isn't confused as to why an active coupon isn't applied.  

screenshot from 2018-06-18 15-10-27

### System configuration
**Solidus Version**:
Using master

johnny243:
**Description**

The following PR fixes #2777 

When a customer tries to apply a promotion coupon code that has no actions, the following will occur: 
- They should get the `The coupon code you entered doesn't exist.` error message 
- A warning is logged mentioning said coupon

**Checklist:**

### DIFF
--- a/core/app/models/spree/promotion_handler/coupon.rb
+++ b/core/app/models/spree/promotion_handler/coupon.rb
@@ -17,6 +17,9 @@ module Spree
             handle_present_promotion(promotion)
           elsif promotion_code && promotion_code.promotion.inactive?
             set_error_code :coupon_code_expired
+          elsif promotion_code && !promotion.actions.exists?
+            set_error_code :coupon_code_not_found
+            Rails.logger.warn I18n.t(:coupon_code_no_promotion_actions, coupon_code: coupon_code)
           else
             set_error_code :coupon_code_not_found
           end
--- a/guides/source/users/promotions/promotion-actions.html.md
+++ b/guides/source/users/promotions/promotion-actions.html.md
@@ -25,8 +25,12 @@ following steps:
    of many types of discount depending on the promotion calculator being used.
    (For example, the amount could be a flat percentage or a flexible rate.)
 
-Note that if you choose the promotion action type **Free Shipping** that steps
-3. and 4. are not applicable.
+### Notes
+1. If you choose the promotion action type **Free Shipping**, then steps
+**3** and **4** are not applicable.
+2. You can set a coupon code to your promotion, but your customers will be 
+unable to use it when no actions are present. In these cases, a warning message
+will be shown in the logs.
 
 [create-a-new-promotion]: overview.html#create-a-new-promotion
 [promotion-calculators]: promotion-calculators.html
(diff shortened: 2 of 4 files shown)


core/app/models/spree/promotion_handler/coupon.rb
```diff
@@ -17,6 +17,11 @@ def apply
             handle_present_promotion(promotion)
           elsif promotion_code && promotion_code.promotion.inactive?
             set_error_code :coupon_code_expired
+          # If the promotion code exists but has no actions,
+          # then log warning and show not found error to user
```
kennyadsl: I think it's better avoid adding code comments unless strictly needed. What happens is already clear by reading the code.

guides/source/users/promotions/promotion-actions.html.md
```diff
@@ -25,8 +25,13 @@ following steps:
    of many types of discount depending on the promotion calculator being used.
    (For example, the amount could be a flat percentage or a flexible rate.)
 
-Note that if you choose the promotion action type **Free Shipping** that steps
-3. and 4. are not applicable.
+### Notes
+1. If you choose the promotion action type **Free Shipping**, then steps
+**3** and **4** are not applicable.
+2. A warning message is logged if the following conditions are true:
```
kennyadsl: I'd use a more admin-friendly language here. They are not supposed to be technical people and maybe a more conversational tone of voice is more appropriate here.

johnny243: How about:

`2. You can specify a coupon code to your promotion, but your customers
will not be able to apply it if no actions are added to them.`

johnny243:
> @johnny243 is this PR still WIP?
> 
> I am asking because the current implementation does not reflect what you wrote on the issue at [#2777 (comment)](https://github.com/solidusio/solidus/issues/2777#issuecomment-540091468)... curious to know if you're interested in going that way... thank you

@spaghetticode I would like to go that way, but I was waiting on some kind of approval/feedback/ideas to make sure I was implementing something that made sense.

I will get back on this issue today. Thanks for following up!

aldesantis:
This seems to be similar to #3596.

seand7565:
Adding more info to this - #2777 initially proposed logging an error (which this PR handles) but it seems the discussion has since been moved elsewhere (for the reason that admin users rarely have access to logs - which makes sense to me). 

#3596 handles the issue in line with the new discussion from #2777

kennyadsl:
Closed with https://github.com/solidusio/solidus/pull/3749

PR #3366, opened 2019-10-07
Ability to unset a taxon icon

### LINKED ISSUE #3349: Ability to unset a taxon icon
Copied the existing icon functionality to let clients upload a banner for taxons as well, and realized after an icon is uploaded, there's no way besides editing the database for a user to remove the icon. You could replace the icon, but sometimes this may not be wanted. For example the site I'm working on has a grid of the taxons with an icon.

Perhaps a little red remove icon + text below the existing icon.

Building a one off route that does this and a link to it. My Rails skills aren't there yet though.

johnny243:
***Description***

The following PR fixes #3349 

Ability to clear a `taxon` icon through the `taxon`'s edit page.

Feedback is very much appreciated.

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/taxons_controller.rb
+++ b/backend/app/controllers/spree/admin/taxons_controller.rb
@@ -72,6 +72,19 @@ module Spree
         respond_with(@taxon) { |format| format.json { render json: '' } }
       end
 
+      def clear_icon
+        @taxon = Spree::Taxon.find(params[:taxon_id])
+        @taxon.icon = nil
+
+        if @taxon.save
+          flash[:success] = flash_message_for(@taxon, :successfully_icon_cleared)
+        end
+
+        respond_with(@taxon) do |format|
+          format.html { redirect_to edit_admin_taxonomy_url(@taxon.taxonomy_id) }
+        end
+      end
+
       private
 
       def taxon_params
--- a/backend/config/routes.rb
+++ b/backend/config/routes.rb
@@ -139,6 +139,7 @@ Spree::Core::Engine.routes.draw do
       collection do
         get :search
       end
+      resource :icon, only: :destroy, to: "taxons#clear_icon"
     end
 
     resources :reimbursement_types, only: [:index]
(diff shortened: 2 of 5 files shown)


backend/config/routes.rb
```diff
@@ -141,6 +141,12 @@
       end
     end
 
+    resource :taxons do
+      member do
+        put "/:id/clear_icon", to: "taxons#clear_icon", as: :clear_icon
```
BenMorganIO: I would prefer to keep this RESTful and instead opt for:

```ruby
resource :taxons, only: [] do
  resource :icon, only: :delete
end
```

backend/config/routes.rb
```diff
@@ -141,6 +141,12 @@
       end
     end
 
+    resource :taxons do
```
BenMorganIO: Note: a `:taxons` resource is already defined.

backend/app/controllers/spree/admin/taxons_controller.rb
```diff
@@ -72,6 +72,19 @@ def destroy
         respond_with(@taxon) { |format| format.json { render json: '' } }
       end
 
+      def clear_icon
+        @taxon = Spree::Taxon.find(params[:taxon_id])
+        @taxon.icon = nil
```
spaghetticode: In order to save one line of code this could be written also like this:
 
```ruby
if @taxon.update(icon: nil)
  flash[:success] = flash_message_for(@taxon, :successfully_icon_cleared)
end

File: backend/config/routes.rb [hunk 4]
```diff
@@ -139,6 +139,7 @@
       collection do
         get :search
       end
+      resource :icon, only: :destroy, to: "taxons#clear_icon"
```
spaghetticode: We don't have a proper `Icon` resource, the resource involved here is a `Taxon`, and the controller method was added to TaxonsController, so I think it may be more appropriate to add the `clear_action` route to the `taxon` resource... what do you think?

seand7565:
This change has already been made with #3446

PR #3368, opened 2019-10-07
Guides (API): Some additional API token info

felixyz:
*UPDATE*: Ignore everything below. All that remains is a small extra tip on where to find your API token and a `curl` example.

**Description**

The [API guide](https://solidus.docs.stoplight.io/authentication) states that the API token should be passed in through the `Authentication` header. I discovered this doesn't work, and from [this line](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/base_controller.rb#L114) it looks like the correct header is `X-Spree-Token`, which also works in practice for me.

Either I misunderstood something, or the guides are wrong.

I also added a `curl` example and spelled out where people can find their admin token.

**Checklist:**

### DIFF
--- a/api/openapi/authentication.md
+++ b/api/openapi/authentication.md
@@ -10,6 +10,14 @@ You can use your API key to access all resources in the API. The API key must be
 Authorization: Bearer API_KEY
 ```
 
+As an admin, you can find your API token in the admin section under Users > Your e-email > API Access (at `admin/users/<user_id>/edit`)
+
+Example:
+
+```
+curl --header "Authorization: Bearer 1a6a9936ad150a2ee345c65331da7a3ccc2de" http://www.my-solidus-site.com/api/stores
+```
+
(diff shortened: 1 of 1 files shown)


kennyadsl:
@felixyz thanks for the contribution. Which Solidus version are you using? Recently, [we deprecated `X-Spree-Token`](https://github.com/solidusio/solidus/pull/3029), so I think guides are right. If the new header is not working for you, you probably are on old Solidus version, isn't it?

felixyz:
@kennyadsl :man_facepalming:  haha you caught me there, now you can infer just how far behind I am with my updates.

How about the other stuff? It's tiny but perhaps worth merging?

kennyadsl:
Sure, I like the rest of the PR. This also needs a rebase to fix pending CI checks I guess. Thanks Felix!

kennyadsl:
@felixyz do you have time to complete this PR or it's better to close it?

felixyz:
@kennyadsl Should be ready now.

aldesantis:
Looking good, thanks!

PR #3369, opened 2019-10-08
Allow to easily extend `Auth#store_location` behavior

spaghetticode:
**Description**

Often makes sense change the behavior of the method `Spree::Core::ControllerHelpers::Auth#store_location`, which is responsible for setting `session[:spree_user_return_to`]` value to the current URL. When necessity arises, the
only way ATM is to monkey patch Solidus version of this method.

This PR allows adding custom behavior by less obtrusive means, ie. by using custom rules.

**Checklist:**

### DIFF
--- a/core/lib/generators/spree/install/templates/config/initializers/spree.rb.tt
+++ b/core/lib/generators/spree/install/templates/config/initializers/spree.rb.tt
@@ -93,3 +93,11 @@ Spree.user_class = <%= (options[:user_class].blank? ? "Spree::LegacyUser" : opti
 # just uncomment the following code and change it as you need.
 #
 # Spree::Model.whitelisted_ransackable_attributes << 'field'
+
+# Rules for avoiding to store the current path into session for redirects
+# When at least one rule is matched, the request path will not be stored
+# in session.
+# You can add your custom rules by uncommenting this line and changing
+# the class name:
+#
+# Spree::UserLastUrlStorer.rules << 'Spree::UserLastUrlStorer::Rules::AuthenticationRule'
--- a/core/app/models/spree/user_last_url_storer.rb
+++ b/core/app/models/spree/user_last_url_storer.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+
+module Spree
+  # This service object is responsible for storing the current path into
+  # into `session[:spree_user_return_to]` for redirects after successful
+  # user/admin authentication.
+  class UserLastUrlStorer
+    # Lists all the rules that will be evaluated before storing the
+    # current path value into the session.
+    #
+    # @return [Spree::Core::ClassConstantizer::Set] a set of rules
+    #  that, when matched, will prevent session[:spree_user_return_to]
+    #  to be set
+    #
+    # @example This method can be used also to add more rules
+    #  Spree::UserLastUrlStorer.rules << 'CustomRule'
+    #
+    # @example it can be used also for removing unwanted rules
+    #  Spree::UserLastUrlStorer.rules.delete('CustomRule')
+    #
+    def self.rules
+      Spree::Config.user_last_url_storer_rules
+    end
+
+    # @param controller [ApplicationController] an instance of ApplicationController
+    #  or its subclasses. The controller will be passed to each rule for matching.
+    def initialize(controller)
+      @controller = controller
+    end
+
+    # Stores into session[:spree_user_return_to] the request full path for
+    # future redirects (to be used after successful authentication). When
+    # there is a rule match then the request full path is not stored.
+    def store_location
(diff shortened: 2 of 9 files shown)


core/app/models/spree/user_return_to_storer.rb
```diff
@@ -0,0 +1,54 @@
...
+    # current path value into the session.
+    #
+    # @return [Spree::Core::ClassConstantizer::Set] a set of rules
+    #  that, when matched, will prevent session[:spree_user_return_to]
+    #  to be set
+    #
+    # @example This method can be used also to add more rules
+    #  Spree::UserReturnToStorer.rules << 'CustomRule'
+    #
+    # @examle it can be used also removing unwanted rules
```
kennyadsl: `examle` -> `example`

core/app/models/spree/user_return_to_storer.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module Spree
+  # This service object is responsible for storing the current path into
+  # into `session[:spree_user_return_to]` for redirects after successful
+  # user/admin authentication.
+  class UserReturnToStorer
```
kennyadsl: This name is a bit hard to understand maybe, at least for me at the first look. I immediately thought about Returns and I'm wondering if we can use another terminology, which conflicts less with other elements of the Solidus domain, like `UserLastLocationStorer` maybe? I think it could be better even because this proposed name is not connected with how we'll use it, while `ReturnTo` is. Not a strong opinion here, what's yours?

core/lib/generators/spree/install/templates/config/initializers/spree.rb.tt
```diff
@@ -93,3 +93,11 @@ Spree.user_class = <%= (options[:user_class].blank? ? "Spree::LegacyUser" : opti
...
 #
 # Spree::Model.whitelisted_ransackable_attributes << 'field'
+
+# Rules for avoiding to store the current path into session for redirects
+# When at least one rule is matched, the request path will not be stored
+# in session.
+# You can add your custom rules by uncommenting this line and changing
+# the class name:
+#
+# Spree::UserReturnToStorer.rules << 'Spree::UrlReturnToStorer::Rules::AuthenticationRule'
```
kennyadsl: I'd use an explicit example class here, like `Spree::UrlReturnToStorer::Rules::YourCustomRule`

core/app/models/spree/user_return_to_storer.rb
```diff
@@ -0,0 +1,54 @@
...
+    class Configuration
+      def rules
+        @rules ||= ::Spree::Core::ClassConstantizer::Set.new.tap do |set|
+          set << 'Spree::UserReturnToStorer::Rules::AuthenticationRule'
```
kennyadsl: I think it could be better setting the default rule into `core/lib/spree/app_configuration.rb` [where we are creating the preference](https://github.com/solidusio/solidus/pull/3369/files#diff-3c6f02645e665491dd8ba3931be30b9fR480-R482) to better split the implementation and how we are actually using it in Solidus and also for consistency with the other preferences that seem to have defaults set there.

core/lib/generators/spree/install/templates/config/initializers/spree.rb.tt
spaghetticode: I used a real one so a developer interested in knowing how these rules work can quickly find that real existing module in the codebase and use it as a blueprint for their custom rules... or at least that's what I would do. 

Differently, one should look into the documentation or use their own means to find a real example, ending up on that very same file (it's the only rule we have so far) but probably spending more time and brain resources in the process.

core/app/models/spree/user_return_to_storer.rb
spaghetticode: My strong opinion here is that I don't like that name either 😄 

I think `UserLastLocationStorer` is good and better than mine, but it's a bit long. What about `UserLastUrlStorer` or `UserLocationStorer`?

As soon as we agree on a name I'll revisit the PR with the other suggestions as well... so much namespace and filenames to change 😄

kennyadsl: I think `UserLastUrlStorer` is fine. Let's see if we have some other opinion here :)

core/lib/generators/spree/install/templates/config/initializers/spree.rb.tt
kennyadsl: That makes sense!

core/app/models/spree/user_return_to_storer.rb
spaghetticode: Sounds good 👍

PR #3372, opened 2019-10-09
Link to guides after installation instructions

jarednorman:
We get a good amount of people completing the installation and setup of a Solidus store, but unsure about how to start customizing it showing up in Slack and asking questions which might be answered by these guides.

The logical next step after setting up Solidus store is to start customizing it, so I think it makes sense to send people to the customization guides here as a first step.

**Checklist:**


spaghetticode:
@jarednorman can you please rebase with master, so we get a green CI? Thank you 🙏 💚

PR #3373, opened 2019-10-10
Fix Sprockets 4 support for extensions

aldesantis:
**Description**

This fixes an issue with extension tests failing due to Sprockets 4 requiring an `app/assets/images` directory to be present.

**Checklist:**

### DIFF
--- a/core/lib/generators/spree/install/install_generator.rb
+++ b/core/lib/generators/spree/install/install_generator.rb
@@ -58,6 +58,9 @@ module Spree
 
     def setup_assets
       @lib_name = 'spree'
+
+      empty_directory 'app/assets/images'
+
       %w{javascripts stylesheets images}.each do |path|
         empty_directory "vendor/assets/#{path}/spree/frontend" if defined? Spree::Frontend || Rails.env.test?
         empty_directory "vendor/assets/#{path}/spree/backend" if defined? Spree::Backend || Rails.env.test?


kennyadsl:
We'll need this PR to fix all extensions specs, but if #3379 will be merged only. Otherwise, we can close this PR and backport the Sprockets 3 lock to all the supported Solidus versions.

aldesantis:
I'll close this in the meantime since we've locked Solidus to Sprockets 3. If https://github.com/rails/sprockets-rails/pull/446 is merged, Solidus may become compatible with Sprockets 4, in which case I'll reopen it.

aldesantis:
Reopening this, since the fix in https://github.com/rails/sprockets-rails/pull/446 only solves the issue for the Solidus dummy app, while this one solves a different issue with extensions.

core/lib/generators/spree/install/install_generator.rb
```diff
@@ -58,6 +58,9 @@ def additional_tweaks
 
     def setup_assets
       @lib_name = 'spree'
+
+      empty_directory 'app/assets/images'
```
jacobherrington: Do you think we could add a comment here linking to this PR or an issue that explains why this is here?

aldesantis: @jacobherrington I think the commit message already explains the rationale behind the change pretty well. I'm not against adding a comment, just not sure it's necessary and afraid it would unnecessarily pollute the code here. What do you think?

jacobherrington: I'm comfortable with that -- I think it's a good commit message. :+1:

PR #3375, opened 2019-10-10
Pass stock location to inventory unit factory

pelargir:
**Description**
When building fulfilled stock packages in specs using the stock package factory, this passes the given transient stock location to the inventory units that get built. This prevents the inventory unit factory from building unique stock locations for each inventory unit that gets built, and ensures that the stock package is using the same stock location as the inventory units it contains.

**Checklist:**


pelargir:
Rebased from master.

pelargir:
The build is now passing. What else needs to happen before this PR can be merged?

kennyadsl:
Thank you!

PR #3377, opened 2019-10-10
Add initial value to reduce function for tab widths

fastjames:
**Description**
When calculating total tab width, we recently saw a case where there
were no tabs and thus no IV for reduce. This adds a default starting
value of 0 which prevents a js exception in that case. I checked the rest of the js and found no other occurrences of a `reduce` call without an initial value.

**Checklist:**


fastjames:
Edited desc to be clear: I have not added tests to this for two reasons:

1. The addition of a default initial value of 0 for a summation function seems pretty obvious
2. I didn't see much if any existing test code for javascript in this area, so I would have to start from scratch

spaghetticode:
@fastjames can you please rebase with master, so we get a green CI? thank you 🙏

fastjames:
Rebased and force-pushed.

Issue #3374, opened 2019-10-10
Sprockets 4 compatibility


kennyadsl:
Sprockets 4 requires a configuration manifest file to be created in the application. This is fine for new stores but it's not working with our DummyApp at the moment since it is an in-memory application, and not a real application on the filesystem, like most engines do in `spec/dummy` with `rake test_app`.

At the moment, [Sprockets 4 is not able to allow setting a custom location](https://github.com/rails/sprockets-rails/issues/369) for this configuration manifest file and it is looking for our manifest in `spec/dummy/app/assets/config/manifest.js`, which does not exist. Spec fails consequently.

I'm trying to propose a change to sprockets-rails that allows setting a custom location for the manifest with [this PR](https://github.com/rails/sprockets-rails/pull/446), even because there's a Rails config named `config.assets.manifest`, which I think is only used for this reason, but I'm not super sure since I can't find a lot of documentation around.

The only alternative solution I can think about is locking Sprockets to 3.x on all the currently supported Solidus versions.

elia:
Fixed by #3379

PR #3378, opened 2019-10-11
Lock Sprockets to v3.x

spaghetticode:
**Description**

Ref https://github.com/solidusio/solidus/issues/3374

This is a temporary fix that will allow master to be green again (so we can continue working on PRs) while we wait for https://github.com/rails/sprockets-rails/pull/446 to be merged (or some other working solution).
    
**Checklist:**

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,11 @@ group :backend, :frontend, :core, :api do
   rails_version = ENV['RAILS_VERSION'] || '~> 6.0.0'
   gem 'rails', rails_version, require: false
 
+  # Temporarily locking sprockets to v3.x
+  # see https://github.com/solidusio/solidus/issues/3374
+  # and https://github.com/rails/sprockets-rails/issues/369
+  gem 'sprockets', '~> 3'
+
   platforms :ruby do
     case ENV['DB']
     when /mysql/


Gemfile
```diff
@@ -8,6 +8,13 @@ group :backend, :frontend, :core, :api do
   rails_version = ENV['RAILS_VERSION'] || '~> 6.0.0'
   gem 'rails', rails_version, require: false
 
+  if ENV['CIRCLE_TEST_REPORTS']
```
cedum: Just curious about `CIRCLE_TEST_REPORTS` var: why not using one of the built-in env flags: https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables ? 🤔(`CI` or `CIRCLECI`, both booleans).

spaghetticode: I didn't know exactly what env variables were set in circleCi by default, so I went with one that's explicitly set in the file [circle.yml](https://github.com/solidusio/solidus/blob/master/.circleci/config.yml)

After looking at the list I think the best candidate is `CI`, wdyt?

cedum: Agree 👍 I usually see `ENV['CI']` used in these cases

kennyadsl:
@spaghetticode I'd lock it anyway (not on CI only), we have this failure even running tests locally and on extensions due to this issue.

spaghetticode:
@kennyadsl locally one can use `CI=true bundle exec rake`, while (most of) extensions use circleCI already, so I guess the problem may be minimal. I was trying to avoid possible issues to people already on Sprockets 4, but maybe scenario actually does not exist 🤔 
Also, consistency is a valuable thing, so I'm chaging this to lock the gem everywhere. If somebody shows some concerns then I can change it again before merging.
.

kennyadsl:
@spaghetticode can you please rebase against upstream master to have Rails 5.1 specs out of the required checks?

spaghetticode:
@kennyadsl it's already updated. I removed the requirements for Rails 5.1 branches by updating github configuration. PR is now ready to merge.

Schermata 2019-10-11 alle 14 37 48

Issue #3363, opened 2019-10-03
Add tooltips on calculators in the admin panel


kennyadsl:
Currently what calculators do is not clear in our admin panel. I think we could easily add a tooltip when calculators are displayed to tell users how to use them. It's quite easy to get some text for them since we have calculators well described in our Guides:

- https://guides.solidus.io/developers/calculators/promotion-calculators.html#promotion-calculators
- https://guides.solidus.io/developers/calculators/shipping-calculators.html#shipping-calculators
- https://guides.solidus.io/developers/calculators/tax-calculator.html#tax-calculator

octoxan:
I'll jump on this tonight or tomorrow night.

seand7565:
Looks like this one has been solved by #3382 - can we close this?

PR #3382, opened 2019-10-15
Add tooltips to admin calculators

codykaup:
**Description**

This change adds tooltips for some calculators in the admin panel. I tried to create some basic text based on the guides so they can easily be changed in the future. If you have further ideas on what these should say, I'm all ears!!

_Side Note_: Since this is just changing a small piece of the frontend view, I wasn't totally sure on tests as there's no real extra functionality. As always, if you feel there should be some, don't hesitate to let me know!

This PR is in reference to: issue #3363

Thanks in advance!

**Quick view of how these look**

Screen Shot 2019-10-14 at 9 05 04 PM
Screen Shot 2019-10-14 at 9 04 01 PM
Screen Shot 2019-10-14 at 9 03 40 PM

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/shared/_calculator_fields.html.erb
+++ b/backend/app/views/spree/admin/shared/_calculator_fields.html.erb
@@ -4,6 +4,11 @@
   <div id="preference-settings">
     <div class="field">
       <%= f.label(:calculator_type, Spree::Calculator.model_name.human) %>
+
+      <% if local_assigns.has_key? :hint %>
+        <%= admin_hint t("spree.#{hint}"), t(hint, scope: [:spree, :hints, "spree/calculator"]) %>
+      <% end %>
+
       <%= f.select(:calculator_type, @calculators.map { |c| [c.description, c.name] }, {}, {class: 'custom-select fullwidth js-calculator-type'}) %>
     </div>
 
--- a/backend/app/views/spree/admin/promotions/_actions.html.erb
+++ b/backend/app/views/spree/admin/promotions/_actions.html.erb
@@ -7,6 +7,7 @@
       <% if can?(:update, @promotion) %>
         <div class="field">
           <%= label_tag :action_type, t('spree.adjustment_type')%>
+          <%= admin_hint t('spree.adjustment_type'), t(:promotions, scope: [:spree, :hints, "spree/calculator"]) %>
           <%= select_tag 'action_type', options, include_blank: t(:choose_promotion_action, scope: 'spree'), class: 'custom-select fullwidth' %>
         </div>
         <div class="filter-actions actions">
--- a/backend/app/views/spree/admin/shipping_methods/_form.html.erb
+++ b/backend/app/views/spree/admin/shipping_methods/_form.html.erb
@@ -100,7 +100,7 @@
   </div>
   <div class="col-5">
     <div data-hook="admin_shipping_method_form_calculator_fields">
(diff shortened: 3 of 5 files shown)


backend/app/views/spree/admin/shared/_calculator_fields.html.erb
```diff
@@ -4,6 +4,7 @@
   <div id="preference-settings">
     <div class="field">
       <%= f.label(:calculator_type, Spree::Calculator.model_name.human) %>
+      <%= admin_hint t("spree.#{@calculator_type}"), t(@calculator_type, scope: [:spree, :hints, "spree/calculator"]) %>
```
codykaup: I couldn't find a great way to grab the parent module of the calculator without it getting hacky so I figured an instance method in the controller would be a decent alternative!

kennyadsl:
Hey @codykaup, thanks a lot for the PR. I think this issue was more to explain what every single calculator does, rather than the generic calculator type. Anyway, I think this is still an improvement, and I'd like to go ahead.

What about passing the tooltip content as a partial variable from the outside, for example [here](https://github.com/solidusio/solidus/blob/b5016ebb2d873123866e1be0aa6446abc2146d43/backend/app/views/spree/admin/shipping_methods/_form.html.erb#L103), so we don't have to set the instance variable in the controller? This variable could be optional so, when we need to call the partial without the tooltip, we can still do that. What do you think?

codykaup:
> Hey @codykaup, thanks a lot for the PR. I think this issue was more to explain what every single calculator does, rather than the generic calculator type. Anyway, I think this is still an improvement, and I'd like to go ahead.
> 
> What about passing the tooltip content as a partial variable from the outside, for example [here](https://github.com/solidusio/solidus/blob/b5016ebb2d873123866e1be0aa6446abc2146d43/backend/app/views/spree/admin/shipping_methods/_form.html.erb#L103), so we don't have to set the instance variable in the controller? This variable could be optional so, when we need to call the partial without the tooltip, we can still do that. What do you think?

Definitely not a bad idea. Do you have a suggestion that would prevent hardcoding the calculator names?

Right now, I'm assuming it would be a new partial that defaults to `base_controller` unless supplied like promotions:
```
<%= admin_hint t('spree.adjustment_type'), t(:promotions, scope: [:spree, :hints, "spree/calculator"]), locals: { calculator: "promotions" } %>
```

kennyadsl:
I mean more something like adding the hint we want to show where we call the partial:

```erb
<%= render partial: 'spree/admin/shared/calculator_fields', locals: { f: f, hint: : shipping_methods } %>
```

and into that partial:

```erb
<% if local_assigns.has_key? :hint %>
  <%= admin_hint t("spree.#{hint}"), t(hint, scope: [:spree, :hints, "spree/calculator"]) %>
<% end %>
```

WDYT?

codykaup:
> I mean more something like adding the hint we want to show where we call the partial:
> 
> ```
> <%= render partial: 'spree/admin/shared/calculator_fields', locals: { f: f, hint: : shipping_methods } %>
> ```
> 
> and into that partial:
> 
> ```
> <% if local_assigns.has_key? :hint %>
>   <%= admin_hint t("spree.#{hint}"), t(hint, scope: [:spree, :hints, "spree/calculator"]) %>
> <% end %>
> ```
> 
> WDYT?

I like that idea a lot. It's essentially the same concept without including an instance variable. I just updated my commit to the new version and it works the exact same!

codykaup:
Hey @kennyadsl, not sure if you saw or not but the PR has been updated! 🎉

PR #3388, opened 2019-10-22
Fix deprecation message for Spree::CreditCard

spaghetticode:
**Description**

Deprecation message on `CreditCard#default` is a bit misleading, as it
prints out `CreditCard.default` but it's referencing an instance method
not a class method.

**Checklist:**

### DIFF
--- a/core/app/models/spree/credit_card.rb
+++ b/core/app/models/spree/credit_card.rb
@@ -47,13 +47,13 @@ module Spree
     }.freeze
 
     def default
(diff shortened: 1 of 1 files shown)


spaghetticode:
@jacobherrington I think this is probably an isolated case, caused by cut & paste from the deprecation of the class method just a few lines above:

```ruby
    def self.default
      Spree::Deprecation.warn("CreditCard.default is deprecated. Please use Spree::Wallet instead.")
      joins(:wallet_payment_sources).where(spree_wallet_payment_sources: { default: true })
    end
```

PR #3389, opened 2019-10-23
Reload product before assigning images to variants

### LINKED ISSUE #2978: Rake task for seeding sandbox app aborted
### Steps to reproduce

* Run `DB=postgresql bundle exec rake sandbox` on your Solidus source folder

### Expected behavior

* Sandbox app should be created without errors

### Actual behavior

* The following error raises:

`NoMethodError: undefined method 'downcase' for nil:NilClass`

And the sandbox's front-end looks like this:

image

### System configuration

* Latest master revision (318ff36)

**Extensions in use**:

* `solidus_auth_devise`

JDutil:
**Description**

While this doesn't actually fix the underlying problem as I wasn't able to determine what it was these changes appear to fix the flakiness of the sample loading (in my dev environment anyways).

This change adds a reload to products before assigning their variant images on the sample seeds loading. 

It also implements a couple minor rubocop and performance improvements noticed while troubleshooting the issue.

* Adds include statements to improve asset sample file performance, which also fixed option value lookup for the `solidus_tshirt` & `solidus_long` products. 
* Unfortunately the include statements just moved the bug to the `solidus_girly` product, but adding a `reload` before the variant lookup which fixed all three products works as a hack to fix the issue.
* Fixes Rubocop guard clause warnings
* Removes unused option value loading

Fixes #2978 

**Checklist:**

### DIFF
--- a/sample/db/samples/assets.rb
+++ b/sample/db/samples/assets.rb
@@ -128,7 +128,7 @@ products[:solidus_long].variants.each do |variant|
   end
 end
 
-products[:solidus_girly].variants.each do |variant|
+products[:solidus_girly].reload.variants.each do |variant|
   color = variant.option_value("tshirt-color").downcase
   main_image = image("solidus_girly_#{color}", "png")
   File.open(main_image) do |f|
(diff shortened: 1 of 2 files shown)


kennyadsl:
@JDutil I think failures will be fixed rebasing against master, thanks!

PR #3391, opened 2019-10-23
Display error if editing non-current order

### LINKED ISSUE #2078: Edit path for completed orders
If I visit the path of a completed order, e.g., `http://localhost:3000/orders/R465318167`, I can see a receipt of the order. If I then append `edit` to that, and navigate to `http://localhost:3000/orders/R465318167/edit`, I am shown the Cart page.

This is unexpected behavior, IMO. I’d expect to either be able to edit the order (i.e., change address or cancel or something like that), _or_ to be told that I can’t edit this order because it’s locked or finalized, etc.

But seeing the current_order Cart page rendered at the `/edit` URL of a completely different order is definitely not expected!

JDutil:
Users can input any order id within the orders controller routing, but
we simply render the cart with the current order. This can cause some
users confusion if they are expecting to edit an existing order after
viewing it's receipt.

So let's simply display a flash message indicating you cannot edit
existing orders if the routing id does not match the current order.

fixes #2078 

**Checklist:**

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1086,6 +1086,7 @@ en:
     cannot_create_payment_without_payment_methods_html: You cannot create a payment
       for an order without any payment methods defined. %{link}
     cannot_create_returns: Cannot create returns as this order has no shipped units.
+    cannot_edit_orders: You may only edit your current shopping cart.
     cannot_perform_operation: Cannot perform requested operation
     cannot_rebuild_shipments_order_completed: Cannot rebuild shipments for a completed
       order.
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -42,6 +42,10 @@ module Spree
       @order = current_order || Spree::Order.incomplete.find_or_initialize_by(guest_token: cookies.signed[:guest_token])
       authorize! :read, @order, cookies.signed[:guest_token]
       associate_user
+      if params[:id] && @order.number != params[:id]
+        flash[:error] = t('spree.cannot_edit_orders')
+        redirect_to cart_path
+      end
     end
 
     # Adds a new item to the order (creating a new order if none already exists)
--- a/frontend/spec/controllers/spree/orders_controller_spec.rb
+++ b/frontend/spec/controllers/spree/orders_controller_spec.rb
@@ -102,6 +102,31 @@ describe Spree::OrdersController, type: :controller do
       end
     end
 
+    context '#edit' do
+      before do
+        allow(controller).to receive :authorize!
+        allow(controller).to receive_messages current_order: order
+      end
+
+      it 'should render cart' do
+        get :edit, params: { id: order.number }
+
+        expect(flash[:error]).to be_nil
+        expect(response).to be_ok
+      end
+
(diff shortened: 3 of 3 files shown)


core/config/locales/en.yml
```diff
@@ -1086,6 +1086,7 @@ en:
     cannot_create_payment_without_payment_methods_html: You cannot create a payment
       for an order without any payment methods defined. %{link}
     cannot_create_returns: Cannot create returns as this order has no shipped units.
+    cannot_edit_completed_orders: Cannot edit completed orders.
```
JDutil: Not sure what wording we'd like to display here, but after submitting this I'm thinking it should be more broad than "completed" orders. Perhaps something more like `Cannot edit existing orders.` or `You may only edit your current shopping cart.`

dportalesr: `Cannot edit placed orders`, maybe?

JDutil: @dportalesr I changed to `You may only edit your current shopping cart.` which I think is most appropriate. It doesn't matter what id you enter you'll see your current cart. We also don't want to indicate to people what existing order numbers are in our system or what state they happen to be in.

frontend/app/controllers/spree/orders_controller.rb
```diff
@@ -42,6 +42,10 @@ def edit
       @order = current_order || Spree::Order.incomplete.find_or_initialize_by(guest_token: cookies.signed[:guest_token])
       authorize! :read, @order, cookies.signed[:guest_token]
       associate_user
+      # It's possible another error may already be assigned so we must join them.
```
kennyadsl: I think we should also redirect the user when this happens if possible. Allowing to change the current order presenting an old order's URL could be confusing for users.

If possible, what about doing that right after the `@order` assignment on line 42? This way, we can maybe even ignore merging the error message with others?

What do you think?

JDutil: @kennyadsl I think the redirect definitely makes sense so I'll do that, but I don't think I can drop the error message merging without losing one of the messages along the way although it's worth a shot.

JDutil: @kennyadsl I think the redirect definitely makes sense so I'll do that, but I don't think I can drop the error message merging without losing one of the messages along the way although it's worth a shot.

kennyadsl: Do you know which kind of other messages could we have there?

JDutil: When a product becomes unavailable, and is removed from your cart.

kennyadsl: When the product becomes unavailable we are redirected to the cart with another flash message but we should not see the new `cannot_edit_completed_orders` flash, right?

I think that this is happening because we are not checking the existence of the `params[:id]` in the if statement. We should print the error message only when `@order.number != params[:id]` and `params[:id]` is set otherwise, as it is now, it will be always be displayed in the cart page.

I think that we can get rid of the messages join if we do this and redirect the user to `spree.order_path(params[:id])`.

JDutil: @kennyadsl updated 👍

PR #3392, opened 2019-10-23
Taking advantage of rails insert all to create inventory units

softr8:
**Description**
By default, rails saves automatically all the new objects recursively, in the case of
solidus, whenever the shipments are assigned to the order,  all dependent objects are
saved automatically,  but speaking about the inventory units, they get inserted one by one, 
we can take advantage of new Rails 6 insert_all functionality here to speed up the creation

**Checklist:**

### DIFF
--- a/core/app/models/spree/inventory_unit.rb
+++ b/core/app/models/spree/inventory_unit.rb
@@ -118,6 +118,10 @@ module Spree
       on_hand?
     end
 
+    def value_attributes
+      attributes.slice('pending', 'state', 'variant_id', 'line_item_id', 'shipment_id')
+    end
+
     private
 
     def fulfill_order
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -535,7 +535,33 @@ module Spree
         raise CannotRebuildShipments.new(I18n.t('spree.cannot_rebuild_shipments_shipments_not_pending'))
       else
         shipments.destroy_all
-        self.shipments = Spree::Config.stock.coordinator_class.new(self).shipments
+
+        new_shipments = Spree::Config.stock.coordinator_class.new(self).shipments
+        # By default, rails saves automatically all the new objects recursively
+        # whenever the shipments are assigned to the order, but in terms of inventory
+        # units, it inserts them one by one, we can take advantage of new Rails 6
+        # insert_all functionality here
+        if Spree::InventoryUnit.respond_to?(:insert_all)
+          self.shipments = new_shipments.map do |shipment|
+            # We need to save temporarily the inventory unit attributes and
+            # unset  them from shipment so they do not get saved
+            inventory_units = shipment.inventory_units.map(&:dup)
+            shipment.inventory_units = []
+            shipment.save
+
+            if inventory_units.any?
+              if inventory_units.detect { |iu| iu.shipment = shipment; !iu.valid? }
+                shipment.inventory_units = inventory_units # delegate to rails error reporting
+              else
+                Spree::InventoryUnit.insert_all(inventory_units.map(&:value_attributes))
+                shipment.reload # shipment needs to know about new objects
+              end
+            end
+            shipment
+          end
+        else
+          self.shipments = new_shipments
+        end
       end
     end
 
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -1157,6 +1157,20 @@ RSpec.describe Spree::Order, type: :model do
 
       expect { shipment.reload }.not_to raise_error
     end
+
+    it "creates the right inventory units" do
+      order = create(:order_with_line_items)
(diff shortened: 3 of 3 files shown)


core/app/models/spree/order.rb
```diff
@@ -535,7 +535,26 @@ def create_proposed_shipments
...
+        # insert_all functionality here
+        if Spree::InventoryUnit.respond_to?(:insert_all)
+          self.shipments = new_shipments.map do |shipment|
+            # We need to save temporarily the inventory unit attributes and
+            # unset  them from shipment so they do not get saved
+            inventory_units_attributes = shipment.inventory_units.map(&:attributes)
+            shipment.inventory_units = []
+            shipment.save
+            # Then insert them all at once
+            Spree::InventoryUnit.insert_all(inventory_units_attributes.map {|iu| iu.merge(shipment_id: shipment.id)})
```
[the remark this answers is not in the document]
dportalesr: Shouldn't `inventory_units_attributes ` already have the `shipment_id` set for the shipment?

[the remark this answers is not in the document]
softr8: it is not saved yet, they are just objects built in memory

core/app/models/spree/order.rb
```diff
@@ -535,7 +535,30 @@ def create_proposed_shipments
...
+          self.shipments = new_shipments.map do |shipment|
+            # We need to save temporarily the inventory unit attributes and
+            # unset  them from shipment so they do not get saved
+            inventory_units_attributes = shipment.inventory_units.map(&:value_attributes)
+            shipment.inventory_units = []
+            shipment.save
+            # Then insert them all at once
+            if inventory_units_attributes.size.positive?
+              Spree::InventoryUnit.insert_all(
+                  inventory_units_attributes.map { |iu| iu.merge(shipment_id: shipment.id) }
```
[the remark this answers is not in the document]
cone: I would move this code to its own method, and add some testing :)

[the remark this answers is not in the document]
cone: nvm, testing is not checked in the list yet, so I guess this is a WIP

core/app/models/spree/inventory_unit.rb
```diff
@@ -118,6 +118,10 @@ def allow_ship?
       on_hand?
     end
 
+    def value_attributes
+      attributes.slice('pending', 'state', 'variant_id', 'line_item_id')
```
jarednorman: Probably could use the symbols here.

core/app/models/spree/order.rb
```diff
@@ -535,7 +535,30 @@ def create_proposed_shipments
...
+        # insert_all functionality here
+        if Spree::InventoryUnit.respond_to?(:insert_all)
+          self.shipments = new_shipments.map do |shipment|
+            # We need to save temporarily the inventory unit attributes and
+            # unset  them from shipment so they do not get saved
+            inventory_units_attributes = shipment.inventory_units.map(&:value_attributes)
+            shipment.inventory_units = []
+            shipment.save
+            # Then insert them all at once
+            if inventory_units_attributes.size.positive?
```
jarednorman:
```suggestion
            if inventory_units_attributes.any?
```

core/app/models/spree/order.rb
```diff
@@ -535,7 +535,30 @@ def create_proposed_shipments
...
+              Spree::InventoryUnit.insert_all(
```
jarednorman: I haven't played with `insert_all` yet. Are you able to do the `insert_all` direction on the association to avoid the map?

```suggestion
              shipment.inventory_units.insert_all(inventory_units_attributes)
```

kennyadsl: I've literally just tried that locally and it doesn't seem to work, unfortunately.

softr8:
I've sent another proposal that improves a bit the approach and it actually runs validations, let me know what you think @kennyadsl !

core/app/models/spree/inventory_unit.rb
softr8: attributes is not a hash with indifferent access :(

kennyadsl:
Thanks Edwin, we discussed this into a past core team meeting and we think this change is great but it's better not to go in this direction. 

The reason is that, with the current version, we still instantiate all the records in memory to perform validation and we think this will reduce almost all the performance improvement introduced by `insert_all`. 

At the same time, we can't avoid running validations, since we don't have database constraints and we can't add them for the same reason [this PR](https://github.com/solidusio/solidus/pull/2036#issuecomment-390004302) has been closed.

I think that stores that need to improve performances in this specific part of Solidus should add database constraints to their database and override this method with your initial implementation, but for now, this is more some content for a blog post probably. 

If you agree I'm going to close this PR, thanks again for highlighting this great solution.

aldesantis:
Closing this since we decided not go this way. Thanks for the spike @softr8! 🙏

PR #3394, opened 2019-10-23
Add permission check for admins updating user passwords

### LINKED ISSUE #438: Remove admin ability to set a password, replace reset password email trigger.

JDutil:
For security purposes administrators should not be able to set a
users password. Only the accounts owner should be able to directly set
their password. administrators should only have the ability to send a
password reset email to the account owner. Otherwise someone working in
customer service or another role could take over a users account in
order to make illegal purchases with their stored credit card
information.

In order to maintain backwards compatibility, and leave more power in
control of the store owner this will leave the current admin role
behavior the same, but anyone creating custom roles will no longer
be able to update passwords unless they explicitly add a change
password permission.

Fixes #438 

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/users_controller.rb
+++ b/backend/app/controllers/spree/admin/users_controller.rb
@@ -131,6 +131,10 @@ module Spree
           attributes += [{ spree_role_ids: [] }]
         end
 
+        unless can? :update_password, @user
+          attributes -= [:password, :password_confirmation]
+        end
+
         params.require(:user).permit(attributes)
       end
 
--- a/backend/app/views/spree/admin/users/_form.html.erb
+++ b/backend/app/views/spree/admin/users/_form.html.erb
@@ -60,18 +60,20 @@
   </div>
 
   <div data-hook="admin_user_form_password_fields" class="col-6">
-    <%= f.field_container :password do %>
-      <%= f.label :password %>
-      <%= f.password_field :password, class: 'fullwidth' %>
-      <%= f.error_message_on :password %>
-    <% end %>
-
-
-    <% if f.object.respond_to?(:password_confirmation) %>
+    <% if can? :update_password, @user %>
       <%= f.field_container :password do %>
-        <%= f.label :password_confirmation %>
-        <%= f.password_field :password_confirmation, class: 'fullwidth' %>
-        <%= f.error_message_on :password_confirmation %>
+        <%= f.label :password %>
+        <%= f.password_field :password, class: 'fullwidth' %>
+        <%= f.error_message_on :password %>
+      <% end %>
+
+
+      <% if f.object.respond_to?(:password_confirmation) %>
+        <%= f.field_container :password do %>
+          <%= f.label :password_confirmation %>
+          <%= f.password_field :password_confirmation, class: 'fullwidth' %>
+          <%= f.error_message_on :password_confirmation %>
+        <% end %>
       <% end %>
     <% end %>
   </div>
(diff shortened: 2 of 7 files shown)


backend/app/views/spree/admin/users/_form.html.erb
```diff
@@ -58,21 +58,4 @@
     <% end %>
 
   </div>
-
-  <div data-hook="admin_user_form_password_fields" class="col-6">
```
JDutil: I'm not sure if we should delete this div with the form fields or not. There maybe stores using this data-hook so I'm thinking that perhaps we should leave it in for backwards compatibility. Although I'm also annoyed the data-hook would have a bad name.

JDutil:
@tvdeyen a button to send a password reset email was added to https://github.com/solidusio/solidus_auth_devise/pull/146 I can modify this to work based on permissions though.

JDutil:
@tvdeyen updated to be permission based 👍

JDutil:
@tvdeyen Yeah I also think it'd be great to start shipping Solidus with some sort of customer service role. Giving every "admin" super user permissions is a bit much by default. That would also help for testing many permission related features.

tvdeyen:
@JDutil agreed, I made #3404

backend/spec/features/admin/users_spec.rb
```diff
@@ -15,7 +15,9 @@
   let(:order_2) do
     create(:completed_order_with_totals, user: user_a, number: "R456").tap do |o|
       li = o.line_items.last
+      # rubocop:disable Rails/SkipsModelValidations
```
kennyadsl: we can remove these Rubocop rules now, right?

backend/app/controllers/spree/admin/users_controller.rb
```diff
@@ -131,6 +131,11 @@ def user_params
           attributes += [{ spree_role_ids: [] }]
         end
 
+        # Don't allow users to set another users password.
+        unless can? :update_password, @user
```
kennyadsl: I'd remove the code comment, it's pretty clear what it does. Also, not sure about this but what about adding this rule in the permission sets, set to true just for documentation? At least in the `Spree::PermissionSets::UserManagement` file?

backend/app/views/spree/admin/users/_form.html.erb
kennyadsl: If I got it correctly, I think this `div` should go into the `if` statement. I think it will still be evaluated even if the `if` statement is `false`, since Deface should do its magic before the erb is rendered.

JDutil: My thought here was that someone may have a deface customization that's adding non-password related fields here simply because it would be a convenient place to add more user fields. I'm fine putting the div into the conditional statement as well, but I think that may cause customizations to be hidden unintentionally if people are using this data hook for non-password stuff.

backend/app/controllers/spree/admin/users_controller.rb
JDutil: I thought it would be better to only allow admins to update the password, and have people opt into adding `:update_password` permissions. I've updated `UserManagement` to have this permission though to follow the update_email permissions.

backend/app/views/spree/admin/users/_form.html.erb
kennyadsl: I think that if someone is adding some non-password related fields into the password dedicated div it's ok to make this change, we can't handle every scenario. The problem is if someone is using this hook in the `insert_after` or `insert_before` deface action, which is legit and the added parts will be hidden at that point. :( 

That said, your solution is probably the safest one in terms of backward compatibility.

PR #3395, opened 2019-10-23
Remove Deprecated EmailValidator

### LINKED ISSUE #1794: Replace E-mail Validator

JDutil:
**Description**

The email validator was deprecated Mar 15th 2018 almost 18 months ago
as part of: 6d8fd3c

The deprecation warning was part of v2.6.0rc1, and remains to this day
in v2.10.0.beta1 I think it's time to remove it altogether. It's been
4 minor version upgrades now.

Fixes #1794

**Checklist:**

### DIFF
--- a/core/lib/spree/core/validators/email.rb
+++ b/core/lib/spree/core/validators/email.rb
@@ -21,10 +21,3 @@ module Spree
     end
   end
 end
-
-# @private
-EmailValidator = ActiveSupport::Deprecation::DeprecatedConstantProxy.new(
-  'EmailValidator',
-  'Spree::EmailValidator',
-  message: "EmailValidator is deprecated! Use Spree::EmailValidator instead.\nChange `validates :email, email: true` to `validates :email, 'spree/email' => true`\n"
-)
--- a/core/lib/spree/core/validators/email.rb
+++ b/core/lib/spree/core/validators/email.rb
@@ -15,7 +15,7 @@ module Spree
     EMAIL_REGEXP = /\A([^@\.]|[^@\.]([^@\s]*)[^@\.])@([^@\s]+\.)+[^@\s]+\z/
 
     def validate_each(record, attribute, value)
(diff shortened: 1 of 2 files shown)


spaghetticode:
While I agree it's ok to remove `EmailValidator`, I'm not sure about completely removing the email validator `Spree::EmailValidator` as well. 

I think a more conservative approach may be to deprecate `Spree::EmailValidator` first, as somebody may be using it, and remove it with a later version of Solidus.

core/lib/spree/core/validators/email.rb
```diff
@@ -1,30 +0,0 @@
...
-    def validate_each(record, attribute, value)
-      unless value =~ EMAIL_REGEXP
-        record.errors.add(attribute, :invalid, { value: value }.merge!(options))
-      end
-    end
-  end
-end
-
-# @private
-EmailValidator = ActiveSupport::Deprecation::DeprecatedConstantProxy.new(
```
kennyadsl: This is the only part to remove if we want to get rid of the deprecation message.

JDutil: Yeah makes sense I didn't notice that it was a proxy 👍

JDutil:
@kennyadsl updated 👍

kennyadsl:
I think specs are failing due to the `core/lib/spree/core/validators/email.rb → core/lib/spree/validators/email.rb` rename, which should not happen. Also, can you please remove the changelog entry? It will be autogenerated using the PR title, so it's better to change it and remove the entry, thanks!

JDutil:
@kennyadsl updated 👍

PR #3396, opened 2019-10-23
Update custom search sample query

peterberkenbosch:
Change the query to have an uppercase R to actually return the sample products.

**Description**

**Checklist:**

### DIFF
--- a/guides/source/developers/preferences/class-extension-points.html.md
+++ b/guides/source/developers/preferences/class-extension-points.html.md
@@ -16,7 +16,7 @@ begins with `class_name_attribute` defines a different extension point.
 ## Example usage
 
 When you extend a class, you can change the behavior of a single feature or your
-entire store. 
+entire store.
 
 For example, you can change what appears in a customer's search results by
 extending the [`Spree::Core::Search::Base`][search-base]. This way, you do not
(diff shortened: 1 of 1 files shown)


guides/source/developers/preferences/class-extension-points.html.md
```diff
@@ -51,7 +51,7 @@ module MyStore
     private
 
     def get_base_scope
-      super.where("name LIKE '%rails%'")
+      super.where("name LIKE '%Rails%'")
```
JDutil: Are my sandbox seeds not loading all products? I only see "Ruby" products, and not any "Rails" products. If the seeds only populate "Ruby" products perhaps we should change this query to `%Ruby%`.

JDutil: Looks like the sandbox only has "Ruby" products  https://github.com/solidusio/solidus/blob/master/sample/db/samples/products.rb while the

JDutil: Looks like RoR products we're recently removed: https://github.com/solidusio/solidus/commit/8221dcb5a462cf25a8117dcc326cf2db148bf862 I've created a PR to remove the old sample images: https://github.com/solidusio/solidus/pull/3397

kennyadsl:
@peterberkenbosch can you please take a look at the @JDutil comment here? Thanks!

peterberkenbosch:
This should be all good now @kennyadsl, thanks @JDutil

PR #3398, opened 2019-10-24
More eager loading in admin and api

softr8:
**Description**
With recent changes, some previous eager loading attempts are no longer applicable, I'm trying to update as much as possible.

I've added the option to specify `includes` in the resource controller when the resource depends of another one and is specified via `belongs_to`. We can now use it everywhere

**Checklist:**


core/app/models/concerns/spree/user_methods.rb
```diff
@@ -72,8 +72,10 @@ def total_available_store_credit
     end
     deprecate total_available_store_credit: :available_store_credit_total, deprecator: Spree::Deprecation
 
-    def available_store_credit_total(currency:)
-      store_credits.reload.to_a.
+    def available_store_credit_total(currency:, force_reload: true)
+      # This reload makes eager loading not to work when hitting read only endpoints
+      store_credits.reload if force_reload
+      store_credits.to_a.
```
spaghetticode: while we are at it, I think it may be worth, in a dedicated commit, to slightly optimize this method... I don't see any reason to not use ActiveRecord for selecting the store credits with the right currency:

`store_credits.where(currency: currency).sum(&:amount_remaining)`

what do you think?

core/app/models/spree/order.rb
```diff
@@ -671,13 +672,13 @@ def add_store_credit_payments
 
     def covered_by_store_credit?
       return false unless user
-      user.available_store_credit_total(currency: currency) >= total
+      user.available_store_credit_total(currency: currency, force_reload: false) >= total
```
spaghetticode: I think we're changing the behavior here, as it was `reload`ing previously... can you explain the reason for this?

core/app/models/spree/order.rb
```diff
@@ -671,13 +672,13 @@ def add_store_credit_payments
...
     end
     alias_method :covered_by_store_credit, :covered_by_store_credit?
 
     def total_available_store_credit
       return 0.0 unless user
-      user.available_store_credit_total(currency: currency)
+      user.available_store_credit_total(currency: currency, force_reload: false)
```
spaghetticode: same as above, I think the behavior is changed now

core/app/models/spree/order.rb
```diff
@@ -686,9 +687,9 @@ def order_total_after_store_credit
 
     def total_applicable_store_credit
       if can_complete? || complete?
-        payments.store_credits.valid.sum(:amount)
+        valid_store_credit_payments.to_a.sum(&:amount)
```
spaghetticode: Like above, I think the previous version was more performant, as it did all calculations in the DB without instantiating and cycling ruby objects

core/app/models/spree/order.rb
```diff
@@ -686,9 +687,9 @@ def order_total_after_store_credit
...
       else
-        [total, (user.try(:available_store_credit_total, currency: currency) || 0.0)].min
+        [total, (user.try(:available_store_credit_total, currency: currency, force_reload: false) || 0.0)].min
```
spaghetticode: I see another change of behavior... by the way, this may be totally legit, I did not investigate, can you share your motivation for this change? 🙏

api/app/views/spree/api/orders/_order.json.jbuilder
```diff
@@ -3,7 +3,7 @@
 json.cache! [I18n.locale, order] do
   json.(order, *order_attributes)
   json.display_item_total(order.display_item_total.to_s)
-  json.total_quantity(order.line_items.sum(:quantity))
+  json.total_quantity(order.line_items.to_a.sum(&:quantity))
```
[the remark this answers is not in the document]
tvdeyen: The order should already have preloaded the line items. If we do a `sum` in the database than we always load it from the database. And even if the line items have not been preloaded yet, they would be loaded by this call. It is always preferable to use array methods instead of `where` statements as the latter will always - no matter what - load from the database, while the former would only load if necessary.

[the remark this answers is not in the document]
tvdeyen: Great write up https://www.speedshop.co/2019/01/10/three-activerecord-mistakes.html

core/app/models/spree/order.rb
tvdeyen: It is not more performant this way. Quite contrary this only does the math in the database if the payments have not been loaded yet. If they already were loaded the sum in database approach is actually slower, as it unnecessarily loads the payments again.

core/app/models/concerns/spree/user_methods.rb
softr8: Not sure about it, the thing is that we're trying to re-use as much as possible all data pre-loaded with eager loading, and if we do it, it'd execute extra queries to the db.

api/app/views/spree/api/orders/_order.json.jbuilder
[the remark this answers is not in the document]
spaghetticode: Oh, I see:

```
order = create :order_with_line_items, line_items_count: 10

Benchmark.bm do |bm|
  bm.report { order.line_items.sum :quantity }
  bm.report { order.line_items.to_a.sum &:quantity }
  bm.report { order.line_items.reload.to_a.sum &:quantity }
end

   user     system      total        real
0.000908   0.000217   0.001125 (  0.001124)
0.000036   0.000006   0.000042 (  0.000041)
0.000956   0.000162   0.001118 (  0.001122)
```
so, as long as the association has already been loaded from the DB, using ruby is so much faster... and when not, the numbers seem to be in the same ballpark. Then this makes definitely sense 👍

spaghetticode:
@softr8 I reviewed the changes, now I understand using `to_a` should be beneficial for performances 👍 

I see specs are red, can you look if the issue is related to these changes?

PR #3406, opened 2019-10-25
Document the real meaning of checkout#set_state_if_present

elia:
**Description**

The fact that this method intended to stop customers from skipping ahead is not
easily drawn out by the method name or the contents.
Renaming & refactoring both have risks, so, for now, I'll content
myself with a comment.

**Checklist:**

### DIFF
--- a/frontend/app/controllers/spree/checkout_controller.rb
+++ b/frontend/app/controllers/spree/checkout_controller.rb
@@ -137,6 +137,10 @@ module Spree
       redirect_to(spree.cart_path) && return unless @order
     end
 
+    # Allow the customer to only go back or stay on the current state
+    # when trying to change it via params[:state]. It's not allowed to
+    # jump forward and skip states (unless #skip_state_validation? is
+    # truthy).
     def set_state_if_present
       if params[:state]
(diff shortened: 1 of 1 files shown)


frontend/app/controllers/spree/checkout_controller.rb
```diff
@@ -137,6 +137,10 @@ def load_order
       redirect_to(spree.cart_path) && return unless @order
     end
 
+    # Allow the customer to only go back or stay on the current state
+    # when trying to change it via params[:state]. It's not allowed to
+    # jump forward and skip states (unless #skip_state_validation? is
+    # truthy).
     def set_state_if_present
```
kennyadsl: What about also changing this method name? We could keep this one so that it only prints the deprecation message and call the new one, WDYT?

elia: I'd be very much in favor of doing so! I'll make a note to do another PR with the full deprecation work.

elia: https://github.com/solidusio/solidus/pull/3496 ✨

PR #3407, opened 2019-10-25
Introduce ActionView::Component for backend components

elia:
**Description**

Creating a set of components for the backend will help in having both
a more consistent UI and will give developers a cleaner way to extend Solidus.

https://github.com/github/actionview-component#readme

_ActionView::Component will come as a standard Rails feature in v6.1_

The provided changes regarding the use of the Pill component are incomplete and I'll finish converting existing view once reasonable consensus from the core-team is reached.

Further components can be added and introduced in future PRs, unless requested otherwise.

**Checklist:**

### DIFF
--- a/backend/app/components/spree/admin/pill_component.rb
+++ b/backend/app/components/spree/admin/pill_component.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class Spree::Admin::PillComponent < ActionView::Component::Base
+  STATES = %i[
+    active
+    complete
+    error
+    inactive
+    neutral
+    pending
+    text
+    warning
+  ]
+
+  validates :state, inclusion: STATES, unless: -> { state.is_a? String }
+
+  def initialize(state: "neutral", text: nil)
+    @state = state
+    @text = text
+  end
+
+  attr_reader :state
+
+  def text
+    @text || content || I18n.t("spree.#{state}", default: nil) || @state
+  end
+end
--- a/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
@@ -33,9 +33,7 @@
             <%= adjustment_reason.name %>
           </td>
           <td>
-            <span class="pill pill-<%= adjustment_reason.active? ? 'active' : 'inactive' %>">
-              <%= t(adjustment_reason.active? ? :active : :inactive, scope: 'spree') %>
-            </span>
+            <%= render Spree::Amin::PillComponent, status: adjustment_reason.active? ? :active : :inactive %>
           </td>
           <td class="actions">
             <%= link_to_edit adjustment_reason, no_text: true %>
--- a/backend/app/components/spree/admin/pill_component.html.erb
+++ b/backend/app/components/spree/admin/pill_component.html.erb
@@ -1,0 +1,1 @@
+<span class="pill pill-<%= state %>"><%= text %></span>
--- a/backend/app/views/spree/admin/cancellations/index.html.erb
+++ b/backend/app/views/spree/admin/cancellations/index.html.erb
@@ -37,9 +37,9 @@
           </td>
           <td class="inventory-unit-quantity align-center">1</td>
           <td class="inventory-unit-state">
-            <span class="pill pill-<%= inventory_unit.state %>">
+            <%= render Spree::Admin::PillComponent, state: inventory_unit.state do %>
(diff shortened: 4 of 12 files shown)


backend/app/views/spree/admin/adjustment_reasons/index.html.erb
```diff
@@ -33,9 +33,7 @@
             <%= adjustment_reason.name %>
           </td>
           <td>
-            <span class="pill pill-<%= adjustment_reason.active? ? 'active' : 'inactive' %>">
-              <%= t(adjustment_reason.active? ? :active : :inactive, scope: 'spree') %>
-            </span>
+            <%= render Spree::Amin::PillComponent, status: adjustment_reason.active? ? :active : :inactive %>
```
tvdeyen: typo: Should be `Spree::Admin::PillComponent`

elia: Thanks for catching this! I guess that view is not really covered by specs 😅

backend/app/components/spree/admin/pill_component.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(state: "neutral", text: nil)
+    @state = state
+    @text = text
+  end
+
+  attr_reader :state
+
+  def text
+    @text || content || I18n.t("spree.#{state}", default: nil) || @state
```
spaghetticode: `@state` => `state`

backend/app/views/spree/admin/adjustment_reasons/index.html.erb
kennyadsl: and `status` -> `state`

tvdeyen: Now that I see this I think we should rename `state` into `status` and `STATES` into `STATUSES` (Or even `STATUS` as the plural of Status is Status in Latin, German and even formal English, see https://english.stackexchange.com/questions/877/what-is-the-plural-form-of-status).

The problem with `state` is that it can be misunderstood as State (part of country) we even have `Spree::State` as model for a Part of Country.

backend/app/components/spree/admin/pill_component.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class Spree::Admin::PillComponent < ActionView::Component::Base
+  STATES = %i[
```
tvdeyen: This should be `STATUS` to avoid confusion with `Spree::State` (part of Country) and the formal plural of status is status (https://english.stackexchange.com/questions/877/what-is-the-plural-form-of-status)

backend/app/components/spree/admin/pill_component.rb
```diff
@@ -0,0 +1,27 @@
...
+    inactive
+    neutral
+    pending
+    text
+    warning
+  ]
+
+  validates :state, inclusion: STATES, unless: -> { state.is_a? String }
+
+  def initialize(state: "neutral", text: nil)
```
tvdeyen:
```suggestion
  def initialize(status: "neutral", text: nil)
```

backend/app/components/spree/admin/pill_component.rb
```diff
@@ -0,0 +1,27 @@
...
+    @state = state
+    @text = text
+  end
+
+  attr_reader :state
```
tvdeyen:
```suggestion
  attr_reader :status
```

backend/app/components/spree/admin/pill_component.rb
```diff
@@ -0,0 +1,27 @@
...
+    neutral
+    pending
+    text
+    warning
+  ]
+
+  validates :state, inclusion: STATES, unless: -> { state.is_a? String }
+
+  def initialize(state: "neutral", text: nil)
+    @state = state
```
tvdeyen:
```suggestion
    @status = status
```

backend/app/views/spree/admin/adjustment_reasons/index.html.erb
kennyadsl: I agree, and can also be confused with the React State, since we are adopting the component terminology from that world.

tvdeyen: Yes, that's true :)

kennyadsl:
Closing this. We will work on a strategy for the future admin shortly.

elia:
😄 definitely!

PR #3408, opened 2019-10-25
Provide a fast, easily accessible, development server

elia:
**Description**
This is intended for when a fast development cycle or manual
inspection is needed, the server starts in a handful of seconds and
leaves no traces.

- the DB is an in-memory SQLite
- logs are sent to STDOUT
- defaults to port 3000
- easy access to /admin with a dummy admin-user (pre signed-in)
- easy to discover with bin/server
- compatible with [puma-dev](https://github.com/puma/puma-dev)

**Checklist:**

image

_with puma-dev_

image

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -1,0 +1,137 @@
+# Suggested rackup options: -O Threads=1:1 -p 3000
+
+require 'bundler/setup'
+require 'rails'
+require 'rails/all'
+require 'solidus'
+
+module Paperclip::Storage::InMemory
+  FS = Hash.new { raise "image not found" }
+
+  Paperclip.interpolates(:data_uri) do |attachment, style|
+    %{data:#{attachment.content_type};base64,#{FS[attachment.path(style)]}}
+  end
+
+  def self.extended base
+    base.instance_eval do
+      @options[:path] = @options[:path].gsub(':rails_root/public/system/', '')
+      @options[:url] = ':data_uri'
+    end
+  end
+
+  def exists?(style_name = default_style)
+    FS.key?(path(style_name))
+  end
+
+  def flush_writes #:nodoc:
+    @queued_for_write.each do |style_name, file|
+      FS[path(style_name)] = Base64.encode64(file.read)
+      file.rewind
+    end
+    after_flush_writes # allows attachment to clean up temp files
+    @queued_for_write = {}
+  end
+
+  def flush_deletes #:nodoc:
+    @queued_for_delete.each { |path| FS.delete(path) }
+    @queued_for_delete = []
+  end
+
+  def copy_to_local_file(style, local_dest_path)
+    File.write(local_dest_path, Base64.decode64(FS[path(style)]))
+  end
+end
+
+module RailsApp
+  class Application < Rails::Application
+    config.root                                       = __dir__
+    config.cache_classes                              = true
+    config.eager_load                                 = false
+    config.public_file_server.enabled                 = true
+    config.public_file_server.headers                 = { 'Cache-Control' => 'public, max-age=3600' }
+    config.consider_all_requests_local                = true
+    config.action_controller.allow_forgery_protection = false
+    config.action_controller.perform_caching          = false
+    config.action_dispatch.show_exceptions            = false
+    config.action_mailer.perform_deliveries           = false
+    config.active_support.deprecation                 = :stderr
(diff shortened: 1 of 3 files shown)


config.ru
```diff
@@ -0,0 +1,93 @@
...
+    config.action_controller.perform_caching          = false
+    config.action_dispatch.show_exceptions            = false
+    config.action_controller.allow_forgery_protection = false
+    config.active_support.deprecation                 = :stderr
+    config.secret_key_base                            = '49837489qkuweoiuoqwe'
+
+    config.logger = ActiveSupport::Logger.new(STDOUT)
+
+    config.middleware.delete Rack::Lock
+    config.middleware.delete ActionDispatch::Flash
```
tvdeyen: I think we should leave this in. If we want to use this a as an easy way to review PRs (instead of generating a sandbox) we want to have flash messages.

config.ru
```diff
@@ -0,0 +1,93 @@
...
+      mount Spree::Core::Engine, at: '/'
+      get '/application/with_assignments' => 'application#with_assignments'
+    end
+
+    %w[core frontend backend api].each do |lib|
+      config.assets.paths << "#{__dir__}/#{lib}/lib/spree/testing_support/dummy_app/assets/javascripts"
+      config.assets.paths << "#{__dir__}/#{lib}/lib/spree/testing_support/dummy_app/assets/stylesheets"
+    end
+
+    config.assets.debug = true
```
tvdeyen: Setting this to false would be much faster.

spaghetticode:
@elia this is a great idea! I wonder if we could make the FE fully usable as well somehow, but I guess the effort would not be small.

elia:
@tvdeyen thanks for the review, I've followed your suggestions!
@spaghetticode I've updated the code, it now has two additional routes that will load seeds and samples: `localhost:3000/solidus/seed` and `localhost:3000/solidus/load_samples`, that way they can be run depending on the situation.

Also Paperclip is kept in memory as much as possible (identify and convert still use tempfiles(

config.ru
```diff
@@ -0,0 +1,121 @@
...
+  retrieve_connection.tap { |connection| define_singleton_method(:connection) { connection }   }
+end
+
+# Migrations
+ActiveRecord::Migration.verbose = false
+# Prepare the db with a basic spree structure for LegacyUser
+ActiveRecord::Base.connection.migration_context.run(:up, '20160101010000_solidus_one_four'.to_i)
+ActiveRecord::Tasks::DatabaseTasks.migrate # Complete the migration
+
+Spree::LegacyUser.class_eval do
```
tvdeyen: Why is this class_eval necessary?

config.ru
```diff
@@ -0,0 +1,121 @@
...
+    routes.append do
+      mount Spree::Core::Engine, at: '/'
+      get '/solidus/seed' => ->env { Spree::Core::Engine.load_seed; [200, {}, ['Solidus seeds loaded']] }
+      get '/solidus/load_samples' => ->env { SpreeSample::Engine.load_samples; [200, {}, ['Solidus samples loaded']] }
+    end
+  end
+end
+
+class ApplicationController < ActionController::Base
+  include Rails.application.routes.url_helpers
```
tvdeyen: Why do we need to include the routes here? Isn't Rails including them for us?

kennyadsl:
@elia thanks, this is great. I have a final question: do you think it's worth mentioning somewhere in guides or in the README?

tvdeyen:
> `@kennyadsl` `@tvdeyen` you think this could still be helpful? The sandbox improved a lot since 2019, it's easy to regenerate, and conveniently responds to `bin/rails` commands, but there might still be some use cases for this.
> 
> Please let me know and I can extract the first commit to a different PR if needed, or add info to the readme if we want to merge it.

@elia bare with me, but this has been a while. Can you elaborate what would be the benefit compared to the sandbox? Is it just for "don't need a sandbox"? If the sandbox already provides anything we need as a fast local dev server I would like to reduce the maintenance burden of maintain two dev environments, although I really appreciate the effort put into this.

elia:
@tvdeyen yeah, I think a bunch of things from this actually ended up in the sandbox, I just wanted to be sure I was not gonna close this if it had a legit use case (looks like there isn't), and the double maintenance is a great point 👍

tvdeyen:
👍🏻

PR #3415, opened 2019-10-30
Remove last migration's spec file

kennyadsl:
**Description**

This migration spec has done its job for users upgrading up to `v2.9` and now I think we can safely remove it assuming users are upgrading Solidus one version at the time.

Also, the migration is quite complex since it uses some ActiveRecord internals that may change with time.

**Checklist:**


kennyadsl:
@spaghetticode I'd love to keep it but, while testing a branch with another migration, I've found some issue that I didn't consider adding this spec. For example, as is, it gives for granted that the migration under test is the last one, and if/when we'll add a new migration this spec will need some adjustment. This random experiment showed the weakness of testing migrations.

Also, there's a lot of boilerplate code, which I tried to extract into some RSpec migration helpers [here](https://github.com/nebulab/solidus/pull/27/files), which I leave for reference if we want to add it back for new migrations. 

Going to merge this since it seems like I have a stronger opinion here. 🙂

PR #3416, opened 2019-10-30
Several small refactors to promotions code

kennyadsl:
**Description**
I've worked on these refactors while working on #3355, which is closed now but I think these small preliminary changes could still be beneficial for the codebase. 

**Checklist:**

### DIFF
--- a/core/spec/models/spree/calculator/shipping/price_sack_spec.rb
+++ b/core/spec/models/spree/calculator/shipping/price_sack_spec.rb
@@ -5,7 +5,7 @@ require 'shared_examples/calculator_shared_examples'
 
 RSpec.describe Spree::Calculator::Shipping::PriceSack, type: :model do
   let(:calculator) do
-    calculator = Spree::Calculator::PriceSack.new
+    calculator = described_class.new
     calculator.preferred_minimal_amount = 5
     calculator.preferred_normal_amount = 10
     calculator.preferred_discount_amount = 1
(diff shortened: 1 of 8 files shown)


core/spec/models/spree/calculator/shipping/price_sack_spec.rb
```diff
@@ -14,22 +14,31 @@
...
 
-  # Regression test for https://github.com/spree/spree/issues/714 and https://github.com/spree/spree/issues/739
-  it "computes with an order object" do
-    calculator.compute(order)
-  end
+    before do
+      # This hack is due to our factories not being so smart to understand
+      # that they should create line items with the price of the associated
+      # variant by default.
+      allow_any_instance_of(Spree::Stock::ContentItem).to receive(:price) { amount }
```
spaghetticode: Do you think it would make sense (be possible, simple,  and useful for other scenarios) to change the factory in order to avoid this stubbing?

kennyadsl: I had the same thought but I think we have to change the line item price having a dynamic price, based on the variant price. I think that this would break a lot of specs in stores that are relying on the static price of line items in our core factories, and that's why I think this is the best trade-off. 

I'm open to making an attempt to create an alternative line_item factory with a dynamic price and trying to deprecate the current one. Do you think it's worth it?

spaghetticode: I think it is the normal scenario for line items to have prices bound to the ones of their variants... I would consider a price difference to be more of an edge case, so I think that factory could be later changed to reflect a more regular use case. 
But of course I would not consider this an urgent matter at all. We may just open an issue and see how it goes from there.

kennyadsl: Agree, let me do that!

PR #3425, opened 2019-11-10
Remove RABL remnants

JuanCrg90:
**Description**
RABL was removed from the project on version 2.4.0 (See #2147 #2146).
This updates the responder template to match with Jbuilder and updates
the README file.

**Checklist:**

### DIFF
--- a/api/lib/spree/api/responders/jbuilder_template.rb
+++ b/api/lib/spree/api/responders/jbuilder_template.rb
@@ -3,7 +3,7 @@
 module Spree
   module Api
     module Responders
-      module RablTemplate
+      module JbuilderTemplate
         def to_format
           if template
             render template, status: options[:status] || 200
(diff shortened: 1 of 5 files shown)


api/lib/spree/api/responders/jbuilder_template.rb
```diff
@@ -3,7 +3,7 @@
 module Spree
   module Api
     module Responders
-      module RablTemplate
+      module JbuilderTemplate
```
kennyadsl: I think we should keep the `RablTemplate` class here and deprecate it since it could be used in host applications and it's just matter of adding a `DeprecatedConstantProxy`:

```ruby
RablTemplate = ActiveSupport::Deprecation::DeprecatedConstantProxy.new('RablTemplate', 'JbuilderTemplate')
```

 and a spec that ensures it's working as expected.

JuanCrg90: Updated 😄 what do you think?

PR #3426, opened 2019-11-14
Caching product results in frontend

softr8:
Taking advantage of [Rails collection caching](https://guides.rubyonrails.org/caching_with_rails.html#collection-caching) to perform multi get if possible
it fetches multiple items at once instead of one by one if the cache storage driver allows it.

Added as point of reference about how to make Solidus stores even faster.

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/caching.rb
+++ b/core/lib/spree/testing_support/caching.rb
@@ -10,6 +10,15 @@ module Spree
       def clear_cache_events
         @cache_write_events = []
       end
+
+      def render_collection_cache_hits_for(partial = nil)
+        @render_collection_events.inject(0) do |final, event|
+          if event[:identifier].include?(partial)
+            final += event[:cache_hits]
+          end
+          final
+        end
+      end
     end
   end
 end
@@ -23,6 +32,11 @@ RSpec.configure do |config|
     ActiveSupport::Notifications.subscribe("write_fragment.action_controller") do |_event, _start_time, _finish_time, _, details|
       @cache_write_events ||= []
       @cache_write_events << details
+    end
+
+    ActiveSupport::Notifications.subscribe("render_collection.action_view") do |_event, _start_time, _finish_time, _, details|
+      @render_collection_events ||= []
+      @render_collection_events << details
     end
   end
 
--- a/frontend/app/views/spree/shared/_product_tile.html.erb
+++ b/frontend/app/views/spree/shared/_product_tile.html.erb
@@ -1,0 +1,15 @@
+<% url = spree.product_path(product, taxon_id: @taxon.try(:id)) %>
+<li id="product_<%= product.id %>" class="columns three <%= cycle("alpha", "secondary", "", "omega secondary", name: "classes") %>" data-hook="products_list_item" itemscope itemtype="http://schema.org/Product">
+  <div class="product-image">
+    <%= link_to(render('spree/shared/image', image: product.gallery.images.first, size: :small, itemprop: "image"), url, itemprop: 'url') %>
+  </div>
+  <%= link_to truncate(product.name, length: 50), url, class: 'info', itemprop: "name", title: product.name %>
+  <div itemprop="offers" itemscope itemtype="http://schema.org/Offer">
+    <% if price = product.price_for_options(current_pricing_options)&.money %>
+      <span class="price selling" itemprop="price" content="<%= price.to_d %>">
+        <%= price.to_html %>
+      </span>
+    <% end %>
+    <span itemprop="priceCurrency" content="<%= current_pricing_options.currency %>"></span>
+  </div>
+</li>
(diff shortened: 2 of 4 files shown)


frontend/app/views/spree/shared/_products.html.erb
```diff
@@ -23,25 +23,10 @@
...
-            <% end %>
-            <span itemprop="priceCurrency" content="<%= current_pricing_options.currency %>"></span>
-          </span>
-        <% end %>
-      </li>
-    <% end %>
+    <%= render partial: 'spree/shared/product_tile',
+               collection: products,
+               as: :product,
+               cached: Proc.new { |product| [I18n.locale, current_pricing_options, product, @taxon] } %>
```
tvdeyen: TIL 👍

frontend/app/views/spree/shared/_product_tile.html.erb
```diff
@@ -0,0 +1,17 @@
+<% url = spree.product_path(product, taxon_id: @taxon.try(:id)) %>
+<li id="product_<%= product.id %>" class="columns three <%= cycle("alpha", "secondary", "", "omega secondary", name: "classes") %>" data-hook="products_list_item" itemscope itemtype="http://schema.org/Product">
+  <% cache(@taxon.present? ? [I18n.locale, current_pricing_options, @taxon, product] : [I18n.locale, current_pricing_options, product]) do %>
+    <div class="product-image">
+      <%= link_to(render('spree/shared/image', image: product.gallery.images.first, size: :small, itemprop: "image"), url, itemprop: 'url') %>
+    </div>
+    <%= link_to truncate(product.name, length: 50), url, class: 'info', itemprop: "name", title: product.name %>
+    <span itemprop="offers" itemscope itemtype="http://schema.org/Offer">
+            <% if price = product.price_for(current_pricing_options) %>
```
kennyadsl: Indentation is a bit off here. Can you please take a look?

frontend/app/views/spree/shared/_product_tile.html.erb
```diff
@@ -0,0 +1,17 @@
+<% url = spree.product_path(product, taxon_id: @taxon.try(:id)) %>
+<li id="product_<%= product.id %>" class="columns three <%= cycle("alpha", "secondary", "", "omega secondary", name: "classes") %>" data-hook="products_list_item" itemscope itemtype="http://schema.org/Product">
+  <% cache(@taxon.present? ? [I18n.locale, current_pricing_options, @taxon, product] : [I18n.locale, current_pricing_options, product]) do %>
```
kennyadsl: Is this cache different from the outer one defined when we call the partial? Do you think we could use the same one? 

Maybe it's also worth extracting the cache key somewhere else if they are identical? Having a shared/uniform place where to store how we cache things could be a good idea to explore, probably into another PR.

softr8: ooh!! good catch, I just moved the code from the other partial without really studying it, will improve it

JuanCrg90:
looks legit

```
Your job ran 5570 tests with 1 failure
products busts the cache when a product is updated - spec.features.caching.products_spec
spec/features/caching/products_spec.rb
Failure/Error: expect(cache_writes.count).to eq(2)

  expected: 2
       got: 1

  (compared using ==)
./spec/features/caching/products_spec.rb:27:in `block (2 levels) in <top (required)>'
```

core/lib/spree/testing_support/caching.rb
```diff
@@ -24,6 +28,11 @@ def clear_cache_events
       @cache_write_events ||= []
       @cache_write_events << details
     end
+
+    ActiveSupport::Notifications.subscribe("render_collection.action_view") do |_event, _start_time, _finish_time, _, details|
```
kennyadsl: Sorry, I quite don't get this. Can you please explain why we should treat `render_collection.action_view` as a cache write event?

softr8: Rails fires this event whenever a collection is rendered, and it also includes information about how many elements were fetched from the cache and how many rendered from scratch, that's why I'm using it to count how many were fetched from the cache in the method above.

kennyadsl: Ok, maybe I get it better. Still not sure why putting those events into the `@cache_write_events` array since they are not "writes". Maybe we could create a new object and put into it just the relevant information that we need? Something like:

```
ActiveSupport::Notifications.subscribe("render_collection.action_view") do |_event, _start_time, _finish_time, _, details|
  @cached_collection_items_count ||= 0
  @cached_collection_items_count += details[: cache_hits]
```

What do you think?

gsmendoza:
Hello @softr8 ! Can you port this to https://github.com/solidusio/solidus_starter_frontend? SolidusStarterFrontend is the new default frontend of Solidus. We will be removing `solidus_frontend` from the Solidus meta-package gem in v4. Please see https://github.com/solidusio/solidus/blob/a7de972427a86f1955cd7edb5bd0f0c6a688c347/CHANGELOG.md#new-solidus-starter-frontend.

kennyadsl:
@softr8 closing this, thanks!

softr8:
Ok, will do!

PR #3428, opened 2019-11-15
Set error flash when unsuccesful destroy using HTML format

mamhoff:
**Description**

When destroying a resource using the admin resource controller and failing because of some `before_destroy` callback throwing `:abort` with an `ActiveModel::Error`, and using HTML, the user will now see an error message after being redirected to the list of resources. 

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/resource_controller.rb
+++ b/backend/app/controllers/spree/admin/resource_controller.rb
@@ -107,8 +107,14 @@ class Spree::Admin::ResourceController < Spree::Admin::BaseController
     else
       invoke_callbacks(:destroy, :fails)
       respond_with(@object) do |format|
-        format.html { redirect_to location_after_destroy }
-        format.js { render status: :unprocessable_entity, plain: @object.errors.full_messages.to_sentence }
+        message = @object.errors.full_messages.to_sentence
+        format.html do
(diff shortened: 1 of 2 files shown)


backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -99,16 +99,27 @@ def destroy
...
       respond_with(@object) do |format|
-        format.html { redirect_to location_after_destroy }
-        format.js   { render partial: "spree/admin/shared/destroy" }
+        message = flash_message_for(@object, :successfully_removed)
+        format.html do
+          flash[:success] = message
+          redirect_to location_after_destroy
+        end
+        format.js do
+          render plain: message
```
tvdeyen: How do we show the success message in success case for a javascript request? The `.delete-resource` click handler only sets the message in error case not success case. https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/admin.js#L89

Can we keep the partial or update the handler to it also supports flash messages?

mamhoff: Fixed, thank you!

kennyadsl:
Thanks, Martin!

PR #3431, opened 2019-11-19
Improve the extensibility of Rules::ItemTotal

elia:
**Description**

- Make operators extendable without reopening the method
- Allow to overwrite/extend the total
- Allow to overwrite/extend the threshold
- Simplify #eligible? making it more readable

This change stems from the need to add the `:eq`/`===` operator and to modify the total, beyond that need it should make the code more readable and linear.

**Checklist:**

### DIFF
--- a/core/app/models/spree/promotion/rules/item_total.rb
+++ b/core/app/models/spree/promotion/rules/item_total.rb
@@ -5,6 +5,9 @@ module Spree
     module Rules
       # A rule to apply to an order greater than (or greater than or equal to)
       # a specific amount
+      #
+      # To add extra operators please override `self.operators_map` or any other helper method.
+      # To customize the error message you can also override `ineligible_message`.
       class ItemTotal < PromotionRule
         include ActiveSupport::Deprecation::DeprecatedConstantAccessor
 
@@ -71,10 +74,13 @@ module Spree
         end
 
         def ineligible_message
-          if preferred_operator.to_s == 'gte'
+          case preferred_operator.to_s
+          when 'gte'
             eligibility_error_message(:item_total_less_than, amount: formatted_amount)
-          else
+          when 'gt'
             eligibility_error_message(:item_total_less_than_or_equal, amount: formatted_amount)
+          else
+            eligibility_error_message(:item_total_doesnt_match_with_operator, amount: formatted_amount, operator: preferred_operator)
           end
         end
 
--- a/core/app/models/spree/promotion/rules/item_total.rb
+++ b/core/app/models/spree/promotion/rules/item_total.rb
@@ -8,7 +8,7 @@ module Spree
       class ItemTotal < PromotionRule
         preference :amount, :decimal, default: 100.00
         preference :currency, :string, default: ->{ Spree::Config[:currency] }
-        preference :operator, :string, default: '>'
+        preference :operator, :string, default: 'gt'
 
         OPERATORS = ['gt', 'gte']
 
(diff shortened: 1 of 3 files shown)


core/app/models/spree/promotion/rules/item_total.rb
```diff
@@ -6,20 +6,36 @@ module Rules
...
       # a specific amount
       class ItemTotal < PromotionRule
+        include ActiveSupport::Deprecation::DeprecatedConstantAccessor
+
         preference :amount, :decimal, default: 100.00
         preference :currency, :string, default: ->{ Spree::Config[:currency] }
         preference :operator, :string, default: '>'
 
-        OPERATORS = ['gt', 'gte']
+        OPERATORS_MAP = {
```
kennyadsl: Is there any reason to don't use a method instead of a constant if we are deprecating this?

elia: No reason, I used a constant just because there was a constant before, I'll change it to a method 👍

core/app/models/spree/promotion/rules/item_total.rb
```diff
@@ -6,20 +6,43 @@ module Rules
       # A rule to apply to an order greater than (or greater than or equal to)
       # a specific amount
       class ItemTotal < PromotionRule
+        include ActiveSupport::Deprecation::DeprecatedConstantAccessor
+
         preference :amount, :decimal, default: 100.00
         preference :currency, :string, default: ->{ Spree::Config[:currency] }
-        preference :operator, :string, default: '>'
+        preference :operator, :string, default: 'gt'
```
kennyadsl: What about existing rules that are using the old operator syntax (`>`, `>=`) instead of `gt`, `gte`?

kennyadsl: Oh right, now I recall it! I think a separate commit would be enough.

elia:
>  I'd prefer to be free to change/remove private methods in core classes without any concern.

The original drive for this PR was to enable creating rules such as the following:

```rb
      class ItemTotalBeforeTax < ItemTotal
        # Rely on the item total with the adjustments already applied
        def total_for(order)
          order.item_total_before_tax
        end
      end
```

and

```rb
      module ItemTotalDecorator
        def operators_map
          super.merge(eq: :==)
        end

        Spree::Promotion::Rules::ItemTotal.prepend self
      end
```

where it's really easy to cherry-pick just the method you want to target. 
Would it be better to have completely independent classes and let them have their own life instead of subclassing or decorating ItemTotal?

core/app/models/spree/promotion/rules/item_total.rb
elia: ✅ Done.

aldesantis:
On the matter of private methods, I think we may want to make them public, but in this case having private methods may also be an acceptable exception. The API seems to make 100% sense to me.

spaghetticode (approved):
@elia thank you! Looks good, I left a comment on a small improvement, let me know your opinion!

spaghetticode:
I discussed with Elia via chat about the improvement (replace the method `ItemTotal::operators_map` with a class attribute)
and we agreed the downside (unexpected behavior when changing in subclasses) is more relevant than the advantages.

kennyadsl:
Works for me, can we just add some documentation/example on how people should use this?

jarednorman (approved):

elia:
> For what it's worth, the second commit contains the word "extendability", which should probably be ["extensibility"](https://en.wikipedia.org/wiki/Extensibility). The meaning is clear, so I don't care at all if you change it.

Fixed, thanks!

jarednorman (approved):

kennyadsl:
Thanks @elia !

PR #3442, opened 2019-11-28
Use RESTful routing for users' API key management

kennyadsl:
**Description**

This PR refactors how we are using routes and controller actions to handle users' API key. In order to use RESTful actions, this is moving the non-RESTful endpoints into a separate controller.

This changes the default routing structure in this way:

<table>
  <tr>
    <th colspan="2">Operation</th>
    <th>Rails Url Helper</th>
    <th>Request</th>
  </tr>
  <tr>
    <td rowspan="2">Create a key</td>
    <td>Before</td>
    <td><code>generate_api_key_admin_user_path</code></td>
    <td><code>PUT /admin/users/:id/generate_api_key</code></td>
  </tr>
  <tr>
    <td>After</td>
    <td><code>admin_user_api_key_path</code></td>
    <td><code>POST /admin/users/:user_id/api_key</code></td>
  </tr>
  <tr>
    <td rowspan="2">Destroy a key</td>
    <td>Before</td>
    <td><code>clear_api_key_admin_user_path</code></td>
    <td><code>PUT /admin/users/:id/clear_api_key</code></td>
  </tr>
  <tr>
    <td>After</td>
    <td><code>admin_user_api_key_path</code></td>
    <td><code>DELETE /admin/users/:user_id/api_key</code></td>
  </tr>
</table>

This has the benefit to have a more organized codebase, but also to improve the granularity of how we can set permissions. For example, now we can have a permission set that does not allow to change or revoke users' API key but allows management of other users things. For this reason, a new ability is required and has been added in the provided UserManagement permission set.

This PR also deprecates the old controller actions, printing a message in the log of users that are still using the old actions.

The controller tests for the old actions have been deprecated as well, and copied over the new controller specs file but it is only checking controller related stuff now, since the real functionality is already covered by unit and feature specs elsewhere.

**Checklist:**

### DIFF
--- a/backend/spec/controllers/spree/admin/users/api_key_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/users/api_key_controller_spec.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe Spree::Admin::Users::ApiKeyController, type: :controller do
+  let(:user) { create(:user) }
+
+  describe '#create' do
+    context "with ability to manage users and API keys" do
+      stub_authorization! do |_user|
+        can [:manage], Spree.user_class
+        can [:manage], :api_key
+      end
+
+      it "allows admins to create a new user's API key" do
+        post :create, params: { user_id: user.id }
+
+        expect(flash[:success]).to eq I18n.t('spree.admin.api.key_generated')
+        expect(response).to redirect_to(spree.edit_admin_user_path(user))
+      end
+    end
+
+    context "without ability to manage users and API keys" do
+      stub_authorization! do |_user|
+      end
+
+      it 'denies access' do
+        delete :destroy, params: { user_id: user.id }
+
+        expect(flash[:error]).to eq I18n.t('spree.authorization_failure')
+        expect(response).to redirect_to '/unauthorized'
+      end
+    end
+  end
+
+  describe '#destroy' do
+    context "with ability to manage users and API keys" do
+      stub_authorization! do |_user|
+        can [:manage], Spree.user_class
+        can [:manage], :api_key
+      end
+
+      it "allows admins to clear an existing user's API key" do
+        user.generate_spree_api_key!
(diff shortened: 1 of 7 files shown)


backend/spec/controllers/spree/admin/users/api_key_controller_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      it "allows admins to create a new user's API key" do
+        post :create, params: { user_id: user.id }
+
+        expect(flash[:success]).to eq I18n.t('spree.admin.api.key_generated')
+        expect(response).to redirect_to(spree.edit_admin_user_path(user))
+      end
+    end
+
+    context "without ability to manage users and API keys" do
+      stub_authorization! do |_user|
```
spaghetticode: 🤔 is there any reason for having the empty block here?

kennyadsl: Yes, it's weird but if no block is passed a full authorization will be granted, [see the method definition](https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/core/lib/spree/testing_support/authorization_helpers.rb#L10).

kennyadsl:
@spaghetticode For the integraton tests question, we already [have integration tests for this](https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/backend/spec/features/admin/users_spec.rb#L242-L285) and I'm satisfied since that they are green without the need to change them anyhow. If you think we should still add more specs, just let me know!

backend/spec/controllers/spree/admin/users/api_key_controller_spec.rb
spaghetticode: interesting, thank you for clarifying it to me!

spaghetticode:
@kennyadsl I agree those specs are all we need, thank you

PR #3446, opened 2019-12-03
Adds ability to remove attachments from taxons

softr8:
This takes advantage of the pluggable attachment definitions
and implements the paperclip version

This is another approach of #3366, original approach and source of inspiration

**Checklist:**

### DIFF
--- a/backend/config/routes.rb
+++ b/backend/config/routes.rb
@@ -132,7 +132,9 @@ Spree::Core::Engine.routes.draw do
       collection do
         post :update_positions
       end
-      resources :taxons
+      resources :taxons do
+        resource :attachment, controller: 'taxons/attachment', only: [:destroy]
+      end
     end
 
     resources :taxons, only: [:index, :show] do
--- a/backend/spec/features/admin/taxons_spec.rb
+++ b/backend/spec/features/admin/taxons_spec.rb
@@ -75,4 +75,16 @@ describe "Taxonomies and taxons", type: :feature do
       end
     end
   end
+
+  scenario "Removes attachments from taxon" do
+    taxon = create(:taxon)
+    taxon.update(icon: File.new(file_fixture('ror_ringer.jpeg')))
+
+    visit spree.edit_admin_taxonomy_taxon_path(taxon.taxonomy, taxon.id)
+    within('#taxon_icon_field') do
+      click_on 'Remove'
+    end
+
+    expect(page).to have_content("Icon has been successfully removed!")
+  end
 end
--- a/core/app/models/spree/taxon/paperclip_attachment.rb
+++ b/core/app/models/spree/taxon/paperclip_attachment.rb
@@ -22,4 +22,13 @@ module Spree::Taxon::PaperclipAttachment
   def attachment_partial_name
     'paperclip'
   end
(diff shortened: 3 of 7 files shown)


core/config/locales/en.yml
```diff
@@ -2114,6 +2114,7 @@ en:
       (i.e. if the tax rate is 5% then enter 0.05)
     tax_rates: Tax Rates
     taxon: Taxon
+    taxon_attachment_error: There was an error removing the attachment
```
EduardoGHdez: Don't forget the spanish version 🙌

backend/config/routes.rb
```diff
@@ -132,7 +132,9 @@
       collection do
         post :update_positions
       end
-      resources :taxons
+      resources :taxons do
+        resource :attachment, controller: '/spree/admin/taxons/attachment', only: [:destroy]
```
kennyadsl: I think you can just use `taxons/attachment` as the `controller` value.

softr8: You're right! Thanks! 👍

core/app/models/spree/taxon/paperclip_attachment.rb
```diff
@@ -22,4 +22,13 @@ def icon_present?
...
   end
+
+  def destroy_attachment(definition)
+    return false unless respond_to?(definition)
+
+    attached_file = send(definition)
+    return false unless attached_file
+
+    attached_file.destroy
+  end
```
spaghetticode: I think it may be worth writing a few model specs for this method. The integration test is great, but of course it covers only the happy path.

core/config/locales/en.yml
spaghetticode: what about a more descriptive `taxon_attachment_removal_error` or something similar?

backend/spec/features/admin/taxons_spec.rb
```diff
@@ -75,4 +75,16 @@ def only_one_selected_tab_inside?(sub_tab_selector, tab_name, tab_path)
       end
     end
   end
+
+  scenario "Removes attachments from taxon" do
+    taxon = create(:taxon)
+    taxon.update(icon: File.new(Rails.root.join('..', '..', 'spec', 'fixtures', 'files', 'ror_ringer.jpeg')))
```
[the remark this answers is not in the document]
softr8: didn't work :(
`No handler found for #<Pathname:spec/fixtures/files/ror_ringer.jpeg>`

core/app/models/spree/taxon/paperclip_attachment.rb
softr8: added!

backend/spec/features/admin/taxons_spec.rb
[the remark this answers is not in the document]
spaghetticode: sorry, my suggestion was not very clear. This will work:
```
taxon.update(icon: File.new(file_fixture("ror_ringer.jpeg")))
```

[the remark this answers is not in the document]
softr8: oh! of course... my bad, done!

Issue #3234, opened 2019-06-19
RFC: Remove firstname and lastname from adresses


tvdeyen:
## The problem

Names :)

Seriously, read https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

Also issues like https://github.com/solidusio/solidus_paypal_braintree/issues/226 happen

## The proposal

We should add a `name` column to `Spree::Address` then deprecate and later remove `firstname` and `lastname`.

ericsaupe:
I'm all for this. It simplifies a lot of the edge cases surrounding names and how to display them. I can see use cases where they want a bit more of a restricted naming experience a gem can be created to handle those cases.

jacobherrington:
I have nothing against this -- I think it's a great change.

kennyadsl:
I also in favor of this.

skukx:
Here Here

filippoliverani:
I really like this change and I started to work on it in #3447 but with the help of @kennyadsl I realized it was too much for a single PR.

I'm tring to split the work into four different steps: 
1. add `name` virtual attribute which still relies on `firstname` and `lastname` #3458 
2. use `name` in views e api responses instead of `lastname` and `firstname` if an ad hoc `show_address_deprecated_attributes` preference is set to `true`
3. add `name` column to Adress db table, add a rake data migration task and keep attributes in sync during updates
4. deprecate `show_address_deprecated_attributes` preference, `firstname`, `lastname` and `full_name`

What do you think?

kennyadsl:
At the current stage we have 3 options:

Option 1:

Add a way to do: `address.update_attributes(name:)`. Now it's not possible since `name` is not a column in the database. We can override how rails sets attributes and split `name` into two db fields. 

Option 2:

Add `name` to the database and provide a fallback when `name` is not set. The job to migrate can be performed async.

Option 3:

Add `name` to the database and provide a migration/rake task that combines all the old `firstname` and `lastname` into `name. Can take forever on old stores with a lot of records.

kennyadsl:
We discussed this during the core team, we'd go with option 2 (adding a fallback when `name` is not present) and creating a migration that has multiple options and each store can select the best one based on their needs. We did something similar [here](https://github.com/solidusio/solidus/blob/9864746a3b5d8493f990ef63b28e4989f6534673/core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb) that we can replicate.

The handlers we were thinking of are:
[...]

PR #3447, opened 2019-12-05
Add name to Spree::Address and deprecate firstname and lastname

filippoliverani:
**Description**
This PR proposes a first step to solve #3234.

The main idea is to add a new `name` attribute to `Spree::Address` and to deprecate `firstname`, `lastname` and `full_name` usage while keeping full backward compatibility during deprecation period.

1. I created a migration to add the new column and rake task to backfill and index the column to leave more control over data migration, it could be time consuming on a large db. The rake task should be mentioned in the changelog, e.g.:
    > Run`rake solidus:migrations:migrate_address_names:up` to migrate `Spree::Address` `firstnmame` and `lastname` data to the new `name` field
2. To favor backward compatibility during deprecation period, writes on `firstname` and `lastname` update name too. Writes on `name` keeps `firstname` and `lastname` updated splitting on the first space to have a fallback. Presence validation on `firstname` and `name` has been updated accordingly.
I updated `Spree::Address` to make lookup, merge and equality methods work with any combination of new and old name attributes. I added a `Spree::Name` value object to handle name parsing during deprecation period.
3. I added the preference flag `show_deprecated_attributes` to leave the possibility of viewing deprecated attributes in HTML views and API responses (and potentially anywhere else) and ease the way into the upgrade. The default value is `true` to minimize impact on existing Solidus installations, specs and new Solidus installations set the preference to `false`.
4. I introduced deprecation warnings on `firstname`, `lastname` and `full_name` and updated all specs to be more specific when checking expectaions on `Spree::Deprecation` mocks to avoid overlapping warnings. 
I also upgraded api docs to OpenAPI Specification 3 to be able to mark properties as deprecated in responses and inputs schemas.

Here are some screenshots of the updated default HTML views

Frontend address step*Frontend address step*

Backend orders search*Backend orders search*

Backend order customer details*Backend order customer details*

Backend user addresses*Backend user addresses*

**Checklist:**


core/app/models/spree/address.rb
```diff
@@ -11,7 +11,9 @@ class Address < Spree::Base
     belongs_to :country, class_name: "Spree::Country", optional: true
     belongs_to :state, class_name: "Spree::State", optional: true
 
-    validates :firstname, :address1, :city, :country_id, presence: true
+    validates :firstname, presence: true, if: -> { name.blank? || lastname? }
```
kennyadsl: Why checking `lastname` as well?

core/app/models/spree/address.rb
```diff
@@ -174,11 +173,25 @@ def country_iso=(iso)
...
+    end
+
+    def lastname=(value)
+      super
+      update_name
     end
 
     private
 
+    def update_name
```
kennyadsl: I'd use another name for this method. `update_name` makes me think that an update (save) will be performed while we are only setting the attribute here relying on something else for persisting it. Also, maybe we should set the name only if it is empty?

core/app/models/spree/address.rb
```diff
@@ -11,23 +11,24 @@ class Address < Spree::Base
...
     validate :validate_state_matches_country
 
     alias_attribute :first_name, :firstname
     alias_attribute :last_name, :lastname
     alias_attribute :full_name, :name
 
+    DEPRECATION = { firstname: :name, lastname: :name, full_name: :name }
     DB_ONLY_ATTRS = %w(id updated_at created_at)
     TAXATION_ATTRS = %w(state_id country_id zipcode)
+    DEPRECATED_ATTRS = DEPRECATION.stringify_keys.keys
```
kennyadsl: Why `stringify_keys` is needed here?

core/spec/models/spree/address_spec.rb
```diff
@@ -23,6 +23,33 @@
...
       expect(address.name).to eq('Ryan Small')
     end
+
+    it 'updates firstname and lastname when updating name' do
+      address.name = 'Jane Von Doe'
+
+      expect(address.firstname).to eq('Jane')
+      expect(address.lastname).to eq('Von Doe')
+    end
+  end
```
kennyadsl: Can we add a test when both `firstname`, `lastname` and `name` are tried to be set in the same instance? I know it should not happen but I'd like to anticipate a default behavior for people messing up this thing in their own applications.

api/spec/requests/spree/api/checkouts_controller_spec.rb
```diff
@@ -353,7 +353,10 @@ module Spree
       end
 
       it "can apply a coupon code to an order" do
+        allow(Spree::Deprecation).to receive(:warn)
```
kennyadsl: Why this `allow` is needed here? Shouldn't `expect` provide both stub and expectation?

core/lib/spree/app_configuration.rb
```diff
@@ -204,6 +204,11 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Boolean] Request instructions/info for shipping (default: +false+)
     preference :shipping_instructions, :boolean, default: false
 
+    # @!attribute [rw] show_deprecated_attributes
+    #   @return [Boolean] Show deprecated model attributes in HTML views and
+    #   API responses. (default: +true+)
+    preference :show_deprecated_attributes, :boolean, default: true
```
kennyadsl: What about `show_address_deprecated_attributes` or `show_deprecated_naming_attributes` instead? I feel like this is too generic.

backend/app/views/spree/admin/orders/index.html.erb
```diff
@@ -65,14 +64,21 @@
...
+                  <%= label_tag :q_bill_address_firstname_start, t('spree.first_name_begins_with') %>
+                  <%= f.text_field :bill_address_firstname_start, size: 25 %>
+                </div>
+                <div class="field">
+                  <%= label_tag :q_bill_address_lastname_start, t('spree.last_name_begins_with') %>
+                  <%= f.text_field :bill_address_lastname_start, size: 25 %>
+                </div>
+              <% else %>
+                <div class="field">
+                  <%= label_tag :q_bill_address_name_start, t('spree.name_begins_with') %>
```
kennyadsl: We are removing the ability to search by the initial part of the lastname here right? 😱

backend/app/views/spree/admin/search/users.json.jbuilder
```diff
@@ -4,7 +4,13 @@ json.array!(@users) do |user|
   json.id user.id
   json.email user.email
 
-  address_fields = [:firstname, :lastname, :address1, :address2, :city, :zipcode, :phone, :state_name, :state_id, :country_id, :company]
+  address_fields = [
+    :name, :address1, :address2, :city, :zipcode, :phone, :state_name,
+    :state_id, :country_id, :company
+  ]
```
kennyadsl: I'd take advantage of this change to move one attribute per line here, so we won't change git history for all of them if only one changes in the future.

api/openapi/main.hub.yml
```diff
@@ -62,4 +62,4 @@ pages:
           route:
             path: /api-reference
           data:
-            $ref: ./api.oas2.yml
+            $ref: ./solidus.yml
```
kennyadsl: I'd split this commit into a preparative one that move the yaml to the new format (maybe also a separate PR) and the actual change so that it is well documented.

api/app/helpers/spree/api/api_helpers.rb
```diff
@@ -121,12 +121,18 @@ def required_fields_for(model)
         :id, :number, :state, :order_id, :memo, :created_at, :updated_at
       ]
 
-      @@address_attributes = [
-        :id, :firstname, :lastname, :full_name, :address1, :address2, :city,
-        :zipcode, :phone, :company, :alternative_phone, :country_id, :country_iso,
-        :state_id, :state_name, :state_text
+      @@non_deprecated_address_attributes = [
```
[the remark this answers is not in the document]
filippoliverani: ApiHelpers explicitly lists all attributes to show in jbuilder templates, I needed the current list and the list including the deprecated ones

core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -40,7 +40,7 @@ def self.setup(gem_root:, lib_name:, auto_migrate: true)
     ENV["LIB_NAME"] = lib_name
     DummyApp::Application.config.root = File.join(gem_root, 'spec', 'dummy')
 
-    DummyApp::Application.initialize!
+    DummyApp::Application.initialize! unless DummyApp::Application.initialized?
```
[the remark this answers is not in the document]
filippoliverani: When something goes wrong during ActiveRecord connection initialization, Rails workflows marks Application initialized before reaching that line and when it executes it returns an `Application has been already initialized` error swallowing the original db connection error making it harder to debug the problem.
I'll move it to its own commit to explain this 😛

api/app/helpers/spree/api/api_helpers.rb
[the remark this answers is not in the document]
kennyadsl: I mean that we can get the list of the deprecated attributes by removing the deprecated ones from the current list. Just because we'll have less things to change if we add other attributes to address in the future.

core/spec/lib/tasks/migrations/migrate_address_names_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+path = Spree::Core::Engine.root.join('lib/tasks/migrations/migrate_address_names.rake')
+
+RSpec.describe 'solidus:migrations:migrate_address_names' do
+  describe 'up', unless: :mysql? do
```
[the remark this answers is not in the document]
filippoliverani: I wasn't able to test a schema changing operation on MySQL without breaking all subsequent specs run in the same suite. If the spec is run on its own it works without any problem but if it's run with other specs it completly breaks MySQL transaction handling. I tried to run it in every way: without transaction, with manual setup and cleaning and with a dedicated transaction but the result is always the same.
I resorted to skipping the spec on MySQL 🏳️.

sample/db/samples/addresses.rb
```diff
@@ -27,9 +27,13 @@
...
 2.times do
+  first_name = first_names.sample
+  last_name = last_names.sample
+  name = "#{first_name} #{last_name}"
   Spree::Address.create!(
-    firstname: first_names.sample,
-    lastname: last_names.sample,
+    name: name,
+    firstname: first_name,
+    lastname: last_name,
```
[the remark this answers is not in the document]
filippoliverani: Yes, should already be like that in the last version of that commit https://github.com/solidusio/solidus/pull/3447/commits/93db0be1d74d5666145d6f0e7f6a052a98c06e28#diff-a7dcd0d0a3b5ed0283a4ab06d77c43c4R29

core/app/models/spree/address.rb
filippoliverani: The latest version of the validation is lightly different: https://github.com/solidusio/solidus/pull/3447/commits/93db0be1d74d5666145d6f0e7f6a052a98c06e28#diff-14b461f472ff6adbfd491ed5674e4e3aR252.
We need to check `lastname` too to avoid a corner case in which `name` is blank, `lastname` is present and someone tries to set `firstname` to blank.

core/app/models/spree/address.rb
filippoliverani: In a following commit I added that check https://github.com/solidusio/solidus/pull/3447/commits/93db0be1d74d5666145d6f0e7f6a052a98c06e28#diff-14b461f472ff6adbfd491ed5674e4e3aR203.
I agree, the `update_name` could be misleading, maybe `fill_name` would be better 🤔 ?

core/app/models/spree/address.rb
filippoliverani: To be consistent with `SpreeAddress.value_attributes` and its callers, they work only with string hash keys https://github.com/solidusio/solidus/pull/3447/files/16fc264d3acc1ff743c8aac33b5cc93dd47bf737..93db0be1d74d5666145d6f0e7f6a052a98c06e28#diff-14b461f472ff6adbfd491ed5674e4e3aR65

core/spec/models/spree/address_spec.rb
filippoliverani: Yes, absolutely 👍

api/spec/requests/spree/api/checkouts_controller_spec.rb
filippoliverani: This should be a partial mock, it can recevive warnings about anything (`firstname` usage included) but verify the expectation only on the `CouponCodesController#create` deprecation its checking.

core/lib/spree/app_configuration.rb
filippoliverani: Yes I wasn't sure about this, I'd go with  `show_address_deprecated_attributes`

backend/app/views/spree/admin/orders/index.html.erb
filippoliverani: Yes 😢 , I was thinking of switching the `starts with` searches to `contains` searches but it could have a big impact on db load and performances. The `LIKE` predicate used in `starts with` could apply the standard index,  but if we swtich to `contains` we would need a full text index. Full text indexes are different on many DBMS and I'm not sure if we can expect that they would be correctly configured.
Is there another way?

mdesantis:
Hi!! We're going to love this change!

I have one concern:

> To favor backward compatibility during deprecation period, writes on firstname and lastname update name too. Writes on name keeps firstname and lastname updated splitting on the first space to have a fallback.

What about names containing spaces? Such as "Pier Giorgio"? Are we just going to ignore them because they should be uncommon?

api/openapi/main.hub.yml
filippoliverani: I slipped it in here to make the deprecation self consistent but it is an important change that needs its own PR 👍

backend/app/views/spree/admin/orders/index.html.erb
filippoliverani: Yes 😢 , I was thinking of switching the starts with searches to contains searches but it could have a big impact on db load and performances. The LIKE predicate used in starts with could apply the standard index, but if we swtich to contains we would need a full text index. Full text indexes are different on many DBMS and I'm not sure if we can expect that they would be correctly configured.
Is there another way?

filippoliverani:
> Hi!! We're going to love this change!
> 
> I have one concern:
> 
> > To favor backward compatibility during deprecation period, writes on firstname and lastname update name too. Writes on name keeps firstname and lastname updated splitting on the first space to have a fallback.
> 
> What about names containing spaces? Such as "Pier Giorgio"? Are we just going to ignore them because they should be uncommon?

The code uses full `name` everywhere so all the spaces are always kept and shown in views and apis. The backfill of `firstname` and `lastname` when writing on `name` is a "just in case" fallback, it is not accurate and it should not be used. During deprecation period if an app writes on the deprecated fields I expect it to read from them too, do you think that someone could mix and match writes and reads from new and deprecated fileds?

api/app/helpers/spree/api/api_helpers.rb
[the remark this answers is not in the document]
filippoliverani: Sorry I misunderstood, yes that would be better 👍

core/spec/lib/tasks/migrations/migrate_address_names_spec.rb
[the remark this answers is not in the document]
kennyadsl: We discussed this IRL and we should try to use [`reset_column_information`](https://apidock.com/rails/ActiveRecord/ModelSchema/ClassMethods/reset_column_information) here to make it work with MySQL as well.

sample/db/samples/addresses.rb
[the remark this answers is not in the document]
kennyadsl: You are right, I reviewed commit by commit and I should have to come back and remove this comment.

core/app/models/spree/address.rb
kennyadsl: Isn't that method already calling `stringify_keys` on the first line?

```ruby
      base = base_attributes.stringify_keys.merge(merge_attributes.stringify_keys)
```

filippoliverani: Yes but it is used in `except` on string keyed hashes in in `self.factory` and `==`

core/app/models/spree/address.rb
kennyadsl: Ah, makes sense!

core/app/models/spree/address.rb
kennyadsl: You are right, I thought it was passed to the method as an argument.

mdesantis:
@filippoliverani no, it makes sense :+1:

core/spec/models/spree/address_spec.rb
filippoliverani: Having a default here is tricky. The ways I can think of to update the three attributes in a single method call are AR methods taking an hash in input like  `attributes=`, `update` and `update_columns`.
`attributes=` and `update` use model accessors and run trough the input hash in sequence so the final outcome will be different depending on the order of the elements in the hash. 
`update_columns` skips accessors and could write non coherent `name`, `firstname` and `lastname` values.
I think that overriding all AR setters to handle corner cases will be overkill and unmaintainable, another solution could be to use `after_save` hooks but I don't like it either. I'm more inclined to leaving it this way, WDYT?

kennyadsl: what about removing `firstname` and `lastname` from attributes if `name` is also in there? Is it somehow possible at a model level?

filippoliverani: We could decorate `ActiveModel::ForbiddenAttributesProtection#sanitize_for_mass_assignment` and reject them if `name` is present. `sanitize_for_mass_assignment` is called during every mass assignment, it should work everywhere and it's not too invasive.

filippoliverani:
I realized that this PR is too big to handle, I'll close it and try to break it into smaller ones, thanks for the feedback and the helpful reviews received so far.

Issue #3448, opened 2019-12-05
TaxRate included_in_price has two conflicting meanings


pedantic-git:
**Solidus Version**: 2.10.0.beta1

**To Reproduce**

1. Create a tax rate with `included_in_price` set to true and non-zero rate.
2. Generate a sale or use the admin to create VAT prices.

**Current behavior**

* When `included_in_price` is true, the VAT is added to the gross price so the net price is always the same, but the amount the customer pays varies according to the tax rate in their locale.
* When `included_in_price` is false, the `Spree::Price` object is never created by the VatPriceGenerator so the user is charged the same price for every purchase. **However, the user is no longer shown the VAT at checkout.**

**Expected behavior**

According to the documentation at https://guides.solidus.io/developers/taxation/value-added-tax.html, these two should be the other way round.

The expected behaviour as far as I can tell from the documentation:

* When `included_in_price` is true, the VatPriceGenerator **does not** generate Spree::Price objects for different locales, so that users pay the same gross price (and different net prices per locale). Or, it generates Spree::Price objects but they all have the same gross price.
* When `included_in_price` is false, the VatPriceGenerator **does** generate Spree::Price objects for different locales, so that users pay different gross prices (and the same net price) in different locales.
* VAT should be shown to users subjected to a TaxRate at checkout whether `included_in_price` is true or not.

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Chrome
 - Version: 79

**Additional context**

I think the main offending line is [here](https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/core/app/models/spree/variant/vat_price_generator.rb#L59) - it should be `.where(included_in_price: false)`.

But I appreciate many stores will have got things back-to-front now so changing this is tricky!

Thanks in advance for your help!!

pedantic-git:
Thinking about this further, I wonder if the simplest solution is to add a second flag to TaxRate called something like `keep_gross_price_the_same`. If that flag is set to true, the `VatPriceGenerator` knows to create new locale `Spree::Price` objects with the same value (but different net amounts). That way it's backwards compatible and gives people the option of whether they want the gross or net price to stay the same between locales. The documentation will need updating too.

Let me know if you think this is a good idea and I can have a go at implementing it myself.

aldesantis:
[...]

PR #3451, opened 2019-12-10
Allow configuring VAT Price Generator class

kennyadsl:
**Description**

The current VAT generation logic provided by Solidus is not flexible enough to match any business needs, see #3448.

This PR allows to easily create your own custom class to generate VAT prices and use it in Solidus. It also adds a simple explanation in Guides with a realistic example, taken from the issue mentioned above.

**Checklist:**

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -439,7 +439,7 @@ module Spree
     end
 
     def build_vat_prices
-      VatPriceGenerator.new(self).run
+      Spree::Config.variant_vat_prices_generator_class.new(self).run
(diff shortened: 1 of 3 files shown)


kennyadsl:
That a great idea. Wondering if this should be done at a TaxRate level. I mean when we decide that a tax rate is incuded in price we could also give admin the ability to choose if they want it constant net or constant gross. Do you think it makes sense this way?

kennyadsl:
Opened https://github.com/solidusio/solidus/issues/3461 to track the @mamhoff suggestion. Thanks!

PR #3457, opened 2019-12-13
Standardize spree event context parameter

skukx:
When calling `Spree::Event.fire 'event'`. It is possible to pass in
parameters. The parameter is usually the context to the event such as an
order in order_finalized or reimbursement when a reimbursement has been
created.

In order to standardize this context rather than calling
`Spree::Event.fire 'order_finalized', order: @order`, call
`Spree::Event.fire 'order_finalized', context: @order`.

This will also allow subscribers to always know which key will contain
the context to the event.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -435,7 +435,7 @@ module Spree
 
       touch :completed_at
 
-      Spree::Event.fire 'order_finalized', order: self
+      Spree::Event.fire 'order_finalized', context: self
     end
 
     def fulfill!
--- a/core/app/models/spree/reimbursement.rb
+++ b/core/app/models/spree/reimbursement.rb
@@ -103,11 +103,11 @@ module Spree
 
       if unpaid_amount_within_tolerance?
         reimbursed!
-        Spree::Event.fire 'reimbursement_reimbursed', reimbursement: self
+        Spree::Event.fire 'reimbursement_reimbursed', context: self
         reimbursement_success_hooks.each { |hook| hook.call self }
       else
         errored!
-        Spree::Event.fire 'reimbursement_errored', reimbursement: self
+        Spree::Event.fire 'reimbursement_errored', context: self
         reimbursement_failure_hooks.each { |hook| hook.call self }
       end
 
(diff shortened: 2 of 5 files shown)


skukx:
Here is a use case for logging:

```ruby
Spree::Event.module_eval do
  LOG_PATH = Rails.root.join('log', 'spree_event.log')

  module FireMethodWrapper
    ##
    # Spree::Event extends self so we need to use self
    # on the override
    #
    def fire(event_name, opts = {})
      context = opts[:context]

      event = {}
      event[:name] = event_name
      if context.respond_to?(:id)
        event[:context_type] = context.class.name
        event[:context_id] = context.id
      end

      log = { spree: { event: event } }
      logger.info(log.to_json)

      super
    end
  end

  singleton_class.send(:prepend, FireMethodWrapper)

  def logger
    return @logger if @logger.present?
    @logger = ActiveSupport::TaggedLogging.new(Logger.new(LOG_PATH))
    @logger.level = Logger.const_get(Rails.application.config.log_level.to_s.upcase)
    @logger
  end
end
```

kennyadsl:
I'm ok with this change, just thinking how to deal if some events have more than one object in their context. Do you think it's something that could happen?

skukx:
As far as I can think, the events are usually tied to a single context. There is other metadata that can be passed in but as far as the context goes, I think it is one to one.

aldesantis:
I have a few concerns about this:

1. I feel like we may be standardizing things a bit too early here. We still don't know if all events are only going to have one context, since we're not using `Spree::Event` extensively yet.
2. Renaming `order` (and similar) to `context` removes a lot of information about the interface the context will respond to, so it makes it a (tiny) bit harder to understand what I can do with that object.
3. I would treat this as a breaking change. Extensions and user code that subscribes to these events will not work anymore if we change the name of the key, although solution to this would be to simply continue passing the old key as well (or is there such a thing as a `DeprecatedHashProxy`?).

Overall, I feel like the cons here far outweigh the pros. For instance, if I want to implement your example with the current interface, I can simply `#inspect` the event's payload, or if I want to be more concise I can just print the class and ID of each AR object and `#inspect` everything else.

spaghetticode:
@skukx thanks for proposing this change. 

I have mixed feelings on it, one issue is this is a breaking change (already pointed out by @aldesantis ), but I'm more concerned by the fact that in this PR there's nothing that enforces the use of `context` at the code level, so `context` would be just a bland convention, easily forgettable.
Even when this PR gets merged, I would not trust any code like the `FireMethodWrapper` you added as an example to work reliably in the future.

PR #3458, opened 2019-12-16
Add name to Spree::Address

filippoliverani:
**Description**
This PR proposes a first step to solve #3234 .

The main idea is to add a new `name` attribute to `Spree::Address` and to deprecate `firstname`, `lastname` and `full_name` usage while keeping full backward compatibility.

This PR adds a virtual attribute `name` that replaces `full_name` and favors reading form it when accessing Address data. It uses a `Spree::Address::Name` value object to handle name parsing during deprecation period.

**Checklist:**

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -20,6 +20,7 @@ module Spree
 
     alias_attribute :first_name, :firstname
     alias_attribute :last_name, :lastname
+    alias_attribute :full_name, :name
 
     DB_ONLY_ATTRS = %w(id updated_at created_at)
     TAXATION_ATTRS = %w(state_id country_id zipcode)
@@ -62,10 +63,13 @@ module Spree
     def self.value_attributes(base_attributes, merge_attributes = {})
       base = base_attributes.stringify_keys.merge(merge_attributes.stringify_keys)
 
-      # TODO: Deprecate these aliased attributes
-      base['firstname'] = base['first_name'] if base.key?('first_name')
-      base['lastname'] = base['last_name'] if base.key?('last_name')
-
+      name_from_attributes = Spree::Address::Name.from_attributes(base)
+      if base['firstname'].presence || base['first_name'].presence
+        base['firstname'] = name_from_attributes.first_name
+      end
+      if base['lastname'].presence || base['last_name'].presence
+        base['lastname'] = name_from_attributes.last_name
+      end
       excluded_attributes = DB_ONLY_ATTRS + %w(first_name last_name)
 
       base.except(*excluded_attributes)
@@ -80,18 +84,13 @@ module Spree
       self.class.value_attributes(attributes.slice(*TAXATION_ATTRS))
     end
 
-    # @return [String] the full name on this address
-    def full_name
-      "#{firstname} #{lastname}".strip
-    end
-
     # @return [String] a string representation of this state
     def state_text
       state.try(:abbr) || state.try(:name) || state_name
     end
 
     def to_s
-      "#{full_name}: #{address1}"
+      "#{name}: #{address1}"
     end
 
     # @note This compares the addresses based on only the fields that make up
@@ -130,7 +129,7 @@ module Spree
     # @return [Hash] an ActiveMerchant compatible address hash
     def active_merchant_hash
       {
-        name: full_name,
+        name: name,
         address1: address1,
         address2: address2,
         city: city,
@@ -170,6 +169,19 @@ module Spree
 
     def country_iso
       country && country.iso
+    end
+
+    # @return [String] the full name on this address
+    def name
+      Spree::Address::Name.new(firstname, lastname).value
+    end
+
+    def name=(value)
+      return if value.nil?
+
+      name_from_value = Spree::Address::Name.new(value)
+      write_attribute(:firstname, name_from_value.first_name)
+      write_attribute(:lastname, name_from_value.last_name)
     end
 
     private
--- a/core/app/models/spree/address/name.rb
+++ b/core/app/models/spree/address/name.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module Spree
+  class Address
+    # Provides a value object to help transitioning from legacy
+    # firstname and lastname fields to a unified name field.
+    class Name
+      attr_reader :first_name, :last_name, :value
+
+      # Creates an instance of Spree::Address::Name parsing input attributes.
+      # @param attributes [Hash] an hash possibly containing name-related
+      #   attributes (name, firstname, lastname, first_name, last_name)
+      # @return [Spree::Address::Name] the object created
+      def self.from_attributes(attributes)
+        params = attributes.with_indifferent_access
+
+        if params[:name].present?
+          Spree::Address::Name.new(params[:name])
+        elsif params[:firstname].present?
+          Spree::Address::Name.new(params[:firstname], params[:lastname])
+        elsif params[:first_name].present?
+          Spree::Address::Name.new(params[:first_name], params[:last_name])
+        else
+          Spree::Address::Name.new
+        end
+      end
+
+      def initialize(*components)
+        @value = components.join(' ').strip
+        initialize_name_components(components)
+      end
+
+      def to_s
+        @value
+      end
+
+      private
+
+      def initialize_name_components(components)
+        if components.size == 2
+          @first_name = components[0].to_s
+          @last_name = components[1].to_s
+        else
+          @first_name, @last_name = @value.split(/[[:space:]]/, 2)
+        end
+      end
+    end
+  end
+end
--- a/core/app/models/spree/credit_card.rb
+++ b/core/app/models/spree/credit_card.rb
@@ -163,7 +163,7 @@ module Spree
     #   the object to ActiveMerchant.
     # @return [String] the first name on this credit card
     def first_name
-      name.to_s.split(/[[:space:]]/, 2)[0]
+      Spree::Address::Name.new(name).first_name
     end
 
     # @note ActiveMerchant needs first_name/last_name because we pass it a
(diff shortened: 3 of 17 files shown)


core/app/models/spree/address.rb
```diff
@@ -177,6 +168,19 @@ def country_iso
...
 
+    # @return [String] the full name on this address
+    def name
+      Spree::Name.new(firstname, lastname).full_name
+    end
+
+    def name=(value)
+      return if value.nil?
+
+      current_name = Spree::Name.new(value)
```
kennyadsl: `current_name`: we can probably find a better variable name here. I feel like `current` has not a contextual meaning but I can't find a better one. 🤔

core/spec/lib/spree/name_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    it 'prioritizes name over firstname' do
+      attributes = {
+        name: 'Jane Doe',
+        firstname: 'Joe',
+        lastname: 'Bloggs'
+      }
+      name = described_class.from_attributes(attributes)
+
+      expect(name.first_name).to eq('Jane')
```
kennyadsl: Just for documentation, I'd check the `last_name` value as well in these examples.

core/app/models/spree/address.rb
```diff
@@ -58,22 +58,18 @@ def self.immutable_merge(existing_address, new_attributes)
...
-      base = base_attributes.dup
-
-      base.stringify_keys!
-
-      if merge_attributes
-        base.merge!(merge_attributes.stringify_keys)
+    def self.value_attributes(base_attributes, merge_attributes = {})
+      base = base_attributes.stringify_keys.merge(merge_attributes.stringify_keys)
+      name_from_attributes = Spree::Name.from_attributes(base)
+      if base.dig('firstname') || base.dig('first_name')
```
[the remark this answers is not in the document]
filippoliverani: Because we're checking the presence of the keys and the presence of the values of those keys

core/app/models/spree/address.rb
```diff
@@ -177,6 +168,19 @@ def country_iso
       country && country.iso
     end
 
+    # @return [String] the full name on this address
+    def name
+      Spree::Name.new(firstname, lastname).full_name
+    end
+
+    def name=(value)
+      return if value.nil?
```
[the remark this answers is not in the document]
filippoliverani: I think that setting `firstname` and `lastname` to an empty string could make sense even if the model would be invalid. That guard condition was meant to avoid setting them to an empty string when `name` was forced to `nil` by a mass assignment with default values.

core/app/models/spree/address.rb
filippoliverani: It's not perfect but I can't come up we something better either

core/spec/lib/spree/name_spec.rb
filippoliverani: Yes good idea 👍

core/app/models/spree/address.rb
```diff
@@ -58,22 +58,18 @@ def self.immutable_merge(existing_address, new_attributes)
       end
     end
 
-    # @return [Hash] hash of attributes contributing to value equality with optional merge
-    def self.value_attributes(base_attributes, merge_attributes = nil)
-      # dup because we may modify firstname/lastname.
-      base = base_attributes.dup
```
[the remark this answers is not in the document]
filippoliverani: We're not modifing in place `base` hash anymore, `base.except(*excluded_attributes)` is without side-effects

core/app/models/spree/address.rb
```diff
@@ -58,22 +58,18 @@ def self.immutable_merge(existing_address, new_attributes)
       end
     end
 
-    # @return [Hash] hash of attributes contributing to value equality with optional merge
```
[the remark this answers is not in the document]
filippoliverani: I think it was a braino, I didn't want to remove it 😱

filippoliverani:
@kennyadsl Thanks for the feedback!
> Also, I've found this [code into ActiveMerchant](https://github.com/activemerchant/active_merchant/blob/6e115b52721fb3320eba87994a45de61dccf4bf7/lib/active_merchant/billing/credit_card.rb#L250-L254). Curious how they do the opposite to split the full name into different parts, do you prefer this way for a specific reason?
> 
I just extracted the code used in [`Spree::CreditCard#first_name`](https://github.com/solidusio/solidus/blob/4c076f34936b2e821c750a26a7cc63c9319d3a63/core/app/models/spree/credit_card.rb#L166) to avoid reinventing the wheel, no preferences about one style or the other.

> Still not sure about the Rubocop fixes merged in the PR. I'd love to keep them but these PRs are very critical and I'd focus their content on the changes only.
>
They seemed small enough to me but I can keep only the ones enforced by Hound to reduce changes.

core/app/models/spree/address.rb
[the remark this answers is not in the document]
tvdeyen: Nice usage of `dig` for that purpose :ok_hand: 
Potentially we still could have an empty string in the database, right? So probably we still need to check for `.presence` here as well?

core/lib/spree/name.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'active_support/core_ext/hash'
+
+module Spree
```
kennyadsl: A last-minute additional comment: what about naming this value object `Spree::Address::Name`? I think it less generic since `Name` could be a lot of things.

core/app/models/spree/address.rb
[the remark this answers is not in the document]
filippoliverani: Right! even better 💪

core/app/models/spree/address.rb
filippoliverani: `name_from_value` could be the best one 😉

core/lib/spree/name.rb
filippoliverani: Absolutely, it's safer with an additional namespace

core/app/models/spree/address.rb
```diff
@@ -7,7 +7,6 @@ module Spree
   #
   class Address < Spree::Base
     extend ActiveModel::ForbiddenAttributesProtection
-
```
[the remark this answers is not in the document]
filippoliverani: Me too, I think it was an auto-format gone band 👍

core/lib/spree/name.rb
```diff
@@ -0,0 +1,44 @@
...
+      elsif params[:firstname].present?
+        Spree::Name.new(params[:firstname], params[:lastname])
+      elsif params[:first_name].present?
+        Spree::Name.new(params[:first_name], params[:last_name])
+      else
+        Spree::Name.new
+      end
+    end
+
+    def initialize(*components)
```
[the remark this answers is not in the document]
filippoliverani: Should be nicer extracted to a private method 👍

core/app/models/spree/order.rb
```diff
@@ -31,8 +31,9 @@ class CannotRebuildShipments < StandardError; end
 
     extend Spree::DisplayMoney
     money_methods :outstanding_balance, :item_total, :adjustment_total,
-      :included_tax_total, :additional_tax_total, :tax_total,
-      :shipment_total, :total, :order_total_after_store_credit, :total_available_store_credit
+                  :included_tax_total, :additional_tax_total, :tax_total,
```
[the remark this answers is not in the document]
filippoliverani: Yes @kennyadsl made the same suggestion, I will go easier on code linting

filippoliverani:
@spaghetticode Thank you for your feedback!

core/app/models/spree/address.rb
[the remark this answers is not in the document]
filippoliverani: To separate and clarify the refactoring I moved it to #3465

core/app/models/spree/address/name.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module Spree
+  class Address
+    # Provides a value object to help transitioning from legacy
+    # firstname and lastname fields to a unified name field.
+    class Name
+      attr_reader :first_name, :last_name, :full_name
```
tvdeyen: I think the attribute should be just `name`, since `full_name` is a legacy name. If `Spree::Address::Name.new.name` seems odd (because of the duplication) we could name it `value`.

core/app/models/spree/credit_card.rb
```diff
@@ -163,7 +163,7 @@ def has_payment_profile?
     #   the object to ActiveMerchant.
     # @return [String] the first name on this credit card
     def first_name
-      name.to_s.split(/[[:space:]]/, 2)[0]
+      Spree::Address::Name.new(name).first_name
```
tvdeyen: We should deprecate this method and move the logic into `solidus_gateway`, respectively into the payment method extensions. Having ActiveMerchant related logic in core seems odd.

core/app/models/spree/credit_card.rb
```diff
@@ -172,7 +172,7 @@ def first_name
     #   the object to ActiveMerchant.
     # @return [String] the last name on this credit card
     def last_name
-      name.to_s.split(/[[:space:]]/, 2)[1]
+      Spree::Address::Name.new(name).last_name
```
tvdeyen: Ditto deprecate this method.

core/app/models/spree/address/name.rb
kennyadsl: Agree, `value` is a good alternative.

core/app/models/spree/credit_card.rb
kennyadsl: I think this could be part of another PR since there are also other (unrelated) methods that we should deprecate under the same motivation. I'd say to keep this as it's minimum since it's already an important/critical change.

filippoliverani: Yes, I was thinking to split the work into 4-5 PRs to ease the review work, this is a proposed roadmap https://github.com/solidusio/solidus/issues/3234#issuecomment-566138962

core/app/models/spree/address/name.rb
filippoliverani: 👍

Issue #3450, opened 2019-12-09
Explore using ApplicationRecord instead of ActiveRecord::Base as models parent class


kennyadsl:
Rails 5 introduced the `ApplicationRecord` concept and started suggesting using it instead of inheriting from `ActiveRecord::Base` directly. 

This is to allow developers to apply some configurations to all the application models without the need to monkey-patch `ActiveRecord::Base`, since `ApplicationRecord` should be present in the host application, and it is added by the Rails generator during the installation (after Rails 5, so stores that started from Rails 4 and upgraded may not have it yet). 

If we want to be aligned with what Rails does, we should start relying on it, being sure our host applications have the `ApplicationRecord` file in place, in order to avoid breaking them. 

*References*

- https://github.com/rails/rails/pull/22567
- https://github.com/solidusio/solidus/blob/6aa743680e17287e177893fb203a8dbc9bc40803/core/app/models/spree/base.rb#L3

tvdeyen:
I like that 👍

jarednorman:
While we didn't do _exactly_ what this ticket outlines, I think with #3476 we can close this.

PR #3476, opened 2020-01-10
Use ApplicationRecord as models base class

filippoliverani:
**Description**
Ref #3450 

This PR introduces `Spree::ApplicationRecord` class and makes every ActiveRecord model previously inheriting from `ActiveRecord::Base` inherit from it. 
To leverage the new hierarchy, an existing `Spree::Core::Permalinks` monkey patch on `ActiveRecord::Base` has been moved to `Spree::ApplicationRecord`.

This way Solidus is more compliant with Rails 5+ recommendations and the risk of unwanted side effects on `ActiveRecord::Base` is mitigated.

**Checklist:**

### DIFF
--- a/core/app/models/spree/base.rb
+++ b/core/app/models/spree/base.rb
@@ -2,6 +2,7 @@
 
 class Spree::Base < ActiveRecord::Base
   include Spree::Preferences::Preferable
+  include Spree::Core::Permalinks
   serialize :preferences, Hash
 
   include Spree::RansackableAttributes
--- a/core/app/models/spree/promotion_code_batch.rb
+++ b/core/app/models/spree/promotion_code_batch.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Spree
-  class PromotionCodeBatch < ActiveRecord::Base
+  class PromotionCodeBatch < Spree::Base
     class CantProcessStartedBatch < StandardError
     end
 
--- a/core/app/models/spree/promotion_rule_role.rb
+++ b/core/app/models/spree/promotion_rule_role.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Spree
-  class PromotionRuleRole < ActiveRecord::Base
+  class PromotionRuleRole < Spree::Base
     belongs_to :promotion_rule, class_name: 'Spree::PromotionRule', optional: true
     belongs_to :role, class_name: 'Spree::Role', optional: true
   end
--- a/core/app/models/spree/shipping_rate_tax.rb
+++ b/core/app/models/spree/shipping_rate_tax.rb
@@ -6,7 +6,7 @@ module Spree
   # @attr [Spree::TaxRate] tax_rate The tax rate used to calculate the tax amount
   # @since 1.3.0
   # @see Spree::Tax::ShippingRateTaxer
-  class ShippingRateTax < ActiveRecord::Base
+  class ShippingRateTax < Spree::Base
     belongs_to :shipping_rate, class_name: "Spree::ShippingRate", optional: true
     belongs_to :tax_rate, class_name: "Spree::TaxRate", optional: true
 
--- a/core/db/default/spree/countries.rb
+++ b/core/db/default/spree/countries.rb
@@ -5,7 +5,7 @@ require 'carmen'
 # Insert Countries into the spree_countries table, checking to ensure that no
 # duplicates are created, using as few SQL statements as possible (2)
 
-connection = ApplicationRecord.connection
+connection = Spree::Base.connection
 
 country_mapper = ->(country) do
   name            = connection.quote country.name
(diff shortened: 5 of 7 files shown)


core/app/models/spree/application_record.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Spree::ApplicationRecord < ActiveRecord::Base
```
tvdeyen: We might want to take into account, that the app itself has an `ApplicationRecord` as well?

```suggestion
base_class = defined?(ApplicationRecord) ? ApplicationRecord : ActiveRecord::Base
class Spree::ApplicationRecord < base_class
```

core/app/helpers/spree/base_helper.rb
```diff
@@ -28,7 +28,7 @@ def meta_data
       object = instance_variable_get('@' + controller_name.singularize)
       meta = {}
 
-      if object.is_a? ActiveRecord::Base
+      if object.is_a? Spree::ApplicationRecord
```
tvdeyen: I think it is fine to keep it as is. Maybe the store has a custom model that also is an `ActiveRecord` and implements `meta_description`?

core/app/models/spree/application_record.rb
kennyadsl: Agree, but it could also happen that the app has an `ApplicationRecord` that should not be shared with Solidus' models maybe?

tvdeyen: Good point. It's hard to tell, because there might be code inside the ApplicationRecord that is related to Solidus' tables as well and there might be cases where this is not the case. I think this can be solved by subclassing ApplicationRecord in the app so those models have a seperate config.

filippoliverani: Thank you for the quick review 🙇 
> This is a good change, thanks 👏
> 
> But, I recommend to name not name this class `ApplicationRecord` even though it is namespaced it potentially conflicts with the top level `ApplicationRecord` new Rails apps have.
> 
> What about `Spree::BaseRecord`? Also, we already have `Spree::Base`, right? 
What do we do about it?

I think that is important to keep `ApplicationRecord` name because its has become a Rails convention, app and engine generators since Rails 5 create this class and almost everybody expects this name.
If we are not sure that namespacing is not enough we could use the `require_dependency` trick suggested here https://guides.rubyonrails.org/engines.html#app-directory

filippoliverani:
@tvdeyen Thank you for the quick review 🙇 
> This is a good change, thanks 👏
> 
> But, I recommend to name not name this class `ApplicationRecord` even though it is namespaced it potentially conflicts with the top level `ApplicationRecord` new Rails apps have.
> 
> What about `Spree::BaseRecord`? Also, we already have `Spree::Base`, right? 
What do we do about it?

I think that is important to keep `ApplicationRecord` name because its has become a Rails convention, app and engine generators since Rails 5 create this class and almost everybody expects this name.
If we are not sure that namespacing is not enough we could use the `require_dependency` trick suggested here https://guides.rubyonrails.org/engines.html#app-directory

core/app/models/spree/application_record.rb
filippoliverani: I'm not convinced that `Spree::ApplicationRecord` should subclass the app one or vice-versa, one of the reasons why `ApplicationRecord` was introduced by Rails was to keep app models hierarchies separated from engines models hierarchies.
And the same goes for `ApplicationController`, `ApplicationMailer` and `ApplicationJob`, they all are [meant to be app or engine specific](https://guides.rubyonrails.org/engines.html#app-directory).

kennyadsl: I think you are right, this is the suggested way: in the documentation linked it states:

> The ArticlesController class inherits from Blorgh::ApplicationController, not the application's ApplicationController.

kennyadsl:
@filippoliverani we discussed this PR during the last core team meeting. We think that probably it's enough to just use `Spree::Base` as our ApplicationRecord, which is doing the same thing and it's an abstract class as well. 

It has a different naming convention than what Rails users expect though but maybe changing the name of that class to `Spree::ApplicationRecord` is not worth it. 

We propose to keep the rest of the changes (using it for models that were inheriting from `ActiveRecord::Base` directly + move the monkey patch into `Spree::Base`) and get rid of the ApplicationRecord file at all. 

What do you think?

filippoliverani:
@kennyadsl thank you for the update, the main point was to stop using `ActiveRecord::Base` directly, I will update the code to reflect the preferred naming convention 👍

kennyadsl:
@tvdeyen `Spree::Base` is already the parent class for all our models. There were just a few that are inheriting from `ActiveRecord::Base` but doesn't seem to exist a specific reason for that.

filippoliverani:
@tvdeyen There's a lot of stuff coming along with Spree::Base, not sure if that could be a problem. At this point is a trade off between having a new lean Spree::ApplicationRecord class or staying with an existing fat Spree::Base class.
Otherwise we could include Ransack and Preferences support as concerns only where needed.

tvdeyen:
@kennyadsl @filippoliverani 

Sorry, I wasn't so sure. Then it makes sense to use it for now. Next step could be to use a lean base class for models that do not need ransack and preferences.

PR #3482, opened 2020-01-22
Remove user address reference when removing address from the address

SamuelMartini:
**Description**

The address book should provide the user the ability to manage all his addresses.

`ship_address_id` and `bill_address_id` are persisted on the `Spree::User` record after transitioning from `address`: `Spree::User#persist_order_address`.
https://github.com/solidusio/solidus/blob/18f6ebc1171155c85594f072e675ea6e51ceb1f9/core/app/models/concerns/spree/user_address_book.rb#L70-L77
These address fields are used to assign user default addresses to the order before transitioning to `address`: `Spree::Order#assign_default_user_addresses`.
https://github.com/solidusio/solidus/blob/d0d1e3d56330c3ccda440d07dd20cbba5aa5934d/core/app/models/spree/order.rb#L716-L726

Currently, when a user empties his address book the `user` record still has the `ship_address_id` and `bill_address_id` references. 
In the `address` checkout step because of `assign_default_user_addresses ` the order addresses are filled with addresses linked on the user record.

As a **user** I expect that when I empty my address book no address will be shown during my next checkout.

As a **developer** I expect to use the address book to provide the user full management capability on his addresses.

This commit removes the address reference from the user record when an address is removed from the address book.

**Checklist:**

### DIFF
--- a/core/spec/models/spree/concerns/user_address_book_spec.rb
+++ b/core/spec/models/spree/concerns/user_address_book_spec.rb
@@ -221,6 +221,7 @@ module Spree
       let(:address1) { create(:address) }
       let(:address2) { create(:address, firstname: "Different") }
       let(:remove_id) { address1.id }
+
       subject { user.remove_from_address_book(remove_id) }
 
       before do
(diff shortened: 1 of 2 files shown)


core/spec/models/spree/concerns/user_address_book_spec.rb
```diff
@@ -242,6 +242,35 @@ module Spree
       it "returns false if the addresses is not there" do
         expect(user.remove_from_address_book(0)).to be false
       end
+
+      context 'when user has previous order addresses' do
```
spaghetticode: I think that besides all these cases when the attributes match and so they are then changed to `nil`, it would be nice to have at least one test for the other scenarios when they don't match and they are not changed... what do you think?

SamuelMartini: Hey @spaghetticode, thank you for the comment! 
Added example when the address does not match any user addresses: https://github.com/solidusio/solidus/pull/3482/files#diff-e930e6c49d94b00d1756dfaf242e4d68R252-R264

PR #3483, opened 2020-01-22
Fix order state with customer returns when receiving return items

elia:
_📝 Notes on reviewing: I suggest to review this PR commit by commit, knowing that the core of the fix is in the last commits:_

1. **Let Order#all_inventory_units_returned? always use fresh data** (88b1220)
  _Which fixes the check on the Order state transition toward "returned"._
2. **Let the order transition to Returned after return items are all Received** (6228fe9)
  _This one contains the feature specs demonstrating the problem along with the changes that make them pass._
3. **Deprecate skip_customer_return_processing** (0dd92fd)
  _Walk back the attempted fix introduced by #3199._

---

**Description**

_Why?_

Currently when a customer return transitions from "In transit" toward "Received" doesn't trigger the transition of the Order status to "Returned".

_How_

The problem is a composite of multiple bugs.

The first one was that `order.inventory_units` was memoized with completely different instances from `ReturnItem#inventory_unit`. So the check to see if they were all returned was, at times, ineffective.

The second part is that the `state_machines` library can halt callback chains without telling anything to anyone. That was happening in some cases when `ReturnItem#attempt_accept` was failing. In those cases, just to increase confusion, the acceptance state-machine was interfering with the reception state-machine.

The third part is that the introduction of the accessor for `skip_customer_return_processing` was fixing one code path (the one still visible in the [original PR](https://github.com/solidusio/solidus/pull/3199)) but was disrupting another one, namely marking the order as returned after the return-item state-machine was getting the "receive" event.

Ref #3199 

**Checklist:**

### DIFF
--- a/backend/spec/controllers/spree/admin/return_items_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/return_items_controller_spec.rb
@@ -8,22 +8,49 @@ describe Spree::Admin::ReturnItemsController, type: :controller do
   describe '#update' do
     let(:customer_return) { create(:customer_return) }
     let(:return_item) { customer_return.return_items.first }
-    let(:old_acceptance_status) { 'pending' }
-    let(:new_acceptance_status) { 'rejected' }
 
-    subject do
-      put :update, params: { id: return_item.to_param, return_item: { acceptance_status: new_acceptance_status } }
-    end
+    describe ':acceptance_status' do
+      let(:old_acceptance_status) { 'pending' }
+      let(:new_acceptance_status) { 'rejected' }
 
-    it 'updates the return item' do
-      expect {
+      subject do
+        put :update, params: { id: return_item.to_param, return_item: { acceptance_status: new_acceptance_status } }
+      end
+
+      it 'updates the return item' do
+        expect {
+          subject
+        }.to change { return_item.reload.acceptance_status }.from(old_acceptance_status).to(new_acceptance_status)
+      end
+
+      it 'redirects to the customer return' do
         subject
-      }.to change { return_item.reload.acceptance_status }.from(old_acceptance_status).to(new_acceptance_status)
+        expect(response).to redirect_to spree.edit_admin_order_customer_return_path(customer_return.order, customer_return)
+      end
     end
 
-    it 'redirects to the customer return' do
-      subject
-      expect(response).to redirect_to spree.edit_admin_order_customer_return_path(customer_return.order, customer_return)
+    describe ':reception_status' do
+      let(:old_reception_status) { 'in_transit' }
+      let(:new_reception_status) { 'received' }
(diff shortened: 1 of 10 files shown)


spaghetticode:
@elia thank you for this fix 🎉 

I see you added many tests at the controller and integration level, those are very good as they expose the functionality and the application flow in the `admin` area, but I'm thinking that, due to the changes to `Spree::ReturnItem` model, maye we should add some model specs in `core` as well in order to document those changes... WDYT?

elia:
@spaghetticode I've added some model-level specs, let me know if you think there's enough coverage now 🔍

kennyadsl:
@elia failures seem legit here, can you please take a look?

core/app/models/spree/order.rb
```diff
@@ -288,7 +288,12 @@ def allow_cancel?
     end
 
     def all_inventory_units_returned?
-      inventory_units.all?(&:returned?)
+      # Inventory units are transitioned to the return state by ReturnItem
+      # across a different association path goes through CustomerReturn and
+      # ReturnItem instead of using Order#inventory_units. That makes any
+      # the cached association on order stale and require a *reload* before
+      # checking if they're returned.
```
aldesantis: I think you could improve the English here a bit, looks like it has been rewritten a few times and the grammar doesn't make much sense anymore. Same goes for the commit.

backend/spec/controllers/spree/admin/return_items_controller_spec.rb
```diff
@@ -34,7 +34,13 @@
       let(:new_reception_status) { 'received' }
       let(:reception_status_event) { 'receive' }
 
-      before { return_item.update reception_status: 'in_transit'}
+      before do
+        allow(Spree::Deprecation).to receive(:warn) do |message|
```
aldesantis: FYI you can use a regexp here, but it's up to you:
```ruby
allow(Spree::Deprecation).to receive(:warn).with(/#process_inventory_unit! will not call/)
```

backend/spec/controllers/spree/admin/return_items_controller_spec.rb
```diff
@@ -34,7 +34,13 @@
...
+          expect(message).to match('#process_inventory_unit! will not call')
+        end
+
+        return_item.update reception_status: 'in_transit'
```
aldesantis: Usually I use `update!` when setting up a test. That way, if the setup code fails for some reason, it does so loudly instead of making the expectations fail.

spaghetticode:
@aldesantis can you review the PR again after Elia's changes?

kennyadsl:
@elia merging. Can you help me writing the changelog entry for this? A comment here is enough, I will take care of moving it into a full changelog entry once we are approaching the release.

Issue #1449, opened 2016-09-16
Order Merging Deleting Orders in Backend Unexpectedly


ericsaupe:
Here is how to recreate the issue:

Logged in user creates an order with something in their cart using the frontend
Admin user creates order in the backend
Admin adds something to order
Admin associates order with Logged in user on the Customer tab
Logged in user refreshes page (or really does anything) causing the orders to be merged
Admin user no longer has that order and the page throws a 404 due to the order being deleted.

I'm not sure what the proper fix is for it we've just seen it as unexpected order behavior for our admins.

aitbw:
Hello there @ericsaupe, tried to reproduce this issue, but no success with the current `master` revision (7665973)

Can you confirm if this issue is still occuring? If so, could you provide more details so I can have a further look on it?

DanielePalombo:
I was able to reproduce this issue. 

The old order [is destroyed](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_merger.rb#L63) (not canceled) after the merge, I was not able to replace the `destroy` with `cancel` because the order not completed can't be canceled.

aldesantis:
We've seen this happening too, where customers would sign in causing guest and authenticated orders to be merged, and one of them (can't remember which one right now) would be deleted, so we'd have non-existing orders in our logs, which makes debugging things a nightmare.

This is expected behavior, but it's far from ideal. I think we could solve it by creating a special `merged` status for orders and then linking them to the new order. That way, we can also very easily redirect admins to the new order when scenario described by @ericsaupe happens.

I could take a stab at this if everyone agrees with the approach.

DanielePalombo:
@aldesantis I started working on the same solution proposed by you!

cesartalves:
Hey @spaghetticode! Does closing https://github.com/nebulab/solidus/issues/37 also entail we should perhaps have closed this issue as well? Or do we need to merge #3486 to fix this issue?

fthobe:
#37 was never actually concluded and #3486 not merged

PR #3486, opened 2020-01-24
Change order state to `merged` instead of destroying it when merging orders

spaghetticode:
**Description**

This is an (opinionated, see later) attempt at https://github.com/solidusio/solidus/issues/1449.

Destroying orders when merging them is not ideal, as we lose the ability to debug what happened, which may be very important in some edge cases. 

This implementation adds the `merged` state to `Spree::Order` and updates the `Spree::OrderMerger` class accordingly. The merged order is associated with the resulting order via the `#merged_to_order` association.

The opinionated part is I kept the existing `Spree::Order#merge!` functionality and interface rather than switching to the one provided by default by the state machine when adding the merge event, see https://github.com/solidusio/solidus/compare/master...nebulab:spaghetticode/order-merged-state?expand=1#diff-2c3e70899d4f22d0569021b01b0e307fR62 (though the auto-generated`Spree::Order#merge` method is retained and used in `Spree::OrderMerge`).
Doing things differently would have required more extensive code changes, deprecations, and other tradeoffs.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -54,6 +54,9 @@ module Spree
     deprecate temporary_credit_card: :temporary_payment_source, deprecator: Spree::Deprecation
     deprecate :temporary_credit_card= => :temporary_payment_source=, deprecator: Spree::Deprecation
 
+    has_many :from_merged_orders, foreign_key: :merged_to_order_id, class_name: 'Spree::Order', inverse_of: :merged_to_order
+    belongs_to :merged_to_order, foreign_key: :merged_to_order_id, class_name: 'Spree::Order', inverse_of: :from_merged_orders, optional: true
+
     # Customer info
     belongs_to :user, class_name: Spree::UserClassHandle.new, optional: true
     belongs_to :bill_address, foreign_key: :bill_address_id, class_name: 'Spree::Address', optional: true
@@ -177,6 +180,14 @@ module Spree
 
     def self.not_canceled
       where.not(state: 'canceled')
+    end
+
+    def self.merged
+      where(state: 'merged')
+    end
+
+    def self.not_merged
+      where.not(state: 'merged')
     end
 
     # Use this method in other gems that wish to register their own custom logic
@@ -846,7 +857,7 @@ module Spree
 
     # Determine if email is required (we don't want validation errors before we hit the checkout)
     def require_email
-      true unless new_record? || ['cart', 'address'].include?(state)
+      true unless new_record? || ['cart', 'address', 'merged'].include?(state)
     end
 
     def ensure_inventory_units
@@ -873,6 +884,12 @@ module Spree
     def validate_line_item_availability
       availability_validator = Spree::Stock::AvailabilityValidator.new
       raise InsufficientStock unless line_items.all? { |line_item| availability_validator.validate(line_item) }
+    end
+
+    def ensure_merged_to_order_present
+      if merged_to_order.blank?
+        errors.add(:base, :merged_to_order_must_be_present) && (return false)
+      end
     end
 
     def ensure_line_items_present
--- a/core/db/migrate/20200117125833_add_merged_to_order_id.rb
+++ b/core/db/migrate/20200117125833_add_merged_to_order_id.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddMergedToOrderId < ActiveRecord::Migration[5.2]
+  def change
+    add_column :spree_orders, :merged_to_order_id, :integer, limit: 4
+    add_foreign_key :spree_orders, :spree_orders, column: :merged_to_order_id
+  end
+end
--- a/backend/app/controllers/spree/admin/orders_controller.rb
+++ b/backend/app/controllers/spree/admin/orders_controller.rb
@@ -49,7 +49,7 @@ module Spree
           params[:q][:completed_at_lt] = params[:q].delete(:created_at_lt)
         end
 
-        @search = Spree::Order.accessible_by(current_ability, :index).ransack(params[:q])
+        @search = Spree::Order.not_merged.accessible_by(current_ability, :index).ransack(params[:q])
         @orders = @search.result.includes([:user]).
           page(params[:page]).
           per(params[:per_page] || Spree::Config[:orders_per_page])
--- a/core/app/models/spree/order/checkout.rb
+++ b/core/app/models/spree/order/checkout.rb
@@ -59,6 +59,10 @@ module Spree
               transition to: :canceled, if: :allow_cancel?
             end
 
+            event :merge do
+              transition to: :merged
+            end
+
             event :return do
               transition to: :returned, from: [:returned, :complete, :awaiting_return, :canceled], if: :all_inventory_units_returned?
             end
@@ -89,7 +93,9 @@ module Spree
               # be added in the correct sequence.
             end
 
-            before_transition from: :cart, do: :ensure_line_items_present
+            before_transition from: :cart, except_to: :merged, do: :ensure_line_items_present
+
+            before_transition to: :merged, do: :ensure_merged_to_order_present
 
             if states[:address]
               before_transition to: :address, do: :assign_default_user_addresses
@@ -105,6 +111,11 @@ module Spree
 
             before_transition to: :resumed, do: :ensure_line_item_variants_are_not_deleted
             before_transition to: :resumed, do: :validate_line_item_availability
+
+            before_transition from: :merged do |order|
+              order.errors.add(:base, :cannot_transition_from_merged)
+              false
+            end
 
             # Sequence of before_transition to: :complete
             # calls matter so that we do not process payments
--- a/core/app/models/spree/order_merger.rb
+++ b/core/app/models/spree/order_merger.rb
@@ -31,15 +31,16 @@ module Spree
     # will add the quantity of the incoming line item to the existing line item.
     # Otherwise, it will assign the line item to the new order.
     #
-    # After the orders have been merged the `other_order` will be destroyed.
+    # After the orders have been merged the `other_order` will be marked as merged.
     #
     # @example
     #   initial_order = Spree::Order.find(1)
     #   order_to_merge = Spree::Order.find(2)
     #   merger = Spree::OrderMerger.new(initial_order)
     #   merger.merge!(order_to_merge)
-    #   # order_to_merge is destroyed, initial order now contains the line items
-    #   # of order_to_merge
+
+    # order_to_merge is marked as merged, initial order now contains the line
+    # items of order_to_merge
     #
     # @api public
     # @param [Spree::Order] other_order An order which will be merged in to the
@@ -58,9 +59,8 @@ module Spree
       set_user(user)
       persist_merge
 
-      # So that the destroy doesn't take out line items which may have been re-assigned
-      other_order.line_items.reload
-      other_order.destroy
+      other_order.merged_to_order = order
+      other_order.merge && other_order.save!
     end
 
     private
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -469,6 +469,12 @@ en:
             price:
               not_a_number: is not valid
               does_not_match_order_currency: Line item price currency must match order currency!
+        spree/order:
+          attributes:
+            base:
+              cannot_transition_from_merged: Cannot transition from merged
+              merged_to_order_must_be_present: The order must be associated with
+                the order that is merged into via the `merged_to_order` association
         spree/price:
           attributes:
             currency:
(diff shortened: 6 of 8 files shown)


core/db/migrate/20200117125833_add_merged_to_order_id.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddMergedToOrderId < ActiveRecord::Migration[5.2]
+  def change
+    add_column :spree_orders, :merged_to_order_id, :integer, limit: 4
+    add_foreign_key :spree_orders, :spree_orders, column: :merged_to_order_id
```
elia: I would also consider adding a `merged_at` timestamp, especially when reconstructing weird edge cases that can be a precious information. In alternative the default order merger should at least add a note mentioning the date of the merge.

core/app/models/spree/order.rb
```diff
@@ -54,6 +54,9 @@ class CannotRebuildShipments < StandardError; end
     deprecate temporary_credit_card: :temporary_payment_source, deprecator: Spree::Deprecation
     deprecate :temporary_credit_card= => :temporary_payment_source=, deprecator: Spree::Deprecation
 
+    has_one :from_merged_order, foreign_key: :merged_to_order_id, class_name: 'Spree::Order', inverse_of: :merged_to_order
```
elia: I recall that a single orders can be the result of merging multiple orders. Should this be a `has_many`?

https://github.com/solidusio/solidus/blob/9f99aa6f084ea3c9d1015edc6c39fb119b169b8d/core/lib/spree/core/controller_helpers/order.rb#L60-L66

elia (dismissed):
👍 this is great, debugging merged orders is like trying to catch ghosts 👻

I left a couple suggestions/considerations.

core/app/models/spree/order.rb
spaghetticode: @elia thank you for pointing this out, I did not realize this could happen as well.

core/db/migrate/20200117125833_add_merged_to_order_id.rb
spaghetticode: @elia I think that would be redundant. We have `state_changes` on orders, and since `merged` is a new state, a new record pops up after merging the order:
```ruby
2.6.3 :001 > order2.state_changes
=> [#<Spree::StateChange:0x00007fea97cd4f90
  id: 1,
  name: "order",
  previous_state: "cart",
  stateful_id: 2,
  user_id: 2,
  stateful_type: "Spree::Order",
  next_state: "merged",
  created_at: Fri, 07 Feb 2020 11:28:08 UTC +00:00,
  updated_at: Fri, 07 Feb 2020 11:28:08 UTC +00:00>]
```

core/app/models/spree/order.rb
```diff
@@ -179,6 +182,14 @@ def self.not_canceled
       where.not(state: 'canceled')
     end
 
+    def self.merged
```
ericsaupe: Is there a reason these aren't scopes?

mamhoff:
We've come across this as well, but we have opted not to add more stuff to the order state machine (which IMO does way too many things already). What we did was create a `frontend_viewable` boolean column on the orders table. Admin-created orders get that set to false, and only "frontend viewable" orders can be merged.

core/app/models/spree/order.rb
spaghetticode: Only for consistency with the `::cancelled` and `::not_cancelled` class methods defined just above.

spaghetticode:
@mamhoff thanks for the feedback. I think you have a point about the state machine, though I don't see adding this new state as a big concern... still this solution goes in the opposite direction of simplifying the SM (BTW, I like your PR!).

I think one big plus for the state machine is that the merged order is locked down, so it will not transition to other states. IMHO this cannot be achieved in a similarly straight forward way by using other approaches. 

But more in general, all the possible solutions to this issue have a problem in common: there are (irrelevant?) orders that are not destroyed anymore, and DB queries will start picking them up, unless stores actively remove them in their custom code. So, stopping to destroy these orders looks to me like a significant breaking change.

I'm wondering if, instead of providing a new default approach to the problem, we should just offer to stores a few alternative merging strategies, built-in in Solidus, like this one (and yours) and let them chose their preference. But I'm not sure that a full-featured solution can be limited to just setting one preference, for example the merge strategy class, as different solutions will also need different ways to exclude merged orders from DB queries... maybe an extension would be a more suitable place 🤔

fthobe:
@kennyadsl has there been made any decision on this? I find merging orders a compelling feature but would like some guidance as it's not something we would extensively invest into. `Merged` Seems to be a desirable state, but I am not clear if it is wanted by core team.

kennyadsl:
I'm ok with this change, but this is a big one and potentially very problematic in terms of backward compatibility, so I guess it need to go in the next major release (upon @solidusio/core-team consensus).

elia:
@fthobe the non-breaking way of merging this would be to put it behind a configuration flag that's not active by default. That would be great although not always trivial. 

In any case someone needs to update the PR and rebase on the latest version, so if anyone is willing to take up both task (rebasing and adding the configuration flag) I would be absolutely in favor 👍.

fthobe:
> @fthobe the non-breaking way of merging this would be to put it behind a configuration flag that's not active by default. That would be great although not always trivial. 
> 

A question came up:
As @spaghetticode said the merge state does not create any destinations from there, where do you see the breaking change other than keeping more orders in DB?

tvdeyen:
Wait. Can we please step back a little bit and talk about the use case of this?

A bit of background on the usual user flow will be helpful I guess. 

So, a yet anonymous user comes to the store and puts something in to the cart. A new order record gets created. This is an unusual, but powerful feature of Solidus. Even the cart is already an order record.

The user then either creates a new account or logs into an existing account. Either way the anonymous order (the cart from current session) gets merged with all incomplete orders (due to the fact this happens every time the user logs in it's usually one incomplete order). And the previous anonymous and now merged order gets destroyed, because it does not serve any purpose anymore. The cart has now been transferred into a users order.

If we now keep the anonymous order it will produce a lot of garbage in the database. 

What is the use case of keeping them?

This behavior needs to be at least behind a feature flag or even in a separate order state machine. And users need to be informed about this behavior. 

But in the first place I would love to hear about the benefits.

fthobe:
Honestly, pure data collection and profiling. 

- after how many days does a cart lead to a purchase 
- how often a product leads to a registration but not to a purchase 

Half of today's recommendation engines are either based on sessions or abandoned / concluded carts. 

Actually I would go many steps further: 
- is there a desire to bring over the cart state date over to the new order
- maybe even track line item events with timestamps 

Throw it in google data studio or tableau and you have some fascinating metrics to display. 

There's a whole order life cycle and long term abandonment data story to tell here and I wouldn't want to drop any of it.

tvdeyen:
All stores I work on use third party tools for that. Even if we would use the state machine for that we would need to track the order the anonymous cart has been merged in. If there is a need for that, you can implement it in your store or create an extension. The order merger class and the state machine are configurable classes and easily adopted to specific needs that way.

fthobe:
I can see your point why this can be done in my store, but I still struggle for any number of reasons with destroying any data without proper documentation. I feel taking up the decision to destroy data is massive compared to cost of storage.

A fully agree to keep a flag, a separate state machine I see as drastically complicating things (might be wrong though). Can we take a look at the current PR and see if it's salvageable (later this year) and put it behind a setting.

kennyadsl:
If I recall correctly, the main driver for this change was that it's very hard to debug issues on orders that are merged and destroyed, so suddenly disappear from the database while still being present in logs prior to the merge.

spaghetticode:
@kennyadsl you're right, that was the original driver for the PR. 
We discussed the change internally a few times. We never reached a consensus about moving it forward - as others already said this is a significant behavior change, it's already a sensible default, and custom behavior can always be added on top of it. I see there's some interest in moving it forward, but due to the concerns I'm thinking it may end up in a small extension.

fthobe:
What spaghetti Said.

fthobe:
@tvdeyen A stupid question: Why is it not possible to Carry over the Order, why do we need to destroy it in the First Place?

tvdeyen:
> @tvdeyen A stupid question: Why is it not possible to Carry over the Order, why do we need to destroy it in the First Place?

I was not involved or even around when this was introduced into Spree decades ago. My guess is, because it was considered as not useful information, regarding that this is a stale anonymous cart after it has been merged. Without keeping track of where it has been merged into it does not make sense to keep, I guess. It leads to false data, because the cart content has been merged over to a new order and we do not know who it belongs to. It could give a false impression of having lots of abandoned carts, although they have been converted into a payed order later on.

fthobe:
Funny enough this might be a thing that is not breaking if we just update the user of an order instead of destroying it, or am I wrong?

elia:
@spaghetticode if I recall correctly we also had a case in which some payments were attached to the then destroyed order, that alone to me would be a good reason to reverse the default in the next major. Customization can still be implemented either to remove the new state or to add other behaviors.



> Funny enough this might be a thing that is not breaking if we just update the user of an order instead of destroying it, or am I wrong?

No, I think this is triggered when you have an anonymous cart and you login to a customer account that also have a current cart, and you want to merge the items from both.

fthobe:
Seems like only bad options here. 

@tvdeyen can you live with hiding it behind a setting?

spaghetticode:
@elia that's quite likely - the merged orders issue emerged multiple times while working on that particular store. 
And thanks for pointing that out, I'd say that in general, anything attached to the order can become either an orphan record or get destroyed depending on how the association is configured.

fthobe:
> And thanks for pointing that out, I'd say that in general, anything attached to the order can become either an orphan record or get destroyed depending on how the association is configured.

Do you have some examples: I mean this applies only to uncompleted orders.

elia:
@fthobe where exactly the payment happens can vary and I'm sure some older payment methods tried to mark an order as complete _after_ the payment was successful. I think something else at that point caused the order to reset to the initial state and then it got merged. 

It's admittedly an edge case but we really wished to have a better paper trail at the time.

tvdeyen:
> we also had a case in which some payments were attached to the then destroyed order

This seems to be a bug rather than anything else. Only incomplete orders get merged. I would like to not keep garbage just because there might be potential bugs in customers implementations. There are plenty of options (like [paper_trail](https://github.com/paper-trail-gem/paper_trail)) to keep track of records changes.

fthobe:
It would not be something we touch before June / July. Do we want to leave this open or is it definitively something that has no chance on merging?

elia:
@tvdeyen it was due to how the state machine handles payments rather than a customer implementation bug, and the old concept that an order is complete only after a payment is attached to it.

We are not cleaning up abandoned carts by default, which is something that was needed in every single store I worked on eventually. In that case it's easy to add a recurring cleanup job that drops merged orders older than some date.

Do you see any negative impact besides having additional order data around?

tvdeyen:
> @tvdeyen it was due to how the state machine handles payments rather than a customer implementation bug, and the old concept that an order is complete only after a payment is attached to it.
> 
> We are not cleaning up abandoned carts by default, which is something that was needed in every single store I worked on eventually. In that case it's easy to add a recurring cleanup job that drops merged orders older than some date.
> 
> Do you see any negative impact besides having additional order data around?

No, if we make sure that the orders are clearly marked as what they are I have no objections.

An unobtrusive way could be to introduce a `merged_into_order_id` foreign key and maybe a `merged_at` timestamp, so that the data can be reasoned about. Using the state machine does not seem to make much sense to me.

fthobe:
> > @tvdeyen it was due to how the state machine handles payments rather than a customer implementation bug, and the old concept that an order is complete only after a payment is attached to it.
> > We are not cleaning up abandoned carts by default, which is something that was needed in every single store I worked on eventually. In that case it's easy to add a recurring cleanup job that drops merged orders older than some date.
> > Do you see any negative impact besides having additional order data around?
> 
> No, if we make sure that the orders are clearly marked as what they are I have no objections.
> 
> An unobtrusive way could be to introduce a `merged_into_order_id` foreign key and maybe a `merged_at` timestamp, so that the data can be reasoned about. Using the state machine does not seem to make much sense to me.

Yes, that would be the absolute dream. Can we settle on that we list this for July?

fthobe:
> An unobtrusive way could be to introduce a `merged_into_order_id` foreign key and maybe a `merged_at` timestamp, so that the data can be reasoned about. Using the state machine does not seem to make much sense to me.

I think that was pretty much what was made in this PR
`has_one :from_merged_order, foreign_key: :merged_to_order_id, class_name: 'Spree::Order', inverse_of: :merged_to_order`

It was pointed out by @Elia that merged could come from multiple orders. 
I would like to point out that this could also make for a real nice feature later on to: 
- complete orders on the phone with customers
- merge multiple stale orders 
- pay per link

Has lot of potential!

jarednorman:
Closing due to no activity in the last year. I'm not against the change, but we need a store that needs and wants to pick this up, get it working, and offer to test it.

PR #3487, opened 2020-01-24
Fix stock configuration examples in documentation

gugaiz:
Fix an issue with the documentation related to the custom configuration
of the stock location filter and sorter. The way it was documented
didn't work.

**Checklist:**

### DIFF
--- a/guides/source/developers/shipments/stock-allocator.html.md
+++ b/guides/source/developers/shipments/stock-allocator.html.md
@@ -50,9 +50,8 @@ For example, you can register it in your `/config/initializer/spree.rb` initiali
 ```ruby
 # /config/initializer/spree.rb
 Spree.config do |config|
-    # ...
-
-    config.stock.allocator_class = 'Spree::Stock::Allocator::CustomAllocator'
-  end
+  # ...
+  config.stock.allocator_class = 'Spree::Stock::Allocator::CustomAllocator'
+  # ...
 end
 ```
--- a/guides/source/developers/shipments/stock-locations-filter.html.md
+++ b/guides/source/developers/shipments/stock-locations-filter.html.md
@@ -41,7 +41,10 @@ the split shipments logic.
 For example, you can register it in your `/config/initializers/spree.rb` initializer:
 
 ```ruby
-# /config/initializers/spree.rb
-
-Rails.application.config.spree.stock.location_filter_class = 'Spree::Stock::LocationFilter::SameOrderCountry'
+# /config/initializer/spree.rb
+Spree.config do |config|
+  # ... 
+  config.stock.location_filter_class = 'Spree::Stock::LocationFilter::SameOrderCountry'
+  # ... 
+end
 ```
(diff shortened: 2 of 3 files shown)


gugaiz:
@kennyadsl, I tried that but didn't work for me neither.

kennyadsl:
Just to be sure, are you restarting the rails server after these changes?

gugaiz:
yes

kennyadsl:
Sorry, my fault, it should be:

```
Spree::Config.stock.location_filter_class = 'Spree::Stock::LocationFilter::SameOrderCountry'
```

gugaiz:
Ok, that worked. I can change it with a new commit if you prefer that way

kennyadsl:
Yes, I think it's better. And please, can you also squash commits into a single one before pushing? Thanks again! 🙂

gugaiz:
Done

spaghetticode:
Out of curiosity, I tried to verify the issue locally. 

So, I generated a sandbox app with `rake sandbox`, then edited the `config/initializers/spree.rb` like this:

```ruby
class CustomAllocator
end
class CustomFilter
end
class CustomSorter
end

Spree.config do |config|
  config.stock.allocator_class = 'CustomAllocator'
  config.stock.location_filter_class = 'CustomFilter'
  config.stock.location_sorter_class = 'CustomSorter'
  # ...
end
```

then starting the sandbox console with `rails c` I got the expected custom configuration values:

```ruby
Loading development environment (Rails 6.0.2.1)
2.6.3 :001 > Spree::Config.stock.allocator_class
 => CustomAllocator 
2.6.3 :002 > Spree::Config.stock.location_filter_class
 => CustomFilter 
2.6.3 :003 > Spree::Config.stock.location_sorter_class
 => CustomSorter 
```

So, in this case it seems that following the existing documentation work properly. @gugaiz can you explain in detail what kind of issues you experienced?

gugaiz:
@spaghetticode that works and that was my first proposal, but @kennyadsl preferred the way is documented on this PR ( `Spree::Config.stock.location_filter_class` )
The current documentation suggest to use `Rails.application.config.spree.stock.location_filter_class` which does not work.

spaghetticode:
@gugaiz thanks for the clarification.  So, the documentation is wrong only for the two files that used `Rails.application.config`. I verified in the sandbox console that the correct way to access that configuration via `Rails.application.config` would be with:
```
Rails.application.config.spree.preferences.stock.location_filter_class
```

and 

```
Rails.application.config.spree.preferences.stock.location_sorter_class
```

I have no idea if we should prefer one way or the other(s) for this case, maybe we should document them all?

gugaiz:
@spaghetticode I guess someone else might answer that. If you ask me, I prefer the way below because is what you find when you open the `config/initializer/spree.rb` file 

```
Spree.config do |config|
  config.stock.allocator_class = 'CustomAllocator'
  config.stock.location_filter_class = 'CustomFilter'
  config.stock.location_sorter_class = 'CustomSorter'
  # ...
end
```

tvdeyen:
> @spaghetticode I guess someone else might answer that. If you ask me, I prefer the way below because is what you find when you open the `config/initializer/spree.rb` file
> 
> ```
> Spree.config do |config|
>   config.stock.allocator_class = 'CustomAllocator'
>   config.stock.location_filter_class = 'CustomFilter'
>   config.stock.location_sorter_class = 'CustomSorter'
>   # ...
> end
> ```

I also prefer that way

kennyadsl:
I'm fine with both, so 👍 if @gugaiz has time to change it.

gugaiz:
@kennyadsl I think is done

Issue #3393, opened 2019-10-23
RFC: Remove Paranoia


jarednorman:
We committed previously to fully moving from paranoia to discard in #2354. The initial work was completed and all models that "act as paranoid" include `Spree::ParanoiaDeprecations` which warns users that we are removing the paranoia gem from the project and that `destroy` is actually going to destroy things in the future. All of those models also include `Discard::Model` and are ready to be fully switched over.

In order to complete the switchover, I believe we need to:

1. Make sure we are using the `kept` scope as appropriate everywhere we need to. This isn't a small task, but is totally achievable. *This work can be merged prior to actually removing the paranoia gem as it won't actually affect the behaviour of the system.*
1. Remove the `paranoia` gem, all the references to `acts_as_paranoid`, and the `Spree::ParanoiaDeprecations` mixin. *As the deprecation warns, this is a breaking change and should likely be included in next _major_ release after it is merged.*

Does anyone have any concerns about following through on this work and removing paranoia in the next major release?

awinabi:
@jacobherrington I can take an attempt at this, if its open for work

jarednorman:
I've cleaned out the outstanding issues/pull requests on the Discard project (I maintain it) and have started working on this, but assistance will be helpful. We're looking at using something like the functionality mentioned in [this Discard issue](https://github.com/jhawthorn/discard/issues/46) to ensure that we don't miss any queries. I'm going to post a checklist of all the models that need to be migrated in the description of this task to make sure we don't duplicate work or miss anything.

cedum:
I started looking into this too. While I like a lot the simplicity of `discard`, I think that migrating the Solidus' codebase, extensions, and client's codebases requires quite an effort: we've to manually review and test every query that touches in any way (`joins` for example) these tables.
I'm also not sure that enforcing an explicit scope/query condition on all these queries is the right thing to do. AFAIK, there're very few cases where we need to work with soft-deleted data, most of the time we're working with non-deleted records.

I'm not sure what are the alternatives. Maybe we should try to evaluate a completely different data structure for soft-deleted records, something similar to [`paper_trail`](https://github.com/paper-trail-gem/paper_trail) (never used it in production, not sure if it's the right way to go, just an 💡).

jarednorman:
[...]

Issue #2354, opened 2017-11-04
Replace acts_as_paranoid/paranoia with discard


jhawthorn:
Our acts_as_paranoid models have been the cause of a lot of pain and confusion.

`paranoia` overrides activerecord's `delete` and `destroy` methods, which can surprise users expecting a real delete, but also surprises users when a record isn't declared `acts_as_paranoid` (or they use a method acts_as_paranoid doesn't override, like `delete_all`). It also has callback issues, `dependent: :destroy` and any other `after_destroy` callbacks happen on a soft-destroy or a real destroy.

Ideally, these would all be switched to use [discard](https://github.com/jhawthorn/discard), my extremely simple alternative: <40 LOC and no ActiveRecord overrides. But it's in no way a drop-in replacement.

A possible stepping stone is to switch to `paranoia_delete` first, which has much more similar semantics to discard (no `dependent: :destroy` callbacks run).

To start planning how to replace it I've cataloged all the models in core it is used on, and the effects either hard or soft deletes have.

* [x] Product/Variant/StockItem/Price #2396
* [x] Promotion actions #2398
* [x] Shipping methods #2487  
* [x] Payment methods #2487
* [x] Tax rates #2487 
* [x] Store credits #2489
* [x] ~Stock Transfers~ (Removed by #2379)

## Product models
* Product
  * Hard-deletes: variants, product_promotion_rules, classifications, product_properties, product_option_types
  * Soft-deletes: variants
  * Orphaned on real destroy: variant_property_rules
* Variant
  * Hard-deletes: images
  * Soft-deletes: stock_items, prices
  * Orphaned on real destroy: inventory_units, line_items, variant_property_rules
  * Hard-deletes on real destroy: option_values_variants
* StockItem
  * Orphaned: stock_movements
* Price

This is the most complicated group of models. Also the most frequently used, with the most possible admin interactions.

I'm not 100% sure why `Price` is `acts_as_paranoid`, but I suspect it's just to prevent them being destroyed when a variant is soft-destroyed. Other than that I think it would be safe for them to be hard-destroyed. Maybe there is a need to keep them around for auditing (in which case we should represent them better and show the "historic" prices in the admin)

Product and Variant should only ever be soft-destroyed. Soft-destroying a product will soft-destroy all the variants, which seems reasonable (although makes it harder to distinguish variants which were removed from a product from variants whose product was deleted. Might not matter.)

[...]

PR #3488, opened 2020-01-24
Discard Paranoia

cedum:
**Description**
This removes [paranoia](https://github.com/rubysherpas/paranoia) and leaves [discard](https://github.com/jhawthorn/discard) as the only option for soft-deleting records.

Ref issues: #3393, #2354 

**TODO:**
- [x] Deprecate paranoia methods and alias them to discard (e.g. `with_deleted`). PR #3554
- [x] Remove `paranoia` gem from dependencies once [`solidus_auth_devise`](https://github.com/solidusio/solidus_auth_devise/blob/c9293c13998917cf04a0d0b0fd00c7f470fa49ae/app/models/spree/user.rb#L11) dependency will be fixed: it's using paranoia but not declaring it explicitly as a dependency. Created PR on `solidus_auth_devise` solidusio/solidus_auth_devise#183
- [x] [`Spree::Api::ResourceController#destroy`](https://github.com/solidusio/solidus/blob/f6844dfcd7ec252fc80b3cb4dfb174483df015a4/api/app/controllers/spree/api/resource_controller.rb#L56-L60) should support models with `discard` the same way as [`Spree::Admin::ResourceController#destroy`](https://github.com/solidusio/solidus/blob/master/backend/app/controllers/spree/admin/resource_controller.rb#L91-L98) PR #3854

**Checklist:**

### Upgrade Notes

[Paranoia](https://github.com/rubysherpas/paranoia) has been replaced by [Discard](https://github.com/jhawthorn/discard) gem for models having the soft-delete feature enabled.\
For more details about this move check [this issue](https://github.com/solidusio/solidus/issues/2354) where the rationale has been discussed.

The soft-delete feature in Solidus, beyond being used as an archival tool, is mostly required to prevent hard-deletion of records other critical models depends on (i.e. Order).

NOTE: the soft-deletable models still have a **default scope** that filters out soft-deleted records to ease the migration.

Starting with Solidus v3 the following methods introduced by Paranoia are replaced by their Discard equivalents:

#### Class Methods

- [**`with_deleted`**][with_deleted] becomes [**`with_discarded`**][with_discarded]\
Returns both active and soft-deleted records.

- [**`only_deleted`**][only_deleted] becomes [**`discarded`**][discarded]\
Returns only soft-deleted records.

- [**`restore(id_or_ids)`**][restore] becomes [**`where(id: [id(s)]).undiscard_all`**][undiscard_all]\
Restores a list of soft-deleted records.

#### Instance Methods

- [**`destroy`**][paranoia_destroy] (also [**`paranoia_destroy`**][paranoia_destroy]) becomes [**`discard`**][discard]\
Soft-deletes a record by firing also any callbacks defined for the model.\
_Before Solidus v3_ when calling `#destroy`, hence soft-deleting a record, any `[before|around|after]_destroy` callbacks defined for that model were also fired.\
_With Solidus v3_ any callbacks when **soft-deleting** a record must be defined using [`[before|around|after]_discard` API](https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L21).\
`[before|around|after]_destroy` will be fired instead when **hard-deleting** a record.

- [**`delete`**][paranoia_delete] (also [**`paranoia_delete`**][paranoia_delete]) becomes [**`discard`**][discard]* _(*see notes below)_\
Soft-deletes a record without firing the callbacks.\
Discard doesn't provide a direct equivalent for soft-deleting without firing the `[before|around|after]_discard` callbacks.\
The same behavior may be obtained by updating directly the soft-delete column:\
`update_attribute(self.class.discard_column, Time.current)`

- [**`restore`**][restore!] (also [**`restore!`**][restore!]) becomes [**`undiscard`**][undiscard] (and [**`undiscard!`**][undiscard!]) \
restores a soft deleted record. The methods with a bang `!` raise an exception when failed.

- [**`really_destroy!`**][really_destroy!] becomes [**`destroy`**][destroy]\
hard-deletes a record.

- [**`paranoia_destroyed?`**][paranoia_destroyed?] (also [**`deleted?`**][paranoia_destroyed?]) becomes [**`discarded?`**][discarded?]\
checks if a record has been soft-deleted.

#### Model Callbacks

- [**`[before|around|after|]_destroy`**][paranoia_destroy_callbacks] becomes [**`[before|around|after|]_discard`**][discard_callbacks]
fired when soft-deleting a record.
- [**`[before|around|after|]_real_destroy`**][real_destroy_callbacks] becomes [**`[before|around|after|]_destroy`**][ar_destroy_callbacks]
fired when hard-deleting a record.
- [**`[before|around|after|]_restore`**][restore_callbacks] becomes [**`[before|around|after|]_undiscard`**][undiscard_callbacks]
fired when restoring a soft-deleted a record.

[with_deleted]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L27-L32
[with_discarded]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L19
[only_deleted]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L34-L46
[discarded]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L18
[restore]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L48-L57
[undiscard_all]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L65-L82
[paranoia_destroy]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L60-L77
[discard]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L115-L123
[paranoia_delete]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L84-L96
[restore!]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L98-L122
[undiscard]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L137-L145
[undiscard!]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L147-L157
[really_destroy!]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L140-L163
[destroy]: https://apidock.com/rails/ActiveRecord/Persistence/destroy
[paranoia_destroyed?]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L135-L138
[discarded?]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L104-L107
[paranoia_destroy_callbacks]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L62
[discard_callbacks]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L21
[real_destroy_callbacks]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L235
[ar_destroy_callbacks]: https://apidock.com/rails/ActiveRecord/Callbacks
[restore_callbacks]: https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L235
[undiscard_callbacks]: https://github.com/jhawthorn/discard/blob/f391785f5d9898d6ce67cee0b9929051e1b5bb44/lib/discard/model.rb#L22


core/app/models/spree/payment_method.rb
```diff
@@ -108,9 +103,9 @@ def active?
         where(type: to_s, active: true).count > 0
       end
 
-      # @deprecated Use .with_deleted.find instead
+      # @deprecated Use .with_discarded.find instead
       def find_with_destroyed(*args)
```
spaghetticode: Not sure if this PR would be the best place for doing it, but I think that this deprecated method may just be removed. 
Since this PR already has many breaking changes, it may make sense to take advantage of that in order to simply remove this old code as well.

core/app/models/concerns/spree/soft_deletable.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module Spree
+  module SoftDeletable
+    extend ActiveSupport::Concern
+
+    included do
+      include Discard::Model
+      self.discard_column = :deleted_at
+
+      default_scope { kept }
```
spaghetticode: `default_scope` has a bit of bad rap, see for example https://rails-bestpractices.com/posts/2013/06/15/default_scope-is-evil/

I'm not implying we should not use it at all, but we should make store maintainers explicitly aware of its existence, with a message such as please verify your discard models overrides don't include a `default_scope` and that the new `default_scope` does not interact weirdly with your custom code.

jarednorman: Yeah, I understand the desire to use it here as it makes the migration significantly easier. I'd prefer we not use it by default for any future models we make soft-deletable though.

core/app/models/concerns/spree/soft_deletable.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    module ClassMethods
+      def with_deleted
+        Spree::Deprecation.warn <<-WARN.strip_heredoc, caller
```
jarednorman: Not really significant, but I assume our Ruby version support is now such that we can use squiggly heredocs, making the `.strip_heredoc` unnecessary in this file.

core/lib/spree/core.rb
```diff
@@ -14,7 +14,6 @@
 require 'state_machines-activerecord'
 
 require 'spree/deprecation'
-require 'spree/paranoia_deprecations'
```
jarednorman: 🎉

kennyadsl (approved):
The best PR's title ever! Thanks @cedum

PR #3493, opened 2020-01-29
Update images sizes to the new defaults

elia:
**Description**
Let the customization example use the old pre 2.10 defaults to make it
easier to copy/paste the code for people landing on this page from the 
2.10 changelog.

**Checklist:**

### DIFF
--- a/guides/source/developers/products-and-variants/product-images.html.md
+++ b/guides/source/developers/products-and-variants/product-images.html.md
@@ -50,24 +50,20 @@ You can check the default settings by calling the `attachment_definitions`
 method on `Spree::Image` in your Rails console:
 
 ```ruby
-Spree::Image.attachment_definitions[:attachment][:styles]
-=> {
-  mini=>"48x48>",
-  small=>"100x100>",
-  product=>"240x240>",
-  large=>"600x600>"
-}
+>> Spree::Image.attachment_definitions[:attachment][:styles]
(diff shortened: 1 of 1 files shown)


guides/source/developers/products-and-variants/product-images.html.md
```diff
@@ -50,24 +50,19 @@ You can check the default settings by calling the `attachment_definitions`
...
-  small=>"100x100>",
-  product=>"240x240>",
-  large=>"600x600>"
-}
+>> Spree::Image.attachment_definitions[:attachment][:styles]
+=> {:mini=>"48x48>", :small=>"400x400>", :product=>"680x680>", :large=>"1200x1200>"}
 ```
 
 The default sizes can be changed in an initializer. For example, in your
 `config/initializers/paperclip.rb` file. You can set new defaults like this:
```
spaghetticode: Would it make sense to make it super clear that those were the old defaults? For example by changing the wording to something like:
> You can set new defaults like this (these were the default values for Solidus up to version 2.9.x):

kennyadsl: @spaghetticode that's a great suggestion, it makes total sense for me. @elia thoughts?

elia:
@kennyadsl @spaghetticode thanks for the comment!
I added a comment in the code block specifying those are coming from Solidus pre 2.9.

PR #3494, opened 2020-01-29
Use create_if_necessary instead of a simple find_or_initialize

### LINKED ISSUE #3467: Inconsistent Order Lookup Between Controllers
Hey folks,

I am looking for some insight here, perhaps you can help me. We ran into a case in production where an order shows up in the cart but upon trying to proceed with checkout, no order is found and the user is redirected back to the cart. 

This is fundamentally because in `Spree::OrdersController#edit`:
https://github.com/solidusio/solidus/blob/master/frontend/app/controllers/spree/orders_controller.rb#L42

```
@order = current_order || Spree::Order.incomplete.find_or_initialize_by(guest_token: cookies.signed[:guest_token])
```

However, in `Spree::CheckoutController#load_order`: 
https://github.com/solidusio/solidus/blob/master/frontend/app/controllers/spree/checkout_controller.rb#L136
``
@order = current_order
``

If we unpack `current_order` behavior further, you will see it uses a different query to load:
https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/controller_helpers/order.rb#L86

```
order = Spree::Order.incomplete.lock(options[:lock]).find_by(current_order_params)
```

and

```
        def current_order_params
          { currency: current_pricing_options.currency, guest_token: cookies.signed[:guest_token], store_id: current_store.id, user_id: try_spree_current_user.try(:id) }
        end
```

In our specific case, the miss was happening because we had a state with the Order have a non-nil `user_id` and the query kept asking for `nil` because the other session did not exist. 

**Solidus Version**:
`master`

**To Reproduce**
I do not have all of the specific to reproduce but am hoping for some insightful discussion.

**Current behavior**
Orders and Checkout controllers load orders differently. 

**Expected behavior**
I would expect them to abide by the same business logic and am curious why that's not the case. I understand this behavior is inherited from a much older Spree release, but hoping folks have extra insight. 

Thanks, y'all!

elia:
**Description**

Using current_order with create_if_necessary creates order that are
much more complete adding info about the ip, creator and other meta
data that is better to have rather than not. Not to mention that it's
also more readable.

Fixes #3467

I did a historical check and no specific reason for doing it this way came about.
- https://github.com/solidusio/solidus/commit/3516aa39c903faeca1d18746bfeefc7b138b1cae
- https://github.com/solidusio/solidus/commit/4efa7898003f30ad265eba8a39e2f59dc3f5f24b

**Checklist:**

### DIFF
--- a/core/spec/lib/spree/core/controller_helpers/order_spec.rb
+++ b/core/spec/lib/spree/core/controller_helpers/order_spec.rb
@@ -75,6 +75,23 @@ RSpec.describe Spree::Core::ControllerHelpers::Order, type: :controller do
         }.from(nil).to("0.0.0.0")
       end
     end
+
+    context 'build_order_if_necessary option is true' do
+      subject { controller.current_order(build_order_if_necessary: true) }
+
+      it 'builds a new order' do
+        expect { subject }.not_to change(Spree::Order, :count).from(0)
+        expect(subject).not_to be_persisted
+      end
+
+      it 'assigns the current_store id' do
+        expect(subject.store_id).to eq store.id
+      end
+
+      it 'records last_ip_address' do
+        expect(subject.last_ip_address).to eq("0.0.0.0")
+      end
+    end
   end
 
   describe '#associate_user' do
--- a/core/spec/lib/spree/core/controller_helpers/order_spec.rb
+++ b/core/spec/lib/spree/core/controller_helpers/order_spec.rb
@@ -2,12 +2,13 @@
 
 require 'rails_helper'
 
-class FakesController < ApplicationController
-  include Spree::Core::ControllerHelpers::Order
-end
-
 RSpec.describe Spree::Core::ControllerHelpers::Order, type: :controller do
-  controller(FakesController) {}
+  controller(ApplicationController) {
+    include Spree::Core::ControllerHelpers::Store
+    include Spree::Core::ControllerHelpers::Pricing
+    include Spree::Core::ControllerHelpers::Auth
+    include Spree::Core::ControllerHelpers::Order
+  }
 
   let(:user) { create(:user) }
   let(:order) { create(:order, user: user, store: store) }
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -39,7 +39,7 @@ module Spree
 
     # Shows the current incomplete order from the session
     def edit
-      @order = current_order || Spree::Order.incomplete.find_or_initialize_by(guest_token: cookies.signed[:guest_token])
+      @order = current_order(build_order_if_necessary: true)
       authorize! :read, @order, cookies.signed[:guest_token]
       associate_user
       if params[:id] && @order.number != params[:id]
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -778,7 +778,9 @@ module Spree
     deprecate assign_default_credit_card: :add_default_payment_from_wallet, deprecator: Spree::Deprecation
 
     def record_ip_address(ip_address)
-      if last_ip_address != ip_address
+      if new_record?
+        self.last_ip_address = ip_address
+      elsif last_ip_address != ip_address
         update_column(:last_ip_address, ip_address)
       end
     end
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -31,17 +31,18 @@ module Spree
         # The current incomplete order from the guest_token for use in cart and during checkout
         def current_order(options = {})
           should_create = options[:create_order_if_necessary] || false
+          should_build = options[:build_order_if_necessary] || should_create
 
           return @current_order if @current_order
 
           @current_order = find_order_by_token_or_user(lock: options[:lock])
 
-          if should_create && (@current_order.nil? || @current_order.completed?)
+          if should_build && (@current_order.nil? || @current_order.completed?)
             @current_order = Spree::Order.new(new_order_params)
             @current_order.user ||= try_spree_current_user
             # See issue https://github.com/spree/spree/issues/3346 for reasons why this line is here
             @current_order.created_by ||= try_spree_current_user
-            @current_order.save!
+            @current_order.save! if should_create
           end
 
           if @current_order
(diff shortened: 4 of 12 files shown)


kennyadsl:
Please also read https://github.com/solidusio/solidus/issues/3467.

kennyadsl:
Anyway, I was trying to make something similar and the only reason why I don't like this change is that we are writing something into the database on a GET request, which is not ideal (and could be an easy endpoint to target for DoS attacks).

I think another valid way to handle this would be redirecting the user to the `root_path` if they are trying to see the cart without a `current_order` in place. Or maybe the empty cart is something that customers want to see anyway?

elia:
I think it's legitimate to have an empty cart page, and I actually saw it on many stores, usually with links to a product-list or with inline suggestions.

If we don't want to create an order upon visiting `/cart` it it's missing we I would suggest to let `current_order` learn how to initialize one. For example by accepting a `build_if_necessary` argument.

core/spec/lib/spree/core/controller_helpers/order_spec.rb
```diff
@@ -74,6 +75,23 @@ class FakesController < ApplicationController
         }.from(nil).to("0.0.0.0")
       end
     end
+
+    context 'build_order_if_necessary option is true' do
+      subject { controller.current_order(build_order_if_necessary: true) }
+
+      it 'creates new order' do
```
spaghetticode: since the order is not persisted, IMO this should read `it builds a new order` instead.

core/spec/lib/spree/core/controller_helpers/order_spec.rb
```diff
@@ -2,12 +2,13 @@
...
-  include Spree::Core::ControllerHelpers::Order
-end
-
 RSpec.describe Spree::Core::ControllerHelpers::Order, type: :controller do
-  controller(FakesController) {}
+  controller(ApplicationController) {
+    include Spree::Core::ControllerHelpers::Store
+    include Spree::Core::ControllerHelpers::Pricing
+    include Spree::Core::ControllerHelpers::Auth
+    include Spree::Core::ControllerHelpers::Order
```
spaghetticode: It's not evident to me why we need to include all these modules now, can you explain the reason?

elia: Previously there was the FakesController class that was shared among all controller helpers spec files. During the file loading (but before the specs were ran) each helper was including itself into that class. So, at the time the specs were ready to run, the FakesController class had all the helpers included. 

Now, instead, should be more clear on which helpers each helper depends, e.g. the `Order` helper needs methods from `Store`, `Pricing`, and `Auth`.

frontend/app/controllers/spree/orders_controller.rb
```diff
@@ -39,7 +39,7 @@ def update
 
     # Shows the current incomplete order from the session
     def edit
-      @order = current_order || Spree::Order.incomplete.find_or_initialize_by(guest_token: cookies.signed[:guest_token])
+      @order = current_order(build_order_if_necessary: true)
       authorize! :read, @order, cookies.signed[:guest_token]
       associate_user
```
bitberryru: Great work, thank you!
Coincidentally, I also researched current_order behaviour and found some interesting points in legacy code.
1. Remove reduntant `find_or_initialize_by` and just build order (you fixed it, great!)
2. Remove `associate_user` call in OrdersController#edit

That code was added in this commit https://github.com/spree/spree/commit/a3d1f9acdbd5f0ec6c5eceb07116e30cd953948b

In that version when you call `current_order(true)` you create new order without any attributes (without current user, without store_id, without currency, without guest_token and so on, check it https://github.com/spree/spree/blob/a3d1f9acdbd5f0ec6c5eceb07116e30cd953948b/core/lib/spree/core/current_order.rb#L27). The logic of order creation was divided - in one place you create empty order, at another place you set necessary attributes such as user_id via `associate_user` call. This mistake is still here for 8 years despite the fact that in current version necessary attributes are set at `current_order`.

This method call can be safely removed.

P.S. `associate_user` defined at this concern `Spree::Core::ControllerHelpers::Order` and called only at two places - this one (reduntant) and before any action at `Spree::CheckoutController`. Perhaps it is also unnecessary, I will investigate)

elia: @bitberryru thanks for the analysis and the hint!

So basically, if we inline `associate_user` inside `spree/orders#edit` we get this:

```rb
    # Shows the current incomplete order from the session
    def edit
      @order = current_order(build_order_if_necessary: true)
      authorize! :read, @order, cookies.signed[:guest_token]

      # vvv Inlined version of #associate_user 
      if try_spree_current_user && (@order.user.blank? || @order.email.blank?)
        @order.associate_user!(try_spree_current_user)
      end

      if params[:id] && @order.number != params[:id]
        flash[:error] = t('spree.cannot_edit_orders')
        redirect_to cart_path
      end
    end
```

The idea is that it's the responsibility of the authentication system to do the association in case the order was created before the login, and we can safely assume that we'll never encounter a persisted order that is authorized (via guest_token) and not already associated to the current user.

Here's the solidus_auth_devise warden callback:

```rb
# Merges users orders to their account after sign in and sign up.
Warden::Manager.after_set_user except: :fetch do |user, auth, _opts|
  if auth.cookies.signed[:guest_token].present?
    if user.is_a?(Spree::User)
      Spree::Order.incomplete.where(guest_token: auth.cookies.signed[:guest_token], user_id: nil).each do |order|
        order.associate_user!(user)
      end
    end
  end
end
```

---

About the checkout controller seems like the code is executed in a different sequence:

```rb
  class CheckoutController < Spree::StoreController
    before_action :load_order                   # <<<< calls #current_order
    around_action :lock_order

    before_action :ensure_order_is_not_skipping_states
    before_action :ensure_order_not_completed
    before_action :ensure_checkout_allowed
    before_action :ensure_sufficient_stock_lines
    before_action :ensure_valid_state

    before_action :associate_user               # <<<< calls #associate_user
    before_action :check_authorization          # <<<< checks against the guest_token
    before_action :apply_coupon_code
```

I'd like to more research to be sure not to break something for some store… 

@kennyadsl @bitberryru what about dealing with removing `#associate_user` in another PR?

kennyadsl: Sure, another PR is perfect for this change.

kennyadsl: Moved the discussion in a separate issue: https://github.com/solidusio/solidus/issues/3559

PR #3495, opened 2020-01-29
Improve sample data for the returned/reimbursed order

kennyadsl:
**Description**

In our sample data script, we are creating two orders. One of them is also returned and refunded in order to be able to see the reimbursement mailer preview, which requires an actual reimbursement in the database to work. 

The order is set in a way that allows the email preview to work but it's actually a bit messed up. In a real UX flow, it's not possible to have the order in that state, so I tried to adjust it to be as similar as possible with how a real refunded order looks like. 

This also improves the documentation since the sample code flow used can be a guideline to understand how the complex returns flow works. 

Ref: #1146

**Checklist:**

### DIFF
--- a/sample/db/samples/reimbursements.rb
+++ b/sample/db/samples/reimbursements.rb
@@ -2,21 +2,43 @@
 
 Spree::Sample.load_sample("orders")
 
-order          = Spree::Order.last
-inventory_unit = order.inventory_units.first
+order = Spree::Order.last
+inventory_unit = order.inventory_units.take!
 stock_location = inventory_unit.find_stock_item.stock_location
+return_reason = Spree::ReturnReason.active.take!
+preferred_reimbursement_type = Spree::ReimbursementType.where(name: 'Original').take!
 
-return_item = Spree::ReturnItem.create(inventory_unit: inventory_unit)
+# Mark the order paid and shipped
+order.payments.pending.each(&:complete)
+order.shipments.each do |shipment|
+  shipment.suppress_mailer = false
+  shipment.ship!
+end
 
-return_item.exchange_variant = return_item.eligible_exchange_variants.last
-return_item.build_exchange_inventory_unit
-return_item.accept!
-
-customer_return = Spree::CustomerReturn.create(
-  stock_location: stock_location,
-  return_items: [return_item]
+# Create a return authorization
+return_item = Spree::ReturnItem.new(
+  inventory_unit: inventory_unit,
+  preferred_reimbursement_type: preferred_reimbursement_type
 )
 
+order.return_authorizations.create!(
+  reason: return_reason,
+  return_items: [return_item],
+  stock_location: stock_location
+)
+
+# Create a customer return and mark it as received
+customer_return = Spree::CustomerReturn.create!(
+  return_items: [return_item],
+  stock_location: stock_location
+)
+return_item.reload
+return_item.skip_customer_return_processing = true
+return_item.receive!
+customer_return.process_return!
+
+# Accept the customer return and reimburse it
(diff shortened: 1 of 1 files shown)


sample/db/samples/reimbursements.rb
```diff
@@ -2,22 +2,46 @@
...
+# Create a return authorization
+return_item = Spree::ReturnItem.new(
+  inventory_unit: inventory_unit,
+  preferred_reimbursement_type: preferred_reimbursement_type
+)
 
-customer_return = Spree::CustomerReturn.create(
-  stock_location: stock_location,
-  return_items: [return_item]
+order.return_authorizations.create(
```
spaghetticode: If a new validation is added to the model then `#create` may fail silently, so what about using `#create!`?

kennyadsl: Done, and added the `!` to some other methods as well, thanks!

sample/db/samples/reimbursements.rb
```diff
@@ -2,21 +2,43 @@
 
 Spree::Sample.load_sample("orders")
 
-order          = Spree::Order.last
-inventory_unit = order.inventory_units.first
+order = Spree::Order.last
+inventory_unit = order.inventory_units.take!
```
vzqzac: Hey @kennyadsl I think we're not handling exceptions when this `takes!` method fails, should we still use the bang method?

sample/db/samples/reimbursements.rb
```diff
@@ -2,21 +2,43 @@
...
 stock_location = inventory_unit.find_stock_item.stock_location
+return_reason = Spree::ReturnReason.active.take!
+preferred_reimbursement_type = Spree::ReimbursementType.where(name: 'Original').take!
```
vzqzac: As far as I remember the `take` method expects the collection to contain at least one element, this will probably break if we don't have any `Spree::ReimbursementType` in the database

sample/db/samples/reimbursements.rb
```diff
@@ -38,8 +43,5 @@
 customer_return.process_return!
 
 # Accept the customer return and reimburse it
-return_item.accept!
-order.reimbursements.create(
-  customer_return: customer_return,
-  return_items: [return_item]
-)
+reimbursement = Spree::Reimbursement.build_from_customer_return(customer_return)
```
vzqzac: Thanks for the commit description! That made me know what the `build_from_customer_return` method does without having to do a lot of research

sample/db/samples/reimbursements.rb
```diff
@@ -7,6 +7,11 @@
 stock_location = inventory_unit.find_stock_item.stock_location
 return_reason = Spree::ReturnReason.active.take!
 preferred_reimbursement_type = Spree::ReimbursementType.where(name: 'Original').take!
+admin_user = if defined?(Spree::Auth)
+  Spree.user_class.admin.take!
+else
+  Spree.user_class.find_or_create_by!(email: 'admin@example.com')
+end
```
vzqzac: This is a clever way to set an user, but if we need an actual admin user, shouldn't we create it with more information than just the email?

sample/db/samples/reimbursements.rb
vzqzac: Ah nevermind, I just saw spaghetticode's comment and that solved my doubt here

sample/db/samples/reimbursements.rb
kennyadsl: This will be executed when `Auth::Devise` is not present, so it will create a LegacyUser which does not need any more information to be valid.

Issue #3472, opened 2020-01-09
Promotions that apply automatically shouldn't allow the user to create promo codes


aldesantis:
When I create a promotion that is automatically applied to all orders, I shouldn't be able to manage the promo codes for that promotion, since creating a code will result in the promotion getting into a broken state and failing validation.

**Solidus Version**:
2.9, but I guess it affects all versions

**To Reproduce**
1. Create a promotion that applies automatically
2. Click on View codes list
3. Click on Create promotion code
4. Create a new code

**Current behavior**
The promotion displays:
```
All orders will attempt to use this promotion

This promotion uses the promotion code: c
```
Furthermore, the promotion cannot be updated in any way, since it fails validation with the following error: "Apply Automatically Disallowed for promotions with a code".

**Expected behavior**
I shouldn't even be allowed to create the promo code.

**Screenshots**
<img width="485" alt="Screenshot 2020-01-09 14 57 06" src="https://user-images.githubusercontent.com/395073/72073498-56525e80-32f0-11ea-9bef-5808fc837682.png">
<img width="299" alt="Screenshot 2020-01-09 14 57 07" src="https://user-images.githubusercontent.com/395073/72073500-56525e80-32f0-11ea-8e89-39c759459331.png">

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

kennyadsl:
What about promotions that did already have promo codes and are set to `apply automatically` afterwards? If we add a validation, admin users will need to remove the promo codes before changing that.

I think we should try to change the logic so that this kind of promotion doesn't try to use the promo code at all and be clear in the Admin UI that if you check apply automatically promo codes will be ignored. 

Promotion code is very complex, though. If the above optimal (IMO) solution is not possible, what you proposed is a good trade-off.

MassimilianoLattanzio:
I think the best way is disallow the creation of promo code.

If the promo codes is ignored when "Apply to all orders" is checked, the operation of creation is useless so why not disable it?

But is also important notify the user that selecting "Apply to all orders", promo codes will be disabled. This can be done in Admin UI while creating a new promotion and also in editing.

Only for backward compatibility, we can change the logic so this kind of promotion ignore the promo code. Otherwise we can try a simpler solution such as create a command/task to delete the promo codes if some promotions applies automatically.

aldesantis:
[...]

PR #3502, opened 2020-02-02
Disable codes on apply automatically promo

MassimilianoLattanzio:
**Description**
Disabled creation of promotion code if a promotion applies automatically to all orders. Managed also old "apply-automatically" promotions with and without associated promotion codes.

Solve #3472

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/promotion_codes_controller.rb
+++ b/backend/app/controllers/spree/admin/promotion_codes_controller.rb
@@ -23,7 +23,12 @@ module Spree
 
       def new
         @promotion = Spree::Promotion.accessible_by(current_ability, :read).find(params[:promotion_id])
-        @promotion_code = @promotion.promotion_codes.build
+        if @promotion.apply_automatically
+          flash[:error] = t('activerecord.errors.models.spree/promotion_code.attributes.base.disallowed_with_apply_automatically')
+          redirect_to admin_promotion_promotion_codes_url(@promotion)
+        else
+          @promotion_code = @promotion.promotion_codes.build
+        end
       end
 
       def create
--- a/backend/app/views/spree/admin/promotions/_activations_new.html.erb
+++ b/backend/app/views/spree/admin/promotions/_activations_new.html.erb
@@ -4,7 +4,7 @@
     <div class="form-check">
       <label class="form-check-label">
         <%= radio_button_tag('activation_type', 'auto', activation_type == 'auto') %>
-        <%= t('.auto') %>
+        <%= t('.auto') %> <%= f.field_hint :promo_code_will_be_disabled %>
       </label>
     </div>
     <div class="form-check">
--- a/backend/spec/controllers/spree/admin/promotion_codes_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/promotion_codes_controller_spec.rb
@@ -29,4 +29,11 @@ describe Spree::Admin::PromotionCodesController do
     expect(flash[:error]).not_to be_nil
     expect(Spree::PromotionCode.where(promotion: promotion).count).to eql(3)
   end
+
+  it "can't create a new code on promotions that apply automatically" do
+    apply_automatically_promotion = create(:promotion, apply_automatically: true)
+    get :new, params: { promotion_id: apply_automatically_promotion.id }
+    expect(response).to redirect_to(spree.admin_promotion_promotion_codes_path(apply_automatically_promotion))
+    expect(flash[:error]).not_to be_nil
+  end
 end
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -32,7 +32,7 @@ module Spree
     validates :usage_limit, numericality: { greater_than: 0, allow_nil: true }
     validates :per_code_usage_limit, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
     validates :description, length: { maximum: 255 }
-    validate :apply_automatically_disallowed_with_codes_or_paths
+    validate :apply_automatically_disallowed_with_paths
 
     before_save :normalize_blank_values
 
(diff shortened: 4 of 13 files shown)


backend/app/controllers/spree/admin/promotion_codes_controller.rb
```diff
@@ -23,7 +23,12 @@ def index
 
       def new
         @promotion = Spree::Promotion.accessible_by(current_ability, :read).find(params[:promotion_id])
-        @promotion_code = @promotion.promotion_codes.build
+        if @promotion.apply_automatically
```
kennyadsl: What about also performing this check on the create? I know we have the model validation there but I think it's quite useless to build promotion codes if we already know they are not valid. Additionally, we already have this code in the same controller that could easily become a `before_action` for both. What do you think?

backend/app/views/spree/admin/promotions/_activations_new.html.erb
```diff
@@ -4,7 +4,7 @@
     <div class="form-check">
       <label class="form-check-label">
         <%= radio_button_tag('activation_type', 'auto', activation_type == 'auto') %>
-        <%= t('.auto') %>
+        <%= t('.auto') %> <small class="text-muted">(<%= t('.auto_warning') %>)</small>
```
kennyadsl: Maybe we could use pills for that?

core/app/models/spree/promotion.rb
```diff
@@ -257,10 +257,9 @@ def match_all?
       match_policy == "all"
     end
 
-    def apply_automatically_disallowed_with_codes_or_paths
+    def apply_automatically_disallowed_with_paths
       return unless apply_automatically
 
-      errors.add(:apply_automatically, :disallowed_with_code) if codes.any?
```
kennyadsl: Can you please expand a little bit more in its commit description? I'm not sure to get why this is needed.

core/config/locales/en.yml
```diff
@@ -865,6 +864,7 @@ en:
           single_code_html: 'This promotion uses the promotion code: <code>%{code}</code>'
         activations_new:
           auto: Apply to all orders
+          auto_warning: Promo codes will be disabled
```
kennyadsl: This message is not very informative, we should expand a little bit more.

core/app/models/spree/promotion_code.rb
```diff
@@ -37,6 +38,10 @@ def usage_limit
     promotion.per_code_usage_limit
   end
 
+  def promotion_not_apply_automatically
+    errors.add(:base, I18n.t('spree.errors.messages.could_not_create_promotion_code_on_apply_automatically_promotion')) if promotion.apply_automatically
```
kennyadsl: There's no need to use a generic key here, you can have the same result by using:

```ruby
errors.add(:base, :disallaowed_with_apply_automatically) if promotion.apply_automatically
```
and moving the translation into:

```
en.activerecord.errors.models.spree/promotion_code.attributes.base.disallaowed_with_apply_automatically
```

(this probably needs a double check on documentation 🙂)

backend/spec/controllers/spree/admin/promotion_codes_controller_spec.rb
```diff
@@ -29,4 +29,11 @@
     expect(flash[:error]).not_to be_nil
     expect(Spree::PromotionCode.where(promotion: promotion).count).to eql(3)
   end
+
+  it "can't create a new code on promotion that apply automatically" do
```
kennyadsl: `promotion` -> `promotions`

core/config/locales/en.yml
MassimilianoLattanzio: I chose a short text because of an error in UI with long texts:
Schermata 2020-02-06 alle 11 48 50

Can I modify the layout to enlarge that col? Or maybe I can modify the pill's css to disable white-space nowrap

kennyadsl: I think we need this component there:

Screenshot 2020-02-06 at 11 59 17

It should be called `hint`, my fault!

core/app/models/spree/promotion_code.rb
MassimilianoLattanzio: Do you mean the solidus developers guide? In that guide there isn't any section about the promotion code, maybe I can create a new page  for the promotion codes.

backend/app/controllers/spree/admin/promotion_codes_controller.rb
MassimilianoLattanzio: Yep you're absolutely right, I think is the best way

core/app/models/spree/promotion.rb
MassimilianoLattanzio: I've added a short message to explain why this commit is needed, let me know if is clear enough

MassimilianoLattanzio: @kennyadsl did you read the new commit message? Let me know what do you think

core/app/models/spree/promotion_code.rb
kennyadsl: > this probably needs a double check on documentation

No, I meant in the Rails documentation.

core/app/models/spree/promotion.rb
kennyadsl: Thanks for updating it, I think the second part can be rewritten as:

```
Also, this improves backward compatibility since if we keep validating
against promotions with codes as well, it will be impossible to update 
existing apply-auto promotions that already have a code associated.
```

While in the first part I still don't get this sentence:

```
Since promotions creation is a two-step process and promo codes
are added in the second step, the validation have to be there.
```

Why do you say that the validation has to be there if we are removing it in this commit?

kennyadsl:
@MassimilianoLattanzio can you please fixup some commits together where it makes sense?

core/app/models/spree/promotion.rb
MassimilianoLattanzio: I modified the commit description to explain why we need to remove that validation and move it in the promo code creation step, let me know if now is ok

core/app/models/spree/promotion_code.rb
MassimilianoLattanzio: Sorry, my fault, I misunderstood 🙂

MassimilianoLattanzio:
> @MassimilianoLattanzio can you please fixup some commits together where it makes sense?

Done 👍

PR #3504, opened 2020-02-04
Ensure payment methods are ordered correctly

AlistairNorman:
**Description**

Previously we were not respecting the payment method order in the admin. This orders them and tests that they are being ordered properly.

### DIFF
--- a/backend/app/controllers/spree/admin/payments_controller.rb
+++ b/backend/app/controllers/spree/admin/payments_controller.rb
@@ -80,11 +80,11 @@ module Spree
 
       def load_data
         @amount = params[:amount] || load_order.total
-        @payment_methods = Spree::PaymentMethod.active.available_to_admin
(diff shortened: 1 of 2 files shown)


backend/app/controllers/spree/admin/payments_controller.rb
```diff
@@ -80,11 +80,11 @@ def object_params
 
       def load_data
         @amount = params[:amount] || load_order.total
-        @payment_methods = Spree::PaymentMethod.active.available_to_admin
+        @payment_methods = Spree::PaymentMethod.active.available_to_admin.ordered_by_position
         if @payment && @payment.payment_method
           @payment_method = @payment.payment_method
         else
-          @payment_method = @payment_methods.first
+          @payment_method = @payment_methods.ordered_by_position.first
```
jarednorman: Hmm, this is already ordered by position, since it's using `@payment_methods` already. I believe it should be fine as it was.

PR #3505, opened 2020-02-04
Remove conditional code that targets Rails 5.1

kennyadsl:
**Description**

🧹🧹🧹🧹

We still have some conditional code that targets Rails 5.1, which has been deprecated and we do not support anymore. 

This PR removes that code in order to keep the codebase as clean as possible.

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -52,6 +52,7 @@ module DummyApp
     config.whiny_nils = true
     config.consider_all_requests_local = true
     config.action_controller.allow_forgery_protection = true
+    config.action_controller.default_protect_from_forgery = true
     config.action_controller.perform_caching = false
     config.action_dispatch.show_exceptions = false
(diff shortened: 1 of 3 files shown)


core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -12,14 +12,11 @@
...
 
-RAILS_52_OR_ABOVE = Gem::Version.new(Rails.version) >= Gem::Version.new('5.2')
 RAILS_6_OR_ABOVE = Gem::Version.new(Rails.version) >= Gem::Version.new('6.0')
 
 # @private
 class ApplicationController < ActionController::Base
-  unless RAILS_52_OR_ABOVE
-    protect_from_forgery with: :exception
-  end
+  protect_from_forgery with: :exception
```
spaghetticode: This code was executed only when `RAILS_52_OR_ABOVE` was false, so I think the line should be removed

kennyadsl: You are right!

PR #3507, opened 2020-02-05
Fix reimbursement_category_name ending up in a translation missing error

elia:
**Description**

While debugging the reason for which after the upgrade to Solidus 2.9 an application was still marking store credit with the "Gift Card" store credit category, I found that `Spree::StoreCreditCategory.reimbursement_category_name` contained the string `"translation missing: en.spree.store_credit_category.default"`, although the I18n key was already set to "Reimbursement".

The problem is that the I18n lookup is done before the Rails app is able to load the application translations.

**Checklist:**

### DIFF
--- a/core/app/models/spree/reimbursement.rb
+++ b/core/app/models/spree/reimbursement.rb
@@ -163,6 +163,23 @@ module Spree
       perform!(created_by: created_by)
     end
 
+    # The returned category is used as the category
+    # for Spree::Reimbursement::Credit.default_creditable_class.
+    #
+    # @return [Spree::StoreCreditCategory]
+    def store_credit_category
+      if Spree::Config.use_legacy_store_credit_reimbursement_category_name
+        Spree::Deprecation.warn("Using the legacy reimbursement_category_name is deprecated. "\
+          "Set Spree::Config.use_legacy_store_credit_reimbursement_category_name to false to use "\
+          "the new version instead.", caller)
+
+        name = Spree::StoreCreditCategory.reimbursement_category_name
+        return Spree::StoreCreditCategory.find_by(name: name) || Spree::StoreCreditCategory.first
+      end
+
+      Spree::StoreCreditCategory.find_by(name: Spree::StoreCreditCategory::REIMBURSEMENT)
+    end
+
     private
 
     def calculate_total
--- a/core/app/models/spree/store_credit_category.rb
+++ b/core/app/models/spree/store_credit_category.rb
@@ -1,18 +1,45 @@
 # frozen_string_literal: true
 
 class Spree::StoreCreditCategory < Spree::Base
+  GIFT_CARD = 'Gift Card'
+  REIMBURSEMENT = 'Reimbursement'
+
   class_attribute :non_expiring_credit_types
   self.non_expiring_credit_types = [Spree::StoreCreditType::NON_EXPIRING]
 
+  # @deprecated
   class_attribute :reimbursement_category_name
   self.reimbursement_category_name = I18n.t('spree.store_credit_category.default')
 
-  def self.reimbursement_category(_reimbursement)
-    Spree::StoreCreditCategory.find_by(name: reimbursement_category_name) ||
-      Spree::StoreCreditCategory.first
+  # @deprecated
+  def self.reimbursement_category(reimbursement)
+    reimbursement.store_credit_category
   end
 
   def non_expiring?
     self.class.non_expiring_credit_types.include? name
+  end
+
+  public_instance_methods.grep(/^reimbursement_category_name/).each do |method|
+    deprecate(
+      method => 'Use Spree::Reimbursement#store_credit_category.name instead',
+      deprecator: Spree::Deprecation
+    )
+  end
+
+  class << self
+    public_instance_methods.grep(/^reimbursement_category_name/).each do |method|
+      deprecate(
+        method => 'Use Spree::Reimbursement.store_credit_category.name instead',
+        deprecator: Spree::Deprecation
+      )
+    end
+
+    public_instance_methods.grep(/^reimbursement_category$/).each do |method|
+      deprecate(
+        method => 'Use Spree::Reimbursement.store_credit_category instead',
+        deprecator: Spree::Deprecation
+      )
+    end
   end
 end
--- a/core/spec/models/spree/reimbursement_spec.rb
+++ b/core/spec/models/spree/reimbursement_spec.rb
@@ -275,4 +275,30 @@ RSpec.describe Spree::Reimbursement, type: :model do
       expect { subject }.to change { reimbursement.refunds.count }.by(1)
     end
   end
+
+  describe '#store_credit_category' do
+    let(:reimbursement) { create(:reimbursement) }
+
+    before do
+      create(:store_credit_category, name: 'foo')
+      create(:store_credit_category, :reimbursement)
+    end
+
+    context 'when using the legacy version' do
+      before do
+        stub_spree_preferences(use_legacy_store_credit_reimbursement_category_name: true)
+      end
+
+      it 'issues a deprecation warning and returns the first category created' do
+        expect(Spree::StoreCreditCategory).to receive(:reimbursement_category_name)
+        expect(Spree::Deprecation).to receive(:warn)
+        expect(reimbursement.store_credit_category.name).to eq('foo')
+      end
+    end
+
+    it 'fetches the the default reimbursement store category' do
+      expect(Spree::Config.use_legacy_store_credit_reimbursement_category_name).to eq(false)
+      expect(reimbursement.store_credit_category.name).to eq('Reimbursement')
+    end
+  end
 end
--- a/core/app/models/spree/reimbursement_type/reimbursement_helpers.rb
+++ b/core/app/models/spree/reimbursement_type/reimbursement_helpers.rb
@@ -61,7 +61,7 @@ module Spree
       Spree::Reimbursement::Credit.default_creditable_class.new(
         user: reimbursement.order.user,
         amount: unpaid_amount,
-        category: Spree::StoreCreditCategory.reimbursement_category(reimbursement),
+        category: reimbursement.store_credit_category,
         created_by: created_by,
         memo: "Refund for uncreditable payments on order #{reimbursement.order.number}",
         currency: reimbursement.order.currency
--- a/core/db/default/spree/store_credit.rb
+++ b/core/db/default/spree/store_credit.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-Spree::StoreCreditCategory.find_or_create_by!(name: I18n.t('spree.store_credit_category.default'))
 
 Spree::PaymentMethod.create_with(
   name: "Store Credit",
@@ -18,6 +17,7 @@ Spree::StoreCreditType.create_with(priority: 2).find_or_create_by!(name: Spree::
 Spree::ReimbursementType.create_with(name: "Store Credit").find_or_create_by!(type: 'Spree::ReimbursementType::StoreCredit')
 Spree::ReimbursementType.create_with(name: "Original").find_or_create_by!(type: 'Spree::ReimbursementType::OriginalPayment')
 
-Spree::StoreCreditCategory.find_or_create_by!(name: 'Gift Card')
+Spree::StoreCreditCategory.find_or_create_by!(name: Spree::StoreCreditCategory::GIFT_CARD)
+Spree::StoreCreditCategory.find_or_create_by!(name: Spree::StoreCreditCategory::REIMBURSEMENT)
 
 Spree::StoreCreditReason.find_or_create_by!(name: 'Credit Given In Error')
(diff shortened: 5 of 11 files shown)


spaghetticode:
After some investigation with @elia we discovered the problem is generated when there's a decorator for the class `Spree::StoreCreditCategory` defined in the store code. 

Basically the issue happens because decorator files are loaded before [I18n](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/i18n_railtie.rb) translations are completely initialized.

kennyadsl:
@spaghetticode not sure if you are saying that this not needed anymore and it's an issue of how you are decorating files in your application. If it's needed we can take a look at the [`default` option of `class_attribute`](https://apidock.com/rails/Class/class_attribute) but I think it does the exact same thing and won't solve the issue.

spaghetticode:
@kennyadsl I think the fix is still needed, it happens to every store that uses a `_decorator.rb` file to decorate `Spree::StoreCreditCategory`.
 I agree that using the `default` option will not solve the issue, though I did not try it... just guessing. 

Elia's solution works great but is specific for this situation; I'd prefer a more generic solution that would work for all instances where this problem could happen, both present and future. 

One solution that comes to my mind  would be to load decorators after loading I18n translations... but not sure if this is feasible, so what we have now may be a good compromise... @elia any thoughts on this?

elia:
@kennyadsl @spaghetticode I would avoid setting a value (unless static) until after the initialization, can be this way or via an actual initializer. So the options I see are:

- Lazily assign as it's done in this PR
- Move the default value to an initializer
- Use some sort of code instead of a display value like name to fetch the record
- Turn the configuration value into a proc that can be easily loaded and fully customizable without resorting to decorators, that BTW can be used to configure not just the name but the whole fetching strategy.

Since I also object to falling back to `.first` if no category is found I would say that the last option is superior because it can easily default to the current behavior and scale up to very thick code for fetching the reimbursement category.

kennyadsl:
@elia maybe we can just redefine the getter? This way the preference would work with the default value no matter where you call it?

elia:
@kennyadsl I changed route altogether and moved the responsibility of fetching the store credit category for a reimbursement to the reimbursement itself via `Spree::Reimbursement#store_credit_category` that in turn relies on `Spree::Reimbursement.store_credit_category_proc`.

Hope this makes more sense and looks cleaner (especially once we'll throw away the deprecated stuff).

core/app/models/spree/reimbursement.rb
```diff
@@ -59,6 +59,20 @@ class IncompleteReimbursementError < StandardError; end
     class_attribute :reimbursement_failure_hooks
     self.reimbursement_failure_hooks = []
 
+    # The reimbursement_category_proc allows to fetch a specific reimbursement
```
aldesantis, line 62: Looks like the name of the option needs to be updated?

core/app/models/spree/reimbursement.rb
```diff
@@ -59,6 +59,20 @@ class IncompleteReimbursementError < StandardError; end
...
+    # category for a given reimbursement, the fetched category is used as the
+    # category passed to Spree::Reimbursement::Credit.default_creditable_class.
+    #
+    # @param proc A proc that will accept a single "reimbursement" argument
+    class_attribute :store_credit_category_proc
+    self.store_credit_category_proc = ->(_reimbursement) {
```
aldesantis, line 68: Why is this not defined on `Spree::StoreCreditCategory` instead? It looks like a better place to do it, since the method returns an instance of that class.

core/app/models/spree/store_credit_category.rb
```diff
@@ -4,12 +4,23 @@ class Spree::StoreCreditCategory < Spree::Base
...
+    reimbursement.store_credit_category
+  end
+
+  deprecate :reimbursement_category_name, deprecator: Spree::Deprecation
+  deprecate :reimbursement_category_name=, deprecator: Spree::Deprecation
+  deprecate :reimbursement_category_name?, deprecator: Spree::Deprecation
+  class << self
+    deprecate :reimbursement_category_name, deprecator: Spree::Deprecation
+    deprecate :reimbursement_category_name=, deprecator: Spree::Deprecation
+    deprecate :reimbursement_category, deprecator: Spree::Deprecation
```
aldesantis, line 23: Can you add a message explaining what should be used instead? Otherwise, users may think these have been deprecated with no replacement, unless they read the documentation.

kennyadsl:
@elia did you have some time to take a look at the comments @aldesantis left? 🙏

filippoliverani:
> @elia did you have some time to take a look at the comments @aldesantis left? 🙏

I agreed with @elia to take on the PR, I'm pushing some changes to address @aldesantis' comments 👍

core/app/models/spree/reimbursement.rb
filippoliverani: I think that could work in both classes, it looks like a matter of personal taste. I'd keep the new logic in `Spree::Reimbursement` to maintain @elia's intent of simplyfing its use:

> I changed route altogether and moved the responsibility of fetching the store credit category for a reimbursement to the reimbursement itself via `Spree::Reimbursement#store_credit_category` that in turn relies on `Spree::Reimbursement.store_credit_category_proc`.
> 
> Hope this makes more sense and looks cleaner (especially once we'll throw away the deprecated stuff).

aldesantis:
@elia @filippoliverani this looks good overall. Could you guys just squash the commits? 🙏

elia:
> @elia @filippoliverani this looks good overall. Could you guys just squash the commits? 🙏

@aldesantis done ✅

aldesantis (approved):

core/app/models/spree/reimbursement.rb
```diff
@@ -59,6 +59,20 @@ class IncompleteReimbursementError < StandardError; end
...
     self.reimbursement_failure_hooks = []
 
+    # The store_credit_category_proc allows to fetch a specific store credit
+    # category for a given reimbursement, the fetched category is used as the
+    # category passed to Spree::Reimbursement::Credit.default_creditable_class.
+    #
+    # @param proc A proc that will accept a single "reimbursement" argument
+    class_attribute :store_credit_category_proc
+    self.store_credit_category_proc = ->(_reimbursement) {
+      name = Spree::Deprecation.silence { Spree::StoreCreditCategory.reimbursement_category_name }
```
kennyadsl, line 69: I'd try to avoid this `silence` since it's not thread-safe yet, see https://github.com/rails/rails/pull/37590 which is still only in Rails master. 

This means that if another thread should emit a deprecation at the same time, it will be silenced. @cedum recently worked to a PR to remove all `silence` from our codebase here https://github.com/solidusio/solidus/pull/3753 and I'd keep going in that direction.

kennyadsl (commented):
I generally dislike silencing deprecations outside of the specs. What if the method we are silencing calls another deprecated method? 

Since it turned out that `silence` is not yet thread-safe, I like it even less. 

I'd prefer to explore an alternative.

core/app/models/spree/reimbursement.rb
cedum: Agree, the biggest issue with silence is that, being a class attribute, it might leave `silence` enabled globally after some time running in a threaded environment. Here's a dumb test I've done some time ago: https://gist.github.com/cedum/8315ebdd59781ee7fcc7f452a405542f

elia: Do you have any example of a non-invasive (or just decent) way of going around the call to `silence`? 
Everything I can think of is quite ugly but I'd like to fix it.

kennyadsl: I don't think there's a way, we should just not use deprecated code at all, maybe reviewing the approach we used in this PR.

kennyadsl: I tried locally, I think there are ways to avoid to use procs, like using a preference as @elia suggested at some point. But maybe the best thing to do is remove the hardcoding at all and add a field to the database in the StoreCreditCategory to indicate which one should be used for the reimbursements instead? 
At that point, we can just deprecate setting and getting those class attributes. WDYT?

elia: @kennyadsl I pushed another commit with some changes, if those are ok I'll squash the two commits and we're good to go ✨

elia:
(The rails-master failure is unrelated)

kennyadsl:
@elia I'll take a look tomorrow ™️ 😆  , thanks!

core/app/models/spree/reimbursement.rb
```diff
@@ -59,6 +59,13 @@ class IncompleteReimbursementError < StandardError; end
     class_attribute :reimbursement_failure_hooks
     self.reimbursement_failure_hooks = []
 
+    # The legacy_store_credit_category_name allows to control whether the legacy
+    # way of fetching the category should be used.
+    #
+    # @param [Boolean] enable/disable the legacy way of fetching the store category name
+    class_attribute :legacy_store_credit_category_name
+    self.legacy_store_credit_category_name = false
```
kennyadsl, line 67: we should start with this preference set true, so people using it has the chance to be notified that we are going to switch the value in Solidus 3.0.

core/app/models/spree/store_credit_category.rb
```diff
@@ -1,18 +1,47 @@
...
 
 class Spree::StoreCreditCategory < Spree::Base
+  GIFT_CARD = 'Gift Card'
+  REIMBURSEMENT = 'Reimbursement'
+
   class_attribute :non_expiring_credit_types
   self.non_expiring_credit_types = [Spree::StoreCreditType::NON_EXPIRING]
 
+  # @deprecated
+  # Please use Spree::Reimbursement.store_credit_category_proc= instead.
```
kennyadsl, line 11: We should remove all `store_credit_category_proc` leftovers.

kennyadsl (commented):
Left a couple of things, thanks @elia!

core/app/models/spree/reimbursement.rb
elia: I finally moved it to an app config, that way support for testing and otherwise was much better

DanielePalombo:
This is the changelog provided from @elia 

**Fixed how the default reimbursement store-credit category is fetched**
Before this change the store-credit category for reimbursement was fetched by name using a missing translation (i.e. `en.spree.store_credit_category.default`) that resulted in the name "Default". If no category was found the code fell back on the first category from the database, which wasn't guaranteed to be the right one. Trying to update the translation to the desired category name was also useless due to how code was loaded.
Now it's possible to disable this legacy behavior and switch to a simpler one, in which the code will look for a CreditCategory named "Reimbursement". Here's a list of checks and fixes you can perform to ensure you can enable the correct implementation:
- If you don't use reimbursements you're good to go, skip below to *Disabling the legacy behavior*
- Ensure you din't decorate or patched any of the involved code, especially:
   - `Spree::StoreCreditCategory.reimbursement_category`
   - `Spree::StoreCreditCategory.reimbursement_category_name`
- Ensure your "production" environment is already returning the correct category, you can assess that by running this in your "production" console: `Spree::StoreCreditCategory.reimbursement_category(nil).name == "Reimbursement"`
*Disabling the legacy behavior*
If everything is sound, or you are ok with a different category name for newly created reimbursement store credits you can switch to the new behavior by configuring this Solidus preference in your `spree.rb` initializer:
```ruby
Spree.config do |config|
  config.use_legacy_store_credit_reimbursement_category_name = false
end
```
If you had modifications in your codebase consider disabling the legacy behavior and porting them to a simple overwrite of `Spree::Reimbursement#store_credit_category`. 
The legacy behavior will be removed in the next major version of Solidus.

kennyadsl (approved):

core/app/models/spree/reimbursement.rb
```diff
@@ -163,6 +163,23 @@ def return_all(created_by: nil)
       perform!(created_by: created_by)
     end
 
+    # The returned category is used as the category
+    # for Spree::Reimbursement::Credit.default_creditable_class.
+    #
+    # @return [Spree::StoreCreditCategory]
+    def store_credit_category
```
jarednorman, line 170: Why is this an instance method? It doesn't look like it relies on the instance in any way. Is it so that it can be overridden by people looking to dynamically determine the store credit category.

elia: > Is it so that it can be overridden by people looking to dynamically determine the store credit category.

Yes, exactly. The original implementation was accepting a reimbursement to allow overrides based on it (although the implementation didn't actually use it) and I didn't want to break the promise if anyone did do that.

jarednorman: 👍 Right, that makes sense.

kennyadsl:
@aldesantis @spaghetticode can you please review again when you have time?

core/spec/models/spree/reimbursement_spec.rb
```diff
@@ -275,4 +275,30 @@
       expect { subject }.to change { reimbursement.refunds.count }.by(1)
     end
   end
+
+  describe '#store_credit_category' do
+    let(:reimbursement) { create(:reimbursement) }
+
+    before do
+      create(:store_credit_category, name: 'foo')
+      create(:store_credit_category, name: Spree::StoreCreditCategory::REIMBURSEMENT)
```
spaghetticode, line 284: I think you can use the new `reimbursement` trait here

spaghetticode (changes requested):
Code looks good to me, I left a note for a possible improvement, but I'm blocking because the last commit needs to be squashed.

jarednorman (approved):

spaghetticode (approved):

spaghetticode:
@elia no need to squash the second commit, looks legit 👍 
Thank you!

PR #3508, opened 2020-02-06
Only log basic response information

JDutil:
**Description**

When logging a gateway error response there is a params field,
which provides extra information about the request. That information
can include useful information like the payment amount, but it can
also include PII such as a full billing address.

By logging the full error response in yaml that PII can end up in your
logs, which is not desirable and potentially against the law. Therefore
we should only log the minimal information needed from the response.

**Checklist:**

### DIFF
--- a/core/spec/models/spree/payment_spec.rb
+++ b/core/spec/models/spree/payment_spec.rb
@@ -38,7 +38,12 @@ RSpec.describe Spree::Payment, type: :model do
   end
 
   let(:failed_response) do
-    ActiveMerchant::Billing::Response.new(false, '', {}, {})
+    ActiveMerchant::Billing::Response.new(
+      false,
+      'Declined',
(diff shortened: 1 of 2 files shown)


core/spec/models/spree/payment_spec.rb
```diff
@@ -387,6 +403,14 @@
           expect { payment.purchase! }.to raise_error(Spree::Core::GatewayError)
           expect(payment.capture_events.count).to eq(0)
         end
+
+        it "should not log response params" do
```
spaghetticode: would it be possible to write a shared example with this code? Basically, the whole test except for one single line that differs (`payment.purchase`, `#capture!` and `#authorize!`) is repeated 3 times.

JDutil: Done 👍

Issue #3481, opened 2020-01-21
Spree::Event Support regex/wildcard


skukx:
**Is your feature request related to a problem? Please describe.**

Right now it is not possible to subscribe to all spree events:

```ruby

Spree::Event.subscribe('*') { Analytics.report(event) }

# Or

Spree::Event.subscribe(/.*/) { |event| Analytics.report(event) }
```

You can do a workaround and go directly through `ActiveSupport::Notifications` or whatever the interface but should probably be supported through spree events.

**Describe the solution you'd like**

The wildcard is going to specific to the adapter being used so whatever the final name becomes will need to be parsed at the adapter level.

For example: In order to match multiple events with `ActiveSupport::Notifications`, a regex object is required. However with `Redis`, a string is sufficient.

Redis: `'*.spree'`
Rails: `/.*\.spree$/`

**Additional context**

skukx:
One of the obstacles I'm noticing with this is the [`Spree::Event.name_with_suffix`](https://github.com/solidusio/solidus/blob/master/core/lib/spree/event.rb#L109-L111) method.

The method checks to see if the suffix is already contained, if it isn't then appends the suffix.

However when looking at a regexp object, this gets messy to if the suffix is included in the regexp.

Is it ok to exclude the suffix and expect users to include it in their regexp?

skukx:
## Option 1
```ruby
module Spree::Event
    # ...

    # Resolve Event Name with suffix
    #
    # @param [String, Regexp] name The event name
    # @return [String, Regexp]
    #
    # @example Get name with suffix
    #   self.name_with_suffix('foo')
    #   #=> 'foo.spree'
    #
    # @example Get name with included suffix
    #   self.name_with_suffix('foo.spree')
    #   #=> 'foo.spree'
    #
    # @example Get name with suffix using regexp
    #   self.name_with_suffix(/foo/)
    #   #=> /foo\.spree$/
    #
    # @example When using regexp, don't include the suffix
    #   self.name_with_suffix(/foo\.spree$/)
    #   #=> /foo\.spree$\.spree$/
    #
    def name_with_suffix(name)
      if name.is_a?(Regexp)
        return Regexp.new(name.source + suffix_as_regexp.source)
      end

      name.end_with?(suffix) ? name : [name, suffix].join
    end

    def suffix_as_regexp
      /#{Regexp.quote(suffix)}$/
    end

    # ...
end
```

skukx:
## Option 2

Move logic to adapter

[...]

PR #3512, opened 2020-02-13
Allow multiple events subscription with regexp

spaghetticode:
Ref https://github.com/solidusio/solidus/issues/3481

`Spree::Event` events can now be subscribed also using regexp event names, when using the default  ActiveSupportNotification adapter.
As ActiveSupportNotification can manage regexp when subscribing to events, it makes sense to allow `Spree::Event` to leverage this feature.

Every specific adapter can decide independently their supported event name types. 

**Checklist:**

### DIFF
--- a/core/lib/spree/event/adapters/active_support_notifications.rb
+++ b/core/lib/spree/event/adapters/active_support_notifications.rb
@@ -4,6 +4,8 @@ module Spree
   module Event
     module Adapters
       module ActiveSupportNotifications
+        class InvalidEventNameType < StandardError; end
+
         extend self
 
         def fire(event_name, opts)
@@ -31,11 +33,25 @@ module Spree
         end
 
         # Normalizes the event name according to this specific adapter rules.
-        # @param [String, Symbol] event_name the event name, with or without the
-        #   .spree" suffix)
+        #  When the event name is a string or a symbol, if the suffix is missing, then
+        #  it is added automatically.
+        #  When the event name is a regexp, due to the huge variability of regexps, adding
+        #  or not the suffix is developer's responsibility (if you don't, you will subscribe
+        #  to all internal rails events as well).
+        #  When the event type is not supported, an error is raised.
+        #
+        # @param [String, Symbol, Regexp] event_name the event name, with or without the
+        #  suffix (Spree::Config.events.suffix defaults to `.spree`).
         def normalize_name(event_name)
-          name = event_name.to_s
-          name.end_with?(suffix) ? name : [name, suffix].join
+          case event_name
+          when Regexp
+            event_name
+          when String, Symbol
+            name = event_name.to_s
+            name.end_with?(suffix) ? name : [name, suffix].join
+          else
+            raise InvalidEventNameType, "Invalid event name type: #{event_name.class}"
+          end
         end
 
         # The suffix used for namespacing event names, defaults to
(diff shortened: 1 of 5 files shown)


core/lib/spree/event/adapters/active_support_notifications.rb
```diff
@@ -29,6 +31,17 @@ def listeners_for(names)
             memo[name] = listeners if listeners.present?
           end
         end
+
+        def name_with_suffix(name, suffix)
+          case name
+          when String
+            name.end_with?(suffix) ? name : [name, suffix].join
+          when Regexp
+            Regexp.new("#{name}#{Regexp.escape(suffix)}$")
```
skukx: We should let it be known that different behavior can happen based on the type here:

For example:
```ruby
name_with_suffix('order_finalized.spree', 'spree')
#=> 'order_finalized.spree'

# But this differs if regex is used

name_with_suffix(/order_finalized\.spree/, 'spree')
#=> /order_finalized\.spree\.spree$/
```

The method tries to check if spree is included in the string case. if it isn't then adds it.
However in the case of regexp, adds it no matter what.

skukx: Would it be best to deprecate the explicit inclusion of `suffix` in the event name and always append it in `name_with_suffix`?

spaghetticode: > We should let it be known that different behavior can happen based on the type

Sure. 
BTW I think the different behavior for String and Regexp event names is not much of a real problem, `String` and `Regexp` are different things, so some different behavior can be tolerated IMO.

>Would it be best to deprecate the explicit inclusion of suffix in the event name and always append it in name_with_suffix?

mmm not sure I understand what you mean here, can you explain with more detail?

skukx: Sorry for the confusion, what I mean is instead of conditionally appeneding the suffix, we always append the suffix.

```ruby
# Conditional
name_with_suffix('my_event.spree')
#=> 'my_event.spree'
```

Changes to:

```ruby
# Always
name_with_suffic('my_event.spree')
#=> 'my_event.spree.spree'
```

This would allow the same behaviour to occur for both types passed in since it is expected that spree will be appended always.

I'm ok with the current code as is though, so this is just for discussion

spaghetticode: Thanks for the clarification. I think both approaches have merits, but I am rather undecided about which one is better.  Maybe somebody has more strong opinions on this, but IMO it does not change really much in the end.

kennyadsl: I also don't like that the behavior is different if you pass different things to this method. I prefer to make it explicit by not calling `name_with_suffix` at all if a regex is passed + changing YARD documentation for the `subscribe` method and the guides' text explaining how to use this feature.

spaghetticode:
While working on this PR I noticed the illogic situation exposed in https://github.com/solidusio/solidus/pull/3519. I think that PR needs attention and to be merged before this one.

core/spec/lib/spree/event/adapters/active_support_notifications_spec.rb
```diff
@@ -17,6 +17,22 @@ module Adapters
...
             end
           end
+
+         context "when event name is a regexp" do
+            let(:event_name) { /.*/ }
+
+            it "never changes the regexp" do
+              expect(subject).to eq event_name
+            end
+          end
```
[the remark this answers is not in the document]
spaghetticode: They're now aligned. Hound didn't notice the change.

core/lib/spree/event/adapters/active_support_notifications.rb
kennyadsl: See https://github.com/solidusio/solidus/pull/3519 for resolution of this comment.

PR #3513, opened 2020-02-13
Memoize Spree::User#wallet method

AlessioRocco:
**Description**

We are already using memoization on some `Spree::Order` methods:

- https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L263-L265
- https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L294-L296
- https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L298-L300
- https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L302-L304

and `Spree::User#wallet` seems a good candidate to memoize.
 
Without memoization every time someone calls this method a new instance of `Spree::Wallet` is created but I think we should interact ever with the same instance, It seems logical to me that a user has just one wallet, this way the code reflect the reality and It also improves performance and memory usage.

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/user_methods.rb
+++ b/core/app/models/concerns/spree/user_methods.rb
@@ -39,7 +39,7 @@ module Spree
     end
 
     def wallet
-      Spree::Wallet.new(self)
+      @wallet ||= Spree::Wallet.new(self)
     end
 
     # has_spree_role? simply needs to return true or false whether a user has a role or not.


kennyadsl:
@AlessioRocco I don't get why this is legit, can you expand a little bit more. Also, it would be helpful to understand what's wrong with the current version.

AlessioRocco:
@kennyadsl I've updated the description to better explain because I did this change.

PR #3514, opened 2020-02-13
Improve unstocking inventory units from stock locations

AlessioRocco:
**Description**

This PR fixes the bug reported from this issue https://github.com/solidusio/solidus/issues/3527.

The `Spree::Stock::InventoryUnitsFinalizer#unstock_inventory_units` method was unstocking the `line_item` original `variant` but not the `variant` set in the `inventory_unit`, this PR fixes this error and improve the code a bit.

**Checklist:**

### DIFF
--- a/core/app/models/spree/stock/inventory_units_finalizer.rb
+++ b/core/app/models/spree/stock/inventory_units_finalizer.rb
@@ -34,11 +34,9 @@ module Spree
       end
 
       def unstock_inventory_units
-        inventory_units.group_by(&:shipment_id).each_value do |inventory_units_for_shipment|
-          inventory_units_for_shipment.group_by(&:line_item_id).each_value do |units|
-            shipment = units.first.shipment
-            line_item = units.first.line_item
-            shipment.stock_location.unstock line_item.variant, units.count, shipment
+        inventory_units.group_by(&:shipment).each do |shipment, units_by_shipment|
+          units_by_shipment.group_by(&:variant).each do |variant, units|
+            shipment.stock_location.unstock variant, units.count, shipment
           end
         end
       end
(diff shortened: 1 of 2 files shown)


kennyadsl:
@AlessioRocco do we have any performance improvements with this refactor as well?

AlessioRocco:
@kennyadsl I thought this PR shouldn't be related to the performance, but the fact that the `line item` is used to retrieve the `variant` when we have a direct association on the `inventory_unit` makes confusion to me. 

With this PR I'm trying to simplify the code and reduce this confusion.

It also allows creating a shipment with products that differ from the variant in the line item.

Suppose that you have a variant that represents a pack of products, with the old implementation you can't adapt the code to create a shipment that sends something different from the line item variant instead you can do it with the code proposed in this PR.

kennyadsl:
Before merging, maybe we should add a test to describe this behavior as well?

> Suppose that you have a variant that represents a pack of products, with the old implementation you can't adapt the code to create a shipment that sends something different from the line item variant instead you can do it with the code proposed in this PR.

sechix:
Hi, Could we merge this commit @AlessioRocco , change size is not working ok and I need it . Thanks!

kennyadsl:
@sechix I look forward to merging this but I'd like to wait for some specs for this behavior. This way we'll be sure to don't reintroduce the bug in future changes.

kennyadsl:
@AlessioRocco any news here?

AlessioRocco:
@kennyadsl I've added the specs and squashed the commits into one to make the change clear for the future.
I've also changed the PR title and description, can you please link the related issue https://github.com/solidusio/solidus/issues/3527 to this PR so it will be closed when merge?

Issue #2725, opened 2018-05-02
Paperclip closing, time to move to ActiveStorage?


askl56:
```Seems that Thoughtbot guys decided to retire Paperclip and are stopping all further development and support for it.

A handful of PRs that are marked "merge before deprecating": https://github.com/thoughtbot/paperclip/pulls 

Can't really find announcement or discussion around it though.
```

Is there a migration plan for moving from Solidus to Paperclip?

askl56:
Anyone on this?

gmacdougall:
We are planning on migration away from Paperclip. There was some initial work done in #2328 but it needs some additional work.

swcraig:
The work in #2337 most likely needs to be merged before we are able to easily extract Paperclip into an extension.

askl56:
@gmacdougall Happy to help in any way I can

exocode:
Hi there, any news on this topic. 

Came here by a research for Activestorage. I saw the team on Spree did an update: They already implemented ActiveStorage.

My questionis if there is a near future where this feature will be migrated?

Thank you very much

jacobherrington:
@exocode It looks like @swcraig is continuing work on this change right now!

askl56:
Any update on this?

equivalent:
@askl56  from https://spreecommerce.org/spree-3-5-and-3-6-with-rails-5-2-ruby-2-5-and-activestorage-support-released/

```
ActiveStorage
Active Storage facilitates uploading files to a cloud storage service like Amazon S3, Google Cloud Storage, or Microsoft Azure Storage and attaching those files to Active Record objects. It comes with a local disk-based service for development and testing and supports mirroring files to subordinate services for backups and migrations.

ActiveStorage was introduced in Rails 5.2 and with Spree 3.6 release it’s the default file storage option. We’ll still support Paperclip until Spree 4.0 when it will be removed from Spree core.
```

so I guess it's in

swcraig:
I am no longer actively working on this.
@elia has graciously picked up the work [here](https://github.com/solidusio/solidus/pull/2974).

kennyadsl:
This is now merged in master. It will be available when Rails 6.1 will be released.

PR #3515, opened 2020-02-14
Add Rails 6.1.0.alpha/master support

filippoliverani:
**Description**
This PR adds the ability to run Solidus with Rails master version.

It's useful to be able to start testing Solidus against Rails master to know what to expect and to reduce the time needed to upgrade after each Rails release.

There's also some Solidus [new features](https://github.com/solidusio/solidus/issues/2725) that require work and that will be widely used only from Rails 6.1.0 on.

**Checklist:**

### DIFF
--- a/core/lib/spree/awesome_nested_set_override.rb
+++ b/core/lib/spree/awesome_nested_set_override.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module AwesomeNestedSetOvveride
+  # Add :polimorphic key option only when used to make it work with Rails 6.1+
+  # https://github.com/osmaelo/rails/commit/2c008d9f6311d92b3660193285230505e74a114f
+  # This can be removed when upgrading to an awesome_nested_set version
+  # compliant with Rails 6.1+.
+  def acts_as_nested_set_relate_parent!
+    # Disable Rubocop to keep original code for diffs
+    # rubocop:disable
+    options = {
+      :class_name => self.base_class.to_s,
+      :foreign_key => parent_column_name,
+      :primary_key => primary_column_name,
+      :counter_cache => acts_as_nested_set_options[:counter_cache],
+      :inverse_of => (:children unless acts_as_nested_set_options[:polymorphic]),
+      :touch => acts_as_nested_set_options[:touch]
+    }
+    options[:polymorphic] = true if acts_as_nested_set_options[:polymorphic]
+    options[:optional] = true if ActiveRecord::VERSION::MAJOR >= 5
+    belongs_to :parent, options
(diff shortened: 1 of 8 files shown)


core/lib/spree/awesome_nested_set_override.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module AwesomeNestedSetOvveride
```
spaghetticode: I think it may be worth adding a comment here with the contents of the commit message. This should increase the chances that, when the time is right, somebody passing by will remove this file.

filippoliverani: Absolutely!

core/lib/spree/awesome_nested_set_override.rb
```diff
@@ -0,0 +1,26 @@
...
+      :inverse_of => (:children unless acts_as_nested_set_options[:polymorphic]),
+      :touch => acts_as_nested_set_options[:touch]
+    }
+    options[:polymorphic] = true if acts_as_nested_set_options[:polymorphic]
+    options[:optional] = true if ActiveRecord::VERSION::MAJOR >= 5
+    belongs_to :parent, options
+    # rubocop:disable
+  end
+
+  CollectiveIdea::Acts::NestedSet.prepend self
```
kennyadsl: I'm not sure about adding this kind of patches in Solidus core. I'm thinking that maybe I prefer to keep the build failing and showing exactly which gems need some changes. 

Maybe we should just propose the patch to https://github.com/collectiveidea/awesome_nested_set?

filippoliverani: Opening a PR on `awesome_nested_set` is a good idea, I'll do it as soon as possibile 👍 

Leaving the build broken could be a problem, Rails master is needed to work on some other issues and `awesome_nested_set`, as it is, breaks the code during initialization when Taxon class is loaded.  Is there a less intrusive solution we can try?

kennyadsl: We could create a separate/additional temporary Gemfile for Rails 6.1 only, that points to PRs on GitHub that address the Rails 6.1 issues. But it's not guaranteed that they already exist so maybe we can also create a file that is loaded only with Rails 6.1 and contains all the monkey patch that we have to remove before supporting that specific version.

filippoliverani: Now the monkey patch is loaded only with Rails 6.1+. When [this nested_awesome_set PR](https://github.com/collectiveidea/awesome_nested_set/pull/421) will be merged and released we can remove it altogether.

PR #3517, opened 2020-02-14
Spree::OptionValue#name delegates to Spree::OptionType even when nil

SamuelMartini:
**Description**
[Spree::OptionValue#name](https://github.com/solidusio/solidus/blob/v2.9.5/core/app/models/spree/option_value.rb#L17
) is delegated to `Spree::OptionType` with `option_type` prefix (option_type_name).

The json response of `Spree::Api::OptionValuesController` uses the `*option_value_attributes` helper variable which includes [:option_type_name](https://github.com/solidusio/solidus/blob/v2.9.5/api/app/helpers/spree/api/api_helpers.rb#L78).
If for some reason an option value is created without option_type the above controller will return a 500 error because of the name delegation.

For instance, `Spree::Api::OptionValuesController#create` allows to create an option value record with only `name` and `presentation` params. Once created, the json response view will call `*option_value_attributes` with `option_type_name` breaking the app.

This pr also: 
- remove `check_option_values` from the `option_values_controller_spec.rb`
because is not used anywhere in the file. 
- remove option value resource duplicated route.

**Checklist:**

### DIFF
--- a/api/config/routes.rb
+++ b/api/config/routes.rb
@@ -59,7 +59,6 @@ Spree::Core::Engine.routes.draw do
     end
     resources :option_values
 
-    resources :option_values, only: :index
     get '/orders/mine', to: 'orders#mine', as: 'my_orders'
     get "/orders/current", to: "orders#current", as: "current_order"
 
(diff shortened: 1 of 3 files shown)


kennyadsl:
@SamuelMartini I know this is quite old, but any chance you remember why you needed to create an option value without an option type?

PR #3519, opened 2020-02-21
Replace `Spree::Event#name_with_suffix` with `adapter#normalize_name`

spaghetticode:
`Spree::Event#name_with_suffix` is not generic enough to stay in the `Spree::Event` class,  as different adapters may not have the concept of a name suffix, while, on the other hand, may need some different naming manipulation.
    
So, event name manipulation is now delegated to the specific event adapter using the method `#normalize_name`. 
As we currently have only `ActiveSupportNotifications`, the existing code from `Spree::Event` has been moved in that adapter class.

**Checklist:**

### DIFF
--- a/core/lib/spree/event.rb
+++ b/core/lib/spree/event.rb
@@ -21,7 +21,7 @@ module Spree
     #     @order.finalize!
     #   end
     def fire(event_name, opts = {})
-      adapter.fire name_with_suffix(event_name.to_s), opts do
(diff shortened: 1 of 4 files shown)


kennyadsl:
@skukx if you have time, your feedback is appreciated here. 🙂

skukx:
This looks good to me.
  - [x] Moves logic for normalizing an event name to the adapter

PR #3520, opened 2020-02-21
Rescue state machine exception from api base controller

SamuelMartini:
**Description**
When an order is not completed cannot be transitioned via `cancel!` 

`OrdersController#cancel` call `@order.cancel!` through `canceled_by`:  the state machine
exception results in a 500 error when the action is called with a not completed order. 

Pr changes:
- rescue `StateMachines::InvalidTransition` exception from the `BaseController` returning the default json response and 422 status.
- Remove rescue `StateMachines::InvalidTransition` from Api/CheckoutsController. **Note** that this change remove the log message from the [update action](https://github.com/solidusio/solidus/pull/3520/commits/f6b7b80c17308d14873a7fc2b223c06a041415f3#diff-10992068b70df9a93b07ec08575d29a6L64) when the state machine raise an exception. However the [invaid_transition](https://github.com/solidusio/solidus/pull/3520/files#diff-72139884662b996fcc462020939b7807R193) method has an explicit message that should work:

Screenshot 2020-03-03 at 18 26 02
The first string in the above image is the removed log message. The last string is what is 
logged in `invalid_transition`.

- Remove a double authorization in the `next` action.

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -29,6 +29,7 @@ module Spree
       rescue_from ActiveRecord::RecordNotFound, with: :not_found
       rescue_from CanCan::AccessDenied, with: :unauthorized
       rescue_from Spree::Core::GatewayError, with: :gateway_error
+      rescue_from StateMachines::InvalidTransition, with: :invalid_transition
 
       helper Spree::Api::ApiHelpers
 
@@ -188,6 +189,12 @@ module Spree
 
       def default_per_page
         Kaminari.config.default_per_page
+      end
+
+      def invalid_transition(error)
+        logger.error("invalid_transition #{error.event} from #{error.from} for #{error.object.class.name}. Error: #{error.inspect}")
+
+        render "spree/api/errors/could_not_transition", locals: { resource: error.object }, status: :unprocessable_entity
       end
     end
   end
--- a/api/app/views/spree/api/orders/could_not_transition.json.jbuilder
+++ b/api/app/views/spree/api/orders/could_not_transition.json.jbuilder
@@ -1,4 +1,9 @@
 # frozen_string_literal: true
 
+Spree::Deprecation.warn(
+  'spree/api/orders/could_not_transition is deprecated.' \
+  ' Please use spree/api/errors/could_not_transition'
+)
+
 json.error(I18n.t(:could_not_transition, scope: "spree.api.order"))
 json.errors(@order.errors.to_hash)
--- a/api/app/controllers/spree/api/checkouts_controller.rb
+++ b/api/app/controllers/spree/api/checkouts_controller.rb
@@ -23,9 +23,6 @@ module Spree
         authorize! :update, @order, order_token
         @order.next!
         respond_with(@order, default_template: 'spree/api/orders/show', status: 200)
-      rescue StateMachines::InvalidTransition => error
-        logger.error("invalid_transition #{error.event} from #{error.from} for #{error.object.class.name}. Error: #{error.inspect}")
-        respond_with(@order, default_template: 'spree/api/orders/could_not_transition', status: 422)
       end
 
       def advance
@@ -42,9 +39,6 @@ module Spree
           @order.complete!
           respond_with(@order, default_template: 'spree/api/orders/show', status: 200)
         end
-      rescue StateMachines::InvalidTransition => error
-        logger.error("invalid_transition #{error.event} from #{error.from} for #{error.object.class.name}. Error: #{error.inspect}")
-        respond_with(@order, default_template: 'spree/api/orders/could_not_transition', status: 422)
       end
 
       def update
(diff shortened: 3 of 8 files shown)


api/app/controllers/spree/api/orders_controller.rb
```diff
@@ -27,7 +27,11 @@ class OrdersController < Spree::Api::BaseController
       def cancel
         authorize! :update, @order, params[:token]
         @order.canceled_by(current_api_user)
+
         respond_with(@order, default_template: :show)
+      rescue StateMachines::InvalidTransition => error
```
spaghetticode: With this addition, this controller includes these exact lines of codes 3 times... it's clearly becoming a pattern 😄  
I think it may be a good moment to DRY them up by extracting them in a private method... WDYT?

SamuelMartini: Thank you for the comment @spaghetticode 
I think you mean [Api::CheckoutsController](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/checkouts_controller.rb) where it is used two times. 

I agree that this could be a private method and that we can DRY. We could use a `rescue_from` in [Api::BaseController](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/base_controller.rb#L30) to deal with `StateMachines::InvalidTransition` and return 422 in all api controllers.  

I'll move the rescue from  the api/orders and api/checkouts controllers to the api base controller if the above works. Let me know!

spaghetticode: @SamuelMartini oh yes, it's Api::CheckoutsController 😅 
I think your proposal is good 🎉

api/app/controllers/spree/api/base_controller.rb
```diff
@@ -188,6 +189,11 @@ def paginate(resource)
       def default_per_page
         Kaminari.config.default_per_page
       end
+
+      def invalid_transition(error)
+        logger.error("invalid_transition #{error.event} from #{error.from} for #{error.object.class.name}. Error: #{error.inspect}")
+        respond_with(@order, default_template: 'spree/api/orders/could_not_transition', status: 422)
```
kennyadsl: I think `@order` is not present in any controller that could raise a state machine invalid transition. I like to be generic but can't work only with orders if we have other state machines in the system.

SamuelMartini: Thank you for catching that! @kennyadsl 

I'll change `@order` with `error.object` which should do the trick.

kennyadsl: Yes, but we should also have a generic template, now it's `spree/api/orders/could_not_transition`.

api/app/views/spree/api/orders/could_not_transition.json.jbuilder
```diff
@@ -1,4 +0,0 @@
-# frozen_string_literal: true
-
-json.error(I18n.t(:could_not_transition, scope: "spree.api.order"))
```
kennyadsl: Maybe someone is using this template in their store, we should keep it, print a deprecation warning and render the generic partial.

api/app/controllers/spree/api/base_controller.rb
SamuelMartini: @kennyadsl changed and split into three different commits.

I moved `spree/api/orders/could_not_transition` under errors `spree/api/errors/could_not_transition`. The changes in the controller should not diff in the api response.

api/app/views/spree/api/orders/could_not_transition.json.jbuilder
SamuelMartini: > Maybe we can also add some specs for other resources' controllers that have state machines? Just to be really sure it's working with any of them.

I could not find any other controller which raise an exception because of the failed state machine transition.

There are controllers that prevent any exception by controlling the flow:
[return_authorization#cancel](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/return_authorizations_controller.rb#L60-L67)

[inventory_unit prepare event](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/inventory_units_controller.rb#L37)

[shipment#ready](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/shipments_controller.rb#L60)

They also return a different message so it will be a breaking change to make them use the generic error message.


Added deprecation for order error messages:
https://github.com/solidusio/solidus/pull/3520/files#diff-72139884662b996fcc462020939b7807R195

api/app/controllers/spree/api/base_controller.rb
```diff
@@ -188,6 +189,17 @@ def paginate(resource)
       def default_per_page
         Kaminari.config.default_per_page
       end
+
+      def invalid_transition(error)
+        logger.error("invalid_transition #{error.event} from #{error.from} for #{error.object.class.name}. Error: #{error.inspect}")
+        if error.object.is_a? Spree::Order
```
kennyadsl, line 195: I don't get this, why we should print a deprecation message if an order is passed? I think the deprecation should only be present in the order `.jbuilder` error partial, that we will not use anymore in Solidus, and will be shown if someone is using that file in their own extension/store.

aldesantis:
@SamuelMartini if you have time, could you please finalize this by addressing @kennyadsl's comment? It would be great to get this into 2.11 so we can switch to the new behavior in 3.0!

api/app/views/spree/api/orders/could_not_transition.json.jbuilder
```diff
@@ -1,4 +1,8 @@
 # frozen_string_literal: true
 
+Spree::Deprecation.warn('spree/api/orders/could_not_transition is deprecated.'\
+  ' Please use spree/api/errors/could_not_transition'
+)
```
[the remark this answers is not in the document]
DanielePalombo: Fixed

PR #3524, opened 2020-02-25
Use Address name attribute in views and APIs

filippoliverani:
**Description**
This PR is a step to solve #3234 and a continuation of #3458 

The PR allows to start using the new `Spree::Address#name` attribute instead of `firstname` and `lastname` by adding a `show_address_deprecated_attributes` preference and providing alternate HTML and JSON views. 
The old attributes are not deprecated yet, but this PR discourages their usage and enables `name` usage everywhere.

**Checklist:**


api/app/helpers/spree/api/api_helpers.rb
```diff
@@ -121,11 +121,18 @@ def required_fields_for(model)
...
+        :id, :name, :firstname, :lastname, :full_name, :address1, :address2,
+        :city, :zipcode, :phone, :company, :alternative_phone, :country_id,
+        :country_iso, :state_id, :state_name, :state_text
+      ]
+
+      @@address_attributes = if Spree::Config[:show_address_deprecated_attributes]
+                               @@full_address_attributes
+                             else
+                               @@full_address_attributes -
+                                 Spree::Address::DEPRECATED_ATTRS.map(&:to_sym)
```
ericsaupe: If these are deprecated attributes should the logic here be additive instead? This makes the default list the non-deprecated fields with deprecated fields being added.

For example,

```
@@address_attributes = if Spree::Config[:show_address_deprecated_attributes]
                                               @@full_address_attributes + 
                                                 Spree::Address::DEPRECATED_ATTRS.map(&:to_sym)
                                             else
                                               @@full_address_attributes
                                              end
```

filippoliverani: Yes, it would me more readable, thank you! 👍

core/app/models/spree/address.rb
```diff
@@ -186,6 +189,17 @@ def name=(value)
 
     private
 
+    def validate_name
+      return if firstname.present?
+
+      name_attribute = if Spree::Config[:show_address_deprecated_attributes]
```
kennyadsl: This line makes me think that we can pick a better configuration name, showing deprecated attributes may not be directly related to model validation of addresses. I'm thinking to something like `Spree::Config[:use_combined_first_and_last_name_in_address]`, do you think it could work?

filippoliverani: Yes, makes sense 👍

frontend/spec/features/checkout_spec.rb
```diff
@@ -118,11 +118,11 @@
 
         it "shows the saved addresses" do
           within("#billing") do
-            expect(find_field('First Name').value).to eq 'Bill'
+            expect(find_field('Name').value).to eq 'Bill'
```
kennyadsl: I think we can change the code so that this will look like a full name now. And it can be fun to complete those with last names :)

frontend/spec/controllers/spree/checkout_controller_spec.rb
```diff
@@ -8,8 +8,8 @@
   let(:order) { FactoryBot.create(:order_with_totals) }
 
   let(:address_params) do
-    address = FactoryBot.build(:address)
-    address.attributes.except("created_at", "updated_at")
+    attributes_for(:address, name: 'John Doe')
```
kennyadsl: Shouldn't we change the factory?

backend/spec/controllers/spree/admin/users_controller_spec.rb
```diff
@@ -168,14 +167,14 @@ def user
...
-          put :update, params: { id: user.id, user: { first_name: "Bob", spree_role_ids: [dummy_role.id] } }
+          put :update, params: { id: user.id, user: { name: "Bob", spree_role_ids: [dummy_role.id] } }
         }.not_to change { user.reload.spree_roles.to_a }
       end
 
       it "cannot clear roles" do
         user.spree_roles << dummy_role
         expect {
-          put :update, params: { id: user.id, user: { first_name: "Bob" } }
+          put :update, params: { id: user.id, user: { name: "Bob" } }
```
kennyadsl: Even here, I'd change it to look like a real full name.

backend/app/views/spree/admin/orders/index.html.erb
```diff
@@ -65,14 +65,21 @@
...
-                <%= label_tag :q_bill_address_firstname_start, t('spree.first_name_begins_with') %>
-                <%= f.text_field :bill_address_firstname_start, size: 25 %>
-              </div>
-              <div class="field">
-                <%= label_tag :q_bill_address_lastname_start, t('spree.last_name_begins_with') %>
-                <%= f.text_field :bill_address_lastname_start, size: 25%>
-              </div>
+              <% if Spree::Config.use_combined_first_and_last_name_in_address %>
+                <div class="field">
+                  <%= label_tag :q_bill_address_name_start, t('spree.name_contains') %>
```
kennyadsl: Should this be `q_bill_address_name_cont`, right?

backend/spec/features/admin/orders/customer_details_spec.rb
```diff
@@ -14,8 +14,8 @@
   let!(:product) { create(:product_in_stock) }
 
   # We need a unique name that will appear for the customer dropdown
-  let!(:ship_address) { create(:address, country: country, state: state, first_name: "Rumpelstiltskin") }
-  let!(:bill_address) { create(:address, country: country, state: state, first_name: "Rumpelstiltskin") }
+  let!(:ship_address) { create(:address, country: country, state: state, name: "Rumpelstiltskin") }
+  let!(:bill_address) { create(:address, country: country, state: state, name: "Rumpelstiltskin") }
```
kennyadsl: Maybe here as well?

api/app/helpers/spree/api/api_helpers.rb
```diff
@@ -121,11 +121,18 @@ def required_fields_for(model)
...
+        :id, :name, :address1, :address2, :city, :zipcode, :phone, :company,
+        :alternative_phone, :country_id, :country_iso, :state_id, :state_name,
+        :state_text
+      ]
+
+      @@address_attributes = if Spree::Config.use_combined_first_and_last_name_in_address
+                               @@address_base_attributes
+                             else
+                               @@address_base_attributes +
+                                 Spree::Address::DEPRECATED_ATTRS.map(&:to_sym)
```
kennyadsl: Technically, these are not yet deprecated. Also, we may have other deprecations in the future that will be included/excluded here as well. What about calling this constant in a more semantic way like `SPLIT_NAME_ATTRS` instead?

backend/app/controllers/spree/admin/search_controller.rb
```diff
@@ -16,8 +16,7 @@ def users
           @users = Spree.user_class.ransack({
             m: 'or',
             email_start: params[:q],
-            addresses_firstname_start: params[:q],
-            addresses_lastname_start: params[:q]
+            name_start: params[:q]
```
kennyadsl: `name_cont`?

frontend/spec/features/checkout_spec.rb
filippoliverani: Yes, I'll updates name-related specs to make them more realistic 👍

frontend/spec/controllers/spree/checkout_controller_spec.rb
filippoliverani: You mean to use ''John Doe" as name or to remove timestamps?
In both cases we will have to update a lot of other specs relying on how address is created now, I don't think it's the right moment to do that

backend/spec/controllers/spree/admin/users_controller_spec.rb
filippoliverani: 👍

backend/app/views/spree/admin/orders/index.html.erb
filippoliverani: Not exactly, this part is bit tricky. 
`name` is a virtual attribute and we can't have a ransacker working on it and, to have the least impact on how the system works right now, the search on name is just as an alias of searching by `_firstname_start` or `_lastname_start`.
So it's neither a `_start` nor a `_cont`, but I opted to keep `_start` to keep it close to the ransackers it's aliasing. 
We could use `_firstname_or_startname_start` but it's quite verbose, I'm not sure which would be the best solution..

api/app/helpers/spree/api/api_helpers.rb
filippoliverani: Yes, right now that would be better. I was already thinking about the incoming deprecations but it's premature

backend/spec/features/admin/orders/customer_details_spec.rb
filippoliverani: 👍

backend/app/views/spree/admin/orders/index.html.erb
kennyadsl: Got it, my personal preference is `_firstname_or_lastname_start`

api/app/helpers/spree/api/api_helpers.rb
kennyadsl: That sounds perfect!

backend/app/controllers/spree/admin/search_controller.rb
kennyadsl: Resolved with https://github.com/solidusio/solidus/pull/3524/files#r385699525

backend/spec/controllers/spree/admin/users_controller_spec.rb
```diff
@@ -168,14 +167,14 @@ def user
...
-          put :update, params: { id: user.id, user: { first_name: "Bob", spree_role_ids: [dummy_role.id] } }
+          put :update, params: { id: user.id, user: { name: "Bob Bloggs", spree_role_ids: [dummy_role.id] } }
         }.not_to change { user.reload.spree_roles.to_a }
       end
 
       it "cannot clear roles" do
         user.spree_roles << dummy_role
         expect {
-          put :update, params: { id: user.id, user: { first_name: "Bob" } }
+          put :update, params: { id: user.id, user: { name: "Bob Bloggs" } }
```
kennyadsl: I see this as a missing opportunity to add `Bob Marley` to the repository 😆

filippoliverani: Ahah true 😂 
I tried to use boring names already present in specs to minimize surprise

kennyadsl: That makes perfect sense, I was just kidding!

filippoliverani:
> Thanks, @filippoliverani. This thing is becoming a reality!

Thank you for your help @kennyadsl! 💪

kennyadsl:
@filippoliverani can you check the conflict here. I can't recall merging something in that file lately.. 🤔

filippoliverani:
> @filippoliverani can you check the conflict here. I can't recall merging something in that file lately.. 🤔
Generators has been renamed recently, rebased 👍

core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -103,6 +103,7 @@ class Application < ::Rails::Application
 Spree.config do |config|
   config.mails_from = "store@example.com"
   config.raise_with_invalid_currency = false
+  config.use_combined_first_and_last_name_in_address = true
```
tvdeyen, line 106: This should have been `false`, because we still tested v2.11 here, that has this set to `false` (see above)
That way specs are misleadingly green and issues like #4094 didn't came up in the specs for v2.11

I am proposing to revert this in v2.11 branch

PR #3532, opened 2020-02-28
RfC: Remove some order transitions

mamhoff:
**Description**

This removes many, many transitions from the Order state machine, allowing us to reason more clearly about what the system is supposed to be doing and when.

I have used the wonderful https://github.com/state-machines/state_machines-graphviz gem to make the following visualisations.

Currently, our order state machine is automatically visualized like this: 

Spree::Order_state

With these changes, this image becomes a lot more streamlined and understandable:

Spree::Order_state

I have on purpose not added specs for this, as this is exclusively a feature removal for which I do not see why we should add tests. :)

**Checklist:**

### DIFF
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -94,11 +94,11 @@ RSpec.describe Spree::Order, type: :model do
       let!(:store_credit_payment_method) { create(:store_credit_payment_method) }
       let(:order_total) { 500.00 }
       let(:store_credit) { create(:store_credit, amount: order_total) }
-      let(:order) { create(:order_with_line_items, user: store_credit.user, line_items_price: order_total) }
+      let(:order) { create(:order_ready_to_complete, user: store_credit.user, line_items_price: order_total) }
 
       before do
         order.add_store_credit_payments
-        order.finalize!
+        order.complete!
         order.capture_payments!
       end
 
--- a/core/app/models/spree/order/checkout.rb
+++ b/core/app/models/spree/order/checkout.rb
@@ -60,7 +60,7 @@ module Spree
             end
 
             event :return do
-              transition to: :returned, from: [:returned, :complete, :awaiting_return, :canceled], if: :all_inventory_units_returned?
+              transition to: :returned, from: [:complete, :awaiting_return], if: :all_inventory_units_returned?
             end
 
             event :resume do
--- a/core/app/models/spree/order/checkout.rb
+++ b/core/app/models/spree/order/checkout.rb
@@ -68,7 +68,7 @@ module Spree
             end
 
             event :authorize_return do
-              transition to: :awaiting_return
+              transition to: :awaiting_return, from: :complete
             end
 
             event :complete do
--- a/core/app/models/spree/order/checkout.rb
+++ b/core/app/models/spree/order/checkout.rb
@@ -56,7 +56,7 @@ module Spree
             end
 
             event :cancel do
-              transition to: :canceled, if: :allow_cancel?
+              transition to: :canceled, if: :allow_cancel?, from: :complete
             end
 
             event :return do


tvdeyen:
I am very much in favor of this. But I am concerned about stores relying on this or expecting certain methods to exist. 

What about moving the current state machine into a "legacy" module and ship this as default state machine going forward.

mamhoff:
I would disagree. It's fairly easy to add transitions to the state machine, but impossible to remove them. This would be the perfect thing for a note in the changelog IMO.

mamhoff:
Added fix for a spec that expected an order in cart state to be finalized. 

There is some scope maybe for still adding a transition from `awaiting_return` to `canceled` I think, anyone using these features together?

kennyadsl:
Thanks, @mamhoff, I like this change a lot and that second picture makes it super tempting for our documentation.

I'm also a bit concerned about removing methods/transactions that could be used by someone and even with a good CHANGELOG entry it could be very hard to determine if you are relying on some of those. Deprecating what we plan to remove here would be a nightmare as well. 

I'll be sure to discuss this in the next core team meeting so we can provide a shared position about this change. Thanks again.

core/spec/models/spree/order_spec.rb
```diff
@@ -94,7 +94,7 @@
       let!(:store_credit_payment_method) { create(:store_credit_payment_method) }
       let(:order_total) { 500.00 }
       let(:store_credit) { create(:store_credit, amount: order_total) }
-      let(:order) { create(:order_with_line_items, user: store_credit.user, line_items_price: order_total) }
+      let(:order) { create(:order_with_line_items, state: :complete, user: store_credit.user, line_items_price: order_total) }
```
kennyadsl: What about using the [`order_ready_to_complete`](https://github.com/solidusio/solidus/blob/dab9035e75e76556bb0dc2cbcbf666fed9427d6c/core/lib/spree/testing_support/factories/order_factory.rb#L83-L100) factory and call `complete` in the before block?

mamhoff: Changed, and works as well!

mamhoff:
There is no way to deprecate this, and it wouldn't be meaningful. Well-instrumented stores would get a `StateMachines::InvalidTransition` exception that even details from which to which state the order could not go, and these stores can THEN decide whether they fix the code path that allows a user to cancel an unfinished order or whether they want to add 
```rb
state_machine do 
   transition from: :cart, to: :canceled
end
```
to their order decorator. Like I said, I'm open to adding transition where the business case is explainable, but for incomplete orders, I do not see why anyone would want cancellations or return authorizations. Also, I think those many transitions are mostly there because whoever introduced the `canceled` state did not think of specifying a `from:` option, not because it's an actually needed feature.

kennyadsl:
I totally agree it's just wrong as is, but users may have some untested code that relies on those wrong transitions and they will only face it once in production. 

I think it's about choosing if we consider a breaking change the removal of a public API, no matter if right or wrong, or only the removal of code that we think someone is actually using.

kennyadsl:
In the core team meeting, we decided that this needs a deprecation since is breaking some public methods. 

The approach proposed by @tvdeyen seems good, we can have a preference for new stores that will use the fixed state machine and deprecate the old class. What do you think?

mamhoff:
We already have configurable state machine modules for the other state machines in the system (shipment, inventory unit, return authorization etc.). I'll spike making the order one configurable as well, and define both versions in the source code.

peterberkenbosch:
love this!

spaghetticode:
FYI, just for clarity, the first graph is not accurate _at least_ in regards to the `cancel` transition. Only `complete` orders can currently become `canceled`, as there is a conditional guard in the SM. Please take a look [here](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order/checkout.rb#L58):
```ruby
event :cancel do
  transition to: :canceled, if: :allow_cancel?
end
```
and [here](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L285):
```ruby
def allow_cancel?
  return false unless completed? && state != 'canceled'
  shipment_state.nil? || %w{ready backorder pending}.include?(shipment_state)
end
```

mamhoff:
@spaghetticode good find; this means that we can, in fact, just remove those transitions without much pain. 

Regarding the other big source of weird transitions: The `authorize_return` event is NOWHERE TO BE FOUND in the entire code base...

kennyadsl:
@mamhoff Hey there! I'd love to add this, with the deprecated module, in the next Solidus release.  Did you have some time to continue this work?

spaghetticode:
@mamhoff thank you for this RCF, I moved these changes into #3542 so we can take advantage of the new module introduced there for reducing the number of transitions without breaking the current state machine for Spree::Order.

Issue #3526, opened 2020-02-26
Stop using ActiveJob::TestHelper methods in specs


filippoliverani:
Solidus testing relies almost entirely on RSpec, with a small exception: some specs directly invoke `ActiveJob::TestHelper` methods, to force the execution of mailers and Promotion related jobs.

I think it should be better to stop using them, avoiding Rails TestHelpers stack would make specs easier to maintain and understand.

RSpec's Job Spec assertions, with a combination of `:inline` and `:test` `ActiveJob::QueueAdapter`s, should cover any testing need.

PR #3543, opened 2020-03-08
Replace ActiveJob::TestHelper with own module.

jessetilro:
Issue #3526.

**Description**
This is my first contribution. I decided to make an attempt to resolve this issue as it was flagged as "Good First Issue". Looking forward to your feedback!

The test code does no longer rely on `ActiveJob::TestHelper`.
The only method used from this module was `perform_enqueued_jobs`.
I implemented a method providing this functionality in an own testing helper module,
such that existing test cases do not have to be changed and remain readable.

**Checklist:**

### DIFF
--- a/api/spec/spec_helper.rb
+++ b/api/spec/spec_helper.rb
@@ -63,8 +63,6 @@ RSpec.configure do |config|
     Rails.cache.clear
   end
 
-  config.include ActiveSupport::Testing::Assertions
-
   config.use_transactional_fixtures = true
 
   config.example_status_persistence_file_path = "./spec/examples.txt"
--- a/api/spec/spec_helper.rb
+++ b/api/spec/spec_helper.rb
@@ -30,6 +30,7 @@ require 'spree/testing_support/partial_double_verification'
 require 'spree/testing_support/factories'
 require 'spree/testing_support/preferences'
 require 'spree/testing_support/authorization_helpers'
+require 'spree/testing_support/job_helpers'
 
 require 'spree/api/testing_support/caching'
 require 'spree/api/testing_support/helpers'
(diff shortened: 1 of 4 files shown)


aldesantis:
@kennyadsl assigning you as a reviewer here because you made the original issue and I'm not 100% sure this is what you had in mind? Let us know!

kennyadsl:
@aldesantis the original issue was made by @filippoliverani. We discussed IRL about this and I think it's quite different from what we had in mind but I'm sure he is able to explain it better than me.

filippoliverani:
@kennyadsl The proposed solution in the issue was to stop calling `#perform_enqueued_jobs` by using `:inline` queue adapter by default. That would require to update some specs rewriting their expectations. 

I think that this PR is good first step in that direction, it has a moderate impact on code but effectively removes `ActiveJob::TestHelper` dependency 👍

api/spec/spec_helper.rb
```diff
@@ -62,7 +64,6 @@
   end
 
   config.include ActiveSupport::Testing::Assertions
```
filippoliverani, line 66: I think that without `ActiveJob::TestHelper` we can get rid of `ActiveSupport::Testing::Assertions` module too

backend/spec/spec_helper.rb
```diff
@@ -107,12 +108,12 @@
 
   config.include FactoryBot::Syntax::Methods
   config.include ActiveSupport::Testing::Assertions
```
filippoliverani, line 110: I think that without `ActiveJob::TestHelper` we can get rid of `ActiveSupport::Testing::Assertions` too

core/spec/rails_helper.rb
```diff
@@ -41,7 +42,8 @@
     Rails.cache.clear
   end
 
+  config.include Spree::TestingSupport::JobHelpers
+
   config.include ActiveSupport::Testing::Assertions
```
filippoliverani, line 47: I think that without `ActiveJob::TestHelper` we can get rid of `ActiveSupport::Testing::Assertions` too

kennyadsl:
@filippoliverani thanks for this review! I'm fine with this, @jessetilro can you please just fix the comments left in the review? Thanks again! 🙏

jessetilro:
@kennyadsl I fixed the comments in the review.

PR #3544, opened 2020-03-09
do not use link_to_add_fields for buttons, its text will be removed from minimize menu

hefan:
**Description**

**Problem**:
The link_to_add_fields function will (when used for buttons) create the following markup
`<a ...>
  <span class="text">Add Option Value</span>
</a>`
The span with class text will be removed when the menu is minimized.
So we only see the plus icon with no text when the menu is minimized.

This happens in the option type edit screen and in the product properties index screen.
_/admin/option_types/1/edit_
_/admin/products/ruby-hoodie/product_properties_

See Screenshot:
Screenshot from 2020-03-09 16-59-14

**Solution**:
This PR removes the symbols from the buttons just by using **link_to** instead of **link_to_add_fields**.
The Buttons will have no more symbol but the text stays when the menu is minimized.
These two where the only buttons with symbols in the Backend.

### DIFF
--- a/backend/app/controllers/spree/admin/customer_returns_controller.rb
+++ b/backend/app/controllers/spree/admin/customer_returns_controller.rb
@@ -70,8 +70,6 @@ module Spree
           return_item = item_params[:id] ? Spree::ReturnItem.find(item_params[:id]) : Spree::ReturnItem.new
           return_item.assign_attributes(item_params)
 
-          return_item.skip_customer_return_processing = true
-
           if item_params[:reception_status_event].blank?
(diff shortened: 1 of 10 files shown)


kennyadsl:
There's another occurrence of this into https://github.com/solidusio/solidus/blob/d3be0c490c3627855dc554c7393c7bf5dc87a774/backend/app/views/spree/admin/product_properties/index.html.erb#L58:L58, can you please check if we can remove that as well, and maybe deprecate the helper if we are not using it anymore in the codebase?

hefan:
ok done in #3547

PR #3547, opened 2020-03-10
replace link_to_add_fields usage and deprecate helper function

hefan:
**Description**

**Problem**:
The link_to_add_fields function will (when used for buttons) create the following markup
`<a ...>
  <span class="text">Add Option Value</span>
</a>`
The span with class text will be removed when the menu is minimized.
So we only see the plus icon with no text when the menu is minimized.

This happens in the option type edit screen and in the product properties index screen.
_/admin/option_types/1/edit_
_/admin/products/ruby-hoodie/product_properties_ (2 occurrences)

See Screenshot:
Screenshot from 2020-03-09 16-59-14

**Solution**:
This PR removes the symbols from the buttons just by using **link_to** instead of **link_to_add_fields**.
The Buttons will have no more symbol but the text stays when the menu is minimized.
These three buttons where the only buttons with symbols in the Backend.

The helper function is also deprecated

See also #3544

### DIFF
--- a/backend/app/helpers/spree/admin/base_helper.rb
+++ b/backend/app/helpers/spree/admin/base_helper.rb
@@ -120,6 +120,8 @@ module Spree
         css_classes = options[:class] ? options[:class] + " spree_add_fields" : "spree_add_fields"
         link_to_with_icon('plus', name, 'javascript:', data: { target: target }, class: css_classes)
       end
+      deprecate link_to_add_fields: "Please use button_tag instead, Example: \n" \
+        "button_tag \"Name\", type: \"button\", data: { target: \"tbody#div\" }", deprecator: Spree::Deprecation
 
       # renders hidden field and link to remove record using nested_attributes
       def link_to_remove_fields(name, form, options = {})
--- a/backend/app/views/spree/admin/option_types/edit.html.erb
+++ b/backend/app/views/spree/admin/option_types/edit.html.erb
@@ -6,7 +6,9 @@
 <% content_for :page_actions do %>
   <li>
     <span id="new_add_option_value" data-hook>
-      <%= link_to_add_fields t('spree.add_option_value'), "tbody#option_values", class: 'btn btn-primary' %>
+      <%= button_tag t('spree.add_option_value'), type: 'button',
+            data: { target: 'tbody#option_values' },
+            class: 'btn btn-primary spree_add_fields' %>
     </span>
   </li>
 <% end %>
--- a/backend/app/views/spree/admin/product_properties/index.html.erb
+++ b/backend/app/views/spree/admin/product_properties/index.html.erb
@@ -7,7 +7,9 @@
   <% if can?(:create, Spree::ProductProperty) %>
     <ul class="tollbar inline-menu">
       <li>
-        <%= link_to_add_fields t('spree.add_product_properties'), 'tbody#product_properties', class: 'plus btn btn-primary' %>
+        <%= button_tag t('spree.add_product_properties'), type: 'button',
+              data: { target: 'tbody#product_properties' },
+              class: 'btn btn-primary spree_add_fields' %>
       </li>
     </ul>
   <% end %>
@@ -55,7 +57,9 @@
       <div class="form-buttons filter-actions actions">
         <%= button_tag t('spree.filter_results'), class: 'btn btn-primary' %>
         <% if @option_value_ids.present? %>
-          <%= link_to_add_fields t('spree.add_variant_properties'), 'tbody#variant_property_values', class: 'plus button' %>
+          <%= button_tag t('spree.add_variant_properties'), type: 'button',
+                data: { target: 'tbody#variant_property_values' },
+                class: 'button spree_add_fields' %>
         <% end %>
       </div>
     </fieldset>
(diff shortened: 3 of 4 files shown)


backend/app/views/spree/admin/option_types/edit.html.erb
```diff
@@ -6,7 +6,9 @@
 <% content_for :page_actions do %>
   <li>
     <span id="new_add_option_value" data-hook>
-      <%= link_to_add_fields t('spree.add_option_value'), "tbody#option_values", class: 'btn btn-primary' %>
+      <%= link_to t('spree.add_option_value'), "javascript:",
```
kennyadsl: What about using a `button_tag`? This way we don't have to set that fake path with `javascript:`.

backend/app/helpers/spree/admin/base_helper.rb
```diff
@@ -115,6 +115,7 @@ def preference_fields(object, form)
              "form: f, attribute: \"preferred_\#{key}\", label: t(key, scope: 'spree') %>\n" \
         "<% end %>", deprecator: Spree::Deprecation
 
+      # @deprecated use link_to instead
```
kennyadsl: This won't really put a deprecation warning for users that are using this method in their code. 

We should start using the suggested versiona and add  a real deprecation message, something like:

```
def link_to_add_fields(name, target, options = {})
  name = '' if options[:no_text]
  css_classes = options[:class] ? options[:class] + " spree_add_fields" : "spree_add_fields"
  button_tag(name, data: { target: target }, class: css_classes)
end
deprecate link_to_add_fields: "Please use button_tag instead, Example: \n" \
  "button_tag \"Name\", data: { target: \"tbody#div\" }, deprecator: Spree::Deprecation
```

as we are doing for other methods in this file.

backend/app/views/spree/admin/option_types/edit.html.erb
hefan: ok. done.

backend/app/helpers/spree/admin/base_helper.rb
hefan: ok. done. i had to add type: 'button' as a parameter to get it to work in all 3 contexts.

hefan:
i guess the specs also need a rewrite, click_button instead of click link. will do that later.

hefan:
ok. all works fine now

backend/app/views/spree/admin/product_properties/index.html.erb
```diff
@@ -7,7 +7,9 @@
   <% if can?(:create, Spree::ProductProperty) %>
     <ul class="tollbar inline-menu">
       <li>
-        <%= link_to_add_fields t('spree.add_product_properties'), 'tbody#product_properties', class: 'plus btn btn-primary' %>
+        <%= button_tag t('spree.add_product_properties'), type: 'button',
+                    data: { target: 'tbody#product_properties' },
```
kennyadsl: Why this indent? I think two spaces from the start of the `button_tag` is the recommended way.

hefan: ok, we're getting there. I had the idea to put the arguments in one column (before the change to button_tag)

spaghetticode:
@hefan thanks for working on this! I was thinking, what about adding a new helper method? It would allow to DRY-up a bit your solution, as very similar ERB code is repeated 3 times.

kennyadsl:
@spaghetticode I'm 👎for creating a new helper. We can change a use the one we already have (removing the deprecation) but I personally think we should move to less custom helpers and more what's already provided by Rails, at least in backend views.

spaghetticode:
> I personally think we should move to less custom helpers and more what's already provided by Rails, at least in backend views

@kennyadsl I would like to better understand your opinion, can you explain to me more in detail the reason for this sentence?

hefan:
Aren't there many helper functions in base_helper and navigation_helper for creating Buttons or Links which couldn't be done also with the normal Rails helper Functions.  i found them confusing and they seem to do nearly the same thing. Also they do things no longer needed (the icons on buttons for example seem to be removed).

kennyadsl:
I don't think a wrapper that only adds a couple of attributes to an existing Rails helper is worth keeping, here's why:

- **abstraction**: from the perspective of who is understanding how that part works, there are no notable differences between using a helper method or adding a class to an element. It's just one more step.
- **deprecations**: yes, having a helper could be helpful to print deprecation warnings but no one ensures us that stores are not using the JS part of this functionality only, bypassing the provided helpers already. This will end up having us deprecate two things instead of one. 

Personally, and for consistency, I'd deprecate `link_to_remove_fields` as well. Maybe in another PR?

spaghetticode:
@kennyadsl when a store needs to change the behavior, a helper would be helpful in not making necessary to change any views (the occurences are 3  in 2 view files this time).  Decorating the helper would do... that's the main benefit I see about using a helper here. Then there is reusability, when one wants to add a similar button somewhere else. 

Three instances is the bare minimum number of occurrences to consider DRYing up the code, so I don't see any problem in postponing this later if it will ever make more sense. Also, taking in account how simple is the code (it's a bit verbose on 3 lines, but no `if` clause is present for example), repeating makes probably more sense.

hefan:
So i would do a PR removing and deprecating link_to_remove_fields, ok? there is only one occurence.
About the 3 Lines: Basically the new method produces not much more code than using the old helper method. i found it more readable in 3 lines. i am not sure about the best practice in erb with the long param list. Should i keep it in seperate lines or should i prefer longer lines?

kennyadsl:
@hefan if you have time, a PR to remove `link_to_remove_fields` as well would be appreciated, thanks again!

PR #3550, opened 2020-03-12
[backend] Disable customer returns buttons after first click

AlessioRocco:
**Description**
This PR continue the work started with https://github.com/solidusio/solidus/pull/3342:

> It may happen that admin users click multiple times (purposely or not) on submit buttons in the backend area.
> 
> This can be a problem as at times it may silently generate unwanted duplicate records, or show an error page due to duplicate records. This is much more likely to happen on real shops when unanticipated server load makes the application unresponsive (at first click nothing seems to happen, so then why not click again?) and where some controller create actions may be patched and be slower.
> 
> By changing button_tag to f.submit or submit_tag the button gets disabled after the first click, thanks to the automatically added attribute data-disable-with="Create"

by disabling the missing submit buttons in the customer returns backend page.

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/customer_returns/_return_item_decision.html.erb
+++ b/backend/app/views/spree/admin/customer_returns/_return_item_decision.html.erb
@@ -50,7 +50,7 @@
           <td>
             <%= form_for [:admin, return_item] do |f| %>
               <%= f.hidden_field 'reception_status_event', value: 'receive' %>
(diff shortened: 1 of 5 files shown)


aldesantis:
@AlessioRocco do you have a second to finish this? 🙏

elia (approved):

elia (approved):

rainerdema (approved):
👏

PR #3551, opened 2020-03-12
Enable api checkout spec that was skipped

SamuelMartini:
__NOTE__: https://github.com/solidusio/solidus/pull/3520 introduces a deprecation warning for invalid state transition. If the current pr will be merged before #3520 then #3520 needs to be rebased against master to add the deprecation expectation to the *no more* skipped test described here. Not doing so will break the CI.

**Description**
The skipped spec makes sense when the `ship_address` is set to nil. This is because the state machine is checking on the ship address and not the bill address.
https://github.com/solidusio/solidus/blob/v2.10.0/core/app/models/spree/order/checkout.rb#L100

**Checklist:**

### DIFF
--- a/api/spec/requests/spree/api/checkouts_controller_spec.rb
+++ b/api/spec/requests/spree/api/checkouts_controller_spec.rb
@@ -59,12 +59,8 @@ module Spree
       end
 
       it "will return an error if the order cannot transition" do
-        skip "not sure if this test is valid"
-        order.bill_address = nil
-        order.save
-        order.update_column(:state, "address")
+        order.update!(state: 'address', ship_address: nil)
         put spree.api_checkout_path(order.to_param), params: { order_token: order.guest_token }
-        # Order has not transitioned
         expect(response.status).to eq(422)
       end
 


api/spec/requests/spree/api/checkouts_controller_spec.rb
```diff
@@ -59,8 +59,7 @@ module Spree
       end
 
       it "will return an error if the order cannot transition" do
-        skip "not sure if this test is valid"
-        order.bill_address = nil
+        order.ship_address = nil
```
spaghetticode: What about grouping the order updates in one single instruction so we hit the DB only once? Something like `order.update!(ship_address: nil, state: "address)`

SamuelMartini: Done @spaghetticode . Also, added a _note_ to the pr!

api/spec/requests/spree/api/checkouts_controller_spec.rb
```diff
@@ -59,12 +59,8 @@ module Spree
       end
 
       it "will return an error if the order cannot transition" do
-        skip "not sure if this test is valid"
-        order.bill_address = nil
-        order.save
-        order.update_column(:state, "address")
+        order.update(state: 'address', ship_address: nil)
```
aldesantis: Can you use `update!` instead of `update`? That way, if the spec setup fails for some reason, the test will break loudly instead of failing its expectation.

kennyadsl:
Hey @SamuelMartini, do you have the chance to take a look at the requested changes here? Thanks!

SamuelMartini:
@kennyadsl @aldesantis Forgot about this!
Done! https://github.com/solidusio/solidus/pull/3551/files#diff-2ba290e3442635458400d2c2b868fb0eR62

aldesantis:
Thanks @SamuelMartini!

PR #3553, opened 2020-03-13
Add order_recalculated event

spaghetticode:
This PR adds the new event `order_recalculated`, also deprecating the use of order update hooks and the public method `OrderUpdater#run_hooks`. A conditional check has been added before running `#run_hooks` in order to avoid tons of not-needed deprecation messages.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -440,7 +440,7 @@ module Spree
 
       updater.update_shipment_state
       save!
-      updater.run_hooks
+      updater.run_hooks if update_hooks.any?
 
       touch :completed_at
 
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -183,6 +183,9 @@ module Spree
     # Use this method in other gems that wish to register their own custom logic
     # that should be called after Order#update
     def self.register_update_hook(hook)
+      Spree::Deprecation.warn \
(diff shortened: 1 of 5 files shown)


core/app/models/spree/order_updater.rb
```diff
@@ -26,12 +26,17 @@ def update
           update_shipments
           update_shipment_state
         end
-        run_hooks
+        update_hooks.any? && run_hooks
```
aldesantis, line 29: Could you turn this into a more semantic inline `if` instead?

core/app/models/spree/order.rb
```diff
@@ -437,7 +440,7 @@ def finalize!
 
       updater.update_shipment_state
       save!
-      updater.run_hooks
+      update_hooks.any? && updater.run_hooks
```
aldesantis, line 443: This should also be an inline `if` imo, it's much more intentional. 🙂

aldesantis:
@spaghetticode one thing that just came to mind: how about adding a test for the new event?

spaghetticode:
@aldesantis sure 👍

spaghetticode:
@aldesantis test added. When you confirm it looks good to you I'll merge this PR.

aldesantis:
@spaghetticode looks good to me, feel free to merge!

PR #3555, opened 2020-03-13
DRY paranoia and discard inclusion into models

cedum:
This PR DRYs the soft delete inclusion logic into a reusable concern.
This also deprecates `.with_deleted` and `.only_deleted` Paranoia' scopes, which will be removed in Solidus v3.

This has been extracted from the #3488 PR. After some feedback, I'm extracting the non-breaking bits into separate PRs, so we can merge them without having to wait for Solidus v3.
Once merged, I'll rebase and revisit #3488 to definitely remove Paranoia as a dependency from Solidus v3.

**Checklist:**

### DIFF
--- a/core/spec/models/spree/variant_spec.rb
+++ b/core/spec/models/spree/variant_spec.rb
@@ -159,7 +159,11 @@ RSpec.describe Spree::Variant, type: :model do
       context "and a variant is really deleted" do
         let!(:old_option_values_variant_ids) { variant.option_values_variants.pluck(:id) }
 
-        before { variant.really_destroy! }
+        before do
+          # #really_destroy! will be replaced here with #destroy when Paranoia
+          # will be removed in Solidus 3.0
+          Spree::Deprecation.silence { variant.really_destroy! }
+        end
 
         it "leaves no stale records behind" do
           expect(old_option_values_variant_ids).to be_present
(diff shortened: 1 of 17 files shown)


kennyadsl:
Cool, I think we still have some occurrences of the newly deprecated methods, can you take a look at failing specs? 🙏

cedum:
Ah, I missed that paranoia itself is also using ([see here](https://github.com/rubysherpas/paranoia/blob/0600183a12b1e0c5d74d94000bee08d5fc4a5933/lib/paranoia.rb#L154)) the newly deprecated scopes (`with_deleted` and `only_deleted`) 🤦‍♂
It seems most of the failing cases are related to using `really_destroy!` in some specs. I'll try to find an alternative solution for these specs without removing the deprecations introduced here.

core/spec/models/spree/variant_spec.rb
```diff
@@ -159,7 +159,9 @@
       context "and a variant is really deleted" do
         let!(:old_option_values_variant_ids) { variant.option_values_variants.pluck(:id) }
 
-        before { variant.really_destroy! }
+        before do
+          Spree::Deprecation.silence { variant.really_destroy! }
```
DanielePalombo: Would you mind to add a comment here to specify this will be replaced with `destroy` in solidus 3.0?

aldesantis:
@cedum this should now be good to merge. 🙂

PR #3557, opened 2020-03-13
Add CircleCI job to run tests against Rails master

filippoliverani:
**Description**
This PR adds a CircleCI job to run tests against Rails master. Right now it's needed to bring forward [the work to support ActiveStorage](https://github.com/solidusio/solidus/pull/3501) and could be useful to spot as soon as possible potentially breaking changes brought in by Rails new features.

The job could be disabled or set as a non-blocking status check in Github's branch protection rules to avoid blocking unrelated PRs.

**Checklist:**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,6 +96,15 @@ jobs:
       - setup
       - test
 
+  postgres_rails_master:
+    executor: postgres
+    parallelism: *parallelism
+    environment:
+      RAILS_VERSION: 'master'
+    steps:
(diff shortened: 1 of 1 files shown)


filippoliverani:
> This is not part of the list of workflow's jobs so it won't run yet. Is that intentional?

@kennyadsl nope, missed that, added to workflow, thank you

aldesantis:
@filippoliverani thanks! One of many steps towards making Solidus even more future-proof. ❤️

Issue #3537, opened 2020-03-02
[admin] Prefill user's address when creating a new order for specific user


spaghetticode:
Go to `/admin/orders` and click `New Order`. Click on the `Customer` tab, now you can choose from the autocomplete the user's email. This will populate the form address fields accordingly.

Now go to `/admin/users` and pick a user. You can now click the link `Create order for this user`. You will be redirected to a similar form as the one for the feature above, but this time in the `Customer` tab the email address in the form is prefilled, but the address is not. 

I think in this case it would make sense to have the form already populated with the user's address, when available. 

<img width="1013" alt="Schermata 2020-03-02 alle 15 16 59" src="https://user-images.githubusercontent.com/141220/75684371-f301fe80-5c98-11ea-8e3a-d0e7d5efa7e8.png">

jaimelr:
Hi! I'm going to work on this issue, I hope to have PR up with progress soon. 🙂

jaimelr:
@spaghetticode Hi! I got a PR up for this, reviews are appreciated! #3558

seand7565:
Looks like #3558 solves this one! 🎉 If so, can we close this issue?

jarednorman:
Affirmative! This was resolved by #3558.

PR #3558, opened 2020-03-18
Add support for prefill user address in new order

jaimelr:
**Description**
Ref #3537 
When a new order is created passing a user id we want the address to be prefilled.

This associate the user address (both ship and bill) with the order using `update`

test_address_issue_solidus

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/orders_controller.rb
+++ b/backend/app/controllers/spree/admin/orders_controller.rb
@@ -61,7 +61,11 @@ module Spree
 
       def new
         user = Spree.user_class.find_by(id: params[:user_id]) if params[:user_id]
-        @order = Spree::Core::Importer::Order.import(user, order_params)
+        order_importer_params = order_params
+        order_importer_params[:bill_address] = user&.bill_address
+        order_importer_params[:ship_address] = user&.ship_address
+
+        @order = Spree::Core::Importer::Order.import(user, order_importer_params)
         redirect_to cart_admin_order_url(@order)
       end
 
--- a/backend/spec/controllers/spree/admin/orders_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/orders_controller_spec.rb
@@ -102,7 +102,7 @@ describe Spree::Admin::OrdersController, type: :controller do
       end
 
       context "when a user_id is passed as a parameter" do
-        let(:user)  { mock_model(Spree.user_class) }
+        let(:user)  { mock_model(Spree.user_class, ship_address: mock_model(Spree::Address), bill_address: nil) }
         before { allow(Spree.user_class).to receive_messages find_by: user }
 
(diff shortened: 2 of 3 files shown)


backend/app/controllers/spree/admin/orders_controller.rb
```diff
@@ -62,6 +62,7 @@ def index
       def new
         user = Spree.user_class.find_by(id: params[:user_id]) if params[:user_id]
         @order = Spree::Core::Importer::Order.import(user, order_params)
+        associate_user_address(user)
```
kennyadsl: Did you consider moving this operation into `Spree::Core::Importer::Order` instead? 

I think it would be "just" matter of preparing the `order_params` to also have the `ship_address_attributes` and `bill_address_attributes` fields before calling the `import` method. What do you think?

jaimelr: I agree. We thought about this approach too, it's good to have some backup. I updated the PR, let me know what you think. And sorry for the long wait! 🙇

backend/spec/controllers/spree/admin/orders_controller_spec.rb
```diff
@@ -19,7 +19,8 @@
         total:           100,
         number:          'R123456789',
         all_adjustments: adjustments,
-        ship_address: mock_model(Spree::Address)
+        ship_address: mock_model(Spree::Address),
+        update: true
```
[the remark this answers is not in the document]
jaimelr: It turns out it wasn't needed after moving the logic to the Importer. 😅

backend/app/controllers/spree/admin/orders_controller.rb
```diff
@@ -61,7 +61,7 @@ def index
 
       def new
         user = Spree.user_class.find_by(id: params[:user_id]) if params[:user_id]
-        @order = Spree::Core::Importer::Order.import(user, order_params)
+        @order = Spree::Core::Importer::Order.import(user, order_params(user))
```
kennyadsl, line 64: I'd go with something like this:

```ruby
def new
  user = Spree.user_class.find_by(id: params[:user_id]) if params[:user_id]
  order_importer_params = order_params
  order_importer_params.merge({
    ship_address: user.ship_address,
    bill_address: user.bill_address
  }) if user

  @order = Spree::Core::Importer::Order.import(user, order_importer_params)
  redirect_to cart_admin_order_url(@order)
end
```

Anyway, probably with another PR, it could make sense to deprecate passing the `user` argument as a separate one, while we could just add it the `order_importer_params` hash. I'm not sure why it's this way and may be worth exploring.

jaimelr: Interesting approach. I think I agree with it since passing `user` to `order_params` seems odd.

PR #3561, opened 2020-03-19
adjust option value handling to work like product properties handling

hefan:
also remove last link_to_remove_fields usage and deprecate link_to_remove_fields helper

as discussed in #3547

The handling of the option value list worked in another way like the product properties list (both are the same type of lists with dynamically added rows):

* There was no new item by default
* all new options also had a remove button, which is not necessary

The Option Value list now works like the product property list.

Option value remove tests are also adjusted and work more like the product property remove tests.

The Test "can remove a non-persisted option value from an option type" is no longer needed, there is no more possibilty to remove the non persisted option value.

### DIFF
--- a/backend/app/views/spree/admin/option_types/_option_value_fields.html.erb
+++ b/backend/app/views/spree/admin/option_types/_option_value_fields.html.erb
@@ -1,11 +1,11 @@
 <tr class="option_value fields" id="spree_<%= dom_id(f.object) %>"  data-hook="option_value">
   <td>
-    <% if f.object.persisted? %>
+    <% if f.object.persisted? && can?(:update_positions, f.object) %>
       <span class="handle"></span>
       <%= f.hidden_field :id %>
     <% end %>
   </td>
   <td class="name"><%= f.text_field :name %></td>
   <td class="presentation"><%= f.text_field :presentation %></td>
-  <td class="actions"><%= link_to_remove_fields t('spree.actions.remove'), f, no_text: true %></td>
+  <td class="actions"><%= link_to_delete(f.object, no_text: true) if f.object.persisted? && can?(:destroy, f.object) %></td>
 </tr>
--- a/backend/app/controllers/spree/admin/option_types_controller.rb
+++ b/backend/app/controllers/spree/admin/option_types_controller.rb
@@ -3,7 +3,7 @@
 module Spree
   module Admin
     class OptionTypesController < ResourceController
-      before_action :setup_new_option_value, only: :edit
+      before_action :setup_new_option_value, only: :edit, if: -> { can?(:create, model_class) }
 
       def update_values_positions
         params[:positions].each do |id, index|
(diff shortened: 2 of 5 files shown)


backend/app/views/spree/admin/option_types/_option_value_fields.html.erb
```diff
@@ -1,11 +1,19 @@
...
-    <% if f.object.persisted? %>
+    <% if f.object.persisted? && can?(:update_positions, f.object) %>
       <span class="handle"></span>
       <%= f.hidden_field :id %>
     <% end %>
   </td>
-  <td class="name"><%= f.text_field :name %></td>
-  <td class="presentation"><%= f.text_field :presentation %></td>
-  <td class="actions"><%= link_to_remove_fields t('spree.actions.remove'), f, no_text: true %></td>
+  <td class="name">
```
spaghetticode: Can you keep the previous formatting for these lines? Personally I prefer your formatting, but it just comes down to tastes... I think that we should minimize unnecessary changes, so we keep the git history clean.

hefan: ok. i thought i do it like in the product property fields for the sake of consistency
https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/product_properties/_product_property_fields.html.erb
But if the git history is more important i'll revert it.

hefan: ok. done

kennyadsl:
Hey @hefan, thanks for your contribution. 

It's not 100% clear to me why you do want to cut off the remove button on option values list. Clicking several times on the `Add Option Value` button, you can have this situation:

Screenshot 2020-03-24 at 09 37 26

And no way to come back to a "clean" state without reloading the page. I think it's not intuitive for admin users to understand that it's legit to leave some empty items and they could feel lost, and that could happen after they added some option values already, which may cause to lose unsaved data.

If this is what we have on Product Properties tab, I think we should change that instead. What do you think?

hefan:
i didn't find that is a bad situation considering the circumstances.

When i compared both screens i thought the product properties screen works better. 
The Option Values screen adds a hidden field to remove entries (which is done by the to be removed helper). i guess to remove the non persistent field you need such a trick like the hidden field.

So it was logical to make it work like the product properties screen, which doesn't do it that way.

What i don't like on both screens is the ordering which doesn't work. The new items should at least appear ALL below the last item. Because after save they are all nevertheless below.

So my proposal would be:
1. Do it like in the product property screen 
3. add a remove button for non persistent fields (maybe with hidden field)
3. Add a default empty entry at the end
4. "Add new" produces new items below all other items
5. do not add a sort option before save (like it is now)

Maybe we do not need a default new entry at the end (3).
This would imply changes on both screens.

What do you think? In general i have the feeling that all entries should be also sortable instantly. But this goes a little bit out of scope of that PR.

kennyadsl:
@hefan I like the plan.

for point 2. I think the hidden field is not needed for non-persisted rows. It's all done client-side, just need to remove that row from the HTML, right?

Also, please, try to split these changes into multiple PR so they will be easier (and faster ;) ) to review from the core team.

hefan:
ok. i'll start with the product properties page then and reference this issue and close it now

PR #3568, opened 2020-03-25
Upgrade from 'feature' to 'system' test

### LINKED ISSUE #3564: Swtich `feature` tests to `system` tests
After https://github.com/solidusio/solidus/pull/3552 was merged, it may be a good time to upgrade our integration `feature` tests to `system` tests. 

Hopefully, the test suite execution time may even improve, as with system tests we don't need to use the`database_cleaner` gem anymore. 

This post may help: https://medium.com/table-xi/a-quick-guide-to-rails-system-tests-in-rspec-b6e9e8a8b5f6

gonzar11:
**Description**
I couldn't remove database cleaner because some specs on backends started to fail. Then I realized that it used just to truncate the database before starting run all specs. This is because some migrations are filling the database.

It closes #3564

**Checklist:**


backend/spec/spec_helper.rb
```diff
@@ -44,6 +44,7 @@
 Capybara.exact = true
 
 require "selenium/webdriver"
+Capybara.javascript_driver = (ENV['CAPYBARA_DRIVER'] || :selenium_chrome_headless).to_sym
```
coorasse: why is this line necessary? isn't
```
config.before(:each, type: :system, js: true) do
  driven_by :selenium_chrome_headless
end
```
already taking care of defining the driver?

gonzar11: Rails uses as default :selenium . If I don't define Capybara.javascript_driver, rails will assume that we use :selenium. In my case it fails because I didn't have geckodriver( Mozilla driver used with selenium)

kennyadsl: That's odd. I checked the branch locally and tried commenting this line and all the `driven_by` below and the result is the same. I think that for some reasons it's just using the default ignoring the `driven_by` calls.

gonzar11: I've tried to print which driver is being used. Please add this line `puts Capybara.current_driver` in both `config.before(:each, type: :system) `and `config.before(:each, type: :system, js: true)` before and after using driven_by. You will see the difference between use and don's use driven_by

kennyadsl: Ok thanks, I'm giving it a try

kennyadsl: if I comment out

```ruby
# Capybara.javascript_driver = (ENV['CAPYBARA_DRIVER'] || :selenium_chrome_headless).to_sym
```

no matter what `Capybara.current_driver` prints, it is opening Firefox to run specs so it's definitely not headless. 
I think there's some misconfiguration happening but I couldn't get exactly where yet.

I've also tried with this syntax, which seems to be the default to make exactly the same thing that we do with our custom driver:

```ruby
driven_by :selenium, using: :headless_chrome, screen_size: [1920, 1080]
```

 but it still opens firefox to run specs. 😢

kennyadsl: @gonzar11 did you get the chance to take a look?

blocknotes: @gonzar11, I worked on a similar PR in a Solidus extension ([solidus_starter_frontend](https://github.com/nebulab/solidus_starter_frontend/pull/22/commits/19606c911c36139f83b9883c2911f8d20fa832ba)) and I would like to add my thoughts.

- `Capybara.javascript_driver = ...` is not necessary; [Rails testing guide](https://guides.rubyonrails.org/testing.html#changing-the-default-settings) (but for Minitest) says that default Capybara driver settings should be changed via `driven_by`;
- in my PR I choose to have 2 different variables (`CAPYBARA_DRIVER`/`CAPYBARA_JS_DRIVER`), it can be useful sometimes;
- I made some tests in this PR commenting out the `Capybara.javascript_driver = ...` line in spec_helper.rb; running a JS spec with `CAPYBARA_DRIVER=selenium_chrome bin/rspec backend/spec/system/admin/configuration/payment_methods_spec.rb:36` first opens a Firefox empty window then a Chrome window with the specified test.
The problem here should be related to the lines:
```ruby
    if RSpec.current_example.metadata[:js] && page.driver.browser.respond_to?(:url_blacklist)
      page.driver.browser.url_blacklist = ['http://fonts.googleapis.com']
    end
```
If you comment them it should work as you expect.

kennyadsl:
Closing as this has no activity anymore. Thanks anyway!

PR #3570, opened 2020-03-29
Remove non-existing middleware

coorasse:
Am I missing something? This middleware does not exist and therefore cannot be loaded.
It should be here, right? https://github.com/solidusio/solidus/tree/master/backend/lib/spree/backend
On the other side, the Frontend one is present: https://github.com/solidusio/solidus/tree/master/frontend/lib/spree/frontend/middleware

This instruction makes the installation of solidus_backend fail. 😢

### DIFF
--- a/backend/lib/spree/backend/engine.rb
+++ b/backend/lib/spree/backend/engine.rb
@@ -5,8 +5,6 @@ require 'spree/backend/config'
 module Spree
   module Backend
     class Engine < ::Rails::Engine
-      config.middleware.use "Spree::Backend::Middleware::SeoAssist"
-
       # Leave initializer empty for backwards-compatability. Other apps
       # might still rely on this event.
       initializer "spree.backend.environment", before: :load_config_initializers do; end


kennyadsl:
Hey @coorasse, thanks for this PR. It's quite creepy that we had this middleware call there but the system was not complaining. Which error are you seeing exactly?

coorasse:
```
NoMethodError:
            undefined method `new' for "Spree::Backend::Middleware::SeoAssist":String
```
I can create a gist later on if needed.

kennyadsl:
Does it happen when you start the rails server on your own host application? And, are you requiring the main `gem 'solidus'` or via individual sub-components (`solidus_backend`, `solidus_core`)?

coorasse:
Individual subcomponents. `solidus_backend` and `solidus_core`.
I am currently investigating, it could be related to:
```
mount Spree::Core::Engine, at: '/'
mount Spree::Backend::Engine, at: '/'
```
?

coorasse:
```
git clone git@github.com:renuo/solidus_engine_issue.git
cd solidus_engine_issue
bin/setup
bundle exec rake
cat log/test.log
```

kennyadsl:
Probably unrelated, there's no need to mount backend routes manually. They are injected into the [`Spree::Core`](https://github.com/solidusio/solidus/blob/68cad802be05f1c0fdc0138f1183e917942e3cf6/backend/config/routes.rb#L3) routes.

This makes me think that maybe we never call `Spree::Backend::Engine` in the whole application and, being not loaded or accessed, it never crashes.

PR #3573, opened 2020-03-31
API uploads images via URL

### LINKED ISSUE #3572: Image API should allow upload via URL

The API for uploading images doesn't allow us to specify a URL of an image as a parameter.   I'm not exactly sure how it works (I spoke with @kennyadsl on the #support Slack channel, and he wasn't exactly sure either). 

I'd like the "attachment" parameter to optionally accept a URL in addition to what it does now.

I tried uploading by base64 encoding a file with the current implementation, it didn't work.  There are apparently no examples of it actually uploading a file (including in the specs, which fakes it by using `fixture_file_upload`)

**Additional context**

At a high level my architecture is something like this: I have 2 rails apps, one is running solidus v2.10 unmodified (I'm treating it like a black box component).  The other app talks to solidus via the api, creating products, etc.  The other app takes image uploads and stashes them on S3.  I want to pass the uploaded image to solidus via the api.  At present, there are no docs and no clear way to actually do that (which is ironic, since one might think this is the exact purpose of the API).

calebhaye:
**Description**

This enhancement give the images API the ability to create images from URLs.  A
simple check determines whether or not the attachment parameter is a URL.  If it
is a URL, OpenURI is used to download the image and it subsequently undergoes
the standard post upload processing (via paperclip).  If it is not a URL, then
the attachment parameter is handled in the same manner it was prior to this
commit.

Fixes #3572

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/images_controller.rb
+++ b/api/app/controllers/spree/api/images_controller.rb
@@ -15,13 +15,23 @@ module Spree
 
       def create
         authorize! :create, Image
-        @image = scope.images.create(image_params)
+
+        @image = scope.images.build(image_params.except(:attachment))
+        @image.attachment = prepared_attachment
+        @image.save
+
         respond_with(@image, status: 201, default_template: :show)
       end
 
       def update
         @image = scope.images.accessible_by(current_ability, :update).find(params[:id])
-        @image.update(image_params)
+        if image_params[:attachment].present?
+          @image.assign_attributes(image_params.except(:attachment))
+          @image.attachment = prepared_attachment
+          @image.save
+        else
+          @image.update image_params
+        end
         respond_with(@image, default_template: :show)
       end
 
@@ -43,6 +53,17 @@ module Spree
         elsif params[:variant_id]
           Spree::Variant.find(params[:variant_id])
         end
+      end
+
+      def prepared_attachment
+        uri = URI.parse image_params[:attachment]
+        if uri.is_a? URI::HTTP
+          URI.open(image_params[:attachment])
+        else
+          image_params[:attachment]
+        end
+      rescue URI::InvalidURIError
+        image_params[:attachment]
       end
     end
   end
--- a/api/spec/requests/spree/api/images_controller_spec.rb
+++ b/api/spec/requests/spree/api/images_controller_spec.rb
@@ -32,6 +32,22 @@ module Spree
         end.to change(Image, :count).by(1)
       end
 
+      it 'can upload a new image from a valid URL' do
+        expect do
+          post spree.api_product_images_path(product.id), params: {
+            image: {
+              attachment: 'https://github.com/solidusio/brand/raw/1827e7afb7ebcf5a1fc9cf7bf6cf9d277183ef11/PNG/solidus-logo-dark.png',
+              viewable_type: 'Spree::Variant',
+              viewable_id: product.master.to_param,
+              alt: 'just a test'
+            },
+          }
+          expect(response.status).to eq(201)
+          expect(json_response).to have_attributes(attributes)
+          expect(json_response[:alt]).to eq('just a test')
+        end.to change(Image, :count).by(1)
+      end
+
       context "working with an existing product image" do
         let!(:product_image) { product.master.images.create!(attachment: image('thinking-cat.jpg')) }
 
@@ -67,6 +83,20 @@ module Spree
           expect(response.status).to eq(200)
           expect(json_response).to have_attributes(attributes)
           expect(product_image.reload.position).to eq(2)
+        end
+
+        it "can update image URL" do
+          expect(product_image.position).to eq(1)
+          put spree.api_variant_image_path(product.master.id, product_image), params: {
+            image: {
+              position: 2,
+              attachment: 'https://github.com/solidusio/brand/raw/1827e7afb7ebcf5a1fc9cf7bf6cf9d277183ef11/PNG/solidus-logo-dark.png'
+            },
+          }
+          expect(response.status).to eq(200)
+          expect(json_response).to have_attributes(attributes)
+          expect(product_image.reload.position).to eq(2)
+          expect(product_image.reload.attachment_height).to eq(420)
         end
 
         it "can delete an image" do
--- a/api/openapi/api.oas2.yml
+++ b/api/openapi/api.oas2.yml
@@ -5885,6 +5885,7 @@ definitions:
         type: string
       attachment:
         type: string
+        description: 'This field can be used to pass an image URL. When used in this manner, the image undergoes the standard post upload processing via paperclip.'
       position:
         type: integer
       viewable_type:


kennyadsl:
@calebhaye I think that HoundCI is complaining at valid things, can you please take a look?

api/spec/requests/spree/api/images_controller_spec.rb
```diff
@@ -17,20 +17,34 @@ module Spree
...
-              attachment: upload_image('thinking-cat.jpg'),
-              viewable_type: 'Spree::Variant',
-              viewable_id: product.master.to_param
-            },
-          }
-          expect(response.status).to eq(201)
-          expect(json_response).to have_attributes(attributes)
-        end.to change(Image, :count).by(1)
-      end
+        it "from a valid URL" do
```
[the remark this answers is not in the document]
calebhaye: fixed, squashed into the sole commit.

calebhaye:
Okay, all ready :)

api/app/controllers/spree/api/images_controller.rb
```diff
@@ -44,6 +48,13 @@ def scope
           Spree::Variant.find(params[:variant_id])
         end
       end
+
+      def valid_url?(url)
+        uri = URI.parse url
+        uri.is_a?(URI::HTTP) && uri.host.present?
```
kennyadsl, line 54: Why do we need to check the host here?

api/spec/requests/spree/api/images_controller_spec.rb
```diff
@@ -32,6 +32,20 @@ module Spree
         end.to change(Image, :count).by(1)
       end
 
+      it 'can upload a new image from a valid URL' do
+        expect do
+          post spree.api_product_images_path(product.id), params: {
+            image: {
+              attachment: 'https://www.google.com/images/branding/googlelogo/2x/googlelogo_color_92x30dp.png',
```
kennyadsl, line 39: What about using an image that belongs to Solidus, e.g. https://github.com/solidusio/brand/raw/1827e7afb7ebcf5a1fc9cf7bf6cf9d277183ef11/PNG/solidus-logo-dark.png ?

This way the test suite is safe, whatever Google decides to do with their assets.

api/app/controllers/spree/api/images_controller.rb
calebhaye: Maybe we don't?  Will fix.

api/spec/requests/spree/api/images_controller_spec.rb
calebhaye: I like that suggestion.  I was definitely unsure about this one.   Will fix

api/app/controllers/spree/api/images_controller.rb
calebhaye: fixed in latest commit.

api/spec/requests/spree/api/images_controller_spec.rb
calebhaye: fixed in latest commit.

calebhaye:
Hi @kennyadsl,

The current implementation has zilch for docs, which is probably why neither of us know if it really works or not.  I could add something for docs, but I honestly don't know where/how to do that.  Will you point me in the right direction?  I'd like to wrap this one up soon if we can 👍, and at present it seems like the only blocker for merging is your request for docs. 

Thank you

calebhaye:
> Thanks a lot!
> 
> I left a couple of notes, can we also add some documentation in this API endpoint description?

Hi @kennyadsl,

The notes you left have been addressed.

The current implementation has zilch for documentation in the API endpoint description, which is probably why neither of us know if it really works or not.  I could add something for docs, but I honestly don't know where/how to do that.  Will you please point me in the right direction?  I'd like to wrap this one up soon if we can 👍, and at present it seems like the only blocker for merging is your request for documentation. 

Thank you

api/app/controllers/spree/api/images_controller.rb
```diff
@@ -15,7 +15,11 @@ def show
 
       def create
         authorize! :create, Image
-        @image = scope.images.create(image_params)
+        if valid_url? image_params[:attachment]
+          @image = scope.images.create(attachment: URI.open(image_params[:attachment]))
```
kennyadsl, line 19: Looking twice, is this removing the other attributes in `image_params` and creating an image with only the attachment? I think we should update the `image_params` instead, preserving the rest of the attributes.

kennyadsl:
For the documentation, I mean adding a small description [here](https://github.com/solidusio/solidus/blob/0f76bac523d557d3aff430c50f55d8fa832be5c5/api/openapi/api.oas2.yml#L5887), something like:

```
      attachment:
        type: string
        description: |
          The URL of the image to attach.
```

Also, can you please squash commits into a single one?

Thanks again.

api/app/controllers/spree/api/images_controller.rb
calebhaye: I think it may be.  I'll do some debugging soon and get back to you.

calebhaye:
> For the documentation, I mean adding a small description [here](https://github.com/solidusio/solidus/blob/0f76bac523d557d3aff430c50f55d8fa832be5c5/api/openapi/api.oas2.yml#L5887), something like:
> 
> ```
>       attachment:
>         type: string
>         description: |
>           The URL of the image to attach.
> ```
> 
> Also, can you please squash commits into a single one?
> 
> Thanks again.

Thanks for the pointer.  If that's the prescribed methodology, I'll add the docs there.  I noticed that the other API controllers (like the [create method in the ProductsController](https://github.com/solidusio/solidus/blob/0f76bac523d557d3aff430c50f55d8fa832be5c5/api/app/controllers/spree/api/products_controller.rb#L40), for instance), just have code comments and no actual docs that manifest on [Stoplight](https://solidus.docs.stoplight.io/api-reference/products/create-product).   I'll leave it up to you, should I add the docs as a comment, add them to the aforementioned reference point suggested by you, or both?  Please advise @kennyadsl thx

api/app/controllers/spree/api/images_controller.rb
calebhaye: @kennyadsl  Good catch. I added a fix that addresses your concern.  The line following the line you commented on now updates the resultant `Spree::Image` object with the other  attributes within `image_params`, except for `:attachment`

I tried updateing `image_params` instead, as per your suggestion, but that resulted in the following error when running the spec: 

```
Paperclip::AdapterRegistry::NoHandlerError:
       No handler found for "https://github.com/solidusio/brand/raw/1827e7afb7ebcf5a1fc9cf7bf6cf9d277183ef11/PNG/solidus-logo-dark.png"

```

So, I opted to create the image first using `image_params[:attachment]`, and then update all the other attributes using `ActionController::Parameters` `except` method.

I also added a test that fails without the newly added line

calebhaye:
Should be good to go 🤞

api/app/controllers/spree/api/images_controller.rb
kennyadsl: It sounds weird... I'll try to take a look soon and see if I find why this happens. I'd prefer if we can make a single `INSERT` query here, instead of also updating the record later.

kennyadsl: I'm trying locally and I got what you mean now. It looks like a bug in paperclip: if it doesn't have the adapter with `create` and all the rest of attributes I can't get why it has the adapter if you create the record only with the attachment one...

Anyway, I think we can avoid the double query with this code:

```ruby
# frozen_string_literal: true

module Spree
  module Api
    class ImagesController < Spree::Api::BaseController
      def create
        authorize! :create, Image

        @image = scope.images.build(image_params.except(:attachment))
        @image.attachment = prepared_attachment
        @image.save

        respond_with(@image, status: 201, default_template: :show)
      end

      private

      def prepared_attachment
        uri = URI.parse image_params[:attachment]

        if uri.is_a? URI::HTTP
          URI.open(image_params[:attachment])
        else
          image_params[:attachment]
        end
      rescue URI::InvalidURIError
        image_params[:attachment]
      end
    end
  end
end
```

I'm also wondering if we need something similar in the update method.

aldesantis: @calebhaye I was wondering if you had a chance to look at this? Would be really cool to get this merged. 🙂

calebhaye: @aldesantis yep!  Just pushed the latest, which includes an updated `update` method and a new spec for testing the update method when updating the attachment property.

@kennyadsl please note: I originally updated the `update` method to look like this:

```
@image.assign_attributes(image_params.except(:attachment))
@image.attachment = prepared_attachment if image_params[:attachment].present?
@image.save
```
But I changed it to the following because I feel as though it reads more logically:  
```
if image_params[:attachment].present?
  @image.assign_attributes(image_params.except(:attachment))
  @image.attachment = prepared_attachment
  @image.save
else
  @image.update image_params
end
```
I'd rather have it make more sense to developers who reference the code in the future, than to have fewer lines

PR #3574, opened 2020-04-02
[Backend] Handle errors and flash messages editing a taxon

softr8:
## Description
If there was any error when trying to edit a taxon, it
was not showing any error and just redirecting to taxons index,
this change renders edit form showing errors correctly instead
of just redirect without any flash message.

**Checklist:**

### DIFF
--- a/api/spec/requests/spree/api/taxons_controller_spec.rb
+++ b/api/spec/requests/spree/api/taxons_controller_spec.rb
@@ -189,7 +189,7 @@ module Spree
       end
 
       it "cannot create a new taxon with invalid attributes" do
-        post spree.api_taxonomy_taxons_path(taxonomy), params: { taxon: { foo: :bar } }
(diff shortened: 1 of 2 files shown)


api/spec/requests/spree/api/taxons_controller_spec.rb
```diff
@@ -189,7 +189,7 @@ module Spree
       end
 
       it "cannot create a new taxon with invalid attributes" do
-        post spree.api_taxonomy_taxons_path(taxonomy), params: { taxon: { foo: :bar } }
+        post spree.api_taxonomy_taxons_path(taxonomy), params: { taxon: { name: '' } }
```
spaghetticode, line 192: I'm not sure about the reason why for this change, can you explain to me? thanks 🙏

softr8: It was passing due the wrong check, it was actually passing because `foo` is not a permitted param, and using name it actually runs the validations

PR #3577, opened 2020-04-03
Making simple coordinator insufficient stock to include a message

softr8:
It is hard to know what items were the ones that were
unavailable, this change includes item names that
were unable to be fullfilled, then this message
can be used in the frontend easily.

What do you think about this?

I saw that the frontend does some complex code to achieve kind of the same [here](https://github.com/solidusio/solidus/blob/99c6ba9537526d11b255a58facffbb8cf328767f/frontend/app/controllers/spree/checkout_controller.rb#L251)

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -26,7 +26,14 @@ module Spree
     include Spree::Order::Checkout
     include Spree::Order::Payments
 
-    class InsufficientStock < StandardError; end
+    class InsufficientStock < StandardError
+      attr_reader :items
+
+      def initialize(message = nil, items: {})
+        @items = items
+        super message
+      end
+    end
     class CannotRebuildShipments < StandardError; end
 
     extend Spree::DisplayMoney
--- a/core/app/models/spree/stock/simple_coordinator.rb
+++ b/core/app/models/spree/stock/simple_coordinator.rb
@@ -51,9 +51,7 @@ module Spree
         # Allocate any available on hand inventory and remaining desired inventory from backorders
         on_hand_packages, backordered_packages, leftover = @allocator.allocate_inventory(@desired)
 
-        unless leftover.empty?
-          raise Spree::Order::InsufficientStock
-        end
+        raise Spree::Order::InsufficientStock.new(items: leftover.quantities) unless leftover.empty?
 
         packages = @stock_locations.map do |stock_location|
           # Combine on_hand and backorders into a single package per-location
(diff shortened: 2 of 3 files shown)


core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -52,7 +52,9 @@ def build_shipments
         on_hand_packages, backordered_packages, leftover = @allocator.allocate_inventory(@desired)
 
         unless leftover.empty?
-          raise Spree::Order::InsufficientStock
+          item_names = leftover.variants.map(&:name).to_sentence
+          message = I18n.t('spree.inventory_error_flash_for_insufficient_quantity', names: item_names)
+          raise Spree::Order::InsufficientStock.new(message)
```
spaghetticode, line 57: What about enhancing this spec https://github.com/solidusio/solidus/blob/0f76bac523d557d3aff430c50f55d8fa832be5c5/core/spec/models/spree/stock/simple_coordinator_spec.rb#L130 to check also for the message?

kennyadsl, line 57: I like the idea but I think we can take advantage of creating a custom exception that contains the unavailable items and let the consumer decide what to do with those. 

What I have in mind is changing the [exception definition](https://github.com/solidusio/solidus/blob/78ccf7f7ca6ac625328a9211225a17d71ab3c991/core/app/models/spree/order.rb#L29) to something like:

```ruby
class InsufficientStock < StandardError
  attr_reader :items

  def initialize(message=nil, items: [])
    super(message)

    @items = items
  end
end
```

and push names in it with:

```ruby
raise Spree::Order::InsufficientStock.new(items: leftovers)
```

At that point, we can consume the exception data in different ways depending on the context, even using `rescue_from` since its [handler methods accept an argument](https://apidock.com/rails/ActiveSupport/Rescuable/ClassMethods/rescue_from) that will be the exception containing the items. 

```ruby
rescue_from Spree::Order::InsufficientStock, with: :something 

def something(exception)
  # exception.items is available here!
end
```

What do you think?

softr8: I like it! will refactor my code to take this path

softr8:
What do you think about this new approach @kennyadsl

core/app/models/spree/order.rb
```diff
@@ -26,7 +26,14 @@ class Order < Spree::Base
     include Spree::Order::Checkout
     include Spree::Order::Payments
 
-    class InsufficientStock < StandardError; end
+    class InsufficientStock < StandardError
+      attr_reader :items
+
+      def initialize(message = nil, items: [])
```
jarednorman, line 32: Should the default value of `items:` be an array or a hash?

softr8: True! it returns a hash of items

kennyadsl:
Rerunning specs, it's a flaky one!

Issue #3566, opened 2020-03-23
No customer return endpoint for solidus api


roshita02:
I am working with solidus api and there seems to be no endpoint for creating customer returns for store admins after creating the return authorization.

**Solidus Version**:
2.10.0.beta1 

**To Reproduce**
1. Create return authorization.

**Current behavior**
Currently, solidus api has endpoint for creating return authorization. After i create a return authorization by the given endpoint, corresponding return item is created but with nil customer_return_id.  

**Expected behavior**
I studied the solidus docs and as mentioned in it, After store admin gets the item when customer returns it, Admin has to create customer return with corresponding return item id. But there is no any endpoint to create customer returns with the created return item as attribute.

kennyadsl:
Thanks for pointing this out. There are no endpoints for customer returns yet. Any contribution is welcome to add them!

kennyadsl:
Thanks for opening the issue @roshita02! @seand7565 implemented this and we can close here. Also thanks to SuperGood for sponsoring this PR.

PR #3579, opened 2020-04-08
Add API endpoint for customer returns

seand7565:

**Description**

  Adds an API endpoint that allows you to manage customer returns. Also adds a test suite for this new API endpoint, adds associated jbuilder files, adds customer_return.number to ransackable attributes, and creates a connection between order and customer_return for easier referencing. 

  Just like return_items, when you create the customer_return you'll also need to add attributes for the return_items associated with the return.

  Quick note: The way I'm referencing customer returns from the order is a bit convoluted, because there's no order_id on customer_returns. I think there should be, as it would simplify this a bit (and it makes sense that customer returns would be connected to the order like return authorizations are) - but it felt like it was outside of the scope of this PR.

  Ref #3566 

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/customer_returns_controller.rb
+++ b/api/app/controllers/spree/api/customer_returns_controller.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+module Spree
+  module Api
+    class CustomerReturnsController < Spree::Api::BaseController
+      before_action :load_order
+      around_action :lock_order, only: [:create, :update, :destroy, :cancel]
+
+      rescue_from Spree::Order::InsufficientStock, with: :insufficient_stock_error
+
+      def create
+        authorize! :create, CustomerReturn
+        @customer_return = CustomerReturn.create(customer_return_params)
+        if @customer_return.save
+          respond_with(@customer_return, status: 201, default_template: :show)
+        else
+          invalid_resource!(@customer_return)
+        end
+      end
+
+      def index
+        authorize! :index, CustomerReturn
+
+        @customer_returns = @order.
+          customer_returns.
+          accessible_by(current_ability, :read).
+          ransack(params[:q]).
+          result
+
+        @customer_returns = paginate(@customer_returns)
+
+        respond_with(@customer_returns)
+      end
+
+      def new
+        authorize! :new, CustomerReturn
+      end
+
+      def show
+        authorize! :show, CustomerReturn
+        @customer_return = @order.customer_returns.accessible_by(current_ability, :read).find(params[:id])
+        respond_with(@customer_return)
+      end
+
+      def update
+        authorize! :update, CustomerReturn
+        @customer_return = @order.customer_returns.accessible_by(current_ability, :update).find(params[:id])
+        if @customer_return.update(customer_return_params)
+          respond_with(@customer_return.reload, default_template: :show)
+        else
+          invalid_resource!(@customer_return)
+        end
+      end
+
+      private
+
+      def load_order
+        @order ||= Spree::Order.find_by!(number: order_id)
+        authorize! :read, @order
+      end
+
+      def customer_return_params
+        params.require(:customer_return).permit(permitted_customer_return_attributes)
+      end
+    end
+  end
+end
--- a/api/config/routes.rb
+++ b/api/config/routes.rb
@@ -40,6 +40,8 @@ Spree::Core::Engine.routes.draw do
           put :cancel
         end
       end
+
+      resources :customer_returns, except: :destroy
     end
 
     resources :checkouts, only: [:update], concerns: :order_routes do
--- a/api/spec/requests/spree/api/customer_returns_controller_spec.rb
+++ b/api/spec/requests/spree/api/customer_returns_controller_spec.rb
@@ -1,0 +1,135 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+module Spree
+  describe Api::CustomerReturnsController, type: :request do
+    let!(:order) { create(:shipped_order) }
+    let(:attributes) { [:id, :number, :stock_location_id] }
+
+    before do
+      stub_authentication!
+    end
+
+    context "as a non admin" do
+      before do
+        allow_any_instance_of(Order).to receive_messages user: create(:user)
+      end
+
+      it "cannot see any customer returns" do
+        get spree.api_order_customer_returns_path(order)
+
+        assert_unauthorized!
+      end
+
+      it "cannot see a single customer return" do
+        get spree.api_order_customer_return_path(order, 1)
+
+        assert_unauthorized!
+      end
+
+      it "cannot learn how to create a new customer return" do
+        get spree.new_api_order_customer_return_path(order)
+
+        assert_unauthorized!
+      end
+
+      it "cannot update a customer return" do
+        put spree.api_order_customer_return_path(order, 0)
+
+        assert_unauthorized!
+      end
+
+      it "cannot create a new customer return" do
+        post spree.api_order_customer_returns_path(order)
+
+        assert_unauthorized!
+      end
+    end
+
+    context "as an admin" do
+      sign_in_as_admin!
+
+      it "can show customer return" do
+        customer_return = FactoryBot.create(:customer_return)
+
+        get spree.api_order_customer_return_path(customer_return.order, customer_return.id)
+
+        expect(response.status).to eq(200)
+        expect(json_response).to have_attributes(attributes)
+      end
+
+      it "can get a list of customer returns" do
+        FactoryBot.create(:customer_return, shipped_order: order)
+        FactoryBot.create(:customer_return, shipped_order: order)
+
+        get spree.api_order_customer_returns_path(order), params: { order_id: order.number }
+
+        expect(response.status).to eq(200)
+
+        customer_returns = json_response["customer_returns"]
+
+        expect(customer_returns.first).to have_attributes(attributes)
+        expect(customer_returns.first).not_to eq(customer_returns.last)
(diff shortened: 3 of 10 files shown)


seand7565:
@kennyadsl Squashed & documented! I added to the stoplight documentation, let me know how it looks. Thanks!

ccarruitero:
Thanks @seand7565 for the PR.

In admin workflow, in order to create a `CustomerReturn`, you need to create a `ReturnAuthorization` first.
https://guides.solidus.io/developers/returns/overview.html
https://guides.solidus.io/developers/returns/customer-returns.html

Until I understand, that is because you have to accept or not the return, and setup the `ReturnReason` and `ReimbursementType` for the `ReturnItem`

And after generate the `CustomerReturn`, you associate the `ReturnItem`s that you will receive.

So, I think we should receive a `return_authorization_id` and list of `ReturnItem`s parameters for create a `CustomerReturn`. Then, we should validate that the `ReturnAuthorization` is accepted for generate the new ``CustomerReturn and associate the given return items with the generated `CustomerReturn`.

Also, we should need to generate a `Reimbursement` for the items that the customer return

api/app/controllers/spree/api/customer_returns_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+        @customer_return = CustomerReturn.create(customer_return_params)
+        if @customer_return.save
+          respond_with(@customer_return, status: 201, default_template: :show)
+        else
+          invalid_resource!(@customer_return)
+        end
+      end
+
+      def index
+        authorize! :admin, CustomerReturn
```
ccarruitero, line 22: I think all actions should require admin authorization, so maybe can use `authorize_resource` on top for set the authorization for all actions

api/app/controllers/spree/api/customer_returns_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+          result
+
+        @customer_returns = paginate(@customer_returns)
+
+        respond_with(@customer_returns)
+      end
+
+      def new
+        authorize! :admin, CustomerReturn
+      end
```
ccarruitero, line 37: I think this action is useless for an api

api/app/controllers/spree/api/customer_returns_controller.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module Spree
+  module Api
+    class CustomerReturnsController < Spree::Api::BaseController
```
ccarruitero, line 5: In think should be better inherit from `ResourceController`

seand7565:
Hey @ccarruitero, thanks for the review!

On requiring RMAs - I tried to model the API the same way the backend works. Currently, an RMA is _NOT_ required for creating a Customer Return. You can create a customer return with no RMA, and it will create the customer return successfully. However, it will generate an “acceptance error” of `{:rma_required=>"Return item requires an RMA”}` - this does not prevent you from saving the Customer Return, nor will it prevent you from “receiving” it - it will toss the Customer Return into the “Rejected” column though, and will not create a reimbursement. 

While I don’t like the way this is handled (there are a lot of Return Item statuses that don’t make sense with RMAs like “given to customer” or “lost in transit” - you wouldn’t be expecting a return in those situations but might still want a reimbursement), changing the functionality of how this works is another issue - one that I might tackle very shortly, but outside of the scope of this PR. 

Reimbursements AFAIK are handled by return_item state changes - setting the reception status of the customer return should trigger the reimbursement handler. 

I’ll take a look at your comments on the code a bit later - thanks for the input!

api/app/controllers/spree/api/customer_returns_controller.rb
kennyadsl: We are using the `new` endpoint to educate the client about which fields are available in this resource. I don't know where this pattern comes from and if it's legit but I think we should keep the convention here.

api/app/controllers/spree/api/customer_returns_controller.rb
kennyadsl: I checked and `Spree::Api::ResourceController` is only used in `Spree::Api::UsersController`. Maybe it was the wrong abstraction and it's better to deprecate it at this point (in another PR)?

kennyadsl:
Thanks @seand7565 and @ccarruitero for the review!

The "RMA required" part is a bit tricky. By default, you need to have a RMA to create valid Return Items and Customer Returns. 

This depends on a specific validator set [here](https://github.com/solidusio/solidus/blob/138bf619647b125a53c8898fce6d01a9dee2c7a2/core/app/models/spree/return_item/eligibility_validator/default.rb#L11) that can be changed depending on the needs.

For now, I'll keep it as in admin even if it's not ideal.

api/app/controllers/spree/api/customer_returns_controller.rb
softr8: We had to deal with a lot of permission sets in solidus marketplace extension, `:admin` is not that friendly when it comes to grant specific actions, maybe use the same idea than resources controller and authorize each action by its name? that way we can allow to show customer returns to specific roles but without the ability to update the record

seand7565: @softr8 Yes good point, I'll make that change and push shortly! (also have to update the documentation which will take some time 😄 ) 

@ccarruitero Sorry for taking so long to get back to you! I think `authorize_resource` _could_ be something that simplifies a lot of our API actions, but we don't use it anywhere in our code currently, and I'd prefer to stick with existing conventions for this PR. However, if you want to open an issue to migrate to using `authorize_resource`, please do! I think it would be a welcome addition, removing lots of lines of code from the controllers that currently authorize on a per-action basis.

api/config/routes.rb
```diff
@@ -40,6 +40,8 @@
           put :cancel
         end
       end
+
+      resources :customer_returns
```
aldesantis, line 44: Can you exclude the `:destroy` action from here? Otherwise it will also generate a `DELETE` route that raises an error. In fact, we have a ton of such "ghost" routes in the API unfortunately...

core/app/models/spree/order.rb
```diff
@@ -105,6 +105,8 @@ def states
 
     # Returns
     has_many :return_authorizations, dependent: :destroy, inverse_of: :order
+    has_many :return_items, through: :inventory_units
+    has_many :customer_returns, through: :return_items
```
aldesantis, line 109: Have you only added these for the tests? If so, I think you should move the logic in the tests. I like having these associations but I don't know if they belong to this PR. Also, I'm not 100% sure this is the right way to associate returns to orders, it would probably be much easier and less error-prone to do it directly as you have proposed. I think we need to be mindful of people doing weird stuff (e.g. moving inventory units between orders maybe? 😄).

api/spec/requests/spree/api/customer_returns_controller_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        assert_unauthorized!
+      end
+    end
+
+    context "as an admin" do
+      sign_in_as_admin!
+
+      it "can show customer return" do
+        customer_return = FactoryBot.create(:customer_return)
+        get spree.api_order_customer_return_path(customer_return.order, customer_return.id)
```
aldesantis, line 50: Can you add empty lines between the setup, exercise and verify phases of this and all other tests? It makes them much easier to read. Thanks!

core/app/models/spree/order.rb
seand7565: About this - I felt the way that customer returns were accessed from the admin side was too convoluted for what I wanted to write (https://github.com/solidusio/solidus/blob/1aa554fb04d2e28d279ec28a6ddb36b6ca2e5956/backend/app/controllers/spree/admin/customer_returns_controller.rb#L43) so I added the associations to be able to reference the customer_return from the order - inventory_units was the closest association I could find that has an order_id, but I had to use return_items as an intermediary connection. It's necessary for the API endpoint to be able to find the return_items for any given order.

I would much rather add an order_id to customer_returns (as I think they don't make sense outside the context of an order anyway) but I thought it might be a bit out of the scope of this PR (and I would want to update the way the backend route fetches the customer_returns as well, which is definitely out of the scope of this PR), so I was going to open an issue about it when this PR was merged... but if you'd prefer, I can definitely add the order_id to simplify this association.

api/config/routes.rb
seand7565: Yep, good point - thanks!

api/spec/requests/spree/api/customer_returns_controller_spec.rb
seand7565: Done - let me know how that looks!

api/spec/requests/spree/api/customer_returns_controller_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+                         reception_status_event: "receive",
+                       }] }
+
+        post spree.api_order_customer_returns_path(order), params: { order_id: order.number, customer_return: cr_params }
+
+        expect(response.status).to eq(201)
+        expect(json_response).to have_attributes(attributes)
+
+        customer_return = Spree::CustomerReturn.last
+
```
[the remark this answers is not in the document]
seand7565: Working with a new IDE, still trying to get the defaults down 🙄 I'll have this fixed in a bit!

core/app/models/spree/order.rb
aldesantis: Didn't realize this was being used in the controller. I think the new association makes sense as a first step, and we can always add the new DB column later (which I still think is the best way to go).

PR #3582, opened 2020-04-09
Guides: add basic Google Analytics integration

kennyadsl:
**Description**

📊 

This PR adds a new page that describes how to set up a basic integration with Google Analytics Enhanced Ecommerce with gtag.js.

It's relevant because we archived solidus_trackers extension. It wasn't really providing any value since it was just a matter of adding a file in your store.


seandawes:
Looks great. Another good one to piggyback off this could be Tag Manager as its simply to do and great addition to GA. Blog post would be great about this + tag manager as well. Similar to @aldesantis blog post, if stores can do tag manager/GA non devs can implement a lot on their own

PR #3584, opened 2020-04-10
Deprecate Spree::Address firstname, lastname and full_name

filippoliverani:
**Description**
This PR is a step to solve #3234 and a continuation of #3524.

`Address#name` is ready to be used instead of `firstname`, `lastname` and `full_name`, we should start to deprecate legacy attributes to be ready to remove them in the next major release.
This PR adds deprecation warnings and removes any remaining reference to legacy attributes. 

**Checklist:**

### DIFF
--- a/api/openapi/api.oas2.yml
+++ b/api/openapi/api.oas2.yml
@@ -5078,12 +5078,15 @@ definitions:
         type: string
       firstname:
         type: string
+        description: '*Deprecated: will be removed in Solidus 3.0, please use `name` property*'
       full_name:
         type: string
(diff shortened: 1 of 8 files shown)


api/openapi/api.oas2.yml
```diff
@@ -5078,12 +5078,15 @@ definitions:
         type: string
       firstname:
         type: string
+        description: '*Deprecated: will be removed in Solidus 3.0, please use `name` property*'
```
kennyadsl, line 5081: I think there's a `deprecate` keyword that we can use here.

PR #3585, opened 2020-04-13
Do not display non-eligible adjustments in the admin cart overview

coorasse:
See #2860 and #2484 .
This PR tackles the points discussed in the previous PR from 2018.
In particular it adds a regression test for the issue.
It also improves another test, since it was testing nothing

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/order/details_adjustments.js
+++ b/backend/app/assets/javascripts/spree/backend/views/order/details_adjustments.js
@@ -10,11 +10,7 @@ Spree.Views.Order.DetailsAdjustments = Backbone.View.extend({
     collection
       .map(function(item) {
         return (item.get("adjustments") || [])
-          .filter(function(adjustment) {
-            if (adjustment.eligible === true) {
-              return adjustment;
-            }
-          });
+          .filter(function(adjustment) { return (adjustment.eligible === true); });
       })
       .flatten(true)
       .each(function(adjustment){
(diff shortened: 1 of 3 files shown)


backend/spec/features/admin/orders/adjustments_spec.rb
```diff
@@ -40,8 +41,11 @@
       end
     end
 
-    it "only shows eligible adjustments" do
-      expect(page).not_to have_content("ineligible")
```
coorasse, line 44: This page shows both eligible and non-eligible adjustments, and that's correct, because it highlights if they are not eligiblew by adding the class `adjustment-ineligible`. This test now checks that.

backend/spec/features/admin/orders/order_details_spec.rb
```diff
@@ -225,6 +225,21 @@
...
+            o.adjustments.create!(order: order, label: 'Rebate', amount: 10)
+          end
+        end
+
+        it "shows only eligible adjustments" do
+          visit spree.cart_admin_order_path(order)
+          expect(page).to have_content("Rebate")
+          expect(page).not_to have_content("Non-Eligible")
+        end
+      end
```
coorasse, line 241: this is the regression test.

backend/app/assets/javascripts/spree/backend/views/order/details_adjustments.js
```diff
@@ -8,7 +8,14 @@ Spree.Views.Order.DetailsAdjustments = Backbone.View.extend({
     var totals = {};
     var collection = this.collection ? this.collection.chain() : _.chain([this.model]);
     collection
-      .map(function(item){ return item.get("adjustments") || [] })
+      .map(function(item) {
+        return (item.get("adjustments") || [])
+          .filter(function(adjustment) {
```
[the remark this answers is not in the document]
kennyadsl: @coorasse I think @houndci-bot is right here, we should return something anyway.

kennyadsl:
P.S. I've re-run failed Netlify checks, should be all green now but please take a look at the HoundCI complain.

aldesantis:
@kennyadsl Netlify is still failing here. 😢

PR #3592, opened 2020-04-21
Unified Handling of Option Values and Product Properties List

hefan:
1. Add a default empty entry if there is no entry
2. "Add new" dynamically produces new items below all other items
3. do never add a sort option before save
4. add a remove button for non persistent fields
5. deprecate remove_button helper

see also #3561 and #3581

The PR removes the mix of removing(nonpersistent) and deleting(persistent) records.
It also unifies the way new records are added (the are always marked as new and don't copy ids of other table rows)

One Catch is remaining (as already before this PR): 
If you delete the first empty default entry you cannot add new entries. You have to reload the Page to see a new empty entry

possible solutions:
1. do never add remove buttons for non persistent fields. You voted against that in #3561. With Commit a08b0223a2d442f9845e4b257f5e8cfcadc05a01 this would be done, the following commits handle the remove functionality.

2. Do somehow rewrite the JS `$('.spree_add_fields').click(function() {})` here https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/admin.js to not need a template or to get the template for a new row from another place.
Any Suggestions for that?

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -73,7 +73,7 @@ Spree.ready(function(){
       var el = $(this);
       el.prop('href', '#');
     })
-    $(target).prepend(new_table_row);
+    $(target).append(new_table_row);
   })
 
   $('body').on('click', '.delete-resource', function() {
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -58,6 +58,14 @@ Spree.ready(function(){
   $('.spree_add_fields').click(function() {
     var target = $(this).data("target");
     var new_table_row = $(target + ' tr:visible:last').clone();
+
+    // remove id
+    new_table_row.attr("id", "");
+
+    // Remove sort handle
+    new_table_row.find("td").first().empty();
+
+    // set unique form ids and names of new item
     var new_id = new Date().getTime() + (uniqueId++);
     new_table_row.find("input, select").each(function () {
       var el = $(this);
@@ -66,13 +74,10 @@ Spree.ready(function(){
       el.prop("id", el.prop("id").replace(/\d+(?=[^\d]*$)/, new_id))
       el.prop("name", el.prop("name").replace(/\d+(?=[^\d]*$)/, new_id))
     })
-    // When cloning a new row, set the href of all icons to be an empty "#"
-    // This is so that clicking on them does not perform the actions for the
-    // duplicated row
-    new_table_row.find("a").each(function () {
-      var el = $(this);
-      el.prop('href', '#');
-    })
+
+    // Add a remove button to the actions column of the new row
+    new_table_row.find("td").last().empty().append('<a class="spree_remove_fields no-text with-tip fa fa-trash icon_link with-tip" data-action="remove" href="#" data-original-title="Remove"><span class="text"></span></a>');
+
     $(target).append(new_table_row);
   })
 
--- a/backend/app/assets/javascripts/spree/backend/admin.js
+++ b/backend/app/assets/javascripts/spree/backend/admin.js
@@ -106,26 +106,10 @@ Spree.ready(function(){
 
   $('body').on('click', 'a.spree_remove_fields', function() {
     var el = $(this);
-    el.prev("input[type=hidden]").val("1");
-    el.closest(".fields").hide();
-    if (el.prop("href").substr(-1) == '#') {
-      el.parents("tr").fadeOut('hide');
-    }else if (el.prop("href")) {
-      Spree.ajax({
-        type: 'POST',
-        url: el.prop("href"),
-        data: {
-          _method: 'delete',
-        },
-        success: function(response) {
-          el.parents("tr").fadeOut('hide');
-        },
-        error: function(response, textStatus, errorThrown) {
-          show_flash('error', response.responseText);
-        }
-
-      })
-    }
+    var table_row = el.parents("tr").first();
+    table_row.fadeOut("hide", function() {
+      table_row.remove();
+    });
     return false;
   });
 
--- a/backend/app/views/spree/admin/option_types/_option_value_fields.html.erb
+++ b/backend/app/views/spree/admin/option_types/_option_value_fields.html.erb
@@ -7,5 +7,12 @@
   </td>
   <td class="name"><%= f.text_field :name %></td>
   <td class="presentation"><%= f.text_field :presentation %></td>
-  <td class="actions"><%= link_to_remove_fields t('spree.actions.remove'), f, no_text: true %></td>
+  <td class="actions">
+    <% if f.object.persisted? %>
+      <%= link_to_delete f.object, no_text: true %>
+    <% else %>
+      <%= link_to_with_icon('trash', 'remove', '#', no_text: true, class: "spree_remove_fields",
+            data: { action: 'remove' }, title: t('spree.actions.remove')) %>
+    <% end %>
+  </td>
 </tr>
--- a/backend/app/views/spree/admin/product_properties/_product_property_fields.html.erb
+++ b/backend/app/views/spree/admin/product_properties/_product_property_fields.html.erb
@@ -12,8 +12,13 @@
     <%= f.text_field :value %>
   </td>
   <td class="actions">
-    <% if f.object.persisted? && can?(:destroy, f.object) %>
-      <%= link_to_delete f.object, no_text: true %>
+    <% if f.object.persisted? %>
+      <% if can?(:destroy, f.object) %>
+        <%= link_to_delete f.object, no_text: true %>
+      <% end %>
+    <% else %>
+      <%= link_to_with_icon('trash', 'remove', '#', no_text: true, class: "spree_remove_fields",
+            data: { action: 'remove' }, title: t('spree.actions.remove')) %>
     <% end %>
   </td>
 </tr>
--- a/backend/spec/features/admin/products/properties_spec.rb
+++ b/backend/spec/features/admin/products/properties_spec.rb
@@ -88,25 +88,30 @@ describe "Properties", type: :feature do
     it "successfully create and then remove product property" do
       fill_in_property
 
-      check_property_row_count(2)
+      check_persisted_property_row_count(1)
 
       delete_product_property
 
-      check_property_row_count(1)
+      check_persisted_property_row_count(0)
     end
 
     # Regression test for https://github.com/spree/spree/issues/4466
     it "successfully remove and create a product property at the same time" do
       fill_in_property
 
-      expect(page).to have_css('tr.product_property', count: 2)
+      expect(page).to have_css('tr.product_property', count: 1)
 
-      within '#spree_new_product_property' do
+      click_button "Add Product Properties"
+      within '.product_property:first-child' do
         find('[id$=_property_name]').set("New Property")
         find('[id$=_value]').set("New Value")
       end
+      expect(page).to have_css('tr.product_property', count: 2)
 
-      delete_product_property
+      # delete 2nd line (persisted product)
+      accept_alert do
+        within_row(2) { click_icon :trash }
+      end
 
       # Give fadeOut time to complete
       expect(page).to have_css('tr.product_property', count: 1)
@@ -115,7 +120,7 @@ describe "Properties", type: :feature do
 
       expect(page).not_to have_content("Product is not found")
 
-      check_property_row_count(2)
+      check_persisted_property_row_count(1)
     end
 
     def fill_in_property
@@ -133,9 +138,9 @@ describe "Properties", type: :feature do
       expect(page).to have_content 'successfully removed'
     end
 
-    def check_property_row_count(expected_row_count)
+    def check_persisted_property_row_count(expected_row_count)
       click_link "Product Properties"
-      expect(page).to have_css("tbody#product_properties tr", count: expected_row_count)
+      expect(page).to have_css("tbody#product_properties tr:not(#spree_new_product_property)", count: expected_row_count)
     end
   end
 end
--- a/backend/spec/features/admin/products/properties_spec.rb
+++ b/backend/spec/features/admin/products/properties_spec.rb
@@ -108,10 +108,7 @@ describe "Properties", type: :feature do
       end
       expect(page).to have_css('tr.product_property', count: 2)
 
-      # delete 2nd line (persisted product)
-      accept_alert do
-        within_row(2) { click_icon :trash }
-      end
+      delete_product_property
 
       # Give fadeOut time to complete
       expect(page).to have_css('tr.product_property', count: 1)
@@ -120,7 +117,7 @@ describe "Properties", type: :feature do
 
       expect(page).not_to have_content("Product is not found")
 
-      check_persisted_property_row_count(1)
+      check_persisted_property_row_count(0)
     end
 
     def fill_in_property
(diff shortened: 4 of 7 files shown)


backend/spec/features/admin/products/properties_spec.rb
```diff
@@ -128,14 +130,14 @@ def fill_in_property
...
     def delete_product_property
       accept_alert do
-        click_icon :trash
+        find('.delete-resource').click
       end
       expect(page).to have_content 'successfully removed'
     end
 
-    def check_property_row_count(expected_row_count)
+    def check_persistent_property_row_count(expected_row_count)
```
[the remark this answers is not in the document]
hefan: you are right

backend/spec/features/admin/products/properties_spec.rb
```diff
@@ -128,14 +130,14 @@ def fill_in_property
 
     def delete_product_property
       accept_alert do
-        click_icon :trash
+        find('.delete-resource').click
```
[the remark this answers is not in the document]
hefan: before the nonpersistent icons didn't have a remove button. So here are 2 trash icons and we need to distinguish between remove and delete.

hefan:
1. You mean by removing  `id="spree_<%= dom_id(f.object) %>"` from non persistent table rows? That makes sense.

2. i understood that small commits are preferred. So i tried to put it in a sequence somehow. There is code removed to get it to work without the remove option (which could be a way to handle this also) and then put back in. i would also prefer to do it all in one commit.

Maybe i squash and rebase it all into one commit. This would be also my preferred way.

hefan:
ok. added draggable, renamed spec methods and squashed/rebased

kennyadsl:
For draggable, I didn't mean that. I'm not sure what that class actually changes, why do you think it's needed?

The problem I'd like to solve is this one:

Screenshot 2020-05-11 at 10 40 55

As you can see, dragging an element with the handle at the end of the table, when there's a non-persisted element, will make the non-persisted element move up in the table order. If you read the sortable plugin documentation, there's a draggable option that you can use when initializing the plugin to mark which elements are draggable and which aren't. 

For commits, it's not true that you can't use just parts of this PR, for example, the deprecation can go in a separate commit and also the controller check that doesn't build the extra row if not empty. Isn't it?

hefan:
1. Split Commits

I assume commits which would have to be reverted with the next commit are not wanted (the tests needs to be changed depending on sortable nonpersistent items or removable nonpersistent items). Also the system should not be let in a poorer state when only applying some commits. So the only commits

1. Sortable

I understood the problem but was not sure about the solution nor i know much about Sortable. I somehow thought that when adding draggable means that non sortable items are no more draggable also. This was wrong. But non draggable also does not mean they are not changed in position when draaging other items below.

To change the initialization of the Plugin we would need to mess with https://github.com/solidusio/solidus/blob/master/backend/vendor/assets/javascripts/solidus_admin/Sortable.js
or at least this
https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/components/sortable_table.js
right?
There is no draggable option as i understood from here.
"By default, sortable items share draggable properties."

This could be a working solution for the problem
https://stackoverflow.com/questions/13885665/how-to-exclude-an-element-from-being-dragged-in-sortable-list

kennyadsl:
This is the plugin used: https://github.com/SortableJS/Sortable

kennyadsl:
Anyway, commits are needed for documentation. When a developer (even the core team reviewing the code) will look at commits, it's a great source of information. They should also contain the reason why that change happened. I suggest you to read this great article: https://mislav.net/2014/02/hidden-documentation/.

Anyway, it's pretty personal to balance how to break PRs code down into commits. It's very up to the developer submitting the patch. I use to walk in the shoes of other developers that, some months from now, will debug that code and git blame on a line seeing what changed and why. The smallest and indipendent that piece of information is, the easiest it will be for them to understand and quickly fix the code.

hefan:
ok. thank you. i will read and do a commit remix tomorrow!

The sortable/draggable stuff is also an extra commit i assume. Couldn't it be also an extra PR? Because it is also an isolated improvement compared to the current code (for both lists)!

hefan:
Now its regrouped in some isolated commits which each make sense. I hope its ok.

The draggable problem remains (just like before). i would like to tackle that in another Commit. Just altering the 
https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/components/sortable_table.js
with a draggable option seemed to dangerous for me. There are many other Backend Tables which rely on these sortable tables and would need to add that element also.
Any Suggestions?

backend/app/views/spree/admin/product_properties/_product_property_fields.html.erb
```diff
@@ -12,8 +12,11 @@
     <%= f.text_field :value %>
   </td>
   <td class="actions">
-    <% if f.object.persisted? && can?(:destroy, f.object) %>
```
kennyadsl, line 15: Why this `can?` check has been removed?

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -105,26 +105,10 @@ Spree.ready(function(){
 
   $('body').on('click', 'a.spree_remove_fields', function() {
```
kennyadsl, line 106: Could this be called on persisted items as well?

backend/spec/features/admin/products/properties_spec.rb
```diff
@@ -128,14 +130,14 @@ def fill_in_property
 
     def delete_product_property
       accept_alert do
-        click_icon :trash
+        all('.delete-resource').last.click
```
kennyadsl: I think we can use `within_row(1) do` here instead.

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -58,6 +58,14 @@ Spree.ready(function(){
   $('.spree_add_fields').click(function() {
     var target = $(this).data("target");
     var new_table_row = $(target + ' tr:visible:last').clone();
+
+    // id of table row
+    new_table_row.attr("id", "spree_new_"+new_table_row.data("hook"));
```
kennyadsl, line 63: `id` of HTML should be unique on the page, can you explain why this is needed to be this way and why we need to set an id for the row?

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -58,6 +58,14 @@ Spree.ready(function(){
...
+
+    // sort handle
```
kennyadsl, line 65: Maybe reduce the time needed to understand what's happening without having for everyone to also read the code?

```
// Remove sort handle
```

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -58,6 +58,14 @@ Spree.ready(function(){
...
+    new_table_row.find("td").first().empty();
+
+    // form ids
```
kennyadsl, line 68: I can't recall what the following code does to form ids, if we add a comment can we try to explain that, `form ids` is not really helping to understand.

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -66,13 +74,9 @@ Spree.ready(function(){
...
       el.prop("name", el.prop("name").replace(/\d+(?=[^\d]*$)/, new_id))
     })
-    // When cloning a new row, set the href of all icons to be an empty "#"
-    // This is so that clicking on them does not perform the actions for the
-    // duplicated row
-    new_table_row.find("a").each(function () {
-      var el = $(this);
-      el.prop('href', '#');
-    })
+    // actions
```
kennyadsl, line 77: Maybe `Add a remove button to the actions column of the new row` can help more?

backend/app/assets/javascripts/spree/backend/admin.js
```diff
@@ -66,14 +74,10 @@ Spree.ready(function(){
...
-    // This is so that clicking on them does not perform the actions for the
-    // duplicated row
-    new_table_row.find("a").each(function () {
-      var el = $(this);
-      el.prop('href', '#');
-    })
-    $(target).prepend(new_table_row);
+    // actions
+    new_table_row.find("td").last().empty().append('<a class="spree_remove_fields no-text with-tip fa fa-trash icon_link with-tip" data-action="remove" href="#" data-original-title="Remove"><span class="text"></span></a>');
+
```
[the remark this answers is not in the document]
kennyadsl: @hefan did you get why is Hound complaining here?

backend/app/views/spree/admin/product_properties/_product_property_fields.html.erb
hefan: For unifying the both approaches for option values or product properties. You are right, but: Shouldn't these can? constraints all be always implemented on both objects? i would prefer it this way (so can?(:destroy) and can?(:update_positions) in both views..

backend/app/assets/javascripts/spree/backend/admin.js
hefan: No. "spree_remove_fields" is for nonpersistent rows, "delete-resource" for persistent.

backend/spec/features/admin/products/properties_spec.rb
hefan: ok. i would change that in the next refresh.

backend/app/assets/javascripts/spree/backend/admin.js
hefan: Thats right. But at least we need to remove the id from the persistent copied item. Also we need to mark this row somehow as new nonpersistent item, a (for example in properties spec for check_persisted_property_row_count). We could do that via a class instead. Would that be ok?

backend/app/assets/javascripts/spree/backend/admin.js
hefan: ok, then comments like "remove id", "remove handle", "add class"..

backend/app/assets/javascripts/spree/backend/admin.js
hefan: I assume a unique id is given to make sure it does not interfere with given ids of persistent items. Its needed to be created when saving new items. Maybe comment it with "set unique form ids for new items"?

backend/app/assets/javascripts/spree/backend/admin.js
hefan: yep. i would improve that comment also this way!

hefan:
Are the above assumptions/suggestions consensual?

backend/spec/features/admin/products/properties_spec.rb
hefan: ahh this does not work. before delete_product_property can delete any line doesn't matter. But in that specific state of the test it would have to delete the second line! the other tests would have to delete the first line. So i would need to add a param for the line number. i decide to not use the function for my case!

hefan:
i updated the PR accordingly to the suggestions.

backend/app/views/spree/admin/product_properties/_product_property_fields.html.erb
kennyadsl: I think that now that we have the `else`, we need to separate the two if conditions, otherwise, when the object is persisted but the user can't destroy the object, the `.spree_remove_fields` versions of the link will be shown. Isn't it?

hefan:
Yes you are right.
i assume that the user should not have the possibility to remove nor delete if he cannot destroy. right?

another thing came to my mind: 
i guess if someone had previously overriden the `_option_value_fields.html.erb` he would rely on the old method of removing non persistent fields. Should this kind of change then noted in a changelog or something?

hefan:
ok, all done..

backend/app/views/spree/admin/option_types/_option_value_fields.html.erb
```diff
@@ -7,5 +7,12 @@
...
   <td class="name"><%= f.text_field :name %></td>
   <td class="presentation"><%= f.text_field :presentation %></td>
-  <td class="actions"><%= link_to_remove_fields t('spree.actions.remove'), f, no_text: true %></td>
+  <td class="actions">
+    <% if f.object.persisted? %>
+      <%= link_to_delete f.object, no_text: true %>
+    <% else %>
+      <%= link_to_with_icon('trash', 'remove', '#', no_text: true, class: "spree_remove_fields",
+            data: { action: 'remove' }, title: t('spree.actions.remove')) %>
+    <% end %>
```
aldesantis, line 16: How about extracting this into a helper, so that it can be easily reused?

aldesantis:
Thanks @hefan! I just left a small comment about extracting the remove links into a helper.

backend/app/views/spree/admin/option_types/_option_value_fields.html.erb
hefan: I like that, but i am unsure why 
`backend/app/views/spree/admin/option_types/_option_value_fields.html.erb`  
does not use the can? helpers like 
`backend/app/views/spree/admin/product_properties/_product_property_fields.html.erb`

then it should be that functionality to be extracted:
```
    <% if f.object.persisted? %>
      <% if can?(:destroy, f.object) %>
        <%= link_to_delete f.object, no_text: true %>
      <% end %>
    <% else %>
      <%= link_to_with_icon('trash', 'remove', '#', no_text: true, class: "spree_remove_fields",
            data: { action: 'remove' }, title: t('spree.actions.remove')) %>
    <% end %>
```
Also we should use then
`can?(:update_positions, f.object)`
there in `backend/app/views/spree/admin/option_types/_option_value_fields.html.erb` as well.

The PR is about Unifying the Handling of Option Values and Product Properties List.
This i would miss..
Or are there any caveats with using the can helpers there?

aldesantis: Hmm, yeah, I hadn't noticed one uses `can?` while the other doesn't. [Admins can manage all models](https://github.com/solidusio/solidus/blob/master/core/lib/spree/permission_sets/super_user.rb), so I think we can safely add the `can?` statement here. Can you try and see if anything breaks?

As for the `update_positions` action, I think we could also add the `can?` for that one, although I'm not sure that change is in the scope of this PR.

kennyadsl: I'm in favor of extracting, but I'd use a partial instead.

hefan: Thanks @aldesantis @kennyadsl. i am quite unsure. Please let me know what to do, you have a deeper understanding of the Rules and Code Structure of the Backend:

1. Leave it as is for this PR
2. add and test can? for option_value_fields and leave it for this PR
3. add and test can? for option_value_fields and put it all in a partial (in shared folder)
4. Just put the "uncanned" version in a partial.

Aside from that: Putting it in a partial could be an additional commit, right?
i would vote for 1 and would be happy to tackle additional challenges (like remove dragging from nonpersistent elements and put funcionality in partials) in following PRs.

kennyadsl: I'm also ok for merging this as is and iterate with new PRs when we see something that we can improve more.

kennyadsl: @aldesantis is this ok for you as well?

aldesantis: Sounds good to me.

PR #3593, opened 2020-04-22
Prevent OpenURI redirection & private hosts during API image upload

calebhaye:
**Description**
Paperclip’s CVE-2017–0889 Server Side Request Forgery (SSRF)
vulnerability (https://medium.com/in-the-weeds/all-about-paperclips-cve-2017-0889-server-side-request-forgery-ssrf-vulnerability-8cb2b1c96fe8) was likely introduced in our attempts to allow API image uploads via URL.  This patch ensures that redirection does not take place in the context of
uploading an image via the api.  It also does a check to make sure the IP/host aren't private.  Lastly, it uses VCR to stub out the http requests for images.

Ref #3573

**Checklist:**


aldesantis:
@calebhaye correct me if I'm wrong, but we still need to ensure the user cannot pass a private IP as the hostname, is that right?

calebhaye:
> @calebhaye correct me if I'm wrong, but we still need to ensure the user cannot pass a private IP as the hostname, is that right?
> 
> Also, would it be possible to stub the HTTP requests with VCR? That way we're not calling live URLs in our tests, which may easily break/fail randomly.

calebhaye:
> @calebhaye correct me if I'm wrong, but we still need to ensure the user cannot pass a private IP as the hostname, is that right?

Thanks @aldesantis, I added a private IP/host check.

> Also, would it be possible to stub the HTTP requests with VCR? That way we're not calling live URLs in our tests, which may easily break/fail randomly.

I also added VCR, but I left it as a separate commit, for now.  Of course, if we're happy with the implementation I'll squash the commits and we can merge.  But if we decide for some reason that VCR isn't the way we want to go, rebasing out the commit will be easy.

Thank you for your comment, I'm looking forward to hearing your feedback.

api/spec/requests/spree/api/images_controller_spec.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'spec_helper'
+require "private_address_check/tcpsocket_ext"
```
aldesantis, line 4: Just curious why this is needed?

api/app/controllers/spree/api/images_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'private_address_check'
```
aldesantis, line 3: We should either add this to the gemspec _or_ only require it when the attachment is a URI, making the functionality and dependency optional.

aldesantis:
@calebhaye left a couple of comments, thanks!

kennyadsl:
I'm starting to think this shouldn't be handled by Solidus. What about keeping URL upload disabled and allow enabling it with a preference? As far as I got, it's just a matter of adding `Paperclip::UriAdapter` in a Rails initializer, as suggested here: https://github.com/thoughtbot/paperclip#io-adapters. 

It may sound drastic but, what about reverting the code in the controller and just adding some documentation on how to enable this adapter in Paperclip?

Another thing that I prefer not to do is adding code specific for Paperclip which is deprecated and we are trying to slowly replace with ActiveStorage.

What are your thoughts about this?

aldesantis:
@kennyadsl my 2 cents: I like the feature and would like to keep it for the time being, and I think [we could easily replicate it with ActiveStorage](https://blog.eq8.eu/til/upload-remote-file-from-url-with-activestorage-rails.html) when the time comes.

With that said, it would probably be a good idea to make the behavior optional by adding a configuration option, so that the user is fully aware of the security implications and we are not potentially slipping a vulnerability into the codebase without them knowing.

calebhaye:
@kennyadsl @aldesantis 

Honestly, I have mixed feelings about it.  I'm obviously biased because a) I wrote the code, and b) it serves the needs of my business, but beyond my selfishness there may be a valid reason for keeping the feature in place.

I guess one relevant question would be can anyone show me an example of the image api working without the feature in place?  I wasn't able to get it to work, there is no documentation on it, and when I asked the Slack channel no one was able to give me an example there either.  My point is, proving that the alternative works and providing docs for that would be a good place to start.

Beyond that, I think we should consider non-ruby devs who want to use Solidus.  Having this feature work out of the box without having to write a Rails initializer might be a good thing.

That being said, I think it is easy enough to write an initializer to get this sort of functionality, but honestly if I had, I probably would have exposed my app to vulnerabilities.  Having the group-think we've had around these implementations and making a decision in the best interest of the end-user seems like the responsible thing to do.

api/spec/requests/spree/api/images_controller_spec.rb
calebhaye: This is the code that contains the error that will be raised, and that the spec tests for.

[The gem itself](jtdowney/private_address_check) seems pretty simple, and we could just copy or fork it, but it has [undergone some review](https://www.cvedetails.com/vulnerability-list/vendor_id-18563/product_id-47091/opbyp-1/Private-Address-Check-Project-Private-Address-Check.html), so I figured maybe it's worth keeping.  I'm open to a better solution, if anyone has any ideas.

api/app/controllers/spree/api/images_controller.rb
calebhaye: Yeah, I felt like I might be doing this wrong.  Any advice on a proper implementation is welcomed.  For now I'll do your latter suggestion.

api/spec/requests/spree/api/images_controller_spec.rb
calebhaye: I could require it only in the sole spec where it was needed

kennyadsl:
We discussed this via DM on Slack and we considered closing this and reverting the other PR that introduces this security concern (https://github.com/solidusio/solidus/pull/3573). 

The API image upload is working, we tried with:

```sh
curl -v -H "Content-Type: multipart/form-data" -H "Authorization: Bearer your_token" -i -F image[attachment]=@test_image.png http://localhost:3000/api/products/1/images
```

and the image is uploaded correctly. 

Anyway, we agreed that we should improve the documentation around this with the following:

- Add a better description in our API documentation for the endpoints that require `"Content-Type: multipart/form-data"`.
- Explain how to configure your app to allow other types of upload,  like uploading via URLs (for paperclip we can link https://github.com/thoughtbot/paperclip/#io-adapters).

calebhaye:
I personally still think we should keep it, but only enable it with a config flag.  Just so that others don't introduce security vulnerabilities when they try to add the adapter as per the paperclip docs you referenced, and so we don't waste the thought we've put into this thus far.  Showing how to do something safely seems like the best approach, plus we get a handy feature that many developers will use, and it doesn't require writing a custom adapter

@kennyadsl what do you think of doing that, rather than throwing away all the code and a feature that is arguably valuable?

aldesantis:
@calebhaye how about extracting this into an extension? It should be fairly easy to do and would solve a lot of problems: users would have to explicitly install the extension, so they'd know the security implications of the feature, but they'd still be able to benefit from all the work and thought that went into the feature.

calebhaye:
> @calebhaye how about extracting this into an extension? It should be fairly easy to do and would solve a lot of problems: users would have to explicitly install the extension, so they'd know the security implications of the feature, but they'd still be able to benefit from all the work and thought that went into the feature.

Sounds good!  I'll do that.  Closing this now.

PR #3595, opened 2020-04-26
Show only active promotions filter

wildbillcat:
**Description**
This adds active as a ransackable scope to be used in the admin search. 

Motivation is that for stores that have a large number of promotions build up, and it is helpful to just see what is active. This is helpful to make sure old/test promotions aren't mistakenly enabled, as they can be costly if not monitored closely. 

Standard scope:
image

Active filter enabled:
image

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/promotions/index.html.erb
+++ b/backend/app/views/spree/admin/promotions/index.html.erb
@@ -16,31 +16,38 @@
   <div data-hook="admin_promotions_index_search">
     <%= search_form_for [:admin, @search] do |f| %>
       <div class="row">
-        <div class="col-3">
+        <div class="col-4">
           <div class="field">
             <%= label_tag :q_name_cont, Spree::Promotion.human_attribute_name(:name) %>
             <%= f.text_field :name_cont, tabindex: 1 %>
           </div>
         </div>
 
-        <div class="col-3">
+        <div class="col-2">
           <div class="field">
             <%= label_tag :q_codes_value_cont, Spree::Promotion.human_attribute_name(:code) %>
             <%= f.text_field :codes_value_cont, tabindex: 1 %>
           </div>
         </div>
 
-        <div class="col-3">
+        <div class="col-2">
           <div class="field">
             <%= label_tag :q_path_cont, Spree::Promotion.human_attribute_name(:path) %>
             <%= f.text_field :path_cont, tabindex: 1 %>
           </div>
         </div>
 
-        <div class="col-3">
+        <div class="col-2">
           <div class="field">
             <%= label_tag :q_promotion_category_id_eq, Spree::PromotionCategory.model_name.human %><br>
             <%= f.collection_select(:promotion_category_id_eq, @promotion_categories, :id, :name, { include_blank: t('spree.match_choices.all') }, { class: 'custom-select fullwidth' }) %>
+          </div>
+        </div>
+
+        <div class="col-2">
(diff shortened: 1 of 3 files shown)


backend/app/views/spree/admin/promotions/index.html.erb
```diff
@@ -16,33 +16,40 @@
...
+        <div class="col-2">
           <div class="field">
             <%= label_tag :q_promotion_category_id_eq, Spree::PromotionCategory.model_name.human %><br>
             <%= f.collection_select(:promotion_category_id_eq, @promotion_categories, :id, :name, { include_blank: t('spree.match_choices.all') }, { class: 'custom-select fullwidth' }) %>
           </div>
         </div>
+
+        <div class="col-2">
+          <div class="field">
+            <%= label_tag :active, 'Show Only Active' %><br>
```
aldesantis, line 49: Could you add a translation for this?

backend/app/views/spree/admin/promotions/index.html.erb
```diff
@@ -16,33 +16,40 @@
...
+            <%= f.check_box :active, label: false, as: :boolean, inline_label: "#{ t 'spree.active'}", checked_value: true %>
```
aldesantis, line 50: Do we need the `inline_label` option here?

backend/app/controllers/spree/admin/promotions_controller.rb
```diff
@@ -62,6 +63,10 @@ def promotion_code_batch_params
       def promotion_includes
         [:promotion_actions]
       end
+
+      def clear_boolean(query_params, condition)
+        query_params.delete(condition) if query_params[condition] == "0"
```
aldesantis, line 68: Do we actually need this? The [Ransack documentation](https://github.com/activerecord-hackery/ransack#using-scopesclass-methods) states that scopes are only applied for `true` values (unless the scope accepts arguments), so a value of `0` should cause Ransack to ignore the scope.

backend/app/views/spree/admin/promotions/index.html.erb
jarednorman: And even if we do, we probably just want to pass in `t('spree.active')` rather than `"#{t 'spree.active'}"` since they should produce the same value.

backend/app/controllers/spree/admin/promotions_controller.rb
wildbillcat: You're right! Will make the change

backend/app/views/spree/admin/promotions/index.html.erb
wildbillcat: Sure, though a opposed to creating a translation I leveraged spree.active. Here is an image:
image
I can change to a new translation if preferred, but looking at it I think inference is pretty strong.

backend/app/views/spree/admin/promotions/index.html.erb
wildbillcat: It is not needed. I stripped it out.

aldesantis:
@wildbillcat one last thing before we merge: could you squash the commits together?

kennyadsl:
Adding a holding label while waiting for the commits squash, thanks again!

wildbillcat:
Sorry for the delay, things got a bit crazy at work yesterday. Rewrote history to be 1 commit.

PR #3596, opened 2020-04-27
2777 inactive actionless promotions

wildbillcat:
**Description**
This is a proposed solution to https://github.com/solidusio/solidus/issues/2777 where action-less promotions were considered active. As opposed to my original thought of implementing validation, given that there is no state toggle on the promotion that it would make sense to embrace the inferred state. So this implementation just causes a promotion to be considered inactive until it has actions, in addition to the other required material. Otherwise it is considered inactive. 

In the process I found the coupon promotion handler already treated action-less coupons as inactive, so I updated it to continue this behavior. 

**Checklist:**

### DIFF
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -206,6 +206,10 @@ module Spree
     #   @return [Integer] Promotions to show per-page in the admin (default: +15+)
     preference :promotions_per_page, :integer, default: 15
 
+    # @!attribute [rw] disable_actionless_promotion_validation
+    #   @return [Boolean] Promotions should have actions associated before activation (default: +true+)
+    preference :actionless_promotion_inactive, :boolean, default: false
+
     # @!attribute [rw] raise_with_invalid_currency
     #   Whether to raise an exception if trying to set a line item currency
     #   different from the order currency. When false a validation error
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -39,10 +39,21 @@ module Spree
     scope :coupons, -> { joins(:codes).distinct }
     scope :advertised, -> { where(advertise: true) }
     scope :active, -> do
+      if Spree::Config.actionless_promotion_inactive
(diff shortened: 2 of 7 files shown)


core/lib/spree/app_configuration.rb
```diff
@@ -202,6 +202,10 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Integer] Promotions to show per-page in the admin (default: +15+)
     preference :promotions_per_page, :integer, default: 15
 
+    # @!attribute [rw] disable_actionless_promotion_validation
+    #   @return [Boolean] Promotions should have actions associated before activation (default: +true+)
+    preference :actionless_promotion_inactive, :boolean, default: true
```
kennyadsl, line 207: I didn't really review the PR yet but in the meantime, I think this should be changed. The default value should be the one that represents the old behavior (`false` in this case). This way, users that upgrade Solidus will keep having the same behavior and will see a deprecation warning. They will have the chance to take a look at what happens before switching the value.

New Solidus users will have the new version of `spree.rb` initializer instead and they will start their apps with the new default. 

How does it sound?

wildbillcat: No worries and sounds fine. This will probably take me a little while as the change of function appears to break a lot of tests, so I need to reconcile the change across functions.

aldesantis:
@wildbillcat I'd love to get this into 2.11, which will be the last release in the 2.x branch. That way, we can completely remove the old behavior in 3.0. Let me know if you think it's doable!

wildbillcat:
No Idea haha. But holiday weekend, so I'll see if I can sort out the issues this weekend. 
If not I'll note where I could use some help, the change of behavior results in a good portion of the test suite failing out. So I have to reconcile that back into green :+1:

wildbillcat:
@aldesantis Apologies, had some work stuff come up over the weekend that kept me from working on this. I'll pick this up this evening and try to have an update for you Friday. Cheers,

kennyadsl:
Closing in favor of #3749, thanks!

PR #3598, opened 2020-04-28
Remove a N+1 query on shipment model

albanv:
**Description**
In `to_package` method, I've added `product` to includes.
We have discovered that when we call `Spree::Stock::Estimator#calculate_shipping_rates` with a `shipment.to_package`, it currently makes a `product` DB query for each `variant`.
In our test, this subtle change has drastically reduced the number of queries to `product` by a factor equivalent to `variant` number.

Concerning the test to add, since I'm not very sure on how to proceed, I would like some guidance. (I've looked at estimator_spec and shipment_spec but since this issue arise when both are coupled, I'm not sure where is the place to start)

**Checklist:**

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -262,7 +262,7 @@ module Spree
     def to_package
       package = Stock::Package.new(stock_location)
       package.shipment = self
-      inventory_units.includes(:variant).joins(:variant).group_by(&:state).each do |state, state_inventory_units|
+      inventory_units.includes(variant: :product).joins(:variant).group_by(&:state).each do |state, state_inventory_units|
         package.add_multiple state_inventory_units, state.to_sym
       end
       package


aldesantis:
@albanv it looks like CircleCI didn't run here for some reason. Could you push an empty commit and then remove it?

albanv:
@aldesantis just done it, but the CI didn't run as well 😞

albanv:
There is something weird, the CI used here is the one from my org (epicery) and not solidus one

kennyadsl:
@albanv I think you should opt-in to use external orbs in your `epicery/solidus` CircleCI project.

aldesantis:
@albanv looks like the CI is still failing here, any chance you can take a look? 🙏 I'm also happy to create another PR myself with your code if you want, just let me know!

kennyadsl:
@albanv hey there! Just wanted to check if you had time to take a look, I'd love to merge this one, thanks!

albanv:
Sorry for the late reply, I don't work for Epicery anymore and so goes with solidus.
I have rebased my PR to trigger the CI. If it doesn't go through, I summon @stem to take my follow

stem:
@kennyadsl I'm not sure about the CircleCI current failures. Is it the same than before ? Does this PR needs more work before merging ?

kennyadsl:
Nope, the only failing check is with Rails master (which is unreleased so it's not an issue). We are fine now, thanks @stem !

kennyadsl (approved):
Thanks!

aldesantis (approved):

PR #3600, opened 2020-04-30
Adding missing paginator when listing all stock locations

softr8:
Current implementation only list all stock locations limited
by the defaulted per page value, but there's no way to navigate
to multiple pages.

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/stock_locations_controller.rb
+++ b/backend/app/controllers/spree/admin/stock_locations_controller.rb
@@ -13,6 +13,10 @@ module Spree
         flash[:error] = t('spree.stock_locations_need_a_default_country')
         redirect_to(admin_stock_locations_path) && return
       end
+
+      def collection
+        super.page(params[:page]).per(Spree::Config[:admin_products_per_page])
+      end
     end
   end
 end
--- a/backend/app/views/spree/admin/stock_locations/index.html.erb
+++ b/backend/app/views/spree/admin/stock_locations/index.html.erb
@@ -16,6 +16,7 @@
 <% end %>
 
 <% if @stock_locations.any? %>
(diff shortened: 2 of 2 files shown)


backend/app/controllers/spree/admin/stock_locations_controller.rb
```diff
@@ -13,6 +13,12 @@ def set_country
         flash[:error] = t('spree.stock_locations_need_a_default_country')
         redirect_to(admin_stock_locations_path) && return
       end
+
+      def collection
+        super.
```
jarednorman, line 18: Minor style thing: The project seems to use leading periods in multiline method chains over trailing periods. I also think the following two lines are indented too far.

aldesantis:
@softr8 feel free to ping me for a review once the style violations addressed by @jarednorman have been fixed!

backend/app/controllers/spree/admin/stock_locations_controller.rb
softr8: I was trying to follow [this one](https://github.com/solidusio/solidus/blob/master/backend/app/controllers/spree/admin/products_controller.rb#L108) that's why I used this style, what do you suggest?

jarednorman: Oh weird, I checked a couple of random places and saw the other way. I'm not a big stickler, though I tend to agree with the outcome of [the debate they a had about this in the standardrb issue](https://github.com/testdouble/standard/issues/75). Up to you what you want to do. 

And leading or trailing, I think just indenting it 2 spaces is the way to go.

aldesantis:
Thanks @softr8!

PR #3603, opened 2020-05-01
Refactoring Admin::ProductsController to use ResourcesController#update

softr8:
It takes advantage of all validations and error handling that
ResourcesController already does, it looks like the code was
just copied and added a bit that could have been handled with
before action and resources controller callbacks.

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/products_controller.rb
+++ b/backend/app/controllers/spree/admin/products_controller.rb
@@ -9,6 +9,7 @@ module Spree
       update.before :update_before
       helper_method :clone_object_url
       before_action :split_params, only: [:create, :update]
+      before_action :normalize_variant_property_rules, only: [:update]
 
       def show
         redirect_to action: :edit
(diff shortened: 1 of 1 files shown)


backend/app/controllers/spree/admin/products_controller.rb
```diff
@@ -138,6 +116,21 @@ def variant_scope
...
       end
+
+      def render_after_update_error
+        # Stops people submitting blank slugs, causing errors when they try to
+        # update the product again
+        @product.slug = @product.slug_was if @product.slug.blank?
+        render action: 'edit'
+      end
+
+      def check_if_updating_variant_property_rules
```
aldesantis, line 127: This doesn't look like this is a check, it seems more like a normalization? Perhaps a better name for this method would be something like `normalize_variant_property_rules`?

aldesantis:
Looking good, thanks @softr8!

Issue #3405, opened 2019-10-25
RFC: Deprecate And Remove Spree.routes


jarednorman:
As documented in #2113, the route helpers in `Spree.routes` are poorly and inconsistently named. [Hawth also stopped using them in new admin code](https://github.com/solidusio/solidus/issues/2113#issuecomment-317821762), instead preferring to use `pathFor` directly. As mentioned in #2113, this is a pretty low priority issue, but making a decision allows us to close the old issue and move on with a solution.

**Solution:**
Deprecate `Spree.routes` and replace all uses of it with `pathFor`. Remove it in a future major version. I prefer this solution because `Spree.routes` doesn't provide much value and is a small burden to keep up to date.

**Alternative:**
Warn users that we will be making `Spree.routes` consistent with the Rails route naming. Update all the names in the next major version. Given that _this is still a breaking change_, I'm less a fan of it, but if people truly find `Spree.routes` valuable, then I'm not opposed to keeping it.

jarednorman:
Now that we have this issue, someone with permissions can close #2113.

seand7565:
#3605 is merged and should close this issue! 🎉

PR #3605, opened 2020-05-04
Replace Spree.routes with Spree.pathFor

seand7565:

**Description**
This PR replaces all instances of Spree.routes with Spree.pathFor. Spree.routes still
exists, we're just preferring to use Spree.pathFor per #3405.

I've added some lines about deprecation to the JS files where Spree.routes live. We should probably output a warning message to console on use, but I'm not 100% sure what the best way to go about it is, since it's just a variable. How do you console.warn when calling a variable in Javascript? 🤔 If there's a good way to do this, please let me know and I'll add it in.

**Checklist:**

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/routes.js
+++ b/backend/app/assets/javascripts/spree/backend/routes.js
@@ -1,27 +1,46 @@
-Spree.routes.checkouts_api = Spree.pathFor('api/checkouts')
-Spree.routes.classifications_api = Spree.pathFor('api/classifications')
-Spree.routes.option_value_search = Spree.pathFor('api/option_values')
-Spree.routes.option_type_search = Spree.pathFor('api/option_types')
-Spree.routes.orders_api = Spree.pathFor('api/orders')
-Spree.routes.product_search = Spree.pathFor('api/products')
-Spree.routes.admin_product_search = Spree.pathFor('admin/search/products')
-Spree.routes.shipments_api = Spree.pathFor('api/shipments')
-Spree.routes.checkouts_api = Spree.pathFor('api/checkouts')
-Spree.routes.stock_locations_api = Spree.pathFor('api/stock_locations')
-Spree.routes.taxon_products_api = Spree.pathFor('api/taxons/products')
-Spree.routes.taxons_search = Spree.pathFor('api/taxons')
-Spree.routes.user_search = Spree.pathFor('admin/search/users')
-Spree.routes.variants_api = Spree.pathFor('api/variants')
-Spree.routes.users_api = Spree.pathFor('api/users')
+/**
+ * @deprecated Spree.routes will be removed in a future release. Please use Spree.pathFor instead.
+ * See: https://github.com/solidusio/solidus/issues/3405
+ */
 
-Spree.routes.line_items_api = function(order_id) {
-  return Spree.pathFor('api/orders/' + order_id + '/line_items')
+var admin_routes = {
+  classifications_api: Spree.pathFor('api/classifications'),
+  option_value_search: Spree.pathFor('api/option_values'),
+  option_type_search: Spree.pathFor('api/option_types'),
+  orders_api: Spree.pathFor('api/orders'),
+  product_search: Spree.pathFor('api/products'),
+  admin_product_search: Spree.pathFor('admin/search/products'),
+  shipments_api: Spree.pathFor('api/shipments'),
+  checkouts_api: Spree.pathFor('api/checkouts'),
+  stock_locations_api: Spree.pathFor('api/stock_locations'),
+  taxon_products_api: Spree.pathFor('api/taxons/products'),
+  taxons_search: Spree.pathFor('api/taxons'),
+  user_search: Spree.pathFor('admin/search/users'),
+  variants_api: Spree.pathFor('api/variants'),
+  users_api: Spree.pathFor('api/users'),
+
+  line_items_api: function(order_id) {
+    return Spree.pathFor('api/orders/' + order_id + '/line_items')
+  },
+
+  payments_api: function(order_id) {
+    return Spree.pathFor('api/orders/' + order_id + '/payments')
+  },
+
+  stock_items_api: function(stock_location_id) {
+    return Spree.pathFor('api/stock_locations/' + stock_location_id + '/stock_items')
+  }
 }
 
-Spree.routes.payments_api = function(order_id) {
-  return Spree.pathFor('api/orders/' + order_id + '/payments')
+var frontend_routes = {
+  states_search: Spree.pathFor('api/states'),
+  apply_coupon_code: function(order_id) {
+    return Spree.pathFor("api/orders/" + order_id + "/coupon_codes");
+  }
 }
 
-Spree.routes.stock_items_api = function(stock_location_id) {
-  return Spree.pathFor('api/stock_locations/' + stock_location_id + '/stock_items')
+if(typeof Proxy == "function") {
+  Spree.routes = new Proxy(Object.assign(admin_routes, frontend_routes), Spree.routesDeprecationProxy);
+} else {
+  Object.assign(Spree.routes, admin_routes)
 }
--- a/core/app/assets/javascripts/spree.js.erb
+++ b/core/app/assets/javascripts/spree.js.erb
@@ -51,12 +51,41 @@ Spree.ajax = function(url, options) {
   return $.ajax(url, options);
 };
 
-Spree.routes = {
+/**
+ * @deprecated Spree.routes will be removed in a future release. Please use Spree.pathFor instead.
+ * See: https://github.com/solidusio/solidus/issues/3405
+ */
+Spree.routesDeprecationProxy = {
+  get: function(obj, prop) {
+    <% if Rails.env != "production" %>
+      console.log("Spree.routes is deprecated, please use pathFor instead. See: https://github.com/solidusio/solidus/issues/3405");
+    <% end %>
+
+    return obj[prop]
+  },
+
+  set: function(obj, prop, value) {
+    <% if Rails.env != "production" %>
+      console.log("Spree.routes is deprecated, please use pathFor instead. See: https://github.com/solidusio/solidus/issues/3405");
+    <% end %>
+
+    obj[prop] = value
+    return value;
+  }
+};
+
+var frontend_routes = {
   states_search: Spree.pathFor('api/states'),
   apply_coupon_code: function(order_id) {
     return Spree.pathFor("api/orders/" + order_id + "/coupon_codes");
   }
-};
+}
+
+if(typeof Proxy == "function") {
+  Spree.routes = new Proxy(frontend_routes, Spree.routesDeprecationProxy);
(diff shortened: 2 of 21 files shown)


aldesantis:
This looks good! I think moving forward we may even want to use something such as https://github.com/railsware/js-routes which works really well. As for the deprecation, I don't think there's a way to do it, because @kennyadsl's suggestion turns `Spree.routes` into a function, changing the API to call it (it would become `Spree.routes().something`) and breaking backwards compatibility.

kennyadsl:
I think this should do the trick:

```js
const Spree.routes_deprecation_proxy = {
  get: function(obj, prop) {
    console.log("This is deprecated, please use ...");

    return obj[prop]
  },

  set: function(obj, prop, value) {
    console.log("This is deprecated, please use ...");

    obj[prop] = value
    return value;
  }
};

Spree.routes = new Proxy({
  states_search: Spree.pathFor('api/states'),
  apply_coupon_code: function(order_id) {
    return Spree.pathFor("api/orders/" + order_id + "/coupon_codes");
  }
}, Spree.routes_deprecation_proxy);
```

The only bad thing is that, when we add new routes in the backend, we need to create a new `Spree.routes` object, instead of adding properties to it. This is needed to avoid emitting deprecation warnings with our own code. Something like

```js
# backend/app/assets/javascripts/spree/backend/routes.js

Spree.routes = new Proxy({
  states_search: Spree.pathFor('api/states'),
  apply_coupon_code: function(order_id) {
    return Spree.pathFor("api/orders/" + order_id + "/coupon_codes");
  },

  checkouts_api = Spree.pathFor('api/checkouts'),
  classifications_api = Spree.pathFor('api/classifications'),
  option_value_search = Spree.pathFor('api/option_values')
}, Spree.routes_deprecation_proxy);
```

but I don't think it's a big deal. 


Thanks @dinohamzic for pointing out [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy).

seand7565:
@kennyadsl That's perfect! I had no idea about Proxy. I'll plug it in and push again later. Thanks!

kennyadsl:
@seand7565 👍. Since that's a `.js.erb` file, we can even conditionally print the deprecation warning only if it's not in production.

seand7565:
@kennyadsl Added the deprecation warnings, they work great! Thanks for the heads up about Proxy!

core/app/assets/javascripts/spree.js.erb
```diff
@@ -51,12 +51,35 @@ Spree.ajax = function(url, options) {
   return $.ajax(url, options);
 };
 
-Spree.routes = {
+/**
+ * @deprecated Spree.routes will be removed in a future release. Please use Spree.pathFor instead.
+ * See: https://github.com/solidusio/solidus/issues/3405
+ */
+const routes_deprecation_proxy = {
```
kennyadsl, line 58: I think this should be `Spree.routesDeprecationProxy` in order to be visible everywhere and to reflect JS naming conventions.

backend/app/assets/javascripts/spree/backend/routes.js
```diff
@@ -1,27 +1,37 @@
...
-  return Spree.pathFor('api/stock_locations/' + stock_location_id + '/stock_items')
-}
+  payments_api: function(order_id) {
+    return Spree.pathFor('api/orders/' + order_id + '/payments')
+  },
+
+  stock_items_api: function(stock_location_id) {
+    return Spree.pathFor('api/stock_locations/' + stock_location_id + '/stock_items')
+  }
+},routes_deprecation_proxy);
```
kennyadsl, line 37: I think we should use `}, Spree.routes_deprecation_proxy);`, not sure how `routes_deprecation_proxy` can be visible here.

core/app/assets/javascripts/spree.js.erb
seand7565: Yes, you're absolutely right! Thanks!

backend/app/assets/javascripts/spree/backend/routes.js
seand7565: 👍

backend/app/assets/javascripts/spree/backend/routes.js
```diff
@@ -1,27 +1,37 @@
...
-Spree.routes.users_api = Spree.pathFor('api/users')
+/**
+ * @deprecated Spree.routes will be removed in a future release. Please use Spree.pathFor instead.
+ * See: https://github.com/solidusio/solidus/issues/3405
+ */
 
-Spree.routes.line_items_api = function(order_id) {
-  return Spree.pathFor('api/orders/' + order_id + '/line_items')
-}
+Spree.routes = new Proxy({
```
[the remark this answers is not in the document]
seand7565: I'm not sure what it's expecting here? Proxy is a javascript object and shouldn't be undefined.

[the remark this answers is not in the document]
jarednorman: Yeah, it's new-ish API that it probably doesn't know exists... which makes me realize, do we care that it has zero support on IE?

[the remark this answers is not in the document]
seand7565: IE still accounts for 2 to 4% of all web usage according to Wikipedia, but there are only two routes on the frontend - for applying coupon codes and searching states, and I think IE is old enough at this point to justify not supporting. I'm curious to see what others think though.

There's a polyfill for Proxy available, but it looks a bit iffy to me. Babel docs say: 
> Due to the limitations of ES5, Proxies cannot be transpiled or polyfilled.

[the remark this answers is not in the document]
aldesantis: This is an interesting consideration. We don't officially support any browsers, so breaking compatibility with IE wouldn't be a breaking change, at least technically. Also, I wonder if there are any other parts of the JS that use unsupported features.

Would it be possible to only use Proxy when it's available, and revert to a regular object otherwise? The only purpose of Proxy here is to print a deprecation message, so if we can just revert to regular behavior, I think that's fine.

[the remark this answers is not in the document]
seand7565: Check out the latest version, let me know how it looks - on IE11 `typeof Proxy` returns undefined, so that _should_ work... but I don't have the capabilities of testing it on IE11 right now, so I can't be 100% sure.

seand7565:
Hey all - the latest version should fix the hound violation - and also falls back to not using Proxy if the user is on IE11. If anyone has a windows machine and can test that out... I'd really appreciate it. 🙏

core/app/assets/javascripts/spree.js.erb
```diff
@@ -51,13 +51,45 @@ Spree.ajax = function(url, options) {
...
+    return value;
   }
 };
 
+if(typeof Proxy == "function") {
+  Spree.routes = new Proxy({
+    states_search: Spree.pathFor('api/states'),
+    apply_coupon_code: function(order_id) {
+      return Spree.pathFor("api/orders/" + order_id + "/coupon_codes");
+    }
```
aldesantis, line 82: Can we extract this into a separate variable so there's less duplication in the code?

core/app/assets/javascripts/spree.js.erb
```diff
@@ -51,13 +51,45 @@ Spree.ajax = function(url, options) {
...
-    return Spree.pathFor("api/orders/" + order_id + "/coupon_codes");
+/**
+ * @deprecated Spree.routes will be removed in a future release. Please use Spree.pathFor instead.
+ * See: https://github.com/solidusio/solidus/issues/3405
+ */
+Spree.routesDeprecationProxy = {
+  get: function(obj, prop) {
+    <% if Rails.env != "production" %>
+      console.log("Spree.routes is deprecated, please use pathFor instead. See: https://github.com/solidusio/solidus/issues/3405");
+    <% end %>
```
aldesantis, line 62: Would it be better to actually pass the environment to the JS code at runtime? I'm not a huge fan of mixing Ruby and JS like this, I think it could potentially generate a lot of confusion and subtle issues.

backend/app/assets/javascripts/spree/backend/routes.js
```diff
@@ -1,27 +1,66 @@
...
-Spree.routes.user_search = Spree.pathFor('admin/search/users')
-Spree.routes.variants_api = Spree.pathFor('api/variants')
-Spree.routes.users_api = Spree.pathFor('api/users')
+/**
+ * @deprecated Spree.routes will be removed in a future release. Please use Spree.pathFor instead.
+ * See: https://github.com/solidusio/solidus/issues/3405
+ */
+if(typeof Proxy == "function") {
+  Spree.routes = new Proxy({
+    states_search: Spree.pathFor('api/states'),
```
aldesantis, line 7: Can we extract all of this into a variable to remove duplication? In the `else` condition, you should be able to mass-assign these properties via `Object.assign`: https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Global_Objects/Object/assign.

core/app/assets/javascripts/spree.js.erb
seand7565: Absolutely! Check out the latest version.

backend/app/assets/javascripts/spree/backend/routes.js
seand7565: Done! If Proxy is available, we're rebuilding the whole Spree.routes object (to avoid triggering the deprecation). If not, it uses Object.assign to add the admin routes to the already assigned frontend routes. Take a look and let me know if that works for you!

core/app/assets/javascripts/spree.js.erb
seand7565: I'm not quite sure how to go about this - do you have an example? We could assign a JS variable at the top of the file `var production = <%= Rails.env == "production" %>`, is that what you mean?

Also, I noticed in the frontend JS file, there's another deprecation done like this:
```
  if (console && console.warn) {
    console.warn('Spree.url is deprecated, and will be removed from a future Solidus version.');
  }
```
I hadn't noticed it before - would it be better to follow the already existing example set there? 

Thanks!

aldesantis: @seand7565 100%, I didn't realize we had done it before. This will print the warning even in production, but IMO it's okay to be consistent and follow the existing pattern. If we want to only print the warnings in production, let's make sure we change it everywhere, and maybe in a separate PR.

@kennyadsl what do you think?

kennyadsl: Considering it only happens in the admin, I think it's not too bad if we print a deprecation warning in the console. So 👍 on making them more uniform. 

For the other PR to use the `Rais.env` check, I guess the JS file in the frontend is not an `.erb` and I wouldn't add ruby to it since people can be including frontend files in unexpected ways (maybe with webpack?) and this would be a breaking change.

aldesantis:
Thanks @seand7565! <3

jarednorman:
Thank you @seand7565! 🙂

PR #3607, opened 2020-05-05
Fix error message on email preview

coorasse:
Email preview had a wrong error message (typo)


aldesantis:
Rerunning the CI here since the failure is not related.

Issue #3606, opened 2020-05-04
Unable to specify a custom attachment definition via config.taxon_attachment_module


softr8:
With the combination of some extensions and the recent code merged that included support for ActiveStorage, it is no longer possible to specify a custom taxon attachment definition, the changes are in this commit d802e5b6c11ae45a3f2a03322b4f1fc906bbff68

For example, solidus_paypal_braintree gem has an initializer, that adds a [custom helper](https://github.com/solidusio/solidus_paypal_braintree/blob/master/config/initializers/braintree.rb#L2) that calls Solidus Payments Controller, which inherits from Admin Base Controller, which also inherits from BaseController, and it has the Spree::Config.active_storage_enabled? check, it compares the current taxon_attachment_module with ActiveStorages one, this initializer runs before any of my initializers including config/initializers/spree.rb, that's where I usually override default taxon_attachment_module to use my custom module, but it just sets the value but it was already loaded in taxon model

Steps to reproduce

* Create a new solidus app with active storage changes included
* Add solidus_paypal_braintree to gemfile
* Create another taxon/paperclip_attachment with a custom method and add it to spree.rb initializer
* Open rails console, find a taxon then try to call the custom method, it says undefined method since it loaded the default one instead of the custom one

Not sure how to fix it, maybe the most compatible one with current extensions ecosystem is to modify this [check](https://github.com/solidusio/solidus/blob/master/core/lib/spree/app_configuration.rb#L591) to not invoke the module?

What do you think?

kennyadsl:
Yep, I think that `active_storage_enabled?` should not check the classes set to determine if we are using ActiveStorage, we can also be using it without using the modules provided by default in Solidus. 

@filippoliverani can you please take a look next Friday? 🙏

filippoliverani:
> Yep, I think that `active_storage_enabled?` should not check the classes set to determine if we are using ActiveStorage, we can also be using it without using the modules provided by default in Solidus.
> 
> @filippoliverani can you please take a look next Friday? 🙏

I agree, I'll update the check as soon as possible, thank you!

PR #3613, opened 2020-05-08
Always set ActiveStorage::Current.host in base controllers

filippoliverani:
**Description**
Ref #3606 

`ActiveStorage` disk service needs current host to be correctly set.
`ActiveStorage::SetCurrent` concern is not availabile in Rails < 6.0
so we need a custom one. The concern should always be included in
in controllers, there's no need to check if we are currently using
`ActiveStorage` based attachment adapters since Solidus already loads
`ActiveStorage` by default. This way we avoid loading uneccessary modules
during initialization process that could potentially conflict with
custom attachment adapters.

**Checklist:**


kennyadsl:
@softr8 do you think this is ok to fix your issue?

softr8:
> @softr8 do you think this is ok to fix your issue?

Yes!

PR #3614, opened 2020-05-08
Fix specs on Rails master/6.1.0.alpha

filippoliverani:
**Description**

`Search::Base` specs started to fail  because `ActiveRecord`-generated SQL has changed in Rails master/6.1.0.alpha.
Generated SQL is bound to change and should not be used in expectations. Search results expectations now verify actual search output only.
When filtering on ranges `between` Arel operator must be used.
`in` operator implementation isn't converted anymore in a `between` SQL query, it
was a deprecated optimization now removed in rails/rails#39163

`Spree::Api::ImagesController` request specs started to fail because `ActionDispatch::TestProcess::FixtureFile#fixture_file_upload` has been changed in Rails master/6.1.0.alpha and rspec-rails is not compatible anymore with it.
`Rack::Test::UploadedFile` handles our use-case in the same away and doesn't have any compatibility issues.

**Checklist:**

### DIFF
--- a/api/lib/spree/api/testing_support/helpers.rb
+++ b/api/lib/spree/api/testing_support/helpers.rb
@@ -38,7 +38,7 @@ module Spree
         end
 
         def upload_image(filename)
-          fixture_file_upload(image(filename).path, 'image/jpg')
+          Rack::Test::UploadedFile.new(File.open(image(filename).path), 'image/jpg')
         end
       end
     end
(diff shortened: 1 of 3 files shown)


jarednorman:
Nice. I see this is still failing the Rails master build, but I like the approach.

filippoliverani:
> Nice. I see this is still failing the Rails master build, but I like the approach.
Thank you, almost there!

aldesantis:
Thanks @filippoliverani!

Issue #3611, opened 2020-05-07
Ruby 2.7 warnings


softr8:
The following warnings are thrown when running Solidus with Ruby 2.7:

* core/lib/spree/preferences/preferable.rb:162: warning: deprecated Object#=~ is called on TrueClass; it always returns nil
* core/app/models/concerns/spree/user_address_book.rb:8: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
* core/lib/spree/i18n.rb:34: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call

jacquesporveau:
I've opened up a PR for the first warning.

softr8:
Fist and second were fixed here #3737

PR #3616, opened 2020-05-08
Only perform regexes on Strings

jacquesporveau:
**Description**
We were getting warnings from Ruby 2.7

warning: deprecated Object#=~ is called on TrueClass; it always returns
nil

In the case that true was passed in as a preference here we would try to
perform a regex on it and get nil from value =~ /<regex>/.

This meant that the conditional would not be met and this case would not
return false but instead return true.

This is the correct behaviour but it's pretty sketchy. We should not be
relying on what calling regex matching methods on things that aren't
strings return here.

Ref https://github.com/solidusio/solidus/issues/3611

**Checklist:**

### DIFF
--- a/core/lib/spree/preferences/preferable.rb
+++ b/core/lib/spree/preferences/preferable.rb
@@ -158,8 +158,7 @@ module Spree
           value.to_i
         when :boolean
           if !value ||
-             value == 0 ||
-             value =~ /\A(f|false|0)\Z/i ||
+             value.to_s =~ /\A(f|false|0|^)\Z/i ||
              (value.respond_to?(:empty?) && value.empty?)
             false
           else


aldesantis:
I like @jarednorman's idea of a shortcircuit, I think it's slightly better than checking the class. I haven't seen any examples in the wild, but I imagine there are cases where users have passed something that is not a `String` but still implements `#=~` — we should still play nice in that case.

The only problem I see with the shortcircuit approach is that we will still call `=~` on all other objects that are neither `true` nor `false`, so if the user passes an object that is neither a string nor a boolean, we'll still get that warning. Since that's not a supported use case and would be the user's mistake though, I'm not terribly worried about it.

jacquesporveau:
Sounds like there is a learning opportunity here for me. I'd like to try and understand why we don't like to check classes. Could you give me a run down? Maybe that will help me get on the same page as far as looking for another solution.

jarednorman:
There's often a desire in Ruby to embrace [duck-typing](https://en.m.wikipedia.org/wiki/Duck_typing). This usually means preferring to check object responds to the method in question and not worrying about whether the object is of a particular class. Our could should just care that the responds to the methods we want to call ("quacks like a duck").

softr8:
It looks like bad idea, but what about:

```ruby
when :boolean
  ( !value || value.to_s =~ /\A(f|false|0|^)\Z/i ) ? false : true
```

it checks for '0', 'f', 'false' and ''

jarednorman:
It does look bad, but I'm having trouble thinking of a reason that we shouldn't do it.

jacquesporveau:
Sry I've been inactive. I'll give that a go and see if we need to add anymore tests shortly.

jacquesporveau:
I think this is what we are looking for. Couldn't just paste in Edwin's solution because apparently we also supported empty arrays here too so needed to cover that as well.

core/lib/spree/preferences/preferable.rb
```diff
@@ -158,8 +158,7 @@ def convert_preference_value(value, type)
           value.to_i
         when :boolean
           if !value ||
-             value == 0 ||
-             value =~ /\A(f|false|0)\Z/i ||
+             value.to_s =~ /\A(f|false|0|^)\Z/i ||
```
aldesantis, line 161: Does it make sense to use this helper instead?
```ruby
ActiveModel::Type::Boolean.new.cast(value)
```
Looking at the [documentation](https://api.rubyonrails.org/classes/ActiveModel/Type/Boolean.html), it seems to do what we need.

jarednorman: Yes, I prefer that. It's arguably a breaking change though and one that might be weird to use a config option to control. Thoughts?

aldesantis: Hmm, yeah, we may decide to either hold off on this until 3.1 or merge the current version now and migrate to ActiveModel's API in 3.1.

@solidusio/core-team any other thoughts?

jarednorman: I'm fine with merging as is, though I'm curious why you'd want to make the switch the the `ActiveModel` API in 3.1 rather than 3.0. Wouldn't that be a breaking change that would be better off rolling into 3.0?

aldesantis: Sorry, that comment didn't make any sense. 😅 I meant 3.0, not 3.1. The problem is that ideally we don't want any changes at all in 3.0: we should just remove all deprecations introduced in previous versions of Solidus. The idea is that, if you're on 2.11 and not seeing any deprecations, you can safely upgrade to 3.0.

I think one approach could be to print a deprecation warning if the value of a configuration option is one of the values that is now being treated as true, but would be treated as false by ActiveModel. Something like the following:

```ruby
if value.in?(ActiveRecord::Type::Boolean::FALSE_VALUES - ['f', 'false', '0'])
  Spree::Deprecation.warn "#{value.inspect} will be treated as false starting in Solidus 3.0."
end

# ...
```
The check would have to be a bit more complex, since it looks like we currently treat any object that responds to `#empty?` and returns `true` as false, while ActiveRecord only seems to do it for empty strings, and even then, they return `nil` rather than `false`.

jarednorman: Yeah, I agree; that would be ideal. I think we're getting away from the original intent of this PR, though. The goal here was to address a deprecation where `true` is being passed into this class which isn't actually handled by the above deprecation suggestion.

It's more work then I'm sure @jacquesporveau had intended to do here, but maybe we want to leave the existing logic, but wrap it in a conditional deprecation warning that's fired if the existing logic doesn't produce the same value as the `ActiveModel` version (potentially with a config option to use the new `ActiveModel` logic instead.) In v3.0 we remove the old logic and move forward with only the `ActiveModel` version.

@jacquesporveau Apologies for all the back and forth on this one. We're trying to be careful in deciding what belongs in v2.11 since it's our last chance to deprecate things ahead of Solidus v3.0.

jacquesporveau: Sa'll good

jacquesporveau:
Looks like I had to authorize circle to have my code ran by CI. Weird.

Anyways it looks like everything but `ci/circleci: postgres_rails_master_activestorage` is passing which seems to be a common theme looking at the last couple of PRs. Seems like the CI workflow might be having a version problem or something like that.

I am not going to look into why that CI check is failing.

aldesantis:
No worries, it's not related to the changes. Thanks @jacquesporveau!

Issue #3619, opened 2020-05-13
Problems with products sorting with Postgresql


thomasrossetto:
Hi, 
In my case i want to order by master price all of my products in products index page.

I have analyzed the code, and I chose to use an already exists product filers to sort by price. In particular this two:

https://github.com/solidusio/solidus/blob/4df3f052c7f13f85fda93968cea6ea2480d27cc4/core/app/models/spree/product/scopes.rb#L31

https://github.com/solidusio/solidus/blob/4df3f052c7f13f85fda93968cea6ea2480d27cc4/core/app/models/spree/product/scopes.rb#L35

Products controller use a Base Search to search products
https://github.com/solidusio/solidus/blob/4df3f052c7f13f85fda93968cea6ea2480d27cc4/frontend/app/controllers/spree/products_controller.rb#L13

And this one permit to add more scopes via parameters.
https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/core/lib/spree/core/search/base.rb#L59
https://github.com/solidusio/solidus/blob/475d9db5d0291dd4aeddc58ec919988c336729bb/core/lib/spree/core/search/base.rb#L110

So, i  tried to pass the name of scope via parameters and works fine in demo app!

But, in my project i found some troubles with Postresql. In particular with the output query.

I found the same problem documented on code here https://github.com/solidusio/solidus/blob/4df3f052c7f13f85fda93968cea6ea2480d27cc4/core/app/models/spree/product/scopes.rb#L220

I fixed it added .select(...) on those scopes.

You can find my PR here: https://github.com/solidusio/solidus/pull/3620

**Solidus Version**:
2.9.2

**To Reproduce**

1 Launch an instance of Solidus with postgresql under the hood
2 Go to '/products' page
3 Try to use one of those filter `/products?search%5B%5D=descend_by_master_price` or `/products?search%5B%5D=ascend_by_master_price`
4 Everything goes wrong :D 

**Current behavior**
Something goes wrong with a select query

**Expected behavior**
Products will be order correctly

**Screenshots**
[Link](https://ibb.co/jZXxncp)

**Additional context**
I'm running postgresql 11.5

PR #3620, opened 2020-05-13
Added select on product's scopes for price sorting

thomasrossetto:
**Description**
You can find all details [here](https://github.com/solidusio/solidus/issues/3619)

### DIFF
--- a/core/app/models/spree/product/scopes.rb
+++ b/core/app/models/spree/product/scopes.rb
@@ -29,11 +29,13 @@ module Spree
           scope :descend_by_name, -> { order(name: :desc) }
 
           add_search_scope :ascend_by_master_price do
-            joins(master: :default_price).order(Spree::Price.arel_table[:amount].asc)
+            joins(master: :default_price).select('spree_products.* , spree_prices.amount')
+                                         .order(Spree::Price.arel_table[:amount].asc)
           end
 
           add_search_scope :descend_by_master_price do
-            joins(master: :default_price).order(Spree::Price.arel_table[:amount].desc)
+            joins(master: :default_price).select('spree_products.* , spree_prices.amount')
+                                         .order(Spree::Price.arel_table[:amount].desc)
           end
 
           add_search_scope :price_between do |low, high|
--- a/core/spec/lib/search/base_spec.rb
+++ b/core/spec/lib/search/base_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe Spree::Core::Search::Base do
     expect(searcher.retrieve_products.count).to eq(3)
   end
 
-  context "when include_images is included in the initalization params" do
+  context "when include_images is included in the initialization params" do
     let(:params) { { include_images: true, keyword: @product1.name, taxon: @taxon.id } }
     subject { described_class.new(params).retrieve_products }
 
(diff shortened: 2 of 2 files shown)


core/app/models/spree/product/scopes.rb
```diff
@@ -28,12 +28,14 @@ def self.property_conditions(property)
           scope :ascend_by_name, -> { order(name: :asc) }
           scope :descend_by_name, -> { order(name: :desc) }
 
-          add_search_scope :ascend_by_master_price do
-            joins(master: :default_price).order(Spree::Price.arel_table[:amount].asc)
+          Spree::Product.add_search_scope :ascend_by_master_price do
```
kennyadsl, line 31: Is `Spree::Product.` necessary here?

thomasrossetto: Nope 😄  It's just a monkey copy/paste by my decorator 😅

thomasrossetto:
@kennyadsl Added test + commented intentionally the right logic. Now you can see postgresql and (maybe) mysql problems.

core/spec/lib/search/base_spec.rb
```diff
@@ -37,26 +37,28 @@
...
   context "when ascend_by_master_price scope is included in the initalization params" do
-    let(:params) { { search:{
-                     ascend_by_master_price: nil }
-                     }
-                   }
+    let(:params) {
+      { search: {
+                     ascend_by_master_price: nil
+} }
+    }
```
jarednorman, line 44: I'm not sure I'd call this style "fixed". Probably fine to do something more like this, same as below.

```suggestion
    let(:params) {
      { search: { ascend_by_master_price: nil } }
    }
```

It's preferred if you squash these changes into the original commit to avoid cluttering up the history with reformatting commits.

thomasrossetto: Sorry @jarednorman, you are right! I have only launched rubocop to shut up Hound 😅
Now I'll squash all changes in one commit with the fix and you'll can see the CI green with a well formated code. Thanks and sorry for this mess 😄

core/spec/lib/search/base_spec.rb
```diff
@@ -36,6 +36,26 @@
...
+      expect(subject.map { |product| product.price.to_i }).to eq [9, 11, 16]
+    end
+  end
+
+  context "when descend_by_master_price scope is included in the initalization params" do
+    let(:params) { { search: { descend_by_master_price: nil } } }
+
+    subject { described_class.new(params).retrieve_products }
+
+    it "returns products witout errors" do
```
aldesantis, line 54: I think this should say "returns products in descending order" instead, since that's the format used in the other.

core/spec/lib/search/base_spec.rb
```diff
@@ -36,6 +36,26 @@
     end
   end
 
+  context "when ascend_by_master_price scope is included in the initalization params" do
+    let(:params) { { search: { ascend_by_master_price: nil } } }
+
+    subject { described_class.new(params).retrieve_products }
+
+    it "returns product in ascending order" do
```
aldesantis, line 44: Small typo here, should be "products" 🙂

PR #3621, opened 2020-05-13
Allow to configure guest_token cookie options

tvdeyen:
**Description**

The guest_token cookie is currently always only allowed for the current domain (including subdomain).

If you want to use the cookie on a static frontend communicating with your Solidus API you want to share the cookie with all subdomains (ie. www.example.com and api.example.com) in order for the cart session to still work.

With this configuration you can do that.

**Checklist:**

### DIFF
--- a/core/lib/spree/core/controller_helpers/auth.rb
+++ b/core/lib/spree/core/controller_helpers/auth.rb
@@ -42,10 +42,10 @@ module Spree
 
         def set_guest_token
           unless cookies.signed[:guest_token].present?
-            cookies.permanent.signed[:guest_token] = {
+            cookies.permanent.signed[:guest_token] = Spree::Config[:guest_token_cookie_options].merge(
               value: SecureRandom.urlsafe_base64(nil, false),
               httponly: true
-            }
+            )
           end
         end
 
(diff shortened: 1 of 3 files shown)


aldesantis:
Looking good @tvdeyen, thanks! Re-run the tests because the failure seems random and not related.

core/lib/spree/core/controller_helpers/auth.rb
```diff
@@ -42,10 +42,10 @@ def redirect_back_or_default(default)
 
         def set_guest_token
           unless cookies.signed[:guest_token].present?
-            cookies.permanent.signed[:guest_token] = {
+            cookies.permanent.signed[:guest_token] = Spree::Config[:guest_token_cookie_options].merge(
```
kennyadsl, line 45: If we want users to be able to also change `value` and `httponly` options we could merge the config hash into the default one but I'm not sure if that could be actually something needed. What do you think?

core/spec/lib/spree/core/controller_helpers/auth_spec.rb
```diff
@@ -45,6 +45,25 @@ def controller.index
...
+      it 'sends cookie with these options' do
+        stub_spree_preferences(guest_token_cookie_options: {
+          domain: :all,
+          path: '/api'
+        })
+        get :index
+        expect(response.headers["Set-Cookie"]).to match(/domain=\.test\.host; path=\/api/)
+      end
+
+      it 'never overwrites httponly' do
```
kennyadsl, line 59: Just saw this test related to my previous comment. Why do we need this?

tvdeyen: We do not want that devs shoot theirselves in the foot. Setting `httponly` to `false` means the cookie is readable by javascript in the client browser, making the door open for session highjacking attacks.

core/lib/spree/core/controller_helpers/auth.rb
tvdeyen: We must never allow this. See comment below 👇

PR #3624, opened 2020-05-15
Admin UI for shipping methods - stock locations association

cedum:
This introduces the admin UI allowing to associate a shipping method to specific stock locations.
This makes editable also `Spree::ShippingMethod#available_to_all` from admin, being an attribute directly related to stock locations association behavior.

This feature is mostly used when a shipping method must be available to a specific subset of stock locations.
The shipping methods for a package [are queried](https://github.com/solidusio/solidus/blob/61110460baaac24b56160a4037f56320fb7870c7/core/app/models/spree/stock/package.rb#L109-L115) using the following methods:
- [`Spree::ShippingMethod.with_all_shipping_category_ids`](https://github.com/solidusio/solidus/blob/89f25d1519128c8f33b2f944f297f4081335e8c9/core/app/models/spree/shipping_method.rb#L40-L55)
- [`Spree::ShippingMethod.available_in_stock_location`](https://github.com/solidusio/solidus/blob/89f25d1519128c8f33b2f944f297f4081335e8c9/core/app/models/spree/shipping_method.rb#L57-L77)

The second query method filters the shipping methods having `available_to_all` enabled OR having the stock location associated.
By disabling `available_to_all` it becomes possible to make specific shipping methods available for specific stock locations.

Screenshot 2020-05-15 at 11 45 13

**Checklist:**

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -319,6 +319,8 @@ en:
         name: Name
         service_level: Service Level
         tracking_url: Tracking URL
+        stock_locations: Stock Locations
+        available_to_all: Available to all stock locations
       spree/shipping_rate:
         amount: Amount
         label: Label
@@ -884,6 +886,9 @@ en:
         expired: Expired
         inactive: Inactive
         not_started: Not started
+      shipping_methods:
+        form:
+          stock_locations_placeholder: 'Choose stock locations'
       stock_locations:
         form:
           address: Address
@@ -1412,6 +1417,9 @@ en:
           If no value is specified, the promotion will be immediately available.
         promo_code_will_be_disabled: Selecting this option, promo codes will be disabled for this promotion
           because all its rules / actions will be applied automatically to all orders.
+      spree/shipping_method:
+        available_to_all: Uncheck to select specific stock locations this
+          shipping method will be available in.
       spree/stock_location:
         active: 'This determines whether stock from this location can be used when
           building packages.<br/> Default: Checked'
--- a/backend/app/assets/javascripts/spree/backend.js
+++ b/backend/app/assets/javascripts/spree/backend.js
@@ -61,3 +61,4 @@
 //= require spree/backend/user_picker
 //= require spree/backend/variant_autocomplete
 //= require spree/backend/zone
+//= require spree/backend/shipping_methods/stock_locations_picker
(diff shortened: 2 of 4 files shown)


aldesantis:
@cedum thanks, this is awesome! I was wondering if we can easily clear and disable the Stock Locations field when Available to all is checked? It would be a nice UX addition.

cedum:
I've introduced a few interactions between the checkbox and the select:
- when checking the checkbox, the selected stock locations get cleared;
- when selecting a stock location, the checkbox gets unchecked automatically;
- when clearing all stock locations, the checkbox gets checked automatically.

What do you think?

Here's a demo:

aldesantis:
@cedum maybe it's not a good idea to auto-select the checkbox when clearing all stock locations? I'm thinking about a scenario where a store owner wants to temporarily disable a shipping method by removing it from all stock locations. They may not notice that the "Available to all stock locations" checkbox got auto-selected, and they would have the opposite result of what they were expecting. 😄

cedum:
> @cedum maybe it's not a good idea to auto-select the checkbox when clearing all stock locations? I'm thinking about a scenario where a store owner wants to temporarily disable a shipping method by removing it from all stock locations. They may not notice that the "Available to all stock locations" checkbox got auto-selected, and they would have the opposite result of what they were expecting. 😄

I think the scope of this feature is to support the main two cases UI-wise:
- adding a stock locations whitelist;
- removing the stock locations whitelist.
Actually, I'm thinking that the "Available to all" checkbox may be completely removed 😄 We can switch that attribute via a hidden input, it's mostly an implementation detail that probably should not be exposed outside.

Disabling a shipping method by removing the stock locations whitelist and disabling the checkbox IMHO is an edge-case, and I think it may be a surprise for a user. If there's a need for such a feature, we may think about introducing a dedicated and clear action for that.

What do you think? Not being a heavy Solidus admin user I may have misinterpreted something 😄

aldesantis:
I think it makes sense — if the main use case is the whitelist, we can remove the user-facing "Available to all stock locations" input and just add a hint under the "Stock locations" input.

core/config/locales/en.yml
```diff
@@ -1411,6 +1416,9 @@ en:
           If no value is specified, the promotion will be immediately available.
         promo_code_will_be_disabled: Selecting this option, promo codes will be disabled for this promotion
           because all its rules / actions will be applied automatically to all orders.
+      spree/shipping_method:
+        available_to_all: Uncheck to select specific stock locations this
+          shipping method will be available.
```
aldesantis, line 1421: I think it should say "will be available in".

aldesantis:
This is amazing, thanks @cedum!

PR #3625, opened 2020-05-15
keep admin routes in backend

ccarruitero:
**Description**
I found that admin api_key routes are inside api routes file. I didn't find that these routes were used in api.

So I think we should move that routes into admin routes file.

This doesn't change any functionality, so I think doesn't need any extra spec.

**Checklist:**

### DIFF
--- a/api/config/routes.rb
+++ b/api/config/routes.rb
@@ -1,17 +1,6 @@
 # frozen_string_literal: true
 
 Spree::Core::Engine.routes.draw do
-  namespace :admin do
-    resources :users do
-      resource :api_key, controller: 'users/api_key', only: [:create, :destroy]
-
-      member do
-        put :generate_api_key # Deprecated
-        put :clear_api_key # Deprecated
-      end
-    end
-  end
-
   namespace :api, defaults: { format: 'json' } do
     resources :promotions, only: [:show]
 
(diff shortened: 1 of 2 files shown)


jarednorman:
The API itself doesn't "use" it's own routes, since it's just an API. The backend makes use of many API routes, but that doesn't mean those routes belong in the API.

aldesantis:
Agreed, those routes should stay in solidus_api, whose job is to expose API routes for everyone else's convenience.

kennyadsl:
Thanks a lot for your contribution but I agree with @jarednorman and @aldesantis. 

I'd actually move [this controller](https://github.com/solidusio/solidus/blob/afd7f5b3bc1f3701012b7932725941aa772e04f8/backend/app/controllers/spree/admin/users/api_key_controller.rb) in API, instead. It should be defined in the same gem where routes and [migrations](https://github.com/solidusio/solidus/tree/afd7f5b3bc1f3701012b7932725941aa772e04f8/api/db/migrate) are defined.

ccarruitero:
Thanks all for your comments.

I think the description in the PR wasn't very clear.

My concern about these routes, is that are useless for any other than admin.

If you look into the [controller](https://github.com/solidusio/solidus/blob/master/backend/app/controllers/spree/admin/users/api_key_controller.rb), these actions ended with a redirection into admin.

I think move the controller into `solidus_api` could be a better approach; always that modify the controller to return the `api_key` instead the redirection.

I'll work on that change.

PR #3626, opened 2020-05-15
Improving stock items management

softr8:
The following items were addressed
 
* Filtering was working by single stock location, made it multiple
* Available stock items by variant was being filtered as well
  when specified a stock location
* Improved available stock locations list in new stock item select box

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/stock_items_controller.rb
+++ b/backend/app/controllers/spree/admin/stock_items_controller.rb
@@ -43,21 +43,24 @@ module Spree
         @stock_locations = Spree::StockLocation.accessible_by(current_ability)
         @stock_item_stock_locations = params[:stock_location_id].present? ? @stock_locations.where(id: params[:stock_location_id]) : @stock_locations
         @variant_display_attributes = self.class.variant_display_attributes
-        @variants = Spree::Config.variant_search_class.new(params[:variant_search_term], scope: variant_scope).results
-        @variants = @variants.includes(:images, stock_items: :stock_location, product: :variant_images)
-        @variants = @variants.includes(option_values: :option_type)
-        @variants = @variants.order(id: :desc).page(params[:page]).per(params[:per_page] || Spree::Config[:orders_per_page])
+        @variants = Spree::Config.variant_search_class.new(params[:variant_search_term], scope: variant_scope).results.
+            order(id: :desc).page(params[:page]).per(params[:per_page] || Spree::Config[:orders_per_page])
       end
 
       def variant_scope
-        scope = Spree::Variant.accessible_by(current_ability)
-        if @product
-          scope = scope.where(
-            product: @product,
-            is_master: !@product.has_variants?
+        scope = Spree::Variant
+          .accessible_by(current_ability)
+          .distinct.order(:sku)
+          .includes(
+            :images,
+            stock_items: :stock_location,
+            product: :variant_images,
+            option_values: :option_type
           )
-        end
-        scope = scope.order(:sku)
+
+        scope = scope.where(product: @product, is_master: !@product.has_variants?) if @product
+        scope = scope.by_stock_location(params[:stock_location_id]) if params[:stock_location_id].present?
+
         scope
       end
 
--- a/backend/app/controllers/spree/admin/stock_items_controller.rb
+++ b/backend/app/controllers/spree/admin/stock_items_controller.rb
@@ -40,8 +40,14 @@ module Spree
       end
 
       def load_stock_management_data
-        @stock_locations = Spree::StockLocation.accessible_by(current_ability)
-        @stock_item_stock_locations = params[:stock_location_id].present? ? @stock_locations.where(id: params[:stock_location_id]) : @stock_locations
+        @stock_locations = Spree::StockLocation.accessible_by(current_ability, :read)
+        @stock_item_stock_locations = Spree::DeprecatedInstanceVariableProxy.new(
+          view_context,
+          :@stock_locations,
+          :stock_item_stock_locations,
+          Spree::Deprecation,
+          "Please, do not use @stock_item_stock_locations anymore in the views, use @stock_locations",
+        )
         @variant_display_attributes = self.class.variant_display_attributes
         @variants = Spree::Config.variant_search_class.new(params[:variant_search_term], scope: variant_scope).results.
             order(id: :desc).page(params[:page]).per(params[:per_page] || Spree::Config[:orders_per_page])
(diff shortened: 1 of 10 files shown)


aldesantis:
@softr8 any chance you could split the improvements into multiple commits? It would make the Git history much more useful! 🙏

softr8:
@kennyadsl what about now?

kennyadsl (commented):
Left a possible improvement in terms of backward compatibility. I'd really love to merge this! Thanks, Edwin.

kennyadsl:
Hey @softr8, do you have any update here? Does my comment make sense for you as well?

softr8:
@kennyadsl done! let me know if this needs more work 👍

kennyadsl (commented):
Thanks @softr8, works for me and I think we can also backport the deprecation in Solidus 2.11.x so we can remove the deprecated instance variable in Solidus 3.0. Can you please rebase against solidusio/solidus? There's an extra commit in the PR that should not be present.

softr8:
Done

backend/app/controllers/spree/admin/stock_items_controller.rb
```diff
@@ -43,22 +43,29 @@ def load_stock_management_data
...
       end
 
       def collection
         []
       end
+
+      def add_stock_location_filtering(scope)
+        return scope if params[:stock_location_id].blank?
+
+        scope.joins(:stock_locations).where(spree_stock_locations: { id: params[:stock_location_id] })
```
jarednorman, line 73: Perhaps we could move this to a scope? I usually prefer keeping query logic out of the controller, so we could add a scope like:

```ruby
scope :by_stock_location, -> (stock_location_id) {
  if stock_location_id
    joins(:stock_locations).where(spree_stock_locations: { id: params[:stock_location_id] })
  else
    all
  end
}
```

Then in the controller we'd just add `.by_stock_location(params[:stock_location_id].presence)` to the scope query.

Thoughts?

jarednorman (changes requested):
Good change, one minor suggestion.

backend/app/controllers/spree/admin/stock_items_controller.rb
softr8: Done! Thanks @jarednorman

jarednorman (approved):

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -111,14 +111,20 @@ def configure_application
       end
     end
 
-    def install_default_plugins
+    def plugin_install_preparation
```
softr8, line 114: Wait, not sure what happened here, I did not touch this file

softr8: Fixed

kennyadsl:
@softr8 Thanks for the update Edwin. There's a spec failing that seems related to the PR, can you please take a look?

softr8:
done! @kennyadsl

spaghetticode (approved):
Looks good to me, thanks @softr8 👍

backend/app/views/spree/admin/stock_items/_stock_management.html.erb
```diff
@@ -71,18 +71,24 @@
...
+                  data-variant-id="<%= variant.id %>"
+                  data-stock-item="<%= item.to_json %>"
+                  data-stock-location-name="<%= item.stock_location.name %>"
+                  data-track-inventory="<%= variant.should_track_inventory? %>"
+                  data-can-edit="<%= can?(:admin, Spree::StockItem) %>">
+                <%# This is rendered in JS %>
+              </tr>
             <% end %>
-            <% locations_without_items = @stock_item_stock_locations - variant.stock_items.flat_map(&:stock_location) %>
+            <% locations_without_items = stock_locations - variant.stock_items.flat_map(&:stock_location) %>
```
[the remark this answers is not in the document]
kennyadsl: @softr8 did you have the chance to take a look at this commit? If I wasn't clear let me know and I'll try to rephrase it in a better way. 🙂

jarednorman:
There's some conflicts that need resolving here.

waiting-for-dev:
Hey @softr8, I rebased from master, and it looks like some tests are failing. I know it's been a while, but do you have a clue which could be the problem?

core/app/models/spree/variant/scopes.rb
```diff
@@ -10,6 +10,14 @@ def self.prepended(base)
             order(Arel.sql("COALESCE((SELECT COUNT(*) FROM  #{Spree::LineItem.quoted_table_name} GROUP BY #{Spree::LineItem.quoted_table_name}.variant_id HAVING #{Spree::LineItem.quoted_table_name}.variant_id = #{Spree::Variant.quoted_table_name}.id), 0) DESC"))
           }
 
+          scope :by_stock_location, ->(stock_location_id) {
+            if stock_location_id
+              joins(:stock_locations).where(spree_stock_locations: { id: stock_location_id })
+            else
+              all
```
BenMorganIO, line 17: I believe with scopes, if `nil` is returned it essentially acts as an `all` anyways.

kennyadsl: @BenMorganIO do you see any issue using `all` here?

waiting-for-dev (approved):

core/app/models/spree/variant/scopes.rb
BenMorganIO: No, I just thought to point out the `else` section might be able to be deleted entirely.

kennyadsl: Ah, cause it will fallback to `nil` and will act as `.all`. Up to @elia (currently assigned to this PR) if he wants to change it.

elia: @BenMorganIO @kennyadsl good call, I updated the code to avoid the check altogether and added a spec for `Variant.by_stock_location`.

backend/app/controllers/spree/admin/stock_items_controller.rb
```diff
@@ -43,21 +43,24 @@ def load_stock_management_data
...
 
       def variant_scope
-        scope = Spree::Variant.accessible_by(current_ability)
-        if @product
-          scope = scope.where(
-            product: @product,
-            is_master: !@product.has_variants?
+        scope = Spree::Variant
+          .accessible_by(current_ability)
+          .distinct.order(:sku)
```
kennyadsl, line 59: @softr8 👋 We are having some issues with this `distinct` call. Getting `for SELECT DISTINCT, ORDER BY expressions must appear in select list (PG::InvalidColumnReference)`. Do you remember the reason why this was added?

tvdeyen: @kennyadsl @softr8 getting the same issue here while searching by variant name. 

```
ActiveRecord::StatementInvalid (PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1: ...e_option_values"."name" ILIKE '%blish%') ORDER BY "spree_var...
```

will open an issue as this is a regression we introduced in 4.2

PR #3633, opened 2020-05-21
refactor api_key routes into solidus_api

ccarruitero:
**Description**
This continue from #3625

The current api_key routes are useless outside admin, because return a redirection into admin and you will need to set the admin csrf token to use.

I'm adding new routes and controller for api_key in solidus_api and updated the admin users edit view to use solidus_api and backbone for update user's api_key.

I found an edge case here (in admin), if you clear the api_key for the current user you wouldn't generate a new one since you need an api_key to send the request to solidus_api.

I was thinking in how to solve this and become with two options:

- Don't allow user to clear there own api_key.

- Add a similar route into admin to generate an api_key.

I prefer the first option to avoid repetitive code.

What do you think?

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/users/api_key_controller.rb
+++ b/api/app/controllers/spree/api/users/api_key_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module Spree
+  module Api
+    module Users
+      class ApiKeyController < Spree::Api::BaseController
+        def create
+          authorize! :manage, model_class
+          user.generate_spree_api_key!
+          respond_with(user, default_template: 'spree/api/users/api_key')
+        end
+
+        def destroy
+          authorize! :manage, model_class
+          user.clear_spree_api_key!
+          head 204
+        end
+
+        private
+
+        def user
+          @user ||= model_class.find(params[:user_id])
+        end
+
+        def model_class
+          Spree.user_class
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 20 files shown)


kennyadsl:
Thanks for your contribution @ccarruitero but my opinion is that we don't need this since the Rest API should not be responsible for creating authorization tokens that should be delegated to some other tools. I think what we need is just moving the admin controllers that take care of api keys generation into the `api` subcomponent. If you mount the API, it just adds the ability for an admin to generate API keys.

One further improvement could be to create an adapter to handle which kind of API authentication someone wants to use, by default we could ship the current one and if needed it can be changed in JWT, for example.

ccarruitero:
@kennyadsl I think it's ok to not expose the token generation into api, but I'm not agree to just move admin controller into api.

I think could be better keep each component routes and controllers inside each them. In my opinion, the code looks more orderly this way and is more easy find for something related to one component in just one place.

So, I prefer move the admin routes, for generate api token, into backend instead.

> One further improvement could be to create an adapter to handle which kind of API authentication someone wants to use, by default we could ship the current one and if needed it can be changed in JWT, for example.

This sounds very interesting.

waiting-for-dev:
Changing how API authentication works is on our roadmap, and probably it'll be done in parallel to the current implementation (which would be deprecated). For those reasons, I'll close it for now, but let me thank you again for your collaboration, @ccarruitero.

PR #3641, opened 2020-05-26
Stop calling perform! as Spree::Refund after_create callback 

kennyadsl:
**Description**

This PR allows creating Refunds without `perform!` being automatically called by a Rails callback. 

This is a combination of two leftover PRs that we want to bring into Solidus `v2.11`:

- #1415 Remove callbacks from refund model by @swively
- #3181 Introduce deprecation warning when using `Spree::Refund#perform!` callback by @aitbw

They were both incomplete: the first one was doing the job but without taking care of adding a deprecation warning, the second one was only adding the deprecation warning without actually change the behavior. If this is an accepted approach we can close both of them.

I merged them together and added some missing pieces. In 3.0 we'll introduce a deprecation warning for that attribute so that people can remove that since it will be useless.

**Some notes for the CHANGELOG entry:**

```
From Solidus v3.0 onwards, #perform! will need to be explicitly called when creating new refunds. Please, change your code from:

Spree::Refund.create(your: attributes)

to:

Spree::Refund.create(your: attributes, perform_after_creation: false).perform!

The `perform_after_creation` attribute will be deprecated in Solidus 3.x 
```

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/refunds_controller.rb
+++ b/backend/app/controllers/spree/admin/refunds_controller.rb
@@ -10,6 +10,21 @@ module Spree
 
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
+      def create
+        @refund.attributes = refund_params.merge(perform_after_creation: false)
+        if @refund.save && @refund.perform!
+          flash[:success] = flash_message_for(@refund, :successfully_created)
+          respond_with(@refund) do |format|
+            format.html { redirect_to location_after_save }
+          end
+        else
+          flash.now[:error] = @refund.errors.full_messages.join(", ")
+          respond_with(@refund) do |format|
+            format.html { render action: 'new' }
+          end
+        end
+      end
+
       private
 
       def location_after_save
@@ -23,6 +38,10 @@ module Spree
 
       def refund_reasons
         @refund_reasons ||= Spree::RefundReason.active.all
+      end
+
+      def refund_params
+        params.require(:refund).permit!
       end
 
       def build_resource
--- a/backend/app/controllers/spree/admin/refunds_controller.rb
+++ b/backend/app/controllers/spree/admin/refunds_controller.rb
@@ -11,7 +11,7 @@ module Spree
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
       def create
-        @refund.attributes = refund_params.merge(perform_after_creation: false)
+        @refund.attributes = refund_params.merge(perform_after_create: false)
         if @refund.save && @refund.perform!
           flash[:success] = flash_message_for(@refund, :successfully_created)
           respond_with(@refund) do |format|
--- a/backend/spec/controllers/spree/admin/refunds_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/refunds_controller_spec.rb
@@ -6,22 +6,48 @@ describe Spree::Admin::RefundsController do
   stub_authorization!
 
   describe "POST create" do
-    context "a Spree::Core::GatewayError is raised" do
-      let(:payment) { create(:payment) }
+    let(:refund_reason) { create(:refund_reason) }
+    let(:refund_amount) { 100.0 }
 
-      subject do
-        post :create,
-          params: {
-          refund: { amount: "50.0", refund_reason_id: "1" },
+    let(:payment) { create(:payment, amount: payment_amount) }
+    let(:payment_amount) { refund_amount * 2 }
+
+    subject do
+      post(
+        :create,
+        params: {
+          refund: {
+            amount: refund_amount,
+            refund_reason_id: refund_reason.id,
+            transaction_id: nil
+          },
           order_id: payment.order_id,
           payment_id: payment.id
         }
+      )
+    end
+
+    context "and no Spree::Core::GatewayError is raised" do
+      it "creates a refund record" do
+        expect{ subject }.to change(Spree::Refund, :count).by(1)
       end
 
-      before(:each) do
-        def controller.create
-          raise Spree::Core::GatewayError.new('An error has occurred')
-        end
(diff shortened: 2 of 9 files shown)


backend/app/controllers/spree/admin/refunds_controller.rb
```diff
@@ -25,6 +40,10 @@ def refund_reasons
         @refund_reasons ||= Spree::RefundReason.active.all
       end
 
+      def refund_params
+        params.require(:refund).permit!
```
aldesantis, line 44: Is there a reason we are not explicitly whitelisting parameters here?

backend/spec/controllers/spree/admin/refunds_controller_spec.rb
```diff
@@ -6,22 +6,46 @@
...
+          to receive(:perform!).
+          and_raise(Spree::Core::GatewayError.new('An error has occurred'))
       end
 
-      before(:each) do
-        def controller.create
-          raise Spree::Core::GatewayError.new('An error has occurred')
-        end
+      it "does not create a refund record" do
+        expect{ subject }.to_not change { Spree::Refund.count }
```
aldesantis, line 48: If you use `change(Spree::Refund, :count)`, RSpec will produce a more meaningful error message when the expectation fails.

core/app/models/spree/refund.rb
```diff
@@ -37,22 +38,57 @@ def description
...
+
+      Spree::Deprecation.warn <<-WARN.strip_heredoc, caller
+        From Solidus v3.0 onwards, #perform! will need to be explicitly called when creating new
+        refunds. Please, change your code from:
+
+          Spree::Refund.create(your: attributes)
+
+        to:
+
+          Spree::Refund.create(your: attributes, perform_after_creation: false).perform!
```
aldesantis, line 79: I have a small concern about this approach. In 3.0, `#perform!` will never be called after creation. That means the `perform_after_creation` attribute will become useless. Is our plan for 3.0 to introduce a deprecation warning when `perform_after_creation` is passed to `create`? We can't simply remove the attribute without a deprecation, because passing an unknown attribute to `create` will raise an error IIRC (and also because we want to make sure people clean up their code).

core/app/models/spree/refund.rb
```diff
@@ -10,13 +10,14 @@ class Refund < Spree::Base
...
     validates :reason, presence: true
-    validates :transaction_id, presence: true, on: :update # can't require this on create because the before_create needs to run first
     validates :amount, presence: true, numericality: { greater_than: 0 }
 
     validate :amount_is_less_than_or_equal_to_allowed_amount, on: :create
 
+    attr_accessor :perform_after_creation
+    after_create :set_perform_after_creation_default
     after_create :perform!
-    after_create :create_log_entry
```
aldesantis, line 19: Was this moved somewhere else or simply removed? I couldn't find it in the PR so I'm assuming the latter?

backend/app/controllers/spree/admin/refunds_controller.rb
kennyadsl: I don't think we are whitelisting params in the admin for any controller. It's how we do in [ResourceController](https://github.com/solidusio/solidus/blob/ac325ef25194a61bf97a7ad975cd50853ac921ce/backend/app/controllers/spree/admin/resource_controller.rb#L281), which is inherited by all the other controllers.

core/app/models/spree/refund.rb
kennyadsl: That's the plan! In 3.0 we'll introduce a deprecation warning for that attribute so that people can remove that since it will be useless.

core/app/models/spree/refund.rb
kennyadsl: This is the code of that method:

```
    def create_log_entry
      log_entries.create!(details: @response.to_yaml)
    end
```

but `@response` is only populated during [`perform!` execution](https://github.com/solidusio/solidus/blob/ac325ef25194a61bf97a7ad975cd50853ac921ce/core/app/models/spree/refund.rb#L49). This means that if `perform!` is not called that LogEntry was empty. That stuff has been moved into the `perform!` method and `create_log_entry` is a private method so I thought it was ok just getting rid of that. Do you think we need to keep and deprecate it instead?

backend/spec/controllers/spree/admin/refunds_controller_spec.rb
kennyadsl: Done, didn't know this. Thanks!

core/app/models/spree/refund.rb
aldesantis: I think it's good, I didn't realize you had moved it to `#perform!`.

core/app/models/spree/refund.rb
aldesantis: Got it. I think that makes sense. Trying to figure out if there's a way to avoid introducing a new deprecation in 3.0 but I think this is the best way to do it.

core/app/models/spree/reimbursement_type/reimbursement_helpers.rb
```diff
@@ -37,7 +37,7 @@ def create_refund(reimbursement, payment, amount, simulate)
         reason: Spree::RefundReason.return_processing_reason
       })
 
-      simulate ? refund.readonly! : refund.save!
+      simulate ? refund.readonly! : refund.save! && refund.perform!
```
jarednorman, line 41: I'd lean towards making this a normal conditional now that the falsy path contains multiple statements. `refund.save!` doesn't ever return false so this code is somewhat misleading looking at a glance.

kennyadsl: You are right, it's a good chance to improve code here.

PR #3645, opened 2020-05-29
Add ability to run validations in order updater

kennyadsl:
**Description**

This PR is a continuation of the old #2208. 

It's only adding to the original work a preference that allows people to knowingly switch to the new behavior with:

```ruby
Spree::Config.run_order_validations_on_order_updater = true
```

It also adds a deprecation warning booting the application if the value used for the preference is false. 

**Checklist:**

### DIFF
--- a/core/app/models/spree/order_updater.rb
+++ b/core/app/models/spree/order_updater.rb
@@ -186,7 +186,7 @@ module Spree
     end
 
     def persist_totals
-      order.save!
+      order.save!(validate: Spree::Config[:run_order_validations_on_order_updater])
     end
 
     def log_state_change(name)
--- a/core/app/models/spree/order_updater.rb
+++ b/core/app/models/spree/order_updater.rb
@@ -186,7 +186,7 @@ module Spree
     end
 
     def persist_totals
-      order.save!(validate: false)
+      order.save!
     end
 
     def log_state_change(name)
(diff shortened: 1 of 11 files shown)


core/lib/spree/core/engine.rb
```diff
@@ -64,6 +64,15 @@ class Engine < ::Rails::Engine
             caller
           )
         end
+
+        if Spree::Config.run_order_validations_on_order_updater == false
```
spaghetticode, line 68: Any reason for not allowing also other falsy values (ie. `nil`) here?

core/app/models/spree/order_updater.rb
```diff
@@ -186,7 +186,7 @@ def update_item_total
     end
 
     def persist_totals
-      order.save!(validate: false)
+      order.save!(validate: Spree::Config[:run_order_validations_on_order_updater])
```
spaghetticode, line 189: Out of curiosity, is there a reason for choosing the `[]` syntax here?

core/lib/spree/core/engine.rb
kennyadsl: I don't think it can be `nil` or any other thing since it's initialized with `true` as default. Of course, unless you initialize this with a non-boolean value. Anyway, the only accepted value that should not raise a deprecation error is `true` so I'm going to change this, thanks!

core/app/models/spree/order_updater.rb
kennyadsl: You are right, the `.` version is the one used in the codebase. Thanks!

Issue #3615, opened 2020-05-08
Add ability for store admin to set BCC for order confirmations


seand7565:
**Is your feature request related to a problem? Please describe.**
Magento and others allow the ability to send a BCC of order confirmations to store owners, but this isn't possible in vanilla Solidus.

**Describe the solution you'd like**
Have a field where you can set a BCC email address for any order confirmation emails sent out. It probably makes sense to put this field on the store edit page. 

![image](https://user-images.githubusercontent.com/5720486/81413552-3d758080-910b-11ea-846a-80fb39bc82a0.png)

**Describe alternatives you've considered**
Maybe also have a boolean for whether or not to send a BCC? Though the presence of an email address in the BCC field indicates that as well, so it's probably unnecessary. 

**Additional context**
I've seen this feature requested 3 times - twice from stores I've worked on, and once on the #support channel, so it's probably a feature that will get some use.

seand7565:
#3646 is merged and closes this issue! 🎉

PR #3646, opened 2020-05-30
Add BCC email to spree_store

seand7565:

**Description**
Adds a BCC email field to the spree store edit page.
This indicates a BCC email that you want all order confirmation
emails to be BCCed to. As indicated in #3615 , this is a feature
that I think a lot of eCommerce stores are interested in, at
the very least to server as a notification of a new order.

image

Also adds the ability to add/update this field through the API

**Checklist:**

### DIFF
--- a/api/spec/requests/spree/api/stores_controller_spec.rb
+++ b/api/spec/requests/spree/api/stores_controller_spec.rb
@@ -33,6 +33,7 @@ module Spree
             "meta_keywords" => nil,
             "seo_title" => nil,
             "mail_from_address" => "spree@example.org",
+            "bcc_email" => nil,
             "default_currency" => nil,
             "code" => store.code,
             "default" => true,
@@ -46,6 +47,7 @@ module Spree
             "meta_keywords" => nil,
             "seo_title" => nil,
             "mail_from_address" => "spree@example.org",
+            "bcc_email" => nil,
             "default_currency" => nil,
             "code" => non_default_store.code,
             "default" => false,
@@ -64,6 +66,7 @@ module Spree
           "meta_keywords" => nil,
           "seo_title" => nil,
           "mail_from_address" => "spree@example.org",
+          "bcc_email" => nil,
           "default_currency" => nil,
           "code" => store.code,
           "default" => true,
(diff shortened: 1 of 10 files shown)


api/spec/requests/spree/api/stores_controller_spec.rb
```diff
@@ -76,7 +79,7 @@ module Spree
           code: "spree123",
           name: "Hack0rz",
           url: "spree123.example.com",
-          mail_from_address: "me@example.com"
+          mail_from_address: "me@example.com",
```
spaghetticode, line 82: the comma without any subsequent line makes me think that maybe you meant to add the `bcc_email` field here?

seand7565: Good catch! I ultimately decided it was unnecessary to add the field there (since bcc_email isn't required to create a store, which is what it's testing), but forgot to remove the comma. 

As for the BCC email not being capitalized, I uploaded an outdated screenshot - serves me right for not checking first. It's properly capitalized in this version. I'll update my PR with the latest screenshot.

core/app/mailers/spree/order_mailer.rb
```diff
@@ -7,7 +7,7 @@ def confirm_email(order, resend = false)
       @store = @order.store
       subject = build_subject(t('.subject'), resend)
 
-      mail(to: @order.email, from: from_address(@store), subject: subject)
+      mail(to: @order.email, bcc: bcc_address(@store), from: from_address(@store), subject: subject)
```
peterberkenbosch, line 10: When there is no `bcc_address` present, would this cause any issues?

peterberkenbosch:
awesome! Thanks Sean. I think we had this present in the past as well, not sure why it got removed then, but this is super useful!

core/app/mailers/spree/order_mailer.rb
seand7565: ActionMailer only uses the bcc if it's not `blank?` so nil or empty strings don't break anything
image

peterberkenbosch: perfect!

aldesantis:
Thanks @seand7565!

afdev82:
I already had a PR for that feature, but it was refused because it was not enough general/flexible: https://github.com/solidusio/solidus/pull/2905
I'm glad that now it's finally landed in the core and I can again use the upstream version of Solidus! :+1:

PR #3655, opened 2020-06-05
Allow Importer::Order to accept array of line items and stock_location_id

ccarruitero:
**Description**
This PR is for improve `Importer::Order` to allow `stock_location_id` when create shipments and allow use an array for `line_item_attributes`.

`Importer::Order` is used internally for `solidus_api` to create orders, and like is [documented](https://solidus.docs.stoplight.io/api-reference/orders/create-order) should allow `stock_location_id` in `shipments_attributes` and an array for `line_item_attributes`.

**Checklist:**

### DIFF
--- a/core/lib/spree/core/importer/order.rb
+++ b/core/lib/spree/core/importer/order.rb
@@ -57,8 +57,8 @@ module Spree
             shipment = Shipment.new
             shipment.tracking       = target[:tracking]
             shipment.stock_location = Spree::StockLocation.find_by(id: target[:stock_location_id]) ||
-            Spree::StockLocation.find_by(admin_name: target[:stock_location]) ||
-            Spree::StockLocation.find_by!(name: target[:stock_location])
+              Spree::StockLocation.find_by(admin_name: target[:stock_location]) ||
+              Spree::StockLocation.find_by!(name: target[:stock_location])
 
             inventory_units = target[:inventory_units] || []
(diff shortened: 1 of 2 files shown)


core/lib/spree/core/importer/order.rb
```diff
@@ -98,19 +100,31 @@ def self.create_shipments_from_params(shipments_hash, order)
...
-            line_item = order.contents.add(Spree::Variant.find(line_item[:variant_id]), line_item[:quantity])
-            # Raise any errors with saving to prevent import succeeding with line items failing silently.
-            if extra_params.present?
-              line_item.update!(extra_params)
-            else
-              line_item.save!
+
+          case line_items_hash
+          when Hash
+            line_items_hash.each_key do |key|
```
aldesantis, line 106: What's the reason you're not using `each_value` instead? Or better yet, converting `line_items_hash` so that it's always an array, and _then_ iterating on the converted array to call `create_line_item`?

ccarruitero: Thanks for notice that @aldesantis 

That comes from the original implementation, but you are right could be better use `each_value` instead.

aldesantis: Sorry, I didn't realize it came from the original implementation. I think it makes sense not to change it in this PR, if that's the case!

PR #3656, opened 2020-06-07
Remove blacklist and whitelist usage

peterberkenbosch:
For obvious reasons. Open to other naming suggestions of course.

There are other parts in the codebase that use blacklist, these are from 3rd party software that need to be updated
when we can.

- [x] replace blacklist with deny
- [x] replace whitelist with allow
- [ ] track 3rd party software used and try to change it there as well.
- [ ] add deprecations for these methods
- [x] update documention and guides

**Description**

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/deny_urls.rb
+++ b/core/lib/spree/testing_support/deny_urls.rb
@@ -2,8 +2,8 @@
 
 module Spree
   module TestingSupport
-    module BlacklistUrls
-      def setup_url_blacklist(browser)
+    module DenyUrls
+      def setup_url_denied_for(browser)
         if browser.respond_to?(:url_blacklist)
           browser.url_blacklist = ['http://fonts.googleapis.com']
         end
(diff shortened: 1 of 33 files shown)


core/lib/spree/testing_support/deny_urls.rb
```diff
@@ -2,8 +2,8 @@
 
 module Spree
   module TestingSupport
-    module BlacklistUrls
-      def setup_url_blacklist(browser)
+    module DenyUrls
```
jarednorman, line 5: I'd probably call this `BlockUrls` or `BlockedUrls` or something. That's probably even a more explanatory name than the original.

core/lib/spree/testing_support/deny_urls.rb
```diff
@@ -14,10 +14,10 @@ def setup_url_blacklist(browser)
 
 RSpec.configure do |config|
   config.before(:each, type: :feature) do
-    setup_url_blacklist(page.driver.browser)
+    setup_url_denied_for(page.driver.browser)
```
jarednorman, line 17: Similarly this could be `setup_blocked_urls_for`.

waiting-for-dev:
Hey @peterberkenbosch, having this work finished would be great. Is there anything else to do here? If not, could you please rebase from master and set it ready for review? :pray:

Issue #3559, opened 2020-03-18
Investigate removal of associate_user in OrdersController#edit


kennyadsl:
Apparently, we can stop [calling this method](https://github.com/nebulab/solidus/blob/dd62a3b83298bd77c3673fd13e86d30eab1207f8/frontend/app/controllers/spree/orders_controller.rb#L44), see [here](https://github.com/solidusio/solidus/pull/3494#discussion_r386117331) for a more in-depth discussion.

bitberryru:
My mail server continues to flag github mails as spam :( I have not seen the discussions, I would have answered earlier.

I’ll try to explain again, in the old versions building order and association user with it was in different places, now when you build/create an order (by calling current_order) it is immediately associated with the current user, now there is no need to separately associate the user with the order. This call (https://github.com/solidusio/solidus/blob/master/frontend/app/controllers/spree/orders_controller.rb#L44) has survived from the old days and has no sense now.

FYI: There is only one case when it is required to associate an anonymous order with a user - after user authentication. This task is solved by solidus_auth_devise in warden hook. And this is the responsibility of the authentication system.

carlosep:
Thanks, I understand now.
Based on this I opened **https://github.com/solidusio/solidus/pull/3667** removing the `associate_user` line from the `OrdersController#edit`. It also seems to make sense to remove it from the `CheckoutController` but I'm worried about the mentioned order of execution and would like some comments from @elia on the PR about that and maybe about some way we could test it before I could remove that line as well if it is the case.

elia:
@carlosep I agree with @bitberryru about being time to get rid of that. 

About testing it I would suggest going through a deprecation cycle (which I will make it into 2.11) using a configuration flag to get the old behavior. In this scenario any stori having an OrderController decorator that defines a method like this:

```rb
    def associate_user
      super

      # Update the order to reset the adjustments on the order once a user has
      # been associated with it.
      @order.recalculate
    end
```

should see a deprecation and be able to move its customization elsewhere before upgrading. If done properly that should also cover the CheckoutController before-action and we should be able to cleanup things in a short cycle.

carlosep:
This seems like a good solution. Would that mean just a `warn` in the `associate_user` method or is there a standard way we should go through?

elia:
[...]

PR #3667, opened 2020-06-14
Removes deprecated usage of associate_user on OrdersController

carlosep:
**Description**
This Pull Request solves the issue https://github.com/solidusio/solidus/issues/3559

Based on [this discussion](https://github.com/solidusio/solidus/pull/3494#discussion_r386117331) this `associate_user` call is deprecated due to changes during the creation or building of an order via `current_order` where before the creation and association of an user were made in different places now it is done immediately. This seems to classify as refactoring, and for that reason I didn't add tests

There is also a call of `associate_user` in `CheckoutController` which is called after `load_order` that calls `current_order` but I didn't remove it here due to the discussion regarding the order of execution and its impacts on some stores. I would like @elia's comment regarding this, and maybe we could do it on this PR as well.

**Checklist:**

### DIFF
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -41,7 +41,6 @@ module Spree
     def edit
       @order = current_order(build_order_if_necessary: true)
       authorize! :read, @order, cookies.signed[:guest_token]
-      associate_user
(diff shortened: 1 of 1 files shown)


carlosep:
This was closed due to the opening of another one:
https://github.com/solidusio/solidus/pull/3677

PR #3671, opened 2020-06-18
Fix bug related to free shipping not being applied correctly

### LINKED ISSUE #3664: FreeShipping promotion action doesn't work with multiple shipments
While working on the [new Solidus guides](https://edgeguides.solidus.io), I noticed what I believe to be a bug in the `FreeShipping` promotion action.

The `#perform` method of the action goes through each shipment on the order and discounts the ones that haven't been already discounted. Each shipment is mapped to either `false` or `true`, depending on whether the shipment was discounted or not, so that the return value of the method can be computed properly (`#perform` should return `true` when any actions have been taken, and `false` otherwise).

Unfortunately, [this `return` statement](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/promotion/actions/free_shipping.rb#L12), which is meant to map a shipment to `false` when it's already been discounted, will actually *jump out of the entire `#perform` method* and make it return `false`. This means that, if an order has multiple shipments, the action will stop executing on the first shipment that returns `false`, rather than going through all of them. It will also make `#perform` return `false` even when some shipments on the order have been discounted.

This is caused by a common misunderstanding of how `return` works in blocks: when you call `return` from inside a block, it will actually exit from the *method*, not just the block. In order to exit from the block, we should call `next false` instead. As an example, take the following snippet:

```ruby
def test
  value = proc { return false }.call
  puts "The block returned #{value}"
end

# The output is never printed, before method
# execution stops at the `.call` statement.
test # => false
```

And then this one:

```ruby
def test
  value = proc { next false }.call
  puts "The block returned #{value}"
end

# Prints "The block returned false".
test # => nil
```

Also see [Ruby: A small summary of what return, break and next means for blocks](https://makandracards.com/makandra/46939-ruby-a-small-summary-of-what-return-break-and-next-means-for-blocks).

**Solidus Version**: Looks like this has been there for as long as the action existed.

**To Reproduce**

I haven't tried to reproduce it yet and I suspect it may be hard to do so, since it's not common for an order's shipments to change mid-course. This may also be why the bug has gone unnoticed for so long.

**Current behavior**
The `#perform` method stops at the first non-discountable shipment.

**Expected behavior**
The `#perform` method should cycle through all the shipments, not stopping at the first non-discountable shipment.

jacquesporveau:
**Description**

Fixes #3664 

I would leave a more detailed PR description but the author of #3664 did a really good job in describing the problem at hand and this PR is just a solution for that.

I kind of took a bit of liberty in improving the spec quality. Let me know if you don't like that style or something like that and I can rework it if need be.

**Checklist:**

### DIFF
--- a/core/app/models/spree/promotion/actions/free_shipping.rb
+++ b/core/app/models/spree/promotion/actions/free_shipping.rb
@@ -8,8 +8,9 @@ module Spree
           order = payload[:order]
           promotion_code = payload[:promotion_code]
 
-          results = order.shipments.map do |shipment|
-            return false if promotion_credit_exists?(shipment)
+          created_adjustments = order.shipments.map do |shipment|
+            next if promotion_credit_exists?(shipment)
+
             shipment.adjustments.create!(
               order: shipment.order,
               amount: compute_amount(shipment),
@@ -17,11 +18,11 @@ module Spree
               promotion_code: promotion_code,
               label: label
             )
-            true
           end
-          # Did we actually end up applying any adjustments?
+
+          # Did we actually end up creating any adjustments?
           # If so, then this action should be classed as 'successful'
-          results.any? { |result| result == true }
+          created_adjustments.any?
         end
 
         def label
--- a/core/spec/models/spree/promotion/actions/free_shipping_spec.rb
+++ b/core/spec/models/spree/promotion/actions/free_shipping_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe Spree::Promotion::Actions::FreeShipping, type: :model do
   let(:promotion_code) { promotion.codes.first! }
 
   # From promotion spec:
-  context "#perform" do
+  describe "#perform" do
     before do
       order.shipments << create(:shipment)
       promotion.promotion_actions << action
--- a/core/spec/models/spree/promotion/actions/free_shipping_spec.rb
+++ b/core/spec/models/spree/promotion/actions/free_shipping_spec.rb
@@ -31,12 +31,55 @@ RSpec.describe Spree::Promotion::Actions::FreeShipping, type: :model do
       end
     end
 
-    context "when order already has one from this promotion" do
-      it "should not create a discount" do
-        expect(action.perform(payload)).to be true
-        expect(action.perform(payload)).to be false
-        expect(promotion.usage_count).to eq(1)
-        expect(order.shipment_adjustments.count).to eq(2)
+    context "when a number of the orders shipments already have free shipping" do
+      let(:shipments) { order.shipments }
+
+      context "when all the shipments have free shipping" do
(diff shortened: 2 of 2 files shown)


core/spec/models/spree/promotion/actions/free_shipping_spec.rb
```diff
@@ -31,12 +31,55 @@
...
-    context "when order already has one from this promotion" do
-      it "should not create a discount" do
-        expect(action.perform(payload)).to be true
-        expect(action.perform(payload)).to be false
-        expect(promotion.usage_count).to eq(1)
-        expect(order.shipment_adjustments.count).to eq(2)
+    context "when a number of the orders shipments already have free shipping" do
+      let(:shipments) { order.shipments }
+
+      context "all the shipments have free shipping" do
```
spaghetticode, line 37: I think this context description would sound better by adding `when` at the beginning

core/spec/models/spree/promotion/actions/free_shipping_spec.rb
```diff
@@ -31,12 +31,55 @@
...
+            order.shipment_adjustments.count
+          }
+        end
+
+        it "should return false" do
+          expect(action.perform(payload)).to eq(false)
+        end
+      end
+
+      context "some of the shipments have free shipping" do
```
spaghetticode, line 61: same as above, I'd add `when` at the beginning

core/app/models/spree/promotion/actions/free_shipping.rb
```diff
@@ -8,20 +8,21 @@ def perform(payload = {})
...
+          created_adjustments = order.shipments.map do |shipment|
+            if !promotion_credit_exists?(shipment)
+              shipment.adjustments.create!(
+                order: shipment.order,
+                amount: compute_amount(shipment),
+                source: self,
+                promotion_code: promotion_code,
+                label: label
+              )
+            end
```
jarednorman, line 20: Up to you which you think is better, but I probably would have switched this to use `next` to avoid unnecessary code churn.

```suggestion
          created_adjustments = order.shipments.map do |shipment|
            next if promotion_credit_exists?(shipment)
            
            shipment.adjustments.create!(
              order: shipment.order,
              amount: compute_amount(shipment),
              source: self,
              promotion_code: promotion_code,
              label: label
            )
```

jacquesporveau: 👍

jacquesporveau:
Thanks gang. [I've addressed the feedback.](https://github.com/solidusio/solidus/compare/08c47bffa3e11610b6573cc8ac48ebed168ed1a1..caf8957a584b3e1d612432af15ff42b11c33e4b2)

Issue #3666, opened 2020-06-14
Add functionality and links to stock movement page


seand7565:
**Is your feature request related to a problem? Please describe.**
As someone who has managed stock in the past, being able to see a detailed list of stock movements for any given product is essential. If I need to figure out why the website says we have 1,000 of a product but there are only 3 on the shelf, having a list of stock movements is key to figuring that out. This is something that Solidus already does, but it's not very useful or visible in its current state.

**Describe the solution you'd like**
There's this handy page that I don't think a lot of admin even know about: Stock locations stock movements. It's located at `/admin/stock_locations/#{id}/stock_movements`. It displays the stock movements for the location in a nice list that contains the order/refund the movement happened for, and how much stock was moved.

![image](https://user-images.githubusercontent.com/5720486/84593900-6c6ec880-ae14-11ea-9fe1-82dfddde41ee.png)

This is already pretty useful, but there are a few things we can do to improve it:

- [x] Add an originator for stock movements initiated by admin users. Admin stock movements currently have no originator, but it's useful to see _who_ made the movement, so the originator in this case should just be the admin user who initiated it.
- [x] Add search functionality for variant.sku and originator so we can see stock movements for any given item or created by any admin user.
- [x] Add a link to this page from each stock_item on the product stock page, scoped to the variant ("See Recent Movements" or something, that page has limited space, so maybe just an icon would be better for the link)

With all this in place, I'd be able to go back through stock movements for a product to figure out why there's a discrepancy between what Solidus says we have on hand and what we actually have on hand - very useful as a stock manager!

jacquesporveau:
I've started to look into this. 

I created a stock movement via the admin interface as described in the docs:
https://guides.solidus.io/developers/inventory/stock-movements.html

After that I noticed that the record created associated the user that made the stock movement correctly.

```
irb(main):005:0> Spree::StockMovement.last
  Spree::StockMovement Load (0.6ms)  SELECT "spree_stock_movements".* FROM "spree_stock_movements" ORDER BY "spree_stock_movements"."id" DESC LIMIT ?  [["LIMIT", 1]]
=> #<Spree::StockMovement id: 38, stock_item_id: 6, quantity: -1, action: nil, created_at: "2020-06-20 18:20:06", updated_at: "2020-06-20 18:20:06", originator_type: "Spree::User", originator_id: 1>
```

[...]

PR #3673, opened 2020-06-20
Display originator email in stock movement admin

jacquesporveau:
**Description**

This gets #3666 started.

From the issue:

> Add an originator for stock movements initiated by admin users. Admin stock movements currently have no originator, but it's useful to see who made the movement, so the originator in this case should just be the admin user who initiated it.

This PR is meant to remedy that trouble.

One point that might be worth discussing is whether or not we should link to the actual user. I think it would be kind of difficult to figure out whether the originator is a user unless we do a class check and I think those are not in favour.

Since we can't really figure out if the originator is a user then providing a link in the admin to the originator becomes difficult. It may be worth going after the most common use case which is that if the originator responds to email then it's probably a `Spree::User` and we could do another conditional and provide a link if it's a user.

Let me know what you think!

Screenshot:

Screen Shot 2020-06-20 at 12 06 25 PM

**Checklist:**

### DIFF
--- a/backend/app/helpers/spree/admin/stock_movements_helper.rb
+++ b/backend/app/helpers/spree/admin/stock_movements_helper.rb
@@ -12,6 +12,8 @@ module Spree
           else
             originator.number
           end
+        elsif originator.respond_to?(:email)
+          originator.email
         else
           ""
         end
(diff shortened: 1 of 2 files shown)


kennyadsl:
Thank you @jacquesporveau, there's an unrelated failing spec that we need to take a look at but I think this is safe to merge.

PR #3674, opened 2020-06-21
Return API Users with distinct result when using Ransack

### LINKED ISSUE #3183: User picker proposes the same users at promotion edit page

hefan:
**Description**
Return API Users ransack searches with a distinct result.

Fixes #3183
-->

**Checklist:**

### DIFF
--- a/api/spec/requests/spree/api/users_controller_spec.rb
+++ b/api/spec/requests/spree/api/users_controller_spec.rb
@@ -144,6 +144,21 @@ module Spree
         expect(response.status).to eq(422)
         expect(json_response).to eq({ "error" => "Cannot delete record." })
       end
+
+      it "returns distinct search results" do
+        distinct_user = create(:user, email: 'distinct_test@solidus.com')
+        distinct_user.addresses << create(:address)
+        distinct_user.addresses << create(:address)
+        get spree.api_users_path, params: {
+          q: {
+            m: 'or',
+            email_start: 'distinct_test',
+            firstname_or_lastname_start: 'distinct_test'
+          }
+        }
+        expect(json_response['count']).to eq(1)
+        expect(json_response['users'].first['email']).to eq distinct_user.email
+      end
     end
   end
 end
--- a/api/app/controllers/spree/api/users_controller.rb
+++ b/api/app/controllers/spree/api/users_controller.rb
@@ -1,6 +1,19 @@
 # frozen_string_literal: true
 
 class Spree::Api::UsersController < Spree::Api::ResourceController
+  def index
+    user_scope = model_class.accessible_by(current_ability, :read)
+    if params[:ids]
+      ids = params[:ids].split(",").flatten
+      @users = user_scope.where(id: ids)
+    else
+      @users = user_scope.ransack(params[:q]).result
+    end
+
+    @users = paginate(@users.distinct)
+    respond_with(@users)
+  end
+
   private
 
   attr_reader :user


hefan:
I was not quite sure to write one.
The Request Spec for spree/api/users_controller has no scenario for that.
Should i write one which tests that each user is only once returned when using a ransack search like the one in #3183?
Can you point me to where there is a similar test like this so i could follow the best practices from the start? And where the test should go to (i guess only to requests/spree/api/users_controller_spec.rb)?

jarednorman:
Yeah, I think a request spec in the file `requests/spree/api/users_controller_spec.rb` would be appropriate. Even just a single spec that fails without this change and passes with it, showing that the endpoint returns non-duplicated users would be great.

spaghetticode:
I was wondering if having distinct results may be something we may want to move upstream, ie. avoid that any subclass of Spree::Api::ResourceController returns duplicates, not only this one.

kennyadsl:
@spaghetticode this was the approach used in [another PR](https://github.com/solidusio/solidus/pull/3195) for the same issue. Please see this comment https://github.com/solidusio/solidus/pull/3195#issuecomment-500366500 and let me know any thoughts on that.

hefan:
I think about the considerations from @kennyadsl in #3195:

1. i do not understand why not and how does `distinct: true` as ransack option does the same like the activerecord distinct. But i assume the activerecord distinct is not a problem from what i read there.
2. The special handling in the backend users controller is gone in master.

However, about the use in the resource controller i am not sure. i imagine scenarios where one could want none distinct results when searching for product/variants or orders/line items etc.

My first idea was to just remove 
line 31: `firstname_or_lastname_start: term` here
https://github.com/solidusio/solidus/blob/16ed773a6554782e33476e2ffc41b70320ddb46e/backend/app/assets/javascripts/spree/backend/user_picker.js#L31

because i never would search for an user by an adress when i want to give promotions on a user basis. Its the only place where the user picker is used.

Please tell me about the preferred way to go here..

hefan:
OK. i've added a spec test for the original approach of this PR with a distinct API Users ransack search result for users

api/spec/requests/spree/api/users_controller_spec.rb
```diff
@@ -144,6 +144,21 @@ module Spree
         expect(response.status).to eq(422)
         expect(json_response).to eq({ "error" => "Cannot delete record." })
       end
+
+      it "returns distinct search results" do
+        uuser = create(:user, email: 'distincttest@solidus.com')
```
aldesantis, line 149: Small typo here, should be `user`?

hefan: It was meant to be an abbreviation of unique_user. i have changed the spec it to be more precise.

aldesantis: Ah, got it! Yeah, I think the full version is more easily understandable.

spaghetticode:
@hefan can you please rebase with master? This build is failing due to an annoying flaky spec that was fixed recently. After the rebase your build should become green 💚 ... thanks 🙏

hefan:
done

kennyadsl:
Thank you Stefan!

Issue #3652, opened 2020-06-03
Encrypted Preferences


elia:
**Is your feature request related to a problem? Please describe.**
Credentials for payment methods are stored as preferences. That means that if anyone gains access to the database or to a database backup can potentially perform operations on the payment provider.
Although this is not a very common scenario it would still be better to have those credentials encrypted in the database.

**Describe the solution you'd like**
I'd like to see the encryption capabilities that Rails already uses for cookies and secrets applied to preferences (when requested). E.g.:

```rb
module SolidusMyPaymentProvider
  class Gateway < SolidusSupport.payment_method_parent_class
    preference :client_id, :encrypted_string
    preference :client_secret, :encrypted_string
    # ...
  end
end
```

**Describe alternatives you've considered**
I initially tried to write getter/setters for those preference, and leverage `Rails::Secrets.encrypt` and `Rails::Secrets.decrypt`, but those are meant to be used in their context, and eventually I thought this is a problem common to almost all payment methods.

aldesantis:
Encrypted credentials would be a good addition, but to be honest I think we should just move away from using database-stored preferences for credentials in the first place, and encourage people to use environment variables instead. Having to change a gateway's credentials is a rare occurrence and you can ask a developer to do it, especially considering how quick it is to change an environment variable.

elia:
@aldesantis I agree, whenever possible we should avoid storing credentials in the database. There are, however, some cases in which that's is not possible or would require an unjustified burden on the user. 

The use case I have in mind, and from which this issue emerged, is the case in which the payment method is created through an inline onboarding process. Of course I'm working on an extension that does just that 😄.

kennyadsl:
I think it's a good idea to add this to Solidus but I'm a bit concerned about adding things that can be critical for security. Is there a way to use/delegate some Rails internal mechanism, so that we don't add extra security concerns in Solidus?

cedum:
I think [`MessageEncryptor`](https://api.rubyonrails.org/v6.0.3.1/classes/ActiveSupport/MessageEncryptor.html) (used also by `Rails::Secrets` in this case) is fairly easy to use as a simple encryption mechanism.
The biggest issue here IMO in most the encryption key management: safe generation, storage, and rotation, even by reusing the Rails secrets key.

elia:
> The biggest issue here IMO is the encryption key management: safe generation, storage, and rotation, even by reusing the Rails secrets key.

[...]

PR #3676, opened 2020-06-22
Add new type encrypted_string for preferences

stefano-sarioli:
**Description**

This pull request wants to resolve the issue Ref #3652 adding a new type encrypted_string as a possible type when declaring preferences.

The new type encrypted_string would encrypt the value assigned to the preference so that when saved to a Database the value would be safe if a malicious actor gets access to the DB or a dump.

The value to be encrypted need to be set/get using the preferences setter and getter method, the encryption would be skipped if the preferences hash is accessed directly.
The idea behind this is that the value would always be encrypted inside the preferences' hash so that the end-user doesn't need to change the way it interacts with the preferences.
 
This implementation would not handle the encryption key, it's up to the end-user to safely generate, store, and rotate the key.
The key can be passed as an option to the preference (encryption_key), it the option is missing the system would look for an environment variable called SOLIDUS_PREFERENCES_MASTER_KEY, if even this variable is missing the system will use the Rails secret key.

I'm not sure that this solution would cover all the use cases that the author of the original issue had in mind, so I open the pull request as a draft to ask for feedback about possible missing use cases. 
 
**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/shared/preference_fields/_encrypted_string.html.erb
+++ b/backend/app/views/spree/admin/shared/preference_fields/_encrypted_string.html.erb
@@ -1,0 +1,12 @@
+<% label = local_assigns[:label].presence %>
+<% html_options = {class: 'input_string fullwidth'}.update(local_assigns[:html_options] || {}) %>
+
+<div class="field">
+  <% if local_assigns[:form] %>
+    <%= form.label attribute, label %>
+    <%= form.text_field attribute, html_options %>
+  <% else %>
+    <%= label_tag name, label %>
+    <%= text_field_tag name, value, html_options %>
+  <% end %>
+</div>
--- a/core/lib/spree/encryptor.rb
+++ b/core/lib/spree/encryptor.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Spree
+  # Spree::Encryptor is a thin wrapper around ActiveSupport::MessageEncryptor.
+  class Encryptor
+    # @param key [String] the 256 bits signature key
+    def initialize(key)
+      @crypt = ActiveSupport::MessageEncryptor.new(key)
+    end
+
+    # Encrypt a value
+    # @param value [String] the value to encrypt
+    # @return [String] the encrypted value
+    def encrypt(value)
+        @crypt.encrypt_and_sign(value)
+    end
+
+    # Decrypt an encrypted value
+    # @param encrypted_value [String] the value to decrypt
+    # @return [String] the decrypted value
+    def decrypt(encrypted_value)
+        @crypt.decrypt_and_verify(encrypted_value)
+    end
+  end
+end
--- a/core/lib/spree/preferences/preferable_class_methods.rb
+++ b/core/lib/spree/preferences/preferable_class_methods.rb
@@ -11,6 +11,7 @@ module Spree::Preferences
       password
       string
       text
+      encrypted_string
     )
 
     def defined_preferences
--- a/core/lib/spree/preferences/preferable_class_methods.rb
+++ b/core/lib/spree/preferences/preferable_class_methods.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'spree/encryptor'
+
 module Spree::Preferences
   module PreferableClassMethods
     DEFAULT_ADMIN_FORM_PREFERENCE_TYPES = %i(
@@ -16,7 +18,10 @@ module Spree::Preferences
     end
 
     def preference(name, type, options = {})
-      options.assert_valid_keys(:default)
+      options.assert_valid_keys(:default, :encryption_key)
+      preference_encryptor = preference_encryptor(options) if type == :encrypted_string
+
+      options[:default] = preference_encryptor.encrypt(options[:default]) if type == :encrypted_string
       default = options[:default]
       default = ->{ options[:default] } unless default.is_a?(Proc)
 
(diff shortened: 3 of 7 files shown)


core/spec/models/spree/preferences/preferable_spec.rb
```diff
@@ -266,6 +266,44 @@ def self.allowed_admin_form_preference_types
...
+        A.preference :secret, :encrypted_string,
+                     encryption_key: 'VkYp3s6v9y$B?E(H+MbQeThWmZq4t7w!'
+
+        @a.set_preference(:secret, 'secret_client_id')
+        expect(@a.get_preference(:secret)).to eq('secret_client_id')
+        expect(@a.preferences[:secret]).not_to eq('secret_client_id')
+      end
+
+      it "with string, encryption key provided as env variable" do
+        ENV['SOLIDUS_PREFERENCES_MASTER_KEY'] = 'VkYp3s6v9y$B?E(H+MbQeThWmZq4t7w!'
```
spaghetticode, line 281: I am not sure but I guess that this ENV variable setting may leak in subsequent specs run after this one 🤔

core/spec/models/spree/preferences/preferable_spec.rb
```diff
@@ -266,6 +266,44 @@ def self.allowed_admin_form_preference_types
...
+        A.preference :secret, :encrypted_string
+
+        @a.set_preference(:secret, 'secret_client_id')
+        expect(@a.get_preference(:secret)).to eq('secret_client_id')
+        expect(@a.preferences[:secret]).not_to eq('secret_client_id')
+      end
+
+      it "with string, encryption key provided as option, setted using syntactic sugar method" do
```
spaghetticode, line 289: setted -> set

core/lib/spree/preferences/preferable_class_methods.rb
```diff
@@ -9,14 +11,18 @@ module PreferableClassMethods
...
     def defined_preferences
       []
     end
 
     def preference(name, type, options = {})
-      options.assert_valid_keys(:default)
+      options.assert_valid_keys(:default, :encryption_key)
+      preference_encryptor = preference_encryptor(options) if type == :encrypted_string
+
+      options[:default] = preference_encryptor.encrypt(options[:default]) if type == :encrypted_string
```
spaghetticode, line 25: These postfixed `if` are the same for the 2 lines above, what about grouping them together in a single `if`?

spaghetticode:
@stefano-sarioli thanks for this PR. I left a few comments, let me know what you think about them. 
Also, the Mysql build is failing due to a flaky spec, rebasing with master will fix the problem. thanks!

core/spec/models/spree/preferences/preferable_spec.rb
stefano-sarioli: You were totally right, I've stubbed the request to ENV to avoid this.

core/spec/models/spree/preferences/preferable_spec.rb
stefano-sarioli: Fixed

core/lib/spree/preferences/preferable_class_methods.rb
stefano-sarioli: Usually, I don't like multiline if, but I've refactored the code using a multiline if and it's not bad, thanks for the feedback.

stefano-sarioli:
@spaghetticode Hi, I've pushed my fix to your comment, thanks again for the feedback.

core/lib/spree/encryptor.rb
```diff
@@ -0,0 +1,25 @@
...
+    # @param key [String] the 256 bits signature key
+    def initialize(key)
+      @crypt = ActiveSupport::MessageEncryptor.new(key)
+    end
+
+    # Encrypt a value
+    # @param value [String] the value to encrypt
+    # @return [String] the encrypted value
+    def encrypt(value)
+        @crypt.encrypt_and_sign(value)
```
jarednorman, line 15: The body of the encrypt and decrypt methods is indented 2 spaces too far.

backend/app/views/spree/admin/shared/preference_fields/_encrypted_string.html.erb
```diff
@@ -0,0 +1,12 @@
+<% label = local_assigns[:label].presence %>
+<% html_options = {class: 'input_string fullwidth'}.update(local_assigns[:html_options] || {}) %>
```
jarednorman, line 2: `Hash#update` just does the same thing as `Hash#merge` and we use `merge` a ton in the codebase already so I think we should prefer `merge` for consistency.

core/lib/spree/encryptor.rb
stefano-sarioli: Good catch, my bad fixed. Thank you 😄.

backend/app/views/spree/admin/shared/preference_fields/_encrypted_string.html.erb
stefano-sarioli: Hi @jarednorman, thanks for the feedback. I've looked at the other preference_fields partial on the backend and all of them are using `Hash#update` instead of `Hash#merge`. Probably we should change the method to `Hash#merge` for all the partials to be more consistent with the overall usage of the method inside the codebase, but probably I would prefer to do tath in another PR, what do you think about it?

Thanks again for your feedback 😄.

jarednorman: Interesting, I had not realized we had consistently used that in the other partials, though I tend to think we should still prefer `merge` as we use it pretty heavily in other parts of the codebase.

stefano-sarioli: I don't have a strong opinion about it, so I've switched from `update` to `merge`

stefano-sarioli:
Hi @jarednorman, does the change I made thanks to your feedbacks looks good to you?

Issue #3679, opened 2020-06-23
[RFC] Remove required validation from address phone numbers


seand7565:
**Is your feature request related to a problem? Please describe.**
Solidus requires phone numbers in addresses by default. Since these phone numbers aren't used by Solidus (besides being displayed to admin) it's not obvious why this field is required, and it being required has the possibility of [decreasing conversions](https://baymard.com/blog/explain-phone-number-field). 

Additionally, as a consumer, required phone numbers in eCommerce is a bit of a pet peeve of mine - you already have my email address if you need to contact me!

**Describe the solution you'd like**
I think this should not be a required field by default, and the documentation should be updated to describe how to make it required if desired by the store.

seandawes:
Many stores will have this as a requirement as it is a required piece of information to run instant financing approvals in the cart such as Affirm. cc @peterberkenbosch 

We ran into issues with paypal/braintree solidus gem and phone number formatting so we don't pass that data point anymore.

Marketing requirements may want it for order tracking or marketing customers wish to opt in to (think flash sales etc) which is done via twilio or other services such as [Klaviyo ](https://www.klaviyo.com/features/sms-marketing) which is on the roadmap

seand7565:
I totally see why it would be a requirement for some stores, especially for cases like Affirm - but IMO Solidus shouldn't be opinionated on what a customer _might_ integrate with. For the handful of Solidus stores I've worked on, I've never seen a case where phone number was required for anything important, so I think it may be overkill to require it on all Solidus stores by default - especially when it has an impact on conversions.

seandawes:
Its gonna be needed for any mainstream enterprise store but Im in favor of stripping it from core and having core do less. This could also be addressed in overhauling the full checkout which is bad for CRO at its current stage. ie. needs single page checkout

aldesantis:
I agree that the phone number shouldn't be required by default. I think the idea was to also use the information for fraud detection, but it's a decision that should be left to the individual store.

jarednorman:
I'm fine making that validation optional. Some stores need to require phone numbers and others don't. Seems like the perfect case for adding an option.

peterberkenbosch:
totally onboard with this. For some shipping providers they like to have a phone number to text delivery time updates or just call to get the delivery made. That being said, having that as an optional validation sounds like the best way forward.

PR #3685, opened 2020-06-26
Add preference for phone validation

seand7565:
**Description**
Adds phone number validation preference so users can easily indicate
whether or not they want to require phone numbers for addresses.
Ref: #3679

**Checklist:**

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -149,9 +149,9 @@ module Spree
 
     # @todo Remove this from the public API if possible.
     # @return [true] whether or not the address requires a phone number to be
-    #   valid
+    #   present
     def require_phone?
-      true
+      Spree::Config[:address_requires_phone]
     end
 
     # @todo Remove this from the public API if possible.
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -155,7 +155,7 @@ module Spree
     end
 
     # @todo Remove this from the public API if possible.
-    # @return [true] whether or not the address requires a zipcode to be valid
+    # @return [true] whether or not the address requires a zipcode to be present
     def require_zipcode?
       true
     end
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -26,6 +26,10 @@ module Spree
   class AppConfiguration < Preferences::Configuration
     # Preferences (alphabetized to more easily lookup particular preferences)
 
+    # @!attribute [rw] address_requires_phone
+    #   @return [Boolean] should phone number be required (default: +true+)
+    preference :address_requires_phone, :boolean, default: true
+
     # @!attribute [rw] address_requires_state
     #   @return [Boolean] should state/state_name be required (default: +true+)
     preference :address_requires_state, :boolean, default: true
--- a/core/spec/models/spree/address_spec.rb
+++ b/core/spec/models/spree/address_spec.rb
@@ -136,9 +136,9 @@ RSpec.describe Spree::Address, type: :model do
     end
 
     context "phone not required" do
-      before { allow(address).to receive_messages require_phone?: false }
+      before { stub_spree_preferences(address_requires_phone: false) }
 
-      it "shows no errors when phone is blank" do
+      it "is valid when phone is blank" do
         address.phone = ""
         address.valid?
         expect(address.errors[:phone].size).to eq 0
--- a/guides/source/developers/users/addresses.html.md
+++ b/guides/source/developers/users/addresses.html.md
@@ -64,21 +64,17 @@ You may want to alter Solidus's address requirements for your store. For
 example, if you do not require customer phone numbers in order for them to check
 out.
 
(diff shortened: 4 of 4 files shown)


core/app/models/spree/address.rb
```diff
@@ -148,14 +148,14 @@ def active_merchant_hash
     end
 
     # @todo Remove this from the public API if possible.
-    # @return [true] whether or not the address requires a phone number to be
-    #   valid
+    # @return [false] whether or not the address requires a phone number to be
+    #   present
     def require_phone?
-      true
+      false
```
aldesantis, line 154: What do we do for stores that relied on this being `true` by default? If I require the phone number in my store, I have not overridden `require_phone?`. When I update Solidus, the phone will become optional without me having touched anything, which is not ideal.

I think we should formally deprecate this. We can introduce a new preference called `require_phone_on_address` (or just `require_phone`?) which this method relies on.

The default value for the preference would be `nil` and, when it's `nil`, we cast it to `true` and print a deprecation warning notifying the user that they need to set an explicit value, because the new behavior in 3.0 will be for phone numbers to be optional.

In 3.0, we just switch the default value for the preference to `false`, and the phone number will become optional.

@kennyadsl is our deprecation expert, so he should also chime in probably. 😜

core/app/models/spree/address.rb
```diff
@@ -148,14 +148,14 @@ def active_merchant_hash
...
     end
 
     # @todo Remove this from the public API if possible.
-    # @return [true] whether or not the address requires a zipcode to be valid
+    # @return [true] whether or not the address requires a zipcode to be present
```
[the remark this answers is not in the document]
seand7565: Yes, will do!

core/app/models/spree/address.rb
seand7565: Sorry for the late response here - the new version of the PR addresses this with a first pass at deprecation. How does it look now?

seand7565:
I've had the opposite experience - The stores I've worked with prefer not to collect phone numbers (more fields on checkout -> lower conversion, at least in theory) and personally it's something that I'd prefer not to give out as a consumer, especially if there's no obvious reason for it - it just feels like data collection for the sake of data collection. 

I'd like Solidus to be less opinionated about what is and isn't required during checkout - IMO anything not immediately necessary to pay for and ship the order should be optional by default.

jarednorman:
All fair points. I wish we had a way to gather info about what stores needed more generally. I can only speak to the fact that all of the stores we're currently managing (five or so) require phone and will have to change this setting from the new default if we adopt this. It's obviously only an extremely minor inconvenience, but it makes me second-guess whether it's the right call or whether this is going to be a minor inconvenience to quite a lot of stores.

core/lib/spree/app_configuration.rb
```diff
@@ -26,6 +26,10 @@ module Spree
   class AppConfiguration < Preferences::Configuration
     # Preferences (alphabetized to more easily lookup particular preferences)
 
+    # @!attribute [rw] address_requires_phone
+    #   @return [Boolean] should phone number be required (default: +nil+)
+    preference :address_requires_phone, :boolean
```
kennyadsl, line 31: What about setting a default anyway? I think that being explicit with false is better.

kennyadsl (commented):
I'd also like to ask less info to users during checkout but I think it's better to keep the old behavior as default without real data to confirm our ideas.

core/lib/spree/app_configuration.rb
seand7565: The reason it's nil here is so we can check to see if it's been explicitly set by the user. If it hasn't been set, a deprecation warning will display, letting the user know that this function will default to false in the future, and if they want to maintain the existing functionality, they will need to explicitly set this preference to "true". 

The idea being that we add "false" as the default when 3.0 rolls around and remove the deprecation warning.

Personally, I think the change is innocuous enough - since Solidus doesn't depend on phone numbers for anything, and it's easy enough for stores to enable the validation with this change. We're also making sure that existing stores are aware of the change, and they'll have a chance to adjust the functionality however they want. However, I understand that there are some strong opinions about this change, so I'll do whatever the core team decides here. If you'd rather this be a PR that just adds a preference that defaults to "true", then that's what we'll do.

kennyadsl: This makes sense Sean but we have no guarantees that devs will look at the deprecation warning before deploying so, in this form, this is considered a breaking change. I'll talk with the rest of the Core Team today in our meeting and see what they think about it.

kennyadsl: @seand7565 we agreed that it's better to keep the existing behavior as the default. I think what you did to force users to choose can work but if they don't we can't change their current behavior.

seand7565: Very well! I'll refactor this PR as just an added preference that defaults to true.

seand7565: Done - let me know how it looks!

jarednorman (approved):

aldesantis (approved):

PR #3688, opened 2020-07-01
Do not attempt to create multiple records with `global_zone` factory

spaghetticode:
**Description**

The factory linting spec is breaking many PR builds lately with this error:

```
  FactoryBot::InvalidFactoryError:
    The following factories are invalid:

    * global_zone - Validation failed: Name has already been taken
    (ActiveRecord::RecordInvalid)

```
I think this has something to do with the fact that the factory `global_zone` can be used only once, as the name is fixed to "Global Zone" and the model has a uniqueness validation on `name`. Still, I don't really understand why the problem happens, as the DB should be clean when this spec example is run.

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/factories/zone_factory.rb
+++ b/core/lib/spree/testing_support/factories/zone_factory.rb
@@ -5,7 +5,7 @@ require 'spree/testing_support/factories/country_factory'
 
 FactoryBot.define do
   factory :global_zone, class: 'Spree::Zone' do
-    name { 'GlobalZone' }
+    initialize_with { Spree::Zone.find_or_initialize_by(name: 'GlobalZone') }
     zone_members do |proxy|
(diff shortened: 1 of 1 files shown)


aldesantis:
How about this?
```ruby
factory :global_zone do
  initialize_with { Spree::Zone.find_or_initialize_by(name: name) }
  # ...
end
```

spaghetticode:
@aldesantis thank you, that's exactly what I was looking for but somehow I was not able to find it... this seems to be such a better solution. I'm updating the PR.

softr8:
ah! I had to do the same in this commit: https://github.com/solidusio/solidus/pull/3626/commits/3180cbf96af81d81a46e9b1bc0b1aa5c65af6f69

Will rebase from master

PR #3693, opened 2020-07-07
Fix typo in comment in promotion.rb

brchristian:


brchristian:
Absolutely, sorry about that. I've made the update in the PR and the commit.

Issue #3692, opened 2020-07-06
Square logos look bigger than we expect in the admin


SamBelmor:
**Solidus Version**:

2.11.0.alpha

**To Reproduce**

1. Change the admin logo by a square logo.
2. Restart the server to watch the change.

**Current behavior**

1. Square images look bigger than we expect.

**Expected behavior**

1. The square logo should keep in their container.

**Screenshots**

**Desktop (please complete the following information):**
### Chrome
![Screen Shot 2020-07-06 at 16 04 37](https://user-images.githubusercontent.com/13225845/86643607-76629200-bfa2-11ea-9c9c-7ffd7afe99f1.png)

**Smartphone (please complete the following information):**
### Iphone 6
![Screen Shot 2020-07-06 at 12 34 11](https://user-images.githubusercontent.com/13225845/86643892-af026b80-bfa2-11ea-8e6d-5b6bee379089.png)
![Screen Shot 2020-07-06 at 13 23 27](https://user-images.githubusercontent.com/13225845/86643902-b0339880-bfa2-11ea-904f-f57b25361e87.png)

**Additional context**

When only icons with no text are shown in the admin section, square images look bigger than expected, which causes a problem because the logo covers the elements close to it. I don't know if it is necessary to specify the required size of the image in the documentation in order to have the expected behavior, or if it is necessary to readjust the image in this scenario, I mean, by disappearing the image when only the icons are shown or just readjusted the logo.

PR #3702, opened 2020-07-10
[Admin] Fix Square Logos appearance

michaelmichael:
**Description**
Ref #3692

This just adds an explicit height to the `.brand-link` div in admin. This height is controlled by the `$main-header-height` variable. Then it sets the image to have a max-width/height of 100%. It should prevent logos of any size or aspect ratio from overflowing and preventing access to the ui. See square, tall, and wide logos below:

Screen Shot 2020-07-10 at 3 50 29 PM
Screen Shot 2020-07-10 at 3 53 23 PM
Screen Shot 2020-07-10 at 3 53 55 PM

**Checklist:**

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
@@ -355,8 +355,11 @@ nav.menu {
 
 .brand-link {
   display: block;
+  height: $main-header-height - 1px;
+  overflow: hidden;
 
   img {
-    max-width: 125px;
+    max-height: 100%;
+    max-width: 100%;
   }
 }


backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
```diff
@@ -355,8 +355,11 @@ nav.menu {
 
 .brand-link {
   display: block;
+  height: $main-header-height - 1;
```
jarednorman, line 358: Might be worth giving the `1` a unit. I imagine other things might break if someone overrode the height with a `rem` value, but we might not as well not add to that.

```suggestion
  height: $main-header-height - 1px;
```

michaelmichael:
Ok, squashed into a single commit.

Issue #3697, opened 2020-07-09
Payment method auto-capture boolean provides 3 options to the user


seand7565:
The value of auto-capture is a boolean, but has three possible values in the following image: `false`, `yes`, and `no`.

![image](https://user-images.githubusercontent.com/5720486/87042756-b6e82880-c1b9-11ea-982a-2263c9bae57c.png)

IMO we should use `true` and `false` instead of `yes` and `no`, but we could also set the value of app default to `yes` and `no` - as long as we're consistent with our options.

aldesantis:
@seand7565 hmm, not sure this is an actual issue. The first option corresponds to `nil`, which tells Solidus to use the application's default `auto_capture` setting. In this case, that setting is `false`, so it's displayed next to "Use App Default" for the user's convenience.

seand7565:
@aldesantis Right, behind the scenes the value is either `true` or `false`, but the dropdown indicates to the user that there are 3 possible values - `false`, `yes` or `no`. As a user, I'd expect to see something like this:
![image](https://user-images.githubusercontent.com/5720486/87049502-abe5c600-c1c2-11ea-9529-80a412857670.png)
Or this:
![image](https://user-images.githubusercontent.com/5720486/87049610-d5065680-c1c2-11ea-9775-b9188ef147f5.png)

(I prefer the first option, but maybe that's just me)

No problem with the logic, just nitpicking at the text in the provided options. 😛

aldesantis:
Ah, sorry, didn't realize you were referring to the label! I think the best option is to use yes/no consistently, since true/false only makes sense for developers.

jarednorman:
Yeah, for consistency I think the ideal options are probably:

- Use App Default (No)
- Yes
- No

michaelmichael:
I believe I fixed this in #3703 according to the consensus here.

- Use App Default (No)
- Yes
- No

However, I wonder if it would be better to remove the redundant option from the dropdown to reduce cognitive overhead. 

While having the third option makes sense to the store's developer, it could be confusing to users. Why have three answers to a yes or no question? The user may wonder if there is any difference between "Use App Default (No)" and "No"—At least, I wondered this when first setting up payment methods.

So, my pull request as submitted makes the verbiage more consistent. That said, I could update it if others agree that reducing it to just two options is preferable, as in:

- Use App Default (No)
- Yes

To take it one step further, I think something like the following would be easiest to scan and parse:
- No (Default)
- Yes

[...]

PR #3703, opened 2020-07-11
Change to true/false to yes/no in Auto Capture Select Text

michaelmichael:
**Description**
  Ref #3697

This modifies the text in the Auto Capture dropdown menu to make the verbiage more consistent. It checks whether `Spree::Config.auto_capture` is set to true or false. If it is set to true, the primary dropdown will read "Use App Default (Yes)". If it is set to "false"—or if it is not set (the default config state)—it will read "Use App Default (No)".

`Spree::Config.auto_capture = true`:
Screen Shot 2020-07-10 at 4 38 30 PM

`Spree::Config.auto_capture = false`:
Screen Shot 2020-07-10 at 4 39 07 PM

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/payment_methods/_form.html.erb
+++ b/backend/app/views/spree/admin/payment_methods/_form.html.erb
@@ -31,7 +31,12 @@
         </div>
         <div data-hook="auto_capture" class="field">
           <%= f.label :auto_capture %>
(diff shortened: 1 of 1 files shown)


aldesantis:
Looks good and I agree with @jarednorman. @michaelmichael perhaps you can just put the arguments on different lines? I think that alone would make them easier to parse. 🙏

kennyadsl:
@michaelmichael let us know if you have time to make the suggested improvements, even in another PR so we can go ahead merging this. Thanks a lot for your contributions!

PR #3704, opened 2020-07-14
Selectively disable stock item rows when some do not track inventory

felixyz:
When some variants of a product have "track inventory" set to `true` and at least one variant of the same product has "track inventory" set to `false`, all the rows on the "Product Stock" tab become disabled, and show the `"Track Inventory" disabled for this variant` popup.

After this change, only the rows for variants with "track inventory" set to `false` are disabled.

I don't think there are any frontend tests that would cover this, but would be happy to add tests if someone can point me in the right direction.

### DIFF
--- a/backend/app/controllers/spree/admin/stock_items_controller.rb
+++ b/backend/app/controllers/spree/admin/stock_items_controller.rb
@@ -52,6 +52,7 @@ module Spree
       def variant_scope
         scope = Spree::Variant.accessible_by(current_ability)
         scope = scope.where(product: @product) if @product
+        scope = scope.order(:sku)
         scope
       end
 
--- a/backend/spec/features/admin/products/stock_management_spec.rb
+++ b/backend/spec/features/admin/products/stock_management_spec.rb
@@ -13,10 +13,13 @@ describe "Product Stock", type: :feature do
     let!(:stock_location) { create(:stock_location, name: 'Default') }
     let!(:product) { create(:product, name: 'apache baseball cap', price: 10) }
     let!(:variant) { create(:variant, product: product) }
+    let!(:variant2) { create(:variant, product: product, track_inventory: false) }
     let(:stock_item) { variant.stock_items.find_by(stock_location: stock_location) }
+    let(:stock_item2) { variant2.stock_items.find_by(stock_location: stock_location) }
 
     before do
       stock_location.stock_item(variant).update_column(:count_on_hand, 10)
+      stock_location.stock_item(variant2).update_column(:count_on_hand, 13)
 
       click_nav "Products"
       within_row(1) { click_icon :edit }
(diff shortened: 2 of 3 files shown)


aldesantis:
@felixyz thanks! For the test, I think you could write a feature spec that reproduces the scenario and verifies that only the expected rows are disabled rather than all of them. You can take a look at the backend's existing specs for inspiration, or let me know if you need any help!

felixyz:
@aldesantis Will do, thanks for the pointer!

backend/app/controllers/spree/admin/stock_items_controller.rb
```diff
@@ -52,6 +52,7 @@ def load_stock_management_data
       def variant_scope
         scope = Spree::Variant.accessible_by(current_ability, :read)
         scope = scope.where(product: @product) if @product
+        scope = scope.order(:sku)
```
felixyz, line 55: Had to introduce this to make the spec deterministic. (The bug only manifests when the row which does not track inventory appears in the list *after* the row which doesn't.) AFAICS, there there is no pre-existing supposition about the ordering, so this shouldn't break anything (maybe a slight improvement in itself).

backend/spec/features/admin/products/stock_management_spec.rb
```diff
@@ -42,46 +45,64 @@
...
 
       click_link "Product Stock"
       within("tr#spree_variant_#{variant.id}") do
         expect(find(:css, "input[type='checkbox']")).not_to be_checked
       end
     end
 
-    def adjust_count_on_hand(count_on_hand)
-      within("tr#spree_variant_#{variant.id}") do
+    def adjust_count_on_hand(variant_id, count_on_hand)
```
felixyz, line 72: I ended up not making use of the changes to `adjust_count_on_hand` and `toggle_backorderable` but I think this is a slight improvement anyway so I'd leave it here unless it's considered bad form.

backend/spec/features/admin/products/stock_management_spec.rb
```diff
@@ -42,46 +45,64 @@
...
+    def toggle_backorderable(variant_id, value: true)
+      within("tr#spree_variant_#{variant_id}") do
         find(:css, "input[type='checkbox']").set(value)
         click_icon :check
       end
       expect(page).to have_content('Updated Successfully')
     end
 
+    context "with two variants, one of which tracks inventory while the other doesn't" do
+      it "allows modifying backorderable only on the variant which tracks inventory", js: true do
```
felixyz, line 89: I tried doing a hover here as well, like in the next example below, but for some reason Selenium/ChromeDriver doesn't pick up the tooltip on this one. Didn't seem worth the effort to solve this.

felixyz:
@aldesantis I finally got around to writing some specs, please have a look when you can.

(The failing CI test seems unrelated.)

aldesantis (approved):
I think this looks good, thanks @felixyz!

kennyadsl (approved):
Thanks @felixyz !

PR #3710, opened 2020-07-16
[v2.8] Permit address attributes via payment source attributes

kennyadsl:
**Description**

This is a complementary PR for the last security fix. Bill address can be passed as params when submitting a new source.

We are not going to wait 24h before merging this.


kennyadsl:
Specs are passing locally:

```text
solidus/backend on v2.8-permit_address-source
> be rspec spec/features/admin/orders/payments_spec.rb

...

Finished in 18.12 seconds (files took 6.26 seconds to load)
17 examples, 0 failures

Randomized with seed 51535
```

Also, looks like a flaky since on this job there are different failures (only the one related to frontend): https://app.circleci.com/pipelines/github/solidusio/solidus/1331/workflows/40de94c9-83f6-42cf-96f9-dcc12941cc55/jobs/19454/parallel-runs/2?filterBy=ALL.

I'm going to merge and release the patch for v2.8 and investigate later why this happens.

PR #3711, opened 2020-07-16
[v2.9] Permit address attributes via payment source attributes

kennyadsl:
**Description**
This is a complementary PR for the last security fix.

Bill address can be passed as params when submitting a new source.

This also resumes a spec that was removed a while ago and apparently it was the only one testing this feature.


kennyadsl:
Merging, the Rails 5.1 on Postgres failure is a flaky we fixed in master.

Issue #3669, opened 2020-06-16
Support "backend only" applications


tmtrademarked:
**Is your feature request related to a problem? Please describe.**
One of our services is logically a "backend only" application using Solidus. We're using Solidus to power the database models, etc as well as provide a friendly framework for an admin user interface. We then use other internal APIs to communicate with it (gRPC, Kafka, etc)

But it seems like the `solidus_backend` gem ends up including the `solidus_api` gem under the hood - which means our application technically exposes all of the API routes exposed by a normal Solidus application, which isn't what we want. We don't want to have routes that we aren't explicitly using, since then we have to take care to secure/maintain/monitor those routes as well.

**Describe the solution you'd like**
Ideally, I'd love to see a world where `solidus_backend` *only* mounts the routes needed for the admin interface, not *all* of the routes of `solidus_api`. (At least, this is what seems to be happening - I apologize if my understanding of the route framework is incorrect!)

**Describe alternatives you've considered**
I've looked at removing/aliasing/etc, but thus far, no solution has felt maintainable, especially if `solidus_api` adds more functionality in the future.

aldesantis:
@tmtrademarked 👋 Unfortunately, I don't think there's a good way to do what you're asking right now (other than removing all the routes from solidus_api with some hack that leverages Rails internals, but that doesn't seem very robust).

One way we could do it is by creating a preference in solidus_api that determines whether the engine should only create the routes that are required by the backend or all of them, e.g.
```ruby
# config/initializers/spree.rb
Spree::Api::Config[:backend_routes_only] = true
```
Would something like that work for your use case?

tmtrademarked:
Something like that sounds like it would work perfectly! That would be a totally fine solution from my perspective.

aldesantis:
Got it, I think that makes sense too and we shouldn't have any issues implementing in a backwards-compatible manner, since the setting would be `false` by default. Can't promise anything right now because I haven't thought through the edge cases yet, but I'll try to do a spike in the next few days. 🙂

kennyadsl:
I think the way to go here is to remove the dependency we have in solidus_backend from the API by just using Rails. I closed the PR with the implementation proposed by @aldesantis because it didn't go ahead after some reviews, and I think we should change the approach, resolving the issue at its root.

aldesantis:
I agree. Thinking about this at a higher level, our goal should be to make each gem completely independent of the others (other than solidus_core).

jarednorman:
[...]

PR #3715, opened 2020-07-17
Add disable_api_routes API config

gsmendoza:
## Description

Implements #3669.

## Goal

```
As a `solidus_backend` user
I want `solidus_backend` to only mount the routes needed for the admin interface,
  and not all of the routes of `solidus_api`
So that we don't have to take care of securing/maintaining/monitoring those API routes
```

### Scenario: Setting `Spree::Api::Config[:disable_api_routes]` excludes api routes from being mounted

```
Given I have a Rails app
And I added `solidus_backend` to the app
When I set `Spree::Api::Config[:disable_api_routes]` to true
And I run `rails routes`
Then I should not see any of the `Spree::Core::Engine` api routes
And I should continue seeing the `Spree::Core::Engine` backend routes
```

## Demo

- Please check https://github.com/gsmendoza/solidus-3669-demo.

## Notes on implementation

* I changed the proposed name of `backend_routes_only` to `disable_api_routes` because I think it's ideal if `solidus_api` doesn't know anything about backend logic.
* I checked a similar configuration, `Spree::Api::Config[:requires_authentication]`, and didn't find any specs that directly test the configuration. That being said, I was able to add a routing spec for  `disable_api_routes`.
* I'm not sure if I need to add a default setting for the preference in `solidus/core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt`.
* I didn't see any documentation on `requires_authentication` in the repo.

## Checklist

### DIFF
--- a/api/spec/routing/api_routes_spec.rb
+++ b/api/spec/routing/api_routes_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'api', type: :routing do
+  context 'when disable_api_routes is enabled' do
+    before do
+      stub_spree_preferences(Spree::Api::Config, disable_api_routes: true)
+      Rails.application.reload_routes!
+    end
+
+    it "does not mount the api routes" do
+      aggregate_failures do
+        expect(get: '/api/orders').to_not be_routable
+        expect(get: '/api/products').to_not be_routable
+      end
+    end
+  end
+
+  context 'when disable_api_routes is disabled' do
+    before do
+      stub_spree_preferences(Spree::Api::Config, disable_api_routes: false)
+      Rails.application.reload_routes!
+    end
+
+    it "mounts the api routes" do
+      aggregate_failures do
+        expect(get: '/api/orders').to be_routable
+        expect(get: '/api/products').to be_routable
+      end
+    end
+  end
+end
--- a/api/spec/routing/api_routes_spec.rb
+++ b/api/spec/routing/api_routes_spec.rb
@@ -3,6 +3,10 @@
 require 'spec_helper'
 
 describe 'api', type: :routing do
+  let!(:current_disable_api_routes_setting) do
+    Spree::Api::Config[:disable_api_routes]
+  end
+
   context 'when disable_api_routes is enabled' do
     before do
       stub_spree_preferences(Spree::Api::Config, disable_api_routes: true)
@@ -29,5 +33,14 @@ describe 'api', type: :routing do
         expect(get: '/api/products').to be_routable
       end
     end
+  end
+
+  after do
+    stub_spree_preferences(
+      Spree::Api::Config,
+      disable_api_routes: current_disable_api_routes_setting
+    )
+
+    Rails.application.reload_routes!
   end
 end
--- a/api/config/routes.rb
+++ b/api/config/routes.rb
@@ -12,133 +12,135 @@ Spree::Core::Engine.routes.draw do
     end
   end
 
-  namespace :api, defaults: { format: 'json' } do
-    resources :promotions, only: [:show]
+  unless Spree::Api::Config[:disable_api_routes]
+    namespace :api, defaults: { format: 'json' } do
+      resources :promotions, only: [:show]
 
-    resources :products do
-      resources :images
-      resources :variants
-      resources :product_properties
-    end
+      resources :products do
+        resources :images
+        resources :variants
+        resources :product_properties
+      end
 
-    concern :order_routes do
-      resources :line_items
-      resources :payments do
+      concern :order_routes do
+        resources :line_items
+        resources :payments do
+          member do
+            put :authorize
+            put :capture
+            put :purchase
+            put :void
+            put :credit
+          end
+        end
+
+        resources :addresses, only: [:show, :update]
+
+        resources :return_authorizations do
+          member do
+            put :cancel
+          end
+        end
+
+        resources :customer_returns, except: :destroy
+      end
+
+      resources :checkouts, only: [:update], concerns: :order_routes do
         member do
-          put :authorize
-          put :capture
-          put :purchase
-          put :void
-          put :credit
+          put :next
+          put :advance
+          put :complete
         end
       end
 
-      resources :addresses, only: [:show, :update]
+      resources :variants do
+        resources :images
+      end
 
-      resources :return_authorizations do
+      resources :option_types do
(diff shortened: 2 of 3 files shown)


api/spec/routing/api_routes_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  after do
+    stub_spree_preferences(
+      Spree::Api::Config,
+      disable_api_routes: current_disable_api_routes_setting
+    )
+
+    Rails.application.reload_routes!
```
gsmendoza, line 44: It seems the routes have to be reloaded otherwise succeeding specs might not see the api routes. See https://app.circleci.com/pipelines/github/solidusio/solidus/1345/workflows/d4444e92-db5e-4f7a-a4db-acdf4723aa59/jobs/19514.

spaghetticode:
@gsmendoza thanks for this PR. Did you consider that some API endpoints are used in the admin area as well? The ones I can think of are mostly AJAX requests. For example, with the API routes disabled, trying to choose a customer for a new order would fail like in this screenshot:

Screen Shot 2020-07-17 at 2 42 01 PM

gsmendoza:
Hi @spaghetticode!

Yeah, I didn't consider that some API endpoints are used in the admin area. Going back to the [issue](https://github.com/solidusio/solidus/issues/3669), it seems my initial understanding of the @aldesantis's `backend_routes_only` wasn't correct:

* What I initially understood - `backend_routes_only` would only allow admin routes. It would exclude api routes.
* What I now think @aldesantis actually meant - `backend_routes_only` would only allow _api_ routes used in the backend.

Grepping the code, I can see that there are at least 17 api routes called from the backend:

* `Spree.pathFor('api/orders`
* `Spree.pathFor('api/checkouts`
* `Spree.pathFor('api/option_types`
* `Spree.pathFor('api/option_values`
* `Spree.pathFor('api/classifications`
* `Spree.pathFor('api/products`
* `Spree.pathFor('api/shipments`
* `Spree.pathFor('api/stock_locations`
* `Spree.pathFor('api/taxons/products`
* `Spree.pathFor('api/taxons`
* `Spree.pathFor('api/variants`
* `Spree.pathFor('api/users`
* `Spree.pathFor('api/states`
* `Spree.pathFor('api/shipments/transfer_to_`location
* `Spree.pathFor('api/shipments/`transfer_to_shipment
* `Spree.pathFor('api/states`
* `Spree.pathFor('api/taxonomies`

If we move forward with implementing `backend_routes_only`, I think it would add maintenance costs to Solidus:

1. If a new api call is added to the backend, the api route has to be added to `backend_routes_only` as well. Otherwise, the api call would break if `backend_routes_only` is enabled (like what we're seeing now).
2. Likewise, if an api call is removed from the backend, the api route has be removed from `backend_routes_only` as well.

It would also be ideal if we can test that the backend Ajax calls are working if `backend_routes_only` is enabled.

To summarize, in order to get this PR done, I would have to:

1. Rename `disable_api_routes` to `backend_routes_only`.
1. Identify which api calls to include to `backend_routes_only`, and manually test them individually.
1. Provide a mechanism to test backend with `backend_routes_only` enabled.

Finally, just to give you a background of how I got involved with this PR, I was asked by @mtylty to work on this PR as a technical assessment for my application to Nebulab. Considering that this PR seems more complicated than I initially assessed, I have asked @mtylty to proceed with my technical assessment even if I haven't completed this PR. This would allow me to work on this PR on a purely voluntary basis. However, if you know of other issues that might be more suitable for me as a first time contributor, please let me know.

I appreciate your guidance!

spaghetticode:
I think the most relevant part of the problem here is that ATM we don't have a clear way to discriminate if a route is used in the backend, except for digging into the code (most likely only JS) that actually uses it. 

One possible course of action may be to find out all these routes and duplicate them under a new scope, something like this:

```ruby
# api/config/routes.rb

Spree::Core::Engine.routes.draw do
  # new admin scope
  scope :admin do
    namespace :api do
      resources :states, only: [:index], controller: 'states'
    end
   # ...
  
  # existing API routes:
  namespace :api, defaults: { format: 'json' } do
    resources :states, only: [:index, :show]
  end
end
```

And then we start to use these `admin` routes in the JS, for example:

```js
// backend/app/assets/javascripts/spree/backbone/collections/states.js

Spree.Collections.States = Backbone.Collection.extend({
  # ...
  url: function () {
    return Spree.pathFor('admin/api/states?country_id=' + this.country_id)
  }
}
```

Of course, this would mean many duplicate API routes, but off the top of my head, I can't think about a cleaner solution.

Once the routes are grouped like that, it would be easy to remove the ones not used in the backend the way you did in this PR. Maybe someone else has a nicer solution?

gsmendoza:
@spaghetticode Thanks! I think that's a better approach than the conditional config. If there's no other suggestions, I'll give it a try tomorrow.

jarednorman:
I mostly worried about the maintainability of any of these solutions. You could always decorate the API base controller to add a filter that rejects any request that isn't from an admin user and that would likely address the problem at hand, though maybe not in the most ideal way.

brchristian:
I share the concerns of https://github.com/solidusio/solidus/issues/3669, so it's very exciting to see the progress being made here!

This may be a naive question, but is there a way to refactor the `backend` code so that `solidus_backend` avoids using the Solidus API altogether? If possible, that would seem to be a rather clean way to do it such that a Solidus store can truly just use the `frontend`, `backend`, and/or `api` totally independently of one another.

jarednorman:
I'm not sure what the path would be to accomplishing that, but I personally would prefer if the admin didn't depend on the API at all either. I'd be concerned about the churn required to get there, though. It would likely make upgrading very difficult for stores with a lot of customizations.

brchristian:
@jarednorman re. removing dependency on the API, I wonder if it’s a matter of identifying the API calls being made by backend JS, and then for each call creating a `backend` Controller action that responds to the `js` format and that spits out whatever JSON, etc., those API calls would.

Because they are backend controller routes, they should naturally check for admin authentication.

jarednorman:
It would also require changes to the assets to hit the right endpoints, as well as various refactoring to avoid creating unnecessary duplication with the added endpoints. I think it would be a significant enough change that the rest of the core team would need to way in before we head this direction.

kennyadsl:
I also think the path here is to have the admin not dependant on the API anymore, and I think there are ways of doing this now.  Going to close, thanks anyway for the great work here!

Issue #3670, opened 2020-06-17
Solidus uses the incorrect states for Italy


seand7565:
Italian addresses [utilize provinces for states](https://italian.stackexchange.com/questions/4039/how-to-write-address-in-italian), however Solidus default states for Italy use regions. This leads to addresses in Solidus like:
`Strada Comunale Piana, 1, 65129 Pescara, Abruzzo, Italy`
which is incorrect and might lead to some lost packages or third party validation issues - the correct address is:
`Strada Comunale Piana, 1, 65129 Pescara PE, Italy`
where PE is the abbreviation for the province of Pescara.

We should swap to using provinces for states, as this is the standard for Italian addresses.

seand7565:
Looking into this a bit further - we get our states from Carmen. We get the `Carmen::Country` and then the subregions of that country. For Italy, we actually want to get the subregions subregions. `Carmen::Country.coded("IT").subregions.map(&:subregions)`

There are only a handful of countries with nested subregions, and only nested up to twice - there are no countries with subregions that have subregions that have subregions. My guess is that for each of these countries, we actually need the nested subregions instead of the top level subregions. I will attempt to do some research for each of these to verify that theory. If it's true, then this should be a simple fix. 

Countries with nested subregions: 

```
<#Carmen::Country name="Albania">
<#Carmen::Country name="Azerbaijan">
<#Carmen::Country name="Bosnia and Herzegovina">
<#Carmen::Country name="Bangladesh">
<#Carmen::Country name="Belgium">
<#Carmen::Country name="Burkina Faso">
<#Carmen::Country name="Canada">
<#Carmen::Country name="Cabo Verde">
<#Carmen::Country name="Czechia">
<#Carmen::Country name="Spain">
<#Carmen::Country name="France">
<#Carmen::Country name="United Kingdom">
<#Carmen::Country name="Guinea">
<#Carmen::Country name="Equatorial Guinea">
<#Carmen::Country name="Greece">
<#Carmen::Country name="Guinea-Bissau">
<#Carmen::Country name="Indonesia">
<#Carmen::Country name="Ireland">
<#Carmen::Country name="Italy">
<#Carmen::Country name="Saint Kitts and Nevis">
<#Carmen::Country name="Sri Lanka">
<#Carmen::Country name="Morocco">
<#Carmen::Country name="Marshall Islands">
<#Carmen::Country name="Maldives">
<#Carmen::Country name="Malawi">
<#Carmen::Country name="Mexico">
[...]

PR #3722, opened 2020-07-29
Fix Italy state seed generation

seand7565:
**Description**
Previously, Italy was using the top-level `subregions` - `regions`
in Carmen to generate states, however for a mailing address this
is incorrect, and it instead needs to use `provinces`, which is
nested as the second level of `subregions` in Carmen. This fixes the
seed generation, allowing for more countries to make this switch
as necessary (there are a handful of other countries that need this)

<strike>Also, adds a `state:regenerate` rake task that updates states for
countries that were generated incorrectly, like Italy. This is so
people updgrading from v2.10 to v2.11 can easily update to the
correct state list.</strike> This has been moved to a rake task, if you want
this functionality, you'll need to copy it over into your own code! 
https://gist.github.com/seand7565/6342d7ce7a89b2412fcb4eab7f2c8a61

This PR only fixes Italy, but lays down some structure to allow fixing 
other countries that need this treatment. For instance, if Spain needs 
to use the nested subregions instead of the top-level subregions, we 
just need to add Spains ISO code to `countries_that_use_nested_subregions` 
in `states.rb` and the ISO array on line 7 of `states.rake`. 

Ref: #3670 

**Checklist:**

### DIFF
--- a/core/db/default/spree/states.rb
+++ b/core/db/default/spree/states.rb
@@ -1,14 +1,22 @@
 # frozen_string_literal: true
 
+def create_states(subregions, country)
+  subregions.each do |subregion|
+    Spree::State.where(abbr: subregion.code, country: country).first_or_create!(
+      name: subregion.name
+    )
+  end
+end
+
 ActiveRecord::Base.transaction do
   Spree::Country.all.each do |country|
     carmen_country = Carmen::Country.coded(country.iso)
     next unless carmen_country.subregions?
 
-    carmen_country.subregions.each do |subregion|
-      Spree::State.where(abbr: subregion.code, country: country).first_or_create!(
-        name: subregion.name
-      )
+    if Spree::Config[:countries_that_use_nested_subregions].include? country.iso
+      create_states(carmen_country.subregions.flat_map(&:subregions), country)
+    else
+      create_states(carmen_country.subregions, country)
     end
   end
 end
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -248,6 +248,14 @@ module Spree
     #   (default: +['admin']+)
     preference :roles_for_auto_api_key, :array, default: ['admin']
 
+    # @!attribute [rw] countries_that_use_nested_subregions
+    #   @return [Array] An array of countries that use nested subregions, instead
+    #   of the default subregions that come with Carmen. Will be used on store creation
+    #   to ensure the correct states are generated, and when running the states
+    #   regenerate rake task.
+    #   (default: +['IT']+)
+    preference :countries_that_use_nested_subregions, :array, default: ['IT']
+
     # @!attribute [rw] run_order_validations_on_order_updater
     #   @return [Boolean] Whether to run validation when updating an order with the OrderUpdater
     preference :run_order_validations_on_order_updater, :boolean, default: false


core/lib/tasks/states.rake
```diff
@@ -0,0 +1,57 @@
...
+    unless state_names.include? state.name
+      Spree::State.where(abbr: state.code, country: country).first_or_create!(
+        name: state.name
+      )
+    end
+  end
+end
+
+namespace :states do
+  desc "Regenerates states in cases where previous Solidus verions generated incorrect states"
```
spaghetticode, line 37: small typo: verions -> versions

core/lib/tasks/states.rake
```diff
@@ -0,0 +1,57 @@
...
+  task regenerate: :environment do
+    Spree::Country.all.each do |country|
+      if carmen_states = get_carmen_states(country.iso)
+        if (state_names = country.states.pluck(:name)) != (carmen_names = carmen_states.map(&:name).sort)
```
spaghetticode, line 41: Should we sort `country.states.pluck(:name)` as well in order to make sure that the ordering on this side is always correct?

core/lib/tasks/states.rake
seand7565: Ah, thanks! :)

core/lib/tasks/states.rake
seand7565: `country.states` is sorted through the association, do I don't think this would be an issue, but let me know if I'm wrong about that! https://github.com/solidusio/solidus/blob/7c76993eaafd6224f5c2e98fff5e8a33c2fd71b3/core/app/models/spree/country.rb#L5

core/db/default/spree/states.rb
```diff
@@ -1,14 +1,24 @@
 # frozen_string_literal: true
 
+countries_that_use_nested_subregions = %w(IT)
```
kennyadsl, line 3: If we use this in two places it's probably better to move it into a shared space like a preference?

core/lib/tasks/states.rake
```diff
@@ -0,0 +1,57 @@
...
+def user_agrees(destroy_states, create_states, country)
+  puts "\e[31mStates for #{country.name} are incorrect. If you proceed, the following changes will happen:\e[0m"
+  puts "\e[31mThese states will be destroyed:\e[0m #{destroy_states}"
+  puts "\e[31mThese states will be created:\e[0m #{create_states}"
+  puts "\e[31mDo you want to proceed? [y/n]\e[0m "
+  ["y", "yes"].include? STDIN.gets.strip.downcase
+end
+
+def delete_states(country, carmen_names)
+  country.states.where.not(name: carmen_names).destroy_all
```
kennyadsl, line 23: I'm hesitant doing this since the addresses stored for old orders will lose the database reference to these states. I suppose we can't soft delete states, right?

kennyadsl (commented):
Left a couple of comments, thanks @seand7565

core/db/default/spree/states.rb
seand7565: Yeah, I couldn't really think of a good shared spot to put this that made sense. Preferences seems like a weird spot for it IMO, but maybe it's the place that makes the most sense!

core/lib/tasks/states.rake
seand7565: We can, AFAIK we'd just have to add a "discarded_at" index, as per the discard readme: https://github.com/jhawthorn/discard

I think that might be a good option, but if that's too much we can also just warn the user that the state was unable to be deleted, and let them handle it however they want.

core/db/default/spree/states.rb
seand7565: Yeah, preferences seems like the best place for it after review! I've added a preference in, how does that look?

core/lib/tasks/states.rake
kennyadsl: Honestly, I would not provide this rake task to users. It's too dangerous. I think that leaving this code in the PR's description as one of the possible migration paths (making clear that states will be destroyed and addresses that use those states will lose references) is enough for people that want to migrate their data after this change. I don't expect to be a lot since if they sell in those regions they probably already did the change. 

For the others that start a new Solidus store this is not an issue since they'll have their code updated

seand7565: Okay, I've moved it to a gist and added a link to the PR description.

jarednorman (approved):

aldesantis:
@kennyadsl if this good to go for you? Feel free to merge if yes!

kennyadsl (approved):

Issue #3604, opened 2020-05-03
Inconsistent handling of resource errors on admin


seand7565:
For resource errors where the parent resource doesn't exist, IMO we should be throwing a `404` and redirecting to the index for the parent. Instead, here's what happens, given a non-existent parent resource.

**products**
`/admin/products/nil-product/edit` - 404, redirect to /products ✅ 
`/admin/products/nil-product/stock` - 404, redirect to /stock_items ✅ 
`/admin/products/nil-product/images` - 500, undefined method `admin_images_url` - #3143  does a great job breaking this down. ❌ 
`/admin/products/nil-product/variants` - 500, undefined method `variants` for nil:NilClass ❌ 

**taxons**
`/admin/taxonomies/nil-taxon/taxons/3/edit` - 404, no redirect ❌ 
`/admin/taxonomies/1/taxons/nil-taxon/edit` - 404, no redirect ❌ 

**users**
`/admin/users/nil-user/edit` - 404, redirect to /users ✅ 
`/admin/users/nil-user/addresses` -404, redirect to /users ✅ 
`/admin/users/nil-user/orders` -404, redirect to /users ✅ 
`/admin/users/nil-user/items` -404, redirect to /users ✅ 
`/admin/users/nil-user/store_credits` - 500, undefined method `store_credits` for nil:NilClass ❌ 

As far as I can tell, all of the single resource routes (like `/shipping_method/nil/edit`) all work correctly - 404 with a redirect.

**Solidus Version**:
Master, though I think this has been a problem for a while.

**To Reproduce**
Visit any of the urls marked with an ❌ above on a fresh Solidus install

**Current behavior**
There's a mixture of 404s with redirects, 404s without redirects, and 500 errors that happen.

**Expected behavior**
Each of the URLs should 404 and redirect to parent resource

seand7565:
Can we close #3143 in favour of this issue?

seand7565:
This seems to have been handled by #3728 - thanks @ikraamg !

Issue #3143, opened 2019-03-14
NoMethodError: undefined method `admin_images_url`


edmunteanu:
When logged in as an admin, attempting to send requests to the `ImagesController` with an inexistent product doesn't redirect to the 404 page, but to an undefined `admin_images_url`, which throws an error.

**Solidus Version**:
2.6.3

**To Reproduce**
In order to reproduce this problem, simply follow this path: `admin/products/:inexistent_product/images`.

**Current behavior**
The actual implementation of the `ResourceController` rescues the `RecordNotFound` error with the `resource_not_found` method, which redirects you to the `collection_url`. This `collection_url` method then creates a polymorphic path out of the admin namespace and the resource, which creates the undefined `admin_images_url`.

**Expected behavior**
What I would expect to happen, when you send such a request, is a 404 page telling you that the product/page you are looking for doesn't exist.

**Desktop / Smartphone:**
This is not an issue which depends on a device / browser.

**Stacktrace:**
```
NoMethodError: undefined method `admin_images_url' for #<ActionDispatch::Routing::RoutesProxy:0x00007fb7a48c4f70>
Did you mean?  admin_product_images_url
  from action_dispatch/routing/routes_proxy.rb:47:in `method_missing'
  from action_dispatch/routing/polymorphic_routes.rb:230:in `polymorphic_method'
  from action_dispatch/routing/polymorphic_routes.rb:114:in `polymorphic_url'
  from spree/admin/resource_controller.rb:262:in `collection_url'
  from spree/admin/resource_controller.rb:130:in `resource_not_found'
  from active_support/rescuable.rb:107:in `call'
  from active_support/rescuable.rb:107:in `block in handler_for_rescue'
  from active_support/rescuable.rb:91:in `rescue_with_handler'
  from active_support/rescuable.rb:164:in `rescue_with_handler'
  from action_controller/metal/rescue.rb:23:in `rescue in process_action'
  from action_controller/metal/rescue.rb:19:in `process_action'
  from action_controller/metal/instrumentation.rb:32:in `block in process_action'
  from active_support/notifications.rb:166:in `block in instrument'
  from active_support/notifications/instrumenter.rb:21:in `instrument'
  from active_support/notifications.rb:166:in `instrument'
  from action_controller/metal/instrumentation.rb:30:in `process_action'
  from action_controller/metal/params_wrapper.rb:252:in `process_action'
  from active_record/railties/controller_runtime.rb:22:in `process_action'
[...]

PR #3728, opened 2020-08-10
Fixes for consistent handling of resource errors on admin 

ikraamg:
**Description**

Referring to the detailed examples in issues #3604 and #3143, there was inconsistency in the way errors were handled due to non-existent resources on the admin panel. The aim was to bring consistency to these errors by redirecting all errors to the relevant page with a relevant flash_message.

This PR fixes all the mentioned issues as well as additional inconsistencies and creates mulitple tests to ensure all the fixes are not affected in the future.

### Previously, these nil-resource routes were redirected correctly:
 **products**
 `/admin/products/nil-product/edit`
 `/admin/products/nil-product/stock`
**users**
 `/admin/users/nil-user/edit` 
 `/admin/users/nil-user/addresses` 
 `/admin/users/nil-user/orders`
 `/admin/users/nil-user/items` 
**promotions**
`/admin/promotion_categories/nil-category/edit`
`/admin/promotions/nil-promotion/edit`
**option types**
`/admin/option_types/nil-type/edit`
**property types**
`/admin/properties/nil-prop/edit`
**taxonomies**
`/admin/taxonomies/nil-taxonomy/edit`
  
### After this PR, the following are fixed and redirected correctly as well:
**orders**
`/admin/orders/nil-order/edit`
`/admin/orders/nil-order/customer_returns`
`/admin/orders/nil-order/customer_returns/1/edit`
`/admin/orders/nil-order/cart`
`/admin/orders/nil-order/customer/edit`
`/admin/orders/nil-order/adjustments`
`/admin/orders/nil-order/payments`
`/admin/orders/nil-order/return_authorizations`
`/admin/orders/nil-order/cancellations`
**products**
`/admin/products/nil-product/images`
`/admin/products/nil-product/prices`
`/admin/products/nil-product/product_properties`
`/admin/products/nil-product/variants`
 **taxons**
`/admin/taxonomies/nil-taxon/taxons/3/edit`
`/admin/taxonomies/1/taxons/nil-taxon/edit`
**users**
`/admin/users/nil-user/store_credits`

**Checklist:**
- [ ] ~~I have updated Guides and README accordingly to this change (if needed)~~
- [ ] ~~I have attached screenshots to this PR for visual changes (if needed)~~

### DIFF
--- a/backend/app/controllers/spree/admin/base_controller.rb
+++ b/backend/app/controllers/spree/admin/base_controller.rb
@@ -62,6 +62,12 @@ module Spree
       def order_mutex_redirect_path
         edit_admin_order_path(@order)
       end
+
+      def resource_not_found(flash_class:, redirect_url:)
+        flash[:error] = flash_message_for(flash_class.new, :not_found)
+        redirect_to redirect_url
+        nil
+      end
     end
   end
 end
--- a/backend/app/controllers/spree/admin/cancellations_controller.rb
+++ b/backend/app/controllers/spree/admin/cancellations_controller.rb
@@ -36,6 +36,8 @@ module Spree
       def load_order
         @order = Spree::Order.find_by!(number: params[:order_id])
         authorize! action, @order
+      rescue ActiveRecord::RecordNotFound
+        resource_not_found(flash_class: Spree::Order, redirect_url: admin_orders_path)
       end
 
       def model_class
--- a/backend/app/controllers/spree/admin/images_controller.rb
+++ b/backend/app/controllers/spree/admin/images_controller.rb
@@ -4,7 +4,6 @@ module Spree
   module Admin
     class ImagesController < ResourceController
       before_action :load_data
-
       create.before :set_viewable
       update.before :set_viewable
 
@@ -24,6 +23,8 @@ module Spree
           [variant.sku_and_options_text, variant.id]
         end
         @variants.insert(0, [t('spree.all'), @product.master.id])
+      rescue ActiveRecord::RecordNotFound
+        resource_not_found(flash_class: Spree::Product, redirect_url: admin_products_path)
       end
 
       def set_viewable
--- a/backend/app/controllers/spree/admin/orders_controller.rb
+++ b/backend/app/controllers/spree/admin/orders_controller.rb
@@ -10,7 +10,6 @@ module Spree
       around_action :lock_order, only: [:update, :advance, :complete, :confirm, :cancel, :resume, :approve, :resend]
 
       rescue_from Spree::Order::InsufficientStock, with: :insufficient_stock_error
-
       respond_to :html
 
       def index
@@ -168,6 +167,8 @@ module Spree
       def load_order
         @order = Spree::Order.includes(:adjustments).find_by!(number: params[:id])
         authorize! action, @order
+      rescue ActiveRecord::RecordNotFound
+        resource_not_found(flash_class: Spree::Order, redirect_url: admin_orders_path)
       end
 
       # Used for extensions which need to provide their own custom event links on the order details view.
--- a/backend/app/controllers/spree/admin/resource_controller.rb
+++ b/backend/app/controllers/spree/admin/resource_controller.rb
@@ -5,7 +5,9 @@ class Spree::Admin::ResourceController < Spree::Admin::BaseController
 
   helper_method :new_object_url, :edit_object_url, :object_url, :collection_url
   before_action :load_resource, except: :update_positions
-  rescue_from ActiveRecord::RecordNotFound, with: :resource_not_found
+  rescue_from ActiveRecord::RecordNotFound do
+    resource_not_found
+  end
   rescue_from ActiveRecord::RecordInvalid, with: :resource_invalid
 
   respond_to :html
@@ -134,9 +136,8 @@ class Spree::Admin::ResourceController < Spree::Admin::BaseController
     end
   end
 
-  def resource_not_found
-    flash[:error] = flash_message_for(model_class.new, :not_found)
-    redirect_to collection_url
+  def resource_not_found(flash_class: model_class, redirect_url: collection_url)
+    super
   end
 
   def model_class
@@ -193,12 +194,14 @@ class Spree::Admin::ResourceController < Spree::Admin::BaseController
     if parent?
       @parent ||= self.class.parent_data[:model_class]
                     .includes(self.class.parent_data[:includes])
-                    .find_by(self.class.parent_data[:find_by] => params["#{parent_model_name}_id"])
+                    .find_by!(self.class.parent_data[:find_by] => params["#{parent_model_name}_id"])
       instance_variable_set("@#{parent_model_name}", @parent)
     else
       Spree::Deprecation.warn "Calling #parent is deprecated on a ResourceController which has not defined a belongs_to"
       nil
     end
+  rescue ActiveRecord::RecordNotFound => e
+    resource_not_found(flash_class: e.model.constantize, redirect_url: spree.polymorphic_url([:admin, parent_model_name.pluralize]))
   end
 
   def parent?
@@ -222,7 +225,8 @@ class Spree::Admin::ResourceController < Spree::Admin::BaseController
   end
 
   def collection
-    return parent.send(controller_name) if parent?
+    return parent.send(controller_name) if parent? && parent
+
     if model_class.respond_to?(:accessible_by) && !current_ability.has_block?(params[:action], model_class)
       model_class.accessible_by(current_ability, action)
     else
--- a/backend/app/controllers/spree/admin/taxons_controller.rb
+++ b/backend/app/controllers/spree/admin/taxons_controller.rb
@@ -3,6 +3,7 @@
 module Spree
   module Admin
     class TaxonsController < Spree::Admin::BaseController
+      rescue_from ActiveRecord::RecordNotFound, with: :resource_not_found
       respond_to :html, :json, :js
 
       def index
@@ -82,6 +83,10 @@ module Spree
 
       def taxon_params
         params.require(:taxon).permit(permitted_taxon_attributes)
+      end
+
+      def resource_not_found
+        super(flash_class: Taxon, redirect_url: admin_taxonomies_path)
       end
     end
   end
--- a/backend/app/controllers/spree/admin/variants_controller.rb
+++ b/backend/app/controllers/spree/admin/variants_controller.rb
@@ -43,8 +43,10 @@ module Spree
       end
 
       def parent
-        @parent ||= Spree::Product.with_discarded.find_by(slug: params[:product_id])
+        @parent ||= Spree::Product.with_discarded.find_by!(slug: params[:product_id])
         @product = @parent
+      rescue ActiveRecord::RecordNotFound
+        resource_not_found(flash_class: Spree::Product, redirect_url: admin_products_path)
       end
     end
   end
--- a/backend/spec/controllers/spree/admin/cancellations_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/cancellations_controller_spec.rb
@@ -17,6 +17,14 @@ describe Spree::Admin::CancellationsController do
         expect(response).to be_ok
       end
     end
+
+    context "existent order id not given" do
+      it "redirects and flashes about the non-existent order" do
+        get :index, params: { order_id: 'non-existent-order' }
+        expect(response).to redirect_to(spree.admin_orders_path)
+        expect(flash[:error]).to eql("Order is not found")
+      end
+    end
   end
 
   describe "#cancel" do
--- a/backend/spec/controllers/spree/admin/customer_returns_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/customer_returns_controller_spec.rb
@@ -26,6 +26,14 @@ module Spree
         end
       end
 
+      context "existent order id not given" do
+        it "redirects and flashes about the non-existent order" do
+          get :index, params: { order_id: 'non-existent-order' }
+          expect(response).to redirect_to(spree.admin_orders_path)
+          expect(flash[:error]).to eql("Order is not found")
+        end
+      end
+
       describe "#new" do
         let(:order) { create(:shipped_order, line_items_count: 1) }
         let!(:inactive_reimbursement_type)      { create(:reimbursement_type, active: false) }
--- a/backend/spec/controllers/spree/admin/payments_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/payments_controller_spec.rb
@@ -152,6 +152,14 @@ module Spree
             expect(response).to redirect_to(spree.edit_admin_order_customer_path(order))
           end
         end
+
+        context "existent order id not given" do
+          it "redirects and flashes about the non-existent order" do
+            get :index, params: { order_id: 'non-existent-order' }
+            expect(response).to redirect_to(spree.admin_orders_path)
+            expect(flash[:error]).to eql("Order is not found")
+          end
+        end
       end
 
       describe '#fire' do
--- a/backend/spec/controllers/spree/admin/product_properties_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/product_properties_controller_spec.rb
@@ -64,6 +64,21 @@ module Spree
               expect(assigns(:variant_property_rule)).to eq first_rule
             end
           end
+
+          context "existent product id not given" do
+            let(:parameters) do
+              {
+                product_id: 'non-existent-product'
+              }
+            end
+
+            before { subject }
+
+            it "cannot find non-existent product" do
+              expect(response).to redirect_to(spree.admin_products_path)
+              expect(flash[:error]).to eql("Product is not found")
+            end
+          end
         end
       end
     end
--- a/backend/spec/controllers/spree/admin/store_credits_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/store_credits_controller_spec.rb
@@ -308,4 +308,12 @@ describe Spree::Admin::StoreCreditsController do
       end
     end
   end
+  context 'User does not exist' do
+    let(:store_credit) { create(:store_credit, user: user, category: a_credit_category) }
+    it "cannot find a store-credit for non-existent user" do
+      get :index, params: { user_id: 'non-existent-user', id: store_credit.id }
+      expect(response).to redirect_to(spree.admin_users_path)
+      expect(flash[:error]).to eql("User is not found")
+    end
+  end
 end
--- a/backend/app/controllers/spree/admin/customer_returns_controller.rb
+++ b/backend/app/controllers/spree/admin/customer_returns_controller.rb
@@ -42,6 +42,8 @@ module Spree
 
       def collection
         parent # trigger loading the order
+        return unless @order
+
         @collection ||= Spree::ReturnItem
           .accessible_by(current_ability, :read)
           .where(inventory_unit_id: @order.inventory_units.pluck(:id))
--- a/backend/app/controllers/spree/admin/orders/customer_details_controller.rb
+++ b/backend/app/controllers/spree/admin/orders/customer_details_controller.rb
@@ -61,6 +61,8 @@ module Spree
 
         def load_order
           @order = Spree::Order.includes(:adjustments).find_by!(number: params[:order_id])
+        rescue ActiveRecord::RecordNotFound
+          resource_not_found(flash_class: Spree::Order, redirect_url: admin_orders_path)
         end
 
         def model_class
--- a/backend/app/controllers/spree/admin/payments_controller.rb
+++ b/backend/app/controllers/spree/admin/payments_controller.rb
@@ -92,6 +92,8 @@ module Spree
         @order = Spree::Order.find_by!(number: params[:order_id])
         authorize! action, @order
         @order
+      rescue ActiveRecord::RecordNotFound
+        resource_not_found(flash_class: Spree::Order, redirect_url: admin_orders_path)
       end
 
       def load_payment
--- a/backend/spec/controllers/spree/admin/images_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/images_controller_spec.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Spree::Admin::ImagesController, type: :controller do
+  stub_authorization!
+
+  it "cannot find a non-existent product" do
+    get :index, params: { product_id: "non-existent-product" }
+    expect(response).to redirect_to(spree.admin_products_path)
+    expect(flash[:error]).to eql("Product is not found")
+  end
+end
--- a/backend/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/orders/customer_details_controller_spec.rb
@@ -53,6 +53,13 @@ describe Spree::Admin::Orders::CustomerDetailsController, type: :controller do
           end
         end
       end
+      context "existent order id not given" do
+        it "redirects and flashes about the non-existent order" do
+          get :edit, params: { order_id: 'non-existent-order' }
+          expect(response).to redirect_to(spree.admin_orders_path)
+          expect(flash[:error]).to eql("Order is not found")
+        end
+      end
     end
 
     context "#update" do
--- a/backend/spec/controllers/spree/admin/orders_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/orders_controller_spec.rb
@@ -396,13 +396,13 @@ describe Spree::Admin::OrdersController, type: :controller do
     end
   end
 
-  context "order number not given" do
+  context "existent order number not given" do
     stub_authorization!
 
-    it "raise active record not found" do
-      expect {
-        get :edit, params: { id: 0 }
-      }.to raise_error ActiveRecord::RecordNotFound
+    it "cannot find non-existent order" do
+      get :edit, params: { id: 0 }
+      expect(response).to redirect_to(spree.admin_orders_path)
+      expect(flash[:error]).to eql("Order is not found")
     end
   end
 end
--- a/backend/spec/controllers/spree/admin/prices_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/prices_controller_spec.rb
@@ -41,5 +41,14 @@ describe Spree::Admin::PricesController do
         expect(assigns(:product)).to eq(product)
       end
     end
+    context "existent product id not given" do
+      subject { get :index, params: { product_id: 'non-existent-product' } }
+
+      it "cannot find non-existent product" do
+        subject
+        expect(response).to redirect_to(spree.admin_products_path)
+        expect(flash[:error]).to eql("Product is not found")
+      end
+    end
   end
 end
--- a/backend/spec/controllers/spree/admin/taxons_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/taxons_controller_spec.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Spree::Admin::TaxonsController, type: :controller do
+  stub_authorization!
+
+  describe "#edit" do
+    let!(:taxon) { create(:taxon) }
+
+    it "finds existing taxon" do
+      get :edit, params: { taxonomy_id: taxon.taxonomy, id: taxon.id }
+      expect(response.status).to eq(200)
+    end
+
+    it "cannot find a non-existing taxon with existent taxonomy" do
+      get :edit, params: { taxonomy_id: taxon.taxonomy, id: 'non-existent-taxon' }
+      expect(response).to redirect_to(spree.admin_taxonomies_path)
+      expect(flash[:error]).to eql("Taxon is not found")
+    end
+
+    it "cannot find a existing taxon with non-existent taxonomy" do
+      get :edit, params: { taxonomy_id: 'non-existent-taxonomy', id: taxon.id }
+      expect(response).to redirect_to(spree.admin_taxonomies_path)
+      expect(flash[:error]).to eql("Taxon is not found")
+    end
+  end
+end
--- a/backend/spec/controllers/spree/admin/variants_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/variants_controller_spec.rb
@@ -53,6 +53,15 @@ module Spree
                 expect(assigns(:collection)).to include deleted_variant
               end
             end
+            context "existent product id not given" do
+              let(:params) { { product_id: 'non-existent-product' } }
+
+              it "cannot find non-existent product" do
+                subject
+                expect(response).to redirect_to(spree.admin_products_path)
+                expect(flash[:error]).to eql("Product is not found")
+              end
+            end
           end
         end
       end


ikraamg:
Hi Solidus Team!

This is my first contribution to solidus, I would be glad to receive any feedback for improving the PR. 

I have a few questions:

- The bin/build did pass all tests locally but with a stream of deprecation warnings, is this expected? (I see that the CircleCI tests are failing as well due to what seems to be the deprecation warnings) 

- The warning `Calling #parent is deprecated on a ResourceController which has not defined a belongs_to` was the most common, should I be concerned about this?

- I realised after making these commits that it may have been useful in cases where there was no 404 error raised, to manually `raise ActionController::RoutingError.new('Not Found')` or something similar to get a 404. Then 'rescue' before redirecting to the correct 'path'. Do you advise I should refactor the code to do this or is that unnecessary since the code works as expected?

Thanks,
Ikraam

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -222,7 +222,8 @@ def build_resource
   end
 
   def collection
-    return parent.send(controller_name) if parent?
+    return parent.send(controller_name) unless parent.nil?
```
jarednorman, line 234: I feel like this _shouldn't_ happen. In what situations might it? Would a better solution be to change the implementation of `#parent?`?

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -134,7 +134,16 @@ def belongs_to(model_name, options = {})
...
 
+  def product_exists?
+    return unless parent.nil?
+
+    flash[:error] = flash_message_for(Spree::Product.new, :not_found)
+    redirect_to admin_products_path
+  end
+
   def resource_not_found
+    return product_exists? if model_class == Spree::Image
```
jarednorman, line 145: This seems like we should be making these changes to `Spree::Admin::ImagesController` rather than making a conditional like this in the generic base resource controller.

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -134,7 +134,16 @@ def belongs_to(model_name, options = {})
     end
   end
 
+  def product_exists?
```
jarednorman, line 137: This method only makes sense if the parent is a product, but the logic could be generalized to handle any kind of parent, right?

backend/app/controllers/spree/admin/taxons_controller.rb
```diff
@@ -83,6 +84,11 @@ def destroy
       def taxon_params
         params.require(:taxon).permit(permitted_taxon_attributes)
       end
+
+      def resource_not_found
+        flash[:error] = flash_message_for(Taxon.new, :not_found)
+        redirect_to admin_taxonomies_path
+      end
```
jarednorman, line 91: Seems like we could generalize this logic instead of duplicating it across controllers?

backend/app/controllers/spree/admin/store_credits_controller.rb
```diff
@@ -74,7 +74,15 @@ def permitted_resource_params
           merge(created_by: try_spree_current_user)
       end
 
+      def user_exists?
+        return unless parent.nil?
+
+        flash[:error] = flash_message_for(model_class.new, :not_found)
+        redirect_to admin_users_path
```
jarednorman, line 81: If we generalized this not found logic for parent classes we wouldn't need these specialized implementations that vary only in the path.

backend/app/controllers/spree/admin/resource_controller.rb
ikraamg: Thanks for this suggestion, I refactored to generalize the flashes and redirects into the base_controller :+1:

ikraamg:
> Thanks for contributing! This PR adds a lot of special cases that I think could be better handled by generalized implementations that are part of `Spree::Admin::ResourceController`. I'd like to make this stuff more consistent by using the same code everywhere, not by adding more special cases when many of them do fundamentally the same thing.
> 
> Side note: We like to follow [these conventions](https://beams.io/posts/git-commit/) for git commit formatting. Your commits offer good details and are definitely nothing to complain about, but if you happened to update them to follow the wrapping and summary line conventions, it would be appreciated.

Thanks for the suggestions and pointers, I have made an attempt to refactor all the code into the base_controller. Kindly let me know if this can be further improved.

I am still looking into the    ` return parent.send(controller_name) if parent?` vs 
`return parent.send(controller_name) unless parent.nil?`. I will mention my findings inside the code conversation above. Using parent.nil? also seems to be the reason why there are so many deprecation warnings in CircleCI.

I also trust the latest git commit is formatted as per the conventions :smile:

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -134,9 +134,15 @@ def belongs_to(model_name, options = {})
     end
   end
 
+  def nil_parent?(options = {})
```
jarednorman, line 137: This is confusing because the convention in Ruby is that methods that end in question marks (known as predicate methods) return boolean values.

backend/spec/controllers/spree/admin/product_properties_controller_spec.rb
```diff
@@ -64,6 +64,21 @@ module Admin
               expect(assigns(:variant_property_rule)).to eq first_rule
             end
           end
+
+          context "existent product id not given" do
+            let(:parameters) do
+              {
+                  product_id: 'non-existent-product'
```
jarednorman, line 71: Too many indents on this line.

backend/spec/controllers/spree/admin/store_credits_controller_spec.rb
```diff
@@ -308,4 +308,12 @@
       end
     end
   end
+  context 'User does not exist' do
+    let!(:store_credit) { create(:store_credit, user: user, category: a_credit_category) }
```
jarednorman, line 312: I think this can just a be a `let` rather than a `let!`.

backend/app/controllers/spree/admin/store_credits_controller.rb
```diff
@@ -74,7 +74,15 @@ def permitted_resource_params
           merge(created_by: try_spree_current_user)
       end
 
+      def user_exists?
```
jarednorman, line 77: I don't think we still need this method?

backend/app/controllers/spree/admin/base_controller.rb
```diff
@@ -62,6 +62,14 @@ def lock_order
       def order_mutex_redirect_path
         edit_admin_order_path(@order)
       end
+
+      def resource_not_found(options = {})
+        options[:custom_class] ||= model_class
+        options[:custom_redirect] ||= admin_url
+        flash[:error] = flash_message_for(options[:custom_class].new, :not_found)
+        redirect_to options[:custom_redirect]
+      end
```
jarednorman, line 71: Refactored to use keyword arguments. I also don't think `custom_class` is a very good name for this argument because it doesn't tell you anything about what the argument is to be used for. I'm not sure what a better name would be though.

```suggestion
      def resource_not_found(custom_class: model_class, redirect_url: admin_url)
        flash[:error] = flash_message_for(custom_class.new, :not_found)
        redirect_to redirect_url
      end
```

ikraamg:
Thanks for the detailed review, I feel a huge difference in how I am able to approach a solution after each review :smile: 

In reply to some of your notes:

> 1. `ActiveRecord::RecordNotFound` errors expose the specific model that was not found. This can be used to infer whether it was the parent or the main model that was not found and generate a reasonable default redirect path. It's possible this might be able to cover all the cases, right?

The`ActiveRecord::RecordNotFound` error seems to only occur for the 'orders' and 'taxonomies' controller, they both inherit from the 'base_controller'. 
Most of the other errors are 'NoMethodErrors' that occur in the first return statement of the resources#collection method. Though, I think the principle of this approach is the same as I am rescuing errors from both types and redirecting them either via resource_not_found and resource_error_filter.


> 2. I don't think that `parent` should ever be `nil` when `parent?` is true. I'm still missing why/how this might be possible and why we need to handle it.

I have refactored the code to avoid having to deal with this but I can give an overview of what I found.

It seems that for particularly these links the issue occurs:
```
/admin/products/nil-product/prices
/admin/products/nil-product/product_properties
/admin/products/nil-product/variants
/admin/users/nil-user/store_credits

```
The issue seems to occur because `#parent?` checks to see if the `#parent_data` exists and it does in these cases, so it returns `true`:
```
def parent?
    self.class.parent_data.present?
end
```
But the confusion comes in because of `#parent` since `parent?` is `true` for these links and it is supposed to run the query that I have not wrapped my head around. I am assuming it checks if `#belongs_to` was defined on 'children' of `ResourceController`. But then it returns `nil` for some reason, creating the `NoMethodError for nil:class` in the #collection method:

`return parent.send(controller_name) if parent?`

I have posted an image of this scenario in the PR description tested with the BetterErrors console.

```
def parent
    if parent?
      @parent ||= self.class.parent_data[:model_class]
                    .includes(self.class.parent_data[:includes])
                    .find_by(self.class.parent_data[:find_by] => params["#{parent_model_name}_id"])
      instance_variable_set("@#{parent_model_name}", @parent)
    else
      Spree::Deprecation.warn "Calling #parent is deprecated on a ResourceController which has not defined a belongs_to"
      nil
    end
  end
```


> Also, forgive me if I'm being picky about how we might change classes like `Spree::Admin::ResourceController`. Because stores and extensions add customizations that depend on these classes (the names of the methods in them, the options those methods take, etc.) we must be very careful that changes here are exactly what we want them to be to avoid creating unnecessary churn and work for stores trying to upgrade.

That makes complete sense. I assumed that it would challenge to meet the standards required in order to contribute, so thank you for your time in guiding me through this.

I will push the changes I have made with a rebase soon, I am getting familiar with rebasing so this might take a short while.

backend/app/controllers/spree/admin/base_controller.rb
ikraamg: This makes much more sense, I will make the change. I tried to do this with the `#resourse_error_filter` (new name for the `nil_parent?`) but since I had to pass the options into `resource_not_found`, i found it difficult to accomplish. Perhaps you can have a look once I push the changes.

backend/app/controllers/spree/admin/store_credits_controller.rb
ikraamg: Yeah, I indeed forgot to remove it.

backend/app/controllers/spree/admin/resource_controller.rb
ikraamg: Ah I see, it definitely needed to be changed then. It's changed to `resource_error_filter` now

backend/app/controllers/spree/admin/base_controller.rb
```diff
@@ -62,6 +62,12 @@ def lock_order
       def order_mutex_redirect_path
         edit_admin_order_path(@order)
       end
+
+      def resource_not_found(flash_class: model_class, redirect_url: admin_url)
```
jarednorman, line 66: I think we could remove the default values, see my review comment.

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -134,9 +134,14 @@ def belongs_to(model_name, options = {})
     end
   end
 
+  def resource_error_filter(options = {})
+    options[:flash_class] ||= "Spree::#{parent_model_name.classify}".constantize
+    options[:redirect_url] ||= spree.new_polymorphic_url([:admin, parent_model_name])
+    base_resource_not_found(options)
+  end
```
jarednorman, line 141: I think if we capture the exceptions specifically in the methods that would raise them, we can make this method unnecessary.

ikraamg:
So I have finally made some time to have a go at it again and I hope we are really close now.
The suggestions really helped. 
There were a couple of challenges that I encountered and I had to make some other changes to get the fixes to work. I wrote up the detailed context in the commit message.

Cheers!

backend/app/controllers/spree/admin/base_controller.rb
```diff
@@ -62,6 +62,11 @@ def lock_order
       def order_mutex_redirect_path
         edit_admin_order_path(@order)
       end
+
+      def resource_not_found(flash_class:, redirect_url:)
+        flash[:error] = flash_message_for(flash_class.new, :not_found)
+        redirect_to redirect_url and return
```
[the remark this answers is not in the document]
jarednorman: This `and return` doesn't actually do anything here.

backend/app/controllers/spree/admin/variants_controller.rb
```diff
@@ -43,7 +43,11 @@ def redirect_on_empty_option_values
       end
 
       def parent
-        @parent ||= Spree::Product.with_discarded.find_by(slug: params[:product_id])
+        begin
+          @parent ||= Spree::Product.with_deleted.find_by!(slug: params[:product_id])
+        rescue ActiveRecord::RecordNotFound => e
+          resource_not_found(e, flash_class: Spree::Product, redirect_url: admin_products_path)
+        end
```
jarednorman, line 50: You can rescue this at the method level:

```ruby
def parent
  @parent ||= Spree::Product.with_deleted.find_by!(slug: params[:product_id])
  @product = @parent
rescue ActiveRecord::RecordNotFound => e
  resource_not_found(e, flash_class: Spree::Product, redirect_url: admin_products_path)
end

File: backend/app/controllers/spree/admin/taxons_controller.rb [hunk 15]
```diff
@@ -83,6 +84,10 @@ def destroy
       def taxon_params
         params.require(:taxon).permit(permitted_taxon_attributes)
       end
+
+      def resource_not_found
+        super({ redirect_url: admin_taxonomies_path, flash_class: Taxon })
```
jarednorman, line 89: You can remove the curly braces here. It doesn't make much sense to include them especially since we're using keyword args.

backend/app/controllers/spree/admin/orders_controller.rb
```diff
@@ -166,7 +165,11 @@ def order_params
       end
 
       def load_order
-        @order = Spree::Order.includes(:adjustments).find_by!(number: params[:id])
+        begin
+          @order = Spree::Order.includes(:adjustments).find_by!(number: params[:id])
+        rescue ActiveRecord::RecordNotFound
+          resource_not_found(flash_class: Spree::Order, redirect_url: admin_orders_path)
+        end
```
jarednorman, line 172: Similarly to my other comment, we can avoid unnecessary git churn by rescuing this at the method level:

```ruby
def load_order
  @order = Spree::Order.includes(:adjustments).find_by!(number: params[:id])
  authorize! action, @order
rescue ActiveRecord::RecordNotFound
  resource_not_found(flash_class: Spree::Order, redirect_url: admin_orders_path)
end
```

backend/app/controllers/spree/admin/images_controller.rb
```diff
@@ -19,7 +18,12 @@ def location_after_save
       end
 
       def load_data
-        @product = Spree::Product.friendly.find(params[:product_id])
+        begin
+          @product = Spree::Product.friendly.find(params[:product_id])
+        rescue ActiveRecord::RecordNotFound => e
+          resource_not_found(e, flash_class: Spree::Product, redirect_url: admin_products_path)
```
jarednorman, line 24: Why are you passing the error through if you are not using it?

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -134,9 +134,10 @@ def belongs_to(model_name, options = {})
     end
   end
 
-  def resource_not_found
-    flash[:error] = flash_message_for(model_class.new, :not_found)
-    redirect_to collection_url
+  def resource_not_found(_err, options = {})
+    options[:flash_class] ||= model_class
+    options[:redirect_url] ||= collection_url
+    super(options)
```
jarednorman, line 140: Since you're not using the error, you can remove that argument, continue using keyword arguments, add default values, and rely on the fact that super passes through the arguments implicitly:

```suggestion
  def resource_not_found(flash_class: model_class, redirect_url: collection_url)
    super
```

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -191,10 +192,14 @@ def parent_data
...
-      @parent ||= self.class.parent_data[:model_class]
-                    .includes(self.class.parent_data[:includes])
-                    .find_by(self.class.parent_data[:find_by] => params["#{parent_model_name}_id"])
-      instance_variable_set("@#{parent_model_name}", @parent)
+      begin
+        @parent ||= self.class.parent_data[:model_class]
+                        .includes(self.class.parent_data[:includes])
+                        .find_by!(self.class.parent_data[:find_by] => params["#{parent_model_name}_id"])
+        instance_variable_set("@#{parent_model_name}", @parent)
+      rescue ActiveRecord::RecordNotFound => e
```
jarednorman, line 200: For consistency I'd also pull this rescue to the method level. If this method was more complex I'd want to localize the rescue, but I think it's fine to prefer avoiding the git churn. It should be obvious to the reader where the exception might come from.

backend/app/controllers/spree/admin/variants_controller.rb
ikraamg: Very cool! Did not know this was possible.

backend/app/controllers/spree/admin/resource_controller.rb
ikraamg: This looks so much neater, thank you. I was so stuck on passing arguments into super.

If I don't have the _err argument then for almost all previously working redirects, including the images, prices and product_properties controllers I am getting an 
`ArgumentError at /admin/products/ruby-asdf/product_properties
wrong number of arguments (given 1, expected 0)`

I believe that this is due to an error message is being passed to the #resource_not_found method by the default rescue method.
 
Is there a more elegant solution than putting `_err = nil` in the argument list for both the base and resource controllers? The other option I imagine is a `*` as the first argument in the methods for both controllers.

backend/app/controllers/spree/admin/resource_controller.rb
ikraamg: Will do, thanks for the explanation :+1:

backend/app/controllers/spree/admin/base_controller.rb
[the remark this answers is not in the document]
ikraamg: If I don't `and return` it creates a 
`AbstractController::DoubleRenderError at /admin/products/ruby-54654/images
Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".` 
for all the redirect fixes we made in this commit.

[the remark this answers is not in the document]
ikraamg: It seems if I just return nil before the end of the method, it achieves the same purpose and Rubocop doesn't complain.

ikraamg:
I hope my explanation of using `_err` is understandable after the reply to your comment above. I ended up using the splat argument as it seemed more concise than `_error = nil`. 

I found some other 'Order' URL's that also created 404's or 500's. I've updated the list in the PR description and added fixes to the code in the same style that we discussed. I am quite sure it's all covered now :100: 

Cheers!

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -134,9 +134,8 @@ def belongs_to(model_name, options = {})
     end
   end
 
-  def resource_not_found
-    flash[:error] = flash_message_for(model_class.new, :not_found)
-    redirect_to collection_url
+  def resource_not_found(*, flash_class: model_class, redirect_url: collection_url)
```
jarednorman, line 137: I'm okay with `_error = nil` here. I had not thought about the fact that `rescue_from` will pass the error to the method if it takes arguments.

backend/app/controllers/spree/admin/base_controller.rb
```diff
@@ -62,6 +62,12 @@ def lock_order
       def order_mutex_redirect_path
         edit_admin_order_path(@order)
       end
+
+      def resource_not_found(*, flash_class:, redirect_url:)
+        flash[:error] = flash_message_for(flash_class.new, :not_found)
+        redirect_to redirect_url
+        nil
```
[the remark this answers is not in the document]
ikraamg: If I don't use the `and return` it creates a 
`AbstractController::DoubleRenderError at /admin/products/ruby-54654/images
Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".` 
for all the redirect fixes we made in this commit.

If I replace `and return` with a `nil` return at the end of the method it also achieves the purpose of avoiding the 'DoubleRender' and Rubocop doesn't complain either.

I am not exactly sure what is the cause for the double render.

backend/app/controllers/spree/admin/resource_controller.rb
ikraamg: Ah, I was unaware that `rescue_from` does that, it was quite confusing until I realized what was happening.
I will replace the 'splat argument' with `_error=nil` then :+1:

backend/app/controllers/spree/admin/base_controller.rb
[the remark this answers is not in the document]
jarednorman: Oh, I see. If you redirect in a regular `before_action` or something it stops subsequent things from happening but in these error handlers you have to explicitly return a falsy value. Got it! Thanks for clarifying.

ikraamg:
> Alright, I think I'm happy with this change now. I don't think it'll cause any serious incompatibility issues for this code, but we'll have to see if someone else from core has any thoughts oon that.

Perfect. Thanks Jared 👍

backend/app/controllers/spree/admin/cancellations_controller.rb
```diff
@@ -36,6 +36,8 @@ def created_by
       def load_order
         @order = Spree::Order.find_by!(number: params[:order_id])
         authorize! action, @order
+      rescue ActiveRecord::RecordNotFound => e
+        resource_not_found(flash_class: e.model.constantize, redirect_url: admin_orders_path)
```
aldesantis, line 40: Why are you not passing the error to `resource_not_found` here? I saw the commit message but it's still not entirely clear.

backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -222,7 +223,8 @@ def build_resource
   end
 
   def collection
-    return parent.send(controller_name) if parent?
+    return parent.send(controller_name) if parent? && parent
```
aldesantis, line 226: Correct me if I'm wrong, but the `&&` parent shouldn't be necessary here, since `parent` will either return a record or throw an exception.

aldesantis (changes requested):
Thanks for the contribution @ikraamg, I left a couple of comments.

backend/app/controllers/spree/admin/cancellations_controller.rb
ikraamg: If I understand your question correctly, you asking why I did not pass `e` into the `_error` of the resource_not_found of the base_controller?
I thought it is not necessary as I am not using the `_error` in the base_controller#resource_not_found. The `_error` is needed because some controllers use the `rescue_from` method which passes the exception object to the handler if it accepts an argument. This results in an `ArgumentError ... wrong number of arguments (given 1, expected 0)` if `_error` is not there.

On that note, I overlooked the fact that I do not even need the `e` at all in this cancellations_controller rescue. I should just pass `Spree::Orders` instead of unnecessarily passing `e.model.contantize` for the flash_class, since `e.model.constantise` will always return `Spree::Order`. Oddly enough, I seem to have appropriately done this in the other controller rescue methods but overlooked this one. I will make the change.

backend/app/controllers/spree/admin/resource_controller.rb
ikraamg: I am not sure how else to correct this ... I have followed the code execution to see that the `record_not_found` is thrown, rescued and the redirect is called in the base_controller#resource_not_found as expected.
However, confusingly, the `parent.send(controller_name)` is still called and then the `No_method_error` caused by `undefined method ... for nil:NilClass` occurs again because parent is `nil` in `parent.send(controller_name)`.

This is further baffling since returning the falsy value `nil` (or using [`and return`](https://guides.rubyonrails.org/layouts_and_rendering.html#avoiding-double-render-errors)) after a redirect, as done in the base_controller#resource_not_found, is meant to stop all subsequent actions from executing (it is successfully avoiding "double render errors" already).


The following redirect tests fail as a result of removing `&& parent`:
```
rspec ./spec/controllers/spree/admin/prices_controller_spec.rb:47 # Spree::Admin::PricesController#index existent product id not given cannot find non-existent product
rspec ./spec/controllers/spree/admin/variants_controller_spec.rb:59 # Spree::Admin::VariantsController#index the value of @collection with soft-deleted variants existent product id not given cannot find non-existent product
rspec ./spec/controllers/spree/admin/product_properties_controller_spec.rb:77 # Spree::Admin::ProductPropertiesController#index option values are provided existent product id not given cannot find non-existent product
rspec ./spec/controllers/spree/admin/store_credits_controller_spec.rb:313 # Spree::Admin::StoreCreditsController User does not exist cannot find a store-credit for non-existent user
```

backend/app/controllers/spree/admin/cancellations_controller.rb
aldesantis: @ikraamg how about this, then?
```ruby
rescue_from ActiveRecord::RecordNotFound do
  resource_not_found
end
```
It looks weird to have a method parameter that is not used at all in the method.

backend/app/controllers/spree/admin/resource_controller.rb
aldesantis: @ikraamg have you checked whether `parent` is being overridden anywhere? Perhaps there's an override that still uses `find_by`, causing the method to return `nil`.

backend/app/controllers/spree/admin/cancellations_controller.rb
ikraamg: To keep the format similar to how the rescue is done in the other controllers, I thought this would work for line 39 and 40:

```ruby
rescue ActiveRecord::RecordNotFound
  resource_not_found(flash_class: Spree::Order, redirect_url: admin_orders_path)
```

backend/app/controllers/spree/admin/resource_controller.rb
ikraamg: Out of the above-mentioned controllers causing this `no_method_error`, the variants_controller.rb is the only one with `parent` that is overridden, and this commit has already changed the `find_by` to `find_by!` in the variants controller and then rescues appropriately. 

I have placed breakpoints to follow the execution and in all cases, the `find_by!` does work but after the redirect, the `parent.send(controller_name)` is still called:
> ... I have followed the code execution to see that the `record_not_found` is thrown, rescued and the redirect is called in the base_controller#resource_not_found as expected.
> However, confusingly, the `parent.send(controller_name)` is still called and then the `No_method_error` caused by `undefined method ... for nil:NilClass` occurs again because parent is `nil` in `parent.send(controller_name)`.
> 
> This is further baffling since returning the falsy value `nil` (or using [`and return`](https://guides.rubyonrails.org/layouts_and_rendering.html#avoiding-double-render-errors)) after a redirect, as done in the base_controller#resource_not_found, is meant to stop all subsequent actions from executing (it is successfully avoiding "double render errors" already).

backend/app/controllers/spree/admin/taxons_controller.rb
```diff
@@ -3,6 +3,7 @@
 module Spree
   module Admin
     class TaxonsController < Spree::Admin::BaseController
+      rescue_from ActiveRecord::RecordNotFound, with: :resource_not_found
```
jarednorman, line 6: @ikraamg I believe the suggestion @aldesantis was making is to modify the calls to `rescue_from` (like this one) to avoid triggering the passing of the exception object by calling it like this:

```suggestion
      rescue_from ActiveRecord::RecordNotFound do
        resource_not_found
      end
```

This way we do not need the unused optional exception argument.

backend/app/controllers/spree/admin/resource_controller.rb
kennyadsl: Maybe there's another `rescue_from ActiveRecord::RecordNotFound` somewhere?

backend/spec/controllers/spree/admin/cancellations_controller_spec.rb
```diff
@@ -17,6 +17,14 @@
         expect(response).to be_ok
       end
     end
+
+    context "existent product id not given" do
```
kennyadsl, line 21: This sould be `order id`, right?

backend/spec/controllers/spree/admin/cancellations_controller_spec.rb
```diff
@@ -17,6 +17,14 @@
...
+      it "redirects and flashes about the non-existent order" do
+        get :index, params: { order_id: 'non-existent-product' }
```
kennyadsl, line 23: `non-existent-order`

backend/spec/controllers/spree/admin/customer_returns_controller_spec.rb
```diff
@@ -26,6 +26,14 @@ module Admin
         end
       end
 
+      context "existent product id not given" do
```
kennyadsl, line 29: Same here

kennyadsl (commented):
Left some comments, thanks for the great work here!

backend/app/controllers/spree/admin/taxons_controller.rb
ikraamg: @jarednorman thanks for clarifying the suggestion, I completely misunderstood! 
This worked great @aldesantis ! I agree that the unused argument was an oddity.

backend/spec/controllers/spree/admin/payments_controller_spec.rb
```diff
@@ -152,6 +152,14 @@ module Admin
             expect(response).to redirect_to(spree.edit_admin_order_customer_path(order))
           end
         end
+
+        context "existent product id not given" do
```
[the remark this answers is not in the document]
ikraamg: Thanks, @kennyadsl for catching these mistakes. They are all supposed to be `order` yes.

backend/app/controllers/spree/admin/resource_controller.rb
ikraamg: Hi @kennyadsl,  I could not find another `rescue_from ActiveRecord::RecordNotFound` in the controllers or the controllers' parents in which the issue occurs.

Perhaps you may be able to see where the issue lies if I share a step-through of the code below. From what I see, it redirects successfully and then continues executing by attempting to find the `controller_name` which is called by the `return parent.send(controller_name)`, and immediately thereafter the [error](https://user-images.githubusercontent.com/34813339/91338813-c78a4780-e7d5-11ea-9d29-81080b597f13.png) is raised.

```Ruby
Started GET "/admin/products/ruby-5646/prices" for ::1 at 2020-08-26 19:37:29 +0200
   (1.9ms)  SELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC
Processing by Spree::Admin::PricesController#index as HTML
  Parameters: {"product_id"=>"ruby-5646"}
  Spree::User Load (1.1ms)  SELECT "spree_users".* FROM "spree_users" WHERE "spree_users"."deleted_at" IS NULL AND "spree_users"."id" = $1 ORDER BY "spree_users"."id" ASC LIMIT $2  [["id", 1], ["LIMIT", 1]]
  Spree::Role Load (1.6ms)  SELECT "spree_roles".* FROM "spree_roles" INNER JOIN "spree_roles_users" ON "spree_roles"."id" = "spree_roles_users"."role_id" WHERE "spree_roles_users"."user_id" = $1  [["user_id", 1]]
/home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/awesome_nested_set-3.2.1/lib/awesome_nested_set/awesome_nested_set.rb:105: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/associations.rb:1657: warning: The called method `belongs_to' is defined here
/home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/awesome_nested_set-3.2.1/lib/awesome_nested_set/awesome_nested_set.rb:90: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/associations.rb:1370: warning: The called method `has_many' is defined here

[227, 236] in /home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/solidus-3dc450477ae5/backend/app/controllers/spree/admin/resource_controller.rb
   227:     end
   228:   end
   229: 
   230:   def collection
   231:     byebug
=> 232:     return parent.send(controller_name) if parent?
   233: 
   234:     if model_class.respond_to?(:accessible_by) && !current_ability.has_block?(params[:action], model_class)
   235:       model_class.accessible_by(current_ability, action)
   236:     else
(byebug) next

[191, 200] in /home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/solidus-3dc450477ae5/backend/app/controllers/spree/admin/resource_controller.rb
   191:     self.class.parent_data
   192:   end
   193: 
   194:   def parent
   195:     byebug
=> 196:     if parent?
   197:       @parent ||= self.class.parent_data[:model_class]
   198:                     .includes(self.class.parent_data[:includes])
   199:                     .find_by!(self.class.parent_data[:find_by] => params["#{parent_model_name}_id"])
   200:       instance_variable_set("@#{parent_model_name}", @parent)
(byebug) next

[192, 201] in /home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/solidus-3dc450477ae5/backend/app/controllers/spree/admin/resource_controller.rb
   192:   end
   193: 
   194:   def parent
   195:     byebug
   196:     if parent?
=> 197:       @parent ||= self.class.parent_data[:model_class]
   198:                     .includes(self.class.parent_data[:includes])
   199:                     .find_by!(self.class.parent_data[:find_by] => params["#{parent_model_name}_id"])
   200:       instance_variable_set("@#{parent_model_name}", @parent)
   201:     else
(byebug) next
  Spree::Product Load (1.7ms)  SELECT "spree_products".* FROM "spree_products" WHERE "spree_products"."deleted_at" IS NULL AND "spree_products"."slug" = $1 LIMIT $2  [["slug", "ruby-5646"], ["LIMIT", 1]]

[202, 211] in /home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/solidus-3dc450477ae5/backend/app/controllers/spree/admin/resource_controller.rb
   202:       Spree::Deprecation.warn "Calling #parent is deprecated on a ResourceController which has not defined a belongs_to"
   203:       nil
   204:     end
   205:   rescue ActiveRecord::RecordNotFound => e
   206:     byebug
=> 207:     resource_not_found(flash_class: e.model.constantize, redirect_url: spree.polymorphic_url([:admin, parent_model_name.pluralize]))
   208:   end
   209: 
   210:   def parent?
   211:     self.class.parent_data.present?
(byebug) next
Redirected to http://localhost:3000/admin/products
Return value is: nil

[65, 74] in /home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/solidus-3dc450477ae5/backend/app/controllers/spree/admin/base_controller.rb
   65: 
   66:       def resource_not_found(flash_class:, redirect_url:)
   67:         flash[:error] = flash_message_for(flash_class.new, :not_found)
   68:         redirect_to redirect_url
   69:         nil
   70:         byebug
=> 71:       end
   72:     end
   73:   end
   74: end
(byebug) step

[136, 145] in /home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/bundler/gems/solidus-3dc450477ae5/backend/app/controllers/spree/admin/resource_controller.rb
   136:     end
   137:   end
   138: 
   139:   def resource_not_found(flash_class: model_class, redirect_url: collection_url)
   140:     super
=> 141:     byebug
   142:   end
   143: 
   144:   def model_class
   145:     "Spree::#{controller_name.classify}".constantize
(byebug) step
Return value is: nil

[139, 148] in /home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/actionpack-6.0.3.2/lib/action_controller/metal.rb
   139:       false
   140:     end
   141: 
   142:     # Delegates to the class' <tt>controller_name</tt>.
   143:     def controller_name
=> 144:       self.class.controller_name
   145:     end
   146: 
   147:     attr_internal :response, :request
   148:     delegate :session, to: "@_request"
(byebug) next

[19, 28] in /home/ikraam/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/actionpack-6.0.3.2/lib/action_controller/metal/rescue.rb
   19: 
   20:     private
   21:       def process_action(*args)
   22:         super
   23:       rescue Exception => exception
=> 24:         request.env["action_dispatch.show_detailed_exceptions"] ||= show_detailed_exceptions?
   25:         rescue_with_handler(exception) || raise
   26:       end
   27:   end
   28: end
(byebug) next
```

ikraamg:
The latest commit is rebased onto master and except for the `&& parent`, the corrections discussed are made. :+1:

backend/app/controllers/spree/admin/resource_controller.rb
kennyadsl: Maybe I got it. From [the official rescue_from documentation](https://apidock.com/rails/ActiveSupport/Rescuable/ClassMethods/rescue_from):

> Handlers are inherited. They are searched from right to left, from bottom to top, and up the hierarchy. The handler of the first class for which exception.is_a?(klass) holds true is the one invoked, if any.

This seems to mean (or at least it's what I hope it means 🙂 ) that if you have multiple handlers (methods with the same name) in the hierarchy they will be all executed with that priority described above. 

Could this explain this odd behavior?

ikraamg: It seems that `rescue_from` and `rescue` act differently after the redirection.  The `rescue_from` works as expected but the `NoMethodError` is raised when using `rescue` in the method.

The context is lost by removing the `rescue` in the `parent` method and using the `rescue_from`. I have attempted a working solution below, but is there a simple way to find out in the `rescue_from` handler if the error was raised in the `parent` ? 

```ruby
rescue_from ActiveRecord::RecordNotFound do |e|
    if params['controller'].match?(/prices|store_credits|product_properties|adjustments|return_authorizations|customer_returns/)
      resource_not_found(flash_class: e.model.constantize, redirect_url: spree.polymorphic_url([:admin, parent_model_name.pluralize]))
    else
      resource_not_found
    end
  end
```

ikraamg: @kennyadsl, @aldesantis I have made an attempt to pass the context into the `rescue_from` by raising a "custom"  `ActiveRecord::RecordNotFound`.

Shall we go with this implementation and remove `&& parent`?

Implementation of rescue_from:
```ruby
rescue_from ActiveRecord::RecordNotFound do |e|
    e.message == 'parent' ? resource_not_found(flash_class: e.model.constantize, redirect_url: spree.polymorphic_url([:admin, parent_model_name.pluralize])) : resource_not_found
end
```
Rescue in parent:
```ruby
rescue ActiveRecord::RecordNotFound => e
    raise ActiveRecord::RecordNotFound.new('parent', e.model, e.primary_key, e.id)
end
```

ikraamg:
I believe the only blocking item this PR is the use of the `&& parent` condition in `#collection`. 

In order to summarise the discussion, I have made a [temporary commit](https://github.com/solidusio/solidus/pull/3728/commits/48d5b8c7fe60454d5ede207f3dd9083d0fe709d7) which implements and explains the alternative implementation for `&& parent` as mentioned in the [above thread](https://github.com/solidusio/solidus/pull/3728#discussion_r478895648).

jarednorman:
Thanks for your hard work on this. I've been a little short on time this week, but I've made a note to come back and take a look at this.

jarednorman (commented):
I think the final commit makes this less elegant... I'd rather stick with the extra condition. I think it makes sense.

kennyadsl:
I agree with @jarednorman, there's something going on there that we don't get now. Keeping the extra condition seems a good trade-off. If we go with removing the last commit and keep the extra condition, can we document well why are we keeping it via the commit description? Thanks, this work is awesome!

ikraamg:
Thanks @jarednorman and @kennyadsl for helping close the discussion. I agree that the alternative does not look very elegant.

I trust the commit description is adequate as it now explains the `&& parent`. I am excited to have this contribution pulled in! 

@jarednorman I appreciate the thorough guidance throughout the PR :handshake:

jarednorman (commented):
Thanks for all your work on this! I'm happy here once @aldesantis's thread is resolved.

aldesantis (approved):
Good to go as far as I'm concerned!

aldesantis:
Thanks so much @ikraamg, this is great. 🙂

Issue #3668, opened 2020-06-15
Admin user controller doesn't respect some abilities


halilim:
1. `#collection` does not check the current admin's abilities and returns all users. It can probably be fixed by using `super.ransack...` which makes use of `#accessible_by`.
 https://github.com/solidusio/solidus/blob/4079b58fe4a1917c1a5c81db86427e90f1d918f8/backend/app/controllers/spree/admin/users_controller.rb#L127

2. Similar issue for `#load_roles`, `#load_stock_locations` & `#set_stock_locations`, though `super` wouldn't cut it in these cases.
 https://github.com/solidusio/solidus/blob/4079b58fe4a1917c1a5c81db86427e90f1d918f8/backend/app/controllers/spree/admin/users_controller.rb#L163-L164
 https://github.com/solidusio/solidus/blob/4079b58fe4a1917c1a5c81db86427e90f1d918f8/backend/app/controllers/spree/admin/users_controller.rb#L170-L171
  `#load_stock_locations` also doesn't respect `user_params` and uses `params` directly instead.
 https://github.com/solidusio/solidus/blob/4079b58fe4a1917c1a5c81db86427e90f1d918f8/backend/app/controllers/spree/admin/users_controller.rb#L180-L181

3. `#set_roles` checks the general ability but not the roles given in params. I.e. a role that can't be managed could be given to a user by manually passing its id in params.
 https://github.com/solidusio/solidus/blob/4079b58fe4a1917c1a5c81db86427e90f1d918f8/backend/app/controllers/spree/admin/users_controller.rb#L175-L176

**Solidus Version**:
2.10.1

**To Reproduce**
In Admin > Users, with an admin whose permissions defined as:
```ruby
cannot :manage, ::Spree::User, ['some SQL'], &:foo?
cannot :manage, ::Spree::Role, ['some SQL'], &:bar?
```

**Current behavior**
1. Users with `foo? => true` are listed
2. Roles with `bar? => true` are included in Search > Roles, New User > Roles etc.
3. Roles with `bar? => true` can be assigned to a user

**Expected behavior**
1. Users with `foo? => true` are not listed
2. Roles with `bar? => true` are excluded from Search > Roles, New User > Roles etc.
3. Roles with `bar? => true` cannot be assigned to a user

kennyadsl:
We can close this one. Thanks @igorbp and @halilim for reporting!

PR #3732, opened 2020-08-15
Respect current ability in users controllers

igorbp:
**Description**

Admin users were able to see users, roles, and stock locations that they don't have permission to manage in admin users index and form. This fixes those authorization inconsistencies reported in issue #3668.

* Users that the admin user can't manage are not listed anymore.
* Roles that the admin user can't manage are not listed in the search input nor in the users' create/update form.
* Stock locations that the admin user can't manage are not listed in the user's create/update form.
* Fixed problem that not accessible roles and stock locations passed in `:spree_role_ids` and `:stock_locations_ids` params were still being assigned to the user.

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/users_controller.rb
+++ b/backend/app/controllers/spree/admin/users_controller.rb
@@ -168,7 +168,7 @@ module Spree
       end
 
       def load_stock_locations
-        @stock_locations = Spree::StockLocation.all
+        @stock_locations = Spree::StockLocation.accessible_by(current_ability)
       end
 
       def set_roles
--- a/backend/app/controllers/spree/admin/users_controller.rb
+++ b/backend/app/controllers/spree/admin/users_controller.rb
@@ -161,7 +161,7 @@ module Spree
       end
 
       def load_roles
-        @roles = Spree::Role.all
+        @roles = Spree::Role.accessible_by(current_ability)
         if @user
           @user_roles = @user.spree_roles
         end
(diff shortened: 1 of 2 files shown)


backend/app/controllers/spree/admin/users_controller.rb
```diff
@@ -178,7 +182,10 @@ def set_roles
       end
 
       def set_stock_locations
-        @user.stock_locations = Spree::StockLocation.where(id: (params[:user][:stock_location_ids] || []))
+        if user_params[:stock_location_ids]
+          @user.stock_locations =
+            Spree::StockLocation.accessible_by(current_ability).where(id: user_params[:stock_location_ids])
+        end
```
spaghetticode, line 188: I think the behavior was slightly changed. The old code always set the user stock locations, also when the parameter was not present, in that case by stock locations to an empty array. This does not happen anymore here. Is this change wanted?

spaghetticode (commented):
@igorbp thanks for these fixes 🎉 

It looks good to me, I left only one question about a change that I'm not 100% sure about.

backend/app/controllers/spree/admin/users_controller.rb
igorbp: This is a good question. I took this decision because with these changes admins that can't manage stock locations would remove user's stock locations since the params wouldn't be there.

aldesantis (approved):
Thanks so much @igorbp, looks great!

PR #3734, opened 2020-08-16
Make payment.rb methods more concise

brchristian:
**Description**

In payment.rb, the `source_required?` method uses the `foo.present? && foo.bar` pattern, which can more concisely be expressed in Ruby >= 2.3 using the safe navigation operator as `foo&.bar`.

Likewise, the `profiles_supported?` method uses the `foo.respond_to?(:bar) && foo.bar` pattern, which can more concisely be expressed using Rails's `try` method as `foo.try(:bar)`.

### DIFF
--- a/core/app/models/spree/payment.rb
+++ b/core/app/models/spree/payment.rb
@@ -179,11 +179,11 @@ module Spree
     end
 
     def source_required?
-      payment_method.present? && payment_method.source_required?
+      !!payment_method&.source_required?
     end
 
     def profiles_supported?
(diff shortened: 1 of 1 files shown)


brchristian:
One additional thought is that we could also prepend with `!!` if we wanted to coerce the return value to a boolean.

jarednorman:
That might be worth it to be extra safe. You never know whether someone might be relying on `== false` or something. 😅

brchristian:
@jarednorman Done! 👍

spaghetticode (approved):
@brchristian thanks! 👍

PR #3736, opened 2020-08-21
Upgrade acts_as_list gem dependency to allow v1.x

marcrohloff:
**Description**
Updates the acts_as_list gem dependency for solidus-core from `~0.3` (which matches up to v0.9.19) to `< 2.0` 
(the current acts_as_list version is 1.0 released 2019-09-26)

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -29,7 +29,7 @@ Gem::Specification.new do |s|
   end
 
   s.add_dependency 'activemerchant', '~> 1.66'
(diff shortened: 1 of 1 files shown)


kennyadsl (approved):
The only dangerous thing I can see in the `acts_as_list` [CHANGELOG](https://github.com/brendon/acts_as_list/blob/master/CHANGELOG.md) is dropping support for older Rails versions that we are not supporting anyway. 👍 Thank you!

PR #3737, opened 2020-08-21
Fix ruby 2.7 warnings on core

stefano-sarioli:
**Description**

Ruby 2.7 enabled some new warnings on how you can pass default and hash parameters to a function.
This PR fixes the warning inside our core code, the remaining on the core are coming from some external gems, I'm going to open issues on the affected gems about this.

I thought that doing one component at a time, instead of all the code at once, can create enable reviews and avoid some small errors that can be missed when doing repetitive tasks like this one.

**Checklist:**


aldesantis:
Thanks @stefano-sarioli!

PR #3741, opened 2020-08-24
Allow for HTML options on image partial

markmead:
**Description**
Allows the passing of HTML attributes to the image partial. 

Recently added this to a project as I needed a class passed down to the image.

Little unsure if there's a better alternative to `options ||= {}` so any guidance there would be great! 🙌

**Checklist:**


aldesantis (approved):
This looks good, thanks! I think `options ||= {}` is fine since it follows the convention for other local variables.

PR #3744, opened 2020-08-26
Skip payment method ext in sandbox and app template

elia:
**Description**

For the sandbox and template apps we should handle the payment method in the same way we handle authentication.

**Checklist:**

### DIFF
--- a/bin/rails-application-template
+++ b/bin/rails-application-template
@@ -19,6 +19,7 @@ end
 
 gem 'solidus'
 gem 'solidus_auth_devise'
+gem 'solidus_paypal_commerce_platform'
 gem 'rails_12factor'
 
 gem_group :heroku do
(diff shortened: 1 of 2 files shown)


kennyadsl:
Is the dummy_app (the one we use in the test suite) also missing this configuration?

peterberkenbosch:
@elia @kennyadsl is right. we need to add this here https://github.com/solidusio/solidus/blob/master/core/lib/spree/testing_support/common_rake.rb#L17 as well.

waiting-for-dev:
Thanks, @elia! Closing by https://github.com/solidusio/solidus/commit/a7ba5250f313ca867f0691221f487bc1a2ac69d1. It's still missing adding it to the application template. Feel free to create a PR for that if somebody needs it.

PR #3748, opened 2020-08-28
Revert "Add payment method selection to the install generator"

elia:
**Description**

Apparently introducing the payment is causing issues on the dummy app generation within extensions.
We need to revert the PR and reintroduce it when the issue is solved.

See, as an example, https://app.circleci.com/pipelines/github/solidusio-contrib/solidus_static_content/170/workflows/b770419d-d80e-4048-8820-b0f404896fd4/jobs/402.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,7 +5,7 @@ require 'bundler'
 task default: :spec
 
 def print_title(gem_name = '')
-  title = ["Solidus", gem_name].join(' ').strip
+  title = ["Solidus", gem_name].join(' ')
   puts "\n#{'-' * title.size}\n#{title}\n#{'-' * title.size}"
 end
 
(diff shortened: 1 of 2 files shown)


aldesantis (approved):

peterberkenbosch:
👍 There are a few changes here that actually make sense, but the main issue is described here: https://github.com/solidusio/solidus/pull/3744 the dummy app installer is using `verbose=false` so the error output is hidden. The main line that is causing issues with extensions is: https://github.com/solidusio/solidus/blob/master/core/lib/spree/testing_support/common_rake.rb#L17, adding the `"--payment-method=none"` to that line solves it.

kennyadsl:
@elia & @peterberkenbosch if I got it correctly we can close this one now after we merged #3751.

peterberkenbosch:
@kennyadsl yes, this is fine to close now indeed.

PR #3749, opened 2020-08-28
Do not consider promotions without actions as active

DanielePalombo:
This PR is based on https://github.com/solidusio/solidus/pull/3596.

From original PR:

This is a proposed solution to #2777 where action-less promotions were considered active. As opposed to my original thought of implementing validation, given that there is no state toggle on the promotion that it would make sense to embrace the inferred state. So this implementation just causes a promotion to be considered inactive until it has actions, in addition to the other required material. Otherwise it is considered inactive.

In the process I found the coupon promotion handler already treated action-less coupons as inactive, so I updated it to continue this behavior.

---

I have only added the spec and refactored it.

Thank you @wildbillcat  

**Checklist:**

### DIFF
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -48,9 +48,9 @@ module Spree
     scope :started_and_unexpired, -> do
       table = arel_table
       time = Time.current
-      joins(:promotion_actions).
-          where(table[:starts_at].eq(nil).or(table[:starts_at].lt(time))).
-          where(table[:expires_at].eq(nil).or(table[:expires_at].gt(time)))
+
+      where(table[:starts_at].eq(nil).or(table[:starts_at].lt(time))).
+        where(table[:expires_at].eq(nil).or(table[:expires_at].gt(time)))
     end
     scope :has_actions, -> do
       joins(:promotion_actions)
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -39,10 +39,21 @@ module Spree
     scope :coupons, -> { joins(:codes).distinct }
     scope :advertised, -> { where(advertise: true) }
     scope :active, -> do
+      if Spree::Config.actionless_promotion_inactive
+        has_actions.started_and_unexpired
+      else
+        started_and_unexpired
+      end
+    end
+    scope :started_and_unexpired, -> do
       table = arel_table
       time = Time.current
-      where(table[:starts_at].eq(nil).or(table[:starts_at].lt(time))).
-        where(table[:expires_at].eq(nil).or(table[:expires_at].gt(time)))
+      joins(:promotion_actions).
+          where(table[:starts_at].eq(nil).or(table[:starts_at].lt(time))).
+          where(table[:expires_at].eq(nil).or(table[:expires_at].gt(time)))
+    end
+    scope :has_actions, -> do
+      joins(:promotion_actions)
     end
     scope :applied, -> { joins(:order_promotions).distinct }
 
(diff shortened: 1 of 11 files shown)


wildbillcat:
Thanks for hopping on this @DanielePalombo ! Apologies, I've been underwater with work and haven't been able to make time back to this.

core/app/models/spree/promotion.rb
```diff
@@ -84,7 +93,7 @@ def not_expired?
     end
 
     def active?
-      started? && not_expired?
+      started? && not_expired? && (Spree::Config.actionless_promotion_inactive ? actions.present? : true)
```
kennyadsl, line 96: What about a different method (`has_actions?`) for this?

kennyadsl (approved):
Left a non-blocking comment. 👍 

Thanks to both @wildbillcat and @DanielePalombo.

core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
```diff
@@ -42,6 +42,11 @@ Spree.config do |config|
   # their previous location first.
   config.redirect_back_on_unauthorized = true
 
+  # Set this configuration to `false` to allow promotions
+  # with no associated actions to be active for use by customers.
+  # See https://github.com/solidusio/solidus/issues/2777 for more info.
```
aldesantis, line 47: Can you link to the new PR here instead?

core/lib/spree/app_configuration.rb
```diff
@@ -213,6 +213,10 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Integer] Promotions to show per-page in the admin (default: +15+)
     preference :promotions_per_page, :integer, default: 15
 
+    # @!attribute [rw] disable_actionless_promotion_validation
+    #   @return [Boolean] Promotions should have actions associated before activation (default: +true+)
+    preference :actionless_promotion_inactive, :boolean, default: false
```
aldesantis, line 218: How about making this more explicit by calling it `consider_actionless_promotions_active` instead?

aldesantis (changes requested):

core/app/models/spree/promotion.rb
DanielePalombo: I think that `has_actions?` would be confused because it will return `true` is the preference `actionless_promotion_inactive` is false.

I have considered this suggestion and following the @aldesantis comments I have named it `consider_active_by_actions_presence`

core/lib/spree/app_configuration.rb
DanielePalombo: Done!

https://github.com/solidusio/solidus/pull/3749/commits/24f884cc9705c0c4987a82fbd0fa83357e21b43f

core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
DanielePalombo: 👍

core/lib/spree/app_configuration.rb
```diff
@@ -217,6 +217,10 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Integer] Promotions to show per-page in the admin (default: +15+)
     preference :promotions_per_page, :integer, default: 15
 
+    # @!attribute [rw] disable_actionless_promotion_validation
+    #   @return [Boolean] Promotions should have actions associated before activation (default: +true+)
```
kennyadsl, line 221: I'd  use `being considered active` to be consistent with the language used in the rest of the PR. `activation` makes me think that there's something to activate with an action.

kennyadsl (approved):
Left a couple of comments about some terms. The rest looks good to me, thanks!

kennyadsl:
@aldesantis can you please review again when you have some time? 🙏

aldesantis (approved):

Issue #3750, opened 2020-08-29
Can't make a dummy app for testing on the master branch


kainio:
I can't make a dummy app for testing in the master branch knowing that it's working correctly in the v2.10.2 branch,  I am having this error 

**Solidus Version**:
V2.11.0alpha

**To Reproduce**
Execute `bundle exec bin/rake extension:test_app extension:specs` inside a solidus extension

**Current behavior**
All the migrations are copied but just the extension's migration are run.

```
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

SQLite3::SQLException: no such table: spree_products
...
[Spree WARNING] Missing migrations.
[Spree WARNING] add_amount_remaining_to_store_credit_events from spree is missing.
[Spree WARNING] add_apply_to_all_to_variant_property_rule from spree is missing.
[Spree WARNING] add_available_locales_to_stores from spree is missing.
[Spree WARNING] add_available_to_columns_and_remove_display_on_from_payment_methods from spree is missing.
[Spree WARNING] add_available_to_users_and_remove_display_on_from_shipping_methods from spree is missing.
[Spree WARNING] add_bcc_email_to_spree_stores from spree is missing.
[Spree WARNING] add_default_billng_flag_to_user_addresses from spree is missing.
[Spree WARNING] add_index_to_spree_payments_number from spree is missing.
[Spree WARNING] add_join_characters_to_promotion_code_batch from spree is missing.
[Spree WARNING] add_lft_and_rgt_indexes_to_taxons from spree is missing.
[Spree WARNING] add_roles_unique_constraints from spree is missing.
[Spree WARNING] add_time_range_to_tax_rate from spree is missing.
[Spree WARNING] create_promotion_rule_stores from spree is missing.
[Spree WARNING] create_spree_promotion_code_batch from spree is missing.
[Spree WARNING] create_spree_store_credit_reasons_table from spree is missing.
[Spree WARNING] create_spree_wallet_payment_sources from spree is missing.
[Spree WARNING] create_store_shipping_methods from spree is missing.
[Spree WARNING] drop_spree_store_credit_update_reasons from spree is missing.
[Spree WARNING] migrate_credit_cards_to_wallet_payment_sources from spree is missing.
[Spree WARNING] remove_code_from_spree_promotions from spree is missing.
[Spree WARNING] remove_currency_from_line_items from spree is missing.
[Spree WARNING] remove_default_tax_from_spree_zones from spree is missing.
[...]

PR #3751, opened 2020-08-31
Set payment method to none

peterberkenbosch:
**Description**
By default Solidus is now looking for `solidus_paypal_commerce_platform`,
however, with the sandbox and the dummy generator we need to set this
to `none`, otherwise these scripts are not running successfully.

This commit fixes that. It does not update the application template for Heroku
though. That will be a seperate PR.

Ref #3750 #3748 #3744 

**Checklist:**


kennyadsl (approved):
👍

aldesantis (approved):
Thanks @peterberkenbosch!

PR #3753, opened 2020-09-04
Explicit deprecation warning mocks

cedum:
When testing, instead of silencing/mocking all deprecation warnings, test the specific deprecation warning needed for the specific test case.
Since we make fail the CI build when there're unhandled deprecation warnings, this ensures we have expectations only for specific deprecation warnings instead of silencing/mocking all warnings.

**Checklist:**

### DIFF
--- a/api/spec/requests/spree/api/checkouts_controller_spec.rb
+++ b/api/spec/requests/spree/api/checkouts_controller_spec.rb
@@ -307,10 +307,13 @@ module Spree
             }
           end
 
+          before do
+            expect(Spree::Deprecation).to receive(:warn).
+              with(/^Passing existing_card_id to PaymentCreate is deprecated/, any_args)
+          end
+
           it 'succeeds' do
-            Spree::Deprecation.silence do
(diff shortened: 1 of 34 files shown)


spaghetticode:
@cedum thank you for this PR, I think it's a good improvement 👍 

Have you considered the idea of adding a helper method for improving readability? Something that would allow us to replace this:

```
expect(Spree::Deprecation).to receive(:warn).
              with(/^Passing existing_card_id to PaymentCreate is deprecated/, any_args)
```

with something like:
```expect_deprecation_with(/^Passing existing_card_id to PaymentCreate is deprecated/)```

it adds a bit of overhead, so I'm not really sure it's worth the effort, still I'd like to know others' opinions.

kennyadsl (approved):
Super in favor of this! Let us know what do you think about the @spaghetticode comment.

jarednorman (approved):
I like this too.

cedum:
Agree with @spaghetticode, a helper might simplify a bit the work with deprecations in tests.

PR #3758, opened 2020-09-11
Change internal mapping for event subscriber registrations

### LINKED ISSUE #3757: Event subscribers loaded twice with Rails 6/Zeitwerk
**Solidus Version**:
master

**Description**

Custom defined event subscriber modules may end up being subscribed twice (or more!) on Solidus apps based on Rails 6.x. I was not able to reproduce this issue on Rails 5.2. 

The problem happens only in development (or more generally when classes are not cached) and when event subscribers are loaded using initializers, as we used to suggest in the Event section of Solidus documentation. 

Apparently, the issue is related to the new Zeitwerk autoload. 

The simplest fix for this is to simply remove the offending initializers. Subscribers will be loaded automatically anyways thanks to #3571.

If for any reason somebody does not want to remove the initializers or wants to experiment with the issue, by explicitly requiring the subscriber file in the initializer will fix it:

```ruby
require Rails.root.join('app/subscribers/custom_subscriber')

CustomSubscriber.subscribe!
```

I'm working on a more robust solution (a centralized way to track event subscribers subscriptions) that will completely remove the issue.

spaghetticode:
Fixes #3757 

The previous way of registering subscribers could generate some issues with rails 6 applications, due to how Zeitwerk (re)loads classes in initializers. The previous approach was also scattering the information into every single subscriber, while now it's centralized in a single place.

Currently, `Spree::Event.subscribers` includes only the list of subscriber classes, while the rest of the information is delegated to each subscriber. This is how information can be accessed with the current interface:

```ruby
# list of subscribers:
$ Spree::Event.subscribers
 => #<Spree::Core::ClassConstantizer::Set:0x00007ff62dd69a18 @collection=#<Set: {"AppEventSubscriber", "Spree::MailerSubscriber"}>> 

# internal mapping for AppEventSubscriber:
$ AppEventSubscriber.event_actions
 => {:app_event=>"app_event"} 

# mapping to the ActiveSupport::Notifications subscription:

```ruby
AppEventSubscriber.app_event_handler
 => #<ActiveSupport::Notifications::Fanout::Subscribers::Timed:0x00007ff67e582880 @pattern="app_event.spree", @delegate=#<Proc:0x00007ff67e582df8@/Users/andrea/nebulab/solidus_event_subscribers_autoload/vendor/gems/solidus/core/lib/spree/event/adapters/active_support_notifications.rb:18>, @can_publish=false> 
```

After merging this PR, the functionality provided by `Spree::Event.subscribers` will be managed by the new `Spree::Event.subscriber_registry`. The registry stores information with the format:

```ruby
{ subscriber_class_name => { subscriber_method => ActiveSupportNotifications_subscription } }
```
Real world example:

```ruby
pp Spree::Event.subscriber_registry
#<Spree::Event::SubscriberRegistry:0x00007fce3e80d798
 @registry=
  {"Spree::MailerSubscriber"=>
    {:order_finalized=>
      #<ActiveSupport::Notifications::Fanout::Subscribers::Timed:0x00007fce3ec86e78
       @can_publish=false,
       @delegate=
        #<Proc:0x00007fce3ec86f40@/Users/andrea/nebulab/solidus/core/lib/spree/event/adapters/active_support_notifications.rb:18>,
       @pattern="order_finalized.spree">,
     :send_reimbursement_email=>
      #<ActiveSupport::Notifications::Fanout::Subscribers::Timed:0x00007fce3ec86a18
       @can_publish=false,
       @delegate=
        #<Proc:0x00007fce3ec86ab8@/Users/andrea/nebulab/solidus/core/lib/spree/event/adapters/active_support_notifications.rb:18>,
       @pattern="reimbursement_reimbursed.spree">},
   "AppSubscriber"=>
    {:app_event=>
      #<ActiveSupport::Notifications::Fanout::Subscribers::Timed:0x00007fce3ec851e0
       @can_publish=false,
       @delegate=
        #<Proc:0x00007fce3ec85280@/Users/andrea/nebulab/solidus/core/lib/spree/event/adapters/active_support_notifications.rb:18>,
       @pattern="app_event.spree">}},
 @semaphore=#<Thread::Mutex:0x00007fce3e80d6a8>
```

This PR requires a small change in solidus_support too, see https://github.com/solidusio/solidus_support/pull/53

These changes should be transparent to stores as they only modify how event subscribers are mapped and managed internally in Solidus.

Many thanks to @cedum and @DanielePalombo who helped in moving forward the PR.

**Checklist:**


core/lib/spree/event/subscriber.rb
```diff
@@ -66,9 +73,11 @@ def event_action(method_name, event_name: nil)
       def subscribe!
         unsubscribe!
         event_actions.each do |event_action, event_name|
-          send "#{event_action}_handler=", Spree::Event.subscribe(event_name) { |event|
-            send event_action, event
-          }
+          subscription = Spree::Event.subscribe(event_name) { |event| send event_action, event }
+          # old mappings, to be removed when Solidus 2.10 is not supported anymore:
```
kennyadsl, line 77: Can we use the word deprecated instead of `old`? Just because it will be easier to find it if we perform a search of all the deprecated code within the codebase with a text editor.

core/lib/spree/event/subscriber.rb
```diff
@@ -66,9 +73,11 @@ def event_action(method_name, event_name: nil)
...
+          send "#{event_action}_handler=", subscription
+          # new mappings:
```
kennyadsl, line 79: I think we can get rid of this comment, probably improving a little bit the one with the deprecated code to explain that this other way is the new way of doing things.

core/lib/spree/testing_support/preferences.rb
```diff
@@ -14,7 +14,9 @@ module Preferences
       #
       # @deprecated
       def reset_spree_preferences(&config_block)
-        Spree::Config.instance_variables.each { |iv| Spree::Config.remove_instance_variable(iv) }
+        Spree::Config.instance_variables.
+          reject { |iv| iv == :@events_configuration }.
```
kennyadsl, line 18: I'm not sure about this. What if some specs need to reset that configuration? It's applicable? How do people know that all the preferences except that one will be reset when calling `reset_spree_preferences` in their code?

core/lib/spree/core/engine.rb
```diff
@@ -48,15 +48,11 @@ class Engine < ::Rails::Engine
       initializer 'spree.core.initialize_subscribers' do |app|
         app.reloader.to_prepare do
           Spree::Event.require_subscriber_files
-          Spree::Event.subscribers.each_key do |klass|
-            klass.constantize.subscribe!
-          end
+          Spree::Event.subscribe_subscribers!
```
kennyadsl, line 51: The more I look into this and the more I think we need another class called `Spree::Event::Subscribers` that acts as a collection of subscribers. No matter if it's hash inside but the interface will probably look better, having something like:

`Spree::Event.subscribers.subscribe_all`. 

WDYT?

core/lib/spree/event/subscriber.rb
```diff
@@ -54,7 +54,14 @@ def self.included(base)
       #     end
       #   end
       def event_action(method_name, event_name: nil)
-        mattr_accessor "#{method_name}_handler"
+        mattr_writer "#{method_name}_handler"
```
kennyadsl, line 57: Why don't we need to deprecate the writer as well?

kennyadsl (commented):
Left some comments, I think we are close to a solution, thanks @spaghetticode !

core/lib/spree/event/subscriber.rb
spaghetticode: We still use the writer internally to keep compatibility, if somebody is still using the reader in their codebase (we don't use the reader anymore). So, if we deprecate the writer, then we'll need to silence the deprecation [when using it](https://github.com/solidusio/solidus/pull/3758/files#diff-d6a7a37486ba2958ea8f0d091d2df618R78). So, deprecating the writer would be impractical at this point.

But I think we don't really need to deprecate the writer as well. The only reason for some developers to be using the writer in their codebase is to set a value to be accessed from the reader. When this happens, they will see the deprecation when the matching reader is called.

core/lib/spree/testing_support/preferences.rb
spaghetticode: This helper method is deprecated, so it won't be around for long. We don't use it anymore in our codebase, but we have some specs that verify its behavior. I'm not sure if creating a better workaround for making it work properly with the new event mappings is worth the effort at this point.

core/lib/spree/core/engine.rb
spaghetticode: sure, makes sense! I'll work on your suggestion 👍

core/lib/spree/testing_support/preferences.rb
kennyadsl: Oh right, it's deprecated. 👍

core/lib/spree/event/subscriber.rb
kennyadsl: Makes sense, thanks Andrea.

core/lib/spree/event/subscriber.rb
```diff
@@ -54,32 +54,35 @@ def self.included(base)
...
-      # Unsubscribes all declared event actions from their events. This means that when an event
+      # @deprecated Unsubscribes all declared event actions from their events. This means that when an event
       # fires then none of its unsubscribed event actions will be called.
       # @example unsubscribe all event actions for module 'EmailSender'
       #    EmailSender.unsubscribe!
       def unsubscribe!
-        event_actions.keys.each do |event_action|
-          Spree::Event.unsubscribe send("#{event_action}_handler")
-        end
+        Spree::Deprecation.warn("#{name}.unsubscribe! is deprecated. Please move your event subscribers to folder `app/subscribers/*_subscriber.rb` and they will be unloaded automatically. ", caller)
```
kennyadsl, line 84: I think the message here is incorrect. You have to delete the subscriber file to unsubscribe, isn't it? Also, I'd keep the method since maybe you don't have control over where the file has been added (core or an extension).

core/lib/spree/event/subscribers.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Spree
+  module Event
+    class Subscribers
+      attr_reader :mappings
```
kennyadsl, line 6: `mappings` seems like an implementation detail concept, I'd avoid using it as the main public interface of this class.

kennyadsl: I made a quick attempt [here](https://github.com/solidusio/solidus/commit/d786b56428eb5e55754523bb53aba60aa68f1a4d) to explain what I was thinking (and being sure I'm not missing something big). Let me know your thoughts!

kennyadsl: The downside of my proposal is that we are exposing all the hash methods to users that in theory will be able to play too much with this mapping of subscribers. But at the end of the day, it's the same effect we have making `mappings` public, which is a hash.

spaghetticode: I think subclassing core ruby classes is generally considered an antipattern. A couple of old articles on that topic: 
* https://tenderlovemaking.com/2014/06/02/yagni-methods-are-killing-me.html 
* https://steveklabnik.com/writing/beware-subclassing-ruby-core-classes

core/lib/spree/event/subscriber.rb
spaghetticode: Yes, you're right about the message. And also the remainder of the comment is right. I think we need to keep offering that kind of granularity, even though it will be rarely used.

core/spec/models/spree/order_spec.rb
```diff
@@ -28,11 +28,16 @@
         end.to change(order, :confirmation_delivered).to true
       end
 
+      it 'does send the email' do
```
jarednorman, line 31: Simpler:

```suggestion
      it 'sends the email' do
```

core/lib/spree/core/engine.rb
```diff
@@ -47,12 +47,11 @@ class Engine < ::Rails::Engine
...
       initializer 'spree.core.initialize_subscribers' do |app|
         app.reloader.to_prepare do
-          Spree::Event.require_subscriber_files
-          Spree::Event.subscribers.each(&:subscribe!)
+          Spree::Event.activate_autoloadable_subscribers
         end
 
         app.reloader.before_class_unload do
-          Spree::Event.subscribers.each(&:unsubscribe!)
+          Spree::Event.deactivate_subscribers
```
kennyadsl, line 54: `deactivate_all_subscribers`?

kennyadsl (changes requested):
Found something that looks like a mistake, let me know if I'm missing something instead.

spaghetticode:
@kennyadsl thanks for spotting the bug ❤️ 

I have just tested this PR with a newly generated sandbox app using `solidus_support` with [this branch](https://github.com/solidusio/solidus_support/pull/53), and it seems to be working properly when adding a custom autoloadable subscriber.

kennyadsl (approved):
Thanks @spaghetticode!

jarednorman (approved):

elia:
I'm getting these deprecations while running a generator (although I don't believe that to be relevant), I think the caller should be refined to point at the right level of the backtrace, instead of showing `-e`.

```
DEPRECATION WARNING: `Spree::Event.subscribers` is deprecated. Please use `Spree::Event.subscriber_registry` instead. (called from <main> at -e:1)
DEPRECATION WARNING: `Spree::Config.events.subscribers` is deprecated. Please use `Spree::Config.events.subscriber_registry`. (called from <main> at -e:1)
DEPRECATION WARNING: Spree::MailerSubscriber.subscribe! is deprecated. Please use `Spree::MailerSubscriber.activate`. (called from <main> at -e:1)
DEPRECATION WARNING: `Spree::Event.subscribers` is deprecated. Please use `Spree::Event.subscriber_registry` instead. (called from <main> at -e:1)
DEPRECATION WARNING: `Spree::Config.events.subscribers` is deprecated. Please use `Spree::Config.events.subscriber_registry`. (called from <main> at -e:1)
DEPRECATION WARNING: Spree::MailerSubscriber.subscribe! is deprecated. Please use `Spree::MailerSubscriber.activate`. (called from <main> at -e:1)
DEPRECATION WARNING: `Spree::Event.subscribers` is deprecated. Please use `Spree::Event.subscriber_registry` instead. (called from <main> at -e:1)
DEPRECATION WARNING: `Spree::Config.events.subscribers` is deprecated. Please use `Spree::Config.events.subscriber_registry`. (called from <main> at -e:1)
DEPRECATION WARNING: Spree::MailerSubscriber.subscribe! is deprecated. Please use `Spree::MailerSubscriber.activate`. (called from <main> at -e:1)
DEPRECATION WARNING: `Spree::Event.subscribers` is deprecated. Please use `Spree::Event.subscriber_registry` instead. (called from <main> at -e:1)
DEPRECATION WARNING: `Spree::Config.events.subscribers` is deprecated. Please use `Spree::Config.events.subscriber_registry`. (called from <main> at -e:1)
DEPRECATION WARNING: Spree::MailerSubscriber.subscribe! is deprecated. Please use `Spree::MailerSubscriber.activate`. (called from <main> at -e:1)
```

Unless I'm mistaken and this is correct for some reason we should probably turn this comment into an issue

cc @spaghetticode

PR #3759, opened 2020-09-14
Fix TaxCategory default not showing up in admin

vl3:
**Description**
When visiting New products page, Spree::TaxCategory with _is_default_ set to true is not being shown as the first selected option in the tax category select box. Additionally, the hint of the select box is saying that the default is None despite having other Spree::TaxCategory as default. This PR aims to fix it.

**Without change**
tax_category_failed

**With change**
tax_category_success

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/products_controller.rb
+++ b/backend/app/controllers/spree/admin/products_controller.rb
@@ -74,6 +74,7 @@ module Spree
 
       def load_data
         @tax_categories = Spree::TaxCategory.order(:name)
+        @default_tax_category = @tax_categories.detect(&:is_default)
         @shipping_categories = Spree::ShippingCategory.order(:name)
       end
 
--- a/backend/config/initializers/form_builder.rb
+++ b/backend/config/initializers/form_builder.rb
@@ -13,8 +13,8 @@ class ActionView::Helpers::FormBuilder
   end
 
   def field_hint(method, options = {})
-    title = options[:title] || @object.class.human_attribute_name(method)
-    text = options[:text] || I18n.t(method, scope: [:spree, :hints, @object.class.model_name.i18n_key])
+    title = options.delete(:title) || @object.class.human_attribute_name(method)
+    text = options.delete(:text) || I18n.t(method, scope: [:spree, :hints, @object.class.model_name.i18n_key], **options)
     @template.admin_hint(title, text)
   end
 end
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -146,8 +146,8 @@
       <div data-hook="admin_product_form_tax_category">
         <%= f.field_container :tax_category do %>
           <%= f.label :tax_category_id, Spree::TaxCategory.model_name.human %>
-          <%= f.field_hint :tax_category %>
-          <%= f.collection_select(:tax_category_id, @tax_categories, :id, :name, { include_blank: t('spree.match_choices.none') }, { class: 'custom-select' }) %>
+          <%= f.field_hint :tax_category, default_tax_category: @default_tax_category&.name %>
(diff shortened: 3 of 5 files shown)


kennyadsl:
Thanks! It seems to make sense but I can't find in the codebase a single place where we are using the default tax category. I mean in the tax calculation... Can you help me understand if it's better to remove the concept of the default tax category entirely?

jarednorman:
I think it _is_ useful to have a tax category that's the default for creating new products. Beyond that I don't think it has any meaning, but it's still useful for that.

backend/config/initializers/form_builder.rb
```diff
@@ -14,7 +14,8 @@ def error_message_on(method, options = {})
 
   def field_hint(method, options = {})
     title = options[:title] || @object.class.human_attribute_name(method)
-    text = options[:text] || I18n.t(method, scope: [:spree, :hints, @object.class.model_name.i18n_key])
+    text_interpolation_args = options.fetch(:text_interpolation_args, {})
+    text = options[:text] || I18n.t(method, scope: [:spree, :hints, @object.class.model_name.i18n_key], **text_interpolation_args)
```
kennyadsl, line 18: I think there may be another way to do this: if we delete the other options we can just pass the remaining ones to the `translate` method instead of using a specific option for this text interpolation. Something like:

```ruby
title = options.delete(:title) || @object.class.human_attribute_name(method)
text = options.delete(:text) || I18n.t(method, scope: [:spree, :hints, @object.class.model_name.i18n_key], **options)
```

The idea is that we could need to pass also other `i18n.t` options along the way (like `default`, for example), and this way it's a bit more flexible, even if less explicit. What do you think?

vl3: Yeah, agree, I wasn't entirely sure what would be preferred. I will make this change.

kennyadsl:
Thanks, I think it's ready to be merged so I'm going to approve. 

An extra thing that we could do in terms of UX is explaining what's the meaning of the default checkbox in the Tax Category edit page with another hint:

Screenshot 2020-09-24 at 11 31 40

Of course, we can also do that with another PR. Let me know if you want to take care of that or I'll open a new issue to keep track of this. 

Thanks again! 🙏

kennyadsl (approved):
👍 Thanks!

backend/app/controllers/spree/admin/products_controller.rb
```diff
@@ -74,6 +74,7 @@ def location_after_save
 
       def load_data
         @tax_categories = Spree::TaxCategory.order(:name)
+        @default_tax_category = @tax_categories.default
```
aldesantis, line 77: Correct me if I'm wrong, but this will execute an additional SQL query, right? How about selecting the default tax category with `detect(&:is_default)` instead? Small improvement, but no reason why we shouldn't do it.

aldesantis (changes requested):
Thanks @vl3, just left a comment on an extra SQL query that I think we can get rid of.

backend/app/controllers/spree/admin/products_controller.rb
vl3: You are correct, _default_ was executing an additional query. I have changed to use _detect_ as you suggested! 
Thanks!

vl3:
> Thanks, I think it's ready to be merged so I'm going to approve.
> 
> An extra thing that we could do in terms of UX is explaining what's the meaning of the default checkbox in the Tax Category edit page with another hint:
> 
> Screenshot 2020-09-24 at 11 31 40
> 
> Of course, we can also do that with another PR. Let me know if you want to take care of that or I'll open a new issue to keep track of this.
> 
> Thanks again!

I can take care of that next week. Is it ok if I merge this and then open another PR?

aldesantis (approved):

aldesantis:
@vl3 thanks, looking great! Also looking forward to your next PR. ❤️

PR #3763, opened 2020-09-18
Fix address validation having a country w/o states

cedum:
This is an extract from #3129 and addresses the [concern](https://github.com/solidusio/solidus/pull/3129#issuecomment-689655524) having a separate commit fixing #3098 and enabling the stores to backport the fix more easily without being forced to upgrade.

This mainly fixes the address' state validation bug when switching from a country with states (regions) to one without.
Before validating that the associated state belongs to the associated country, ensure the country does have states; if it doesn't, nullify the `state_id` since we don't need one here.

This is quite a tricky bug that has been partially introduced when the `validate_state_matches_country` validation method has been merged.
Note: both state validation methods are being deprecated in #3129 and partially refactored into a configurable validation class allowing to override the default behavior and handling any region-based edge case around the address' state normalization and validation logic.

**Checklist:**

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -253,6 +253,9 @@ module Spree
     end
 
     def validate_state_matches_country
+      return unless country
+
+      self.state = nil if country.states.empty?
       if state && state.country != country
         errors.add(:state, :does_not_match_country)
       end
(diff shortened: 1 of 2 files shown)


aldesantis (approved):

kennyadsl (approved):
👍 thanks Dumitru!

kennyadsl:
@cedum if I understand correctly we can partially revert [this PR](https://github.com/solidusio/solidus/pull/2371) now? I mean, at least the JS and HTML parts that look like hacks to fix the same issue client side.

PR #3767, opened 2020-09-21
Make order customer email links consistent

brchristian:
**Description**
Currently when looking at the table of orders on the backend, the customer email will always be displayed as a link. However, the behavior of this link is completely different based on whether the customer has a user account or not. If they _do_, it will link to that user's User page on the backend. If they _don't_, it is a simple mailto link.

(Can you tell that clicking one of these will do one thing and clicking the other will do something totally different?)

Screen Shot 2020-09-21 at 11 42 53 AM

This behavior is inconsistent, and in my opinion, confusing. There are two ways to make the behavior consistent:

(1) We could make all links function as mailto.
(2) We could make emails with User accounts link to those user pages, and emails with no user accounts display as plain text.

In my opinion, (2) is the slightly less pretty, but much more intuitive and helpful of these two options:

Screen Shot 2020-09-21 at 11 43 52 AM

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/orders/index.html.erb
+++ b/backend/app/views/spree/admin/orders/index.html.erb
@@ -184,8 +184,9 @@
           <% if order.user %>
             <%= link_to order.email, edit_admin_user_path(order.user) %>
           <% else %>
-            <%= mail_to order.email %>
+            <%= order.email %>
           <% end %>
+          <%= link_to_with_icon('email', t('spree.actions.send_email'), "mailto:#{order.email}", no_text: true) %>
         </td>
         <td class="align-right"><%= order.display_total.to_html %></td>
         <td class='actions align-center' data-hook="admin_orders_index_row_actions">
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -793,6 +793,7 @@ en:
       refund: Refund
       remove: Remove
       save: Save
+      send_email: Send Email
       ship: Ship
       split: Split
       update: Update


jarednorman:
I'm fine with this change, but I'm wondering if anyone actually cares about that mailto functionality. I doubt any of the stores I work with would mind this change, but I do want to be mindful of existing stores.

brchristian:
@jarednorman Sounds good to let folks weigh in if there are strong feelings here.

Note that you _can’t_ easily send mail to logged-in users who have placed orders, however! The links just go to their profiles instead! Which makes me suspect that few people rely on it for their workflow, but I’m totally open to dissenting perspectives if they are out there in the community.

aldesantis:
How about having the best of both worlds and using the current solution, but _also_ adding a mailto link next to the email (maybe in the form of an envelope icon)? I don't know if that would pollute the UI too much, but it certainly helps us keep the experience consistent without disrupting any existing workflows.

If it doesn't look good, I'm also okay with the current solution — as @brchristian noted, no one is probably relying on the functionality because of its inconsistence.

brchristian:
There's another issue here that just occurred to me, which is that there may be cases where `order.email != order.user.email`. Probably in most cases they will be the same, but I can think of at least two scenarios where they would diverge:

(1) It's an old order, and the user has since changed their email from what it was at the time they placed the order.

(2) It's an order purchased as a gift for someone else, and perhaps a store uses the `order.email` field to hold the _recipient's_ email, rather than the _sender's_ email.

In cases where  `order.email != order.user.email`, I'm not sure what the behavior should be here!

aldesantis:
I think we should always display `order.email`, since that's the email the user entered on the order (or the user had on their profile at the time they made the order).

Even if a user changes their email after placing the order, I think it would make sense to try the email on the order and then try contacting the user on the email on their profile if they're not replying or if the email doesn't exist anymore.

brchristian:
Okay, following the suggestions from @aldesantis, I have updated and force pushed. Here is the new behavior; it will always show `order.email`, will link the text to the user account if there is an `order.user`, and mailto links will appear to the right in both cases like so:

Screen Shot 2020-09-27 at 9 47 11 AM

I also gave the link a hover text like so:

Screen Shot 2020-09-27 at 9 47 18 AM

Let me know how this looks to you!

aldesantis (approved):

spaghetticode (approved):
@brchristian thank you 👍

kennyadsl (approved):
Looks great, thanks for taking care of this and for keep improving the Solidus Admin. ❤️

Issue #3766, opened 2020-09-21
DefaultTax Calculator considers invalid TaxRates when calculating tax for an item whose tax is included in price


jugglinghobo:
When calculating the tax for line items whose tax is included in the price, solidus uses TaxRates in the calculation even though they are invalid according to their `start_date` and/or `end_date`. This results in incorrect tax calculations if there are multiple tax rates for the same tax category and only some of them are currently valid.

**Solidus Version**:
`2.9.4` up to the current master version.

**To Reproduce**
1. In the admin backend, create a new TaxRate
2. Set the following attributes: Included in Price: true, Zone: North America, Tax Category: Default, name: "Active"
3. Create a second TaxRate with: Included in Price: true, Zone: North America, Tax Category: Default, name: "Inactive", end_date: sometime in the past
4. Delete the originally existing tax rate (to not interfere with this repro)

5. Go to the shop, add any product to cart and proceed to checkout
6. On the address page, enter a valid address for US
7. On the delivery page, you can now see the calculated tax.
8. As shown on the screenshot, this calculation includes the invalid tax rate

**Screenshots**
![Screenshot 2020-09-21 at 14 34 27](https://user-images.githubusercontent.com/814278/93772201-d6bab480-fc1e-11ea-9ba2-f29e2fd723a3.png)

**Current behavior**
The current calculation for line item tax that's included in the price is based on these methods (in `DefaultTax` calculator):

```
def deduced_total_by_rate(item, rate)
  round_to_two_places(
    rate.amount * item.total_before_tax / (1 + sum_of_included_tax_rates(item))
  )
end

def sum_of_included_tax_rates(item)
  rates_for_item(item).sum(&:amount)
end
```

`rates_for_item(item)` returns all tax rates whose category matches the categories for this item, and __does not consider whether this tax rate is valid or not__.

This leads to incorrect calculations, for example:

* item price: 119
* active tax rate amount: 0.16
* inactive tax rate amount: 0.19
* calculated tax: 14.103703704 = `0.16 * 119 / (1 + 0.16 + 0.19)`

Note how the inactive tax rate is part of the divisor in the above equation.

**Expected behavior**
The calculation for line item tax that's included in the price takes into account if a tax rate is valid or invalid, and calculates the tax only based on the valid tax rates.

Example for the correct calculation:

* item price: 119
* active tax rate amount: 0.16
* inactive tax rate amount: 0.19
[...]

PR #3768, opened 2020-09-22
TaxHelpers#rates_for_item now respects the validity period of tax rates

jugglinghobo:
**Description**
When calculating the tax for line items whose tax is included in the price, solidus uses TaxRates in the calculation even though they are invalid according to their start_date and/or end_date. This results in incorrect tax calculations if there are multiple tax rates for the same tax category and only some of them are currently valid.

This PR ensures that `TaxHelpers#rates_for_item` only returns tax rates that are currently valid. Before, it would return any tax rate that matched the items' tax category.

This addresses https://github.com/solidusio/solidus/issues/3766

**Checklist:**

### DIFF
--- a/core/spec/models/spree/tax/tax_helpers_spec.rb
+++ b/core/spec/models/spree/tax/tax_helpers_spec.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+class DummyClass
+  include Spree::Tax::TaxHelpers
+
+  def valid_rates(item)
+    rates_for_item(item)
+  end
+end
+
+RSpec.describe Spree::Tax::TaxHelpers do
+  let(:tax_category) { create(:tax_category) }
+  let(:irrelevant_tax_category) { create(:tax_category) }
+
+  let(:item) { create(:line_item, tax_category: tax_category) }
+  let(:tax_address) { item.order.tax_address }
+  
+  let(:zone) { create(:zone, name: "Country Zone", countries: [tax_address.country]) }
+
+  let!(:tax_rate) do
+    create(:tax_rate, tax_categories: [tax_category], zone: zone)
+  end
+
+  describe '#rates_for_item', :focus do
+
+    it 'returns tax rates that match the tax category of the given item' do
+      expect(DummyClass.new.valid_rates(item)).to contain_exactly(tax_rate)
+    end
+
+    context 'when multiple rates exist that are currently not valid' do
+      let(:starts_at) { 1.day.from_now }
+      let(:expires_at) { 2.days.from_now }
+
+      let!(:invalid_tax_rate) do
+        create(:tax_rate, tax_categories: [tax_category], zone: zone,
+               starts_at: starts_at, expires_at: expires_at)
+      end
+      
+      it 'returns only active rates that match the tax category of given item' do
+        expect(Spree::TaxRate.for_address(tax_address)).to contain_exactly(tax_rate, invalid_tax_rate)
+
+        expect(DummyClass.new.valid_rates(item)).to contain_exactly(tax_rate)
+      end
+    end
+  end
+end
--- a/core/spec/models/spree/tax/tax_helpers_spec.rb
+++ b/core/spec/models/spree/tax/tax_helpers_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe Spree::Tax::TaxHelpers do
     create(:tax_rate, tax_categories: [tax_category], zone: zone)
   end
 
-  describe '#rates_for_item', :focus do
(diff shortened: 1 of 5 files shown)


jugglinghobo:
As suggested in https://github.com/solidusio/solidus/issues/3766, I have opened a PR and moved the change to the `TaxHelpers#rates_for_item` method.

I am currently still unsure about two things:

* Maybe it would make sense to add an `.active` scope to `TaxRate`. If so, would you like me to do this in this PR directly?
* I didn't create a new spec file for the `TaxHelpers` module and instead added a test to the `DefaultTax` calculator which is one of two places affected by the change. Would you like me to create a spec file for the `TaxHelpers` module? If so, where should it be (since `TaxHelpers` is in `models/` for some reason...?).

I'm very open for suggestions and improvements.

aldesantis:
@jugglinghobo thanks for this! Answers to your concerns below:

> * Maybe it would make sense to add an `.active` scope to `TaxRate`. If so, would you like me to do this in this PR directly?

Yes, I think it would be useful. 🙂  Feel free to do it in this PR!

> * I didn't create a new spec file for the `TaxHelpers` module and instead added a test to the `DefaultTax` calculator which is one of two places affected by the change. Would you like me to create a spec file for the `TaxHelpers` module? If so, where should it be (since `TaxHelpers` is in `models/` for some reason...?).

Yes, I think the test for this should be on the `TaxHelpers` module itself. You can put it in `spec/models/spree/tax/tax_helpers_spec.rb`. Testing modules can be a bit tricky (you'd usually create a "dummy" test class and include the module there, then test the method somehow), so let me know if you need any help.

mamhoff:
@aldesantis @jugglinghobo I think this is actually a pretty important fix functionality-wise. I appreciate wanting to have very good specs that respect all the things, and maybe also a scope, but can we make issues for that and get this in? I'm seeing clients write workarounds in their exporter code for tax advisors because this doesn't work. I'd also be happy to take this over if necessary. (cc @tvdeyen @mntmn).

jugglinghobo:
@mamhoff @aldesantis I'm sorry this fell off my radar. If you can wait, I'll try to finish this tonight or tomorrow. If you don't hear back until sunday or if you need a solution quicker, feel free to take over, and if you leave out anything (e.g. scope), feel free to assign an issue to me or whatever you'd like.

Does that work for you, or would you like me to do anything else? This is my first solidus / open source contribution.

aldesantis:
@jugglinghobo even if this is merged today, it still wouldn't be available until the next Solidus release, so tonight/tomorrow works perfectly. Thanks for your help on this one!

mamhoff:
I just didn't want this to sit for another couple of weeks. If you're on it, great! :+1:

jugglinghobo:
@aldesantis 

As discussed I've added a spec for `TaxHelpers#rates_for_item`. IMO this is all that's required to fix the bug.
I also added the `TaxRate.active` scope (identical to the one in `Promotion`), but it is unused at the moment.

Please suggest improvements and advise on how to proceed. I didn't squash yet because I thought maybe you'd like to add the scope in a separate PR.

PS: AFAICT there are no specs for `Promotion.active`, if the new specs are accepted then these could be reused there.

core/spec/models/spree/tax/tax_helpers_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+require 'rails_helper'
+
+class DummyClass
+  include Spree::Tax::TaxHelpers
+
+  def valid_rates(item)
+    rates_for_item(item)
+  end
+end
```
aldesantis, line 11: Instead of defining the class outside of the tests, can you do with `before` and `stub_const`?
```ruby
before do
  stub_const('DummyClass', (Class.new do
    include Spree::Tax::TaxHelpers

    def valid_rates(item)
      rates_for_item(item)
    end
  end))
end
```
The reason is that a class defined with the `class` keyword will leak between tests, while one defined with `stub_const` will be unloaded as soon as the test has finished executing.

aldesantis (changes requested):
@jugglinghobo thanks, the tests are okay except for one minor thing. After that, GTM!

core/spec/models/spree/tax/tax_helpers_spec.rb
jugglinghobo: Thanks for the feedback, I've pushed the requested changes.

aldesantis (approved):
Looks great, thanks so much! 🙏

mamhoff (approved):
Nicely done!

kennyadsl (approved):
Thanks!

jugglinghobo:
Thank you guys!

PR #3772, opened 2020-09-24
Make admin order event_links translatable

tvdeyen:
**Description**

The admin/orders_helper event_links didn't had enough scope to be translatable for the particular order event. In some languages canceling and order is something different than clicking a "cancel" button in a form.

The solidus_i18n gem already has the translation scopes for these events, so I implemeted it that it will Just Work™

**Checklist:**

### DIFF
--- a/backend/spec/helpers/admin/orders_helper_spec.rb
+++ b/backend/spec/helpers/admin/orders_helper_spec.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe Spree::Admin::OrdersHelper, type: :helper do
+  describe "#event_links" do
+    subject { helper.event_links }
+
+    before do
+      helper.class.include Spree::Admin::NavigationHelper
+      helper.class.include Spree::Core::Engine.routes.url_helpers
+      @order_events = %w{approve cancel resume}
+    end
+
+    context "with an uncompleted order" do
+      before do
+        @order = create(:order)
+      end
+
+      it "renders link to approve order" do
+        is_expected.to have_button("Approve")
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


kennyadsl:
@tvdeyen I think the specs are related to the change but it probably just needs a capitalization.

tvdeyen:
@kennyadsl yes, will fix them soon

tvdeyen:
@kennyadsl should now be fixed

backend/spec/helpers/admin/orders_helper_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      @order_events = %w{approve cancel resume}
+    end
+
+    context "with an uncompleted order" do
+      before do
+        @order = create(:order)
+      end
+
+      it "renders link to approve order" do
+        is_expected.to have_button("Approve")
```
aldesantis, line 21: Should we call `I18n.t` in these tests, so that we don't have to update them if we change the translations? I think that's unlikely to happen, just wondering if it's better to do it that way from the beginning.

aldesantis (approved):
@tvdeyen thanks, looking great. Left a non-blocking comment but I'm going to approve.

backend/spec/helpers/admin/orders_helper_spec.rb
tvdeyen: I am no fan of computed values for test expectations. They can provide more problems than they seem to solve.

aldesantis: Agree that it's a gray area. Nevermind! 🙂

kennyadsl (approved):
Thanks Thomas!

Issue #3775, opened 2020-09-25
Default payment method sometimes fails to install


seand7565:
When installing Solidus using the generator, I got this error:

```shell
  You can select a payment method to be included in the installation process.
  Please select a payment method name: [paypal, none] (paypal) paypal
     gemfile  solidus_paypal_commerce_platform
      append  db/seeds.rb
     copying  migrations
Traceback (most recent call last):
	17: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
	16: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
	15: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/setup.rb:20:in `<top (required)>'
	14: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/ui/shell.rb:88:in `silence'
	13: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/ui/shell.rb:136:in `with_level'
	12: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/setup.rb:20:in `block in <top (required)>'
	11: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler.rb:149:in `setup'
	10: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/runtime.rb:20:in `setup'
	 9: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/runtime.rb:101:in `block in definition_method'
	 8: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/definition.rb:226:in `requested_specs'
	 7: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/definition.rb:237:in `specs_for'
	 6: from /Users/seandenny/.rvm/rubies/ruby-2.7.0/lib/ruby/2.7.0/bundler/definition.rb:170:in `specs'
[...]

PR #3743, opened 2020-08-26
Replace --auto-accept with --interactive

elia:
**Description**

The current solidus install generator will ask for a number of things unless `--auto-accept` is provided. Since many of the things asked are not simple yes/no questions, and by default Rails doesn't require any interaction when generating a new app, we should move to an opt-in `--interactive` flag.

The current PR only switches to the new CLI option and cleans up the generator to make it more maintainable by solidus core and usable by end-users.

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
- [ ] I have updated Guides and README accordingly to this change (if needed)
- [ ] I have added tests to cover this change (if needed)
- [ ] I have attached screenshots to this PR for visual changes (if needed)

PR #3777, opened 2020-09-25
Fix the install generator

seand7565:
ref #3775

Fixes the issues with the install generator by bundling after adding the
extensions to the gemfile, stopping spring, and then running the install
generators for each of the installed extensions.

Currently this works a bit wonky as the extensions will ask if you want
to run migrations, and that's not desirable as migrations are already run
after these extensions are installed. Will fix and update shortly, just
wanted to get this into a PR!

**Checklist:**

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -111,14 +111,20 @@ module Solidus
       end
     end
 
-    def install_default_plugins
+    def plugin_install_preparation
+      @plugins_to_be_installed = []
+      @plugin_generators_to_run = []
+    end
+
+    def install_auth_plugin
       if options[:with_authentication] && (options[:auto_accept] || !no?("
-  Solidus has a default authentication extension that uses Devise.
-  You can find more info at https://github.com/solidusio/solidus_auth_devise.
+        Solidus has a default authentication extension that uses Devise.
+        You can find more info at https://github.com/solidusio/solidus_auth_devise.
 
-  Would you like to install it? (y/n)"))
+        Would you like to install it? (y/n)"))
 
-        gem 'solidus_auth_devise'
+        @plugins_to_be_installed << 'solidus_auth_devise'
+        @plugin_generators_to_run << 'solidus:auth:install'
       end
     end
 
@@ -135,7 +141,10 @@ module Solidus
 
       gem_name = PAYMENT_METHODS.fetch(name)
 
-      gem gem_name if gem_name
+      if gem_name
+        @plugins_to_be_installed << gem_name
+        @plugin_generators_to_run << "#{gem_name}:install"
+      end
     end
 
     def include_seed_data
@@ -154,6 +163,19 @@ module Solidus
     def create_database
       say_status :creating, "database"
       rake 'db:create'
+    end
+
+    def run_bundle_install_if_needed_by_plugins
+      @plugins_to_be_installed.each do |plugin_name|
+        gem plugin_name
+      end
+
+      bundle_cleanly{ run "bundle install" } if @plugins_to_be_installed.any?
+      run "spring stop"
+
+      @plugin_generators_to_run.each do |plugin_generator_name|
+        generate "#{plugin_generator_name} --skip_migrations=false"
+      end
     end
 
     def run_migrations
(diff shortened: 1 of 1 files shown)


seand7565:
Currently, it's impossible to run extension installers 
A. without running migrations and
B. without being asked whether or not you want to run migrations

Ideally, with the overall Solidus installer, the user would be able to just run it and have a working solidus store, while installing the default extensions. Currently that's not possible.

The two solutions I can think of are:
1. Running the install generators separately. In previous versions of Solidus, we had to run the install generator for solidus_auth_devise separately. We could go back to that functionality - although it gets weird when we have multiple payment methods able to be installed.

2. Change the way we're migrating in extension installers.

Here's an example of an extension installer - https://github.com/solidusio/solidus_auth_devise/blob/8b65779f02a4fe46d81bbf0c78347b39f8f93205/lib/generators/solidus/auth/install/install_generator.rb#L23

If you pass in `--auto-run-migrations=false` it will still ask you if you want to run migrations. We _could_ update that to a `run-migrations` flag and if it's not explicitly set, then ask if they want to run migrations. We could also just default to automatically running migrations and have a flag that turns that functionality off. I think the use case of not running migrations after installation is pretty small (pretty much just useful in the context of this PR AFAIK).

However that requires making changes to a lot of extensions, and to solidus_dev_support, which is a lot of work. We'd also have to create new releases for some of these extensions after doing that. (though solidus_auth_devise needs a new release anyways).

What do you think? @kennyadsl @elia

kennyadsl:
What if you run:

`run "rails generate #{plugin_generator_name} --auto-run-migrations=false"`

instead of:

`generate "#{plugin_generator_name} --auto-run-migrations=false"`

Will the message be prompted anyway?

kennyadsl:
Also, I think the [generate](https://edgeguides.rubyonrails.org/generators.html#generate) syntax is wrong. Should be something like:

```
generate "#{plugin_generator_name}", "--auto-run-migrations=false"
``

**Comment by seand7565** [comment]
Unfortunately the migration question gets asked regardless of syntax - I tried both ways. Thanks for pointing out the generate syntax!

The problem is this line:
```ruby
run_migrations = options[:auto_run_migrations] || ['', 'y', 'Y'].include?(ask('Would you like to run the migrations now? [Y/n]'))
```
if you set auto_run_migrations to false, then the question will always be asked.

seand7565:
Maybe it's not that big of a deal though, since there are several other prompts during installation. It is a bit weird to ask to migrate 3 times during the initial installation, but probably easier to keep it that way than change every extension that uses that line. :P

elia:
@seand7565 I think it's worth fixing it, not for the seasoned Solidus dev, but for everyone trying to install it for the first time, who I suspect would be rather confused by that.

I also hope to come back to https://github.com/solidusio/solidus/pull/3743 and finish it to further improve the usability of the installer.

kennyadsl:
@seand7565 got it. If we don't want to auto-run migrations the message will be prompted. 

I'd start modifying these two extensions for now, adding another parameter that force not asking and not migrating. Maybe we can move this shared logic into a `solidus_dev_support` module that we can include in both and gradually adopt in other extensions when needed?

seand7565:
> @seand7565 got it. If we don't want to auto-run migrations the message will be prompted.
> 
> I'd start modifying these two extensions for now, adding another parameter that force not asking and not migrating. Maybe we can move this shared logic into a `solidus_dev_support` module that we can include in both and gradually adopt in other extensions when needed?

Yeah, let's do that! I have a PR on SPCP that handles this change, take a look if you get a chance. If it looks good, I'll also port it into solidus_devise - at which point this PR should be mergeable (though we'll need a re-release of both SPCP and solidus_devise) https://github.com/solidusio-contrib/solidus_paypal_commerce_platform/pull/94

Then we can add it into solidus_dev_support so any new extensions are built with this change. 

I opted for a `skip-migrations` option because I think it makes the most sense here - I thought copying the new syntax from #3743 might be acceptable too, but it would still run the migrations 3 times, and I'd prefer to skip the migrations entirely until the final migration happens as part of the apps setup.

seand7565:
Here's the change in solidus_auth_devise: https://github.com/solidusio/solidus_auth_devise/pull/194

seand7565:
The devise PR needs to be merged, then a new release cut.
The SPCP PR just needs a new release.

But other than that, this PR is good to go, so I'll un-draft it. Once a new release is cut for both extensions, this should fix the installers issues.

kennyadsl:
@seand7565 I just released `solidus_auth_devise` with the installer enhancement. Can you take care of the SPCP one or do you need my assistance there?

seand7565:
@kennyadsl It's been released! 🎉 This PR should be good to go. I'll give it a quick rebase first though.

kennyadsl (approved):

jarednorman (approved):

spaghetticode (approved):

PR #3781, opened 2020-09-28
Update order cancellation, sets inventory units to canceled

### LINKED ISSUE #3719: Inventory units stay `on_hand` after an order is canceled

When I cancel and order, the inventory units remain `on_hand`
```
# Canceled order
order.inventory_units.pluck(:state) => ["on_hand"]
```
**Steps to recreate**
- Create an order and complete it
- Cancel the order by clicking the "Cancel" button
- Notice that the order state is `canceled`, the shipment state it `canceled`, but the inventory units states remain `on_hand` (see screenshot below)

When I cancel an order, the inventory units state should also change to `canceled`
```
# Desired return for canceled order
order.inventory_units.pluck(:state) => ["canceled"]`
```

**Additional context**

The issue we've specifically had has been with our warehouse system. The inventory units _have_ to be canceled so our CSRs have to cancel and order, _and_ cancel all the items under the "Cancel Items" tab. But it does seem strange that when you cancel an order the inventory units are not also canceled. We set the shipment state to cancelled so I would assume we do the same for the inventory units. 
cancelled-order

stefano-sarioli:
This PR fix #3719.

Before this commit when an order was canceled the inventory units
of the order will remain untouched. So there were canceled orders with
inventory units with the state on_hand or backordered.

This situation was not only inconsistent but can lead to subtle bugs
with the stock management system.
How to reproduce the bug: 

- Create an order for a single variant, but select a number of items greater than the available amount.
- The order generates 2 shipments, one with the inventory units on_hand and one with the inventory units back-ordered.
- Cancel the order.
- The inventory units on_hand remains on_hand.
- The inventory units back-ordered go to the on_hand state. 
- If you resume the order all the inventory units will be on_hand, and the shipment will be ready to be shipped.

This problem is caused by the fact that when the order is canceled, a movement is created to restock the stock_item correlated with the inventory_units. As soon as the stock_items get back to a positive value it will try to convert as much as possible back-ordered inventory units to on_hand.

order_complete_before
order_canceled_before
order_resumed_before

With this commit the inventory units related to an order are canceled when the order is canceled.

When an order is resumed we recalculate the state of the connected
inventory units, based on the actual stocks available.

order_complete_after
order_canceled_after
order_resumed_after

What do you think about this issue, and my solution?
I'm not sure that the testing is enough, probably I can add an E2E test, what do you think? any idea on what can be the best place to put this test?

**Checklist:**

### DIFF
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -76,9 +76,11 @@ module Spree
 
     def after_cancel
       manifest.each { |item| manifest_restock(item) }
+      cancel_inventory_units
     end
 
     def after_resume
+      resume_inventory_units
       manifest.each { |item| manifest_unstock(item) }
     end
 
@@ -384,6 +386,33 @@ module Spree
 
     def manifest_unstock(item)
       stock_location.unstock item.variant, item.quantity, self
+    end
+
+    def cancel_inventory_units
+      inventory_units.where.not(state: 'canceled').each(&:cancel!)
+    end
+
+    def resume_inventory_units
+      manifest.each do |item|
+        number_of_items_to_restore = item.states["on_hand"].to_i + item.states["backordered"].to_i
+
+        if item.variant.should_track_inventory?
+          units_available, units_to_back_order = stock_location.fill_status(item.variant, number_of_items_to_restore)
+
+          inventory_units.canceled.
+            where(line_item: item.line_item).
+            limit(units_available).
+            each(&:on_hand!)
+
+          inventory_units.canceled.
+            where(line_item: item.line_item).
+            limit(units_to_back_order).
+            each(&:backorder!)
+        else
+          inventory_units.canceled.limit(number_of_items_to_restore).each(&:on_hand!)
+        end
+      end
+      inventory_units.reload
     end
 
     def set_cost_zero_when_nil
--- a/core/lib/spree/core/state_machines/inventory_unit.rb
+++ b/core/lib/spree/core/state_machines/inventory_unit.rb
@@ -34,6 +34,14 @@ module Spree
             event :cancel do
               transition to: :canceled, from: ::Spree::InventoryUnit::CANCELABLE_STATES.map(&:to_sym)
             end
+
+            event :on_hand do
+              transition to: :on_hand, from: :canceled
+            end
+
+            event :backorder do
+              transition to: :backordered, from: :canceled
+            end
           end
         end
       end
--- a/core/spec/models/spree/shipment_spec.rb
+++ b/core/spec/models/spree/shipment_spec.rb
@@ -432,6 +432,18 @@ RSpec.describe Spree::Shipment, type: :model do
       shipment.after_cancel
     end
 
+    it 'sets inventory unit to canceled' do
+      shipment.stock_location = mock_model(Spree::StockLocation)
+      allow(shipment.stock_location).to receive(:restock)
+      allow(shipment.stock_location).to receive(:fill_status)
+      inventory_unit = shipment.inventory_units.first
+
+      shipment.cancel!
+
+      inventory_unit.reload
+      expect(inventory_unit).to be_canceled
+    end
+
     context "with backordered inventory units" do
       let(:order) { create(:order) }
       let(:variant) { create(:variant) }
@@ -462,9 +474,7 @@ RSpec.describe Spree::Shipment, type: :model do
   end
 
   context "#resume" do
-    let(:inventory_unit) { create(:inventory_unit) }
-
-    before { shipment.state = 'canceled' }
+    before { shipment.cancel! }
 
     context "when order cannot ship" do
       before { allow(order).to receive_messages(can_ship?: false) }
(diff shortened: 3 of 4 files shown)


core/app/models/spree/shipment.rb
```diff
@@ -386,6 +388,21 @@ def manifest_unstock(item)
       stock_location.unstock item.variant, item.quantity, self
     end
 
+    def resume_inventory_units
+      manifest.each do |item|
+        if item.variant.should_track_inventory?
+          units_available, units_to_back_order = stock_location.fill_status(item.variant, item.quantity)
+
+          inventory_units.canceled.limit(units_available).each(&:on_hand!)
+          inventory_units.canceled.limit(units_to_back_order).each(&:backorder!)
```
kennyadsl, line 397: Is canceling an order the only way to cancel an IU at the moment? Just wondering if other IUs can have the state canceled and they would be resumed, even if not part of the order cancelation.

kennyadsl (commented):
Left a preliminary concern, also some specs seem to be failing.

core/app/models/spree/shipment.rb
stefano-sarioli: No, you are right, there are other places.
I have to pick only the canceled units for the variant reported on the manifest items.

stefano-sarioli: I've used the line_item to be sure.

stefano-sarioli: BTW I'm double-checking this to be sure. Sorry for the message spam 🙇 .

core/app/models/spree/shipment.rb
```diff
@@ -386,6 +388,33 @@ def manifest_unstock(item)
...
+    def cancel_inventory_units
+      inventory_units.where.not(state: 'canceled').each(&:cancel!)
+    end
+
+    def resume_inventory_units
+      manifest.each do |item|
+        number_of_items_to_restore = item.states["on_hand"].to_i + item.states["backordered"].to_i
+
+        if item.variant.should_track_inventory?
+          units_available, units_to_back_order = stock_location.fill_status(item.variant, number_of_items_to_restore)
```
kennyadsl, line 400: Are these variables filled with the current availability in the stock location, right?

core/spec/models/spree/shipment_spec.rb
```diff
@@ -476,38 +486,78 @@
...
-      before { allow_any_instance_of(Spree::InventoryUnit).to receive(:allow_ship?).and_return(false) }
-      it "should result in a 'ready' state" do
+      before { allow(shipment.stock_location).to receive(:fill_status).and_return([0, 1]) }
+      it "should result in a 'pending' state" do
         shipment.resume!
         expect(shipment.state).to eq 'pending'
       end
     end
 
+    context "when any inventory was already canceled" do
```
kennyadsl, line 503: Thanks for adding this spec!

kennyadsl (commented):
Just left a question, overall I can't see any issues with this approach.

core/spec/models/spree/shipment_spec.rb
stefano-sarioli: Thanks to you for pointing out the edge case 🙇, I totally missed it!

core/app/models/spree/shipment.rb
stefano-sarioli: Exactly, this will update the inventory units state based on the current stock level.

This is why inside #after_resume I have to call #resume_inventory_units before unstocking the items. If you call #resume_inventory_units after the unstocking the states of the inventory units will be wrong. 

There is a test covering this use-case to avoid regression.

kennyadsl (approved):
Thanks @stefano-sarioli !

core/lib/spree/core/state_machines/inventory_unit.rb
```diff
@@ -34,6 +34,14 @@ module InventoryUnit
             event :cancel do
               transition to: :canceled, from: ::Spree::InventoryUnit::CANCELABLE_STATES.map(&:to_sym)
             end
+
+            event :on_hand do
```
spaghetticode, line 38: I'm not 100% sure about this event name. 

Semantically speaking, event names should be verbs, as they actively change the state of an object. In this case, the action name is the same as the target state, and this may lead to some confusion too.  
Also, there's the query method #can_on_hand? added automatically, which again does not make 100% sense to me. 

What about changing the name to something like `set_on_hand` or `make_on_hand`? And if you find a better naming, just go for it.

core/app/models/spree/shipment.rb
```diff
@@ -386,6 +388,33 @@ def manifest_unstock(item)
       stock_location.unstock item.variant, item.quantity, self
     end
 
+    def cancel_inventory_units
+      inventory_units.where.not(state: 'canceled').each(&:cancel!)
+    end
+
+    def resume_inventory_units
```
kennyadsl, line 395: We discussed this within the core team. We think that resuming concerns should be moved into a separate PR. There are a lot of opinions on how resuming an order should work and probably none of them will work as the majority of stores expect (everyone wants to restock items in different ways, depending on if they are using Solidus to manage stocks, which itself has a lot of variables, or they are relying on an external system). 

We are thinking to actually deprecate the Resume order functionality. It's not used a lot and gives us more headaches than the benefits it provides to stores (none of us can think of a single store that ever used this functionality). The way to go here is moving the current behavior into a legacy extension, and deprecate the functionality in core (removing the Resume button from the admin UI). WDYT?

stefano-sarioli: I agree with you, especially if it's an underused functionality. 
My solution is far too general and makes a lot of assumptions about the stock management system.

I think that extracting the resume functionality into an external gem, maybe with this fix on it, would be the best solution.

Probably we can close this PR, I'll wait for the final decision about the resume functionality and will come back to contribute to that one as soon as it's ready.

Thank a lot for your feedback and consideration 🙏.

kennyadsl: The cancellation part of this PR is well done and something that we definitely need in core. Please don't close the PR but just remove the part that resumes the inventory units. 🙏

core/lib/spree/core/state_machines/inventory_unit.rb
stefano-sarioli: I agree with you, I was not convinced too by the name of that event but couldn't think of something better. 

I like `set_on_hand` proposed by you ❤️.

Sadly, as pointed out in the last comment of Kennyadsl, I'm probably going to close this PR to move the whole resume functionality out of Solidus.

Thanks a lot for your feedback, I'll note it down to use it when the work on the extracted functionality will start 🙏.

PR #3782, opened 2020-10-01
Get all absolute adjustments on an order at once

duartemvix:
This is a small change that I hope can stay with Solidus for a long time. It simplifies the call at the `orders/_form.html.erb` and `order.rb` to use one straight call to the database in order to get adjustments.

I haven't tested in other databases except for Postgres. I also could not be able to run tests on my own and I hope there's already tests written for editing orders that I could quickly run through.

Visual changes in pictures:

Captura de tela de 2020-09-30 22-27-47

and

Captura de tela de 2020-09-30 22-28-15


core/app/models/spree/order.rb
```diff
@@ -190,6 +190,12 @@ def self.not_canceled
       where.not(state: 'canceled')
     end
 
+    # => Get all nonzero (reference to :nonzero scope of Spree::Adjustment) adjustments for # => an instance of Spree::Order and its LineItem and Shipment associations
+    def absolute_adjustments
+      Spree::Adjustment.includes(:order).where(order: self, adjustable_type: ['Spree::LineItem', 'Spree::Shipment'])
+      .where.not(amount: 0).or(Spree::Adjustment.includes(:order).where(adjustable: self).where.not(amount: 0))
+    end
```
jarednorman, line 197: I believe that the we can do this without using `or` by relying on the `all_adjustments` association on the order. We can also use the existing scope that you reference in the comment. Isn't this equivalent? In what situations is it not?

```suggestion
    def absolute_adjustments
      all_adjustments.includes(:order).nonzero
    end
```

jarednorman (changes requested):

core/app/models/spree/order.rb
duartemvix: If you look closely the all_adjustments association of `Spree::Orders` doesn't contemplate adjustments that have the `Spree::LineItem` and `Spree::Shipment` associations of `Spree::Order` as their adjustable target model. These different adjustments are called at `<% if order.adjustments.nonzero.exists? || order.line_item_adjustments.nonzero.exists? || order.shipment_adjustments.nonzero.exists? || order.shipments.any? %>` 

What `.or` does in this circunstance is that it replaces the logical union of the 3 different `Exists?` queries and makes just one of them on the database rather than in Ruby. Before `.or` you have all adjustments to `LineItem` and `Shipment` that have an `order_id` of the current order. But that order is not the adjustable object of the query.

duartemvix:
```ruby
# => Get all nonzero (reference to :nonzero scope of Spree::Adjustment) adjustments for an instance of Spree::Order and its LineItem and Shipment associations
def absolute_adjustments
  all_adjustments.nonzero.or(Spree::Adjustment.where(adjustable: order).nonzero)
end
```

I thought of something that accomplishes a similar thing with just 1 call rather than 3 as it was before. I'm pushing to see test results.

core/app/models/spree/order.rb
```diff
@@ -190,6 +190,11 @@ def self.not_canceled
       where.not(state: 'canceled')
     end
 
+    # => Get all nonzero (reference to :nonzero scope of Spree::Adjustment) adjustments for an instance of Spree::Order and its LineItem and Shipment associations
+    def absolute_adjustments
+      all_adjustments.nonzero.or(Spree::Adjustment.where(adjustable: self).nonzero)
```
jarednorman, line 195: I'm having trouble understanding in what situation is the `order_id` not set to the current order on an order-level adjustment. `all_adjustments` uses the `order_id` foreign key, so it should return all the adjustments associated with the order, regardless of whether the adjustable is the order itself or a shipment/line item.

jarednorman: I'm also concerned about naming here. Here are my concerns:

1. I think this method should probably be called `nonzero_adjustments`. An "absolute" value of x is ["the non-negative value of x without regard to its sign"](https://en.wikipedia.org/wiki/Absolute_value) which isn't really a good description of this. These are just "non-zero" adjustments. We already use the term `nonzero` to describe them elsewhere so I would need to see a good argument for introducing a second name of `nonzero` to the codebase.
2. To be consistent with the naming of the associations (`adjustments` are the order-level adjustments, `all_adjustments` are all the adjustments associated with the order in any way) this method should probably be called `all_nonzero_adjustments`.
3. If we were to call this method `all_nonzero_adjusments`, there's not much value in this method. To me at least, `order.all_nonzero_adjustments` isn't any easier to write or remember than `order.all_adjustments.nonzero`.

jarednorman (changes requested):
I don't think this method adds very much, so I'd personally prefer we just call `order.all_adjustments.nonzero.exists?` or `order.all_adjustments.nonzero.any?` (depending on if the association is loaded) in the view. I'm open to hearing why we might need a method for this though.

core/app/models/spree/order.rb
duartemvix: > I'm having trouble understanding in what situation is the `order_id` not set to the current order on an order-level adjustment. `all_adjustments` uses the `order_id` foreign key, so it should return all the adjustments associated with the order, regardless of whether the adjustable is the order itself or a shipment/line item.

That's not wrong. The call to the method could be replaced by `order.all_adjustments.nonzero.any?`. However, when I wrote this I was thinking of passing the same arguments to the query that Solidus was already doing and that's why there's an `.or` uniting already existent scopes.

I'll change locally and see if it yields the same results.

jarednorman (changes requested):
Cool, looks good to me. I think the only thing this needs for a 👍 from me is for the commits to get squashed down to a single commit that makes the final change.

PR #3783, opened 2020-10-01
Show the admin/settings menu for any of its elements

elia:
**Description**

It's possible to create permission sets the only give access to a
subset of the settings menu items, without this the settings menu was
kept hidden.

**Checklist:**

### DIFF
--- a/backend/spec/lib/spree/backend_configuration_spec.rb
+++ b/backend/spec/lib/spree/backend_configuration_spec.rb
@@ -18,5 +18,26 @@ RSpec.describe Spree::BackendConfiguration do
         expect(stock_menu_item.match_path).to eq('/stock_items')
       end
     end
+
+    describe 'menu tab for settings' do
+      let(:menu_item) { subject.find { |item| item.label == :settings } }
+      let(:view) { double("view") }
+
+      it 'is shown if any of its submenus are present' do
+        allow(view).to receive(:can?).and_return(true, false)
+
+        result = view.instance_exec(&menu_item.condition)
+
+        expect(result).to eq(true)
+      end
+
+      it 'is not shown if none of its submenus are present' do
+        expect(view).to receive(:can?).exactly(12).times.and_return(false)
+
+        result = view.instance_exec(&menu_item.condition)
+
+        expect(result).to eq(false)
+      end
+    end
   end
 end
--- a/backend/lib/spree/backend_configuration.rb
+++ b/backend/lib/spree/backend_configuration.rb
@@ -131,7 +131,20 @@ module Spree
         MenuItem.new(
           CONFIGURATION_TABS,
           'wrench',
-          condition: -> { can?(:admin, Spree::Store) },
+          condition: -> {
+            can?(:admin, Spree::Store) ||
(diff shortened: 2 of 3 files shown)


backend/spec/lib/spree/backend_configuration_spec.rb
```diff
@@ -18,5 +18,17 @@
...
+
+    describe 'menu tab for settings' do
+      let(:menu_item) do
+        subject.find { |item| item.label == :settings }
+      end
+
+      it 'is shown if any of its submenus are present' do
+        view = double("view", can?: false)
+        expect(view).to receive(:can?).at_least(12).times
+        view.instance_exec(&menu_item.condition)
```
aldesantis, line 30: Wouldn't it be better to test the result of `view.instance_exec(&menu_item.condition)` in a couple of different scenarios? With your current test, one could change all the `||` to `&&` and still have the test pass.

aldesantis (changes requested):

backend/spec/lib/spree/backend_configuration_spec.rb
elia: Of course, this is a trade-off between simplicity and completeness. Although I think the change to `&&` would make the test fail (`can?` would be called just once), I see the point, I'm just not sure how much more detail we want to put in there, given the complete matrix in all its combinations is probably too much.

I think the real fix would be to redesign the admin menu items to have sub-items wrapped in objects and check if any of those is `#visible?`, but this is clearly beyond the scope of this PR. 

**That'd be fine if I opened an issue describing the underlying problem?**

aldesantis: > Although I think the change to && would make the test fail (can? would be called just once)
Right, my bad. 🤦 

Anyway, I agree that we need to find a compromise, but I feel like we could do a bit better without writing too much code.

I think a good compromise could be just two tests such as the following:
```ruby
it 'is shown if any of its submenus are present' do
  view = double("view")
  allow(view).to receive(:can?).and_return(true, false)

  result = view.instance_exec(&menu_item.condition)

  expect(result).to eq(true)
end

it 'is not shown if none of its submenus are present' do
  view = double("view", can?: false)

  result = view.instance_exec(&menu_item.condition)

  expect(result).to eq(false)
end
```
The main benefits over the existing tests, in my opinion, are:

1. You're testing the underlying logic more directly. Rather than setting an expectation on how many times `can?` is called, you're setting it on whether the condition works properly.
2. The tests are more expressive. Reading your test, it's unclear how the underlying implementation works. With this version, it's a bit clearer what's going on in the implementation.
3. The tests follow a standard setup/exercise/verify pattern and are easier for a human to parse.

Let me know your thoughts!

aldesantis (approved):

kennyadsl (approved):
Thanks @elia!

backend/spec/lib/spree/backend_configuration_spec.rb
elia: Thanks for the suggestion, it was useful!
The issue is here https://github.com/solidusio/solidus/issues/3785

PR #3786, opened 2020-10-04
Update backend New Image link for consistency

brchristian:
This makes the New Image link consistent with visual style of other backend views.

Other Views:

Screen Shot 2020-10-03 at 5 31 58 PM

Screen Shot 2020-10-03 at 5 32 07 PM

Current Behavior:

Screen Shot 2020-10-03 at 5 31 40 PM

New Behavior:

Screen Shot 2020-10-03 at 5 33 07 PM

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/images/index.html.erb
+++ b/backend/app/views/spree/admin/images/index.html.erb
@@ -5,7 +5,9 @@
 
 <% content_for :page_actions do %>
   <% if can?(:create, Spree::Image) %>
-    <li><%= link_to_with_icon('plus', t('spree.new_image'), new_admin_product_image_url(@product), id: 'new_image_link', class: 'btn btn-primary') %></li>
+    <li>
+      <%= link_to t('spree.new_image'), new_admin_product_image_url(@product), id: 'new_image_link', class: 'btn btn-primary' %>
+    </li>
   <% end %>
 <% end %>
 


aldesantis (approved):
Thanks @brchristian!

backend/app/views/spree/admin/images/index.html.erb
```diff
@@ -5,7 +5,9 @@
 
 <% content_for :page_actions do %>
   <% if can?(:create, Spree::Image) %>
-    <li><%= link_to_with_icon('plus', t('spree.new_image'), new_admin_product_image_url(@product), id: 'new_image_link', class: 'btn btn-primary') %></li>
+    <li id="new_image_link">
+      <%= link_to t('spree.new_image'), new_admin_product_image_url(@product), id: 'new_image_link', class: 'btn btn-primary' %>
```
jarednorman, line 9: Looks like you've applied the `new_image_link` ID to both the `li` and the `a` elements. Since IDs should be unique and it's conceivable that someone could be using deface overrides to target this, I think it's best to leave the ID on the `a` and not also add it to the `li`.

jarednorman (changes requested):

backend/app/views/spree/admin/images/index.html.erb
brchristian: @jarednorman Good catch!

Looking around at other backend views, it seems like in https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/prices/index.html.erb#L5-L7 and https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/variants/index.html.erb#L7-L13, the convention is to put the ID on the `li` element. I’d suggest that we do the same here for consistency with those other views.

I’ve force-pushed the latest; LMK what you think.

kennyadsl (approved):
Thanks @brchristian !

brchristian:
Ahh, actually there is more to the story here, because there are both specs and JS that expect the ID to live on the `a` element:

https://github.com/solidusio/solidus/blob/10a119940eca56953c6af92b63150359199f24bf/backend/spec/features/admin/products/edit/images_spec.rb#L43

https://github.com/solidusio/solidus/blob/10a119940eca56953c6af92b63150359199f24bf/backend/app/assets/javascripts/spree/backend/images/index.js#L2

For now, I have reversed my earlier decision and have force-pushed a version that puts the ID on the `a` and not the `li`. This is inconsistent with admin/prices.index and admin/variants/index, which use the `li`:

https://github.com/solidusio/solidus/blob/10a119940eca56953c6af92b63150359199f24bf/backend/app/views/spree/admin/prices/index.html.erb#L5-L7

https://github.com/solidusio/solidus/blob/10a119940eca56953c6af92b63150359199f24bf/backend/app/views/spree/admin/variants/index.html.erb#L7-L13

...so my OCD is triggered, but on the other hand, putting the ID on the `a` element: (1) doesn't change from the current behavior, which already has the ID on the `a` element, and (2) doesn't require changing the specs or JS.

Probably a more principled solution would be to move the `"new_image_link"` ID to the `li` for consistency with other backend views, but to then come up with some other unique ID to put on the `a` element and have _that_ ID in the spec and JS. However I'm not exactly sure what that other ID should be, because "new image link" is such an obvious name. :)

jarednorman (approved):

aldesantis:
Thanks @brchristian! This triggered my OCD too... I guess another solution would be to update both the JS and the spec to just take the link that is contained in the `li` element (`#new_image_link > a`), but it's definitely not a priority. 🙂

PR #3788, opened 2020-10-07
Explicitly require spree/auth_configuration

elia:
**Description**
Otherwise the corresponding constant will be autoloaded during
initialization, causing in turn a warning from Rails 6 / Zeitwerk.

```
DEPRECATION WARNING: Initialization autoloaded the constant Spree::AuthConfiguration.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload Spree::AuthConfiguration, for example,
the expected changes won't be reflected in that stale Class object.

This autoloaded constant has been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at /********/config/environment.rb:5)

Randomized with seed 9913
```

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/authorization_helpers.rb
+++ b/core/lib/spree/testing_support/authorization_helpers.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'cancan'
+require 'spree/auth_configuration'
 
 module Spree
   module TestingSupport


kennyadsl:
@elia can you please post the warning as well so it's searchable if someone has the same issue?

elia:
@kennyadsl sure, added, I'll make the whole case clear before marking it as ready for review, I'm still investigating the issue, this change might not be enough after all

elia:
The problem was in solidus_auth_devise, as explained in https://github.com/solidusio/solidus_auth_devise/pull/197

Issue #3756, opened 2020-09-11
Calculator should always return a BigDecimal, but in same scenarios that's not happening   


stefano-sarioli:
The calculator should always return a BigDecimal, but that's not always the case.

For example the [flat_rate calculator](https://github.com/solidusio/solidus/blob/v2.5.2/core/app/models/spree/calculator/flat_rate.rb) return 0 as Integer on the else branch.

If the calculator returns a type other than BigDecimal a warning will be issued by the adjustment action for the promotions,
for example [create_quantity_adjustments](https://github.com/solidusio/solidus/blob/v2.5.2/core/app/models/spree/promotion/actions/create_quantity_adjustments.rb#L56).

I think that we should document this expected interface on the Calculator, and update our actual implementation to respect the expected interface.

I wasn't able to find a test that raises this warning on the Solidus codebase, we should probably add some tests to cover this case.

**Solidus Version**:
First found out on Solidus 2.5.2.

**To Reproduce**
- Write a test that creates a create_quantity_adjustments using a flat_rate calculator giving in input a nil object.
- Launch it.
- A Warning should appear in the console about results not being a BigDecimal. ("Spree::Calculator::FlatRate#compute returned 0, it should return a BigDecimal")

**Current behavior**
- Raises a warning when using a standard calculator.

**Expected behavior**
- Doesn't raise any warning when using a standard calculator.

Ilyeo:
Hi! I did this PR https://github.com/solidusio/solidus/pull/3789, let me know if I made a mistake or if can I improve something thanks.

redencion:
Hi all, I tested  create_quantity_adjustments  I added nil object but it is not return 0 either bigdecimal. 
![testing](https://user-images.githubusercontent.com/6812808/137589285-df7c7c87-04ae-4ce4-b99e-81ac11333778.png)
this is I got 
![added_testing_233](https://user-images.githubusercontent.com/6812808/137589307-005816a1-22e4-44a3-814b-c7ebbf91986f.png)
So, I think as I saw in description it should get back bigdecimal, but It'll not to be, because there exist another class that catching object be nil or not, this is PartialLineItem here 
![quality_compute](https://user-images.githubusercontent.com/6812808/137590483-f59e8f6a-0359-4b01-b153-14dbcfb9ec29.png)
If I'm lost appreciate your help, thanks

PR #3789, opened 2020-10-07
Calculator should always return a big decimal

Ilyeo:
**Description**
We want to be sure to return a big decimal always that a calculator computes a quantity adjustment.
We added some specs to remove a warning message that informs when the calculator returned a non-big decimal result.

Ref. https://github.com/solidusio/solidus/issues/3756

**Checklist:**

### DIFF
--- a/core/app/models/spree/calculator/flat_rate.rb
+++ b/core/app/models/spree/calculator/flat_rate.rb
@@ -8,10 +8,12 @@ module Spree
     preference :currency, :string, default: ->{ Spree::Config[:currency] }
 
     def compute(object = nil)
+      return BigDecimal(0) if object.nil?
+
       if object && preferred_currency.casecmp(object.currency).zero?
         preferred_amount
       else
-        0
+        BigDecimal(0)
       end
     end
   end
--- a/core/app/models/spree/promotion/actions/create_quantity_adjustments.rb
+++ b/core/app/models/spree/promotion/actions/create_quantity_adjustments.rb
@@ -56,10 +56,12 @@ module Spree
         # adjustment. +adjustment_amount * 3+ or $15.
         #
         def compute_amount(line_item)
+          return BigDecimal(0) if line_item.nil?
+
           adjustment_amount = calculator.compute(PartialLineItem.new(line_item))
-          if !adjustment_amount.is_a?(BigDecimal)
-            Spree::Deprecation.warn "#{calculator.class.name}#compute returned #{adjustment_amount.inspect}, it should return a BigDecimal"
-          end
+          # if !adjustment_amount.is_a?(BigDecimal)
+          #   Spree::Deprecation.warn "#{calculator.class.name}#compute returned #{adjustment_amount.inspect}, it should return a BigDecimal"
+          # end
           adjustment_amount ||= BigDecimal(0)
           adjustment_amount = adjustment_amount.abs
 
--- a/core/spec/models/spree/promotion/actions/create_quantity_adjustments_spec.rb
+++ b/core/spec/models/spree/promotion/actions/create_quantity_adjustments_spec.rb
@@ -25,6 +25,31 @@ module Spree::Promotion::Actions
     describe "#compute_amount" do
       subject { action.compute_amount(line_item) }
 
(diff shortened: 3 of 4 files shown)


core/app/models/spree/promotion/actions/create_quantity_adjustments.rb
```diff
@@ -56,10 +56,12 @@ class CreateQuantityAdjustments < CreateItemAdjustments
...
         def compute_amount(line_item)
+          return BigDecimal(0) if line_item.nil?
+
           adjustment_amount = calculator.compute(PartialLineItem.new(line_item))
-          if !adjustment_amount.is_a?(BigDecimal)
-            Spree::Deprecation.warn "#{calculator.class.name}#compute returned #{adjustment_amount.inspect}, it should return a BigDecimal"
-          end
+          # if !adjustment_amount.is_a?(BigDecimal)
+          #   Spree::Deprecation.warn "#{calculator.class.name}#compute returned #{adjustment_amount.inspect}, it should return a BigDecimal"
+          # end
```
Ilyeo, line 64: do we want to remove these lines right?

core/app/models/spree/calculator/flat_rate.rb
```diff
@@ -8,10 +8,12 @@ class Calculator::FlatRate < Calculator
     preference :currency, :string, default: ->{ Spree::Config[:currency] }
 
     def compute(object = nil)
+      return BigDecimal(0) if object.nil?
```
jarednorman, line 11: I don't think this is necessary since when `object` is `nil` we still end up in the second part of the conditional below. This just causes us to check that twice.

core/app/models/spree/promotion/actions/create_quantity_adjustments.rb
jarednorman: I believe they should be kept, so that custom calculators will still generate warnings even if the ones in core don't.

core/app/models/spree/promotion/actions/create_quantity_adjustments.rb
```diff
@@ -56,10 +56,12 @@ class CreateQuantityAdjustments < CreateItemAdjustments
         # adjustment. +adjustment_amount * 3+ or $15.
         #
         def compute_amount(line_item)
+          return BigDecimal(0) if line_item.nil?
```
jarednorman, line 59: For roughly the same reason as the other, I don't think this is necessary.

core/spec/models/spree/promotion/actions/create_quantity_adjustments_spec.rb
```diff
@@ -25,6 +25,31 @@ module Spree::Promotion::Actions
...
+          let(:line_item) { nil }
+
+          before do
+            expect(Spree::Deprecation).to_not receive(:warn).
+              with(/^Spree::Calculator::FlatRate#compute returned 0, it should return a BigDecimal/, any_args)
+          end
+
+          it "returns a 0" do
+            expect(action.compute_amount(line_item)).to eq 0
+          end
```
jarednorman, line 41: We can simplify this by using the subject via `is_expected` and relying on the return type rather than the warning which could change.

```suggestion
          it { is_expected.to be_a BigDecimal }
          it { is_expected.to eq 0 }
```

core/spec/models/spree/promotion/actions/create_quantity_adjustments_spec.rb
```diff
@@ -25,6 +25,31 @@ module Spree::Promotion::Actions
...
+        end
+
+        context "with a valid line item " do
+          let(:line_item) { order.line_items.first }
+
+          it "returns a big decimal" do
+            expect(action.compute_amount(line_item)).to eq BigDecimal(0)
+          end
```
jarednorman, line 49:
```suggestion
          it { is_expected.to be_a BigDecimal }
          it { is_expected.to eq 0 }
```

jarednorman (changes requested):

waiting-for-dev:
I'll close it for now, but if you find the time to work on it I'll be more than happy to repent 🙂

PR #3792, opened 2020-10-08
Add CSS selector for datetime-local

### LINKED ISSUE #3755: Styling for datetime_field Does Not Work
Styling for the `datetime_field` does not work. 

image

This occurs because the `datetime_field` becomes an HTML field of type `datetime-local` and the CSS definition [here](https://github.com/solidusio/solidus/blob/16113356b9b21636d3e635705a468a5410deecc9/backend/app/assets/stylesheets/spree/backend/shared/_forms.scss) does not include this type.

**Solidus Version**:
2.10.1

jacobherrington:
**Description**

Relevant: https://apidock.com/rails/v6.0.0/ActionView/Helpers/FormHelper/datetime_field

You could argue that because we don't use this method in the vanilla backend, it doesn't need to be in the CSS, but I don't see any harm in adding it. :slightly_smiling_face: 

Closes #3755

**Checklist:**
~- [ ] I have attached screenshots to this PR for visual changes (if needed)~


aldesantis (approved):

aldesantis:
I think that makes sense, thanks @jacobherrington!

jarednorman (approved):

Issue #465, opened 2015-10-28
Implement available_until on products / variants


athal7:
To be able to easily set an end availability date

jasonfb:
do you want me to just pull request Spree PR #6794 against Solidus?

without un-deleting already deleted records, you still retain a host of bugs related to nil objects. The bulk of the heavy lifting in Spree PR 6794 is to backport the change to eliminate the ```undefined method for nil bugs```, which was the primary driver for me (the feature of discontinuing in the future was an added bonus).

jhawthorn:
@jasonfb thanks, but I have a few reservations about spree/spree#6794. I don't think the migration is appropriate (turning deleted products into discontinued products), as that may have significant behaviour differences in stores' custom code. `acts_as_paranoid` was not removed from the product, despite some of its necessary functionality being removed. There were also unrelated changes I would prefer to be reviewed separately (fixing line items of deleted products, for example).

What we'd like to see, at least initially, is a minimal PR implementing the `available_until` without breaking existing functionality.

jasonfb:
Yes I understand the reasoning behind moving slowly. However, adding a discontinued_on (or available_until) was not the primary objective of my work, the primary objective was to fix the dozens and dozens of bugs I have seen across multiple stores that are related to deleted objects. That issue has cost literally hundreds and hundreds of hours of developer time. That was the primary objective, the added feature was a side-benefit.

You're right it is a significant change, and upgraders would need to re-think their scoping mechanisms if they aren't already using the patched ```available``` scope. 

>> acts_as_paranoid was not removed from the product, despite some of its necessary functionality being removed

I'm not sure what you mean by this. By "its necessary functionality" are you referring to the ruby magic that happens to mysteriously re-show deleted objects? That's an anti-pattern, and leads to very bad bugs, and no one should code like that. It's true I left ```acts_as_paranoid``` in there, for when you really do want to delete something, but you shouldn't make deleting a normal part of a workflow. 

As far as "fixing line items of deleted products" I think you are referring to the rake task, which only happens if you run it explicitly, of course, and you will note the migration does not edit any line items. 

I do realize it's "seems" like a drastic change, but in my view simply adding a small ```available_until``` feature is just a small band-aid when the real problem is the flawed design and the dozens and dozens of bugs created by the design. Without address the root problem it seems like you aren't actually fixing what is arguably one of the worst bugs in Spree itself.

athal7:
[...]

PR #3793, opened 2020-10-08
Add discontinue on to products

tvdeyen:
**Description**

Adds a `discontinue_on` attribute to products. This accompanies the `available_on` attribute to complete the "Time Based Availability" feature. The `Product.available` scope and `Product#avaliable?` method take this new date filed into account to calculate the availability of products.

Refs #465

Monosnap 2020-10-09 12-55-16

_Note_ 

Contrary to the suggestions in #465 I used the `discontinue_on` attribute introduced in Spree 3.1 for compatibility reasons. If we think this is a mistake we can also rename the attribute, but this will make migrations for former Spree users harder.

**Checklist:**
- [ ] I have updated Guides and README accordingly to this change

### DIFF
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -168,11 +168,13 @@ module Spree
     end
 
     # Determines if product is available. A product is available if it has not
-    # been deleted and the available_on date is in the past.
+    # been deleted, the available_on date is in the past
+    # and the discontinue_on date is nil or in the future.
     #
     # @return [Boolean] true if this product is available
     def available?
-      !deleted? && available_on&.past?
+      !deleted? && available_on&.past? &&
+        (discontinue_on.nil? || discontinue_on.future?)
     end
 
     # Groups variants by the specified option type.
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -173,8 +173,17 @@ module Spree
     #
     # @return [Boolean] true if this product is available
     def available?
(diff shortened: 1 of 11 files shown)


core/app/models/spree/product.rb
```diff
@@ -168,11 +168,13 @@ def deleted?
     end
 
     # Determines if product is available. A product is available if it has not
-    # been deleted and the available_on date is in the past.
+    # been deleted, the available_on date is in the past
+    # and the available_on date is nil or in the future.
```
spaghetticode, line 172: I think we need to change this second `available_on` to `discontinue_on`.

spaghetticode (changes requested):
@tvdeyen thanks for the new feature, it looks good but there's a comment that needs a small change.

core/app/models/spree/product.rb
tvdeyen: Thanks for the early feedback :)

This PR is still in Draft, though.

tvdeyen:
@kennyadsl @aldesantis there are lots of failing specs that don't seem to be related to this PR. Do I need to do something or do we have CI issues?

jarednorman (approved):
Nice, definitely a good change because I know I'm going to have remove the custom discontinued tracking from multiple existing stores and switch to this.

spaghetticode (approved):
@tvdeyen thanks for this neat new feature ❤️

aldesantis:
😍 Thanks @tvdeyen!

PR #3794, opened 2020-10-09
Allow builds with Rails master to fail on Circle CI

tvdeyen:
Do not fail the circle ci step if the `RAILS_VERSION` is set to `master`.

It is not very welcoming for contributors to get a mail from Circle CI that the build has failed, just because we also test for latest unreleased version of Rails

build (d5a7dc4) — Inbox 2020-10-09 23-44-07

Also the list of open PRs looks a bit discouraging with all of them failing.

solidus 2020-10-09 23-46-18

**Checklist:**

### DIFF
--- a/bin/build-ci
+++ b/bin/build-ci
@@ -95,7 +95,7 @@ class Project
       log("- #{project.name} #{build ? 'SUCCESS' : 'FAILURE'}")
     end
 
-    builds.all?
+    builds.all? || ENV['RAILS_VERSION'] == 'master'
   end
   private_class_method :test
 


kennyadsl (dismissed):
Thank you!!!! 😍😍

kennyadsl:
Wait, does this make the individual build check against master green, even if it actually fails? At this point maybe it's better to remove the check al all? Or maybe use a neutral status code if exists.

tvdeyen:
> Wait, does this make the individual build check against master green, even if it actually fails? 

Yes. Since Rails master builds are not important for us, we need to check them individually. We still log the results, so devs can look into it, if they like.

> At this point maybe it's better to remove the check al all?

Not sure. It is still valid information.

> Or maybe use a neutral status code if exists.

No, I am not aware of any neutral status in GitHub PRs.

tvdeyen:
@kennyadsl 
The current status is not pleasing at all. The Readme displays a failing build, although our master is green, only the Rails master part is failing. All PRs are red and contributions get emails about failing builds. Not very welcoming ;)

kennyadsl:
I agree it's not welcoming but I don't see the point of spending CI time to have no information back. At this point, if someone is curious to know if rails master is working, they can run the build locally. 

The neutral status should be the one posted by Netlify for example:

Screenshot 2020-10-13 at 09 51 04

We'd still consume CI time but at least we are not faking a success status.

tvdeyen:
Then let's remove it for now. We can figure out how to report neutral checks later.

kennyadsl:
Works for me, as soon as it's always red it's also not providing useful information, and it's also non-welcoming.

tvdeyen:
Superseded by #3796

PR #3796, opened 2020-10-13
Disable Rails master builds on CircleCI

tvdeyen:
They are failing all the time and should not fail a PR. Since we
do not know how to send neutral status checks we should remove
them.

**Checklist:**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ workflows:
       - mysql
       - postgres_rails52
       - mysql_rails52
-      - postgres_rails_master_activestorage
+      # - postgres_rails_master_activestorage
       - stoplight/push:
           context: "Solidus Core Team"
           project: solidus/solidus-api


kennyadsl (approved):
We can reintroduce this check once we have the green check green again, or find a way to report a less aggressive state in the GitHub checks.

stem:
Hey @tvdeyen, what do you think about running specs against rails master on the master branch once per week ?
I'm not a CircleCI configuration expert, but [it seems doable](https://circleci.com/docs/2.0/triggers/#scheduled-builds) and would enable us to see if it works with rails master without compromising build status of solidus' PRs.

tvdeyen:
> Hey @tvdeyen, what do you think about running specs against rails master on the master branch once per week ?
> I'm not a CircleCI configuration expert, but [it seems doable](https://circleci.com/docs/2.0/triggers/#scheduled-builds) and would enable us to see if it works with rails master without compromising build status of solidus' PRs.

Yes, but we need to make sure that the README badge is not effected by this. Currently we point this to the latest master branch build.

jarednorman (approved):

spaghetticode (approved):

PR #3797, opened 2020-10-13
Try another CircleCI configuration

stem:
Sorry, I tried to make it on our fork to try a CircleCI configuration for #3796 but instead, make it on the official repository...

This keeps the spec running against rails master, but fake the result. I don't think what you prefer : 
1. not having automated test on rails master
2. having spec that are wrongly green and need to check the output to know for sure if they passed or failed

@tvdeyen or @kennyadsl , you can safely close this PR once you'll read this message if you decide to keep going the road of #3796

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,23 @@ commands:
       - store_test_results:
           path: /tmp/test-results
 
+  test_without_failing:
+    steps:
+      - run:
+          name: Run Tests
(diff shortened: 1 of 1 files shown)


stem:
I see that #3796 is merged so I'm closing this one

kennyadsl:
Yes, I was reporting here that we decided to take that path, mainly to avoid spending CI time with an information that is not reliable. Thanks anyway, it's appreciated!

Issue #3803, opened 2020-10-21
Germany and other countries are seeded with `states_required` true


seand7565:
German addresses do not typically use states. See: https://www.angloinfo.com/how-to/germany/housing/postal-system

This can cause validation issues on things like `solidus_paypal_braintree` because PayPal will not return a state when the users selects a German address in PayPal. 

Valid German address in PayPal (no state):
<img width="412" alt="Screen Shot 2020-10-21 at 2 52 03 PM" src="https://user-images.githubusercontent.com/5720486/96775345-07704400-13ad-11eb-9448-808dc51e91c6.png">

the below error happens on submit with `solidus_paypal_braintree` cart checkout on submit:
<img width="645" alt="Screen Shot 2020-10-21 at 2 52 10 PM" src="https://user-images.githubusercontent.com/5720486/96775354-093a0780-13ad-11eb-80dc-97f215cc341a.png">

I'm sure there are other countries like this, but Germany is the only one that I'm currently aware of. 

`states_required` is set to true automatically if the country has any subregions in Carmen, but it appears that that shouldn't happen in some cases. [See here](https://github.com/solidusio/solidus/blob/a363a7484f0d537887dfb23937615ecb20da4a42/core/db/default/spree/countries.rb#L16) - I think that's where it's being set.

stuffmatic:
Another example of country where `states_required` should be `false` is Sweden. It's easy enough to set this flag manually if need be, but it would be nice if this information was kept and maintained centrally in Solidus. As far as I can tell, the value of `states_required` can't be deduced from Carmen data alone, so out of curiosity I had a quick look at the Spree code base and it seems like they're using a [hardcoded list of countries](https://github.com/spree/spree/blob/a1ad03b5335aa5882456feafa3e1d03a0b45eb7b/core/app/models/spree/address.rb#L14) for which `states_required` should be `true`. Would something like this be a reasonable solution?

victorth:
The same is true for the Netherlands and I believe for, most European countries.

AlistairNorman:
[...]

PR #3807, opened 2020-10-23
Change state_required detections

DanielePalombo:
`Carmen::Country#subregions?` returns true even if the subregions aren't states.
With this commit state_required variable will set true only when there is at least one nested state.

To update existing countries you can execute the following command on the console.
```
require 'carmen'

carmen_countries = Carmen::Country.all

Spree::Country.all.each do |country|
  next if carmen_countries.detect { |c| c.name == country.name }.subregions.any? { |c| c.subregions.detect { |r| r.type == 'state' } }
  
  country.update(states_required: false)
end
```

Ref https://github.com/solidusio/solidus/issues/3803

**Checklist:**

### DIFF
--- a/core/db/default/spree/countries.rb
+++ b/core/db/default/spree/countries.rb
@@ -13,7 +13,7 @@ country_mapper = ->(country) do
   iso             = connection.quote country.alpha_2_code
   iso_name        = connection.quote country.name.upcase
   numcode         = connection.quote country.numeric_code
-  states_required = connection.quote country.subregions?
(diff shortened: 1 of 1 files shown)


kennyadsl:
I'm not sure to get this. How having states as subregions and subregion being required are connected. For example, in Italy we have provinces, which are not states and they are required in the address.

waiting-for-dev:
Closing in favor of #3842 to properly define all the logic involved here.

PR #3813, opened 2020-10-29
Move UserClassHandle to lib/

elia:
**Description**

This class is potentially loaded during initialization when used by
factories and that can generate problems with reloaders.

In Rails 6 / zeitwerk that resulted in this warning message:

```
DEPRECATION WARNING: Initialization autoloaded the constant Spree::UserClassHandle.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload Spree::UserClassHandle, for example,
the expected changes won't be reflected in that stale Class object.

This autoloaded constant has been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at /my-app/config/environment.rb:5)
```

**Checklist:**


spaghetticode (approved):
@elia thanks 👍

jarednorman (approved):

PR #3815, opened 2020-10-29
Use the right method in the AddApplyToAllToVariantPropertyRule migration

### LINKED ISSUE #3809: Invalid method used in the the AddApplyToAllToVariantPropertyRule migration
I think this one should be `up` instead of `change`:

https://github.com/solidusio/solidus/blob/44da6ff1061ac8dfd339f02228bbd921b4a4c006/core/db/migrate/20180416083007_add_apply_to_all_to_variant_property_rule.rb#L4

Otherwise it's impossible to reverse this migration.

ok32:
**Description**

I think `change` is used here by mistake. Among other things this renders the migration irreversible.

Closes #3809


kennyadsl (approved):
👍 Thank you!

jarednorman (approved):

PR #3818, opened 2020-10-29
Remove all code deprecated in Solidus v2.x

### LINKED ISSUE #3816: Remove all deprecated code for Solidus 3.0
Before releasing Solidus 3.0 we have to remove all deprecated code from the codebase.

kennyadsl:
Closes #3816 

Now that master points to Solidus 3.0, we can remove all code deprecated in Solidus 2.x and introduce the breaking changes of this major.

### Removals without deprecation

- `Spree::LineItem::CurrencyMismatch` exception: we are not using it anymore since the behavior we had with `Spree::Config.raise_with_invalid_currency = true` has been removed. 
- `Spree::Order::Checkout` is not used anymore. 
- `Spree::Admin::PaymentsHelper` is not used anymore.
- `UserPaymentSource` is empty after removing all deprecated methods inside it. 
- `Spree::Refund#perform_after_create` attribute, if you are still using it, please stop, it does nothing now.
- `Spree::TaxCalculator#ShippingRate`: it is now always `nil`
- `Spree::Money::RUBY_NUMERIC_STRING`: was only used in a deprecated code path. 

#### Preferences

We removed the following preferences without deprecations. They were just controlling a deprecated flow and have no effect so, assuming you already switched to the only accepted value, you can safely remove them.

- `Spree::Config.raise_with_invalid_currency`
- `Spree::Config.redirect_back_on_unauthorized preference`
- `Spree::Config.run_order_validations_on_order_updater preference `
- `Spree::Config.use_legacy_order_state_machine`
- `Spree::Config.use_legacy_store_credit_reimbursement_category_name`
- `Spree::Config.consider_actionless_promotion_active`
- `Spree::Config.use_legacy_address_state_validator` (removed in #3847)
- `Spree::Config.use_combined_first_and_last_name_in_address` (removed in #3820)


api/app/controllers/spree/api/checkouts_controller.rb
```diff
@@ -11,9 +11,6 @@ class CheckoutsController < Spree::Api::BaseController
 
       include Spree::Core::ControllerHelpers::Order
 
-      # TODO: Remove this after deprecated usage in #update is removed
-      include Spree::Core::ControllerHelpers::PaymentParameters
```
DanielePalombo, line 15: This can't be removed because `update` uses `set_payment_parameters_amount`. I'm going to remove the comment about deprecation.

kennyadsl: Yes, makes sense. The comment didn't take that into account. 👍

DanielePalombo: The method has been added later. I'm going to squash the commit `https://github.com/solidusio/solidus/pull/3818/commits/51597744f1291a828aaa3ef3c9d7f9544559c790`

kennyadsl: Seems like it [was already there](https://github.com/solidusio/solidus/commit/a173f23884ef3901ceef3f7fd4fb1dff1ae42223#diff-45fff904f0af543763fdd5e09e339a490d759b46b78fabb150bd0f41e532570bR91), but nevermind. We know what to do!

api/spec/requests/spree/api/checkouts_controller_spec.rb
```diff
@@ -399,7 +399,7 @@ module Spree
         end
 
         it "returns a sensible error when no payment method is specified" do
-          # put :complete, id: order.to_param, order_token: order.token, order: {}
+          # put :complete, id: order.to_param, order_token: order.guest_token, order: {}
```
kennyadsl, line 402: What about removing these comments at all?

core/lib/spree/core/class_constantizer.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'active_support'
```
kennyadsl, line 3: Not sure to understand why it was working before this change, did you get that? Also, please use a permalink in the commit otherwise that file could change in the future and not point anymore to the right line.

api/spec/requests/spree/api/checkouts_controller_spec.rb
kennyadsl: @DanielePalombo ☝️

core/lib/spree/core/class_constantizer.rb
DanielePalombo: This worked because `active_support` was implicit required by `Spree::Core::Search::Base` when [spree/deprecation](https://github.com/solidusio/solidus/blob/ba78d7358c81baeeadab9a08d0bab9c21ea500ba/core/lib/spree/core/search/base.rb#L3) was required. I'll change the commit link. 👍

api/spec/requests/spree/api/checkouts_controller_spec.rb
DanielePalombo: Removed 👍

core/lib/spree/core/class_constantizer.rb
kennyadsl: Makes sense, thanks!

jarednorman:
CleanShot 2021-01-03 at 14 51 52@2x

🔪🔪🔪

jarednorman (approved):
I've gone over this commit by commit now and I think it looks good. Thanks for the hard work on tracking all this stuff down. 🙌🏻

spaghetticode (approved):
looks great, thanks @kennyadsl and @DanielePalombo 💯 ✂️ 🎉

kennyadsl:
Rebased, going to merge this as soon as specs pass

PR #3820, opened 2020-10-30
Remove deprecated name-related Address fields

filippoliverani:
**Description**

Ref #3816 and #3234

This PR removes deprecated public-accessible `Address` name fields and their aliases:
- `full_name`
- `firstname`
- `first_name`
- `lastname`
- `last_name`.

`Address` `name` attribute is still interally persisted on the DB as two separate `firstname` and `lastname` columms, in a future PR we can move them to a new `name` column and provide a rake task to help migrate existing data.

**Checklist:**
- ~~[] I have added tests to cover this change (if needed)~~

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -14,34 +14,23 @@ module Spree
     belongs_to :country, class_name: "Spree::Country", optional: true
     belongs_to :state, class_name: "Spree::State", optional: true
 
-    validates :address1, :city, :country_id, presence: true
+    validates :address1, :city, :country_id, :name, presence: true
     validates :zipcode, presence: true, if: :require_zipcode?
     validates :phone, presence: true, if: :require_phone?
-
-    validate :validate_name
 
     validate do
       self.class.state_validator_class.new(self).perform
     end
 
-    alias_attribute :first_name, :firstname
-    alias_attribute :last_name, :lastname
-    alias_attribute :full_name, :name
-
     DB_ONLY_ATTRS = %w(id updated_at created_at)
     TAXATION_ATTRS = %w(state_id country_id zipcode)
-    LEGACY_NAME_ATTRS = %w(firstname lastname full_name)
+    LEGACY_NAME_ATTRS = %w(firstname lastname)
 
     self.whitelisted_ransackable_attributes = %w[firstname lastname]
 
     scope :with_values, ->(attributes) do
       where(value_attributes(attributes))
     end
-
-    Spree::Deprecation.deprecate_methods(
-      Spree::Address,
-      LEGACY_NAME_ATTRS.product([:name]).to_h
-    )
 
     # @return [Address] an address with default attributes
     def self.build_default(*args, &block)
@@ -74,17 +63,7 @@ module Spree
     # @return [Hash] hash of attributes contributing to value equality with optional merge
     def self.value_attributes(base_attributes, merge_attributes = {})
       base = base_attributes.stringify_keys.merge(merge_attributes.stringify_keys)
-
-      name_from_attributes = Spree::Address::Name.from_attributes(base)
-      if base['firstname'].presence || base['first_name'].presence
-        base['firstname'] = name_from_attributes.first_name
-      end
-      if base['lastname'].presence || base['last_name'].presence
-        base['lastname'] = name_from_attributes.last_name
-      end
-      excluded_attributes = DB_ONLY_ATTRS + %w(first_name last_name)
-
-      base.except(*excluded_attributes)
+      base.except(*DB_ONLY_ATTRS)
     end
 
     # @return [Hash] hash of attributes contributing to value equality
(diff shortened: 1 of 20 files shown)


core/app/models/spree/address.rb
```diff
@@ -85,16 +74,15 @@ def self.immutable_merge(existing_address, new_attributes)
...
 
-      name_from_attributes = Spree::Address::Name.from_attributes(base)
-      if base['firstname'].presence || base['first_name'].presence
-        base['firstname'] = name_from_attributes.first_name
-      end
-      if base['lastname'].presence || base['last_name'].presence
-        base['lastname'] = name_from_attributes.last_name
-      end
-      excluded_attributes = DB_ONLY_ATTRS + %w(first_name last_name)
+      # name_from_attributes = Spree::Address::Name.from_attributes(base)
```
kennyadsl, line 77: Do we need to remove those comments?

core/app/models/spree/address.rb
```diff
@@ -34,25 +32,16 @@ class Address < Spree::Base
...
     end
 
-    alias_attribute :first_name, :firstname
-    alias_attribute :last_name, :lastname
-    alias_attribute :full_name, :name
-
     DB_ONLY_ATTRS = %w(id updated_at created_at)
     TAXATION_ATTRS = %w(state_id country_id zipcode)
-    LEGACY_NAME_ATTRS = %w(firstname lastname full_name)
+    LEGACY_NAME_ATTRS = %w(firstname lastname)
```
kennyadsl, line 37: Do we really need to keep this constant? I saw it's used in `as_json` but I can't get why, can you help me understand?

core/app/models/spree/address.rb
filippoliverani: Yes, thanks!

core/app/models/spree/address.rb
filippoliverani: I've kept it to make explicit why we remove them from the json hash.
Since they're still on the db they are added by default to any attributes representation, including the json one, but we want to keep them private and avoid that they creep outside the model polluting API responses and breaking equality checks.

kennyadsl: Got it, it's clear now. I wasn't thinking that the list was coming magically from db columns. Have you considered using Rails' `ignore_columns` for this? There's also a [similar example](https://github.com/rails/rails/blob/b80645003796690b76aef5c289923db543d84673/activerecord/test/models/developer.rb#L25) in Rails specs. 😆

filippoliverani: Yes, I thought about it but we still need those columns to be loaded, our internal implementation of `name` accessor still rely on them, I plan to remove them all togheter in the next PR 👍

DanielePalombo (approved):
LGTM

kennyadsl (approved):
Thanks, @filippoliverani!

spaghetticode (approved):
@filippoliverani can you please rebase with the recent changes in `address.rb`?
Otherwise, it looks good to me 👍

peterberkenbosch:
how about we keep the first & lastname and add `has_person_name` dependency? This way we can still use firstname etc for integrations that need that but also just have a name field.

DanielePalombo:
@spaghetticode I have rebase this PR

filippoliverani:
@peterberkenbosch @kennyadsl @spaghetticode 
> how about we keep the first & lastname and add `has_person_name` dependency? This way we can still use firstname etc for integrations that need that but also just have a name field.

An important principle behind the whole refactoring is that a name in an address doesn't have to map to a person name, I'm not sure we should encourage users to think otherwise.
Internally we were [already splitting a full name to make legacy address-name based integrations work](https://github.com/solidusio/solidus/blob/a17109b07a817ae34bc61491c3ed1ea824ee32c5/core/app/models/spree/credit_card.rb#L165), we can keep using the same technique.

Moreover, splitting a name in multiple components make sense only for some kind of names and for some locales, Basacamp's name_of_person works only for english names unfortunately.

I think that, in general, developers should add their opinionated or locale-dependent way of saving persons' names to their User models. Address names instead can be more flexible and future-proof if kept as generic as possible.

What do you think?

kennyadsl:
As already suggested by someone else (maybe @elia), I also think we should have a `name` field on the user, which should represent how the user wants to be called. This could be optional so that we have a migration path for stores that want to keep that information present. 

Also, we are still on time to keep both code paths (`firstname+lastname` and `fullname`) in Solidus 3, by removing the deprecation and allowing people to choose what works best in their store while we work on providing a better alternative for having a valid person name stored instead of using the shipping address name.

filippoliverani:
> Also, we are still on time to keep both code paths (`firstname+lastname` and `fullname`) in Solidus 3, by removing the deprecation and allowing people to choose what works best in their store while we work on providing a better alternative for having a valid person name stored instead of using the shipping address name.

If we  want to keep`firstname+lastname` in Address, we have to explicitly write and save them to two separate attributes, otherwise we won't know how they should be split in the first place. If this is what we want, we should disregard #3234 and revert the work done up to now.

kennyadsl:
No wait, I didn't mean to revert what we did up to now 😅 , I meant (and I'm not even sure it's possible) to keep allowing both paths based on the `use_combined_first_and_last_name_in_address ` preference value, until we have a clear alternative for those who were using the first name for specific reasons, like marketing emails or stuff like that. Of course, if we have a solution in mind we can go ahead, merge this, and release Solidus 3.0, but it sounds to me like something that a lot of stores will need to solve on their own before upgrading to Solidus 3.0.

peterberkenbosch:
I'm late to the party around the naming, I don't really understand or see the need to make this generic tbh. What I see right now personally is that it is causing more problems then it's solving. I do like the idea of moving the person name (firstname, lastname, mabye even middlename) to the user. That seems to be semanticly more correct. However this will cause some issues on what to do when a user wants to ship to someone else. What do we put on the shipping information? We do need a name there. Just a generic name field on address does not make sense to me tbh. Again, I'm late to the party, I might have missed the initial discussion around that (reading up on that right now). 

Either way, it's easy enough to manually override this or make a tiny extension that does what I need :) Just a bit confused about the need. Feel free to ignore the above, the PR itself looks good to me.

peterberkenbosch:
> I don't really understand or see the need to make this generic tbh.

And now I do.. 👍 any specific needs should be implemented by users indeed.

PR #3821, opened 2020-10-30
Implicitly include required spec modules

blocknotes:
**Description**
In _solidus_starter_frontend_ we had an issue with `Spree::TestingSupport::Translations` because the module file was required in the spec_helper but it wasn't included in RSpec configuration. Reference PR [here](https://github.com/nebulab/solidus_starter_frontend/pull/121).

I'm going to propose with this PR to automatically include the spec modules that we are requiring to avoid these kinds of conflicts.

**Checklist:**

### DIFF
--- a/api/spec/spec_helper.rb
+++ b/api/spec/spec_helper.rb
@@ -54,7 +54,6 @@ RSpec.configure do |config|
   config.extend Spree::Api::TestingSupport::Setup, type: :request
   config.include Spree::Api::TestingSupport::Helpers, type: :controller
   config.extend Spree::Api::TestingSupport::Setup, type: :controller
-  config.include Spree::TestingSupport::Preferences
   config.include Spree::TestingSupport::JobHelpers
 
   config.extend WithModel
--- a/backend/spec/spec_helper.rb
+++ b/backend/spec/spec_helper.rb
@@ -95,7 +95,6 @@ RSpec.configure do |config|
 
   config.include FactoryBot::Syntax::Methods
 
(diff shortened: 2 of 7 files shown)


kennyadsl:
I understand the concern but I don't like that we have different behaviors. For example, why `config.include Spree::TestingSupport::Flash` needs a manual include?

blocknotes:
> I understand the concern but I don't like that we have different behaviors. For example, why `config.include Spree::TestingSupport::Flash` needs a manual include?

Because the other modules (like Flash - also if I didn't miss anything) are not calling a method that they define in their `RSpec.configure` block (if any).

But it is a valid point. From my standpoint, we could include also those modules in their respective files.

WDYT?

kennyadsl:
Yes, we could. Another thing to consider is extensions that are currently both requiring files and manually including those modules. Do those extensions end up having them included twice? Do we have a way to handle that?

waiting-for-dev:
Hey @blocknotes, do you think it's still an issue?

blocknotes:
> Hey @blocknotes, do you think it's still an issue?

Hey @waiting-for-dev.
Honestly I don't know 🤷‍♂️ `solidus_starter_frontend` could have been updated quite a lot during this time.
So, feel free to do anything you prefer with this PR.

waiting-for-dev:
:+1: thanks, @blocknotes, I'll close it for now, and we can revisit if we find a problem. Pinging @gsmendoza in case you want to check something on startar_solidus_frontend.

PR #3823, opened 2020-10-30
Multiple domains are considered

johvo:
**Description**
If you have two shops configured like my-shop.net and other-shop.net, while other-shop should respond to two domains (other-shop.net and www.other-shop.net) this seems not to work at the moment. The fix is less strict while checking the domain name.

**Checklist:**

### DIFF
--- a/core/app/models/spree/store_selector/by_server_name.rb
+++ b/core/app/models/spree/store_selector/by_server_name.rb
@@ -22,7 +22,7 @@ module Spree
         # We select a store which either matches our server name, or is default.
         # We sort by `default ASC` so that a store matching SERVER_NAME will come
         # first, and we will find that instead of the default.
-        store = Spree::Store.where(url: server_name).or(Store.where(default: true)).order(default: :asc).first
+        store = Spree::Store.by_url(server_name).or(Store.where(default: true)).order(default: :asc).first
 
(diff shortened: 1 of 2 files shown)


aldesantis:
@johvo thanks for the contribution! FYI your tests are failing here because you're using `by_url`, which was deprecated. We're not supposed to call deprecated code in the Solidus codebase, since it will be removed in the next major.

Also, I think this might be considered a breaking change. Someone might rely on the fact that www.shop.net returns a store, while shop.net returns a different store. I think your improvement should be implemented as a new store selector rather than a change to the default one.

johvo:
Hallo @aldesantis,
thank you for your feedback, I will check my implementation and suggest a improved version.
Should I close this pull-request for the time being?

aldesantis:
@johvo that's not necessary, you can keep this open and work here. 🙂

kennyadsl:
@johvo 👋  

Are you still interested in finishing this PR or we can close it? Anyway, thank you for the contribution!

johvo:
Hi, thank you for asking.
I still plan to finish it. I put the task on my to-do list again. I hope to
get it soon done.

Best regards Tilman


Alberto Vena <notifications@github.com> schrieb am Fr., 19. Feb. 2021,
16:00:

PR #3824, opened 2020-10-31
Move "thinking cat" fixture to lib folder

mamhoff:
## Description

With the exclusion of the spec folder from the packaged solidus_core gem
in 80bea87, all factories that use an image can now not access that
image when loaded from the lib folder of a packaged gem.

The reason for that is that the image referenced in those factories
lives under spec/fixtures - in the spec folder.

This commit does the simplest thing that could possibly work: Move the
image to the lib/fixtures folder instead and adapt the factories so that
they work even if loaded from the packaged gem.

One could imagine a single-pixel image file instead of an actual JPEG of a thinking cat to conserve space on the many machines this gem runs on. Would that still be in scope? 

I would very much like a new version of Solidus to be cut with this change, as 2.11.0 has broken factories because of this.

**Checklist:**


aldesantis:
Thanks @mamhoff! 🙏 Do you think it would make more sense to move the image to `lib/spree/testing_support/fixtures`, since that's where all test-related files live (including the factories)?

mamhoff:
Sure! but you know what: When I do that, I might as well also do the 1-pixel-image thing? I mean I don't think it's worth distributing those 20kB that I have not looked at in the however many years I've spent with this software.

aldesantis:
Go for it! 👍

kennyadsl:
@mamhoff thanks for this PR! For some reason, CircleCI did not receive the hook for running the CI after the last push. Can you please push again to retrigger it?

kennyadsl (approved):
Thanks, Martin!

aldesantis (approved):

kennyadsl:
I'm going to backport and release Solidus 2.11.1 tomorrow, thanks again!

PR #3827, opened 2020-11-05
Refactor and add specs to stock locations helper

gabrielbaldao:
**Checklist:**

### DIFF
--- a/backend/app/helpers/spree/admin/stock_locations_helper.rb
+++ b/backend/app/helpers/spree/admin/stock_locations_helper.rb
@@ -4,9 +4,7 @@ module Spree
   module Admin
     module StockLocationsHelper
       def admin_stock_location_display_name(stock_location)
-        name_parts = [stock_location.admin_name, stock_location.name]
-        name_parts.delete_if(&:blank?)
-        name_parts.join(' / ')
+        [stock_location.admin_name, stock_location.name].reject(&:blank?).join(' / ')
       end
     end
   end
--- a/backend/spec/helpers/admin/stock_locations_helper_spec.rb
+++ b/backend/spec/helpers/admin/stock_locations_helper_spec.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Spree::Admin::StockLocationsHelper, type: :helper do
+  describe "#admin_stock_location_display_name" do
+    subject { helper.admin_stock_location_display_name(stock_location) }
+
+    context "without admin_name" do
+      let(:stock_location) { create(:stock_location_with_items, name: "NY Warehouse") }
+
+      it { is_expected.to eq "NY Warehouse" }
+    end
+
+    context "with admin_name" do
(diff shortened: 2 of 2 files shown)


backend/app/helpers/spree/admin/stock_locations_helper.rb
```diff
@@ -4,9 +4,7 @@ module Spree
   module Admin
     module StockLocationsHelper
       def admin_stock_location_display_name(stock_location)
-        name_parts = [stock_location.admin_name, stock_location.name]
-        name_parts.delete_if(&:blank?)
-        name_parts.join(' / ')
+        [stock_location.admin_name, stock_location.name].compact.join(' / ')
```
jarednorman, line 7: `.compact` and `delete_if(&:blank?)` don't do the same thing. `.compact` only removes `nil` values, not empty strings.

```suggestion
        [stock_location.admin_name, stock_location.name].reject(&:blank?).join(' / ')
```

jarednorman (changes requested):
This changes the behaviour when the `admin_name` or `name` are empty/blank strings rather than `nil` values.

Ideally the new specs should enforce this behaviour that your changes might otherwise have broken.

waiting-for-dev:
Hey @gabrielbaldao, it's been a while, but would you like to finish it with Jared's suggestion?

waiting-for-dev:
> Hey @gabrielbaldao, it's been a while, but would you like to finish it with Jared's suggestion?

I did it myself. It should be ready for review :slightly_smiling_face:

gabrielbaldao:
thank you @waiting-for-dev i appreciate that

spaghetticode (approved):
@gabrielbaldao @waiting-for-dev thanks!

backend/spec/helpers/admin/stock_locations_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Spree::Admin::StockLocationsHelper, type: :helper do
+  describe "#admin_stock_location_display_name" do
+    subject { helper.admin_stock_location_display_name(stock_location) }
+
+    context "without admin_name" do
+      let(:stock_location) { create(:stock_location_with_items) }
```
aldesantis, line 10: Can we set the name explicitly, both here and in the `let` statement below? Right now "NY Warehouse" is a magic string and relies on the factory's defaults—it would be good to make it explicit, for clarity and so that any changes in the factory don't affect the test.

aldesantis (changes requested):

jarednorman (commented):
Looks good, pending Alessandro's request to update the tests so that the value the output depends on is explicit in the setup.

backend/spec/helpers/admin/stock_locations_helper_spec.rb
waiting-for-dev: It makes a lot of sense. Thanks for the suggestion. Done!

jarednorman (approved):

aldesantis (approved):

PR #3828, opened 2020-11-05
Refact removing switch case in callback class

gabrielbaldao:
I like to avoid switch cases because of the interpreter when it reaches the low level, but I know that apparently it is not as readable, feel free to close this PR if you think this is unnecessary

**Checklist:**

### DIFF
--- a/backend/lib/spree/backend/callbacks.rb
+++ b/backend/lib/spree/backend/callbacks.rb
@@ -41,11 +41,11 @@ module Spree
       def invoke_callbacks(action, callback_type)
         callbacks = self.class.callbacks || {}
         return if callbacks[action].nil?
-        case callback_type.to_sym
-        when :before then callbacks[action].before_methods.each { |method| send method }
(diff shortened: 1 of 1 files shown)


aldesantis:
@gabrielbaldao I'm not entirely clear on the rationale behind this change, could you explain?

Generally speaking, I like to avoid this kind of "magic" in the code because it's harder to read and it may come back to bite us if we're not careful. However, I'm happy to consider a change if it's well motivated. 🙂

jarednorman (changes requested):
The original has two main benefits over the proposed change:

1. The original makes it clear what the valid values are for `callback_type`.
2. The original will only call the three methods it's meant to call. The new version could call any method ending in `_methods` erroneously.

I could imagine a refactoring to make this method clearer/simpler, but I don't think this is an improvement as it is now.

kennyadsl:
Going to close this due to a lack of response from the author. Thanks anyway!

PR #3831, opened 2020-11-05
Fix ability to perform refunds after a first failed attempt

kennyadsl:
**Description**

With #1415 we introduced a bug in the Admin:

When attempting to create a refund for a payment that fails with a gateway error, it's no more possible to create refunds for that payment. 

To reproduce:

1. Click the refund link on any payment

Screenshot 2020-11-05 at 15 41 27

2. Refund credit allowed is correct, let's proceed with the refund **using a gateway that will fail**

Screenshot 2020-11-05 at 15 41 38

At this point, we have the flash message correctly printed with the error:

Screenshot 2020-11-05 at 15 42 02

3. A refund is actually created and we can't create new refunds:

Screenshot 2020-11-05 at 15 42 16

This is due to the split of record creation and calling `perform!` in `Spree::Admin::RefundsController#create`. In fact, when we create a refund in the first place, it is added to the database but when we call `perform!` right after the creation, and it fails, the record created is not marked as failed anyhow. The system assumes all refunds are successful and consider them done if when they failed, disabling us to create new ones for those payments.

In fact, the refundable amount is calculated with 

```
payment.credit_allowed = payment total - refunds amount
```

and this formula is true also when the refund has not been successful.

This comes from our past behavior of creating and calling perform! on refunds at the same time. Records weren't created if the refund failed to perform.

A more clean solution would be considering in that formula only refunds that are completed but at the moment there's no way to do that. This way we can also keep track of the failed attempts to perform refunds, saving all the information we get back from the gateway. But this is way more invasive and we can't do this in a patch level fix. 

This change is supposed to be merged in master (for 3.0) and backported to v2.11, which is the version that introduces the bug.

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/refunds_controller.rb
+++ b/backend/app/controllers/spree/admin/refunds_controller.rb
@@ -11,8 +11,9 @@ module Spree
       rescue_from Spree::Core::GatewayError, with: :spree_core_gateway_error
 
       def create
-        @refund.attributes = refund_params.merge(perform_after_create: false)
-        if @refund.save && @refund.perform!
+        @refund.attributes = refund_params
+
+        if @refund.valid? && @refund.perform!
           flash[:success] = flash_message_for(@refund, :successfully_created)
           respond_with(@refund) do |format|
             format.html { redirect_to location_after_save }
--- a/backend/spec/controllers/spree/admin/refunds_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/refunds_controller_spec.rb
@@ -42,7 +42,7 @@ describe Spree::Admin::RefundsController do
     context "a Spree::Core::GatewayError is raised" do
       before do
         expect_any_instance_of(Spree::Refund).
-          to receive(:perform!).
+          to receive(:process!).
           and_raise(Spree::Core::GatewayError.new('An error has occurred'))
       end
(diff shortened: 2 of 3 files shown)


jarednorman (changes requested):
What if we initialize but don't save the refund before calling `perform!` and have `perform!` save the record (if it doesn't already)?

This would avoid this weird code that recreates the refund.

Regardless, I agree that tracing the success/fail state would be a better solution in the long run. Would probably be worth having three states: pending (default state), complete and failed.

aldesantis:
I would also prefer not to save the refund at all unless it's been successfully performed. Would that be possible?

kennyadsl:
I think it's a great idea, working on that and will let you know, thanks!

kennyadsl:
@aldesantis @jarednorman please review again when you have time, and thanks again for pointing in the right direction. The first attempt was so bad!

spaghetticode (approved):
@kennyadsl makes sense to me, thanks 👍

aldesantis (approved):
Looking good, thanks @kennyadsl!

kennyadsl:
@jarednorman will wait for your review before merging if you have some time. Please let me know!

jarednorman (approved):

PR #3832, opened 2020-11-05
Lock Rails version to < 6.1.x

kennyadsl:
**Description**

This way it doesn't try to use Rails 6.1, which has not yet been released yet and we are not yet compatible with. Not sure why the CI isn't failing on master, I suspect it's for some cached gems.

This is causing all extensions specs to fail and needs to be backported in all versions that are compatible with Rails 6.


jarednorman (approved):

aldesantis (approved):

kennyadsl:
Merging without waiting 24 hours, this is breaking all extensions specs.

PR #3833, opened 2020-11-06
Fix install instructions in Solidus Guides

ikraamg:
**Description**

As adopted from @kennyadsl commit on the edge guides solidusio/edgeguides#12:

Using --skip-webpack-install is only preventing the Rails install
generator from running webacker:install, but the webpacker gem is
still added to the Gemfile causing an error when starting the server:
'Webpacker configuration file not found'

In the context of a fresh app that uses the bare minimum set of tools to
run Solidus we don't need that. People can still add it later if needed.

**Checklist:**
~~- [ ] I have added tests to cover this change (if needed)~~

### DIFF
--- a/guides/source/developers/getting-started/first-time-installation.html.md
+++ b/guides/source/developers/getting-started/first-time-installation.html.md
@@ -57,12 +57,13 @@ You can find more details on Solidus' dependencies in the
 First, we need a new Rails project:
 
 ```bash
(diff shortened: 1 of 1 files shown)


kennyadsl (approved):

spaghetticode (approved):
@ikraamg thanks! 🎉

PR #3835, opened 2020-11-08
Update order_tabs Order number format

brchristian:
Many Solidus objects have a "number" field, including Orders, Shipments, Payments, etc. Counterintuitively this "number" generally begins with a letter. :) It’s "R" in the case of Orders, "H" in the case of Shipments, etc.

In almost all cases we simply display this as a string:

Screen Shot 2020-11-08 at 2 16 54 PM

Screen Shot 2020-11-08 at 2 21 54 PM

Screen Shot 2020-11-08 at 2 21 46 PM

However, there is one exception, which is the Order breadcrumb, which prepends a "#" symbol:

Screen Shot 2020-11-08 at 2 22 49 PM

This PR removes the "#" in the Order breadcrumb for consistency with other Order views as well as Shipments, Payments, etc.


jarednorman (approved):
Thanks!

spaghetticode (approved):
makes sense, thanks @brchristian 👍

kennyadsl:
Thanks @brchristian, we will merge this in Solidus 3.1!

PR #3836, opened 2020-11-09
Update governance with latest changes to the organization

kennyadsl:
With time, the stakeholders' group is changing allowing us to transform this organization into something more mature and more close to what we actually need for Solidus. Thus, we recently changed in an official way the structure of the group and, for transparency, we are reporting here what changed.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -49,7 +49,7 @@ New Core Team members are appointed by the Core Team.
 
 To apply for the Core Team, you must be proposed by an existing member.
 
-If you don't know anyone, simply send a Slack DM to Jacob Herrington (@Jacob Herrington), explaining
+If you don't know anyone, simply send a Slack DM to Alberto Vena (@kennyadsl), explaining
 why you think you should be in the Core Team and what kind of contribution you expect to bring.
 
 ### Leaving the Core Team
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -9,7 +9,12 @@ This file documents Solidus' governance model. It is meant for all users, but pa
 ## Project Director
 
 At present, [Nebulab](https://www.nebulab.it) is the main code contributor and director of Solidus.
-They provide technical guidance and coordinate community efforts and activities.
+
+Project Directors are in charge of the organizational and business aspects of the platform. They handle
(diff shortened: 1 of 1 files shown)


spaghetticode (approved):
@kennyadsl thanks!

aldesantis:
@kennyadsl should we also clarify the responsibilities of each group, as explained in the new Google Doc we wrote? I think it would be helpful to clear up some confusion.

kennyadsl:
@aldesantis I was thinking the same but the document already covers Core Team and Project Director in the sections above and it doesn't make a lot of sense to re-explain what is already stated in other paragraphs. That's why I omitted that part but I think the definitions of Advisors and Partners are still needed. 

I'll copy here the text from the document, just as a reference:

```text
Core team: the core team is ultimately responsible for the technical direction of Solidus. Members of the core team make the final decision as to what goes into the core and any other non-marketing material (e.g. extensions, development tools) hosted in the official Solidus GitHub organizations, although they may delegate part of this responsibility to other entities (e.g. a specific company may be put in charge of maintaining an extension it has originally developed).

Project directors: the project directors are in charge of the organizational and business aspects of the platform. They handle the social media communication and marketing/business direction, and advertise Solidus to the software community and larger eCommerce industry. The project directors commit to hearing all interested parties around sensitive business development decisions, although they reserve the right to always act in the best interest of the platform and its community.

Partners: the partners are individuals or companies that are financially contributing to the platform’s Open Collective organization, in exchange for a pre-established set of benefits, mainly the ability to vote during stakeholder meetings.

Advisors: the advisors are individuals or companies that consistently provide value to the Solidus community through their industry experience, professional network or other means. They participate in the stakeholder meetings and other stakeholder ceremonies with the purpose of informing the group’s decisions and the platform’s direction. Advisors hold no decisional power. They are elected during stakeholder meetings and their status must be re-confirmed every 6 months.
```

aldesantis:
@kennyadsl sorry, what I meant is that the new document does a much better job than the current governance at explaining the separation of responsibilities between project directors, core team and stakeholders. I think we should find a way to integrate the governance with that information by replacing the descriptions of each group's responsibilities with the new version.

MFRWDesign:
LGTM

jarednorman (approved):

Issue #3802, opened 2020-10-21
Improve developer experience when changing payment methods.


kennyadsl:
It happens that stores change their payment method over the years. When switching to a new extension it's a good practice to remove the old one once the migration is done. But by doing so, we'll have a lot of references in the database that points, through STI, to a class that doesn't exist anymore. 

When this happens, we should find a way to avoid the application to raise an exception similar to:

```
ActiveRecord::SubclassNotFound (The single-table inheritance mechanism failed to locate the subclass: 'Solidus::Gateway::BraintreeGateway'. This error is raised because the column 'type' is reserved for storing the class in case of inheritance. Please rename this column if you didn't intend it to be used for storing the inheritance class or overwrite Spree::PaymentMethod.inheritance_column to use another column for that information.)
```

I'm not sure it's possible, I'd make an attempt to explore some possible solutions though. 

**Solidus Version**:
All solidus versions

**To Reproduce**

See this issue as an example: https://github.com/solidusio/solidus_paypal_braintree/issues/184

**Additional context**

An alternative could be documenting best practices for switching from a payment method to another in our guides, in order to point users to a good resource when they are facing this scenario.

peterberkenbosch:
@kennyadsl this has been on my radar as well here: https://github.com/solidusio-contrib/solidus_affirm_v2/issues/9

fthobe:
@jarednorman solved

PR #3837, opened 2020-11-09
Improve payment service providers switching errors

luca-landa:
**Description**

This PR is a solution for https://github.com/solidusio/solidus/issues/3802.

Since old payment methods can't just be deleted after switching payment service provider, I provided a script to reset their `type` to a default `Spree::PaymentMethod`, to avoid breaking Single Table Inheritance while building `ActiveRecord` instances.
I added a `type_before_removal` field which will store the actual `type` value when the script is run. 

Also, the script sets `active`, `available_to_admin` and `available_to_users` flags to false.

Moreover, I replaced the typical  `ActiveRecord::SubclassNotFound` with a more meaningful error, which should save time to developers facing this issue and points them to the script.

If I got something wrong please tell me, this was my first experience with the project :)

**Checklist:**

### DIFF
--- a/core/app/models/spree/payment_method.rb
+++ b/core/app/models/spree/payment_method.rb
@@ -13,6 +13,7 @@ module Spree
   #
   class PaymentMethod < Spree::Base
     include Spree::Preferences::Persistable
+    class UnsupportedPaymentMethod < StandardError; end
 
     preference :server, :string, default: 'test'
     preference :test_mode, :boolean, default: true
@@ -59,6 +60,16 @@ module Spree
     class << self
       def model_name
         ModelName.new(self, Spree)
+      end
+
+      def find_sti_class(type_name)
+        super(type_name)
+      rescue ActiveRecord::SubclassNotFound
+        raise UnsupportedPaymentMethod, "Found invalid payment type '#{type_name}'.\n"\
+          "This may happen after switching payment service provider, when payment methods "\
+          "reference old types that are not supported any more.\n"\
+          "If that is the case, consider running 'rake payment_method:deprecate_unsupported_payment_methods' "\
+          "to fix the issue."
       end
     end
 
--- a/core/lib/tasks/payment_method.rake
+++ b/core/lib/tasks/payment_method.rake
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+namespace :payment_method do
+  desc "Deactivates old payment methods and fixes ActiveRecord::SubclassNotFound error, "\
+  "which happens after switching Payment Service Provider."
+  task deactivate_unsupported_payment_methods: :environment do
+    Spree::PaymentMethod.pluck(:id, :type).select do |id, type|
+      ActiveSupport::Dependencies.constantize(type)
+    rescue NameError
+      fix_payment_method_record(id, type)
+    end
+  end
+
+  def fix_payment_method_record(id, previous_type)
+    connection = ActiveRecord::Base.connection
+    false_value = connection.quoted_false
+    connection.exec_update(<<-SQL
+      UPDATE spree_payment_methods
+      SET
+        type='#{Spree::PaymentMethod.name}',
+        type_before_removal='#{previous_type}',
+        active=#{false_value},
+        available_to_users=#{false_value},
+        available_to_admin=#{false_value}
+      WHERE id=#{id};
+    SQL
+    )
+  end
+end
--- a/core/spec/models/spree/payment_method_spec.rb
+++ b/core/spec/models/spree/payment_method_spec.rb
@@ -233,4 +233,20 @@ RSpec.describe Spree::PaymentMethod, type: :model do
       end
     end
   end
+
+  describe "#find_sti_class" do
+    context "with an unexisting type" do
+      context "while retrieving records" do
+        let!(:unsupported_payment_method) { create(:payment_method, type: 'UnsupportedPaymentMethod') }
+
+        it "raises an UnsupportedPaymentMethod error" do
+          expect { Spree::PaymentMethod.all.to_json }
+            .to raise_error(
+              Spree::PaymentMethod::UnsupportedPaymentMethod,
+              /Found invalid payment type 'UnsupportedPaymentMethod'/
+            )
+        end
+      end
+    end
+  end
 end
--- a/guides/source/developers/payments/payment-service-providers.html.md
+++ b/guides/source/developers/payments/payment-service-providers.html.md
@@ -66,3 +66,16 @@ You would need to extend or rewrite this class with your preferred PSP
 integration.
 
 [credit-card-base]: https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment_method/credit_card.rb
+
+### Switching payment service provider
+
+After switching payment service provider, there may be `Spree::PaymentMethod` 
+records referencing a `type` class that does not exist anymore. Trying to 
+retrieve these records through an `ActiveRecord` query raises a 
+`Spree::PaymentMethod::UnsupportedPaymentMethod` error.
+
+If you cannot delete these records, you can deactivate them running 
+`rake payment_method:deactivate_unsupported_payment_methods`. 
+This way, their `type` will be set to `Spree::PaymentMethod`, allowing for 
+records retrieval without errors. Also, their real `type` value will be stored
+in the `type_before_removal` attribute.
--- a/core/db/migrate/20201127212108_add_type_before_removal_to_spree_payment_methods.rb
+++ b/core/db/migrate/20201127212108_add_type_before_removal_to_spree_payment_methods.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddTypeBeforeRemovalToSpreePaymentMethods < ActiveRecord::Migration[5.2]
+  def change
+    add_column :spree_payment_methods, :type_before_removal, :string
+  end
+end
--- a/core/spec/lib/tasks/payment_method_spec.rb
+++ b/core/spec/lib/tasks/payment_method_spec.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'payment_method:deactivate_unsupported_payment_methods' do
+  include_context(
+    'rake',
+    task_name: 'payment_method:deactivate_unsupported_payment_methods',
+    task_path: Spree::Core::Engine.root.join('lib/tasks/payment_method.rake'),
+  )
+
+  let!(:unsupported_payment_method_id) { 0 }
+  let!(:unsupported_payment_method) { create(:payment_method, id: unsupported_payment_method_id) }
+  let!(:supported_payment_method) { create(:payment_method, id: 1) }
+
+  def unsupported_payment_method_reloaded
+    Spree::PaymentMethod.find_by(id: unsupported_payment_method_id)
+  end
+
+  before do
+    unsupported_payment_method.update type: 'UnsupportedPaymentMethod'
+  end
+
+  context "with an unsupported payment method" do
+    it "allows payment method records retrieval" do
+      task.invoke
+
+      expect {
+        Spree::PaymentMethod.find_by(id: unsupported_payment_method_id)
+      }.not_to raise_error
+    end
+  end
+
+  context "on an unsupported payment method" do
+    before { task.invoke }
+    subject { unsupported_payment_method_reloaded }
+
+    it "sets payment method type to 'Spree::PaymentMethod'" do
+      expect(subject.type).to eq 'Spree::PaymentMethod'
+    end
+
+    it "sets payment method type_before_removal correctly" do
+      expect(subject.type_before_removal).to eq 'UnsupportedPaymentMethod'
+    end
+
+    it "resets payment method active flag" do
+      expect(subject.active).to be false
+    end
+
+    it "resets payment method available_to_users flag" do
+      expect(subject.available_to_users).to be false
+    end
+
+    it "resets payment method available_to_admin flag" do
+      expect(subject.available_to_admin).to be false
+    end
+  end
+
+  context "on a supported payment method" do
+    it "does not change payment method attributes" do
+      expect { task.invoke }.not_to change { supported_payment_method.reload }
+    end
+  end
+end
--- a/guides/source/developers/payments/payment-methods.html.md
+++ b/guides/source/developers/payments/payment-methods.html.md
@@ -30,6 +30,10 @@ service providers][payment-service-providers] article.
 - `available_to_users`: Determines if the payment method is visible to users.
 - `available_to_admin`: Determines if the payment method is visible to
   administrators.
+- `type_before_removal`: Contains the previous real payment type, in case `type` has 
+  been removed after switching Payment Service Provider. Defaults to `nil`. For more 
+  information, see the [Payment service providers][payment-service-providers] 
+  article.
 
 [payment-method-base]: https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment_method.rb
 [payment-service-providers]: payment-service-providers.html


luca-landa:
My bad, I didn't pay attention to failing checks.
I currently have no idea of what is causing the errors, I have run the full suite locally on postgresql without any trouble.
I will investigate and fix it.

spaghetticode:
@luca-landa I'm wondering if we may be interested in keeping for reference the original payment method information.

For example, if we change the boolean field to a string, we can populate it with the previous `type` class name... what do you think?

luca-landa:
@spaghetticode it seems a good idea to keep track of the previous `type`.

But personally I would add it in another field, instead of converting the `deprecated` field. 
I thought that the new flag could be useful to query for all current payment methods or for the deprecated ones. For these purposes, I wouldn't find it intuitive to write `.where.not(previous_type: nil)`, I would much rather do `.where(deprecated: false)` instead.

Do you think it can be a better solution?

spaghetticode:
@luca-landa I don't have a strong opinion on this, I think there are pros and cons with both approaches. 

I don't see `where.not(previous_type: nil)` as a big issue, as it would probably just end up in a scope  such as
`scope :deprecated, -> {where.not(deprecated_type: nil)}`
Same goes for instance methods such as
```ruby
def deprecated?
  deprecated_type.present?
end
```
which would need to be explicitly written instead of relying on AR automatic definitions. The advantage in this case would be we just need to maintain a single field on the DB instead of 2.  Also, we already have the boolean field `active`, adding another one can potentially make things a bit messy when they're not kept in sync.

On the other hand, the burden to maintain 2 fields is rather minimal, and it would probably allow us to write simpler code in ruby.

luca-landa:
@spaghetticode I agree on the potential problem of keeping `active` and `deprecated` in sync, I didn't really consider it at first and I'd like to avoid the risk.

I like your proposal with `deprecated_type` and the scope you defined (since `active` will help filter out deprecated records).

But I have another change in mind that would make the AR scope a bit more readable, and it would make the record more explicitly marked as "deprecated".
What if the new `type` set by the task is a new `Spree::PaymentMethod::DeprecatedMethod`model class, instead of `Spree::PaymentMethod`? The scope would also become `scope :deprecated, -> {where(type: "Spree::PaymentMethod::DeprecatedMethod")}`.

I have another (OT) question about this PR: I don't know how exactly should I fix an error that I made, I could add another commit but it wouldn't satisfy the guidelines request that the specs must pass for each commit. Should I create another PR for another branch?

core/db/migrate/20201108142424_add_deprecated_to_payment_method.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddDeprecatedToPaymentMethod < ActiveRecord::Migration[6.0]
+  def change
+    add_column :spree_payment_methods, :deprecated, :boolean, default: false
```
jarednorman, line 5: I don't think "deprecated" is the right name for this field. "Deprecated" means "usable but regarded as obsolete and best avoided, typically due to having been superseded". I think in this case the payment method will no longer be usable, so we should find a better way to describe it.

jarednorman (changes requested):
I think we should find a better name for the "deprecated" column that aligns better with how it works. Otherwise, this seems like a reasonable approach if the method we're overriding has the desired effect.

core/lib/tasks/payment_method.rake
```diff
@@ -0,0 +1,59 @@
...
+    # Let ActiveRecord handle boolean format, which may vary
+    # depending on the DBMS in use.
+    set_unsupported_payment_methods_state(unsupported_ids)
+  end
+
+  def set_unsupported_payment_methods_type(unsupported_ids)
+    where_clause = get_where_clause('id', unsupported_ids)
+    ActiveRecord::Base.connection.execute(
+      "UPDATE spree_payment_methods "\
+      "SET type='#{Spree::PaymentMethod.name}' "\
```
kennyadsl, line 19: Why use `.name` instead of a string? Not saying it's wrong, I just would like to understand the reasoning behind it.

core/lib/tasks/payment_method.rake
```diff
@@ -0,0 +1,59 @@
...
+      "WHERE #{where_clause};"
+    )
+  end
+
+  def set_unsupported_payment_methods_state(unsupported_ids)
+    Spree::PaymentMethod
+      .where(id: unsupported_ids)
+      .update(deprecated: true,
+              active: false,
```
kennyadsl, line 28: Is there a reason why we can't set the type here as well?

core/lib/tasks/payment_method.rake
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+namespace :payment_method do
+  desc "Deprecates old payment methods and fixes ActiveRecord::SubclassNotFound error, "\
+  "which happens after switching Payment Service Provider."
+  task deprecate_unsupported_payment_methods: :environment do
+    unsupported_ids = get_unsupported_payment_method_ids
```
kennyadsl, line 7: Maybe I'm missing something but what about removing these methods and just leave the code to execute the change when the exception is raised? Something like:

```ruby
    Spree::PaymentMethod.pluck(:type).select do |type|
      begin
        ActiveSupport::Dependencies.constantize(type)
      rescue NameError
        # do the updates here
      end
    end
```

core/app/models/spree/payment_method.rb
```diff
@@ -11,6 +11,8 @@ module Spree
   # This class is not meant to be instantiated. Please create instances of concrete payment methods.
   #
   class PaymentMethod < Spree::Base
+    class UnsupportedPaymentMethod < StandardError; end
```
kennyadsl, line 14: I think this could be a good spot for https://api.rubyonrails.org/classes/ActiveSupport/ActionableError/ClassMethods.html, which is the same class used to run migrations directly in the browser when needed. What do you think?

guides/source/developers/payments/payment-service-providers.html.md
```diff
@@ -64,3 +64,13 @@ You would need to extend or rewrite this class with your preferred PSP
 integration.
 
 [credit-card-base]: https://github.com/solidusio/solidus/blob/master/core/app/models/spree/payment_method/credit_card.rb
+
+### Switching payment service provider
```
kennyadsl, line 68: 👍 adding this documentation, thanks!

kennyadsl (commented):
Thanks! I left some questions on the implementation. Looking forward to know what you think about them. 

 Also, can you please squash the last commit into the one that introduced the code style issue?

🙏

core/db/migrate/20201108142424_add_deprecated_to_payment_method.rb
luca-landa: As I said discussing with @spaghetticode, I'd like the db row to be clearly marked as "deprecated" (or another term). I didn't understand from the solidus docs in which cases should a payment method be flagged with `active = false`, and nor if there are cases where it makes sense to set it to `false` at some point and again to `true` after a while. So, I wasn't really sure that the `active` could be used for this case.

Given that, if you think that `active` can be used to express this state I would be glad to remove the new flag, or even better to change it into `previous_type` as suggested from @spaghetticode to keep track of the previous real type of the payment method.

core/lib/tasks/payment_method.rake
luca-landa: I usually prefer to reference constants or variables when possible, just in case they may change in the future (even tough `Spree::PaymentMethod` is very unlikely to be renamed). Even having the tests, being able to reference a class to fix an error that requires that field to contain a valid class name seemed a bit safer to me.

core/lib/tasks/payment_method.rake
luca-landa: That's the point of this fix actually: even running a query `Spree::PaymentMethod.where(...)` before changing the `type` via raw sql would make AR immediately raise the STI error, while creating the result instances, so `type` needs to be changed before that.
So, I could have updated the record fully in the first raw sql update, but I preferred to separate this second update to let AR handle boolean values via its connection adapter.

core/lib/tasks/payment_method.rake
luca-landa: I'm a bit confused by what you quoted, I'll try to answer for what I understood.
I personally wouldn't apply the fix automatically when the STI error happens: that error reports an inconsistent situation on the db and fixing it actually changes a bit of the behavior of the application (querying for payment methods will return instances of another class after that).
I don't like the idea of this happening without the store devs understanding this problem, and I thought that they may need to take more actions than just this when this problem occurs.

But I lack of experience on this project, so please correct me if my assumption is wrong.

About the code you inserted: that rescue instruction is where the task understands what types are not supported anymore, and not where the original error is caught to raise a more understandable error.
The error is caught in `Spree::PaymentMethod#find_sti_class,` so it should be caught there in order to make this automatic. Did I understand correctly what you meant?

core/app/models/spree/payment_method.rb
luca-landa: It is a good idea, I actually didn't think about it at first (and didn't know about it until now). 

But that's a Rails 6 feature and Solidus minimum requirement is for Rails 5.0.0, so it can't be implemented, am I right?

spaghetticode:
> But I have another change in mind that would make the AR scope a bit more readable, and it would make the record more explicitly marked as "deprecated".
> What if the new `type` set by the task is a new `Spree::PaymentMethod::DeprecatedMethod`model class, instead of `Spree::PaymentMethod`? The scope would also become `scope :deprecated, -> {where(type: "Spree::PaymentMethod::DeprecatedMethod")}`.

I guess that would make sense, although it's not strictly needed... but I think it may make it easier to discriminate _removed_ payment methods and their payments? If while doing the PR you can come up with good reasons for doing (or not doing) it, you can add some comments in the commit message, that would be very valuable.

> I have another (OT) question about this PR: I don't know how exactly should I fix an error that I made, I could add another commit but it wouldn't satisfy the guidelines request that the specs pass for each commit. Should I create another PR for another branch?

No need, you can make all the fixes and changes you need in this PR, as long as they're related to it. 
You can also rebase, do whatever makes sense so that, eventually, commits make sense when we'll need to navigate the git history and understand the changes.

luca-landa:
Well the `Spree::PaymentMethod::DeprecatedMethod` class was just a proposal, I don't see any more good reasons than those I already explained, but I actually see a reason for not doing it: we would be adding a subclass which does not define any specific behavior.

I may just discard the idea, but stick with the idea of changing `deprecated` to `previous_type` if the other reviewers approve this change too.

core/app/models/spree/payment_method.rb
kennyadsl: Yes, you are right, but if we like the idea we could conditionally include `ActiveSupport::ActionableError` and its related code only when that class is defined. Up to you!

core/db/migrate/20201108142424_add_deprecated_to_payment_method.rb
kennyadsl: These concerns make sense and I don't have a clear answer as well. Just thinking aloud: is there any occurrence when we want `active` and `deprecated` to have different values at the same time. I mean does it make sense having `deprecated: true` without also having `active: false`? I don't think so, since the payment method class has been removed from the project, having that payment method active would just produce this error more often. This makes me think that active is just fine for this purpose as well. 

If we also add `previous_type`, this new field populated would be enough to understand which records are in this specific state.

But again, I don't have strong opinions here and I'd leave this decision to you. I'm fine with both as soon as we don't use the `deprecated` term for the reasons above.

core/lib/tasks/payment_method.rake
kennyadsl: Sorry, my comment wasn't clear at all: I meant to do something like this directly in the rake task:

```ruby
  task deprecate_unsupported_payment_methods: :environment do
    Spree::PaymentMethod.pluck(:id, :type).each do |id, type|
      begin
        ActiveSupport::Dependencies.constantize(type)
      rescue NameError
        # we have the id of the problematic type here, can we avoid the extra query and just update the record?
      end
    end
  end
```

core/lib/tasks/payment_method.rake
kennyadsl: Sure, you are right. I'd try to make a single query though, maybe we can move these attributes settings to the raw SQL update?

kennyadsl (commented):
Left a second review with other thoughts. Let me know if those make sense. 

Another thing: I was wondering if there's something we could do when deleting a payment method. I'm thinking of a simple PORO that handles the PaymentMethod destruction and maybe sets these attributes for us when we (soft-)delete a payment method? This way we will also prevent this error from happening, anyway, maybe it's something for another PR. 🙂

core/db/migrate/20201108142424_add_deprecated_to_payment_method.rb
luca-landa: Then adding `previous_type` and removing `deprecated` seems a good choice. I'm just not 100% sure on the name, can you think of something more expressive and not too long? Maybe `removed_type`?

core/lib/tasks/payment_method.rake
luca-landa: Updating in a single query (and every single row individually, as you suggested below), would make the code something like this.
```ruby
def update_row(id)
    connection = ActiveRecord::Base.connection
    false_value = connection.adapter_name[/sqlite/i] ? '0' : 'false'
    true_value = connection.adapter_name[/sqlite/i] ? '1' : 'true'
    connection.exec_query(
      "UPDATE spree_payment_methods "\
      "SET type='#{Spree::PaymentMethod.name}', deprecated=#{true_value}, "\
      "active=#{false_value}, available_to_users=#{false_value}, available_to_admin=#{false_value} "\
      "WHERE id=#{id}"
    )
end
```
The task becomes really shorter, I like it, but I don't find really nice to manually choose what to write depending on the db in use (it's a db adapter's job), since doing it via AR would be more stable (for example, in case another db will be supported in the future that another "true" value).
Do you think I'm just worrying too much? It's unlikely to support another db that will require another boolean value different from "0"/"1" and "false"/"true", isn't it?

core/lib/tasks/payment_method.rake
luca-landa: It's possible, the records will be a few, so I can see no performance issue. It would make much more sense merging the two update statements in the first one, as you asked above. I'll wait for you to answer to that before changing this.

core/app/models/spree/payment_method.rb
luca-landa: Just to be sure, you mean like a module to include that could include all the stuff related to this PR? Including a static method to do the actual updates, the `ActiveSupport::ActionableError` and maybe the overriding of `find_sti_class` to catch the actual error?

core/lib/tasks/payment_method.rake
spaghetticode: ActiveRecord provides facilities for avoiding hardcoding those values, you can use `ActiveRecord::Base.connection.quoted_true` and `ActiveRecord::Base.connection.quoted_false`, this should make the SQL generic enough IMO.

core/app/models/spree/payment_method.rb
kennyadsl: I mean something like:

```ruby
class UnsupportedPaymentMethod < StandardError
  if defined?(ActiveSupport::ActionableError)
    include ActiveSupport::ActionableError

    action "Bla bla" do
      Our::Rake::Task.execute
    end
  end
end

File: core/db/migrate/20201108142424_add_deprecated_to_payment_method.rb (hunk 1 above)
Comment (luca-landa) [reply core/db/migrate/20201108142424_add_deprecated_to_payment_method.rb]: I changed `deprecated` to `previous_type` and updated the doc accordingly.

File: core/lib/tasks/payment_method.rake (hunk 3 above)
Comment (luca-landa) [reply core/lib/tasks/payment_method.rake]: Thanks! I couldn't find it myself, this helped me changing the script as suggested by @kennyadsl, I just force pushed the change.

File: core/app/models/spree/payment_method.rb (hunk 5 above)
Comment (luca-landa) [reply core/app/models/spree/payment_method.rb:14]: I tried to implement it without success: the button doesn't appear if the error was raised in `ActiveRecord.find_sti_class`. I could only find base examples of this feature usage and nothing in official docs that helped my solve my specific case. I would leave this for another PR for now (and maybe I'd open an issue after a bit more digging).

File: core/db/migrate/20201118200630_add_previous_type_to_spree_payment_methods.rb [hunk 7]
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPreviousTypeToSpreePaymentMethods < ActiveRecord::Migration[5.2]
+  def change
+    add_column :spree_payment_methods, :previous_type, :string
```
kennyadsl, line 5: What about `type_before_removal`? I think it expresses better what this field contains.

core/lib/tasks/payment_method.rake
```diff
@@ -0,0 +1,27 @@
...
+    end
+  end
+
+  def fix_payment_method_record(id, previous_type)
+    connection = ActiveRecord::Base.connection
+    false_value = connection.quoted_false
+    connection.exec_update(<<-SQL
+      UPDATE spree_payment_methods
+      SET type='#{Spree::PaymentMethod.name}', previous_type='#{previous_type}',
+      active=#{false_value}, available_to_users=#{false_value}, available_to_admin=#{false_value}
```
kennyadsl, line 22: Not mandatory, but I'd move all these attributes to separate lines. This will allow us to add/change/remove individual lines in the future without impacting git history for all of them.

guides/source/developers/payments/payment-service-providers.html.md
```diff
@@ -64,3 +64,13 @@ You would need to extend or rewrite this class with your preferred PSP
...
+
+After switching payment service provider, there may be `Spree::PaymentMethod` 
+records referencing a `type` class that does not exist anymore.
+
```
kennyadsl, line 72: I'd specify that this prevents the system to raise the `ActiveRecord::SubclassNotFound` error. This way, it will also be easier to find if someone is searching using the exception name as a keyword.

kennyadsl (commented):
Left some other comments, we are almost there, thanks again!

core/db/migrate/20201118200630_add_previous_type_to_spree_payment_methods.rb
luca-landa: It sounds better actually, fixed.

core/lib/tasks/payment_method.rake
luca-landa: Fixed it, actually it was a bit ugly before.

guides/source/developers/payments/payment-service-providers.html.md
luca-landa: Nice suggestion, I didn't consider it at first. I just added it.

core/spec/models/spree/payment_method_spec.rb
```diff
@@ -395,4 +395,22 @@ def gateway_class
       expect(described_class::DISPLAY).to eq([:both, :front_end, :back_end])
     end
   end
+
+  describe "#find_sti_class" do
+    context "with an unexisting type" do
+      context "while retrieving records" do
+        let!(:unsupported_payment_method) { create(:payment_method, type: 'UnsupportedPaymentMethod') }
+        before do
+          unsupported_payment_method.update type: 'UnsupportedPaymentMethod'
```
spaghetticode, line 404: I think this `before` block has no effect on the record, as it's updating the type but the value is the same as the initial one from line 402.

luca-landa: Right, I removed the block.

spaghetticode:
@luca-landa the PR is almost ready now, thanks for the work you've done so far! There are still a couple of cosmetic issues raised by Hound, and this [merge commit](https://github.com/luca-landa/solidus/commit/fb03992a0fbd0212d9f1935ba414700acb792007) that should be dealt with.

luca-landa:
@spaghetticode I'm sorry I completely lost track of this PR 😅 it should be fixed now

spaghetticode (approved):
LGTM, thanks Luca!

waiting-for-dev:
Hey @jarednorman, I think [your concern](https://github.com/solidusio/solidus/pull/3837#pullrequestreview-530851981) was addressed :slightly_smiling_face:  Is there anything else we need to tackle here?

jarednorman (approved):
No, I'm happy here.

PR #3563, opened 2020-03-20
Uniform bill_address and ship_address behaviour in Spree::UserAddressBook module

oldjackson:
Responding to the issue [#3539](https://github.com/solidusio/solidus/issues/3539), the association `has_one :bill_address` provided by `UserAddressBook` to `LegacyUser` has been made similar to the one already in place for `default_address` _alias_ `ship_address`.

Namely, `user.bill_address` is now the `Address` pointed by the only `UserAddress` having the new `default_billing` flag column set to `true`, which plays the same role as the `default` one (which retains the function of selecting the one and only default _shipping_ address).

Being now possible to set the two default addresses independently, 

- `#persist_order_address` now calls `#save_in_address_book` in the same way for the two addresses, solving the issue.

- the new public methods `#mark_default_ship_address` and `#mark_default_bill_address` are provided. `#mark_default_address` just calls the former for backward compatibility, and can be deprecated.

PR #3838, opened 2020-11-11
Migrate default billing addresses to address book

mamhoff:
**Description**

`Spree::User` records have a `bill_address_id` that denotes their saved
billing address to be reused in the next checkout. That column, however,
is not in use since Solidus 2.11.0. Instead, we rely on a
`default_billing` Boolean on the `spree_user_addresses` join table.

This migration sets the `default_billing` flag on that join table for
addresses matching the user's default billing address and the user's ID.

**Checklist:**

### DIFF
--- a/core/lib/tasks/upgrade.rake
+++ b/core/lib/tasks/upgrade.rake
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+namespace :solidus do
+  namespace :upgrade do
+    desc "Upgrade Solidus to version 2.11.0"
+    task two_point_eleven: [
+        'solidus:migrations:migrate_default_billing_addresses_to_address_book:up'
+      ] do
+      puts "Your Solidus install is ready for Solidus 2.11.0"
+    end
+  end
+end
--- a/core/spec/lib/tasks/migrations/migrate_default_billing_addresses_to_address_book_spec.rb
+++ b/core/spec/lib/tasks/migrations/migrate_default_billing_addresses_to_address_book_spec.rb
@@ -1,0 +1,103 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'solidus:migrations:migrate_default_billing_addresses_to_address_book' do
+  describe 'up' do
+    include_context(
+      'rake',
+      task_path: Spree::Core::Engine.root.join('lib/tasks/migrations/migrate_default_billing_addresses_to_address_book.rake'),
+      task_name: 'solidus:migrations:migrate_default_billing_addresses_to_address_book:up',
+    )
+
+    context "migrate from Solidus 2.11" do
+      let(:user1) { create(:user) }
+      let(:user2) { create(:user) }
+      let(:bill_address1) { FactoryBot.create(:address) }
+      let(:bill_address2) { FactoryBot.create(:address) }
+      let(:bill_address3) { FactoryBot.create(:address) }
+
+      before do
+        # Set two billing addresses for User1, the second address is the default.
+        user1.save_in_address_book(bill_address1.attributes, false, :billing)
+        user1.save_in_address_book(bill_address2.attributes, true, :billing)
+
+        # Update the "bill_address_id" for user1 to be different from the address_book's default address.
+        user1.update!(bill_address_id: bill_address1.id)
+
+        # Set user2's bill address using old `bill_address_id` method.
+        user2.save_in_address_book(bill_address3.attributes, false, :billing)
+        user2.update!(bill_address_id: bill_address3.id)
+        Spree::UserAddress.where(user_id: user2.id).first.update!(default_billing: false)
+      end
+
+      it 'runs' do
+        expect { task.invoke }.to output(
+          "Migrating default billing addresses to address book ... Success\n"
+        ).to_stdout
+      end
+
+      it "does not migrate a user's `bill_address_id` when a user already has a default `bill_address` in the address book" do
+        task.invoke
+        expect(user1.bill_address_id).not_to eq bill_address2.id
+        expect(user1.bill_address).to eq bill_address2
+      end
+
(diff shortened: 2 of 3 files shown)


kennyadsl:
Ref: https://github.com/solidusio/solidus/pull/3563

tvdeyen (approved):
Approving although this ideally should be a data migration that runs in a potential `solidus:upgrade` rake task. We had this rake task once (back in 1.x times) but this was unfortunately removed.

Anyhow. Thanks 👍

core/db/migrate/20201111133031_migrate_default_billing_addresses_to_address_book.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class MigrateDefaultBillingAddressesToAddressBook < ActiveRecord::Migration[5.2]
+  def up
+    adapter_type = connection.adapter_name.downcase.to_sym
```
tvdeyen, line 5: Thinking about this again, I think we should add a safe guard here. Since we already shipped this code maybe someone already migrated their data.

kennyadsl: This is true also for people that already updated to 2.11 and has some address with default_billing to true, isn't it?

mamhoff: Yes. I've added a safeguard so that the migration does not do anything if any default_billing flags are already set.

kennyadsl: But that only avoids the migration (and the fix) to run for people that used 2.11 for some time before this will be released, isn't it? The ideal would be something that fixes their database as well. Do you think it's possible?

kennyadsl: @mamhoff any thoughts on this? Not sure I was clear enough, I meant that I think this will prevent the issue to happen only for users upgrading from Solidus `2.10.x` to `2.11.4` (assuming `2.11.4` will contain this fix), but doesn't cover who upgrades from `2.11.3` to `2.11.4`. Does it make sense?

ikraamg:
Hi, this is a great bug catch 🐛  👍 . below I have a suggestion  in order to account for any codebase that already uses the `default_billing`,  should correctly be able to migrate and rollback

mamhoff:
Hey @ikraamg thank you - However, for a very large Solidus installation, this will be prohibitively slow. I've purposefully set this to run in pure SQL, and I'm sure it's possible to do this without iterating over ActiveRecord collections. Sometime this week :)

ikraamg:
Hi @mamhoff, @DanielePalombo helped out with a solution where we exclude the user_addresses that are already using the new format by setting the bill_address_id to nil before running the query:

```ruby 
if Spree::UserAddress.where(default_billing: true).any?

Spree::User.joins(:bill_address).update_all(bill_address_id: nil)

end

....
```

ikraamg:
The task is created in a PR to this branch, however, I am unsure where to run the task as we do not have the `upgrade` task anymore that was previously used for similar tasks. 

https://github.com/mamhoff/solidus/pull/1

mamhoff:
@ikraamg @DanielePalombo Thank you for moving this ahead! I've merged @ikraamg s Pull Request and fixed a few whitespace issues, but I think this should be good to go now :)

core/lib/tasks/upgrade.rake
```diff
@@ -0,0 +1,10 @@
+namespace :solidus do
```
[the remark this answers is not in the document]
kennyadsl: @mamhoff why is this solved?

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Solidus 3.0.0 (master, unreleased)
 
+Migrated default billing addresses to address book
```
kennyadsl, line 3: This should not go into the 3.0 changelog, we need to backport this task into 2.11.x. I'll take care of the changelog, you can remove this entirely.

core/spec/lib/tasks/migrations/migrate_default_billing_addresses_to_address_book_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+      task.invoke
+      expect(user1.bill_address_id).not_to eq bill_address2.id
+      expect(user1.bill_address).to eq bill_address2
+    end
+
+    it "migrates a user's `bill_address_id` when a user does not have a default `bill_address` in the address book" do
+      task.invoke
+      expect(user2.bill_address_id).to eq bill_address3.id
+      expect(user2.bill_address).to eq bill_address3
+    end
```
kennyadsl, line 49: I was talking with @ikraamg on Slack and I'm not sure if he wanted to add some more test cases. Isn't it?

ikraamg: I will add a scenario which covers an upgrade from before 2.11 on another PR to this branch 👍

CHANGELOG.md
ikraamg: Okay, I will remove this when I add the test case.

ikraamg:
@mamhoff I have made the fixes that @kennyadsl requested in new PR: https://github.com/mamhoff/solidus/pull/2. 

Then after a squash here, I think we are ready to merge this PR 👍

kennyadsl:
@mamhoff any chance to take a look at the other PR that @ikraamg submitted on this branch? Let me know if you don't have time and we'll merge this and apply the patches directly with another PR against the main repo. Thanks!

mamhoff:
@ikraamg @DanielePalombo @kennyadsl merged and squashed!

kennyadsl (approved):
Thanks @mamhoff @ikraamg @DanielePalombo !

PR #3840, opened 2020-11-12
Fix permissions to see admin menu items

### LINKED ISSUE #3834: Settings menu is visible in admin login screen before logging in
After a fresh install, the settings menu is visible in the admin login screen. Clicking on any of its items just redirects back to the login screen.

**Solidus Version**: 2.11.1

**To Reproduce**
* Clone the solidus repo (master branch commit 304da6d1a72da1c2014bf809480e0c455ddbefe1)
* Create a new sandbox app following the steps in the readme. 
* Go to localhost:3000/admin/login
* Notice that the settings menu is visible in the sidebar to the left

**Current behavior**

The settings menu is visible when the admin user is not logged in.

**Expected behavior**

The settings menu should not be visible when the admin user is not logged in.

**Screenshots**

Screenshot 2020-11-08 at 20 00 42

kennyadsl:
**Description**

Closes #3834.

Right now a user without roles like logged users or guests user can see the admin menu partially, see https://github.com/solidusio/solidus/issues/3834. 

This behavior has been introduced with https://github.com/solidusio/solidus/commit/295df5e8d92e85b3c5a7fc6c761c32b691b81b75#diff-757abc24f29994c5e06e233c60ca1f3972c2f5019fba628df718eea681adb61fR68 and it happens because a non-admin user actually has permission to show zones and shipping methods; the current code is checking against the `:show` ability, so it seems to be legit.

We use `:admin` for the rest of the menu items checks though, and this is also what we use in the controller to determine if
we can access that page, see: https://github.com/solidusio/solidus/blob/3c8ffcc34f9248b286a9d4ca94d1f9a3197ac7b2/backend/app/controllers/spree/admin/base_controller.rb#L29

Test setup lines have been removed since they are useless now.

**Checklist:**

### DIFF
--- a/backend/spec/features/admin/homepage_spec.rb
+++ b/backend/spec/features/admin/homepage_spec.rb
@@ -73,8 +73,8 @@ describe "Homepage", type: :feature do
     custom_authorization! do |_user|
       can [:admin, :home], :dashboards
       can [:admin, :edit, :index, :show], Spree::Order
-      cannot [:show], Spree::StockLocation
-      cannot [:show], Spree::Zone
+      cannot [:admin], Spree::StockLocation
+      can [:admin], Spree::Zone
     end
 
     it 'should only display tabs fakedispatch has access to' do
@@ -82,7 +82,9 @@ describe "Homepage", type: :feature do
       expect(page).to have_link('Orders')
       expect(page).not_to have_link('Products')
       expect(page).not_to have_link('Promotions')
-      expect(page).not_to have_link('Settings')
+      expect(page).to have_link('Settings')
+      expect(page).not_to have_link('Stock Locations', visible: false)
+      expect(page).to have_link('Zones', visible: false)
     end
   end
 end
--- a/backend/app/views/spree/admin/shared/_settings_sub_menu.html.erb
+++ b/backend/app/views/spree/admin/shared/_settings_sub_menu.html.erb
@@ -3,24 +3,24 @@
     <%= tab :stores, label: :stores, url: spree.admin_stores_path %>
   <% end %>
 
-  <% if can?(:show, Spree::PaymentMethod) %>
+  <% if can?(:admin, Spree::PaymentMethod) %>
     <%= tab :payments, url: spree.admin_payment_methods_path %>
   <% end %>
 
-  <% if can?(:show, Spree::TaxCategory) || can?(:show, Spree::TaxRate) %>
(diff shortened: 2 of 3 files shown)


peterberkenbosch (approved):
👍 nice!

backend/spec/features/admin/homepage_spec.rb
```diff
@@ -73,8 +73,6 @@
     custom_authorization! do |_user|
       can [:admin, :home], :dashboards
       can [:admin, :edit, :index, :show], Spree::Order
-      cannot [:show], Spree::StockLocation
-      cannot [:show], Spree::Zone
```
elia, line 77: Shouldn't these two be changed to:

```rb
      cannot [:admin], Spree::StockLocation
      cannot [:admin], Spree::Zone
```

So we can be sure that the settings tab is still visible even when the user is not able to access some of its sub-resources?

elia (changes requested):

jarednorman (approved):

backend/spec/features/admin/homepage_spec.rb
kennyadsl: Even with the suggested change, the settings tab is not visible. I could change it this way to make sure the settings tab is displayed if one of the inner tab can be accessed but this won't test that the rest of the settings are excluded:

```ruby
context 'as fakedispatch user' do
    before do
      allow_any_instance_of(Spree::Admin::BaseController).to receive(:try_spree_current_user).and_return(nil)
    end

    custom_authorization! do |_user|
      can [:admin, :home], :dashboards
      can [:admin, :edit, :index, :show], Spree::Order
      can [:admin], Spree::Zone
    end

    it 'should only display tabs fakedispatch has access to' do
      visit spree.admin_path
      expect(page).to have_link('Orders')
      expect(page).not_to have_link('Products')
      expect(page).not_to have_link('Promotions')
      expect(page).to have_link('Settings')
    end
  end
```

Let me think about something to test that.

kennyadsl: @elia maybe I've got a possible test, let me know your thoughts!

elia: Thanks @kennyadsl this look like a nice canary to catch if anything is going wrong with permissions without checking all the possible combinations 👍

elia (approved):

spaghetticode (approved):

PR #3844, opened 2020-11-17
Raise canceling a payment when try_void is not implemented

kennyadsl:
**Description**

The previous (deprecated) behavior was still calling cancel without trying to void the payment first.
The actual raise will be handled in Spree::PaymentMethod since it's already present in that class.

**Note:** This is part of #3816, but since it's not trivial I'd like to have a more accurate review from the core team, especially from @tvdeyen. 🙏


tvdeyen (approved):
Yes. Thanks! 🙏

jarednorman (approved):
Looks good! Typo in the commit message though `s/prevoius/previous/`.

PR #3845, opened 2020-11-19
Fix use_legacy_address_state_validator deprecation message

kennyadsl:
To activate the new behavior the value should be false, not true.


jarednorman (approved):

spaghetticode (approved):
@kennyadsl thanks!

PR #3850, opened 2020-11-25
Remove all usage of FooAbility and BarAbility when testing abilities

kennyadsl:
**Description**

This is preparatory work for #3816. We need to remove `BarAbility` that is deprecated, but it's actually still used in the codebase. 

This PR removes its usage and takes advantage of the change to also remove `FooAbility` class, which is very similar but wasn't under the radar in a [previous refactor](https://github.com/solidusio/solidus/pull/2629) because it's not defining a top-level class in our documentation.

Needs backport on v2.11.

**Checklist:**


spaghetticode (approved):
@kennyadsl thank you 👍

aldesantis (approved):

PR #3851, opened 2020-11-27
Remove the gray border inside a product image

mfrecchiami:
It wasn't visible, since there is the gray border applied by CSS, around every product image. Inside the Polo image there was an unwanted extra gray border, now removed.
ruby_polo_back
ruby_polo


spaghetticode (approved):
@mfrecchiami thank you 👍

jarednorman (approved):

PR #3852, opened 2020-12-01
Remove Spree::UserAddress#archived flag

mamhoff:
**Description**

**Checklist:**
- [x] I have ~added~ removed tests to cover this change


kennyadsl (commented):
@mamhoff thanks!

I think that your reasoning makes total sense, it took me a while to get it only reading the PR description because of a mistake:

> Why not remove the join table entry directly? 

Should be

> Why not remove the record from the join table entry directly? 

I think that fixing it could help readers from the future to avoid confusion.

In terms of PR's content, I also can't find a reason not to remove the record, instead of resurrecting an old one but I think we should deprecate the usage of `archived` instead of removing the column, which in my opinion is too invasive and gives for granted that people are not using that field with some custom code. 

I also considered using Discard (which at least would be consistent with the rest of the code) + remove the "resurrecting" logic but honestly, I'd avoid it here since it's just a record on a join table and I can't see any point where we may have some value in soft-deleting.

core/app/models/spree/user_address.rb
```diff
@@ -6,19 +6,25 @@ class UserAddress < Spree::Base
     belongs_to :address, class_name: "Spree::Address", optional: true
 
     validates_uniqueness_of :address_id, scope: :user_id
-    validates_uniqueness_of :user_id, conditions: -> { active.default_shipping }, message: :default_address_exists, if: :default?
```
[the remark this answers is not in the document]
mamhoff: No, I meant in line 8.

mamhoff:
Rebased this PR on the current `master` branch. I've also added deprecation warnings for the scopes `Spree::UserAddress.archived` and `Spree::UserAddress.all_historical`.

mamhoff:
@kennyadsl I changed the wording of the commit message. 

I've also kept the scopes to return all records instead of their custom behavior. If stores make use of that `archived` Boolean (I really doubt it), they are free to empty the migration on upgrading and overriding the new to the old behavior in their stores. I have a really, really hard time imagining this was used for anything anywhere.

mamhoff:
From @kennyadsl : Check if default billing / default shipping addresses are preserved with this change

kennyadsl (approved):

api/spec/requests/spree/api/address_books_spec.rb
```diff
@@ -59,8 +59,7 @@ module Spree::Api
             put "/api/users/#{user.id}/address_book",
               params:  { address_book: harry_address_attributes.merge('id' => address.id) },
               headers: { Authorization: 'Bearer galleon' }
-          }.to change { Spree::UserAddress.count }.from(1).to(2)
-
+          }.not_to change { [Spree::UserAddress.count, user.ship_address.id] }
```
kennyadsl, line 62: Shouldn't this be 

```
}.not_to change { [Spree::UserAddress.count, user.reload.ship_address.id] }
```

otherwise doesn't get the fresh value?

mamhoff:
Yes, i think that is what happens :) I'll provide a separate PR

mamhoff:
@kennyadsl Extracted the behavior change to #4332, care to re-review? :)

core/db/migrate/20201201170826_remove_archived_user_addresses.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class RemoveArchivedUserAddresses < ActiveRecord::Migration[5.2]
+  def up
+    Spree::UserAddress.where(archived: true).delete_all
```
kennyadsl, line 5: I'm still a little bit scared about this. We are assuming people are only using this field in the way we think but it's not guaranteed. 

Maybe we can keep the `archived` column (ignoring it on the Rails side) and use the `bin/rails g solidus:update` task to be sure people are understanding that those records will be deleted from their database. What do you think?

kennyadsl: @waiting-for-dev correct me if I'm wrong here, but if I recall correctly it's a matter of adding a new method in the [update generator](https://github.com/solidusio/solidus/blob/master/core/lib/generators/solidus/update/update_generator.rb), isn't it?

mamhoff: I really can't imagine that - and if we leave the field, we'll have to deal with it in every upgrade from now on (similar to the `Spree::Shipment#deprecated_address_id` field. I'd rather add a large changelog entry, and be done with it. People who use the `archived` flag for something will still be able to edit the migration on their project to not delete the records / column.

core/db/migrate/20201201170826_remove_archived_user_addresses.rb
```diff
@@ -0,0 +1,12 @@
...
+    remove_column :spree_user_addresses, :archived, :boolean, default: false
+  end
+
+  def down
+    add_column :spree_user_addresses
```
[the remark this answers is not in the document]
mamhoff: Oops. Yes.

core/db/migrate/20201201170826_remove_archived_user_addresses.rb
kennyadsl: Using [ignored_columns](https://apidock.com/rails/v6.1.3.1/ActiveRecord/ModelSchema/ClassMethods/ignored_columns) I think it will be pretty transparent for us. 

In my opinion, it's not a matter of imagining how the feature could be used, but more of a formal process of making this software as compliant with SemVer as possible. And removing a table column in a minor release is definitely a breaking change from all possible points of view.

For `Spree::Shipment#deprecated_address_id` (https://github.com/solidusio/solidus/pull/4333), it is a bit different because we are not using that field in the code from 2016 as you mentioned in the PR, so it's very unlikely that it is being used. BTW, I think that PR should go in a major release as well. I didn't comment because I just missed it. 🙂

waiting-for-dev: > @waiting-for-dev correct me if I'm wrong here, but if I recall correctly it's a matter of adding a new method in the [update generator](https://github.com/solidusio/solidus/blob/master/core/lib/generators/solidus/update/update_generator.rb), isn't it?

Do you mean adding a message? Yes, we could do that by adding a new method. However, the actual code is already [redirecting people to the Changelog](https://github.com/solidusio/solidus/blob/547d4eab818796ae77d0bd743359865ab189d069/core/lib/generators/solidus/update/update_generator.rb#L66). I think it'd be better to add any information there and keep using it as the unique source of truth for upgrades info.

waiting-for-dev: Basically, this way we won't have to remember removing the new message before the next Solidus version :wink:

tvdeyen (approved):
I think we should be less strict on semver with this change, because this change has never been visible to users and or admins. It is completely hidden from any UI. We even do not have an address book UI. The only potential users of this "feature" would have been developers and or data analysts. But I bet this is not used by anyone. This causes a lot of confusion and complex code.

What this does is, that it mimics `discard` where it is not necessary, because addresses are not actually removed, but just removed from the users addresses. This causes lots of confusion.

We recently migrated a store from spree to solidus and were very confused still seeing deleted addresses in the users address books.

I am for removing this even in a minor version. 👍🏻

kennyadsl:
I'm ok removing the feature entirely in a minor version. Just saying that maybe we can postpone the database column removal to the next major by just using `ignored_columns` on the model. Any eventual confusion will be gone because that field is explicitly ignored, and we can keep any unexpected usage of that column safe.

On the other hand, when `archived = true` the record will be removed from the database, so if someone is using that column for some reason, they will lose a lot of those records anyway, unless they change the migration manually, but at that point, they can also change the removal of the field if needed.

I will bring this to the core team meeting later today and will get back with a shared decision. Thanks again!

mamhoff:
> I'm ok removing the feature entirely in a minor version. Just saying that maybe we can postpone the database column removal to the next major by just using `ignored_columns` on the model. Any eventual confusion will be gone because that field is explicitly ignored, and we can keep any unexpected usage of that column safe.

If we use `ignored_columns`, then anyone actually using the column would have to metaprogram themselves out of us having invoked it. Not sure that would improve the (imo extremely unlikely) situation.

kennyadsl:
With `metaprogram themselves out of us having invoked it` do you mean decorating the model and removing the field from the `ignored_columns` list, or there's anything more complex to do that I am missing? I thought it was just a matter of doing that but maybe  I'm not considering some other things. 

By the way, even if it's just that, it's not less complex than manually changing the migration so that nothing happens.

mamhoff:
Should I remove the deprecated scopes here entirely then, or deprecate for Solidus 5?

kennyadsl:
If there's a way of deprecating this without removing the database column (eg. as with `ignored_columns`), I think we can merge this immediately, and only remove the column with Solidus 4. Makes sense?

core/app/models/spree/user_address.rb
```diff
@@ -6,18 +6,24 @@ class UserAddress < Spree::Base
...
-    scope :all_historical, -> { unscope(where: :archived) }
+    scope :all_historical, -> {
+      Spree::Deprecation.warn("This scope does not do anything and will be removed from Solidus 4.")
+      all
+    }
     scope :default_shipping, -> { where(default: true) }
     scope :default_billing, -> { where(default_billing: true) }
-    scope :active, -> { where(archived: false) }
+    scope :active, -> {
+      Spree::Deprecation.warn("This scope does not do anything and will be removed from Solidus 4.")
```
kennyadsl, line 24:
```suggestion
      Spree::Deprecation.warn("This scope does not do anything and will be removed from Solidus 5.")
```

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@ suffix.
 
 ### Core
 
+- Remove Spree::UserAddress#archived flag [#3852](https://github.com/solidusio/solidus/pull/3852) ([mamhoff](https://github.com/mamhoff))
+  If you happen to use archived addresses for anything, please be sure to edit the provided migrations in your Solidus project.
```
kennyadsl, line 49:
```suggestion
```

kennyadsl (changes requested):

mamhoff:
Rebased fromn `master`, removed the changelog entries (as I assume they're added back in when y'all release).

CHANGELOG.md
```diff
@@ -3417,4 +3417,4 @@ See https://github.com/solidusio/solidus/releases/tag/v1.0.1
 
 ## Solidus 1.0.0 (2015-08-11)
 
-See https://github.com/solidusio/solidus/releases/tag/v1.0.0
\ No newline at end of file
+See https://github.com/solidusio/solidus/releases/tag/v1.0.0
```
waiting-for-dev, line 3420: Non-blocking, but ideally, we shouldn't have any diff with the Changelog.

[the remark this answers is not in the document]
mamhoff: Fixed

tvdeyen (approved):
@mamhoff can you please rebase this PR? 🙏

tvdeyen:
@kennyadsl I strongly believe that this is good as is. Not necessary to add deprecations for imo. This is a pressing issue and it has been pushed back for too long now.

mamhoff:
> @mamhoff can you please rebase this PR? 🙏

Done!

kennyadsl (approved):
I'm ok merging this.

tvdeyen:
🎉 💯

PR #3853, opened 2020-12-03
Remove deprecated sass color-#{state} variables usage

kennyadsl:
**Description**

This is a preparatory PR for #3816 and needs backport.

With 9552cccd3e825f9bf3fe9c1f1df0b4a4da72c23e we deprecated the following variables in favor of their boostrap counterpart:

|  Deprecated | New  |   
|---|---|
| `$color-success`  | `theme-color(success)`  |
| `$color-notice` | `theme-color(warning)`  | 
| `$color-error`  | `theme-color(danger)`  | 

This PR is quite complex because we were using these deprecated variables in `variables.scss` itself and we can't just swap them to the new version since when this file is imported we didn't import the Boostrap file yet. And there's a reason for that: we define Boostrap colors using the variables defined in `variables.scss`. 🤯 

I propose to move the actions related variable in the already existing `actions.scss` file, so that we can define them later, after Boostrap files have been loaded. 

We preserved the ability to customize those variables per store because they will be still loaded after the `variables_override.scss` file.

There's some slight visual change in the colors, due to using the boostrap built-in `theme-color-level()` function, instead of sass `lighten()`:

# Flash Messages

Screenshot 2020-12-03 at 18 15 06 |

# Table rows with highlighted colors

|  Success  | Warning  | Danger |
|---|---|---|
| Screenshot 2020-12-03 at 16 47 33 |  Screenshot 2020-12-03 at 16 47 49 |   Screenshot 2020-12-03 at 16 47 59 |
**Checklist:**


tvdeyen (approved):
Very great work!

jarednorman (approved):

PR #3854, opened 2020-12-04
Add soft-delete support for Api::ResourceController

cedum:
With Paranoia being replaced by Discard (#3488) this adds soft-delete support also to `Api::ResourceController` the same way it is [implemented in `Admin::ResourceController`](https://github.com/solidusio/solidus/blob/47dacbad7b7d056058dc78a3cc8da45334da0a2e/backend/app/controllers/spree/admin/resource_controller.rb#L94-L101). This way, when deleting a record having Discard included, it will soft-delete the record instead of hard-deleting it.

**Checklist:**


jarednorman (approved):
Nice, change. This makes sense to me.

kennyadsl (approved):
👍

PR #3858, opened 2020-12-09
Remove ability to void invalid payments

vl3:
**Description**

My reasoning behind this is that attempts to void invalid payments will fail because _response_code_ is not set for invalid payments. There are several things that suggests that transitioning from invalid to void should be forbidden (although that the payment could be reattempted, but that's another story).

- Invalid payments will not have **response_code**
- **response_code** is sent as argument to _#void_ when trying to void, potentially causing other gems to fail (in my case, it's the stripe gateway in _active_merchant_ . 
- https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/state_machines/payment.rb#L41 suggest that transitions to void should only happen from [:pending, :processing, :completed, :checkout]

So everything suggests that invalid state should be final and, other than reattempting the payment , we should block this transition. Attached is an screenshot of how the admin panel will understand  transitions.

**Before**

invalid_payment_before

**After**
invalid_payment_after

**Checklist:**


core/app/models/spree/payment_source.rb
```diff
@@ -23,9 +23,9 @@ def can_capture?(payment)
     end
 
     # @param payment [Spree::Payment] the payment we want to know if can be voided
-    # @return [Boolean] true when the payment is not failed or voided
+    # @return [Boolean] true when the payment is not failed, voided or invalid
     def can_void?(payment)
-      !payment.failed? && !payment.void?
+      !payment.failed? && !payment.void? && !payment.invalid?
```
kennyadsl, line 28: Have you considered using `payment.can_void?`. It should just use a [metaprogrammed `state_machine` helper](https://github.com/state-machines/state_machines/blob/5cdb6783edd4e14323276bdb14fb5bcac077d469/lib/state_machines/event.rb#L209) that should return the same value since the only allowed states would be the remaining ones as you linked in the PR description.

kennyadsl (commented):
I think it makes sense, thanks! I left a question about the implementation that I think we can DRY a bit, let me know!

core/app/models/spree/payment_source.rb
vl3: Yeah, it should have the same behaviour. I'll update the PR and remove both tests, given that the logic will not be in payment source anymore and this behaviour is tested in other place.
Thanks for the review!

jarednorman (approved):
Makes sense to me!

kennyadsl (approved):
👍

PR #3865, opened 2020-12-13
Update canonical-rails syntax for latest version

brchristian:
**Description**
canonical-rails version 0.2.10 changed the method `whitelisted_parameters` to `allowed_parameters`. (See https://github.com/jumph4x/canonical-rails/pull/58/files.)

Any user running `bundle update` and going from version <= 0.2.9 to 0.2.10 will fail to compile their asset pipeline, resulting in the error `NoMethodError: undefined method 'whitelisted_parameters=' for CanonicalRails:Module`.

This updates to the current syntax, and also sets the dependency at 0.2.10 to prevent any incompatibility issues that might arise from using the new `allowed_parameters` with old  versions that still expect `whitelisted_parameters`.

**Checklist:**

### DIFF
--- a/frontend/solidus_frontend.gemspec
+++ b/frontend/solidus_frontend.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'solidus_api', s.version
   s.add_dependency 'solidus_core', s.version
 
-  s.add_dependency 'canonical-rails', '~> 0.2.0'
+  s.add_dependency 'canonical-rails', '~> 0.2.10'
   s.add_dependency 'font-awesome-rails', '~> 4.0'
   s.add_dependency 'jquery-rails'
   s.add_dependency 'kaminari', '~> 1.1'
--- a/frontend/config/initializers/canonical_rails.rb
+++ b/frontend/config/initializers/canonical_rails.rb
@@ -12,5 +12,5 @@ CanonicalRails.setup do |config|
   # Parameter spamming can cause index dilution by creating seemingly different URLs with identical or near-identical content.
   # Unless whitelisted, these parameters will be omitted
 
-  config.whitelisted_parameters = [:keywords, :page, :search, :taxon]
+  config.allowed_parameters = [:keywords, :page, :search, :taxon]
 end


kennyadsl:
Thanks, I created https://github.com/jumph4x/canonical-rails/pull/65 to avoid backporting this to old Solidus versions. I think we should wait for their next release if this is accepted. Thanks!

cc @jumph4x

brchristian:
@kennyadsl I like your solution even better!

kennyadsl:
@brchristian they are both needed otherwise we'll have a deprecation warning here! Thanks!

jumph4x:
I see!

jumph4x:
Thank you for the contribution. I'll cut a PATCH version release to make sure existing Gemfiles will continue working. Sorry for breaking it!

jumph4x:
https://rubygems.org/gems/canonical-rails/versions/0.2.11

peterberkenbosch:
nice. My bad for not adding deprecations. Thank @kennyadsl for fixing that!

peterberkenbosch:
This is good to go right? Would love to start new project on Rails 6.1 / Solidus master ;) So let me know if there is anything I can do to move this forward.

frontend/solidus_frontend.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'solidus_api', s.version
   s.add_dependency 'solidus_core', s.version
 
-  s.add_dependency 'canonical-rails', '~> 0.2.0'
+  s.add_dependency 'canonical-rails', '~> 0.2.10'
```
kennyadsl, line 27: we can probably revert this gemspec change now?

brchristian: This PR moves from the deprecated `whitelisted_params` to the new `allowed_params`, which works in 0.2.10 and 0.2.11, so I think that we are going to want something to lock the version at 0.2.10 or 0.2.11 in order to ensure compatibility. The `allowed_params` method is not compatible with 0.2.9 and lower.

kennyadsl: Oh, you are totally right.

kennyadsl (approved):

spaghetticode (approved):
@brchristian thank you, and thanks also to @kennyadsl for his PR on `canonical-rails`!

peterberkenbosch:
Can we merge this? Or waiting for the backport?

gmacdougall (approved):

PR #3869, opened 2020-12-18
Update taxon icon to use present instead of exists

seand7565:
`exists?` on a paperclip object attempts to access the image where
it's stored, whereas `present?` just checks if the object exists on
Solidus.

Using `exists?` here causes an error if the image doesn't exist. Luckily,
paperclip double checks that the image filename is present before
attempting to fetch it from the storage service, but paperclip_cloudinary
does not do this, meaning that any store using paperclip_cloudinary is
unable to view the admin taxon edit page for taxons without icons.

[This was an issue on the demo store](https://github.com/solidusio/solidus-demo/issues/69), which uses paperclip_cloudinary. I've patched it up there, but figured I should add a fix here as well.

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/taxons/attachment_forms/_paperclip.html.erb
+++ b/backend/app/views/spree/admin/taxons/attachment_forms/_paperclip.html.erb
@@ -2,7 +2,7 @@
   <%= f.field_container attachment do %>
     <%= f.label attachment %><br>
     <%= f.file_field attachment %>
-    <% if f.object.send(attachment).exists? %>
+    <% if f.object.send(attachment).present? %>
       <%= image_tag f.object.send(attachment, definition[:default_style]) %>
       <%= link_to t('spree.actions.remove'),
                   admin_taxonomy_taxon_attachment_path(@taxonomy,


kennyadsl (approved):
Thanks Sean!

spaghetticode (approved):
@seand7565 thanks 👍

backend/spec/features/admin/taxons_spec.rb
```diff
@@ -76,6 +76,14 @@ def only_one_selected_tab_inside?(sub_tab_selector, tab_name, tab_path)
     end
   end
 
+  it "should not send exists to attachments" do
+    taxon = create(:taxon)
+
+    expect_any_instance_of(Paperclip::Storage::Filesystem).not_to receive(:exists?)
+
+    visit spree.edit_admin_taxonomy_taxon_path(taxon.taxonomy, taxon.id)
+  end
```
jarednorman, line 85: This test is _extremely_ fragile and totally dependent on the implementation details of a deprecated gem. I like the change itself, but as I see it this test guards against nothing other than someone reverting this exact change which I don't think we need to worry about.

I'm open to hearing alternative perspectives on it, but I don't currently see any value in maintaining this test.

jarednorman (changes requested):

backend/spec/features/admin/taxons_spec.rb
aldesantis: I agree, the test here is not very meaningful. The only alternative I see is testing this in integration by creating a record that has an attachment, but whose the attachment doesn't actually exist. However, I think it would be much more effort than the value we'd get out of it.

seand7565: Makes sense to me! I've ditched the spec, let me know if that works for you.

jarednorman (approved):

PR #3880, opened 2021-01-04
Fix Search::Base#respond_to_missing? arity

elia:
ref https://github.com/solidusio/solidus/commit/9219f1e384aa68f049eb0dd984ac7b780dd3884d

**Description**

The method's signature was wrong and ended up in an error when actually called through `respond_to?`.

https://ruby-doc.org/core-3.0.0/Object.html#method-i-respond_to_missing-3F

**Checklist:**

### DIFF
--- a/core/lib/spree/core/search/base.rb
+++ b/core/lib/spree/core/search/base.rb
@@ -46,8 +46,8 @@ module Spree
           end
         end
 
-        def respond_to_missing?(name)
-          @properties.key?(name) || super(name)
+        def respond_to_missing?(name, include_private)
+          @properties.key?(name) || super
         end
 
         protected


core/lib/spree/core/search/base.rb
```diff
@@ -46,8 +46,8 @@ def method_missing(name)
           end
         end
 
-        def respond_to_missing?(name)
-          @properties.key?(name) || super(name)
+        def respond_to_missing?(name, include_private)
```
jarednorman, line 49: I think that second argument is optional, right?

```suggestion
        def respond_to_missing?(name, include_private = false)
```

jarednorman (changes requested):

core/lib/spree/core/search/base.rb
elia: That's true for `respond_to?` but not for `respond_to_missing?` which should never be used directly and works just as an internal hook:

image

jarednorman (approved):

aldesantis (approved):

spaghetticode:
@elia this is good to merge but needs rebasing, can you do it?

elia:
Fixed by da36bc13

PR #3883, opened 2021-01-07
Deprecate UserLastUrlStorer and replace `redirect_back_or_default` with Rails 5's `redirect_back`

cincospenguinos:
**Description**
Given that Rails 5 introduces `redirect_back` and Solidus currently uses `redirect_back_or_default` for the same behavior, this PR replaces usages of `redirect_back_or_default` with `redirect_back`, and deprecates the public API class that provides `redirect_back_or_default`. This was done for Issue #3883.

**Checklist:**

### DIFF
--- a/core/app/models/spree/user_last_url_storer.rb
+++ b/core/app/models/spree/user_last_url_storer.rb
@@ -4,8 +4,6 @@ module Spree
   # This service object is responsible for storing the current path into
   # into `session[:spree_user_return_to]` for redirects after successful
   # user/admin authentication.
-  #
-  # @deprecated Please use #redirect_back provided by Rails 5+
   class UserLastUrlStorer
     # Lists all the rules that will be evaluated before storing the
     # current path value into the session.
--- a/core/app/models/spree/user_last_url_storer.rb
+++ b/core/app/models/spree/user_last_url_storer.rb
@@ -4,6 +4,8 @@ module Spree
   # This service object is responsible for storing the current path into
   # into `session[:spree_user_return_to]` for redirects after successful
   # user/admin authentication.
+  #
+  # @deprecated Please use #redirect_back provided by Rails 5+
   class UserLastUrlStorer
     # Lists all the rules that will be evaluated before storing the
     # current path value into the session.
@@ -26,6 +28,15 @@ module Spree
     #  or its subclasses. The controller will be passed to each rule for matching.
     def initialize(controller)
       @controller = controller
+    end
+
+    # Stores into session[:spree_user_return_to] the request full path for
+    # future redirects (to be used after successful authentication). When
+    # there is a rule match then the request full path is not stored.
+    def store_location
+      return if self.class.rules.any? { |rule| rule.match? controller }
+
+      session[:spree_user_return_to] = request.fullpath.gsub('//', '/')
     end
 
     private
--- a/core/app/models/spree/user_last_url_storer.rb
+++ b/core/app/models/spree/user_last_url_storer.rb
@@ -28,15 +28,6 @@ module Spree
       @controller = controller
     end
 
-    # Stores into session[:spree_user_return_to] the request full path for
-    # future redirects (to be used after successful authentication). When
-    # there is a rule match then the request full path is not stored.
-    def store_location
-      return if self.class.rules.any? { |rule| rule.match? controller }
-
-      session[:spree_user_return_to] = request.fullpath.gsub('//', '/')
-    end
-
     private
 
     attr_reader :controller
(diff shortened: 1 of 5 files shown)


jarednorman (commented):
I'm generally in favour of removing this now unnecessary complexity and using the built in Rails functionality.

core/app/models/spree/user_last_url_storer.rb
```diff
@@ -28,15 +28,6 @@ def initialize(controller)
       @controller = controller
     end
 
-    # Stores into session[:spree_user_return_to] the request full path for
-    # future redirects (to be used after successful authentication). When
-    # there is a rule match then the request full path is not stored.
-    def store_location
```
kennyadsl, line 34: Unfortunately, we can't just remove public methods from classes since people may be still using them. A better approach could be deprecating the method in order to alert people that we are going to remove this method in the next major version. I think that `store_location` is the only method used so we can probably deprecate the whole class usage.

core/lib/spree/core/controller_helpers/auth.rb
```diff
@@ -48,11 +48,6 @@ def current_ability
           @current_ability ||= Spree::Ability.new(try_spree_current_user)
         end
 
-        def redirect_back_or_default(default)
```
kennyadsl, line 51: Same here, we can't remove this method because we are not sure how/if people are using it. Better to deprecate it.

core/lib/spree/core/controller_helpers/auth.rb
```diff
@@ -57,10 +57,6 @@ def set_guest_token
           end
         end
 
-        def store_location
```
kennyadsl, line 65: Did you get a chance to see where this method was called from?

kennyadsl (commented):
Thanks, I think we are on the right track! I know it may still be WIP but I left a preliminary review, hope it helps in this phase.

core/lib/spree/core/controller_helpers/auth.rb
cincospenguinos: As far as I could tell, it was only used in spec files which made it ripe for removal. Is this another public method in use?

kennyadsl: Oh, it is used by our authentication plugin (using devise) [here](https://github.com/solidusio/solidus_auth_devise/search?q=store_location). I think this should be taken into account to understand if we can remove the method and we are using the right approach.

cincospenguinos: Agreed. Personally I would opt to just replace our usage of `redirect_back_or_default` with Rails's implementation, but indicate that the class that maintains that behavior has been deprecated. I'll ensure deprecation warnings are in place and put the specs back.

core/lib/spree/core/controller_helpers/auth.rb
```diff
@@ -48,6 +48,12 @@ def current_ability
           @current_ability ||= Spree::Ability.new(try_spree_current_user)
         end
 
+        # @deprecated Use #redirect_back provided in Rails 5+
+        def redirect_back_or_default(default)
+          redirect_to(session["spree_user_return_to"] || default)
+          session["spree_user_return_to"] = nil
+        end
```
jarednorman: We use Rails' deprecation system, so with this and the other deprecated method you can do something like:

```ruby
deprecate redirect_back_or_default: "Please use redirect_back provided in Rails 5+ instead", deprecator: Spree::Deprecation
```

core/lib/spree/core/controller_helpers/auth.rb
kennyadsl: I'm just concerned that we are deprecating a core function without providing an alternative for extensions using this method. This could even be just instructions on how to replace `store_location` with something else.

kennyadsl: Hey Andre, I just found [this page on the Devis Wiki](https://github.com/heartcombo/devise/wiki/How-To:-Redirect-back-to-current-page-after-sign-in,-sign-out,-sign-up,-update), which seems very relevant. I suspect we needed the custom code when the Devise gem did not have those methods available. Apparently we just can't rely on the header referral here, so some sort of storage mechanism for the last visited URL (except the authentication ones to avoid redirect loops) is somehow required.

jarednorman (changes requested):
Looking great. I think the only thing I'd like to see here is for the commits to be squashed down to remove the extra noise from homing on the solution.

waiting-for-dev:
We are closing as we need to add Rails 7' `redirect_back_or_to` to the equation. This PR can be used as an excellent starting point. @cincospenguinos, thanks for your collaboration, and feel free to work on it if you're still interested :heart:

PR #3885, opened 2021-01-08
Adjust CircleCI config to reflect Rails versions that we support

kennyadsl:
**Description**

We recently added support for Rails 6.1 but we still need to run specs against Rails 6.0. This PR takes care of that.

To reduce the amount of CI time needed, I decided to remove MySQL tests for older Rails versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,15 +128,6 @@ jobs:
       - setup
       - test
 
-  mysql_rails52:
-    executor: mysql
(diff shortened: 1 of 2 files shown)


kennyadsl:
The pending CI check should not be expected actually. I removed mysql_rails52 in this PR and I think it's CircleCI not being able to get it.

jarednorman (approved):

spaghetticode (approved):

PR #3886, opened 2021-01-08
Enable ActiveStorage specs against Rails 6.1

kennyadsl:
**Description**

This PR is only useful to highlight the specs failures we have enabling ActiveStorage on Rails 6.1 and should be merged only after we fixed the existing issues.

Some ActiveStorage internals changed since we worked on this (see #2974 #3501) and we have to make some changes to reflect the new API.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,6 +98,8 @@ jobs:
   postgres:
     executor: postgres
     parallelism: &parallelism 3
+    environment:
+      ENABLE_ACTIVE_STORAGE: true
     steps:
       - setup
       - test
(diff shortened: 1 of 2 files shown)


kennyadsl:
@elia can you please elaborate a little bit about why we are using the paperclip partial for ActiveStorage as well?

kennyadsl:
@elia I did check the code and I think I've got it. We are now using the `paperclip` partial for both Paperclip and Activestorage. I think we can remove this abstraction in a separate PR as you probably thought.

spaghetticode (approved):
@elia thanks!

jarednorman (approved):

Issue #3873, opened 2020-12-22
Switch allow_forgery_protection to false in test


kennyadsl:
With [this commit](https://github.com/solidusio/solidus/commit/3b8311230cd8ec208a89c526b03cd04bea1818f1) I switched the default value of `allow_forgery_protection` to true in our spec suite to mimic what we see in a new Rails application after https://github.com/rails/rails/commit/ec4a836. 

But this doesn't mean that Rails wants us to test the application with that value set to `true`, in fact, the default is `false` in the [test environment template](https://github.com/rails/rails/blob/56c9be11ad1cd75f9d0114947cbad36ad2b5901e/railties/lib/rails/generators/rails/app/templates/config/environments/test.rb.tt#L39).

Other than not adhere to a Rails convention, it's important that we change this in order to be able to use system specs in Solidus frontend, which we are trying to convert in the new [starter frontend project](https://github.com/nebulab/solidus_starter_frontend/pull/90).

FrancescoAiello01:
Hey @kennyadsl taking a look at this issue.

I'm wondering if `dummy_app.rb` should always turn off forgery protection, just as real Rails apps turn it off in their test env? Or should it still be true in certain circumstances?

Thanks,
Francesco

kennyadsl:
Hey @FrancescoAiello01, I can't see any circumstance when we need to keep it enabled at the moment. I think we are good if we just disable it.

FrancescoAiello01:
Cool @kennyadsl sounds good. Just opened a PR over [here](https://github.com/solidusio/solidus/pull/3887). Let me know what you think.

PR #3887, opened 2021-01-08
Set dummy app forgery protection to false

FrancescoAiello01:
**Description**
This reverts a previous change setting dummy app forgery protection to true. Rails prefers testing to be performed with forgery protection set to false.

Verification effort: test suite passes

Ref #3873

**Checklist:**

### DIFF
--- a/core/lib/generators/spree/dummy/templates/rails/test.rb
+++ b/core/lib/generators/spree/dummy/templates/rails/test.rb
@@ -20,6 +20,9 @@ Dummy::Application.configure do
   # Raise exceptions instead of rendering exception templates
   config.action_dispatch.show_exceptions = false
 
+  # Disable request forgery protection in test environment
+  config.action_controller.allow_forgery_protection = false
+
   # Tell Action Mailer not to deliver emails to the real world.
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -17,6 +17,7 @@ RAILS_6_OR_ABOVE = Rails.gem_version >= Gem::Version.new('6.0')
 
 # @private
 class ApplicationController < ActionController::Base
+  protect_from_forgery with: :exception
 end
 
 # @private
@@ -52,8 +53,8 @@ module DummyApp
     config.public_file_server.headers = { 'Cache-Control' => 'public, max-age=3600' }
     config.whiny_nils = true
     config.consider_all_requests_local = true
-    config.action_controller.allow_forgery_protection = true
-    config.action_controller.default_protect_from_forgery = true
+    config.action_controller.allow_forgery_protection = false
+    config.action_controller.default_protect_from_forgery = false
     config.action_controller.perform_caching = false
     config.action_dispatch.show_exceptions = false
     config.active_support.deprecation = :stderr


core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -17,6 +17,7 @@
 
 # @private
 class ApplicationController < ActionController::Base
+  protect_from_forgery with: :exception
```
kennyadsl, line 20: Do you think this is necessary?

core/lib/generators/spree/dummy/templates/rails/test.rb
```diff
@@ -20,6 +20,9 @@
   # Raise exceptions instead of rendering exception templates
   config.action_dispatch.show_exceptions = false
 
+  # Disable request forgery protection in test environment
+  config.action_controller.allow_forgery_protection = false
```
kennyadsl, line 24: I feel like this is not necessary since we already set this value when defining the dummy app.

kennyadsl (commented):
Thanks @FrancescoAiello01! I have the feeling we can skip some lines of code. I know they were there before but maybe they were useless already. Let me know!

core/lib/generators/spree/dummy/templates/rails/test.rb
FrancescoAiello01: That's a good point. It is redundant, I just pushed a change removing the line. Thanks

core/lib/spree/testing_support/dummy_app.rb
FrancescoAiello01: I tried to not include that line, but I ran into an issue in `frontend/app/controllers/spree/orders_controller.rb` on line 14: `skip_before_action :verify_authenticity_token, only: [:populate]`

The error was: 
`ArgumentError: Before process_action callback :verify_authenticity_token has not been defined`

I think the problem was that `:verify_authenticity_token` doesn't exist, so it's raising an exception when trying to skip it. My first thought was just to add `raise: false` like this:

`skip_before_action :verify_authenticity_token, only: [:populate], raise: false`

Which works, but seems really hacky and not ideal. Instead, I thought if I added the line:

`protect_from_forgery with: :exception`

It would add `verify_authenticity_token` to the `before_filter` list, which can then be skipped in other controllers without an error.

It's possible I'm misunderstanding, so please let me know what you think.

Thanks so much

kennyadsl: Does this error happen in the test suite? If yes can you please point a spec that is failing for reference?

FrancescoAiello01: This is the error output:
```
An error occurred while loading ./spec/controllers/spree/orders_controller_ability_spec.rb.
Failure/Error: skip_before_action :verify_authenticity_token, only: [:populate]

ArgumentError:
  Before process_action callback :verify_authenticity_token has not been defined
# ./app/controllers/spree/orders_controller.rb:14:in `<class:OrdersController>'
# ./app/controllers/spree/orders_controller.rb:4:in `<module:Spree>'
# ./app/controllers/spree/orders_controller.rb:3:in `<top (required)>'
# ./spec/controllers/spree/orders_controller_ability_spec.rb:6:in `<module:Spree>'
# ./spec/controllers/spree/orders_controller_ability_spec.rb:5:in `<top (required)>'


Finished in 0.00005 seconds (files took 4.88 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples

```
It looks like the error is happening when loading the order controller before the tests.

kennyadsl: Oh ok, it makes sense. I think that we need that `protect_from_forgery with: :exception` then. 
If we add `, raise: false` it will also change the behavior outside the test environment so I don't think it's a viable solution.

FrancescoAiello01: Yeah good point about keeping it in the test environment. Thanks for taking a look

core/lib/generators/spree/dummy/templates/rails/test.rb
SamuelMartini: Hey!
I have tried to run the [solidus_starter_frontend](https://github.com/nebulab/solidus_starter_frontend/pull/90) from this branch and they are failing because of the forgery protection.

I think we need to disable the `allow_forgery_protection` configuration here as well.

The reason is that there are two dummy apps that can be generated.

 1 - A `Dummy::Application` to test extensions that is created by running [this](https://github.com/solidusio/solidus/blob/v2.11.4/core/lib/spree/testing_support/extension_rake.rb) task and use this [DummyGenerator](https://github.com/solidusio/solidus/blob/v2.11.4/core/lib/generators/spree/dummy/dummy_generator.rb#L61). This generator creates a full rails application (Dummy::Application) and overrides the rails `environments/test.rb` file with the above templates/rails/test.rb configuration. This is the reason `allow_forgery_protection` needs to be disabled in `core/lib/generators/spree/dummy/templates/rails/test.rb `.
 
2 - A `DummyApp::Application` to test the solidus framework and it is an application initialized in-memory [here](https://github.com/solidusio/solidus/blob/v2.11.4/core/lib/spree/testing_support/dummy_app.rb#L35) when the [spec/rails_helper.rb](https://github.com/solidusio/solidus/blob/master/core/spec/rails_helper.rb#L8) file is loaded. This is the reason we need to disable in `core/lib/spree/testing_support/dummy_app.rb`

kennyadsl: You are right @SamuelMartini, we need to re-add it to make it work with Dummy apps generated in extensions. 

@FrancescoAiello01 sorry for the wrong suggestion here! 😬

FrancescoAiello01: That makes sense, sorry for the confusion! Going to push a fix now.

kennyadsl:
@FrancescoAiello01 please rebase against master to have the spec suite passing!

kennyadsl (approved):
Thanks!

jarednorman (approved):

PR #3888, opened 2021-01-10
Enhances Layout Config to render different layouts if needed

### LINKED ISSUE #3866: Let Config.layout accept all the arguments allowed by Rails

I need to use a simplified layout for the checkout and for other places of the store. 
It's very common to have less UI during the checkout to avoid distractions.

Currently the `Common` controller helper is the one setting the layout and the config entry gives no extension point to dynamically select the layout based on the current controller/action/request etc. 
As we all know "decorating" a helper is a pain in the neck, and requires the module to be re-included in all the classes its included in. It wouldn't be much of a problem if we didn't have 6 controllers including that module, if we take auth-devise into account. 

I'd like to see `Spree::Config.layout` accept a proc that takes the current controller instance as an argument and/or is evaluated with `instance_eval` inside it as it's done by Rails. Even better would be to just accept anything allowed by [`ActionView::Layouts::ClassMethods#layout`](https://api.rubyonrails.org/v6.0.0/classes/ActionView/Layouts/ClassMethods.html#method-i-layout).

I would also suggest to switch the default layout name from `spree_application` to `application` as most of the store I've seen only used the solidus layout and `spree_application` is very confusing for newcomers to solidus.

- conditionals inside the layout: they would work but that's yet another a workaround fighting solidus instead of making things right
- decorating the `Common` controller helper: it's a nightmare
- decorating all the individual controller involved in the checkout: equally overkill and not very maintainable

devchris:
**Description**
Closes #3866 

Added the ability for Spree::Config.layout accept anything allowed by ActionView::Layouts::ClassMethods#layout.

**Checklist:**


devchris:
@elia This is a slightly different approach which I believe is a little bit more straight forward to implement. Since the whole reason behind this is that you can use a simplified layout during checkout and some other places, you could now create a new layout, ex. `checkout_layout` and declare it inside the initializer like 
```
  config.layout_conditions = {
    'spree/checkout': 'checkout_layout'
  }
```

Let me know what you think about this approach, open for any feedback 👍

aldesantis:
Hey @devchris, thanks for the work on this! I have a couple of comments:

First of all, could you split the layout renaming into a separate commit or, even better, a separate PR? It seems like it's completely unrelated to the other changes?

Secondly, while your `layout_conditions` preference fixes the problem outlined by @elia in #3866, I'm not sure it's an ideal solution: with your change, it's only possible to differentiate layouts by controller. What if I want to use some other condition, such as an attribute on the user's profile, or the state of a feature flag?

Another aspect to keep in mind is that each additional configuration option increases the user's cognitive load and the system's complexity.

Accepting the same range of inputs as Rails would get rid of the additional option, provide maximum flexibility to the user and get us closer to the Rails conventions, which means people can just do things the Rails way without having to learn the Solidus way.

Let me know your thoughts!

devchris:
> First of all, could you split the layout renaming into a separate commit or, even better, a separate PR? It seems like it's completely unrelated to the other changes?

Will do

> Secondly, while your layout_conditions preference fixes the problem outlined by @elia in #3866, I'm not sure it's an ideal solution: with your change, it's only possible to differentiate layouts by controller. What if I want to use some other condition, such as an attribute on the user's profile, or the state of a feature flag?

That makes sense. Haven't thought about those use cases - I'll get it changed

core/lib/spree/core/controller_helpers/common.rb
```diff
@@ -73,8 +73,22 @@ def set_user_language
...
+            layout
+          when Symbol
+            send(layout)
+          when Proc
+            Spree::Config.layout.call(self)
+          when true
+            raise ArgumentError, 'Layouts must be specified as a String, Symbol, Proc, false, or nil'
+          when nil
+            'spree/layouts/spree_application'
+          end
```
devchris, line 91: My initial thought was to get rid of the `get_layout` method and just call `Spree::Config.layout` directly on line 15. This kinda worked when I was testing it manually. But I did not get the tests to pass like this method. No matter what I would set for the layout in the config file, it wouldn't take effect, maybe I was missing something?

devchris: Anyways, this works without having to deprecate the layout preference or breaking any current applications

aldesantis: @devchris I think the reason that it's not working in tests is that the controller has already been loaded by the time you attempt to stub the preference, and the `layout` call has already been made. This is usually resolved either by reloading the class (i.e., unloading the constant associated to the file and `load`ing it again) or by moving the logic into a method (which is what you've done).

devchris: @aldesantis This seems to be a little bit more tricky than just simply "reloading" it in this case. But regardless, keeping it in a method gives us some flexibility when we decide to go the route to deprecate 'spree_application.html.erb' 😄

aldesantis: Makes sense!

aldesantis:
@elia what do you think about this?

elia (changes requested):
@devchris looks great 💪!

I would ask you just a couple of minor things:
1. can you please update the commit(s) to make them more atomic and specific to the feature?
2. can you add specs for all the possible values, including using a method, and `false` (for which I'm not even sure what the behavior would be

Thanks!

PR #3894, opened 2021-01-12
Renames Spree Application to Application

devchris:
**Description**
#3866 Mentioned a renaming of the `spree_application` to `application` to avoid confusion.

This could potentially be a "breaking change" for some existing applications. Could be a part of v3.

**Checklist:**


aldesantis:
@devchris thanks! FYI, our plan is not to introduce any changes in 3.0: it should only remove deprecated code. The way we achieve this kind of changes is by always having a deprecation period before rolling them out.

In this case, what do you think would be a good approach?

devchris:
@aldesantis Thanks for the info, I didn't know that this was the plan yet. This would only be a breaking change if someone specifically said that the layout should be `spree_application.html.erb`. While this is also the default value, we can't guarantee  that no user didn't set the default value again.
To resolve this, we could add a deprecation warning that the layout default "preference" will change from `spree_application.html.erb` to `application.html.erb` in an upcoming version which doesn't have any visual impact but will be removed in the future. We probably don't need to copy paste the same exact file but just check if someone is looking for `spree_application.html.erb` or `application.html.erb` and render `application.html.erb`

It would probably also be a good idea do wait for those changes in this PR here until #3888 gets merged.

elia:
This seems a bit trickier than anticipated… and maybe a good approach would be: 
- update the generated initializer so that will point to to the default layout (for newly installed apps)
- leave the pref default as it is (to support legacy apps)
- use a generator to add just the required pieces to the default rails layout (e.g. `spree/shared/head`) and leave everything else as vanilla as possible (js/webpack includes  etc.)

I know this widens the scope of this PR, but I think this will bring some fresh air to new stores, but I'm not _core_ so I'll defer to @aldesantis and the rest of the core-team to make the final call on this 👍

aldesantis:
@elia that sounds like a good approach! 🙂

@devchris what are your thoughts on the above?

PR #3896, opened 2021-01-17
Support multiple errors in Promotion Handlers

### LINKED ISSUE #3227: Support multiple errors in Promotion Handlers
At the moment, applying a promotion coupon (but I think it is an issue present also in the other promotion handlers) consider a single error possible.

Setting the error happens through [setting an `error` attribute on the class, along with a `status_code`](https://github.com/solidusio/solidus/blob/01f71957302915a639d25e76a3a2376b243a854e/core/app/models/spree/promotion_handler/coupon.rb#L7).

The problem is where we have multiple errors for the same promotion, for example, if the coupon code is added to an order that does not meet multiple promotion rules for that promotion, as discussed [here](https://github.com/solidusio/solidus/pull/3193#issuecomment-494035406).

**A possible solution**

I think we should change `error` to `errors` adding a new class that represent that `PromotionHandlerError` concept, or something similar. We should also try to deprecate the current usage of `error` and `status_code` printing a deprecation warning and falling back to the first `PromotionHandlerError` associated with that PromotionHandler.

cc @fkoessler

svyatogor:
**Description:**
Fixes #3227
This PR adds support for multiple errors in Promotion Handlers by merging error messages of the underlying rules and introducing support for multiple errors on the Coupon. For backward compatibility `#error` method is maintained and falls back to the first error. In case there is at least one error `#status_code` will also return the code of the first error.

Note, that the merge behaviour is only implemented when promotion is set to match all rules, as with match any it is not very obvious what should be the visual presentation and if all possible rules should be exposed/visible to the customer.

Screenshot 2021-01-18 at 11 56 33 PM

**Checklist:**

### DIFF
--- a/api/app/views/spree/api/promotions/handler.json.jbuilder
+++ b/api/app/views/spree/api/promotions/handler.json.jbuilder
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 json.success(@handler.success)
+json.error(@handler.error)
 json.errors(@handler.errors.details[:base])
 json.successful(@handler.successful?)
 json.status_code(@handler.status_code)
--- a/api/app/views/spree/api/promotions/handler.json.jbuilder
+++ b/api/app/views/spree/api/promotions/handler.json.jbuilder
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
 json.success(@handler.success)
-json.error(@handler.error)
+json.errors(@handler.errors.details[:base])
 json.successful(@handler.successful?)
-json.status_code(@handler.status_code)
--- a/api/app/views/spree/api/promotions/handler.json.jbuilder
+++ b/api/app/views/spree/api/promotions/handler.json.jbuilder
@@ -3,3 +3,4 @@
 json.success(@handler.success)
 json.errors(@handler.errors.details[:base])
 json.successful(@handler.successful?)
+json.status_code(@handler.status_code)
--- a/api/openapi/solidus-api.oas.yml
+++ b/api/openapi/solidus-api.oas.yml
@@ -6174,18 +6174,26 @@ components:
           type: string
         order_number:
           type: string
+    promotion-handler-error:
+      type: object
+      title: Promotion handler error
+      properties:
+        error:
+          type: string
+        error_code:
+          type: string
     coupon-code-handler:
       type: object
       title: Coupon code handler
       properties:
         success:
           type: string
-        error:
-          type: string
+        errors:
+          type: array
+          items:
+            $ref: '#/components/schemas/promotion-handler-error'
         successful:
           type: boolean
-        status_code:
-          type: string
     address-book:
       type: array
       title: Address book
--- a/api/openapi/solidus-api.oas.yml
+++ b/api/openapi/solidus-api.oas.yml
@@ -6188,12 +6188,18 @@ components:
       properties:
         success:
           type: string
+        error:
+          type: string
+          deprecated: true
         errors:
           type: array
           items:
             $ref: '#/components/schemas/promotion-handler-error'
         successful:
           type: boolean
+        status_code:
+          type: string
+          deprecated: true
     address-book:
       type: array
       title: Address book
--- a/core/app/models/spree/promotion_handler/coupon.rb
+++ b/core/app/models/spree/promotion_handler/coupon.rb
@@ -3,12 +3,15 @@
 module Spree
   module PromotionHandler
     class Coupon
-      attr_reader :order, :coupon_code
-      attr_accessor :error, :success, :status_code
+      extend ActiveModel::Naming
+
+      attr_reader :order, :coupon_code, :errors
+      attr_accessor :success
 
       def initialize(order)
         @order = order
         @coupon_code = order.coupon_code && order.coupon_code.downcase
+        @errors = ActiveModel::Errors.new(self)
       end
 
       def apply
@@ -40,13 +43,13 @@ module Spree
       end
 
       def set_success_code(status_code)
-        @status_code = status_code
         @success = I18n.t(status_code, scope: 'spree')
+        @success_status_code = status_code
       end
 
       def set_error_code(status_code, options = {})
-        @status_code = status_code
-        @error = options[:error] || I18n.t(status_code, scope: 'spree')
+        @errors.clear
+        @errors.add(:base, options[:error] || I18n.t(status_code, scope: 'spree'), error_code: status_code)
       end
 
       def promotion
@@ -58,7 +61,16 @@ module Spree
       end
 
       def successful?
-        success.present? && error.blank?
+        success.present? && errors.empty?
+      end
+
+      def error
+        Spree::Deprecation.warn "#error is deprecated. Please start using #errors."
+        errors.full_messages.first
+      end
+
+      def status_code
+        errors.any? ? errors.details[:base].first[:error_code] : @success_status_code
       end
 
       private
@@ -72,8 +84,8 @@ module Spree
         return promotion_applied if promotion_exists_on_order?(order, promotion)
 
         unless promotion.eligible?(order, promotion_code: promotion_code)
-          set_promotion_eligibility_error_code(promotion)
-          return (error || ineligible_for_this_order)
+          @errors = promotion.eligibility_errors
+          return (errors.any? || ineligible_for_this_order)
         end
 
         # If any of the actions for the promotion return `true`,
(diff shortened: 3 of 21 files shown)


api/app/views/spree/api/promotions/handler.json.jbuilder
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
 json.success(@handler.success)
-json.error(@handler.error)
```
kennyadsl, line 4: I think we should keep both `error` and `errors` here otherwise people consuming the API and relying on `error` in the JSON response will have their consumer broken.

api/openapi/solidus-api.oas.yml
```diff
@@ -6174,18 +6174,26 @@ components:
...
+          type: string
+        error_code:
+          type: string
     coupon-code-handler:
       type: object
       title: Coupon code handler
       properties:
         success:
           type: string
-        error:
```
kennyadsl, line 6183: We should keep this and mark it as deprecated, it will be removed in the next major bump.

api/openapi/solidus-api.oas.yml
```diff
@@ -6174,18 +6174,26 @@ components:
...
-          type: string
+        errors:
+          type: array
+          items:
+            $ref: '#/components/schemas/promotion-handler-error'
         successful:
           type: boolean
-        status_code:
```
kennyadsl, line 6187: Same here

core/app/models/spree/promotion_handler/coupon.rb
```diff
@@ -40,13 +43,13 @@ def remove
...
       def set_success_code(status_code)
-        @status_code = status_code
         @success = I18n.t(status_code, scope: 'spree')
+        @success_status_code = status_code
       end
 
       def set_error_code(status_code, options = {})
-        @status_code = status_code
-        @error = options[:error] || I18n.t(status_code, scope: 'spree')
+        @errors.clear
```
kennyadsl, line 51: Can't we use `errors` instead of `@errors` now that we have the `attr_reader`? Also I can't find what `.clear` does in the [`ActiveModel::Errors` documentation](https://www.rubydoc.info/docs/rails/ActiveModel/Errors). Can you help me understand?

core/app/models/spree/promotion_handler/coupon.rb
```diff
@@ -72,8 +84,8 @@ def handle_present_promotion(promotion)
         return promotion_applied if promotion_exists_on_order?(order, promotion)
 
         unless promotion.eligible?(order, promotion_code: promotion_code)
-          set_promotion_eligibility_error_code(promotion)
-          return (error || ineligible_for_this_order)
+          @errors = promotion.eligibility_errors
```
kennyadsl, line 87: If we need to set `@errors` we can probably make it an attr_accessor and use `errors =` ? What do you think?

frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js
```diff
@@ -30,7 +30,11 @@ Spree.onCouponCodeApply = function(e) {
     var handler;
     // handler = JSON.parse(xhr.responseText)
     handler = xhr.responseJSON;
-    couponStatus.addClass(errorClass).html(handler["error"]);
+    const errorMessages = $.map(
+      handler["errors"],
+      function (err) { return err.error }
+    )
+    couponStatus.addClass(errorClass).html(errorMessages.join('<br>'));
```
kennyadsl, line 37: Do you have a screenshot of how it looks? Doesn't need to be good, just wanted to be sure it doesn't break the page layout.

frontend/app/controllers/spree/coupon_codes_controller.rb
```diff
@@ -18,7 +20,7 @@ def create
               flash[:success] = handler.success
               redirect_to cart_path
             else
-              flash.now[:error] = handler.error
+              flash.now[:error] = safe_join(handler.errors.full_messages, " <br /> ".html_safe)
```
kennyadsl, line 23: What about providing a generic coupon failure message and also set a `@coupon_handler` instance variable that can be used in the view layer to retrieve the errors? I'm just concerned that some store flash implementation is not ready to have multi-lines messages.

core/spec/spec_helper.rb
```diff
@@ -6,7 +6,6 @@
 end
 
 require 'rspec/core'
-
```
kennyadsl, line 9: Not sure if this is intentional but I'd avoid this just to have a cleaner git history.

core/app/models/spree/promotion_handler/coupon.rb
```diff
@@ -58,7 +61,16 @@ def promotion
...
+        success.present? && errors.empty?
+      end
+
+      def error
+        Spree::Deprecation.warn "#error is deprecated. Please start using #errors."
+        errors.full_messages.first
+      end
+
+      def status_code
+        errors.any? ? errors.details[:base].first[:error_code] : @success_status_code
```
kennyadsl, line 73: Wondering if we need to keep the `status_code` at the handler level for some specific reason or we can move the `status_code` attribute into each error and deprecate this as well. Thoughts?

kennyadsl (commented):
Great work and thanks @svyatogor! 

I left some questions, let me know your thoughts! 

On a side note, I think the CI is not working because you created the project under your username as well in CircleCI. I think that if you remove it, it will be executed again under the Solidusio organization context, with all the permissions in place. Let me know if this fixes it.

frontend/app/assets/javascripts/spree/frontend/checkout/coupon-code.js
kennyadsl: Saw the screenshot in the PR description. All good here!

core/app/models/spree/promotion_handler/coupon.rb
svyatogor: Does exactly what the name suggests, removes all errors: https://apidock.com/rails/v6.0.0/ActiveModel/Errors/clear

core/app/models/spree/promotion_handler/coupon.rb
svyatogor: This will make it writable from outside the class, I tend to keep setters private even to avoid confusion in the future.

core/app/models/spree/promotion_handler/coupon.rb
svyatogor: status_code also contains codes for successful action, I wasn't sure if it is used by someone or it can be useful in some applications, hence I kept it. Doesn't seem to be used currently though.

frontend/app/controllers/spree/coupon_codes_controller.rb
svyatogor: Great idea! I'll try to do it a bit later.

core/app/models/spree/promotion_handler/coupon.rb
kennyadsl: Thanks, not sure why it's not present in the link I posted. A couple of things here:

1. I think my first comment is still valid. I think it's better to hide the instance variable here, there's no reason to use it if we have the `attr_reader`.
2. I'm not sure to get why we need to clear all the other errors when setting a new one. I mean, if I need to set multiple errors for a promotions handling and I call `set_error_code` twice, does the second `set_error_code` remove the first error just set? This is the core of this change and specs are successful so I suspect I'm just missing something. 🙂

frontend/app/controllers/spree/coupon_codes_controller.rb
```diff
@@ -17,9 +19,12 @@ def create
             if handler.successful?
               flash[:success] = handler.success
               redirect_to cart_path
-            else
-              flash.now[:error] = handler.error
+            elsif handler.errors.count == 1
```
kennyadsl, line 22: I have mixed feelings here but I probably prefer to use a generic solution (a generic message for failed coupon code application and always relying on the views for explaining details). I get why you think this is the best approach but I think we can make this change safely, it's not a big deal for backward compatibility to show a custom message here and we can save some `if`s in the view as well.

kennyadsl (commented):
Made a second review, let me know your thoughts and thanks again!

waiting-for-dev:
Hey @svyatogor, this is a bit old, but would you like to retake this work? Thanks! :slightly_smiling_face:

waiting-for-dev:
Closing it, but if you find the time to keep working on it, I'll be more than happy to reopen 🙂

PR #3899, opened 2021-01-19
Convert jQuery data attribute to number

spaghetticode:
JQuery numeric string data attributes are not casted to numbers consistently. 

There are instances when they are kept as string, this behavior seems to be related to the presence of insignificant
zeros in the original string:

```javascript
  $('<div data-price="12.00"></div>').data('price')
  > "12.00"

  $('<div data-price="12.10"></div>').data('price')
  > "12.10"
```

but:

```javascript
  $('<div data-price="12.1"></div>').data('price')
  > 12.1
```

So, we need to enforce the typecast when summing those attributes in the function `totalSelectedReturnItemAmount`.

After the sum, since we want to show a 2 decimal number, we need to convert the total amount with `toFixed(2)`.

|  Before   |  After   |
|---|---|
| Screen Shot 2021-01-19 at 9 40 02 AM | Screen Shot 2021-01-19 at 9 45 02 AM |

**Checklist:**

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/tables/selectable_table/sum_return_item_amount.js
+++ b/backend/app/assets/javascripts/spree/backend/views/tables/selectable_table/sum_return_item_amount.js
@@ -15,8 +15,11 @@ Spree.Views.Tables.SelectableTable.SumReturnItemAmount = Backbone.View.extend({
   },
 
   totalSelectedReturnItemAmount: function () {
-    var totalAmount = 0.00;
+    var totalAmount = 0;
     var selectedItems = [];
+    var decimals = 0;
+    var separator = Spree.t('currency_separator');
+    var amount, decimalAmount;
 
     if(this.model.get('allSelected')) {
       selectedItems = $('.selectable');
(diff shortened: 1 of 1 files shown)


kennyadsl:
@spaghetticode nice finding. Did you have the chance to understand if this has different implications with other locales?

spaghetticode:
@kennyadsl I did not. I'm assuming everyone is converting to 2 decimals as of today, if that's what concerns you about locales. It would not be that hard to change it to use the digits that are found in the data attribute numbers.

Also, I was able to find the reason for this strange behavior, see this line: https://github.com/jquery/jquery/blob/a32cf6324f8f2190e66a687e94be9687ebf840b7/src/data.js#L34
```javascript
"12.01" === +"12.01" + ""
> true

"12.00" === +"12.00" + ""
> false
```

aldesantis:
🤦 

Thanks for the fix @spaghetticode!

As for the precision, I'm pretty sure there are currencies that use a different numbers of decimals (e.g., the Japanese Yen comes to mind — as far as I know, they use no decimals at all). Perhaps we can get this information from the Money gem?

spaghetticode:
@aldesantis thanks for the input, it now works properly also for those countries that don't have decimal amounts, and the number is properly localized too. Unfortunately, we don't have access to the currency here, and adding it would make things a bit too complicated, still can leverage some locale information such as the currency delimiter and separator that IMHO will suffice.

aldesantis (approved):
@spaghetticode awesome, thanks!

kennyadsl (approved):
Thanks!

kennyadsl:
Failing spec is already addressed in master

PR #3901, opened 2021-01-19
Fix ActionMailer preview loading

### LINKED ISSUE #3898: Deprecation Warning pertaining to Mailer Previews being loaded from an app which loads the Solidus engine
It appears that the way in which Solidus loads ActionMailer previews from a containing app during initialization is today causing a deprecation warning, but in a future version of Rails will be a hard error.

**Solidus Version**:
2.11.3

**To Reproduce**
Create an app which loads Solidus. Ensure that the app has its own ActionMailer previews and then attempt to run the app server/console. Tail the development log to see the following:

DEPRECATION WARNING: Initialization autoloaded the constants ...

Included in this list are MailerPreviews from the containing application 

**Current behavior**
A deprecation warning is presented during boot

**Expected behavior**
No such deprecation warning

 - OS: OSX 11.1
 - Browser N/A
 - Version N/A

aldesantis:
**Description**

This fixes #3898, by moving the loading of the app's previews (and our own) to an `after_initialize` block, which gets rid of the deprecation warning about constant loading during app initialization.

I also managed to write a test that verifies that both the engine's and the app's previews are properly loaded. I had to enable preview loading in `Rails.env.test?` as well in order for the test to pass, but I think that won't be a problem.

Btw, it seems like we'll be able to get rid of this hack if/when https://github.com/rails/rails/pull/31595 gets merged.

**Checklist:**

### DIFF
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -85,8 +85,9 @@ module Spree
         # We need to make sure we call `Preview.all` before requiring our
         # previews, otherwise any previews the app attempts to add need to be
         # manually required.
-        if Rails.env.development?
+        if Rails.env.development? || Rails.env.test?
           ActionMailer::Preview.all
+
           Dir[root.join("lib/spree/mailer_previews/**/*_preview.rb")].each do |file|
             require_dependency file
           end
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -80,14 +80,12 @@ module Spree
             caller
           )
         end
-      end
 
-      # Load in mailer previews for apps to use in development.
-      # We need to make sure we call `Preview.all` before requiring our
-      # previews, otherwise any previews the app attempts to add need to be
-      # manually required.
-      if Rails.env.development?
-        initializer "spree.mailer_previews" do
+        # Load in mailer previews for apps to use in development.
+        # We need to make sure we call `Preview.all` before requiring our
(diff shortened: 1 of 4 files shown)


jarednorman:
Is this seeking to resolve #3898?

PeteTheSadPanda:
I believe it would as it would push load_previews' descendants call into runtime and out of intializer time which results in autoloading of the Preview classes.

aldesantis:
@jarednorman @PeteTheSadPanda yep, this solves #3898. It needs some more love — mostly, I want to implement some kind of caching/memoization and give that anonymous module its own place — but I've tested it in a real Solidus store and both Solidus' and the store's previews are loaded correctly.

PeteTheSadPanda:
Given how frequently this gets called, I'm not sure caching/memorization is really needed. This call primarily gets called when rendering /rails/previews which itself does some level of caching by short circuiting is descendants are populated, so really you might only be requiring Solidus previews.

I'm not advocating against your proposal which is fine, I'm just stating that if it proves too much a pain it wouldn't be a blocker (to me). I believe (but am not certain) that any subsequent calls to the same file via require_dependency would be cached assuming the constant is still in memory and hasn't been unloaded due to it being changed.

aldesantis:
This should be ready for a review!

kennyadsl (approved):
👍 Thanks Ale!

spaghetticode (approved):
@aldesantis thanks!

kennyadsl:
Failing specs already addressed, merging anyway!

Issue #3906, opened 2021-01-21
2.11.4 breaks modified factories in default paths


tmtrademarked:
In 2.11.4, attempting to modify a factory defined in the `spec/factories/` directory of an application ends up failing with a `Factory not registered` error. This same factory worked just fine in 2.11.3, but fails in 2.11.4.

For context, here's an example of one of our factories which extends a built in Solidus factory:
```
# file: spec/factories/option_types.rb

# frozen_string_literal: true
require 'spree/testing_support/factories'

FactoryBot.modify do
  factory :option_type do
    description { 'Testing testing testing' }
  end
end
```

**Solidus Version**:
2.11.4

**To Reproduce**
1) Author a factory that modifies a built-in factory as described above.
2) Run any spec of your application.

**Current behavior**
Fails to run with `Factory not registered` error.

**Expected behavior**
Specs should run and find the modified factories.

**Context**

It looks like the problem is this line here: https://github.com/solidusio/solidus/blob/master/core/lib/spree/testing_support.rb#L36

Assuming I'm understanding the behavior of the FactoryBot railtie correctly, I believe this puts "spec/factories/<foo>_factory.rb" into the `FactoryBot.definition_file_paths` list *first*, which means it attempts to load these factories *before* it loads the default Spree factories.

It seems like this may be fixable either by reordering the traversal of the file definitions for FactoryBot, or perhaps by documenting a better naming convention/location for overrides of Solidus default factories.

jarednorman:
@elia This seems to relate to your work on #3814. Do you have any thoughts here?

elia:
> Assuming I'm understanding the behavior of the FactoryBot railtie correctly, I believe this puts "spec/factories/_factory.rb" into the `FactoryBot.definition_file_paths` list _first_, which means it attempts to load these factories _before_ it loads the default Spree factories.
> 
> It seems like this may be fixable either by reordering the traversal of the file definitions for FactoryBot, or perhaps by documenting a better naming convention/location for overrides of Solidus default factories.

This is correct, unfortunately all the different workarounds that were employed to make factories work with solidus in all the applications out there are going to pose an issue, and the loading sequence was not addressed in the original fix.

Anyway the solution seems to involve unshifting the solidus factory path instead of appending them, I'm going to send a PR to fix that briefly, in the meantime you can try this in your app:

Add this into `config/application.rb` (not inside an initializer because of the loading sequence):

```rb
# config/application.rb

module MyApp
  class Application < Rails::Application
[...]

PR #3907, opened 2021-01-22
Fix factory loading

elia:
**Description**

The current FactoryBot support code doesn't take into account that Solidus factories need to be loaded before app-level factories, to allow modifications.

With this change using `factory_bot_rails` it will work out of the box (the initializer works with v4.8 up to v6).
Nothing is done if `FactoryBot` is used directly, assuming the a more fine grained control/configuration is being used.
The code will be activated only if FactoryBotRails is defined, so to avoid running it in production is enough to keep the gem in the developmen/test groups as advised by the factory_bot readme.

Ref https://github.com/solidusio/solidus/issues/3906 (this issue will be fixed by a back port of this PR to v2.11).

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/factory_bot.rb
+++ b/core/lib/spree/testing_support/factory_bot.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+require "factory_bot"
+begin
+  require "factory_bot_rails"
+rescue LoadError
+end
+
+module Spree
+  module TestingSupport
+    module FactoryBot
+      SEQUENCES = ["#{::Spree::Core::Engine.root}/lib/spree/testing_support/sequences.rb"]
+      FACTORIES = Dir["#{::Spree::Core::Engine.root}/lib/spree/testing_support/factories/**/*_factory.rb"].sort
+      PATHS = SEQUENCES + FACTORIES
+
+      def self.definition_file_paths
+        @paths ||= PATHS.map { |path| path.sub(/.rb\z/, '') }
+      end
+
+      def self.deprecate_cherry_picking
+        # All good if the factory is being loaded by FactoryBot.
+        return if caller.find { |line| line.include? "/factory_bot/find_definitions.rb" }
+
+        Spree::Deprecation.warn(
+          "Please do not cherry-pick factories, this is not well supported by FactoryBot. " \
+          'Use `require "spree/testing_support/factories"` instead.', caller(2)
+        )
+      end
+
+      def self.check_version
+        require "factory_bot/version"
+
+        requirement = Gem::Requirement.new("~> 4.8")
+        version = Gem::Version.new(::FactoryBot::VERSION)
+
+        unless requirement.satisfied_by? version
+          Spree::Deprecation.warn(
+            "Please be aware that the supported version of FactoryBot is #{requirement}, " \
+            "using version #{version} could lead to factory loading issues.", caller(2)
+          )
+        end
+      end
+
+      def self.add_definitions!
+        ::FactoryBot.definition_file_paths.unshift(*definition_file_paths).uniq!
+      end
+
+      def self.add_paths_and_load!
+        add_definitions!
+        ::FactoryBot.reload
+      end
+    end
+  end
+end
--- a/core/lib/spree/testing_support/factory_bot.rb
+++ b/core/lib/spree/testing_support/factory_bot.rb
@@ -17,13 +17,27 @@ module Spree
         @paths ||= PATHS.map { |path| path.sub(/.rb\z/, '') }
       end
 
+      def self.when_cherry_picked
+        callsites = caller
+
+        # All good if the factory is being loaded by FactoryBot or from `testing_support/factories.rb`.
+        return if callsites.find do |line|
+          line.include?("/factory_bot/find_definitions.rb") ||
+          line.include?("/spree/testing_support/factories.rb")
+        end
+
+        yield
+      end
+
       def self.deprecate_cherry_picking
-        # All good if the factory is being loaded by FactoryBot.
-        return if caller.find { |line| line.include? "/factory_bot/find_definitions.rb" }
+        callsites = caller
+        core_root = Spree::Core::Engine.root.to_s
+        index = callsites.index { |line| !line.start_with? core_root }
 
         Spree::Deprecation.warn(
-          "Please do not cherry-pick factories, this is not well supported by FactoryBot. " \
-          'Use `require "spree/testing_support/factories"` instead.', caller(2)
+          "Please do not cherry-pick factories, this is not well supported by FactoryBot, " \
+          'follow the changelog instructions on how to migrate your current setup.',
+          callsites[index..-1]
         )
       end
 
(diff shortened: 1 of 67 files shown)


core/lib/spree/core/engine.rb
```diff
@@ -94,6 +94,14 @@ class Engine < ::Rails::Engine
           end
         end
       end
+
+      if defined?(FactoryBotRails)
```
kennyadsl, line 98: I don't understand why we can't use the default way of setting factories for an Engine as described [here](https://github.com/thoughtbot/factory_bot_rails#automatic-factory-definition-loading)? Is this approach excluded for a reason?

kennyadsl (commented):
Left a preliminary question about the approach used, thanks @elia!

core/lib/spree/core/engine.rb
elia: If you're referring to the use of the Rails-style `config.factory_bot.definition_file_paths` that's not available for FactoryBot v4 (don't remember it it was introduced in 5 or 6), by updating directly `FactoryBot.definition_file_paths` we're compatible with all versions in the 4..6 range.

The `defined?` check is necessary because FactoryBot might not be available in some environments (e.g. production) and the check is done on `FactoryBotRails` with the assumption that if you're not using it, you probably also have some custom setup we don't want to mess with. In this way we can rely on the initializer as hook to inject the Solidus paths.

About using an initializer with an `after:` condition, the main reason for this is that custom factories may depend on the solidus factories, so we need a way to enforce loading them before the custom ones. Specifically `FactoryBotRails` has an initializer named `"factory_bot.set_factory_paths"` that will reset the paths, so if we add the solidus paths before that they'll be wiped out by this initializer.

kennyadsl: Ok, it makes sense now, thanks! So, next question: why still draft?

elia: I was waiting for the CI to be green, but when I pushed last week master was temporarily red, so time passed by 😉 
I just fixed the linting suggestion and repushed, will mark it as ready in a moment 👍

core/lib/spree/testing_support/factory_bot.rb
```diff
@@ -0,0 +1,54 @@
...
+        @paths ||= PATHS.map { |path| path.sub(/.rb\z/, '') }
+      end
+
+      def self.deprecate_cherry_picking
+        # All good if the factory is being loaded by FactoryBot.
+        return if caller.find { |line| line.include? "/factory_bot/find_definitions.rb" }
+
+        Spree::Deprecation.warn(
+          "Please do not cherry-pick factories, this is not well supported by FactoryBot. " \
+          'Use `require "spree/testing_support/factories"` instead.', caller(2)
```
kennyadsl, line 26: `caller(1)` is giving me more accurate details when running specs in a real store: 

```
# caller(2) example result
DEPRECATION WARNING: Please do not cherry-pick factories, this is not well supported by FactoryBot. Use `require "spree/testing_support/factories"` instead. (called from block in <top (required)> at /Users/kennyadsl/Code/nebulab/solidus/core/lib/spree/testing_support/factories.rb:11)

# caller(1) example result
DEPRECATION WARNING: Please do not cherry-pick factories, this is not well supported by FactoryBot. Use `require "spree/testing_support/factories"` instead. (called from <top (required)> at /Users/kennyadsl/Code/nebulab/solidus/core/lib/spree/testing_support/factories/zone_factory.rb:4)

```

elia: I updated the code to mention the changelog instruction as a way to fix the deprecation and to point it to the first non-core line from the backtrace so ideally it will point right in the place where the cherry-picking is happening (i.e. inside the rails application).

tvdeyen:
FYI: This change and the PR that introduced it breaks every gem/app in our eco system in a minor version that loads individual factories. We probably should revert it in 2.11 or add a fix that still supports loading individual factories and only remove it in 3.0

kennyadsl:
@tvdeyen thanks for the feedback, the goal of this PR is to still provide support for loading individual factories without breaking any application. Do you have any evidence that it is breaking things? I tried this PR with a couple of real stores and they just give deprecation warnings. Of course, I don't expect to have covered all possible scenarios for factory_bot configuration.

tvdeyen:
@kennyadsl I have not tested this PR, but this CI run is from a recent Solidus v2.11

https://github.com/AlchemyCMS/alchemy-solidus/runs/1888940075?check_suite_focus=true

tvdeyen:
@elia @kennyadsl the switch to mutate FactoryBot factory load paths to load factories breaks the current way of requiring individual factories in tests suites of gems and apps. The errors you are getting are missing factories, duplicated factories and missing sequences. I think there is no easy way of supporting both ways the same time.

I agree that the FactoryBot way is better and we should not require individual factories in test suites, but this is an widely used approach and we should figure out a way to a) support both or b) revert this change in 2.11

kennyadsl:
For gems, I made [this PR on Solidus Support](Fix loading factories in extensions after the last changes in Solidus solidusio/solidus_dev_support#169) that, once this is merged, "should" fix all extensions. I only tried with a couple so I'm not really sure it works for everyone. 

For stores, reverting breaking changes is exactly what this PR has been made for. If you can help us understand the exact problem you are facing, we'd be happy to explore it or if not solvable, I agree we should revert and add this breaking behavior in a major.

kennyadsl:
@tvdeyen I'm taking a look at `alchemy-solidus` (checking that out to the commit before your fix in master) with this branch and I think you are right. 

There's one big thing we didn't consider: when cherry-picking a file we are showing a deprecation message but not loading the specific dependency of that factory anymore. 

For example when requiring the [product_factory](https://github.com/solidusio/solidus/blob/70979e8dd1df543f515314a317787f90b28eae1a/core/lib/spree/testing_support/factories/product_factory.rb) individually, we print the deprecation message but we are not re-requiring all the factories that this file was relying on, which were:

```
require 'spree/testing_support/sequences'
require 'spree/testing_support/factories/shipping_category_factory'
require 'spree/testing_support/factories/stock_location_factory'
require 'spree/testing_support/factories/tax_category_factory'
require 'spree/testing_support/factories/product_option_type_factory'
```

I'll talk with @elia to see if there's a way to make it work but it's becoming uncomfortable to keep adding code again and again to fix all the possible cases, so I'd strongly consider a revert. 

Thanks again for pointing this out.

kennyadsl:
@tvdeyen the last pushed code restores backward compatibility to require factories individually, I tested it with alchemy-solidus (checking out to a commit previous to your fix in https://github.com/AlchemyCMS/alchemy-solidus/pull/72). 

I think that after some git commit improvements we can merge and backport this into 2.11. Projects that are requiring factories individually will keep their stores/gems working and we also print a deprecation warning since this behavior will not be supported anymore with Solidus 3.0.

What do you think?

tvdeyen:
@kennyadsl wow. Thanks. This looks promising 🙏🏻 @mamhoff and I will test this and provide feedback asap.

tvdeyen (approved):
Looks great.

kennyadsl (approved):
Worked together with @elia here. 

We tested this change on a real application, 2-3 extensions and an external gem (`alchemy-solidus`) that was cherry-picking individual factories. Apparently fixed the backward-compatibility issues in all of them, printing a deprecation warning as well.

PR #3908, opened 2021-01-22
Move Spree::Address#name attribute to the db

filippoliverani:
**Description**
Ref https://github.com/solidusio/solidus/issues/3234

This PR pushes forward the migration to `Spree::Address#name` to replace `firstname` and `lastname` attributes. It adds a correseponding `name` column to the DB and replaces all the remaining direct attribute accesses to `firstname` and `lastname` with `name`.

~The DB migration contains a blocking warning to force users to evaluate the impact of the upgrade and to choose the correct strategy to handle it in their store. In a future PR we will provide a complementary rake task to migrage existing data from `firstname` and `lastname` columns to the new `name` columns.  New stores won't receive any warning and will be able to start working with the `name` field from scratch.~

☝️ This piece will be moved into a separate rake task.

**Checklist:**

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -22,6 +22,7 @@ module Spree
       self.class.state_validator_class.new(self).perform
     end
 
+    self.ignored_columns = %w(firstname lastname)
     DB_ONLY_ATTRS = %w(id updated_at created_at).freeze
     TAXATION_ATTRS = %w(state_id country_id zipcode).freeze
 
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -24,9 +24,8 @@ module Spree
 
     DB_ONLY_ATTRS = %w(id updated_at created_at).freeze
     TAXATION_ATTRS = %w(state_id country_id zipcode).freeze
-    LEGACY_NAME_ATTRS = %w(firstname lastname).freeze
 
-    self.whitelisted_ransackable_attributes = %w[firstname lastname]
+    self.whitelisted_ransackable_attributes = %w[name]
 
     scope :with_values, ->(attributes) do
       where(value_attributes(attributes))
@@ -160,29 +159,6 @@ module Spree
 
     def country_iso
       country && country.iso
-    end
-
-    # @return [String] the full name on this address
-    def name
-      Spree::Address::Name.new(
-        read_attribute(:firstname),
-        read_attribute(:lastname)
-      ).value
-    end
-
-    def name=(value)
-      return if value.nil?
-
-      name_from_value = Spree::Address::Name.new(value)
-      write_attribute(:firstname, name_from_value.first_name)
-      write_attribute(:lastname, name_from_value.last_name)
-    end
-
-    def as_json(options = {})
-      super.tap do |hash|
-        hash.except!(*LEGACY_NAME_ATTRS)
-        hash['name'] = name
-      end
     end
   end
 end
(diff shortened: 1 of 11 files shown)


core/app/models/spree/address.rb
```diff
@@ -161,28 +160,5 @@ def country_iso=(iso)
...
-
-    def name=(value)
-      return if value.nil?
-
-      name_from_value = Spree::Address::Name.new(value)
-      write_attribute(:firstname, name_from_value.first_name)
-      write_attribute(:lastname, name_from_value.last_name)
-    end
-
-    def as_json(options = {})
```
kennyadsl, line 181: @filippoliverani so address's JSON representation will still contain firstname and lastname?

kennyadsl: We added `Spree::Address.ignore_columns` to address this.

core/db/migrate/20210122110141_add_name_to_spree_addresses.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class AddNameToSpreeAddresses < ActiveRecord::Migration[5.2]
+  class AddressForMigration < ApplicationRecord
+    self.table_name = 'spree_addresses'
+  end
+
+  def up
+    if AddressForMigration.where.not(firstname: [nil, '']).exists?
+      AddNameToSpreeAddresses.handle_exisiting_data
+    end
```
tvdeyen, line 11: How should a store be able to add the `name` column if we stop the migration here when existing addresses exist? Move all names into the `lastname` column first?

I think adding the `name` column first should be fine.

tvdeyen (commented):
Should we just introduce the `name` column first?

core/db/migrate/20210122110141_add_name_to_spree_addresses.rb
kennyadsl: @tvdeyen this migration is intentionally stopped before the change for people that do have at least one record that needs manual attention. They can change the migration according to their context. We will provide some alternatives in following PRs.

tvdeyen: But adding a new column should be ok, no? The exception text talks about moving the data. Where should the data be moved to if there is no column? Maybe just print the message out after we introduced the column?

kennyadsl: Adding the new column and then raise will make the migration execute and rollback. That works but I can't see a lot of difference now. 

But yes, when we'll introduce some alternatives approaches for people to choose (as we did in [this other migration](https://github.com/solidusio/solidus/blob/bfc998da4712ad3b04372462439e0853b84cf9c3/core/db/migrate/20190106184413_remove_code_from_spree_promotions.rb#L30)), it will probably be better to have this check after the column addition so that we don't ask to also move the code around.

kennyadsl: @tvdeyen We removed the data migration helper code. It is better to provide an upgrade rake task that will handle this and other things needed for the major upgrade. We can expect broken apps if that task will not be executed but after all, that's what a major does, isn't it?

tvdeyen: Agreed 👍🏻

tvdeyen (approved):

spaghetticode (approved):
@filippoliverani @kennyadsl thanks 👍 🎉

jarednorman (approved):
Yes, looks good to me.

Issue #3912, opened 2021-01-30
Searching deleted products by SKU does not work


brchristian:
This search box for Products allows the searching of deleted products.

<img width="552" alt="Screen Shot 2021-01-29 at 4 18 30 PM" src="https://user-images.githubusercontent.com/2460418/106340121-a1b61280-624d-11eb-84af-35fe3a6ca189.png">

Searching for deleted products by name works as expected. However, searching for deleted products by SKU does not work.

**Solidus Version**:
Current `master` (b0092d0f3bda447494a586bf8eb96a18322b8bed)

**To Reproduce**
Start sandboxed Solidus store.
Go to store_url/admin/products.
Delete a product.
Click the "Show Deleted" checkbox.
Search for that product by name: it is in the results.
Search for that product by SKU: it is not in the results.

**Current behavior**
The product appears in the results only when searching by name, not by SKU.

**Expected behavior**
The product appears in the results when searching by name or SKU.

**Screenshots**

Here we see that searching by name shows one deleted product and one undeleted product:
<img width="507" alt="Screen Shot 2021-01-29 at 4 21 57 PM" src="https://user-images.githubusercontent.com/2460418/106340286-1e48f100-624e-11eb-9c47-3b9e52297147.png">

But here we see that searching by the beginning of the SKU shows ONLY the undeleted product:
<img width="515" alt="Screen Shot 2021-01-29 at 4 22 15 PM" src="https://user-images.githubusercontent.com/2460418/106340304-2acd4980-624e-11eb-894f-772568892717.png">

And here we see that searching by the exact SKU shows no results at all:
<img width="518" alt="Screen Shot 2021-01-29 at 4 22 30 PM" src="https://user-images.githubusercontent.com/2460418/106340310-31f45780-624e-11eb-8e51-0aa9b98fb8ce.png">

PR #3917, opened 2021-02-03
Ensured deleted products are visible in the admin product index page when searching by sku

mikekibz59:
**Description**

 Changed the `Spree::SoftDeletable` default_scope from `default_scope { kept }` to 
 `default_scope -> { kept }`.  I did this because the initial implementation caused select
  queries to include the default scope even when directed not to. Such that when you would
  use the `with_discarded` method.
  Also, change the `with_variant_sku_cont`  scope to include variants that have been deleted so as to
  allow deleted products to be shown on the index page.
  
  Pictures of changes:
   
Screen Shot 2021-02-03 at 1 16 25 PM

  Ref #https://github.com/solidusio/solidus/issues/3912

**Checklist:**

### DIFF
--- a/core/app/models/spree/product/scopes.rb
+++ b/core/app/models/spree/product/scopes.rb
@@ -197,7 +197,8 @@ module Spree
           def self.with_variant_sku_cont(sku)
             sku_match = "%#{sku}%"
             variant_table = Spree::Variant.arel_table
-            subquery = Spree::Variant.where(variant_table[:sku].matches(sku_match).and(variant_table[:product_id].eq(arel_table[:id])))
+            subquery = Spree::Variant.with_discarded
+                      .where(variant_table[:sku].matches(sku_match).and(variant_table[:product_id].eq(arel_table[:id])))
             where(subquery.arel.exists)
           end
 
--- a/core/spec/models/spree/product/scopes_spec.rb
+++ b/core/spec/models/spree/product/scopes_spec.rb
@@ -183,4 +183,28 @@ RSpec.describe "Product scopes", type: :model do
       end
     end
   end
+
+  describe '.with_variant_sku_cont' do
+    let!(:product) { create(:product, name: 'ring cap') }
+    let!(:variant_1) { create(:variant, product: product, sku: 'H100') }
+    let!(:variant_2) { create(:variant, product: product, sku: 'H290') }
+
+    context 'query products' do
+      it 'scope returns empty array' do
+        product.discard
+        expect(Spree::Product.with_variant_sku_cont('H1').to_a).to eq []
+      end
+
+      it 'scope returns product' do
+        product = Spree::Product.with_variant_sku_cont('H1').first
+        expect(product.name).to eq 'ring cap'
+      end
+
+      it 'returns product when with_discarded is passed on product' do
+        product.discard
+        product = Spree::Product.with_discarded.with_variant_sku_cont('H1').first
+        expect(product.name).to eq 'ring cap'
+      end
+    end
+  end
 end
--- a/backend/spec/features/admin/products/products_spec.rb
+++ b/backend/spec/features/admin/products/products_spec.rb
@@ -151,6 +151,29 @@ describe "Products", type: :feature do
         expect(page).not_to have_content("zomg shirt")
         expect(page).to have_css('#listing_products > tbody > tr', count: 1)
       end
+
+      it "should be able to search for deleted products by sku" do
+        create(:product, name: 'nginx cap ', sku: 'B201', deleted_at: "2021-01-02 18:21:13")
+        create(:product, name: 'solidus shirt', sku: 'GG201')
+
+        click_nav "Products"
+        expect(page).to have_content("solidus shirt")
+        expect(page).not_to have_content("nginx cap ")
+
+        # when show deleted is checked
+        fill_in 'SKU', with: "B2"
+        check 'Show Deleted'
+        click_button 'Search'
+        expect(find('input[name="q[with_discarded]"]')).to be_checked
+        expect(page).to have_content('nginx cap ')
+        expect(page).not_to have_content('solidus shirt')
+
+        fill_in 'SKU', with: 'GG'
+        click_button 'Search'
+        expect(find('input[name="q[with_discarded]"]')).to be_checked
+        expect(page).not_to have_content('nginx cap ')
+        expect(page).to have_content('solidus shirt')
+      end
     end
 
     context "creating a new product" do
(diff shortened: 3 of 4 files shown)


core/app/models/spree/product/scopes.rb
```diff
@@ -197,7 +197,8 @@ def self.available(available_on = nil)
           def self.with_variant_sku_cont(sku)
             sku_match = "%#{sku}%"
             variant_table = Spree::Variant.arel_table
-            subquery = Spree::Variant.where(variant_table[:sku].matches(sku_match).and(variant_table[:product_id].eq(arel_table[:id])))
```
kennyadsl, line 200: Not sure to get why we need the scope to be defined here. This way every call to `with_variant_sku_cont` will give us also discarded results, even if not called with the `Show Deleted` option. Can you please help me understand?

kennyadsl (commented):
Seems to make sense, left one question. Thanks!

core/app/models/spree/product/scopes.rb
mikekibz59: when you soft delete a product,  the corresponding variant is deleted(soft delete). This won't be a problem when searching an active SKU but when searching a deleted product, the default select query for the variant won't account for the deleted SKU. And thus we won't get back the deleted products.

kennyadsl: I think we may add some unit tests here to describe this behavior. What do you think?

mikekibz59: yes absolutely. I'll get right to it.

mikekibz59: I have just pushed a commit for tests covering the `with_variant_sku_cont` scope. As for test covering the deletion of the product with their variances, it is covered here:
REF https://github.com/solidusio/solidus/blob/b0092d0f3bda447494a586bf8eb96a18322b8bed/backend/spec/controllers/spree/admin/products_controller_spec.rb#L261

core/spec/models/spree/product/scopes_spec.rb
```diff
@@ -183,4 +183,28 @@
...
+        expect(product.name).to eq 'ring cap'
+      end
+
+      it 'returns product when with_discarded is passed on product' do
+        product.discard
+        product = Spree::Product.with_discarded.with_variant_sku_cont('H1').first
+        expect(product.name).to eq 'ring cap'
+      end
+    end
+  end
```
kennyadsl, line 209: Thanks for adding these specs. What if we also discard the variant instead? Like:

```ruby
result_with_variant_not_discarded = Spree::Product.with_variant_sku_cont('H1').to_a
variant_1.discard
result_with_variant_discarded = Spree::Product.with_variant_sku_cont('H1').to_a
expect(result_with_variant_not_discarded).to eq result_with_variant_discarded
```

I think this expectation should fail and it doesn't with the current code otherwise we have no way of deciding if we want to return discarded variants or not. What do you think?

mikekibz59: yes, it should fail, since the corresponding product isn't discarded. But since the scope is only used within the context of returning products, deleted or not,  with a specific `SKU`; I think that should be the expected behaviour. I might be wrong on this. When we discard a product it also discards its corresponding variants. If we need to search for that product with a `SKU` which has been discarded, we will not get anything back. So I would propose to maybe change this method's name to reflect that it is for searching variants discarded or not; Or we can create a new scope for that purpose.

kennyadsl: > When we discard a product it also discards its corresponding variants. 

But it's not true the other way around. Discarding a variant without discarding the whole product is totally legit.

My concern here is that the search form will return products with discarded variants (not the whole product) searching for the SKU of a discarded variant, even when `Show deleted` is not selected. Can you please verify?

On the technical implementation, I was also thinking at another scope that explicitly tells that we will return also discarded variants, like `.with_all_variant_sku_cont` and maybe we can rename `with_variant_sku_cont` to `with_kept_variant_sku_cont` to reflect discard gem's terminology.

mikekibz59: I have looked into it by discarding a variant of an active(non-discarded) product. 
`Index Action`
It will return that product with that `SKU` even if the variant is discarded. This is the default behaviour whether or not you check the `Show deleted` checkbox and without invoking the `with_variant_sku_cont` scope.
`Search Functionality`
with invoking the `with_variant_sku_cont` scope. It will also return that product with the discarded variant.

So either way, if we discard a variant, but a product is linked to that `SKU`, that variant will be returned. Regardless if we are using the `search functionality` or just the `index`

kennyadsl: > I would suggest to move the SKU to the product model and make it unique since SKUs should be unique to a specific product. That way we will solve both problems.

I don't think we can do that, each variant has a different SKU.

mikekibz59: Then I think it leaves us with few options: 
`first`: to leave the implementation as it is
`second`: to return products that only have undiscarded variances
`Third`: to find a middle ground between the two. (we may need to brainstorm on this one)
 what do you think?

kennyadsl: What about creating a new scope called `with_all_variant_sku_cont` that also check for discarded variants and connect it to the search when `Show deleted` is checked? When `Show deleted` is not selected we'll keep using the `with_variant_sku_cont` scope. Do you think this solution can work or am I missing something?

mikekibz59: yes, it can work. But discarded variants will still be available visible as long as the product is still linked to it.

mikekibz59: The highlighted product is active but its variance is discarded. This is the behavior I was hinting about. 

Screen Shot 2021-02-05 at 9 35 41 AM

kennyadsl: Products can have multiple variants. I think that we can consider product not discarded with only one variant (called master variant) discarded as an edge case. In real life you discard a variant of a product when you have at least another one non discarded. Can you try to see if it make sense discarding a variant of a product that has multiple variants?

mikekibz59: okay, let me check it out.

mikekibz59: The search_form and index action only return master variants. When you delete a different variance of the same product and search it with the old `with_variance_sku_cont` scope it will return that product. The same can be said after adding in my changes. It seems like when searching priority is given to master variance. 
Would it be possible if we can schedule a video call because this looks quite big of a problem

mikekibz59: I have been reviewing your proposed options for adding a scope to explicitly search discarded variances and I think that would be the best approach.

Issue #3427, opened 2019-11-14
Unify `Spree::User` and `Spree::Order` email validation 


spaghetticode:
_(not sure if this is more a bug or a new feature... I'm going for the latter as I think the solution will end up being a new feature 😄 )_

At the moment we have a custom validator for the email in `Spree::Order`. This validator has different logic than the one used in `solidus_auth_devise`, so there may be instances when the user is valid, but the order is not. This creates confusion and can affect shop sales negatively when the issue goes unnoticed.

One possible course of action may be to use the same validator from `solidus_auth_devise`, which happens to be the one used by `devise`, as it's more used (if not respectful of standards) than ours... but only when the application is using `solidus_auth_devise` as well. More in general, if a user validator for email exists, the order should use that one instead of the one provided by Solidus.  

```ruby
email = 'foo.@example.com'

u = Spree::User.new(email: email)
u.valid?
u.errors[:email]
# => [] 

o = Spree::Order.new(email: email)
o.valid?
o.errors[:email]
# => ["is invalid"] 
```

spaghetticode:
A bit more context on why this can be insidious, again from Solidus sandbox:

```ruby
u = Spree::User.create email: 'foo.@example.com', password: 'password', password_confirmation: 'password'

o = u.orders.create
o.persisted?
# => true
o.valid?
# => false
o.errors[:email]
 => ["is invalid"] 
```

The order is valid before being persisted, as the email field is blank. After the order is persisted, it  becomes invalid as the email field, copied from the user, is not valid according to the validator.

rubenochiavone:
Hi there, I took a look at `devise` email validation and it turns out to be a `regexp` similar to `Spree::EmailValidator::EMAIL_REGEXP`. See

https://github.com/plataformatec/devise/blob/b52e642c0131f7b0d9f2dd24d8607a186f18223e/lib/devise.rb#L116

and

https://github.com/plataformatec/devise/blob/07f2712a22aa05b8da61c85307b80a3bd2ed6c4c/lib/devise/models/validatable.rb#L34

Then, replacing `Spree::EmailValidator::EMAIL_REGEXP` by `Devise::email_regexp` when `Devise` is defined would fix the issue and all commands you presented would be okay.

**core/lib/spree/core/validators/email.rb**

```diff
-      unless EMAIL_REGEXP.match? value
+      email_regexp = defined?(Devise) && defined?(Devise::email_regexp) ? Devise::email_regexp : EMAIL_REGEXP
[...]

PR #3920, opened 2021-02-04
Use Devise email_regexp for order e-mail validations when using spree_auth_devise

cesartalves:
## Description

Background: #3427 

The `EmailValidator` provided by spree and the format validation provided by Devise use different regexps. This will lead to a Order containing an email format error, after persistence, if the email is copied from `Spree::User` in [spree_auth_devise](https://github.com/spree/spree_auth_devise), such as when `associate_user!` is called.

It would be logic to use the same email validation for both `Spree::User` and `Order::Entities`.

This is currently only an outline of the problem. I have added a test where the Order's email is copied from an spree-invalid User email, and followed the suggestion from a comment in the issue to sketch a quick solution.

However I am considering the following alternative approaches:
1. change the default `email_regexp used` in `spree_auth_devise` to be the same as the one in `Spree::EmailValidator`. This might be a good option as I feel that the email_regexp in `Spree::EmailValidator` is waaaay too permissible (see References)
2. use `defined?(Devise)` inside [`Spree::Order`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L132) instead of patching `Spree::EmailValidator` (this is my prefered approach in the case of changing solidus itself, but it might not be easy to test as Devise is evaluated on class definition. But a method / lambda validator would make the test added pass)
3. decorate / subclass `Spree::EmailValidator` and use it in the validation of the Order
4. Use this very, VERY black magic to get the email format validator:
```
 Spree.user_class.send(:_validators)[:email]
                            .select { |x| x.class == ActiveModel::Validations::FormatValidator }
                            .first.instance_eval { @options[:with] }
```

I will be committing my preferred solution this weekend ~~and removing the trailing white spaces vscode left~~ :)

## Checklist

### References

https://deliverbility.com/valid-email-address-format/
https://help.xmatters.com/ondemand/trial/valid_email_format.htm#:~:text=A%20valid%20email%20address%20consists,the%20left%20of%20the%20%40%20symbol.&text=For%20example%2C%20in%20the%20address,com%22%20is%20the%20email%20domain.
https://github.com/heartcombo/devise/blob/07f2712a22aa05b8da61c85307b80a3bd2ed6c4c/lib/devise/models/validatable.rb#L34
https://github.com/heartcombo/devise/blob/b52e642c0131f7b0d9f2dd24d8607a186f18223e/lib/devise.rb#L116

### DIFF
--- a/core/lib/spree/core/validators/email.rb
+++ b/core/lib/spree/core/validators/email.rb
@@ -15,7 +15,9 @@ module Spree
     EMAIL_REGEXP = /\A([^@\.]|[^@\.]([^@\s]*)[^@\.])@([^@\s]+\.)+[^@\s]+\z/
 
     def validate_each(record, attribute, value)
-      unless EMAIL_REGEXP.match? value
+      email_regexp = defined?(Devise) ? Devise.email_regexp : EMAIL_REGEXP
+
+      unless email_regexp.match? value
         record.errors.add(attribute, :invalid, { value: value }.merge!(options))
       end
     end
--- a/core/spec/models/spree/order/validations_spec.rb
+++ b/core/spec/models/spree/order/validations_spec.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'rails_helper'
+require "rails_helper"
 
 module Spree
   RSpec.describe Spree::Order, type: :model do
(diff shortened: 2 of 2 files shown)


cesartalves:
Closing this PR as I drafted a more flexible solution at https://github.com/solidusio/solidus/pull/3927

PR #3922, opened 2021-02-05
Skip adding webpacker gem when generating dummyapp

SamuelMartini:
**Description**

Currently, the dummy app generated to test extensions has the `javascript_pack_tag` helper included in `layouts/application.html.erb`. This is because `skip-webpack-install` only prevents rails to run the webpacker install task but being webpacker the default for Rails 6, it is still included in the generated view.

`--skip-javascript` avoids webpacker to be included in the rails generated app.[1]

Related pr https://github.com/solidusio/solidus/pull/3326 where `--skip-javascript` is used in the sandbox generation.

[1]: https://github.com/rails/rails/issues/35484#issuecomment-469798114

**Checklist:**

### DIFF
--- a/core/lib/generators/spree/dummy/dummy_generator.rb
+++ b/core/lib/generators/spree/dummy/dummy_generator.rb
@@ -43,7 +43,7 @@ module Spree
       opts[:skip_test] = true
       opts[:skip_yarn] = true
       opts[:skip_bootsnap] = true
-      opts[:skip_webpack_install] = true
+      opts[:skip_javascript] = true
 
       puts "Generating dummy Rails application..."
       invoke Rails::Generators::AppGenerator,


core/lib/generators/spree/dummy/dummy_generator.rb
```diff
@@ -44,6 +44,7 @@ def generate_test_dummy
       opts[:skip_yarn] = true
       opts[:skip_bootsnap] = true
       opts[:skip_webpack_install] = true
+      opts[:skip_javascript] = true
```
kennyadsl, line 47: Is `skip_webpack_install` needed if we have `skip_javascript`?

SamuelMartini: Actually not! 👍 
Pushing the `skip_webpack_install` removal.

SamuelMartini: Reference https://github.com/rails/rails/blob/6-1-stable/railties/lib/rails/generators/app_base.rb#L408

kennyadsl (approved):
Thanks!

spaghetticode (approved):
@SamuelMartini thanks for the fix! 👍

PR #3923, opened 2021-02-05
Deprecate try_spree_current_user

elia:
**Description**

It was introduced 9 years ago to overcome an obscure call loop in the early days of Spree (a00d703) and to support both a custom helper (`spree_current_user`) and the helper generated by Devise (`current_spree_user`).

Then, 8 years ago, `solidus_auth_devise` actually added support for the official custom helper (`spree_current_user`) making
`try_spree_current_user` finally unnecessary (https://github.com/solidusio/solidus_auth_devise/commit/69269765).

**Checklist:**

### DIFF
--- a/core/lib/spree/core/controller_helpers/auth.rb
+++ b/core/lib/spree/core/controller_helpers/auth.rb
@@ -33,7 +33,7 @@ module Spree
 
         # Needs to be overriden so that we use Spree's Ability rather than anyone else's.
         def current_ability
-          @current_ability ||= Spree::Ability.new(try_spree_current_user)
+          @current_ability ||= Spree::Ability.new(spree_current_user)
         end
 
         def redirect_back_or_default(default)
--- a/core/lib/spree/core/controller_helpers/auth.rb
+++ b/core/lib/spree/core/controller_helpers/auth.rb
@@ -18,6 +18,7 @@ module Spree
         included do
           before_action :set_guest_token
           helper_method :try_spree_current_user
+          helper_method :spree_current_user
 
           class_attribute :unauthorized_redirect
           self.unauthorized_redirect = -> do
@@ -65,6 +66,8 @@ module Spree
             current_spree_user
           end
         end
+
+        deprecate try_spree_current_user: :spree_current_user, deprecator: Spree::Deprecation
       end
     end
   end
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -18,6 +18,9 @@ RAILS_6_OR_ABOVE = Rails.gem_version >= Gem::Version.new('6.0')
 # @private
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+
+  def spree_current_user
+  end
 end
 
 # @private
--- a/backend/app/controllers/spree/admin/base_controller.rb
+++ b/backend/app/controllers/spree/admin/base_controller.rb
@@ -33,7 +33,7 @@ module Spree
       # Need to generate an API key for a user due to some backend actions
       # requiring authentication to the Spree API
       def generate_admin_api_key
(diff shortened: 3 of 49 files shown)


spaghetticode (approved):
I like this deprecation, @elia thank you 👍

jarednorman (commented):
Would it be preferred to fix the custom user generator? We get people in the support channel asking for help around this use case.

elia:
(sorry for the hiatus)
@jarednorman I agree we should fix the custom user generator, but in the meanwhile I believe it's better not to have docs rather than misleading/broken ones. Anyway I removed that commit and will move it to a different issue/PR so it can be dealt with properly 👍

jarednorman (approved):

waiting-for-dev (approved):
It's always so lovely removing tech debt. Thanks! :slightly_smiling_face:

waiting-for-dev:
Hey @elia, not sure why this PR got staled. Would you mind rebasing from master? :pray:

core/lib/spree/core/controller_helpers/auth.rb
```diff
@@ -18,6 +18,7 @@ module Auth
         included do
           before_action :set_guest_token
           helper_method :try_spree_current_user
+          helper_method :spree_current_user
```
jarednorman, line 21: The commit on solidus_auth_devise referenced in the commit message has a link that isn't valid; it should be `https://github.com/solidusio/solidus_auth_devise/commit/69269765`.

jarednorman (changes requested):
What's the deal with the final commit? It doesn't seem related to the PR title.

waiting-for-dev:
> What's the deal with the final commit? It doesn't seem related to the PR title.

@elia, what about moving it to a separated PR? :pray:

core/lib/spree/core/controller_helpers/auth.rb
elia: Fixed, thanks for catching it!

elia:
@jarednorman @waiting-for-dev extracted the last commit to #4440, the reason behind that was to remove all ties to devise and reduce the amount of hidden coupling we have between solidus and solidus_auth_devise.

core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -18,6 +18,9 @@
 # @private
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+
+  def spree_current_user
```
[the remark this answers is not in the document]
elia: When using `try_spree_current_user` you can get away with not setting up any authentication and it will just return `nil`. By not using it we're relying more on the expectation on `ApplicationController` to implement this method. For test purposes having a no-op method is equivalent of using `try_spree_current_user` without an auth implementation.

jarednorman (approved):

waiting-for-dev:
@elia, do we need to update [solidus_auth_devise calls](https://github.com/solidusio/solidus_auth_devise/search?q=try_spree_current_user) now?

waiting-for-dev:
I wonder if we still need to support the method through `solidus_support`. Otherwise, newer versions of `solidus_auth_devise` won't be compatible with older `solidus` versions.

elia:
> I wonder if we still need to support the method through `solidus_support`. Otherwise, newer versions of `solidus_auth_devise` won't be compatible with older `solidus` versions.

I'm quite sure that since auth-devise will implement `spree_current_user` and that is also a requirement for implementing a custom auth it's not needed. Any extension could switch to be using `spree_current_user` even now without this PR, it's just that we kept that helper around and people went on using it because it _looks_ safer.



> @elia, do we need to update [solidus_auth_devise calls](https://github.com/solidusio/solidus_auth_devise/search?q=try_spree_current_user) now?

Good call, I'll send a pr for auth-devise 👍

PR #3925, opened 2021-02-05
Let the PriceSelector return a Spree::Price

swively:
**Description**
Issue laid out in [this discussion](https://github.com/solidusio/solidus/discussions/3919)

The `Variant::PriceSelector` currently returns a `Spree::Money`, which is misleading and somewhat limiting.
The `Spree::Price` model is useful for its ability to be extended to support things like sales prices, partial prices, etc.

**Checklist:**

### DIFF
--- a/core/app/models/spree/variant/price_selector.rb
+++ b/core/app/models/spree/variant/price_selector.rb
@@ -26,11 +26,24 @@ module Spree
       # @param [Spree::Variant::PricingOptions] price_options Pricing Options to abide by
       # @return [Spree::Money, nil] The most specific price for this set of pricing options.
       def price_for(price_options)
+        Spree::Deprecation.warn(
+          "price_for is deprecated and will be removed. The price_for method
+          should return a Spree::Price as described. Please use
+          #price_for_options and adjust your frontend code to explicitly call
+          &.money where required"
+        )
+        price_for_options(price_options)&.money
+      end
+
+      # The variant's Spree::Price record, given a set of pricing options
+      # @param [Spree::Variant::PricingOptions] price_options Pricing Options to abide by
+      # @return [Spree::Price, nil] The most specific price for this set of pricing options.
+      def price_for_options(price_options)
         variant.currently_valid_prices.detect do |price|
-          ( price.country_iso == price_options.desired_attributes[:country_iso] ||
-            price.country_iso.nil?
+          (price.country_iso == price_options.desired_attributes[:country_iso] ||
+           price.country_iso.nil?
           ) && price.currency == price_options.desired_attributes[:currency]
-        end.try!(:money)
+        end
       end
     end
   end
--- a/api/app/views/spree/api/products/_product.json.jbuilder
+++ b/api/app/views/spree/api/products/_product.json.jbuilder
@@ -4,8 +4,8 @@
 json.cache! [I18n.locale, @current_user_roles.include?('admin'), current_pricing_options, @product_attributes, @exclude_data, product] do
   json.(product, *(@product_attributes - [:total_on_hand]))
   json.total_on_hand(total_on_hand_for(product))
-  json.price(product.price_for(current_pricing_options).try(:to_d))
-  json.display_price(product.price_for(current_pricing_options).to_s)
+  json.price(product.price_for_options(current_pricing_options)&.amount)
+  json.display_price(product.price_for_options(current_pricing_options)&.money&.to_s)
 
   @exclude_data ||= {}
   unless @exclude_data[:variants]
--- a/api/app/views/spree/api/variants/_small.json.jbuilder
+++ b/api/app/views/spree/api/variants/_small.json.jbuilder
@@ -2,8 +2,8 @@
 
 json.cache! [I18n.locale, current_pricing_options, variant] do
   json.(variant, *variant_attributes)
-  json.price(variant.price_for(current_pricing_options).try(:to_d))
-  json.display_price(variant.price_for(current_pricing_options).to_s)
+  json.price(variant.price_for_options(current_pricing_options)&.amount)
+  json.display_price(variant.price_for_options(current_pricing_options)&.money&.to_s)
   json.options_text(variant.options_text)
   json.track_inventory(variant.should_track_inventory?)
   json.in_stock(variant.in_stock?)
--- a/core/app/helpers/spree/base_helper.rb
+++ b/core/app/helpers/spree/base_helper.rb
@@ -130,7 +130,7 @@ module Spree
     end
 
     def display_price(product_or_variant)
-      product_or_variant.price_for(current_pricing_options).to_html
+      product_or_variant.price_for_options(current_pricing_options)&.money&.to_html
     end
 
     def pretty_time(time, format = :long)
--- a/core/app/helpers/spree/products_helper.rb
+++ b/core/app/helpers/spree/products_helper.rb
@@ -38,7 +38,7 @@ module Spree
                   .with_prices(current_pricing_options)
                   .all? { |variant_with_prices| variant_with_prices.price_same_as_master?(current_pricing_options) }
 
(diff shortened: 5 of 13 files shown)


jarednorman:
It also occurred to me that it's perhaps the case that the reason that this returns a `Spree::Money` instance is to allow for dynamic pricing when persisting the price might be undesireable. For example if you sold stickers that were priced dynamically based on the size that customer inputted and also the volume.

I don't think this is an issue, since you can just return an unpersisted price, but it should be understood that this API is allowed to return unpersisted prices to allow for that use case. 🤔

core/app/models/spree/variant/price_selector.rb
```diff
@@ -24,13 +24,28 @@ def initialize(variant)
...
       # The variant's price, given a set of pricing options
       # @param [Spree::Variant::PricingOptions] price_options Pricing Options to abide by
-      # @return [Spree::Money, nil] The most specific price for this set of pricing options.
-      def price_for(price_options)
-        variant.currently_valid_prices.detect do |price|
+      # @param [Boolean] return_price Flag to return a Spree::Price instead of
+      #   a Spree::Money
+      # @return [Spree::Price, Spree::Money, nil] The most specific price for
+      #   this set of pricing options in the desired format.
+      def price_for(price_options, return_price: false)
```
aldesantis, line 31: I'm a bit concerned by this approach to deprecating the method. Our approach to major upgrades is to ensure that someone who's on the latest 3.x will always be able to update to 4.0 _without touching their code_ as long as they don't get any deprecation warnings.

Unless I'm missing something, with this change people will have to call `price_for(..., return_price: true)` in the latest 3.x in order not to get the deprecation warning. But then, in 4.0, we are going to remove the `return_price:` argument altogether, and anyone who passes it will start getting an error.

In other words, people who upgrade from the latest 3.x to 4.0 will have to change something in their code in order to complete the upgrade, which is something we should try to avoid if at all possible.

How about removing this, and only using the value of `Spree::Config. price_selector_returns_price` instead? It's also more straightforward.

aldesantis (dismissed):

aldesantis:
@swively thanks, I think the change makes sense but the deprecation strategy should probably be adjusted a little bit.

Also, we'll have to address a bunch of deprecation warnings that this change causes in the Solidus codebase instead — the calling code will have to be updated in order to use the `Spree::Money` API instead of `Spree::Price`.

As an alternative, have you also considered simply implementing a new price selector class that returns a `Spree::Money` object, and using that as the default in all new applications?

core/app/models/spree/variant/price_selector.rb
kennyadsl: The idea was to deprecated `return_price` after the change in the next major (4.0). This kind of deprecation allows keeping the system working for example if the host application switched to returning price but some extensions don't.

I'm also ok using `Spree::Config. price_selector_returns_price` alone.

swively:
Having seen a few client projects encounter issues with the price selector returning a `Spree::Money`, the typical solution has been to add a method `price_record_for` that returns the `Spree::Price`. For this reason I think having the return value be configurable by the `return_price` flag is useful.

To avoid forcing developer intervention when bumping to 4.0, I could remove the deprecation altogether and just let this method's return value be controlled via the config/flag, but I also think that the method in its current state is misleading and not as useful as it would be if it just returned a price object.

Happy to go either way on this, the less impactful option being to remove the deprecation and leave it up to the user to configure the method to suit their use case.

aldesantis:
I think just using the global configuration would be better, since it doesn't create more work for us and our users in the future to also remove the new method parameter.

@jarednorman @kennyadsl what do you think? Do you see any drawbacks?

swively:
I've encountered it on two client projects in the last month. The first was using the [sales prices](https://github.com/solidusio-contrib/solidus_sale_prices) extension, which has the sale price for a variant on the `Spree::Price`. Anywhere that you would need to grab the price (using `#price_for(pricing_options)`) would return a `Spree::Money`, not allowing us to reach for the sales price.

On a different project they used partial prices for items that are part of a bundle, so each item would have a regular standalone price, as well as a `partial_price` that was discounted when in a bundle. Again you would ideally grab the price with `#price_for(pricing_options)` to get one that is accurate for the store and currency, but we were limited by not being able to access the `partial_price` for the given variant.

core/app/models/spree/variant/price_selector.rb
jarednorman: Interesting, I'd not really considered this. So @kennyadsl are you thinking we do this:

- In 3.x you get a deprecation warning if you use the `return_price: false` behaviour, but the behaviour remains.
- In 4.x we remove the `return_price: false` behaviour and you get a deprecation warning if you pass `return_price:` to the method.
- In 5.x we remove the `return_price:` argument.

That's not unreasonable, but 5.0 seems so far in the distance. 😅

kennyadsl: That's what I was thinking but, totally unrelated, I think we can change the removal deprecation process starting from Solidus 3.0 so that we are allowed to remove deprecated code in the next minor version. That's what Rails does, and also (I discovered this yesterday 😓) what we kinda state [here](https://guides.solidus.io/developers/upgrades/versioning-guidelines.html):

> We would like to also follow the Rails approach: deprecating functionality in one minor version and removing it in the next.

Anyway, I think @aldesantis has a valid proposal to avoid this parameter at all, we were discussing it yesterday and I guess he will post some comments soon.

aldesantis:
@swively after discussing the deprecation approach for this with @kennyadsl, we decided adding a new method may be the best path for moving forward. The additional parameter doesn't play well with a situation where someone has implemented their custom price selector, without knowing that it needs to comply with the `return_price: true` argument.

You could add a `PriceSelector#price_for_options` method that returns a `Spree::Price` instead of a `Spree::Money`, and update `PriceSelector#price_for` to raise a deprecation warning and simply call `#price_for_options(...).money`.

The only concern is properly handling a situation where someone has implemented their own price selector and has only implemented `#price_for`, but not `#price_for_options`.

To fix that, you might want to also implement `Spree::Variant#price_for_options` (and have a `delegate :price_for_options, to: :find_or_build_master` in `Spree::Product`) and deprecate `Spree::Variant#price_for`. `Spree::Variant#price_for_options` would check if the price selector responds to `#price_for_options`: if it does, it would simply return its value. If not, it would print a deprecation warning, then build an unpersisted `Spree::Money` with the value returned by `#price_for` and return that.

Finally, you would update all the places in Solidus where we call `#price_for` and call `#price_for_options` instead.

Something else to keep in mind: `Spree::Variant#price_for` was implemented as the successor to `Spree::Variant#price_in` and `Spree::Variant#amount_in`, so you need to make sure the deprecation warnings for those two methods tell people to call `#price_for_options` instead. Otherwise, we'll be telling people to solve a deprecation by calling another deprecated method. 😅

kennyadsl:
@swively sorry, my fault for initially pointing you in a path that is not ideal. 🙏

jarednorman:
Ah, that's a good catch. I had not thought of that. I don't love the name `price_for_options`, but I don't hate it either.

aldesantis:
Yeah, to be honest it's the only one we could think of. 🤷 Open to better suggestions.

One thing we should keep in mind is that we're out of names for deprecating that method, so let's make sure we get it right this time. :trollface:

jarednorman:
Yeah, I think in one app I used `spree_price_for` to distinguish it from `price_for` by making it clear it returns a `Spree::Price`, but I don't think we should do that in core.

jarednorman (approved):

kennyadsl (approved):
Thanks, this PR is perfect! 🙏 🙏 

We are going to merge this as soon as 3.0 is released so it will be part of 3.1.

jarednorman (approved):

spaghetticode (approved):
@swively thanks!

PR #3927, opened 2021-02-06
Use Spree::Order.user email format validators on Spree::EmailValidator

cesartalves:
## Description

Background: #3427 

This patch solves a problem where the email for Spree::Order is copied from Spree.user_class, specifically when spree_auth_devise is used.
Without it, `Spree::Order` might be invalid after persistence when the Spree.user_class email does not match the standard `EMAIL_REGEXP` of `Spree::EmailValidator`

It would be logic to use the same email validation for both `Spree.user_class` and `Spree::Order`.

I came to the conclusion there were three variations of possible solutions:

1. use `defined?(Devise)` to check whether we should validate the email format against the old `EMAIL_REGEXP` or against `Devise.email_regexp`. This check could be done on `Spree::EmailValidator`, such as in [this PR](https://github.com/solidusio/solidus/pull/3920), upon defining the email validation on `Spree::Order` or somewhere else. Cons of this approach are referencing Devise on the code base;

2. Use ActiveRecord's reflection to reflect upon the Order's user association and use its :email format validator. The advantage is that this works whether spree_auth_devise or another user gem is used, so **it was the used approach in this PR**. Cons are its a bit more complex than solution 1. and a bit more tricky to write automated tests for, it may also not be compatible with active-modal < 3.0 but this can be fixed by not relying on ActiveRecord::Reflection and doing it "crudely" (namely, :send and :instance_eval);

3. Change default `email_regexp` config on [spree_auth_devise's Devise initializer](https://github.com/solidusio/solidus_auth_devise/blob/master/config/initializers/devise.rb) to use the same regexp as `Spree::EmailValidator`:
```
config.email_regexp = /\A([^@\.]|[^@\.]([^@\s]*)[^@\.])@([^@\s]+\.)+[^@\s]+\z/
```
I really like this approach as it minimizes changes, but it might lead to unsuspecting users of `solidus_auth_devise` having some of their User database invalidated if upgrading the gem (though this might happen only in edge cases)

So, which choice do you guys prefer?

## Checklist
- [x] ~~I have updated Guides and README accordingly to this change (if needed)~~
- [x] ~~I have attached screenshots to this PR for visual changes (if needed)~~

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -129,7 +129,7 @@ module Spree
     before_create :link_by_email
 
     validates :email, presence: true, if: :require_email
-    validates :email, 'spree/email' => true, allow_blank: true
+    validates :email, 'spree/user_email' => true, allow_blank: true
     validates :guest_token, presence: { allow_nil: true }
     validates :number, presence: true, uniqueness: { allow_blank: true, case_sensitive: true }
     validates :store_id, presence: true
--- a/core/lib/spree/core.rb
+++ b/core/lib/spree/core.rb
@@ -90,6 +90,7 @@ require 'spree/core/role_configuration'
 require 'spree/core/state_machines'
 require 'spree/core/stock_configuration'
 require 'spree/core/validators/email'
+require 'spree/core/validators/user_email'
 require 'spree/permission_sets'
 require 'spree/user_class_handle'
 
--- a/core/lib/spree/core/validators/user_email.rb
+++ b/core/lib/spree/core/validators/user_email.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+require_relative 'email'
+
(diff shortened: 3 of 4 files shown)


kennyadsl:
@cesartalves Thanks a lot for working on this, I think this is a good solution but I'd suggest considering another approach, especially if you like number 3: we could use this file: https://github.com/solidusio/solidus_auth_devise/blob/09841c164e1d0e6ab7afbb80494737fc692eb806/lib/generators/solidus/auth/install/templates/config/initializers/devise.rb, which should be copied over the host application at install time. This means that any change to that fill will be only available to new users, old ones that already have that file will not be impacted. What do you think?

cesartalves:
@kennyadsl I think that works great! I will close this PR then and open a new one on solidus_auth_devise ASAP.

PR #3932, opened 2021-02-12
Inline the attachment form for taxon icons

elia:
**Description**

There's no explicit support for multiple taxon attachment anywhere else in the solidus codebase, and the attachment_partial_name was using "paperclip" even for active_storage.

Given attachment adapters are substantially dormant prior to Rails 6.1 we just need to explain the change in the changelog without the burden of an overcomplex deprecation cycle.

Original PR: https://github.com/solidusio/solidus/pull/3308

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/taxons/_form.html.erb
+++ b/backend/app/views/spree/admin/taxons/_form.html.erb
@@ -20,7 +20,24 @@
       </div>
     <% end %>
 
-    <%= render "spree/admin/taxons/attachment_forms/#{f.object.attachment_partial_name}", f: f %>
+    <% if f.object.class.attachment_definitions.size > 1 %>
+      <%= render "spree/admin/taxons/attachment_forms/#{f.object.attachment_partial_name}", f: f %>
+    <% else %>
+      <%= f.field_container :icon do %>
+        <%= f.label :icon %><br>
+        <%= f.file_field :icon %>
+        <% if f.object.icon_present? %>
+          <%= image_tag f.object.icon(:mini) %>
+          <%= link_to t('spree.actions.remove'),
+                      admin_taxonomy_taxon_attachment_path(@taxonomy,
+                                                           @taxon.id,
+                                                           attachment_definition: :icon,
+                                                           authenticity_token: form_authenticity_token),
+                      method: :delete,
+                      class: 'btn btn-sm btn-danger' %>
+        <% end %>
+      <% end %>
+    <% end %>
   </div>
 
   <div class="col-5">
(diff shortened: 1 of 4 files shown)


backend/app/views/spree/admin/taxons/_form.html.erb
```diff
@@ -20,7 +20,20 @@
       </div>
     <% end %>
 
-    <%= render "spree/admin/taxons/attachment_forms/#{f.object.attachment_partial_name}", f: f %>
+    <%= f.field_container :icon do %>
```
softr8, line 23: IMO It still needs to iterate over the attachment definitions, it is pretty common to add more than icon attachment to taxons, hero image, mobile hero image, brand logo, etc

elia: I'm not against that in principle, I just thought that if done it should have its own documentation/guide with complete support. 

On the same note might make sense to completely remove the icon from the taxon and let individual store add it if needed, allowing to manage the whole attachment configuration (e.g. pick sizes, attach a video instead, etc.).

In my experience it's always the case that a store has it's own rules for images and attachments.

@softr8 would any of these alternative work for you?

stem: hi folks

I'm jumping in as we've added a lot of different attachments to our taxon object and do not use the default icon.
So as @elia, I'm in favour of removing the icon from the core model. And maybe move it into an extension if some store use only that icon, but my guess is that it won't be used that much.

aldesantis: I'm open to someone trying to change my mind, but it feels like removing support for multiple taxon attachments, even if not documented/explicitly supported, would effectively constitute a breaking change.

As soon as we release some kind of public API, we _must_ assume that people are using it and we can't just pull the rug from under their feet, no matter how well documented it is, or whether it was introduced intentionally or by mistake.

With that said, I agree that we should probably remove the default "icon" attachment altogether and simply let people define their own attachments — but I also think that should be considered a breaking change.

Curious to hear if anyone from @solidusio/core-team feels differently about this.

elia: I added the usual deprecation/configuration harness 👍

backend/app/views/spree/admin/taxons/_form.html.erb
```diff
@@ -20,7 +20,24 @@
       </div>
     <% end %>
 
-    <%= render "spree/admin/taxons/attachment_forms/#{f.object.attachment_partial_name}", f: f %>
+    <% if Spree::Config.keep_legacy_backend_support_for_multiple_taxon_attachments %>
```
kennyadsl, line 23: I don't really get why the preference is needed. I think deprecating the partial and those methods in the `_attachment` files is enough.

elia: Fixed

kennyadsl (approved):
👍

aldesantis (approved):
Thanks @elia!

jarednorman (approved):

spaghetticode (approved):
@elia looks good! 👍

kennyadsl:
We are going to backport this one since 2.11 is ActiveStorage friendly

Issue #3902, opened 2021-01-19
Product promotion rule input only shows first page of products


AlistairNorman:
The product promotion rule select2 only shows the first page of products regardless of how many products are associated with the rule on the admin promotion edit page. There is no way to see the rest of the products associated with the rule.

**Solidus Version**:
2.9.6

**To Reproduce**

1. Go to /admin/promotions/:promotion_id/edit
2. Add 13 products to a rule
3. See that only 12 products show up

On my store the limit is 12, configured on the product model decorator. This will vary for other configurations.

**Current behavior**
It only shows the first 12 products in the input.

**Expected behavior**
All products associated with the rule are shown in the input.

**Screenshots**
![image](https://user-images.githubusercontent.com/272581/105097257-29856b00-5a5d-11eb-939b-93936eb94c2c.png)

**Desktop (please complete the following information):**
 - OS: MacOS 10.15.7
 - Browser Chrome 
 - Version 87.0.4280.141

**Additional Details**
We have tracked down the issue to the pagination in the admin search controller in the product action.
https://github.com/solidusio/solidus/blob/d51518a3ac75e15b19088afbf6e397f5698abf57/backend/app/controllers/spree/admin/search_controller.rb#L24

adammathys:
@jarednorman I believe this was fixed by #3934.

jarednorman:
Thanks!

PR #3934, opened 2021-02-14
Promotion rule product limit improvements

nirnaeth:
**Description**
  Ref #3902 

Adding a further parameter `show_all` to the `select2` rule when picking products in the admin interface. 
The parameter should act only when the selection is initialized at page load and should not have an impact on the product drop-down.
Now the promotion rule section for products should properly show all products associated with the rule and not just up to a certain limit as described in #3902 .

**Screenshots**
With 8 products associated to the rule:

_Before_
Screenshot 2021-02-14 at 12 41 52

_After_
Screenshot 2021-02-14 at 12 42 14

**Checklist:**

**Credits**
@aldesantis for the lead

### DIFF
--- a/backend/spec/controllers/spree/admin/search_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/search_controller_spec.rb
@@ -87,11 +87,29 @@ describe Spree::Admin::SearchController, type: :controller do
       end
     end
 
-    context 'when idds param is not present' do
+    context 'when ids param is not present' do
       let(:params) { { q: { name_cont: 'jersey' } } }
 
       it_should_behave_like 'product search' do
         let(:expected_products) { [product_one, product_two] }
+      end
+    end
+
+    context 'when all products are requested to be shown' do
+      context 'when a per page limit is set' do
+        let(:params) { { show_all: true, per_page: 1 } }
+
+        it_should_behave_like 'product search' do
+          let(:expected_products) { [product_one, product_two] }
+        end
+      end
+
+      context 'when a per page limit is not set' do
+        let(:params) { { show_all: true } }
+
+        it_should_behave_like 'product search' do
+          let(:expected_products) { [product_one, product_two] }
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


nirnaeth:
let me know if I have to squash or reformulate the commit messages

kennyadsl (commented):
👋 Hey Monica! 

This looks great, thanks. 🙏 Please, can you squash commits into a single one?

nirnaeth:
sure, on it!

nirnaeth:
@kennyadsl do I need to do anything else?

kennyadsl:
Nope, this is good thanks. We will wait for the next release before merging this. Going to add the right labels/milestone so it's clear for everybody. Thanks again! 🙏

backend/spec/controllers/spree/admin/search_controller_spec.rb
```diff
@@ -87,12 +87,30 @@ def starting_letters(string)
...
+    context 'when all products are requested to be shown' do
+      context 'when a per page limit is set' do
+        let(:params) { { show_all: true, per_page: 1 } }
+
+        it_should_behave_like 'product search' do
+          let(:expected_products) { [product_one, product_two] }
+        end
+      end
+
+      context 'when a per page limit is not set' do
```
spaghetticode, line 107: It seems to me that this `context` expectations are the same as the `context` above, only the `params` are slightly different, so it's not evident to me what difference in result the different params should produce... so I'm missing the reason for this second test 😅. Can you help me with that?

kennyadsl: It is basically testing that, no matter which `per_page` you are requesting, when there's the `show_all` option it will return all the records, ignoring the `per_page` param. Makes sense @spaghetticode?

spaghetticode: @kennyadsl sure... thanks!

kennyadsl (approved):
Thanks, Monica!

spaghetticode (approved):
@nirnaeth thank you 👍

PR #3936, opened 2021-02-15
Pre-select current tax category on product form

spaghetticode:
The `tax_category` select on the product form page should pre-select the default category only when the product does not have a tax category yet. If it does, then the current product tax category should be pre-selected, instead of the default one.

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -158,7 +158,7 @@
         <%= f.field_container :tax_category do %>
           <%= f.label :tax_category_id, Spree::TaxCategory.model_name.human %>
           <%= f.field_hint :tax_category, default_tax_category: @default_tax_category&.name %>
-          <%= f.collection_select(:tax_category_id, @tax_categories, :id, :name, { include_blank: t('spree.match_choices.none'), selected: @default_tax_category&.id }, { class: 'custom-select' }) %>
+          <%= f.collection_select(:tax_category_id, @tax_categories, :id, :name, { include_blank: t('spree.match_choices.none') }, { class: 'custom-select' }) %>
           <%= f.error_message_on :tax_category %>
         <% end %>
       </div>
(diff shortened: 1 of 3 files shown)


backend/app/views/spree/admin/products/_form.html.erb
```diff
@@ -158,7 +158,7 @@
         <%= f.field_container :tax_category do %>
           <%= f.label :tax_category_id, Spree::TaxCategory.model_name.human %>
           <%= f.field_hint :tax_category, default_tax_category: @default_tax_category&.name %>
-          <%= f.collection_select(:tax_category_id, @tax_categories, :id, :name, { include_blank: t('spree.match_choices.none'), selected: @default_tax_category&.id }, { class: 'custom-select' }) %>
+          <%= f.collection_select(:tax_category_id, @tax_categories, :id, :name, { include_blank: t('spree.match_choices.none'), selected: f.object.tax_category_id || @default_tax_category&.id }, { class: 'custom-select' }) %>
```
kennyadsl, line 161: What about setting the default on `@product.tax_category` at the controller level?

spaghetticode: sure, that an option. Both solutions look rather equivalent to me. While the view gets simpler (we can entirely remove the `selected` specification here) the controller gains a `before_action` for both edit and new. 

I've updated the code, let me know what you think!

kennyadsl (approved):

backend/app/views/spree/admin/products/_form.html.erb
kennyadsl: I like this way! Let see if others have opinions as well.

jarednorman (approved):

MohammadAli1992:
Hello all,

I cant under stand what you sent here ????

*Mohammad Ali*
*CEO & Founder | YalaShoghol*
Phone:     *0792291616* <0792291616>
Website:  *www.yalashoghol.com* <http://www.yalashoghol.com/>
Email:       *Mohammada@Yalashoghol.com* <Mohammada@Yalashoghol.com>
Address:  *King Hussein Business Park #7 First Floor Yalashoghol*

PR #3938, opened 2021-02-15
Install Active Storage by default on new stores

kennyadsl:
**Description**

We are ready to default to Active Storage and it would be great for Solidus 3.0 to ship with this new default. 

We can also investigate removing paperclip as a dependency of core and re-adding it optionally with the installer in the host application Gemfile.


jarednorman (approved):

aldesantis (approved):
@kennyadsl +1 for removing Paperclip as a core dependency and moving it to the Gemfile instead. Should we do it in this PR or another one?

kennyadsl:
@aldesantis it would be great but, as always, it is more complex than expected. Paperclip is still required when Solidus runs with Rails <= 6.0, so if we remove it from core dependencies we will break backward compatibility for people updating Solidus but not Rails. Also, our CI uses it when running specs against Rails 5.2 and 6.0.

peterberkenbosch:
💯  :shipit:

PR #3940, opened 2021-02-17
Only use #original_message in Api::BaseController#parameter_missing_error if defined

dividedharmony:
While working on https://github.com/solidusio-contrib/solidus_related_products/pull/88 for solidus_related_products, I came across a bug that has caused every branch on that repository to fail its specs. 

Sometimes, for reasons that I don't completely understand, the `ParameterMissing` error class does not define `#original_message`, causing a `NoMethodError` to be raised, which obscures the original execption and returns a 500 status response.

Further background is detailed in #3941 

As error handling should not raise its own obscuring error, use the #try method to obtain #original_message, if defined.

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/base_controller.rb
+++ b/api/app/controllers/spree/api/base_controller.rb
@@ -76,7 +76,8 @@ module Spree
       end
 
       def parameter_missing_error(exception)
-        message = exception.original_message || exception.message
+        # use original_message to remove DidYouMean suggestions, if defined
+        message = exception.try(:original_message) || exception.message
         render json: {
           exception: message,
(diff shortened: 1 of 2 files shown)


kennyadsl:
Last week I was working on trying to fix `solidus_related_products` spec suite and I was literally going mad on this issue. I also couldn't understand why this happens but I'm glad you find a working solution, thanks!

kennyadsl (approved):
Thanks!

jarednorman (approved):

tvdeyen (approved):
Nice work, Larry!

Issue #3924, opened 2021-02-05
Remove deprecated error enumeration


seand7565:
The `validate_source` method in `payment.rb` throws a deprecation warning, and will need to be fixed before Rails 6.2.
The warning:
```
DEPRECATION WARNING: Enumerating ActiveModel::Errors as a hash has been deprecated.
In Rails 6.1, `errors` is an array of Error objects,
therefore it should be accessed by a block with a single block
parameter like this:

person.errors.each do |error|
  attribute = error.attribute
  message = error.message
end

You are passing a block expecting two parameters,
so the old hash behavior is simulated. As this is deprecated,
this will result in an ArgumentError in Rails 6.2.
```

PR #3946, opened 2021-02-19
Remove Deprecation Warning in ActiveModel Errors

Azeem838:
**Description**

This PR is a solution for #3924

Enumerating ActiveModel::Errors as a hash has been deprecated. In Rails 6.1, `errors` is an array of Error objects, therefore it should be accessed by a block with a single block parameter. 

This resulted in deprecation warnings in the payment and product models for the validate_source and validate_master methods respectively.

Since this is deprecated, it will result in an ArgumentError in Rails 6.2.

**Checklist:**


aldesantis (changes requested):
@Azeem838 thanks for the work here!

In order to make the code work with Rails 5.2 and 6.0 as well, you can simply check the current Rails version in an `if/else` branch.

core/app/models/spree/payment.rb
```diff
@@ -165,9 +165,16 @@ def source_actions
 
     def validate_source
       if source && !source.valid?
-        source.errors.each do |field, error|
-          field_name = I18n.t("activerecord.attributes.#{source.class.to_s.underscore}.#{field}")
-          errors.add(I18n.t(source.class.to_s.demodulize.underscore, scope: 'spree'), "#{field_name} #{error}")
+        if Rails.version.to_f >= 6.1
```
kennyadsl, line 168: I think the most idiomatic way of doing this is with:

```ruby
Gem::Requirement.new(">= 6.1").satisfied_by?(Rails.gem_version)
```

kennyadsl (commented):
Thanks, I left an improvement for checking the Rails version requirement.

core/app/models/spree/payment.rb
Azeem838: Thank you @kennyadsl, this is a much better way of checking the Rails version requirement. I have implemented your suggestion in the latest commit.

kennyadsl:
Thanks! Can you please just squash commits into a single one and we are good to go! 🙏

Azeem838:
@aldesantis and @kennyadsl, thank you both for your advice on this PR. Please let me know if there are any other changes that need to be made.

kennyadsl (approved):
Thank you! 🙏

aldesantis (approved):
Awesome stuff, thanks @Azeem838!

kennyadsl:
I backported this to `v2.11` so we remove the warning for users using that Solidus version.

PR #3947, opened 2021-02-19
Add docker-compose development environment

waiting-for-dev:
This allows a quick hands-on approach to contributing to Solidus:

```bash
docker-compose up -d
```

The ruby version in the image can be configured through the
`ruby_version` docker build argument (2.7.2 is used by default):

```bash
docker-compose build --build-arg ruby_version=2.6 app
docker-compose up -d
```

The rails version can be set when the project is booted:

```bash
RAILS_VERSION='~> 5.0' docker-compose up -d
```

Tests can be run setting the corresponding `DB` environment variable:

```bash
docker-compose exec app bin/rspec
docker-compose exec app env DB=postgres bin/rspec
docker-compose exec app env DB=mysql bin/rspec
```

Port `3000` is exposed to allow installing the sandbox application as
usual.

Some noticeable considerations:

- The file `database.yml` in the Dummy app has been changed to
accommodate the docker-compose setup. Taking the occasion it has been
refactor to be more consistent. The changes should not affect any local
setup, but it's worth that some people check their setup keeps working
without issues.

- The simplest configuration forces us to use `root` as the postgres
user. MySQL's docker image creates the `root` user and grants all the
privileges to it. Using the same user (and password) for both engines
simplifies `database.yml` configuration. In case another username was to
be chosen, MySQL wouldn't allow it to create the databases needed to do
the testing. If we decide to change it in the future, we would need to
provision a SQL script to the image on boot time:

```
services:
  mysql:
    # ...
    volumes:
      - ./docker/provision/mysql/init:/docker-entrypoint-initdb.d
```

```sql
-- docker/provision/mysql/init/01_databases.sql
CREATE DATABASE IF NOT EXISTS `solidus_api_test`;
CREATE DATABASE IF NOT EXISTS `solidus_backend_test`;
CREATE DATABASE IF NOT EXISTS `solidus_core_test`;
CREATE DATABASE IF NOT EXISTS `solidus_frontend_test`;

GRANT ALL ON `solidus_api_test`.* TO 'solidus_user'@'%';
GRANT ALL ON `solidus_backend_test`.* TO 'solidus_user'@'%';
GRANT ALL ON `solidus_core_test`.* TO 'solidus_user'@'%';
GRANT ALL ON `solidus_frontend_test`.* TO 'solidus_user'@'%'
```

- Dummy application's `database.yml` file no longer checks whether the
`CI` environment variable is set. Instead, it's CircleCI configuration
the one that sets `DB_USERNAME` to `root`. This is part of the work of
making the file more consistent.

- At first, I explored having two different images, one for each database
engine. But I found the final solution simpler. However, it's another
option to consider.

- The `Gemfile` has been modified to allow mysql2, postgres &
sqlite3/fast_sqlite gems to be installed at the same time. They will
install everything in case the `DB_ALL` environment variable is set (as
we are doing in the docker-compose file). This change should be backward
compatible.

- A new selenium capybara driver has been added to make tests pass
inside the docker container.

**Description**

**Checklist:**

### DIFF
--- a/.dockerdev/Dockerfile
+++ b/.dockerdev/Dockerfile
@@ -1,0 +1,57 @@
+ARG RUBY_VERSION
+FROM ruby:$RUBY_VERSION-slim-buster
+
+ARG PG_VERSION
+ARG MYSQL_VERSION
+ARG NODE_VERSION
+ARG BUNDLER_VERSION
+
+RUN apt-get update -qq \
+  && DEBIAN_FRONTEND=noninteractive apt-get install -yq --no-install-recommends \
+    build-essential \
+    gnupg2 \
+    curl \
+    git \
+    imagemagick \
+    libmariadb-dev \
+    sqlite3 \
+    libsqlite3-dev \
+    chromium \
+    chromium-driver \
+  && rm -rf /var/cache/apt/lists/*
+
+RUN curl -sSL https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
+  && echo 'deb http://apt.postgresql.org/pub/repos/apt/ buster-pgdg main' $PG_VERSION > /etc/apt/sources.list.d/pgdg.list
+
+RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys 8C718D3B5072E1F5 \
+ && echo "deb http://repo.mysql.com/apt/debian/ buster mysql-"$MYSQL_VERSION > /etc/apt/sources.list.d/mysql.list
+
+RUN curl -sSL https://deb.nodesource.com/setup_$NODE_VERSION.x | bash -
+
+RUN apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get -yq dist-upgrade && \
+  DEBIAN_FRONTEND=noninteractive apt-get install -yq --no-install-recommends \
+    libpq-dev \
+    postgresql-client-$PG_VERSION \
+    mysql-client \
+    nodejs \
+  &&  rm -rf /var/lib/apt/lists/*
+
+ENV APP_USER=solidus_user \
+    LANG=C.UTF-8 \
+    BUNDLE_JOBS=4 \
+    BUNDLE_RETRY=3
+ENV GEM_HOME=/home/$APP_USER/gems
+ENV APP_HOME=/home/$APP_USER/app
+ENV PATH=$PATH:$GEM_HOME/bin
+
+RUN useradd -ms /bin/bash $APP_USER
+
+RUN gem update --system \
+  && gem install bundler:$BUNDLER_VERSION \
+  && chown -R $APP_USER:$(id -g $APP_USER) /home/$APP_USER/gems
+
+USER $APP_USER
+
+RUN mkdir -p /home/$APP_USER/history
+
+WORKDIR /home/$APP_USER/app
--- a/backend/spec/spec_helper.rb
+++ b/backend/spec/spec_helper.rb
@@ -54,6 +54,15 @@ Capybara.register_driver :selenium_chrome_headless do |app|
   browser_options.args << '--window-size=1920,1080'
   Capybara::Selenium::Driver.new(app, browser: :chrome, options: browser_options)
 end
+Capybara.register_driver :selenium_chrome_headless_docker_friendly do |app|
+  browser_options = ::Selenium::WebDriver::Chrome::Options.new
+  browser_options.args << '--headless'
+  browser_options.args << '--disable-gpu'
+  # Sandbox cannot be used inside unprivileged Docker container
+  browser_options.args << '--no-sandbox'
+  browser_options.args << '--window-size=1240,1400'
+  Capybara::Selenium::Driver.new(app, browser: :chrome, options: browser_options)
+end
 
 Capybara.javascript_driver = (ENV['CAPYBARA_DRIVER'] || :selenium_chrome_headless).to_sym
 
--- a/core/lib/spree/testing_support/dummy_app/database.yml
+++ b/core/lib/spree/testing_support/dummy_app/database.yml
@@ -1,34 +1,54 @@
-<%
-database_prefix = ENV['LIB_NAME'].presence || "solidus"
-%>
-<% case ENV['DB']
+<% db = case ENV['DB']
+        when 'mysql'
+          'mysql'
+        when 'postgres', 'postgresql'
+          'postgres'
+        when 'sqlite', '', nil
+          'sqlite'
+        else
+          raise "Invalid DB specified: #{ENV['DB']}"
+        end %>
+<% db_host = case db
+             when 'mysql'
+               ENV['DB_MYSQL_HOST'] || ENV['DB_HOST']
+             when 'postgres'
+               ENV['DB_POSTGRES_HOST'] || ENV['DB_HOST']
+             else
+               ENV['DB_HOST']
+             end %>
+<% db_prefix = ENV['LIB_NAME'].presence || "solidus" %>
+<% db_username = ENV['DB_USERNAME'] %>
+<% db_password = ENV['DB_PASSWORD'] %>
+<% case db
 when 'mysql' %>
 test:
   adapter: mysql2
-  <% if ENV['CI'] %>
-  username: root
-  password:
+  database: <%= db_prefix %>_test
+  <% unless db_username.blank? %>
+  username: <%= db_username %>
   <% end %>
-  database: <%= database_prefix %>_test
+  <% unless db_password.blank? %>
+  password: <%= db_password %>
+  <% end %>
+  <% unless db_host.blank? %>
+  host: <%= db_host %>
+  <% end %>
   encoding: utf8
-  <% unless ENV['DB_HOST'].blank? %>
-  host: <%= ENV['DB_HOST'] %>
-  <% end %>
-<% when 'postgres', 'postgresql' %>
-<% db_host = ENV['DB_HOST'] %>
+<% when 'postgres' %>
 test:
   adapter: postgresql
-  database: <%= database_prefix %>_test
-  username: postgres
-  min_messages: warning
-<% unless db_host.blank? %>
+  database: <%= db_prefix %>_test
+  username: <%= db_username.presence || "postgres" %>
+  <% unless db_password.blank? %>
+  password: <%= db_password %>
+  <% end %>
+  <% unless db_host.blank? %>
   host: <%= db_host %>
-<% end %>
-<% when 'sqlite', '', nil %>
+  <% end %>
+  min_messages: warning
+<% when 'sqlite' %>
 test:
   adapter: sqlite3
-  database: db/<%= database_prefix %>_test.sqlite3
+  database: db/<%= db_prefix %>_test.sqlite3
   timeout: 10000
-<% else %>
-  <% raise "Invalid DB specified: #{ENV['DB']}" %>
 <% end %>
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,62 @@
+version: '3.7'
+
+services:
+  mysql:
+    image: mysql:8.0
+    command: --default-authentication-plugin=mysql_native_password
+    environment:
+      MYSQL_ROOT_PASSWORD: password
+    volumes:
+      - mysql:/var/lib/mysql:cached
+
+  postgres:
+    image: postgres:13.2
+    environment:
+      POSTGRES_USER: root
+      POSTGRES_PASSWORD: password
+    volumes:
+      - postgres:/var/lib/postgresql/data:cached
+
+  app:
+    build:
+      context: .dockerdev
+      dockerfile: Dockerfile
+      args:
+        RUBY_VERSION: "2.7.2"
+        PG_VERSION: 13
+        NODE_VERSION: 14
+        MYSQL_VERSION: "8.0"
+        BUNDLER_VERSION: 2
+    image: solidus-3.0.0.rc1
+    command: bash -c "(bundle check || bundle) && tail -f /dev/null"
+    environment:
+      CAPYBARA_DRIVER: selenium_chrome_headless_docker_friendly
+      DB_USERNAME: root
+      DB_PASSWORD: password
+      RAILS_VERSION: ${RAILS_VERSION:-~> 6.1.0}
+      DB_ALL: "1"
+      DB_MYSQL_HOST: mysql
+      DB_POSTGRES_HOST: postgres
+      HISTFILE: "/home/solidus_user/history/bash_history"
+      MYSQL_HISTFILE: "/home/solidus_user/history/mysql_history"
+      RAILS_ENV: development
+    ports:
+      - "${SANDBOX_PORT:-3000}:${SANDBOX_PORT:-3000}"
+    volumes:
+      - .:/home/solidus_user/app:delegated
+      - bundle:/home/solidus_user/gems:cached
+      - history:/home/solidus_user/history:cached
+      - .dockerdev/.psqlrc:/home/solidus_user/.psqlrc:cached
+    tty: true
+    stdin_open: true
+    tmpfs:
+      - /tmp
+    depends_on:
+      - mysql
+      - postgres
+
+volumes:
+  bundle:
+  history:
+  postgres:
+  mysql:
--- a/frontend/spec/spec_helper.rb
+++ b/frontend/spec/spec_helper.rb
@@ -45,6 +45,15 @@ Capybara.default_max_wait_time = ENV['DEFAULT_MAX_WAIT_TIME'].to_f if ENV['DEFAU
 
 require "selenium/webdriver"
 require 'webdrivers'
+Capybara.register_driver :selenium_chrome_headless_docker_friendly do |app|
+  browser_options = ::Selenium::WebDriver::Chrome::Options.new
+  browser_options.args << '--headless'
+  browser_options.args << '--disable-gpu'
+  # Sandbox cannot be used inside unprivileged Docker container
+  browser_options.args << '--no-sandbox'
+  browser_options.args << '--window-size=1240,1400'
+  Capybara::Selenium::Driver.new(app, browser: :chrome, options: browser_options)
+end
 Capybara.javascript_driver = (ENV['CAPYBARA_DRIVER'] || :selenium_chrome_headless).to_sym
 
 ActiveJob::Base.queue_adapter = :test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,7 @@ executors:
       <<: *environment
       DB: mysql
       DB_HOST: 127.0.0.1
+      DB_USERNAME: root
     docker:
       - image: *image
       - image: circleci/mysql:5.7-ram
--- a/.dockerdev/.psqlrc
+++ b/.dockerdev/.psqlrc
@@ -1,0 +1,1 @@
+\set HISTFILE ~/history/psql_history
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ tags
 node_modules
 yarn.lock
 package-lock.json
+.env
--- a/Gemfile
+++ b/Gemfile
@@ -19,12 +19,13 @@ group :backend, :frontend, :core, :api do
   gem 'sprockets', '~> 3'
 
   platforms :ruby do
-    case ENV['DB']
-    when /mysql/
+    if /mysql/.match?(ENV['DB']) || ENV['DB_ALL']
       gem 'mysql2', '~> 0.5.0', require: false
-    when /postgres/
+    end
+    if /postgres/.match?(ENV['DB']) || ENV['DB_ALL']
       gem 'pg', '~> 1.0', require: false
-    else
+    end
+    if ENV['DB_ALL'] || !/mysql|postgres/.match?(ENV['DB'])
       gem 'sqlite3', require: false
       gem 'fast_sqlite', require: false
     end
--- a/README.md
+++ b/README.md
@@ -238,6 +238,8 @@ and/or customizations to the Solidus admin. Use at your own risk.
   cd solidus
   ```
 
+### Without Docker
+
 * Install the gem dependencies
 
   ```bash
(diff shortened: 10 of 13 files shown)


backend/spec/spec_helper.rb
```diff
@@ -54,6 +54,16 @@
   browser_options.args << '--window-size=1920,1080'
   Capybara::Selenium::Driver.new(app, browser: :chrome, options: browser_options)
 end
+Capybara.register_driver :selenium_chrome_headless_docker_friendly do |app|
+  Capybara::Selenium::Driver.load_selenium
```
kennyadsl, line 58: This line should not be required since `selenium/webdriver` is already loaded at line 47. Or there's something else in that method that we could need here? If we need it, it's probably better to call it outside all the `register_driver` definitions?

frontend/spec/spec_helper.rb
```diff
@@ -45,6 +45,16 @@
 
 require "selenium/webdriver"
 require 'webdrivers'
+Capybara.register_driver :selenium_chrome_headless_docker_friendly do |app|
```
kennyadsl, line 48: Do we need to replicate the config here as well?

core/lib/spree/testing_support/dummy_app/database.yml
```diff
@@ -1,33 +1,39 @@
-<%
-database_prefix = ENV['LIB_NAME'].presence || "solidus"
-%>
-<% case ENV['DB']
+<% db = ENV['DB'] %>
+<% db_prefix = ENV['LIB_NAME'].presence || "solidus" %>
+<% db_host = ENV['DB_HOST'] || ENV['DB'] %>
```
kennyadsl, line 3: Why this should be `ENV['DB']` if `ENV['DB_HOST']` is empty?

kennyadsl (commented):
Hey @waiting-for-dev, this is great. I'm not a Docker expert but this seems to be well done, thanks!

I left some preliminary questions + I think we should improve the documentation by adding some notes about Docker setup in Solidus Guides here: https://guides.solidus.io/developers/getting-started/develop-solidus.html.

backend/spec/spec_helper.rb
waiting-for-dev: Good catch. We don't need it. Removed.

frontend/spec/spec_helper.rb
waiting-for-dev: I'm afraid we need it. They belong to two different sub-systems: backend & frontend. In case I remove it from here it will make tests fail when executed from within the `frontend` directory. I haven't tried it, but it's possible that they pass if executed globaly by `bin/rspec` script. However, it would rely on a very brittle source code load order.

It doesn't happen with `:selenium_chrome_headless` because it's a driver already provided out of the box by selenium.

core/lib/spree/testing_support/dummy_app/database.yml
waiting-for-dev: Good question, I had forgotten to mention it in the PR description. This is kind of a trick to make it possible for the docker service to run both MySQL & Postgress suites through the usual `DB` environment variable. If we only use `DB_HOST` then we should force the `app` service to link either to `mysql` or `postgres` services or otherwise we would have to provide both `DB_HOST` & `DB` when running tests. Defaulting to `DB` when `DB_HOST` is not present, makes the docker container match the name of the database service name (which is the same as the name seen for it as a host from app).

waiting-for-dev:
Thanks for your review @kennyadsl .

> I left some preliminary questions + I think we should improve the documentation by adding some notes about Docker setup in Solidus Guides here: https://guides.solidus.io/developers/getting-started/develop-solidus.html.

I updated the guides, thanks for pointing me to them. I have addressed the issues separately in different commits. If it's ok with you I'll squash them together when everything is ready for release.

core/lib/spree/testing_support/dummy_app/database.yml
waiting-for-dev: Thinking about it a second time, we can communicate better if we allow overriding `DB_HOST` through new `DB_MYSQL_HOST` & `DB_POSTGRES_HOST` env vars, respectively for MySQL and Postgres. This way we can define both on docker-compose.yml and we don't need to overuse `DB`. If you don't think otherwise I'll change it in this way.

kennyadsl: The result is not yet super clear to me but if you think it's better let's try, I'll take a look later after the change is made, thanks!

peterberkenbosch:
There are some good pointers here https://evilmartians.com/chronicles/ruby-on-whales-docker-for-ruby-rails-development and https://evilmartians.com/chronicles/reusable-development-containers-with-docker-compose-and-dip that we could implement here too. Keeping the docker image small and running fast is key for adaption I think. Using volumes and the correct layering gives a lot of benefit. Having this available is awesome, thanks for that

waiting-for-dev:
> There are some good pointers here https://evilmartians.com/chronicles/ruby-on-whales-docker-for-ruby-rails-development and https://evilmartians.com/chronicles/reusable-development-containers-with-docker-compose-and-dip that we could implement here too. Keeping the docker image small and running fast is key for adaption I think. Using volumes and the correct layering gives a lot of benefit. Having this available is awesome, thanks for that

Thank you @peterberkenbosch for sharing, that's really good stuff. I think we could adopt some of the ideas, while I'd ignore others. I'll try to go through all of them so we can all agree on what to take. Please, tell me if I overlooked something.

:x:  Using dip. It's an opinionated tool that anyway can be used orthogonally with our setup (but looks really cool, I'll study adopting it myself, so thanks for letting me know about it).

:heavy_check_mark:  Using slim image. SURE!

:x: Specifying Debian distribution. I'm not sure it's a good idea. It adds more maintenance burden for a benefit that's not too clear for me.

:heavy_check_mark: Installing postgres-client in the image. Yeah, makes sense.

:heavy_check_mark: Setting default values for build arguments from docker-compose.

:heavy_check_mark: Lock yarn & node versions.

:heavy_check_mark: Lock bundler version.

:x: Install git. I don't think most people use git from within the container. But more on this later.

:heavy_check_mark: Remove apt records. Makes sense.

:heavy_check_mark: noninteractive apt frontend. Makes sense.

:heavy_check_mark: Don't install apt recommendations. Makes sense.

:heavy_check_mark: & :x: Using the `Aptfile` with dependencies. Having some dependencies declared outside the Dockerfile in this way makes sense only for a team that tries to share a Dockerfile between several projects. However, I think it's a neat trick to allow users to install what they want within the container. So, I'd add `apt-get install $(cat /tmp/Aptfile 2> /dev/null | xargs)` & I'd add `.dockerdev/Aptfile` to `.gitignore`. Then a user can decide to add `git`, `vim` or whatever to the image.

:heavy_check_mark: Use `.dockerdev` contenxt in the Dockerfile. That's great!

:x: Define multiple services based on app. I thought about it but I don't think it makes sense here. I thought about adding a service with the sandbox application, but as it needs to be regenerated every time the code changes it'd be quite useless.

:heavy_check_mark: Tagging the image. Definitely!!!

:heavy_check_mark: `:cached` strategy & volumes to speed up Mac. Great!! I don't use Mac so I wasn't aware of these performance settings. Let's add them!!

:heavy_check_mark: Keep psql history. Cool!

:heavy_check_mark: Keep bash history. Even cooler!

:x: Setting default editor. It's user-dependent and can be configured using the `Aptfile` as we talked about before and the `.env` file. However, I'd add `.env` to `.gitignore`.

:heavy_check_mark: Health checks. I didn't know about it. Makes sense!

core/lib/spree/testing_support/dummy_app/database.yml
waiting-for-dev: Hey @kennyadsl , you can take a look at this commit:

https://github.com/solidusio/solidus/pull/3947/commits/cdf038ddfd3dcd0be2001461a383bfcfd9e27337

peterberkenbosch:
Awesome @waiting-for-dev, this helps a lot for development. 👍 

Also really like your reply and rational for your choices here. Love it.

core/lib/spree/testing_support/dummy_app/database.yml
kennyadsl: Got it, I think it makes sense, even if that variables' definition is becoming too complex in my opinion. Maybe we can abstract that in some external class. Not saying we should do that in this PR, but maybe it's something we can start thinking about, wdyt?

waiting-for-dev: Yes, I completely agree.

waiting-for-dev:
@peterberkenbosch I updated the configuration following some of the tips from the article. Some changes from what I said in the previous comment:

- Of course, I had to add MySQL into the equation.
- I did have to specify the Debian version in order to add APT sources for some manually installed packages.
- No need to use yarn.
- I had to install git so that Gemfile sources can be fetched from Github.
- Limited the clean-up of apt sources to `rm -rf /var/cache/apt/lists/*`. It's what [docker recommends](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get) and I want to keep things simple.
- Not using the `Aptfile` trick. Docker `COPY` command doesn't allow for a file not to exist. There are workarounds but again I want to keep things simple, at least for now. Anyway, it's not a standardized workflow with Docker, so it would be far from evident.
- No healthy checks for DB services. docker-compose [no longer supports `condition`](https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on). There are [workarounds](https://docs.docker.com/compose/startup-order/) but I don't think adding extra complexity pays off here. So I only switched `links` for `depends_on`, which will at least ensure the startup order.

Please, look at https://github.com/solidusio/solidus/pull/3947/commits/737464d5a07980264d38111c177f5d37eeca60f3

.dockerdev/Dockerfile
```diff
@@ -0,0 +1,56 @@
+ARG RUBY_VERSION
+FROM ruby:$RUBY_VERSION-slim-buster
```
blocknotes, line 2: Just to share a thought here: Debian is great, but I like more to create my containers using an Alpine image (ex. `ruby:2.7.2-alpine`) to have the bare minimum OS possible; the drawback is that you'll probably need to install some extra package and to switch from bash to sh where needed - so also staying on Debian is a good choice anyway if you prefer it :)

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+      DB_POSTGRES_HOST: postgres
+      HISTFILE: "/home/solidus_user/history/bash_history"
+      MYSQL_HISTFILE: "/home/solidus_user/history/mysql_history"
+    ports:
+      - "3000:3000"
+    volumes:
+      - .:/home/solidus_user/app:cached
+      - bundle:/home/solidus_user/gems
+      - history:/home/solidus_user/history
+      - .dockerdev/.psqlrc:/home/solidus_user/.psqlrc
```
blocknotes, line 48: About volumes: I find it a good choice to have the project path as `delegated` and the cache paths (`gems` and `node_modules`) as `cached`. Some extra info here: https://tkacz.pro/docker-volumes-cached-vs-delegated/

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+    volumes:
+      - mysql:/var/lib/mysql
+
+  postgres:
+    image: postgres:13.2
+    environment:
+      POSTGRES_USER: root
+      POSTGRES_PASSWORD: password
+    volumes:
+      - postgres:/var/lib/postgresql/data
```
blocknotes, line 18: A curiosity here: is there a special reason to have a named volume for DB data?

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+      - bundle:/home/solidus_user/gems
+      - history:/home/solidus_user/history
+      - .dockerdev/.psqlrc:/home/solidus_user/.psqlrc
+    tty: true
+    stdin_open: true
+    tmpfs:
+      - /tmp
+    depends_on:
+      - mysql
+      - postgres
```
blocknotes, line 55: You could also evaluate adding `user` to set the uid/gid (but this must be tested carefully because it can be tricky to handle all the permissions accordingly): `user: "${UID}:${GID}"`

.dockerdev/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+ && echo "deb http://repo.mysql.com/apt/debian/ buster mysql-"$MYSQL_VERSION > /etc/apt/sources.list.d/mysql.list
+
+RUN curl -sSL https://deb.nodesource.com/setup_$NODE_VERSION.x | bash -
+
+RUN apt-get update -qq && DEBIAN_FRONTEND=noninteractive apt-get -yq dist-upgrade && \
+  DEBIAN_FRONTEND=noninteractive apt-get install -yq --no-install-recommends \
+    libpq-dev \
+    postgresql-client-$PG_VERSION \
+    mysql-client \
+    nodejs \
```
blocknotes, line 36: Do we still `nodejs` package here? In the statement above we have just installed it from nodesource.

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+      context: .dockerdev
+      dockerfile: Dockerfile
+      args:
+        RUBY_VERSION: "2.7.2"
+        PG_VERSION: 13
+        NODE_VERSION: 14
+        MYSQL_VERSION: "8.0"
+        BUNDLER_VERSION: 2
+    image: solidus-3.0.0.rc1
+    command: bash -c "(bundle check || bundle) && tail -f Gemfile"
```
blocknotes, line 31: `tail -f Gemfile`? 🤔 
I wonder the meaning of this line.

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+    environment:
+      CAPYBARA_DRIVER: selenium_chrome_headless_docker_friendly
+      DB_USERNAME: root
+      DB_PASSWORD: password
+      RAILS_VERSION: ${RAILS_VERSION:-~> 6.1.0}
+      DB_ALL: "1"
+      DB_MYSQL_HOST: mysql
+      DB_POSTGRES_HOST: postgres
+      HISTFILE: "/home/solidus_user/history/bash_history"
+      MYSQL_HISTFILE: "/home/solidus_user/history/mysql_history"
```
blocknotes, line 41: What about having the `RAILS_ENV` set?

docker-compose.yml
```diff
@@ -0,0 +1,61 @@
...
+    environment:
+      MYSQL_ROOT_PASSWORD: password
+    volumes:
+      - mysql:/var/lib/mysql
+
+  postgres:
+    image: postgres:13.2
+    environment:
+      POSTGRES_USER: root
+      POSTGRES_PASSWORD: password
```
blocknotes, line 16: In a personal project I'm using `POSTGRES_HOST_AUTH_METHOD: trust` to skip `POSTGRES_PASSWORD` and to keep the default Rails database configuration (but for the user I'm doing the same as you).

blocknotes (commented):
Nice work @waiting-for-dev!
I just added some minor considerations but looks good to me.

.dockerdev/Dockerfile
waiting-for-dev: If it's ok I'd prefer to keep Debian. I know that with alpine it would be lighter, but I think that for most people Debian is a more familiar system (I included ;)). When the time comes to do some dirty work in the container I appreciate it!

docker-compose.yml
waiting-for-dev: Thanks for pointing me to that article. That's interesting stuff I didn't have the necessity to study before as I'm a Linux user. But isn't it the other way round? The project path will have a lot of writing from the host, while the `GEM_PATH` will only be written by the container.

docker-compose.yml
waiting-for-dev: It's so that we can keep the data in case we remove the container.

docker-compose.yml
waiting-for-dev: Sorry, but I don't understand this comment :sweat_smile: Where should I add the user?

.dockerdev/Dockerfile
waiting-for-dev: Good catch, fixed!

docker-compose.yml
waiting-for-dev: It's in order to keep the container alive. If I omit `tail...` the process will exit when bundle ends its job. This way I can run `docker-compose exec app ...` and keep the history. Nothing prevents me to do `docker-compose run --rm app..` if I prefer it.

docker-compose.yml
waiting-for-dev: Do you think it's necessary? I mean, we only need it for the sandbox application, and it will always be in `development` environment, which is already the default. And anyway I can do `docker-compose exec app env RAILS_ENV=test ...` if really needed.

docker-compose.yml
waiting-for-dev: In this setup we need both mysql and postgres to match their credentials to work with the generated `database.yml` file.

waiting-for-dev:
In https://github.com/solidusio/solidus/pull/3947/commits/591f6cc49a8ee59c44d563c9142171cfbbe11ab6 I have adapted the `database.yml` for the dummy application that is generated from extensions to be also configurable from docker (and consistent with the one within `spec` directory).

I have removed the reference to Travis as I think we're not using it anymore. I also have removed a duplicated configuration for sqlite. `postgres` is no longer the default username for Postgres, but as it's already the default in PostgresSQL itself it should be a safe change.

As @kennyadsl said, cleaning and unifying how the `database.yml` files are generated should be part of another PR.

.dockerdev/Dockerfile
waiting-for-dev: Reverted. We do need that line. `curl...` is just updating apt sources and we still need to manually install the package.

docker-compose.yml
waiting-for-dev: hey @blocknotes any extra info for this one? :slightly_smiling_face:

docker-compose.yml
waiting-for-dev: hey @blocknotes , what do you think about it?

blocknotes: From the [docs](https://docs.docker.com/engine/reference/commandline/service_create/#add-bind-mounts-volumes-or-memory-filesystems) (scroll down a bit to the consistency parameter):
_"delegated: The container runtime's view of the mount is authoritative. There may be delays before updates made in a container are visible on the host."_
So delegated is good for the project path because with it we give priority to the host changes. I understand that for a Linux user this changes little but on the Mac side, Docker volume sync slowness is an issue, unfortunately.
Cached is the opposite, so it fits well for cache paths.

docker-compose.yml
blocknotes: It's a minor point @waiting-for-dev - but if you are curious I mean with just a `user` key.

In my setup I have a common block like:
```yml
x-defaults:
  app: &default_app
    user: "${UID}:${GID}"
```
And then:
```yml
services:
  app:
    <<: *default_app
```
In my case I also have a .env with:
```
UID=${UID}
GID=${GID}
```

.dockerdev/Dockerfile
blocknotes: Ah ok, so you'll need it in any case... 👍

docker-compose.yml
blocknotes: Checking your README.md changes it looks like you do it to be able to use `docker-compose exec app ...` without a running app, right?

From my standpoint, this line looks like a workaround to keep the process up 🤔 but I haven't a better solution to offer here to keep the same behavior ATM.
What I see often is just to have `rails s` in `command`, which means that you need always a running app to be able to exec something else...

docker-compose.yml
blocknotes: I haven't a strong opinion here, so if you prefer that way it's fine.
Perhaps having it could be useful just to make it explicit (eventually using an env var with development as fallback).

docker-compose.yml
waiting-for-dev: Understood, makes sense! (however, the article you pointed me takes the opposite as conclusions :see_no_evil: )

Changed!

docker-compose.yml
waiting-for-dev: Ok, I didn't know it was allowed to provide the user through docker-compose. However, we're performing some actions in the `Dockerfile` for a `solidus_user` user. If a developer could provide a different user through `docker-compose.yml` it would be inconsistent. If I'm not wrong, to enforce this consistency we should add a `--build-arg` and the user should be aware to make it match with what it's configured in `docker-compose.yml`. It would complicate things. Please, if you know a solution for this let me know. User management in containers is something that keeps bugging me.

docker-compose.yml
waiting-for-dev: Yeah, you need a live process to be able to `exec` on the container. As we can't just assume that the sandbox application exists, we need that trick.

docker-compose.yml
waiting-for-dev: I think you are right. It's better to be explicit. Just added it.

waiting-for-dev:
I added a new commit that allows configuring the exposed port (defaults to `3000`):

```bash
SANDBOX_PORT=4000 docker-compose up -d
docker-compose exec app rails server --binding 0.0.0.0 --port 4000
```

docker-compose.yml
blocknotes: Ah sorry about the article. I read a nice blog post times ago about the volumes types but it has been passed a while so it's probably not that one 😓

waiting-for-dev: No worries :blush:

docker-compose.yml
blocknotes: Having `tty: true`, another option could be to remove the `command` param.
(but perhaps an ENTRYPOINT is needed in your Dockerfile, not sure)

docker-compose.yml
blocknotes: Yeah, probably a `--build-arg` (or perhaps an exported env var) would be necessary to set a specific user. It seems tricky in this case to use the `user` parameter.

docker-compose.yml
waiting-for-dev: I tried and then the running process becomes `irb`, which I guess is defined as entrypoint by the base ruby image. But doing so we would lose the option to automatically `bundle` when waking up, which is something I like. I can change `tail -f Gemfile` to `irb` if you prefer. Sure it's a more expensive process than `tail`, but I have no strong opinion here.

waiting-for-dev:
I added a new commit that generates the sandbox application taken into account what is configured as host/username/password for the dabase engine.

docker-compose.yml
blocknotes: If we are going to replace the command line with another blocking process I suppose that it could be fine with your solution anyway.
What about adding a RUN in the Dockfile to exec `bundle install` instead?

waiting-for-dev: I used to do that but it's a hassle as you need to rebuild the image every time you add a new gem to the Gemfile. You can still do `bundle` in a container that has some process running on it, but then when you remove the container for some reason and wake up the project again the application won't work until you realize that you have to rebuild again! It breaks with the flow you would use outside of docker.

Another option I've used sometimes is to just `bundle` neither on Dockerfile nor docker-compose, and then require it to be done manually, which is just the expected flow outside of docker. But having it done automatically and being able to have a development environment just after a single `docker-compose up -d` is something that is awesome (and anyway it doesn't prevent you to manually run `bundle` so it doesn't break any expectation).

blocknotes: Ah ok - BTW it's up to you @waiting-for-dev, if you prefer the tail way then it's fine. At the moment I haven't a better solution if we want also the bundle already executed

waiting-for-dev: Cool. I made a change, though. I added a commit changing `tail -f Gemfile` to `tail -f /dev/null`. It's better because it doesn't pollute the logs and it communicates slightly better.

kennyadsl (approved):
I'm not a docker user but this seems legit if we want to be more friendly with who does.  Thanks, @waiting-for-dev!

jarednorman (approved):

waiting-for-dev:
> I'm not a docker user but this seems legit if we want to be more friendly with who does. Thanks, @waiting-for-dev!

Thanks @kennyadsl . I rebased from master and I added a new line to the Dockerfile:

```
ENV PATH=$PATH:$GEM_HOME/bin
```

It adds bin executables to the PATH so that we can, for instance, run `rails`.

PR #3950, opened 2021-02-21
Fix order checkout flow completion with custom steps

### LINKED ISSUE #3926: Removing `confirm` checkout step doesn't allow transition to `complete`
**Solidus Version**:
2.11

**To Reproduce**
1. Start with a vanilla solidus app
2. Create this file:
```
# app/decorators/my_app/spree/order_decorator.rb
module MyApp
  module Spree
    module OrderDecorator
      def self.prepended(base)
        base.remove_checkout_step :confirm
        # base.insert_checkout_step :complete
      end

      ::Spree::Order.prepend self
    end
  end
end
```
3. Try to check out
4. Unable to transition to complete
Screen Shot 2021-02-05 at 4 57 26 PM

_HOWEVER_ if you go back to that file you created and uncomment out the line that adds the `complete` step to the checkout flow, then checking out with the `confirm` step works as normal.

The `complete` step is added by default to any checkout flow that doesn't include it, so it seems unintentional to not be able to complete checkout from any state but `confirm` without also adding a `complete` step to your custom flow. As far as I understand it, you should be able to `complete` from any step, so long as the order is actually able to be completed and there are no steps above the step you are on.

**Current behavior**

Checkout is unable to proceed from payment without the `confirm` step, or without `checkout` specifically being added as a checkout step.

**Expected behavior**
You should be able to complete from the last checkout step without having to manually add a `complete` step

nerfologist:
**Checklist:**

**Description:**

The `Spree::Order` checkout flow [is configurable](https://github.com/solidusio/solidus/blame/master/core/lib/spree/core/state_machines/order.rb#L32-L37) and steps (states) can be added or removed. Unfortunately, removing the `confirm` state was causing issue #3926 since the code for the `complete` event was still taking for granted that the `confirm` state was part of the flow.

This PR fixes the problem by plugging in the last step of the configurable checkout flow instead.

Also, a paragraph explaining how to customize the checkout flow has been added to the `Order` documentation, since the feature is available in the codebase.

Fixes #3926.

### DIFF
--- a/core/lib/spree/core/state_machines/order.rb
+++ b/core/lib/spree/core/state_machines/order.rb
@@ -73,7 +73,7 @@ module Spree
               end
 
               event :complete do
-                transition to: :complete, from: :confirm
+                transition to: :complete, from: klass.checkout_steps.keys.last
               end
 
               if states[:payment]
--- a/core/spec/models/spree/order/state_machine_spec.rb
+++ b/core/spec/models/spree/order/state_machine_spec.rb
@@ -119,4 +119,42 @@ RSpec.describe Spree::Order, type: :model do
       end
     end
   end
+
+  describe "#complete" do
+    context "when the confirm step has been taken out of the checkout flow" do
+      let!(:payment) { create(:payment, state: 'checkout', order: order) }
+
+      before :all do
+        class Spree::Order
+          checkout_flow do
+            go_to_state :address
+            go_to_state :delivery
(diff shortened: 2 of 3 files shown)


kennyadsl (approved):
Nice job, and thanks for adding the Guide section about it.

jarednorman (approved):

spaghetticode (approved):
LGTM, thanks 👍

Issue #3872, opened 2020-12-22
Update Rubocop rules


kennyadsl:
We have quite an old set of Rubocop rules. We should evaluate the impact of updating the rules we have to a new set. Also, we should consider that at the moment Hound is performing the checks and we should be compatible with one of the versions they support.

jarednorman:
In my view, maintaining a set of Rubocop rules is more trouble than it is worth when we could use [standard](https://github.com/testdouble/standard) that allows us to avoid style arguments, avoid maintaining our own configs, and just use the default settings. I'm very much not a fan of Rubocop's defaults and it's not like standard has made all the same choices I would have, but I'm happy to go along with their choices, especially given the thoughtfulness they've put into each of them and how they've responded to community feedback and backed out of a couple controversial changes.

aldesantis:
I agree, it's probably better to switch to standard and stop worrying about coding style altogether rather than trying to come up with our own rules for every situation.

As for Hound, I think we should stop using it. Because of its architecture, it has a number of problems (mainly that you can't use any RuboCop plug-ins it doesn't support, and the way it handles comments on PRs) and there are better solutions out there nowadays. Even just running RuboCop as part of our CI would be a better (and much simpler) solution imo.

tejasbubane:
I added `standardrb` in this project and it reports a ton of offenses:

```
1318 files inspected, 13047 offenses detected, 12974 offenses auto-correctable
```
Including things like `Prefer double-quoted strings`, indentation errors, etc.

Does it makes sense to make all these changes?

jarednorman:
Yeah, our current Rubocop configuration doesn't line up well with Standard's preferences. I'm not sure a wholesale update is feasible, but we're going to have to do something here. Do you happen to know what percentage of those offenses are just quotes? We can configure Standard to prefer single quotes and potentially make a couple other compromises if it make this update less prohibitive.

tejasbubane:
@jarednorman I added standard, disabled a few rubocop rules and fixed some warnings. https://github.com/solidusio/solidus/pull/3951

For some reason I am getting a bunch of hound-ci warnings and CI failures.

jarednorman:
I explained over [here](https://github.com/solidusio/solidus/pull/3951#issuecomment-1158240985) that we're going to just stick with Rubocop for now. We can move forward with auditing the rules we're using and updating them.

PR #3951, opened 2021-02-21
Replace rubocop with standard

tejasbubane:
Closes https://github.com/solidusio/solidus/issues/3872

**Checklist:**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,12 @@ commands:
           paths:
             - vendor/bundle
 
+  lint:
+    steps:
+      - run:
+          name: Standard Lint
+          command: bundle exec standardrb
+
   test:
     steps:
       - run:
@@ -94,6 +100,11 @@ jobs:
           root: workspace
           paths:
             - solidus-version
+
+  lint:
+    executor: base
+    steps:
+      - lint
 
   postgres:
     executor: postgres
@@ -138,6 +149,7 @@ workflows:
           filters:
             branches:
               only: /master|v\d\.\d+/
+      - lint
       - postgres
       - mysql
       - postgres_rails60
--- a/.standard.yml
+++ b/.standard.yml
@@ -1,0 +1,57 @@
+parallel: true
+format: progress
+
+ignore:
+  - './**/*':
+    - Layout/ArrayAlignment
+    - Layout/ArgumentAlignment
+    - Layout/ClosingParenthesisIndentation
+    - Layout/DotPosition
+    - Layout/BeginEndAlignment
+    - Layout/ElseAlignment
+    - Layout/EndAlignment
+    - Layout/ExtraSpacing
+    - Layout/FirstHashElementIndentation
+    - Layout/FirstArrayElementIndentation
+    - Layout/HashAlignment
+    - Layout/IndentationWidth
+    - Layout/MultilineOperationIndentation
+    - Layout/MultilineMethodCallIndentation
+    - Layout/SpaceBeforeBlockBraces
+    - Layout/SpaceInsideHashLiteralBraces
+    - Layout/SpaceAroundOperators
+    - Layout/SpaceInsideParens
+    - Layout/RescueEnsureAlignment
+    - Lint/AmbiguousRegexpLiteral
+    - Lint/AssignmentInCondition
+    - Lint/ConstantDefinitionInBlock
+    - Performance/RegexpMatch
+    - Performance/BigDecimalWithNumericArgument
+    - Performance/ConstantRegexp
+    - Performance/BlockGivenWithExplicitBlock
+    - Performance/DeletePrefix
+    - Standard/SemanticBlocks
+    - Style/ArgumentsForwarding
+    - Style/ConditionalAssignment
+    - Style/HashSyntax
+    - Style/NegatedIf
+    - Style/NegatedWhile
+    - Style/SafeNavigation
+    - Style/Alias
+    - Style/Alias
+    - Style/PercentLiteralDelimiters
+    - Style/RedundantAssignment
+    - Style/StringLiterals
+    - Style/StringLiteralsInInterpolation
+    - Style/TrailingCommaInArguments
+    - Style/TrailingCommaInArrayLiteral
+    - Style/TrailingCommaInHashLiteral
+    - Style/RedundantRegexpEscape
+  - './**/*_spec.rb':
+    - Layout/EmptyLineBetweenDefs
+    - Lint/ConstantDefinitionInBlock
+    - Style/EmptyMethod
+    - Style/SingleLineMethods
+    - Style/Semicolon
+  - './**/*.jbuilder':
+    - Style/LambdaCall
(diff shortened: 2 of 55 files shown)


jarednorman:
While I don't want to be the one pushing for a ton of formatting churn, I'm also concerned we're not going to get a huge amount of value from using the tool if it's configured such that so much of its functionality is disabled. Might need further consideration with @solidusio/core-team.

tejasbubane:
>  I'm also concerned we're not going to get a huge amount of value from using the tool if it's configured such that so much of its functionality is disabled.

I agree and I realized this after completing the task. The whole point of using standard is to not have to configure anything. This more or less defeats the purpose.

aldesantis:
I agree. If we move to Standard, we should probably just fix all the offenses. We're basically going to touch every single file in the codebase, but at least we'll only have to do it once.

What does the rest of @solidusio/core-team think?

tvdeyen:
I like the idea of opinionated, non-config code formatters like rufo or standard. The thing is that every contributor would need to use it and we would still need to lint the code with rubocop, otherwise the rules wouldnt be enforced. Its a tricky situation. Unfortunately the code formatting is not build into the language.

tejasbubane:
@tvdeyen 

>  we would still need to lint the code with rubocop, otherwise the rules wouldnt be enforced

you mean lint with `standard`?

> Its a tricky situation. Unfortunately the code formatting is not build into the language.

`standard` or `rubocop` is almost same as built-in formatters like `gofmt` or `rustfmt` right? Users need to install it, here they do it with `bundle install`. CI needs to run them to check for formatting. Or am I missing something?

tvdeyen:
> @tvdeyen
> 
> > we would still need to lint the code with rubocop, otherwise the rules wouldnt be enforced
> 
> you mean lint with `standard`?

Standard uses rubocop, so yes. If we go with the default rubocop rules of standard (what I encourage us to do, otherwise it would not make sense to use it at all), we would use standard to lint as well. 👍🏻 

> 
> > Its a tricky situation. Unfortunately the code formatting is not build into the language.
> 
> `standard` or `rubocop` is almost same as built-in formatters like `gofmt` or `rustfmt` right? Users need to install it, here they do it with `bundle install`. CI needs to run them to check for formatting. Or am I missing something?

Unfortuantely its not. The people would still need to install it and configure their editor to "format on safe" with the tool that we force on them. They might have it disabled or even worse configured differently.

The idea of automatic code formatters is great, but unless build into the language from the beginning, you are screwed.

Standards

tvdeyen:
Take the double-quotes rule as example. Prettierrb has it changed from the Prettierjs default to `false` (to enforce single quotes), just because it is more common in Ruby code to use single quotes. This violates the whole idea of prettier.

The idea was to not fight over code formatting rules anymore, and use a tool for that. Now people fight over the rules. Its ridiculous.

Imagine the following situation:

A dev has configured its editor to use `prettierrb` to format `.rb` files on safe, now wants to contribute to solidus and is enforced to use double-quotes. Giving them a bad time fighting its own editor just to fulfill our rules. Not very welcoming.

The only unobtrusive thing I could imagine is to automatically run `standard` on the CI as an after commit hook or something.

jarednorman:
On our projects we usually use our autoformatters with a tool like `lint-staged` and `husky` to provide automatic and `git add -p`-friendly commit hooks that autoformat everything seemlessly. That's the ideal, I think.

bogdanvlviv:
It is lovely to see how we use rubocop in rails.

- We disable all default cops https://github.com/rails/rails/blob/main/.rubocop.yml#L10 and only enable those ones that reflect our style guide in the project.
- The rubocop.yml file read great https://github.com/rails/rails/blob/main/.rubocop.yml#L10. By reading it we can learn about the style Rails repo uses.

One example of how rubocop gives a hand while improving the codebase:

We noticed that we dropped support for < Ruby 2.7 so that we could use `filter_map` gem to speed up some lines of the codebase. https://github.com/rails/rails/pull/42053.  With rubocop, we were able to catch all the places we could improve by `filter_map` and make sure that we would use this method where possible instead of  `map+compact`, or something in the future, and never return to this topic again.
https://github.com/rails/rails/commit/bbbc861f717689a0a28f031fe2176d0f8a6b07c7.
We might use rubocop in a similar way in solidusio/solidus. WDYT?

P. S.There are more interesting samples https://github.com/rails/rails/pulls?q=is%3Apr+rubocop+is%3Aclosed+author%3Akamipo

kennyadsl:
@bogdanvlviv I really like the approach proposed. We can gradually introduce rules based on the feedback we provide in PR reviews and use Rubocop as a tool to refactor code over time. 👍

jarednorman:
I more in favour of the set it and forget approach of Standard. Continuing to maintain a rubocop configuration, especially across the extension ecosystem (though I know we could rely on solidus_dev_support) doesn't seem like a good use of maintenance effort to me. That said, if there were some rules we wanted to enable on top of the Standard defaults, like ones that deter people from doing insecure things, I'm open it.

spaghetticode:
I'm not really in favor of basically touching every single file in the codebase for cosmetic reasons (or at least debatable reasons). 

While I personally like standard more than rubocop (but in general I'm not a fan of code linters, at least in the ruby ecosystem), I think the change would be pretty invasive and pollute the git history. And there's no guarantee it would be final, in a few years we may be still here evaluating the new best linter and debating if we should change all the files again.

jarednorman:
The concern about touching every file is fair; it's not ideal, but I don't consider it a dealbreaker if what we get in exchange for a one-off formatting commit is to stop worrying about this style formatting stuff.

I don't agree with your concern about further churn down the road. The problem with maintaining a Rubocop config (across the whole Solidus ecosystem) is that Rubocop continues to add, remove and change rules, which requires us to evaluate them and decide whether to integrate with (and at least keep up with changes to rules we use.) Using Rubocop is like agreeing to debate style forever.

The whole point of Standard (especially now that it has hit v1.0) is that they've done the work to debate the preferences of the Ruby community and have adopted a reasonably conservative set of sensible defaults. If we adopt it, we agree to basically never do this again barring some kind of catastrophe, otherwise I agree it's not worth changing.

While I don't like formatting churn standard'll cause, only one of these tools purports to offer the opportunity to just use the tool's default settings and never worry about this again.

jarednorman:
And to be clear, I'm still not 100% sure we should do this, I just really don't want to maintain a Rubocop config. It just seems like a waste of time when there's a "set it and forget it" alternative.

tvdeyen:
@jarednorman 💯 agree. Well said.

aldesantis:
I also agree with @jarednorman. Going with Standard is not just another coding style change - it means not having to worry about coding style ever again, which is a very alluring idea.

jarednorman:
@tejasbubane Thanks for your patience with our back and forth on this issue.

After a Core Team discussion we've decided that _for now_ this change would be a bit too disruptive. It would mean everyone would have to rebase all their open PRs, we'd have to update the whole extension ecosystem to match, and it would generate a lot of noise in the history of all the files in the project.

I think we'd all prefer to be using Standard (and without a bunch of rules disabled), but the benefits don't outweigh the disruption. We think there might be a moment down the road where some other large, disruptive change gets made where we'd sneak this one in too just to get it all over with, so we'll cross that bridge when we come to it.

Issue #3903, opened 2021-01-20
Some currencies formatting changed using the last Money gem version


kennyadsl:
We have a failing spec in master due to a change in the last minor version of the Money gem (v6.14.0). I already opened an issue on their project here: https://github.com/RubyMoney/money/issues/966, trying to describe what changed but I'm not sure it will be considered as a breaking change and reverted, at least partially.

The failing spec on our side is due to a comparison that we make between the formatting provided by Money and another JS formatting library that we use ([accounting.js](http://openexchangerates.github.io/accounting.js/)). This comparison is showing clearly that some currency formatting changed.

PR #3958, opened 2021-02-25
Relax Money dependency in development

kennyadsl:
**Description**

Ref #3903

The CHF currency issue we were having has been fixed and a new version (1.14.1) that contains [the fix](https://github.com/RubyMoney/money/pull/967) has been released. CI should be green now. 

We can backport this commit to v2.11 when merged.

**Checklist:**

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,10 +18,6 @@ group :backend, :frontend, :core, :api do
   # and https://github.com/rails/sprockets-rails/issues/369
   gem 'sprockets', '~> 3'
 
-  # Temporary locking money to 6.13.8.
-  # See https://github.com/solidusio/solidus/issues/3903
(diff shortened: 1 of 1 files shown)


jarednorman:
Maybe not necessary, but should we consider locking it to _not_ the versions we know don't work?

kennyadsl:
I'd avoid that Jared, it's a patch-level fix so it's very unlikely that people have Money locked at that specific version. Or are you thinking at a specific scenario where this could cause issues?

jarednorman:
Nah, if it's just one patch release that's affected then I agree, let's not worry about it.

spaghetticode (approved):
@kennyadsl thanks 👍

aldesantis (approved):

jarednorman (approved):

PR #3960, opened 2021-02-26
Adds a Generic NumberGenerator

matheusteixeira:
**Description**

This PR implements the changes suggested in Discussion https://github.com/solidusio/solidus/discussions/3939

Changes:
- Converts `Spree::Order::NumberGenerator` into a more generic `Spree::Core::NumberGenerator`
- Changes the `make_permalink` method of Spree::Core::Permalinks to also use this `NumberGenerator`
- Changes in the models:
  - CustomerReturn
  - ReturnAuthorization
  - Shipment
  - Carton

**Checklist:**


jarednorman (changes requested):
Overall strategy seems good, but we can't just nuke the existing class. We'll have to deprecate it and then remove in a later release. We can just change it so it calls out to the new class though.

matheusteixeira:
@jarednorman Thanks for the heads up. Do you think it would be a good strategy to deprecated using ActiveSupport::Deprecation?

Something like:
```ruby
# lib/order/deprecations.rb
require 'active_support/deprecation'

module Order
  NumberGeneratorDeprecation = ActiveSupport::Deprecation.new(DEPRECATION_HORIZON, 'NumberGenerator')
end

# order/numer_generator.rb
Order::NumberGenerator = ActiveSupport::Deprecation::DeprecatedConstantProxy.new('Order::NumberGenerator', 'Core::NumberGenerator', Order::NumberGeneratorDeprecation)

```

If so, what version do you consider it would be a good deprecation horizon?

jarednorman:
I'm not sure. We've previously been removing things in major releases, recently @kennyadsl pointed out that our policy allows us to do it in minor releases too, so I think the @solidusio/core-team team needs to sync up on how aggressively we want to deprecate and remove things and what our official strategy is for that.

kennyadsl:
@matheusteixeira I think you can use `Spree::Deprecation`, which inherits from the `ActiveSupport` one. We'll set the deprecation horizon at that level, so that we can avoid thinking about it completely in each PR.

jarednorman:
If you use our deprecation, then yeah, your approach looks good. I'll try to give this a thorough review pass once you've tweaked that.

matheusteixeira:
Hi guys, I ended up removing the spec I had in place to test the Deprecation, I tried a few approaches (I kept the commits, so you can maybe give me some insight) that were successful and running locally but would fail to run on CI. Example that pass locally:

```ruby
# core/spec/models/spree/order/number_generator_spec.rb

# frozen_string_literal: true

require 'rails_helper'

RSpec.describe Spree::Order::NumberGenerator do
  describe '#generate' do
    before do
      Spree::Deprecation.behavior = behavior
    end

    context 'when is called on a depracated class without raising' do
      let(:behavior) { :silence }

      it 'calls the method on the new class' do
        expect_any_instance_of(Spree::Core::NumberGenerator).to receive(:generate)
        described_class.new.generate
      end
    end

    context 'when is called on a depracated class with raising behavior' do
      let(:behavior) { :raise }

      it 'raises DeprecationException' do
        expect { described_class.new }.to raise_error(ActiveSupport::DeprecationException)
      end
    end
  end
end
```

I think this is due to following configuration: 
```ruby

  if ENV['SOLIDUS_RAISE_DEPRECATIONS'].present?
    Spree::Deprecation.behavior = :raise
  end
```

I couldn't find a feasible solution at the moment and didn't find other examples of similar code being tested on the project as a whole, so I ended opting for not having the specific test.

kennyadsl:
@matheusteixeira Hey, that `behavior` is set to `raise` in the CI because we want to be sure we are never using deprecated code in core. Of course, in the specs it requires some extra configuration. I think you can rewrite that code as something similar to:

```ruby
# core/spec/models/spree/order/number_generator_spec.rb

# frozen_string_literal: true

require 'rails_helper'

RSpec.describe Spree::Order::NumberGenerator do
  describe '#generate' do
    context 'when is called on a deprecated class' do
      it 'calls the method on the new class emitting a deprecation warning' do
        expect(Spree::Deprecation).to receive(:warn)
        expect_any_instance_of(Spree::Core::NumberGenerator).to receive(:generate)
        described_class.new.generate
      end
    end
  end
end
```

matheusteixeira:
@kennyadsl I had tried a similiar approach before, but even testing this way is still raises and the tests faild the moment it tries to call the deprecated object. I found it weird because it didn't work with the `expect { .. }.to raise` blocks either.

Here is the result with the `Spree::Deprecation.behavior = :raise` config and the example you suggested:

image

PR #3961, opened 2021-02-26
Don't hack into ActionMailer to add our mail previews path

elia:
**Description**

Instead just append our AM preview path let the app control wether they're enabled or not, without forcing it only on some environments.

From the Rails guides of the versions we support:
- https://guides.rubyonrails.org/v6.1/action_mailer_basics.html#previewing-emails
- https://guides.rubyonrails.org/v6.0/action_mailer_basics.html#previewing-emails
- https://guides.rubyonrails.org/v5.2/action_mailer_basics.html#previewing-emails

The path is used inside a glob, that's why we use the `{path1,path2}` notation to add the path to the existing one.

**Checklist:**

### DIFF
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -55,18 +55,13 @@ module Spree
         end
       end
 
-      config.after_initialize do
-        # Load in mailer previews for apps to use in development.
-        # We need to make sure we call `Preview.all` before requiring our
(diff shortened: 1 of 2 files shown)


aldesantis (approved):
@elia thanks! Shame on me for not thinking about this when I implemented the original fix.

jarednorman (approved):

PR #3962, opened 2021-02-26
Add and start populating `spree_addresses.name` field

spaghetticode:
**Description**

In order to prepare apps for the Solidus 3.0 migration where `firstname` and `lastname` fields are completely removed from the DB, this PR introduces the field `name` and starts using it when creating new records, so backfilling will be required only for historical data prior to incorporating this change.

Also, this PR fixes a possible issue when comparing new addresses with double first names to historical records when the preference `use_combined_first_and_last_name_in_address` is set to true.

**Checklist:**

### DIFF
--- a/core/db/migrate/20210122110141_add_name_to_spree_addresses.rb
+++ b/core/db/migrate/20210122110141_add_name_to_spree_addresses.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class AddNameToSpreeAddresses < ActiveRecord::Migration[5.2]
+  def up
+    add_column :spree_addresses, :name, :string
+    add_index :spree_addresses, :name
+  end
+
+  def down
+    remove_index :spree_addresses, :name
+    remove_column :spree_addresses, :name
+  end
+end
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -92,6 +92,12 @@ module Spree
       if base['lastname'].presence || base['last_name'].presence
         base['lastname'] = name_from_attributes.last_name
       end
+
+      virtual_name = name_from_attributes.value
+      if base['name'].blank? && virtual_name.present?
+        base['name'] = virtual_name
+      end
+
       excluded_attributes = DB_ONLY_ATTRS + %w(first_name last_name)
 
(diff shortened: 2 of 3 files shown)


kennyadsl:
@spaghetticode thanks! I tried this locally in a store with existing addresses in the old format and seems to work well, I just found one unrelated problem (already present in v2.11.4) that I documented [here](https://github.com/solidusio/solidus/issues/3966).

kennyadsl (approved):

aldesantis (approved):
I think this looks good, thanks @spaghetticode!

core/db/migrate/20210122110141_add_name_to_spree_addresses.rb
```diff
@@ -0,0 +1,13 @@
...
+  def up
+    add_column :spree_addresses, :name, :string
+    add_index :spree_addresses, :name
+  end
+
+  def down
+    remove_index :spree_addresses, :name
+    remove_column :spree_addresses, :name
+  end
+end
```
schmijos, line 13: Does it still count as a patch update if I need to run `bin/rails railties:install:migrations`?

kennyadsl: Hey! Yes, unfortunately we had to do this change at a patch level because it is fixing the migration process for switching to the new addresses fields schema. We decided to consider this a patch instead of a new feature for this reason. Theoretically, a new migration doesn't always represent a new feature, although it's strange to have. We are sorry if this is causing issues and we are open to improving the documentation to help users upgrading. Let us know your thoughts.

schmijos: I think having a more detailed [CHANGELOG](https://github.com/solidusio/solidus/blob/master/CHANGELOG.md#solidus-2110-2020-10-23) could have helped me to find the issue.

kennyadsl: You are right. I'm preparing a PR that updates the Changelog [here](https://github.com/solidusio/solidus/pull/3975). Anyway, you can find details of each release in https://github.com/solidusio/solidus/releases.

Issue #3967, opened 2021-03-01
Add ActiveStorage setup to the dummy app generated to test extensions


kennyadsl:
More info at https://github.com/nebulab/solidus_starter_frontend/pull/148#issuecomment-787852628.

PR #3969, opened 2021-03-02
Add Active Storage in Dummy App for extensions

kennyadsl:
**Description**

This PR makes sure Active Storage is properly configured when we are generating dummy apps for extensions. This is done in the extension root with:

```ruby
bin/rake extension:test_app
```

Ref #3967

I tested this PR with solidus_starter_frontend that uses Active Storage in some system specs. See https://github.com/nebulab/solidus_starter_frontend/pull/145

No need to backport this one, it will only affect extensions testing against Solidus 3.0, which will default to using Active Storage.

Needs #3970 to be merged first.


aldesantis (approved):

jarednorman (approved):
Makes sense to me.

PR #3979, opened 2021-03-05
Improve Customizing Attributes documentation

dhughesbc:
The documentation about how to add custom attributes to a model in not accurate. This PR updates the documentation

### DIFF
--- a/guides/source/developers/customizations/customizing-attributes.html.md
+++ b/guides/source/developers/customizations/customizing-attributes.html.md
@@ -73,11 +73,12 @@ field.
 
 ## Allowing Custom Attributes to be Saved
 
-If you try to submit the form with the new attribute field, you will notice
-it will not save to the database. This is due to every model having a list of
-permitted attributes.
+If you try to submit the form with the new attribute field, you may notice
+it will not save to the database. This is due to many models having a list of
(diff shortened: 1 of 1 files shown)


kennyadsl (approved):
Thanks! Which is the model you are not seeing in the list?

dhughesbc:
Spree::Role

kennyadsl:
Thanks, just to be clear about why it is missing, `Spree::Role` doesn't have a UI in the admin currently, so there's no admin controller for that resource. I think the guide covers adding attributes to the existing UI provided by Solidus. If you are adding your own views/controllers you already have access to the new controller added and you can simply set the permitted attributes in that controller. Or maybe you are trying to do something else and I am missing something?

Anyway, I think this change is still an improvement and we should add it to the documentation.

dhughesbc:
Thanks, we are using the solidus_user_roles extension, which provides a controller and views for roles management. It was a less experienced (with Solidus) colleague of mine that got tripped up by this Solidus documentation. He added a custom attribute to the Spree::Role model and was following this documentation on how to it, but he did not realize the solidus_user_roles extension does not use the core Spree::PermittedAttributes module.

PR #3982, opened 2021-03-09
Improve address name migration task output

kennyadsl:
**Description**

It now has a similar output to the other task provided to upgrade to v2.11, and improves some messages.

**Before**

```
> bin/rails solidus:upgrade:two_point_eleven
Migrating default billing addresses to address book ... Success
Your DB contains 6 addresses that may be affected by this task.
do you want to proceed? [Y, N] Y
Data migration will happen in batches. The default value is 100_000, which should take less than 20 seconds on mysql or postgresql.
Please enter a different batch size, or press return to confirm the default
We're going to migrate 6 records in batches of 100000
do you want to proceed? [Y, N] Y
Updating 6 records in batches of 100000, this may take a while...
batch took 0 seconds
Your Solidus install is ready for Solidus 2.11
```

**After**

```
> bin/rails solidus:upgrade:two_point_eleven
Migrating default billing addresses to address book ... Success
Combining addresses' firstname and lastname into name ...
  Your DB contains 6 addresses that may be affected by this task.
  Data migration will happen in batches. The default value is 100_000, which should take less than 20 seconds on mysql or postgresql.
  Please enter a different batch size, or press return to confirm the default:  1
  We're going to migrate 6 records in 6 batches of 1.
  Do you want to proceed? [Y, N] Y
  Updating 6 records in batches of 1, this may take a while...
  Batch 1/6 done in 0 seconds.
  Batch 2/6 done in 0 seconds.
  Batch 3/6 done in 0 seconds.
  Batch 4/6 done in 0 seconds.
  Batch 5/6 done in 0 seconds.
  Batch 6/6 done in 0 seconds.
Your Solidus install is ready for Solidus 2.11
```

This PR is also removing the output produced by the task when running specs. We won't see deprecation warnings this way but I think we are covered since we have `behavior = :raise` when tests are run in the CI. 

**Checklist:**

### DIFF
--- a/core/lib/tasks/migrations/migrate_address_names.rake
+++ b/core/lib/tasks/migrations/migrate_address_names.rake
@@ -9,15 +9,17 @@ namespace :solidus do
         concatenation in order to retain historical data when upgrading to new
         address name format'
       task up: :environment do
+        puts "Combining addresses' firstname and lastname into name ... "
         class Spree::AddressForMigration < ApplicationRecord
           self.table_name = 'spree_addresses'
         end
 
-        count = Spree::AddressForMigration.unscoped.where(name: [nil, '']).count
+        records = Spree::AddressForMigration.unscoped.where(name: [nil, ''])
+        count = records.count
         connection = ActiveRecord::Base.connection
         adapter_name = connection.adapter_name.downcase
         shell = Thor::Shell::Basic.new
-        puts "Your DB contains #{count} addresses that may be affected by this task."
+        puts "  Your DB contains #{count} addresses that may be affected by this task."
         # `trim` is not needed for pg or mysql when using `concat_ws`:
         # select concat_ws('joinstring', 'foo', null);
         #  concat_ws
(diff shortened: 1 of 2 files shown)


spaghetticode (approved):
@kennyadsl thanks!

core/lib/tasks/migrations/migrate_address_names.rake
```diff
@@ -37,115 +38,115 @@ namespace :solidus do
...
+          answer = shell.ask('  Do you want to proceed?', limited_to: ['Y', 'N'], case_insensitive: true)
+          if answer == 'Y'
+            puts "  Updating #{count} records in batches of #{size}, this may take a while..."
 
-              Spree::AddressForMigration.unscoped.where(name: [nil, '']).in_batches(of: size).each do |batch|
-                now = Time.zone.now
-                batch.update_all(concat_statement)
-                puts "batch took #{(Time.zone.now - now).to_i} seconds"
-              end
+            Spree::AddressForMigration.unscoped.where(name: [nil, '']).in_batches(of: size).each_with_index do |batch, index|
```
aldesantis, line 143: Hot tip: you can use `each.with_index(1)` (note the dot) instead of `each_with_index` to make the index start from 1.

core/lib/tasks/migrations/migrate_address_names.rake
```diff
@@ -37,115 +38,115 @@ namespace :solidus do
...
+              now = Time.zone.now
+              batch.update_all(concat_statement)
+              puts "  Batch ##{index + 1} done in #{(Time.zone.now - now).to_i} seconds."
```
aldesantis, line 146: Would it be helpful to also include the total number of batches here (i.e., Batch 1/50)? I'm not sure if that's easy to do with ActiveRecord, but it would be a nice addition.

aldesantis (approved):
@kennyadsl I left a couple of comments about possible improvements, but they're both non-blocking so I'm approving anyway!

PR #3945, opened 2021-02-18
Add a post install/update message

kennyadsl:
**Description**

This is particularly useful now that we require users to run a rake task to complete the data migrations needed to update Solidus to v2.11.

This PR is for v2.11 and needs to be replicated in master for Solidus 3.0

**Checklist:**
- [ ] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [ ] I have added a detailed description into each commit message
- [ ] I have updated Guides and README accordingly to this change (if needed)
- [ ] I have added tests to cover this change (if needed)
- [ ] I have attached screenshots to this PR for visual changes (if needed)

PR #3985, opened 2021-03-10
Add Post-Install message to Solidus 3.0

kennyadsl:
**Description**

Similarly to what we did in the `v2.11` branch with #3945, we can add a gem post-install message that will be shown when `solidus-core` will be installed or updated. We used `solidus-core` so that the message will be visible to people using only some submodules of Solidus, instead of the whole wrapper gem. 

This PR also adds a generic `solidus:upgrade` task that will be installed in host applications. 

Even if we don't have any specific rake task here in 3.0 we can keep this convention so we always have a place where to
put this kind of tasks.

Also, we can be consistent between versions in terms of how we request people to upgrade Solidus:

1. Change Gemfile
2. Run `bin/rails solidus:upgrade`


spaghetticode (approved):
@kennyadsl thanks 👍

jarednorman (approved):

PR #3987, opened 2021-03-12
Do not allow prices with nil amount

waiting-for-dev:
**Description**

This overtolerance was introduced in 58ec5cd5. That commit introduced
the `Spree::Price` model to allow dealing with multiple currencies. It
allowed `amount` to be `nil` for master variants, variants with no
master variant, and variants with a master variant with no price. AFAIK,
these three scenarios are no considered by our business logic.

In fact, the logic was removed 5 days later in 9d563e3.

58ec5cd5 also allowed for `amount` to be `nil` in any other case, but
then it copied it from the master variant:

https://github.com/solidusio/solidus/commit/9d563e38ee4ac3f6d3f656f204bd1cc84348bb4e#diff-c083f45179b62dc8c47e4fdcbf7ee31da80552100fc0258238817086e46e2925L35

However, this was also done in the controller layer and it's still in
place:

https://github.com/solidusio/solidus/blob/3cc2d3e492b0aa6c84ac66814db26bea4f17d2e0/backend/app/controllers/spree/admin/variants_controller.rb#L19

Please, tell me if you'd prefer a less aggressive migration process.

**Checklist:**

### DIFF
--- a/core/lib/tasks/migrations/delete_prices_with_nul_amount.rake
+++ b/core/lib/tasks/migrations/delete_prices_with_nul_amount.rake
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+namespace :solidus do
+  namespace :migrations do
+    namespace :delete_prices_with_nul_amount do
+      task up: :environment do
+        print "Deleting prices wich amount attribute is nil ... "
+        Spree::Price.where(amount: nil).delete_all
+        puts "Success"
+      end
+    end
+  end
+end
--- a/core/lib/tasks/upgrade.rake
+++ b/core/lib/tasks/upgrade.rake
@@ -12,7 +12,7 @@ namespace :solidus do
 
     desc "Upgrade Solidus to version 3.0"
     task three_point_zero: [
-        'solidus:migrations:delete_prices_with_nul_amount:up',
+        'solidus:migrations:delete_prices_with_nil_amount:up',
       ] do
       puts "Your Solidus install is ready for Solidus 3.0"
     end
--- a/core/lib/tasks/upgrade.rake
+++ b/core/lib/tasks/upgrade.rake
@@ -9,5 +9,12 @@ namespace :solidus do
       ] do
       puts "Your Solidus install is ready for Solidus 2.11"
     end
+
+    desc "Upgrade Solidus to version 3.0"
+    task three_point_zero: [
+        'solidus:migrations:delete_prices_with_nul_amount:up',
+      ] do
+      puts "Your Solidus install is ready for Solidus 3.0"
+    end
   end
 end
--- a/core/spec/lib/tasks/migrations/delete_prices_with_nil_amount_spec.rb
+++ b/core/spec/lib/tasks/migrations/delete_prices_with_nil_amount_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+path = Spree::Core::Engine.root.join('lib/tasks/migrations/delete_prices_with_nul_amount.rake')
+
+RSpec.describe 'solidus:migrations:delete_prices_with_nul_amount' do
+  describe 'up' do
+    include_context(
+      'rake',
+      task_path: path,
+      task_name: 'solidus:migrations:delete_prices_with_nul_amount:up'
+    )
+
+    it 'removes all prices which amount column is NULL' do
+      price = create(:price)
+      expect(Spree::Price).to receive(:where).with(amount: nil).and_return(Spree::Price.where(id: price))
+
+      task.invoke
+
+      expect { price.reload }.to raise_error(ActiveRecord::RecordNotFound)
+    end
+  end
+end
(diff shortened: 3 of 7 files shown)


core/spec/models/spree/price_spec.rb
```diff
@@ -14,11 +14,6 @@
     let(:variant) { stub_model Spree::Variant }
     subject { Spree::Price.new variant: variant, amount: amount }
 
-    context 'when the amount is nil' do
```
kennyadsl, line 17: Why not keeping this spec and invert the expectation instead?

kennyadsl (commented):
@waiting-for-dev great work, I think it makes sense. I'd move the data migration into a rake task ([something similar to what we did for 2.11](https://github.com/solidusio/solidus/blob/v2.11/core/lib/tasks/upgrade.rake#L6)). 

Please also take a look at the specs against older versions of Rails in the CI.

core/spec/models/spree/price_spec.rb
waiting-for-dev: Fair enough... Done!

waiting-for-dev:
> @waiting-for-dev great work, I think it makes sense. I'd move the data migration into a rake task ([something similar to what we did for 2.11](https://github.com/solidusio/solidus/blob/v2.11/core/lib/tasks/upgrade.rake#L6)).

Yeah, I agree it's a better solution. I did it in a task named matching 3.0 version. Please, tell me if I should address another version.
 
> Please also take a look at the specs against older versions of Rails in the CI.

Hmm.. all the failures reference a `spree_user_key` column which seems not related to the work here... Do you know what is it about?

screenshot-app circleci com-2021 03 12-18_01_56

kennyadsl:
@waiting-for-dev I think the CI errors are due to the way migration has been defined: you are using `ActiveRecord::Migration[6.1]` as the parent class and this is preventing it to work in old versions. Please take a look at the other migrations as a reference.

waiting-for-dev:
> @waiting-for-dev I think the CI errors are due to the way migration has been defined: you are using `ActiveRecord::Migration[6.1]` as the parent class and this is preventing it to work in old versions. Please take a look at the other migrations as a reference.

Thanks @kennyadsl , it seems everything is fine now :+1:

core/lib/tasks/migrations/delete_prices_with_nul_amount.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+namespace :solidus do
+  namespace :migrations do
+    namespace :delete_prices_with_nul_amount do
```
spaghetticode, line 5: Very small typo, here it should be either `null` (one `l` is missing) or `nil`. For consistency with the rest of the occurrences in this file, I'd personally go with `nil`. Also, the filename should change accordingly.

core/lib/tasks/upgrade.rake
```diff
@@ -9,5 +9,12 @@ namespace :solidus do
       ] do
       puts "Your Solidus install is ready for Solidus 2.11"
     end
+
+    desc "Upgrade Solidus to version 3.0"
+    task three_point_zero: [
+        'solidus:migrations:delete_prices_with_nul_amount:up',
```
spaghetticode, line 15: if you change the name of the task, the change should be reflected here.

core/spec/lib/tasks/migrations/delete_prices_with_nil_amount_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+path = Spree::Core::Engine.root.join('lib/tasks/migrations/delete_prices_with_nul_amount.rake')
```
spaghetticode, line 5: if you change the name of the task, the change should be reflected in this file.

spaghetticode (commented):
@waiting-for-dev thanks for this change, I think it makes total sense. I left a few comments on renaming the rake task, I believe there's a small typo in the name that may be worth addressing.

waiting-for-dev:
> @waiting-for-dev thanks for this change, I think it makes total sense. I left a few comments on renaming the rake task, I believe there's a small typo in the name that may be worth addressing.

Thanks for reviewing it @spaghetticode The typo is fixed now :+1:

spaghetticode:
@waiting-for-dev thank you! 

I think the spec file needs renaming: ```core/spec/lib/tasks/migrations/delete_prices_with_nul_amount_spec.rb -> core/spec/lib/tasks/migrations/delete_prices_with_nil_amount_spec.rb```

waiting-for-dev:
> I think the spec file needs renaming: `core/spec/lib/tasks/migrations/delete_prices_with_nul_amount_spec.rb -> core/spec/lib/tasks/migrations/delete_prices_with_nil_amount_spec.rb`

:facepalm: thanks @spaghetticode , fixed :see_no_evil:

gmacdougall (approved):
This makes sense. I don't think we should have prices with nil amounts.

jarednorman (approved):

spaghetticode (approved):
Thanks @waiting-for-dev 👍

PR #3998, opened 2021-03-17
Allow accessing preferences on models that do not have any set

kennyadsl:
**Description**

We may have models that are supposed to have preferences but are not defining them explicitly because they are not needed.

For example, when defining a custom calculator that does not need any preference. Core code expects that preferences still responds with a Hash instead of nil because that's how it worked before b947015bc28c.

That commit is needed because otherwise Rails would serialize the object differently on models that do not use preferences, because seralize is now lazy executed. Example of the wrong serialization without it:

       expected: #<Spree::Address id: 4, firstname: "John", lastname: "Von Doe", address1: "10 Lovely Street", address...00 +0000", updated_at: "2021-03-17 08:09:00.244459000 +0000", name: "John Von Doe", preferences: {}>
            got: #<Spree::Address id: 4, firstname: "John", lastname: "Von Doe", address1: "10 Lovely Street", address...0 +0000", updated_at: "2021-03-17 08:09:00.244459000 +0000", name: "John Von Doe", preferences: nil>

This commit introduces a hack to have both things. When `preferences` is empty at database level, it's safe to always return a Hash, because that's how the data would have been deserialized anyway. This allows us to call `preferences[:something]` on models that do not explicitly define any preference.

NOTE: This is kind of a hack just to allow backward compatibility. https://github.com/solidusio/solidus/pull/3997 probably provides a better solution to this problem but it is not backward compatible. Long-term it's maybe better to adopt that approach but, in my opinion, it needs a deprecation cycle to allow all stores and extensions to explicitly set the serialization methods when needed.

**Checklist:**

### DIFF
--- a/core/app/models/spree/base.rb
+++ b/core/app/models/spree/base.rb
@@ -5,6 +5,10 @@ class Spree::Base < ActiveRecord::Base
   include Spree::Core::Permalinks
   include Spree::RansackableAttributes
 
+  def preferences
+    read_attribute(:preferences) || self.class.preferences_coder_class.new
+  end
+
   def initialize_preference_defaults
     if has_attribute?(:preferences)
       self.preferences = default_preferences.merge(preferences)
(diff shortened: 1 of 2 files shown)


tvdeyen (approved):
Looks good to me.

Would like to hear from @mamhoff as well.

mamhoff (approved):
Nice work, and thanks for responding so quickly!

spaghetticode (approved):
@kennyadsl thanks 👍

PR #4003, opened 2021-03-19
Switch to the correct ActiveStorage variant syntax

filippoliverani:
**Description**

This PR updates the syntax that we're using to create variants from main `ActiveStorage` images in our attachment decorator.
Since Rails 6.0, [`MiniMagik`](https://github.com/minimagick/minimagick) syntax [has been depreacted](https://github.com/rails/rails/blob/6-0-stable/activestorage/CHANGELOG.md) in favor of [`ImageProcessing`](https://github.com/janko/image_processing) wrapper gem. 

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/active_storage_adapter/attachment.rb
+++ b/core/app/models/concerns/spree/active_storage_adapter/attachment.rb
@@ -9,11 +9,9 @@ module Spree
     class Attachment
       delegate_missing_to :@attachment
 
-      DEFAULT_SIZE = '100%'
-
       def initialize(attachment, styles: {})
         @attachment = attachment
-        @styles = styles
+        @styles = normalize_styles(styles)
       end
 
       def exists?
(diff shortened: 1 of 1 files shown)


core/app/models/concerns/spree/active_storage_adapter/attachment.rb
```diff
@@ -19,11 +19,11 @@ def exists?
       end
 
       def filename
-        blob.filename.to_s
+        blob&.filename.to_s
```
jarednorman, line 22: Should this have two safe operators or are you fine with the empty string when `blob` is nil?

```suggestion
        blob&.filename&.to_s
``

filippoliverani: I'm not sure it is needed, the safe navigation was there to avoid calling `Blob#filename` on `nil`, returning `nil` or an empty string should not make a difference to `Attachement#filename` users

kennyadsl (approved):
Thanks Filippo!

kennyadsl:
@jarednorman can you please review again?

jarednorman (approved):

PR #4005, opened 2021-03-25
Failing spec for calculator with preferences that aren't known

mamhoff:
If you try accessing a calculator with preferences in the DB, and that
calculator hasn't seen the `serialize` call, Rails will try accessing
the preferences column directly and return a string :(

**Checklist:**


kennyadsl:
Shouldn't those preferences be declared in the calculator with `preference :a`?

mamhoff:
I've added this commit to #4013

PR #4007, opened 2021-03-26
Allow customer returns to reference existing `ReturnItem`s on create through API

forkata:
**Description**
This change fixes an issue in the API customer returns controller `create` action when referencing existing return items from a return authorization. This works on `update` but Rails has a limitation where it cannot set the `customer_return_id` on the `ReturnItem` records when they are passed as nested attributes, because that record doesn't exist. This will fail with an `ActiveRecord::RecordNotFound` error. This behaviour is documented, though slightly buried deep in the documentation [here](https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html#module-ActiveRecord::NestedAttributes::ClassMethods-label-One-to-many).

We already allow this in the `CustomerReturnsController` in the backend, however it requires a bit of parameter mangling to get the exact behaviour we want. In this change we have pulled the `build_return_items_from_params` before action from the backend controller and modified it to better fit the API use case. 

The main changes required to the `before_action` were to
* remove the code to exclude items with the `returned` flag, which is specific to the backend form.
* allowed the nested `return_items_attributes` to be passed as a hash of hashes (default for nested forms) as well as an `Array`. I've also added a deprecation when providing these as a hash of hashes.
* remove the custom errors the backend adds when a `reception_status_event` is not provided, this is not required by the API controller, but required in the backend.

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/customer_returns_controller.rb
+++ b/api/app/controllers/spree/api/customer_returns_controller.rb
@@ -4,13 +4,14 @@ module Spree
   module Api
     class CustomerReturnsController < Spree::Api::BaseController
       before_action :load_order
+      before_action :build_customer_return, only: [:create]
       around_action :lock_order, only: [:create, :update, :destroy, :cancel]
 
       rescue_from Spree::Order::InsufficientStock, with: :insufficient_stock_error
 
       def create
         authorize! :create, CustomerReturn
-        @customer_return = CustomerReturn.create(customer_return_params)
+
         if @customer_return.save
           respond_with(@customer_return, status: 201, default_template: :show)
         else
@@ -61,6 +62,30 @@ module Spree
 
       def customer_return_params
         params.require(:customer_return).permit(permitted_customer_return_attributes)
+      end
+
+      def build_customer_return
+        customer_return_attributes = customer_return_params
+        return_items_params = customer_return_attributes.
+          delete(:return_items_attributes)
+
+        if return_items_params.is_a? ActionController::Parameters
+          return_items_params = return_items_params.values
+        end
+
+        @customer_return = CustomerReturn.new(customer_return_attributes)
+
+        @customer_return.return_items = return_items_params.map do |item_params|
+          return_item = if item_params[:id]
+                          Spree::ReturnItem.find(item_params[:id])
+                        else
+                          Spree::ReturnItem.new
+                        end
+
+          return_item.assign_attributes(item_params)
+
+          return_item
+        end
       end
     end
   end
--- a/api/app/controllers/spree/api/customer_returns_controller.rb
+++ b/api/app/controllers/spree/api/customer_returns_controller.rb
@@ -71,6 +71,10 @@ module Spree
 
         if return_items_params.is_a? ActionController::Parameters
           return_items_params = return_items_params.values
+          Spree::Deprecation.warn(
+            "Passing `return_items_attributes` as a hash of hashes is \
+deprecated and will be removed in future versions of Solidus."
+          )
         end
 
         @customer_return = CustomerReturn.new(customer_return_attributes)
--- a/api/spec/requests/spree/api/customer_returns_controller_spec.rb
+++ b/api/spec/requests/spree/api/customer_returns_controller_spec.rb
@@ -183,11 +183,21 @@ module Spree
           end
 
           it "can create a new customer return" do
+            expect(Spree::Deprecation).to receive(:warn)
             expect { subject }.to change { Spree::CustomerReturn.count }.
               from(0).to(1)
 
             expect(response).to have_http_status(:success)
             expect(json_response).to have_attributes(attributes)
+          end
+
+          it "logs a deprecation warning" do
+            expect(Spree::Deprecation).
+              to receive(:warn).
+              with(
+                /Passing `return_items_attributes` as a hash of hashes is deprecated/
+              )
+            subject
           end
         end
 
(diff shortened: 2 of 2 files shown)


kennyadsl:
@forkata Thanks for the PR!

For anyone else reviewing, here's a handy link to [the backend counterpart for this feature](https://github.com/solidusio/solidus/blob/a17109b07a817ae34bc61491c3ed1ea824ee32c5/backend/app/controllers/spree/admin/customer_returns_controller.rb#L68-L79).

> remove the code to exclude items with the returned flag, which is specific to the backend form 

👍 

> remove support for passing the return_items_attributes as a hash of hashes (default for nested forms) and instead expects that field to be an Array. We could add this back as it may be a breaking change if anyone is passing these attributes in that way through the API which the create call on an ActiveRecord object should handle. I don't think this is a documented feature in our API docs as the type is specified as array here).

Do you think there's a way to deprecate passing the hash, just to be sure we are not breaking existing applications? I'm afraid that a lot of people are looking at the source code to understand how to use our APIs, also because at the moment the doc is not super reliable as you saw.

> add custom errors to the CustomerReturn and fail the create call if a reception_status_event is not provided in each the return_items_attributes. This is the behaviour in the backend controller, but I now realize reception_status_event is not required in our API docs 🤔, so I am happy to remove this change or update the docs if the field is actually required.

I'd avoid adding this validation, maybe just because I don't understand why a `reception_status_event` is needed. I'm open to change my mind if having a record built that way can create data inconsistencies. Did you get that?

forkata:
Thank you for your quick feedback @kennyadsl!

> > remove support for passing the return_items_attributes as a hash of hashes (default for nested forms) and instead expects that field to be an Array. We could add this back as it may be a breaking change if anyone is passing these attributes in that way through the API which the create call on an ActiveRecord object should handle. I don't think this is a documented feature in our API docs as the type is specified as array here).
> 
> Do you think there's a way to deprecate passing the hash, just to be sure we are not breaking existing applications? I'm afraid that a lot of people are looking at the source code to understand how to use our APIs, also because at the moment the doc is not super reliable as you saw.

That is a great suggestion! I was contemplating handling both formats to ensure backwards compatibility. What I can do is add a change and corresponding tests to handle `return_items_attributes` as a hash of hashes and add a deprecation warning if we go down that code path.

> > add custom errors to the CustomerReturn and fail the create call if a reception_status_event is not provided in each the return_items_attributes. This is the behaviour in the backend controller, but I now realize reception_status_event is not required in our API docs 🤔, so I am happy to remove this change or update the docs if the field is actually required.
> 
> I'd avoid adding this validation, maybe just because I don't understand why a `reception_status_event` is needed. I'm open to change my mind if having a record built that way can create data inconsistencies. Did you get that?

I was also unsure if this was a `backend` specific requirement, but assumed it was not because of this being a feature of the state machine which is defined in the model. I am going to do some more testing on this and remove if it isn't required to create or update the `CustomerReturn` record or when creating new return items through this call. It would be also good to have this behaviour tested in the specs for this controller so I will try to add to that.

forkata:
Hi @kennyadsl,
After some more testing here I've updated the PR and it should be now ready for review. I made the following changes based on our conversation
* the `reception_status_event` field appears to be optional and we can create the CR without providing that. I've added a test to show usage with and without specifying that in the payload and removed the custom error when this field was omitted.
* i've added a test for passing `return_items_attributes` as a hash of hashes and ensured my change is backwards compatible for this, as well as added a deprecation warning in case this code path is reached. It is relatively simple to support both, but I feel like this was undocumented and probably not something we should support through the API.

kennyadsl (approved):
Thanks! Going to merge this after we release 3.0 (this will be present in 3.1).

api/app/controllers/spree/api/customer_returns_controller.rb
```diff
@@ -62,6 +63,30 @@ def load_order
...
+          return_item = if item_params[:id]
+                          Spree::ReturnItem.find(item_params[:id])
+                        else
+                          Spree::ReturnItem.new
+                        end
+
+          return_item.assign_attributes(item_params)
+
+          return_item
+        end.compact
```
spaghetticode, line 92: I'm not sure why we need `compact` after the `map` cycle, I suppose we're always returning a `Spree::ReturnItem` with each iteration, and when a return item cannot be found from `item_params[:id]` an  exception is raised... but maybe I'm overlooking some scenario?

Also, I think the naming of the method reflects only part of what this method does here. What about a more general `build_customer_return`, since it's actually building the instance stored in`@customer_return`?

spaghetticode (commented):
@forkata thanks for this fix, I really appreciated being guided in the review by the very detailed commits ❤️ 

I left a couple of comments, let me know what you think!

api/app/controllers/spree/api/customer_returns_controller.rb
forkata: > I'm not sure why we need compact after the map cycle
I agree, upon closer inspection I don't think the `.compact` is necessary here.

> I think the naming of the method reflects only part of what this method does here.
That's a great suggestion, I think `build_customer_return` is a better name here, I will update that!

Thanks for the feedback, I will update this change to reflect the above suggestions!

forkata: @spaghetticode I've addressed your comments here and this should be ready for another review 🙏🏼

spaghetticode (approved):
@forkata thank you, LGTM 👍

PR #4012, opened 2021-03-29
Support the selection of currency by customers

waiting-for-dev:
**Description**

A new `Store#currencies` attribute has been added. It contains a set of
currencies that are supported by a store in addition to
`Store#default_currency`.

The currency in the frontend can be switched if it set within the
`'currency'` key in the session. Otherwise, how a default is picked
doesn't change: `Store#default_currency` is taken or, if not present,
`Spree.config.currency`.

A currency switcher has been added to the frontend. In the case a
product has been added to the cart, it's emptied. There're a couple
of reasons why we're not trying to automagically update amounts in the
cart lines:

- Promotions: We can create promotion rules that only apply to a
  currency. It'd be a surprising behavior if we remove a discount to
  the user.
- Customer safety: It's better to explicitly require a user to click on
  the "Add to cart" button while they are seeing the amount of money that
  will be added. Let's imagine that an administrator makes an error and
  sets a super-high price on a product for a specific currency. If a
  user blindly switch to that currency and they perform the checkout,
  they could have a heart attack just afterward.
- The regular flow should be switching currencies and then adding to the
  cart, anyway.

To have a better UX, when requesting a change in the currency an async
call is made to check whether some items have been added to the cart.
When that's the case, the user is prompted to confirm the action.

The work to empty the cart is made from the frontend for now. As a follow-up, it will be moved to a service so that Solidus' users can customize what happens to the order when the currency is switched. We can also consider letting users configure how the default currency is selected, as a common use-case would be selecting one or another depending on the IP.

https://user-images.githubusercontent.com/52650/112824308-aac73280-908a-11eb-96dd-fe460290cb37.mp4

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -724,6 +724,15 @@ module Spree
       end
     end
 
+    # Returns whether have been added some {#line_items}
+    #
+    # It considers both persisted and in-memory line items.
+    #
+    # @return [Boolean]
+    def items?
+      line_items.any?
+    end
+
     private
 
     def process_payments_before_complete
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -724,15 +724,6 @@ module Spree
       end
     end
 
-    # Returns whether have been added some {#line_items}
-    #
-    # It considers both persisted and in-memory line items.
-    #
-    # @return [Boolean]
-    def items?
-      line_items.any?
-    end
-
     private
 
     def process_payments_before_complete
--- a/core/lib/spree/core/controller_helpers/pricing.rb
+++ b/core/lib/spree/core/controller_helpers/pricing.rb
@@ -7,11 +7,45 @@ module Spree
         extend ActiveSupport::Concern
 
         included do
-          helper_method :current_pricing_options
+          # Key where the current currency should be stored in a requests' session
+          CURRENCY_SESSION_KEY = 'currency'
+
+          helper_method :current_pricing_options,
+                        :current_currency,
+                        :switch_currency
         end
 
+        # @see Variant::PricingOptions#from_context
         def current_pricing_options
           Spree::Config.pricing_options_class.from_context(self)
+        end
+
+        # Current currency from {#current_pricing_options}
+        #
+        # @return [String]
+        # @see #current_pricing_options
+        # @example
+        #   'USD'
+        def current_currency
+          current_pricing_options.currency
+        end
+
+        # Currency stored in the session
+        #
+        # @return [String, nil]
+        # @see CURRENCY_SESSION_KEY
+        # @example
+        #   'USD'
+        def currency_in_session
+          session[CURRENCY_SESSION_KEY]
+        end
+
+        # Stores given currency in the session
+        #
+        # @param currency [String]
+        # @see #currency_in_session
+        def switch_currency(currency)
+          session[CURRENCY_SESSION_KEY] = currency
         end
       end
     end
--- a/frontend/app/assets/javascripts/spree/frontend/currency_selector.js
+++ b/frontend/app/assets/javascripts/spree/frontend/currency_selector.js
@@ -1,0 +1,16 @@
+$(function() {
+  $('#currency_selector select').change(function() {
+    var select = this;
+    var form = this.form;
+    Spree.ajax({
+      url: Spree.pathFor('/orders/current_order_has_items'),
+      success: function(data) {
+        if (data['result'] && !confirm($(form).data('confirm-text'))) {
+          $(select).val($(form).data('current-currency'));
+        } else {
+          form.submit();
+        }
+      }
+    });
+  });
+});
--- a/frontend/app/assets/javascripts/spree/frontend/currency_selector.js
+++ b/frontend/app/assets/javascripts/spree/frontend/currency_selector.js
@@ -2,15 +2,10 @@ $(function() {
   $('#currency_selector select').change(function() {
     var select = this;
     var form = this.form;
-    Spree.ajax({
-      url: Spree.pathFor('/orders/current_order_has_items'),
-      success: function(data) {
-        if (data['result'] && !confirm($(form).data('confirm-text'))) {
-          $(select).val($(form).data('current-currency'));
-        } else {
-          form.submit();
-        }
-      }
-    });
+    if ($(form).data('needs-confirm') && !confirm($(form).data('confirm-text'))) {
+      $(select).val($(form).data('current-currency'));
+    } else {
+      form.submit();
+    }
   });
 });
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -101,6 +101,10 @@ module Spree
       end
     end
 
+    def current_order_has_items
+      render json: { result: current_order_in_progress? }
+    end
+
     private
 
     def store_guest_token
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -101,10 +101,6 @@ module Spree
       end
     end
 
-    def current_order_has_items
-      render json: { result: current_order_in_progress? }
-    end
-
     private
 
     def store_guest_token
--- a/frontend/app/views/spree/shared/_currency_selector.html.erb
+++ b/frontend/app/views/spree/shared/_currency_selector.html.erb
@@ -1,0 +1,19 @@
+<% if multicurrency? %>
+  <li id="currency_selector" data-hook>
+    <%= form_tag spree.select_currency_path, class: 'navbar-form', method: :put, data: { current_currency: current_currency, confirm_text: I18n.t('spree.currency_switch_confirm') } do %>
+      <div class="form-group">
+        <label for="switch_currency" class="sr-only">
+          <%= I18n.t('spree.currency') %>
+        </label>
+        <%= select_tag(
+            :switch_to_currency,
+            options_for_select(
+              supported_currencies,
+              selected: current_currency
+            ),
+            class: 'form-control') %>
+        <noscript><%= submit_tag t("spree.select") %></noscript>
+      </div>
+    <% end %>
+  </li>
+<% end %>
--- a/frontend/app/views/spree/shared/_currency_selector.html.erb
+++ b/frontend/app/views/spree/shared/_currency_selector.html.erb
@@ -1,6 +1,13 @@
 <% if multicurrency? %>
   <li id="currency_selector" data-hook>
-    <%= form_tag spree.select_currency_path, class: 'navbar-form', method: :put, data: { current_currency: current_currency, confirm_text: I18n.t('spree.currency_switch_confirm') } do %>
+    <%= form_tag spree.select_currency_path,
+      class: 'navbar-form',
+      method: :put,
+      data: {
+        current_currency: current_currency,
+        confirm_text: I18n.t('spree.currency_switch_confirm'),
+        needs_confirm: current_order_in_progress?
+      } do %>
       <div class="form-group">
         <label for="switch_currency" class="sr-only">
           <%= I18n.t('spree.currency') %>
--- a/backend/app/controllers/spree/admin/stores_controller.rb
+++ b/backend/app/controllers/spree/admin/stores_controller.rb
@@ -3,6 +3,8 @@
 module Spree
   module Admin
     class StoresController < Spree::Admin::ResourceController
+      before_action :cast_currencies, only: %i[create update]
+
       def index
         if Spree::Store.count == 1
           redirect_to edit_admin_store_path(Spree::Store.first)
@@ -19,6 +21,10 @@ module Spree
 
       def permitted_params
         Spree::PermittedAttributes.store_attributes
+      end
+
+      def cast_currencies
+        params['store']['currencies'] = params['store']['currencies']&.to_set || Set[]
       end
     end
   end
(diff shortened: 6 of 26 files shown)


waiting-for-dev:
Following offline feedback by @kennyadsl I have made two kinds of changes:

- Align naming with the previous `locale#set` action.
- Moved session knowledge out of `PricingOptions` and into the controller layer.

frontend/app/views/spree/shared/_currency_selector.html.erb
```diff
@@ -0,0 +1,19 @@
...
+          <%= I18n.t('spree.currency') %>
+        </label>
+        <%= select_tag(
+            :switch_to_currency,
+            options_for_select(
+              supported_currencies,
+              selected: current_currency
+            ),
+            class: 'form-control') %>
+      <noscript><%= submit_tag t("spree.select") %></noscript>
```
kennyadsl, line 15: Should this line be indented more?

frontend/app/assets/javascripts/spree/frontend/currency_selector.js
```diff
@@ -0,0 +1,16 @@
+$(function() {
```
kennyadsl, line 1: I was thinking we could remove all the JS, including the ajax call and the new route/controller action if we use [`data-url`](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#data-url-and-data-params) in combination with the select tag instead.

For the optional confirm message, maybe we can just add `data-confirm` on the element conditionally checking if the order has items directly in the view.

```erb
          <%= select_tag(
            :switch_to_currency,
            options_for_select(
              supported_currencies,
              selected: current_currency
            ),
            class: 'form-control'),
            data: {
              url: spree.select_currency_path,
              method: :put,
              confirm: current_order_in_progress? I18n.t('spree.currency_switch_confirm') : false
            } %>
```

I didn't test it and not 100% sure it works as is. Thoughts?

core/lib/spree/core/controller_helpers/pricing.rb
```diff
@@ -7,12 +7,46 @@ module Pricing
...
         end
+
+        # Current currency from {#current_pricing_options}
+        #
+        # @return [String]
+        # @see #current_pricing_options
+        # @example
+        #   'USD'
+        def current_currency
+          current_pricing_options.currency
```
kennyadsl, line 30: Question: I don't get why we can't use the value of the currency from the session here. Why the currency set on `current_pricing_options` represents the `current_currency` instead?

core/lib/spree/core/controller_helpers/pricing.rb
```diff
@@ -7,12 +7,46 @@ module Pricing
...
+          session[CURRENCY_SESSION_KEY]
+        end
+
+        # Stores given currency in the session
+        #
+        # @param currency [String]
+        # @see #currency_in_session
+        def switch_currency(currency)
+          session[CURRENCY_SESSION_KEY] = currency
+        end
```
kennyadsl, line 49: What about creating a new object to store the currency information that will be used here? Something like:

```ruby
class Spree::Currency::SessionStore
  CURRENCY_SESSION_KEY = 'currency'

  def initialize(context)
    @session = context.session
  end

  def currency
    @session[CURRENCY_SESSION_KEY]
  end

  def currency=
    @session[CURRENCY_SESSION_KEY] = currency
  end
end
```

Setting the currency store in a preference like

```ruby
Spree::Config.currency_store = Spree::Currency::SessionStore
```

And finally, use it in the pricing helper as:

```ruby
        def currency_in_session
          currency_store.currency
        end

        def switch_currency(currency)
          currency_store.currency = currency
        end

        private

        def currency_store
          @currency_store ||= Spree::Config.currency_store.new(self)
        end
```

The benefit of this approach is that it allows people to change the way currency is stored. They would just need to create their own "currency store" class that responds to `currency` and `currency=` and change the preference.

kennyadsl (commented):
Left another review, thanks Marc!

frontend/app/views/spree/shared/_currency_selector.html.erb
waiting-for-dev: :mag: fixed!!!

frontend/app/assets/javascripts/spree/frontend/currency_selector.js
waiting-for-dev: It would change the behavior in two ways:

1. It would change the switch currency action to work asynchronously.
2. We would lose the asynchronous call to check whether the cart is empty.

About 1, I don't like the idea because the frontend architecture follows good old sync requests-responses, so we should add complexity to update the DOM and display the new prices. It would also be inconsistent with how locale switching works.

About 2, using the `data-confirm` value, which is set when the HTML is loaded, would make it possible that we missed products added on another browser tab.

core/lib/spree/core/controller_helpers/pricing.rb
waiting-for-dev: When users haven't selected a currency, the session has no information about it. In this case, the default for the currency is taken. This is the logic that `current_pricing_options.currency` implements.

frontend/app/assets/javascripts/spree/frontend/currency_selector.js
kennyadsl: You are right, I didn't consider we want this request to be sync. In terms of `data-confirm`, personally, I think that for the sake of simplicity, removing the real-time check is a good trade-off. I think we can ship a basic version when someone has multiple tabs open and doesn't see the popup if they change the currency in some of those tabs. If that's important for some businesses, it's very easy to add that portion of code in their app.

core/lib/spree/core/controller_helpers/pricing.rb
waiting-for-dev: Undoubtedly it would add more flexibility. But don't you think it would be overkilling? Do you think we need to provide this flexibility at this point? I'm open to implementing it, it's just that I'm considering complexity as a trade-off.

kennyadsl: You are probably right, I also don't want to anticipate a design before we actually need it. I'm ok with the current version, we can always change this later if needed.

kennyadsl: To be honest, I have been triggered by `currency_in_session` method name, which seems too tied to the implementation. Maybe it's a good thing to be explicit but in this case shouldn't we use the same approach when we set the currency, so using something like `switch_currency_in_session`?

frontend/app/assets/javascripts/spree/frontend/currency_selector.js
waiting-for-dev: Probably it makes sense... I removed the async check in a new commit. Please, take a look at it. As we still need to keep the current currency as the selected option when the action is aborted, we can't rely on Rails `data-confirm` magic attribute, so it's still done through the JS file.

core/lib/spree/core/controller_helpers/pricing.rb
waiting-for-dev: I had the same internal conversation about it :wink: But I decided to go with this version because `switch_currency` is in fact changing the end currency seen by the user (which is indeed the same that it's in the session), while the reader method `currency_in_session` needs to be differentiated from `current_currency`, which will take the one from the session only when it's set. WDYT?

jarednorman (approved):

core/app/models/spree/order.rb
```diff
@@ -721,6 +721,15 @@ def validate_payments_attributes(attributes)
       end
     end
 
+    # Returns whether have been added some {#line_items}
+    #
+    # It considers both persisted and in-memory line items.
+    #
+    # @return [Boolean]
+    def items?
```
spaghetticode, line 729: mmm, I'm not 100% convinced by this method. It's used only once in this PR, and if I look in the codebase for `line_items.any?` I can find only occurrences with a block, which are not replaceable with this implementation.

frontend/app/controllers/spree/orders_controller.rb
```diff
@@ -101,10 +101,6 @@ def accurate_title
       end
     end
 
-    def current_order_has_items
```
spaghetticode: Shouldn't we deprecate this controller action before removing it?

spaghetticode (commented):
@waiting-for-dev thanks for adding this new feature ❤️ 

It looks good to me in general, I left a couple of comments on the code, then I am curious about the naming, which seemed a bit confusing to me at first sight. Before getting to the code, I was expecting `Store#currencies` to include the default currency, as the latter is a currency as well. Wouldn't a different name such as `secondary_currencies` or `optional_currencies` be more descriptive?

core/app/models/spree/order.rb
jarednorman: Yeah, it doesn't strike me as a particularly valuable abstraction over just calling `line_items.any?`, even if the functionality is useful.

frontend/app/controllers/spree/orders_controller.rb
waiting-for-dev: That action is not in master. It's the second commit in this PR, removing the action introduced in the first one. It's discussed in https://github.com/solidusio/solidus/pull/4012#discussion_r607654632. I haven't squashed because I want to be sure that @kennyadsl is ok with the result.

core/app/models/spree/order.rb
waiting-for-dev: Good point. We can also modify the method to accept the block and replace the other occurrences while keeping Demeter happy :slightly_smiling_face: WDYT?

waiting-for-dev:
> I am curious about the naming, which seemed a bit confusing to me at first sight. Before getting to the code, I was expecting `Store#currencies` to include the default currency, as the latter is a currency as well. Wouldn't a different name such as `secondary_currencies` or `optional_currencies` be more descriptive?

Yeah, I agree it's confusing, and I had contradictory thoughts about it. In general, the management of currencies is not very clear on Solidus. The `default_currency` field is not mandatory, and in its absence, it will use `Spree::Config.currency`. But regardless of `default_currency` being set, most parts of the application will only use `Spree::Config.currency`. In fact, `default_currency` is only used in the frontend. For this reason, I don't like a name like `optional_currencies` because `default_currency` is also optional.

`secondary_currencies` is a better fit, but still, I don't like it. The reason is that at some point, I'd like to add the option for users to configure the main currency depending on the request information (in the absence of a currency set in the session). For instance, you may want to default to USD if the visitor is from the US and EUR if they are from Germany. This is not trivial, as, at this point, we're setting it from the model:

```ruby
def main_currency
  default_currency.presence || Spree::Config.currency
end
```

But I'd like to end up with something like (not sure in which layer):

```ruby
def main_currency
  Spree::Config.main_currency(store, request)
end
```

`Spree::Config.main_currency` probably would default to ignore `request` and return ` default_currency.presence || Spree::Config.currency` but users would be able to override it.

So, the other currencies would be secondaries in a default installation, but one could become the main one if a user decides to decide depending on the request.

What do you think about a name like `other_currencies`? It does carry a "secondary" semantics, too, but it's less strongly about it.

core/app/models/spree/order.rb
spaghetticode: I would be more convinced, but still not 100%, for these reasons:
* the current name doesn't directly map to what it aliases, so it increases the cognitive distance (but this can be easily fixed by changing to `line_items_any?`);
* when we add this method, then why not adding a similar one for any occurrence of `whatever_association#any?` we have in the codebase? If it is good this time, it then should be always good... but I think this would add mostly overhead;
* some people don't agree we should keep Demeter happy all the time, for example https://www.tedinski.com/2018/12/18/the-law-of-demeter.html. In this case, I think that the biggest value for this change would be when adding clarity/removing a lot of duplication, but we don't have such thing now, since:
  * there are only 2 occurrences (3 with your change) in the application code, plus another 2 in specs;
  * it seems to be perfectly fine in the Rails ecosystem to call `#association_name.any?` (with or without the block) so I think it's very unlikely that other people (contributors, developers) will start using that method, even when aware of its existence;

By the way, I think it would be totally fine if ActiveRecord added something similar, as it would be at the API level and everybody would be encouraged to know about it and start using it. But I see little value in having this single custom method in this Solidus class.

frontend/app/controllers/spree/orders_controller.rb
spaghetticode: Oh sure, sorry 🤦

core/app/models/spree/order.rb
waiting-for-dev: I agree that the Law of Demeter shouldn't be seen as carved in stone. But, in general, I don't like to ask for data and perform logic afterward, as objects should keep data and behavior together. I also understand this is neither carved in stone, but I think that most of the cases when this doesn't apply reveals you have a problem with your design. It's not the case here, or at least we have little control over it, as they are standard Rails associations.

But I also understand that it can also lead to bloated interfaces. With AR having so many public methods, if we did that for every method, then we would land on that scenario. In my view, in this case, asking whether an order has line items should be packed with the order itself, but I agree it's a moot point, so I've changed the code along with your suggestion.

spaghetticode:
@waiting-for-dev thanks for the detailed explanation on the naming ❤️

>In general, the management of currencies is not very clear on Solidus

unfortunately, I'm definitely a novice when it comes to the way Solidus handles currencies, but based on what you wrote I must agree with you. 

I'm wondering if we should first try to clean up some of that confusion before merging this PR. I am envisioning a scenario where this cleanup would help in simplifying things here as well. For example:

> I don't like a name like optional_currencies because default_currency is also optional.

Without knowing the implementation details and the possible reason for this, at a philosophical level I don't understand why Solidus makes `default_currency` optional. I mean, every store needs **at least** one currency, and that would be the default one, right? Especially if we offer multicurrency... Rails allows for multiple locales, but even when you use only one, `I18n.default_locale` is automatically set for you, if you don't do it explicitly or change to another value than `:en`.

kennyadsl (approved):
Thanks @waiting-for-dev!

PR #4013, opened 2021-03-29
Deprecate calling preferences without serialization

mamhoff:
**Description**

In Solidus < 2.11.4, `Spree::Base` had a `serialize :preferences, Hash` statement. This statement fails in Rails 6.1ff, as after that Rails will try deserializing the column even if it isn't there. 

What this commit does is: 
- Remove the `serialize` call from `Spree::Base`, as many inheriting classes do not have a `preferences` column
- Add the `serialize` and `after_initialize` call to those base classes that need it
  - `Spree::Calculator`
  - `Spree::PromotionAction`
  - `Spree::PaymentMethod`
  - `Spree::PromotionRule`
- Add a new `preferences` getter to `Spree::Base` that exhibits the following behaviour: If `preferences` is not a Hash, we can safely assume that the class of the instance we're trying to get preferences from expects `Spree::Base` to perform the `serialize` call for them. In that case, call `serialize` on the instance's class, ensuring that all future loads of this type of object will have `preferences` serialized, and parse the attribute using `ActiveRecord::Type::Serialized` (as that's how `Spree::Base` worked).
- Remove the `preferences_coder_class` method. I don't think it's prudent to add an extension point here for something that was hardcoded in the past and can easily be circumnavigated by e.g. setting one's custom class to serialize using JSON: 

```rb
class MyPreferences
  serialize :preferences, JSON
  after_initialize :initialize_preference_defaults
  ...
end
```

This PR goes against the `v2.11` branch of Solidus; I would like to uphold my PR #3997 for the `master` branch. 

This means that the deprecation window for this change would be `v2.11.8` => `v3.0.0`, but I think that's doable. There is a bit of work in related projects, but it just consists in adding to lines to configuration classes. 

**Checklist:**
- [ ] <del> I have updated Guides and README accordingly to this change (if needed) </del>
- [ ] <del> II have attached screenshots to this PR for visual changes (if needed) </del>

### DIFF
--- a/core/app/models/spree/base.rb
+++ b/core/app/models/spree/base.rb
@@ -6,23 +6,36 @@ class Spree::Base < ActiveRecord::Base
   include Spree::RansackableAttributes
 
   def preferences
-    read_attribute(:preferences) || self.class.preferences_coder_class.new
+    value = read_attribute(:preferences)
+    if !value.is_a?(Hash)
+      Spree::Deprecation.warn <<~WARN
+        #{self.class.name} has a `preferences` column, but does not explicitly (de)serialize this column.
+        In order to make #{self.class.name} work with future versions of Solidus (and Rails), please add the
+        following to lines to your class:
+        ```
+        class #{self.class.name}
+          serialize :preferences, Hash
+          after_initialize :initialize_preference_defaults
+          ...
+        end
+        ```
+      WARN
+      self.class.serialize :preferences, Hash
+      self.class.after_initialize :initialize_preference_defaults
+
+      ActiveRecord::Type::Serialized.new(
+        ActiveRecord::Type::Text.new,
+        ActiveRecord::Coders::YAMLColumn.new(:preferences, Hash)
+      ).deserialize(value)
+    else
+      value
+    end
   end
 
   def initialize_preference_defaults
     if has_attribute?(:preferences)
       self.preferences = default_preferences.merge(preferences)
     end
-  end
-
-  # Only run preference initialization on models which requires it. Improves
-  # performance of record initialization slightly.
-  def self.preference(*args)
-    # after_initialize can be called multiple times with the same symbol, it
-    # will only be called once on initialization.
-    serialize :preferences, preferences_coder_class
-    after_initialize :initialize_preference_defaults
-    super
   end
 
   if Kaminari.config.page_method_name != :page
@@ -33,10 +46,6 @@ class Spree::Base < ActiveRecord::Base
 
       send Kaminari.config.page_method_name, num
     end
-  end
-
-  def self.preferences_coder_class
-    Hash
   end
 
   self.abstract_class = true
--- a/core/app/models/spree/base.rb
+++ b/core/app/models/spree/base.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'spree/preferences/persistable'
+
 class Spree::Base < ActiveRecord::Base
   include Spree::Preferences::Preferable
   include Spree::Core::Permalinks
@@ -11,17 +13,15 @@ class Spree::Base < ActiveRecord::Base
       Spree::Deprecation.warn <<~WARN
         #{self.class.name} has a `preferences` column, but does not explicitly (de)serialize this column.
         In order to make #{self.class.name} work with future versions of Solidus (and Rails), please add the
-        following to lines to your class:
+        following lines to your class:
         ```
         class #{self.class.name}
-          serialize :preferences, Hash
-          after_initialize :initialize_preference_defaults
+          include Spree::Preferences::Persistable
           ...
         end
         ```
       WARN
-      self.class.serialize :preferences, Hash
-      self.class.after_initialize :initialize_preference_defaults
+      self.class.include Spree::Preferences::Persistable
 
       ActiveRecord::Type::Serialized.new(
         ActiveRecord::Type::Text.new,
@@ -29,12 +29,6 @@ class Spree::Base < ActiveRecord::Base
       ).deserialize(value)
     else
       value
-    end
-  end
-
-  def initialize_preference_defaults
-    if has_attribute?(:preferences)
-      self.preferences = default_preferences.merge(preferences)
     end
   end
 
--- a/core/app/models/spree/base.rb
+++ b/core/app/models/spree/base.rb
@@ -6,6 +6,22 @@ class Spree::Base < ActiveRecord::Base
   include Spree::Core::Permalinks
   include Spree::RansackableAttributes
 
+  def self.preference(*args)
+    Spree::Deprecation.warn <<~WARN
+      #{name} has a `preferences` column, but does not explicitly (de)serialize this column.
+      In order to make #{name} work with future versions of Solidus (and Rails), please add the
+      following line to your class:
+      ```
+      class #{name}
+        include Spree::Preferences::Persistable
+        ...
+      end
+      ```
+    WARN
+    include Spree::Preferences::Persistable
+    preference(*args)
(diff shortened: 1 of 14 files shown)


tvdeyen (approved):
I like this approach.

mamhoff:
The failing spec seems unrelated; can someone retrigger a build?

core/spec/models/spree/base_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    ActiveRecord::Migration[4.2].verbose = false
+    CreatePrefTest.migrate(:up)
+  end
+
+  after(:all) do
+    CreatePrefTest.migrate(:down)
+    ActiveRecord::Migration[4.2].verbose = @migration_verbosity
+  end
+
+  context "with a class that has a preference column but does not explicitly serialize" do
```
kennyadsl, line 30: I guess you omitted to test `a class that has a preference column and does explicitly serialize` because it's already covered in the specific classes specs + `spec/models/spree/preferences/preferable_spec.rb`, isn't it? 

Wondering if having these specs split over multiple files is a good thing. I mean we could move them under preferable, or move the ones in that file that depend on the methods defined in `Spree::Base` here. Thoughts?

kennyadsl (commented):
Great stuff Martin! I only left some comments/questions but overall I think it's good to merge both in 2.11 and 3.0. If that was breaking your application, I think it's safe to consider this a patch.

.circleci/config.yml
```diff
@@ -140,6 +149,7 @@ workflows:
               only: /master|v\d\.\d+/
       - postgres
       - mysql
+      - postgres_rails61
```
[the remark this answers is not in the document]
mamhoff: Oh, d'uh :) thanks for the heads-up :)

core/app/models/spree/base.rb
```diff
@@ -6,7 +6,26 @@ class Spree::Base < ActiveRecord::Base
...
+        class #{self.class.name}
+          serialize :preferences, Hash
+          after_initialize :initialize_preference_defaults
+          ...
+        end
+        ```
+      WARN
+      self.class.serialize :preferences, Hash
+      self.class.after_initialize :initialize_preference_defaults
+      value ? YAML.safe_load(value, [Symbol]) : {}
```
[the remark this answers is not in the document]
mamhoff: This is the behaviour any class exhibits when you call `serialize :preferences, Hash`. See the docs and the implementation here: 
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/attribute_methods/serialization.rb#L19-L139

Crucially, in the Case of Hash, `null` becomes a `{}`, and anything that can be deserialized into a Hash using YAML will be.

core/spec/models/spree/base_spec.rb
mamhoff: Hm, in `base_spec.rb` I am only testing behaviour that is usually provided by `ActiveRecord::Base.serialize`. I think this spec can go once we get rid of the `Spree::Base#preferences` in Solidus 3.

After that, the contract between a class and the `Spree::Preferences::Preferable` module is to provide a `preferences` method that returns a Hash. I do think that the classes themselves should have a test asserting that.

mamhoff (commented):
> What about providing a module/concern for this? I'm not particularly interested in DRYing things, but if we need to make other changes in the future (eg. deprecate the way we serialize/initialize defaults) it could be handy.

I like that idea. Especially because that means we can have the callback and the `after_initialize` call in the same file. Pretty nice!

core/app/models/spree/base.rb
```diff
@@ -1,30 +1,48 @@
...
-  def initialize_preference_defaults
-    if has_attribute?(:preferences)
-      self.preferences = default_preferences.merge(preferences)
+  def preferences
+    value = read_attribute(:preferences)
+    if !value.is_a?(Hash)
+      Spree::Deprecation.warn <<~WARN
+        #{self.class.name} has a `preferences` column, but does not explicitly (de)serialize this column.
+        In order to make #{self.class.name} work with future versions of Solidus (and Rails), please add the
+        following to lines to your class:
```
kennyadsl, line 31: extra `to` before `lines`?

core/app/models/spree/calculator.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
+require 'spree/preferences/persistable'
```
kennyadsl, line 3: I think this will be handled by the Rails autoloader, or is there a reason why this is needed explicitly?

core/app/models/spree/base.rb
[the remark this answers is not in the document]
kennyadsl: Maybe we can use directly `ActiveRecord::Type::Serialized`?

core/spec/models/spree/promotion_action_spec.rb
```diff
@@ -3,6 +3,12 @@
 require 'rails_helper'
 
 RSpec.describe Spree::PromotionAction, type: :model do
+  describe "preferences" do
+    subject { described_class.new.preferences }
+
+    it { is_expected.to eq({}) }
```
kennyadsl, line 9: `be_a(Hash)` here as well?

core/app/models/spree/promotion_rule.rb
```diff
@@ -3,6 +3,9 @@
 module Spree
   # Base class for all promotion rules
   class PromotionRule < Spree::Base
+    serialize :preferences, Hash
+    after_initialize :initialize_preference_defaults
```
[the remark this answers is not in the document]
mamhoff: Done!

core/app/models/spree/base.rb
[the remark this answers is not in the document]
mamhoff: Yes, that was actually possible (verbose, but possible).

core/app/models/spree/base.rb
mamhoff: Removed

core/app/models/spree/calculator.rb
mamhoff: No, `spree/preferences/*` live in `lib`, and that isn't in the autoload path afaik. I did have to add it to make the tests pass :)

core/spec/models/spree/promotion_action_spec.rb
mamhoff: Nah, `{}` is fine as we know it's an empty Hash for this one. I've used `be_a(Hash)` mostly when for some reason it wasn't crystal clear what would be in there.

core/spec/models/spree/base_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    before :all do
+      class PrefTestWithoutSerialization < Spree::Base
+        self.table_name = :pref_tests
+      end
+    end
+
+    before(:each) do
+      allow(Spree::Deprecation).to receive(:warn).
```
kennyadsl, line 38: What about turning this into an `expect`. The concern is that it may not be emitting the deprecation warning when we expect this happens. I mean, stubbing the deprecation should not be used to avoid failures and clean the logs only but to actually test that users get a deprecation message.

kennyadsl (approved):
I left a non-blocking comment, let me know your thoughts but I'm fine merging even as is.

Thanks!

core/spec/models/spree/base_spec.rb
kennyadsl: Got it, and I can't think at a clean way to do it.

gmacdougall:
I like this change and what it's doing. Thanks for the contribution Martin!

jarednorman (approved):

PR #4019, opened 2021-04-07
Add sqlite executor to CI

waiting-for-dev:
The test suite was only run against mysql and postgres. It's good to add
sqlite as it's is commonly used during development.

I worked on this because we recently had a failing test in the task  `core/lib/tasks/migrations/migrate_default_billing_addresses_to_address_book.rake`. However, it was recently removed (https://github.com/solidusio/solidus/commit/f3f58034b7b93be9f565b07bae357c1ea92c2d1a). I spent some time fixing it and then when I pulled for new changes I realized the file wasn't there anymore... :facepalm: Anyway, having sqlite3 in the CI can help prevent problems like that. Please, tell me if you'd like to backport the fix for that task to 2.1.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,14 @@ executors:
       - image: *image
       - image: circleci/mysql:5.7-ram
 
+  sqlite:
+    working_directory: *workdir
+    environment:
+      <<: *environment
+      DB: sqlite
+    docker:
+      - image: *image
+
 commands:
   setup:
     steps:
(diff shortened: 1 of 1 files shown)


waiting-for-dev:
> I worked on this because we recently had a failing test in the task core/lib/tasks/migrations/migrate_default_billing_addresses_to_address_book.rake. However, it was recently removed (f3f5803). I spent some time fixing it and then when I pulled for new changes I realized the file wasn't there anymore... facepalm Anyway, having sqlite3 in the CI can help prevent problems like that. Please, tell me if you'd like to backport the fix for that task to 2.1.

Just in case someone needs it, I pushed the fix in #4020

kennyadsl:
Thanks @waiting-for-dev, but I'm going to close this for now after talking with the rest of the Core Team. The CI will take forever (~1 hour) if we enable another executor and, considering sqlite is not in production for anyone (I hope), I think not having that DB tested is a good trade-off. 

If we will add more jobs/concurrency to CircleCI later, we can definitely resume this work! 🙏

Issue #4017, opened 2021-04-02
Corrupted Active Storage attachments


kennyadsl:
Hey guys!
I have an issue using `ActiveStorageAttachment` for product images.
I'm using an S3-compatible service for storing attachments, and it's working ok. But if the service returns an error (it is down at the moment the user attempts to upload an image or there's a permission error, etc), a product image record and ActiveStorage blob record are created anyway despite the image processing error. So it produces a corrupted state of the system. And then any code which calls `.url(style)` on this image gets an `ActiveStorage::FileNotFoundError` exception, because it tries to run image processing to create the requested image variant, but it can't find the original image because its upload never finished (it happens on any page which needs to display the image).
Ideally, I would like it to be "all or nothing", so the database records get committed only after the storage service responded with a success.
Is it a problem of how ActiveStorage works itself or is it a Solidus wrapper's one? How do I prevent the situation when one failed image upload breaks down the whole store (you can't even delete the corrupted image, because the admin product page is also broken)?

_Originally posted by @ok32 in https://github.com/solidusio/solidus/discussions/3986_

kennyadsl:
I had another issue uploading a not supported file type (`.heic`), not sure it's related but I also had to remove the image from the database to fix it. To reproduce:

```
rails new store --skip-javascript
cd store
bundle add solidus -v 3.0.0.rb2
bin/rails generate solidus:install
```

Login into the admin and upload a `.heic` file in any product. Reload the page to see the `ActiveStorage::InvariableError`.

cpfergus1:
I was able recreate this issue by:
1. creating a new application
2.  add `gem 'solidus', path: '/Users/*_user_*/solidus-3` and:
```
$ bundle install
$ bin/rails generate solidus:install
```
3. setup s3 bucket at AWS (https://docs.aws.amazon.com/AmazonS3/latest/userguide/creating-bucket.html)
4. change root/config/storage.yml to include s3 bucket:
```amazon:
  service: S3
  access_key_id: <%= Rails.application.credentials.dig(:aws, :access_key_id) %>
  secret_access_key: <%= Rails.application.credentials.dig(:aws, :secret_access_key) %>
  region: us-west-2
  bucket: *_bucket_*```
5. change root/config/environments/development.rb `config.active_storage.service = :amazon`
6. slightly change your AWS credentials prior to booting up your server by commenting out part of your ID
7. Boot up server, and navigate to admin page to try and upload a photo - should receive uploading error.
[...]

PR #4021, opened 2021-04-08
Image attachment content type validation fix for ActiveStorage

cpfergus1:
**Description**

Ref #4017 

This PR aimed to solve the bugs described in Issue #4017 by implementing a validation for content type and to handle a FileNotFound error when ActiveStorage fails to upload to a remote service such as S3. 

The validation allows file types jpeg, png, jpg and gif. More file types can be added and removed by editing the accepted_image_types definition found in active_storage_attachment.rb. Validation tests have been added to the image_spec.rb in core.

The handling of the FileNotFound error maybe a temporary fix as we are aiming to discover another pathway to handle the exception and remove the invalid objects simultaneously. Currently when the error is raised on method call, it returns a corrupted image file path.

**Checklist:**

### DIFF
--- a/core/app/models/spree/image/active_storage_attachment.rb
+++ b/core/app/models/spree/image/active_storage_attachment.rb
@@ -7,6 +7,10 @@ module Spree::Image::ActiveStorageAttachment
   delegate :width, :height, to: :attachment, prefix: true
 
   included do
+    validates :attachment, presence: true
+    validate :attachment_is_an_image
+    validate :supported_content_type
+
     has_attachment :attachment,
                    styles: {
                    mini: '48x48>',
@@ -15,7 +19,11 @@ module Spree::Image::ActiveStorageAttachment
                    large: '1200x1200>'
                  },
                  default_style: :product
-    validates :attachment, presence: true
-    validate :attachment_is_an_image
+
+    def supported_content_type
+      unless attachment.content_type.in?(Spree::Config.allowed_image_mime_types)
+        errors.add(:attachment, :content_type_not_supported)
+      end
+    end
   end
 end
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -445,6 +445,14 @@ module Spree
     # Enumerable of images adhering to the present_image_class interface
     class_name_attribute :image_attachment_module, default: 'Spree::Image::ActiveStorageAttachment'
 
+    # @!attribute [rw] allowed_image_mime_types
+    #
+    # Defines which MIME types are allowed for images
+    # `%w(image/jpeg image/jpg image/png image/gif).freeze` is the default.
+    #
+    # @return [Array]
+    class_name_attribute :allowed_image_mime_types, default: %w(image/jpeg image/jpg image/png image/gif).freeze
+
     # Allows switching attachment library for Taxon
     #
     # `Spree::Taxon::ActiveStorageAttachment`
--- a/core/spec/lib/search/base_spec.rb
+++ b/core/spec/lib/search/base_spec.rb
@@ -22,11 +22,12 @@ RSpec.describe Spree::Core::Search::Base do
 
   context "when include_images is included in the initialization params" do
     let(:params) { { include_images: true, keyword: @product1.name, taxon: @taxon.id } }
+    let(:image_file) { File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', 'blank.jpg')) }
     subject { described_class.new(params).retrieve_products }
 
     before do
-      @product1.master.images.create(attachment_file_name: "Test", position: 2)
-      @product1.master.images.create(attachment_file_name: "Test1", position: 1)
+      @product1.master.images.create(attachment_file_name: 'test', attachment: image_file, position: 2)
+      @product1.master.images.create(attachment_file_name: 'test1', attachment: image_file, position: 1)
       @product1.reload
     end
 
--- a/core/spec/lib/search/base_spec.rb
+++ b/core/spec/lib/search/base_spec.rb
@@ -4,6 +4,7 @@ require 'rails_helper'
 require 'spree/core/product_filters'
 
 RSpec.describe Spree::Core::Search::Base do
+  include ImageSpecHelper
   before do
     include Spree::Core::ProductFilters
     @taxon = create(:taxon, name: "Ruby on Rails")
@@ -22,7 +23,7 @@ RSpec.describe Spree::Core::Search::Base do
 
   context "when include_images is included in the initialization params" do
     let(:params) { { include_images: true, keyword: @product1.name, taxon: @taxon.id } }
-    let(:image_file) { File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', 'blank.jpg')) }
+    let(:image_file) { open_image('blank.jpg') }
     subject { described_class.new(params).retrieve_products }
 
     before do
--- a/core/spec/models/spree/image_spec.rb
+++ b/core/spec/models/spree/image_spec.rb
@@ -3,6 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe Spree::Image, type: :model do
+  include ImageSpecHelper
   it_behaves_like 'an attachment' do
     subject { create(:image) }
     let(:attachment_name) { :attachment }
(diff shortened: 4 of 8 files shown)


cpfergus1:
Failed on paperclip

core/spec/models/spree/image_spec.rb
```diff
@@ -9,6 +9,22 @@
     let(:default_style) { :product }
   end
 
+  describe 'validations' do
+    let(:valid_image_file) { File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', 'blank.jpg')) }
+    let(:invalid_image_file) { File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', 'file.txt')) }
+    let(:image) { create(:image) }
```
kennyadsl, line 15: I think that using this syntax the specs should work with both adapters:

`subject { create(:image, attachment: valid_image_file) }`

core/app/models/concerns/spree/active_storage_adapter.rb
```diff
@@ -107,6 +107,8 @@ def filename
 
     def url(style = default_style)
       attachment.url(style)
+    rescue ActiveStorage::FileNotFoundError
+      '/images/corrupted-image.png'
```
kennyadsl, line 111: This image is not provided by the system, and I think it shouldn't because the frontend and the backend have separate assets folders while this concern code would serve both of them. One alternative could be using a base64 string here with a standard broken image icon, but it's not ideal as well.

On a side note: what about removing this fix from this PR and opening another one just for the other issue?

cpfergus1: Removed temporary fix to be brought up in another branch

cpfergus1:
Removed NoFileError catch from this PR will upload that fix to another PR for further Discussion.

core/spec/models/spree/image_spec.rb
```diff
@@ -9,6 +9,26 @@
     let(:default_style) { :product }
   end
 
+  describe 'validations' do
```
kennyadsl, line 12: I feel like there's no need for the `describe` blocks here since we don't have any shared object between the test cases. What about just using two different tests as:

```ruby

it 'is valid when attachment has allowed content type' do
  image = build(:image, 
    attachment: File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', 'blank.jpg'))
  )
  
  expect(image).to be_valid
end

it 'is not valid when attachment has restricted content type' do
  image = build(:image, 
    attachment: File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', 'file.txt'))
  )
  
  expect(image).to_not be_valid
end
```

core/app/models/spree/image/active_storage_attachment.rb
```diff
@@ -3,10 +3,13 @@
 module Spree::Image::ActiveStorageAttachment
   extend ActiveSupport::Concern
   include Spree::ActiveStorageAdapter
-
```
kennyadsl, line 6: Why removing this line?

kennyadsl (commented):
Left a couple of suggestions. When done, can you please also squash all the commits into a single one for a better git history when the PR will get merged?

core/app/models/spree/image/active_storage_attachment.rb
cpfergus1: There was no reasoning for this, it was overlooked and has been returned to the original state.

core/spec/models/spree/image_spec.rb
cpfergus1: This syntax has been implemented!

kennyadsl (approved):
Thanks, @cpfergus1! 🎉

core/app/models/spree/image/active_storage_attachment.rb
```diff
@@ -15,7 +19,15 @@ module Spree::Image::ActiveStorageAttachment
...
-    validate :attachment_is_an_image
+
+    def supported_content_type
+      unless attachment.content_type.in?(accepted_image_types)
+        errors.add(:attachment, :content_type_not_supported)
+      end
+    end
+
+    def accepted_image_types
+      %w(image/jpeg image/jpg image/png image/gif)
```
waiting-for-dev, line 30: What do you think about having the list of accepted mime types in a constant?

core/spec/lib/search/base_spec.rb
```diff
@@ -22,11 +22,12 @@
 
   context "when include_images is included in the initialization params" do
     let(:params) { { include_images: true, keyword: @product1.name, taxon: @taxon.id } }
+    let(:image_file) { File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', 'blank.jpg')) }
```
waiting-for-dev, line 25: As we are using `File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', **)` in several place, what do you think about extracting it to a test helper method?

core/spec/models/spree/image_spec.rb
```diff
@@ -9,6 +9,18 @@
     let(:default_style) { :product }
   end
 
+  it 'is valid when attachment has allowed content type' do
+    image = build(:image,
+                  attachment: File.open(File.join('lib', 'spree', 'testing_support', 'fixtures', 'blank.jpg')))
```
waiting-for-dev, line 14: I like to put a blank line between Given/Then/When contexts in the specs... However, I think this is not consistently used in the codebase, so feel free to ignore the comment

waiting-for-dev (commented):
Good job @cpfergus1 . I think this validation is definitely the first step to have in place before attempting to fix edge cases :clap: :clap:

core/app/models/spree/image/active_storage_attachment.rb
cpfergus1: Defined in a ENV constant or where would the constant be defined?

core/spec/lib/search/base_spec.rb
cpfergus1: I think this is a great idea, I can certainly implement this now.

core/app/models/spree/image/active_storage_attachment.rb
waiting-for-dev: No, no... I mean as a constant in `Spree::Image::ActiveStorageAttachment`, like in:

```ruby
module Spree::Image::ActiveStorageAttachment
  ACCEPTED_IMAGE_TYPES = %w(....).freeze
end
```

I mean, it's information that doesn't depend on the state of an object, and it has no computational logic, so in my view, it belongs to the constants domain.

Sorry not having been clearer.

cpfergus1: I see, thank you for clarifying

core/spec/models/spree/image_spec.rb
cpfergus1: I am not noticing this in the rest of the examples, I think I will leave it as is to conform for the time being. Thanks for the suggestion!

core/app/models/spree/image/active_storage_attachment.rb
```diff
@@ -3,10 +3,15 @@
 module Spree::Image::ActiveStorageAttachment
   extend ActiveSupport::Concern
   include Spree::ActiveStorageAdapter
+  ACCEPTED_IMAGE_TYPES = %w(image/jpeg image/jpg image/png image/gif).freeze
```
kennyadsl, line 6: What about adding an application configuration for this so that it's easier to configure for stores that want to restrict the list further?

kennyadsl: On a side note: we can see also how Paperclip is restricting file types and use the same config for Paperclip as well.

cpfergus1: I replaced both %() in paperclipattachment and activestorage attachment with an not application configuration allowed_image_mime_types.

core/lib/spree/app_configuration.rb
```diff
@@ -444,7 +444,7 @@ def payment_canceller
     # @return [Module] a module that can be included into Spree::Image to allow attachments
     # Enumerable of images adhering to the present_image_class interface
     class_name_attribute :image_attachment_module, default: 'Spree::Image::ActiveStorageAttachment'
-
+    class_name_attribute :allowed_image_mime_types, default: %w(image/jpeg image/jpg image/png image/gif).freeze
```
kennyadsl, line 447: Can you please add some documentation to this configuration as the other ones?

cpfergus1: Implemented!

jarednorman (approved):

PR #4022, opened 2021-04-09
Add global Spree::Config.default_email_regexp

cesartalves:
This change adds a :default_email_regexp preference to be used by Spree::EmailValidator and also solidus_auth_devise, allowing the regex to be fully and easily customizable. With this introduction it will be possible to keep the Devise.email_regexp in sync with solidus's email validator regexp.

**Description**

**Checklist:**

### DIFF
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -150,6 +150,10 @@ module Spree
     #   @return [String] Two-letter ISO code of a {Spree::Country} to assumed as the country of an unidentified customer (default: "US")
     preference :default_country_iso, :string, default: 'US'
 
+    # @!attribute [rw] default_email_regexp
+    #   @return [Regexp] Regex to be used in email validations, for example in Spree::EmailValidator
+    preference :default_email_regexp, :regexp, default: URI::MailTo::EMAIL_REGEXP
+
     # @!attribute [rw] generate_api_key_for_all_roles
     #   @return [Boolean] Allow generating api key automatically for user
     #   at role_user creation for all roles. (default: +false+)
--- a/core/lib/spree/core/validators/email.rb
+++ b/core/lib/spree/core/validators/email.rb
@@ -13,9 +13,11 @@ module Spree
   #
   class EmailValidator < ActiveModel::EachValidator
     EMAIL_REGEXP = URI::MailTo::EMAIL_REGEXP
+    # Use Spree::Config.default_email_regexp instead
+    deprecate_constant :EMAIL_REGEXP
 
     def validate_each(record, attribute, value)
-      unless EMAIL_REGEXP.match? value
+      unless Spree::Config.default_email_regexp.match? value
         record.errors.add(attribute, :invalid, **{ value: value }.merge!(options))
       end
     end


core/lib/spree/app_configuration.rb
```diff
@@ -145,6 +145,10 @@ class AppConfiguration < Preferences::Configuration
     #   @return [String] Two-letter ISO code of a {Spree::Country} to assumed as the country of an unidentified customer (default: "US")
     preference :default_country_iso, :string, default: 'US'
 
+    # @!attribute [rw] default_email_regexp
+    #   @return [Regexp] Regex to be used in email validations, for example in Spree::EmailValidator
+    preference :default_email_regexp, :regexp, default: /\A([^@\.]|[^@\.]([^@\s]*)[^@\.])@([^@\s]+\.)+[^@\s]+\z/
```
softr8, line 150: Might be worth trying with Ruby's built in [URI::MailTo::EMAIL_REGEXP](https://github.com/ruby/ruby/blob/48b94b791997881929c739c64f95ac30f3fd0bb9/lib/uri/mailto.rb#L55)

core/lib/spree/core/validators/email.rb
```diff
@@ -12,10 +12,8 @@ module Spree
   #     end
   #
   class EmailValidator < ActiveModel::EachValidator
-    EMAIL_REGEXP = /\A([^@\.]|[^@\.]([^@\s]*)[^@\.])@([^@\s]+\.)+[^@\s]+\z/
```
kennyadsl, line 15: It's ok to remove code from Solidus core, but since this is public and could be used in stores, we should deprecate its usage instead, and remove it in a future release. See this useful [gist](https://gist.github.com/rafaelfranca/5492531) and our own [Spree::Deprecation class](https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/lib/spree/deprecation.rb#L6).

core/lib/spree/app_configuration.rb
cesartalves: This is just a copy of the regex in Spree::EmailValidator. That regex you mentioned is significantly different from it and I think changing it might have a potential impact for unaware upgraders.

core/lib/spree/core/validators/email.rb
cesartalves: That makes sense, thanks for pointing it out.

core/lib/spree/app_configuration.rb
nirebu: I think it might be worth considering an update of the regex because the current one has a couple of bugs.

[At this link](https://regex101.com/r/JTDNrI/4) you can see and play with different combinations (be aware of the usage of `^` and `$` instead of `\A` and `\z`, it was just to use multiline matches), but looks like that the current regex allows the following:

- `test @example.com`: single whitespace before the `@`
- ` test@example.com`: leading whitespace, although this is a smaller issue since Devise by default auto strips leading and trailing whitespaces on `email` fields. The `Spree::User` class in `solidus_auth_devise` is immune from this bug because of the auto stripping.

Since this would enter a deprecation cycle, would it be ok updating the regex to something like this? [Resulting in this example](https://regex101.com/r/3rC6us/1).

```ruby
/\A([^@\s.]|[^@\s.]([^@\s]*)[^@\s.])@([^@\s]+\.)+[^@\s]+\z/
```

@cesartalves , @kennyadsl , @softr8 thoughts?

cesartalves: I see no problem in updating the regex. We can let users know that it's gonna happen on a release note. Besides, shouldn't break a lot of validations save few edge cases (e.g., white spaces as you mentioned).

waiting-for-dev:
Hey! Any particular reason why we're closing it? We could handle the change in value thanks to the new Solidus update process. I.e.:

```ruby
preference :default_email_regexp, :regexp, default: by_version(/\A([^@\.]|[^@\.]([^@\s]*)[^@\.])@([^@\s]+\.)+[^@\s]+\z/, "3.2.alpha" => URI::MailTo::EMAIL_REGEXP)
```

Reopening to keep visibility on it.

waiting-for-dev:
Hey @cesartalves, would you be willing to retake it?

We should:

- Rebase from master
- Update the regular expression to use `URI::MailTo::REGEXP` (we updated it to get rid of security vulnerabilities, see https://github.com/solidusio/solidus/commit/ff33a8bdbeb66cd47625e6c05af1a10a26399834).
- Address @spaghetticode's concern.

waiting-for-dev:
Hey, I rebased it myself and implemented @spaghetticode suggestion. It should be ready for review.

spaghetticode (approved):
thank you @cesartalves and @waiting-for-dev !

Issue #3988, opened 2021-03-12
UnknownAttributeError creating a product via API with a Permalink


kennyadsl:
**Solidus Version**:
2.11.5, master

ActiveModel::UnknownAttributeError

**To Reproduce**

Trying to create a product with the `permalink` attribute included:

```
curl -si \
--header 'Authorization: Bearer <PLACE_API_KEY_HERE>' \
--header 'content-type: application/json' \
--request POST --url https://<PLACE_HOST_HERE>/api/products \
--data  '{"product": {"name":"Eraseme_2","permalink":null,"description":"To be erased.","meta_description":"","meta_keywords":"","price":"1000.0","sku":"ERS000001", "deleted_at":null,"option_values_hash":null,"weight":"0.0","height":null,"width":null,"depth":null,"shipping_category_id":1}}'
```

**Current behavior**

We get this 500 error.

```HTTP/1.1 500 Internal Server Error Content-Type: text/html; charset=UTF-8 X-Request-Id: 4c8e1aa1-8ce2-4622-91a0-30ea933abb1b X-Runtime: 0.243611 Content-Length: 129551
ActiveModel::UnknownAttributeError in Spree::Api::ProductsController#create
unknown attribute 'permalink' for Spree::Product. 
Extracted source (around line #52)
          public_send(setter, v)
        else
          raise UnknownAttributeError.new(self, k)
        end
      end
  end
```

**Expected behavior**

I think that request should be successful, so we should understand why this is happening in the first place and fix it. 

Alternatively, if the error is legit for some reason, we should not make the application explode but gracefully return an API error.

PR #4024, opened 2021-04-12
Handle permalink attribute on product create

nandita2010:
**Description**
An ActiveModel::UnknownAttributeError error is seen when a product is being created with a permalink attribute.

The reason for this error is because the request contains permalink key that Rails ActiveRecord is looking for in Products table. This was the case in the past but in a [commit](https://github.com/solidusio/solidus/commit/d4d955017102f31980cc3cf3c0ca7ad9f5b053bd), permalink attribute in products was renamed to slug and moved to taxons model. Ideally, permalink attribute should have been removed from permitted_attributes.rb under @@product_attributes array. Looks like that was missed.

This PR removes the attribute from product's permitted attributes and from the document

This PR takes care of issue #[3988](https://github.com/solidusio/solidus/issues/3988)

**Checklist:**
~~-[ ] I have updated Guides and README accordingly to this change (if needed)~~
~~- [ ] I have added tests to cover this change (if needed)~~

### DIFF
--- a/api/openapi/solidus-api.oas.yml
+++ b/api/openapi/solidus-api.oas.yml
@@ -6289,8 +6289,6 @@ components:
           type: string
         available_on:
           type: string
-        permalink:
-          type: string
         meta_description:
           type: string
         meta_keywords:
--- a/core/lib/spree/permitted_attributes.rb
+++ b/core/lib/spree/permitted_attributes.rb
@@ -68,7 +68,7 @@ module Spree
     @@product_properties_attributes = [:property_name, :value, :position]
 
     @@product_attributes = [
-      :name, :description, :available_on, :discontinue_on, :permalink, :meta_description,
+      :name, :description, :available_on, :discontinue_on, :meta_description,
       :meta_keywords, :price, :sku, :deleted_at,
       :option_values_hash, :weight, :height, :width, :depth,
       :shipping_category_id, :tax_category_id,


kennyadsl:
Thanks for your contribution, @nandita2010! 

I'm taking a closer look and couldn't figure out where we are actually using the permalink field for Spree::Product in Solidus. 

Can you help me understand if we really need it or it was just an [API documentation](https://solidus.docs.stoplight.io/api-reference/products/create-product) issue?

nandita2010:
You're right @kennyadsl 

Permalink attribute is not used anywhere in product. 
Reading the issue my understanding was that the client is sending a permalink attribute that was causing the call to fail. Introducing a setter solved this without functionally changing anything.

But, yes. If we can update the document and republish it to the clients so they do not pass the permalink attribute will do the job and avoid this error. My bad, I didn't know this was an option.

Could we consider using permit?  "params.require(: product).permit(:name, :description, :meta_description ...)" as an additional check on the controller. We can then use action_on_unpermitted_parameters to raise an error.

kennyadsl:
Thanks for getting back @nandita2010. I think we already use permitted attributes on the controller level, see [here](https://github.com/solidusio/solidus/blob/master/core/lib/spree/permitted_attributes.rb#L70). Not sure why permalink is there though. Worth exploring the git history and understand if we can just remove that attribute from the list. Also, removing it and see if some specs fail will give us some useful information.

nandita2010:
Thank you @kennyadsl for the pointer. Looks like permalink was an attribute in products which was later renamed to slug in product and moved to taxons model.  Ideally in commit d4d955017102f31980cc3cf3c0ca7ad9f5b053bd where all the related changes have been made, permalink attribute should have been removed from permitted_attributes.rb under @@product_attributes array. Looks like that was missed. 
I will undo the changes from previous commit and create a new commit with the attribute removed from here and ensure all rspecs pass with the new change.

nandita2010:
PS - Force-pushed only because I had not set the right git config for user.name and email for that commit.

jarednorman (commented):
This looks good to me. Could you rebase/squash this down so it's just one commit that makes the intended change. The initial change and revert can be dropped.

nandita2010:
Hi @jarednorman, please have a look at the updated commits.

jarednorman (approved):

nandita2010:
Hi @kennyadsl Please have a look at the PR and let me know the next steps.

kennyadsl (approved):
Thanks @nandita2010!

PR #4026, opened 2021-04-13
Patch for FileNotFoundError bug on failed image downloads.

cpfergus1:
**Description**
This PR aimed to solve the Active::Storage::FileNotFoundError bug described in Issue #4017 by to handling FileNotFound error when ActiveStorage fails to download from a remote service such as S3.

In rare circumstances, an image would fail to upload to a third party storage service, causing the application to hard crash when trying to retrieve an image resource. This commit implements a fix to return a place holder image to alert the user of an error and also prevents the application from crashing. This fix will not, however, save any errors associated with the failed upload such as credential errors. This fix was tested by artificially changing the blob record to force
a FileNotFoundError.

The handling of the FileNotFound error maybe a temporary fix as we are aiming to discover another pathway to handle the exception and remove the invalid objects simultaneously. 

image

**Checklist:**

### DIFF
--- a/backend/spec/features/admin/products/edit/images_spec.rb
+++ b/backend/spec/features/admin/products/edit/images_spec.rb
@@ -69,6 +69,21 @@ describe "Product Images", type: :feature do
       expect(page).not_to have_field "image[alt]", with: "ruby on rails t-shirt"
     end
 
+    context 'Using Active Storage',
+            if: Spree::Config.image_attachment_module == Spree::Image::ActiveStorageAttachment do
+      it "should show a no image placeholder when images fail to download" do
+        click_link "new_image_link"
+        within_fieldset 'New Image' do
+          attach_file('image_attachment', file_path)
+        end
+
+        click_button "Update"
+        Spree::Image.first.attachment.blob.update(key: 11)
+        visit current_path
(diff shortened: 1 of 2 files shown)


cpfergus1:
Failed on Paperclip

backend/spec/features/admin/products/edit/images_spec.rb
```diff
@@ -69,6 +69,18 @@
...
     end
 
+    it "should show a no image placeholder when images fail to download" do
+      click_link "new_image_link"
+      within_fieldset 'New Image' do
+        attach_file('image_attachment', file_path)
+      end
+
+      click_button "Update"
+      Spree::Image.first.attachment.blob.update(key: 11)
```
kennyadsl, line 79: Specs are failing in the CI in jobs where we do not use Active Storage but Paperclip. In that case, `attachment` doesn't respond to `blob` and it fails. 

I think we need to find a way to make this compatible with both or skip the test when we are using Paperclip.

cpfergus1: Skipping test based on which module is used.

kennyadsl (approved):

elia (approved):

jarednorman (approved):

PR #4028, opened 2021-04-14
Update install_generator.rb

iqre8:
Check if the solidus_auth_devise is already in the Gemfile to avoid adding it twice.

**Description**

**Checklist:**

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -134,7 +134,7 @@ module Solidus
 
         Would you like to install it? (y/n)"))
 
-        @plugins_to_be_installed << 'solidus_auth_devise'
+        @plugins_to_be_installed << 'solidus_auth_devise' unless File.open('Gemfile').grep(/gem 'solidus_auth_devise'/)
         @plugin_generators_to_run << 'solidus:auth:install'
       end
     end


core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -134,7 +134,7 @@ def install_auth_plugin
 
         Would you like to install it? (y/n)"))
 
-        @plugins_to_be_installed << 'solidus_auth_devise'
+        @plugins_to_be_installed << 'solidus_auth_devise' unless File.open('Gemfile').grep(/gem 'solidus_auth_devise'/)
```
kennyadsl, line 137: What about avoiding asking to install devise at all if already present?

jarednorman: In what situation would someone be running the install generator but already have `solidus_auth_devise` in their Gemfile?

iqre8: @jarednorman I was under the impression the install was need to create an admin user in my environment. I had already added the solidus_auth_devise  in my dev environment because thats where the install was run. in my acceptance env, I required a new admin account and thus ran the installer again. I assume I will need to do it again in the production env

jarednorman: No, that's not the case. That's done as part of loading the seeds, but the seeds can be loaded separately in environments where you need them. Once you've run the install generator in your codebase you shouldn't need to ever run it again.

iqre8: Well, I just wiped my db, ran just the migrations and the seed. Did not get a prompt for creds. So guessing there's a missing task somewhere? The default creds dont work either.

kennyadsl: I think there's a specific task for creating an admin user: https://github.com/solidusio/solidus_auth_devise/blob/master/lib/tasks/auth.rake#L4

kennyadsl:
Can we close this one? I think the issue has been fixed pointing out the other task coming from `solidus_auth_devise`, isn't it?

iqre8:
We can close this one.

PR #4029, opened 2021-04-18
Update variant.rb

iqre8:
Law of demeter violation.
Get issues since the model doesnt ensure option_value.option_type is not nil

**Description**

**Checklist:**

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -183,7 +183,7 @@ module Spree
     # @return [String] a sentence-ified string of option values.
     def options_text
       values = option_values.includes(:option_type).sort_by do |option_value|
-        option_value.option_type.position
+        option_value.option_type.position unless option_value.option_type.nil?
       end
 
       values.to_a.map! do |ov|
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -183,7 +183,7 @@ module Spree
     # @return [String] a sentence-ified string of option values.
     def options_text
       values = option_values.includes(:option_type).sort_by do |option_value|
-        option_value.option_type.position unless option_value.option_type.nil?
+        option_value.option_type&.position
       end
 
       values.to_a.map! do |ov|


iqre8 (commented):
Updated with safe navigation

core/app/models/spree/variant.rb
```diff
@@ -183,7 +183,7 @@ def is_backorderable?
     # @return [String] a sentence-ified string of option values.
     def options_text
       values = option_values.includes(:option_type).sort_by do |option_value|
-        option_value.option_type.position
+        option_value.option_type&.position
```
stem, line 186: Doesn't `sort_by` raise an error 'cause you're comparing `nil` with other values ?

I know we have a lot of these in our codebase where we need to fake a value to avoir that error. Maybe it's not with `sort_by` and just `sort` 🤷‍♂️  but better safe than sorry and ask.

kennyadsl:
@iqre8 can you please also share how did you add the option value without the option type? Just to be sure we don't have anything in the Admin UI that can lead to this.

iqre8:
Hi,

Not sure how I got this situation. I'm guessing its because of some incomplete csv file uploaded via solidus_importer. I would hope the model logic would prevent a situation like this though. Couldn't get the dev/test env to run to dig and see if I could help write some more tests to cover this.

core/app/models/spree/variant.rb
iqre8: @stem Based on the error message, the sort_by doesn't raise the error.

kennyadsl:
Got it, thanks for sharing. 

> Couldn't get the dev/test env to run to dig and see if I could help write some more tests to cover this.

Did you already read [these instructions](https://github.com/solidusio/solidus#developing-solidus)? Happy to help if you want to share what was the issue.

core/app/models/spree/variant.rb
kennyadsl: If you have an error message, it would be great to share it. If someone has the same message and google it, finding this issue could be useful.

waiting-for-dev:
Having an `option_type` without an `option_value` has been recently deprecated at https://github.com/solidusio/solidus/commit/3dcc3fae3075fff18f3bb216e2a9a29e3a42e128. Therefore, if we still find a place where that happens, we need to fix it there. Anyway, thanks for your collaboration @iqre8 :raised_hands:

PR #4034, opened 2021-04-21
Review install instructions in README and Guides

kennyadsl:
**Description**

For the sake of clarity, we can remove the install instructions for Solidus <= 2.10. It's very unlikely that someone will start a new application with that version now.

TODO:

- [ ] Remove Imagemagick/Paperclip paragraph? We default on ActiveStorage now.

### DIFF
--- a/guides/source/developers/getting-started/first-time-installation.html.md
+++ b/guides/source/developers/getting-started/first-time-installation.html.md
@@ -65,15 +65,17 @@ the JavaScript compiler shipped with Rails by default ([webpacker][webpacker]),
 which is not required for a sample Solidus store.
 You are still free to install and configure it in your Solidus store though.
 
-Once the new project has finished being created, we can open the project's newly
-created `Gemfile` in a text editor and add the required Solidus gems as new
-lines:
+Once the new project is created, you can run:
 
-```ruby
-gem 'solidus'
+```bash
+cd /path/to/your-solidus-project-name
+bundle add solidus
 ```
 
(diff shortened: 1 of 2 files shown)


guides/source/developers/getting-started/first-time-installation.html.md
```diff
@@ -113,37 +111,13 @@ After the gems have been successfully installed, you need to create the
...
-
-```bash
-bin/rails generate spree:install
-```
-
-[solidus-auth-devise]: https://github.com/solidusio/solidus_auth_devise
-
 ### Set the administrator username and password
 
 The `solidus:install` generator prompts you to configure the Solidus administrator
```
cpfergus1, line 123: I believe we might want to change this to:
```The `solidus:install` generator prompts you to configure the default authorization solution, default payment service, and the Solidus administrator```

cpfergus1 (approved):
Following these changes are easy to understand and make sense! I would agree that anyone deciding to use solidus today would most likely not back-port to a stable version less than 3.0.0 so the install guides for 2.10 and 2.11 certainly are not required.

guides/source/developers/getting-started/first-time-installation.html.md
kennyadsl: It's `authentication` and not `authorization` but this proposal makes perfect sense, thanks!

jarednorman (approved):

spaghetticode (approved):
LGTM, thanks @kennyadsl 👍

PR #4035, opened 2021-04-22
Load defaults for the latest Rails minor version in the dummy app

waiting-for-dev:
Before this change, the dummy app used in the test suite was not loading
any new default that Rails has added since 5.0.

However, we still use old defaults for three settings that make our test
suite fail. Each one of them will be worked out in individual commits.

See https://api.rubyonrails.org/classes/Rails/Application/Configuration.html#method-i-load_defaults

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -490,7 +490,10 @@ module Spree
         raise CannotRebuildShipments.new(I18n.t('spree.cannot_rebuild_shipments_shipments_not_pending'))
       else
         shipments.destroy_all
-        self.shipments = Spree::Config.stock.coordinator_class.new(self).shipments
+        # TODO: We can use `self.shipments#=` instead of `#push` when
+        # https://github.com/rails/rails/issues/42102 is fixed and we deprecate
+        # Rails versions with the bug
+        shipments.push(*Spree::Config.stock.coordinator_class.new(self).shipments)
       end
     end
 
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -30,7 +30,7 @@ module Spree
     attr_writer :rebuild_vat_prices
     include Spree::DefaultPrice
 
-    belongs_to :product, -> { with_discarded }, touch: true, class_name: 'Spree::Product', inverse_of: :variants, optional: false
+    belongs_to :product, -> { with_discarded }, touch: true, class_name: 'Spree::Product', inverse_of: :variants_including_master, optional: false
     belongs_to :tax_category, class_name: 'Spree::TaxCategory', optional: true
 
     delegate :name, :description, :slug, :available_on, :discontinue_on, :discontinued?,
--- a/backend/spec/features/admin/products/edit/images_spec.rb
+++ b/backend/spec/features/admin/products/edit/images_spec.rb
@@ -78,7 +78,7 @@ describe "Product Images", type: :feature do
         end
 
         click_button "Update"
-        Spree::Image.first.attachment.blob.update(key: 11)
+        invalidate_attachment(Spree::Image.first.attachment)
         visit current_path
         expect(page).to have_xpath("//img[contains(@src, 'assets/noimage/mini')]")
       end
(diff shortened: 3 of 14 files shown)


waiting-for-dev:
I have started removing the overridden defaults in this same PR. If you'd prefer it to be in another one, please tell me.

kennyadsl:
I think that removed the overridden defaults in this PR is ok.

core/app/models/spree/order.rb
```diff
@@ -490,7 +490,10 @@ def create_proposed_shipments
         raise CannotRebuildShipments.new(I18n.t('spree.cannot_rebuild_shipments_shipments_not_pending'))
       else
         shipments.destroy_all
-        self.shipments = Spree::Config.stock.coordinator_class.new(self).shipments
+        # TODO: We can use `#=` instead of `#push` when
+        # https://github.com/rails/rails/issues/42102 is fixed and we deprecate
+        # Rails versions with the bug
+        self.shipments.push(*Spree::Config.stock.coordinator_class.new(self).shipments)
```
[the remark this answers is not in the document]
kennyadsl: I think @houndci-bot may be right here, we don't need the `self`, isn't it?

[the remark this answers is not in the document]
waiting-for-dev: Yep, it's fixed now

waiting-for-dev:
OK! This was tough, but everything is :green_circle: now :slightly_smiling_face:

When reviewing this PR, I strongly recommend going commit by commit, and reading the commit message first, as I guess it can be hard to follow how the different changes are related otherwise.

kennyadsl (commented):
Stellar work @waiting-for-dev, thanks. Just one note: can we check that the other config used later in `core/lib/spree/testing_support/dummy_app.rb` are not overriding the defaults of some old versions of Rails? Maybe we can take advantage of the change and cleanup a little bit those configs?

jarednorman (approved):

aldesantis (approved):
Outstanding debugging work on this one @waiting-for-dev, thanks so much! 🙇

core/app/models/spree/variant.rb
```diff
@@ -30,7 +30,7 @@ class Variant < Spree::Base
     attr_writer :rebuild_vat_prices
     include Spree::DefaultPrice
 
-    belongs_to :product, -> { with_discarded }, touch: true, class_name: 'Spree::Product', inverse_of: :variants, optional: false
+    belongs_to :product, -> { with_discarded }, touch: true, class_name: 'Spree::Product', inverse_of: :variants_including_master, optional: false
```
spaghetticode, line 33: It seems to me that this change (and the one in product.rb) are not really related to the issue with `ActiveRecord.has_many_inverse`, but they've been an issue for a while. 

What about moving these changes into their own commit? I think it would give the fix more prominence and make the last commit a bit lighter.

kennyadsl: Ops, I merged too fast. Sorry, @spaghetticode! 🤦‍♂️  

I would have agreed with you for what it's worth. Maybe something we can take into account in the next PRs. Sorry again.

waiting-for-dev: Fixing that association solved two test failures:

https://github.com/solidusio/solidus/blob/69b9a0f74c850d1283f24854332036cfa69db071/api/spec/requests/spree/api/products_controller_spec.rb#L251

```
  1) Spree::Api::ProductsController as an admin creating a product creates with embedded variants
     Failure/Error: expect(variants.count).to eq(2)
     
       expected: 2
            got: 3
     
       (compared using ==)
```

https://github.com/solidusio/solidus/blob/69b9a0f74c850d1283f24854332036cfa69db071/core/spec/lib/spree/core/testing_support/factories/variant_factory_spec.rb#L25

```
  1) variant factory master variant builds a master variant properly
     Failure/Error: expect(product.variants).to be_empty
       expected `#<ActiveRecord::Associations::CollectionProxy []>.empty?` to be truthy, got false
```

I didn't debug further, though, as I saw that the change made sense.

Issue #4027, opened 2021-04-14
Extending API resources attributes in applications does not work in development anymore


kennyadsl:
See also: https://github.com/solidusio/solidus/discussions/3956. 

**Versions**:
2.9+, Rails 5.2+

**To Reproduce**

In a new store or in the sandbox add a new field to `Spree::Product`:

```sh
rails g migration AddLikesCountToSpreeProducts likes_count:integer
```

and add this to the `spree.rb` initializer:

```
# config/initializers/spree.rb

Spree::Api::ApiHelpers.product_attributes << :likes_count
```

It is something related to the new Rails Autoloading system that reloads the classes at each request, but the initializer is not reloaded, so the product_attributes is restored to the original value at each request. This means that when the classes are cached (in production for example), this should not happen.

**Current behavior**

```bash
$ rails c
2.6.6 :001 > Spree::Api::ApiHelpers.user_attributes
 => [:id, :email, :created_at, :updated_at]
```

**Expected behavior**

```bash
$ rails c
2.6.6 :001 > Spree::Api::ApiHelpers.user_attributes
 => [:id, :email, :created_at, :updated_at, :likes_count]
```

**Additional context**

This can be temporarily solved by opening the `ApiHelpers` module with

```ruby
# host_app/path/to/your_decorators/api_helpers_decorator.rb

Spree::Api::ApiHelpers.module_eval do
  user_attributes << :likes_count
end
```

But we really need a long-term solution. This is what I attempted, and seems to work:

```ruby
# solidus/api/app/helpers/spree/api/api_helpers.rb

# for each attribute:
def self.user_attributes
 Spree::Deprecation.warn("Please use Spree::Api::Config::user_attributes instead.")
 Spree::Api::Config.user_attributes
end
# We also need to deprecate accessing the ATTRIBUTES constant in this file.
```

```ruby
# solidus/api/lib/spree/api_configuration.rb

module Spree
  class ApiConfiguration < Preferences::Configuration
    # ..
    
    # for each ApiHelpers attribute:
    preference :user_attributes, :array, default: [:id, :email, :created_at, :updated_at]
  end
end
```

With this in the host app we can now just do:

```ruby
# host_app/config/initializers/spree.rb

Spree::Api::Config.configure do |config|
  config.user_attributes += [ :your_custom_attributes ]
end
```

- [ ] We also need to change the [edgeguides](https://edgeguides.solidus.io/customization/customizing-the-api#extending-existing-resources) accordingly.
[...]

PR #4039, opened 2021-04-26
Moving API attribute helpers to API config

snada:
Issue: https://github.com/solidusio/solidus/issues/4027

Edgeguides Pr: https://github.com/solidusio/edgeguides/pull/37

**Description**

As described in [this issue](https://github.com/solidusio/solidus/issues/4027) there is a problem with the way data is autoloaded when pushing new attributes in an initalizer.

A proposed change was to move the attributes location from the helper module to the api config class.

This PR follows this approach.

Albeit this PR does not change existing behavior, it introduces a deprecation when directly accessing attributes from the static helper methods.

To take advantage of the autoload fix, attributes in initializers should be set this way:

```ruby
Spree::Api::Config.configure do |config|
  config.user_attributes += [ :your_custom_attributes ]
end
```

Edgeguides has been updated accordingly.

**Checklist:**

### DIFF
--- a/api/app/helpers/spree/api/api_helpers.rb
+++ b/api/app/helpers/spree/api/api_helpers.rb
@@ -37,7 +37,16 @@ module Spree
         :variant_property_attributes
       ]
 
-      mattr_reader(*ATTRIBUTES)
+      ATTRIBUTES.each do |attribute|
+        define_method attribute do
+          Spree::Api::Config.send(attribute)
+        end
+
+        define_singleton_method attribute do
+          Spree::Deprecation.warn("Please use Spree::Api::Config::#{attribute} instead.")
+          Spree::Api::Config.send(attribute)
+        end
+      end
 
       def required_fields_for(model)
         required_fields = model._validators.select do |_field, validations|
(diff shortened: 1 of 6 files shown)


api/app/helpers/spree/api/api_helpers.rb
```diff
@@ -37,7 +37,16 @@ module ApiHelpers
...
-      mattr_reader(*ATTRIBUTES)
+      ATTRIBUTES.each do |attribute|
+        define_method attribute do
+          Spree::Api::Config.send(attribute)
+        end
+
+        define_singleton_method attribute do
+          Spree::Deprecation.warn("Please use Spree::Api::Config::#{attribute} instead.")
+          Spree::Api::Config.send(attribute)
+        end
```
kennyadsl, line 48: I'm not sure to get why we need both `define_method` and `define_singleton_method` here. Can you please help me with some details?

kennyadsl (commented):
Thanks @snada, looks promising! 

I left a question, and I'd love to see a spec that actually describes the deprecated behavior as well. For example testing that accessing/setting the attribute in the old way still works raising a deprecation warning. Here's [an example at an old ref](https://github.com/nebulab/solidus/blob/b62ced66be2eb8eec8557e8e08a546dbfc17c6fa/core/spec/lib/spree/core/controller_helpers/strong_parameters_spec.rb#L22-L28) where we do something similar. It is no more in the main branch because we recently released Solidus 3.0 and removed deprecated code (and associated specs). What do you think?

api/app/helpers/spree/api/api_helpers.rb
snada: Sure. Methods defined with `define_method` are the ones used in views. Remove that definition and you'll see every test rendering something using these methods fail.

The singleton methods are the ones called when you reference to the helper directly, like so:

```
2.7.0 :015 > Spree::Api::ApiHelpers.user_attributes
DEPRECATION WARNING: Please use Spree::Api::Config::user_attributes instead. (called from irb_binding at (irb):15)
 => [:id, :email, :created_at, :updated_at, :likes_count]
```

You can clearly see the deprecation warning. Without that definition every piece of code currently using will break (like some of our tests, that I changed in this PR). With this code you will get the deprecation warning.

Plus, the very definition of [`mattr_reader`](https://apidock.com/rails/Module/mattr_reader) (previously used to generate these methods) explains how it creates both a class and an instance reader for the provided symbol.

I just pushed a commit with tests regarding the deprecation warning.

kennyadsl (approved):
👍 Thanks!

jarednorman (approved):

PR #4047, opened 2021-05-05
Update defaults in dummy application

waiting-for-dev:
Here it's a description of the reason for each change. Keep in mind that
CI test against Rails 5.2 as the minimal version:

- serve_static_assets: Removed it. It's [not available since Rails
5.0](https://guides.rubyonrails.org/5_0_release_notes.html#railties-removals).

- public_file_server.enabled: Set explicitly to `true`. Even if it was
already `true`, a [real Rails application will default it to `false` in
the production environment unless a env variable
`RAILS_SERVE_STATIC_FILES` is
present](https://github.com/rails/rails/blob/f95c0b7e96eb36bc3efc0c5beffbb9e84ea664e4/railties/lib/rails/generators/rails/app/templates/config/environments/production.rb.tt#L27). This way it's less surprising.

- whiny_nils: Removed. It's [not available since Rails
4.1](https://guides.rubyonrails.org/4_1_release_notes.html#railties-removals).

- action_controller.allow_forgery_protection: Set to true to mimic
  production.

- action_controller.default_protect_from_forgery: Set to true to mimic
  production.

- action_controller.perform_caching: Set to true to mimic production.

- active_support.deprecation: Remove duplicated entry.

- action_mailer.delivery_job: Removed because `load_defaults` take care of
it.

- storage_path: Removed. It wasn't a Rails setting, but used as a
convenience. It was confusing.

- action_controller.include_all_helpers: Removed as it's already the
default.

- log_level: Added. It already defaulted to `:debug`, but a real Rails app
will default to `:info` in production. To make things less surprising
let's be explicit.

- action_mailer.perform_caching: Added to mimic production.

- i18n.fallbacks: Added to mimic production.

- active_record.dump_schema_after_migration: Added to mimic production and
for better performance.

- assets: No need to check whether it's a method on `config`.

See #4035

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -47,35 +47,47 @@ module DummyApp
 
   class Application < ::Rails::Application
     config.load_defaults("#{Rails::VERSION::MAJOR}.#{Rails::VERSION::MINOR}")
-    config.eager_load = false
+    # Make the test environment more production-like:
     config.cache_classes = true
-    config.cache_store = :memory_store
-    config.serve_static_assets = true
-    config.public_file_server.headers = { 'Cache-Control' => 'public, max-age=3600' }
-    config.whiny_nils = true
-    config.consider_all_requests_local = true
     config.action_controller.allow_forgery_protection = false
     config.action_controller.default_protect_from_forgery = false
-    config.action_controller.perform_caching = false
+    config.action_controller.perform_caching = true
+    config.action_mailer.perform_caching = false
+    config.i18n.fallbacks = true
+
+    # Make debugging easier:
+    config.consider_all_requests_local = true
     config.action_dispatch.show_exceptions = false
     config.active_support.deprecation = :stderr
+    config.log_level = :debug
+
+    # Improve test suite performance:
+    config.eager_load = false
+    config.public_file_server.headers = { 'Cache-Control' => 'public, max-age=3600' }
+    config.cache_store = :memory_store
+
+    # We don't use a web server, so we let Rails serve assets.
+    config.public_file_server.enabled = true
+
+    # We don't want to send email in the test environment.
     config.action_mailer.delivery_method = :test
-    config.active_support.deprecation = :stderr
+
+    # No need to use credentials file in a test environment.
     config.secret_key_base = 'SECRET_TOKEN'
 
-    config.action_mailer.delivery_job = "ActionMailer::MailDeliveryJob" if RAILS_6_OR_ABOVE
+    # Set the preview path within the dummy app:
     config.action_mailer.preview_path = File.expand_path('dummy_app/mailer_previews', __dir__)
-    config.action_mailer.show_previews = true
+
     config.active_record.sqlite3.represent_boolean_as_integer = true unless RAILS_6_OR_ABOVE
+    config.active_record.dump_schema_after_migration = false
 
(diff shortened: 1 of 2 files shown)


waiting-for-dev:
Hmm. Some tests failing on CI but not on my local machine. I'll try to figure it out.

waiting-for-dev:
I think it'll be green once #4048 is merged

kennyadsl (approved):
Thanks, @waiting-for-dev.

core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -47,35 +47,33 @@ def self.setup(gem_root:, lib_name:, auto_migrate: true)
...
+    config.action_controller.default_protect_from_forgery = false # mimic production
+    config.action_controller.perform_caching = true # mimim production
+    config.action_dispatch.show_exceptions = false # makes debugging easier
+    config.active_support.deprecation = :stderr # makes debugging easier
+    config.log_level = :debug # makes debugging easier
+    config.action_mailer.delivery_method = :test # don't try to send mails
+    config.secret_key_base = 'SECRET_TOKEN' # no need to use credentials file
+    config.action_mailer.preview_path = File.expand_path('dummy_app/mailer_previews', __dir__) # preview path within the dummy app
+    config.action_mailer.perform_caching = false # mimic production
+    config.i18n.fallbacks = true # mimic production
```
jarednorman, line 66: There's a lot of duplication in the reasoning for these. I also don't think appending all these comments to these lines is very readable. What if we grouped them by the reasoning? Something like:

```suggestion
    # Make the test environment more production-like:
    config.cache_classes = true
    config.action_controller.allow_forgery_protection = false
    config.action_controller.default_protect_from_forgery = false
    config.action_controller.perform_caching = true
    config.action_mailer.perform_caching = false
    config.i18n.fallbacks = true

    # Make debugging easier:
    config.consider_all_requests_local = true
    config.action_dispatch.show_exceptions = false
    config.active_support.deprecation = :stderr
    config.log_level = :debug

    # Improve test suite performance:
    config.eager_load = false
    config.public_file_server.headers = { 'Cache-Control' => 'public, max-age=3600' }
    config.cache_store = :memory_store

    # We don't use a web server, so we let Rails serve assets.
    config.public_file_server.enabled = true

    # We don't want to send email in the test environment.
    config.action_mailer.delivery_method = :test

    # No need to use credentials file in a test environment.
    config.secret_key_base = 'SECRET_TOKEN'

    # Set the preview path within the dummy app:
    config.action_mailer.preview_path = File.expand_path('dummy_app/mailer_previews', __dir__)
```

jarednorman (changes requested):

core/lib/spree/testing_support/dummy_app.rb
waiting-for-dev: I preferred to group by scope, but we can do either way

jarednorman (approved):

kennyadsl:
@waiting-for-dev did we remove the grouping suggested by Jared with the last commit?

waiting-for-dev:
> @waiting-for-dev did we remove the grouping suggested by Jared with the last commit?

Thanks for catching it @kennyadsl . As I directly committed the suggestion but didn't update my local repo, I override when forced the push :see_no_evil: Sorry for that :disappointed: 

Just fixed it.

kennyadsl:
No problem, we are here to review!

jarednorman (approved):

Issue #3010, opened 2018-12-19
RFC: Change *_decorator pattern


kennyadsl:
The Solidus suggested way to extend Ruby classes coming from core and extensions into users' applications is by creating files that end with `_decorator.rb` into the `/app` folder of the main application.

Despite this is not mandatory and users can change this pattern, this is suggested by us with the [install generator](https://github.com/solidusio/solidus/blob/4e293cd8b051ee1d01407b2ffe9587b6d29a9da3/core/lib/generators/spree/install/install_generator.rb#L84-L88) that pushes this code into `config/application.rb`

```ruby
    config.to_prepare do
      # Load application's model / class decorators
      Dir.glob(File.join(File.dirname(__FILE__), "../app/**/*_decorator*.rb")) do |c|
        Rails.configuration.cache_classes ? require(c) : load(c)
      end
```

I think this is not a good pattern and we should stop using/promoting it:
- it is not clear
- it is not really related to the [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) that (I think) it was originally referring to
- it may conflict with naming used by other gems, Draper for example. 

I hope this issue helps to answer these questions:

- Does the community agree that we should change it?
- What is a good pattern/naming convention instead? Some proposals have been already done (https://github.com/solidusio/solidus/pull/3008#issuecomment-448587039)
- Does switching to the new pattern breaks backward compatibility or is it something we should do in a major release (maybe 3.0)?

jacobherrington:
We can deprecate the `*_decorator.rb` pattern now with a warning and add support for using `*_override.rb` (or whatever), then remove support for the old pattern in 4.0? Is that too long term, or does it make sense? 

That's assuming that the 3.0 (https://github.com/solidusio/solidus/milestone/16) isn't too far off.

elia:
This is relevant doc from Rails: https://guides.rubyonrails.org/engines.html#a-note-on-decorators-and-loading-code

Not sure if this habit came first on Rails or on Spree, but it might be good to adhere to whatever standard Rails has (except for extremely good reasons).

kennyadsl:
@elia didn't know that, great find! I think that could be a good reason to keep it as is then.

jacobherrington:
If we keep it the way it is, let's document why we do that here: https://github.com/solidusio/solidus/blob/master/guides/source/developers/extensions/decorators.html.md

elia:
[...]

PR #4054, opened 2021-05-10
Updated decorator pathing for new applications

cpfergus1:
**Description**
The current functionality of finding decorators was by pulling files with the 'decorator' suffix in a file name. This seems redundant and forces modules to end with "Decorator." It makes more sense that decorators should be discovered based on their inclusion in the "app/decorators" sub-folder of the user's application. Updated guides to reflect name spacing changes.

**Checklist:**

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -102,16 +102,20 @@ module Solidus
       empty_directory "app/overrides"
     end
 
+    def create_decorators_directory
+      empty_directory "app/decorators"
+    end
+
     def configure_application
       application <<-RUBY
-    # Load application's model / class decorators
-    initializer 'spree.decorators' do |app|
-      config.to_prepare do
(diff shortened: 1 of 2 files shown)


aldesantis (approved):
Thanks @cpfergus1, this makes sense, and it seems to me like there are no backwards-compatibility issues given that the code lives in an initializer.

tvdeyen (approved):

cpfergus1:
I believe there will be a change from decorators to overrides as well to match with rails conventions. I would say don't merge this PR quite yet
cc @kennyadsl

waiting-for-dev:
> I believe there will be a change from decorators to overrides as well to match with rails conventions.

Some context: https://github.com/solidusio/solidus/issues/3010

waiting-for-dev:
We stopped recommending this pattern (and adding the autoloading code) in https://github.com/solidusio/solidus/pull/4231. See there for the rationale behind it. Anyway, thanks, @cpfergus1, for your work :heart:

Issue #4055, opened 2021-05-11
TypeError: Cannot read property 'version' of undefined


Ruk33:
Unable to complete installation steps.

**Solidus Version**:

3.0.1

**To Reproduce**

1. Follow installation steps
2. Start rails server
3. Go to localhost:3000

**Current behavior**

The following error can be seen on the page

Showing /usr/local/bundle/gems/solidus_frontend-3.0.1/app/views/spree/shared/_head.html.erb where line #7 raised:

TypeError: Cannot read property 'version' of undefined
Trace of template inclusion: /usr/local/bundle/gems/solidus_frontend-3.0.1views/spree/layouts/spree_application.html.erb

Rails.root: /app

**Expected behavior**

The app works properly

**Screenshots**

![image](https://user-images.githubusercontent.com/1719111/117749151-a80e4000-b1e7-11eb-9f45-e682942af7e8.png)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows using WSL and Docker
 - Browser [e.g. chrome, safari] Brave
 - Version [e.g. 22] V1.24.82

**Additional context**

- Tried my own custom docker version
- Tried using plain ruby and rails
- Tried with rails 5 and 6
- Tried using the docker version from master

In all scenarios, I get the same error.

jlacaciom:
I have the same problem :(, i try in two different macbook and nothing. :(

Ruk33:
As an additional note, if the line https://github.com/solidusio/solidus/blob/master/frontend/app/views/spree/shared/_head.html.erb#L8 gets commented out, the page loads correctly (without styles but at least it works)

Ruk33:
Hello @jarednorman , yes I do

```
➜  ~ node -v
v15.4.0
➜  ~ npm -v
7.0.15
``` 

As a note, I also tried the dockerfile you guys have on the project and that didn't worked either (I used the sandbox application)

jlacaciom:
Yes i have v14.16.1 and other mac v14.15.4 but still the error persist.

benjaminwil:
Hello, I saw this same error yesterday.

It seems to be an incompatibility with the new version of `execjs` and Ruby <  2.7.

See https://github.com/rails/execjs/issues/99.

For me, the short-term solution is to lock to an older version of `execjs`

jarednorman:
Either upgrade Ruby or downgrade execjs if you hit this issue. https://github.com/rails/execjs/issues/99#issuecomment-837369385

[...]

PR #4056, opened 2021-05-11
Fix Spree::Promotion.has_actions scope

mamhoff:
The scope would return multiple copies of the same promotion without
this change.

I hope this bug does not result in duplicate promotions being applied to orders. The bug is also present in 2.9.11, and might warrant a backport.

**Checklist:**


jarednorman (approved):

kennyadsl (approved):

kennyadsl:
CI is red due to https://github.com/solidusio/solidus/issues/4055.

PR #4060, opened 2021-05-13
Add UUID to StoreCredit#generate_authorization_code

### LINKED ISSUE #4041: Prevent store credits `authorization_code` duplicates with UUID
The [current implementation](https://github.com/solidusio/solidus/blob/2ea829645b00fcedd5bfd69e045bddab7f40beb9/core/app/models/spree/store_credit.rb#L152
) of `Spree::StoreCredit#generate_authorization_code` may generate duplicates since the code is composed of a timestamp string. While in normal operations this is probably never the case, still it's potentially possible, i.e. when creating records concurrently in batches. One easy way to see the problem in action is with `Timecop.freeze`. 

I think that replacing the timestamp component with a UUID or adding the UUID to the current implementation without removing the timestamp would be a more robust solution. Since Ruby standard library includes `securerandom`, adding `SecureRandom.uuid` would cause no meaningful overhead to applications.

spaghetticode:
Fixes #4041 

The value returned by `StoreCredit#generate_authorization_code` depends on time, which means that it may not be unique when a store credit is saved more than once at the same moment. This is not very likely to happen, but not impossible at all.

By introducing a UUID into the code we can be sure that this code will end up being unique. The previous format of the code is retained, since there may be applications relying on it.

**Checklist:**

### DIFF
--- a/core/app/models/spree/store_credit.rb
+++ b/core/app/models/spree/store_credit.rb
@@ -150,7 +150,12 @@ class Spree::StoreCredit < Spree::PaymentSource
   end
 
   def generate_authorization_code
-    "#{id}-SC-#{Time.current.utc.strftime('%Y%m%d%H%M%S%6N')}"
+    [
+      id,
+      'SC',
+      Time.current.utc.strftime('%Y%m%d%H%M%S%6N'),
+      SecureRandom.uuid
+    ].join('-')
   end
 
   def editable?
--- a/core/spec/models/spree/store_credit_spec.rb
+++ b/core/spec/models/spree/store_credit_spec.rb
@@ -897,4 +897,12 @@ RSpec.describe Spree::StoreCredit do
       end
     end
   end
+
+  describe "#generate_authorization_code" do
+    it "doesn't rely on time for uniqueness" do
+      freeze_time do
+        expect(subject.generate_authorization_code).not_to eq(subject.generate_authorization_code)
+      end
+    end
+  end
 end


core/app/models/spree/store_credit.rb
```diff
@@ -150,7 +150,7 @@ def can_void?(payment)
   end
 
   def generate_authorization_code
-    "#{id}-SC-#{Time.current.utc.strftime('%Y%m%d%H%M%S%6N')}"
+    "#{id}-SC-#{Time.current.utc.strftime('%Y%m%d%H%M%S%6N')}-#{SecureRandom.uuid}"
```
jarednorman, line 153: This has enough segments in it now that it might be nicer to write it more like this, but either way is fine by me.

```suggestion
    [
      id,
      "SC",
      Time.current.utc.strftime('%Y%m%d%H%M%S%6N'),
      SecureRandom.uuid
    ].join("-")
```

jarednorman (approved):

core/spec/models/spree/store_credit_spec.rb
```diff
@@ -897,4 +897,11 @@
       end
     end
   end
+
+  describe "#generate_authorization_code" do
+    it "doesn't rely on time for uniqueness" do
+      expect(Time).to receive(:current).twice { Time.zone.parse("2019/09/05 06:30:00") }
```
aldesantis, line 903: Should we use the Rails helpers for stubbing time here, so the test is not so closely coupled with the implementation?

core/app/models/spree/store_credit.rb
```diff
@@ -150,7 +150,12 @@ def can_void?(payment)
   end
 
   def generate_authorization_code
-    "#{id}-SC-#{Time.current.utc.strftime('%Y%m%d%H%M%S%6N')}"
+    [
+      id,
+      'SC',
+      Time.current.utc.strftime('%Y%m%d%H%M%S%6N'),
+      SecureRandom.uuid
```
aldesantis, line 157: Is the code displayed anywhere? Adding the UUID at the end will make it extreeeemely long. Should we be concerned about that?

aldesantis (changes requested):

core/app/models/spree/store_credit.rb
spaghetticode: I don't think it's displayed anywhere. It's used internally when creating store credit events.

core/spec/models/spree/store_credit_spec.rb
spaghetticode: I actually wanted to explicitly couple it with the implementation, because the problem is with it. The moment we stop using the timestamp, this specific spec implementation may not make much sense anymore (I mean the line after this one), surely it won't make sense its text description.

aldesantis: What I meant is that there's a ton of different ways to get the current time in Ruby/Rails: `Time.current`, `Time.now`, `Time.zone.now`, `DateTime.current`, etc. In the test, we shouldn't worry about what method is being called exactly in the implementation.

A quick way to solve the problem would be to use the [helpers](https://api.rubyonrails.org/v5.2.5/classes/ActiveSupport/Testing/TimeHelpers.html) Rails provides for stubbing time.

Does that make sense?

spaghetticode: Oh, sure! I think I knew about those helpers at a certain point in my life, but I forgot about them since then.

jarednorman (approved):

aldesantis (approved):

PR #4062, opened 2021-05-13
Add preferences to configure product and taxon images style

cpfergus1:
This commit adds Spree configuration options for styles, which will make
it easier to customize product and taxon image styles. Currently, the
user would have to create an override file or supersede the current
modules ActiveStorageAttachment for either class. This will make editing
styles clean and a single step process by allowing users to edit their
spree.rb configuration to modify these settings.

**Description**

**Checklist:**

### DIFF
--- a/core/app/models/spree/image/active_storage_attachment.rb
+++ b/core/app/models/spree/image/active_storage_attachment.rb
@@ -12,13 +12,8 @@ module Spree::Image::ActiveStorageAttachment
     validate :supported_content_type
 
     has_attachment :attachment,
-                   styles: {
-                   mini: '48x48>',
-                   small: '400x400>',
-                   product: '680x680>',
-                   large: '1200x1200>'
-                 },
-                 default_style: :product
+                   styles: Spree::Config.product_image_styles,
+                   default_style: Spree::Config.product_image_style_default
 
     def supported_content_type
       unless attachment.content_type.in?(Spree::Config.allowed_image_mime_types)
--- a/core/app/models/spree/taxon/paperclip_attachment.rb
+++ b/core/app/models/spree/taxon/paperclip_attachment.rb
@@ -5,14 +5,14 @@ module Spree::Taxon::PaperclipAttachment
 
   included do
     has_attached_file :icon,
-      styles: { mini: '32x32>', normal: '128x128>' },
-      default_style: :mini,
+      styles: Spree::Config.taxon_image_styles,
+      default_style: Spree::Config.taxon_image_style_default,
       url: '/spree/taxons/:id/:style/:basename.:extension',
       path: ':rails_root/public/spree/taxons/:id/:style/:basename.:extension',
       default_url: '/assets/default_taxon.png'
 
     validates_attachment :icon,
-      content_type: { content_type: %w[image/jpg image/jpeg image/png image/gif] }
+      content_type: { content_type: Spree::Config.allowed_image_mime_types }
   end
 
   def icon_present?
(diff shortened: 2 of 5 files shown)


core/app/models/spree/image/active_storage_attachment.rb
```diff
@@ -12,13 +12,8 @@ module Spree::Image::ActiveStorageAttachment
...
 
     has_attachment :attachment,
-                   styles: {
-                   mini: '48x48>',
-                   small: '400x400>',
-                   product: '680x680>',
-                   large: '1200x1200>'
-                 },
-                 default_style: :product
+                   styles: Spree::Config.product_image_styles,
```
spaghetticode, line 15: I'm not sure about the naming here. I think this is a generic `image` attachment, so maybe the styles should be simply `Spree::Config.image_styles`. What do you think?

core/app/models/spree/taxon/paperclip_attachment.rb
```diff
@@ -5,14 +5,14 @@ module Spree::Taxon::PaperclipAttachment
...
-      default_url: '/assets/default_taxon.png'
+                      styles: Spree::Config.taxon_image_styles,
+                      default_style: Spree::Config.taxon_image_style_default,
+                      url: '/spree/taxons/:id/:style/:basename.:extension',
+                      path: ':rails_root/public/spree/taxons/:id/:style/:basename.:extension',
+                      default_url: '/assets/default_taxon.png'
 
     validates_attachment :icon,
-      content_type: { content_type: %w[image/jpg image/jpeg image/png image/gif] }
+                         content_type: { content_type: %w[image/jpg image/jpeg image/png image/gif] }
```
spaghetticode, line 15: I think we can replace the array with  `Spree::Config.allowed_image_mime_types` here

spaghetticode (commented):
@cpfergus1 thanks for this PR, I like the new configurations for image styles 🖼️ 
I added a comment for a small optimization, let me know what you think!

core/app/models/spree/image/active_storage_attachment.rb
spaghetticode: Please disregard my comment above. I typed it but then I removed it, at least that's what I thought. While I think the naming for this class and the configurations may not be perfect, I think it's good enough.

core/app/models/spree/taxon/paperclip_attachment.rb
cpfergus1: Thanks for catching this, I agree that it should have been changed

core/app/models/spree/image/active_storage_attachment.rb
cpfergus1: I had difficulties on it too, I felt I had to make a clear distinction from the taxon config and the product-image config

core/app/models/spree/taxon/paperclip_attachment.rb
cpfergus1: Apparently it is breaking the application, I am not sure why yet

spaghetticode: I didn't try it, but the correct format should be  `content_type: { content_type: Spree::Config.allowed_image_mime_types }`

kennyadsl (commented):
Looks good. Can we just remove formatting changes like indentations? I know the style is not consistent but I think a clean git history is more important here.

cpfergus1:
> Looks good. Can we just remove formatting changes like indentations? I know the style is not consistent but I think a clean git history is more important here.

done!

kennyadsl (approved):
Thanks, Connor!

jarednorman (approved):

PR #4064, opened 2021-05-17
Allow using different preference defaults depending on a Solidus version

waiting-for-dev:
This feature is similar to [Rail's
`load_defaults`](https://guides.rubyonrails.org/configuring.html#results-of-config-load-defaults).
However, instead of loading imperatively, the defaults for a given
version keep the history of the value in the preference declaration.
Both systems are upgrade-friendly, as users need to adjust the new
version defaults manually. The main advantage of this commit
implementation is communication (as the code tells about its current
state and history). However, it's architecturally more complex.

`Spree::Core::VersionedValue` implements the core behavior. This class
accepts a specification of how a value has changed in time and returns
the result for a given solidus version. The specification consists of an
initial value and a series of boundaries when it changed:

```ruby
value = Spree::Core::VersionedValue.new(false, "3.0" => true)
value.call("2.0") # => false
value.call("3.0") # => true
```

`Spree::AppConfiguration` bundles the behavior into the preferences
system. Its `.by_version` method builds a `Proc` that accepts a solidus
version and returns the corresponding value. It's meant to be used in
the `default:` keyword argument:

```ruby
preference :foo, :boolean, default: by_version(false, "3.0" => false)
```

Accordingly, `Spree::Preferences::Preferable` has been modified to
provide arguments to the `default:` `Proc`. As this module is not only
included in `Spree::AppConfiguration`, it defaults to supply no
arguments to it, but `Spree::AppConfiguration` overrides it to add a
`#loaded_defaults` attribute.

Users can specify the version defaults they want to use through the
`#load_defaults` method in `Spree::AppConfiguration` (which gets bound
in the `Spree.config` method used in the `spree` initializer):

```ruby
Spree.config do |config|
  config.load_defaults '3.1'
end
```

We have modified the `spree` generator template to add that line of code
defaulting to the current Solidus version. As the sibling step, we
should create an update task creating another initializer à la
`new_framework_defaults` in Rails.

The tricky part in the rollout of this new feature could be the first
upgrade after it, as current Solidus installations don't know about
`load_defaults`. Therefore, an upgrade will leave them with the new
defaults unless they manually add the `load_defaults` line.

Related to discussion: https://github.com/solidusio/solidus/discussions/4045

**Checklist:**

### DIFF
--- a/core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
+++ b/core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
@@ -4,6 +4,9 @@
 Spree.config do |config|
   # Core:
 
+  # Solidus version defaults for preferences that are not overridden
+  config.load_defaults '<%= Spree.solidus_version %>'
+
   # Default currency for new sites
   config.currency = "USD"
 
@@ -62,12 +65,14 @@ end
 
 <% if defined?(Spree::Frontend::Engine) -%>
 Spree::Frontend::Config.configure do |config|
+  config.load_defaults '<%= Spree.solidus_version %>'
   config.locale = 'en'
 end
 <% end -%>
 
 <% if defined?(Spree::Backend::Engine) -%>
 Spree::Backend::Config.configure do |config|
+  config.load_defaults '<%= Spree.solidus_version %>'
   config.locale = 'en'
 
   # Uncomment and change the following configuration if you want to add
@@ -83,6 +88,7 @@ end
 
 <% if defined?(Spree::Api::Engine) -%>
 Spree::Api::Config.configure do |config|
+  config.load_defaults '<%= Spree.solidus_version %>'
   config.requires_authentication = true
 end
 <% end -%>
--- a/core/lib/spree/core/versioned_value.rb
+++ b/core/lib/spree/core/versioned_value.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+module Spree
+  module Core
+    # Wrapper for a value that can be different depending on the Solidus version
+    #
+    # Some configuration defaults can be added or changed when a new Solidus
+    # version is released. This class encapsulates getting the correct value for a
+    # given Solidus version.
+    #
+    # The way it works is you provide an initial value in time, plus the version
+    # boundary where it got changed. Then you can fetch the value providing the
+    # desired Solidus version:
+    #
+    # @example
+    #   value = VersionedValue.new(true, "3.0.0" => false)
+    #   value.call("2.7.0") # => true
+    #   value.call("3.0.0") # => false
+    #   value.call("3.1.0") # => false
+    #
+    # Remember that you must provide the exact boundary when a value got changed,
+    # which could easily be during a pre-release:
+    #
+    # @example
+    #   value = VersionedValue.new(true, "3.0.0" => false)
+    #   value.call("3.0.0.alpha") # => true
+    #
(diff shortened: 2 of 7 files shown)


aldesantis (approved):
@waiting-for-dev this is really impressive, thanks! I agree that it's architecturally more complex, but the implementation is simple and elegant enough to make dealing with that complexity almost a no-brainer.

kennyadsl:
Before merging, I'd like to understand how we want to communicate this to users. Can a section similar to this [Rails one](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#configure-framework-defaults) be enough?

core/spec/lib/spree/app_configuration_spec.rb
```diff
@@ -36,6 +36,18 @@
     expect(prefs.pricing_options_class).to eq Spree::Variant::PriceSelector.pricing_options_class
   end
 
+  it "allows defining diferent defaults depending on the Solidus version" do
```
kennyadsl, line 39: small typo `diferent` -> `different`

kennyadsl (approved):
Left a non-blocking comment. Thank you Marc, this is awesome!

core/spec/lib/spree/app_configuration_spec.rb
waiting-for-dev: :mag: fixed, thanks!

waiting-for-dev:
I added a new commit generalizing the possibility to use different defaults by version to any `Preferable` class, and not only to `Spree::AppConfiguration`. Please, @kennyadsl , @aldesantis , I'd appreciate if you can take a look at the second commit.

waiting-for-dev:
> I added a new commit generalizing the possibility to use different defaults by version to any `Preferable` class, and not only to `Spree::AppConfiguration`. Please, @kennyadsl , @aldesantis , I'd appreciate if you can take a look at the second commit.

I pushed a new commit where, instead of enabling the new feature for any `Preferable` class, it's only done for descendants of `Spree::Preferences::Configuration`. I.e., configuration classes for core, backend, frontend & engine. After talking with @kennyadsl , we decided to set other `Preferable` classes aside, as they are ActiveRecord models that have very different behavior and can be configured through the backend.

aldesantis (approved):

aldesantis:
@waiting-for-dev makes sense to me!

jarednorman (approved):
A lot of moving parts here, but I've gone over it a couple times now and I think it all makes sense.

kennyadsl:
I think we are ready here, can we just rebase to remove commit 2 and 3 (3 reverts 2 and we can probably have the history cleaner by doing so).

waiting-for-dev:
> I think we are ready here, can we just rebase to remove commit 2 and 3 (3 reverts 2 and we can probably have the history cleaner by doing so).

I squashed all the commits and updated the commit message. I also rebased from master.

I made an extra addition, adding the `load_defaults` call in the backend, frontend & API configuration blocks in the `spree.rb` initializer template.

jarednorman (approved):
The approach for handling changes is very clever.

kennyadsl:
Merging, but I think we need to call `config.load_defaults` in the dummy apps as well.

waiting-for-dev:
> Merging, but I think we need to call `config.load_defaults` in the dummy apps as well.

Thanks, @kennyadsl. It's not needed because of `load_defaults` defaulting to the current Solidus version, but I'm introducing it on #4087 to avoid the warning we introduce there.

PR #4065, opened 2021-05-17
Implement `#with_event_subscribers` test helper

aldesantis:
**Description**

One of the big problem with event subscribers is that they're like ActiveRecord callbacks: they are executed every time, even when you don't want to. This makes testing subscribers in isolation extremely hard.

On the other hand, you also want to make sure your subscribers work in integration, e.g. when running system/request specs.

This helper (which we've been using in production for a while now) addresses the problem by providing a concise syntax for deciding what subscribers you want to execute in a given block of code.

**Checklist:**


aldesantis:
Test failure is legit, unfortunately. `_with_subscriber_registry_override` causes subscribers to be activated multiple times.

jarednorman (commented):
I just took a quick pass over this and I really like the idea.

core/lib/spree/testing_support/event_helpers.rb
```diff
@@ -0,0 +1,71 @@
...
+            registry.activate_all_subscribers
+          end
+
+          yield
+        end
+      end
+
+      private
+
+      def _deactivate_all_subscribers
```
jarednorman, line 37: Up to you, but if the concern is that you want to avoid polluting the test context with private methods, what about extracting the helper's whole functionality to an object or class invoked by the helper itself? That's usually my approach for mixins that don't require any access to the context of where they're being mixed in.

You and I know why these methods have prepended underscores, but it seems like something non-obvious to the average developer who might look at this. Giving them underscores also doesn't make them any less callable or visible in a test context either.

It's a tradeoff, since you're adding some indirection, but the whole implementation could still live in-line in this file, so I don't think it would add any confusion.

Not blocking though.

jarednorman (approved):
I like this. I've made one suggestion about how I think it could be improved, but I don't feel that what's there is bad, so I don't consider it blocking.

waiting-for-dev:
Closing, as it's already covered on the new event bus system:

1.- Stub helpers from [Spree::TestingSupport::BusHelpers](https://github.com/solidusio/solidus/blob/master/core/lib/spree/testing_support/bus_helpers.rb).
2.- [Testing methods on Omnes](https://github.com/nebulab/omnes#testing).

PR #4070, opened 2021-05-19
Accept static_model_preferences class name as string

### LINKED ISSUE #4040: `static_model_preferences` doesn't accept class-names as strings, interfering with autoloading
**Solidus Version**:

up to 3.0

**To Reproduce**

1. Create a new store and add `config.static_model_preferences.add('Spree::PaymentMethod::StripeCreditCard', …)` as a static preference, using a string as the class-name
2. Try to start the server

**Current behavior**

The following error is raised:

```
/Users/elia/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/solidus_core-2.11.7/lib/spree/preferences/static_model_preferences.rb:12:in `block in initialize': undefined method `defined_preferences' for "Spree::PaymentMethod::StripeCreditCard":String (NoMethodError)
	from /Users/elia/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/solidus_core-2.11.7/lib/spree/preferences/static_model_preferences.rb:11:in `each'
	from /Users/elia/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/solidus_core-2.11.7/lib/spree/preferences/static_model_preferences.rb:11:in `initialize'
	from /Users/elia/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/solidus_core-2.11.7/lib/spree/preferences/static_model_preferences.rb:41:in `new'
	from /Users/elia/.asdf/installs/ruby/2.7.2/lib/ruby/gems/2.7.0/gems/solidus_core-2.11.7/lib/spree/preferences/static_model_preferences.rb:41:in `add'
	from …/my-store/config/initializers/solidus.rb:60:in `block in <main>'
```

**Expected behavior**

Solidus should use `constantize` appropriately to avoid autoloading constants during the initialization.

willianveiga:
**Description**
Make it possible to accept `static_model_preferences` class name as `string`.

Fixes #4040

**Checklist:**

### DIFF
--- a/core/lib/spree/preferences/static_model_preferences.rb
+++ b/core/lib/spree/preferences/static_model_preferences.rb
@@ -36,9 +36,14 @@ module Spree
       end
 
       def add(klass, name, preferences)
+        constantized_klass = klass.try(:constantize) || klass
+
         # We use class name instead of class to allow reloading in dev
-        raise "Static model preference '#{name}' on #{klass} is already defined" if @store[klass.to_s][name]
-        @store[klass.to_s][name] = Definition.new(klass, preferences)
+        if @store[constantized_klass.to_s][name]
+          raise "Static model preference '#{name}' on #{constantized_klass} is already defined"
+        end
+
+        @store[constantized_klass.to_s][name] = Definition.new(constantized_klass, preferences)
       end
 
       def for_class(klass)
--- a/core/spec/models/spree/preferences/static_model_preferences_spec.rb
+++ b/core/spec/models/spree/preferences/static_model_preferences_spec.rb
@@ -5,7 +5,7 @@ require 'rails_helper'
 module Spree
   RSpec.describe Preferences::StaticModelPreferences do
     let(:preference_class) do
-      Class.new do
+      MyFakeStaticModelPreferencesClass ||= Class.new do
         include Preferences::Preferable
         preference :color, :string
       end
@@ -20,6 +20,11 @@ module Spree
     it "can store preferences" do
       subject.add(preference_class, 'my_definition', {})
       # just testing that it was added here
+      expect(definitions).to have_key('my_definition')
+    end
+
+    it "can add with string class name" do
+      subject.add(preference_class.to_s, 'my_definition', {})
       expect(definitions).to have_key('my_definition')
     end
 


kennyadsl:
Thanks @willianveiga, this looks good but I'd like to better understand why we need this ([as asked in the original issue](https://github.com/solidusio/solidus/issues/4040#issuecomment-845278751)).

I mean, in which circumstances we need to use a string there?

core/lib/spree/preferences/static_model_preferences.rb
```diff
@@ -36,9 +36,14 @@ def initialize
       end
 
       def add(klass, name, preferences)
+        constantized_klass = klass.try(:constantize) || klass
```
elia, line 39: This will trigger an autoload during initialization, which is one of the things that this change should avoid. Thoughts on just converting to a string if it's not already and allowing `Definition` to accept it as well?

core/spec/models/spree/preferences/static_model_preferences_spec.rb
```diff
@@ -5,7 +5,7 @@
 module Spree
   RSpec.describe Preferences::StaticModelPreferences do
     let(:preference_class) do
-      Class.new do
+      MyFakeStaticModelPreferencesClass ||= Class.new do
```
elia, line 8: I think defining a constant here will actually define it under `Spree::`, an alternative could be using stubbed constants https://rspec.info/blog/2012/06/constant-stubbing-in-rspec-2-11/, but I'm not sure it's a good one, you need to check how they'll interact with this setup.

elia (changes requested):
Hey @willianveiga! thanks for tackling this 💯 
I left a couple of comments for possible issues with the proposed implementation 👍

core/lib/spree/preferences/static_model_preferences.rb
waiting-for-dev: Hey @elia, I'm wondering how to fix it. In the end, `add` creates another `Definition` instance, and [it calls `.defined_preferences` on the class](https://github.com/solidusio/solidus/blob/c914bd1abf47029d9295e656c9d4017c34c2d12b/core/lib/spree/preferences/static_model_preferences.rb#L12) :shrug:

PR #4072, opened 2021-05-20
Support Ruby 3

waiting-for-dev:
Besides a couple of [issues related to the new separation between
keyword and positional
arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/),
this bug https://github.com/rails/rails/issues/42098 is currently
hitting us. As we don't need to do anything from our side, we mark our
failing example so that we'll notice when it gets fixed upstream.

**Checklist:**


kennyadsl (approved):
Thanks @waiting-for-dev !

jarednorman (approved):

PR #4073, opened 2021-05-21
Move currently_valid_prices to a method

waiting-for-dev:
Having `currently_valid_prices` as an association between `Variant` and
`Price` made it a source of inconsistencies with the undecorated
`prices` association:

```
Spree::Variant.new.tap do |v|
  v.currently_valid_prices << Spree::Price.new
end.prices.any? # => false
```

Note: Extraction from #3994

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -13,6 +13,13 @@ module Spree
         autosave: true
     end
 
+    # Returns `#prices` prioritized for being considered as default price
+    #
+    # @return [ActiveRecord::Relation<Spree::Price>]
+    def currently_valid_prices
+      prices.currently_valid
+    end
+
     def find_or_build_default_price
       default_price || build_default_price(Spree::Config.default_pricing_options.desired_attributes)
     end
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -24,7 +24,6 @@ module Spree
       stock_items.discard_all
       images.destroy_all
       prices.discard_all
-      currently_valid_prices.discard_all
     end
 
     attr_writer :rebuild_vat_prices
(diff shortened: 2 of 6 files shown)


core/app/models/spree/variant.rb
```diff
@@ -24,7 +24,6 @@ class Variant < Spree::Base
       stock_items.discard_all
       images.destroy_all
       prices.discard_all
-      currently_valid_prices.discard_all
```
kennyadsl, line 27: This is not needed because the line above would still remove those, right?

core/spec/models/spree/price_spec.rb
```diff
@@ -96,6 +96,33 @@
         expect(price.country).to eq(country)
       end
     end
+
+    describe '.currently_valid' do
```
kennyadsl, line 100: If I got it correctly, these are passing even without the change in the PR, isn't it?

core/app/models/concerns/spree/default_price.rb
```diff
@@ -13,6 +13,13 @@ module DefaultPrice
         autosave: true
     end
 
+    # Returns `#prices` prioritized for being considered as default price
+    #
+    # @return [Array<Spree::Price>]
```
kennyadsl, line 18: This should be a `ActiveRecord::Relation`.

core/app/models/spree/variant.rb
waiting-for-dev: Right!

core/spec/models/spree/price_spec.rb
waiting-for-dev: Right!!

core/app/models/concerns/spree/default_price.rb
waiting-for-dev: Right x 3, & fixed!! :slightly_smiling_face:

gmacdougall (approved):

tvdeyen (approved):
Nice. Makes sense

PR #4076, opened 2021-05-24
Simplify `Variant#default_price` logic

waiting-for-dev:
Before this work, `#default_price` was a `has_one` association between
`Variant` and `Product`. As we already have a `has_many` association
between both models, this redundancy was a source of inconsistencies.
E.g.:

```ruby
include Spree
Variant.new(price: Price.new) # price delegates to default_price
Variant.prices # => []
```

From now on, `#default_price` is a regular method that searches within
`#prices` taking into account the criteria for a price to be considered
the default one.

We have renamed previous method `#find_default_price_or_build` to
`default_price_or_build`. The latter feels less standard according to
Rails conventions, but the intention here is, precisely, to communicate
that this is not a Rails association.

No longer being a Rails association makes it impossible to use the default
ransack conventions to build the sort-by-price link on the products
listing page. For this reason, we added
`sort_by_master_default_price_amount_{asc, desc}` scopes to the `Price`
model, which is automatically picked up by ransack. As it's now
explicit, these queries ignore the `ORDER BY` clauses implicit in the
`currently_valid_prices` method, but this was also happening in the
query built by ransack from the `has_one` association:

```SQL
SELECT "spree_products".* FROM "spree_products" LEFT OUTER JOIN
"spree_variants" ON "spree_variants"."is_master" = ? AND "spree
_variants"."product_id" = "spree_products"."id" LEFT OUTER JOIN
"spree_prices" ON "spree_prices"."currency" = ? AND "spree_pric
es"."country_iso" IS NULL AND "spree_prices"."variant_id" =
"spree_variants"."id" WHERE "spree_products"."deleted_at" IS NULL O
RDER BY "spree_prices"."amount" ASC, "spree_products"."id" ASC LIMIT ?
OFFSET ?  [["is_master", 1], ["currency", "USD"], ["LIMI  T", 10],
["OFFSET", 0]]
```

However, the new query doesn't include discarded prices on the result,
but I think it's something desirable:

```SQL
SELECT "spree_products".* FROM "spree_products" LEFT OUTER JOIN
"spree_variants" "master" ON "master"."is_master" = ? AND "mast
er"."product_id" = "spree_products"."id" LEFT OUTER JOIN "spree_prices"
"prices" ON "prices"."deleted_at" IS NULL AND "prices".  "variant_id" =
"master"."id" LEFT OUTER JOIN "spree_variants" "masters_spree_products"
ON "masters_spree_products"."is_master"   = ? AND
"masters_spree_products"."product_id" = "spree_products"."id" WHERE
"spree_products"."deleted_at" IS NULL AND "prices".  "currency" = ? AND
"prices"."country_iso" IS NULL AND "prices"."deleted_at" IS NULL ORDER
BY prices.amount DESC, "spree_product  s"."id" ASC LIMIT ? OFFSET ?
[["is_master", 1], ["is_master", 1], ["currency", "USD"], ["LIMIT", 10],
["OFFSET", 0]]
```

**Checklist:**
- [] I have updated Guides and README accordingly to this change (if needed)

Note: Extracted from #3994

### DIFF
--- a/backend/app/views/spree/admin/products/index.html.erb
+++ b/backend/app/views/spree/admin/products/index.html.erb
@@ -78,7 +78,7 @@
               <%= render 'spree/admin/shared/image', image: product.gallery.images.first, size: :mini %>
             </td>
             <td><%= link_to product.try(:name), edit_admin_product_path(product) %></td>
-            <td class="align-right"><%= product.display_price.to_html %></td>
+            <td class="align-right"><%= product.display_price&.to_html %></td>
             <td class="actions" data-hook="admin_products_index_row_actions">
               <%= link_to_edit product, no_text: true, class: 'edit' if can?(:edit, product) && !product.deleted? %>
               &nbsp;
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -5,12 +5,13 @@ module Spree
     extend ActiveSupport::Concern
 
     included do
-      has_one :default_price,
-        -> { with_discarded.currently_valid.with_default_attributes },
-        class_name: 'Spree::Price',
-        inverse_of: :variant,
-        dependent: :destroy,
-        autosave: true
+      delegate :display_price, :display_amount, :price, to: :default_price, allow_nil: true
+      delegate :price=, to: :default_price_or_build
+
+      # @see Spree::Variant::PricingOptions.default_price_attributes
+      def self.default_price_attributes
+        Spree::Config.default_pricing_options.desired_attributes
+      end
     end
 
     # Returns `#prices` prioritized for being considered as default price
@@ -20,15 +21,60 @@ module Spree
       prices.currently_valid
     end
 
-    def find_or_build_default_price
-      default_price || build_default_price(Spree::Config.default_pricing_options.desired_attributes)
+    # Returns {#default_price} or builds it from {Spree::Variant.default_price_attributes}
+    #
+    # @return [Spree::Price, nil]
+    # @see Spree::Variant.default_price_attributes
+    def default_price_or_build
+      default_price ||
+        prices.build(self.class.default_price_attributes)
     end
 
-    delegate :display_price, :display_amount, :price, to: :find_or_build_default_price
-    delegate :price=, to: :find_or_build_default_price
+    # Select from {#prices} the one to be considered as the default
+    #
+    # This method works with the in-memory association, so non-persisted prices
+    # are taken into account. Discarded prices are also considered.
+    #
+    # A price is a candidate to be considered as the default when it meets
+    # {Spree::Variant.default_price_attributes} criteria. When more than one candidate is
+    # found, non-persisted records take preference. When more than one persisted
+    # candidate exists, the one most recently updated is taken or, in case of
+    # race condition, the one with higher id.
+    #
+    # @return [Spree::Price, nil]
+    # @see Spree::Variant.default_price_attributes
+    def default_price
+      prioritized_default(
+        prices_meeting_criteria_to_be_default(
+          (prices + prices.with_discarded).uniq
+        )
+      )
+    end
 
     def has_default_price?
       default_price.present? && !default_price.discarded?
+    end
+
+    private
+
+    def prices_meeting_criteria_to_be_default(prices)
+      criteria = self.class.default_price_attributes.transform_keys(&:to_s)
+      prices.select do |price|
+        contender = price.attributes.slice(*criteria.keys)
+        criteria == contender
+      end
+    end
+
+    def prioritized_default(prices)
+      prices.min do |prev, succ|
+        contender_one, contender_two = [succ, prev].map do |item|
+          [
+            item.updated_at || Time.zone.now,
+            item.id || Float::INFINITY
+          ]
+        end
+        contender_one <=> contender_two
+      end
     end
   end
 end
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -60,6 +60,17 @@ module Spree
     has_many :line_items, through: :variants_including_master
     has_many :orders, through: :line_items
 
+    scope :sort_by_master_default_price_amount_asc, -> {
+      with_default_price.order('spree_prices.amount ASC')
+    }
+    scope :sort_by_master_default_price_amount_desc, -> {
+      with_default_price.order('spree_prices.amount DESC')
+    }
+    scope :with_default_price, -> {
+      left_joins(master: :prices)
+        .where(master: { spree_prices: Spree::Config.default_pricing_options.desired_attributes })
+    }
+
     def find_or_build_master
       master || build_master
     end
--- a/core/spec/models/spree/product_spec.rb
+++ b/core/spec/models/spree/product_spec.rb
@@ -151,8 +151,7 @@ RSpec.describe Spree::Product, type: :model do
 
       context "with currency set to JPY" do
         before do
-          product.master.default_price.currency = 'JPY'
-          product.master.default_price.save!
+          product.master.default_price.update!(currency: 'JPY')
           stub_spree_preferences(currency: 'JPY')
         end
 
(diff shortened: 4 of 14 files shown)


core/spec/models/spree/product_spec.rb
```diff
@@ -152,8 +152,8 @@ class Extension < Spree::Base
       context "with currency set to JPY" do
         before do
           product.master.default_price.currency = 'JPY'
-          product.master.default_price.save!
           stub_spree_preferences(currency: 'JPY')
+          product.master.default_price.save!
```
kennyadsl, line 156: Why do we need this change?

backend/app/views/spree/admin/products/index.html.erb
```diff
@@ -78,7 +78,7 @@
               <%= render 'spree/admin/shared/image', image: product.gallery.images.first, size: :mini %>
             </td>
             <td><%= link_to product.try(:name), edit_admin_product_path(product) %></td>
-            <td class="align-right"><%= product.display_price.to_html %></td>
+            <td class="align-right"><%= product.display_price&.to_html %></td>
```
kennyadsl, line 81: Why this is needed?

core/app/models/concerns/spree/default_price.rb
```diff
@@ -5,20 +5,54 @@ module DefaultPrice
...
+    #
+    # A price is a candidate to be considered as the default when it meets
+    # {Spree::Variant.default_pricing} criteria. When more than one candidate is
+    # found, non-persisted records take preference. When more than one persisted
+    # candidate exists, the one most recently updated is taken or, in case of
+    # race condition, the one with higher id.
+    #
+    # @return [Spree::Price, nil]
+    # @see Spree::Price.default_pricing
+    def default_price
```
kennyadsl, line 39: This method is a bit complex and hard to read to understand what it does. Any thoughts about moving it to a separate class that is responsible to select the default price only?

kennyadsl (commented):
Left some questions, thanks Marc!

core/spec/models/spree/product_spec.rb
waiting-for-dev: `default_price` was an association with `autosave` option set to `true`, same that `master` on `product`. Now we need to save it manually.

waiting-for-dev: Thinking about it twice, maybe that could have backward compatibility issues. However, at least, not providing the price will result [in a validation error](https://github.com/solidusio/solidus/blob/58f2e0bbcc6c27cff459985bce67e8912f324220/core/app/models/spree/product.rb#L111) when the app is configured to require a default price. I don't think there's a reliable way to detect the situation to emit a warning, though.

backend/app/views/spree/admin/products/index.html.erb
waiting-for-dev: In fact, it's not needed. Probably I used it as an intermediate step, and it leaked to the final result. Thanks for catching it! I forced a push and removed the change.

core/app/models/concerns/spree/default_price.rb
waiting-for-dev: Yeah, I completely agree. I have some reserves because of:

- It would be nice to have it be injectable for the end-user. However, we haven't decided how to do that yet. We could build on top of the old system, though.
- The logic around prices is very complex, and I'd like to simplify it further. However, probably it's not a preference at this point.

WDYT? We can do it now if you think it's better to leave it clean at this point.

core/spec/models/spree/product_spec.rb
kennyadsl: But the change is only moving the save a line above, isn't it?

core/app/models/concerns/spree/default_price.rb
```diff
@@ -5,20 +5,54 @@ module DefaultPrice
...
-        -> { with_discarded.currently_valid.with_default_attributes },
-        class_name: 'Spree::Price',
-        inverse_of: :variant,
-        dependent: :destroy,
-        autosave: true
+      delegate :display_price, :display_amount, :price, to: :default_price, allow_nil: true
+      delegate :price=, to: :default_price_or_build
+
+      # @see Spree::Variant::PricingOptions.default_price_attributes
+      def self.default_pricing
```
mamhoff, line 12: naming nit: Should this be called `self.default_price_attributes`?

core/app/models/spree/product.rb
```diff
@@ -60,6 +60,17 @@ class Product < Spree::Base
...
 
+    scope :sort_by_master_default_price_amount_asc, -> {
+      with_default_price.order('spree_prices.amount ASC')
+    }
+    scope :sort_by_master_default_price_amount_desc, -> {
+      with_default_price.order('spree_prices.amount DESC')
+    }
+    scope :with_default_price, -> {
+      left_joins(master: :prices)
+        .where(master: { spree_prices: Spree::Config.default_pricing_options.desired_attributes })
```
mamhoff, line 71: This didn't work with some database adapters in the past. Can you confirm it does with pg, mysql and sqlite? Because if this works, the cursed `Spree::Variant.with_prices` scope can go.

mamhoff (approved):
I like this.

core/spec/models/spree/product_spec.rb
waiting-for-dev: I'm sorry, I don't understand what you mean. We're adding a new line to the test so that price gets persisted into the database.

core/app/models/concerns/spree/default_price.rb
kennyadsl: As we discussed IRL yesterday. Maybe a good tradeoff could be trying to use more semantic variable names and/or extracting something to a method in this same class to help reading the code.

core/spec/models/spree/product_spec.rb
kennyadsl: This is what I see in the diff:

CleanShot 2021-06-01 at 09 36 22@2x

core/app/models/spree/product.rb
waiting-for-dev: It seems it works. That's the SQL generated for each adapter:

sqlite:
```ruby
=> "SELECT \"spree_products\".* FROM \"spree_products\" LEFT OUTER JOIN \"spree_variants\" \"master\" ON \"master\".\"is_master\" = 1 AND \"master\".\"product_id\" = \"spree_products\".\"id\" LEFT OUTER JOIN \"spree_prices\" ON \"spree_prices\".\"deleted_at\" IS NULL AND \"spree_prices\".\"variant_id\" = \"master\".\"id\" WHERE \"spree_products\".\"deleted_at\" IS NULL AND \"spree_prices\".\"currency\" = 'USD' AND \"spree_prices\".\"country_iso\" IS NULL"
```

Postgres:
```ruby
"SELECT \"spree_products\".* FROM \"spree_products\" LEFT OUTER JOIN \"spree_variants\" \"master\" ON \"master\".\"is_master\" = TRUE AND \"master\".\"product_id\" = \"spree_products\".\"id\" LEFT OUTER JOIN \"spree_prices\" ON \"spree_prices\".\"deleted_at\" IS NULL AND \"spree_prices\".\"variant_id\" = \"master\".\"id\" WHERE \"spree_products\".\"deleted_at\" IS NULL AND \"spree_prices\".\"currency\" = 'USD' AND \"spree_prices\".\"country_iso\" IS NULL"
```

MySQL:
```ruby
"SELECT `spree_products`.* FROM `spree_products` LEFT OUTER JOIN `spree_variants` `master` ON `master`.`is_master` = TRUE AND `master`.`product_id` = `spree_products`.`id` LEFT OUTER JOIN `spree_prices` ON `spree_prices`.`deleted_at` IS NULL AND `spree_prices`.`variant_id` = `master`.`id` WHERE `spree_products`.`deleted_at` IS NULL AND `spree_prices`.`currency` = 'USD' AND `spree_prices`.`country_iso` IS NULL"
```

While the tests using that method work in the three adapters:

- https://github.com/nebulab/solidus/blob/cb87e1666ef7d9dc07d0e0df0b96ffd05e8bab0a/core/spec/models/spree/product_spec.rb#L623
- https://github.com/nebulab/solidus/blob/cb87e1666ef7d9dc07d0e0df0b96ffd05e8bab0a/core/spec/models/spree/product_spec.rb#L634

We could tackle that scope in a separate PR.

core/app/models/concerns/spree/default_price.rb
waiting-for-dev: I tried to make everything clear. I think that the result is much more readable. Thanks for pointing it out. Please, take a look when you have a second.

core/app/models/concerns/spree/default_price.rb
waiting-for-dev: Done!

core/app/models/concerns/spree/default_price.rb
kennyadsl: I think this is perfect, thanks Marc!

core/spec/models/spree/product_spec.rb
waiting-for-dev: Following @kennyadsl suggestion, we updated it a bit to look less confusing:

```ruby
product.master.default_price.update!(currency: 'JPY')
stub_spree_preferences(currency: 'JPY')
```

backend/app/views/spree/admin/products/index.html.erb
kennyadsl: I can still see this change, maybe some mess with git?

kennyadsl (approved):
Thanks @waiting-for-dev!

jarednorman (approved):

PR #4078, opened 2021-05-24
Deprecate callback setting variant price from master

waiting-for-dev:
This simplifies logic, and it can be seen as preparatory work to propose
to the user the inheritance of all, and not just the default, prices
from master when creating a new variant.

Extracted from #3994 

**Checklist:**

### DIFF
--- a/sample/db/samples/variants.rb
+++ b/sample/db/samples/variants.rb
@@ -30,184 +30,218 @@ variants = [
     product: solidus_tshirt,
     option_values: [small, blue],
     sku: "SOL-00003",
-    cost_price: 17
+    cost_price: 17,
+    price: 25
   },
   {
     product: solidus_tshirt,
     option_values: [small, black],
     sku: "SOL-00002",
-    cost_price: 17
+    cost_price: 17,
+    price: 25
   },
   {
     product: solidus_tshirt,
     option_values: [small, white],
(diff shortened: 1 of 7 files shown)


kennyadsl (approved):
Thanks Marc!

kennyadsl:
Maybe we can add a spec that tests the deprecation warning emitted when the price is not passed?

sample/db/samples/variants.rb
```diff
@@ -30,184 +30,218 @@
     product: solidus_tshirt,
     option_values: [small, blue],
     sku: "SOL-00003",
-    cost_price: 17
+    cost_price: 17,
+    price: 10
```
spaghetticode, line 34: it seems a bit strange that `price` is lower than the `cost_price`, shouldn't it be greater than that?

gmacdougall (approved):

mamhoff (approved):

jarednorman (approved):

sample/db/samples/variants.rb
waiting-for-dev: Fair enough, fixed :sweat_smile:

waiting-for-dev:
> Maybe we can add a spec that tests the deprecation warning emitted when the price is not passed?

Makes sense. I added it and rebased from master.

Issue #3977, opened 2021-03-05
Order controller populate method should be able to handle multiple variants


seand7565:
I think there are a lot of circumstances where it's desirable for the order controller to be able to accept multiple variants. For instance, if I sell selectable "add-ons" or extras on a product page, that are separate products, I'd want to be able to add both the initial product and the add-on product to the cart at the same time, if it's selected.

I've made this change in other stores, and it's a pretty small one - order_controller just needs to be able to accept params[:variants] instead of params[:variant] and params[:quantity], then do the same action it's doing now, just looping over params[:variants]. Lots of people have custom frontends, so we'll need to deprecate this carefully, but it's a small change and introduces a lot of flexibility.

RyanofWoods:
Hi @seand7565 ,

Thanks for the raising this feature idea, I like it! 🙂

In order to allow the controller to have :variants, the _cart_form partial will need to instead send over the variants as an array. As you said, a lot of people have custom front-ends, and so if they have _cart_form partial that overrides the gem, it will break in their frontend OrdersController.

Because of this I thought of three solutions:
1. Simply change the form to send the variants as an array and be very clear about this update, so that stores with a custom _cart_form know to update it when upgrading.
2. Like 1. but OrdersController#populate still allows for use of the old params :variant_id and :quantity. This transformation would reduce the risk of breaking stores and could also be also be pulled after a grace period if desired.
3. Keep :variant_id and :quantity as is, but instead allow the controller to also deal with a new possible parameter :add_ons. This allows for the old functionality and no breaking stores, but allows for devs to add add_ons if desired. (Drawback: if someone with a custom front-end for some reason had a page where you can select multiple variants of the main item at the same time, and there isn't a concept of add-ons on that page, they will be forced to use the variant_id and add_ons names which could get confusing)

If you have any thoughts on which would be better or any other ideas/notes, I would greatly appreciated it. :)

PR #4081, opened 2021-05-25
Change frontend to allow multiple variants

RyanofWoods:
Hello there! 😄

Before, the front-end _cart_form partial and OrdersController#populate
only allowed for one variant to be sent. This changes that behaviour
to allow for multiple variants, which increases flexibility for
custom front-ends where add-ons can be included on the product page.

The old parameters :variant_id and :quantity are still allowed,
however gives a deprecation warning. The new parameter is:
`variants: [{ :variant_id, :quantity }]`
The _cart_form partial and other specs that reference #populate
have been updated to use this.

**Description**
PR for issue #3977. 

1. Currently, this PR allows the use of old parameters(`:variant_id` and `:quantity`), however would it be better to remove them, and to only have the new parameter :variants?
2. If do wish to keep it, is there something more appropriate than raising a `Spree:Deprecation` warning? (Pretty sure this is not its intended use)
3. Couldn't see any documentation that needs updating. However, could it be a good idea to add more documentation about adding custom-front-end pages and also include about how you can add add-ons based on this PR? (maybe even a new documentation page)

Also note, there are similar tests because I kept the old parameters(`:variant_id` and `:quantity`).

Example on how you can have add-ons through checkboxes:

```  
  <% add_ons = Spree::Variant.first(2) %>
  <% add_ons.each do |variant|%>
    <li>
      <%= "Add on?" %>
      <%= check_box_tag "variants[][variant_id]", variant.id %>
      <%= "Quantity:" %>
      <%= number_field_tag "variants[][quantity]", 1, class: 'title', min: 1 %>
    </li>
  <% end %>
```

add on
cart

_Note, you also do not need to have the quantity input on add-ons if you wish it to always be set at 1._

**Checklist:**

First Solidus PR 😊

### DIFF
--- a/frontend/app/controllers/spree/orders_controller.rb
+++ b/frontend/app/controllers/spree/orders_controller.rb
@@ -52,17 +52,43 @@ module Spree
       @order = current_order(create_order_if_necessary: true)
       authorize! :update, @order, cookies.signed[:guest_token]
 
-      variant  = Spree::Variant.find(params[:variant_id])
-      quantity = params[:quantity].present? ? params[:quantity].to_i : 1
+      variants = []
+
+      # allow old parameters but raise deprecation warning
+      if params[:variant_id].present?
+        variants << { variant_id: params[:variant_id], quantity: params[:quantity] }
+
+        Spree::Deprecation.warn(
+          "\nYou have a custom front-end _cart_form partial which
+          uses deprecated parameters: :variant_id, :quantity.
+          These will be removed in future versions of Solidus.
+          Please update your parameters according to:
+          https://github.com/solidusio/solidus/pull/4081\n"
+        )
+      end
+
+      # check needed for those who have custom front-end that uses old parameters
+      variants.push(*params[:variants]) if params[:variants].present?
 
       # 2,147,483,647 is crazy. See issue https://github.com/spree/spree/issues/2695.
-      if !quantity.between?(1, 2_147_483_647)
+      variants_has_bad_quantity = variants.any? do |variant|
+        variant[:quantity].present? && !variant[:quantity].to_i.between?(1, 2_147_483_647)
+      end
+
+      if variants_has_bad_quantity
         @order.errors.add(:base, t('spree.please_enter_reasonable_quantity'))
       else
-        begin
-          @line_item = @order.contents.add(variant, quantity)
-        rescue ActiveRecord::RecordInvalid => error
-          @order.errors.add(:base, error.record.errors.full_messages.join(", "))
+        variants.each do |variant|
+          next unless variant.key?(:variant_id)
+
+          v = Spree::Variant.find(variant[:variant_id])
+          q = variant[:quantity].present? ? variant[:quantity].to_i : 1
+
+          begin
+            @line_item = @order.contents.add(v, q)
+          rescue ActiveRecord::RecordInvalid => error
+            @order.errors.add(:base, error.record.errors.full_messages.join(", "))
+          end
         end
       end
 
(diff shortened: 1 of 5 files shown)


kennyadsl (commented):
Thanks @RyanofWoods, this makes a lot of sense even if I'm tempted attempting to create a new controller/action for this. 

Let me explain better: `orders/populate` is a non-RESTful route and it's here since the first versions of Spree/Solidus probably. I think that deprecating some of its functionality could be a good opportunity to deprecate the whole method and use something more isolated and easy to customize. We could use a more RESTful structure, [similar to what we use in the API](https://solidus.docs.stoplight.io/api-reference/line-items/create-order-line-item), like `POST orders/line_items`. 

As a bonus point, if we deprecate the whole method using [`deprecate`](https://github.com/rails/rails/blob/56cfefe0cd6bb6debd2fba4a5a1d839993bf2eac/activesupport/lib/active_support/core_ext/module/deprecation.rb#L5), with something like:

```ruby
def populate
  # ..
end
deprecate populate: 'please use something else...', deprecator: Spree::Deprecation
```

we'll be able to also print the deprecation warning for people that opened the controller and changed the populate code completely, without calling `super`.

If this makes sense for you as well, I think that we can do the change in two separate PRs, the first will just move the action + deprecate the old one, the second will add the multiple variants feature. Let me know your thoughts!

RyanofWoods:
Thanks for the review @kennyadsl. That sounds like a really good idea and makes a lot of sense. It would make more RESTful and single responsible for sure.

Is LineItemsController and create fine? Also, I think the route has to be unnested as the user might not have order yet, and we can do `@order = current_order(create_order_if_necessary: true)` like before. Is this correct?

Lastly, I wanted to ask about OrdersController#populate_redirect, which is based on GET '/orders/populate'. This route doesn't get referenced anywhere, but there is one spec on it and a old issue post #2006. Any thoughts on this?

kennyadsl:
@RyanofWoods 

> Is LineItemsController and create fine? Also, I think the route has to be unnested as the user might not have order yet, and we can do @order = current_order(create_order_if_necessary: true) like before. Is this correct?

POST requests are usually mapped to `create` actions, so yes! `LineItemsController` looks good, otherwise I was think we can create a OrderContentsController, to reflect [what is actually](https://github.com/solidusio/solidus/blob/d6e84ecb19aa1a7eac685db912e0cd048ad95d16/core/app/models/spree/order_contents.rb) called to populate the order.

> Lastly, I wanted to ask about OrdersController#populate_redirect, which is based on GET '/orders/populate'. This route doesn't get referenced anywhere, but there is one spec on it and a old issue post #2006. Any thoughts on this?

I think the issue was https://github.com/spree/spree/issues/2006, from the project we forked from. I'm not sure we want to keep this functionality for the reasons mentioned in it, what do you think?

RyanofWoods:
@kennyadsl 

> @RyanofWoods

> POST requests are usually mapped to `create` actions, so yes! `LineItemsController` looks good, otherwise I was thinking we can create a `OrderContentsController`, to reflect [what is actually](https://github.com/solidusio/solidus/blob/d6e84ecb19aa1a7eac685db912e0cd048ad95d16/core/app/models/spree/order_contents.rb) called to populate the order.

Good point, lets go with `OrderContentsController`

> I think the issue was [spree/spree#2006](https://github.com/spree/spree/issues/2006), from the project we forked from. I'm not sure we want to keep this functionality for the reasons mentioned in it, what do you think?

Lets perhaps remove it then when we deprecate `populate`

I will start the first PR next week 😊

PR #4087, opened 2021-05-28
Introduce Solidus update process

waiting-for-dev:
Following the work done in solidusio#4064, this commits introduces a Rails
generator that creates a new initializer called, by default,
`new_solidus_defaults.rb`.

This initializer works in a very similar way that
[`new_framework_defaults.rb` does in
Rails](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#configure-framework-defaults).
It allows users to preview the defaults that have changed on a new
Solidus version, as they are printed one by one on a commented line.
Users can then keep enabling them while updating their application code.

We're adding the `load_defaults` call with the old version value to the
generated initializer. Users can remove the file altogether when they're
done with the process. At that point, we require them to add
`load_defaults` with the new version to the main initializer file
(`spree.rb`). Even if there's no actual need for that, as
`loaded_defaults` on the configuration class defaults to the current
Solidus version, we want to enforce that so that users are on the safe
side for the next version upgrade. If they don't add it, we emit a
warning.

For now, we're leaving this as a generator, but we could reference it
from a rake task, although probably there's no need.

Be aware that users need to provide the version from which they are
updating. It's an option that offers more flexibility, as users can
update from versions different from the latest one. It also plays well
with our system's flexibility, for instance, to change defaults between
a pre-release and a release. However, we can add some code to default to
the latest minor version, but we should keep that information in our
code, and that's a small burden for our update process.

**Checklist:**

### DIFF
--- a/core/lib/generators/solidus/update/update_generator.rb
+++ b/core/lib/generators/solidus/update/update_generator.rb
@@ -4,19 +4,28 @@ require 'spree/core/preference_changes_between_solidus_versions'
 require 'rails/generators'
 
 module Solidus
+  # @private
   class UpdateGenerator < ::Rails::Generators::Base
+    FROM = Spree.previous_solidus_minor_version
+
     desc 'Generates a new initializer to preview the new defaults for current Solidus version'
 
     source_root File.expand_path('templates', __dir__)
-
-    class_option :from,
-                 type: :string,
-                 banner: 'The version your are updating from. E.g. 2.11.10'
 
     class_option :initializer_basename,
                  type: :string,
                  default: 'new_solidus_defaults',
                  banner: 'The name for the new initializer'
+
+    class_option :previous_version_prompt,
+                 type: :boolean,
+                 default: true,
+                 banner: 'Prompt to warn about only previous version support'
+
+    class_option :from,
+                 type: :string,
+                 default: FROM,
+                 hide: true
 
     class_option :to,
                  type: :string,
@@ -29,6 +38,12 @@ module Solidus
                  hide: true
 
     def create_new_defaults_initializer
+      previous_version_prompt = options[:previous_version_prompt]
+      return if previous_version_prompt && !yes?(<<~MSG, :red)
+        The update process is only supported if you are coming from version #{FROM}. If this is not the case, please, skip it and update your application to use Solidus #{FROM} before retrying.
+        Are you sure you want to continue? (y/N)
+      MSG
+
       from = options[:from]
       to = options[:to]
       @from = from
--- a/core/lib/generators/solidus/update/update_generator.rb
+++ b/core/lib/generators/solidus/update/update_generator.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'spree/core/preference_changes_between_solidus_versions'
+require 'spree/preferences/preference_differentiator'
 require 'rails/generators'
 
 module Solidus
@@ -81,7 +81,7 @@ module Solidus
     end
 
     def changes_template_for(klass, from, to)
-      changes = Spree::Core::PreferenceChangesBetweenSolidusVersions.new(klass).call(from: from, to: to)
+      changes = Spree::Preferences::PreferenceDifferentiator.new(klass).call(from: from, to: to)
       return '# No changes' if changes.empty?
 
       [
--- a/core/lib/generators/solidus/update/update_generator.rb
+++ b/core/lib/generators/solidus/update/update_generator.rb
@@ -1,0 +1,80 @@
+# frozen_string_literal: true
+
+require 'spree/core/preference_changes_between_solidus_versions'
+require 'rails/generators'
+
+module Solidus
+  class UpdateGenerator < ::Rails::Generators::Base
+    desc 'Generates a new initializer to preview the new defaults for current Solidus version'
+
+    source_root File.expand_path('templates', __dir__)
+
+    class_option :from,
+                 type: :string,
+                 banner: 'The version your are updating from. E.g. 2.11.10'
+
+    class_option :initializer_basename,
+                 type: :string,
+                 default: 'new_solidus_defaults',
+                 banner: 'The name for the new initializer'
+
+    class_option :to,
+                 type: :string,
+                 default: Spree.solidus_version,
+                 hide: true
+
+    class_option :initializer_directory,
+                 type: :string,
+                 default: 'config/initializers/',
+                 hide: true
+
+    def create_new_defaults_initializer
+      from = options[:from]
+      to = options[:to]
+      @from = from
+      @core_changes = core_changes_template(from, to)
+      @frontend_changes = frontend_changes_template(from, to)
+      @backend_changes = backend_changes_template(from, to)
+      @api_changes = api_changes_template(from, to)
+
+      template 'config/initializers/new_solidus_defaults.rb.tt',
+               File.join(options[:initializer_directory], "#{options[:initializer_basename]}.rb")
+    end
+
+    private
+
+    def core_changes_template(from, to)
+      changes_template_for(Spree::AppConfiguration, from, to)
+    end
+
+    def frontend_changes_template(from, to)
+      return '' unless defined?(Spree::Frontend::Engine)
+
+      changes_template_for(Spree::FrontendConfiguration, from, to)
+    end
+
+    def backend_changes_template(from, to)
+      return '' unless defined?(Spree::Backend::Engine)
+
+      changes_template_for(Spree::BackendConfiguration, from, to)
+    end
+
+    def api_changes_template(from, to)
+      return '' unless defined?(Spree::Api::Engine)
+
+      changes_template_for(Spree::ApiConfiguration, from, to)
+    end
+
+    def changes_template_for(klass, from, to)
+      changes = Spree::Core::PreferenceChangesBetweenSolidusVersions.new(klass).call(from: from, to: to)
+      return '# No changes' if changes.empty?
+
+      [
+        ["config.load_defaults('#{from}')"] +
+          changes.map do |pref_key, change|
+            "  # config.#{pref_key} = #{change[:to]}"
+          end.flatten
+      ].join("\n")
+    end
+  end
+end
--- a/core/lib/generators/solidus/update/update_generator.rb
+++ b/core/lib/generators/solidus/update/update_generator.rb
@@ -25,7 +25,7 @@ module Solidus
     class_option :from,
                  type: :string,
                  default: FROM,
-                 hide: true
+                 banner: 'Solidus version from which you are upgrading'
 
     class_option :to,
                  type: :string,
@@ -41,6 +41,7 @@ module Solidus
       previous_version_prompt = options[:previous_version_prompt]
       return if previous_version_prompt && !yes?(<<~MSG, :red)
         The update process is only supported if you are coming from version #{FROM}. If this is not the case, please, skip it and update your application to use Solidus #{FROM} before retrying.
+        If you are confident you want to upgrade from a previous version, you must rerun the generator with the "--from={OLD_VERSION}" argument.
         Are you sure you want to continue? (y/N)
       MSG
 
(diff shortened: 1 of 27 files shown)


core/lib/generators/solidus/update/update_generator.rb
```diff
@@ -0,0 +1,95 @@
...
+                 hide: true
+
+    class_option :initializer_directory,
+                 type: :string,
+                 default: 'config/initializers/',
+                 hide: true
+
+    def create_new_defaults_initializer
+      previous_version_prompt = options[:previous_version_prompt]
+      return if previous_version_prompt && !yes?(<<~MSG, :red)
```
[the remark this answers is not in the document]
waiting-for-dev: It's a false positive as the line below belongs to the first argument of the function.

waiting-for-dev:
@kennyadsl in the last commit, I added the warning prompted to the user that we discussed offline yesterday. We still need to keep track of the previous solidus minor version, so once it's merged, we need to update https://github.com/solidusio/solidus/wiki/How-to-release-Solidus

core/lib/spree/preferences/configuration.rb
```diff
@@ -38,18 +38,35 @@ class Configuration
...
 
+    def check_load_defaults_called(instance_constant_name = nil)
+      return if load_defaults_called || !Spree::Core.solidus_installed?
+
+      target_name = instance_constant_name || "#{self.class.name}.new"
+      Spree::Deprecation.warn <<~MSG
+        Please, indicate wich Solidus version defaults the application must use.
+        The method `#{target_name}.load_defaults` must be called from the Solidus initializer. E.g.:
+
+          config.load_defaults('#{Spree.solidus_version}')
```
kennyadsl, line 65: I think it could be more intuitive if we use `Spree::Config.load_defaults(...)` dynamically based on the module user needs to change (`Api`, `Backend`, etc). Do you think it is possible?

core/lib/spree/preferences/configuration.rb
```diff
@@ -38,18 +38,35 @@ class Configuration
...
     # overriden.
     # @see #load_defaults
     def load_defaults(version)
       @loaded_defaults = version
+      @load_defaults_called = true
       reset
     end
 
+    def check_load_defaults_called(instance_constant_name = nil)
+      return if load_defaults_called || !Spree::Core.solidus_installed?
```
kennyadsl, line 58: Not sure to understand why we need to check `!Spree::Core.solidus_installed?` here, can you explain please?

kennyadsl (commented):
Love it! Left some questions, though. 

One extra step may be having a generic load_default function for users that has all the solidus components in place. Maybe we can automatically skip defaults for subcomponents (constants, like `Spree::Backend`) that are just not defined. Of course, if you think it's a good idea we can iterate on another PR. Thoughts?

Thanks!

core/lib/spree/preferences/configuration.rb
waiting-for-dev: This is to avoid the warning being raised when the user runs the install generator `bin/rails g solidus:install` :slightly_smiling_face: It could lead to thinking that there's some problem when in fact, everything would be ok.

kennyadsl: Oh it was happening with the previous deprecation system we had for Config as well and it was annoying. It's very hard to understand why this is needed though. I suggest three options:

1. Move it (or wrap it) into a more semantic method like (`is_currently_installing_solidus?`)
2. Move this change to a separate commit well documented
3. Both ¯\_(ツ)_/¯

core/lib/spree/preferences/configuration.rb
waiting-for-dev: Please, take a look at commit 5f81fee. I've introduced a `Spree.load_defaults` method, which forwards to all available components. I prefer to have it on `Spree` instead of `Spree::Config`, as `Spree::Config` is an instance of the `Core` configuration, and we need fine-grained control to allow users to enable new defaults one by one, component by component.

The new method simplifies our test setup to avoid the warning, so thanks for pointing that out!!

core/lib/spree/preferences/configuration.rb
waiting-for-dev: I think you're right. `solidus_installed?` was confusing, as the gem `solidus` is, of course, always installed. But in my view `is_currently_installing_solidus?` is also confusing because it doesn't need to be forcefully the case. I renamed it to `has_install_generator_been_run?`, which I think captures better what we're trying to detect.

About moving it to a different commit, the problem is that we need the check to make the CI green. Otherwise, it will fail because of the warning on the test suite. Even if we add both commits in this PR, I think it's always nice to keep a green history at every step. However, we can update the message for the single commit for this PR (once they're squashed). WDYT?

waiting-for-dev:
https://user-images.githubusercontent.com/52650/123631577-316ce680-d817-11eb-873d-7331f0c72372.mp4

This is a short demo showcasing how the process works

core/lib/spree/preferences/configuration.rb
kennyadsl: I think the new method name makes sense!

For the different commit, I think that now that the method name is addressed, if we want to keep it in the initial commit, it is not a big deal.

core/lib/spree/core/preference_changes_between_solidus_versions.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Spree
+  module Core
+    class PreferenceChangesBetweenSolidusVersions
```
aldesantis, line 5: Could we maybe have a shorter name for this, e.g. `PreferenceDifferentiator`? Also, this could live under `Spree::Preferences` and be called `Spree::Preferences::Differentiator`. What do you think?

core/lib/generators/solidus/update/update_generator.rb
```diff
@@ -29,6 +38,12 @@ class UpdateGenerator < ::Rails::Generators::Base
                  hide: true
 
     def create_new_defaults_initializer
+      previous_version_prompt = options[:previous_version_prompt]
+      return if previous_version_prompt && !yes?(<<~MSG, :red)
```
aldesantis, line 42: I'm not sure I understand why this is needed. Shouldn't this work with any Solidus version?

aldesantis (changes requested):

core/lib/spree/core/preference_changes_between_solidus_versions.rb
waiting-for-dev: Done @aldesantis!

waiting-for-dev: I think it makes sense :slightly_smiling_face:

core/lib/generators/solidus/update/update_generator.rb
waiting-for-dev: Yes, it should work with any Solidus version. However, we agreed with @kennyadsl that we should only encourage upgrading from the last minor version. On top of that, it allows us to assume the `from` version so that users don't need to provide it manually, which is not too ergonomic.

aldesantis: @waiting-for-dev okay, understood! In that case, should we make it clear that you can still upgrade from an earlier version, but it's not recommended? With the current messaging, it looks like the process won't work at all if you upgrade from an earlier version.

waiting-for-dev: I updated the message and removed the `hide: true` option from the `from` argument. Probably you're right. Still, I can't help feeling a bit uncomfortable :slightly_smiling_face:

core/lib/generators/solidus/update/update_generator.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    class_option :initializer_directory,
+                 type: :string,
+                 default: 'config/initializers/',
+                 hide: true
+
+    def create_new_defaults_initializer
+      previous_version_prompt = options[:previous_version_prompt]
+      return if previous_version_prompt && !yes?(<<~MSG, :red)
+        The update process is only supported if you are coming from version #{FROM}. If this is not the case, please, skip it and update your application to use Solidus #{FROM} before retrying.
```
aldesantis, line 43: What if, instead of having the `previous_version_prompt` option, we always showed the warning, but only when the user passed a `--from` version that's older than the previous minor version?

So if I'm on Solidus 2.11 and run the update generator with no options (or if I run it with `--from=2.10`), I don't get the warning. But if I run it with `--from=2.9`, I get the warning and I can decide if I want to continue or not.

I would also update this message: it doesn't make much sense to tell people that a use case is unsupported but still allow them to continue, imo. We could tell them something like:

> You specified you want to upgrade from Solidus [from version] to Solidus [current version], but the upgrade process is tested to be run one minor release at a time. Instead of skipping versions, we strongly recommend upgrading to Solidus [from version + 1 minor], fixing any deprecations and then continuing from there.
>
> If you are confident you want to upgrade all the way from Solidus [from version] to Solidus [current version] instead, you can still do it, although the process might be more complicated.
> 
> Are you sure you want to continue? (y/N)

core/lib/spree/preferences/configuration.rb
```diff
@@ -38,18 +38,35 @@ class Configuration
...
+      @load_defaults_called = true
       reset
     end
 
+    def check_load_defaults_called(instance_constant_name = nil)
+      return if load_defaults_called || !Spree::Core.has_install_generator_been_run?
+
+      target_name = instance_constant_name || "#{self.class.name}.new"
+      Spree::Deprecation.warn <<~MSG
+        Please, indicate wich Solidus version defaults the application must use.
```
aldesantis, line 62: Typo: wich → which.

Also, I would reword as follows (for simplicity):

> It's recommended that you explicitly load the default configuration for your current Solidus version. You can do it by adding the following to your Solidus initializer:
>
> ...

core/spec/lib/spree/preferences/preference_differentiator_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'spree/preferences/preference_differentiator'
+require 'spree/preferences/configuration'
+
+RSpec.describe Spree::Preferences::PreferenceDifferentiator do
+  it 'includes defaults that have changed' do
```
aldesantis, line 8: Should we add a test for new options too?

aldesantis (changes requested):

core/lib/generators/solidus/update/update_generator.rb
waiting-for-dev: > What if, instead of having the previous_version_prompt option, we always showed the warning, but only when the user passed a --from version that's older than the previous minor version?

That would mean that in the case the `--from` version is not specified, we'll print no warning. However, it's precisely those users who we want to address the most. The idea was to document a seamless upgrade process to bump from a minor version to the next one and then run `bin/rails g solidus:update`. If a user is providing the `--from` option, probably they know what they're doing.

We can go back to square one on this subject and make it mandatory that the `--from` option is supplied. We decided not to do for a couple of reasons:

- It's not ergonomic or common in the community to provide the version you're upgrading from.
- It conveys the idea that you can upgrade from whatever older version (although it'd be heavily mitigated by the warning message).

The benefit would be that users will forcefully be more conscious of what they're doing, but not sure if it outweighs the cons.

core/lib/spree/preferences/configuration.rb
waiting-for-dev: Done!

core/spec/lib/spree/preferences/preference_differentiator_spec.rb
waiting-for-dev: This class takes a single `Configuration` class, which can define values that have changed over time. They are defined with the `by_version` method, which takes an initial value and boundaries where it changed. E.g.:

```ruby
by_version(true, "3.0.0" => false)
```

If we introduce a new preference on Solidus that breaks some behavior, it'll be equivalent to say that the preference had a different value on past versions, so the `by_version` definition could be the same as the previous example.

core/lib/generators/solidus/update/update_generator.rb
aldesantis: You're right... Let's keep it like this for now, it should be easy enough to change it in the future without breaking anything, if needed.

kennyadsl (approved):

Issue #4095, opened 2021-06-09
Corrupted uploads with ActiveStorageAdapter crash the store


ok32:
I'm having a situation with attachments once again.
I've once opened a [discussion](https://github.com/solidusio/solidus/discussions/3986) about corrupted uploads, and the solution that I offered there has already been PR'd and merged into master.

But now I've encountered a slightly different case. Turns out the user uploaded an image successfully, and then tried to do something with the stock – the URL was `/api/stock_locations/1/stock_items/58` and there's `StockItemsController` in the backtrace. And it caused some exception in `ActiveStorage` (as seen in the Airbrake error notification email):

```
ERROR MESSAGE	No such file or directory @ rb_file_s_size - /tmp/image_processing20210608-1-1lsc5ns.jpg

BACKTRACE (a part of)

/usr/local/lib/ruby/2.7.0/tempfile.rb:226 in size
/usr/local/lib/ruby/2.7.0/tempfile.rb:226 in size
/GEM_ROOT/gems/activestorage-6.1.3.2/lib/active_storage/service/s3_service.rb:30 in block in upload

...

/GEM_ROOT/gems/solidus_api-3.0.1/app/controllers/spree/api/stock_items_controller.rb:42 in update

...
```

And all this resulted in a situation where, again, every page which contains images of this product crashes with an exception (admin pages too). The exception is raised here:

```ruby
module Spree
  module ActiveStorageAdapter
    # Decorares AtiveStorage attachment to add methods exptected by Solidus'
    # Paperclip-oriented attachment support.
    class Attachment

      # ...

      def variant(style = nil)
        size = style_to_size(style)
        @attachment.variant(
          resize_to_limit: size,
          strip: true
        ).processed # <---- HERE! : "undefined method `processed' for nil:NilClass"
      end

      # ...

    end
  end
end
```

So `@attachment` itself is not nil (otherwise `.variant()` would not have be called in the first place), but `.variant()` returns `nil`.

So the question is, how come? What could cause such a corruption? Can it be fixed? Where?

**Solidus Version**: 3.0.1

**To Reproduce**
No idea.

**Current behavior**
Corrupted attachment causes all pages containing it to crash. Impossible to fix from the UI.

**Expected behavior**
I expect this situation to be admin-recoverable. If something's gone wrong during attachment manipulation the admin should be able to remove/reupload the corrupted one.

cpfergus1:
[...]

PR #4103, opened 2021-06-15
Prevent UI crash on FileNotFound errors with Active Storage

cpfergus1:
**Description**
In a previous PR #4026, a fix was implemented to prevent an application from crashing by capturing an exception and providing a replacement image to prevent a non-operable UI. In issue #4095 it was found that the implemented fix did not capture all failure points that may be caused by a corrupt or FileNotFound issue.

This commit adds the ability for the application to catch missing images with Active Storage during an API call should a file become corrupt or deleted without removing the record. This compliments PR #4026 which prevented the same error on the backend. The change will return a json image package with attributes equal to nil and noimage replacements if an
image cannot be sourced. This should be back-ported to 3.0 if cleared for merging. 

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/active_storage_adapter/attachment.rb
+++ b/core/app/models/concerns/spree/active_storage_adapter/attachment.rb
@@ -59,6 +59,10 @@ module Spree
         analyze unless analyzed?
 
         @attachment.metadata
+      rescue ActiveStorage::FileNotFoundError => error
+        logger.error("#{error} - Image id: #{attachment.record.id} is corrupted or cannot be found")
+
+        { identified: nil, width: nil, height: nil, analyzed: true }
       end
 
       def styles_to_transformations(styles)
--- a/api/app/views/spree/api/images/_image.json.jbuilder
+++ b/api/app/views/spree/api/images/_image.json.jbuilder
@@ -3,5 +3,5 @@
 json.(image, *image_attributes)
 json.(image, :viewable_type, :viewable_id)
 Spree::Image.attachment_definitions[:attachment][:styles].each do |key, _value|
-  json.set! "#{key}_url", image.attachment.url(key)
+  json.set! "#{key}_url", image.url(key)
 end
--- a/api/spec/requests/spree/api/images_spec.rb
+++ b/api/spec/requests/spree/api/images_spec.rb
@@ -74,6 +74,17 @@ module Spree::Api
           expect(response.status).to eq(204)
           expect { product_image.reload }.to raise_error(ActiveRecord::RecordNotFound)
         end
+
+        it "returns nil attribute values and noimage urls when the image cannot be found",
+          if: Spree::Config.image_attachment_module == Spree::Image::ActiveStorageAttachment do
+          product_image.attachment.blob.update(key: 11)
+          expect(Rails.logger).to receive(:error).with(/Image id: #{product_image.id} is corrupted or cannot be found/).twice
+          get spree.api_variant_images_path(product.master.id)
+          expect(response.status).to eq(200)
+          expect(json_response[:images].first[:attachment_width]).to be_nil
+          expect(json_response[:images].first[:attachment_height]).to be_nil
+          expect(json_response[:images].first[:product_url]).to include("noimage")
+        end
       end
 
       context 'when image belongs to another product' do


api/spec/requests/spree/api/images_controller_spec.rb
```diff
@@ -74,6 +74,16 @@ module Spree
           expect(response.status).to eq(204)
           expect { product_image.reload }.to raise_error(ActiveRecord::RecordNotFound)
         end
+
+        it "It returns nil attribute values and noimage urls when the image cannot be found",
```
kennyadsl, line 78: `It` is not required in the example text

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
```diff
@@ -53,8 +53,9 @@ def destroy
 
       def metadata
         analyze unless analyzed?
-
         @attachment.metadata
+      rescue ActiveStorage::FileNotFoundError
+        { identified: nil, width: nil, height: nil, analyzed: false }
```
kennyadsl, line 58: @elia @filippoliverani any thoughts on this? I can't think of a better way to fix this problem but maybe we already faced something similar or have other ideas?

api/spec/requests/spree/api/images_controller_spec.rb
cpfergus1: done!

spaghetticode (commented):
@cpfergus1 thanks for working on this issue! 

Unless we want to patch [ActiveStorage](https://github.com/rails/rails/blob/581aac754c6cbe9f3f4de396bc8ef3728e56f4f4/activestorage/lib/active_storage/analyzer/image_analyzer/image_magick.rb#L12) directly, this change looks reasonable to me. 
Also, having `nil` instead of `0` for `height` and `width` seems to be the [right way](https://github.com/rails/rails/blob/581aac754c6cbe9f3f4de396bc8ef3728e56f4f4/activestorage/test/analyzer/image_analyzer/image_magick_test.rb#L44) to me. 
On the other hand, not really sure we want `nil` for `identified` and `analyzed`, for example the latter seems to be always set to `true` by [rails](https://github.com/rails/rails/blob/581aac754c6cbe9f3f4de396bc8ef3728e56f4f4/activestorage/app/models/active_storage/blob/analyzable.rb#L51), after extracting the image metadata.
Do you have any opinions on this?

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
waiting-for-dev: @elia @filippoliverani pinging you again to move it forward :slightly_smiling_face:

filippoliverani: The solution makes sense to me, but I'd prefer to let the error bubble up and handle it where the attachment is used. That's a breaking error, and different controllers or views might want to handle it in different ways, returning a null object might be misleading 🤔

kennyadsl: I'm not sure about this solution, but please correct me if I'm wrong. If there are 20 images on the page and only one of them is unavailable because of this problem, bubbling up the exception to the controller/view layer we are not able to display the other 19 images. We can only show an error to the user.

waiting-for-dev: Don't you think that could be the desired behavior sometimes? Depending on the model, you might want to catch that issue instead of silencing it. For instance, as an administrator, you want to notice when you upload an image for a product, but it's corrupted.

But I made an error when I said, "we could manage it at the controller layer with a configurable error handler". We should just make returning a null object an optional behavior. If someone asks for it, we can do more work and make it configurable on a model basis.

kennyadsl: My opinion here is that, if this failure will prevent people to buy things from the store, I think that showing a broken image is a better solution. So maybe storefront and APIs should never break because of this. 

On the other side, the admin should notice that there's something wrong, but if we just show the error, what can they do to fix it? I'm afraid that they need to ask some devs to manually destroy the record and try to reupload the image, and probably it's better to keep the admin working until that happens.

waiting-for-dev: > My opinion here is that, if this failure will prevent people to buy things from the store, I think that showing a broken image is a better solution. So maybe storefront and APIs should never break because of this.

What I meant is: A user wants to buy a shirt, but, as the image is not rendered, they don't buy it. As no error is logged, the admin will never know about the problem unless they revisit the product page for some reason.

> On the other side, the admin should notice that there's something wrong, but if we just show the error, what can they do to fix it?

That's an excellent point. Admin shouldn't break. We need different behavior depending on the component.

A good middle-way solution could be returning a nil image but showing a warning so that it's logged and can be detected. What do you think?

kennyadsl: Yes, or delete the corrupted image and continue with the execution. Could be dangerous, I don't know the exact implications of this proposal but if I got it correctly, admin will need to reupload the image again in this case because it's not saved in the storage provider. Why keep an instance in our system for that image the actual image is nowhere?

waiting-for-dev: Not sure, but maybe the stored metadata could help debug?

kennyadsl: Probably, I don't even know if this is reproducible, to be honest.

waiting-for-dev: @cpfergus1, when you have a moment, could you please add a warning when the image is not found? 🙏

waiting-for-dev: Hey, @cpfergus1, pinging you again in case you have time to put the little extra effort needed to complete this 💪

cpfergus1: Sorry, I missed the request, I will pick it back up!

cpfergus1:
> @cpfergus1 thanks for working on this issue!
> 
> Unless we want to patch [ActiveStorage](https://github.com/rails/rails/blob/581aac754c6cbe9f3f4de396bc8ef3728e56f4f4/activestorage/lib/active_storage/analyzer/image_analyzer/image_magick.rb#L12) directly, this change looks reasonable to me. Also, having `nil` instead of `0` for `height` and `width` seems to be the [right way](https://github.com/rails/rails/blob/581aac754c6cbe9f3f4de396bc8ef3728e56f4f4/activestorage/test/analyzer/image_analyzer/image_magick_test.rb#L44) to me. On the other hand, not really sure we want `nil` for `identified` and `analyzed`, for example the latter seems to be always set to `true` by [rails](https://github.com/rails/rails/blob/581aac754c6cbe9f3f4de396bc8ef3728e56f4f4/activestorage/app/models/active_storage/blob/analyzable.rb#L51), after extracting the image metadata. Do you have any opinion on this?

My thoughts on this are that if the object cannot be found, it cannot be analyzed or identified since it doesn't exist. This behavior can be witnessed on [images_controller_spec.rb](api/spec/requests/spree/api/images_controller_spec.rb) when examining the image blob. If you attempt to run analyze on it - it will fail due to the image not being found, so the object has not been analyzed.

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
cpfergus1: I have added a point to capture this error in the logs, did we still want this to bubble up to the controller?

waiting-for-dev: Thanks, @cpfergus1! No, I think it's better not to bubble up to the controller at this point.

waiting-for-dev (approved):
Many thanks for your continuous work on it, @cpfergus1! A last minor suggestion, it would be nice to add the id of the attachment in the warning message, so that users can't more easily find it.

It'd be also better to rebase from master before merging :pray:

waiting-for-dev (approved):

waiting-for-dev:
Thanks, @cpfergus1. We only need another review. Pinging @kennyadsl as he also looked into it.

kennyadsl (approved):

PR #4104, opened 2021-06-18
Update billing address migration tasks with batch limit

spaghetticode:
**Description**

When running the task on large DBs it makes sense to limit the number of records updated on each transaction in order to avoid locking the records (or the whole table) for too long.

The default batch size is 100'000 and it can be overridden by passing a different value when invoking the task. For example,
this will set the batch size to 1000 rows:

```
rake solidus:migrations:migrate_default_billing_addresses_to_address_book:up[1000]
```

**Checklist:**


kennyadsl (approved):
Thanks, Andrea!

jarednorman (approved):

PR #4105, opened 2021-06-18
Simplify fetching order to calculate rates

### LINKED ISSUE #4046: Simplify fetching the order to calculate rates for proposed shipments
I'm not super-familiar with the logic around this issue, so maybe I'm missing something. However, if my understanding is correct, this is a proposal for an internal simplification.

[`Spree::Stock::SimpleCoordinator#build_shipments`](https://github.com/solidusio/solidus/blob/2ea829645b00fcedd5bfd69e045bddab7f40beb9/core/app/models/spree/stock/simple_coordinator.rb#L50) is meant to build shipments presented as an option to the user. To calculate their associated shipping rates, it [delegates the shipments](https://github.com/solidusio/solidus/blob/2ea829645b00fcedd5bfd69e045bddab7f40beb9/core/app/models/spree/stock/simple_coordinator.rb#L78) to  [`Spree::Stock::Estimator`](https://github.com/solidusio/solidus/blob/2ea829645b00fcedd5bfd69e045bddab7f40beb9/core/app/models/spree/stock/estimator.rb), through a [previously generated `Spree:::Stock::Package`](https://github.com/solidusio/solidus/blob/2ea829645b00fcedd5bfd69e045bddab7f40beb9/core/app/models/spree/stock/simple_coordinator.rb#L65). This service needs to know the order instance those shipments would be assigned to. With the current implementation, this information is taken from the shipment itself as it's already associated with that order when [it's initialized on  `Spree::Stock::Package`](https://github.com/solidusio/solidus/blob/2ea829645b00fcedd5bfd69e045bddab7f40beb9/core/app/models/spree/stock/package.rb#L127).

 Before the `has_many_inversing` new Rails setting (https://github.com/rails/rails/pull/34533 & https://github.com/rails/rails/pull/37429), this initialization wasn't reflected in the inverse association `Spree::Order#shipments`, but it's no longer the case. For this reason, after we have calculated the shipping rates, we need to remove the shipping instances through `order.shipments = order.shipments - shipment`. This is part of what we have done in #4035.

As an improvement, we could pass the order as a parameter to the estimator and avoid us depending on the complex AR associations cache. As users can provide their own package class, it could break behavior if we change the initializer signature. However, we can keep backward compatibility until the next major release if we make the new argument optional and keep associating the order for now.

HendrikLouw:
**Description**

resolves #4046 

**Checklist:**

### DIFF
--- a/core/app/models/spree/stock/estimator.rb
+++ b/core/app/models/spree/stock/estimator.rb
@@ -3,8 +3,13 @@
 module Spree
   module Stock
     class Estimator
+      attr_reader :order
       class ShipmentRequired < StandardError; end
       class OrderRequired < StandardError; end
+
+      def initialize(order = nil)
+        @order = order
+      end
 
       # Estimate the shipping rates for a package.
       #
@@ -15,8 +20,14 @@ module Spree
       #   descending cost, with the least costly marked "selected"
       def shipping_rates(package, frontend_only = true)
         raise ShipmentRequired if package.shipment.nil?
-        raise OrderRequired if package.shipment.order.nil?
+        raise OrderRequired if package.shipment.order.nil? && @order.nil?
 
+        if order && package.shipment.order.nil?
+          package.shipment.order = order
+        elsif package.shipment.order && order.nil?
+          @order = package.shipment.order
+        end
+        
         rates = calculate_shipping_rates(package)
         rates.select! { |rate| rate.shipping_method.available_to_users? } if frontend_only
         choose_default_shipping_rate(rates)
@@ -34,7 +45,7 @@ module Spree
 
       def calculate_shipping_rates(package)
         tax_calculator_class = Spree::Config.shipping_rate_tax_calculator_class
-        tax_calculator = tax_calculator_class.new(package.shipment.order)
+        tax_calculator = tax_calculator_class.new(order)
         shipping_methods(package).map do |shipping_method|
           cost = shipping_method.calculator.compute(package)
           if cost
@@ -55,15 +66,15 @@ module Spree
 
       def shipping_methods(package)
         package.shipping_methods
-          .available_to_store(package.shipment.order.store)
-          .available_for_address(package.shipment.order.ship_address)
+          .available_to_store(order.store)
+          .available_for_address(order.ship_address)
           .includes(:calculator)
           .to_a
           .select do |ship_method|
           calculator = ship_method.calculator
           calculator.available?(package) &&
             (calculator.preferences[:currency].blank? ||
-             calculator.preferences[:currency] == package.shipment.order.currency)
+             calculator.preferences[:currency] == order.currency)
         end
       end
     end
--- a/core/app/models/spree/stock/estimator.rb
+++ b/core/app/models/spree/stock/estimator.rb
@@ -27,7 +27,7 @@ module Spree
         elsif package.shipment.order && order.nil?
           @order = package.shipment.order
         end
-        
+
         rates = calculate_shipping_rates(package)
         rates.select! { |rate| rate.shipping_method.available_to_users? } if frontend_only
         choose_default_shipping_rate(rates)
--- a/core/app/models/spree/stock/estimator.rb
+++ b/core/app/models/spree/stock/estimator.rb
@@ -8,6 +8,9 @@ module Spree
       class OrderRequired < StandardError; end
 
       def initialize(order = nil)
+        Spree::Deprecation.warn('A Spree::Order is required as an argument,
+          please initialize with Spree::Stock::Estimator.new(order),
+          Spree::Stock::Estimator.new will throw an ArgumentError after deprecation')
         @order = order
       end
 
@@ -20,7 +23,7 @@ module Spree
       #   descending cost, with the least costly marked "selected"
       def shipping_rates(package, frontend_only = true)
         raise ShipmentRequired if package.shipment.nil?
-        raise OrderRequired if package.shipment.order.nil? && @order.nil?
+        raise OrderRequired if package.shipment.order.nil? && order.nil?
 
         if order && package.shipment.order.nil?
           package.shipment.order = order
--- a/core/app/models/spree/stock/simple_coordinator.rb
+++ b/core/app/models/spree/stock/simple_coordinator.rb
@@ -75,7 +75,7 @@ module Spree
         # Turn the Stock::Packages into a Shipment with rates
         shipments = packages.map do |package|
           shipment = package.shipment = package.to_shipment
-          shipment.shipping_rates = Spree::Config.stock.estimator_class.new(order).shipping_rates(package)
+          shipment.shipping_rates = Spree::Config.stock.estimator(order).shipping_rates(package)
           shipment
         end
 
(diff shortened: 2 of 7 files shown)


core/app/models/spree/stock/estimator.rb
```diff
@@ -15,8 +20,14 @@ class OrderRequired < StandardError; end
       #   descending cost, with the least costly marked "selected"
       def shipping_rates(package, frontend_only = true)
         raise ShipmentRequired if package.shipment.nil?
-        raise OrderRequired if package.shipment.order.nil?
+        raise OrderRequired if package.shipment.order.nil? && @order.nil?
```
aldesantis, line 23: Is there a reason for using `@order` instead of `order` here?

core/app/models/spree/stock/estimator.rb
```diff
@@ -3,9 +3,14 @@
 module Spree
   module Stock
     class Estimator
+      attr_reader :order
       class ShipmentRequired < StandardError; end
       class OrderRequired < StandardError; end
 
+      def initialize(order = nil)
+        @order = order
```
aldesantis, line 11: Should we print a deprecation warning telling people that initializing the estimator without `order` is not going to be supported in the next major?

core/app/models/spree/stock/estimator.rb
```diff
@@ -15,8 +20,14 @@ class OrderRequired < StandardError; end
...
       def shipping_rates(package, frontend_only = true)
         raise ShipmentRequired if package.shipment.nil?
-        raise OrderRequired if package.shipment.order.nil?
+        raise OrderRequired if package.shipment.order.nil? && @order.nil?
 
+        if order && package.shipment.order.nil?
+          package.shipment.order = order
+        elsif package.shipment.order && order.nil?
+          @order = package.shipment.order
+        end
```
aldesantis, line 29: What if `package.shipment.order` and `order` are both set but are two different objects? Should we think about that case?

core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -75,7 +75,7 @@ def build_shipments
         # Turn the Stock::Packages into a Shipment with rates
         shipments = packages.map do |package|
           shipment = package.shipment = package.to_shipment
-          shipment.shipping_rates = Spree::Config.stock.estimator_class.new.shipping_rates(package)
+          shipment.shipping_rates = Spree::Config.stock.estimator_class.new(order).shipping_rates(package)
```
aldesantis, line 78: People may have a custom stock estimator which does not accept the `order` argument. We need to make sure we play nice with that scenario, otherwise this will be a breaking change.

aldesantis (changes requested):

core/app/models/spree/stock/estimator.rb
HendrikLouw: @aldesantis  I think that if the `order` is passed into the constructor it would make sense to give it preference over `package.shipment.order`, I noted that decision in the commit here https://github.com/solidusio/solidus/pull/4105/commits/7506823349f011769c882ffb07151586f234738c

core/app/models/spree/stock/simple_coordinator.rb
HendrikLouw: The current test coverage provides enough confidence to make sure we're not breaking anything for custom stock estimators, the existing tests still test the case when there is no `order` passed into the constructor and an additional test was added to ensure we use the initializer's `order` if it's passed in https://github.com/solidusio/solidus/blob/7414ae6d2f177610d272130143e92d7e6023dc9d/core/spec/models/spree/stock/estimator_spec.rb#L36

core/app/models/spree/stock/estimator.rb
HendrikLouw: I'll add a deprecation message here, I couldn't find any documentation that specifies how to add your own custom estimator so the usage is probably low.

core/app/models/spree/stock/simple_coordinator.rb
aldesantis: @HendrikLouw what I meant is that, if someone provides a custom stock estimator and overrides the `#initialize` method (or if their estimator doesn't inherit from `Spree::Stock::Estimator` at all), their stock estimator's `#initialize` will not accept an `order` argument, but Solidus will still attempt to initialize it with an order argument, which will break the user's code.

Take the following example:

```ruby
# app/models/custom_estimator.rb
class CustomEstimator
  def shipping_rates
    # ...
  end
end

Spree.config do |config|
  config.stock.estimator_class = 'CustomEstimator'
end

# Somewhere in the code, assuming `order` is a `Spree::Order` instance
Spree::Stock::SimpleCoordinator.new(order).shipments # <= This will raise an error
```

core/app/models/spree/stock/estimator.rb
aldesantis: @HendrikLouw overriding the stock estimator is pretty common practice for stores with advanced stock logic. Some extensions even provide their own estimator, e.g. [solidus_easypost](https://github.com/solidusio-contrib/solidus_easypost/blob/master/lib/solidus_easypost/estimator.rb) does it.

core/app/models/spree/stock/estimator.rb
aldesantis: @HendrikLouw understood, but what if they're both passed and they're different objects? Is that a data integrity issue, is that something we should be concerned about, etc.?

core/app/models/spree/stock/estimator.rb
HendrikLouw: Good spot, I think I used @order before adding the `attr_accessor` and didn't update it before pushing up, all fixed now,

core/app/models/spree/stock/estimator.rb
HendrikLouw: Got it, thank you for the information, I've added a deprecation message now and that is highlighting that a lot of our tests are not passing in the `order` to the initialiser.  Do you think we should update the tests as part of this change?

core/app/models/spree/stock/estimator.rb
HendrikLouw: @aldesantis I think it makes sense to make the order passed into the initializer the default, the aim is to remove the usage of `package.shipment.order` and forcing the usage of the passed in order promotes that goal. There shouldn't be any data integrity issues if all the packages always contain the same order. 

If the assumption that packages always have the same order holds then the tests document the correct behaviour if there can be different orders per package then this whole PR and issue should be abandoned, since I don't have enough context here I'm assuming that the creator of the issue took this into account.

core/app/models/spree/stock/simple_coordinator.rb
HendrikLouw: Thank you for the full explanation, I've added a way to handle this case https://github.com/solidusio/solidus/pull/4105/files#diff-43c0bdb931bf364be432145396b649919cfee1ce27e82af572bf394f47af8cdaR22, happy with the solution?

core/app/models/spree/stock/estimator.rb
HendrikLouw: @aldesantis I see CI doesn't allow the deprecation message, I'll fix this up tonight.

PR #4108, opened 2021-06-21
Ignore discarded prices on Variant#default_price

waiting-for-dev:
**Description**

Fixes bug where discarded prices were taken into account in order to
tell the default price for a variant.

The rationale behind the previous behavior was to be able to render a
price when discarded products are included in the backend products
listing page (see 'Show deleted' checkbox near the top-right edge). It
was introduced in commit 87cef09.

However, this logic was incorrect in two ways:

- It affected non-discarded variants, where we don't want to consider
  discarded prices.
- Even for discarded variants, it could lead to display the wrong
  information. As all prices get discarded when a variant is discarded,
  there's no guarantee that the one that is rendered was the previously
  considered as the default.

The list of products on the backend will render the price field for
discarded products as blank. In case we want to track default prices, we
should add a flag to uniquely identify them.

Extracted from #3994

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -33,7 +33,7 @@ module Spree
     # Select from {#prices} the one to be considered as the default
     #
     # This method works with the in-memory association, so non-persisted prices
-    # are taken into account. Discarded prices are also considered.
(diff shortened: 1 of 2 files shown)


waiting-for-dev:

PR #4109, opened 2021-06-21
Rename currently_valid to prioritized_for_default

waiting-for-dev:
**Description**

It conveys better what it's doing, as the query only consists of `ORDER`
clauses and no record is filtered out.

Extracted from #3994

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/default_price.rb
+++ b/core/app/models/concerns/spree/default_price.rb
@@ -17,8 +17,8 @@ module Spree
     # Returns `#prices` prioritized for being considered as default price
     #
     # @return [ActiveRecord::Relation<Spree::Price>]
-    def currently_valid_prices
-      prices.currently_valid
+    def prioritized_for_default_prices
+      prices.prioritized_for_default
     end
 
     # Returns {#default_price} or builds it from {Spree::Variant.default_price_attributes}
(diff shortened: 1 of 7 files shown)


core/app/models/concerns/spree/default_price.rb
```diff
@@ -17,8 +17,8 @@ def self.default_price_attributes
     # Returns `#prices` prioritized for being considered as default price
     #
     # @return [ActiveRecord::Relation<Spree::Price>]
-    def currently_valid_prices
```
kennyadsl, line 20: I think we should deprecate this method instead of removing it.

core/app/models/spree/price.rb
```diff
@@ -20,7 +20,7 @@ class Price < Spree::Base
     validates :currency, inclusion: { in: ::Money::Currency.all.map(&:iso_code), message: :invalid_code }
     validates :country, presence: true, unless: -> { for_any_country? }
 
-    scope :currently_valid, -> { order(Arel.sql("country_iso IS NULL")).order(updated_at: :DESC, id: :DESC) }
```
kennyadsl, line 23: Let's deprecate this as well!

kennyadsl (changes requested):
Nice, but I think we need some deprecations here.

waiting-for-dev:

PR #4114, opened 2021-06-23
Fix flaky product feature spec #4110

### LINKED ISSUE #4110: Flaky Feature Spec detected
We currently have a flaky feature spec related to products. It can be inspected here: 

https://app.circleci.com/pipelines/github/solidusio/solidus/2519/workflows/b06934e2-9055-44ba-960b-37ed4345939b/jobs/25240

It is happening quite often, see [all the latest failures in this list](https://app.circleci.com/pipelines/github/solidusio/solidus) for more occurrences. At a first look, it doesn't seem related to the database used.

gsmendoza:
## Ticket

https://github.com/solidusio/solidus/pull/4110

## Demo of fix

https://www.dropbox.com/s/6yax0jcknnurbz1/4110-flaky-feature-spec-detected-test-2021-06-23_15.33.37.mp4?dl=0

## Description

I still haven't found the root cause of the issue, but it seems that visiting the root path only once per spec prevents the failure from occurring.

In https://app.circleci.com/pipelines/github/solidusio/solidus/2527/workflows/e98388e1-970b-4600-974b-bd2ced1e095e/jobs/25276/parallel-runs/2?filterBy=ALL, the spec ran 100 times and did not result in a failure.

In contrast, if you run the same spec in https://github.com/nebulab/solidus/tree/gsmendoza/4110-flaky-feature-spec-detected-original-spec-100-times, the same spec is more likely to fail.

## Checklist

- [ ] I have updated Guides and README accordingly to this change (if needed) - N/A
- [ ] I have attached screenshots to this PR for visual changes (if needed) - N/A

### DIFF
--- a/frontend/spec/features/products_spec.rb
+++ b/frontend/spec/features/products_spec.rb
@@ -9,11 +9,8 @@ describe "Visiting Products", type: :feature, inaccessible: true do
     ((first_store = Spree::Store.first) && first_store.name).to_s
   end
 
-  before(:each) do
-    visit spree.root_path
-  end
-
   it "should be able to show the shopping cart after adding a product to it" do
+    visit spree.root_path
     click_link "Ruby on Rails Ringer T-Shirt"
     expect(page).to have_content("$19.99")
 
@@ -31,8 +28,8 @@ describe "Visiting Products", type: :feature, inaccessible: true do
     end
 
     it "should use *_path helper to generate the product links" do
-     visit spree.root_path
-     expect(page).to have_xpath(".//a[@href='#{spree.product_path(product)}']")
+      visit spree.root_path
+      expect(page).to have_xpath(".//a[@href='#{spree.product_path(product)}']")
     end
   end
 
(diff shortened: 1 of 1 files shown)


frontend/spec/features/products_spec.rb
```diff
@@ -9,11 +9,8 @@
     ((first_store = Spree::Store.first) && first_store.name).to_s
   end
 
-  before(:each) do
```
kennyadsl, line 12: This should do exactly the same as manually visiting the path into each test. It is not happening once for all tests (that's `before(:all)` if I recall correctly).

kennyadsl (changes requested):
Thanks but looking at the code I think the issue is [here]( https://github.com/solidusio/solidus/blob/ab980996cf3fcb3c4f9111166798e6a126ec8ddd/frontend/spec/features/products_spec.rb#L194-L202) instead. 

What's happening is (if I recall correctly how the before execution order works):

- page visited with the `before(:each)`
- image created with the `before(:each)` I linked above
- image is not there because it's created after the page is rendered

Still not sure why it happens intermittently though.

gsmendoza:
Thanks for the feedback @kennyadsl. Let me close this first and resubmit it using my nebulab email. I might also ask to discuss with you the PR offline.

PR #4117, opened 2021-06-24
Make frontend ContentController RESTful

RyanofWoods:
This change renames `ContentController` to `CardCvvController`, its `cvv`
method and corresponding view to `index`. This allows the controller to be
RESTful. As well as preventing a confusing similar name to
the new controller `OrderContentsController`.

It also adds a feature spec which was missing: User opens up a new window about card cvv information from the checkout payment section.

**Note:**
The route could be changed from:
```ruby
get '/card_cvv/index', to: 'card_cvv#index', as: :card_cvv
```
to:
```ruby
resources card_cvv, only [:index]
``` 
This would however mean the route would be 'card_cvv_index' as the resource is not plural. If this is more desirable anyway, let me know 😃 

**Checklist:**
- [ ] I have updated Guides and README accordingly to this change (N/A)
- [ ] I have attached screenshots to this PR for visual changes (N/A)


RyanofWoods:
Hi @kennyadsl! 😊 Perhaps it is inappropriate for this controller to be RESTful as it implies that it is based on a model, but instead it is just to render a static view. Would it be better to just keep it non RESTful but have a more descriptive controller name that doesn't so easily clash with `OrderContentsController`?

jarednorman (commented):
This seems like a totally reasonable change, but I'm not sure if there's been any solid discussion into how we want to handle changes to frontend. Changes to frontend are the most liable to be annoying and disruptive to existing stores, and with some effort going into solidus_starter_frontend, I'm still wary to make changes like this that doesn't really meaningfully make anything better for users, but has the possibility to break references to the existing path for current users.

kennyadsl:
@jarednorman yes, we discussed this privately and we are about to move these changes to the starter frontend instead. All these improvements will find a better place in that project. Thanks for raising it.

PR #4121, opened 2021-06-26
Adds conditional validation to `ReturnItem`

### LINKED ISSUE #4086: Return item model: inventory_unit association is set as optional but there's also a validation presence
It looks like there's a conflict between the association definition and a validation rule.

The association in the `return_item` model:
`belongs_to :inventory_unit, inverse_of: :return_items, optional: true`
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/return_item.rb#L33

The validation rule some lines below:
`validates :inventory_unit, presence: true`
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/return_item.rb#L45

**Solidus Version**:
Master @ 5f35714

**Current behavior**
Using `shoulda-matchers` gem to check the association fails with or without `.optional`:
`it { is_expected.to belong_to(:inventory_unit).inverse_of(:return_items).touch(true).optional }`

_Expected Spree::ReturnItem to have a belongs_to association called inventory_unit (and for the record not to fail validation if :inventory_unit is unset; i.e., either the association should have been defined with `optional: true`, or there should not be a presence validation on :inventory_unit)_

While removing `optional`:
_Expected Spree::ReturnItem to have a belongs_to association called inventory_unit (and for the record to fail validation if :inventory_unit is unset; i.e., either the association should have been defined with `required: true`, or there should be a presence validation on :inventory_unit)_

**Expected behavior**
Perhaps we could remove the validation rule introduced with this old commit:
https://github.com/solidusio/solidus/commit/108e01331a35ec4a59e7253d55d3cce3339d0e50#diff-81e13a48112d990ac138e28c8ceebadc3bbb6c8db28b8c9eef50a0b23bcf9efaR20
But we need to make some extra checks, there are references in the model itself that could break in some situations.

Brian-Demon:
**Description**

An issue exists between the association and validation of
`ReturnItem.inventory_unit`. There are three valid states for
the association:

- A ReturnItem is valid when it has no inventory_unit
- A ReturnItem is valid when it has an inventory_unit that is valid
- A ReturnItem is invalid when it has an inventory_unit that is invalid

This adds a conditional to the validation so that it only applies when
`inventory_unit_id` is present.  Tests have been updated to reflect
these three cases.

Resolves #4086

**Checklist:**

### DIFF
--- a/core/app/models/spree/return_item.rb
+++ b/core/app/models/spree/return_item.rb
@@ -30,7 +30,7 @@ module Spree
     self.refund_amount_calculator = Calculator::Returns::DefaultRefundAmount
 
     belongs_to :return_authorization, inverse_of: :return_items, optional: true
-    belongs_to :inventory_unit, inverse_of: :return_items, optional: true
+    belongs_to :inventory_unit, inverse_of: :return_items
     belongs_to :exchange_variant, class_name: 'Spree::Variant', optional: true
     belongs_to :exchange_inventory_unit, class_name: 'Spree::InventoryUnit', inverse_of: :original_return_item, optional: true
     belongs_to :customer_return, inverse_of: :return_items, optional: true
@@ -42,7 +42,6 @@ module Spree
     validate :eligible_exchange_variant
     validate :belongs_to_same_customer_order
     validate :validate_acceptance_status_for_reimbursement
-    validates :inventory_unit, presence: true
     validate :validate_no_other_completed_return_items
 
     after_create :cancel_others, unless: :cancelled?


Brian-Demon:
The CI check failure appears to be a flaky failure outlined here: https://github.com/solidusio/solidus/issues/4110

I would re-run it myself but it doesn't appear to let me.

jarednorman:
I'm not clear on whether this is the desired approach and have left a comment over on the issue regarding this.

waiting-for-dev:
Hey @Brian-Demon. It's been a while, but would you like to re-take this one? As said in the related issue, we should go with option 1 from https://github.com/solidusio/solidus/issues/4086#issuecomment-873673898.

waiting-for-dev:
@solidusio/core-team, could we have a review here? I rebased from master and updated the PR along with the suggestion by @jarednorman. Tests are still green, so it looks like there are no side effects to this. Thanks for your collaboration, @Brian-Demon!

tvdeyen (approved):
Seems good to me.

aldesantis (approved):
Assuming that option 1 _is_ the way to go (which seems to make sense), this looks good to me!

jarednorman (approved):

Issue #4001, opened 2021-03-19
Remove controllers references from API request specs


kennyadsl:
A while ago, we converted API controller specs into requests specs, see #2052. But those specs still have references from their old "life" that we can clean a little bit with a small effort.

In fact, their filenames still have the `_controller_` part, which makes no sense now. Eg. `solidus/api/spec/requests/spree/api/addresses_controller_spec.rb`

Also, they contain references to the previous class tested as the described class. Eg.

```ruby
module Spree
  describe Api::AddressesController, type: :request do
  # ...
```

Let's clean them!

niamtokik:
Hi,

I would like to help. I am looking for a Ruby project to improve my level in this language. Seems a good issue to start.

> In fact, their filenames still have the `_controller_` part, which makes no sense now. Eg. `solidus/api/spec/requests/spree/api/addresses_controller_spec.rb`

Those files should now be renamed like `solidus/api/spec/requests/spree/api/addresses_request_spec.rb`, so, replacing `controller` by `request`? Or `_controller_` term should just be removed?

> Also, they contain references to the previous class tested as the described class. Eg.

If I take your example, module Spree should look like that after the cleanup:

```ruby
module Spree
  describe Api::AddressesRequest, type: :request do
  # ...
```

Other dependencies in the code should be in this case modified too, right?

jarednorman:
The code after your cleanup wouldn't work, as that constant doesn't exist.

@kennyadsl Do you have any thoughts on what the convention should be for those top-level describes in request specs?

kennyadsl:
@niamtokik @jarednorman I didn't have anything in mind. I think we can do some research and apply the current best practices in the Rails ecosystem.

tejasbubane:
I stubled upon this issue while looking for some beginner-level issues. I think this is resolved by #4158 and can be closed now?

kennyadsl:
You are right, thanks!

PR #4122, opened 2021-06-27
Rename files in api/spec/requests/spree/api and in api/spec/controllers/spree/api

AymanJabr:
In this PR I attempt to solve Issue #4001. This Issue is a follow up to PR #2052, where API controller specs have been converted to request specs.

The issue is about cleaning up references to controllers from the previous PR. This is mostly in the file names.

Since this is my first contribuition to an open project, it is a good Issue to try and fix. I did have some issues while trying to understand how everything fit toghether though, and I try to explain these issues below.

The first issue that I have with this Issue is that most of the files changed in the previous PR are not in the original (api/spec/controllers/spree/api/) path. Even the file called addresses_controller_spec.rb, given as an example by @Kennyadsl is not there anymore. This is the case both for the master branch and for the latest v3.0 branch.
However these files can be found elsewhere in the repo, for example, address_controller_spec.rb can be found in api/spec/requests/spree/api/. This change in files is to be expected since this is what the Rspec docs recommend (API specs are in the requests folder): https://relishapp.com/rspec/rspec-rails/docs/directory-structure

@Niamtokik asks how he should name the new files in the Issue, @Kennyadsl responds by saying that it should be done according to Rails convention. From what I am seeing in the example given in the docs for Request type spec: https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec. The name of the file does not contain _request_ in the middle, so a file named "addresses_controller_spec.rb" would become "addresses_spec.rb". Am I right in this assessment? Should I name the files according to some other convention?

The second part of @Kennyadsl Issue request says to "clean-up references inside the specs", and he gives this as an example: 
```
module Spree
  describe Api::AddressesController, type: :request do
```

This part has me a bit confused because the specs are still referring to the controllers, so the spec for AddressesController is still referring to that class. @Niamtokik asks him if he should rename the spec such as "AddressRequest". @Jarednorman tells him that this wouldn't work, because such classes do not exist, and the spec would not know where to look for them. Is this a misunderstanding from @Kennyadsl? Or should the Controllers themselves be named differently? In the case of the latter, should I open a new Issue and ask for someone to rename the controllers?

A minor point that I wish to include just for completeness sake, is that there are some files that were changed in #2052 which are no longer in the project, namely: transfer_items_controller_spec.rb, stock_transfers_controller_spec.rb, protect_product_actions.rb
There are also some files that were not originally in the files from #2052 but that I still changed the file name of since they were found in the requests folder, namely: coupon_codes_controller_spec.rb, and customer_returns_controller_spec.rb

Lastly, out of the 35 commits made by @Jhawthorn, most of them do not reference the controllers, except for the file name. Am I missing something? Are there more details that I should change/remove?


jarednorman:
When I say the constant doesn't exist, it is because in [this comment](https://github.com/solidusio/solidus/issues/4001#issuecomment-821866925), @niamtokik suggests changing:

```ruby
module Spree
  describe Api::AddressesController, type: :request do
  # ...
```

to this:

```ruby
module Spree
  describe Api::AddressesRequest, type: :request do
  # ...
```

`Spree::Api::AddressesController` is a constant that exists; `Spree::Api::AddressesRequest` does not.

This PR addresses the file naming issue, so I'm happy with it as an incremental improvement, but we've not yet addresses the naming convention issue within the files: these are request specs and so likely shouldn't explicitly reference the controller they are testing.

jarednorman (approved):

kennyadsl (changes requested):
Thanks! 🙏 

I think that `api/spec/controllers/spree/api/base_controller_spec.rb`, `api/spec/controllers/spree/api/resource_controller_spec.rb` and `api/spec/controllers/spree/api/soft_deletable_resource_controller_spec.rb` may not need a rename. Or is it a good practice to remove the `_controller` part from controller specs as well? Reading [this page](https://relishapp.com/rspec/rspec-rails/docs/controller-specs/controller-spec), looks like it's a good practice to keep it, isn't it?

I second what @jarednorman said: using a call that does not exist in the main `describe` block is not ideal, [see here for some example](https://relishapp.com/rspec/rspec-rails/v/5-0/docs/request-specs/request-spec).

kennyadsl:
Closing, thanks anyway!

PR #4123, opened 2021-06-28
Exclude canceled orders in the #usage_count of promotions and promotion codes

ikraamg:
**Description**

This change is extracted from a clients project. I thought it might be useful to have this on Solidus because excluding the count of canceled orders in `#usage-count` allows for a more accurate count(and limit) of promotions and promotion_codes.

**Checklist:**

### DIFF
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -192,7 +192,7 @@ module Spree
     def usage_count(excluded_orders: [])
       Spree::Adjustment.promotion.
         eligible.
-        in_completed_orders(excluded_orders: excluded_orders).
+        in_completed_orders(excluded_orders: excluded_orders, exclude_canceled: true).
         where(source_id: actions).
         count(:order_id)
     end
--- a/core/app/models/spree/adjustment.rb
+++ b/core/app/models/spree/adjustment.rb
@@ -55,11 +55,12 @@ module Spree
     #
     # @param excluded_orders [Array<Spree::Order>] Orders to exclude from query
     # @return [ActiveRecord::Relation] Scoped Adjustments
-    def self.in_completed_orders(excluded_orders: [])
-      joins(:order).
-      merge(Spree::Order.complete).
-      where.not(spree_orders: { id: excluded_orders }).
-      distinct
+    def self.in_completed_orders(excluded_orders: [], exclude_canceled: false)
+      result = joins(:order)
+               .merge(Spree::Order.complete)
+               .where.not(spree_orders: { id: excluded_orders })
+               .distinct
+      exclude_canceled ? result.merge(Spree::Order.not_canceled) : result
     end
 
     def finalize!
--- a/core/app/models/spree/promotion_code.rb
+++ b/core/app/models/spree/promotion_code.rb
@@ -30,7 +30,7 @@ class Spree::PromotionCode < Spree::Base
   def usage_count(excluded_orders: [])
     adjustments.
     eligible.
-    in_completed_orders(excluded_orders: excluded_orders).
+    in_completed_orders(excluded_orders: excluded_orders, exclude_canceled: true).
     count(:order_id)
   end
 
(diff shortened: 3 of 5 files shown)


kennyadsl (approved):
Nice catch, the CI will be green once we address #4110

core/app/models/spree/promotion.rb
```diff
@@ -193,6 +193,7 @@ def usage_count(excluded_orders: [])
       Spree::Adjustment.promotion.
         eligible.
         in_completed_orders(excluded_orders: excluded_orders).
+        where(spree_orders: { id: Spree::Order.not_canceled }).
```
spaghetticode, line 196: I think there may be performance issues associated with this subquery, as it yields all orders in the db except for the ones canceled. This could mean millions of records on large enough DBs.

spaghetticode (commented):
I think the feature change makes sense, but I see a couple of issues, the first one is related to large datasets (see the other comment), the second one is we're changing the original behavior, so I'm wondering if we should give the possibility to keep the old one (for example via a config flag) or if the change is so common sense that we should consider the original behavior almost a bug.

core/app/models/spree/promotion.rb
ikraamg: @spaghetticode, thanks for catching this 🙏 

I have tried an alternative approach that changes the `#in_completed_orders` method in `Adjustments`. 
It does not alter the method's default behaviour, but it allows for `canceled` orders to be excluded on passing an optional argument `exclude_canceled_orders`. 
Please see this [commit](https://github.com/solidusio/solidus/pull/4123/commits/1745930de91e1b14a80c0190ad2d06512bbb3789).

Although the old `#usage_count` behaviour can be considered a bug, we can also use a config flag with this change if you feel this is the better way to go.

It does look a bit messy to me, so please let me know if this is okay and if there is a better way to do something like conditional method changing in this case.

I see that I used the non-American spelling of `canceled`(`cancelled`) in the argument 🤦. I will fix that, squash the commits and put some more of this context into the commit message when you confirm this approach is fine.


The query that was running before you raised this issue was:
```sql
SELECT COUNT(DISTINCT "spree_adjustments"."order_id") FROM "spree_adjustments" INNER JOIN "spree_orders" ON "spree_orders"."id" = "spree_adjustments"."order_id" WHERE "spree_adjustments"."promotion_code_id" = ? AND "spree_adjustments"."eligible" = ? AND "spree_orders"."completed_at" IS NOT NULL AND 1=1 AND "spree_orders"."state" != ?  [["promotion_code_id", 1], ["eligible", 1], ["state", "canceled"]]
```

With the new approach, it runs this instead:
```sql
SELECT COUNT(DISTINCT "spree_adjustments"."order_id") FROM "spree_adjustments" INNER JOIN "spree_orders" ON "spree_orders"."id" = "spree_adjustments"."order_id" WHERE "spree_adjustments"."promotion_code_id" = ? AND "spree_adjustments"."eligible" = ? AND "spree_orders"."completed_at" IS NOT NULL AND 1=1 AND "spree_orders"."id" IN (SELECT "spree_orders"."id" FROM "spree_orders" WHERE "spree_orders"."state" != ?)  [["promotion_code_id", 1], ["eligible", 1], ["state", "canceled"]]
```

spaghetticode:
@ikraamg thanks for removing the orders subquery 🙏 

I'm not sure how picky we want to be with the behavior change, which is now moved [here](https://github.com/solidusio/solidus/pull/4123/files#diff-575b646b9f6dd87dd6d0f9800480ad9cfc8c51b59cbc81c88c273c2b74ec4928R195) and [here](https://github.com/solidusio/solidus/pull/4123/files#diff-a091604fe38b9746f35ddc14eb3ed8a3815ee3a3bd65e7802f3e6f4404a1334dR33). I will discuss this with the rest of the core team during the next call and let you know.

core/app/models/spree/promotion.rb
ikraamg: Great! Thanks for double checking. I think I pasted the same query String in both places 🤦

jarednorman (approved):

PR #4130, opened 2021-07-13
Introduce new EventBus adapter

waiting-for-dev:
**Description**

We're deprecating `Spree::Event::Adapters::ActiveSupportNotifications`
for a couple of reasons:

- It uses the same bus that standard Rails notifications. It forces us
  to try to normalize the event names given through `Spree::Event.fire`
  & `Spree::Event.subscribe` to add a `.spree` suffix. However, it's not
  a complete solution as we're missing subscriptions placed through a
  regular expression (like `Spree::Event.subscribe /.*/`, which
  subscribes to all events, including Rails ones).

- It relies heavily on global state. Besides not being able to create
  two different buses (see the previous point), it makes it cumbersome
  to test events in isolation.

This commit introduces a new `Spree::Event::Adapters::EventBus` adapter,
which instances make for independent event buses. As our implementation
of the event bus system relies on a global bus for all Solidus events,
the idea is to initialize a single instance and configure it as
`Spree::Config.events.adapter`. However, we've added a new optional
`adapter:` parameter to all relevant methods in `Spree::Event` to inject
other buses when testing. E.g.:

```ruby
bus = Spree::Event::Adapters::EventBus.new
Spree::Event.subscribe('foo', adapter: bus) { do_something }
Spree::Event.fire('foo', adapter: bus)
```

The old adapter is still the default one due to backward compatibility
reasons. However, we display a deprecation warning to let users know
that they're encouraged to change. There're two critical updates that
they need to be aware of (they're also described in the deprecated
message):

- Event name normalization won't happen anymore. E.g. in order to
  subscribe to all Solidus events probably users will need to change
  from `Spree::Event.subscribe /\.spree$/` to `Spree::Event.subscribe
  /.*/`. There's no way to reliably warn when the `.spree` prefix is
  used, as users can still decide to use it as far as they're consistent
  when performing different operations. On top of that, we still can't
  inspect regular expressions with confidence.

- The old adapter allowed to provide a block on `Spree::Event.fire`, like in:

  ```ruby
  Spree::Event.fire 'foo', order: order do
    order.do_something
  end
  ```

  As this block is entirely unnecessary and it can lead to
  confusion (is it executed before or after the subscribers?), it's no
  longer supported in the new adapter. To provide a safer upgrade
  experience, we raise an `ArgumentError` when the new adapter is being
  used, and a block is provided.

The new `EventBus` adapter will be made the default one on the next
major Solidus release.

The old adapter is leaking the abstraction through the returned values
(`ActiveSupport::Notifications::Fanout::Subscribers::Timed` instances).
In the new adapter, we introduce `Spree::Event::Event` &
`Spree::Event::Listener` objects that the adapters need to return.

The main `Spree::Event` module will still keep a copy of the
subscriptions when the legacy adapter is used. This behavior was
introduced to differentiate Solidus events from others, but it can cause
inconsistencies if we compare it with the copy on the adapter. As
something no longer needed, we're branching to don't perform the logic
when the adapter is the new one.

**Checklist:**

### DIFF
--- a/core/lib/spree/event.rb
+++ b/core/lib/spree/event.rb
@@ -174,6 +174,14 @@ module Spree
       Spree::Config.events.adapter
     end
 
+    def adapter
+      Spree::Deprecation.warn <<~MSG
+        `Spree::Event.adapter` is deprecated. Please, use
+        `Spree::Event.default_adapter` instead.
+      MSG
+      default_adapter
+    end
+
     # @!attribute [r] subscribers
     #   @return <Spree::Event::SubscriberRegistry> The registry for supporting class reloading for Spree::Event::Subscriber instances
     def subscriber_registry
--- a/core/lib/spree/event.rb
+++ b/core/lib/spree/event.rb
@@ -6,44 +6,88 @@ require_relative 'event/configuration'
 require_relative 'event/subscriber'
 
 module Spree
+  # Event bus for Solidus.
+  #
+  # This module serves as the interface to access the Event Bus system in
+  # Solidus. You can use different underlying adapters to provide the core
+  # logic. It's recommended that you use {Spree::Event::Adapters::Default}.
+  #
+  # You use the {#fire} method to trigger an event:
+  #
+  #  @example
+  #    Spree::Event.fire 'order_finalized', order: order
+  #
+  # Then, you can use {#subscribe} to hook into the event:
+  #
+  #   @example
+  #     Spree::Event.subscribe 'order_finalized' do |event|
+  #       # Take the order at event.payload[:order]
+  #     end
+  #
+  # You can also subscribe to an event through a module including
+  # {Spree::Event::Subscriber}:
+  #
+  #  @example
+  #    module MySubscriber
+  #      include Spree::Event::Subscriber
+  #
+  #      event_action :order_finalized
+  #
+  #      def order_finalized(event)
+  #        # Again, take the order at event.payload[:order]
+  #      end
+  #    end
   module Event
     extend self
 
     delegate :activate_autoloadable_subscribers, :activate_all_subscribers, :deactivate_all_subscribers, to: :subscriber_registry
 
-    # Allows to trigger events that can be subscribed using #subscribe. An
-    # optional block can be passed that will be executed immediately. The
-    # actual code implementation is delegated to the adapter.
+    # Allows to trigger events that can be subscribed using {#subscribe}.
     #
-    # @param [String] event_name the name of the event. The suffix ".spree"
-    #  will be added automatically if not present
-    # @param [Hash] opts a list of options to be passed to the triggered event
+    # The actual code implementation is delegated to the adapter.
+    #
+    # @param [String, Symbol] event_name the name of the event.
+    # @param [Hash] opts a list of options to be passed to the triggered event.
+    # They will be made available through the {Spree::Event::Event} instance
+    # that is yielded to the subscribers (see {Spree::Event::Event#payload}).
+    # However, take into account that the deprecated
+    # {Spree::Event::Adapters::ActiveSupportNotifications} adapter will yield a
+    # {ActiveSupport::Notifications::Fanout::Subscribers::Timed} instance
+    # instead.
+    # @option opts [Any] :adapter Reserved to indicate the adapter to use as
+    # event bus. Defaults to {#default_adapter}
+    # @return [Spree::Event::Event] an event object, unless the adapter is
+    # {Spree::Event::Adapters::ActiveSupportNotifications}
     #
     # @example Trigger an event named 'order_finalized'
     #   Spree::Event.fire 'order_finalized', order: @order do
     #     @order.finalize!
     #   end
-    def fire(event_name, opts = {})
-      adapter.fire normalize_name(event_name), opts do
-        yield opts if block_given?
-      end
+    #
+    # TODO: Change signature so that `opts` are keyword arguments, and include
+    # `adapter:` in them. We want to do that on Solidus 4.0. Spree::Deprecation
+    # can't be used because of this: https://github.com/solidusio/solidus/pull/4130#discussion_r668666924
+    def fire(event_name, opts = {}, &block)
+      adapter = opts.delete(:adapter) || default_adapter
+      handle_block_on_fire(block, opts, adapter) if block_given?
+      adapter.fire normalize_name(event_name), opts
     end
 
-    # Subscribe to an event with the given name. The provided block is executed
-    # every time the subscribed event is fired.
+    # Subscribe to events matching the given name.
     #
-    # @param [String, Regexp] event_name the name of the event.
-    #  When String, the suffix ".spree" will be added automatically if not present,
-    #  when using the default adapter for ActiveSupportNotifications.
-    #  When Regexp, due to the unpredictability of all possible regexp combinations,
-    #  adding the suffix is developer's responsibility (if you don't, you will
-    #  subscribe to all notifications, including internal Rails notifications
-    #  as well).
+    # The provided block is executed every time the subscribed event is fired.
     #
-    # @see Spree::Event::Adapters::ActiveSupportNotifications#normalize_name
+    # @param [String, Symbol, Regexp] event_name the name of the event. When it's a
+    # {Regexp} it subscribes to all that match.
+    # @param [Any] adapter the event bus adapter to use.
+    # @yield block to execute when an event is triggered
     #
-    # @return a subscription object that can be used as reference in order
-    #  to remove the subscription
+    # @return [Spree::Event::Listener] a subscription object that can be used as
+    # reference in order to remove the subscription. However, take into account
+    # that the deprecated {Spree::Event::Adapters::ActiveSupportNotifications}
+    # adapter will return a
+    # {ActiveSupport::Notifications::Fanout::Subscribers::Timed} instance
+    # instead.
     #
     # @example Subscribe to the `order_finalized` event
     #   Spree::Event.subscribe 'order_finalized' do |event|
@@ -52,51 +96,81 @@ module Spree
     #   end
     #
     # @see Spree::Event#unsubscribe
-    def subscribe(event_name, &block)
-      name = normalize_name(event_name)
-      listener_names << name
-      adapter.subscribe(name, &block)
+    def subscribe(event_name, adapter: default_adapter, &block)
+      event_name = normalize_name(event_name)
+      adapter.subscribe(event_name, &block).tap do
+        if legacy_adapter?(adapter)
+          listener_names << adapter.normalize_name(event_name)
+        end
+      end
     end
 
     # Unsubscribes a whole event or a specific subscription object
     #
-    # @param [String, Object] subscriber the event name as a string (with
-    #  or without the ".spree" suffix) or the subscription object
+    # When unsubscribing from an event, all previous listeners are deactivated.
+    # Still, you can add new subscriptions to the same event and they'll be
+    # called if the event is fired:
+    #
+    # @example
+    #   Spree::Event.subscribe('foo') { do_something }
+    #   Spree::Event.unsubscribe 'foo'
+    #   Spree::Event.subscribe('foo') { do_something_else }
+    #   Spree::Event.fire 'foo' # `do_something_else` will be called, but
+    #   # `do_something` won't
+    #
+    # @param [String, Symbol, Spree::Event::Listener] subscriber_or_event_name the
+    # event name as a string or the subscription object. Take into account that
+    # if the deprecated {Spree::Event::Adapters::ActiveSupportNotifications}
+    # adapter is used, the subscription object will be a
+    # {ActiveSupport::Notifications::Fanout::Subscribers::Timed} object.
     #
     # @example Unsubscribe a single subscription
-    #   subscription = Spree::Event.fire 'order_finalized'
+    #   subscription = Spree::Event.subscribe('order_finalized') { do_something
+    #   }
     #   Spree::Event.unsubscribe(subscription)
     # @example Unsubscribe all `order_finalized` event subscriptions
     #   Spree::Event.unsubscribe('order_finalized')
-    # @example Unsubscribe an event by name with explicit prefix
-    #   Spree::Event.unsubscribe('order_finalized.spree')
-    def unsubscribe(subscriber)
-      name_or_subscriber = subscriber.is_a?(String) ? normalize_name(subscriber) : subscriber
-      adapter.unsubscribe(name_or_subscriber)
+    def unsubscribe(subscriber_or_event_name, adapter: default_adapter)
+      if subscriber_or_event_name.is_a?(Listener) || subscriber_or_event_name.is_a?(ActiveSupport::Notifications::Fanout::Subscribers::Timed)
+        unsubscribe_listener(subscriber_or_event_name, adapter)
+      else
+        unsubscribe_event(subscriber_or_event_name, adapter)
+      end
     end
 
-    # Lists all subscriptions currently registered under the ".spree"
-    # namespace. Actual implementation is delegated to the adapter
+    # Lists all subscriptions.
     #
-    # @return [Hash] an hash with event names as keys and arrays of subscriptions
-    #  as values
+    # Actual implementation is delegated to the adapter.
+    #
+    # @return [Hash<<String,Regexp>,Spree::Event::Listener>] an hash with
+    # patterns as keys and arrays of subscriptions as values. Take into account
+    # that the deprecated {Spree::Event::Adapters::ActiveSupportNotifications}
+    # adapter will map to
+    # {ActiveSupport::Notifications::Fanout::Subscribers::Timed} instances
+    # instead.
     #
     # @example Current subscriptions
     #  Spree::Event.listeners
-    #    # => {"order_finalized.spree"=> [#<ActiveSupport...>],
-    #      "reimbursement_reimbursed.spree"=> [#<ActiveSupport...>]}
-    def listeners
-      adapter.listeners_for(listener_names)
+    #    # => {"order_finalized"=> [#<Spree::Event::Listener...>],
+    #         "reimbursement_reimbursed"=> [#<Spree::Event::Listener...>]}
+    def listeners(adapter: default_adapter)
+      if legacy_adapter?(adapter)
+        adapter.listeners_for(listener_names)
+      else
+        init = Hash.new { |h, k| h[k] = [] }
+        adapter.listeners.each_with_object(init) do |listener, map|
+          map[listener.pattern] << listener
+        end
+      end
     end
 
-    # The adapter used by Spree::Event, defaults to
-    # Spree::Event::Adapters::ActiveSupportNotifications
+    # The default adapter used by Spree::Event.
     #
     # @example Change the adapter
-    #   Spree::Config.events.adapter = "Spree::EventBus.new"
+    #   Spree::Config.events.adapter = "Spree::OtherAdapter.new"
     #
-    # @see Spree::AppConfiguration
-    def adapter
+    # @see Spree::Event::Configuration
+    def default_adapter
       Spree::Config.events.adapter
     end
 
@@ -108,12 +182,65 @@ module Spree
 
     private
 
-    def normalize_name(name)
-     adapter.normalize_name(name)
+    def unsubscribe_listener(listener, adapter)
+      adapter.unsubscribe(listener)
+    end
+
+    def unsubscribe_event(event_name, adapter)
+      adapter.unsubscribe(normalize_name(event_name))
     end
 
     def listener_names
       @listeners_names ||= Set.new
+    end
+
+    def normalize_name(name)
+      case name
+      when Symbol
+        name.to_s
+      else
+        name
+      end
+    end
+
+    def handle_block_on_fire(block, opts, adapter)
+      example = <<~MSG
+        Please, instead of:
+
+          Spree::Event.fire 'event_name', order: order do
+            order.do_something
+          end
+
+        Use:
+
+          order.do_something
+          Spree::Event.fire 'event_name', order: order
+      MSG
+      if legacy_adapter?(adapter)
+        Spree::Deprecation.warn <<~MSG
+          Blocks on `Spree::Event.fire` are ignored in the new adapter
+          `Spree::Event::Adapters::Default`, and your current adapter
+          (`Spree::Event::Adapters::ActiveSupportNotifications`) is deprecated.
+          For an easier transition it's recommendable to update your code.
+
+          #{example}
+
+        MSG
+        block.call(opts)
+      else
+        raise ArgumentError, <<~MSG
+          Blocks passed to `Spree::Event.fire` are ignored unless the adapter is
+          `Spree::Event::Adapters::ActiveSupportNotifications` (which is
+          deprecated).
+
+          #{example}
+
+        MSG
+      end
+    end
+
+    def legacy_adapter?(adapter)
+      adapter == Adapters::ActiveSupportNotifications
     end
   end
 end
--- a/core/lib/spree/event/configuration.rb
+++ b/core/lib/spree/event/configuration.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+require 'spree/core/versioned_value'
+require 'spree/event/adapters/active_support_notifications'
+require 'spree/event/adapters/default'
+
 module Spree
   module Event
     class Configuration
@@ -14,11 +18,63 @@ module Spree
       end
 
       def adapter
-        @adapter ||= Spree::Event::Adapters::ActiveSupportNotifications
+        @adapter ||= Spree::Core::VersionedValue.new(
+          Spree::Event::Adapters::ActiveSupportNotifications,
+          "4.0.0.alpha" => Spree::Event::Adapters::Default.new
+        ).call.tap do |value|
+          deprecate_if_legacy_adapter(value)
+        end
       end
 
+      # Only used by {Spree::Event::Adapters::ActiveSupportNotifications}.
       def suffix
         @suffix ||= '.spree'
+      end
+
+      private
+
+      def deprecate_if_legacy_adapter(adapter)
+        Spree::Deprecation.warn <<~MSG if adapter == Spree::Event::Adapters::ActiveSupportNotifications
+          `Spree::Event::Adapters::ActiveSupportNotifications` adapter is
+          deprecated. Please, take your time to update it to an instance of
+          `Spree::Event::Adapters::Default`. I.e., in your `spree.rb`:
+
+          require 'spree/event/adapters/default'
+
+          Spree.config do |config|
+            # ...
+            config.events.adapter = Spree::Event::Adapters.Default.new
+            # ...
+          end
+
+          That will be the new default on Solidus 4.
+
+          Take into account there're two critical changes in behavior in the new adapter:
+
+          - Event names are no longer automatically suffixed with `.spree`, as
+          they're no longer in the same bucket that Rails's ones. So, for
+          instance, if you were relying on a global subscription to all Solidus
+          events with:
+
+            Spree::Event.subscribe /.*\.spree$/
+
+          You should change it to:
+
+            Spree::Event.subscribe /.*/
+
+          - Providing a block to `Spree::Event.fire` is no longer supported. If
+          you're doing something like:
+
+            Spree::Event.fire 'event_name', order: order do
+              order.do_something
+            end
+
+            You now need to change it to:
+
+            order.do_something
+            Spree::Event.fire 'event_name', order: order
+
+        MSG
       end
     end
   end
--- a/core/lib/spree/event/listener.rb
+++ b/core/lib/spree/event/listener.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+module Spree
+  module Event
+    # Subscription to an event
+    #
+    # An instance of it is returned on {Spree::Event.subscribe}.
+    #
+    # You're not expected to perform any action with it, besides using it as
+    # a reference to unsubscribe.
+    #
+    # @example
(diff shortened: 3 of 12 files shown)


core/lib/spree/event/listener.rb
```diff
@@ -0,0 +1,51 @@
...
+      end
+
+      # @api private
+      def call(event)
+        @block.call(event)
+      end
+
+      # @api private
+      def matches?(event_name)
+        pattern === event_name &&
```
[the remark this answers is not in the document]
waiting-for-dev: I think that in this case is clearer and better to use polymorphic polymorphism instead of an ad-hoc `#is_a?(Regexp)`... WDYT?

core/lib/spree/event.rb
```diff
@@ -6,44 +6,102 @@
...
+    # instead.
+    # @option opts [Any] :adapter Reserved to indicate the adapter to use as
+    # event bus. Defaults to {#default_adapter}
     #
     # @example Trigger an event named 'order_finalized'
     #   Spree::Event.fire 'order_finalized', order: @order do
     #     @order.finalize!
     #   end
+    #
+    # TODO: Change opts to be keyword arguments and include `adapter:` in them
```
kennyadsl, line 65: Is this `TODO` for this PR?

core/lib/spree/event.rb
```diff
@@ -6,44 +6,102 @@
...
+            end
+
+          Use:
+
+            order.do_something
+            Spree::Event.fire 'event_name', order: order
+
+        MSG
+      elsif block_given?
+        yield opts
```
kennyadsl, line 85: I think we need to emit a deprecation warning here, otherwise people using `AS::N` adapter won't see that they need to remove the block in their fire calls. They will only know this with an exception after switching to the new adapter when `fire` will be called.

core/lib/spree/event.rb
```diff
@@ -52,51 +110,78 @@ def fire(event_name, opts = {})
...
+    #   }
     #   Spree::Event.unsubscribe(subscription)
     # @example Unsubscribe all `order_finalized` event subscriptions
     #   Spree::Event.unsubscribe('order_finalized')
-    # @example Unsubscribe an event by name with explicit prefix
-    #   Spree::Event.unsubscribe('order_finalized.spree')
-    def unsubscribe(subscriber)
-      name_or_subscriber = subscriber.is_a?(String) ? normalize_name(subscriber) : subscriber
+    def unsubscribe(name_or_subscriber, adapter: default_adapter)
+      name_or_subscriber = normalize_name(name_or_subscriber)
```
kennyadsl, line 148: I don't get how `normalize_name` is handling the case when `name_or_subscriber` is a `subscriber`. Will that just be a symbol in that case? In the method documentation I see also `Spree::Event::Listener`

core/lib/spree/event.rb
```diff
@@ -52,51 +110,78 @@ def fire(event_name, opts = {})
...
-    def listeners
-      adapter.listeners_for(listener_names)
+    #    # => {"order_finalized"=> [#<Spree::Event::Listener...>],
+    #         "reimbursement_reimbursed"=> [#<Spree::Event::Listener...>]}
+    def listeners(adapter: default_adapter)
+      if legacy_adapter(adapter)
+        adapter.listeners_for(listener_names)
+      else
+        init = Hash.new { |h, k| h[k] = [] }
+        adapter.listeners.each_with_object(init) do |listener, map|
```
kennyadsl, line 172: I have the feeling Ruby is providing some handy shortcuts for this. For example, I think we can avoid creating the init variable and just use `adapter.listeners.each_with_object(Hash.new([]))`, isn't it?

core/lib/spree/event.rb
```diff
@@ -108,12 +193,21 @@ def subscriber_registry
...
+      when Symbol
+        name.to_s
+      else
+        name
+      end
     end
 
-    def listener_names
-      @listeners_names ||= Set.new
+    def legacy_adapter(adapter)
```
kennyadsl, line 209: Shouldn't this be `legacy_adapter?` Also, did you consider moving this in the adapter itself? I know they have no Base class to inherits from (that would make it easy to set the default for this method to false) but maybe it's a signal we need it?

core/lib/spree/event/configuration.rb
```diff
@@ -14,12 +18,64 @@ def autoload_subscribers
       end
 
       def adapter
-        @adapter ||= Spree::Event::Adapters::ActiveSupportNotifications
+        @adapter ||= Spree::Core::VersionedValue.new(
+          Spree::Event::Adapters::ActiveSupportNotifications,
+          "4.0.0.alpha" => Spree::Event::Adapters::EventBus.new
```
kennyadsl, line 23: Why one is a class and the other is an instance?

core/lib/spree/event/configuration.rb
```diff
@@ -14,12 +18,64 @@ def autoload_subscribers
...
 
+      # Only used by {Spree::Event::Adapters::ActiveSupportNotifications}.
       def suffix
         @suffix ||= '.spree'
       end
+
+      private
+
+      def deprecate_if_legacy_adapter(adapter)
+        Spree::Deprecation.warn <<~MSG if adapter == Spree::Event::Adapters::ActiveSupportNotifications && !ENV['CI']
```
kennyadsl, line 37: Why we need `!ENV['CI']`?

kennyadsl (commented):
Made a first pass and left some questions, great work as usual.

core/lib/spree/event.rb
waiting-for-dev: I'd say no, but let me explain it better.

So, the issue is that I'd like to be consistent with the other methods, as I do think that having `adapter:` as a keyword argument is the most expressive way to inject adapters. However, if I add it to the signature:

```ruby
def fire(event_name, opts = {}, adapter: default_adapter)
```

There's no way for it to work on Ruby < 3 unless we provide the `adapter:` argument when invoking it (which breaks the whole purpose). I.e., both

```ruby
fire('foo', order: order)
```

and

```ruby
fire('foo', { order: order })
```

will raise `ArgumentError (unknown keyword: :order)`.

Given that, we should move `opts` from being a `Hash` to keyword arguments:

```ruby
def fire(event_name, adapter: default_adapter, **kwargs)
```

It will work in RUBY < 3, both with:

```ruby
fire('foo', order: order)
```

and

```ruby
fire('foo', { order: order })
```

Still, Ruby 3 stores are now expecting that the event options are given as a `Hash`, so it will break them. In addition, we can't provide a straightforward deprecation path as if we try to support both options:

```ruby
def fire(event_name, opts = {}, adapter: default_adapter, **kwargs)
```

we're back to square one.

I can think of two workarounds for it:

1. Define different methods depending on `RUBY_VERSION`. E.g. (pseudo-code):

```ruby
if RUBY_VERSION < 3
  def fire(event_name, adapter: default_adapter, **kwargs)
    # ...
else
  def fire(event_name, opts = {}, adapter: default_adapter, **kwargs)
    opts = if opts.any?
                deprecate('Use kwargs instead of hash')
                opts
              else
                kwargs
              end
```

I think this option would deserve a separate commit.

2. Just change the signature to:

```ruby
def fire(event_name, adapter: default_adapter, **kwargs)
```

on Solidus 4 and inform about the change on the upgrade guide.

kennyadsl: Got it. What if we use opts for now and deprecate all the un-recognized options passed? So that when we'll switch to kwargs in Solidus 4 we'll sure to be compliant.



```ruby
# Pseudo-code

def fire(event_name, opts = {}, 
  adapter = opts.delete :adapter
  order = opts.delete :order
  
  opts.each do |key, value|
     Spree::Deprecation.warn("passing #{key} to fire is deprecated, please just pass order and adapter")
  end
end
```

core/lib/spree/event.rb
waiting-for-dev: The main deprecation message, the one they should see when booting up the app or running the test suite, is already informing users to stop providing blocks to `Spree::Event.fire`. Do you think it's not enough? In fact, the block is still supported if they're using the legacy adapter.

kennyadsl: I saw that the main deprecation warning mentions that people need to stop using fire with block, but that code could technically be in some extension and they won't be able to know it until the code breaks. Honestly, I don't think it's super used in the real world but I'd still add a deprecation warning here, just to be 100% sure.

core/lib/spree/event.rb
waiting-for-dev: `normalize_name` is `Spree::Event` is just returning the argument untouched unless it's a Symbol. So, if it's a `Spree::Event::Listener` it won't do anything. I agree it'd be clearer to ask whether it's a `Listener` before dispatching it to `normalize_name`, but as we can be dealing with an `ActiveSupport...Fanout` object from the legacy adapter it'd make the code rather ugly. WDYT?

core/lib/spree/event.rb
waiting-for-dev: Using `Hash.new` with a default value binds the same instance to every key. So if you use a mutable object like an array, you can get into serious troubles :sweat:. E.g.:

```ruby
hash = Hash.new([])
hash[:a] << 1
hash[:b] # => [1]
```

:boom:

core/lib/spree/event.rb
waiting-for-dev: Sure! I just renamed it `legacy_adapter?`.

As for defining a base class for the adapters, I'd respectfully disagree if the reason is to accommodate the `legacy_adapter?` method. This method is here as a hack until we finish the process and get rid of the ASN adapter, so it's okay to look like a hack. TBH I wouldn't like it to leak into other files. As the idea is to eventually leave us only with the `EventBus` adapter, I wouldn't complicate things at this stage by creating a superclass that would act only as an interface definition.

core/lib/spree/event/configuration.rb
waiting-for-dev: One of the main ideas behind building a new adapter is to make it work without relying on global state. Having it as an instance means that we can use its state as storage for the event bus. A clear benefit is that it allows us to use separate buses when testing.

On top of that, even though we're still using a global bus for the whole application to preserve the current API, we could eventually create different buses. For instance:

```ruby
module Spree
  class Order
    include Spree::Event::Adapters::EventBus # we'd want to make it work through `Spree::Event` though

    def method(...)
      fire 'order_finalized', order: self
    end
  end
end
```

core/lib/spree/event/configuration.rb
waiting-for-dev: The current version of Solidus is still using the legacy adapter, even though it's warning about its deprecation. CircleCI is currently configured to raise when any warning is displayed:

https://github.com/solidusio/solidus/blob/7118543095e0db0887f3b851e1107a1afe5b6aba/.circleci/config.yml#L11

The `CI` environment variable is set on CircleCI, making it possible to have a green build.

However, the warnings are indeed displayed when running `bin/rspec` locally. We could create a custom `ENV` variable in the `spec_helper` files and check it instead of `CI`, but I'm not a fan of this solution as it's too easy for that variable definitions to say there once they're no longer needed. Remember that we can't just ignore the warning when `RAILS_ENV` is `test` as we do want it to be displayed on users' application test suite. Do you think we have a better alternative?

core/lib/spree/event.rb
kennyadsl: I respectfully accept your disagreement here. 👍 🙂

core/lib/spree/event/configuration.rb
kennyadsl: Got it, so basically the instance of `Spree::Event::Adapters::EventBus.new` has the same interface of the `Spree::Event::Adapters::ActiveSupportNotifications` class/module.

core/lib/spree/event/configuration.rb
kennyadsl: > Do you think we have a better alternative?

I think we could set up the dummy app to test against the new adapter. This means that the deprecated code won't be tested by default, which is not ideal because people may still use it. We could build a CI matrix (or at least one job) that tests Solidus with the legacy adapter, or we can just create some smoke tests for it, changing the adapter runtime before running those specs. Does it make sense?

core/lib/spree/event.rb
kennyadsl: Ops, you are right. I didn't think about that.

core/lib/spree/event.rb
waiting-for-dev: The problem is that when you fire an event, you can attach to it whatever information you want to make available to the subscribers. Here I used `order`, but it could be anything else :slightly_smiling_face:

core/lib/spree/event.rb
waiting-for-dev: I see. Added!

core/lib/spree/event/configuration.rb
waiting-for-dev: Right :slightly_smiling_face:

core/lib/spree/event/configuration.rb
waiting-for-dev: Yeah, I think that's the best option we have. I updated the code so that we're using the new adapter on the dummy application. As for the smoke tests, we can use the same that we had before, those on `event_spec.rb` that reach the legacy adapter.

core/lib/spree/event.rb
kennyadsl: Oh right, how did I miss it? I see no way to deprecate it before removal then, maybe just changing the method name and deprecate `fire` at all but I don't think it's worth it. We can specify in the TODO that we can remove it in the next major, adding the keyword Deprecation somehow so that we find it along with the others when removing 3.x deprecations. 

Alternatively we could create a new issue for this, and move it into a new 4.0 milestone in GitHub.

core/lib/spree/event.rb
kennyadsl: My only concern is that looking at the `normalize_name` method as is, I assume that name can be a String or a Symbol, even if as you pointed out, it can be any object. Not sure if any action is needed honestly, the code looks good, maybe it's just me. :)

core/lib/spree/event/listener.rb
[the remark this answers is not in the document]
kennyadsl: I don't have strong opinions here, but if I got this correctly we can just use this instead:

```
pattern.match?(event_name)
```

core/lib/spree/event.rb
waiting-for-dev: I added the improved `TODO:` comment with `Spree::Deprecation` on it a link to this comment :slightly_smiling_face:

core/lib/spree/event.rb
waiting-for-dev: I think you're right. We can't let the legacy adapter condition a better design. I refactored the method so that `normalize_name` is only called when it's not a subscriber object from any of the adapters.

core/lib/spree/event.rb
```diff
@@ -108,12 +180,65 @@ def subscriber_registry
...
+
+          order.do_something
+          Spree::Event.fire 'event_name', order: order
+      MSG
+      if legacy_adapter?(adapter)
+        Spree::Deprecation.warn <<~MSG
+          Blocks on `Spree::Event.fire` are ignored in the new adapter
+          `Spree::Event::Adapters::EventBus`, and your current adapter
+          (`Spree::Event::Adapters::ActiveSupportNotifications`) is deprecated.
+          For an easier transition is recommendable to update your code.
```
aldesantis, line 222: is recommendable → it's recommendable

core/lib/spree/event/adapters/event_bus.rb
```diff
@@ -0,0 +1,79 @@
...
+      #   Spree.config do |config|
+      #     # ...
+      #     config.events.adapter = EventBus.new
+      #     # ...
+      #   end
+      #
+      # You won't need to do that from Solidus version 4.0 as this adapter will
+      # be the default one.
+      class EventBus
+        # @api private
```
aldesantis, line 27: You can apply `@api private` to the class, and YARD will mark all children of that class as private. See https://www.rubydoc.info/gems/yard/file/docs/Tags.md#api:
> Note: This tag is transitive. If it is applied on a namespace (module or class), it will immediately be applied to all children objects of that namespace unless it is redefined on the child object.

core/lib/spree/event/configuration.rb
```diff
@@ -14,12 +18,64 @@ def autoload_subscribers
...
+          Spree::Event::Adapters::ActiveSupportNotifications,
+          "4.0.0.alpha" => Spree::Event::Adapters::EventBus.new
+        ).call.tap do |value|
+          deprecate_if_legacy_adapter(value)
+        end
       end
 
+      # Only used by {Spree::Event::Adapters::ActiveSupportNotifications}.
       def suffix
         @suffix ||= '.spree'
```
aldesantis, line 31: Do you think we should display an error/refuse booting if the suffix is set even though the new adapter is being used?

core/lib/spree/event/listener.rb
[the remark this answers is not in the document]
aldesantis: I agree that this can be ignored. I like the idea of the event bus being able to use whatever responds to `#===` as a pattern.

core/lib/spree/event/adapters/event_bus.rb
```diff
@@ -0,0 +1,79 @@
...
+      #
+      #   Spree.config do |config|
+      #     # ...
+      #     config.events.adapter = EventBus.new
+      #     # ...
+      #   end
+      #
+      # You won't need to do that from Solidus version 4.0 as this adapter will
+      # be the default one.
+      class EventBus
```
aldesantis, line 26: Should this be simply called `Default`? The actual event bus is `Spree::Event`, this is just an adapter.

aldesantis (changes requested):
@waiting-for-dev thanks, this is amazing! 😍  I have a few change requests, but it's mostly minor stuff. I also have a couple of comments about the general architecture, but I don't know if these can be realistically implemented in a backwards-compatible way.

I'm not fully convinced about `Spree::Event` being a module, and users having to pass the adapter to each method call if they want to use a non-default adapter. It seems like it would be much cleaner to have a class that you can initialize with the adapter.

When you think about it, `Spree::Event` is the actual event bus, because it's responsible for coordinating everything. So we could have the following API:

```ruby
event_bus = Spree::Events::EventBus.new(adapter: custom_adapter)
event_bus.fire ...
```

Then, in their Solidus initializer, users could just do the following:

```ruby
Spree.config do |config|
  config.event_bus = Spree::Events::EventBus.new(adapter: Spree::Event::Adapters::Default)
end
```

As I said, I realize this is a huge departure from the current API, so if you think it's too complicated/doesn't make sense, feel free to let me know.

core/lib/spree/event/listener.rb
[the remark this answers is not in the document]
waiting-for-dev: The problem with `String.match?(pattern)` is that it coerces `pattern` to regular expression, so:

```ruby
'foo'.match?('o') # => true
```

So, it'd treat `event_name` as a regular expression, and, for instance, we could have a subscriber listening only to `order_canceled_because_of_lack_of_funds` fire when a more generic `order_canceled` is fired.

core/lib/spree/event.rb
waiting-for-dev: Thanks, fixed!

core/lib/spree/event/listener.rb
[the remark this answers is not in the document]
kennyadsl: Makes sense 👍

core/lib/spree/event/adapters/event_bus.rb
waiting-for-dev: Hmm, yeah, but the class is not private. It belongs to the public API, as you can do:

```ruby
Spree.config do |config|
  config.events.adapter = Spree::Events::Adapters::EventBus.new
ènd
```

waiting-for-dev: In fact, I just removed the `@api private` tag from the `initialize` method, as you need to call it to the bus. Please, let me know what you think :slightly_smiling_face:

core/lib/spree/event/adapters/event_bus.rb
waiting-for-dev: Ok, that makes sense. The adapter itself is an `EventBus`, as it could be used standalone, but probably it's easier to understand doing it as you say.

core/lib/spree/event/configuration.rb
waiting-for-dev: I think it'd be too aggressive as the value is set by default. We could render a warning, but still, I'm not convinced. We'd be forcing users to add `config.events.suffix = nil` to their initializer because of a deprecated behavior. On top of that, we'd bother them again to remove it once the legacy adapter is removed.

waiting-for-dev:
> ```ruby
> Spree.config do |config|
>   config.event_bus = Spree::Events::EventBus.new(adapter: Spree::Event::Adapters::Default)
> end
> ```

That would require assigning to `Spree::Event`, like in `Spree::Event = config.event_bus`. But, as we feared, the problem is keeping backward compatibility. If `Spree::Event` is an instance, then we can't define constants under its "namespace". However, `Spree::Event::Subscriber` is part of the public API. I.e.:

```ruby
A = Hash.new
class/module A
  class B
  end
end
# => TypeError (A is not a class/module)
```

waiting-for-dev:
I added another commit making `Spree::Event` to return the `Spree::Event::Event` object instead of the list of listeners.

aldesantis (approved):
@waiting-for-dev thanks, looks great! <3

core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -133,6 +133,9 @@ class Application < ::Rails::Application
 Spree.user_class = 'Spree::LegacyUser'
 Spree.config do |config|
   config.mails_from = "store@example.com"
+  # TODO: Remove on Solidus 4.0 as it'll be the default
+  require 'spree/event/adapters/default'
+  config.events.adapter = Spree::Event::Adapters::Default.new
```
spaghetticode, line 138: I'm not 100% convinced by this. If the old adapter is still the default one used by Solidus, I think we should run the suite against it. I read in a previous discussion that the issue is mostly deprecation warnings on CI: I think there could be a couple of solutions to this issue: 
1) don't deprecate the old behavior right now, but leave that to a subsequent PR (that would be a short lived victory, but it would also move some code out of this PR, making it smaller);
2) I think we could silence the deprecations with something like `allow(Spree::Deprecation).to receive(:warn)` so they don't make CI fail

spaghetticode (commented):
@waiting-for-dev definitely amazing work here! 🤩  

I'm sorry it took so long to review the PR, I started a couple of times before but the changes were harder to digest than I anticipated.

I found reviewing the PR a little more difficult than usual, partly because of the PR subject, then I think because it's basically one single commit. 
I tend to prefer when big PRs are spread among multiple commits, as they help in scoping code changes and their explanations. For example, I spent some amount of time trying to understand if the new implementation of `Spree::Event.normalize_name` could possibly lead to bugs (it was previously delegated to the adapter and the new definition is surely different), so I needed to check in the whole PR what areas were affected by this change, eventually realizing that the AS Notification adapter has been properly updated as well in order to still use the old definition. I think that, if this change was scoped in a single commit, this would have been easier to verify.

core/lib/spree/event.rb
```diff
@@ -52,51 +96,81 @@ def fire(event_name, opts = {})
...
-    # The adapter used by Spree::Event, defaults to
-    # Spree::Event::Adapters::ActiveSupportNotifications
+    # The default adapter used by Spree::Event.
     #
     # @example Change the adapter
-    #   Spree::Config.events.adapter = "Spree::EventBus.new"
+    #   Spree::Config.events.adapter = "Spree::OtherAdapter.new"
     #
-    # @see Spree::AppConfiguration
-    def adapter
```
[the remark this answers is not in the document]
waiting-for-dev: Hmmm, do we consider it a public method? Indeed, it's not marked as `@private`, so it's effectively a public method. But in which situation would a user be calling `Spree::Event.adapter` when we had a single adapter up to this point? I'm not against it, but I'm just wondering if it's necessary. WDYT? If you think it's better to add it, I will.

core/lib/spree/testing_support/dummy_app.rb
waiting-for-dev: Yeah, I'm neither 100% convinced. Ideally, we'd run two jobs, each one with a different adapter, but as we're limited here, we have to decide. I'd go with 2 or with the previous solution (adding a `CI` environment variable check) if we prefer to check the legacy adapter. However, the former would clutter our test suite. Thoughts @kennyadsl ?

waiting-for-dev:
> @waiting-for-dev definitely amazing work here! 
> 
> I'm sorry it took so long to review the PR, I started a couple of times before, but the changes were harder to digest than I anticipated.
> 
> I found reviewing the PR a little more difficult than usual, partly because of the PR subject, then I think because it's basically one single commit.
> I tend to prefer when big PRs are spread among multiple commits, as they help in scoping code changes and their explanations. For example, I spent some amount of time trying to understand if the new implementation of `Spree::Event.normalize_name` could possibly lead to bugs (it was previously delegated to the adapter and the new definition is surely different), so I needed to check in the whole PR what areas were affected by this change, eventually realizing that the AS Notification adapter has been properly updated as well in order to still use the old definition. I think that, if this change was scoped in a single commit, this would have been easier to verify.

Thanks for your review, @spaghetticode! I addressed your comments. Sorry for the difficulties in reviewing it. The truth is that the path to take wasn't clear at the beginning of this work, so the first phase was a good amount of exploratory coding. Fortunately, the rest of the work on top of this commit is well structured in different PRs.

kennyadsl:
Let's hold this for a while. We'll release Solidus 3.1 first so that all the new events-related changes coming will go in a single release.

core/lib/spree/event.rb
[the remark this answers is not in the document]
spaghetticode: Given this is a public API, and that this is a prominent Solidus feature, I would deprecate the method first, unless this creates any issues in our path forward. It doesn't require much work, and it keeps us on the safe side.

core/lib/spree/testing_support/dummy_app.rb
spaghetticode: Another solution, but don't know how feasible would be, is to have the event bus specs as shared examples and run them with both adapters. I think this could be handy also in the future, if we ever decide to change the default adapter again, or if we keep multiple official ones in the codebase.

core/lib/spree/event.rb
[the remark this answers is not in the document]
waiting-for-dev: Makes sense. Done!

core/lib/spree/testing_support/dummy_app.rb
waiting-for-dev: I think that the main problem is not the specs on the event bus itself but those at the integration level. We already have tests for both adapters, and I don't think it's worth trying to extract them to a shared example. The reason is that both adapters don't support the exact number of features. This is more evident in the work that will be built on top of this PR. They're features that are difficult or impossible to build with ASN. On top of that, I already talked with @aldesantis about this concept of adapters for the event bus. Again, the work that will follow this PR adds more complexity and therefore becomes more opinionated, probably making it more difficult or impossible to build good abstractions for other libraries. In my view, we should keep the adapter pattern only until the old adapter is completely deprecated and removed, and from that point on, I'd remove the adapters concept altogether.

Ideally, we should use two CI jobs, one for each adapter. If that's not possible, we should decide in favor of one or the other. I have just been playing with options to provide a rspec helper to run tests for both adapters, but it's tricky, and there is no way to ensure they are used if we add new tests. If we use them at the example level, the adapter will be ignored on any `before`, `let`... blocks, where a `Spree::Event.subscribe` could happen. And using them at the RSpec's class level (wrapping `context`, `before`... blocks) doesn't work as expected, as the examples are declared at the class level (when the RSpec file is loaded) but run at the instance level (once the adapter has been changed again to be the default one).

spaghetticode: Oh, I didn't know we are planning to remove the concept of the adapter from `Spree::Event`... then it definitely doesn't make any sense to build a multi-adapter testing facility. 

I guess that from a pragmatic point of view it would make sense to test the new adapter since it will become the new default and we don't plan to change anything on the old one, but on the other hand, since the old one is still the default, we need to make sure it keeps working properly as we introduce new features in order to avoid breaking Solidus installations... sounds like a tough call.

waiting-for-dev: > Oh, I didn't know we are planning to remove the concept of the adapter from Spree::Event

Well, it's not something 100% decided. It's more that, with the new additions to the event bus, I think that the adapters concept is ceasing to be meaningful, and I talked about it with @aldesantis. But I don't think the decision has been made, so other views are more than welcome :slightly_smiling_face:

jarednorman (approved):

core/lib/spree/testing_support/dummy_app.rb
kennyadsl: I'm also not 100% convinced here. Considering that we are doing to merge other changes in the Events context, I think it's very likely that we could have some effect on the legacy adapter and if we always default the test suite to the new adapter we have really no way to notice that. Even if deprecated, we should still be sure it's working until it's removed. Even in another PR, I'd experiment with adding another CI job for this, maybe removing another existing job (paperclip, which should be removed/unsupported now if I recall correctly).

kennyadsl (approved):
Approved, but left [a comment](https://github.com/solidusio/solidus/pull/4130#discussion_r713824274) that I think we should address soon. In the meantime, we can merge this, thanks a lot Marc!

core/lib/spree/testing_support/dummy_app.rb
waiting-for-dev: I thought that adding a new CI job was off the table, but that's definitely the best solution. I didn't see a job for paperclip, so probably it's been already removed.

Please, take a look at https://github.com/solidusio/solidus/pull/4176

PR #4134, opened 2021-07-21
Revert "Raise canceling a payment when try_void"

senemsoy:
This reverts [this](https://github.com/solidusio/solidus/commit/4e1002c79) commit, as it was supposed to only stay in v3.0 but got accidentally backported to v2.11.


jarednorman (approved):

spaghetticode (approved):
@senemsoy thank you 👍

PR #4136, opened 2021-07-27
Remove unused `ShippingRateTaxer` service object

aldesantis:
**Description**

The `Spree::Tax::ShippingRateTaxer` service object is not being used anywhere. Its original purpose was to apply taxes to estimated shipping rates, but that logic was moved into `Spree::Stock::Estimator` in https://github.com/solidusio/solidus/commit/36395855f7fa6670bd0d63b12cc238b531ea21d9 for performance reasons. 

The only call site was removed, but the service object itself was not, and it's currently sitting unused in the codebase (I found out about it while writing the [Tax calculation](https://edgeguides.solidus.io/advanced-solidus/tax-calculation) guide).

**Checklist:**


jarednorman (approved):

aldesantis:
@solidusio/core-team can I get one more review on this? 🙏

spaghetticode:
@aldesantis I understand this class has not been used for a long time in Solidus codebase, but I'm wondering should we deprecate it before removing it?

aldesantis:
@spaghetticode I gave it some consideration and I'm not totally sure. I'm sure _someone_ might be using it for some reason but... what reason exactly? The class is not used by Solidus in any way, so I guess the only situation is if someone overrode the shipping rate estimator and their custom estimator uses `ShippingRateTaxer` (or a custom class that inherits from it).

I think when you see it this way, the class is a public API and should be deprecated. On the other, it also feels weird to deprecate a class that has no call sites in Solidus.

jarednorman:
I'm also fine with removing this since it's not in use. I don't see a benefit to deprecating an orphaned class. If it breaks something for some hypothetical person who decided to build off of an interrnal class we aren't using then their test suite should catch it and they can copy the class into their codebase.

spaghetticode (approved):

aldesantis:
We posted a message about this in the Solidus Slack and no one objected, so I'm merging! :shipit:

PR #4138, opened 2021-07-30
Make order-related service objects configurable

aldesantis:
**Description**

Solidus has a few "service objects" that are used to manage orders:

- `Spree::OrderContents`, for managing the contents and the state of an order,
- `Spree::OrderShipping`, for shipping a completed order, and
- `Spree::OrderCancellations`, for managing a completed order's inventory units.

In custom stores, it's common to customize these classes to inject your own logic/checks in different flows.

This PR makes it possible to replace these classes with a custom implementation without monkey-patching the default implementations or the corresponding builders on `Spree::Order`.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -269,7 +269,7 @@ module Spree
     end
 
     def contents
-      @contents ||= Spree::OrderContents.new(self)
+      @contents ||= Spree::Config.order_contents_class.new(self)
     end
 
     def shipping
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -277,7 +277,7 @@ module Spree
     end
 
     def cancellations
-      @cancellations ||= Spree::OrderCancellations.new(self)
+      @cancellations ||= Spree::Config.order_cancellations_class.new(self)
     end
 
     # Associates the specified user with the order.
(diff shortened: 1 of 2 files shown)


waiting-for-dev:
Allowing a neat way to replace services is definitely a must!! Related to that, I'd like to point out a POC we have on the Nebulab's fork: https://github.com/nebulab/solidus/pull/115

Even if merging this PR will definitely be an improvement, maybe we'd like to consider if it's worth it to wait until we have a definitive way to inject services to avoid requiring users to change their code again. If we add more service objects, having an auto-registration mechanism would help to have our codebase cleaner. On top of that, I think there's a benefit to having the injection of objects separated from the more generic configuration.

kennyadsl (approved):
I'm surprised that we are referencing these classes only in this file! Good stuff, thanks Ale.

kennyadsl:
> Even if merging this PR will definitely be an improvement, maybe we'd like to consider if it's worth it to wait until we have a definitive way to inject services to avoid requiring users to change their code again. If we add more service objects, having an auto-registration mechanism would help to have our codebase cleaner. On top of that, I think there's a benefit to having the injection of objects separated from the more generic configuration.

@waiting-for-dev this is definitely interesting but I'd wait to have something in place for injection and switch all the existing preference classes over there in one shot when it's time.

spaghetticode (approved):
@aldesantis thanks 👍

PR #4146, opened 2021-08-15
Make option value to variant association unique

jarednorman:
**Description**

I can't think of a reason why someone would want to associate the same option value with a given variant multiple times. This makes the index on the `variant`/`option_value_id` columns of the `spree_option_values_variants` join table a unique index so we can't have a variant that has the same option value multiple times.

**Checklist:**

### DIFF
--- a/core/db/migrate/20210815004823_add_unique_index_to_option_values_variants.rb
+++ b/core/db/migrate/20210815004823_add_unique_index_to_option_values_variants.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class AddUniqueIndexToOptionValuesVariants < ActiveRecord::Migration[5.2]
+  def up
+    remove_index :spree_option_values_variants, [:variant_id, :option_value_id]
+    add_index :spree_option_values_variants, [:variant_id, :option_value_id],
+              name: "index_option_values_variants_on_variant_id_and_option_value_id",
+              unique: true
+  end
+
+  def down
+    remove_index :spree_option_values_variants, [:variant_id, :option_value_id]
+    add_index :spree_option_values_variants, [:variant_id, :option_value_id],
+              name: "index_option_values_variants_on_variant_id_and_option_value_id"
+  end
+end
--- a/core/spec/models/spree/variant_spec.rb
+++ b/core/spec/models/spree/variant_spec.rb
@@ -1042,4 +1042,9 @@ RSpec.describe Spree::Variant, type: :model do
     subject { variant.sku_and_options_text }
     it { is_expected.to eq("EB1 Size: S") }
   end
+
+  it "cannot have the same option value multiple times" do
+    expect { variant.option_values << variant.option_values.first }
+      .to raise_error ActiveRecord::RecordNotUnique
+  end
 end


jarednorman:
I don't think any of the errors have anything to do with my change. 🤔

jarednorman:
Had to scroll back more. I used 6.2 as the migration version, but currently we have to write migrations for the lowest supported ActiveRecord version.

spaghetticode (approved):
🎉 looks good to me, thanks @jarednorman !

core/db/migrate/20210815004823_add_unique_index_to_option_values_variants.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class AddUniqueIndexToOptionValuesVariants < ActiveRecord::Migration[5.2]
+  def up
+    remove_index :spree_option_values_variants, [:variant_id, :option_value_id]
+    add_index :spree_option_values_variants, [:variant_id, :option_value_id],
+              name: "index_option_values_variants_on_variant_id_and_option_value_id",
```
spaghetticode, line 7: out of curiosity, why the explicit index name?

jarednorman: Rails will generate an index name that exceeds the 64 character limit. The name I've used just matches the existing index's name, so presumably whoever created the original index also ran into this.

waiting-for-dev (approved):

mamhoff:
This is not to hold this back: We associate the same option value with multiple variants. We have a third field, an integer, that differentiates the variants from one another. We are a special case though as we sell a lot of stuff in differently sized bags.

jarednorman:
@mamhoff I don't think this will impact you then? You can still have multiple variants per option value, just not the the same option value on one variant multiple times.

mamhoff:
Ah true! I mistakenly thought we need to have different option value combos for each variant. :+1:

kennyadsl:
Hey @jarednorman, can you please rebase this one? Thanks! 🙏

waiting-for-dev:
> Hey @jarednorman, can you please rebase this one? Thanks! pray

I did it myself and fixed the conflict

BenMorganIO:
This would be a good chance to mark the `variant_id` and the `option_value_id` as `NOT NULL` and add the two foreign key constraints. I would assume `ON DELETE CASCADE ON UPDATE CASCADE` would be appropriate here.

kennyadsl:
> This would be a good chance to mark the variant_id and the option_value_id as NOT NULL and add the two foreign key constraints. I would assume ON DELETE CASCADE ON UPDATE CASCADE would be appropriate here.

We decided to skip this kind of changes for this major release. There are already a bunch of database changes and we don't want to be a nightmare to update. 

@BenMorganIO thanks a lot for the suggestion, I think we should design a process to make these changes safer. An idea is a script that runs at boot and checks the database (especially important in production) and gives some hints if you need to fix inconsistencies. BTW, we can discuss in Slack or in a separate issue/PR if you want.

PR #4155, opened 2021-08-26
Hide the master variants from stock management

tmtrademarked:
This PR hides the master variants from the stock management list if the product has variants. This prevents a strange experience when managing inventory for these products - they don't really have SKUs, and aren't sellable in the general case.

Before:
Screen Shot 2021-08-26 at 3 20 29 PM

After:
image


jarednorman (approved):
That's perfect. This looks good to me. I can't think of a reason right now that this would cause any issues for stores I work on.

kennyadsl (approved):
Thank you Tom!

PR #4156, opened 2021-08-31
Fix gateway_error when no order is defined

alexblackie:
If the current controller raises a gateway error, but does not have
`@order` defined, this error handler will raise a NoMethodError when it
tries to access `nil.errors`.

We can resolve this by falling back to an empty Order instance if none
is defined. This maintains the existing behaviour, and as the response
only serializes the errors hash, there shouldn't be any surprises with
null fields, etc.

**Checklist:**

### DIFF
--- a/api/spec/controllers/spree/api/base_controller_spec.rb
+++ b/api/spec/controllers/spree/api/base_controller_spec.rb
@@ -145,7 +145,7 @@ describe Spree::Api::BaseController, type: :controller do
       get :index
     end
 
-    context "with @order defined" do
+    context "with @order not defined" do
       controller(Spree::Api::BaseController) do
         def index
           raise Spree::Core::GatewayError, "Insufficient Funds"
@@ -167,7 +167,7 @@ describe Spree::Api::BaseController, type: :controller do
       end
     end
 
-    context "with @order not defined" do
+    context "with @order defined" do
       controller(Spree::Api::BaseController) do
         def index
           @order = Spree::Order.new
--- a/api/spec/controllers/spree/api/base_controller_spec.rb
+++ b/api/spec/controllers/spree/api/base_controller_spec.rb
@@ -138,4 +138,58 @@ describe Spree::Api::BaseController, type: :controller do
       end
     end
   end
+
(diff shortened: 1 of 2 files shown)


api/spec/controllers/spree/api/base_controller_spec.rb
```diff
@@ -138,4 +138,58 @@ def index
...
+      end
+
+      it "serializes the errors" do
+        expect(JSON.parse(response.body)["errors"]).to(
+          match(hash_including({ "base" => ["Insufficient Funds"] }))
+        )
+      end
+    end
+
+    context "with @order not defined" do
```
jarednorman, line 170: Are your contexts backwards?

jarednorman (changes requested):

api/spec/controllers/spree/api/base_controller_spec.rb
alexblackie: .......

kennyadsl:
Thanks, @alexblackie! Just to add some context, in which circumstances can this happen? I saw the specs but with a standard checkout flow it's very unlikely that we'll have an exception related to the gateway in a controller that is not related to the checkout where, I guess, having an order is mandatory.

alexblackie:
@kennyadsl Ah yeah, for context, this was raised in a custom reimbursements API controller, which interacts with payments (refunds) but does not have an `@order`.

Since the `:authorize_for_order` filter is conditional on params, there is no guarantee `@order` will be set even in Core controllers, unless the child controller assigns it. Also from an architecture/abstraction point of view it seems wrong to have a `rescue_from` in a base controller make assumptions about pre-existing internal state, especially if that state came from a child.

jarednorman (approved):
Makes sense to me. Thanks, Alex!

kennyadsl:
> Also from an architecture/abstraction point of view it seems wrong to have a rescue_from in a base controller make assumptions about pre-existing internal state, especially if that state came from a child.

Agree, thanks for sharing the context!

kennyadsl (approved):

PR #4144, opened 2021-08-09
Prepare CHANGELOG for Solidus 3.1 release

waiting-for-dev:
**Description**

Prepare CHANGELOG for Solidus 3.1 release

The new entries have been generated through the following command, as detailed in the [how-to-release guides](https://github.com/solidusio/solidus/wiki/How-to-release-Solidus#make-sure-changelogmd-is-up-to-date):

```bash
git rev-list v3.0.1...master | github_fast_changelog solidusio/solidus
```
The following items have been removed because of being change/revert pairs or being changes on top of still-unreleased stuff:

- Docker: Change public keys server after deprecation [#4140](https://github.com/solidusio/solidus/pull/4140) ([waiting-for-dev](https://github.com/waiting-for-dev))
- Revert "Deprecate callback setting variant price from master" [#4132](https://github.com/solidusio/solidus/pull/4132) ([waiting-for-dev](https://github.com/waiting-for-dev))
- Deprecate callback setting variant price from master [#4078](https://github.com/solidusio/solidus/pull/4078) ([waiting-for-dev](https://github.com/waiting-for-dev))
- Remove staled comment [#4100](https://github.com/solidusio/solidus/pull/4100) ([waiting-for-dev](https://github.com/waiting-for-dev))
- Disable perform_caching in DummyApp [#4119](https://github.com/solidusio/solidus/pull/4119) ([gsmendoza](https://github.com/gsmendoza))
- Master now tracks 3.1.0 [#4033](https://github.com/solidusio/solidus/pull/4033) ([kennyadsl](https://github.com/kennyadsl))

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
- [ ] I have updated Guides and README accordingly to this change (if needed)
- [ ] I have added tests to cover this change (if needed)
- [ ] I have attached screenshots to this PR for visual changes (if needed)

PR #4157, opened 2021-09-06
Remove the upgrade task and point to additional steps from the update generator

waiting-for-dev:
**Description**

Having both an `upgrade` task and an `update` generator is confusing. On
top of that, it's better to have actionable items, which can be
undesirable in some situations, the most explicit as possible. For this
reason, we leave in place the `update` generator, which is just a safe
path to update preferences, and remove the `upgrade` task. Instead, we
add a generic message into the `update` generator pointing to the
detailed instructions in the Changelog.

In particular, at this point, the `update` task was calling to a single
task introduced in #3987. We still keep the task reachable by end-users,
but we rename it slightly to be more friendly.

**Checklist:**


kennyadsl (commented):
Maybe we want to also update https://github.com/solidusio/solidus/pull/4144 to add this in the important changes section?

waiting-for-dev:
> Maybe we want to also update #4144 to add this in the important changes section?

Yes, I'll do it right next :slightly_smiling_face:

kennyadsl (approved):

kennyadsl (approved):

jarednorman (approved):

PR #4164, opened 2021-09-13
Fix searching deleted products by SKU

### LINKED ISSUE #3912: Searching deleted products by SKU does not work

adammathys:
**Description**

We've added two new scopes to `Spree::Product` and deprecated on existing one.
- `with_variant_sku_cont`: Has been deprecated in favour of `with_kept_variant_sku_cont`
- `with_all_variant_sku_cont`: Will search for products that have a variant containing the provided SKU, even if it has been discarded.
- `with_kept_variant_sku_cont`: Maintains the existing behaviour of `with_variant_sku_cont`, but with a more descriptive name.

This allows us, along with a new Backbone view to help control the two inputs we now have on the page (one for each scope), to correctly return deleted products when searching by SKU. Our view will swap out the inputs based on the value of the "with_deleted" checkbox so that the user experience is unchanged.

While attempting to fix this issue, there are a few other approaches we tried before settling on this one:
- Updating the scope to always include discarded variants. This came close but had an issue when the product had some discarded variants and would return the product regardless of the value of the "with_deleted" checkbox.
- Adding a parameter to the scope to conditionally include discarded variants. Ransack had some trouble with parameterizing the values from this. It would add extra stuff to the input field.

Fixes #3912

**Checklist:**
- [ ] ~~I have updated Guides and README accordingly to this change (if needed)~~
- [ ] ~~I have attached screenshots to this PR for visual changes (if needed)~~

### DIFF
--- a/backend/app/assets/javascripts/spree/backend.js
+++ b/backend/app/assets/javascripts/spree/backend.js
@@ -47,6 +47,7 @@
 //= require spree/backend/orders
 //= require spree/backend/payments/edit
 //= require spree/backend/payments/new
+//= require spree/backend/products/index
 //= require spree/backend/product_picker
 //= require spree/backend/progress
 //= require spree/backend/promotions
--- a/backend/app/assets/javascripts/spree/backend/namespaces.js
+++ b/backend/app/assets/javascripts/spree/backend/namespaces.js
@@ -8,6 +8,7 @@ _.extend(window.Spree, {
     Cart: {},
     Zones: {},
     Payment: {},
+    Product: {},
     Promotions: {},
     Stock: {},
     Tables: {}
--- a/backend/app/assets/javascripts/spree/backend/products/index.js
+++ b/backend/app/assets/javascripts/spree/backend/products/index.js
@@ -1,0 +1,7 @@
+Spree.ready(function() {
+  if ($('[data-hook="admin_products_index_search"]').length) {
+    new Spree.Views.Product.Search({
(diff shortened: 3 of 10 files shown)


jarednorman (approved):
This all looks reasonable to me! Thanks, @adammathys!

kennyadsl (approved):
Thanks for the contribution, @adammathys!

PR #4168, opened 2021-09-14
Respect completed_at timestamp in order factories

DianeLooney:
The order factories that automatically complete orders previously used
Time.current when setting the completed_at timestamp. For developers who want
to pre-date their order completions, this can be confusing. This change allows
orders to be cleanly pre-dated in this way.

**Checklist:**

### DIFF
--- a/core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
+++ b/core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
@@ -122,6 +122,26 @@ RSpec.shared_examples "an order with line items factory" do |expected_order_stat
     end
   end
 end
+
+RSpec.shared_examples 'supplied completed_at is respected' do
+  context 'when passed a completed_at timestamp' do
+    let(:completed_at) { 2.days.ago }
+    let(:order) { create(factory, completed_at: completed_at) }
+
+    it 'respects the timestamp' do
+      expect(order.completed_at).to be_within(5.seconds).of(completed_at)
+    end
+  end
+
+  context 'when no completed_at timestamp is passed' do
+    let(:order) { create(factory) }
+
+    it 'defaults to the current time' do
+      expect(order.completed_at).to be_within(5.seconds).of(Time.current)
+    end
+  end
+end
+
 RSpec.describe 'order factory' do
   let(:factory_class) { Spree::Order }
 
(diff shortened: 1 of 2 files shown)


jarednorman (approved):

DianeLooney:
Mysql tests failed because I was too granular with my timestamp comparison. I updated it to 5 seconds and force-pushed the fix. Sorry if anyone happened to catch the review super early!

DianeLooney:
The failing postgres test looks like spec flake to me, I can reproduce it locally by `.reverse`ing the expected value. I'll open up a fix in another PR while I'm at it.

DianeLooney:
https://github.com/solidusio/solidus/pull/4169

Happy to simply rebase if/when this is merged, or if someone with ci powers wants to re-run for me I'd be happy with that too.

core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
```diff
@@ -122,6 +122,26 @@
...
+    it 'respects the timestamp' do
+      expect(order.completed_at).to be_within(5.seconds).of(completed_at)
+    end
+  end
+
+  context 'when no completed_at timestamp is passed' do
+    let(:order) { create(factory) }
+
+    it 'defaults to the current time' do
+      expect(order.completed_at).to be_within(5.seconds).of(Time.current)
```
jarednorman, line 140: `1.second` should be plenty of leeway.

jarednorman (approved):
Looks good, though 5 seconds is probably excessive.

core/spec/lib/spree/core/testing_support/factories/order_factory_spec.rb
DianeLooney: As mentioned here, I actually ran into an issue with the 1s time when posting the PR. The test ran at the rollover to a new second. Clock time at the start of the test was `xx:yy:01.995`, mysql stored `xx:yy:01.000` and the comparison tested against `xx:yy:02.016` (or something similar to that). Happy to move it back down to 2s which should account for this, if desired.

Perhaps a better test than this would be to record the time before and after the factory runs and test that the completed_at is between those two values?

DianeLooney: Actually the range idea still doesn't work because of the truncation.

I'll just update it to be 2 seconds instead.

DianeLooney:
```suggestion
      expect(order.completed_at).to be_within(2.seconds).of(Time.current)
```

kennyadsl (approved):
Looks great, thanks!

PR #4170, opened 2021-09-15
Add deprecation path for arity-zero preference defaults

### LINKED ISSUE #4165: Error passing procs to Preferences in Solidus 3.1
I tried to upgrade to Solidus 3.1 but I have a blocking issue with some extensions that pass a proc to a preference, not sure if it's not allowed anymore but I haven’t seen anything about that in the migration guide or in a previous deprecation 🤔 

**Solidus Version**: 3.1.0

**To Reproduce**

Extension code:
```ruby
class Spree::AvataxConfiguration < Spree::Preferences::Configuration
  preference :environment, :string, default: -> { default_environment }
  ...
```

Error:
```
/app/models/spree/avatax_configuration.rb:7:in `block i
n <class:AvataxConfiguration>': wrong number of arguments (given 1, expected 0) (ArgumentError)
        from /usr/local/lib/ruby/gems/3.0.0/gems/solidus_core-3.1.0/lib/spree/preferences/preferable_class_methods.rb:64:in `block in preference'
```

Related Solidus change:
https://github.com/solidusio/solidus/commit/b897eccb6cdd8235d782275e9df069d95474cd6b#diff-1d7628c9b0c063921fa05554bc947bcd7d512427a47f6e7d709256ccd66e3601L63-R65

waiting-for-dev:
**Description**

Solidus 3.1 shipped with a new preferences system, where a default can
take different values depending on the Solidus version defaults that
have been loaded. See #4064 for details.

To achieve those above, when a proc is given as the default value
constructor, it now should take the loaded Solidus version as an
argument. That's a breaking change, as some extensions or user-defined
preferences may be using zero-arity lambdas.

This commit deprecates zero-arity lambdas but wraps them into another
lambda, taking and disregarding a single argument. That's only needed
for procs with lambda semantics, as raw procs will ignore the provided
extra argument.

When it comes to the implementation, as it's something to be ditched in
the next major release, we've opted for the more straightforward
solution. I.e., wrapping the lambda into the `Preferable` module even if
it only affects `AppConfiguration` classes. The default-handling logic
is very entangled into the former, and it'd take more work to extract
it.

Fixes #4165

**Checklist:**

### DIFF
--- a/core/lib/spree/preferences/preferable_class_methods.rb
+++ b/core/lib/spree/preferences/preferable_class_methods.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'spree/deprecation'
 require 'spree/encryptor'
 
 module Spree::Preferences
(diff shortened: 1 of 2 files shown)


kennyadsl (approved):
Thanks @waiting-for-dev, we have to release a new patch after this is merged.

jarednorman (approved):
Thanks. Seems like a good solution.

PR #4172, opened 2021-09-15
Recalculate promotions after destroying/refreshing shipments

spaghetticode:
**Description**

This change has basically zero impact on a standard Solidus store, but is beneficial on stores that removed the `address` order state, as in that case [`order.ensure_updated_shipments`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L514) would actually recreate shipments when sending the order to `delivery` state, courtesy of `next!` from [`Order#restart_checkout_flow`](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L522) (in a Solidus default store calling `Order#ensure_updated_shipments` simply destroys the shipments leaving the order in `address` state).

The current order of execution applies all promotions, including the ones for shipping, but these will then be removed by the subsequent call to `Order#ensure_updated_shipments`. 
Swapping the lines solves the issue, otherwise, stores need to re-apply shipping promotions with a local patch.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order_contents.rb
+++ b/core/app/models/spree/order_contents.rb
@@ -42,8 +42,8 @@ module Spree
           # If we do not update first, then the item total will be wrong and ItemTotal
           # promotion rules would not be triggered.
           reload_totals
-          PromotionHandler::Cart.new(order).activate
           order.ensure_updated_shipments
+          PromotionHandler::Cart.new(order).activate
         end
         reload_totals
         true


jarednorman (approved):

kennyadsl (approved):

5minpause:
Hi. I am looking through some of the PRs and wondering why this one wasn't merged. Is it held up by something specific?

jarednorman:
I'm not sure we've discussed whether this belongs in the next minor or major release yet.

mamhoff (approved):
This makes sense.

waiting-for-dev:
Hey @spaghetticode, could we have a regression test for it?

> I'm not sure we've discussed whether this belongs in the next minor or major release yet.

I'd say it's fixing a bug, so it should go to the next minor.

waiting-for-dev:
@kennyadsl, do you agree it's ok to ship it in the next minor?

kennyadsl:
@waiting-for-dev Yes, I think this is a bug.

spaghetticode:
@waiting-for-dev sorry for the late rebase 😓

jarednorman (approved):

waiting-for-dev:
Thanks, @spaghetticode !

PR #4173, opened 2021-09-16
Rename Order#ensure_updated_shipments method

spaghetticode:
**Description**

The naming of this method is confusing and it may induce developers to believe it refreshes shipments while it actually only destroys them on a standard Solidus store.

The method definition is thoroughly tested in the codebase, and the wanted behavior is to actually remove shipments without recreating them back.

The old name is still available but deprecated with a message that suggests using the new one.

FYI if the original order's state machine is patched for example by removing the `address` state, then shipments will be actually recreated since the order will be `next`ed to the `delivery` state and a callback will create them again.

I'm not 100% satisfied with the new name `clean_shipments_and_restart_checkout`, if anybody has a better one I'll be happy to change it.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -502,18 +502,18 @@ module Spree
       recalculate
     end
 
-    # Clean shipments and make order back to address state
-    #
-    # At some point the might need to force the order to transition from address
-    # to delivery again so that proper updated shipments are created.
-    # e.g. customer goes back from payment step and changes order items
-    def ensure_updated_shipments
+    # Clean shipments and make order back to address state (or to whatever state
+    # is set by restart_checkout_flow in case of state machine modifications)
+    def check_shipments_and_restart_checkout
       if !completed? && shipments.all?(&:pending?)
         shipments.destroy_all
         update_column(:shipment_total, 0)
         restart_checkout_flow
       end
     end
+
+    alias_method :ensure_updated_shipments, :check_shipments_and_restart_checkout
+    deprecate ensure_updated_shipments: :check_shipments_and_restart_checkout, deprecator: Spree::Deprecation
 
     def restart_checkout_flow
       return if state == 'cart'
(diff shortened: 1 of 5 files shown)


jarednorman (approved):

kennyadsl:
> I'm not 100% satisfied with the new name clean_shipments_and_restart_checkout, if anybody has a better one I'll be happy to change it.

Actually, "Cleaning shipments" makes me think of some sort of moving them in a cleaner state.

What about `empty_shipments_and_restart_checkout`? "Empty" can be misleading as well, maybe; it could mean that the shipments are still there but their contents are removed.

Another alternative could be `remove_shipments_and_restart_checkout`, WDYT?

spaghetticode:
@kennyadsl I didn't want to use a term that strongly conveys the meaning of absolutely destroying/removing shipments, because this doesn't happen always (i.e. when order is completed it doesn't) and the final outcome also happens depends on what does later `restart_checkout_flow`. That's the reason for my naming 😅

waiting-for-dev:
Hey! This one is ready to be merged, but the naming is holding it back :slightly_smiling_face: 

What about `check_shipments_and_restart_checkout`?

spaghetticode:
@waiting-for-dev looks good to me, I'm going to change the name according to your suggestion 👍

waiting-for-dev:
Hey @spaghetticode, sorry for the hassle, but, can you rebase from master so that specs are green?

kennyadsl:
@spaghetticode last rebase, I swear it will be merged this time! ❤️

kennyadsl:
@spaghetticode I think something is messed with the reabase, maybe it's because of the `master` -> `main` rename?

spaghetticode:
@kennyadsl yes, that was the issue! Finally it's now green 💚

kennyadsl (approved):

elia (approved):

kennyadsl:
> I swear it will be merged this time!

Promise kept! :trollface:

PR #3913, opened 2021-02-01
Switch to maintained Paperclip fork

filippoliverani:
**Description**

Since Paperclip dependency is not going away anytime soon, it would be better switching to the official maintained fork to keep up with security updates and Ruby 3 compability.

Paperclip has been deprecated for 3 years and abandoned for a year, [the creators suggest](https://thoughtbot.com/blog/paperclip-new-maintainers) to migrate to the maintained fork [Paperclip-kt](https://github.com/kreeti/kt-paperclip).

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
- ~~[ ] I have updated Guides and README accordingly to this change (if needed)~~
- ~~[ ] I have added tests to cover this change (if needed)~~
- ~~[ ] I have attached screenshots to this PR for visual changes (if needed)~~

PR #4174, opened 2021-09-21
Backport #3913 to V2.11

spaghetticode:
**Description**

This PR backports #3913 to Solidus 2.11. 

In order to avoid tedious warnings on Ruby 2.7 and greater, we need to switch to the maintained fork of paperclip on 2.11 as well. The deprecation is now [fixed](https://github.com/kreeti/kt-paperclip/commit/f0367733ddf37141ae00517c6697cc3dd233dab9) in kt-paperclip 6.4.x.

**Checklist:**

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -39,7 +39,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'kaminari-activerecord', '~> 1.1'
   s.add_dependency 'mini_magick', '~> 4.10'
   s.add_dependency 'monetize', '~> 1.8'
-  s.add_dependency 'paperclip', '>= 4.2'
+  s.add_dependency 'kt-paperclip', '>= 4.4.0'
   s.add_dependency 'paranoia', '~> 2.4'
   s.add_dependency 'ransack', '~> 2.0'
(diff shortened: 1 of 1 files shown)


kennyadsl:
I'm not totally convinced about changing this dependency on a patch-level release. Are we sure using `kt-paperclip` at the version allowed is preserving full backward compatibility with `paperclip`?

spaghetticode:
@kennyadsl if it helps, I can test it briefly on a real project (specs there were already green, but I'm unsure how much Paperclip is actually tested VS mocked there) unless your issue is with the major version requirements jump (from `>= 4.3` to `~> 6.3`).

In regards to breaking changes with the last version of Paperclip at `6.1`, the new project seems to be following semantic versioning, their [changelog](https://github.com/kreeti/kt-paperclip/blob/master/NEWS) includes only fixes and small functionality additions, so it looks safe to me from that point of view.

tvdeyen (approved):
I think this is a good thing, even in a patch level

spaghetticode:
I'm happy to announce that kt-paperclip maintainers have released 4.x and 5.x versions, so we can now avoid bumping to a major version with this PR 🎉

PR #4176, opened 2021-09-23
Centralize legacy event bus deprecation and test legacy on CI

waiting-for-dev:
**Description**

This commit adds a new job on CI to perform the test suite with the
legacy adapter configured.

Besides that, it extracts the deprecation logic into its module to avoid
having it scattered through different parts of the system.

Relates to discussion:

https://github.com/solidusio/solidus/pull/4130#discussion_r682733694

**Checklist:**

### DIFF
--- a/core/spec/lib/spree/app_configuration_spec.rb
+++ b/core/spec/lib/spree/app_configuration_spec.rb
@@ -135,7 +135,7 @@ RSpec.describe Spree::AppConfiguration do
     expect(prefs.admin_vat_location.country_id).to eq(nil)
   end
 
-  it 'can access event adapter' do
-    expect(prefs.events.adapter).to be_an_instance_of(Spree::Event::Adapters::Default)
+  it 'can access event configuration' do
+    expect(prefs.events).to be_an_instance_of(Spree::Event::Configuration)
   end
 end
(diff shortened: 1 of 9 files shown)


core/spec/lib/spree/app_configuration_spec.rb
```diff
@@ -136,6 +136,6 @@ class DummyClass; end;
   end
 
   it 'can access event adapter' do
-    expect(prefs.events.adapter).to be_an_instance_of(Spree::Event::Adapters::Default)
+    expect(prefs.events.adapter).to be(Spree::Config.events.adapter)
```
jarednorman, line 139: This test is circular. `prefs` is `Spree::Config`, so this is just testing that `Spree::Config.events.adapter` is `Spree::Config.events.adapter`.

jarednorman (changes requested):

core/spec/lib/spree/app_configuration_spec.rb
waiting-for-dev: Oh, correct:sweat_smile: I changed it to test what it's relevant from the app configuration point of view; the event configuration is accessible from the application configuration.

jarednorman (approved):

kennyadsl (approved):
Thanks @waiting-for-dev! Let's monitor how much it adds to the test suite time but I think we should be good.

PR #4177, opened 2021-09-24
Fix Spree::Api:LineItemsController#create handling validation errors

RyanofWoods:
Spree::Api:LineItemsController#create calls #add on an OrderContents
instance which eventually call #save!. This raises errors if validation
fails. #create was not handling these validation errors. In fact, it had
an if statement checking for errors on a LineItem, but this could never
be returned if errors were raised.

-----
I can include changes to the documentation if the code looks good 👍 

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/line_items_controller.rb
+++ b/api/app/controllers/spree/api/line_items_controller.rb
@@ -11,16 +11,15 @@ module Spree
 
       def create
         variant = Spree::Variant.find(params[:line_item][:variant_id])
-        @line_item = @order.contents.add(
-          variant,
-          params[:line_item][:quantity] || 1,
-          options: line_item_params[:options].to_h
-        )
-
-        if @line_item.errors.empty?
+        begin
+          @line_item = @order.contents.add(
+            variant,
+            params[:line_item][:quantity] || 1,
+            options: line_item_params[:options].to_h
+          )
           respond_with(@line_item, status: 201, default_template: :show)
-        else
-          invalid_resource!(@line_item)
+        rescue ActiveRecord::RecordInvalid => error
+          invalid_resource!(error.record)
         end
       end
 
--- a/api/spec/requests/spree/api/line_items_spec.rb
+++ b/api/spec/requests/spree/api/line_items_spec.rb
@@ -95,6 +95,12 @@ module Spree::Api
         expect(response.status).to eq(201)
       end
 
+      it '#create calls #invalid_resource! if adding a line item fails validation' do
+        allow_any_instance_of(Spree::LineItem).to receive(:valid?).and_return(false)
+        expect_any_instance_of(Spree::Api::BaseController).to receive(:invalid_resource!).once
+        post spree.api_order_line_items_path(order), params: { line_item: { variant_id: product.master.to_param, quantity: 1 } }
+      end
+
       it "default quantity to 1 if none is given" do
         post spree.api_order_line_items_path(order), params: { line_item: { variant_id: product.master.to_param } }
         expect(response.status).to eq(201)


spaghetticode:
@RyanofWoods thanks for this PR! I think it makes sense, although I would like to discuss some doubts that come to my mind. 

The first one is about the fact that the focus changes from the line item to the order. Also, I'm wondering if we should add another action (i.e. `batch_create`) for creating multiple line items and leave this one for a single one. 

Then, I wonder if changing the implementation of `OrderContents#add` to call `save` instead of `save!` could fix the issue here without much change to the controller action. I don't see the failure case tested on `OrderContents` specs, so I guess we're underspecifying the behavior and there may be an opportunity to fix that. WDYT?

RyanofWoods:
Hi @spaghetticode, thanks for the review! 🙂 

Having another action for adding multiple items would probably be best 👍 If a method gets added to the [OrderContents](https://github.com/solidusio/solidus/blob/a54f7139b6bc23d71697fcaada0aebba90614c34/core/app/models/spree/order_contents.rb#L3-L4) class then the `API` and `Frontend` can both utilize it.

Having the `OrdersContent` use `#save` instead of `#save!` would allow more simple logic when adding a variant, instead of having to catch the error and then insert it back in like [here](https://github.com/solidusio/solidus/blob/a54f7139b6bc23d71697fcaada0aebba90614c34/frontend/app/controllers/spree/orders_controller.rb#L62-L66). If the change gets done for `#add_to_line_item` then I think we should do it [here](https://github.com/solidusio/solidus/blob/a54f7139b6bc23d71697fcaada0aebba90614c34/core/app/models/spree/order_contents.rb#L110) for `#remove_from_line_item` to have consistency.

Do you think there's more of a risk of developers not handling the errors if it doesn't raise? Or it is something we shouldn't worry about.

Note, this change would mean updating the frontend code, however, I currently have a PR open to move this logic out 🤔 https://github.com/solidusio/solidus/pull/4102

waiting-for-dev:
The idea is to deprecate `solidus_frontend` eventually, so maybe we should not worry too much about it. However, changing `#add` to return false would be a breaking change in itself, so we should manage it accordingly.

I agree that it'd be best to keep the interface for the current end-point and create a new one for the batch action.

waiting-for-dev:
BTW, we should make sure that the raising on `#add` is not used to abort a transaction somewhere else.

waiting-for-dev:
Hey, @RyanofWoods! Would you still like to work on it? As a minimal change, we could fix rendering the validation errors to the end-user instead of raising them.

waiting-for-dev:
Hey, @RyanofWoods, pinging you again, in case you'd like to complete it 🙂

RyanofWoods:
I will look into this again soon (friday or after my two weeks vacation) @waiting-for-dev

api/spec/requests/spree/api/line_items_spec.rb
```diff
@@ -95,6 +95,12 @@ module Spree::Api
         expect(response.status).to eq(201)
       end
 
+      it '#create calls #invalid_resource! if adding a line item fails validation' do
```
waiting-for-dev, line 98: Instead of testing the implementation, I think, in this case, we can test the actual behavior. I mean sending invalid data and checking that the response contains an error. Maybe we could try sending `-1` as quantity?

waiting-for-dev (commented):
I left a suggestion about the test, but the application code is perfect 👍

api/spec/requests/spree/api/line_items_spec.rb
RyanofWoods: I thought of this as well, but I am afraid, we can't do it through variant or quantity. LineItem's quantity gets normalized to zero upon validation if it's `nil` or negative:
https://github.com/solidusio/solidus/blob/6dae6359e5d8cf7a83aa473e59f7888d2090f6f1/core/app/models/spree/line_item.rb#L25
https://github.com/solidusio/solidus/blob/6dae6359e5d8cf7a83aa473e59f7888d2090f6f1/core/app/models/spree/line_item.rb#L137-L139

So in away, this is really a non-occurring problem, but it's problematic if validations on `LineItem` changed.

waiting-for-dev: Oh, I see. It looks like there's no point in testing the unhappy path, as that's something not considered by our business logic (which explains why we never found an issue here). I wonder if it makes sense to make the controller code less defensive and acknowledge that it should never fail.

RyanofWoods: It makes partial sense to simplify the endpoint if cannot currently fail, but then it's always relying on that fact, and therefore if `LineItem` received a new validation, this endpoint would start failing and it would not be realised by the specs 🤔 

Also about perhaps extracting the following line to a validation on `LineItem`? If we did this, we would need this this API endpoint to handle LineItem errors:
https://github.com/solidusio/solidus_starter_frontend/blob/fa62873bf7d06ef8054de1dbba8ca5bee27a1cfc/templates/app/controllers/cart_line_items_controller.rb#L19

waiting-for-dev: I wouldn't be defensive trying to anticipate what can happen in the future. Our system should reflect the current state of things. However, we can also argue if it makes sense to normalize the quantity to zero when a negative number is given. I think a consumer would be better served if an explicit error is given in that case. Happy to hear other thoghts. cc @solidusio/core-team

kennyadsl: I tried to change that part in the past, without a respectable patch to be proposed. I think that normalization shouldn't happen in a before_validation. It would be great to move that code in a validation and I think that this comes from the feature that allows to remove line items if the quantity is 0, which is more of a UX problem and should not be a model responsibility. 

Solving this problem will also clean up several tangential parts of the codebase, like these specs, but of course, it's not part of this PR's purpose. 


That said, for this specific spec, what if we add some fake errors to the line item and check the error is present in the response?

```ruby
allow_any_instance_of(Spree::LineItem).to receive(:errors).and_return("something")
post spree.api_order_line_items_path(order), params: { line_item: { variant_id: product.master.to_param, quantity: 1 } }
expect(json_response["error"]).to eq('something')
```

Of course, this is just pseudo-code I didn't actually check, only useful to explain better what I meant. 😆

RyanofWoods: I am happy to open up a PR to investigate removing the quantity normalization if desired by the core team. If you want, we could keep this PR open until the change is actually required (such as removing the quantity normalization).

kennyadsl: If this PR solves a real store problem I won't block merging it. 

It would be great to see if we can get rid of the normalization and just use some plain validations; if you want to have a quick conversation around that before starting, please ping me privately, I would be happy to help.

waiting-for-dev: Yeah, I agree. We can merge this one and work on the actual fix afterward.

waiting-for-dev: @kennyadsl solution is a good compromise for now

RyanofWoods: I will update the spec to use a quantity of 0 once we update validation.

RyanofWoods:
Rebased

waiting-for-dev (approved):

rainerdema (approved):
👍

api/app/controllers/spree/api/line_items_controller.rb
```diff
@@ -11,16 +11,15 @@ def new
...
+          @line_item = @order.contents.add(
+            variant,
+            params[:line_item][:quantity] || 1,
+            options: line_item_params[:options].to_h
+          )
           respond_with(@line_item, status: 201, default_template: :show)
-        else
-          invalid_resource!(@line_item)
+        rescue ActiveRecord::RecordInvalid => error
+          invalid_resource!(error.record)
```
elia, line 22: I'm wondering ig this could/should be turned into a `rescue_from` in the API::BaseController

elia (approved):
Good catch! 🙌

PR #4179, opened 2021-09-27
Fix loading core on Rails < 6.1

### LINKED ISSUE #4178: undefined method `paths' when running solidus update generator
Running the `rails g solidus:update` with solidus version 3.1.0 and Rails version < 6.1 causes the following error. 

**Solidus Version**:
3.1.1

**To Reproduce**
Create a fresh install with Solidus 3.1.1 and Rails 6.0.0
Run `rails g solidus:update`

**Current behavior**
`NoMethodError: undefined method `paths' for #<Rails::Paths::Path:0x00007f8721d650b0>`

**Expected behavior**
The solidus update generator runs without errors

 - OS: MacOS 11.5.2

**Additional context**
The error is coming from `Core.has_install_generator_been_run?`

waiting-for-dev:
**Description**

To avoid warning about `Spree.load_defaults` not being defined at a
moment when `bin/rails g solidus:install` has not even been called, the
`Spree::Core` engine is checking whether the `spree.rb` initializer
already exists (as it's there where `#load_defaults` should be called).

We're looking for `spree.rb` within the directories associated with the
`Rails::Paths::Path` instance for initializers. However, before this
commit, we were using `Rails::Paths::Path#paths` method, which is only
available on Rails >= 6.1.0. That made the loading of `Spree::Core`
crash.

As `Rails::Paths::Path` is an `Enumberable`, we can use `any?` directly
from the instance and manually build the `Pathname` for each yielded
string instead.

We also added tests to cover the previous failure and changed the naming
to make the purpose clearer.

Fixes #4178

**Checklist:**

### DIFF
--- a/core/lib/spree/core.rb
+++ b/core/lib/spree/core.rb
@@ -62,12 +62,23 @@ module Spree
   end
 
   module Core
-    def self.has_install_generator_been_run?
-      (Rails.env.test? && Rails.application.class.name == 'DummyApp::Application') ||
-        Rails.application.paths['config/initializers'].paths.any? do |path|
-          File.exist?(path.join('spree.rb'))
-        end
+    # @api private
+    def self.has_install_generator_been_run?(rails_paths: Rails.application.paths, initializer_name: 'spree.rb', dummy_app_name: 'DummyApp::Application')
+      does_spree_initializer_exist?(rails_paths, initializer_name) ||
+        running_solidus_test_suite_with_dummy_app?(dummy_app_name)
     end
+
+    def self.running_solidus_test_suite_with_dummy_app?(dummy_app_name)
+      Rails.env.test? && Rails.application.class.name == dummy_app_name
+    end
+    private_class_method :running_solidus_test_suite_with_dummy_app?
+
+    def self.does_spree_initializer_exist?(rails_paths, initializer_name)
+      rails_paths['config/initializers'].any? do |path|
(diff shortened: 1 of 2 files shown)


jarednorman (approved):

core/lib/spree/core.rb
```diff
@@ -62,13 +62,24 @@ def self.config(&_block)
...
+        running_solidus_test_suite_with_dummy_app?(dummy_app_name)
     end
 
+    def self.running_solidus_test_suite_with_dummy_app?(dummy_app_name)
+      Rails.env.test? && Rails.application.class.name == dummy_app_name
+    end
+    private_class_method :running_solidus_test_suite_with_dummy_app?
+
+    def self.does_spree_initializer_exist?(rails_paths, initializer_name)
+      rails_paths['config/initializer'].any? do |path|
```
AlistairNorman, line 77:
```suggestion
      rails_paths['config/initializers'].any? do |path|
```

I think `initializers` needs to be pluralized, that's why CI is failing.

AlistairNorman (changes requested):
Thanks for fixing this!

AlistairNorman (approved):

jarednorman (approved):

kennyadsl (approved):
Thanks @waiting-for-dev and @AlistairNorman for reporting and reviewing accurately. 🤝

Issue #2680, opened 2018-04-06
duplicated payments with solidus_gateway gem and Stripe


spaghetticode:
### Issue

I'm using the gem __solidus_gateway__ to integrate Stripe payments, I noticed that, when paying with an existing credit card (a card previously used and already stored in the DB) then **two payments are created**. The last one seems the right one, while the first has amount 0.0 and is marked as invalid. 

I am on Solidus 2.5.0.rc1, I understand the integration with external gems is generally off topics here, especially when such gems use has been explicitly deprecated in the readme, but still I'd like to properly understand what should be the correct payment flow within Solidus in order to fix my issue and hopefully submit a PR on the right codebase. 

### Debugging

During debugging I found out that the Order::Checkout state machine transition
`before_transition to: :payment, do: :add_default_payment_from_wallet` creates the zero amount payment by calling `add_default_payment_from_wallet`, which eventually initializes a new instance of `Spree::Wallet::DefaultPaymentBuilder` that persists the blank payment on the DB. 

After that, while advancing the checkout process OrdersController#update_order instantiates a `OrderUpdateAttributes` object which instantiates a `PaymentCreate` object. At this moment the second payment with correct amount data is created, and this payment is later used for Stripe payment integration. 

The two payments have a few similarities, for example these attributes `"source_type" => "Spree::CreditCard", "source_id" => 2, "payment_method_id" => 1`
where the payment_method is the `#<Spree::Gateway::StripeGateway id: 1` object.

### Questions

Is there any rationale behind having 2 different payments? Can the first dummy payment be eliminated from the checkout flow or should the `PaymentCreate` class update it with new data instead of creating a new one? 

Thanks for your attention,

Andrea

kennyadsl:
Hello @spaghetticode, if I recall correctly the reason behind this implementation is related with how the checkout works:

- on the payment step it makes an authorization payment without capturing any amount, this create the fake payment and store all the information about the payment source (which will allow user to capture the payment later without readding the credit card information)
- on the confirm step it actually makes the payment using the source from the payment previously authorized, this is why that payment is needed.

Despite the process in not intuitive and it can probably be improved, storing the first null payment in terms of logging could also be useful to understand where payment issues are.

Let us know if this makes sense!

spaghetticode:
@kennyadsl thank you for the answer. 

So, you're basically saying that the procedure is right when it creates 2 payments, the first with zero amount that is invalidated when the second, the real one is created. 

[...]

PR #4185, opened 2021-09-29
Remove Spree::Order#add_default_payment_from_wallet checkout transition

danielpuglisi:
This PR is still work in progress.

I'm currently building a shop with solidus and found some logic that seems unnecessary that originated from a 7 year old PR from before the fork: https://github.com/spree/spree/pull/5148

The current behaviour is this:

1. A user with an existing payment source transitions from `delivery` to `payment` state
2. [Spree::Order#add_default_payment_from_wallet](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L686-L698) gets triggered by the [state machine](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/state_machines/order.rb#L85) and creates a payment record.
3. User transitions to `confirm` state
4. Through how the checkout logic works with [Spree::OrderUpdateAttributes](https://github.com/codegestalt/solidus/blob/ec4bdf33325d29ca4243dbccfd699c628b4a9576/core/app/models/spree/order_update_attributes.rb#L35-L39), that default payment does not get reused and is always replaced with a new payment, resulting every time in an invalid payment record:

image

My expectations would be one of the following:

1. The default payment record has some kind of use (so far I haven't found one) and gets updated instead of invalidated (I'm also not sure how this affects fraud detection)
2. No default payment record gets created as it has no use

I started this PR by removing `Spree::Order#add_default_payment_from_wallet` to see what breaks. So far only the spec removed in this [commit](https://github.com/codegestalt/solidus/commit/ec4bdf33325d29ca4243dbccfd699c628b4a9576#diff-c2434636d311e5db5ab103cdd58bd3f44b947d4d384525b6d0c4c475adf1b1ad) have been affected but it seems to me that these are directly tied to the creation of the default payment source and have no further use. The only spec I'm unsure about is the one about assigning a billing address if none exists yet, but I can't think of a scenario where this should be the case.

Another thing the removal of `Spree::Order#add_default_payment_from_wallet` would affect is that it makes [Spree::Wallet::DefaultPaymentBuilder](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/wallet/default_payment_builder.rb) obsolete. So I guess there is a concern about breaking compatibility with existing shops that rely on that one. Although I'm not sure what a use case could be.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -684,6 +684,11 @@ module Spree
     end
 
     def add_default_payment_from_wallet
+      Spree::Deprecation.warn(
+        "`Spree::Order#add_default_payment_from_wallet` transition is deprecated and will be removed." \
+        "Disable it by setting `Spree::Config.disable_adding_default_payment_to_order = true`."
+      )
+
       builder = Spree::Config.default_payment_builder_class.new(self)
 
       if payment = builder.build
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -374,7 +374,12 @@ module Spree
     # @!attribute [rw] default_payment_builder_class
     # @return [Class] a class with the same public interfaces as
     #   Spree::Wallet::DefaultPaymentBuilder.
-    class_name_attribute :default_payment_builder_class, default: 'Spree::Wallet::DefaultPaymentBuilder'
+    attr_writer :default_payment_builder_class
+    def default_payment_builder_class
+      Spree::Deprecation.warn("`Spree::Config.default_payment_builder_class` is deprecated and will be removed.")
+
+      @default_payment_builder_class ||= Spree::Wallet::DefaultPaymentBuilder
+    end
 
     # Allows providing your own class for managing the contents of an order.
     #
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -263,6 +263,11 @@ module Spree
     #   @return [] Track on_hand values for variants / products. (default: true)
     preference :track_inventory_levels, :boolean, default: true
 
+    # @!attribute [rw] disable_adding_default_payment_to_order
+    #   @return [Boolean] Disable Spree::Order#add_default_payment_from_wallet transition (default: +false+)
+    preference :disable_adding_default_payment_to_order, :boolean, default: false
+
+
     # Other configurations
 
     # Allows restricting what currencies will be available.
--- a/core/lib/spree/core/state_machines/order.rb
+++ b/core/lib/spree/core/state_machines/order.rb
@@ -82,7 +82,9 @@ module Spree
                 end
 
                 after_transition to: :complete, do: :add_payment_sources_to_wallet
-                before_transition to: :payment, do: :add_default_payment_from_wallet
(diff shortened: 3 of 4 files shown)


jarednorman:
This is related to #2680.

adammathys (changes requested):
This generally seems reasonable to me. (Removing the `add_default_payment_from_wallet` method.) I can't see any reason why we would ever want that extra invalid payment and we're always going to create a new one during the payment step of checkout.

I think we should probably start by deprecating the method instead of outright removing it. (Along with maybe deprecating the `Spree::Config.default_payment_builder_class` config option and corresponding class.) Just in case there's anyone out there relying on this behaviour for some reason.

Then we could add a new configuration option to skip this callback step for everyone ready to remove this functionality ahead of it being removed in a future release. e.g.:
```ruby
unless Spree::Config.disable_adding_default_payment_to_order
  before_transition to: :payment, do: :add_default_payment_from_wallet
end
```

danielpuglisi:
@adammathys makes sense. If there are no other concerns, I'll gladly put something together.

jarednorman:
Thanks @danielpuglisi (and @adammathys), that would be great!

danielpuglisi:
Finally got the time to take a jab at this.

Added the deprecation preference and warnings based on @adammathys suggestion. The current default preference although creates a lot of noise when running the specs. What is the preferred way to handle this?

waiting-for-dev:
We appreciate your contribution, @danielpuglisi. After talking with @kennyadsl, we decided to put it on hold for the meantime. It looks like a legit change, although it's entirely possible it'd break some extensions. We feel we need to define a clear path for the evolution of the payment system before introducing more changes.

danielpuglisi:
I understand. Let me know if I can be of any help in the future.

kennyadsl:
Hey @danielpuglisi, first of all thanks for the PR and sorry for taking so long to come back on this. We'd love to move this forward because this is annoying on all levels (Admin UX and for the developer). 

We have a couple of things left to do before we can merge this:

1. Rebase the PR against master
2. Change the preference usage to use the new [versioned preferences](https://github.com/solidusio/solidus/blob/9948b01d0e7ec3b77c6aab6421c52adc37be765d/core/lib/spree/preferences/configuration.rb#L129-L151) ([here](https://github.com/solidusio/solidus/blob/3479d43660330be75ba6774dfc1db426a5c9ca13/core/lib/spree/app_configuration.rb#L297) you can find a usage example).
3. run specs of a couple of payment gateway extensions against the local version of solidus that contains this change. I suggest to try with [solidus_stripe](https://github.com/solidusio/solidus_stripe) and [solidus_paypal_braintree](https://github.com/solidusio/solidus_paypal_braintree), which are the most used. 


Please let us know if you have time to take care of these last things otherwise, we will happily take over this work as soon as we have some availability. 

Thanks again! 🙏

danielpuglisi:
Hey @kennyadsl. I'm a bit flooded right now. Feel free to take over if you get the time and I haven't done anything.

jarednorman:
I'm closing this in favour of #6564. @adammathys is working on broadly improving how payments are handled and this approach is incompatible with his, (besides being suuuuper stale). With payment reuse, we don't need to remove `add_default_payment_from_wallet`. The transition will do what it intends to do.

PR #4188, opened 2021-10-04
Let solidus:install use original environment instead of unbundled one

gsmendoza:
Goal
----

TODO

Background
----------

TODO

Reference
---------

From
https://makandracards.com/makandra/15649-how-to-discard-a-surrounding-bundler-environment:

> Use Bundler.with_original_env to restore the environment's state
before Bundler was launched. Do this whenever you want to execute shell
commands inside other bundles.

> If you want a completely Bundler-free environment, you can use
Bundler.with_unbundled_env on modern Bundler (2.1+). If you started with
an environment that already held env variables related to Bundler, this
will restore the original environment with such keys discarded as well.
On older Bundlers, this method was called Bundler.with_clean_env.

**Description**

**Checklist:**

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -181,7 +181,7 @@ module Solidus
         gem plugin_name
       end
 
-      bundle_cleanly{ run "bundle install" } if @plugins_to_be_installed.any?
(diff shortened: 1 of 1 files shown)


gsmendoza:
Closing PR. May not be needed.

PR #4189, opened 2021-10-04
Fix discarded duplicated products bug

Azeem838:
**Description**
When a product is cloned and then discarded without changing the cloned variants SKU, if the product is ever cloned again then the cloning process will fail with a variant and master variant SKU validation error. This issue occurs because products and their variants are soft-deleted and therefore the copied SKU will not be unique during the second cloning.

This change will allow multiple cloning and deleting of cloned products without an SKU validation error. This change uses the count of duplicated records in the SKU variants name. This will allow every record to be unique even if it is deleted.

**Checklist:**

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -66,7 +66,7 @@ module Spree
 
     validates :cost_price, numericality: { greater_than_or_equal_to: 0, allow_nil: true }
     validates :price,      numericality: { greater_than_or_equal_to: 0, allow_nil: true }
-    validates_uniqueness_of :sku, allow_blank: true, case_sensitive: true, if: :enforce_unique_sku?
+    validates_uniqueness_of :sku, allow_blank: true, case_sensitive: true, conditions: -> { where(deleted_at: nil) }, if: :enforce_unique_sku?
 
     after_create :create_stock_items
     after_create :set_position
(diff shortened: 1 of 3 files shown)


jarednorman (commented):
Would a better solution be to scope the uniqueness validation to only non-discarded records? This can also be done at the database level by including the `deleted_at` column in the uniqueness constraint.

Azeem838:
@jarednorman Thanks for the review! Looking at the variant model, there seems to be an `enforce_unique_sku?` condition which was not being applied for this specific case. 

I did not remove the `enforce_unique_sku?` condition since it might be still be used although the tests still pass without it. I need your advice as to whether to remove it or keep it as is.

waiting-for-dev (approved):

waiting-for-dev:
> This can also be done at the database level by including the deleted_at column in the uniqueness constraint.

I guess we can't do that, as we want that behavior to be configurable (i.e., at the application level)

> I did not remove the enforce_unique_sku? condition since it might be still be used although the tests still pass without it. I need your advice as to whether to remove it or keep it as is.

I think we need to keep it, as it was [added](https://github.com/solidusio/solidus/commit/10b2806e3dd7ab48b84cc9bf7406d219855b56b7) to let users override it.

PR #4191, opened 2021-10-14
Issue#3756/calculator should always return a big decimal

redencion:
**Description**

**Checklist:**
- [X ] I have added a detailed description into each commit message
- [X ] I have added tests to cover this change (if needed)


waiting-for-dev:
Hey, @redencion. Thanks for your contribution.

I don't think that's the right place to fix the issue. Callers should obey the requirements of callees instead of callees trying to sanitize their input.

This is a bit old, but do you recall in which situation was that method called with `nil` as an argument?

waiting-for-dev:
We'll close it for now. However, feel free to get back in the future if you want to revisit it. Again, many thanks for your collaboration :heart:

Issue #4175, opened 2021-09-22
Use Variant Searcher for Autocomplete


jarednorman:
When looking into [this issue](https://github.com/solidusio-contrib/solidus_globalize/issues/144) in solidus_globalize, it occurred to me that a possible solution that would I think be an improvement to Solidus and make that issue easy to fix would be to change the variant autocomplete widget to use the variant searcher, rather than [a custom ransack query](https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/variant_autocomplete.js#L35).

Making the variant autocomplete configurable (so that stores that translate the product name could change that query to `product_translations_name_or_sku_cont` wouldn't be excessively difficult, but our variant search class is already configurable and is already more flexible than this custom query. The solidus_globalize gem would just need to provide an initializer that updates [this configuration](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/search/variant.rb#L15-L21) to overcome the bug I reported *if our variant autocomplete code used the variant search class*.

My proposal is this:

1. We add a new API endpoint or augment the existing API variants index action to allow for queries that search using the configured variant search class.
2. We update the variant autocomplete code to use that new endpoint/parameter.
3. In solidus_globalize, we provide some configuration to replace the fields that break when translated with ones that use the translations.

The end result is that we'd be able to resolve the issue on solidus_globalize and that the variant autocomplete fields in the admin would be a little nicer out of the box.

PR #4197, opened 2021-10-27
Use Variant Searcher for Autocomplete

adammathys:
**Description**

Update `/api/variants` to be able to search either using the configurable `variant_search_class` or Ransack. This makes it possible for users to more easily override the search behaviour of `variantAutocomplete` based on their needs. (e.g.: If they're using Globalize and need to search the translated fields instead of just the base columns that won't be in use.)

The existing behaviour is maintained to ensure the API is completely backwards compatible.

Ref #4175

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/variants_controller.rb
+++ b/api/app/controllers/spree/api/variants_controller.rb
@@ -25,8 +25,14 @@ module Spree
       # we render on the view so we better update it any time a node is included
       # or removed from the views.
       def index
-        @variants = scope.includes(include_list)
-          .ransack(params[:q]).result
+        @variants =
+          if params[:variant_search_term]
+            Spree::Config.variant_search_class.new(
+              params[:variant_search_term], scope: scope
+            ).results.includes(include_list)
+          else
+            scope.includes(include_list).ransack(params[:q]).result
+          end
 
         @variants = paginate(@variants)
         respond_with(@variants)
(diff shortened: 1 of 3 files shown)


jarednorman (approved):
I'm personally fine with merging this in a minor release since it preserves the old behaviour even if you have weird overrides.

waiting-for-dev:
Excellent idea, but I'm not entirely sure about the implementation. What about creating another `variant_search_class` delegating to ransack and letting users choose which one to use in the API controller? It should default to the ransack one to keep current behavior.

adammathys:
@waiting-for-dev I'm not sure I follow. The default [`variant_search_class`](https://github.com/solidusio/solidus/blob/master/core/lib/spree/core/search/variant.rb) already uses Ransack under the hood. The difference between this change and existing behaviour is that we allow passing in a single search term and letting the `variant_search_class` decide how to craft the appropriate Ransack parameters to use for that, instead of having those parameters hardcoded in JavaScript. (Which makes them impossible to easily modify.)

waiting-for-dev:
@adammathys, sorry, I overlooked that the current `variant_search_class` is already delegating to ransack.

It's nice the extra flexibility that this PR adds to the variant autocomplete code for JS. And it makes sense to have the API controller perform both fuzzy and conventional search (as the backend needs the fuzzy one while the API request needs the traditional one).

My comment was more about creating a second `variant_search_class` for the traditional type of search, so just extracting the ransack forwarding without further processing. So, I'd have two types of configurable variant classes: `variant_search_class` and `variant_autocomplete_search_class`, and would apply them not only to the backend as it's done now but also in the API index action.

However, that would broaden this PR's scope as we're discussing a more concrete need. So, I'm ok with merging this straightaway.

waiting-for-dev (approved):

Issue #4004, opened 2021-03-22
AddPaymentSourcesToWallet "randomly" changing the default


adammathys:
As of the introduction of #2913, we no longer set the default payment to one that was used on the current order. Instead, we implicitly pick the one with the biggest ID from all sources in a user's wallet. This can result in inconsistent behaviour where placing an order will change the default payment from the one the customer used and/or the one a customer has explicitly set as their default.

**Solidus Version**:
v2.11+

**To Reproduce**
1. Add two payment sources to a users wallet. Make the one with the smallest ID the default payment source.
2. Complete checkout using the default payment source.

**Current behavior**
After placing an order with the default payment source, the default will be changed to the wallet source with the largest ID. (Because `order.user.wallet_payment_sources.last` should implicitly order by ID.) Meaning we changed the user's default payment source without them ever selecting the new default.

**Expected behavior**
Based on previous behaviour, default payment source should be unchanged after placing the order.

jarednorman:
Am I misunderstanding that the _previous_ behaviour was for the reusable source _on the order_ with the highest ID would previously have been selected as the default. That's how I'm reading [this](https://github.com/solidusio/solidus/pull/2913/files#diff-7fd6876e363aa5cb0d758438a2cf1523d888d90396ad36b1e25931855bb03358L31-L32).

adammathys:
Correct, the previous behaviour would be to pick the source with the largest `Spree::Source#id` that was part of a valid payment on that order. 

The new logic instead picks the wallet source with the largest `Spree::WalletPaymentSource#id` from all of the sources in the wallet.

PR #4198, opened 2021-11-01
Fix AddPaymentSourcesToWallet changing default when reused

RyanofWoods:
Fixes issue #4004 [1]

PR #2913 [2] changed how the new default WalletPaymentSource was set,
from using the last one found/created * within the Order's
PaymentSources to the user's last WalletPaymentSource. A subtle
difference which assumes that the user's last WalletPaymentSource was
just created by add_to_wallet.

However, if the Order's PaymentSource was from the User's wallet default
and add_to_wallet is called, a new WalletPaymentSource would not be
created * and make_default would try to set the new default as the last
WalletPaymentSource. This would change the default if the last
WalletPaymentSource was not the default.

Now, the WalletPaymentSources are scoped to the order's PaymentSources
and the last one is given to Wallet#default_wallet_payment_source=.
Meaning that in this scenario, the default would be given and ignored as it
is already the default.

[1] https://github.com/solidusio/solidus/issues/4004
[2] https://github.com/solidusio/solidus/pull/2913/
* Wallet#add creates or finds the WalletPaymentSource if exists.

**Checklist:**

### DIFF
--- a/core/app/models/spree/wallet/add_payment_sources_to_wallet.rb
+++ b/core/app/models/spree/wallet/add_payment_sources_to_wallet.rb
@@ -24,19 +24,19 @@ class Spree::Wallet::AddPaymentSourcesToWallet
       # add valid sources to wallet and optionally set a default
       if sources.any?
         # arbitrarily sort by id for picking a default
-        sources.sort_by(&:id).each do |source|
+        order_wallet_payment_sources = sources.sort_by(&:id).map do |source|
           order.user.wallet.add(source)
         end
 
-        make_default
+        make_default(order_wallet_payment_sources)
       end
     end
   end
 
   protected
 
-  def make_default
-    order.user.wallet.default_wallet_payment_source = order.user.wallet_payment_sources.last
+  def make_default(sources)
+    order.user.wallet.default_wallet_payment_source = sources.last
   end
 
   private
--- a/core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
+++ b/core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
@@ -4,14 +4,41 @@ require 'rails_helper'
 
 RSpec.describe Spree::Wallet::AddPaymentSourcesToWallet, type: :model do
   let(:order) { create(:order_ready_to_complete) }
+  let(:user) { order.user }
 
   describe '#add_to_wallet' do
     subject { described_class.new(order) }
+
+    before do
+      # Ensure the payment is reusable, otherwise false positives occur
+      allow_any_instance_of(Spree::CreditCard).to receive(:reusable?).and_return(true)
+    end
 
     it 'saves the payment source' do
       expect { subject.add_to_wallet }.to change {
         order.user.wallet.wallet_payment_sources.count
       }.by(1)
+    end
+
+    context "when the default wallet payment source is used and a more recent wallet payment exists" do
+      before do
+        credit_card_one = user.wallet.add(create(:credit_card, user: user))
+        user.wallet.add(create(:credit_card, user: user))
+        user.wallet.default_wallet_payment_source = credit_card_one # must be the first created card
+        order.payments.first.update!(source: credit_card_one.payment_source)
+      end
+
+      it 'does not make a new wallet payment source' do
+        expect { subject.add_to_wallet }.to_not change {
+          order.user.wallet.wallet_payment_sources.count
+        }
+      end
+
+      it 'does not change the default wallet payment source' do
+        expect { subject.add_to_wallet }.to_not change {
+          user.wallet.default_wallet_payment_source
+        }
+      end
     end
   end
 end


waiting-for-dev:
I might be a bit lost here. But isn't it effectively reverting #2913? Wouldn't #2652 still be an issue in this case?

I haven't studied it closely enough, but I guess that ideally, `#add_to_wallet` shouldn't try to set the default unless:
- Only one source exists
- The user checks a "set default" checkbox (so we'd need to push that data from the controller layer down to this method).

RyanofWoods:
> I might be a bit lost here. But isn't it effectively reverting #2913? Wouldn't #2652 still be an issue in this case?
> 
> I haven't studied it closely enough, but I guess that ideally, `#add_to_wallet` shouldn't try to set the default unless:
> 
> * Only one source exists
> * The user checks a "set default" checkbox (so we'd need to push that data from the controller layer down to this method).

Thank you for raising this to my attention. Thinking about it more, I think you are correct. 👍 `#add_to_wallet` shouldn't set the default unless the user gives input to do so. If I change the PR to do this, then it can be to resolve https://github.com/solidusio/solidus/issues/2652 and would automatically resolve https://github.com/solidusio/solidus/issues/4004.

I think perhaps it would be more appropriate for `Spree::Wallet` to set default the `wallet_payment_source` if it is the only one. Then it can be done as callbacks for adding or removing a `wallet_payment_source`, what do you think?

waiting-for-dev:
> I think perhaps it would be more appropriate for Spree::Wallet to set default the wallet_payment_source if it is the only one. Then it can be done as callbacks for adding or removing a wallet_payment_source, what do you think?

Do you mean Active Record callbacks? I'd strongly recommend against using them. Coupling business logic with persistence is an endless source of problems. My feeling is that we're trying to move away from them in Solidus.

I'm not very familiar with that part of the system, so at first sight, I'd leave that piece of logic there if there's nowhere more appropriate. About making that method to know if a user has checked something, I have no idea how hard it'll be :slightly_smiling_face: I'm happy to help here if it's needed.

adammathys (commented):
I agree with @waiting-for-dev on all points here. Ideally we're only setting a default source if there isn't one or if the user prompts to have it changed. Also, I think the logic for setting the initial default fits fairly well within the `AddPaymentSourcesToWallet` class instead of trying to add it to callbacks. e.g.:

```ruby
unless order.user.wallet.default_wallet_payment_source
  order.user.wallet.default_wallet_payment_source = new_default_source
end
```

That being said, I don't think this PR as is reverts #2913. The main purpose of that change was to make it easier to override the `make_default` behaviour so that users could decide to opt out of the default behaviour of having the most recently used credit card becoming the default. (I think the switch to have it be "most recently created" instead of "most recently used" was an unintended side-effect of that change.)

waiting-for-dev:
It makes sense. Another point to consider, as @kennyadsl commented offline, is payments made offsite, where users can't select a default payment method. We definitely should give more thought and study a suitable architecture for a broad solution going beyond the scope of this PR.

core/app/models/spree/wallet/add_payment_sources_to_wallet.rb
```diff
@@ -24,7 +24,7 @@ def add_to_wallet
       # add valid sources to wallet and optionally set a default
       if sources.any?
         # arbitrarily sort by id for picking a default
-        sources.sort_by(&:id).each do |source|
+        @order_wallet_payment_sources = sources.sort_by(&:id).map do |source|
```
waiting-for-dev, line 27:
```suggestion
        order_wallet_payment_sources = sources.sort_by(&:id).map do |source|
```

core/app/models/spree/wallet/add_payment_sources_to_wallet.rb
```diff
@@ -36,7 +36,7 @@ def add_to_wallet
   protected
 
   def make_default
```
waiting-for-dev, line 38:
```suggestion
  def make_default(sources)
```

core/app/models/spree/wallet/add_payment_sources_to_wallet.rb
```diff
@@ -36,7 +36,7 @@ def add_to_wallet
...
-    order.user.wallet.default_wallet_payment_source = order.user.wallet_payment_sources.last
+    order.user.wallet.default_wallet_payment_source = @order_wallet_payment_sources.last
```
waiting-for-dev, line 39:
```suggestion
    order.user.wallet.default_wallet_payment_source = sources.last
```

waiting-for-dev (approved):
I think we can merge it. It's not the complete solution, but at least it's fixing the issue introduced in #2913.

@RyanofWoods, I left some suggestions that, when combined, get rid of the unneeded global state and make use of a local variable instead.

core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
```diff
@@ -13,5 +15,18 @@
...
+
+    it 'keeps the payment source as default when used' do
+      user = order.user
+      user.wallet.add(create(:credit_card, user: user))
+      user.wallet.add(create(:credit_card, user: user))
+      user.wallet.default_wallet_payment_source = user.wallet_payment_sources.first
+      order.payments.first.update!(source: user.wallet.default_wallet_payment_source.payment_source)
+      allow_any_instance_of(Spree::CreditCard).to receive(:reusable?).and_return(true)
+
+      expect { subject.add_to_wallet }.to not_change {
```
kennyadsl, line 27: I think you can just use `.to_not change` here, without defining the custom matcher on line 5.

kennyadsl (commented):
@RyanofWoods hey, thanks a lot for this. I think it's close to something we can merge if you can address comments made by @waiting-for-dev and myself. 🙏

RyanofWoods:
I don't see how my change is going to decrease coverage by 15%? 😅

core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
```diff
@@ -13,5 +14,27 @@
...
       }.by(1)
     end
+
+    context "when the default wallet payment source is used and more a recent wallet payment exists" do
+      before do
+        credit_card_one = user.wallet.add(create(:credit_card, user: user))
+        credit_cart_two = user.wallet.add(create(:credit_card, user: user))
+        user.wallet.default_wallet_payment_source = credit_card_one # must be the first created card
+        order.payments.first.update!(source: credit_card_one.payment_source)
+        flunk unless credit_card_one.payment_source.reusable?
```
waiting-for-dev, line 24: Can't we make sure that the payment_source is reusable in some other way?

waiting-for-dev (approved):
Thanks, @RyanofWoods! I left a question about the test, but it looks good otherwise. I rerun the build, and the issue about the coverage is gone 🙂

core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
RyanofWoods: Hmm, for context, the reason I added this is because if the credit card was not reusable, the spec could become a false positive. Now, instead, I added `expect(user.wallet).to receive(:add).and_call_original` to the specs, WDYT?

waiting-for-dev: I see. My point is, is there a way to ensure that through the factory itself?

RyanofWoods: Not directly:
https://github.com/solidusio/solidus/blob/a5c2e41e22441cf3fe648c21bf438c6d92c4130d/core/app/models/spree/credit_card.rb#L123-L131

Seeing as `reusable?` is already returning `true`. I think it dilutes the spec a lot of we add unnecessary setup to make sure `reusable?` is explicitly `true`. Is `expect(user.wallet).to receive(:add).and_call_original` not a suitable alternative to ensure we're not getting false positives?

core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
```diff
@@ -13,5 +14,28 @@
...
+        expect(user.wallet).to receive(:add).and_call_original
+        expect { subject.add_to_wallet }.to_not change {
+          order.user.wallet.wallet_payment_sources.count
+        }
+      end
+
+      it 'does change the default wallet payment source' do
+        expect(user.wallet).to receive(:add).and_call_original
+        expect { subject.add_to_wallet }.to_not change {
+          user.wallet.default_wallet_payment_source
```
kennyadsl, line 36: The spec description says `does change the default wallet payment source` but seems like it's testing the opposite thing:  `...to_not change { user.wallet.default_wallet_payment_source }`. Am I missing something or something needs to be switched here?

RyanofWoods: Whoops! 😅 Typo. Good catch, thanks!

core/spec/models/spree/wallet/add_payment_sources_to_wallet_spec.rb
```diff
@@ -13,5 +14,28 @@
         order.user.wallet.wallet_payment_sources.count
       }.by(1)
     end
+
+    context "when the default wallet payment source is used and more a recent wallet payment exists" do
```
kennyadsl, line 18:
```suggestion
    context "when the default wallet payment source is used and a more recent wallet payment exists" do
```

kennyadsl (commented):
Found a typo in the test context. Also, there's a hanging discussion with @waiting-for-dev, is that solved?

RyanofWoods:
Resolved both points, @waiting-for-dev and @kennyadsl, thanks!

kennyadsl (approved):
Thanks @RyanofWoods!

PR #4201, opened 2021-11-07
Add configuration flag deprecated associated user

dborovsky:
**Description**
task - #3559 

**Checklist:**

### DIFF
--- a/core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
+++ b/core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
@@ -40,6 +40,8 @@ Spree.config do |config|
   # Template to use when rendering layout
   # config.layout = "spree/layouts/spree_application"
 
+  # manually call associate_user instead of relying on current_order
+  # config.associate_user_in_authentication_extension = false
 
   # Admin:
 
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -169,6 +169,10 @@ module Spree
     #   @return [String] Email address used as +From:+ field in transactional emails.
     preference :mails_from, :string, default: 'solidus@example.com'
 
+    # @!attribute [rw] associate_user_in_authentication_extension
+    #   @return [Boolean] manually call associate_user instead of relying on current_order
+    preference :associate_user_in_authentication_extension, :boolean, default: false
+
     # @!attribute [rw] max_level_in_taxons_menu
     #   @return [Integer] maximum nesting level in taxons menu (default: +1+)
     preference :max_level_in_taxons_menu, :integer, default: 1
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -263,7 +263,6 @@ module Spree
     #   @return [] Track on_hand values for variants / products. (default: true)
     preference :track_inventory_levels, :boolean, default: true
 
-
     # Other configurations
 
     # Allows restricting what currencies will be available.
@@ -524,6 +523,17 @@ module Spree
     # @return [Module] a module that can be included into Spree::Taxon to allow attachments
     # Enumerable of taxons adhering to the present_taxon_class interface
     class_name_attribute :taxon_attachment_module, default: 'Spree::Taxon::ActiveStorageAttachment'
+
+    # Configures the absolute path that contains the Solidus engine
+    # migrations. This will be checked at app boot to confirm that all Solidus
+    # migrations are installed.
+    #
+    # @!attribute [rw] migration_path
+    # @return [Pathname] the configured path. (default: `Rails.root.join('db', 'migrate')`)
+    attr_writer :migration_path
+    def migration_path
+      @migration_path ||= ::Rails.root.join('db', 'migrate')
+    end
 
     # Allows providing your own class instance for generating order numbers.
     #
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -171,7 +171,7 @@ module Spree
 
     # @!attribute [rw] associate_user_in_authentication_extension
     #   @return [Boolean] manually call associate_user instead of relying on current_order
-    preference :associate_user_in_authentication_extension, :boolean, default: false
+    preference :associate_user_in_authentication_extension, :boolean, default: by_version(false, "3.2.0.alpha" => true)
 
     # @!attribute [rw] max_level_in_taxons_menu
     #   @return [Integer] maximum nesting level in taxons menu (default: +1+)
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -171,7 +171,7 @@ module Spree
 
     # @!attribute [rw] associate_user_in_authentication_extension
     #   @return [Boolean] manually call associate_user instead of relying on current_order
-    preference :associate_user_in_authentication_extension, :boolean, default: by_version(false, "3.2.0.alpha" => true)
+    preference :associate_user_in_authentication_extension, :boolean, default: by_version(false, "4.0.0" => true)
 
     # @!attribute [rw] max_level_in_taxons_menu
     #   @return [Integer] maximum nesting level in taxons menu (default: +1+)
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -171,7 +171,7 @@ module Spree
 
     # @!attribute [rw] associate_user_in_authentication_extension
     #   @return [Boolean] manually call associate_user instead of relying on current_order
-    preference :associate_user_in_authentication_extension, :boolean, default: by_version(false, "4.0.0" => true)
+    preference :associate_user_in_authentication_extension, :boolean, default: by_version(false, "3.2.0.alpha" => true)
 
     # @!attribute [rw] max_level_in_taxons_menu
     #   @return [Integer] maximum nesting level in taxons menu (default: +1+)
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -38,9 +38,8 @@ module Spree
 
         def associate_user
           return if Spree::Config.associate_user_in_authentication_extension
-          Spree::Deprecation.behavior = :stderr
           Spree::Deprecation.warn(
-            "Order#associate_user is deprecated and will be removed in 3.2.0.alpha",
+            "Order#associate_user is deprecated and will be removed in 4.0.0",
             caller(1)
           )
           @order ||= current_order
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -38,7 +38,11 @@ module Spree
 
         def associate_user
           return if Spree::Config.associate_user_in_authentication_extension
-          Spree::Deprecation.warn "Order#associate_user is deprecated and will be removed in 2.11"
+          Spree::Deprecation.behavior = :stderr
+          Spree::Deprecation.warn(
+            "Order#associate_user is deprecated and will be removed in 2.11",
+            caller(1)
+          )
           @order ||= current_order
           if try_spree_current_user && @order
             @order.associate_user!(try_spree_current_user) if @order.user.blank? || @order.email.blank?
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -37,6 +37,8 @@ module Spree
         end
 
         def associate_user
+          return if Spree::Config.associate_user_in_authentication_extension
+          Spree::Deprecation.warn "Order#associate_user is deprecated and will be removed in 2.11"
           @order ||= current_order
           if try_spree_current_user && @order
             @order.associate_user!(try_spree_current_user) if @order.user.blank? || @order.email.blank?
--- a/core/lib/spree/core/controller_helpers/order.rb
+++ b/core/lib/spree/core/controller_helpers/order.rb
@@ -40,7 +40,7 @@ module Spree
           return if Spree::Config.associate_user_in_authentication_extension
           Spree::Deprecation.behavior = :stderr
           Spree::Deprecation.warn(
-            "Order#associate_user is deprecated and will be removed in 2.11",
+            "Order#associate_user is deprecated and will be removed in 3.2.0.alpha",
             caller(1)
           )
           @order ||= current_order
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -134,6 +134,7 @@ Spree.user_class = 'Spree::LegacyUser'
 Spree.load_defaults(Spree.solidus_version)
 Spree.config do |config|
   config.mails_from = "store@example.com"
+  config.associate_user_in_authentication_extension = false
   # TODO: Remove on Solidus 4.0 as it'll be the default
   require 'spree/event/adapters/deprecation_handler'
   if Spree::Event::Adapters::DeprecationHandler.legacy_adapter_set_by_env.nil?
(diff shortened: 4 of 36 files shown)


core/lib/spree/app_configuration.rb
```diff
@@ -169,6 +169,10 @@ class AppConfiguration < Preferences::Configuration
     #   @return [String] Email address used as +From:+ field in transactional emails.
     preference :mails_from, :string, default: 'solidus@example.com'
 
+    # @!attribute [rw] associate_user_in_authentication_extension
+    #   @return [Boolean] manually call associate_user instead of relying on current_order
+    preference :associate_user_in_authentication_extension, :boolean, default: false
```
waiting-for-dev, line 174: We can leverage [the new system for defaults by version](https://github.com/solidusio/solidus/blob/master/CHANGELOG.md#major-changes):

```ruby
preference :associate_user_in_authentication_extension, :boolean, default: by_version(false, "3.2.0.alpha" => true)
```

core/lib/spree/core/controller_helpers/order.rb
```diff
@@ -37,6 +37,12 @@ def current_order(options = {})
         end
 
         def associate_user
+          return if Spree::Config.associate_user_in_authentication_extension
+          Spree::Deprecation.behavior = :stderr
```
waiting-for-dev, line 41: Do we want to override whatever preferences users have?

core/lib/spree/core/controller_helpers/order.rb
```diff
@@ -37,6 +37,12 @@ def current_order(options = {})
...
+          Spree::Deprecation.warn(
+            "Order#associate_user is deprecated and will be removed in 2.11",
```
waiting-for-dev, line 43: I think there's an error in the version number, right?

core/lib/spree/testing_support/dummy_app.rb
```diff
@@ -134,6 +134,7 @@ class Application < ::Rails::Application
 Spree.load_defaults(Spree.solidus_version)
 Spree.config do |config|
   config.mails_from = "store@example.com"
+  config.associate_user_in_authentication_extension = false
```
waiting-for-dev, line 137: I guess we can remove this line. If we use the defaults by version, it'll use `true` on the current version so no warning will be emitted.

core/spec/spec_helper.rb
```diff
@@ -11,6 +11,7 @@
 require 'spree/testing_support/preferences'
 require 'spree/config'
 require 'with_model'
+require 'pry'
```
waiting-for-dev, line 14: I guess we should remove this line.

frontend/spec/spec_helper.rb
```diff
@@ -38,6 +38,7 @@
 require 'spree/testing_support/caching'
 require 'spree/testing_support/blacklist_urls'
 require 'spree/testing_support/translations'
+require 'pry'
```
waiting-for-dev, line 41: Same.

core/lib/spree/core/controller_helpers/order.rb
dborovsky: @waiting-for-dev  by default Spree::Deprecation has behavior :warn and return ActiveSupport::DeprecationException  and integration tests capybara are fails. By this reason I've overridden behavior to stderr.

waiting-for-dev: I see your point. However, as we're using `#by_version` in the configuration class now, the master branch will use `true` (as its version is `3.2.0.alpha`), and there's no actual need for it anymore.

core/lib/spree/core/controller_helpers/order.rb
```diff
@@ -37,6 +37,12 @@ def current_order(options = {})
         end
 
         def associate_user
+          return if Spree::Config.associate_user_in_authentication_extension
+          Spree::Deprecation.behavior = :stderr
+          Spree::Deprecation.warn(
+            "Order#associate_user is deprecated and will be removed in 3.2.0.alpha",
```
waiting-for-dev, line 43: The deprecation horizon should be beyond `3.2`, as that version will be the one adding the new option. I think our policy is not breaking code until a major release, so it should be `4.0.0`. cc @kennyadsl

kennyadsl: You are right @waiting-for-dev, the method will be removed in the next major (`4.0.0`).

dborovsky:
@waiting-for-dev @kennyadsl. check please my last commit. I removed also `Spree::Deprecation.behavior = :stderr`

 In my local environment  by default behavior is :stedrr and tests are true.  But circle/ci return error. I see deprecation return ActiveSupport::DeprecationException. How I understood correct it is mean behavior assigned as `:raise`.

core/lib/spree/app_configuration.rb
```diff
@@ -169,6 +169,10 @@ class AppConfiguration < Preferences::Configuration
     #   @return [String] Email address used as +From:+ field in transactional emails.
     preference :mails_from, :string, default: 'solidus@example.com'
 
+    # @!attribute [rw] associate_user_in_authentication_extension
+    #   @return [Boolean] manually call associate_user instead of relying on current_order
+    preference :associate_user_in_authentication_extension, :boolean, default: by_version(false, "4.0.0" => true)
```
waiting-for-dev, line 174:
```suggestion
    preference :associate_user_in_authentication_extension, :boolean, default: by_version(false, "3.2.0.alpha" => true)
```

We're going to remove the setting in v4.0 completely. We want it to be `false` on 3.1, but `true` on master (v3.2.0.alpha) and onward (v3.2.0). When users update v3.1 to v3.2 with `bin/rails g solidus:update`, they will keep having it to `false` as it will be overridden by the generated initializer `new_solidus_defaults.rb`. Then, they'll adapt their app (if needed) and remove the override. Once done, they'll be using `true` (default in v3.2), and their app will be ready for the future update to v4.0, where we'll completely remove all references to `:associate_user_in_authentication_extension`.

Once done this change, no error should be raised on the CI.

core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
```diff
@@ -40,6 +40,8 @@ Spree.config do |config|
   # Template to use when rendering layout
   # config.layout = "spree/layouts/spree_application"
 
+  # manually call associate_user instead of relying on current_order
+  # config.associate_user_in_authentication_extension = false
```
waiting-for-dev, line 44: As we're going to remove the setting in v4.0 altogether, I think it's better not to add it in the generated `spree.rb` for fresh installations. They should be using the new default, and there's no need for them to know anything about the old behavior that will be removed. Thoughts @kennyadsl ?

waiting-for-dev:
Thanks for your continued work on this, @dborovsky! I tried to explain better the deprecation path. I hope it's clearer now. I'm also pinging @kennyadsl for a minor question.

core/spec/lib/spree/core/controller_helpers/order_spec.rb
```diff
@@ -84,6 +84,19 @@
         controller.associate_user
       end
     end
+
+    context "configuration associate_user_in_authentication_extension is false" do
+      before do
+        stub_spree_preferences(associate_user_in_authentication_extension: false)
+      end
```
forkata, line 91: I think we should move the above specs into this context since they will start failing once we update the version and also they are only relevant in the case where the value of this setting is `false`.

We could also add a test for the new behaviour which should not associate the user nor show a deprecation warning.

waiting-for-dev: > We could also add a test for the new behaviour which should not associate the user nor show a deprecation warning.

Agree

> I think we should move the above specs into this context since they will start failing once we update the version and also they are only relevant in the case where the value of this setting is false.

What I've been doing is still to test that the deprecation warning is shown. As it's useful for users, I like to have it covered. We'd remove it alongside the setting on `v4.0`. But, in this case, mine is not a strong opinion, so I'd be good with any outcome.

dborovsky:
@forkata  @waiting-for-dev  Check please my last commit.  Inside the method `associate_user` was called a method `order.associate_user! `  and tests are fails.

dborovsky:
> @forkata @waiting-for-dev Check please my last commit. Inside the method `associate_user` was called a method `order.associate_user! ` and tests are fails.

dborovsky:
@forkata @waiting-for-dev  check please my solution, could you tell check why tests are fails in CI/circleci.

core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
waiting-for-dev: I still would like some thought about it. @kennyadsl, what are we usually doing in these cases?

frontend/spec/controllers/spree/checkout_controller_spec.rb
```diff
@@ -54,6 +54,7 @@
         # The first step for checkout controller is address
         # Transitioning into this state first is required
         order.update_column(:state, "address")
+        stub_spree_preferences(associate_user_in_authentication_extension: false)
```
waiting-for-dev, line 57: How is it that we need to override the new setting value here? Wasn't the method useless?

waiting-for-dev:
> @forkata @waiting-for-dev check please my solution, could you check the tests which are failed in CI/circlec

You need to rebase from master because of https://github.com/solidusio/solidus/pull/4209

dborovsky:
@forkata  @waiting-for-dev  If the method `associate_user `won't be used than in `checkout _controller` method `authorize!(:edit, current_order, cookies.signed[:guest_token])` return
`CanCan::AccessDenied: You are not authorized to access this page.`

https://github.com/solidusio/solidus/blob/9dbcad1178c484edab0a4c2de554ce968107d38a/frontend/app/controllers/spree/checkout_controller.rb#L223

core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt
kennyadsl: I agree with @waiting-for-dev, I'd remove this as well.

core/lib/spree/core/controller_helpers/order.rb
```diff
@@ -37,6 +37,11 @@ def current_order(options = {})
         end
 
         def associate_user
+          return if Spree::Config.associate_user_in_authentication_extension
```
kennyadsl, line 40: I think we should just stop calling this method from inside our core codebase instead of returning if the preference is true.

I can see at least a couple of occurrences:

- https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/frontend/app/controllers/spree/orders_controller.rb#L43
- https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/frontend/app/controllers/spree/checkout_controller.rb#L18


If it's useless, things should keep working as before.

kennyadsl (changes requested):
Thanks for taking care of this. I left a comment on the approach, which I think it's not the right one. Let me know!

core/lib/spree/core/controller_helpers/order.rb
forkata: I think adding a conditional on the value of the preference in multiple places may be slightly worse than just checking it in the method itself.

frontend/spec/controllers/spree/checkout_controller_spec.rb
```diff
@@ -56,10 +56,28 @@
...
+          get :edit
+        end
+      end
+
+      context "new behavior" do
+        before do
+          stub_spree_preferences(associate_user_in_authentication_extension: true)
+        end
+
+        it "should associate the order with a user" do
```
forkata, line 76: This should probably say `"should not associate the order with a user"`.

frontend/spec/controllers/spree/checkout_controller_spec.rb
```diff
@@ -56,10 +56,28 @@
         order.update_column(:state, "address")
       end
 
-      it "should associate the order with a user" do
-        order.update_column :user_id, nil
-        expect(order).to receive(:associate_user!).with(user)
-        get :edit
+      context "current behavior" do
```
forkata, line 59: It might be better if we say `"with associate_user_in_authentication_extension enabled"` as this behaviour will change with the version bump.

frontend/spec/controllers/spree/checkout_controller_spec.rb
```diff
@@ -56,10 +56,28 @@
...
+        end
+
+        it "should associate the order with a user" do
+          order.update_column :user_id, nil
+          expect(order).to receive(:associate_user!).with(user)
+          get :edit
+        end
+      end
+
+      context "new behavior" do
```
forkata, line 71: Similar to the above comment, this may be better as `"with associate_user_in_authentication_extension disabled"`.

core/spec/lib/spree/core/controller_helpers/order_spec.rb
```diff
@@ -71,16 +71,31 @@
...
+      context "user's email is blank" do
+        let(:user) { create(:user, email: '') }
+        it 'calls Spree::Order#associate_user! method' do
+          expect_any_instance_of(Spree::Order).to receive(:associate_user!)
+          controller.associate_user
+        end
+      end
+
+      context "user isn't blank" do
+        it 'does not calls Spree::Order#associate_user! method' do
```
forkata, line 89: This should be `"does not call Spree::Order#associate_user! method"`.

core/lib/spree/core/controller_helpers/order.rb
jarednorman: @forkata I believe @kennyadsl is suggesting not having an option at all. We can simply deprecate it and stop using it in core, then later remove it entirely.

forkata: That makes, sense, I misread the comment here 😅

dborovsky: @waiting-for-dev @forkata @kennyadsl Please read my comment above https://github.com/solidusio/solidus/pull/4201#issuecomment-982833893

kennyadsl:
> @forkata @waiting-for-dev If the method associate_user won't be used in checkout _controller than method authorize!(:edit, current_order, cookies.signed[:guest_token]) return `CanCan::AccessDenied: You are not authorized to access this page`.

Did you get why this is happening? Can you please explain where exactly and in which circumstances?

dborovsky:
> > @forkata @waiting-for-dev If the method associate_user won't be used in checkout _controller than method authorize!(:edit, current_order, cookies.signed[:guest_token]) return `CanCan::AccessDenied: You are not authorized to access this page`.
> 
> Did you get why this is happening? Can you please explain where exactly and in which circumstances?

@kennyadsl 

It happens here:  https://github.com/solidusio/solidus/blob/9dbcad1178c484edab0a4c2de554ce968107d38a/frontend/app/controllers/spree/checkout_controller.rb#L223


Because `current_order.user_id is nil`.

kennyadsl:
I guess we need to keep `associate_user` then. We can't deprecate a method that we are using and we need to make things work. Or am I missing something?

dborovsky:
> I guess we need to keep `associate_user` then. We can't deprecate a method that we are using and we need to make things work. Or am I missing something?

dborovsky:
Yes. The method `associate_user` called the method  `@order.associate_user!(try_spree_current_user) `. This method set user_id to current order.

dborovsky:
@forkata  @waiting-for-dev  @kennyadsl  any news?

waiting-for-dev:
I've been reviewing it, and I think we can close it. Thanks for your work, @dborovsky, and for discovering that the method is still helpful for the checkout controller. We could still remove the call from the orders controller. However, given that we're going to deprecate the current frontend in favor of `solidus_starter_frontend`, probably it's better to fix it there. Feel free to get back to us if you think it's the wrong decision or in case you still want to work on the removal from the orders controller.

dborovsky:
Hi Marc. (@waiting-for-dev ) I agree. I would like to help contributing solidus.  Tell me which tasks are actual and in priority now.

waiting-for-dev:
Hey, @dborovsky, that would be amazing! Would contact me through DM on [Slack](http://slack.solidus.io/) work for you? My nickname is @waiting_for_dev. Let me know if you prefer any other channel.

PR #4203, opened 2021-11-16
Adds observability to the new event bus

waiting-for-dev:
This commit adds observability abilities to the new event bus adapter:

- It adds creation time and caller location to the `Spree::Event::Event`
  instance bound to the subscribers.
- It wraps the execution of listeners within a new
  `Spree::Event::Execution` class, which wraps the execution time, a
  benchmark of the operation, and the result returned by the
  subscription.

Both the `Spree::Event::Event` and the list of `Spree::Event::Execution`
are wrapped within a new `Spree::Event::Firing` object, which is
returned on `Spree::Event.fire`. Therefore, a typical flow could be
getting the event's caller from within the subscriber block and then
inspecting the firing object from there if needed.

Having all the metadata collected in a single place when we do
`Spree::Event.fire` will simplify things if we implement some event
store in the future.

**Checklist:**

### DIFF
--- a/core/lib/spree/event/firing.rb
+++ b/core/lib/spree/event/firing.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module Spree
+  module Event
+    # The result of firing an event
+    #
+    # It encapsulates a fired {Spree::Event::Event} as well as the
+    # {Spree::Event::Execution}s it originated.
+    #
+    # This class is useful mainly for debugging and logging purposes. An
+    # instance of it is returned on {Spree::Event.fire}.
+    class Firing
+      # Fired event
+      #
+      # @return [Spree::Event::Event]
+      attr_reader :event
+
+      # Listener executions that the firing originated
+      #
+      # @return [Array<Spree::Event::Execution>]
+      attr_reader :executions
+
+      # @api private
+      def initialize(event:, executions:)
+        @event = event
+        @executions = executions
+      end
+    end
+  end
+end
--- a/core/spec/lib/spree/event/adapters/default_spec.rb
+++ b/core/spec/lib/spree/event/adapters/default_spec.rb
@@ -7,85 +7,59 @@ module Spree
   module Event
     module Adapters
       RSpec.describe Default do
+        let(:counter) do
+          Class.new do
+            attr_reader :count
+
+            def initialize
+              @count = 0
+            end
+
+            def inc
+              @count += 1
+            end
+          end
+        end
+
         describe '#fire' do
           it 'executes listeners subscribed as a string to the event name' do
             bus = described_class.new
-            dummy = Class.new do
-              attr_reader :run
-
-              def initialize
-                @run = false
-              end
-
-              def toggle
-                @run = true
-              end
-            end.new
-            bus.subscribe('foo') { dummy.toggle }
+            dummy = counter.new
+            bus.subscribe('foo') { dummy.inc }
 
             bus.fire 'foo'
 
-            expect(dummy.run).to be(true)
+            expect(dummy.count).to be(1)
           end
 
           it 'executes listeners subscribed as a regexp to the event name' do
             bus = described_class.new
-            dummy = Class.new do
-              attr_reader :run
-
-              def initialize
-                @run = false
-              end
-
-              def toggle
-                @run = true
-              end
-            end.new
-            bus.subscribe(/oo/) { dummy.toggle }
+            dummy = counter.new
+            bus.subscribe(/oo/) { dummy.inc }
 
             bus.fire 'foo'
 
-            expect(dummy.run).to be(true)
+            expect(dummy.count).to be(1)
           end
 
           it "doesn't execute listeners not subscribed to the event name" do
             bus = described_class.new
-            dummy = Class.new do
-              attr_reader :run
-
-              def initialize
-                @run = false
-              end
-
-              def toggle
-                @run = true
-              end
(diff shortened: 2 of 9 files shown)


adammathys (approved):
This seems pretty reasonable to me.

core/lib/spree/event/firing.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Spree
+  module Event
+    # The result of firing an event
+    #
+    # It encapsulates a fired {Spree::Event::Event} as long as the
+    # {Spree::Event::Execution}s it originated.
```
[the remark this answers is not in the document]
waiting-for-dev: Right :slightly_smiling_face: Thanks for looking into it. Fixed!

aldesantis (approved):

aldesantis:
@waiting-for-dev looks like we have some test failures here, but I'm not sure they're related to your changes. Could you take a look and re-run the CI if needed?

Other than that, this is outstanding work (as I've already said in the original PR), thanks! 👏

core/spec/lib/spree/event/adapters/default_spec.rb
```diff
@@ -99,6 +99,50 @@ def toggle
 
             expect(dummy.box).to eq('foo')
           end
+
+          it 'adds the fired event with given caller location to the firing result object' do
+            bus = described_class.new
+            dummy = Class.new do
```
spaghetticode, line 105: I noticed this file includes 7 repetitions of this class definition, I'm thinking that we can define it only once and reuse it since tests use instances so I don't think we're changing the class state... what do you think? They're about 11 lines each, so we're adding about 70 duplicate lines.

core/spec/lib/spree/event/adapters/default_spec.rb
```diff
@@ -99,6 +99,50 @@ def toggle
...
+            bus.subscribe('foo') { dummy.toggle }
+
+            firing = bus.fire 'foo', caller_location: caller_locations(0)[0]
+
+            expect(firing.event.caller_location.to_s).to include(__FILE__)
+          end
+
+          it 'adds the triggered executions to the firing result object' do
+            bus = described_class.new
+            dummy = Class.new do
```
spaghetticode, line 125: Also here, I think we can avoid introducing a new class definition and reuse the one(s) from above, because it seems to me that this spec doesn't depend on the logic of this class.

core/spec/lib/spree/event_spec.rb
```diff
@@ -40,11 +40,12 @@ def toggle
...
         end
       end.new
-      subject.subscribe('foo', adapter: bus) { dummy.toggle }
+      subject.subscribe('foo', adapter: bus) { dummy.toggle; :done }
 
-      subject.fire 'foo', adapter: bus
+      firing = subject.fire 'foo', adapter: bus
 
       expect(dummy.run).to be(true)
+      expect(firing.executions.first.result).to be(:done)
```
spaghetticode, line 48: given we're now having multiple expectations, I think it makes sense to add the `:aggregate_failures` metadata here.

waiting-for-dev: Makes sense. Done.

core/spec/lib/spree/event/adapters/default_spec.rb
waiting-for-dev: I tend to be quite a bigot when it comes not to sharing context between test examples (they should be basically about defining independent and isolated scenarios). Still, you're right, and I went too far! (I do have exceptions, and class definitions are one of them). So I've updated the code, and all tests use a `counter` class, which is pretty flexible for many situations. Thanks for pointing it out! It looks much tidier now! :tada:

waiting-for-dev:
> @waiting-for-dev looks like we have some test failures here, but I'm not sure they're related to your changes. Could you take a look and re-run the CI if needed?

Hmm, yeah, it seems unrelated. I'll look into it.

core/spec/lib/spree/event_spec.rb
waiting-for-dev: As you recommended in https://github.com/solidusio/solidus/pull/4204#discussion_r755859877, I've changed to use metadata instead of blocks.

kennyadsl (approved):
Thanks @waiting-for-dev, this is great!! 🎉 

We need to squash some commits (maybe all of them since they are just fixup from the previous block of changes introduced in the first commit). 

Also, I think it's time to start writing some guides about this new event adapter. As far as I remember, now we have only [this section](https://edgeguides.solidus.io/customization/customizing-the-core#using-the-event-bus) in the new guides, and it refers to the legacy adapter. All these improvements are worth a mention in the guides and I think we can write things incrementally along with the PR that we release. Maybe there's already a plan for this with @aldesantis? Nothing blocking anyway, I just wanted to share these thoughts. Thanks!!

kennyadsl:
Ah and this also needs a rebase against master, we recently merged #4209, which fixed the test suite.

spaghetticode (approved):
@waiting-for-dev looks great, thank you 👏

waiting-for-dev:
Thanks, folks!!

> We need to squash some commits (maybe all of them since they are just fixup from the previous block of changes introduced in the first commit).

Done

>  I think it's time to start writing some guides about this new event adapter.

I'll work on this

> Ah and this also needs a rebase against master, we recently merged #4209, which fixed the test suite.

Done. The test suite is :green_heart: now!

PR #4048, opened 2021-05-06
Use symbols in polymorphic path for event_links

tvdeyen:
**Description**

Rails 6.0.3.7 introduced a security fix, that forces us to use symbols
in polymorphic paths.

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
- [x] I have added tests to cover this change (if needed)

PR #4204, opened 2021-11-16
Introduce Spree::Event's test interface to run only selected listeners

waiting-for-dev:
Given the global nature of our event bus, we need a system to scope the
execution of a block to a selected list of subscribers. That's useful
for testing purposes, as we need to be sure that others subscribers are
not interfering with our expectations.

This commit introduces a `Spree::Event.performing_only(listeners)`
method. It takes a block during the execution of which only the provided
listeners are subscribed:

```ruby
listener1 = Spree::Event.subscribe('foo') { do_something }
listener2 = Spree::Event.subscribe('foo') { do_something_else }

Spree::Event.performing_only(listener1) do
  Spree::Event.fire('foo') # only listener1 will run
end

Spree::Event.fire('foo') # both listener1 & listener2 will run
```

This behavior is only available for the new
`Spree::Event::Adapters::Default` adapter.

We only need that for testing purposes, so the method is made available
after calling `Spree::Event.enable_test_interface`. It prevents the main
`Spree::Event` API from being bloated and sends a more explicit message
to users.

We also add a `Spree::Subscriber#listeners` method, which returns the
set of generated listeners for a given subscriber module. It's called
automatically by `Spree::Event.performing_only` so that users can
directly specify that they only want the listeners for a given
subscriber module to be run. `Spree::Subscriber#listeners` accepts an
array of event names as arguments in case more fine-grained control is
required.

```ruby
module EmailSubscriber
  include Spree::Event::Subscriber

  event_action :foo
  event_action :bar

  def foo(_event)
    do_something
  end

  def bar(_event)
    do_something_else
  end
end

Spree::Event.performing_only(EmailSubscriber) do
  Spree::Event.fire('foo') # both foo & bar methods will run
end

Spree::Event.performing_only(EmailSubscriber.listeners('foo')) do
  Spree::Event.fire('foo') # only foo method will run
end
```

A specialized `Spree::Event.performing_nothing` method calls
`Spree::Event.performing_only` with no listeners at all.

**Checklist:**

### DIFF
--- a/core/lib/spree/event/adapters/deprecation_handler.rb
+++ b/core/lib/spree/event/adapters/deprecation_handler.rb
@@ -12,7 +12,7 @@ module Spree
 
         CI_LEGACY_ADAPTER_ENV_KEY = 'CI_LEGACY_EVENT_BUS_ADAPTER'
 
-        def self.legacy_adapter?(adapter)
+        def self.legacy_adapter?(adapter = Spree::Config.events.adapter)
           adapter == LEGACY_ADAPTER
         end
 
@@ -20,7 +20,7 @@ module Spree
           return LEGACY_ADAPTER if ENV[CI_LEGACY_ADAPTER_ENV_KEY].present?
         end
 
-        def self.render_deprecation_message?(adapter)
+        def self.render_deprecation_message?(adapter = Spree::Config.events.adapter)
           legacy_adapter?(adapter) && legacy_adapter_set_by_env.nil?
         end
       end
--- a/core/lib/spree/event/subscriber.rb
+++ b/core/lib/spree/event/subscriber.rb
@@ -81,6 +81,47 @@ module Spree
       def deactivate(event_action_name = nil)
         Spree::Event.subscriber_registry.deactivate_subscriber(self, event_action_name)
       end
+
+      # Returns the generated listeners for the subscriber
+      #
+      # This method is only available when using
+      # [Spree::Event::Adapters::Default] adapter
+      #
+      # When a {Subscriber} is registered, the corresponding listeners are
+      # automatically generated, i.e., the returning values for
+      # {Spree::Event.subscribe} that encapsulate the logic to be performed.
+      #
+      # The listeners to obtain can be restricted to only certain events by providing
+      # their names:
+      #
+      # @example
+      #
+      #   module EmailSender
+      #     include Spree::Event::Subscriber
+      #
+      #     event_action :order_finalized
+      #     event_action :confirm_reimbursement
+      #
+      #     def order_finalized(event)
+      #       # ...
+      #     end
+      #
+      #     def confirm_reimbursement(event)
+      #       # ...
+      #     end
+      #   end
+      #
+      #   EmailSender.activate
+      #   EmailSender.listeners.count # => 2
+      #   EmailSender.listeners("order_finalized").count # => 1
+      #
+      # @param event_names [Array<String, Symbol>]
+      # @return [Array<Spree::Event::Listener>]
+      # @raise [RuntimeError] When adapter is
+      # [Spree::Event::Adapters::ActiveSupportNotifications]
+      def listeners(*event_names)
+        Spree::Event.subscriber_registry.listeners(self, event_names: event_names)
+      end
     end
   end
 end
--- a/core/lib/spree/event/test_interface.rb
+++ b/core/lib/spree/event/test_interface.rb
@@ -1,0 +1,93 @@
+# frozen_string_literal: true
+
+require 'spree/event'
+
+module Spree
+  module Event
+    # Test helpers for {Spree::Event}
+    #
+    # This module defines test helpers methods for {Spree::Event}. They can be
+    # made available to {Spree::Event} when {Spree::Event.enable_test_interface}
+    # is called.
+    #
+    # If you prefer, you can directly call them from
+    # `Spree::Event::TestInterface}.
+    module TestInterface
+      # @see {Spree::Event::TestInterface}
+      module Methods
+        # Perform only given listeners for the duration of the block
+        #
+        # Temporarily deactivate all subscribed listeners and listen only to the
+        # provided ones for the duration of the block.
+        #
+        # @example
+        #   Spree::Event.enable_test_interface
+        #
+        #   listener1 = Spree::Event.subscribe('foo') { do_something }
+        #   listener2 = Spree::Event.subscribe('foo') { do_something_else }
+        #
+        #   Spree::Event.performing_only(listener1) do
+        #     Spree::Event.fire('foo') # This will run only `listener1`
+        #   end
+        #
+        #   Spree::Event.fire('foo') # This will run both `listener1` & `listener2`
+        #
+        # {Spree::Event::Subscriber} modules can also be given to unsubscribe from
+        # all listeners generated from it:
+        #
+        # @example
+        #   Spree::Event.performing_only(EmailSubscriber) {}
+        #
+        # You can gain more fine-grained control thanks to
+        # {Spree::Event::Subscribe#listeners}:
+        #
+        # @example
+        #   Spree::Event.performing_only(EmailSubscriber.listeners('order_finalized')) {}
+        #
+        # You can mix different ways of specifying listeners without problems:
+        #
+        # @example
+        #   Spree::Event.performing_only(EmailSubscriber, listener1) {}
+        #
+        # @param listeners_and_subscribers [Spree::Event::Listener,
+        # Array<Spree::Event::Listener>, Spree::Event::Subscriber]
+        # @yield While the block executes only provided listeners will run
+        def performing_only(*listeners_and_subscribers)
+          adapter_in_use = Spree::Event.default_adapter
+          listeners = listeners_and_subscribers.flatten.map(&:listeners)
+          Spree::Config.events.adapter = adapter_in_use.with_listeners(listeners.flatten)
+          yield
+        ensure
+          Spree::Config.events.adapter = adapter_in_use
+        end
+
+        # Perform no listeners for the duration of the block
+        #
+        # It's a specialized version of {#performing_only} that provides no
+        # listeners.
+        #
+        # @yield While the block executes no listeners will run
+        #
+        # @see Spree::Event::TestInterface#performing_only
+        def performing_nothing(&block)
+          performing_only(&block)
+        end
+      end
+
+      extend Methods
+    end
+
+    # Adds test methods to {Spree::Event}
+    #
+    # @raise [RuntimeError] when {Spree::Event::Configuration#adapter} is set to
+    # the legacy adapter {Spree::Event::Adapters::ActiveSupportNotifications}.
+    def enable_test_interface
+      raise <<~MSG if deprecation_handler.legacy_adapter?(default_adapter)
+        Spree::Event's test interface is not supported when using the deprecated
+        adapter 'Spree::Event::Adapters::ActiveSupportNotifications'.
+      MSG
+
+      extend(TestInterface::Methods)
+    end
+  end
+end
--- a/core/spec/lib/spree/event/adapters/default_spec.rb
+++ b/core/spec/lib/spree/event/adapters/default_spec.rb
@@ -189,6 +189,25 @@ module Spree
             expect(dummy2.count).to be(1)
           end
         end
+
+        describe '#with_listeners' do
+          it 'returns a new instance with given listeners', :aggregate_failures do
+            bus = described_class.new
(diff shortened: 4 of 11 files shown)


core/lib/spree/event/subscriber.rb
```diff
@@ -81,6 +81,47 @@ def activate
       def deactivate(event_action_name = nil)
         Spree::Event.subscriber_registry.deactivate_subscriber(self, event_action_name)
       end
+
+      # Returns the generated listeners for the subscriber
+      #
+      # This method is only available when usiing
```
adammathys, line 87: `usiing` => `using`

core/lib/spree/event/test_interface.rb
```diff
@@ -0,0 +1,84 @@
...
+      # all listeners generated from it:
+      #
+      # @example
+      #   Spree::Event.performing_only(EmailSubscriber) {}
+      #
+      # You can gain more fine-grained control thanks to
+      # {Spree::Event::Subscribe#listeners}:
+      #
+      # @example
+      #   Spree::Event.performing_only(EmailSubscriber.listeners('order_finalized') {}
```
adammathys, line 39: Missing an `)`

adammathys (approved):
Overall looks pretty reasonable to me.

core/lib/spree/event/subscriber.rb
waiting-for-dev: Thanks, fixed!

core/lib/spree/event/test_interface.rb
waiting-for-dev: Thanks, fixed!

core/spec/lib/spree/event/adapters/default_spec.rb
```diff
@@ -243,6 +243,37 @@ def toggle
             expect(dummy2.run).to be(true)
           end
         end
+
+        describe '#with_listeners' do
+          it 'returns a new instance with given listeners' do
+            bus = described_class.new
+            dummy1, dummy2, dummy3 = Array.new(3) do
+              Class.new do
```
waiting-for-dev, line 251: @spaghetticode, if you see it (now you will ;)), I'll also use the counter class here as in #4203, but I'll rebase from master once that branch is merged.

core/spec/lib/spree/event/test_interface_spec.rb
```diff
@@ -0,0 +1,193 @@
...
+
+        aggregate_failures do
+          expect(dummy1.count).to be(1)
+          expect(dummy2.count).to be(1)
+          expect(dummy3.count).to be(0)
+        end
+      ensure
+        Spree::Event.unsubscribe(listener1)
+        Spree::Event.unsubscribe(listener2)
+        Spree::Event.unsubscribe(listener3)
```
spaghetticode, line 59: I'm not 100% sure we need those `ensure` blocks: I think that the `before` block from above by calling `Spree::Event.enable_test_interface` sets the adapter to a throwaway one, is that correct?

spaghetticode (commented):
@waiting-for-dev thanks for this useful addition! I've left a few notes in regards to specs... the application code looks great!

core/spec/lib/spree/event/subscriber_registry_spec.rb
```diff
@@ -112,4 +115,44 @@ def other_event(event)
...
+      after { subject.deactivate_subscriber(N) }
+
+      it 'returns all listeners that the subscriber generates' do
+        listeners = subject.listeners(N)
+
+        expect(listeners.count).to be(2)
+        expect(listeners.first).to be_a(Spree::Event::Listener)
+      end
+
+      it 'can restrict by event names' do
```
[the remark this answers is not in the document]
waiting-for-dev: Done. Do you think there is any drawback to having that setting configured globally for RSpec?

core/spec/lib/spree/event/subscriber_registry_spec.rb
```diff
@@ -112,4 +115,44 @@ def other_event(event)
...
+        subject.register(N)
+        subject.activate_subscriber(N)
+      end
+      after { subject.deactivate_subscriber(N) }
+
+      it 'returns all listeners that the subscriber generates' do
+        listeners = subject.listeners(N)
+
+        expect(listeners.count).to be(2)
+        expect(listeners.first).to be_a(Spree::Event::Listener)
```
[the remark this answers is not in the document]
waiting-for-dev: Done

core/spec/lib/spree/event/test_interface_spec.rb
waiting-for-dev: Good point. `#enable_test_interface` only includes some methods to the module, but `#performing_only` has its own `ensure` block to, as you said, restore the previous adapter. So there's really no point for them in the test file. I removed all of them, besides the ones dealing with the subscriber registry.

core/spec/lib/spree/event/subscriber_registry_spec.rb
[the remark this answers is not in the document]
spaghetticode: It seems to be a reasonable default... but I'd be curious to check what happens with Capybara specs in that case 🤔

core/spec/lib/spree/event/adapters/default_spec.rb
waiting-for-dev: Done

jarednorman (approved):

core/lib/spree/event/test_interface.rb
```diff
@@ -0,0 +1,84 @@
...
+    # @raise [RuntimeError] when {Spree::Event::Configuration#adapter} is set to
+    # the legacy adapter {Spree::Event::Adapters::ActiveSupportNotifications}.
+    def enable_test_interface
+      raise <<~MSG if deprecation_handler.legacy_adapter?(default_adapter)
+        Spree::Event's test interface is not supported when using the deprecated
+        adapter 'Spree::Event::Adapters::ActiveSupportNotifications'.
+      MSG
+
+      extend(TestInterface)
+    end
```
forkata, line 82: What are your thoughts on having this as a stand-alone module (something like `Spree::TestingSupport::Event`) rather than extending the `Spree::Event` module? It doesn't seem like we are not using anything other than `default_adapter` and the `deprecation_handler` both of which we have access to.

waiting-for-dev: In my view, having those methods on `Spree::Event` is more convenient and communicate better. Otherwise, to use them, we should do something like:

```ruby
RSpec.describe SomeClass do
  include Spree::TestingSupport::Event

  it 'does something' do
    # ...
    performing_only(listener) {}
    # ...
  end
end
```

We're already introducing a testing helper module for events in https://github.com/solidusio/solidus/pull/4048. However, they are stubbing helpers, and the module includes RSpec matches. My idea was to separate both kinds of testing strategies. Stubbing methods are indeed testing helpers. However, `#performing_only` is just an extension of the core event bus behavior (for instance, it has nothing to do with RSpec). We could have that method within `Spree::Event` itself, but we use ruby's flexibility to communicate intentions better. However, if you feel very strongly about it, we can revisit it.

forkata: Thanks for the explanation @waiting-for-dev, that makes sense as a motivation for the approach you have taken. I think the naming suggestion I used may have not been the best, as I wasn't suggesting we mix this "new" module into our test classes, but rather provide a different class from `Spree::Event` which has the `#performing_only` interface. I thought that may provide some more clarity over enhancing the behaviour of the existing class by calling `enable_test_interface` which isn't a reversible action. I don't feel strongly about this, so no need to change anything here 😄

waiting-for-dev: I see. Yeah, having them called directly from the test module is also clear. I think I reached a good middle ground, and now both systems are supported! So you can both:

```ruby
Spree::Event.enable_test_interface
Spree::Event.performing_only(...) { ... }
```

or

```ruby
Spree::Event::TestInterface.performing_only(...) { ... }
```

We still use the former on the doc examples, but we do support the latter (we mention it on the docs and have a test covering it). WDYT? :slightly_smiling_face:

core/lib/spree/event/adapters/deprecation_handler.rb
```diff
@@ -12,15 +12,15 @@ module DeprecationHandler
 
         CI_LEGACY_ADAPTER_ENV_KEY = 'CI_LEGACY_EVENT_BUS_ADAPTER'
 
-        def self.legacy_adapter?(adapter)
+        def self.legacy_adapter?(adapter = Spree::Config.events.adapter)
```
kennyadsl, line 15: Why this is needed in this PR?

core/lib/spree/event/test_interface.rb
```diff
@@ -0,0 +1,93 @@
...
+        # You can mix different ways of specifying listeners without problems:
+        #
+        # @example
+        #   Spree::Event.performing_only(EmailSubscriber, listener1) {}
+        #
+        # @param listeners_and_subscribers [Spree::Event::Listener,
+        # Array<Spree::Event::Listener>, Spree::Event::Subscriber]
+        # @yield While the block executes only provided listeners will run
+        def performing_only(*listeners_and_subscribers)
+          old_adapter = Spree::Event.default_adapter
```
kennyadsl, line 56: I don't understand why we need to store the old adapter, can you help me?

core/lib/spree/event/test_interface.rb
forkata: That sounds great, thanks @waiting-for-dev! This seems like a good compromise which shouldn't add any maintenance overhead 👍🏼

core/lib/spree/event/test_interface.rb
forkata: I believe this is necessary because `Spree::Event.fire` uses the `default_adapter` unless one is explicitly passed in the options [here](https://github.com/solidusio/solidus/blob/master/core/lib/spree/event.rb#L73), so we have to replace that with one built by the `with_listeners` call for the duration of the tests, and then reset to the adapter instance which is not configured with a filter on the listeners.

waiting-for-dev: Exactly. The event bus is global. We change it for the duration of the `performing_only` block.

core/lib/spree/event/adapters/deprecation_handler.rb
waiting-for-dev: This PR adds other caller locations that have no direct access to the used adapter. It's easier to default it to the global adapter, instead of making the same operation from the caller each time. E.g.:

- https://github.com/solidusio/solidus/blob/5881d866e70687194cbdf4f08ab4b0047296cba5/core/lib/spree/event/subscriber_registry.rb#L56
- https://github.com/solidusio/solidus/blob/5881d866e70687194cbdf4f08ab4b0047296cba5/core/spec/lib/spree/event/subscriber_spec.rb#L95
- https://github.com/solidusio/solidus/blob/5881d866e70687194cbdf4f08ab4b0047296cba5/core/spec/lib/spree/event/test_interface_spec.rb#L37
https://github.com/solidusio/solidus/blob/5881d866e70687194cbdf4f08ab4b0047296cba5/core/spec/lib/spree/event/subscriber_registry_spec.rb#L120

core/lib/spree/event/test_interface.rb
kennyadsl: Got it, I was confused because I wrongly interpreted `old_adapter` as `legacy_adapter`. What about changing it to `adapter_in_use` or something similar, instead?

kennyadsl (approved):
Non blocking:

1) suggested to change a variable name
2) I would create two commits for this PR, moving in a separate one the part that adds the default adapter to methods signature, which is related but IMHO deserves some kind of specific documentation via its commit message.

core/lib/spree/event/test_interface.rb
waiting-for-dev: Oh, I see. Yeah, it can be confusing. I renamed it to `adapter_in_use` now :slightly_smiling_face:

waiting-for-dev:
@kennyadsl 

>     1. suggested to change a variable name

Done!

>     2. I would create two commits for this PR, moving in a separate one the part that adds the default adapter to methods signature, which is related but IMHO deserves some kind of specific documentation via its commit message.

Unfortunately, that change belongs to the same unit of work. It'd leave the first commit in an inconsistent state (tests failing), so I'd prefer not to do that. However, I don't think it's that important, as we should remove all the deprecation logic once we remove the old adapter. WDYT? If you feel very strongly about it, we can try to find a way, though :slightly_smiling_face:

kennyadsl (approved):
Thanks Marc!

PR #4209, opened 2021-11-24
Use SSL on fonts.googleapis.com scss import

RyanofWoods:
Since 23/11/2021, `@import url(//fonts.googleapis.com/css...)` began to
cause errors; Breaking Solidus and all extensions that depend on
solidus_frontend. Explicitly using https protocol fixed this.

Error example:
https://app.circleci.com/pipelines/github/solidusio/solidus/2791/workflows/e62ff646-9ae6-4e65-b20f-e1f7a109d3a1/jobs/26572

Blocking of this URL in testing_support was updated to keep performance
improvement.

**Checklist:**

### DIFF
--- a/backend/spec/spec_helper.rb
+++ b/backend/spec/spec_helper.rb
@@ -39,6 +39,7 @@ require 'spree/testing_support/capybara_ext'
 require 'spree/testing_support/precompiled_assets'
 require 'spree/testing_support/translations'
 require 'spree/testing_support/job_helpers'
+require 'spree/testing_support/blacklist_urls'
 
 require 'capybara-screenshot/rspec'
 Capybara.save_path = ENV['CIRCLE_ARTIFACTS'] if ENV['CIRCLE_ARTIFACTS']
@@ -96,9 +97,6 @@ RSpec.configure do |config|
 
   config.before do
     Rails.cache.clear
-    if RSpec.current_example.metadata[:js] && page.driver.browser.respond_to?(:url_blacklist)
-      page.driver.browser.url_blacklist = ['http://fonts.googleapis.com']
-    end
   end
 
   config.include BaseFeatureHelper, type: :feature
@@ -112,6 +110,7 @@ RSpec.configure do |config|
   config.include Spree::TestingSupport::Flash
   config.include Spree::TestingSupport::Translations
   config.include Spree::TestingSupport::JobHelpers
+  config.include Spree::TestingSupport::BlacklistUrls
 
   config.extend WithModel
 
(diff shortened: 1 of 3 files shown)


RyanofWoods:
I am not really sure why this line suddenly started causing issues. Would be interested to know if anyone has any insight.

waiting-for-dev (changes requested):
Many thanks for looking into it @RyanofWoods! :heart: 

I haven't found anything out there, but I guess that since that date, `fonts.googleapis.com` doesn't support non-SSL connections anymore. Anyway, it's good from our side to request through `HTTPS`.

I think we also need to update the URL here:

https://github.com/solidusio/solidus/blob/e878076f2ed670d07654ab6293a16588743f2fa6/core/lib/spree/testing_support/blacklist_urls.rb#L8

I'd also take the occasion to remove the duplication from this line:

https://github.com/solidusio/solidus/blob/d4925d83ef619c1a752c1f5f4f00e9597dd07510/backend/spec/spec_helper.rb#L100

And instead require `spree/testing_support/blacklsit_urls` as is done from the frontend engine:

https://github.com/solidusio/solidus/blob/d4925d83ef619c1a752c1f5f4f00e9597dd07510/frontend/spec/spec_helper.rb#L39

backend/spec/spec_helper.rb
```diff
@@ -94,13 +95,6 @@
     DatabaseCleaner.clean_with :truncation
   end
 
-  config.before do
-    Rails.cache.clear
```
waiting-for-dev, line 98: We should not remove this line, right?

RyanofWoods: I think not 👍 I added it back now.

waiting-for-dev (approved):
Thanks @RyanofWoods :raised_hands: :raised_hands:

RyanofWoods:
> Thanks @RyanofWoods 🙌 🙌

Thank you for your help @waiting-for-dev! Amazing spot with the `testing_support` 🙌

gmacdougall (approved):

jarednorman (approved):

kennyadsl (approved):
👍

spaghetticode (approved):
@RyanofWoods thank you 👍

kennyadsl:
@RyanofWoods thanks! I think we need to release this in all the supported Solidus versions as well. Can you make the same PR against [v3.1](https://github.com/solidusio/solidus/tree/v3.1), [v3.0](https://github.com/solidusio/solidus/tree/v3.0) and [v2.11](https://github.com/solidusio/solidus/tree/v2.11), please?

RyanofWoods:
> @RyanofWoods thanks! I think we need to release this in all the supported Solidus versions as well. Can you make the same PR against [v3.1](https://github.com/solidusio/solidus/tree/v3.1), [v3.0](https://github.com/solidusio/solidus/tree/v3.0) and [v2.11](https://github.com/solidusio/solidus/tree/v2.11), please?

Done @kennyadsl 👍 Let me know if I could have done it in a more helpful way 🙂

PR #4210, opened 2021-11-25
v3.1 - Use SSL on fonts.googleapis.com scss import

RyanofWoods:
Since 23/11/2021, @import url(//fonts.googleapis.com/css...) began to
cause errors; Breaking Solidus and all extensions that depend on
solidus_frontend. Explicitly using HTTPS protocol fixed this.

Error example:
https://app.circleci.com/pipelines/github/solidusio/solidus/2791/workflows/e62ff646-9ae6-4e65-b20f-e1f7a109d3a1/jobs/26572

Blocking of this URL in testing_support was updated to keep performance
improvement.

Checklist:

- [x]  I have followed Pull Request guidelines
- [x]  I have added a detailed description into each commit message

### DIFF
--- a/backend/spec/spec_helper.rb
+++ b/backend/spec/spec_helper.rb
@@ -39,6 +39,7 @@ require 'spree/testing_support/capybara_ext'
 require 'spree/testing_support/precompiled_assets'
 require 'spree/testing_support/translations'
 require 'spree/testing_support/job_helpers'
+require 'spree/testing_support/blacklist_urls'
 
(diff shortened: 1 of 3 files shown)


waiting-for-dev (approved):

jarednorman (approved):

jarednorman:
I see there's PRs back to v2.11, but we may need to backport this to more versions. It's going to break specs across tons of extensions that support and runs specs against any older versions. 😩

PR #4213, opened 2021-11-25
Make display_price optional on admin variants list

luca-landa:
When there are variants which respond `nil` as `#default_price` (e.g. because admin is creating them and they still don't have all prices set up, or they don't have a price matching `Spree::Variant.default_price_attributes`), the `/admin/products/<product>/variants` page returns 500 and it's not obvious why to the admin.

This small fix allows for the variants list page to be displayed anyway.

**Description**

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/variants/_table.html.erb
+++ b/backend/app/views/spree/admin/variants/_table.html.erb
@@ -27,7 +27,7 @@
         <% end %>
       </td>
       <td><%= variant.options_text %></td>
-      <td><%= variant.display_price.to_html %></td>
+      <td><%= variant.display_price&.to_html %></td>
       <td><%= variant.sku %></td>
       <td class="actions">
         <% if can?(:update, variant) %>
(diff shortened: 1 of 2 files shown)


jarednorman (approved):

waiting-for-dev (commented):
Thanks, @luca-landa! Could we have a test for that to avoid possible future regressions?

luca-landa:
> Thanks, @luca-landa! Could we have a test for that to avoid possible future regressions?

@waiting-for-dev added! I wasn't sure if I should have created a factory for the variant needed in this test, being a unique use case, let me know if you think I should change something 🙏

waiting-for-dev (approved):

waiting-for-dev:
> @waiting-for-dev added! I wasn't sure if I should have created a factory for the variant needed in this test, being a unique use case, let me know if you think I should change something 

Thanks, @luca-landa! I think there's no need for a custom factory :slightly_smiling_face:

jarednorman (approved):

tvdeyen:
This should be backported to v3.1 and v3.0 as well.

PR #4214, opened 2021-11-26
Add stubbing test helpers for the event bus

waiting-for-dev:
**Description**

This commit adds a `Spree::TestingSupport::EventHelpers` module which
adds stubbing helpers for `Spree::Event` when included in an RSpec file.

It adds a RSpec matcher, `have_been_fired`, which can be used after the
event bus has been stubbed through `stub_events`.

Internally, it leans on the `have_received` built-in `rspec-mocks`
matcher. Unfortunately, RSpec's matcher lacks good extensibility, so we
can't provide OOTB with all its bells and whistles (like `once`,
`exactly(2).times`), and the effort is not worth it at this point.
However, we do support a `with` modifier for `have_been_fired` to match
the published payload.

Example:

```ruby
require 'rails_helper'
require 'spree/testing_support/event_helpers'

RSpec.describe MyClass do
  include Spree::TestingSupport::EventHelpers

  it 'does stuff with events' do
    stub_events

    described_class.new.do_stuff_with_events

    expect('foo').to have_been_fired
    expect('bar').to have_been_fired.with(a_hash_containing(foo: :bar))
  end
end
```

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/event_helpers.rb
+++ b/core/lib/spree/testing_support/event_helpers.rb
@@ -1,0 +1,103 @@
+# frozen_string_literal: true
+
+require 'spree/event'
+
+module Spree
+  module TestingSupport
+    # RSpec test helpers for the event bus
+    #
+    # If you want to use the methods defined in this module, include it in your
+    # specs:
+    #
+    # @example
+    #   require 'rails_helper'
+    #   require 'spree/testing_support/event_helpers'
+    #
+    #   RSpec.describe MyClass do
+    #     include Spree::TestingSupport::EventHelpers
+    #   end
+    #
+    # or, globally, in your `spec_helper.rb`:
+    #
+    # @example
+    #   require 'spree/testing_support/event_helpers'
+    #
+    #   RSpec.configure do |config|
+    #     config.include Spree::TestingSupport::EventHelpers
+    #   end
+    module EventHelpers
+      extend RSpec::Matchers::DSL
+
+      # Stubs {Spree::Event}
+      #
+      # After you have called this method in an example, {Spree::Event} will no
+      # longer listen to any event for the duration of that example. All the
+      # method invocations on it will be spied but not performed.
+      #
+      # Internally, it stubs {Spree::Event} to a class spy of itself.
+      #
+      # After you call this method, probably you'll want to call some of the
+      # matchers defined in this module.
+      def stub_spree_events
+        stub_const('Spree::Event', class_spy(Spree::Event))
+      end
+
+      # @!method have_been_fired(event_name)
+      #   Matcher to test that an event has been fired via {Spree::Event#fire}
+      #
+      #   Before using this matcher, you need to call {#stub_spree_events}.
+      #
+      #   Remember that the event listeners won't be performed.
+      #
+      #   @example
+      #     it 'fires foo event' do
+      #       stub_spree_events
+      #
+      #       Spree::Event.fire 'foo'
+      #
+      #       expect('foo').to have_been_fired
+      #     end
+      #
+      #   It can be chain through `with` to match with the published payload:
+      #
+      #   @example
+      #     it 'fires foo event with the expected payload' do
+      #       stub_spree_events
+      #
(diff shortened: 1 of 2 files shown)


kennyadsl (approved):
Sad that we can't use `once` or similar, but this is a great starting point. Thanks @waiting-for-dev!

adammathys (approved):
Very cool. Might be nice to have it print out the difference in arguments when they're provided, but this is a great initial implementation.

waiting-for-dev:
> Very cool. Might be nice to have it print out the difference in arguments when they're provided, but this is a great initial implementation.

Thanks, @adammathys. Yeah, I agree. Unfortunately, there's [no straightforward way to reuse the logic in RSpec's `have_received` matcher](https://github.com/rspec/rspec-mocks/blob/17cf86ab61544b93232c4e9ca9784ab212dccbf6/lib/rspec/mocks/matchers/have_received.rb#L119). So, it's easy to show the payload given in the expectation, but not the actual one :disappointed: For now, I feel that it's not worth it to make the code more complex...

spaghetticode:
@waiting-for-dev this looks so convenient! I was thinking that we could also use rspec metadata so we don't need to explicitly write `stub_events` in the example body, something like this:
```ruby
it 'fires foo event', :stub_events do
  Spree::Event.fire 'foo'

  expect('foo').to have_been_fired
end
```
what do you think?

waiting-for-dev:
> @waiting-for-dev this looks so convenient! I was thinking that we could also use rspec metadata so we don't need to explicitly write `stub_events` in the example body, something like this:
> 
> ```ruby
> it 'fires foo event', :stub_events do
>   Spree::Event.fire 'foo'
> 
>   expect('foo').to have_been_fired
> end
> ```
> 
> what do you think?

@spaghetticode, I think it's a great idea to use it that way! However, I'm not particularly eager to modify RSpec configuration from a library. I think that's something that we could address from the docs point of view. How do you feel about that?

spaghetticode:
> @spaghetticode, I think it's a great idea to use it that way! However, I'm not particularly eager to modify RSpec configuration from a library. I think that's something that we could address from the docs point of view. How do you feel about that?

I think I get what you mean, but I'm not sure I would consider that addition something that intrusive, the possible clash with other metadata names is unlikely, but if this is a concern then the name could be changed to `:stub_spree_events` (actually, while I am writing this I am starting to think it may be a good idea to change also the helper method name to that 🤔)... then Solidus is a Rails engine, not a simple library so some level of convenience at the price of some obtrusive configuration doesn't seem unreasonable to me. But in the end I don't have strong opinions, I think it's reasonable both ways!

spaghetticode (approved):

waiting-for-dev:
> actually, while I am writing this I am starting to think it may be a good idea to change also the helper method name to that

I think you're right, and it makes it consistent with the approach used elsewhere in the codebase. I just renamed the method to `stub_spree_events`

> then Solidus is a Rails engine, not a simple library so some level of convenience at the price of some obtrusive configuration doesn't seem unreasonable to me. But in the end I don't have strong opinions, I think it's reasonable both ways!

Thanks for bringing up those points. That's entirely a valid reason! I think my hesitance grounds in two points:

- As I said, if possible, I always try to avoid changing user configuration (also from an engine)
- I'm not particularly fond of controlling a test context through metadata:
  - It adds indirection
  - Encourages sharing state between tests
  - It makes debugging more difficult as we're using something that is no actual code (so to speak)

Considering that the whole stubbing approach is not my favorite one (I'm more in favor of using #155), I wouldn't like to make it too easy to use if you get my point. For instance, if someone adds `stub_spree_event: true` on the top `describe` block, it'd encourage using it elsewhere in the file.

On top of that, it'll be a one-liner to add it the user application. Otherwise, `before(:each)` blocks can be used. It's only a few keystrokes longer and at least is actual code (again, so to speak).

Now, I understand that those are all opinable reasons. But, if you don't have strong opinions about it, I'd prefer to leave it as it is for now. It'll always be easier to add it than to remove it :slightly_smiling_face:

PR #4222, opened 2021-12-13
Add filter and pagination to tax rates admin view

tvdeyen:
**Description**

Managing a lot of tax rates is currently not very user friendly.

Adding pagination, filtering by zone and tax category and a default
order by zone makes managing tax much nicer. And who does not love
managing taxes.

**Screenshots**

Tax Rates - Taxes - Settings 2021-12-13 21-08-23

Tax Rates - Taxes - Settings 2021-12-13 21-08-34

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/tax_rates/index.html.erb
+++ b/backend/app/views/spree/admin/tax_rates/index.html.erb
@@ -83,7 +83,7 @@
         <td><%=tax_rate.included_in_price? ? t('spree.say_yes') : t('spree.say_no') %></td>
         <td><%=tax_rate.show_rate_in_label? ? t('spree.say_yes') : t('spree.say_no') %></td>
         <td><%=tax_rate.expires_at.to_date.to_s(:short_date) if tax_rate.expires_at %></td>
-        <td><%=tax_rate.calculator.to_s %></td>
+        <td><%=tax_rate.calculator && tax_rate.calculator.class.model_name.human %></td>
         <td class="actions">
           <% if can?(:update, tax_rate) %>
             <%= link_to_edit tax_rate, no_text: true %>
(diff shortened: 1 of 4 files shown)


jarednorman (approved):

spaghetticode (approved):
Thanks @tvdeyen 🎉

backend/app/views/spree/admin/tax_rates/index.html.erb
```diff
@@ -55,7 +83,7 @@
         <td><%=tax_rate.included_in_price? ? t('spree.say_yes') : t('spree.say_no') %></td>
         <td><%=tax_rate.show_rate_in_label? ? t('spree.say_yes') : t('spree.say_no') %></td>
         <td><%=tax_rate.expires_at.to_date.to_s(:short_date) if tax_rate.expires_at %></td>
-        <td><%=tax_rate.calculator.to_s %></td>
+        <td><%=tax_rate.calculator && tax_rate.calculator.class.model_name.human %></td>
```
waiting-for-dev, line 86: What do you think about having it in a helper?

tvdeyen: I prefer it that way. This is established Rails.

What do you hope to gain from a helper?

waiting-for-dev: I like to avoid logic in the template. It helps prevent duplication and makes templates more readable. But that's non-blocking. As you said, it's established Rails, although I disagree with it :slightly_smiling_face:

waiting-for-dev (approved):
Thanks, @tvdeyen!

backend/app/views/spree/admin/tax_rates/index.html.erb
tvdeyen: I see. I would not call this logic in the template. It's a translation with a safe guard (that might be unnecessary, but the test failed without).

Also Solidus being an extendable framework I would even argue against helpers at all. Presenters and templates are way easier to overwrite and extend. 

But that's another discussion. Thanks for the review.

PR #4224, opened 2021-12-16
v2.11 fix(Address): Set name from firstname and lastname on update

### LINKED ISSUE #4094: Solidus 2.11 - it's impossible to update a split address name from the backend
On Solidus 2.11, theres a bug that prevents you from updating firstname and lastname from admin.

[It all starts with the addresses `value_attributes` class method.](https://github.com/solidusio/solidus/blob/07bdb8d180d643fcdf3e769d8f0d197b3d65d44a/core/app/models/spree/address.rb#L85) It combines the old addresses attributes with the attribute changes that you're trying to make. That way, your changes are layered on top of the existing address information. However, if you've set `use_combined_first_and_last_name_in_address` to false, you won't be passing in a `name` attribute (instead using `firstname` and `lastname`). But the old address will still have a `name` attribute.

So the combined attributes would be something like: `{ firstname: 'Bruce', lastname: 'Wayne', name: 'The Batman' }`.

[On line 88 of the method](https://github.com/solidusio/solidus/blob/07bdb8d180d643fcdf3e769d8f0d197b3d65d44a/core/app/models/spree/address.rb#L88), we generate a new name with `Spree::Address::Name.from_attributes`. But if you pass in a `name` attribute, it just uses that instead. So even if you pass in the above example, it will disregard the `firstname` and `lastname` and just use `name` instead - which comes from the old address.

***Additionally***

If that were fixed, there's another issue - [on line 97](https://github.com/solidusio/solidus/blob/07bdb8d180d643fcdf3e769d8f0d197b3d65d44a/core/app/models/spree/address.rb#L97) of the value attributes method, you'll see that we only replace the `name` attribute with the new generated `name` *only* if the `name` attribute is not present. So even if we did generate the correct name above, the new address would still use the incorrect `name`.

***Last note***

The address model knows that it needs to generate a new address when you pass in an updated `firstname` or `lastname`, so it does that in this scenario... but those attributes don't change. Meaning, you generate an exact copy of the address you're trying to change every time you attempt to change it.

https://www.loom.com/share/37ab0c961dbd4ab9b40e39e8fe4bfff3

**Solidus Version**:
2.11

**To Reproduce**
Start a Solidus store on 2.11 from scratch
Set `use_combined_first_and_last_name_in_address` config to false
Go to admin, attempt to change an addresses firstname or lastname.

tvdeyen:
**Description**

If `use_combined_first_and_last_name_in_address` is `false` (default in v2.11) and `firstname` and `lastname` are given in the changeset (ie. from a form) we need to make sure that name is set from it, otherwise data will be inconsistent.

Closes #4094

**Checklist:**

### DIFF
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -84,6 +84,7 @@ module Spree
     # @return [Hash] hash of attributes contributing to value equality with optional merge
     def self.value_attributes(base_attributes, merge_attributes = {})
       base = base_attributes.stringify_keys.merge(merge_attributes.stringify_keys)
+      base.delete("name") unless Spree::Config.use_combined_first_and_last_name_in_address
 
       name_from_attributes = Spree::Address::Name.from_attributes(base)
       if base['firstname'].presence || base['first_name'].presence
--- a/core/spec/models/spree/address_spec.rb
+++ b/core/spec/models/spree/address_spec.rb
@@ -109,12 +109,56 @@ RSpec.describe Spree::Address, type: :model do
     end
   end
 
-  context "when saving a record" do
+  context "when creating a record" do
     context "when the `name` field is not explicitly set" do
       subject { build :address, name: nil, firstname: 'John', lastname: 'Doe' }
 
       it "sets `name` from `firstname` and `lastname`" do
         expect { subject.save }.to change { subject.read_attribute(:name) }.from(nil).to('John Doe')
+      end
+    end
+  end
+
+  context "when updating a record" do
+    let(:address) { create(:address, firstname: "John", lastname: "Doe") }
+
+    context "if use_combined_first_and_last_name_in_address is set to false (default)" do
+      before do
+        allow(Spree::Config).to receive(:use_combined_first_and_last_name_in_address) { false }
+      end
+
+      context "and the `name` attribute is not in changeset" do
+        it "sets `name` from `firstname` and `lastname`" do
+          new_address = described_class.immutable_merge(address, firstname: "Jane", lastname: "Fonda")
+          expect(new_address.read_attribute(:name)).to eq("Jane Fonda")
+        end
+      end
+
+      context "and the `name` attribute is in changeset" do
+        it "does not update name" do
+          new_address = described_class.immutable_merge(address, name: "Jane Fonda")
+          expect(new_address.read_attribute(:name)).to eq("John Von Doe")
+        end
+      end
+    end
+
+    context "if use_combined_first_and_last_name_in_address is set to true" do
+      before do
+        allow(Spree::Config).to receive(:use_combined_first_and_last_name_in_address) { true }
+      end
+
+      context "and the `name` attribute is not in changeset" do
+        it "keeps old name" do
+          new_address = described_class.immutable_merge(address, firstname: "Jane", lastname: "Fonda")
+          expect(new_address.read_attribute(:name)).to eq("John Von Doe")
+        end
+      end
+
+      context "and the `name` attribute is in changeset" do
+        it "updates name" do
+          new_address = described_class.immutable_merge(address, name: "Jane Fonda")
+          expect(new_address.read_attribute(:name)).to eq("Jane Fonda")
+        end
       end
     end
   end


core/app/models/spree/address.rb
```diff
@@ -84,6 +84,7 @@ def self.immutable_merge(existing_address, new_attributes)
     # @return [Hash] hash of attributes contributing to value equality with optional merge
     def self.value_attributes(base_attributes, merge_attributes = {})
       base = base_attributes.stringify_keys.merge(merge_attributes.stringify_keys)
+      base.delete("name") unless Spree::Config.use_combined_first_and_last_name_in_address
```
kennyadsl, line 87: if `use_combined_first_and_last_name_in_address` is false, if I recall correctly `name` should not be part of the attributes provided in an address, but maybe I'm missing something.

kennyadsl (commented):
Left a question to better understand the context where this is a problem. Thanks Thomas!

core/app/models/spree/address.rb
tvdeyen: We have custom forms that only have `firstname` and `lastname` as attributes. But the same is true in `solidus_frontend` v2.11.14 https://github.com/solidusio/solidus/blob/v2.11/frontend/app/views/spree/address/_form.html.erb#L8-L18 btw. 

Somehow Solidus is setting `name` before we reach this code and if `name` is given it will not be updated. See L97-99 below. Maybe there is a better fix, but this is the easiest with the least impact that fixes this issue. Credit to @cpfergus1

core/spec/models/spree/address_spec.rb
```diff
@@ -119,6 +119,47 @@
...
+      end
+
+      context "and the `name` attribute is not in changeset" do
+        let(:address) { create(:address, firstname: 'John', lastname: 'Doe') }
+
+        it "sets `name` from `firstname` and `lastname`" do
+          new_address = described_class.immutable_merge(address, firstname: 'Jane', lastname: 'Fonda')
+          expect(new_address.read_attribute(:name)).to eq('Jane Fonda')
+        end
+      end
```
kennyadsl, line 135: Hey @tvdeyen, did you intentionally leave out the `"and the `name` attribute is in changeset"` context here? Isn't that exactly the scenario we want to fix?

core/spec/models/spree/address_spec.rb
```diff
@@ -119,6 +119,47 @@
...
+      before do
+        allow(Spree::Config).to receive(:use_combined_first_and_last_name_in_address) { true }
+      end
+
+      context "and the `name` attribute is not in changeset" do
+        let(:address) { create(:address, firstname: 'John', lastname: 'Doe') }
+
+        it "keeps old name" do
+          new_address = described_class.immutable_merge(address, firstname: 'Jane', lastname: 'Fonda')
+          expect(new_address.read_attribute(:name)).to eq('John Von Doe')
```
kennyadsl, line 148: Is `John Von Doe` coming from the address factory?

core/spec/models/spree/address_spec.rb
```diff
@@ -119,6 +119,47 @@
...
+        end
+      end
+
+      context "and the `name` attribute is in changeset" do
+        let(:address) { create(:address, firstname: 'John', lastname: 'Doe') }
+
+        it "keeps old name" do
```
kennyadsl, line 155: Reading the test expectation, it doesn't seem the old name is kept, or am I missing something?

kennyadsl (commented):
I'm ok with the fix, thanks, Thomas. Just left a couple of questions on the specs.

core/spec/models/spree/address_spec.rb
tvdeyen: No, we want to test exactly that scenario. Given are only `firstname` and `lastname` in the params. See the issue https://github.com/solidusio/solidus/issues/4094

core/spec/models/spree/address_spec.rb
tvdeyen: Yes https://github.com/solidusio/solidus/blob/master/core/lib/spree/testing_support/factories/address_factory.rb#L18

core/spec/models/spree/address_spec.rb
tvdeyen: Yes, you are right. Fixing

tvdeyen: fixed

core/spec/models/spree/address_spec.rb
kennyadsl: Ah got it, so essentially `name` is there because it is part of the address already (being "in sync" with firstname/lastname).

What do you think about adding the other context as well (passing the `name` explicitly with `use_combined_first_and_last_name_in_address: false`)? I understand that this should never happen but at least we have a spec that describe the state of the system in that case.

kennyadsl (approved):
Left one non-blocking comment. Thanks for taking care of this, @tvdeyen!

jarednorman (approved):

core/spec/models/spree/address_spec.rb
```diff
@@ -119,6 +119,47 @@
...
   end
 
+  context "when updating a record" do
+    context "if use_combined_first_and_last_name_in_address is set to false (default)" do
+      before do
+        allow(Spree::Config).to receive(:use_combined_first_and_last_name_in_address) { false }
+      end
+
+      context "and the `name` attribute is not in changeset" do
+        let(:address) { create(:address, firstname: 'John', lastname: 'Doe') }
```
spaghetticode, line 129: given every context includes this line, what about moving it one `context` above, so we remove the duplication?

spaghetticode (approved):
Looks good to me, only left a minor cleanup note but not blocking... thanks!

core/spec/models/spree/address_spec.rb
tvdeyen: Sure!

core/spec/models/spree/address_spec.rb
tvdeyen: Done

kennyadsl (approved):

PR #4225, opened 2021-12-16
Revert 21300973408b1b02b67d7d05cffaccab8c6d489c

tvdeyen:
In Solidus v2.11 we still use `false` for this setting.
Using `true` in specs per default is giving false positives
and issues like #4094 where not found.

**Description**

**Checklist:**

### DIFF
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -123,7 +123,6 @@ Spree.config do |config|
   config.raise_with_invalid_currency = false
   config.redirect_back_on_unauthorized = true
   config.run_order_validations_on_order_updater = true
-  config.use_combined_first_and_last_name_in_address = true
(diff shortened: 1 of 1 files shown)


tvdeyen:
expected to fail

tvdeyen:
@solidusio/core-team what do we want to do about this? v2.11 tests fail if the dummy app runs with Solidus v2.11 defaults.

kennyadsl:
@tvdeyen I'd say that we should change the code so that specs pass for that version. Did you already have a look at what's needed?

tvdeyen:
@kennyadsl I agree we should fix the code. I did not had time to look into it yet.

kennyadsl:
I don't recall if this has been fixed by https://github.com/solidusio/solidus/pull/4224. BTW, we no longer support 2.11 at this time. I think we can close this @tvdeyen.

PR #4226, opened 2021-12-21
Enforce event registration

waiting-for-dev:
**Description**

Make it mandatory to register an event before being fired, subscribed
, or unsubscribed.

It helps with two scenarios:

- It avoids typo errors, like subscribing to an event `ordr_finalized`
  instead of `order_finalized`.

- It avoids naming collisions between user-defined events and core ones.

Example:

```ruby
Spree::Event.register('foo')
Spree::Event.fire('foo')
```

We add this behavior only to the new, recommended adapter. However, we
add into the deprecation error for the legacy adapter the instructions
to update.

We also add a new `#unregister` method for the test interface.

**Checklist:**

### DIFF
--- a/core/lib/spree/event.rb
+++ b/core/lib/spree/event.rb
@@ -5,6 +5,7 @@ require_relative 'event/configuration'
 require_relative 'event/listener'
 require_relative 'event/subscriber_registry'
 require_relative 'event/subscriber'
+require 'spree/deprecation'
 
 module Spree
   # Event bus for Solidus.
@@ -12,6 +13,12 @@ module Spree
   # This module serves as the interface to access the Event Bus system in
   # Solidus. You can use different underlying adapters to provide the core
   # logic. It's recommended that you use {Spree::Event::Adapters::Default}.
+  #
+  # Before firing, subscribing, or unsubscribing an event, you need to
+  # {#register} it:
+  #
+  # @example
+  #   Spree::Event.register 'order_finalized'
   #
   # You use the {#fire} method to trigger an event:
   #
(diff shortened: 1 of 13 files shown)


core/lib/spree/event.rb
```diff
@@ -43,6 +50,34 @@ module Event
...
+    #
+    # @example
+    #   Spree::Event.register('foo')
+    #
+    # @param [String, Symbol] event_name
+    # @param [Any] adapter the event bus adapter to use.
+    def register(event_name, adapter: default_adapter)
+      warn_registration_on_legacy_adapter if deprecation_handler.render_deprecation_message?(adapter)
+      return if deprecation_handler.legacy_adapter?(adapter)
+
```
aldesantis, line 68: Extra blank line?

core/lib/spree/event/configuration.rb
```diff
@@ -59,6 +59,13 @@ def adapter
               order.do_something
               Spree::Event.fire 'event_name', order: order
 
+          - You need to register your custom events before firing or subscribing
+            to them (no need for Solidus' custom ones). Example:
```
aldesantis, line 63: Solidus' native ones?

aldesantis (approved):
😍

core/lib/spree/event/configuration.rb
jarednorman: Perhaps "not necessary for events provided by Solidus or other extensions" would be more clear.

jarednorman (approved):
Looks good, other than the slightly confusing wording that Alessandro pointed out.

core/lib/spree/event.rb
waiting-for-dev: Thanks! Removed.

core/lib/spree/event/configuration.rb
waiting-for-dev: Fixed, I used @jarednorman's suggestion :+1:

core/lib/spree/event/registry.rb
```diff
@@ -0,0 +1,93 @@
...
+            You can register the new event with:
+
+              Spree::Event.register('#{event_name}')
+        MSG
+      end
+
+      private
+
+      def suggestions(event_name)
+        dictionary = DidYouMean::SpellChecker.new(dictionary: event_names)
```
kennyadsl, line 83: 👍

kennyadsl (approved):
Thanks, Marc!

I have one note: it is not quite clear to me what is the suggested approach at registering events. Considering that people may fire events multiple times in their codebase, is the initializer the right place where people should place these `Spree::Event.register` calls? Maybe worth mentioning or providing an example?

waiting-for-dev:
> it is not quite clear to me what is the suggested approach at registering events.

Good point. I'm checking now, and I'm finding some issues because of the loading of module subscribers. Please, put it on hold until I have a clearer picture.

waiting-for-dev:
> Good point. I'm checking now, and I'm finding some issues because of the loading of module subscribers. Please, put it on hold until I have a clearer picture.

@kennyadsl, no significant issues in the end.

An event must be registered before being fired/subscribed, etc. In theory, that should mean that we can register them anywhere in the code before that. However, we have to add module subscribers into the equation. They're [loaded in a `#to_prepare` block](https://github.com/solidusio/solidus/blob/fda6073aaefbc91ac1b8f6fd48ac9e963340d88e/core/lib/spree/core/engine.rb#L49-L51). That runs before the application code is loaded but after the initializers.

That means that we must register them on an initializer. However, the initializer must run after the [adapter has been configured](https://github.com/solidusio/solidus/blob/fda6073aaefbc91ac1b8f6fd48ac9e963340d88e/core/lib/generators/solidus/install/templates/config/initializers/spree.rb.tt#L37). As Rails loads initializer in alphabetical order, one called `spree_events.rb` would do the trick, but `events.rb` wouldn't. For that reason, I think it's simpler to recommend adding the registrations at the end of the `spree.rb` file. It's what I've done, updating the error and warning messages, as you pointed out.

kennyadsl:
@waiting-for-dev thanks, it works for me.

PR #4227, opened 2021-12-21
Reintroduce inverse_of: :product for variants association

spaghetticode:
**Description**

`inverse_of: :product` was moved to the `variants_including_master` association with 7c37093, but there should be no problem in having this option specified in more than one association, in fact its removal is causing an issue as now the reference to the product changes is lost for records in the `variants` association:

    p = Spree::Product.first
    v = p.variants.first
    m = p.variants_including_master.find_by(is_master: true)

    p.name = "Foobar"

    m.product.name
     # => "Foobar"

    v.product.name
    # => "Solidus T-Shirt"

**Checklist:**

### DIFF
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -45,6 +45,7 @@ module Spree
 
     has_many :variants,
       -> { where(is_master: false).order(:position) },
+      inverse_of: :product,
       class_name: 'Spree::Variant'
 
     has_many :variants_including_master,


waiting-for-dev (approved):
I can't recall anymore why I did that change, but I'd swear it was because of some issue. However, CI is green! I'm ok with merging that and revisiting if we find a problem later on :slightly_smiling_face:

jarednorman (approved):
Seems fine if it doesn't cause issue. If we find an issue we'll just have to make sure we add a spec for it.

PR #4228, opened 2021-12-24
Ensure promotion codes don't trigger ActiveRecord::RecordNotUnique errors on save

jcsanti:
Promotion codes are always stored in the database in lowercase. Up to this PR, the transformation of the in-memory value of a code was performed in a [`before_save`][before_save] callback.

This implementation led to a weird behavior that can easily be reproduced with a fresh Solidus install:

https://user-images.githubusercontent.com/4389323/147362573-a615fcc6-0013-4f02-87cc-46c9d5cb78c9.mov

The expected error handling would be that any input that would end up matching an existing code (after having been stripped of spaces & downcased) should trigger the same flash error rather than crash the page.

This behavior can be achieved by normalizing values earlier in a `before_validation` callback: that way, the validation callback will query the database using a stripped/downcased value and, should a match be found, set an error on the model instance as expected.

**Description**

**Checklist:**

[before_save]: https://github.com/solidusio/solidus/blob/d467123b9b9b303a68b2817c21ca4c72b694f3b0/core/app/models/spree/promotion_code.rb#L12

### DIFF
--- a/core/app/models/spree/promotion_code.rb
+++ b/core/app/models/spree/promotion_code.rb
@@ -5,11 +5,11 @@ class Spree::PromotionCode < Spree::Base
   belongs_to :promotion_code_batch, class_name: "Spree::PromotionCodeBatch", optional: true
   has_many :adjustments
 
+  before_validation :normalize_code
+
   validates :value, presence: true, uniqueness: { allow_blank: true, case_sensitive: true }
   validates :promotion, presence: true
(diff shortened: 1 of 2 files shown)


waiting-for-dev (approved):
Thanks for your contribution and the detailed explanation :heart:

kennyadsl (approved):
Thank you @jcsanti!

PR #4238, opened 2022-01-04
Admin users did not work with custom user models

softr8:
# Description
If Spree.user_class was other than Spree::User, it was failing
to create edit and delete url's in Admin::UsersController#index,
using CustomUser for example, it tried to generate: admin_custom_user_url
instead of using the existing admin_user_url

# Checklist:

### DIFF
--- a/backend/app/views/spree/admin/store_credits/index.html.erb
+++ b/backend/app/views/spree/admin/store_credits/index.html.erb
@@ -70,6 +70,6 @@
   <div class="no-objects-found">
     <%= render 'spree/admin/shared/no_objects_found',
                  resource: Spree::StoreCredit,
-                 new_resource_url: new_object_url %>
+                 new_resource_url: spree.new_admin_user_store_credit_url %>
   </div>
 <% end %>
--- a/backend/spec/controllers/spree/admin/users_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/users_controller_spec.rb
@@ -59,6 +59,33 @@ describe Spree::Admin::UsersController, type: :controller do
         expect(assigns(:collection)).to eq [user]
       end
     end
+
+    context "when Spree.user_class have a different namespace than Spree" do
+      class UserModel < ApplicationRecord
+        self.table_name = 'spree_users'
+        include Spree::UserMethods
+      end
+
+      around do |example|
+        actual_user_class = Spree.user_class
+        Spree.user_class = 'UserModel'
+        UserModel.create(email: "a@solidus.io")
+        example.run
+        Spree.user_class = actual_user_class.name
+      end
+
+      render_views
+
+      it "renders the edit and delete links correctly" do
+        allow(Spree.user_class).to receive(:find_by).
+            with(hash_including(:spree_api_key)).
+            and_return(Spree.user_class.new)
+
+        get :index
+
+        expect(response).to be_successful
+      end
+    end
   end
 
   context "#show" do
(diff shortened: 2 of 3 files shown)


backend/spec/controllers/spree/admin/users_controller_spec.rb
```diff
@@ -59,6 +59,30 @@
...
     end
+
+    context "when Spree.user_class have a different namespace than Spree" do
+      class UserModel < ApplicationRecord
+        self.table_name = 'spree_users'
+        include Spree::UserMethods
+      end
+
+      before do
+        Spree.user_class = 'UserModel'
```
peterberkenbosch, line 70: I think we should do an after block here too to make sure subsequent tests are using the original User.

jarednorman (approved):

backend/spec/controllers/spree/admin/users_controller_spec.rb
```diff
@@ -59,6 +59,33 @@
...
+        self.table_name = 'spree_users'
+        include Spree::UserMethods
+      end
+
+      around do |example|
+        Spree.user_class = 'UserModel'
+        UserModel.create(email: "a@solidus.io")
+        example.run
+        Spree.user_class = 'Spree::LegacyUser'
+      end
```
waiting-for-dev, line 74: Can we take the old value of `user_class` and reset it later instead of hardcoding the value that we are using now? I.e.:


```suggestion
        actual_user_class = Spree.user_class
        Spree.user_class = 'UserModel'
        UserModel.create(email: "a@solidus.io")
        example.run
        Spree.user_class = actual_user_class
      end
```

backend/app/views/spree/admin/store_credits/index.html.erb
```diff
@@ -70,6 +70,6 @@
   <div class="no-objects-found">
     <%= render 'spree/admin/shared/no_objects_found',
                  resource: Spree::StoreCredit,
-                 new_resource_url: new_object_url %>
+                 new_resource_url: spree.new_admin_user_store_credit_url %>
```
waiting-for-dev, line 73: I don't think it's bad to reference the URL helper directly, but it seems that the `new_object_url` shortcut is used extensively in the backend. Besides, it doesn't seem related to this PR. Can you help me understand why we need this change?

waiting-for-dev (changes requested):

backend/app/views/spree/admin/store_credits/index.html.erb
softr8: it's the same issue than the one in users/index, it's a nested resource and it tries to build the url with two AR objects, so if the model that represents Spree.user_class is different than `user`, it does not match anymore with the routes.

`Spree.user_class = 'MyStore::Admin'` is going to generate new_admin_admin_store_credit_url, because it reads the model name, so, that's why we have to 'hardcode' the route name

waiting-for-dev: Thanks!

backend/spec/controllers/spree/admin/users_controller_spec.rb
```diff
@@ -59,6 +59,33 @@
...
     end
+
+    context "when Spree.user_class have a different namespace than Spree" do
+      class UserModel < ApplicationRecord
+        self.table_name = 'spree_users'
+        include Spree::UserMethods
+      end
+
+      around do |example|
+        actual_user_class = Spree.user_class
```
waiting-for-dev, line 70: Oh, my fault. This should fix the CI, right?
```suggestion
        actual_user_class = Spree.user_class.name
```

waiting-for-dev (approved):
Thanks!!

PR #4239, opened 2022-01-07
[RFC] Just touch taxons in Product#touch_taxons

### LINKED ISSUE #3931: `Spree::Product` touching `Spree::Taxon` may cause deadlocks
Solidus V2.10.1

Why do **products** touch **taxons** when touched? 

https://github.com/solidusio/solidus/blob/b0092d0f3bda447494a586bf8eb96a18322b8bed/core/app/models/spree/product.rb#L348-L357

I noticed that in our deployment where we have multiple operators simultaneously reviewing products as well as async jobs creating products (many of which share the same `Spree::Taxon`) this operation seems to be related to (or causing) some of our DB deadlocks. 

```
PG::TRDeadlockDetected: ERROR:  deadlock detected
DETAIL:  Process 18333 waits for ShareLock on transaction 26056034; blocked by process 29223.
Process 29223 waits for ShareLock on transaction 26056296; blocked by process 18333.
HINT:  See server log for query details.
CONTEXT:  while rechecking updated tuple (17,44) in relation "spree_taxons"
```

I am considering removing this operation on our end. However, if this is legacy would it be appropriate to remove it from here as well?

A similar problem is occurring with the touching of Spree::OptionType.

tvdeyen:
**Description**

And make use of the `after_touch` callback of `Spree::Taxon`,
that already handles touching ancestors and its taxonomy.

This hopefully fixes #3931 as well, because it runs each `touch`
callback in it's own transaction instead of one large transaction
that might cause dead locks.

This IS slower than the optimized approach before, but this should not
be that much of an issue, since we are updating a product here
and this will a) not happen that much and b) there will not be many taxons
one single product will be attached to, right? Even in large stores this
will most likely be in single to double digits, instead of hundreds
or thousands. But I might be painfully wrong here.

This also has the advantage that a `after_touch` callback a store or an
extension might have will be triggered and caches will be invalidated
correctly.

~~DISCLAIMER: I did not added any tests on purpose, because I want to discuss if this is an approach we want to try first.~~

~~Actually I added a spec that creates a taxon tree of 1000 taxons nested 10 random levels deep. The product that then will be updated is attached to 50 random taxons. We expect no errors.~~

Removed that spec again

**Checklist:**

### DIFF
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -115,7 +115,7 @@ module Spree
     after_initialize :ensure_master
 
     after_save :run_touch_callbacks, if: :saved_changes?
-    after_touch :touch_taxons
+    after_commit :touch_taxons
 
     before_validation :normalize_slug, on: :update
     before_validation :validate_master
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -349,15 +349,11 @@ module Spree
       run_callbacks(:touch)
     end
 
-    # Iterate through this product's taxons and taxonomies and touch their timestamps in a batch
+    # Iterate through this product's taxons and touch their timestamps in a batch
+    # The `after_touch` callback of `Spree::Taxon` will make sure the ancestors and
+    # the taxonomy will be touched as well.
     def touch_taxons
-      taxons_to_touch = taxons.flat_map(&:self_and_ancestors).uniq
-      unless taxons_to_touch.empty?
-        Spree::Taxon.where(id: taxons_to_touch.map(&:id)).update_all(updated_at: Time.current)
-
-        taxonomy_ids_to_touch = taxons_to_touch.flat_map(&:taxonomy_id).uniq
-        Spree::Taxonomy.where(id: taxonomy_ids_to_touch).update_all(updated_at: Time.current)
-      end
+      taxons.each(&:touch)
     end
 
     def remove_taxon(taxon)
--- a/core/app/models/spree/taxon.rb
+++ b/core/app/models/spree/taxon.rb
@@ -23,8 +23,7 @@ module Spree
     validates :meta_title, length: { maximum: 255 }
     validates :taxonomy_id, uniqueness: { scope: :parent_id, message: :can_have_only_one_root }, if: -> { root? }
 
-    after_save :touch_ancestors_and_taxonomy
-    after_touch :touch_ancestors_and_taxonomy
+    after_commit :touch_ancestors_and_taxonomy
 
     include ::Spree::Config.taxon_attachment_module
 
--- a/core/app/models/spree/taxon.rb
+++ b/core/app/models/spree/taxon.rb
@@ -140,11 +140,11 @@ module Spree
 
     private
 
+    # Touches all ancestors at once to avoid recursive taxonomy touch, and reduce queries.
     def touch_ancestors_and_taxonomy
-      # Touches all ancestors at once to avoid recursive taxonomy touch, and reduce queries.
-      self.class.default_scoped.where(id: ancestors.pluck(:id)).update_all(updated_at: Time.current)
+      ancestors.touch_all
       # Have taxonomy touch happen in #touch_ancestors_and_taxonomy rather than association option in order for imports to override.
-      taxonomy.try!(:touch)
+      taxonomy&.touch
     end
   end
 end


kennyadsl:
@tvdeyen Thanks for taking care of this! 

I'm seeing the test suite for this PR is taking ~1 hour compared to the ~10 minutes of the other builds. Even if the single spec added takes ~5 seconds, I'm concerned that this change added a significant amount of time by getting performance worse on every single test that interacts with products and taxons. If that's the case, I don't think we should merge this as is. 

On a side note, did you consider providing this new behavior optional with a preference? Maybe we could have the past approach as default, with optimization in mind and the new one, more reliable in terms of deadlock as optional. 

I know of stores that have products constantly updated by external tools (for content or inventory management sync), in these cases we will trigger a lot of these touches and I'm afraid that this version would significantly slow down their operations (assuming the current version is barely acceptable for them, which I'm not sure).

jarednorman (dismissed):
This looks good, but I'd still like to move the `touch_taxons` call to an `after_commit`. I don't believe it must be in a transaction.

core/spec/models/spree/product_touch_taxons_spec.rb
```diff
@@ -33,5 +33,9 @@
       expect(product.taxons.count).to eq(50)
       expect { subject }.to_not raise_error
     end
+
+    it "is fast" do
+      puts Benchmark.measure { subject }
```
jarednorman, line 38: 🏎️

tvdeyen:
> I'm seeing the test suite for this PR is taking ~1 hour compared to the ~10 minutes of the other builds. Even if the single spec added takes ~5 seconds, I'm concerned that this change adds a significant amount of time by getting performance worse on every single test that interacts with products and taxons. If that's the case, I don't think we should merge this as is.

@kennyadsl The spec that I added took 5647.54 (1.5h) to run under postgresql on CircleCI. In mysql it finished in 587.86s (9 minutes).
From https://app.circleci.com/pipelines/github/solidusio/solidus/2926/workflows/c1e0911a-07a7-4e2b-b4b8-23af32c3e60f/jobs/27313

```
Top 10 slowest example groups:
  Spree::Product
    5950.8 seconds average (5950.8 seconds / 1 example) ./spec/models/spree/product_touch_taxons_spec.rb:3
```

Most of the time is spent creating 1000 taxons, nested 10 levels deep. The actual spec of updating 50 taxons (and their ancestors) runs in 8ms on my machine.

**Disclaimer:** I never intended to keep the spec. I just wanted to verify it does not cause any deadlocks even with unrealistically high numbers of taxons. Locally this spec runs with postgres in under 10 seconds. I am not sure why this takes so long on CircleCI.

> On a side note, did you consider providing this new behavior optional with a preference? Maybe we could have the past approach as default, with optimization in mind and the new one, more reliable in terms of deadlock as optional.

I thought about this, yes. But is this actually necessary? See the note below.

> I know of stores that have products constantly updated by external tools (for content or inventory management sync), in these cases we will trigger a lot of these touches and I'm afraid that this version would significantly slow down their operations (assuming the current version is barely acceptable for them, which I'm not sure).

My test shows that this should not have an impact on these stores, because even though we touch a lot of products, how many taxons are these products attached to? 1, 2, 10? My test assigns a product to 50 taxons (very unlikely for real life stores IMO) and it takes ~8ms to touch these 50 taxons (and their 10 ancestors).

```
0.000748   0.000021   0.000769 (  0.000764)
```

_These numbers are from an 8 core Apple M1 Pro with 32GB of RAM_. But even if they would be 10 x slower it still would only take 80ms to update 50 taxons. Most products I know of are assigned to 1-5 taxons.

If we are still concerned (and honestly the CircleCI times concern me), we could add the new behavior as an opt-in. For my use-case this would be fine. But shouldn't we ship "reliable" as the default option then? And fast/unreliable as opt-in?

> This looks good, but I'd still like to move the touch_taxons call to an after_commit. I don't believe it must be in a transaction.

@jarednorman I agree, but does a call to `product.touch` also triggers an `after_commit`?

jarednorman:
I would assume it does, but we can check.

jarednorman:
I've confirmed that `after_commit` callbacks run on touch.

kennyadsl:
Hey @tvdeyen! I think we can try to move this forward now. What about rebasing to try to have the test suite green? After we verified that there are no deadlocks, we can remove the slow "benchmark" specs and merge the PR. Sounds good?

tvdeyen:
> I've confirmed that `after_commit` callbacks run on touch.

Unfortunately `after_commit` fails current specs. It seems that it runs no matter the case if changes happened or not. I tried `if: :saved_changes?` but no luck as well. (See failing specs in this PR). Maybe it is related to this unresolved bug in Rails 7.1? https://github.com/rails/rails/issues/51046

Any idea how to get it working with `after_commit` instead of `after_touch`?

> Hey tvdeyen! I think we can try to move this forward now. What about rebasing to try to have the test suite green? After we verified that there are no deadlocks, we can remove the slow "benchmark" specs and merge the PR. Sounds good?

I rebased and removed the benchmark spec.

tvdeyen:
I am still not sure about this one. Let's re-visit this problem if it emerges again someday

PR #4243, opened 2022-01-13
Fix product prices table pagination

tvdeyen:
**Description**

- Move the pagination into the fieldset of its table,
  makes it more easily to distinguish the buttons
- Paginate prices of variants and products independently
- Scopes pricing table headers to the template path. 
  That way it can better be translating in context.

### Before

localhost_3000_admin_products_solidus-long-sleeve_prices(Full-HD TV)

### After

localhost_3000_admin_products_solidus-long-sleeve_prices_variants_page=3(Full-HD TV)

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/prices/_master_variant_table.html.erb
+++ b/backend/app/views/spree/admin/prices/_master_variant_table.html.erb
@@ -1,6 +1,6 @@
 <fieldset class="no-border-bottom <%= "no-border-top" if !variants %>">
   <% if variants %>
-    <legend align="center"><%= I18n.t(:master_variant, scope: :spree) %> <%= admin_hint I18n.t(:master_variant, scope: :spree), I18n.t(:master_variant, scope: [:spree, :hints, "spree/price"]) %></legend>
+    <legend align="center"><%= t(".master_variant") %> <%= admin_hint t(".master_variant"), I18n.t(:master_variant, scope: [:spree, :hints, "spree/price"]) %></legend>
   <% end %>
 
   <%= paginate master_prices, theme: "solidus_admin" %>
--- a/backend/app/views/spree/admin/prices/_master_variant_table.html.erb
+++ b/backend/app/views/spree/admin/prices/_master_variant_table.html.erb
@@ -1,9 +1,9 @@
-<%= paginate master_prices, theme: "solidus_admin" %>
-
 <fieldset class="no-border-bottom <%= "no-border-top" if !variants %>">
   <% if variants %>
     <legend align="center"><%= I18n.t(:master_variant, scope: :spree) %> <%= admin_hint I18n.t(:master_variant, scope: :spree), I18n.t(:master_variant, scope: [:spree, :hints, "spree/price"]) %></legend>
   <% end %>
+
+  <%= paginate master_prices, theme: "solidus_admin" %>
   <table class="index master_prices">
     <colgroup>
       <col style="width: 30%">
@@ -36,6 +36,5 @@
       </tr>
     <% end %>
   </table>
+  <%= paginate master_prices, theme: "solidus_admin" %>
 </fieldset>
-
-<%= paginate master_prices, theme: "solidus_admin" %>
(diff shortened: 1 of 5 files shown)


mamhoff (approved):
I like this.

jarednorman (approved):

waiting-for-dev (approved):

backend/app/views/spree/admin/prices/_master_variant_table.html.erb
```diff
@@ -1,9 +1,9 @@
-<%= paginate master_prices, theme: "solidus_admin" %>
-
 <fieldset class="no-border-bottom <%= "no-border-top" if !variants %>">
   <% if variants %>
-    <legend align="center"><%= I18n.t(:master_variant, scope: :spree) %> <%= admin_hint I18n.t(:master_variant, scope: :spree), I18n.t(:master_variant, scope: [:spree, :hints, "spree/price"]) %></legend>
+    <legend align="center"><%= t(".master_variant") %> <%= admin_hint t(".master_variant"), I18n.t(:master_variant, scope: [:spree, :hints, "spree/price"]) %></legend>
```
spaghetticode, line 3: Can you remove also the last useless`I18n`? 
```suggestion
    <legend align="center"><%= t(".master_variant") %> <%= admin_hint t(".master_variant"), t(:master_variant, scope: [:spree, :hints, "spree/price"]) %></legend>
```

spaghetticode (approved):
@tvdeyen I left a comment for a small change but looks good!

backend/app/views/spree/admin/prices/_master_variant_table.html.erb
tvdeyen: I don't see any benefit in doing that. We are not using any features of the `t` view helper here.

spaghetticode: I'd surely agree the benefit is minimal( it's about being less verbose, having a more coherent codebase, and following the best practice), but it doesn't make much difference, and it can be done later. 

The main reason I asked is because we are already editing that line and you already removed two of the three occurrences of `I18n`, so why not finish the job.

tvdeyen: I used the `t` in these three occasions because I make use of its unique feature (the leading period for automatic templates path scoping of the translation key).

The other occasions in here use their own scope that is not template path based. We don't use the unique `t` helpers feature and that's why I don't think it is necessary and also not wrong in this case. There is nothing to finish imo, because nothing is missing here.

I also don't see the coherence argument here, because we use `I18n.t` in many places and the `t` helper in some views, where we make us of it's unique feature.

waiting-for-dev: However, `t` will work just like `I18n.t` when its unique features are not needed, right? If so, I'd replace all occurrences of the latest. It feels better to call a helper than a static method from another module. But, according to your comment, it wouldn't be consistent with the rest of the views. Then, I would replace them in all the views, but that's something that goes much beyond the scope of this PR :slightly_smiling_face:

tvdeyen: > However, t will work just like I18n.t when its unique features are not needed, right? 

More code, more potential Bugs, so no, I do not agree here.

> It feels better to call a helper than a static method from another module

I do not agree here as well. Rails view helpers are modules mixed into the ActionView instance. Way more potential for conflict and hard to track down bugs.

But, as you said, nothing we need to discuss on this PR. Thanks for the reviews and approvals.

spaghetticode: > I also don't see the coherence argument here, because we use I18n.t in many places and the t helper in some views, where we make us of it's unique feature.

I counted 12 occurrences of `I18n.t` in the backend and 2 in the FE views. We have *way* more occurrences of the `t` view helper (126 in the frontend and 693 in the BE). So yes, we're not being 100% consistent at this time, but quite close to it. Of course, topic for another PR/issue.

PR #4244, opened 2022-01-17
[Admin] Fix permission checks on some links

spaghetticode:
**Description**

In the admin area there are a few links, mostly to `edit` actions, that are not visible unless the user has `update` permissions on the resource. 
Unfortunately, this is preventing read-only users to access the resource data, since often the `edit` page acts as `show`, the latter being missing from routes, see for example [what happens for prices](https://github.com/solidusio/solidus/blob/9f54d19d1ae591dda3b9ad6fe4f1a389fa802a5a/backend/config/routes.rb#L56). 

On a different note, on the adjustments page the `Add coupon` button together with its text field should be visible only to users that have `update` permissions because showing it doesn't add any information to unprivileged users.

A further UX improvement left for the future may be to hide `Update` buttons on edit pages when the user doesn't have `update` abilities.

**Checklist:**

### DIFF
--- a/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
+++ b/backend/app/views/spree/admin/adjustment_reasons/index.html.erb
@@ -38,7 +38,9 @@
             </span>
           </td>
           <td class="actions">
-            <%= link_to_edit adjustment_reason, no_text: true %>
+            <% if can? :edit, adjustment_reason %>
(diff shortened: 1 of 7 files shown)


jarednorman (approved):

forkata (approved):
Thanks Andrea!

tvdeyen (approved):
Thanks 🙏🏻

PR #4245, opened 2022-01-17
Update image settings documentation

### LINKED ISSUE #3700: Not able to change images size
I am trying to change sizes with this in the initializer:
Spree::Image.attachment_definitions[:attachment][:styles] = {
  mini: '48x48>',
  small: '100x100>',
  product: '240x240>',
  large: '600x600>'
}
But 'module Spree::Image::ActiveStorageAttachment' don't take  new sizes. How to do it?

**Solidus Version**:
Last version of master

**To Reproduce**

nbelzer:
Resolves #3700

**Description**

As outlined [in my comment on #3700](https://github.com/solidusio/solidus/issues/3700#issuecomment-1012378830) the guides recommend the use of a code snippet to override the attachment styles used in `Spree::Image`.  However, b1fa1b897fc9eed74b2c5fe77f9b621c6e7daa2e introduced configuration options, that can be set from `Spree::Config` in an initializer like `config/initializers/spree.rb`.

This PR introduces updated documentation for the developer guides to reflect both the introduction of this configuration and the newly introduced Active Storage backend for `Spree::Image` in Solidus 3.0.

**Spec Tests**

To verify that this behavior is correct I have included a set of tests for both of the available backends for `Spree::Image`.

---

The developer guides also mention Paperclip as the reason for the system requirement of ImageMagick.  Perhaps this is also outdated as, since we are using Active Storage by default, it is also possible to use `libvips` (as outlined in the [Active Storage Requirements](https://guides.rubyonrails.org/active_storage_overview.html#requirements) as a faster and less memory intensive replacement for ImageMagick).  I believe we should therefore also update these requirements.

**Checklist:**
- ~~I have attached screenshots to this PR for visual changes (if needed)~~

### DIFF
--- a/guides/source/developers/getting-started/system-requirements.html.md
+++ b/guides/source/developers/getting-started/system-requirements.html.md
@@ -16,6 +16,7 @@ Homebrew][ruby-homebrew] if you need to upgrade from your system's Ruby.
 We also recommend configuring your development environment so that you can
 [install RubyGems without `sudo`][gem-install-without-sudo].
 
+[brew]: https://brew.sh
 [ruby-homebrew]: https://www.ruby-lang.org/en/documentation/installation/#homebrew
 [gem-install-without-sudo]: https://www.michaelehead.com/2016/02/06/installing-gems-without-sudo.html
 


nbelzer:
The same update to the guides should be made for the [EdgeGuides](https://github.com/solidusio/edgeguides), I found [this PR](https://github.com/solidusio/edgeguides/pull/39) which already includes these changes.

core/spec/models/spree/image/preferences_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    before { stub_spree_preferences(image_attachment_module: Spree::Image::ActiveStorageAttachment) }
+
+    it 'correctly sets the image styles' do
+      expect(Spree::Image.attachment_definitions[:attachment][:styles]).to eq(custom_styles)
+      expect(Spree::Image.attachment_definitions[:attachment][:default_style]).to eq(custom_default_style)
+    end
+  end
+
+  context 'When using Spree::Image::PaperclipAttachment as Spree::Config.image_attachment_module' do
+    before { stub_spree_preferences(image_attachment_module: Spree::Image::PaperclipAttachment) }
```
nbelzer, line 41: Similarly as to how the changes to `product_image_styles` are not reflected (because `Spree::Image` has already been loaded, the change in `image_attachment_module` is also ignored.  Therefore this test context runs through the same steps as the one for `Spree::Image::ActiveStorageAttachment`

nbelzer:
> To verify that this behavior is correct I have included a set of tests for both of the available backends for `Spree::Image`.  While they work in isolation they fail when running as part of the entire test suite. This is related to the autoloading of `Spree::Image` (by other tests) which causes the initialization of the `product_image_styles` and `product_image_style_default`.  If loaded before the newly added tests are run, changing the configuration using `stub_spree_preferences` has not effect (causing the tests to fail).  The most straightforward approach to solve this would be to reload the `Spree::Image` class, however, I have been unable to do so successfully and could use some help.

I was able to reload the `Spree::Image` class sucessfully using the `load` method, provided with the relative path to the definition of `Spree::Image`.  However, it does not reload the included `image_attachment_module` if that module has already been loaded before.

```ruby
load './app/models/spree/image.rb'
```

It moves the problem from having an outdated `Spree::Image` to having an outdated `Spree::Image::ActiveStorageAttachment` that is not reloaded when `Spree::Image` is.

guides/source/developers/products-and-variants/product-images.html.md
```diff
@@ -41,25 +41,29 @@ for a specific variant or for **All** variants.
 If set to **All**, the `viewable_id` is set to the master variant for the
 current product.
 
-## Paperclip settings
+## Image settings
 
-[Paperclip][paperclip-gem] handles the creation and storage of product images.
-By default, it creates several versions of each image at specific sizes.
+[Active Storage][activestorage] and [Paperclip][paperclip-gem] handle the
+creation and storage of product images.  By default, they creates several
```
kennyadsl, line 47: Not the endo of the world, but there's an extra space before `By`

kennyadsl (approved):
> The developer guides also mention Paperclip as the reason for the system requirement of ImageMagick. Perhaps this is also outdated as, since we are using Active Storage by default, it is also possible to use libvips (as outlined in the Active Storage Requirements as a faster and less memory intensive replacement for ImageMagick). I believe we should therefore also update these requirements.

@nbelzer thanks. 👆 please mention this as well.

guides/source/developers/products-and-variants/product-images.html.md
nbelzer: I replaced all the occurrences with a single space 👍 

This is something I do by default when writing.

nbelzer:
@kennyadsl [d739f87](https://github.com/solidusio/solidus/pull/4245/commits/d739f872dbc6d4de3e9682edb8064d08f8fba6df) includes an update to the requirements section that includes the changed requirements.  I made sure to mention that the changes are related to the chosen image backend and that stores that are using paperclip are still referred to using ImageMagick.

nbelzer:
We might also want to change the following lines to instead include libvips instead of imagemagick in the example brew install command:

https://github.com/solidusio/solidus/blob/6743f2e2bd9ad09a41391fe711f7a9f3303da132/guides/source/developers/getting-started/system-requirements.html.md#L25-L31

guides/source/developers/products-and-variants/product-images.html.md
```diff
@@ -41,25 +41,29 @@ for a specific variant or for **All** variants.
 If set to **All**, the `viewable_id` is set to the master variant for the
 current product.
 
-## Paperclip settings
+## Image settings
 
-[Paperclip][paperclip-gem] handles the creation and storage of product images.
-By default, it creates several versions of each image at specific sizes.
+[Active Storage][activestorage] and [Paperclip][paperclip-gem] handle the
+creation and storage of product images. By default, they creates several
```
spaghetticode, line 47:
```suggestion
creation and storage of product images. By default, they create several
```

nbelzer:
I have cleaned up the git history.

spaghetticode (approved):
@nbelzer thank you! 🙌

core/spec/models/spree/image/preferences_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  it 'correctly sets the image styles Paperclip' do
+    stub_spree_preferences(
+      product_image_styles: custom_styles,
+      product_image_style_default: custom_style_default
+    )
+
+    # We make use of a custom class, such that the preferences loaded
+    # are the mocked ones.
+    class Spree::PaperclipImage < Spree::Asset
```
waiting-for-dev, line 39: We can use an anonymous class here. This way, we won't leave a dangling top-level class for the rest of the test suite. I.e.:

```ruby
paperclip_image = Class.new(Spree::Asset) do
  # ...
end

expect(paperclip_image.attachment_defintions #...
```

core/spec/models/spree/image/preferences_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  it 'correctly sets the image styles ActiveStorage' do
+    stub_spree_preferences(
+      product_image_styles: custom_styles,
+      product_image_style_default: custom_style_default
+    )
+
+    # We make use of a custom class, such that the preferences loaded
+    # are the mocked ones.
+    class Spree::ActiveStorageImage < Spree::Asset
```
waiting-for-dev, line 23: Same as https://github.com/solidusio/solidus/pull/4245/files#r891032263

nbelzer:
Thanks for the suggestion @waiting-for-dev, I think that is a way better approach, learned something new today!

waiting-for-dev:
Thanks for your work, @nbelzer!

PR #4247, opened 2022-01-19
V2.11 - Fix non auto populated customer info

### LINKED ISSUE #3966: Customer info are not auto-populated correctly in admin creating a new order
When `use_combined_first_and_last_name_in_address` is set to `false`, creating a new order in admin, in the customer tab, it is not populating First Name and Last Name fields when you pick an existing customer from the autosuggestion dropdown.

**Solidus Version**:
v2.11.4

**To Reproduce**
1. Install Solidus at the specified version, 
2. Change `use_combined_first_and_last_name_in_address` to `false` in `spree.rb` initializer
3. Login in the admin panel
4. Go to Orders and click `New order`
5. Click `Customer` tab
6. Select any user 
7. `First Name` and `Last Name` are not populated

**Expected behavior**
`First Name` and `Last Name` fields of the auto-populated address are not filled with customer's info and left blank.

**Screenshots**
CleanShot 2021-03-01 at 12 20 44@2x

nbelzer:
Resolves #3966

**Description**
- Includes a fix for #3966.
- Includes spec tests that verify the render behavior of the Backbone View [Spree.Views.Order.Address](https://github.com/solidusio/solidus/blob/3fac51e1d6307eb5768f11eb87870699312eca5c/backend/app/assets/javascripts/spree/backend/views/order/address.js).

The spec tests verify the behavior of the Address Backbone View based on two fixtures, these are extracts from the partial `spree/admin/shared/address_form` as I was unable to render the partial itself (due to the country and state fields).  Because of this, the fixtures do not render the country and state fields, nor does the spec test verify that they are updated.  Since this is unrelated to the original issue I don't think this is a problem for this PR.

It looks like this issue also exists on Solidus v3.x, so perhaps we should also merge these changes into the master branch?  I would need some advice on how to do this correctly.

**Checklist:**
- ~~I have updated Guides and README accordingly to this change (if needed)~~
- ~~I have attached screenshots to this PR for visual changes (if needed)~~

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/views/order/address.js
+++ b/backend/app/assets/javascripts/spree/backend/views/order/address.js
@@ -1,5 +1,5 @@
 Spree.Views.Order.Address = Backbone.View.extend({
-  initialize: function(options) {
+  initialize: function() {
     // read initial values from page
     this.onChange();
 
--- a/backend/app/assets/javascripts/spree/backend/views/order/address.js
+++ b/backend/app/assets/javascripts/spree/backend/views/order/address.js
@@ -23,8 +23,9 @@ Spree.Views.Order.Address = Backbone.View.extend({
 
   eachField: function(callback){
     var view = this;
-    var fields = ["name", "company", "address1", "address2",
-      "city", "zipcode", "phone", "country_id", "state_name"];
+    var fields = ["name", "firstname", "lastname", "company", "address1",
+                  "address2", "city", "zipcode", "phone", "country_id",
+                  "state_name"];
     _.each(fields, function(field) {
       var el = view.$('[name$="[' + field + ']"]');
       if (el.length) callback(field, el);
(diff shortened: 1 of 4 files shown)


kennyadsl:
@nbelzer thanks for this!!

3.0 shouldn't have the `use_combined_first_and_last_name_in_address` preference. The default behavior is having it `true`, but I didn't verify that this feature is working correctly on those versions. If you can check it would be awesome.

nbelzer:
@kennyadsl You are correct, the preference does not exist on `Spree::AppConfiguration` and the issue also doesn't appear as the field shown is `Name` which was already included in the Address Backbone view:

https://github.com/solidusio/solidus/blob/7e0a386f4345629829dc47b4049100667dccab4f/backend/app/assets/javascripts/spree/backend/views/order/address.js#L24-L32

kennyadsl (approved):
Thanks @nbelzer!

jarednorman (approved):

spaghetticode (approved):
@nbelzer thank you!

PR #4250, opened 2022-01-26
fix backbone shipment item view when split

ccarruitero:
This PR is sponsored by [MagmaLabs](https://www.magmalabs.io/)

**Description**
This will fix current build for `solidus_product_assembly` https://app.circleci.com/pipelines/github/solidusio-contrib/solidus_product_assembly/194/workflows/a9e6929f-4d14-4234-bdcf-5266a359875a/jobs/464

Since we updated to use jquery3 in https://github.com/solidusio/solidus/commit/bdfa09409477d94198f5db96605600b93f066ef8, we need to update ajax callback to avoid use `success`, `error` and `complete` that are deprecated now https://api.jquery.com/jquery.ajax/

**Checklist:**

### DIFF
--- a/backend/spec/features/admin/orders/shipments_spec.rb
+++ b/backend/spec/features/admin/orders/shipments_spec.rb
@@ -142,4 +142,24 @@ describe "Shipments", type: :feature do
       end
     end
   end
+
+  context "when split shipment", js: true do
+    let(:location_name) { "WA Warehouse" }
+    let!(:location) { create(:stock_location, name: location_name)}
+
+    it "can split item" do
+      visit spree.edit_admin_order_path(order)
+      find_all('.split-item').first.click
+      within '.stock-item-split' do
+        find('.save-split').click
+        alert_text = page.driver.browser.switch_to.alert.text
+        expect(alert_text).to include 'Please select the split destination'
+        page.driver.browser.switch_to.alert.accept
+        find('.select2-container').click
+        find(:xpath, '//body').all('.select2-drop li.select2-result', text: "#{location_name} (0 on hand)")[1].click
+        find('.save-split').click
+      end
+      expect(page).to have_content /Pending package from '#{location_name}'/i
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


waiting-for-dev:
Should we add a regression test for that?

kennyadsl:
I second what @waiting-for-dev is asking. How didn't we notice that before? We are probably missing some specs for that part and if you could add it @ccarruitero, it would be great. Thanks!

backend/spec/features/admin/orders/shipments_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'spec_helper'
 
-describe "Shipments", type: :feature do
+describe "Shipments", type: :feature, js: true do
```
waiting-for-dev, line 5: Can we add this only to the new test addition? Tests that use javascript are slower, so it's better not to affect all of them.

backend/spec/features/admin/orders/shipments_spec.rb
```diff
@@ -142,4 +142,25 @@ def ship_shipment
...
+      within '.stock-item-split' do
+        find('.save-split').click
+        alert_text = page.driver.browser.switch_to.alert.text
+        expect(alert_text).to include 'Please select the split destination'
+        page.driver.browser.switch_to.alert.accept
+        find('.select2-container').click
+        find(:xpath, '//body').all('.select2-drop li.select2-result', text: "#{location_name} (0 on hand)")[1].click
+        find('.save-split').click
+      end
+      sleep 5
```
waiting-for-dev, line 162: We definitely should find a way to don't need to `sleep`. The reason is not making the suite slower.

waiting-for-dev (changes requested):

ccarruitero:
@waiting-for-dev I just updated PR according your comments. Please take a look when you have a chance

waiting-for-dev (approved):
Thanks! :rocket:

kennyadsl (approved):
Thanks, this is probably worth for a backport.

PR #4253, opened 2022-01-31
Fix for advancing carts correctly in admin checkout

tmtrademarked:
**Description:**

This PR adds a minor fix for cart behavior in Admin. When updating the customer details, the current code only calls `next`. This will move th order *one* state down the checkout flow - so, for instance, from `address` to `delivery`. But if the order could move further, it won't. And this can leave orders in a slightly unexpected state.

As an example, in our store, when the order is in the `delivery` state, there is no admin UI element to cause it to move into the `payment` state, short of selecting a different shipping method or adding a new item to the cart. And since for us, taxes aren't calculated until the order reaches the `payment` state, we can't render the correct cost in the payment detail admin page to allow our admins to enter the correct payment.

This fix is rather simple - rather than moving only *one* state, move as far as possible via `advance`. This essentially preserves the same behavior as before if the order could not advance, but allows the order to move forward into payment if the shipments all have shipping rates selected. By default, the default estimator should be selecting a shipping rate for every shipment when the shipments are proposed.

I tested this in my store with shipments which had multiple shipment rates, and observed no problem advancing to the `payment` state. We could potentially make this opt-in if desired, but that felt like slight overkill.

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/orders/customer_details_controller.rb
+++ b/backend/app/controllers/spree/admin/orders/customer_details_controller.rb
@@ -31,7 +31,7 @@ module Spree
             end
 
             unless @order.completed?
-              @order.next
+              @order.contents.advance
               @order.refresh_shipment_rates
             end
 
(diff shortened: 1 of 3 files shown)


tmtrademarked:
It seems like the remaining failure might be a flake of some sort - seems unlikely my change in a controller broke a core spec. Is anyone able to re-run that failing test?

tmtrademarked:
It seems like the tests are green - anyone able to review?

waiting-for-dev:
Seems legit, but I don't have enough background to confirm we aren't missing something. Thoughts, @kennyadsl?

kennyadsl (approved):
I think this makes total sense. Thanks @tmtrademarked!

waiting-for-dev (approved):

PR #4256, opened 2022-02-02
Wrap the checkout transitions within a database transaction

waiting-for-dev:
**Description**

Previous to this commit, the order's checkout flow transitions weren't
running, by default, within a database transaction (e.g.,
`order.complete!` wasn't safe). The reason is a limitation on the state
machine itself: When enabled, defined transitions, including all
callbacks, like `on_failure`, run within a database transaction (see
https://github.com/state-machines/state_machines-activerecord/blob/94662c61eeea4783b9d0b2899693c43610246900/lib/state_machines/integrations/active_record.rb#L261).

However, sometimes we need to make a persistent change on a record when
a transition fails. For instance, when we go from `confirm` to
`complete`, we need to roll back and update the state field back to
`payment` if payment can't be processed.

On this commit, we prepend a
`Spree::Core::StateMachines::Order::AfterTransactionOverrides` module to
`Spree::Order` to work around that limitation. The module is meant to
override the needed generated transition methods (only `complete` at
this point), delegating to them and performing any needed
post-transaction logic. For that, the failure reason needs to be
collected during the transition. Errors meant to be added to the record
need to be processed after-transition persistence, as it will
reset the record errors.

A more solid solution will imply breaking up data and persistence
responsibilities, taking complete control of the transition with
something like a service object.

The change in a test expectation that we can see on the diff is not an
actual change in business behavior. The final state was indeed
`payment`, previous to enabling transactions, but the cache needed to be
reloaded (`order.reload`).

https://github.com/nebulab/solidus/commit/cb1e1ff2611247cd213bca8a0aa24d7ee16c4440 is the commit were transactions
were disabled. The parent commit, https://github.com/nebulab/solidus/commit/bf0aca0d46cd92b12254d4c716f47db26b24a077,
reveals the hidden intention.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -735,8 +735,7 @@ module Spree
         true
       else
         saved_errors = errors[:base]
-        payment_failed!
-        saved_errors.each { |error| errors.add(:base, error) }
+        fail_transition(:payment_processing, saved_errors)
         false
       end
     end
@@ -773,11 +772,12 @@ module Spree
         !adjustment.calculate_eligibility
       end
       if adjustment_changed
-        restart_checkout_flow
-        recalculate
-        errors.add(:base, I18n.t('spree.promotion_total_changed_before_complete'))
+        errors = [I18n.t('spree.promotion_total_changed_before_complete')]
+        fail_transition(:promotion_eligibility, errors)
+        false
+      else
+        true
       end
-      errors.empty?
     end
 
     def validate_line_item_availability
--- a/core/spec/models/spree/order/state_machine_spec.rb
+++ b/core/spec/models/spree/order/state_machine_spec.rb
@@ -30,7 +30,7 @@ RSpec.describe Spree::Order, type: :model do
 
           it "should not complete the order" do
             expect(order.complete).to be false
-            expect(order.state).to eq("confirm")
+            expect(order.state).to eq("payment")
           end
         end
       end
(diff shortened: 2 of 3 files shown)


core/app/models/spree/order.rb
```diff
@@ -735,8 +735,7 @@ def process_payments_before_complete
         true
       else
         saved_errors = errors[:base]
-        payment_failed!
-        saved_errors.each { |error| errors.add(:base, error) }
+        fail_transition(:payment_processing, saved_errors)
```
kennyadsl, line 738: Did you consider passing a block to the method instead? This way we can keep the logic here here.

core/spec/models/spree/order/state_machine_spec.rb
```diff
@@ -30,7 +30,7 @@
 
           it "should not complete the order" do
             expect(order.complete).to be false
-            expect(order.state).to eq("confirm")
+            expect(order.state).to eq("payment")
```
kennyadsl, line 33: If I understood correctly, the order was in the payment state even before, but we were missing an `order.reload` to see that. What is not clear is why we don't need it anymore now. BTW, I think that a preparatory commit, in this case, would help a lot to understand what happens.

waiting-for-dev:
Thanks for your feedback, @kennyadsl. Let me put this PR on hold, as I'm doing related work with events, and maybe I'll change its direction altogether.

waiting-for-dev:
Closing in favor of #4262

PR #4262, opened 2022-02-10
Wrap the checkout transitions within a database transaction

waiting-for-dev:
**Description**

Previous to this commit, the order's checkout flow transitions weren't
running, by default, within a database transaction (e.g.,
`order.complete!` wasn't safe). The reason is a limitation on the state
machine itself: When enabled, defined transitions, including all
callbacks, like `on_failure`, run within a database transaction (see
https://github.com/state-machines/state_machines-activerecord/blob/94662c61eeea4783b9d0b2899693c43610246900/lib/state_machines/integrations/active_record.rb#L261).

However, sometimes we need to make a persistent change on a record when
a transition fails. For instance, when we go from `confirm` to
`complete`, we need to roll back and update the state field back to
`payment` if payment can't be processed.

This hack allows us to add a per-instance `after_rollback` callback.
We're creating an object that will quack like a record for that purpose,
adding it to the current transaction.

We prefer to leave the class definition inline to avoid giving
first-class citizenship to the hack. The sound solution will imply
breaking up data and persistence responsibilities, taking complete
control of the transition with something like a service object.

We considered another approach that consisted of monkey-patching the
state machine itself (see
https://github.com/solidusio/solidus/pull/4256). However, we'll need to
add another `after_commit` hook in upcoming work related to the firing
of events, which will also be required outside the state machine
transitions. Therefore, it's better to use a single type of hack for
everything.

cb1e1ff2611247cd213bca8a0aa24d7ee16c4440 is the commit were transactions
were disabled. The parent commit, bf0aca0d46cd92b12254d4c716f47db26b24a077,
reveals the hidden intention.

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -726,8 +726,10 @@ module Spree
         true
       else
         saved_errors = errors[:base]
-        payment_failed!
-        saved_errors.each { |error| errors.add(:base, error) }
+        after_rollback do
+          payment_failed!
+          saved_errors.each { |error| errors.add(:base, error) }
+        end
         false
       end
     end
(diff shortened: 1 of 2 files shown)


waiting-for-dev:
I am closing for now. After talking with the Core Team, we decided that implementing a solution that is so hacky could be risky.

PR #4265, opened 2022-02-11
Incorrect display store credit reason

### LINKED ISSUE #4264: Incorrect display of StoreCreditReason
There seems to be a layout bug for StoreCreditEvents, where the reason is displayed in a column that is given no width.  The column is also given an undescriptive name of 'name', where instead something like 'reason' would be more appropriate.

**Solidus Version**: 3.1.1

**To Reproduce**
1. Go to 'users' on the admin panel
2. Select a user and go to the 'Store Credits' tab
3. Add store credit
4. Edit the newly created store credit, for example by changing the amount, make sure to choose one of the reasons.
5. Voila, the chosen reason is now displayed as described.

**Current behavior**
Cramped column and non-descriptive name for the StoreCreditReason.

**Expected behaviour**
The column header should have a more descriptive name, such as 'reason'.  As well as be more easy to read by giving it some more room.
 
**Screenshots**
cramped-credit-event-showcase

Not applicable, see additional context below.

**Smartphone (please complete the following information):**
Not applicable, see additional context below.

**Additional context**
The issue is related to the following snippet, for every store credit event there are 7 columns defined while in the colgroup only 6 columns are defined.  In addition the name of the specific column for the store credit reason is taken from the `:name` attribute, which is why it results in 'name'.

https://github.com/solidusio/solidus/blob/68e97f35b6aafb0d69e1f998dcf4369a11a19aaa/backend/app/views/spree/admin/store_credits/show.html.erb#L68-L106

nbelzer:
Resolves #4264

**Description**
This change resolves a layout issue in store credit events table.  Here the last column is used to present the name of the `Spree::StoreCreditReason` associated with the event.  However, by default the column is given no width and therefore looks cramped.

In addition the column name used is not very descriptive, I've made use of an existing (but unused) locale key under `store_credits` to provide the column with a more descriptive name.

See #4264 for the before image.  See below for the after image.

cramped-credit-event-after

**Checklist:**
- ~I have updated Guides and README accordingly to this change (if needed)~
- ~I have added tests to cover this change (if needed)~

### DIFF
--- a/backend/app/views/spree/admin/store_credits/show.html.erb
+++ b/backend/app/views/spree/admin/store_credits/show.html.erb
@@ -85,7 +85,7 @@
         <th><%= t('spree.admin.store_credits.created_by') %></th>
         <th><%= Spree::StoreCreditEvent.human_attribute_name(:user_total_amount) %></th>
         <th><%= Spree::StoreCreditEvent.human_attribute_name(:amount_remaining) %></th>
-        <th><%= Spree::StoreCreditReason.human_attribute_name(:name) %></th>
+        <th><%= t('spree.admin.store_credits.reason_for_updating') %></th>
       </tr>
     </thead>
     <tbody>
--- a/backend/app/views/spree/admin/store_credits/show.html.erb
+++ b/backend/app/views/spree/admin/store_credits/show.html.erb
@@ -72,9 +72,10 @@
       <col style="width: 20%;" />
       <col style="width: 10%;" />
       <col style="width: 10%;" />
-      <col style="width: 25%;" />
+      <col style="width: 20%;" />
       <col style="width: 10%;" />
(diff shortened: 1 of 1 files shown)


spaghetticode (approved):
🎉  thanks Nick!

kennyadsl (approved):
👍 Thanks!!!

kennyadsl:
There's something with the CI though. Would you mind re-pushing to see if it starts this time?

nbelzer:
I updated the branch, but again there doesn't seem to be an update in the CI status.

kennyadsl:
@nbelzer I saw this happening to contributors that have set up their Solidus' fork project in CircleCI. In that case, CircleCI will try to run specs on that other project instead of the main one. Maybe it's that?

Issue #4270, opened 2022-02-15
Fix TestingSupport Country Factory states_required


RyanofWoods:
The mentioned attribute is commented out due to issues with Solidus' tests. 

https://github.com/solidusio/solidus/blob/a6c2ceeb2ee6bff832080ebb11773f7b6e4c2df0/core/lib/spree/testing_support/factories/country_factory.rb#L23-L24

This should be fixed so that anyone developers can use the factory and not have to modify the factory to get its intended behaviour.

**Solidus Version**:

1.2 and above.

**Current behavior**

The attribute will always become `false` as this is the attribute's default value.

**Expected behavior**

Use carmen to get the correct `states_required` value.

RyanofWoods:
WIP

kennyadsl:
This bugfix will be available from Solidus 3.3!

Issue #3842, opened 2020-11-13
Rework country/state handling around ISO 3166-2 subdivisions


elia:
## Is your feature request related to a problem? Please describe.

This is an umbrella issue on top of:
- https://github.com/solidusio/solidus/issues/1378
- https://github.com/solidusio/solidus/issues/3670
- https://github.com/solidusio/solidus/pull/3722
- https://github.com/solidusio/solidus/pull/3807
- https://github.com/solidusio/solidus/issues/3803

**After some research seems that in ISO terms, what Solidus is trying to map is _subdivisions_ as defined in ISO-3166-2** (https://en.wikipedia.org/wiki/ISO_3166-2, https://www.iso.org/iso-3166-country-codes.html#2012_iso3166-2).

As clearly noticed by some of the issues the term state is not right for most countries. In addition to that the way we're fetching data from Carmen might not match the correct _subdivision_ level, or we might be requiring a "state" presence when we shouldn't.

## Describe the solution you'd like

A path forward for this would be to fix the data generation basing it on the _subdivisions_ as defined in ISO 3166-2 and maybe  err on the side of not requiring the field presence unless we're 100% sure it's the right thing.

With time the database column should probably be migrated to a name that's more generic, like "subdivision" or "principal_subdivision".

A way of obtaining a list of all subdivisions should be found within Carmen (maybe contributing it back) or elsewhere. In the latter case the best candidate seems to be the `iso-codes` Debian package (https://salsa.debian.org/iso-codes-team/iso-codes/).

## Describe alternatives you've considered

I had a look at existing gems referring to ISO-3166, none of them seem reassuringly maintained or with the data we need, best bet would be to add that into Carmen itself, given its popularity.

## Additional context

On a similar topic would be great to have a pluggable system of address-validation with a storage that's enough generic to cover all the address variations of the world, possibly removing altogether the countries and states tables, delegating the whole thing outside of Solidus.

elia:
Related: https://github.com/countries/countries

AlistairNorman:
I think additional data about subdivisions will require some more discussion about how we would go about obtaining that data but in the meantime removing states_required from some countries during seeding and renaming seem like easy and positive changes to make.

PR #4272, opened 2022-02-15
Fix Country factory states_required attribute

RyanofWoods:
**Description**
Resolves issue: https://github.com/solidusio/solidus/issues/4270

Some tests were modified to give a state with the address parameters
as they used a country which will now get the states_required attribute
of true, meaning a state must be present on the address.

**Checklist:**

### DIFF
--- a/api/spec/requests/spree/api/orders_spec.rb
+++ b/api/spec/requests/spree/api/orders_spec.rb
@@ -523,14 +523,15 @@ module Spree::Api
       let(:billing_address) {
         { name: "Tiago Motta", address1: "Av Paulista",
                                 city: "Sao Paulo", zipcode: "01310-300", phone: "12345678",
-                                country_id: country.id }
+                                country_id: country.id, state_id: state.id }
       }
       let(:shipping_address) {
         { name: "Tiago Motta", address1: "Av Paulista",
                                  city: "Sao Paulo", zipcode: "01310-300", phone: "12345678",
-                                 country_id: country.id }
+                                 country_id: country.id, state_id: state.id }
       }
       let(:country) { create(:country, { name: "Brazil", iso_name: "BRAZIL", iso: "BR", iso3: "BRA", numcode: 76 }) }
+      let(:state) { create(:state, country_iso: 'BR')  }
 
       before { allow_any_instance_of(Spree::Order).to receive_messages user: current_api_user }
 
--- a/api/spec/requests/spree/api/users_spec.rb
+++ b/api/spec/requests/spree/api/users_spec.rb
@@ -44,6 +44,8 @@ module Spree::Api
 
       it "can update own details" do
         country = create(:country)
+        state = create(:state)
+
         put spree.api_user_path(user.id), params: { token: user.spree_api_key, user: {
           email: "mine@example.com",
           bill_address_attributes: {
@@ -51,7 +53,7 @@ module Spree::Api
             address1: '1 Test Rd',
             city: 'City',
             country_id: country.id,
-            state_id: 1,
+            state_id: state.id,
             zipcode: '55555',
             phone: '5555555555'
           },
@@ -60,7 +62,7 @@ module Spree::Api
             address1: '1 Test Rd',
             city: 'City',
             country_id: country.id,
-            state_id: 1,
+            state_id: state.id,
             zipcode: '55555',
             phone: '5555555555'
           }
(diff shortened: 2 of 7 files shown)


RyanofWoods:
This change has potential to break codebase's tests if they have not modified the 'country' factory 'states_required' attribute to be the intended behaviour, because the attribute can be `true`. If this fix was released on a major release, this would be very confusing, as developers might think it's a breaking change causing the failed tests, and not the factory.

waiting-for-dev:
Thanks, @RyanofWoods!

What about a warning in the address factory? Something like:

```ruby
after(:build) do |address, evaluator|
  carmen_country = Carmen::Country.code(adress.country.iso)
  Spree::Deprecation.warn <<~MSG if address.state.nil? && Spree::Config.address_require_state && carmen_country.subregions? && address.country.states_required == false
    You built an inconsistent country......
    whether use a country that doesn't require subregions (e.g. ?) or add a state to the address
  MSG
end
```

RyanofWoods:
> Thanks, @RyanofWoods!
> 
> What about a warning in the address factory? Something like:
> 
> ```ruby
> after(:build) do |address, evaluator|
>   carmen_country = Carmen::Country.code(adress.country.iso)
>   Spree::Deprecation.warn <<~MSG if address.state.nil? && Spree::Config.address_require_state && carmen_country.subregions? && address.country.states_required == false
>     You built an inconsistent country......
>     whether use a country that doesn't require subregions (e.g. ?) or add a state to the address
>   MSG
> end
> ```
@waiting-for-dev 
I don't think this warner is aligned with the scenario that I mentioned. The problem is that if they used the Solidus Country factory logic, `states_required` was always `false`, but now it can become `true`, which could breaking tests for them. If they used a build_strategy of `create`, then they will probably get an error of `State cannot be blank`. But it is not guaranteed and they could use other strategies.

I am not really sure there is something we can do, other than perhaps realise it on a minor release and make a note of it in the changelog 🤔 

(I think this warning could be annoying if the have custom logic where some countries's `state_required` is being set to `false` even though it has subregions)

waiting-for-dev:
Hey @RyanofWoods!

> I don't think this warner is aligned with the scenario that I mentioned. The problem is that if they used the Solidus Country factory logic, states_required was always false, but now it can become true, which could breaking tests for them. If they used a build_strategy of create, then they will probably get an error of State cannot be blank.

I'm not familiar with this part of the codebase, but I tried to dig a bit, and it seemed to be that the only place where that could happen is on the address model because of a [custom validator](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/address/state_validator.rb#L64).

> But it is not guaranteed and they could use other strategies.

AFAIK, `after(:build)` will only run for the build and creation strategies. I thought using it on `: build` was a good idea, as they may be asserting that the instance is valid. We could be more localized and use a `before(:create)`.

> (I think this warning could be annoying if the have custom logic where some countries's state_required is being set to false even though it has subregions)

It seemed that there's no frontend for `states_required`, and it's only set during the [population of seeds](https://github.com/solidusio/solidus/blob/a6c2ceeb2ee6bff832080ebb11773f7b6e4c2df0/core/db/default/spree/countries.rb#L16). So users should not be messing with it. If they do not require addresses to have a state when the country is `states_required`, they must have `Spree::Config.address_require_state` to `false`, a condition included in the guard.

WDYT? Let's also wait for the opinion of others.

kennyadsl:
I'd explore if you are making the right assumption in the first place: are we sure that just having subregions is enough to determine if states are required to make an address valid in a given country? That information doesn't seem to live in Carmen (for a good reason, it's just a repository of geographies, doesn't include addresses concerns), but I guess there are places where we have subregions and they are not mandatory for address validity.

RyanofWoods:
That's a good point @kennyadsl 🙂  I don't think it is for the reasons you mentioned. The following issue touches upon this:
https://github.com/solidusio/solidus/issues/3842

I can look into this more 👍

waiting-for-dev:
> are we sure that just having subregions is enough to determine if states are required to make an address valid in a given country?

@kennyadsl, it seems that's the condition we're checking on our seeds:

https://github.com/solidusio/solidus/blob/a6c2ceeb2ee6bff832080ebb11773f7b6e4c2df0/core/db/default/spree/countries.rb#L16

waiting-for-dev (approved):
Thanks, @RyanofWoods. I think now the logic is better covered. However, I'd like another review by someone with more experience than me dealing with this part of Solidus. cc @solidusio/core-team

kennyadsl (approved):
👍 Thanks Ryan, this is a nice improvement!

PR #4275, opened 2022-02-22
Fix n+1 possibilities in Promotion#blacklisted?

mamhoff:
**Description**

This removes some n+1 possibilities in `Spree::Promotion#blacklisted?`. First, we go through a line items' variant to get its product as that is more likely to be preloaded, and second, we use `Enumerable` to find out whether any line item's products are non-promotionable rather than issuing a DB query for every promotion / every line item when using an order-level promotion rule.

**Checklist:**
- n/a I have updated Guides and README accordingly to this change (if needed)
- n/a I have added tests to cover this change (if needed)
- n/a I have attached screenshots to this PR for visual changes (if needed)


jarednorman (approved):

tvdeyen (commented):
Nice work. There are some specs failing that seem to be related. Mind looking at those?

mamhoff:
Done :sweat_drops:

tvdeyen (approved):

PR #4280, opened 2022-02-24
Remove n+1 in Spree::Tax::TaxLocation class

mamhoff:
This class was initialized with a country. But it forgot the country
after initialization, making a roundtrip to the database necessary. The
change in this commit keeps the country and state in memory instead.

**Checklist:**
- [x I have added a detailed description into each commit message

### DIFF
--- a/core/app/models/spree/tax/tax_location.rb
+++ b/core/app/models/spree/tax/tax_location.rb
@@ -8,7 +8,7 @@ module Spree
     # @attr_reader [Integer] country_id the ID of a Spree::Country object
     # @attr_reader [Integer] state_id the ID of a Spree::State object
     class TaxLocation
-      attr_reader :country_id, :state_id
+      attr_reader :country, :state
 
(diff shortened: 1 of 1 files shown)


tvdeyen (approved):
Nice work. Thanks

core/app/models/spree/tax/tax_location.rb
```diff
@@ -19,18 +19,15 @@ class TaxLocation
       #
       # @return [Spree::Tax::TaxLocation] a Spree::Tax::TaxLocation object
       def initialize(country: nil, state: nil)
-        @country_id = country && country.id
-        @state_id = state && state.id
+        @country, @state = country, state
       end
+      delegate :id, to: :state, prefix: true, allow_nil: true
```
waiting-for-dev, line 24: Non-blocking. WDYT about putting both `delegate` at the top of the class?

waiting-for-dev (approved):
Nice, thanks!!

PR #4282, opened 2022-02-24
Product Promotion Rule: Use in-memory objects

mamhoff:
The chance for any request to preload `order.products` is very slim, and
the chance for any request to preload `order.line_items` is much better.

**Checklist:**

### DIFF
--- a/core/app/models/spree/promotion/rules/product.rb
+++ b/core/app/models/spree/promotion/rules/product.rb
@@ -31,17 +31,19 @@ module Spree
           return true if eligible_products.empty?
 
           case preferred_match_policy
-          when 'all'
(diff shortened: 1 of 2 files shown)


core/app/models/spree/promotion/rules/product.rb
```diff
@@ -68,6 +70,16 @@ def product_ids_string
         def product_ids_string=(product_ids)
           self.product_ids = product_ids.to_s.split(',').map(&:strip)
         end
+
+        def product_ids
```
tvdeyen, line 74: This should already be handled by active record, no?

tvdeyen (commented):
There is probably an unnecessary change. but this is 😗👌🏻

tvdeyen (approved):

PR #4285, opened 2022-02-28
Promotion Rule CreateItemAdjustments: Use in-memory objects

mamhoff:
This changes the function that obtains the line items of an order to
create adjustments for to use in-memory objects rather than querying the
database for line items that have no adjustments with a source from the
current rule.

Promotion rules can easily be preloaded, and there's no need for the DB
call in the previous code.

**Checklist:**


jarednorman (approved):

tvdeyen (approved):
Thanks

PR #4286, opened 2022-02-28
OrderContents: Initialize line item with empty adjustments

mamhoff:
If we do not explicitly add an empty array here, a subsequent call to
`line_item.adjustments` will trigger an unnecessary DB hit.

**Checklist:**


mamhoff:
I think the failing spec is spurious and not related. Can an admin please trigger a rebuild?

tvdeyen:
@mamhoff Thanks! triggered a re-run

jarednorman (approved):
Loving all these optimizations.

tvdeyen (approved):

Issue #2807, opened 2018-07-20
Moving from Integer to Bigint or UUID for Primary Key


BenMorganIO:
Hey guys, with Rails using bigint as the default now, perhaps it would be good to talk about moving Solidus from integer PK to bigint or UUID PK. I personally would prefer UUID, but Rails has standardized to bigint.

Thoughts on this?

skukx:
I think this is a good idea. Using `BigInt` would probably be the best since it is what rails defaults to.  We recently hit an issues where one of our databases hit the `int` limit and it crashed a whole lot of services.

ericsaupe:
I think keeping with Rails standard is probably the best. I'm usually for keeping things "standard" in the community following Rails' lead unless there is a strong reason giving great benefits using UUID. I'd love to hear them.

BenMorganIO:
If we add a migration that converts _all_ of the primary keys to bigint, we could also probably add an option to make it uuid since it will all be abstracted out enough. If not, bigint is an improvement.

BenMorganIO:
@ericsaupe there's really three main reasons why people use UUIDs over bigints:

1. Faster writes.
2. Security.
3. Checking IDs in tests is easier.

If Rails does the UUID generation, you don't need to do a round trip and ask for what the ID is on return. This saves a round trip after insertion. So in SQL you just do `INSERT x` instead of `INSERT x RETURNING id`.

With the security benefit, you can hide data that reveals information about your system. This is an amazing bonus.

The main drawback with UUIDs is the storage requirement, which is only about ~150MB for 10,000,000 rows. I could be wrong though. Since that's not much storage space for the benefits of performance and security, plus storage is always getting cheaper and the speed of light ain't getting faster through a vacuum let alone mediums, I, therefore, use UUIDs.

cedum:
UUIDs as a primary are required mostly in distributed architectures/datastores, because generating correctly serial numbers in such architectures is much more challenging. IMHO this is the main reason when choosing UUIDs becomes sound from a practical perspective.
Solidus, being basically a monolith, has no such limitations.

When evaluating "faster writes", keep in mind that when having primary UUIDs in almost all cases it'll require to have also a "sort index" on a column like `created_at`. This means every `INSERT` will require (at least) two writes on disk.

I'd stick with autoincrementing integers as a primary. There's quite a big cost of migrating to UUIDs and honestly I don't see strong benefits to justify the conversion of all the primary indexes on a production database to a completely different datatype.

[...]

PR #4287, opened 2022-03-01
Authorize uuid for existing object in sortable table

julienanne:
**Description**

In the backend section, from Ref #3591 we can't sort object list when we use uuid for the object's id.
We need to allow `objId` to be an uuid.
This change need to be backported until the version 2.11

Hope it can help.

PR v2.11 : #4299 
PR v3.0 : #4300 
PR v3.1 : #4301

**Checklist:**
~~- [ ] I have updated Guides and README accordingly to this change (if needed)~~
~~- [ ] I have added tests to cover this change (if needed)~~ On this point i have encoutered many issues with teaspoon on docker and i don't see existing tests about that part of the backend.

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/components/sortable_table.js
+++ b/backend/app/assets/javascripts/spree/backend/components/sortable_table.js
@@ -1,6 +1,14 @@
 //= require solidus_admin/Sortable
 /* eslint no-unused-vars: "off" */
 
+/* Check if string is valid UUID */
+function isAValidUUID(str) {
+  // https://stackoverflow.com/a/13653180/8170555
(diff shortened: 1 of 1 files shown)


julienanne:
The test marked as failed is a flaky test but i can't rerun the workflow from start ;)

julienanne:
> Thanks, if you need this in 2.11, can you please backport this into 2.11, 3.0 and 3.1 branches as well with different PRs?
> 
> Also, non-blocking but do you think there's a way to add some specs around this behavior? Just wanted to be sure we won't be removing this feature in the future without noticing.

Ok I open the three PRs one for each version with the same content.
For the testing part I don't know, if I have more time I will check if I can do it in another PRs.
I noticed that you also have a big discussion about uuid on this issue #2807 (https://github.com/solidusio/solidus/issues/2807), maybe this PR can be part of that discussion, maybe it is not the only breaking changes on UI with uuid ?
Thanks for approval.

spaghetticode (approved):
@julienanne thank you!

PR #4290, opened 2022-03-03
Fix(OrderMerger): Do not carry line item adjustments to new order

mamhoff:
**Description**

Adjustments are connected to the spree_orders table via two ways:
- via their adjustable's (line item's) order ID
- via their own order ID

If we move a line item with adjustments to new order, the adjustment's
order ID will not be changed to the new order.

Instead of changing the foreign key of an existing line item here, let's
build a new line item with the old line item's attributes.

**Checklist:**

### DIFF
--- a/core/spec/models/spree/order_merger_spec.rb
+++ b/core/spec/models/spree/order_merger_spec.rb
@@ -141,6 +141,20 @@ module Spree
         expect(order_1.line_items.pluck(:quantity)).to match_array([1, 1])
         expect(order_1.line_items.pluck(:variant_id)).to match_array([variant.id, variant_2.id])
       end
+
+      context "with line item promotion applied to order 2" do
+        let!(:promotion) { create(:promotion, :with_line_item_adjustment, apply_automatically: true) }
+
+        before do
+          Spree::PromotionHandler::Cart.new(order_2).activate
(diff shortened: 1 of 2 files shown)


core/spec/models/spree/order_merger_spec.rb
```diff
@@ -141,6 +141,21 @@ module Spree
...
+        let!(:promotion) { create(:promotion, :with_line_item_adjustment, apply_automatically: true) }
+
+        before do
+          Spree::PromotionHandler::Cart.new(order_2).activate
+          expect(order_2.line_items.flat_map(&:adjustments)).not_to be_empty
+        end
+
+        it "does not carry a line item adjustments with the wrong order ID over" do
+          subject.merge!(order_2)
+          expect { order_2.reload }.to raise_exception(ActiveRecord::RecordNotFound)
```
tvdeyen, line 155: Is this necessary for this case? It is kinda confusing

tvdeyen (commented):
This is a nice find.

core/spec/models/spree/order_merger_spec.rb
mamhoff: Removed

tvdeyen (approved):

PR #4292, opened 2022-03-04
Allow OrderWalkThrough to take a user

mamhoff:
The OrderWalkThrough creates an order that can't be completed because
the credit card associated with the order has a different user (some
user) than the order it's being associated to (nil). This leads to
failures on trying to add that credit card to the user's wallet.

**Checklist:**


tvdeyen (approved):
Makes sense. Thank you.

waiting-for-dev (approved):

jarednorman (approved):

PR #4293, opened 2022-03-04
Stock location optimizations

mamhoff:
This allows the stock location to operate mostly on preloaded stock items, rather than issuing `where` to the DB a lot. 

**Checklist:**

### DIFF
--- a/core/app/models/spree/stock_location.rb
+++ b/core/app/models/spree/stock_location.rb
@@ -51,11 +51,14 @@ module Spree
 
     # Returns an instance of StockItem for the variant id.
     #
-    # @param variant_id [String] The id of a variant.
+    # @param variant_or_id [String] The id of a variant.
     #
     # @return [StockItem] Corresponding StockItem for the StockLocation's variant.
-    def stock_item(variant_id)
-      stock_items.where(variant_id: variant_id).order(:id).first
+    def stock_item(variant_or_id)
+      stock_items.detect do |stock_item|
+        stock_item.variant_id == variant_or_id ||
+          stock_item.variant == variant_or_id
+      end
     end
 
     # Attempts to look up StockItem for the variant, and creates one if not found.
--- a/core/app/models/spree/stock_location.rb
+++ b/core/app/models/spree/stock_location.rb
@@ -38,15 +38,13 @@ module Spree
     end
 
     # Wrapper for creating a new stock item respecting the backorderable config
-    def propagate_variant(variant)
-      stock_items.create!(variant: variant, backorderable: backorderable_default)
-    end
-
-    # Return either an existing stock item or create a new one. Useful in
-    # scenarios where the user might not know whether there is already a stock
-    # item for a given variant
-    def set_up_stock_item(variant)
-      stock_item(variant) || propagate_variant(variant)
+    def propagate_variant(variant_or_id)
+      vid = if variant_or_id.is_a?(Variant)
+        variant_or_id.id
+      else
+        variant_or_id
+      end
+      stock_items.create!(variant_id: vid, backorderable: backorderable_default)
     end
 
     # Returns an instance of StockItem for the variant id.
@@ -67,14 +65,15 @@ module Spree
     # but controller actions can pass just the variant id as a parameter.
     #
     # @return [StockItem] Corresponding StockItem for the StockLocation's variant.
-    def stock_item_or_create(variant)
-      vid = if variant.is_a?(Variant)
-        variant.id
+    def stock_item_or_create(variant_or_id)
+      vid = if variant_or_id.is_a?(Variant)
+        variant_or_id.id
       else
-        variant
+        variant_or_id
       end
-      stock_item(vid) || stock_items.create(variant_id: vid)
+      stock_item(vid) || propagate_variant(vid)
     end
+    alias_method :set_up_stock_item, :stock_item_or_create
 
     def count_on_hand(variant)
       stock_item(variant).try(:count_on_hand)
--- a/backend/spec/features/admin/orders/order_details_spec.rb
+++ b/backend/spec/features/admin/orders/order_details_spec.rb
@@ -563,8 +563,8 @@ describe "Order Details", type: :feature, js: true do
       end
 
       describe 'line item sort order' do
-        let(:product2) { create(:product) }
-        let(:product3) { create(:product) }
+        let!(:product2) { create(:product) }
+        let!(:product3) { create(:product) }
 
         before do
           # grab this one first and then create others that should end up before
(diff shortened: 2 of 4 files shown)


core/app/models/spree/stock_location.rb
```diff
@@ -51,11 +51,14 @@ def set_up_stock_item(variant)
 
     # Returns an instance of StockItem for the variant id.
     #
-    # @param variant_id [String] The id of a variant.
+    # @param variant_or_id [String] The id of a variant.
     #
     # @return [StockItem] Corresponding StockItem for the StockLocation's variant.
-    def stock_item(variant_id)
-      stock_items.where(variant_id: variant_id).order(:id).first
```
jarednorman, line 58: I can't think of why it matters, but could losing the order on ID matter?

core/app/models/spree/stock_location.rb
```diff
@@ -38,15 +38,13 @@ def state_text
...
-    # scenarios where the user might not know whether there is already a stock
-    # item for a given variant
-    def set_up_stock_item(variant)
-      stock_item(variant) || propagate_variant(variant)
+    def propagate_variant(variant_or_id)
+      vid = if variant_or_id.is_a?(Variant)
+        variant_or_id.id
+      else
+        variant_or_id
+      end
```
jarednorman, line 46: I know that this would be a breaking change, but I wish that there was an elegant way to transition to using keyword arguments instead of multi-purpose arguments. I'd love this to be like:

```ruby
def propagate_variant(variant: nil, variant_id: nil)
  vid = variant_id || variant&.id or raise ArgumentError

  stock_items.create!(variant_id: vid, backorderable: backorderable_default)
end
```

One option could be to deprecate `propagate_variant` for a method called `propagate` (since I think it would still be clear due to the use of keyword args) and maintain `propagate_variant`'s API until it gets removed. Similarly we could create a `stock_item_for(variant: nil, variant_id: nil)` method.

I don't really feel strongly about this; I'm just not a fan of having these arguments that can be different things and inspecting them.

forkata: I think this pattern of allowing both `id` or record is a bit unfortunate and results in the type checking we now have to have in multiple places in this class. We would have a much nicer API if we require a full `Variant` record for all public methods in this class that require it. If we are willing to deprecate anything it should be the ability to pass in `id` in `stock_item_or_create` so we can remove the type checking there, rather than propagating that throughout other existing methods. In my experience passing id's around usually results in more unnecessary queries because often the calling code already has access to a fully loaded record.

core/app/models/spree/stock_location.rb
```diff
@@ -51,11 +51,14 @@ def set_up_stock_item(variant)
...
     # Returns an instance of StockItem for the variant id.
     #
-    # @param variant_id [String] The id of a variant.
+    # @param variant_or_id [String] The id of a variant.
     #
     # @return [StockItem] Corresponding StockItem for the StockLocation's variant.
-    def stock_item(variant_id)
-      stock_items.where(variant_id: variant_id).order(:id).first
+    def stock_item(variant_or_id)
+      stock_items.detect do |stock_item|
```
forkata, line 56: 👋🏼 @mamhoff I am curious, have we done any real world benchmarking for how this is going to impact stores with a large number of variants/stock items. I work on a store which has ~90K stock items in each stock location and an isolated call to this method is _much_ slower in Ruby due to the need to load all records in memory before iterating over them. The benchmark below is just from my local development environment so doesn't account for network traffic/database performance etc. This is also a "worst case" scenario where the stock item is the last one in the list (based on the order the db returns them).
```ruby
[1] pry(main)> Benchmark.measure { Spree::StockLocation.first.stock_item(93917) }
=> #<Benchmark::Tms:0x00007fc663c98098 @cstime=0.0, @cutime=0.0, @label="", @real=0.2266759998165071, @stime=0.026478, @total=0.07949200000000012, @utime=0.05301400000000012>
[2] pry(main)> Benchmark.measure { Spree::StockLocation.first.stock_items.detect { |i| i.variant_id == 93917 } }
=> #<Benchmark::Tms:0x00007fc64944ccf0 @cstime=0.0, @cutime=0.0, @label="", @real=2.5800430001690984, @stime=0.06728000000000023, @total=2.5387829999999996, @utime=2.4715029999999993>
```
I realize this example isn't indicative of real world usage, that's why I am curious if you've seen a performance increase with the ruby code when adding variants etc? An alternative here could be to use lazy enumeration to save us from having to load all records in memory 🤔

mamhoff: Good point; I came to this by going through all the callbacks on `Spree::LineItem` and found `Spree::OrderInventory` to make heavy use of this method. We have some 4000 stock items per stock location, so it's not that bad here, but we also have really large orders with some 200 line items; and we provide bulk adding / removing items to those orders, so anything that creates n+1 queries on line items really hurts us. 

My instinct so far has been that the combination of preloading records and iterating is often a lot faster than even 10 DB queries, especially when looking at things in production, and it's gotten me really really far with getting some of our more expensive endpoints faster. 

I guess this is a question of Solidus policy: We do want to serve high-traffic sites, but do we also want to serve super-large-catalog sites? What are the trade-offs? :)

core/app/models/spree/stock_location.rb
mamhoff: Oh it's awful. I just kept it because it seemed like the least amount of change - absolutely happy to add `propagate` and deprecate the rest. I'd even go for a signature like `stock_item_for(variant: nil) signature; less stuff to worry about, and less mucking about with IDs that stop meaning anything up or down the callstack.

core/app/models/spree/stock_location.rb
mamhoff: Hm, there can only be one undeleted stock item per stock location and variant, so I'd say no:
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/stock_item.rb#L12

core/app/models/spree/stock_location.rb
forkata: > We do want to serve high-traffic sites, but do we also want to serve super-large-catalog sites? 

I think whenever possible we should try to do both 😸 I am 💯 with you on this work to reduce DB queries by operating on in-memory preloaded records, it will improve performance significantly!

I am curious if in this case we can invert the lookup by loading all stock items for a variant and looking for one with a matching stock location where this is called from - 
https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_inventory.rb#L65 🤔 It is less likely that a store has tens of thousands of stock locations, whereas that is a totally reasonable number of stock items imho.

mamhoff: https://github.com/solidusio/solidus/pull/4303

mamhoff:
I think I'm going to close this. @forkata is right: Given that we want to support stores with tens of thousands of stock items per stock location, using DB lookups here is smart. One shouldn't really look up stock items through their stock location when the variant is known unless administering them.

PR #4303, opened 2022-03-17
OrderInventory: Use variant stock items

mamhoff:
This class, which is executed as a callback when saving line items,
should get any inventory information not through all stock location's
stock items, but much rather through the stock items connected to its
variant. Less data to care about.

**Checklist:**


tvdeyen (approved):
Thanks. This helps a lot.

forkata (approved):
Thanks @mamhoff, this looks great!

jarednorman (approved):
Yeah, this seems good.

PR #4307, opened 2022-03-19
Refactor promotion usage counts

mamhoff:
Refactors Spree::Promotion#usage_count, Spree::PromotionCode#usage_count and Spree::Promotion#used_by? to use a common method on Spree::Promotion. 

This is extracted work from #4296 and makes that work easier. 

All of these methods are well-tested, so I'm not adding more specs here. 

**Checklist:**

### DIFF
--- a/core/app/models/spree/adjustment.rb
+++ b/core/app/models/spree/adjustment.rb
@@ -62,6 +62,7 @@ module Spree
                .distinct
       exclude_canceled ? result.merge(Spree::Order.not_canceled) : result
     end
+    deprecate :in_completed_orders, "Please don't use this and rather go through Spree::Promotion#discounted_orders"
 
     def finalize!
       update!(finalized: true)
--- a/core/app/models/spree/promotion.rb
+++ b/core/app/models/spree/promotion.rb
@@ -67,6 +67,19 @@ module Spree
       ).first
     end
 
+    # All orders that have been discounted using this promotion
+    def discounted_orders
+      Spree::Order.
+        joins(:all_adjustments).
+        where(
+          spree_adjustments: {
+            source_type: "Spree::PromotionAction",
+            source_id: actions.map(&:id),
+            eligible: true
+          }
+        ).distinct
+    end
+
     def as_json(options = {})
       options[:except] ||= :code
       super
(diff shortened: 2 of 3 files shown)


core/app/models/spree/promotion.rb
```diff
@@ -67,7 +67,7 @@ def self.with_coupon_code(val)
       ).first
     end
 
-    # All orders that have been discounted using this promotion
+    # All orders that have been discounted using this promotionp
```
jarednorman, line 70: Typo

jarednorman:
Besides the typo, this all seems sensible to me.

core/app/models/spree/promotion.rb
mamhoff: Thanks, fixes

tvdeyen (approved):
Thank you.

core/app/models/spree/adjustment.rb
```diff
@@ -62,6 +62,7 @@ def self.in_completed_orders(excluded_orders: [], exclude_canceled: false)
                .distinct
       exclude_canceled ? result.merge(Spree::Order.not_canceled) : result
     end
+    deprecate :in_completed_orders, "Please don't use this and rather go through Spree::Promotion#discounted_orders"
```
kennyadsl, line 65: Why deprecating this? It is no more used?

kennyadsl (approved):
Left a non-blocking question, thanks for this!

jarednorman (approved):

core/app/models/spree/promotion.rb
```diff
@@ -67,6 +67,19 @@ def self.with_coupon_code(val)
       ).first
     end
 
+    # All orders that have been discounted using this promotion
+    def discounted_orders
```
spaghetticode, line 71: This looks like a relationship, I'm wondering if it would be preferable to define it that way... it looks possible to me, but I think we'd need to add a few relationships on other models in that case. What do you think?

mamhoff: I'm a bit wary of defining a large amount of relationships. They offer a bunch of new APIs that I would not want to exist: 
`order.applied_promotions.create!(...)`? Doesn't look like a step forward to me. This simple method does the job :)

spaghetticode (approved):
Thanks Martin 👍

PR #4309, opened 2022-03-21
Stay below v7 with kt-paperclip

### LINKED ISSUE #4308: kt-paperclip version mismatch
`kt-paperclip` is open ended in v2.11.5 but constrained to v6 in latest versions.
This allows paperclip v7 in 2.11.15 and only v6 in 3.0 and above. Not sure if this intentional
or a mistake. It surely is confusing for folks updating from v2 to v3

**Solidus Version**:

`>= v2.11.15`

**To Reproduce**

See the core gemspec in `master` https://github.com/solidusio/solidus/blob/master/core/solidus_core.gemspec#L44
The gemspec in v2.11.15 https://github.com/solidusio/solidus/blob/v2.11/core/solidus_core.gemspec#L42

Not sure what is the expected version. Happy to provide a PR if we agree on a version we want to support.

tvdeyen:
**Description**

We backported the usage of kt-paperclip to v2.11.15,
but did not added a upper boundary. Since v3.0.5 only
allows v6.3 and above, we do the same for v2.11 as well.

Closes #4308 

**Checklist:**

### DIFF
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -39,7 +39,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'kaminari-activerecord', '~> 1.1'
   s.add_dependency 'mini_magick', '~> 4.10'
   s.add_dependency 'monetize', '~> 1.8'
-  s.add_dependency 'kt-paperclip', '>= 4.4.0'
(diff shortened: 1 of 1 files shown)


jarednorman (approved):

spaghetticode (approved):
@tvdeyen thanks!

PR #4313, opened 2022-03-22
Fix flaky spec for Spree::ShippingMethod#available_to_store

mamhoff:
This spec had subject and expectations somehow mixed up (at least in my
mind), and it did fail occasionally, probably because of that.

See https://app.circleci.com/pipelines/github/solidusio/solidus/3095/workflows/16ada36d-0342-4b15-9e38-cde85df111a2/jobs/28315/tests#failed-test-0

**Checklist:**

### DIFF
--- a/core/spec/models/spree/shipping_method_spec.rb
+++ b/core/spec/models/spree/shipping_method_spec.rb
@@ -219,10 +219,10 @@ RSpec.describe Spree::ShippingMethod, type: :model do
 
   describe '.available_to_store' do
     let(:store) { create(:store) }
-    let(:first_shipping_method) { create(:shipping_method, stores: [store]) }
-    let(:second_shipping_method) { create(:shipping_method, stores: [store]) }
+    let!(:first_shipping_method) { create(:shipping_method, stores: [store]) }
+    let!(:second_shipping_method) { create(:shipping_method, stores: [store]) }
 
-    subject { [first_shipping_method, second_shipping_method] }
+    subject { described_class.available_to_store(store) }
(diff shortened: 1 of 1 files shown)


core/spec/models/spree/shipping_method_spec.rb
```diff
@@ -231,20 +231,21 @@ def matching(categories)
...
     end
 
     context 'when the store has shipping methods associated' do
       before { create(:shipping_method) }
 
       it 'returns the associated records' do
-        expect(store.shipping_methods).to eq(subject)
-        expect(described_class.available_to_store(store)).to match_array(subject)
+        expect(store.shipping_methods).to eq([first_shipping_method, second_shipping_method])
+        expect(subject).to contain_exactly(first_shipping_method, second_shipping_method)
```
jarednorman, line 248: If there's no order on these associations (I assume there isn't), then the spurious failures are likely due to the database returning the records in a different order.

This last assertion uses `contain_exactly`. Like `match_array`, it doesn't care about order. Maybe the other `.eq` assertions should be traded for one of those matchers?

jarednorman (changes requested):

jarednorman (approved):

jarednorman:
Thanks Martin!

core/spec/models/spree/shipping_method_spec.rb
mamhoff: Thank you, good observation!

tvdeyen (approved):

Issue #3357, opened 2019-09-30
Legacy backend shows both carton and shipment as shipped package


cedum:
The admin order->shipments UI shows "Shipped package ..." for both Cartons and Shippings. While in reality, the "real" shipped packages are represented only by Cartons (afaik).

**Solidus Version**:
I think it's present in all maintained versions.

**To Reproduce**
- create an order with at least 2 products
- cancel one product and ship the other one (having a canceled item highlights better this issue)
- it now shows two shipped packages (first is the shipped `Carton`, the second one is the `Shipment`)

**Screenshots**

![](https://cedum.s3-eu-west-2.amazonaws.com/NQq2D7Lpu/Screen-Shot-2019-09-30-12-00-14.40.jpg)

tvdeyen:
I have a nearly done PR laying around somewhere. Made this for a customer once and we are happy to share this with the community

dportalesr:
I could take care of this. I would just need to know what's the actual expected outcome in these cases. I mean, the shipment in the image do have the shipped status, so not really sure what's the required change.

seandawes:
@cedum Do not believe this should change as this relies on the data found in spree_shipments table and more so specifically spree_shipments.state ? As spree_shipments.tracking contains the shipment containers tracking number if available and can be mapped to your product catalog via spree_shipments.order_id where as spree_cartons does not contain anything to connect the tables? 

First thing I thought of and maybe Im wrong here.

jarednorman:
Cartons should also have tracking and can be mapped back to the order through the inventory units.

tvdeyen:
I think the most confusing thing about cartons is that they can hold multiple orders. This is not what most of our users do and we should think about moving that feature into an extension instead. In that extension Cartons should then probably be its own Admin Tab, because they do not belong to an individual order any more.

For the current backend I think we should remove that feature and nest all cartons under each shipment and remove most of the duplication. That's what I did in the draft PR.

PR #4315, opened 2022-03-25
Move carton in another tab in admin UI

MassimilianoLattanzio:
**Description**
The existing UX for a shipped order is somewhat confusing. The "Shipments" tab of the completed order renders both shipments and cartons, which ends up making it seem like there are duplicated shipments. Instead, maybe it's better to split this out. 

This PR adds a separate tab to the order bar for cartons, and render the cartons separately. It also removes the carton rendering from the default shipment view.

Ref: https://github.com/solidusio/solidus/issues/3357

**Result**
Screenshot 2022-03-25 at 12 38 55
Screenshot 2022-03-25 at 12 38 47

**Checklist:**

### DIFF
--- a/backend/app/controllers/spree/admin/orders_controller.rb
+++ b/backend/app/controllers/spree/admin/orders_controller.rb
@@ -4,7 +4,7 @@ module Spree
   module Admin
     class OrdersController < Spree::Admin::BaseController
       before_action :initialize_order_events
-      before_action :load_order, only: [:edit, :update, :complete, :advance, :cancel, :resume, :approve, :resend, :unfinalize_adjustments, :finalize_adjustments, :cart, :confirm]
+      before_action :load_order, only: [:edit, :update, :complete, :advance, :cancel, :resume, :approve, :resend, :unfinalize_adjustments, :finalize_adjustments, :cart, :confirm, :cartons]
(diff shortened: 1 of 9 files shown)


tvdeyen (dismissed):
I don't think this is good UX. Users will be even more confused by that. Unfortunately I didn't had time to finish my PR yet, but I think we should nest cartons under each shipment, so it is clear where they belong to.

tvdeyen:
I think the most confusing thing about cartons is that they can hold multiple orders. This is not what most of our users do and we should think about moving that feature into an extension instead. In that extension Cartons should then probably be its own Admin Tab, because they do not belong to an individual order any more.

For the current backend I think we should remove that feature and nest all cartons under each shipment and remove most of the duplication. I am preparing a PR right now, that can happily be took over by someone with more time.

MassimilianoLattanzio:
Hey @tvdeyen, thanks for your messages.

I created this PR because a client asked me for this update (move carton to another tab) and I saw an open issue for, more or less, the same problem.

As you said, probably this is not the best way to proceed so maybe it's better to discuss it in the [other draft PR](https://github.com/solidusio/solidus/pull/4319) you created and close this one? WDYT?

waiting-for-dev:
As suggested, closing in favor of #4319. Feel free to get back if you think it's the wrong decision.

PR #4316, opened 2022-03-28
Fixes using ActiveStorage adapter with libvips as variant processor

waiting-for-dev:
**Description**

[Rails 6.0 introduced](https://github.com/rails/rails/blob/6-0-stable/activestorage/CHANGELOG.md#rails-600beta1-january-18-2019) the option of using libvips as an alternative to imagemagick as a backend image processor.

Both processors are handled using the [image_processing](https://github.com/janko/image_processing) library as adapter. However, the API for both is not fully compatible. What it's important for us when generating a variant, the `strip` option needs to be within the `saver` group to work with both adapters. See https://gist.github.com/brenogazzola/a4369965a1da426d50f11d080fe2e563#1-move-everything-that-has-to-do-with-compression-to-a-saver-hash for details. We also need to explicitly coerce size values to integers for libvips.

We're also adding libvips to the development docker image and allowing to run tests with that adapter by setting the
`ACTIVE_STORAGE_VARIANT_PROCESSOR` env variable. However, we still default it to `:mime_magick` (Ruby's wrapper for imagemagick), Rail's default previous to 7.0. We'll update it once we integrate support for Rails 7.0.

I also manually tested that metadata is still stripped when using both adapters.

Needs to be backported.

**Checklist:**

### DIFF
--- a/.dockerdev/Dockerfile
+++ b/.dockerdev/Dockerfile
@@ -13,6 +13,7 @@ RUN apt-get update -qq \
     curl \
     git \
     imagemagick \
+    libvips \
     libmariadb-dev \
     sqlite3 \
     libsqlite3-dev \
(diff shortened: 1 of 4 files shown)


waiting-for-dev:
Closing in favor of #4317

PR #4318, opened 2022-03-28
Fixes defining thumbnail sizes through ActiveStorage adapter

waiting-for-dev:
**Description**

The adapter intends to make it possible to work with ActiveStorage as it was Paperclip. We had been using Paperclip until Rails 6.1 introduced ActiveStorage's public links. We moved to ActiveStorage as the default file attachment backend from that point on, but we tried to make its API compatible with Paperclip's one. See https://github.com/solidusio/solidus/pull/3501 for details.

When trying to translate [Paperclip's style definitions](https://github.com/solidusio/solidus/blob/564b4fe54648537b6b1420eed5a05cba6854e915/core/lib/spree/app_configuration.rb#L494) (based on [ImageMagick](https://legacy.imagemagick.org/Usage/thumbnails/)) to what ActiveStorage expects (which is [image_processing syntax](https://github.com/janko/image_processing), common for ImageMagick and libvips), we were transforming everything to a [`resize_to_limit`](https://github.com/janko/image_processing/blob/master/doc/minimagick.md#resize_to_limit)
option. However, that corresponds to the `>` symbol in ImageMagick, but not to other options that can be given.

Whit this commit, we add support for defining other kinds of transformations in ActiveStorage. However, we're still missing more
advanced options that can be given when using [standalone Paperclip](https://www.rubydoc.info/gems/kt-paperclip/7.1.1) (example `{ geometry: '100x100', format: :jpg }`)  or [standalone ActiveStorage](https://github.com/janko/image_processing/blob/master/doc/minimagick.md#methods) (example `rotate: 90`). In the long run, we should aim to completely
differentiate how definitions are given for each library.

Needs to be backported.

**Checklist:**

### DIFF
--- a/core/app/models/concerns/spree/active_storage_adapter/attachment.rb
+++ b/core/app/models/concerns/spree/active_storage_adapter/attachment.rb
@@ -4,14 +4,16 @@ require 'mini_magick'
 
 module Spree
   module ActiveStorageAdapter
-    # Decorares AtiveStorage attachment to add methods exptected by Solidus'
+    # Decorates ActiveStorage attachment to add methods expected by Solidus'
     # Paperclip-oriented attachment support.
     class Attachment
       delegate_missing_to :@attachment
 
+      attr_reader :attachment
+
       def initialize(attachment, styles: {})
         @attachment = attachment
-        @styles = normalize_styles(styles)
+        @transformations = styles_to_transformations(styles)
       end
 
       def exists?
@@ -27,13 +29,13 @@ module Spree
       end
 
       def variant(style = nil)
-        size = style_to_size(style)
-        @attachment.variant(
-          resize_to_limit: size,
+        transformation = @transformations[style] || default_transformation(width, height)
+
+        @attachment.variant({
           saver: {
             strip: true
           }
-        ).processed
+        }.merge(transformation)).processed
       end
 
       def height
@@ -59,12 +61,23 @@ module Spree
         @attachment.metadata
       end
 
-      def normalize_styles(styles)
-        styles.transform_values { |v| v.split('x').map(&:to_i) }
+      def styles_to_transformations(styles)
+        styles.transform_values(&method(:imagemagick_to_image_processing_definition))
       end
 
-      def style_to_size(style)
-        @styles.fetch(style&.to_sym) { [width, height] }
+      def imagemagick_to_image_processing_definition(definition)
+        width_height = definition.split('x').map(&:to_i)
+
+        case definition[-1].to_sym
+        when :^
+          { resize_to_fill: width_height }
+        else
+          default_transformation(*width_height)
+        end
+      end
+
+      def default_transformation(width, height)
+        { resize_to_limit: [width, height] }
       end
     end
   end
--- a/core/spec/models/spree/concerns/active_storage_adapter/attachment_spec.rb
+++ b/core/spec/models/spree/concerns/active_storage_adapter/attachment_spec.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+unless ENV['DISABLE_ACTIVE_STORAGE']
+  RSpec.describe Spree::ActiveStorageAdapter::Attachment do
+    describe '#variant' do
+      it "converts to resize_to_limit when definition doesn't contain any special symbol" do
+        image = create(:image)
+
+        attachment = described_class.new(image.attachment.attachment, styles: { mini: '10x10' })
+
+        expect(
+          attachment.variant(:mini).variation.transformations
+        ).to include(resize_to_limit: [10, 10])
+      end
+
+      it 'converts to resize_to_limit when definition ends with >' do
+        image = create(:image)
+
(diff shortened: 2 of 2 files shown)


core/app/models/concerns/spree/active_storage_adapter/attachment.rb
```diff
@@ -56,15 +58,28 @@ def destroy
...
+        @decorated_attachment.metadata
+      end
+
+      def styles_to_transformations(styles)
+        styles.transform_values(&method(:paperclip_to_image_processing_definition))
       end
 
-      def normalize_styles(styles)
-        styles.transform_values { |v| v.split('x').map(&:to_i) }
+      def paperclip_to_image_processing_definition(definition)
```
tvdeyen, line 68: Nitpick: The string is a imagemagick transform string and not a paperclip one. Paperclip just happens to use imagemagick for transfomations.

tvdeyen (approved):
makes sense

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
```diff
@@ -4,14 +4,16 @@
 
 module Spree
   module ActiveStorageAdapter
-    # Decorares AtiveStorage attachment to add methods exptected by Solidus'
+    # Decorates AtiveStorage attachment to add methods expected by Solidus'
```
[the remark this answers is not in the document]
waiting-for-dev: Thanks, fixed.

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
```diff
@@ -4,14 +4,16 @@
...
-      delegate_missing_to :@attachment
+      delegate_missing_to :@decorated_attachment
 
-      def initialize(attachment, styles: {})
-        @attachment = attachment
-        @styles = normalize_styles(styles)
+      attr_reader :decorated_attachment
+
+      def initialize(decorated_attachment, styles: {})
+        @decorated_attachment = decorated_attachment
```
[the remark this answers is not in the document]
waiting-for-dev: Fair enough. I'm keeping the old name now.

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
waiting-for-dev: It makes sense. Updated. Thanks.

spaghetticode (approved):
Thank you Marc 🎉

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
```diff
@@ -59,12 +61,25 @@ def metadata
...
+        styles.transform_values(&method(:imagemagick_to_image_processing_definition))
+      end
+
+      def imagemagick_to_image_processing_definition(definition)
+        width_height = definition.split('x').map(&:to_i)
+
+        case definition[-1].to_sym
+        when :>
+          { resize_to_limit: width_height }
+        when :^
```
tvdeyen, line 74: This actually has to be

```suggestion
        when :#
```

for `resize_to_fill`, because it crops the image to fill the space. See https://github.com/janko/image_processing/wiki/Resize-To-Fill

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
```diff
@@ -59,12 +61,25 @@ def metadata
...
+          { resize_to_fill: width_height }
+        else
+          { resize_to_fit: width_height }
```
tvdeyen, line 77: This is the case for `:^`, BUT this upscales the image, if the original dimensions are smaller than the requested dimensions. Not sure if we want that as default? We probably want `resize_to_limit` as default.

tvdeyen (changes requested):
Currently working on porting Alchemy from Dragonfly to active storage and I think this needs to be changed.

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
waiting-for-dev: >  We probably want resize_to_limit as default.

Thanks, I think you're right, plus it won't change the previous behavior. Fixed.

core/app/models/concerns/spree/active_storage_adapter/attachment.rb
waiting-for-dev: Actually, `image_processing` transforms `resize_to_fill` into `^`:

https://github.com/janko/image_processing/blob/12e7cf50b512d0f4c0735138013c5ea6d87fdee5/lib/image_processing/mini_magick.rb#L74

See also [docs in imagemagick](https://legacy.imagemagick.org/Usage/resize/#fill).

Do I get it right?

tvdeyen: Yes. I see, this seems to be a Dragonfly thing. Really confusing actually, because Dragonfly just forwards the string to `convert -thumbnail` https://markevans.github.io/dragonfly/imagemagick#thumb

tvdeyen (approved):

PR #4320, opened 2022-03-30
Deprecate SolidusFrontend

gsmendoza:
Goal
----

We want to deprecate SolidusFrontend so that developers, especially those
building fresh Solidus apps, will choose to SolidusStarterFrontend instead.

## Background

This is part of an epic to replace the SolidusFrontend gem with SolidusStarterFrontend.

## Screenshots

```
frontend 14:51:29 $ gem install solidus_frontend-3.2.0.alpha.gem
----------------------------------------------------------------------------
DEPRECATION WARNING: SolidusFrontend is deprecated. It will be removed from
the solidus meta-package gem in a Solidus v4. Furthermore, its code will be
extracted from https://github.com/solidusio/solidus to a new repo. Once
extracted, you'll need to explicitly add `solidus_frontend` to your Gemfile
in order to continue using it.

For fresh Solidus applications, we recommend you use SolidusStarterFrontend
instead.
----------------------------------------------------------------------------
Successfully installed solidus_frontend-3.2.0.alpha
Parsing documentation for solidus_frontend-3.2.0.alpha
Done installing documentation for solidus_frontend after 0 seconds
1 gem installed
```

**Checklist:**
- N/A: I have updated Guides and README accordingly to this change (if needed)
- N/A: I have added tests to cover this change (if needed)
- N/A: I have attached screenshots to this PR for visual changes (if needed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,18 @@ config.allow_promotions_any_match_policy = true
 - Add configuration option for `migration_path` [#4190](https://github.com/solidusio/solidus/pull/4190) ([SuperGoodSoft](https://github.com/supergoodsoft/))
 - Deprecate Promotion `any` Match Policy [#4304](https://github.com/solidusio/solidus/pull/4304) ([mamhoff](https://www.github.com/mamhoff))
 
+### Frontend
+
+DEPRECATION WARNING: SolidusFrontend is deprecated. It will be removed from the
+solidus meta-package gem in Solidus v4. Furthermore, its code will be extracted
+from https://github.com/solidusio/solidus to a new repo. Once extracted, you'll
+need to explicitly add `solidus_frontend` to your Gemfile in order to continue
+using it.
+
+For fresh Solidus applications, we recommend you use 
+[SolidusStarterFrontend](https://github.com/solidusio/solidus_starter_frontend) 
+instead.
+
 ## Solidus 3.1.5 (v3.1, 2021-12-20)
 
 - Fix CSRF forgery protection bypass for Spree::OrdersController#populate [GHSA-h3fg-h5v3-vf8m](https://github.com/solidusio/solidus/security/advisories/GHSA-h3fg-h5v3-vf8m)
--- a/frontend/solidus_frontend.gemspec
+++ b/frontend/solidus_frontend.gemspec
@@ -23,6 +23,19 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.5.0'
   s.required_rubygems_version = '>= 1.8.23'
 
+  s.post_install_message = <<~MESSAGE
+    ----------------------------------------------------------------------------
+    DEPRECATION WARNING: SolidusFrontend is deprecated. It will be removed from
+    the solidus meta-package gem in a Solidus v4. Furthermore, its code will be
+    extracted from https://github.com/solidusio/solidus to a new repo. Once
+    extracted, you'll need to explicitly add `solidus_frontend` to your Gemfile
+    in order to continue using it.
+
+    For fresh Solidus applications, we recommend you use SolidusStarterFrontend
+    instead.
+    ----------------------------------------------------------------------------
+  MESSAGE
+
   s.add_dependency 'solidus_api', s.version
   s.add_dependency 'solidus_core', s.version
 


waiting-for-dev:
Thanks, @gsmendoza!

Any chance we could reuse `Spree::Deprecation` module for that? Example: https://github.com/peterberkenbosch/solidus/blob/d2035bfb6644d913e4b76ddae7d72799ef27e936/core/app/models/spree/order.rb#L385

Also, if you like, we could talk about this change and its motivation in the[ `Other important changes` in the Changelog](https://github.com/solidusio/solidus/blob/master/CHANGELOG.md#solidus-320alpha-master-unreleased). If we add it now, we won't have to remember later :slightly_smiling_face:

gsmendoza:
Thanks @waiting-for-dev ! I'll make the following updates:

* Use `Spree::Deprecation`.
* Update Changelog.

frontend/lib/solidus_frontend.rb
```diff
@@ -1,3 +1,12 @@
 # frozen_string_literal: true
 
+require 'spree/deprecation'
+
+deprecation_message = <<~MSG
+  NOTE: SolidusFrontend is deprecated; use SolidusStarterFrontend instead.
```
kennyadsl, line 6: I think this message is not clear. We don't want people currently using `solidus_frontend` to start using the new frontend because they would have to rebuild their frontend entirely. I think what we want to say to people is that, this gem will be removed from the meta-package `solidus` and if they still want to use it, they need to add it manually to the Gemfile, isn't it?

gsmendoza: @kennyadsl How about this:

> NOTE: SolidusFrontend is deprecated. It will be moved out of the `solidus` meta-package gem in a future version of Solidus. Once moved out, you'll need to manually add `solidus_frontend` to your Gemfile in order to continue using it.

> For fresh Solidus applications, we recommend you use SolidusStarterFrontend instead.

kennyadsl: I think the text of the message may work. I'm only a bit concerned that, with this approach, there's no way to remove the deprecation message until we actually remove `solidus_frontend` from `solidus`. Is there a way to check that solidus_frontend is present in the Gemfile and print the message only otherwise? Or do you have any other idea about how to fix this?

waiting-for-dev: > Is there a way to check that solidus_frontend is present in the Gemfile and print the message only otherwise?

That question intrigued me :slightly_smiling_face: It seems we can get the information through:

```ruby
Bundler.locked_gems.dependencies
```

However, it seems to be very slow. I don't know if there's a way to cache it thanks to the `bundler/setup` Rails apps do in `config/boot.rb`

kennyadsl: and probably move at the boot phase so that, even if it's slow it only happens once when the application boots (but maybe it's the same in this file?).

gsmendoza: > I'm only a bit concerned that, with this approach, there's no way to remove the deprecation message until we actually remove solidus_frontend from solidus. 

I'm currently extracting `solidus_frontend` to an external gem. By pointing `solidus_frontend` in the Gemfile to that external gem, we'd be able to remove the deprecation message.

To demonstrate, please see the Solidus app in https://github.com/gsmendoza/external_solidus_frontend_store. In that app's [Gemfile](https://github.com/gsmendoza/external_solidus_frontend_store/blob/main/Gemfile#L62), all the Solidus gems except for `solidus_frontend` point to this branch. In contrast, `solidus_frontend` points to the `solidus_frontend` repo I'm working on: https://github.com/gsmendoza/solidus_frontend.

If we run that app's server, then it wouldn't load the `solidus_frontend` in this Solidus meta package, and as such, wouldn't display the deprecation message.

Does this resolve the concern you have @kennyadsl?

waiting-for-dev: @gsmendoza, I'm not sure I follow. `solidus_frontend` is already a [standalone gem](https://rubygems.org/gems/solidus_frontend). Then, it's a dependency on [solidus meta-gem](https://rubygems.org/gems/solidus).

kennyadsl: Yes, this does solve! 🙂 

You are right, if people switch to the gem in the Gemfile they will be able to get the version without the deprecation notice. There's just one thing to address: we need to understand how the versioning of those two gems will work, because both need to be on RubyGems, and they can't have the same version.

kennyadsl: @waiting-for-dev we can move `solidus_frontend` to another GitHub repository, remove it as a dependency of `solidus`, and start publishing new versions on RubyGems of it from its own new repository only. Do you see any issue with this approach?

waiting-for-dev: Oh, I get it. Yeah, makes sense :slightly_smiling_face:

gsmendoza: > we need to understand how the versioning of those two gems will work, because both need to be on RubyGems, and they can't have the same version.

@kennyadsl I remember your idea before is to give the external gem a different name e.g. `solidus_deprecated_frontend`. If we do that, we can maintain both `solidus_frontend` and `solidus_deprecated_frontend` in RubyGems :)

gsmendoza:
@kennyadsl @waiting-for-dev I updated the deprecation message based on our offline discussion. Appreciate if you can take a look. Thanks!

CHANGELOG.md
```diff
@@ -47,6 +47,10 @@ suffix.
 
 - Add configuration option for `migration_path` [#4190](https://github.com/solidusio/solidus/pull/4190) ([SuperGoodSoft](https://github.com/supergoodsoft/))
 
+### Frontend
+
+- SolidusFrontend is deprecated; use [SolidusStarterFrontend](https://github.com/solidusio/solidus_starter_frontend) instead. SolidusFrontend will be removed in a future version of Solidus.
```
waiting-for-dev, line 52: Again, maybe we can be cautious here and make it clearer that we're not recommending old storefronts built on top of `solidus_frontend` to replace it.

CHANGELOG.md
```diff
@@ -47,6 +47,18 @@ suffix.
 
 - Add configuration option for `migration_path` [#4190](https://github.com/solidusio/solidus/pull/4190) ([SuperGoodSoft](https://github.com/supergoodsoft/))
 
+### Frontend
+
+DEPRECATION WARNING: SolidusFrontend is deprecated. It will be removed from the
+Solidus meta-package gem in a future version of Solidus. Furthermore, its code
```
waiting-for-dev, line 53: I'd say "solidus meta-package" (with initial "s" in lower case), as that's the name of the gem, and people will be less confused.

Maybe we can specify that the Solidus version where it'll be removed will be 4.0.

Other than that, it's very clear :+1:

gsmendoza: > I'd say "solidus meta-package" (with initial "s" in lower case), as that's the name of the gem, and people will be less confused.

I'm okay with that, since "solidus" specifically refers to the gem. But how about SolidusFrontend and Solidus in "in a future version of Solidus"? For those, let's keep the camel case?

waiting-for-dev (approved):
Thanks, @gsmendoza!!

kennyadsl (approved):
[This repository](https://github.com/gsmendoza/eng_310_2022_07_05_store) is a demonstration that the solidus_frontend "as-an-extension" works by just [unpacking](https://github.com/gsmendoza/eng_310_2022_07_05_store/commit/f758f2fc38666d0221cb2beb1754ab84593440eb) the full `solidus` gem and [adding it as a standalone dependency](https://github.com/gsmendoza/eng_310_2022_07_05_store/commit/d8c8f1870d9e226b85a1787e17ff4cfc9da99ec3).


We can go ahead!

PR #4330, opened 2022-04-09
Make install generator compatible with Rails Engines

basicfeatures:
In Rails Engines tasks must be prefixed with `app`.

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -52,10 +52,19 @@ module Solidus
       template 'config/initializers/spree.rb.tt', 'config/initializers/spree.rb'
     end
 
+    def is_engine
+      Dir["lib/*/engine.rb"].any?
+    end
+
     def install_file_attachment
       if options[:active_storage]
         say "Installing Active Storage", :green
-        rake 'active_storage:install'
+
+        unless is_engine
+          rake "active_storage:install"
+        else
+          rake "app:active_storage:install"
+        end
       else
         say "Installing Paperclip", :green
         gsub_file 'config/initializers/spree.rb', "ActiveStorageAttachment", "PaperclipAttachment"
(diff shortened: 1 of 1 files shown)


kennyadsl:
Thanks for the contribution. In which scenario do we need this exactly?

basicfeatures:
Hi! In my case I'm trying to split up my app, it's getting too big. But at the same time I don't want to turn it into a bunch of separate full apps either. Plus they're meant to be sharing a lot of the same resources. So Rails Engines just seemed like the right thing to do :thumbsup:

kennyadsl:
Sorry, I still don't understand well. Solidus is already a Rails Engine and the install generator only runs the first time you install it into your main app.

jarednorman:
@basicfeatures Correct me if I'm wrong.

@kennyadsl I think @basicfeatures has extracted parts of their own application into Rails engines and is running into issues installing Solidus into one of those engines.

For what it's worth, that approach sounds like more trouble than it's worth, but if this is all that's required to support it, that's not so bad.

basicfeatures:
Hey guys!

@jarednorman: Cheers! You can see my use case here:

https://gist.github.com/basicfeatures/d01d773de471fa3bb685ca798adff5b5

Feedback welcome!

It's kinda cool since Solidus already is a Rails Engine, although I can understand the concern if I'm the only one needing this feature.

jarednorman:
After some discussion, we're not sure we want to maintain changes to core to support this use case. Once a change is in core, we have to maintain support for it, and we've never seen a store take this approach and generally don't recommend it. That's not to say that it's not a valid approach, but I think there's a workaround without needing Solidus to support this officially.

As I understand it, you (and anyone else who wants to do this) should be able to follow these steps to get Solidus installed in an engine:

1. Add Solidus to your Gemfile, but point it to your branch on GitHub on your fork with the changes you made.
2. Run the install generator.
3. Update your Gemfile to use the official version of Solidus.

At that point you no longer need the install generator changes and can carry on using the official release of Solidus.

Happy to discuss further if you think we've misjudged the situation. We do appreciate the PR, since it means if anyone else wants to do this they can use this example as a reference. 🙏🏻

basicfeatures:
Hi,

I completely agree. After giving this some more thought, as well as reading stuff like [When to use Engines for breaking up monolith?](https://www.reddit.com/r/rails/comments/gqjmjv/when_to_use_engines_for_breaking_up_monolith/), I've decided to drop engines altogether and just go for a namespaced monolith instead.

Thanks a lot for the help -- more PRs coming up shortly 👍

PR #4332, opened 2022-04-12
Bugfix: Changing Default Addresses

mamhoff:
Solidus' immutable addresses allow "changing" an address by passing its
ID to the `save_in_address_book` method. If one does that with an
address that was marked as default shipping or billing of the user, the
old address would be archived, and the new address would not become the
user's default address. This change addresses that by handling that
situation for both default billing and default shipping addresses.

Prerequisite for #3852 

**Checklist:**


jarednorman (approved):

jarednorman (approved):

kennyadsl:
Thanks @mamhoff!

PR #4334, opened 2022-04-13
Update in-memory shipments of order in order_shipping

tvdeyen:
**Description**

Since we update the shipments state in the database,
but we read the shipsments from memory in the
OrderUpdater#determine_shipment_state we need to set the new
values here on the object as well.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order_shipping.rb
+++ b/core/app/models/spree/order_shipping.rb
@@ -62,18 +62,15 @@ class Spree::OrderShipping
     end
 
     inventory_units.map(&:shipment).uniq.each do |shipment|
-      # Temporarily propagate the tracking number to the shipment as well
-      # TODO: Remove tracking numbers from shipments.
-      shipment.update!(tracking: tracking_number)
-
-      next unless shipment.inventory_units.reload.all? { |iu| iu.shipped? || iu.canceled? }
-      # TODO: make OrderShipping#ship_shipment call Shipment#ship! rather than
-      # having Shipment#ship! call OrderShipping#ship_shipment. We only really
-      # need this `update_columns` for the specs, until we make that change.
-      shipment.update_columns(state: 'shipped', shipped_at: Time.current)
+      if shipment.inventory_units.reload.all? { |iu| iu.shipped? || iu.canceled? }
+        shipment.update!(state: "shipped", shipped_at: Time.current, tracking: tracking_number)
+      else
+        shipment.update!(tracking: tracking_number)
+      end
     end
 
     send_shipment_emails(carton) if stock_location.fulfillable? && !suppress_mailer # e.g. digital gift cards that aren't actually shipped
+    @order.shipments.reload
     @order.recalculate
 
     carton
--- a/core/spec/models/spree/order_shipping_spec.rb
+++ b/core/spec/models/spree/order_shipping_spec.rb
@@ -214,6 +214,21 @@ RSpec.describe Spree::OrderShipping do
       end
     end
 
+    context "when a second shipment is shipped" do
+      let(:order) { create(:order_ready_to_ship) }
+
+      it "sets the order to shipped state" do
+        order.shipping.ship_shipment(order.shipments.first)
+
+        unshipped_shipment = create(:shipment, order: order, state: "ready")
+
+        order.reload
+        order.recalculate
+
+        expect { order.shipping.ship_shipment(unshipped_shipment) }.to(change { order.shipment_state }.from("partial").to("shipped"))
+      end
+    end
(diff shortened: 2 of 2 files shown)


tvdeyen:
@solidusio/core-team I would like to back port this into 2.11, since this is a long standing bug.

tvdeyen:
Failing specs should be fixed with #4338

core/app/models/spree/order_shipping.rb
```diff
@@ -70,7 +70,10 @@ def ship(inventory_units:, stock_location:, address:, shipping_method:,
       # TODO: make OrderShipping#ship_shipment call Shipment#ship! rather than
       # having Shipment#ship! call OrderShipping#ship_shipment. We only really
       # need this `update_columns` for the specs, until we make that change.
-      shipment.update_columns(state: 'shipped', shipped_at: Time.current)
+      attributes = { state: "shipped", shipped_at: Time.current }
+      shipment.update_columns(attributes)
+      # Make sure the in-memory shipments of the order have the correct state for OrderUpdate#determine_shipment_state
+      @order.shipments.detect { |s| shipment.id == s.id }&.assign_attributes(attributes)
```
mamhoff, line 76: A few lines above we use `update!`, so I don't see why we need to go with `update_columns` here. I understand that we want the changes from these inventory unit shipments to be propagated to the order's shipments, but I think we can accomplish that much easier by doing

```
shipment.update!(state: :shipped, shipped_at: Time.current)
@order.shipments.reset
```

The advantage of that approach would be: definitely no changed shipments in the order updater, and more understandability of this code for people looking at it :) What do you think?

tvdeyen: We cannot be certain that there are no changes on the in-memory shipments we gather from the inventory units (that potentially could be from other orders).

If @solidusio/core-team agrees that we take that risk I would love to got with this approach instead.

core/app/models/spree/order_shipping.rb
```diff
@@ -70,7 +70,10 @@ def ship(inventory_units:, stock_location:, address:, shipping_method:,
       # TODO: make OrderShipping#ship_shipment call Shipment#ship! rather than
       # having Shipment#ship! call OrderShipping#ship_shipment. We only really
       # need this `update_columns` for the specs, until we make that change.
-      shipment.update_columns(state: 'shipped', shipped_at: Time.current)
```
kennyadsl, line 73: The comment above state that we need this `update_columns` for specs only. Do you think it's still valid?

core/app/models/spree/order_shipping.rb
```diff
@@ -70,7 +70,10 @@ def ship(inventory_units:, stock_location:, address:, shipping_method:,
       # TODO: make OrderShipping#ship_shipment call Shipment#ship! rather than
```
jarednorman, line 70: This is my new favourite comment in Solidus. Props to @kennyadsl for managing to read it out without stumbling in the core team meeting today. 👏🏻 

CleanShot 2022-04-13 at 08 18 26

tvdeyen: You made my day

core/app/models/spree/order_shipping.rb
waiting-for-dev: We should trust the test suite. Otherwise, there's no point in it :man_shrugging: Do you think we need to add a more specific test?

kennyadsl: I made a test locally with the following code (improvable for sure):

```ruby
if shipment.inventory_units.reload.all? { |iu| iu.shipped? || iu.canceled? }
  shipment.update!(state: 'shipped', shipped_at: Time.current, tracking: tracking_number)
else
  shipment.update!(tracking: tracking_number)
end
```

and made some tests with `binding.irb` right after this execution. Apparently, at this point, the value of the shipment associated to the `@order` (taken with`@order.shipments.detect { |s| shipment.id == s.id }`) is already updated with the value just set with the `update!` call. No reload needed and, if I got it correctly, without the reload we won't have any risk to clear existing changes on the in-memory shipment. isn't it?

tvdeyen: Tried that.

Unfortunately the `order.shipment_state` is still `"partial"`, unless I also call

```rb
@order.shipments.reload
```

right before `@order.recalculate`

I pushed an update. Tell me what you think @kennyadsl @mamhoff @jarednorman @waiting-for-dev

jarednorman (approved):

waiting-for-dev (approved):

PR #4338, opened 2022-04-13
Respect new preloader API in Rails 7

mamhoff:
The API for ActiveRecord's Preloader changes in Rails 7. This commit
makes it work in both Rails 6 and 7.

**Checklist:**


tvdeyen (approved):
Thanks 🙏🏻

kennyadsl (approved):
We missed [this deprecation warning](https://github.com/rails/rails/blob/16e78466c5551cc68bd3df2adbd7baa8bbffd8d5/activerecord/lib/active_record/associations/preloader.rb#L107) in the test logs. It happens, thanks for promptly fixing the issue. ❤️

PR #4339, opened 2022-04-14
[v2.11] Fix deprecated deliver_order_confirmation_email method

mamhoff:
This method on the Spree::Order class was trying to call the `order`
method. I think what it means to say is  `self`.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -476,8 +476,8 @@ module Spree
         "Spree::MailerSubscriber#order_finalized.",
         caller(1)
 
-      Spree::Config.order_mailer_class.confirm_email(order).deliver_later
-      order.update_column(:confirmation_delivered, true)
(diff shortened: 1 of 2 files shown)


tvdeyen (approved):

kennyadsl (approved):
This needs to be released on 2.11

core/app/models/spree/order.rb
```diff
@@ -476,7 +476,7 @@ def deliver_order_confirmation_email
         "Spree::MailerSubscriber#order_finalized.",
         caller(1)
 
-      Spree::Config.order_mailer_class.confirm_email(order).deliver_later
+      Spree::Config.order_mailer_class.confirm_email(self).deliver_later
       order.update_column(:confirmation_delivered, true)
```
mamhoff, line 480: here as well. I'm writing a spec.

waiting-for-dev:
Thanks!

PR #4252, opened 2022-01-28
Add public interface to fetch registered events

waiting-for-dev:
**Description**

It adds value so that users don't need to scan source code or consult
external docs.

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
- [x] I have updated Guides and README accordingly to this change (if needed)
- [x] I have added tests to cover this change (if needed)
- [ ] I have attached screenshots to this PR for visual changes (if needed)

PR #4246, opened 2022-01-18
Cosmetic changes to the unknown event message

waiting-for-dev:
- Remove line break.
- Wrap each known event within single quotes (').

- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
- [ ] I have updated Guides and README accordingly to this change (if needed)
- [x] I have added tests to cover this change (if needed)
- [ ] I have attached screenshots to this PR for visual changes (if needed)

PR #4216, opened 2021-11-29
Use default spree event adapter

peterberkenbosch:
**Description**
On installing this version (currenlty unreleased) of Solidus, we
no longer support the ActiveSupport Notification adaptor, but rather
the new `Spree::Event::Adapters::Default`

This commit add that configuration to the spree.rb template so every
generated install is using the new defaults.

**Checklist:**
- [ ] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [ ] I have added a detailed description into each commit message
- [ ] I have updated Guides and README accordingly to this change (if needed)
- [ ] I have added tests to cover this change (if needed)
- [ ] I have attached screenshots to this PR for visual changes (if needed)

PR #4342, opened 2022-04-15
Use Omnes for pub/sub

waiting-for-dev:
**Description**

This PR removes the new event bus adapter and delegates the pub/sub backbone to an external dependency: [omnes](https://github.com/nebulab/omnes). Omnes is, in fact, an extraction of that code into a standalone library, which has given us the possibility to add more features from which Solidus can benefit.

The changes can be more easily reviewed commit by commit:

- To begin with, we revert all the changes we've been doing so far regarding the new event bus system. That undoes:
  - https://github.com/solidusio/solidus/pull/4130
  - https://github.com/solidusio/solidus/pull/4203
  - https://github.com/solidusio/solidus/pull/4204
  - https://github.com/solidusio/solidus/pull/4226
  - https://github.com/solidusio/solidus/pull/4176
  - https://github.com/solidusio/solidus/pull/4130
  - https://github.com/solidusio/solidus/pull/4246
  - https://github.com/solidusio/solidus/pull/4252
  - https://github.com/solidusio/solidus/pull/4215
  - https://github.com/solidusio/solidus/pull/4216
- Add add omnes as a dependency. As said, that contains all the reverted code and new features.
- Use omnes internally by default from Solidus 3.2.0 while still supporting the legacy system through a configuration setting. Creates an omnes subscriber to replace the legacy subscriber module.
- Update test helpers to stub the new `Spree::Bus` instance. That is something that won't be on Omnes, as it's a specific stub for our global bus instance.
- Warn when the legacy system is used. We'll update the deprecation message with a link to an upgrade link once that's ready.

**How will Solidus benefit from it?**

Being free from the need to support two completely different adapters through the same API enormously simplifies our code. We've also been able to add new features without constraints. Omnes has better support for debugging, testing, and async (see its README for details). We can potentially benefit from contributors using it on other kinds of projects.

There's still a place to take more advantage of omnes in future work:

- Create an internal configurable adapter to allow users to run Solidus events in their preferred async adapter.
- Use instance events and add an option to store their serialization in the database.

**How will people upgrade from the legacy system to omnes?**

As pointed out, we'll write an upgrade guide and add a link to it in the deprecation message. As heads-up:

- Replace references of `Spree::Event` to `Spree::Bus`.
- Rename `#fire` to `#publish`.
- Register custom events in an initializer.
- Transform module subscribers into Omnes subscribers and load them in an initializer.
- Don't provide a block to `#publish`. Instead, run whatever you want before it.
- Don't use regular expression subscriptions as they are an anti-pattern (they can catch undesired events if they're added in the future). Probably, what most people will need instead is omnes' `#subscribe_to_all`. If they really want it, they can still create a custom matcher on omnes.

**How will people upgrade from master to omnes?**

All the work on the new event bus adapter was on master, but it hadn't been released yet.

Anyway, the migration from master will be pretty straightforward. We'll happily support it, and once this gets merged, we'll post a message on Slack to let people know they can come to us (more explicitly to @waiting-for-dev) if they have any issues.

**Checklist:**


waiting-for-dev:
cc @forkata

kennyadsl (approved):
Nicely implemented, thanks @waiting-for-dev. 

🚌 🚌 🚌

core/lib/spree/bus.rb
```diff
@@ -0,0 +1,20 @@
...
+  # system {see Spree::AppConfiguration#use_legacy_events}.
+  Bus = Omnes::Bus.new.tap do |bus|
+    bus.define_singleton_method(:publish) do |*args, **kwargs, &block|
+      if Spree::Config.use_legacy_events
+        Spree::Event.fire(*args, **kwargs, &block)
+      else
+        super(*args, **kwargs, caller_location: caller_locations(1)[0], &block)
+      end
+    end
+  end
```
jarednorman, line 19: Not a strong opinion, but I know `tap` can be a stumbling block for people not familiar with it. I like `tap`, but only use it when it actually makes the code cleaner or avoids polluting a scope with too many local variables. The use here doesn't really seem to add much to me, since we could write it like this instead and I think it's just as good, but doesn't require you to be familiar with `tap`:

```suggestion
  Bus = Omnes::Bus.new
  Bus.define_singleton_method(:publish) do |*args, **kwargs, &block|
    if Spree::Config.use_legacy_events
      Spree::Event.fire(*args, **kwargs, &block)
    else
      super(*args, **kwargs, caller_location: caller_locations(1)[0], &block)
    end
  end
```

waiting-for-dev: Fair enough. Changed. Thanks, @jarednorman.

waiting-for-dev:
Updated with a fix for autoloading subscribers (see the addition of `Spree::Bus.clear` on the engine file). It required bumping omnes to v0.2.1.

core/lib/spree/bus.rb
```diff
@@ -0,0 +1,19 @@
...
+  # use it.
+  #
+  # It has some modifications to support internal usage of the legacy event
+  # system {see Spree::AppConfiguration#use_legacy_events}.
+  Bus = Omnes::Bus.new
+  Bus.define_singleton_method(:publish) do |*args, **kwargs, &block|
+    if Spree::Config.use_legacy_events
+      Spree::Event.fire(*args, **kwargs, &block)
+    else
+      super(*args, **kwargs, caller_location: caller_locations(1)[0], &block)
```
forkata, line 16: I was a bit puzzled as to why we are overriding the `caller_location` here. Would it be worth leaving a comment that this is intended to omit this from the stack trace, or just dropping that in order to ease debugging for anyone who is getting the new bus implementation and trying to track down why?

forkata (commented):
Nice work here @waiting-for-dev!

core/lib/spree/bus.rb
waiting-for-dev: I went with the addition of a comment. We do need to go one level higher in the stack. Otherwise, it'll be useless for debugging, as all events will look as published from this line (see https://github.com/nebulab/omnes/blob/bb73fce75414816e304a7402efbea655b82c857b/lib/omnes/bus.rb#L187)

forkata: Thanks for clarifying this @waiting-for-dev, your decision here makes a lot more sense now.

core/lib/spree/core/engine.rb
```diff
@@ -44,14 +44,29 @@ class Engine < ::Rails::Engine
...
+          app.reloader.to_prepare do
+            Spree::Event.activate_autoloadable_subscribers
+          end
+
+          app.reloader.before_class_unload do
+            Spree::Event.deactivate_all_subscribers
+          end
+        else
+          app.reloader.to_prepare do
+            Spree::Bus.clear
```
forkata, line 59: ❓ Does this change anything with regards to where applications or Solidus extensions should be registering any custom events they want to introduce that use the `Spree::Bus` 🤔

waiting-for-dev: That's a great question. More than a change, it's a fix. To survive code reloading, events need to be registered, and subscriptions must be added within a `#to_prepare` block (which runs one on production but at every reload on development). Solidus owns `Spree::Bus`, so it clears it at every reload. Then, solidus itself, extensions, and applications can just call `#register` & `#subscribe_to` within their own `#to_prepare` blocks, which will run later in time.

waiting-for-dev:
Rebased from master, and updated Omnes version to 0.2.2 to support Ruby 2.5 & Ruby 2.6, which are still supported on Solidus.

spaghetticode:
This looks good to me, thanks Marc for the usual great work! What about removing the spurious fixup commit that updates `omnes`? Also, there's another commit that changes its version after the commit that introduces the gem.

waiting-for-dev:
> This looks good to me, thanks Marc for the usual great work! What about removing the spurious fixup commit that updates `omnes`? Also, there's another commit that changes its version after the commit that introduces the gem.

Thanks, @spaghetticode. Oh, sorry, I missed that (forgot the rebase). I'll fix it 🙂

jarednorman (approved):

waiting-for-dev:
> This looks good to me, thanks Marc for the usual great work! What about removing the spurious fixup commit that updates `omnes`? Also, there's another commit that changes its version after the commit that introduces the gem.

Thanks for spotting it, @spaghetticode. Fixed!

spaghetticode:
> Thanks for spotting it, @spaghetticode. Fixed!

I still see the fixup commit: https://github.com/solidusio/solidus/pull/4342/commits/cb0e9101bb8fb9282f770e74ffc6908000740c55

To be super explicit, here's the three commits related to the omnes dependency:

* introduces 0.20: https://github.com/solidusio/solidus/pull/4342/commits/690a19813f57071e160cdbd5f8c5e1b37ab2142d

* introduces 0.2.1: https://github.com/solidusio/solidus/pull/4342/commits/07512498380113099f198067b3a492bd64a4be19

* introduces 0.2.2: https://github.com/solidusio/solidus/pull/4342/commits/cb0e9101bb8fb9282f770e74ffc6908000740c55

waiting-for-dev:
> > Thanks for spotting it, @spaghetticode. Fixed!
> 
> I still see the fixup commit: [cb0e910](https://github.com/solidusio/solidus/commit/cb0e9101bb8fb9282f770e74ffc6908000740c55)
> 
> To be super explicit, here's the three commits related to the omnes dependency:
> 
>     * introduces 0.20: [690a198](https://github.com/solidusio/solidus/commit/690a19813f57071e160cdbd5f8c5e1b37ab2142d)
> 
>     * introduces 0.2.1: [0751249](https://github.com/solidusio/solidus/commit/07512498380113099f198067b3a492bd64a4be19)
> 
>     * introduces 0.2.2: [cb0e910](https://github.com/solidusio/solidus/commit/cb0e9101bb8fb9282f770e74ffc6908000740c55)

My bad @spaghetticode. I forced the push to the wrong repo. Everything should be fine now.

spaghetticode (approved):
thank you Marc!

waiting-for-dev:
Rebased from master and made a minor change to adopt the new way to define versioned preferences (see https://github.com/solidusio/solidus/pull/4375).

waiting-for-dev:
See https://github.com/solidusio/solidus/discussions/4380

PR #4343, opened 2022-04-20
Unify sandbox creation source

Naokimi:
**Description**
The sandbox apps for [solidus](https://github.com/solidusio/solidus/blob/master/bin/sandbox) and [solidus' extensions](https://github.com/solidusio/solidus_dev_support/blob/master/lib/solidus_dev_support/templates/extension/bin/sandbox.tt) have different code leading to different creation outcomes. For example solidus' sandbox has a docker-compose development environment, but solidus extension's sandbox doesn't. 

This PR fixes the discrepancies by [creating a new repo with a bash script on it](https://github.com/Naokimi/solidus-sandbox-script/tree/master), which can be called to create the sandbox app. Since future updates will be made at a single point, this will ensure alignment.

Once the PR is approved I will transfer ownership of the bash script to Solidus and modify the PR to point to the correct URL before merging.

Related PRs:
- https://github.com/solidusio/solidus_dev_support/pull/182

**Checklist:**


jarednorman (changes requested):
~Hardcoding this so that it sources a file that's on a repository that we don't control is a serious security vulnerability and not an acceptable approach.~ Should have read the description better.

jarednorman:
It would be great if this could still live in one of these repositories and if the correct version was used for the runnning version of Solidus. Can you think of an approach that would accomplish that? I don't really like the idea of it being hardcoded to use the latest version in a separate repo in case there are eventually incompatibilities.

waiting-for-dev:
Thanks for your contribution, @Naokimi. You're right; some inconsistencies ideally should be fixed. However, my concern is that we're trying to encompass two different things with a common abstraction that could easily break. The script for Solidus is intended to be used as it is. However, the script for extensions belongs to a template that extensions creators can modify. If they need to add something in-between the core script, they would need to copy-paste it, and we'd be back at square one.

As @jarednorman said, the discrepancies between versions are also a problem.

I also think we can improve something here, but I'm not sure exactly how. Maybe we could have that script "rubyfied" and use the bits we need from `solidus_dev_support` to build its own sandbox application.

kennyadsl:
@Naokimi we are designing a new way of dealing with extensions, and the sandbox generation for core and extensions will be part of that plan. We will consider this PR as inspiration in that work, but for the moment, it's better not to change the status quo, so I'm going to close this one, thanks!

PR #4356, opened 2022-04-22
Readd `config.cache_classes` on test env and remove `with_model` dep

waiting-for-dev:
Rails' config setting `cache_classes` was recently set to `false` in the
dummy app (see
https://github.com/solidusio/solidus/commit/ef9b5ab9fcbc20290d2d6ec7a5437e73dd0e1db2).
That was due to a misunderstanding, as Rails recommends it be true
except when `spring` is present. From
https://guides.rubyonrails.org/configuring.html#config-cache-classes:

> Controls whether or not application classes and modules should be
> reloaded if they change. When the cache is enabled (true), reloading
> will not occur. Defaults to false in the development environment, and
> true in production. In the test environment, the default is false if
> Spring is installed, true otherwise.

That wasn't such a big deal until now. However, if we try to use the new
Ruby 3.1 in conjunction with `cache_classes` in the test environment,
many errors related to STI classes not being found pop out.

However, having the correct `false` value on Rails 7 raises an error
sourced on the `with_model` gem. See
https://github.com/Casecommons/with_model/issues/35.

Because of that, we remove the `with_model` dependency so that our path
to support Ruby 3.1 is unblocked. We take different decisions to remove
the occurrences of `with_model` usage on Solidus:

- Tests on `Spree::Api::ResourceController`. That controller was only
used on `Spree::Api::UsersController`. Therefore, it was no longer a
valid abstraction:
  - We removed the tests altogether.
  - We moved the missing pieces to `UsersController`.
  - We removed the soft deletable specific test and added one to `UsersController`.
  - We marked `ResourceController` as deprecated (not removing it just in case some store uses it).

- Tests on `Spree::Admin::ResourceController`, `Spree::CalculatedAdjustements`,
`Spree::Validations::DbMaximumLengthValidtor` &
`Spree::WalletPaymentSource`: we use a custom logic for the same
behavior provided by `with_model`.

- Tests on `Spree::SoftDeletable`: we use `Spree::Product` as a fixture.


waiting-for-dev:
Moved to https://github.com/solidusio/solidus/pull/4358

PR #4363, opened 2022-04-26
Rename spree_products.discontinue_on into spree_products.available_until

mamhoff:
In our business, the semantics of availability and discontinuation are
not aligned with Solidus' nomenclature: A product can both be available
and discontinued (i.e. we still have stock of it, but won't be producing
any more).

When communicating the meaning of the discontinue_on column, it's often
hard (because there is a date when we decide to discontinue, and there's
another date when we sell out and the product becomes unavailable).

From reading the code, Solidus' `discontinue_on` is actually a mirror of
`available_until`.

What this commit does is it adds an `available_until` column, populates that with the data from `discontinue_on` and changes all scopes within Solidus to reference the new column. In the next major upgrade, we can then remove the `discontinue_on` column.

**Checklist:**
- [x] I have adapted tests to cover this change (if needed)

### DIFF
--- a/backend/app/views/spree/admin/products/_form.html.erb
+++ b/backend/app/views/spree/admin/products/_form.html.erb
@@ -123,14 +123,14 @@
         <% end %>
       </div>
 
-      <div data-hook="admin_product_form_discontinue_on">
-        <%= f.field_container :discontinue_on do %>
-          <%= f.label :discontinue_on %>
-          <%= f.field_hint :discontinue_on %>
+      <div data-hook="admin_product_form_available_until">
+        <%= f.field_container :available_until do %>
+          <%= f.label :available_until %>
+          <%= f.field_hint :available_until %>
 
-          <%= render "spree/admin/shared/datepicker", f: f, date_attr: :discontinue_on %>
+          <%= render "spree/admin/shared/datepicker", f: f, date_attr: :available_until %>
 
-          <%= f.error_message_on :discontinue_on %>
+          <%= f.error_message_on :available_until %>
         <% end %>
       </div>
 
--- a/backend/spec/features/admin/products/products_spec.rb
+++ b/backend/spec/features/admin/products/products_spec.rb
@@ -314,12 +314,12 @@ describe "Products", type: :feature do
         expect(Spree::Product.last.available_on).to eq('Tue, 25 Dec 2012 00:00:00 UTC +00:00')
       end
 
-      it 'should correctly update discontinue_on', :js do
+      it 'should correctly update available_until', :js do
         visit spree.admin_product_path(product)
-        fill_in "product_discontinue_on", with: "2020/12/4"
+        fill_in "product_available_until", with: "2020/12/4"
         click_button "Update"
         expect(page).to have_content("successfully updated!")
-        expect(product.reload.discontinue_on.to_s).to eq('2020-12-04 00:00:00 UTC')
+        expect(product.reload.available_until.to_s).to eq('2020-12-04 00:00:00 UTC')
       end
 
       context "when there is a default tax category" do
--- a/core/app/models/spree/product.rb
+++ b/core/app/models/spree/product.rb
@@ -72,6 +72,9 @@ module Spree
         .where(master: { spree_prices: Spree::Config.default_pricing_options.desired_attributes })
     }
 
+    alias_attribute :discontinue_on, :available_until
+    deprecate discontinue_on: :available_until, deprecator: Spree::Deprecation
+
     def find_or_build_master
       master || build_master
     end
@@ -176,7 +179,7 @@ module Spree
     #
     # @return [Boolean] true if this product is available
     def available?
-      !deleted? && available_on&.past? && !discontinued?
+      !deleted? && available_on&.past? && !unavailable?
     end
 
     # Determines if product is discontinued.
@@ -185,9 +188,12 @@ module Spree
     # is not nil and in the past.
     #
     # @return [Boolean] true if this product is discontinued
-    def discontinued?
-      !!discontinue_on&.past?
+    def unavailable?
+      !!available_until&.past?
     end
+
+    alias_method :discontinued?, :unavailable?
+    deprecate discontinued?: :unavailable?, deprecator: Spree::Deprecation
 
     # Poor man's full text search.
     #
--- a/core/app/models/spree/product/scopes.rb
+++ b/core/app/models/spree/product/scopes.rb
@@ -185,8 +185,8 @@ module Spree
           def self.available(available_on = nil)
             with_master_price
               .where("#{Spree::Product.quoted_table_name}.available_on <= ?", available_on || Time.current)
-              .where("#{Spree::Product.quoted_table_name}.discontinue_on IS NULL OR" \
-                     "#{Spree::Product.quoted_table_name}.discontinue_on >= ?", Time.current)
+              .where("#{Spree::Product.quoted_table_name}.available_until IS NULL OR" \
+                     "#{Spree::Product.quoted_table_name}.available_until >= ?", Time.current)
           end
           search_scopes << :available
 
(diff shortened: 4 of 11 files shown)


BenAkroyd (approved):
:shipit: beautiful

tvdeyen (approved):
Makes much more sense. Thank you

jarednorman (dismissed):
Agreed.

waiting-for-dev:
Agree, but technically that's not a backward compatible change, as queries that go directly to the database, bypassing `alias_attribute` will be broken. Also, I haven't checked it, but I'm afraid some other features, like dirty attributes or method `#attribute_names`, won't pick aliases, so it could also break some code.

We could put it on hold after the next major or create a soft deprecation strategy adding instead of renaming a column.

mamhoff:
If we do that, how do we deal with the two columns going out of sync? And how would things be different than now after the next major version? When is the next major release planned?

It took years for Solidus 3 to be released. I appreciate the commitment to stability in this project, but in cases like this where there's cheap progress to be had at the expense of two search and replace operations in the host app, I sometimes think the deprecation strategies we try to find up with generate a large (and tbh frustrating) burden on contributors. 

Solidus apps are complicated beasts with many interfaces (as ruby allows metaprogramming). Which parts of the interface do we consider stable API, and where do we allow ourselves change? Is this documented somewhere?

kennyadsl:
> Solidus apps are complicated beasts with many interfaces (as ruby allows metaprogramming). Which parts of the interface do we consider stable API, and where do we allow ourselves change? Is this documented somewhere?

Due to the nature of the project, as you said, this is left to the core team common sense. I think we can do an effort, trying to document it a little bit, at least as a series of principles. 

I know how frustrating this could be because I was on the other side (and @mamhoff, you were on this side 🙂) but I think stability and doing our best to be SemVer compliant is still one of the best things about Solidus and what distinguishes us from the other platforms.

On a side note, let me also add that in the past, we were releasing new Solidus major versions based on some more marketing initiatives and I think it's time to stop doing that and release when we have something breaking. We were discussing releasing the next major soon because we want to dismiss some old things like the fronted as a gem. With soon, I think this could be right after the next minor, so after 3.2 there could be Solidus 4.

That said, we discussed this in the core team meeting and we agreed that this can definitely wait for the next major in this form. If we find a way to have this at least compatible with the Rails public methods (like dirty etc) we can re-evaluate this. 

But let me thank you on the behalf of the whole community for the great changes you are lately proposing to the platform, we really value your contributions and the last thing we want is to make you frustrated about continuing contributing. We will do our best to have these changes in Solidus as soon as possible.

mamhoff:
I've reworked the PR to not rename the column, but to add an additional `available_until` column that is updated alongside the `discontinue_on` column. That way we can safely deprecate and delete the old column with the next major version. What do you think?

core/db/migrate/20220426165309_add_available_until_to_spree_products.rb
```diff
@@ -0,0 +1,14 @@
...
+  class SpreeProduct < ActiveRecord::Base
+    self.table_name = "spree_products"
+  end
+  def up
+    add_column :spree_products, :available_until, :datetime
+    SpreeProduct.update_all("available_until = discontinue_on")
+  end
+
+  def down
+    SpreeProduct.update_all("discontinue_on = available_until")
```
kennyadsl, line 11: This should not be needed if they are in sync, right?

kennyadsl (commented):
Thanks, I left a question but overall, I think this should work now!

core/db/migrate/20220426165309_add_available_until_to_spree_products.rb
mamhoff: True, removed!

kennyadsl (approved):
Maybe OT for this PR, since we already had the same issue before, but I noticed that we forgot to add this attribute in the [products API response](https://github.com/solidusio/solidus/blob/0e410a57b62792e119c7e20f5bb0a29233af9286/api/lib/spree/api_configuration.rb#L8). Maybe we can take advantage of the change and add it, but I'm also fine doing this later in another PR. 

Thanks, @mamhoff. I will raise this again in the next core team meeting (tomorrow), just to be sure we have the approval of everyone who was concerned.

kennyadsl:
I was discussing this offline with @waiting-for-dev and we realized that this still won't work with `update_column` and `update_all`. Not saying it's a trade-off we won't accept but reporting here for completeness.

waiting-for-dev:
> I was discussing this offline with @waiting-for-dev and we realized that this still won't work with `update_column` and `update_all`. Not saying it's a trade-off we won't accept but reporting here for completeness.

For more context, this is a script that reproduces how syncing columns can be bypassed using public AR API. Just execute with `ruby file.rb`. The second assertion will fail:

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org/'
  gem 'activerecord'
  gem 'sqlite3'
end

require 'active_record'

ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: 'bypassing.db')

ActiveRecord::Schema.define do
  create_table :dummies, force: true do |t|
    t.string :field_one
    t.string :field_two
    t.timestamps
  end
end

class Dummy < ActiveRecord::Base
  def field_one=(value)
    write_attribute(:field_two, value)
    super
  end
end

require 'minitest/autorun'

class BypassingTest < Minitest::Test
  def test_bypass
    dummy = Dummy.create(field_one: 'foo')
    dummy.update_column(:field_one, 'bar')
    dummy.reload

    assert_equal(
      'bar',
      dummy.field_one
    )
    assert_equal(
      'bar',
      dummy.field_two
    )
  end
end
```

mamhoff:
I'm fine adding the attribute to the API, no problem. Regarding the deprecation strategy though: Do you have a suggested solution?

kennyadsl:
I don't think there's a solution here, or I don't see it now. I think it's a matter to understand if we want to accept this trade-off or wait for the next major. I will let you know tomorrow.

kennyadsl:
Hey @mamhoff, thanks again for this work and for adjusting the PR to try to make it backward-compatible. We think that this still needs to wait for the next major to be merged. We are discussing releasing very soon (let's say in the next couple of months, just to give you an idea). Thanks for understanding.

mamhoff:
Should I revert to the previous version of this PR then for the major release? It is unfortunate that users won't be warned about the disappearance of the column, but I don't think I'm willing to keep updating this - or a related - PR until Solidus 5 comes out :)

kennyadsl:
@mamhoff Yes, can you please revert, please. We won't make any deprecation warning for this change, but I think it's legit to expect this kind of change in a major release.

mamhoff:
Reverted. I've left the `alias_attribute` and deprecation warnings in, so at least in some circumstances people get deprecation warnings.

mamhoff:
Would Solidus 4 then have this version, or the previous iteration?

Alberto Vena ***@***.***> schrieb am Fr., 6. Mai 2022, 12:02:

waiting-for-dev:
@mamhoff, yes, this is on hold until v4

kennyadsl:
@mamhoff hello! With a lot of late, but we are close to releasing Solidus 4.0. Could you please let me know if you are still interested in having this merged? Please, let me know if you have time to work on rebasing this, otherwise we'll see if we can move ahead on our own. Thanks again for your patience!

mamhoff:
Rebased from `master`. In our store, we have now settled on `retired_on` with a `discontinue_on` state before that, but for Solidus core I think that having the two states mirror each other makes a lot of sense still.

kennyadsl:
Given that you seem to have found a solution "good enough", we would feel more confident not merging this. 

The reality is that we couldn't find a safe way to rename a column in use with a simple migration, and the change here, although important, doesn't look to be that critical. We'll put some thoughts on finding a generic solution for this problem, and apply that here. 

Sorry for "promising" to have this merged in v4! 🙏

mamhoff:
Ok, closing.

PR #4373, opened 2022-05-16
Handle empty values as quantity adding to cart via API

kennyadsl:
**Description**

Right now, when empty values are passed as quantity in the API call that creates line items, the line item is populated with quantity 0.

This is not the expected behavior, as it makes no sense to have a line item with 0 as quantity.

Still, if someone wants to force the value to 0, this is still possible but 0 needs to be explicitly passed as quantity value. A spec has been added for this scenario as well.

**Checklist:**

### DIFF
--- a/api/app/controllers/spree/api/line_items_controller.rb
+++ b/api/app/controllers/spree/api/line_items_controller.rb
@@ -13,7 +13,7 @@ module Spree
         variant = Spree::Variant.find(params[:line_item][:variant_id])
         @line_item = @order.contents.add(
           variant,
-          params[:line_item][:quantity] || 1,
+          params[:line_item][:quantity].presence || 1,
           options: line_item_params[:options].to_h
         )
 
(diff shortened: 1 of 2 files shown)


jarednorman (approved):

waiting-for-dev:
Thanks, @kennyadsl. However, wouldn't it make more sense to completely ignore the item instead of setting its quantity to 1? I'm afraid it could have undesired consequences for storefronts that poorly sanitize what they send to the server.

kennyadsl:
Leaving a small report before closing, I didn't have the proper time and mindset to complete this exploration. By the way, this is not the right approach to solve the problem.

`normalize_quanity` is there mainly for when someone edits a line item. having `0` is acceptable, because having the line item to 0 is what triggers the actual removal of the line item from the order. 

That's probably the first thing we need to change: line items need to be removed from the cart in a proper, specific way, which is not setting their quantity to 0.

PR #4375, opened 2022-05-17
Fix update generator taking non-comparable values as changes in a default

waiting-for-dev:
Preference defaults are declared at the class level. To allow them to
reference other parts of the code base regardless of the loading order,
they're internally wrapped within a proc (whey they aren't already one)
to be lazily evaluated. When the default for a preference is itself a
proc, it needs to be declared one level nested to bypass that first
outer layer. E.g.

```ruby
preference :foo, :proc, default: proc { proc { true } }
```

The `solidus:update` generator compared the resolved preferences for
equality for different Solidus versions to add changes in the
`new_solidus_defaults.rb` initializer. However, procs can't be compared
in that way (and the same is true for other mutable types). E.g:

```ruby
proc { true } == proc { true } => false
```

Because of that, the update generator was fooled and added those kinds
of settings as a change. To reproduce, run `bin/rails g solidus:update`
from the parent commit and see how the generated
`Spree::Backend::Config.configure` block in
`config/initializers/new_solidus_defaults.rb` contains something similar
to:

```ruby
config.frontend_product_path = #<Proc:0x00007facd4c30928 /home/solidus_user/app/backend/lib/spree/backend_configuration.rb:12 (lambda)>
```

With this commit, we're moving the information of the changes in default
values from the value itself (which, as told, is not reliable) to make
it explicit to the class that defines the preference. In that way, the
update generator can directly ask for those and do its job. We're
introducing a new `.versioned_preference` method in the configuration
class that, besides doing the same as the `.preference` method, helps
keep track.

**Checklist:**

### DIFF
--- a/core/lib/spree/preferences/configuration.rb
+++ b/core/lib/spree/preferences/configuration.rb
@@ -38,6 +38,7 @@ module Spree::Preferences
     #     defaults. Set via {#load_defaults}
     attr_reader :loaded_defaults
 
+    # @api private
     attr_reader :load_defaults_called
 
     def initialize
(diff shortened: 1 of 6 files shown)


aldesantis (approved):

kennyadsl (approved):
👍

PR #4376, opened 2022-05-17
Fix issues loading serialized logs

waiting-for-dev:
`Spree::LogEntry` loads details from a YAML string that can potentially
contain any serialized Ruby object if users have overridden some part of
the codebase.

New Psych (YAML parser) version packed with Ruby 3.1 requires that, by
default, users are explicit about which type of classes are allowed to
be loaded (see ruby/psych@cb50aa8 and ruby/psych@1764942).

On 008168c, we updated our code to allow instances of
`ActiveMerchant::Billing::Response`. However, as [noted in a
comment](https://github.com/solidusio/solidus/commit/008168c962bb75370b70c12315515710f35fe1bc#r73483078),
it seems we're missing, at least, instances of
`ActiveSupport::TimeWithZone`.

We need to add all classes used internally and leave a door open for
users to add their own classes (via a new `log_entry_permitted_classes`
preference).

On top of that, by default, `YAML.safe_load` won't allow YAML aliases to be parsed.
However, we need them at least for older versions of serialized
`ActiveSupport::TimeWithZone`.

We add a configuration option so that users can still disable it.
Depending on the source of the YAML, parsing aliases could open gates to
entity expansion attacks (a DoS created by nesting self-references).

**Checklist:**

### DIFF
--- a/core/app/models/spree/log_entry.rb
+++ b/core/app/models/spree/log_entry.rb
@@ -39,6 +39,30 @@ module Spree
       end
     end
 
+    # Raised when YAML contains aliases and they're not enabled
+    class BadAlias < RuntimeError
+      attr_reader :psych_exception
+
+      def initialize(psych_exception:)
+        @psych_exception = psych_exception
+        super(default_message)
+      end
+
+      private
+
+      def default_message
+        <<~MSG
+          #{psych_exception.message}
+
+          You can explicitly enable aliases in config/initializers/spree.rb. E.g:
+
+          Spree.config do |config|
+            config.log_entry_allow_aliases = true
+          end
+        MSG
+      end
+    end
+
     def self.permitted_classes
       CORE_PERMITTED_CLASSES + Spree::Config.log_entry_permitted_classes.map(&:constantize)
     end
@@ -46,9 +70,15 @@ module Spree
     belongs_to :source, polymorphic: true, optional: true
 
     def parsed_details
-      @details ||= YAML.safe_load(details, permitted_classes: self.class.permitted_classes)
+      @details ||= YAML.safe_load(
+        details,
+        permitted_classes: self.class.permitted_classes,
+        aliases: Spree::Config.log_entry_allow_aliases
+      )
     rescue Psych::DisallowedClass => e
       raise DisallowedClass.new(psych_exception: e)
+    rescue Psych::BadAlias => e
+      raise BadAlias.new(psych_exception: e)
     end
   end
 end
--- a/core/app/models/spree/log_entry.rb
+++ b/core/app/models/spree/log_entry.rb
@@ -2,10 +2,53 @@
 
 module Spree
   class LogEntry < Spree::Base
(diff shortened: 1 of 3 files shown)


aldesantis (approved):

kennyadsl (approved):
Thanks @waiting-for-dev!

waiting-for-dev:
> We are putting it on hold, as we'll rescue from Psych exception about unknown classes and print a message redirecting to the new preference.

Done. @albertoalmagro, @aldesantis, do you want to take a final look?

core/app/models/spree/log_entry.rb
```diff
@@ -2,10 +2,57 @@
...
+    # @see Spree::AppConfiguration#log_entry_permitted_classes
+    CORE_PERMITTED_CLASSES = [
+      ActiveMerchant::Billing::Response,
+      ActiveSupport::TimeWithZone,
+      Time,
+      ActiveSupport::TimeZone
+    ].freeze
+
+    # Raised when a disallowed class is tried to be loaded
+    class DisallowedClass < ArgumentError
```
aldesantis, line 19: This is a nit, but maybe a good old `RuntimeError` is semantically better here? It doesn't seem like the exception is caused by a bad method argument.

core/app/models/spree/log_entry.rb
```diff
@@ -2,10 +2,57 @@
...
+    end
+
     belongs_to :source, polymorphic: true, optional: true
 
     def parsed_details
-      @details ||= YAML.safe_load(details, permitted_classes: [ActiveMerchant::Billing::Response])
+      @details ||= YAML.safe_load(
+        details,
+        permitted_classes: self.class.permitted_classes,
+        aliases: Spree::Config.log_entry_allow_aliases
```
aldesantis, line 52: Should we also raise our own exception for aliases, so that the user knows about the new preference?

aldesantis (commented):
@waiting-for-dev the implementation looks good, but I've left a couple of suggestions. Let me know your thoughts!

core/app/models/spree/log_entry.rb
waiting-for-dev: Oh, yeah, it makes total sense. Changed!

core/app/models/spree/log_entry.rb
waiting-for-dev: I thought about it. Aliases are on by default, so users must know about the preference if they're off. For the sake of simplicity, I decided not to add it. However, it's also true that it won't hurt, and it can help another developer. WDYT?

aldesantis: @waiting-for-dev I think it's a nice little touch and doesn't require a lot of work, so I'd say let's just do it! 🙂

waiting-for-dev: Done!! 🙂

aldesantis: Look sgood!

PR #4333, opened 2022-04-13
Remove deprecated_address_id column from shipments

mamhoff:
We have not used this column since we removed all references to it in
2016.

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message

PR #4377, opened 2022-05-18
Revert "Remove deprecated_address_id column from shipments"

waiting-for-dev:
This reverts commit e81d213bee3d3ea89d1fd3734956d6a8588c543f.

As it's not backward compatible, we'll put it on hold until v4.0.

Ref #4333, cc @mamhoff


waiting-for-dev:
After it's merged, we'll create a PR identical to #4333 and label it for 4.0.

kennyadsl (approved):

PR #4274, opened 2022-02-19
Replace expired GPG key for mysql install in dev Dockerfile

nemeth:
**Description**

Docker build fails with:
```
W: GPG error: http://repo.mysql.com/apt/debian buster InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 467B942D3A79BD29
E: The repository 'http://repo.mysql.com/apt/debian buster InRelease' is not signed.
``` 
The key `8C718D3B5072E1F5` expired in 2022-02-16:
https://keyserver.ubuntu.com/pks/lookup?search=mysql+release&fingerprint=on&exact=on&op=index

**Checklist:**
- [ ] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [ ] I have added a detailed description into each commit message
- [ ] I have updated Guides and README accordingly to this change (if needed)
- [ ] I have added tests to cover this change (if needed)
- [ ] I have attached screenshots to this PR for visual changes (if needed)

PR #4381, opened 2022-05-20
[v3.1] Replace expired GPG key for mysql install in dev Dockerfile

waiting-for-dev:
Backports #4274


kennyadsl:
@waiting-for-dev don't we need to backport this in 3.0 as well?

waiting-for-dev:
> @waiting-for-dev don't we need to backport this in 3.0 as well?

There's no Docker development support there 🙂

PR #4385, opened 2022-05-26
Deprecate dangling option_values and duplicated routes

waiting-for-dev:
Up until now, we were allowing the creation of option_values without an associated option_type. It makes no sense from the domain logic perspective. Ref https://github.com/solidusio/solidus/pull/3309 & https://github.com/solidusio/solidus/pull/3517.

On top of that, we're deprecating the nested version for member routes, as there's no need to fetch the parent id for those.

### DIFF
--- a/api/app/controllers/spree/api/option_values_controller.rb
+++ b/api/app/controllers/spree/api/option_values_controller.rb
@@ -18,6 +18,13 @@ module Spree
       end
 
       def create
+        Spree::Deprecation.warn <<~MSG unless request.path.include?('option_types')
+          This route is deprecated, as it'll be no longer possible to create an
+          option_value without an associated option_type. Please, use instead:
+
+            POST api/option_types/{option_type_id}/option_values
+        MSG
+
         authorize! :create, Spree::OptionValue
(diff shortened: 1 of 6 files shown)


waiting-for-dev:
> I asked for some context here, [#3517 (comment)](https://github.com/solidusio/solidus/pull/3517#issuecomment-1138295958). It's quite old, so not sure we can get anything, but I thought it was worth trying.

It looks like the referenced PR didn't introduce the behavior. It only added a test to make it explicit and fixed an issue when `option_type` is `nil`. As per #3309, the behavior has been that way because it used to be the default for all `belongs_to` associations on Solidus.

waiting-for-dev:
By the way, for some reason, CircleCI result is not being reported, although it was successful. See https://app.circleci.com/pipelines/github/nebulab/solidus?branch=waiting-for-dev%2Fduplicated_option_value_endpoints&filter=all

aldesantis (approved):

kennyadsl (approved):
Ah got it, thanks @waiting-for-dev!

PR #4387, opened 2022-05-30
Deprecate creating new shipment with an item via API

kennyadsl:
**Description**

There's no known reason why we need to force creating a shipment with exactly one variant into it.

There's a dedicated endpoint for that:

```
PUT /api/shipments/{shipment_number}/add
```

This feature is not used anymore in our backend UI and, if someone is using this endpoint, they will receive a deprecation warning with the instructions on how to adjust their flow to be compatible with the next major version.

TODO:

- [x] deprecate fields in [the related API documentation endpoint](https://solidus.stoplight.io/docs/solidus/6e41c2d519a2b-create-shipment)

**Checklist:**
- ~I have attached screenshots to this PR for visual changes (if needed)~

### DIFF
--- a/api/app/controllers/spree/api/shipments_controller.rb
+++ b/api/app/controllers/spree/api/shipments_controller.rb
@@ -61,9 +61,10 @@ module Spree
           variant = Spree::Variant.unscoped.find(params[:variant_id])
           @order.contents.add(variant, quantity, { shipment: @shipment })
           @shipment.save!
+          @shipment.reload
         end
 
-        respond_with(@shipment.reload, default_template: :show)
+        respond_with(@shipment, default_template: :show)
       end
 
       def update
--- a/api/app/controllers/spree/api/shipments_controller.rb
+++ b/api/app/controllers/spree/api/shipments_controller.rb
@@ -41,6 +41,8 @@ module Spree
       def create
         authorize! :create, Shipment
         quantity = params[:quantity].to_i
+        variant = Spree::Variant.unscoped.find(params[:variant_id])
+
         @shipment = @order.shipments.create(stock_location_id: params.fetch(:stock_location_id))
         @order.contents.add(variant, quantity, { shipment: @shipment })
 
--- a/api/app/controllers/spree/api/shipments_controller.rb
+++ b/api/app/controllers/spree/api/shipments_controller.rb
@@ -40,13 +40,28 @@ module Spree
 
       def create
         authorize! :create, Shipment
-        quantity = params[:quantity].to_i
-        variant = Spree::Variant.unscoped.find(params[:variant_id])
 
         @shipment = @order.shipments.create(stock_location_id: params.fetch(:stock_location_id))
-        @order.contents.add(variant, quantity, { shipment: @shipment })
 
-        @shipment.save!
+        if passing_deprecated_params_on_create?
+          Spree::Deprecation.warn <<~MSG
+          Passing `quantity` or `variant_id` to
+
+              POST /api/shipments
+
+          is deprecated and won't be allowed anymore starting from Solidus 4.0.
+          Instead, create an empty shipment and add items to it subsequently using
+          the dedicated endpoint:
+
+              PUT /api/shipments/{shipment_number}/add
+
+          MSG
+
+          quantity = params[:quantity].to_i
+          variant = Spree::Variant.unscoped.find(params[:variant_id])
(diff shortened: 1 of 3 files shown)


api/app/controllers/spree/api/shipments_controller.rb
```diff
@@ -40,13 +40,31 @@ def select_shipping_method
...
+
+              PUT /api/shipments/{shipment_number}/add
+
+          MSG
+
+          quantity = params[:quantity].to_i
+          variant = Spree::Variant.unscoped.find(params[:variant_id])
+          @order.contents.add(variant, quantity, { shipment: @shipment })
+          @shipment.save!
+          @shipment.reload
```
waiting-for-dev, line 64: Do we need `reload` here?

kennyadsl: I don't know (see https://github.com/solidusio/solidus/pull/4387/commits/d018c6e4f0c0b699a8b65012457332bef96627db). But being code that we are going to remove soon, I think it's not even worth spending too much time to understand if there was a reason for that reload to be there.

api/openapi/solidus-api.oas.yml
```diff
@@ -5579,6 +5579,10 @@ paths:
         Creates a shipment.
 
         Please note that this request can be only performed by users with the `create` permission on the shipment.
+
+        **Deprecation Warning**: Adding items to the shipment via this endpoint
+        is deprecated. Instead, create an empty shipment and populate it with
+        the dedicated endpoint [to add items to the shipment](/docs/solidus/7078dbcf415ac-add-shipment-item).
```
kennyadsl, line 5585: I don't know if this way of adding link will work. I saw [the docs here](https://meta.stoplight.io/docs/platform/hosted-docs/stoplight-urls) but I don't know how to test it until it's published.

api/app/controllers/spree/api/shipments_controller.rb
waiting-for-dev: Yeah, you're right :+1:

api/openapi/solidus-api.oas.yml
waiting-for-dev: Oh, interesting. I was looking for something like it for #4385 & #4388. If it works, I'll update the references I added there.

waiting-for-dev (approved):
Thanks!!

spaghetticode (approved):
Looks good to me, thank you!

api/app/controllers/spree/api/shipments_controller.rb
```diff
@@ -41,6 +41,8 @@ def select_shipping_method
       def create
         authorize! :create, Shipment
         quantity = params[:quantity].to_i
+        variant = Spree::Variant.unscoped.find(params[:variant_id])
```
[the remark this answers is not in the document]
kennyadsl: Good point. It was added [here](https://github.com/solidusio/solidus/commit/8efd648a019ea5053caaf71cc2e23f25f3eb36d9), it seems to allow removing discarded items from a shipment. It doesn't seem necessary for adding them though, changing this, thanks for the question, Andrea!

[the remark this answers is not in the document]
kennyadsl: Wait, thinking twice, I'd prefer not to make that change. For the same reason described on [another commet](https://github.com/solidusio/solidus/pull/4387#discussion_r884874986), I think changing this deprecated code now could give us more pain than benefits. Are you ok keeping this as is?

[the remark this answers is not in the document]
spaghetticode: That makes sense! Thanks for double-checking this.

api/openapi/solidus-api.oas.yml
waiting-for-dev: What do you mean? The deprecation warning is only rendered on `master`, and I can switch the version with the top-left selector without problems.

I submitted https://github.com/solidusio/solidus/pull/4397 to update my references.

kennyadsl: I mean that if you click on the link from the `master` version, it will move you to the default one (currently 3.1) because the `/branches/master/` part of the URL is not reflected in the link.

waiting-for-dev: Oh, I see :facepalm:

waiting-for-dev: Yeah, I thought exactly that :wink:

PR #4388, opened 2022-05-30
Deprecate duplicated variant routes

waiting-for-dev:
As in https://github.com/solidusio/solidus/pull/4385:

- We keep shallow member routes. We deprecate nested member routes.
- We keep creation nested route. We deprecate creation shallow route.
- We keep both shallow and nested collection routes (more flexibility
for storefronts).

We're also tweaking the deprecation strategy on #4385 to make it less vulnerable
to false positives.


kennyadsl (approved):
Thanks Marc!

jarednorman (approved):

PR #4394, opened 2022-05-30
Make the inventory unit builder configurable

jarednorman:
**Description**

This adds the inventory unit builder (`Spree::Stock::InventoryUnitBuilder` by default) to the `Spree::Config.stock` configuration, so that it can be overridden.

This is desirable for any store that doesn't have a one-to-one mapping between purchased variants and the fulfilled inventory units. The extension `solidus_product_assembly` heavily overrides this class to accomplish this, so with this change it will be able to stop decorating `Spree::Stock::InventoryUnitBuilder` and instead provide its own.

I'll admit that I'm not a huge fan of how I tested the functionality in the order class, but there's a reason for stubbing the preference like that. `Spree::Config.stock` isn't a normal config object, so it doesn't support our test helpers for stubbing preferences within the context of only a single test.

**Checklist:**

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -486,7 +486,8 @@ module Spree
     end
 
     def create_shipments_for_line_item(line_item)
-      units = Spree::Stock::InventoryUnitBuilder.new(self).missing_units_for_line_item(line_item)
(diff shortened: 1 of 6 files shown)


jarednorman:
Oops, pushed this to the wrong remote. Re-opening from my own fork.

jarednorman:
Real PR: #4395

PR #4402, opened 2022-06-02
Only install ActiveStorage adapter on supported Rails versions

tvdeyen:
**Description**

We only support the ActiveStorage adapter for Rails >= 6.1,
but we install it for any supported Rails versions by default.

If generating a dummy test app for an extension this leads to
failing test runs, because Solidus quits oif the Rails version
is not Rails 6.1 if using the ActiveStorage adapter.

Solving by installing the ActiveStorage adapter only if the Rails
version is >= 6.1.

**Checklist:**

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails/generators'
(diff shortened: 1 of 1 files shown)


jarednorman (approved):

waiting-for-dev (approved):
Thanks!

PR #4409, opened 2022-06-08
Fix option_values nested attributes behavior on the API

waiting-for-dev:
**Description**

Both for option types and variants, `:option_values_attributes` was defined as a plain Symbol. As the [Rails documentation makes clear](https://api.rubyonrails.org/classes/ActionController/StrongParameters.html), nested attributes must be explicitly defined. Otherwise, they default to an empty Hash. On top of that, there's not a nested attributes definition for option values in variants.

We fix it for option types and remove it altogether for variants. Variants already support [`:option_value_ids` array](https://github.com/solidusio/solidus/blob/07c88ddd1603699939ac0343965fa88dc2e1851a/core/lib/spree/permitted_attributes.rb#L135), which covers the most common use case (adding option values to an already created variant), so there's no need to provide such a nested hierarchy.

**Checklist:**

### DIFF
--- a/api/openapi/solidus-api.oas.yml
+++ b/api/openapi/solidus-api.oas.yml
@@ -7185,10 +7185,6 @@ components:
           type: array
           items:
             type: integer
-        option_values_attributes:
-          type: array
-          items:
-            $ref: '#/components/schemas/option-value-input'
         options:
           type: object
           description: '`Name` will be the name/presentation of the option type and `Value` will be the name/presentation of the option value. They will be created if not exist.'
--- a/api/spec/requests/spree/api/option_types_spec.rb
+++ b/api/spec/requests/spree/api/option_types_spec.rb
@@ -89,6 +89,22 @@ module Spree::Api
         expect(response.status).to eq(201)
       end
 
+      it "can create an option type with nested option values" do
+        post spree.api_option_types_path, params: {
+          option_type: {
+            name: "Option Type",
+            presentation: "Option Type",
+            option_values_attributes: [
+              name: "foo", presentation: "Foo"
+            ]
+          }
+        }
+
+        option_value = Spree::OptionType.find(json_response["id"]).option_values.first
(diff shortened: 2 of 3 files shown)


waiting-for-dev:
> But, if someone was using option_values_attributes on the variant endpoints, would their requests break? 

No. The attribute will be ignored (not permitted). The difference is that both keys and values will now be ignored, while until now, the given value reached the model layer but was ignored once there. Notice that there's no definition of nested option values within variants:

```ruby
irb(main):005:0> Spree::Variant.new.option_values_attributes=[]
/home/solidus_user/gems/gems/activemodel-7.0.3/lib/active_model/attribute_methods.rb:458:in `method_missing': undefined method `option_values_attributes=' for #<Spree::Variant id: nil, sku: "", weight: 0.0, height: nil, width: nil, depth: nil, deleted_at: nil, is_master: false, product_id: nil, cost_price: nil, position: nil, cost_currency: nil, track_inventory: true, tax_category_id: nil, updated_at: nil, created_at: nil> (NoMethodError)     
Did you mean?  option_values_variants=                                                               
               option_values_variants
```

> Also, please take a look at the API docs [here](https://solidus.stoplight.io/docs/solidus/681aa6cb75b1e-create-product-variant). If I recall correctly that variant_attributes is also used to define the API available params.

I'm not sure I follow. I removed it from the docs right now. Is that what you meant?

kennyadsl:
> The difference is that both keys and values will now be ignored, while until now, the given value reached the model layer but was ignored once there.

Ah, now it's clear, thanks.

> I'm not sure I follow. I removed it from the docs right now. Is that what you meant?

Yes, exactly. Thanks Marc!

kennyadsl (approved):

luiz-m-affonso (approved):
💯

aldesantis (approved):

jarednorman (approved):

PR #4415, opened 2022-06-14
Fix delete response in admin users controller

tvdeyen:
**Description**

If a user with orders gets deleted in the admin an exception is raised. This exception is handled by the `user_destroy_with_orders_error method`.

This method uses the deprecated (and later removed) `text` response. Instead we need to use `plain`.

**Checklist:**


tvdeyen:
> Ah yeah, `:text` doesn't work on Rails 5.1+.

Yeah, and we didn't noticed...

tvdeyen:
> Hey @tvdeyen, could you rebase from master so that CI is green? 🙂

Done

jarednorman (approved):

waiting-for-dev (approved):
Thanks!

Issue #4368, opened 2022-05-09
Legacy backend user address update silently ignores validation errors


iLucker93:
gem 'solidus_backend', '~> 3.1'

https://drive.google.com/file/d/1fCWoJh98L1_d73qh5VAtCowEW7QMUSpf/view?usp=sharing

iLucker93:
any suggestion?

jarednorman:
Have you done any debugging to investigate the issue? What Rails version?

jarednorman:
Looks like a bug that needs to be looked into by someone. Would appreciate any more information on this you found in debugging it. What version of Rails are you running?

iLucker93:
any suggestion?

waiting-for-dev:
It looks like https://github.com/nebulab/solidus/blob/3b30854712cc7c50fb49a2ee461b848d938c5ca7/backend/app/controllers/spree/admin/users_controller.rb#L58 will return `true` no matter what. @iLucker93, do you want to work on it? I'll happily provide any help you need.

PR #4431, opened 2022-06-20
#4368 Admin panel - users - edit exist user - update address field always return 200

dborovsky:
**Issue:** https://github.com/solidusio/solidus/issues/4368

### DIFF
--- a/backend/app/controllers/spree/admin/addresses_controller.rb
+++ b/backend/app/controllers/spree/admin/addresses_controller.rb
@@ -38,16 +38,16 @@ module Spree
       end
 
       def message_validation_errors_for_addresses(bill_address, ship_address)
-        message = "Validation errors: "
+        message = "Validation errors: </br>"
         if(!bill_address.valid?)
-          message = message + "Billing address errors: #{map_errors(bill_address.errors.messages)}"
+          message = message + "Billing address errors: #{map_errors(bill_address.errors.messages)} <br/>"
         end
 
         if(!ship_address.valid?)
-          message = message + "Shipping address errors: #{map_errors(ship_address.errors.messages)}"
+          message = message + "Shipping address errors: #{map_errors(ship_address.errors.messages)}<br/>"
         end
 
-        message
+        message.html_safe
       end
 
       def map_errors(errors)
--- a/backend/app/controllers/spree/admin/addresses_controller.rb
+++ b/backend/app/controllers/spree/admin/addresses_controller.rb
@@ -3,9 +3,7 @@ module Spree
     class AddressesController < Spree::Admin::BaseController
       before_action :initial_user
 
-      def addresses
-        render :addresses
-      end
+      def addresses; end
 
       def update
         new_shipping_address = Spree::Address.immutable_merge(@user.ship_address, user_params[:ship_address_attributes])
@@ -16,11 +14,11 @@ module Spree
           @user.ship_address = new_shipping_address
 
           flash.now[:success] = t('spree.account_updated')
+          redirect_to :admin_user_addresses
         else
           flash.now[:error] = message_validation_errors_for_addresses(new_bill_address, new_shipping_address)
+          render "addresses"
         end
-
-        render :addresses
       end
 
       def model_class
--- a/backend/app/controllers/spree/admin/addresses_controller.rb
+++ b/backend/app/controllers/spree/admin/addresses_controller.rb
@@ -23,10 +23,6 @@ module Spree
         end
       end
 
-      def model_class
-        Spree.user_class
-      end
-
       private
 
       def initial_user
--- a/backend/app/controllers/spree/admin/addresses_controller.rb
+++ b/backend/app/controllers/spree/admin/addresses_controller.rb
@@ -3,7 +3,9 @@ module Spree
     class AddressesController < Spree::Admin::BaseController
       before_action :initial_user
 
-      def addresses; end
+      def show
+        render 'addresses'
+      end
 
       def update
         new_shipping_address = Spree::Address.immutable_merge(@user.ship_address, user_params[:ship_address_attributes])
--- a/backend/app/controllers/spree/admin/addresses_controller.rb
+++ b/backend/app/controllers/spree/admin/addresses_controller.rb
@@ -1,0 +1,38 @@
+module Spree
+  module Admin
+    class AddressesController < Spree::Admin::BaseController
+      def addresses
+        @user = Spree.user_class.find_by(id: params[:user_id]) if params[:user_id]
+
+        if request.put?
+          begin
+            new_shipping_address = Spree::Address.immutable_merge(@user.ship_address, user_params[:ship_address_attributes])
+            new_bill_address = Spree::Address.immutable_merge(@user.bill_address, user_params[:bill_address_attributes])
+
+            @user.bill_address = new_bill_address
+            @user.ship_address = new_shipping_address
+          rescue ActiveModel::StrictValidationFailed => e
+            flash.now[:error] = e.message
+            return render :addresses
+          end
+
+          flash.now[:success] = t('spree.account_updated')
+        end
+
+        render :addresses
+      end
+
+      def model_class
+        Spree.user_class
+      end
+
+      private
+
+      def user_params
+        attributes = permitted_user_attributes
+
+        params.require(:user).permit(attributes)
+      end
+    end
+  end
+end
--- a/backend/app/controllers/spree/admin/addresses_controller.rb
+++ b/backend/app/controllers/spree/admin/addresses_controller.rb
@@ -10,18 +10,18 @@ module Spree
       def update
         @user = Spree.user_class.find_by(id: params[:user_id]) if params[:user_id]
 
-        begin
-          new_shipping_address = Spree::Address.immutable_merge(@user.ship_address, user_params[:ship_address_attributes])
-          new_bill_address = Spree::Address.immutable_merge(@user.bill_address, user_params[:bill_address_attributes])
+        new_shipping_address = Spree::Address.immutable_merge(@user.ship_address, user_params[:ship_address_attributes])
+        new_bill_address = Spree::Address.immutable_merge(@user.bill_address, user_params[:bill_address_attributes])
 
+        if (new_shipping_address.valid? && new_bill_address.valid?)
           @user.bill_address = new_bill_address
           @user.ship_address = new_shipping_address
-        rescue ActiveModel::StrictValidationFailed => e
-          flash.now[:error] = e.message
-          return render :addresses
+
+          flash.now[:success] = t('spree.account_updated')
+        else
+          flash.now[:error] = message_validation_errors_for_addresses(new_bill_address, new_shipping_address)
         end
 
-        flash.now[:success] = t('spree.account_updated')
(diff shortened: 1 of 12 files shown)


backend/app/controllers/spree/admin/users_controller.rb
```diff
@@ -153,6 +155,19 @@ def set_stock_locations
             Spree::StockLocation.accessible_by(current_ability).where(id: user_params[:stock_location_ids])
         end
       end
+
+      def address_valid?
+        bill_address_params = user_params[:bill_address_attributes]
+        ship_address_params = user_params[:ship_address_attributes]
+
+        if(Address.new(bill_address_params).valid? &&
+          Address.new(ship_address_params).valid?)
```
jarednorman, line 164: I don't _think_ that this approach makes sense given that the attributes are immutably merged with the existing address. I'm pretty sure that's how that works, anyway.

jarednorman (changes requested):

backend/app/controllers/spree/admin/users_controller.rb
tvdeyen: With this approach we do not have error messages on the addresses that we could display.

I am puzzling with

```rb
@user.update(user_params) && @user.bill_address.valid? && @user.ship_address.valid?
```

but it does not seem to work either.

tvdeyen: We use `nested_attributes` for users ship_address and bill_address, but since address a immutable it silently (thanks rails) rejects. The address is always valid, because it never updates. We need a way to use `Spree::Address.immutable_merge` on an existing `user.x_address` instances somehow.

tvdeyen: I also believe we should add a `Admin::Users::AddressesController` to directly work on the address records in stead of proxying through the user.

tvdeyen:
@dborovsky Thanks for the contribution, but I think the issue is in how addresses work in Solidus (they are immutable). The solution is a bit more involved. Do you still want to try working on it? I left some comments about what I currently think could solve the issue.

dborovsky:
@tvdeyen Hi 

Thanks for your comment. I've already started working on it Have already covered with tests update addresses in user_controller. 

Yes, validation is not calling and for another attributes, for example if city is empty @user.update(user_params) return true too.

backend/app/controllers/spree/admin/users_controller.rb
dborovsky: > Admin::Users::AddressesController

Yes, I think better update address by callling new AddressesController.

backend/app/controllers/spree/admin/addresses_controller.rb
```diff
@@ -0,0 +1,65 @@
+module Spree
+  module Admin
+    class AddressesController < Spree::Admin::BaseController
+      before_action :initial_user
+
+      def show
```
waiting-for-dev, line 6: Why do we need a `#show` action?

backend/app/controllers/spree/admin/addresses_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+
+          flash.now[:success] = t('spree.account_updated')
+          redirect_to :admin_user_addresses
+        else
+          flash.now[:error] = message_validation_errors_for_addresses(new_bill_address, new_shipping_address)
+          render "addresses"
+        end
+      end
+
+      def model_class
```
waiting-for-dev, line 26: Sorry, why do we need it? I don't see it used anywhere else, right?

backend/app/controllers/spree/admin/addresses_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+        render 'addresses'
+      end
+
+      def update
+        new_shipping_address = Spree::Address.immutable_merge(@user.ship_address, user_params[:ship_address_attributes])
+        new_bill_address = Spree::Address.immutable_merge(@user.bill_address, user_params[:bill_address_attributes])
+
+        if (new_shipping_address.valid? && new_bill_address.valid?)
+          @user.bill_address = new_bill_address
+          @user.ship_address = new_shipping_address
```
waiting-for-dev, line 16: Did you check that it's actually updating the user and not only assigning it in memory?

backend/app/controllers/spree/admin/users_controller.rb
```diff
@@ -53,16 +53,6 @@ def update
         end
       end
 
-      def addresses
```
waiting-for-dev, line 56: I'm afraid we need to deprecate the endpoint before removing it straight away, as some users might have overridden it. You take a look at how we did something similar here: https://github.com/solidusio/solidus/pull/4388/files

backend/app/views/spree/admin/addresses/_addresses_form.html.erb
```diff
@@ -4,7 +4,7 @@
     <fieldset class="no-border-bottom">
       <legend align="center"><%= t('spree.billing_address') %></legend>
       <%= f.fields_for :bill_address, @user.bill_address || Spree::Address.build_default do |ba_form| %>
-        <%= render partial: 'spree/admin/shared/address_form', locals: { f: ba_form, type: "billing" } %>
+      <%= render partial: 'spree/admin/shared/address_form', locals: { f: ba_form, type: "billing" } %>
```
waiting-for-dev, line 7:
```suggestion
        <%= render partial: 'spree/admin/shared/address_form', locals: { f: ba_form, type: "billing" } %>
```

backend/app/views/spree/admin/addresses/_addresses_form.html.erb
```diff
@@ -13,7 +13,7 @@
     <fieldset class="no-border-bottom">
       <legend align="center"><%= t('spree.shipping_address') %></legend>
       <%= f.fields_for :ship_address, @user.ship_address || Spree::Address.build_default do |sa_form| %>
-        <%= render partial: 'spree/admin/shared/address_form', locals: { f: sa_form, type: "shipping" } %>
+      <%= render partial: 'spree/admin/shared/address_form', locals: { f: sa_form, type: "shipping" } %>
```
waiting-for-dev, line 16:
```suggestion
        <%= render partial: 'spree/admin/shared/address_form', locals: { f: sa_form, type: "shipping" } %>
```

backend/app/views/spree/admin/addresses/addresses.html.erb
```diff
@@ -15,14 +15,14 @@
...
-      <%= render partial: 'addresses_form', locals: { f: f } %>
+    <%= form_for [:admin, @user], as: :user, url: admin_user_addresses_update_url(@user), method: :put do |f| %>
+    <%= render partial: 'addresses_form', locals: { f: f } %>
 
-      <% if can?(:update, @user) %>
-        <div data-hook="admin_user_edit_form_button">
-          <%= render partial: 'spree/admin/shared/edit_resource_links', locals: { collection_url: admin_users_url } %>
-        </div>
-      <% end %>
+    <% if can?(:update, @user) %>
```
waiting-for-dev, line 21: We also need to restore the indentation here.

waiting-for-dev (changes requested):
Sorry for taking so long before looking into it, @dborovsky 🙏 It's looking good. I left some preliminary comments. Let me know what you think.

backend/app/controllers/spree/admin/users_controller.rb
kennyadsl: Yes, people may still have this path in their customized views. We should deprecate this instead (and not remove the route),

backend/app/controllers/spree/admin/addresses_controller.rb
dborovsky: I separated in two actions: one is the get method `show`(when user clicks to tab Adresses) and another method post `update`(when user clicks button **update**)

backend/app/controllers/spree/admin/addresses_controller.rb
dborovsky: yes, new values will be saved for  user.  I tested. I reloaded the page and addresses have already been displayed with new values.

backend/app/controllers/spree/admin/addresses_controller.rb
waiting-for-dev: Oh, I see. Understood.

backend/app/controllers/spree/admin/users_controller.rb
dborovsky: @waiting-for-dev @kennyadsl  I added deprecation warning only to put endpoint `addresses`.

New get endpoint path `addresses` is the same as the old one

backend/config/routes.rb
```diff
@@ -168,9 +168,12 @@
       member do
         get :orders
         get :items
-        get :addresses
```
kennyadsl, line 171: I still have a concern related to backward compatibility related to these routes changes: am I right that the route helper method `addresses_admin_user_path` is no more available (it's called `admin_user_addresses_path` now)? If yes, I think we should keep it somehow.

backend/config/routes.rb
```diff
@@ -168,9 +168,12 @@
...
         put :addresses
       end
+
+      get 'addresses', action: :show, controller: 'addresses', as: :addresses
+      put 'addresses/update', action: :update, controller: 'addresses', as: :addresses_update
```
kennyadsl, line 175: As mentioned [here](https://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing), it's probably fine to skip the `_update` part in the `as:` argument. The PUT verb is already telling routes which action to take. Can we try without it please?

backend/spec/controllers/spree/admin/addresses_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'pry'
```
kennyadsl, line 4: Can you please remove this `require`?

kennyadsl (changes requested):
Thanks @dborovsky! I left some other comments related to backward compatibility and routes shape, but overall this is a good job. I think we are close to merging, thanks again!

tvdeyen (commented):
This is a great step in the right direction. Thanks for addressing this (pun indented)

I would like this to be more RESTful and check for the existence of a `@user` for the edit and update actions. In all other cases it should handle single addresses.

BTW. this is a great step into finally having a Admin -> User -> Addressbook feature 👏🏻

backend/config/routes.rb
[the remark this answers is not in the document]
tvdeyen: Actually

```suggestion
      get :addresses, to: 'addresses#edit', as: :addresses
      put :addresses, to: 'addresses#update, as: :addresses
```

backend/app/controllers/spree/admin/addresses_controller.rb
dborovsky: @tvdeyen  sorry, I don't fully understand what you mean? 

> _Then update the one record (passed as id), to be inline with the rest of the resource controllers and be RESTful in those cases?_

fthobe:
@jarednorman old admin closed won't fix

PR #4432, opened 2022-06-21
Skip failing test because of unsupported feature on Rails < 6.1

waiting-for-dev:
`ActiveStorageAdapter` is only supposed to work on Rails versions
greater than 6.1.

CI was silently failing on Rails 5.2 because of [calling `abort` on our
code](https://github.com/solidusio/solidus/blob/aa77140e8c5b07ab597126ba76524a3931bbb449/core/app/models/concerns/spree/active_storage_adapter.rb#L13).

I'll create a follow-up PR using `raise` instead of `abort`.

### DIFF
--- a/core/spec/models/spree/image/preferences_spec.rb
+++ b/core/spec/models/spree/image/preferences_spec.rb
@@ -12,20 +12,22 @@ RSpec.describe Spree::Image, type: :model do
   }
   custom_style_default = :other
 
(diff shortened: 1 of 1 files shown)


kennyadsl (approved):

waiting-for-dev:
I will merge it even if there's only a single approval, as it's a fix for the CI and blocking all other open PRs.

PR #4434, opened 2022-06-21
Fail with `raise` when the storage adapter is not supported

waiting-for-dev:
When `abort` is raised on a CI test, it's skipped, but it's not counted
as a failure, giving an obscure error result on the corresponding job.

See f1f9ca1a01c46ceaaec9d2ba5d9d7c7982742aa8 for an example.


tvdeyen (approved):
Interesting

jarednorman (approved):

PR #4435, opened 2022-06-21
Update CircleCI configuration

waiting-for-dev:
Images prepended with `circleci/` [are deprecated](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034) in favor of `cimg/` ones.

Adding the `-browsers` suffix will pull node & selenium, but not the actual browsers. For that, we need to add the `browser-tools` orb and install the browser & driver in new steps.

The [`-ram` suffix](https://circleci.com/docs/2.0/circleci-images#service-image-variant) in mysql is [no longer supported](https://circleci.com/developer/images/image/cimg/mysql).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  browser-tools: circleci/browser-tools@1.3
+
 executors:
   base:
     working_directory: &workdir ~/solidus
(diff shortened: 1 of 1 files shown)


jarednorman (approved):

jarednorman:
I wonder if we should consider at some point moving to GitHub Actions.

kennyadsl (approved):

waiting-for-dev:
> I wonder if we should consider at some point moving to GitHub Actions.

We talked about it a few weeks ago. In short, we didn't see it as a priority at this point (the value it'd provide is not clear), but it might be something worth investigating in the future.

PR #4187, opened 2021-10-03
Normalize i18n keys using i18n-tasks gem issue: #3978

shubham9411:
**Description**

Post install message from i18n-tasks:
```
Post-install message from i18n-tasks:
# Install default configuration:
cp $(bundle exec i18n-tasks gem-path)/templates/config/i18n-tasks.yml config/
# Add an RSpec for missing and unused keys:
cp $(bundle exec i18n-tasks gem-path)/templates/rspec/i18n_spec.rb spec/
```

We have translations in `api/` and `core/` only so added the config and test for them only. 

`i18n-tasks` provides 4 type of assessments for translations, checking missing and unused keys, normalization and checking inconsistent interpolations of translations. For `core/` I have removed check on `unused` keys because it has 1285 unused i18n keys that are not used inside `core/` but used in `backend/` and `frontend/`.

|  folder/ |  missing |  unused | normalization | inconsistent interpolations |
|---|---|---|---|---|
| api  |  ✅  |  ✅  | ✅   | ✅   |
|  core | ✅  | ❌  |  ✅  |  ✅  |

If you have any feedback please let me know.

Ref #3978 

**Checklist:**
- [x] I have followed [Pull Request guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines)
- [x] I have added a detailed description into each commit message
- [ ] I have updated Guides and README accordingly to this change (if needed)
- [x] I have added tests to cover this change (if needed)
- [ ] I have attached screenshots to this PR for visual changes (if needed)

PR #4437, opened 2022-06-22
Sanitize i18n keys using i18n-tasks gem issue: #3978

### LINKED ISSUE #3978: Normalize i18n keys
`i18n-tasks` gem now comes with the following post-install message:

```
Post-install message from i18n-tasks:
# Install default configuration:
cp $(bundle exec i18n-tasks gem-path)/templates/config/i18n-tasks.yml config/
# Add an RSpec for missing and unused keys:
cp $(bundle exec i18n-tasks gem-path)/templates/rspec/i18n_spec.rb spec/
```

It's a good opportunity to follow these instructions and improve i18n in our platform. I tried and there are a couple of things to configure and fix, but in general, it would be great to have the suggested specs to run in the CI.

waiting-for-dev:
i18n-tasks gem helps in managing translations. For now we don't use the
tasks in our test pipeline. Added config and test for `api/` and
`core/`.

Taking over #4187 

Closes #3978

### DIFF
--- a/api/spec/i18n_spec.rb
+++ b/api/spec/i18n_spec.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'i18n/tasks'
+
+RSpec.describe I18n do
+  let(:i18n) { I18n::Tasks::BaseTask.new }
+  let(:missing_keys) { i18n.missing_keys }
+  let(:unused_keys) { i18n.unused_keys }
+  let(:inconsistent_interpolations) { i18n.inconsistent_interpolations }
+
+  it 'does not have missing keys' do
+    expect(missing_keys).to be_empty,
+                            "Missing #{missing_keys.leaves.count} i18n keys, run `i18n-tasks missing' to show them"
+  end
+
+  it 'does not have unused keys' do
+    expect(unused_keys).to be_empty,
+                           "#{unused_keys.leaves.count} unused i18n keys, run `i18n-tasks unused' to show them"
+  end
+
+  it 'files are normalized' do
+    non_normalized = i18n.non_normalized_paths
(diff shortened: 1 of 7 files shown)


tvdeyen (approved):
Nice!

api/spec/i18n_spec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'i18n/tasks'
+
+RSpec.describe I18n do
```
aldesantis, line 5: Just a nit, but do you think it would make sense to turn this into a set of shared examples?

Bonus: would it make sense to add this to [solidus_dev_support](https://github.com/solidusio/solidus_dev_support)?

aldesantis (approved):
Approved with a suggestion!

api/spec/i18n_spec.rb
waiting-for-dev: > Just a nit, but do you think it would make sense to turn this into a set of shared examples?

As it is code automatically generated by the `i18n-tasks` gem, I think it's better to leave things as they're now. For instance, they could add a new task to update that code, which could not work in our case if we changed it.

> Bonus: would it make sense to add this to [solidus_dev_support](https://github.com/solidusio/solidus_dev_support)?

Yeah, that's an excellent point for the extensions.

waiting-for-dev: Hey @aldesantis, are you ok with that? :slightly_smiling_face:

kennyadsl: @aldesantis ping, so we can merge this one!

waiting-for-dev: Merging, we can always extract it as a follow-up work :slightly_smiling_face:

PR #4439, opened 2022-06-23
Deprecate Rails 5.2

waiting-for-dev:
[Rails 5.2 reached EOL](https://guides.rubyonrails.org/maintenance_policy.html) on June 1st, 2022 [1]. We're going to deprecate it
on Solidus v3.2, and remove support on the next major.

### DIFF
--- a/core/lib/spree/core.rb
+++ b/core/lib/spree/core.rb
@@ -94,6 +94,15 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.7")
   HEREDOC
 end
 
+if Gem::Version.new(Rails.version) < Gem::Version.new('6.0')
+  Spree::Deprecation.warn <<~HEREDOC
+    Rails 5.2 (EOL) is deprecated and will not be supported anymore from the next Solidus version.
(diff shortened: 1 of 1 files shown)


core/lib/spree/core.rb
```diff
@@ -85,6 +85,15 @@ class GatewayError < RuntimeError; end
   end
 end
 
+if Gem::Version.new(Rails.version) < Gem::Version.new('5.2')
```
jarednorman, line 88: `<=`?

jarednorman (changes requested):

core/lib/spree/core.rb
waiting-for-dev: Oh, fixed it! :sweat: Thanks for spotting it! I've gone with `< 6.0` to catch patch-level versions.

jarednorman: Smart, I actually almost suggested that. 😅

jarednorman (approved):

kennyadsl (approved):

spaghetticode (approved):

Issue #2947, opened 2018-11-13
Incorrect behaviour for Cancel Item


gerritwessels:
### Steps to reproduce
Get the Solidus sandbox up and running.
Ensure that 'track_inventory' is set to true for variants. ( I believe this is the default setting anyway )
Create an order with a single line item and capture the payment.
At this point in time, the order statuses should be:
- Status: Complete
- Shipment: Ready
- Payment: Paid

**In the database:**
New Spree::StockMovement record will be present for this line_item, with quantity set to -1.
New Spree::Shipment record with state of 'Ready'
New Spree::InventoryUnit record with state 'on_hand' that belongs to the Spree::Shipment

**Trigger the Issue**
Go to the admin interface and cancel the item.

### Expected behavior
I would expect to see the item canceled, the payment balance to be updated and the spree_stock to be replenished with the canceled item.

If there was only one item in the Spree::Shipment, I would expect the shipment state to be changed to 'Canceled' or Spree::Shipment removed.

If there was multiple items in the same Spree::Shipment, I would expect the shipment state to remain as 'Ready'

### Actual behavior
**Database**
No Spree::StockMovement record gets created to replenish the stock, which seems incorrect.
The Spree::Shipment state transitions to 'shipped' and populates the shipped_at column, which seems incorrect.
The Spree::InventoryUnit state transitions to 'canceled', which is as expected.

**Question**
This behaviour seems wrong to me, am I missing something?

### System configuration
**Solidus Version**: v.2.7.0

**Extensions in use**: 
N/A

gerritwessels:
Regarding Spree::Shipment records that transition to 'shipped' when all the inventory_units are 'canceled'

I quickly poked around in the code to see if I could see anything obvious.
This line:
https://github.com/solidusio/solidus/blob/35010c23483086e1be832708bd06d30bb05b1a88/core/app/models/spree/order_cancellations.rb#L141

Quick fix for this could be:
```
# Before we update a shipment to 'shipped':
# Ensure that ALL the inventory_units for this shipment are in either a 'shipped' or 'canceled' state AND ensure that they are NOT ALL in a 'canceled' state
if (shipment.inventory_units.all? { |iu| iu.shipped? || iu.canceled? }) && (!shipment.inventory_units.all? { |iu| iu.canceled? })
```

jacobherrington:
Hey @gerritwessels if you have an idea to fix this, we'd love to get a PR for it. 👍

aitbw:
[...]

PR #4446, opened 2022-07-12
fix: update_shipped_shipments now cancels properly

NewAlexandria:
also refactor method parts for easier overwriting,
in a way that also encourages the use of super for the main method

**Description**
[We](https://github.com/suvie-eng) were having this same issue, and I found the related issue when searching.

This solution may a fait bit more ruby algo gymnastics, but if it's acceptable I think it could make it easier to overwrite this pattern, again while still trying to encourage people to call `super` when overwriting the `update_shipped_shipments` method.  

I also use an `update_all` call, to avoid N+1 queries.

I can include other comments, tests, etc, if this code looks like it'd be acceptable, and if anyone has suggestions.  Until I confirmed receptiveness I didn't want to setup the test env.   This PR is basically WIP, pending feedback.

Ref: https://github.com/solidusio/solidus/issues/2947

**Checklist:**

### DIFF
--- a/core/app/models/spree/order_cancellations.rb
+++ b/core/app/models/spree/order_cancellations.rb
@@ -112,15 +112,29 @@ class Spree::OrderCancellations
 
   # if any shipments are now fully shipped then mark them as such
   def update_shipped_shipments(inventory_units)
+    shipped_shippments_available_states.reduce({}) do |h,state_name|
+      h.merge({
+        state_name => updateable_shipped_shipments.select { |shipment| all_ius_match_state?(shipment, state_name) }
+      })
+    end.map {|state_name, shipments|} do
+      Spree::Shipment.update_all(shipments, state: state_name, shipped_at: Time.current)
+    end
+  end
+
+  def all_ius_match_state?(shipment, state_name)
+    shipment.inventory_units.all? do |iu|
+      iu.send "#{state_name}?".to_sym
+    end
+  end
+
+  def updateable_shipped_shipments
     shipments = Spree::Shipment.
       includes(:inventory_units).
       where(id: inventory_units.map(&:shipment_id)).
       to_a
+  end
 
-    shipments.each do |shipment|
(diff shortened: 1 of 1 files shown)


waiting-for-dev:
Hey @NewAlexandria, it looks like tests are failing. Would you like to keep working on it? :slightly_smiling_face:

NewAlexandria:
> Hey @NewAlexandria, it looks like tests are failing. Would you like to keep working on it? 🙂

Sure, I'm glad to.  As I mentioned in the PR body, I just wanted to make sure it seemed like a reasonable change that would be accepted given tests pass, etc. 

I'll be slow to respond because we're pushing through to a launch, but I can commit to finishing this PR's tests, so it's mergeable

kennyadsl:
Hey @NewAlexandria! Thanks for your contribution. Before evaluating the solution proposed (which looks good at a first pass), can you please add a failing spec for this bug to be sure this regression will never come back in the future? Thanks in advance!

NewAlexandria:
I will get back to this PR.  We're just in the middle of a release.  Thanks.

core/app/models/spree/order_cancellations.rb
```diff
@@ -112,15 +112,29 @@ def short_ship_unit(inventory_unit, created_by: nil)
...
       to_a
+  end
 
-    shipments.each do |shipment|
-      if shipment.inventory_units.all? { |iu| iu.shipped? || iu.canceled? }
-        shipment.update!(state: 'shipped', shipped_at: Time.current)
-      end
-    end
+  def shipped_shippments_available_states
+    %w(shipped canceled)
```
NewAlexandria, line 138: sketch of thought on tests, pending what's there already:
* these states
* more valid states
* no states
* invalid state/s (the `send` fail.  rescue it?)
* idempotency? (maybe exclude those already in `shipped_shippments_available_states` ?)

jarednorman:
Thanks for this work. Looks like it was superceded by #6314, though.

PR #4448, opened 2022-07-13
Allow Symbol as value in serialized preferences column for newly generated apps

tvdeyen:
Fixes an issue with latest Rails update for newly generated apps.

https://discuss.rubyonrails.org/t/cve-2022-32224-possible-rce-escalation-bug-with-serialized-columns-in-active-record/81017#releases-2

**Checklist:**
- [ ] I have added tests to cover this change

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -112,7 +112,7 @@ module Solidus
       end
       application <<~RUBY
         # Allow Symbol as value in serialized preferences column of Solidus models
-        config.active_record.yaml_column_permitted_classes = [Symbol]
+        config.active_record.yaml_column_permitted_classes = [Symbol, BigDecimal]
       RUBY
     end
 
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -110,6 +110,10 @@ module Solidus
     I18n.enforce_available_locales = #{options[:enforce_available_locales]}
         RUBY
       end
+      application <<~RUBY
+        # Allow Symbol as value in serialized preferences column of Solidus models
+        config.active_record.yaml_column_permitted_classes = [Symbol]
+      RUBY
     end
 
     def plugin_install_preparation
--- a/core/lib/generators/spree/dummy/dummy_generator.rb
+++ b/core/lib/generators/spree/dummy/dummy_generator.rb
@@ -78,7 +78,7 @@ module Spree
         inject_into_file("config/application.rb", after: /config.generators.system_tests = nil\n/, verbose: true) do
           <<-RUBY
             # Allow Symbol as value in serialized preferences column of Solidus models
-            config.active_record.yaml_column_permitted_classes = [Symbol]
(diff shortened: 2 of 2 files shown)


gsmendoza:
Hi @tvdeyen ! I was also looking into this and found that we may need to permit these classes in order to get the Solidus specs the pass:

* `ActiveSupport::HashWithIndifferentAccess`
* `BigDecimal`
* `Symbol`
* `Spree::Role`

However, I wasn't able to add `Spree::Role` to `config.active_record.yaml_column_permitted_classes` because I couldn't get `core/lib/spree/core/engine.rb` to load it.

Here is the spike and its build.

* Spike: https://github.com/solidusio/solidus/commit/b3975f0db34c9e35caba1dd9fd40c582516ab088 
* Build: https://app.circleci.com/pipelines/github/gsmendoza/solidus/19/workflows/c9417980-30a3-4c72-8c95-dfbe68c077d3/jobs/112. It only fails because of the specs that load `Spree::Role` objects.

gsmendoza:
@tvdeyen Based on discussions with @kennyadsl , @waiting-for-dev , and @aldesantis :

* We use serialize in:
  * [Spree::Preference](https://github.com/solidusio/solidus/blob/df51df62fa9b829216958b21b20514b7a3d87b30/core/app/models/spree/preference.rb#L4)
  * [Spree::ReturnItem](https://github.com/solidusio/solidus/blob/df51df62fa9b829216958b21b20514b7a3d87b30/core/app/models/spree/return_item.rb#L70)
  * [Spree::Preference::Persistable](https://github.com/nebulab/solidus/blob/df51df62fa9b829216958b21b20514b7a3d87b30/core/lib/spree/preferences/persistable.rb#L10) models
* There's another model using YAML serialized data, Spree::LogEntry, but [we're already using ](https://github.com/nebulab/solidus/blob/df51df62fa9b829216958b21b20514b7a3d87b30/core/app/models/spree/log_entry.rb#L73)safe_load[ there](https://github.com/nebulab/solidus/blob/df51df62fa9b829216958b21b20514b7a3d87b30/core/app/models/spree/log_entry.rb#L73).
* We need to make sure that we update the `active_record.yaml_column_permitted_classes` config only if the setting exists so that we don't break installations on unpatched Rails' versions.
* Unlike what I did in the [spike](https://github.com/gsmendoza/solidus/tree/gsmendoza/eng-395-update-solidus-to-support-cve-2022-32224), it would be preferrable to push to `active_record.yaml_column_permitted_classes` instead of assigning an array to it, so that we don't replace the configuration's existing contents.
* We may need to ensure that the fix works for all Rails versions that applied this security update: 7.0.3.1, 6.1.6.1, 6.0.5.1, and 5.2.8.1.

Finally, based on the [CircleCI orb](https://github.com/solidusio/circleci-orbs-extensions/blob/master/src/commands/run-tests-solidus-older.yml) we use for testing our extensions, we may need to apply the fix down to Solidus 2.11.

tvdeyen:
@gsmendoza thanks for the reply. I am more than happy to hand this over to you.

gsmendoza:
Thank you @tvdeyen ! I'll give it a try :)

tvdeyen:
Close in favor of #4451

PR #4449, opened 2022-07-13
Support code reloading when configuring static preferences sources

waiting-for-dev:
This is how [we usually recommend configuring sources](https://github.com/solidusio/solidus_stripe/blob/3fab36511a9d02ebd64d03571704ac5a9031b63b/README.md#usage) for static
preferences:

```ruby
Spree.config do |config|
  config.static_model_preferences.add(
    AmazingStore::AmazingPaymentMethod,
    'amazing_payment_method_credentials',
    credentials: ENV['AMAZING_PAYMENT_METHOD_CREDENTIALS'],
    server: Rails.env.production? ? 'production' : 'test',
    test_mode: !Rails.env.production?
  )
end
```

However, it's no longer possible to directly [reference an autoloadable](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoload-on-boot-and-on-each-reload)
class from an initializer.

In this case, we can't change the `#add` method to take the class name
instead. The reason is that it ends up [invoking a sanitization check](https://github.com/solidusio/solidus/blob/df51df62fa9b829216958b21b20514b7a3d87b30/core/lib/spree/preferences/static_model_preferences.rb#L13) on
the given preferences against the defined preferences, and that, of
course, needs access to the actual class.

Therefore, it's better to start recommending nesting the configuration
within a `.to_prepare` block, as they [run every time](https://guides.rubyonrails.org/configuring.html#initialization-events
) the code is
reloaded:

```ruby
Rails.application.config.to_prepare do
  Spree::Config.static_model_preferences.add(
    AmazingStore::AmazingPaymentMethod,
    'amazing_payment_method_credentials',
    credentials: ENV['AMAZING_PAYMENT_METHOD_CREDENTIALS'],
    server: Rails.env.production? ? 'production' : 'test',
    test_mode: !Rails.env.production?
  )
end
```

However, that means running the `#add` method multiple times, so we need
to change it to replace the definition instead of raising an error.

Closes #4070 & #4040

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@
 
 - Fix CSRF forgery protection bypass for Spree::OrdersController#populate [GHSA-h3fg-h5v3-vf8m](https://github.com/solidusio/solidus/security/advisories/GHSA-h3fg-h5v3-vf8m)
 
-**Other important changes**
+### Other important changes
+
+#### No more autoload of decorators in fresh applications
 
 New Solidus applications won't autoload files matching `app/**/*_decorator*.rb`
 pattern anymore. For previous Solidus applications, it's something that will
@@ -43,7 +45,7 @@ of Solidus recommendations (that files are monkey patches; they don't use the
 you can place those files in `app/overrides/` and remove the `decorator`
 suffix.
 
-### Changes to the promotion system
+#### Changes to the promotion system
 
 Promotions with a `match_policy` of `any` are deprecated. If you have promotions
 with such a match policy, try running the following rake task:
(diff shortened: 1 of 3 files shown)


aldesantis (approved):

kennyadsl (approved):
Approved assuming specs fails for a different reason. I think this is an important change that we should highlight in the Changelog.

waiting-for-dev:
> Approved assuming specs fails for a different reason. I think this is an important change that we should highlight in the Changelog.

Good point! I updated the Changelog accordingly.

PR #4451, opened 2022-07-14
Support CVE-2022-32224 Rails security updates

gsmendoza:
## Summary

Rails Versions 7.0.3.1, 6.1.6.1, 6.0.5.1, and 5.2.8.1 have been released to address CVE-2022-32224, documented at https://discuss.rubyonrails.org/t/cve-2022-32224-possible-rce-escalation-bug-with-serialized-columns-in-active-record/81017.

Currently, Solidus fails with the "Psych::DisallowedClass: Tried to load unspecified class: Symbol" error on those Rails versions. See https://app.circleci.com/pipelines/github/gsmendoza/solidus/14/workflows/88f440b9-3887-45c0-a013-593379d56ee5/jobs/85/steps.

## References

* https://rubyonrails.org/2022/7/12/Rails-Versions-7-0-3-1-6-1-6-1-6-0-5-1-and-5-2-8-1-have-been-released
* https://discuss.rubyonrails.org/t/cve-2022-32224-possible-rce-escalation-bug-with-serialized-columns-in-active-record/81017

## Previous solutions

* https://github.com/solidusio/solidus/pull/4448

## Post-merge tasks

Once approved and merged, we'll need to backport this PR to older Solidus versions: v3.1, v3.0, and v2.11.

**Checklist:**
- N/A: I have updated Guides and README accordingly to this change (if needed)
- N/A: I have added tests to cover this change (if needed)
- N/A: I have attached screenshots to this PR for visual changes (if needed)

### DIFF
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -15,6 +15,9 @@ module Spree
         generator.test_framework :rspec
       end
 
+      config.active_record.yaml_column_permitted_classes ||= []
+      config.active_record.yaml_column_permitted_classes |= [Symbol]
+
       initializer "spree.environment", before: :load_config_initializers do |app|
         app.config.spree = Spree::Config.environment
       end
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -16,7 +16,7 @@ module Spree
       end
 
       config.active_record.yaml_column_permitted_classes ||= []
-      config.active_record.yaml_column_permitted_classes |= [Symbol]
+      config.active_record.yaml_column_permitted_classes |= [Symbol, BigDecimal]
 
       initializer "spree.environment", before: :load_config_initializers do |app|
         app.config.spree = Spree::Config.environment
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -15,9 +15,11 @@ module Spree
         generator.test_framework :rspec
       end
 
-      config.active_record.yaml_column_permitted_classes ||= []
-      config.active_record.yaml_column_permitted_classes |=
-        [Symbol, BigDecimal, ActiveSupport::HashWithIndifferentAccess]
+      if ActiveRecord.respond_to?(:yaml_column_permitted_classes) || ActiveRecord::Base.respond_to?(:yaml_column_permitted_classes)
+        config.active_record.yaml_column_permitted_classes ||= []
+        config.active_record.yaml_column_permitted_classes |=
+          [Symbol, BigDecimal, ActiveSupport::HashWithIndifferentAccess]
+      end
 
       initializer "spree.environment", before: :load_config_initializers do |app|
         app.config.spree = Spree::Config.environment
--- a/core/lib/spree/core/engine.rb
+++ b/core/lib/spree/core/engine.rb
@@ -16,7 +16,8 @@ module Spree
       end
 
       config.active_record.yaml_column_permitted_classes ||= []
-      config.active_record.yaml_column_permitted_classes |= [Symbol, BigDecimal]
+      config.active_record.yaml_column_permitted_classes |=
+        [Symbol, BigDecimal, ActiveSupport::HashWithIndifferentAccess]
 
       initializer "spree.environment", before: :load_config_initializers do |app|
         app.config.spree = Spree::Config.environment
--- a/core/solidus_core.gemspec
+++ b/core/solidus_core.gemspec
@@ -42,6 +42,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'mini_magick', '~> 4.10'
   s.add_dependency 'monetize', '~> 1.8'
   s.add_dependency 'kt-paperclip', ['>= 6.3', '< 8']
+  s.add_dependency 'psych', ['>= 3.1.0', '< 5.0']
   s.add_dependency 'ransack', '~> 2.0'
   s.add_dependency 'sprockets-rails'
   s.add_dependency 'state_machines-activerecord', '~> 0.6'
--- a/core/spec/models/spree/promotion/rules/user_role_spec.rb
+++ b/core/spec/models/spree/promotion/rules/user_role_spec.rb
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe Spree::Promotion::Rules::UserRole, type: :model do
-  let(:rule) { described_class.new(preferred_role_ids: roles_for_rule) }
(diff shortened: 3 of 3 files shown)


core/lib/spree/core/engine.rb
```diff
@@ -15,6 +15,12 @@ class Engine < ::Rails::Engine
         generator.test_framework :rspec
       end
 
+      if ActiveRecord.respond_to?(:yaml_column_permitted_classes) || ActiveRecord::Base.respond_to?(:yaml_column_permitted_classes)
```
waiting-for-dev, line 18: Is there any scenario where the setting is defined in `ActiveRecord::Base` but not in `ActiveRecord`?

waiting-for-dev (commented):
Thanks, @gsmendoza! I left a doubt.

core/lib/spree/core/engine.rb
gsmendoza: @waiting-for-dev Yes, for Rails 6.1 and older, the setting is defined in `ActiveRecord::Base`. You can find the setting in the commit https://github.com/rails/rails/commit/8ce4bd1be83c08c30c34af4d0f1a726066128176. Actually, the build initially failed without the `ActiveRecord::Base` check. Please see https://app.circleci.com/pipelines/github/gsmendoza/solidus/21/workflows/5d0dafe5-418c-416a-8660-39f352df23c8.

gsmendoza:
Hello @tvdeyen ! Do you have any concerns about setting `active_record.yaml_column_permitted_classes` in `Spree::Core::Engine` instead of in `Solidus::InstallGenerator#configure_application`? It seems that setting the configuration in `Solidus::InstallGenerator#configure_application` doesn't allow the specs to pass.

waiting-for-dev (approved):

tvdeyen (approved):
Awesome. Great work.

core/lib/spree/core/engine.rb
```diff
@@ -15,6 +15,12 @@ class Engine < ::Rails::Engine
...
+        config.active_record.yaml_column_permitted_classes ||= []
+        config.active_record.yaml_column_permitted_classes
+          .concat([Symbol, BigDecimal, ActiveSupport::HashWithIndifferentAccess]).uniq!
```
waiting-for-dev, line 21:
```suggestion
        config.active_record.yaml_column_permitted_classes |= [Symbol, BigDecimal, ActiveSupport::HashWithIndifferentAccess]
```

waiting-for-dev, line 21:
```suggestion
        config.active_record.yaml_column_permitted_classes |= [Symbol, BigDecimal, ActiveSupport::HashWithIndifferentAccess]
```

Looking at https://github.com/rails/rails/commit/8ce4bd1be83c08c30c34af4d0f1a726066128176, it looks like the setting already defaults to an empty array. Plus, we can leverage the [union operator](https://ruby-doc.org/core-3.1.0/Array.html#method-i-7C) to simplify the code. What do you think, @gsmendoza?

gsmendoza: @waiting-for-dev I I tested your suggestion and found that `config.active_record.yaml_column_permitted_classes` is nil by default. This is implied by [this line](https://github.com/rails/rails/commit/8ce4bd1be83c08c30c34af4d0f1a726066128176#diff-184a4c3f4a4726fe0cafd0f3504ca57f53a3d87e5315c5e34b58b830e9b5ef21R294) in the commit.

In contrast, [this line](https://github.com/rails/rails/commit/8ce4bd1be83c08c30c34af4d0f1a726066128176#diff-4411c7993e3f8ccfd8e09f44a81f10dfac3ea7dbd44cabd60b369040a00be324R164) is the one that sets `ActiveRecord::Base.yaml_column_permitted_classes` to an empty array.

I think the "config.active_record.yaml_column_permitted_classes is [] by default" spec in the commit should be corrected since it actually tests `ActiveRecord::Base.yaml_column_permitted_classes` instead of `config.active_record.yaml_column_permitted_classes` :/

As for the union operator, I don't think we can switch to it yet since it's only been introduced in Ruby 2.6. We still haven't removed Ruby 2.5 from Solidus (see https://github.com/solidusio/solidus/pull/4442).

waiting-for-dev: > @waiting-for-dev I I tested your suggestion and found that config.active_record.yaml_column_permitted_classes is nil by default.

I see. Thanks for checking.

> As for the union operator, I don't think we can switch to it yet since it's only been introduced in Ruby 2.6. We still haven't removed Ruby 2.5 support from Solidus (see https://github.com/solidusio/solidus/pull/4442).

I think it's also [available on 2.5](https://devdocs.io/ruby~2.5/array#method-i-7C) :slightly_smiling_face:

gsmendoza: > I think it's also [available on 2.5](https://devdocs.io/ruby~2.5/array#method-i-7C) slightly_smiling_face

@waiting-for-dev You're right :) It seems the `|` operator has been around for awhile. It's the `union` method that's only been added in [2.6.1](https://ruby-doc.org/core-2.6.1/Array.html). I'll update the configuration. Todo.

core/solidus_core.gemspec
```diff
@@ -42,6 +42,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'mini_magick', '~> 4.10'
   s.add_dependency 'monetize', '~> 1.8'
   s.add_dependency 'kt-paperclip', ['>= 6.3', '< 8']
+  s.add_dependency 'psych', ['>= 3.1.0', '< 5.0']
```
gsmendoza, line 45: @waiting-for-dev @tvdeyen I added this line in https://github.com/solidusio/solidus/pull/4451/commits/10da2e7ff02da6f45cf11d635e6d600b09767215. From that commit: 

The permitted_classes keyword argument was introduced to psych starting from psych 3.1.0. See https://github.com/ruby/psych/commit/682abf20f0ddbc8d83e816022953536dad80c05b.

Ruby 2.5.9 installs 3.0.2 by default.

This fixes errors like this when running the specs:

```
  1416) Taxation system integration tests selling from new york to new york an order with a download and a shipment has a shipment with additional tax of 0.40
        Failure/Error: preferences[name] = value

        ArgumentError:
          unknown keywords: permitted_classes, aliases
        # ./lib/spree/preferences/preferable_class_methods.rb:86:in `block in preference'
        # ./lib/spree/testing_support/factories/shipping_method_factory.rb:29:in `block (3 levels) in <top (required)>'
        # ./spec/models/spree/tax/taxation_integration_spec.rb:597:in `block (3 levels) in <top (required)>'
```

waiting-for-dev (approved):

kennyadsl (approved):
Thanks @gsmendoza!

PR #4456, opened 2022-07-15
Fix stores with no authentication

waiting-for-dev:
We deprecated `#try_spree_current_user` in favor of
`#spree_current_user` in #3923. The former defaulted to `nil` when
no authentication library was present, but the latter was simply not
defined under those circumstances.

This commit restores the old behavior, as the code base works with the
assumption that current user is `nil` when no extension has defined it
(see [one example](https://github.com/solidusio/solidus/blob/fee8a860c068b6ad050f9b2ff9052aadf9d56174/backend/app/controllers/spree/admin/base_controller.rb#L36) or [another](https://github.com/solidusio/solidus/blob/fee8a860c068b6ad050f9b2ff9052aadf9d56174/backend/app/controllers/spree/admin/cancellations_controller.rb#L33)).

The issue was [raised in the support channel](https://solidusio.slack.com/archives/C0JBKDF35/p1657030519259159?thread_ts=1657019907.668419&cid=C0JBKDF35
).

### DIFF
--- a/core/lib/spree/core/controller_helpers/auth.rb
+++ b/core/lib/spree/core/controller_helpers/auth.rb
@@ -54,6 +54,11 @@ module Spree
           Spree::UserLastUrlStorer.new(self).store_location
         end
 
+        # Auth extensions are expected to define it, otherwise it's a no-op
+        def spree_current_user
+          defined?(super) ? super : nil
+        end
+
         # proxy method to *possible* spree_current_user method
         # Authentication extensions (such as spree_auth_devise) are meant to provide spree_current_user
         def try_spree_current_user
--- a/core/lib/spree/testing_support/dummy_app.rb
+++ b/core/lib/spree/testing_support/dummy_app.rb
@@ -18,9 +18,6 @@ RAILS_6_OR_ABOVE = Rails.gem_version >= Gem::Version.new('6.0')
 # @private
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
-
-  def spree_current_user
-  end
 end
 
 # @private
--- a/core/spec/lib/spree/core/controller_helpers/auth_spec.rb
+++ b/core/spec/lib/spree/core/controller_helpers/auth_spec.rb
@@ -114,4 +114,27 @@ RSpec.describe Spree::Core::ControllerHelpers::Auth, type: :controller do
       expect(response).to redirect_to('/unauthorized')
     end
   end
+
+  describe "#spree_current_user" do
+    context "when an ancestor defines it" do
+      it "delegates" do
+        controller = Class.new(ApplicationController) do
+          include (Module.new do
+            def spree_current_user
+              :user
+            end
+          end)
+          include Spree::Core::ControllerHelpers::Auth
+        end.new
+
+        expect(controller.spree_current_user).to eq :user
+      end
+    end
+
+    context "when no ancestor defines it" do
+      it "returns nil" do
+        expect(controller.spree_current_user).to eq nil
+      end
+    end
+  end
 end


core/lib/spree/core/controller_helpers/auth.rb
```diff
@@ -54,6 +54,11 @@ def store_location
           Spree::UserLastUrlStorer.new(self).store_location
         end
 
+        # Auth extensions are expected to define it, otherwise it's a no-op
+        def spree_current_user
+          defined?(super) ? super : nil
+        end
```
kennyadsl, line 60: Some lines later, in `try_spree_current_user`, we have:

```ruby
        def try_spree_current_user
          # This one will be defined by apps looking to hook into Spree
          # As per authentication_helpers.rb
          if respond_to?(:spree_current_user, true)
            spree_current_user
          # This one will be defined by Devise
          elsif respond_to?(:current_spree_user, true)
            current_spree_user
          end
        end```

With the proposed change, will the `respond_to?(:spree_current_user, true)` always be true?

**Review by kennyadsl — COMMENTED** [review]
Left a small comment, thanks!

**Comment by waiting-for-dev** [comment]
cc @elia

File: core/lib/spree/core/controller_helpers/auth.rb (hunk 1 above)
Comment (waiting-for-dev) [reply core/lib/spree/core/controller_helpers/auth.rb:60]: Yes, but I think that's ok. Then `spree_current_user` will delegate to the auth library, and `solidus_auth_devise` is already [implementing](https://github.com/solidusio/solidus_auth_devise/commit/69269765) it as noted in the linked PR.

My understanding is that `#current_spree_user` is never defined (`solidus_auth_devise` delegates to `#current_user` instead), so it's safe to forgo its support.

Comment (kennyadsl) [reply core/lib/spree/core/controller_helpers/auth.rb:60]: I'm not sure to fully understand here. If we are 100% sure that the `elsif` will never be executed, why keep it?

Comment (waiting-for-dev) [reply core/lib/spree/core/controller_helpers/auth.rb:60]: Yes, I guess we can remove it. Pinging @elia to be 100% sure.

**Review by jarednorman — APPROVED** [review]

File: core/lib/spree/core/controller_helpers/auth.rb (hunk 1 above)
Comment (elia) [reply core/lib/spree/core/controller_helpers/auth.rb:60]: @waiting-for-dev you're right, we shouldn't care about `current_spree_user` and this will work.
I'm just a bit hesitant on having a cover for a missing authentication implementation. Thoughts on having a sort of `NoAuth` module that we inject from within an initializer like we do for `Spree::AuthenticationHelpers` in auth-devise?

_solidus_auth_devise/lib/spree/auth/engine.rb_

```rb
module Spree
  module Auth
    class Engine < Rails::Engine
      …
      config.to_prepare do
        Spree::Auth::Engine.prepare_backend if SolidusSupport.backend_available?
        Spree::Auth::Engine.prepare_frontend if SolidusSupport.frontend_available?

        ApplicationController.include Spree::AuthenticationHelpers
      end
```

in the no-auth case would be something like:

_my-app/config/initializers/solidus_auth.rb_

```rb
Rails.application.reloader.to_prepare do
  ApplicationController.include Spree::NoAuth
end
```

aldesantis: @elia given our new storefront customization strategy is never to inject anything, but rather to copy-paste code into the host application. We don't want to do anything weird with the storefront as that makes customization harder.

waiting-for-dev: Thanks for looking into it, @elia! On top of what @aldesantis said, I think it's not worth it to make things complex. I agree that we don't have a clear separation of concerns for the authentication system. But extracting things to separated modules would give an *apparent* impression of orthogonality, but the truth is that all the components expect something to be there. Until we don't revisit it, I think it's better to have the no-op behavior as visible as possible.

aldesantis (approved):

PR #4460, opened 2022-07-22
Configurable promotion adjuster

mamhoff:
## Summary

This adds an extension point for a configurable promotion adjuster to the app configuration, and uses that extension point to point to a new  `Spree::Promotion::LegacyAdjuster`. This LegacyAdjuster does all the things that the order updater in conjunction with the Spree::Adjustment class was doing. 

## Checklist


core/app/models/spree/promotion/legacy_adjuster.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+module Spree
+  class Promotion < Spree::Base
+
+    # This class encapsulates all the things the promotion system does to
+    # an order.
```
kennyadsl, line 7: Not mandatory, but a more detailed (high-level) explanation could help readers in understanding what's going on here without reading the whole class code.

core/app/models/spree/promotion/legacy_adjuster.rb
```diff
@@ -0,0 +1,88 @@
...
+    class LegacyAdjuster
```
kennyadsl, line 8: If this is the only solution provided in core, I don't think we should call it "Legacy". Or am I missing something?

core/lib/spree/app_configuration.rb
```diff
@@ -342,6 +342,9 @@ def default_pricing_options
     # promotion_chooser_class allows extensions to provide their own PromotionChooser
     class_name_attribute :promotion_chooser_class, default: 'Spree::PromotionChooser'
 
+    # promotion_chooser_class allows extensions to provide their own PromotionChooser
```
kennyadsl, line 345: This comment needs to be changed to `promotion_adjuster_class`

kennyadsl (changes requested):
Love it! I left a couple of things I think should be changed, though. 

Also, can you please provide a real-world example of when someone could use a custom version of this? Not because I don't think there will be, but it will be super useful when writing a how-to for this preference based on a real need.

Thanks Martin! 🙏

core/app/models/spree/promotion/legacy_adjuster.rb
mamhoff: The plan is another adjuster that does not call `Spree::Adjustment#recalculate`, as that code path is very difficult to optimize and quite error-prone in some situations. This one does call that method, which I think should be deprecated once we have an alternative.

mamhoff: The adjustment should not be recalculated using its source; I'd much rather load all promotions and determine their eligibilty first, and then add, remove or change adjustments. The way it is done right now falls apart when using multi-level (orde,r line item, shipment, shipping rate) rules.

kennyadsl: I agree it's not the best way to handle the recalculation, but we risk that this legacy class ends up in a new version without the new option. My proposal is:

- rename it into something that works no matter if there's the new option available. Some semantic name that describes what it does would be awesome (and I know, hard to find as hell 😄 ). 
- merge it
- if we merge the new option before the next minor/major, we are still allowed to rename this class without deprecating it, because it has never been released
- if we don't merge the new option before the next minor/major, we can still deprecate the class, even if doesn't have the legacy suffix in its name. 

What do you think?

adammathys (approved):

core/spec/lib/spree/app_configuration_spec.rb
```diff
@@ -32,6 +32,10 @@
     expect(prefs.variant_price_selector_class).to eq Spree::Variant::PriceSelector
   end
 
+  it "uses legacy promotion adjuster class by default" do
```
kennyadsl, line 35: Last occurrence of legacy here

kennyadsl (approved):
Approved but if you have time to fix that occurrence of legacy it would be great!

core/spec/lib/spree/app_configuration_spec.rb
mamhoff: fixed

jarednorman (approved):

core/app/models/spree/order_updater.rb
```diff
@@ -163,6 +162,7 @@ def update_adjustment_total
       order.included_tax_total = all_items.sum(&:included_tax_total) + order_tax_adjustments.select(&:included?).sum(&:amount)
       order.additional_tax_total = all_items.sum(&:additional_tax_total) + order_tax_adjustments.reject(&:included?).sum(&:amount)
 
+      # TODO: Delete this line in Solidus 4.0, when it is run in `update_promotions`.
```
kennyadsl, line 165: @mamhoff while taking a look at this PR again, I notice this line. Is this comment still valid? I think we are already doing this in the `update_promotion` method so we can probably already remove this line?

kennyadsl: I diwscussed with @mamhoff about this, and it is actually valid. People might be calling `update_item_promotions` in their own store, which doesn't include this `promo_total` calculation.

mamhoff: https://github.com/solidusio/solidus/pull/4890

PR #4465, opened 2022-07-25
Fix expectations about solidus_auth_devise order in the Gemfile

waiting-for-dev:
## Summary

solidus_frontend tries to [load](https://github.com/solidusio/solidus_frontend/blob/dae732ed7442a755cc0426cd2afa6ac662de7105/app/views/spree/shared/_nav_bar.html.erb#L4) a [partial from solidus_auth_devise](https://github.com/solidusio/solidus_auth_devise/blob/bfbf616ef625aa51c6b74040d96998082f736e13/lib/views/frontend/spree/shared/_login_bar_items.html.erb). However, it provided a [fallback](https://github.com/solidusio/solidus_frontend/blob/dae732ed7442a755cc0426cd2afa6ac662de7105/app/views/spree/shared/_login_bar_items.html.erb) when the latter was missing.

The same is true in solidus_backend; it also tries to [load](https://github.com/solidusio/solidus/blob/a22723079f88387f53f72b503db875b9d97aa3f5/backend/app/views/spree/admin/shared/_navigation.html.erb#L10) a [partial from solidus_auth_devise](https://github.com/solidusio/solidus_auth_devise/blob/master/lib/views/backend/spree/admin/shared/_navigation_footer.html.erb), but provides its [fallback](https://github.com/solidusio/solidus/blob/master/backend/app/views/spree/admin/shared/_navigation_footer.html.erb)

Unfortunately, that worked with the assumption that solidus_auth_devise went below solidus_frontend and solidus_backend respectively in the Gemfile. Otherwise, the fallback templates took precedence.

For solidus_frontend, we're removing the placeholder template and rendering the one from solidus_auth_devise only when present. Similarly, for solidus_backend we're renaming the fallback template and rendering the one from solidus_auth_devise when present.

## Checklist

- ~[ ] I have updated the readme to account for my changes.~


gsmendoza (approved):

kennyadsl:
I don't know if I like **not** assuming that `solidus_backend` can live before `solidus_auth_devise` in the Gemfile, because it changes some behavior provided by the core gems. 

For the frontend, it is another story. If we remove all the frontend code from the extension, the order should no longer matter.

What about changing the `solidus_frontend` gem only?

kennyadsl:
> I'm not sure I follow. With these changes, solidus_backend can live both before and after solidus_auth_devise in the Gemfile

Yes, but my point is that there's no need to optimize `solidus_backend` to live after `solidus_auth_devise` because that should not be done. Mailny this should not be done because it's expected (and probably happens) that `solidus_auth_devise` changes things set by `solidus_backend` previously. Please let me know if I am missing anything.

waiting-for-dev:
> Mailny this should not be done because it's expected (and probably happens) that solidus_auth_devise changes things set by solidus_backend previously.

As far as I can see, `solidus_auth_devise` explicitly asks for the backend twice:

- [When injecting the routes](https://github.com/solidusio/solidus_auth_devise/blob/3addfb3521756f9eaff711f7eb0d6fea6eb4bc2d/config/routes.rb#L40).
- [When configuring the controller, within a `to_prepare` block](https://github.com/solidusio/solidus_auth_devise/blob/3addfb3521756f9eaff711f7eb0d6fea6eb4bc2d/lib/spree/auth/engine.rb#L25).

Both situations happen after all the gems have been required. It should be independent of their order.

As per the implicit work from solidus_support, we fixed the dependency on the Gemfile order in https://github.com/solidusio/solidus_support/commit/202e77c2d7887ee1edb7bf352b5a942e71b5176c.

If we're missing something, I think we should try fixing it as this PR does. IMHO, requiring the gems to be declared in any particular order is unacceptable. Users could find a tough time debugging a problem with a root outside of the code space, so to speak.

FWIW, the work's origin here is the changes that we will submit to support the installation of solidus_starter_frontend in the Solidus installer. The current installer will add solidus_auth_devise after Solidus. But we must run the new frontend installation script after installing Solidus. When users choose it, we need to remove the solidus meta-gem and replace it with solidus_core, solidus_backend, solidus_api & solidus_sample, so they don't pull solidus_frontend. As built-in bundler utilities that modify the Gemfile work by appending, the process switches the order and puts solidus_backend after solidus_auth_devise.

kennyadsl:
Got it, I think it's a matter of choosing if we want to encourage adding extensions wherever in the Gemfile, or only after the core gems. Let me explain what I mean: if a new developer ends up the installation process with a Gemfile that looks like this:

```

gem 'solidus_auth_devise`

gem 'solidus_core`
gem 'solidus_api`
...
```

where do we think they are going to add the other extensions later? Is it acceptable at the moment? Do the other extensions currently support living before the core gem? I think they don't (even if I'm not 100% sure). 


> FWIW, the work's origin here is the changes that we will submit to support the installation of solidus_starter_frontend in the Solidus installer. The current installer will add solidus_auth_devise after Solidus. But we must run the new frontend installation script after installing Solidus. When users choose it, we need to remove the solidus meta-gem and replace it with solidus_core, solidus_backend, solidus_api & solidus_sample, so they don't pull solidus_frontend. As built-in bundler utilities that modify the Gemfile work by appending, the process switches the order and puts solidus_backend after solidus_auth_devise.

Thanks for adding more context. As a curiosity, did you think about adding `solidus_auth_devise` after the `frontend/starter_frontend` choice in the installer?

waiting-for-dev:
> where do we think they are going to add the other extensions later?

I think that's not our call. Users should be able to add them wherever they want. The Gemfile order should not determine the resulting runtime.

If we look at the docs about [what is a Gemfile](https://bundler.io/man/gemfile.5.html) on Bundler's site, they don't mention that the gems are required in any particular order. Therefore, I don't think we should treat it as something deterministic. They already changed how they're [required in the past](https://github.com/rubygems/bundler/commit/f17183b53edc96162cd2e20255d6bc315779b5cf). True, now that happens in declaration order, as it's the most sensible default (as we can see here), but it's not part of the documentation. That `bundler add` can only append and doesn't provide `-before`, `-after` modifiers points in the same direction.

The Gemfile order should be more seen as the result of a style guide. For instance, Rubocop has the [`Bundler/OrderedGems` cop](https://docs.rubocop.org/rubocop/cops_bundler.html#bundlerorderedgems) which, in fact, would place `solidus_auth_devide` before `solidus_backend`.

> Thanks for adding more context. As a curiosity, did you think about adding solidus_auth_devise after the frontend/starter_frontend choice in the installer?

Yeah, that would be possible, but it'll result in a less linear installation flow and, sometimes, an extra `bundle install` call. What I tried to achieve is 1) Install solidus with its base components and its dependencies (solidus_auth_devise for the backend) & 2) Install the frontend.

kennyadsl:
You got me, this is the correct way to go. If other extensions won't work listed before the core gems, we should fix that somewhere else.

kennyadsl (approved):

waiting-for-dev:
> You got me, this is the correct way to go. If other extensions won't work listed before the core gems, we should fix that somewhere else.

Yeah, I think that's the way to go. Thanks for bearing with me :slightly_smiling_face:

aldesantis (approved):

aldesantis:
Looking good, thank you!

jarednorman (approved):

PR #4471, opened 2022-07-26
Allow to extend user deletion logic

tvdeyen:
## Summary

Currently we restrict to delete a user with an exception if
the user has any orders. Some businesses might want to
allow to delete users with incomplete orders.

This allows to override the can_be_deleted? method in the user
class.

Refs #3138

## Checklist

### DIFF
--- a/backend/app/views/spree/admin/users/index.html.erb
+++ b/backend/app/views/spree/admin/users/index.html.erb
@@ -89,7 +89,7 @@
           <% if can?(:edit, user) %>
             <%= link_to_edit user, no_text: true, url: spree.admin_user_path(user) %>
           <% end %>
-          <% if can?(:destroy, user) && user.orders.none? %>
+          <% if can?(:destroy, user) && user.can_be_deleted? %>
             <%= link_to_delete user, no_text: true, url: spree.admin_user_path(user) %>
           <% end %>
         </td>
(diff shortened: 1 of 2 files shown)


jarednorman (approved):

waiting-for-dev:
Thanks, @tvdeyen. Could we have tests for this new behavior? Also, I'm not a fan of having it as an AR callback, but we can revisit it later if we introduce more service objects.

tvdeyen:
> Thanks, @tvdeyen. Could we have tests for this new behavior? Also, I'm not a fan of having it as an AR callback, but we can revisit it later if we introduce more service objects.

There is no change in behavior. It behaves exactly like before. We have tests for this and they are green. 

What exactly do you like me to write tests for?

waiting-for-dev:
Hey, @tvdeyen. I mean, now it's possible to modify the default behavior by overriding the `#can_be_deleted?` method. We don't have tests for that. It'd be nice to reflect the expectation for users to modify it. WDYT?

waiting-for-dev:
In hindsight, I'm ok not adding them. We don't need to test every possible way our code can be overridden :slightly_smiling_face:

waiting-for-dev (approved):

jarednorman (approved):

PR #4475, opened 2022-07-26
Configurable cancelation adjuster

mamhoff:
## Summary

While working on the promotion system, I realized that unit cancels use `Spree::Adjustment#recalculate`. This is unfortunate, as that method does not work well for the improved promotions system, but is still used for cancellation adjustments. 

Knowing how taxes work, I was surprised to see that cancellation adjustments are recalculated after taxes are applied. This PR fixes that, and can thus simplify the calculation of the unit cancellation amount significantly: Before taxes, the amount of a cancellation must be simply the line item's price, since an inventory unit always has a quantity of 1. 

Tax adjustments will recalculate regardless of their `finalized` state. I believe this should be the case for cancellation adjustments as well. While it is unlikely that unit cancels change in value, taxes might very well change between order completion and unit cancellation. 

## Checklist

### DIFF
--- a/core/app/models/spree/unit_cancel.rb
+++ b/core/app/models/spree/unit_cancel.rb
@@ -34,6 +34,7 @@ class Spree::UnitCancel < Spree::Base
   # This method is used by Adjustment#update to recalculate the cost.
   def compute_amount(line_item)
     raise "Adjustable does not match line item" unless line_item == inventory_unit.line_item
-    -line_item.price
+    rest = [line_item.total_before_tax, adjustment&.amount&.abs].compact.sum
+    -([line_item.price, rest].min)
   end
 end
--- a/core/app/models/spree/unit_cancel.rb
+++ b/core/app/models/spree/unit_cancel.rb
@@ -34,20 +34,6 @@ class Spree::UnitCancel < Spree::Base
   # This method is used by Adjustment#update to recalculate the cost.
   def compute_amount(line_item)
     raise "Adjustable does not match line item" unless line_item == inventory_unit.line_item
-
-    -weighted_line_item_amount(line_item)
-  end
-
-  private
-
-  def weighted_line_item_amount(line_item)
-    quantity_of_line_item = quantity_of_line_item(line_item)
-    raise ZeroDivisionError, "Line Item does not have any inventory units available to cancel" if quantity_of_line_item.zero?
-
-    line_item.total_before_tax / quantity_of_line_item
-  end
-
-  def quantity_of_line_item(line_item)
-    BigDecimal(line_item.inventory_units.not_canceled.reject(&:original_return_item).size)
+    -line_item.price
   end
 end
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -426,6 +426,13 @@ module Spree
     #   Spree::OrderCancellations.
     class_name_attribute :order_cancellations_class, default: 'Spree::OrderCancellations'
 
+    # Allows providing your own class for adjusting items with cancellation adjustments
+    #
+    # @!attribute [rw] cancelled_items_adjuster
+    # @return [Class] a class with the same public interfaces as
+    #   Spree::OrderCancellations::CancelledItemsAdjuster.
+    class_name_attribute :cancelled_items_adjuster, default: 'Spree::OrderCancellations::CancelledItemsAdjuster'
+
     # Allows providing your own class for canceling payments.
     #
     # @!attribute [rw] payment_canceller
--- a/core/spec/models/spree/order_cancellations/cancelled_items_adjuster_spec.rb
+++ b/core/spec/models/spree/order_cancellations/cancelled_items_adjuster_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Spree::OrderCancellations::CancelledItemsAdjuster do
+  let(:order) { create(:order_ready_to_ship) }
+  subject { described_class.new(order).adjust! }
+
+  it "does not change anything on an order without canceled items" do
+    expect { subject }.not_to change { order.line_items.sum(&:total_before_tax) }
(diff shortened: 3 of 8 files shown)


core/spec/models/spree/unit_cancel_spec.rb
```diff
@@ -38,21 +38,8 @@
...
-    end
-
-    context "some inventory on the line item is canceled" do
-      before { inventory_unit2.cancel! }
-
-      it "divides the line item total by the uncanceled units size" do
         expect(subject).to eq(-10.0)
       end
-
-      it "raises an error if dividing by 0" do
```
forkata, line 52: Should we preserve the behaviour where this fails if all units are already cancelled? Either way, I think we should have a spec for that behaviour.

core/spec/models/spree/order_cancellations/cancelled_items_adjuster_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  context "if there is a canceled item" do
+    before do
+      Spree::OrderCancellations.new(order).short_ship([order.inventory_units.first])
+      # manually update the value of the cancellation adjustment
+      order.line_items.first.adjustments.detect(&:cancellation?).update(amount: -2)
+    end
+
+    it "changes the line item's total before tax" do
+      expect { subject }.to change { order.reload.line_items.sum(&:total_before_tax) }
```
forkata, line 21: Would love if we can assert what the amount is changed from/to as this could still pass if we make changes to this class, while doing the wrong thing. At least having the amounts there will surface any unintended changes to the behaviour.

core/app/models/spree/unit_cancel.rb
```diff
@@ -34,20 +34,6 @@ def adjust!
...
-  def weighted_line_item_amount(line_item)
-    quantity_of_line_item = quantity_of_line_item(line_item)
-    raise ZeroDivisionError, "Line Item does not have any inventory units available to cancel" if quantity_of_line_item.zero?
-
-    line_item.total_before_tax / quantity_of_line_item
-  end
-
-  def quantity_of_line_item(line_item)
-    BigDecimal(line_item.inventory_units.not_canceled.reject(&:original_return_item).size)
+    -line_item.price
```
forkata, line 37: Will this work for line items which have promotion adjustmets?

forkata:
Thanks @mamhoff! I left a couple of small comments, but love the simplified solution you have proposed here and the new configuration preference ❤️

core/spec/models/spree/unit_cancel_spec.rb
mamhoff: I think recalculating the amount should not fail, but adjusting the line item should. OTOH, `OrderCancellations` will already fail if the inventory unit is already canceled: https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order_cancellations.rb#L108 (in a transaction, will fail with an invalid state transition. Not sure we should add a spec to the recalculator.

core/spec/models/spree/order_cancellations/cancelled_items_adjuster_spec.rb
mamhoff: It's a bit problematic because I'm essentially changing the adjustment amount in the `before` block to something random, and then it corrects it back. I can add the amounts, but really only the `to` dimension is interesting.

core/app/models/spree/unit_cancel.rb
mamhoff: Hm hm. Yes and no. Promotion adjustments are `finalized`, and that means they won't get recalculated. The're also no method to find out the line item total before promotions, as cancelations are not part of the line item amount stuff (yet?).

IMO promotions should actually recalculate their amounts when cancelations happen.

core/spec/models/spree/order_cancellations/cancelled_items_adjuster_spec.rb
forkata: That makes sense, setting up realistic taxes which get recalculated correctly in specs is a pain 😩

core/app/models/spree/unit_cancel.rb
forkata: > IMO promotions should actually recalculate their amounts when cancelations happen.

Yeah I guess there are two separate approaches here, if promotion adjustments are not recalculated, then we need the cancellation adjustment to be for the discounted amount.
If we say they can be recalculated, then we need them to be calculated after the cancellation adjustment is applied on the full price of the item. That will lead to the promo adjustment amount being reduced to only discount non-cancelled quantities from the line.

core/lib/spree/app_configuration.rb
```diff
@@ -426,6 +426,13 @@ def default_pricing_options
     #   Spree::OrderCancellations.
     class_name_attribute :order_cancellations_class, default: 'Spree::OrderCancellations'
 
+    # Allows providing your own class for adjusting items with cancellation adjustments
+    #
+    # @!attribute [rw] cancelled_items_adjuster
+    # @return [Class] a class with the same public interfaces as
+    #   Spree::OrderCancellations::CancelledItemsAdjuster.
+    class_name_attribute :cancelled_items_adjuster, default: 'Spree::OrderCancellations::CancelledItemsAdjuster'
```
kennyadsl, line 434: It would be nice to have a versioned preference here and to change this behavior only for new stores. Unfortunately, this is still not possible for `class_name_attribute`.

kennyadsl (approved):
Thanks Martin!

mamhoff:
I'm hesitant to keep this open after discussions with @forkata on Slack. It seems some stores rely on the pre-existing solution, and Solidus as a whole (i.e. the order updater and `Spree::OrderContents` all have implicit dependencies on the existing order of things. Here be dragons.

waiting-for-dev:
Hey, @mamhoff, do you think that we should therefore close it?

mamhoff:
Yeah, let's close it for now; I'll probably get something up in the future that works better.

PR #4490, opened 2022-08-05
Deprecate solidus_frontend & allow installing solidus_starter_frontend

waiting-for-dev:
## Summary

We're recommending new stores use [solidus_starter_frontend](https://github.com/solidusio/solidus_starter_frontend) as their storefront. However, we still want to give the option to install solidus_frontend for now, as some extensions rely on it. We're modifying the Solidus installer to allow installing one or the other.

Nonetheless, [solidus_frontend is part of the solidus meta-gem](https://github.com/solidusio/solidus/blob/a22723079f88387f53f72b503db875b9d97aa3f5/solidus.gemspec#L27), which is what is usually bundled in a new installation. At this point, we can't remove solidus_frontend from the meta-gem, as users upgrading from old Solidus versions could see their storefronts suddenly gone. As that will be a breaking change, we're deferring it until v4.0.

Until that happens, we must prevent users choosing the new frontend from pulling solidus_frontend. Because of that, we're breaking down the solidus gem into solidus_core, solidus_api, solidus_backend & solidus_sample. We leverage [Bundler's injector](https://github.com/rubygems/bundler/blob/master/lib/bundler/injector.rb) to add dependencies matching the main solidus gem constraints in the Gemfile. We also use it to remove the meta-gem. Still, we need to extend the upstream injector to support the `path:` option, which is required for local testing (e.g., the sandbox application). We also must break down the meta-gem when solidus_frontend is installed until we remove the frontend directory from this repo. That prevents Bundler from resolving it from there when `path:` or `git:` are used as sources.

As per the logic to detect which frontend gets installed, the first of the following conditions met wins:

- It's explicitly selected with the `FRONTEND` environment variable
  (useful for sandbox applications).
- It's explicitly selected with the `frontend` option.
- If `solidus_frontend` is in the Gemfile, that's the one used.
- If the `auto_accept` option is given, `solidus_starter_frontend` is
  chosen.
- The user is prompted to select one or the other.

Extensions have been modified to explicitly depend on `solidus_frontend` on their Gemfile, so the dummy app generator used for testing will work OOO matching the third condition. As per its sandbox application, they'll default to `solidus_starter_frontend` because they use `auto_accept`, but they can select `solidus_frontend` through the `FRONTEND` variable.

## Checklist

- ~[ ] I have updated the readme to account for my changes.~


jarednorman (approved):

aldesantis (approved):
This is some top-notch work, thanks!

gsmendoza:
Hi @waiting-for-dev ! I'm encountering the following error when I try to install SolidusFrontend:

image

Here's the Rails app with SolidusFrontend: https://github.com/gsmendoza/rails_7_store/tree/solidus--solidus-449--solidus-frontend.

I didn't encounter any issues with the SolidusStarterFrontend version: https://github.com/gsmendoza/rails_7_store/tree/solidus--solidus-4490-solidus-starter-frontend. I was able to pass the specs and complete a checkout.

gsmendoza:
For comparison, here's the Rails app without any frontend: https://github.com/gsmendoza/rails_7_store/tree/solidus--solidus-4490--no-frontend.

waiting-for-dev:
> Hi @waiting-for-dev ! I'm encountering the following error when I try to install SolidusFrontend:

Thanks for checking it out, @gsmendoza!

That's weird. @kennyadsl also found something similar. The issue comes from the fact that solidus_frontend 3.2.0.alpha can come from two sources: the mono-repo & RubyGems (released from the forked repo). Only the latter contains the install generator. In my tests, that was the one selected by Bundler. However, fetching from `git:` instead of `path:` showed a warning about conflicting sources, but it picked RubyGems nonetheless. That is a short-lived issue, as there won't be conflicting sources right next when we remove the `frontend` folder from the mono-repo.

I've updated the code so that solidus_frontend is added to the Gemfile with an explicit source pointing to RubyGems. @gsmendoza, could you give it another try? :pray:

waiting-for-dev:
I also rebased from master after #4494

gsmendoza:
Thanks @waiting-for-dev! I am now able to install Solidus with SolidusFrontend. I used the following Rails app: https://github.com/gsmendoza/rails_7_store/tree/solidus--solidus-4490--solidus-frontend--explicit-source.

waiting-for-dev:
I also updated the message prompting to install solidus_auth_devise so that users are aware that it'll be installed nonetheless if they choose solidus_starter_frontend

kennyadsl (approved):
> I also updated the message prompting to install solidus_auth_devise so that users are aware that it'll be installed anyway if they choose solidus_starter_frontend

It works for now! I think it's acceptable for this transition period.

Thanks @waiting-for-dev, great work as always!

PR #4517, opened 2022-08-18
Update previous minor version

waiting-for-dev:
This is used by the update generator to figure out changes in
default preferences between releases:

https://github.com/solidusio/solidus/blob/ba316ea3a5580c80938449c77c37abf788e62f71/core/lib/generators/solidus/update/update_generator.rb#L9


waiting-for-dev:
Merging to go ahead with the release

PR #4563, opened 2022-09-05
Only default to activestorage adapter if Rails version is supported

tvdeyen:
## Summary

We only support the activestorage adapter with Rails 6.1 and above,
but we default to it, even if the Rails version is 6.0 and below.
Since we still support Rails 5.2 and 6.0, we cannot default to
the active storage adapter. The app won't boot unless we also
change the adapter in the spree initializer.

If the initializer, though, is nested in a `config.to_prepare` hook
(as recommended from Rails new autoloader zeitwerk [1]) the app
refuses to start anyhow, because the config is then taken from
the `app_configuration` defaults.

[1](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#use-case-1-during-boot-load-reloadable-code)

## Checklist

### DIFF
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -21,6 +21,7 @@ require "spree/core/search/base"
 require "spree/core/search/variant"
 require 'spree/preferences/configuration'
 require 'spree/core/environment'
+require 'spree/rails_compatibility'
 
 module Spree
   class AppConfiguration < Preferences::Configuration
@@ -497,7 +498,7 @@ module Spree
     # @!attribute [rw] image_attachment_module
     # @return [Module] a module that can be included into Spree::Image to allow attachments
     # Enumerable of images adhering to the present_image_class interface
-    class_name_attribute :image_attachment_module, default: 'Spree::Image::ActiveStorageAttachment'
+    class_name_attribute :image_attachment_module, default: Spree::RailsCompatibility.default_image_attachment_module
 
     # @!attribute [rw] allowed_image_mime_types
     #
@@ -555,7 +556,7 @@ module Spree
     # @!attribute [rw] taxon_attachment_module
     # @return [Module] a module that can be included into Spree::Taxon to allow attachments
     # Enumerable of taxons adhering to the present_taxon_class interface
-    class_name_attribute :taxon_attachment_module, default: 'Spree::Taxon::ActiveStorageAttachment'
+    class_name_attribute :taxon_attachment_module, default: Spree::RailsCompatibility.default_taxon_attachment_module
 
     # Configures the absolute path that contains the Solidus engine
     # migrations. This will be checked at app boot to confirm that all Solidus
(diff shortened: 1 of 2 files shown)


kennyadsl:
Hey @tvdeyen. Can you help me understand from where this is coming? New stores won't start on Rails 6.0 and if I really correctly there were a deprecation message telling people to explicitly set that preference to Paperclip for existing stores. 

Just to understand if there's something I am missing, but of course the changes make sense!

waiting-for-dev:
It makes sense, although I neither understand how it was a problem in new stores with the latest Rails. BTW, what about using https://github.com/solidusio/solidus/blob/master/core/lib/spree/rails_compatibility.rb to return the value? Having all the stuff that depends on Rails versions encapsulated will help to deprecate things while we move forward.

tvdeyen:
@kennyadsl sure.

This is a Rails 6.0 application, running Solidus 3.0.

The initializer is set up like this:

```rb
Rails.application.config.to_prepare do
  Spree.config do |config|
    [..]
    config.image_attachment_module = "Spree::Image::PaperclipAttachment"
    config.taxon_attachment_module = "Spree::Taxon::PaperclipAttachment"
  end
```

Trying to install the latest migrations via:

```
bin/rake railties:install:migrations
```

blows up with

```
Configuration Error: Solidus ActiveStorage attachment adpater requires Rails >= 6.1.0.

Spree::Config.image_attachment_module preference is set to Spree::Image::ActiveStorageAttachment
Spree::Config.taxon_attachment_module preference is set to Spree::Taxon::ActiveStorageAttachment
Rails version is 6.0.5.1

To solve the problem you can upgrade to a Rails version greater than or equal to 6.1.0
or use legacy Paperclip attachment adapter by editing `config/initialiers/spree/rb`:
config.image_attachment_module = 'Spree::Image::PaperclipAttachment'
config.taxon_attachment_module = 'Spree::Taxon::PaperclipAttachment'
```

> New stores won't start on Rails 6.0 

What do you mean by that? [We still support rails 5.2 and 6.0 in our Gemspec](https://github.com/solidusio/solidus/blob/master/core/solidus_core.gemspec#L30).

waiting-for-dev:
> @waiting-for-dev what do you mean by this comment? Do you agree with the change and I should try using that module instead?

That's it 🙂

tvdeyen:
> 

@waiting-for-dev thanks. Unfortunately the module is not available in Solidus 3.0 and this bug exists in Solidus 3.0 and above. Any ideas?

waiting-for-dev:
> > 
> 
> @waiting-for-dev thanks. Unfortunately the module is not available in Solidus 3.0 and this bug exists in Solidus 3.0 and above. Any ideas?

Good point. What about using it only where it's available?

kennyadsl:
I didn't get that the issue was there when the migration runs, no matter what preference are you using. That's definitely a bug that we should address soon. Thanks!

tvdeyen:
> > @waiting-for-dev thanks. Unfortunately the module is not available in Solidus 3.0 and this bug exists in Solidus 3.0 and above. Any ideas?
> 
> Good point. What about using it only where it's available?

@waiting-for-dev done for `master` and `v3.2`

waiting-for-dev (approved):
Thanks!

jarednorman (approved):

PR #4629, opened 2022-09-15
Frontend installers with app-templates

elia:
## Summary

This change makes it trivial to add/remove/change frontend or to pass a custom template to be used. This among other use-cases can become handy for testing out branches of solidus_starter_frontend without touching the code of the installer.

- It's closer to what you would expect coming from `rails new` options.
- `--frontend=none` was fixed and now properly removes `solidus_frontend`.
- The bundler splitter was compacted into a single file and always adds `solidus_frontend` leaving the choice to other templates if to keep it.
- A job testing the installer has been added to the CI

## Checklist


forkata (commented):
Nice work here @elia! This is probably not in scope for this PR, but I am curious if we have plans to enable similar behaviour for the dummy app generator as well. We've run into some challenges adding the `solidus_starter_frontend` to a dummy app (for extension testing) post generation due to the fact that there is no `Gemfile`.

elia:
@forkata that's a good point, can you please help give me the ropes on how to replicate that issue?

(on the same topic I'm also experimenting with using app-templates for more stuff in the installer and for turning some extensions into a sort of recipes, I'll send another PR for that if I end up with something viable)

bin/sandbox
```diff
@@ -3,6 +3,10 @@
 # Used in the sandbox rake task in Rakefile
 
 set -e
+if [ ! -z $DEBUG ]
```
waiting-for-dev, line 6: Not a big deal, but I think it won't catch when `$DEBUG` is set to the empty string:

```bash
> [ ! -z '' ]
> echo $?
1
```

I've usually seen to check for the presence of an env variable is `if [ -z ${var+x} ]`, which returns false if the variable is set to the empty string or any other string.

bin/sandbox
```diff
@@ -92,7 +96,7 @@ unbundled bin/rails generate solidus:install \
   --auto-accept \
   --user_class=Spree::User \
   --enforce_available_locales=true \
-  --with_authentication=false
+  --with_authentication=false \
```
waiting-for-dev, line 99: Thanks for fixing it!

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -1,40 +1,38 @@
...
-require_relative 'install_generator/bundler_context'
-require_relative 'install_generator/support_solidus_frontend_extraction'
-require_relative 'install_generator/install_frontend'
+require 'rails/generators'
+require 'rails/generators/app_base'
 
 module Solidus
   # @private
-  class InstallGenerator < Rails::Generators::Base
+  class InstallGenerator < Rails::Generators::AppBase
```
waiting-for-dev, line 9: I'm a bit concerned that `Rails::Generators::AppBase` is a bit private. It makes sense to use it in our context, but [Rails guides](https://guides.rubyonrails.org/generators.html) always reference `Rails::Generators::Base`.

The risk is mitigated thanks to the new CI job, though.

core/lib/generators/solidus/install/frontend_templates/solidus_frontend.rb
```diff
@@ -0,0 +1,3 @@
+apply "#{__dir__}/break_down_solidus_gem.rb"
```
waiting-for-dev, line 1: That changes the current behavior. We weren't breaking down the solidus meta-gem when solidus_frontend was chosen, as there's not actual need. Am I missing something?

core/lib/generators/solidus/install/frontend_templates/break_down_solidus_gem.rb
```diff
@@ -0,0 +1,59 @@
...
+  solidus_sample
+].map(&to_dependency))
+
+bundler_injector.remove(%w[
+  solidus
+])
+
+# `solidus_frontend` is not sourced from rubygems if the solidus gem was not.
+github_solidus_frontend = '--github solidusio/solidus_frontend' unless solidus.source.is_a? Bundler::Source::Rubygems
+bundle_command("add solidus_frontend --skip-install #{github_solidus_frontend}")
```
waiting-for-dev, line 58: Instead of adding and later removing, why not add only when solidus_frontend is chosen? 🤔

core/lib/generators/solidus/install/frontend_templates/break_down_solidus_gem.rb
```diff
@@ -0,0 +1,59 @@
...
+  solidus_backend
+  solidus_api
+  solidus_sample
+].map(&to_dependency))
+
+bundler_injector.remove(%w[
+  solidus
+])
+
+# `solidus_frontend` is not sourced from rubygems if the solidus gem was not.
```
waiting-for-dev, line 56: I'm not sure I get it. Could you explain it to me? 🙏

waiting-for-dev (changes requested):
Thanks for working on it, @elia! There are very nice improvements here.

However, I think something is wrong when installing solidus_starter_frontend, as when I go to `http://localhost:3000` I see Rails' welcome page instead. It's quite possible the error is on solidus_starter_frontend's side, but we need to investigate it.

I see the benefits of using rails templates for the frontend installers, but I'm afraid we lose proper encapsulation in this way. For instance, we're now duplicating how we check for gems being present in the Gemfile and, more importantly, we no longer respect `--auto-accept` when delegating to the solidus_frontend installer. What do you think about it?

elia:
Thanks for the review @waiting-for-dev 🙏 

I solved the the issue with the welcome page, it was applying the engine route to the Core routes block that is prepended by the SSF. I added running of the sandbox specs in the solidus-installer CI job, still a sort of smoke test but should be able to catch similar errors in the future.

Also now it forwards both `--auto-accept` (from `options[:auto_accept]`) and `--auto-run-migrations` (from `options[:migrate]`).
```
Use `bundle info [gemname]` to see where a bundled gem is installed.
    generate    solidus_frontend:install
       rails    generate solidus_frontend:install --auto-accept --auto-run-migrations
      create  config/initializers/solidus_frontend.rb
```

core/lib/generators/solidus/install/install_generator.rb
elia: AppBase is the one used by templates, another way would be to have a template file for the whole installation process and call it from a `Rails::Generators::Base` but I don't think it will gain us much. Also it's seems unlikely that AppBase is going away in any new Rails version. Open to rethink this though 👍

bin/sandbox
elia: I think it's ok to assume it will have some contents when set, that said maybe I'm doing some bash mistake here:

```
solidus:elia/frontend-templates ⤑ bash -l -c 'DEBUGz="asdf" test -z ${DEBUG+x}; echo $?'
0
solidus:elia/frontend-templates ⤑ bash -l -c 'DEBUG="asdf" test -z ${DEBUG+x}; echo $?'
0
solidus:elia/frontend-templates ⤑ bash -l -c 'DEBUG="" test -z ${DEBUG+x}; echo $?'
0
```

core/lib/generators/solidus/install/frontend_templates/break_down_solidus_gem.rb
elia: The idea is that if you have added solidus from rubygems you want to take the latest compatible `solidus_frontend` from rubygems as well.

On the contrary if you added it from git/github/path then it's better if we add it from GitHub as well.

You think it's fine to always get it from Rubygems?

core/lib/generators/solidus/install/frontend_templates/break_down_solidus_gem.rb
elia: This way we have a common baseline that is supposedly equivalent to the original but with exploded list of components, up to the scripts that don't require `solidus_frontend` to remove it.

core/lib/generators/solidus/install/frontend_templates/solidus_frontend.rb
elia: That's true, I think I did it for parity with the others but it's fine to keep the meta-gem intact in this case, I'll update the template 👍

bin/sandbox
waiting-for-dev: It looks like the parameter expansion is evaluated before the command, so $DEBUG is never defined at that point. But if you run it as part of a script, it'll work.

core/lib/generators/solidus/install/install_generator.rb
waiting-for-dev: Fair enough! The new CI job will tell us if something is broken.

core/lib/generators/solidus/install/frontend_templates/solidus_frontend.rb
waiting-for-dev: Thanks!

bin/sandbox
elia: Thanks for the explanation, let me know if you would still prefer to support an empty string there

core/lib/generators/solidus/install/frontend_templates/break_down_solidus_gem.rb
waiting-for-dev: Given that is the component we want to deprecate and that I wouldn't say undoing an action (or definition of action) is a good pattern for expansion, I would still have it only where it's needed. But not a blocker. I'll be fine with whatever you decide 🙂

core/lib/generators/solidus/install/frontend_templates/break_down_solidus_gem.rb
waiting-for-dev: Oh, yeah, it makes sense.

However, we should skip adding it when it's already in the Gemfile, right?

waiting-for-dev:
> I solved the the issue with the welcome page

Maybe you forgot to push it? It's not working for me yet.

.circleci/config.yml
```diff
@@ -122,22 +122,65 @@ commands:
...
           paths:
-            - solidus-version
+            - /tmp/my_app
+            - /tmp/bundle
+      - run:
+          name: "Run `solidus:install` with `<<parameters.flags>>`"
+          command: |
+            cd /tmp/my_app
+            bundle add solidus --git "file://$(ruby -e"puts File.expand_path ENV['CIRCLE_WORKING_DIRECTORY']")"
+            export SKIP_SOLIDUS_BOLT='true' # workaround for solidus_frontend not respecting --auto-accept when asking about bolt
```
waiting-for-dev, line 167: Now that we're respecting `--auto-accept`, we can:

- Remove the comment and keep skipping solidus_bolt.
- Remove the line and install solidus_bolt.

bin/sandbox
waiting-for-dev: I think it makes sense to support them. I know the code looks weirder, but it's not our fault that bash is bizarre 😄 Up to you! 🙂

elia: Done! TIL 🙏

core/lib/generators/solidus/install/frontend_templates/break_down_solidus_gem.rb
elia: Done, now the `starter_frontend` just adds the gem, and the break-down recipe doesn't do that anymore.

core/lib/generators/solidus/install/frontend_templates/break_down_solidus_gem.rb
elia: Adding `solidus_frontend` was removed from this app-template.

elia (commented):
@waiting-for-dev thanks for the reviews, I should have addressed everything 🙏

benjaminwil:
@elia:

> @forkata that's a good point, can you please help give me the ropes on how to replicate that issue?

We just created reproduction instructions and started a discussion for it: #4634

bin/sandbox
```diff
@@ -3,6 +3,7 @@
 # Used in the sandbox rake task in Rakefile
 
 set -e
+test -z "${DEBUG+empty_string}" && set -x
```
waiting-for-dev, line 6: Careful! Length is zero (`-z`) when `$DEBUG is not set, so it's actually doing the opposite that we want.

elia: 🤦‍♂️ thanks, fixed

waiting-for-dev (approved):
Thanks, @elia! You're great.

kennyadsl (approved):
Wow, thanks Elia!

PR #4636, opened 2022-09-22
Promote orders with any `promotionable?` line-item

ikraamg:
**Description**

Promote orders with any `promotionable?` line-item

The current behaviour seems incorrect, and also, there is a spec which is not clear about what the behaviour should be as the setup does not align with the test name.

**Checklist:**

### DIFF
--- a/core/spec/models/spree/promotion_spec.rb
+++ b/core/spec/models/spree/promotion_spec.rb
@@ -693,12 +693,12 @@ RSpec.describe Spree::Promotion, type: :model do
         let!(:line_item) { create(:line_item) }
         let!(:line_item2) { create(:line_item) }
 
-        context "and at least one item is non-promotionable" do
+        context "and at least one item is promotionable" do
           before do
             line_item.variant.product.promotionable = false
           end
 
-          it { is_expected.to be false }
+          it { is_expected.to be true }
         end
 
         context "and the items are all non-promotionable" do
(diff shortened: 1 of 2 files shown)


mamhoff:
I think this is not correct behavior. The idea is here is that line items that should not be promoted will never be promoted, not even through an order-level promotion. An order-level promotion will promote all line items in the order, and if an order contains ANY non-promotionable items, it should not be promoted.

core/spec/models/spree/promotion_spec.rb
```diff
@@ -709,10 +709,6 @@
 
           it { is_expected.to be false }
         end
-
-        context "and at least one item is promotionable" do
-          it { is_expected.to be true }
-        end
```
ikraamg, line 715: Either way, we need to clarify or remove this spec as is testing "and the items are all non-promotionable"

ikraamg (commented):
> I think this is not correct behavior. The idea is here is that line items that should not be promoted will never be promoted, not even through an order-level promotion. An order-level promotion will promote all line items in the order, and if an order contains ANY non-promotionable items, it should not be promoted.

@mamhoff I don't have a strong opinion here. Though, a change like this seems more intuitive as the boolean exists on the line items model and it would allow all promotions to be more flexible.

However, the current behaviour is important so perhaps a migration path that adds a new rule, "All line items are promotionable" to existing promotions and the `blacklisted?` (perhaps renamed) method only does:
`promotable.is_a?(Spree::LineItem) && !promotable.variant.product.promotionable?`

What do you think?

PR #4639, opened 2022-09-23
Fix variant price performance regressions 

mamhoff:
## Summary

This PR fixes a performance regression originating from #4076 and #4073. These two PRs fixed some inconsistencies by converting `variant.default_price` and  `variant.currently_valid_prices` from associations to methods. Unfortunately, they also introduced n+1 queries by attaching scopes to the `prices` association, essentially rendering any preloading of the `prices` association futile. 

What this PR does is address both scope usages. 

The first commit addresses the default price accessing `prices.with_discarded` so that a discarded variant keeps its default price. We can get around this scope usage by simply not discarding prices when a variant is discarded. Because we only really load prices through variants and not the other way around, this should have no real-world impact (and will make the unlikely case of someone actually un-deleting a variant more predictable).   

The second commit addresses the `currently_valid_prices` method. It sorts the `prices` association by priority in the database. By handling the sorting in ruby, we can get rid of the scope usage here. 

The third commit realizes that the `default_price` is the first price of the `currently_valid_prices` that fits the `default_price_attributes` and refactors some duplication out of the `Spree::DefaultPrice` concern. 

## Checklist


mamhoff:
Waiting on #4642 to fix the "coverage regression" :D

tvdeyen (approved):
As someone having several hundreds prices per product in the database (welcome to Europe!) I like this very much. 

For the "breaking" change: I cannot imagine a single scenario where I would want to keep prices for deleted variants anyway. And with just keeping them I am fine because as you said they will not appear anywhere anyway.

Thanks 👍🏻

core/app/models/concerns/spree/default_price.rb
```diff
@@ -16,9 +16,15 @@ def self.default_price_attributes
 
     # Returns `#prices` prioritized for being considered as default price
     #
-    # @return [ActiveRecord::Relation<Spree::Price>]
+    # @return [Array<Spree::Price>]
     def currently_valid_prices
-      prices.currently_valid
+      prices.sort_by do |price|
```
jarednorman, line 21: If we don't use the `currently_valid` scope here... do we use it anywhere? I believe my stores rely on the behaviour of `currently_valid`, though we can obviously move things around if needed.

waiting-for-dev, line 21: Any chance we can reorganize logic so that it's not duplicated with https://github.com/nebulab/solidus/blob/60f446c4343ad3929e249af0fb4ecf948eb3fbad/core/app/models/concerns/spree/default_price.rb#L68?

mamhoff: `prioritized_default` is deleted in the following commit ;)

mamhoff: We use it in the admin prices controller. We could probably get rid of it, but I would say that's outside of the scope of this PR :)

waiting-for-dev: Oh... 🙂

waiting-for-dev:
@mamhoff, you can now rebase from master to fix the codecov issue 🙂

waiting-for-dev (approved):
Thanks!!

jarednorman (approved):
I begrudgingly admit this is a good change, even if it's going to break one of my apps. 😅

waiting-for-dev:
> I begrudgingly admit this is a good change, even if it's going to break one of my apps. 😅

Hey @jarednorman, in which way will your store be broken?

jarednorman:
We have custom pricing logic and have overridden the `currently_valid` scope on prices, but this changes the `currently_valid_prices` method to no longer delegate to `currently_valid`.

mamhoff:
Maybe the whole thing belongs into the `variant_price_selector_class`.

jarednorman:
Yeah, there's still too many touch points for doing custom pricing right now.

mamhoff:
Rebased again. I've also provided backports for 3.1 and 3.2. 

Regarding the discussion with @jarednorman I totally think we should move the `currently_valid_prices` and `default_price` methods to the `variant_price_selector_class`, but I think that's something for a different PR that needs not to be backported.

kennyadsl:
@mamhoff what if we deprecate the current `currently_valid_prices` and `default_price` and start using the new methods moved into `variant_price_selector_class`? That would be a nice move in terms of backward compatibility, isn't it?

mamhoff:
@kennyadsl I'll provide a separate PR for `master` and `v3.2` once this is merged, alright?

kennyadsl:
I'm actually more concerned about shipping a patch-level version with the current change. If we consider this a plain bug, I think that adding the deprecation only in master (3.3 at this point) might just be useless, because people already had to deal with that code change. If this is not a bug, we should deprecate it on all versions. 

@mamhoff do you think this is a bug?

mamhoff:
I don't understand the question. It's a performance regression. It depends on Solidus' core's set of values whether that's a bug. 

Moving the two methods into another class in order to make it easier to customize pricing is something I would consider a feature. I'd move the implementation and call from the `DefaultPrice` module. After that we can think about deprecation; I don't really care. What I care about is removing this n+1 query from our endpoints.

mamhoff:
I paired with @tvdeyen on this again today, and we found that we could actually move a lot of the logic into the `Spree::Variant::PriceSelector`, keeping the changeset smaller. This also allows us to deprecate `Spree::Variant.currently_valid_prices` in this PR.

mamhoff:
We've tested this approach in our app and ran into errors! Turns out we have a bunch of discarded variants with discarded prices (as Solidus' current behavior is to discard all variants when a variant is discarded). These variants were now without a default price, as the code in this PR previously did not consider variants with exclusively discarded prices. 

I considered modifying the `Spree::Variant#prices` association in such a way that it includes discarded prices if the variant is soft-deleted, but excludes them when the variant is live. That proved to be difficult without making the association instance-dependent, which precludes preloading it (the whole point of this PR). So here's what I did not do: 

```ruby
# spree/variant.rb

# problematic as preloading is now impossible
has_many :prices, -> (variant) { variant.discarded? ? unscoped.with_discarded : kept }, ...
```

Let me know what you think. Oh soft-deletion.

tvdeyen:
@kennyadsl what do you think [about the latest changes](https://github.com/CandleScience/CSpree/commit/3ae149acfd56cf9a9c39a49a2fcdd0d29e3901ef)?

We recorded a video showing the bug where a deleted price is still used as a default price for a variant.

### before

[variant deleted price is default.webm](https://user-images.githubusercontent.com/42868/195567715-8971ecba-8ee9-4c21-918d-e563af3890e6.webm)

### after

[variant deleted price after.webm](https://user-images.githubusercontent.com/42868/195567993-d306e742-b0cf-4de6-b4bc-a647e6680ea8.webm)

This PR fixes that as well! 🎉

kennyadsl:
I'd wait for @waiting-for-dev re-review here. He worked on this part so I'm sure we will have a better review. 

> We recorded a video showing the bug where a deleted price is still used as a default price for a variant.

@tvdeyen This definitely looks like a bug, is this regression covered by some specs now?

tvdeyen:
> I'd wait for @waiting-for-dev re-review here. He worked on this part so I'm sure we will have a better review.
> 
> > We recorded a video showing the bug where a deleted price is still used as a default price for a variant.
> 
> @tvdeyen This definitely looks like a bug, is this regression covered by some specs now?

@mamhoff I am sure it is, right?

mamhoff:
@kennyadsl It's a bug but not a regression, in that it has been a bug for a long time. It's covered in this spec: 
https://github.com/solidusio/solidus/pull/4639/files#diff-f02d323d23d4cc20f7d7985235fa4ca043c81357b78a0fae2e41fe3bb4289480R307-R312

The more I look into this, the unhappier I am with the current way prices are handled (and I'm at least partly to blame here, as I wrote some of the confusing behavior. Especially the consideration of `updated_at` for the validity of a price is something we should look at, as well as the ability to soft-delete. IMO it shouldn't be possible at all :) I do think this PR represents an improvement over how things used to be. The weird behavior is now confined to deleted variants and deleted prices, where it does the least damage IMO.

waiting-for-dev (approved):
Thanks, it looks good to me!

kennyadsl:
I changed the label of this PR from Enhancement to Bug to reflect its content. @mamhoff going to merge, should we update the backport PRs as well?

mamhoff:
I can update them on Wednesday, if you want to Backport before that, go ahead! Thank you!

kennyadsl:
@mamhoff we could try using https://github.com/solidusio/solidus/pull/4678 for this. What do you think?

mamhoff:
Go ahead!

waiting-for-dev:
Done! Please, @mamhoff, take a look to be sure nothing is off.

PR #4642, opened 2022-09-26
Improve variant spec coverage

mamhoff:
## Summary

This adds some specs for a few instance methods on Spree::Variant. No functional changes.

## Checklist

### DIFF
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -211,7 +211,7 @@ module Spree
     # Assign given options hash to option values.
     #
     # @param options [Array] array of hashes with a name and value.
-    def options=(options = {})
+    def options=(options = [])
       options.each do |option|
         set_option_value(option[:name], option[:value])
       end
--- a/core/app/models/spree/variant.rb
+++ b/core/app/models/spree/variant.rb
@@ -210,7 +210,7 @@ module Spree
 
     # Assign given options hash to option values.
     #
-    # @param options [Array] array of hashes with a name and value.
+    # @param options [Array<Hash{name: String, value: String}>] array of hashes with a name and value.
     def options=(options = [])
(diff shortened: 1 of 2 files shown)


tvdeyen (approved):
Thanks

kennyadsl (approved):
Danke Martin!

core/spec/models/spree/variant_spec.rb
```diff
@@ -45,6 +45,28 @@
       variant_with_same_sku = build(:variant, sku: variant.sku)
       expect(variant_with_same_sku).to be_invalid
     end
+
+    context "should have a price if master variant" do
```
kennyadsl, line 49: Just one nit re-reviewing: this sounds a bit off as a context title, isn't it?

mamhoff: Fixed!

jarednorman (approved):

core/app/models/spree/variant.rb
```diff
@@ -211,7 +211,7 @@ def deleted?
     # Assign given options hash to option values.
     #
     # @param options [Array] array of hashes with a name and value.
```
waiting-for-dev, line 213: The doc is also wrong, as it should be `[Hash]`. Nice if we fix it now.

waiting-for-dev (approved):
Minor suggestion to go the extra mile 🙂

core/app/models/spree/variant.rb
mamhoff: I think the brackets only indicate that this is the type part of the line, not that it's an array. I can be specific and do `[Array<Hash>]`.

waiting-for-dev: Oh, yeah, sorry, I was confused. Thanks for your change.

PR #4646, opened 2022-09-27
Fix the dummy app usage of the generator

elia:
## Summary

Currently the installer generator relies on `Rails.root` which is not available when running it on the dummy app from extensions. 

- CI coverage was added
- It skips solidus_bolt
- Uses solidus_frontend by default, can be overridden setting `ENV['FRONTEND']`
- Sets the generator to `exit(false)` if an error occurs while running it

Example of a broken CI from one extension: https://app.circleci.com/pipelines/github/solidusio-contrib/solidus_paypal_commerce_platform/545/workflows/05ea9623-3caa-45bb-9724-99ad3ca91dfe/jobs/1417

## Checklist

### DIFF
--- a/core/lib/generators/solidus/install/frontend_templates/solidus_frontend.rb
+++ b/core/lib/generators/solidus/install/frontend_templates/solidus_frontend.rb
@@ -1,7 +1,12 @@
 solidus = Bundler.locked_gems.dependencies['solidus']
 
-# `solidus_frontend` is not sourced from rubygems if the solidus gem was not.
-github_solidus_frontend = '--github solidusio/solidus_frontend' unless solidus.nil? || solidus.source.is_a?(Bundler::Source::Rubygems)
+if Bundler.locked_gems.dependencies['solidus_frontend']
+  say_status :skipping, "solidus_frontend is already in the bundle", :blue
+else
+  # `solidus_frontend` is not sourced from rubygems if the solidus gem was not.
+  github_solidus_frontend = '--github solidusio/solidus_frontend' unless solidus.nil? || solidus.source.is_a?(Bundler::Source::Rubygems)
 
-bundle_command("add solidus_frontend #{github_solidus_frontend}")
+  bundle_command("add solidus_frontend #{github_solidus_frontend}")
+end
+
 generate 'solidus_frontend:install'
--- a/core/lib/spree/testing_support/common_rake.rb
+++ b/core/lib/spree/testing_support/common_rake.rb
@@ -9,12 +9,15 @@ class CommonRakeTasks
     namespace :common do
       task :test_app, :user_class do |_t, args|
         args.with_defaults(user_class: "Spree::LegacyUser")
-        require ENV['LIB_NAME']
+        lib_name = ENV['LIB_NAME'] or
+          raise "Please provide a library name via the LIB_NAME environment variable."
+
+        require lib_name
 
         force_rails_environment_to_test
 
         Spree::DummyGenerator.start [
-          "--lib-name=#{ENV['LIB_NAME']}",
+          "--lib-name=#{lib_name}",
           "--quiet",
         ]
 
@@ -28,7 +31,7 @@ class CommonRakeTasks
           'generate',
           'solidus:install',
           Dir.pwd, # use the current dir as Rails.root
-          "--lib-name=#{ENV['LIB_NAME']}",
+          "--lib-name=#{lib_name}",
           "--auto-accept",
           "--with-authentication=false",
           "--payment-method=none",
--- a/core/lib/spree/testing_support/common_rake.rb
+++ b/core/lib/spree/testing_support/common_rake.rb
@@ -34,7 +34,6 @@ class CommonRakeTasks
           "--lib-name=#{lib_name}",
           "--auto-accept",
           "--with-authentication=false",
-          "--payment-method=none",
           "--migrate=false",
           "--seed=false",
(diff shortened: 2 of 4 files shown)


spaghetticode (approved):
Thanks @elia !

forkata (approved):
Thanks @elia, this change looks great!

core/lib/generators/solidus/install/frontend_templates/solidus_frontend.rb
```diff
@@ -1,5 +1,10 @@
 solidus = Bundler.locked_gems.dependencies['solidus']
 
+if Bundler.locked_gems.dependencies['solidus_frontend']
+  say_status :skipping, "solidus_frontend is already in the bundle", :blue
+  return
```
waiting-for-dev, line 5: We need to skip adding solidus_frontend to the Gemfile, but we still need to run its generator.

core/lib/spree/testing_support/common_rake.rb
```diff
@@ -13,12 +9,38 @@ def initialize
...
 
-        Spree::DummyGenerator.start ["--lib_name=#{ENV['LIB_NAME']}", "--quiet"]
-        Solidus::InstallGenerator.start ["--lib_name=#{ENV['LIB_NAME']}", "--auto-accept", "--with-authentication=false", "--payment-method=none", "--migrate=false", "--seed=false", "--sample=false", "--quiet", "--user_class=#{args[:user_class]}"]
+        Spree::DummyGenerator.start [
+          "--lib-name=#{lib_name}",
+          "--quiet",
+        ]
+
+        # After the dummy app is generated the working directory
+        # is changed to that of the dummy app.
```
waiting-for-dev, line 25: I'm not sure I understand that. It looks like I'm still in the directory from where I run the command (which is desirable). 🤔

waiting-for-dev (changes requested):
Thanks, @elia! I tried in an actual extension, and it doesn't work because we're skipping running the solidus_frontend generator when it's already in the Gemfile. We only need to fix that 🙂

core/lib/spree/testing_support/common_rake.rb
```diff
@@ -13,12 +9,38 @@ def initialize
...
+          'FRONTEND' => ENV['FRONTEND'] || 'solidus_frontend',
+        }, [
+          'bin/rails',
+          'generate',
+          'solidus:install',
+          Dir.pwd, # use the current dir as Rails.root
+          "--lib-name=#{lib_name}",
+          "--auto-accept",
+          "--with-authentication=false",
+          "--payment-method=none",
```
waiting-for-dev, line 37: If you want to do it as part of this PR, this option is no longer used in the install generator.

core/lib/spree/testing_support/common_rake.rb
elia: Yeah I was bitten by this in the past and so I though it was good to (at least) warn my future self, here's a demonstration:

```
solidus_paypal_commerce_platform:master * ⤑ bin/rake clobber extension:test_app --trace
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
** Invoke clobber (first_time)
** Invoke clean (first_time)
** Execute clean
** Execute clobber
rm -r /Users/elia/Code/Nebulab/solidus_paypal_commerce_platform/spec/dummy
** Invoke extension:test_app (first_time)
** Execute extension:test_app
** Invoke common:test_app (first_time)
** Execute common:test_app
Generating dummy Rails application...
      create  config/master.key

From: /Users/elia/Code/Nebulab/solidus/core/lib/spree/testing_support/common_rake.rb @ line 23 :

    18: 
    19:         Spree::DummyGenerator.start [
    20:           "--lib-name=#{lib_name}",
    21:           "--quiet",
    22:         ]
 => 23:         binding.irb
    24: 
    25:         # After the dummy app is generated the working directory
    26:         # is changed to that of the dummy app.
    27:         sh({
    28:           'SKIP_SOLIDUS_BOLT' => '1',

>> Dir.pwd
=> "/Users/elia/Code/Nebulab/solidus_paypal_commerce_platform/spec/dummy"
>> 
```

core/lib/generators/solidus/install/frontend_templates/solidus_frontend.rb
elia: I did that thinking that if you're handling it yourself you would also run the generator. I can see both sides of this what would be your preference?

elia (commented):
Thanks for the review @waiting-for-dev, great call on the `--payment-method` option 👍

core/lib/generators/solidus/install/frontend_templates/solidus_frontend.rb
waiting-for-dev: I see your point. However, at this point, the dummy application errors because of that. So we can either run the generator as part of the task or as part of the Solidus installer. I think I'd go with the latter in the spirit of having it create a ready-to-go app.

core/lib/spree/testing_support/common_rake.rb
waiting-for-dev: Oh, I see. Would it be better to say "While the dummy app is generated" instead of "After the dummy app is generated"? I mean, when I run the task from the CLI, my directory is not changed after the task has run.

waiting-for-dev (approved):
Thanks for addressing those issues, @elia!

PR #4654, opened 2022-09-30
Use app-templates to handle authentication options in the installer

elia:
## Summary

This is a followup to the PR introducing the use of app-templates for installing different frontends, coming from the realization that going forward the so-called extensions can take up many forms (like solidus_starter_frontend) not necessarily being gems. This gives much more flexibility and the opportunity to offer more features and quality out of the box with reduced maintenance burden.

- Fix the custom_user generator
- Add `--authentication` options for `solidus_auth_devise` `none` `existing_devise` (the latter is actually what the `custom_user` generator is about, but feel free to suggest a different name)
- Deprecate `--with-authentication` (boolean) in favor of `--authentication` (string)
- Opens the way to applying the same pattern to:
  - storage (active-storage vs. paperclip)
  - payment (stripe/bolt/paypal/…)
  - seed (only seed vs seed + sample)
  - …

## Checklist


core/lib/generators/solidus/install/app_templates/authentication/none.rb
```diff
@@ -0,0 +1,23 @@
+initializer 'solidus_authentication.rb', <<~RUBY
+  module SolidusAuthenticationControllerHelper
+    def spree_current_user
+      Spree::LegacyUser.first ||
+        Spree::LegacyUser.create!(spree_roles: [
```
waiting-for-dev, line 5: Why do we want to create a user if there's none (at least without checking flags like `--sample`)?

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -14,9 +14,17 @@ class InstallGenerator < Rails::Generators::AppBase
...
     FRONTENDS = {
-      'none' => "#{__dir__}/frontend_templates/none.rb",
-      'solidus_frontend' => "#{__dir__}/frontend_templates/solidus_frontend.rb",
-      'solidus_starter_frontend' => "#{__dir__}/frontend_templates/solidus_starter_frontend.rb",
+      'none' => "#{__dir__}/app_templates/frontend/none.rb",
+      'solidus_frontend' => "#{__dir__}/app_templates/frontend/solidus_frontend.rb",
+      'solidus_starter_frontend' => "#{__dir__}/app_templates/frontend/solidus_starter_frontend.rb",
+    }
+
+    LEGACY_AUTHENTICATION = 'solidus_auth_devise'
```
waiting-for-dev, line 22: Is that correct?

core/lib/generators/spree/custom_user/custom_user_generator.rb
```diff
@@ -22,10 +22,12 @@ def generate
       migration_template 'migration.rb.tt', "db/migrate/add_spree_fields_to_custom_user_table.rb"
       template 'authentication_helpers.rb.tt', "lib/spree/authentication_helpers.rb"
 
-      file_action = File.exist?('config/initializers/spree.rb') ? :append_file : :create_file
-      send(file_action, 'config/initializers/spree.rb') do
-        "Rails.application.config.to_prepare do\n  require_dependency 'spree/authentication_helpers'\nend\n"
-      end
+      initializer 'solidus_authentication', <<~RUBY
+        Rails.application.config.to_prepare do
+          ApplicationController.include Spree::AuthenticationHelpers, Spree::CurrentUserHelpers
```
waiting-for-dev, line 27: I might be missing something, but why do we need to include those helpers in the user's main controller?

core/lib/generators/solidus/install/install_generator.rb
elia: Good catch, I initially intended to have a pure `devise` option that would install and configure devise directly in the rails app, but ended up skipping it. This is a leftover, removing 👍

core/lib/generators/solidus/install/app_templates/authentication/none.rb
elia: The intent is to ensure the resulting store is functional and can be explored, with this setup I was able to complete a checkout. 

But if you think it's too much I'm ready to remove it as I was also not sure it was worth it.

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -30,9 +38,12 @@ class InstallGenerator < Rails::Generators::AppBase
     class_option :admin_email, type: :string
     class_option :admin_password, type: :string
     class_option :lib_name, type: :string, default: 'spree'
-    class_option :with_authentication, type: :boolean, default: true
     class_option :enforce_available_locales, type: :boolean, default: nil
     class_option :frontend, type: :string, enum: FRONTENDS.keys, default: nil, desc: "Indicates which frontend to install."
+    class_option :authentication, type: :string, enum: AUTHENTICATIONS.keys, default: nil, desc: "Indicates which authentication system to install."
+
+    # DEPRECATED
+    class_option :with_authentication, type: :boolean, hide: true, default: nil
```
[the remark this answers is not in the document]
elia: This is what I thought initially as well, but the ecosystem is also using those options and a blunt removal can end up in bugs creeping in (we saw that with the `--payment-method` option).

In any case it should be easy to clean up in the next major.

core/lib/generators/spree/custom_user/custom_user_generator.rb
elia: Short (incomplete) answer: we're just keeping the previous behavior

Longer version: We could probably include them in `Spree::BaseController` but the currently released `solidus_auth_devise` has the default `ApplicationController` set to be the `Devise.parent_controller` used for all devise controller. So I'm pretty sure including in BaseController would be a problem, although I didn't try.

elia (commented):
@waiting-for-dev thanks for the review, I left a couple of comments, please let me know your thoughts on them 🙏

core/lib/generators/solidus/install/app_templates/authentication/none.rb
waiting-for-dev: I understand that `none` means: "We are not going to install an authentication system for you now, but you'll need to do it yourself". If I understand correctly, having a functional Solidus without authentication is impossible. If I'm correct, we could:
- Warn when `none` is selected, or be more explicit in the prompt message.
- Raise a meaningful error here in `spree_curent_user`.

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -256,9 +276,22 @@ def detect_frontend_to_install
...
+          Selecting #{DEFAULT_FRONTEND} is recommended.
+          However, some extensions are still only compatible with the now deprecated #{LEGACY_FRONTEND}.
+
+        MSG
+    end
+
+    def detect_authentication_to_install
+      ENV['AUTHENTICATION'] ||
+        options[:authentication] ||
+        (Bundler.locked_gems.dependencies['solidus_auth_devise'] && 'solidus_auth_devise') ||
```
waiting-for-dev, line 289: Should we use `DEFAULT_AUTHENTICATION` instead of `'solidus_auth_devise'` in this line?

core/lib/generators/solidus/install/install_generator.rb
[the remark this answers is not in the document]
waiting-for-dev: Makes sense, thanks for clarifying.

core/lib/generators/spree/custom_user/custom_user_generator.rb
waiting-for-dev: I see. Thanks. Something for another PR.

core/lib/generators/solidus/install/app_templates/authentication/none.rb
elia: Ok, I pushed a new version that will both allow complete access to /admin and the ability to place orders with some inline documentation in the initializer that should give the ropes of what to do to roll your own authentication.

core/lib/generators/solidus/install/install_generator.rb
elia: Done thanks! 👍

core/lib/generators/solidus/install/app_templates/authentication/solidus_auth_devise.rb
```diff
@@ -0,0 +1,12 @@
...
+  say_status :skipping, 'solidus_auth_devise is already in the gemfile'
+else
+  bundle_command("add solidus_auth_devise")
+end
+
+if options[:auto_accept]
+  migrations_flag = options[:migrate] ? '--auto-run-migrations' : '--skip-migrations'
+end
+
+generate "solidus:auth:install #{migrations_flag}"
```
waiting-for-dev, line 12: I haven't tried, but I think it'll error unless `options[:auto_accept]` is given.

waiting-for-dev: TIL

```ruby
if false
  a = 1
end
a # => nil 🤯 
```

waiting-for-dev (approved):
Thanks, @elia! I did a few tests and everything worked fine, also when it comes to the dummy app on extensions.

adammathys (approved):

kennyadsl (approved):
Thanks Eli, looks refreshing and I look forward for the next steps in this direction!

PR #4673, opened 2022-10-12
Restore `--payment-method=` for `solidus:install` on v3.2

elia:
## Summary

Solidus PCP is now compatible with Ruby 3 and can be re-enabled, since the CLI flag has been restored we can also attach bolt to it and let user select the payment method with CLI flags.

I included the minimum required to make the installer work, since we don't have tests for it I tried with:

```
bin/sandbox --frontend=solidus_frontend --payment-method=bolt --with-authentication 
bin/sandbox --frontend=solidus_frontend --payment-method=none --with-authentication 
```

and variations, also including the same options but interactively.

## Checklist

- ~~[ ] I have added automated tests to cover my changes.~~
- ~~[ ] I have attached screenshots to demo visual changes.~~
- ~~[ ] I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).~~
- ~~[ ] I have updated the readme to account for my changes.~~

### DIFF
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -92,7 +92,8 @@ unbundled bin/rails generate solidus:install \
   --auto-accept \
   --user_class=Spree::User \
   --enforce_available_locales=true \
-  --with_authentication=false
+  --with_authentication=false \
+  --payment-method=none
   $@
 
 unbundled bin/rails generate solidus:auth:install \
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -52,17 +52,8 @@ fi
 
 cd ./sandbox
 cat <<RUBY >> Gemfile
-
 gem 'solidus', path: '..'
-gem 'solidus_auth_devise', '>= 2.1.0'
-gem 'rails-i18n'
-gem 'solidus_i18n'
-
-group :test, :development do
-  platforms :mri do
-    gem 'pry-byebug'
-  end
-end
+gem 'pry-byebug', group: [:test, :development], platforms: :mri
 RUBY
 
 replace_in_database_yml() {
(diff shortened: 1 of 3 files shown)


jarednorman (approved):

bin/sandbox
```diff
@@ -88,12 +79,7 @@ fi
 
 unbundled bundle install --gemfile Gemfile
 unbundled bin/rails db:drop db:create
-unbundled bin/rails generate solidus:install \
-  --auto-accept \
```
waiting-for-dev, line 92: I'm not super happy removing `auto-accept` by default, as it slows the development feedback. What do you think about keeping it?

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -248,6 +274,13 @@ def complete
     private
 
     def detect_frontend_to_install(bundler_context)
+      if options[:payment_method] != 'none'
+        say_status :warning, set_color(
+          "You selected a payment method that might require manual integration with `solidus_starter_frontend`",
```
waiting-for-dev, line 279: I think a better experience would be warning before the payment has been selected.

core/lib/generators/solidus/install/install_generator/install_frontend.rb
```diff
@@ -30,7 +30,14 @@ def install_solidus_frontend
           end
         end
 
-        @generator_context.generate("solidus_frontend:install #{@generator_context.options[:auto_accept] ? '--auto-accept' : ''}")
+        # Solidus bolt will be handled in the installer as a payment method.
+        begin
+          skip_solidus_bolt = ENV['SKIP_SOLIDUS_BOLT']
+          ENV['SKIP_SOLIDUS_BOLT'] = 'true'
```
waiting-for-dev, line 36: If we move it here, I think we should remove the option from solidus_frontend altogether. Otherwise we're dealing with unneeded complexity here. WDYT?

waiting-for-dev (commented):
Some preliminary comments. Thanks!

bin/sandbox
elia: I think it's no big deal either way, easy to add or remove locally as needed, so I'll add it back (I had to remove it in order to test out the interactive `solidus:install` wizard).

elia:
@jarednorman @waiting-for-dev I had to change a couple of things to check auth options for `.nil?` instead of using `||=` so to correctly support both true and false when provided as CLI flags.

jarednorman (approved):

waiting-for-dev (approved):
:+1:

PR #4675, opened 2022-10-14
Installer UI improvements

elia:
## Summary

- A few installer cleanups and removals of legacy options
- Created some helpers for handling interactive options
- Shortened the names of options to be provided to flags (e.g. `solidus_starter_frontend` → `starter`)
- Improved the layout of interactive questions (see screenshots)

Screen Shot 2022-10-04 at 12 44 36

## Checklist


sample/db/samples/assets.rb
```diff
@@ -20,7 +20,7 @@
 products[:solidus_tote] = Spree::Product.find_by!(name: "Solidus Tote")
 products[:ruby_tote] = Spree::Product.find_by!(name: "Ruby Tote")
 
-def image(name, type = "jpg")
+image = ->(name, type = "jpg") do
```
waiting-for-dev, line 23: :heart:

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -143,11 +143,6 @@ def configure_application
       end
     end
 
-    def plugin_install_preparation
```
waiting-for-dev, line 146: Are you going to need all of the old plugin installation system for https://github.com/solidusio/solidus/pull/4673?

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -61,23 +63,25 @@ def prepare_options
...
 
       # No reason to check for their presence if we're about to install them
       ENV['SOLIDUS_SKIP_MIGRATIONS_CHECK'] = 'true'
 
-      if options[:with_authentication] != nil
+      if options[:enforce_available_locales] != nil
         warn \
-          "DEPRECATION WARNING: using `solidus:install --with-authentication` is now deprecated. " \
-          "Please use --authentication instead (see --help for the full list of options)."
+          "DEPRECATION WARNING: using `solidus:install --enforce-available-locales` is now deprecated and has no effect. " \
```
waiting-for-dev, line 77: Why not remove the option without deprecating it first if it has no effect? The generator will just ignore unknown options.

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -61,23 +63,25 @@ def prepare_options
...
 
-        if options[:authentication].blank? && options[:with_authentication] == 'false'
-          # Don't use the default authentication if the user explicitly
-          # requested no authentication system.
-          options[:authentication] = 'none'
-        end
+      if options[:lib_name] != nil
+        warn \
+          "DEPRECATION WARNING: using `solidus:install --lib-name` is now deprecated and has no effect. " \
+          "The option is legacy and should be removed from scripts still using it."
```
waiting-for-dev, line 84: I'd like to understand... which kind of script could be using it?

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -61,25 +58,15 @@ def prepare_options
...
+      @selected_authentication = detect_authentication_to_install
 
       # Silence verbose output (e.g. Rails migrations will rely on this environment variable)
       ENV['VERBOSE'] = 'false'
 
       # No reason to check for their presence if we're about to install them
       ENV['SOLIDUS_SKIP_MIGRATIONS_CHECK'] = 'true'
 
-      if options[:with_authentication] != nil
-        warn \
```
waiting-for-dev, line 72: If we're removing the warning, can't we remove the option altogether now?

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -11,15 +11,12 @@ class InstallGenerator < Rails::Generators::AppBase
 
     CORE_MOUNT_ROUTE = "mount Spree::Core::Engine"
 
-    LEGACY_FRONTEND = 'solidus_frontend'
-    DEFAULT_FRONTEND = 'solidus_starter_frontend'
```
waiting-for-dev, line 15: Up to you, but do you think it is ok to remove the constants (same for `DEFAULT_AUTHENTICATION` below)? It helps to check for string values that can be repeated multiple times (e.g., avoiding a typo that would create a mysterious bug).

waiting-for-dev (approved):
Thanks, Elia! I left some non-blocking comments. I haven't tested it, though, but I'm sure you have :slightly_smiling_face:

core/lib/generators/solidus/install/install_generator.rb
elia: Being #4673 targeted at 3.2 I don't think we need to keep this old system in place, and I don't see a scenario in which we'll keep needing it… but even in that case it should be easy to add it back.

core/lib/generators/solidus/install/install_generator.rb
elia: That's a good point, I'll try with a local variable first and fall back to a constant if that doesn't work

Context:
Those were removed because the role was a bit confusing, and with checks on the gemfile the default is actually dynamic (e.g. if you have the legacy one in the gemfile and `auto_accept` it will "default" to it, etc.).

core/lib/generators/solidus/install/install_generator.rb
elia: The warning is still there, just moved it to `detect_authentication_to_install`.

core/lib/generators/solidus/install/install_generator.rb
elia: `Rails::Generators::AppBase` is capable of accepting one argument as the app name/place (like in `rails new`), we don't need to specify it (defaults to `Rails.root`) but that also means that any unrecognized option will be interpreted as it. 

I would also argue that any existing script with formerly supported arguments (e.g. sandbox scripts) will be better off being notified it's no longer a supported option.

core/lib/generators/solidus/install/install_generator.rb
elia: Scripts for generating the sandbox inside extensions are the most likely to have legacy options, also I think we had an example app to for the heroku button (not sure it's still maintained tho).

Over time would probably be safer to reject unknown options instead of silently accepting them, wdyt?

elia (commented):
@waiting-for-dev added answers to all your comments, even if optional I'd like to have your take in an additional review if you have time 🙏

core/lib/generators/solidus/install/install_generator.rb
waiting-for-dev: Oh, yeah, sorry, forget about it.

core/lib/generators/solidus/install/install_generator.rb
waiting-for-dev: Yeah, that's a good point :+1:

core/lib/generators/solidus/install/install_generator.rb
waiting-for-dev: Makes sense.

About rejecting unknown options, I think that should be the default.

core/lib/generators/solidus/install/install_generator.rb
waiting-for-dev: Cool. That's non-blocking, though :slightly_smiling_face:

kennyadsl (approved):
Thanks Eli, the installer looks way better now!

PR #4677, opened 2022-10-17
Add support for sorting store credits with different algorithms

tmtrademarked:
## Summary

In our store, we have a set of custom eligibility rules for store credits based on the contents of the order. So we need a hook to inject a different way to prioritize store credits, as opposed to simply the type priority. While it might be possible to monkey patch the `in_priority_order` scope, that feels somewhat wrong - and the `add_store_credit_payment` method of `Spree::Order` does a *lot* that we would prefer not to monkey patch.

This PR adds a very simple configuration hook in the middle of this method to allow stores to inject some new behavior. By default, this simply does what the existing method did - sort by credit priority.

## Checklist

- [ X ] I have written a thorough PR description.
- [ X ] I have kept my commits small and atomic.
- [ X ] I have used clear, explanatory commit messages.
- [ X ] I have added automated tests to cover my changes.

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -596,8 +596,9 @@ module Spree
 
       if matching_store_credits.any?
         payment_method = Spree::PaymentMethod::StoreCredit.first
+        sorter = Spree::Config.store_credit_prioritizer_class.new(matching_store_credits, self)
 
-        matching_store_credits.order_by_priority.each do |credit|
+        sorter.call.each do |credit|
           break if remaining_total.zero?
           next if credit.amount_remaining.zero?
 
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -532,6 +532,15 @@ module Spree
                                                           small: '400x400>',
                                                           product: '680x680>',
                                                           large: '1200x1200>' }
+
+    # Allows providing your own class for prioritizing store credit application
+    # to an order.
+    #
+    # @!attribute [rw] store_credit_prioritizer_class
+    # @return [Class] a class with the same public interfaces as
+    #   Spree::StoreCreditPrioritizer.
+    class_name_attribute :store_credit_prioritizer_class, default: 'Spree::StoreCreditPrioritizer'
+
     # @!attribute [rw] taxon_image_style_default
     #
     # Defines which style to default to when style is not provided
--- a/core/app/models/spree/store_credit_prioritizer.rb
+++ b/core/app/models/spree/store_credit_prioritizer.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Spree
+  class StoreCreditPrioritizer
+    def initialize(credits, _order)
+      @credits = credits
+    end
+
+    def call
+      credits.order_by_priority
+    end
+
+    private
+
+    attr_reader :credits
+  end
+end
--- a/core/spec/models/spree/store_credit_prioritizer_spec.rb
+++ b/core/spec/models/spree/store_credit_prioritizer_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Spree::StoreCreditPrioritizer, type: :model do
+  let(:order) { create(:order) }
+  let(:credits) { Spree::StoreCredit.all }
+  let(:sorter) { described_class.new(credits, order) }
+
+  describe '#call' do
+    subject { sorter.call }
+
+    let(:credit_type_1) { create(:primary_credit_type, priority: '30') }
+    let!(:credit_1) { create(:store_credit, credit_type: credit_type_1) }
+    let(:credit_type_2) { create(:primary_credit_type, priority: '20') }
+    let!(:credit_2) { create(:store_credit, credit_type: credit_type_2) }
+    let(:credit_type_3) { create(:primary_credit_type, priority: '10') }
+    let!(:credit_3) { create(:store_credit, credit_type: credit_type_3) }
+
+    it 'returns the credits ordered by their priority' do
+      expect(subject.to_a).to eq([credit_3, credit_2, credit_1])
+    end
+  end
+end


gsmendoza:
Hi @tmtrademarked ! I think your PR makes sense! I only have a few suggestions to get it approved faster :)

1. Do squash the "Fix spec and remove usless order spec" commit into the "Adding support for sorting store credits with different algorithms" commit.
2. Check the links in https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines about commit message formatting. In particular,
  a. Write your commit message in the imperative: "Fix bug" and not "Fixed bug"
or "Fixes bug."
  b. Wrap the explanatory to about 72 characters or so.
  
Cheers!

core/app/models/spree/order.rb
```diff
@@ -596,8 +596,9 @@ def add_store_credit_payments
 
       if matching_store_credits.any?
         payment_method = Spree::PaymentMethod::StoreCredit.first
+        sorter = Spree::Config.order_credit_sorter_class.new(self)
```
kennyadsl, line 599: What about passing store credits to sort in the initializer instead of the order? Having the sorter know about the order is something that only applies in your domain, while the store credits will be needed in any scenario.

core/app/models/spree/order.rb
```diff
@@ -596,8 +596,9 @@ def add_store_credit_payments
...
 
-        matching_store_credits.order_by_priority.each do |credit|
+        sorter.in_sorted_order(matching_store_credits).each do |credit|
```
kennyadsl, line 601: What about just using `call` here? It is a sorter, and sorting is the only thing it is expected to do, so I guess `sorter.call` is enough semantic. 

Also, if you think my previous suggestion makes sense, this could be without any argument in the core, you could change the signature in your own implementation, passing the order. WDYT?

kennyadsl (commented):
Thanks Tom, this is a good change! 

I left a couple of ideas, let me know what you do think!

tmtrademarked:
> 2\. https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines

Thanks, @gsmendoza! I've squashed the commits, and tweaked the commit message. Let me know what you think?

core/app/models/spree/order.rb
tmtrademarked: I'm definitely fine with passing the credits in the constructor! Will update after we're aligned on the other comment, though.

kennyadsl:
> But I'm slightly confused by your suggestion, which makes me think I don't quite understand something. 

No, you are totally right! The core would continue to pass no arguments to the `call` method so it won't work. I guess that passing both in the constructor is fine. Just another thought: maybe the signature will be better having `credits` as the first argument and `order` as the second? I know it might look counterintuitive because `order` seems like a more broad object which contains `credits` (even if at this stage this is not true) but, for this method, `order` is more of an extra parameter. In fact, it is not even used in the default implementation, and it would be a bit weird being the first argument. Also, if we want to add more extra parameters in the future, it would be strange to have these extra parameters with `credits`, the main one, in between, eg `.new(order, credits, subscription)`. WDYT?

tmtrademarked:
> > But I'm slightly confused by your suggestion, which makes me think I don't quite understand something.
> 
> No, you are totally right! The core would continue to pass no arguments to the `call` method so it won't work. I guess that passing both in the constructor is fine. Just another thought: maybe the signature will be better having `credits` as the first argument and `order` as the second? I know it might look counterintuitive because `order` seems like a more broad object which contains `credits` (even if at this stage this is not true) but, for this method, `order` is more of an extra parameter. In fact, it is not even used in the default implementation, and it would be a bit weird being the first argument. Also, if we want to add more extra parameters in the future, it would be strange to have these extra parameters with `credits`, the main one, in between, eg `.new(order, credits, subscription)`. WDYT?

Yeah, that makes sense to me! I can understand the argument that `credits` is primary in this circumstance. Let me apply the suggestion!

core/app/models/spree/order_credit_sorter.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Spree
+  class OrderCreditSorter
+    def initialize(credits, _order)
+      @credits = credits
+    end
+
+    def sort
```
tmtrademarked, line 9: @kennyadsl - I went with `sort` as the method name which felt like a slightly clearer verb to me (personally). Let me know if you think `call` would be more inline with the overall conventions of Solidus, though?

tmtrademarked:
It *seems* like the spec failures here are flakes unrelated to this change, from all the logs I can see.

gsmendoza:
> Thanks, @gsmendoza! I've squashed the commits, and tweaked the commit message. Let me know what you think?

Looks good @tmtrademarked ! Thank you!

tmtrademarked:
It seems like a few different tests have failed, but none appear related to my changes. I don't have the ability to re-run the tests myself, unfortunately. Is anyone from Solidus able to trigger a re-run to see if that makes it pass?

tmtrademarked:
@kennyadsl @gsmendoza - any further thoughts about this one? Would love to be able to land this in a forthcoming release if at all possible. :-)

core/app/models/spree/order_credit_sorter.rb
kennyadsl: @waiting-for-dev What do you think? I'm still for using `call` for consistency, but we still don't have a real convention in Solidus.

waiting-for-dev: Yeah, I'm more in favor of using `#call`. That's the convention we want to use for objects that do a single action, where the class name describes that action.

core/lib/spree/app_configuration.rb
```diff
@@ -415,6 +415,14 @@ def default_pricing_options
     #   Spree::OrderContents.
     class_name_attribute :order_contents_class, default: 'Spree::OrderContents'
 
+    # Allows providing your own class for prioritizing store credit application
+    # to an order.
+    #
+    # @!attribute [rw] order_credit_sorter_class
+    # @return [Class] a class with the same public interfaces as
+    #   Spree::OrderCreditSorter.
+    class_name_attribute :order_credit_sorter_class, default: 'Spree::OrderCreditSorter'
```
waiting-for-dev, line 424: What about `:store_credit_sorter_class` & `Spree::StoreCreditSorter` as the names? I'm not very familiar with that sub-system, so reading "order credit" made me think that maybe it was something different than store credits.

core/spec/models/spree/order_credit_sorter_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    let!(:primary_credit) { create(:store_credit) }
+    let!(:secondary_credit) { create(:store_credit, credit_type: create(:secondary_credit_type)) }
+    
+    it 'returns the expected credits' do
+      expect(subject).to contain_exactly(primary_credit, secondary_credit)
+    end
+
+    it 'uses the default sorting order' do
+      expect(credits).to receive(:order_by_priority)
```
waiting-for-dev, line 21: Maybe a matter of taste, but I'd prefer to see the actual behavior being tested instead of the implementation details. I'd assert on the returned order instead.

waiting-for-dev (commented):
Thanks @tmtrademarked, I left some suggestions 🙂

core/lib/spree/app_configuration.rb
tmtrademarked: I'm fine with that name if we prefer - my only possible reservation about it is that there could be multiple contexts in which you'd want to "sort store credits". For example, maybe the UI for rendering the list of a users credits. That could be slightly confusing, potentially.

What do you think, @waiting-for-dev ? I'm more than happy to change the name if you think that makes the purpose more obvious!

core/spec/models/spree/order_credit_sorter_spec.rb
tmtrademarked: I can do that - will update!

core/lib/spree/app_configuration.rb
waiting-for-dev: That's a good point. What about something along the lines of `Spree::StoreCreditUtilizationSorter`? (feel free to find a better name 🙂 )

tmtrademarked: Renamed to `Spree::StoreCreditPrioritizer` - thoughts?

waiting-for-dev (approved):
It looks perfect now. Thanks, @tmtrademarked! Last ask, could you please squash the commits so that we don't pollute the git history with the details of the discussion in this PR?

kennyadsl (approved):
I second what @waiting-for-dev asked about squashing. Excellent work @tmtrademarked, thanks!

tmtrademarked:
Thanks guys - commits are squashed!

jarednorman (approved):

PR #4680, opened 2022-10-20
Update product_option_types Seed File

Naokimi:
## Summary

When tinkering with a freshly created solidus app, I was surprised to see that Solidus Women's T-Shirt was missing an option to add variants.

image

Further investigation revealed that this due to Spree::ProductOptionType not being created when running seeds and that it impacts all clothing objects. I understand that this might have been done due to lack of images for the other variants, but one would expect to see the option to manually add the variants (and related images) through the admin panel.

With this commit all clothing options are assigned variants and the `New Variant` button i made visible in the admin panel:

image

It also brings the option types in line with the ones displayed in demo.solidus.io

image

## Checklist

- [ ] ~~I have added automated tests to cover my changes.~~
- [x] I have attached screenshots to demo visual changes.
- [ ] ~~I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).~~
- [ ] ~~I have updated the readme to account for my changes.~~

### DIFF
--- a/sample/db/samples/product_option_types.rb
+++ b/sample/db/samples/product_option_types.rb
@@ -5,6 +5,12 @@ Spree::Sample.load_sample("products")
 size = Spree::OptionType.find_by!(presentation: "Size")
 color = Spree::OptionType.find_by!(presentation: "Color")
 
-solidus_tshirt = Spree::Product.find_by!(name: "Solidus T-Shirt")
-solidus_tshirt.option_types = [size, color]
-solidus_tshirt.save!
+colored_clothes = [
+  "Solidus T-Shirt", "Solidus Long Sleeve", "Solidus Women's T-Shirt"
+]
+
+Spree::Product.all.each do |product|
+  product.option_types = [size]
+  product.option_types << color if colored_clothes.include?(product.name)
+  product.save!
+end
--- a/sample/db/samples/variants.rb
+++ b/sample/db/samples/variants.rb
@@ -157,6 +157,96 @@ variants = [
     option_values: [medium, black],
     sku: "SOL-WM006",
     cost_price: 17
+  },
+  {
+    product: solidus_snapback_cap,
+    option_values: [small],
+    sku: "SOL-SNC02",
+    cost_price: 17
+  },
+  {
+    product: solidus_snapback_cap,
+    option_values: [medium],
+    sku: "SOL-SNC03",
+    cost_price: 17
+  },
+  {
+    product: solidus_snapback_cap,
+    option_values: [large],
+    sku: "SOL-SNC04",
+    cost_price: 17
+  },
+  {
+    product: solidus_hoodie,
+    option_values: [small],
+    sku: "SOL-HD02",
+    cost_price: 27
+  },
+  {
+    product: solidus_hoodie,
+    option_values: [medium],
+    sku: "SOL-HD03",
+    cost_price: 27
+  },
+  {
+    product: solidus_hoodie,
+    option_values: [large],
+    sku: "SOL-HD04",
+    cost_price: 27
+  },
+  {
+    product: ruby_hoodie,
+    option_values: [small],
+    sku: "RUB-HD02",
+    cost_price: 27
+  },
+  {
+    product: ruby_hoodie,
+    option_values: [medium],
+    sku: "RUB-HD03",
+    cost_price: 27
+  },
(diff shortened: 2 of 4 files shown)


sample/db/samples/product_option_types.rb
```diff
@@ -5,6 +5,14 @@
...
 
-solidus_tshirt = Spree::Product.find_by!(name: "Solidus T-Shirt")
-solidus_tshirt.option_types = [size, color]
-solidus_tshirt.save!
+clothes = Spree::Product.where(
+  name: [
+    "Solidus T-Shirt", "Solidus Long Sleeve", "Solidus Women's T-Shirt",
+    "Solidus Snapback Cap", "Solidus Hoodie Zip", "Ruby Hoodie",
+    "Ruby Hoodie Zip", "Ruby Polo"
+  ]
```
kennyadsl, line 13: Do all of these products have size/color variants? If yes 👍, if no what's the benefit of adding the option types when there's only the master variant?

Naokimi: So I'll try to express again the thought process behind these changes:

1. I installed a fresh solidus app locally
2. Noticed that Solidus Women's T-Shirt did not have any variants (size or colour)
3. Went to admin panel to add the variants manually
4. Couldn't find how to add a variant

After further investigation I understood that it was because the option types were not assigned to the product. 

I just thought it would be a good idea to skip the step of adding the option type and be able to add the variant from the get go.

kennyadsl: The thing is that you add one or more option types to a product when that product has variants. Having an option type assigned but no variants might be considered an inconsistency because it means we are assuming that the product has some variants, but we didn't create them yet. 

On a side note, in terms of consistency, with the clothing products we are creating in the demo, it's acceptable not to have the color option type, but maybe the size option type and its variants (eg. s/m/l) should always be present.

My proposal is to add the size option types and variants to all of them and keep the `Solidus T-Shirt` with both. WDYT?

Naokimi: ok, will modify as suggested

kennyadsl: By the way, maybe it's not clear or easy to add a variant to a product without option types. If you have any suggestions on how to improve it, let me know!

Naokimi: Fixed in https://github.com/solidusio/solidus/pull/4680/commits/5875a59b43db3c69e4508fe03f297070b9511170

waiting-for-dev: > My proposal is to add the size option types and variants to all of them and keep the Solidus T-Shirt with both. WDYT?

If that's what we want to communicate, I think it'd be better to refactor instead of:

1. Manually select clothing products and add size & color option types.
2. Manually select the rest of the clothing products and add the size option type.

Do:

1. Manually select clothing products and add color option type.
2. Select all clothing products and add the size option type.

That way, we don't need to visually confirm that we do not forget any clothing product to add the size, plus we're safe if we add new items.

WDYT?

sample/db/samples/variants.rb
```diff
@@ -157,6 +157,96 @@
     option_values: [medium, black],
     sku: "SOL-WM006",
     cost_price: 17
+  },
+  {
+    product: solidus_snapback_cap,
+    option_values: [small],
```
waiting-for-dev, line 163: Now that we're adding a bunch of new variants, it'd be suitable to leverage automation to create all the possible combinations. However, you're building on top of what we already had, so feel free to ignore this suggestion if you think it deserves a separate PR.

waiting-for-dev (commented):
Thanks, @Naokimi! Left two suggestions, let me know what you think :slightly_smiling_face:

jarednorman (approved):

sample/db/samples/variants.rb
Naokimi: would personally prefer to refactor it in a separate PR

sample/db/samples/product_option_types.rb
Naokimi: Implemented you suggestion in https://github.com/solidusio/solidus/pull/4680/commits/9eda486ef6821034f6957802c7473ff672aba51a

waiting-for-dev (approved):
❤️

PR #4694, opened 2022-10-25
[v3.2] make frontend installer shell-out commands more compatible

elia:
## Summary

- Avoid using <code>Kernel#`</code> for shelling out
- Track the corresponding branch in SSF

Fixes https://github.com/solidusio/solidus/discussions/4687

## Checklist

### DIFF
--- a/core/lib/generators/solidus/install/install_generator/install_frontend.rb
+++ b/core/lib/generators/solidus/install/install_generator/install_frontend.rb
@@ -22,12 +22,20 @@ module Solidus
 
       private
 
+      def bundle_command(command, env = {})
+        @generator_context.say_status :run, "bundle #{command}"
+        BundlerContext.bundle_cleanly do
+          system(
+            Gem.ruby,
+            Gem.bin_path("bundler", "bundle"),
+            *command.shellsplit,
+          )
+        end
+      end
+
       def install_solidus_frontend
         unless @bundler_context.component_in_gemfile?(:frontend)
-          BundlerContext.bundle_cleanly do
-            `bundle add solidus_frontend`
-            `bundle install`
-          end
+          bundle_command 'add solidus_frontend'
         end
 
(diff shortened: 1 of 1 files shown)


core/lib/generators/solidus/install/install_generator/install_frontend.rb
```diff
@@ -11,23 +11,31 @@ def initialize(bundler_context:, generator_context:)
...
           install_solidus_frontend
         when 'solidus_starter_frontend'
-          install_solidus_starter_frontend(installer_adds_auth)
+          install_solidus_starter_frontend
         end
       end
 
       private
 
+      def bundle_command(command, env = {})
```
waiting-for-dev, line 25: Not sure I understand the need for this change. In which situation is <code>Kernel#`</code> incompatible? It looks like we're adding unneeded complexity, but I'm sure I'm missing something.

waiting-for-dev (commented):
Thanks, @elia! I left a question regarding the last commit. I probably miss some context.

core/lib/generators/solidus/install/install_generator/install_frontend.rb
elia: The incompatibility comes down to Kernel#` sending the whole command to the current shell (which is not guaranteed to be `bash` on windows). The reason for this helper is surely questionable, but I went straight to the original Rails implementation instead of trying to find it out on my own. That's where I got that we could use `Gem.ruby` etc. for composing the command to be sent to system. I did that of course because I know Rails works on windows and it seemed the safest bet.

waiting-for-dev (approved):
🚀

core/lib/generators/solidus/install/install_generator/install_frontend.rb
waiting-for-dev: Fair enough! Let's go with it!

Issue #4433, opened 2022-06-21
[Admin] Unsupported use of can? with array of actions


spaghetticode:
**Solidus Version**:
all including master

**Description**
In some views we're probably using `cancancan` in an unsupported way. 

Instead of passing a single action, we provide an array with multiple values. For example, in [app/views/spree/admin/users/orders.html.erb](https://github.com/solidusio/solidus/blob/aa77140e8c5b07ab597126ba76524a3931bbb449/backend/app/views/spree/admin/users/orders.html.erb#L41) we're doing:
```erb
<% if can?([:admin, :edit], order) %>
```
which seems to mean that if the user has permissions for both actions, then they should be able to pass the check. 

Unfortunately, this works only for users that can `:manage` the resource.

Let's take a look at the [cancancan codebase](https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/relevant.rb#L8), where the rule check happens: 

```ruby
    def relevant?(action, subject)
      subject = subject.values.first if subject.class == Hash
      @match_all || (matches_action?(action) && matches_subject?(subject))
    end

    private

    def matches_action?(action)
      @expanded_actions.include?(:manage) || @expanded_actions.include?(action)
    end
```
 
we're mostly interested in the `matches_action?` method, where the first part is verified if the user has `:manage` permissions, while the second is verified when the one dimensional array `@expanded_actions` includes `action`, which is an array in our case... and this will never happen.

I don't know if the array match was dropped at some point in the past or if the feature was never there (the [documentation](https://github.com/CanCanCommunity/cancancan/blob/develop/docs/define_check_abilities.md) doesn't mention that feature), anyway I think we should change our codebase in order to support the current behavior.

Also, checking `:admin` seems to be a bit redundant, as it's going to [be checked at the controller level](https://github.com/solidusio/solidus/blob/master/backend/app/controllers/spree/admin/base_controller.rb#L29) when checking permissions for the `edit` action.

jacobherrington:
Glancing at the docs it seems like `can` is able to receive an array, but `can?` cannot :smile: 

```
# valid
can [:admin, :manage], Spree::Order, number: 'R987654321'
```

It's not too much of a leap to:

```
# invalid
can?([:create, :update], Spree::ProductProperty)
```

adammathys:
It looks like this was fixed by #4705

PR #4705, opened 2022-11-07
Do not pass arrays to can?

jacobherrington:
## Summary

#4433 notes that we are, in several places, passing arrays to the `can?` method. According to the documentation, that's not supported, so this low touch change doesn't do that and behaves more like what we might expect.

@spaghetticode you mentioned that checking for `:admin` might be redundant; should I include removing that check in this change? Should we rely only on the controller level permissions?

## Checklist

~- [ ] I have added automated tests to cover my changes.~ (maybe we need tests for this?)
~- [ ] I have attached screenshots to demo visual changes.~
~- [ ] I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).~
~- [ ] I have updated the readme to account for my changes.~

### DIFF
--- a/backend/app/views/spree/admin/product_properties/index.html.erb
+++ b/backend/app/views/spree/admin/product_properties/index.html.erb
@@ -36,7 +36,7 @@
       </tbody>
     </table>
 
-    <% if can?([:create, :update], Spree::ProductProperty) %>
+    <% if can?(:create, Spree::ProductProperty) && can?(:update, Spree::ProductProperty) %>
       <%= render 'spree/admin/shared/edit_resource_links' %>
     <% end %>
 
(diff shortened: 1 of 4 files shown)


spaghetticode:
@jacobherrington thanks for working on this 🙏 

In regards to your question, I'm not an expert on CanCan authorizations, so I would not recommend that change based only on my opinion. Also, I think that's at best a minor issue, so I'd say let's keep the `:admin` permission check, unless somebody has a more informed opinion.

jacobherrington:
@spaghetticode That makes sense to me, I was hoping you were the informed one :laughing: 

Thanks!

waiting-for-dev (approved):
Thanks for your collaboration ❤️ 

As per the redundancy of `:admin`, it certainly looks like that. However, as this is something related to security and not comprehensively tested, I'd prefer to err on the safe side and double-check on both layers.

PR #4719, opened 2022-11-15
Automatically update the release draft when merging PRs

waiting-for-dev:
## Summary

We add a GitHub action to automatically update the next [release draft](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository) whenever a PR is merged.

A PR is categorized under at least one of the changelog categories [according to its labels](https://github.com/solidusio/solidus/pull/4756). Labels are already automatically added by the [labeler GitHub action](https://github.com/actions/labeler). A PR can be skipped when the `changelog:skip` label is added.

The chosen format mimics what we've done in the [Changelog](https://github.com/solidusio/solidus/blob/0cae150e4f347cd9f232001bb945ecc9974abb29/CHANGELOG.md), so we can leverage this work further to automate its generation.

We prepend the draft content with a comment warning about its automatic generation and append a link to the list of new commits since the previous tag.

The following tag is computed like this:

- We take the current tag in the base branch thanks to a custom callable workflow that extracts the information from the pipeline.
- For PRs to master, the following tag is the subsequent minor or major version depending on the value of `Spree::VERSION`.
- The following tag is the next patch version for PRs to branches corresponding to supported versions.

We considered using [Release Drafter action](https://github.com/marketplace/actions/release-drafter) as an alternative, but it reaches GitHub API abuse limit in our repo.

## Checklist

### DIFF
--- a/.github/workflows/extract_pipeline_context.yml
+++ b/.github/workflows/extract_pipeline_context.yml
@@ -1,0 +1,28 @@
+name: "Extract pipeline context"
+
+on:
+  workflow_call:
+    outputs:
+      current_tag:
+        description: "Current tag"
+        value: ${{ jobs.extract.outputs.current_tag }}
+      candidate_tag:
+        description: "Candidate tag"
+        value: ${{ jobs.extract.outputs.candidate_tag }}
+
+jobs:
+  extract:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: git fetch --tags origin
+      - id: extraction
+        name: "Extract context for the current pipeline"
+        run: |
+          cd dev_tools/ && bin/extract-pipeline-context \
+            $GITHUB_OUTPUT \
+            ${{ github.ref_name }} \
+            `git tag --list`
+    outputs:
+      current_tag: ${{ steps.extraction.outputs.current_tag }}
+      candidate_tag: ${{ steps.extraction.outputs.candidate_tag }}
--- a/dev_tools/lib/solidus/release_drafter.rb
+++ b/dev_tools/lib/solidus/release_drafter.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+require_relative 'release_drafter/client'
+require_relative 'release_drafter/builder'
+
+module Solidus
+  # Coordinates updating the release draft on GitHub when a PR is merged.
+  #
+  # @api private
+  class ReleaseDrafter
+    LABEL_PREFIX = 'changelog:'
+
+    LABELS = {
+      "#{LABEL_PREFIX}solidus_core" => 'Solidus Core',
+      "#{LABEL_PREFIX}solidus_backend" => 'Solidus Backend',
+      "#{LABEL_PREFIX}solidus_api" => 'Solidus API',
+      "#{LABEL_PREFIX}solidus_sample" => 'Solidus Sample',
+      "#{LABEL_PREFIX}solidus" => 'Solidus'
+    }.freeze
+
+    SKIP_LABEL = "#{LABEL_PREFIX}skip"
+
+    NO_EDIT_WARNING = <<~TXT
+      <!-- Please, don't edit manually. The content is automatically generated. -->
+    TXT
+
+    def initialize(github_token:, repository:,
+                   client: Client.new(github_token: github_token, repository: repository))
+      @repository = repository
+      @client = client
+    end
+
+    def call(pr_number:, current_tag:, candidate_tag:, branch:)
+      pr = @client.fetch_pr(pr_number: pr_number)
+      pr_labels = pr.labels.map(&:name)
+      matching_labels = LABELS.keys & pr_labels
+      return if pr_labels.include?(SKIP_LABEL) || matching_labels.empty?
+
+      draft = @client.fetch_draft(tag: candidate_tag)
+      Builder.new(draft: draft, categories: LABELS.values, prepend: NO_EDIT_WARNING, append: full_changelog(current_tag, candidate_tag))
+        .then { |builder| add_pr(builder, pr, matching_labels) }
+        .then { |draft| save_release(draft, candidate_tag, branch) }
+    end
+
+    private
+
+    def builder(draft, current_tag, candidate_tag)
+      Builder.new(draft: draft, categories: LABELS.values, prepend: NO_EDIT_WARNING, append: full_changelog(current_tag, candidate_tag))
+    end
+
+    def add_pr(builder, pr, labels)
+      builder.add(
+        number: pr.number,
+        categories: LABELS.values_at(*labels),
+        title: pr.title,
+        user: pr.user.login
+      )
+    end
+
+    def save_release(draft, candidate_tag, branch)
+      if draft.new?
+        @client.create_draft(draft: draft, tag: candidate_tag, branch: branch)
+      else
+        @client.update_draft(draft: draft, tag: candidate_tag, branch: branch)
+      end
+    end
+
+    def full_changelog(current_tag, candidate_tag)
+      <<~TXT
+
+
+        **Full Changelog**: https://github.com/#{@repository}/compare/#{current_tag}...#{candidate_tag}
+      TXT
+    end
+  end
+end
--- a/dev_tools/lib/solidus/release_drafter/builder.rb
+++ b/dev_tools/lib/solidus/release_drafter/builder.rb
@@ -1,0 +1,139 @@
+# frozen_string_literal: true
+
+require 'set'
+
+module Solidus
+  class ReleaseDrafter
+    # Defines the template for releases and adds entries to it.
+    #
+    # @api private
+    class Builder
+      TITLE_PREFIX = '## '
+
+      TITLE_TEMPLATE = lambda do |category|
+        TITLE_PREFIX + category
+      end.freeze
+
+      SEPARATOR = "\n"
+
+      TITLE_SEPARATOR = SEPARATOR
+
+      ENTRY_TEMPLATE = lambda do |title, number, user|
+        "- #{title} ##{number} (@#{user})"
+      end.freeze
+
+      ENTRY_SEPARATOR = SEPARATOR
+
+      SECTION_SEPARATOR = SEPARATOR + SEPARATOR
+
+      def initialize(draft:, categories:, prepend: '', append: '')
+        @categories = Set[*categories]
+        @prepend = prepend
+        @prepend_pattern = /^#{Regexp.quote(@prepend)}/m
+        @append = append
+        @append_pattern = /#{Regexp.quote(@append)}$/m
+        @draft = draft.new? ? draft.with(content: initial_content) : draft
+        sanitize_prepend_append
+        @ast = build_ast
+        sanitize_categories
+      end
+
+      def add(title:, number:, user:, categories:)
+        ENTRY_TEMPLATE.(title, number, user)
+          .then { |entry| add_to_ast(entry, categories) }
+          .then { |ast| @draft.with(content: unparse(ast)) }
+      end
+
+      private
+
+      def initial_content
+        @prepend + @categories.map do |category|
+          TITLE_TEMPLATE.(category)
+        end.join(SECTION_SEPARATOR) + @append
+      end
+
+      def build_ast
+        generated_lines.reduce([nil, {}]) do |(current_category, ast), line|
+          maybe_next_category = line.match(/^#{TITLE_PREFIX}(.*)$/)&.[](1)
+          if line.strip.empty?
+            [current_category, ast]
+          elsif maybe_next_category
+            [maybe_next_category, ast.merge(maybe_next_category => [])]
+          elsif current_category
+            [current_category, ast.merge(current_category => ast[current_category] + [line])]
+          else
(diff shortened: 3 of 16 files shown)


bin/update_release_draft
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env ruby
```
elia, line 1: nit: maybe worth naming this file with dashes for consistency with other files in bin/

core/lib/spree/release_drafter.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
```
elia, line 1: This set of files seems completely detached from solidus shipped code, can we find a better place for it?
I guess the problem could be where to put the specs… but it also looks like they don't need any special setup and can be run independently, so maybe an extra job in circleci would be fine.

I don't have great locations to suggest, one would be a subfolder of bin/, or of .github/workflows/, or maybe a new top-level folder?

waiting-for-dev: I also felt that there was something off in that regard. I didn't want to do a complex setup only for that, but it probably makes sense. As you said, nothing needs to be shipped to the user. That we need to branch on the ruby version is also a smell of the loose cohesion.

I'll investigate creating a new `tools` top directory with its own Gemfile. This way, we won't need to branch on the main Gemfile either.

bin/update_release_draft
waiting-for-dev: Thanks! Fixed!

core/lib/spree/release_drafter.rb
waiting-for-dev: We discussed it in the Core Team meeting and decided to give it more thought. We commented on different options:

- Top `tools` directory.
- Folder within `bin`.
- Separate gem
- Part of solidus_dev_support

For now, I've put it within a top `tools` directory for a better structure in the review. However, I still haven't added the CircleCI job, which means that the tests are not running now—putting the PR as a draft until we decide.

dev_tools/spec/lib/solidus/release_drater_spec.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'ostruct'
+require 'solidus/release_drafter'
+
+RSpec.describe Solidus::ReleaseDrafter do
```
gsmendoza, line 6: @waiting-for-dev Spec file name and folder name refers to `release_drater`; should be `release_drafter`.

gsmendoza (changes requested):
@waiting-for-dev Found one minor issue so far; still reviewing the PR...

dev_tools/spec/lib/solidus/release_drater_spec.rb
waiting-for-dev: Oh, thanks! Fixed 🙂

core/lib/spree/release_drafter.rb
waiting-for-dev: I've gone with a dedicated `dev_tools` folder. We're going to need more code for further automation, so I think it's good to have a proper structure for it. Going for a gem or making it part of soliidus_dev_support looks overkill at this point. Do you agree @elia, @kennyadsl?

waiting-for-dev: @kennyadsl, about your concern about a `dev_tools` being confusing for contributors, I've added a README file within with an explicit description. Let me know if that would address your concern.

gsmendoza (commented):
@waiting-for-dev I still have to go through the specs, and I'm hoping those would help me understand the classes better (especially the Builder class). It seems template-building software like this tend to become hard to read, simply because it's hard to imagine how the template looks like. I'll take some time tomorrow to read and run the specs. But please go ahead and continue your work on this PR; don't consider me a blocker :)

core/lib/spree/release_drafter.rb
elia: `solidus_dev_support` is meant to help supporting multiple solidus versions and not well suited for this, an external gem is definitely overkill 💯.

It's good anyway that this has no consequence on versions and we're free to adjust as we go, so I'm not too concerned with the name or structure 🙌.

kennyadsl: @waiting-for-dev works for me, thanks for adding the README. I think we also want to exclude this folder from the files we ship in the gem via the gemspec, right?

dev_tools/lib/solidus/release_drafter.rb
```diff
@@ -0,0 +1,70 @@
...
+    attr_reader :repository,
+                :next_tag,
+                :branch
+
+    def initialize(github_token:, repository:, latest_tag:, branch:, next_tag_resolver: NextTagResolver.new, client: Client)
+      @repository = repository
+      @latest_tag = latest_tag
+      @branch = branch
+      @next_tag = next_tag_resolver.(latest_tag: @latest_tag, branch: @branch)
+      @client = client.new(github_token: github_token, repository: repository, tag: @next_tag, branch: @branch)
```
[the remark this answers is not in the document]
waiting-for-dev: Great point. I implemented your suggestion 👍

waiting-for-dev:
> @waiting-for-dev I still have to go through the specs, and I'm hoping those would help me understand the classes better (especially the Builder class). It seems template-building software like this tend to become hard to read, simply because it's hard to imagine how the template looks like. I'll take some time tomorrow to read and run the specs. But please go ahead and continue your work on this PR; don't consider me a blocker :)

Yeah! Specs can help a lot here. Let me know if you need any help. And thanks for your detailed review ❤️

dev_tools/lib/solidus/release_drafter/builder.rb
```diff
@@ -0,0 +1,139 @@
...
+            [current_category, ast]
+          elsif maybe_next_category
+            [maybe_next_category, ast.merge(maybe_next_category => [])]
+          elsif current_category
+            [current_category, ast.merge(current_category => ast[current_category] + [line])]
+          else
+            [current_category, ast]
+          end
+        end[1]
+      end
```
gsmendoza, line 68: It seems the AbstractSyntaxTree methods (`build_ast`, `add_to_ast`, `unparse`) could be extracted to a separate AbstractSyntaxTree class. Doing so might simplify the Builder class and allow us to test the AbstractSyntaxTree methods directly. What do you think?

dev_tools/lib/solidus/release_drafter/builder.rb
```diff
@@ -0,0 +1,139 @@
...
+  class ReleaseDrafter
+    # Defines the template for releases and adds entries to it.
+    #
+    # @api private
+    class Builder
+      TITLE_PREFIX = '## '
+
+      TITLE_TEMPLATE = lambda do |category|
+        TITLE_PREFIX + category
+      end.freeze
```
gsmendoza, line 15: I'm curious to learn more about this pattern of implementing a small piece of functionality with a lambda. How does it differ from implementing the functionality with a class method (see below)?

```ruby
def self.title_template(category)
  TITLE_PREFIX + category
end
```

If you can point me to any docs or articles, I would appreciate it :)

gsmendoza (commented):
@waiting-for-dev Great work! I have some suggestions, but no blockers :)

dev_tools/lib/solidus/release_drafter/builder.rb
waiting-for-dev: @gsmendoza, I can't name any specific pattern. I guess it's just going all the way down to treating a function just like any other value. In this case, we have other constants that are not functions that play a very similar role, so it felt more natural to me that all of them were constants. We could also argue that we could define everything as class methods, like in:

```ruby
def self.title_prefix
  '##'
end
```

The outcome would be exactly the same 🙂

gsmendoza: @waiting-for-dev I can't remember this rule of thumb I picked up before, but it's like "use the simplest tool for the problem at hand." So, for example, if you can implement a solution with both a function and a class, consider the function solution first, since functions are simpler than classes.

Using this rule of thumb, it makes more sense to implement `'##'` as a constant instead of as a class method, since a constant is simpler than a class method. 

A corollary to this idea is that, if you choose a more complex tool to implement a solution, it gives the impression that there was a reason behind it.

Going back to the example, if you choose to implement `'##'` as a class method, then readers of your code might assume that you might have had a reason behind that, since class methods are more complex than constants.

With lambdas versus class methods, I think that lambdas is the more powerful/fancier/esoteric of the two.

Your thoughts?

waiting-for-dev: Absolutely, @gsmendoza. Sorry if I wasn't clear enough, but I wasn't arguing to move those constants to class methods. I was using it to illustrate why it shouldn't be shocking to see lambdas as class methods, as they're another class of value, just like a string 🙂

gsmendoza: @waiting-for-dev I won't say "shocking". More like, "intriguing" :smile:

dev_tools/lib/solidus/release_drafter/next_tag_resolver.rb
```diff
@@ -0,0 +1,25 @@
...
+    #
+    # @api private
+    class NextTagResolver
+      MAIN_BRANCH = 'master'
+
+      TAG_PREFIX = 'v'
+
+      TAG_SEPARATOR = '.'
+
+      def call(latest_tag:, branch:)
```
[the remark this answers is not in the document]
waiting-for-dev: Makes sense. Done.

dev_tools/lib/solidus/release_drafter/builder.rb
waiting-for-dev: I see pros & cons. For now, I'd prefer to leave it as it is as I'm worried about over-engineering it plus I already have spend a lot of time with this PR 😅

waiting-for-dev:
This is ready for review. I changed a few things:

- Skipping the addition when the `changelog:skip` label is present.
- Detecting the context for a major release from the `Spree::VERSION` constant.
- Updated to the new changelog label names.
- Minor refactors.

waiting-for-dev:
This is ready for review (once more 🙂 ). I had to re-do how we fetch the current tag in the pipeline, but the good news is that's going to be reused in #4778

elia (approved):
It's a lot of code! 🥵 🙌 

I left a couple of comments, no need to take action except for the typo, but given the amount of code maybe the evaluation for/against extracting this to a gem changes… in fact now I would be ok with doing that, could be in a different PR too, thoughts?

Anyway I'm approving, I'm also fine with merging, after all dev_tools is pretty self contained and is not adding too much noise 😊 ✅ 

(I'm assuming it works, didn't have the heart to try it 😅)

.github/workflows/extract_pipeline_context.yml
```diff
@@ -0,0 +1,28 @@
+name: "Expract pipeline context"
```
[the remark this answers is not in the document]
waiting-for-dev: Thanks ✅

dev_tools/lib/solidus/release_drafter/client.rb
```diff
@@ -0,0 +1,52 @@
...
+      def update_draft(draft:, tag:, branch:)
+        @client.update_release(
+          draft.url,
+          name: tag,
+          body: draft.content,
+          draft: true,
+          tag_name: tag,
+          target_commitish: branch
+        ) && draft
+      end
```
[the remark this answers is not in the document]
waiting-for-dev: Not sure I get it. The `Solidus::ReleaseDrafter::Client` class wraps both all the external requests and the Octokit API. On top of that, it works with our custom `Draft` data structure. We're also defining other methods, like `#fetch_pr` or `#fetch_draft`.

waiting-for-dev:
> but given the amount of code maybe the evaluation for/against extracting this to a gem changes… in fact now I would be ok with doing that, could be in a different PR too, thoughts?

I couldn't agree more that's a separate gem in spirit. However, at this point, it's highly tailored to our project. Extracting it right now would only make our processes more difficult. I'd like to stop and add a few abstractions to make it more generic, but that would require a considerable amount of time for something that it's not a high priority.

As there's other stuff coming next, I propose waiting until we have a clearer picture of everything. At that point, we can talk about what makes more sense.

> (I'm assuming it works, didn't have the heart to try it 😅)

I did 🙂 Hopefully, comprehensively enough 😅

dev_tools/lib/solidus/release_drafter/client.rb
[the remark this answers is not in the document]
elia: I could have misread it, the impression I got was that it could have inlined everywhere it's used since each method is only used in one place and the wrapping is rather thin. But as I said I have no issue in keeping it as it is 🙌

elia:
> As there's other stuff coming next, I propose waiting until we have a clearer picture of everything. At that point, we can talk about what makes more sense.

@waiting-for-dev agreed 💯

kennyadsl (approved):
I have to be honest: I didn't go too much into the details here, but the general logic and concepts flow well to me. Also, there's great test coverage, which makes me feel safe. Looking forward to testing this in the real world. Thanks, @waiting-for-dev!

gsmendoza (approved):

Issue #4720, opened 2022-11-16
Context for a proc given as a preference default changed from v3.0 to v3.1


waiting-for-dev:
Before v3.1, the proc given as the default value in a preference had its context bound to the initialized instance. That allowed, for example, referencing other preferences. From v3.1., the context is bound to the class.

**Solidus Version**:
v3.2, v3.1, v3.0

**To Reproduce**

On v3.0 you can do the following:

```ruby
Spree::Calculator::FlatRate.preference(:foo, :string, default: -> { preferred_currency })
a = Spree::Calculator::FlatRate.new
a.preferred_foo #=> "USD"
```

However, on Solidus v3.1:

```ruby
Spree::Calculator::FlatRate.preference(:foo, :string, default: -> { preferred_currency })
a = Spree::Calculator::FlatRate.new
# undefined local variable or method `preferred_currency' for main:Object (NameError)
```

**Current behavior**
The proc is bound to the class context, so a `NameError` is raised because the generated `preferred_` methods are not found.

**Expected behavior**
It should keep working as in v3.0.

**Additional context**
First reported on Slack: https://solidusio.slack.com/archives/C0JBKDF35/p1668417688273419

Roddoric:
I did a git bisect with the example above and found the commit introducing the behavior change
https://github.com/solidusio/solidus/commit/b897eccb6cdd8235d782275e9df069d95474cd6b

I'll try yo look into it has I need it fixed to update our Solidus version

waiting-for-dev:
I also suspected of that commit, but looking to the diff I don't see how it could break that 🤷 Thanks for looking into it, I'm curious about what's the source of the problem here 🙂

Roman2K:
The culprit does indeed seem to be that the commit found by @Roddoric.

It changes the context of execution of the default block when [defining the default getter method](https://github.com/solidusio/solidus/commit/b897eccb6cdd8235d782275e9df069d95474cd6b#diff-1d7628c9b0c063921fa05554bc947bcd7d512427a47f6e7d709256ccd66e3601R63):

Before that commit, the default block was passed as method body, so its execution context was the "preferable" instance:

```ruby
define_method preference_default_getter_method(name), &default
```

After that commit, the default block is merely called from another method body, so its execution context is wherever it was created at (before being passed to `preference(...)`):

```ruby
define_method preference_default_getter_method(name) do
  default.call(*context_for_default)
end
```

---

I see 2 solutions (not tested :nerd_face:):

[...]

PR #4721, opened 2022-11-16
Fix call context when a preference default is a proc

Roddoric:
## Summary

Before v3.1, the proc given as the default value in a preference had its context bound to the initialized instance. That allowed, for example, referencing other preferences.

From v3.1., the context is bound to the class.

We are coming back to the behavior by calling the proc with `instance_exec`

More context in the linked [Issue](https://github.com/solidusio/solidus/issues/4720)

## Checklist

### DIFF
--- a/core/lib/spree/preferences/preferable.rb
+++ b/core/lib/spree/preferences/preferable.rb
@@ -10,6 +10,9 @@ module Spree
     #
     # A class including Preferable must implement #preferences which should return
     # an object responding to .fetch(key), []=(key, val), and .delete(key).
+    # If #preferences is initialized with `default_preferences` and one of the
+    # preferences is another preference, it will cause a stack level too deep error.
+    # To avoid it do not memoize #preferences.
     #
     # It may also define a `#context_for_default` method. It should return an
     # array with the arguments to be provided to a proc used as the `default:`
@@ -111,6 +114,8 @@ module Spree
       end
 
       # @return [Hash{Symbol => Object}] Default for all preferences defined on this class
+      # This may raise an infinite loop error if any of the defaults are
+      # dependent on other preferences defaults.
       def default_preferences
         Hash[
           defined_preferences.map do |preference|
--- a/core/spec/models/spree/preferences/preferable_spec.rb
+++ b/core/spec/models/spree/preferences/preferable_spec.rb
@@ -3,105 +3,110 @@
 require 'rails_helper'
 
 RSpec.describe Spree::Preferences::Preferable, type: :model do
-  before :all do
-    class A
+  let(:config_class_a) do
+    Class.new do
       include Spree::Preferences::Preferable
       attr_reader :id
 
       def initialize
         @id = rand(999)
+        initialize_preference_defaults
       end
 
       def preferences
-        @preferences ||= default_preferences
+        @preferences ||= {}
+      end
+
+      def initialize_preference_defaults
+        @preferences = default_preferences
       end
 
       preference :color, :string, default: 'green'
     end
+  end
 
-    class B < A
+  let(:config_class_b) do
+    Class.new(config_class_a) do
       preference :flavor, :string
     end
   end
 
-  before :each do
-    @a = A.new
-    @b = B.new
-  end
+  let(:a) { config_class_a.new }
+  let(:b) { config_class_b.new }
 
   describe "preference definitions" do
     it "parent should not see child definitions" do
-      expect(@a.has_preference?(:color)).to be true
-      expect(@a.has_preference?(:flavor)).not_to be true
+      expect(a.has_preference?(:color)).to be true
+      expect(a.has_preference?(:flavor)).not_to be true
     end
 
     it "child should have parent and own definitions" do
-      expect(@b.has_preference?(:color)).to be true
-      expect(@b.has_preference?(:flavor)).to be true
+      expect(b.has_preference?(:color)).to be true
+      expect(b.has_preference?(:flavor)).to be true
     end
 
     it "instances have defaults" do
-      expect(@a.preferred_color).to eq 'green'
-      expect(@b.preferred_color).to eq 'green'
-      expect(@b.preferred_flavor).to be_nil
+      expect(a.preferred_color).to eq 'green'
+      expect(b.preferred_color).to eq 'green'
+      expect(b.preferred_flavor).to be_nil
     end
 
     it "can be asked if it has a preference definition" do
-      expect(@a.has_preference?(:color)).to be true
-      expect(@a.has_preference?(:bad)).to be false
+      expect(a.has_preference?(:color)).to be true
+      expect(a.has_preference?(:bad)).to be false
     end
 
     it "can be asked and raises" do
       expect {
-        @a.has_preference! :flavor
+        a.has_preference! :flavor
       }.to raise_error(NoMethodError, "flavor preference not defined")
     end
 
     it "has a type" do
-      expect(@a.preferred_color_type).to eq :string
-      expect(@a.preference_type(:color)).to eq :string
+      expect(a.preferred_color_type).to eq :string
+      expect(a.preference_type(:color)).to eq :string
     end
 
     it "has a default" do
-      expect(@a.preferred_color_default).to eq 'green'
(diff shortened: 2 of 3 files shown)


waiting-for-dev (dismissed):
Thanks!! ❤️

core/spec/models/spree/preferences/preferable_spec.rb
```diff
@@ -151,6 +151,16 @@ def self.allowed_admin_form_preference_types
       end
     end
 
+    context "when a default is a proc" do
+      before do
+        A.preference(:context, :string, default: -> { self.to_s })
+      end
+
+      it "calls it from the instance" do
+        expect(@a.preferred_context).to eq(@a.to_s)
```
kennyadsl, line 160: I'm a bit concerned that with time this could become a false positive (pretty sure it is not now, but wanted to ask anyway): what is the content of  `@a.to_s` at this point? 

I thought that maybe we could also (or instead) test that `@a` receives `to_s` so no matter what the value is, we are sure that `self` is the instance of `A`, which is what we want to cover here. 

Let me know your thoughts.

kennyadsl (commented):
Thanks, this is great! I left a question about the testing approach, let me know! ❤️

core/spec/models/spree/preferences/preferable_spec.rb
Roddoric: If I rollback the modification in `core/lib/spree/preferences/preferable_class_methods.rb` the test will fail like this:
```
expected: "#<A:0x000000010f6ff3c8>"
got: "#<RSpec::ExampleGroups::SpreePreferencesPreferable::PreferenceAccess::WhenADefaultIsAProc:0x000000010f684858>"
``` 
The `#to_s` runs in the test context, for me it looked enough

waiting-for-dev: I also thought that this test was a bit indirect. Another option that would allow us to avoid stubbing is testing for something in the state of the `A` instance. Like:

```ruby
A.preference(:context, :string, default: -> { preferred_color })
# ...
expect(@a.preferred_context).to eq('green')
```

The above would have the added benefit of covering probably the most common use case.

What do you think, @Roddoric?

Roddoric: Soooo, it was a good catch as it proved the fix did not actually work...

But that lead to others problems, so I resolve this conversation and start a new one...

core/spec/models/spree/preferences/preferable_spec.rb
```diff
@@ -301,6 +316,7 @@ def self.allowed_admin_form_preference_types
                      default: 'my_default_secret',
                      encryption_key: 'VkYp3s6v9y$B?E(H+MbQeThWmZq4t7w!'
 
+        @a = A.new
```
Roddoric, line 319: If this is not here and the test is not the first to run in the context the test will fail.

And I do no understand why it's necessary...

Roddoric:
I changed the test `#preferences` from
```ruby
      def preferences
         @preferences ||= default_preferences
       end
```
to
```ruby
      def preferences
         @preferences ||= {}
       end

       def initialize_preference_defaults
         @preferences = default_preferences.merge(preferences)
       end
```

Because `#default_preferences` will enter an infinite loop trying to evaluate the value of `preferred_context`
( default_preferences calls preferences which calls default_preferences...)

Any thoughts about it ?
Are the comments clear and should they stay ?

waiting-for-dev:
Thanks, @Roddoric. We definitely need to understand better what's going on here 🤔

core/spec/models/spree/preferences/preferable_spec.rb
waiting-for-dev: Hey, @Roddoric, did you find out why that's needed?

Roddoric: No, I'm not looking into it right now
I'll try to look into it during the weekend or next week 😬

waiting-for-dev: Sure! Let us know your findings 🙏

Roddoric: Okk

I think I found the problem...
Context:
- `class A` is declared once in a `before(:all)` so the class WILL NOT be destroyed/recreate for each test
- `@a = A.new` is in a `before(:each)` therefore the variable is initialized before the test start
- The 4 examples from l.285 define the same preference `:secret` but the last one is the only using a default

If the test run first:
- `@a` is instantiated without `preference :secret`
- `@a.preferences` looks like `{:color=>"green"}`
- We define the `preference :secret` with a default
- Calling `@a.get_preference(:secret)` will work because it will look for the default 

If the test doesn't run first:
- `@a` is instantiated with a `preference :secret` without `:default`
- `@a.preferences` looks like `{:color=>"green", :secret=>"rJ0s--jPH2d2iyyB9KseLU--HbYfXFtRB36+yj9L3fSDFA=="}
` (`:secret` value is `nil` but encrypted)
- We override `preference :secret` with a default
- `@a.get_preference(:secret)` returns `nil` as it's initialized and does not look for the default


Setting `@a = A.new` after the override works because it'll be initialized with the new default

So there is multiple ways to fix it:

1. Defining `class A` and `class B` in a `before(:each)` instead of a `before(:all)` which will make the tests truly independents
2. Using different names for the secrets which will avoid conflict
3. Changing `@a` and `@b` initialization from:
```ruby
  before :each do
    @a = A.new
    @b = B.new
  end
```
To:
```ruby
  let(:a) { A.new }
  let(:b) { B.new }
```
It will initialize the variable once called (after the override)

IMO I would go for 1 + 3 (as 2 looks more like a workaround)

Roddoric:
Hello @waiting-for-dev @kennyadsl 
Did you have some time to check the last changes and the comment ?

core/lib/spree/preferences/preferable.rb
```diff
@@ -10,6 +10,9 @@ module Preferences
     #
     # A class including Preferable must implement #preferences which should return
     # an object responding to .fetch(key), []=(key, val), and .delete(key).
+    # If #preferences is initialized with `default_preferences` and one of the
+    # preferences is another preference, it will cause a stack level too deep error.
+    # To avoid it do not momoize #preferences.
```
tvdeyen, line 15: Typo

```suggestion
    # To avoid it do not memoize #preferences.
```

Roddoric: Thx !

waiting-for-dev:
> Hello @waiting-for-dev @kennyadsl Did you have some time to check the last changes and the comment ?

Thanks, @Roddoric. I answered in the comment.

core/spec/models/spree/preferences/preferable_spec.rb
```diff
@@ -151,158 +156,170 @@ def self.allowed_admin_form_preference_types
...
 
       it "with string, default value" do
-        A.preference :secret, :encrypted_string,
+        config_class_a.preference :secret, :encrypted_string,
                      default: 'my_default_secret',
                      encryption_key: 'VkYp3s6v9y$B?E(H+MbQeThWmZq4t7w!'
 
-        expect(@a.get_preference(:secret)).to eq('my_default_secret')
-        expect(@a.preferences[:secret]).not_to eq('my_default_secret')
+        a = A.new
```
waiting-for-dev, line 320:
```suggestion
        a = config_class_a.new
```

I think that'll make tests green

waiting-for-dev (changes requested):

Roddoric:
Sorry for the delay...

@waiting-for-dev it should be fine now, all tests pass in local but I don't see the CI triggering

waiting-for-dev:
Thanks, @Roddoric. Last ask, could you squash all the commits? 🙏 

Yeah, we have some issues with CI. Once it's fixed we can go ahead.

waiting-for-dev:
@Roddoric, I know I said last ask, but as any good movie there's still a VERY last scene 😅 Could you please rebase from master to trigger CI?

Roddoric:
I think it did not work 😅
image

jarednorman (approved):

waiting-for-dev (approved):
Thanks! ❤️

PR #4732, opened 2022-11-22
[v3.2] Fix the dummy app usage of the generator

waiting-for-dev:
## Summary

Backports part of #4646

We already shell-out for setting the environment and running migrations, this way we can avoid requiring the generator and we run it in a standard app environment instead of from within a library.

Before this change it was failing to set Rails.root as the app_path for the generator.

## Checklist

### DIFF
--- a/core/lib/spree/testing_support/common_rake.rb
+++ b/core/lib/spree/testing_support/common_rake.rb
@@ -1,9 +1,5 @@
 # frozen_string_literal: true
 
-unless defined?(Solidus::InstallGenerator)
-  require 'generators/solidus/install/install_generator'
-end
-
 require 'generators/spree/dummy/dummy_generator'
 
 class CommonRakeTasks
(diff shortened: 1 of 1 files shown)


core/lib/spree/testing_support/common_rake.rb
```diff
@@ -17,8 +13,31 @@ def initialize
...
+        Spree::DummyGenerator.start [
+          "--lib-name=#{ENV['LIB_NAME']}",
+          "--quiet",
+        ]
+
+        # While the dummy app is generated the current directory
+        # within ruby is changed to that of the dummy app.
+        sh({
+          'SKIP_SOLIDUS_BOLT' => '1',
+          'FRONTEND' => 'solidus_frontend',
```
gsmendoza, line 25: The [master version](https://github.com/nebulab/solidus/blob/29ddabbee282369daf990743dad96d2ce170ac0b/core/lib/spree/testing_support/common_rake.rb#L28) sets `FRONTEND` to `ENV['FRONTEND']` if that is available. Shouldn't we do that here alos? The [SSF 3.2 patch](https://github.com/solidusio/solidus_starter_frontend/blob/09dc12a4d8df600e9002586297f5e1408e964a2e/solidus_3_2_patch.rb#L17) accepts the `FRONTEND` environment variable.

elia: Good point! Although it's probably kinda optional as we confirmed that is not actually feasible to build this kind of dummy app with SSF (it breaks due to modifications to the Gemfile and other changes).

So we might even revert that change on master (maybe eventually deprecate the whole spec/dummy approach or externalize it.

elia (approved):
Feels a bit like self-approving but it's a 👍 anyways 😅

kennyadsl (approved):
Thanks

spaghetticode (approved):

core/lib/spree/testing_support/common_rake.rb
gsmendoza: Okay. Thanks for the explanation @elia !

PR #4733, opened 2022-11-22
Improve variant and product autocomplete functions flexibility with Ransack

RyanofWoods:
## Summary
From this change, when initializing a productAutocomplete, a callback can be given to the searchCallback option. This callback should return a hash object that will be used to merge onto the existing parameters in the select2 ajax function. The callback also allows for one argument, which is the string typed in the select2 (term). Custom Ransack can be achieved by defining this option with the Ransack query on the `q` key.

```JavaScript
$('#product-dropdown').productAutocomplete({
  multiple: false,
  searchCallback: (searchString) => ({
    q: {
      name_cont: searchString,
      available_on_not_null: true
    })
  }
})
```

Similar behavior was already existing on the `variantAutocomplete`, however accepted a simple `hash` object instead of a function, and so one didn't have access to the select2 search term, so this was improved.

```JavaScript
$('#variant-dropdown).variantAutocomplete({
  searchCallback: (searchString) => ({
    q: {
      variant_search_term: null,
      sku_cont: searchString
    })
  }
})
```

## Checklist

----
Autocompletes are still working:

### Orders#index
Screenshot 2022-11-28 at 14 10 33

### Orders#edit
Screenshot 2022-11-28 at 14 10 45

### Promotion Rules - Product
Screenshot 2022-11-28 at 14 12 10

### Promotion Rules - Option Values
Screenshot 2022-11-28 at 14 12 47

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/product_picker.js
+++ b/backend/app/assets/javascripts/spree/backend/product_picker.js
@@ -1,9 +1,32 @@
+/**
+  * Make the element a select2 dropdown used for finding Products. By default,
+  * it allows the Products to be found by its name and its Variants' SKUs.
+  * @param  {Object} options Options
+  * @param  {Boolean} [options.multiple=true] Allow multiple products to be selectable
+  * @param  {Function|undefined} options.searchCallback Returns a hash object for params to merge on the select2 ajax request
+  *                                                     Accepts an argument of the select2 search term. To use custom Ransack
+  *                                                     define q on the hash and add your custom terms. Note, you need to
+  *                                                     ensure that the attributes are allowed to be Ransacked.
+  */
 $.fn.productAutocomplete = function (options) {
   'use strict';
 
   // Default options
   options = options || {}
   var multiple = typeof(options['multiple']) !== 'undefined' ? options['multiple'] : true
+  function extraParameters(term) {
+    if (typeof(options['searchCallback']) === 'function') {
+      return options['searchCallback'](term)
+    }
+
+    return {
+      q: {
+        name_cont: term,
+        variants_including_master_sku_start: term,
+        m: 'or'
+      }
+    }
+  }
 
   function formatProduct(product) {
     return Select2.util.escapeMarkup(product.name);
@@ -26,15 +49,11 @@ $.fn.productAutocomplete = function (options) {
       datatype: 'json',
       params: { "headers": {  'Authorization': 'Bearer ' + Spree.api_key } },
       data: function (term, page) {
-        return {
-          q: {
-            name_cont: term,
-            variants_including_master_sku_start: term,
-            m: 'or'
-          },
+        const params = {
           token: Spree.api_key,
           page: page
         };
+        return _.extend(params, extraParameters(term));
       },
       results: function (data, page) {
         var products = data.products ? data.products : [];
--- a/api/spec/requests/spree/api/variants_spec.rb
+++ b/api/spec/requests/spree/api/variants_spec.rb
@@ -47,7 +47,7 @@ module Spree::Api
 
       it 'can query the results through ransack' do
         expected_result = create(:variant, sku: 'FOOBAR')
-        get spree.api_variants_path, params: { q: { sku_cont: 'FOO' } }
+        get spree.api_variants_path, params: { variant_search_term: nil, q: { sku_cont: 'FOO' } }
         expect(json_response['count']).to eq(1)
         expect(json_response['variants'].first['sku']).to eq expected_result.sku
       end
--- a/backend/app/assets/javascripts/spree/backend/orders/edit.js
+++ b/backend/app/assets/javascripts/spree/backend/orders/edit.js
@@ -1,6 +1,8 @@
 Spree.ready(function () {
   'use strict';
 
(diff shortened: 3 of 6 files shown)


RyanofWoods:
I added an example to the description @kennyadsl 👍  I also changed the option variable name to use lower camelcase
`ransack_q_callback` -> `ransackQCallback`

backend/app/assets/javascripts/spree/backend/product_picker.js
```diff
@@ -4,6 +4,15 @@ $.fn.productAutocomplete = function (options) {
   // Default options
   options = options || {}
   var multiple = typeof(options['multiple']) !== 'undefined' ? options['multiple'] : true
+  function ransackQuery(term) {
+    if (typeof(options['ransackQCallback']) === 'function') return options['ransackQCallback'](term)
```
tvdeyen, line 8: I really like that the `term` is passed into the function ( a bug we currently have in the `variantAutocomplete` actually)

But could we use a more approachable name for the option and maybe use the same as in the `variantAutocomplete`? https://github.com/solidusio/solidus/blob/master/backend/app/assets/javascripts/spree/backend/variant_autocomplete.js#L11

```suggestion
    if (typeof(options['searchOptions']) === 'function') return options['searchOptions'](term)
```

**Extra bonus points** for fixing the `variantAutocomplete` in this PR as well ;)

tvdeyen (commented):
Loving that! Thanks. I have some doubts about the name and it would be nice to fix the `variantAutocomplete` as well :)

RyanofWoods:
> Thanks, @RyanofWoods!! ❤️ A couple of questions/suggestions:
> 
> * Can you point us to the point where it's sent to the database? I'm worried about possible XSS attacks.
> * Would it be possible to add a test for it?
> * In the boy scout spirit, adding a few docs on top of the file would be nice.

Thanks for your feedback @waiting-for-dev ❤️ 

I will happily add tests and docs for it after discussing some more 👍 

Here's the endpoint it touches:
https://github.com/solidusio/solidus/blob/d582f545fb51eeb1c4b4cf318009c0a7f28db774/backend/app/controllers/spree/admin/search_controller.rb#L29

Security of this is a great point, this change would make it highly dependant on how well `.ransack` protects attacks. We could perhaps add a configuration such as `allowed_product_autocomplete_rancsack_matchers`, and then do param filters in the controller, WDYT?

RyanofWoods:
On further inspection, if there would be any security issues with allowing full control of parameters to `ransack`, we would have this problem right now:
https://github.com/solidusio/solidus/blob/d582f545fb51eeb1c4b4cf318009c0a7f28db774/api/app/controllers/spree/api/variants_controller.rb#L42

waiting-for-dev:
> Security of this is a great point, this change would make it highly dependant on how well `.ransack` protects attacks

Ok, it all boils down to the regular usage of Ransack, so we should be fine.

backend/app/assets/javascripts/spree/backend/product_picker.js
RyanofWoods: I think I might move the variant work to another PR @tvdeyen. Also, I am wondering if/how we should change this `variantComplete`. Technically, it one can currently initialize it to pass to the `ransack` method directly, but doesn't have access to the `select2` search term. Also, if doesn't go down this route, it will end up using a search variant class (`Spree::Core::Search::Variant` by default) which does has a lot of search logic already.

Having all these different paths is adding a lot of complexity. If we start accepting a function in the JS instead of a hash, having the search variant class route almost becomes redundant 🤔 
https://github.com/solidusio/solidus/blob/d582f545fb51eeb1c4b4cf318009c0a7f28db774/api/app/controllers/spree/api/variants_controller.rb#L35
https://github.com/solidusio/solidus/blob/d582f545fb51eeb1c4b4cf318009c0a7f28db774/core/lib/spree/core/search/variant.rb

RyanofWoods:
Specs are not getting actioned here @waiting-for-dev 🤔

waiting-for-dev:
> Specs are not getting actioned here @waiting-for-dev 🤔

Can you `git commit --amend --allow-empty` & `git push -f`? I've seen this kind of hiccup in the past 🤷

RyanofWoods:
Done this twice now, but same thing @waiting-for-dev 🤷

waiting-for-dev:
> Done this twice now, but same thing @waiting-for-dev 🤷

For some reason, you are unauthorized to run the build. Let's investigate that further. For now, I re-run CI and we can see the results. BUT, same that master, it's failing because the image suddenly can't find the libvips package 🤷

waiting-for-dev:
Great, @RyanofWoods, you're no longer unauthorized! 🥳 Just rebase from master to fix the CI errors.

RyanofWoods:
Nope, happened again @waiting-for-dev, wondering it's something to do with force pushing or that I am pushing to my own fork 🤔

RyanofWoods:
I couldn't get the CircleCI tests to trigger 🤷 So I made a new PR via the Nebulab fork which did the trick.
https://github.com/solidusio/solidus/pull/4767

@tvdeyen @waiting-for-dev

PR #4748, opened 2022-11-29
Don't install solidus_auth_devise if it's already in the Gemfile

### LINKED ISSUE #4766: Fix: Ensure that SSF 3.2 doesn't add solidus_auth_devise to the gemfile twice
Goal
----

Given I have a Rails app

When I add Solidus 3.2.4 to the Gemfile

And I install the default installation of Solidus

Then I should see only one `solidus_auth_devise` line in the Gemfile.

Current behavior
----------------

The Gemfile has two line for `solidus_auth_devise` :

```
# FIXME: Please remove this line if `solidus_auth_devise` appears anywhere else in the gemfile
#        or replace it with a simple `gem 'solidus_auth_devise'` otherwise.
gem 'solidus_auth_devise' unless File.read(__FILE__).lines[__LINE__..-1].grep(/solidus_auth_devise/).any?

gem "solidus_auth_devise"
```

Cause
-----

This is a known workaround. From <https://github.com/solidusio/solidus_starter_frontend/pull/277/commits/b9e8a7fa04f05718b5953619d4a3b377f0e017e3>:

> ```
> Adding solidus_auth_devise to the Gemfile was conflicting with it
> being added by the solidus installer again unless
> `--with-authentication=false` was provided. But the auth extension
> is also required by SSF. This forced a workaround in which we skip
> the `gem` entry in the Gemfile if another entry for the same gem is
> present in any of the following lines. It's ugly but puts a bandaid
> that will at least allow the installation to succeed.
> ```

Possible solutions
------------------

-   Revert <https://github.com/solidusio/solidus/commit/3b7ec0afa13b131cd1e4b2dc6c7516fd36e1ed78>. Leave SSF v3.2 without the responsibility to install solidus_auth_devise so we can remove the check on the Gemfile.

-   Alternatively, update the installer on v3.2 to try to install solidus_auth_devise after the frontend.

    -   Done on master.

-   Check if auth devise is already installed.

    -   Done on master.

Once fixed, we can update the template as follows:

```
gem 'solidus_auth_devise' unless Bundler.locked_gems.dependencies['solidus_auth_devise']
bundle_command["install"]
generate 'solidus:auth:install'
```

gsmendoza:
Fixes #4766.

Goal
----

Given I have a Rails app
When I add Solidus 3.2.4 to the Gemfile
And I install Solidus with Solidus Starter Frontend and Solidus Auth Devise
Then I should see only one `solidus_auth_devise` line in the Gemfile.

Current behavior
----------------

The Gemfile has two lines for `solidus_auth_devise` :

```ruby
# FIXME: Please remove this line if `solidus_auth_devise` appears anywhere else in the gemfile
#        or replace it with a simple `gem 'solidus_auth_devise'` otherwise.
gem 'solidus_auth_devise' unless File.read(__FILE__).lines[__LINE__..-1].grep(/solidus_auth_devise/).any?

gem "solidus_auth_devise"
```

Cause
-----

This is a workaround to an issue where `Solidus#InstallGenerator` and SolidusStarterFrontend both add `solidus_auth_devise` to the Gemfile. From <https://github.com/solidusio/solidus_starter_frontend/pull/277/commits/b9e8a7fa04f05718b5953619d4a3b377f0e017e3>:

> ```
> Adding solidus_auth_devise to the Gemfile was conflicting with it
> being added by the solidus installer again unless
> `--with-authentication=false` was provided. But the auth extension
> is also required by SSF. This forced a workaround in which we skip
> the `gem` entry in the Gemfile if another entry for the same gem is
> present in any of the following lines. It's ugly but puts a bandaid
> that will at least allow the installation to succeed.
> ```

## Testing

I have confirmed manually that, in conjunction with the [SolidusStarterFrontend update](https://github.com/solidusio/solidus_starter_frontend/commits/gsmendoza/sol-547-fix-ensure-that-ssf-32-doesnt-add), installing Solidus in v3.2 now installs `solidus_auth_devise` only once. See

* https://github.com/solidusio/solidus_starter_frontend/commits/gsmendoza/sol-547-fix-ensure-that-ssf-32-doesnt-add
  - This update to SolidusStarterFrontend removes the SolidusAuthDevise workaround and FIXME comment.
* https://github.com/nebulab/solidus/tree/gsmendoza/sol-547-fix-ensure-that-ssf-32-doesnt-add--with-test-ssf-branch
  - A branch of this PR that ties InstallGenerator to a branch of SolidusStarterFrontend without the workaround.
* https://github.com/gsmendoza/solidus_store/tree/rails-7-0-4--solidus-3-2-4--sol-547-fix-ensure-that-ssf-32-doesnt-add
  - The app generated by the Solidus installation I did to test this fix.
* [solidus-install-log.txt](https://github.com/solidusio/solidus/files/10111919/solidus-install-log.txt)
  - The log of the Solidus installation. Note the "skipping  solidus_auth_devise is already in the gemfile" line.

## Checklist

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -184,6 +184,12 @@ module Solidus
         InstallFrontend
           .new(bundler_context: bundler_context, generator_context: self)
           .call(frontend)
+
+        # The DEFAULT_FRONTEND installation makes changes to the
+        # bundle without updating the bundler context. As such, we need to
+        # reset the bundler context to get the latest dependencies from the
+        # context.
+        reset_bundler_context if frontend == DEFAULT_FRONTEND
       end
     end
 
@@ -290,6 +296,11 @@ module Solidus
     end
 
     def install_plugin(plugin_name:, plugin_generator_name:)
+      if bundler_context.dependencies[plugin_name]
+        say_status :skipping, "#{plugin_name} is already in the gemfile"
+        return
+      end
+
       gem plugin_name
       BundlerContext.bundle_cleanly { run "bundle install" }
       run "spring stop" if defined?(Spring)
@@ -298,6 +309,10 @@ module Solidus
 
     def bundler_context
       @bundler_context ||= BundlerContext.new
+    end
+
+    def reset_bundler_context
+      @bundler_context = nil
     end
   end
 end
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -187,14 +187,12 @@ module Solidus
     end
 
     def install_frontend
-      bundler_context = BundlerContext.new
-
       if options[:frontend] == 'none'
-        support_solidus_frontend_extraction(bundler_context)
+        support_solidus_frontend_extraction
       else
-        frontend = detect_frontend_to_install(bundler_context)
+        frontend = detect_frontend_to_install
 
-        support_solidus_frontend_extraction(bundler_context) unless frontend == LEGACY_FRONTEND
+        support_solidus_frontend_extraction unless frontend == LEGACY_FRONTEND
 
         say_status :installing, frontend
 
@@ -285,7 +283,7 @@ module Solidus
 
     private
 
-    def detect_frontend_to_install(bundler_context)
+    def detect_frontend_to_install
       ENV['FRONTEND'] ||
         options[:frontend] ||
         (bundler_context.component_in_gemfile?(:frontend) && LEGACY_FRONTEND) ||
@@ -299,12 +297,16 @@ module Solidus
         MSG
     end
 
-    def support_solidus_frontend_extraction(bundler_context)
+    def support_solidus_frontend_extraction
       say_status "break down", "solidus"
 
       SupportSolidusFrontendExtraction.
         new(bundler_context: bundler_context).
         call
+    end
+
+    def bundler_context
+      @bundler_context ||= BundlerContext.new
     end
   end
 end
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -126,14 +126,9 @@ module Solidus
       end
     end
 
-    def plugin_install_preparation
-      @plugins_to_be_installed = []
-      @plugin_generators_to_run = []
-    end
-
-    def install_auth_plugin
-      with_authentication = options[:with_authentication]
-      with_authentication.nil? and with_authentication = (options[:auto_accept] || !no?("
+    def select_auth_plugin
+      @with_authentication = options[:with_authentication]
+      @with_authentication.nil? and @with_authentication = (options[:auto_accept] || !no?("
         Solidus has a default authentication extension that uses Devise.
         You can find more info at https://github.com/solidusio/solidus_auth_devise.
 
@@ -141,14 +136,9 @@ module Solidus
         solidus_starter_frontend as your storefront in a later step.
 
         Would you like to install it? (Y/n)"))
-
-      if with_authentication
-        @plugins_to_be_installed << 'solidus_auth_devise'
-        @plugin_generators_to_run << 'solidus:auth:install'
-      end
     end
 
-    def install_payment_method
+    def select_payment_method
       say_status :warning, set_color(
         "Selecting a payment along with `solidus_starter_frontend` might require manual integration.",
         :yellow
@@ -160,12 +150,7 @@ module Solidus
   You can select a payment method to be included in the installation process.
   Please select a payment method name:", limited_to: PAYMENT_METHODS.keys, default: PAYMENT_METHODS.keys.first)
 
-      gem_name = PAYMENT_METHODS.fetch(name)
-
-      if gem_name
-        @plugins_to_be_installed << gem_name
-        @plugin_generators_to_run << "#{gem_name}:install"
-      end
+      @payment_method_gem_name = PAYMENT_METHODS.fetch(name)
     end
 
     def include_seed_data
@@ -202,17 +187,16 @@ module Solidus
       end
     end
 
-    def run_bundle_install_if_needed_by_plugins
-      @plugins_to_be_installed.each do |plugin_name|
-        gem plugin_name
-      end
+    def install_authentication_plugin
+      return unless @with_authentication
 
-      BundlerContext.bundle_cleanly { run "bundle install" } if @plugins_to_be_installed.any?
-      run "spring stop" if defined?(Spring)
+      install_plugin(plugin_name: 'solidus_auth_devise', plugin_generator_name: 'solidus:auth:install')
+    end
 
-      @plugin_generators_to_run.each do |plugin_generator_name|
-        generate "#{plugin_generator_name} --skip_migrations=true"
-      end
+    def install_payment_method_plugin
+      return unless @payment_method_gem_name
+
+      install_plugin(plugin_name: @payment_method_gem_name, plugin_generator_name: "#{@payment_method_gem_name}:install")
     end
 
     def run_migrations
@@ -303,6 +287,13 @@ module Solidus
       SupportSolidusFrontendExtraction.
         new(bundler_context: bundler_context).
         call
+    end
+
+    def install_plugin(plugin_name:, plugin_generator_name:)
+      gem plugin_name
+      BundlerContext.bundle_cleanly { run "bundle install" }
+      run "spring stop" if defined?(Spring)
+      generate "#{plugin_generator_name} --skip_migrations=true"
     end
 
     def bundler_context
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -302,7 +302,7 @@ module Solidus
       end
 
       gem plugin_name
-      BundlerContext.bundle_cleanly { run "bundle install" }
+      run_bundle
       run "spring stop" if defined?(Spring)
       generate "#{plugin_generator_name} --skip_migrations=true"
     end
--- a/core/lib/generators/solidus/install/install_generator/bundler_context.rb
+++ b/core/lib/generators/solidus/install/install_generator/bundler_context.rb
@@ -49,7 +49,7 @@ module Solidus
       end
 
       def initialize
-        @dependencies = Bundler.locked_gems.dependencies
+        @dependencies = uncached_dependencies
         @injector = InjectorWithPathSupport
       end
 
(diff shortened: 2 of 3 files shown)


core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -187,34 +172,37 @@ def create_database
...
 
         InstallFrontend
           .new(bundler_context: bundler_context, generator_context: self)
           .call(frontend)
+
+        # The DEFAULT_FRONTEND installation makes changes to the
+        # bundle without updating the bundler context. As such, we need to
+        # reset the bundler context to get the latest dependencies from the
+        # context.
+        reset_bundler_context if frontend == DEFAULT_FRONTEND
```
gsmendoza, line 192: From the comment:

```
        # The DEFAULT_FRONTEND installation makes changes to the
        # bundle without updating the bundler context. As such, we need to
        # reset the bundler context to get the latest dependencies from the
        # context.
```

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -299,12 +287,32 @@ def detect_frontend_to_install(bundler_context)
...
     end
+
+    def install_plugin(plugin_name:, plugin_generator_name:)
+      if bundler_context.dependencies[plugin_name]
+        say_status :skipping, "#{plugin_name} is already in the gemfile"
+        return
+      end
+
+      gem plugin_name
+      BundlerContext.bundle_cleanly { run "bundle install" }
```
gsmendoza, line 305: As I wrote in the commit message:

>   * Installing the plugin sequentially would call `bundle install` twice,
      but this is okay since InstallGenerator in master also adds gems to the
      bundle one at a time.

core/lib/generators/solidus/install/install_generator/bundler_context.rb
```diff
@@ -49,7 +49,11 @@ def self.bundle_cleanly(&block)
       end
 
       def initialize
-        @dependencies = Bundler.locked_gems.dependencies
+        # We want to get the locked gems from `#definition` instead of
+        # `Bundler.locked_gems` so that the locked gems won't be memoized. See
+        # https://rubydoc.info/gems/bundler/2.3.22/Bundler.locked_gems.
+        @dependencies = Bundler.definition.locked_gems.dependencies
```
gsmendoza, line 55: Highlighting the comment about calling `locked_gems` from `Bundler.definition`:

```
        # We want to get the locked gems from `#definition` instead of
        # `Bundler.locked_gems` so that the locked gems won't be memoized. See
        # https://rubydoc.info/gems/bundler/2.3.22/Bundler.locked_gems.
```

core/lib/generators/solidus/install/install_generator.rb
elia: nit: since it's now using `AbbBase` as the superclass I think the whole line can be replaced with `run_bundle`

elia (approved):
Left an optional nit, fine by me if we merge without changing anything 🙌

core/lib/generators/solidus/install/install_generator.rb
waiting-for-dev: I guess then we could remove the whole `BundlerContext.bundle_cleanly` method, right?

waiting-for-dev (commented):
Leaving some comments to understand everything, but it looks great ❤️

core/lib/generators/solidus/install/install_generator.rb
gsmendoza: @waiting-for-dev It's needed because BundlerContext stores the dependencies in an instance variable. If we want to get the latest dependencies, we'd have to create a new BundlerContext.

One alternative is to change `BundlerContext#dependencies` so that it always returns the latest dependencies:

```ruby
class BundlerContext
  def dependencies
    Bundler.definition.locked_gems.dependencies
  end 
end
```

However, I'm guessing BundlerContext stores dependencies in an instance variable for performance reasons i.e. caching.

Your thoughts?

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -202,17 +187,16 @@ def install_frontend
...
+      install_plugin(plugin_name: 'solidus_auth_devise', plugin_generator_name: 'solidus:auth:install')
+    end
 
-      @plugin_generators_to_run.each do |plugin_generator_name|
-        generate "#{plugin_generator_name} --skip_migrations=true"
-      end
+    def install_payment_method_plugin
+      return unless @payment_method_gem_name
+
+      install_plugin(plugin_name: @payment_method_gem_name, plugin_generator_name: "#{@payment_method_gem_name}:install")
```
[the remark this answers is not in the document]
gsmendoza: Acknowledged :) I would prefer to keep the current implementation, since it's good enough for now. Giving the generator name  would probably require storing it in the `PAYMENT_METHODS`. It's not a complicated change, but at the same time, it's still a change that we can postpone making.

core/lib/generators/solidus/install/install_generator.rb
gsmendoza: I'll try changing that call to `run_bundle` and removing `bundle_cleanly` :+1:

core/lib/generators/solidus/install/install_generator/install_frontend.rb
```diff
@@ -24,22 +24,11 @@ def call(frontend)
...
-          )
-        end
-      end
-
       def install_solidus_frontend
         unless @bundler_context.component_in_gemfile?(:frontend)
-          bundle_command 'add solidus_frontend'
+          # `Rails::Generator::AppBase#bundle_command` is protected so have to `send` it.
+          # See https://api.rubyonrails.org/v3.2.16/classes/Rails/Generators/AppBase.html#method-i-run_bundle
+          @generator_context.send :bundle_command, 'add solidus_frontend'
```
gsmendoza, line 31: @elia @waiting-for-dev Any concerns with using `#send` here?

core/lib/generators/solidus/install/install_generator.rb
gsmendoza: I was able to remove `bundle_cleanly` but I had to call `bundle_command` with `#send` in InstallFrontend. Please see https://github.com/solidusio/solidus/pull/4748#discussion_r1036860556. Thanks!

core/lib/generators/solidus/install/install_generator.rb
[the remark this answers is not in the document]
waiting-for-dev: Agree 🙂

core/lib/generators/solidus/install/install_generator/install_frontend.rb
waiting-for-dev: As we own the `@generator_context`, we can wrap `bundle_command` in another method and call it from `InstallFrontend`. However, it adds up to my fears that we're leveraging private API by relying on `AppBase` 😨

core/lib/generators/solidus/install/install_generator.rb
waiting-for-dev: Ah, I see. Then, what isn't needed is the change in https://github.com/solidusio/solidus/pull/4748/files#r1034556084, right? (I'd still memorize and invalidate as it probably makes sense for performance reasons)

gsmendoza: We still need https://github.com/solidusio/solidus/pull/4748/files#r1034556084 (changing `Bundler.locked_gems` to `Bundler.definition.locked_gems`) since `Bundler.locked_gems` also memoizes the locked gems (see https://rubydoc.info/gems/bundler/2.3.22/Bundler.locked_gems). If we kept `Bundler.locked_gems`, a new instance of BundlerContext would load the memoized locked gems from Bundler, with its corresponding dependencies.

waiting-for-dev: Got it. Thanks for the explanation, @gsmendoza.

core/lib/generators/solidus/install/install_generator/install_frontend.rb
gsmendoza: @waiting-for-dev I'm not sure if I understood you correctly, but if I try to add a `public_bundle_command` method to InstallGenerator, it causes `solidus:install` to fail:

```ruby
module Solidus
  class InstallGenerator < Rails::Generators::AppBase
    def public_bundle_command(command)
      bundle_command(command)
    end
  end
end
```

```sh
solidus 14:19:05 $ bin/sandbox --frontend=solidus_frontend
# ...
/home/gsmendoza/repos/solidusio/solidus/core/lib/generators/solidus/install/install_generator.rb:61:in `public_bundle_command': rails public_bundle_command takes 1 argument, but it should not. (ArgumentError)
```

I think this is because `solidus:install` attempts to run all of the public methods in InstallGenerator, including the `public_bundle_command` helper.

It doesn't seem like there's a way to avoid using `#send` here. Do you have any other ideas?

waiting-for-dev: Oh, too bad 😢 Let's go with `send` or revert to what we had before—your choice.

gsmendoza: @waiting-for-dev Let's go with `send`. We're already using AppBase anyway hehehe...

gsmendoza:
@waiting-for-dev @solidusio/core-team I don't have any more updates to this PR. Can you take a look again? Thank you!

core/lib/generators/solidus/install/install_generator/bundler_context.rb
waiting-for-dev: @kennyadsl suggested wrapping that in a `uncached_dependencies` method and moving the comment into the commit message.

waiting-for-dev (approved):

core/lib/generators/solidus/install/install_generator/bundler_context.rb
gsmendoza: @kennyadsl @waiting-for-dev I'm not sure about the desired outcome for `uncached_dependencies`. Is our goal to extract `uncached_dependencies` in a private method, like this?

```ruby
module Solidus
  class InstallGenerator < Rails::Generators::AppBase
    def initialize
        @dependencies = uncached_dependencies
        @injector = InjectorWithPathSupport
    end

    private

    # We want to get the locked gems from `#definition` instead of
    # `Bundler.locked_gems` so that the locked gems won't be memoized. See
    # https://rubydoc.info/gems/bundler/2.3.22/Bundler.locked_gems.
    def uncached_dependencies
      Bundler.definition.locked_gems.dependencies
    end
  end
end
```

For this case, I would still prefer to keep the comment inline with the method since it's pertains specifically to the method.

waiting-for-dev: Yeah, I think that's what @kennyadsl suggested. I don't have a strong opinion about the comment 🙂

gsmendoza:
@waiting-for-dev I extracted the method to `uncached_dependencies`. Can you take one more look? Thank you!

waiting-for-dev (approved):

PR #4754, opened 2022-11-30
Remove position column from spree_taxons

mamhoff:
## Summary

This column has not been used since 8d7d4d7, which had been merged into Spree 1.1.0 in 2013.

Background: Apparently, taxons used to be a list in very early versions of Spree, and have then been refactored to be a nested set. Nested sets are strict supersets of lists, so this column is entirely unnecessary.

## Checklist

- [ ] ~I have added automated tests to cover my changes.~
- [ ] ~I have updated the README to account for my changes.~


core/db/migrate/20221130103509_remove_taxon_position.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveTaxonPosition < ActiveRecord::Migration[7.0]
```
tvdeyen, line 1: We still need to support old Rails versions, as we are [still depending on rails >=5.2](https://github.com/solidusio/solidus/blob/master/core/solidus_core.gemspec#L30)

```suggestion
class RemoveTaxonPosition < ActiveRecord::Migration[ActiveRecord::Migration.current_version]
```

tvdeyen (commented):
Thanks 🙏🏻 

Unfortunately we cannot assume Rails 7.0

core/db/migrate/20221130103509_remove_taxon_position.rb
waiting-for-dev: To ensure syntax compatibility, we're creating migrations always with the oldest version in mind:

https://github.com/solidusio/solidus/blob/efe2b3a20d0577413e4810c008ea564e637c36a3/core/db/migrate/20220805202442_add_level_to_spree_tax_rates.rb#L1

core/db/migrate/20221130103509_remove_taxon_position.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveTaxonPosition < ActiveRecord::Migration[ActiveRecord::Migration.current_version]
```
waiting-for-dev, line 1: To ensure syntax compatibility, we're creating migrations always with the oldest version in mind:

https://github.com/solidusio/solidus/blob/efe2b3a20d0577413e4810c008ea564e637c36a3/core/db/migrate/20220805202442_add_level_to_spree_tax_rates.rb#L1

waiting-for-dev (changes requested):
Thanks, @mamhoff! To ensure compatibility, we need to rely on the oldest Rails version for the database file, though. In this case, it won't make a difference, but it'll be consistent with what we've done so far. We can discuss and change in separately for all the migration files.

waiting-for-dev:
Also, please, rebase from master to see if the CircleCI issues are gone.

waiting-for-dev (approved):
Thanks!!

jarednorman (approved):

waiting-for-dev:
Rebasing to launch CI

mamhoff:
Rebased from `master`; added a commit to remove the `position` attribute from the taxon sample file.

PR #4767, opened 2022-12-05
Improve variant and product autocomplete functions flexibility with Ransack

RyanofWoods:
## Summary
From this change, when initializing a productAutocomplete, a callback can be given to the searchCallback option. This callback should return a hash object that will be used to merge onto the existing parameters in the select2 ajax function. The callback also allows for one argument, which is the string typed in the select2 (term). Custom Ransack can be achieved by defining this option with the Ransack query on the `q` key.

```JavaScript
$('#product-dropdown').productAutocomplete({
  multiple: false,
  searchCallback: (searchString) => ({
    q: {
      name_cont: searchString,
      available_on_not_null: true
    })
  }
})
```

Similar behavior was already existing on the `variantAutocomplete`, however accepted a simple `hash` object instead of a function, and so one didn't have access to the select2 search term, so this was improved.

```JavaScript
$('#variant-dropdown).variantAutocomplete({
  searchCallback: (searchString) => ({
    q: {
      variant_search_term: null,
      sku_cont: searchString
    })
  }
})
```

## Checklist

----
Autocompletes are still working:

### Orders#index
Screenshot 2022-11-28 at 14 10 33

### Orders#edit
Screenshot 2022-11-28 at 14 10 45

### Promotion Rules - Product
Screenshot 2022-11-28 at 14 12 10

### Promotion Rules - Option Values
Screenshot 2022-11-28 at 14 12 47

### DIFF
--- a/backend/app/assets/javascripts/spree/backend/product_picker.js
+++ b/backend/app/assets/javascripts/spree/backend/product_picker.js
@@ -1,9 +1,32 @@
+/**
+  * Make the element a select2 dropdown used for finding Products. By default,
+  * it allows the Products to be found by its name and its Variants' SKUs.
+  * @param  {Object} options Options
+  * @param  {Boolean} [options.multiple=true] Allow multiple products to be selectable
+  * @param  {Function|undefined} options.searchParameters Returns a hash object for params to merge on the select2 ajax request
+  *                                                       Accepts an argument of the select2 search term. To use custom Ransack
+  *                                                       define q on the hash and add your custom terms. Note, you need to
+  *                                                       ensure that the attributes are allowed to be Ransacked.
+  */
 $.fn.productAutocomplete = function (options) {
   'use strict';
 
   // Default options
   options = options || {}
   var multiple = typeof(options['multiple']) !== 'undefined' ? options['multiple'] : true
+  function extraParameters(term) {
+    if (typeof(options['searchParameters']) === 'function') {
+      return options['searchParameters'](term)
+    }
+
+    return {
+      q: {
+        name_cont: term,
+        variants_including_master_sku_start: term,
+        m: 'or'
+      }
+    }
+  }
 
   function formatProduct(product) {
     return Select2.util.escapeMarkup(product.name);
@@ -26,15 +49,11 @@ $.fn.productAutocomplete = function (options) {
       datatype: 'json',
       params: { "headers": {  'Authorization': 'Bearer ' + Spree.api_key } },
       data: function (term, page) {
-        return {
-          q: {
-            name_cont: term,
-            variants_including_master_sku_start: term,
-            m: 'or'
-          },
+        const params = {
           token: Spree.api_key,
           page: page
         };
+        return _.extend(params, extraParameters(term));
       },
       results: function (data, page) {
         var products = data.products ? data.products : [];
--- a/backend/app/assets/javascripts/spree/backend/variant_autocomplete.js
+++ b/backend/app/assets/javascripts/spree/backend/variant_autocomplete.js
@@ -10,16 +10,30 @@
   /**
     * Make the element a select2 dropdown used for finding Variants. By default, the search term will be
     * passed to the defined Spree::Config.variant_search_class by the controller with its defined scope.
-    * @param  {Object} options Options
+    * @param  {Object|undefined|null} options Options
     * @param  {Function|undefined} options.searchParameters Returns a hash object for params to merge on the select2 ajax request
     *                                                       Accepts an argument of the select2 search term. To use Ransack, define
     *                                                       variant_search_term as a falsy value, and q as the Ransack query. Note,
     *                                                       you need to ensure that the attributes are allowed to be Ransacked.
     */
-  $.fn.variantAutocomplete = function(options = {}) {
+  $.fn.variantAutocomplete = function(options) {
     function extraParameters(term) {
-      if (typeof(options['searchParameters']) === 'function') {
-        return options['searchParameters'](term)
+      if (typeof(options) === 'object') {
+        if (typeof(options['searchParameters']) === 'function') {
+          return options['searchParameters'](term)
+        } else {
+          console.warn(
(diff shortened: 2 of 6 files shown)


waiting-for-dev (approved):
Awesome, @RyanofWoods!

RyanofWoods:
Note, the `variantAutocomplete` changes are breaking to any stores that have implemented it. They'll need to switch from `searchOptions` to `searchCallback` appropriately.

backend/app/assets/javascripts/spree/backend/product_picker.js
```diff
@@ -1,9 +1,32 @@
+/**
+  * Make the element a select2 dropdown used for finding Products. By default,
+  * it allows the Products to be found by its name and its Variants' SKUs.
+  * @param  {Object} options Options
+  * @param  {Boolean} [options.multiple=true] Allow multiple products to be selectable
+  * @param  {Function|undefined} options.searchCallback Returns a hash object for params to merge on the select2 ajax request
```
tvdeyen, line 6: Nice work on the documentation, this is 💯 

I am not very happy with the name, though. It is very generic and too technical imo. Sure, it is - technically - a callback, but it is not very descriptive about what it is used for or what it shall return. What about:

- `searchParameters` - My favorite
- `extraParameters` (would fit to the function it is used in)

tvdeyen (commented):
Nice work and thanks for also tackling the variantAutocomplete 👏🏻 

I still think we can improve the name of the option, to make it more intuitive.

backend/app/assets/javascripts/spree/backend/product_picker.js
RyanofWoods: Understood. I found the naming difficult. I do agree that `searchCallback` is not perfect, though to me `searchParameters` sounds like it should be a hash object rather than a function that should return a hash? 🤔 @tvdeyen

RyanofWoods: On further thoughts, the documentation does show `Function`. So I think the suggestions are acceptable. Note, I made it so the function's returning object is extended onto the request parameters, so it doesn't have to just be just the ransack `q` param. Interested on your thoughts @waiting-for-dev

RyanofWoods: `searchCallback` changed to `searchParameters`

RyanofWoods:
I pulled a spec change out into an new commit and changed other commits to rename `searchCallback` to `searchParameters`

tvdeyen (approved):
Thanks 🙏🏻

adammathys (approved):

kennyadsl (commented):
Hey @RyanofWoods! This is great but I have a request: what if we try to make this change backward compatible converting the previous input format for `variantAutocomplete` into the new format programmatically and print a deprecation warning so that people have the opportunity to change it?

RyanofWoods:
I can do that @kennyadsl, when you mean `print`, do you mean like `console.warn`? I believe we don't have a JS deprecation API in place or that we've done formal JS deprecations before 🤔

kennyadsl:
> when you mean `print`, do you mean like `console.warn`?

Yes, I think that the console warning, combined with an entry in our CHANGELOG would be enough to let people know. We do not have a specific API for that, despite we did [something similar](https://github.com/solidusio/solidus/commit/5d55e10c23fadfcdcb39436593c4fcc3c5a244b1#diff-a6e1aff0e25ccb5e822aa6c73f4c5b92dc2a5b411f2b7dcf6b87189d9d1c2068R61) in the past.

RyanofWoods:
Hmm, looks we're not adding manual changes to the changelog anymore @kennyadsl and @waiting-for-dev. So I removed them for now and left them below for reference.

---

- productAutocomplete's options parameter now accepts the `searchParameters` key, which allows access to the select2 search term and modification of the parameters sent to the API endpoint.

---
#### variantAutocomplete Changes
DEPRECATION:
- Passing an object of parameters is deprecated, instead, please declare the `searchParameters` key that's a function returning the parameters.

Deprecated usage:
```JavaScript
$('#id').variantAutocomplete({
  suppliable_only: true
})
```

New usage:
```JavaScript
$('#id').variantAutocomplete({
  searchParameters: function (_selectSearchTerm) { return { suppliable_only: true  } }
})
```

backend/app/assets/javascripts/spree/backend/variant_autocomplete.js
```diff
@@ -7,10 +7,28 @@
...
+    *                                                       variant_search_term as a falsy value, and q as the Ransack query. Note,
+    *                                                       you need to ensure that the attributes are allowed to be Ransacked.
+    */
+  $.fn.variantAutocomplete = function(options) {
+    function extraParameters(term) {
+      if (typeof(options) === 'object') {
+        if (typeof(options['searchParameters']) === 'function') {
+          return options['searchParameters'](term)
+        } else {
+          console.warn("Solidus deprecation: Passing an object of parameters to variantAutocomplete is deprecated. Instead, on the options object, please declare `searchParameters` as a function returning the parameters.")
```
kennyadsl, line 25: Nice, what about adding a usage example of the new syntax? I think it might help a lot.

kennyadsl (commented):
Thanks @RyanofWoods, left one more last suggestion.

backend/app/assets/javascripts/spree/backend/variant_autocomplete.js
RyanofWoods: Done @kennyadsl 
Screenshot 2022-12-22 at 10 08 31

kennyadsl (approved):
Nice! I removed the `release:major` label, we can merge this in a minor now. Thanks!

RyanofWoods:
Didn't want the squash commit squashed? 🤔

waiting-for-dev:
Oh, sorry, I missed that 🤦

RyanofWoods:
No worries! I could have said something @waiting-for-dev

PR #4802, opened 2022-12-22
Skip update of release notes at the job definition level

waiting-for-dev:
## Summary

We're already skipping adding entries for PRs merged with the `changelog:skip` label. However, we were doing it only from the [code](https://github.com/solidusio/solidus/blob/f8cb8517b6dae37a0e1a00eae72409ac0b54c37a/dev_tools/lib/solidus/release_drafter.rb#L37) called by the job definition. We're now optimizing by also checking it at the job definition level to skip the execution of previous slow steps (i.e., checking out the code).

## Checklist

### DIFF
--- a/.github/workflows/update_release_draft.yml
+++ b/.github/workflows/update_release_draft.yml
@@ -15,7 +15,9 @@ jobs:
   update:
     runs-on: ubuntu-latest
     needs: extract_pipeline_context
-    if: github.event.pull_request.merged == true
+    if: |
+      (github.event.pull_request.merged == true) &&
+        !contains(github.event.pull_request.labels.*.name, 'changelog:skip')
     steps:
       - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1


.github/workflows/update_release_draft.yml
```diff
@@ -15,7 +15,9 @@ jobs:
   update:
     runs-on: ubuntu-latest
     needs: extract_pipeline_context
-    if: github.event.pull_request.merged == true
+    if: |
+      (github.event.pull_request.merged == true) &&
+        !contains(github.event.pull_request.labels.*.name, 'changelog:skip')
```
kennyadsl, line 20: Will skipping this step at this point prevent the rest of the action from being executed?

kennyadsl (approved):
👍 just left a question, thanks!

.github/workflows/update_release_draft.yml
waiting-for-dev: It skips the `update` job, but not the previous `extract_pipeline_context` one. I'm tackling that separately, as I'd like to convert that job into a composite action to avoid checking out the code multiple times.

gsmendoza (approved):

PR #4805, opened 2022-12-22
Extract pipeline context into a composite action

### LINKED ISSUE #4798: Extract the pipeline context workflow into a composite action
At this point, the pipeline context extraction is run as a separate workflow. That means that we're checking out the code twice and making the automated tasks slower. We are neither reusing the if: conditions on the called workflows. We need to investigate if we can extract it into a composite action and call it from within a step.

waiting-for-dev:
## Summary

We move the [reused workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to extract the pipeline context to a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action). There're a few benefits of doing that:

- High win on performance, as we don't need to check out the code twice or wait in the queue.
- The dependent workflows only have one job now, so that we can skip all the workflows with a single `if:` declaration.
- More straightforward, as there're no job dependencies anymore.

## Checklist

Closes #4798


kennyadsl (approved):
Thanks @waiting-for-dev

elia (approved):

PR #4827, opened 2022-12-31
[v3.1] Add EN percentage number format

tvdeyen:
backport of #4824 into 3.1

PR #4829, opened 2023-01-02
[v3.1] Test with Ruby 2.7

tvdeyen:
## Summary

The latest Bundler version (2.4) that automatically installs on CI does not support Ruby 2.5 anymore, also Ruby 2.5 is not receiving any updates from the Ruby team anymore.

We still support using Ruby 2.5 with Solidus 3.1, but we do not test it with later versions. We should test with Ruby 2.7 instead.

## Checklist


tvdeyen:
See https://github.com/solidusio/solidus/pull/4827 for a failing build because of that.

jarednorman (approved):

PR #4834, opened 2023-01-03
Misc text cleanup

basicfeatures:


jarednorman:
The first commit appears to be adding blank lines, not newlines to the end of these files. We want newlines, but we don't need all these blank lines.

jarednorman (changes requested):
1. We don't need blank lines, just new lines.
2. Let's not apply any of these changes to vendored files.

basicfeatures:
Good thing you pointed it out. I sorta have it fixed already with a Ruby version of the sadly buggy script that did this. I'll get to it ASAP.

PR #4741, opened 2022-11-24
Ignore big reformatting commits in GitHub blame UI

kennyadsl:
## Summary

By adding a `.git-blame-ignore-revs` file that contains the commits that git blame will ignore, we are able to remove them from the blame view when using this feature.

See https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view for extra details.

From that link it's worth mentioning that:

>If the blame view for a file shows Ignoring revisions in .git-blame-ignore-revs, you can still bypass .git-blame-ignore-revs and see the normal blame view. In the URL, append a ~ to the SHA and the Ignoring revisions in .git-blame-ignore-revs will disappear.

For now, I added just a couple of big commits, but we can add more with time.

This same behavior can be enabled locally with:

	git config blame.ignoreRevsFile .git-blame-ignore-revs

This config is not suitable as a global config though, because it will fail on repositories that do not have the file set.

Other resources:

- https://blog.testdouble.com/posts/2022-11-21-rubocop-git-blame/

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/.github/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [x] I have written a thorough PR description.
- [x] I have kept my commits small and atomic.
- [x] I have used clear, explanatory commit messages.

The following are not always needed (~cross them out~ if they are not):

- [ ] ~I have added automated tests to cover my changes.~
- [ ] ~I have attached screenshots to demo visual changes.~
- [ ] ~I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).~
- [ ] ~I have updated the README to account for my changes.

PR #4837, opened 2023-01-05
Misc cleanup of all text files

basicfeatures:
- Remove CR+LF (^M)
- Remove trailing whitespace
- Compress two or more consecutive blank lines
- Ensure newline at EOF

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file


jarednorman (dismissed):

dev_tools/lib/solidus/release_drafter.rb
```diff
@@ -68,7 +68,6 @@ def save_release(draft, candidate_tag, branch)
     def full_changelog(current_diff_source_tag, candidate_tag)
       <<~TXT
 
-
```
kennyadsl, line 71: This extra space is intentional, see https://app.circleci.com/pipelines/github/solidusio/solidus/3952/workflows/a82c4d50-c4ee-426b-92b4-5cccee0ad4fe/jobs/36006. We can probably get rid of it, but the specs should reflect the change, cc @waiting-for-dev.

jarednorman: Ha, good catch.

waiting-for-dev (changes requested):
Thanks for this clean-up ❤️ 

Let's revert the changes in the `dev_tools` directory out for now, as all blank lines were intentionally there.

jarednorman (commented):
We should also make sure that this commit gets added to the commits that are ignored in blames.

waiting-for-dev:
> We should also make sure that this commit gets added to the commits that are ignored in blames.

See https://github.com/solidusio/solidus/pull/4741 for context

PR #4844, opened 2023-01-10
Raise a custom extension passing invalid search params

kennyadsl:
## Summary

Right now, passing a string (or any other thing that is not a Hash) to the searcher in the `:search` parameter will raise a `NoMethodError`.

For example: `/products?per_page=&keywords=&search=blub` would raise a NoMethodError.

This commit avoids that a generic NoMethodError is raised and instead raises a very specific exception that serves two purposes:

1. Helps debugging: the error message contains the name of the param that is invalid, making it easier to spot eventual issues.
2. Allows to rescue the error safely at the storefront level. For example, to respond with a 400, one can easily do:

       rescue_from Spree::Core::Search::Base::InvalidOptions do |error|
         raise ActionController::BadRequest.new, error.message
       end

### Backward compatibility impact

If a store was rescuing from NoMethodError trying to catch this extension, their rescuing will fail. There's nothing to do about that. 

--- 

🙏  This issue was originally reported in HackerOne by our friends at @renuo, but we agreed to treat this as a regular bug because of the low security impact this might have.

## Checklist

### DIFF
--- a/core/lib/spree/core/search/base.rb
+++ b/core/lib/spree/core/search/base.rb
@@ -4,6 +4,12 @@ module Spree
   module Core
     module Search
       class Base
+        class InvalidOptions < ArgumentError
+          def initialize(option)
+            super("Invalid option passed to the searcher: '#{option}'")
+          end
+        end
+
         attr_accessor :properties
         attr_accessor :current_user
         attr_accessor :pricing_options
@@ -57,16 +63,19 @@ module Spree
         end
 
         def add_search_scopes(base_scope)
-          if @properties[:search]
-            @properties[:search].each do |name, scope_attribute|
-              scope_name = name.to_sym
-              if base_scope.respond_to?(:search_scopes) && base_scope.search_scopes.include?(scope_name.to_sym)
-                base_scope = base_scope.send(scope_name, *scope_attribute)
-              else
-                base_scope = base_scope.merge(Spree::Product.ransack({ scope_name => scope_attribute }).result)
-              end
+          return base_scope unless @properties[:search].present?
+          raise InvalidOptions.new(:search) unless @properties[:search].respond_to?(:each_pair)
+
+          @properties[:search].each_pair do |name, scope_attribute|
+            scope_name = name.to_sym
+
+            if base_scope.respond_to?(:search_scopes) && base_scope.search_scopes.include?(scope_name.to_sym)
+              base_scope = base_scope.send(scope_name, *scope_attribute)
+            else
+              base_scope = base_scope.merge(Spree::Product.ransack({ scope_name => scope_attribute }).result)
             end
           end
+
           base_scope
         end
 
--- a/core/spec/lib/search/base_spec.rb
+++ b/core/spec/lib/search/base_spec.rb
@@ -89,6 +89,13 @@ RSpec.describe Spree::Core::Search::Base do
     expect(searcher.retrieve_products.count).to eq(2)
   end
 
+  it "raises a proper exception when search param is invalid" do
+    params = { per_page: "",
+               search: "blub" }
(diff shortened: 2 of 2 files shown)


core/lib/spree/core/search/base.rb
```diff
@@ -57,16 +57,18 @@ def add_eagerload_scopes(scope)
...
-            @properties[:search].each do |name, scope_attribute|
-              scope_name = name.to_sym
-              if base_scope.respond_to?(:search_scopes) && base_scope.search_scopes.include?(scope_name.to_sym)
-                base_scope = base_scope.send(scope_name, *scope_attribute)
-              else
-                base_scope = base_scope.merge(Spree::Product.ransack({ scope_name => scope_attribute }).result)
-              end
+          return base_scope unless @properties[:search].presence.respond_to?(:each_pair)
+
+          @properties[:search].each_pair do |name, scope_attribute|
```
kennyadsl, line 62: It's a bit hidden in the git diff here, but I changed `each` into `each_pair` to reflect the new `respond_to?` check we added to this method. Leaving `each` would work anyway, but it's a bit strange to check that the object responds to a method and then use another, even if they are aliased. I couldn't have used `respond_to?(each)` because even other kinds of enumerables (like arrays) would be valid at that point while the rest of the code expects a key/value enumerable.

waiting-for-dev (approved):
Thanks!

waiting-for-dev:
I just approved the PR, but I have a question. Shouldn't we treat when something that is not a Hash is given as an error and raise instead of being silent about it?

kennyadsl:
That's a good concern, @waiting-for-dev. 

The final result should probably be responding with a **400 Bad Request**:
The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing).

I was too hurried and didn't think about the possibility of raising another kind of error, even though we probably need to change something on the controller level of the frontend.

adammathys (approved):

kennyadsl:
> Shouldn't we treat when something that is not a Hash is given as an error and raise instead of being silent about it?

@waiting-for-dev please take another look now.

core/lib/spree/core/search/base.rb
```diff
@@ -4,6 +4,12 @@ module Spree
   module Core
     module Search
       class Base
+        class InvalidOptions < StandardError
```
waiting-for-dev, line 7:
```suggestion
        class InvalidOptions < ArgumentError
```
Thoughts about being more specific?

waiting-for-dev (approved):
Left a suggestion, up to your judgement 🙂

waiting-for-dev:
> Not sure if we still want to consider this a bug we want to backport. What do you think @solidusio/core-team?

I don't see it as very critical. In the end, Ruby is a dynamic language, so probably there are scores of other possible similar errors.

kennyadsl:
> I don't see it as very critical. In the end, Ruby is a dynamic language, so probably there are scores of other possible similar errors.

Backport labels removed! 👍

core/lib/spree/core/search/base.rb
kennyadsl: Nice suggestion, going to change it.

kennyadsl:
@waiting-for-dev what if we avoid the custom exception and just raise a ArgumentError?

waiting-for-dev:
> @waiting-for-dev what if we avoid the custom exception and just raise a ArgumentError?

I'd also be fine with that 🙂

kennyadsl:
I'll go with the custome exception: I think that allowing people to intercept it in the frontend and so something specific when users craft some malformed URLs has some value. They can also catch `ArgumentError` but it's too generic.

PR #4848, opened 2023-01-11
Remove Ruby v2.6 support

### LINKED ISSUE #4841: Remove support for Ruby 2.6
And previous versions, if we are still supporting them.

waiting-for-dev:
## Summary

Remove support for ruby v2.6

[Ruby v2.6 is EOL](https://www.ruby-lang.org/en/downloads/branches/) since 12 Apr 2022.

Our [new release policy](https://solidus.io/release_policy/) makes it clear that we don't support Ruby versions that are no longer maintained.

We also make a basic sanitization check in our code base to update to newer standards:

1 - Use [beginless ranges](https://ruby-doc.org/core-2.7.0/Range.html#class-Range-label-Beginless-2FEndless+Ranges) when possible.

Fixes #4841 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -342,10 +342,10 @@ workflows:
       - test_solidus:
           context: slack-secrets
           name: *name
-          matrix: { parameters: { rails: ['6.0'], ruby: ['2.6'], database: ['sqlite'], paperclip: [true] } }
+          matrix: { parameters: { rails: ['6.0'], ruby: ['2.7'], database: ['sqlite'], paperclip: [true] } }
       - test_solidus:
           context: slack-secrets
           name: *name
-          matrix: { parameters: { rails: ['5.2'], ruby: ['2.6'], database: ['sqlite'], paperclip: [true] } }
+          matrix: { parameters: { rails: ['5.2'], ruby: ['2.7'], database: ['sqlite'], paperclip: [true] } }
       - dev_tools:
           context: slack-secrets
--- a/Gemfile
+++ b/Gemfile
@@ -41,7 +41,7 @@ group :backend do
   # - https://github.com/ruby/net-protocol/issues/10
   # - https://stackoverflow.com/a/72474475
   v = ->(string) { Gem::Version.new(string) }
-  if Gem::Requirement.new(['>= 2.6', '< 3']) === Gem::Version.new(RUBY_VERSION)
+  if Gem::Requirement.new(['>= 2.7', '< 3']) === Gem::Version.new(RUBY_VERSION)
     gem 'net-http', require: false
   end
 
--- a/api/solidus_api.gemspec
+++ b/api/solidus_api.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
     f.match(%r{^(spec|script)/})
   end
 
-  s.required_ruby_version = '>= 2.6.0'
+  s.required_ruby_version = '>= 2.7.0'
   s.required_rubygems_version = '>= 1.8.23'
 
   s.add_dependency 'jbuilder', '~> 2.8'
--- a/backend/app/views/spree/admin/taxons/_form.html.erb
+++ b/backend/app/views/spree/admin/taxons/_form.html.erb
@@ -12,7 +12,7 @@
         <% if @taxon.parent %>
           <div class="input-group-prepend">
             <span class="input-group-text">
-              <%= @taxon.permalink.split('/')[0...-1].join('/') + '/' %>
+              <%= @taxon.permalink.split('/')[...-1].join('/') + '/' %>
             </span>
           </div>
         <% end %>
--- a/backend/solidus_backend.gemspec
+++ b/backend/solidus_backend.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
     f.match(%r{^(spec|script)/})
   end
 
-  s.required_ruby_version = '>= 2.6.0'
+  s.required_ruby_version = '>= 2.7.0'
   s.required_rubygems_version = '>= 1.8.23'
 
   s.add_dependency 'solidus_api', s.version
--- a/core/lib/spree/core.rb
+++ b/core/lib/spree/core.rb
@@ -84,15 +84,6 @@ module Spree
   end
 end
 
-if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.7")
-  Spree::Deprecation.warn <<~HEREDOC
-    Ruby 2.6, which reached EOL, is deprecated and will not be supported anymore from the next Solidus version.
-    Please, upgrade to a more recent Ruby version.
-    Read more on the release notes for different Ruby versions here:
-    https://www.ruby-lang.org/en/downloads/releases/
-  HEREDOC
-end
-
 if Gem::Version.new(Rails.version) < Gem::Version.new('6.0')
   Spree::Deprecation.warn <<~HEREDOC
     Rails 5.2 (EOL) is deprecated and will not be supported anymore from the next Solidus version.
--- a/core/lib/spree/testing_support/order_walkthrough.rb
+++ b/core/lib/spree/testing_support/order_walkthrough.rb
@@ -34,7 +34,7 @@ module Spree
                               states
                             else
                               end_state_position = states.index(state.to_sym)
-                              states[0..end_state_position]
+                              states[..end_state_position]
                             end
 
         states_to_process.each do |state_to_process|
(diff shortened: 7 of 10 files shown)


jarednorman (approved):
I honestly don't think numbered block parameters improve readability, but will go with the team if others thing it's an improvement. Otherwise this all looks good to me.

tvdeyen:
I am not sure about this. What is the version we want to ship this in? 4.0? Or already 3.3?

We still support Rails 5.2 and [even Rails 6.1 still supports Ruby >= 2.5](https://github.com/rails/rails/blob/6-0-stable/rails.gemspec#L12). For a good reason.

We as an engine need to support a wide range of Ruby versions even if they are EOL. This does not matter for us, because we do not run an app, we provide software for apps. This is a whole different thing. The apps decide when to update which dependency and we do not want to force a Ruby update, a Rails update and Solidus update the same time on our users. We want them to be able to make updates for fundamental dependencies piece by piece.

We can stop building for Ruby 2.6, ok. We can even stop accepting PRs related to Ruby 2.6, but we should not break apps like that. What if an store needs an important Fix or Feature and they cannot update because they are still running Ruby 2.6?

I am not sure we should do that.

waiting-for-dev:
> I honestly don't think numbered block parameters improve readability, but will go with the team if others thing it's an improvement. Otherwise this all looks good to me.

@jarednorman, @kennyadsl, @spaghetticode, I removed the commit introducing numbered parameters as per our talk offline. We decided to do it opportunistically. @elia, tagging you as you also were interested in it.

jarednorman:
I assume this is going into the next major and won't be backported?

waiting-for-dev:
> I am not sure about this. What is the version we want to ship this in? 4.0? Or already 3.3?

@tvdeyen, the idea is v3.3. As per our [new release policy](https://solidus.io/release_policy/):

> Solidus is in strict compliance with [Semantic Versioning](https://semver.org/). Our definition of a Breaking Change is any update to the Solidus code that can break a host application, except for:
> ...
> - Issues due to the application using a Ruby, Rails or any other dependency version that is no longer maintained by the corresponding community.

> We still support Rails 5.2 and [even Rails 6.1 still supports Ruby >= 2.5](https://github.com/rails/rails/blob/6-0-stable/rails.gemspec#L12). For a good reason.

For the same reason, our idea includes removing support for Rails 5.2. Our plan is consistent with what Rails does, as we're not going to remove support for older rubys in older Solidus versions. I.e., we're not going to backport the removals. I agree we should be clearer about what we support and what we don't, through a compatibility table or something.
> 
> We as an engine need to support a wide range of Ruby versions even if they are EOL. This does not matter for us, because we do not run an app, we provide software for apps. This is a whole different thing. The apps decide when to update which dependency and we do not want to force a Ruby update, a Rails update and Solidus update the same time on our users. We want them to be able to make updates for fundamental dependencies piece by piece.
> 
> We can stop building for Ruby 2.6, ok. We can even stop accepting PRs related to Ruby 2.6, but we should not break apps like that. What if an store needs an important Fix or Feature and they cannot update because they are still running Ruby 2.6?

Our reasoning is that we don't want to support insecure practices: i.e., running on Ruby and Rails versions that don't get security updates anymore. An app is a stack of dependencies, and they need to be updated from bottom to top. If Ruby is not updated, you can't expect to be able to run the latest version of its libraries. On top of that, if we need to support EOL versions, that means that we'll be locked forever and won't be able to make Solidus evolve.

jarednorman:
I often think about how best we handle this kind of thing. On the one hand, we certainly don't want to hang any stores out to dry, but we also want there to be some pressure to keep people upgrading regularly. Much of the value of Solidus is lost when people fall behind on updates, so there should be _some_ pressure there.

tvdeyen:
Urgh, you are all faster then me. So my answer might be a bit outdated 😄 

@waiting-for-dev 

> Our reasoning is that we don't want to support insecure practices: i.e., running on Ruby and Rails versions that don't get security updates anymore.

This is noble goal and we should encourage our users to do that. But we cannot force them, we need to support them in updating it by having a wide range of versions supported.

> An app is a stack of dependencies, and they need to be updated from bottom to top. If Ruby is not updated, you can't expect to be able to run the latest version of its libraries. 

Exactly and updating an app from the bottom up needs to be done piece by piece. Let me tell you that from experience. I made "a couple" of major updates of Rails, Spree, Solidus and what not. Forcing an dependency update on users should be avoided by any means if you provide infrastructure as we do.

> On top of that, if we need to support EOL versions, that means that we'll be locked forever and won't be able to make Solidus evolve.

What feature in Ruby 2.6 is harming Solidus evolving?. We do not need to support it, but still allow it. There is a major difference here. Forcing an update vs. allowing an update. We can encourage it, by ie. not running specs for it. But the app need to be able to `bundle install` an  update of Solidus, without forcing a Ruby update as well.

> I assume this is going into the next major and won't be backported?

@jarednorman major updates of Solidus is a task of its own. We do not want to force a Ruby update on them as well right?

jarednorman:
> @jarednorman major updates of Solidus is a task of its own. We do not want to force a Ruby update on them as well right?

Yes, but this isn't like the Spree releases that contained many breaking changes _as well as bumping the Rails version requirement_. In this case stores can upgrade Ruby totally independently of when we do this change.

The improvements we get from dropping support are definitely minor, though. I'm torn. I think we should definitely "officially" drop support for EOL versions of Ruby, but how we handle that is a different question.

waiting-for-dev:
> Exactly and updating an app from the bottom up needs to be done piece by piece. Let me tell you that from experience. I made "a couple" of major updates of Rails, Spree, Solidus and what not. Forcing an dependency update on users should be avoided by any means if you provide infrastructure as we do.

To be clear, we won't be forcing stores to update more than one piece at a time. A store running Solidus v3.2 with Ruby 2.6, would need to:

- Update from Solidus **v3.2** on Ruby **2.6** to Solidus **v3.2** on Ruby **2.7**
- Update from Solidus **v3.2** on Ruby **2.7** to Solidus **v3.3** on Ruby **2.7**

elia:
FWIW **I'd be open to not bumping the required ruby version in the gemspec and just use a message in the changelog.** Just to be sure I understood correctly, is that what you're suggesting @tvdeyen?

Anyway I think that the moment we stop testing against it breaking support is around the corner.
If I was on a difficult situation like the one described I would probably end up with back porting the security-fix / patch / feature manually inside the app …or outright using my own copy of solidus, because it's unlikely to be the only thing that breaks.

waiting-for-dev:
@tvdeyen, I understand this example you gave summarizes the problem we're trying to solve:

> @elia let's consider the following: A major security issue in Solidus is fixed in 3.3 - or even an important feature has shipped that we do not want to or can back port. Do you also want to force a Ruby update on them? Ruby 2.6 has some potential breaking changes in it.

- For security fixes, they will be backported, so there's no problem here.
- For features: yes, a Ruby update would also be needed. But, again, we're **not** requiring to update everything at once. An app can update to a supported Ruby and, after that, to the new Solidus with the new feature. FWIW, that's something that the Ruby community is pretty used to as that's also what [recommends does](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#ruby-versions) (the highlight is mine):

> Rails generally stays close to the latest released Ruby version when it's released:
>
> - Rails 7 requires Ruby 2.7.0 or newer.
> - Rails 6 requires Ruby 2.5.0 or newer.
> - Rails 5 requires Ruby 2.2.2 or newer.
>
> It's a good idea to upgrade Ruby and Rails separately. **Upgrade to the latest Ruby you can first, and then upgrade Rails.**

We need to understand that's a compromise. Of course, ideally, we should support all the matrix of possibilities, but that wouldn't leave our ecosystem in a healthy situation for the reasons Alberto mentioned above.

> FWIW I'd be open to not bumping the required ruby version in the gemspec and just use a message in the changelog. Just to be sure I understood correctly, is that what you're suggesting @tvdeyen?

@elia, I wouldn't like either to leave the ruby requirement in the `gemspec` untouched while allowing ourselves to stop supporting Ruby 2.6 code. As an app developer, I'd prefer to notice that in development rather than in some incompatible runtime run in production.

tvdeyen:
Thanks for all the input and answers. It seems it has been considered very well. I am fine with this then.

tvdeyen (approved):

waiting-for-dev:
> Thanks for all the input and answers. It seems it has been considered very well. I am fine with this then.

Thanks for your being receptive to this, @tvdeyen. I'm happy your concerns have been addressed. Needless to say, feel free to reach out if there's something that needs clarification.

Issue #4849, opened 2023-01-11
Problematic Taxons can be created


RyanofWoods:
A lot of these problems are caused by incorrectly Taxons via code rather than using the admin page, but some simple validations could prevent these issues.

**Solidus Version**:
all/master

**Issues**
1.
A Taxonomy should have one root Taxon, but still multiple Taxon's can be created with a `parent_id` of `nil`.
https://github.com/solidusio/solidus/blob/2b79f72aa53f5caa850c587888fff46c1c91f7b7/core/app/models/spree/taxonomy.rb#L10

2. You can make "duplicate" Taxons, i.e. Taxons that have the same `name` and parent.

These both can be fixed with simple validation.

3. If you have taxons with the same `lft` and `rgt` attributes, deleting one, deletes all of them. I am not really sure how this occurred on the project I am working on, but just wanted to share, while I figure out the cause.

PR #4851, opened 2023-01-12
Improve Taxon validations and factory

RyanofWoods:
## Summary

To resolve the two main issues outlined in https://github.com/solidusio/solidus/issues/4849. This will help stores creating problematic Taxons when creating them manually as well as make it easier to setup Taxons correctly via its factory.

## Checklist


core/lib/spree/testing_support/factories/taxon_factory.rb
```diff
@@ -9,12 +9,19 @@
...
+
     name { 'Ruby on Rails' }
-    taxonomy
-    parent_id { nil }
+    taxonomy_id { taxonomy.id }
+    parent_id { parent&.id || taxonomy.root.id }
 
     trait :with_icon do
-      icon { Spree::Core::Engine.root.join('lib', 'spree', 'testing_support', 'fixtures', 'blank.jpg').open }
+      after(:create) do |taxon|
```
RyanofWoods, line 22: Note, `after(:build)` and setting the icon manually does not work 🤔  This change could break tests in stores if they were building taxons with icons?

core/app/models/spree/taxon.rb
```diff
@@ -21,6 +21,7 @@ class Taxon < Spree::Base
     validates :meta_keywords, length: { maximum: 255 }
     validates :meta_description, length: { maximum: 255 }
     validates :meta_title, length: { maximum: 255 }
+    validates :taxonomy_id, uniqueness: { message: 'can only have one root Taxon' }, if: :root?
```
jarednorman, line 24: It would be better to rely on I18n for this so that it can be set separately per language for stores that need that. https://guides.rubyonrails.org/i18n.html#error-message-scopes

jarednorman (changes requested):

kennyadsl (commented):
Love it! I've been hit by this issue in the past, thanks for taking care of this. I left some comments but overall this is a great change I look forward to merge.

core/lib/spree/testing_support/factories/taxon_factory.rb
RyanofWoods: Sorry I mean, I had to change it to `after(:create)` for icons to be set correctly for nested taxons. If you checkout to this branch and revert this local change you can see that `core/spec/models/spree/taxon_spec.rb` fails because of this.

kennyadsl: I don't think this could be breaking for existing stores as long as they just use the `with_icon` trait. By the way, were you able to identify the cause of the spec failure without this change?

RyanofWoods: Hmmm, it could if they are building taxons with icon, because the icon won't be set, whereas it was before. I haven't been able to identify the cause yet. 🤔

kennyadsl: Sorry, not sure to understand. Can you help with a spec in an existing store that was passing before this change and would fail if we merge?

RyanofWoods: Hard to think of a plausible reason stores are using built taxons with icons and testing the icons behavior, so it's probably not of concern. But as a demonstration:

```ruby
  context "a built taxon's icon" do
    it "can be removed" do
      taxon = build(:taxon, :with_icon)

      # because the icon is no longer set
      expect(taxon.destroy_attachment(:icon)).to be_truthy
    end
  end
```

kennyadsl: But why do you say it's no longer set? I think it is in the `after(:create)` block, isn't it?

RyanofWoods: No, because `after(:create)` is not triggered for build records @kennyadsl 
https://thoughtbot.com/blog/aint-no-calla-back-bot

kennyadsl: Ahhhhh! I wasn't thinking about the build vs create, but now I got it and I think you are right. 

Yes, that's not optimal. I think it's worth some more detailed investigation for the root cause of the error. I'm available to help with that if needed.

core/lib/spree/testing_support/factories/taxon_factory.rb
```diff
@@ -9,12 +9,19 @@
 
 FactoryBot.define do
   factory :taxon, class: 'Spree::Taxon' do
+    transient do
+      parent { nil }
+      taxonomy { parent&.taxonomy || create(:taxonomy) }
```
waiting-for-dev, line 14: Why does it need to be within the `transient` block? 🤔

RyanofWoods: Interesting, I thought it was needed because `parent` and `taxonomy` are not attributes of Taxon. But the following works if you prefer it. I guess it works because `parent` and `taxonomy` are associations!
```ruby
  factory :taxon, class: 'Spree::Taxon' do
    name { 'Ruby on Rails' }
    taxonomy_id { (parent&.taxonomy || create(:taxonomy)).id }
    parent_id { parent&.id || taxonomy.root.id }

    trait :with_icon do
      after(:create) do |taxon|
        taxon.update(icon: Spree::Core::Engine.root.join('lib', 'spree', 'testing_support', 'fixtures', 'blank.jpg').open)
      end
    end
  end
```

waiting-for-dev: Yes, you can refer to the associations from the regular context 🙂

RyanofWoods: Done, thanks!

RyanofWoods:
Regarding the old taxon factory `with_icon` trait @kennyadsl 
https://github.com/solidusio/solidus/blob/2223700f392c2e9da1b8ef53f6ba1e5676a35f35/core/lib/spree/testing_support/factories/taxon_factory.rb#L16-L18

The old code for setting the icon no longer works as it sets the icon and an `after_save` callback that comes from `awesome_nested_set` causes the icon to be lost.

[after_save callback](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/awesome_nested_set.rb#L69) > [move_to_new_parent](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/model/movable.rb#L122) > [move_to_child_of](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/model/movable.rb#L38) > [move_to](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/model/movable.rb#L99) > [reload_nested_set](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/model.rb#L263) (the cause!)

---

Doesn't make much sense to have save the icon on build records imo, seeing as in admin when you go to create a new taxon, it creates a dummy one and only then can you add an icon.

core/lib/spree/testing_support/factories/taxon_factory.rb
RyanofWoods: @kennyadsl 
The old code for setting the icon no longer works as it sets the icon before building, and an `after_save` callback that comes from `awesome_nested_set` causes the icon to be lost.

[after_save callback](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/awesome_nested_set.rb#L69) > [move_to_new_parent](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/model/movable.rb#L122) > [move_to_child_of](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/model/movable.rb#L38) > [move_to](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/model/movable.rb#L99) > [reload_nested_set](https://github.com/collectiveidea/awesome_nested_set/blob/6409bab837c9fa786509616bc409422b47ddbece/lib/awesome_nested_set/model.rb#L263) (the cause!)

---

Doesn't make much sense to have save the icon on build records imo, seeing as in admin when you go to create a new taxon, it creates a dummy one and only then can you add an icon.

kennyadsl: Sorry for taking a lot of time to get back here. Yes, I think it's an acceptable compromise, and this sounds more like a bug for which we can introduce a breaking change.

kennyadsl:
@RyanofWoods would love to move this forward. Can you please rebase?

RyanofWoods:
I rebased and added one new commit @kennyadsl. It fixes a mistake I made in the Taxon factory. On approval, I can squash it into the `Improve Taxon factory` commit.

core/spec/lib/spree/core/testing_support/factories/taxon_factory_spec.rb
```diff
@@ -13,6 +13,12 @@
     context "when no taxonomy is given" do
       subject { create(:taxon) }
 
+      before do
+        # ensure that the subject is not accidentally
+        # getting valid parent and taxonomy ids
```
kennyadsl, line 18: This comment is clear only in combination with the commit description. So I propose to have still a separate commit that adds this `before` only and explains why we need it, plus, squashing the other part of the commit (the one related to  `core/lib/spree/testing_support/factories/taxon_factory.rb`) in a previous commit, so this one only contains this extra safeguard.

RyanofWoods: I will change the commit title, thanks!

core/app/models/spree/taxon.rb
```diff
@@ -17,10 +17,11 @@ class Taxon < Spree::Base
     before_update :set_permalink
     after_update :update_child_permalinks, if: :saved_change_to_permalink?
 
-    validates :name, presence: true
+    validates :name, presence: true, uniqueness: { scope: :parent_id, message: :must_be_unique_under_same_parent }
     validates :meta_keywords, length: { maximum: 255 }
     validates :meta_description, length: { maximum: 255 }
     validates :meta_title, length: { maximum: 255 }
+    validates :taxonomy_id, uniqueness: { message: :can_have_only_one_root }, if: :root?
```
waiting-for-dev, line 24: Isn't it more `:root_must_be_unique`? I mean, the error is not that the taxon has many roots, but that the taxonomy has already been taken as a parent.

Regardless, should it be something that users can change? As far as I can see, even if it's [permitted by the controller](https://github.com/nebulab/solidus/blob/5d119ba3d1e05c508a6e991bdc5649174a6096c7/core/lib/spree/permitted_attributes.rb#L121), that's not part of the backend's UI:

**Creation**
screenshot-localhost_3000-2023 02 23-05_44_18

**Update**
screenshot-localhost_3000-2023 02 23-05_44_49

However, it can be modified via API both in [creation](https://solidus.stoplight.io/docs/solidus/2e0e5c45af59c-create-taxonomy-taxon) and [update](https://solidus.stoplight.io/docs/solidus/18c69a84a5e88-update-taxonomy-taxon).

Given that, it makes sense to add the validation now but should we deprecate the `:taxonomy_id` attribute from the list of modifiable fields in the API?

core/config/locales/en.yml
```diff
@@ -498,6 +498,12 @@ en:
           attributes:
             base:
               cannot_destroy_default_store: Cannot destroy the default Store.
+        spree/taxon:
+          attributes:
+            name:
+              must_be_unique_under_same_parent: must be unique under the same parent Taxon
+            taxonomy_id:
+              can_have_only_one_root: can only have one root Taxon
```
waiting-for-dev, line 506: Same consideration about the error message.

waiting-for-dev (commented):
Thanks once more, @RyanofWoods, I left one consideration.

core/app/models/spree/taxon.rb
RyanofWoods: Hmm, I disagree. I think the error and validation is about not allowing multiple roots under one Taxonomy. Parents are not so relevant, as parents are just describing how the Taxons are nested under the Taxonomy. The only relevancy is that root Taxons have no parent.

I am not 100% set on `root_must_be_unique`. It's half correct, as we are not testing only the uniqueness of `taxonomy_id`. We are testing that it's unique when `parent_id` is `nil`, or that all roots have a unique Taxonomy. The following Taxons would be considered "unique" but are not valid as they are both trying to be roots for the same Taxonomy:
- `taxonomy_id: 1, parent_id: nil, name: 'Categories', permalink: 'Categories'`
- `taxonomy_id: 1, parent_id: nil, name: 'Brands', permalink: 'Brands'`.

But this is just semantics over a message name, so I am open to other suggestions.

---

What part would you consider to be changed by the user? I don't think we should let the validation be optional, as Taxonomies assume they only have one root:
https://github.com/solidusio/solidus/blob/2b79f72aa53f5caa850c587888fff46c1c91f7b7/core/app/models/spree/taxonomy.rb#L10

Note, in admin you cannot make attempt to make multiple roots for one Taxonomy. You also can't modify the `taxonomy_id` in the update page. `parent_id` still needs to be permissable as you can move a Taxon's nesting.

waiting-for-dev: > Hmm, I disagree. I think the error and validation is about not allowing multiple roots under one Taxonomy. Parents are not so relevant, as parents are just describing how the Taxons are nested under the Taxonomy. The only relevancy is that root Taxons have no parent.

Oh, I see. I was confusing root with parent. The problem is that it's a message that refers to another model, Taxonomy, and not to the Taxon one, so it reads weird. But I can't think of a better name now 🤷 

> What part would you consider to be changed by the user? I don't think we should let the validation be optional, as Taxonomies assume they only have one root:

I mean that the creation of the root taxon is something automatic (unless I'm wrong), so users shouldn't deal with it. But that would be part of another PR.

waiting-for-dev (approved):
Thanks again, @RyanofWoods 🙌 @jarednorman, I think that your concern has been addressed 🙂

core/app/models/spree/taxon.rb
RyanofWoods: Fair point @waiting-for-dev. I wanted to note, making root Taxons is something users should generally don't have to worry about. When you make a Taxonomy, it automatically makes the root Taxon:
https://github.com/solidusio/solidus/blob/5d119ba3d1e05c508a6e991bdc5649174a6096c7/core/app/models/spree/taxonomy.rb#L27

Though, I think it would still be beneficial to have validation to ensure multiple Taxons under the same Taxonomy don't have the `parent_id` as `nil`. Which would mess up the finding of the root. This can happen easily if you make Taxons via code and forget to set the `parent_id`.

Regarding the message, what do you think about keeping the `can_have_only_one_root` name unless we think of something better and changing the message to: "must be unique if parent is set to nil (Taxonomy can only have one root Taxon)"?

waiting-for-dev: Yeah, @RyanofWoods, I'm good with the name 🙂

RyanofWoods: But did you want to keep the message as is or? 🤔

waiting-for-dev: Oh, sorry, yeah, I'm good 👍

forkata (commented):
This is great @RyanofWoods, thank you for implementing this validation in core 🎉 

I have just one question regarding this change, is it worth including a simple script in the Changelog along side this PR, so users who are upgrading can ensure that they don't have invalid records in the database after this new validation is introduced? 

Although unlikely for admin created taxons, it is possible that this change invalidates existing records and prevents updates to those after an upgrade, which could easily be unnoticed until an admin tries to edit the records.

RyanofWoods:
What's your opinion on @forkata's suggestion @waiting-for-dev?

waiting-for-dev:
> What's your opinion on @forkata's suggestion @waiting-for-dev? I mean after the upgrade, developers would be able to iterate over them and find invalid Taxons. So they could do this locally if they have Taxon + Taxonomy prod data.
> 
> But would it be better to release this change in the a major, and add a script for checking Taxons in a minor version before, so they can fix them before the upgrade? thinking

As that's something not expected by our business domain, I'd say it's ok to release in a minor. The task would be a nice addition, or we could also write the instructions to manually do the checking on the release notes. However, I'll forward decisions of what needs to go into a major to @kennyadsl :wink:

kennyadsl:
I think the safest way to take this scenario into account (never easy to add validations, thanks for raising this @forkata), is to add a new preference called something like `Spree::Config.extra_taxon_validations`, which is true for new installations and who upgrades needs to turn it on manually, so they are aware of what's happening, and they have time to check their taxons.

Including that simple script in the release notes is enough at that point, in my opinion, users will know how to use that script in their context.

RyanofWoods:
Thanks @kennyadsl, I will investigate this next week. Thanks for raising this @forkata! 🙌

core/lib/spree/app_configuration.rb
```diff
@@ -155,6 +155,18 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Regexp] Regex to be used in email validations, for example in Spree::EmailValidator
     preference :default_email_regexp, :regexp, default: URI::MailTo::EMAIL_REGEXP
 
+    # @!attribute [rw] extra_taxon_validations
+    #   Use extra validations on Taxons in 3.4.0, but default to false so stores can
+    #   upgrade and then fix any now invalid Taxons before enabling.
+    #   @return [Boolean]
+    versioned_preference :extra_taxon_validations, :boolean, initial_value: false, boundaries: { "3.4.0" => true }
```
waiting-for-dev, line 162:
```suggestion
    versioned_preference :extra_taxon_validations, :boolean, initial_value: false, boundaries: { "3.4.0.dev" => true }
```

You can use the current version in `core/lib/spree/core/version.rb` to fix CI :slightly_smiling_face:

core/lib/spree/app_configuration.rb
```diff
@@ -155,6 +155,18 @@ class AppConfiguration < Preferences::Configuration
...
+
+    # @!attribute [rw] extra_taxonomy_validations
+    #   Use extra validations on Taxonomies in 3.4.0, but default to false so stores can
+    #   upgrade and then fix any now invalid Taxonomies before enabling.
+    #   @return [Boolean]
+    versioned_preference :extra_taxonomy_validations, :boolean, initial_value: false, boundaries: { "3.4.0" => true }
```
waiting-for-dev, line 168: Same here!

waiting-for-dev (changes requested):
Thanks, @RyanofWoods, I left a comment to fix the failing tests!

waiting-for-dev (approved):
Awesome!!!

RyanofWoods:
Note @waiting-for-dev, as well as fixing the tests, I made the new preferences deprecated as suggested by @kennyadsl privately.

kennyadsl (approved):
Nice, that's the way to go!

kennyadsl:
@jarednorman can you please review it again? 🙏

RyanofWoods:
An explanation on the new validations, how to proceed with them, and checking invalid Taxons + Taxonomies will be needed

jarednorman (approved):

RyanofWoods:
WDYT about the release notes in the description @kennyadsl and @waiting-for-dev?

kennyadsl:
@RyanofWoods Nice, just one thing:

>Using version 3.4 new defaults will enable these validations unless you override them with false.

I don't think this is correct. By upgrading, we aren't going to change the default. People will have to change that value when ready explicitly. On new installations, instead, the new value will be used.

RyanofWoods:
@kennyadsl Does the boundary only mean that only `3.4.0.dev` has the preference as `true` and not `3.4.0.dev+`? So stores using the `3.4.0` defaults will have it as `false`? I am little confused on the boundaries 🤔

kennyadsl:
The boundaries indicate from which version that will be the default value. In other words: new applications will get that value if they are created with a version >= 3.4.0.dev. Existing applications will have the other value, but in the upgrade process (running `bin/rails g solidus:update`), we will copy an initializer that will guide the user in updating their app to the new defaults. We've written more info about this at: https://guides.solidus.io/getting-started/upgrading-solidus#updating-preferences. Let us know if it's clear.

RyanofWoods:
Thanks for the documentation @kennyadsl, now I understand it much better. Please see the updated release notes (PR description history helps!). If you have a better idea on how stores can find and fix Taxons on production other than pulling them down and testing them, let me know 🤔

kennyadsl:
> If you have a better idea on how stores can find and fix Taxons/Taxonomies on production other than pulling them down and testing them, let me know

Can't see any better path, but because it's only documentation, we can update even later if we have ideas.

Thanks a lot, @RyanofWoods!

PR #4858, opened 2023-01-16
Allow storing static preferences using string class names

elia:
## Summary

This will make unnecessary to wrap these definitions under `.to_prepare` blocks.

Additionally the error message for unexpected keys has improved and now lists both all the unexpected keys along with the expected keys, mentioning the name associated to the definition.

The only thing that changed is that it doesn't do so when the configuration is added but rather when it's fetched.

Completes https://github.com/solidusio/solidus/pull/4070 and addresses the original request of https://github.com/solidusio/solidus/issues/4040.

## Checklist


core/lib/spree/preferences/static_model_preferences.rb
```diff
@@ -27,20 +22,30 @@ def []=(key, value)
...
         end
+
+        delegate :keys, to: :@preferences
       end
 
       def initialize
-        @store = Hash.new do |data, klass|
-          data[klass] = {}
-        end
+        @store = {}
```
waiting-for-dev, line 30: I do think it's cleaner to create the Hash with a fresh hash as the default value, for a few reasons:

- Otherwise, we need to sanitize it every time we access a key, as we're doing now twice in `#add` & `#fetch`.
- Could break something in edge cases where code relies on returning a Hash.
- In my view, it communicates better as it nails down the data structure we need.

Thoughts?

core/spec/models/spree/preferences/static_model_preferences_spec.rb
```diff
@@ -13,6 +13,8 @@ module Spree
     let(:other_class){ Class.new }
     let(:definitions){ subject.for_class(preference_class) }
 
+    before { stub_const("SomeClass", preference_class) }
```
waiting-for-dev, line 16: We only need this stub within the new test case. What do you think about eliminating it and interpolating `preference_class.to_s` in the error message?

waiting-for-dev (commented):
Thanks for working on it, @elia. That's something that needs to be addressed 🙌 

However, did you happen to check if we will still render the error message on initialization, or otherwise, will it happen only for affected runtimes? Keeping the first behavior is optimal, as it saves from more headaches on production. If that changes, we could work around that by having one `#to_prepare` block on our engine doing the sanitization check for all configured preferences.
 
What do you think?

core/lib/spree/preferences/static_model_preferences.rb
elia: The reason I touched this is that while storing prefs for solidus_stripe I noticed that trying a non existing key would add the key to the hash, e.g.:

```
>> sp = Spree::Preferences::StaticModelPreferences.new
=> #<Spree::Preferences::StaticModelPreferences:0x00000001124dac70 @store={}>
>> sp.for_class("Foo")
=> {}
>> sp
=> 
#<Spree::Preferences::StaticModelPreferences:0x00000001124dac70
 @store={"Foo"=>{}}>          
>> 
```

I'll change it to a simpler `Hash.new { {} }` so it behaves this way:

```
>> h = Hash.new { {} }
=> {}
>> h[:asdf]
=> {}
>> h
=> {}
>> h[:asdf] = 1234
=> 1234
>> h
=> {:asdf=>1234}
>> 
```

elia: I had to walk back on this, my second example wasn't correct, and after all we only need to set the empty hash in `#add` and fallback to a default in `#for_class` which is not much.

The whole preferences system could probably be simplified but didn't want to launch into a campaign, and this will at least provide a consistent the experience within initializers.

core/spec/models/spree/preferences/static_model_preferences_spec.rb
elia: Done, thanks for the suggestion!

waiting-for-dev:
Hey, @elia, it looks nice, but what do you think about:

> However, did you happen to check if we will still render the error message on initialization, or otherwise, will it happen only for affected runtimes? Keeping the first behavior is optimal, as it saves from more headaches on production. If that changes, we could work around that by having one #to_prepare block on our engine doing the sanitization check for all configured preferences.

elia:
> However, did you happen to check if we will still render the error message on initialization, or otherwise, will it happen only for affected runtimes? Keeping the first behavior is optimal, as it saves from more headaches on production. If that changes, we could work around that by having one `#to_prepare` block on our engine doing the sanitization check for all configured preferences.
> 
> What do you think?

@waiting-for-dev sorry I though I addressed that! I updated the PR moving the check to a `validate!` method that is being run once inside an `after_initializer`. Since the static preferences are not reloaded I thought it's not strictly necessary to check at every code change.

core/lib/spree/core/engine.rb
```diff
@@ -87,6 +87,7 @@ class Engine < ::Rails::Engine
 
       config.after_initialize do
         Spree::Config.check_load_defaults_called('Spree::Config')
+        Spree::Config.static_model_preferences.validate!
```
kennyadsl, line 90: What if you set the preference value after the app is initialized? Do you need to call `validate!` manually as well?

elia: After the app is initialized you could change the model configuration (e.g. adding or removing a preference) not the static preference value, but I regarded that as infrequent.

If we want to do it at every request in development we could move it to a `to_prepare` block, would you prefer that?

kennyadsl: You are right, they live in memory and we can't change their value at runtime. At least, not easily and it's not an expected usage.

core/lib/spree/preferences/static_model_preferences.rb
waiting-for-dev: Hmm, I see. I never realized that Ruby had this behavior. What about the change in the API (i.e., returning `nil` instead of `Hash`? Do you think it's plausible it could break stores? cc @kennyadsl

elia: I'd say that the whole API a bit sparse and could be streamlined, it also only serves one model inside solidus (although it's probably more across the ecosystem). But I wanted to surgically solve only the problem of not being able to pass strings for class names and finally make it consistent with other parts of the configuration.

There's also probably a case for making some of these APIs to be private and so not susceptible to backward compatibility support.

waiting-for-dev (approved):
❤️

kennyadsl (approved):

PR #4861, opened 2023-01-19
Improve Bogus (test) Credit Card voiding

kennyadsl:
Follow up for #4843 and #4859.

⚠️ This doesn't change any behavior of real credit card processing. This change has been made to improve DX, when someone is trying to simulate a realistic order flow (create order, capture payment, cancel order) with the provided BogusCreditCard placeholder, which we use in our demo and sandbox.

## Summary

Before this change, all the attempts to void a payment in the admin panel would be succeeding. This is not realistic because, in real life, you can't void captured credit card payments. 

With this change, we can simulate a failure (by trying to cancel an order with captured payment). This will leave the payment in the complete state and will create a refund for that payment. 

## Checklist

### DIFF
--- a/core/app/models/spree/payment_method/bogus_credit_card.rb
+++ b/core/app/models/spree/payment_method/bogus_credit_card.rb
@@ -57,8 +57,12 @@ module Spree
       end
     end
 
-    def void(_response_code, _credit_card, _options = {})
-      ActiveMerchant::Billing::Response.new(true, SUCCESS_MESSAGE, {}, test: true, authorization: AUTHORIZATION_CODE)
+    def void(_response_code, _credit_card, options = {})
+      if options[:originator].completed?
+        ActiveMerchant::Billing::Response.new(false, FAILURE_MESSAGE, {}, test: true, authorization: AUTHORIZATION_CODE)
+      else
+        ActiveMerchant::Billing::Response.new(true, SUCCESS_MESSAGE, {}, test: true, authorization: AUTHORIZATION_CODE)
+      end
     end
 
     # @see Spree::PaymentMethod#try_void
--- a/core/app/models/spree/payment_method/bogus_credit_card.rb
+++ b/core/app/models/spree/payment_method/bogus_credit_card.rb
@@ -65,11 +65,6 @@ module Spree
       end
     end
 
-    # @see Spree::PaymentMethod#try_void
-    def try_void(_payment)
-      ActiveMerchant::Billing::Response.new(true, SUCCESS_MESSAGE, {}, test: true, authorization: AUTHORIZATION_CODE)
-    end
-
     def test?
       # Test mode is not really relevant with bogus gateway (no such thing as live server)
       true
--- a/core/spec/models/spree/order/state_machine_spec.rb
+++ b/core/spec/models/spree/order/state_machine_spec.rb
@@ -92,31 +92,42 @@ RSpec.describe Spree::Order, type: :model do
       expect(mail.subject).to include "Cancellation"
     end
 
-    context "resets payment state" do
-      let!(:payment) { create(:payment, order: order, amount: order.total, state: "completed") }
+    it "tries to cancel payments when the order has a complete payment" do
+      payment = create(:payment, order: order, amount: order.total, state: "completed")
 
-      context "without shipped items" do
-        it "should set payment state to 'void'" do
-          expect { order.cancel! }.to change{ order.reload.payment_state }.to("void")
-        end
-      end
+      expect(payment).to receive(:cancel!)
+      expect(order).to be_allow_cancel
 
-      context "with shipped items" do
-        before do
-          order.shipments[0].ship!
-        end
+      order.cancel!
+    end
 
-        it "should not alter the payment state" do
-          expect(order).to_not be_allow_cancel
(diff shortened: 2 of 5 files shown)


elia (approved):

core/spec/models/spree/order/state_machine_spec.rb
```diff
@@ -113,9 +113,18 @@
...
+        let!(:payment) { create(:payment, order: order, amount: order.total, state: "pending") }
+
+        it "should change the payment state whe the payment is not completed" do
+          expect(order).to be_allow_cancel
+          expect { order.cancel! }.to change{ payment.reload.state }.to("void")
+        end
+      end
+
+      it "should not change the payment state whe the payment is completed" do
         expect(order).to be_allow_cancel
```
waiting-for-dev, line 126: Shouldn't this expectation fail now?

core/spec/models/spree/order/state_machine_spec.rb
```diff
@@ -113,9 +113,18 @@
         end
       end
 
-      it "should automatically refund all payments" do
+      context "with a pending payment" do
```
waiting-for-dev, line 116: If I understand correctly, the tested behavior depends on the underlying payment method. Do these tests make sense here?

core/app/models/spree/payment_method/bogus_credit_card.rb
```diff
@@ -57,13 +57,12 @@ def capture(_money, authorization, _gateway_options)
...
     end
 
-    def void(_response_code, _credit_card, _options = {})
-      ActiveMerchant::Billing::Response.new(true, SUCCESS_MESSAGE, {}, test: true, authorization: AUTHORIZATION_CODE)
-    end
-
-    # @see Spree::PaymentMethod#try_void
-    def try_void(_payment)
-      ActiveMerchant::Billing::Response.new(true, SUCCESS_MESSAGE, {}, test: true, authorization: AUTHORIZATION_CODE)
+    def void(_response_code, _credit_card, options = {})
```
waiting-for-dev, line 60: Should we add/update tests for that in the bogus credit card specs?

waiting-for-dev (approved):
Thanks, @kennyadsl. Leaving some comments. It looks like the tests don't reflect the current responsibilities. Feel free to merge if I'm missing something or if you see it as something that we can't address at this point (this PR is already an improvement on its own).

kennyadsl:
@waiting-for-dev I think your suggestions make sense, I'll take a look. To be honest, I didn't expect any specs to fail after this change but apparently we have a lot of specs relying on the BogusCreditCard behavior, being that the parent class of the factories we are using in specs for credit card payments.

core/spec/models/spree/order/state_machine_spec.rb
kennyadsl: Mhh, no. The completed payment state doesn't change. It stays `completed` but a refund will be issued.

kennyadsl: Anyway, there's a typo in the spec description which I'm going to fix.

core/app/models/spree/payment_method/bogus_credit_card.rb
kennyadsl: At the moment, there are no specs for this class: https://github.com/solidusio/solidus/blob/c6050d71707063d1e46693f7d388ebc4c07d1a60/core/spec/models/spree/payment_method/bogus_credit_card_spec.rb

core/spec/models/spree/order/state_machine_spec.rb
kennyadsl: You are right, this should probably go in the specific payment method specs. And we should just test that the outer interface method is called here. Let me see if I can do the change.

kennyadsl: Again, you were right. This behavior is already tested in [order specs](https://github.com/solidusio/solidus/blob/36340a8785294b7287b03599232be8bf44de548e/core/spec/models/spree/order_spec.rb#L43-L125) and [payment cancellation specs](https://github.com/solidusio/solidus/blob/36340a8785294b7287b03599232be8bf44de548e/core/spec/models/spree/payment/cancellation_spec.rb). This change gives us the opportunity to remove the duplication.

In [payment method specs](https://github.com/solidusio/solidus/blob/36340a8785294b7287b03599232be8bf44de548e/core/spec/models/spree/payment_method_spec.rb#L223-L329) instead, we now have the opportunity to just use the bogus credit card instead of the anonymous classes. The downside is that we'll connect these bogus implementations to the tests, but I think it also has an advantage: if we make some changes that make these specs fail, it most likely will be the same for other payment gateways with the same interface, and at least we will notice immediately.

Please re-review now, commit-by-commit should be easier.

core/app/models/spree/payment_method/bogus_credit_card.rb
kennyadsl: This is tested now, but in the payment method specs, see the second commit. We could move these tests into the bogus own spec files, but I tend to think it has more value if we leave them together in the payment_method specs. Still open to changing my mind, though.

waiting-for-dev (approved):
Thanks, @kennyadsl, for improving this 🙏

PR #4869, opened 2023-01-24
Remove buildkite test collector until issue is fixed upstream

waiting-for-dev:
See buildkite/test-collector-ruby#160. Until the issue is fixed it's better to remove instead of supporting a fork. We can readd it later by reverting this commit.

[See workflow](https://app.circleci.com/pipelines/github/solidusio/solidus/4002/workflows/c1a72c05-ddba-4fa4-b127-176685bcf896) for an example on how it's hitting us for Ruby 3.1 & 3.2.


kennyadsl (approved):

waiting-for-dev:
Not sure why is not being reported, but the [build is green](https://app.circleci.com/pipelines/github/nebulab/solidus/753/workflows/c1da3883-85c1-4345-9003-25e1e5a3be5a).

gsmendoza (approved):

core/spec/spec_helper.rb
```diff
@@ -10,9 +10,6 @@
   SimpleCov.start('rails')
 end
 
-require 'spree/testing_support/buildkite_test_collector'
-Spree::TestingSupport::BuildkiteTestCollector.enable
```
elia, line 14: When/if we'll reinstate it we could try driving the call to enable with an environment variable, so temporary hiccups can be handled in the CircleCI settings…

elia (approved):

core/spec/spec_helper.rb
waiting-for-dev: Good call 👍

PR #4666, opened 2022-10-10
Test matrix of ruby/rails versions on different Databases

elia:
## Summary

_Continuing from #4652, most of the work done by @cpfergus1, I'm just taking this through the finish line_

- parameterize the main job
- add matrixes to better cover what we officially support
- try to be efficient and avoid combinatorial explosion (i.e. skip combos sufficiently covered by other jobs)
- this doesn't consider ruby versions older than 2.7 although the gemspec allows installing solidus with 2.5

Probably @solidusio/core-team should decide on how to interpret semver with regards to EOL dependencies such as Ruby and Rails, my suggestion would be to automatically drop support for versions that are no longer maintained.

Open to suggestions on how to streamline the matrixes 👂 

_ℹ️ The higher amount of runs increases the chance to incur into a flakey, this is not a fault of more coverage per-se but something to be aware of, and possibly motivation to fix them_

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/solidus/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [ ] I have written a thorough PR description.
- [ ] I have kept my commits small and atomic.
- [ ] I have used clear, explanatory commit messages.

The following are not always needed (~cross them out~ if they are not):

- [ ] I have added automated tests to cover my changes.
- [ ] I have attached screenshots to demo visual changes.
- [ ] I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).
- [ ] I have updated the readme to account for my changes.

PR #4880, opened 2023-01-26
Fix CI only testing with the legacy event system adapter

waiting-for-dev:
## Summary

From https://github.com/solidusio/solidus/pull/4666 all CI jobs have
been running with the legacy event system. We were defaulting the
`legacy` parameter to `0`, which we copied over to the
`USE_LEGACY_EVENTS` environment variable. However, the test suite was
checking for the mere existence of the environment variable to select
the legacy adapter and not for an actual `1` value.

We are now:

- Renaming the `legacy` parameter to `legacy_events` to be clearer.
- Making the parameter a boolean one, defaulting to false.
- Checking that `USE_LEGACY_EVENTS` is the string `"true"` before
  selecting the legacy adapter.
- Using the legacy adapter for the job on Ruby 2.7 & Rails 6.0.
- Adding `-legacy_events` to the job name when selected.

## Checklist

- ~[ ] I have documented new code [with YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md).~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -263,9 +263,9 @@ jobs:
       paperclip:
         type: boolean
         default: true
-      legacy:
-        type: string
-        default: "0"
+      legacy_events:
+        type: boolean
+        default: false
     executor:
       name: << parameters.database >>
       ruby: << parameters.ruby >>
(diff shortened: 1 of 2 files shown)


elia (approved):

kennyadsl:
Probably I'm missing something but shouldn't CI fail at 2.7 because of #4875?

waiting-for-dev:
> Probably I'm missing something but shouldn't CI fail at 2.7 because of #4875?

No, as that's for an edge case (event responding to `#to_hash`) that we were not using. I'll add a regression test there.

adammathys (approved):

PR #4883, opened 2023-01-26
Risk analysis box update

elia:
## Summary

While investigating how an order and its payments are marked as risky as part of the solidus_stripe rewrite I bumped into AVS/CCV codes.

Any modern payment provider and risk analysis tool will probably provide some structured data and a score (with thresholds) as the outcome of a risk analysis. 

This is a first step toward distancing ourselves from the ruins of solidus_gateway toward per payment-method and per-order risk analysis tools.

--- 

## Changes
- The original risk banner was showing information for the last payment, despite `Spree::Order#is_risky?` was based on having any of the payments to be considered risky → now it lists information for all risky payments
- The fieldset was changed to an expandable details/summary box → vanilla HTML FTW

image
image
image
image
image

<details>
<summary>before</summary>
image
image
image

</details>

## Checklist

- [ ] I have documented new code [with YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md).

### DIFF
--- a/backend/app/views/spree/admin/orders/_risk_analysis.html.erb
+++ b/backend/app/views/spree/admin/orders/_risk_analysis.html.erb
@@ -1,52 +1,26 @@
-<% latest_payment = @order.payments.reorder("created_at DESC").first %>
+<details id="risk_analysis">
+  <summary><%= t('spree.risk_analysis') %>: <%= t('spree.risky') %></summary>
 
-<fieldset class="no-border-bottom" id="risk_analysis">
-  <legend><%= "#{t('spree.risk_analysis')}: #{t('spree.not') unless @order.is_risky?} #{t('spree.risky')}" %></legend>
   <table>
     <thead>
+      <th><%= t('spree.payment') %></th>
       <th><%= t('spree.risk') %></th>
       <th><%= t('spree.status') %></th>
     </thead>
-    <tbody id="risk-analysis" data-hook="order_details_adjustments" class="with-border">
-      <tr class="">
-        <td><strong>
-          <%= t('spree.failed_payment_attempts') %>:
-        </strong></td>
-        <td>
-          <span class="pill pill-<%= @order.payments.failed.count > 0 ? 'warning' : 'complete' %>">
-            <%= t(
-              'spree.payments_failed_count',
-              count: @order.payments.failed.count
-            ) %>
-          </span>
-        </td>
-      </tr>
-
-      <tr>
-        <td><strong><%= t('spree.avs_response') %>:</strong></td>
-        <td>
-          <span class="pill pill-<%= latest_payment.is_avs_risky? ? 'warning' : 'complete' %>">
-            <% if latest_payment.is_avs_risky? %>
-              <%= avs_response_code[latest_payment.avs_response] %>
-            <% else %>
-              <%= t('spree.success') %>
-            <% end %>
-          </span>
-        </td>
-      </tr>
-
-      <tr>
-        <td><strong><%= t('spree.cvv_response') %>:</strong></td>
-        <td>
-          <span class="pill pill-<%= latest_payment.is_cvv_risky? ? 'warning' : 'complete' %>">
-            <% if latest_payment.is_cvv_risky? %>
-              <%= cvv_response_code[latest_payment.cvv_response_code] %>
-            <% else %>
-              <%= t('spree.success') %>
-            <% end %>
-          </span>
-        </td>
-      </tr>
+    <tbody id="risk-analysis" data-hook="order_details_adjustments"  class="with-border">
+      <% order.payments.risky.each do |payment| %>
+        <tr class="">
+          <td><%= link_to payment.number, admin_order_payment_path(order, payment) %></td>
+          <td>
+            <span class="pill pill-warning"><%= t('spree.risky') %></span>
+          </td>
+          <td>
+            <span class="pill pill-<%= payment.state %>">
+              <%= t(payment.state, scope: 'spree.payment_states') %>
+            </span>
+          </td>
+        </tr>
+      <% end %>
     </tbody>
   </table>
-</fieldset>
+</details>
--- a/backend/app/views/spree/admin/orders/_risk_analysis.html.erb
+++ b/backend/app/views/spree/admin/orders/_risk_analysis.html.erb
@@ -1,3 +1,5 @@
+<% latest_payment = @order.payments.reorder("created_at DESC").first %>
+
 <fieldset class="no-border-bottom" id="risk_analysis">
   <legend><%= "#{t('spree.risk_analysis')}: #{t('spree.not') unless @order.is_risky?} #{t('spree.risky')}" %></legend>
   <table>
--- a/backend/app/views/spree/admin/orders/cart.html.erb
+++ b/backend/app/views/spree/admin/orders/cart.html.erb
@@ -16,7 +16,7 @@
 </div>
 
 <% if @order.is_risky? %>
-  <%= render 'spree/admin/orders/risk_analysis' %>
+  <%= render 'spree/admin/orders/risk_analysis', order: @order %>
 <% end %>
 
 <div data-hook="admin_order_edit_form">
--- a/backend/app/views/spree/admin/orders/cart.html.erb
+++ b/backend/app/views/spree/admin/orders/cart.html.erb
@@ -15,8 +15,8 @@
   <%= render partial: 'spree/shared/error_messages', locals: { target: @order } %>
 </div>
 
-<% if @order.payments.exists? && @order.is_risky? %>
-    <%= render 'spree/admin/orders/risk_analysis' %>
+<% if @order.is_risky? %>
+  <%= render 'spree/admin/orders/risk_analysis' %>
 <% end %>
 
 <div data-hook="admin_order_edit_form">
--- a/backend/app/views/spree/admin/orders/cart.html.erb
+++ b/backend/app/views/spree/admin/orders/cart.html.erb
@@ -16,7 +16,7 @@
 </div>
 
 <% if @order.payments.exists? && @order.is_risky? %>
-    <%= render 'spree/admin/orders/risk_analysis', latest_payment: @order.payments.reorder("created_at DESC").first %>
+    <%= render 'spree/admin/orders/risk_analysis' %>
 <% end %>
 
 <div data-hook="admin_order_edit_form">
--- a/backend/app/views/spree/admin/payments/_list.html.erb
+++ b/backend/app/views/spree/admin/payments/_list.html.erb
@@ -22,7 +22,10 @@
   <tbody>
     <% payments.each do |payment| %>
       <tr id="<%= dom_id(payment) %>" data-hook="payments_row" class="payment vertical-middle" data-payment-id="<%= payment.id %>">
-        <td><%= link_to payment.number, spree.admin_order_payment_path(@order, payment) %></td>
+        <td>
+          <%= tag :i, class: "fa fa-warning red", title: t('spree.risky') if payment.risky? %>
+          <%= link_to payment.number, spree.admin_order_payment_path(@order, payment) %>
+        </td>
         <td><%= l(payment.created_at, format: :short) %></td>
         <td><%= payment.payment_method.name %></td>
         <td><%= payment.transaction_id %></td>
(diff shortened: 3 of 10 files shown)


backend/app/views/spree/admin/orders/_risk_analysis.html.erb
```diff
@@ -1,50 +1,28 @@
-<fieldset class="no-border-bottom" id="risk_analysis">
-  <legend><%= "#{t('spree.risk_analysis')}: #{t('spree.not') unless @order.is_risky?} #{t('spree.risky')}" %></legend>
+<% risky_payments = @order.payments.risky.reorder("created_at DESC") %>
```
waiting-for-dev, line 1: Would you have any thoughts on pushing it to the controller? Queries on the view layer are always a red flag for me, and now is a good opportunity to move it.

waiting-for-dev, line 1: Would you have any thoughts on pushing it to the controller? Queries on the view layer are always a red flag for me, and now is a good opportunity to move it.

backend/app/views/spree/admin/orders/_risk_analysis.html.erb
```diff
@@ -1,50 +1,28 @@
...
+
+<details id="risk_analysis">
```
waiting-for-dev, line 3: 🙌

backend/app/views/spree/admin/orders/cart.html.erb
```diff
@@ -11,13 +11,12 @@
...
 <% admin_breadcrumb(t('spree.cart')) %>
 
-
 <div data-hook="admin_order_edit_header">
   <%= render partial: 'spree/shared/error_messages', locals: { target: @order } %>
 </div>
 
-<% if @order.payments.exists? && @order.is_risky? %>
-    <%= render 'spree/admin/orders/risk_analysis', latest_payment: @order.payments.reorder("created_at DESC").first %>
+<% if @order.is_risky? %>
```
waiting-for-dev, line 18: For context, I checked, and that's backward compatible 👍 

```ruby
def is_risky?
  payments.risky.count > 0
end
```

When there're no payments, that'll return `false` (`0 > 0`)

core/README.md
```diff
@@ -40,7 +40,6 @@ source (e.g. `Spree::CreditCard`) using a specific payment method (e.g
 `Solidus::Gateway::Braintree`).
 * `Spree::PaymentMethod` - A base class which is used for implementing payment methods.
 * `Spree::PaymentMethod::CreditCard` - An implementation of a `Spree::PaymentMethod` for credit card payments.
-See https://github.com/solidusio/solidus_gateway/ for officially supported payment method implementations.
```
waiting-for-dev, line 43: Nice clean-up 👍

core/app/models/spree/payment.rb
```diff
@@ -127,19 +127,26 @@ def payment_source
       res || payment_method
     end
 
-    # @return [Boolean] true when this payment is risky based on address
+    # @return [Boolean] true when this payment is risky
+    def risky?
```
waiting-for-dev, line 131: Where are we using this new method? Unless I'm missing something, we're now using [the scope](https://github.com/solidusio/solidus/blob/1c4b1ffbcf08ff6d907bdb2c77699b67a331b4fe/core/app/models/spree/payment.rb#L57), which doesn't go through it. Is is there so that payment implementations can use it? 🤔

core/spec/models/spree/payment_spec.rb
```diff
@@ -1207,51 +1207,49 @@
     end
   end
 
-  describe "is_avs_risky?" do
+  describe "risky?" do
```
waiting-for-dev, line 1210: I think there're some conditions that we're not testing, like the `failed` state. However, as I said in the other comment, I'm not sure I understand why we need the method.

waiting-for-dev (commented):
Thanks, @elia; I left some comments to understand it better and a non-blocking suggestion about moving the query up.

backend/app/views/spree/admin/orders/cart.html.erb
elia: Yes, that was my assessment too 👍

core/app/models/spree/payment.rb
elia: After talking with @kennyadsl I went for not deprecating the avs/cvv risky methods and using `#risky?` in the `_gateway` partial to give a notice for risky payments and in the payment list.

core/spec/models/spree/payment_spec.rb
elia: See above, now it's being used elsewhere, and properly tested 👍 good call

elia (commented):
@waiting-for-dev did a few changes, mostly pushing details of the risk analysis to the partial that is specific to the payment method, so we can stop assuming all payments will have AVS and CVV informations returned by payment providers.

backend/app/views/spree/admin/payments/_list.html.erb
```diff
@@ -22,7 +22,10 @@
   <tbody>
     <% payments.each do |payment| %>
       <tr id="<%= dom_id(payment) %>" data-hook="payments_row" class="payment vertical-middle" data-payment-id="<%= payment.id %>">
-        <td><%= link_to payment.number, spree.admin_order_payment_path(@order, payment) %></td>
+        <td>
+          <%#= tag :i, class: "fa fa-warning red", title: t('spree.risky') if payment.risky? %>
```
waiting-for-dev, line 26: Is this line commented by error?

waiting-for-dev (approved):
@elia, it looks cohesive now 👍 Please, see my comment, I'm not sure if there's a line that is commented by error.

backend/app/views/spree/admin/orders/_risk_analysis.html.erb
```diff
@@ -1,3 +1,5 @@
+<% latest_payment = @order.payments.reorder("created_at DESC").first %>
```
kennyadsl: what if we use `latest_payment ||=` for backward compatibility? Just in case someone is using this partial with a different local variable.

backend/app/views/spree/admin/payments/_list.html.erb
kennyadsl: 👆

kennyadsl (commented):
Left a couple of questions, but overall this change works for me, thanks Elia.

backend/app/views/spree/admin/payments/_list.html.erb
elia: Yes! Thank you, I must have done it to grab the screenshot and then committed with the rest 😓

backend/app/views/spree/admin/orders/_risk_analysis.html.erb
elia: This was done in a preparatory commit, but the final version doesn't use `latest_payment` at all. As of backwards compatibility we should be sort of good since partials don't complain about extra locals.

kennyadsl:
@elia a rebase is needed here after merging #4891 🙏

backend/app/views/spree/admin/orders/_risk_analysis.html.erb
```diff
@@ -1,50 +1,26 @@
...
-          <span class="pill pill-<%= latest_payment.is_cvv_risky? ? 'warning' : 'complete' %>">
-            <% if latest_payment.is_cvv_risky? %>
-              <%= cvv_response_code[latest_payment.cvv_response_code] %>
-            <% else %>
-              <%= t('spree.success') %>
-            <% end %>
-          </span>
-        </td>
-      </tr>
+      <% @order.payments.risky.each do |payment| %>
```
gsmendoza, line 11: Nit: Have you considered moving the `@order` instance variable out of the partial? From https://andycroll.com/ruby/dont-use-instance-variables-in-partials/: 

image

core/spec/models/spree/payment_spec.rb
```diff
@@ -1231,26 +1246,17 @@
...
-
-    it "returns false if cvv_response_message == ''" do
-      payment.update_attribute(:cvv_response_message, '')
-      expect(payment.is_cvv_risky?).to eq(false)
+  describe "#is_cvv_risky?" do
+    it "returns false if cvv_response_code is 'M' or nil" do
+      ['M', nil].each do |char|
+        payment.cvv_response_code = char
+        expect(payment.is_cvv_risky?).to eq(false)
+      end
```
gsmendoza, line 1254: Nit: I'd prefer separate examples for nil and 'M' like in the original code since if the example right now fails, you'll have to debug the example to see if it failed because the character was "M" or nil.

core/spec/models/spree/payment_spec.rb
```diff
@@ -1231,26 +1246,17 @@
...
 
-    it "returns true if cvv_response_code == [A-Z], omitting D" do
-      # should use cvv_response_code helper
-      (%w{N P S U} << "").each do |char|
-        payment.update_attribute(:cvv_response_code, char)
+    it "returns true if cvv_response_code is not 'M' or nil" do
+      ['', *('A'...'M'), *('N'..'Z')].each do |char|
+        payment.cvv_response_code = char
         expect(payment.is_cvv_risky?).to eq(true)
       end
```
gsmendoza, line 1261: Nit: Similar to https://github.com/solidusio/solidus/pull/4883/files#r1094171649, I'd prefer separate examples for empty string and non-M alpha characters. For the non-M scenario, I'd usually just choose one character and call it a day :D

backend/app/views/spree/admin/orders/_risk_analysis.html.erb
elia: I'm deeply aware of this problem, [in fact have been for 10 years already](https://www.slideshare.net/elia.schito/rails-oo-views). In this case I would say it's a rather benign situation and the whole risk support is not very useful currently, but I'll pass the order from outside 👍 

BTW Rails 7.1 will allow listing accepted locals for partials, which will alleviate the biggest issue with using locals 🙌

core/spec/models/spree/payment_spec.rb
elia: I'll move the block outside of the `it` which should give us the best of both worlds 👍

backend/app/views/spree/admin/orders/_risk_analysis.html.erb
```diff
@@ -1,50 +1,26 @@
-<fieldset class="no-border-bottom" id="risk_analysis">
-  <legend><%= "#{t('spree.risk_analysis')}: #{t('spree.not') unless @order.is_risky?} #{t('spree.risky')}" %></legend>
+<details id="risk_analysis">
+  <summary><%= t('spree.risk_analysis') %>: <%= t('spree.risky') %></summary>
```
gsmendoza, line 2: As a side note, if you say the translation very fast, it sounds like "spreeesky" :laughing:

backend/app/views/spree/admin/orders/_risk_analysis.html.erb
```diff
@@ -1,50 +1,26 @@
...
-            <% end %>
-          </span>
-        </td>
-      </tr>
+    <tbody id="risk-analysis" data-hook="order_details_adjustments"  class="with-border">
+      <% order.payments.risky.each do |payment| %>
+        <tr class="">
+          <td><%= link_to payment.number, admin_order_payment_path(order, payment) %></td>
+          <td>
+            <span class="pill pill-warning"><%= t('spree.risky') %></span>
```
gsmendoza, line 15: Nit: At the moment, this column seems redundant, since it just shows "Risky" for every row. I'm guessing this column will show various degrees of riskiness in the future (Risky, Very Risky, ...)?

gsmendoza (approved):

backend/app/views/spree/admin/orders/_risk_analysis.html.erb
elia: Yes that's 100% true, in fact in the long run would make sense to turn risk assessment into a score with threshold, which is the most common solution outside and have different factors contribute to it (payments being just one part of this). 

But for now this is just a dent in that direction starting to move away from the old ways of 10 years ago 😊

kennyadsl: Good finding. Another UX/UI improvement (if we remove that redundant table column) could be moving the risky section in the right sidebar, close to some more global order information and not related to the currently open tab.

elia: @kennyadsl that would make sense, especially if paired with some order level way to convey risk information. 
You think it's fine to handle it separately or would you like to see it as part of this PR?

kennyadsl: Up to you @elia! Just let me know if you plan to do it separately and I'll proceed with merging this.

elia: @kennyadsl Yes! Please merge it as it is 🙏 
I would prefer to keep it separate and provide a good integration point for risk assessment and giving it more thought 💭

PR #4909, opened 2023-02-06
Allow providing an existing payment during the checkout

elia:
## Summary

This allows modern payment methods (relying on tokens and frontend confirmation) to have a record on solidus' side that can track events and the payment status.

Creating a payment before or during the payment step of the checkout was already possible, but resulted in an additional payment to be created before transitioning to the confirm step and invalidating all previous payments.

This way it can simply be provided via its id, and fetched from the payments currently associated with the order.

If payment methods will handle their own payment creation in the future we can distance ourselves from the credit-card coupled code and have a generic payment interface that can better serve modern payment methods.

## Checklist


core/spec/models/spree/payment_create_spec.rb
```diff
@@ -122,5 +122,39 @@ module Spree
...
+
+      it 'picks up the payment specified by the id' do
+        expect(payment_create.build).to eq(payment)
+      end
+    end
+
+    context 'with an unknown payment' do
+      let(:attributes) { { id: 123 } }
+
+      it 'picks up the payment specified by the id' do
```
waiting-for-dev, line 155: Wrong description 🙂

core/app/models/spree/payment_create.rb
```diff
@@ -24,13 +32,14 @@ def initialize(order, attributes, payment: nil, request_env: {})
...
     def build
+      @payment ||= order.payments.find(attributes[:id]) if attributes[:id]
       @payment ||= order.payments.new
       @payment.request_env = @request_env if @request_env
       @payment.attributes = @attributes
 
       if source_attributes[:wallet_payment_source_id].present?
         build_from_wallet_payment_source
-      else
+      elsif source_attributes.present?
```
waiting-for-dev, line 42: Thoughts on using a guard clause instead?

waiting-for-dev (commented):
Thanks for tackling the improvement of the payment system, @elia. That's not an easy task! 🙂 

I left a couple of minor comments. However, I'm concerned about the security repercusions of this change. If I understand well, the payment id could come from the attributes submitted by the user. I see that they are scoped from the payments belonging to the order. At the very least, we should add a test for that, ensuring that no other existent payment can be used. However, do you think it could create some mess in the case a user provides the id for a failed payment, or any other edge case that could be exploited?

core/app/models/spree/payment_create.rb
```diff
@@ -3,16 +3,24 @@
...
     #     * :wallet_payment_source_id (Integer): The id of a {WalletPaymentSource} to use
     # @param request_env [Hash] rack env of user creating the payment
     # @param payment [Payment] Internal use only. Instead of making a new payment, change the attributes for an existing one.
-    def initialize(order, attributes, payment: nil, request_env: {})
+    def initialize(order, attributes, payment: PAYMENT_NOT_PROVIDED, request_env: {})
       @order = order
-      @payment = payment
+
+      if payment != PAYMENT_NOT_PROVIDED
+        Spree::Deprecation.warn("Passing `payment:` to PaymentCreate.new is no longer supported.")
```
kennyadsl, line 19: Does this class continue to work exactly as before in this path? If yes I think this deprecation warning is misleading, seems like immediate action is needed. If we deprecate passing that parameter, we should say that its support will be removed in the future.

elia: Good call, I thought the deprecation horizon was being added by active support, changing that to use `deprecation_warning` which builds a better message:

```
DEPRECATION WARNING: Passing `payment:` to PaymentCreate.new is deprecated and will be removed from Solidus 4.0 (called from <main> at bin/rails:4)
```

core/spec/models/spree/payment_create_spec.rb
elia: 😓 fixing thanks

core/app/models/spree/payment_create.rb
elia: I gave it a try but wasn't really happy with the result, and preferred to have that if/elsif as the central body of the method, but thanks for suggesting it 👍

elia:
> If I understand well, the payment id could come from the attributes submitted by the user. I see that they are scoped from the payments belonging to the order. 

Yes, this is mainly a tool for payment methods to create the payment before (or while) the payment page is shown.

> At the very least, we should add a test for that, ensuring that no other existent payment can be used. 

I'll add one 👍 

> I'm concerned about the security repercussions of this change […] do you think it could create some mess in the case a user provides the id for a failed payment, or any other edge case that could be exploited?

Currently Solidus allows completing an order even with a payment with a status of `:checkout` and the enforcement is left to the frontend. 

I think being able to provide existing payments associated to the order shouldn't change much, except not creating a new one, since the existing payments are already associated and count toward the balance.

@waiting-for-dev

core/app/models/spree/payment_create.rb
```diff
@@ -3,16 +3,24 @@
...
   class PaymentCreate
+    PAYMENT_NOT_PROVIDED = Object.new.freeze
+
     # @param order [Order] The order for the new payment
     # @param attributes [Hash,ActionController::Parameters] attributes which are assigned to the new payment
     #   * :payment_method_id Id of payment method used for this payment
     #   * :source_attributes Attributes used to build the source of this payment. Usually a {CreditCard}
     #     * :wallet_payment_source_id (Integer): The id of a {WalletPaymentSource} to use
     # @param request_env [Hash] rack env of user creating the payment
     # @param payment [Payment] Internal use only. Instead of making a new payment, change the attributes for an existing one.
```
kennyadsl, line 14: If this is deprecated now, we should mark it as such in the YARD documentation

core/app/models/spree/payment_create.rb
```diff
@@ -3,16 +3,24 @@
 module Spree
   # Service object for creating new payments on an Order
   class PaymentCreate
+    PAYMENT_NOT_PROVIDED = Object.new.freeze
+
     # @param order [Order] The order for the new payment
     # @param attributes [Hash,ActionController::Parameters] attributes which are assigned to the new payment
     #   * :payment_method_id Id of payment method used for this payment
```
kennyadsl, line 10: What about adding documentation for the new `id` attribute, required in case we want to use an existing payment?

core/spec/models/spree/payment_create_spec.rb
```diff
@@ -122,5 +122,39 @@ module Spree
...
+        end
+      }
+      let(:attributes) { { id: payment.id } }
+
+      it 'picks up the payment specified by the id' do
+        expect(payment_create.build).to eq(payment)
+      end
+    end
+
+    context 'with an unknown payment' do
```
kennyadsl, line 152: What if we add a test of a payment coming from another order as well?

core/app/models/spree/payment_create.rb
```diff
@@ -3,16 +3,24 @@
...
     #   * :source_attributes Attributes used to build the source of this payment. Usually a {CreditCard}
     #     * :wallet_payment_source_id (Integer): The id of a {WalletPaymentSource} to use
     # @param request_env [Hash] rack env of user creating the payment
     # @param payment [Payment] Internal use only. Instead of making a new payment, change the attributes for an existing one.
-    def initialize(order, attributes, payment: nil, request_env: {})
+    def initialize(order, attributes, payment: PAYMENT_NOT_PROVIDED, request_env: {})
       @order = order
-      @payment = payment
+
+      if payment != PAYMENT_NOT_PROVIDED
```
kennyadsl, line 18: Why not just `if payment.present?` here?

kennyadsl (commented):
Second pass!

core/app/models/spree/payment_create.rb
kennyadsl: Got it now, thanks @spaghetticode (https://github.com/solidusio/solidus/pull/3338#discussion_r1129169244)

elia:
Was split into other PRs

PR #4950, opened 2023-02-20
Keep all the LogEntry serialization logic in one place

elia:
## Summary

Moving all the serialization/deserialization of Spree::LogEntry details to one place and using a consistent proxy attribute.

This begs for [proper ActiveRecord serialization](https://edgeapi.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize) but before doing that we need to think through the migration path for existing stores.

This change is intentionally minimal, enough to allow all writers of log entries to use the same serialization method.

## Checklist

### DIFF
--- a/core/app/models/spree/payment/processing.rb
+++ b/core/app/models/spree/payment/processing.rb
@@ -206,7 +206,7 @@ module Spree
       end
 
       def record_response(response)
-        log_entries.create!(details: response.to_yaml)
+        log_entries.create!(parsed_details: response)
       end
 
       def protect_from_connection_error
--- a/core/spec/models/spree/log_entry_spec.rb
+++ b/core/spec/models/spree/log_entry_spec.rb
@@ -4,7 +4,7 @@ require 'rails_helper'
 
 RSpec.describe Spree::LogEntry, type: :model do
   describe '#parsed_details' do
-    it 'allow aliases by default' do
+    it 'allows aliases by default' do
       x = []
       x << x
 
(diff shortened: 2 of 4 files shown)


core/app/models/spree/payment/processing.rb
```diff
@@ -206,7 +206,7 @@ def handle_response(response)
       end
 
       def record_response(response)
-        log_entries.create!(details: response.to_yaml)
+        log_entries.create!(parsed_details: response)
```
waiting-for-dev, line 209: I had to double-check that `create!` will accept the `parsed_details` attribute even if it's not a table field. Non-blocking, but what do you think if we make the implementation and interface clearer by creating another method like `Spree::LogEntry.log` or similar?

core/spec/models/spree/log_entry_spec.rb
```diff
@@ -53,4 +53,16 @@
...
     end
   end
+
+  describe '#parsed_details=' do
+    it 'serializes the provided value to YAML' do
+      value = double("an object", to_yaml: "---\nfoo: bar")
+
+      log_entry = described_class.new(parsed_details: value)
+
+      expect(value).to have_received(:to_yaml)
```
waiting-for-dev, line 63: Also non-blocking, but it's easy to decouple the test from the implementation by using any Ruby object as fixture (as `Object#to_yaml` is part of core). Thoughts?

waiting-for-dev (approved):
Thanks @elia, that's a nice improvements. I left two suggestions to your judgment 🙂

elia:
@waiting-for-dev I added a few more commits, because despite the change I ended up with an exception in solidus_stripe while trying to parse a serialized hash that had symbol keys. So now we'll also `safe_dump` when using `parsed_details=`. We don't have backward compatibility concerns as `parsed_details=` is a new method.

I adapted all the tests for `#parsed_details` and shared the error handling between the two methods. 

> creating another method like Spree::LogEntry.log or similar?

I think this is a great idea, although the interface should probably go on the association or on the associated object. I'd say this is beyond the scope of what I meant to fix with this PR 😅

waiting-for-dev (approved):

kennyadsl (approved):
Nice, thanks @elia!

PR #4954, opened 2023-02-23
Ensure target shipments are evaluated in order of creation (fix flakey)

elia:
## Summary

This solves a long standing flakey spec which was uncovering an actual issue with the method being dependent on the implicit db order and how it returned records.

I have added `:id` to the sorting to disambiguate ordering of shipments were created quickly one after the other and ended up having the same creation date (e.g. in specs).

Adding `.reverse` after `order.shipments.order(:created_at, :id)` is consistently reproducing the failure.

## Checklist

### DIFF
--- a/core/app/models/spree/order_inventory.rb
+++ b/core/app/models/spree/order_inventory.rb
@@ -57,7 +57,7 @@ module Spree
     # first unshipped that's leaving from a stock_location that stocks this variant, with availability check
     # first unshipped that's leaving from a stock_location that stocks this variant, without availability check
     def determine_target_shipment(quantity)
-      potential_shipments = order.shipments.select(&:ready_or_pending?)
+      potential_shipments = order.shipments.order(:created_at, :id).select(&:ready_or_pending?)
 
       potential_shipments.detect do |shipment|
         shipment.include?(variant)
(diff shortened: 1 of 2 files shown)


waiting-for-dev (approved):
That's the way to go! 🚀

core/app/models/spree/order_inventory.rb
```diff
@@ -57,7 +57,7 @@ def remove(quantity, shipment = nil)
     # first unshipped that's leaving from a stock_location that stocks this variant, with availability check
     # first unshipped that's leaving from a stock_location that stocks this variant, without availability check
     def determine_target_shipment(quantity)
-      potential_shipments = order.shipments.select(&:ready_or_pending?)
+      potential_shipments = order.shipments.order(:created_at, :id).select(&:ready_or_pending?)
```
kennyadsl, line 60: I am a bit worried about how this can change the shipment used at the end of the checkout, when there are multiple available. Can you please help me understand why did you pick specifically this order?

kennyadsl (commented):
Thanks! Left a question, I hope we can merge this soon!

core/app/models/spree/order_inventory.rb
elia: Assuming the spec is describing the desired behavior, the issue was clearly reproducible when reversing the order of creation. I think most databases will fetch records in order of creation when no `ORDER BY` clause is provided, but not all of them (see [this answer](https://stackoverflow.com/a/60582276) for example).

By explicitly requesting the creation order fixed the spec, and reversing it consistently made it fail.

The order is also disambiguated by the adding `:id` for when, like in tests, the `:created_at` time of two records can be the same because of lower date-time precision on the database.

kennyadsl (approved):

PR #4957, opened 2023-02-27
Copy new migrations as part of the update task

kennyadsl:
## Summary
Running `bin/rails solidus:update`, we are not currently copying the new migrations introduced with the new Solidus version. This new install step will copy migrations and it's left to the user to run `db:migrate`, before checking them.

This commit also updates some copy of the task to reflect the new content of the file.

## Checklist

### DIFF
--- a/core/lib/generators/solidus/update/update_generator.rb
+++ b/core/lib/generators/solidus/update/update_generator.rb
@@ -8,7 +8,7 @@ module Solidus
   class UpdateGenerator < ::Rails::Generators::Base
     FROM = Spree.previous_solidus_minor_version
 
-    desc 'Generates a new initializer to preview the new defaults for current Solidus version'
+    desc 'Generates a new initializer to preview the new defaults for current Solidus version and copy new migrations'
 
     source_root File.expand_path('templates', __dir__)
 
@@ -40,7 +40,7 @@ module Solidus
     def create_new_defaults_initializer
       previous_version_prompt = options[:previous_version_prompt]
       return if previous_version_prompt && !yes?(<<~MSG, :red)
-        The update process is only supported if you are coming from version #{FROM}. If this is not the case, please, skip it and update your application to use Solidus #{FROM} before retrying.
+        The default preferences update process is only supported if you are coming from version #{FROM}. If this is not the case, please, skip it and update your application to use Solidus #{FROM} before retrying.
         If you are confident you want to upgrade from a previous version, you must rerun the generator with the "--from={OLD_VERSION}" argument.
         Are you sure you want to continue? (y/N)
       MSG
@@ -55,6 +55,11 @@ module Solidus
 
       template 'config/initializers/new_solidus_defaults.rb.tt',
                File.join(options[:initializer_directory], "#{options[:initializer_basename]}.rb")
+    end
+
+    def install_migrations
+      say_status :copying, "migrations"
+      rake 'spree:install:migrations'
     end
 
     def print_message


core/lib/generators/solidus/update/update_generator.rb
```diff
@@ -40,7 +40,7 @@ class UpdateGenerator < ::Rails::Generators::Base
     def create_new_defaults_initializer
       previous_version_prompt = options[:previous_version_prompt]
       return if previous_version_prompt && !yes?(<<~MSG, :red)
-        The update process is only supported if you are coming from version #{FROM}. If this is not the case, please, skip it and update your application to use Solidus #{FROM} before retrying.
+        The defaults update process is only supported if you are coming from version #{FROM}. If this is not the case, please, skip it and update your application to use Solidus #{FROM} before retrying.
```
waiting-for-dev, line 43:
```suggestion
        The default update process is only supported if you are coming from version #{FROM}. If this is not the case, please, skip it and update your application to use Solidus #{FROM} before retrying.
```

core/lib/generators/solidus/update/update_generator.rb
```diff
@@ -57,6 +57,11 @@ def create_new_defaults_initializer
                File.join(options[:initializer_directory], "#{options[:initializer_basename]}.rb")
     end
 
+    def install_migrations
+      say_status :copying, "migrations"
+      rake 'railties:install:migrations'
```
waiting-for-dev, line 62: What about running `bin/rails railties:install:migrations FROM=spree` so that we don't get migrations from other engines?

waiting-for-dev (commented):
Thanks, @kennyadsl!

Did you try to add a test for the new behavior? That generator is [tested](https://github.com/solidusio/solidus/blob/master/core/spec/lib/generators/solidus/update/update_generator_spec.rb) and it'd be good to add it. However, if that's going to require a complex setup (which I'm afraid could be the case) I'm also ok with not doing that.

kennyadsl:
I didn't remember we have tests for that task, thanks for pointing out. I'll definitely try!

core/lib/generators/solidus/update/update_generator.rb
kennyadsl: That's a valid concern, and I was thinking about it, but if there are pending migrations from other engines, why shouldn't we bring them in? I couldn't think of scenarios when it might be a problem.

waiting-for-dev: I think it's a matter of expectations. The Solidus update task should only deal with its stuff. We don't know in which scenario that task could be run, so I think it's good if we don't mess with other engines. Users might want to go step by step for any reason. What do you think?

kennyadsl:
@waiting-for-dev I'm losing any hope that we can test copying the migrations. My only idea was to remove the last existing migration from the dummy app in the test setup and then run the update task to see it generated again. But this won't work because the dummy app actually doesn't hold any migration; [they are used directly from the engine, injected programmatically in the App via `config.paths['db/migrate']`](https://github.com/solidusio/solidus/blob/aa3063ef7ae577ba0d1b1fdcb5d287445da353e7/core/lib/spree/testing_support/dummy_app.rb#L110-L119).

core/lib/generators/solidus/update/update_generator.rb
kennyadsl: Changed to "The default preferences update process..". Thoughts?

core/lib/generators/solidus/update/update_generator.rb
kennyadsl: You are right, done!

core/lib/generators/solidus/update/update_generator.rb
```diff
@@ -57,6 +57,11 @@ def create_new_defaults_initializer
                File.join(options[:initializer_directory], "#{options[:initializer_basename]}.rb")
     end
 
+    def install_migrations
+      say_status :copying, "migrations"
+      rake 'spree:install:migrations'
```
waiting-for-dev, line 62: Using `spree:` as the namespace is even better than using the `FROM` env variable :clap: :clap:

waiting-for-dev (commented):
> @waiting-for-dev I'm losing any hope that we can test copying the migrations. My only idea was to remove the last existing migration from the dummy app in the test setup and then run the update task to see it generated again. But this won't work because the dummy app actually doesn't hold any migration; [they are used directly from the engine, injected programmatically in the App via config.paths['db/migrate']](https://github.com/solidusio/solidus/blob/aa3063ef7ae577ba0d1b1fdcb5d287445da353e7/core/lib/spree/testing_support/dummy_app.rb#L110-L119).

No worries. The only clean way I can think would be to have the `install:migrations` accept an injectable directory and use `tmp/` for the tests, but that comes from Rails.

Nice work!!

waiting-for-dev (approved):
Oh, sorry, I thought I had already approve this one :+1:

core/lib/generators/solidus/update/update_generator.rb
kennyadsl: I randomly faced a similar problem in another context and realized how scoping the copy of migrations only for the spree namespace might be an issue: if solidus depends on another Rails engine, and solidus bumps it in a new version, and the new required version of that engine has new migrations, users will need to run the install migrations command for that engine or the system won't work. Do you think it's a realistic scenario?

adammathys (approved):

core/lib/generators/solidus/update/update_generator.rb
waiting-for-dev: However, Solidus doesn't depend on any Rails engine with migrations, right? If so, we should then be explicit and copy the migrations from its namespace. What do you think?

PR #4996, opened 2023-03-28
improve strictness of spree_line_items

BenMorganIO:
## Summary

The goal of this migration is to add `NOT NULL` where applicable and to add foreign key constraints. The foreign keys on_delete are copies of what the `dependent` option would likely replicate.

## Checklist

### DIFF
--- a/core/db/migrate/20230328042835_improve_strictness_of_spree_line_items.rb
+++ b/core/db/migrate/20230328042835_improve_strictness_of_spree_line_items.rb
@@ -1,0 +1,42 @@
+class ImproveStrictnessOfSpreeLineItems < ActiveRecord::Migration[7.0]
+  def up
+    change_table(:spree_line_items, bulk: true) do |t|
+      t.change :variant_id, :integer, null: false
+      t.change :order_id, :integer, null: false
+
+      t.change :created_at, :datetime, null: false
+      t.change :updated_at, :datetime, null: false
+
+      t.change :adjustment_total, :decimal, precision: 10, scale: 2, default: '0.0', null: false
+      t.change :additional_tax_total, :decimal, precision: 10, scale: 2, default: '0.0', null: false
+      t.change :promo_total, :decimal, precision: 10, scale: 2, default: '0.0', null: false
+    end
+
+    add_foreign_key :spree_line_items, :spree_orders, column: :order_id,
+      on_delete: :cascade, on_update: :cascade
+
+    add_foreign_key :spree_line_items, :spree_variants, column: :variant_id,
+      on_delete: :restrict, on_update: :cascade
+
+    add_foreign_key :spree_line_items, :spree_tax_categories, column: :tax_category_id,
+      on_delete: :restrict, on_update: :cascade
+  end
+
+  def down
+    remove_foreign_key :spree_line_items, :spree_orders, column: :order_id
+    remove_foreign_key :spree_line_items, :spree_variants, column: :variant_id
+    remove_foreign_key :spree_line_items, :spree_tax_categories, column: :tax_category_id
+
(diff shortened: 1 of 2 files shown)


tvdeyen (commented):
I like that change, but we should at least give owners of very old stores some hints, what to do if the migration fail for them (mostly concerned about deleted variants)

core/db/migrate/20230328042835_improve_strictness_of_spree_line_items.rb
```diff
@@ -0,0 +1,42 @@
+class ImproveStrictnessOfSpreeLineItems < ActiveRecord::Migration[7.0]
```
kennyadsl, line 1:
```suggestion
class ImproveStrictnessOfSpreeLineItems < ActiveRecord::Migration[5.2]
```

kennyadsl: ☝️ this will likely make the test pass on all Rails versions.

tvdeyen: I know that we do this in all our migrations, but I want to suggest that we do something like this

```rb
class ImproveStrictnessOfSpreeLineItems < ActiveRecord::Migration[ActiveRecord::Migration.current_version]
```

to make sure that new Stores get all features (like `t.bigint`, `NOT NULL` on timestamps, etc.) for later Rails versions.

To make it easier we could introduce a `Spree::Migration` base class.

kennyadsl: The only problem with this approach is that we won't be sure of how the database of users looks like anymore. A middle-ground could be using the last version of Rails that we support (currently 6.0) for new migrations?

kennyadsl: I mean, but correct me if I'm wrong, we won't know which migrations have been executed on different stores, depending on the version of Rails they are using. This way, we could have a future migration (or application code) that will need conditional paths depending on the Rails version used when the migration has been executed.

BenMorganIO: We could just do a large migration and set all the timestamps as `not null`. That seems to be the appropriate solution. Then we could do another migration later where we make all the primary/foreign keys bigints.

BenMorganIO:
@tvdeyen @kennyadsl I think this should be reserved for a major version number change and it can be highlighted in the changelog. I think we should create several migrations that focus on what we want to improve:

- `NOT NULL` changes only on the timestamps
- `NOT NULL` changes on non-timestamp columns
- `FOREIGN KEY CONSTRAINT` for all the missing relations
- `bigint` for primary key/foreign key columns

core/db/migrate/20230328042835_improve_strictness_of_spree_line_items.rb
elia: This should be fine after merging https://github.com/solidusio/solidus/pull/5012

waiting-for-dev:
We should write very detailed release instructions about how to check if a store will be affected by the change. Something like we did in https://github.com/solidusio/solidus/pull/4851. The text would be copied to the corresponding upgrade instructions on our guides.

waiting-for-dev:
Rebasing from master and marking a test as pending, as we can't simulate an invalid line item with this change

kennyadsl:
Closing this for the same reasons we did close https://github.com/solidusio/solidus/pull/2036. Thanks again for the contribution @BenMorganIO; I'm sure it will come the day we will be able to improve the strictness of our db, and that day we will resume this PR considering your proposal solution. 

🙏

PR #4999, opened 2023-03-29
[backend] Add dark mode to the backend UI

MassimilianoLattanzio:
## Summary

_Incorporates #5113_

_Compation SolidusAdmin PR #5511_

This PR aims to add a dark theme switcher for the Solidus backend. The theme can be switched using a new select added in the left sidebar, and the choice is saved in local-storage.

| System in dark mode | System in light mode |
|--------|--------|
| image | image |
| image | image | 
| image | image |

| System in dark mode (new navbar) | System in light mode (new navbar) |
|--------|--------|
| image | image |
| image | image |
| image | image|

This is just an easy way to invert theme colors as suggested by @elia following [this StackOverflow example](https://stackoverflow.com/questions/17741629/how-can-i-invert-color-using-css#21106143).

## Checklist

### DIFF
--- a/backend/app/views/spree/admin/shared/_theme_selection.html.erb
+++ b/backend/app/views/spree/admin/shared/_theme_selection.html.erb
@@ -1,0 +1,21 @@
+<% theme_options_for_select = Spree::Backend::Config.themes.keys.map { |theme| [theme.to_s.humanize, theme] }.sort %>
+
+<%= form_tag(admin_set_theme_path(format: :html), method: :put, style: "width: 100%;", class: "light-only") do %>
+  <%= hidden_field_tag :system_theme, :light %>
+  <label class="admin-navbar-selection">
+    <i class="fa fa-sun-o fa-fw" title="<%= I18n.t('spree.choose_dashboard_theme') %>"></i>
+    <select name="switch_to_theme" class="custom-select fullwidth" onchange="this.form.requestSubmit()">
+      <%= options_for_select(theme_options_for_select, session[:admin_light_theme] || Spree::Backend::Config.theme) %>
+    </select>
+  </label>
+<% end %>
+
+<%= form_tag(admin_set_theme_path(format: :html), method: :put, style: "width: 100%;", class: "dark-only") do %>
+  <%= hidden_field_tag :system_theme, :dark %>
+  <label class="admin-navbar-selection">
+    <i class="fa fa-moon-o fa-fw" title="<%= I18n.t('spree.choose_dashboard_theme') %>"></i>
+    <select name="switch_to_theme" class="custom-select fullwidth" onchange="this.form.requestSubmit()">
+      <%= options_for_select(theme_options_for_select, session[:admin_dark_theme] || Spree::Backend::Config.dark_theme) %>
+    </select>
+  </label>
+<% end %>
--- a/backend/app/assets/config/solidus_backend_manifest.js
+++ b/backend/app/assets/config/solidus_backend_manifest.js
@@ -1,9 +1,7 @@
 //= link_tree ../images
+//= link_tree "../stylesheets/spree/backend/themes" .css
 
 //= link spree/backend/all.js
 //= link spree/backend/all.css
 
 //= link solidus_admin/select2_locales
-
-//= link spree/backend/themes/classic.css
-//= link spree/backend/themes/solidus_admin.css
--- a/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/_navigation.scss
@@ -301,7 +301,8 @@ nav.menu {
 }
 
 .admin-navbar-selection {
-  margin: 0 1.5em;
+  padding: 0 1.5em;
+  width: 100%;
   position: relative;
   overflow: hidden;
   display: inline-flex;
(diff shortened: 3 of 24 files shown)


backend/app/assets/stylesheets/spree/backend/globals/_dark.scss
```diff
@@ -0,0 +1,14 @@
...
+  filter: invert(85%) hue-rotate(180deg);
+
+  img {
+    filter: invert(100%) hue-rotate(-180deg);
+  }
+
+  .brand-link img {
+    filter: invert(0%) hue-rotate(0deg);
+  }
+}
```
elia, line 14: nit: Missing EOF newline

backend/app/assets/stylesheets/spree/backend/globals/_dark.scss
```diff
@@ -0,0 +1,14 @@
+html.dark-mode {
```
elia, line 1: Thoughts on using `prefers-color-scheme` instead?

```css
@media (prefers-color-scheme: dark) {
  ...
}
```

Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

Looks like you can also use the `color-scheme` property to force a scheme on a particular element: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme

This means that it could follow the system by default and allow opt-in to force it to be either way via JS by setting that property on `<html>`.

MassimilianoLattanzio:
Some additional notes:
- There is an issue with the images using the dark theme that look darker than the relative ones in the default theme.
- The user preference can be saved in the DB instead of the cookie. Otherwise, whenever the user logs out will have to switch the theme again.
- Probably, the best way to add a new theme is to create a new bootstrap `_variables_overrides` file, but it will take a lot of time. We can use this way if the solution in this PR needs to be better.

backend/app/views/spree/admin/shared/_theme_selection.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="admin-theme-selection">
+  <select class="js-theme-selection custom-select fullwidth">
+    <option value="system"><%= t("spree.admin.themes.system") %></option>
+    <option value="light"><%= t("spree.admin.themes.light") %></option>
+    <option value="dar"><%= t("spree.admin.themes.dark") %></option>
```
MassimilianoLattanzio, line 5: @elia there is a typo on the option value

elia:
CI is failing because of a bad cache key fallback, fixed by https://github.com/solidusio/solidus/pull/5005

elia (approved):

kennyadsl:
I noticed this little glitch when the sidebar is minimized: 

Screenshot 2023-04-12 at 07 31 37@2x

This is already a bit of a mess, but I think we could improve it slightly by using the same approach we have with the other dropdown (language). Can you please take a look @MassimilianoLattanzio?

MassimilianoLattanzio:
> I noticed this little glitch when the sidebar is minimized:
> 
> Screenshot 2023-04-12 at 07 31 37@2x
> 
> This is already a bit of a mess, but I think we could improve it slightly by using the same approach we have with the other dropdown (language). Can you please take a look @MassimilianoLattanzio?

[This commit](https://github.com/solidusio/solidus/pull/4999/commits/0379cb3354aecd119244ca4f686ca87717f5fc89) fixes this issue. Here is the result:

Screenshot 2023-04-12 at 17 19 03

MassimilianoLattanzio:
I also added [another commit](https://github.com/solidusio/solidus/pull/4999/commits/8abf73845ab2cb487157fde6efd9bbaf811949f9) to fix an issue with hidden icons on the minimized menu that exceeds the window height. Example:

Screenshot 2023-04-12 at 16 54 33

kennyadsl:
@MassimilianoLattanzio I think this still needs a rebase, there's a fixup commit in the list.

MassimilianoLattanzio:
> @MassimilianoLattanzio I think this still needs a rebase, there's a fixup commit in the list.

Done!

kennyadsl:
@MassimilianoLattanzio thanks for you contributions here, but I think we are not going to merge this right now. We are in the process of starting a complete redesign of the admin, and adding Dark Mode now would influence design processes and decisions that we don't want to anticipate now. 

We will keep this on hold and use as a reference when it will be a good time to add this to the new admin. I hope this makes sense fo you as well, and in the meantime thanks again for your great work!

MassimilianoLattanzio:
> @MassimilianoLattanzio thanks for you contributions here, but I think we are not going to merge this right now. We are in the process of starting a complete redesign of the admin, and adding Dark Mode now would influence design processes and decisions that we don't want to anticipate now.
> 
> We will keep this on hold and use as a reference when it will be a good time to add this to the new admin. I hope this makes sense fo you as well, and in the meantime thanks again for your great work!

Sure! No problem at all! Makes sense to me.

Hope this can help other people and also for future updates of the admin UI.

waiting-for-dev (approved):
:night_with_stars:

elia:
In order to avoid having to support changes after the release this is on hold until the new admin UI and dark-mode implementation will be defined (e.g. cookie vs session vs db).

rainerdema (approved):

PR #5008, opened 2023-04-11
Deprecate `Spree::Payment` offsets

waiting-for-dev:
## Summary

The old system for refunds was removed in [2014](https://github.com/spree/spree/pull/4883).

We deprecate the `.offset_payment` scope, which is in turn used by the `.offsets` association. We inline its usage on a couple of methods until we perform the full removal.

## Checklist

### DIFF
--- a/core/app/models/spree/payment.rb
+++ b/core/app/models/spree/payment.rb
@@ -46,7 +46,18 @@ module Spree
     scope :from_credit_card, -> { where(source_type: 'Spree::CreditCard') }
     scope :with_state, ->(state) { where(state: state.to_s) }
     # "offset" is reserved by activerecord
-    scope :offset_payment, -> { where("source_type = 'Spree::Payment' AND amount < 0 AND state = 'completed'") }
+    # TODO: When removing the method we can also:
+    #   - Remove the `.offsets` association
+    #   - Remove the `#offsets_total` method
+    #   - Remove offsets count from the `#credit_allowed` method
+    #   - Remove offsets check from `Spree::Order#has_non_reimbursement_related_refunds?
+    def self.offset_payment
+      Spree::Deprecation.warn <<~MSG
+        `Spree::Payment offsets` are deprecated. Use the refund system (`Spree::Refund`) instead.
+      MSG
+
+      where("source_type = 'Spree::Payment' AND amount < 0 AND state = 'completed'")
+    end
 
     scope :checkout, -> { with_state('checkout') }
     scope :completed, -> { with_state('completed') }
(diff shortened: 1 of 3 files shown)


core/app/models/spree/payment.rb
```diff
@@ -17,6 +17,14 @@ class Payment < Spree::Base
     belongs_to :payment_method, -> { with_discarded }, class_name: 'Spree::PaymentMethod', inverse_of: :payments, optional: true
 
     has_many :offsets, -> { offset_payment }, class_name: "Spree::Payment", foreign_key: :source_id
```
kennyadsl, line 19: What about also deprecating the `offset_payment` scope?

core/app/models/spree/payment.rb
```diff
@@ -101,7 +109,10 @@ def offsets_total
     # @return [BigDecimal] the amount of this payment minus the offsets
     #   (old-style refunds) and refunds
     def credit_allowed
-      amount - (offsets_total.abs + refunds.sum(:amount))
+      amount - (
+        (Spree::Deprecation.silence { offsets_total.abs }) +
```
kennyadsl, line 113: What if we avoid the deprecation silencing and use a more verbose `self.class.where("source_type = 'Spree::Payment' AND amount < 0 AND state = 'completed' AND source_id = '#{self.id}'").pluck(:amount).sum`, which should do the same?

kennyadsl (commented):
Left a couple of questions to better understand if we have alternatives. Thanks!

core/app/models/spree/payment.rb
waiting-for-dev: Good call. I deprecated `.offset_payment`, which is in turn use by the `.offsets` association.

core/app/models/spree/payment.rb
```diff
@@ -101,7 +111,10 @@ def offsets_total
     # @return [BigDecimal] the amount of this payment minus the offsets
     #   (old-style refunds) and refunds
     def credit_allowed
-      amount - (offsets_total.abs + refunds.sum(:amount))
+      amount - (
+        (self.class.where("source_type = 'Spree::Payment' AND amount < 0 AND state = 'completed' AND source_id = ?", id).sum(:amount)).abs +
```
kennyadsl, line 115: In this case, maybe a variable makes sense? We could add a comment to easily recall which part of the method we need to remove when `offset_payment` will be gone.

kennyadsl (approved):
Left another small note for our future selves, but I'm also ok as is.

core/app/models/spree/payment.rb
waiting-for-dev: Given that I needed to repeat the inline on `Spree::Order#has_non_reimbursement_related_refunds?`, and that the code should really be very short-lived (we should remove it in #4989), I'd prefer to keep it simple if you think it's ok :slightly_smiling_face:

rainerdema (approved):
💯

core/app/models/spree/payment.rb
elia: I had the same suggestion as alberto, but also fine with this give its expected short life

elia (approved):

PR #5009, opened 2023-04-12
Add a tooltip for default currency in store settings

kennyadsl:
## Summary

This PR comes from some support requests we had in Slack, highlighting some confusion when a currency switch is required. I hope this will mitigate that confusion.  By the way, this is just a little help, a proper page on the guide would be a good next step.

Screenshot 2023-04-12 at 11 27 55@2x

## Checklist

### DIFF
--- a/core/config/locales/en.yml
+++ b/core/config/locales/en.yml
@@ -1644,6 +1644,7 @@ en:
         available_locales: This determines which locales are available for your customers to choose from in the storefront.
         cart_tax_country_iso: 'This determines which country is used for taxes on carts (orders which don''t yet have an address).<br> Default: None.'
         code: An identifier for your store. Developers may need this value if you operate multiple storefronts.
+        default_currency: This determines which currency will be used for the storefront's product prices. Please, be aware that changing this configuration, only products that have prices in the selected currency will be listed on your storefront. <br><br>This setting won't change the default currency used when you create a product. For that, only the global `Spree::Config.currency` is taken into account.
       spree/tax_category:
         is_default: When checked, this tax category will be selected by default when creating new products or variants.
       spree/tax_rate:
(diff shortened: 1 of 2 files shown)


core/config/locales/en.yml
```diff
@@ -1644,6 +1644,7 @@ en:
         available_locales: This determines which locales are available for your customers to choose from in the storefront.
         cart_tax_country_iso: 'This determines which country is used for taxes on carts (orders which don''t yet have an address).<br> Default: None.'
         code: An identifier for your store. Developers may need this value if you operate multiple storefronts.
+        default_currency: This determines which currency will be used to retrieve product prices for this store. If you want to change it globally, set `Spree::Config.currency` instead. <br><br>Please, be aware that only products that have prices in the selected currency will be listed.
```
waiting-for-dev, line 1647: I'm still a bit confused. What do we mean by "retrieve"? Do we mean from the frontend (although we should be agnostic)? Is it also true from the API? Do we mean the default product searcher? On top of that, I think (not 100% sure), that `Spree::Config.currency` is the only one used when creating products on the backend.

waiting-for-dev (commented):
Thanks, @kennyadsl! That confusion also hit me in the past.

core/config/locales/en.yml
kennyadsl: It's used by the `current_pricing_options` helper, which is used by all frontends, API and base searcher, so I think it's safe to assume that it's a valid general rule for any "storefront" instance (in fact, it is available in the store preference). 

It seems to be a different thing for the backend, and probably `Spree::Config.currency` is used to set the default currency because the same product can be available in different stores.

If you have any suggestions on how to make it clearer, please help me! 🙂

kennyadsl: Rephrased, what about now?

waiting-for-dev (approved):
:heart:

core/config/locales/en.yml
waiting-for-dev: Cool!

rainerdema (approved):

PR #5040, opened 2023-04-24
Get fresh data for update_positions

julienanne:
## Summary

In the backend section, from Ref https://github.com/solidusio/solidus/commit/248b39f9b5c797fd0836277c2fdb7fe50380f045 we have a bug depending of the order choose for many objects.
Params will be sorted by keys (I guess) and so we update records in a certain order (position are updated for more than one record, `increment_all` and `decrement_all` methods of `acts_as_list gem`).
But we take records from an "not fresh" request/array and so, the after update callback of acts_as_list will not be executed as we want.
The N+1 introduce seems to be the way that acts_as_list want to be used, but in fact I think acts_as_list should give an other method than only `set_list_position`, a more "global" method will be appreciate.

Again my commentary in code is improvable I think.

And if performance is more important than this bug feel free to close this PR ;)

This change need to be backported until the version 3.1.0 if I'm right ;)

Hope it can help.

## Checklist

### DIFF
--- a/backend/app/controllers/spree/admin/resource_controller.rb
+++ b/backend/app/controllers/spree/admin/resource_controller.rb
@@ -79,12 +79,16 @@ class Spree::Admin::ResourceController < Spree::Admin::BaseController
   def update_positions
     ActiveRecord::Base.transaction do
       positions = params[:positions]
-      records = model_class.where(id: positions.keys).to_a
 
       positions.each do |id, index|
-        # To permit the use of UUID as id, we compare models id in database
-        # and in params under the string representation instead of integer
-        records.find { |r| r.id.to_s == id }&.set_list_position(index)
+        # Yes here there is a N+1 but acts_as_list use after update callback
+        # so we can't keep not reloaded data without create a bug
+        # (spec : backend/spec/controllers/spree/admin/resource_controller_spec.rb)
+        # "with take care of acts_as_list's after update callback" test
+        #
+        # TODO : create a global set_list_position on all concerned objects
+        # maybe in the acts_as_list gem
+        model_class.where(id: id).first&.set_list_position(index)
       end
     end
 
--- a/backend/spec/controllers/spree/admin/resource_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/resource_controller_spec.rb
@@ -237,6 +237,41 @@ describe Spree::Admin::WidgetsController, type: :controller do
         expect(Widget.all.order('position')).to eq [widget_2, widget_1]
       end
     end
+
+    context "with take care of acts_as_list's after update callback" do
+      let(:widget_1) { Widget.create!(id: 4, name: 'widget 4', position: 1) }
+      let(:widget_2) { Widget.create!(id: 2, name: 'widget 3', position: 2) }
+      let(:widget_3) { Widget.create!(id: 1, name: 'widget 2', position: 3) }
+      let(:widget_4) { Widget.create!(id: 3, name: 'widget 1', position: 4) }
+
+      subject do
+        post :update_positions, params: { id: widget_1.to_param,
+          positions: { widget_3.id => '4', widget_2.id => '3', widget_4.id => '1', widget_1.id => '2' }, format: 'js' }
+      end
+
+      it 'updates the position of all widgets' do
+        subject
+        expect(Widget.all.order('position').map(&:id)).to eq [widget_4, widget_1, widget_2, widget_3].map(&:id)
+      end
+
+      context "not passing all elements" do
+        subject do
+          post :update_positions, params: { id: widget_1.to_param,
+            positions: { widget_3.id => '1', widget_2.id => '2', widget_4.id => '3' }, format: 'js' }
+        end
+
+        it 'updates the position of all widgets' do
+          widget_1 # trigger the creation
+          widget_2 # trigger the creation
+          widget_3 # trigger the creation
+          widget_4 # trigger the creation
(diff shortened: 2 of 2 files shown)


backend/app/controllers/spree/admin/resource_controller.rb
```diff
@@ -79,12 +79,16 @@ def create
   def update_positions
     ActiveRecord::Base.transaction do
       positions = params[:positions]
-      records = model_class.where(id: positions.keys).to_a
```
elia, line 82: Thoughts on using `index_by` to avoid running a query for each position?

```rb
# use strings to normalize ids being string vs. integer
records = model_class.where(id: positions.keys).index_by { _1.id.to_s }
  # …
  records[id.to_s]&.set_list_position(index)
```

elia (commented):
I know it's early as the PR is still in draft, but I was perusing it and I thought was worth suggesting on performance improvement 😊

backend/app/controllers/spree/admin/resource_controller.rb
julienanne: Thanks for help.

I'm not sure of what I say but `index_by` from https://api.rubyonrails.org/classes/Enumerable.html#method-i-index_by seems to reproduce the "previous" behavior aka one request and no take care of after update acts_as_list callback and so keep the bug or I miss something ?

Spec https://github.com/solidusio/solidus/pull/5040/files#diff-5025be293c7a07819979e3d765405a1fe07fc4d24f4d4aec17c59e358357d273R241-R256 still red with this implementation, maybe spec is wrong ?

Can you explain to me if I'm going wrong please?

elia: Oh snap! That's right!

I wonder if we should just disable it and use update_column https://github.com/brendon/acts_as_list#disabling-temporarily

julienanne: Oh At the first read I think like you but I add a spec to show the different behavior (correct or not I'm not sure) between with block `acts_as_list_no_update` and without it : https://github.com/solidusio/solidus/pull/5040/files#diff-5025be293c7a07819979e3d765405a1fe07fc4d24f4d4aec17c59e358357d273R257-R273

This spec with the actual code (without `acts_as_list_no_update` block) is green and with the `acts_as_list_no_update` is red, because we have a duplicate position number the "1" because not passed to params so no update and so not increment/decrement because of the block no update, so it's a trade-off situation between performance and real order and a choice between different behaviors :smile:  : 
```
solidus_user@4b984c481e10:~/app/backend$ bundle exec rspec spec/controllers/spree/admin/resource_controller_spec.rb 

Randomized with seed 7646
..............F..

Failures:

  1) Spree::Admin::WidgetsController#update_positions with take care of acts_as_list's after update callback not passing all elements updates the position of all widgets
     Failure/Error: expect(Widget.all.order('position').pluck(:position).uniq).to eq [1, 2, 3, 4] # trivial test to show no duplicate position integer
     
       expected: [1, 2, 3, 4]
            got: [1, 2, 3]
     
       (compared using ==)
     # ./spec/controllers/spree/admin/resource_controller_spec.rb:270:in `block (5 levels) in <top (required)>'
     # /home/solidus_user/gems/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:124:in `block in run'
     # /home/solidus_user/gems/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `loop'
     # /home/solidus_user/gems/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:110:in `run'
     # /home/solidus_user/gems/gems/rspec-retry-0.6.2/lib/rspec_ext/rspec_ext.rb:12:in `run_with_retry'
     # /home/solidus_user/gems/gems/rspec-retry-0.6.2/lib/rspec/retry.rb:37:in `block (2 levels) in setup'

Finished in 0.47797 seconds (files took 1.54 seconds to load)
17 examples, 1 failure

Failed examples:

rspec ./spec/controllers/spree/admin/resource_controller_spec.rb:263 # Spree::Admin::WidgetsController#update_positions with take care of acts_as_list's after update callback not passing all elements updates the position of all widgets

Randomized with seed 7646

```

For this PR I let you choose the right approach (on the project where I work I choose the non performant approach because I can ;) )

Happy to discuss.
Thanks again for help.

elia: @julienanne I did an experiment starting from this PR and was able to make it work with a fixed number of queries with this code:

```rb
  def update_positions
    ActiveRecord::Base.transaction do
      position_column = model_class.new.position_column # `acts_as_list` defines this method only on instances
      positions = params[:positions].reject { |id, position| id =~ /\D/ || position =~ /\D/ } # clean the input
      missing_ids = model_class.where.not(id: positions.keys).reorder(position_column => :asc).pluck(:id)
      
      updates = []
      positions # Fill in provided positions, removing duplicates, with zero-based indexing
        .each { |id, position| updates[position.to_i - 1] = id }
      updates # Fill in the missing positions
        .each.with_index { |value, position| updates[position] = missing_ids.shift unless value }
      updates += missing_ids # Add any remaining missing ids to the end
      updates.compact!

      model_class.update(
        updates, 
        (0...updates.size).map { { position_column => _1 + 1 } }
      )
    end

    respond_to do |format|
      format.js { head :no_content }
    end
  end
```

But it's probably a bit too much, I'm ok with the non-performant version.
If doubling the queries is a problem, probably it was already a problem with the original code (they're both `O(n)`).

elia (approved):

kennyadsl (approved):
Thanks, approving but we need a rebase. I think @elia is already on it.

julienanne:
Thanks all

PR #5023, opened 2023-04-18
Remove deprecated factories usage

kennyadsl:
This PR is part of [3.x Deprecations Removal](https://github.com/solidusio/solidus/issues/4983).

## Summary

This PR removes support for requiring individual factories from Solidus core using: 

```
require 'spree/testing_support/factories/some_factory'
```

Now, you should use the provided `FactoryBot.definition_file_paths`. We opened this [GitHub Discussion](https://github.com/solidusio/solidus/discussions/4014) with more details, and instructions to upgrade.

Ref https://github.com/solidusio/solidus/pull/3907

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/.github/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [ ] I have written a thorough PR description.
- [ ] I have kept my commits small and atomic.
- [ ] [I have used clear, explanatory commit messages](https://github.com/solidusio/.github/blob/main/CONTRIBUTING.md#writing-good-commit-messages).

The following are not always needed:

- 📖 I have updated the README to account for my changes.
- 📑 I have documented new code [with YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md).
- 🛣️ I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).
- ✅ I have added automated tests to cover my changes.
- 📸 I have attached screenshots to demo visual changes.

PR #5059, opened 2023-04-28
Deprecate other code related to old factories loading

kennyadsl:
## Summary

Ref #5023 

People should have started to use `require 'spree/testing_support/factory_bot'`now, so this workaround is no longer needed.

Also, no one is calling the methods that check for the version of factory_bot, so we can safely deprecate that code.

## Checklist

### DIFF
--- a/core/lib/spree/testing_support.rb
+++ b/core/lib/spree/testing_support.rb
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 
 require 'spree/testing_support/factory_bot'
+require 'spree/deprecation'
+
+Spree::Deprecation.warn <<-WARN
+  Using `require 'spree/testing_support'` is deprecated and will be removed in
+  Solidus 4.0.
+WARN
 
 module Spree
   module TestingSupport
@@ -11,5 +17,6 @@ module Spree
     def check_factory_bot_version
       Spree::TestingSupport::FactoryBot.check_version
     end
+    deprecate :check_factory_bot_version, deprecator: Spree::Deprecation
   end
 end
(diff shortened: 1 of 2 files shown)


elia (approved):

elia:
@kennyadsl would make sense to add a warning in the file instructing people to stop requiring it?

rainerdema (approved):

kennyadsl:
@elia I tried but I think it's still loaded by Zeitwerk when someone require an inner module like 'Spree::TestingSupport::FactoryBot'. Do you know any way to avoid this?

kennyadsl:
> Are you sure?

Nope, my attempt was removing the file completely and the CI was failing, but maybe thinking twice, it was for another reason, see [here](https://github.com/solidusio/solidus_starter_frontend/pull/338/files#diff-1ceb42e9dd4ccfd7b01826cb5e570004ec897791b4f43d0117d2a6e4d5717ea5L118). Let me try again. 

> However, shouldn't we deprecate the method itself?

Yes, you are right, even if it's [not used](https://github.com/search?q=repo%3Asolidusio%2Fsolidus%20check_factory_bot_version&type=code). 

Considering that it's better to add deprecations to push people to not require this file at all, marking it and its content for removal, does this mean we can do this whole PR later after releasing v4.0? I think it does, no rush on this cleanup.

waiting-for-dev:
> Considering that it's better to add deprecations to push people to not require this file at all, marking it and its content for removal, does this mean we can do this whole PR later after releasing v4.0? I think it does, no rush on this cleanup.

We can also release another 3.4.x version with the warning and remove the file altogether on v4, but either way it's not something critical

kennyadsl:
@waiting-for-dev This PR is now deprecating those methods/files instead of removing them. I think the idea to backport and release a patch is acceptable because we still didn't release v4.0.

waiting-for-dev (approved):
:+1:

PR #5068, opened 2023-05-04
Add `SolidusAdmin`

elia:
Add SolidusAdmin to the list solidus sub-frameworks, for now it will live in the repo and be released separately while we complete coverage for the existing backend sections.

During the intermediate phase it will be possible to keep SolidusAdmin and Spree::Backend in the same rails application and benefit immediately from the updated UI/UX of the sections covered by the new admin dashboard.

### Features

- ✨ Updated modern UI, more intuitive and familiar to modern users
- 🧰 A dedicated UI Kit that can be used to easily create custom sections with high quality and low effort
- 🌱 A built-in switch that allows to fall back on the legacy admin dashboard that allows a soft transition and early feedback 
- 🎛️ A components registry that will enable controlled updates to the UI/UX and to completely abandon UI monkey patching (e.g. with `deface`)

**Action Items:**

- [x] Consider creating an hosted demo for the new admin interface before the upcoming release. This would provide an interactive way to showcase the improvements and new features.


admin/README.md
```diff
@@ -0,0 +1,4 @@
+# solidus_admin (WIP)
+
+A Rails engine that provides a backend administration interface to the Solidus ecommerce platform.
```
waiting-for-dev, line 3: Thoughts on getting rid of the word "backend" completely? slightly_smiling_face Just saying "provides an administration interface" should be enough and less confusing.

waiting-for-dev:
Should we add `solidus_admin` as a dependency in `solidus_core`'s gemspec?

admin/README.md
elia: Sure, can I blame Copilot? 😄

waiting-for-dev: :rofl:

rainerdema (approved):
🚢

PR #5070, opened 2023-05-05
Install & configure Tailwind for solidus_admin

waiting-for-dev:
## Summary

We add [tailwindcss-rails](https://github.com/rails/tailwindcss-rails) as a dependency for solidus_admin. It's worth noticing that it's only a development dependency. The host application won't need to use tailwind, as it'll just consume the compiled CSS through the assets pipeline.

As with the other components, the development environment is configured on the top Gemfile. We also need to add `solidus_admin` gem as a dependency for `solidus`.

We mimic how tailwindcss is [configured in a regular Rails application](https://tailwindcss.com/docs/guides/ruby-on-rails):

- `admin/config/tailwind.config.js` configure which paths are scanned for compilation. For now, we leave the default configuration.
- `admin/app/assets/stylesheets/application.tailwind.css` configures tailwind directives. For now, we leave the default configuration.
- We use `admin/app/assets/builds/solidus_admin.css` for the compiled CSS, which is added to the Rails manifest.

We've prepared an example to show how tailwind is being used. It can be rendered at `/admin/orders`, and we make use of it both in the layout and action templates. We need to isolate the engine so that the right controller is used.

We're also adding a task to automatically watch tailwind on solidus_admin. To help with development, we're adding a `bin/dev` script that automates the process of running the sandbox app and watching tailwind.

## Checklist


admin/app/assets/builds/solidus_admin.css
```diff
@@ -0,0 +1 @@
...}*,::backdrop,:after,:before{--tw-border-spacing-x:0;--tw-border-spacing-y:0;--tw-translate-x:0;--tw-translate-y:0;--tw-rotate:0;--tw-skew-x:0;--tw-skew-y:0;--tw-scale-x:1;--tw-scale-y:1;--tw-pan-x: ;--tw-pan-y: ;--tw-pinch-zoom: ;--tw-scroll-snap-strictness:proximity;--tw-gradient-from-position: ;--tw-gradient-via-position: ;--tw-gradient-to-position: ;--tw-ordinal: ;--tw-slashed-zero: ;--tw-numeric-figure: ;--tw-numeric-spacing: ;--tw-numeric-fraction: ;--tw-ring-inset: ;--tw-ring-offset-width:0px;--tw-ring-offset-color:#fff;--tw-ring-color:#3b82f680;--tw-ring-offset-shadow:0 0 #0000;--tw-ring-shadow:0 0 #0000;--tw-shadow:0 0 #0000;--tw-shadow-colored:0 0 #0000;--tw-blur: ;--tw-brightness: ;--tw-contrast: ;--tw-grayscale: ;--tw-hue-rotate: ;--tw-invert: ;--tw-saturate: ;--tw-sepia: ;--tw-drop-shadow: ;--tw-backdrop-blur: ;--tw-backdrop-brightness: ;--tw-backdrop-contrast: ;--tw-backdrop-grayscale: ;--tw-backdrop-hue-rotate: ;--tw-backdrop-invert: ;--tw-backdrop-opacity: ;--tw-backdrop-saturate: ;--tw-backdrop-sepia: }.text-3xl{font-size:1.875rem;line-height:2.25rem}.text-4xl{font-size:2.25rem;line-height:2.5rem}.font-bold{font-weight:700}.underline{text-decoration-line:underline}
```
elia: Should we also git-ignore this file?

admin/config/tailwind.config.js
```diff
@@ -0,0 +1,23 @@
+const defaultTheme = require('tailwindcss/defaultTheme')
+
+module.exports = {
+  content: [
+    './public/*.html',
+    './app/helpers/**/*.rb',
+    './app/javascript/**/*.js',
+    './app/views/**/*.{erb,haml,html,slim}'
+  ],
```
elia, line 9: In my understanding compiling with this configuration would mean that custom views won't have access to the full set of utility classes… should we change the config to include everything?

Procfile.dev
```diff
@@ -0,0 +1,2 @@
+web: env RUBY_DEBUG_OPEN=true bin/rails server
+admin_tailwind: cd admin && bundle exec rake solidus_admin:tailwind:watch
```
elia, line 2: I think this can be replaced by `bin/rails solidus_admin:tailwind:watch`

admin/Rakefile
```diff
@@ -15,3 +15,17 @@ DummyApp::RakeTasks.new(
...
+namespace :solidus_admin do
+  namespace :tailwind do
+    task :watch do
+      require "tailwindcss-rails"
+      require "solidus_admin/engine"
+      system "#{Tailwindcss::Engine.root.join("exe/tailwindcss")} \
+         -i #{SolidusAdmin::Engine.root.join("app/assets/stylesheets/application.tailwind.css")} \
+         -o #{SolidusAdmin::Engine.root.join("app/assets/builds/solidus_admin.css")} \
+         -c #{SolidusAdmin::Engine.root.join("config/tailwind.config.js")} \
+         --minify -w"
```
tvdeyen, line 28: Could it be beneficial to not minify the output in order to debug CSS issues? At least during rebuilding the admin?

admin/app/assets/builds/solidus_admin.css
waiting-for-dev: I don't think so. By adding the compiled CSS, we make tailwindrails-css a development dependency for solidus_admin, but not something end users need to worry about. Otherwise, they would need to add it to their dependencies and compile a file we're already able to generate.

admin/config/tailwind.config.js
waiting-for-dev: That's related to the last point. I think is correct not to try to scan the host application files. We're only using tailwindcss-rails as a development dependency. Users can use whatever they want if they want to override parts of solidus_admin. If they also want to use tailwind, they still can, and they just need to configure it however they prefer.

Procfile.dev
waiting-for-dev: Using `bin/rails` here won't find the task. I think that's because it is not defined within `lib/tasks`. Instead, it's placed directly in the `Rakefile`. I did that purposefully because that task shouldn't be exposed to the host application. They don't need to compile a file that we use for development. Let me know if you think otherwise :pray:

admin/Rakefile
waiting-for-dev: Honestly, I think that most debugging will happen through the browser development utilities. If we find something, we'll need to touch the HTML, not the compiled file. At this point, I see no value in skipping the minification, but we can reconsider if we think otherwise. Happy also to hear what others think :pray:

admin/config/tailwind.config.js
elia: Taking the use case of building a custom component that I need to integrate in the solidus admin I think I would want to have access to the same CSS framework that was used for the existing components instead of reverse engineer the spacing, colors, etc. 

If not possible to have a generating CSS (e.g. some classes might be dynamically generated) maybe we should have an option to run tailwind in the target application?

admin/app/assets/builds/solidus_admin.css
elia: This could become part of the gem release process, but I guess it's also fine, many JS libraries host a `dist` directory with the compiled artifacts (not pretty, but… whatever)

admin/config/tailwind.config.js
waiting-for-dev: Oh, I see what you mean. That's a valid point. In that case, it would be enough for users to point to our `app/assets/stylesheets/application.tailwind.css` when compiling their CSS, right? :thinking:

admin/app/assets/builds/solidus_admin.css
elia: Yes, exactly, we could enhance the `release` task, but of course there's always the option of going lower level and skip steps, but it's also true that the release won't be done by random people, and we have SOPs and scripts in place.

Anyway it's not super important where we put the file 😊

admin/config/tailwind.config.js
elia: I guess if I add a custom component (e.g. in `mystore/app/components/mystore/button_component.html.erb`) and inside it I use tailwind classes that were tree-shaken by the compilation I would need to recompile the CSS starting from the tailwind configuration. Alternatively we should check if it's possible to have a bigger CSS without tree shaking that can be used for additional components.

elia: See this section on 3rd party libraries providing TW styled components https://tailwindcss.com/docs/content-configuration#working-with-third-party-libraries

waiting-for-dev: Users can probably `@import` solidus_admin's `application.tailwind.css` to access the same set of utilities. However, we might want to give more thoughts about the best way to provide extensibility (for instance, do we want to use a prefix?). As this PR was more in the lines of a POC for using tailwind from the admin, are you ok if we leave that for a follow-up PR?

kennyadsl (approved):

admin/Rakefile
kennyadsl: If we start to feel the need of having it not minified when developing, we can change this anytime. 👍

PR #5072, opened 2023-05-08
Make Spree::MigrationHelpers Ruby 3.0 compatible

RyanofWoods:
## Summary
Due to the separation of positional and keyword arguments in Ruby 3.0, the options default value of a hash can no longer be coerced into keyword arguments. Using double splat avoids this by only allowing optional arguments as keywords.

## Checklist

### DIFF
--- a/core/spec/lib/spree/migration_helpers_spec.rb
+++ b/core/spec/lib/spree/migration_helpers_spec.rb
@@ -1,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe Spree::MigrationHelpers do
+  let(:helper) do
+    double.as_null_object.tap do |object|
+      object.extend(described_class)
+      object.extend(ActiveRecord::ConnectionAdapters::SchemaStatements)
+    end
+  end
+
+  subject { helper.safe_add_index(double, double) }
+
+  # regression test
+  describe "#safe_add_index" do
+    context "when the column exists" do
+      context "and the index does" do
+        it "passes compatible arguments to index_exists?" do
+          expect { subject }.to_not raise_error(ArgumentError)
+        end
+      end
+
+      context "and the index does not" do
+        before do
+          allow(helper).to receive(:index_exists?).with(any_args).and_return(false)
+        end
+
+        it "passes compatible arguments to add_index" do
+          expect { subject }.to_not raise_error(ArgumentError)
+        end
+      end
+    end
+  end
+end
--- a/core/lib/spree/migration_helpers.rb
+++ b/core/lib/spree/migration_helpers.rb
@@ -6,9 +6,9 @@ module Spree
       remove_index(table, column) if index_exists?(table, column)
     end
 
-    def safe_add_index(table, column, options = {})
-      if columns_exist?(table, column) && !index_exists?(table, column, options)
-        add_index(table, column, options)
+    def safe_add_index(table, column, **options)
+      if columns_exist?(table, column) && !index_exists?(table, column, **options)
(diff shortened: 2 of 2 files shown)


core/spec/lib/spree/migration_helpers_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  subject { helper.safe_add_index(double, double) }
+
+  # regression test
+  describe "#safe_add_index" do
+    context "when the column exists" do
+      context "and the index does" do
+        it "is compatible with Ruby 3.0" do
+          expect { subject }.to_not raise_error(ArgumentError)
```
RyanofWoods, line 18: RFC. I don't normally write specs with `to_not raise_error(...)` as be misleading if the test fails for other reasons, yet the spec passes. But, avoiding this, means a much more complicated test and setup.

RyanofWoods:
@kennyadsl previously discussed on Slack, that we might want to remove this logic from Solidus as it's not used by Solidus.

I agree with the idea that stores should be responsible for maintaining their own migration logic.

WDYT? Should I make a new PR to deprecate the module and its methods?

core/spec/lib/spree/migration_helpers_spec.rb
kennyadsl: I think this is more than enough for something that we do no longer want to maintain. Thanks!

core/spec/lib/spree/migration_helpers_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+  end
+
+  subject { helper.safe_add_index(double, double) }
+
+  # regression test
+  describe "#safe_add_index" do
+    context "when the column exists" do
+      context "and the index does" do
+        it "is compatible with Ruby 3.0" do
```
kennyadsl, line 17: Not blocking, but I wouldn't be specific about the version under test in the `it` description. That's something that is controlled by the CI matrix; for example, if this spec is run in the CI with Ruby 2.7, it will still pass. My suggestion is to just document the behavior, like `it "add the index correctly" do`, if it will fail when testing against Ruby 3.0, we'll get to know by the right CI step. Thoughts?

kennyadsl (approved):
Thanks @RyanofWoods! I'm good as is, but I left a suggestion. 🙏

core/spec/lib/spree/migration_helpers_spec.rb
RyanofWoods: This is a good point @kennyadsl. Though I think `add the index correctly` is misleading, as we aren't testing that it works fully. Due to it's incomplete setup, the method actually raises other errors related to the migration. So we are specifically only testing argument compatibility.

WDYT about:
- `it "passes compatible arguments to index_exists?" do`; and
- `it "passes compatible arguments to add_index" do`

kennyadsl: Both Make sense, that's what I wanted to propose 👍. Go with the one you like the most.

RyanofWoods:
Please backlog to supported versions: v3.4, v3.3 and v3.2. Though note, this will mean v3.1 which is now unsupported will have a breaking change as this is the Solidus version where Ruby 3.0 became supported.

kennyadsl:
@RyanofWoods we could backport this change to the 3.1 branch in GH (people could still reference to it) without releasing a new version.

waiting-for-dev (approved):

PR #5080, opened 2023-05-12
Allow extending Admin's tailwind

waiting-for-dev:
## Summary

Allow host app & extensions to make use & extend admin's tailwind classes as well as adding new CSS.

As a requisite for this change, the compiled CSS is now generated in the host app's assets directory, and tailwindcss-rails needs to be an explicit dependency for solidus_admin.

With this change, the host app and extensions can make use of the tailwind classes defined in the admin. They can also extend the tailwind configuration by adding their own classes.

The way it works is that we convert the tailwind.config.js file into a `.js.erb` template. We populate the `content` property by reading from a `SolidusAdmin::Config#tailwind_content` setting. Through custom `solidus_admin:tailwindcss:build` and `solidus_admin:tailwindcss:watch` tasks, we generate the actual tailwind configuration file on the fly.

By default, the `tailwind_content` setting contains both the admin's and the host app's paths scoped to `solidus_admin` directories, so a host application is not required to do further configuration. On the other side, extensions can manually push new paths on an initializer. As solidus_admin's configuration file requires access to `Rails.root` to create those defaults, the file must be required after the Rails application is created.

The solidus_admin's tailwind tasks have been moved to the `lib/tasks` directory so they can be picked by the `bin/rails` command from within the host app.

When solidus_admin is installed, tailwind is automatically built and the compiled file is added to the `.gitignore` file. The compilation is also added as an enhancement for the `assets:precompile` task.

In a similar way, we're now allowing the host app and extensions to add their own Tailwind CSS.

This is done by changing the `application.tailwind.css` file to an ERB template. Both the host application and extensions can add their own stylesheets through the `tailwind_stylesheets` configuration option. During compilation time, the content for those `tailwind_stylesheets` is appended to the dynamically generated CSS file.

As adding new styles should be thoroughly thought as per tailwind recomendations, we're not creating a default stylesheet for the host app and, instead, it needs to be manually created.

The way we're doing this can be seen as a bit clumsy, but the alternative requires [using `postcss-import`](https://tailwindcss.com/docs/using-with-preprocessors#build-time-imports) and adding all the [npm jazz to the mix](https://github.com/rails/tailwindcss-rails/issues/133). There have been some [discussion](https://github.com/rails/tailwindcss-rails/issues/133) on tailwindcss-rails about this, but there's no built-in way as of today.

If we go forward with this proposal, we can add new settings to allow configuring other parts of tailwind's config file, like plugins.

## Checklist


waiting-for-dev:
I have added three extra commits:

## Allow overriding SolidusAdmin's configuration file

`app/config/solidus_admin/tailwind.config.js.erb` will take precedence if present. That can be helpful when more advanced configuration is needed, like adding custom plugins or extending the default theme.

A `bin/rails tailwind:override_config` task has been added to help with this.

## Allow overriding SolidusAdmin's Tailwind stylesheet

Same for the Tailwind stylesheet. That can be helpful for advanced CSS customizations.

We're adding a `bin/rails tailwind:override_stylesheet` task that copies the default stylesheet to `app/assets/stylesheets/solidus_admin/application.tailwind.css.erb`.

## Create an initializer when installing Solidus admin

That will help users to know what they can configure about the admin interface. For now, tailwind options are the only ones available.

admin/lib/generators/solidus_admin/install/install_generator.rb
```diff
@@ -10,6 +10,14 @@ def install_solidus_core_support
           }
         RUBY
       end
+
+      def ignore_tailwind_build_files
+        append_file(".gitignore", "app/assets/stylesheets/solidus_admin/builds/") if File.exist?(Rails.root.join(".gitignore"))
+      end
+
+      def build_tailwind
+        rake "solidus_admin:tailwindcss:build"
```
kennyadsl, line 21: Why is this task needed during the installation?

kennyadsl (approved):
> there's a typo in the first commit ("configuratio")

Fixed, thanks!!

admin/lib/generators/solidus_admin/install/install_generator.rb
waiting-for-dev: Because the compiled CSS is generated in the host application, that's to allow users to extend the admin style, for instance, using other tailwind classes or stylesheets.

kennyadsl: Got it! I somehow thought that was already part of the watch task. Do you think it might be a good idea to always run it when the watch task is executed?

admin/app/views/layouts/solidus_admin/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <%= stylesheet_link_tag "solidus_admin", "data-turbo-track": "reload" %>
+    <%= stylesheet_link_tag "solidus_admin_manifest", "data-turbo-track": "reload" %>
```
elia, line 4: Usually the manifest is not used directly, but just to tell sprockets what assets it needs to compile and keep track of, see app/assets/config/manifest.js and app/assets/javascripts/application.js in a vanilla rails app).

Open to discussing this in case I'm missing an angle, and/or to fix it in a follow up PR

admin/lib/tasks/tailwind.rake
```diff
@@ -0,0 +1,39 @@
+namespace :solidus_admin do
+  namespace :tailwindcss do
```
elia, line 2: optional nit: thoughts on using "tailswindcss" also for the filename?

admin/lib/tasks/tailwind.rake
```diff
@@ -0,0 +1,39 @@
...
+    require "tailwindcss-rails"
+
+    def run(args = "")
```
elia, line 31: I don't remember how rake works with regards to methods, is this defined on top-level or inside the rake task?

admin/lib/tasks/tailwind.rake
```diff
@@ -3,14 +3,21 @@ namespace :solidus_admin do
     require "tailwindcss-rails"
 
     def config_app_path
-      Rails.root.join("config", "tailwind.config.js.erb")
```
elia: optional nit: Oh, I think this removal goes to the previous commit…

admin/lib/generators/solidus_admin/install/install_generator.rb
elia: @waiting-for-dev would it still be possible to ship a compiled CSS that won't require tailwind to be installed for all those not needing to customize?

elia (approved):
Great stuff!

I left some questions and doubts, but also approving, up to you if it's worth updating this PR or use a follow up PR (if needed).

admin/lib/generators/solidus_admin/install/install_generator.rb
waiting-for-dev: I gave more thought to it, and I think it'd be possible, but at the expense of a more complex setup.

We could keep the generation of the CSS for the solidus_admin engine within the development process of it. Then, let's consider a host application that needs to reuse. It should install tailwind, and we could have a different task using the same tailwind config file (the erb template) from the engine but generating the CSS output file to an application path matching the one in solidus_admin.

If we add other engines into the mix, each one should do something similar to what I described for the host application. However, each of them should ship an isolated CSS file, so we should change the solidus_admin manifest to take all of them.

I wonder if it pays off, as now customization is more streamlined for both scenarios. Happy to explore it further if you think that would be a good idea, but I'd prefer to do that on another PR. We can keep it in the back of our minds as a future optimization if we feel the need for it.

admin/lib/tasks/tailwind.rake
waiting-for-dev: Makes total sense. Changed!

admin/lib/tasks/tailwind.rake
waiting-for-dev: Great call! I was under the assumption that it created an anonymous class and bound to it, but it was polluting the `Object` namespace. I extracted to a helper module. Thanks for pointing out.

admin/lib/tasks/tailwind.rake
```diff
@@ -2,6 +2,15 @@ namespace :solidus_admin do
   namespace :tailwindcss do
     require "tailwindcss-rails"
 
+    def config_app_path
+      Rails.root.join("config", "tailwind.config.js.erb")
+      Rails.root.join("config", "solidus_admin", "tailwind.config.js.erb")
```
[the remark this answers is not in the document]
waiting-for-dev: See https://github.com/solidusio/solidus/pull/5080#discussion_r1202034494

admin/lib/tasks/tailwind.rake
waiting-for-dev: Thanks! Fixed.

admin/app/views/layouts/solidus_admin/application.html.erb
elia: Oh that's right, I got confused because in sprockets parlance the manifest is the one in `config/` with _link_ directives, and not the final file with a require in it. 

Not fundamental, but what would you think about removing the `_manifest` suffix?

waiting-for-dev: Makes sense, @elia. I changed the configuration of the assets to follow the conventions used in a host application, but namespacing everything within a `solidus_admin` directory. The main CSS manifest file is `app/assets/solidus_admin/application.css`, and the compiled file is saved to `app/assets/build/solidus_admin/tailwind.css`, to follow [tailwindcss-rails conventions](https://github.com/rails/tailwindcss-rails#configuration).

PR #5086, opened 2023-05-24
Update Taxon PaperClip attributes on attachment destroy

RyanofWoods:
## Summary
When using Paperclip, calling destroy on the attachment removes it from
from the storage location, such as S3, and leaves dirty changes on the
Paperclip attributes, such as icon_file_name.

If a store used `.present?` on the icon, it would return continue to
return `true`, as it looks at the file_name attribute [1], yet the
icon no longer exists on storage (`.exists? => false`).

A view such as the backend Taxon edit view, could therefore attempt to
show the icon, but it would fail and show a broken icon image.

Note:
- an assertion was not added for `.present?` as ActiveStorage's API
  works a little differently;
- stores using Paperclip may have Taxons with Paperclip attributes
  present, yet the image on storage no longer exists. For these Taxons, a broken image and the remove button will appear in the backend. Frontend could be affected as well.

[1] https://github.com/kreeti/kt-paperclip/blob/91923a6ee1bd692696b51ea839f2f2578cb485b4/lib/paperclip/attachment.rb#L357-L361

### Screenshots
| Description | Screenshot |
|--|--|
| Attached icon before deletion | Screenshot 2023-05-24 at 08 36 19|
| Attached icon after deletion but before fix | Screenshot 2023-05-24 at 08 37 55 |
| Attempting to remove a broken Taxon's icon which does not exist | Screenshot 2023-05-24 at 09 58 46 |
| Attached icon after deletion and after fix | Screenshot 2023-05-24 at 09 59 14 |

## Checklist

### DIFF
--- a/core/lib/generators/solidus/install/install_generator.rb
+++ b/core/lib/generators/solidus/install/install_generator.rb
@@ -96,7 +96,7 @@ module Solidus
         rake 'active_storage:install'
       else
         say_status :assets, "Paperclip", :green
-        gsub_file 'config/initializers/spree.rb', "ActiveStorageAttachment", "PaperclipAttachment"
+        gsub_file 'config/initializers/spree.rb', "::ActiveStorageAttachment", "::PaperclipAttachment"
       end
     end
 
--- a/core/spec/models/spree/taxon_spec.rb
+++ b/core/spec/models/spree/taxon_spec.rb
@@ -20,15 +20,14 @@ RSpec.describe Spree::Taxon, type: :model do
   describe "#destroy_attachment" do
     context "when trying to destroy a valid attachment definition" do
       context "and taxon has a file attached " do
-        it "removes the attachment" do
-          taxon = create(:taxon, :with_icon)
+        let(:taxon) { create(:taxon, :with_icon)}
 
+        it "removes the attachment" do
           expect(taxon.destroy_attachment(:icon)).to be_truthy
         end
 
         if Spree::Config.taxon_attachment_module == Spree::Taxon::PaperclipAttachment
           it "resets paperclip attributes when using Paperclip", aggregate_failures: true do
-            taxon = create(:taxon, :with_icon)
             expect(taxon.destroy_attachment(:icon)).to be_truthy
             expect(taxon.reload.icon_file_name).to_not be_present
             expect(taxon.reload.icon_content_type).to_not be_present
@@ -39,18 +38,18 @@ RSpec.describe Spree::Taxon, type: :model do
       end
 
       context "and the taxon does not have any file attached yet" do
-        it "returns false" do
-          taxon = create(:taxon)
+        let(:taxon) { create(:taxon)}
 
+        it "returns false" do
           expect(taxon.destroy_attachment(:icon)).to be_falsey
         end
       end
     end
 
     context "when trying to destroy an invalid attachment" do
-      it 'returns false' do
-        taxon = create(:taxon)
+      let(:taxon) { create(:taxon)}
 
+      it 'returns false' do
         expect(taxon.destroy_attachment(:foo)).to be_falsey
       end
     end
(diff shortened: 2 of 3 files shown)


core/spec/models/spree/taxon_spec.rb
```diff
@@ -20,15 +20,15 @@
   describe "#destroy_attachment" do
     context "when trying to destroy a valid attachment definition" do
       context "and taxon has a file attached " do
+        let(:taxon) { create(:taxon, :with_icon)}
+
         it "removes the attachment" do
           taxon = create(:taxon, :with_icon)
```
kennyadsl, line 26: Should we remove this line?

core/lib/generators/solidus/install/install_generator.rb
```diff
@@ -96,7 +96,7 @@ def install_file_attachment
         rake 'active_storage:install'
       else
         say_status :assets, "Paperclip", :green
-        gsub_file 'config/initializers/spree.rb', "ActiveStorageAttachment", "PaperclipAttachment"
+        gsub_file 'config/initializers/spree.rb', "::ActiveStorageAttachment", "::PaperclipAttachment"
```
kennyadsl, line 99: ❤️

kennyadsl (approved):
Thanks @RyanofWoods!!! This makes perfect sense, and we will probably backport this bug fix in all the supported Solidus versions.

I left a question/comment for a line in the spec that might not be necessary anymore, but let me know if that's correct or I missed something. 

🙏

core/spec/models/spree/taxon_spec.rb
RyanofWoods: Definitely 🤦 😁 Thanks!

RyanofWoods:
One more note, a Taxon's PaperClip icon attachment will now trigger some Taxon callbacks due to the `save`:
https://github.com/solidusio/solidus/blob/a8f56d585dd50f802827df53630e4a08e10bcfe2/core/app/models/spree/taxon.rb#L27

A bit unfortunate, but I don't think it will be an issue.

kennyadsl:
> but I don't think it will be an issue.

I don't think either. On the other side, it will invalidate the cache which might be a good thing when the icon is removed.

RyanofWoods:
Good question @waiting-for-dev
1. I am not really sure why paperclip is not saving the associated record. Note, that the documented `save` is actually for the attachment though. Also, it seems that the maintainers don't really know why these attributes are not being set to `nil` 🤔 https://github.com/kreeti/kt-paperclip/pull/91#discussion_r1027060962. Let's hold off on merging this, while I  investigate further 
2. Good idea, I will do that now 👍

RyanofWoods:
A few findings @waiting-for-dev. Note `exists?` checks the actual storage, while `.present?` checks the attachment's filename attribute

Dec 6, 2019:
Replaced taxons/_form attachment `.exists?` to `.present?`
https://github.com/solidusio/solidus/commit/484f7d58f731060ae2a08c069ac99f7e205bed38

Jan 6, 2021:
Replaced taxons/_form attachment `.present?` to `.exists?`
https://github.com/solidusio/solidus/commit/6e19cd795db4779088275bc2965455917f8d29e9

Because of the latter, I think it's important to continue using `.present?`. We therefore need  the taxon's `icon_file_name` to be saved to `nil`, so the storage is in sync with attribute, and `present?` is correct.

From the [kt-paperclip](https://github.com/kreeti/kt-paperclip) documentation, it seems that they intentionally do not call save on the taxon instance because they want the attachment to act like an "attribute".
```
The intent behind it was to keep setup as easy as possible and to treat files as much like other attributes as possible. This means they aren't saved to their final locations on disk, nor are they deleted if set to nil, until ActiveRecord::Base#save is called.
```

waiting-for-dev (approved):
Thanks for your investigation, @RyanofWoods!

waiting-for-dev:
Oh, backport failed probably because the last commit in this PR had already been backported :disappointed: We'll need to do it manually...

RyanofWoods:
Sorry @waiting-for-dev 😢 I removed the commit, but I see that I pushed to my fork instead of `nebulab` 🙁

waiting-for-dev:
No worries, @RyanofWoods, my fault for not checking before. Luckily, it has easy solution :slightly_smiling_face:

PR #5091, opened 2023-05-25
Add support for admin themes

elia:
## Summary

We're going to refresh the UX/UI for the admin in the next future and we want to match it in the current admin so that switching between the two versions won't be too confusing.

The preference is split between light and dark because most likely the new UI will have dark mode support, and also to allow stores that desire it to easily add their own dark theme.

Selecting different themes as an individual preference for each user (or session) has been considered out of scope for this PR but it's something that will eventually be added.

## Checklist

### DIFF
--- a/backend/app/assets/stylesheets/spree/backend.css
+++ b/backend/app/assets/stylesheets/spree/backend.css
@@ -1,7 +1,6 @@
 /*
- * This is a manifest file that'll automatically include all the stylesheets available in this directory
- * and any sub-directories. You're free to add application-wide styles to this file and they'll appear at
- * the top of the compiled file, but it's generally better to create a new file per style scope.
-
- *= require spree/backend/spree_admin
-*/
+ * This is a manifest file that is here only for legacy support purposes, please refer to themes direcly via 
+ * appropriate `Spree::Config` settings.
+ *
+ *= require spree/backend/themes/classic
+ */
--- a/backend/app/assets/stylesheets/spree/backend.css
+++ b/backend/app/assets/stylesheets/spree/backend.css
@@ -3,9 +3,5 @@
  * and any sub-directories. You're free to add application-wide styles to this file and they'll appear at
  * the top of the compiled file, but it's generally better to create a new file per style scope.
 
- *= require solidus_admin/select2
- *= require solidus_admin/flatpickr/flatpickr
- *= require prism
-
  *= require spree/backend/spree_admin
 */
--- a/backend/app/assets/stylesheets/spree/backend/vendor/index.scss
+++ b/backend/app/assets/stylesheets/spree/backend/vendor/index.scss
@@ -1,0 +1,11 @@
+@import "solidus_admin/select2";
+@import "solidus_admin/flatpickr/flatpickr";
+@import "prism";
+
+@import 'spree/backend/bootstrap_custom';
+@import 'solidus_admin/bootstrap/bootstrap';
+@import 'font-awesome';
+
+// These mixins will modify existing Bootstrap ones and need to be imported
+// after the Bootstrap has been processed.
+@import 'spree/backend/globals/mixins';
--- a/backend/app/views/spree/admin/shared/_head.html.erb
+++ b/backend/app/views/spree/admin/shared/_head.html.erb
@@ -6,7 +6,7 @@
 <title><%= admin_page_title %></title>
 
 <%= favicon_link_tag 'favicon.ico' %>
-<%= stylesheet_link_tag 'spree/backend/all', media: 'all', data: {turbolinks_track: 'reload'} %>
+<%= stylesheet_link_tag Spree::Backend::Config.theme_path, media: 'all', data: {turbolinks_track: 'reload'} %>
 <%= javascript_include_tag 'spree/backend/all', data: {turbolinks_track: 'reload'} %>
 
 <%- if Rails.env.test? %>
--- a/backend/lib/spree/backend_configuration.rb
+++ b/backend/lib/spree/backend_configuration.rb
@@ -6,6 +6,20 @@ module Spree
   class BackendConfiguration < Preferences::Configuration
     preference :locale, :string, default: I18n.default_locale
 
+    # @!attribute [rw] themes
+    #   @return [Hash] A hash containing the themes that are available for the admin panel
+    preference :themes, :hash, default: {
+      classic: 'spree/backend/all',
+    }
+
+    # @!attribute [rw] theme
+    #   @return [String] Default admin theme name
+    versioned_preference :theme, :string, initial_value: 'classic', boundaries: { "4.1.0.a" => "classic" }
+
+    def theme_path(user_theme = nil)
+      user_theme ? themes.fetch(user_theme.to_sym) : themes.fetch(theme.to_sym)
+    end
+
     preference :frontend_product_path,
       :proc,
       default: proc {
--- a/backend/spec/lib/spree/backend_configuration_spec.rb
+++ b/backend/spec/lib/spree/backend_configuration_spec.rb
@@ -79,4 +79,41 @@ RSpec.describe Spree::BackendConfiguration do
       end
     end
   end
+
+  describe '#theme_path' do
+    it 'returns the default theme path' do
+      subject.themes = { foo: 'foo-theme-path' }
+      subject.theme = :foo
+
+      expect(subject.theme_path).to eq('foo-theme-path')
+    end
+
+    it 'returns the default theme path when the theme is a string' do
+      subject.themes = { foo: 'foo-theme-path' }
+      subject.theme = 'foo'
+
+      expect(subject.theme_path).to eq('foo-theme-path')
+    end
+
+    it 'returns the fallback theme path when the default theme is missing' do
+      subject.themes = { foo: 'foo-theme-path', classic: 'classic-theme-path' }
+      subject.theme = :bar
+
+      expect{ subject.theme_path }.to raise_error(KeyError)
+    end
+
+    it 'gives priority to the user defined theme' do
+      subject.themes = { foo: 'foo-theme-path', user: 'user-theme-path' }
+      subject.theme = :foo
+
+      expect(subject.theme_path(:user)).to eq('user-theme-path')
+    end
+
+    it 'raises an error if the user theme is missing' do
+      subject.themes = { foo: 'foo-theme-path', classic: 'classic-theme-path' }
+      subject.theme = :foo
+
+      expect{ subject.theme_path(:bar) }.to raise_error(KeyError)
+    end
+  end
 end
--- a/backend/app/assets/config/solidus_backend_manifest.js
+++ b/backend/app/assets/config/solidus_backend_manifest.js
@@ -4,3 +4,5 @@
 //= link spree/backend/all.css
 
 //= link solidus_admin/select2_locales
+
+//= link_directory ../stylesheets/spree/backend/themes .css
--- a/backend/app/assets/stylesheets/spree/backend/components/index.scss
+++ b/backend/app/assets/stylesheets/spree/backend/components/index.scss
@@ -1,0 +1,15 @@
+@import 'spree/backend/components/actions';
+@import 'spree/backend/components/breadcrumb';
+@import 'spree/backend/components/buttons';
+@import 'spree/backend/components/number_with_currency';
+@import 'spree/backend/components/hint';
+@import 'spree/backend/components/image_placeholder';
+@import 'spree/backend/components/list_group';
+@import 'spree/backend/components/messages';
+@import 'spree/backend/components/progress';
+@import 'spree/backend/components/table-filter';
+@import 'spree/backend/components/navigation';
+@import 'spree/backend/components/sidebar';
+@import 'spree/backend/components/editable_table';
+@import 'spree/backend/components/pills';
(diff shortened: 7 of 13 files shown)


core/lib/spree/app_configuration.rb
```diff
@@ -140,6 +146,14 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Integer] Customer returns to show per-page in the admin (default: +15+)
     preference :customer_returns_per_page, :integer, default: 15
 
+    # @!attribute [rw] default_admin_light_theme
+    #   @return [String] Default admin light theme name
+    preference :default_admin_light_theme, :string, default: :classic
+
+    # @!attribute [rw] default_admin_dark_theme
+    #   @return [String] Default admin dark theme name
+    preference :default_admin_dark_theme, :string, default: :classic
```
kennyadsl, line 155: The dark theme is no more in this PR, right?

backend/app/views/spree/admin/shared/_head.html.erb
```diff
@@ -6,7 +6,15 @@
 <title><%= admin_page_title %></title>
 
 <%= favicon_link_tag 'favicon.ico' %>
-<%= stylesheet_link_tag 'spree/backend/all', media: 'all', data: {turbolinks_track: 'reload'} %>
+
+<% Spree::Config.admin_themes.each do |name, path| %>
+  <% media_queries = [] %>
+  <% media_queries << '(prefers-color-scheme: dark)' if Spree::Config.default_admin_dark_theme.to_s == name.to_s %>
+  <% media_queries << '(prefers-color-scheme: light)' if Spree::Config.default_admin_light_theme.to_s == name.to_s %>
```
kennyadsl, line 13: Same here, there's no dark theme yet if I got it correctly.

elia: Yes exactly, it's just the option to set different themes for when the system is in dark/light mode, similarly to how it works for other software (GitHub, Slack, etc).

rainerdema (approved):
Excellent work! 👏

backend/app/views/spree/admin/shared/_head.html.erb
kennyadsl: Got it. I think it's the right approach, but honestly, it still seems a bit out of context for this PR. Considering the end goal is to dismiss the old backend, I can't see much value in adding this feature now. 

It will take time to implement the full switch, so do you think it's essential to have it in the meantime? If you ask me, I personally don't, but at the same time, I don't think it will be the end of the world if we merge this enhancement so I won't block the PR for this.

elia: > do you think it's essential to have it in the meantime?

Yes, although "essential" is a stronger word than I would pick. Having been through many rewrites and migrations I know that holding on sensible and quality of life fixes doesn't pay off, since the turning point seems around the corner but usually proves to be just a mirage and the release date is shifted by another 3 months, and so on and so forth. 

This is also very portable to the new admin in which seems likely that we'll use [tailwind that lends itself very well to recompiling with custom palettes and themes](https://tailwindcss.com/docs/theme) that can be light and dark, allowing stores that are inclined to do so to match the admin to their brand.

Here's for example how GitHub allows you to pick a theme for each of dark and light modes:

image

backend/app/assets/stylesheets/spree/backend.css
```diff
@@ -3,9 +3,5 @@
  * and any sub-directories. You're free to add application-wide styles to this file and they'll appear at
  * the top of the compiled file, but it's generally better to create a new file per style scope.
```
waiting-for-dev, line 4: What do you think if we take the occasion to change this misleading comment? It's not true that the manifest will `automatically include all the stylesheets available in this directory and any sub-directories`. It'll only take what is referenced in `spree/backend/spree_admin`.

core/lib/spree/app_configuration.rb
```diff
@@ -42,6 +42,12 @@ class AppConfiguration < Preferences::Configuration
     #   @return [Integer] Number of products to display in admin (default: +10+)
     preference :admin_products_per_page, :integer, default: 10
 
+    # @!attribute [rw] admin_themes
+    #   @return [Hash] A hash containing the themes that are available for the admin panel
+    preference :admin_themes, :hash, default: {
```
waiting-for-dev, line 47: What do you think if we define those new settings in `Spree::BackendConfiguration`? I know that we have other `admin` related settings in `Spree::Config`, but we also have some of them in the backend file, and it seems to make more sense. Thoughts?

backend/app/views/spree/admin/shared/_head.html.erb
```diff
@@ -6,7 +6,15 @@
 <title><%= admin_page_title %></title>
 
 <%= favicon_link_tag 'favicon.ico' %>
-<%= stylesheet_link_tag 'spree/backend/all', media: 'all', data: {turbolinks_track: 'reload'} %>
+
+<% Spree::Config.admin_themes.each do |name, path| %>
```
waiting-for-dev, line 10: I'm a bit confused with this loop... Shouldn't we just render one media query, light or dark, for the active theme?

backend/app/views/spree/admin/shared/_head.html.erb
waiting-for-dev: I'm also a bit concerned that we don't need it right now. Plus, it introduces a warning to end users, but I won't block it here either if you think it can help! In the end, it's an improvement...

core/lib/spree/app_configuration.rb
waiting-for-dev: What if it's `nil` to better communicate it's not supported yet?

waiting-for-dev (approved):
Thanks, @elia! I left some non-blocking comments, and I'm also confused about how the media query is rendered in the HTML. :raised_hands:

core/lib/spree/app_configuration.rb
elia: @waiting-for-dev we still need to set a theme for when the system is in dark mode, and by default (like it is now) we'll use the same light theme for when the system is either in dark or light mode. Hope this makes sense.

backend/app/assets/stylesheets/spree/backend/vendor/index.scss
```diff
@@ -0,0 +1,11 @@
...
+@import "solidus_admin/flatpickr/flatpickr";
+@import "prism";
+
+@import 'spree/backend/bootstrap_custom';
+@import 'solidus_admin/bootstrap/bootstrap';
+@import 'font-awesome';
+
+// These mixins will modify existing Bootstrap ones and need to be imported
+// after the Bootstrap has been processed.
+@import 'spree/backend/globals/mixins';
```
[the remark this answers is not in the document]
elia: Yes thanks for noticing, we should be good, all my checks were ok, but I'll re-run the reasoning below:

- _globals_ only contain variables and Sass mixins, and select2, flatpicker and prism are raw css files, not affected by any Sass specific variable or function
- `globals/mixins` are overrides of bootstrap mixins, so it's just needs to come after bootstrap and before our solidus_backend specific CSS
- font awesome is also a `.css.erb` file, that is unaffected by anything else

I blocked myself from cleaning up further because this is already EOL, just did the bare minimum to make it easy to define themes as variations of the current style sheet (e.g. via variable adjustment).

backend/app/views/spree/admin/shared/_head.html.erb
elia: The idea is to have a single link tag in case the theme is the same for both modes, and two link tags if they differ:

pseudo html for when it's the same theme:

```html
<link media="light, dark" src="theme1.css"/>
```

and two tags when they differ:

```html
<link media="light" src="theme1.css"/>
<link media="dark" src="theme2.css"/>
```

--- 

But I'll look into it again and see if I can simplify and ditch the loop 👍

backend/app/views/spree/admin/shared/_head.html.erb
waiting-for-dev: I think I'd prefer to add it. Otherwise, it's like having something half-done. We should probably add a new section about how to create your own theme here: https://guides.solidus.io/customization/customizing-the-backend

waiting-for-dev (approved):

elia:
After live reviewing this with @kennyadsl we convened to just have a single configuration for enabling the new theme and handle dark mode in the original PR

backend/lib/spree/backend_configuration.rb
```diff
@@ -6,6 +6,22 @@ module Spree
   class BackendConfiguration < Preferences::Configuration
     preference :locale, :string, default: I18n.default_locale
 
+    # @!attribute [rw] themes
+    #   @return [Hash] A hash containing the themes that are available for the admin panel
+    preference :themes, :hash, default: {
+      classic: 'spree/backend/all',
+    }
+
+    FALLBACK_THEME = :classic
```
waiting-for-dev, line 15: I think we forgot to use the constant, as other references use `:classic` every time :slightly_smiling_face:

backend/lib/spree/backend_configuration.rb
```diff
@@ -6,6 +6,22 @@ module Spree
...
+
+    # @!attribute [rw] theme
+    #   @return [String] Default admin light theme name
+    preference :theme, :string, default: :classic
+
+    def theme_path
+      themes.fetch(theme.to_sym) { themes.fetch(:classic) }
```
waiting-for-dev, line 22: Given that `theme` will return `:classic` by default, probably there's no need to provide a block to `fetch`.

backend/lib/spree/backend_configuration.rb
```diff
@@ -6,6 +6,22 @@ module Spree
...
+    # @!attribute [rw] themes
+    #   @return [Hash] A hash containing the themes that are available for the admin panel
+    preference :themes, :hash, default: {
+      classic: 'spree/backend/all',
+    }
+
+    FALLBACK_THEME = :classic
+
+    # @!attribute [rw] theme
+    #   @return [String] Default admin light theme name
```
waiting-for-dev, line 18: The final work in this PR is agnostic to light/dark concepts. What do you think if we remove references to it (also in the specs below)?

waiting-for-dev (commented):
Thanks, @elia! I left a few nits.

backend/lib/spree/backend_configuration.rb
```diff
@@ -6,6 +6,27 @@ module Spree
...
+    def default_theme
+      :classic
+    end
+
+    def theme_path(user_theme = nil)
+      theme_path = themes.fetch(user_theme.to_sym, nil) if user_theme
+      theme_path ||= themes.fetch(theme.to_sym, nil) if theme
+      theme_path ||= themes.fetch(default_theme, nil)
+      theme_path
+    end
```
waiting-for-dev, line 28: I'm probably missing something, but it looks like too complicated logic. Shouldn't we have a `default: :classic` for the `theme` preference, and default the `user_theme` argument in the `theme_path` function to the `theme` preference value? :thinking:

backend/lib/spree/backend_configuration.rb
```diff
@@ -6,6 +6,27 @@ module Spree
   class BackendConfiguration < Preferences::Configuration
     preference :locale, :string, default: I18n.default_locale
 
+    # @!attribute [rw] themes
+    #   @return [Hash] A hash containing the themes that are available for the admin panel
+    preference :themes, :hash, default: {
```
kennyadsl, line 11: Not blocking, but if we can use versioned preferences here instead, we'll have the Solidus upgrade process automatically suggest people to use the new default starting from the next version. Did you already consider this?

kennyadsl (commented):
Left a couple of suggestions, thanks @elia!

backend/spec/lib/spree/backend_configuration_spec.rb
```diff
@@ -79,4 +79,41 @@ def product_path(product)
       end
     end
   end
+
+  describe '#theme_path' do
+    it 'returns the default light theme path' do
```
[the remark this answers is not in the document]
elia: Sure, I missed this!

backend/lib/spree/backend_configuration.rb
elia: @waiting-for-dev maybe I was too defensive here, I can remove some of the fallback layers and expect that if you set a theme that theme is available (maybe raise an exception via `fetch` if it's not).

backend/lib/spree/backend_configuration.rb
elia: Done!

backend/lib/spree/backend_configuration.rb
```diff
@@ -6,6 +6,20 @@ module Spree
...
+      classic: 'spree/backend/all',
+    }
+
+    # @!attribute [rw] theme
+    #   @return [String] Default admin theme name
+    versioned_preference :theme, :string, initial_value: 'classic', boundaries: { "4.1.0.a" => "classic" }
```
kennyadsl, line 17: Are we going to change this value when we'll introduce the new theme, right?

kennyadsl (approved):
👍 Thanks!

backend/lib/spree/backend_configuration.rb
elia: yes!

PR #5099, opened 2023-05-26
Use and extend view components in Solidus Admin

waiting-for-dev:
## Summary

We're introducing [ViewComponent](https://viewcomponent.org/) for the new admin along with proposing a way to extend them from the host app.

Please, go through each commit description for a more complete walk-through.

As a recap, admin components can be replaced by the host app by creating a new component with a matching path within the `app/components/{app_name}/solidus_admin/` directory. Example:

```ruby
# app/components/sandbox/solidus_admin/main_nav_item_component.rb

module Sandbox
  module SolidusAdmin
    class MainNavItemComponent < ViewComponent::Base
  
      attr_reader :item
  
      def initialize(item:)
        @item = item
      end
  
      erb_template <<~ERB
        <li style="background-color: red">
          <a href="#">
            <%= item.title %>
          </a>
        </li
      ERB
    end
  end
end
```

In the case of desired fine-grained control, the created component can inherit the original one. For instance, here we're just changing the template while keeping all the rest of view logic in place (and the opposite could also be possible):

```ruby
# app/components/sandbox/solidus_admin/main_nav_item_component.rb

module Sandbox
  module SolidusAdmin
    class MainNavItemComponent < ::SolidusAdmin::MainNavItemComponent
      erb_template <<~ERB
        <li style="background-color: red">
          <a href="#">
            <%= item.title %>
          </a>
        </li
      ERB
    end
  end
end
```

Unfortunately, view_component[ doesn't allow overriding templates](https://github.com/ViewComponent/view_component/issues/411) from the host application by just having the same relative path. For now, we'll use the [inline templates](https://viewcomponent.org/guide/templates.html#inline) feature

Worth noting that we're injecting nested components as initialization dependencies for their parent components instead of going with slots. That's to avoid the problem with [deeply nested slots](https://github.com/ViewComponent/view_component/discussions/867) and also to make customization easier by just overriding the initializer if needed:

```ruby
# app/components/sandbox/solidus_admin/main_nav_component.rb

module Sandbox
  module SolidusAdmin
    class MainNavComponent < ::SolidusAdmin::MainNavComponent
      def initialize(main_nav_item_component: component("my_main_nav_item"), **kwargs)
        super
      end
    end
  end
end
```
``
The same registry that is used to keep track of the components is also used to contain data entities, like the one used to wrap the main navigation items. For instance, a new nav item could be registered like this:

```ruby
require "solidus_admin/nav_item"
stats = SolidusAdmin::Stats.new(title: "Stats", position: 15)
SolidusAdmin::Container.register("main_nav.stats", stats)
```

Not showcased now, but we can also use the registry for the nav item itself, so users would have also full control over them.

## Checklist


admin/app/components/solidus_admin/main_nav_component.rb
```diff
@@ -3,5 +3,14 @@
 module SolidusAdmin
   # Renders the main navigation of Solidus Admin.
   class MainNavComponent < BaseComponent
+    erb_template <<~ERB
+      <nav>
+        <%=
+          render component("main_nav_item").with_collection(
+            container.all("main_nav").sort_by(&:position)
```
tvdeyen: What do you think about?

```rb
components.all("main_nav")...
```

as helper for the `container` implementation detail?

admin/app/components/solidus_admin/base_component.rb
```diff
@@ -4,13 +4,6 @@ module SolidusAdmin
...
     include ViewComponent::InlineTemplate
-
-    def container
-      SolidusAdmin::Container
-    end
-
-    def component(name)
-      container.resolve("#{name}_component")
-    end
+    include SolidusAdmin::ContainerHelper
```
tvdeyen, line 7: What about `SolidusAdmin::ComponentsHelper`? I would like to avoid adding a technical implementation detail to our public API. I prefer naming that explains its usecase rather than its implementation.

admin/app/helpers/solidus_admin/container_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+
+require "solidus_admin/container"
+
+module SolidusAdmin
+  module ContainerHelper
+    def container
+      SolidusAdmin::Container
+    end
+
+    def component(name)
```
tvdeyen, line 11: 👏🏻 I like this as this hides the implementation.

admin/lib/solidus_admin/system/providers/main_nav.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "solidus_admin/container"
+require "solidus_admin/main_nav_item"
+
+module SolidusAdmin
+  module Providers
```
tvdeyen, line 7: Maybe a `Components` namespace would help here as well to understand the usecase of this "Provider" a bit better?

admin/app/components/solidus_admin/main_nav_component.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module SolidusAdmin
+  # Renders the main navigation of Solidus Admin.
+  class MainNavComponent < BaseComponent
+    def initialize(main_nav_item_component: component("main_nav_item"), items: container.within_namespace("main_nav"))
```
tvdeyen, line 6: If we would add a `components` helper that uses `container.within_namespace` internally, we could make this much more userfriendly and would hide the technical detail of a `Dry::Container` from the user.

```suggestion
    def initialize(main_nav_item_component: component("main_nav_item"), items: components("main_nav"))
```

WDYT?

tvdeyen (commented):
Yeah `view_component` 🎉 

I really like that we are going this route 🥳 

First of all I want to emphasize the great writing and git work on this PR. This is how PRs and commits should be written. It has been a pleasure to review and future version of ourslevs will thank you for this. 🎖️ 

I have some thoughts on naming.

admin/lib/solidus_admin/container.rb
```diff
@@ -0,0 +1,23 @@
...
+
+require "dry/system"
+require "dry/system/container"
+
+module SolidusAdmin
+  # Global registry for host-injectable components.
+  #
+  # We use this container to register all the components that can be
+  # overridden by the host application.
+  class Container < Dry::System::Container
```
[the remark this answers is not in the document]
waiting-for-dev: The issue is that, probably, we'll use the container for registering other kinds of dependencies. For instance, if, at some point, we introduce service objects only for the admin panel, they would also need to be registered here. In fact, we're already using it for an entity-like structure: the data wrapped within a `SolidusAdmin::MainNavItem` shouldn't be seen as something that only belongs to components.

admin/lib/solidus_admin/main_nav_item.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module SolidusAdmin
+  # Encapsulates the data for a main nav item.
+  class MainNavItem
```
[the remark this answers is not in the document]
waiting-for-dev: As pointed in https://github.com/solidusio/solidus/pull/5099#discussion_r1207065776, a `MainNavItem` is just something similar to a data object. We probably don't want to see it as something that can only belong to a view component. For instance, if at some point we dedice to persist them, they could become database records.

admin/app/components/solidus_admin/main_nav_component/main_nav_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <nav>
   <%=
-    render SolidusAdmin::MainNavItemComponent.with_collection(
+    render container["main_nav_item_component"].with_collection(
```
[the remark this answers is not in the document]
waiting-for-dev: Fixed in subsequent commits :slightly_smiling_face:

admin/app/helpers/solidus_admin/container_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "solidus_admin/container"
+
+module SolidusAdmin
+  module ContainerHelper
+    def container
+      SolidusAdmin::Container
```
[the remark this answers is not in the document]
waiting-for-dev: I also have some doubts about the name, what do you think about `SolidusAdmin::Registry`?

admin/app/components/solidus_admin/base_component.rb
```diff
@@ -6,5 +6,9 @@ class BaseComponent < ViewComponent::Base
     def container
       SolidusAdmin::Container
     end
+
+    def component(name)
```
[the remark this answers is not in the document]
waiting-for-dev: See https://github.com/solidusio/solidus/pull/5099#discussion_r1207108938

admin/app/components/solidus_admin/main_nav_component.rb
waiting-for-dev: See https://github.com/solidusio/solidus/pull/5099#discussion_r1207108938

admin/app/components/solidus_admin/base_component.rb
waiting-for-dev: See https://github.com/solidusio/solidus/pull/5099#discussion_r1207108938

admin/lib/solidus_admin/system/providers/main_nav.rb
waiting-for-dev: See https://github.com/solidusio/solidus/pull/5099#discussion_r1207065776

admin/app/components/solidus_admin/main_nav_component.rb
waiting-for-dev: Please, notice that `container.within_namespace("main_nav")` is not resolving `ViewContainer` classes nor instances, but instances of `SolidusAdmin::MainNavItem` entities. I don't think it's a big problem exposing the container (happy to rename it to registry) to the components, as it's precisely intended to resolve dependencies. However, I agree that there's a problem having it also available in the templates, but I can't see a way to avoid that, as it's more of a `view_component` limitation. What do you think?

On top of that, I'd love to use a more declarative specification of dependencies via [`dry-system`'s auto-injection](https://dry-rb.org/gems/dry-system/1.0/dependency-auto-injection/) feature. However, that would require having to call `super` everytime we define the `initialize` method for a component. I didn't go that way because I think it's good to stick as much as we can to bare `view_components` so that ours users don't have a hard time extending our components.

```ruby
module SolidusAdmin
  class MainNavComponent < BaseComponent
    include Import["main_nav_item_component"]

    def initialize(items: container.within_namespace("main_nav"))
      super
      @items = items
    end

    erb_template <<~ERB
      <nav>
        <%=
          render main_nav_item_component.with_collection(
            sorted_items
          )
        %>
      </nav>
    ERB

    private

    def sorted_items
      @items.sort_by(&:position)
    end
  end
end
```

waiting-for-dev:
> I really like that we are going this route partying_face

Happy to hear that! I hope our broader community will feel the same :slightly_smiling_face: 

> First of all I want to emphasize the great writing and git work on this PR. This is how PRs and commits should be written. It has been a pleasure to review and future versions of ourselves will thank you for this. medal_military

Thanks for your nice words :bow: :slightly_smiling_face:

admin/app/components/solidus_admin/main_nav_component.rb
```diff
@@ -0,0 +1,27 @@
...
+      @main_nav_item_component = main_nav_item_component
+      @items = items
+    end
+
+    erb_template <<~ERB
+      <nav>
```
kennyadsl, line 12: Is tailwind able to catch the classes of these HTML elements even if they are in the .rb file?

admin/lib/solidus_admin/container.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require "dry/system"
+require "dry/system/container"
+require "dry/system/component"
```
kennyadsl, line 5: I'm not against adding this dependency, but just to be aware of all the pieces of the tradeoff we are doing here, do you have an idea of the effort needed to create something similar internally? I have the feeling it's not a giant effort, and maybe it might make sense only if we use small parts of what's provided in dry container/component; interested in your take on this.

admin/lib/solidus_admin/system/providers/main_nav.rb
```diff
@@ -0,0 +1,18 @@
...
+    Container.register_provider("main_nav") do
+      start do
+        container.namespace("main_nav") do
+          register("first_item", MainNavItem.new(title: "First item", position: "10"))
```
kennyadsl, line 11: Do you envision any problem with when we'll need to use I18n here for the title?

kennyadsl (commented):
Thanks @waiting-for-dev, I left some questions before approving, but I'm overall excited to merge this!

admin/app/components/solidus_admin/main_nav_component.rb
waiting-for-dev: We should be fine if we add the `components` directory to the Tailwind configuration. [Tailwind uses ](https://tailwindcss.com/docs/content-configuration#class-detection-in-depth)a very simple regexp matching to extract the needed classes.

admin/lib/solidus_admin/system/providers/main_nav.rb
waiting-for-dev: I don't think so, as `I18n` acts globally. Alternatively, we could use the I18n key and translate from the component.

admin/lib/solidus_admin/container.rb
kennyadsl: Makes sense, let's re-evaluate if we really need it once we are ok with the architecture and we see it in action. Thanks!

kennyadsl (approved):
Approving, but there's an issue with the installer related to the version of the `solidus_admin`. Do we already have a fix for that in mind?

waiting-for-dev:
> Approving, but there's an issue with the installer related to the version of the `solidus_admin`. Do we already have a fix for that in mind?

Not sure what's going on. I tried to reproduce locally but it worked. I'll look into it.

waiting-for-dev:
> Approving, but there's an issue with the installer related to the version of the `solidus_admin`. Do we already have a fix for that in mind?

@kennyadsl, bumping the cache version on the CI to fetch the latest changes on SSF to allow `view_components ~> 3.0` fixed the issue (ref: https://github.com/solidusio/solidus_starter_frontend/pull/343)

admin/app/components/solidus_admin/main_nav_component.rb
waiting-for-dev: FYI, I rebased from the admin branch, which now contains tailwindcss, and pushed a new commit adding the component directories to the list of scanned paths.

kennyadsl: 💯

elia (commented):
It's great to see this coming along!

Left a couple of comments, mainly food for thought, happy to discuss it with a shared screen!

admin/lib/solidus_admin/container.rb
```diff
@@ -0,0 +1,32 @@
...
+  #
+  # @api private
+  class Container < Dry::System::Container
+    configure do |config|
+      config.root = Pathname(__FILE__).dirname.join("../..").realpath
+      config.component_dirs.add("app/components") do |dir|
+        dir.loader = System::Loaders::HostOverridableConstant.method(:call).curry["components"]
+        dir.namespaces.add "solidus_admin", key: nil
+      end
+    end
```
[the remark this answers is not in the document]
waiting-for-dev: Hey @elia, as pointed out in a [previous comment](https://github.com/solidusio/solidus/pull/5099#discussion_r1210069301), there's definitely nothing we couldn't build on our own. However, for the sake of this spike, I wanted to focus on the architecture and the interface to interact with it instead of getting lost in implementation details. dry-system allowed me to do that, but I have no problems revisiting it and evaluating if it's enough to build something like a Hash-based registry.

I'm not sure which direction we're going to take on other topics, like service objects on the new admin, but having dry-system in place can help us with moving and changing things faster, as its declarative configuration is very flexible. Then, it can also give us other valuable features like logging, monitoring, or testing support (stubbing a component while testing). We can assess how much of it we end up using, and then decide whether to go with or without it. What do you think?

admin/app/views/layouts/solidus_admin/application.html.erb
```diff
@@ -5,6 +5,7 @@
   </head>
   <body>
     <h1 class="text-4xl">Layout is rendered</h1>
+    <%= render component("main_nav").new %>
```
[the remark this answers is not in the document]
waiting-for-dev: Hey @elia, I pointed out some of the befits and features that dry-system provides in [another answer](https://github.com/solidusio/solidus/pull/5099#discussion_r1217434038).

> <%= render component("main_nav").new(items: SolidusAdmin::NavItem.all) %>

I think it's good having the default value for `items` already defined on the initializer of `MainNavComponent`. In this case, it doesn't make such a difference, but maybe other components will be reused, which would avoid boilerplate code. In that situation, dry-system's [auto_inject](https://dry-rb.org/gems/dry-system/1.0/dependency-auto-injection/) would also help.

> class SolidusAdmin::NavItem < ActiveModel::Model
>  def self.all
>    @all ||= SolidusAdmin::Config.nav_items.map { new _1 }
>  end
> end

I know that's not a formal proposal, but regardless of using dry-system or not, I'd go with a PORO for nav item entities. In the end, we don't need to deal with validations, which would be a case for `ActiveModel::Model`.

> SolidusAdmin::Config.nav_items += [
>  { name: "Extra item", icon: :gift, … }
> ]

I'd also like to avoid having `SolidusAdmin::Config` as a hodgepodge. For anything else that configuring simple primitives, I think it'd be good having a richer interface, whether that one is provided by dry-system or something we own.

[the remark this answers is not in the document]
waiting-for-dev: @elia, please, take a look at the last commit I pushed (__Auto-inject dependencies into components__). I also included [`dry-sytem`'s auto_inject](https://dry-rb.org/gems/dry-system/1.0/dependency-auto-injection/) within this spike, to better showcase both how having defaults defined in the `#initialize` method, and having nav items registered in the container help with avoiding boilerplate. Also pinging @kennyadsl.

admin/app/components/solidus_admin/main_nav_component.rb
waiting-for-dev: Finally, I also added auto-injection to this spike. In the end, `super` needs to be called whenever you override an initializer, and rubocop will complain if you don't.

admin/lib/solidus_admin/container.rb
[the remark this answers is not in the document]
elia: Maybe it's unfamiliarity with DRY, but looks like for our case it's an additional layer of injections on top of the already sufficient initializer params, which I also find more explicit.

I think having a more complete context might help in appreciating pros and cons of each solution, thoughts on holding a little bit on merging and keeping this PR as a reference maybe coming back to it once we start implementing? 

_I think we can merge right away the first commits setting up the Base component class_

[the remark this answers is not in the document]
waiting-for-dev: After talking offline, we decided to merge everything now and re-evaluate along the way :+1:

PR #5100, opened 2023-05-29
Deprecate Paperclip Adapter

### LINKED ISSUE #5033: Consider removing Paperclip adapter
We're not supporting Rails v5.2 anymore, so all stores using Solidus can use the newer ActiveStorage adapter. However, we might first want to hear about the experience of stores using it. One option can be extracting it to a separate gem and archive it.

kennyadsl:
## Summary

Closes #5033

Add `solidus:paperclip_adapter:install` generator, which allows incorporating the paperclip adapter into the host application, in order to remove support for it from the core.

This generator does two things:

1. Copy the Image and Taxon Paperclip adapters into the host application;
2. Set the right configurations in the spree.rb (path configurable) initializer in order to use the newly created files.

This PR also deprecates the Paperclip adapters, so that we can remove it in the next major version.

### TODOs

- [ ] install gem on the host  app's Gemfile because we will remove it from dependencies.

## Checklist

### DIFF
--- a/core/lib/generators/solidus/paperclip_adapter/install/install_generator.rb
+++ b/core/lib/generators/solidus/paperclip_adapter/install/install_generator.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'rails/generators'
+
+module Solidus
+  # @private
+  module PaperclipAdapter
+    class InstallGenerator < Rails::Generators::Base
+      source_root File.expand_path('templates', __dir__)
+
+      class_option :set_preferences,
+        type: :boolean,
+        default: true,
+        banner: 'Wheter or not to set the preferences in the initializer for the attachment adapters'
+
+      class_option :app_directory,
+        type: :string,
+        default: Rails.root,
+        hide: true
+
+      class_option :initializer_directory,
+        type: :string,
+        default: 'config/initializers/',
+        hide: true
+
+      class_option :initializer_basename,
+        type: :string,
+        default: 'spree',
+        banner: 'The name of the initializer containing the attachment adapters settings'
+
+      class_option :app_name,
+        type: :string,
+        default: Rails.application.class.module_parent.name,
+        banner: 'The name of the host application'
+
+      def copy_templates
+        say_status :template, "Paperclip Adapters", :green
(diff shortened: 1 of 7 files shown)


core/lib/generators/solidus/paperclip_adapter/install/install_generator.rb
```diff
@@ -0,0 +1,53 @@
...
+        hide: true
+
+      class_option :initializer_basename,
+        type: :string,
+        default: 'spree',
+        banner: 'The name of the initializer containing the attachment adapters settings'
+
+      class_option :app_name,
+        type: :string,
+        default: 'MyStore',
```
waiting-for-dev, line 28: I think we can automate that with `Rails.application.class.module_parent.name`

waiting-for-dev (commented):
I think we need to change CI for the paperclip adapter so the generator is run, right? :thinking:

core/lib/generators/solidus/paperclip_adapter/install/install_generator.rb
kennyadsl: It was on my metal TODO list, and of course, I forgot about it. Thanks!

kennyadsl:
> I think we need to change CI for the paperclip adapter so the generator is run, right? 🤔

At the moment, the paperclip CI jobs should use the old adapter, which I don't think it's working because tests are failing and I don't understand why. Your proposed approach seems interesting, this way we'll stop using the deprecated adapter and just rely on the generated files. I'll give it a try.

core/lib/generators/solidus/paperclip_adapter/install/install_generator.rb
waiting-for-dev: > metal TODO list

:metal:

kennyadsl: lol 🤘

jarednorman:
We should do this, but seeing as there's an open issue for it and this PR hasn't been touched in three years, I'm closing it. I'll add the issue to the roadmap so we address this for Solidus 5.

PR #5101, opened 2023-05-30
Ensure to return false on Paperclip attachment destroy failure

RyanofWoods:
## Summary
I recently changed the `destroy_attachment` method, to save the Spree::Taxon instance afterwards.  However, this caused it to always return `true`, regardless of if the attachment was actually destroyed. Causing potential incorrect feedback to the user.

---

If the attachment does exist, destroying the attachment will be attempted. However, destroying is not guaranteed by Paperclip. In case of failure, it will return `false` [1][2]. It should therefore return `false` in this case, so callees know it failed. Before, due to the `save` call afterwards, it was always returning `true`.

[1] https://github.com/kreeti/kt-paperclip/blob/v7.1.1/lib/paperclip/attachment.rb#L263
[2] https://github.com/kreeti/kt-paperclip/blob/v7.1.1/lib/paperclip/attachment.rb#L234-L235

## Checklist

### DIFF
--- a/core/spec/models/spree/taxon_spec.rb
+++ b/core/spec/models/spree/taxon_spec.rb
@@ -27,6 +27,11 @@ RSpec.describe Spree::Taxon, type: :model do
         end
 
         if Spree::Config.taxon_attachment_module == Spree::Taxon::PaperclipAttachment
+          it "returns false if destroying the attachment fails" do
+            allow(taxon.icon).to receive(:destroy).and_return(false)
+            expect(taxon.destroy_attachment(:icon)).to be_falsey
+          end
+
           it "resets paperclip attributes when using Paperclip", aggregate_failures: true do
             expect(taxon.destroy_attachment(:icon)).to be_truthy
             expect(taxon.reload.icon_file_name).to_not be_present
--- a/core/app/models/spree/taxon/paperclip_attachment.rb
+++ b/core/app/models/spree/taxon/paperclip_attachment.rb
@@ -25,7 +25,6 @@ module Spree::Taxon::PaperclipAttachment
     attached_file = send(definition)
     return false unless attached_file.exists?
 
-    attached_file.destroy
-    save
+    attached_file.destroy && save
   end
 end


core/spec/models/spree/taxon_spec.rb
```diff
@@ -27,6 +27,11 @@
         end
 
         if Spree::Config.taxon_attachment_module == Spree::Taxon::PaperclipAttachment
+          it "returns false if destroying the attachment fails" do
+            allow(taxon.icon).to receive(:destroy).and_return(false)
+            expect(taxon.destroy_attachment(:icon)).to be_falsey
```
kennyadsl, line 32: Just to be sure, do you think it's worth checking if the file_name is still there?

```
expect(taxon.reload.icon_file_name).to_not be_present
```

kennyadsl (approved):
Left a question, but 👍 as is as well. Thanks for taking care of this!

core/spec/models/spree/taxon_spec.rb
RyanofWoods: @kennyadsl the paperclip attributes such as `icon_file_name` will still be present because the destroy failed and the attachment still exists. I don't think it's necessary to test this, because paperclip does the setting of these attributes to `nil` on destroying. If Solidus did this, I think it would be necessary.

I am happy to add it if you feel strong about it, and I am interested to hear more opinions about it.

kennyadsl: Sorry I meant:

```
expect(taxon.reload.icon_file_name).to be_present
```

By the way, I got your point. If you think this is already covered by Paperclip, we are good as is. 👍

kennyadsl:
@RyanofWoods merged, but I have a follow-up question somewhat unrelated: do you know why we don't have that method defined in the [Image's paperclip_attachment](https://github.com/solidusio/solidus/blob/4cf0ca494d8d5c2396ad4ef596af9425ba67d8ee/core/app/models/spree/image/paperclip_attachment.rb)?

PR #5129, opened 2023-06-07
Integrate Role and Permission Management from `solidus_user_roles` (solidus_core step)

### LINKED ISSUE #5105: Incorporate roles management code into the core

rainerdema:
## Summary

  Fixes #5105

This PR integrates role and permission management capabilities from the [solidus_user_roles](https://github.com/boomerdigital/solidus_user_roles) gem directly into Solidus, providing greater flexibility and control over user roles and permissions.

Major changes include:
1. **`Permission Sets`**: A new `Spree::PermissionSet` model has been added to the database schema. This model represents sets of permissions that can be assigned to roles. Each permission set has a name and a set identifier.

2. **`Role Permissions`**: A new `Spree::RolePermission` model has been added to the database schema. This model represents the association between roles and permission sets.

3. **`Data Migration`**: [Pull Request](https://github.com/nebulab/solidus/pull/361)
This rake task creates `Spree::PermissionSet` records for each existing permission set subclass in `Spree::PermissionSets::*`, and `Spree::RolePermission` records for each role's associated permission sets. 
(It is designed to work with both new installations and existing Solidus stores that have custom roles and permissions).
```
bin/rails solidus:import_existing_permission_sets
```

4. ~**`Backend interface`**: [closed](https://github.com/nebulab/solidus/pull/362)~

5. ~**`Admin interface`**: Will only be merged into `nebulab/admin` branch~

## Checklist

### DIFF
--- a/core/app/models/spree/permission_set.rb
+++ b/core/app/models/spree/permission_set.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module Spree
+  class PermissionSet < Spree::Base
+    has_many :role_permissions
+    has_many :roles, through: :role_permissions
+
+    validates :name, :group, presence: true
+  end
+end
--- a/core/app/models/spree/role.rb
+++ b/core/app/models/spree/role.rb
@@ -2,13 +2,29 @@
 
 module Spree
   class Role < Spree::Base
+    RESERVED_ROLES = ['admin', 'default'].freeze
+
     has_many :role_users, class_name: "Spree::RoleUser", dependent: :destroy
+    has_many :role_permissions, dependent: :destroy
+    has_many :permission_sets, through: :role_permissions
     has_many :users, through: :role_users
 
+    scope :non_base_roles, -> { where.not(name: RESERVED_ROLES) }
+
     validates_uniqueness_of :name, case_sensitive: true
+    validates :name, uniqueness: true
+    after_save :assign_permissions
 
     def admin?
       name == "admin"
+    end
+
+    def permission_sets_constantized
+      permission_sets.map(&:name).map(&:constantize)
+    end
+
+    def assign_permissions
+      ::Spree::Config.roles.assign_permissions name, permission_sets_constantized
     end
   end
 end
--- a/core/db/migrate/20230607100048_create_spree_permission_sets.rb
+++ b/core/db/migrate/20230607100048_create_spree_permission_sets.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateSpreePermissionSets < ActiveRecord::Migration[7.0]
+  def change
+    create_table :spree_permission_sets do |t|
+      t.string :name
+      t.string :group
+      t.timestamps
+    end
+  end
+end
--- a/core/app/models/spree/role_permission.rb
+++ b/core/app/models/spree/role_permission.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module Spree
+  class RolePermission < Spree::Base
+    belongs_to :role
+    belongs_to :permission_set
+  end
+end
--- a/core/db/migrate/20230607100109_create_spree_roles_permissions.rb
+++ b/core/db/migrate/20230607100109_create_spree_roles_permissions.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateSpreeRolesPermissions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :spree_role_permissions do |t|
+      t.references :role
+      t.references :permission_set
+      t.timestamps
+    end
+  end
+end
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -605,6 +605,10 @@ module Spree
       @roles ||= Spree::RoleConfiguration.new.tap do |roles|
         roles.assign_permissions :default, ['Spree::PermissionSets::DefaultCustomer']
         roles.assign_permissions :admin, ['Spree::PermissionSets::SuperUser']
+
+        Spree::Role.non_base_roles.each do |role|
+          roles.assign_permissions role.name, role.permission_sets_constantized
+        end
       end
     end
 
--- a/core/lib/tasks/solidus/import_existing_permission_sets.rake
+++ b/core/lib/tasks/solidus/import_existing_permission_sets.rake
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+namespace :solidus do
+  desc "Import existing permission sets to role permissions table"
+  task import_existing_permission_sets: :environment do
+    Zeitwerk::Loader.eager_load_all unless Rails.env.test?
+
+    ActiveRecord::Base.transaction do
+      Spree::PermissionSets::Base.descendants.each do |permission|
+        Spree::PermissionSet.find_or_create_by(name: permission.to_s, group: permission.to_s.split("PermissionSets::").last.gsub(/Display|Management/i, ""))
+      end
+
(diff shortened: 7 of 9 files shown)


core/db/migrate/20230607100229_import_existing_permission_sets.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class ImportExistingPermissionSets < ActiveRecord::Migration[7.0]
```
kennyadsl, line 3: Did you consider adding this code as a rake task instead? Not everyone will need it and if we can save a migration I think it's generally a good principle to follow. 

Also because they still need to remove the subclasses manually after the migration is run otherwise the permissions will be loaded twice (?).

We could make sure people run it with some check at rails startup, informing that they can and should run the command (in all environments) and after that, redeploy their code with the subclasses removed. 

Even better if we can make subclasses and `Spree::Role` instances live together, giving priority to the latter so that they don't have to rush deleting the subclasses. 

Does it make sense?

kennyadsl (commented):
Thanks @rainerdema! I know it's a draft but I was reading and I thought it was worth commenting. Other than that, what about moving the migration part (which seems quite complex) to a subsequent PR?

core/db/migrate/20230607100229_import_existing_permission_sets.rb
rainerdema: Thank you for your feedback @kennyadsl!

> Does it make sense?

Yes! I agree with your points and believe your suggestions will make the transition smoother.

So to recap your suggestions:
- Moving the migration to a subsequent PR
- Rake task instead of migration
- Startup check with a message informing devs about the rake task (+ subclasses to remove)
- Coexistence of subclasses and `Spree::Role` Instances

>Also because they still need to remove the subclasses manually after the migration is run otherwise the permissions will be loaded twice (?).

I'll check but probably yes 🤔

core/app/models/spree/permission_set.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    validates :name, :set, presence: true
+
+    scope :display_permissions, -> { where('name LIKE ?', '%Display') }
+    scope :management_permissions, -> { where('name LIKE ?', '%Management') }
+
+    scope :custom_permissions, -> {
+      where.not(id: display_permissions)
+      .where.not(id: management_permissions)
+    }
```
waiting-for-dev, line 16: What do you think about testing these scopes?

core/app/models/spree/role.rb
```diff
@@ -3,12 +3,26 @@
 module Spree
   class Role < Spree::Base
     has_many :role_users, class_name: "Spree::RoleUser", dependent: :destroy
+    has_many :role_permissions, dependent: :destroy
+    has_many :permission_sets, through: :role_permissions
     has_many :users, through: :role_users
 
+    scope :non_base_roles, -> { where.not(name: ['admin']) }
```
waiting-for-dev, line 10: Same as above, should we add tests here?

core/app/models/spree/role.rb
```diff
@@ -3,12 +3,26 @@
...
     def admin?
       name == "admin"
     end
+
+    def permission_sets_constantized
+      permission_sets.map(&:set).map(&:constantize)
+    end
+
+    def assign_permissions
+      ::Spree::Config.roles.assign_permissions name, permission_sets_constantized
```
waiting-for-dev, line 25: If I'm not missing something, that will only update the permissions in-memory, and it won't survive an application reload https://github.com/solidusio/solidus/blob/83a62431494d8a8ed0c241247c43d3dc37bba98d/core/lib/spree/core/role_configuration.rb#L67.

core/db/migrate/20230607100048_create_spree_permission_sets.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class CreateSpreePermissionSets < ActiveRecord::Migration[7.0]
+  def change
+    create_table :spree_permission_sets do |t|
+      t.string :name
+      t.string :set
```
waiting-for-dev, line 7: I still fail to understand why we need both `name` and `set`. Don't they represent the same information? :thinking:

waiting-for-dev (commented):
Thanks, @the-krg! :raised_hands: I left some comments. Besides, it looks like a merge commit slipped in the changeset.

core/db/migrate/20230607100048_create_spree_permission_sets.rb
the-krg: While going through and writing the scope tests this I've noticed a flaw related to this question:
The scope searches by the `name`, meaning that the permission must end with "Display" or "Management" on its `name` attribute in order to be shown by the scope.

I've changed the scope to search by the `set`. This field represents the class name with its namespaces. Which is the key.

The `name` should be something for labeling purposes only, especially for custom permissions. 
Eg.: a user creates a custom permission named "Read/Edit-only custom data" as label on a permission that only allows editing and not managing.

What do you think?

core/app/models/spree/role.rb
the-krg: You're right! 
This is an important thing that I've added on the [backend interface pull request](https://github.com/nebulab/solidus/pull/362/files#diff-8f107b502c1d91ce09dcb78e3514f1e4503a7276f98a82a5f545281843e3e34a) that I forgot (and must) bring to this one!

core/db/migrate/20230607100048_create_spree_permission_sets.rb
the-krg: I'm also changing `set` to something else. It's a reserved word for many databases, and escaping it doesn't feel right.

the-krg: Update:

I'm changing the `set` attribute to `name` and adding a new attribute named `group` (or label...? WDYT?).

1. `set` is a reserved word for some databases;
2. `name` will now take place as `set` was before; (eg.: `name: "Spree::PermissionSets::UserDisplay"`)
3. `group` will be used for grouping the permissions; (eg.: `Spree::PermissionSet.where(group: "Users")` will return all permissions related to `Users`);
4.  when presenting the permissions for the user when creating a role, we'll displaying the groups with their permissions such as Edit, View or Custom.

Image 31-08-23 at 04 08

waiting-for-dev: Thanks, it makes sense

core/app/models/spree/role.rb
```diff
@@ -3,12 +3,26 @@
 module Spree
   class Role < Spree::Base
     has_many :role_users, class_name: "Spree::RoleUser", dependent: :destroy
+    has_many :role_permissions, dependent: :destroy
+    has_many :permission_sets, through: :role_permissions
     has_many :users, through: :role_users
 
+    scope :non_base_roles, -> { where.not(name: ['admin', 'default']) }
```
rainerdema, line 10: What do you think about storing the reserved role names like 'admin' and 'default' in a constant? 🤔 like `RESERVED_ROLES` or `STATIC_ROLES` something like that.

rainerdema (commented):
Great work! 👏  Thanks, @the-krg! ✅

elia:
I'm putting this on <kbd>hold</kbd> until we figure out a few things on how it will be used in the new admin dashboard.

fthobe:
@jarednorman I assume old and I do not know if @MadelineCollier carried this over to the new admin, so I'd close it depending on her reply

fthobe:
@jarednorman is this superseeded by @MadelineCollier ?

jarednorman:
I'm closing this as it's extremely stale, but I'm opening to revisiting and having a discussion about this.

PR #5131, opened 2023-06-08
Prepare admin container for sidecar components

waiting-for-dev:
## Summary

We prepare the admin container to resolve components from within sidecar
directories and we namespace all components within a `components` key.

Example, from the engine directory:

    app/components/solidus_admin/main_nav/component.rb

gets registered as:

    components.main_nav.component

and can be overriden by the following in the host application:

    app/components/sandbox/solidus_admin/main_nav/component.rb

## Checklist

### DIFF
--- a/admin/app/helpers/solidus_admin/container_helper.rb
+++ b/admin/app/helpers/solidus_admin/container_helper.rb
@@ -9,7 +9,7 @@ module SolidusAdmin
     end
 
     def component(name)
-      container.resolve("#{name}_component")
+      container.resolve("components.#{name}.component")
     end
   end
 end
--- a/admin/app/components/solidus_admin/main_nav/component.rb
+++ b/admin/app/components/solidus_admin/main_nav/component.rb
@@ -2,9 +2,9 @@
 
 module SolidusAdmin
   # Renders the main navigation of Solidus Admin.
-  class MainNavComponent < BaseComponent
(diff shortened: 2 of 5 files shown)


admin/app/helpers/solidus_admin/container_helper.rb
```diff
@@ -9,7 +9,7 @@ def container
     end
 
     def component(name)
-      container.resolve("#{name}_component")
+      container.resolve("components.#{name}.component")
```
elia, line 12: Does it make sense to change the keys to use `/` instead of `.` in the context of this PR or it's better to use a different one?

waiting-for-dev: Thanks for your feedback. Let's talk about it in a different place. I'm a bit reticent, as the identifier doesn't need to map one-to-one to a filesystem path, so it could lead to confusion :thinking: Actually, that's already the case, as `components.main_nav.component` maps to `components/solidus_admin/main_nav/component.rb`.

tvdeyen:
Just FYI: Do we support to allowing components from extensions as well?

waiting-for-dev:
> Just FYI: Do we support to allowing components from extensions as well?

Technically, that would be possible, as they could register their own components. Let's see if that would be a good idea :slightly_smiling_face:

tvdeyen:
I think it is fine to register them.

PR #5134, opened 2023-06-08
[admin] Use slash as the component registry key separator

elia:
## Summary

With this change referencing components will be more familiar to rails developers used to partial paths and by default retains a clear mapping between the file-system and the keys.

Also keeps all components from a section under the same namespace in order to reduce noise whenever the number of components will start to grow.

## Checklist


admin/app/components/solidus_admin/main_nav/component.rb
```diff
@@ -1,27 +1,18 @@
...
-        <%=
-          render main_nav_item_component.with_collection(
-            sorted_items
-          )
-        %>
-      </nav>
-    ERB
+  erb_template <<~ERB
+    <nav>
+      <%= render component('main_nav/item').with_collection(sorted_items) %>
```
waiting-for-dev, line 9: Caution! By calling the component from here, we're breaking the possibility of injecting new components on initialization. It'd be nice for a host application to be able to switch which nested component a parent component renders by just changing the initializer and without the need to touch the parent template. That would help for situations where more than one parent can use a nested component, and you only want to do the substitution in one place. In general, it would make the components less coupled and make testing easier and faster.

I'll prioritize working on the "Readme" about component extensibility to clarify the intentions behind using those patterns here. I'll try to have it ready by the beginning of next week :crossed_fingers:

admin/app/helpers/solidus_admin/container_helper.rb
```diff
@@ -9,6 +9,9 @@ def container
     end
 
     def component(name)
+      # The dry namespace separator is not intended to be user-configurable.
+      name = name.tr('/', '.')
```
waiting-for-dev, line 13: We're not actually changing the container separator. But I think that's good for a couple of reasons:

- We're not actually mapping one-to-one filesystem with identifiers. E:g., we're skipping the `SolidusAdmin` namespace, and a path `components/solidus_admin/foo.rb` is registered as `components.foo`. Using `components/foo` would be confusing.
- The container is also used to manually register other dependencies with no many-to-one mapping to a file, like different instances of `SolidusAdmin::MainNavItem`. In that case, using something like `solidus_admin/main_nav_item/first_item` would also be confusing.

However, I think it's a great idea to hide all of that from within the `component` helper, as you did. But the comment and the commit titles are confusing as it looks as we did the opposite.

waiting-for-dev (commented):
Thanks, @elia. I left some considerations. On top of that, the last commit breaks the loading of the page, probably a last-minute addition :slightly_smiling_face:

admin/app/helpers/solidus_admin/container_helper.rb
elia: > But the comment and the commit titles are confusing as it looks as we did the opposite.

Good call, I didn't recheck the title after discovering it was not configurable on DRY.rb, I'll update it 👍 

> We're not actually mapping one-to-one filesystem with identifiers. E:g., we're skipping the SolidusAdmin namespace, and a path components/solidus_admin/foo.rb is registered as components.foo. Using components/foo would be confusing.

To me rendering components is very similar to rendering partials, while the dots make me think of translations. It's clearly coming from what I'm used to in Rails, but I think that can be generalized to other Rails users. Even considering (comparatively rare) cases in which components are overridden making it look like a path is easily translatable to the file position in the file system.

> The container is also used to manually register other dependencies

With the code in this PR that is still the case (i.e. accessing using dots), but I think we would probably be better off with different spaces for data and components, be it a container, a configuration, or something else.

admin/app/components/solidus_admin/main_nav/component.rb
elia: I'll move the commit out of this PR as it probably needs it's own space 👍

admin/app/components/solidus_admin/main_nav/item/component.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# Menu item within a {MainNavComponent}
+class SolidusAdmin::MainNav::Item::Component < SolidusAdmin::BaseComponent
```
waiting-for-dev, line 4: I'm still trying to figure out the nesting components. It can lead to very lengthy namespaces. For instance, in #5154, we're nesting the `main_nav` component within a `sidebar` component. Should we then name this one `SolidusAdmin::Sidebar::MainNav::Item`? What if we add a main nav item icon component at some point? `SolidusAdmin::Sidebar::MainNav::Item::Icon`?
On top of that, what if we move the main navigation outside the sidebar and into the header at some point? Making no assumptions about the components' structure regarding their namespaces will help us refactor. Trying to find some rules might end up being a hassle to follow. In this case, it might seem clear that a main nav item belongs to a main nav, but it could be less evident in other cases.
Thoughts?

elia: I can report the experience and the steps we went through for the last project I used components in.

We started out having components "organized organically", i.e. with no specific rule except individual common sense. That was fine initially but as the number of components grew we had to figure out a better way.

Looking at the components we had it was pretty evident that we had sort of two sets of components:
- page specific components (think the product details for the PDP)
- ui elements (think of the classic button)

We came up with a soft rule for which every component that was used across multiple pages was probably a UI element and we started a `UI::` namespace for them. We also decided to keep all components specific to a page in a single namespace (e.g. `app/components/products` or `app/components/layout`). That brought back `app/components` to a manageable state.

It also made very important to be able to move components around swiftly as we realized that things are evolving and we didn't want to spend time fixing class nesting and constant references or renaming files.

---

With all of that as context I'll say that adopting this principles/rules the tree structure could probably look like this:

a base layout folder: `app/components/solidus_admin/layout` will contain a `nav/` subfolder inside which we can decide to keep things flatten if we want. But personally I don't see issues with deeper nestings, we're far from the limits of Ruby and constant lookups are cached by the VM anyway.

In the same way I think we need to start with less granular components and extract as we see a direct benefit instead of going with a very fine granularity upfront. 

Our need for basic components is a perfect benchmark of the level we need to reach as customizations of the admin will need to build similar interfaces. In other words I cannot say right now if we need a component for a table row or not, but will be almost self evident once we try to build two or three index tables.

kennyadsl: Thanks for sharing this @elia! I think it might work but I have one concern related to 

> and extract as we see a direct benefit instead of going with a very fine granularity upfront.

which might be harder than in a specific store, as we should handle backward compatibility. Of course, even if we go with the more granular approach, no one ensures us that that's the correct way of structuring components and that they will never change location. What I'd like to understand (even later) is how we want to approach having this flexibility of extracting components when we see they are used in multiple places taking into account backward compatibility, which should be a lean practice to work.

elia: @kennyadsl yes, in the end, when we officially release, we'll need the right amount of granularity to allow extendibility.

When I propose we start simple with bigger components I'm referring mostly to the grace time in which the admin is not released or is still experimental. Lowering the cost of renaming and extracting components (through things like 1-level nesting, the generator, etc.) I think we'll be able to achieve it during that time, taking informed (and more confident) decisions based on our direct experience.

waiting-for-dev: Thanks, @elia, for your detailed description of the process you followed.

I'm ok with following the `UI` vs. page namespaced convention.

> a base layout folder: app/components/solidus_admin/layout will contain a nav/ subfolder inside which we can decide to keep things flatten if we want. But personally I don't see issues with deeper nestings, we're far from the limits of Ruby and constant lookups are cached by the VM anyway.

For that, I'd go with a flattened structure. FWIW, I'm not concerned about Ruby's limits or performance. As you said, we'll discover the level of granularity along the way, and I'd like to avoid expensive refactors. As in the example I gave above, it's good if I can introduce the `Sidebar` namespace without needing to rename `MainNav` to `Sidebar::MainNav` and `MainNav::Item` to `Sidebar::MainNav::Item`. I also think it makes things more ergonomic. Are you ok with that?

admin/app/helpers/solidus_admin/container_helper.rb
waiting-for-dev: Hey, @elia, we still forgot to remove the comment here. dry-system's separator is indeed configurable, but we're not changing it there but in the component helper (which makes sense as I pointed out above).

admin/lib/solidus_admin/system/loaders/host_overridable_constant.rb
```diff
@@ -60,7 +60,7 @@ def override_constant(component, namespace)
             "#{application_name}::SolidusAdmin::#{component_segment(component, namespace).classify}".constantize
           end
 
-          # "components.foo.component" => "foo/component"
+          # "components/foo/component" => "foo/component"
```
waiting-for-dev, line 61: This is wrong, as we didn't change the namespace separator in the container, only in the helper.

admin/app/helpers/solidus_admin/container_helper.rb
waiting-for-dev: Oh, ok, I didn't know that. Anyway, I'd remove the comment because of the reasons I detailed above. Non-blocking.

rainerdema (approved):
I find the approach of using the slash and the concept of dividing components into UI vs page components to be sensible. 👏
Thank you all for the insights presented!

waiting-for-dev (commented):
@elia, trying it locally raises an error, as the `components.main_nav_item.component` key is not found in the container for this version of the code. We should either flat the item component or change the identifier to `components.main_nav.item.component`.

waiting-for-dev (approved):

PR #5135, opened 2023-06-08
[admin] Add a SolidusAdmin component generator

elia:
## Summary

- Components changes in https://github.com/solidusio/solidus/pull/5134 are compatible with the generator in this PR

Add a component generator specialized for SolidusAdmin that can be used either to generate new components within SolidusAdmin and custom components in the host application.

## Checklist

### DIFF
--- a/admin/README.md
+++ b/admin/README.md
@@ -3,3 +3,21 @@
 A Rails engine that provides an administrative interface to the Solidus ecommerce platform.
 
 - [Customizing tailwind](docs/customizing_tailwind.md)
+
+## Development
+
+### Adding components to Solidus Admin
+
+When using the component generator from within the admin folder it will generate the component in the library
+instead of the sandbox application.
+
+```bash
+cd admin
+# the `solidus_admin/` namespace is added by default
+bin/rails g solidus_admin:component foo
+      create  app/components/solidus_admin/foo/component.rb
+      create  app/components/solidus_admin/foo/component.html.erb
+      create  app/components/solidus_admin/foo/component.yml
+      create  app/components/solidus_admin/foo/component.js
+      create  spec/components/solidus_admin/foo/component_spec.rb
+```
--- a/admin/app/components/solidus_admin/base_component.rb
+++ b/admin/app/components/solidus_admin/base_component.rb
@@ -8,8 +8,8 @@ module SolidusAdmin
     include ViewComponent::InlineTemplate
     include SolidusAdmin::ContainerHelper
 
-    def identifier
-      @identifier ||= module_parent.to_s.underscore.dasherize.gsub(%r{/}, '--')
+    def stimulus_id
+      @stimulus_id ||= module_parent.to_s.underscore.dasherize.gsub(%r{/}, '--')
     end
   end
 end
--- a/admin/app/components/solidus_admin/base_component.rb
+++ b/admin/app/components/solidus_admin/base_component.rb
@@ -7,5 +7,9 @@ module SolidusAdmin
   class BaseComponent < ViewComponent::Base
     include ViewComponent::InlineTemplate
     include SolidusAdmin::ContainerHelper
+
+    def identifier
+      @identifier ||= module_parent.to_s.underscore.dasherize.gsub(%r{/}, '--')
+    end
   end
 end
--- a/admin/lib/generators/solidus_admin/component/USAGE
+++ b/admin/lib/generators/solidus_admin/component/USAGE
@@ -1,0 +1,13 @@
+Description:
+
+    Creates a SolidusAdmin component.
+
+Example:
+
+    $ bin/rails generate solidus_admin:component my_namespace/my_module_name my_attribute
+
+          create  app/components/my_namespace/my_module_name/component.rb
+          create  app/components/my_namespace/my_module_name/component.html.erb
+          create  app/components/my_namespace/my_module_name/component.yml
+          create  app/components/my_namespace/my_module_name/component.js
+          create  spec/components/my_namespace/my_module_name/component_spec.rb
--- a/admin/lib/generators/solidus_admin/component/component_generator.rb
+++ b/admin/lib/generators/solidus_admin/component/component_generator.rb
@@ -12,8 +12,13 @@ module SolidusAdmin
     class_option :spec, type: :boolean, default: true
 
     def create_component_files
+      template "component.html.erb", destination(".html.erb")
+      unless options["html"]
+        say_status :inline, destination(".html.erb"), :blue
+        @inline_html = File.read(destination(".html.erb"))
+        shell.mute { remove_file(destination(".html.erb")) }
+      end
       template "component.rb", destination(".rb")
-      template "component.html.erb", destination(".html.erb") if options["html"]
       template "component.yml", destination(".yml") if options["i18n"]
       template "component.js", destination(".js") if options["js"]
       template "component_spec.rb", destination("_spec.rb", root: "spec/components") if options["spec"]
@@ -71,6 +76,10 @@ module SolidusAdmin
         (stimulus_html if options["js"]),
         (i18n_html if options["i18n"]),
       ].compact.join("\n  ")
+    end
+
+    def inline_html(indent: '')
+      @inline_html.gsub!(/^/, indent).strip
     end
   end
 end
--- a/admin/lib/generators/solidus_admin/component/component_generator.rb
+++ b/admin/lib/generators/solidus_admin/component/component_generator.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+module SolidusAdmin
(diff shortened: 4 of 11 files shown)


admin/app/components/solidus_admin/base_component.rb
```diff
@@ -7,5 +7,9 @@ module SolidusAdmin
   class BaseComponent < ViewComponent::Base
     include ViewComponent::InlineTemplate
     include SolidusAdmin::ContainerHelper
+
+    def identifier
+      @identifier ||= module_parent.to_s.underscore.dasherize.gsub(%r{/}, '--')
```
waiting-for-dev, line 12: Two considerations:

- Should we remove the `solidus_admin` prefix?
- Following what you did in #5134, why not skip dasherizing and use `/` instead of `--` for the identifier?

The result of both is that the `SolidusAdmin::MainNav::Item::Component` component could be referenced with `main_nav/item`.

admin/lib/generators/solidus_admin/component/USAGE
```diff
@@ -0,0 +1,13 @@
+Description:
+
+    Creates a SolidusAdmin component.
+
+Example:
+
+    $ bin/rails generate solidus_admin:component my_namespace/my_module_name my_attribute
```
waiting-for-dev, line 7: What do you think if we use `Rails.application.module_parent` to automatically fetch `my_namespace`?

We should also add `solidus_admin` within (like in `my_namespace/solidus_admin/my_module_name`) so that the generated component is automatically registered by the container.

In summary, we could:

```
bin/rails generate solidus_admin:component my_module_name my_attribute
```

have to generate:

`solidus_admin/my_namespace/my_module_name/component.rb` and siblings.

What do you think?

waiting-for-dev (commented):
Thanks, @elia, that's going to be very helpful.

I left some considerations. The main problem with the current implementation is that the generated component doesn't follow the implicit container registration.

admin/lib/generators/solidus_admin/component/USAGE
elia: Maybe we can add an option for that, I'd like to still leave the door open for defining components in an arbitrary namespace (e.g. `my_namespace_admin`) as I had some bad experiences with prependers using the same schema and introducing the requirement of using leading colons every time we had to reference `::Spree` constants (Basically every reference to `Spree` was looked up as `MyNamespace::Spree` creating constant headaches).

I'm also not sure if we're able to discern within the generator when we're working within the engine and when it's the host app (in order to understand when we should avoid the namespace), but maybe there's a way.

@waiting-for-dev please let me know if this is something you think should be addressed now or we can evaluate later.

admin/app/components/solidus_admin/base_component.rb
elia: We need to check if using `/` is valid for stimulus controller names, in that case I think it would be much clearer, although would create a discrepancy when controller names are used in data-attributes:

```html
<div data-controller="foo/bar">
  <span data-foo-bar-target="baz"></span>
</div>
```

As I don't think we can do:

```html
<div data-controller="foo/bar">
  <span data-foo/bar-target="baz"></span>
</div>
```

waiting-for-dev: > As I don't think we can do:
> 
> ```
> <div data-controller="foo/bar">
>   <span data-foo/bar-target="baz"></span>
> </div>
> ```

That's a good point. Maybe it's better to be consistent here?

admin/README.md
```diff
@@ -2,3 +2,19 @@
...
+## Development
+
+### Adding components to Solidus Admin
+
+When using the component generator from within the admin folder it will generate the component in the library
+instead of the sandbox application.
+
+```bash
+cd admin
+bin/rails g solidus_admin:component foo
```
kennyadsl, line 14: After reading https://github.com/solidusio/solidus/pull/5135/files#r1223834576, I'm not sure if this does need to have the namespace as well? Shouldn't it be

```
bin/rails g solidus_admin:component solidus_admin/foo
```

?

admin/lib/generators/solidus_admin/component/templates/component.html.erb.tt
```diff
@@ -0,0 +1,4 @@
+<div class="<%%= identifier %>"<%= stimulus_attributes if options[:js] %>>
+  <p>Add <%= class_name %> HTML here</p>
```
kennyadsl, line 2: Maybe it is good to use the placeholder i18n key here, though?

admin/lib/generators/solidus_admin/component/component_generator.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+module SolidusAdmin
+  class ComponentGenerator < Rails::Generators::NamedBase
+    source_root File.expand_path('templates', __dir__)
+
+    argument :attributes, type: :array, default: [], banner: "attribute"
+
+    class_option :html, type: :boolean, default: true
```
kennyadsl, line 9: What's the use case to generate the template without html? If it's the inline render, shouldn't we add that block to the generated component?

admin/lib/generators/solidus_admin/component/component_generator.rb
```diff
@@ -0,0 +1,63 @@
...
+    class_option :i18n, type: :boolean, default: true
```
kennyadsl, line 10: I'd like to enforce the good practice of always using i18n, especially if it's so easy to get it modular for the component. Do you think it's common for people to not use i18n, even if it's so straightforward?

admin/lib/generators/solidus_admin/component/component_generator.rb
```diff
@@ -0,0 +1,63 @@
...
+    class_option :js, type: :boolean, default: true
+    class_option :spec, type: :boolean, default: true
```
kennyadsl, line 12: If I recall correctly, we are still undecided if we want to keep the component's tests in the sidecar directory. Do we want to start the discussion here in this context? Alternatively, we can extract this part in another PR/issue to facilitate merging this one.

kennyadsl (commented):
Excellent work @elia! I left some points, let me know your thoughts!

admin/README.md
elia: Rails has a built-in mechanism (I think coming from the name-generator) that will add the namespace for engines automatically when called with an ENGINE_ROOT set. 

So this is correct, but adding it explicitly would work as well as there's also a check that won't add it if it's already there.

elia (commented):
@kennyadsl thanks for the review, I'll update and ask for another one once ready

admin/lib/generators/solidus_admin/component/component_generator.rb
kennyadsl: Ah right, I think it's hard to anticipate that you won't need any text before building the component but could happen and as long as the default is `true`, I think it's ok to keep the option.

kennyadsl (approved):
Left a non-blocking one. Thanks!

admin/app/components/solidus_admin/base_component.rb
elia: > Maybe it's better to be consistent here?

I'm not sure I understand what you're suggesting, can you expand on it?

waiting-for-dev: I was reconsidering if it's worth changing how identifiers look like if we can't use the same format everywhere (the data attribute you suggested").

elia: Oh I see, now it's clear 👍 

I think we should not conflate the registry keys and the internal component identifier, which should be independent and for stimulus is tied to the path of the javascript file. 

Maybe we can rename the method and call it `js_id` or `stimulus_id` (or `js_identifier` or `javascript_identifier` etc.), but since it's likely to be used often I would take shortness in consideration.

Thoughts on this?

[Norm Macdonald — ID (A Strange Abbreviation) [Izp52QUeSzQ].webm](https://github.com/solidusio/solidus/assets/1051/e97737e5-8397-4501-a5be-71b338e0d45d)

waiting-for-dev: > I think we should not conflate the registry keys and the internal component identifier, which should be independent and for stimulus is tied to the path of the javascript file.

I completely agree. Same for the `components` helper.

> Maybe we can rename the method and call it js_id or stimulus_id (or js_identifier or javascript_identifier etc.), but since it's likely to be used often I would take shortness in consideration.

Maybe `stimulus_id` makes everything clearer? No strong opinion, though.

elia: 👍 Renamed to `stimulus_id`

waiting-for-dev (approved):

rainerdema (approved):
👍

PR #5136, opened 2023-06-09
[admin] Delay use of auto injection for components

elia:
## Summary

We might want to add back some auto-injection mechanism in the future but for now vanilla ruby is easier to understand and work with.

Is worth noting that is among ViewComponent conventions [to have a base class (for host applications is expected to be called `ApplicationComponent`)](https://viewcomponent.org/guide/getting-started.html#conventions) and [to let each component handle its own `initialize` method](https://viewcomponent.org/guide/collections.html#collection-counter) _(thanks @kennyadsl for asking about this)_.

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/main_nav_item/component.rb
+++ b/admin/app/components/solidus_admin/main_nav_item/component.rb
@@ -5,15 +5,13 @@ module SolidusAdmin
   class MainNavItem::Component < BaseComponent
     with_collection_parameter :item
 
-    attr_reader :item
-
     def initialize(item:)
       @item = item
     end
 
     erb_template <<~ERB
       <a href="#">
-        <%= item.title %>
+        <%= @item.title %>
       </a>
     ERB
   end
--- a/admin/app/components/solidus_admin/main_nav/component.rb
+++ b/admin/app/components/solidus_admin/main_nav/component.rb
@@ -3,25 +3,19 @@
 module SolidusAdmin
   # Renders the main navigation of Solidus Admin.
   class MainNav::Component < BaseComponent
-    include Import[
-      main_nav_item_component: "components.main_nav_item.component",
-      items: "main_nav_items"
-    ]
+    def initialize(items: container["main_nav_items"], item_component: component("main_nav_item"))
+      @items = items
+      @item_component = item_component
+    end
 
     erb_template <<~ERB
       <nav>
-        <%=
-          render main_nav_item_component.with_collection(
-            sorted_items
-          )
-        %>
+        <%= render @item_component.with_collection(items) %>
       </nav>
     ERB
 
-    private
-
-    def sorted_items
-      items.sort_by(&:position)
+    def items
+      @items.sort_by(&:position)
     end
   end
 end


rainerdema (approved):
I agree with this proposal 👍

waiting-for-dev:
> We might want to add back some auto-injection mechanism in the future but for now vanilla ruby is easier to understand and work with.

Sorry @elia, what is what you find particularly difficult to work with? As per the understanding part, could it be something easy to fix with documentation?

I'm not entirely against it. I didn't add that in the first version of the spike. However, I'd like to document component extensibility, including auto_injection. It could be too boilerplate without it, and I'd like to explain better how it works (also to clear out some stuff on my end). Can we hold it until the Readme is ready? I can work on it on Monday.

elia:
> what is what you find particularly difficult to work with?

I had to dig into DRY in order to understand what was going on, surely better documentation could have helped, vanilla ruby on the other hand won't require documentation at all. There are surely good use cases in which DRY.rb would reduce boiler plate code, but in this case I've found both the number of lines and the complexity reduced.

We can surely hold off as much as we want 👍

tvdeyen (approved):
Like this better as well

To quote https://peps.python.org/pep-0020/#the-zen-of-python 

> Explicit is better than implicit.

admin/app/components/solidus_admin/main_nav_item/component.rb
```diff
@@ -5,16 +5,13 @@ module SolidusAdmin
   class MainNavItem::Component < BaseComponent
     with_collection_parameter :item
 
-    attr_reader :item
```
kennyadsl, line 8: A random note about using `ivar` or `attr_reader`: `ivar` is harder to deprecate against the `attr_reader`. For the former, we will need to use [DeprecatedInstanceVariableProxy](https://api.rubyonrails.org/classes/ActiveSupport/Deprecation/DeprecatedInstanceVariableProxy.html) which is complex and doesn't support deprecating without a replacement if I recall correctly. The `attr_reader` is easier to deprecate, because it's a plain method.

kennyadsl: That's a good point, but if we allow the specific granularity in customizing only the component template without touching the rest of the component, and the template relies on that instance variable, it's already a public API that people might be using in a legit way.

waiting-for-dev (approved):
Given the quorum of this suggestion, let's move forward with it. In my experience, using auto_inject greatly reduces boilerplate code, leveraging Ruby's flexibility to make transparent dependency injection a pleasure to work with (as opposed to Java's hell). We can always add it later if we feel it's worth it.

By the way, I wrote the documentation following the change in this PR. Please, see https://github.com/solidusio/solidus/pull/5144.

admin/app/components/solidus_admin/main_nav_item/component.rb
elia: @kennyadsl yes, that's something we should seriously consider before committing to allowing replacing just the template

kennyadsl (approved):

PR #5139, opened 2023-06-09
Configurable cancelation adjuster class

mamhoff:
## Summary

This adds a class that handles recalculating adjustments originating from a Spree::UnitCancel. The goal is to eliminate uses of `Spree::Adjustment#recalculate`, similar to #4460 .

## Checklist

- 📖 ~I have updated the README to account for my changes.~
- 🛣️ ~I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).~

### DIFF
--- a/core/lib/spree/app_configuration.rb
+++ b/core/lib/spree/app_configuration.rb
@@ -334,6 +334,14 @@ module Spree
 
     class_name_attribute :shipping_rate_selector_class, default: 'Spree::Stock::ShippingRateSelector'
 
+    # Allows implementing custom cancellations recalculation.
+    # @!attribute [rw] cancellations_recalculator_class
+    # @ see Spree::UnitCancel::OrderCancellationsRecalculator
+    # @return [Class] a class that conforms to the API of the
+    #   standard cancellations recalculator class,
+    #   `Spree::UnitCancel::OrderCancellationsRecalculator`
+    class_name_attribute :cancellations_recalculator_class, default: 'Spree::UnitCancel::OrderCancellationsRecalculator'
+
     # Allows providing your own class for calculating taxes on a shipping rate.
     #
     # @!attribute [rw] shipping_rate_tax_calculator_class
--- a/core/spec/models/spree/unit_cancel/order_cancellations_recalculator_spec.rb
+++ b/core/spec/models/spree/unit_cancel/order_cancellations_recalculator_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Spree::UnitCancel::OrderCancellationsRecalculator do
+  let(:inventory_unit) { create(:inventory_unit, line_item: line_item) }
+  let(:line_item) { create(:line_item) }
+  let(:unit_cancel) { Spree::UnitCancel.create!(inventory_unit: inventory_unit, reason: Spree::UnitCancel::SHORT_SHIP) }
+  let(:cancellation) { unit_cancel.adjust! }
+
+  before do
+    cancellation.update_columns(amount: 10, finalized: false)
+    cancellation.reload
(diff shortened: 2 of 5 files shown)


core/spec/models/spree/unit_cancel/order_cancellations_recalculator_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  before do
+    cancellation.update_columns(amount: 10, finalized: false)
+    cancellation.reload
+    line_item.reload
+  end
+
+  subject { described_class.new(line_item.order).call }
+
+  it "recalculates the cancellation adjustment to the correct amount" do
+    expect { subject }.to change { cancellation.reload.amount }.from(10).to(-20)
```
mamhoff, line 20: Weird, actually. There's only one item to cancel, and only one canceled, yet the amount changes to -20? Turns out Unit Cancel adjustments are not even meant to be recalculated; they're always "finalized": https://github.com/solidusio/solidus/blob/main/core/app/models/spree/unit_cancel.rb#L27. Also, recalculating them does the wrong thing, as unit cancel adjustments do not take themselves into account: https://github.com/solidusio/solidus/blob/main/core/app/models/spree/unit_cancel.rb#L47

Should we make the cancellations recalculator a no-op by default then?

mamhoff (commented):
Should we make the recalculator a no-op by default? It usually doesn't do anything because Unit Cancels are always finalized, and finalized adjustments do not get recalculated.

core/spec/models/spree/unit_cancel/order_cancellations_recalculator_spec.rb
elia: > Should we make the cancellations recalculator a no-op by default then?

So the code accounting for when the adjustment is not finalized is just to keep the current behavior, which seems to never be called in real stores, and it's only a contrived scenario in the specs?

I was wondering that also in relation to the optimizations used to avoid the extra queries, are those done off the spec only too?

If the answer is yes to both I think a preference+deprecation cycle in which we raise if the cancelation adjustment is not finalized is in good order ✂️ ✂️ ✂️

core/lib/spree/app_configuration.rb
```diff
@@ -334,6 +334,14 @@ def default_pricing_options
 
     class_name_attribute :shipping_rate_selector_class, default: 'Spree::Stock::ShippingRateSelector'
 
+    # Allows implementing custom cancellations recalculation.
+    # @!attribute [rw] cancellations_recalculator_class
+    # @ see Spree::UnitCancel::OrderCancellationsRecalculator
+    # @return [Class] a class that conforms to the API of the
+    #   standard cancellations recalculator class,
+    #   `Spree::UnitCancel::OrderCancellationsRecalculator`
+    class_name_attribute :cancellations_recalculator_class, default: 'Spree::UnitCancel::OrderCancellationsRecalculator'
```
tvdeyen, line 343: What about making this obvious?

```suggestion
    class_name_attribute :cancellations_recalculator_class, default: 'Spree::UnitCancel::WrongOrderCancellationsRecalculator'
```

tvdeyen (commented):
IMO There is no benefit in supporting buggy and dead code for a long deprecation cycle. Stores that reach this code, most likely have the bug fixed, will have it overwritten anyway.

What about keeping the legacy calculator and introduce a no-op (as proposed from Martin) and keeping the old code for backwards compatibility but make the no-op class the default in the next version?

kennyadsl:
I'm generally in favor of just deprecating the recalculation of cancellations without a replacement. As far as I understand, it's always already a no-op, because they are always finalized. 

But keeping the current core a part, do we know any other scenario when it might make sense to recalculate a cancellation?

mamhoff:
Closing in favor of #5633

PR #5142, opened 2023-06-12
Don't require calling super for components

elia:
## Summary

It's not a standard practice for view components that are meant to be very minimal and almost POROs.

Move the configuration _excludes_ from the rubocop todo to the main config as they seem legitimate exceptions.

## Checklist

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -334,3 +334,9 @@ Naming/MemoizedInstanceVariableName:
 
 Lint/SuppressedException:
   Enabled: false
+
+Lint/MissingSuper:
+  Exclude:
+    - '*/app/components/**/*' # components need pristine initializer methods
+    - 'core/lib/spree/deprecation.rb' # this is a known class that doesn't require super
+    - 'core/lib/spree/preferences/configuration.rb' # this class has no superclass defining `self.inherited`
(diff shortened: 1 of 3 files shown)


kennyadsl (approved):
Considering that each component could (and should) have totally different initialization arguments, I think this is a good call, and I think it's ok to consider this an exception in the Rubocop config. Slightly concerned that we are losing some point of contact with the base class in case for example we want to deprecate something in all components initialization using inheritance, but I'm sure we'll find a way even without calling super in that case.

elia:
> in case for example we want to deprecate something in all components initialization using inheritance, but I'm sure we'll find a way even without calling super in that case.

Yes, we'll surely find a way if that problem arises, which is hardly the case given there's not much you want to inherit from other components except from occasional template-helpers that can be either live in the base component or moved to modules.

.rubocop.yml
```diff
@@ -334,3 +334,9 @@ Naming/MemoizedInstanceVariableName:
 
 Lint/SuppressedException:
   Enabled: false
+
+Lint/MissingSuper:
+  Exclude:
+    - '*/app/components/**/*' # components need pristine initializer methods
+    - 'core/lib/spree/deprecation.rb' # this is a known class that doesn't require super
+    - 'core/lib/spree/preferences/configuration.rb' # this class has no superclass defining `self.inherited`
```
waiting-for-dev, line 342: Should we add those to the `main` branch?

waiting-for-dev (approved):
Makes sense!

PR #5144, opened 2023-06-12
Document admin's view components extensibility

waiting-for-dev:
## Summary

This document describes the working implementation of the admin's view components extensibility. See [the rendered content by GitHub](https://github.com/nebulab/solidus/blob/waiting-for-dev/document_admin_extensibility/admin/docs/customizing_view_components.md) for better readability.

It considers the changes that will be introduced in:

- https://github.com/solidusio/solidus/pull/5136
- https://github.com/solidusio/solidus/pull/5134

## Checklist

### DIFF
--- a/admin/docs/customizing_view_components.md
+++ b/admin/docs/customizing_view_components.md
@@ -1,0 +1,189 @@
+# Customizing view components
+
+Solidus Admin uses [view components](https://viewcomponent.org/) to render the views. Components are
+a pattern for breaking up the view layer into small, reusable pieces, easy to
+reason about and test.
+
+All the components Solidus Admin uses are located in the [`app/components`](../app/components) folder of the
+`solidus_admin` gem. As you can see, they are organized in a particular folder structure:
+
+- All of them are under the `SolidusAdmin` namespace.
+- They are grouped in sidecar directories, where the main component file and
+  all its related files (assets, i18n files, etc.) live together.
+
+For instance, the component for the main navigation is located in
+[`app/components/solidus_admin/main_nav/component.rb`](../app/components/solidus_admin/main_nav/component.rb).
+
+Solidus Admin components are designed to be easily customizable by the host
+application. Because of that, if you look at how they are designed, you'll find
+a series of patterns are followed consistently:
+
+- Components are always resolved from a global registry in
+  `SolidusAdmin::Container` instead of being referenced by their constant. For
+  example, we call `component('main_nav')` instead of referencing
+  `SolidusAdmin::MainNav::Component` directly. As you'll see later, this makes
+  it easy to replace or tweak a component.
+- When a component needs to render another component, it gets it as an argument
+  for its `#initialize` method and assigns it to an instance variable ending in
+  `_component`. When rendering from the template, it calls the `render` method
+  on the instance variable instead of resolving the component in-line. Because
+  of that, switching a nested component only requires modifying its
+  initializer.
+- Those components given on initialization are always defaulted to its expected
+  value resolved from the container. That means that callers don't need to
+  provide them explicitly, and that they can be replaced by custom components.
+- In any case, Solidus Admin components initializers only take keyword
+  arguments.
+
+A picture is worth a thousand words, so let's depict how this works with an
+example:
+
+```ruby
+# app/components/solidus_admin/foo/component.rb
+class SolidusAdmin::Foo::Component < SolidusAdmin::BaseComponent
+  def initialize(bar_component: component('bar'))
+    @bar_component = bar_component
+  end
+  
+  erb_template <<~ERB
+    <div>
+      <%= render @bar_component.new %>
+    </div>
+  ERB
+end
+# render component('foo').new
+```
+
+## Customizing components
+
+Some of the customizations detailed below require you to match Solidus Admin's
+component paths in your application. For instance, if we talk about the
+component in `solidus_admin` gem's
+`app/components/solidus_admin/main_nav/component.rb`, the matching path in your
+application would be
+`app/components/my_application/solidus_admin/main_nav/component.rb`, where
+`my_application` is the underscored name of your application (you can get it by
+running `Rails.application.class.module_parent_name`).
+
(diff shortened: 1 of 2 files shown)


admin/docs/customizing_view_components.md
```diff
@@ -0,0 +1,190 @@
...
+  it easy to replace or tweak a component.
+- When a component needs to render another component, it gets it as an argument
+  for its `#initialize` method and assigns it to an instance variable ending in
+  `_component`. When rendering from the template, it calls the `render` method
+  on the instance variable instead of resolving the component in-line. Because
+  of that, switching a nested component only requires modifying its
+  initializer.
+- Those components given on initialization are always defaulted to its expected
+  value resolved from the container. That means that callers don't need to
+  provide them explicitly, and that they can be replaced by custom components.
```
elia, line 34: I think injecting the component through the initializer is redundant with the registry, if we need to replace it we can use the registry itself.

For testing is easy to play with the registry already:

```rb
component = described_class.new
allow(component).to receive(:components).with(:bar).and_return(double('bar', render_in: '<span></span>')
…
```

There maybe are exceptions to this rule (e.g. for components that act as wrappers) but in those cases you usually end up passing an instance rather than the class:

```erb
<%= SolidusAdmin::UI::Modal::Component.new(
  content: SolidusAdmin::Products::Component.new(product: @product)
) %>
```

or even

```erb
<%= SolidusAdmin::UI::Modal::Component.new do %>
  <%= SolidusAdmin::Products::Component.new(product: @product) %>
<% end %>
```

---

Thoughts?

admin/docs/customizing_view_components.md
```diff
@@ -0,0 +1,190 @@
...
+argument from the top-level nav component:
+
+```ruby
+# app/components/my_application/solidus_admin/main_nav/component.rb
+class MyApplication::SolidusAdmin::MainNav::Component < ::SolidusAdmin::MainNav::Component
+  def initialize(main_nav_item_component: components('main_nav_item_foo'), **kwargs)
+    super
+  end
+end
+```
```
elia, line 190: I think this is a worthy feature to have, but I think we can achieve it tapping into the `components` helper, for example we might have an optional separate registry for each component and fallback to the global one.

What follow is pseudo-code:

```rb
# helper.rb
def components(name)
  container[name]
end

def container
  Container.for(self.class) || Container.global
end
```

```rb
# initializer in the host app
SolidusAdmin.configure do |config|
  config.component_registry_for('foo/bar')['baz'] = 'MyAlternativeBaz::Component'
end
```

elia (commented):
@waiting-for-dev thanks for tackling this, it's an awesome starting point for documentation wherever we'll end up 🙌 

I left one consideration about injecting component classes into initializers vs. using the registry.

admin/docs/customizing_view_components.md
waiting-for-dev: Hey @elia :wave: I don't think they're redundant. Besides the stub vs. mock stuff, it makes it impossible to inject a different dependency for a single instance. We'll reuse some components in more than a single place. If someone wants only to replace a dependency for one of them, there would be no way if everything is resolved straightaway from the container. In general, not having dependency injection generates a more coupled code. That has more or less effective alternatives as we're used in common Rails applications. However, that's not our case, and if we want to provide a system that can be extended, coupling is our worst enemy!

admin/docs/customizing_view_components.md
elia: > We can try something and see how it looks!

👍 I can give it a go and se if things come out naturally 😌 or if it's a monster 👹

admin/docs/customizing_view_components.md
elia: > If someone wants only to replace a dependency for one of them, there would be no way if everything is resolved straightaway from the container.

I'll do an experiment with having `components` results be overridable by subclasses of the BaseComponent as mentioned in the other comment. Looks feasible to me but I might be proved wrong by actually attempting it 👍

elia (approved):

admin/docs/customizing_view_components.md
waiting-for-dev: Let me know if you need anything. However, consider that I'm not talking about subclasses but instances. Even if we had a per-class registry, multiple instances of the same class would use the same dependency. Injecting it on initialization is how we can change that. There's also value in using well-known architectural patterns instead of building our utilitarian helpers. The former makes it easier to reason about once you know the pattern, and they are battled tested when it comes to building loosely decoupled systems. I'm sure, at this point, we can't envision all the needs we'll face.

admin/docs/customizing_view_components.md
```diff
@@ -0,0 +1,190 @@
...
+For instance, the component for the main navigation is located in
+[`app/components/solidus_admin/main_nav/component.rb`](../app/components/solidus_admin/main_nav/component.rb).
+
+Solidus Admin components are designed to be easily customizable by the host
+application. Because of that, if you look at how they are designed, you'll find
+a series of patterns are followed consistently:
+
+- Components are always resolved from a global registry in
+  `SolidusAdmin::Container` instead of being referenced by their constant. For
+  example, we call `components('main_nav')` instead of referencing
```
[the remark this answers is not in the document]
waiting-for-dev: Oh, thanks for catching it, fixed!

admin/docs/customizing_view_components.md
```diff
@@ -0,0 +1,190 @@
...
+In some situations, you might only need to add some markup before or after a
+component. You can easily do that by rendering the Solidus Admin component and
+adding your markup before or after it.
+
+```erb
+<%# app/components/my_application/solidus_admin/main_nav/template.html.erb %>
+<h1>MY STORE ADMINISTRATION</h1>
+<%= render SolidusAdmin::MainNav::Component.new %>
+</nav>
+```
```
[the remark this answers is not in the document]
waiting-for-dev: Removed!

rainerdema (approved):
💯

PR #5152, opened 2023-06-13
Allow lambda in menu item :match_path option and URL

mamhoff:
## Summary

This allows setting a path for a menu item via a lambda executed at runtime. If engines other than spree would like their route in the main menu, this is how they can do dynamically without having to hard-code the path. 

Similarly, we allow the `match_path` option for Menu Items to be a callable taking a request. 

## Checklist

### DIFF
--- a/backend/lib/spree/backend_configuration/menu_item.rb
+++ b/backend/lib/spree/backend_configuration/menu_item.rb
@@ -4,7 +4,7 @@ module Spree
   class BackendConfiguration < Preferences::Configuration
     # An item which should be drawn in the admin menu
     class MenuItem
-      attr_reader :icon, :label, :partial, :condition, :sections, :url, :match_path
+      attr_reader :icon, :label, :partial, :condition, :sections, :match_path
 
       attr_accessor :position
 
@@ -21,8 +21,9 @@ module Spree
       # @param url [String] A url where this link should send the user to
       # @param position [Integer] The position in which the menu item should render
       #   nil will cause the item to render last
-      # @param match_path [String, Regexp] (nil) If the {url} to determine the active tab is ambigous
-      #   you can pass a String or Regexp to identify this menu item
+      # @param match_path [String, Regexp, callable] (nil) If the {url} to determine the active tab is ambigous
+      #   you can pass a String, Regexp or callable to identify this menu item. The callable
+      #   accepts a request object and returns a Boolean value.
       def initialize(
         sections,
         icon,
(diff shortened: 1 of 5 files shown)


backend/lib/spree/backend_configuration/menu_item.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'spree/preferences/configuration'
+require 'spree/backend_configuration/menu_item'
```
tvdeyen, line 4: These requires should not be necessary

tvdeyen (approved):
Nice feature. This helps engines like [alchemy-solidus](https://github.com/AlchemyCMS/alchemy-solidus) to hook into the main navigation much nicer 👍🏻 

_Rubocop still needs some attention here._

mamhoff:
Rubocop issues should be addressed.

backend/lib/spree/backend_configuration/menu_item.rb
elia: `require 'spree/backend_configuration/menu_item'` isn't this a circular require?

Ruby handles it fine, but we should remove just that line, the other one is a direct dependency and it's ok to keep it there 👍

elia (commented):
I like this change a lot, I think this is the right direction 🙌 

I'm also wondering if we can also integrating some of the logic in `backend/app/views/spree/admin/shared/_tabs.html.erb`.

Maybe we want to evaluate the url in the context of the view like we do for the `condition`: `instance_exec(&menu_item.condition)`.

And we might also think of ways to move this ternary from the view to the `#url` helper `menu_item.url.is_a?(Symbol) ? spree.public_send(menu_item.url) : menu_item.url`.

@mamhoff thoughts on this kind of changes?

backend/lib/spree/backend_configuration/menu_item.rb
tvdeyen: > isn't this a circular require?

Yep, thats what I thought as well

mamhoff: Both are unnecessary as the file is required from a file that has both. Removed, thank you Thomas, and sorry for the misunderstanding

tvdeyen: No worries

mamhoff:
While I think this API reorganization would be a nice idea, I don't think the level of effort and change involved is worth the pain at this point...

jarednorman (approved):

elia (approved):
Fair enough, that's admittedly a lot 👍👍👍

PR #5153, opened 2023-06-14
[admin] Add importmap + hotwire support to SolidusAdmin

elia:
## Summary

This PR sets up a separate importmap instance for SolidusAdmin, any configuration that could be shared was not overridden.

There is a PR open for making the official helper flexible enough to support this use-case, but until it's merged we need to keep our version: https://github.com/rails/importmap-rails/pull/187.

An example `hello_controller.js` is added as a demonstration, since the whole layout will be revised we can get rid of it once we start actually building components.

## Checklist


admin/app/assets/config/solidus_admin_manifest.js
```diff
@@ -0,0 +1,3 @@
+//= link_tree ../images
+//= link_tree ../stylesheets .css
+//= link_tree ../../javascript .js
```
waiting-for-dev, line 3: Shouldn't it be `../` instead of `../../`? Also, Rails uses `javascripts` (pluralized) as the default for new applications. Should we use the same convention?

admin/app/javascript/solidus_admin/application.js
```diff
@@ -0,0 +1 @@
+console.log("Hello from application.js")
```
waiting-for-dev, line 3: Is this intentional to make sure it works, or can we remove it? :thinking:

admin/lib/tasks/importmap.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+namespace :solidus_admin do
+  namespace :importmap do
+    desc "Build Solidus Admin's Tailwind's css"
```
waiting-for-dev, line 5: Wrong description :slightly_smiling_face:

waiting-for-dev (commented):
Excellent work! :heart:

admin/lib/tasks/importmap.rake
elia: LOL, now you know from where I copied it! 😜

admin/app/javascript/solidus_admin/application.js
elia: Yes, this is intentional, once we start adding controllers in the form of `component.js` files I think we can cleanup, and maybe tweak a few more things.

admin/app/assets/config/solidus_admin_manifest.js
elia: This is the default for new apps when you use importmaps (which is also default), I think the rationale is that app/assets is the realm of sprockets/propshaft and app/javascript is where you have the source files that will be managed by your JS bundler (in this case `importmap-rails`).

```
⤑ rails new foobar                                                          /tmp
…[CUT]
⤑ cd foobar                                                                                                                                                                                                /tmp
foobar:master ⤑ tree app                                                                                                                                                                              /t/foobar
app
├── assets
│   ├── config
│   │   └── manifest.js
│   ├── images
│   └── stylesheets
│       └── application.css
├── channels
│   └── application_cable
│       ├── channel.rb
│       └── connection.rb
├── controllers
│   ├── application_controller.rb
│   └── concerns
├── helpers
│   └── application_helper.rb
├── javascript
│   ├── application.js
│   └── controllers
│       ├── application.js
│       ├── hello_controller.js
│       └── index.js
├── jobs
│   └── application_job.rb
├── mailers
│   └── application_mailer.rb
├── models
│   ├── application_record.rb
│   └── concerns
└── views
    └── layouts
        ├── application.html.erb
        ├── mailer.html.erb
        └── mailer.text.erb

17 directories, 16 files
```

waiting-for-dev: Oh, cool :+1:

waiting-for-dev (approved):
:+1:

admin/app/helpers/solidus_admin/importmap_helper.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module SolidusAdmin::ImportmapHelper
+  def javascript_solidus_admin_importmap_tags(entry_point = "solidus_admin/application", shim: false, importmap: SolidusAdmin.importmap)
```
waiting-for-dev, line 4: Sorry, @elia, any reason why we defaulted `shim` to `false` in our version? At first, glance, adding the `shim` looks nicer for greater compatibility.

elia: Yes of course! It was set to false for two reasons:
1. we are in a controlled environment in which is reasonable to expect evergreen browsers that will support import maps out of the box https://caniuse.com/import-maps
2. this allowed to remove an extra import in the component stimulus loader code that is still visible in the original file https://github.com/hotwired/stimulus-rails/blob/498902a3687d32bd6589f7d3c7e7d72eabd822c0/app/assets/javascripts/stimulus-loading.js#L1-L2

admin/lib/solidus_admin/configuration.rb
```diff
@@ -38,6 +38,15 @@ class Configuration < Spree::Preferences::Configuration
     #
     # Recompile with `bin/rails solidus_admin:tailwindcss:build` after changing this list.
     preference :tailwind_stylesheets, :array, default: []
+
+    preference :importmap_cache_sweepers, :array, default: [
```
waiting-for-dev, line 42: It'd be nice if we document this preference and also add it to the [generated initializer](https://github.com/solidusio/solidus/blob/nebulab/admin/admin/lib/generators/solidus_admin/install/templates/config/initializers/solidus_admin.rb).

Do you think it would make sense to add `Rails.root.join("app", "javascript", "solidus_admin")` & `Rails.root.join("app", "assets", "javascripts", "solidus_admin")` to the defaults?

admin/lib/solidus_admin/configuration.rb
```diff
@@ -38,6 +38,15 @@ class Configuration < Spree::Preferences::Configuration
...
+      SolidusAdmin::Engine.root.join("app", "assets", "javascripts"),
+      SolidusAdmin::Engine.root.join("app", "javascript"),
+    ]
+
+    preference :importmap_paths, :array, default: [
```
waiting-for-dev, line 47: Same as https://github.com/solidusio/solidus/pull/5153/commits/38b7ef50841573d26ce9881f0ddee83e8859cf7e#r1234972715 :slightly_smiling_face:

PR #5158, opened 2023-06-16
Fix solidus_core making use of responders API

waiting-for-dev:
## Summary

We removed the responders dependency from solidus_core in https://github.com/solidusio/solidus/pull/2090. However, we forgot to remove the use of the responders API from solidus_core. We weren't catching this because all consumers of solidus_core's `Spree::BaseController` had the responders gen as a dependency (see https://github.com/solidusio/solidus/pull/3336)

## Checklist

### DIFF
--- a/backend/app/controllers/spree/admin/base_controller.rb
+++ b/backend/app/controllers/spree/admin/base_controller.rb
@@ -8,6 +8,8 @@ module Spree
 
       before_action :authorize_admin
 
+      respond_to :html
+
       private
 
       # Overrides ControllerHelpers::Common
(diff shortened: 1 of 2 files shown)


elia (approved):
Love the change, just wondering if this is a breaking change, since we didn't have a deprecation

waiting-for-dev:
> Love the change, just wondering if this is a breaking change, since we didn't have a deprecation

I think it's a breaking change if somebody inherited from `Spree::BaseController` and relies on `:html` having been added as a format.

However, it's also a bug fix, as somebody inheriting from `Spree::BaseController` and lacking `responders` as a dependency will see an error.

I'm not sure there's an easy way to deprecate that:

1. Somebody is inheriting from `Spree::BaseController`
2. They are making use of `responders`
3. They are using the `:html` format
4. They are not adding themselves the `:html` format

In this case, I'd treat it as a bug fix and call it a day :slightly_smiling_face:

rainerdema (approved):
Thanks @waiting-for-dev! 👏

kennyadsl (approved):
I agree we should treat this as a bug fix, mainly because it wasn't expected to be there and using responder inheriting from a core controller should not be allowed.

jarednorman (approved):

elia (approved):

PR #5167, opened 2023-06-19
[Admin] Add the `UI` namespace

### LINKED ISSUE #5151: [Admin] Avoid polluting `SolidusAdmin` namespace with components
As we'll have many components, it's good if they are not defined without any apparent pattern within modules under the `SolidusAdmin` namespace. E.g., we could have a `SolidusAdmin::OrdersManagament::Component` component, but there are chances that a module in any other place could also want a `SolidusAdmin::OrdersManagement` name. We have talked about defining modules within `SolidusAdmin::UI`, or we could also establish a `ModuleNameComponent` convention.

elia:
## Summary

Setup `UI` as an acronym and fix the component generator to properly use it.

- Fixes #5151 

## Checklist

### DIFF
--- a/admin/lib/generators/solidus_admin/component/component_generator.rb
+++ b/admin/lib/generators/solidus_admin/component/component_generator.rb
@@ -11,6 +11,13 @@ module SolidusAdmin
     class_option :js, type: :boolean, default: true
     class_option :spec, type: :boolean, default: true
 
+    def setup_inflections
+      # This is needed because the generator won't run the initialization process,
+      # in order to ensure that UI is not rendered as Ui we need to setup inflections
+      # manually.
+      SolidusAdmin::Engine.initializers.find { _1.name =~ /inflections/ }.run
+    end
+
     def create_component_files
       template "component.html.erb", destination(".html.erb")
       unless options["html"]
--- a/admin/lib/solidus_admin/engine.rb
+++ b/admin/lib/solidus_admin/engine.rb
@@ -12,6 +12,11 @@ module SolidusAdmin
       require "solidus_admin/configuration"
     end
 
+    initializer "solidus_admin.inflections" do
+      # Support for UI as an acronym
+      ActiveSupport::Inflector.inflections { |inflect| inflect.acronym 'UI' }
+    end
+
     initializer "solidus_admin.importmap" do
       SolidusAdmin::Config.importmap_paths.each { |path| SolidusAdmin.importmap.draw(path) }
     end
--- a/admin/lib/solidus_admin/container.rb
+++ b/admin/lib/solidus_admin/container.rb
@@ -15,6 +15,7 @@ module SolidusAdmin
   # @api private
   class Container < Dry::System::Container
     configure do |config|
+      config.inflector = ActiveSupport::Inflector
       config.root = Pathname(__FILE__).dirname.join("../..").realpath
       config.component_dirs.add("app/components/solidus_admin") do |dir|
         dir.loader = System::Loaders::HostOverridableConstant.method(:call).curry["components"]


admin/lib/generators/solidus_admin/component/component_generator.rb
```diff
@@ -11,6 +11,11 @@ class ComponentGenerator < Rails::Generators::NamedBase
     class_option :js, type: :boolean, default: true
     class_option :spec, type: :boolean, default: true
 
+    def setup_inflections
+      # This is needed to ensure that UI is not rendered as Ui
+      SolidusAdmin::Engine.initializers.find { _1.name =~ /inflections/ }.run
```
waiting-for-dev, line 16: Can you help me understand why we need this one? Where is it consumed? :thinking:

waiting-for-dev: I see, thanks for the clarification!

waiting-for-dev (commented):
Thanks for working on that, @elia. This work will also solve #5151.

I'm concerned about modifying the global inflections, as they will also impact the ones used by the host application. I've been experimenting a bit and see no way to have the inflection only available for our engine. Given that, I think the easiest solution is to choose another module name that doesn't require dealing with inflections. An extra benefit will be that we won't need to use a hack from the components generator. What do you think? I don't have a strong opinion about alternatives, but a few come to my mind:
- `Components`
- `Interface`
- `Display`

elia:
> What do you think?

I asked my self the same question, and eventually decided it was ok:
- we are already setting up one acronym for correctly loading files for [RMA](https://github.com/solidusio/solidus/pull/3236/files#diff-38fc1d04f955c7d2c7af5ff20975dd952759a79465f535b188d520b36d68dab3R4)
- I would expect that a host application would have added it already or not use it at all
- having a short one reduces friction when writing (I know, I'm laaazy! 😝) 

That's the background story, but if you want we can add an issue so we don't forget to reassess it before releasing, a massive rename should be rather easy given we always go through the `component` helper.

admin/lib/generators/solidus_admin/component/component_generator.rb
kennyadsl: Another workaround that doesn't require to duplicate the code (but requires a specific file for the inflector initialization) could be to create a separate initializer file in the gem and require it in the generator. I think it might feel less hacky but I'm also ok with this solution, just wanted to point it out.

admin/lib/solidus_admin/engine.rb
```diff
@@ -12,6 +12,11 @@ class Engine < ::Rails::Engine
       require "solidus_admin/configuration"
     end
 
+    initializer "solidus_admin.inflections" do
+      # Suport for UI as an acronym
```
kennyadsl, line 16:
```suggestion
      # Support for UI as an acronym
```

kennyadsl (approved):
I left a couple of comments. 

For the `UI` inflector defined globally, I also don't like it particularly because people might not expect it, but Rails already comes with some magic and I don't think people will be shocked if we add some extra exception to the list. Also, I think that if people have issues it will be quite easy for them to understand where is the problem and fix it (probably there will be some suggestions offered by Ruby itself?).

waiting-for-dev:
> * we are already setting up one acronym for correctly loading files for [RMA](https://github.com/solidusio/solidus/pull/3236/files#diff-38fc1d04f955c7d2c7af5ff20975dd952759a79465f535b188d520b36d68dab3R4)
> * I would expect that a host application would have added it already or not use it at all
> * having a short one reduces friction when writing (I know, I'm laaazy! stuck_out_tongue_closed_eyes)

To be honest, I don't find them strong enough in front of such an easy fix, but your call! :+1:

waiting-for-dev (approved):
:rocket:

elia:
@waiting-for-dev I'm looking into rails autoloaders / zeitwerk to see if there's a way to avoid it 👍

elia:
@waiting-for-dev @kennyadsl I switched to a Zeitwerk-based implementation that should be way less intrusive being active only for `solidus_admin/ui` paths

kennyadsl:
@elia any thoughts on [this](https://github.com/solidusio/solidus/pull/5167#discussion_r1235103988)? Just referencing in case you missed it.

elia:
@kennyadsl sorry, I didn't miss it, it's just the same stackoverflow answer I used for the original solution, which I then changed to use a block instead of a file. 

I think the main concern was in setting `ui` as an acronym for the host app and possible conflicts with existing `Ui` namespaces. Even switching to a file wouldn't have solved it, that's why I didn't push that alternative.

kennyadsl:
Ah got it. I thought they were two different concerns. One related to the inflector name, and another related to adding some custom code in the generator to make it work, due to the fact that the initializers code is not loaded by generators. Anyway, we are using some code in the `before_initialize` hook now so I don't think my suggestion of extracting the inflector initializer in a standalone file is relevant anymore. 👍

waiting-for-dev:
> @waiting-for-dev @kennyadsl I switched to a Zeitwerk-based implementation that should be way less intrusive being active only for `solidus_admin/ui` paths

Honestly, I don't think that monkey-patching the inflector to be able to use the "UI" namespace is acceptable. I'd go with the previous solution if using a non-problematic name is not an option. At least, that would be easier to debug in case of troubles, and leaves the hack as our responsibility.

kennyadsl:
I agree with @waiting-for-dev here. I prefer the old version, the hack is less deep and in a way easier to find for the final user.

elia:
@waiting-for-dev @kennyadsl 👍 

BTW we can surely also go with a different name, I just wanted to give `UI` another try because it is both short and immediately understandable. The other option could be to use `Ui` (lowercase `i`) although at that point if the app is already using it as a namespace they also probably marked it as an acronym already.

In any case switching to a different namespace should be a matter of just a few renames and search/replace operations, I'll revert to the previous proposal and merge.

kennyadsl:
@elia I think reverting is ok. 

I'd avoid `Ui` because if used in the host application in another context it's very likely that its inflector is changed there to `UI`, causing the opposite issue.

PR #5173, opened 2023-06-22
[Admin] products#index empty shell

elia:
## Summary

Just sets up a simple controller with an almost empty component for products#index.

Also adds a `#spree` method to the base component to allow links to go across engines from the new admin back to the old backend (the component context has no access to engine routes natively).

image

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/base_component.rb
+++ b/admin/app/components/solidus_admin/base_component.rb
@@ -9,5 +9,9 @@ module SolidusAdmin
     def stimulus_id
       @stimulus_id ||= self.class.module_parent.to_s.underscore.dasherize.gsub(%r{/}, '--')
     end
+
+    def spree
+      @spree ||= Spree::Core::Engine.routes.url_helpers
+    end
   end
 end
--- a/admin/app/components/solidus_admin/products/index/component.html.erb
+++ b/admin/app/components/solidus_admin/products/index/component.html.erb
@@ -1,0 +1,7 @@
+<div class="<%= stimulus_id %>">
+  <header class="mb-[24px]">
+    <div class="text-[20px] font-[600] leading-[24px]">
+      <%= Spree::Product.model_name.human.pluralize %>
+    </div>
+  </header>
+</div>
--- a/admin/app/components/solidus_admin/products/index/component.rb
+++ b/admin/app/components/solidus_admin/products/index/component.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class SolidusAdmin::Products::Index::Component < SolidusAdmin::BaseComponent
+  def initialize(products:)
+    @products = products
+  end
+end
--- a/admin/spec/components/solidus_admin/base_component_spec.rb
+++ b/admin/spec/components/solidus_admin/base_component_spec.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe SolidusAdmin::BaseComponent, type: :component do
+  describe "#spree" do
+    it "gives access to spree routing helpers" do
+      # The spree/core routes start as empty, so we need to add a route to test
+      Spree::Core::Engine.routes.draw { get '/foo/bar', to: 'foo#bar', as: :foo_bar }
+
+      component = described_class.new
+
+      expect(component.spree).to respond_to(:foo_bar_path)
+    ensure
+      Spree::Core::Engine.routes.clear!
+    end
+  end
+end
--- a/admin/app/controllers/solidus_admin/products_controller.rb
+++ b/admin/app/controllers/solidus_admin/products_controller.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module SolidusAdmin
+  class ProductsController < SolidusAdmin::BaseController
+    def index
+      @products = Spree::Product.all
+    end
+  end
+end
--- a/admin/app/views/solidus_admin/products/index.html.erb
+++ b/admin/app/views/solidus_admin/products/index.html.erb
@@ -1,0 +1,2 @@
+<h1>Products#index</h1>
+<p>Find me in app/views/solidus_admin/products/index.html.erb</p>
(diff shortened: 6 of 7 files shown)


admin/app/components/solidus_admin/products/index/component.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class SolidusAdmin::Products::Index::Component < SolidusAdmin::BaseComponent
+  def initialize(products:)
+    @products = products
+  end
+
+  # @!visibility private
```
kennyadsl, line 8: Can you help me understand we need this line, please?

admin/spec/components/solidus_admin/base_component_spec.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe SolidusAdmin::BaseComponent, type: :component do
+  describe "#spree" do
+    it "gives access to spree routing helpers" do
+      # The spree/core routes start as empty, so we need to add a route to test
```
kennyadsl, line 8: Didn't we add the products resource routes in the previous commit?

rainerdema (approved):
👏

admin/app/components/solidus_admin/base_component.rb
```diff
@@ -9,5 +9,9 @@ class BaseComponent < ViewComponent::Base
     def stimulus_id
       @stimulus_id ||= self.class.module_parent.to_s.underscore.dasherize.gsub(%r{/}, '--')
     end
+
+    def spree
```
kennyadsl, line 13: Did we evaluate adding

```
helper Spree::Core::Engine.routes.url_helpers
```

to admin base controller to access all the Solidus engine routes without being explicit about it?

Btw, it's so annoying to call this method `spree`, do you think we can find a valid alternative name if we keep it?

kennyadsl (commented):
Left a couple of questions/discussion points before approval.

admin/app/components/solidus_admin/base_component.rb
elia: > Did we evaluate adding`helper Spree::Core::Engine.routes.url_helpers` to admin base controller to access all the Solidus engine routes without being explicit about it?
> Btw, it's so annoying to call this method `spree`, do you think we can find a valid alternative name if we keep it?

We picked `spree` because that's the original engine name and matches the way you would call spree routes from any other engine or from the host application itself. Since we're not sure if all the routes between the `Spree::Backend` and `SolidusAdmin` will match I think it's better to be explicit. 

I think that mixing `Spree::Core` url helpers with those in `SolidusAdmin` would create more confusion and that is better to keep them clearly separated. We'll need the `spree` helper only while the two backend/admin are coexisting, once `SolidusAdmin` is complete we should be able to only use helpers coming `SolidusAdmin` routes.

In this case we're gonna need to link from `SolidusAdmin`'s `products#index` to `Spree::Backend` `products#edit`.

Does it make sense?

admin/spec/components/solidus_admin/base_component_spec.rb
elia: For `SolidusAdmin` we have routes that are isolated from `Spree::Core`, and the previous commit only added the resource to the `SolidusAdmin::Engine` routes, not the `Spree::Core::Engine` ones.

admin/app/components/solidus_admin/products/index/component.rb
elia: There's no strict need here, but it will be used in the next PR when we'll add local helper methods.
This Yard directive will tell that all the methods declared after it will be documented as private.

I'll move the addition to the next PR 👍

admin/spec/components/solidus_admin/base_component_spec.rb
kennyadsl: Ah, got it! Makes sense, thanks.

admin/app/components/solidus_admin/base_component.rb
kennyadsl: Ok, so if I understand correctly, while we two backend/admin won't be coexisting anymore, we can deprecate this `spree` method and ask developers just to call the routes straight?

elia: Exactly, once we finish SolidusAdmin we won't need to link back to the old backend and we'll be able to eliminate that helper. Maybe we won't even need a deprecation because the new admin will be experimental while we use it.

admin/app/components/solidus_admin/products/index/component.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="<%= stimulus_id %>">
```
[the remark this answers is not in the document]
elia: For bigger templates using `erb_template` is very uncomfortable 😒 

@waiting-for-dev do you know if that issue applies despite this section in the docs? https://viewcomponent.org/guide/templates.html#inherited

admin/app/components/solidus_admin/base_component.rb
elia: I guess it depends on how we want it to be used, my understanding was that the main nav would have a mix of new and old links on both sides, then if a URL is covered by the new admin we would have shown it.

Cross-linking makes usage seamless, at the cost of less UI/UX consistency. If we set expectations right I think we can go either way.

Thoughts?

admin/app/components/solidus_admin/products/index/component.html.erb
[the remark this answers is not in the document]
elia: I need to try this out, but looks like it's not an issue if you use a template path that's consistent with the inheriting component path

> The Blacklight gem has the file app/components/blacklight/constraint_layout_component.html.erb. I'm placing a modified variant of this file in the same relative path and it's not getting picked up.

https://github.com/ViewComponent/view_component/issues/411#issuecomment-662607011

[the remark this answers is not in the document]
waiting-for-dev: I didn't try, let's investigate

PR #5174, opened 2023-06-22
[Admin] Add basic product listing

elia:
## Summary

- depends on https://github.com/solidusio/solidus/pull/5173

image

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/ui/table/component.html.erb
+++ b/admin/app/components/solidus_admin/ui/table/component.html.erb
@@ -1,0 +1,43 @@
+<div
+  class="
+    <%= stimulus_id %>
+    rounded-lg
+    border
+    border-gray-100
+    border-gray-100
+    overflow-hidden
+  "
+  data-controller="<%= stimulus_id %>"
+>
+  <table class=" table-auto w-full border-collapse ">
+    <thead class="bg-gray-15 color-gray-100">
+      <tr>
+        <% @columns.each do |column| %>
+          <th
+            class="
+              border-b
+              border-gray-100
+              py-[12px]
+              px-[16px]
+              text-[#4f4f4f]
+              text-left
+              text-[14px]
+              font-[600]
+              line-[120%]
+            "
+          ><%= column.name %></th>
+        <% end %>
+      </tr>
+    </thead>
+
+    <tbody class="bg-white text-[14px] line-[150%] text-black">
+      <% @rows.each do |row| %>
+        <tr class="border-b border-gray-100">
+          <% @columns.each do |column| %>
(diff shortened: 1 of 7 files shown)


rainerdema (approved):

admin/app/components/solidus_admin/ui/table/component.rb
```diff
@@ -0,0 +1,26 @@
...
+    name =
+      case name
+      when Symbol then @model_class.human_attribute_name(name)
+      else name
+      end
+
+    @columns << Column.new(name: name, content: block, class_name: class_name)
+  end
+
+  Column = Struct.new(:name, :content, :class_name, keyword_init: true)
```
kennyadsl, line 24: What are the alternatives at defining a class into a component? What about a simple component for the column as well?

admin/app/components/solidus_admin/ui/table/component.html.erb
```diff
@@ -0,0 +1,57 @@
...
+              text-[14px]
+              font-[600]
+              line-[120%]
+            "
+          ><%= column.name %></th>
+        <% end %>
+      </tr>
+    </thead>
+
+    <tbody class="bg-white text-[14px] line-[150%] text-black">
```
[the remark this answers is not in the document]
kennyadsl: I think that if some of us are using it and some don't we'll end up having a lot of troubles when the file is changed later, because it's going to reformat the whole file, even on lines unrelated to the current changes. My vote is to disable it or enable it for anyone (no strong opinion on them, but I don't know the implications of having it enabled on a open source project)

admin/app/components/solidus_admin/ui/table/component.rb
elia: > I'm sure we can extract it even later if we see that it might be a good idea.

💯 

Right now the contents of the cells are complete responsibility of the code calling the `ui/table` component, while the overall table layout, borders, etc. are a responsibility of the `ui/table` component itself.

We can certainly render components inside cells, not sure if we'll ever need to have an alternative row component, which is by definition coupled with the table itself, especially with features like batch actions and live search.

waiting-for-dev: I think the decision here (components vs DSL) is important as it paves the way for a lot of future work and it has effects on the type of product we want to build. Let's talk about it offline! :slightly_smiling_face: :muscle:

waiting-for-dev: My feedback is better given as a follow-up PR :slightly_smiling_face: Please, see https://github.com/solidusio/solidus/pull/5176

Issue #5106, opened 2023-05-30
[Admin] Rebuild the navigation sidebar component


chrean:
Since we are reworking the UX for the Admin, one of the first pieces we want to refresh is the navigation sidebar, using view components and Tailwind if applicable.

This task is quite big because it requires new components to be created.
We will create other tasks, one for each component.

PR #5175, opened 2023-06-22
Style store link from the admin sidebar

waiting-for-dev:
## Summary

A few decisions are made here when it comes to following the provided Figma design. TL;DR: Figma, from which we have less control, wins over Tailwind, which we can change at any time:

- We start configuring Tailwind to avoid using hardcoded `[style]` attributes. We'll continue this whenever we find the need to add new styles.
- Even when both Tailwind and Figma use the same style name and value, we still explicitly configure it in Tailwind. This is to avoid future discrepancies and to make it easier to change. Example: `fontSize/sm`.
- When Tailwind and Figma use the same style value but with different names, we add the Figma name to Tailwind. Ideally, we should try to change Figma to use the same name as Tailwind, but, for now, it'll be easy to compare styles in this way, and a rename from the code part will be straightforward. Example: `fontSize/tiny`.
- When Tailwind and Figma use different style values for the same style name, we change Tailwind to use the Figma value. Example: `fontWeight/bold`, which is `600` for the Figma style and `700` for Tailwind.
- For spacing, we didn't find the need to configure anything different from the default Tailwind spacing scale, so we use it as is. That keeps things much simpler in something that could be cumbersome to think about.

Another rule we followed is that when needing to separate different blocks, we use the top margin from the element below.

We also followed a loose rule where we stack the CSS classes in different lines, from top to bottom:

- Spacing in relation to other elements: display, flex, etc...
- Spacing in relation to the element itself: padding, margin, etc...
- Border styles: color, width, radius...
- Text styles: font size, color...
- Font styles: familiy, weight...
- Background styles: color, image...

screenshot-localhost_3000-2023 06 22-17_14_35

## Checklist

### DIFF
--- a/admin/app/assets/images/solidus_admin/arrow_right_up_line.svg
+++ b/admin/app/assets/images/solidus_admin/arrow_right_up_line.svg
@@ -1,0 +1,5 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g id="arrow-right-up-line">
+<path id="Vector" d="M10.6691 6.27614L4.93141 12.0139L3.98861 11.0711L9.72633 5.33333H4.66915V4H12.0025V11.3333H10.6691V6.27614Z" fill="#A3A3A3"/>
+</g>
+</svg>
--- a/admin/app/components/solidus_admin/sidebar/component.rb
+++ b/admin/app/components/solidus_admin/sidebar/component.rb
@@ -3,6 +3,7 @@
 # Renders the sidebar
 class SolidusAdmin::Sidebar::Component < SolidusAdmin::BaseComponent
   def initialize(
+    store:,
     logo_path: SolidusAdmin::Config.logo_path,
     items: container["main_nav_items"],
     item_component: component("sidebar/item")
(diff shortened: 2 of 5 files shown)


elia (approved):
Awesome!

About the rules followed for this PR I would comment that:

> We also followed a loose rule where we stack the CSS classes in
> different lines, from top to bottom:
> 
> - Spacing in relation to other elements: display, flex, etc...
> - Spacing in relation to the element itself: padding, margin, etc...
> - Border styles: color, width, radius...
> - Text styles: font size, color...
> - Font styles: familiy, weight...
> - Background styles: color, image...

is great, as long as we automate it in some way, as maintaining it manually is super time-consuming.

I'm sure we could add some TW specific rule to erb-formatter and have it handled automatically. But any other suggestions is as good.

rainerdema (approved):
Thanks @waiting-for-dev, LGTM!

waiting-for-dev:
Ref. #5106

waiting-for-dev:
> is great, as long as we automate it in some way, as maintaining it manually is super time-consuming.
>
> I'm sure we could add some TW specific rule to erb-formatter and have it handled automatically. But any other suggestions is as good.

Yeah, I agree. Because of that I set the rule as to be loosely followed... but I think it helps at least when related styles are in the same line. Let's talk more about it. Having a cop for erb-formatter would be super cool.

PR #5182, opened 2023-06-26
[Admin] local component override

elia:
## Summary

This will allow overriding component classes locally to a component without the overhead of setting up keyword arguments for any used component upfront.

The method name is inline with the existing [`with_content`](https://viewcomponent.org/guide/getting-started.html#with_content) that can be used instead of passing a block to the `render` method. 

## Checklist


waiting-for-dev:
Thanks for getting back to my points, @elia! You're right in that you can achieve the same level of flexibility by calling `#with_components` from the initializer, and we could use the convention of always going through the class builder method. Still, I don't feel the need to change it because of a few reasons:

- Maintaining that list of components in the initializer is, in fact, a good thing. Explicitness is, most of the time, better than implicitness, and, in our case, having a clear picture of the used nested components (vs. needing to scan both the component and template files in search of `component` calls) will help users when customizing.
- We'd be adding some complexity for something achievable with built-in Ruby: the `#initialize` method plus keyword arguments.
- It would extend the API surface that users need to know to extend our system.

Happy to hear more thoughts about all of that!

elia:
@waiting-for-dev yes, I generally agree, fundamentally comes down to how much we expect this type of extension to be used. My goal with the whole admin component system is that it will never be necessary to extend a component in this way.

I'll add a couple of counters anyway:

- I'm not sure to understand why anyone would need a list of the components used within another component, if you really need to customize a piece of HTML you'll look into the corresponding template and find what you need
- Maintaining a list of components as keyword arguments + instance variables is a different form of complexity (maybe complicacy would be more correct here)

I'm also fine to drop this feature until its proven that we need it, seems to me that this is an edge case of replacing a component class directly in the registry, which in turn could be way less popular than we think. On this I'm looking forward to when we'll try to re-create some of the existing real-world customization that we'll gather from existing stores.

elia:
This was just a demo/spike, and has served its purpose 🫡

Issue #5187, opened 2023-06-27
Root taxons with at least a child are not valid


elia:
**Solidus Version**:
v4

**To Reproduce**

- Create a taxonomy
- Create a taxon with `taxonomy.root` as its parent
- Check `taxonomy.root.valid?`

**Current behavior**

It returns `false`

**Expected behavior**

Should return `true`

**Additional context**

Looks like creating products with associated taxons having those conditions might trigger that validation in some cases.

cc @filippoliverani

PR #5190, opened 2023-06-28
[v3.4] Fix Taxon taxonomy id validation regression

RyanofWoods:
## Summary

Fixes https://github.com/solidusio/solidus/issues/5187.

In v3.4 I added optional Taxon validations which become mandatory in v4.0.
This caused a regression which made all root taxons with children always invalid.

Fix for v4:
https://github.com/solidusio/solidus/pull/5189

This optional validation for root taxons (parent_id is nil), was
checking for uniqueness of taxonomy_id among all Taxons. This is wrong
however, as the children Taxons of this root Taxon will point to the
same Taxonomy. Instead, we should only look among other root Taxons.

## Checklist

### DIFF
--- a/core/app/models/spree/taxon.rb
+++ b/core/app/models/spree/taxon.rb
@@ -22,7 +22,7 @@ module Spree
     validates :meta_keywords, length: { maximum: 255 }
     validates :meta_description, length: { maximum: 255 }
     validates :meta_title, length: { maximum: 255 }
-    validates :taxonomy_id, uniqueness: { message: :can_have_only_one_root }, if: -> { Spree::Config.extra_taxon_validations && root? }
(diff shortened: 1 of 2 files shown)


RyanofWoods:
Specs seed to have been cancelled after changing the PR base @kennyadsl 🙁

elia (approved):
@RyanofWoods so if I understand correctly the `if` check deceived all of us because it looked like it was applied to the query, while it wasn't, right?

RyanofWoods:
Correct @elia, the `if` check may have caused some to think that only root Taxons are looked at for what `taxonomy_id`s are used. However, instead, it's only used to determine if this validation should be triggered or not. It doesn't affect the scope/query, and the uniqueness scope is needed for this.

I hope my explanations have been clear.

Issue #5186, opened 2023-06-27
[Admin] Integrate arbitrary Tailwind styles into the theme


waiting-for-dev:
Some of the work done so far uses arbitrary Tailwind styles (`-[]` classes). We need to move it to the Tailwind style configuration so that our theme is more easy to change.

components to check
- [ ] ui/badge
- [ ] products/index

PR #5213, opened 2023-07-03
[Admin] Admin UI showcase

elia:
## Summary

Now that we're getting into the nitty gritty of building the new UI with components we need.

This PR sets up component previews for SolidusAdmin and adds a basic [Lookbook](https://lookbook.build) setup to the sandbox. The reason lookbook wasn't added to SolidusAdmin directly is doesn't support being mounted at multiple paths with dedicated configurations. For that reason it's better to prioritize host application setting up their lookbook installation for storefront components if they wish to.

A basic component preview boilerplate is setup when generating components, mainly consisting in an `overview` and a `playground` preview scenarios.

The **playground** scenario will allow playing with dynamic params live from Lookbook, e.g. changing the color of a badge.

The **overview** scenario is intended to showcase the component in different modes (e.g. a button with all its states and options in a single matrix/table).
The `overview` can also act as a simple smoke test for the component and can act as basic spec coverage for the component that ensures is not breaking when provided with different options.

### Other approaches

Before settling to starting with the `overview` and `playground` scenarios I tried a bunch of different approaches using [Lookbook groups](https://lookbook.build/guide/previews/groups), including: 
- typing out all the scenarios manually: `def badge_with_green_color_and_m_size…`
- using a local macro `def self.add_scenario…` followed by `add_scenario color: :green, size: :m`
and other in-between approaches.

Eventually it was very time consuming and repetitive, even when using the `# !@macro` directive of YARD.
The solution I settled for is generating the overview template with a simple layout similar to the one used for Lookbook groups, and for the badge component using a simple table resembling the Figma designs I used to build it.

That seemed the best compromise that also minimized effort and duplication.

<details><summary>Details</summary>
<p>

```rb
  # @!macro [attach] with_color_and_size
  #   @!method $1_$2
  def self.with_color_and_size(color, size)
    define_method "#{color}_#{size}" do
      configurable(color: color, size: size)
    end
  end

  # @!group Graphite Light
  with_color_and_size :graphite_light, :s
  with_color_and_size :graphite_light, :m
  with_color_and_size :graphite_light, :l
  # @!endgroup

  # @!group Red
  with_color_and_size :red, :s
  with_color_and_size :red, :m
  with_color_and_size :red, :l
  # @!endgroup

  # @!group Green
  with_color_and_size :green, :s
  with_color_and_size :green, :m
  with_color_and_size :green, :l
  # @!endgroup

  # @!group Blue
  with_color_and_size :blue, :s
  with_color_and_size :blue, :m
  with_color_and_size :blue, :l
  # @!endgroup

  # @!group Black
  with_color_and_size :black, :s
  with_color_and_size :black, :m
  with_color_and_size :black, :l
  # @!endgroup

  # @!group Yellow
  with_color_and_size :yellow, :s
  with_color_and_size :yellow, :m
  with_color_and_size :yellow, :l
  # @!endgroup

  # # @!group Medium
  # with_color_and_size :graphite_light, :m
  # with_color_and_size :red, :m
  # with_color_and_size :green, :m
  # with_color_and_size :blue, :m
  # with_color_and_size :black, :m
  # with_color_and_size :yellow, :m
  # # @!endgroup

  # # @!group Small
  # with_color_and_size :graphite_light, :s
  # with_color_and_size :red, :s
  # with_color_and_size :green, :s
  # with_color_and_size :blue, :s
  # with_color_and_size :black, :s
  # with_color_and_size :yellow, :s
  # # @!endgroup
```

</p>
</details> 

## Screenshots

image
image

### ViewComponent previews

The original VC previews are still accessible at `http://localhost:3000/rails/view_components`

image

`http://localhost:3000/rails/view_components/solidus_admin/ui/badge/component/overview`

image

`http://localhost:3000/rails/view_components/solidus_admin/ui/badge/component/playground`

image

`http://localhost:3000/rails/view_components/solidus_admin/ui/badge/component/playground?name=Foo`

image

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/ui/badge/component.rb
+++ b/admin/app/components/solidus_admin/ui/badge/component.rb
@@ -12,12 +12,18 @@ class SolidusAdmin::UI::Badge::Component < SolidusAdmin::BaseComponent
     yellow: 'text-yellow bg-[#feecd4]',
   }.freeze
 
-  def initialize(name:, color: :graphite_light)
+  SIZES = {
+    s: 'leading-[16px] px-[8px] py-[2px] text-[12px] font-[500]',
+    m: 'leading-[20px] px-[12px] py-[2px] text-[14px] font-[500]',
+    l: 'leading-[24px] px-[12px] py-[2px] text-[16px] font-[500]',
+  }.freeze
+
+  def initialize(name:, color: :graphite_light, size: :m)
     @name = name
 
     @class_name = [
       'inline-flex items-center rounded-full', # layout
-      'leading-[20px] px-[12px] py-[2px] text-[14px] font-[500]', # size
+      SIZES.fetch(size.to_sym), # size
       COLORS.fetch(color.to_sym), # color
     ].join(' ')
   end
--- a/admin/lib/generators/solidus_admin/install/install_generator.rb
+++ b/admin/lib/generators/solidus_admin/install/install_generator.rb
@@ -3,7 +3,9 @@
 module SolidusAdmin
   module Generators
     class InstallGenerator < Rails::Generators::Base
-      source_root File.expand_path('templates', __dir__)
+      class_option :lookbook, type: :boolean, default: !!ENV['SOLIDUS_ADMIN_LOOKBOOK'], desc: 'Install Lookbook for component previews'
+
+      source_root "#{__dir__}/templates"
 
       def install_solidus_core_support
         route <<~RUBY
@@ -23,6 +25,18 @@ module SolidusAdmin
 
       def build_tailwind
         rake "solidus_admin:tailwindcss:build"
+      end
+
+      def install_lookbook
+        return unless options[:lookbook]
+
+        gem_group :development, :test do
+          gem "lookbook"
+          gem "listen"
+          gem "actioncable"
+        end
+
+        route "mount Lookbook::Engine, at: '/lookbook' if Rails.env.development?"
       end
     end
   end
--- a/admin/lib/solidus_admin/configuration.rb
+++ b/admin/lib/solidus_admin/configuration.rb
@@ -27,6 +27,7 @@ module SolidusAdmin
       SolidusAdmin::Engine.root.join("app/assets/javascripts/**/*.js"),
       SolidusAdmin::Engine.root.join("app/views/**/*.erb"),
       SolidusAdmin::Engine.root.join("app/components/**/*.{rb,erb,js}"),
+      SolidusAdmin::Engine.root.join("spec/components/previews/**/*.erb"),
 
       Rails.root.join("public/solidus_admin/*.html"),
       Rails.root.join("app/helpers/solidus_admin/**/*.rb"),
--- a/admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
+++ b/admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# @component "ui/badge"
+class SolidusAdmin::UI::Badge::ComponentPreview < ViewComponent::Preview
+  layout "solidus_admin/preview"
+  include SolidusAdmin::ContainerHelper
+
+  # @param name [String]
+  def overview(name: "Label")
+    render_with_template(locals: { name: name, component: component("ui/badge") })
+  end
+
+  # @param name [String]
+  # @param color select :color_options
+  # @param size select :size_options
+  def playground(name: "Label", color: :green, size: :m)
+    render component("ui/badge").new(name: name, color: color, size: size)
+  end
+
+  private
+
+  def size_options
+    component('ui/badge')::SIZES.keys
+  end
+
+  def color_options
+    component('ui/badge')::COLORS.keys
+  end
+end
--- a/admin/spec/components/solidus_admin/ui/badge/component_spec.rb
+++ b/admin/spec/components/solidus_admin/ui/badge/component_spec.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe SolidusAdmin::UI::Badge::Component, type: :component do
+  it "renders the overview preview" do
+    render_preview(:overview)
+  end
+end
--- a/bin/sandbox
+++ b/bin/sandbox
@@ -95,9 +95,8 @@ fi
 unbundled bin/rails db:drop db:create
 
 echo "~~~> Running the solidus:install generator"
-unbundled bin/rails generate solidus:install \
-  --auto-accept \
-  $@
+export SOLIDUS_ADMIN_LOOKBOOK=true
+unbundled bin/rails generate solidus:install --auto-accept $@
 
 echo "
 🚀 This app is intended for test purposes. If you're interested in running
--- a/admin/app/views/layouts/solidus_admin/preview.html.erb
+++ b/admin/app/views/layouts/solidus_admin/preview.html.erb
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <%= stylesheet_link_tag "solidus_admin/application.css", "inter-font", "data-turbo-track": "reload" %>
+    <%= javascript_importmap_tags "solidus_admin/application", shim: false, importmap: SolidusAdmin.importmap %>
+  </head>
+  <body class="p-3">
+    <%= yield %>
+  </body>
+</html>
--- a/admin/lib/generators/solidus_admin/component/component_generator.rb
+++ b/admin/lib/generators/solidus_admin/component/component_generator.rb
@@ -10,6 +10,7 @@ module SolidusAdmin
     class_option :i18n, type: :boolean, default: true
     class_option :js, type: :boolean, default: true
     class_option :spec, type: :boolean, default: true
+    class_option :preview, type: :boolean, default: true
 
     def setup_inflections
       # This is needed because the generator won't run the initialization process,
@@ -29,9 +30,19 @@ module SolidusAdmin
       template "component.yml", destination(".yml") if options["i18n"]
       template "component.js", destination(".js") if options["js"]
       template "component_spec.rb", destination("_spec.rb", root: "spec/components") if options["spec"]
+
+      if options["preview"]
+        preview_destination_path = destination("_preview.rb", root: "spec/components/previews")
+        template "component_preview.rb", preview_destination_path
+        template "component_preview_overview.html.erb", preview_destination_path.sub(/\.rb/, '/overview.html.erb')
+      end
     end
 
     private
+
+    def component_registry_id
+      [class_path.presence, file_name].compact.join("/")
+    end
 
     def destination(suffix, root: "app/components")
       File.join(root, class_path, file_name, "component#{suffix}")
@@ -61,6 +72,32 @@ module SolidusAdmin
 
     def initialize_body
       attributes.map { |attr| "@#{attr.name} = #{attr.name}" }.join("\n    ")
+    end
+
+    def preview_signature
+      return if attributes.blank?
+
+      signature = attributes.map { |attr| "#{attr.name}: #{attr.name.to_s.inspect}" }.join(", ")
+      "(#{signature})"
+    end
+
+    def preview_playground_yard_tags
+      return if attributes.blank?
+
+      attributes.map { |attr| "@param #{attr.name} [String]" }.join("\n  ")
+    end
+
+    def preview_playground_body
+      render_signature = attributes.map { |attr| "#{attr.name}: #{attr.name}" }.join(", ")
+      render_signature = "(#{render_signature})" if render_signature.present?
+
+      "render component(#{component_registry_id.inspect}).new#{render_signature}"
+    end
+
+    def preview_overview_body
+      component_registry_id = [class_path.presence, file_name].compact.join("/")
+
+      "render component(#{component_registry_id.inspect}).new#{preview_signature}"
     end
 
     def attributes_html
--- a/admin/lib/generators/solidus_admin/component/templates/component_preview.rb.tt
+++ b/admin/lib/generators/solidus_admin/component/templates/component_preview.rb.tt
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+# @component <%= component_registry_id.inspect %>
+class <%= File.join(*[namespaced_path, file_path].compact).classify %>::ComponentPreview < ViewComponent::Preview
+  layout "solidus_admin/preview"
+  include SolidusAdmin::ContainerHelper
+
+  def overview
+    render_with_template(locals: { name: name, component: component(<%= component_registry_id.inspect %>) })
+  end
+
+  <%= preview_playground_yard_tags %>
+  def playground<%= preview_signature %>
+    <%= preview_playground_body %>
+  end
+end
--- a/admin/lib/generators/solidus_admin/component/templates/component_preview_overview.html.erb
+++ b/admin/lib/generators/solidus_admin/component/templates/component_preview_overview.html.erb
@@ -1,0 +1,7 @@
+<div class="mb-8">
+  <h6 class="text-gray-500 mb-3 mt-0">
+    Scenario 1
+  </h6>
+
+  <%%= render component.new<%= preview_signature %> %>
+</div>
--- a/admin/lib/generators/solidus_admin/component/templates/component_spec.rb.tt
+++ b/admin/lib/generators/solidus_admin/component/templates/component_spec.rb.tt
@@ -3,7 +3,9 @@
 require "spec_helper"
 
 RSpec.describe <%= File.join(*[namespaced_path, file_path].compact).classify %>::Component, type: :component do
-  pending "add some examples to (or delete) #{__FILE__}"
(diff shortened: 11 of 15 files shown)


bin/sandbox
```diff
@@ -99,7 +99,11 @@ unbundled bin/rails generate solidus:install \
   --auto-accept \
   $@
 
+unbundled bundle add lookbook listen actioncable --group development,test
+ruby -e "File.write('config/routes.rb', File.read('config/routes.rb').gsub(/draw do/, %{draw do\n  mount Lookbook::Engine, at: '/lookbook' if Rails.env.development?}))"
```
rainerdema, line 103: Just a note on this, as we discussed during the live review, this could be moved from here probably 🤔
Even if to be honest, I don't have a better idea for the moment on how to do it differently.

rainerdema (approved):
This will be extremely useful for developing various UI components and for developers who want to do their own integration on the new Solidus admin! 😍
Looks great 👏

bin/sandbox
```diff
@@ -99,7 +99,11 @@ unbundled bin/rails generate solidus:install \
   --auto-accept \
   $@
 
+unbundled bundle add lookbook listen actioncable --group development,test
```
waiting-for-dev, line 102: For some reason, it looks like the gems are not added to the sandbox's `Gemfile`.

admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# @component "ui/badge"
+class SolidusAdmin::UI::Badge::ComponentPreview < ViewComponent::Preview
+  layout "solidus_admin/preview"
+  include SolidusAdmin::ContainerHelper
```
waiting-for-dev, line 6: What do you think about moving it into a `SolidusAdmin::BaseComponentPreview` class? I guess we'll need those for all other previews...

admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  # @param name [String]
+  def overview(name: "Label")
+    render_with_template(locals: { name: name, component: component("ui/badge") })
```
waiting-for-dev, line 10: If we add the `BaseComponentPreview`, maybe we can extract that `component("ui/badge")` into a helper method `component`? We can do it later, though.

admin/spec/components/solidus_admin/ui/badge/component_spec.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe SolidusAdmin::UI::Badge::Component, type: :component do
+  it "renders the overview preview" do
+    render_preview(:overview)
```
waiting-for-dev, line 7: I tried it locally and it's failing with:

```ActionView::MissingTemplate:
       Missing template layouts/solidus_admin/preview with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby]}.
     
       Searched in:
         * "/home/solidus_user/app/admin/spec/components/previews"
         * "/home/solidus_user/app/admin/spec/dummy/app/views"
         * "/home/solidus_user/gems/gems/view_component-3.2.0/app/views"
         * "/home/solidus_user/gems/gems/railties-7.0.4.3/lib/rails/templates"
     
       Did you mean?  solidus_admin/ui/badge/component_preview/overview
                      view_components/preview
                      view_components/previews
                      layouts/application
                      rails/info/properties
                      solidus_admin/ui/badge/component_preview
```

For some reason, it looks like CI is only running `core` & `backend` for this PR :thinking: It probably has nothing to do with the work here, so we need to investigate...

waiting-for-dev (commented):
Thanks, @elia! I love having this feature, and I think the overview/playground structure can work well. Left some comments, and I think there are a few issues that need to be fixed.

admin/app/components/solidus_admin/ui/badge/component.rb
```diff
@@ -12,12 +12,18 @@ class SolidusAdmin::UI::Badge::Component < SolidusAdmin::BaseComponent
     yellow: 'text-yellow bg-[#feecd4]',
   }.freeze
 
-  def initialize(name:, color: :graphite_light)
+  SIZES = {
+    s: 'leading-[16px] px-[8px] py-[2px] text-[12px] font-[500]',
+    m: 'leading-[20px] px-[12px] py-[2px] text-[14px] font-[500]',
+    l: 'leading-[24px] px-[12px] py-[2px] text-[16px] font-[500]',
```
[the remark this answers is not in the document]
elia: Yes, I would consider that out of the scope of this PR. I added this component and product index to a list of think to be checked in #5186 👍

admin/lib/solidus_admin/configuration.rb
```diff
@@ -27,6 +27,7 @@ class Configuration < Spree::Preferences::Configuration
       SolidusAdmin::Engine.root.join("app/assets/javascripts/**/*.js"),
       SolidusAdmin::Engine.root.join("app/views/**/*.erb"),
       SolidusAdmin::Engine.root.join("app/components/**/*.{rb,erb,js}"),
+      SolidusAdmin::Engine.root.join("spec/components/previews/**/*.{rb,erb,js,html,md}"),
```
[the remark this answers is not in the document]
elia: This was to easily style surrounding html in the overview scenario or for potential other scenarios in which we want to show the component in a specific context.

Of course some of those styles might not be strictly needed in the actual admin, but I think it's just a few classes and it's not worth having a separate TW build or using inline style attributes.

bin/sandbox
elia: Is it a silent error or you get anything more specific? 
Also I suggest checking if it's the latest version of bundler, just to be sure.

bin/sandbox
elia: I'll try to move both operations to the solidus_admin install generator 👍

admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
elia: This bothers me as well, 😞 and I tried, but for some reason the preview broke, after a while I stopped spending time on it and just added the two lines to the generator. I think over time we might give it another crack and see if we succeed. 

I can try again if you think this is needed right away.

admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
elia: Good call! You think would be worth it having a local helper with some duplication assuming we won't be able to extract the superclass?

admin/spec/components/solidus_admin/ui/badge/component_spec.rb
elia: I tried locally and it works fine, maybe we used different commands?

```
solidus_admin:elia/admin/ui-showcase ⤑ bin/rspec admin/spec/components/solidus_admin/ui/badge/component_spec.rb                                   ~/C/N/solidus_admin
$ cd admin; bundle exec rspec /Users/elia/Code/Nebulab/solidus_admin/admin/spec/components/solidus_admin/ui/badge/component_spec.rb; cd -

Randomized with seed 64336
.

Finished in 0.24157 seconds (files took 2.02 seconds to load)
1 example, 0 failures

Randomized with seed 64336
```

admin/lib/solidus_admin/configuration.rb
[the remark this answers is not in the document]
waiting-for-dev: I see. It's probably an acceptable compromise. However, we probably only need the `erb` extension now.

bin/sandbox
waiting-for-dev: Very silent :slightly_smiling_face: Do you get it added in your sandbox?

admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
waiting-for-dev: It's not needed now, but it probably will be needed for the next one :slightly_smiling_face: We can work on it separately, though.

admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
waiting-for-dev: I'm not sure I follow... if we had the `BaseComponentPreview` (which it looks it's not that easy) we should be able to extract the component from the preview class name, right? `class SolidusAdmin::UI::Badge::ComponentPreview` => `ui/badge`

elia: Yes, if we had the BaseComponentPreview that would be super easy, I was wondering if would make sense to have it in every preview class while we don't have the base class. Sorry about the confusion.

bin/sandbox
elia: It is now part of the install generator, does it work for you now?

admin/spec/components/solidus_admin/ui/badge/component_spec.rb
elia: > For some reason, it looks like CI is only running `core` & `backend` for this PR 🤔 It probably has nothing to do with the work here, so we need to investigate...

@waiting-for-dev I was fooled by CircleCI as well at first, you can find `admin` in one of the other parallel runs, `bin/build-ci` tries to split the work among parallel runs.

image

waiting-for-dev: That's really weird. @rainerdema, @kennyadsl, could you try it and confirm who's seeing ghosts here? :slightly_smiling_face: I tried once more, and it fails for me...`q

admin/spec/components/previews/solidus_admin/ui/badge/component_preview.rb
waiting-for-dev: Yeah, that works if the base class is not an option.

bin/sandbox
kennyadsl: I tried it yesterday and it was there.

admin/lib/generators/solidus_admin/install/install_generator.rb
```diff
@@ -24,6 +24,19 @@ def ignore_tailwind_build_files
...
+        # The dummy app has no Gemfile
+        return unless in_root { File.exist?("Gemfile") }
+
+        gem_group :development, :test do
+          gem "lookbook"
+          gem "listen"
+          gem "actioncable"
+        end
+
+        route "mount Lookbook::Engine, at: '/lookbook' if Rails.env.development?"
```
waiting-for-dev, line 38: That will affect every `solidus_admin` installation unless I'm missing something. What about controlling through an environment variable and we set it in the sandbox script?

admin/spec/components/solidus_admin/ui/badge/component_spec.rb
kennyadsl: Let me try again!

bin/sandbox
waiting-for-dev: I tried now and it works, but I left a comment about the approach.

admin/lib/generators/solidus_admin/install/install_generator.rb
kennyadsl: That's a good idea!

admin/spec/components/solidus_admin/ui/badge/component_spec.rb
elia: @rainerdema had the same issue, looks like there's a fix for that issue in ViewComponent 3.3, I'll bump the dependency in solidus_admin.gemspec 👍 

Probably me and @kennyadsl already have 3.3.

admin/lib/generators/solidus_admin/install/install_generator.rb
elia: I think it's fine to install it anyway, and the starter should use it as well, that said I added a class option with the default dependant on the `SOLIDUS_ADMIN_LOOKBOOK` env variable, and changed the sandbox script accordingly.

admin/lib/generators/solidus_admin/install/install_generator.rb
```diff
@@ -24,6 +26,21 @@ def ignore_tailwind_build_files
       def build_tailwind
         rake "solidus_admin:tailwindcss:build"
       end
+
+      def install_lookbook
+        return unless options[:lookbook]
+
+        # The dummy app has no Gemfile
+        return unless in_root { File.exist?("Gemfile") }
```
waiting-for-dev, line 34: Nit, but maybe we can now remove these lines, as we won't provide `SOLIDUS_ADMIN_LOOKBOOK` variable for the dummy app.

admin/spec/components/solidus_admin/ui/badge/component_spec.rb
waiting-for-dev: I bumped the dependency to 3.3, but I'm still experiencing the problem. Is it the same for you, @rainerdema? :thinking:

admin/lib/generators/solidus_admin/install/install_generator.rb
```diff
@@ -3,7 +3,9 @@
 module SolidusAdmin
   module Generators
     class InstallGenerator < Rails::Generators::Base
-      source_root File.expand_path('templates', __dir__)
+      class_option :lookbook, type: :boolean, default: !!ENV['SOLIDUS_ADMIN_LOOKBOOK'], banner: 'Install Lookbook for component previews'
```
kennyadsl, line 6: I tried to run ` bin/rails g solidus_admin:install --help`

```
Running Rails from: /Users/kennyadsl/Code/nebulab/solidus/sandbox
Usage:
  rails generate solidus_admin:install [options]

Options:
  [--skip-namespace], [--no-skip-namespace]                              # Skip namespace (affects only isolated engines)
  [--skip-collision-check], [--no-skip-collision-check]                  # Skip collision check
  [--lookbook=Install Lookbook for component previews], [--no-lookbook]  # Indicates when to generate lookbook

Runtime options:
  -f, [--force]                    # Overwrite files that already exist
  -p, [--pretend], [--no-pretend]  # Run but do not make any changes
  -q, [--quiet], [--no-quiet]      # Suppress status output
  -s, [--skip], [--no-skip]        # Skip files that already exist

Description:
    Create solidus admin files for install generator.
```

And I think there's something off with the `lookbook` option. It's setting the banner in the value of the option, which doesn't seem correct to me.

kennyadsl (commented):
I think there's something to review in the install generator options, and there's a commit to rebase. Other than that, I'm good to go, looking forward to approval and merging!

admin/lib/generators/solidus_admin/install/install_generator.rb
elia: @kennyadsl good catch, I copied it from the core install generator, but maybe being a different superclass the syntax is different, fixing it 👍

admin/spec/components/solidus_admin/ui/badge/component_spec.rb
elia: @waiting-for-dev the dependency must be updated on the root Gemfile, not the sandbox one, I thought it was worth mentioning since this tricked me at first.

rainerdema: @waiting-for-dev I can confirm that I was experiencing the same exception yesterday.
But after I switched from version 3.2.0 to 3.3.0, the error ceased to occur 🤔

The fix for the problem released in 3.3.0 seems to be this specifically:
https://github.com/ViewComponent/view_component/pull/1777

admin/lib/generators/solidus_admin/install/install_generator.rb
```diff
@@ -24,6 +26,18 @@ def ignore_tailwind_build_files
...
+      def install_lookbook
+        return unless options[:lookbook]
+
+        gem_group :development, :test do
+          gem "lookbook"
+          gem "listen"
+          gem "actioncable"
+        end
+
+        route "mount Lookbook::Engine, at: '/lookbook' if Rails.env.development?"
```
kennyadsl, line 39: @elia can you help me understand why we need `if Rails.env.development?"` if we also have the ENV variable?

kennyadsl (commented):
👍 left a last, non-blocking thing. Thanks!

kennyadsl (approved):
Approving because the last thing was not blocking 😅

admin/lib/generators/solidus_admin/install/install_generator.rb
elia: That is how lookbook suggests to add it to routes, the same goes for the bundler groups the gems will belong to. Also the env variable is about whether we'll install it, not for when it should be active.

I think it's good to stick to lookbook's instructions when we install it to the host app.

admin/spec/components/solidus_admin/ui/badge/component_spec.rb
waiting-for-dev: Bumping to v3.3.0 alone didn't fix my issue, but after re-generating my `Gemfile.lock` file the issue is gone.

waiting-for-dev (approved):
:rocket:

PR #5214, opened 2023-07-03
[Admin] Authentication configuration spike/demo

elia:
## Summary

**This PR is not intended for merging (yet)**

Having the long term goal of separating the storefront customers form the admin users, would be nice to create the layer of configuration that will then allow us to replace with a different setup when the time comes.

_The configuration options were inspired from the extensive experience [gathered by ActiveAdmin](https://activeadmin.info/1-general-configuration.html#authentication) over the years._

A default setup that taps in the existing `solidus_backend` authentication is provided and will be helpful during the transition from `backend` to `admin`. Eventually `solidus_auth_devise` can support `solidus_admin` directly or an alternative system can be provided.

## Checklist


admin/lib/solidus_admin/configuration.rb
```diff
@@ -59,6 +59,19 @@ class Configuration < Spree::Preferences::Configuration
...
+    preference :authentication_method, :string, default: :authenticate_solidus_backend_user!
+
+    # The method used to retrieve the current user in the admin interface.
+    preference :current_user_method, :string, default: :spree_current_user
+
+    # The path used to logout the user in the admin interface.
+    preference :logout_link_path, :string, default: :destroy_solidus_user_session_path
+
+    # The HTTP method used to logout the user in the admin interface.
+    preference :logout_link_method, :string, default: :delete
```
rainerdema, line 74: I like the idea of having the destroy session path and the action to use for it here 👍
(even if they do not seem to be used at the moment 👀 )

About the logout_link_method, this seems similar to the sign_out_via option on Devise:
https://github.com/heartcombo/devise/blob/ec0674523e7909579a5a008f16fb9fe0c3a71712/test/rails_app/config/initializers/devise.rb#L172-L173

rainerdema (commented):
Thank you, @elia! Having that simple configuration for authentication seems like a good approach to me 👍

kennyadsl:
This might make a lot of sense @elia! 

Can you please share a high-level overview of how you envision the long-term plan to separate operators and users? I understand this is just the first piece but I miss all the rest of the strategy to leave some feedback that makes sense.

admin/app/controllers/solidus_admin/backend_auth_support.rb
```diff
@@ -0,0 +1,28 @@
...
+  rescue CanCan::AccessDenied
+    instance_exec(&Spree::Admin::BaseController.unauthorized_redirect)
+  end
+
+  # Needs to be overriden so that we use Spree's Ability rather than anyone else's.
+  def current_ability
+    @current_ability ||= Spree::Ability.new(spree_current_user)
+  end
+
+  def store_location
```
waiting-for-dev, line 25: Where is this method used? :thinking:

admin/app/controllers/solidus_admin/backend_auth_support.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module SolidusAdmin::BackendAuthSupport
```
waiting-for-dev, line 3: Shouldn't we also define the `spree_current_user` method? Or is it already available from somewhere else?

admin/app/controllers/solidus_admin/auth.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module SolidusAdmin::Auth
+  extend ActiveSupport::Concern
+
+  # Support the legacy Solidus Backend authentication method
+  include SolidusAdmin::BackendAuthSupport
```
waiting-for-dev, line 7: Given that `SolidusAdmin::BaseController` includes `SolidusAdmin::Auth`, we include `BackendAuthSupport` no matter what users configure. I know this is in the preliminary phase, but have you thought about how we would deal with this `include` line to be dependent on the configuration? I'm thinking that maybe what should be configurable is the name of the module, and that we should expect it to define the expected interface. Then, for now, we can default to `SolidusAdmin::BackendAuthSupport`. Does it make sense?

```ruby
# admin/lib/solidus_admin/configuration.rb

class_name_attribute  :authentication_module, default: `SolidusAdmin::BackendAuthSupport`
```

waiting-for-dev (commented):
Thanks for raising it now, @elia. We're at the right moment to make solidus admin authentication decoupled from the one for customers, as it'll guide us in making the right architectural decisions. I'd also like to hear about the rest of the plan when that's ready, but this is a significant first step! I left some questions.

admin/app/controllers/solidus_admin/backend_auth_support.rb
elia: It's coming from `spree_auth_devise`

admin/app/controllers/solidus_admin/auth.rb
```diff
@@ -0,0 +1,35 @@
...
+    helper_method :current_solidus_admin_user
+    helper_method :solidus_admin_logout_path
+    helper_method :solidus_admin_logout_method
+  end
+
+  private
+
+  def authenticate_solidus_admin_user!
+    send SolidusAdmin::Config.authentication_method if SolidusAdmin::Config.authentication_method
+  end
```
waiting-for-dev, line 22: Why not define those methods in the included module adapter?

admin/lib/solidus_admin/configuration.rb
```diff
@@ -59,6 +59,23 @@ class Configuration < Spree::Preferences::Configuration
...
+    preference :authentication_method, :string, default: :authenticate_solidus_backend_user!
+
+    # The method used to retrieve the current user in the admin interface.
+    preference :current_user_method, :string, default: :spree_current_user
+
+    # The path used to logout the user in the admin interface.
+    preference :logout_link_path, :string, default: :admin_logout_path
+
+    # The HTTP method used to logout the user in the admin interface.
+    preference :logout_link_method, :string, default: :delete
```
waiting-for-dev, line 74: In the same direction as the above comment, if we expect the adapter to define those methods, there's no need to define them in the initializer.

admin/app/controllers/solidus_admin/base_controller.rb
```diff
@@ -1,7 +1,11 @@
 # frozen_string_literal: true
 
 module SolidusAdmin
-  class BaseController < Spree::BaseController
+  class BaseController < ApplicationController
+    include ActiveStorage::SetCurrent
+    include ::SolidusAdmin::Auth
+    include Spree::Core::ControllerHelpers::Store
```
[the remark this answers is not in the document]
waiting-for-dev: I understand they'll be used, but are they needed in the context of the PR configuring authentication? :thinking:

admin/app/controllers/solidus_admin/backend_auth_support.rb
waiting-for-dev: I see. Do you think it's worth it to drop a comment so it's clear for those reading the code?

admin/app/controllers/solidus_admin/backend_auth_support.rb
waiting-for-dev: As above, do you think a comment will make things clearer?

kennyadsl:
@elia reading the plan, I have a question. I know it's hard to predict at this stage but if you already wrapped your mind around this, what's the DX for making the switch? I imagine at some point, we will release a rake task that converts all the users that have some admin/operator roles to the other model, with the right roles assigned. That rake task should be run right before the switch of the `authentication_adapter` preference (otherwise, for example, people could change the roles on the backend side and that not be reflected on the admin side), and after changing that preference, we have to be sure that the functions related to roles are disabled on the backend, to avoid errors (could be that people use the new admin auth for admins with the old backend? 🤯 ).

elia:
> could be that people use the new admin auth for admins with the old backend?

I think in this regard we have an opportunity to leverage the new admin for the switch, and the old backend will still run on the spree_users table

So once the apps are fully switched we'll offere a rake task for the migration and preferences, but I think for some time the core will need to still allow using the old table. That shouldn't be a big problem if everything related to admins has stopped assuming it's the same model as customers (that's probably the biggest lift).

admin/app/controllers/solidus_admin/base_controller.rb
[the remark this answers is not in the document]
elia: _Been a long time, but I'll add an answer anyway 😊._

Those are not unused, `current_store` is necessary for the sidebar, while ActiveStorage::SetCurrent is needed in order to generate blob urls, removing it would cause this error (I erroneously thought it was only needed for uploading):

image

See https://github.com/rails/rails/blob/7-0-stable/activestorage/app/controllers/concerns/active_storage/set_current.rb#L3-L6

admin/app/controllers/solidus_admin/auth.rb
elia: This file contains the adapter mechanics, providing the hooks that different adapter can use. The backend one is of course the one we need while migrating out of it, but eventually I think other will pop up (devise, maybe 3rd party single sign on, etc.) according to "the plan" in the description.

admin/spec/features/accounts_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe "Accounts", type: :feature do
+  let(:user) { create(:admin_user, email: 'admin@example.com') }
+
+  before do
+    allow_any_instance_of(SolidusAdmin::BaseController).to receive(:spree_current_user).and_return(user)
```
elia, line 9: I think eventually we'll need a helper method for doing the sign in programmatically, but for now this is ok 👍

elia (commented):
✅ Thanks for taking care of this @the-krg!

the-krg (approved):

admin/app/views/solidus_admin/accounts/show.html.erb
```diff
@@ -3,4 +3,5 @@
   class="text-4xl"
 ><a href="#" data-action="hello#greet">Account</a></h1>
 
-<h2 class="text-3xl underline">Your account info</h1>
+<p>Logged in as <%= current_solidus_admin_user.email %></p>
+<p><%= button_to "Log out", solidus_admin_logout_path, method: solidus_admin_logout_method, class: 'underline' %></p>
```
rainerdema, line 7: This view could be part of a new "account/show" component to be rendered (like "products/index").
Non-blocking for now, I've added a dedicated point in the list of the final things to do before the release:
- [[Admin] Release SolidusAdmin v0.0.1.alpha.1 #5280](https://github.com/solidusio/solidus/issues/5280)
- [[Admin] Refactor "accounts/show" view into a separate component #5292](https://github.com/solidusio/solidus/issues/5292)

rainerdema (approved):
Great job! 🥇

waiting-for-dev:
Thanks, @elia & @the-krg, for working on that! This is a very nice step towards having a decoupled authentication system :heart:

I'm only left wondering if there's a benefit to the extra indirection we're adding through the option to configure the adapter method names. If the adapter is configurable, and we expect users to create those interfaces, why not fix the method names?

In short, instead of:

```ruby
# admin/lib/solidus_admin/configuration.rb
preference :authentication_method, :string, default: :authenticate_solidus_backend_user!
preference :current_user_method, :string, default: :spree_current_user
preference :logout_link_path, :string, default: :admin_logout_path
preference :logout_link_method, :string, default: :delete
preference :authentication_adapter, :string, default: 'SolidusAdmin::AuthAdapters::Backend'
```

We could only have the following:

```ruby
preference :authentication_adapter, :string, default: 'SolidusAdmin::AuthAdapters::Backend'
```

With:

```ruby
# admin/app/controllers/solidus_admin/auth_adapters/backend.rb
def authentication_method
  authenticate_solidus_backend_user!
end

def current_user_method
  spree_current_user
end

def logout_link_path
  admin_logout_path
end

def logout_link_method
  :delete
end
```

That's to say: simply require the configured module to implement the expected interface. `SolidusAdmin::Auth` would also be simplified by not needing to define those methods.

Am I missing some context? Thoughts, @elia, @the-krg, @rainerdema?

elia:
@waiting-for-dev that's a completely valid alternative, and that's why it made sense to pick either the configurations or the adapter (and we picked the configuration).

The reason the configuration was preferred is that generally any existing authentication system won't require any adapter and will have those methods in place (one way or another). The only reason that made auth-devise need an adapter is that weird entanglement of authentication and authorization that we have.

In other words we're building on the experience accrued by active-admin with the following benefits:
- auth systems won't need any glue code
- auth systems will add (or let you add) those authentication methods to ApplicationControlller or some other base controller
- generally with this solution we increase the number of configuration points but we don't need to write any code or patch anything
- in the worst case (or in case of a completely custom auth system) it's still easy to plug into it

Hope this makes sense, but happy to answer to more questions 🙌

waiting-for-dev:
Thanks for your answer, @elia! That makes sense. To better understand real-world usage, do you have an example you can share about how other authentication systems different than devise are used in active admin?

elia:
Yes, looking at https://www.ruby-toolbox.com/categories/rails_authentication and picking the most popular ones:
- https://doorkeeper.gitbook.io/guides/ruby-on-rails/protecting-your-resources => `before_action :doorkeeper_authorize!`
- authlogic seems to require some manual redirecting by design: https://stackoverflow.com/a/19991617
- https://github.com/thoughtbot/clearance#access-control => `before_action :require_login`
- https://github.com/sorcery/sorcery#core => `before_action :require_login`

waiting-for-dev:
Thanks, @elia. I saw that the option to configure the adapter was removed altogether in another PR. I'm a bit concerned that it will require users to patch `SolidusAdmin::BaseController` for authentication systems that don't natively integrate with Rails and that probably need some glue code (thinking of https://github.com/jeremyevans/rodauth, but probably there are others). Probably that's something we can change in the future.

elia:
@waiting-for-dev I don't think we need to optimize for every possible authentication system, should be easy to use the popular ones and possible to use anything else. 

That said I think even for `rodauth` the tune is the same:

- https://github.com/janko/rodauth-rails#current-account
- https://github.com/janko/rodauth-rails#requiring-authentication

and you can `delegate :require_account, :current_accout, to: :rodauth, prefix: true` and configure solidus_admin to use `:rodauth_require_account` as the authentication method and `:rodauth_current_account` as the current user method.

PR #5216, opened 2023-07-03
[Admin] Prepare main navigation

waiting-for-dev:
## Summary

This PR sets up the main navigation for the admin when it comes to styles,
behavior, data, and configuration. Please, see individual commits for full
details.

- Add basic state styles to the main navigation
- Move nav item markup to a template
- Set up nested main nav items for the admin
- Style :hover on admin main navigation
- Render subnav items from the admin only when hovering over the parent item
- Set up admin's main navigation routes
- Handle active admin menu items from admin main navigation
- Hide container details from the user to configure admin main navigation
- Document admin main menu configuration

## Checklist


admin/lib/solidus_admin/providers/main_nav.rb
```diff
@@ -8,20 +8,21 @@ module Providers
...
+              .with_child(key: "option_types", route: :option_types_path, position: 10)
           )
-          register("promotions", MainNavItem.new(key: "promotions", icon: "solidus_admin/megaphone-line.svg", position: 30))
-          register("stock", MainNavItem.new(key: "stock", icon: "solidus_admin/stack-line.svg", position: 40))
-          register("users", MainNavItem.new(key: "users", icon: "solidus_admin/user-line.svg", position: 50))
-          register("settings", MainNavItem.new(key: "settings", icon: "solidus_admin/settings-line.svg", position: 60))
+          register("promotions", MainNavItem.new(key: "promotions", route: :promotions_path, icon: "solidus_admin/megaphone-line.svg", position: 30))
+          register("stock", MainNavItem.new(key: "stock", route: :stock_items_path, icon: "solidus_admin/stack-line.svg", position: 40))
+          register("users", MainNavItem.new(key: "users", route: :users_path, icon: "solidus_admin/user-line.svg", position: 50))
+          register("settings", MainNavItem.new(key: "settings", route: :stores_path, icon: "solidus_admin/settings-line.svg", position: 60))
```
elia: This setup is a bit hard to grok for me, I'm also not sure to understand the advantage of a fallback system over using different helpers for each menu item. Can you expand on it?

admin/app/components/solidus_admin/sidebar/item/component.rb
```diff
@@ -49,9 +61,13 @@ def nested_items
     return unless @item.children?
 
     tag.nav(
-      class: "hidden group-hover:block"
+      class: "hidden group-hover:block group-[.active]:block"
```
elia, line 64: I'm trying to understand if this can be replaced with ` class: "#{:hidden unless active?} group-hover:block"`, since `#active?` is fully static

admin/lib/solidus_admin/configuration/main_nav.rb
```diff
@@ -0,0 +1,56 @@
...
+
+require "solidus_admin/container"
+require "solidus_admin/main_nav_item"
+
+module SolidusAdmin
+  class Configuration < Spree::Preferences::Configuration
+    # Configuration for the main navigation menu.
+    class MainNav
+      NAMESPACE = "main_nav"
+      private_constant :NAMESPACE
```
elia, line 11: This is only used in `#container_key` would make sense to just inline it?

admin/app/components/solidus_admin/sidebar/item/component.rb
```diff
@@ -10,9 +10,24 @@ def initialize(item:)
...
+  # Arbitrary Tailwind background images is not working:
+  # https://github.com/tailwindlabs/tailwindcss/discussions/6617
   erb_template <<~ERB
-    <a href="#">
-      <%= @item.title %>
-    </a>
+    <div class="
+        py-2 px-3 mb-1
+        bg-no-repeat bg-left bg-origin-content
+      " style="background-image: url('<%= image_path(@item.icon) %>')">
```
elia: You think would be possible to embed the whole remix icon font and avoid adding the SVGs one by one? 

https://remixicon.com/download

elia (commented):
Wow, this is great! 🙌 

I left a few comments here and there, mostly minor things, except for the URL fallback for which I have a strong feeling that we can go with a much simpler setup.

I was also wondering about the sidebar namespace, now we have part of the components under `Sidebar` and part under `MainNav`. If we stick with `main_nav` as the name of this section you think we can bring everything under a single namespace?

Plus, a very small change I think the Sidebar component background should be `bg-gray-100` instead of `bg-white` (it's not part of the diff so I couldn't comment inline).

admin/docs/customizing_main_navigation.md
```diff
@@ -0,0 +1,48 @@
...
+      icon: "solidus_admin/my_custom_link.svg",
+      position: 80
+    )
+  end
+end
+```
+
+- The key you provide will be used to translate the link's label under the
+`solidus_admin.main_nav.#{key}` key. 
+- Icon needs to be a path to a SVG file accessible from the asset pipeline.
```
kennyadsl, line 21: I think we need to provide and document the full icon set somehow (let's wait for the answer to the related Elia's comment) because it's unrealistic that people will build their own icon if they want to add a new item to the menu.

kennyadsl (commented):
Thanks Marc, I don't have big concerns but curious to see the answers to Elia's comments.

admin/app/components/solidus_admin/sidebar/item/component.rb
```diff
@@ -43,7 +43,9 @@ def link_level_classes
   def nested_items
     return unless @item.children?
 
-    tag.nav do
+    tag.nav(
+      class: "hidden group-hover:block"
```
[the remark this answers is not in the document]
elia: hahah sorry, I left that part out! 😅 

In my interpretation there was not expansion on hover, only on click, and if that's the case the implementation will be simpler. I think I didn't think of expanding on hover also because it might be tricky to navigate, with things moving under the pointer. 

That's why I was suggesting a check with a designer.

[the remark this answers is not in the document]
waiting-for-dev: Oh, I see. Will check :slightly_smiling_face:

rainerdema (approved):
Overall looks great to me! 👏 I don't have anything additional to add beyond what's already under discussion.
(Pending @elia's and @kennyadsl's comments)

admin/app/components/solidus_admin/sidebar/item/component.rb
```diff
@@ -10,9 +10,24 @@ def initialize(item:)
...
+        text-sm text-black
+        font-sans font-bold
+      ">
+        <%= name %>
+      </a>
+    </div>
   ERB
+
+  def name
+    t("main_nav.#{@item.key}")
```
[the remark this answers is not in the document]
waiting-for-dev: Thanks, @elia! Fixed :raised_hands: :raised_hands:

admin/app/components/solidus_admin/sidebar/item/component.rb
```diff
@@ -11,4 +11,30 @@ def initialize(item:)
...
   end
+
+  # Arbitrary Tailwind background images is not working:
+  # https://github.com/tailwindlabs/tailwindcss/discussions/6617
+  def background_image_style_attribute
+    return unless @item.icon
+
+    "style=\"background-image: url('".html_safe +
+      image_path(@item.icon) +
+      "')\"".html_safe
```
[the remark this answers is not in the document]
waiting-for-dev: Good point. However, in this case, I wanted to be picky and not mark the URL as safe, in case someone implements another kind of main menu item fetching data from the database. In any case, that's not part of the final version anymore :slightly_smiling_face:

admin/app/components/solidus_admin/sidebar/item/component.rb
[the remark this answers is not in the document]
waiting-for-dev: You were right! I updated the code accordingly.

admin/app/components/solidus_admin/sidebar/item/component.rb
waiting-for-dev: Thanks for pointing that out, it's a great improvement. Done!

admin/docs/customizing_main_navigation.md
waiting-for-dev: Done!

admin/app/components/solidus_admin/sidebar/item/component.rb
waiting-for-dev: Good call! Fixed.

admin/lib/solidus_admin/configuration/main_nav.rb
waiting-for-dev: Yeah, that's an option. If you don't feel strongly about that, I prefer to leave it as it is now. I prefer clarity over terseness for this kind of hardcoded string.

admin/lib/solidus_admin/providers/main_nav.rb
kennyadsl: Shouldn't we also consider that some routes will be available in both engines but controlled by a preference so that users may still stick to the old admin until they are ready to switch? I'm not sure what this adds to the mix here but I wanted to point it out in case it's relevant.

elia: > I want to hide our need to use two route proxies (spree & solidus_admin) until all the new admin has been done.

I thought it was better to keep it loud and clear, it's an existing and somewhat known pattern coming from rails itself, and we can refer to the original documentation instead of explaining how the proxy would work so that they can use it.

When it comes to the proxy I would propose to just evaluate it in the view_context (or the controller context) as that's the two natural environments in which urls are usually built:

```rb
register("users", MainNavItem.new(key: "users", route: -> { spree.users_path }, icon: "solidus_admin/user-line.svg", position: 50))
```

If I was a user that would look really straightforward and for us as developer we can skip a whole bunch of code.

elia: @kennyadsl we don't have an option yet for allowing users to select which sections they want to enable on the new admin (actually we barely have one section), so I think we should start with routes "hardcoded" in the proc.

The proc is flexible enough to accomodate whatever system we'll want to use for selecting the right route (a preference, cookies, a mix of both, etc.).

Whenever we'll add new sections we'll be able to also update the menu with new routes. At least until we setup a way to let the user cherry-pick the desired sections.

waiting-for-dev: After thinking about it, I agree it's better to require being explicit, as we still don't know how to implement all the stuff around the two types of routes.

Please, take a look at the last commit. Instead of providing the complete view context, I've limited it to the needed URL helper's proxies. It didn't feel well to me that a data object like `MainNavItem` had access to many things happening in the view context. In addition, this way, we don't need to worry about supporting weird stuff that could be done with the view context, and we don't need to worry about potential code injection in case somebody uses a database-backed main menu item.

waiting-for-dev:
> I was also wondering about the sidebar namespace, now we have part of the components under `Sidebar` and part under `MainNav`. If we stick with `main_nav` as the name of this section you think we can bring everything under a single namespace?

As far as I can see, everything when it comes to components is under the `Sidebar` namespace. Then, we have other stuff (used by the components) using the `MainNav` name. It would feel weird to me having the main navigation component know stuff about the logo, and logout form... I think that's pointing out that the Sidebar component has too many responsibilities. Do you think the time has come to extract the main navigation component from the sidebar one? I think that would make things more manageable. Anyway, I'd leave that as part of another PR, as this one is already large enough.

> Plus, a very small change I think the Sidebar component background should be `bg-gray-100` instead of `bg-white` (it's not part of the diff so I couldn't comment inline).

Thanks for catching that. If I'm not wrong, it's `bg-gray-15`?

screenshot-www figma com-2023 07 05-14_12_33

admin/lib/solidus_admin/configuration/main_nav.rb
elia: whatever, super-weak preference here, I thought it was a leftover 👍

admin/spec/solidus_admin/main_nav_item_spec.rb
```diff
@@ -3,6 +3,13 @@
 require "spec_helper"
 
 RSpec.describe SolidusAdmin::MainNavItem do
+  def url_helpers(solidus_admin: {}, spree: {})
+    Struct.new(:solidus_admin, :spree).new(
+      solidus_admin.any? ? Struct.new(*solidus_admin.keys).new(*solidus_admin.values) : nil,
+      spree.any? ? Struct.new(*spree.keys).new(*spree.values) : nil
+    )
```
elia, line 10: nit: you might also use a double here which might be a little more readable:

```rb
double(
  solidus_admin: double(**solidus_admin),
  spree: double(**spree),
)
```

admin/app/components/solidus_admin/sidebar/item/component.rb
```diff
@@ -4,7 +4,10 @@
 class SolidusAdmin::Sidebar::Item::Component < SolidusAdmin::BaseComponent
   with_collection_parameter :item
 
-  def initialize(item:, url_helpers: solidus_admin_with_fallbacks)
+  def initialize(
+    item:,
+    url_helpers: Struct.new(:spree, :solidus_admin).new(spree, solidus_admin)
```
elia, line 9: Not needed right now, but might make sense in the future to also give access to `main_app` and other engines. 

A nice thing I just discovered is that `Rails.application.routes.mounted_helpers` is a module that will give access to all those helpers
 https://github.com/rails/rails/blob/495cbe9f8fec01ed3f2c0128dda39ca07bad00b3/actionpack/lib/action_dispatch/routing/route_set.rb#L473-L479

For sure I'll open a PR to include it in the base component

elia (approved):
@waiting-for-dev Thanks for tackling all my concerns! 🙌 

I left a fully optional nit and a consideration for the future, none of them are blocking.

---

> As far as I can see, everything when it comes to components is under the `Sidebar` namespace. Then, we have other stuff (used by the components) using the `MainNav` name. 

You're right my bad, I confused `MainNavItem` with `Sidebar::Item::Component`. At some point we might still need to rename `Sidebar` to something else to avoid mixing the main navigation with the content sidebar that I saw on some of the design. But for now it's fine as is. 

> It would feel weird to me having the main navigation component know stuff about the logo, and logout form... I think that's pointing out that the Sidebar component has too many responsibilities. Do you think the time has come to extract the main navigation component from the sidebar one? I think that would make things more manageable. 

The main navigation sidebar and it's internal components is unlikely to be reused anywhere else and I think it's ok to split the however you like within the Sidebar namespace.

As long as there's a sound way to configure items customization will never need to touch it.

> Anyway, I'd leave that as part of another PR, as this one is already large enough.

Absolutely! Let's move on!

kennyadsl (approved):
Things are coming to life! Not blocking, but at this point we might consider (even in another PR) to enable previews for the newly introduced components.

admin/lib/solidus_admin/providers/main_nav.rb
kennyadsl: We don't have a ton of URLs yet, we always have time to change them in a separate PR that tackles exactly that problem, if any. This PR is already taking into account too many things, I think it's a wise choice to opt for this solution without anticipating too much. 👍

admin/spec/solidus_admin/main_nav_item_spec.rb
waiting-for-dev: I like it. Changed!! Thanks for the suggestion :raised_hands:

admin/app/components/solidus_admin/sidebar/item/component.rb
waiting-for-dev: Yeah, I thought about `main_app`, but as it's not already part of the available helpers and I'm not sure it's going to be needed, I preferred postponing it for the future.

> A nice thing I just discovered is that Rails.application.routes.mounted_helpers is a module that will give access to all those helpers

I spent some time looking for something like that but I didn't find it :laughing: It looks like it needs something else, though. `Rails.application.routes.mounted_helpers.solidus_admin` will raise an error on the console :thinking:

waiting-for-dev:
> Things are coming to life! Not blocking, but at this point we might consider (even in another PR) to enable previews for the newly introduced components.

Agree! I created https://github.com/solidusio/solidus/issues/5221 to tackle it along with the other missing components (and stop this fat PR :sweat: )

admin/app/components/solidus_admin/sidebar/item/component.rb
elia: Yes, because `Rails.application.routes.mounted_helpers` is a module that needs to be included in a class, that said I just tried on the BaseComponent and errors out if used before rendering (because it needs a reference to the controller). 

💯 for now we can stick to those two

PR #5217, opened 2023-07-04
Remove unused ActiveRecord join class `Spree::PromotionRuleRole`

mamhoff:
## Summary

This class never had a database table to back it. The promotion rule "User Role" uses a preferences array of role IDs to store the roles it needs.

## Checklist

### DIFF
--- a/core/app/models/spree/promotion_rule_role.rb
+++ b/core/app/models/spree/promotion_rule_role.rb
@@ -1,8 +1,0 @@
-# frozen_string_literal: true
-
-module Spree
-  class PromotionRuleRole < Spree::Base
-    belongs_to :promotion_rule, class_name: 'Spree::PromotionRule', optional: true
-    belongs_to :role, class_name: 'Spree::Role', optional: true
-  end
-end


kennyadsl:
@mamhoff Thanks, I think you are right! Do you think using the preference to store roles is enough? Or is there any reason to move to the database version, instead of removing the table?

mamhoff:
We could go either way here. I went for the simpler path: Remove the join ActiveRecord class. More complicated would be: keep the class, add the table, add a migration moving the store IDs from the preference to the table. The advantage of the table would be referential integrity at the DB level if that's something we care about. I'd say this is your call as maintainers, I'm happy to do either.

kennyadsl:
@AlessioRocco I know it's a long time, but do you remember why we took this direction?

gmacdougall (approved):
We discussed this in the the core team meeting and agree that this is useless and shouldn't be in the code base.

Thanks for the contribution Martin!

waiting-for-dev (approved):

kennyadsl (approved):
👍

PR #5219, opened 2023-07-04
ResourceController: Allow overriding routes proxy

mamhoff:
## Summary

Within Solidus, we always want to use the `spree` routes proxy as that is where all routes are defined. However, many people use the Resource controller as a parent controller in their main app, and because of this they need to override all the URL generation methods in the Resource Controller.

This allows one to specify the routes proxy in a custom controller as follows:

```rb
module MyApp
  class WidgetsController < Spree::Admin::ResourceController
  private

  def routes_proxy
    main_app
  end
end
```

Engine Developers can use this for engines, as well:

```rb
module MyEngine
  class WidgetsController < Spree::Admin::ResourceController
  private

  def routes_proxy
    my_engine
  end
end
```

## Checklist

### DIFF
--- a/backend/spec/controllers/spree/admin/resource_controller_spec.rb
+++ b/backend/spec/controllers/spree/admin/resource_controller_spec.rb
@@ -22,10 +22,12 @@ describe Spree::Admin::WidgetsController, type: :controller do
     # Database
     class CreateWidgets < ActiveRecord::Migration[5.1]
       def change
-        create_table(:widgets) do |t|
-          t.string :name
-          t.integer :position
-          t.timestamps null: false
+        unless table_exists?(:widgets)
+          create_table(:widgets) do |t|
+            t.string :name
+            t.integer :position
+            t.timestamps null: false
+          end
         end
       end
     end
@@ -288,6 +290,15 @@ describe Spree::Admin::WidgetsController, type: :controller do
 
       expect(response).to redirect_to('/admin/widgets')
       expect(flash[:error]).to eql('Product is not found')
+    end
+  end
+
+  describe "#routes_proxy" do
+    subject { controller.send(:routes_proxy) }
+
+    it "forwards to the #spree routing proxy" do
+      expect(controller).to receive(:spree).and_call_original
+      expect(subject).to be_a(ActionDispatch::Routing::RoutesProxy)
     end
   end
 end
--- a/backend/app/controllers/spree/admin/resource_controller.rb
+++ b/backend/app/controllers/spree/admin/resource_controller.rb
@@ -193,7 +193,7 @@ class Spree::Admin::ResourceController < Spree::Admin::BaseController
                   .find_by!(self.class.parent_data[:find_by] => params["#{parent_model_name}_id"])
     instance_variable_set("@#{parent_model_name}", @parent)
   rescue ActiveRecord::RecordNotFound => e
-    resource_not_found(flash_class: e.model.constantize, redirect_url: spree.polymorphic_url([:admin, parent_model_name.pluralize.to_sym]))
+    resource_not_found(flash_class: e.model.constantize, redirect_url: routes_proxy.polymorphic_url([:admin, parent_model_name.pluralize.to_sym]))
   end
 
   def parent?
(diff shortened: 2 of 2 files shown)


backend/spec/controllers/spree/admin/resource_controller_spec.rb
```diff
@@ -64,12 +82,14 @@ def check_destroy_constraints
     Object.send(:remove_const, :Widget)
```
kennyadsl, line 82: Shouldn't we also remove the `MyEngine` constant at this point?

kennyadsl (commented):
Thanks @mamhoff, excellent work as always. I left some initial questions.

waiting-for-dev (commented):
Thanks, @mamhoff! That will be helpful! I am only concerned about the chosen test strategy, where we're modifying the Dummy app routes mounting an engine that it's later removed. From Solidus's point of view, this is more of a refactor, so I don't think there's the need to change anything. At most, we can just test the new `routes_proxy` public method.

backend/spec/controllers/spree/admin/resource_controller_spec.rb
mamhoff: Done, thanks!

backend/spec/controllers/spree/admin/resource_controller_spec.rb
```diff
@@ -2,23 +2,34 @@
 
 require 'spec_helper'
 
-module Spree
+module MyEngine
```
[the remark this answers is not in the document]
mamhoff: I think the `spree` routes proxy is implicitly tested in all of Solidus' admin controllers, so I believe that's not really necessary. If you insist though I can go that route.

kennyadsl:
@mamhoff We discussed internally and we think we can drop those specs changes. The setup complexity/impact ratio is probably not worth it. As @waiting-for-dev suggested, probably we can just unit test the new method.


On a side note, I checked existing extensions and clients' code and I fund some occurrences of using the AdminResource controller without the need of this change (e.g. https://github.com/solidusio-contrib/solidus_content/blob/da2de159fa3c14c139f9f2aedff80636a5398ff8/app/controllers/solidus_content/resource_controller.rb#L4). Do you know why it's not needed there? Maybe it's just because of [this](https://github.com/solidusio-contrib/solidus_content/blob/da2de159fa3c14c139f9f2aedff80636a5398ff8/lib/solidus_content/engine.rb#L10)?

mamhoff:
> On a side note, I checked existing extensions and I found some occurrences of using the AdminResource controller without the need for this change (e.g. https://github.com/solidusio-contrib/solidus_content/blob/da2de159fa3c14c139f9f2aedff80636a5398ff8/app/controllers/solidus_content/resource_controller.rb#L4). Do you know why it's not needed there? Maybe it's just because of [this](https://github.com/solidusio-contrib/solidus_content/blob/da2de159fa3c14c139f9f2aedff80636a5398ff8/lib/solidus_content/engine.rb#L10)?

In most Solidus extensions, we create routes within the `Spree` engine. I have one that isolates the namespace to `SolidusFriendlyPromotions`, and there it just won't work. Besides, in that example, all the relevant methods (`new_object_url`, `edit_object_url` etc.) are overridden, thus sending to whatever the local resource proxy is in that controller, rather than directly addressing a resource proxy.

kennyadsl:
> Besides, in that example, all the relevant methods (new_object_url, edit_object_url etc.) are overridden, thus sending to whatever the local resource proxy is in that controller, rather than directly addressing a resource proxy.

Yep, I saw that but reading the git history, I think it's for another reason, more specific to that extension's purpose and would happen anyway.

> In most Solidus extensions, we create routes within the Spree engine. I have one that isolates the namespace to SolidusFriendlyPromotions, and there it just won't work.

Isolating the engine is legit in the Rails world, even if maybe not a common practice in the Soldius' extensions ecosystem. I think it's more than legit for us to support this use case. Thanks for adding some context. 👍

kennyadsl (approved):

waiting-for-dev (approved):

waiting-for-dev:
Thanks, @mamhoff!

PR #5220, opened 2023-07-05
mark a shipment as canceled if all the inventory units have been canceled

BenMorganIO:
## Summary

I noticed that if a multi-shipment order goes through and you cancel one of the shipments, an order recalculation will set the shipment as pending/ready. See:

https://github.com/solidusio/solidus/blob/main/core/app/models/spree/shipment.rb#L208

This is because it's only checking if the order is canceled. While this is a correct method to check if a shipment should be marked as canceled, it falls short if the it's only 1 shipment in an order that needs to be canceled out of many. If I cancel a shipment, then the shipment will be marked as shipped and that can be quite surprising for customers and admins.

Instead, let's increase the guarantee on canceling a shipment for now by checking if the order is canceled _or_ all of the inventory units for said shipment have been canceled.

In the future, I think it would be better that when a shipment is canceled, an inventory unit cancellation should be done on the inventory units it's responsible for. That would help create a guarantee for resolving the bug highlighted above. But for now, just marking a shipment as canceled when the inventory units have been canceled is progress in the right direction.

## Checklist

### DIFF
--- a/core/app/models/spree/order_cancellations.rb
+++ b/core/app/models/spree/order_cancellations.rb
@@ -118,7 +118,9 @@ class Spree::OrderCancellations
       to_a
 
     shipments.each do |shipment|
-      if shipment.inventory_units.all? { |iu| iu.shipped? || iu.canceled? }
+      if shipment.inventory_units_canceled?
+        shipment.update!(state: 'canceled', shipped_at: nil)
+      elsif shipment.inventory_units_shipped_or_canceled?
         shipment.update!(state: 'shipped', shipped_at: Time.current)
       end
     end
--- a/core/app/models/spree/shipment.rb
+++ b/core/app/models/spree/shipment.rb
@@ -200,19 +200,24 @@ module Spree
 
     # Determines the appropriate +state+ according to the following logic:
     #
-    # canceled   if order is canceled
+    # canceled   if all of the inventory units are canceled
+    # shipped    if all of the inventory units are either shipped or canceled
     # pending    unless order is complete and +order.payment_state+ is +paid+
-    # shipped    if already shipped (ie. does not change the state)
     # ready      all other cases
     def determine_state(order)
-      return 'canceled' if order.canceled?
+      return 'canceled' if inventory_units_canceled? || order.canceled?
       return 'shipped' if shipped?
       return 'pending' unless order.can_ship?
-      if can_transition_from_pending_to_ready?
-        'ready'
-      else
-        'pending'
-      end
+
+      can_transition_from_pending_to_ready? ? 'ready' : 'pending'
+    end
+
+    def inventory_units_canceled?
+      inventory_units.all?(&:canceled?)
+    end
+
+    def inventory_units_shipped_or_canceled?
+      inventory_units.all? { |iu| iu.shipped? || iu.canceled? }
     end
 
     def set_up_inventory(state, variant, _order, line_item)
--- a/core/spec/models/spree/order_cancellations_spec.rb
+++ b/core/spec/models/spree/order_cancellations_spec.rb
@@ -83,15 +83,15 @@ RSpec.describe Spree::OrderCancellations do
     end
 
     it "cancels the inventory unit" do
-      expect { subject }.to change { inventory_unit.state }.to "canceled"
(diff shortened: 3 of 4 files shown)


kennyadsl:
Thanks for working on this @BenMorganIO! I tried this locally with @MassimilianoLattanzio, who worked on something similar for one of our stores.

We think this works, but we are not sure about if it's a good thing to consider a **shipped** shipment as canceled if all of its content has been canceled. E.g. I ship an empty package by mistake, so I will cancel all the items in it. Is it correct to consider that shipment canceled, even if it has been shipped? Yes, there's always the carton that represents the shipped shipment, but maybe in this case it's not clear what actually happened looking at the order's shipment state. What do you think?

Also, I think there was another issue (already there, not related to this PR), because we are updating the shipment date of the shipment after canceling one of the items present in a shipped shipment, so in that case, the shipping date will represent the "all-items cancelation" date.

BenMorganIO:
If a shipped shipment has all of its items marked as cancelled, it should remain shipped since the shipped state is supposed to be immutable.

kennyadsl:
But I don't see this happening with your code. Maybe I'm missing something?

kennyadsl:
@BenMorganIO let me show that with a quick video (I'm using your branch):


https://github.com/solidusio/solidus/assets/167946/407c0ad6-b20b-4b28-a88c-c4c7c92e561b

BenMorganIO:
@kennyadsl I can't understand why someone would cancel all of the items in a shipment after it has been shipped. I think a return would be started if that was the case. However, I want to keep the code so that once a shipment has been shipped, it is effectively immutable and should not transition to `canceled`.

core/app/models/spree/order_cancellations.rb
```diff
@@ -118,7 +118,9 @@ def update_shipped_shipments(inventory_units)
       to_a
 
     shipments.each do |shipment|
-      if shipment.inventory_units.all? { |iu| iu.shipped? || iu.canceled? }
+      if shipment.inventory_units_canceled?
```
BenMorganIO, line 121: @kennyadsl I likely need to add a check here if the shipment is not in a shipped state.

kennyadsl:
> I can't understand why someone would cancel all of the items in a shipment after it has been shipped. I think a return would be started if that was the case.

You are right; it's hard to find a use case that resonates with this and the most realistic thing I could think of is having two items in the package (e.g., two glasses), both broken. Probably, you don't want to make a return for broken items, and canceling them would work. Do you think it's something worth taking into account?

core/app/models/spree/order_cancellations.rb
kennyadsl: I'm having a look at the state machine of the shipment and it seems like it's not allowed to move a shipment from `shipped` to `canceled` already: 
 https://github.com/solidusio/solidus/blob/df68c0c748ded6a595d28c37a9460126c6d12ecb/core/lib/spree/core/state_machines/shipment.rb#L36

Do you know if there's a specific reason not to rely on the state machine here? Same for when we set the state to `shipped`. I think we are also skipping the rest of the callbacks defined in the state machine.

BenMorganIO:
@kennyadsl I think if we keep shipped shipments retained as shipped, then it shouldn't be an issue. I can see a customer returning all items from a shipment or an admin canceling the items if the shipment was lost or something. Either way, we do have a comment in the code that the shipped state is supposed to be immutable.

core/app/models/spree/order_cancellations.rb
BenMorganIO: No, I should be able to use the state machine call. I was mimicking the original code and thought that there must have been a reason to jump the state machine. Although, we should likely use it.

For the code on line 124, should we update that code as well to utilize the state machine?

kennyadsl: Yes, I think the best path forward at this point is to make another PR that just changes `OrderCancelations` to try to use the state machine. If that works and we can't find blockers, changing this PR will be smooth. What do you think?

tvdeyen:
@solidusio/core-team now that we merged and shipped #6314 can we close this?

PR #5222, opened 2023-07-05
[Admin] `ui/table` component preview

elia:
## Summary

This PR adds a preview and specs to the table component. ~~At the same time generalizes the rendering of both header and data cells.~~ _(Moved to https://github.com/solidusio/solidus/pull/5223)_

The preview demonstrates how header and data cells support different types of content:

image

The preview is also used in the spec to ensure it's not breaking and to provide basic coverage.

_The changes in this PR were extracted from the work on batch actions for which we'll need to fit a checkbox in a header cell, but I figured was easier to review it separately._

**Props to @waiting-for-dev for pushing me to try again the extraction of common things among SolidusAdmin preview classes 🙌 .**

## Checklist

### DIFF
--- a/admin/lib/solidus_admin/engine.rb
+++ b/admin/lib/solidus_admin/engine.rb
@@ -3,6 +3,7 @@
 require "view_component"
 require "solidus_admin/container"
 require "solidus_admin/importmap_reloader"
+require "solidus_admin/preview"
 
 module SolidusAdmin
   class Engine < ::Rails::Engine
@@ -16,6 +17,14 @@ module SolidusAdmin
 
     initializer "solidus_admin.view_component" do |app|
       app.config.view_component.preview_paths << SolidusAdmin::Engine.root.join("spec/components/previews").to_s
+
+      app.config.to_prepare do
+        preview_controller_class = app.config.view_component.preview_controller.constantize
+
+        # This is needed to make the preview controller have access to the same
+        # set of helpers that are available to the Preview class.
+        preview_controller_class.include SolidusAdmin::Preview::ControllerHelper
+      end
     end
 
     initializer "solidus_admin.inflections" do
--- a/admin/lib/solidus_admin/preview.rb
+++ b/admin/lib/solidus_admin/preview.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+# This module will add all the necessary harness to a ViewComponent::Preview
+# to be able to render Solidus components.
+#
+# Adds a `current_component` helper method that will return the the component class
+# based on the preview class name. The component class name is inferred by removing
+# the `Preview` suffix from the preview class name.
+#
+# Adds a `helpers` method that will return the helpers module from SolidusAdmin::BaseController
+# making them available to the preview class (which is not a standard controller). All helpers
+# are available to the preview class via `method_missing`.
+#
+# @example
+#   class SolidusAdmin::UI::Badge::ComponentPreview < ViewComponent::Preview
+#     include SolidusAdmin::Preview
+#
+#     def default
+#       render current_component.new(name: time_ago_in_words(1.day.ago))
+#     end
+#   end
+#
+module SolidusAdmin::Preview
+  extend ActiveSupport::Concern
+
+  module ControllerHelper
+    extend ActiveSupport::Concern
+
+    included do
+      helper ActionView::Helpers
+      helper_method :current_component
+    end
+
+    private
+
+    def current_component
+      @current_component ||= begin
+        # Lookbook sets the @preview instance variable with a PreviewEntity instance, while ViewComponent uses the preview class.
+        # Lookbook's PreviewEntity has `#preview_class_name` as part of its public api.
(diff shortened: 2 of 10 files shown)


admin/lib/solidus_admin/preview.rb
```diff
@@ -0,0 +1,71 @@
...
+        preview_class_name = @preview.respond_to?(:preview_class_name) ? @preview.preview_class_name : @preview.name
+        preview_class_name.chomp("Preview").constantize
+      end
+    end
+  end
+
+  included do
+    layout "solidus_admin/preview"
+
+    delegate :helpers, to: SolidusAdmin::BaseController
```
waiting-for-dev, line 50: I wonder if it would be more correct to delegate to the `SolidusAdmin::BaseComponent` instead. What do you think? I'm not sure we'll need everything that is exposed to the base controller.

admin/lib/solidus_admin/engine.rb
```diff
@@ -16,6 +17,12 @@ class Engine < ::Rails::Engine
 
     initializer "solidus_admin.view_component" do |app|
       app.config.view_component.preview_paths << SolidusAdmin::Engine.root.join("spec/components/previews").to_s
+
+      app.config.to_prepare do
+        preview_controller_class = app.config.view_component.preview_controller.constantize
+
+        preview_controller_class.include SolidusAdmin::Preview::ControllerHelper
```
waiting-for-dev, line 24: I'm failing to understand why we need to include that module in the view component's controller. Can you help me understand?

waiting-for-dev (commented):
Thanks, @elia! That will help a lot to avoid duplication. I left some comments/questions to help my understanding.

I'd also like to ask if it would be possible to submit the two final commits in a separate PR :pray:

elia (commented):
@waiting-for-dev I hope I clarified your doubts. 

I also extracted the last two commits although it's not completely clear why you had this preference, if you can explain I'll be happy to anticipate the need right away 😊.

admin/lib/solidus_admin/preview.rb
```diff
@@ -0,0 +1,71 @@
...
+    private :helpers # hide it from preview "example" methods
+  end
+
+  def current_component
```
kennyadsl, line 54: Is there a reason not to call this method just `component`? I feel like this will be the most used interface and maybe we can save some complexity removing the `current_` prefix. 

I'm not sure about `component` either, because it matches with the `component` method name outside the preview world, but I feel like `current_` add some specificity without telling anything about why it's needed or what it does.

Thinking loud and ready to change my mind, no strong opinions here, thoughts?

admin/lib/solidus_admin/engine.rb
waiting-for-dev: > So, for instance, in order to have current_component available to both inline-renders (from within the preview class) and inside preview templates (that run in the context of the controller) we need to add the method to two different places.

I see. Thanks for the clarification. Would it be good to add a comment about that, or otherwise in the commit message?

waiting-for-dev (approved):
Thanks, @elia!

> I also extracted the last two commits although it's not completely clear why you had this preference, if you can explain I'll be happy to anticipate the need right away blush.

Sure! Sorry for not being clearer. Those commits change the product's component design. I think they're too far from the PR title, but most importantly they modify how columns can be added/removed to a component, so I think it deserves more visibility (mainly to help our future selves). I hope that makes more sense! :slightly_smiling_face:

admin/lib/solidus_admin/preview.rb
waiting-for-dev: I think that's to avoid the name collision with the [`component` helper](https://github.com/solidusio/solidus/blob/64232793a1945eb80c6083e99d3d2d59c9ef3b9e/admin/app/helpers/solidus_admin/container_helper.rb#L13), which is now available to the preview classes.

elia: 😄 I actually called it `component` initially, but soon realized what you pointed out, i.e. that we also have the `component` helper. And it's already the case that it is very useful to show components collaborating with other components for either context or to provide content.

Right now it's used to show a badge in table cells, but you can also think of a modal or a sidebar component, both needing some content to be showcased in the preview.

I picked `current_` because it seemed natural (and is very familiar) but we can switch to something like `previewed_component` or something else if you think it's worth it.

kennyadsl: I think I'd slightly prefer something like `previewed_component`, but I'm also ok keeping `current_component` with the above in mind. Up to you, and we can also change it later.

elia:
> Sure! Sorry for not being clearer. Those commits change the product's component design. I think they're too far from the PR title, but most importantly they modify how columns can be added/removed to a component, so I think it deserves more visibility (mainly to help our future selves). I hope that makes more sense! 🙂

@waiting-for-dev makes total sense, thanks for clarifying. 🙏 

I was wondering about the link with PR titles, and I'm not sure it will be retained given we're rebasing the dev-branch regularly, so except for the review/merging phase all we'll have is the context of the single commit I think.

PR #5225, opened 2023-07-06
[Admin] Add a `products#index` feature spec

elia:
## Summary

Improve our coverage and protection against regression by adding a feature spec for products#index.

Eventually we'll need feature parity with the legacy version and/or cover for newly added features. In the meantime at least we ensure that we can visit the page and products are listed.

The first commit is also added to `main` in https://github.com/solidusio/solidus/pull/5224 and is needed because there's no `Spree::Store` in the database unless we add it.

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/products/index/component.rb
+++ b/admin/app/components/solidus_admin/products/index/component.rb
@@ -8,8 +8,10 @@ class SolidusAdmin::Products::Index::Component < SolidusAdmin::BaseComponent
   # @!visibility private
 
   def image_column(product)
+    image = product.gallery.images.first or return
+
     link_to(
-      image_tag(product.gallery.images.first.url(:small), class: 'h-10 w-10 max-w-min'),
+      image_tag(image.url(:small), class: 'h-10 w-10 max-w-min'),
       spree.edit_admin_product_path(product),
     )
   end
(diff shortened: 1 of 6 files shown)


core/lib/spree/testing_support/dummy_app/routes.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
 DummyApp::Application.routes.draw do
+  mount SolidusAdmin::Engine, at: '/admin' if defined?(SolidusAdmin) && ENV['LIB_NAME'] == 'solidus_admin'
```
waiting-for-dev, line 4: I wanted to investigate passing that somehow as an [argument for the `DummyApp`](https://github.com/nebulab/solidus/blob/5f9c960982146b4489d70c781637c1bc31bf4d91/core/lib/spree/testing_support/dummy_app.rb#L35) class. That would be cleaner, as here we're having the callee knowing about the caller, but I don't know if that would be possible... :thinking:

elia: Maybe we could pass a block to `.setup` for doing engine-local stuff before the app is initialized, but that would still be a one-off case just for solidus_admin as the other engines won't require it.

I think I've found a way to adjust the routes from within the `spec_helper`.

waiting-for-dev (approved):
:rocket:

kennyadsl (approved):

PR #5227, opened 2023-07-07
[Admin] Add component preview for the sidebar item component

### LINKED ISSUE #5221: [Admin] Create preview for all components
After https://github.com/solidusio/solidus/pull/5213 has been merged, we might want to create previews for the rest of the components in the codebase.

waiting-for-dev:
## Summary

We mock the url helpers so the dummy route we use points to the current page.

At the same time, we change the item component to get the request object on initialization so we can mock the full path and set the active state from outside. However, as the actual request object is only available on render time for the component, we need to set it on a `before_render` hook unless a mocked one has been provided.

Closes #5221

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/sidebar/item/component.rb
+++ b/admin/app/components/solidus_admin/sidebar/item/component.rb
@@ -4,12 +4,17 @@
 class SolidusAdmin::Sidebar::Item::Component < SolidusAdmin::BaseComponent
   with_collection_parameter :item
 
+  # @param item [SolidusAdmin::MainNavItem
+  # @param fullpath [String] the current path
+  # @param url_helpers [#solidus_admin, #spree] context for generating paths
   def initialize(
     item:,
+    fullpath: "#",
     url_helpers: Struct.new(:spree, :solidus_admin).new(spree, solidus_admin)
   )
     @item = item
     @url_helpers = url_helpers
+    @fullpath = fullpath
   end
 
   def name
@@ -67,11 +72,11 @@ class SolidusAdmin::Sidebar::Item::Component < SolidusAdmin::BaseComponent
     tag.nav(
       class: nested_nav_active_classes
     ) do
-      render self.class.with_collection(@item.children, url_helpers: @url_helpers)
+      render self.class.with_collection(@item.children, url_helpers: @url_helpers, fullpath: @fullpath)
     end
   end
 
   def active?
-    @item.active?(@url_helpers, request.fullpath)
+    @item.active?(@url_helpers, @fullpath)
   end
 end
--- a/admin/spec/components/previews/solidus_admin/sidebar/item/component_preview.rb
+++ b/admin/spec/components/previews/solidus_admin/sidebar/item/component_preview.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require "solidus_admin/main_nav_item"
+
+# @component "sidebar/item"
+class SolidusAdmin::Sidebar::Item::ComponentPreview < ViewComponent::Preview
+  include SolidusAdmin::Preview
+
+  DUMMY_ROUTE = :foo_path
+
+  DUMMY_PATH = "#"
+
+  # @param active toggle { description: "Whether the item is active" }
+  # @param key text { description: "ID also used for i18n" }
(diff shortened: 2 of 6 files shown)


admin/spec/components/previews/solidus_admin/sidebar/item/component_preview.rb
```diff
@@ -0,0 +1,47 @@
...
+require "solidus_admin/main_nav_item"
+
+# @component "sidebar/item"
+class SolidusAdmin::Sidebar::Item::ComponentPreview < ViewComponent::Preview
+  include SolidusAdmin::Preview
+
+  DUMMY_ROUTE = :foo_path
+
+  DUMMY_PATH = "#"
+
```
kennyadsl, line 12: What about using the notes to document a bit the `url_helpers` and `request` parameters, as we did [here](https://github.com/solidusio/solidus/blob/a9bba258ef3ca2ac72d854e60927ddaff1a6dd89/admin/spec/components/previews/solidus_admin/ui/forms/checkbox/component_preview.rb#L7-L47)?

kennyadsl (approved):
Left a possible documentation improvement, but maybe I'm missing something.

admin/app/components/solidus_admin/sidebar/item/component.rb
```diff
@@ -74,4 +76,12 @@ def nested_items
...
     @item.active?(@url_helpers, request.fullpath)
   end
+
+  private
+
+  # We have the request object injected on initialization for testing purposes.
+  # However, the actual Rails request object is only available at render time.
+  def before_render
+    @request ||= request
+  end
```
elia, line 86: This smells like an indication that the component can be dumbed down a little bit. 

A good technique that yielded good results in the past for me was to only accept values and use a class method to build the component starting from a more complex object.

In this case it could look like this:

```rb
  def self.from_item(item, request, url_helpers)
    new(
      key: item.key,
      icon: item.icon,
      href: item.path(url_helpers)
      active: item.active?(url_helpers, request.fullpath)
      top_level: item.top_level,
      children: item.children.map { |child| from_item(child, request, url_helpers) }
    )
  end
```

Which would remove the need to provide `request`, `url_helpers` and all that jazz to the initializer and have a preview which is really easy to play with and demonstrate in different states.

elia (commented):
Good to see more and more stuff popping up in lookbook!

Left a suggestion that might in streamlining the component item API and preview.

admin/app/components/solidus_admin/sidebar/item/component.rb
waiting-for-dev: Thanks for raising that, @elia. I agree, there's something we can extract about how the item deals with the request and URL generation contexts. I already had it at the back of my mind, but I'd like to think more about it and, in any case, tackle it in the context of a separate PR. I want to move some of the stuff to the main navigation configuration.

For now, I revisited the stuff and removed the `before_render` "hack" by just passing around the whole path string instead of the entire request object.

Does it work for you?

admin/spec/components/previews/solidus_admin/sidebar/item/component_preview.rb
waiting-for-dev: I feel a bit uneasy documenting those parameters, which are private in nature, in the lookbook documentation. In any case, I added docs about how to build the component from a main navigation item, and also documented all the parameters (including `url_helpers` & `request`) in the component class. What do you think?

admin/app/components/solidus_admin/sidebar/item/component.rb
elia: > I already had it at the back of my mind, but I'd like to think more about it and, in any case, tackle it in the context of a separate PR. I want to move some of the stuff to the main navigation configuration.

That's awesome, you have the best knowledge about where to move the logic, and specs and previews always benefit from accepting simpler objects. Tackling it in a separate PR works great! 

> For now, I revisited the stuff and removed the `before_render` "hack" by just passing around the whole path string instead of the entire request object.

Great call 🙌

elia (approved):

kennyadsl (approved):
👍

waiting-for-dev:
Forced a push to fix a test after the change from `request` to `fullpath` as parameter.

PR #5234, opened 2023-07-11
[Admin] add a `ui/pagination` component 

rainerdema:
## Summary

We are using [Geared Pagination](https://github.com/basecamp/geared_pagination) as it is much simpler to configure and use in this context compared to [Kaminari](https://github.com/kaminari/kaminari).

Happy to discuss and evaluate this choice if needed.
Screenshot 2023-07-17 at 00 37 24
Screenshot 2023-07-17 at 00 37 16
Screenshot 2023-07-13 at 20 16 34
Screenshot 2023-07-17 at 00 30 48

---

Screenshot 2023-07-17 at 01 38 26
Screenshot 2023-07-17 at 01 52 38

## Checklist


admin/app/components/solidus_admin/ui/pagination/component.html.erb
```diff
@@ -0,0 +1,2 @@
+<div>
+</div>
```
waiting-for-dev: Do you think we can squash the commit with these changes with the fourth one (the actual implementation for the component)?

admin/app/controllers/solidus_admin/products_controller.rb
```diff
@@ -3,7 +3,10 @@
 module SolidusAdmin
   class ProductsController < SolidusAdmin::BaseController
     def index
-      @products = Spree::Product.all
+      set_page_and_extract_portion_from(
+        Spree::Product.order(created_at: :desc),
+        per_page: Spree::Config[:admin_products_per_page]
```
waiting-for-dev, line 8: Thoughts about moving this setting to `SolidusAdmin::Configuration`? We could default it to the value in `Spree::Config` as part of the process for deprecating the one in core.

admin/app/components/solidus_admin/ui/pagination/component.html.erb
```diff
@@ -1,2 +1,33 @@
 <div>
+  <nav>
+    <ul class="flex items-center -space-x-px h-10 text-base">
```
waiting-for-dev, line 3: Is the `-` before `space-x-px` a typo? :thinking:

waiting-for-dev: Also, I wonder if `text-base` makes any difference here, as we only use `icon_tag` within the container. Is it because of the line-height?

admin/app/components/solidus_admin/ui/pagination/component.rb
```diff
@@ -1,4 +1,42 @@
...
 
 class SolidusAdmin::UI::Pagination::Component < SolidusAdmin::BaseComponent
+  def initialize(page:, path: nil)
+    @page = page
+    @path = path
+  end
+
+  def path_for(page_number)
+    if @path.nil?
+      "#"
```
waiting-for-dev, line 11: Are we considering `@path` to be `nil` because of the preview? I think that maybe we can isolate that within the preview itself by creating a dummy path resolver there: `->(_page_number) { "#" }`. What do you think?

admin/app/components/solidus_admin/ui/pagination/component.rb
```diff
@@ -1,4 +1,42 @@
 # frozen_string_literal: true
 
 class SolidusAdmin::UI::Pagination::Component < SolidusAdmin::BaseComponent
+  def initialize(page:, path: nil)
```
waiting-for-dev, line 4: It took me a couple of minutes to understand that `path` is a `Proc`. What do you think if we add yard tags for the `#initialize` parameters to make it super clear?

admin/spec/components/previews/solidus_admin/ui/pagination/component_preview.rb
```diff
@@ -4,4 +4,27 @@
...
+    render current_component.new(
+      page: page_proc.call(left, right)
+    )
+  end
+
+  private
+
+  def page_proc
+    lambda { |left, right|
+      Struct.new(:records, :number, :next_param, :first?, :last?).new(nil, 1, '#', !left, !right)
```
waiting-for-dev, line 27: Probably we don't need to stub `:records`, as we're not using it in the component, but I might be missing something.

admin/app/components/solidus_admin/ui/table/component.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 class SolidusAdmin::UI::Table::Component < SolidusAdmin::BaseComponent
-  def initialize(rows:, model_class:, columns: [])
+  def initialize(rows:, model_class:, columns: [], pagination_component: nil)
```
waiting-for-dev, line 4: Coupling the table component with a page at this point is probably worth it. We have the `rows` and the `pagination_component` parameters requiring the caller to extract information from a page. I also wonder if extracting the `model_class` also from a page is possible (maybe not if the collection is empty?). Do you think we can explore that solution? It also would make the code more consistent, as we're providing component classes in the initializer but not component instances as in this case.

If we go the extra mile, we could probably split the component into two: the one coupled with a `page`, and a pure UI one. But we can wait for this one until we find its need.

waiting-for-dev (commented):
Thanks, @rainerdema! Very happy to see more horsepower in building components :horse: :muscle: I left some questions and suggestions. I'm eager to read your feedback!

waiting-for-dev:
It would be worth it to rebase on top of `nebulab/admin`, as we now have automatic a11y checks, and something could break after the merge.

admin/app/components/solidus_admin/ui/table/component.rb
rainerdema: Thank you for bringing up this point @waiting-for-dev 🙏 

>Coupling the table component with a page at this point is probably worth it. We have the rows and the pagination_component parameters requiring the caller to extract information from a page.
Do you think we can explore that solution? It also would make the code more consistent, as we're providing component classes in the initializer but not component instances as in this case.

>If we go the extra mile, we could probably split the component into two: the one coupled with a page, and a pure UI one. But we can wait for this one until we find its need.

So you're thinking about having something like this at the end, right?
```ruby
def initialize(page:, columns: [], pagination_component: component("ui/pagination"))
  .
  .
  .
end
  ```

I think in general it could be a good idea for now 🤔
We can start by coupling these logics and exploring the possibilities of a more generic table component in the future.

My main concern is how to determine the path (e.g., `product_path` in this example) that I'm currently passing as a proc to the pagination component.

We can consider the following solution:
1. Create a separate method in the table component to handle generating the path based on the provided proc.
2. Use a specific approach to generate the path. For example, if the table component is used for displaying `Spree::Product`s, we can assume that the path helper would follow the pattern `product_path`. This way, we can automatically generate the path without explicitly passing it as a proc.
3. Allow the flexibility to override the path generation logic by accepting a lambda as an optional argument to the table component. This way, users can provide their own path generation logic if needed.

The last point raises my main concern that by simplifying the configuration too much, we risk creating a less reusable table component since it is a UI component.


> I also wonder if extracting the model_class also from a page is possible (maybe not if the collection is empty?).

Yeah, I have to explore this point better, but I think it's possible 👍 

Let me know WDYT 🙌

Comment (waiting-for-dev) [reply admin/app/components/solidus_admin/ui/table/component.rb:4]: > So you're thinking about having something like this at the end, right?
> 
> ```ruby
> def initialize(page:, columns: [], pagination_component: component("ui/pagination"))
>   .
>   .
>   .
> end
> ```

Yeah! That would be great :slightly_smiling_face:

> * Create a separate method in the table component to handle generating the path based on the provided proc.
> * Use a specific approach to generate the path. For example, if the table component is used for displaying `Spree::Product`s, we can assume that the path helper would follow the pattern `product_path`. This way, we can automatically generate the path without explicitly passing it as a proc.
> * Allow the flexibility to override the path generation logic by accepting a lambda as an optional argument to the table component. This way, users can provide their own path generation logic if needed.

I think I'd go with the third option, dispatching the argument from the table to the pagination component. As you said, we could add some conventions to make the discovery automatic, but I won't be bothered if we need to pass the proc every time :+1:

> The last point raises my main concern: by simplifying the configuration too much, we risk creating a less reusable table component since it is a UI component.

I completely agree. We're now making explicit the coupling we already had (having three parameters that were deducible from a page). A nice next step would be creating a pure table UI component, without knowledge of a page or ActiveRecord. Then the current one should compose the UI one.

**Comment by rainerdema** [comment]
ℹ️  Specs are failing because of: https://github.com/solidusio/solidus/pull/5251

File: admin/spec/components/previews/solidus_admin/ui/table/component_preview.rb [hunk 8]
```diff
@@ -17,11 +17,17 @@ def simple
       model_class.new(id: n, name: "Product #{n}", price: n * 10.0, available_on: n.days.ago)
     end
 
+    page = Struct.new(:records, :number, :next_param, :first?, :last?).new(rows, 1, '#', true, false)
+
+    page.records.define_singleton_method(:model) do
+      model_class
+    end
```
waiting-for-dev, line 24: If there were a way to create a collection proxy from the above-generated array, we wouldn't need to do that. I didn't find a way, but leaving here the comment in case someone has a better idea :slightly_smiling_face:

waiting-for-dev (approved):
Thanks, @rainerdema! Great job! :clap: :clap: 

Left two non-blocking comments.

admin/spec/components/previews/solidus_admin/ui/table/component_preview.rb
rainerdema: Yeah, I don't like it either. I've also tried to come up with some ideas to simplify this part, but unfortunately, I couldn't find a clearer solution. 😕

admin/app/components/solidus_admin/ui/table/component.rb
rainerdema: > A nice next step would be creating a pure table UI component, without knowledge of a page or ActiveRecord. Then the current one should compose the UI one.

Definitely, that will be the next step. I don't want to continue these changes in this PR to avoid deviating too much from the integration purpose. However, once we have the need, we should definitely proceed with something like that.

PR #5180, opened 2023-06-26
[Admin] add batch actions to the `products/index` component

elia:
## Summary

- based on #5259 

Add support for batch actions to the `ui/table` component and connects it to some example actions in the `products/index` component.

- the list of batch actions for products is tentative, but adding multiple batch actions was instrumental to testing the harness
- selecting beyond the current page is left for future development because it needs to involve filters and search

https://github.com/solidusio/solidus/assets/1051/8118c5bc-b404-4732-967a-cc4517fa23b6

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/.github/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [ ] I have written a thorough PR description.
- [ ] I have kept my commits small and atomic.
- [ ] [I have used clear, explanatory commit messages](https://github.com/solidusio/.github/blob/main/CONTRIBUTING.md#writing-good-commit-messages).

The following are not always needed:

- 📖 I have updated the README to account for my changes.
- 📑 I have documented new code [with YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md).
- 🛣️ I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).
- ✅ I have added automated tests to cover my changes.
- 📸 I have attached screenshots to demo visual changes.

PR #5237, opened 2023-07-11
[Admin] `ui/tab` component

elia:
## Summary

image

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/ui/tab/component.rb
+++ b/admin/app/components/solidus_admin/ui/tab/component.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+class SolidusAdmin::UI::Tab::Component < SolidusAdmin::BaseComponent
+  SIZES = {
+    s: %w[h-7 px-1.5 body-small-bold],
+    m: %w[h-9 px-3 body-small-bold],
+    l: %w[h-12 px-4 body-text-bold],
+  }
+
+  TAG_NAMES = {
+    a: :a,
+    button: :button,
+  }
+
+  def initialize(text:, size: :m, tag: :a, disabled: false, active: false, **attributes)
+    @tag = tag
+    @text = text
+    @size = size
+    @active = active
+    @disabled = disabled
+    @attributes = attributes
+  end
+
+  def call
+    class_name = [
+      @attributes.delete(:class),
+      SIZES.fetch(@size.to_sym),
+      %w[
+        rounded justify-start items-center inline-flex py-1.5 cursor-pointer
+        bg-transparent text-gray-500
+
+        hover:bg-gray-75 hover:text-gray-700
+        focus:bg-gray-25 focus:text-gray-700
+
+        active:bg-gray-50 active:text-black
+        data-[ui-active]:bg-gray-50 data-[ui-active]:text-black
+
+        disabled:bg-gray-100 disabled:text-gray-400
+        data-[ui-disabled]:bg-gray-100 data-[ui-disabled]:text-gray-400
+      ]
+    ].join(" ")
+
+    @attributes["data-ui-active"] = true if @active
+    @attributes["data-ui-disabled"] = true if @disabled
+    @attributes[:disabled] = true if @disabled && @tag == :button
+
+    content_tag(
+      TAG_NAMES.fetch(@tag.to_sym),
+      @text,
+      class: class_name,
+      **@attributes
+    )
+  end
+end
--- a/admin/spec/components/previews/solidus_admin/ui/tab/component_preview.rb
+++ b/admin/spec/components/previews/solidus_admin/ui/tab/component_preview.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+# @component "ui/tab"
+class SolidusAdmin::UI::Tab::ComponentPreview < ViewComponent::Preview
+  include SolidusAdmin::Preview
+
+  # @param text text
+  def overview(text: "text")
+    render_with_template locals: { text: text }
+  end
+
+  # @param text text
+  # @param size select { choices: [s, m, l] }
+  # @param active toggle
+  # @param disabled toggle
+  def playground(text: "Tab", size: :m, active: false, disabled: false)
+    render current_component.new(text: text, size: size, active: active, disabled: disabled)
+  end
+end
--- a/admin/spec/components/previews/solidus_admin/ui/tab/component_preview/overview.html.erb
+++ b/admin/spec/components/previews/solidus_admin/ui/tab/component_preview/overview.html.erb
@@ -1,0 +1,18 @@
+<table>
+  <tr>
+    <td></td>
+    <% current_component::SIZES.keys.each do |size| %>
+      <td class="px-3 py-1 text-gray-500 text-center text-[16px]"><%= size.to_s.humanize %></td>
+    <% end %>
+  </tr>
+  <% %i[default active disabled].each do |state| %>
+    <tr>
+      <td class="font-bold px-3 py-1"><%= state.to_s.humanize %></td>
+      <% current_component::SIZES.keys.each do |size| %>
+        <td class="px-3 py-1 text-center">
+          <%= render current_component.new(size: size, text: text, disabled: state == :disabled, active: state == :active) %>
+        </td>
+      <% end %>
+    </tr>
+  <% end %>
+</table>
--- a/admin/spec/components/solidus_admin/ui/tab/component_spec.rb
+++ b/admin/spec/components/solidus_admin/ui/tab/component_spec.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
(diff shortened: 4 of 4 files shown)


admin/spec/components/previews/solidus_admin/ui/tab/component_preview.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# @component "ui/tab"
+class SolidusAdmin::UI::Tab::ComponentPreview < ViewComponent::Preview
+  include SolidusAdmin::Preview
+
+  def overview(text: "text")
```
waiting-for-dev, line 7: Is it missing the `@param` definition? :thinking:

admin/spec/components/previews/solidus_admin/ui/tab/component_preview.rb
```diff
@@ -0,0 +1,18 @@
...
+    render_with_template locals: { text: text }
+  end
+
+  # @param text text
+  # @param size select { choices: [s, m, l] }
+  # @param state select { choices: [default, active, disabled] }
```
waiting-for-dev, line 13: What about using a `toggle` type? Given that `default` & `active` mean the same, and we indeed act as if the param was a boolean (we don't consider the value can be `:default`). We already used this type in the item component preview: https://github.com/solidusio/solidus/blob/a81065132489f025ff9c42e4228e3aecaffc766e/admin/spec/components/previews/solidus_admin/sidebar/item/component_preview.rb#L13

admin/app/components/solidus_admin/ui/tab/component.rb
```diff
@@ -0,0 +1,54 @@
...
+    m: %w[h-9 px-3 body-small-bold],
+    l: %w[h-12 px-4 body-text-bold],
+  }
+
+  TAG_NAMES = {
+    a: :a,
+    button: :button,
+  }
+
+  def initialize(text:, size: :m, tag: :a, disabled: nil, active: nil, **attributes)
```
waiting-for-dev, line 15:
```suggestion
  def initialize(text:, size: :m, tag: :a, disabled: false, active: false, **attributes)
```

What do you think about :arrow_up: ? They're boolean parameters in spirit :slightly_smiling_face:

admin/app/components/solidus_admin/ui/tab/component.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+class SolidusAdmin::UI::Tab::Component < SolidusAdmin::BaseComponent
+  SIZES = {
+    s: %w[h-7 px-1.5 body-small-bold],
+    m: %w[h-9 px-3 body-small-bold],
+    l: %w[h-12 px-4 body-text-bold],
```
waiting-for-dev, line 7: Non-blocking nit: should we move `body-small-bold` to the list of common classes? Or do you foresee other sizes where we'll change the typo?

waiting-for-dev (commented):
Thanks, @elia! We already have a good amount of UI components :clap: :clap: 

I left some comments, but it looks great.

admin/app/components/solidus_admin/ui/tab/component.rb
elia: Large (`l`) uses `body-text-small`, while the other two use `body-small-bold`, so I think it's easier to keep them here.

admin/app/components/solidus_admin/ui/tab/component.rb
elia: Done, thanks!

admin/app/components/solidus_admin/ui/tab/component.rb
waiting-for-dev: Oh, I overlooked that :+1:

admin/app/components/solidus_admin/ui/tab/component.rb
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  TAG_NAMES = {
+    a: :a,
+    button: :button,
+  }
+
+  def initialize(text:, size: :m, tag: :a, disabled: false, active: false, **attributes)
```
waiting-for-dev, line 15: I overlooked that in the first round, but aren't `disabled` & `active` mutually exclusive? Do you think we could reflect that better by having a `state` "enum" parameter as input with `:default`, `:active` or `:disabled` as possible values?

elia: I started out with a state, but then moved to independent args, mainly because I noticed that for a button `disabled` would be an attribute and I would expect to be able to pass `disabled: true` like in other components.

I'm clearly conflicted in the decision, but in the current state `disabled` takes precedence over `active` which seems correct.

Thoughts?

kennyadsl: I didn't think about that, but maybe we should not see a tab as a button and couple our UI element's behavior with the HTML element. No strong feelings here, but I think that we should change the interface of the component only because one of its elements might expect a certain pattern.

waiting-for-dev: Yeah, I also think that we should build the interface that makes sense for this given component :+1:

rainerdema (approved):
Great job 👏

admin/app/components/solidus_admin/ui/tab/component.rb
elia: I'm not sure what you are suggesting (sorry, it's Friday! 😅).

I think both options make sense, and passing `disabled:` will be common to every other `input` element.

If you want to see the component in its usage context here it is:

```erb
   <div class="<%= toolbar_classes %>" data-<%= stimulus_id %>-target="scopesToolbar">
     <%= render component('ui/tab').new(text: "All", active: true, scheme: :secondary, href: "") %>
   </div>
```

https://github.com/solidusio/solidus/pull/5180/files#diff-b52bed4cf337144516f6db24019235696a9e901ab2cd0f96ef06fbd06560e4a2R12-R14

kennyadsl: I see the tab more as a navigational element than a for object, so I think having the `a` element is enough to cover all the cases. But until we see it in action we don't know yet and it might be also useful as a `button` sometimes. Up to you, I'm also ok with the current version.

kennyadsl (approved):

admin/app/components/solidus_admin/ui/tab/component.rb
waiting-for-dev: I understand your point about the `button` tag. What I find annoying is something like:

```ruby
component('ui/tab').new(text: "All", active: true, disabled: true)
```

The method signature does not define it, and I need to read the implementation to understand the behavior. That's because, in nature, we're dealing with an enum type.

@elia, do we have somewhere in the design files where a tab acts like a button? That could help to make the decision. If we have one, or we foresee we'll have one, would it be better to create another button component? In the end, matching styles should be seen as a coincidence, but a I think a component should wrap the UI or domain nature of the elements, not its style. Thoughts?

elia: @waiting-for-dev 

My basic thought about this is that I don't consider any of this final, and I would spend as little as possible guessing vs. sourcing the best API we can from actual usage in the application. Right now we only have one use case for this component which is inside `ui/table` for the scopes. 

The only reason I went for separate fields is that is reasonable to expect `disabled` to be an independent attribute which would also be reflected in the DOM if used with a button element.

Since I don't know yet what kind of interaction we'll have between that "scopes bar" and filters/searches/sorting on the table I can't predict if it's gonna be one or the other. So I quickly did what kind of made sense delaying a refinement of the API to when we'll have more information. I'm certain that using a button element is possible as the filter/search/sorting system _can_ end up using an underlying form, but again, I'm not sure yet.

I also already pointed out that when both `active` and `disabled` are set te component appears as disabled. To me the current situation is good enough as a basis for further refinement down the road and I would move along as fast as possible.

The other option, also popular among teams building components, is to be more strict and not extract anything to the "UI" library until it's used at least in two different components. That would include `ui/table` become `products/index/table` and so on, until we have another usage place. I would be ok with that as long as we're able to swiftly rename components.

_Sorry for the wall of text, I how this gives you a better picture of how I came up with this code and why I'm resisting a conversation on which API would be "better". 🙏_

waiting-for-dev: Fair enough, @elia. Let's go ahead and move forward. In my experience, these refinements left for later are not often reviewed later on. Because of that, I always prefer to build an API that makes sense for the current code version, as it's always easier to revisit when driven by an actual need to extend. But you're more familiar with these components as part of your work with batch actions, so let's merge this!

waiting-for-dev (approved):

admin/app/components/solidus_admin/ui/tab/component.rb
elia: Thanks @waiting-for-dev, really appreciate ❤️

PR #5240, opened 2023-07-12
[Admin] Add account navigation component

waiting-for-dev:
## Summary

It's not still fully functional, as authentication is not yet in place. For now, we use dummy data to show the component.

We extract the helper to render SVG remix icons as we need to reuse it.

We're also changing the layout so the main element occupies the full height of the viewport. That fixes an issue where the sidebar wasn't filling it due to being in a shorter container.

screenshot-localhost_3000-2023 07 12-12_07_42

[screencast-localhost_3000-2023.07.12-12_06_52.webm](https://github.com/solidusio/solidus/assets/52650/90f8cd07-753c-44d0-8d38-17df7f25b82d)

Closes #5106 

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/.github/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [x] I have written a thorough PR description.
- [x] I have kept my commits small and atomic.
- [x] [I have used clear, explanatory commit messages](https://github.com/solidusio/.github/blob/main/CONTRIBUTING.md#writing-good-commit-messages).

The following are not always needed:

- 📖 I have updated the README to account for my changes.
- 📑 I have documented new code [with YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md).
- 🛣️ I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).
- ✅ I have added automated tests to cover my changes.
- 📸 I have attached screenshots to demo visual changes.

PR #5241, opened 2023-07-12
[Admin] Extract sidebar item's icon to a proper helper 

kennyadsl:
## Summary

We will use this `icon_tag` a lot, so it might make sense to extract it into a common place.

This PR also improves the sidebar's item fill color to use the [`fill-current`](https://v2.tailwindcss.com/docs/fill#usage) property, which is now the default. 

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/base_component.rb
+++ b/admin/app/components/solidus_admin/base_component.rb
@@ -5,6 +5,23 @@ module SolidusAdmin
   class BaseComponent < ViewComponent::Base
     include SolidusAdmin::ContainerHelper
 
+    # Renders a remixincon svg.
+    #
+    # @param name [String] the name of the icon
+    # @option attrs [String] :class the class to add to the svg
+    # @return [String] the svg tag
+    # @see https://remixicon.com/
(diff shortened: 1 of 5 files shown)


waiting-for-dev:
Thanks, @kennyadsl! I did something similar in #5240, but I agree having a separate PR is better. A couple of considerations when comparing with the other helper:

- Do you think it's worth having `inline-block align-text-bottom` as default classes?
- We can copy the unit test I added for the helper.
- We can also add some yarddocs to the method.

:raised_hands:

elia (approved):
👍👍👍 

> Do you think it's worth having `inline-block align-text-bottom` as default classes?

@waiting-for-dev I'd say that _in this case_ if we have default classes we should also have a way to completely override them as they're both about positioning, which brings me to think that we could start without and add a default if we notice that we keep adding the same classes when we use `icon_tag`.

kennyadsl:
Done everything except the `inline-block align-text-bottom` as default. I'd stick to @elia's suggestion and wait to see if that's really something we use all the time. It will be easy to change it even after we have many occurrences of the `icon_tag`.

PR #5247, opened 2023-07-13
[Admin] Automatically test a11y in the new admin

### LINKED ISSUE #5132: [Admin] Use an accessibility linter for the new admin
We want to build a new admin which is accessible to anyone. We can perform a simple task toward that goal using an accessibility linter.

In quick research, [axe](https://github.com/dequelabs/axe-core-gems) looks like a good candidate. We would probably need both `axe-core-capybara` & `axe-core-rspec`. However, it could be possible to find a better tool, so let's also do a bit more of investigation.

waiting-for-dev:
## Summary

We use the [axe set of gems](https://github.com/dequelabs/axe-core-gems) to test for automatic testing of accessibility issues in the new admin. This commit adds the necessary gems and configuration to the admin Gemfile and spec_helper.rb files. It also adds a test to the products listing.

As accessibility testing requires javascript, which makes tests slow, we'll need to be selective about where we add these tests.

We're also fixing some accessibility issues we had so far. Please, check individual commits for more details.

Closes #5132

- Adds HTML titles to admin pages
- Fix accessibility for empty column headers in admin tables
- Fix HTML tag not specifying language in the admin
- Fix nested main menu items wrapped within "nav" instead of "ul"
- Add h1 to products index page

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/sidebar/item/component.rb
+++ b/admin/app/components/solidus_admin/sidebar/item/component.rb
@@ -61,7 +61,7 @@ class SolidusAdmin::Sidebar::Item::Component < SolidusAdmin::BaseComponent
   def nested_items
     return unless @item.children?
 
-    tag.nav(
+    tag.ul(
       class: nested_nav_active_classes
     ) do
       render self.class.with_collection(@item.children, url_helpers: @url_helpers, fullpath: @fullpath)
--- a/admin/app/components/solidus_admin/ui/table/component.html.erb
+++ b/admin/app/components/solidus_admin/ui/table/component.html.erb
@@ -19,19 +19,7 @@
     <thead class="bg-gray-15 color-gray-100">
       <tr>
         <% @columns.each do |column| %>
-          <th
-            class="
-              border-b
-              border-gray-100
-              py-3
-              px-4
-              text-[#4f4f4f]
-              text-left
-              text-3.5
-              font-[600]
-              line-[120%]
-            "
-          ><%= render_header_cell(column.header) %></th>
+          <%= render_header_cell(column.header) %>
         <% end %>
       </tr>
     </thead>
@@ -40,7 +28,7 @@
       <% @rows.each do |row| %>
         <tr class="border-b border-gray-100">
           <% @columns.each do |column| %>
-            <td class="py-2 px-4 "><%= render_data_cell(column.data, row) %></td>
+            <%= render_data_cell(column.data, row) %>
           <% end %>
         </tr>
       <% end %>
--- a/admin/app/helpers/solidus_admin/layout_helper.rb
+++ b/admin/app/helpers/solidus_admin/layout_helper.rb
@@ -3,6 +3,11 @@
 module SolidusAdmin
   # Helpers for the admin layout
   module LayoutHelper
+    # @return [Symbol]
+    def current_locale(backend: I18n)
+      backend.locale
+    end
+
     # @param store_name [String]
     # @param controller_name [String]
     # @return [String] HTML title
--- a/admin/app/helpers/solidus_admin/layout_helper.rb
+++ b/admin/app/helpers/solidus_admin/layout_helper.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module SolidusAdmin
+  # Helpers for the admin layout
+  module LayoutHelper
+    # @param store_name [String]
+    # @param controller_name [String]
+    # @return [String] HTML title
+    def solidus_admin_title(store_name: current_store.name, controller_name: controller.controller_name)
+      "#{store_name} - #{t("solidus_admin.#{controller_name}")}"
+    end
+  end
+end
--- a/admin/spec/features/products_spec.rb
+++ b/admin/spec/features/products_spec.rb
@@ -3,12 +3,13 @@
 require 'spec_helper'
 
 describe "Products", type: :feature do
-  it "lists products" do
+  it "lists products", :js do
     create(:product, name: "Just a product", price: 19.99)
 
     visit "/admin/products"
 
     expect(page).to have_content("Just a product")
     expect(page).to have_content("$19.99")
+    expect(page).to be_axe_clean
   end
 end
--- a/admin/spec/helpers/solidus_admin/layout_helper_spec.rb
+++ b/admin/spec/helpers/solidus_admin/layout_helper_spec.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+RSpec.describe SolidusAdmin::LayoutHelper, :helper do
+  describe '#solidus_admin_title' do
+    it "includes the store name" do
+      expect(
+        helper.solidus_admin_title(store_name: "My Store")
+      ).to include("My Store")
+    end
+
+    it "includes the translated controller base name" do
+      expect(
+        helper.solidus_admin_title(store_name: "My Store", controller_name: "orders")
+      ).to include("Orders")
+    end
+  end
+end
--- a/admin/spec/helpers/solidus_admin/layout_helper_spec.rb
+++ b/admin/spec/helpers/solidus_admin/layout_helper_spec.rb
@@ -3,6 +3,14 @@
 require "spec_helper"
 
 RSpec.describe SolidusAdmin::LayoutHelper, :helper do
+  describe '#current_locale' do
+    it "returns the current locale" do
+      expect(
+        helper.current_locale(backend: double(locale: :en))
+      ).to eq(:en)
+    end
+  end
+
   describe '#solidus_admin_title' do
     it "includes the store name" do
       expect(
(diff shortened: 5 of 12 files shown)


admin/app/helpers/solidus_admin/layout_helper.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module SolidusAdmin
+  # Helpers for the admin layout
+  module LayoutHelper
+    # @param store_name [String]
+    # @param controller_name [String]
+    # @return [String] HTML title
+    def solidus_admin_title(store_name: current_store.name, controller_name: controller.controller_name)
+      "#{store_name} - #{t("solidus_admin.#{controller_name}")}"
```
elia, line 15: Having the title scoped to the controller is great, by using a relative key `t('.title')` rails will check for an action level translation first, giving even more granularity in case it's needed:

- `"en.solidus_admin.#{controller_name}.#{action_name}.title"`
- `"en.solidus_admin.#{controller_name}.title"`

See https://api.rubyonrails.org/classes/ActionView/Helpers/TranslationHelper.html#method-i-translate

admin/app/components/solidus_admin/ui/table/component.html.erb
```diff
@@ -31,7 +31,7 @@
               font-[600]
               line-[120%]
             "
-          ><%= render_header_cell(column.header) %></th>
+          ><%= render_header_cell(column.header) %></<%= header_cell_tag(column) %>>
```
elia, line 34: Great catch!

If we move the whole tag generation inside the `render_header_cell` helper we can make it work even when a header will return blank content through a block or a method call (Symbol).

I have a preference for seeing all the classes in the template file, but in this case would be fine to move them to the class.

admin/spec/helpers/solidus_admin/layout_helper_spec.rb
```diff
@@ -3,6 +3,14 @@
 require "spec_helper"
 
 RSpec.describe SolidusAdmin::LayoutHelper, :helper do
+  describe '#current_locale' do
+    it "returns the current locale" do
+      expect(
+        helper.current_locale(backend: double(locale: :en))
```
elia, line 9: With a simple `double` RSpec won't apply spies to check if the original method exist, that would make it more robust. 

To me would also be fine if we stubbed directly on `I18n`, is there any utility in injecting the constant other than testing?

admin/app/components/solidus_admin/sidebar/item/component.rb
```diff
@@ -64,7 +64,7 @@ def link_level_classes
   def nested_items
     return unless @item.children?
 
-    tag.nav(
+    tag.ul(
```
elia, line 67: I guess every `li` needs its `ul`! ☺️

admin/spec/features/products_spec.rb
```diff
@@ -3,12 +3,13 @@
...
 describe "Products", type: :feature do
-  it "lists products" do
+  it "lists products", :js do
     create(:product, name: "Just a product", price: 19.99)
 
     visit "/admin/products"
 
     expect(page).to have_content("Just a product")
     expect(page).to have_content("$19.99")
+    expect(page).to be_axe_clean
```
elia, line 13: 🙌

elia (commented):
Just a few comments, awesome to have this check right from the beginning! 🚀

admin/app/helpers/solidus_admin/layout_helper.rb
waiting-for-dev: I'm afraid that doesn't work in our context, as the method is called from within a helper (at least, I didn't manage to make it work). In any case, I only wanted to leave an acceptable starting point and make the a11y happy. We probably need to define how we want our HTML titles to look, maybe in the context of breadcrumbs or not :slightly_smiling_face:

admin/app/components/solidus_admin/ui/table/component.html.erb
waiting-for-dev: Thanks for your suggestion! I didn't want to refactor the work for the table component further for such a small thing, as we have both the `render_header_cell` & `render_data_cell` going through a common `render_cell` method that doesn't deal with HTML tags. But I'm entirely ok with changing it now. Please, can you take a look at the last commit I pushed? I will squash it with the relevant previous commit if you like it. Otherwise, I can remove it and leave it as the initial PR.

admin/app/helpers/solidus_admin/layout_helper.rb
elia: Oh, that's right, it probably works only if generated in the controller or in the (non-layout) template.
This is fine for now

admin/spec/helpers/solidus_admin/layout_helper_spec.rb
waiting-for-dev: I'm not worried about the test not checking the original interface, which will be fully covered at the integration layer, and that's not something the unit level should worry about. I like this pattern where we limit hidden pre-conditions for a method to its parameter defaults. However, if you feel strongly about it, I won't die because of doing it the other way :slightly_smiling_face:

admin/app/components/solidus_admin/sidebar/item/component.rb
waiting-for-dev: `li` :heart: `ul`

:slightly_smiling_face:

kennyadsl (approved):
Nice! Pending conversations with @elia, but approving since nothing seems blocking to me.

admin/spec/helpers/solidus_admin/layout_helper_spec.rb
elia: I don't feel strongly at all, but I'm interested un understanding it more deeply, please stay alive and keep it ❤️ 

I guess the things that confuse me the most are: 
- what would be the use-case for passing a different thing as the i18n backend? For example it's fairly easy to `allow(I18n).to receive(:locale).and_respond(:en)` without all this harness.
- what are the hidden preconditions in this context? 
- why the unit level shouldn't worry that its default value meet its expectations? 


_Thanks in advance for your patience!_

admin/app/components/solidus_admin/ui/table/component.html.erb
elia: I had a look and it's way better, I think there's still something that can be improved on the parity between `render_header_cell` and `render_data_cell`: 
- they should both render their `td` or `th` tag
- `render_header_cell` lost the ability to render components, previously done by `render_cell`
- you can leverage `content_tag` and pass the tag from outside

```rb
def render_header_cell(cell)
  cell =
    case cell
    when Symbol
      @model_class.human_attribute_name(cell)
    when Proc
      cell.call
    end
  cell = cell.render_in(self) if cell.respond_to?(:render_in)
  tag_name = cell.blank? ? :td : :th

  content_tag(tag_name, cell, class: %{...})
end
```

The same pattern (rendering the `td` tag) should be applied to data cells, so they look even ⚖️

elia (commented):
Thanks!

I left one more request for the cell rendering methods, but they look great already and that's the right direction 🙌

rainerdema (approved):
(Pending @elia's comments)

Excellent work! 👏

admin/app/components/solidus_admin/ui/table/component.html.erb
waiting-for-dev: Oh, all right! Please, see my last commit and I'll squash everything together after your :heavy_check_mark:

admin/spec/helpers/solidus_admin/layout_helper_spec.rb
waiting-for-dev: > please stay alive and keep it

I'll [stay alive](https://www.youtube.com/watch?v=fNFzfwLM72c) anyway! :laughing:

As with so many things, that's just opinionated stuff:

> what would be the use-case for passing a different thing as the i18n backend? For example it's fairly easy to allow(I18n).to receive(:locale).and_respond(:en) without all this harness

In this case, it's only useful for testing purposes. I'm not fond of stubbing, and I try to avoid `allow... to receive...` and leave it for some integration cases. In my view, it's too tied to the method implementation, and I prefer providing external interfaces.

> what are the hidden preconditions in this context?

The hidden precondition is the I18n module. The method implementation expects it to be there and to have been configured. Of course, Ruby is not pure, but we can limit those pre-conditions to the method signature. It helps me reason about methods.

> why the unit level shouldn't worry that its default value meet its expectations?

In my view, it should worry about the interface that its collaborator implements, but not about its nature.

Again, I think there can be different views about this stuff and none is the right or wrong one :slightly_smiling_face:

elia: Thanks, this helps me a lot understanding the reasoning behind this style 👍👍👍

elia (approved):
🙌

PR #5249, opened 2023-07-14
Reorganize simple stock coordinator for improve debugging

BenMorganIO:
## Summary

This is all about making it easier for a developer to debug issues when they get the dreaded

```
     Failure/Error: order.next!

     StateMachines::InvalidTransition:
       Cannot transition state via :next from :address (Reason(s): We are unable to calculate shipping rates for the selected items.)
```

The first step we wanted to take was to make the instance variables accessible to the developer. A developer's journey to this issue is likely going to be discovering `Spree::Order#create_proposed_shipments` is what's failing. Then discovering that it builds an instance of this class. After that, they're going to want to start looking into the data that is set. Since they're all instance variables, they would likely want to be able to access some of the data.

However, what the developer will soon realize is that it's the shipping rates that are holding back the shipments from being validated. So the real issue likely lies in an empty array being returned from:

```ruby
shipment.shipping_rates = Spree::Config.stock.estimator_class.new.shipping_rates(package)
```

If this is the issue, we'll want them to be able to easily recreate the packages the same way the class builds the packages. We've reorchestrated how the packages are built into a `build_packages` method and stored it there. I've also taken the time to simplify what `build_shipments` does so they can utilize that code as well.

One other issue I noticed was that there was a global `Spree::StockLocation.all`. In a normal store, there will likely not be many, but in a marketplace, there could be thousands - so let's give them a method to easily make that code more intelligent if needed.

I hope this work also makes it easier for the developer to extend this class.

## Checklist

### DIFF
--- a/core/app/models/spree/stock/simple_coordinator.rb
+++ b/core/app/models/spree/stock/simple_coordinator.rb
@@ -22,39 +22,65 @@ module Spree
     class SimpleCoordinator
       attr_reader :order
 
+      # @api private
+      attr_reader :inventory_units, :splitters, :stock_locations,
+        :filtered_stock_locations, :inventory_units_by_variant, :desired,
+        :availability, :allocator, :packages
+
       def initialize(order, inventory_units = nil)
         @order = order
         @inventory_units =
           inventory_units || Spree::Config.stock.inventory_unit_builder_class.new(order).units
         @splitters = Spree::Config.environment.stock_splitters
 
-        filtered_stock_locations = Spree::Config.stock.location_filter_class.new(Spree::StockLocation.all, @order).filter
+        @filtered_stock_locations = Spree::Config.stock.location_filter_class.new(load_stock_locations, order).filter
         sorted_stock_locations = Spree::Config.stock.location_sorter_class.new(filtered_stock_locations).sort
         @stock_locations = sorted_stock_locations
 
         @inventory_units_by_variant = @inventory_units.group_by(&:variant)
-        @desired = Spree::StockQuantities.new(@inventory_units_by_variant.transform_values(&:count))
+        @desired = Spree::StockQuantities.new(inventory_units_by_variant.transform_values(&:count))
         @availability = Spree::Stock::Availability.new(
-          variants: @desired.variants,
-          stock_locations: @stock_locations
+          variants: desired.variants,
+          stock_locations: stock_locations
         )
 
-        @allocator = Spree::Config.stock.allocator_class.new(@availability)
+        @allocator = Spree::Config.stock.allocator_class.new(availability)
       end
 
       def shipments
-        @shipments ||= build_shipments
+        @shipments ||= begin
+          @packages = build_packages
+          shipments = build_shipments
+
+          # Make sure we don't add the proposed shipments to the order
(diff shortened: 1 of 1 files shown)


waiting-for-dev (approved):
Thanks for taking the time to improve this and submitting your collaboration :heart: I don't see any breaking change and I'm sure that will help other developers!

kennyadsl (approved):
👍 Nice, and now those variables are easier to deprecate if we need to change or eliminate them one day.

kennyadsl:
@BenMorganIO I saw that the coverage decreased a bit with this PR (see failing check). I think it's related to all the new public methods added that are never tested. Maybe it can be fixed "easily" by testing the `#initialize` method and checking that those attributes contain the expected value. Would you mind giving this approach a try?

jarednorman (changes requested):
I'm not in favour of adding to the API of our objects for debugging purposes, but I've spent enough time debugging this class that I'm willing to go along regardless. 😄

That said, is there a way to make it explicit that these methods are not part of the official public API of this class? I would rather we not end up going through a deprecation cycle for a bunch of methods that are public only for debugging purposes when this class inevitably changes.

BenMorganIO:
@jarednorman we could prefix them with an underscore? That seems to tell devs not to trust that piece of the API from remaining consistent.

core/app/models/spree/stock/simple_coordinator.rb
```diff
@@ -20,41 +20,64 @@ module Stock
     # Finally, shipping rates are calculated using the class configured as
     # Spree::Config.stock.estimator_class.
     class SimpleCoordinator
-      attr_reader :order
+      attr_reader :order, :inventory_units, :splitters, :stock_locations,
+        :filtered_stock_locations, :inventory_units_by_variant, :desired,
+        :availability, :allocator, :packages
```
jarednorman, line 25:
```suggestion
      attr_reader :order
      
      # This class is notoriously opaque. To aid in debugging, we've provided
      # the following methods. These are not considered part of the public API
      # of this class and may be removed without deprecation in a future
      # version of Solidus. Do not use them in production code.
      attr_reader :inventory_units, :splitters, :stock_locations,
        :filtered_stock_locations, :inventory_units_by_variant, :desired,
        :availability, :allocator, :packages
```

waiting-for-dev: `@api private` should also do the trick :slightly_smiling_face:

jarednorman: 🙏🏻 Perfect!

kennyadsl: @BenMorganIO did you see this comment about using `@api private`?

waiting-for-dev: @BenMorganIO, what do you think about just using `@api private`? This one is very close to being merged :slightly_smiling_face:

waiting-for-dev: nvm, I did it myself :slightly_smiling_face: 

@jarednorman, are you ok with merging it now?

jarednorman (approved):

PR #5257, opened 2023-07-18
[Admin] Implement search bar in product listing index

### LINKED ISSUE #5252: [Admin] Live search functionality
We are planning to add live search functionality to the Solidus Admin interface, specifically in the product listing.

The goal is to allow users to dynamically search through the product listing and observe results updating in real-time without any page refresh.

To achieve this, we can leverage the Hotwire's Turbo Frames.
Here are the steps to implement this solution:
 - Turbo Frame integration: Embed the product listing inside a Turbo Frame. This provides the ability to refresh the listing independently from the rest of the page.
 - Search Form submission: Configure the search form to submit on user input. This could be accomplished with a new component by using Stimulus to capture input events and trigger form submission.

rainerdema:
## Summary

Depends on:
- https://github.com/solidusio/solidus/pull/5286

Fixes #5252
Screenshot 2023-07-27 at 12 29 07

https://github.com/solidusio/solidus/assets/19948291/b83a0215-794b-48ef-a588-f0d1928fe89d

## Checklist

# Questions

**Why wasn't a separate search bar component created?**
 Due to the anticipated complexity from numerous variants and upcoming integrations, a standalone search bar component wasn't created. Instead, the search functionality is embedded within the UI/table component.
This approach was taken to avoid prematurely committing to a design without fully understanding future requirements (these include features such as saving search views with different predefined and custom scopes).


admin/app/components/solidus_admin/ui/button/component.rb
```diff
@@ -54,6 +55,8 @@ def initialize(tag: :button, text: nil, class_name: nil, size: :m, scheme: :prim
   end
 
   def call
-    content_tag(@tag, @text, class: @class_name, **@attributes)
+    content_tag(@tag, class: @class_name, **@attributes) do
+      safe_join([@icon_tag, @text].compact, " ")
+    end
```
rainerdema, line 60: 👀 @elia, could https://github.com/solidusio/solidus/pull/5284 possibly address this issue?

admin/app/components/solidus_admin/ui/table/component.rb
```diff
@@ -136,6 +139,90 @@ def render_data_cell(cell, data)
...
+      data: {
+        "#{stimulus_id}_target": "searchForm"
+      }
+    }
+  end
+
+  def search_field_options
+    {
+      class: "border-2 border-gray-100 bg-white h-10 px-10 rounded-lg text-sm focus:outline-none flex-grow",
+      placeholder: "Search",
```
kennyadsl, line 161: Should this go under `i18n`? Also, because the search is by `name_or_slug_cont`, I think we could add those fields in the placeholder, eg `Search by Name or Slug`, so maybe it's a good idea to create a i18n key the search fields configuration included. 

Screenshot 2023-07-26 at 10 13 40@2x

On a side note, currently Solidus allows searching for Name and SKU with the text fields, shouldn't we reflect that?

admin/app/components/solidus_admin/ui/table/component.rb
```diff
@@ -136,6 +139,90 @@ def render_data_cell(cell, data)
...
+
+  def render_search_button
+    @button_component.new(
+      tag: :button,
+      type: :submit,
+      disabled: true,
+      class_name: "left-0 ml-0 border-none bg-transparent absolute hover:bg-transparent hover:text-gray-700",
+      icon_tag: icon_tag("search-line", class: "text-gray-600 h-4 w-4 fill-current"),
+      scheme: :ghost,
+      "aria-label": "Search"
```
kennyadsl, line 195: Same here for i18n.

admin/app/components/solidus_admin/ui/table/component.rb
```diff
@@ -136,6 +139,90 @@ def render_data_cell(cell, data)
...
+    ).call
+  end
+
+  def render_close_button
+    @button_component.new(
+      tag: :button,
+      class_name: "right-0 mr-0 border-none bg-transparent absolute hover:bg-transparent hover:text-gray-700",
+      icon_tag: icon_tag("close-circle-fill", class: "text-gray-600 h-4 w-4 fill-current"),
+      scheme: :ghost,
+      "aria-label": "Close",
```
kennyadsl, line 205: Same here for i18n.

admin/app/components/solidus_admin/ui/table/component.rb
```diff
@@ -136,6 +139,90 @@ def render_data_cell(cell, data)
...
+    ).call
+  end
+
+  def render_filter_button
+    @button_component.new(
+      tag: :button,
+      class_name: "flex items-center justify-center w-9 h-9 pt-1.5 pr-1.5 pb-1.5 pl-1.5",
+      icon_tag: icon_tag("filter-3-line", class: "text-gray-600 w-5 h-5 fill-current"),
+      scheme: :secondary,
+      "aria-label": "Filter",
```
kennyadsl, line 228: Same here for i18n.

admin/app/components/solidus_admin/ui/table/component.js
```diff
@@ -10,6 +24,33 @@ export default class extends Controller {
...
   }
 
+  toggleSearch(event) {
+    this.filterButtonTarget.classList.toggle('hidden')
+    this.cancelButtonTarget.classList.toggle('hidden')
+    this.searchFormTarget.classList.toggle('hidden')
+    this.checkField()
+  }
+
+  search() {
```
kennyadsl, line 34: Not blocking, but we could restrict the search to start after at least 2 or 3 chars have been typed. I suspect this search to be quite expensive in terms of performances but we can also address this later. Just wanted to point it out.

kennyadsl (commented):
I like this, left some comments but overall seems almost ready to ship!

admin/app/components/solidus_admin/ui/table/component.rb
rainerdema: You're right. I've adjusted the search key by using the variant association, enabling the search bar to look up by product name or variants SKU.

As for the placeholder, we're now using an i18n key related to that specific search key:
Screenshot 2023-07-27 at 12 28 44

Thank you @kennyadsl and let me know what do you think 👍

admin/app/components/solidus_admin/ui/table/component.rb
rainerdema: 👍 for this specific one, I also included an `"aria-disabled": true` to reflect the search button status.

admin/app/components/solidus_admin/ui/table/component.js
rainerdema: Thank you for bringing up this point 🙌  
Indeed, initiating a search after a few characters have been typed could reduce the load.

However, I've implemented a debouncing mechanism for the search function.
This mechanism waits until the user has finished typing (200ms after the last keystroke) before executing the search:
As a result, even if the user types quickly, the search won't be triggered until they've paused by limiting the frequency of the search requests.
This mechanism should help optimize performance, but we'll certainly keep an eye on this and reconsider if necessary.

I also added a comment on the search function to explain this mechanism. 👍

admin/app/components/solidus_admin/ui/table/component.yml
```diff
@@ -3,3 +3,9 @@ en:
   rows_selected: 'selected'
   select_all: 'Select all'
   select_row: 'Select row'
+  search: Search
+  close: Close
+  filter: Filter
+  placeholders:
+    search_key:
+      name_or_variants_including_master_sku_cont: Search by Name or Sku
```
kennyadsl, line 11: How about including a line in the preference comment that clarifies the need to create a new key with the same value if changed?

admin/app/components/solidus_admin/ui/table/component.js
kennyadsl: Makes sense, I saw the debouncing and I think it's already a great thing. Let's keep this in mind if needed in the future, thanks!

admin/app/components/solidus_admin/products/index/component.html.erb
```diff
@@ -21,6 +21,7 @@
     search_page_url: search_page_url,
     previous_page_url: previous_page_url,
     next_page_url: next_page_url,
+    turbo_frame_id: 'products-list',
```
kennyadsl, line 24: I have some doubts about passing the `turbo_frame_id` explicitly from outside. It seems something related to the internal implementation that shouldn't be visible from the outside. What if we call it just `id` on the interface and use it internally for all the turbo-related stuff?

kennyadsl (approved):
Left another round of comments. Nothing is blocking but I'd like to know your thoughts about some of them before merging.

admin/app/components/solidus_admin/products/index/component.html.erb
kennyadsl: Related, Rails has these [`dom_id`](https://api.rubyonrails.org/classes/ActionView/RecordIdentifier.html#method-i-dom_id) and [`dom_class`](https://api.rubyonrails.org/classes/ActionView/RecordIdentifier.html#method-i-dom_class) helpers which we could use to generate an HTML ids for resources. See also https://boringrails.com/articles/rails-dom-id-the-most-underrated-helper/ .

admin/app/components/solidus_admin/ui/table/component.yml
rainerdema: That makes sense 👍  I've incorporated a comment and assigned a default value for the translation.

admin/app/components/solidus_admin/products/index/component.html.erb
rainerdema: 🙏 I was looking for a way to remove that id. I initialized the id inside the initializer of the `ui/table`. Let me know what you think.

kennyadsl (approved):
👏 Thanks @rainerdema!

elia:
🤦‍♂️ merging the pagination one closed this, I'll take care of reopening it one way or another

elia:
Reopening!

PR #5258, opened 2023-07-19
[Admin] Spike for data & configuration layer for tables

waiting-for-dev:
## Summary

**This PR is a proof of concept on adding a data and configuration layer on top of the table rendering. It's not meant to be merged, and it serves as a way to understand tradeoffs**

Currently, users need to replace a given table-based component in order to add a column. For instance, adding the `discontinue_on` column to the products table could look like this:

```ruby
# app/components/my_application/solidus_admin/products/component.rb
class MyApplication::SolidusAdmin::Products::Component < ::SolidusAdmin::Products::Component
  def columns
    super + [{
      header: :disconue_on,
      data: ->{ discontinue_on_column(_1) }
    }]
  end
  
  def discontinue_on_column(product)
    tag.span(class: "text-bold") { product.discontiue_on }
  end
end
```

The above code requires the user to know the internals of the table component. It has the potential to make use of any available method in it, making the new component brittle against future upgrades.

An interface that allows adding columns from outside would be more flexible and less brittle. This proof of concept shows how that could look like:

```ruby
# config/initializers/solidus_admin.rb
SolidusAdmin::Config.products.index do |columns|
  columns.add(
    name: :discontinue_on,
    header: :discontinue_on,
    renderer: ->(product) { tag.span(class: "text-bold") { product.discontiue_on } },
    position: 100
  )
end
```

It's important to notice that the `renderer` lambda is not executed in the context of the table component, which would expose the same kind of problems as the previous example, only making the experience more user-friendly. Instead, the available context is defined in `SolidusAdmin::Configuration::Products::Index::RenderContext`. The current version only gives access to Rails' tag helpers and SolidusAdmin's helpers to render other components, but it could be better tailored or even made configurable.

### Implementation

The `SolidusAdmin::Column` object has been extracted from within the table component. It accepts a `render_context` as an argument, which is set, in the example, to the products component for the columns defined from within the `solidus_admin` engine, and to the constrained `SolidusAdmin::Configuration::Products::Index::RenderContext` for user-defined columns.

To render a column cell, the `SolidusAdmin::Column` object calls the `renderer` lambda in the context of the defined `render_context`. By default, the `renderer` calls a `:"#{column_name}_column"` method on the context, which is leveraged by the `solidus_admin` components. Users can instead define their renderers. The `renderer` lambda takes as an argument the result of extracting data from every row, which is specified by the `data` parameter in `SolidusAdmin::Column`. By default, that's the complete row, but it can be trimmed down via a proc for a better separation of concerns.

For now, the behavior to render a column header is similar but with assumptions about the associated model class. Especially this part should not be seen as definitive, as it would still need a lot of refinements.

The component previews haven't been updated and are not working for now.

### Future work

In the case of this one moving forward, we should work on:

- Bringing the data definitions within the components sidecar directory.
- Better handling of column headers.
- New API to remove columns or modify the rendering for the ones already present.
- Tweakings the render context that is bound.

### Tradeoffs

Good:

- More flexibility for users.
- It gives us more license to treat some parts of table components as black boxes.
- The API can be used to add columns in any position, which is not that easy to do with the current implementation.
- The interface could easily be extended to also remove columns or modify the rendering of existing ones.

Bad:

- Adds a lot of complexity to tables (a lot of indirections).
- Makes contributions harder, as everything is much more abstracted.

### DIFF
--- a/admin/app/components/solidus_admin/products/index/component.html.erb
+++ b/admin/app/components/solidus_admin/products/index/component.html.erb
@@ -27,17 +27,10 @@
         icon: 'play-circle-line',
       },
     ],
-    columns: [
-      {
-        class_name: "w-10",
-        header: tag.span('aria-label': t('.product_image'), role: 'text'),
-        data: -> { image_column(_1) },
-      },
-      { header: :name, data: -> { name_column(_1) } },
-      { header: :status, data: -> { status_column(_1) } },
-      { header: :price, data: -> { price_column(_1) } },
-      { header: :stock, data: -> { stock_column(_1) } }
-     ]
+    columns: @columns,
+    column_classes: {
+      image: "w-10"
+    }
    )
   %>
 </div>
--- a/admin/app/components/solidus_admin/products/index/component.html.erb
+++ b/admin/app/components/solidus_admin/products/index/component.html.erb
@@ -28,7 +28,11 @@
       },
     ],
     columns: [
-      { header: "", data: -> { image_column(_1) }, class_name: "w-10" },
+      {
+        class_name: "w-10",
+        header: tag.span('aria-label': t('.product_image'), role: 'text'),
+        data: -> { image_column(_1) },
+      },
       { header: :name, data: -> { name_column(_1) } },
       { header: :status, data: -> { status_column(_1) } },
       { header: :price, data: -> { price_column(_1) } },
--- a/admin/app/components/solidus_admin/products/index/component.html.erb
+++ b/admin/app/components/solidus_admin/products/index/component.html.erb
@@ -7,6 +7,26 @@
 
   <%= render component("ui/table").new(
     page: @page,
+    batch_actions: [
+      {
+        display_name: t('.batch_actions.delete'),
+        action: solidus_admin.products_path,
+        method: :delete,
+        icon: 'delete-bin-7-line',
+      },
+      {
+        display_name: t('.batch_actions.discontinue'),
+        action: solidus_admin.discontinue_products_path,
+        method: :put,
+        icon: 'pause-circle-line',
+      },
+      {
+        display_name: t('.batch_actions.activate'),
+        action: solidus_admin.activate_products_path,
+        method: :put,
+        icon: 'play-circle-line',
+      },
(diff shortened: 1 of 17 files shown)


admin/app/components/solidus_admin/products/index/component.html.erb
```diff
@@ -27,17 +27,10 @@
...
-        data: -> { image_column(_1) },
-      },
-      { header: :name, data: -> { name_column(_1) } },
-      { header: :status, data: -> { status_column(_1) } },
-      { header: :price, data: -> { price_column(_1) } },
-      { header: :stock, data: -> { stock_column(_1) } }
-     ]
+    columns: @columns,
+    column_classes: {
+      image: "w-10"
```
kennyadsl, line 32: I'm sure this is intentional but what's the reasoning behind not having the classes in the Column itself? It's true that this file might be responsible for styling and the data layer shouldn't know about it, but might have a specific style depending on the data of the column. Seeing it from this angle, I probably see this class more close to the column definition. If people add a new column, where should they define their class?

kennyadsl (commented):
Nice proposal, I like this interface for the final user. 

True that it will be harder for the contributors and I'm ok with that because we want to prioritize being friendly with the end user. Still, I'd add to the cons section that it will be also harder for people that wants to customize the table component going out of the API that we offer. In that case, if they have some custom needs and they want to override the table component, they will inherit all of this complexity in their application without a real need of it. Do you think there's a way to avoid that?

admin/app/components/solidus_admin/products/index/component.html.erb
waiting-for-dev: Yeah, that also deserves more attention. It can probably be added to the column definition.

kennyadsl:
@waiting-for-dev at the end, the component itself is not incredibly more complex than the previous version, so maybe it's not a giant problem.

waiting-for-dev:
For context, we decided to go for it, but we'll wait until we have at least one more table component. We also want to study how other features, like filters or forms, could be extended to see if the rendering part has a standard extraction.

PR #5259, opened 2023-07-19
[Admin] Misc `ui/table` fixes and improvements

elia:
## Summary

A few commits extracted after the first review on the batch actions PR (#5180).

The only one slightly out of the realm of `ui/table` or `products/index` is the one adding support for the `hidden` attribute. After using the `hidden` tailwind class for showing/hiding the batch action toolbars I realized that there is a native way hide/show elements, but that I needed a TW support to properly style it.

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/products/index/component.html.erb
+++ b/admin/app/components/solidus_admin/products/index/component.html.erb
@@ -10,7 +10,7 @@
     columns: [
       {
         class_name: "w-[72px]",
-        header: "",
+        header: tag.span('aria-label': t('.product_image'), role: 'text'),
         data: -> { image_column(_1) },
       },
       { header: :name, data: -> { name_column(_1) } },
--- a/admin/app/components/solidus_admin/products/index/component.html.erb
+++ b/admin/app/components/solidus_admin/products/index/component.html.erb
@@ -8,7 +8,11 @@
   <%= render component("ui/table").new(
     page: @page,
     columns: [
-      { header: "", data: -> { image_column(_1) }, class_name: "w-10" },
+      {
+        class_name: "w-[72px]",
+        header: "",
+        data: -> { image_column(_1) },
+      },
       { header: :name, data: -> { name_column(_1) } },
       { header: :status, data: -> { status_column(_1) } },
       { header: :price, data: -> { price_column(_1) } },
--- a/admin/spec/helpers/solidus_admin/layout_helper_spec.rb
+++ b/admin/spec/helpers/solidus_admin/layout_helper_spec.rb
@@ -19,9 +19,7 @@ RSpec.describe SolidusAdmin::LayoutHelper, :helper do
     end
 
     it "includes the translated controller base name" do
-      expect(
-        helper.solidus_admin_title(store_name: "My Store", controller_name: "orders")
-      ).to include("Orders")
+      expect(helper.solidus_admin_title(store_name: "My Store", controller_name: 'products')).to eq("My Store - Products")
     end
   end
 end
--- a/admin/app/components/solidus_admin/products/index/component.rb
+++ b/admin/app/components/solidus_admin/products/index/component.rb
@@ -9,8 +9,9 @@ class SolidusAdmin::Products::Index::Component < SolidusAdmin::BaseComponent
     image = product.gallery.images.first or return
 
     link_to(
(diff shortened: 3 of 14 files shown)


admin/app/components/solidus_admin/products/index/component.html.erb
```diff
@@ -8,7 +8,11 @@
   <%= render component("ui/table").new(
     page: @page,
     columns: [
-      { header: "", data: -> { image_column(_1) }, class_name: "w-10" },
+      {
+        class_name: "w-[72px]",
```
kennyadsl, line 12: Can you help me understand why we can't use `px-4` here? I understand that having the content `flex items-center` now it will be centered, but isn't the same to use padding if we know the image size upfront?

admin/config/solidus_admin/tailwind.config.js.erb
```diff
@@ -82,5 +83,7 @@ module.exports = {
     require('@tailwindcss/aspect-ratio'),
     require('@tailwindcss/typography'),
     require('@tailwindcss/container-queries'),
+    plugin(({ addVariant }) => addVariant('hidden', '&([hidden])')),
```
kennyadsl, line 86: Am I right that we are not using these modifiers anywhere?

kennyadsl (commented):
Left a couple of questions, thanks @elia!

admin/app/components/solidus_admin/products/index/component.html.erb
elia: The table layout was changed from `auto` to `fixed`, that means that the browser will try to have equally sized columns. This creates less jumping when the header is hidden but also requires forcing the width for some columns.

kennyadsl: 👍 but isn't that auto-centering without jumping working anyway by using the padding? Just to be clear, I know your solution works, I'm just trying to stay away from custom values in px as much as possible and stick to what's defined in the figma file, but if there are good reasons not to do that, I'm totally ok with this version.

admin/config/solidus_admin/tailwind.config.js.erb
kennyadsl: Thoughts about moving this to that other PR? If, for some reason, the reviews of that one will change what we are anticipating, we might end up not using these modifiers and forget about them.

admin/app/components/solidus_admin/products/index/component.html.erb
kennyadsl: This is solved. I didn't know that the fixed table layout would set the width of the column fixed only if specified directly on the column (or first row). This sounds like the only way to make it work without jumps. 👍

admin/spec/helpers/solidus_admin/layout_helper_spec.rb
```diff
@@ -19,9 +19,7 @@
     end
 
     it "includes the translated controller base name" do
-      expect(
-        helper.solidus_admin_title(store_name: "My Store", controller_name: "orders")
-      ).to include("Orders")
+      expect(helper.solidus_admin_title(store_name: "My Store", controller_name: 'products')).to eq("My Store - Products")
```
elia, line 22: While updating this spec I was wondering if we shouldn't ensure that the defaults work, i.e. checking that `current_store` (which is provided by `Spree::Core::ControllerHelpers::Store`) is available as a helper and so on. 

Also the previous expectation was passing even if the translation was missing, should the translation backend be injected as well?

My feeling is that we're somewhat fighting the system with this approach and we would be better off not trying to inject every dependency and only do that if there's a significant advantage over a more common use of instance and partial doubles:

```rb
    it "includes the translated controller base name" do
      helper.extend Spree::Core::ControllerHelpers::Store

      allow(helper).to receive(:current_store).and_return(build(:store, name: "My Store"))
      allow(helper).to receive(:controller_name).and_return("products")

      expect(helper.solidus_admin_title).to eq("My Store - Products")
    end
```

@waiting-for-dev and @solidusio/nebulab I'd be interested in knowing how you would approach it given the context above 💭

waiting-for-dev: Feel free to change it, @elia.

waiting-for-dev (approved):

kennyadsl (approved):

rainerdema (approved):
👏

PR #5275, opened 2023-07-21
Stop setting line items with quantity of zero

RyanofWoods:
## Summary

In order to order to be able to update LineItem's validations to stop allowing a quantity of zero, we need to remove code that saves LineItems with a quantity of 0 (or less which gets normalized to 0).

## Checklist

### DIFF
--- a/core/app/models/spree/order_contents.rb
+++ b/core/app/models/spree/order_contents.rb
@@ -35,9 +35,10 @@ module Spree
     end
 
     def update_cart(params)
+      params = destroy_non_positive_line_items(params)
+
       if order.update(params)
         unless order.completed?
-          order.line_items = order.line_items.select { |li| li.quantity > 0 }
           # Update totals, then check if the order is eligible for any cart promotions.
           # If we do not update first, then the item total will be wrong and ItemTotal
           # promotion rules would not be triggered.
@@ -81,6 +82,20 @@ module Spree
 
     def reload_totals
       @order.recalculate
+    end
+
+    def destroy_non_positive_line_items(params)
+      line_items_to_destroy_ids = []
+      # Could be a single LineItem or multiple via nesting
+      if id = params[:line_items_attributes][:id].present?
+        line_items_to_destroy_ids << id if params[:line_items_attributes][:quantity].to_i <= 0
+      else
+        params[:line_items_attributes].reject! do |_index, attributes|
+          line_items_to_destroy_ids << attributes[:id] if attributes[:quantity].to_i <= 0
+        end
+      end
+      order.line_items.where(id: line_items_to_destroy_ids).destroy_all
+      params
     end
 
     def add_to_line_item(variant, quantity, options = {})
--- a/api/app/controllers/spree/api/line_items_controller.rb
+++ b/api/app/controllers/spree/api/line_items_controller.rb
@@ -13,7 +13,7 @@ module Spree
         variant = Spree::Variant.find(params[:line_item][:variant_id])
         @line_item = @order.contents.add(
           variant,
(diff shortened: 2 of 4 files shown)


core/app/models/spree/order_contents.rb
```diff
@@ -83,6 +84,20 @@ def reload_totals
...
+      line_items_to_destroy_ids = []
+      # Could be a single LineItem or multiple via nesting
+      if id = params[:line_items_attributes][:id].present?
+        line_items_to_destroy_ids << id if params[:line_items_attributes][:quantity].to_i <= 0
+      else
+        params[:line_items_attributes].reject! do |_index, attributes|
+          line_items_to_destroy_ids << attributes[:id] if attributes[:quantity].to_i <= 0
+        end
+      end
+      order.line_items.where(id: line_items_to_destroy_ids).destroy_all
```
kennyadsl, line 97: Thoughts on using https://github.com/solidusio/solidus/blob/df68c0c748ded6a595d28c37a9460126c6d12ecb/core/app/models/spree/order_contents.rb#L32?

core/app/models/spree/order_contents.rb
```diff
@@ -83,6 +84,20 @@ def reload_totals
...
 
+    def destroy_non_positive_line_items(params)
+      line_items_to_destroy_ids = []
+      # Could be a single LineItem or multiple via nesting
+      if id = params[:line_items_attributes][:id].present?
+        line_items_to_destroy_ids << id if params[:line_items_attributes][:quantity].to_i <= 0
+      else
+        params[:line_items_attributes].reject! do |_index, attributes|
+          line_items_to_destroy_ids << attributes[:id] if attributes[:quantity].to_i <= 0
+        end
```
kennyadsl, line 95: In which cases do we have this difference in passing the attributes?

kennyadsl (commented):
I know it's a draft, but I was taking a look and left a couple of comments.

core/app/models/spree/order_contents.rb
RyanofWoods: I would prefer not to use this method because it would cause a DELETE SQL for every destroy rather than just one. It would also cause two extra order recalculates for every destroy. Note that `update_cart` explicitly handles recalculating anyway.

core/app/models/spree/order_contents.rb
RyanofWoods: Do you mean whether it's nested or not? Like so:
```ruby
# Case 1:
line_item_attributes: {
  id: 1,
  quantity: 1
}

# Case 2:
line_item_attributes: {
  "01" => {
    id: 1,
    quantity: 1
  },
  "02" => {
    id: 3,
    quantity: 0
  }
}
```

I am supporting single nesting for `Api::LineItemsController`, though it seems that Rails supports these different levels by default - though I couldn't find any explicit documentation about it 🤔

kennyadsl: Hey @RyanofWoods, I'm not able to detect any code in the `Api::LineItemsController` or its specs that seem to support this way of passing attributes. Can you please help me adding more details about your discovery?

RyanofWoods: Case 1 is needed for `Spree::Api::LineItemsController`:
https://github.com/solidusio/solidus/blob/834347e6477ecb0b64c622d7c9c574050a41b804/api/app/controllers/spree/api/line_items_controller.rb#L56-L62

Case 2 was needed for how the OrderContents spec is setup.

Both these formats are supported by the nested_attributes in Rails.

fthobe:
@RyanofWoods this was a really cool change, can we get this over the finish line?

PR #5299, opened 2023-08-03
[Admin] Add toggle switch component 

the-krg:
## Summary

Screen Shot 2023-08-03 at 03 07 44

https://github.com/solidusio/solidus/assets/31260493/01cf4114-4065-48bd-92ca-cfc6c617a1c0

Screen Shot 2023-08-03 at 10 59 30

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/ui/forms/switch/component.rb
+++ b/admin/app/components/solidus_admin/ui/forms/switch/component.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+class SolidusAdmin::UI::Forms::Switch::Component < SolidusAdmin::BaseComponent
+  SIZES = {
+    s: '
+      w-8 h-5
+
+      after:w-4 after:h-4
+      after:top-0.5 after:left-0.5
+
+      active:after:w-4
+      after:checked:left-[1.875rem]
+    ',
+    m: '
+      w-10 h-6
+
+      after:w-5 after:h-5
+      after:top-0.5 after:left-0.5
+
+      active:after:w-5
+      after:checked:left-[2.375rem]
+    ',
+  }.freeze
+
+  def initialize(size: :m, **attributes)
+    @size = size
+    @attributes = attributes
+  end
+
+  def call
+    tag.input(
+      type: 'checkbox',
+      class: "
+        #{SIZES.fetch(@size)}
+        appearance-none	outline-0 cursor-pointer bg-gray-200 inline-block rounded-full relative
+
+        after:content-[''] after:absolute after:bg-white
+        after:duration-300 after:rounded-full after:checked:-translate-x-full
+
+        hover:bg-gray-300
+        disabled:opacity-40 disabled:cursor-not-allowed
+        checked:bg-gray-500 checked:hover:bg-gray-700
+      ",
+      **@attributes,
+    )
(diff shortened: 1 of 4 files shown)


admin/app/components/solidus_admin/ui/forms/switch/component.rb
```diff
@@ -0,0 +1,50 @@
...
+    ',
+  }.freeze
+
+  def initialize(id: SecureRandom.hex, size: :m, **attributes)
+    @id = id
+    @size = size
+    @attributes = attributes
+    @states = []
+    @states << "checked" if @attributes[:checked]
+    @states << "disabled" if @attributes[:disabled]
```
elia, line 31: If we collect any attribute in the initializer we should apply it to the element, otherwise we can specifically collect just `checked` and `disabled`. Of course is unclear if the collected attributes would be applied to the label or the input.

admin/app/components/solidus_admin/ui/forms/switch/component.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      hover:bg-gray-300
+      peer-disabled:opacity-40 peer-disabled:cursor-not-allowed
+      peer-checked:bg-gray-500 peer-checked:hover:bg-gray-700
+    "
+  end
+
+  erb_template <<~ERB
+    <input type="checkbox" class="h-0 w-0 invisible peer" id="<%= @id %>" <%= @states.join(' ') %>/>
+    <label for="<%= @id %>" class="<%= @label_class %>"></label>
```
elia, line 48: I saw this is the most common implementation of a toggle switch, would be great if there was an alternative only using the `input` element (e.g. by using SVG backgrounds, like the checkbox does) or one in which there was a single root (e.g. the label `<label><input …/></label>`).

Having a single root makes it simpler to organize the component from outside (e.g. putting it in a flex container and expect it to be a single element).

At worst maybe we can wrap it into a `div` or `span`.

Thoughts?

elia (commented):
Great work! It's awesome to see components come alive ✨ 

I left a couple of comments over internal details, but looks good overall 🙌

admin/app/components/solidus_admin/ui/forms/switch/component.rb
the-krg: Agreed! I'll change it to specifically collect `checked` and `disabled`.

admin/app/components/solidus_admin/ui/forms/switch/component.rb
the-krg: the problem on putting the `<input>` inside the `<label>` tag is that the Tailwind `peer` won't work, [since it requires the peer to be defined first](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-sibling-state) (which is the `<input>` tag), so the label can change depending on its status.

the <input/> tag on the current solution is 0x0 with an `invisible` property, which shouldn't affect much, but it's definitely better to keep as one tag. 

I'll study your first suggestion, in case I'm not successful I'll stick with wrapping it on a `<div>` tag.

elia (approved):
🚀 🚀 🚀

rainerdema (approved):
Great job! 👏

PR #5324, opened 2023-08-09
Prevent hard-errors when VatPriceGenerator is invoked without a defau…

firemind:
Guard against exceptions when calling `Price#net_amount` without a price.

Skip trying to calculate foreign prices in `VatPriceGenerator` when the `default_price` does not have an amount set.

Because the `VatPriceGenerator` is invoked in a `before_validate` hook when creating a product, this can lead to exceptions when an admin tries to create a product without specifying a price (and there are non-default countries for which a price needs to be calculated).

This PR allows the validations to run without raising exceptions so the admin can see the validation errors.

### DIFF
--- a/core/app/models/spree/price.rb
+++ b/core/app/models/spree/price.rb
@@ -46,6 +46,8 @@ module Spree
     end
 
     def net_amount
+      return nil unless amount
+
       amount / (1 + sum_of_vat_amounts)
     end
 
--- a/core/app/models/spree/variant/vat_price_generator.rb
+++ b/core/app/models/spree/variant/vat_price_generator.rb
@@ -25,6 +25,7 @@ module Spree
       def run
         # Early return if there is no VAT rates in the current store.
         return if !variant.tax_category || variant_vat_rates.empty?
+        return unless variant.default_price.net_amount
 
         country_isos_requiring_price.each do |country_iso|
           # Don't re-create the default price


waiting-for-dev (approved):

jarednorman (approved):

waiting-for-dev:
I labeled the PR as a breaking change, as it's changing a behavior some people could be relying on. It'll be merged in the next major.

core/app/models/spree/price.rb
```diff
@@ -46,6 +46,8 @@ def price=(price)
     end
 
     def net_amount
+      return nil unless amount
```
adammathys, line 49: I don't believe it's possible for `amount` to be nil as of v3.1. (See the "Other important changes" section in the [`CHANGELOG.md`](https://github.com/solidusio/solidus/blob/main/CHANGELOG.md#solidus-310-v31-2021-09-10) entry for v3.1, or the PR that removes `nil` as a valid value: https://github.com/solidusio/solidus/pull/3987)

Which means this shouldn't be possible on recent versions of Solidus. (Assuming the task to remove any records with a `nil` amount has been run and there isn't any invalid data left in the database from before that change.)

adammathys (changes requested):

core/app/models/spree/price.rb
adammathys: Oops, nevermind. I missed that we were trying to fix a `before_validation` callback on `Spree::Variant`. 

In which case I think the preferred fix should be skipping that callback instead of modifying this public method. The callback already has a conditional associated to it, so I believe it would be a less invasive change to prevent it from running in the case when a price is invalid. (And we avoid any future confusion from having a `nil` check in a method where a `nil` value is considered invalid.)

jarednorman:
I've finished this work in #6566. Thanks @firemind.

PR #5330, opened 2023-08-11
Fix error added when resetting order flow on an empty order

waiting-for-dev:
## Summary

https://github.com/solidusio/solidus/pull/4369 introduced a regression where when calling `Spree::Order#restart_checkout_flow` on an empty order, an error would be added to the order. That happened when calling `#next` and the validation failed because no line items were present.

We restore the previous behavior where we only try going to the `"address"` state if the order has line items.

## Checklist

### DIFF
--- a/core/spec/models/spree/order_spec.rb
+++ b/core/spec/models/spree/order_spec.rb
@@ -590,9 +590,16 @@ RSpec.describe Spree::Order, type: :model do
     end
 
     context "without line items" do
+      let(:order) { create(:order, state: "delivery", line_items: []) }
+
       it "updates the state column to cart" do
-        order = create(:order, state: "delivery")
         expect{ order.restart_checkout_flow }.to change{ order.state }.from("delivery").to("cart")
+      end
+
+      it "doesn't add errors to the order" do
+        order.restart_checkout_flow
+
+        expect(order.errors).to be_empty
       end
     end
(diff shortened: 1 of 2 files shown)


core/spec/models/spree/order_spec.rb
```diff
@@ -594,6 +594,14 @@ def merge!(other_order, user = nil)
         order = create(:order, state: "delivery")
         expect{ order.restart_checkout_flow }.to change{ order.state }.from("delivery").to("cart")
       end
+
+      it "doesn't add errors to the order" do
+        order = create(:order, state: "delivery", line_items: [])
```
spaghetticode, line 599: What do you think about moving this line in a `let` so that we can share the same `order` definition with both specs inside this `context`? This would also remove the confusion about the different approaches for creating the empty order in delivery state (this line is explicit about empty line items, while in the previous spec, this remains implicit).

spaghetticode (approved):
@waiting-for-dev thanks for the fix, looks good to me, and I left a suggestion for the spec, please let me know if it works for you.

rainerdema (approved):
👍

core/spec/models/spree/order_spec.rb
waiting-for-dev: Changed!

PR #5359, opened 2023-08-26
Rails 7.1 support

peterberkenbosch:
## Summary

Preparation for supporting Rails 7.1. Very much work in progress, although it looks like the changes needed are minimal so far. The sandbox is running successfully, there are some failing specs that need to be addressed. 

- [x] https://github.com/activerecord-hackery/ransack/issues/1420 
- [x] Add support to Circle to run Rails from edge
- [x] Add support to Circle to run Rails 7.1
- [x] Fix all the specs
- [x] Fix or document all deprecation warnings for 7.1
- [x] Use Rails 7.1 for installer tests on CircleCI

One of the current RSpec errors are related to this https://github.com/rails/rails/issues/49591 and it looks that this will be fixed in the upcoming Rails 7.1.2 release.

Ransack is pain still. The custom predicates with formatters is failing hard currently:

`core/app/models/spree/product.rb`:

```ruby
# The :variants_option_values ransacker filters Spree::Product based on
    # variant option values ids.
    #
    # Usage:
    # Spree::Product.ransack(
    #   variants_option_values_in: [option_value_id1, option_value_id2]
    # ).result
    ransacker :variants_option_values, formatter: proc { |v|
      if OptionValue.exists?(v)
        joins(variants_including_master: :option_values)
          .where(spree_option_values: { id: v })
          .distinct
          .select(:id).arel
      end
    } do |parent|
      parent.table[:id]
    end
```

```
1) Spree::Product ransacker :variants_option_values filters products based on option values of their variants
     Failure/Error: result = Spree::Product.ransack(variants_option_values_in: [option_value_1.id]).result
     
     NoMethodError:
       undefined method `value' for #<Arel::SelectManager:0x00000001160b22c0 @ast=#<Ar
```
Figuring out where this value method call used to be. 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -358,10 +358,10 @@ workflows:
       # https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html.
       - test_solidus:
           name: &name "test-rails-<<matrix.rails>>-ruby-<<matrix.ruby>>-<<matrix.database>>-<<#matrix.paperclip>>paperclip<</matrix.paperclip>><<^matrix.paperclip>>activestorage<</matrix.paperclip>>"
-          matrix: { parameters: { rails: ['7.0'], ruby: ['3.0'], database: ['mysql'], paperclip: [true] } }
+          matrix: { parameters: { rails: ['7.0', '7.1'], ruby: ['3.0'], database: ['mysql'], paperclip: [true] } }
       - test_solidus:
           name: *name
-          matrix: { parameters: { rails: ['7.0'], ruby: ['3.1'], database: ['postgres'], paperclip: [false] } }
+          matrix: { parameters: { rails: ['7.0', '7.1'], ruby: ['3.1'], database: ['postgres'], paperclip: [false] } }
       - test_solidus:
           name: *name
-          matrix: { parameters: { rails: ['7.0'], ruby: ['3.2'], database: ['sqlite'], paperclip: [false] } }
+          matrix: { parameters: { rails: ['7.0', '7.1'], ruby: ['3.2'], database: ['sqlite'], paperclip: [false] } }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,7 +165,7 @@ commands:
             ruby -v >> /tmp/.ruby-versions
             gem --version >> /tmp/.gems-versions
             bundle --version >> /tmp/.gems-versions
-            gem search -eq rails -v "~> 7" -v"< 7.1" >> /tmp/.gems-versions # get the latest rails from rubygems
+            gem search -eq rails -v "~> 7" -v "< 7.2" >> /tmp/.gems-versions # get the latest rails from rubygems
             gem search -eq solidus >> /tmp/.gems-versions # get the latest solidus from rubygems
 
             cat /tmp/.ruby-versions
--- a/core/app/models/spree/address.rb
+++ b/core/app/models/spree/address.rb
@@ -28,8 +28,10 @@ module Spree
 
     self.allowed_ransackable_attributes = %w[name]
 
-    scope :with_values, ->(attributes) do
-      where(value_attributes(attributes))
+    unless ActiveRecord::Relation.method_defined? :with_values # Rails 7.1+
+      scope :with_values, ->(attributes) do
+        where(value_attributes(attributes))
+      end
     end
 
     # @return [Address] an address with default attributes
--- a/core/app/models/spree/credit_card.rb
+++ b/core/app/models/spree/credit_card.rb
@@ -21,9 +21,6 @@ module Spree
 
     scope :with_payment_profile, -> { where('gateway_customer_profile_id IS NOT NULL') }
 
-    # needed for some of the ActiveMerchant gateways (eg. SagePay)
-    alias_attribute :brand, :cc_type
-
     # Taken from ActiveMerchant
     # https://github.com/activemerchant/active_merchant/blob/2f2acd4696e8de76057b5ed670b9aa022abc1187/lib/active_merchant/billing/credit_card_methods.rb#L5
     CARD_TYPES = {
@@ -94,6 +91,12 @@ module Spree
       else type
       end
     end
+
+    # needed for some of the ActiveMerchant gateways (eg. SagePay)
+    alias_attribute :brand, :cc_type
+
+    # Rails 7.1+ won't use the custom setter with alias_attribute
+    alias_method :brand=, :cc_type=
 
     # Sets the last digits field based on the assigned credit card number.
     def set_last_digits
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -72,11 +72,15 @@ module Spree
 
     # Customer info
     belongs_to :user, class_name: Spree::UserClassHandle.new, optional: true
+
     belongs_to :bill_address, foreign_key: :bill_address_id, class_name: 'Spree::Address', optional: true
-    alias_attribute :billing_address, :bill_address
+    alias_method :billing_address, :bill_address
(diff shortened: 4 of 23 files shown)


core/app/models/spree/address.rb
```diff
@@ -28,10 +28,6 @@ class Address < Spree::Base
 
     self.allowed_ransackable_attributes = %w[name]
 
-    scope :with_values, ->(attributes) do
-      where(value_attributes(attributes))
-    end
-
```
jarednorman, line 34: Is it available in all versions of Rails we currently support?

peterberkenbosch: No, it's introduced in 7.1 (I think). Probably needs a version check. It's breaking with 7.1 though with a scope already defined error.

peterberkenbosch:
specs are almost passing. Figuring out why we now get different apostrophe's back from our api..

Took me a few to figure out why this was not matching: 

```
expected: "can't be blank"
            got: "can’t be blank"
```

`'` vs `’`

peterberkenbosch:
Intentional Rails update:

```
*   Improve typography of user facing error messages. In English contractions,
    the Unicode APOSTROPHE (`U+0027`) is now RIGHT SINGLE QUOTATION MARK
    (`U+2019`). For example, "can't be blank" is now "can’t be blank".

    *Jon Dufresne*
```

core/app/models/spree/shipping_rate.rb
```diff
@@ -17,9 +17,11 @@ class ShippingRate < Spree::Base
     delegate :order, :currency, to: :shipment
     delegate :name, :tax_category, :tax_category_id, to: :shipping_method
     delegate :code, to: :shipping_method, prefix: true
-    alias_attribute :amount, :cost
 
-    alias_method :total_before_tax, :amount
+    alias_attribute :amount, :cost
+    def total_before_tax
```
kennyadsl, line 22: This is related to https://github.com/rails/rails/issues/49250. There does not seem to be a proposed solution, and apparently, this is the expected behavior. 

As you mentioned in the commit description, we could just use `cost`, but if anyone customized the `amount` method in their store, it's going to be breaking. Also, `amount` is used as `money_method` below, which we should also change, but the rest of Solidus uses `amount` as money method, so it's not ideal. 

I think that for now your solution is ok, thanks Peter.

core/lib/solidus_core.rb
```diff
@@ -1,3 +1,9 @@
 # frozen_string_literal: true
 
 require 'spree_core'
+
+class Arel::Table
```
kennyadsl, line 5: 🤞 that they ship a patch, so we don't need this.

kennyadsl:
@peterberkenbosch Hey Peter, did you have any chance to keep up with this? I'd volunteer to continue the work if you want, just let me know please.

kennyadsl:
Made another pass to fix other deprecation warnings and errors. As far as I can see the biggest thing left are ransack and previews, which require more time.

core/app/models/spree/order.rb
```diff
@@ -73,10 +73,9 @@ class CannotRebuildShipments < StandardError; end
     # Customer info
     belongs_to :user, class_name: Spree::UserClassHandle.new, optional: true
     belongs_to :bill_address, foreign_key: :bill_address_id, class_name: 'Spree::Address', optional: true
-    alias_attribute :billing_address, :bill_address
-
+    alias_method :billing_address, :bill_address
     belongs_to :ship_address, foreign_key: :ship_address_id, class_name: 'Spree::Address', optional: true
-    alias_attribute :shipping_address, :ship_address
+    alias_method :billing_address, :bill_address
```
spaghetticode, line 78: I think this line should read `alias_method shipping_address, :ship_address`, as currently it's the same as the changes from line 76 🤔

core/lib/spree/core/engine.rb
```diff
@@ -68,11 +68,22 @@ class Engine < ::Rails::Engine
...
-        original_preview_path = app.config.action_mailer.preview_path
-        solidus_preview_path = Spree::Core::Engine.root.join 'lib/spree/mailer_previews'
+        if Rails.gem_version < Gem::Version.new('7.1.0')
+          original_preview_path = app.config.action_mailer.preview_path
+          solidus_preview_path = Spree::Core::Engine.root.join 'lib/spree/mailer_previews'
+
+          app.config.action_mailer.preview_path = "{#{original_preview_path},#{solidus_preview_path}}"
+          ActionMailer::Base.preview_path = app.config.action_mailer.preview_path
+        else
+          app.config.action_mailer.preview_paths << solidus_preview_path.to_s
```
spaghetticode, line 78: `solidus_preview_path` is nil here, as it's defined at line 73 which is inside the other branch of the `if`. When generating the sanbox, this creates an interesting bug by adding the filesystem root to the Rails dependencies autoload path, so when the app loads, it tries to find all rb files on the whole disk, creating 1) huge slowdown, 2) weird file permission access errors (at least on a recent MacOS computer).

core/app/models/spree/order.rb
```diff
@@ -73,10 +73,9 @@ class CannotRebuildShipments < StandardError; end
     # Customer info
     belongs_to :user, class_name: Spree::UserClassHandle.new, optional: true
     belongs_to :bill_address, foreign_key: :bill_address_id, class_name: 'Spree::Address', optional: true
-    alias_attribute :billing_address, :bill_address
-
+    alias_method :billing_address, :bill_address
```
spaghetticode, line 76: Unfortunately, this doesn't fully match the existing behavior. For example, previously we could do  (and we do it [here](https://github.com/solidusio/solidus/blob/682033394bb1cec88ef6da58223710b870b5dc15/sample/db/samples/orders.rb#L17), for example) 
```ruby
Spree::Order.new(billing_addres: Spree::Address.new)
``` 

but now this raises `ActiveModel::UnknownAttributeError: unknown attribute 'billing_address' for Spree::Order. (ActiveModel::UnknownAttributeError)`

core/app/models/spree/credit_card.rb
```diff
@@ -95,6 +95,10 @@ def cc_type=(type)
       end
     end
 
+    def brand=(type)
+      cc_type = type
```
spaghetticode, line 99: I'm thinking this should be `self.cc_type =` otherwise looks like a variable assignment to me 🤔

.circleci/config.yml
```diff
@@ -347,4 +347,7 @@ workflows:
       - test_solidus:
           name: *name
           matrix: { parameters: { rails: ['7.0'], ruby: ['3.2'], database: ['sqlite'], paperclip: [false] } }
+      - test_solidus:
```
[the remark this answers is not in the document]
elia: I'm moving this to a different PR so we can focus on 7.1

core/app/models/spree/shipping_rate.rb
elia: Looks like the proposed alternative in the rails ticket is to use `alias_attribute` twice, I'll update that

core/app/models/spree/order.rb
elia: adding `alias_method :billing_address=, :bill_address=` should fix this issue

core/app/models/spree/credit_card.rb
elia: I turned it into an `alias_method`

kennyadsl (commented):
Thanks for completing the work here, @elia @spaghetticode!

kennyadsl (approved):

rainerdema (approved):
👏👏👏

tvdeyen:
Great work! Thanks 🙏🏻

peterberkenbosch:
great!! thanks all for pushing this over the finish line! 🎉

PR #5360, opened 2023-08-29
[Admin] Add form component

waiting-for-dev:
## Summary

The first commit belongs to #5358.

Introducing a form component brings a two-fold benefit:

1. It wraps all the form component dependencies in a single place so that we don't need to redeclare them in every other form (e.g., the product creation form).
2. It allows defining a form from the outside, which is a first step towards making the admin forms configurable.

Internally, it uses Rails' [`form_with` helper](https://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-form_with), and all the given arguments are passed to it. The only exception is the `elements:` key, which is used to define the form elements.

The form elements are simple builders of renderable things that happen within a form context. That's to say, to be rendered, both the form component instance and the Rails' form builder are given to a `#call` method.

We ship with out-of-the-box form elements for fields, fieldsets and arbitrary components or HTML.

Notice that there's a circular dependency between form and form elements. While the form calls the elements to get the renderable content, the elements can use the form to get the configured dependencies. For instance, when rendering a text field element, it'll use the configured text_field dependency in the form. This complexity is an acceptable trade-off for a double benefit: the ability to change all the form dependencies at once (e.g., using a different text field component across a form), and the removal of boilerplate from the form element definitions. Nonetheless, the default behavior can be overridden by passing a different component to the form element.

This commit also makes some tangential changes that were necessary:

- We rename the `form:` parameter to `builder:` in the components within the form namespace. As we have now a form component, that makes it more explicit that it corresponds to the Rails' form builder.
- We add `.rb` files within the preview directories to Tailwind watchable files.
- We modify the `mock_component` test helper to accept not providing a block.

Ref. #5329

## Checklist


admin/spec/components/previews/solidus_admin/ui/forms/text_field/component_preview.rb
```diff
@@ -6,7 +6,11 @@ class SolidusAdmin::UI::Forms::TextField::ComponentPreview < ViewComponent::Prev
 
   # The text field component is used to render a text field in a form.
   #
-  # It must be used within the block context yielded in the [`form_with`
+  # Most commonly, it'll be used indirectly through the definition given to a
+  # [form component](../form/overview).
+  #
+  # For standalone usage, it must be used within the block context yielded in
+  # the [`form_with`
   # ](https://api.rubyonrails.org/v5.1/classes/ActionView/Helpers/FormHelper.html#method-i-form_with)
```
rainerdema, line 14: I think there is an extra line here

rainerdema (approved):
This is amazing, thanks @waiting-for-dev!

admin/spec/components/previews/solidus_admin/ui/forms/text_field/component_preview.rb
waiting-for-dev: Well, that was the automatic formatting to make the line fit the max length :slightly_smiling_face:

PR #5364, opened 2023-08-31
[Admin] Add modal component

the-krg:
## Summary
Those slots allow the developer to pass a block to the component, this enables an in-depth customization of the component, especially for the modal, in which the body should be dynamic. Since modals also carry action confirmations,
I've added the possibility to change the action button as well.

The modal is actually just a UI trick to show regular content and the close button (along with clicking on the backdrop) are simple links and just point to a URL which will need to be provided explicitly.

Currently we don't have a use case for `alert()` / `confirm()` style modals that are transactional, but we'll bake in that functionality if we'll bump into such cases.

image
image
image

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/ui/modal/component.html.erb
+++ b/admin/app/components/solidus_admin/ui/modal/component.html.erb
@@ -1,0 +1,37 @@
+<dialog
+  data-controller="<%= stimulus_id %>"
+  <%= tag.attributes(
+    "aria-label": @title,
+    class: "relative z-10",
+    **@attributes
+  ) %>
+>
+  <a href="<%= @close_path %>" aria-hidden="true" class="cursor-default fixed inset-0 bg-gray-100/30 backdrop-blur-sm overflow-y-auto"></a>
+  <div class="fixed inset-0 z-10 pointer-events-none flex min-h-full justify-center p-4 text-center items-center">
+    <div class="pointer-events-auto cursor-auto relative transform overflow-auto rounded-lg bg-white text-left shadow-xl max-w-lg divide-y divide-gray-100">
+
+      <header class="flex items-center justify-between p-4">
+        <h3 class="text-xl font-semibold text-gray-900">
+          <%= @title %>
+        </h3>
+        <%= render component('ui/button').new(
+          tag: :a,
+          href: @close_path,
+          icon: 'close-line',
+          scheme: :ghost,
+          title: t('.close'),
+        ) %>
+      </header>
+
+      <div class="p-4 overflow-auto max-h-96">
+        <%= content %>
+      </div>
+
+      <% if actions? %>
+        <footer class="p-4 flex gap-2 justify-end">
+          <%= actions %>
+        </footer>
+      <% end %>
+    </div>
+  </div>
+</dialog>
--- a/admin/app/components/solidus_admin/ui/modal/component.rb
+++ b/admin/app/components/solidus_admin/ui/modal/component.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class SolidusAdmin::UI::Modal::Component < SolidusAdmin::BaseComponent
+  renders_one :actions
+
+  def initialize(title:, close_path: nil, open: true, **attributes)
+    @title = title
+    @close_path = close_path
+    @attributes = attributes
+    @attributes[:open] = open
+  end
+end
--- a/admin/app/components/solidus_admin/ui/modal/component.yml
+++ b/admin/app/components/solidus_admin/ui/modal/component.yml
@@ -1,0 +1,2 @@
+en:
+  close: Close
--- a/admin/spec/components/previews/solidus_admin/ui/modal/component_preview.rb
+++ b/admin/spec/components/previews/solidus_admin/ui/modal/component_preview.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+# @component "ui/modal"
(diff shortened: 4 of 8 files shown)


admin/app/components/solidus_admin/ui/modal/component.html.erb
```diff
@@ -0,0 +1,33 @@
+<div data-controller=<%= stimulus_id %> id=<%= @id %> class="relative z-10 hidden" aria-labelledby="modal-title" role="dialog" aria-modal="true">
```
elia, line 1: Thoughts on using the `<dialog>` element directly instead of using `role="dialog"`?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

elia (commented):
Great! Left a couple of suggestions / questions 👍

admin/app/components/solidus_admin/ui/modal/component.js
```diff
@@ -0,0 +1,9 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static values = { modal: String }
+
+  toggleModal(e) {
+    document.getElementById(e.currentTarget.dataset.modalId).classList.toggle('hidden');
```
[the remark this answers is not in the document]
the-krg: Whoa, great insights! 🚀 

I decided to stop using the hidden attribute and go with the "close" one, provided by the `<dialog>` element :) 
_Those variants are great tho!_

WDYT?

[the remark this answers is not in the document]
the-krg: >Also if I understand this correctly in the presence of multiple modals in the same page the code would be executed by each of them and potentially multiple toggles could happen.

Hmmm 🤔 I don't think I understood. The target is the modal's id, so it wound only trigger the one that is on the `data-ui--modal-modal-id-param` attribute.
Also on the Lookbok's Overview I render 3 different modals on the same page. They all are correctly targeted.

admin/app/components/solidus_admin/ui/modal/component.rb
```diff
@@ -0,0 +1,22 @@
...
+  )
+    @id = id
+    @title = title
+    @close_message = close_message
+    @button_component = button_component
+  end
+
+  def render_cancel_button
+    render @button_component.new(text: @close_message, scheme: :secondary, class: 'sm:mt-0 sm:w-auto', 'data-action': "#{stimulus_id}#toggleModal", "data-#{stimulus_id}-modal-id-param": @id)
+  end
```
rainerdema, line 21: Could we inline the close button rendering in the `component.html.erb` file instead of using a separate method?
Given its simplicity, keeping it in the main template could enhance readability at this stage 🤔

admin/app/components/solidus_admin/ui/modal/component.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class SolidusAdmin::UI::Modal::Component < SolidusAdmin::BaseComponent
+  renders_one :body
+  renders_one :actions
+
+  def initialize(
+    title:,
+    id: SecureRandom.hex,
+    close_message: "Close",
+    button_component: component("ui/button")
```
rainerdema, line 11: In line with the recently merged PR #5372, we should now avoid injecting components through the initializer.

admin/app/components/solidus_admin/ui/modal/component.html.erb
```diff
@@ -0,0 +1,33 @@
...
+      <div class="relative transform overflow-hidden rounded-lg bg-white text-left shadow-xl sm:my-8 sm:w-full sm:max-w-lg divide-y">
+        <div class="flex justify-between pb-4 pt-5 sm:p-6 sm:pb-4">
+          <div class="mt-3 text-center sm:mt-0 sm:text-left">
+            <h3 class="text-xl font-semibold leading-6 text-gray-900" id="modal-title">
+              <%= @title %>
+            </h3>
+          </div>
+          <button type="button" class="close" data-action=<%= "#{stimulus_id}#toggleModal" %> aria-label="Close" <%= "data-#{stimulus_id}-modal-id-param=#{@id}" %>>
+            <span aria-hidden="true">&times;</span>
+          </button>
```
rainerdema, line 15: Why not use a `component('ui/button')` for this button?

rainerdema (commented):
Overall, great work! 👏 I've left a few comments 👍

admin/spec/components/previews/solidus_admin/ui/modal/component_preview/overview.html.erb
```diff
@@ -0,0 +1,71 @@
+<div class="flex justify-center items-center h-screen">
+  <table>
+    <tr>
+      <td class="px-3 py-1 text-center">
+        <%= render component('ui/button').new(
+          scheme: :primary,
+          text: 'Modal',
+          'data-action': 'click->ui--modal#toggleModal',
+          'data-controller': 'ui--modal',
```
elia, line 9: We should find a way to avoid reusing the same controller for both the modal and the button opening it, but I think we can fix it later when we start using them.

elia (approved):
Great! I left some more notes, but looks good generally, up to you on how to act on each of them. 

In any case some stuff will be adjusted once we start using it!

admin/app/components/solidus_admin/ui/modal/component.html.erb
the-krg: Unfortunately our button component only allows primary, secondary and ghost "layouts"; the `x` button on the modal does not match any of those.

rainerdema (approved):
👏

PR #5368, opened 2023-09-04
[Admin] Stop depending on DRY.rb

elia:
## Summary

`dry-system` was initially added out of convenience with the shared agreement to review and likely eliminate the dependency down the road. Now that a first version has been released for public testing the time has come to clean up before moving to new sections of the admin.

The CI failure should be fixed by https://github.com/solidusio/solidus/pull/5369

## Checklist


rainerdema (approved):
👍

waiting-for-dev (commented):
:+1: removing a dependency, but there's no feature parity yet. Some considerations that are worth to be highlighted:

- The new admin is losing the ability to automatically register components by creating a matching file in the host application. That can be easily fixed by making changes to the magic Hash registry.
- The logic must be duplicated if the new admin needs to allow objects beyond components to be registered (thinking about the data layer).
- The configuration class is becoming a god object, like in solidus core. It would be nice to split the responsibilities.
- Modifying the main nav items configurations to be an array (by the way, not related to removing `dry-system`) works for the current API. However, it'll penalize if a more advanced interface is desired. Thinking about replacing, modifying, or deleting items, which, for sure, would be great to have to allow customizations.
- If the new admin owns that code, it should test it.

elia:
> 👍 removing a dependency, but there's no feature parity yet. Some considerations that are worth to be highlighted:

Thanks, although I'm not aiming at full feature parity I think it's a good call to check which features we want to bring over or we're ok with leaving behind 🙏 

---

> * The new admin is losing the ability to automatically register components by creating a matching file in the host application. That can be easily fixed by making changes to the magic Hash registry.

I think this is ok for now, the registry is mainly useful for replacing components provided by the library with custom versions, that can be done explicitly by setting a key/value pair in the registry. Inside a host application other local components can be referenced directly be their own class. 

> * The logic must be duplicated if the new admin needs to allow objects beyond components to be registered (thinking about the data layer).

I think it's good to keep the component registry and a potential data-layer separated, anyway as long as components are able to receive data from outside we can shuffle around the source of that data without any impact on the components/view-layer.

> * The configuration class is becoming a god object, like in solidus core. It would be nice to split the responsibilities.

Might make sense to have more configuration layers over time, right now, despite the increased file size it only added a configuration point for components, which being a configuration should live there. But I see the point and we should start grouping some configurations (e.g. authentication is comprised of 4 preferences that can be grouped).

> * Modifying the main nav items configurations to be an array (by the way, not related to removing `dry-system`) works for the current API. However, it'll penalize if a more advanced interface is desired. Thinking about replacing, modifying, or deleting items, which, for sure, would be great to have to allow customizations.

So far the array worked well for the backend and the only feature available on the wrapper class was `#add`, eventually we can add methods for manipulating it and/or move it further inside (e.g. `Config.menu.items # => […]` combined with `Config.menu.add(…)` etc.)

> * If the new admin owns that code, it should test it.

Yes! I'm adding the missing specs 👍 thanks for pointing that out

PR #5373, opened 2023-09-06
[Admin] Add `products#show`

elia:
## Summary

There are still a few rough edges, but we can go from here. We're keeping track of lacking or broken things in an internal GH project (e.g. we'll build a component that will replace the original `select2` behavior).

The `Manage …` links are pointing to sections in the original admin that still need to be ported to the new UI, e.g. adding a media uploading component (which is better handled in a separate PR).

image
image

## Checklist


admin/app/components/solidus_admin/products/show/component.html.erb
```diff
@@ -0,0 +1,110 @@
...
+      ) %>
+      <%= render component('ui/button').new(
+        tag: :button,
+        text: t('.save'),
+        form: form_id,
+      ) %>
+    </div>
+  </header>
+
+  <%= form_for @product, url: product_path(@product), html: { id: form_id } do |f| %>
```
waiting-for-dev, line 38: FYI, we added the product component to lay down the architecture to define forms from the outside, which would later enable a data layer on top of it. Please see https://github.com/solidusio/solidus/pull/5360

tvdeyen:
Wondering if this should be `product#edit` instead, since we display forms? 🤔

elia:
> Wondering if this should be `product#edit` instead, since we display forms? 🤔

@tvdeyen yes, that's a good call, I did some research on how REST conceptualizes this stuff because I was torn between the this page being the only representation we have for the resource in the admin (there's no traditional `show` page) and this containing a form to modify the resource. I came out of that with the conclusion that `/edit` is an _alternative_ representation of the same resource and that having no main representation (i.e. `show` page) that would become the main one.

With this in mind there are other considerations:
- do we want to stick to the legacy solidus_backend url?
- should we use `get "/product/:id", as: :edit` so only the url changes?

and so on and so forth.

I'm still conflicted so I'd like to hear from you, given the context above. 
Renaming at this stage is cheap anyway ☺️

rainerdema (approved):
🌟 Amazing work! 🌟

tvdeyen:
That are good arguments. What is the point of displaying a resource in an admin area anyway, right? And if we also consider authorization into account it makes even more sense to use the show route as it might be the case an unauthorized user only sees static values anyway. 

If we go with show routes, what about using it for all ongoing routes as well? So that we at least have a pattern here?

elia:
> That are good arguments. What is the point of displaying a resource in an admin area anyway, right? And if we also consider authorization into account it makes even more sense to use the show route as it might be the case an unauthorized user only sees static values anyway. If we go with show routes, what about using it for all ongoing routes as well? So that we at least have a pattern here?

For sure, we'll head that way for all the routes of the new dashboard and ensure that you can go back and forth during the transition time 👍

rainerdema (approved):

PR #5413, opened 2023-10-04
Use bundler gem tasks instead of gem_release

elia:
## Summary

Now that we also have a solidus_admin that follows a different release track we need more control over what gems are released. `gem_release` has no such control available and simply has a configuration for recursing. Also all the features of `gem_relerase` regarding version numbers manipulation are no longer relevant as it all happens within GitHub actions using dev_tools.

Bundler gem tasks are now the standard for all our extensions and are a simpler, already available, tool that can be leveraged here as well.

image

## Checklist

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'bundler'
+require 'bundler/gem_tasks'
 
 task default: :spec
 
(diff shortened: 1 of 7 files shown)


Rakefile
```diff
@@ -53,55 +53,17 @@ task :clean do
...
-  desc "Install all solidus gems"
-  task install: :build do
-    for_each_gem do |gem_path|
-      Bundler.with_clean_env do
-        sh "gem install #{gem_path}"
-      end
-    end
-  end
-
-  desc "Release all gems to rubygems"
```
kennyadsl, line 98: I was under the impression that all of this code was not used at all by `gem-release`. Instead, this was the previous system used to release (see https://github.com/solidusio/solidus/pull/83). I think the subgems release was managed by the `recurse` option set in the `gem-release` config.

That said, I'm in favor of removing the dependency and simplifying the process but I have two concerns:

1. I suspect with the proposed changes we are not releasing the main `solidus` gem but only the subgems.
2. Having two different ways of releasing for new/older versions might create complications. Is there a way to uniform the two? Both having the same interface or backporting is ok for me, but if you have other ideas, they are welcome.

elia: Thanks, yes the recurse option is what got in the way with solidus_admin.

1. Good catch! Fixed 👍 
2. Added back ports labels for all supported versions

rainerdema (approved):

kennyadsl (approved):
👍

PR #5418, opened 2023-10-04
Call empty only on unshipped orders

### LINKED ISSUE #4438: [API] empty cart endpoint raises 500 error with shipped order
**Solidus Version**:
all, including master

**How to reproduce**
Trying to `empty` a shipped order via API raises a 500 error, which is something we should try to avoid.

This can be easily demonstrated in the admin area:
* on the `Cart` tab, add an item to a new order, and keep this browser window as it is, we'll get back to it later;
* add the address, payment, and finalize the order in a different browser window/tab;
* ship the order;
* go back to the `Cart` browser window and click the `Empty Cart` button. Web developer tools will show a 500 error, and no message is presented to the user.

**Context**

This is the method definition from the [API orders controller](https://github.com/solidusio/solidus/blob/master/api/app/controllers/spree/api/orders_controller.rb):
```ruby
      def empty
        authorize! :update, @order, order_token
        @order.empty!
        render plain: nil, status: 204
      end
```

`@order.empty!` [tries to reset the cart](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/order.rb#L450) by destroying the associated records, but shipped shipments [cannot be destroyed](https://github.com/solidusio/solidus/blob/master/core/app/models/spree/shipment.rb#L340). The error bubbles up to the controller, where it's not handled. 
Screen Shot 2022-06-22 at 11 18 08 AM

nirnaeth:
## Summary

Fixes #4438.

The change triggers `@order.empty!` only when the shipping status is false.
The 500 is not triggered from the FE, but only from the BE. Nevertheless I tested manually for both cases.

**Notes:** 
1. I was not able to redirect to the Shipping action (`admin/order/xxxxxx/edit`), directions are welcome.
2. The test is failing on my local environment on interaction with the alert for confirming the emptying action. I am also not sure if that is the best location for the test itself. Please advise :) 

## Checklist
- [x] II have added tests

### DIFF
--- a/api/app/controllers/spree/api/orders_controller.rb
+++ b/api/app/controllers/spree/api/orders_controller.rb
@@ -46,7 +46,9 @@ module Spree
 
       def empty
         authorize! :update, @order, order_token
-        @order.empty!
+
+        @order.empty! unless @order.shipped?
+
         respond_with(@order, default_template: :show)
       end
 
--- a/backend/spec/features/admin/orders/order_details_spec.rb
+++ b/backend/spec/features/admin/orders/order_details_spec.rb
@@ -399,6 +399,33 @@ describe "Order Details", type: :feature, js: true do
               expect(page).not_to have_selector('.fa-arrows-h')
               expect(page).not_to have_selector('.fa-trash')
             end
+
+            context 'and on the cart page the cart is emptied' do
+              it 'should show the empty cart page' do
+                order = create(:order_ready_to_ship)
+
+                visit spree.cart_admin_order_path(order)
+                order.fulfill!
+
+                ## simulate shipping order in another tab
+                shipment = order.shipments.first
+                order.shipping.ship(
+                  inventory_units: shipment.inventory_units,
+                  stock_location: shipment.stock_location,
+                  address: order.ship_address,
+                  shipping_method: shipment.shipping_method
+                )
+
+                click_on 'Empty Cart'
+
+                accept_alert 'Are you sure you want to delete this record?' do
+                  click_icon :ok
+                end
+
+                expect(page.current_path).to eq(spree.cart_admin_order_path(order))
+                expect(page).not_to have_content(order.line_items.first.variant.sku)
+              end
+            end
           end
         end
 


nirnaeth:
by looking at the issue, I think this is more a case of an accident. You happen to have 2 tabs open and you complete the order in one. I don't think it's something really intentional.

happy to close the PR if you feel this is a non-problem or change the PR according to new specs :)

api/app/controllers/spree/api/orders_controller.rb
```diff
@@ -46,7 +46,9 @@ def create
 
       def empty
         authorize! :update, @order, order_token
-        @order.empty!
+
+        @order.empty! unless @order.shipped?
```
kennyadsl, line 50:
```suggestion
        @order.empty! unless @order.completed?
```

kennyadsl:
@nirnaeth my comment was a bit twisted actually, this is what I meant: https://github.com/solidusio/solidus/pull/5418/files#r1348457754. 

I don't think anyone wants to empty a completed order, as they don't want to empty a shipped order. Does it make sense?

spaghetticode:
@nirnaeth thanks for working on this! ❤️ 

I've checked out the code, now the API call succeeds with a 200 and the order is not emptied, so no 500 error anymore 🎉. Unfortunately, there's no error message shown and cart items disappear from the page, just as if it was emptied for real, so this may be confusing for users. For this reason, I'm thinking a more clear approach may be to fail with a 422 HTTP error and show an error message to the user, what do you think?

nirnaeth:
> @nirnaeth my comment was a bit twisted actually, this is what I meant: https://github.com/solidusio/solidus/pull/5418/files#r1348457754.
> 
> I don't think anyone wants to empty a completed order, as they don't want to empty a shipped order. Does it make sense?

gotcha, let me see if the change you suggested fix both cases.

nirnaeth:
> @nirnaeth thanks for working on this! ❤️
> 
> I've checked out the code, now the API call succeeds with a 200 and the order is not emptied, so no 500 error anymore 🎉. Unfortunately, there's no error message shown and cart items disappear from the page, just as if it was emptied for real, so this may be confusing for users. For this reason, I'm thinking a more clear approach may be to fail with a 422 HTTP error and show an error message to the user, what do you think?

makes perfect sense! mine was the minimum to get to the goal of not having a 500, but I do agree it makes more sense for UX to have a proper message explaining what's going on. let me see what I can do :)

nvandoorn:
Hey @nirnaeth thanks for the contribution. I pulled down your changes and made a few changes, and then pushed them back up here: https://github.com/nvandoorn/solidus/tree/empty_cart_error_with_shipped_order

Changes are as follows:
1) Call `order.incomplete?` instead of `order.shipped?` as suggested by @kennyadsl 
2) Updated the new feature spec to make it pass.

You're welcome to pull my branch and push here, or I can open a new pull request if you would prefer.

fthobe:
@jarednorman merged in different PR and can be closed

PR #5441, opened 2023-10-17
Add an `orders/cart` component

elia:
## Summary

- add variants
- change quantity
- remove line-items

https://github.com/solidusio/solidus/assets/1051/8267a50e-2d0f-4e1a-814f-4551c82388dc

## Checklist


admin/app/components/solidus_admin/orders/cart/component.js
```diff
@@ -0,0 +1,129 @@
+import { Controller } from "@hotwired/stimulus"
+import { useClickOutside, useDebounce } from "stimulus-use"
+
+const LOADING = "Loading..."
+const INITIAL = "Type to search"
+const EMPTY = "No results"
```
rainerdema, line 6: nit: How about passing these messages as I18n data values?

rainerdema (approved):
🌟 Amazing work! 🌟

admin/app/components/solidus_admin/orders/cart/component.html.erb
```diff
@@ -0,0 +1,115 @@
...
+  data-controller="<%= stimulus_id %>"
+  data-<%= stimulus_id %>-products-url-value="<%= solidus_admin.variants_for_order_path(@order) %>"
+  data-action="
+    keydown-><%= stimulus_id %>#selectResult
+  "
+>
+  <%= render component('ui/panel').new do |panel| %>
+    <div>
+      <div class="peer">
+      <%= render component("ui/forms/search_field").new(
```
tvdeyen, line 12: Are there any plans to make this reusable components? We currently have the `productPicker` and `variantSelect` that can be used outside of the admin cart. Would be great if we could keep this functionality in the future.

admin/app/components/solidus_admin/orders/cart/component.js
tvdeyen: Great catch. Would even call this mandatory and not a nit pick imo ;)

admin/app/components/solidus_admin/orders/cart/component.js
```diff
@@ -0,0 +1,129 @@
...
+const QUERY_KEY = "q[name_or_variants_including_master_sku_cont]"
```
tvdeyen, line 7: In the current admin we are able to change the query in order to reuse the variant select in other areas of the admin. We should make this possible here as well.

admin/app/components/solidus_admin/orders/cart/result/component.js
```diff
@@ -0,0 +1,8 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  submit(event) {
+    debugger
```
tvdeyen, line 5: debugging leftover

admin/app/components/solidus_admin/orders/show/component.html.erb
```diff
@@ -1,9 +1,7 @@
 <%= page do %>
   <%= page_header do %>
-    <%= page_header_back solidus_admin.orders_path %>
-
-    <%= page_header_title t(".create_order") %>
-
+    <%= page_header_back(solidus_admin.orders_path) %>
+    <%= page_header_title("Order #{@order.number}") %>
```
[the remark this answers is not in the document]
tvdeyen: Not a nit, mandatory. Let's not repeat the Spree v2 days, where every PR introduced not translatable strings all over the admin. I18n is a must

admin/app/components/solidus_admin/orders/cart/component.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  async search() {
+    const query = this.query
+
+    if (query) {
+      this.resultsValue = LOADING
+      this.render()
+
+      this.resultsValue = (await (await fetch(`${this.productsUrlValue}?${QUERY_KEY}=${query}`)).text()) || EMPTY
```
tvdeyen, line 73: Is this query configurable somewhere? We need to be able to change it as we are in the current variant select

admin/app/components/solidus_admin/orders/cart/component.html.erb
```diff
@@ -0,0 +1,115 @@
...
+        "data-action": "#{stimulus_id}#search focus->#{stimulus_id}#showResults #{stimulus_id}#showResults",
+        placeholder: t(".search_placeholder"),
+        "data-#{stimulus_id}-target": "searchField",
+      ) %>
+      </div>
+
+      <%# results popover %>
+      <details class="px-6 relative overflow-visible">
+        <summary class="hidden"></summary>
+        <div
```
tvdeyen, line 22: A view component would be great.

tvdeyen (changes requested):
Nice work, but there are some mandatory things to be updated

admin/app/components/solidus_admin/orders/cart/result/component.js
elia: Thanks!  🤦‍♂️ 
We should have a linter for these things 🤔

admin/app/components/solidus_admin/orders/show/component.html.erb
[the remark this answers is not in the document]
elia: Already on it, agree 👍

admin/app/components/solidus_admin/orders/cart/component.js
elia: Currently no, but this is not the only place in which we'll have to fetch products/variants, once we bump into the next one we'll surely get out something more generic and possibly have a `ui/…` component rather than one specific for the `orders/…` pages.

admin/app/components/solidus_admin/orders/cart/component.html.erb
elia: As above, we're driving the component generation/extraction from bottom up, "generalizing" too early means missing what the "right" parameters are, that's a risk also with when you have 2-3 concrete cases to extract from but is much closer to something truly reusable.

The _section comments_ like `# result popover` are a clear hint that you correctly noticed, but we need to hold a little bit more 😊

admin/app/components/solidus_admin/orders/cart/result/component.js
elia: See https://github.com/solidusio/solidus/pull/5457

admin/app/components/solidus_admin/orders/cart/component.html.erb
tvdeyen: > The general plan also relies on the expectation that the building blocks we'll end up with at the end of the projects will be good enough and generic enough to be useful for custom interfaces as well. This approach doesn't exclude doing a "final" pass on the components looking for additional extractions or missing pieces but we want to give priority to sourcing them from concrete needs rather than imagining how we'll use them in the future.

Very well thought. Thanks for explaining!

admin/app/components/solidus_admin/orders/cart/component.js
tvdeyen: Makes sense.

admin/app/components/solidus_admin/orders/cart/component.html.erb
tvdeyen: This makes total sense.

tvdeyen (approved):

PR #5442, opened 2023-10-17
[Admin] Enhance toast component: Background color and animations

rainerdema:
## Summary

This PR brings two enhancements to the toast UI component:
1. Updated the default background from `bg-gray-800` to `bg-full-black` for better visibility.
2. Added smooth slide-in and fade-out animations for a better UX when the toast is displayed or dismissed.

## Checklist

### DIFF
--- a/admin/app/components/solidus_admin/ui/toast/component.html.erb
+++ b/admin/app/components/solidus_admin/ui/toast/component.html.erb
@@ -14,7 +14,7 @@
 >
   <%= icon_tag(@icon, class: 'w-[1.125rem] h-[1.125rem] mr-2 fill-current') if @icon %>
 
-  <p class="body-tiny-bold"><%= @text %></p>
+  <p class="body-tiny-bold leading-none"><%= @text %></p>
 
   <button
     class="ml-2 align-text-bottom"
--- a/admin/app/components/solidus_admin/ui/toast/component.html.erb
+++ b/admin/app/components/solidus_admin/ui/toast/component.html.erb
@@ -1,5 +1,6 @@
 <div
   class="
+    flex items-center justify-between
     rounded inline-block px-3 py-2
     <%= SCHEMES.fetch(@scheme.to_sym).join(' ') %>
   "
@@ -10,12 +11,12 @@
   aria-label="<%= t(".#{@scheme}_label") %>"
   aria-live="polite"
 >
-  <%= icon_tag(@icon, class: 'inline-block w-[1.125rem] h-[1.125rem] mr-3 fill-current') if @icon %>
+  <%= icon_tag(@icon, class: 'w-[1.125rem] h-[1.125rem] mr-2 fill-current') if @icon %>
 
-  <p class="inline-block body-tiny-bold"><%= @text %></p>
+  <p class="body-tiny-bold"><%= @text %></p>
 
   <button
-    class="inline-block ml-3 align-text-bottom"
+    class="ml-2 align-text-bottom"
     title="<%= t('.close_text') %>"
     data-action="<%= stimulus_id %>#close"
     aria-label="<%= t('.close_text') %>"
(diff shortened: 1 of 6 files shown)


admin/app/components/solidus_admin/ui/toast/component.rb
```diff
@@ -3,16 +3,19 @@
 class SolidusAdmin::UI::Toast::Component < SolidusAdmin::BaseComponent
   SCHEMES = {
     default: %w[
-      bg-gray-800 text-white
+      bg-black text-white
```
mfrecchiami, line 6: @rainerdema the class `xxx-black` corresponds to `rgb(34, 34, 34)`, HEX color `#222222` which is the color used for the primary button, for example. I've added the full black `#000000` to fill the toast background. I think is worth adding, probably we'll need it in the future when we'll develop the dark/light theme mode.
We could call it `full-black` or `pure-black`, and it can stay under the [Secondary palette](https://github.com/solidusio/solidus/blob/main/admin/config/solidus_admin/tailwind.config.js.erb#L28).

Let me know your thoughts on it!

admin/app/components/solidus_admin/ui/toast/component.js
```diff
@@ -2,16 +2,18 @@ import { Controller } from '@hotwired/stimulus'
 
 export default class extends Controller {
   static targets = ['closeButton']
-  static classes = ['closing']
-  static values = { transition: Number }
+  static values = { transition: Number, animationClasses: String }
```
elia, line 5: Stimulus provides a specific set of attributes for classes, better to use that instead of values:

https://stimulus.hotwired.dev/reference/css-classes#definitions

admin/app/components/solidus_admin/ui/toast/component.html.erb
```diff
@@ -1,11 +1,13 @@
 <div
   class="
     rounded inline-block px-3 py-2
+    transform translate-y-full opacity-0 transition-all duration-500
     <%= SCHEMES.fetch(@scheme.to_sym).join(' ') %>
+    <%= @animation_classes %>
```
elia, line 6: Any reason for not using the constant directly? (here and below)

admin/app/components/solidus_admin/ui/toast/component.rb
```diff
@@ -10,9 +10,12 @@ class SolidusAdmin::UI::Toast::Component < SolidusAdmin::BaseComponent
     ],
   }
 
+  ANIMATION_CLASSES = %w[translate-y-full opacity-0]
```
elia, line 13: Although we did it in other components I think it's better to switch to strings instead of arrays whenever we can so we don't put unnecessary pressure on memory and CPU

elia (approved):

admin/app/components/solidus_admin/ui/toast/component.rb
```diff
@@ -3,7 +3,7 @@
 class SolidusAdmin::UI::Toast::Component < SolidusAdmin::BaseComponent
   SCHEMES = {
     default: %w[
-      bg-gray-800 text-white
+      bg-fullBlack text-white
```
mfrecchiami, line 6: @rainerdema which naming convention do we want to use for color variables? 
Probably we can safely use hyphens instead of camelCase. I don't think there'll be another fullBlack i.e. `fullBlack-100`. What are your thoughts about it?

rainerdema: Hey @mfrecchiami, I initially used camelCase for 'fullBlack' because I noticed other similar colors in the Tailwind configuration were formatted this way (likely to avoid the need for wrapping them in quotes 🤔).
However, based on your feedback, I've now changed all the colors to kebab-case in the latest commit.
Let me know WDYT 👍

mfrecchiami (approved):
Thanks for working on this @rainerdema 🚀

admin/app/components/solidus_admin/ui/toast/component.rb
mfrecchiami: Awesome! Yes, Tailwind uses the camelCase format but the output is a classname in kebab-case. Thanks for this update ❤️

PR #5459, opened 2023-10-25
Update pinned chrome version

tvdeyen:
## Summary

The current version does not exist anymore. We need to update it.

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/.github/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [x] I have written a thorough PR description.
- [x] I have kept my commits small and atomic.
- [x] [I have used clear, explanatory commit messages](https://github.com/solidusio/.github/blob/main/CONTRIBUTING.md#writing-good-commit-messages).

The following are not always needed:

- 📖 I have updated the README to account for my changes.
- 📑 I have documented new code [with YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md).
- 🛣️ I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).
- ✅ I have added automated tests to cover my changes.
- 📸 I have attached screenshots to demo visual changes.

PR #5463, opened 2023-10-27
[4.1] Update pinned chrome version

tvdeyen:
# Backport

This will backport the following commits from `main` to `v4.1`:
 - [Merge pull request #5459 from tvdeyen/revert-5367](https://github.com/solidusio/solidus/pull/5459)

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  browser-tools: circleci/browser-tools@1.3
+  browser-tools: circleci/browser-tools@1.4.6
   codecov: codecov/codecov@3.2.3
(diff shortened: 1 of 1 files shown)


tvdeyen:
The spec only fails because we cannot install Solidus 4.2 in a Solidus 4.0 branch. Not sure why we test this in all branches. Does not make sense to me. Merging anyhow, because currently all backport PRs fail becaus of this.

Also, not sure what the issue with rubocop is. The same branch passes locally. It only fails on CI. Maybe caching issues? Since this change has not touched and Ruby code we can simply ignore this.

PR #5380, opened 2023-09-13
Adjust after rubocop changes

elia:
## Summary

Rubocop 1.56 was released introducing some changes around `Rails/FindEach`.

Most changes are in specs and not-critical code, the few changes in core models seem innocuous or even improvements for some edge cases (e.g. a store with thousands of stock locations, which is admittedly unlikely anyway).

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/.github/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [ ] I have written a thorough PR description.
- [ ] I have kept my commits small and atomic.
- [ ] [I have used clear, explanatory commit messages](https://github.com/solidusio/.github/blob/main/CONTRIBUTING.md#writing-good-commit-messages).

The following are not always needed:

- 📖 I have updated the README to account for my changes.
- 📑 I have documented new code [with YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md).
- 🛣️ I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).
- ✅ I have added automated tests to cover my changes.
- 📸 I have attached screenshots to demo visual changes.

PR #5477, opened 2023-11-01
CI: Skip install_solidus check on non-current version

tvdeyen:
## Summary

The installer always uses the current released minor version (4.2 time writing this), but the check runs on all pipeline runs, wasting CI time and precious resources. Also it fails anyway, because of conflicting dependencies.

So, we only run the check on main and latest minor now.

## Checklist

Check out our [PR guidelines](https://github.com/solidusio/.github/blob/master/CONTRIBUTING.md#pull-request-guidelines) for more details.

The following are mandatory for all PRs:

- [x] I have written a thorough PR description.
- [x] I have kept my commits small and atomic.
- [x] [I have used clear, explanatory commit messages](https://github.com/solidusio/.github/blob/main/CONTRIBUTING.md#writing-good-commit-messages).

The following are not always needed:

- 📖 I have updated the README to account for my changes.
- 📑 I have documented new code [with YARD](https://www.rubydoc.info/gems/yard/file/docs/Tags.md).
- 🛣️ I have opened a PR to update the [guides](https://github.com/solidusio/edgeguides).
- ✅ I have added automated tests to cover my changes.
- 📸 I have attached screenshots to demo visual changes.

PR #5484, opened 2023-11-02
[4.1] Use ActiveRecord's `.find_each` instead of `.each` whenever possible

tvdeyen:
## Summary

Backport of https://github.com/solidusio/solidus/pull/5380/ for v4.1

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -339,3 +339,7 @@ Lint/MissingSuper:
   Exclude:
     - 'core/lib/spree/deprecation.rb' # this is a known class that doesn't require super
     - 'core/lib/spree/preferences/configuration.rb' # this class has no superclass defining `self.inherited`
+
+Rails/FindEach:
+  Exclude:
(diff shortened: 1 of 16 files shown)


elia (approved):
👍👍👍 
pending the installer specs, tbh I'm not sure why the rackup executable is causing trouble, maybe there's a `-y` or `-f` that we can add to overcome the problem.

tvdeyen:
> pending the installer specs

@elia we would need to backport https://github.com/solidusio/solidus/pull/5477 to v4.1 (and 4.1, 4.2 and 3.4 tbh), because all those branches test the installation of latest solidus, what does not make any sense

elia:
@tvdeyen I'm down with backporting it, I think the backport labels will trigger the gh action even after merge 👍

tvdeyen:
That's good to know. @kennyadsl agree?

kennyadsl:
Works for me!

PR #5498, opened 2023-11-02
Use configurable promo adjuster in callback

mamhoff:
## Summary

This deprecates `Spree::Adjustment#calculate_eligibility` by refactoring `Spree::Order#ensure_promotions_eligible` to use the configurable promotion adjuster rather than `Spree::Adjustment#calculate_eligibility`.

## Checklist

- ✅ I have added automated tests to cover my changes. (Really I didn't add any because exisiting specs would catch misbehavior).

### DIFF
--- a/core/app/models/spree/order.rb
+++ b/core/app/models/spree/order.rb
@@ -788,10 +788,9 @@ module Spree
     end
 
     def ensure_promotions_eligible
-      adjustment_changed = all_adjustments.eligible.promotion.any? do |adjustment|
-        !adjustment.calculate_eligibility
-      end
-      if adjustment_changed
+      Spree::Config.promotion_adjuster_class.new(self).call
+
+      if promo_total_changed?
         restart_checkout_flow
         recalculate
         errors.add(:base, I18n.t('spree.promotion_total_changed_before_complete'))
--- a/core/app/models/spree/adjustment.rb
+++ b/core/app/models/spree/adjustment.rb
@@ -132,6 +132,7 @@ module Spree
         eligible?
       end
     end
+    deprecate :calculate_eligibility, deprecator: Spree.deprecator
 
     private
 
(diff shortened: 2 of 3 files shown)


jarednorman (approved):

adammathys (approved):

core/app/models/spree/order.rb
```diff
@@ -788,10 +788,9 @@ def ensure_inventory_units
     end
 
     def ensure_promotions_eligible
-      adjustment_changed = all_adjustments.eligible.promotion.any? do |adjustment|
-        !adjustment.calculate_eligibility
-      end
-      if adjustment_changed
+      Spree::Config.promotion_adjuster_class.new(self).call
```
kennyadsl, line 791: What [this class](https://github.com/solidusio/solidus/blob/00031a2c1844e881a08128847289924573bb6b48/core/app/models/spree/promotion/order_adjustments_recalculator.rb#L17-L41) does seems pretty different from the `calculate_eligibility` method's content. Can you help me understand in which way they are equivalent in this context?

mamhoff: It does check the eligibility of each adjustment in `recalculate(adjustment)` and especially [`calculate_eligibility(adjustment)`](https://github.com/solidusio/solidus/blob/00031a2c1844e881a08128847289924573bb6b48/core/app/models/spree/promotion/order_adjustments_recalculator.rb#L83-L89). The class does a little bit more - it calculates not only eligibility, but also checks whether the adjustment's sum of amount is still correct. So it actually checks the promo total that the error message indicates it should have checked.

kennyadsl: I got it; I am a bit worried about degrading the performance of this method with the call to the `promotion_adjuster_class` because I guess (I didn't check) it will hit the database many more times. Considering that it only happens on the order completion, it might be a good trade-off. If you or others are not concerned about that, I'm okay with merging this change.

mamhoff: There's a slight performance degradation for sure. From my experience with the promotion system though, the expensive part is checking the rules for eligibility, and not (re)calculating the discounts. 
Also: I fully agree that it's a tradeoff, and that it's okay because it only happens on order completion.

